PR #4327, opened 2022-07-19
layers: Make Device version less than instance an INFO message

charles-lunarg:


ncesario-lunarg (approved):
LGTM (pending @paul-lunarg's approval)!

Are the version macro changes related, or just too tempting to ignore ;) ?

charles-lunarg:
Much too tempting to ignore. 
Its not like its a big deal to have to update the if-chain on minor version releases, but it certainly is nice to have one less place to think about.

Aaaand formatting errors, but of course.

paul-lunarg (approved):
LGTM, it does make sense to not have a warning that can't be fixed... Though you'll need to fix the commit message thats too long before merging

PR #4328, opened 2022-07-19
layers: Use a thread for QUEUE_STATE submission processing

### LINKED ISSUE #4308: False positives after concurrently waiting for a timeline semaphore
Sometimes, after multiple threads wait for (or check the value of) a timeline semaphore, one of the threads gets false positive errors, e.g. `VUID-vkDestroyBuffer-buffer-00922` (*All submitted commands that refer to buffer, either directly or via a `VkBufferView`, **must** have completed execution*).

This issue seems similar to #4271, but this case does not involve a fence, #4281 does not fix it, and it is more about command buffer use counts.

After some debugging, I'm pretty sure this is what's happening.
Let there be a command buffer `CB`, a buffer `Buffer` and a timeline semaphore `Sema`.
Some thread submits work:
- `vkCmdDoSomething(CB, Buffer, ...)`
- `vkQueueSubmit(CB, signalSemaphore=Sema, signalSemaphoreValue=10)`

Then two threads observe the semaphore's value and act on it (I might be wrong about the exact order of internal operations and locking):

| T0 | T1 |
|-|-|
| `vkGetSemaphoreCounterValue(Sema) == 10` | `vkGetSemaphoreCounterValue(Sema) == 10` |
| > `SEMAPHORE_STATE::RetireTimeline(Sema, 10)` | > `SEMAPHORE_STATE::RetireTimeline(Sema, 10)` |
| > > `SEMAPHORE_STATE::Retire` (locks `SEMAPHORE_STATE` first) | |
| > > > remove and return operations to retire | |
| | > > `SEMAPHORE_STATE::Retire` (locks `SEMAPHORE_STATE` second) |
| | > > > remove and return nothing |
| | > (return from `vkGetSemaphoreCounterValue`) |
| | `vkDestroyBuffer(Buffer)` |
| | > check `BUFFER_STATE::InUse()` |
| | > > check `CMD_BUFFER_STATE::InUse() == true` |
| | > **ERROR** `VUID-vkDestroyBuffer-buffer-00922` |
| > > (without the lock) actually retire the returned operations | |
| > > > `CMD_BUFFER_STATE::Retire()` and `CMD_BUFFER_STATE::EndUse()` | |

I haven't checked this, but judging by the code, the same error could occur if two threads wait for the same fence or for a fence and semaphore signalled by the same operation.

I attempted to fix the issue by locking `SEMAPHORE_STATE`'s lock for the entire duration of `RetireTimeline()`. But even after replacing the `shared_mutex` with a `recursive_mutex`, deadlocks still occurred between threads locking `SEMAPHORE_STATE` and `QUEUE_STATE` in different orders (at least that's what it looked like).

**Environment:**
 - OS: Windows 10
 - GPU: NVIDIA RTX 2080
 - Checked with SDK 1.3.216.0, the current `master` @ adf933c0, and with #4281 applied
 - Checked with and without `GPU_ASSISTED` and `SYNCHRONIZATION_VALIDATION` enabled

jeremyg-lunarg:
Rewrite QUEUE_STATE to use a worker thread to update state. This was necessary to fix a race condtion if multiple threads wait on the same timeline semaphore.
    
    Thread 1:
    vkWaitSemaphores() : wait on semaphore for value N
    
    Thread 2:
    vkQueueSubmit()    : signal semaphore for value N
    vkWaitSemaphores() : wait on semaphore for value N
    
Thread 1 starts waiting before we have any information about which queues(s) will need to make progress before value N will be complete. In the old implementation, that might cause it to not correctly update the queue(s) state after it woke up. That would then lead to false 'object in use' errors. In theory, it might have been possible to fix this problem, but the logic was already confusing. Also, this opens up the possiblity of moving execute time validation and state updates to the queue thread, which might improve performance.
 
Fixes #4308

### DIFF
--- a/layers/queue_state.cpp
+++ b/layers/queue_state.cpp
@@ -30,6 +30,43 @@
 
 using SemOp = SEMAPHORE_STATE::SemOp;
 
+// This timeout is for all queue threads to update their state after we know
+// (via being in a PostRecord call) that a fence, semaphore or wait for idle has
+// completed. Hitting it is almost a certainly a bug in this code.
+static std::chrono::time_point<std::chrono::steady_clock> GetCondWaitTimeout() {
+    return std::chrono::steady_clock::now() + std::chrono::seconds(10);
+}
+
+void CB_SUBMISSION::BeginUse() {
+    for (auto &wait : wait_semaphores) {
+        wait.semaphore->BeginUse();
+    }
+    for (auto &cb_node : cbs) {
+        cb_node->BeginUse();
+    }
+    for (auto &signal : signal_semaphores) {
+        signal.semaphore->BeginUse();
+    }
+    if (fence) {
+        fence->BeginUse();
+    }
+}
+
+void CB_SUBMISSION::EndUse() {
+    for (auto &wait : wait_semaphores) {
+        wait.semaphore->EndUse();
+    }
+    for (auto &cb_node : cbs) {
+        cb_node->EndUse();
+    }
+    for (auto &signal : signal_semaphores) {
+        signal.semaphore->EndUse();
+    }
+    if (fence) {
+        fence->EndUse();
+    }
+}
+
 uint64_t QUEUE_STATE::Submit(CB_SUBMISSION &&submission) {
     for (auto &cb_node : submission.cbs) {
         auto cb_guard = cb_node->WriteLock();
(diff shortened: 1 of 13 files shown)


ncesario-lunarg (approved):
LGTM!

build-android/jni/Application.mk
```diff
@@ -16,6 +16,6 @@
 APP_ABI := armeabi-v7a arm64-v8a x86 x86_64
 # APP_ABI := arm64-v8a   # just build for pixel2  (don't check in)
 APP_PLATFORM := android-26
-APP_STL := c++_static
+APP_STL := c++_shared
```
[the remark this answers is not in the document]
jeremyg-lunarg: it is a separate commit:

> build-android: Use c++_shared
> https://developer.android.com/ndk/guides/cpp-support recommends using
> c++_shared for applications that use more than one shared library.
> If multiple libraries using c++_static are loaded you end up with
> several copies of the globals in the C++ runtime. This also happens
> if the same library is dlopen/dlclosed several times. Some of the
> c++ runtime globals are thread_local, so each copy consumes a
> TLS key. There are only 128 TLS keys allowed on android, and the
> unit tests can hit this because of repeatedly loading and unloading
> VVL.
> 
> See https://github.com/android/ndk/issues/789 and the many issues
> linked to it for more info.

This was necessary to get CI to pass on android with this change.

layers/queue_state.h
```diff
@@ -247,30 +259,73 @@ struct CB_SUBMISSION {
...
+
+    ~QUEUE_STATE() { Destroy(); }
+    void Destroy() override;
 
     VkQueue Queue() const { return handle_.Cast<VkQueue>(); }
 
     uint64_t Submit(CB_SUBMISSION &&submission);
 
-    bool HasWait(VkSemaphore semaphore, VkFence fence) const;
+    // Tell the queue thread that work up to until_seq has finished
```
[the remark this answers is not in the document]
jeremyg-lunarg: I reworded this comment, does the new one make more sense?

layers/queue_state.h
```diff
@@ -247,30 +259,73 @@ struct CB_SUBMISSION {
...
-    std::deque<CB_SUBMISSION> submissions_;
-    uint64_t seq_;
-    mutable ReadWriteLock lock_;
+    using LockGuard = std::unique_lock<std::mutex>;
+    void ThreadFunc();
+
+    // state related to submitting to the queue, all data members must
+    // be accessed with SubmitState::lock_ held
+    class SubmitState {
+      public:
```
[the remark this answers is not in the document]
jeremyg-lunarg: had to touch this header so changed them.

paul-lunarg (approved):
LGTM

layers/cmd_buffer_state.cpp
```diff
@@ -441,6 +441,11 @@ void CMD_BUFFER_STATE::ResetPushConstantDataIfIncompatible(const PIPELINE_LAYOUT
 }
 
 void CMD_BUFFER_STATE::Destroy() {
+    BASE_NODE::Destroy();
```
jzulauf-lunarg, line 444: Isn't destruction unwound child-then-base usually?

build-android/jni/Application.mk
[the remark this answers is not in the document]
ncesario-lunarg: 👍 thanks, I should have looked at the individual commits :)

layers/queue_state.h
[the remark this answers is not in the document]
ncesario-lunarg: 👍 yep, thanks!

layers/cmd_buffer_state.cpp
jeremyg-lunarg: Yeah. This is a leftover from when I was beating my head against the android c++ runtime problem. I can try undoing it. It really shouldn't be needed.

jeremyg-lunarg: I mean, the lock is needed, but the original ordering should work.

jeremyg-lunarg:
This PR has multiple problems running CTS timeline_semaphore tests. I'm going to pull out some of the commits into individual PRs to get them out of the way.

layers/cmd_buffer_state.cpp
jeremyg-lunarg: This is moved to https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4407

layers/queue_state.cpp
```diff
@@ -30,6 +30,43 @@
 
 using SemOp = SEMAPHORE_STATE::SemOp;
 
+// This timeout is for all queue threads to update their state after we know
+// (via being a PostRecord call) that a fence, semaphore or wait for idle has
```
ncesario-lunarg, line 34: ~~being~~?

layers/queue_state.cpp
```diff
@@ -38,138 +75,151 @@ uint64_t QUEUE_STATE::Submit(CB_SUBMISSION &&submission) {
...
-                return true;
-            }
-        }
+std::shared_future<void> QUEUE_STATE::Wait(uint64_t until_seq) {
+    auto guard = Lock();
+    if (until_seq == UINT64_MAX) {
+        until_seq = seq_;
+    }
+    if (submissions_.empty() || until_seq < submissions_.begin()->seq) {
+        std::promise<void> already_done;
```
ncesario-lunarg, line 117: I was under the impression that the promise's lifetime needs to encapsulate the lifetime of associated futures? Does this work because `set_value` is called here before the function returns?

Asking for my own education :)

layers/queue_state.cpp
```diff
@@ -38,138 +75,151 @@ uint64_t QUEUE_STATE::Submit(CB_SUBMISSION &&submission) {
...
-    return false;
 }
 
-static void MergeResults(SEMAPHORE_STATE::RetireResult &results, const SEMAPHORE_STATE::RetireResult &sem_result) {
-    for (auto &entry : sem_result) {
-        auto &last_seq = results[entry.first];
-        last_seq = std::max(last_seq, entry.second);
+uint64_t QUEUE_STATE::Notify(uint64_t until_seq) {
+    auto guard = Lock();
+    if (until_seq == UINT64_MAX) {
```
ncesario-lunarg, line 142: Not recommending for this PR, but we still have no consistency with the `*_MAX` macros. Maybe something like the following could be used in general?
```cpp
template<typename T>
constexpr T MaxType(T) { return std::numeric_limits<T>::max(); }
```
So this would look like
```cpp
if (until_seq == layer_data::MaxType(until_seq)) {
```

ncesario-lunarg (approved):
This looks scary...but awesome :)

Tony-LunarG (approved):
LGTM but I had to look up promise and future,

mark-lunarg (approved):
🥇

layers/queue_state.cpp
jeremyg-lunarg: There's a shared state for storing the value which both the future(s) and the promise have references to. So it is ok if the completed promise goes away before the future.

jeremyg-lunarg:
> LGTM but I had to look up promise and future,

I started out with condition variables and mutex in the semaphore state, but ran into tons of problems with recursive locking. Promises and futures (which I also had to research to do this change) avoid much of the locking.

layers/queue_state.cpp
jeremyg-lunarg: this is a good idea. std::numeric_limits<uint64_t>::max is an awful lot of typing.

ncesario-lunarg:
> > LGTM but I had to look up promise and future,
> 
> I started out with condition variables and mutex in the semaphore state, but ran into tons of problems with recursive locking. Promises and futures (which I also had to research to do this change) avoid much of the locking.

+1 for futures/promises. Makes everything clearer/easier IMHO.

dj2:
Is this PR intended to find new validation errors, or only fixup issues with the old implementation?

If it's the former, after appling this CL we're seeing timeouts in our tests which do not happen at the revision prior to this one.

I haven't attempted to determine what is causing the timeout (Seeing `VkSemaphore-state-timeout Timeout waiting for timeline semaphore state to update` errors). But, wanted to verify if it was expected that this change could uncover erorrs which were not present previously?

jeremyg-lunarg:
> Is this PR intended to find new validation errors, or only fixup issues with the old implementation?
> 
> If it's the former, after appling this CL we're seeing timeouts in our tests which do not happen at the revision prior to this one.
> 
> I haven't attempted to determine what is causing the timeout (Seeing `VkSemaphore-state-timeout Timeout waiting for timeline semaphore state to update` errors). But, wanted to verify if it was expected that this change could uncover erorrs which were not present previously?

Can you please file an issue with any details of the problem you're seeing? that error indicates a problem with the code in the PR, and it is issued instead of deadlocking.   This PR was mostly intended to fix false positives from the exiting validation.

Issue #4333, opened 2022-07-20
Linux CFI failures caused by recent changes


romanl-g:
**Describe the Issue**
Hi, we ran into CFI failures on Linux in Chromium testing ANGLE, very similarly to this previous report: https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/2825

It started happening as our bots tried rolling these changes:

```
cb13f0f layers: Avoid decoding ranges in AnyInRange() by Jeremy Gebben · 26 hours ago
c08f650 layers: Remove Descriptor type and class members by Jeremy Gebben · 26 hours ago
f4816f7 layers: Move Descriptor::updated to DescriptorBinding by Jeremy Gebben · 26 hours ago
7b963d8 layers: Add back placement new for DescriptorSet allocations by Jeremy Gebben · 26 hours ago
1b9fdb8 layers: Move Descriptor iteration to class DescriptorSet by Jeremy Gebben · 26 hours ago
94b7c2f layers: Use small_vector for storing Descriptors by Jeremy Gebben · 26 hours ago
080210f layers: Store descriptors in per-binding vectors by Jeremy Gebben · 26 hours ago
cd63625 tests: Add VkPositiveLayerTest.FenceSemThreadRace by Jeremy Gebben · 26 hours ago
98faf72 layers: Fix a race condition in FENCE_STATE::Retire() by Jeremy Gebben · 26 hours ago
```

@jeremyg-lunarg

I'm trying to narrow it down further but we're missing symbols for some reason, all I have for now is this error in case that rings any bells:

```
../../third_party/vulkan-deps/vulkan-validation-layers/src/layers/descriptor_sets.h:1042:56: runtime error: control flow integrity check for type 'cvdescriptorset::DescriptorBindingImpl<cvdescriptorset::ImageSamplerDescriptor>' failed during cast to unrelated type (vtable address 0xabababababababab)
0xabababababababab: note: invalid vtable
<memory cannot be printed>
../../third_party/vulkan-deps/vulkan-validation-layers/src/layers/descriptor_sets.h:1042:56: note: check failed in angledata/../libVkLayer_khronos_validation.so, vtable located in (unknown)
```

**Environment:**
 - OS: Linux with CFI checks

jeremyg-lunarg:
Do you know which ANGLE test(s) hit this error?

romanl-g:
We don't have CFI enabled in ANGLE and only caught it in Chromium tests:
[...]

PR #4336, opened 2022-07-20
layers: Revert "Use small_vector for storing Descriptors"

jeremyg-lunarg:
This reverts commit 94b7c2feac812e060041dff0c243b6950e122a78.

It is also causing CFI errors. Based on:

https://clang.llvm.org/docs/ControlFlowIntegrity.html#bad-cast-checking

this is probably only showing up for Descriptors because:

>    The checks are currently only introduced where the class being casted
>    to is a polymorphic class.

Presumably the other objects stored in small_vector are not polymorphic, but Descriptors are.

This is also for Issue #4333

### DIFF
--- a/layers/descriptor_sets.h
+++ b/layers/descriptor_sets.h
@@ -741,8 +741,7 @@ class DescriptorBinding {
     const VkDescriptorBindingFlags binding_flags;
     const uint32_t count;
     const bool has_immutable_samplers;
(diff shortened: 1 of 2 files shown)


Tony-LunarG (approved):
LGTM

ncesario-lunarg (approved):

PR #4338, opened 2022-07-20
tests: Fix fsr framebuffer dimensions test

ziga-lunarg:
If `layeredShadingRateAttachments` is `VK_FALSE` then VUID 04551 will be hit.


Tony-LunarG (commented):
Does this mean we can stop skipping this test on AMD_PROPRIETARY ?  If so, can you remove the skip?

ziga-lunarg:
> Does this mean we can stop skipping this test on AMD_PROPRIETARY ? If so, can you remove the skip?

I did not test, so I left the skip in. I ran into this on linux.

Tony-LunarG (approved):
LGTM

PR #4339, opened 2022-07-21
layers: Fix crash in build acc struct

ziga-lunarg:
Moved checks using `src_as_state` inside `if (src_as_state) {`


jeremyg-lunarg (approved):
LGTM

PR #4126, opened 2022-05-17
layers: Store descriptors in per-binding vectors

jeremyg-lunarg:
Storing descriptors in type specific, per binding vectors saves
memory (and improves cache utilization). Most descriptor objects are
are 48 bytes or less, but MutableDescriptor is almost 200 bytes.
Per binding storage also allows tight inner loops, which helps
performance for large bindless descriptor sets.

PR #4344, opened 2022-07-22
layers: Fix mutable descriptor state active type state and checking

jeremyg-lunarg:
Fix a problem with tracking the active descriptor type, which was introduced by https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4126

Fix the check for VUID 04613, which was a preexisting condition. Improve the test case to show what the problem was.


Tony-LunarG:
Did you mean to check in layers/descriptor_sets.h.old?

jeremyg-lunarg:
> Did you mean to check in layers/descriptor_sets.h.old?

No.  That's a mistake. I'll re push without it.

Tony-LunarG (approved):
LGTM

PR #4348, opened 2022-07-25
layers: Improve VUID 02697 error message

### LINKED ISSUE #4347: Hard to understand grammar mistake in validation error.
I am getting the following validation error:

"Validation Error: [ VUID-vkCmdDraw-None-02697 ] Object 0: handle = 0x89e60f0000000042, name = pipeline [0x555557093790], type = VK_OBJECT_TYPE_PIPELINE; Object 1: handle = 0x2e2cd000000002b, name = m_pipeline_layout2 [0x555557093790], type = VK_OBJECT_TYPE_PIPELINE_LAYOUT; Object 2: handle = 0xba7514000000002a, name = m_pipeline_layout1 [0x555557093790], type = VK_OBJECT_TYPE_PIPELINE_LAYOUT; | MessageID = 0x28d281ae | vkCmdDraw(): VkPipeline 0x89e60f0000000042[pipeline [0x555557093790]] defined with VkPipelineLayout 0x2e2cd000000002b[m_pipeline_layout2 [0x555557093790]] is not compatible for maximum set statically used 2 with bound descriptor sets, last bound with VkPipelineLayout 0xba7514000000002a[m_pipeline_layout1 [0x555557093790]] The Vulkan spec states: For each set n that is statically used by the VkPipeline bound to the pipeline bind point used by this command, a descriptor set must have been bound to n at the same pipeline bind point, with a VkPipelineLayout that is compatible for set n, with the VkPipelineLayout used to create the current VkPipeline, as described in Pipeline Layout Compatibility (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdDraw-None-02697)"

The problematic part here:

    VkPipeline xyz defined with VkPipelineLayout foo is not compatible for maximum set statically used 2 with bound descriptor sets, ...

I believe this should read:

    VkPipeline xyz defined with VkPipelineLayout foo is not compatible for maximum set statically used with 2 bound descriptor sets, ...

although even then I can hardly read it :p. Perhaps some comma's are missing? I'd think it is "compatible with", not "compatible for". So perhaps this should be something like:

    VkPipeline xyz, defined with VkPipelineLayout foo, is not compatible with, for maximum set statically used, 2 bound descriptor sets, ...

At the very least it is extremely hard (read: not) to understand for someone not deep into this material, but it seems that the '2' is in the wrong place - aka, an actual grammar error.

sjfricke:
Closes #4347 

I agree with @CarloWood on the issue that the message isn't good, I purposed this fix that changes it from

> vkCmdDispatch(): VkPipeline 0x967dd1000000000e[] defined with VkPipelineLayout 0xead9370000000008[] is not compatible for maximum set statically used 0 with bound descriptor sets, last bound with VkPipelineLayout 0xe88693000000000c[]

to 

> vkCmdDispatch(): The VkPipeline 0x967dd1000000000e[] (created with VkPipelineLayout 0xead9370000000008[]) maximum descriptor set statically used (index #0) is not compatible with the currently bound descriptor sets' pipeline layout (VkPipelineLayout 0xe88693000000000c[])

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -1547,8 +1547,8 @@ bool CoreChecks::ValidateCmdBufDrawState(const CMD_BUFFER_STATE *cb_node, CMD_TY
         }
         objlist.add(state.pipeline_layout);
         result |= LogError(objlist, vuid.compatible_pipeline,
-                           "%s(): %s defined with %s is not compatible for maximum set statically used %" PRIu32
-                           " with bound descriptor sets, last bound with %s",
+                           "%s(): The %s (created with %s) statically uses descriptor set (index #%" PRIu32
+                           ") which is not compatible with the currently bound descriptor set's pipeline layout (%s)",
                            CommandTypeString(cmd_type), report_data->FormatHandle(pipe->pipeline()).c_str(),
                            pipe_layouts_log.str().c_str(), pipe->max_active_slot,
                            report_data->FormatHandle(state.pipeline_layout).c_str());


layers/core_validation.cpp
```diff
@@ -1547,8 +1547,8 @@ bool CoreChecks::ValidateCmdBufDrawState(const CMD_BUFFER_STATE *cb_node, CMD_TY
         }
         objlist.add(state.pipeline_layout);
         result |= LogError(objlist, vuid.compatible_pipeline,
-                           "%s(): %s defined with %s is not compatible for maximum set statically used %" PRIu32
-                           " with bound descriptor sets, last bound with %s",
+                           "%s(): The %s (created with %s) maximum descriptor set statically used (index #%" PRIu32
+                           ") is not compatible with the currently bound descriptor sets' pipeline layout (%s)",
```
ncesario-lunarg, line 1551: The word `maximum` is a little confusing--maybe even misleading--to me when read along side the VUID text. I would propose a slight change:
```
The {pipeline} (created with {layout}) statically uses descriptor set {max_active_slot} which is not compatible with the currently bound descriptor set's pipeline layout ({state.pipeline_layout}).
```
I think even more ideal is if `CompatForSet` returned an error message giving more information as to _why_ there was an incompatibility. e.g., "only m sets currently bound, but pipeline uses set n > m," or "set n in currently bound descriptor set layout has type a, but  set n in currently bound pipeline has type b which is incompatible."

ncesario-lunarg (approved):
I left a purely subjective comment, but I'm good with the current change as is.

layers/core_validation.cpp
sjfricke: agree :100: and applied suggestions

ncesario-lunarg (approved):

PR #4353, opened 2022-07-26
NVIDIA Best Practices

rlocatti-nv:
This PR adds "Best Practices" validation for NVIDIA devices. I've separated each VUID and test in its own commit, so it should be easier to review per commit instead of viewing the edited files.

There are existing recommendations that align with other vendors, so they are included here as well. For ONE_TIME_SUBMIT we've found it'd be better to check whether the command buffer has been submitted once without using this flag, instead of reusing the existing code. ARM may want the same behavior, but I didn't want to intrusively change their recommendations.

The tests use dynamic rendering (through Vulkan 1.3), so the layer code related to it has been updated to avoid crashes.


scripts/layer_chassis_generator.py
```diff
@@ -304,7 +304,11 @@ class LAYER_PHYS_DEV_PROPERTIES {
 typedef enum ValidationCheckEnables {
     VALIDATION_CHECK_ENABLE_VENDOR_SPECIFIC_ARM,
     VALIDATION_CHECK_ENABLE_VENDOR_SPECIFIC_AMD,
+<<<<<<< HEAD
```
jeremyg-lunarg, line 307: These merge conflict markers are causing you to fail CI. Also, it looks like you may have edited the generated files to remove them. CI will also fail if the files in layers/generated/ do not match the output of the generate_source.py script, so please make sure you check in the new output after fixing this file and rerunning the script.

layers/best_practices_utils.cpp
```diff
@@ -1586,6 +1849,501 @@ void BestPractices::PostCallRecordCmdBindPipeline(VkCommandBuffer commandBuffer,
...
+            VendorSpecificTag(kBPVendorNVIDIA), report_data->FormatHandle(cmd_state->nv.zcull_scope.image).c_str(), good_mode,
+            bad_mode);
+    }
+
+    return skip;
+}
+
+static std::array<uint32_t, 4> GetRawClearColor(VkFormat format, const VkClearColorValue& clear_value) {
+    std::array<uint32_t, 4> raw_color{};
+    memcpy(raw_color.data(), clear_value.uint32, sizeof(raw_color));
```
ncesario-lunarg, line 2219: I'd vote for something with some type information compared to `memcpy` (e.g., `std::copy(std::begin(clear_value.uint32), std::end(clear_value.uint32), raw_color.begin());`). Same comment for other instances of `memcpy` below.

layers/best_practices_utils.cpp
```diff
@@ -1586,6 +1849,501 @@ void BestPractices::PostCallRecordCmdBindPipeline(VkCommandBuffer commandBuffer,
...
+
+    const std::array<uint32_t, 4> raw_color = GetRawClearColor(format, clear_value);
+    if (IsClearColorZeroOrOne(format, raw_color)) {
+        return skip;
+    }
+
+    const auto it = std::find(kCustomClearColorCompressedFormatsNVIDIA.begin(), kCustomClearColorCompressedFormatsNVIDIA.end(), format);
+    if (it == kCustomClearColorCompressedFormatsNVIDIA.end()) {
+        // The format is not compressible
+        std::string format_list = MakeCompressedFormatListNVIDIA();
```
ncesario-lunarg, line 2299: Would it make sense for `format_list` to be `static`?

layers/best_practices_utils.cpp
```diff
@@ -3570,6 +4820,74 @@ bool BestPractices::PreCallValidateCreatePipelineLayout(VkDevice device, const V
...
+                device, kVUID_BestPractices_CreatePipelineLayout_SeparateSampler,
+                "%s Consider using combined image samplers instead of separate samplers for marginally better performance.",
+                VendorSpecificTag(kBPVendorNVIDIA));
+        }
+
+        if (fast_space_usage > kPipelineLayoutFastDescriptorSpaceNVIDIA) {
+            skip |= LogPerformanceWarning(
+                device, kVUID_BestPractices_CreatePipelinesLayout_LargePipelineLayout,
+                "%s Pipeline layout size is too large, prefer using pipeline-specific descriptor set layouts. "
+                "Aim for consuming less than %d bytes to allow fast reads for all non-bindless descriptors. "
```
ncesario-lunarg, line 4882: `%d` -> `PRIu32` (in general)

tests/vklayertests_nvidia_best_practices.cpp
```diff
@@ -0,0 +1,1266 @@
...
+      private:
+        VkDevice device = VK_NULL_HANDLE;
+    };
+}
+
+TEST_F(VkNvidiaBestPracticesLayerTest, PageableDeviceLocalMemory) {
+    InitBestPracticesFramework(kEnableNVIDIAValidation);
+    InitState();
+
+    if (!DeviceExtensionSupported(gpu(), nullptr, VK_EXT_PAGEABLE_DEVICE_LOCAL_MEMORY_EXTENSION_NAME)) {
```
ncesario-lunarg, line 57: In general when adding an extension, it's best to use `AddRequiredExtensions(...)`, as this will add all necessary dependent extensions as well, before calling `InitFramework/InitBestPracticesFramework`. Then, after `InitBestPracticesFramework`, check:
```cpp
    }
```

Also, @paul-lunarg recently added support for the `VK_LAYER_TESTS_DISABLE_CORE_VALIDATION` environment variable. It would probably be good to run these new tests with `VK_LAYER_TESTS_DISABLE_CORE_VALIDATION=false`. That will be the default in an upcoming change, but currently best practices tests do not enable core validation.

tests/vklayertests_nvidia_best_practices.cpp
```diff
@@ -0,0 +1,1266 @@
...
+    }
+
+    VkDeviceQueueCreateInfo queue_ci = { VK_STRUCTURE_TYPE_DEVICE_QUEUE_CREATE_INFO };
+    queue_ci.queueCount = 1;
+
+    VkDeviceCreateInfo device_ci = { VK_STRUCTURE_TYPE_DEVICE_CREATE_INFO };
+    device_ci.queueCreateInfoCount = 1;
+    device_ci.pQueueCreateInfos = &queue_ci;
+
+    VkPhysicalDevicePageableDeviceLocalMemoryFeaturesEXT pageable = { VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PAGEABLE_DEVICE_LOCAL_MEMORY_FEATURES_EXT };
```
ncesario-lunarg, line 69: Use `auto pageable = LvlInitStruct<VkPhysicalDevicePageableDeviceLocalMemoryFeaturesEXT>()`, as this will initialize the `.sType` appropriately and make sure the `.pNext` field is null (and similar comment for all other `VK_STRUCTURE_TYPE_...` initializations).

tests/vklayertests_nvidia_best_practices.cpp
```diff
@@ -0,0 +1,1266 @@
...
+    pageable.pageableDeviceLocalMemory = VK_TRUE;
+
+    {
+        m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_PERFORMANCE_WARNING_BIT_EXT,
+                                             "UNASSIGNED-BestPractices-CreateDevice-PageableDeviceLocalMemory");
+        SafeDevice test_device;
+        vk::CreateDevice(gpu(), &device_ci, nullptr, &test_device);
```
ncesario-lunarg, line 76: Can `vk_testing::Device` be used instead? e.g.,
```cpp
vk_testing::Device test_device(gpu());
test_device.init(device_ci);
```
If not, can `vk_testing::Device` be modified to work here? I'd personally add a constructor so `vk_testing::Device test_device(gpu(), device_ci);` (not sure why that doesn't exist already), but if there are functional reasons it doesn't work, it would be good to address them if we can.

tests/vklayertests_nvidia_best_practices.cpp
```diff
@@ -0,0 +1,1266 @@
...
+        m_errorMonitor->VerifyFound();
+    }
+
+    // Now enable the expected features
+    const char* pageableExtensionName = VK_EXT_PAGEABLE_DEVICE_LOCAL_MEMORY_EXTENSION_NAME;
+    device_ci.enabledExtensionCount = 1;
+    device_ci.ppEnabledExtensionNames = &pageableExtensionName;
```
ncesario-lunarg, line 83: If `AddRequiredExtensions` is used above, you can use `m_device_extension_names` here (which will include any dependent extensions).

tests/vklayertests_nvidia_best_practices.cpp
```diff
@@ -0,0 +1,1266 @@
...
+    image_ci.mipLevels = 1;
+    image_ci.arrayLayers = 1;
+    image_ci.samples = VK_SAMPLE_COUNT_1_BIT;
+    image_ci.tiling = VK_IMAGE_TILING_OPTIMAL;
+    image_ci.usage = VK_IMAGE_USAGE_SAMPLED_BIT;
+
+    {
+        m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_PERFORMANCE_WARNING_BIT_EXT,
+                                             "UNASSIGNED-BestPractices-CreateImage-TilingLinear");
+        VkImage image = VK_NULL_HANDLE;
```
ncesario-lunarg, line 112: In general there are wrappers for non-dispatchable objects under the `vk_testing` namespace. So here you could use `vk_testing::Image image(*m_device, image_ci);` or `vk_testing::Image image(*m_device, image_ci, vk_testing::no_mem)` if you don't want to bind memory to the image. These wrappers are generally preferred as they will ensure objects do not leak, which can lead to sporadic test failures in later tests.

ncesario-lunarg (commented):
his looks really good, thanks for the contribution! I've gone through it relatively quickly due to the size, and just some minor nits on my part and more general comments in the tests (we're in the process of doing some test refactoring/cleanup).

layers/best_practices_utils.cpp
```diff
@@ -647,6 +710,20 @@ void BestPractices::PostCallRecordFreeDescriptorSets(VkDevice device, VkDescript
...
 }
 
+void BestPractices::PreCallRecordAllocateMemory(VkDevice device, const VkMemoryAllocateInfo* pAllocateInfo,
+                                                const VkAllocationCallbacks* pAllocator, VkDeviceMemory* pMemory) {
+    if (VendorCheckEnabled(kBPVendorNVIDIA)) {
+        WriteLockGuard guard{memory_free_events_lock_};
+
+        // Release old allocations to avoid overpopulating the container
+        const auto now = std::chrono::high_resolution_clock::now();
+        const auto last_old = std::find_if(memory_free_events_.rbegin(), memory_free_events_.rend(), [now](const auto& event) {
```
ncesario-lunarg, line 720: As much as I would love to use auto for function parameters, we're stuck on C++11 for now (this only works on Windows because MSVC automatically uses C++17, IIRC), so this will fail on everything except Windows.

layers/best_practices_utils.cpp
```diff
@@ -268,6 +302,15 @@ bool BestPractices::PreCallValidateCreateDevice(VkPhysicalDevice physicalDevice,
             VendorSpecificTag(kBPVendorArm), VendorSpecificTag(kBPVendorAMD), VendorSpecificTag(kBPVendorIMG));
     }
 
+    if (VendorCheckEnabled(kBPVendorNVIDIA) && !enabled_pageable_device_local_memory &&
+        std::find(extensions.begin(), extensions.end(), VK_EXT_PAGEABLE_DEVICE_LOCAL_MEMORY_EXTENSION_NAME) != extensions.end()) {
```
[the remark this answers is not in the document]
rlocatti-nv: The intention is to only signal the message when the extension is available. I thought it'd be annoying for users if they get a message to enable an extension they don't have access to.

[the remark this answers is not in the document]
ncesario-lunarg: Fair enough, though it might be useful information if they don't have access to it as well?

layers/best_practices_utils.cpp
rlocatti-nv: For the other ones I'd need `std::bit_cast`, which is C++20.

rlocatti-nv:
Pushed a new commit addressing some of the feedback. I'll squash it into the individual things once things settle down.

Tony-LunarG:
Do you think you could run through CTS with this enabled? I see an assert running 
dEQP-VK.api.image_clearing.core.clear_depth_stencil_attachment.multiple_layers.s8_uint: 

best_practices_utils.cpp 2018 assert((subresource_range.aspectMask & VK_IMAGE_ASPECT_DEPTH_BIT) != 0U);
The test gets no validation errors running core validation

and then running dEQP-VK.dynamic_rendering.suballocation.attachment.1.13:

best_practices_utils.cpp 3421 const auto& attachment = pColorAttachments[clear_attachment.colorAttachment];
pColorAttachement is NULL
This test does get core validation errors, but unrelated to the code around the crash.

I've run no farther than this test, so I don't know if anything else follows.

rlocatti-nv:
> Do you think you could run through CTS with this enabled? I see an assert running dEQP-VK.api.image_clearing.core.clear_depth_stencil_attachment.multiple_layers.s8_uint:

I did a pass through the CTS. The latest commit should fix the asserts and crashes caused.

Looking into the CI issues now.

tests/vklayertests_nvidia_best_practices.cpp
```diff
@@ -0,0 +1,1139 @@
...
+    sparse_buffer_bind.pBinds = &bind;
+
+    VkBindSparseInfo bind_info = {};
+    bind_info.bufferBindCount = 1;
+    bind_info.pBufferBinds = &sparse_buffer_bind;
+
+    {
+        m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_PERFORMANCE_WARNING_BIT_EXT,
+                                             "UNASSIGNED-BestPractices-QueueBindSparse-NotAsync");
+        vk::QueueBindSparse(transfer_queue, 1, &bind_info, nullptr);
```
ncesario-lunarg, line 227: This is causing a compiler error:
```
error C2664: 'VkResult (VkQueue,uint32_t,const VkBindSparseInfo *,VkFence)': cannot convert argument 4 from 'nullptr' to 'VkFence' [C:\j\vz4\Release32\Vulkan-ValidationLayers\build32\tests\vk_layer_validation_tests.vcxproj]
```

rlocatti-nv:
Any idea why `VkSyncValTest.SyncQSBufferCopyHazards` is running and crashing on Linux builds? I don't see that test running on other PRs.

ncesario-lunarg:
> Any idea why `VkSyncValTest.SyncQSBufferCopyHazards` is running and crashing on Linux builds? I don't see that test running on other PRs.

I believe the crash is specific to the mock ICD (which can be built from Vulkan-Tools and used on Windows as well). That said, we're seeing some odd failures on all of our desktop CI platforms (Nvidia and AMD), but I'm not sure why yet.

paul-lunarg:
VkSyncValTest.SyncQSBufferCopyHazards doesn't run core validation. With core on it generates this error:
```
Validation Error: Attempt to free VkCommandBuffer which is in use.
```
Which could very well crash the driver. I am in the process of fixing these kinds of bugs but unfortunately there are a bunch, however I should get to the ones that cause CI to fail first and get them merged ASAP.

ReinUsesLisp:
I see `SyncQSBufferCopyHazards` is running and passing on other PRs now. Can CI be re-run?

tests/vklayertests_nvidia_best_practices.cpp
```diff
@@ -0,0 +1,1141 @@
...
+    }
+}
+
+TEST_F(VkNvidiaBestPracticesLayerTest, QueueBindSparse_NotAsync) {
+    InitBestPracticesFramework(kEnableNVIDIAValidation);
+    InitState();
+
+    if (!m_device->phy().features().sparseBinding) {
+        // No sparse binding support
+        return;
```
ncesario-lunarg, line 140: Please use `GTEST_SKIP() << "<message>"` to skip tests (same comment for similar instances below).

tests/vklayertests_nvidia_best_practices.cpp
```diff
@@ -0,0 +1,1141 @@
...
+
+TEST_F(VkNvidiaBestPracticesLayerTest, AccelerationStructure_NotAsync) {
+    InitBestPracticesFramework(kEnableNVIDIAValidation);
+    InitState();
+
+    const char *extensions[] = {
+        VK_KHR_ACCELERATION_STRUCTURE_EXTENSION_NAME,
+    };
+
+    if (!DeviceExtensionSupported(gpu(), nullptr, VK_KHR_ACCELERATION_STRUCTURE_EXTENSION_NAME)) {
```
ncesario-lunarg, line 248: Is this test intentionally not enabling dependent extensions? If not, then this should probably use `AddRequiredExtensions` _before_ `InitBestPracticesFramework` and then use `m_device_extension_names` in `device_ci` below. This test is crashing on AMD drivers, though not sure if this is why.

tests/vklayertests_nvidia_best_practices.cpp
```diff
@@ -0,0 +1,1141 @@
...
+
+        command_buffer.end();
+    }
+}
+
+TEST_F(VkNvidiaBestPracticesLayerTest, AllocateMemory_SetPriority) {
+    InitBestPracticesFramework(kEnableNVIDIAValidation);
+    InitState();
+
+    VkMemoryAllocateInfo memory_ai = {VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_INFO};
```
ncesario-lunarg, line 379: Please use `LvlInitStruct<VkMemoryAllocateInfo>()` to ensure `pNext` is initialized to `nullptr`. Same comment for similar instances below.

tests/vklayertests_nvidia_best_practices.cpp
```diff
@@ -0,0 +1,1141 @@
...
+
+// Tests for NVIDIA-specific best practices
+const char *kEnableNVIDIAValidation = "VALIDATION_CHECK_ENABLE_VENDOR_SPECIFIC_NVIDIA";
+
+TEST_F(VkNvidiaBestPracticesLayerTest, PageableDeviceLocalMemory) {
+    AddRequiredExtensions(VK_EXT_PAGEABLE_DEVICE_LOCAL_MEMORY_EXTENSION_NAME);
+    InitBestPracticesFramework(kEnableNVIDIAValidation);
+    InitState();
+
+    if (!AreRequiredExtensionsEnabled()) {
```
ncesario-lunarg, line 30: In general this test can be made before `InitState` (which will create the logical device).

layers/best_practices_utils.cpp
```diff
@@ -665,6 +734,54 @@ bool BestPractices::PreCallValidateAllocateMemory(VkDevice device, const VkMemor
...
+                                              "has just been released, and it could have been reused in place of this allocation.",
+                                              VendorSpecificTag(kBPVendorNVIDIA));
+                } else {
+                    const uint32_t seconds = static_cast<uint32_t>(time_delta.count() / 1000);
+                    const uint32_t milliseconds = static_cast<uint32_t>(time_delta.count() % 1000);
+
+                    skip |= LogPerformanceWarning(
+                        device, kVUID_BestPractices_AllocateMemory_ReuseAllocations,
+                        "%s Reuse memory allocations instead of releasing and reallocating. A memory allocation has been released "
+                        "%d.%03d seconds ago, and it could have been reused in place of this allocation.",
```
ncesario-lunarg, line 778: I believe these should also be `PRIu32`.

ncesario-lunarg (commented):
Just a few more minor comments and some other unresolved comments from before, but this generally LGTM.

@rlocatti-nv were you able to run this change with core checks enabled (setting the environment variable `VK_LAYER_TESTS_DISABLE_CORE_VALIDATION=false`)? Some of the tests are crashing on AMD drivers.

rlocatti-nv:
Core validation should be good now.

tests/vklayertests_nvidia_best_practices.cpp
```diff
@@ -0,0 +1,1242 @@
...
+
+    {
+        m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_PERFORMANCE_WARNING_BIT_EXT,
+                                             "UNASSIGNED-BestPractices-QueueBindSparse-NotAsync");
+        vk::QueueBindSparse(graphics_queue, 1, &bind_info, VK_NULL_HANDLE);
+        m_errorMonitor->VerifyFound();
+    }
+}
+
+TEST_F(VkNvidiaBestPracticesLayerTest, AccelerationStructure_NotAsync) {
```
ncesario-lunarg, line 242: This test is still failing on RT AMD hardware/drivers (which I do not have locally to test). I can try to pull a stack trace off of the CI machine tomorrow.

ncesario-lunarg: I think I found the problem:
```diff
diff --git a/tests/vklayertests_nvidia_best_practices.cpp b/tests/vklayertests_nvidia_best_practices.cpp
index 9750a3b8f..fbe1000e2 100644
--- a/tests/vklayertests_nvidia_best_practices.cpp
+++ b/tests/vklayertests_nvidia_best_practices.cpp
@@ -240,14 +240,17 @@ TEST_F(VkNvidiaBestPracticesLayerTest, QueueBindSparse_NotAsync) {
 }

 TEST_F(VkNvidiaBestPracticesLayerTest, AccelerationStructure_NotAsync) {
+    SetTargetApiVersion(VK_API_VERSION_1_1);
     AddRequiredExtensions(VK_KHR_ACCELERATION_STRUCTURE_EXTENSION_NAME);
     AddRequiredExtensions(VK_KHR_BUFFER_DEVICE_ADDRESS_EXTENSION_NAME);
     InitBestPracticesFramework(kEnableNVIDIAValidation);
-    InitState();
-
+    if (DeviceValidationVersion() < VK_API_VERSION_1_1) {
+        GTEST_SKIP() << "Vulkan >= 1.1 required";
+    }
     if (!AreRequiredExtensionsEnabled()) {
         GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported";
     }
+    InitState();

     VkDeviceQueueCreateInfo general_queue_ci = LvlInitStruct<VkDeviceQueueCreateInfo>();
     general_queue_ci.queueFamilyIndex = m_device->QueueFamilyMatching(VK_QUEUE_GRAPHICS_BIT, 0);
```

Looks like this crashes on the mock ICD as well, but only when running with core checks enabled (env var: `VK_LAYER_TESTS_DISABLE_CORE_VALIDATION=false`)

@paul-lunarg looks like this is more motivation to get this enabled by default sooner rather than later ;)

rlocatti-nv: Thanks for looking into it. Even with this, the internal CI still seems to be failing.
If possible, I'm inclined to guard this test with `#if 0` for now.

rlocatti-nv:
Testing with `AccelerationStructure_NotAsync` disabled.

tests/vklayertests_nvidia_best_practices.cpp
ncesario-lunarg: It may make sense to skip this test, but I don't think we need to turn it off entirely. I mainly want to make sure this is a testing and/or hardware/driver issue, and there is nothing wrong with the changes going into the layer code. If it does make sense to turn this off, can please use something like the following after the `InitFramework` call in that test rather than `#if 0`?
```cpp
if (!IsDriver(VK_DRIVER_ID_AMD_PROPRIETARY)) {
    GTEST_SKIP() << "Test is crashing on AMD hardware for unknown reasons.";
}
```

rlocatti-nv: Added it.

tests/vklayertests_nvidia_best_practices.cpp
```diff
@@ -0,0 +1,1249 @@
...
+        vk_testing::DeviceMemory memory(*m_device, memory_ai);
+    }
+
+    m_errorMonitor->VerifyFound();
+}
+
+TEST_F(VkNvidiaBestPracticesLayerTest, BindMemory_NoPriority) {
+    AddRequiredExtensions(VK_EXT_PAGEABLE_DEVICE_LOCAL_MEMORY_EXTENSION_NAME);
+    AddRequiredExtensions(VK_KHR_MAINTENANCE_4_EXTENSION_NAME);
+    InitBestPracticesFramework(kEnableNVIDIAValidation);
```
ncesario-lunarg, line 465: This also needs the 1.1 check:
```diff
diff --git a/tests/vklayertests_nvidia_best_practices.cpp b/tests/vklayertests_nvidia_best_practices.cpp
index a5e535ec0..567c9ff3c 100644
--- a/tests/vklayertests_nvidia_best_practices.cpp
+++ b/tests/vklayertests_nvidia_best_practices.cpp
@@ -460,14 +460,17 @@ TEST_F(VkNvidiaBestPracticesLayerTest, AllocateMemory_ReuseAllocations) {
 }
 
 TEST_F(VkNvidiaBestPracticesLayerTest, BindMemory_NoPriority) {
+    SetTargetApiVersion(VK_API_VERSION_1_1);
     AddRequiredExtensions(VK_EXT_PAGEABLE_DEVICE_LOCAL_MEMORY_EXTENSION_NAME);
     AddRequiredExtensions(VK_KHR_MAINTENANCE_4_EXTENSION_NAME);
     InitBestPracticesFramework(kEnableNVIDIAValidation);
-    InitState();
-
+    if (DeviceValidationVersion() < VK_API_VERSION_1_1) {
+        GTEST_SKIP() << "Vulkan >= 1.1 required";
+    }
     if (!AreRequiredExtensionsEnabled()) {
         GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported";
     }
+    InitState();
 
     VkDeviceQueueCreateInfo queue_ci = LvlInitStruct<VkDeviceQueueCreateInfo>();
     queue_ci.queueFamilyIndex = 0;
```

Tony-LunarG:
I'm seeing crashes in:
dEQP-VK.api.pipeline.pipeline_invalid_pointers_unused_structs.graphics
dEQP-VK.image.texel_view_compatible.graphic.basic.2d_image.texture_read.bc1_rgba_unorm_block.r16g16b16a16_unorm
dEQP-VK.image.texel_view_compatible.graphic.basic.2d_image.texture_read.bc1_rgba_srgb_block.r16g16b16a16_unorm
but only with VALIDATION_CHECK_ENABLE_VENDOR_SPECIFIC_NVIDIA

KarenGhavam-lunarG:
> I'm seeing crashes in:
dEQP-VK.api.pipeline.pipeline_invalid_pointers_unused_structs.graphics
dEQP-VK.image.texel_view_compatible.graphic.basic.2d_image.texture_read.bc1_rgba_unorm_block.r16g16b16a16_unorm
dEQP-VK.image.texel_view_compatible.graphic.basic.2d_image.texture_read.bc1_rgba_srgb_block.r16g16b16a16_unorm
but only with VALIDATION_CHECK_ENABLE_VENDOR_SPECIFIC_NVIDIA

@Tony-LunarG Is there specific information about the hardware configuration and configuration you had enabled with the validation layers that you ran that could be helpful for @rlocatti-nv ?

rlocatti-nv:
Tony, Karen, the revision I pushed after that message had the failures fixed.

Tony-LunarG:
Yes, I see the three tests I mentioned above all run now - thanks.  I'll do one last full run this evening.
And for the record, I'm running Win 11, RTX 2060, 472.81, khronos_validation.enables = VK_VALIDATION_FEATURE_ENABLE_BEST_PRACTICES_EXT,VALIDATION_CHECK_ENABLE_VENDOR_SPECIFIC_NVIDIA

Tony-LunarG:
CTS run looks good

Tony-LunarG (approved):
LGTM

ncesario-lunarg:
Thanks @Tony-LunarG! And thank you @rlocatti-nv for these changes!

rlocatti-nv:
Thanks for helping get this merged.

PR #4356, opened 2022-07-28
Queue Submit Time Support for RenderPass based operations (default off)

jzulauf-lunarg:
Update the first use/replay models to support renderpass and subpass operation validation at queue submit time.

### DIFF
--- a/layers/synchronization_validation.cpp
+++ b/layers/synchronization_validation.cpp
@@ -269,6 +269,17 @@ struct NoopBarrierAction {
     const bool layout_transition = false;
 };
 
+static void InitSubpassContexts(VkQueueFlags queue_flags, const RENDER_PASS_STATE &rp_state, const AccessContext *external_context,
+                                std::vector<AccessContext> &subpass_contexts) {
+    const auto &create_info = rp_state.createInfo;
+    // Add this for all subpasses here so that they exsist during next subpass validation
+    subpass_contexts.clear();
+    subpass_contexts.reserve(create_info.subpassCount);
+    for (uint32_t pass = 0; pass < create_info.subpassCount; pass++) {
+        subpass_contexts.emplace_back(pass, queue_flags, rp_state.subpass_dependencies, subpass_contexts, external_context);
+    }
+}
+
 // NOTE: Make sure the proxy doesn't outlive from, as the proxy is pointing directly to access contexts owned by from.
 CommandBufferAccessContext::CommandBufferAccessContext(const CommandBufferAccessContext &from, AsProxyContext dummy)
     : CommandBufferAccessContext(from.sync_state_) {
@@ -325,6 +336,7 @@ std::string CommandExecutionContext::FormatHazard(const HazardResult &hazard) co
     out << ", " << FormatUsage(hazard.tag) << ")";
     return out.str();
 }
+
 
 bool CommandExecutionContext::ValidForSyncOps() const {
     bool valid = GetCurrentEventsContext() && GetCurrentAccessContext();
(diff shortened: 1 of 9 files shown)


layers/synchronization_validation.h
```diff
@@ -787,19 +758,19 @@ class SyncOpBeginRenderPass : public SyncOpBase {
...
+    const RenderPassAccessContext *GetRenderPassAccessContext() const { return rp_context_; }
 
   protected:
     safe_VkRenderPassBeginInfo renderpass_begin_info_;
     safe_VkSubpassBeginInfo subpass_begin_info_;
     std::vector<std::shared_ptr<const IMAGE_VIEW_STATE>> shared_attachments_;
     std::vector<const IMAGE_VIEW_STATE *> attachments_;
     std::shared_ptr<const RENDER_PASS_STATE> rp_state_;
-    std::shared_ptr<ReplayRenderpassContext> replay_context_;  // Shared across all subpasses for same renderpass instance
+    const RenderPassAccessContext *rp_context_;
```
jeremyg-lunarg, line 773: This change freaked me out at first but it seems like it is ok because *rp_context_ is owned by the cb access context and the syncops won't outlive it... Right?

layers/synchronization_validation.cpp
```diff
@@ -326,6 +337,20 @@ std::string CommandExecutionContext::FormatHazard(const HazardResult &hazard) co
     return out.str();
 }
 
+void CommandExecutionContext::BeginRenderPassReplay(const SyncOpBeginRenderPass &begin_op, ResourceUsageTag tag) {
+    assert("Must override if use by derived type is valid" == nullptr);
```
jeremyg-lunarg, line 341: Would it be better to have these asserts inline in the header so it is more likely that whoever is making a new derived type sees what they need to do?

jeremyg-lunarg (approved):
LGTM.  The few inline comments are nice to haves, I don't think any of them are actual bugs.

tests/vksyncvaltests.cpp
```diff
@@ -4873,3 +4908,74 @@ TEST_F(VkSyncValTest, SyncQSOBarrierHazard) {
...
+    // Single renderpass barrier  (sanity check)
+    cb0.begin();
+    do_clear(cb0);
+    m_errorMonitor->VerifyNotFound();
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, "SYNC-HAZARD-WRITE_AFTER_WRITE");
+    do_begin_rp(cb0);
+    m_errorMonitor->VerifyFound();
+    // No "end render pass" as the begin fails
+
+    m_errorMonitor->ExpectSuccess();
```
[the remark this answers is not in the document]
ncesario-lunarg: Oh yeah...I should merge that 😬

tests/vksyncvaltests.cpp
```diff
@@ -4828,3 +4908,74 @@ TEST_F(VkSyncValTest, SyncQSOBarrierHazard) {
...
+                                     VK_ACCESS_COLOR_ATTACHMENT_READ_BIT, 0U});
+
+    rp_helper.InitRenderPass();
+    rp_helper.InitFramebuffer();
+    rp_helper.InitBeginInfo();
+
+    VkCommandBufferObj cb0(m_device, m_commandPool);
+    VkCommandBufferObj cb1(m_device, m_commandPool);
+
+    auto do_begin_rp = [&rp_helper](VkCommandBufferObj& cb_obj) { cb_obj.BeginRenderPass(rp_helper.render_pass_begin); };
```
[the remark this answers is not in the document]
jzulauf-lunarg: the test used to be more complex :)  the goal of the lambda's is to make it clear that we are doing the exact same thing in each version of the sequence...

layers/synchronization_validation.h
jzulauf-lunarg: The deleted line was dead code (never used, only on used in conjunction with a incorrect partial implementation).  The added line does reflect that both the sync_op and the renderpass context have a common owner, and thus a common lifespan.

ncesario-lunarg (approved):

PR #4357, opened 2022-07-28
Fix, then remove the fragment shader portion of 06896

ncesario-lunarg:
The first commit fixes 06896 to only trigger the missing fragment shader error when rasterization is enabled. The second commit removes this check (see internal issue 3178 for details). I've left the "fix commit" in history because it is my understanding this change will be coming back in the near future, just gated behind something like a maintenance extension check.

### DIFF
--- a/tests/vklayertests_graphics_library.cpp
+++ b/tests/vklayertests_graphics_library.cpp
@@ -1171,7 +1171,7 @@ TEST_F(VkGraphicsLibraryLayerTest, MissingShaderStages) {
         GTEST_SKIP() << "VkPhysicalDeviceGraphicsPipelineLibraryFeaturesEXT::graphicsPipelineLibrary not supported.";
     }
 
-    ASSERT_NO_FATAL_FAILURE(InitState(nullptr, &gpl_features));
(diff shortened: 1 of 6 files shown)


jeremyg-lunarg (approved):
Both the fix and revert commits LGTM

tests/vklayertests_graphics_library.cpp
```diff
@@ -1222,13 +1222,82 @@ TEST_F(VkGraphicsLibraryLayerTest, MissingShaderStages) {
...
+        vk::CmdDraw(m_commandBuffer->handle(), 3, 1, 0, 0);
         m_errorMonitor->VerifyFound();
+        m_commandBuffer->EndRenderPass();
+        m_commandBuffer->end();
     }
-}
\ No newline at end of file
+
+    {
+        // If rasterization is enabled, a fragment shader is not necessary
```
Tony-LunarG: Should "enabled" be "disabled" ? Or if discard is enabled?

Tony-LunarG (approved):
LGTM but I think a comment may be reversed

tests/vklayertests_graphics_library.cpp
ncesario-lunarg: 🤦‍♂️ yup

ncesario-lunarg:
The internal CI failure is one of the remaining spurious ones.

PR #4359, opened 2022-07-28
build: Update to header 1.3.223

mikes-lunarg:
- Update known-good
- Generate source


ncesario-lunarg (approved):
Looks like an unexciting one :)

ziga-lunarg (approved):

PR #4363, opened 2022-08-01
layers: Use KHR RayTracing shader alias

sjfricke:
Currently doing more validation work for ray tracing (`spirv-val` and now here) and updated some enum to use the `KHR` alias where it is used as a general enum (left if directly dealing with `VK_NV_ray_tracing`) as makes grepping much easier

### DIFF
--- a/layers/debug_printf.cpp
+++ b/layers/debug_printf.cpp
@@ -26,8 +26,8 @@
 #include "cmd_buffer_state.h"
 
 static const VkShaderStageFlags kShaderStageAllRayTracing =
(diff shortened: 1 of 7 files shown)


sjfricke:
>  CI Vulkan-ValidationLayers build # 8046 failed.

The only difference between this and build `8045` was I had to update the copyright so not sure why it failed now

ncesario-lunarg:
> > CI Vulkan-ValidationLayers build # 8046 failed.
> 
> The only difference between this and build `8045` was I had to update the copyright so not sure why it failed now

Yep, the failure is spurious. While we've "fixed" the test framework, the reporting/failure of tests failing in unexpected ways is only enforced for core tests, but BP and sync tests are still not enforcing this yet, and we know there are more tests we need to fix.

ncesario-lunarg (approved):
LGTM, but will wait for a quick look from @Tony-LunarG.

Tony-LunarG (approved):
LGTM

PR #4364, opened 2022-08-01
layers: Add missing indirect draw offset checks

sjfricke:
1. Adds 2 simple VU checks for `VUID-vkCmdDrawIndirect-offset-02710` and `VUID-vkCmdDrawIndexedIndirect-offset-02710` (not sure why the other `vkCmd` calls were checked and not these)

2. Add missing checks for `VK_AMD_draw_indirect_count`

3. Unified how `is a multiple of 4` logic is done

4. Used the `CMD_TYPE` for the common function


ncesario-lunarg (commented):
LGMT, thanks!

PR #4365, opened 2022-08-01
tests: Add missing device waits in SyncQS* tests

jeremyg-lunarg:
Many of these tests submit commands to multiple queues, but only detect errors on one queue. They should wait before ending to ensure that the successful commands complete before vulkan handles are destroyed.

These errors were hidden by a bug in CMD_BUFFER_STATE::Reset(), which shouldn't have been resetting the object in use count.

### DIFF
--- a/tests/vksyncvaltests.cpp
+++ b/tests/vksyncvaltests.cpp
@@ -4474,6 +4474,8 @@ TEST_F(VkSyncValTest, SyncQSBufferCopyHazards) {
     m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, "SYNC-HAZARD-WRITE_AFTER_READ");
     vk::QueueSubmit(m_device->m_queue, 1, &submit1, VK_NULL_HANDLE);
     m_errorMonitor->VerifyFound();
+
+    m_device->wait();
 }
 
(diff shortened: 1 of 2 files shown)


ncesario-lunarg (approved):

jeremyg-lunarg:
Re-pushed to fix commit message typos, not gonna wait for another CI

tests/vksyncvaltests.cpp
```diff
@@ -4474,6 +4474,8 @@ TEST_F(VkSyncValTest, SyncQSBufferCopyHazards) {
     m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, "SYNC-HAZARD-WRITE_AFTER_READ");
     vk::QueueSubmit(m_device->m_queue, 1, &submit1, VK_NULL_HANDLE);
     m_errorMonitor->VerifyFound();
+
+    m_device->wait();
```
jzulauf-lunarg, line 4478: Is their a performance hit and wait to the device destructor... I mean if there's nothing pending, should noop, yes?

jeremyg-lunarg: I don't think that would solve this problem in all cases. If there's a local vulkan handle in a RAII wrapper, wouldn't its destructor would get called before m_device's?  If so, then the local handle could still get destroyed while it was pending in a command buffer.

Issue #4267, opened 2022-06-30
Add validation for VK_EXT_multisampled_render_to_single_sample


Tony-LunarG:
**Valid Usage IDs requested**
VUID-VkGraphicsPipelineCreateInfo-multisampledRenderToSingleSampled-06853
VUID-VkGraphicsPipelineCreateInfo-renderPass-06854
VUID-vkCmdBindPipeline-pipeline-06856
VUID-VkRenderingInfo-multisampledRenderToSingleSampled-06857
VUID-VkRenderingInfo-imageView-06858
VUID-VkRenderingInfo-imageView-06859
VUID-VkRenderingAttachmentInfo-imageView-06861
VUID-VkRenderingAttachmentInfo-imageView-06862
VUID-VkRenderingAttachmentInfo-imageView-06863
VUID-VkRenderingAttachmentInfo-imageView-06864
VUID-VkRenderingAttachmentInfo-imageView-06865
VUID-VkSubpassDescription-pColorAttachments-06868
VUID-VkSubpassDescription2-multisampledRenderToSingleSampled-06869
VUID-VkSubpassDescription2-pNext-06870
VUID-VkSubpassDescription2-pNext-06871
VUID-VkSubpassDescription2-multisampledRenderToSingleSampled-06872
VUID-VkSubpassDescriptionDepthStencilResolve-pNext-06873
VUID-VkSubpassDescriptionDepthStencilResolve-pNext-06874
VUID-VkSubpassDescriptionDepthStencilResolve-pNext-06875
VUID-VkSubpassDescriptionDepthStencilResolve-pNext-06876
VUID-VkSubpassDescriptionDepthStencilResolve-pNext-06877
VUID-VkMultisampledRenderToSingleSampledInfoEXT-rasterizationSamples-06878
VUID-VkMultisampledRenderToSingleSampledInfoEXT-pNext-06879
VUID-VkMultisampledRenderToSingleSampledInfoEXT-pNext-06880
VUID-VkFramebufferCreateInfo-samples-06881
VUID-VkImageCreateInfo-multisampledRenderToSingleSampled-06882
VUID-VkImageCreateInfo-flags-06883

Tony-LunarG:
Still needs validation for VUID-VkMultisampledRenderToSingleSampledInfoEXT-pNext-06879
which is under question here https://gitlab.khronos.org/vulkan/vulkan/-/issues/3169

PR #4367, opened 2022-08-01
Validation and tests for VK_EXT_multisampled_render_to_single_sampled

Tony-LunarG:
Addresses #4267 

Does not include validation for VUID-VkMultisampledRenderToSingleSampledInfoEXT-pNext-06879
which is under question here https://gitlab.khronos.org/vulkan/vulkan/-/issues/3169

### DIFF
--- a/layers/buffer_validation.cpp
+++ b/layers/buffer_validation.cpp
@@ -2030,6 +2030,22 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
         }
     }
 
+    if ((pCreateInfo->flags & VK_IMAGE_CREATE_MULTISAMPLED_RENDER_TO_SINGLE_SAMPLED_BIT_EXT) != 0) {
+        if (!(enabled_features.multisampled_render_to_single_sampled_features.multisampledRenderToSingleSampled)) {
+            skip |= LogError(
+                device, "VUID-VkImageCreateInfo-multisampledRenderToSingleSampled-06882",
+                "vkCreateImage(): pCreateInfo.flags contains VK_IMAGE_CREATE_MULTISAMPLED_RENDER_TO_SINGLE_SAMPLED_BIT_EXT "
+                "but the multisampledRenderToSingleSampled feature is not enabled");
+        }
+        if (pCreateInfo->samples != VK_SAMPLE_COUNT_1_BIT) {
+            skip |= LogError(
+                device, "VUID-VkImageCreateInfo-flags-06883",
+                "vkCreateImage(): pCreateInfo.flags contains VK_IMAGE_CREATE_MULTISAMPLED_RENDER_TO_SINGLE_SAMPLED_BIT_EXT "
+                "but samples (%s) is not equal to VK_SAMPLE_COUNT_1_BIT",
+                string_VkSampleCountFlagBits(pCreateInfo->samples));
+        }
+    }
+
     skip |= ValidateImageFormatFeatures(pCreateInfo);
 
     // Check compatibility with VK_KHR_portability_subset
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -2114,13 +2114,13 @@ bool CoreChecks::ValidatePipeline(std::vector<std::shared_ptr<PIPELINE_STATE>> c
         const auto msrtss_info = LvlFindInChain<VkMultisampledRenderToSingleSampledInfoEXT>(subpass_desc->pNext);
         if (msrtss_info && msrtss_info->multisampledRenderToSingleSampledEnable) {
             if (msrtss_info->rasterizationSamples != pipeline->MultisampleState()->rasterizationSamples) {
-                skip |= LogError(device, "VUID-VkGraphicsPipelineCreateInfo-renderPass-06854",
-                                 "vkCreateGraphicsPipelines(): VkMultisampledRenderToSingleSampledInfoEXT in the pNext chain of "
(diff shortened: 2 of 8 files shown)


layers/buffer_validation.cpp
```diff
@@ -2040,6 +2040,22 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
         }
     }
 
+    if ((pCreateInfo->flags &VK_IMAGE_CREATE_MULTISAMPLED_RENDER_TO_SINGLE_SAMPLED_BIT_EXT) != 0) {
```
ncesario-lunarg, line 2043: Minor nit: no space after binary `&`.

layers/core_validation.cpp
```diff
@@ -8623,6 +8640,106 @@ bool CoreChecks::ValidateCmdBeginRendering(VkCommandBuffer commandBuffer, const
                          pRenderingInfo->viewMask, phys_dev_props_core11.maxMultiviewViewCount);
     }
 
+    if (IsExtEnabled(device_extensions.vk_ext_multisampled_render_to_single_sampled)) {
+        const auto msrtss_info = LvlFindInChain<VkMultisampledRenderToSingleSampledInfoEXT>(pRenderingInfo->pNext);
+        for (uint32_t j = 0; j < pRenderingInfo->colorAttachmentCount; ++j) {
+            if (pRenderingInfo->pColorAttachments[j].imageView != VK_NULL_HANDLE) {
```
ncesario-lunarg, line 8646: Not suggesting any change here, but this reminds me of a general question I've had: should the `!= VK_NULL_HANDLE` check be moved to the `Get<...>` function, and then just check if the state is non-null?

layers/core_validation.cpp
```diff
@@ -2111,6 +2111,21 @@ bool CoreChecks::ValidatePipeline(std::vector<std::shared_ptr<PIPELINE_STATE>> c
...
 
+    if (rp_state && rp_state->renderPass() != VK_NULL_HANDLE &&
+        IsExtEnabled(device_extensions.vk_ext_multisampled_render_to_single_sampled)) {
+        const auto msrtss_info = LvlFindInChain<VkMultisampledRenderToSingleSampledInfoEXT>(subpass_desc->pNext);
+        if (msrtss_info && msrtss_info->multisampledRenderToSingleSampledEnable) {
+            if (msrtss_info->rasterizationSamples != pipeline->MultisampleState()->rasterizationSamples) {
+                skip |= LogError(
+                    device, "VUID-VkGraphicsPipelineCreateInfo-renderPass-06854",
+                    "vkCreateGraphicsPipelines(): A VkMultisampledRenderToSingleSampledInfoEXT struct in the pNext chain of "
+                    "pipelines[%" PRIu32 "], subpass index %u's VkSubpassDescription2 has a rasterizationSamples of (%" PRIu32
```
ncesario-lunarg, line 2122: `%u` -> `PRIu32`? Same comment for other instances.

tests/vklayertests_descriptor_renderpass_framebuffer.cpp
```diff
@@ -12738,3 +12739,426 @@ TEST_F(VkLayerTest, ImagelessFramebufferWith3DImage) {
...
+    }
+
+    image_create_info.samples = VK_SAMPLE_COUNT_2_BIT;
+    image_create_info.flags = VK_IMAGE_CREATE_MULTISAMPLED_RENDER_TO_SINGLE_SAMPLED_BIT_EXT;
+    image_create_info.imageType = VK_IMAGE_TYPE_2D;
+    image_create_info.format = VK_FORMAT_B8G8R8A8_UNORM;
+    VkImage bad_flag_image;
+    // VK_IMAGE_CREATE_MULTISAMPLED_RENDER_TO_SINGLE_SAMPLED_BIT_EXT requires VK_SAMPLE_COUNT_1_BIT
+    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkImageCreateInfo-flags-06883");
+    vk::CreateImage(device(), &image_create_info, nullptr, &bad_flag_image);
```
ncesario-lunarg, line 13143: `vk_testing::Image`? Not as big of a deal to use the wrappers now that we are making sure leaked objects are reported, though.

ncesario-lunarg (commented):
Other than the formatting nit, LGTM!

layers/buffer_validation.cpp
Tony-LunarG: Fixed

layers/core_validation.cpp
Tony-LunarG: That's probably  more efficient, especially if you're just going to check state for null too.  But I left this as is.

layers/core_validation.cpp
Tony-LunarG: Fixed

tests/vklayertests_descriptor_renderpass_framebuffer.cpp
Tony-LunarG: I'm expecting the CreateImage to fail.  I usually only use the wrappers for calls I expect to succeed.  The DestroyImage was left over from test development and I removed it.

ncesario-lunarg: 👍 sounds good...though, there's no harm in using `vk_testing::Image` if the creation fails (as long as it fails due to validation).

ncesario-lunarg (approved):

PR #4368, opened 2022-08-02
Add fence wait support to the queue submit feature.  

jzulauf-lunarg:
Alpha Candidate for phase II -- with known gaps (missing Submit2)

### DIFF
--- a/layers/synchronization_validation.cpp
+++ b/layers/synchronization_validation.cpp
@@ -860,6 +860,15 @@ void AccessContext::ForAll(Action &&action) {
     }
 }
 
+template <typename Predicate>
+void AccessContext::EraseIf(Predicate &&pred) {
+    for (const auto address_type : kAddressTypes) {
+        auto &accesses = GetAccessStateMap(address_type);
+        // Note: Don't forward, we don't want r-values moved, since we're going to make multiple calls.
+        layer_data::EraseIf(accesses, pred);
+    }
+}
+
 // A recursive range walker for hazard detection, first for the current context and the (DetectHazardRecur) to walk
 // the DAG of the contexts (for example subpasses)
 template <typename Detector>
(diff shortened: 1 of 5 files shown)


layers/vk_layer_data.h
```diff
@@ -986,5 +986,23 @@ bool Contains(const Container &container, const Key &key) {
...
 }
 
+// EraseRemoveIf is not implemented in STL style for two reasons
+// 1) Robin Hood containers don't support two-argument erase functions
+// 2) STL remove_if requires the predicate to be const w.r.t the value-type, and we are doing mutating removing traversals
+//
+// Any STL style erase-remove-if implementation should be named to indicate the const value restrictions (and whether
+// it's a performance win to do so is unknown
+template <typename Container, typename Predicate>
+void EraseRemoveIf(Container &c, Predicate &&p) {
```
jeremyg-lunarg, line 996: Why isn't this called EraseIf() since the STL equivalent is std::erase_if()? Or is this the equivalent of some combination of both erase and remove? like discussed in: https://stackoverflow.com/questions/799314/difference-between-erase-and-remove

jeremyg-lunarg (commented):
Other than a method name that confuses me, this looks good. I am a bit sad we won't have a wait worm anymore. :worm: :worm: :worm:

layers/vk_layer_data.h
jzulauf-lunarg: Essentially, I was trying to implement erase_if using the erase-remove approach, and discovered Robin Hood couldn't handle, and ended up writing erase_if (without noting it's existence in '20).  If also looks like the STL erase_if doesn't have the same constraints on const-ness of the predicate argument (at least they aren't mentioned here: https://en.cppreference.com/w/cpp/container/map/erase_if )

jeremyg-lunarg (approved):

layers/vk_layer_data.h
jzulauf-lunarg: @jeremyg-lunarg  -- redid this as EraseIf with the same return semantics as erase_if

layers/synchronization_validation.cpp
```diff
@@ -860,6 +860,15 @@ void AccessContext::ForAll(Action &&action) {
     }
 }
 
+template <typename Predicate>
+void AccessContext::EraseIf(Predicate &&pred) {
```
ncesario-lunarg, line 864: Shouldn't this definition go into the header, or is there a particular reason to restrict the types `Predicate` can represent to this file?

jzulauf-lunarg: @ncesario-lunarg -- Given that all of syncval is implemented is a single file, I've kept the implementations in the cpp to the extent possible.  If we were to break this up with appropriate namespace containment, that would change.

PR #4369, opened 2022-08-02
tests: Fix how device_profile_api layer is enabled

sjfricke:
This came up as I was touching a GPU validation test (which doesn't use devsim) and realized `EnableDeviceProfileLayer` was never called it still was working

- Renamed `EnableDeviceProfileLayer` to a more accurate name `OverrideDevsimForDeviceProfileLayer`
- Everything using `GTEST_SKIP` now
- Create Test doc updated
- Moved all tests to use the `LoadDeviceProfileLayer` function


ncesario-lunarg (approved):
LGTM, thanks!

PR #4370, opened 2022-08-02
tests: Add submit2 GPU-Val test

sjfricke:
So I originally thought there was a bug with `vkQueueSubmit2` for GPU-Validation, but after writing a test I realize there isn't.

This test still adds a missing test for `VUID-vkCmdDrawIndexedIndirectCount-countBuffer-02717` and cleans up the `GpuDrawIndirectCountDeviceLimit` a bit. Also, I can see using the `submit2` for future tests

### DIFF
--- a/tests/vklayertests_gpu.cpp
+++ b/tests/vklayertests_gpu.cpp
@@ -1698,7 +1698,7 @@ TEST_F(VkGpuAssistedLayerTest, GpuBuildAccelerationStructureValidationRestoresSt
 TEST_F(VkGpuAssistedLayerTest, GpuDrawIndirectCountDeviceLimit) {
     TEST_DESCRIPTION("GPU validation: Validate maxDrawIndirectCount limit");
     SetTargetApiVersion(VK_API_VERSION_1_3);
-    AddRequiredExtensions(VK_KHR_DRAW_INDIRECT_COUNT_EXTENSION_NAME);
+    AddRequiredExtensions(VK_KHR_DRAW_INDIRECT_COUNT_EXTENSION_NAME);  // instead of enabling feature
     InitGpuAssistedFramework(false);
     if (!AreRequiredExtensionsEnabled()) {
         GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported.";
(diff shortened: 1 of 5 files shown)


tests/vklayertests_gpu.cpp
```diff
@@ -1698,11 +1698,7 @@ TEST_F(VkGpuAssistedLayerTest, GpuBuildAccelerationStructureValidationRestoresSt
 TEST_F(VkGpuAssistedLayerTest, GpuDrawIndirectCountDeviceLimit) {
     TEST_DESCRIPTION("GPU validation: Validate maxDrawIndirectCount limit");
     SetTargetApiVersion(VK_API_VERSION_1_3);
-    AddRequiredExtensions(VK_KHR_DRAW_INDIRECT_COUNT_EXTENSION_NAME);
```
Tony-LunarG, line 1701: If you use the api level to enable indirect count, you need to enable the feature or get VUID-vkCmdDrawIndexedIndirectCount-None-04445

tests/vklayertests_gpu.cpp
```diff
@@ -1808,6 +1798,83 @@ TEST_F(VkGpuAssistedLayerTest, GpuDrawIndirectCountDeviceLimit) {
     }
 }
 
+TEST_F(VkGpuAssistedLayerTest, GpuDrawIndexedIndirectCountDeviceLimitSubmit2) {
+    TEST_DESCRIPTION("GPU validation: Validate maxDrawIndirectCount limit using vkQueueSubmit2");
+    SetTargetApiVersion(VK_API_VERSION_1_3);
+    InitGpuAssistedFramework(false);
+    if (DeviceValidationVersion() < VK_API_VERSION_1_3) {
+        GTEST_SKIP() << "At least Vulkan version 1.3 is required";
```
Tony-LunarG, line 1806: This needs to enable drawIndirectCount (04445)  and Synchronization2 (03866)

tests/vklayertests_gpu.cpp
```diff
@@ -1808,6 +1798,83 @@ TEST_F(VkGpuAssistedLayerTest, GpuDrawIndirectCountDeviceLimit) {
...
+    VkCommandBufferBeginInfo begin_info = LvlInitStruct<VkCommandBufferBeginInfo>();
+    m_commandBuffer->begin(&begin_info);
+    m_commandBuffer->BeginRenderPass(m_renderPassBeginInfo);
+    vk::CmdBindPipeline(m_commandBuffer->handle(), VK_PIPELINE_BIND_POINT_GRAPHICS, pipe.handle());
+    VkViewport viewport = {0, 0, 16, 16, 0, 1};
+    vk::CmdSetViewport(m_commandBuffer->handle(), 0, 1, &viewport);
+    VkRect2D scissor = {{0, 0}, {16, 16}};
+    vk::CmdSetScissor(m_commandBuffer->handle(), 0, 1, &scissor);
+
+    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-vkCmdDrawIndexedIndirectCount-countBuffer-02717");
```
Tony-LunarG, line 1865: It needs to have an index buffer bound

Tony-LunarG (changes requested):
Just a couple things needed.  A suggestion when developing GPU-AV tests is to enable core validation by setting features.disabledValidationFeatureCount = 0 in line 26 of vklayertests_gpu.cpp until your new test is clean from validation errors.

ncesario-lunarg:
> A suggestion when developing GPU-AV tests is to enable core validation by setting features.disabledValidationFeatureCount = 0 in line 26 of vklayertests_gpu.cpp until your new test is clean from validation errors.

Should/can this be the default?

jeremyg-lunarg:
> > A suggestion when developing GPU-AV tests is to enable core validation by setting features.disabledValidationFeatureCount = 0 in line 26 of vklayertests_gpu.cpp until your new test is clean from validation errors.
> 
> Should/can this be the default?

I think so. Once we get a batch of tests clean with that option it does seem like we should keep it on. The only downside is the risk of implicit dependency on having core validation, which hides failures when GPU-AV (or whatever) is run alone. But given recent history that seems like a lower risk  to me.

Tony-LunarG:
> 

If I remember correctly, and it's been long enough that I probably don't, we want to test GPU-AV without CoreChecks because
- That's the way we tell end users to run it
- There were cases where CoreChecks "fixed" something in GPU-AV and we want to catch those cases
- There's no reason to validate the same test that hasn't changed in years 10 times a day
But I can see the other side too

tests/vklayertests_gpu.cpp
sjfricke: I agree... but not why is that producing an error for this test... I thought we fixed this, will look into this

sjfricke:
> Should/can this be the default?

I agree with Tony reasoning as I doubt too many new people will be attempting to write more GPU Validation tests randomly.

Maybe worth adding a flag to `vk_layer_validation_tests` to quickly be able to go between the 2 options... regardless, should for sure add this to [gpu_validation.md#gpu-assisted-validation-testing](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/docs/gpu_validation.md#gpu-assisted-validation-testing) about the preferred way to toggle them off and on... but not all the tests even use `InitGpuAssistedFramework` to start with so there isn't even currently a single location all the tests could toggle them

ncesario-lunarg:
> If I remember correctly, and it's been long enough that I probably don't, we want to test GPU-AV without CoreChecks because
> 
> * That's the way we tell end users to run it
> * There were cases where CoreChecks "fixed" something in GPU-AV and we want to catch those cases
> * There's no reason to validate the same test that hasn't changed in years 10 times a day
>   But I can see the other side too

Hmm...I wonder if this means we should add "yet another permutation" to CI :/

Tony-LunarG (approved):
LGTM

PR #4371, opened 2022-08-02
layers: Fix VK_IMAGE_CREATE_2D_VIEW_COMPATIBLE_BIT_EXT handling

jeremyg-lunarg:
This new flag also needs to make IsDepthSliced() return true,
so it works just like VK_IMAGE_CREATE_2D_ARRAY_COMPATIBLE_BIT.


Tony-LunarG (approved):
LGTM and fixes CTS crash

PR #4372, opened 2022-08-02
layers: Avoid crashes when dynamic_rendering is enabled

jeremyg-lunarg:
Add RENDER_PASS_STATE::UsesDynamicRendering() because several code
paths were going in the 'normal' RenderPass code when
use_dynamic_rendering_inherited == true but use_dynamic_rendering ==
false.


ncesario-lunarg (approved):
Yikes, LGTM!

Tony-LunarG (approved):
LGTM and fixes the CTS crash I saw.

PR #4379, opened 2022-08-04
build: Add GitHub Actions dependency caching

karl-lunarg:
Cache the builds of components that this project depends on to speed up
GitHub Actions builds.

The dependent components are specified in scripts/known_good.json whose
hash is used as a key for the cache.  When the known good file is
updated, a cache won't be found for that hash/key and the components are
rebuilt and cached again.

For the more common case for when the known good file does not change,
the cache is found and the dependent components don't need to be
rebuilt.

Documentation for this GitHub Actions feature can be found at:
https://docs.github.com/en/actions/using-workflows
in the "Caching dependencies to speed up workflows" section.  There are
also APIs available for managing the caches.

### DIFF
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -1,5 +1,5 @@
-# Copyright (c) 2021 Valve Corporation
-# Copyright (c) 2021 LunarG, Inc.
+# Copyright (c) 2021-2022 Valve Corporation
+# Copyright (c) 2021-2022 LunarG, Inc.
 
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


ncesario-lunarg (approved):
LGTM! Great to see this only effects the github yml files and no cmake changes were necessary :)

karl-lunarg:
> LGTM! Great to see this only effects the github yml files and no cmake changes were necessary :)

Part of it is being able to run `update_deps.py` on the retrieved cache content without causing any damage or recompilations.

mikes-lunarg (approved):
Looks great. I had always assumed this would involve saving off the install directories and making `update_deps.py` cache aware. That might still be an option if we find the caches getting too big, but this approach is much simpler and the cache sizes seem reasonable.

karl-lunarg:
> I had always assumed this would involve saving off the install directories

Right!  That would save a ton of cache space.  I verified that the `path: external/**/build/install` syntax would work.  And we'd have to cache helper.cmake too.  But yes, we'd have to make update_deps notice that the install dirs are already there.

The current cache usage is about 4.1GB for a distinct revision of the known good file, and the limit is 10GB.  How well this works will depend on how many distinct revisions of the known good file get built.  Old PR branches (that have not been rebased recently) could increase that number.

PR #4380, opened 2022-08-04
Fix compatibility with VK_EXT_shader_module_identifier

Tony-LunarG:
Validation to follow.

### DIFF
--- a/layers/pipeline_state.cpp
+++ b/layers/pipeline_state.cpp
@@ -90,12 +90,15 @@ PIPELINE_STATE::StageStateVec PIPELINE_STATE::GetStageStates(const ValidationSta
                     // If module is null and there is a VkShaderModuleCreateInfo in the pNext chain of the stage info, then this
                     // module is part of a library and the state must be created
                     const auto shader_ci = LvlFindInChain<VkShaderModuleCreateInfo>(shader_stage.pNext);
+                    const uint32_t unique_shader_id = 0;  
                     if (shader_ci) {
-                        const uint32_t unique_shader_id = 0;  // TODO GPU-AV rework required to get this value properly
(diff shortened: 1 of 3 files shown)


jeremyg-lunarg (approved):
LGTM

layers/pipeline_state.cpp
```diff
@@ -90,12 +90,15 @@ PIPELINE_STATE::StageStateVec PIPELINE_STATE::GetStageStates(const ValidationSta
                     // If module is null and there is a VkShaderModuleCreateInfo in the pNext chain of the stage info, then this
                     // module is part of a library and the state must be created
                     const auto shader_ci = LvlFindInChain<VkShaderModuleCreateInfo>(shader_stage.pNext);
+                    const uint32_t unique_shader_id = 0;  
                     if (shader_ci) {
-                        const uint32_t unique_shader_id = 0;  // TODO GPU-AV rework required to get this value properly
+                        // TODO GPU-AV rework required to get this value properly
```
ncesario-lunarg, line 95: This is one of my questions regarding getting GPU-AV working with GPL :)

ncesario-lunarg (approved):
LGTM!

layers/pipeline_state.cpp
Tony-LunarG: Ugh, and I'm just realizing that shader_module_identifier has all the same GPU-AV problems.  I think I should at least warn if an identifier is retrieved for an instrumented shader.  
Off the top of my head, I think since the instrumented modules in the library already have a unique shader id, the library itself is a don't care, since the instrumentation will use the id from the modules.  But I'd want to step through a test to verify that.

PR #4381, opened 2022-08-04
build: Update to header 1.3.224

mikes-lunarg:
- Update known-good
- Generate source
- Remove check and test for deleted VUIDs 06413/06414


ziga-lunarg (approved):

ncesario-lunarg (approved):
LGTM!

PR #4382, opened 2022-08-05
SyncQS Alpha fix: Make RP access context pointers stable

jzulauf-lunarg:
Store renderpass access contexts as unique pointers, s.t. refernces held stay valid.

### DIFF
--- a/layers/synchronization_validation.cpp
+++ b/layers/synchronization_validation.cpp
@@ -2332,8 +2332,9 @@ ResourceUsageTag CommandBufferAccessContext::RecordBeginRenderPass(CMD_TYPE cmd_
     // Create an access context the current renderpass.
     const auto barrier_tag = NextCommandTag(cmd_type, ResourceUsageRecord::SubcommandType::kSubpassTransition);
     const auto load_tag = NextSubcommandTag(cmd_type, ResourceUsageRecord::SubcommandType::kLoadOp);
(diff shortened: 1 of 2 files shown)


layers/synchronization_validation.cpp
```diff
@@ -2332,8 +2332,9 @@ ResourceUsageTag CommandBufferAccessContext::RecordBeginRenderPass(CMD_TYPE cmd_
     // Create an access context the current renderpass.
     const auto barrier_tag = NextCommandTag(cmd_type, ResourceUsageRecord::SubcommandType::kSubpassTransition);
     const auto load_tag = NextSubcommandTag(cmd_type, ResourceUsageRecord::SubcommandType::kLoadOp);
-    render_pass_contexts_.emplace_back(rp_state, render_area, GetQueueFlags(), attachment_views, &cb_access_context_);
-    current_renderpass_context_ = &render_pass_contexts_.back();
+    render_pass_contexts_.emplace_back(
+        std::make_unique<RenderPassAccessContext>(rp_state, render_area, GetQueueFlags(), attachment_views, &cb_access_context_));
```
ncesario-lunarg, line 2336: No >= C++14 yet, so I think this needs to be `layer_data::make_unique`.

ncesario-lunarg (approved):
Other than making the compiler happy, LGTM!

layers/synchronization_validation.cpp
jzulauf-lunarg: pushed.

jeremyg-lunarg (approved):
LGTM

PR #4386, opened 2022-08-05
Fix drawing with dynamic rendering in secondary command buffers

ziga-lunarg:


tests/vklayertests_command.cpp
```diff
@@ -11338,3 +11338,73 @@ TEST_F(VkLayerTest, ResolveInvalidUsage) {
...
+    pipeline_rendering_info.colorAttachmentCount = 1;
+    pipeline_rendering_info.pColorAttachmentFormats = &format;
+
+    CreatePipelineHelper pipe(*this);
+    pipe.InitInfo();
+    pipe.gp_ci_.pNext = &pipeline_rendering_info;
+    pipe.InitState();
+    pipe.CreateGraphicsPipeline();
+
+    const VkCommandBufferInheritanceRenderingInfoKHR inheritanceRenderingInfo = {
```
ncesario-lunarg, line 11381: I think we're trying to use the `LvlInitStruct` helper for initializing these structs in general (that also makes it a little easier to read which fields are initialized to what IMHO).

ncesario-lunarg (approved):
LGTM!

jeremyg-lunarg (approved):
LGTM

tests/vklayertests_command.cpp
ziga-lunarg: Missed this because it was copy pasted from CTS

PR #4389, opened 2022-08-08
layers: Clean up CB state used in GPU-AV

sjfricke:
So I want to use `has_dispatch_cmd` but currently, a `vkCmdDraw` is setting it so this change is to refactor/cleanup those variables

- Changed variable names to match code style
- When `vkCmdExecuteCommands` is called, updates the state for all the values (before only did for `draw` in the best practice layers)
- Moved the `PostCallRecordCmdTraceRays*` calls to `GpuAssistedBase` as they are just there to call `ValidationStateTracker` and noticed the GPU-AV was missing the `PostCallRecordCmdTraceRaysIndirect2KHR` call (but debug_printf had it)
  -  edit:  `ValidationStateTracker` is a base class, so just removed functions all together

(cc @jeremyg-lunarg if you were touching this)

### DIFF
--- a/layers/best_practices_utils.cpp
+++ b/layers/best_practices_utils.cpp
@@ -2028,7 +2028,8 @@ void BestPractices::RecordCmdBeginRenderPass(VkCommandBuffer commandBuffer, Rend
                                              const VkRenderPassBeginInfo* pRenderPassBegin) {
     // Reset the renderpass state
     auto cb = GetWrite<bp_state::CommandBuffer>(commandBuffer);
-    cb->hasDrawCmd = false;
+    // TODO - move this logic to the Render Pass state as cb->has_draw_cmd should stay true for lifetime of command buffer
+    cb->has_draw_cmd = false;
     assert(cb);
     auto& render_pass_state = cb->render_pass_state;
     render_pass_state.touchesAttachments.clear();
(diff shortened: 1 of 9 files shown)


ncesario-lunarg (approved):
Minor comment nit, but LGTM!

layers/gpu_utils.cpp
```diff
@@ -743,6 +743,46 @@ void GpuAssistedBase::PreCallRecordDestroyPipeline(VkDevice device, VkPipeline p
     ValidationStateTracker::PreCallRecordDestroyPipeline(device, pipeline, pAllocator);
 }
 
+void GpuAssistedBase::PostCallRecordCmdTraceRaysNV(VkCommandBuffer commandBuffer, VkBuffer raygenShaderBindingTableBuffer,
```
jeremyg-lunarg, line 746: are these PostCallRecord* overrides doing anything? It looks to me like they're just calling the base implementation

jeremyg-lunarg (commented):
Looks to me like the additions to gpu_utils.{h,cpp} aren't needed since they're just calling the base implementation with nothing else happening.

layers/gpu_utils.cpp
sjfricke: o wow, didn't see the `class GpuAssistedBase : public ValidationStateTracker `... makes this change even nicer, will remove

jeremyg-lunarg (approved):
LGTM

PR #4390, opened 2022-08-08
layers: Add static assert for instrument.hpp values

sjfricke:
I commented this in the code well I hope, but this prevents the case when `instrument.hpp` is updated and we will fail the `static_assert` and then only need to update the single `#define`


ncesario-lunarg (approved):
LGTM!

Tony-LunarG (approved):
LGTM!

PR #4397, opened 2022-08-10
Add remaining VK_EXT_blend_operation_advanced VUID

sjfricke:
- VUID-VkPipelineColorBlendAttachmentState-colorBlendOp-01406
- VUID-VkPipelineColorBlendAttachmentState-advancedBlendIndependentBlend-01407
- VUID-VkPipelineColorBlendAttachmentState-advancedBlendIndependentBlend-01408
- VUID-VkPipelineColorBlendAttachmentState-colorBlendOp-01410

For the test I found it would be easier to set `VkPhysicalDeviceBlendOperationAdvancedPropertiesEXT` so added a commit to add support for `SetPhysicalDeviceProperties2EXT` to the Device Profile Layer

### DIFF
--- a/layers/vk_layer_utils.h
+++ b/layers/vk_layer_utils.h
@@ -286,6 +286,11 @@ static inline uint32_t GetPlaneIndex(VkImageAspectFlags aspect) {
     }
 }
 
+// all "advanced blend operation" found in spec
+static inline bool IsAdvanceBlendOperation(const VkBlendOp blend_op) {
+    return (static_cast<int>(blend_op) >= VK_BLEND_OP_ZERO_EXT) && (static_cast<int>(blend_op) <= VK_BLEND_OP_BLUE_EXT);
+}
+
 // Perform a zero-tolerant modulo operation
 static inline VkDeviceSize SafeModulo(VkDeviceSize dividend, VkDeviceSize divisor) {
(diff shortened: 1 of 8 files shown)


layers/vk_layer_utils.h
```diff
@@ -286,6 +286,11 @@ static inline uint32_t GetPlaneIndex(VkImageAspectFlags aspect) {
     }
 }
 
+// all "advanced blend operation" found in spec
+static inline bool IsAdvanceBlendOperation(const VkBlendOp blend_op) {
```
ncesario-lunarg, line 290: Looks like this could be generated from the xml, though not sure if it's worth the extra effort here.

ncesario-lunarg (approved):
LGTM!

layers/vk_layer_utils.h
sjfricke: thought the same thing, but felt like it was more trouble then worth unless another extension with more operations is added in the future

ncesario-lunarg:
@sjfricke I'm pretty sure the CI failure is illegitimate. Could you please try rebasing onto master (specifically to include https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/6ace8025bb07f2bdac5ff60a2f3bbf53e6c56779)?

PR #4398, opened 2022-08-10
Add indirect dispatch to GPU-AV

sjfricke:
Adds

- VUID-VkDispatchIndirectCommand-x-00417
- VUID-VkDispatchIndirectCommand-y-00418
- VUID-VkDispatchIndirectCommand-z-00419

In process 2 other changes to GPU-AV
- `GpuAssistedCmdDrawIndirectState` to `GpuAssistedCmdIndirectState` (named `cdi_state` to `indirect_state`)
- created a `GpuAssistedAccelerationStructureBuildValidationState::Destroy` to be consistent with how the GPU-VA validation state is destroyed


lunarpapillo:
I suspect the CI run failed because Windows decided it was a good time to update the machine.  Re-running.

layers/gpu_validation.cpp
```diff
@@ -1570,9 +1631,102 @@ void GpuAssisted::AllocatePreDrawValidationResources(GpuAssistedDeviceMemoryBloc
...
+        pipeline_ci.stage = pipeline_stage_ci;
+        pipeline_ci.layout = pre_dispatch_validation_state.pipeline_layout;
+
+        result = DispatchCreateComputePipelines(device, VK_NULL_HANDLE, 1, &pipeline_ci, nullptr,
+                                                &pre_dispatch_validation_state.pipeline);
+        if (result != VK_SUCCESS) {
+            ReportSetupProblem(device, "Failed to create compute pipeline for pre dispatch validation.");
+        }
+
+        pre_dispatch_validation_state.initialized = true;
```
ncesario-lunarg, line 1696: It's not immediately clear to me why this isn't predicated on `result == VK_SUCCESS` at the end of this function?

ncesario-lunarg (commented):
This LGTM, but will wait for @Tony-LunarG's input.

Great to know there's another GPUAV expert!

layers/gpu_validation.cpp
sjfricke: Then needs a better comment, but we only need to create the pipeline/shader once for the entire life of the VkDevice because it can just reuse the same VkPipeline each vkCmdDispatchIndirect... Another way to do it would be to create it once upfront at vkCreateDevice time always instead of waiting for the first dispatch indirect call

Tony-LunarG (changes requested):
Looks good, but the new setting needs to be added to .\layers\json\VkLayer_khronos_validation.json.in and then @christophe-lunarg needs to add the setting to vkconfig

sjfricke:
added to `VkLayer_khronos_validation.json.in` ... is `layers/vk_layer_settings.txt` still being used then?

christophe-lunarg:
It looks great and should work with vkconfig and this is merged.

christophe-lunarg (approved):

Tony-LunarG (approved):
LGTM

PR #4409, opened 2022-08-11
Add back 'outfilename' arg to generate_spirv.py

amaiorano:
Useful for CIs with strict sandboxing rules (i.e. @google).

The recent change to this script has broken our internal CI builds. This adds back the optional 'outfilename' arg that we need so that we can specify a safe path.

### DIFF
--- a/scripts/generate_spirv.py
+++ b/scripts/generate_spirv.py
@@ -48,7 +48,7 @@ def compile(filename, glslang_validator):
         args = [glslang_validator, "-V", "-H", "-o", tmpfile, filename]
         output = subprocess.check_output(args, universal_newlines=True)
     except subprocess.CalledProcessError as e:
-        raise(e.output)
+        raise Exception(e.output)
 
     # read the temp file into a list of SPIR-V words
     words = []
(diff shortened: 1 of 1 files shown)


ncesario-lunarg (approved):
This LGTM, but FYI @sjfricke as I _think_ you got rid of the `outfilename` arg, but can't recall the details.

sjfricke:
So `generate_source.py` does not have a "out file arg" neither. It uses `common_codegen.repo_relative('layers/generated')` which I did as well for `generate_spirv.py`

Also, why is `generate_spirv.py` being called in a CI flow, seems that script is really only designed for people making manual changes

amaiorano:
> So `generate_source.py` does not have a "out file arg" neither. It uses `common_codegen.repo_relative('layers/generated')` which I did as well for `generate_spirv.py`
> 
> Also, why is `generate_spirv.py` being called in a CI flow, seems that script is really only designed for people making manual changes

Fair question. I'm not actually super familiar with the internal setup for this project, my role last week was to make sure VVL imports (among others) were working. I'll need to take a closer look to see why we run `generate_spirv.py`, and whether we could get rid of it.

PR #4410, opened 2022-08-12
layers: Update drawIndirectFirstInstance VUID

sjfricke:
I noticed that `VUID-VkDrawIndirectCommand-firstInstance-00501` and `VUID-VkDrawIndexedIndirectCommand-firstInstance-00554` were not "implemented" but realized there are 2 VUIDs doing the same thing, talking to Tobias offline, also agreed the function VUID should be removed in favor of the structure VUID. Raised issue to remove from the spec as well https://github.com/KhronosGroup/Vulkan-Docs/pull/1913

### DIFF
--- a/layers/gpu_vuids.h
+++ b/layers/gpu_vuids.h
@@ -53,7 +53,7 @@ struct GpuVuidsCmdDrawIndirect : GpuVuid {
     GpuVuidsCmdDrawIndirect() : GpuVuid() {
         uniform_access_oob = "VUID-vkCmdDrawIndirect-None-02705";
         storage_access_oob = "VUID-vkCmdDrawIndirect-None-02706";
-        first_instance_not_zero = "VUID-vkCmdDrawIndirect-firstInstance-00478";
+        first_instance_not_zero = "VUID-VkDrawIndirectCommand-firstInstance-00501";
     }
 };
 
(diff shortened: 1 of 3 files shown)


layers/gpu_vuids.h
```diff
@@ -53,15 +53,15 @@ struct GpuVuidsCmdDrawIndirect : GpuVuid {
     GpuVuidsCmdDrawIndirect() : GpuVuid() {
         uniform_access_oob = "VUID-vkCmdDrawIndirect-None-02705";
         storage_access_oob = "VUID-vkCmdDrawIndirect-None-02706";
-        first_instance_not_zero = "VUID-vkCmdDrawIndirect-firstInstance-00478";
```
ncesario-lunarg, line 56: Do I understand correctly that we do not check (nor never have checked) 00478?

ncesario-lunarg (approved):
LGTM!

layers/gpu_vuids.h
sjfricke: We were checking it, but Im about to remove it from the spec in favor of the other VUID (see Vulkan-Doc PR linked) (trying to proactively save the energy of figuring this out at the next header PR)

Tony-LunarG (approved):
LGTM

PR #4419, opened 2022-08-15
Update GPU-AV VMA to version 3.0.1

Tony-LunarG:
This version of VMA enables much better GPU-AV performance as measured by running game traces with GPU-AV enabled.  This is because the VMA author implemented a Two-Level Segregated Fit (TLSF) allocation algorithm.


layers/gpu_utils.h
```diff
@@ -21,6 +21,7 @@
 #include "shader_validation.h"
 #include "cmd_buffer_state.h"
 #include "state_tracker.h"
+#define VMA_VULKAN_VERSION 1001000
```
ncesario-lunarg, line 24: Is this something that is required for vma or us?

ncesario-lunarg (approved):
LGTM!

layers/gpu_utils.h
Tony-LunarG: It tells VMA what version of Vulkan it can use.  Since GPU-AV requires 1.1, I set this to 1.1 too.

PR #4420, opened 2022-08-16
layers: Rerun spec const fold pass temporarily

sjfricke:
cc @amaiorano - should fix your issue until I get a fix in SPIRV-Tools and we pull it in... the test I change reproduces the issue you were having


amaiorano:
> cc @amaiorano - should fix your issue until I get a fix in SPIRV-Tools and we pull it in... the test I change reproduces the issue you were having

Thanks! Once this lands, I'll post here if it fixes our build.

ncesario-lunarg (approved):
LGTM!

PR #4421, opened 2022-08-16
layers: pass VkImageFormatListCreateInfo when iterating modifiers

zmike:
this otherwise breaks mutable format checks in drivers

### DIFF
--- a/layers/buffer_validation.cpp
+++ b/layers/buffer_validation.cpp
@@ -1813,9 +1813,11 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
     } else {
         auto modifier_list = LvlFindInChain<VkImageDrmFormatModifierListCreateInfoEXT>(pCreateInfo->pNext);
         auto explicit_modifier = LvlFindInChain<VkImageDrmFormatModifierExplicitCreateInfoEXT>(pCreateInfo->pNext);
+        /* format list must be passed to avoid failing driver checks for mutability */
+        auto format_list_info = LvlFindInChain<VkImageFormatListCreateInfo>(pCreateInfo->pNext);
         if (modifier_list) {
             for (uint32_t i = 0; i < modifier_list->drmFormatModifierCount; i++) {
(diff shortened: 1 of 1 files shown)


layers/buffer_validation.cpp
```diff
@@ -1823,6 +1824,7 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
                 image_format_info.tiling = pCreateInfo->tiling;
                 image_format_info.usage = pCreateInfo->usage;
                 image_format_info.flags = pCreateInfo->flags;
+                image_format_info.pNext = &format_list_info;
```
ncesario-lunarg, line 1827: I think this should just be `format_list_info` (without `&`), as `format_list_info` is already a pointer.

layers/buffer_validation.cpp
```diff
@@ -1813,6 +1813,7 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
     } else {
         auto modifier_list = LvlFindInChain<VkImageDrmFormatModifierListCreateInfoEXT>(pCreateInfo->pNext);
         auto explicit_modifier = LvlFindInChain<VkImageDrmFormatModifierExplicitCreateInfoEXT>(pCreateInfo->pNext);
+        auto format_list_info = LvlFindInChain<VkImageFormatListCreateInfo>(pCreateInfo->pNext);
```
ncesario-lunarg, line 1816: Could you please add a comment indicating why this is necessary?

ncesario-lunarg (commented):
Couple questions, but otherwise LGTM.

Is there a test case that could/should be added for this?

layers/buffer_validation.cpp
zmike: I don't think so. That would cause drivers to crash.

zmike: Ah, I think I have other issues here.

ncesario-lunarg:
Closing this as the "correct" solution is going to be a bit more involved: https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4435

PR #4424, opened 2022-08-16
syncval: Add queue submit apply barriers support

jzulauf-lunarg:
Modify the apply barriers functors to reflect queue submit time
validation needs.

### DIFF
--- a/layers/synchronization_validation.cpp
+++ b/layers/synchronization_validation.cpp
@@ -860,6 +860,16 @@ void AccessContext::ForAll(Action &&action) {
     }
 }
 
+template <typename Action>
+void AccessContext::ConstForAll(Action &&action) const {
+    for (const auto address_type : kAddressTypes) {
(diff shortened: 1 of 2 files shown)


layers/synchronization_validation.cpp
```diff
@@ -860,6 +860,16 @@ void AccessContext::ForAll(Action &&action) {
     }
 }
 
+template <typename Action>
+void AccessContext::ConstForAll(Action &&action) const {
```
jeremyg-lunarg, line 864: I can't find where this method gets used. Did something get left out or did it get left in by accident?

jeremyg-lunarg: sounds like it is useful for debugging

jeremyg-lunarg:
I did mean to hit Approve on this before it got merged but apparently I didn't push the button hard enough.

Issue #4425, opened 2022-08-16
Pixel 6 crashes and failures


lunarpapillo:
We're preparing to add the Goole **Pixel_6** to CI.  Running VVL on a **Pixel_6** with the July 15, 2022 Google system update:
```
[ro.build.description]: [oriole-user 12 SQ3A.220705.004 8836240 release-keys]
```
provokes 13 crashes and 16 failures.  These are being removed from CI testing until VVL engineers have opportunity to address them.  When the issues have all been addressed, please assign this back to the originator, to remove the CI blacklist.

#### 13 crashing tests
- `VkLayerTest.AndroidHardwareBufferMemoryAllocation`
- `VkLayerTest.AndroidHardwareBufferImportBuffer`
- `VkLayerTest.AndroidHardwareBufferExporttBuffer` (should fix spelling on this too)
- `VkLayerTest.AndroidHardwareBufferImportBufferHandleType`
- `VkLayerTest.InlineUniformBlockEXT`
- `VkLayerTest.TransformFeedbackCmdBindTransformFeedbackBuffersEXT`
- `VkLayerTest.TransformFeedbackCmdBeginTransformFeedbackEXT`
- `VkLayerTest.TransformFeedbackCmdEndTransformFeedbackEXT`
- `VkGpuAssistedLayerTest.GpuValidationInlineUniformBlockAndMiscGpu`
- `VkPositiveLayerTest.AndroidHardwareBufferDepthStencil`
- `VkPositiveLayerTest.AndroidHardwareBufferBindBufferMemory`
- `VkPositiveLayerTest.OpCopyObjectSampler`
- `VkPositiveLayerTest.TestShaderInputAndOutputStructComponents`

#### 16 failing tests
- `VkLayerTest.CreatePipelineLayoutExcessPerStageDescriptors`
- `VkLayerTest.CreatePipelineLayoutExcessDescriptorsOverall`
- `VkLayerTest.CreateImageMaxLimitsViolation`
- `VkLayerTest.AndroidHardwareBufferInvalidBindBufferMemory`
- `VkLayerTest.ImageDrmFormatModifer`
- `VkLayerTest.FramebufferCreateErrors`
- `VkLayerTest.DrawIndirectByteCountEXT`
- `VkLayerTest.InvalidMixingProtectedResources`
- `VkLayerTest.BindImageMemorySwapchain`
[...]

PR #4429, opened 2022-08-17
tests: Use new GetPhysicalDevice helpers

sjfricke:
should help address some failures in https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4425

### DIFF
--- a/tests/positive/command.cpp
+++ b/tests/positive/command.cpp
@@ -602,8 +602,7 @@ TEST_F(VkPositiveLayerTest, DrawIndirectCountWithFeature) {
     auto features12 = LvlInitStruct<VkPhysicalDeviceVulkan12Features>();
     features12.drawIndirectCount = VK_TRUE;
     auto features2 = LvlInitStruct<VkPhysicalDeviceFeatures2>(&features12);
-
-    vk::GetPhysicalDeviceFeatures2(gpu(), &features2);
+    GetPhysicalDeviceFeatures2(features2);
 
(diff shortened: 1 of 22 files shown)


tests/positive/command.cpp
```diff
@@ -602,8 +602,7 @@ TEST_F(VkPositiveLayerTest, DrawIndirectCountWithFeature) {
     auto features12 = LvlInitStruct<VkPhysicalDeviceVulkan12Features>();
     features12.drawIndirectCount = VK_TRUE;
     auto features2 = LvlInitStruct<VkPhysicalDeviceFeatures2>(&features12);
-
-    vk::GetPhysicalDeviceFeatures2(gpu(), &features2);
+    GetPhysicalDeviceFeatures2(features2);
```
ncesario-lunarg, line 605: No change needed, but just an FYI, this function is specialized such that if you pass in a non-`VkPhysicalDeviceFeatures2` struct, one will be created and returned with the struct you do pass in set to the pNext of the PDF2 struct. IOW, you can remove the previous line and just do `auto features2 = GetPhysicalDeviceFeatures2(features12);`.

ncesario-lunarg (approved):
LGTM, thanks!

lunarpapillo:
FYI, the LunarG CI run will fail due to machine lockup.  I'm restarting.

lunarpapillo:
And the last failure *also* seems to be a CI machine problem.  Restarting.

PR #4431, opened 2022-08-17
build: Disable -Wnullability-completeness for VMA

ShabbyX:
Fixes build in Chromium


ncesario-lunarg (approved):

lunarpapillo:
The LunarG CI failure seems to be due to a machine issue.  Restarting.

PR #4432, opened 2022-08-17
Fix new VMA by supplying Get*ProcAddr

Tony-LunarG:
Fixes GPU-AV on RADV


ncesario-lunarg (approved):

ncesario-lunarg:
Looks like this will also fix things on lavapipe as well.

Tony-LunarG:
> Looks like this will also fix things on lavapipe as well.

Cool!  Thanks for your help!

PR #4434, opened 2022-08-17
layers: Add zero-size alloc case

ncesario-lunarg:
Adds an allowed zero-sized allocation case based on VUID 01874.


jeremyg-lunarg (approved):
LGTM

Tony-LunarG (approved):
LGTM

sjfricke (approved):
This is my least favorite part of the spec, but after caching back in how this code worked and reading the VU multiple times, seems like the right fix to me!

PR #4437, opened 2022-08-18
layers: Remove todo and code workaround

sjfricke:
I added this workaround in the original version of the script, can't see the gitlab, but XML seems updated to `16` now and this can be safely removed

### DIFF
--- a/scripts/format_utils_generator.py
+++ b/scripts/format_utils_generator.py
@@ -206,12 +206,6 @@ extern "C" {
             'components' : []
         }
 
-        # TODO - The current XML has a bug, remove when new headers with fix are added
-        # Was setting theses as 8 instead of 16
-        # https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/4975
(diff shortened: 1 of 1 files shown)


ncesario-lunarg:
IIRC, we would be seeing test failures (on Android I think) if the issue was _not_ fixed in the XML?

ncesario-lunarg (approved):
LGTM, but just a clarifying question to make sure we were hitting the case where this would be a problem in our tests.

sjfricke:
@ncesario-lunarg I know this was old, did my research, 
https://github.com/KhronosGroup/Vulkan-Docs/commit/ac23aa229fd9b8ea06aa99cf07b79cdc15af406f

>   * Correct etext:VK_FORMAT_ETC2_R8G8B8A8_*_BLOCK block sizes to 128 bits in
    the XML format tags (internal merge request 4975).

it was fixed in 1.2.203 spec for safe to know it is there now correctly

PR #4438, opened 2022-08-18
layers: Use helper for CooperativeMatrix

sjfricke:
I don't have a `SPV_NV_cooperative_matrix` capable device currently to test, but assuming CI passes this should be a trivial cleanup using the grammar file helper

### DIFF
--- a/layers/shader_validation.cpp
+++ b/layers/shader_validation.cpp
@@ -1327,38 +1327,8 @@ bool CoreChecks::ValidateCooperativeMatrix(const SHADER_MODULE_STATE &module_sta
     bool seen_coopmat_capability = false;
 
     for (auto insn : module_state) {
-        // Whitelist instructions whose result can be a cooperative matrix type, and
-        // keep track of their types. It would be nice if SPIRV-Headers generated code
-        // to identify which instructions have a result type and result id. Lacking that,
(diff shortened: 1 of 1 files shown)


layers/shader_validation.cpp
```diff
@@ -1327,38 +1327,8 @@ bool CoreChecks::ValidateCooperativeMatrix(const SHADER_MODULE_STATE &module_sta
...
-            case spv::OpConvertSToF:
-            case spv::OpConvertUToF:
-            case spv::OpUConvert:
-            case spv::OpSConvert:
-            case spv::OpFConvert:
-                id_to_type_id[insn.word(2)] = insn.word(1);
-                break;
-            default:
-                break;
+        if (OpcodeHasType(insn.opcode()) && OpcodeHasResult(insn.opcode())) {
```
ncesario-lunarg, line 1330: Looks like this will include a bunch more cases that weren't covered before, is that correct?

ncesario-lunarg (approved):
LGTM!

layers/shader_validation.cpp
sjfricke: (I assume Jeff B wrote this) and from the comment wanted a way to check if the opcode had a result/type but instead manually just wrote what he knew would be seen

PR #4255, opened 2022-06-23
Add runtime SPIRV 8/16bit features

sjfricke:
Addresses #3288 
Adds validation for

- VUID-RuntimeSpirv-storageBuffer16BitAccess-06331
- VUID-RuntimeSpirv-storageBuffer8BitAccess-06328
- VUID-RuntimeSpirv-storageInputOutput16-06334
- VUID-RuntimeSpirv-storagePushConstant16-06333
- VUID-RuntimeSpirv-storagePushConstant8-06330
- VUID-RuntimeSpirv-uniformAndStorageBuffer16BitAccess-06332
- VUID-RuntimeSpirv-uniformAndStorageBuffer8BitAccess-06329

----

For the tests there is a nasty subtleness, if you in GLSL do

```glsl
#extension GL_EXT_shader_8bit_storage: enable
layout(set = 0, binding = 0) buffer SSBO { int8_t x; } data;
```

you get

```swift
OpCapability Int8
OpCapability StorageBuffer8BitAccess
OpExtension "SPV_KHR_8bit_storage"
```

but the question becomes if something not GLSL generates
```swift
OpCapability Int8
OpExtension "SPV_KHR_8bit_storage"
```

This is still a 100% legal/valid SPIR-V module as `Int8` is a superset of `StorageBuffer8BitAccess`

----

The original tests (now named `VkLayerTest.Storage8and16bitCapability`) will be triggered from the auto-generated SPIR-V capability checking. The new test will only get the runtime error as there are no invalid capabilities being used

PR #4439, opened 2022-08-18
layers: Handle OpTypeBool in structs

sjfricke:
address issue in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4255#issuecomment-1219214198

### DIFF
--- a/layers/shader_module.cpp
+++ b/layers/shader_module.cpp
@@ -1891,7 +1891,8 @@ uint32_t SHADER_MODULE_STATE::GetTypeBytesSize(const spirv_inst_iter &iter) cons
 // Returns the base type (float, int or unsigned int) or struct (can have multiple different base types inside)
 uint32_t SHADER_MODULE_STATE::GetBaseType(const spirv_inst_iter &iter) const {
     const uint32_t opcode = iter.opcode();
-    if (opcode == spv::OpTypeFloat || opcode == spv::OpTypeInt || opcode == spv::OpTypeStruct) {
+    if (opcode == spv::OpTypeFloat || opcode == spv::OpTypeInt || opcode == spv::OpTypeBool || opcode == spv::OpTypeStruct) {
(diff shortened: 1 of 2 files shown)


layers/shader_module.cpp
```diff
@@ -1906,6 +1907,9 @@ uint32_t SHADER_MODULE_STATE::GetBaseType(const spirv_inst_iter &iter) const {
         const auto& type = get_def(iter.word(3));
         return GetBaseType(type);
     }
+    // if we assert, we are missing a valid base type and need to handle or else returning zero will produce a hard bug to track
+    // down as a value of zero means nothing useful
```
ncesario-lunarg, line 1911: Had to read this a couple times to understand it. Would this also be accurate to say?
```
If we assert here, we are missing a valid base type that must be handled. Without this assert, a return value of 0 will produce a hard bug to track...
```
This also makes me wonder if we can/should assert in `SHADER_MODULE_STATE::get_def` that `id != 0`, but I didn't see anything explicitly stating result IDs in SPIRV must be > 0.

ncesario-lunarg (approved):
One comment nit and question about SPIR-V IDs, but otherwise LGTM!

Tony-LunarG (approved):
LGTM and fixes the CTS tests I saw problems with

layers/shader_module.cpp
sjfricke: > Would this also be accurate to say?

Yes, will update

> if we can/should assert in SHADER_MODULE_STATE::get_def

I guess I have no idea if anywhere in the code is relying on `get_def` to return `0` to know it is invalid... I think this is a bigger issue to tackle, I do see things like 

```cpp
auto def = get_def(id);
assert(def != end());
```

but also see things like `GetConstantDef` check and pass back the `end()` to it's caller

I think before we assert at that level, someone (happy to do) will need to examine the 113 cases that currently call it and make sure... might be worth just creating a proper issue for this

PR #4441, opened 2022-08-18
layers: Fix hangs in NV best practices

MarkY-LunarG:
The best practice layer had a few cases wher GetRead<> was called
sequentially and resulted in a hang as it waited on a lock that
the callstack it was already in had already locked on.

### DIFF
--- a/layers/best_practices_utils.cpp
+++ b/layers/best_practices_utils.cpp
@@ -1707,9 +1707,12 @@ bool BestPractices::ValidateCmdPipelineBarrierImageBarrier(VkCommandBuffer comma
 
     bool skip = false;
 
+    const auto cmd_state = GetRead<bp_state::CommandBuffer>(commandBuffer);
+    assert(cmd_state);
+
     if (VendorCheckEnabled(kBPVendorNVIDIA)) {
         if (barrier.oldLayout == VK_IMAGE_LAYOUT_UNDEFINED && barrier.newLayout != VK_IMAGE_LAYOUT_UNDEFINED) {
-            skip |= ValidateZcull(commandBuffer, barrier.image, barrier.subresourceRange);
+            skip |= ValidateZcull(*cmd_state, barrier.image, barrier.subresourceRange);
         }
     }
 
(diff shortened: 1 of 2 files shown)


layers/best_practices_validation.h
```diff
@@ -1000,8 +1000,11 @@ class BestPractices : public ValidationStateTracker {
 
     void RecordZcullDraw(bp_state::CommandBuffer& cmd_state);
 
-    bool ValidateZcullScope(VkCommandBuffer commandBuffer) const;
-    bool ValidateZcull(VkCommandBuffer commandBuffer, VkImage image, const VkImageSubresourceRange& subresource_range) const;
+    bool ValidateZcullScope(VkCommandBuffer commandBuffer,
+                            const LockedSharedPtr<const bp_state::CommandBuffer, ReadLockGuard>& cmd_state) const;
+    bool ValidateZcull(VkCommandBuffer commandBuffer,
+                       const LockedSharedPtr<const bp_state::CommandBuffer, ReadLockGuard>& cmd_state, VkImage image,
```
jeremyg-lunarg, line 1006: This should be a `const  bp_state::CommandBuffer& cmd_state`

layers/best_practices_utils.cpp
```diff
@@ -4772,9 +4783,11 @@ bool BestPractices::PreCallValidateCmdClearDepthStencilImage(VkCommandBuffer com
                     "vkCmdClearAttachments instead",
                     VendorSpecificTag(kBPVendorAMD));
     }
+    const auto cmd_state = GetRead<bp_state::CommandBuffer>(commandBuffer);
+    assert(cmd_state);
     if (VendorCheckEnabled(kBPVendorNVIDIA)) {
         for (uint32_t i = 0; i < rangeCount; i++) {
-            skip |= ValidateZcull(commandBuffer, image, pRanges[i]);
+            skip |= ValidateZcull(commandBuffer, cmd_state, image, pRanges[i]);
```
jeremyg-lunarg, line 4790: This should be *cmd_state

layers/best_practices_validation.h
jeremyg-lunarg: Also, just get rid of the commandBuffer parameter while you're here.  It'll require more changes in the method body but it makes it clearer that you already have a state object so you don't need to call Get<>/GetRead<>.

jeremyg-lunarg (changes requested):

layers/best_practices_validation.h
MarkY-LunarG: 👍🏽

layers/best_practices_utils.cpp
MarkY-LunarG: 👍🏽

ncesario-lunarg (approved):
LGTM! I wonder how we never ran into this on our Linux + Nvidia CI setup?

jeremyg-lunarg (approved):
LGTM

jeremyg-lunarg:
> LGTM! I wonder how we never ran into this on our Linux + Nvidia CI setup?

One possibility is that we use std::mutex when compiling for c++11, but std::shared_timed_mutex with c++14 or std::shared_mutex with c++17 or higher. Multiple calls to GetRead<> will deadlock with std::mutex but not with the r/w locks.   Or it could be something completely different.

ncesario-lunarg:
> > LGTM! I wonder how we never ran into this on our Linux + Nvidia CI setup?
> 
> One possibility is that we use std::mutex when compiling for c++11, but std::shared_timed_mutex with c++14 or std::shared_mutex with c++17 or higher. Multiple calls to GetRead<> will deadlock with std::mutex but not with the r/w locks. Or it could be something completely different.

Right, my understanding is that the existing code should have been fine on C++17 or higher (maybe C++14, too?) due to shared_mutex, but I thought our CI machines were compiling against C++11, which I thought would have hit the same scenario?

jeremyg-lunarg:
> I thought our CI machines were compiling against C++11, which I thought would have hit the same scenario?

I thought so too. The only other possibility I can think of is the test gets skipped in CI.

PR #4442, opened 2022-08-18
layers: Add vendor check in ClearDepthStencilImage

### LINKED ISSUE #4440: `assert(VendorCheckEnabled(kBPVendorNVIDIA));` getting hit in Debug
We're seeing asserts in Debug builds on all platforms in "BestPractices::RecordResetZcullDirection" when not using the NVidia driver here:
[Vulkan-ValidationLayers/layers/best_practices_utils.cpp](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/27f1e075c41daf1b3c229b48f913eb5649b28530/layers/best_practices_utils.cpp#L2055)

`assert(VendorCheckEnabled(kBPVendorNVIDIA));` is getting hit.

 - OS: All
 - GPU: Any non NVidia
 - SDK or header version if building from repo: ToT
 - Options enabled (synchronization, best practices, etc.): Debug

rlocatti-nv:
Add NVIDIA vendor check for vkCmdClearDepthStencilImage before calling
RecordResetZcullDirection.

Attempts to fix #4440.


ncesario-lunarg (approved):
LGTM, thanks for the quick fix!

PR #4443, opened 2022-08-18
Use current renderpass to create pre-draw GPU-AV pipeline

Tony-LunarG:
Use the current renderpass to create the pre-draw pipeline instead of the renderpass used to create the currently bound pipeline, which may have been destroyed


ncesario-lunarg (approved):

jeremyg-lunarg (approved):
LGTM. Is this what was causing the DoomEternal trace to crash?

Tony-LunarG:
No, but it's what was causing Wolfenstein 2 to crash

PR #4445, opened 2022-08-19
tests: Fix ValidateImportMemoryHandleType test on Nvidia laptop

MarkY-LunarG:
Nvidia exposes multiple memory types that have 0 properties.  This
causes a problem if one of those memory types is first for the
ValidateImportMemoryHandleType test because the test will still
select it and then the allocation for the buffer and image
memory import memory will fail with VK_OUT_OF_DEVICE_MEMORY.
Fix this by forcing device local memory flag to be used when querying
a proper memory allocation for this test.

Verified this still works for integrated Intel as well.


ncesario-lunarg (approved):
LGTM!

MarkY-LunarG:
Thnx @ncesario-lunarg

PR #4447, opened 2022-08-19
Enable QueueSubmit2 and 2KHR in SyncVal

jzulauf-lunarg:
Refactored the Sync1 QueueSubmit code to support both paths.  Simplifying some of the template madness by simply converting the Sync 1 parameter to the Sync 2 set, and using plain common functions for the validate and record functions.

Some minor flattening of twoliner functions that were called once.

A simple Submit2 test case with test helper object refinements.

### DIFF
--- a/layers/synchronization_validation.cpp
+++ b/layers/synchronization_validation.cpp
@@ -3896,6 +3896,32 @@ QueueBatchContext::BatchSet SyncValidator::GetQueueBatchSnapshot() {
     return snapshot;
 }
 
+struct QueueSubmitCmdState {
+    std::shared_ptr<const QueueSyncState> queue;
+    std::shared_ptr<QueueBatchContext> last_batch;
+    std::string submit_func_name;
+    AccessLogger logger;
+    SignaledSemaphores signaled;
+    QueueSubmitCmdState(const char *func_name, const AccessLogger &parent_log, const SignaledSemaphores &parent_semaphores)
+        : submit_func_name(func_name), logger(&parent_log), signaled(parent_semaphores) {}
+};
+
+bool QueueBatchContext::DoQueueSubmitValidate(const SyncValidator &sync_state, QueueSubmitCmdState &cmd_state,
+                                              const VkSubmitInfo2 &batch_info) {
+    bool skip = false;
+
+    //  For each submit in the batch...
+    for (const auto &cb : command_buffers_) {
+        if (cb.cb->GetTagLimit() == 0) continue;  // Skip empty CB's
+        skip |= cb.cb->ValidateFirstUse(*this, cmd_state.submit_func_name.c_str(), cb.index);
+
+        // The barriers have already been applied in ValidatFirstUse
+        ResourceUsageRange tag_range = ImportRecordedAccessLog(*cb.cb);
+        ResolveSubmittedCommandBuffer(*cb.cb->GetCurrentAccessContext(), tag_range.begin);
+    }
+    return skip;
+}
+
 bool SignaledSemaphores::SignalSemaphore(const std::shared_ptr<const SEMAPHORE_STATE> &sem_state,
                                          const std::shared_ptr<QueueBatchContext> &batch,
                                          const VkSemaphoreSubmitInfo &signal_info) {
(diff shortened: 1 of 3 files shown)


layers/synchronization_validation.cpp
```diff
@@ -7237,27 +7263,23 @@ void SyncValidator::PostCallRecordDeviceWaitIdle(VkDevice device, VkResult resul
...
-        : logger(&parent_log), signaled(parent_semaphores) {}
-};
-
 template <>
 thread_local layer_data::optional<QueueSubmitCmdState> layer_data::TlsGuard<QueueSubmitCmdState>::payload_{};
 
 bool SyncValidator::PreCallValidateQueueSubmit(VkQueue queue, uint32_t submitCount, const VkSubmitInfo *pSubmits,
                                                VkFence fence) const {
+    SubmitInfoConverter submit_info(submitCount, pSubmits);
+    return ValidateQueueSubmit(queue, submitCount, submit_info.info2s.data(), fence, "VkQueueSubmit");
```
jeremyg-lunarg, line 7272: since submit_info.info2s is only used right here and only once, would it be better to have something like:
```
     std::vector<VkSubmitInfo> info2s = submit_info.GetSubmitInfoVec();
```

tests/vksyncvaltests.cpp
```diff
@@ -4418,71 +4428,85 @@ void QSTestContext::Submit(VkQueue q, VkCommandBufferObj& cb, VkSemaphore wait,
...
 
-    m_device->wait();
+    test.DeviceWait();
 }
 
-TEST_F(VkSyncValTest, SyncQSBufferCopyVsIdle) {
-    // TODO (jzulauf)
-    // GTEST_SKIP() << "this test is causing a sporadic crash on nvidia 32b release. Skip until further investigation";
+TEST_F(VkSyncValTest, SyncQSSubmit2) {
+    if (VK_API_VERSION_1_3 != SetTargetApiVersion(VK_API_VERSION_1_3)) {
```
jeremyg-lunarg, line 4507: I think this check isn't correct which is causing the CI crashes. Instead you should have something like:
```
    ASSERT_NO_FATAL_FAILURE(Init());
    if (DeviceValidationVersion() < VK_API_VERSION_1_3) {
        GTEST_SKIP() << "At least Vulkan version 1.3 is required";
    }

```

jeremyg-lunarg (changes requested):

layers/synchronization_validation.cpp
jzulauf-lunarg: Noting that each info needs backing stores for its p<Foo>Infos with invariant pointers ... the object seemed the best way to organize them

layers/synchronization_validation.h
```diff
@@ -1456,23 +1459,16 @@ class QueueBatchContext : public CommandExecutionContext {
...
-    QueueBatchContext(const SyncValidator &sync_state, const QueueSyncState &queue_state);
-    QueueBatchContext() = delete;
+    void SetupAccessContext(const std::shared_ptr<const QueueBatchContext> &prev, const VkSubmitInfo2 &submit_info,
+                            SignaledSemaphores &signaled_semaphores);
+    void SetupCommandBufferInfo(const VkSubmitInfo2 &submit_info);
 
-    template <typename BatchInfo>
-    void Setup(const std::shared_ptr<const QueueBatchContext> &prev_batch, const BatchInfo &batch_info,
-               SignaledSemaphores &signaled);
+    bool DoQueueSubmtValidate(const SyncValidator &sync_state, QueueSubmitCmdState &cmd_state, const VkSubmitInfo2 &submit_info);
```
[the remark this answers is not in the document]
jzulauf-lunarg: Fixed I literally could *not* spell Submit all day that day.  Sumbit was the most common form...

layers/synchronization_validation.cpp
jeremyg-lunarg: Where is that requirement coming from? Maybe a comment explaining it is needed.

jeremyg-lunarg (approved):
LGTM

layers/synchronization_validation.cpp
jzulauf-lunarg: It's the nature of the info2s -- similar to what we do for the safe_Vk<foo> structs.  I'll add a comment though.

jzulauf-lunarg: Comment added @jeremyg-lunarg

jeremyg-lunarg (approved):

Issue #4312, opened 2022-07-14
Update descriptor/render pass layout validation


mikes-lunarg:
The following VUIDs are no longer in the spec and should be updated in validation:
```
  VUID-VkAttachmentDescription-format-03288
  VUID-VkAttachmentDescription-format-03289
  VUID-VkAttachmentDescription2-finalLayout-03061
  VUID-VkAttachmentDescription2-format-03294
  VUID-VkAttachmentDescription2-format-03295
  VUID-VkAttachmentDescription2-format-03296
  VUID-VkAttachmentDescription2-format-03297
  VUID-VkAttachmentDescription2-format-03300
  VUID-VkAttachmentDescription2-format-03301
  VUID-VkAttachmentDescription2-format-03302
  VUID-VkAttachmentDescription2-format-03303
  VUID-VkAttachmentDescription2-format-03304
  VUID-VkAttachmentDescription2-format-03305
  VUID-VkAttachmentDescription2-format-03306
  VUID-VkAttachmentDescription2-format-03307
  VUID-VkAttachmentDescription2-format-06701
  VUID-VkAttachmentDescription2-format-06702
  VUID-VkAttachmentDescription2-separateDepthStencilLayouts-03298
  VUID-VkAttachmentDescription2-separateDepthStencilLayouts-03299
  VUID-VkAttachmentDescription2-separateDepthStencilLayouts-06556
  VUID-VkAttachmentDescription2-separateDepthStencilLayouts-06557
  VUID-VkAttachmentReference-layout-00857
  VUID-VkAttachmentReference2-attachment-04754
  VUID-VkAttachmentReference2-attachment-04755
  VUID-VkAttachmentReference2-attachment-04756
  VUID-VkAttachmentReference2-attachment-04757
  VUID-VkSubpassDescription-None-04437
  VUID-VkSubpassDescription2-None-04439
```

This looks like it comes from fairly large valid usage refactor in this spec MR: https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/4686

MarkY-LunarG:
I found some additional info while working on this:

Removed (mentioned above):
```
  VUID-VkAttachmentDescription2-finalLayout-03061
  VUID-VkAttachmentDescription2-format-03294
  VUID-VkAttachmentDescription2-format-03295
  VUID-VkAttachmentDescription2-format-03296
  VUID-VkAttachmentDescription2-format-03297
[...]

PR #4448, opened 2022-08-19
Add validation and tests for VK_EXT_attachment_feedback_loop_layout

Tony-LunarG:
This will still get false positives for 04437/04439, but those have been removed from the spec and will be addressed by the solution to #4312

### DIFF
--- a/layers/buffer_validation.cpp
+++ b/layers/buffer_validation.cpp
@@ -452,6 +452,43 @@ bool CoreChecks::ValidateRenderPassLayoutAgainstFramebufferImageUsage(RenderPass
                          report_data->FormatHandle(image_view).c_str());
     }
 
+    if (layout == VK_IMAGE_LAYOUT_ATTACHMENT_FEEDBACK_LOOP_OPTIMAL_EXT) {
+        if (((image_usage & (VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT | VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT)) == 0) ||
+            ((image_usage & (VK_IMAGE_USAGE_SAMPLED_BIT | VK_IMAGE_USAGE_INPUT_ATTACHMENT_BIT)) == 0)) {
+            vuid = use_rp2 ? "VUID-vkCmdBeginRenderPass2-initialLayout-07002" : "VUID-vkCmdBeginRenderPass-initialLayout-07000";
+            LogObjectList objlist(image);
+            objlist.add(renderpass);
+            objlist.add(framebuffer);
+            objlist.add(image_view);
+            skip |=
+                LogError(objlist, vuid,
+                         "%s: Layout/usage mismatch for attachment %" PRIu32
+                         " in %s"
+                         " - the %s is %s but the image attached to %s via %s"
+                         " was not created with either the VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT or "
+                         "VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT usage bits, and the VK_IMAGE_USAGE_INPUT_ATTACHMENT_BIT or "
+                         "VK_IMAGE_USAGE_SAMPLED_BIT usage bits",
+                         function_name, attachment_index, report_data->FormatHandle(renderpass).c_str(), variable_name,
+                         string_VkImageLayout(layout), report_data->FormatHandle(framebuffer).c_str(),
+                         report_data->FormatHandle(image_view).c_str());
+        }
+        if (!(image_usage & VK_IMAGE_USAGE_ATTACHMENT_FEEDBACK_LOOP_BIT_EXT)) {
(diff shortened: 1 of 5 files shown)


layers/buffer_validation.cpp
```diff
@@ -827,9 +866,15 @@ bool CoreChecks::ValidateBarriersToImages(const Location &outer_loc, const CMD_B
         if (!image_state) {
             continue;
         }
-        VkImageUsageFlags usage_flags = image_state->createInfo.usage;
-        skip |= ValidateBarrierLayoutToImageUsage(loc.dot(Field::oldLayout), img_barrier.image, img_barrier.oldLayout, usage_flags);
-        skip |= ValidateBarrierLayoutToImageUsage(loc.dot(Field::newLayout), img_barrier.image, img_barrier.newLayout, usage_flags);
+
+        if ((img_barrier.srcQueueFamilyIndex != img_barrier.dstQueueFamilyIndex) ||
```
jeremyg-lunarg, line 870: this if guard seems like an important fix even without the new layout & usage. Was not having it causing false positives in CTS?

jeremyg-lunarg (approved):
LGTM

layers/buffer_validation.cpp
Tony-LunarG: I meant to mention that.  It looks like, at some point, all of the layout vs usage VUIDs (not just the new one) got prefaced with "If srcQueueFamilyIndex and dstQueueFamilyIndex define a queue family ownership transfer or oldLayout and newLayout define an image layout transition" so this was my attempt to implement that.  But I'm not at all sure that this is all there is to it and I'm open to suggestions.

layers/buffer_validation.cpp
```diff
@@ -452,6 +452,41 @@ bool CoreChecks::ValidateRenderPassLayoutAgainstFramebufferImageUsage(RenderPass
...
+        if (((image_usage & (VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT | VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT)) == 0) ||
+            ((image_usage & (VK_IMAGE_USAGE_SAMPLED_BIT | VK_IMAGE_USAGE_INPUT_ATTACHMENT_BIT)) == 0)) {
+            vuid = use_rp2 ? "VUID-vkCmdBeginRenderPass2-initialLayout-07002" : "VUID-vkCmdBeginRenderPass-initialLayout-07000";
+            LogObjectList objlist(image);
+            objlist.add(renderpass);
+            objlist.add(framebuffer);
+            objlist.add(image_view);
+            skip |=
+                LogError(objlist, vuid,
+                         "%s: Layout/usage mismatch for attachment %u in %s"
```
ncesario-lunarg, line 465: `%u` -> `PRIu32` (same for below)?

Every time I see this I'm tempted to stop everything and refactor the logging :'(

layers/buffer_validation.cpp
jeremyg-lunarg: It looks like you got it right to me.  Looks like it came in here https://github.com/KhronosGroup/Vulkan-Docs/commit/dd7b521af03ed3ce13b7d2a54c4542f5af7cf370  (see last bullet point in the commit message).  I think that this was done to get ready for synchronization2.

ncesario-lunarg (approved):
LGTM!

tests/vklayertests_descriptor_renderpass_framebuffer.cpp
```diff
@@ -2263,17 +2263,25 @@ TEST_F(VkLayerTest, RenderPassBeginLayoutsFramebufferImageUsageMismatches) {
...
         maintenance2Supported = true;
     }
 
-    ASSERT_NO_FATAL_FAILURE(InitState(nullptr, nullptr, VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT));
+    auto attachment_feedback_loop_layout_features =
+        LvlInitStruct<VkPhysicalDeviceAttachmentFeedbackLoopLayoutFeaturesEXT>();
+    auto features2 = LvlInitStruct<VkPhysicalDeviceFeatures2>(&attachment_feedback_loop_layout_features);
+    auto vkGetPhysicalDeviceFeatures2KHR = reinterpret_cast<PFN_vkGetPhysicalDeviceFeatures2KHR>(
+        vk::GetInstanceProcAddr(instance(), "vkGetPhysicalDeviceFeatures2KHR"));
+    vkGetPhysicalDeviceFeatures2KHR(gpu(), &features2);
```
paul-lunarg, line 2283: You could use VkLayerTest::GetPhysicalDeviceFeatures2 instead, which does exactly this and also checks for Vk1.1

paul-lunarg (approved):

layers/buffer_validation.cpp
Tony-LunarG: Ugh - I've been trying not to do that.  Really!  Fixed

tests/vklayertests_descriptor_renderpass_framebuffer.cpp
Tony-LunarG: Good idea - I forgot about that.  Fixed

PR #4450, opened 2022-08-20
layers: Refactor PreCallValidateCreateDescriptorSetLayout

sjfricke:
This function I feels like it was from the original style of the VVL but duct taped with more stuff

removed the unneeded indirect function passing in all the object pointers. Moved the extension enum check out to stateless as well

### DIFF
--- a/layers/descriptor_validation.cpp
+++ b/layers/descriptor_validation.cpp
@@ -353,147 +353,131 @@ bool CoreChecks::PreCallValidateCmdBindDescriptorSets(VkCommandBuffer commandBuf
     return skip;
 }
 
-// Validate descriptor set layout create info
-static bool ValidateDescriptorSetLayoutCreateInfo(
-    const ValidationObject *val_obj, const VkDescriptorSetLayoutCreateInfo *create_info, const bool push_descriptor_ext,
-    const uint32_t max_push_descriptors, const bool descriptor_indexing_ext,
-    const VkPhysicalDeviceVulkan12Features *core12_features, const VkPhysicalDeviceVulkan13Features *core13_features,
-    const VkPhysicalDeviceInlineUniformBlockPropertiesEXT *inline_uniform_block_props,
-    const VkPhysicalDeviceAccelerationStructureFeaturesKHR *acceleration_structure_features,
-    const DeviceExtensions *device_extensions) {
+bool CoreChecks::PreCallValidateCreateDescriptorSetLayout(VkDevice device, const VkDescriptorSetLayoutCreateInfo *pCreateInfo,
+                                                          const VkAllocationCallbacks *pAllocator,
+                                                          VkDescriptorSetLayout *pSetLayout) const {
     bool skip = false;
     layer_data::unordered_set<uint32_t> bindings;
     uint64_t total_descriptors = 0;
 
-    const auto *flags_create_info = LvlFindInChain<VkDescriptorSetLayoutBindingFlagsCreateInfo>(create_info->pNext);
-
-    const bool push_descriptor_set = !!(create_info->flags & VK_DESCRIPTOR_SET_LAYOUT_CREATE_PUSH_DESCRIPTOR_BIT_KHR);
-    if (push_descriptor_set && !push_descriptor_ext) {
-        skip |= val_obj->LogError(
-            val_obj->device, kVUID_Core_DrawState_ExtensionNotEnabled,
-            "vkCreateDescriptorSetLayout(): Attempted to use %s in %s but its required extension %s has not been enabled.\n",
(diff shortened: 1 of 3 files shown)


layers/parameter_validation_utils.cpp
```diff
@@ -4319,7 +4319,7 @@ bool StatelessValidation::manual_PreCallValidateCreateDescriptorSetLayout(VkDevi
         mutable_descriptor_type_features && mutable_descriptor_type_features->mutableDescriptorType == VK_TRUE;
 
     // Validation for parameters excluded from the generated validation code due to a 'noautovalidity' tag in vk.xml
-    if ((pCreateInfo != nullptr) && (pCreateInfo->pBindings != nullptr)) {
```
sjfricke, line 4322: we dereference `pCreateInfo` a few lines above this, so it obviously wasn't ever being `null` (nor should it be tbh)

layers/descriptor_validation.cpp
```diff
@@ -653,47 +636,34 @@ static bool ValidateDescriptorSetLayoutCreateInfo(
...
+            skip |= LogError(device, "VUID-VkDescriptorSetLayoutCreateInfo-descriptorType-03001",
+                             "vkCreateDescriptorSetLayout(): binding (%" PRIi32
+                             ") has VK_DESCRIPTOR_BINDING_UPDATE_AFTER_BIND_BIT "
+                             "flag, but binding (%" PRIi32 ") has descriptor type VK_DESCRIPTOR_TYPE_STORAGE_BUFFER_DYNAMIC.",
+                             update_after_bind, storage_buffer_dynamic);
         }
     }
 
-    if ((push_descriptor_set) && (total_descriptors > max_push_descriptors)) {
-        const char *undefined = push_descriptor_ext ? "" : " -- undefined";
```
sjfricke, line 676: removed this `undefined` as we if the extension is not enabled it will fail in stateless layer before ever getting here

layers/parameter_validation_utils.cpp
```diff
@@ -4401,28 +4401,46 @@ bool StatelessValidation::manual_PreCallValidateCreateDescriptorSetLayout(VkDevi
...
-        }
-        if (pCreateInfo->flags & VK_DESCRIPTOR_SET_LAYOUT_CREATE_HOST_ONLY_POOL_BIT_VALVE &&
-            !mutable_descriptor_type_features_enabled) {
-            skip |= LogError(device, "VUID-VkDescriptorSetLayoutCreateInfo-flags-04596",
-                             "vkCreateDescriptorSetLayout(): pCreateInfo->flags contains "
-                             "VK_DESCRIPTOR_SET_LAYOUT_CREATE_HOST_ONLY_POOL_BIT_VALVE, but "
-                             "VkPhysicalDeviceMutableDescriptorTypeFeaturesVALVE::mutableDescriptorType feature is not enabled.");
-        }
+
+    // TODO - Remove these 2 extension checks once the enum-to-extensions logic is generated
```
ncesario-lunarg, line 4405: This one https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4228?

ncesario-lunarg (approved):
This LGTM, but would be good if @jeremyg-lunarg takes a look as I think he did some refactoring related to this.

layers/parameter_validation_utils.cpp
sjfricke: Ya, I'm assuming these were explicitly added for a reason, but these would be replaced by that PR in theory

jeremyg-lunarg (approved):

ncesario-lunarg:
For some reason this is failing on the following test (mesa RADV and recent nvidia drivers):
```
[ RUN      ] VkLayerTest.MutableDescriptorSetLayout
Unexpected: Validation Error: [ UNASSIGNED-CoreValidation-DrawState-ExtensionNotEnabled ] Object 0: handle = 0x7af4030, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x1382f0d8 | vkCreateDescriptorSetLayout(): Attemped to use VK_DESCRIPTOR_SET_LAYOUT_CREATE_UPDATE_AFTER_BIND_POOL_BIT in VkDescriptorSetLayoutCreateInfo::flags but its required extension VK_EXT_descriptor_indexing has not been enabled.
```
Not immediately sure why this change has this affect, though.

sjfricke:
@ncesario-lunarg that is failing from the https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4228 thing I mentioned above

So I moved the extension check out from CoreValdation to StatelessValidation. If we hit something in Stateless, it fails early instead of going into core... will fix the `MutableDescriptorSetLayout` test

PR #4451, opened 2022-08-21
Fix queries in render pass with multiview

ziga-lunarg:
From the spec:

> If queries are used while executing a render pass instance that has multiview enabled, the query uses N consecutive query indices in the query pool (starting at query) where N is the number of bits set in the view mask in the subpass the query is used in.

### DIFF
--- a/layers/state_tracker.cpp
+++ b/layers/state_tracker.cpp
@@ -3098,30 +3098,47 @@ void ValidationStateTracker::PreCallRecordCmdPipelineBarrier2(VkCommandBuffer co
 void ValidationStateTracker::PostCallRecordCmdBeginQuery(VkCommandBuffer commandBuffer, VkQueryPool queryPool, uint32_t slot,
                                                          VkFlags flags) {
     if (disabled[query_validation]) return;
-
-    QueryObject query = {queryPool, slot};
     auto cb_state = GetWrite<CMD_BUFFER_STATE>(commandBuffer);
-    cb_state->RecordCmd(CMD_BEGINQUERY);
-    if (!disabled[query_validation]) {
-        cb_state->BeginQuery(query);
+
+    uint32_t num_queries = 1;
+    // If render pass instance has multiview enabled, query uses N consecutive query indices
+    if (cb_state->activeRenderPass) {
+        uint32_t bits = cb_state->activeRenderPass->GetViewMaskBits(cb_state->activeSubpass);
+        num_queries = std::max(num_queries, bits);
     }
-    if (!disabled[command_buffer_state]) {
-        auto pool_state = Get<QUERY_POOL_STATE>(query.pool);
-        cb_state->AddChild(pool_state);
+    for (uint32_t i = 0; i < num_queries; ++i) {
+        QueryObject query = {queryPool, slot};
+        cb_state->RecordCmd(CMD_BEGINQUERY);
+        if (!disabled[query_validation]) {
+            cb_state->BeginQuery(query);
+        }
+        if (!disabled[command_buffer_state]) {
+            auto pool_state = Get<QUERY_POOL_STATE>(query.pool);
+            cb_state->AddChild(pool_state);
+        }
     }
 }
 
 void ValidationStateTracker::PostCallRecordCmdEndQuery(VkCommandBuffer commandBuffer, VkQueryPool queryPool, uint32_t slot) {
     if (disabled[query_validation]) return;
-    QueryObject query_obj = {queryPool, slot};
     auto cb_state = GetWrite<CMD_BUFFER_STATE>(commandBuffer);
-    cb_state->RecordCmd(CMD_ENDQUERY);
-    if (!disabled[query_validation]) {
-        cb_state->EndQuery(query_obj);
+    uint32_t num_queries = 1;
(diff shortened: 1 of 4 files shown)


tests/positive/render_pass.cpp
```diff
@@ -1132,3 +1132,121 @@ TEST_F(VkPositiveLayerTest, BeginRenderPassWithViewMask) {
...
+TEST_F(VkPositiveLayerTest, QueriesInMultiviewRenderPass) {
+    TEST_DESCRIPTION("Use queries in a render pass instance with multiview enabled.");
+
+    SetTargetApiVersion(VK_API_VERSION_1_2);
+    ASSERT_NO_FATAL_FAILURE(InitFramework());
+
+    auto vulkan_11_features = LvlInitStruct<VkPhysicalDeviceVulkan11Features>();
+    auto features2 = LvlInitStruct<VkPhysicalDeviceFeatures2>(&vulkan_11_features);
+
+    vk::GetPhysicalDeviceFeatures2(gpu(), &features2);
```
ncesario-lunarg, line 1145: This should probably move below the `DeviceValidationVersion` check.

Also, I think @sjfricke just went through some cleanup to use `GetPhysicalDeviceFeatures2` (i.e., the test class member function) as it will attempt to use the correct function (core vs KHR) based on API version.  So something like `auto features2 = GetPhysicalDeviceFeatures2(vulkan_11_features);`.

tests/positive/render_pass.cpp
```diff
@@ -1132,3 +1132,121 @@ TEST_F(VkPositiveLayerTest, BeginRenderPassWithViewMask) {
...
+
+    uint32_t viewMasks[] = {0x3u};
+    uint32_t correlationMasks[] = {0x1u};
+    auto rpmvci = LvlInitStruct<VkRenderPassMultiviewCreateInfo>();
+    rpmvci.subpassCount = 1;
+    rpmvci.pViewMasks = viewMasks;
+    rpmvci.correlationMaskCount = 1;
+    rpmvci.pCorrelationMasks = correlationMasks;
+
+    VkRenderPassCreateInfo rpci = {VK_STRUCTURE_TYPE_RENDER_PASS_CREATE_INFO, &rpmvci, 0, 1, &attachment, 1, &subpass, 0, nullptr};
```
ncesario-lunarg, line 1178: `LvlInitStruct` for consistency (same comment for instances below)?

ncesario-lunarg (commented):
A couple minor nits, but otherwise LGTM!

ncesario-lunarg (approved):

layers/state_tracker.cpp
```diff
@@ -4102,18 +4119,36 @@ void ValidationStateTracker::PostCallRecordGetPhysicalDeviceDisplayPlaneProperti
...
+    }
 }
 
 void ValidationStateTracker::PostCallRecordCmdEndQueryIndexedEXT(VkCommandBuffer commandBuffer, VkQueryPool queryPool,
                                                                  uint32_t query, uint32_t index) {
-    QueryObject query_obj = {queryPool, query, index};
     auto cb_state = GetWrite<CMD_BUFFER_STATE>(commandBuffer);
-    cb_state->RecordCmd(CMD_ENDQUERYINDEXEDEXT);
-    cb_state->EndQuery(query_obj);
+    uint32_t N = 1;
```
jeremyg-lunarg, line 4140: might be better to have this as num_queries rather than N

layers/state_tracker.cpp
```diff
@@ -4102,18 +4119,36 @@ void ValidationStateTracker::PostCallRecordGetPhysicalDeviceDisplayPlaneProperti
 
 void ValidationStateTracker::PostCallRecordCmdBeginQueryIndexedEXT(VkCommandBuffer commandBuffer, VkQueryPool queryPool,
                                                                    uint32_t query, VkQueryControlFlags flags, uint32_t index) {
-    QueryObject query_obj = {queryPool, query, index};
     auto cb_state = GetWrite<CMD_BUFFER_STATE>(commandBuffer);
-    cb_state->RecordCmd(CMD_BEGINQUERYINDEXEDEXT);
-    cb_state->BeginQuery(query_obj);
+    uint32_t N = 1;
```
jeremyg-lunarg, line 4123: might be better to have this as num_queries rather than N

layers/state_tracker.cpp
```diff
@@ -3098,30 +3098,47 @@ void ValidationStateTracker::PreCallRecordCmdPipelineBarrier2(VkCommandBuffer co
...
                                                          VkFlags flags) {
     if (disabled[query_validation]) return;
-
-    QueryObject query = {queryPool, slot};
     auto cb_state = GetWrite<CMD_BUFFER_STATE>(commandBuffer);
-    cb_state->RecordCmd(CMD_BEGINQUERY);
-    if (!disabled[query_validation]) {
-        cb_state->BeginQuery(query);
+
+    uint32_t N = 1;
```
jeremyg-lunarg, line 3103: might be better to have this as num_queries rather than N

jeremyg-lunarg (approved):
Code looks correct to me. I worry about the local variables named N as they don't really follow the coding standard.

PR #4454, opened 2022-08-22
Fix ShaderModuleIdentifier test

Tony-LunarG:
I'm finding that some drivers wont create a pipeline from a shader module id without creating one with the shader module

### DIFF
--- a/tests/vklayertests_pipeline_shader.cpp
+++ b/tests/vklayertests_pipeline_shader.cpp
@@ -17594,14 +17594,31 @@ TEST_F(VkLayerTest, ShaderModuleIdentifier) {
     pipe.CreateGraphicsPipeline();
     m_errorMonitor->VerifyFound();
 
-    // Now really create a pipeline with a smid
     sm_id_create_info.identifierSize = get_identifier.identifierSize;
+    // Trying to create a pipeline with a shader module identifier at this point can result in a VK_PIPELINE_COMPILE_REQUIRED from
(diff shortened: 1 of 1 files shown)


tests/vklayertests_pipeline_shader.cpp
```diff
@@ -17594,14 +17594,31 @@ TEST_F(VkLayerTest, ShaderModuleIdentifier) {
     pipe.CreateGraphicsPipeline();
     m_errorMonitor->VerifyFound();
 
-    // Now really create a pipeline with a smid
     sm_id_create_info.identifierSize = get_identifier.identifierSize;
+    // Trying to create a pipeline with a shader module identifier at this point can result in a VK_PIPELINE_COMPILE_REQUIRED from
```
ncesario-lunarg, line 17598: Is this something that you think should be clarified in the spec? Or this is more a lack of driver support?

ncesario-lunarg (approved):

tests/vklayertests_pipeline_shader.cpp
Tony-LunarG: I think it's an artifact of the way I'm using shader module ids in the test.  In "real" usage, on one run you'd use shader modules to create all your pipelines and save the associated shader module ids to use in subsequent runs.  Since the test is only running once, it has to get the shader module id and use it in the same run, which isn't the envisioned workflow for the extension.  I guess I'm surprised that it worked without creating a pipeline using a module, and this is what the test should have been doing all along.

PR #4458, opened 2022-08-23
tests: Fix SpirvGroupDecorations test

sjfricke:
The shader from `spv_source` in the test was never actually being used in the test, Luckily it seems to work correctly locally.

Not sure how this never triggered any `-Wunused-variable` warning (including locally)

### DIFF
--- a/layers/shader_module.cpp
+++ b/layers/shader_module.cpp
@@ -491,6 +491,10 @@ void SHADER_MODULE_STATE::PreprocessShaderBinary(const spv_target_env env) {
             // **THIS ONLY HAPPENS ON INITIALIZATION**. words should remain const for the lifetime
             // of the SHADER_MODULE_STATE instance.
             *const_cast<std::vector<uint32_t> *>(&words) = std::move(optimized_binary);
+            // Will need to update static data now the words have changed or else the def_index will not align
+            // It is really rare this will get here as Group Decorations have been deprecated and before this was added no one ever
+            // raised an issue for a bug that would crash the layers that was around for many releases
+            *const_cast<SpirvStaticData *>(&static_data_) = SpirvStaticData(*this);
         }
     }
 }
--- a/tests/positive/shaderval.cpp
+++ b/tests/positive/shaderval.cpp
@@ -809,15 +809,15 @@ TEST_F(VkPositiveLayerTest, SpirvGroupDecorations) {
         dslb[i].stageFlags = VK_SHADER_STAGE_COMPUTE_BIT | VK_SHADER_STAGE_ALL;
     }
     if (m_device->props.limits.maxPerStageDescriptorStorageBuffers < dslb_size) {
-        printf("%sNeeded storage buffer bindings exceeds this devices limit.  Skipping tests.\n", kSkipPrefix);
-        return;
+        GTEST_SKIP() << "Needed storage buffer bindings (" << dslb_size << ") exceeds this devices limit of "
+                     << m_device->props.limits.maxPerStageDescriptorStorageBuffers;
     }
 
     CreateComputePipelineHelper pipe(*this);
     pipe.InitInfo();
     pipe.dsl_bindings_.resize(dslb_size);
     memcpy(pipe.dsl_bindings_.data(), dslb, dslb_size * sizeof(VkDescriptorSetLayoutBinding));
-    pipe.cs_.reset(new VkShaderObj(this, bindStateMinimalShaderText, VK_SHADER_STAGE_COMPUTE_BIT));
+    pipe.cs_.reset(new VkShaderObj(this, spv_source.c_str(), VK_SHADER_STAGE_COMPUTE_BIT, SPV_ENV_VULKAN_1_0, SPV_SOURCE_ASM));
     pipe.InitState();
     pipe.CreateComputePipeline();
 }
--- a/layers/shader_module.h
+++ b/layers/shader_module.h
@@ -234,6 +234,8 @@ struct SHADER_MODULE_STATE : public BASE_NODE {
     struct SpirvStaticData {
         SpirvStaticData() = default;
         SpirvStaticData(const SHADER_MODULE_STATE &module_state);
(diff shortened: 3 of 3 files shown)


ncesario-lunarg:
> The shader from `spv_source` in the test was never actually being used in the test, Luckily it seems to work correctly locally.
> 
> Not sure how this never triggered any `-Wunused-variable` warning (including locally)

Something weird is going on here. _All_ debug builds are crashing on that test now in our CI.

tests/positive/shaderval.cpp
```diff
@@ -817,7 +817,7 @@ TEST_F(VkPositiveLayerTest, SpirvGroupDecorations) {
     pipe.InitInfo();
     pipe.dsl_bindings_.resize(dslb_size);
     memcpy(pipe.dsl_bindings_.data(), dslb, dslb_size * sizeof(VkDescriptorSetLayoutBinding));
-    pipe.cs_.reset(new VkShaderObj(this, bindStateMinimalShaderText, VK_SHADER_STAGE_COMPUTE_BIT));
+    pipe.cs_.reset(new VkShaderObj(this, spv_source.c_str(), VK_SHADER_STAGE_COMPUTE_BIT, SPV_ENV_VULKAN_1_0, SPV_SOURCE_ASM));
```
ncesario-lunarg, line 820: Running `spv_source` through spirv-val shows that it requires at least SPV 1.3. However, after changing this to `SPV_ENV_VULKAN_1_1` (and updating the test to request 1.1), this is now crashing here: https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/9c08a43f6770011edea0cb23a2f19cc9346172fb/layers/shader_validation.cpp#L1818

So now there are 3 questions:
1. Why did `-Wunused-variable` not complain about this?
2. Why is spirv-val within the validation layers not complaining about the SPV assembly needing at least SPV 1.3?
3. Why is this crashing as if it were invalid SPV?

Sadly I have answers to none of these at the moment...

sjfricke: > Running spv_source through spirv-val shows that it requires at least SPV 1.3

I ran

```
spirv-as test.spvasm --target-env vulkan1.0 -o test.spv
spirv-val --target-env vulkan1.0 test.spv
```

with no issue

-----

I do see it asserting now, but not in the same location... will look more into it

layers/generated/spirv_grammar_helper.cpp
```diff
@@ -625,9 +625,9 @@ bool OpcodeHasResult(uint32_t opcode) {
 // Will be 0 if there is no result.
 uint32_t OpcodeResultWord(uint32_t opcode) {
     uint32_t position = 0;
-    if (OpcodeHasResult(opcode)) {
```
sjfricke, line 628: @ncesario-lunarg So I realized when I added this I mixed this up

image

There will be a type THEN the result

So seems `OpDecorationGroup` is the only instruction I could find that has a Type but no Result and was causing the indexing into the shader module to go bad

tests/positive/shaderval.cpp
sjfricke: > Why did -Wunused-variable not complain about this?

So if I do the following

```c++
TEST_F(VkPositiveLayerTest, test) {
    const std::string x = R"()";
    int y = 2;
}
```

it will only complain about `y` so I guess the raw string will not trigger the warning

layers/shader_module.cpp
```diff
@@ -491,6 +494,10 @@ void SHADER_MODULE_STATE::PreprocessShaderBinary(const spv_target_env env) {
             // **THIS ONLY HAPPENS ON INITIALIZATION**. words should remain const for the lifetime
             // of the SHADER_MODULE_STATE instance.
             *const_cast<std::vector<uint32_t> *>(&words) = std::move(optimized_binary);
+            // Will need to update static data now the words have changed or else the def_index will not align
+            // It is really rare this will get here as Group Decorations have been deprecated and before this was added no one ever
+            // raised an issue for a bug that would crash the layers that was around for many releases
+            static_data_.Update(*this, true);
```
sjfricke, line 500: @ncesario-lunarg why it was segfaulting

tests/positive/shaderval.cpp
ncesario-lunarg: Hm, I'm guessing if it's a `const char*` instead of `std::string` it would complain. The compiler probably can't (easily) determine whether or not the constructor and destructor of `std::string` (or any other struct/class) has side effects.

ncesario-lunarg (commented):
Some questions here and there, but this LGTM. I didn't realize this case is only hit by deprecated spirv, but I'm glad that's the case!

layers/shader_module.cpp
```diff
@@ -419,7 +422,7 @@ SHADER_MODULE_STATE::SpirvStaticData::SpirvStaticData(const SHADER_MODULE_STATE
         }
     }
 
-    entry_points = SHADER_MODULE_STATE::ProcessEntryPoints(module_state);
+    entry_points = ProcessEntryPoints(module_state);
```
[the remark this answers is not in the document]
sjfricke: I think this is leftover from me trying to hack together different ways to change this all around, good catch

layers/shader_module.cpp
sjfricke: ok, ya I think the `const_cast` is better as it actually suppose to be const EXCEPT this rare case which is now has a comment explaining why it is the exeception

sjfricke: > without the need for an "Update" or "Reset" function

So the issue here is if we are in the `static_data_.has_group_decoration` section we need to "reset" or else when `variable_inst.push_back(insn);` is called it will be appended to the previous list

I think what might work (if you were already on this mental path sorry for repeating) is to have only the constructor and if we have to "reset" it we just delete and reconstruct it

lunarpapillo:
CI failure was likely due to a driver issue.  Rerunning.

ncesario-lunarg (approved):

PR #4459, opened 2022-08-23
Update descriptor/renderpass layout validation

MarkY-LunarG:


MarkY-LunarG:
Fixes issue #4312 .

layers/core_validation.cpp
```diff
@@ -13240,125 +13240,72 @@ bool CoreChecks::ValidateAttachmentReference(RenderPassCreateVersion rp_version,
...
+                                         function_name, error_type, string_VkImageLayout(stencil_layout));
                     }
                 }
-                break;
+            }
+            break;
+        case VK_IMAGE_LAYOUT_ATTACHMENT_OPTIMAL_KHR:
+        case VK_IMAGE_LAYOUT_READ_ONLY_OPTIMAL_KHR:
+            if (!enabled_features.core13.synchronization2) {
+                // TODO: Add tests!!!! Brainpain
```
jeremyg-lunarg, line 13299: Does this TODO still need to be fixed?

tests/vklayertests_descriptor_renderpass_framebuffer.cpp
```diff
@@ -499,112 +499,52 @@ TEST_F(VkLayerTest, RenderPassCreateAttachmentDescriptionInvalidFinalLayout) {
     if (separate_depth_stencil_layouts_features.separateDepthStencilLayouts) {
         attach_desc.initialLayout = VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_OPTIMAL_KHR;
         TestRenderPassCreate(m_errorMonitor, m_device->device(), &rpci, rp2Supported, "VUID-VkAttachmentDescription-format-03286",
```
jeremyg-lunarg, line 501: Would it be too confusing to do something like this?
```
void TestRenderPassCreate(ErrorMonitor *error_monitor, const VkDevice device,
                                            const VkRenderPassCreateInfo *create_info,
                                            bool rp2_supported, const char *rp1_vuid, 
                                             const char *rp2_vuid = nullptr) {

if (!rp2_vuid) {
    rp2_vuid = rp1_vuid;
}

TestRenderPassCreate(m_errorMonitor, m_device->device(), &rpci, rp2Supported,
                                     "VUID-VkAttachmentDescription-format-03286");
```

jeremyg-lunarg (commented):
LGTM except for the TODO BrainPain comment. :)  

It seems like this could cause big changes in the nightly vvltraces error counts and / or CTS.

layers/core_validation.cpp
MarkY-LunarG: Crap, missed that.  Thanks

tests/vklayertests_descriptor_renderpass_framebuffer.cpp
MarkY-LunarG: I actually like that.  Then you reduce duplication.  👍🏽

MarkY-LunarG: I had to adjust a few tests because doing this broke the existing behavior if rp_vuid was null (below this).  So we started getting more validation functions.  So I just set the "rp2_supported" flag in those limited cases to false to rememdy this.

layers/core_validation.cpp
MarkY-LunarG: Fixed.

tests/vklayertests_descriptor_renderpass_framebuffer.cpp
MarkY-LunarG: Suggested changes applied.

tests/vklayertests_descriptor_renderpass_framebuffer.cpp
```diff
@@ -1341,6 +1212,52 @@ TEST_F(VkLayerTest, RenderPassCreateAttachmentReferenceInvalidLayout) {
     }
 }
 
+TEST_F(VkLayerTest, RenderPassCreateAttachmentReferenceInvalidSync2Layout) {
+    TEST_DESCRIPTION("Attachment reference uses sync2 and ATTACHMENT_OPTIMAL_KHR or READ_ONLY_OPTIMAL_KHR layouts");
+
+    SetTargetApiVersion(VK_API_VERSION_1_3);
+    AddRequiredExtensions(VK_KHR_SYNCHRONIZATION_2_EXTENSION_NAME);
+
+    ASSERT_NO_FATAL_FAILURE(InitFramework(m_errorMonitor));
```
ncesario-lunarg, line 1221: If a 1.3 driver is really required, I'm guessing there should be a
```cpp
if (DeviceValidationVersion() < VK_API_VERSION_1_3) {
    GTEST_SKIP() ...;
}
```
after `InitFramework`?

ncesario-lunarg (commented):
One test question, but otherwise LGTM!

tests/vklayertests_descriptor_renderpass_framebuffer.cpp
MarkY-LunarG: Thanks!  I also applied the "AreRequiredExtensionsEnabled check and skip to.  But that will be pushed up after the current CI run has completed.

tests/layer_validation_tests.cpp
```diff
@@ -201,25 +201,25 @@ void TestRenderPassCreate(ErrorMonitor *error_monitor, const VkDevice device, co
     VkRenderPass render_pass = VK_NULL_HANDLE;
     VkResult err;
 
-    if (rp1_vuid) {
-        // Some tests mismatch attachment type with layout
-        error_monitor->SetUnexpectedError("VUID-VkSubpassDescription-None-04437");
+    // If the second VUID is not provided, set it equal to the first VUID.  In this way,
+    // we can check both vkCreateRenderPass and vkCreateRenderPass2 with the same VUID.
+    if (!rp2_vuid) {
```
jeremyg-lunarg, line 206: based on the rp1_vuid checks below it looks like there could be cases where `rp1_vuid == nullptr && rp2_vuid != nullptr`.  You could handle those by changing this check to be:` if (rp1_vuid && !rp2_vuid)`

jeremyg-lunarg: Not sure if this is needed based on your comments in tests/vklayertests_buffer_image_memory_sampler.cpp

jeremyg-lunarg (approved):
LGTM once internal CI passes

tests/vklayertests_descriptor_renderpass_framebuffer.cpp
MarkY-LunarG: I just saw sync2 was promoted into 1.3.  So I'll remove the additional extension  requirement and sheck.

tests/layer_validation_tests.cpp
MarkY-LunarG: Good point.  I see that is the case too, in "RenderPassCreateInvalidSubpassDependencies" at least.  I'll just put back the rp2_vuid check down below just in case.  I still like doing the check the new way.  I did move this if under the "if (rp1_vuid)" if, though so you're only setting the rp2_vuid value to non-NULL if rp1_vuid is not NULL now.

Tony-LunarG (approved):
LGTM

PR #4462, opened 2022-08-23
layers: Check if feature is enabled for VUID 06858 and 06859

tomowc01-ARM:
Add check if multisampledRenderToSingleSampled feature is enabled for VUID 06858 and 06859.

https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkRenderingInfo-imageView-06858
https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkRenderingInfo-imageView-06859

Currently when the VkPhysicalDeviceMultisampledRenderToSingleSampledFeaturesEXT is enabled for current device the checks are performed even when feature is not in use (disabled).


Tony-LunarG:
@tomowc01-ARM I've guarded the checks with VkMultisampledRenderToSingleSampledInfoEXT::multisampledRenderToSingleSampledEnable in #4470  Is that acceptable in lieu of this PR?

tomowc01-ARM:
@Tony-LunarG yes #4470 fixed the issue.
Many thanks for addressing it.

PR #4463, opened 2022-08-24
layers: Refactor StorageImageFormat tests

sjfricke:
So I am 99% sure no one (including CI) was running these tests as they did not have valid SPIR-V and I just ran on a device without `shaderStorageImageReadWithoutFormat` feature support and realized this

- The `UNASSIGNED-features-shaderStorageImageReadWithoutFormat` (and `write`) VU is redundant as `spirv-val` is checking it and there is no way to hit this VU in the VVL https://github.com/KhronosGroup/SPIRV-Tools/pull/2486/files
- With the `UNASSIGNED` VU gone, the only thing being checked in `ValidateShaderStorageImageFormats()` was the `OpVariable` so moved that function into the `ValidateVariables` to save from call it every instruction

### DIFF
--- a/layers/shader_validation.cpp
+++ b/layers/shader_validation.cpp
@@ -1064,86 +1064,52 @@ bool CoreChecks::ValidateShaderStageInputOutputLimits(const SHADER_MODULE_STATE 
     return skip;
 }
 
-bool CoreChecks::ValidateShaderStorageImageFormats(const SHADER_MODULE_STATE &module_state, const spirv_inst_iter &insn) const {
+bool CoreChecks::ValidateShaderStorageImageFormatsVariables(const SHADER_MODULE_STATE &module_state,
+                                                            const spirv_inst_iter &insn) const {
     bool skip = false;
+    // Go through all variables for images and check decorations
+    assert(insn.opcode() == spv::OpVariable);
+    // spirv-val validates this is an OpTypePointer
+    const spirv_inst_iter pointer_def = module_state.get_def(insn.word(1));
+    if (pointer_def.word(2) != spv::StorageClassUniformConstant) {
+        return skip;  // Vulkan Spec says storage image must be UniformConstant
+    }
+    spirv_inst_iter type_def = module_state.get_def(pointer_def.word(3));
 
-    switch (insn.opcode()) {
-        // Go through all ImageRead/Write instructions
-        case spv::OpImageSparseRead:
-        case spv::OpImageRead: {
-            // spirv-val validates this is an OpTypeImage
-            const uint32_t image = module_state.GetTypeId(insn.word(3));
-            const spirv_inst_iter image_def = module_state.get_def(image);
+    // Unpack an optional level of arraying
+    if (type_def.opcode() == spv::OpTypeArray || type_def.opcode() == spv::OpTypeRuntimeArray) {
+        type_def = module_state.get_def(type_def.word(2));
+    }
 
-            const uint32_t dim = image_def.word(3);
-            const uint32_t image_format = image_def.word(8);
-            // If the Image Dim operand is not SubpassData, the Image Format must not be Unknown, unless the
-            // StorageImageReadWithoutFormat Capability was declared.
-            if (dim != spv::DimSubpassData && image_format == spv::ImageFormatUnknown) {
-                skip |= RequireFeature(enabled_features.core.shaderStorageImageReadWithoutFormat,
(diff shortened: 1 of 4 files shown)


tests/vklayertests_pipeline_shader.cpp
```diff
@@ -2502,6 +2498,7 @@ TEST_F(VkLayerTest, MissingNonReadableDecorationStorageImageFormatRead) {
          %21 = OpConstant %19 1
          %22 = OpConstantComposite %20 %21 %21 %21
           %4 = OpFunction %2 None %3
+          %l = OpLabel
```
sjfricke, line 2501: without this `label` the `OpVariable` is not in a Block and not valid with `spirv-val`

tests/vklayertests_pipeline_shader.cpp
```diff
@@ -1856,13 +1855,13 @@ TEST_F(VkLayerTest, MissingStorageImageFormatRead) {
...
     if (DeviceExtensionSupported(gpu(), nullptr, VK_KHR_FORMAT_FEATURE_FLAGS_2_EXTENSION_NAME)) {
-        printf("%s %s supported, skipping.\n", kSkipPrefix, VK_KHR_FORMAT_FEATURE_FLAGS_2_EXTENSION_NAME);
-        return;
+        GTEST_SKIP() << "VK_KHR_format_feature_flags2 is supported";
     }
 
     // Make sure compute pipeline has a compute shader stage set
     const char *csSource = R"(
                OpCapability Shader
+               OpCapability StorageImageReadWithoutFormat
```
sjfricke, line 1864: Without this `OpCapability` it is invalid SPIR-V (`spirv-val` blows up) and with it here, the runtime needs to have either 
image

Both of which are being checked in the auto-generated SPIR-V capability check (`VUID 01091`)

ncesario-lunarg: Yup, this test is skipped _everywhere_ :/

layers/shader_validation.cpp
```diff
@@ -1064,86 +1064,52 @@ bool CoreChecks::ValidateShaderStageInputOutputLimits(const SHADER_MODULE_STATE
...
-                                       "shaderStorageImageReadWithoutFormat", kVUID_Features_shaderStorageImageReadWithoutFormat);
-            }
-            break;
-        }
-        case spv::OpImageWrite: {
-            // spirv-val validates this is an OpTypeImage
-            const uint32_t image = module_state.GetTypeId(insn.word(1));
-            const spirv_inst_iter image_def = module_state.get_def(image);
+    // Unpack an optional level of arraying
+    if (type_def.opcode() == spv::OpTypeArray || type_def.opcode() == spv::OpTypeRuntimeArray) {
```
ncesario-lunarg, line 1080: Are we guaranteed the "base type" of an array is not also an array? Or maybe it doesn't matter if it is in this case?

ncesario-lunarg (commented):

layers/shader_validation.cpp
sjfricke: This is a fair question

```cpp
    if (pointer_def.word(2) != spv::StorageClassUniformConstant) {
        return skip;  // Vulkan Spec says storage image must be UniformConstant
    }
```

and Vulkan also bans 2D arrays of some storage classes `VUID-StandaloneSpirv-UniformConstant-04655` as I remember trying to write a shader to break this assumption and ran into that

ncesario-lunarg: This seems to be valid:
```glsl
#version 450

layout(set = 0, binding = 0) uniform T { float v[1][1]; } t;

void main() {
	gl_Position = vec4(t.v[0][0]);
}
```
And produces
```asm
%_arr_float_uint_1_0 = OpTypeArray %float %uint_1
%_arr__arr_float_uint_1_0_uint_1 = OpTypeArray %_arr_float_uint_1_0 %uint_1
```
Though maybe we don't have to worry about that in this particular case?

sjfricke: this is a 

```
OpVariable %_ptr_Uniform_T Uniform
```

and not a `UniformConstant` storage class which is for image writes, so if you go

```glsl
layout(set = 0, binding = 0, r32ui) uniform uimage2D storageImage[2][2];
```

you will get an error

> WARNING: a.vert:2: '[][]' : Generating SPIR-V array-of-arrays, but Vulkan only supports single array level for this resource

ncesario-lunarg (approved):

PR #4466, opened 2022-08-25
Fix validation of VK_EXT_rasterization_order_attachment_access

ShabbyX:
VUs that enforced render pass being valid where not actually checking
that, incorrectly triggering when render pass was indeed valid.


ncesario-lunarg (approved):
LGTM, thanks!

We need to do a local CI reboot, but will merge after that.

PR #4470, opened 2022-08-25
Fix multisampled_render_to_single_sampled interpretation of "enabled"

Tony-LunarG:
Previous interpretation of "If multisampled-render-to-single-sampled is enabled" was that the extension was enabled. This changes the interpretation to mean that it is enabled for the subpass.

### DIFF
--- a/tests/vklayertests_descriptor_renderpass_framebuffer.cpp
+++ b/tests/vklayertests_descriptor_renderpass_framebuffer.cpp
@@ -12952,8 +12952,6 @@ TEST_F(VkLayerTest, MultisampledRenderToSingleSampled) {
 
     color_attachment.imageView = two_count_image_view.handle();
     color_attachment.resolveImageView = VK_NULL_HANDLE;
-    // Image samples can't be COUNT_1 or we get 06861, and has to be COUNT_1, or we get 06858, use COUNT_2 and expect 06858
-    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkRenderingInfo-imageView-06858");
     // If resolveMode is not VK_RESOLVE_MODE_NONE, resolveImageView must not be VK_NULL_HANDLE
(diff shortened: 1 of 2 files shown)


tests/vklayertests_descriptor_renderpass_framebuffer.cpp
```diff
@@ -12952,8 +12952,6 @@ TEST_F(VkLayerTest, MultisampledRenderToSingleSampled) {
 
     color_attachment.imageView = two_count_image_view.handle();
     color_attachment.resolveImageView = VK_NULL_HANDLE;
-    // Image samples can't be COUNT_1 or we get 06861, and has to be COUNT_1, or we get 06858, use COUNT_2 and expect 06858
-    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkRenderingInfo-imageView-06858");
```
Tony-LunarG, line 12956: Can no longer get 06858 unless msrtss enabled for subpass

paul-lunarg (approved):
LGTM. Though you talk about subpass, this check only changes VkRenderingInfo validation. For subpasses, it would be these VUIDs : https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkSubpassDescription2-pNext-06870
I assume there's no need to change validation for these ?

jeremyg-lunarg (approved):

Tony-LunarG:
> LGTM. Though you talk about subpass, this check only changes VkRenderingInfo validation. For subpasses, it would be these VUIDs : https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkSubpassDescription2-pNext-06870 I assume there's no need to change validation for these ?

Yes, those VUIDs are guarded by the presence of the info structure rather than saying "If multisampled-render-to-single-sampled is enabled", so I was more sure what to do with those, and they're correct.

ncesario-lunarg (approved):
LGTM!

PR #4471, opened 2022-08-25
docs: fix default value of report_flags

lunarpapillo:
layers/vk_layer_config.cpp defaults "report_flags" to just "error":

    value_map_["khronos_validation.report_flags"] = "error";

while "error,warn,perf" is listed as the default in both vk_layer_settings.txt:

    khronos_validation.report_flags = error,warn,perf

and layers/json/VkLayer_khronos_validation.json.in:

    "default": [
        "error",
        "warn",
        "perf"
    ]

This change makes the latter two files match the current code by changing the
default to be just "error".

This affects shipped SDK files as well as vkconfig behavior and SDK documentation.

### DIFF
--- a/layers/json/VkLayer_khronos_validation.json.in
+++ b/layers/json/VkLayer_khronos_validation.json.in
@@ -282,9 +282,7 @@
                         }
                     ],
                     "default": [
-                        "error",
-                        "warn",
-                        "perf"
+                        "error"
                     ]
                 },
                 {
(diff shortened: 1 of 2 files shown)


ncesario-lunarg (approved):
LGTM!

jeremyg-lunarg (approved):
LGTM, seems like we should wait for Christophe to weigh in though.

christophe-lunarg (approved):

Issue #4449, opened 2022-08-19
tests: SetTargetApiVersion() should return void


jeremyg-lunarg:
This function returns the minimum of the requested version and the version supported by the instance. Sometimes it is used to skip tests that require a version. But this is incorrect.  We also need the physical device version to meet the requirement.

So... this function should return void and DeviceValidationVersion() should be checked instead. 

For even more fun, can we make InitState() return a BAD_VERSION or MISSING_EXTENSION error to make it easier to skip?

arno-lunarg:
So I modified SetTargetApiVersion() like you said, and replaced every call where the return value was used with a subsequent call do DeviceValidationVersion() after the init step (except when that call was already done).

I am not sure about changing InitState(), because in at least one test, `VkPositiveLayerTest.MultiplaneImageTests`, the test just proceed to other steps if the API version is not 1.1, so should not be skipped?

jeremyg-lunarg:
> I am not sure about changing InitState(), because in at least one test, `VkPositiveLayerTest.MultiplaneImageTests`, the test just proceeds to other steps if the API version is not 1.1, so should not be skipped? :

Yes, trying to change InitState() is probably too tricky and would have consequences in many tests.

ncesario-lunarg:
I'd like to eventually change `InitFramework` to accept a list of extensions to enable and return either success or failure + an error message. The `Add(Required|Optional)Extensions` must occur _before_ `InitFramework` for it to work properly, so it is one more thing that _can_ be called in an incorrect order and therefore must be checked at code review time. So something like:
```cpp
struct TestRequirements { // <- or a better name
    std::vector<const char*> required_extensions; // <- will effectively just call AddRequiredExtensions on each element
    std::vector<const char*> optional_extensions; // <- will effectively just call AddOptionalExtensions on each element
    uint32_t requested_version = VK_API_VERSION_1_0;
    void* p_next = nullptr;
    // perhaps a lambda or some other means of checking features and properties?
    ...
    std::string error_msg; // <- non-empty on failure. Would include the RequiredExtensionsNotSupported() << " not supported." string
};
```
Then a test would look like
```cpp
TestRequirements reqs;
reqs.required_extensions = { VK_KHR_8BIT_STORAGE_EXTENSION_NAME, ... };
...
InitFramework(reqs);

// Ideally would want some way to streamline the following as well
if (!reqs.error_msg.empty()) {
[...]

PR #4473, opened 2022-08-26
Fix Issue #4449

arno-lunarg:
Fix for issue [#4449](https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4449)
Using DeviceValidationVersion() is the correct way to get the supported API version, since it returns the API version supported by the device. SetTargetApiVersion() used to return the API version supported by the Vulkan instance. It now returns void to prevent further harm.
Replaced every call where the return value was used with a subsequent call to DeviceValidationVersion() after the init step (except when that call was already done).

### DIFF
--- a/tests/positive/image_buffer.cpp
+++ b/tests/positive/image_buffer.cpp
@@ -2445,11 +2445,7 @@ TEST_F(VkPositiveLayerTest, PlaneAspectNone) {
 TEST_F(VkPositiveLayerTest, ImageCompressionControl) {
     TEST_DESCRIPTION("Checks for creating fixed rate compression image.");
 
-    uint32_t version = SetTargetApiVersion(VK_API_VERSION_1_2);
-    if (version < VK_API_VERSION_1_2) {
-        printf("%s At least Vulkan version 1.2 is required, skipping test.\n", kSkipPrefix);
-        return;
-    }
+    SetTargetApiVersion(VK_API_VERSION_1_2);
 
     AddRequiredExtensions(VK_EXT_IMAGE_COMPRESSION_CONTROL_EXTENSION_NAME);
     AddRequiredExtensions(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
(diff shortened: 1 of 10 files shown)


tests/vklayertests_dynamic_rendering.cpp
```diff
@@ -20,11 +20,7 @@
 TEST_F(VkLayerTest, DynamicRenderingCommandBufferInheritanceRenderingInfo) {
     TEST_DESCRIPTION("VkCommandBufferInheritanceRenderingInfoKHR Dynamic Rendering Tests.");
 
-    uint32_t version = SetTargetApiVersion(VK_API_VERSION_1_2);
-    if (version < VK_API_VERSION_1_2) {
-        printf("%s At least Vulkan version 1.2 is required, skipping test.\n", kSkipPrefix);
-        return;
-    }
+    SetTargetApiVersion(VK_API_VERSION_1_2);
```
sjfricke, line 23: test in this file missing the `if (DeviceValidationVersion() < VK_API_VERSION_1_X) {` ?

sjfricke: same for some of the vklayertests_others.cpp

arno-lunarg: It is done line 32!

tests/positive/image_buffer.cpp
```diff
@@ -2459,6 +2455,11 @@ TEST_F(VkPositiveLayerTest, ImageCompressionControl) {
 
     ASSERT_NO_FATAL_FAILURE(InitFrameworkAndRetrieveFeatures(features2));
 
+    if (DeviceValidationVersion() < VK_API_VERSION_1_2) {
+        printf("%s At least Vulkan version 1.2 is required, skipping test.\n", kSkipPrefix);
```
jeremyg-lunarg, line 2459: Since you're touching these printfs(), it would be nice to upgrade them to `GTEST_SKIP() << ...;`  so that we get better stats out of gtests about which tests run on each system.

jeremyg-lunarg (approved):
LGTM. It would be nice to upgrade the printfs(), but not completely necessary. And yeah, trying to change InitState() would be tricky and error prone.

tests/vklayertests_dynamic_rendering.cpp
sjfricke: I see, LGTM too! Always enjoy these test cleanups!

tests/positive/image_buffer.cpp
arno-lunarg: Will do!

arno-lunarg: Can I do this in a separate commit?

jeremyg-lunarg: sure!

jeremyg-lunarg (approved):

ncesario-lunarg (approved):
LGTM!

Tony-LunarG (commented):
LGTM

PR #4474, opened 2022-08-26
Make VkRenderFramework::physDevProps() return const ref

arno-lunarg:
Given how it is used, VkRenderFramework::physDevProps() should return a const reference to avoid unnecessary copy of a not so small structure. Besides, the copy not being deep, copying the `char*` may lead to memory issues in the future (and we are not so far from one at the moment).

This method is used twice:

https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/6086f7993dde6e6d8a9e0f8eef29584289bd216f/tests/vkrenderframework.cpp#L757-L759

https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/6086f7993dde6e6d8a9e0f8eef29584289bd216f/tests/layer_validation_tests.cpp#L1027-L1030

### DIFF
--- a/tests/vkrenderframework.cpp
+++ b/tests/vkrenderframework.cpp
@@ -318,7 +318,7 @@ VkPhysicalDevice VkRenderFramework::gpu() {
     return gpu_;
 }
 
-VkPhysicalDeviceProperties VkRenderFramework::physDevProps() {
+const VkPhysicalDeviceProperties &VkRenderFramework::physDevProps() {
     EXPECT_NE((VkPhysicalDevice)0, gpu_);  // Invalid to request physical device properties before gpu
     return physDevProps_;
 }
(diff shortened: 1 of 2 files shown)


paul-lunarg (approved):
Good first change 👍

jeremyg-lunarg (approved):
LGTM.  One of the uses looks problematic for another reason:
```
 } else { 
     return (!vk_gpu_table.find(platform)->second.compare(physDevProps().deviceName)); 
 } 
```
If find() fails we will have a bad time. But maybe that's a second PR?

arno-lunarg:
> LGTM. One of the uses looks problematic for another reason:
> 
> ```
>  } else { 
>      return (!vk_gpu_table.find(platform)->second.compare(physDevProps().deviceName)); 
>  } 
> ```
> 
> If find() fails we will have a bad time. But maybe that's a second PR?

Oh yes did not see that. Will fix it in an other PR.

ncesario-lunarg (commented):
LGTM, nice!

PR #4475, opened 2022-08-26
Fix VkRenderFramework::IsPlatform()

arno-lunarg:
Spotted by @jeremy-lunarg 
The return value of find can be invalid, thus must be checked before use.

### DIFF
--- a/tests/vkrenderframework.cpp
+++ b/tests/vkrenderframework.cpp
@@ -755,7 +755,11 @@ bool VkRenderFramework::IsPlatform(PlatformType platform) {
     if (VkRenderFramework::IgnoreDisableChecks()) {
         return false;
     } else {
-        return (!vk_gpu_table.find(platform)->second.compare(physDevProps().deviceName));
+        const auto search = vk_gpu_table.find(platform);
+        if (search != vk_gpu_table.end()) {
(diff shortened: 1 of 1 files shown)


tests/vkrenderframework.cpp
```diff
@@ -755,7 +755,11 @@ bool VkRenderFramework::IsPlatform(PlatformType platform) {
     if (VkRenderFramework::IgnoreDisableChecks()) {
         return false;
     } else {
-        return (!vk_gpu_table.find(platform)->second.compare(physDevProps().deviceName));
+        const auto search = vk_gpu_table.find(platform);
+        if (search != vk_gpu_table.end()) {
+            return 0 == search->second.compare(physDevProps().deviceName);
+        }
+        return false;
```
arno-lunarg, line 762: Is returning false ok? Or assert instead?

ncesario-lunarg: I think returning `false` is necessary due to release builds, but also makes sense here since `vk_gpu_table` is manually maintained by us, and it's probably best for a test to run on an "unknown gpu" rather than skip it. If the test fails we can update the gpu table accordingly.

ncesario-lunarg (approved):
LGTM!

Issue #4478, opened 2022-08-27
Missing essential information in Validation Error.


CarloWood:
I am currently running into the following validation error:

```
$4 = {sType = VK_STRUCTURE_TYPE_DEBUG_UTILS_MESSENGER_CALLBACK_DATA_EXT, pNext = 0x0, flags = 0, 
  pMessageIdName = 0x7fffec046a20 "VUID-VkGraphicsPipelineCreateInfo-layout-00756", messageIdNumber = 1165064310, 
  pMessage = 0x7fffec046c70 "Validation Error: [ VUID-VkGraphicsPipelineCreateInfo-layout-00756 ] Object 0: handle = 0x7fffe4113eb0, name = LogicalDevice, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x45717876 | Shader uses descriptor slot 0.1 (expected `VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER`) but not declared in pipeline layout The Vulkan spec states: layout must be consistent with all shaders specified in pStages (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkGraphicsPipelineCreateInfo-layout-00756)", queueLabelCount = 0, 
  pQueueLabels = 0x0, cmdBufLabelCount = 0, pCmdBufLabels = 0x0, objectCount = 1, pObjects = 0x7fffec088bc0}
```

Aka,

```Shader uses descriptor slot 0.1 (expected `VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER`) but not declared in pipeline layout The Vulkan spec states: layout must be consistent with all shaders specified in pStages```

Where `pObjects` contains just

```
1: {objectType:Device, objectHandle:7fffe4113eb0, pObjectName:"LogicalDevice"}
```

What is missing here is: which shader? Which pipeline layout? Even if only -say- hexadecimal values are available, they are still better than nothing.

**Comment by sjfricke** [comment]
agree this is lacking the useful information you want from the validation error.

So a while ago a nice new `LogObjectList` struct was added so we could go

```cpp
LogObjectList objlist(module_state.vk_shader_module());
objlist.add(pipeline_layout->layout());
skip |= LogError(objlist, ....
```

but not every error message has been updated yet, but just pushed a PR up to address the shader validation to print out better object handle information

CarloWood:
[...]

PR #4481, opened 2022-08-29
layers: Improve objects reported in shader val messages

sjfricke:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4478

Make use of `LogObjectList` and putting the actual `VkShaderModule` handle

Also removed `pipeline->pipeline()` as this is pre-pipeline validation so that handle is going to be NULL

### DIFF
--- a/layers/shader_validation.cpp
+++ b/layers/shader_validation.cpp
@@ -464,11 +464,12 @@ bool CoreChecks::ValidateBuiltinLimits(const SHADER_MODULE_STATE &module_state, 
                 uint32_t length = static_cast<uint32_t>(module_state.GetConstantValueById(type.word(3)));
                 // Handles both the input and output sampleMask
                 if (length > phys_dev_props.limits.maxSampleMaskWords) {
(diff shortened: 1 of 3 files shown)


layers/shader_validation.cpp
```diff
@@ -479,15 +480,16 @@ bool CoreChecks::ValidateBuiltinLimits(const SHADER_MODULE_STATE &module_state,
...
+                                         safe_VkPipelineShaderStageCreateInfo const *info) const {
     bool skip = false;
 
     const auto *spec = info->pSpecializationInfo;
 
     if (spec) {
         for (auto i = 0u; i < spec->mapEntryCount; i++) {
             if (spec->pMapEntries[i].offset >= spec->dataSize) {
-                skip |= LogError(device, "VUID-VkSpecializationInfo-offset-00773",
+                skip |= LogError(module_state.vk_shader_module(), "VUID-VkSpecializationInfo-offset-00773",
```
ncesario-lunarg, line 492: This may not be super useful when graphics pipeline libraries are in use, as this will be null, but not a whole lot we can do about that at the moment.

ncesario-lunarg (approved):
LGTM!

PR #4482, opened 2022-08-29
CMAKE_COMPILER_IS_GNUCC is deprecated

arno-lunarg:
CMAKE_COMPILER_IS_GNUCC usage is not recommended by official doc. Besides, it may be true for clang, potentially leading to errors. So use CMAKE_GNU_COMPILER_ID instead, a variable already available in CMake 3.10

[Official CMAKE_COMPILER_IS_GNUCC  doc](https://cmake.org/cmake/help/v3.10/variable/CMAKE_COMPILER_IS_GNUCC.html)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,7 +193,7 @@ if(MAKE_C_COMPILER_ID MATCHES "Clang")
                         -Wstring-conversion)
 
 endif()
-if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "Clang")
+if(CMAKE_GNU_COMPILER_ID OR CMAKE_C_COMPILER_ID MATCHES "Clang")
     add_compile_options(-Wall
                         -Wextra
                         -Wno-unused-parameter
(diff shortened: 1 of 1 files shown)


arno-lunarg:
the CMAKE_Clang_COMPILER_ID variable does not seem to be recognised, looking at this.

ncesario-lunarg:
I think these conditions are fairly old. If you're interested, it might be good to investigate why these conditions must still be so specific, as it would be better if we could just keep flags as consistent as possible "per platform" rather than "per compiler verison" :)

arno-lunarg:
> I think these conditions are fairly old. If you're interested, it might be good to investigate why these conditions must still be so specific, as it would be better if we could just keep flags as consistent as possible "per platform" rather than "per compiler verison" :)

I will take a look at, it would indeed be a welcome change!

arno-lunarg:
Is the cmake minimum version set in stone?

ncesario-lunarg:
Not exactly. The current minimum is set for purely historical reasons. If there's a good argument to raise the minimum, we should be able to.

PR #4486, opened 2022-08-30
CMAKE_COMPILER_IS_GNUCC is deprecated

arno-lunarg:
CMAKE_COMPILER_IS_GNUCC usage is not recommended by official doc. Besides, it may be true for clang, potentially leading to errors. So use CMAKE_GNU_COMPILER_ID instead, a variable already available in CMake 3.10

[Official CMAKE_COMPILER_IS_GNUCC  doc](https://cmake.org/cmake/help/v3.10/variable/CMAKE_COMPILER_IS_GNUCC.html)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -181,42 +181,42 @@ endif()
 
 # Platform-specific compiler switches
 option(BUILD_WERROR "Treat compiler warnings as errors" ON)
-if(${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang")
+if(${CMAKE_C_COMPILER_ID} STREQUAL "Clang")
     add_compile_options(-Wconversion
-
-                        # TODO These warnings also get turned on with -Wconversion in some versions of clang.
-                        #      Leave off until further investigation.
-                        -Wno-sign-conversion
-                        -Wno-shorten-64-to-32
-                        -Wno-implicit-int-conversion
-                        -Wno-enum-enum-conversion
-                        -Wstring-conversion)
+      # TODO These warnings also get turned on with -Wconversion in some versions of clang.
+      #      Leave off until further investigation.
+      -Wno-sign-conversion
+      -Wno-shorten-64-to-32
+    -Wno-implicit-int-conversion
+    -Wno-enum-enum-conversion
+    -Wstring-conversion)
 
 endif()
+
 if(${CMAKE_CXX_COMPILER_ID} MATCHES "(GNU|Clang)")
-    add_compile_options(-Wall
-                        -Wextra
-                        -Wno-unused-parameter
-                        -Wno-missing-field-initializers
-                        -fno-strict-aliasing
-                        -fno-builtin-memcmp
-                        -fvisibility=hidden)
+  add_compile_options(-Wall
+    -Wextra
+    -Wno-unused-parameter
+    -Wno-missing-field-initializers
+    -fno-strict-aliasing
+    -fno-builtin-memcmp
+    -fvisibility=hidden)
 
-    # Treat warnings as errors for versions of GCC and c++11-compliant Clang versions that are shipped on Ubuntu 18.04 or older.
-    if(BUILD_WERROR)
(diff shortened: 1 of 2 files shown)


sjfricke:
this added a new error

> /home/runner/work/Vulkan-ValidationLayers/Vulkan-ValidationLayers/layers/buffer_validation.cpp:6280:46: error: implicit conversion from 'unsigned int' to 'float' may lose precision [-Werror,-Wimplicit-int-float-conversion]
            if (image_view_min_lod->minLod > max_level) {
            
seems you will need to update `buffer_validation.cpp` to be 

```patch
-auto max_level = (pCreateInfo->subresourceRange.baseMipLevel + (pCreateInfo->subresourceRange.levelCount - 1));
+auto max_level = static_cast<float>(pCreateInfo->subresourceRange.baseMipLevel + (pCreateInfo->subresourceRange.levelCount - 1));
if (image_view_min_lod->minLod > max_level) {
```

arno-lunarg:
Yes I saw that when building locally... This one is easy to fix, but I have also another warning for an unused variable ,`AllVkFormatFeatureFlagBits`,in `layers/generated/parameter_validation.cpp`.
It is only needed on `fuchsia`, thus I would like to enclose it with `#ifdef VK_USE_PLATFORM_FUCHSIA`, but looking at the generation scripts I do not find how to do that : /

sjfricke:
Ok, so I see only `vkSetBufferCollectionImageConstraintsFUCHSIA` and `vkSetBufferCollectionBufferConstraintsFUCHSIA` use it

I think there are a few ways to solve this, but good to take in consideration is sometime in the future, another non-Fuchsia function might need it as well.... few possible options are

1. (**my favorite**) just wrap all the `AllVk*` flags in the generated file with a "don't mark as unused"
2. (**easiest**) Add `VkFormatFeatureFlags` to the "skip" list as we do for `VkGeometryInstanceFlagsKHR` and then manually check it later (or not at all and mark as a TODO as doubt this never hit that struct)
3. (**robust but complex**) Have a way to add a combine the `self.struct_feature_protect` and `self.flag_values_definitions` that could detect if the struct is only in a the `VK_USE_PLATFORM_FUCHSIA` and then do what you purpose and wrap `AllVkFormatFeatureFlagBits` as well

@ncesario-lunarg can weigh in on what he would be happy with

arno-lunarg:
I vote for 1) too!
Can I do this using code from https://www.fluentcpp.com/2019/08/30/how-to-disable-a-warning-in-cpp/
(excerpt at the end of the page)
I could create a header `platform_defines.h` for this, or put it in `vk_layer_utils.h`

ncesario-lunarg (commented):
After an offline discussion with Arno, I think we've determined that the warnings in the generated code do not occur under "normal" build conditions. Once CI is happy, I think this is good to go.

arno-lunarg:
Updated PR. I just would like to bring attention to one thing that puzzles me. The present version of the CMakeLists.txt has this `if` to enable/disable `-Werror` on gcc/clang:
```
if(BUILD_WERROR)
        if ((CMAKE_COMPILER_IS_GNUCXX AND NOT (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 7.3.0)) OR
           (("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang") AND (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 6.0.0)))
            add_compile_options(-Werror)
        endif()
endif()
```

Basically you cannot _enforce_ `-Werror`. So I changed it to this:
``` 
if(BUILD_WERROR OR
      (${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS_EQUAL 7.3.0) OR
      (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 6.0.0))
        add_compile_options(-Werror)
endif()
```

Any objection?

arno-lunarg:
See https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4492 for updated PR

PR #4488, opened 2022-08-30
Fix warning for loading deprecated extension

### LINKED ISSUE #4293: Best practices layer warns about deprecated extension based on instance version
**Describe the Issue**
When creating an instance with (in my case) 1.3.216 and then calling vkCreateDevice for a device that supports 1.1.217, I get the following warning from the best practices layer:
> CreateDevice(): Attempting to enable deprecated extension VK_KHR_timeline_semaphore, but this extension has been promoted to VK_VERSION_1_2.

This is nonsense, as determining the proper extensions for device creation is device specific and depends on the api version of the device, not the instance. The device does not know of `KHR_timeline_semaphore` being deprecated as it doesn't know about 1.2. Currently, the message is absolutely useless. Although, simply by checking by physical device's apiVersion, the message makes sense again and is actually something a developer should consider. (Also there's a `vk` prefix missing on the CreateDevice() at the start of the string.)

paul-lunarg:
When requesting an API version that deprecates an extension that is enabled on the device, a BP warning is always triggered, even if the device doesn't support the requested API version and hasn't deprecated the extension.
This fix compares the extension deprecation version to the device version instead of the instance API version.

Closes #4293

### DIFF
--- a/layers/best_practices_utils.cpp
+++ b/layers/best_practices_utils.cpp
@@ -265,7 +265,7 @@ bool BestPractices::PreCallValidateCreateDevice(VkPhysicalDevice physicalDevice,
     for (uint32_t i = 0; i < pCreateInfo->enabledExtensionCount; i++) {
         const char *extension_name = pCreateInfo->ppEnabledExtensionNames[i];
 
-        uint32_t extension_api_version = device_api_version;
+        uint32_t extension_api_version = std::min(api_version, device_api_version);
 
         if (white_list(extension_name, kInstanceExtensionNames)) {
             skip |= LogWarning(instance, kVUID_BestPractices_CreateDevice_ExtensionMismatch,
--- a/layers/best_practices_utils.cpp
+++ b/layers/best_practices_utils.cpp
@@ -265,13 +265,16 @@ bool BestPractices::PreCallValidateCreateDevice(VkPhysicalDevice physicalDevice,
     for (uint32_t i = 0; i < pCreateInfo->enabledExtensionCount; i++) {
         const char *extension_name = pCreateInfo->ppEnabledExtensionNames[i];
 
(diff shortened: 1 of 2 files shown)


layers/best_practices_utils.cpp
```diff
@@ -271,7 +271,7 @@ bool BestPractices::PreCallValidateCreateDevice(VkPhysicalDevice physicalDevice,
                                extension_name);
         }
 
-        skip |= ValidateDeprecatedExtensions("CreateDevice", extension_name, api_version,
+        skip |= ValidateDeprecatedExtensions("CreateDevice", extension_name, device_api_version,
```
ncesario-lunarg, line 274: Is this sufficient? I vaguely remember looking at this issue and it seemed like we needed to know whether or not `extension_name` is an instance or device extension, and then compare to `api_version` or `device_api_version` respectively. This would likely require adding some information in the generated code.

paul-lunarg: Well I thought that since it is in CreateDevice it wouldn't make sense for there to be instance extensions

paul-lunarg: Since this same function checks if the extension is instance and issues a warning (I would have thought it would be an error) I used that to decide to compare with `api_version` or `device_api_version`.

ncesario-lunarg: Yeah, this looks good. I was either over-thinking it or thinking about a different issue.

ncesario-lunarg (approved):

layers/best_practices_utils.cpp
paul-lunarg: There's actually a case where api version is lower than device version so I'm taking the minimum of the two to validate device extensions

jeremyg-lunarg (approved):

PR #4493, opened 2022-08-31
Undisable core on CI

paul-lunarg:
This PR reenables core validation on all tests by default.
The aim is to get validation errors on CI and fix them progressively so maybe we can officially reenable core by default on master.


juan-lunarg:
Closing due to age, @paul-lunarg no longer being active at lunarg, and a draft PR.

PR #4501, opened 2022-09-01
WIP Two object debug naming conditional code

jzulauf-lunarg:
For Reviewers:
 * Useful?
 * What would be needed to commit *off* or to commit *on*?
 

Added two debugging compile time options useful for debugging VVL.

Define DEFAULT_OBJECT_NAMES
	Gives each object a unqiue name <type_name>_<seq_no>.
	Sequence number is atomic and monotonic in creation order.
	Useful when unique handles is disabled and driver returns are
	inconsistent. BASE_NODE::SetName wraps debug extension for
	convenience.
Define BASE_NODE_DEBUG_NAME
	Stores the last name from SetName as the first member of
	BASE_NODE.

### DIFF
--- a/layers/base_node.cpp
+++ b/layers/base_node.cpp
@@ -28,6 +28,17 @@
 #include "base_node.h"
 #include "vk_layer_utils.h"
 
+void BASE_NODE::SetName(debug_report_data* report_data, const char* new_name) {
+    auto name_arg = lvl_init_struct<VkDebugMarkerObjectNameInfoEXT>();
+    name_arg.object = handle_.handle;
+    name_arg.objectType = get_debug_report_enum[handle_.type];
+    name_arg.pObjectName = new_name;
+    report_data->DebugReportSetMarkerObjectName(&name_arg);
+#ifdef BASE_NODE_DEBUG_NAME
+    debug_name_ = std::string(new_name);
+#endif
+}
+
 BASE_NODE::~BASE_NODE() { Destroy(); }
 
 void BASE_NODE::Destroy() {
--- a/layers/state_tracker.h
+++ b/layers/state_tracker.h
@@ -263,6 +263,35 @@ enum PushConstantByteState {
     PC_Byte_Not_Updated = 2,
 };
 
+#ifdef DEFAULT_OBJECT_NAMES
+template <typename Category, typename Counter = uint64_t>
+class CategoryCounter {
+  public:
+    Counter Count(const Category& category) {
+        auto& count = counts_[category];
+        return count.fetch_add(1);
+    }
+
+  private:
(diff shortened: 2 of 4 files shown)


jeremyg-lunarg (commented):
I'm in need of something that is *almost* this.  I'd like to have a per-object type id (CB1, CB2, CB3, CB4, IMG1, IMG2, etc...) for purposes other than user visible names.  Then you could store that ID in BASE_NODE and generate names on the fly when you needed.   This would be more memory efficient as well.  With an arbitrary BASE_NODE you should be able to generate a name with a static lookup table of base_node->Handle().type -> friendly type name + base_node->Id().

jzulauf-lunarg:
@jeremyg-lunarg  -- BASE_NODE_DEBUG_NAME stores the default name in the BASE_NODE... how does that differ from your requirement?

jeremyg-lunarg:
I want the integer ID to remain constant over the life of the object no matter what the current name is set to.

Also, I think BASE_NODE_DEBUG_NAME would need to be off by default, storing a unique string in every object will be a lotta mallocs.

layers/base_node.cpp
```diff
@@ -28,6 +28,17 @@
 #include "base_node.h"
 #include "vk_layer_utils.h"
 
+void BASE_NODE::SetName(debug_report_data* report_data, const char* new_name) {
+    auto name_arg = lvl_init_struct<VkDebugMarkerObjectNameInfoEXT>();
```
ncesario-lunarg, line 32: `lvl_init_struct` -> `LvlInitStruct` (for consistency; the former will be removed "at some point in the future")

layers/state_tracker.h
```diff
@@ -263,6 +263,35 @@ enum PushConstantByteState {
...
+
+  private:
+    layer_data::unordered_map<Category, std::atomic<Counter>> counts_;
+};
+class DebugNames {
+  public:
+    std::string GetName(const std::string& type_name) {
+        char buff[81];
+        snprintf(buff, 81, "_%" PRIu64, unique_.Count(type_name));
+        return type_name + std::string(buff);
```
ncesario-lunarg, line 283: Why not just `return type_name + std::to_string(unique_.Count(type_name))` and remove `buff` and the `snprintf` call?

ncesario-lunarg (commented):
Seems like this could be very useful, but I agree with @jeremyg-lunarg that this should probably be off by default, at least initially.

juan-lunarg:
@jzulauf-lunarg  could you rebase this PR?

It's using the old CI flow which is impacting up the github actions cache.

juan-lunarg:
Closing due to age / staleness / impacting CI caching

If still valuable please open new PR

PR #4504, opened 2022-09-01
tests: Fix layer/extension enumeration

mikes-lunarg:
adf933c0d3f1ff2cffaefb21093176c0bd0af40a broke extension enumeration by returning {} for VK_INCOMPLETE. Found because header 1.3.226 pushed the device extension count > 256.


mark-lunarg (approved):

ncesario-lunarg (commented):
LGTM, thanks @mikes-lunarg!

Issue #4295, opened 2022-07-11
vkCmdBindVertexBuffers2-bindingCount-arraylength is incorrectly generated when bindingCount == 0 & (pSizes and pStrides are nullptr)


DominikWitczakAMD:
**Describe the Issue**
When validation is enabled, calling vkCmdBindVertexBuffers2EXT(cmd buffer handle, 0, 0, nullptr, nullptr, nullptr, nullptr) generates the following vallidation error:

``
vkCmdBindVertexBuffers2EXT: parameter bindingCount must be greater than 0. The Vulkan spec states: If any of pSizes, or pStrides are not NULL, bindingCount must be greater than 0 (https://vulkan.lunarg.com/doc/view/1.3.216.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdBindVertexBuffers2-bindingCount-arraylength)
``

The thing is, the VU the error is referring to does not contain the "bindingCount must be greated than 0" bit. It only contains the latter part, ie "If any of pSizes, or pStrides are not NULL, bindingCount must be greater than 0"

The call has to be valid in order for apps to be able to issue draw calls using GFX pipelines, that do not take any input VAs *but* have been created with the "vertex input binding stride" dynamic state enabled , without generating validation errors.

**Environment:**
 - OS:  Win10
 - GPU: irrelevant
 - SDK or header version if building from repo: 1.3.216.0

PR #4505, opened 2022-09-02
Fix VUID-vkCmdBindVertexBuffers2-bindingCount-arraylength validation

arno-lunarg:
Fix for issue https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4295

The way validation is done currently is a bit complex and seems to miss the [VUID definition](https://vulkan.lunarg.com/doc/view/1.3.216.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdBindVertexBuffers2-bindingCount-arraylength):
>VUID-vkCmdBindVertexBuffers2-bindingCount-arraylength
> If any of pSizes, or pStrides are not NULL, bindingCount must be greater than 0

This PR proposes a simpler validation, that correctly handles the call mentioned in the issue, and the VUID in general. The call was:
```
vkCmdBindVertexBuffers2EXT(cmd_buffer_handle, 0, 0, nullptr, nullptr, nullptr, nullptr);
```

Some questions:
1) The same validation is done at two spots, should it be factorised into a function?
2) The formation used for the `if`s, the variable names, the use of snake_case, is it ok?

### DIFF
--- a/layers/generated/parameter_validation.cpp
+++ b/layers/generated/parameter_validation.cpp
@@ -10185,10 +10185,23 @@ bool StatelessValidation::PreCallValidateCmdBindVertexBuffers2(
     const VkDeviceSize*                         pSizes,
     const VkDeviceSize*                         pStrides) const {
     bool skip = false;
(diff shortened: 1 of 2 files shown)


arno-lunarg:
Oops missing modif in the generation scripts

PR #4506, opened 2022-09-02
Add check for VkImageSubresourceLayers.layerCount == VK_REMAINING_ARRAY_LAYERS

### LINKED ISSUE #4485: VUIDs that catch baseArrayLayer + layerCount > arrayLayers may not work if layerCount== VK_REMAINING_ARRAY_LAYERS
It was recently reported to me that an application used VK_REMAINING_ARRAY_LAYERS with [VkImageSubresourceLayers.layerCount](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkImageSubresourceLayers.html), but the VVL didn't report a VUID violation like:
[VUID-vkCmdCopyImage-srcSubresource-01698](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdCopyImage-srcSubresource-01698)
> The srcSubresource.baseArrayLayer + srcSubresource.layerCount of each element of pRegions must be less than or equal to the arrayLayers specified in [VkImageCreateInfo](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkImageCreateInfo) when srcImage was created

Since VK_REMAINING_ARRAY_LAYERS is defined as 0xFFFFFFFF it's possible that doing baseArrayLayer+layerCount will overflow and wrap. Eyeballing the VVL code, I notice that potential happening here:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/67b79bcb00a5f56cb6403b4c2546bbca08da9818/layers/buffer_validation.cpp#L2697

It is valid to use VK_REMAINING_ARRAY_LAYERS with [VkImageSubresourceRange.layerCount](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkImageSubresourceRange.html).

Since using VK_REMAINING_ARRAY_LAYERS with VkImageSubresourceLayers.layerCount is likely to be a common mistake, there should be special handling in the VVL code to detect this.

paul-lunarg:
This PR adds a check and a test for using VK_REMAINING_ARRAY_LAYERS with VkImageSubresourceLayers.layerCount, which is unsupported. 
This check will trigger in place of the currently established check for VUID-vkCmdCopyImage-srcSubresource-01698 (for example in the case of vkCmdCopyImage), since there is no VUID for this yet. 

Closes #4485

### DIFF
--- a/layers/buffer_validation.cpp
+++ b/layers/buffer_validation.cpp
@@ -2751,11 +2751,18 @@ bool CoreChecks::ValidateImageArrayLayerRange(const CMD_BUFFER_STATE *cb_node, c
     bool skip = false;
     if (base_layer >= img->createInfo.arrayLayers || layer_count > img->createInfo.arrayLayers ||
         (base_layer + layer_count) > img->createInfo.arrayLayers) {
-        skip |= LogError(cb_node->commandBuffer(), vuid,
-                         "In %s, pRegions[%u].%s.baseArrayLayer is %u and .layerCount is "
-                         "%u, but provided %s has %u array layers.",
-                         function, i, member, base_layer, layer_count, report_data->FormatHandle(img->image()).c_str(),
-                         img->createInfo.arrayLayers);
+        if (layer_count == VK_REMAINING_ARRAY_LAYERS) {
+            skip |= LogError(cb_node->commandBuffer(), vuid,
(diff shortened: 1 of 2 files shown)


jeremyg-lunarg (approved):
LGTM

Issue #4477, opened 2022-08-27
Validation error triggers when binding pipelines with different attachment count in two different render passes (w/ dynamic_rendering)


loryruta:
**Describe the Issue**
Hi,
I'm using the dynamic_rendering feature of Vulkan 1.3; I have two rendering blocks (`vkCmdBeginRendering`/`vkCmdEndRendering` or "render passes") where in each of them I'm binding and using a graphics pipeline. The graphics pipeline within the first block has 3 color attachments while the second pipeline has only 1 color attachment.

Even if I'm using the two pipelines in two different begin/end blocks, I'm still getting this validation error which states that I shouldn't be using graphics pipelines with different attachment count in the same "render pass". 

I've tried padding the attachments of the second render pass to 3 by adding null attachments but the validation error changes and states the attachment 0 format mismatches between the two pipelines and it shouldn't because it's the same render pass.

Should I use the exact same attachments also among different rendering blocks? Or is that a validation layer possible bug that likely doesn't detect properly rendering blocks?

**Valid Usage ID**
```
Validation Error: [ VUID-vkCmdBindPipeline-pipeline-06195 ] Object 0: handle = 0x3f36830000000078, name = debug_renderer_pipeline, type = VK_OBJECT_TYPE_PIPELINE; | MessageID = 0x6c16bfb4 | vkCmdBindPipeline(): Binding pipeline with VkPipelineRenderingCreateInfo::colorAttachmentCount 1, but previously bound pipeline VkPipelineRenderingCreateInfo::colorAttachmentCount was 3. The Vulkan spec states: If pipeline is a graphics pipeline, this command has been called inside a render pass instance started with vkCmdBeginRendering, and commands using the previously bound graphics pipeline have been recorded within the render pass instance, then the value of VkPipelineRenderingCreateInfo::colorAttachmentCount specified by this pipeline must match that set in the previous pipeline (https://vulkan.lunarg.com/doc/view/1.3.224.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdBindPipeline-pipeline-06195)
```

**Environment:**
 - OS: Windows 10
 - GPU: NVIDIA GeForce RTX 2060 SUPER, driver version: 516.94
 - SDK or header version if building from repo: 1.3.224.0
 - Options enabled (synchronization, best practices, etc.): VkConfig's "Standard Preset"

**Additional context**

[...]

PR #4509, opened 2022-09-02
layers: Fix check for dynamic rendering attachment format use

ncesario-lunarg:
The spec says to check for the current render pass instance,
but code was checking if there were any draw calls at all in the command
buffer.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -9551,8 +9551,8 @@ bool CoreChecks::PreCallValidateCmdBindPipeline(VkCommandBuffer commandBuffer, V
                                  msrtss_info->rasterizationSamples, pipeline_rasterization_samples);
             }
         }
-        if ((cb_state->commands_since_begin_rendering > 0) && cb_state->activeRenderPass &&
-            cb_state->activeRenderPass->UsesDynamicRendering() && cb_state->has_draw_cmd) {
+        if (cb_state->activeRenderPass && cb_state->activeRenderPass->UsesDynamicRendering() &&
+            cb_state->has_draw_cmd_in_current_render_pass) {
             const auto rendering_struct = LvlFindInChain<VkPipelineRenderingCreateInfo>(pipeline_state->PNext());
             const auto last_pipeline = cb_state->GetCurrentPipeline(VK_PIPELINE_BIND_POINT_GRAPHICS);
             const auto *last_rendering_struct =
--- a/tests/positive/dynamic_rendering.cpp
+++ b/tests/positive/dynamic_rendering.cpp
@@ -205,9 +205,10 @@ TEST_F(VkPositiveLayerTest, DynamicRenderingDrawMultiBind) {
     m_commandBuffer->EndRendering();
 
     m_commandBuffer->BeginRendering(begin_rendering_info);
-    vk::CmdBindPipeline(m_commandBuffer->handle(), VK_PIPELINE_BIND_POINT_GRAPHICS, pipe2.handle());
+    // NOTE: Setting dynamic state does not count as "using" the currently bound pipeline.
     vk::CmdSetViewport(m_commandBuffer->handle(), 0, 1, &viewport);
     vk::CmdSetScissor(m_commandBuffer->handle(), 0, 1, &scissor);
+    vk::CmdBindPipeline(m_commandBuffer->handle(), VK_PIPELINE_BIND_POINT_GRAPHICS, pipe2.handle());
     vk::CmdDraw(m_commandBuffer->handle(), 3, 1, 0, 0);
     m_commandBuffer->EndRendering();
     m_commandBuffer->end();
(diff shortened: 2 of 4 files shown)


layers/core_validation.cpp
```diff
@@ -9480,8 +9480,8 @@ bool CoreChecks::PreCallValidateCmdBindPipeline(VkCommandBuffer commandBuffer, V
                                  msrtss_info->rasterizationSamples, pipeline_rasterization_samples);
             }
         }
-        if ((cb_state->commands_since_begin_rendering > 0) && cb_state->activeRenderPass &&
-            cb_state->activeRenderPass->UsesDynamicRendering() && cb_state->has_draw_cmd) {
+        if (cb_state->activeRenderPass && cb_state->activeRenderPass->UsesDynamicRendering() && cb_state->has_draw_cmd &&
```
jeremyg-lunarg, line 9483: seems like just checking has_draw_cmd_in_current_render_pass should be enough. Doesn't that imply that has_draw_cmd is also true?

tests/positive/dynamic_rendering.cpp
```diff
@@ -205,9 +205,9 @@ TEST_F(VkPositiveLayerTest, DynamicRenderingDrawMultiBind) {
     m_commandBuffer->EndRendering();
 
     m_commandBuffer->BeginRendering(begin_rendering_info);
-    vk::CmdBindPipeline(m_commandBuffer->handle(), VK_PIPELINE_BIND_POINT_GRAPHICS, pipe2.handle());
     vk::CmdSetViewport(m_commandBuffer->handle(), 0, 1, &viewport);
     vk::CmdSetScissor(m_commandBuffer->handle(), 0, 1, &scissor);
+    vk::CmdBindPipeline(m_commandBuffer->handle(), VK_PIPELINE_BIND_POINT_GRAPHICS, pipe2.handle());
```
jeremyg-lunarg, line 210: might be good to add some comments here about setting dynamic state not counting as 'using' (as discussed in the issue).  It wasn't very clear to me.

HansKristian-Work:
This should work, yes.

tests/positive/dynamic_rendering.cpp
paul-lunarg: Looking at [vkCmdSetViewport](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCmdSetViewport.html), "This command sets the viewport transformation parameters state for subsequent drawing commands when the graphics pipeline is created with VK_DYNAMIC_STATE_VIEWPORT" and there is no VUID stating it's illegal to call vkCmdSetViewport after binding a pipeline that doesn't have a dynamic viewport. All this considered, it makes sense that setting dynamic viewport doesn't actually "use" the viewport, just sets the state that *may* be used later...

paul-lunarg: This also fixes arno's issue #4477

paul-lunarg (approved):
LGTM

tests/positive/dynamic_rendering.cpp
arno-lunarg: > ... when THE graphics pipeline ...
Maybe not related to this PR, just need a clarification: the wording here makes me think that this command will only have effect on the currently bound pipeline? Calling `vkCmdSetViewport` before binding a pipeline created with VK_DYNAMIC_STATE_VIEWPORT won't change its viewport?
If this is true, my opinion is that then `vkCmdSetViewport` could be considered using the currently bound pipeline

paul-lunarg: vkCmdSetViewport will affect "subsequent drawing commands" and their bound pipelines. Since there are no VUIDs validating the use of vkCmdSetViewport with a pipeline that doesn't support it, it makes sense that you can call it before or after binding the pipeline you want to use it with. The VUIDs checking for dynamic state are at draw time, such as VUID-vkCmdDraw-commandBuffer-02701 and VUID-vkCmdDraw-None-02859

paul-lunarg: >Calling vkCmdSetViewport before binding a pipeline created with VK_DYNAMIC_STATE_VIEWPORT won't change its viewport?
It actually will, according to VUID-vkCmdDraw-commandBuffer-02701 :
> If the VkPipeline object bound to the pipeline bind point used by this command requires any dynamic state, that state must have been set or inherited for commandBuffer, and done so after any previously bound pipeline with the corresponding state not specified as dynamic

arno-lunarg: That was the clarification I was looking for, merci Paul : )

Tony-LunarG (approved):
LGTM with same question as Jeremy

tests/positive/dynamic_rendering.cpp
ncesario-lunarg: Yes, @paul-lunarg is correct. There was a related issue filed recently that I think did something very similar and triggered a VUID _incorrectly_; I can't recall which issue that was at the moment, but we should probably comment on it with this clarification.

ncesario-lunarg: > might be good to add some comments here about setting dynamic state not counting as 'using' (as discussed in the issue). It wasn't very clear to me.

Yep, will do.

PR #4511, opened 2022-09-02
Add support for GPU-AV support when using graphics pipeline libraries

ncesario-lunarg:


ncesario-lunarg:
@Tony-LunarG can you please take a look at this? I still have some TODOs that I need to possibly address, which is why this is a draft, but this _seems_ to be working (at least for the one test case I added so far).

tests/vklayertests_gpu.cpp
```diff
@@ -3039,4 +3039,196 @@ TEST_F(VkGpuAssistedLayerTest, DispatchIndirectWorkgroupSize) {
     m_commandBuffer->QueueCommandBuffer();
     ASSERT_VK_SUCCESS(vk::QueueWaitIdle(m_device->m_queue));
     m_errorMonitor->VerifyFound();
-}
\ No newline at end of file
+}
+
```
Tony-LunarG, line 3043: Will there be a debug printf test too?

Tony-LunarG (commented):
This all looks really good to me.  Is the pipeline layout question all that's left?  And maybe a debug printf test?

tests/vklayertests_gpu.cpp
ncesario-lunarg: Yep, forgot to add it.

ncesario-lunarg:
> Is the pipeline layout question all that's left? And maybe a debug printf test?

Basically yes. I think there's at least one `const_cast` that is sketchy, but I didn't see a straightforward way around it. The pipeline layout is probably the biggest question, but I think I can write another test to figure out whether or not this approach will work.

layers/debug_printf.cpp
```diff
@@ -675,10 +669,17 @@ void DebugPrintf::AllocateDebugPrintfResources(const VkCommandBuffer cmd_buffer,
...
     if (pipeline_state) {
-        const auto &pipeline_layout = pipeline_state->PipelineLayoutState();
+        const auto pipeline_layout = pipeline_state->PipelineLayoutState();
+        // If GPL is used, it's possible the pipeline layout used at pipeline creation time is null. If CmdBindDescriptorSets has
+        // not been called yet (i.e., state.pipeline_null), then fall back to the layout associated with pre-raster state.
+        // PipelineLayoutState should be used for the purposes of determining the number of sets in the layout, but this layout
+        // may be a "pseudo layout" used to represent the union of pre-raster and fragment shader layouts, and therefore have a
+        // null handle.
+        const auto pipeline_layout_handle =
+            (last_bound.pipeline_layout) ? last_bound.pipeline_layout : pipeline_state->PreRasterPipelineLayoutState()->layout();
```
ncesario-lunarg, line 679: One question here (that also applies to the same change in `gpu_validation.cpp`), is the case where independent sets are used *and* `CmdBindDescriptorSets` is never called. In that case, though, the descriptor sets that are not bound cannot be used and therefore cannot trigger any errors, no? The only place I saw `CmdBindDescriptorSets` _not_ called was in a test utilizing device addresses + push constants. I'm assuming that is the usual use case when not calling `CmdBindDescriptorSets`, in which case the descriptor set layout doesn't matter (if I understand correctly).

layers/gpu_utils.cpp
```diff
@@ -824,41 +818,84 @@ void GpuAssistedBase::PreCallRecordPipelineCreations(uint32_t count, const Creat
...
+                        // If the shader module's handle is non-null, then it was defined with CreateShaderModule and covered by the
+                        // case above. Otherwise, it is being defined during CGPL time
+                        if (cgpl_state.shader_states.size() <= pipeline) {
+                            cgpl_state.shader_states.resize(pipeline + 1);
+                        }
+                        auto &csm_state = cgpl_state.shader_states[pipeline][stage.stage_flag];
+                        const auto pass =
+                            InstrumentShader(module_state->words_, csm_state.instrumented_pgm, &csm_state.unique_shader_id);
+                        if (pass) {
+                            const_cast<SHADER_MODULE_STATE *>(module_state.get())->gpu_validation_shader_id =
```
ncesario-lunarg, line 870: These `const_cast`s are related to assigning the "unique shader ID" after instrumentation; I'd like to do this a bit differently to avoid this, but don't have a good solution yet.

layers/debug_printf.cpp
jeremyg-lunarg: 🤯 Which test does this?

tests/layer_validation_tests.cpp
```diff
@@ -1605,7 +1605,7 @@ void CreatePipelineHelper::InitRasterizationInfo() {
     rs_state_ci_.rasterizerDiscardEnable = VK_FALSE;
     rs_state_ci_.polygonMode = VK_POLYGON_MODE_FILL;
     rs_state_ci_.cullMode = VK_CULL_MODE_BACK_BIT;
-    rs_state_ci_.frontFace = VK_FRONT_FACE_COUNTER_CLOCKWISE;
+    rs_state_ci_.frontFace = VK_FRONT_FACE_CLOCKWISE;
```
jeremyg-lunarg, line 1608: why did this need to change?

layers/generated/vk_object_types.h
```diff
@@ -990,5 +990,6 @@ struct VulkanTypedHandle {
     VulkanTypedHandle() :
         handle(CastToUint64(VK_NULL_HANDLE)),
         type(kVulkanObjectTypeUnknown) {}
+    operator bool() const { return handle != 0; }
```
jeremyg-lunarg, line 993: This is handy and probably can be used in lots of places.

layers/gpu_utils.cpp
jeremyg-lunarg: you could always make the module member be non-const... or would that give too much potential for messing things up.  You could also use std::const_pointer_cast<> which would be slightly more concise.

layers/gpu_utils.cpp
```diff
@@ -824,41 +818,84 @@ void GpuAssistedBase::PreCallRecordPipelineCreations(uint32_t count, const Creat
...
+
+                            // Now we need to find the corresponding VkShaderModuleCreateInfo and update its shader code
+                            auto &stage_ci = GetShaderStageCI<SafeCreateInfo, safe_VkPipelineShaderStageCreateInfo>(
+                                new_pipeline_ci, stage.stage_flag);
+                            // We're modifying the copied, safe create info, which is ok to be non-const
+                            auto sm_ci =
+                                const_cast<VkShaderModuleCreateInfo *>(LvlFindInChain<VkShaderModuleCreateInfo>(stage_ci.pNext));
+                            // module_state->Handle() == VK_NULL_HANDLE should imply sm_ci != nullptr, but checking here anyway
+                            if (sm_ci) {
+                                // "safe" structs manage "raw" pointers, so we need to also manage that here when copying the
```
jeremyg-lunarg, line 881: since you have safe_VkPipelineShaderStageCreateInfo, doesn't that mean all of its pNext was set up by SafePnextCopy()? And if that's the case, sm_ci is really a safe_ struct.  So you could add a method on that class to update pCode and hide this gunk?

layers/debug_printf.cpp
ncesario-lunarg: I believe it was https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/d5bc5ca269e548716ed19237aac6618bd20566dc/tests/vklayertests_gpu.cpp#L1143 that made me aware of this :)

layers/gpu_utils.cpp
ncesario-lunarg: Yeah, I was trying to avoid making the module member non-cast, as everything _should_ be const _unless_ GPUAV is involved. That might be the better way to go, though, and just clarify this through comments.

I could have sworn I had used `std::const_pointer_cast`, but now I don't see it. It might be in my "union layouts" version.

layers/gpu_utils.cpp
ncesario-lunarg: I think so! I'll give that a try as that would be much better.

tests/layer_validation_tests.cpp
ncesario-lunarg: In short: fragments were not getting generated with back-face culling.

TL;DR: GPUAV generally (only?) uses `VkPipelineObj` for testing, and I only used `CreatePipelineHelper` for all of the graphics pipeline library tests. `VK_FRONT_FACE_CLOCKWISE` is used in `VkPipelineObj` and works for the tests that need to produce fragments in that code. Specifically vertex shader code like https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/d5bc5ca269e548716ed19237aac6618bd20566dc/tests/vklayertests_gpu.cpp#L255-L258

This was counter-intuitive to me since the order in which those vertices are specified looks counter-clockwise, but I didn't investigate too much as to why this is the case.

jeremyg-lunarg (commented):
This LGTM, but I'd prefer @Tony-LunarG  give the big thumbs up / thumbs down.

Tony-LunarG (approved):
Nice work - LGTM.  Actually fixes a VMA memory leak I was chasing running CTS.

tests/vktestbinding.h
```diff
@@ -1187,6 +1190,55 @@ inline VkCommandBufferAllocateInfo CommandBuffer::create_info(VkCommandPool cons
...
+
+    GraphicsPipelineFromLibraries(const Device &dev, layer_data::span<VkPipeline> libs, VkGraphicsPipelineCreateInfo *ci = nullptr)
+        : libs(libs) {
+        link_info = LvlInitStruct<VkPipelineLibraryCreateInfoKHR>();
+        link_info.libraryCount = libs.size();
+        link_info.pLibraries = libs.data();
+
+        if (ci) {
+            // This is an ugly cast, but only used in testing for convenience of extending pNext chain
+            VkBaseOutStructure **prev =
```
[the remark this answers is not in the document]
ncesario-lunarg: I wasn't sure if it was 100% safe to pass down `safe_*` structs to the driver, so didn't want to use them here, but that's probably fine.

That said, I'm trying to remember why I wanted to attach the `link_info` structure at the _end_ of the pNext chain, when I'm pretty sure just I could have just done `link_info->pNext = ci->pNext; ci->pNext = &link_info;` :thinking:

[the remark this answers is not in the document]
jeremyg-lunarg: yeah that seems like it'd work too

layers/gpu_utils.cpp
ncesario-lunarg: Looking at this a bit more, I think the "right" thing to do here is move as much of the GPUAV-specific state (declared near the beginning of state_tracker.h) into derived (`gpu_utils_state::`) state. There will probably still be some `const_cast`s involved, but the non-const items will only be in the GPUAV derived state. I've started this, but it's looking like it'll be a larger change, so probably going to save it for a follow up PR.

layers/gpu_utils.h
```diff
@@ -76,6 +76,15 @@ class CommandBuffer : public CMD_BUFFER_STATE {
...
     virtual void Process(VkQueue queue) = 0;
 };
+
+class ShaderModuleState : public SHADER_MODULE_STATE {
+  public:
+    ShaderModuleState() : SHADER_MODULE_STATE() {}
+    ShaderModuleState(const VkShaderModuleCreateInfo &create_info, VkShaderModule shader_module, spv_target_env env)
+        : SHADER_MODULE_STATE(create_info, shader_module, env) {}
+
+    uint32_t gpu_validation_shader_id{std::numeric_limits<uint32_t>::max()};
```
ncesario-lunarg, line 86: This avoids a couple `const_cast`s for now.

After trying to move the rest of the state in here, it looks like I need to either use something like John's `TlsGuard` or move all of the instrumentation to CGPL time, and I'll be doing the latter in a subsequent change anyway.

ncesario-lunarg:
Doing one more CTS run after the CTS crash fix went in, and then I'm going to merge if there are no objections.

PR #4514, opened 2022-09-03
build: Revert addition of aligned_alloc()

brad0:
Vulkan-ValidationLayers builds as C++11. aligned_alloc() was only added
in C++17. C headers are not guarenteed to expose the function until
C++17 mode, as noted on OpenBSD, but other implementations such as
glibc do the same.

### DIFF
--- a/layers/vk_mem_alloc.h
+++ b/layers/vk_mem_alloc.h
@@ -2688,7 +2688,9 @@ remove them if not needed.
    #define VMA_NULL   nullptr
 #endif
 
-#if defined(__ANDROID_API__) && (__ANDROID_API__ < 16)
+#if defined(_WIN32)
+static void* vma_aligned_alloc(size_t alignment, size_t size) { return _aligned_malloc(size, alignment); }
+#elif defined(__ANDROID_API__) && (__ANDROID_API__ < 16)
 #include <cstdlib>
 static void* vma_aligned_alloc(size_t alignment, size_t size)
 {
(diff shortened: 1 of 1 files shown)


jeremyg-lunarg (changes requested):
vk_mem_alloc.h is from https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/

It looks like they have a history of fixing up [problems with aligned_alloc on various OSes](https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/search?q=aligned_alloc&type=issues). Can you try negotiating a fix with them? This PR undoes a bunch of their fixes for MacOS and other platforms we support.

jeremyg-lunarg:
@brad0 we have recently switched to using c++17 by default and not supporting older standards. Does that solve your problem?

brad0:
> @brad0 we have recently switched to using c++17 by default and not supporting older standards. Does that solve your problem?

I'll have to double check, but it should. But also having VulkanMemoryAllocator itself upstream building as C++17 instead of C++14.

jeremyg-lunarg:
> > @brad0 we have recently switched to using c++17 by default and not supporting older standards. Does that solve your problem?
> 
> I'll have to double check, but it should. But also having VulkanMemoryAllocator itself upstream building as C++17 instead of C++14.

Hmm, isn't it header only? But, if you need us to update to a newer version of it, we can.

brad0:
> > > @brad0 we have recently switched to using c++17 by default and not supporting older standards. Does that solve your problem?
> > 
> > 
> > I'll have to double check, but it should. But also having VulkanMemoryAllocator itself upstream building as C++17 instead of C++14.
> 
> Hmm, isn't it header only? But, if you need us to update to a newer version of it, we can.

I thought maybe the header was used by the bit of C++ code there but looking around it appears it does not. So you're right. But it would be good to move the rest of the codebase over either way. I noticed there is a PR for SPIRV-Tools, what about Vulkan-Loader?

jeremyg-lunarg:
> I noticed there is a PR for SPIRV-Tools, what about Vulkan-Loader?

[Vulkan-Loader](https://github.com/KhronosGroup/Vulkan-Loader is C99 and only uses C++ for tests,  I don't think it is very urgent to upgrade it. We've upgraded several of the other repositories that we maintain already. If there are any projects  where using an older C++ standard is causing problems, please file an issue against that repository and we'll give it priority.

PR #4518, opened 2022-09-05
Fix vkCmdBindVertexBuffers2-bindingCount-arraylength

arno-lunarg:
Fix for issue https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4295

`VUID-vkCmdBindVertexBuffers2-bindingCount-arraylength` is not validated correctly. The problem is in the generated code for the check.
Note that the VUID ends with `-arraylength`. Looking at `parameter_validation_generator.py` all the VUIDs ending with this suffix have some validation code generated to check the value of count, the length of the/an array in the associated command. This makes perfect sense, since they all have the same form: count must be greater that 0. No additional condition.

Except `VUID-vkCmdBindVertexBuffers2-bindingCount-arraylength`. The spec states additional conditions:
>If any of pSizes, or pStrides are not NULL, bindingCount must be greater than 0. 

This VUID seems to be the only exception to the previous rule. And it makes calling the generated code invalid in that case.

So this PR provides manual checking for this VUID. 

It also updates the test `ValidateExtendedDynamicStateEnabled` to also check the null call case.


arno-lunarg:
Adding an exception for this VUID in the generation script is way simpler than renaming the VUID, the root issue.

paul-lunarg (approved):
LGTM

mark-lunarg (approved):

jeremyg-lunarg (approved):
LGTM, thanks for the detailed explanation in the PR

Tony-LunarG (approved):
LGTM

PR #4519, opened 2022-09-05
Add check for VUID-vkCmdDraw-None-02686

### LINKED ISSUE #4494: VUID-vkCmdDraw-None-02686 validation missing
The layers do not currently validate this VUID at all.

When using input attachments in a render pass, if the input attachment binding is used in the shader, then a validation error will result if nothing is bound to that binding at draw time. This is the same as any other binding that is used by the shader but not bound at draw time.

But according to the VUID, an input attachment must be bound if it's used by the subpass, regardless of whether it's used in the shader. Thus, there should be a validation error if no image is bound to an input attachment via a descriptor set, even if the shader never uses it.

**Valid Usage IDs requested**
VUID-vkCmdDraw-None-02686
> Every input attachment used by the current subpass **must** be bound to the pipeline via a descriptor set

paul-lunarg:
Add check for not binding all input attachments that are used in a subpass.
Also adds a check for when an input attachment descriptor image view isn't used as input attachment in current subpass/framebuffer. Is this already a VUID ? I can't find it anywhere, and seems like it should be.

Closes #4494

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -1590,7 +1590,7 @@ bool CoreChecks::ValidateCmdBufDrawState(const CMD_BUFFER_STATE *cb_node, CMD_TY
 
             // Check if all input attachments are bound
             // VUID-vkCmdDraw-None-02686
-            std::set<VkImageView> subpass_input_views;
+            std::unordered_map<VkImageView, uint32_t> subpass_input_views;
             const auto &subpass = cb_node->activeRenderPass->createInfo.pSubpasses[cb_node->activeSubpass];
             for (const auto &stage : pipe->stage_state) {
                 if (stage.stage_flag == VK_SHADER_STAGE_FRAGMENT_BIT) {
@@ -1600,7 +1600,7 @@ bool CoreChecks::ValidateCmdBufDrawState(const CMD_BUFFER_STATE *cb_node, CMD_TY
                         if (index != VK_ATTACHMENT_UNUSED) {
                             VkImageView input_attachment_view =
                                 cb_node->activeFramebuffer->attachments_view_state[index]->Handle().Cast<VkImageView>();
-                            subpass_input_views.insert(input_attachment_view);
+                            subpass_input_views.insert(std::make_pair(input_attachment_view, i));
                         }
                     }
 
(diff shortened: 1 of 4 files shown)


jeremyg-lunarg:
FYI it looks like there was checking for 02686 but it was removed by https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/2244

layers/core_validation.cpp
```diff
@@ -1587,6 +1587,59 @@ bool CoreChecks::ValidateCmdBufDrawState(const CMD_BUFFER_STATE *cb_node, CMD_TY
...
+                        }
+                    }
+
+                    break;
+                }
+            }
+
+            // Report error for every unbound input attachment
+            for (auto view : subpass_input_views) {
+                result |= LogError(view, vuid.subpass_input,
```
jeremyg-lunarg, line 1639: Please remove this comment from core_validation.h (line 62)

```
    // subpass input doesn't validate anything because those checks were done in ValidateCreateGraphicsPipelines
    const char* subpass_input = kVUIDUndefined;

```

jeremyg-lunarg (approved):
LGTM

ncesario-lunarg (approved):
LGTM!

layers/core_validation.cpp
```diff
@@ -1587,6 +1587,61 @@ bool CoreChecks::ValidateCmdBufDrawState(const CMD_BUFFER_STATE *cb_node, CMD_TY
...
+                                    }
+                                    // Input attachment is ImageDescriptor
+                                    const auto image_descriptor = static_cast<const cvdescriptorset::ImageDescriptor *>(descriptor);
+                                    VkImageView image_view = image_descriptor->GetImageView();
+
+                                    // Image view isn't used by subpass as an input attachment with this framebuffer, is this a VUID
+                                    // ?
+                                    if (subpass_input_views.count(image_view) == 0) {
+                                        result |=
+                                            LogError(image_view, "UNASSIGNED-input-attachment-descriptor-not-in-subpass",
```
ncesario-lunarg, line 1624: Be sure to add this in `layers/core_validation_error_enums.h` and use the constant here.

ncesario-lunarg (changes requested):

PR #4525, opened 2022-09-07
tests: fix how global list of extensions/layers is obtained

arno-lunarg:
Saw that while working on another PR.
The way all extensions/layers are queried is not really what you are supposed to do. And `GetGlobalExtensions` was not using its `pLayerName` parameter.

### DIFF
--- a/tests/vktestbinding.cpp
+++ b/tests/vktestbinding.cpp
@@ -144,7 +144,7 @@ std::vector<VkExtensionProperties> GetGlobalExtensions(const char *pLayerName) {
     uint32_t extension_count = 32;
     std::vector<VkExtensionProperties> extensions(extension_count);
     for (;;) {
-        err = vk::EnumerateInstanceExtensionProperties(nullptr, &extension_count, extensions.data());
+        err = vk::EnumerateInstanceExtensionProperties(pLayerName, &extension_count, extensions.data());
         if (err == VK_SUCCESS) {
             extensions.resize(extension_count);
             return extensions;


arno-lunarg:
I miss error check, will add it

paul-lunarg:
Here is the commit that implemented the old method : https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/adf933c0d3f1ff2cffaefb21093176c0bd0af40a
@charles-lunarg What do you think ? 
Is it possible that EnumerateInstanceLayerProperties will return VK_INCOMPLETE even though the layer_count is the same as returned by the previous call to EnumerateInstanceLayerProperties ?
Also it seems that multiplying by 2 is unnecessary in this case since layer_count should be the correct amount after the first call.
Although the new version is more correct, I think it is possible to preserve the slight speedup of Charles' version by keeping the initial size of 32 layers and only calling EnumerateInstanceLayerProperties a second time *if* VK_INCOMPLETE was returned.

arno-lunarg:
I imagine that the call to get the count is faster, since no data needs to be returned. Also, the current version can return vectors bigger than the needed size.

arno-lunarg:
Oh but reading Charles comment, my assumption is wrong apparently!

charles-lunarg:
> Also it seems that multiplying by 2 is unnecessary in this case since layer_count should be the correct amount after the first call.

No, multiplying by 2 is correct. When pProperties is null, the 'count' is written, but when pProperties is non-null, the number of written entries is stored in count. This means that if the count is 32 but there are more layers, VK_INCOMPLETE should be returned and count should still be 32 (aka, 32 layer props were written to pProperties).

> I imagine that the call to get the count is faster, since no data needs to be returned.
Ironically, both calls do just about the same amount of work. Only, the first call just returns a number, and in the second call do you get the actual layer properties. The difference is about a handful of memcpys. https://github.com/KhronosGroup/Vulkan-Loader/blob/0a783b5d3e3378528b73abc155e856750a67e50a/loader/loader.c#L6518 is where the code paths diverge.

arno-lunarg:
Well I was wrong, thanks for the explanation!
I am still going to submit the PR to correct the forgotten `pLayerName`

paul-lunarg:
> but when pProperties is non-null, the number of written entries is stored in count

Ugh...
But in this case if there are more than 64 layers/extensions then it is going to be slower than only calling twice anyway, since it will call thrice, right ?

charles-lunarg:
> I think those types of functions were designed to have those 2 successive calls, one for the count, the other to get the data ?

You are definitely correct, that was how they were intended to be used. The rationale for the speedup is to elide unnecessary calls by providing a large enough buffer first. If these API functions are only called a handful of times, its a bit pointless to perform this optimization, since it is much weirder code. In the case of device extensions they are queried many times if I recall correctly. Maybe a better approach is to 'cache' this extension list rather than try to speedup its querying.

And yes, pLayerName being wrong is my fault. I do wonder if pLayerName was ever being used, as it seems that the changes passed...

charles-lunarg:
> But in this case if there are more than 64 layers/extensions then it is going to be slower than only calling twice anyway, since it will call thrice, right ?

Yep! This optimization is called "I assume that there will not be more than 32 layers on a system". Seems that for extensions, 256 was not enough, and MikeS bumped it up to 512.

arno-lunarg:
> And yes, pLayerName being wrong is my fault. I do wonder if pLayerName was ever being used, as it seems that the changes passed...

I am surprised it was not catch by with an "unused variable" warning : /

charles-lunarg (approved):

charles-lunarg:
Also should say that I should of better documented the changes I made. Considering you went through and created a PR to change it back, that shows a lack of comments where comments should have been.

arno-lunarg:
I should have checked the commit history, you described your intent in the commit description, but I just went straight into "let's fix it" mode (～￣▽￣)～

Issue #4274, opened 2022-07-05
VkSwapchainCreateInfoKHR flags unchecked


danilw:
**Issue description:**
As I saw - `VkSwapchainCreateInfoKHR.flags` ignored by Validation layers. (**Vulkan SDK 1.3.216.0** tested today)

**Valid Usage IDs requested**
VUID-VkSwapchainCreateInfoKHR-physicalDeviceCount-01429
VUID-VkSwapchainCreateInfoKHR-flags-03168
VUID-VkSwapchainCreateInfoKHR-flags-04100
VUID-VkSwapchainCreateInfoKHR-flags-03187

**Additional Context**
When `VkSwapchainCreateInfoKHR.flags` is not `0` it results in "very weird behavior" - I had to debug my application line by line for many hours, just to find that I set flag on incorrect value.
When `flags` is 1 - broken UV on small resolution, when `flags` is >1 application crash, when `flags` negative driver crash.

**Visual it looks like this**: (screenshots from Nvidia)

Vulkan samples
![1](https://user-images.githubusercontent.com/24825887/177331285-7718c106-b81e-40f3-b93a-d290b79720a0.png)

My test application in RenderDoc, as you can see "screenshot" of swapchain in RenderDoc not broken, this also made me very confused about source of this bug
![2](https://user-images.githubusercontent.com/24825887/177331546-77e9b2d9-a4c6-4212-9c9a-55f0917b4afc.png)

**Produced bug** (when flags incorrect) - when window resized to "small" UV of swapchain become broken, on every GPU (every OS), on Nvidia its "black vertical lines", on Intel and AMD it "broken UV"(random UV shifts).

It is hard to find reason of bugs like this when Validation layers completely silent, visual result can make you think this is because "wrong memory flags" or some wrong not initialized structure... and checking all of used Vulkan parameters require way too much time.

Test of this - just add to any VkSwapchainCreateInfoKHR parameters `.flags = 1;` and resize window to 85 height or less.

This `VkSwapchainCreateInfoKHR.flags` look completely unchecked by ValidationLayers and result when `flags` is wrong very much ruin whole application, so I think it should be checked by ValidationLayers.

ncesario-lunarg:
[...]

PR #4527, opened 2022-09-08
Fixes for swapchain flags VUIDs

arno-lunarg:
Fixes for issue https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4274

This PR aims at fixing checks for 2 VUIDs:
`VUID-VkSwapchainCreateInfoKHR-physicalDeviceCount-01429`:
> If the logical device was created with VkDeviceGroupDeviceCreateInfo::physicalDeviceCount equal to 1, flags must not contain VK_SWAPCHAIN_CREATE_SPLIT_INSTANCE_BIND_REGIONS_BIT_KHR

`VUID-VkSwapchainCreateInfoKHR-flags-03187`:
> If flags contains VK_SWAPCHAIN_CREATE_PROTECTED_BIT_KHR, then VkSurfaceProtectedCapabilitiesKHR::supportsProtected must be VK_TRUE in the VkSurfaceProtectedCapabilitiesKHR structure returned by vkGetPhysicalDeviceSurfaceCapabilities2KHR for surface

## For 01429

The root issue is how `device_group_create_info`, a member variable that was only correctly initialised when a `VkDeviceGroupDeviceCreateInfo` member was in the `pNext` chain of `VkDeviceCreateInfo`. When it is not, according to [the spec](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkDeviceGroupDeviceCreateInfo.html), this member is implictly defined with `physicalDeviceCount == 1` and `pPhysicalDevices = &physical_device`. But it was not done in `state_tracker.cpp`.

I checked references to `device_group_create_info`, they are in the following functions:
	- `ValidateBindBufferMemory`
	- `ValidateBindImageMemory`
	- `PreCallValidateGetDeviceGroupSurfacePresentModes2EXT`
	- `PreCallValidateGetDeviceGroupSurfacePresentModesKHR`
It seems that nothing seems to be done here after `device_group_create_info` initialisation update.

I updated the validation message in case `VkDeviceGroupDeviceCreateInfo` is implicitly defined, otherwise the original message would be confusing.

Updating `device_group_create_info` showed that `VUID-VkBindImageMemoryDeviceGroupInfo-deviceIndexCount-01634` was not triggered in the test `VkLayerTest.InvalidImageSplitInstanceBindRegionCount`, when it should have been.
VUID description is:
> deviceIndexCount must either be zero or equal to the number of physical devices in the logical device

The number of physical device is 1 in this test, but `deviceIndexCount` is 2. The number of physical device used to be 0, so the VUID was not triggered.

## For 03187

This VUID was only checked if the instance extension `VK_KHR_surface_protected_capabilities` was activated by the user, but the check should be performed if the extension is supported, not necessarily enabled. If the extension is not supported and the corresponding swapchain flag is present, a validation error will be triggered.

## For both VUIDs

I added a new test for the 2 VUIDs, `VkLayerTest.InitSwapchainPotentiallyIncompatibleFlag` in `vklayertest_wsi.cpp` (is it a good place to put it?)

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -17744,11 +17744,14 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
         }
     }
 
-    if ((pCreateInfo->flags & VK_SWAPCHAIN_CREATE_SPLIT_INSTANCE_BIND_REGIONS_BIT_KHR) &&
-        device_group_create_info.physicalDeviceCount == 1) {
+    if ((pCreateInfo->flags & VK_SWAPCHAIN_CREATE_SPLIT_INSTANCE_BIND_REGIONS_BIT_KHR) && physical_device_count == 1) {
         if (LogError(device, "VUID-VkSwapchainCreateInfoKHR-physicalDeviceCount-01429",
                      "%s called with flags containing VK_SWAPCHAIN_CREATE_SPLIT_INSTANCE_BIND_REGIONS_BIT_KHR"
-                     "but logical device was created with VkDeviceGroupDeviceCreateInfo::physicalDeviceCount equal to 1",
+                     "but logical device was created with VkDeviceGroupDeviceCreateInfo::physicalDeviceCount equal to 1."
+                     "The logical device may have been created without explicitly using VkDeviceGroupDeviceCreateInfo, or with"
+                     "VkDeviceGroupDeviceCreateInfo::physicalDeviceCount equal to zero. "
+                     "It is equivalent to using VkDeviceGroupDeviceCreateInfo with "
+                     "VkDeviceGroupDeviceCreateInfo::physicalDeviceCount equal to 1",
                      func_name)) {
             return true;
         }
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -17523,16 +17523,25 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
         }
     }
 
-    if (IsExtEnabled(device_extensions.vk_khr_surface_protected_capabilities) &&
-        (pCreateInfo->flags & VK_SWAPCHAIN_CREATE_PROTECTED_BIT_KHR)) {
-        VkPhysicalDeviceSurfaceInfo2KHR surface_info = LvlInitStruct<VkPhysicalDeviceSurfaceInfo2KHR>();
-        surface_info.surface = pCreateInfo->surface;
-        VkSurfaceProtectedCapabilitiesKHR surface_protected_capabilities = LvlInitStruct<VkSurfaceProtectedCapabilitiesKHR>();
-        VkSurfaceCapabilities2KHR surface_capabilities = LvlInitStruct<VkSurfaceCapabilities2KHR>();
-        surface_capabilities.pNext = &surface_protected_capabilities;
-        DispatchGetPhysicalDeviceSurfaceCapabilities2KHR(physical_device_state->PhysDev(), &surface_info, &surface_capabilities);
+    if (pCreateInfo->flags & VK_SWAPCHAIN_CREATE_PROTECTED_BIT_KHR) {
+        const bool is_required_ext_enabled = IsExtEnabled(instance_extensions.vk_khr_surface_protected_capabilities);
 
-        if (!surface_protected_capabilities.supportsProtected) {
+        // Assume that the "protected" flag is not supported if VK_KHR_surface_protected_capabilities is not enabled
+        bool log_error = !is_required_ext_enabled;
+
+        if (is_required_ext_enabled) {
+            VkPhysicalDeviceSurfaceInfo2KHR surface_info = LvlInitStruct<VkPhysicalDeviceSurfaceInfo2KHR>();
+            surface_info.surface = pCreateInfo->surface;
+            VkSurfaceProtectedCapabilitiesKHR surface_protected_capabilities = LvlInitStruct<VkSurfaceProtectedCapabilitiesKHR>();
+            VkSurfaceCapabilities2KHR surface_capabilities =
+                LvlInitStruct<VkSurfaceCapabilities2KHR>(&surface_protected_capabilities);
+            DispatchGetPhysicalDeviceSurfaceCapabilities2KHR(physical_device_state->PhysDev(), &surface_info,
+                                                             &surface_capabilities);
+
+            log_error = !surface_protected_capabilities.supportsProtected;
+        }
+
+        if (log_error) {
             if (LogError(device, "VUID-VkSwapchainCreateInfoKHR-flags-03187",
                          "%s: pCreateInfo->flags contains VK_SWAPCHAIN_CREATE_PROTECTED_BIT_KHR but the surface "
                          "capabilities does not have VkSurfaceProtectedCapabilitiesKHR.supportsProtected set to VK_TRUE.",
--- a/tests/vklayertests_wsi.cpp
+++ b/tests/vklayertests_wsi.cpp
@@ -29,6 +29,89 @@
 #include "layer_validation_tests.h"
 #include "core_validation_error_enums.h"
 
+TEST_F(VkLayerTest, InitSwapchainPotentiallyIncompatibleFlag) {
+    TEST_DESCRIPTION("Initialize swapchain with potentially incompatible flags");
+
+    SetTargetApiVersion(VK_API_VERSION_1_1);
+
+    AddRequiredExtensions(VK_KHR_SURFACE_PROTECTED_CAPABILITIES_EXTENSION_NAME);
+    AddSurfaceExtension();
+
+    ASSERT_NO_FATAL_FAILURE(InitFramework(m_errorMonitor));
+    if (DeviceValidationVersion() < VK_API_VERSION_1_1) {
+        GTEST_SKIP() << "At least Vulkan version 1.1 is required";
+    }
+    if (!AreRequiredExtensionsEnabled()) {
+        GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported";
+    }
+
+    ASSERT_NO_FATAL_FAILURE(InitState());
+
+    ASSERT_NO_FATAL_FAILURE(InitRenderTarget());
+    if (!InitSurface()) {
+        GTEST_SKIP() << "Cannot create surface, skipping test";
+    }
+    InitSwapchainInfo();
+
+    VkSwapchainCreateInfoKHR swapchain_ci = LvlInitStruct<VkSwapchainCreateInfoKHR>();
+    swapchain_ci.surface = m_surface;
+    swapchain_ci.minImageCount = m_surface_capabilities.minImageCount;
+    swapchain_ci.imageFormat = m_surface_formats[0].format;
+    swapchain_ci.imageColorSpace = m_surface_formats[0].colorSpace;
+    swapchain_ci.imageExtent = {m_surface_capabilities.minImageExtent.width, m_surface_capabilities.minImageExtent.height};
+    swapchain_ci.imageArrayLayers = 1;
+    swapchain_ci.imageUsage = VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT;
+    swapchain_ci.imageSharingMode = VK_SHARING_MODE_EXCLUSIVE;
+    swapchain_ci.preTransform = VK_SURFACE_TRANSFORM_IDENTITY_BIT_KHR;
+    swapchain_ci.compositeAlpha = m_surface_composite_alpha;
+    swapchain_ci.presentMode = m_surface_non_shared_present_mode;
+    swapchain_ci.clipped = VK_FALSE;
+    swapchain_ci.oldSwapchain = 0;
+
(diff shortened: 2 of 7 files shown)


layers/core_validation.cpp
```diff
@@ -17523,16 +17523,50 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
...
-        VkPhysicalDeviceSurfaceInfo2KHR surface_info = LvlInitStruct<VkPhysicalDeviceSurfaceInfo2KHR>();
-        surface_info.surface = pCreateInfo->surface;
-        VkSurfaceProtectedCapabilitiesKHR surface_protected_capabilities = LvlInitStruct<VkSurfaceProtectedCapabilitiesKHR>();
-        VkSurfaceCapabilities2KHR surface_capabilities = LvlInitStruct<VkSurfaceCapabilities2KHR>();
-        surface_capabilities.pNext = &surface_protected_capabilities;
-        DispatchGetPhysicalDeviceSurfaceCapabilities2KHR(physical_device_state->PhysDev(), &surface_info, &surface_capabilities);
+    if (pCreateInfo->flags & VK_SWAPCHAIN_CREATE_PROTECTED_BIT_KHR) {
+        // Look for VK_KHR_surface_protected_capabilities
+        bool is_required_ext_enabled = IsExtEnabled(instance_extensions.vk_khr_surface_protected_capabilities);
+#if 0
```
arno-lunarg, line 17529: this is where I would like some insights!

tests/vklayertests_wsi.cpp
```diff
@@ -361,19 +448,14 @@ TEST_F(VkLayerTest, ValidSwapchainImageParams) {
         GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported";
     }
 
-    if (!AreRequiredExtensionsEnabled()) {
-        GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported.";
```
arno-lunarg, line 365: this check is needlessly duplicated, so I removed it

layers/core_validation.cpp
```diff
@@ -17744,12 +17778,24 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
         }
     }
 
-    if ((pCreateInfo->flags & VK_SWAPCHAIN_CREATE_SPLIT_INSTANCE_BIND_REGIONS_BIT_KHR) &&
-        device_group_create_info.physicalDeviceCount == 1) {
+    if ((pCreateInfo->flags & VK_SWAPCHAIN_CREATE_SPLIT_INSTANCE_BIND_REGIONS_BIT_KHR) && physical_device_count == 1) {
+        const auto *device_group_ci = LvlFindInChain<VkDeviceGroupDeviceCreateInfo>(pCreateInfo->pNext);
```
jeremyg-lunarg, line 17782: VkDeviceGroupDeviceCreateInfo  would be attached to the create info for vkCreateDevice(), but you're looking for it here in VkSwapchainCreateInfoKHR. See VUID-VkSwapchainCreateInfoKHR-pNext-pNext

layers/core_validation.cpp
jeremyg-lunarg: Why can't you use IsExtEnabled(instance_extensions.vk_khr_get_surface_capabilities2) here?  Also VK_KHR_surface_protected_capabilities requires capabilities2 so I'm pretty sure some other part of validation will catch if you have one but not the other.

arno-lunarg: Oh I agree there is no problem here, my problem is that I need the extension to be enabled to get `VkSurfaceProtectedCapabilitiesKHR surface_protected_capabilities`, and I am not sure about how to activate it myself when the extension is disabled.

layers/core_validation.cpp
arno-lunarg: Oops, to fix that I removed this additional logic, and just updated the message:
```cpp
    if ((pCreateInfo->flags & VK_SWAPCHAIN_CREATE_SPLIT_INSTANCE_BIND_REGIONS_BIT_KHR) && physical_device_count == 1) {
        if (LogError(device, "VUID-VkSwapchainCreateInfoKHR-physicalDeviceCount-01429",
                     "%s called with flags containing VK_SWAPCHAIN_CREATE_SPLIT_INSTANCE_BIND_REGIONS_BIT_KHR"
                     "but logical device was created with VkDeviceGroupDeviceCreateInfo::physicalDeviceCount equal to 1."
                     "The logical device may have been created without explicitly using VkDeviceGroupDeviceCreateInfo, or with"
                     "VkDeviceGroupDeviceCreateInfo::physicalDeviceCount equal to zero. "
                     "It is equivalent to using VkDeviceGroupDeviceCreateInfo with "
                     "VkDeviceGroupDeviceCreateInfo::physicalDeviceCount equal to 1",
                     func_name)) {
            return true;
        }
    }
```

layers/core_validation.cpp
jeremyg-lunarg: From vk.xml:
```
 <extension name="VK_KHR_surface_protected_capabilities" number="240" type="instance" 
         requiresCore="1.1" requires="VK_KHR_get_surface_capabilities2" ...
```

So if the user enabled VK_KHR_surface_protected_capabilities, VK_KHR_get_surface_capabilities2 must be enabled. If they did not enable VK_KHR_surface_protected_capabilities, you should assume `surface_protected_capabilities.supportsProtected = VK_FALSE;`

If you wanted to force an extension on, you'd have to do during vkCreateInstance(), which would be a mess.

arno-lunarg: Ok, I was trying to enable VK_KHR_surface_protected_capabilities because of what Mike suggested when we last talked:
> I think the check for VUID 03187 is wrong, we should query the surface caps if VK_KHR_surface_protected_capabilities is supported, not enabled

jeremyg-lunarg: @mikes-lunarg is there a better way to handle this problem?

tests/vklayertests_wsi.cpp
```diff
@@ -29,6 +29,93 @@
 #include "layer_validation_tests.h"
 #include "core_validation_error_enums.h"
 
+TEST_F(VkLayerTest, InitSwapchainPotentiallyIncompatibleFlag) {
+    TEST_DESCRIPTION("Initialize swapchain with potentially incompatible flags");
+
+    SetTargetApiVersion(VK_API_VERSION_1_1);
+
+    // VK_KHR_surface_protected_capabilities requires Vulkan 1.1 and instance extension VK_KHR_get_surface_capabilities2
+    if (!InstanceExtensionSupported(VK_KHR_GET_SURFACE_CAPABILITIES_2_EXTENSION_NAME)) {
```
sjfricke, line 38: Please use the new `AreRequiredExtensionsEnabled` pattern (I realize not all the tests have fully embraced this yet)

```patch
-    if (!InstanceExtensionSupported(VK_KHR_GET_SURFACE_CAPABILITIES_2_EXTENSION_NAME)) {
-        GTEST_SKIP() << "Did not find required instance extension " VK_KHR_GET_SURFACE_CAPABILITIES_2_EXTENSION_NAME;
-    }
-    if (!InstanceExtensionSupported(VK_KHR_SURFACE_PROTECTED_CAPABILITIES_EXTENSION_NAME)) {
-        GTEST_SKIP() << "Did not find required instance extension " VK_KHR_SURFACE_PROTECTED_CAPABILITIES_EXTENSION_NAME;
-    }
-
+    AddRequiredExtensions(VK_KHR_GET_SURFACE_CAPABILITIES_2_EXTENSION_NAME);
+    AddRequiredExtensions(VK_KHR_SURFACE_PROTECTED_CAPABILITIES_EXTENSION_NAME);
    AddSurfaceExtension();
-    instance_extensions_.push_back(VK_KHR_GET_SURFACE_CAPABILITIES_2_EXTENSION_NAME);
-    instance_extensions_.push_back(VK_KHR_SURFACE_PROTECTED_CAPABILITIES_EXTENSION_NAME);

    }
```

ncesario-lunarg: :heavy_plus_sign:

tests/vkrenderframework.cpp
```diff
@@ -657,6 +657,9 @@ bool VkRenderFramework::AddRequestedInstanceExtensions(const char *ext_name) {
     if (is_instance_ext) {
         const auto &info = InstanceExtensions::get_info(ext_name);
         for (const auto &req : info.requirements) {
+            if (0 == strncmp(req.name, "VK_VERSION", 10)) {
+                continue;
```
arno-lunarg, line 661: This is an hack to prevent fake extensions "VK_VERSION_1_x" from being added to the instance extension list.
@ncesario-lunarg modifying `InstanceExtensionSupported` by adding:
```cpp
if (0 == strncmp(extension_name, "VK_VERSION", 10)) return true;
```
does not prevent those extensions from going into the list given the current code flow, unfortunately

layers/core_validation.cpp
```diff
@@ -17523,16 +17523,25 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
...
+
+        // Assume that the "protected" flag is not supported if VK_KHR_surface_protected_capabilities is not enabled
+        bool log_error = !is_required_ext_enabled;
+
+        if (is_required_ext_enabled) {
+            VkPhysicalDeviceSurfaceInfo2KHR surface_info = LvlInitStruct<VkPhysicalDeviceSurfaceInfo2KHR>();
+            surface_info.surface = pCreateInfo->surface;
+            VkSurfaceProtectedCapabilitiesKHR surface_protected_capabilities = LvlInitStruct<VkSurfaceProtectedCapabilitiesKHR>();
+            VkSurfaceCapabilities2KHR surface_capabilities = LvlInitStruct<VkSurfaceCapabilities2KHR>();
+            surface_capabilities.pNext = &surface_protected_capabilities;
```
ncesario-lunarg, line 17537: Just an FYI that you can pass in a pNext value to `LvlInitStruct`. e.g., `LvlInitStruct<VkSurfaceCapabilities2KHR>(&surface_protected_capabilities)`.

tests/vklayertests_wsi.cpp
```diff
@@ -29,6 +29,84 @@
...
 
+TEST_F(VkLayerTest, InitSwapchainPotentiallyIncompatibleFlag) {
+    TEST_DESCRIPTION("Initialize swapchain with potentially incompatible flags");
+
+    SetTargetApiVersion(VK_API_VERSION_1_1);
+
+    AddRequiredExtensions(VK_KHR_SURFACE_PROTECTED_CAPABILITIES_EXTENSION_NAME);
+    AddSurfaceExtension();
+
+    ASSERT_NO_FATAL_FAILURE(InitFramework(m_errorMonitor));
```
ncesario-lunarg, line 40: Probably want a `DeviceValidationVersion() < VK_API_VERSION_1_1` check after this.

tests/vklayertests_wsi.cpp
```diff
@@ -29,6 +29,84 @@
...
+    swapchain_ci.presentMode = m_surface_non_shared_present_mode;
+    swapchain_ci.clipped = VK_FALSE;
+    swapchain_ci.oldSwapchain = 0;
+
+    // "protected" flag support is device defined
+    {
+        swapchain_ci.flags = VK_SWAPCHAIN_CREATE_PROTECTED_BIT_KHR;
+
+        PFN_vkGetPhysicalDeviceSurfaceCapabilities2KHR vkGetPhysicalDeviceSurfaceCapabilities2KHR =
+            (PFN_vkGetPhysicalDeviceSurfaceCapabilities2KHR)vk::GetInstanceProcAddr(instance(),
```
ncesario-lunarg, line 71: This should be a `reinterpret_cast` per the [guidelines](https://google.github.io/styleguide/cppguide.html#Casting).

ncesario-lunarg (commented):
A couple minor nits, but otherwise LGTM!

tests/vklayertests_wsi.cpp
```diff
@@ -29,6 +29,87 @@
...
+
+    AddRequiredExtensions(VK_KHR_SURFACE_PROTECTED_CAPABILITIES_EXTENSION_NAME);
+    AddSurfaceExtension();
+
+    ASSERT_NO_FATAL_FAILURE(InitFramework(m_errorMonitor));
+    if (!AreRequiredExtensionsEnabled()) {
+        GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported";
+    }
+    ASSERT_NO_FATAL_FAILURE(InitState());
+    if (DeviceValidationVersion() < VK_API_VERSION_1_1) {
```
ncesario-lunarg, line 45: This check can be moved up to before `InitState` (where the logical device is created), but _after_ `InitFramework` (where the physical device is chosen).

ncesario-lunarg (approved):
LGTM!

tests/vklayertests_wsi.cpp
arno-lunarg: Oh ok, I was wrongly thinking that logical device was needed for this test

arno-lunarg:
Fixes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4274

PR #4528, opened 2022-09-08
Fix ForEachSubresource

paul-lunarg:
Was looking to use ForEachSubresource and saw something weird.
the function would crash when using VK_REMAINING_ARRAY_LAYERS and baseArrayLayer > 0  or VK_REMAINING_MIP_LEVELS and baseMipLevel > 0.
Fix to correctly calculate layerCount/levelCount with REMAINING values.
Modify test to use layer index 1 to test this new behavior.


paul-lunarg:
@rlocatti-nv Does this fix look good to you ?

rlocatti-nv:
It looks good to me.

PR #4534, opened 2022-09-09
Add missing VUIDs and tests for VUID-VkPipelineRobustnessCreateInfoEXT-*

JarredHDavies:


tests/vklayertests_pipeline_shader.cpp
```diff
@@ -17683,3 +17683,218 @@ TEST_F(VkLayerTest, ShaderModuleIdentifierFeatures) {
     vkGetShaderModuleCreateInfoIdentifierEXT(device(), &sm_ci, &get_identifier);
     m_errorMonitor->VerifyFound();
 }
+
+TEST_F(VkLayerTest, PipelineRobustnessDisabled) {
+    TEST_DESCRIPTION("Test for VK_EXT_pipeline_robustness extension.");
+    TEST_DESCRIPTION("Create a pipeline using VK_EXT_pipeline_robustness but with pipelineRobustness == VK_FALSE");
```
ncesario-lunarg, line 17689: I can't recall any tests that call `TEST_DESCRIPTION` more than once. What effect does this have?

ncesario-lunarg (commented):
This is great, thanks @JarredHDavies! Some minor nits and questions, but this looks really good.

I am curious about the "supported" vs "enabled" features as this seems to digress from the [norm](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap41.html), specifically:
> Features are optional, and must be explicitly enabled before use.

and
> Some features, such as robustBufferAccess, may incur a runtime performance cost. Application writers should carefully consider the implications of enabling all supported features.

For this reason I would think that we want to check whether the user enabled these features rather than whether or not the device supports them, but that seems to conflict with the language of the VUIDs.

Tony-LunarG:
Has GPU-AV out of bounds read / write checking been considered?  It's turned off device wide if robust buffer access is enabled, but it would seem to need to be turned off per stage / pipeline with this extension.

layers/stateless_validation.h
```diff
@@ -118,6 +118,15 @@ class StatelessValidation : public ValidationObject {
     };
     DeviceExtensionProperties phys_dev_ext_props = {};
 
+    // VK_EXT_pipeline_robustness functionality changes depending on whether a device supports the
+    // robustBufferAccess2 or robustImageAccess features, these features do not need to be enabled
+    // for the functionality to be used
+    struct DeviceRobustnessExtensionFeatures {
```
sjfricke, line 124: why is this added in `statelass_validation.h` and not in `device_state.h` `DeviceFeatures` struct like all the other features?

layers/parameter_validation_utils.cpp
```diff
@@ -358,6 +366,18 @@ void StatelessValidation::PostCallRecordCreateDevice(VkPhysicalDevice physicalDe
         phys_dev_ext_props.maintenance4_props = maintance4_props;
     }
 
+    if (IsExtEnabled(device_extensions.vk_ext_pipeline_robustness)) {
```
sjfricke, line 369: related to my other comment, there is already a `enabled_features.*` system for getting features which we do in `ValidationStateTracker::CreateDevice()`

layers/parameter_validation_utils.cpp
```diff
@@ -1765,6 +1785,101 @@ bool StatelessValidation::manual_PreCallValidateCreatePipelineLayout(VkDevice de
...
+bool StatelessValidation::ValidatePipelineRobustnessCreateInfo(const char *func_name, const char *parameter_name,
+                                                               const VkPipelineRobustnessCreateInfoEXT *pCreateInfo) const {
+    bool skip = false;
+
+    if (pCreateInfo) {
+        const auto *pipeline_robustness_features =
+            LvlFindInChain<VkPhysicalDevicePipelineRobustnessFeaturesEXT>(device_createinfo_pnext);
+
+        if (pipeline_robustness_features && pipeline_robustness_features->pipelineRobustness) {
+            if (!phys_dev_robustness_ext_features.image_robustness_features.robustImageAccess) {
```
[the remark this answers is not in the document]
JarredHDavies: Yeah these features are an exception now, I probably should have added extra properties looking back at the extension now but when it was written it seemed sensible to me that you could just look at what features a device supports.

tests/vklayertests_pipeline_shader.cpp
JarredHDavies: The ShaderModuleIdentifier test above this test does it and I just based the description from that. It looks like the description attribute get's overwritten so I'll change that.

layers/stateless_validation.h
JarredHDavies: It's my understanding that the DeviceFeatures is only for tracking enabled features (or at least I can't see anywhere where it's used otherwise), where as this extension needs to know whether a device supports a feature and doesn't care whether it's enabled or not. That kind of makes those features similar to a device property which is why put it with the property tracking code.

sjfricke: so this already was done with `VK_KHR_format_feature_flags2` which also is in the "it matters if its supported, not enabled" camp

So my suggestion is look where `has_format_feature2` and make `has_pipeline_robustness` so both are in the same spot incase they need to get moved in the future

Tony-LunarG:
> Has GPU-AV out of bounds read / write checking been considered? It's turned off device wide if robust buffer access is enabled, but it would seem to need to be turned off per stage / pipeline with this extension.

We are planning on refactoring GPU-AV for better descriptor indexing coverage, and we have an idea for a way to do this as a first step in that refactor.  As such, we'd propose that LunarG make the GPU-AV changes to deal with per-pipeline buffer out of bounds reads and writes.  @JarredHDavies is that ok with you?

JarredHDavies:
> > Has GPU-AV out of bounds read / write checking been considered? It's turned off device wide if robust buffer access is enabled, but it would seem to need to be turned off per stage / pipeline with this extension.
> 
> We are planning on refactoring GPU-AV for better descriptor indexing coverage, and we have an idea for a way to do this as a first step in that refactor. As such, we'd propose that LunarG make the GPU-AV changes to deal with per-pipeline buffer out of bounds reads and writes. @JarredHDavies is that ok with you?

That sounds great for me. I'm always happy for other people to do my work for me :D

layers/stateless_validation.h
ncesario-lunarg: This probably _should_ go into `state_tracker.h` if possible. That may mean that some of the checks need to move out of `parameter_validation_utils.cpp` and into `core_validation.cpp`, which isn't ideal, but would be more "correct" if the checks are indeed reading state.

ncesario-lunarg (commented):
I think once @sjfricke's comments are resolved this should be good to go.

KarenGhavam-lunarG:
> @JarredHDavies Was wondering if we could get this finished... The GPU side of this validation was merged and it would be good to get this PR merged to finish the validation for this extension.

Pinging @JarredHDavies again. This extension has been released now for  6 months. Would be good to get the validation layer implementation completed.

spencer-lunarg:
closing, rebased here https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5539

PR #4538, opened 2022-09-09
Add new VK_EXT_multisampled_render_to_single_sampled VUIDs

Tony-LunarG:


layers/drawdispatch.cpp
```diff
@@ -1019,6 +1060,9 @@ struct DispatchVuidsCmdDrawMeshTasksIndirectCountNV : DrawDispatchVuid {
         dynamic_rendering_06189            = "VUID-vkCmdDrawMeshTasksIndirectCountNV-pDepthAttachment-06189";
```
Tony-LunarG, line 1060: Does anyone know why a chunk of these are initialized twice?

ncesario-lunarg: I'm guessing that is a mistake and the duplicate should be removed.

ncesario-lunarg (approved):
LGTM!

layers/drawdispatch.cpp
Tony-LunarG: Duplicates removed

jeremyg-lunarg (approved):
LGTM

PR #4540, opened 2022-09-12
Add OpImageWrite component count check

sjfricke:
Adds new `VUID-RuntimeSpirv-OpImageWrite-07112` support

This only validates of the `OpImageType` format is not `Unknown`, if it is then `VUID-vkCmdDispatch-None-04115` and `VUID-vkCmdDispatch-OpImageWrite-04469` are used... I was going to add these 2 as well in here, but some other refactoring was needed and bloated this PR

New `CompatibleSpirvImageFormat` generated to map the table in the Spec which SPIR-V image formats go with which `VkFormat`

### DIFF
--- a/build-android/jni/Android.mk
+++ b/build-android/jni/Android.mk
@@ -29,6 +29,7 @@ LOCAL_SRC_FILES += $(SRC_DIR)/layers/generated/vk_format_utils.cpp
 LOCAL_C_INCLUDES += $(VULKAN_INCLUDE) \
                     $(LOCAL_PATH)/$(SRC_DIR)/layers/generated \
                     $(LOCAL_PATH)/$(SRC_DIR)/layers \
+                    $(LOCAL_PATH)/$(THIRD_PARTY)/shaderc/third_party/spirv-tools/external/spirv-headers/include \
                     $(LOCAL_PATH)/$(THIRD_PARTY)/robin-hood-hashing/src/include
 LOCAL_CPPFLAGS += -std=c++11 -Wall -Werror -Wno-unused-function -Wno-unused-const-variable
 LOCAL_CPPFLAGS += -DVK_ENABLE_BETA_EXTENSIONS -DVK_USE_PLATFORM_ANDROID_KHR -DVK_PROTOTYPES -DUSE_ROBIN_HOOD_HASHING -fvisibility=hidden
--- a/layers/generated/vk_format_utils.cpp
+++ b/layers/generated/vk_format_utils.cpp
@@ -30,6 +30,7 @@
 #include "vk_layer_utils.h"
 #include <map>
 #include <vector>
+#include <spirv/unified1/spirv.hpp>
 
 
 enum class COMPONENT_TYPE {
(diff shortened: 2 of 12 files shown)


sjfricke:
@ncesario-lunarg not sure looking at the `build.gn` script where it would want the SPIRV-Header path as there is nothing explicit for the other files in it (also never worked with the `gn` build system)

layers/generated/vk_format_utils.cpp
```diff
@@ -1927,3 +1928,94 @@ bool FormatHasAlpha(VkFormat format) {
     return FormatHasComponent(format, COMPONENT_TYPE::A);
 }
 
+// Will return the Vulkan format for a given SPIR-V image format value
+// Note: will return VK_FORMAT_UNDEFINED if non valid input
+VkFormat CompatibleSpirvImageFormat(int spirv_image_format) {
```
ncesario-lunarg, line 1933: `int` -> `uint32_t` for `spirv_image_format` type?

tests/positive/shaderval.cpp
```diff
@@ -2934,3 +2934,384 @@ TEST_F(VkPositiveLayerTest, OpTypeStructRuntimeArray) {
     };
     CreateComputePipelineHelper::OneshotTest(*this, set_info, kErrorBit);
 }
+
+TEST_F(VkPositiveLayerTest, StorageImageWriteMoreComponent) {
+    TEST_DESCRIPTION("Test writing to image with less compoents.");
```
ncesario-lunarg, line 2939: `compoents` -> `components` (and similarly below).

ncesario-lunarg (commented):
Couple minor nits, but LGTM!

build-android/jni/Android.mk
```diff
@@ -29,6 +29,7 @@ LOCAL_SRC_FILES += $(SRC_DIR)/layers/generated/vk_format_utils.cpp
 LOCAL_C_INCLUDES += $(VULKAN_INCLUDE) \
                     $(LOCAL_PATH)/$(SRC_DIR)/layers/generated \
                     $(LOCAL_PATH)/$(SRC_DIR)/layers \
+                    $(LOCAL_PATH)/$(THIRD_PARTY)/shaderc/third_party/spirv-tools/external/spirv-headers/include \
```
ncesario-lunarg, line 32: I'm guessing something similar is needed at https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/b48bd2b501be8f63de7279da1d675ec6fef82c3e/BUILD.gn#L271

sjfricke: Ok, so there is a `./build-android/secondary/build_overrides/vulkan_validation_layers.gni` that has the paths I am looking for

ncesario-lunarg: Ah, good to know...though, it doesn't really make sense for that to be under the `build-android` directory.

ncesario-lunarg (approved):

build-android/jni/Android.mk
sjfricke: sorry, for record it is in `build-gn/secondary/build_overrides/vulkan_validation_layers.gni` not `build-android` so make sense :laughing:

Issue #3823, opened 2022-03-09
practices: massive memory use and performance problems from ARM Image layout checks


jeremyg-lunarg:
**Describe the Issue**
The vector of function objects in  BestPractices::QueueValidateImage can become quite huge if validation is run against an application with large descriptor sets.  The checking here is very similar to the image layout validation in CoreChecks and it should be restuctured to match. Namely, keep a map or set of the images in use in the command buffer (because they often repeat) and use the sparse range maps to track information about subresources. 

**Valid Usage ID**
NONE

**Environment:**
 - OS: any
 - GPU: any
 - SDK or header version if building from repo: 1.3.204 (but really any)
 - Options enabled (synchronization, best practices, etc.):
best practices
**Additional context**

PR #4542, opened 2022-09-12
Implement check for ownership transfer

### LINKED ISSUE #2412: Missing validation for queue ownership transfer for image with VK_SHARING_MODE_EXCLUSIVE
Issue description:
I use a graphics and a compute queue with different queueFamilyIndices.
1. Create an image with sharingMode = VK_SHARING_MODE_EXCLUSIVE
2. Using the image as a framebuffer attachment in a renderpass on graphics queue.
3. Submitting on graphics queue with signal semaphore
4. Binding image as sampled image in a compute shader with a DescriptorUpdateTemplate.
5. Submitting work on compute queue with wait on aforementioned semaphore

I expected some validation error in the sort of: VkImage with sharingMode=VK_SHARING_MODE_EXCLUSIVE is used on queues with different queueFamilyIndices without explicit ownership transfer.

Additional Context:
The image layout conversions of the image are done with renderpass subpass dependencies
I don't use any VkImageMemoryBarrier for this image as layout transitions are done in the renderpass and making the memory visible is done with the semaphore.

paul-lunarg:
Keeps the last used queue family with usage per subresource for images
Whenever a new usage appears on queue on a different queue family, check if it is a read operation because it will be undefined
Check for an image pipeline barrier defining an ownership acquisition operation and change the queue family index
Only for images for now, buffers later

Closes #2412

### DIFF
--- a/layers/best_practices_utils.cpp
+++ b/layers/best_practices_utils.cpp
@@ -2652,8 +2652,7 @@ void BestPractices::ValidateImageInQueueArmImg(const char* function_name, const 
 void BestPractices::ValidateImageInQueue(const ValidationStateTracker& vst, const QUEUE_STATE& qs, const CMD_BUFFER_STATE& cbs,
                                          const char* function_name, bp_state::Image& state, IMAGE_SUBRESOURCE_USAGE_BP usage,
                                          uint32_t array_layer, uint32_t mip_level) {
-    auto last_usage = state.UpdateUsage(array_layer, mip_level, usage);
-    auto last_queue_family = state.UpdateQueueFamily(qs.queueFamilyIndex);
+    auto last_usage = state.UpdateUsage(array_layer, mip_level, usage, qs.queueFamilyIndex);
 
     // When image was discarded with StoreOpDontCare but is now being read with LoadOpLoad
     if (last_usage == IMAGE_SUBRESOURCE_USAGE_BP::RENDER_PASS_DISCARDED &&
--- a/layers/best_practices_utils.cpp
+++ b/layers/best_practices_utils.cpp
@@ -2568,9 +2568,9 @@ void BestPractices::QueueValidateImage(QueueCallbacks& funcs, const char* functi
 
 void BestPractices::QueueValidateImage(QueueCallbacks& funcs, const char* function_name, std::shared_ptr<bp_state::Image>& state,
                                        IMAGE_SUBRESOURCE_USAGE_BP usage, uint32_t array_layer, uint32_t mip_level) {
-    funcs.push_back([this, function_name, state, usage, array_layer, mip_level](const ValidationStateTracker&, const QUEUE_STATE&,
-                                                                                const CMD_BUFFER_STATE&) -> bool {
-        ValidateImageInQueue(function_name, *state, usage, array_layer, mip_level);
+    funcs.push_back([this, function_name, state, usage, array_layer, mip_level](
+                        const ValidationStateTracker& vst, const QUEUE_STATE& qs, const CMD_BUFFER_STATE& cbs) -> bool {
+        ValidateImageInQueue(vst, qs, cbs, function_name, *state, usage, array_layer, mip_level);
         return false;
     });
 }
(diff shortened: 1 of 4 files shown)


layers/best_practices_validation.h
```diff
@@ -177,26 +177,30 @@ class Image : public IMAGE_STATE {
...
+            mip_vec.resize(createInfo.mipLevels, std::make_pair(IMAGE_SUBRESOURCE_USAGE_BP::UNDEFINED, VK_QUEUE_FAMILY_IGNORED));
         }
     }
     // A 2d vector for all the array layers and mip levels.
     // This does not split usages per aspect.
     // Aspects are generally read and written together,
     // and tracking them independently could be misleading.
-    std::vector<std::vector<IMAGE_SUBRESOURCE_USAGE_BP>> usages_;
+    // second/uint32_t is last queue family usage
+    std::vector<std::vector<std::pair<IMAGE_SUBRESOURCE_USAGE_BP, uint32_t>>> usages_;
```
jeremyg-lunarg, line 203: It would be clearer to make a struct for this, especially if it is likely that there will be more information you want to track here in the future.

layers/best_practices_utils.cpp
```diff
@@ -2649,9 +2649,31 @@ void BestPractices::ValidateImageInQueueArmImg(const char* function_name, const
...
 
-void BestPractices::ValidateImageInQueue(const char* function_name, bp_state::Image& state, IMAGE_SUBRESOURCE_USAGE_BP usage,
-                                         uint32_t array_layer, uint32_t mip_level) {
-    auto last_usage = state.UpdateUsage(array_layer, mip_level, usage);
+void BestPractices::ValidateImageInQueue(const QUEUE_STATE& qs, const CMD_BUFFER_STATE& cbs, const char* function_name,
+                                         bp_state::Image& state, IMAGE_SUBRESOURCE_USAGE_BP usage, uint32_t array_layer,
+                                         uint32_t mip_level) {
+    auto last_queue_family = state.GetLastQueueFamily(array_layer, mip_level);
+    auto queue_family = qs.queueFamilyIndex;
+    auto last_usage = state.UpdateUsage(array_layer, mip_level, usage, queue_family);
```
jeremyg-lunarg, line 2657: This is the only call location for UpdateUsage(). You could change it to return your pair/struct without worrying about any other code. This would save the need for duplicating the same lookup in GetLastQueueFamily().

jeremyg-lunarg (commented):
Eventually the usage info in BP will need to be converted to work like the ImageLayoutMap in core validation. See https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/3823 
That doesn't have to happen for this PR to go in, but you should be aware of it.

layers/best_practices_validation.h
ncesario-lunarg: ➕

layers/best_practices_utils.cpp
```diff
@@ -2649,9 +2649,31 @@ void BestPractices::ValidateImageInQueueArmImg(const char* function_name, const
...
+        // if UNDEFINED then first use/acquisition of subresource
+        if (last_usage != IMAGE_SUBRESOURCE_USAGE_BP::UNDEFINED) {
+            // If usage might read from the subresource, as contents are undefined
+            // so write only is fine
+            if (usage == IMAGE_SUBRESOURCE_USAGE_BP::RENDER_PASS_READ_TO_TILE || usage == IMAGE_SUBRESOURCE_USAGE_BP::BLIT_READ ||
+                usage == IMAGE_SUBRESOURCE_USAGE_BP::COPY_READ || usage == IMAGE_SUBRESOURCE_USAGE_BP::DESCRIPTOR_ACCESS ||
+                usage == IMAGE_SUBRESOURCE_USAGE_BP::RESOLVE_READ) {
+                LogWarning(
+                    state.Handle().Cast<VkImage>(), kVUID_BestPractices_ConcurrentUsageOfExclusiveImage,
+                    "%s : Subresource (arrayLayer: %u, mipLevel: %u) of image is used on queue family index %u after being used on "
```
ncesario-lunarg, line 2670: `%u` -> `PRIu32`

ncesario-lunarg (commented):
LGTM once JeremyG's comments have been resolved.

paul-lunarg:
I addressed your comments, it is much better with a struct!

ncesario-lunarg (approved):

jeremyg-lunarg (approved):
LGTM

PR #4546, opened 2022-09-13
Use correct attachment storage for imageless

paul-lunarg:
Check was using framebuffer state to check attachment image views which was causing a crash in CTS, because imageless framebuffers have no image views.


Tony-LunarG (approved):
LGTM and fixes CTS crash

PR #4548, opened 2022-09-13
Add checks for 06774 and 06775

### LINKED ISSUE #4543: Dynamic rendering missing VUIDs
While working on CTS found a VUID that should be triggered by some tests but I was not hitting it. Found it's yet to be implemented. VUID list:
- [ ] VUID-vkCmdExecuteCommands-pDepthAttachment-06774
- [ ] VUID-vkCmdExecuteCommands-pStencilAttachment-06775

paul-lunarg:
Changed dynamic rendering messages to stop using the VkRenderPass object as it is always VK_NULL_HANDLE

Closes #4543

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -15912,6 +15912,38 @@ bool CoreChecks::PreCallValidateCmdExecuteCommands(VkCommandBuffer commandBuffer
                             }
                         }
 
+                        if (cb_state->activeRenderPass->dynamic_rendering_begin_rendering_info.pDepthAttachment == nullptr ||
+                            cb_state->activeRenderPass->dynamic_rendering_begin_rendering_info.pDepthAttachment->imageView ==
+                                VK_NULL_HANDLE) {
+                            VkFormat format = sub_cb_state->activeRenderPass->inheritance_rendering_info.depthAttachmentFormat;
+                            if (format != VK_FORMAT_UNDEFINED) {
+                                skip |= LogError(
+                                    pCommandBuffers[i], "VUID-vkCmdExecuteCommands-pDepthAttachment-06774",
+                                    "vkCmdExecuteCommands(): Secondary %s is executed within a dynamic renderpass "
+                                    "instance scope begun by %s(), and VkRenderingInfo::pDepthAttachment does not define an "
+                                    "image view but VkCommandBufferInheritanceRenderingInfo::depthAttachmentFormat "
+                                    "is %s instead of VK_FORMAT_UNDEFINED.",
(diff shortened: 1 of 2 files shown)


layers/core_validation.cpp
```diff
@@ -15902,33 +15892,63 @@ bool CoreChecks::PreCallValidateCmdExecuteCommands(VkCommandBuffer commandBuffer
...
+                                                 cb_state->begin_rendering_func_name.c_str());
+                            }
+                        }
+
+                        if (cb_state->activeRenderPass->dynamic_rendering_begin_rendering_info.pDepthAttachment == nullptr ||
+                            cb_state->activeRenderPass->dynamic_rendering_begin_rendering_info.pDepthAttachment->imageView ==
+                                VK_NULL_HANDLE) {
+                            if (sub_cb_state->activeRenderPass->inheritance_rendering_info.depthAttachmentFormat !=
+                                VK_FORMAT_UNDEFINED) {
+                                LogObjectList objlist(pCommandBuffers[i]);
```
ncesario-lunarg, line 15911: If logging a single handle, is there a reason to use a `LogObjectList` here (same question for instance below)?

layers/core_validation.cpp
```diff
@@ -15902,33 +15892,63 @@ bool CoreChecks::PreCallValidateCmdExecuteCommands(VkCommandBuffer commandBuffer
...
+                                skip |= LogError(
+                                    objlist, "VUID-vkCmdExecuteCommands-pDepthAttachment-06774",
+                                    "vkCmdExecuteCommands(): Secondary %s is executed within a dynamic renderpass "
+                                    "instance scope begun by %s(), and VkRenderingInfo::pDepthAttachment does not define an "
+                                    "image view but VkCommandBufferInheritanceRenderingInfo::depthAttachmentFormat "
```
ncesario-lunarg, line 15916: Would be good to print out the value of (`depthAttachmentFormat`); same for the stencil case below.

ncesario-lunarg (commented):
A couple nits, but otherwise LGTM!

It would be good to have the "Changed dynamic rendering messages to stop using the VkRenderPass object as it is always VK_NULL_HANDLE" changes in their own commit.

layers/core_validation.cpp
paul-lunarg: Is there a way to print the name of the format ?

ncesario-lunarg: `string_VkFormat(...)`?

paul-lunarg: Done!

ncesario-lunarg (approved):

PR #4549, opened 2022-09-13
Fix fragment shading rate attachment with multiview

ziga-lunarg:
Layer count 1 is allowed for the fragment shading rate attachment according to the VUID


ncesario-lunarg (approved):
LGTM!

Tony-LunarG (approved):
LGTM

jeremyg-lunarg (approved):

PR #4554, opened 2022-09-15
Fix checking if zero initialize workgroup memory feature is enabled

ziga-lunarg:


layers/shader_validation.cpp
```diff
@@ -2478,7 +2478,7 @@ bool CoreChecks::ValidateVariables(const SHADER_MODULE_STATE &module_state) cons
         if (storage_class == spv::StorageClassWorkgroup) {
             // If Workgroup variable is initalized, make sure the feature is enabled
             if (insn.len() > 4 &&
-                !enabled_features.zero_initialize_work_group_memory_features.shaderZeroInitializeWorkgroupMemory) {
+                !enabled_features.core13.shaderZeroInitializeWorkgroupMemory) {
```
ncesario-lunarg, line 2481: Is there any need now for `zero_initialize_work_group_memory_features` in device_state.h? Looks like `core13.shaderZeroInitializeWorkgroupMemory` will get set whether using 1.3 or the extension (if I understand correctly).

ncesario-lunarg (approved):
One question, but LGTM!

layers/shader_validation.cpp
ziga-lunarg: You are correct.

Tony-LunarG (approved):
LGTM

PR #4558, opened 2022-09-15
threading: Use a better thread id

### LINKED ISSUE #4552: TID value on Linux is not optimal
While attempting to debug an issue with an Android application and running into `UNASSIGNED-Threading-MultipleThreads`, we realized that the value returned by `loader_platform_get_thread_id()` on Linux is a pointer as returned by `pthread_self()` which isn't particularly useful as debuggers don't use this value as the TID and you'd need to evaluate `pthread_self()` on every single thread to determine the faulty threads.

The real TID should be obtained by `gettid` syscall which should likely be implemented in a similar manner as [Bionic's `gettid` wrapper](https://android.googlesource.com/platform/bionic/+/a4edf7a57cf818d19546c86b74146335468ea53f/libc/bionic/gettid.cpp#33) to avoid constant syscall overhead here by storing it in TLS.

jeremyg-lunarg:
Use std::this_thread::get_id(), rather than pthread_self() as get_id() usually maps to a system wide thread id rather than a pthread_t pointer.

Fixes #4552

### DIFF
--- a/layers/generated/thread_safety.h
+++ b/layers/generated/thread_safety.h
@@ -76,7 +76,7 @@ public:
         int64_t count;
     };
 
-    ObjectUseData() : thread(0), writer_reader_count(0) {
+    ObjectUseData() : thread(), writer_reader_count(0) {
         // silence -Wunused-private-field warning
         padding[0] = 0;
     }
@@ -108,7 +108,7 @@ public:
         }
     }
 
-    std::atomic<loader_platform_thread_id> thread;
+    std::atomic<std::thread::id> thread;
 
 private:
     // need to update write and read counts atomically. Writer in high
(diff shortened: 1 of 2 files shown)


layers/generated/thread_safety.h
```diff
@@ -108,15 +108,15 @@ class ObjectUseData
         }
     }
 
-    std::atomic<loader_platform_thread_id> thread;
```
ncesario-lunarg, line 111: Should `loader_platform_thread_id` be removed entirely to avoid future use, or is this still required elsewhere?

ncesario-lunarg (approved):

layers/generated/thread_safety.h
jeremyg-lunarg: hmm. I would argue that ./layers/vk_loader_platform.h should be removed entirely. It is a copy of the same file from[ Vulkan-Loader ](https://github.com/KhronosGroup/Vulkan-Loader/blob/master/loader/vk_loader_platform.h) and I'm not sure if there is a real reason we need it.

ncesario-lunarg: Looks like that file could be trivially removed if not for the use of `loader_platform_open_library` when creating the dispatch table. That said, I think that is the _only_ bit of code that is referenced from that file, and we should still remove that file.

PR #4559, opened 2022-09-15
Rework VUID 02686

### LINKED ISSUE #4555: Input attachment descriptor image view is not a subpass input attachment
The recent patch `7acdcd6304b34aac41ef0b0e4bffaf8b7849aca4` revived input attachment validation. I found an example when checking gives false positive result and triggers error:

```
Input attachment descriptor image view is not a subpass input attachment
```

This happens after finishing render pass with input attachments. The timeline is following:

1. Command buffer start
1. Begin render pass with two subpasses. 5 attachments. 4 color render targets, 1 Depth.
1. Start first subpass
1. Rasterization drawcalls (geometry pass)
1. Start second subpass. 4 input attachments. 1 color render target.
1. Rasterization drawcalls (ligh pass)
1. End render pass
1. Start render pass with one subpass. 1 attachment. 1 color image from swapchain
1. Screen quad rasterization

Now issue happens here:

image

For some reason validation layers try to check render pass without input attachments and found candidate to fail. `subpass_input_views` set is empty.

Would you kindly help?

**Valid Usage ID**
```
2022-09-15 04:47:18.309 E/android_vulkan::C++: Renderer::OnVulkanDebugReport:
    category: VK_DEBUG_REPORT_ERROR_BIT_EXT
    object type: VK_DEBUG_REPORT_OBJECT_TYPE_IMAGE_VIEW_EXT
    object: 0x980b0000000002e
    location: 18446744073316843103
    message code: 0
    layer prefix: Validation
    message: Validation Error: [ UNASSIGNED-input-attachment-descriptor-not-in-subpass ] Object 0: handle = 0x980b0000000002e, type = VK_OBJECT_TYPE_IMAGE_VIEW; | MessageID = 0xe897be5f | Input attachment descriptor image view is not a subpass input attachment
```

**Environment:**
 - OS: _Android 11_
 - GPU: _Mali G76_
 - SDK or header version if building from repo: _SHA-1_ `92ed21f6a188b02dfc62a88c0ac40d5153429704`
 - Options enabled (synchronization, best practices, etc.): `VK_VALIDATION_FEATURE_ENABLE_BEST_PRACTICES_EXT` and `VK_VALIDATION_FEATURE_ENABLE_SYNCHRONIZATION_VALIDATION_EXT`

**Additional context**

Previous compatible version _SHA-1_ `bc7cacd509df1f3ac89949a586031d7105b27f97`.

paul-lunarg:
Breaks early if no input attachments in subpass.
Looks for input attachment descriptors in layout instead of descriptors, then accesses descriptor state correctly this time (I hope ?)
Closes #4555

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -1601,38 +1601,42 @@ bool CoreChecks::ValidateCmdBufDrawState(const CMD_BUFFER_STATE *cb_node, CMD_TY
 
             // Check if all input attachments are bound
             // VUID-vkCmdDraw-None-02686
-            std::unordered_map<VkImageView, uint32_t> subpass_input_views;
-            const auto &subpass = cb_node->activeRenderPass->createInfo.pSubpasses[cb_node->activeSubpass];
-            for (const auto &stage : pipe->stage_state) {
-                if (stage.stage_flag == VK_SHADER_STAGE_FRAGMENT_BIT) {
-                    // Get all subpass input attachment indices
-                    for (uint32_t i = 0; i < subpass.inputAttachmentCount; ++i) {
-                        auto index = subpass.pInputAttachments[i].attachment;
-                        if (index != VK_ATTACHMENT_UNUSED) {
-                            auto &view_states = *cb_node->active_attachments;
-                            if (view_states.size() > index && view_states[index] != nullptr) {
-                                VkImageView input_attachment_view = view_states[index]->image_view();
-                                subpass_input_views.insert(std::make_pair(input_attachment_view, i));
-                            }
+            {
+                const auto &subpass = cb_node->activeRenderPass->createInfo.pSubpasses[cb_node->activeSubpass];
+                std::unordered_map<VkImageView, uint32_t> subpass_input_views(subpass.inputAttachmentCount);
+                // Get all subpass input attachment indices
+                for (uint32_t i = 0; i < subpass.inputAttachmentCount; ++i) {
+                    auto index = subpass.pInputAttachments[i].attachment;
+                    if (index != VK_ATTACHMENT_UNUSED) {
+                        auto &view_states = *cb_node->active_attachments;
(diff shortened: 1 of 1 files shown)


lunarpapillo:
The Android failure appears to be spurious.  Re-running.

layers/core_validation.cpp
```diff
@@ -1619,21 +1619,25 @@ bool CoreChecks::ValidateCmdBufDrawState(const CMD_BUFFER_STATE *cb_node, CMD_TY
                         }
                     }
 
+                    if (subpass_input_views.empty()) {
+                        break;
```
sjfricke, line 1623: if breaking early please leave a comment otherwise when this funciton is altered in the future someone will see that `subpass_input_views` is not used until later and probably safe to move this where `subpass_input_views.count()` is currently called

paul-lunarg: Done

Tony-LunarG:
With this modification, performance is much better than without it, but there is still some degradation.
All traces slowed some, and 3 traces dropped below the 5% fail threshold at -6.16%, -5.51%, -5.22% and baselines would need to be reset.

There are still many new errors in CTS testing. Looking at 02686 it says that every input attachment used by the renderpass must be bound, but I don't see in the validation code where usage is determined. Is that determined elsewhere?  Could that be the cause of the new errors?

paul-lunarg:
The VUID just says the input attachment must be bound, so I list attachment views corresponding to input attachment indexing into the framebuffer and put them in a hashmap, and remove each input attachment descriptor bound by sets from the hashmap. In theory it should work but if for some reason a descriptor is missed it will trigger an error.
For performance I've optimized as much as I think I can, but it is a check that happens for every draw call. It only iterates on input attachments, though I now realize just the creation of the unordered map might be cause for a lot of delay... I should reduce it's size to the amount of input attachments.

ncesario-lunarg (commented):
If perf and CTS look good with is, LGTM.

PR #4560, opened 2022-09-15
layers: Remove vk_loader_platform.h

jeremyg-lunarg:
This file was duplicated from Vulkan-Loader and we were only using a small part of it. Here's what we were using and how it got resolved:

- dlopen wrappers: moved to lvt_function_pointers.cpp
- NOEXCEPT: the noexcept keyword is part of c++11 so just use it 
- API_NAME: was not used in the code
- PRINTF_SIZE_T_SPECIFIER: we are already using %zu for this in many places.


mark-lunarg (approved):
Sweet!

layers/generated/lvt_function_pointers.h
```diff
@@ -24,7 +24,18 @@
 
 
 #include <vulkan/vulkan.h>
-#include "vk_loader_platform.h"
+
+#ifdef _WIN32
+/* Windows-specific common code: */
+// WinBase.h defines CreateSemaphore and synchapi.h defines CreateEvent
+//  undefine them to avoid conflicts with VkLayerDispatchTable struct members.
```
ncesario-lunarg, line 31: I wish C++ modules were mainstream by now :(

ncesario-lunarg (approved):
Nice!

ziga-lunarg (approved):

PR #4561, opened 2022-09-15
v1.3.228 update

mikes-lunarg:
The header update was straightforward, but the VK_EXT_mutable_descriptor rename was larger than I expected so I put it in a separate commit.


Tony-LunarG (approved):
LGTM - Thanks for handling the mutable descriptor promotion.

ncesario-lunarg (approved):

PR #4562, opened 2022-09-16
layers: Calculate OpTypeBool as 32-bit ints for limits

sjfricke:
The 1.3.228 Vulkan Spec added a new clarification that `OpTypeBool` are the same size as 32-bit int for calculating the limits, prior they were being counted as `0` bytes

https://github.com/KhronosGroup/Vulkan-Docs/commit/355367640f2eabd2a0d492010a0afc166696aa72#diff-d753f04f744ff6c91e6b57979b21d920c15f1b59f955cff57d7cc942a0f96e6dR484-R485

### DIFF
--- a/tests/vklayertests_pipeline_shader.cpp
+++ b/tests/vklayertests_pipeline_shader.cpp
@@ -13909,6 +13909,35 @@ TEST_F(VkLayerTest, ComputeSharedMemoryOverLimit) {
     m_errorMonitor->VerifyFound();
 }
 
+TEST_F(VkLayerTest, ComputeSharedMemoryBooleanOverLimit) {
+    TEST_DESCRIPTION("Validate compute shader shared memory does not exceed maxComputeSharedMemorySize with booleans");
+
+    ASSERT_NO_FATAL_FAILURE(Init());
+
+    const uint32_t max_shared_memory_size = m_device->phy().properties().limits.maxComputeSharedMemorySize;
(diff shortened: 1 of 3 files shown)


tests/vklayertests_pipeline_shader.cpp
```diff
@@ -13909,6 +13909,35 @@ TEST_F(VkLayerTest, ComputeSharedMemoryOverLimit) {
...
+    const uint32_t max_shared_memory_size = m_device->phy().properties().limits.maxComputeSharedMemorySize;
+    // "Boolean values considered as 32-bit integer values for the purpose of this calculation."
+    const uint32_t max_shared_bools = max_shared_memory_size / 4;
+
+    std::stringstream csSource;
+    // Make sure compute pipeline has a compute shader stage set
+    csSource << R"glsl(
+        #version 450
+        shared bool a[)glsl";
+    csSource << (max_shared_bools + 16);
```
ncesario-lunarg, line 13926: I'm sure I'm missing something obvious here, but what is the significance of `16` (e.g., why is `1` not sufficient)?

ncesario-lunarg (approved):
LGTM!

tests/vklayertests_pipeline_shader.cpp
sjfricke: Because the other test used 16 😂

Issue #2756, opened 2021-04-29
vkQueueWaitIdle is not considered as a fence w.r.t. command buffer lifetime


hevrard:
**Describe the Issue**

The code conceptually does this:

```
vkQueueSubmit(myQueue, myCmdBuffer, ..., VkFence(0));
vkQueueWaitIdle(myQueue);
vkDestroyCommandPool(myCmdPool); // myCmdBuffer was allocated from myCmdPool
```

and I get this validation error:

```
04-29 15:20:52.624 11920 12185 I VALIDATION: VUID-vkDestroyCommandPool-commandPool-00041(ERROR / SPEC): msgNum: -1387836198 - Validation Error: [ VUID-vkDestroyCommandPool-commandPool-00041 ] Object 0: handle = 0x749b8bde70, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xad474cda | Attempt to destroy command pool with VkCommandBuffer 0x749b8bde70[] which is in use. The Vulkan spec states: All VkCommandBuffer objects allocated from commandPool must not be in the pending state (https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VUID-vkDestroyCommandPool-commandPool-00041)
```

By changing the code to the following, the error is no longer reported:

```
vkQueueSubmit(myQueue, myCmdBuffer, ..., myFence);
vkWaitForFences(myDevice, 1, &myFence, true, 1000000000);
vkDestroyCommandPool(myCmdPool); // myCmdBuffer was allocated from myCmdPool
```

The spec says:

> vkQueueWaitIdle is equivalent to submitting a fence to a queue and waiting with an infinite timeout for that fence to signal.

So I believe the error should not be reported with the original code.

**Valid Usage ID**

VUID-vkDestroyCommandPool-commandPool-00041

**Environment:**
 - OS: Android11
 - GPU: Adreno640
 - SDK or header version if building from repo: using current VVL master branch ToT (commit 5363130b75709e96e75cfecb75750c7ffe458b3e)
 - Options enabled (synchronization, best practices, etc.): default

jzulauf-lunarg:
@hevrard -- by code inspection, the two retire command buffers using the same underlying mechanism.  It looks like there's a bookkeeping bug re: what to retire.

Is there a repro case I could use to help me find it?

hevrard:
[...]

PR #4564, opened 2022-09-16
tests: Add VkPositiveLayerTest.SubmitFenceButWaitIdle

jeremyg-lunarg:
Test case for Issue #2756. I cannot get the error to reproduce, even running against the old version of VVL mentioned in the Issue. But it is still a good test to have.


Tony-LunarG (approved):
LGTM

ncesario-lunarg (approved):

PR #4569, opened 2022-09-19
Remove validated_descriptor_sets

paul-lunarg:
Remove validated_descriptor_sets as it was only used in one place and was not cleared correctly, causing a crash.
Replaced usage with lastBound[].per_set


jeremyg-lunarg:
I'm going to do a perf run on this change, just in case.

jeremyg-lunarg (approved):
LGTM. No changes to perf from this PR

PR #4570, opened 2022-09-19
Fix unexpected messages

paul-lunarg:
Fail the test on reception of unexpected message

### DIFF
--- a/tests/positive/other.cpp
+++ b/tests/positive/other.cpp
@@ -30,6 +30,7 @@
 #include <algorithm>
 #include <array>
 #include <chrono>
+#include <cstdlib>
 #include <memory>
 #include <mutex>
 #include <thread>
@@ -836,7 +837,23 @@ TEST_F(VkPositiveLayerTest, TestSwapchainImageFenceWait) {
 }
 
 TEST_F(VkPositiveLayerTest, EnumeratePhysicalDeviceGroups) {
-    TEST_DESCRIPTION("Test using VkPhysicalDevice handles obtained with vkEnumeratePhysicalDeviceGroups");
+    TEST_DESCRIPTION("Test using VkPhysicalDevice handles obtained with vkEnumeratePhysicalDeviceGroups"); 
+
+#ifdef __linux__
+    if (std::getenv("NODEVICE_SELECT") == nullptr)
+    {
+        // Currently due to a bug in MESA this test will fail.
+        // https://gitlab.freedesktop.org/mesa/mesa/-/commit/4588453815c58ec848b0ff6f18a08836e70f55df
+        //
+        // It's fixed as of v22.7.1:
+        // https://gitlab.freedesktop.org/mesa/mesa/-/tree/mesa-22.1.7/src/vulkan/device-select-layer
+        //
+        // To avoid impacting local users, skip this TEST unless NODEVICE_SELECT is specified.
+        // NODEVICE_SELECT enables/disables the implicit mesa layer which has illegal code:
+        // https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/vulkan/device-select-layer/VkLayer_MESA_device_select.json
+        GTEST_SKIP();
+    }
+#endif
 
     SetTargetApiVersion(VK_API_VERSION_1_1);
 
@@ -850,7 +867,7 @@ TEST_F(VkPositiveLayerTest, EnumeratePhysicalDeviceGroups) {
     uint32_t physical_device_group_count = 0;
     vk::EnumeratePhysicalDeviceGroups(test_instance, &physical_device_group_count, nullptr);
     std::vector<VkPhysicalDeviceGroupProperties> device_groups(physical_device_group_count,
-                                                               {VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_GROUP_PROPERTIES});
+                                                               LvlInitStruct<VkPhysicalDeviceGroupProperties>());
     vk::EnumeratePhysicalDeviceGroups(test_instance, &physical_device_group_count, device_groups.data());
 
     if (physical_device_group_count > 0) {
--- a/tests/vklayertests_best_practices.cpp
+++ b/tests/vklayertests_best_practices.cpp
@@ -1869,6 +1869,10 @@ TEST_F(VkBestPracticesLayerTest, LoadDeprecatedExtension) {
 
     const char *extension = VK_KHR_SYNCHRONIZATION_2_EXTENSION_NAME;
 
+    if (!DeviceExtensionSupported(extension)) {
+        GTEST_SKIP() << extension << " not supported.";
+    }
+
     VkDeviceQueueCreateInfo qci = LvlInitStruct<VkDeviceQueueCreateInfo>();
     qci.queueFamilyIndex = 0;
     float priority = 1;
--- a/tests/vklayertests_best_practices.cpp
+++ b/tests/vklayertests_best_practices.cpp
@@ -100,7 +100,6 @@ TEST_F(VkBestPracticesLayerTest, UseDeprecatedInstanceExtensions) {
     TEST_DESCRIPTION("Create an instance with a deprecated extension.");
 
     SetTargetApiVersion(VK_API_VERSION_1_1);
-    AddRequiredExtensions(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
 
     ASSERT_NO_FATAL_FAILURE(InitBestPracticesFramework());
 
@@ -111,6 +110,8 @@ TEST_F(VkBestPracticesLayerTest, UseDeprecatedInstanceExtensions) {
     if (!AreRequiredExtensionsEnabled()) {
         GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported";
     }
+
+    AddRequiredExtensions(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
 
     // Create a 1.1 vulkan instance and request an extension promoted to core in 1.1
     m_errorMonitor->SetDesiredFailureMsg(kWarningBit, "UNASSIGNED-BestPractices-vkCreateInstance-deprecated-extension");
--- a/tests/vklayertests_buffer_image_memory_sampler.cpp
+++ b/tests/vklayertests_buffer_image_memory_sampler.cpp
@@ -9249,30 +9249,14 @@ TEST_F(VkLayerTest, CreateImageMiscErrors) {
         image_ci.flags = VK_IMAGE_CREATE_CUBE_COMPATIBLE_BIT;
         image_ci.arrayLayers = 6;
         image_ci.imageType = VK_IMAGE_TYPE_1D;
-        m_errorMonitor->SetUnexpectedError("VUID-VkImageCreateInfo-imageType-00954");
         image_ci.extent = {64, 1, 1};
         CreateImageTest(*this, &image_ci, "VUID-VkImageCreateInfo-flags-00949");
 
         image_ci = safe_image_ci;
         image_ci.flags = VK_IMAGE_CREATE_CUBE_COMPATIBLE_BIT;
         image_ci.imageType = VK_IMAGE_TYPE_3D;
-        m_errorMonitor->SetUnexpectedError("VUID-VkImageCreateInfo-imageType-00954");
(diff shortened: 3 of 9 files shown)


ncesario-lunarg (dismissed):
LGTM, but we'll want to double check the CI to make sure there are not failing tests that are "passing."

juan-lunarg:
Without testing it myself repeatedly I'd say it's safer to keep on skipping GalaxyS10 for the `AndroidHardwareBufferInvalidBindBufferMemory` since it's completely unclear what Paul's intent was. Everything else looks fine and makes sense though.

ncesario-lunarg:
> Without testing it myself repeatedly I'd say it's safer to keep on skipping GalaxyS10 for the `AndroidHardwareBufferInvalidBindBufferMemory` since it's completely unclear what Paul's intent was. Everything else looks fine and makes sense though.

Sounds good to me. We should have an issue to track getting that test working again, though. It can either be a new issue or perhaps it would fit into one of the ones Bob already made (https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4587, https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4425)?

ncesario-lunarg:
> Added a fix to `DSUpdateOutOfBounds` since it had the same issue as `InUseDestroyedSignaled`
> 
> Except instead of specifying 32 I think it's more robust to use VK_WHOLE_SIZE instead:
> 
> ```
> When range is VK_WHOLE_SIZE the effective range is calculated at [vkUpdateDescriptorSets](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkUpdateDescriptorSets.html) is by taking the size of buffer minus the offset.
> ```

👍 as long as that keeps the test valid.

juan-lunarg:
> +1 as long as that keeps the test valid.

Still keeps the test valid. Just never mattered in practice since the test doesn't actually use the buffer in `DSUpdateOutOfBounds`, versus it does in `InUseDestroyedSignal`.

Still I just wanted to make sure the same pattern wasn't repeated.

juan-lunarg:
Now that unexpected messages give an error `vk::EnumeratePhysicalDeviceGroups` is failing.

When nullptr is passed in for the `VkPhysicalDeviceGroupProperties` pointer it still tries to validate the `sType` even though there is nothing to validate on the first pass.

```
[ RUN      ] VkPositiveLayerTest.EnumeratePhysicalDeviceGroups
Unexpected: Validation Error: [ VUID-VkPhysicalDeviceGroupProperties-sType-sType ] Object 0: VK_NULL_HANDLE, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0xc9edee8b | vkEnumeratePhysicalDeviceGroups: parameter pPhysicalDeviceGroupProperties[0].sType must be VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_GROUP_PROPERTIES The Vulkan spec states: sType must be VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_GROUP_PROPERTIES (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkPhysicalDeviceGroupProperties-sType-sType)
/home/lunarg/.jenkins/vz1/Debug64/Vulkan-ValidationLayers/tests/vkrenderframework.cpp:130: Failure
Failed
Received unexpected error.
Unexpected: Validation Error: [ VUID-VkPhysicalDeviceGroupProperties-sType-sType ] Object 0: VK_NULL_HANDLE, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0xc9edee8b | vkEnumeratePhysicalDeviceGroups: parameter pPhysicalDeviceGroupProperties[0].sType must be VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_GROUP_PROPERTIES The Vulkan spec states: sType must be VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_GROUP_PROPERTIES (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkPhysicalDeviceGroupProperties-sType-sType)
/home/lunarg/.jenkins/vz1/Debug64/Vulkan-ValidationLayers/tests/vkrenderframework.cpp:130: Failure
Failed
Received unexpected error.
```

juan-lunarg:
> Now that unexpected messages give an error `vk::EnumeratePhysicalDeviceGroups` is failing.

This was due to Mesa passing the incorrect sType in their code which resulted in an error for our code:
https://gitlab.freedesktop.org/mesa/mesa/-/commit/4588453815c58ec848b0ff6f18a08836e70f55df

tests/vkrenderframework.cpp
```diff
@@ -127,6 +127,10 @@ VkBool32 ErrorMonitor::CheckForDesiredMsg(const char *const msgString) {
         if (!found_expected) {
             result = VK_TRUE;
             printf("Unexpected: %s\n", msgString);
+            // TODO: Fix unexpected android failures
+#if !defined(ANDROID)
+            ADD_FAILURE() << "Received unexpected error.";
+#endif
```
juan-lunarg, line 133: Is it fine if to handle Android failures in another PR? Since this PR has already gotten pretty long, and I am not set up for it?

tests/positive/other.cpp
```diff
@@ -836,7 +837,18 @@ TEST_F(VkPositiveLayerTest, TestSwapchainImageFenceWait) {
 }
 
 TEST_F(VkPositiveLayerTest, EnumeratePhysicalDeviceGroups) {
-    TEST_DESCRIPTION("Test using VkPhysicalDevice handles obtained with vkEnumeratePhysicalDeviceGroups");
+    TEST_DESCRIPTION("Test using VkPhysicalDevice handles obtained with vkEnumeratePhysicalDeviceGroups"); 
+
+#ifdef __linux__
+    if (std::getenv("NODEVICE_SELECT") == nullptr)
```
juan-lunarg, line 843: @ncesario-lunarg like we discussed I'll make sure to get CI updated.

tests/vklayertests_buffer_image_memory_sampler.cpp
```diff
@@ -9249,30 +9249,14 @@ TEST_F(VkLayerTest, CreateImageMiscErrors) {
...
-        image_ci.extent = {8, 6, 1};
-        image_ci.arrayLayers = 6;
-        CreateImageTest(*this, &image_ci, "VUID-VkImageCreateInfo-imageType-00954");
-
-        image_ci = safe_image_ci;
-        image_ci.flags = VK_IMAGE_CREATE_CUBE_COMPATIBLE_BIT;
-        image_ci.imageType = VK_IMAGE_TYPE_2D;
-        image_ci.extent = {8, 8, 1};
-        image_ci.arrayLayers = 4;
-        CreateImageTest(*this, &image_ci, "VUID-VkImageCreateInfo-imageType-00954");
```
juan-lunarg, line 9275: Remove duplicate testing. See CreateImageMinLimitsViolation

tests/vklayertests_buffer_image_memory_sampler.cpp
```diff
@@ -9249,30 +9249,14 @@ TEST_F(VkLayerTest, CreateImageMiscErrors) {
         image_ci.flags = VK_IMAGE_CREATE_CUBE_COMPATIBLE_BIT;
         image_ci.arrayLayers = 6;
         image_ci.imageType = VK_IMAGE_TYPE_1D;
-        m_errorMonitor->SetUnexpectedError("VUID-VkImageCreateInfo-imageType-00954");
```
juan-lunarg, line 9252: These can be safely removed now that the image type is properly checked now.

tests/positive/other.cpp
juan-lunarg: I figured out the exact release this bug got fixed in.

It's been fixed as of [22.7.1](https://gitlab.freedesktop.org/mesa/mesa/-/tree/mesa-22.1.7/src/vulkan/device-select-layer)

ncesario-lunarg: @lunarpapillo any chance we could get a "mesa driver update exercise" scheduled in the near future?

tests/vkrenderframework.cpp
juan-lunarg: Considering Windows/Linux builds are now fixed, I'd say this PR has still achieved quite a bit.

juan-lunarg: Agreed. I plan to make an issue listing them after the PR is merged.

```
GalaxyS10
[  FAILED  ] VkBestPracticesLayerTest.UseDeprecatedInstanceExtensions

Pixel6
[  FAILED  ] VkLayerTest.AndroidHardwareBufferInvalidBindBufferMemory
[  FAILED  ] VkBestPracticesLayerTest.UseDeprecatedInstanceExtensions
[  FAILED  ] VkPositiveLayerTest.AndroidHardwareBufferMemoryRequirements
[  FAILED  ] VkPositiveLayerTest.AndroidHardwareBufferBindBufferMemory

Pixel3:
[  FAILED  ] VkLayerTest.AndroidHardwareBufferInvalidBindBufferMemory
[  FAILED  ] VkBestPracticesLayerTest.UseDeprecatedInstanceExtensions
[  FAILED  ] VkPositiveLayerTest.AndroidHardwareBufferDepthStencil

SheildTv
[  FAILED  ] VkLayerTest.AndroidHardwareBufferInvalidBindBufferMemory
[  FAILED  ] VkBestPracticesLayerTest.UseDeprecatedInstanceExtensions
[  FAILED  ] VkPositiveLayerTest.AndroidHardwareBufferDepthStencil
```

tests/layer_validation_tests.cpp
```diff
@@ -575,6 +575,13 @@ void CreateBufferTest(VkLayerTest &test, const VkBufferCreateInfo *create_info,
 }
 
 void CreateImageTest(VkLayerTest &test, const VkImageCreateInfo *create_info, std::string code) {
+    VkImageFormatProperties props{};
+    VkResult result =
+        vk::GetPhysicalDeviceImageFormatProperties(test.gpu(), create_info->format, create_info->imageType, create_info->tiling,
+                                                   create_info->usage, create_info->flags, &props);
+    if (result == VK_ERROR_FORMAT_NOT_SUPPORTED) {
+        return;
```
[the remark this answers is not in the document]
juan-lunarg: After looking at every reference of `CreateImageTest`, the only thing this function is used for is checking the `VkImageCreateInfo` struct. In which case the check for unsupported format can actually reduce the code written by the caller.

What about renaming the function `CreateImageIfSupportedTest`?

tests/vklayertests_best_practices.cpp
```diff
@@ -1868,6 +1869,21 @@ TEST_F(VkBestPracticesLayerTest, LoadDeprecatedExtension) {
...
+
+    bool found = false;
+    for (uint32_t i = 0; i < count; ++i) {
+        if (strcmp(props[i].extensionName, extension) == 0) {
+            found = true;
+        }
+    }
+    if (!found) {
+        GTEST_SKIP() << extension << " not supported.";
+    }
```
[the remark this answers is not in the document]
juan-lunarg: Good call I'll fix that.

tests/layer_validation_tests.cpp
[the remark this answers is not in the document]
ncesario-lunarg: Yes, if changing the behavior of the function, it should also be renamed.

My only reservation is concerning "positive" tests (i.e., CreateImage should pass), where the test is never even run due to the format not being supported, and the author not actually realizing their code is not being tested. This may be an unreasonable concern, though.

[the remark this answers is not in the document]
juan-lunarg: Thinking about it again, CreateImageMiscErrors passes locally on my machine now. And I did end up making actual fixes to the code besides that. Maybe this early return isn't needed at all. I'm pushing up a change.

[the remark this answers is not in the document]
juan-lunarg: Currently testing PR without this change to `CreateImageTest` since it may not be needed at all, considering my other fixes to `CreateImageMiscErrors`.

juan-lunarg:
> CI Vulkan-ValidationLayers build # 9055 failed.

I checked to make sure my change to `CreateImageMiscErrors` didn't cause any regressions.

The failures is due to `VkPositiveLayerTest.GetTimelineSemThreadRace` which is known to be flaky, and the S10 timing out. Both of which appear to be unrelated.

I'll re-run since it appears to be unrelated.

PR #4571, opened 2022-09-19
 layers: Fix CB locking during query callback functions

### LINKED ISSUE #4243: VkLayerTest.QueryPerformanceReleaseProfileLockBeforeSubmit test hangs
**Describe the Issue**
When running `VkLayerTest.QueryPerformanceReleaseProfileLockBeforeSubmit` it is hanging in `CoreChecks::ValidatePerformanceQuery` at 

```c++
auto cb_state = state_data->GetRead<CMD_BUFFER_STATE>(commandBuffer);
```

Not familiar with the performance query part of the code, but I always have to disable this test when trying to run the full suite of tests locally

**Valid Usage ID**

The tests aims to test `VUID-vkQueueSubmit-pCommandBuffers-03220`

**Environment:**
 - OS: Ubuntu 20.04
 - GPU: Intel Xe Graphics
 - SDK or header version if building from repo: ToT (but also for the last 2 months I have tried as well)

jeremyg-lunarg:
There were deadlocks and race conditions when running the callbacks in CMD_BUFFER_STATE::queryUpdates. This was because secondary command buffer callbacks were copied into the primary's callback list. Then the callbacks were all called with the primary buffer's lock locked, but not the secondary. Instead, make a callback that loops through the secondary's callbacks in place rather than copying. Then the callback can safely lock the secondary.

Hopefully this is only an interim fix and we can find a better way to implement this without callback lambda voodoo.

Fixes #4243

This PR also un-deletes and updates the tests, and fixes another problem with how perf passes were tracked (see individual commit messages).

### DIFF
--- a/layers/cmd_buffer_state.cpp
+++ b/layers/cmd_buffer_state.cpp
@@ -562,6 +562,13 @@ void CMD_BUFFER_STATE::EndQuery(const QueryObject &query_obj) {
     updatedQueries.insert(query_obj);
 }
 
+bool CMD_BUFFER_STATE::UpdatesQuery(const QueryObject &query_obj) const {
+    // Clear out the perf_pass from the caller because it isn't known when the command buffer is recorded.
+    auto key = query_obj;
+    key.perf_pass = 0;
+    return updatedQueries.find(key) != updatedQueries.end();
+}
+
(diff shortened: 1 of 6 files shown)


sjfricke:
can confirm `--gtest_filter=*QueryPerformance*` works for me now with these changes

ziga-lunarg (approved):
LGTM

ncesario-lunarg (approved):

PR #4563, opened 2022-09-16
Fix checking pipelineCreationCacheControl feature

ziga-lunarg:

PR #4573, opened 2022-09-19
Validate the attachment_feedback_loop_layout feature

Tony-LunarG:
New VUIDs were added, requiring the feature to use the layout.

### DIFF
--- a/layers/buffer_validation.cpp
+++ b/layers/buffer_validation.cpp
@@ -5610,6 +5610,17 @@ bool CoreChecks::ValidateImageBarrier(const LogObjectList &objects, const Locati
                          layout_loc.Message().c_str());
         }
     }
+    
+    if (mem_barrier.newLayout == VK_IMAGE_LAYOUT_ATTACHMENT_FEEDBACK_LOOP_OPTIMAL_EXT) {
+        if (!enabled_features.attachment_feedback_loop_layout_features.attachmentFeedbackLoopLayout) {
+            auto layout_loc = loc.dot(Field::newLayout);
+            const auto &vuid = sync_vuid_maps::GetImageBarrierVUID(loc, sync_vuid_maps::ImageError::kBadAttFeedbackLoopLayout);
+            skip |= LogError(cb_state->commandBuffer(), vuid,
+                             "%s Image Layout cannot be transitioned to VK_IMAGE_LAYOUT_ATTACHMENT_FEEDBACK_LOOP_OPTIMAL_EXT if "
+                             "the attachmentFeedbackLoopLayout feature is not enabled",
+                             layout_loc.Message().c_str());
+        }
+    }
 
     auto image_data = Get<IMAGE_STATE>(mem_barrier.image);
     if (image_data) {
--- a/layers/stateless_validation.h
+++ b/layers/stateless_validation.h
@@ -1025,6 +1025,12 @@ class StatelessValidation : public ValidationObject {
                 separate_depth_stencil_layouts = separate_depth_stencil_layouts_features->separateDepthStencilLayouts;
         }
 
+        VkBool32 attachment_feedback_loop_layout = false;
+        const auto *attachment_feedback_loop_layout_features =
+                LvlFindInChain<VkPhysicalDeviceAttachmentFeedbackLoopLayoutFeaturesEXT>(device_createinfo_pnext);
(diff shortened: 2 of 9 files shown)


tests/vklayertests_descriptor_renderpass_framebuffer.cpp
```diff
@@ -2181,11 +2181,16 @@ TEST_F(VkLayerTest, RenderPassBeginLayoutsFramebufferImageUsageMismatches) {
...
 
-    auto attachment_feedback_loop_layout_features =
-        LvlInitStruct<VkPhysicalDeviceAttachmentFeedbackLoopLayoutFeaturesEXT>();
-    auto features2 = LvlInitStruct<VkPhysicalDeviceFeatures2>(&attachment_feedback_loop_layout_features);
-    GetPhysicalDeviceFeatures2(features2);
-    ASSERT_NO_FATAL_FAILURE(InitState(nullptr, &features2, VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT));
+    if (feedback_loop_layout) {
+        auto attachment_feedback_loop_layout_features = LvlInitStruct<VkPhysicalDeviceAttachmentFeedbackLoopLayoutFeaturesEXT>();
+        auto features2 = LvlInitStruct<VkPhysicalDeviceFeatures2>(&attachment_feedback_loop_layout_features);
+        GetPhysicalDeviceFeatures2(features2);
```
ncesario-lunarg, line 2187: You can collapse this into `auto features2 = GetPhysicalDeviceFeatures2(attachment_feedback_loop_layout_features);` if desired.

ncesario-lunarg (approved):
LGTM!

layers/buffer_validation.cpp
```diff
@@ -5611,6 +5611,17 @@ bool CoreChecks::ValidateImageBarrier(const LogObjectList &objects, const Locati
         }
     }
 
+    if (!enabled_features.attachment_feedback_loop_layout_features.attachmentFeedbackLoopLayout) {
+        if (mem_barrier.newLayout == VK_IMAGE_LAYOUT_ATTACHMENT_FEEDBACK_LOOP_OPTIMAL_EXT) {
```
sjfricke, line 5615: nit: I expect layout to not be `VK_IMAGE_LAYOUT_ATTACHMENT_FEEDBACK_LOOP_OPTIMAL_EXT` 99.9% of the time for apps, and I think for other "the feature enables this random enum value" we flip the `if` statements around

```suggestion
    if (mem_barrier.newLayout == VK_IMAGE_LAYOUT_ATTACHMENT_FEEDBACK_LOOP_OPTIMAL_EXT) {
        if (!enabled_features.attachment_feedback_loop_layout_features.attachmentFeedbackLoopLayout) {
```

Tony-LunarG: Good point - fixed

tests/vklayertests_descriptor_renderpass_framebuffer.cpp
Tony-LunarG: Done.  Now if I can just remember that for next time....

layers/stateless_validation.h
```diff
@@ -1025,6 +1025,12 @@ class StatelessValidation : public ValidationObject {
                 separate_depth_stencil_layouts = separate_depth_stencil_layouts_features->separateDepthStencilLayouts;
         }
 
+        VkBool32 attachment_feedback_loop_layout = false;
+        const auto *attachment_feedback_loop_layout_features =
+                LvlFindInChain<VkPhysicalDeviceAttachmentFeedbackLoopLayoutFeaturesEXT>(device_createinfo_pnext);
```
jeremyg-lunarg, line 1030: Between this and https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4563 it sure seems like it would be nice to have the enabled_features struct available to stateless.

jeremyg-lunarg (approved):
LGTM

layers/stateless_validation.h
ncesario-lunarg: Yeah, I've been tempted in the past. The amount of "state tracking" in "stateless" is starting to grow, but maybe that's not a bad thing; just need some guidelines as to what "should" and "should not" be tracked there (since the name kind of implies _nothing_ should be tracked...).

PR #4574, opened 2022-09-20
layers: Label RP1 version of Attachment Ref VU

sjfricke:
labels `VUID-VkAttachmentReference2-layout-03077` and `VUID-VkAttachmentReference2-synchronization2-06910`

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -215,7 +215,7 @@ void TestRenderPassCreate(ErrorMonitor *error_monitor, const VkDevice device, co
         error_monitor->VerifyFound();
     }
 
-    if (rp2_supported && nullptr == rp2_vuid) {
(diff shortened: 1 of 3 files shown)


tests/layer_validation_tests.cpp
```diff
@@ -215,7 +215,7 @@ void TestRenderPassCreate(ErrorMonitor *error_monitor, const VkDevice device, co
         error_monitor->VerifyFound();
     }
 
-    if (rp2_supported && nullptr == rp2_vuid) {
+    if (rp2_supported && nullptr != rp2_vuid) {
```
sjfricke, line 218: unless I am missing something, this seems to have been wrong

ncesario-lunarg: That would be my assumption as well.

ncesario-lunarg (approved):
LGTM!

PR #4575, opened 2022-09-20
Track physical device handles coming from groups

arno-lunarg:
Physical device handles obtained from `vkEnumeratePhysicalDeviceGroups` were not tracked, thus validation did not know about them and would throw errors when using them.

The tracking code was not generated because the C struct `VkPhysicalDeviceGroupProperties` is not an handle (but a set of handles), and from my understanding object tracking code is generated only for handles. So I hacked in object tracking generation for this struct in the code generation script.

As a note, this issue may not pop when using old version of the Vulkan loader. The old version of `vkEnumeratePhysicalDeviceGroups` would call `vkEnumeratePhysicalDevices`, a call that would get validated, and cause physical devices handles to be tracked.


jeremyg-lunarg (approved):
LGTM

scripts/object_tracker_generator.py
```diff
@@ -744,6 +745,8 @@ def generate_create_object_code(self, indent, proto, params, cmd_info, allocator
                 indent = self.decIndent(indent)
                 create_obj_code += '%s}\n' % indent
             indent = self.decIndent(indent)
+        elif handle_type.text == 'VkPhysicalDeviceGroupProperties':
```
ncesario-lunarg, line 748: Might be good to add a comment as to why this is needed here.

scripts/object_tracker_generator.py
```diff
@@ -147,6 +147,7 @@ def __init__(self,
             'vkCreateInstance',
             'vkCreateDevice',
             'vkEnumeratePhysicalDevices',
+            'vkEnumeratePhysicalDeviceGroups',
```
ncesario-lunarg, line 150: Can this be removed to keep `PreCallValidateEnumeratePhysicalDeviceGroups` auto-generated? Looks like the auto-generated and manually written functions are the same. That would also keep things consistent with `PreCallValidateEnumeratePhysicalDeviceGroupsKHR`.

ncesario-lunarg (commented):
A couple questions, but otherwise LGTM!

scripts/object_tracker_generator.py
arno-lunarg: I did this because otherwise `PostCallRecordEnumeratePhysicalDeviceGroups` body would end up in generated code, when I wanted it to be in `object_tracker_utils.cpp`.
But thinking about it your call is the right one, so I move the code generation for this function in the script. It also now relies on `PostCallRecordEnumeratePhysicalDevices` to record the handles. Doing so seems legitimate since device groups are, well, just device groups!

tests/positive/other.cpp
```diff
@@ -834,3 +834,29 @@ TEST_F(VkPositiveLayerTest, TestSwapchainImageFenceWait) {
 
     vk::QueueWaitIdle(m_device->m_queue);
 }
+
+TEST_F(VkPositiveLayerTest, EnumeratePhysicalDeviceGroups) {
+    TEST_DESCRIPTION("Test using VkPhysicalDevice handles obtained with vkEnumeratePhysicalDeviceGroups");
```
Tony-LunarG, line 839: This test passes even if it gets errors because the errors are not going to the callback, although it's not immediately clear to me why or how to fix it. Maybe a debug utils messenger via the debug utils extension?

ncesario-lunarg: Is this related to https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4570?

scripts/object_tracker_generator.py
```diff
@@ -744,6 +745,13 @@ def generate_create_object_code(self, indent, proto, params, cmd_info, allocator
                 indent = self.decIndent(indent)
                 create_obj_code += '%s}\n' % indent
             indent = self.decIndent(indent)
+        # Physical device groups are not handles, but a set of handles, they need to be tracked as well
+        elif handle_type.text == 'VkPhysicalDeviceGroupProperties':
+            create_obj_code += '%sif (%s) {\n' % (indent, cmd_info[-1].name)
```
ncesario-lunarg, line 750: This is largely subjective, but I find it way easier to read "f-strings" rather than printf style style strings for this. e.g., `
```python
create_obj_code += f'''{indent}if ({cmd_info[-1].name}) {
{indent}{indent}for (... ;device_group_index < *{cmd_info[-2].name}; ...) {
...
'''`
```

ncesario-lunarg (approved):

tests/positive/other.cpp
Tony-LunarG: Not exactly, or at least #4570 doesn't fix the problem

scripts/object_tracker_generator.py
arno-lunarg: I absolutely did not know about that, this is cool!

tests/positive/other.cpp
arno-lunarg: I am not sure I understand what to do here... Do I need to add a `DebugReporter` object, like this for instance:
```cpp
    VkInstance test_instance = VK_NULL_HANDLE;
    ASSERT_VK_SUCCESS(vk::CreateInstance(&ici, nullptr, &test_instance));
    DebugReporter debug_reporter;
    debug_reporter.Create(test_instance);
```
?

arno-lunarg: It seems that doing this is not enough, when forgetting to call `debug_reporter.Destroy(test_instance)`, I indeed get a validation error regarding this but the test still passes...

PR #4576, opened 2022-09-20
Add check to avoid crash in CTS

paul-lunarg:
Since it is not technically illegal to not bind as many sets as set by the pipeline layout, need to check that we are not reading more

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -1621,7 +1621,7 @@ bool CoreChecks::ValidateCmdBufDrawState(const CMD_BUFFER_STATE *cb_node, CMD_TY
                     // Remove input attachment views that are bound
                     // Look into pipeline layout for input attachments
                     const auto &set_layouts = pipe->PipelineLayoutState()->set_layouts;
-                    for (uint32_t set_index = 0; set_index < set_layouts.size(); ++set_index) {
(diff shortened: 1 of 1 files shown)


layers/core_validation.cpp
```diff
@@ -1626,6 +1626,7 @@ bool CoreChecks::ValidateCmdBufDrawState(const CMD_BUFFER_STATE *cb_node, CMD_TY
                         for (const auto &binding : set_layout->GetBindings()) {
                             if (binding.descriptorType == VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT) {
                                 // Layout binding is input so get descriptors at that set/binding
+                                if (state.per_set.size() <= set_index) continue;
```
Tony-LunarG, line 1629: state.per_set.size() is loop invariant in both for loops, and set_index is only increasing, so wouldn't you want to break out of the set_index for loop instead of continuing in the binding for loop? Or use a compound condition in the set_index for loop to compare set_index to set_layouts.size() and state.per_set.size() ?

ncesario-lunarg (commented):
LGTM if all of @Tony-LunarG's concerns are resolved.

layers/core_validation.cpp
paul-lunarg: You're right! Thanks

Tony-LunarG (approved):
LGTM - thanks for the quick fix

PR #4578, opened 2022-09-20
Reduce long term memory use growth for SyncVal with submit time validation enabled

jzulauf-lunarg:
Reorganize access log storage s.t. unreferenced command buffer access logs can be discared.

Needed

* detailed review of the BatchLog trimming algorithm
* detailed review of sparse_container::consolidate  (which seemed way simpler than I had thought it would be)

### DIFF
--- a/layers/range_vector.h
+++ b/layers/range_vector.h
@@ -1824,6 +1824,45 @@ bool update_range_value(Map &map, const Range &range, MapValue &&value, value_pr
     return updated;
 }
 
+//  combines directly adjacent ranges with equal RangeMap::mapped_type .
+template <typename RangeMap>
+void consolidate(RangeMap &map) {
+    using Value = typename RangeMap::value_type;
+    using Key = typename RangeMap::key_type;
+    using It = typename RangeMap::iterator;
+
+    It current = map.begin();
+    const It map_end = map.end();
+
+    // To be included in a merge range there must be no gap in the Key space, and the mapped_type values must match
+    auto can_merge = [](const It &last, const It &cur) {
+        return cur->first.begin == last->first.end && cur->second == last->second;
+    };
+
+    while (current != map_end) {
+        // Establish a trival merge range at the current location, advancing current. Merge range is inclusive of merge_last
+        const It merge_first = current;
+        It merge_last = current;
+        ++current;
+
+        // Expand the merge range as much as possible
+        while (current != map_end && can_merge(merge_last, current)) {
+            merge_last = current;
+            ++current;
+        }
+
+        // Current isn't in the active merge range. If there is a non-trivial merge range, we resolve it here.
+        if (merge_first != merge_last) {
+            // IFF there is more than one range in (merge_first, merge_last)  <- again noting the *inclusive* last
+            // Create a new Val spanning (first, last), substitute it for the multiple entries.
+            Value merged_value = std::make_pair(Key(merge_first->first.begin, merge_last->first.end), merge_last->second);
+            // Note that current points to merge_last + 1, and is valid even if at map_end for these operations
+            map.erase(merge_first, current);
+            map.insert(current, std::move(merged_value));
+        }
+    }
+}
+
 }  // namespace sparse_container
 
 #endif
--- a/layers/synchronization_validation.cpp
+++ b/layers/synchronization_validation.cpp
@@ -3876,6 +3876,7 @@ void SyncValidator::ApplyTaggedWait(QueueId queue_id, ResourceUsageTag tag) {
     QueueBatchContext::BatchSet queue_batch_contexts = GetQueueBatchSnapshot();
     for (auto &batch : queue_batch_contexts) {
         batch->ApplyTaggedWait(queue_id, tag);
+        batch->Trim();
     }
 }
 
@@ -7650,6 +7651,7 @@ void QueueBatchContext::ApplyTaggedWait(QueueId queue_id, ResourceUsageTag tag) 
 void QueueBatchContext::ApplyDeviceWait() {
     access_context_.Reset();
     events_context_.ApplyTaggedWait(GetQueueFlags(), ResourceUsageRecord::kMaxIndex);
+    Trim();
 }
 
 HazardResult QueueBatchContext::DetectFirstUseHazard(const ResourceUsageRange &tag_range) {
--- a/layers/synchronization_validation.cpp
+++ b/layers/synchronization_validation.cpp
@@ -841,12 +841,10 @@ void AccessContext::Trim() {
     auto normalize = [](AccessAddressType address_type, ResourceAccessRangeMap::value_type &access) { access.second.Normalize(); };
     ForAll(normalize);
 
-    // TODO consolidate map after normalization
-#if 0
+    // Consolidate map after normalization, combines directly adjacent ranges with common values.
     for (auto& map : access_state_maps_) {
-        map.consolidate();
+        sparse_container::consolidate(map);
     }
-#endif
 }
 
 void AccessContext::AddReferencedTags(ResourceUsageTagSet &used) const {
--- a/layers/synchronization_validation.cpp
+++ b/layers/synchronization_validation.cpp
@@ -287,7 +287,7 @@ CommandBufferAccessContext::CommandBufferAccessContext(const CommandBufferAccess
     cb_state_ = from.cb_state_;
     queue_flags_ = from.queue_flags_;
     destroyed_ = from.destroyed_;
-    access_log_ = from.access_log_;  // potentially large, but no choice given tagging lookup.
+    access_log_ = std::make_shared<AccessLog>(*from.access_log_);  // potentially large, but no choice given tagging lookup.
     command_number_ = from.command_number_;
     subcommand_number_ = from.subcommand_number_;
     reset_count_ = from.reset_count_;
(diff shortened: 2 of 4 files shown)


layers/synchronization_validation.cpp
```diff
@@ -7882,3 +7912,70 @@ SubmitInfoConverter::SubmitInfoConverter(uint32_t count, const VkSubmitInfo *inf
...
+    return tag_limit;
+}
+
+void BatchAccessLog::Import(const BatchAccessLog &other) {
+    for (const auto &entry : other.log_map_) {
+        log_map_.insert(entry);
+    }
+}
+
+void BatchAccessLog::Trim(const ResourceUsageTagSet &used_tags) {
```
jzulauf-lunarg, line 7930: This funky code here is a bit of a "your in a maze of twisty little passages, all different" parallel iteration which attempts to remove all ranges (which correspond to the access log of one command buffer recording) if none of the tags are referenced by set of tags passed in.

Clearly I need to boil some of ^^^ into a good comment (things you'd like to see mentioned, please include in your comments.

@jeremyg-lunarg  @ncesario-lunarg @Tony-LunarG

jzulauf-lunarg: @ncesario-lunarg -- yes.  The range_map contains shared pointers to the access logs, such that once no `QueueBatchContext::access_context_` references any tag in a range, the underlying access logs will destruct. 

It's another example of the "parallel iteration" pattern that populate sync val.  However in this case we are iterating dissimilar data -- a set of tags, and a map keyed by tag ranges, thus the novel construction.

ncesario-lunarg: A comment similar to ^ (what you just said) seems like it would be good to me :)

layers/synchronization_validation.cpp
```diff
@@ -7882,3 +7933,85 @@ SubmitInfoConverter::SubmitInfoConverter(uint32_t count, const VkSubmitInfo *inf
...
+        log_map_.insert(entry);
+    }
+}
+
+// In order to contain memory growth in the AccessLog information regarding prior submitted command buffers
+// the Trim call removes any AccessLog references that do not correspond to any tags in use (the input list being generated from
+// the AccessContext and EventContext of the parent QueueBatchContext).
+//
+// The approach here is another example of the "parallel iteration" pattern common within SyncVal.  In this case we are
+// traversing the ordered range_map containing the AccessLog references and the order set of "tags in use. The traversal is
```
ncesario-lunarg, line 7956: Missing `"`?

layers/synchronization_validation.cpp
```diff
@@ -7882,3 +7933,85 @@ SubmitInfoConverter::SubmitInfoConverter(uint32_t count, const VkSubmitInfo *inf
...
+// optimized in the following ways:
```
ncesario-lunarg, line 7957: Just to clarify, the purpose of this function is to optimize all future traversals _after this function returns_, and this comment is referring to how the traversal in this particular function is optimized?

layers/synchronization_validation.cpp
```diff
@@ -7882,3 +7933,85 @@ SubmitInfoConverter::SubmitInfoConverter(uint32_t count, const VkSubmitInfo *inf
...
+//  * when ranges are detected that have no tags reference, all ranges between the last tag and the current tag are erased
```
ncesario-lunarg, line 7958: "tags referenced"? "tag referenced"?

jzulauf-lunarg:
@jeremyg-lunarg -- new coded to review "consolidate" -- which seems simple enough I'm worried I'm missing something. ;)

layers/synchronization_validation.cpp
jeremyg-lunarg: would this functionality be of use to the other uses of range maps? Image Layout and I think core bound memory tracking.  At a glance it looks like some sort of set intersection operation.

layers/range_vector.h
```diff
@@ -1824,6 +1824,65 @@ bool update_range_value(Map &map, const Range &range, MapValue &&value, value_pr
     return updated;
 }
 
+//  combines directly adjacent ranges with equal RangeMap::mapped_type .
+template <typename RangeMap>
+void consolidate(RangeMap &map) {
```
jeremyg-lunarg, line 1829: Do you think we should be doing this to the global layout range maps after they are updated in CoreChecks::UpdateCmdBufImageLayouts()?

layers/range_vector.h
```diff
@@ -1824,6 +1824,65 @@ bool update_range_value(Map &map, const Range &range, MapValue &&value, value_pr
...
+    if (map.empty()) return;
+
+    It current = map.begin();
+    const It map_end = map.end();
+    // first and last are *inclusive* of the ranges to be merged
+    It merge_first;
+    It merge_last;
+
+    // Define two common operations the happen in and around the while
+    auto new_merge_range = [&current, &merge_first, &merge_last]() {
```
jeremyg-lunarg, line 1843: couldn't `new_merge_range()`; be replaced with `merge_first = merge_last = current++;`  To me it seems like that would be clearer, for understanding how the loop works.

layers/range_vector.h
```diff
@@ -1824,6 +1824,65 @@ bool update_range_value(Map &map, const Range &range, MapValue &&value, value_pr
...
+        }
+    };
+
+    // Set up the first trivial merge range -- for which merge is a noop, with current pointing after it
+    new_merge_range();
+    while (current != map_end) {
+        // To be included in a merge range there must be no gap in the Key space, and the mapped_type values must match
+        if ((current->first.begin == merge_last->first.end) && (current->second == merge_first->second)) {
+            // Expand the merge range, but don't merge until current isn't included.
+            merge_last = current;
```
jeremyg-lunarg, line 1873: I'd be tempted to rewrite this as 2 loops:
```
     auto can_merge = [] (It first, It last, It cur) {   //lamdas with side effects hurt my noggin
           return cur->first.begin == last->first.end && cur->second == first->second;
      }
     while (current != map_end) {
            merge_first = merge_last = current++;
            while (current != map_end && can_merge(merge_first, merge_last, current)) {
                  merge_last = current++; 
            }
            merge();  // but just do it inline since now it only needs to happen once.
    }
```

layers/synchronization_validation.cpp
```diff
@@ -3453,6 +3470,20 @@ void ResourceAccessState::ClearRead() {
     last_read_stages = VK_PIPELINE_STAGE_2_NONE;
 }
 
+void ResourceAccessState::ClearPending() {
+    pending_write_dep_chain = VK_PIPELINE_STAGE_2_NONE;
+    pending_layout_transition = false;
+    ;
```
jeremyg-lunarg, line 3476: semi-colon only line?

layers/synchronization_validation.cpp
```diff
@@ -7882,3 +7931,85 @@ SubmitInfoConverter::SubmitInfoConverter(uint32_t count, const VkSubmitInfo *inf
...
+    auto current_tag = used_tags.cbegin();
+    const auto end_tag = used_tags.cend();
+    auto current_range = log_map_.begin();
+    const auto end_map = log_map_.end();
+
+    while (current_range != end_map) {
+        if (current_tag == end_tag) {
+            // We're out of tags, the rest of the map isn't referenced, so erase it
+            current_range = log_map_.erase(current_range, end_map);
+            assert(current_range == end_map);
```
jeremyg-lunarg, line 7974: should this assert be in range_map<>::erase()? If it was failing the bug would be in that code

layers/synchronization_validation.cpp
```diff
@@ -7882,3 +7931,85 @@ SubmitInfoConverter::SubmitInfoConverter(uint32_t count, const VkSubmitInfo *inf
...
+            if (tag < range.begin) {
+                // Skip to the next tag potentially in range
+                // if this is end_tag, we'll handle that next iteration
+                current_tag = used_tags.lower_bound(range.begin);
+            } else if (tag >= range.end) {
+                // This tag is beyond the current range, delete all ranges between current_range,
+                // and the next that includes the tag.  Next is not erased.
+                auto next_used = log_map_.lower_bound(ResourceUsageRange(tag, tag + 1));
+                current_range = log_map_.erase(current_range, next_used);
+                assert(current_range == next_used);
```
jeremyg-lunarg, line 7987: same assert.

layers/synchronization_validation.cpp
```diff
@@ -7882,3 +7931,85 @@ SubmitInfoConverter::SubmitInfoConverter(uint32_t count, const VkSubmitInfo *inf
...
+//  * when a tag is found within the range, that range is skipped (and thus kept in the map), and further used tags within the range
+//    are skipped.
+//
+// Note that for each subcase, any "next steps" logic is designed to be handled within the subsequent iteration -- meaning that
+// each subcase simply handles the specific of the current update/skip/erase action needed, and leave the iterators in a sensible
+// state for the top of loop... intentionally eliding special case handling.
+void BatchAccessLog::Trim(const ResourceUsageTagSet &used_tags) {
+    auto current_tag = used_tags.cbegin();
+    const auto end_tag = used_tags.cend();
+    auto current_range = log_map_.begin();
```
jeremyg-lunarg, line 7967: maybe name this current_map_range, so that it is clear that it is part of the map stream and not the tag stream. For better stream crossing.

layers/range_vector.h
jzulauf-lunarg: it could, though stylistically I've never liked the a <- b <- op(c)  because one has to read it right-to-left.  (I also don't know if I've tested (or even written) the post increment operator for range_map.)

layers/range_vector.h
jzulauf-lunarg: ^^^ this... I think more people will grok this.  Thanks.

layers/synchronization_validation.cpp
jzulauf-lunarg: DON'T CROSS THE STREAMS!

layers/synchronization_validation.cpp
jzulauf-lunarg: Yes, this is just me being paranoid after not writing range_map code in a while. I'll pull it.

jzulauf-lunarg (commented):
I'll integrate this comments and retest.

jzulauf-lunarg:
@jeremyg-lunarg @ncesario-lunarg -- I think I've integrated all your feedback

ncesario-lunarg (commented):
LGTM pending Jeremy's approval.

jeremyg-lunarg (approved):
LGTM

PR #4582, opened 2022-09-21
layers: Fix spurious VUID-vkCmdEndQueryIndexedEXT-queryType-06696

rg3igalia:
Several CTS test in the dEQP-VK.transform_feedback.* group were incorrectly reporting VUID-vkCmdEndQueryIndexedEXT-queryType-06696. The cause is some of these tests using two query pools and the layers not verifying the query pool being checked matched the expected one.


ziga-lunarg (approved):

ziga-lunarg:
@rg3igalia  I think rebasing from master will fix the CI error.

Tony-LunarG (approved):
LGTM

PR #4583, opened 2022-09-21
layers: Save multisample state if render pass is not null

rg3igalia:
Several tests in the dEQP-VK.transform_feedback.* group are passing a valid pMultisampleState pointer when creating pipelines with rasterization disabled, and the layers are still generating VUID-VkGraphicsPipelineCreateInfo-renderpass-06631.

The VUID requires multisample state to be passed if the pipeline contains fragment shader state and a non-NULL render pass, but the layers do not save the provided multisample state info due to rasterization being disabled and later emit the validation error incorrectly.

This commit makes the layers save the multisample state if a render pass is provided.

### DIFF
--- a/layers/generated/vk_safe_struct.cpp
+++ b/layers/generated/vk_safe_struct.cpp
@@ -4025,7 +4025,7 @@ safe_VkGraphicsPipelineCreateInfo::safe_VkGraphicsPipelineCreateInfo(const VkGra
         pRasterizationState = new safe_VkPipelineRasterizationStateCreateInfo(in_struct->pRasterizationState);
     else
         pRasterizationState = NULL;
-    if (in_struct->pMultisampleState && (has_rasterization || is_graphics_library))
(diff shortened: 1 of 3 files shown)


ncesario-lunarg (approved):
Thanks @rg3igalia, would you be able to rebase this on top of master? This is currently failing on our internal CI due to an unrelated test that has just been fixed.

This change LGTM, though I am curious what parts of fragment shader state are used when rasterization is disabled? Mainly because I'm wondering if there is other state we should not be setting to null as well.

Also this description is a little misleading now in the spec:
>pMultisampleState is a pointer to a [VkPipelineMultisampleStateCreateInfo](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap27.html#VkPipelineMultisampleStateCreateInfo) structure defining multisample state used when rasterization is enabled.

The
> ...when rasterization is enabled

Should probably be removed or updated.

Issue #4530, opened 2022-09-08
Add interactions for dynamic rendering in vkCmdClearAttachments


mikes-lunarg:
**Describe the Issue**
vkCmdClearAttachments validation changes in 1.3.227

**Valid Usage ID**
VUID-vkCmdClearAttachments-aspectMask-02501
VUID-vkCmdClearAttachments-aspectMask-02502
VUID-vkCmdClearAttachments-aspectMask-02503

**Environment:**
Header v1.3.227

**Additional context**
Spec Diff: https://github.com/KhronosGroup/Vulkan-Docs/commit/09d120580acb942d9cc3a96c863815a05990893c#diff-51b3a7d73dfaaf16f2b4cfc74503c6e8da60f8a391edba14cd6f3172bb0a014eL302-L316
Spec MR: https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5240

PR #4595, opened 2022-09-27
Update vkCmdClearAttachments VUIDs

arno-lunarg:
Issue https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4530

Those three VUIDs were removed:

**VUID-vkCmdClearAttachments-aspectMask-02501**
    If the *aspectMask* member of any element of *pAttachments*
    contains VK_IMAGE_ASPECT_COLOR_BIT, then the *colorAttachment*
    member of that element must either refer to a color attachment which is
    VK_ATTACHMENT_UNUSED, or must be a valid color attachment

**VUID-vkCmdClearAttachments-aspectMask-02502**
    If the *aspectMask* member of any element of *pAttachments*
    contains VK_IMAGE_ASPECT_DEPTH_BIT, then the current subpass'
    depth/stencil attachment must either be VK_ATTACHMENT_UNUSED, or
    must have a depth component

**VUID-vkCmdClearAttachments-aspectMask-02503**
    If the *aspectMask* member of any element of *pAttachments*
    contains VK_IMAGE_ASPECT_STENCIL_BIT, then the current subpass'
    depth/stencil attachment must either be VK_ATTACHMENT_UNUSED, or
    must have a stencil component

And those two added:

**VUID-vkCmdClearAttachments-pAttachments-07270**
For each element of *pAttachments*, the corresponding attachment in the current render pass instance must either not be backed by an image view, or contain each of the aspects specified in *aspectMask*

**VUID-vkCmdClearAttachments-aspectMask-07271**
If the *aspectMask* member of any element of *pAttachments* contains VK_IMAGE_ASPECT_COLOR_BIT, the *colorAttachment* must be a valid color attachment index in the current render pass instance

Those two better account for dynamic rendering. Other related VUIDs saw their name change, I updated them.

To make validation easier, I setup all necessary variables at the beginning of `PreCallValidateCmdClearAttachments`, based on whether or not dynamic rendering is used, and do the VUID checks using those variables.


layers/buffer_validation.cpp
```diff
@@ -3580,123 +3580,187 @@ bool CoreChecks::PreCallValidateCmdClearAttachments(VkCommandBuffer commandBuffe
...
+                    if (color_view_state &&
+                        !(color_view_state->safe_create_info.subresourceRange.aspectMask & VK_IMAGE_ASPECT_COLOR_BIT)) {
+                        skip |= LogError(commandBuffer, "VUID-vkCmdClearAttachments-pAttachments-07270",
+                                         "vkCmdClearAttachments() pAttachments[%u].colorAttachment=%u is backed by an "
+                                         "image view that does not have aspect VK_IMAGE_ASPECT_COLOR_BIT%s.",
+                                         attachment_index, clear_desc->colorAttachment, renderpass_info.c_str());
                     }
+                } else {
+                    skip |= LogError(commandBuffer, "VUID-vkCmdClearAttachments-aspectMask-07271",
+                                     "vkCmdClearAttachments() pAttachments[%u].colorAttachment=%u out of range%s."
```
ncesario-lunarg, line 3678: Should there be a space between `range` and `%s` (assuming `renderpass_info` does not start with whitespace, but maybe it does?)? Same questions for similar instances below.

ncesario-lunarg (approved):
Some minor nits, but LGTM!

layers/buffer_validation.cpp
```diff
@@ -3580,123 +3580,187 @@ bool CoreChecks::PreCallValidateCmdClearAttachments(VkCommandBuffer commandBuffe
...
+                }
+
+                renderpass_info = " for " + report_data->FormatHandle(cb_node->activeRenderPass->renderPass()) + " subpass " +
+                                  std::to_string(cb_node->activeSubpass);
+            }
+
             if (aspect_mask & VK_IMAGE_ASPECT_METADATA_BIT) {
                 skip |= LogError(commandBuffer, "VUID-VkClearAttachment-aspectMask-00020",
-                                 "vkCmdClearAttachments() pAttachments[%u] mask contains VK_IMAGE_ASPECT_METADATA_BIT",
+                                 "vkCmdClearAttachments() pAttachments[%u] mask contains VK_IMAGE_ASPECT_METADATA_BIT.",
```
jeremyg-lunarg, line 3659: I'm pretty sure the logging subsystem already adds a "." to the messages. Double check that you don't have ".." on the messages you've changed.

layers/buffer_validation.cpp
```diff
@@ -3580,123 +3580,187 @@ bool CoreChecks::PreCallValidateCmdClearAttachments(VkCommandBuffer commandBuffe
...
+            const IMAGE_VIEW_STATE *depth_view_state = nullptr;
+
+            bool has_valid_stencil_attachment = false;
+            const IMAGE_VIEW_STATE *stencil_view_state = nullptr;
+
+            uint32_t view_mask = 0;
+
+            std::string renderpass_info;
+
+            if (cb_node->activeRenderPass->use_dynamic_rendering) {
```
jeremyg-lunarg, line 3608: Should this be UsesDynamicRendering()?  Unless you need to distinguish between the normal and inherited case, the method that checks both should be used.

ShabbyX:
Could we please revert this change? It's regressing in ANGLE due to a bug.

In particular, it's complaining about the following situation (for example):

- Render pass has two attachments
- Subpass has the following attachments: {0, UNUSED, UNUSED, 1}
- vkCmdClearAttachments with attachments {0, 3}

This is valid Vulkan.

arno-lunarg:
Hi @ShabbyX, and thank you for pointing this error.
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4621 should be a valid fix.

PR #4599, opened 2022-09-28
layers: Don't auto-validate SPIR-V after optimization

ShabbyX:
In CoreChecks::ValidatePipelineShaderStage, the SPIR-V optimizer is run followed by validation of the optimized shader.  However, the SPIR-V optimizer itself is also validating the optimized SPIR-V.

This change disables the validation done by the optimizer, and lets it be done by CoreChecks::ValidatePipelineShaderStage.

As a result, pipeline-creation-heavy tests in ANGLE see ~10% speed up.


ncesario-lunarg (approved):
LGTM!

ncesario-lunarg:
Internal CI failure is unrelated and spurious, so merging this.

PR #4600, opened 2022-09-29
layers: Add check for 02267, 02268 and 02269

zzoon:
Found these are missing when implementing relevant CTSes.


spencer-lunarg:
@zzoon sorry for late reply, the code looks good, but we need to have a test written for each of these VUIDs before we can merge it 

https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/docs/creating_tests.md

PR #4601, opened 2022-09-29
layers: Disable friendly names in SPIR-V validation

ShabbyX:
The friendly-name-mapper in spirv-val is slow. Disabling it in VVL improves performance of an ANGLE test that creates numerous shaders by ~10%.  The cost of SPIR-V validation itself is reduced by ~40%.

### DIFF
--- a/layers/shader_validation.cpp
+++ b/layers/shader_validation.cpp
@@ -3841,4 +3841,7 @@ void AdjustValidatorOptions(const DeviceExtensions &device_extensions, const Dev
         // --allow-localsizeid
         options.SetAllowLocalSizeId(true);
     }
+
+    // Faster validation without friendly names.
+    options.SetFriendlyNames(false);
 }


ShabbyX:
Note: this needs https://github.com/KhronosGroup/SPIRV-Tools/pull/4951 to land first

layers/shader_validation.cpp
```diff
@@ -3841,4 +3841,7 @@ void AdjustValidatorOptions(const DeviceExtensions &device_extensions, const Dev
         // --allow-localsizeid
         options.SetAllowLocalSizeId(true);
     }
+
+    // Faster validation without friendly names.
+    options.SetFriendlyNames(false);
```
ncesario-lunarg, line 3846: @Tony-LunarG could this effect GPUAV in any way?

scripts/known_good.json
```diff
@@ -37,7 +37,7 @@
       "cmake_options": [
         "-DSPIRV-Headers_SOURCE_DIR={repo_dir}/../SPIRV-Headers"
       ],
-      "commit": "4c456f7da67c5437a6fb7d4d20d78e2a5ae2acf2"
+      "commit": "4dbc66380dd63aabbd33c38198008449d0a5807a"
```
ncesario-lunarg, line 40: Is this required? This usually gets updated along with SPIRV-Headers, but if there's a reason to update it on its own (e.g., this includes an important bug fix), we can do a one-off here.

ncesario-lunarg: If this isn't required here, I've added this change to #4609, as we need to also update glslang and SPIRV-Headers for the tests in the upcoming mesh shader PR.

layers/shader_validation.cpp
Tony-LunarG: It doesn't appear to change any of the messages and that's the only place I would expect any changes, so no.

ShabbyX: Thanks for checking

scripts/known_good.json
ShabbyX: Not required, I just didn't know when you would make an update. Since one is in progress, I'll wait for that to land and rebase.

ShabbyX:
This is ready to land

PR #4606, opened 2022-09-29
VK_EXT_micromap and associated

ewerness-nv:
With the new sync2 validation generation, the generate script changes to add new bits are needed for the build to not break when updating. Version 230 adds two extensions with a new bit, so this branch right now has four changes:

1. Update to .230
2. Cherry-pick Carsten's change to add optical flow bit parsing
3. VK_EXT_opacity_micromap validation
4. Temporary change before the SPIRV-Headers PR is submitted to point at the new SPIRV-Headers


ewerness-nv:
Updating a few more of the known_good.json now hits issues with other extensions. I'm abandoning my attempt to create a PR that holistically builds and will instead post one that I believe will build once all the other changes are brought together.

PR #4602, opened 2022-09-29
layers: Changes required for VK_NV_optical_flow implementation

crohdewm:

PR #4607, opened 2022-09-29
layers: VK_EXT_opacity_micromap

ewerness-nv:
Note: This requires https://github.com/KhronosGroup/SPIRV-Headers/pull/292 to be merged and updated in the various known_good.json and requires the version of Vulkan-Headers to be updated to .230, both of which appear to require multiple changes to be merged at once (confirmed at least this one and https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4602 and it looks like something else about an ARM SPIR-V extension)


sjfricke:
heads up @ewerness-nv you will need to re-run the code generation to have your `./scripts/` changes take effect 

`python3 ./scripts/generate_source.py ./external/Vulkan-Headers/registry/ ./external/SPIRV-Headers/include/spirv/unified1/`

ewerness-nv:
@sjfricke My conclusion from https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4606 was that it seemed like multiple changes would have to be generated and submitted together for dependency ordering reasons, which is why I'm keeping this one as the basic change. 

If there's some dependency ordering that I can use to make this one pass by itself that would be great - I just don't know of it.

ewerness-nv:
@ncesario-lunarg Do you have an opinion on how we should land these changes that seem to be co-dependent? 

See #4606 for the previous attempt. Updating Vulkan-Headers requires the new bits be handled in the generate script for all extensions that add stage bits (this and optical flow) but taking this change before Vulkan-Headers fails because parameter_validation_utils uses the new enum extensions. The set of 4 changes there passes some builds, but fails some others that I don't have locally, so it wasn't easy for me to finalize how many more changes needed to be pulled in together to get a passing CI.

ncesario-lunarg:
1.3.230 is merged, so I think if you rebase this change and run the generate script as @sjfricke mentioned, you should be good. I tried this locally and had to do some manual fixup (which I may not have done correctly) to sync_val_gen.py as there were some overlapping changes made in #4615.

ncesario-lunarg (approved):
This LGTM and runs clean on all of our CI devices. @ewerness-nv Should I create a separate issue for tracking the addition of validation, or is that being tracked over on gitlab?

PR #4612, opened 2022-09-30
git: Ignore CMakerUserPresets.json

juan-lunarg:
Allows users to have their own custom presets to speed up local workflows.


ncesario-lunarg (approved):
LGTM! I don't even know how to use that file :/

PR #4615, opened 2022-09-30
v1.3.230 update

mikes-lunarg:


mikes-lunarg:
@jeremyg-lunarg I could use a sanity check on the sync_val_gen.py updates

ziga-lunarg (approved):

layers/core_validation_error_enums.h
```diff
@@ -40,7 +40,6 @@ static const char DECORATE_UNUSED *kVUID_Core_Bound_Resource_FreedMemoryAccess =
 // static const char DECORATE_UNUSED *kVUID_Core_MemTrack_InvalidMap = "UNASSIGNED-CoreValidation-MemTrack-InvalidMap";
 static const char DECORATE_UNUSED *kVUID_Core_MemTrack_InvalidState = "UNASSIGNED-CoreValidation-MemTrack-InvalidState";
 // static const char DECORATE_UNUSED *kVUID_Core_MemTrack_InvalidUsageFlag = "UNASSIGNED-CoreValidation-MemTrack-InvalidUsageFlag";
-static const char DECORATE_UNUSED *kVUID_Core_MemTrack_RebindObject = "UNASSIGNED-CoreValidation-MemTrack-RebindObject";
```
ncesario-lunarg, line 43: Should this just be commented out rather than removed?

ncesario-lunarg (approved):

layers/core_validation_error_enums.h
mikes-lunarg: Maybe? I'm not really sure why these were originally commented out. In general, I'm against commenting out code because we have version control.

ncesario-lunarg: I agree, but @mark-lunarg told me: "this is the way" :)

mark-lunarg: I agree also -- in these cases we were leaving the old definitions for traceability if VUIDs changed, and as none of us seem to remember why let's take this opporutnity to delete  (all?) these things.

mikes-lunarg: Looks like it started 4 years ago when we converted our internal error enums to strings, and some of them weren't used even then. https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/51653904ad8c78598c9eb6a476cc6ad42cba1b80

It would be my pleasure to rip out the rest

mark-lunarg: Yes, please!

Issue #4614, opened 2022-09-30
Warnings from macOS build "built for newer macOS version (11.7) than being linked (10.12)"


juan-lunarg:
MacOs builds on CI are seeing multiple linker warnings. Here is an example:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/actions/runs/3160572537/jobs/5145179242
```
ld: warning: object file (/Users/runner/work/Vulkan-ValidationLayers/Vulkan-ValidationLayers/external/glslang/build/install/lib/libSPVRemapper.a(SPVRemapper.cpp.o)) was built for newer macOS version (11.7) than being linked (10.12)
ld: warning: object file (/Users/runner/work/Vulkan-ValidationLayers/Vulkan-ValidationLayers/external/glslang/build/install/lib/libSPIRV.a(GlslangToSpv.cpp.o)) was built for newer macOS version (11.7) than being linked (10.12)
ld: warning: object file (/Users/runner/work/Vulkan-ValidationLayers/Vulkan-ValidationLayers/external/glslang/build/install/lib/libMachineIndependent.a(ShaderLang.cpp.o)) was built for newer macOS version (11.7) than being linked (10.12)
ld: warning: object file (/Users/runner/work/Vulkan-ValidationLayers/Vulkan-ValidationLayers/external/googletest/build/install/lib/libgtest.a(gtest-all.cc.o)) was built for newer macOS version (11.7) than being linked (10.12)
ld: warning: object file (/Users/runner/work/Vulkan-ValidationLayers/Vulkan-ValidationLayers/external/SPIRV-Tools/build/install/lib/libSPIRV-Tools.a(binary.cpp.o)) was built for newer macOS version (11.7) than being linked (10.12)
ld: warning: object file (/Users/runner/work/Vulkan-ValidationLayers/Vulkan-ValidationLayers/external/SPIRV-Tools/build/install/lib/libSPIRV-Tools.a(table.cpp.o)) was built for newer macOS version (11.7) than being linked (10.12)
```

**Environment:**
 - OS:  MacOS

This is occurring due to `CMAKE_OSX_DEPLOYMENT_TARGET` not being properly propagated. Unsure of severity of warning.

juan-lunarg:
Also worth pointing out that if this is fixed to make linker warnings as errors to prevent this issue in the future.

ncesario-lunarg:
FYI @richard-lunarg.

[...]

PR #4618, opened 2022-10-03
build: Minor build fixes

juan-lunarg:
- Add MacOS to CI build status
- Use CMake idioms
- Fix invalid CMake warning, CMAKE_GENERATOR_PLATFORM only makes sense for VS builds in the context of update_deps.py

### DIFF
--- a/.github/workflows/build_macos.yml
+++ b/.github/workflows/build_macos.yml
@@ -56,4 +56,4 @@ jobs:
       - name: Generate build files and dependencies
         run: cmake -S. -Bbuild -DCMAKE_BUILD_TYPE=${{matrix.config}} -DBUILD_TESTS=ON -DUPDATE_DEPS=ON -DUPDATE_DEPS_SKIP_EXISTING_INSTALL=ON
       - name: Build Vulkan-ValidationLayers
-        run: cmake --build build --parallel $(sysctl -n hw.logicalcpu)
+        run: cmake --build build --parallel $(sysctl -n hw.logicalcpu) --config ${{matrix.config}}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,7 +42,7 @@ option(BUILD_TESTS "Build the tests" OFF)
 # Enable beta Vulkan extensions
 add_definitions(-DVK_ENABLE_BETA_EXTENSIONS)
 
-set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
 if (UPDATE_DEPS)
     find_package(PythonInterp 3 REQUIRED)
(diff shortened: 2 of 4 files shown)


juan-lunarg:
Originally I was trying to fix https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4614 but the scope got out of hand. 

I didn't want to keep these fixes for 1 giant PR, where these changes would complicate the discussion.

.github/workflows/build_macos.yml
```diff
@@ -56,4 +56,4 @@ jobs:
       - name: Generate build files and dependencies
         run: cmake -S. -Bbuild -DCMAKE_BUILD_TYPE=${{matrix.config}} -DBUILD_TESTS=ON -DUPDATE_DEPS=ON -DUPDATE_DEPS_SKIP_EXISTING_INSTALL=ON
       - name: Build Vulkan-ValidationLayers
-        run: cmake --build build --parallel $(sysctl -n hw.logicalcpu)
+        run: cmake --build build --parallel $(sysctl -n hw.logicalcpu) --config ${{matrix.config}}
```
juan-lunarg, line 59: This file doesn't explicitly say what the generator being used is. So for safety make the code work with multi/single config generators.

CMakeLists.txt
```diff
@@ -42,15 +42,17 @@ option(BUILD_TESTS "Build the tests" OFF)
 # Enable beta Vulkan extensions
 add_definitions(-DVK_ENABLE_BETA_EXTENSIONS)
 
-set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
```
juan-lunarg, line 45: Minor CMake idiom

CMakeLists.txt
```diff
@@ -42,15 +42,17 @@ option(BUILD_TESTS "Build the tests" OFF)
...
 if (UPDATE_DEPS)
     find_package(PythonInterp 3 REQUIRED)
 
     if (CMAKE_GENERATOR_PLATFORM)
         set(_target_arch ${CMAKE_GENERATOR_PLATFORM})
     else()
-        message(WARNING "CMAKE_GENERATOR_PLATFORM not set. Using x64 as target architecture.")
+        if (MSVC_IDE)
+            message(WARNING "CMAKE_GENERATOR_PLATFORM not set. Using x64 as target architecture.")
+        endif()
```
juan-lunarg, line 55: It only makes sense to warn about this for the VS generator (and technically Green Hills Multi, but I've never seen anyone use that)

CMakeLists.txt
```diff
@@ -138,6 +140,9 @@ if(USE_CCACHE)
 endif()
 
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+set(CMAKE_CXX_VISIBILITY_PRESET "hidden")
+set(CMAKE_C_VISIBILITY_PRESET "hidden")
+set(CMAKE_VISIBILITY_INLINES_HIDDEN "YES")
```
juan-lunarg, line 145: CMake idiom that replaces -fvisibility=hidden, no affect on Windows

CMakeLists.txt
juan-lunarg: Also Green Hills Multi use of CMAKE_GENERATOR_PLATFORM would just break the python script from what I can tell

ncesario-lunarg (approved):
LGTM!

PR #4620, opened 2022-10-04
tests: Fix test group and whitespace

sjfricke:
I had the positive test as `VkLayerTest`

also found some white space my editor noticed that figured could go well with this nit cleanup PR


ncesario-lunarg:
The CI failure looks pretty suspicious. Re-running for sanity.

PR #4625, opened 2022-10-05
layers: Fix null descriptor set crash

ziga-lunarg:


ncesario-lunarg (approved):
LGTM!

ncesario-lunarg:
Is there a test case that should be added to cover this?

layers/cmd_buffer_state.cpp
```diff
@@ -1045,6 +1045,9 @@ void CMD_BUFFER_STATE::UpdatePipelineState(CMD_TYPE cmd_type, const VkPipelineBi
             if (set_index >= state.per_set.size()) {
                 continue;
             }
+            if (!state.per_set[set_index].bound_descriptor_set) {
+                continue;
```
arno-lunarg, line 1049: Nit: this test could use the following variable

arno-lunarg:
LGTM! (One NIT)

arno-lunarg (approved):

jeremyg-lunarg (approved):

layers/cmd_buffer_state.cpp
ziga-lunarg: The following variable is a refence, if this is nullptr it is not valid to take a reference from it.

ziga-lunarg:
> Is there a test case that should be added to cover this?

I am seeing these crashes in my own application, using multiple threads with indirect draws, but I have not been able to reproduce this with a test yet.

layers/cmd_buffer_state.cpp
arno-lunarg: To my knowledge, I think it is valid to do so? `auto &descriptor_set` refers to the variable `state.per_set[set_index].bound_descriptor_set`, and his just a "synonym" for this variable, unrelated to its potential values.  Or am I missing something.
This code compiles and runs just fine: https://godbolt.org/z/xj6ejcE87

ziga-lunarg: It's not valid to have a reference to a nullptr, but this isn't even a regular pointer, it's a `shared_ptr` so I guess I was looking at the wrong code. The problem comes in the line after this.

layers/cmd_buffer_state.cpp
```diff
@@ -1045,8 +1045,14 @@ void CMD_BUFFER_STATE::UpdatePipelineState(CMD_TYPE cmd_type, const VkPipelineBi
...
+            }
             // Pull the set node
             auto &descriptor_set = state.per_set[set_index].bound_descriptor_set;
+            if (descriptor_set) {
```
arno-lunarg, line 1053: I think the `!` is missing

layers/cmd_buffer_state.cpp
```diff
@@ -1045,8 +1045,14 @@ void CMD_BUFFER_STATE::UpdatePipelineState(CMD_TYPE cmd_type, const VkPipelineBi
             if (set_index >= state.per_set.size()) {
                 continue;
             }
+            if (!state.per_set[set_index].bound_descriptor_set) {
```
arno-lunarg, line 1048: This can now be removed

PR #4629, opened 2022-10-05
syncval: Rework async hazard check range bounds

jzulauf-lunarg:
Ensure async hazard checks apply to all accesses since most recent sync point (including indirectly through a semaphore chain).  Previously only checked "last_batch" operations.

### DIFF
--- a/layers/synchronization_validation.cpp
+++ b/layers/synchronization_validation.cpp
@@ -19,10 +19,12 @@
  * Author: Jeremy Gebben <jeremyg@lunarg.com>
  */
 
-#include <limits>
-#include <vector>
-#include <memory>
+#include <algorithm>
 #include <bitset>
+#include <limits>
+#include <memory>
+#include <vector>
+
 #include "synchronization_validation.h"
 #include "sync_utils.h"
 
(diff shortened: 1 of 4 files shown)


ncesario-lunarg (approved):
LGTM, but would be good to get a sign off from JeremyG first.

layers/synchronization_validation.cpp
```diff
@@ -4493,10 +4514,10 @@ void SyncValidator::CreateDevice(const VkDeviceCreateInfo *pCreateInfo) {
     SetCommandBufferResetCallback([this](VkCommandBuffer command_buffer) -> void { ResetCommandBufferCallback(command_buffer); });
     SetCommandBufferFreeCallback([this](VkCommandBuffer command_buffer) -> void { FreeCommandBufferCallback(command_buffer); });
 
-    QueueId queue_id = QueueSyncState::kQueueIdBase;
-    ForEachShared<QUEUE_STATE>([this, &queue_id](const std::shared_ptr<QUEUE_STATE> &queue_state) {
+    ForEachShared<QUEUE_STATE>([this](const std::shared_ptr<QUEUE_STATE> &queue_state) {
         auto queue_flags = physical_device_state->queue_family_properties[queue_state->queueFamilyIndex].queueFlags;
-        std::shared_ptr<QueueSyncState> queue_sync_state = std::make_shared<QueueSyncState>(queue_state, queue_flags, queue_id++);
+        std::shared_ptr<QueueSyncState> queue_sync_state =
+            std::make_shared<QueueSyncState>(queue_state, queue_flags, queue_id_limit_++);
```
jeremyg-lunarg, line 4520: Is this one of the reasons you eventually want per-type IDs in the state tracker?

jeremyg-lunarg (approved):
LGTM

layers/synchronization_validation.cpp
jzulauf-lunarg: @jeremyg-lunarg This one is a bit of a unique case.  The total number of queues available is fixed at device create time, allowing the code to index into vector, as opposed to finding in a map. Also the queue states are guaranteed to endure for device scope allowing queue references by index rather than shared_ptr in most cases.

Issue #4579, opened 2022-09-20
Incorrect warning for unconsumed attributes related to VK_KHR_maintenance4/Vulkan 1.3


pdoane:
A completely unconsumed vertex output produces the following message when VK_KHR_maintenance4 is disabled:

```
vertex shader writes to output location 0.2 which is not consumed by fragment shader.  Enable VK_KHR_maintenance4 device extension to allow relaxed interface matching between input and output vectors.
```

VK_KHR_maintenance4 relaxes interface matching between vectors of different sizes (e.g. vertex shader writes `vec4` but fragment shader reads `vec3`).  It is unrelated to whether outputs are consumed.

Two problems with the current behavior:
  - The warning should always be generated even when VK_KHR_maintenance4 is enabled.  It is still a potential performance issue to have unconsumed output.
  - The warning should not mention VK_KHR_maintenance4.  That is not related to unconsumed outputs but different vector sizes.

See #4148 and #4352 for additional context.  Both of these closed with the assumption that enabling VK_KHR_maintenance4 should disable the performance warning.

arno-lunarg:
Hi @pdoane,

You mention completely unconsumed vertex output, but your message reads `location 0.2`. In the case of a completely unconsumed output, I would expect `location 0.0` to be printed instead. So maybe there is a bug on top of the issues you mention.

Are you able to give me the full out/in interface from both shaders so I can know more please?

pdoane:
Oops I hand typed in the error message and transposed that.

```
struct VertexOutput
{
  float4 ndcPos : SV_Position;
  uint textureOffset : TEXTURE_OFFSET;  // Location 0
  float2 zw : ZW;  // Location 1
  float2 tc : TEXCOORD0; // Location 2
};

struct PixelInput
{
  float4 screenPos : SV_Position;
  uint textureOffset : TEXTURE_OFFSET;
  float2 zw : ZW;  // Location 1
  float2 tc : TEXCOORD0;
};
```

The fragment shader does not read location 0 or 2 - it's completely unconsumed and generates the error message above but fixed to say `location 0.0` and `location 2.0`.

arno-lunarg:
Thank you for the clarification!

spencer-lunarg:
> The fragment shader does not read location 0 or 2 - it's completely unconsumed and generates the error message above

[...]

PR #4630, opened 2022-10-06
Issue 4579 - Unconsumed attributes warning&error behavior

arno-lunarg:
Fix for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4579

Completely unconsumed attributes are allowed, and should result in a performance warning if the best practices layer is enabled.
Partially consumed attributes are allowed if `maintenance4` is enabled, but should still result in a performance warning. They are illegal when `maintenance4` is disabled, and result in an error.

### DIFF
--- a/layers/best_practices_utils.cpp
+++ b/layers/best_practices_utils.cpp
@@ -1095,6 +1095,44 @@ void BestPractices::ManualPostCallRecordCreateComputePipelines(VkDevice device, 
     pipeline_cache_ = pipelineCache;
 }
 
+// This function validates that the output of one shader is fully consumed by the consumer shader in the pipeline for performance
+// reasons
+bool BestPractices::ValidateInterfaceBetweenStages(const SHADER_MODULE_STATE& producer, spirv_inst_iter producer_entrypoint,
+                                                   shader_stage_attributes const* producer_stage,
+                                                   const SHADER_MODULE_STATE& consumer, spirv_inst_iter consumer_entrypoint,
+                                                   shader_stage_attributes const* consumer_stage) const {
+    bool skip = false;
+
+    auto check_unconsumed_attrib = [this](const SHADER_MODULE_STATE& producer, shader_stage_attributes const* producer_stage,
+                                          const ShaderAttribute& out, const SHADER_MODULE_STATE& consumer,
+                                          shader_stage_attributes const* consumer_stage, const ShaderAttribute& in) -> bool {
+        bool skip = false;
+
+        if (out.IsPartiallyUnconsumedBy(in)) {
+            if (enabled_features.core13.maintenance4) {
+                std::string msg = std::string{producer_stage->name} + " writes to output location " + std::to_string(out.location) +
+                                  '.' + std::to_string(out.component) + " which is not consumed by " + consumer_stage->name + ". ";
+                skip |= LogPerformanceWarning(producer.vk_shader_module(), kVUID_Core_Shader_OutputNotConsumed, "%s", msg.c_str());
+            }
+        }
+
+        if (out.IsFullyUnconsumedBy(in)) {
+            // out attribute is completely unconsumed
+            std::string msg = std::string{producer_stage->name} + " writes to output location " + std::to_string(out.location) +
+                              '.' + std::to_string(out.component) + " which is completely not consumed by " + consumer_stage->name +
+                              ". ";
+            skip |= LogPerformanceWarning(producer.vk_shader_module(), kVUID_Core_Shader_OutputNotConsumed, "%s", msg.c_str());
+        }
+
+        return skip;
+    };
+
+    skip |= IterateInterfaceBetweenStages(check_unconsumed_attrib, producer, producer_entrypoint, producer_stage, consumer,
+                                          consumer_entrypoint, consumer_stage);
+
+    return skip;
+}
+
 bool BestPractices::PreCallValidateCreateGraphicsPipelines(VkDevice device, VkPipelineCache pipelineCache, uint32_t createInfoCount,
                                                            const VkGraphicsPipelineCreateInfo* pCreateInfos,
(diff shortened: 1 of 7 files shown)


juan-lunarg (changes requested):

layers/best_practices_utils.cpp
```diff
@@ -1095,6 +1104,97 @@ void BestPractices::ManualPostCallRecordCreateComputePipelines(VkDevice device,
...
+    while (output_it != shader_outputs.end() || input_it != shader_inputs.end()) {
+        const bool output_at_end = output_it == shader_outputs.end();
+        const bool input_at_end = input_it == shader_inputs.end();
+
+        ShaderAttribute out_attrib;
+        out_attrib.is_valid = !output_at_end;
+        if (out_attrib.is_valid) {
+            out_attrib.location = output_it->first.first;
+            out_attrib.written_components_count = producer.GetNumComponentsInBaseType(producer.get_def(output_it->second.type_id));
+        }
```
ncesario-lunarg, line 1145: Looks like `out_attrib` is read-only in this function? Might make sense to put this in a constructor and make `out_attrib` `const`, but definitely not necessary. Same comment for `in_attrib` below (might be able to consolidate that initialization code a bit as well then).

layers/best_practices_utils.cpp
```diff
@@ -1113,6 +1213,33 @@ bool BestPractices::PreCallValidateCreateGraphicsPipelines(VkDevice device, VkPi
...
+                if (consumer.module_state->has_valid_spirv && producer.module_state->has_valid_spirv) {
+                    auto producer_id = GetShaderStageId(producer.stage_flag);
+                    auto consumer_id = GetShaderStageId(consumer.stage_flag);
+                    skip |= ValidateInterfaceBetweenStages(*producer.module_state.get(), producer.entrypoint,
+                                                           &shader_stage_attribs[producer_id], *consumer.module_state.get(),
+                                                           consumer.entrypoint, &shader_stage_attribs[consumer_id]);
+                }
+            }
+        }
+    }
```
ncesario-lunarg, line 1241: Not a ton of code here, but another duplication of CoreChecks code. Again, might be more effort than it's worth to consolidate, but if we can, that would be good.

ncesario-lunarg (changes requested):
Overall this change LGTM, and I appreciate the comments :)

Just a lot of duplication that would be nice to consolidate. Some of it I think we can consolidate pretty easily, but other parts may be more effort than it's worth.

arno-lunarg:
I am going to improve on duplication, thank you for the comments @juan-lunarg and @ncesario-lunarg!

layers/best_practices_utils.cpp
```diff
@@ -1095,6 +1104,97 @@ void BestPractices::ManualPostCallRecordCreateComputePipelines(VkDevice device,
     pipeline_cache_ = pipelineCache;
 }
 
+static uint32_t GetShaderStageId(VkShaderStageFlagBits stage) {
+    uint32_t bit_pos = uint32_t(u_ffs(stage));
+    return bit_pos - 1;
+}
```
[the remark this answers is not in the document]
jeremyg-lunarg: This is already defined in shader_validation.cpp, it should be made available outside that file somehow.

arno-lunarg:
Ok so I refactored `CoreChecks::ValidateInterfaceBetweenStages` and `BestPractices::ValidateInterfaceBetweenStages`, As well as the for loop you mentioned [here](https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4630#discussion_r989556763) @ncesario-lunarg.
It took way more time than anticipated, because of how I wanted to refactor the ValidateInterface functions. I wanted to separate the iterators increments part from the checks part, so that those two functions only need to worry about checking a (out, in) attribute pair. Turns out redoing iterators increments was kind of hard. Also, the check parts now heavily rely on the `ShaderAttribute` struct, it makes the code simpler and more explicit.

Some bugs appeared along the way, and some still are very much alive, because I did not know about the `component` attribute layout qualifier. This keyword is a nightmare, especially when combined with the special rules for types like `dvec3` and `dvec4`. ATM, I still do not handle correctly those types, since I do not know how to see if an attribute has such a type?

arno-lunarg:
Failing test is `VkLayerTest.PipelineInUseDestroyedSignaled` on tcwinnvi2 debug 64, seems unrelated to those changes

layers/shader_validation.cpp
```diff
@@ -3146,108 +3140,55 @@ bool CoreChecks::ValidateInterfaceBetweenStages(const SHADER_MODULE_STATE &produ
...
-            if ((input_first.first > output_first.first) || input_at_end || (output_component + 1 == output_length)) {
-                output_it++;
-                output_component = 0;
-            } else {
-                output_component++;
+                std::string msg = std::string{producer_stage->name} + " writes to output location " + std::to_string(out.location) +
+                                  '.' + std::to_string(out.component) + " which is not consumed by " + consumer_stage->name +
+                                  ". Enable VK_KHR_maintenance4 device extension to allow relaxed interface matching between "
+                                  "input and output vectors. ";
+                skip |= LogError(producer.vk_shader_module(), kVUID_Core_Shader_OutputNotConsumed, "%s", msg.c_str());
```
spencer-lunarg, line 3154: I think this is `VUID-RuntimeSpirv-OpTypeVector-06816`/`VUID-RuntimeSpirv-OpTypeVector-06817`

>    Any OpTypeVector output interface variables must: not have a higher
    Component Count than a matching OpTypeVector input
    interface variable
    
so would make this

```cpp
const char* vuid = IsExtEnabled(device_extensions.vk_khr_maintenance4) ? "VUID-RuntimeSpirv-OpTypeVector-06817" : "VUID-RuntimeSpirv-OpTypeVector-06816";
```

(and then apply to the test case below to check for this)

arno-lunarg:
Closed for now, spotted few more problems related to 64 bits types, I have to make some changes.

PR #4633, opened 2022-10-07
layers: Update xxhash to 0.8.1

juan-lunarg:
While the main improvements have mainly been about new hashing algorithms. There have been performance improvments to XXH32 since 0.6.2, which is what VVL currently uses.

See release notes frome details:
https://github.com/Cyan4973/xxHash/releases


juan-lunarg:
Also while XXH32 is the appropriate algorithm for usage with VUID's which require a 32 bit hash due to `VkDebugUtilsMessengerCallbackDataEXT` it may be worth looking into the newer hash algorithms for shader hashing. Which doesn't have the same limitation.

Metrics on hash performance:
https://github.com/Cyan4973/xxHash/tree/v0.8.1#benchmarks

arno-lunarg (approved):

jeremyg-lunarg (commented):
LGTM. Should we at some point consider moving this to known_good.json?

ncesario-lunarg:
> LGTM. Should we at some point consider moving this to known_good.json?

I'm guessing the rationale was that, "it's just 2 files, so might as well include it," but I do think it would be a bit clearer (and perhaps a bit easier) to have this be part of known_good.json.

juan-lunarg:
> I'm guessing the rationale was that, "it's just 2 files, so might as well include it," but I do think it would be a bit clearer (and perhaps a bit easier) to have this be part of known_good.json.

I had 2 problems with having xxhash in known_good.json 

xxhash has CMake support, but in a subdirectory named `cmake_unofficial`
Also the xxhash files get installed as part of VVL for some reason, I'm not sure I understand.

Issue #4355, opened 2022-07-27
Descriptor validation for OpTypeImage prematurely checks signedness


aitor-lunarg:
**Describe the Issue**
As stated in the title, if we have an OpTypeImage (https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpTypeImage) with a mismatching `Sampled Type` of the bound image we will get `VUID-vkCmdDispatch-None-02699` even if the access respects `Image Format`'s signedness.

The spec (https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#spirvenv-format-type-matching) states that `The signedness must match the signedness of any access to the image`. Spir-v 1.4 introduces `SignExtend` and `ZeroExtend` to force signedness when accessing the resource, so until resource access, we won't know if the access is correct as stated in https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#spirvenv-image-signedness

**Valid Usage ID**
VUID-vkCmdDispatch-None-02699

**Environment:**
 - OS: Any
 - GPU: Any
 - SDK or header version if building from repo: ToT
 - Options enabled (synchronization, best practices, etc.):

**Additional context**
Here's a test to replicate the issue:
```
TEST_F(VkPositiveLayerTest, ImageSampleResultMismatch) {
TEST_DESCRIPTION("Try to sample an image with mismatching sample result from image format.");

    m_errorMonitor->ExpectSuccess();

    SetTargetApiVersion(VK_API_VERSION_1_1);
    AddRequiredExtensions(VK_KHR_SPIRV_1_4_EXTENSION_NAME);
    ASSERT_NO_FATAL_FAILURE(InitFramework(m_errorMonitor));

    if (DeviceValidationVersion() < VK_API_VERSION_1_1) {
        GTEST_SKIP() << "At least Vulkan version 1.1 is required";
    }

    ASSERT_NO_FATAL_FAILURE(InitState());

    if (!AreRequiredExtensionsEnabled()) {
        GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported.";
    }

    const char *csSource = R"(
                          OpCapability Shader
                          OpCapability StorageImageExtendedFormats
                          %std = OpExtInstImport "GLSL.std.450"
                          OpMemoryModel Logical GLSL450
                          OpEntryPoint GLCompute %main "main" %image_ptr
                          OpExecutionMode %main LocalSize 1 1 1
[...]

PR #4634, opened 2022-10-07
layers: SPIR-V Instruction class

sjfricke:
This adds a new `Instruction` class in favor of the old `spirv_inst_iter` method. 

- **What this provides** 
   - Having a capsulated `Instruction`  class makes the code more readable that only a single instruction is being passed around and not a random iterator.
   - This makes the code more like `SPIRV-Tools`/`spirv-val`
   - Opens up for more refactoring to move Instruction logic (ex. `Instruction::GetStorageClass` or `Instruction::GetBitWidth`) to be added
   - Will allow us to finally solve issues such as https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4355#issuecomment-1266428301
 - **What is not done**
   - I didn't remove `spirv_inst_iter` completely because both GPU-AV and debug-printf make use of the iterator more and wanted to test it separately as it might include more invasive changes to the code. Plan to do that in a follow-up PR

### DIFF
--- a/layers/shader_instruction.h
+++ b/layers/shader_instruction.h
@@ -1,0 +1,70 @@
+/* Copyright (c) 2022 The Khronos Group Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Author: Spencer Fricke <spencerfricke@gmail.com>
+ *
+ * The Shader Instruction file is in charge of holding instruction information
+ */
+#ifndef VULKAN_SHADER_INSTRUCTION_H
+#define VULKAN_SHADER_INSTRUCTION_H
+
+#include <stddef.h>  // size_t for gcc
+#include <stdint.h>
+#include <vector>
+#include <spirv/unified1/spirv.hpp>
+
+struct SHADER_MODULE_STATE;
+
+struct atomic_instruction_info {
+    uint32_t storage_class;
+    uint32_t bit_width;
+    uint32_t type;  // ex. OpTypeInt
+};
+
+// Holds information about a single SPIR-V instruction
+// Provides easy access to len, opcode, and content words without the caller needing to care too much about the physical SPIRV module layout.
+//
+// For more information of the physical module layout to help understand this struct:
+// https://github.com/KhronosGroup/SPIRV-Guide/blob/master/chapters/parsing_instructions.md
+class Instruction {
+    public:
+    Instruction(std::vector<uint32_t>::const_iterator it);
+    ~Instruction() {}
+
+    // The word used to define the Instruction
+    uint32_t word(size_t index) const { return words_[index]; }
+
+    // The words used to define the Instruction
+    const std::vector<uint32_t>& words() const { return words_; }
+
+    uint32_t length() const { return words_[0] >> 16; }
+
+    uint32_t opcode() const { return words_[0] & 0x0ffffu; }
+
+    // operand id index, return 0 if no result
+    uint32_t ResultId() const { return result_id_; }
+    // operand id index, return 0 if no type
+    uint32_t TypeId() const { return type_id_; }
+
+    char const * GetAsString(uint32_t operand) const { return (char const *)&words_.at(operand); }
+    atomic_instruction_info GetAtomicInfo(const SHADER_MODULE_STATE& module_state) const;
+    spv::BuiltIn GetBuiltIn() const;
+
+    private:
+        std::vector<uint32_t> words_;
+        uint32_t result_id_;
+        uint32_t type_id_;
+};
+
+#endif  // VULKAN_SHADER_INSTRUCTION_H
--- a/layers/shader_instruction.h
+++ b/layers/shader_instruction.h
@@ -22,11 +22,12 @@
 #include <stddef.h>  // size_t for gcc
 #include <stdint.h>
 #include <vector>
+#include <assert.h>
 #include <spirv/unified1/spirv.hpp>
 
 struct SHADER_MODULE_STATE;
 
-struct atomic_instruction_info {
+struct AtomicInstructionInfo {
     uint32_t storage_class;
     uint32_t bit_width;
     uint32_t type;  // ex. OpTypeInt
@@ -43,22 +44,29 @@ class Instruction {
     ~Instruction() {}
 
     // The word used to define the Instruction
-    uint32_t word(size_t index) const { return words_[index]; }
+    uint32_t Word(size_t index) const { return words_[index]; }
 
     // The words used to define the Instruction
-    const std::vector<uint32_t>& words() const { return words_; }
+    const std::vector<uint32_t>& Words() const { return words_; }
 
-    uint32_t length() const { return words_[0] >> 16; }
+    uint32_t Length() const { return words_[0] >> 16; }
 
-    uint32_t opcode() const { return words_[0] & 0x0ffffu; }
+    uint32_t Opcode() const { return words_[0] & 0x0ffffu; }
 
     // operand id index, return 0 if no result
     uint32_t ResultId() const { return result_id_; }
     // operand id index, return 0 if no type
     uint32_t TypeId() const { return type_id_; }
 
-    char const * GetAsString(uint32_t operand) const { return (char const *)&words_.at(operand); }
-    atomic_instruction_info GetAtomicInfo(const SHADER_MODULE_STATE& module_state) const;
+    // Only used to get strings in SPIR-V instructions
+    // SPIR-V spec (and spirv-val) ensure:
+    // "A string is interpreted as a nul-terminated stream of characters"
+    char const* GetAsString(uint32_t operand) const {
+        assert(operand < Length());
+        return (char const*)&words_.at(operand);
+    }
+
+    AtomicInstructionInfo GetAtomicInfo(const SHADER_MODULE_STATE& module_state) const;
     spv::BuiltIn GetBuiltIn() const;
 
     private:
--- a/layers/shader_instruction.h
+++ b/layers/shader_instruction.h
@@ -45,7 +45,7 @@ class Instruction {
     ~Instruction() = default;
 
     // The word used to define the Instruction
-    uint32_t Word(size_t index) const { return words_[index]; }
+    uint32_t Word(uint32_t index) const { return words_[index]; }
 
     uint32_t Length() const { return words_[0] >> 16; }
 
--- a/layers/shader_instruction.h
+++ b/layers/shader_instruction.h
@@ -23,6 +23,7 @@
 #include <stdint.h>
 #include <vector>
 #include <assert.h>
+#include "vk_layer_data.h"
 #include <spirv/unified1/spirv.hpp>
 
 struct SHADER_MODULE_STATE;
@@ -41,13 +42,10 @@ struct AtomicInstructionInfo {
 class Instruction {
     public:
     Instruction(std::vector<uint32_t>::const_iterator it);
-    ~Instruction() {}
+    ~Instruction() = default;
 
     // The word used to define the Instruction
     uint32_t Word(size_t index) const { return words_[index]; }
-
-    // The words used to define the Instruction
-    const std::vector<uint32_t>& Words() const { return words_; }
 
     uint32_t Length() const { return words_[0] >> 16; }
 
@@ -63,7 +61,7 @@ class Instruction {
     // "A string is interpreted as a nul-terminated stream of characters"
     char const* GetAsString(uint32_t operand) const {
         assert(operand < Length());
-        return (char const*)&words_.at(operand);
+        return (char const*)&words_[operand];
     }
 
     AtomicInstructionInfo GetAtomicInfo(const SHADER_MODULE_STATE& module_state) const;
@@ -73,7 +71,15 @@ class Instruction {
     bool operator!=(Instruction const& other) const { return words_ != other.words_; }
 
   private:
-    std::vector<uint32_t> words_;
+    // When this class was created, for SPIR-V Instructions that could be used in Vulkan,
+    //   414 of 423 had 6 or less operands
+    //   361 of 423 had 5 or less operands
+    //   287 of 423 had 4 or less operands
+    // An extra word is still needed because each insturction has one word prior to the operands
+    static constexpr uint32_t word_vector_length = 7;
+
+    // Max capacity needs to be uint32_t because an instruction can have a string operand that is (2^16)-1 bytes long
+    small_vector<uint32_t, word_vector_length, uint32_t> words_;
     uint32_t result_id_;
     uint32_t type_id_;
 };
--- a/layers/shader_instruction.h
+++ b/layers/shader_instruction.h
@@ -69,10 +69,13 @@ class Instruction {
     AtomicInstructionInfo GetAtomicInfo(const SHADER_MODULE_STATE& module_state) const;
     spv::BuiltIn GetBuiltIn() const;
 
-    private:
-        std::vector<uint32_t> words_;
-        uint32_t result_id_;
-        uint32_t type_id_;
+    bool operator==(Instruction const& other) const { return words_ == other.words_; }
+    bool operator!=(Instruction const& other) const { return words_ != other.words_; }
+
+  private:
+    std::vector<uint32_t> words_;
+    uint32_t result_id_;
+    uint32_t type_id_;
 };
 
 #endif  // VULKAN_SHADER_INSTRUCTION_H
--- a/layers/shader_module.cpp
+++ b/layers/shader_module.cpp
@@ -898,7 +898,8 @@ void SHADER_MODULE_STATE::DefineStructMember(const Instruction* insn, std::vecto
     data.size = local_offset + data1.size * total_array_length;
 }
 
-static uint32_t UpdateOffset(uint32_t offset, const std::vector<uint32_t> &array_indices, const shader_struct_member &data) {
+uint32_t SHADER_MODULE_STATE::UpdateOffset(uint32_t offset, const std::vector<uint32_t>& array_indices,
+                                           const shader_struct_member& data) const {
     int array_indices_size = static_cast<int>(array_indices.size());
     if (array_indices_size) {
         uint32_t array_index = 0;
@@ -912,7 +913,8 @@ static uint32_t UpdateOffset(uint32_t offset, const std::vector<uint32_t> &array
     return offset;
 }
 
-static void SetUsedBytes(uint32_t offset, const std::vector<uint32_t> &array_indices, const shader_struct_member &data) {
+void SHADER_MODULE_STATE::SetUsedBytes(uint32_t offset, const std::vector<uint32_t>& array_indices,
+                                       const shader_struct_member& data) const {
     int array_indices_size = static_cast<int>(array_indices.size());
     uint32_t block_memory_size = data.size;
     for (uint32_t i = static_cast<int>(array_indices_size); i < data.array_length_hierarchy.size(); ++i) {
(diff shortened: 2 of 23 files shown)


layers/shader_module.h
```diff
@@ -267,61 +237,70 @@ struct SHADER_MODULE_STATE : public BASE_NODE {
...
-    // The spirv image itself
-    // NOTE: this _must_ be initialized first.
-    // NOTE: this may end up being an _optimized_ version of what was passed in at initialization time.
-    const std::vector<uint32_t> words;
-
-    const SpirvStaticData static_data_;
+    // This is the SPIR-V module data content
+    const std::vector<uint32_t> words_;
+    // List of all instructions in the order they appear in the binary
+    std::vector<Instruction> instructions_;
```
ncesario-lunarg, line 243: If I understand correctly this will effectively duplicate all of `words_`. Might make sense to use `span` to store the words associated with each instruction, as you should end up with effectively the same code, save _some_ space, and avoid a few copies, but probably doesn't make too much of a difference one way or the other here.

layers/shader_module.h
```diff
@@ -429,7 +396,7 @@ struct SHADER_MODULE_STATE : public BASE_NODE {
     // Used to populate the shader module object
     void PreprocessShaderBinary(spv_target_env env);
 
-    static std::unordered_multimap<std::string, EntryPoint> ProcessEntryPoints(const SHADER_MODULE_STATE &module_state);
+    StaticData static_data_;
```
ncesario-lunarg, line 399: Does this need to be non-const now? If so, the name "static data" probably doesn't fit. The original idea behind this class was to put all "const data" within it and initialize everything within its constructor. It looks like the function `SetStaticData` is only called within constructors, so it seems like this should still be possible?

layers/shader_module.h
```diff
@@ -267,61 +237,70 @@ struct SHADER_MODULE_STATE : public BASE_NODE {
...
 
     const bool has_valid_spirv{false};
     const uint32_t gpu_validation_shader_id{std::numeric_limits<uint32_t>::max()};
 
+    bool has_group_decoration{false};
+
     SHADER_MODULE_STATE(const uint32_t *code, std::size_t count, spv_target_env env = SPV_ENV_VULKAN_1_0)
         : BASE_NODE(static_cast<VkShaderModule>(VK_NULL_HANDLE), kVulkanObjectTypeShaderModule),
-          words(code, code + (count / sizeof(uint32_t))),
-          static_data_(*this) {
```
ncesario-lunarg, line 283: This should probably remain the same if possible or be renamed (see previous comments).

layers/shader_validation.cpp
```diff
@@ -48,38 +48,38 @@ static shader_stage_attributes shader_stage_attribs[] = {
     {"fragment shader", false, false, VK_SHADER_STAGE_FRAGMENT_BIT},
 };
 
-static const spirv_inst_iter GetBaseTypeIter(const SHADER_MODULE_STATE &module_state, uint32_t type) {
-    const auto &insn = module_state.get_def(type);
+static const Instruction *GetBaseTypeIter(const SHADER_MODULE_STATE &module_state, uint32_t type) {
```
ncesario-lunarg, line 51: Should this be renamed to `GetBaseTypeInstruction` now?

layers/shader_module.h
```diff
@@ -267,61 +237,70 @@ struct SHADER_MODULE_STATE : public BASE_NODE {
...
-
-    const SpirvStaticData static_data_;
+    // This is the SPIR-V module data content
+    const std::vector<uint32_t> words_;
+    // List of all instructions in the order they appear in the binary
+    std::vector<Instruction> instructions_;
+    // Instructions that can be referenced by Ids
+    // A mapping of <id> to the first word of its def. this is useful because walking type
+    // trees, constant expressions, etc requires jumping all over the instruction stream.
+    layer_data::unordered_map<uint32_t, const Instruction *> definitions_;
```
ncesario-lunarg, line 247: Looks like this is const data, and therefore could go into `static_data_`.

layers/shader_instruction.h
```diff
@@ -0,0 +1,70 @@
...
+    uint32_t ResultId() const { return result_id_; }
+    // operand id index, return 0 if no type
+    uint32_t TypeId() const { return type_id_; }
+
+    char const * GetAsString(uint32_t operand) const { return (char const *)&words_.at(operand); }
+    atomic_instruction_info GetAtomicInfo(const SHADER_MODULE_STATE& module_state) const;
+    spv::BuiltIn GetBuiltIn() const;
+
+    private:
+        std::vector<uint32_t> words_;
```
ncesario-lunarg, line 65: See previous comment. Might make sense to make this a `span`, but also might not.

layers/shader_instruction.h
```diff
@@ -0,0 +1,70 @@
...
+    uint32_t length() const { return words_[0] >> 16; }
+
+    uint32_t opcode() const { return words_[0] & 0x0ffffu; }
+
+    // operand id index, return 0 if no result
+    uint32_t ResultId() const { return result_id_; }
+    // operand id index, return 0 if no type
+    uint32_t TypeId() const { return type_id_; }
+
+    char const * GetAsString(uint32_t operand) const { return (char const *)&words_.at(operand); }
```
ncesario-lunarg, line 60: This seems a little scary if the wrong operand is passed in; are we guaranteed a null value at least at the end of the SPIR-V content?

layers/shader_instruction.h
```diff
@@ -0,0 +1,70 @@
...
+#define VULKAN_SHADER_INSTRUCTION_H
+
+#include <stddef.h>  // size_t for gcc
+#include <stdint.h>
+#include <vector>
+#include <spirv/unified1/spirv.hpp>
+
+struct SHADER_MODULE_STATE;
+
+struct atomic_instruction_info {
```
ncesario-lunarg, line 29: `CamelCase`.

ncesario-lunarg (commented):
This is awesome, I've been wanting something like this for a long time!

Most of these comments are style-related nits and some questions about constness (as there was a pretty large effort to make as much data "const" as possible in the past), but overall I think this is great. Thanks @sjfricke!

layers/shader_instruction.h
```diff
@@ -0,0 +1,69 @@
...
+//
+// For more information of the physical module layout to help understand this struct:
+// https://github.com/KhronosGroup/SPIRV-Guide/blob/master/chapters/parsing_instructions.md
+class Instruction {
+    public:
+    Instruction(std::vector<uint32_t>::const_iterator it);
+    ~Instruction() {}
+
+    // The word used to define the Instruction
+    uint32_t word(size_t index) const { return words_[index]; }
```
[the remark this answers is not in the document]
sjfricke: So happy to make these function names match normal code convention, just not the diff will be MUCH bigger as I will have to touch everywhere that uses it

layers/shader_module.h
```diff
@@ -225,40 +205,30 @@ struct shader_struct_member {
 
 struct SHADER_MODULE_STATE : public BASE_NODE {
     struct EntryPoint {
-        uint32_t offset;  // into module to get OpEntryPoint instruction
+        const Instruction *insn;  // OpEntryPoint instruction
```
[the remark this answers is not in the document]
sjfricke: These should not be null pointers, unless we are given a bogus SPIR-V module that `spirv-val` didn't catch, but realize the shader_module code can have bugs... let me play around with this and get a more robust solution

layers/shader_module.h
sjfricke: So the only 2 places we still actually even need `words_` is

- GPU-AV (because didn't touch it yet)
- When we run `spirv-val`/`spirv-opt` and need to give it the fully buffer

`span` is a c++ 20 feature I think, are we at c++ 20 yet?

layers/shader_instruction.h
sjfricke: So I guess all this was doing was moving the "scary" cast to a function... let me just add an bound check by making sure `operand` is not above `length` and that should make sure nothing aweful happens

layers/shader_instruction.h
sjfricke: This was moved here, but about to do a full "make the legacy names match what they should" pass now :muscle:

layers/shader_instruction.h
sjfricke: Added in a comment above the function, but [SPIR-V spec](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Literal) says

> A string is interpreted as a nul-terminated stream of characters

layers/shader_module.h
sjfricke: There are 2 issues I am running into

1. Unless we do an initial loop prior, I don't know how big to make `span` at init time
2. In the future we will want `Instruction` to be copied and owned outside the life of the `SHADER_MODULE_STATE` and I think you run into the issue where sometimes the `SHADER_MODULE_STATE` owns the `words` and sometimes the `Instruction` do, not sure if a more clever way to work around that

I don't like having essentially a copy of the `words` neither, but not 100% at this moment which is the best approach

layers/shader_module.cpp
```diff
@@ -487,60 +464,59 @@ void SHADER_MODULE_STATE::PreprocessShaderBinary(const spv_target_env env) {
...
             // NOTE: We need to update words with the result from the spirv-tools optimizer.
             // **THIS ONLY HAPPENS ON INITIALIZATION**. words should remain const for the lifetime
             // of the SHADER_MODULE_STATE instance.
-            *const_cast<std::vector<uint32_t> *>(&words) = std::move(optimized_binary);
+            *const_cast<std::vector<uint32_t>*>(&words_) = std::move(optimized_binary);
             // Will need to update static data now the words have changed or else the def_index will not align
             // It is really rare this will get here as Group Decorations have been deprecated and before this was added no one ever
             // raised an issue for a bug that would crash the layers that was around for many releases
-            *const_cast<SpirvStaticData *>(&static_data_) = SpirvStaticData(*this);
+            *const_cast<StaticData*>(&static_data_) = StaticData(*this);
```
sjfricke, line 477: CI is not happy because this seems to not work with `EntryPoint::insn` being a reference and blowing up trying to copy `std::unordered_multimap<std::string, EntryPoint> entry_points;`... can reproduce on Android, will need to pull out some extra std template foo to figure out what operator needs to be given an override

(tried `Instruction &operator=(const Instruction &other)` but seems to not be any luck)

layers/shader_instruction.h
```diff
@@ -0,0 +1,81 @@
...
+
+// Holds information about a single SPIR-V instruction
+// Provides easy access to len, opcode, and content words without the caller needing to care too much about the physical SPIRV module layout.
+//
+// For more information of the physical module layout to help understand this struct:
+// https://github.com/KhronosGroup/SPIRV-Guide/blob/master/chapters/parsing_instructions.md
+class Instruction {
+    public:
+    Instruction(std::vector<uint32_t>::const_iterator it);
+    ~Instruction() {}
```
juan-lunarg, line 44:
```suggestion
    ~Instruction() = default;
```

layers/shader_module.cpp
```diff
@@ -320,20 +320,51 @@ layer_data::optional<VkPrimitiveTopology> SHADER_MODULE_STATE::GetTopology(const
...
-SHADER_MODULE_STATE::SpirvStaticData::SpirvStaticData(const SHADER_MODULE_STATE &module_state) {
-    for (auto insn : module_state) {
-        const uint32_t result_word = OpcodeResultWord(insn.opcode());
-        if (result_word != 0) {
-            def_index[insn.word(result_word)] = insn.offset();
+SHADER_MODULE_STATE::StaticData::StaticData(const SHADER_MODULE_STATE& module_state) {
+    // Parse the words first so we have instruction class objects to use
+    {
+        std::vector<uint32_t>::const_iterator it = module_state.words_.cbegin();
+        it += 5;  // skip first 5 word of header
```
juan-lunarg, line 332: Is there a constant that could be used instead?

sjfricke: There is no value in the SPIR-V headers, but this is just a constant in the spec, `SPIRV-Tools` does the same with a literal

layers/shader_instruction.h
```diff
@@ -0,0 +1,81 @@
...
+    uint32_t ResultId() const { return result_id_; }
+    // operand id index, return 0 if no type
+    uint32_t TypeId() const { return type_id_; }
+
+    // Only used to get strings in SPIR-V instructions
+    // SPIR-V spec (and spirv-val) ensure:
+    // "A string is interpreted as a nul-terminated stream of characters"
+    char const* GetAsString(uint32_t operand) const {
+        assert(operand < Length());
+        return (char const*)&words_.at(operand);
```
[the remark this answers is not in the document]
sjfricke: before the assert, but now you are right, will turn to index

layers/shader_instruction.h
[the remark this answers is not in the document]
ncesario-lunarg: Yeah, I kind of assumed some of the choices were to avoid widespread changes. I still have hopes of getting clang-tidy integration working at some point, and this would get changed as part of that effort, so up to you on whether or not to do it here :)

layers/shader_module.h
ncesario-lunarg: So my assumption was that (2) was _not_ the case (i.e., an `Instruction` would _not_ outlive the `SHADER_MODULE_STATE` it belongs to). If that assumption is false, then I think the current solution is the only way to go (i.e., copy the words into the instruction).

layers/shader_module.cpp
ncesario-lunarg: It would be nice if this information was in one of the json files in SPIRV-Headers, but it looks like this is currently missing.

layers/shader_instruction.h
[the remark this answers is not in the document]
ncesario-lunarg: It might not be terrible if we ended up with an abort in release here rather than somewhere down the road if operand is invalid. Since we have exceptions disabled, I would be surprised if `at` makes much of a perf difference here (via branch prediction and other dark arts...), but my assumption may very well be invalid.

layers/shader_module.cpp
sjfricke: Raised issue, for now should be good, but :100: agree would be better to have the const in the official headers
https://github.com/KhronosGroup/SPIRV-Headers/issues/297

layers/shader_instruction.h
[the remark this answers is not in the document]
sjfricke: well with the `assert` I can't imagine where indexing would ever go bad and I would almost change it to be consistent instead of making the next dev waste anytime going "why is this `at` and not `[]`"

layers/shader_module.cpp
ncesario-lunarg: If going the reference route, I think you'll need to use something like `std::reference_wrapper<Instruction>` in order to get copy semantics. Again, I'm not sure this is the "right" way to go as I haven't personally used `std::reference_wrapper` a whole lot, but that seems like one way to indicate `insn` is non-null with code (though it would be nice if a plain old reference "just worked"...).

layers/shader_module.h
sjfricke: > Instruction would not outlive the SHADER_MODULE_STATE it belongs to)

So the only time have to worry is

```c++
vkCreateShaderModule(&module);
vkCreatePipeline(module, &pipeline);
vkDestroyShaderModule(module);

vkCmdBindPipeline(pipeline);
vkCmdDraw(); // Invalid draw time instruction
```

To my knowledge `SHADER_MODULE_STATE` would be gone at this moment (if not, well then yes, `2` is not an issue) and we still want to be able to save the Instruction

I am still looking into a way to `Instruction` both be able to easily hold a its own `vector<uint32_t>` or `span<uint32_t>`

sjfricke: One way would be using the `operator=` and it `Instruction` is constructed it is a `span` but if copied, we do a deep copy and grab the memory over

ncesario-lunarg: > So the only time have to worry is....

Yep, [this](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap9.html#vkDestroyShaderModule) bit in the spec makes it pretty clear:
> A shader module can be destroyed while pipelines created using its shaders are still in use.

So yeah, definitely need to copy. That said...I'm pretty sure this would crash in the validation layers currently 🤔

ncesario-lunarg: > I am still looking into a way to Instruction both be able to easily hold a its own vector<uint32_t> or span<uint32_t>

Looks like we need to copy, so not sure that is necessary. `layer_data::small_vector` might be good to use here, too, since I'm assuming most instructions will be about the same size and we could then avoid some heap allocations.

I suppose something like the following _could_ (??) work, but it would have to be with unions since we don't have access to `std::variant`:
```cpp
template<int N = 0>
struct Instruction {
        ...
        std::span<const uint32_t> Words() const {
                const auto& s = std::get<N>(store_);
                return std::span<const uint32_t>(s.begin(), s.end());
        }
        ...
        std::variant<std::vector<uint32_t>, std::span<uint32_t>> store_;
};

using InstructionDeepCopy = Instruction<0>;
using InstructionShallowCopy = Instruction<1>;
```

layers/vk_layer_data.h
```diff
@@ -198,6 +198,18 @@ class small_vector {
...
+        if (size_ != rhs.size_) return true;
+        auto value = begin();
+        for (const auto &rh_value : rhs) {
+            if (!(*value == rh_value)) {
+                return true;
+            }
+            ++value;
+        }
+        return false;
+    }
```
juan-lunarg, line 211: Couldn't we leverage the existing `==` operator?

```suggestion
    bool operator!=(const small_vector &rhs) const { return !(*this == rhs); }
```

juan-lunarg (changes requested):
minor nitpick

juan-lunarg (approved):
LGTM but @ncesario-lunarg is much more familiar with the code at hand

layers/shader_instruction.h
```diff
@@ -0,0 +1,87 @@
...
+    spv::BuiltIn GetBuiltIn() const;
+
+    bool operator==(Instruction const& other) const { return words_ == other.words_; }
+    bool operator!=(Instruction const& other) const { return words_ != other.words_; }
+
+  private:
+    // When this class was created, for SPIR-V Instructions that could be used in Vulkan,
+    //   414 of 423 had 6 or less operands
+    //   361 of 423 had 5 or less operands
+    //   287 of 423 had 4 or less operands
```
ncesario-lunarg, line 77: Good to know!

ncesario-lunarg:
I think the following should get things compiling:
```diff
diff --git a/layers/shader_instruction.h b/layers/shader_instruction.h
index 92ee89a46..54be749ac 100644
--- a/layers/shader_instruction.h
+++ b/layers/shader_instruction.h
@@ -45,7 +45,7 @@ class Instruction {
     ~Instruction() = default;

     // The word used to define the Instruction
-    uint32_t Word(size_t index) const { return words_[index]; }
+    uint32_t Word(uint32_t index) const { return words_[index]; }

     uint32_t Length() const { return words_[0] >> 16; }

diff --git a/layers/shader_module.cpp b/layers/shader_module.cpp
index d085acbf8..072c47a16 100644
--- a/layers/shader_module.cpp
+++ b/layers/shader_module.cpp
@@ -429,7 +429,7 @@ SHADER_MODULE_STATE::StaticData::StaticData(const SHADER_MODULE_STATE& module_st

                 auto range = entry_points.equal_range(entrypoint_name);
                 for (auto it = range.first; it != range.second; ++it) {
-                    if (it->second.insn == insn) {
+                    if (insn == it->second.insn.get()) {
                         entry_point = &(it->second);
                         break;
                     }
@@ -595,7 +595,7 @@ layer_data::optional<Instruction> SHADER_MODULE_STATE::FindEntrypoint(char const
     auto range = static_data_.entry_points.equal_range(name);
     for (auto it = range.first; it != range.second; ++it) {
         if (it->second.stage == stageBits) {
-            assert(it->second.insn.Opcode() == spv::OpEntryPoint);
+            assert(it->second.insn.get().Opcode() == spv::OpEntryPoint);
             result.emplace(it->second.insn);
             break;
         }
diff --git a/layers/shader_module.h b/layers/shader_module.h
index f52f97002..f0d03c21b 100644
--- a/layers/shader_module.h
+++ b/layers/shader_module.h
@@ -205,7 +205,7 @@ struct shader_struct_member {

 struct SHADER_MODULE_STATE : public BASE_NODE {
     struct EntryPoint {
-        const Instruction &insn;  // OpEntryPoint instruction
+        std::reference_wrapper<const Instruction> insn;  // OpEntryPoint instruction
         VkShaderStageFlagBits stage;
         std::vector<function_set> function_set_list;
         shader_struct_member push_constant_used_in_shader;
```

ncesario-lunarg (approved):

spencer-lunarg:
CI Tests are unrelated and being discussed inside to prevent the false positive

PR #4636, opened 2022-10-08
layers: Remove temp fix for re-running spirv-opt

sjfricke:
We added this re-run in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4420 but now with the changes from `spirv-opt` pulled in (https://github.com/KhronosGroup/SPIRV-Tools/pull/4932) there is no need to re-run `spirv-opt`

Note: Some operations are still not supported in the `spirv-opt` pass ([CompositeInsert](https://github.com/KhronosGroup/SPIRV-Tools/pull/4943), non-32 bit ints, and 16 bit floats) but I am actively working on adding support. By "not supported" I mean it will remain as a `OpSpecConstantOp` and not get turned into the desirable `OpConstant`


ncesario-lunarg:
The CI failure is only seen on the nvidia Shield and is likely unrelated. Re-running for sanity and will then merge.

ncesario-lunarg (approved):

PR #2294, opened 2020-11-04
layers: Add Android 24 and 25 support

sfricke-samsung:
closes #2287 

- This adds a `AHB_VALIDATION_SUPPORT` macro depending on building for Android and the version
  - replaced `VK_USE_PLATFORM_ANDROID_KHR` with new macro where needed
- BUILD.md update how to switch to `android-24` (which will work with `android-25`)
- added a `CheckAndroidVersion` call to run before `dlopen` so that if another developer runs into this same issue it will point them to what is wrong and hopefully save them time

----

@izhido I applied the following patch and built the layers, here are binary drops so they should work on your device

```patch
diff --git a/build-android/jni/Application.mk b/build-android/jni/Application.mk
index 243c28538..db1323a15 100644
--- a/build-android/jni/Application.mk
+++ b/build-android/jni/Application.mk
@@ -13,9 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.

-APP_ABI := armeabi-v7a arm64-v8a x86 x86_64
+APP_ABI := armeabi-v7a arm64-v8a
 # APP_ABI := arm64-v8a   # just build for pixel2  (don't check in)
-APP_PLATFORM := android-26
+APP_PLATFORM := android-24
 APP_STL := c++_static
 NDK_TOOLCHAIN_VERSION := clang
 NDK_MODULE_PATH := .
diff --git a/build-android/jni/shaderc/Application.mk b/build-android/jni/shaderc/Application.mk
index 5447415eb..214d05a7e 100644
--- a/build-android/jni/shaderc/Application.mk
+++ b/build-android/jni/shaderc/Application.mk
@@ -1,4 +1,4 @@
-APP_ABI := all
+APP_ABI := armeabi-v7a arm64-v8a
 APP_BUILD_SCRIPT := Android.mk
 APP_STL := c++_static
 APP_PLATFORM := android-23
```

[PR_build.zip](https://github.com/KhronosGroup/Vulkan-ValidationLayers/files/5489680/PR_build.zip)

PR #2651, opened 2021-03-24
layers: Add an Android default log force option via setprop

neelbedekar:
It's useful for certain downstream tools/systems(like automation)
to parse the log output of validation layers. On Android via logcat,
the default validation layer callback only prints when the Vulkan app
explicitly doesn't subscribe to the VK_EXT_debug_utils callback via
vkCreateDebugUtilsMessengerEXT. When the app does subscribe, then only
the app's callback is executed. To allow such downstream tools to
parse logcat for validation layer errors, this PR adds a mechanism to
force the default log to always be printed, regardless of whether the
app subscribes via vkCreateDebugUtilsMessengerEXT. This PR is currently
limited to Android-only, and opt-in with an Android system property.
This allows downstream tools to set the property and parse the logcat
in a known form(accessing key info like VUID), and every other existing
app relying on validation layers will remain unaffected.

PR #4637, opened 2022-10-08
layers: Don't spawn process for android prop lookup

### LINKED ISSUE #4605: Android: CheckAndroidVersion spawns subprocesses when it doesn't have to
**Describe the Issue**
On Android, GetEnvironment helper function forks extra processes/threads, which confuses developers.
It uses popen to fork a thread to run "getprop".  But GetEnvironment could have used the getenv C library call instead.

**Valid Usage ID**
n/a

**Environment:**
 - OS: Android
 - GPU: n/a
 - SDK or header version if building from repo: 

**Additional context**

A developer was debugging an Android Vulkan application and was confused when they threads being spawned.

The CheckAndroidVersion call at https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/e81f04ef6b6ad186bd771aca2d761a92d2277069/layers/vk_layer_config.cpp#L426 is called during dlopen.
That in turn calls GetEnvironment:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/e81f04ef6b6ad186bd771aca2d761a92d2277069/layers/vk_layer_config.cpp#L87 

That spawns a subprocess to run 'getprop'. It could have used getenv instead.

sjfricke:
(might) close #4605 

Removes the Android `GetEnvironment` logic added from https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/2294 as in Android "Environment" and "Property" values are different things

### DIFF
--- a/layers/vk_layer_config.cpp
+++ b/layers/vk_layer_config.cpp
@@ -60,6 +60,7 @@ class ConfigFile {
 
 static ConfigFile layer_config;
 
+// Note: Android doesn't use environment and instead uses properties values
 std::string GetEnvironment(const char *variable) {
 #if !defined(__ANDROID__) && !defined(_WIN32)
     const char *output = getenv(variable);
@@ -74,38 +75,22 @@ std::string GetEnvironment(const char *variable) {
     string output = buffer;
     delete[] buffer;
     return output;
-#elif defined(__ANDROID__)
-    string command = "getprop " + string(variable);
-    FILE *pPipe = popen(command.c_str(), "r");
-    if (pPipe != nullptr) {
-        char value[256];
-        fgets(value, 256, pPipe);
-        pclose(pPipe);
-
-        // Make sure its not an empty line and does not end in a newline
-        const auto str_len = strcspn(value, "\r\n");
-        if (str_len == 0) {
-            return "";
-        } else {
-            value[str_len] = '\0';
-            return string(value);
-        }
-    } else {
-        return "";
-    }
 #else
     return "";
 #endif
 }
 
 const char *getLayerOption(const char *option) { return layer_config.GetOption(option); }
-const char *GetLayerEnvVar(const char *option) {
-    // NOTE: new code should use GetEnvironment directly. This is a workaround for the problem
-    // described in https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/3048
-    static std::string result;
-    result = GetEnvironment(option);
-    return result.c_str();
+#if defined(__ANDROID__)
+#include <sys/system_properties.h>
+VK_LAYER_EXPORT int GetAndroidProperty(const char *prop, char *out) {
+    // __system_property_get deprecated from Android O, but still using
+    // See https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4605
+    return __system_property_get(prop, out);
 }
+#endif  // defined(__ANDROID__)
+
+VK_LAYER_EXPORT const char *getLayerOption(const char *option) { return layer_config.GetOption(option); }
 
 const SettingsFileInfo *GetLayerSettingsFileInfo() { return &layer_config.settings_info; }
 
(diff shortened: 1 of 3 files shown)


layers/vk_layer_config.cpp
```diff
@@ -82,25 +83,6 @@ VK_LAYER_EXPORT std::string GetEnvironment(const char *variable) {
     string output = buffer;
     delete[] buffer;
     return output;
```
jeremyg-lunarg, line 85: There are several other calls to GetEnvironment() during VVL startup.  See https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/layers/layer_options.cpp#L395
Probably CI will break without the ability to set some of the properties through the code block that is removed here.

However, there are other uses of GetEnvironment() where VVL is really looking for environment variables like $HOME. It seems like a new abstraction for layer options is needed.

layers/vk_layer_config.cpp
```diff
@@ -423,19 +405,25 @@ VK_LAYER_EXPORT void PrintMessageType(VkFlags vk_flags, char *msg_flags) {
...
-    if ((target_version != 0) && (target_version < 26)) {
-        LOGCONSOLE(
-            "ERROR - Targeted Android version is %d and needs to be 26 or above. Please read "
-            "https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/BUILD.md for how to build the Validation Layers "
-            "for Android 25 and below",
-            target_version);
+    // __system_property_get deprecated from Android O, but still using
+    // See https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4605
+    char version[8];
+    int str_length = __system_property_get("ro.build.version.sdk", version);
```
jeremyg-lunarg, line 415: This part is great though, since it is android specific code looking at a standard property.

layers/vk_layer_config.cpp
jeremyg-lunarg: There's code in the internal CI scripts that does this:
```
      # Layer settings.
        for variable, value in layer_settings:
            # Android sends layer settings via setprop.
            # ? Do we have to restore the original values later?
            with log_lock:
                log.info('%s: setting layer setting: setprop %s %s', device, variable, value)
            command = [adb, '-s', device.serial, 'shell', 'setprop', variable, value]
            unsetprop.append(variable)
            _run(command)
```
And I remember but cannot find code that translated environment variable names into things that look more like properties? @lunarpapillo Is the above block of code legacy or important? That is, does internal CI use the setting file or properties to configure VVL on Android?

ncesario-lunarg: _If_ `__system_property_get` can reliably be used to get Android properties (I wasn't sure if this was necessarily the case), then perhaps we need to do the following for `GetEnvironment`:
1. Check result of `getenv`, if non-null, return the result
2. Otherwise, check the result of `__system_property_get` with `debug.vvl.` pre-pended to `variable`, return the result

Then `setprop debug.vvl.*` can be used to set validation layer settings in the same way that setting environment variables are used on desktop...if I understand correctly.

ncesario-lunarg: Looks like we maybe want to use `__system_property_find` and `__system_property_read_callback` (https://android.googlesource.com/platform/bionic/+/master/libc/include/sys/system_properties.h)? I don't see `__system_property_get` in the header anymore (it was deprecated...).

sjfricke: `__system_property_read_callback` was not added until Android 26, I don't have an Android 24/25 device to test, but I assume it will not load the layers when it sees this symbol and has no internal implementation of it... That being said, we could have 2 solutions, one for just getting the Android SDK version and one for general property value getting

sjfricke: Ok, I forgot about https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/2651 and adding the `debug.vvl.` namespace. I refactored this to be a common function and moved what should be the only 2 usages over to use it

ncesario-lunarg: > `__system_property_read_callback` was not added until Android 26, I don't have an Android 24/25 device to test, but I assume it will not load the layers when it sees this symbol and has no internal implementation of it... That being said, we could have 2 solutions, one for just getting the Android SDK version and one for general property value getting

I think we have to do something crazy along the lines of https://codereview.chromium.org/393923002/diff/20001/base/sys_info_android.cc (which I think @mikes-lunarg found). i.e., see if the `__system_property_read_callback` is present, and if not, use `__system_property_get`...and I guess if neither one is present we give up.

PR #4641, opened 2022-10-10
Fix for PR #4595 - Secondary cmd buffers not accounted for when using dynamic rendering

arno-lunarg:
I was forgetting to account for secondary cmd buffers when using dynamic rendering. For the classic render pass model, the way the state it managed makes handling thins "automatic", so no changes should be needed here.
This PR also adds a test case for secondary cmd buffers.


layers/cmd_buffer_state.h
```diff
@@ -523,6 +523,28 @@ class CMD_BUFFER_STATE : public REFCOUNTED_NODE {
     uint32_t GetDynamicDepthResolveAttachmentImageIndex() const { return 2 * GetDynamicColorAttachmentCount() + 1; }
     uint32_t GetDynamicStencilAttachmentImageIndex() const { return 2 * GetDynamicColorAttachmentCount() + 2; }
     uint32_t GetDynamicStencilResolveAttachmentImageIndex() const { return 2 * GetDynamicColorAttachmentCount() + 3; }
+    bool HasValidDynamicDepthAttachment() const {
+        if (activeRenderPass) {
+            if (activeRenderPass->use_dynamic_rendering_inherited) {
```
ncesario-lunarg, line 528: Should/can this check the `depthAttachmentFormat != VK_FORMAT_UNDEFINED` in the inheritance info?

layers/cmd_buffer_state.h
```diff
@@ -523,6 +523,28 @@ class CMD_BUFFER_STATE : public REFCOUNTED_NODE {
...
+            }
+            if (activeRenderPass->use_dynamic_rendering) {
+                return activeRenderPass->dynamic_rendering_begin_rendering_info.pDepthAttachment != nullptr;
+            }
+        }
+        return false;
+    }
+    bool HasValidDynamicStencilAttachment() const {
+        if (activeRenderPass) {
+            if (activeRenderPass->use_dynamic_rendering_inherited) {
```
ncesario-lunarg, line 539: Same question as with depth above.

ncesario-lunarg (commented):

Tony-LunarG (approved):
The changes LGTM but there seems to be an empty merge commit - we usually rebase.

ncesario-lunarg (approved):

ncesario-lunarg:
> The changes LGTM but there seems to be an empty merge commit - we usually rebase.

I've checked the "linear history" option for the master branch on this repo, which supposedly blocks PRs with merge commits from getting merged. Regardless, something doesn't seem right about ab8ad92f84d1e4ff68d4e7f71e38aabd05178b82.

jeremyg-lunarg (approved):

PR #4643, opened 2022-10-10
build: more xxhash fixes

### LINKED ISSUE #4640: MSVC warning in 64bit following xxHash update.
After the update of xxHash in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4633 the roll of VVLs in Dawn and ANGLE is failing because of an MSVC warning being triggered. Looking at the code I don't really understand what the problem is so I'll add a suppression in the BUILD.gn file. Out of curiosity, @kaadam, do you think that https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4639 would fix this issue (seems unlikely, but we never know).

The exact warning emitted is:
```
ninja -t msvc -e environment.x64 -- "C:\b\s\w\ir\cache\win_toolchain\vs_files\1023ce2e82\VC\Tools\MSVC\14.29.30133\bin\Hostx64\x64\cl.exe" /c ../../third_party/vulkan-deps/vulkan-validation-layers/src/layers/xxhash.c /nologo /showIncludes -DUSE_AURA=1 -D_HAS_EXCEPTIONS=0 -D__STD_C -D_CRT_RAND_S -D_CRT_SECURE_NO_DEPRECATE -D_SCL_SECURE_NO_DEPRECATE -D_ATL_NO_OPENGL -D_WINDOWS -DCERT_CHAIN_PARA_HAS_EXTRA_FIELDS -DPSAPI_VERSION=2 -DWIN32 -D_SECURE_ATL -DWINAPI_FAMILY=WINAPI_FAMILY_DESKTOP_APP -DWIN32_LEAN_AND_MEAN -DNOMINMAX -DNTDDI_VERSION=NTDDI_WIN10_FE -D_WIN32_WINNT=0x0A00 -DWINVER=0x0A00 -DNDEBUG -DNVALGRIND -DDYNAMIC_ANNOTATIONS_ENABLED=0 -D_CRT_NONSTDC_NO_WARNINGS -D_WINSOCK_DEPRECATED_NO_WARNINGS -DVULKAN_NON_CMAKE_BUILD -DVK_ENABLE_BETA_EXTENSIONS -DVK_USE_PLATFORM_WIN32_KHR -I../../third_party/vulkan-deps/vulkan-validation-layers/src/layers -I../../third_party/vulkan-deps/vulkan-validation-layers/src/layers/generated -I../../third_party/vulkan-deps/spirv-headers/src/include -I../.. -Igen -I../../third_party/vulkan-deps/vulkan-headers/src/include /Gy /FS /bigobj /utf-8 /Zc:sizedDealloc- /wd4117 /D__DATE__= /D__TIME__= /D__TIMESTAMP__= /O1 /Ob2 /Oy- /Zc:inline /Gw /Oi /Zi /MT /wd4800 /wd4267 /WX /wd4244 /std:c11 /TC /Foobj/third_party/vulkan-deps/vulkan-validation-layers/src/vulkan_layer_utils/xxhash.obj /Fd"obj/third_party/vulkan-deps/vulkan-validation-layers/src/vulkan_layer_utils_c.pdb"
C:\b\s\w\ir\cache\builder\dawn\third_party\vulkan-deps\vulkan-validation-layers\src\layers\xxhash.h(4578): error C2220: the following warning is treated as an error
C:\b\s\w\ir\cache\builder\dawn\third_party\vulkan-deps\vulkan-validation-layers\src\layers\xxhash.h(4578): warning C4028: formal parameter 4 different from declaration
```

y-novikov:
Fixes #4640

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -296,6 +296,7 @@ source_set("vulkan_layer_utils") {
     "layers/vk_layer_utils.cpp",
     "layers/vk_layer_utils.h",
     "layers/vk_mem_alloc.h",
+    "layers/vk_xxhash.h",
     "layers/xxhash.c",
     "layers/xxhash.h",
   ]
--- a/layers/xxhash.c
+++ b/layers/xxhash.c
@@ -40,4 +40,4 @@
 #define XXH_STATIC_LINKING_ONLY   /* access advanced declarations */
 #define XXH_IMPLEMENTATION   /* access definitions */
 
-#include "xxhash.h"
+#include "vk_xxhash.h"


y-novikov:
@ncesario-lunarg, could you PTAL?
@Kangz, @juan-lunarg FYI

ncesario-lunarg (approved):
LGTM, just need clang-format to run on this.

@juan-lunarg this is probably another good reason to move xxhash to `known_good.json`, as we don't really need/want to run clang-format on it. We could add an exception, going forward, but I think it makes more sense to just move it into external.

y-novikov:
I've formatted it, but I don't think that's the right thing to do, since the rest of the file is formatted differently.

juan-lunarg:
> @juan-lunarg this is probably another good reason to move xxhash to known_good.json, as we don't really need/want to run clang-format on it. We could add an exception, going forward, but I think it makes more sense to just move it into external.

I agree. Just requires knowing a commit that would work downstream (since the stable tag currently won't). And addressing the fact the CMakeLists.txt isn't in the top level directory for xxhash.

layers/xxhash.h
```diff
@@ -4520,16 +4520,13 @@ XXH3_hashLong_64b_withSeed_internal(const void* input, size_t len,
 /*
  * It's important for performance that XXH3_hashLong is not inlined.
  */
-XXH_NO_INLINE XXH64_hash_t
-XXH3_hashLong_64b_withSeed(const void* input, size_t len,
-                           XXH64_hash_t seed, const xxh_u8* secret, size_t secretLen)
-{
+XXH_NO_INLINE XXH64_hash_t XXH3_hashLong_64b_withSeed(const void* XXH_RESTRICT input, size_t len, XXH64_hash_t seed,
+                                                      const xxh_u8* XXH_RESTRICT secret, size_t secretLen) {
```
juan-lunarg, line 4524: Why is this code being compiled? XXH_NO_LONG_LONG should disable this code from being compiled at all.

ncesario-lunarg: clang-format probably isn't taking that `#define` into account.

y-novikov: I think because XXH_NO_LONG_LONG is not defined in xxhash.c.
Let me try and see if including vk_xxhash.h in xxhash.c helps.

juan-lunarg: The best approach is to define XXH_NO_LONG_LONG in the build systems and remove vk_xxhash.h

That way we don't have to modify xxhash code at all.

y-novikov: I've changed this to include "vk_xxhash.h" from "xxhash.c", this seems to work, PTAL.
Or maybe you can implement the build systems approach.

y-novikov:
> Apologies @y-novikov could you test this PR #4644?
> 
> This adds XXH_NO_LONG_LONG correctly to the builds.

Your fix works https://chromium-review.googlesource.com/c/angle/angle/+/3944893
Thanks a lot!

PR #4645, opened 2022-10-10
build: Use vcpkg

### LINKED ISSUE #3866: Add support for CMake toolchain file in update_deps.py to ease cross compilation
Hi,

I need to cross compile the Vulkan validation layers for a ARM64 Linux target system. Therefore, I need to set the  -DCMAKE_TOOLCHAIN_FILE command line option when invoking cmake.
The script **update_deps.py** doesn't support to set this cmake command line option.
Can you please add support for this.

You can find my modified **update_deps.py** in the linked pull request.

Regards,
Oliver

### LINKED ISSUE #4614: Warnings from macOS build "built for newer macOS version (11.7) than being linked (10.12)"

juan-lunarg:
Using vcpkg would replace update_deps.py

- Use industry standard
- Works much better with CMake
- Very fast
- Supports multi-configuration generators
- Cleaner builds (vcpkg detects errors very quickly)
- Custom ports in case upstream doesn't work

closes #4614
closes #3866

### DIFF
--- a/cmake/ports/glslang/portfile.cmake
+++ b/cmake/ports/glslang/portfile.cmake
@@ -1,0 +1,46 @@
+vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
+
+vcpkg_from_github(
+  OUT_SOURCE_PATH SOURCE_PATH
+  REPO KhronosGroup/glslang
+  REF 28b53119bdfbc43eae532641337a7adbb315b273
+  SHA512 52039c54c75286499294bc03fcfa4ae1377e24924ba6fe31160baa3bbfdb3389437b0b68a7519e6d4df47b8bcecf553f07b8686c9f36efd280e5bd7a74100c8a
+  HEAD_REF master
+)
+
+vcpkg_find_acquire_program(PYTHON3)
+get_filename_component(PYTHON_PATH ${PYTHON3} DIRECTORY)
+vcpkg_add_to_path("${PYTHON_PATH}")
+
+if(VCPKG_TARGET_IS_IOS)
+  # this case will report error since all executable will require BUNDLE DESTINATION
+  set(BUILD_BINARIES OFF)
+else()
+  set(BUILD_BINARIES ON)  
+endif()
+
+vcpkg_cmake_configure(
+  SOURCE_PATH "${SOURCE_PATH}"
+  OPTIONS
+    -DSKIP_GLSLANG_INSTALL=OFF
+    -DBUILD_EXTERNAL=OFF
+    -DENABLE_GLSLANG_BINARIES=${BUILD_BINARIES}
+)
+
+vcpkg_cmake_install()
+vcpkg_cmake_config_fixup(CONFIG_PATH lib/cmake)
+
+vcpkg_copy_pdbs()
+
+if(NOT BUILD_BINARIES)
+  file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/bin")
+else()
+  vcpkg_copy_tools(TOOL_NAMES glslangValidator spirv-remap AUTO_CLEAN)
+endif()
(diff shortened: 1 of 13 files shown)


juan-lunarg:
Using `vcpkg` instead of `update_deps.py` closes the aforementioned issues because vcpkg can properly propagate `toolchain` files to dependencies, and properly propagate other CMake state information that is important to dependencies like `CMAKE_OSX_DEPLOYMENT_TARGET`

vcpkg works on all platforms (as long as the dependency uses CMake) since it builds it from scratch.

juan-lunarg:
note: While I raise the cmake min that's for FetchContent and nothing else. Technically I could write around that.

juan-lunarg:
vcpkg also has fantastic binary caching. Save a lot of times on builds.

```
Restored 8 package(s) from /home/juan/.cache/vcpkg/archives in 4.336 s. Use --debug to see more details.
```

cmake/ports/glslang/portfile.cmake
```diff
@@ -0,0 +1,46 @@
...
+vcpkg_cmake_configure(
+  SOURCE_PATH "${SOURCE_PATH}"
+  OPTIONS
+    -DSKIP_GLSLANG_INSTALL=OFF
+    -DBUILD_EXTERNAL=OFF
+    -DENABLE_GLSLANG_BINARIES=${BUILD_BINARIES}
+)
+
+vcpkg_cmake_install()
+vcpkg_cmake_config_fixup(CONFIG_PATH lib/cmake)
```
lunarpapillo, line 31: I don't really belong here, but I'm curious... what does `vcpkg_cmake_config_fixup()` do, and why is it needed?

juan-lunarg: I mainly copied the existing port on this one. But essentially vcpkg has a standard way of laying out it's installs in the cmake binary directory. And from what I can tell the glslang puts things in a slightly different manner.

juan-lunarg:
Closing this draft PR. I'll still leave the branch for reference

PR #4647, opened 2022-10-11
build: Add build-only definitions

ncesario-lunarg:
This is a temporary workaround for building some functions that should not be included in external repos that consume some VVL code.


ncesario-lunarg:
@Tony-LunarG can you check if this addresses your external build issues?

Tony-LunarG:
> @Tony-LunarG can you check if this addresses your external build issues?

No - #include <spirv/unified1/spirv.hpp> is still there.

BUILD.gn
```diff
@@ -51,6 +51,7 @@ config("vulkan_internal_config") {
   defines = [
     "VULKAN_NON_CMAKE_BUILD",
     "VK_ENABLE_BETA_EXTENSIONS",
+    "VVL_BUILD_ONLY_DEFINITIONS",
```
juan-lunarg, line 54: You need Android.mk as well

juan-lunarg (changes requested):

sjfricke (commented):
So in terms of what this is trying to accomplish, it looks good to me, but curious if it would just be easier to move `CompatibleSpirvImageFormat` to `spirv_validation_helper` instead. It already has both the SPIR-V Grammar and VK XML to use... The reason I originally put it in `vk_format_utils` was it seems like the most logical place, but didn't realize adding the `SPIRV-Headers` dependency would cause an issue, granted I feel that it is causing this issue for this PR means other things are in need of fixing too

ncesario-lunarg:
Closing this in favor of #4651.

PR #4648, opened 2022-10-11
Add validation for VK_EXT_pipeline_protected_access

Tony-LunarG:


layers/core_validation.cpp
```diff
@@ -3529,6 +3530,39 @@ bool CoreChecks::ValidatePipeline(std::vector<std::shared_ptr<PIPELINE_STATE>> c
                                          "VkPipelineShaderStageModuleIdentifierCreateInfoEXT and identifierSize not equal to 0",
                                          pipe_index, string_VkGraphicsPipelineLibraryFlagsEXT(lib->graphics_lib_type).c_str());
                     }
+
+                    std::vector<std::pair<VkPipelineCreateFlagBits, std::pair<std::string, std::string>>> test_infos = {
```
jeremyg-lunarg, line 3534: this should be static const, if possible and maybe an array instead of a vector.

jeremyg-lunarg (approved):
LGTM once test_infos is consty

layers/core_validation.cpp
juan-lunarg: constexpr std::array?

layers/core_validation.cpp
```diff
@@ -3529,6 +3530,39 @@ bool CoreChecks::ValidatePipeline(std::vector<std::shared_ptr<PIPELINE_STATE>> c
                                          "VkPipelineShaderStageModuleIdentifierCreateInfoEXT and identifierSize not equal to 0",
                                          pipe_index, string_VkGraphicsPipelineLibraryFlagsEXT(lib->graphics_lib_type).c_str());
                     }
+
+                    static const std::array<std::pair<VkPipelineCreateFlagBits, std::pair<std::string, std::string>>, 2> check_infos =
+                        {std::make_pair(VK_PIPELINE_CREATE_NO_PROTECTED_ACCESS_BIT_EXT,
+                                        std::make_pair("VUID-VkPipelineLibraryCreateInfoKHR-pipeline-07404",
+                                                       "VUID-VkPipelineLibraryCreateInfoKHR-pipeline-07405")),
```
ncesario-lunarg, line 3537: Might be worth creating a struct just above to make this a bit clearer what each field is vs nested pairs, but I'm good either way here.

ncesario-lunarg (approved):
LGTM!

layers/core_validation.cpp
Tony-LunarG: Fixed

layers/core_validation.cpp
Tony-LunarG: Fixed

PR #4650, opened 2022-10-11
build: CMake cleanup

juan-lunarg:
- Specify the languages don't just use the default behavior
- enable_testing called unconditionally
- Use ALL instead of add_dependencies for vvl_update_deps
- Cleanup robinhood code
- Use find_package(glslang) instead of custom code, also move code to tests
- Move code to appropriate directories to avoid cluttering top level CMakeLists.txt
- Remove code that has no reason to exist
- etc.


juan-lunarg:
I'm trying to prepare the repo for vcpkg in the future. Without pulling the trigger this cleans up the codebase that is backwards/future compatible.

layers/generated/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
+# ~~~
```
ncesario-lunarg, line 1: I'm pretty sure _all_ files in `layers/generated` are automatically generated files (save for the `.clang-format` file); can this go into the `cmake` directory?

tests/CMakeLists.txt
```diff
@@ -162,7 +157,14 @@ target_include_directories(vk_layer_validation_tests PRIVATE ${SPIRV_HEADERS_INC
 
 target_link_libraries(vk_layer_validation_tests PRIVATE
     VkLayer_utils
-    ${GLSLANG_LIBRARIES}
+    glslang::glslang
```
ncesario-lunarg, line 160: Great to see these move to targets!

CMakeLists.txt
```diff
@@ -103,10 +99,13 @@ else()
     message("********************************************************************************")
     find_package(PythonInterp 3 QUIET)
 endif()
-#backwards compatability settings for other parts of the build system that don't use find_package() variables yet
+# Dependencies can be installed in arbitrary locations. This is done by update_deps.py / users.
```
ncesario-lunarg, line 102: I've meant to go back and add this as an additional cmake option, but that never happened...

ncesario-lunarg (commented):
One question, but I'd love to see this cleanup get merged. My only concern is is if external dependencies on VVL will be impacted by this, but since CI is passing, it looks promising this is good to go!

CMakeLists.txt
juan-lunarg: No worries, this just really confused my while I was writing my vcpkg experiment

tests/CMakeLists.txt
juan-lunarg: 100% agree

layers/generated/CMakeLists.txt
juan-lunarg: np

jeremyg-lunarg:
> > One question, but I'd love to see this cleanup get merged. My only concern is is if external dependencies on VVL will be impacted by this, but since CI is passing, it looks promising this is good to go!
> 
> My goal with this PR is too maintain back compat, and not break any existing processes while doing cleanup. Anything that causes a breakage will be reverted/fixed.

It might be worthwhile to check that this doesn't break ubuntu or fedora package generation as well (or pull in people to review @johnzupin can probably check ubuntu).   This looks like a great clean up to me though!

juan-lunarg:
> It might be worthwhile to check that this doesn't break ubuntu or fedora package generation as well (or pull in people to review @johnzupin can probably check ubuntu). This looks like a great clean up to me though!

I'm going to run testing to ensure that the SDK package generation isn't affected.

ncesario-lunarg (approved):

juan-lunarg:
Going to have to change the robinhood code slightly since I relied on a 3.11 feature

```
New in version 3.11: Allow setting INTERFACE items on [IMPORTED targets](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#imported-targets).
```

juan-lunarg:
CI is passing. I'll merge if I don't break the SDK build.

ncesario-lunarg:
> Going to have to change the robinhood code slightly since I relied on a 3.11 feature
> 
> ```
> New in version 3.11: Allow setting INTERFACE items on [IMPORTED targets](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#imported-targets).
> ```

I'm tempted to just bump the minimum cmake version and see what happens. @lunarpapillo would appreciate that, I'm sure ;)

PR #4651, opened 2022-10-12
scripts: Move CompatibleSpirvImageFormat

sjfricke:
This "reverts" (really moves) `CompatibleSpirvImageFormat` to not be in `vk_format_utils` this would be a replacement for https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4647 as a "temp fix" until we properly deal with how the format utils folder is being used outside the scope of the VVL 

cc @ncesario-lunarg

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -268,7 +268,6 @@ source_set("vulkan_layer_utils") {
   include_dirs = [
     "layers",
     "layers/generated",
-    "$vvl_spirv_headers_dir/include",
   ]
   sources = [
     "$vulkan_headers_dir/include/vulkan/vk_layer.h",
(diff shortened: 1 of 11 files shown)


sjfricke:
@Tony-LunarG can you check if this addresses your external build issues? (part 2)

ncesario-lunarg (approved):
MUCH nicer, thanks!

ncesario-lunarg:
Merging this for now. We can revert if it doesn't resolve the original issue, but I think this function makes just as much sense inside `spirv_validation_helper` as it does `vk_format_utils`.

Tony-LunarG:
This works for me (and VulkanTools).  Thanks!

PR #4653, opened 2022-10-12
ci: Use MACOS_DEPLOYMENT_TARGET

### LINKED ISSUE #4614: Warnings from macOS build "built for newer macOS version (11.7) than being linked (10.12)"

juan-lunarg:
- This approach matches what our SDK release process does
- CMAKE_MACOS_DEPLOYMENT_TARGET doesn't propagate to dependencies

closes #4614

### DIFF
--- a/.github/workflows/build_macos.yml
+++ b/.github/workflows/build_macos.yml
@@ -55,5 +55,9 @@ jobs:
           key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ matrix.config }}-${{ hashfiles('scripts/known_good.json') }}
       - name: Generate build files and dependencies
         run: cmake -S. -Bbuild -DCMAKE_BUILD_TYPE=${{matrix.config}} -DBUILD_TESTS=ON -DUPDATE_DEPS=ON -DUPDATE_DEPS_SKIP_EXISTING_INSTALL=ON
+        env:
+          # Specify the minimum version of macOS on which the target binaries are to be deployed.
+          # https://cmake.org/cmake/help/latest/envvar/MACOSX_DEPLOYMENT_TARGET.html
+          MACOSX_DEPLOYMENT_TARGET: 10.12
       - name: Build Vulkan-ValidationLayers
         run: cmake --build build --parallel $(sysctl -n hw.logicalcpu) --config ${{matrix.config}}
--- a/BUILD.md
+++ b/BUILD.md
@@ -304,15 +304,6 @@ on/off options currently supported by this repository:
 | BUILD_WSI_WAYLAND_SUPPORT | Linux | `ON` | Build the components with Wayland support. |
 | USE_CCACHE | Linux | `OFF` | Enable caching with the CCache program. |
 
-The following is a table of all string options currently supported by this repository:
-
-| Option | Platform | Default | Description |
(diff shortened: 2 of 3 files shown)


juan-lunarg:
More of an explanation:

Currently we have linker warnings on CI due to a MacOS version mismatch. This was due to `CMAKE_OSX_DEPLOYMENT_TARGET` not getting propagated to dependencies.

The best approach IMO is to use the environment variable `MACOSX_DEPLOYMENT_TARGET` which is what we currently do for our SDK releases.

Note: The linker warnings will still happen for a bit due to CI caching dependencies. But `10.12` will be actively tested for the validation layers which is the point, and when future dependencies are re-built they will use the correct version as well.

juan-lunarg:
It's also worth pointing out by specifying `CMAKE_OSX_DEPLOYMENT_TARGET` in the CMake the environment variable would get ignored. Which is undesirable from the SDK build perspective (I would think).

juan-lunarg:
Linux build failed due to `VkLayerTest.SampledInvalidImageViews`?

juan-lunarg:
Seems re-running CI fixed the spurious failure with `VkLayerTest.SampledInvalidImageViews`

ncesario-lunarg (approved):

PR #4656, opened 2022-10-13
layers: Use GetIndexAlignment everywhere

spencer-lunarg:
Found a place not using `GetIndexAlignment` and went through and updated everyone to use it (and added `VK_INDEX_TYPE_NONE_KHR` support)

Changed from `VkDeviceSize` to `uint32_t` as only one location actually needed `VkDeviceSize` directly


ncesario-lunarg (approved):
LGTM!

spencer-lunarg:
Only test failing in CI is `VkPositiveLayerTest.WaitTimelineSemThreadRace` on Pixel 6 which is a known issue

PR #4663, opened 2022-10-13
syncval: Fix copy/paste error in PreCallRecordCmdExecuteCommands

jeremyg-lunarg:
This caused an assert to fire when running CTS in:
 dEQP-VK.memory.pipeline_barrier.host_write_uniform_texel_buffer.1024
and similar tests.


ncesario-lunarg (approved):

Tony-LunarG (approved):
LGTM and fixes CTS tests

PR #4665, opened 2022-10-13
layers: Fix a vkQueueSubmit()/vkQueueWaitIdle() race

jeremyg-lunarg:
Close a race condition in very common code in the test suite that does:

vkQueueSubmit(..., VK_NULL_HANDLE /*fence*/);
vkQueueWaitIdle(...);

While the thread function is processing a CB_SUBMISSION structure, it needs to stay on the deque so that QUEUE_STATE::Wait() can find it.

### DIFF
--- a/layers/queue_state.cpp
+++ b/layers/queue_state.cpp
@@ -164,9 +164,9 @@ void QUEUE_STATE::Destroy() {
     BASE_NODE::Destroy();
 }
 
-layer_data::optional<CB_SUBMISSION> QUEUE_STATE::NextSubmission() {
-    layer_data::optional<CB_SUBMISSION> result;
-    // Pop the next submission off of the queue so that the thread function doesn't need to worry
+CB_SUBMISSION *QUEUE_STATE::NextSubmission() {
+    CB_SUBMISSION *result = nullptr;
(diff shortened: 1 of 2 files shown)


ncesario-lunarg (approved):

layers/queue_state.cpp
```diff
@@ -164,28 +164,35 @@ void QUEUE_STATE::Destroy() {
...
     auto guard = Lock();
     while (!exit_thread_ && (submissions_.empty() || request_seq_ < submissions_.front().seq)) {
         // The queue thread must wait forever if nothing is happening, until we tell it to exit
         cond_.wait(guard);
     }
     if (!exit_thread_) {
-        result.emplace(std::move(submissions_.front()));
-        submissions_.pop_front();
+        result = &submissions_.front();
+        // NOTE: the submission must remain on the dequeue until we're done processing it so that
```
ziga-lunarg, line 178: Is dequeue meant to be deque?

jeremyg-lunarg: https://en.cppreference.com/w/cpp/container/deque

ziga-lunarg (approved):

PR #4666, opened 2022-10-14
cmake: Fix spirv dependency

juan-lunarg:
Currently builds break when `-D BUILD_LAYERS=OFF -D BUILD_TESTS=OFF`

Because of vvlgeneratedsourcecode depending on spirv/unified1


Tony-LunarG (approved):
LGTM and should fix the VT build

juan-lunarg:
> Because of vvlgeneratedsourcecode depending on spirv/unified1

To be clear this dependency already existed. It just breaks loudly now due to the error handling I added.

PR #4668, opened 2022-10-15
layers: Fix crash when validating GetDrmDisplayEXT

Joshua-Ashton:
This creates the `display` object so it needs to be added to this list.


spencer-lunarg (approved):
LGTM

ncesario-lunarg (approved):

PR #4672, opened 2022-10-18
Issue 4579 - Unconsumed attributes warning&error behavior

arno-lunarg:
Fix for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4579

Completely unconsumed attributes are allowed, and should result in a performance warning if the best practices layer is enabled.
Partially consumed attributes are allowed if `maintenance4` is enabled, but should still result in a performance warning. They are illegal when `maintenance4` is disabled, and result in an error.

Follow up from PR https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4630, had to make some fixes.
From this previous PR, I re-did `ShaderAttribute` struct and (out, in) attributes iteration again, it should be more simple (and more correct)


layers/shader_validation.h
```diff
@@ -39,6 +41,180 @@ struct shader_stage_attributes {
     VkShaderStageFlags stage;
 };
 
+extern const std::array<shader_stage_attributes, 5> shader_stage_attribs;
```
juan-lunarg, line 44: Suggestion feel free to ignore.

Might be cleaner to have this variable as a static member of `ShaderAttribute`. Since it will act as a namespace essentially.

EX:

```
// Feel free to tweak name to reduce redundancy
ShaderAttribute::shader_stage_attribs
```

layers/shader_validation.h
```diff
@@ -39,6 +41,180 @@ struct shader_stage_attributes {
...
+
+struct ShaderAttribute {
```
juan-lunarg, line 46: feel free to ignore

I think it may be cleaner to have `ShaderAttribute` in it's own files. Especially since this header file already defines another class.

ShaderAttribute.h / ShaderAttribute.cpp

juan-lunarg (changes requested):

layers/shader_validation.h
```diff
@@ -39,6 +41,180 @@ struct shader_stage_attributes {
...
+        Uninitialized,
+        Seen,
+        Unseen,
+    };
+
+    ShaderAttribute(const SHADER_MODULE_STATE &shader_module,
+                    const std::map<location_t, interface_var>::const_iterator &shader_attributes_iterator,
+                    const std::map<location_t, interface_var>::const_iterator &shader_attributes_iterator_end)
+        : is_valid(shader_attributes_iterator != shader_attributes_iterator_end) {
+        if (is_valid) {
```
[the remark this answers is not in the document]
juan-lunarg: This is a question, since I don't fully understand the context. It may be the correct approach. It's just counter intuitive to the regular RAII approach used with classes in C++.

layers/best_practices_utils.cpp
```diff
@@ -1095,6 +1095,43 @@ void BestPractices::ManualPostCallRecordCreateComputePipelines(VkDevice device,
...
+                                                   const SHADER_MODULE_STATE& consumer, spirv_inst_iter consumer_entrypoint,
+                                                   shader_stage_attributes const* consumer_stage) const {
+    bool skip = false;
+
+    auto check_unconsumed_attrib = [this](const SHADER_MODULE_STATE& producer, shader_stage_attributes const* producer_stage,
+                                          const ShaderAttribute& out, const SHADER_MODULE_STATE& consumer,
+                                          shader_stage_attributes const* consumer_stage, const ShaderAttribute& in) -> bool {
+        bool skip = false;
+        if (out.IsScanCompleted() &&
+            ((enabled_features.core13.maintenance4 && out.IsOnlyPartiallyConsumed()) || out.IsFullyUnconsumed())) {
```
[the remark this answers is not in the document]
arno-lunarg: `out` is one the lambda's parameter!

layers/shader_validation.h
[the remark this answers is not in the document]
arno-lunarg: Incrementing an iterator in `IterateInterfaceBetweenStages` looks like this:
```cpp
    const auto increment_out_iterator = [&]() {
        if (out_iterator == shader_outputs.end()) return;
        ++out_iterator;
        out = ShaderAttribute(producer, out_iterator, shader_outputs.end());
        if (out.DoesOverflowOnNextLocation()) {
            assert(out_iterator != shader_outputs.end());
            ++out_iterator;
        }
    };
```

But the iterator given to `ShaderAttribute` can be invalid, thus the check. The validity of `ShaderAttribute` is checked when comparing it to another one, here when comparing `out` and `in`. This logic happens inside the `ShaderAttribute` class, I found that it made the code in `IterateInterfaceBetweenStages` more readable.
I could try to always pass a valid iterator, but I think it would make `IterateInterfaceBetweenStages` messier?

layers/shader_validation.h
arno-lunarg: yep it would be better!

layers/shader_validation.h
arno-lunarg: yes I think you are right... And the name too, `ShaderAttribute` still not sure about it, since this class was made to help check shader attributes, it sort of wraps parts of a shader attribute, plus things related to performing the check, so I should find another name.

layers/best_practices_utils.cpp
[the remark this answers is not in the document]
juan-lunarg: Oh sorry I read this wrong...

layers/shader_validation.h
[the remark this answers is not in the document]
juan-lunarg: Gotcha, that makes sense. I'll mark this as resolved.

layers/shader_validation.h
arno-lunarg: In the end I moved it in the only function it is used, `ValidateInterfaceBetweenAllPipelineStages`

layers/shader_validation.h
arno-lunarg: I Added `shader_attribute_validation.h`, and renamed `ShaderAttribute` to `ShaderAttributeValidation`

layers/best_practices_validation.h
```diff
@@ -23,6 +23,7 @@
 
 #include "chassis.h"
 #include "state_tracker.h"
+#include "shader_validation.h"
```
juan-lunarg, line 26: Since everything is passed by pointer or reference you can just forward declare instead of including.

layers/shader_attribute_validation.h
```diff
@@ -0,0 +1,308 @@
...
+        Uninitialized,
+        Seen,
+        Unseen,
+    };
+
+    ShaderAttributeValidation(const SHADER_MODULE_STATE &shader_module,
+                              const std::map<location_t, interface_var>::const_iterator &shader_attributes_iterator,
+                              const std::map<location_t, interface_var>::const_iterator &shader_attributes_iterator_end)
+        : is_valid_(shader_attributes_iterator != shader_attributes_iterator_end) {
+        if (is_valid_) {
```
juan-lunarg, line 41: Return early to avoid indenting the majority of the logic.

```suggestion
        if (!is_valid_) { return; }
```

layers/shader_attribute_validation.h
```diff
@@ -0,0 +1,308 @@
...
+        }
+    }
+
+    static void TagMatchingComponentsAsSeen(ShaderAttributeValidation &lhs, ShaderAttributeValidation &rhs) {
+        const sparse_container::range<uint32_t> lhs_range(lhs.current_global_component_,
+                                                          lhs.current_global_component_ + lhs.current_components_left_);
+        const sparse_container::range<uint32_t> rhs_range(rhs.current_global_component_,
+                                                          rhs.current_global_component_ + rhs.current_components_left_);
+        const sparse_container::range<uint32_t> range_intersection = lhs_range & rhs_range;
+        if (range_intersection.non_empty()) {
```
juan-lunarg, line 74: Return early to avoid indenting the majority of the logic.

```suggestion
        if (range_intersection.empty()) { return; }
```

layers/shader_attribute_validation.h
```diff
@@ -0,0 +1,308 @@
...
+    uint32_t current_components_left_ = 0;
+    uint32_t components_count_ = 0;
+    std::array<ComponentStatus, 16> components_ = {{ComponentStatus::Uninitialized}};
+    interface_var const *interface_ = nullptr;
+};
+
+template <typename PipeStageState, typename PipeStageStateVec, typename ValidateInterfaceBetweenStagesFunc>
+bool ValidateInterfaceBetweenAllPipelineStages(const PipeStageStateVec &pipeline_stage_states, const PipeStageState *fragment_stage,
+                                               const ValidateInterfaceBetweenStagesFunc &validation_func) {
+    static const std::array<shader_stage_attributes, 5> shader_stage_attribs = {
```
juan-lunarg, line 208: nitpick

```suggestion
    constexpr std::array<shader_stage_attributes, 5> shader_stage_attribs = {
```

juan-lunarg (changes requested):

layers/shader_attribute_validation.h
```diff
@@ -0,0 +1,308 @@
...
+
+  private:
+    bool is_valid_ = false;
+    bool is_scan_completed_ = false;
+    uint32_t start_location_ = 0;
+    uint32_t start_component_ = 0;
+    uint32_t current_global_component_ = 0;
+    uint32_t current_components_left_ = 0;
+    uint32_t components_count_ = 0;
+    std::array<ComponentStatus, 16> components_ = {{ComponentStatus::Uninitialized}};
```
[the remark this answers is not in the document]
arno-lunarg: a `mat4` occupies 16 components, or 4 components on 4 consecutive locations. But writing that, I now realise that a `dmat4` would occupy 32 components, each scalar occupying 2 components... need to have a look at this. And will add a comment regarding where this array's size comes from

layers/shader_attribute_validation.h
```diff
@@ -0,0 +1,308 @@
...
+
+#pragma once
+
+#include <cstdlib>
+#include <array>
+
+#include "shader_module.h"
+#include "range_vector.h"
+
+class ShaderAttributeValidation {
```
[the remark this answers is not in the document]
arno-lunarg: I think this small is too small for the decrease in compile time to be a problem, and with the present form it gives the compiler the choice of inlining or not. So I think it is ok to keep it as an header only. But maybe the other reviewers have another opinion!

layers/shader_attribute_validation.h
```diff
@@ -0,0 +1,308 @@
...
+    ShaderAttributeValidation out(producer, out_iterator, shader_outputs.end());
+    if (out.DoesOverflowOnNextLocation()) {
+        assert(out_iterator != shader_outputs.end());
+        ++out_iterator;
+    }
+    ShaderAttributeValidation in(consumer, in_iterator, shader_inputs.end());
+    if (in.DoesOverflowOnNextLocation()) {
+        assert(in_iterator != shader_inputs.end());
+        ++in_iterator;
+    }
```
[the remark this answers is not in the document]
arno-lunarg: I added a comment on `DoesOverflowOnNextLocation` to explain the method:
```cpp
// True for vector types with more than 2 components and 64 bits scalar type (dvec3, dvec4, ...)
    bool DoesOverflowOnNextLocation() const { return components_count_ > 4; }
```
But on this specific block, I need to add another comment, stating why I need to increment iterators if this is true: in that case, tne next iterator holds the same info as the current one but with an incremented `location`. Since it is still the same out/in attribute, I skip it.

layers/best_practices_validation.h
arno-lunarg: I think that according to the style guide, I should keep the include: https://google.github.io/styleguide/cppguide.html#Forward_Declarations

layers/shader_attribute_validation.h
arno-lunarg: yes it looks better this way!

layers/shader_attribute_validation.h
arno-lunarg: I think I still need the `static` keyword here? Does `constexpr` implies `static`?

juan-lunarg: After checking godbolt and cppreference, constexpr doesn't imply static

https://godbolt.org/z/G7Wb35594

You still need the static keyword.

layers/best_practices_validation.h
juan-lunarg: Ah. I was looking at a different style guide.

layers/shader_attribute_validation.h
juan-lunarg: FFTI this request. I thought constexpr did imply static. I learned something new.

layers/shader_attribute_validation.h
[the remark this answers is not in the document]
arno-lunarg: Actually, it seems that I only need a size of 8. Each column of a matrix occupies its own location, so just need a max of 8 to be able to fit a `dvec4`

layers/shader_attribute_validation.h
```diff
@@ -0,0 +1,322 @@
...
+
+  private:
+    bool is_valid_ = false;
+    bool is_scan_completed_ = false;
+    uint32_t start_location_ = 0;
+    uint32_t start_component_ = 0;
+    uint32_t current_global_component_ = 0;
+    uint32_t current_components_left_ = 0;
+    uint32_t components_count_ = 0;
+    std::array<ComponentStatus, 8> components_ = {{ComponentStatus::Uninitialized}};
```
juan-lunarg, line 200: Add comment explaining size 8

juan-lunarg (approved):
1 minor nitpick other than that LGTM

layers/shader_attribute_validation.h
arno-lunarg: oopsy forgot to add it

layers/shader_attribute_validation.h
```diff
@@ -0,0 +1,324 @@
...
+
+#pragma once
+
+#include <cstdlib>
+#include <array>
+
+#include "shader_module.h"
+#include "range_vector.h"
+
+class ShaderAttributeValidation {
```
spencer-lunarg, line 29: It takes quit a bit to fully understand what this class/file is actually for. 

Also I think `attribute` is the wrong name for this. In Vulkan an `attribute` is only for the input of the vertex shader

layers/shader_module.h
```diff
@@ -284,6 +284,11 @@ struct SHADER_MODULE_STATE : public BASE_NODE {
 
     SHADER_MODULE_STATE() : BASE_NODE(static_cast<VkShaderModule>(VK_NULL_HANDLE), kVulkanObjectTypeShaderModule) {}
 
+    static uint32_t GetShaderStageId(VkShaderStageFlagBits stage) {
```
spencer-lunarg, line 287: I don't think this belongs in `SHADER_MODULE_STATE`, I didn't see the issue of keeping it in the .cpp file as it was

layers/shader_module.cpp
```diff
@@ -1848,6 +1848,23 @@ uint32_t SHADER_MODULE_STATE::GetTypeBitsSize(const Instruction* insn) const {
 
 uint32_t SHADER_MODULE_STATE::GetTypeBytesSize(const Instruction* insn) const { return GetTypeBitsSize(insn) / 8; }
 
+uint32_t SHADER_MODULE_STATE::GetArraySize(const Instruction* insn) const {
+    const uint32_t opcode = insn->Opcode();
+    if ( opcode == spv::OpTypePointer )
```
spencer-lunarg, line 1853: nit: This ran with clang-format? not sure how this is passing

tests/vklayertests_pipeline_shader.cpp
```diff
@@ -14183,8 +14207,7 @@ TEST_F(VkLayerTest, TestInvalidShaderInputAndOutputComponents) {
         const auto set_info = [&](CreatePipelineHelper &helper) {
             helper.shader_stages_ = {vs.GetStageCreateInfo(), fs.GetStageCreateInfo()};
         };
-        CreatePipelineHelper::OneshotTest(*this, set_info, kPerformanceWarningBit,
-                                          "UNASSIGNED-CoreValidation-Shader-OutputNotConsumed");
+        CreatePipelineHelper::OneshotTest(*this, set_info, kErrorBit, "VUID-RuntimeSpirv-OpTypeVector");
```
spencer-lunarg, line 14210: this needs to be `VUID-RuntimeSpirv-OpTypeVector-06816` (also you have `VUID-RuntimeSpirv-OpTypeVector-06817` but it is `VUID-RuntimeSpirv-maintenance4-06817`)

layers/shader_validation.cpp
```diff
@@ -3141,108 +3129,83 @@ bool CoreChecks::ValidateInterfaceBetweenStages(const SHADER_MODULE_STATE &produ
...
-            } else {
-                input_component++;
+                const auto &out_components = out.GetComponents();
+                for (uint32_t component_i = 0; component_i < out.GetComponentsCount(); component_i++) {
+                    if (out_components[component_i] == ShaderAttributeValidation::ComponentStatus::Unseen) {
+                        const uint32_t out_location = out.LocationFromComponentsIndex(component_i);
+                        const uint32_t out_component = out.ComponentFromComponentsIndex(component_i);
+
+                        std::string msg =
+                            std::string{producer_stage->name} + " writes to output location " + std::to_string(out_location) + '.' +
```
spencer-lunarg, line 3147: This is not what `VUID-RuntimeSpirv-maintenance4-06817` is checking

> If [maintenance4](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#features-maintenance4) is not enabled, any OpTypeVector output interface variables must not have a higher Component Count than a matching OpTypeVector input interface variable

This is VU is suppose to be used to check if the vertex shader writes `vec4` but fragment shader reads `vec3` at the same `Location` and nothing to do with being consumed

layers/shader_validation.cpp
```diff
@@ -3141,108 +3129,83 @@ bool CoreChecks::ValidateInterfaceBetweenStages(const SHADER_MODULE_STATE &produ
                                                 shader_stage_attributes const *consumer_stage) const {
     bool skip = false;
 
-    auto outputs =
-        producer.CollectInterfaceByLocation(producer_entrypoint, spv::StorageClassOutput, producer_stage->arrayed_output);
-    auto inputs = consumer.CollectInterfaceByLocation(consumer_entrypoint, spv::StorageClassInput, consumer_stage->arrayed_input);
-
-    auto output_it = outputs.begin();
-    auto input_it = inputs.begin();
+    auto validate_attrib = [this](const SHADER_MODULE_STATE &producer, shader_stage_attributes const *producer_stage,
```
spencer-lunarg, line 3132: what is the advantage of doing this as a lambda here and not having this just be a separate function?

layers/shader_attribute_validation.cpp
```diff
@@ -0,0 +1,68 @@
...
+    const SHADER_MODULE_STATE &shader_module, const std::map<location_t, interface_var>::const_iterator &shader_attributes_iterator,
+    const std::map<location_t, interface_var>::const_iterator &shader_attributes_iterator_end)
+    : is_valid_(shader_attributes_iterator != shader_attributes_iterator_end) {
+    if (!is_valid_) return;
+
+    start_location_ = shader_attributes_iterator->first.first;
+    start_component_ = shader_attributes_iterator->first.second;
+    current_global_component_ = start_location_ * 4 + start_component_;
+    const auto def = shader_module.FindDef(shader_attributes_iterator->second.type_id);
+    const auto base_type_components_count = shader_module.GetNumComponentsInBaseType(def);
```
arno-lunarg, line 15: chat with spencer

layers/shader_validation.cpp
arno-lunarg: I need to pass this function to `IterateInterfaceBetweenStages`, and since it calls methods on a `CoreChecks` object, better to capture `this` using a lambda

layers/shader_module.cpp
arno-lunarg: Yes I am surprise it ran without issue... Will do the change manually

layers/shader_module.h
arno-lunarg: This function is now used in `ValidateInterfaceBetweenAllPipelineStages` (`shader_attribute_validation.h`), it need to be moved somewhere. I am open to a better location though!

layers/shader_attribute_validation.h
arno-lunarg: I do not like this name either, and I am open to suggestions. Reading [the glsl spec](https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.4.60.pdf), page 51:
> It is expected that graphics hardware will have a small number of fixed vector locations for passing
vertex inputs. Therefore, the OpenGL Shading Language defines each non-matrix input variable as
taking up one such vector location.

So I could maybe rename it to `VectorLocationValidation`? Add still add a comment

arno-lunarg: I think `InterfaceLocationValidation` better captures the intent, and the names of the 2 other functions in this .h

layers/shader_validation.cpp
```diff
@@ -3141,108 +3129,111 @@ bool CoreChecks::ValidateInterfaceBetweenStages(const SHADER_MODULE_STATE &produ
...
+// layout(location = 0, component = 0) in float y1;
+// layout(location = 0, component = 3) in float y3;
+// Or:
+// // VS
+// layout(location = 0) out dvec4 x;
+// // FS
+// layout(location = 0, component = 1) flat in float hello_there;
+//
+// Until the issue is resolved, the following code should be inactive
+#if 0
```
arno-lunarg, line 3164: Until maintenance4 description is clarified, this part is disabled

arno-lunarg: @ncesario-lunarg and @spencer-lunarg for visibility

spencer-lunarg:
This will need a LOT of rebasing with new shader module refactoring. Going to close this as it has got stale, tracking issue still has reference for when we need to find this code

PR #4675, opened 2022-10-18
build: Use -Wpedantic

juan-lunarg:
Fixes minor ISO compliance
- Semicolon at end of func definition
- Using C++14 binary literals
- Using C++14 default parameter values


juan-lunarg:
This PR makes me think it's safe to update to C++14, but technically we do promise only needing C++11.

Let me know what you think.

PR #4677, opened 2022-10-19
layers: Generate CB dynamic status flags

spencer-lunarg:
`CBStatus` and `CBStatusFlags` are really only for dynamic status (after we pulled out index buffer from it which didn't belong)

This renames them to `CB_DYNAMIC_STATUS` and `CBDynamicFlag` and generates them (and the helper functions)

This exposed a few incorrect names (ex `CBSTATUS_STENCIL_READ_MASK_SET` should have been `CBSTATUS_STENCIL_COMPARE_MASK_SET`)

There are now a LOT of dynamic states and I plan to move more to generation after as this will prevent the copy and paste prone errors in the future


layers/generated/command_validation.h
```diff
@@ -479,3 +480,81 @@ static const std::array<const char *, CMD_RANGE_SIZE> kGeneratedCommandNameList
     "vkCmdWriteTimestamp2",
     "vkCmdWriteTimestamp2KHR",
 }};
+
+typedef enum CB_DYNAMIC_STATUS {
```
jeremyg-lunarg, line 484: Should this enum get renamed as well?

Should the values get renamed to match https://google.github.io/styleguide/cppguide.html#Enumerator_Names ?

jeremyg-lunarg (approved):
LGTM. I was hoping this would happen, but didn't think it would happen this fast. :) Only question is if we should take this opportunity to move closer to the coding standard.

Tony-LunarG:
I think this should probably be picked into the SDK branch, as it fixes a CTS crash when an unrecognized dynamic state enum is used in a CBStatusFlags bitset

jeremyg-lunarg:
This fixes a crash in CTS. Merging now to get it into the SDK.

layers/generated/command_validation.h
ncesario-lunarg: :heavy_plus_sign: let's get this in post-SDK.

layers/generated/command_validation.h
```diff
@@ -479,3 +480,81 @@ static const std::array<const char *, CMD_RANGE_SIZE> kGeneratedCommandNameList
...
+    CB_DYNAMIC_COVERAGE_MODULATION_MODE_NV_SET = 63,
+    CB_DYNAMIC_COVERAGE_MODULATION_TABLE_ENABLE_NV_SET = 64,
+    CB_DYNAMIC_COVERAGE_MODULATION_TABLE_NV_SET = 65,
+    CB_DYNAMIC_SHADING_RATE_IMAGE_ENABLE_NV_SET = 66,
+    CB_DYNAMIC_REPRESENTATIVE_FRAGMENT_TEST_ENABLE_NV_SET = 67,
+    CB_DYNAMIC_COVERAGE_REDUCTION_MODE_NV_SET = 68,
+    CB_DYNAMIC_STATUS_NUM = 69
+} CB_DYNAMIC_STATUS;
+
+typedef std::bitset<CB_DYNAMIC_STATUS_NUM> CBDynamicFlags;
```
ncesario-lunarg, line 556: If we're "standardizing," aliases should be `using <name> = <value>;` (https://google.github.io/styleguide/cppguide.html#Aliases).

layers/generated/command_validation.h
spencer-lunarg: I have more in store to generate that would use this, happy to rename there!

PR #4683, opened 2022-10-19
corechecks: Add execution-time checking timeline semaphores

jeremyg-lunarg:
There are requirements on the signal value for timeline semaphores that can only be fully checked at execution time. Now that QUEUE_STATE::ThreadFunc provides a location where we know execution dependencies for each submission have completed, we can check it there.
    
This affects the following VUIDs:
    
    VUID-VkBindSparseInfo-pSignalSemaphores-03249
    VUID-VkSubmitInfo-pSignalSemaphores-03242
    VUID-VkSubmitInfo2-semaphore-03882
    
It is important to note that execution time checking cannot block execution of invalid signal operations. To do that we'd have to we'd have to defer queue submissions until all of their dependencies were known to have completed. This would be very tricky and probably slow.
   
There are quite a few other VUIDs related to image layouts, swapchains, and other things that can only be accurately checked this way. They'll be added in future PRs.

### DIFF
--- a/layers/core_checks/core_validation.h
+++ b/layers/core_checks/core_validation.h
@@ -216,6 +216,7 @@ struct SubresourceRangeErrorCodes {
 
 typedef vvl::unordered_map<const IMAGE_STATE*, std::optional<GlobalImageLayoutRangeMap>> GlobalImageLayoutMap;
 
+namespace core_state {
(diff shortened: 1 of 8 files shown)


juan-lunarg:
Closing due to age of draft PR and lack of recent activity

PR #4685, opened 2022-10-19
build: Use C++17 standard

ncesario-lunarg:
@juan-lunarg showed that we are already using C++14 features, so making the minimum explicit.

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -64,6 +64,12 @@ config("vulkan_internal_config") {
     "VK_ENABLE_BETA_EXTENSIONS",
   ]
 
+  if (!is_win) {
+    cflags_cc = [ "-std=c++17" ]
+  } else {
+    cflags_cc = [ "/std:c++17" ]
+  }
+
   cflags = []
   if (is_clang || !is_win) {
     cflags += [ "-Wno-unused-function" ]
--- a/BUILD.md
+++ b/BUILD.md
@@ -530,8 +530,8 @@ layer validation tests use the "googletest" testing framework.)
 ### Linux Build Requirements
 
 This repository has been built and tested on the two most recent Ubuntu LTS
-versions. Currently, the oldest supported version is Ubuntu 16.04, meaning
-that the minimum officially supported C++17 compiler version is GCC 5.4.0,
+versions. Currently, the oldest supported version is Ubuntu 18.04, meaning
+that the minimum officially supported C++17 compiler version is GCC 7.3.0,
 although earlier versions may work. It should be straightforward to adapt this
 repository to other Linux distributions.
 
--- a/BUILD.md
+++ b/BUILD.md
@@ -16,7 +16,7 @@
 
 1. Python >= 3.7 (3.6 may work, 3.5 and earlier is not supported)
 1. CMake >= 3.10.2
-1. C++ >= c++11 compiler. See platform-specific sections below for supported compiler versions.
+1. C++ >= c++17 compiler. See platform-specific sections below for supported compiler versions.
 
 ## Building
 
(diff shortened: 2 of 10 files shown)


juan-lunarg:
> CI Vulkan-ValidationLayers build # 9100 passed.

Probably a good idea to update build.gn and the android build as well

ncesario-lunarg:
> > CI Vulkan-ValidationLayers build # 9100 passed.
> 
> Probably a good idea to update build.gn and the android build as well

I don't see where to make this change, but that is probably just due to my ignorance with GN.

juan-lunarg:
> I don't see where to make this change, but that is probably just due to my ignorance with GN.

Yeah, I was a bit confused by that as well.

juan-lunarg:
The GN build has a minor issue:
```
-std=c++17 -Wno-undefined-bool-conversion -Wno-tautological-undefined-compare -std=c++14
```

Looks like -std=c++14 is overriding the -std=c++17 you set

BUILD.gn
```diff
@@ -86,6 +86,13 @@ config("vulkan_internal_config") {
   if (is_clang) {
     cflags += [ "-Wno-extra-semi" ]
   }
+
+  # TODO is there a more "standard" way to set this?
+  if (!is_win) {
+    cflags += [ "-std=c++17" ]
+  } else {
+    cflags += [ "/std:c++17" ]
+  }
```
juan-lunarg, line 95: I found an example for setting the C++ standard:
https://fuchsia.googlesource.com/fuchsia/+/refs/heads/main/build/config/BUILD.gn

Looks like there is a kinda standard approach. Couldn't find anything in the docs.

```
import("//build_overrides/vulkan_validation_layers.gni")

config("language") {
  if (!is_win) {
    cflags_cc += [ "-std=c++17" ]
  } else {
    cflags_cc += [ "/std:c++17" ]
  }
}
```

juan-lunarg: Note you want to use the `cflags_cc` variable otherwise you'll affect the compilation of `xxhash` which is a c library

juan-lunarg: Hmm, now I'm not so sure after looking at this chromium build.
https://github.com/timniederhausen/gn-build/blob/923683700a39a53ecf467974cbf29f3405358292/config/BUILD.gn#L807

juan-lunarg: Looks like the chromium builds don't have  a standardized approach like `CXX_STANDARD` in CMake. So each repo makes up their own approach.

ncesario-lunarg: :+1: thanks! I can cherry pick the changes you made in #4791 here.

The internal build failure looks spurious.

juan-lunarg: I wouldn't cherry pick yet since my change failed to build.

juan-lunarg: Latest version of my PR appears to pass.

BUILD.md
```diff
@@ -531,7 +531,7 @@ layer validation tests use the "googletest" testing framework.)
 
 This repository has been built and tested on the two most recent Ubuntu LTS
 versions. Currently, the oldest supported version is Ubuntu 16.04, meaning
-that the minimum officially supported C++11 compiler version is GCC 5.4.0,
+that the minimum officially supported C++17 compiler version is GCC 5.4.0,
```
juan-lunarg, line 534: The Ubuntu version and GCC version don't seem correct.

juan-lunarg (changes requested):

BUILD.md
juan-lunarg, line 534: Min Ubuntu version should be 18.04, not sure what the min GCC version should be.

juan-lunarg (approved):

Issue #4689, opened 2022-10-21
UBSAN error in PIPELINE_STATE::GetStageStates


jbauman42:
**Describe the Issue**
At https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/layers/pipeline_state.cpp#L86, (1<<31) is cast to a VkShaderStageFlagBits and loaded. That value is > VK_SHADER_STAGE_FLAG_BITS_MAX_ENUM, so that's undefined behavior, which is caught by ubsan.

**Environment:**
 - OS: Fuchsia
 - GPU: Intel
 - SDK or header version if building from repo: 1.3.231
 - Options enabled (synchronization, best practices, etc.): default

**Additional context**
Callstack (as a Fuchsia log):
[00031.211991][58875][58877][klog] INFO: [1m../../third_party/Vulkan-ValidationLayers/layers/pipeline_state.cpp:86:39:[1m[31m runtime error: [1m[0m[1mload of value
[00031.211993][58875][58877][klog] INFO:  2147483648, which is not a valid value for type 'const VkShaderStageFlagBits'[1m[0m
[00031.657452][58875][58877][klog] INFO:    #0    0x000021abf7484393 in PIPELINE_STATE::GetStageStates(const ValidationStateTracker&, const PIPELINE_STATE&) ../../third_party/Vulkan-ValidationLayers/layers/pipeline_state.cpp:86 <VkLayer_khronos_validation.so>+0x295a393
[00031.657500][58875][58877][klog] INFO:    #1.2  0x0000208d349d83ff in ubsan_GetStackTrace() compiler-rt/lib/ubsan/ubsan_diag.cpp:41 <libclang_rt.asan.so>+0x3d3ff
[00031.657500][58875][58877][klog] INFO:    #1.1  0x0000208d349d83ff in MaybePrintStackTrace() compiler-rt/lib/ubsan/ubsan_diag.cpp:51 <libclang_rt.asan.so>+0x3d3ff
[00031.657500][58875][58877][klog] INFO:    #1    0x0000208d349d83ff in ~ScopedReport() compiler-rt/lib/ubsan/ubsan_diag.cpp:387 <libclang_rt.asan.so>+0x3d3ff
[...]

PR #4690, opened 2022-10-21
layers: Fix PIPELINE_STATE::GetStageStates bug

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4689

### DIFF
--- a/layers/pipeline_state.cpp
+++ b/layers/pipeline_state.cpp
@@ -73,13 +73,14 @@ PipelineStageState::PipelineStageState(const safe_VkPipelineShaderStageCreateInf
       has_input_attachment_capability(module_state->HasInputAttachmentCapability()) {}
 
 // static
-PIPELINE_STATE::StageStateVec PIPELINE_STATE::GetStageStates(const ValidationStateTracker &state_data,
-                                                             const PIPELINE_STATE &pipe_state) {
-    PIPELINE_STATE::StageStateVec stage_states;
+PIPELINE_STATE::StageStateList PIPELINE_STATE::GetStageStates(const ValidationStateTracker &state_data,
+                                                              const PIPELINE_STATE &pipe_state) {
+    PIPELINE_STATE::StageStateList stage_states;
     // shader stages need to be recorded in pipeline order
     const auto stages = pipe_state.GetShaderStages();
 
-    for (uint32_t stage_idx = 0; stage_idx < 32; ++stage_idx) {
+    // (1 << 31) is greater than VK_SHADER_STAGE_FLAG_BITS_MAX_ENUM
+    for (uint32_t stage_idx = 0; stage_idx < 31; ++stage_idx) {
         bool stage_found = false;
         const auto stage = static_cast<VkShaderStageFlagBits>(1 << stage_idx);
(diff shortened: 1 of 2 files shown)


layers/pipeline_state.cpp
```diff
@@ -73,13 +73,14 @@ PipelineStageState::PipelineStageState(const safe_VkPipelineShaderStageCreateInf
...
-    PIPELINE_STATE::StageStateVec stage_states;
+PIPELINE_STATE::StageStateList PIPELINE_STATE::GetStageStates(const ValidationStateTracker &state_data,
+                                                              const PIPELINE_STATE &pipe_state) {
+    PIPELINE_STATE::StageStateList stage_states;
     // shader stages need to be recorded in pipeline order
     const auto stages = pipe_state.GetShaderStages();
 
-    for (uint32_t stage_idx = 0; stage_idx < 32; ++stage_idx) {
+    // (1 << 31) is greater than VK_SHADER_STAGE_FLAG_BITS_MAX_ENUM
+    for (uint32_t stage_idx = 0; stage_idx < 31; ++stage_idx) {
```
ncesario-lunarg, line 83: Does this actually fix the original issue? I'm not sure I see what changed here. I would think we would need something like `stage_idx = highest_bit(VK_SHADER_STAGE_FLAG_BITS_MAX_ENUM);` to make UBSAN happy here, but I'm probably missing something.

ncesario-lunarg (commented):
If this fixes the original issue, then LGTM, but I'm missing what actually changed.

Also, what is the rationale behind the rename from `Vec` -> `List`? Not necessarily arguing for or against it, just curious :)

layers/pipeline_state.cpp
spencer-lunarg: So this code logic needs to loop through all the `VkShaderStageFlagBits` values

`VK_SHADER_STAGE_FLAG_BITS_MAX_ENUM == 0x7FFFFFFF`

so I assumed the `1 << 30` is `0x40000000` which is the max any valid Vulkan FlagBits can be. I assume this change would correctly fix UBSAN, but honestly I would rather have a way to loop through all the `VkShaderStageFlagBits`, I was thinking how we have these vectors in `paramater_validation.cpp` such as `const std::vector<VkFormat> AllVkFormatEnums` and that maybe it would be smarter to just generate something similar for this code instead

> rationale behind the rename

I didn't know what `Vec` stood for, wasn't an intuitive abbreviation for me, but now realizing we use `*Vec` else where, rather not inconsistently change it here

spencer-lunarg:
closing for a better solution

layers/pipeline_state.cpp
ncesario-lunarg: Ah, I completely missed the 32 -> 31. I'm good with this if it fixes the original issue. The comment wasn't intuitive to me the first time I read it, but I probably just read it too quickly.

spencer-lunarg: the fact you can't at a glance know if it is correct or not **is** the problem IMO, going to just generate an array to loop through stages flags

PR #4695, opened 2022-10-22
layers: Use array for enum list

spencer-lunarg:
I realized we generate a `std::vector` for what is a known length

Also the vector were using the aliased values so there were extra enums in them not needed


layers/stateless_validation.h
```diff
@@ -53,18 +53,18 @@ extern const VkGeometryFlagsKHR AllVkGeometryFlagBitsKHR;
 extern const VkPipelineColorBlendStateCreateFlags AllVkPipelineColorBlendStateCreateFlagBits;
 extern const VkPipelineDepthStencilStateCreateFlags AllVkPipelineDepthStencilStateCreateFlagBits;
 
-extern const std::vector<VkGeometryTypeKHR> AllVkGeometryTypeKHREnums;
```
spencer-lunarg, line 56: This is the part I am not a fan of and would not merge as is

I rather not make this "yet another" place to update on spec changes, but I couldn't get around how we have to `extern` these because of the nature of `parameter_validation_utils`

ncesario-lunarg: If we're going to change these (and might as well do the flag bits above if it makes sense), I _think_ I'd vote for making them static inside a struct (don't think it will work outside of a struct/class, but I don't know for sure), and putting in a separate header. So something like:

`layers/generated/enum_flag_bits.h`:
```cpp
struct AllValues {
    static const std::array<VkGeometryTypeKHR, 3> GeometryTypeKHREnums;
    ...
};
```

`layers/generated/parameter_validation.cpp` (and other files that depend on these values):
```cpp
#include "enum_flag_bits.h"

// s/AllVk/AllValues::/g <- modulo special cases...
```

layers/stateless_validation.h
```diff
@@ -701,9 +701,9 @@ class StatelessValidation : public ValidationObject {
      * @param value Enumeration value to validate.
      * @return Boolean value indicating that the call should be skipped.
      */
-    template <typename T>
+    template <typename T, size_t N>
```
ncesario-lunarg, line 704: Since `value_type` is baked into `std::array`, I don't think the additional template parameter is needed, and the parameter types can change to:

`std::array<T, N>` -> `T`

`T` -> `typename T::value_type`

ncesario-lunarg (commented):
I like where this is going...especially if it removes those `extern` variables :)

layers/stateless_validation.h
spencer-lunarg: seems I didn't need the `struct AllValues`

juan-lunarg:
@spencer-lunarg looks like this needs to be rebased on master due to merge conflicts

juan-lunarg (approved):
LGTM

PR #4698, opened 2022-10-23
tests: Pixel6 failure tests fixed

spencer-lunarg:
Part of https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4425

Should fix

- VkArmBestPracticesLayerTest.RedundantRenderPassStore
- VkLayerTest.InvalidMixingProtectedResources
- VkLayerTest.CreateImageMaxLimitsViolation

These were cases where the test logic was wrong

### DIFF
--- a/tests/layer_validation_tests.cpp
+++ b/tests/layer_validation_tests.cpp
@@ -1292,7 +1292,9 @@ VkRenderPass VkArmBestPracticesLayerTest::CreateRenderPass(VkFormat format, VkAt
     rpinf.dependencyCount = 0;
     rpinf.pDependencies = nullptr;
 
-    vk::CreateRenderPass(m_device->handle(), &rpinf, nullptr, &renderpass);
+    VkResult result = vk::CreateRenderPass(m_device->handle(), &rpinf, nullptr, &renderpass);
+    assert(result == VK_SUCCESS);
+    (void)result;
 
     return renderpass;
 }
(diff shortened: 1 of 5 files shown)


tests/layer_validation_tests.cpp
```diff
@@ -1309,7 +1310,8 @@ VkFramebuffer VkArmBestPracticesLayerTest::CreateFramebuffer(const uint32_t widt
     framebuffer_create_info.height = height;
     framebuffer_create_info.layers = 1;
 
-    vk::CreateFramebuffer(m_device->handle(), &framebuffer_create_info, nullptr, &framebuffer);
+    VkResult result = vk::CreateFramebuffer(m_device->handle(), &framebuffer_create_info, nullptr, &framebuffer);
```
spencer-lunarg, line 1313: This was failing because it was hitting another best practice warning, this assert makes it easier to spot where the issue was

tests/vklayertests_buffer_image_memory_sampler.cpp
```diff
@@ -9565,7 +9565,6 @@ TEST_F(VkLayerTest, CreateImageMaxLimitsViolation) {
         }
 
         if (dev_limits.maxFramebufferHeight != UINT32_MAX) {
-            image_ci.usage = VK_IMAGE_USAGE_INPUT_ATTACHMENT_BIT;  // try different one too
```
spencer-lunarg, line 9568: changing the usage here changes the value of `vkGetPhysicalDeviceImageFormatProperties`  and `img_limits.maxExtent.height` was changing... this usage change here doesn't add anything worth wild to the test IMO anyway

juan-lunarg (approved):
LGTM

Issue #4594, opened 2022-09-27
`ValidateCreateSwapchain` doesn't account for `VkSurfaceFullScreenExclusiveInfoEXT` in checks


Rua:
The `ValidateCreateSwapchain` function uses the functions `GetPresentModes`, `GetFormats` and `GetCapabilities` as well as `DispatchGetPhysicalDeviceSurfaceCapabilitiesKHR` and `DispatchGetPhysicalDeviceSurfaceCapabilities2KHR` to query what the physical device supports, in order to validate the arguments provided by the user. But the spec of `VkPhysicalDeviceSurfaceInfo2KHR` says the following:
> Additional capabilities of a surface **may** be available to swapchains created with different full-screen exclusive settings - particularly if exclusive full-screen access is application controlled. These additional capabilities **can** be queried by adding a [VkSurfaceFullScreenExclusiveInfoEXT](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkSurfaceFullScreenExclusiveInfoEXT.html) structure to the `pNext` chain of this structure when used to query surface properties. [...] These additional capabilities only apply to swapchains created with the same parameters included in the `pNext` chain of [VkSwapchainCreateInfoKHR](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkSwapchainCreateInfoKHR.html).

None of the above functions use the `VkSurfaceFullScreenExclusiveInfoEXT` extension struct in their queries, even when the user provided that struct as part of `VkSwapchainCreateInfoKHR`. That means that the validation checks may miss out on specific values in `VkSwapchainCreateInfoKHR` that are valid only with this struct, and trigger errors for usage that is actually valid. For example, if additional formats were supported by the device for a specific `fullScreenExclusive` value, the user would still get a validation error for using them.

PR #4703, opened 2022-10-25
Account for `VkSurfaceFullScreenExclusiveInfoEXT` when validating swapchain creation

arno-lunarg:
- Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4594

The main thing about the fix is an update to `SURFACE_STATE` to use `VkSurfaceFormat2KHR`, `VkSurfaceCapabilities2KHR` and `VkPhysicalDeviceSurfaceInfo2KHR`. It now relies on `DispatchGetPhysicalDeviceSurfaceFormats2KHR` and `DispatchGetPhysicalDeviceSurfaceCapabilities2KHR` when `vk_khr_get_surface_capabilities2` is enabled. Otherwise it fall backs to the original functions.

I am still not sure about 3 things:
1) I do a deep copy of the `pCreateInfo->pnext` part that concerns full screen info ([code here](https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4703/files#r1021863471)), is it needed?

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -18139,7 +18139,7 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
     }
 #endif
     const auto surface_caps2 = surface_state->GetCapabilities(IsExtEnabled(instance_extensions.vk_khr_get_surface_capabilities2),
-                                                              physical_device_state->PhysDev(), full_screen_info_copy_addr);
+                                                              physical_device_state->PhysDev(), full_screen_info_copy_addr, this);
 
     bool skip = false;
     VkSurfaceTransformFlagBitsKHR current_transform = surface_caps2.surfaceCapabilities.currentTransform;
@@ -18187,8 +18187,9 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
                               surface_caps2.surfaceCapabilities.maxImageExtent)) {
         safe_VkSurfaceCapabilities2KHR cached_capabilities{};
         if (surface_state) {
-            cached_capabilities = surface_state->GetCapabilities(IsExtEnabled(instance_extensions.vk_khr_get_surface_capabilities2),
-                                                                 physical_device_state->PhysDev(), full_screen_info_copy_addr);
+            cached_capabilities =
+                surface_state->GetCapabilities(IsExtEnabled(instance_extensions.vk_khr_get_surface_capabilities2),
+                                               physical_device_state->PhysDev(), full_screen_info_copy_addr, this);
         } else if (IsExtEnabled(instance_extensions.vk_google_surfaceless_query)) {
             cached_capabilities = physical_device_state->surfaceless_query_state.capabilities;
         }
@@ -18313,7 +18314,7 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
         layer_data::span<const safe_VkSurfaceFormat2KHR> formats{};
         if (surface_state) {
             formats = surface_state->GetFormats(IsExtEnabled(instance_extensions.vk_khr_get_surface_capabilities2),
-                                                physical_device_state->PhysDev(), full_screen_info_copy_addr);
+                                                physical_device_state->PhysDev(), full_screen_info_copy_addr, this);
         } else if (IsExtEnabled(instance_extensions.vk_google_surfaceless_query)) {
             formats = physical_device_state->surfaceless_query_state.formats;
         }
@@ -18353,7 +18354,7 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
 
     std::vector<VkPresentModeKHR> present_modes{};
     if (surface_state) {
-        present_modes = surface_state->GetPresentModes(physical_device);
+        present_modes = surface_state->GetPresentModes(physical_device, this);
     } else if (IsExtEnabled(instance_extensions.vk_google_surfaceless_query)) {
         present_modes = physical_device_state->surfaceless_query_state.present_modes;
     }
@@ -18827,7 +18828,7 @@ bool CoreChecks::ValidateAcquireNextImage(VkDevice device, const AcquireVersion 
         safe_VkSurfaceCapabilities2KHR caps{};
         if (swapchain_data->surface) {
             caps = swapchain_data->surface->GetCapabilities(IsExtEnabled(device_extensions.vk_khr_get_surface_capabilities2),
-                                                            physical_device);
+                                                            physical_device, nullptr, this);
         } else if (IsExtEnabled(instance_extensions.vk_google_surfaceless_query)) {
             caps = physical_device_state->surfaceless_query_state.capabilities;
         }
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -18109,10 +18109,40 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
         }
     }
 
-    VkSurfaceCapabilitiesKHR capabilities{};
-    DispatchGetPhysicalDeviceSurfaceCapabilitiesKHR(physical_device_state->PhysDev(), pCreateInfo->surface, &capabilities);
+    // Will be assigned &full_screen_info_copy if a VkSurfaceFullScreenExclusiveInfoEXT is found in pCreateInfo->pNext chain
+    const void *full_screen_info_copy_addr = nullptr;
+#if defined(VK_USE_PLATFORM_WIN32_KHR)
+    auto full_screen_info_copy = LvlInitStruct<VkSurfaceFullScreenExclusiveInfoEXT>();
+    auto win32_full_screen_info_copy = LvlInitStruct<VkSurfaceFullScreenExclusiveWin32InfoEXT>();
+    if (IsExtEnabled(device_extensions.vk_ext_full_screen_exclusive)) {
+        const auto *full_screen_info = LvlFindInChain<VkSurfaceFullScreenExclusiveInfoEXT>(pCreateInfo->pNext);
+        full_screen_info_copy = LvlInitStruct<VkSurfaceFullScreenExclusiveInfoEXT>();
+        if (full_screen_info) {
+            full_screen_info_copy_addr = &full_screen_info_copy;
+            full_screen_info_copy = *full_screen_info;
+            full_screen_info_copy.pNext = nullptr;
+
+            const auto *win32_full_screen_info = LvlFindInChain<VkSurfaceFullScreenExclusiveWin32InfoEXT>(pCreateInfo->pNext);
+            if (!win32_full_screen_info) {
+                if (LogError(device, "VUID-VkSwapchainCreateInfoKHR-pNext-02679",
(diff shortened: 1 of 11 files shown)


layers/core_validation.cpp
```diff
@@ -18005,10 +18005,40 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
         }
     }
 
-    VkSurfaceCapabilitiesKHR capabilities{};
-    DispatchGetPhysicalDeviceSurfaceCapabilitiesKHR(physical_device_state->PhysDev(), pCreateInfo->surface, &capabilities);
+    auto full_screen_info_copy = LvlInitStruct<VkSurfaceFullScreenExclusiveInfoEXT>();
```
arno-lunarg, line 18008: Here for question 1)

layers/image_state.cpp
```diff
@@ -697,43 +697,69 @@ std::vector<VkPresentModeKHR> SURFACE_STATE::GetPresentModes(VkPhysicalDevice ph
...
-    result.resize(count);
-    DispatchGetPhysicalDeviceSurfaceFormatsKHR(phys_dev, surface(), &count, result.data());
-    formats_[phys_dev] = result;
-    return result;
+
+    std::vector<VkSurfaceFormat2KHR> result;
+    if (vk_khr_get_surface_capabilities2_enabled) {
+        uint32_t count = 0;
+        const auto surface_info2 = GetSurfaceInfo2(surface_info2_pnext);
+        DispatchGetPhysicalDeviceSurfaceFormats2KHR(phys_dev, &surface_info2, &count, nullptr);
```
arno-lunarg, line 720: in the process of adding checks for `Dispatch***` return value

layers/image_state.h
```diff
@@ -346,6 +346,12 @@ class SURFACE_STATE : public BASE_NODE {
     }
 
     VkSurfaceKHR surface() const { return handle_.Cast<VkSurfaceKHR>(); }
+    VkPhysicalDeviceSurfaceInfo2KHR GetSurfaceInfo2(const void *surface_info2_pnext = nullptr) const {
```
jeremyg-lunarg, line 349: seems like this method is only used in 1 place (GetCapabilities2()) Does it really need to be separate?

layers/vk_layer_utils.h
```diff
@@ -42,6 +42,9 @@
 #define STRINGIFY_HELPER(s) #s
 
 #ifdef __cplusplus
+template <typename... Types>
+void UNUSED(Types...) {}
```
jeremyg-lunarg, line 46: why not https://en.cppreference.com/w/cpp/language/attributes/maybe_unused ?

juan-lunarg:
Re-running failed linux jobs due to loader causing a regression: https://github.com/KhronosGroup/Vulkan-Loader/pull/1042

layers/image_state.cpp
```diff
@@ -691,49 +691,101 @@ std::vector<VkPresentModeKHR> SURFACE_STATE::GetPresentModes(VkPhysicalDevice ph
     }
     std::vector<VkPresentModeKHR> result;
     uint32_t count = 0;
-    DispatchGetPhysicalDeviceSurfacePresentModesKHR(phys_dev, surface(), &count, nullptr);
+    VkResult err = DispatchGetPhysicalDeviceSurfacePresentModesKHR(phys_dev, surface(), &count, nullptr);
+    assert(err == VK_SUCCESS || err == VK_INCOMPLETE);
```
spencer-lunarg, line 695: I guess not sure why we would want to assert here

if we would get a `VK_ERROR_OUT_OF_HOST_MEMORY` or `VK_ERROR_SURFACE_LOST_KHR` type error, I wouldn't want to crash on the user, I think it would be better to just return

layers/image_state.h
```diff
@@ -359,11 +365,14 @@ class SURFACE_STATE : public BASE_NODE {
     void SetPresentModes(VkPhysicalDevice phys_dev, std::vector<VkPresentModeKHR> &&modes);
     std::vector<VkPresentModeKHR> GetPresentModes(VkPhysicalDevice phys_dev) const;
 
-    void SetFormats(VkPhysicalDevice phys_dev, std::vector<VkSurfaceFormatKHR> &&fmts);
```
spencer-lunarg, line 362: For `VkFormatProperties2`/`VkFormatProperties3` we don't use the `2` suffix

a `SetFormats2` seems strange without a `SetFormats` along with it. I don't see a real need for a name change of these functions

layers/state_tracker.cpp
```diff
@@ -4005,15 +4009,18 @@ void ValidationStateTracker::PostCallRecordGetPhysicalDeviceSurfaceFormatsKHR(Vk
     if ((VK_SUCCESS != result) && (VK_INCOMPLETE != result)) return;
 
     if (pSurfaceFormats) {
+        std::vector<VkSurfaceFormat2KHR> fmts(*pSurfaceFormatCount);
```
spencer-lunarg, line 4012: nit: for naming consistency can this be named `formats` so its

`pd_state->surfaceless_query_state.formats = std::move(formats);`

layers/image_state.cpp
```diff
@@ -691,49 +691,101 @@ std::vector<VkPresentModeKHR> SURFACE_STATE::GetPresentModes(VkPhysicalDevice ph
...
 
-void SURFACE_STATE::SetFormats(VkPhysicalDevice phys_dev, std::vector<VkSurfaceFormatKHR> &&fmts) {
+void SURFACE_STATE::SetFormats2(VkPhysicalDevice phys_dev, std::vector<VkSurfaceFormat2KHR> &&fmts) {
     auto guard = Lock();
     assert(phys_dev);
     formats_[phys_dev] = std::move(fmts);
 }
 
-std::vector<VkSurfaceFormatKHR> SURFACE_STATE::GetFormats(VkPhysicalDevice phys_dev) const {
+layer_data::span<const VkSurfaceFormat2KHR> SURFACE_STATE::GetFormats2(bool vk_khr_get_surface_capabilities2_enabled,
```
spencer-lunarg, line 715: `vk_khr_get_surface_capabilities2_enabled` -> `get_surface_capabilities2` as we don't pass the whole ext name as a bool in anywhere else and it nice if another extension also enables this logic in the future as well

(another example: `PickSpirvEnv`)

layers/image_state.cpp
```diff
@@ -691,49 +691,101 @@ std::vector<VkPresentModeKHR> SURFACE_STATE::GetPresentModes(VkPhysicalDevice ph
...
+    if (!(err == VK_SUCCESS || err == VK_INCOMPLETE)) {
+        return result;
+    }
     result.resize(count);
-    DispatchGetPhysicalDeviceSurfacePresentModesKHR(phys_dev, surface(), &count, result.data());
+    err = DispatchGetPhysicalDeviceSurfacePresentModesKHR(phys_dev, surface(), &count, result.data());
+    assert(err == VK_SUCCESS);
+    if (err != VK_SUCCESS) {
+        result.resize(0);
+        return result;
```
spencer-lunarg, line 704: If we do this, looking where `GetPresentModes()` gets called it will raise a error for `VUID-VkSwapchainCreateInfoKHR-presentMode-01281` which I feel will be more confusing

layers/image_state.cpp
spencer-lunarg: Thinking more about this, we don't check a lot of `Dispatch` calls we make... but I am in favor of doing it the way `CoreChecks::GetPhysicalDeviceImageFormatProperties` does it

layers/image_state.h
arno-lunarg: It is also used in GetFormats2() !

layers/image_state.cpp
arno-lunarg: Yes agreed, but at the same time at this point the `result` vec would not be in a correct state and should not be used : /

spencer-lunarg: I think this goes back to https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4703#discussion_r1005279778 where we should leave an error stating that our internal `Dispatch` failed instead of trying to add extra complexity handling these cases (which I assume will be rare to none)

arno-lunarg: Ok makes sense. Error checking can be a rabbit hole, if `result` is invalid the error should somehow get propagated for true error handling I guess.
But adding a block similar to:
```
    if (image_properties_result != VK_SUCCESS) {
        skip |= LogError(device, vuid_string,
                         "vkGetPhysicalDeviceImageFormatProperties() or vkGetPhysicalDeviceImageFormatProperties2() unexpectedly "
                         "failed with result = %s, "
                         "when called for validation with following params: "
                         "format: %s, imageType: %s, "
                         "tiling: %s, usage: %s, "
                         "flags: %s.",
                         string_VkResult(image_properties_result), string_VkFormat(image_create_info.format),
                         string_VkImageType(image_create_info.imageType), string_VkImageTiling(image_create_info.tiling),
                         string_VkImageUsageFlags(image_create_info.usage).c_str(),
                         string_VkImageCreateFlags(image_create_info.flags).c_str());
    }
```
For each failed Dispatch*** seems really tedious, and too verbose

layers/vk_layer_utils.h
arno-lunarg: It requires c++ 17 : /

layers/image_state.cpp
arno-lunarg: Maybe a middle ground would be to allow `GetPresentModes`, `GetFormats` and `GetCapabilites` to return a bool to check for internal failure, and only try to do validation when they succeed. And no actual message to signal this internal error, only `assert` in debug builds.
I am aware of the "But then, if we do this here for those 3 functions, we should do it everywhere", and I do not know what to respond to that...

layers/core_validation.cpp
```diff
@@ -18107,10 +18107,40 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
         }
     }
 
-    VkSurfaceCapabilitiesKHR capabilities{};
-    DispatchGetPhysicalDeviceSurfaceCapabilitiesKHR(physical_device_state->PhysDev(), pCreateInfo->surface, &capabilities);
+    // Will be assigned &full_screen_info_copy if a VkSurfaceFullScreenExclusiveInfoEXT is found in pCreateInfo->pNext chain
```
arno-lunarg, line 18110: Here for question 1)

layers/device_state.h
```diff
@@ -113,9 +113,9 @@ class QUEUE_FAMILY_PERF_COUNTERS {
 
 class SURFACELESS_QUERY_STATE {
   public:
-    std::vector<VkSurfaceFormatKHR> formats;
+    std::vector<VkSurfaceFormat2KHR> formats;
     std::vector<VkPresentModeKHR> present_modes;
-    VkSurfaceCapabilitiesKHR capabilities;
+    VkSurfaceCapabilities2KHR capabilities;
```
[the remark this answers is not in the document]
arno-lunarg: Yes I need to update to the safe_ versions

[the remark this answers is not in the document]
arno-lunarg: Yes I need to update to the safe_ versions

layers/vk_layer_utils.h
jeremyg-lunarg: we have c++17 now!

PR #4706, opened 2022-10-25
tests: Don't wait for all the nanoseconds

jeremyg-lunarg:
A UINT64_MAX timeout is about 585 years. We want tests to complete before then, even if everything has gone wrong.

### DIFF
--- a/tests/positive/image_buffer.cpp
+++ b/tests/positive/image_buffer.cpp
@@ -1986,8 +1986,8 @@ TEST_F(VkPositiveLayerTest, TransferImageToSwapchainDeviceGroup) {
     VkFence fence_handle = fence.handle();
 
     uint32_t image_index;
-    vk::AcquireNextImageKHR(m_device->handle(), m_swapchain, UINT64_MAX, VK_NULL_HANDLE, fence_handle, &image_index);
-    vk::WaitForFences(m_device->device(), 1, &fence_handle, VK_TRUE, std::numeric_limits<uint64_t>::max());
+    vk::AcquireNextImageKHR(m_device->handle(), m_swapchain, kWaitTimeout, VK_NULL_HANDLE, fence_handle, &image_index);
+    vk::WaitForFences(m_device->device(), 1, &fence_handle, VK_TRUE, kWaitTimeout);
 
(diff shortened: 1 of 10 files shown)


tests/vkrenderframework.h
```diff
@@ -47,6 +47,8 @@ class VkImageObj;
 
 using vk_testing::MakeVkHandles;
 
+static constexpr uint64_t kWaitTimeout{10000000000};  // 10 seconds in ns
```
jeremyg-lunarg, line 50: Should this constant go in one of the test framework classes?

tests/positive/image_buffer.cpp
```diff
@@ -1986,8 +1986,8 @@ TEST_F(VkPositiveLayerTest, TransferImageToSwapchainDeviceGroup) {
     VkFence fence_handle = fence.handle();
 
     uint32_t image_index;
-    vk::AcquireNextImageKHR(m_device->handle(), m_swapchain, UINT64_MAX, VK_NULL_HANDLE, fence_handle, &image_index);
-    vk::WaitForFences(m_device->device(), 1, &fence_handle, VK_TRUE, std::numeric_limits<uint64_t>::max());
+    vk::AcquireNextImageKHR(m_device->handle(), m_swapchain, kWaitTimeout, VK_NULL_HANDLE, fence_handle, &image_index);
```
jeremyg-lunarg, line 1989: Should all wait calls have an ASSERT_VK_SUCCESS(err) after them?

tests/vkrenderframework.h
ncesario-lunarg: I don't have a strong preference.

ncesario-lunarg (approved):
LGTM!

spencer-lunarg (approved):

tests/positive/image_buffer.cpp
spencer-lunarg: I guess not sure if some tests can allow `VK_TIMEOUT`

jeremyg-lunarg: I'm going to skip this part for now, it seems like it could end up being subjective.

PR #4707, opened 2022-10-25
tests: Fix PresentIdWait minor issues

juan-lunarg:
- Destroy Win32 window handle
- InitSurface ignoring parameters
- Allow InitSurface to be used twice

### DIFF
--- a/tests/vklayertests_wsi.cpp
+++ b/tests/vklayertests_wsi.cpp
@@ -1843,21 +1843,18 @@ TEST_F(VkLayerTest, DisplayPresentInfoSrcRect) {
 TEST_F(VkLayerTest, PresentIdWait) {
     TEST_DESCRIPTION("Test present wait extension");
     SetTargetApiVersion(VK_API_VERSION_1_1);
-    AddSurfaceExtension();
     auto present_id_features = LvlInitStruct<VkPhysicalDevicePresentIdFeaturesKHR>();
     auto present_wait_features = LvlInitStruct<VkPhysicalDevicePresentWaitFeaturesKHR>(&present_id_features);
     auto features2 = LvlInitStruct<VkPhysicalDeviceFeatures2KHR>(&present_wait_features);
-    m_device_extension_names.push_back(VK_KHR_PRESENT_WAIT_EXTENSION_NAME);
-    m_device_extension_names.push_back(VK_KHR_PRESENT_ID_EXTENSION_NAME);
-    m_device_extension_names.push_back(VK_KHR_SWAPCHAIN_EXTENSION_NAME);
-    bool retval = InitFrameworkAndRetrieveFeatures(features2);
+    AddRequiredExtensions(VK_KHR_PRESENT_WAIT_EXTENSION_NAME);
+    AddRequiredExtensions(VK_KHR_PRESENT_ID_EXTENSION_NAME);
+    AddSurfaceExtension();
+    const bool retval = InitFrameworkAndRetrieveFeatures(features2);
     if (!retval) {
-        printf("%s Error initializing extensions or retrieving features, skipping test\n", kSkipPrefix);
-        return;
+        GTEST_SKIP() << "Error initializing extensions or retrieving features, skipping test.";
     }
     if (!present_id_features.presentId || !present_wait_features.presentWait) {
-        printf("%s presentWait feature is not available, skipping test\n", kSkipPrefix);
-        return;
+        GTEST_SKIP() << "presentWait feature is not available, skipping test.";
     }
 
     ASSERT_NO_FATAL_FAILURE(InitState(nullptr, &features2, VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT));
@@ -1868,7 +1865,7 @@ TEST_F(VkLayerTest, PresentIdWait) {
     ASSERT_TRUE(InitSwapchain());
     VkSurfaceKHR surface2;
     VkSwapchainKHR swapchain2;
-    InitSurface(m_width, m_height, surface2);
+    InitSurface(surface2);
     InitSwapchain(surface2, VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT, VK_SURFACE_TRANSFORM_IDENTITY_BIT_KHR, swapchain2);
 
     auto vkWaitForPresentKHR = (PFN_vkWaitForPresentKHR)vk::GetDeviceProcAddr(m_device->device(), "vkWaitForPresentKHR");
--- a/tests/vkrenderframework.cpp
+++ b/tests/vkrenderframework.cpp
@@ -862,16 +862,15 @@ LRESULT CALLBACK WindowProc(HWND hwnd, UINT uMsg, WPARAM wParam, LPARAM lParam) 
 bool VkRenderFramework::InitSurface(VkSurfaceKHR &surface) {
 #if defined(VK_USE_PLATFORM_WIN32_KHR)
     HINSTANCE window_instance = GetModuleHandle(nullptr);
-    if (m_win32Window == nullptr) {
-        const char class_name[] = "test";
-        WNDCLASS wc = {};
-        wc.lpfnWndProc = WindowProc;
-        wc.hInstance = window_instance;
-        wc.lpszClassName = class_name;
-        RegisterClass(&wc);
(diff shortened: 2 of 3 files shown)


tests/vklayertests_wsi.cpp
```diff
@@ -1843,21 +1843,18 @@ TEST_F(VkLayerTest, DisplayPresentInfoSrcRect) {
...
     SetTargetApiVersion(VK_API_VERSION_1_1);
-    AddSurfaceExtension();
     auto present_id_features = LvlInitStruct<VkPhysicalDevicePresentIdFeaturesKHR>();
     auto present_wait_features = LvlInitStruct<VkPhysicalDevicePresentWaitFeaturesKHR>(&present_id_features);
     auto features2 = LvlInitStruct<VkPhysicalDeviceFeatures2KHR>(&present_wait_features);
     m_device_extension_names.push_back(VK_KHR_PRESENT_WAIT_EXTENSION_NAME);
     m_device_extension_names.push_back(VK_KHR_PRESENT_ID_EXTENSION_NAME);
-    m_device_extension_names.push_back(VK_KHR_SWAPCHAIN_EXTENSION_NAME);
-    bool retval = InitFrameworkAndRetrieveFeatures(features2);
+    AddSurfaceExtension();
```
juan-lunarg, line 1851: AddSurfaceExtension implies VK_KHR_SWAPCHAIN_EXTENSION_NAME

tests/vklayertests_wsi.cpp
```diff
@@ -1868,7 +1865,7 @@ TEST_F(VkLayerTest, PresentIdWait) {
     ASSERT_TRUE(InitSwapchain());
     VkSurfaceKHR surface2;
     VkSwapchainKHR swapchain2;
-    InitSurface(m_width, m_height, surface2);
+    InitSurface(surface2);
```
juan-lunarg, line 1868: This is the main problem. Currently (depending on the platform), surface2 just ends up being VK_NULL_HANDLE. Since it actually just gets ignored.

tests/vkrenderframework.cpp
```diff
@@ -848,78 +851,85 @@ void VkRenderFramework::InitViewport(float width, float height) {
...
+            XCreateSimpleWindow(m_surface_dpy, RootWindow(m_surface_dpy, s), 0, 0, static_cast<int>(m_width),
+                                static_cast<int>(m_height), 1, BlackPixel(m_surface_dpy, s), WhitePixel(m_surface_dpy, s));
+    }
+
+    if (m_surface_dpy) {
         VkXlibSurfaceCreateInfoKHR surface_create_info = LvlInitStruct<VkXlibSurfaceCreateInfoKHR>();
         surface_create_info.dpy = m_surface_dpy;
         surface_create_info.window = m_surface_window;
-        VkResult err = vk::CreateXlibSurfaceKHR(instance(), &surface_create_info, nullptr, &m_surface);
+        VkResult err = vk::CreateXlibSurfaceKHR(instance(), &surface_create_info, nullptr, &surface);
```
juan-lunarg, line 903: Currently PresentWaitId breaks on XLIB builds due to not using the function parameter properly.

tests/vkrenderframework.h
```diff
@@ -257,8 +257,7 @@ class VkRenderFramework : public VkTestFramework {
     void InitViewport(float width, float height);
     void InitViewport();
     bool InitSurface();
-    bool InitSurface(float width, float height);
-    bool InitSurface(float width, float height, VkSurfaceKHR &surface);
```
juan-lunarg, line 261: I removed passing width/height as arguments due to `InitSwapchainInfo` / `InitSwapchain` implicitly using member variables regardless.

tests/vkrenderframework.cpp
```diff
@@ -848,78 +851,85 @@ void VkRenderFramework::InitViewport(float width, float height) {
...
 #endif
-    return (m_surface == VK_NULL_HANDLE) ? false : true;
+    return (surface != VK_NULL_HANDLE);
 }
 
 // Makes query to get information about swapchain needed to create a valid swapchain object each test creating a swapchain will need
 void VkRenderFramework::InitSwapchainInfo() {
     const VkPhysicalDevice physicalDevice = gpu();
 
+    assert(m_surface != VK_NULL_HANDLE);
```
juan-lunarg, line 931: I added this assert in case you try to create a surface and then create a swapchain, without having already initialized the member variables. Otherwise all the swapchain/surface code doesn't work.

juan-lunarg:
Updated with more concise commit message

spencer-lunarg (approved):

tests/vklayertests_wsi.cpp
```diff
@@ -1843,21 +1843,18 @@ TEST_F(VkLayerTest, DisplayPresentInfoSrcRect) {
 TEST_F(VkLayerTest, PresentIdWait) {
     TEST_DESCRIPTION("Test present wait extension");
     SetTargetApiVersion(VK_API_VERSION_1_1);
-    AddSurfaceExtension();
     auto present_id_features = LvlInitStruct<VkPhysicalDevicePresentIdFeaturesKHR>();
     auto present_wait_features = LvlInitStruct<VkPhysicalDevicePresentWaitFeaturesKHR>(&present_id_features);
     auto features2 = LvlInitStruct<VkPhysicalDeviceFeatures2KHR>(&present_wait_features);
     m_device_extension_names.push_back(VK_KHR_PRESENT_WAIT_EXTENSION_NAME);
     m_device_extension_names.push_back(VK_KHR_PRESENT_ID_EXTENSION_NAME);
```
ncesario-lunarg, line 1850: Might as well update this to use `AddRequiredExtensions`.

tests/vkrenderframework.cpp
```diff
@@ -1018,6 +1028,12 @@ void VkRenderFramework::DestroySwapchain() {
         }
     }
 
+#if defined(VK_USE_PLATFORM_WIN32_KHR)
+    if (m_win32Window != nullptr) {
+        assert(DestroyWindow(m_win32Window) != 0);
```
ncesario-lunarg, line 1033: Won't `DestroyWindow` only get called for debug builds? If so, is that the intention?

ncesario-lunarg (commented):
A couple questions/comments, but otherwise LGTM!

tests/vklayertests_wsi.cpp
juan-lunarg: These are the extensions being tested, and from what I can tell using vulkanCapsViewer aren't always supported.

tests/vkrenderframework.cpp
juan-lunarg: Good catch, that's a bug.

tests/vklayertests_wsi.cpp
juan-lunarg: I read your comment wrong. I'll use AddRequiredExtension :+1:

tests/vkrenderframework.cpp
```diff
@@ -848,78 +851,87 @@ void VkRenderFramework::InitViewport(float width, float height) {
...
+    VkResult err = vk::CreateAndroidSurfaceKHR(instance(), &surface_create_info, nullptr, &surface);
     if (err != VK_SUCCESS) return false;
 #endif
 
 #if defined(VK_USE_PLATFORM_XLIB_KHR)
-    assert(m_surface_dpy == nullptr);
-    m_surface_dpy = XOpenDisplay(NULL);
+    if (m_surface_dpy == nullptr) {
+        m_surface_dpy = XOpenDisplay(NULL);
+        if (m_surface_dpy != nullptr) {
```
juan-lunarg, line 893: Added check for XOpenDisplay failing. Wasn't able to repro locally, but XOpenDisplay does fail on server. Latest force push addresses this.

juan-lunarg:
Closing because PresentWaitId relies on Win32 window handles being leaked
Swapchain creation will fail if you re-use the WIN32 window handle with 2 swapchains

juan-lunarg:
Also PresentIdWait just doesn't actually work on Linux with Nvidia GPUs. Working with @charles-lunarg it appears to be a driver bug.

jeremyg-lunarg:
Is it worth still merging your cleanups to the surface code?

juan-lunarg:
> Is it worth still merging your cleanups to the surface code?

There is still some cleanup worth merging. I'll make a separate PR though.

PR #4709, opened 2022-10-26
layers: Cleanup Shader Module interface

spencer-lunarg:
Tried moving more things into `StaticData`, move things into `SHADER_MODULE_STATE`, and rename structs to match coding style


ncesario-lunarg (approved):
This is awesome!

PR #4713, opened 2022-10-26
build: Fix disabling BUILD_WERROR

jbauman42:
After commit 3bfe7b52a55c568d19a5d5cf53dc830aad895a2d, BUILD_WERROR=OFF doesn't do anything if the clang version is new enough. That change doesn't seem intentional and prevents building on extremely new clang versions with new warnings added.


juan-lunarg:
See https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4736

PR #4719, opened 2022-10-27
tests: Remove pointless try/catch

juan-lunarg:
Currently the ViewportInheritanceTestData constructor tries to catch and re-throw an exception.

However, none of the functions throws an exception.

### DIFF
--- a/tests/vklayertests_viewport_inheritance.cpp
+++ b/tests/vklayertests_viewport_inheritance.cpp
@@ -66,7 +66,7 @@ class ViewportInheritanceTestData {
     // Set to a failure message if initialization failed.
     const char* m_failureReason = nullptr;
 
-    void PickColorFormat(VkPhysicalDevice physical_device) {
+    void PickColorFormat(VkPhysicalDevice physical_device) noexcept {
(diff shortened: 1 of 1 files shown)


jeremyg-lunarg (approved):
LGTM.

tests/vklayertests_viewport_inheritance.cpp
```diff
@@ -247,28 +247,24 @@ class ViewportInheritanceTestData {
...
-            if (m_failureReason) return;
-            CreateFramebuffer();
-            if (m_failureReason) return;
-            CreatePipelineLayout();
-            if (m_failureReason) return;
-            CreateShaderStages();
-            if (m_failureReason) return;
-        }
-        catch (...) {
-            Cleanup();
```
juan-lunarg, line 269: Was the intention to call Cleanup if any of these failed?

juan-lunarg: Actually calling cleanup in this scenario would cause issues due to un-initialized memory. So the current change is fine, since it's completely consistent with what was practically being used.

PR #4721, opened 2022-10-27
layers: Improve ImportSemphoreFUCHSIA validation

jbauman42:
This patch fixes compilation and adds validation of more properties, including that the object isn't in use and that it's not a timeline semaphore.


jeremyg-lunarg (approved):
LGTM

PR #4725, opened 2022-10-27
layers: Fix missing linear_color_attachment validation

pdaniell-nv:
There were a few places where VVL would throw false failures when the application tried to use linear color attachments with the VK_NV_linear_color_attachment extension and linearColorAttachment feature enabled. Implementations that can't support general color attachment usage of linear images can expose this extension if they can support it with the restriction that all color attachments to the framebuffer are either optimal or linear, and not a mix of the two. Because of this restriction, they can't advertise the regular VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT format feature for linear images, they instead advertise VK_FORMAT_FEATURE_2_LINEAR_COLOR_ATTACHMENT_BIT_NV.


layers/core_validation.cpp
```diff
@@ -3648,7 +3648,7 @@ bool CoreChecks::ValidatePipeline(std::vector<std::shared_ptr<PIPELINE_STATE>> c
                                      pipe_index);
                 }
 
-                if (!IsExtEnabled(device_extensions.vk_nv_linear_color_attachment)) {
+                if (!enabled_features.linear_color_attachment_features.linearColorAttachment) {
```
spencer-lunarg, line 3651: So the reason we use `IsExtEnabled` is depending which extension is given determines the URL link to the spec

I would expect somewhere else to make sure `VK_FORMAT_FEATURE_2_LINEAR_COLOR_ATTACHMENT_BIT_NV` can't be used without `linearColorAttachment` and with that assumption this code could be simplified to

```cpp
if ((format_features & (VK_FORMAT_FEATURE_2_COLOR_ATTACHMENT_BIT | VK_FORMAT_FEATURE_2_LINEAR_COLOR_ATTACHMENT_BIT_NV)) == 0) {
    IsExtEnabled(device_extensions.vk_nv_linear_color_attachment)
    const char *vuid = IsExtEnabled(device_extensions.vk_nv_linear_color_attachment)
                    ? "VUID-VkGraphicsPipelineCreateInfo-renderPass-06582"
                    : "VUID-VkGraphicsPipelineCreateInfo-renderPass-06581";
    skip |= LogError(device, vuid,
                    "vkCreateGraphicsPipelines() pCreateInfos[%" PRIu32
                    "]: pColorAttachmentFormats[%" PRIu32 "] color format (%s) must be a format with potential format features that include "
                    "VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT or VK_FORMAT_FEATURE_2_LINEAR_COLOR_ATTACHMENT_BIT_NV",
                    pipe_index, color_index, string_VkFormat(color_format));
}
```

layers/core_validation.cpp
```diff
@@ -9600,11 +9600,29 @@ bool CoreChecks::PreCallValidateBeginCommandBuffer(VkCommandBuffer commandBuffer
...
+                                        "VkCommandBufferInheritanceRenderingInfo->pColorAttachmentFormats[%u] (%s) must be a "
+                                        "format with potential format features that include VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT",
+                                        i, string_VkFormat(attachment_format));
+                                }
+                            } else {
+                                if ((potential_format_features & (VK_FORMAT_FEATURE_2_COLOR_ATTACHMENT_BIT_KHR |
+                                                                  VK_FORMAT_FEATURE_2_LINEAR_COLOR_ATTACHMENT_BIT_NV)) == 0) {
+                                    skip |= LogError(
+                                        commandBuffer,
+                                        "VUID-VkCommandBufferInheritanceRenderingInfoKHR-pColorAttachmentFormats-06492",
```
spencer-lunarg, line 9619: random side note: `Linear Color attachment` in the spec for this VUID doesn't actually point anywhere

layers/core_validation.cpp
```diff
@@ -9600,11 +9600,29 @@ bool CoreChecks::PreCallValidateBeginCommandBuffer(VkCommandBuffer commandBuffer
...
+                                if ((potential_format_features & VK_FORMAT_FEATURE_2_COLOR_ATTACHMENT_BIT_KHR) == 0) {
+                                    skip |= LogError(
+                                        commandBuffer, "VUID-VkCommandBufferInheritanceRenderingInfo-pColorAttachmentFormats-06006",
+                                        "vkBeginCommandBuffer(): "
+                                        "VkCommandBufferInheritanceRenderingInfo->pColorAttachmentFormats[%u] (%s) must be a "
+                                        "format with potential format features that include VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT",
+                                        i, string_VkFormat(attachment_format));
+                                }
+                            } else {
+                                if ((potential_format_features & (VK_FORMAT_FEATURE_2_COLOR_ATTACHMENT_BIT_KHR |
```
spencer-lunarg, line 9615: This VU seems to only be checking for `VK_FORMAT_FEATURE_2_LINEAR_COLOR_ATTACHMENT_BIT_NV` and seems this would not catch a case where `COLOR_ATTACHMENT_BIT` is only set instead

layers/core_validation.cpp
```diff
@@ -13942,14 +13960,27 @@ bool CoreChecks::ValidateRenderpassAttachmentUsage(RenderPassCreateVersion rp_ve
...
                     const VkFormatFeatureFlags2KHR format_features = GetPotentialFormatFeatures(attachment_format);
-                    if ((format_features & valid_flags) == 0) {
-                        vuid = use_rp2 ? "VUID-VkSubpassDescription2-pInputAttachments-02897"
-                                       : "VUID-VkSubpassDescription-pInputAttachments-02647";
-                        skip |=
-                            LogError(device, vuid,
-                                     "%s: Input attachment %s format (%s) does not contain VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT "
-                                     "| VK_FORMAT_FEATURE_DEPTH_STENCIL_ATTACHMENT_BIT.",
-                                     function_name, error_type.c_str(), string_VkFormat(attachment_format));
+                    if (!enabled_features.linear_color_attachment_features.linearColorAttachment) {
```
spencer-lunarg, line 13963: I'm slightly confused, so if `linearColorAttachment` is enabled the format could not have `VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT` and still be valid?

pdaniell-nv: Right. VK_NV_linear_color_attachment was created because we can't advertise VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT in general for linear tiling image because the Vulkan spec allows framebuffers to be created with a mixture of linear and optimal color image attachments, but we only support all optimal or all linear attachments. So this extension created a new format feature bit to indicate that the linear image can be used as a color attachment, but in this more limited way.

layers/core_validation.cpp
```diff
@@ -9600,11 +9600,29 @@ bool CoreChecks::PreCallValidateBeginCommandBuffer(VkCommandBuffer commandBuffer
...
                         const VkFormat attachment_format = p_inherited_rendering_info->pColorAttachmentFormats[i];
                         if (attachment_format != VK_FORMAT_UNDEFINED) {
-                            const VkFormatFeatureFlags2KHR potential_format_features = GetPotentialFormatFeatures(attachment_format);
-                            if ((potential_format_features & VK_FORMAT_FEATURE_2_COLOR_ATTACHMENT_BIT_KHR) == 0) {
-                                skip |= LogError(commandBuffer, "VUID-VkCommandBufferInheritanceRenderingInfo-pColorAttachmentFormats-06006",
-                                    "vkBeginCommandBuffer(): VkCommandBufferInheritanceRenderingInfo->pColorAttachmentFormats[%u] (%s) must be a format with potential format features that include VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT",
-                                    i, string_VkFormat(attachment_format));
+                            const VkFormatFeatureFlags2KHR potential_format_features =
+                                GetPotentialFormatFeatures(attachment_format);
+                            if (!enabled_features.linear_color_attachment_features.linearColorAttachment) {
```
spencer-lunarg, line 9605: Looking that these VUs, I think the logic for all of these should be the following to help make the it less error prone for a future dev touching this code

```c++
if ((potential_format_features & VK_FORMAT_FEATURE_2_COLOR_ATTACHMENT_BIT_KHR) == 0) {
    if (!enabled_features.linear_color_attachment_features.linearColorAttachment) {
        LogError(06006);
    } else if ((potential_format_features & VK_FORMAT_FEATURE_2_LINEAR_COLOR_ATTACHMENT_BIT_NV) == 0) {
        LogError(06492);
    }
}
```

spencer-lunarg (changes requested):
@pdaniell-nv it all LGTM to merge but first

1. I would inverse the logic of the `if/else` to more match the wording of the VUID (prevent someone in the future from messing it up)
2. Can we get a test to help prevent a regression here, should be able to hopefully just take the logic of the current tests for the other VUs and just add a

```c++
AddRequiredExtensions(VK_NV_LINEAR_COLOR_ATTACHMENT_EXTENSION_NAME);
// ...
auto linear_color_features = LvlInitStruct<VkPhysicalDeviceLinearColorAttachmentFeaturesNV>();
VkPhysicalDeviceFeatures2 features2 = GetPhysicalDeviceFeatures2(linear_color_features);
if (linear_color_features.linearColorAttachment == VK_FALSE) {
    GTEST_SKIP() << "Test requires (unsupported) linearColorAttachment";
}

// Use the formats you can't expose COLOR_ATTACHMENT on
```

and it should hit the other VUs for you

pdaniell-nv:
> @pdaniell-nv it all LGTM to merge but first
> 
> 1. I would inverse the logic of the `if/else` to more match the wording of the VUID (prevent someone in the future from messing it up)
> 2. Can we get a test to help prevent a regression here, should be able to hopefully just take the logic of the current tests for the other VUs and just add a
> 
> ```c++
> AddRequiredExtensions(VK_NV_LINEAR_COLOR_ATTACHMENT_EXTENSION_NAME);
> // ...
> auto linear_color_features = LvlInitStruct<VkPhysicalDeviceLinearColorAttachmentFeaturesNV>();
> VkPhysicalDeviceFeatures2 features2 = GetPhysicalDeviceFeatures2(linear_color_features);
> if (linear_color_features.linearColorAttachment == VK_FALSE) {
>     GTEST_SKIP() << "Test requires (unsupported) linearColorAttachment";
> }
> 
> // Use the formats you can't expose COLOR_ATTACHMENT on
> ```
> 
> and it should hit the other VUs for you

Thanks for the review. I've addressed your two requests.

spencer-lunarg (approved):

PR #2326, opened 2020-11-16
Correct for non-device parent object validation in Object Lifetimes

mark-lunarg:
Many of these were being incorrectly validated against the owning vkDevice objects. These are now special case and the code-gen'd device checks are properly skipped.  

Fixes #2256. 

[Passed internal CI](http://erusea:4040/#/c/7749/).

PR #4728, opened 2022-10-28
tests: LeakASwapchain

### LINKED ISSUE #4608: Validation doesn't report undestroyed VkSwapchainKHR when VkDevice and VkInstance are destroyed
Issue description:

Steps:
1. Creating VkInstance with `VK_EXT_DEBUG_UTILS_EXTENSION_NAME` and `VK_LAYER_KHRONOS_validation` enabled
2. Creating VkDebugUtilsMessengerCreateInfoEXT ... creating VkDevice ...
3. Creating vkCreateSwapchainKHR
4. Call vkDestroyDevice ... call vkDestroyInstance

Wanted:
I think there should be a warning that VkSwapchain is not destroyed, but there is nothing except normal debug info. So I made a test that comments out such as `vkDestroyCommandPool`, and some Validation Error turn up.

juan-lunarg:
Ensure that a leaked swapchain is reported by the validation layers.

closes #4608

### DIFF
--- a/layers/generated/object_tracker.cpp
+++ b/layers/generated/object_tracker.cpp
@@ -39,7 +39,6 @@ WriteLockGuard ObjectLifetimes::WriteLock() { return WriteLockGuard(validation_o
 bool ObjectLifetimes::ReportUndestroyedInstanceObjects(VkInstance instance, const std::string& error_code) const {
     bool skip = false;
     skip |= ReportLeakedInstanceObjects(instance, kVulkanObjectTypeSurfaceKHR, error_code);
-    skip |= ReportLeakedInstanceObjects(instance, kVulkanObjectTypeSwapchainKHR, error_code);
     // No destroy API -- do not report: skip |= ReportLeakedInstanceObjects(instance, kVulkanObjectTypeDisplayKHR, error_code);
     // No destroy API -- do not report: skip |= ReportLeakedInstanceObjects(instance, kVulkanObjectTypeDisplayModeKHR, error_code);
     skip |= ReportLeakedInstanceObjects(instance, kVulkanObjectTypeDebugReportCallbackEXT, error_code);
@@ -72,6 +71,7 @@ bool ObjectLifetimes::ReportUndestroyedDeviceObjects(VkDevice device, const std:
     skip |= ReportLeakedDeviceObjects(device, kVulkanObjectTypeSamplerYcbcrConversion, error_code);
     skip |= ReportLeakedDeviceObjects(device, kVulkanObjectTypeDescriptorUpdateTemplate, error_code);
     skip |= ReportLeakedDeviceObjects(device, kVulkanObjectTypePrivateDataSlot, error_code);
+    skip |= ReportLeakedDeviceObjects(device, kVulkanObjectTypeSwapchainKHR, error_code);
     skip |= ReportLeakedDeviceObjects(device, kVulkanObjectTypeVideoSessionKHR, error_code);
     skip |= ReportLeakedDeviceObjects(device, kVulkanObjectTypeVideoSessionParametersKHR, error_code);
     skip |= ReportLeakedDeviceObjects(device, kVulkanObjectTypeDeferredOperationKHR, error_code);
@@ -90,7 +90,6 @@ bool ObjectLifetimes::ReportUndestroyedDeviceObjects(VkDevice device, const std:
 
 void ObjectLifetimes::DestroyLeakedInstanceObjects() {
     DestroyUndestroyedObjects(kVulkanObjectTypeSurfaceKHR);
-    DestroyUndestroyedObjects(kVulkanObjectTypeSwapchainKHR);
     DestroyUndestroyedObjects(kVulkanObjectTypeDisplayKHR);
     DestroyUndestroyedObjects(kVulkanObjectTypeDisplayModeKHR);
     DestroyUndestroyedObjects(kVulkanObjectTypeDebugReportCallbackEXT);
@@ -121,6 +120,7 @@ void ObjectLifetimes::DestroyLeakedDeviceObjects() {
     DestroyUndestroyedObjects(kVulkanObjectTypeSamplerYcbcrConversion);
     DestroyUndestroyedObjects(kVulkanObjectTypeDescriptorUpdateTemplate);
     DestroyUndestroyedObjects(kVulkanObjectTypePrivateDataSlot);
+    DestroyUndestroyedObjects(kVulkanObjectTypeSwapchainKHR);
     DestroyUndestroyedObjects(kVulkanObjectTypeVideoSessionKHR);
     DestroyUndestroyedObjects(kVulkanObjectTypeVideoSessionParametersKHR);
     DestroyUndestroyedObjects(kVulkanObjectTypeDeferredOperationKHR);
--- a/scripts/object_tracker_generator.py
+++ b/scripts/object_tracker_generator.py
@@ -394,6 +394,9 @@ class ObjectTrackerOutputGenerator(OutputGenerator):
         self.type_categories = GetTypeCategories(self.registry.tree)
         self.is_aliased_type = GetHandleAliased(self.registry.tree)
 
+        # Fix the parent of VkSwapchainKHR
+        self.handle_parents['VkSwapchainKHR'] = 'VkDevice'
+
         header_file = (genOpts.filename == 'object_tracker.h')
         source_file = (genOpts.filename == 'object_tracker.cpp')
 
--- a/tests/vklayertests_wsi.cpp
+++ b/tests/vklayertests_wsi.cpp
@@ -1840,6 +1840,34 @@ TEST_F(VkLayerTest, DisplayPresentInfoSrcRect) {
     m_errorMonitor->VerifyFound();
 }
 
+TEST_F(VkLayerTest, LeakASwapchain) {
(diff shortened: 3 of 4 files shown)


juan-lunarg:
Currently have this as a Draft PR, since I'm unsure about 2 things.

- 1.) How to fix the code generation

Swapchain's are created by the device, but for some reason are listed as under instance in `object_tracker.cpp`

When I manually change this I get the expected VVL error desired in the issue.

However, even after looking at `object_tracker_generator.py` I'm lost on how to fix this.

- 2.) I made changes to `ShutdownFramework` which may not be acceptable

My test relies on leaking a surface/swapchain so I need to destroy an instance/device

However, ShutdownFramework() currently calls `debug_reporter_.error_monitor_.Finish();` which makes it impossible to use the existing framework with my testing. Unless I move the finish call to the destructor which should be fine?

Otherwise I'll need to write a lot more code to handle swapchain/surface creation/destruction to avoid altering the framework.

juan-lunarg:
> Swapchain's are created by the device, but for some reason are listed as under instance in `object_tracker.cpp`

For conxtext this is what I mean

```
// Provided by VK_KHR_swapchain
VkResult vkCreateSwapchainKHR(
    VkDevice                                    device,
    const VkSwapchainCreateInfoKHR*             pCreateInfo,
    VkSwapchainKHR*                             pSwapchain);
```

tests/vkrenderframework.cpp
```diff
@@ -890,7 +891,6 @@ bool VkRenderFramework::InitSurface(VkSurfaceKHR &surface) {
 #endif
 
 #if defined(VK_USE_PLATFORM_XLIB_KHR)
-    assert(m_surface_dpy == nullptr);
```
jeremyg-lunarg, line 893: why is removing the asserts necessary? Are surfaces getting re-initialized?

tests/vklayertests_wsi.cpp
```diff
@@ -1840,6 +1840,34 @@ TEST_F(VkLayerTest, DisplayPresentInfoSrcRect) {
...
+
+    VkSurfaceKHR surface = VK_NULL_HANDLE;
+    VkSwapchainKHR swapchain = VK_NULL_HANDLE;
+    ASSERT_TRUE(InitSurface(surface));
+    ASSERT_TRUE(InitSwapchain(surface, VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT, VK_SURFACE_TRANSFORM_IDENTITY_BIT_KHR, swapchain));
+
+    // Warn about the surface/swapchain not being destroyed
+    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-vkDestroyInstance-instance-00629");
+    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-vkDestroyDevice-device-00378");
+    ShutdownFramework();  // Destroy Instance/Device
```
jeremyg-lunarg, line 1867: This is cool and probably there are other tests that make their own device / instances because of the problem you fixed here.

layers/generated/object_tracker.cpp
```diff
@@ -48,6 +47,7 @@ bool ObjectLifetimes::ReportUndestroyedInstanceObjects(VkInstance instance, cons
 }
 bool ObjectLifetimes::ReportUndestroyedDeviceObjects(VkDevice device, const std::string& error_code) const {
     bool skip = false;
+    skip |= ReportLeakedDeviceObjects(device, kVulkanObjectTypeSwapchainKHR, error_code);
```
jeremyg-lunarg, line 50: The changes to this file look right to me, you 'just' gotta make the script do them.

jeremyg-lunarg: I'm pretty sure [this is the line](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/scripts/object_tracker_generator.py#L812) you need to change

tests/vkrenderframework.cpp
juan-lunarg: Currently InitSurface relies on memory leaks to utilize more than 1 surface. It wasn't just WIN32 see line 879

juan-lunarg: This is the problem I tried to highlight in my previous PR regarding `PresentIdWait`.

I'm thinking having a vector of `wsi_window` objects that get cleaned up. But I'm not sure if that's overkill/desirable.

layers/generated/object_tracker.cpp
juan-lunarg: That code is responsible for another section of the code generation: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/2326

juan-lunarg: I believe I found the problem causing the incorrect code gen. And a workaround.

scripts/object_tracker_generator.py
```diff
@@ -394,6 +394,9 @@ def beginFile(self, genOpts):
         self.type_categories = GetTypeCategories(self.registry.tree)
         self.is_aliased_type = GetHandleAliased(self.registry.tree)
 
+        # Fix the parent of VkSwapchainKHR
+        self.handle_parents['VkSwapchainKHR'] = 'VkDevice'
```
juan-lunarg, line 398: Currently vk.xml has this line:

```
<type category="handle" parent="VkSurfaceKHR"     objtypeenum="VK_OBJECT_TYPE_SWAPCHAIN_KHR"><type>VK_DEFINE_NON_DISPATCHABLE_HANDLE</type>(<name>VkSwapchainKHR</name>)</type>
```

The parent should be VkDevice. Not VkSurfaceKHR.

tests/vklayertests_wsi.cpp
juan-lunarg: There are a decent amount of tests that make their own device/instance because of the current setup.

tests/vkrenderframework.cpp
jeremyg-lunarg: oh ok. probably having an object that represents a window and surface together and does the clean up is the best eventual solution.

scripts/object_tracker_generator.py
jeremyg-lunarg: oof. Good find.

jeremyg-lunarg (approved):
LGTM. @ncesario-lunarg can you please double check that the ShutdownFramework() change doesn't undo any of the test cleanups that were made over the last couple of months?

scripts/object_tracker_generator.py
charles-lunarg: I'm not sure this is a spec bug? There is no indication that the parent *has* to be a dispatchable handle. And to an extent, the surface is the parent to the swapchain, just not the direct parent (cause swapchain is a device level object).

So I'm fine with declaring this is a spec bug, since the immediate parent is a device, just not 100% sure of that.

juan-lunarg: I'm not 100% sure either. I made a MR and will see what happens.

ncesario-lunarg (approved):
One question, but LGTM!

tests/vkrenderframework.cpp
```diff
@@ -306,7 +306,10 @@ VkRenderFramework::VkRenderFramework()
     m_clear_color.float32[3] = 0.0f;
 }
 
-VkRenderFramework::~VkRenderFramework() { ShutdownFramework(); }
+VkRenderFramework::~VkRenderFramework() {
+    ShutdownFramework();
+    debug_reporter_.error_monitor_.Finish();
```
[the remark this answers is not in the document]
juan-lunarg: Yes they still get reported and cause the test to fail.

```
[ RUN      ] VkLayerTest.LeakASwapchain
Unexpected: Validation Error: [ VUID-vkDestroyDevice-device-00378 ] Object 0: handle = 0x55555e2a6300, type = VK_OBJECT_TYPE_DEVICE; Object 1: handle = 0xcb3ee80000000007, type = VK_OBJECT_TYPE_SWAPCHAIN_KHR; | MessageID = 0x71500fba | OBJ ERROR : For VkDevice 0x55555e2a6300[], VkSwapchainKHR 0xcb3ee80000000007[] has not been destroyed. The Vulkan spec states: All child objects created on device must have been destroyed prior to destroying device (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkDestroyDevice-device-00378)
/home/juan/projects/layers/tests/vkrenderframework.cpp:132: Failure
Failed
Received unexpected error.
Unexpected: Validation Error: [ VUID-vkDestroyInstance-instance-00629 ] Object 0: handle = 0x55555da85b70, type = VK_OBJECT_TYPE_INSTANCE; Object 1: handle = 0xf443490000000006, type = VK_OBJECT_TYPE_SURFACE_KHR; | MessageID = 0x8b3d8e18 | OBJ ERROR : For VkInstance 0x55555da85b70[], VkSurfaceKHR 0xf443490000000006[] has not been destroyed. The Vulkan spec states: All child objects created using instance must have been destroyed prior to destroying instance (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkDestroyInstance-instance-00629)
/home/juan/projects/layers/tests/vkrenderframework.cpp:132: Failure
Failed
Received unexpected error.
[  FAILED  ] VkLayerTest.LeakASwapchain (4673 ms)
```

PR #4731, opened 2022-10-30
Add VUID 06417

spencer-lunarg:
` VUID-VkDeviceImageMemoryRequirementsKHR-pCreateInfo-06417`

### DIFF
--- a/layers/parameter_validation_utils.cpp
+++ b/layers/parameter_validation_utils.cpp
@@ -8688,9 +8688,9 @@ bool StatelessValidation::ValidateDeviceImageMemoryRequirements(VkDevice device,
 
         if ((pInfo->pCreateInfo->flags & VK_IMAGE_CREATE_DISJOINT_BIT) != 0) {
             if (FormatIsMultiplane(pInfo->pCreateInfo->format) && (pInfo->planeAspect == VK_IMAGE_ASPECT_NONE_KHR)) {
-                skip |=
-                    LogError(device, "VUID-VkDeviceImageMemoryRequirementsKHR-pCreateInfo-06417",
-                             "%s(): Can not use VK_IMAGE_ASPECT_NONE_KHR with a multiplanar format with disjoint flag.", func_name);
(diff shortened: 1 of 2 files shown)


layers/parameter_validation_utils.cpp
```diff
@@ -8684,6 +8684,14 @@ bool StatelessValidation::ValidateDeviceImageMemoryRequirements(VkDevice device,
                              "%s(): pInfo->pCreateInfo->pNext chain contains VkImageDrmFormatModifierExplicitCreateInfoEXT.",
                              func_name);
         }
+
+        if ((pInfo->pCreateInfo->flags & VK_IMAGE_CREATE_DISJOINT_BIT) != 0) {
+            if (FormatIsMultiplane(pInfo->pCreateInfo->format) && (pInfo->planeAspect == VK_IMAGE_ASPECT_NONE_KHR)) {
+                skip |=
+                    LogError(device, "VUID-VkDeviceImageMemoryRequirementsKHR-pCreateInfo-06417",
+                             "%s(): Can not use VK_IMAGE_ASPECT_NONE_KHR with a multiplanar format with disjoint flag.", func_name);
```
juan-lunarg, line 8692: I think this reads a bit easier.

```suggestion
                             "%s(): Must not specify VK_IMAGE_ASPECT_NONE_KHR with a multi-planar format and disjoint flag", func_name);
```

juan-lunarg (approved):
LGTM

layers/parameter_validation_utils.cpp
juan-lunarg: I realize we don't do it consistently but I think it would be good practice to use `Must` in the error messages if the VUID does. Since it's part of the vulkan specification [normative terminology](https://registry.khronos.org/vulkan/specs/1.3/html/chap2.html#introduction-normative-terminology)

spencer-lunarg: agree 100%, good catch

PR #4737, opened 2022-10-31
tests: Fix WSI memory leaks

juan-lunarg:
Fix memory leaks caused by InitSurface


tests/vkrenderframework.cpp
```diff
@@ -876,11 +867,9 @@ bool VkRenderFramework::InitSurface(VkSurfaceKHR &surface) {
     surface_create_info.hinstance = window_instance;
     surface_create_info.hwnd = window;
     VkResult err = vk::CreateWin32SurfaceKHR(instance(), &surface_create_info, nullptr, &surface);
```
jeremyg-lunarg, line 869: If any of the vkCreate*SurfaceKHR() calls fail it looks like we could still leak stuff.

tests/vkrenderframework.h
```diff
@@ -350,18 +350,20 @@ class VkRenderFramework : public VkTestFramework {
     VkFramebufferCreateInfo m_framebuffer_info;
     std::vector<VkImageView> m_framebuffer_attachments;
 
-    // WSI items
     VkSurfaceKHR m_surface = VK_NULL_HANDLE;
+    struct WsiHandle {
```
jeremyg-lunarg, line 354: Could this struct use RAII to make sure things get cleaned up?  The downside is that then you'd need to make sure it doesn't unintentionally get copied.

juan-lunarg: What about when the constructor fails? That was what I had trouble with. Since swapchain code can just fail on headless machines, and our tests just skip when the swapchain can't be constructed.

At AMD we worked around this by just having the constructor be a no-op, and then have an init function you check the return value of.

```
WsiHandle() = default;
bool Init(); // check if Init succeeded
```

Since we don't throw.

jeremyg-lunarg: That would work. Really as long as the members are initialized to nullptr (or the windows handle equivalent) in the constructor and non-null members are destroyed in the destructor, I think you could set the 'real' pointers however you wanted.

tests/vkrenderframework.cpp
juan-lunarg: Address possible leak.

PR #4738, opened 2022-11-01
corechecks: Add VUID 03881 checking

jeremyg-lunarg:
This specifies that within a single queue submission, any waits on a timeline semaphore must use a lower value than any signals for that same semaphore. This is only a required for vkQueueSubmit2(), but it could also apply to vkQueueSubmit() and vkQueueBindSparse(), if we wanted to make an UNASSIGNED check for it.


ncesario-lunarg (approved):
LGTM!

PR #4739, opened 2022-11-01
 corechecks: Improve timeline semaphore VUID checks

jeremyg-lunarg:
Restructure timeline semaphore checking so that it is easier to be aware of of timeline values within a single submission.
Redo vkQueueBindSparse() to use the common semaphore submission checking code.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -5078,14 +5078,30 @@ void CoreChecks::PostCallRecordQueueSubmit2(VkQueue queue, uint32_t submitCount,
     RecordQueueSubmit2(queue, submitCount, pSubmits, fence, result);
 }
 
+struct TimelineMaxDiffCheck {
+    TimelineMaxDiffCheck(uint64_t value_, uint64_t max_diff_) : value(value_), max_diff(max_diff_) {}
+
+    // compute the differents between 2 timeline values, without rollover if the difference is greater than INT64_MAX
+    uint64_t AbsDiff(uint64_t a, uint64_t b) { return a > b ? a - b : b - a; }
+
+    bool operator()(const SEMAPHORE_STATE::SemOp &op, bool is_pending) { return AbsDiff(value, op.payload) > max_diff; }
+
+    uint64_t value;
+    uint64_t max_diff;
+};
+
 struct SemaphoreSubmitState {
     const CoreChecks *core;
+    VkQueue queue;
     VkQueueFlags queue_flags;
     layer_data::unordered_set<VkSemaphore> signaled_semaphores;
     layer_data::unordered_set<VkSemaphore> unsignaled_semaphores;
     layer_data::unordered_set<VkSemaphore> internal_semaphores;
+    layer_data::unordered_map<VkSemaphore, uint64_t> timeline_signals;
+    layer_data::unordered_map<VkSemaphore, uint64_t> timeline_waits;
 
-    SemaphoreSubmitState(const CoreChecks *core_, VkQueueFlags queue_flags_) : core(core_), queue_flags(queue_flags_) {}
+    SemaphoreSubmitState(const CoreChecks *core_, VkQueue q_, VkQueueFlags queue_flags_)
+        : core(core_), queue(q_), queue_flags(queue_flags_) {}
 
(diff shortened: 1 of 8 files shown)


Tony-LunarG (approved):
Tricky stuff, but LGTM.  No test for 03245?

jeremyg-lunarg:
> Tricky stuff, but LGTM. No test for 03245?

Oooh good point. And I think that the code to check that would output the wrong VUID as it is currently written.

jeremyg-lunarg:
> Tricky stuff, but LGTM. No test for 03245?

This should be fixed now. I pushed 2 commits onto the PR, the previous commits didn't change.

layers/core_validation.cpp
```diff
@@ -5078,14 +5078,28 @@ void CoreChecks::PostCallRecordQueueSubmit2(VkQueue queue, uint32_t submitCount,
     RecordQueueSubmit2(queue, submitCount, pSubmits, fence, result);
 }
 
+struct TimelineMaxDiffCheck {
+    TimelineMaxDiffCheck(uint64_t value_, uint64_t max_diff_) : value(value_), max_diff(max_diff_) {}
+    uint64_t TimelineDiff(uint64_t a, uint64_t b) { return a > b ? a - b : b - a; }
+
+    bool operator()(const SEMAPHORE_STATE::SemOp &op, bool is_pending) { return TimelineDiff(value, op.payload) > max_diff; }
```
spencer-lunarg, line 5085: This is the only place calling `TimelineDiff` and to be honest, at first glance I don't know why it is ` a > b ? a - b : b - a;`

Might be better to spell out the reasoning behind this logic in this `operator()`

ncesario-lunarg: `std::abs`?

jeremyg-lunarg: Yep that's what it is but it deserves a comment.

ncesario-lunarg (approved):
Couple comments/questions, but LGTM.

tests/vklayertests_others.cpp
```diff
@@ -7848,140 +7847,486 @@ TEST_F(VkLayerTest, QueueSubmitTimelineSemaphoreBadValue) {
...
 
+    // Check for re-signalling an already completed value (5)
     m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkSubmitInfo-pSignalSemaphores-03242");
-    vk::QueueSubmit(m_device->m_queue, 1, submit_info, VK_NULL_HANDLE);
+    vk::QueueSubmit(m_device->m_queue, 1, &submit_info, VK_NULL_HANDLE);
     m_errorMonitor->VerifyFound();
 
-    vk::DestroySemaphore(m_device->device(), semaphore, nullptr);
+    // Submit (6)
+    signalValue++;
```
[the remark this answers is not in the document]
jeremyg-lunarg: I don't think that struct gets used  after the one call above. I'm adding scoping to check (and will update it if I can).  Really I should make semaphore.signal(signalValue); but that's another PR.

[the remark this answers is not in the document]
jeremyg-lunarg: scopes added to show that this a single use struct.

layers/core_validation.cpp
jeremyg-lunarg: I added comments and renamed the diff method to AbsDiff()

layers/core_validation.cpp
```diff
@@ -5247,21 +5331,21 @@ bool CoreChecks::ValidateSemaphoresForSubmit(SemaphoreSubmitState &state, VkQueu
                                  "not include an instance of VkTimelineSemaphoreSubmitInfo",
                                  loc.Message().c_str(), report_data->FormatHandle(semaphore).c_str());
                 continue;
```
spencer-lunarg, line 5333: why do these `continue`, seems like it will just end up spamming the same error for how big `waitSemaphoreCount`/`signalSemaphoreCount` it is

spencer-lunarg (approved):
lgtm

layers/core_validation.cpp
jeremyg-lunarg: good point.

PR #4744, opened 2022-11-02
layers: Fix validation of raytracing pipeline creation with null shader module

VaderY:
Fix for false positive VUID-VkPipelineShaderStageCreateInfo-module-parameter error.

VUID-VkPipelineShaderStageCreateInfo-module-parameter allows the shader module to be null everywhere else except for raytracing pipelines.

### DIFF
--- a/layers/object_tracker_utils.cpp
+++ b/layers/object_tracker_utils.cpp
@@ -1104,7 +1104,7 @@ bool ObjectLifetimes::PreCallValidateCreateRayTracingPipelinesKHR(VkDevice devic
         for (uint32_t index0 = 0; index0 < createInfoCount; ++index0) {
             if (pCreateInfos[index0].pStages) {
                 for (uint32_t index1 = 0; index1 < pCreateInfos[index0].stageCount; ++index1) {
-                    skip |= ValidateObject(pCreateInfos[index0].pStages[index1].module, kVulkanObjectTypeShaderModule, false,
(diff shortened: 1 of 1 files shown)


layers/core_validation.cpp
```diff
@@ -19592,7 +19592,7 @@ bool CoreChecks::PreCallValidateCreateSampler(VkDevice device, const VkSamplerCr
                                  "customBorderColorWithoutFormat feature being enabled");
             }
         }
-        if (custom_border_color_sampler_count >= phys_dev_ext_props.custom_border_color_props.maxCustomBorderColorSamplers) {
+        if (custom_border_color_sampler_count > phys_dev_ext_props.custom_border_color_props.maxCustomBorderColorSamplers) {
```
spencer-lunarg, line 19595: so the issue is `custom_border_color_sampler_count` is updated "after" `vkCreateSampler` is called successfully, while this is checking it prior to being called

so if `maxCustomBorderColorSamplers` is `2`

```
// custom_border_color_sampler_count = 0
vkCreateSampler();
// custom_border_color_sampler_count = 1
vkCreateSampler();
// custom_border_color_sampler_count = 2
vkCreateSampler(); // invalid, throw 04012
```

So I don't think this change is correct

VaderY: Fair enough, Thank you. I removed that commit

PR #4745, opened 2022-11-02
Move pool destroy in GPU-AV to after state tracker call

Tony-LunarG:
Also remove an assert that an app leaking in flight command buffers can reach and trigger
Fixes a crash playing back a game trace with GPU-AV


jeremyg-lunarg (approved):
LGTM

Issue #2918, opened 2021-06-28
[Best Practices] Validate invalid access/layout combinations for VkImageMemoryBarrier{2KHR}


Tobski:
**Describe the situation in which you encountered the missing validation**
There are a number of situations in which the set of access flags in the source/destination mask make no sense with an old/new layout in an image memory barrier. While these are not invalid by the spec, it would be helpful to check that the access mask matches up with the specified layout and issue a warning if they don't match.

The following table lists the set of "expected" access flags for a given image layout - anything outside of this should issue a warning.

| Layout | "Valid" access bits |
| ------ | ------ |
| UNDEFINED | Any (can be 0) |
| GENERAL | Any (must not be 0) |
| COLOR_ATTACHMENT_OPTIMAL | COLOR_ATTACHMENT_READ_BIT<br>COLOR_ATTACHMENT_WRITE_BIT<br>COLOR_ATTACHMENT_READ_NONCOHERENT_BIT_EXT |
| DEPTH_STENCIL_ATTACHMENT_OPTIMAL | DEPTH_STENCIL_ATTACHMENT_READ_BIT<br>DEPTH_STENCIL_ATTACHMENT_WRITE_BIT |
| DEPTH_STENCIL_READ_ONLY_OPTIMAL | SHADER_READ_BIT<br>DEPTH_STENCIL_ATTACHMENT_READ_BIT |
| SHADER_READ_ONLY_OPTIMAL | INPUT_ATTACHMENT_READ_BIT<br>SHADER_READ_BIT |
| TRANSFER_SRC_OPTIMAL | TRANSFER_READ_BIT |
| TRANSFER_DST_OPTIMAL | TRANSFER_WRITE_BIT |
| PREINITIALIZED | HOST_WRITE_BIT |
| DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL | Depth Aspect:<br>SHADER_READ_BIT<br>DEPTH_STENCIL_ATTACHMENT_READ_BIT<br><br>Stencil Aspect:<br>DEPTH_STENCIL_ATTACHMENT_READ_BIT<br>DEPTH_STENCIL_ATTACHMENT_WRITE_BIT |
[...]

PR #4746, opened 2022-11-02
practices: Fix BestPractices-ImageBarrierAccessLayout again

### LINKED ISSUE #4735: Incorrect best practices message BestPractices-ImageBarrierAccessLayout
RAR and WAR hazards don't require a memory barrier (https://github.com/KhronosGroup/Vulkan-Docs/wiki/Synchronization-Examples#first-dispatch-reads-from-a-storage-buffer-second-dispatch-writes-to-that-storage-buffer)

So if the image layout is e.g. `VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL`, it's perfectly fine for `srcAccessMask` to be 0. Best practices could warn about this for `dstAccessMask`, but shouldn't for `srcAccessMask`.

Also, there is syncval, what's the point of this?

jeremyg-lunarg:
It turns out that the table from #2918 had an implicit "(can be 0)" on almost every line. The only weird case is:

- PRESENT_SRC_KHR - Must be 0

Fixes #4735

### DIFF
--- a/layers/best_practices_utils.cpp
+++ b/layers/best_practices_utils.cpp
@@ -1615,15 +1615,13 @@ bool BestPractices::ValidateAccessLayoutCombination(const std::string& api_name,
                                                     VkImageAspectFlags aspect) const {
     bool skip = false;
 
-    const VkAccessFlags all = VK_ACCESS_FLAG_BITS_MAX_ENUM;
-    bool none_allowed = false;
-    VkAccessFlags allowed = 0;
+    const VkAccessFlags2 all = UINT64_MAX;  // core validation is responsible for detecting undefined flags.
+    VkAccessFlags2 allowed = 0;
 
     // Combinations taken from https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/2918
     switch (layout) {
         case VK_IMAGE_LAYOUT_UNDEFINED:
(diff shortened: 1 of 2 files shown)


layers/best_practices_utils.cpp
```diff
@@ -1615,18 +1615,18 @@ bool BestPractices::ValidateAccessLayoutCombination(const std::string& api_name,
...
 
     // Combinations taken from https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/2918
     switch (layout) {
         case VK_IMAGE_LAYOUT_UNDEFINED:
             allowed = all;
-            none_allowed = true;
             break;
         case VK_IMAGE_LAYOUT_GENERAL:
             allowed = all;
+            none_allowed = false;  // PR table says "Any (must not be 0)"
```
Tobski, line 1629: Looking back at my table and per my comment on the new issue this is incorrect, this is allowed to be zero because of semaphore interactions just like everything else, so this whole "none_allowed" path should be removed per my comment.

Tobski (changes requested):
One comment as mentioned on the issue, but otherwise LGTM.

Tobski (approved):
LGTM, thanks!

PR #4752, opened 2022-11-03
build: Update to header 1.3.233

mikes-lunarg:
- Update known-good
- Remove CoreBuiltinsARM from spirv_validation_generator capability exclude list and add the corresponding feature struct to state_tracker
- Add ShaderInvocationReorderNV to spirv_validation_generator capability exclude list
- Re-number VUID 01276 to be covered under the existing VUID 01427
- Generate source


ziga-lunarg (approved):

ncesario-lunarg (approved):
LGTM!

PR #4754, opened 2022-11-04
Import Operations Minor Cleanup

juan-lunarg:
Cleanup I've done while investigating `VkLayerTest.ValidateImportMemoryHandleType`

Split into 2 commits

### DIFF
--- a/layers/parameter_validation_utils.cpp
+++ b/layers/parameter_validation_utils.cpp
@@ -29,11 +29,60 @@
 
 static const int kMaxParamCheckerStringLength = 256;
 
+namespace {
 template <typename T>
 inline bool in_inclusive_range(const T &value, const T &min, const T &max) {
     // Using only < for generality and || for early abort
     return !((value < min) || (max < value));
 }
+
+struct ImportOperationsInfo {
+    const VkImportMemoryHostPointerInfoEXT *host_pointer_info_ext;
+    uint32_t total_import_ops;
+};
+
+ImportOperationsInfo GetNumberOfImportInfo(const VkMemoryAllocateInfo *pAllocateInfo) {
+    uint32_t count = 0;
+
+#ifdef VK_USE_PLATFORM_WIN32_KHR
+    // VkImportMemoryWin32HandleInfoKHR with a non-zero handleType value
+    auto import_memory_win32_handle = LvlFindInChain<VkImportMemoryWin32HandleInfoKHR>(pAllocateInfo->pNext);
+    count += (import_memory_win32_handle && import_memory_win32_handle->handleType);
+#endif
+
+    // VkImportMemoryFdInfoKHR with a non-zero handleType value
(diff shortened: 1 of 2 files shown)


layers/parameter_validation_utils.cpp
```diff
@@ -29,12 +29,59 @@
...
+
+#ifdef VK_USE_PLATFORM_FUCHSIA
+    // VkImportMemoryZirconHandleInfoFUCHSIA with a non-zero handleType value
+    auto import_zircon_fushcia = LvlFindInChain<VkImportMemoryZirconHandleInfoFUCHSIA>(pAllocateInfo->pNext);
+    count += (import_zircon_fushcia && import_zircon_fushcia->handleType);
+
+    // VkImportMemoryBufferCollectionFUCHSIA
+    auto import_buffer_collection_fushcia = LvlFindInChain<VkImportMemoryBufferCollectionFUCHSIA>(pAllocateInfo->pNext);
+    count += (import_buffer_collection_fushcia && import_buffer_collection_fushcia->);
+#endif
```
juan-lunarg, line 75: + @jbauman42 for commentary on Fuschia

juan-lunarg: Is there a reason we don't check the `VkBufferCollectionFUCHSIA` handle to be consistent with the other checks for import operations?

jbauman42: Thanks for checking with me. There's no handleType on `VkImportMemoryBufferCollectionFUCHSIA`, so we can't check that, and from the "Valid Usage (Implicit)" `collection` has to always be valid (though maybe it makes sense to  check that here).

layers/parameter_validation_utils.cpp
```diff
@@ -29,12 +29,59 @@
...
+
+#ifdef VK_USE_PLATFORM_ANDROID_KHR
+    // VkImportAndroidHardwareBufferInfoANDROID with a non-NULL buffer value
+    auto import_memory_ahb = LvlFindInChain<VkImportAndroidHardwareBufferInfoANDROID>(pAllocateInfo->pNext);
+    count += (import_memory_ahb && import_memory_ahb->buffer);
+#endif
+
+#ifdef VK_USE_PLATFORM_FUCHSIA
+    // VkImportMemoryZirconHandleInfoFUCHSIA with a non-zero handleType value
+    auto import_zircon_fushcia = LvlFindInChain<VkImportMemoryZirconHandleInfoFUCHSIA>(pAllocateInfo->pNext);
```
jbauman42, line 69: It should be spelled "fuchsia", not  "fushcia" - a very easy mistake to make unless you're typing it a million times a day.

layers/parameter_validation_utils.cpp
juan-lunarg: That clears things up. I'll add a note explaining this. Since the function reads strangely if you don't know this.

jeremyg-lunarg (approved):
LGTM as long as @jbauman42  is ok with the fuchsia part

jbauman42 (approved):
The Fuchsia changes lgtm.

PR #4755, opened 2022-11-04
syncval: Improve access log information

jzulauf-lunarg:
Enables additional handle information for each tagged access.  Limited usage to test, and as exmplar.

The specific additional information was added as when debugging a vkquake hazard, there just wasn't the needed information.  More is needed, for example the layout transitions, loads, and stores, could be individually tagged with the affected view handles and attachment id captured. But that would require plumbing down the tag generation and access log info into the renderpass access context.

Some of the ideas may be of general interest to improve VVL message generation, both the NamedHandle captures and the Formatter/FormatterState to enable "streamed out which needs external referants.

### DIFF
--- a/layers/synchronization_validation.cpp
+++ b/layers/synchronization_validation.cpp
@@ -255,6 +255,24 @@ std::ostream &operator<<(std::ostream &out, const SyncNodeFormatter &formatter) 
     return out;
 }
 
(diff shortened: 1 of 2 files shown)


jeremyg-lunarg (approved):
LGTM

PR #4756, opened 2022-11-04
tests: Fix ValidateImportMemoryHandleType

juan-lunarg:
Code doesn't properly check if external memory supports dedicated allocation

I used [`VkPositiveLayerTest.ExternalMemory`](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/226b595aeec0dc014bfc610f1df986eaf78f3578/tests/positive/image_buffer.cpp#L977) as the primary reference for correctness.

### DIFF
--- a/tests/vklayertests_others.cpp
+++ b/tests/vklayertests_others.cpp
@@ -10320,36 +10320,37 @@ TEST_F(VkLayerTest, ValidateImportMemoryHandleType) {
 #endif
     const auto wrong_handle_type = VK_EXTERNAL_MEMORY_HANDLE_TYPE_HOST_ALLOCATION_BIT_EXT;
 
-    // Check for external memory instance extensions
-    std::vector<const char *> reqd_instance_extensions = {
-        {VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME, VK_KHR_EXTERNAL_MEMORY_CAPABILITIES_EXTENSION_NAME}};
-    for (auto extension_name : reqd_instance_extensions) {
-        if (InstanceExtensionSupported(extension_name)) {
-            m_instance_extension_names.push_back(extension_name);
-        } else {
-            printf("%s Required instance extension %s not supported, skipping test\n", kSkipPrefix, extension_name);
-            return;
-        }
-    }
+    AddRequiredExtensions(ext_mem_extension_name);
+    AddRequiredExtensions(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
+    AddRequiredExtensions(VK_KHR_EXTERNAL_MEMORY_CAPABILITIES_EXTENSION_NAME);
+    AddRequiredExtensions(VK_KHR_BIND_MEMORY_2_EXTENSION_NAME);
+    AddRequiredExtensions(VK_KHR_EXTERNAL_MEMORY_EXTENSION_NAME);
+    AddRequiredExtensions(VK_KHR_GET_MEMORY_REQUIREMENTS_2_EXTENSION_NAME);
+
+    AddOptionalExtensions(VK_KHR_DEDICATED_ALLOCATION_EXTENSION_NAME);
 
     ASSERT_NO_FATAL_FAILURE(InitFramework(m_errorMonitor));
 
-    auto vkGetPhysicalDeviceExternalBufferPropertiesKHR =
-        (PFN_vkGetPhysicalDeviceExternalBufferPropertiesKHR)vk::GetInstanceProcAddr(
-            instance(), "vkGetPhysicalDeviceExternalBufferPropertiesKHR");
+    if (!AreRequiredExtensionsEnabled()) {
+        GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported";
+    }
+
+    auto vkGetPhysicalDeviceExternalBufferPropertiesKHR = reinterpret_cast<PFN_vkGetPhysicalDeviceExternalBufferPropertiesKHR>(
+        vk::GetInstanceProcAddr(instance(), "vkGetPhysicalDeviceExternalBufferPropertiesKHR"));
+    ASSERT_TRUE(vkGetPhysicalDeviceExternalBufferPropertiesKHR != nullptr);
 
     // Check for import/export capability
     // export used to feed memory to test import
-    VkPhysicalDeviceExternalBufferInfoKHR ebi = {VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_EXTERNAL_BUFFER_INFO_KHR, nullptr, 0,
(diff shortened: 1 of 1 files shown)


tests/vklayertests_others.cpp
```diff
@@ -10353,12 +10353,16 @@ TEST_F(VkLayerTest, ValidateImportMemoryHandleType) {
...
     // Always use dedicated allocation
-    if (DeviceExtensionSupported(gpu(), nullptr, VK_KHR_DEDICATED_ALLOCATION_EXTENSION_NAME)) {
-        m_device_extension_names.push_back(VK_KHR_DEDICATED_ALLOCATION_EXTENSION_NAME);
-        m_device_extension_names.push_back(VK_KHR_GET_MEMORY_REQUIREMENTS_2_EXTENSION_NAME);
-    } else {
-        printf("%s Dedicated allocation extension not supported, skipping test\n", kSkipPrefix);
-        return;
+    {
+        const bool dedicated_allocation =
+            ebp.externalMemoryProperties.externalMemoryFeatures & VK_EXTERNAL_MEMORY_FEATURE_DEDICATED_ONLY_BIT;
```
juan-lunarg, line 10358: On my Nvidia machine that's failing `externalMemoryFeatures` only supports `VK_EXTERNAL_MEMORY_FEATURE_EXPORTABLE_BIT` and `VK_EXTERNAL_MEMORY_FEATURE_IMPORTABLE_BIT` yet we use the dedicated allocation extension regardless.

juan-lunarg (commented):
On my Nvidia machine that's failing externalMemoryFeatures only supports VK_EXTERNAL_MEMORY_FEATURE_EXPORTABLE_BIT and VK_EXTERNAL_MEMORY_FEATURE_IMPORTABLE_BIT yet we use the dedicated allocation extension regardless.

Now, I check if the feature is supported and use it if needed. This allows the test to pass/run on local Nvidia machine.

MarkY-LunarG (approved):
Looks good to me.  But I have no official say over this repo.

tests/vklayertests_others.cpp
```diff
@@ -10402,10 +10406,15 @@ TEST_F(VkLayerTest, ValidateImportMemoryHandleType) {
     auto alloc_info = vk_testing::DeviceMemory::get_resource_alloc_info(*m_device, buffer_export.memory_requirements(), mem_flags);
 
     // Add export allocation info to pNext chain
+    VkExportMemoryAllocateInfoKHR export_info = {VK_STRUCTURE_TYPE_EXPORT_MEMORY_ALLOCATE_INFO_KHR, nullptr, handle_type};
```
spencer-lunarg, line 10409: I would use `LvlInitStruct<VkExportMemoryAllocateInfoKHR>();`

tests/vklayertests_others.cpp
```diff
@@ -10352,13 +10352,17 @@ TEST_F(VkLayerTest, ValidateImportMemoryHandleType) {
...
+    // Check if dedicated allocation is required
+    const bool dedicated_allocation =
+        ebp.externalMemoryProperties.externalMemoryFeatures & VK_EXTERNAL_MEMORY_FEATURE_DEDICATED_ONLY_BIT;
+
+    if (dedicated_allocation) {
+        if (DeviceExtensionSupported(gpu(), nullptr, VK_KHR_DEDICATED_ALLOCATION_EXTENSION_NAME)) {
+            m_device_extension_names.push_back(VK_KHR_DEDICATED_ALLOCATION_EXTENSION_NAME);
+            m_device_extension_names.push_back(VK_KHR_GET_MEMORY_REQUIREMENTS_2_EXTENSION_NAME);
+        } else {
+            GTEST_SKIP() << "Dedicated allocation extension not supported, skipping test";
```
spencer-lunarg, line 10364: Is there a known device that supports external memory and not `VK_KHR_dedicated_allocation`? Curious why it was not easier to always assume and require `VK_KHR_dedicated_allocation` to be supported

juan-lunarg: My Nvidia GPU on Linux supports external memory but doesn't support dedicated allocation. I could just skip, but decided it would be useful to have the test run on more GPUs instead. Also it would be consistent with [VkPositiveLayerTest.ExternalMemory](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/226b595aeec0dc014bfc610f1df986eaf78f3578/tests/positive/image_buffer.cpp#L977)

tests/vklayertests_others.cpp
```diff
@@ -10352,13 +10352,17 @@ TEST_F(VkLayerTest, ValidateImportMemoryHandleType) {
...
-        printf("%s Dedicated allocation extension not supported, skipping test\n", kSkipPrefix);
-        return;
+    // Check if dedicated allocation is required
+    const bool dedicated_allocation =
+        ebp.externalMemoryProperties.externalMemoryFeatures & VK_EXTERNAL_MEMORY_FEATURE_DEDICATED_ONLY_BIT;
+
+    if (dedicated_allocation) {
+        if (DeviceExtensionSupported(gpu(), nullptr, VK_KHR_DEDICATED_ALLOCATION_EXTENSION_NAME)) {
+            m_device_extension_names.push_back(VK_KHR_DEDICATED_ALLOCATION_EXTENSION_NAME);
+            m_device_extension_names.push_back(VK_KHR_GET_MEMORY_REQUIREMENTS_2_EXTENSION_NAME);
```
ncesario-lunarg, line 10362: Ideally `AddOptionalExtensions` is used instead of modifying `m_device_extension_names` directly, then checking `IsExtensionsEnabled` to see if the requested extension is supported.

If that doesn't work for some reason here, we should probably figure out why and maybe modify the existing helper functions to handle it; otherwise, we're back to manually checking extension requirements.

ncesario-lunarg (commented):
One minor test nit aside from other foks' comments, but LGTM otherwise!

tests/vklayertests_others.cpp
juan-lunarg: `AddOptionalExtensions` modifies instance extensions. Doesn't seem correct to use for device extensions.

```
void VkRenderFramework::AddOptionalExtensions(const char *ext_name) {
    m_optional_extensions.push_back(ext_name);
    AddRequestedInstanceExtensions(ext_name);
}
```

ncesario-lunarg: That is to collect any instance extensions that `ext_name` depends on. `AddOptionalExtensions` is used for device extensions as well (search the `tests` directory for examples). You'll need to use it in the same way as `AddRequiredExtensions`; i.e., call before `InitFramework` and then check afterwards.

This is one of the things I'd like to streamline in the not-too-distant-future...

juan-lunarg: Updated PR to use `AddOptionalExtensions` and `AddRequiredExtensions`

spencer-lunarg (approved):

jeremyg-lunarg (approved):
LGTM

PR #4760, opened 2022-11-07
layers: Fix PIPELINE_STATE::GetStageStates bug

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4689

Better version of https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4690


scripts/parameter_validation_generator.py
```diff
@@ -705,11 +712,12 @@ def endFeature(self):
...
                 if flagBits in self.flagBits:
                     bits = self.flagBits[flagBits]
-                    decl = 'const {} All{} = {}'.format(flag, flagBits, bits[0])
-                    for bit in bits[1:]:
-                        decl += '|' + bit
-                    decl += ';'
+                    decl = 'const {} All{} = {};'.format(flag, flagBits, '|'.join(bits))
                     self.flag_values_definitions[flag] = Guarded(self.featureExtraProtect, decl)
+                    if flag in self.flagBitsAsArray:
+                        decl = 'const std::array<%s, %d> All%s = {%s};' % (flag, len(bits), flag, ','.join(bits))
```
arno-lunarg, line 718: Nit: maybe use formatted string literals instead (not 100% sure about the syntax):
```python
decl = 'const std::array<{flag}, {len(bits)}> All{flag} = {{{','.join(bits)}}}'
```

arno-lunarg (commented):
LGTM! Only one nit

scripts/parameter_validation_generator.py
spencer-lunarg: So yes, the python string format debate :laughing: 

I have zero preference how we `printf` in the scripts, but as long as it is consistent everywhere... which is far from currently. I just mimicked the styles in the function above and where the other `std::array` string is

There is a long needed TODO to align all these to a single style

juan-lunarg (approved):

layers/generated/enum_flag_bits.h
```diff
@@ -140,6 +140,7 @@ const VkOpticalFlowGridSizeFlagsNV AllVkOpticalFlowGridSizeFlagBitsNV = VK_OPTIC
...
 const VkAccelerationStructureCreateFlagsKHR AllVkAccelerationStructureCreateFlagBitsKHR = VK_ACCELERATION_STRUCTURE_CREATE_DEVICE_ADDRESS_CAPTURE_REPLAY_BIT_KHR|VK_ACCELERATION_STRUCTURE_CREATE_MOTION_BIT_NV;
+const std::array<VkShaderStageFlags, 25> AllVkShaderStageFlags = {VK_SHADER_STAGE_VERTEX_BIT,VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT,VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT,VK_SHADER_STAGE_GEOMETRY_BIT,VK_SHADER_STAGE_FRAGMENT_BIT,VK_SHADER_STAGE_COMPUTE_BIT,VK_SHADER_STAGE_ALL_GRAPHICS,VK_SHADER_STAGE_ALL,VK_SHADER_STAGE_RAYGEN_BIT_KHR,VK_SHADER_STAGE_ANY_HIT_BIT_KHR,VK_SHADER_STAGE_CLOSEST_HIT_BIT_KHR,VK_SHADER_STAGE_MISS_BIT_KHR,VK_SHADER_STAGE_INTERSECTION_BIT_KHR,VK_SHADER_STAGE_CALLABLE_BIT_KHR,VK_SHADER_STAGE_RAYGEN_BIT_NV,VK_SHADER_STAGE_ANY_HIT_BIT_NV,VK_SHADER_STAGE_CLOSEST_HIT_BIT_NV,VK_SHADER_STAGE_MISS_BIT_NV,VK_SHADER_STAGE_INTERSECTION_BIT_NV,VK_SHADER_STAGE_CALLABLE_BIT_NV,VK_SHADER_STAGE_TASK_BIT_NV,VK_SHADER_STAGE_MESH_BIT_NV,VK_SHADER_STAGE_TASK_BIT_EXT,VK_SHADER_STAGE_MESH_BIT_EXT,VK_SHADER_STAGE_SUBPASS_SHADING_BIT_HUAWEI};
```
ncesario-lunarg, line 143: Should `VK_SHADER_STAGE_ALL*` enums be in this list?

ncesario-lunarg (commented):

layers/generated/enum_flag_bits.h
spencer-lunarg: So for sake of being "generated with future headers without worrying what might be added" I think it should stay in `AllVkShaderStageFlags`

The users of `AllVkShaderStageFlags` should be aware of it, will add a note in the code consuming it

But it will not be an issue as the code using it is only looking for an exact match, not a bit boolean logic

PR #4763, opened 2022-11-07
layers: fix typo in semaphore export handle type checks

emersion:
These messages seem to be copy-pasted from the VkFence logic.


jeremyg-lunarg (commented):
LGTM once the commit subject is fixed. Reminder: you can run ./scripts/check_code_format.py locally to avoid hitting this in CI (but I forget about 50% of the time)

emersion:
Fixed!

PR #4764, opened 2022-11-07
tests: Rework require_success to REQUIRE_SUCCESS

jzulauf-lunarg:
So to handle the timeout in my acquire/present testing, initially had a lamba doing the timeout check, but didn't actually "FAIL" correctly (i.e. return from the test).

This refactors that logic into a macro which will, and incorporates cleanup support.  Relocated to tests_common as REQUIRE_SUCCESS is likely useful anywhere we're failing on timeout, as it includes that cleanup capability.

### DIFF
--- a/tests/test_common.h
+++ b/tests/test_common.h
@@ -134,4 +134,24 @@ static inline void test_error_callback(const char *expr, const char *file, unsig
     ADD_FAILURE_AT(file, line) << "Assertion: `" << expr << "'";
 }
 
+// Wrap FAIL:
+//  * DRY for common messages
+//  * for test stability reasons sometimes cleanup code is required *prior* to the return hidden in FAIL
+//  * result_arg_ *can* (should) have side-effect, but is referenced exactly once
+//  * label_ must be converitble to bool, and *should* *not* have side-effects
+//  * clean_ *can* (should) have side-effects
+//    * "{}" or ";" are valid clean_ values for noop
+#define REQUIRE_SUCCESS(result_arg_, label_, clean_)                    \
+    {                                                                   \
+        const VkResult result_ = (result_arg_);                         \
+        if (result_ != VK_SUCCESS) {                                    \
+            { clean_; }                                                 \
+            if (bool(label_)) {                                         \
+                FAIL() << string_VkResult(result_) << ": " << (label_); \
+            } else {                                                    \
+                FAIL() << string_VkResult(result_);                     \
+            }                                                           \
+        }                                                               \
+    }
+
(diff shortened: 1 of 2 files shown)


tests/test_common.h
```diff
@@ -134,4 +134,24 @@ static inline void test_error_callback(const char *expr, const char *file, unsig
...
+//  * result_arg_ *can* (should) have side-effect, but is referenced exactly once
+//  * label_ must be converitble to bool, and *should* *not* have side-effects
+//  * clean_ *can* (should) have side-effects
+//    * "{}" or ";" are valid clean_ values for noop
+#define REQUIRE_SUCCESS(result_arg_, label_, clean_)                    \
+    {                                                                   \
+        const VkResult result_ = (result_arg_);                         \
+        if (result_ != VK_SUCCESS) {                                    \
+            { clean_; }                                                 \
+            if (bool(label_)) {                                         \
```
jeremyg-lunarg, line 149: It seems like [SCOPED_TRACE()](https://github.com/google/googletest/blob/e07617d6c692a96e126f11f85c3e38e46b10b4d0/docs/reference/testing.md#scoped_trace-scoped_trace) might be another way to get the label out.

jzulauf-lunarg: FAIL already calls GTEST_MESSAGE_AT_(__FILE__, __LINE__, message, result_type), so SCOPED_TRACE would duplicate the file/line information.

tests/test_common.h
```diff
@@ -134,4 +134,24 @@ static inline void test_error_callback(const char *expr, const char *file, unsig
...
 }
 
+// Wrap FAIL:
+//  * DRY for common messages
+//  * for test stability reasons sometimes cleanup code is required *prior* to the return hidden in FAIL
+//  * result_arg_ *can* (should) have side-effect, but is referenced exactly once
+//  * label_ must be converitble to bool, and *should* *not* have side-effects
+//  * clean_ *can* (should) have side-effects
+//    * "{}" or ";" are valid clean_ values for noop
+#define REQUIRE_SUCCESS(result_arg_, label_, clean_)                    \
```
ncesario-lunarg, line 144: Would really rather not have a big, multi-line macro. Possible alternative:
```cpp
template<typename Label>
inline void RequireSuccess(const VkResult result, Label label = {}, std::function<void()> clean = {}) {
    if (result != VK_SUCCESS) {
        if (clean) { clean(); }
        const auto result_string = string_VkResult(result);
        if (label) {
            FAIL() << result_string << ": " << label;
        } else {
            FAIL() << result_string;
        }
    }
}
```

PR #1393, opened 2019-11-22
build: Fix Vulkan headers detection with CMake.

digit-google:
NOTE: This is a port of https://github.com/KhronosGroup/Vulkan-Loader/pull/286
      to Khronos/Vulkan-ValidationLayers

The logic used to find the location of the Vulkan headers and
registry used simple find_path() calls with HINTS. Unfortunately,
this doesn't work when using a custom toolchain file that sets
CMAKE_FIND_ROOT_PATH_INCLUDE to ONLY, which can happen when
cross-compiling with a custom sysroot.

This patch changes the logic so that:

If VULKAN_HEADERS_INSTALL_DIR is specified on the command-line,
or if it is set in the environment, or if VULKAN_SDK is set in
the environment, the corresponding paths will be searched for
the Vulkan headers and register, ignoring the
CMAKE_FIND_ROOT_PATH. This does not affect other parts of the
build.

Otherwise, use a regular find_path() call which may be restricted
to the CMAKE_FIND_ROOT_PATH depending on previous configuration
settings.

This should be equivalent to the previous behaviour, except when
CMAKE_FIND_ROOT_PATH_INCLUDE is set to ONLY.

PR #4765, opened 2022-11-07
cmake: Use new VulkanHeaders cmake support

juan-lunarg:
Now VulkanHeaders is finally consistent with all of the other dependencies in Vulkan-ValidationLayers :tada:


CMakeLists.txt
```diff
@@ -112,10 +112,8 @@ if (GOOGLETEST_INSTALL_DIR)
     list(APPEND CMAKE_PREFIX_PATH ${GOOGLETEST_INSTALL_DIR})
 endif()
 
-find_package(VulkanHeaders REQUIRED)
-add_library(Vulkan-Headers INTERFACE)
-target_include_directories(Vulkan-Headers INTERFACE ${VulkanHeaders_INCLUDE_DIRS})
-add_library(Vulkan::Headers ALIAS Vulkan-Headers)
+list(APPEND CMAKE_PREFIX_PATH "/home/juan/projects/pub-headers/build/install")
```
mikes-lunarg, line 115: What happens to the `VULKAN_HEADERS_INSTALL_DIR` option? Will that end up here?

juan-lunarg: See updated commit. It would be treated like the other dependencies.

layers/CMakeLists.txt
```diff
@@ -319,20 +319,14 @@ if(BUILD_LAYERS)
...
 
     # Fetch header version from vulkan_core.h ----------------------------------------------------------------------------------------
-    file(STRINGS "${VulkanHeaders_INCLUDE_DIRS}/vulkan/vulkan_core.h" lines REGEX "^#define VK_HEADER_VERSION [0-9]+")
-    list(LENGTH lines len)
-    if(${len} EQUAL 1)
-        string(REGEX MATCHALL "[0-9]+" vk_header_version ${lines})
-    else()
-        message(FATAL_ERROR "Unable to fetch version from vulkan_core.h")
-    endif()
+    set(vk_header_version "${VulkanHeaders_VERSION}")
```
juan-lunarg, line 322: Because of `VulkanHeadersConfigVersion.cmake` we don't have to parse the version anymore. <PackageName>_VERSION is provided by find_package now.

juan-lunarg:
@mikes-lunarg since we only use released tags for vulkan headers, we should not update our code yet until the next release correct?

mikes-lunarg:
> @mikes-lunarg since we only use released tags for vulkan headers, we should not update our code yet until the next release correct?

That tags are convenient but SHA is fine too. There is no reason to hold this up until the next header drop.

juan-lunarg:
> > @mikes-lunarg since we only use released tags for vulkan headers, we should not update our code yet until the next release correct?
> 
> The tags are convenient but SHA is fine too. There is no reason to hold this up until the next header drop.

Great I'll mark this ready for review then :+1:

cmake/VVLGenerateSourceCode.cmake
```diff
@@ -16,6 +16,7 @@
 find_package(PythonInterp 3 QUIET)
 
 if(PYTHONINTERP_FOUND)
+    get_target_property(VulkanRegistry_DIR Vulkan::Registry INTERFACE_INCLUDE_DIRECTORIES)
```
juan-lunarg, line 19: This is the only think I would label as awkward. Since Vulkan::Registry is treated as a header only library, so to grab the registry directory we grab the include directory.

layers/CMakeLists.txt
```diff
@@ -318,21 +316,12 @@ if(BUILD_LAYERS)
...
 
-    # Fetch header version from vulkan_core.h ----------------------------------------------------------------------------------------
-    file(STRINGS "${VulkanHeaders_INCLUDE_DIRS}/vulkan/vulkan_core.h" lines REGEX "^#define VK_HEADER_VERSION [0-9]+")
-    list(LENGTH lines len)
-    if(${len} EQUAL 1)
-        string(REGEX MATCHALL "[0-9]+" vk_header_version ${lines})
-    else()
-        message(FATAL_ERROR "Unable to fetch version from vulkan_core.h")
-    endif()
-
```
juan-lunarg, line 329: This isn't needed because Vulkan-Headers now provides proper versioning support.

CMakeLists.txt
```diff
@@ -111,11 +111,11 @@ endif()
 if (GOOGLETEST_INSTALL_DIR)
     list(APPEND CMAKE_PREFIX_PATH ${GOOGLETEST_INSTALL_DIR})
 endif()
+if (VULKAN_HEADERS_INSTALL_DIR)
+    list(APPEND CMAKE_PREFIX_PATH ${VULKAN_HEADERS_INSTALL_DIR})
```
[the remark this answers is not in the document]
juan-lunarg: CMAKE_PREFIX_PATH takes [priority](https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure) over system/sys-root packages. So because VULKAN_HEADERS_INSTALL_DIR gets added to CMAKE_PREFIX_PATH it will look there first.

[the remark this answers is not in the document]
juan-lunarg: The CMake documentation covers the search procedure quite well: https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure

juan-lunarg:
> LGTM. It might be good to run this by @johnzupin to make sure this doesn't cause problems for building .debs

Check with @johnzupin, change seems fine. If anything Vulkan-Headers is now consistent with our repos now.

ncesario-lunarg (approved):
Wow, that change to Vulkan-Headers (that I completely missed) is awesome, as is this change!

CMakeLists.txt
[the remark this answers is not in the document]
mikes-lunarg: OK good to know. I think the Android NDK toolchain file does some funny stuff so my `find_package(VulkanHeaders)` didn't work until I added `-DCMAKE_FIND_ROOT_PATH_MODE_PACKAGE=BOTH` to my configure command. We've gotten traffic on this before (https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/1393) but I'm satisfied with the workaround.

johnzupin:
I know this was merged already but I tested building packages with this change and they built successfully.

PR #4774, opened 2022-11-09
layers: Remove unused ValidateViConsistency

spencer-lunarg:
I realized that `ValidateViConsistency` is not being checked because there is already 

> VUID-VkPipelineVertexInputStateCreateInfo-pVertexBindingDescriptions-00616
>
> All elements of pVertexBindingDescriptions must describe distinct binding numbers

That is validating this

Also while at it, took the `vkCmdSetVertexInputEXT` logic and made it consistent and improved the error message

### DIFF
--- a/layers/parameter_validation_utils.cpp
+++ b/layers/parameter_validation_utils.cpp
@@ -2266,8 +2266,8 @@ bool StatelessValidation::manual_PreCallValidateCreateGraphicsPipelines(VkDevice
                                          "vkCreateGraphicsPipelines: parameter "
                                          "pCreateInfo[%" PRIu32 "].pVertexInputState->pVertexBindingDescription[%" PRIu32
                                          "].binding "
-                                         "(%" PRIu32 ") is not distinct.",
-                                         i, d, vertex_bind_desc.binding);
+                                         "(%" PRIu32 ") is already in pVertexBindingDescription[%" PRIu32 "]",
+                                         i, d, vertex_bind_desc.binding, *binding_it);
                     }
                     vertex_bindings.insert(vertex_bind_desc.binding);
 
@@ -2299,9 +2299,9 @@ bool StatelessValidation::manual_PreCallValidateCreateGraphicsPipelines(VkDevice
                     if (location_it != attribute_locations.cend()) {
                         skip |= LogError(device, "VUID-VkPipelineVertexInputStateCreateInfo-pVertexAttributeDescriptions-00617",
                                          "vkCreateGraphicsPipelines: parameter "
-                                         "pCreateInfo[%" PRIu32 "].pVertexInputState->vertexAttributeDescriptions[%" PRIu32
-                                         "].location (%" PRIu32 ") is not distinct.",
-                                         i, d, vertex_attrib_desc.location);
+                                         "pCreateInfo[%" PRIu32 "].pVertexInputState->pVertexAttributeDescriptions[%" PRIu32
+                                         "].location (%" PRIu32 ") is already in pVertexAttributeDescriptions[%" PRIu32 "].",
+                                         i, d, vertex_attrib_desc.location, *location_it);
                     }
                     attribute_locations.insert(vertex_attrib_desc.location);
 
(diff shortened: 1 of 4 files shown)


layers/parameter_validation_utils.cpp
```diff
@@ -8354,31 +8354,34 @@ bool StatelessValidation::manual_PreCallValidateCmdSetVertexInputEXT(
...
-                    skip |= LogError(device, "VUID-vkCmdSetVertexInputEXT-pVertexBindingDescriptions-04794",
-                                     "vkCmdSetVertexInputEXT(): binding description for binding %" PRIu32 " already specified",
-                                     binding_value);
-                }
+    // check for distinct values
+    {
+        layer_data::unordered_set<uint32_t> vertex_bindings(vertexBindingDescriptionCount);
+        for (uint32_t i = 0; i < vertexBindingDescriptionCount; ++i) {
+            const uint32_t binding = pVertexBindingDescriptions[i].binding;
+            auto const &binding_it = vertex_bindings.find(binding);
```
juan-lunarg, line 8362: `find` doesn't return a const reference

layers/parameter_validation_utils.cpp
```diff
@@ -8354,31 +8354,34 @@ bool StatelessValidation::manual_PreCallValidateCmdSetVertexInputEXT(
...
-            for (uint32_t next_attribute = attribute + 1; next_attribute < vertexAttributeDescriptionCount; ++next_attribute) {
-                if (location == pVertexAttributeDescriptions[next_attribute].location) {
-                    skip |= LogError(device, "VUID-vkCmdSetVertexInputEXT-pVertexAttributeDescriptions-04795",
-                                     "vkCmdSetVertexInputEXT(): attribute description for location %" PRIu32 " already specified",
-                                     location);
-                }
+        layer_data::unordered_set<uint32_t> vertex_locations(vertexBindingDescriptionCount);
+        for (uint32_t i = 0; i < vertexBindingDescriptionCount; ++i) {
+            const uint32_t location = pVertexAttributeDescriptions[i].location;
+            auto const &location_it = vertex_locations.find(location);
```
juan-lunarg, line 8376: `find` doesn't return a const reference

juan-lunarg (dismissed):

layers/parameter_validation_utils.cpp
ncesario-lunarg: I believe it is legal to assign an rvalue to a const ref and this _can_ (??) avoid copy construction, so I think this is fine, but it's quite possible I'm not remembering this detail correctly.

ncesario-lunarg (commented):
LGTM once Juan's comments are resolved.

layers/parameter_validation_utils.cpp
juan-lunarg: @ncesario-lunarg is correct. I forgot the nuance for `const` references. [Source FWIW](https://stackoverflow.com/questions/22845167/const-reference-and-lvalue)

juan-lunarg (dismissed):

juan-lunarg (changes requested):
```
[ctest] The following tests FAILED:
[ctest] 	594 - VkLayerTest.ValidateVertexInputDynamicStateEnabled (SEGFAULT)
[ctest] 	595 - VkLayerTest.ValidateVertexInputDynamicStateDivisor (SEGFAULT)
```

layers/parameter_validation_utils.cpp
```diff
@@ -8354,31 +8354,34 @@ bool StatelessValidation::manual_PreCallValidateCmdSetVertexInputEXT(
...
-        for (uint32_t attribute = 0; attribute < vertexAttributeDescriptionCount - 1; ++attribute) {
-            uint32_t location = pVertexAttributeDescriptions[attribute].location;
-            for (uint32_t next_attribute = attribute + 1; next_attribute < vertexAttributeDescriptionCount; ++next_attribute) {
-                if (location == pVertexAttributeDescriptions[next_attribute].location) {
-                    skip |= LogError(device, "VUID-vkCmdSetVertexInputEXT-pVertexAttributeDescriptions-04795",
-                                     "vkCmdSetVertexInputEXT(): attribute description for location %" PRIu32 " already specified",
-                                     location);
-                }
+        layer_data::unordered_set<uint32_t> vertex_locations(vertexBindingDescriptionCount);
+        for (uint32_t i = 0; i < vertexBindingDescriptionCount; ++i) {
```
juan-lunarg, line 8374:
```suggestion
        layer_data::unordered_set<uint32_t> vertex_locations(vertexAttributeDescriptionCount);
        for (uint32_t i = 0; i < vertexAttributeDescriptionCount; ++i) {
```

juan-lunarg (changes requested):

juan-lunarg (approved):
LGTM

PR #4776, opened 2022-11-09
cmake: Remove BUILD_WSI_*

juan-lunarg:
When building for linux, we should always support all 3 of the WSI interfaces. That's currently what we do for all SDK releases.

Micro-optimizing the WSI libraries will only cause issues.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -310,9 +310,6 @@ on/off options currently supported by this repository:
 | BUILD_TESTS | All | `OFF` | Controls whether or not the validation layer tests are built. |
 | INSTALL_TESTS | All | `OFF` | Controls whether or not the validation layer tests are installed. This option is only available when `BUILD_TESTS` is `ON` |
 | BUILD_WERROR | All | `ON` | Controls whether or not to treat compiler warnings as errors. |
-| BUILD_WSI_XCB_SUPPORT | Linux | `ON` | Build the components with XCB support. |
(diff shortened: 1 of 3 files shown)


BUILD.md
```diff
@@ -541,7 +538,7 @@ repository to other Linux distributions.
 
     sudo apt-get install git build-essential libx11-xcb-dev \
         libxkbcommon-dev libwayland-dev libxrandr-dev \
-        libegl1-mesa-dev
+        libegl1-mesa-dev pkg-config
```
juan-lunarg, line 541: By default most distros already have pkg-config, but just in case there is a distro that doesn't

BUILD.md
```diff
@@ -628,16 +625,6 @@ You can also use
     cmake --build .
 
 ### Linux Notes
-
-#### WSI Support Build Options
```
juan-lunarg, line 632: While adding this customization sounds nice, it just splinters the linux deliverable for minimal gains, and maximum potential for confusion.

ncesario-lunarg (commented):
I actually do make use of these options as I will often only build with wayland support (which I find is pretty easy to get all dependencies from source as opposed to X as well) on a new linux system and avoid installing all of the X dependencies if I can. I agree it adds to maintenance, though I'm not sure how how much.

juan-lunarg:
closing for developer convenience

PR #4777, opened 2022-11-09
Fix out of scope variable in ValidationArrayOOBRayTracingShaders

Tony-LunarG:
Fixes crash.


juan-lunarg (approved):
Verified this fixes the test failure on my GTX 3050

ncesario-lunarg (approved):

jeremyg-lunarg (approved):
LGTM. Yikes that's a non-obvious problem. Thanks pNext!

PR #4779, opened 2022-11-10
layers: Fix vkDestroyQueryPool edge case

### LINKED ISSUE #4704: Crash due to validation of deleted VkQueryPool
**Describe the Issue**
I'm seeing a crash in the validation layer when trying to lock a mutex on the QUERY_POOL_STATE variable.

The crash call stack is this, with an access violation for address 0xC:
```
ntdll.dll!RtlAcquireSRWLockExclusive() + 16 bytes	Unknown
[Inline Frame] VkLayer_khronos_validation.dll!std::shared_mutex::lock() Line 49	C++
[Inline Frame] VkLayer_khronos_validation.dll!ReadWriteLock::lock() Line 381	C++
[Inline Frame] VkLayer_khronos_validation.dll!std::unique_lock<ReadWriteLock>::{ctor}(ReadWriteLock &) Line 150	C++
[Inline Frame] VkLayer_khronos_validation.dll!QUERY_POOL_STATE::WriteLock() Line 87	C++
[Inline Frame] VkLayer_khronos_validation.dll!QUERY_POOL_STATE::SetQueryState(unsigned int query, unsigned int perf_pass, QueryState) Line 63	C++
VkLayer_khronos_validation.dll!CMD_BUFFER_STATE::Retire(unsigned int perf_submit_pass, const std::function<bool __cdecl(QueryObject const &)> & is_query_updated_after) Line 1416	C++
VkLayer_khronos_validation.dll!QUEUE_STATE::Retire(unsigned __int64 until_seq) Line 157	C++
VkLayer_khronos_validation.dll!SEMAPHORE_STATE::RetireTimeline(unsigned __int64 payload) Line 364	C++
VkLayer_khronos_validation.dll!ValidationStateTracker::RecordWaitSemaphores(VkDevice_T * device, const VkSemaphoreWaitInfo * pWaitInfo, unsigned __int64 timeout, VkResult result) Line 1817	C++
VkLayer_khronos_validation.dll!vulkan_layer_chassis::WaitSemaphores(VkDevice_T * device, const VkSemaphoreWaitInfo * pWaitInfo, unsigned __int64 timeout) Line 4583	C++
 ```

I think I've narrowed down the order of operations leading to the crash, and believe this to be a bug of the validation layer.

1. vkCreateQueryPool
2. vkResetQueryPool
3. add queries to QueryPool (in this case vkCmdWriteAccelerationStructuresPropertiesKHR)
4. submit command buffer
5. (one frame later) vkGetQueryPoolResults which returns success as the GPU is done already
6. vkDestroyQueryPool
7. vkWaitSemaphores on semaphore that indicates command buffer in step 4 is done -> crash

The [Vulkan spec](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkDestroyQueryPool.html) says it is ok to destroy the query pool once "vkGetQueryPoolResults() without the VK_QUERY_RESULT_PARTIAL_BIT flag returns VK_SUCCESS", but I suspect the validation layer is still referencing the pool until the command buffer is verified to be completed, hence crashing in the retiring of the command buffer above.

**Environment:**
 - OS: Windows 10
 - GPU: NVIDIA A6000
 - SDK or header version if building from repo: 1.3.224.0
 - Options enabled (synchronization, best practices, etc.): "Standard Preset" for KHRONOS_validation, reproed with or without "Fine Grained Locking"

juan-lunarg:
Applications can verify that queryPool can be destroyed by checking that vkGetQueryPoolResults() without the VK_QUERY_RESULT_PARTIAL_BIT flag returns VK_SUCCESS for all queries that are used in command buffers submitted for execution.

closes #4704

### DIFF
--- a/layers/cmd_buffer_state.cpp
+++ b/layers/cmd_buffer_state.cpp
@@ -1273,7 +1273,10 @@ void CMD_BUFFER_STATE::Retire(uint32_t perf_submit_pass, const std::function<boo
     for (const auto &query_state_pair : local_query_to_state_map) {
         if (query_state_pair.second == QUERYSTATE_ENDED && !is_query_updated_after(query_state_pair.first)) {
             auto query_pool_state = dev_data->Get<QUERY_POOL_STATE>(query_state_pair.first.pool);
-            query_pool_state->SetQueryState(query_state_pair.first.query, query_state_pair.first.perf_pass, QUERYSTATE_AVAILABLE);
+            if (query_pool_state) {
+                query_pool_state->SetQueryState(query_state_pair.first.query, query_state_pair.first.perf_pass,
+                                                QUERYSTATE_AVAILABLE);
+            }
         }
     }
 }
--- a/tests/positive/command.cpp
+++ b/tests/positive/command.cpp
@@ -260,6 +260,80 @@ TEST_F(VkPositiveLayerTest, BasicQuery) {
     vk::DestroyQueryPool(m_device->handle(), query_pool, NULL);
 }
 
+TEST_F(VkPositiveLayerTest, DestroyQueryPoolBasedOnQueryPoolResults) {
+    TEST_DESCRIPTION("Destroy a QueryPool based on vkGetQueryPoolResults");
+    ASSERT_NO_FATAL_FAILURE(InitFramework(m_errorMonitor));
+    VkCommandPoolCreateFlags pool_flags = VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT;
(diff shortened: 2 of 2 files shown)


juan-lunarg:
> @juan-lunarg - if you want to take a crack at this one it looks like it should be fairly straightforward. There's missing nullptr checks on handle lookups in CMD_BUFFER_STATE::Retire() (and ::Submit).

@jeremy-lunarg I reproduced a test case but didn't see a reason to add a `nullptr` check in `Submit` based on the issue described. Let me know what you think.

spencer-lunarg (approved):

Issue #4775, opened 2022-11-09
vkCmdPushDescriptorSetKHR validation issues?


fhoenig:
**Describe the Issue**
I call vkCmdPushDescriptorSetKHR for set 0, then set 2 then do a draw call.
After the draw call, the third call to vkCmdPushDescriptorSetKHR the validation layer crashes on "image_info was nullptr". 
Digging into the callstack I can clearly see that I pass the right arguments. However deeper into the validation layer in the call stack, I see some structs that changed the type to Image. See screenshot.

**Valid Usage ID**
The first draw call spits out this following error. Even though set 0 was bound via a successful `vkCmdPushDescriptorSetKHR` call.
```
"[Validation] Code 0 : Validation Error: [ UNASSIGNED-CoreValidation-DrawState-DescriptorSetNotBound ] Object 0: handle = 0x287dcf800c0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xcde11083 | vkCmdDraw(): VkPipeline 0xec25c90000000093[] uses set #0 but that set is not bound."
```

**Environment:**
 - OS: Windows 10
 - GPU: Quadro RTX 6000
 - SDK or header version if building from repo: *VulkanSDK-1.3.231.1*
 - Options enabled (synchronization, best practices, etc.):
```
Vulkan Development Status:
- Layers override: "Validation" configuration
- VULKAN_SDK environment variable: C:\Vulkan
- Vulkan Loader version: 1.3.201
- User-Defined Layers locations:
    - VK_LAYER_PATH variable: None
    - Per-configuration paths: None
    - VK_ADD_LAYER_PATH variable: None
- `vk_layer_settings.txt` uses the default platform path:
    C:\Users\Florian\AppData\Local\LunarG\vkconfig\override
- Available Layers:
    - VK_LAYER_NV_optimus
    - VK_LAYER_RENDERDOC_Capture
    - VK_LAYER_NV_nsight
    - VK_LAYER_VALVE_steam_overlay
    - VK_LAYER_VALVE_steam_fossilize
    - VK_LAYER_EOS_Overlay
    - VK_LAYER_EOS_Overlay
    - VK_LAYER_OBS_HOOK
    - VK_LAYER_OBS_HOOK
    - VK_LAYER_LUNARG_api_dump
    - VK_LAYER_LUNARG_gfxreconstruct
    - VK_LAYER_KHRONOS_synchronization2
[...]

PR #4781, opened 2022-11-10
Descriptor Set Tracking code cleanup

spencer-lunarg:
While looking into https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4775 I noticed a lot of small code cleanup that would be nice. Each commit breaks down the changes step by step


mark-lunarg (approved):
Very nice cleanup!

Issue #3363, opened 2021-09-30
Unexpected UNASSIGNED-CoreValidation-Shader-PointSizeMissing when emitting no points


rg3igalia:
**Describe the Issue**
In mesh shaders with the NVIDIA extension, the number of primitives that a given work group emits is controlled by the gl_PrimitiveCountNV variable. It's legal to emit no primitives at all by assigning 0 to it. When assigning 0, the shader should be allowed to not write anything to any output array, but doing so triggers UNASSIGNED-CoreValidation-Shader-PointSizeMissing if  the output primitive is points. Maybe the VUID could be changed to a warning for mesh shaders?

Note: for the future multi-vendor extension this will be even more important, as the mesh shader code is not **allowed** to write to any output array position beyond the limit specified by the dynamic count.

Example triggering the issue:
```
#version 450 
#extension GL_NV_mesh_shader : enable

layout (local_size_x=1) in; 
layout (points) out;
layout (max_vertices=1, max_primitives=1) out;

void main () {
    gl_PrimitiveCountNV = 0u; 
}
```

**Valid Usage ID**
UNASSIGNED-CoreValidation-Shader-PointSizeMissing

**Environment:**
 - OS: Linux
 - GPU: RTX 2060
 - SDK or header version if building from repo: cd72d751a1c9602b61132001e1ac147157187a8d
 - Options enabled (synchronization, best practices, etc.):

**Additional context**

spencer-lunarg:
raised [internal issue](https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5502), I would want a spec clarification for the `PointSize` builtin

spencer-lunarg:
@rg3igalia the Vulkan 1.3.235 spec now says

> VUID-StandaloneSpirv-MeshEXT-07728
>
> In mesh shaders using the MeshEXT or MeshNV Execution Model and the OutputPoints Execution Mode, a PointSize decorated variable must be written to

So this shader is invalid as it does need to write to a `PointSize` if using `layout (points) out;`

rg3igalia:
Thanks for the update @spencer-lunarg.  The VK_NV_mesh_shader test can be fixed like you suggest, but the new VUID affects both MeshNV and MeshEXT. In the case of MeshEXT we can't do that. For example, see the equivalent `dEQP-VK.mesh_shader.ext.misc.no_points` test:

[...]

Issue #4657, opened 2022-10-13
erroneous UNASSIGNED-CoreValidation-Shader-PointSizeMissing


zmike:
**Describe the Issue**
When running `KHR-GL46.shader_image_load_store.multiple-uniforms` with zink, I get this error from a geometry shader . The shader never emits any vertices, however, so it's pointless to be throwing an error.

**Valid Usage ID**
`UNASSIGNED-CoreValidation-Shader-PointSizeMissing(ERROR / SPEC): msgNum: -211210120 - Validation Error: [ UNASSIGNED-CoreValidation-Shader-PointSizeMissing ] Object 0: handle = 0xcc25c90000000082, type = VK_OBJECT_TYPE_SHADER_MODULE; | MessageID = 0xf3693078 | Pipeline topology is set to POINT_LIST, but PointSize is not written to in the shader corresponding to VK_SHADER_STAGE_GEOMETRY_BIT.`

**Environment:**
 - OS: Linux
 - GPU: N/A
 - SDK or header version if building from repo: e5afd16e31388ade3585a44af19388f4a7c18d10

**Additional context**
I think validation for this should verify that the GS emits vertices before triggering the error.

ziga-lunarg:
I believe the same applies for `dEQP-VK.geometry.emit.points_emit_0_end_0`

spencer-lunarg:
being `unassigned` this was probably added from early days. Can look "why" it was added, maybe get a proper VUID added to spec, and most important add the "are we emitting vertices" check

spencer-lunarg:
@zmike so as discussed in https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5502#note_385816 it seems that for this (assuming you can't just disable `shaderTessellationAndGeometryPointSize`) this is still an invalid usage and either need

1. convincing to relax the VUID more
2. fix the CTS test to write something to pointsize
3. having Zink inject the `PointSize` write

spencer-lunarg:
closing as stated above that the CTS is now invalid with regards to the new VUIDs

PR #3374, opened 2021-10-01
Fix issue 3363, validating mesh shader points output

ziga-lunarg:
Closes #3363

PR #4782, opened 2022-11-10
layers: Fix PointSize logic

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/3363 and https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4657 and removed more unassigned VUIDs!

There are [approved VUIDs](https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5502) that will be added probably in the next header, will go back and label when they are up streamed.

adds 
- VUID-VkGraphicsPipelineCreateInfo-Vertex-07722
-  VUID-VkGraphicsPipelineCreateInfo-TessellationEvaluation-07723
-  VUID-VkGraphicsPipelineCreateInfo-TessellationEvaluation-07724
-  VUID-VkGraphicsPipelineCreateInfo-Geometry-07725
-  VUID-VkGraphicsPipelineCreateInfo-Geometry-07726


layers/pipeline_state.h
```diff
@@ -188,6 +188,14 @@ class PIPELINE_STATE : public BASE_NODE {
     // Additional metadata needed by pipeline_state initialization and validation
     using StageStateVec = std::vector<PipelineStageState>;
     const StageStateVec stage_state;
+    bool HasShaderStage(VkShaderStageFlagBits stage_flag) const {
+        for (auto &state : stage_state) {
```
ziga-lunarg, line 192: Could be `const`

ziga-lunarg:
Is there a check if any points are actually emitted?

spencer-lunarg:
> Is there a check if any points are actually emitted?

Good questions, I realize you have https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/3374 already.

For Geometry shaders, I raised the question if `OpEmitVertex` is not called

For Mesh Shaders, if you have

```glsl
layout (points) out;
void main () {
    gl_PrimitiveCountNV = 0u; 
}
```

A valid tool like glslang will not produce a `BuiltIn PointSize` so it will not hit any validation (in `spirv-val`)

tests/positive/shaderval.cpp
```diff
@@ -3352,4 +3352,48 @@ TEST_F(VkPositiveLayerTest, UnnormalizedCoordinatesNotSampled) {
 
     m_commandBuffer->EndRenderPass();
     m_commandBuffer->end();
+}
+
+TEST_F(VkPositiveLayerTest, MeshShaderNoPointsEmitted) {
+    TEST_DESCRIPTION("Test mesh shader with not emiiting any points.");
```
arno-lunarg, line 3358: Nit: spelling

layers/shader_validation.cpp
```diff
@@ -2126,19 +2122,60 @@ bool CoreChecks::ValidatePointListShaderState(const PIPELINE_STATE *pipeline, co
...
+                    break;
+                } else if (mode == spv::ExecutionModePointMode) {
+                    point_mode = true;  // for tessellation
+                    break;
+                }
+            }
+        }
+    }
+
+    if (stage == VK_SHADER_STAGE_GEOMETRY_BIT && output_points) {
```
arno-lunarg, line 2146: Just a candid question: why mesh shaders are not among those various checks in this function?

arno-lunarg: Oh reading you comments, it seems that you actually need some input on what to do with mesh shaders

spencer-lunarg: Will update the comment in the code but basically

Vertex - Need to read input topology in pipeline
Geo/Tess - Need to know the feature bit is on
Mesh - Can detect if written or not all inside the shader module itself

So I am adding the mesh shader pointsize check in `spirv-val` because it can be detected offline

tests/positive/shaderval.cpp
juan-lunarg: Nit: Explain why we don't emit a VUID if a mesh shader doesn't emit points. I know this is mentioned in `ValidatePointSizeShaderState` but it's non-obvious if you are reading this test standalone.

tests/positive/shaderval.cpp
```diff
@@ -3352,4 +3352,48 @@ TEST_F(VkPositiveLayerTest, UnnormalizedCoordinatesNotSampled) {
...
+
+    AddRequiredExtensions(VK_NV_MESH_SHADER_EXTENSION_NAME);
```
juan-lunarg, line 3360: I'm assuming we aren't using `VK_EXT_MESH_SHADER_EXTENSION_NAME` for simplicity?

juan-lunarg (approved):
:+1:

juan-lunarg (approved):

PR #4783, opened 2022-11-10
layers: Fix null DSLs

ncesario-lunarg:
Fix a crash in
dEQP-VK.pipeline.pipeline_library.graphics_library.misc.other .null_descriptor_set_in_monolithic_pipeline
when independent sets/null DSLs are used in.


Tony-LunarG (approved):
LGTM, fixes the crash

PR #4784, opened 2022-11-10
build: Update to header 1.3.234

mikes-lunarg:
- Update known-good
- Generate source


ziga-lunarg (approved):

scripts/known_good.json
```diff
@@ -17,7 +17,7 @@
             "sub_dir": "Vulkan-Headers",
             "build_dir": "Vulkan-Headers/build",
             "install_dir": "Vulkan-Headers/build/install",
-            "commit": "ef96d229cb3d9d58b71c832aafd79f9dda306375"
+            "commit": "v1.3.234"
```
juan-lunarg, line 20: Stable tags set my heart at ease :+1:

juan-lunarg (approved):
LGTM

build-android/known_good.json
```diff
@@ -16,13 +16,13 @@
...
-            "commit": "v1.3.233"
+            "commit": "v1.3.234"
         },
         {
             "name": "Vulkan-Tools",
             "url": "https://github.com/KhronosGroup/Vulkan-Tools.git",
             "sub_dir": "Vulkan-Tools",
-            "commit": "v1.3.233"
+            "commit": "v1.3.234"
         },
```
juan-lunarg, line 26: Why does Android need Vulkan-Tools? `scripts/known_good.json` doesn't list Vulkan-Tools as a dependency.

mikes-lunarg: I don't remember. The only thing I see is that the validation layer tests include vulkan_wrapper.h. I wonder how hard it would be to remove that dependency...

PR #4786, opened 2022-11-10
build: Remove Android dependency on Vulkan-Tools

mikes-lunarg:
@juan-lunarg asked why the Android build needs Vulkan-Tools and I couldn't find a good reason!

### DIFF
--- a/build-android/jni/Android.mk
+++ b/build-android/jni/Android.mk
@@ -131,7 +131,6 @@ LOCAL_C_INCLUDES += $(VULKAN_INCLUDE) \
                     $(LOCAL_PATH)/$(SRC_DIR)/layers/generated \
                     $(LOCAL_PATH)/$(SRC_DIR)/layers \
                     $(LOCAL_PATH)/$(SRC_DIR)/libs \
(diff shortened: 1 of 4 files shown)


mark-lunarg:
Prolly used to be for running the android tests with devsim maybe?

mikes-lunarg:
> Prolly used to be for running the android tests with devsim maybe?

I thought devsim lived in the *other* VulkanTools. Best I can tell is that the Android VLTs used to use the vulkan_wrapper from KhronosGroup/Vulkan-Tools (https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/b57f0901b72063c16ebf044c809c31955a90193a) but the only thing left was the header include.

juan-lunarg (commented):
Change LGTM, but since I don't actively work on Android I defer to @spencer-lunarg

spencer-lunarg:
My guess from looking when added https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/4041d8f843d4764144e02390ec9d57079a998108 is that it was just there to match with the other `known_good.json` at the time, which used the MocICD/devsim

spencer-lunarg (approved):

PR #4788, opened 2022-11-11
tests: Move AHB tests to own file

spencer-lunarg:
Simple moving of the AndroidHardwareBuffer tests to be in their own file


arno-lunarg (approved):

tests/CMakeLists.txt
```diff
@@ -82,6 +82,7 @@ set(COMMON_CPP
     vklayertests_viewport_inheritance.cpp
     vklayertests_wsi.cpp
     vklayertests_graphics_library.cpp
+    vklayertests_android_hardware_buffer.cpp
```
jeremyg-lunarg, line 85: Should we just not add this file for non-android? I'm guessing that doing it with this way will make it easier to eventually switch to using cmake for android.

jeremyg-lunarg (approved):
Splitting giant test files into smaller files ? :100: yes
Splitting platform specific tests into their own file? :100: * :100: yes

tests/CMakeLists.txt
juan-lunarg: I personally prefer the approach @spencer-lunarg uses with #ifdef since it reduces complexity for the build system(s) and results in the file being viewable in Visual-Studio/XCode projects.

PR #4791, opened 2022-11-11
build-gn: Set++17

juan-lunarg:


juan-lunarg:
Latest version of PR pass gn build

BUILD.gn
```diff
@@ -15,6 +15,14 @@
 
 import("//build_overrides/vulkan_validation_layers.gni")
 
+config("language") {
+  if (is_win) {
+    cflags_cc = [ "/std=c++17" ]
+  } else {
+    cflags_cc = [ "-std:c++17" ]
+  } 
+}
```
juan-lunarg, line 24: Note the assignment instead of +=

PR #4792, opened 2022-11-12
tests: Move ErrorMonitor class into it's own files

juan-lunarg:


tests/vkerrormonitor.h
```diff
@@ -43,14 +43,6 @@ class ErrorMonitor {
...
     void SetDesiredFailureMsg(const VkFlags msgFlags, const char *const msgString);
 
-    // ErrorMonitor will look for an error message containing the specified string(s)
-    template <typename Iter>
-    void SetDesiredFailureMsg(const VkFlags msgFlags, Iter iter, const Iter end) {
-        for (; iter != end; ++iter) {
-            SetDesiredFailureMsg(msgFlags, *iter);
-        }
-    }
-
```
juan-lunarg: Removed this templated member function since it was only used in 1 location and better served by using `std::for_each`

tests/vkerrormonitor.cpp
```diff
@@ -113,14 +113,10 @@ VkBool32 ErrorMonitor::CheckForDesiredMsg(const char *const msgString) {
     return result;
 }
 
-std::vector<std::string> ErrorMonitor::GetOtherFailureMsgs() const { return other_messages_; }
```
juan-lunarg: Removed this function for making a pointless copy of other_messages_

juan-lunarg: I do realize this function was most likely being inlined/optimized but still, it's a bit confusing to read given the simplicity.

spencer-lunarg (approved):
LGTM :100:

jeremyg-lunarg (approved):
LGTM. It always freaks me out that none of the test code is in BUILD.gn

Issue #4793, opened 2022-11-13
Test: VkLayerTest.InvalidMemoryRequirements failed out of CI environment


sl1pkn07:
**Describe the Issue**
ctest failed when try to run in local machine (out of CI specs)

~~~
379/1536 Testing: VkLayerTest.InvalidMemoryRequirements
379/1536 Test: VkLayerTest.InvalidMemoryRequirements
Command: "/tmp/makepkg/sl1-vulkan-git/src/Vulkan-ValidationLayers/build64/tests/vk_layer_validation_tests" "--gtest_filter=VkLayerTest.InvalidMemoryRequirements" "--gtest_also_run_disabled_tests"
Directory: /tmp/makepkg/sl1-vulkan-git/src/Vulkan-ValidationLayers/build64/tests
"VkLayerTest.InvalidMemoryRequirements" start time: Nov 13 02:11 CET
Output:
----------------------------------------------------------
[0;33mNote: Google Test filter = VkLayerTest.InvalidMemoryRequirements
[m[0;32m[==========] [mRunning 1 test from 1 test suite.
[0;32m[----------] [mGlobal test environment set-up.
[0;32m[----------] [m1 test from VkLayerTest
[0;32m[ RUN      ] [mVkLayerTest.InvalidMemoryRequirements
/usr/include/c++/12.2.0/bits/stl_vector.h:1123: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = VkDrmFormatModifierProperties2EXT; _Alloc = std::allocator<VkDrmFormatModifierProperties2EXT>; reference = VkDrmFormatModifierProperties2EXT&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.
<end of output>
Test time =   0.29 sec
----------------------------------------------------------
Test Failed.
"VkLayerTest.InvalidMemoryRequirements" end time: Nov 13 02:11 CET
"VkLayerTest.InvalidMemoryRequirements" time elapsed: 00:00:00
----------------------------------------------------------
~~~

**Valid Usage ID**
If applicable, please include the validation messages encountered leading up to the issue

**Environment:**
 - OS: Archlinux
 - GPU: GTX1070ti FE (525.53)
 - SDK or header version if building from repo: 781c919ddf59701ccdb5b2532a7108d38fc21758
 - Options enabled (synchronization, best practices, etc.):

**Additional context**

[...]

PR #4795, opened 2022-11-13
tests: Disable DRM Format Modifier disjoint check

spencer-lunarg:
Temp fix as pointed out in https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4793

### DIFF
--- a/tests/vklayertests_buffer_image_memory_sampler.cpp
+++ b/tests/vklayertests_buffer_image_memory_sampler.cpp
@@ -11605,29 +11605,8 @@ TEST_F(VkLayerTest, InvalidMemoryRequirements) {
 
         vk::DestroyImage(m_device->device(), image, nullptr);
 
-        if (drm_format_modifier) {
-            constexpr std::array<uint64_t, 1> dummy_modifiers = {0};
-            auto drm_format_info = LvlInitStruct<VkImageDrmFormatModifierListCreateInfoEXT>();
-            drm_format_info.drmFormatModifierCount = dummy_modifiers.size();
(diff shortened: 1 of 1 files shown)


tests/vklayertests_buffer_image_memory_sampler.cpp
```diff
@@ -11605,29 +11605,31 @@ TEST_F(VkLayerTest, InvalidMemoryRequirements) {
...
-            mem_req_info2.pNext = nullptr;
-            mem_req_info2.image = image;
-
-            m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkImageMemoryRequirementsInfo2-image-02279");
-            vkGetImageMemoryRequirements2Function(device(), &mem_req_info2, &mem_req2);
-            m_errorMonitor->VerifyFound();
-
-            vk::DestroyImage(m_device->device(), image, nullptr);
-        }
+        // TODO - Test needs to use a real drm format and not try using ALIAS_BIT
```
jeremyg-lunarg, line 11608: Do we have a preferred way of disabling code blocks (and if so is it written down)? Seems like we can use C++ comments //, C /* */, #if 0 or (DISABLED_SOMETHING_SOMETHING), or just delete the block and put it back later.    I slightly prefer /* */ or #if 0, but would prefer consistency more.

jeremyg-lunarg (approved):
LGTM

spencer-lunarg:
> There is another location in the code where we use dummy formats:

Yes, but the issue here is the nature is

1. Wanting a successful `VkImage`
2. Doing multi-planar stuff

juan-lunarg:
Thanks for the clarification :+1:

PR #4800, opened 2022-11-14
test: Increase gtest discovery timeout

charles-lunarg:
The short default timeout of 1 second caused CI runs to fail when the call to gtest_discover_tests would error. Setting it to 100 makes the timeout be one hundred seconds, giving ample time for the command to run.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -117,7 +117,7 @@ add_executable(vk_layer_validation_tests
 if (CMAKE_CROSSCOMPILING)
     gtest_add_tests(TARGET vk_layer_validation_tests)
 else()
-    gtest_discover_tests(vk_layer_validation_tests)
+    gtest_discover_tests(vk_layer_validation_tests DISCOVERY_TIMEOUT 100)
 endif()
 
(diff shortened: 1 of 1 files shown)


juan-lunarg (approved):

ncesario-lunarg (commented):
LGTM, but will wait for an approval from @juan-lunarg.

tests/CMakeLists.txt
```diff
@@ -117,7 +117,7 @@ add_executable(vk_layer_validation_tests
 if (CMAKE_CROSSCOMPILING)
     gtest_add_tests(TARGET vk_layer_validation_tests)
 else()
-    gtest_discover_tests(vk_layer_validation_tests)
+    gtest_discover_tests(vk_layer_validation_tests PROPERTIES DISCOVERY_TIMEOUT 100)
```
juan-lunarg, line 120:
```suggestion
    gtest_discover_tests(vk_layer_validation_tests DISCOVERY_TIMEOUT 100)
```

juan-lunarg (changes requested):

juan-lunarg (approved):

Issue #4780, opened 2022-11-10
uninitialised variable warning on fedora rawhide


airlied:
Building for rawhide/f38 I get this problem.

/builddir/build/BUILD/Vulkan-ValidationLayers-sdk-1.3.231.1/layers/best_practices_utils.cpp: In member function 'BestPractices::RecordCmdBeginRenderingCommon(VkCommandBuffer_T*)':
/builddir/build/BUILD/Vulkan-ValidationLayers-sdk-1.3.231.1/layers/best_practices_utils.cpp:1972:57: error: 'MEM <unsigned int> [(struct optional *)&load_op]' may be used uninitialized [-Werror=maybe-uninitialized]
 1972 |             if (*load_op == VK_ATTACHMENT_LOAD_OP_CLEAR || *load_op == VK_ATTACHMENT_LOAD_OP_DONT_CARE) {
      |                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/builddir/build/BUILD/Vulkan-ValidationLayers-sdk-1.3.231.1/layers/best_practices_utils.cpp:1925:50: note: 'MEM <unsigned int> [(struct optional *)&load_op]' was declared here
 1925 |         layer_data::optional<VkAttachmentLoadOp> load_op;
      |                                                  ^~~~~~~

*Environment:**
 - OS: Fedora rawhide

spencer-lunarg:
The fix is trivial, but want to prevent in future, what compiler version are you using to build?

airlied:
gcc version 12.2.1 20221103 (Red Hat 12.2.1-3) (GCC)

ncesario-lunarg:
@airlied I'm on Fedora 37 (I don't know how to access 38) and have GCC version `gcc (GCC) 12.2.1 20220819 (Red Hat 12.2.1-2)`, but cannot reproduce this.

What version of Vulkan-ValidationLayers are you building? That line number doesn't seem to match up with where that line of currently exists: https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/6cd52525d3014d2b8f3abaa73c64daceba1a5eb9/layers/best_practices_utils.cpp#L2086

I'm not also not seeing what is potentially uninitialized on that line.

airlied:
[...]

PR #4801, opened 2022-11-14
CMake: BUILD_WERROR and CMakePresets.json

juan-lunarg:
Have BUILD_WERROR=OFF by default, add CMakePresets.json to facilitate easier workflow for developers.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -309,10 +309,19 @@ on/off options currently supported by this repository:
 | BUILD_LAYER_SUPPORT_FILES | All | `OFF` | Controls whether or not layer support files are installed. |
 | BUILD_TESTS | All | `OFF` | Controls whether or not the validation layer tests are built. |
 | INSTALL_TESTS | All | `OFF` | Controls whether or not the validation layer tests are installed. This option is only available when `BUILD_TESTS` is `ON` |
-| BUILD_WERROR | All | `ON` | Controls whether or not to treat compiler warnings as errors. |
+| BUILD_WERROR | All | `OFF` | Controls whether or not to treat compiler warnings as errors. |
 | BUILD_WSI_XCB_SUPPORT | Linux | `ON` | Build the components with XCB support. |
 | BUILD_WSI_XLIB_SUPPORT | Linux | `ON` | Build the components with Xlib support. |
 | BUILD_WSI_WAYLAND_SUPPORT | Linux | `ON` | Build the components with Wayland support. |
+
+### CMakePresets.json (3.21+)
+
+CMakePresets.json can save developer time by specifying common build flags.
+
+```bash
+# Enables tests, enable werror, etc.
+cmake -S . -B build/ --preset dev
+```
 
 ### CCACHE
 
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,21 @@
+{
+  "version": 3,
+  "cmakeMinimumRequired": {
+    "major": 3,
+    "minor": 21,
+    "patch": 0
+  },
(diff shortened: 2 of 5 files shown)


BUILD.md
```diff
@@ -309,11 +309,20 @@ on/off options currently supported by this repository:
...
 | BUILD_WSI_WAYLAND_SUPPORT | Linux | `ON` | Build the components with Wayland support. |
 
+### CMakePresets.json (3.21+)
+
+CMakePresets.json can save developer time by specifying common build flags.
+
+```bash
+# Enables tests, enable werror, etc.
+cmake -S . -B build/ --preset dev
+```
```
juan-lunarg, line 324: This is much nicer than

```
cmake -S . -B build -D BUILD_TESTS=ON -D BUILD_WERROR=ON -D UPDATE_DEPS=ON -D CMAKE_BUILD_TYPE=Debug
```

CMakePresets.json
```diff
@@ -0,0 +1,21 @@
+{
```
spencer-lunarg, line 1: is it normal to have a file like this in root directory or in the `cmake` folder?

spencer-lunarg (approved):
Assuming `scripts/common_ci.py` is being ran correclty, all for this!

I think this should close https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4780 as well then?

CMakePresets.json
arno-lunarg: I think you have to: https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html

CMakePresets.json
```diff
@@ -0,0 +1,21 @@
...
+  "version": 3,
+  "cmakeMinimumRequired": {
+    "major": 3,
+    "minor": 21,
```
arno-lunarg, line 5: In CMakeLists.txt, the minimum version is 3.10.2, is there a clash?

CMakePresets.json
juan-lunarg: Yes it has to be in this root directory.

CMakePresets.json
juan-lunarg: Presets are a relatively new feature to CMake. In order to use it your CMake version must be at least `3.21+`. This is intentional, assuming developers are using new version of CMake.

ncesario-lunarg (approved):
One question, but LGTM!

scripts/common_ci.py
```diff
@@ -79,6 +79,8 @@ def BuildVVL(args, build_tests=False):
     utils.make_dirs(VVL_BUILD_DIR)
     print("Run CMake for Validation Layers")
     cmake_cmd = f'cmake -DUPDATE_DEPS=ON -DUPDATE_DEPS_SKIP_EXISTING_INSTALL=ON -DCMAKE_BUILD_TYPE={args.configuration.capitalize()} {args.cmake} ..'
+    # By default BUILD_WERROR is OFF, CI should always enable it.
+    cmake_cmd = cmake_cmd + ' -DBUILD_WERROR=ON'
```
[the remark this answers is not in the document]
juan-lunarg: I don't think so. The Android.mk file doesn't have a `BUILD_WERROR` option AFAIK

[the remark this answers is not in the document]
juan-lunarg: My thought when I made this was the audience willing to build VVL was much smaller for Android and wouldn't be as big of a deal. Since I don't imagine as many developers wanting to try and package up the Android build.

jeremyg-lunarg (approved):
LGTM

PR #4803, opened 2022-11-15
cmake: Layers should be MODULEs

### LINKED ISSUE #2550: Question for windows build: install folder and symbols export
**Describe the Issue**

I don't know if it's the good place to ask. Maybe I miss something, but:
- why `VkLayer_khronos_validation` is built as `SHARED` and not `MODULE`, and why its symbols are exported? I thought that it was not supposed to be linked, but dynamically loaded by `vulkan-loader`. AFAIR, import lib is not even provided in Vulkan SDK. Moreover `-fvisibility=hidden` is passed to GNU compilers: https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/f9eec56cc82f4b3c268caac39be736105a209282/CMakeLists.txt#L124
- why `VkLayer_khronos_validation.dll` and its jsonfile are not installed in bin folder? It's usually where runtime (dll/exe) are installed on Windows (and vulkan sdk of LunarG puts all its validation layers -including this one- in bin folder).

juan-lunarg:
Changes library type from SHARED -> MODULE

MODULE's unlike SHARED libraries are intended to primarily be loaded dynamically at runtime using dlopen-like functionality.

This is the design intent of the validation layers.

closes #2550


juan-lunarg:
+ @lunarpapillo for feedback on the Windows SDK

juan-lunarg:
From the WIN32 CI:

```
-- Installing: D:/a/Vulkan-ValidationLayers/Vulkan-ValidationLayers/build/./install/release/lib/VkLayer_khronos_validation.json
-- Installing: D:/a/Vulkan-ValidationLayers/Vulkan-ValidationLayers/build/./install/release/lib/VkLayer_khronos_validation.dll
```

Note how the lib is no longer being installed.

juan-lunarg:
NOTE: The other change this causes is for MacOS builds. Instead of a `.dylib` it's now a `.so`

juan-lunarg:
@richard-lunarg for MacOS builds

richard-lunarg:
Whoa... so you are changing the macOS layers to be .so instead of .dylib? This is bound to break a lot of downstream stuff...

richard-lunarg:
Not so much functionality wise, I see the difference, and were we starting from scratch, the .so on macOS makes sense. I think a great deal of tooling and scripts however are looking for .dylib extensions.

juan-lunarg:
closing this PR, didn't intend the MacOS build side effect

SpaceIm:
@juan-lunarg some workaround for macOS in https://gitlab.kitware.com/cmake/cmake/-/issues/21189. Basically changing `SHARED_MODULE_SUFFIX` property of `VkLayer_khronos_validation` target should be sufficient.

juan-lunarg:
> @juan-lunarg some workaround for macOS in https://gitlab.kitware.com/cmake/cmake/-/issues/21189. Basically changing `CMAKE_SHARED_MODULE_SUFFIX` to `.dylib` if `APPLE` should be sufficient.

Added that info to the issue. Thanks for pointing this out :+1:

PR #4806, opened 2022-11-15
cmake: Install during CI builds

juan-lunarg:
Mainly useful for logging/verification.


juan-lunarg:
This can also catch CMake installation bugs, in case they happen again.

spencer-lunarg (approved):

PR #4813, opened 2022-11-16
cmake: Avoid duplicate VkLayer_khronos_validation.json

juan-lunarg:
Now the build directory will only contain 1 json file named VkLayer_khronos_validation.json

This caused an issue for CI projects that searched the build dir for the json file.

image

### DIFF
--- a/layers/CMakeLists.txt
+++ b/layers/CMakeLists.txt
@@ -312,7 +312,8 @@ target_link_libraries(VkLayer_khronos_validation PRIVATE ${SPIRV_TOOLS_TARGET} S
 # This code generates the appropriate json for both local testing and the installation.
 # NOTE: For WIN32 the JSON and dll MUST be placed in the same location, due to Win32 using a relative path for installation.
 set(INPUT_FILE "${CMAKE_CURRENT_SOURCE_DIR}/json/VkLayer_khronos_validation.json.in")
-set(OUTPUT_FILE "${VVL_BINARY_DIR}/json/VkLayer_khronos_validation.json")
+set(INTERMEDIATE_FILE "${CMAKE_CURRENT_BINARY_DIR}/json/validation.json")
+set(OUTPUT_FILE_FINAL_NAME "VkLayer_khronos_validation.json")
 set(LAYER_INSTALL_DIR ${CMAKE_INSTALL_LIBDIR})
 
 if (WIN32)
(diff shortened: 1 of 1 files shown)


layers/CMakeLists.txt
```diff
@@ -312,7 +312,8 @@ target_link_libraries(VkLayer_khronos_validation PRIVATE ${SPIRV_TOOLS_TARGET} S
 # This code generates the appropriate json for both local testing and the installation.
 # NOTE: For WIN32 the JSON and dll MUST be placed in the same location, due to Win32 using a relative path for installation.
 set(INPUT_FILE "${CMAKE_CURRENT_SOURCE_DIR}/json/VkLayer_khronos_validation.json.in")
-set(OUTPUT_FILE "${VVL_BINARY_DIR}/json/VkLayer_khronos_validation.json")
+set(OUTPUT_FILE "${CMAKE_CURRENT_BINARY_DIR}/json/validation.json")
```
lunarpapillo, line 315: A comment here explaining why a different name was chosen for the intermediary file would keep someone from inadvertently changing it back in the future.

It might be helpful to name it `INTERMEDIATE_FILE` or something like that, so it's clear from usage that this isn't the final output file.

layers/CMakeLists.txt
```diff
@@ -327,19 +328,19 @@ configure_file(${INPUT_FILE} ${OUTPUT_FILE} @ONLY)
 
 # To support both multi/single configuration generators just copy the json to the correct directory
 add_custom_command(TARGET VkLayer_khronos_validation POST_BUILD
-    COMMAND ${CMAKE_COMMAND} -E copy ${OUTPUT_FILE} $<TARGET_FILE_DIR:VkLayer_khronos_validation>
+    COMMAND ${CMAKE_COMMAND} -E copy_if_different ${OUTPUT_FILE} $<TARGET_FILE_DIR:VkLayer_khronos_validation>/${OUTPUT_FILE_FINAL_NAME}
 )
 
 # For UNIX-based systems, `library_path` should not contain a relative path (indicated by "./") before installing to system directories
 # This json isn't used for regular local development, it's used installation.
```
lunarpapillo, line 335: `used installation` => `used for installation`

lunarpapillo (approved):
A clever solution.  I had thought of some others, but this is probably cleaner.

juan-lunarg:
Updated based on feedback.

layers/CMakeLists.txt
juan-lunarg: Changed the name to `INTERMEDIATE_FILE`.

Tony-LunarG (approved):
LGTM - Thanks!

PR #4816, opened 2022-11-17
layers: Move accessible ids to shader module state

spencer-lunarg:
It didn't seem like `PipelineStageState` needed to own the per-function access id variables, now store them inside a per-EntryPoint level, makes future multi-entry-point support nicer too

- move `accessible_ids`/`MarkAccessibleIds` from a per-`PIPELINE_STATE` to a per-`SHADER_MODULE_STATE` in the `OpEntryPoint`
- Redo `EntryPoint` to only be what is needed
- Remove need for `std::reference_wrapper`


jeremyg-lunarg (commented):
LGTM but @ncesario-lunarg and / or @Tony-LunarG should also review this

ncesario-lunarg (approved):
Great cleanup!

PR #4819, opened 2022-11-17
cpp: Use std::make_unique (C++14)

juan-lunarg:


jeremyg-lunarg (approved):

spencer-lunarg:
are we still using`layerdata::make_unique` anywhere or can it be removed from `vk_layer_data.h`?

juan-lunarg:
> are we still using`layerdata::make_unique` anywhere or can it be removed from `vk_layer_data.h`?

I removed it entirely

PR #4822, opened 2022-11-17
layers: Fix fuchsia buffer import check

jbauman42:
The check was attempting to add a pointer to an integer. Instead we want to check if the pointer is present, and if so add 1 to the integer.


spencer-lunarg (approved):
LGTM, the CI failure is unrelated (issue with recent mesh shading PR it seems)

juan-lunarg (approved):

juan-lunarg:
> LGTM, the CI failure is unrelated (issue with recent mesh shading PR it seems)

Agreed, CI failure is unrelated. I'll merge the change in.

PR #4823, opened 2022-11-17
cmake: Treat xxhash/vma as SYSTEM includes

juan-lunarg:
This is consistent with our approach for all other 3rd-party code.

With this change -Wpedantic can be added to the build.


CMakeLists.txt
```diff
@@ -236,6 +237,8 @@ if(BUILD_LAYERS OR BUILD_TESTS)
     endif()
 endif()
 
+include_directories(SYSTEM ${CMAKE_CURRENT_SOURCE_DIR}/layers/third_party)
```
juan-lunarg, line 240: `SYSTEM` includes mean the header files won't be analyzed for warnings. Which is how all of our other dependencies are compiled.

IE it will use `-isystem` instead of `-I`

juan-lunarg:
To be clear this wasn't just for `-Wpedantic`, this same problem occurs with `-Wimplicit-fallthrough`

CMakeLists.txt
Tony-LunarG: Cool - Do you think we should remove all the changes we added for warnings?

juan-lunarg: Currently `SYSTEM` only works for GCC/CLANG, MSVC until recently didn't have system include functionality. It's been added in newer VS/CMake but we can't rely on it for now.

juan-lunarg: CMake 3.24+ is when `SYSTEM` is usable for us.

The [Visual Studio Generators](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#visual-studio-generators) now support SYSTEM headers when using VS 2019 Update 11 or later.

https://learn.microsoft.com/en-us/cpp/build/reference/external-external-headers-diagnostics?view=msvc-170

juan-lunarg:
Closing due to -Wpedantic being buggy on my local system... I can't reproduce the same errors CI is despite using a newer GCC version.

PR #4824, opened 2022-11-18
layers: Ensure shader stage validation for GPL 

### LINKED ISSUE #4811: VK_EXT_graphics_pipeline_library - not checking stage masks for descriptors leading to DEVICE_LOST
When using VK_EXT_graphics_pipeline_library in Dota a pre-rasterization pipeline library is created with the following situation:

* The vertex shader fetches from a sampler in descriptor set 3, binding 17
* The descriptor set layout containing that descriptor has stage masks of PIXEL | COMPUTE

No validation error is given either when creating the pre-rasterization pipeline library OR when creating the final linked pipeline library.  This leads to a VK_ERROR_DEVICE_LOST on NVIDIA.

If I run my application without VK_EXT_graphics_pipeline_library (Dota 2), validation does give an error in this case as it should:

VK ERROR : VALIDATION - Message Id Number: 1165064310 | Message Id Name: VUID-VkGraphicsPipelineCreateInfo-layout-00756
	Validation Error: [ VUID-VkGraphicsPipelineCreateInfo-layout-00756 ] Object 0: handle = 0xb2ec7c0000001b92, name = grasstile.vfx_vs, type = VK_OBJECT_TYPE_SHADER_MODULE; Object 1: handle = 0xefa4b60000038a40, type = VK_OBJECT_TYPE_PIPELINE_LAYOUT; | MessageID = 0x45717876 | Set 3 Binding 17 in shader uses descriptor slot but descriptor not accessible from stage VK_SHADER_STAGE_VERTEX_BIT The Vulkan spec states: layout must be consistent with all shaders specified in pStages (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkGraphicsPipelineCreateInfo-layout-00756)

The same or similar error should be given when creating the pre-rasterization pipeline library.

**Valid Usage IDs requested**
VUID-VkGraphicsPipelineCreateInfo-layout-00756

**Additional Context**
Any additional information: GTX 2080 NVIDIA 526.47

ncesario-lunarg:
Fixes the IsGraphicsLibrary and a shader module identifier check that
prevented all shader stage validation from happening when GPL was in
use.

Closes #4811.

### DIFF
--- a/layers/pipeline_state.cpp
+++ b/layers/pipeline_state.cpp
@@ -258,10 +258,11 @@ static layer_data::unordered_set<uint32_t> GetFSOutputLocations(const PIPELINE_S
     return result;
 }
 
-static VkPrimitiveTopology GetTopologyAtRasterizer(const PIPELINE_STATE::StageStateVec &stage_states,
-                                                   const safe_VkPipelineInputAssemblyStateCreateInfo *assembly_state) {
-    VkPrimitiveTopology result = assembly_state ? assembly_state->topology : static_cast<VkPrimitiveTopology>(0);
-    for (const auto &stage : stage_states) {
+static VkPrimitiveTopology GetTopologyAtRasterizer(const PIPELINE_STATE &pipeline) {
+    auto result = (pipeline.vertex_input_state && pipeline.vertex_input_state->input_assembly_state)
+                      ? pipeline.vertex_input_state->input_assembly_state->topology
+                      : VK_PRIMITIVE_TOPOLOGY_MAX_ENUM;
+    for (const auto &stage : pipeline.stage_state) {
         if (!stage.entrypoint) {
             continue;
         }
@@ -536,7 +537,7 @@ PIPELINE_STATE::PIPELINE_STATE(const ValidationStateTracker *state_data, const V
       active_slots(GetActiveSlots(stage_state)),
       max_active_slot(GetMaxActiveSlot(active_slots)),
       active_shaders(GetActiveShaders(stage_state)),
-      topology_at_rasterizer(GetTopologyAtRasterizer(stage_state, create_info.graphics.pInputAssemblyState)),
+      topology_at_rasterizer(GetTopologyAtRasterizer(*this)),
       uses_shader_module_id(UsesShaderModuleId(stage_state)),
       descriptor_buffer_mode((create_info.graphics.flags & VK_PIPELINE_CREATE_DESCRIPTOR_BUFFER_BIT_EXT) != 0),
       uses_pipeline_robustness(UsesPipelineRobustness(PNext(), stage_state)),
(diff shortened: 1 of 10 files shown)


layers/pipeline_state.cpp
```diff
@@ -232,10 +232,12 @@ static layer_data::unordered_set<uint32_t> GetFSOutputLocations(const PIPELINE_S
     return result;
 }
 
-static VkPrimitiveTopology GetTopologyAtRasterizer(const PIPELINE_STATE::StageStateVec &stage_states,
-                                                   const safe_VkPipelineInputAssemblyStateCreateInfo *assembly_state) {
-    VkPrimitiveTopology result = assembly_state ? assembly_state->topology : static_cast<VkPrimitiveTopology>(0);
-    for (const auto &stage : stage_states) {
+static VkPrimitiveTopology GetTopologyAtRasterizer(const PIPELINE_STATE &pipeline) {
+    // Choose VK_PRIMITIVE_TOPOLOGY_POINT_LIST as an arbitrary default if input assembly state is missing
```
spencer-lunarg, line 236: why not `VK_PRIMITIVE_TOPOLOGY_MAX_ENUM`? 

Also trying to understand why not all `topology_at_rasterizer` are initialize with an explicit default value

spencer-lunarg (approved):

layers/pipeline_state.cpp
ncesario-lunarg: Good question! I was trying to keep the previous behavior and didn't think casting 0 made sense, but looking at the usage of `topology_at_rasterizer`, I think an invalid value like `VK_PRIMITIVE_TOPOLOGY_MAX_ENUM` makes a lot more sense

layers/shader_validation.cpp
```diff
@@ -2786,10 +2786,12 @@ bool CoreChecks::ValidatePipelineShaderStage(const PIPELINE_STATE *pipeline, con
     const auto *pStage = stage_state.create_info;
     const SHADER_MODULE_STATE &module_state = *stage_state.module_state.get();
     auto entrypoint_optional = stage_state.entrypoint;
+    const auto module_identifier = LvlFindInChain<VkPipelineShaderStageModuleIdentifierCreateInfoEXT>(stage_state.create_info);
 
     skip |= ValidateShaderModuleId(module_state, stage_state, pStage, pipeline->GetPipelineCreateFlags());
 
-    if (module_state.vk_shader_module() == VK_NULL_HANDLE) return skip;  // No real shader for further validation
+    if (module_identifier && module_state.vk_shader_module() == VK_NULL_HANDLE)
```
ncesario-lunarg, line 2793: @Tony-LunarG is this still ok for `VK_EXT_shader_module_identifier`?

Tony-LunarG (approved):
LGTM

PR #4828, opened 2022-11-18
Add VK_EXT_descriptor_buffer

archimedus:
Add support into validation layers for VK_EXT_descriptor_buffer and appropriate tests.
Mostlikely [defs: Update VK_EXT_descriptor_buffer in known_good.json](https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/5b3b80df983a6c353cf0c282f3e972579b3aea6a) should be removed

PR #4831, opened 2022-11-18
layers: Fix invalid memory access in UpdatePipelineState()

jeremyg-lunarg:
This method in CMD_BUFFER_STATE was dereferencing a vector before checking that the index in bounds.


jeremyg-lunarg:
This was part of https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4828

juan-lunarg (approved):

PR #4835, opened 2022-11-18
layers: Always use std::shared_mutex now that we're c++17

jeremyg-lunarg:


layers/vk_layer_utils.h
```diff
@@ -354,60 +354,10 @@ static inline int u_ffs(int val) {
 #endif
 
 #ifdef __cplusplus
-// clang sets _MSC_VER to 1800 and _MSC_FULL_VER to 180000000, but we only want to clean up after MSVC.
-#if defined(_MSC_FULL_VER) && !defined(__clang__)
-// Minimum Visual Studio 2015 Update 2, or libc++ with C++17
-// But, before Visual Studio 2017 version 15.7, __cplusplus is not set
-// correctly. See:
-//   https://docs.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-160
```
jeremyg-lunarg, line 362: I will miss this epic dunk on early MSVC lack of regard for standards.

ncesario-lunarg (approved):
(proverbial sigh of relief)

layers/vk_layer_utils.h
juan-lunarg: You can actually get MSVC to use `__cplusplus` like GCC/Clang. You have to use [/permissive-](https://learn.microsoft.com/en-us/cpp/build/reference/permissive-standards-conformance?view=msvc-170)

juan-lunarg (approved):

layers/vk_layer_utils.h
ncesario-lunarg: Seems like we should be using that. That requires VS >= 2017, though. We've talked about dropping 2015 support in the past; we should probably add this to our Windows builds as soon as that happens (or add it now behind a compiler version check).

juan-lunarg: Does VS 2015 support C++17?

ncesario-lunarg: > Does VS 2015 support C++17?

It does in the sense our CI running VS 2015 passes :)

I think https://learn.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-140 lists what's supported.

jeremyg-lunarg: I think moving to c++17 solves this specific problem, which was the worst one that I knew about.

jeremyg-lunarg:
CI failure is another `ERROR: [WS-CLEANUP] Cannot delete workspace: remote file operation failed`

PR #4837, opened 2022-11-19
cmake: Layers should be MODULE instead of SHARED

### LINKED ISSUE #4804: Layers should be MODULE instead of SHARED
Currently `VkLayer_khronos_validation` is a SHARED library type. Since layers are intended to be dynamically loaded at runtime it should be a MODULE library.

```
add_library(VkLayer_khronos_validation MODULE)
```

This mainly impacts windows builds, since by default installs a static lib is not even provided in Vulkan SDK.

```
-- Installing: D:/.../build/./install/lib/VkLayer_khronos_validation.json
-- Installing: D:/.../build/./install/lib/VkLayer_khronos_validation.lib
-- Installing: D:/.../build/./install/lib/VkLayer_khronos_validation.dll
```

However, changing the type to `MODULE` has the side affect of changing `MacOS` binaries to `.so` from `.dylib`. This can be mitigated, but must be taken into account.

Changing the `SUFFIX` property of `VkLayer_khronos_validation` target should be sufficient.

juan-lunarg:
closes #4804

### DIFF
--- a/layers/CMakeLists.txt
+++ b/layers/CMakeLists.txt
@@ -235,7 +235,7 @@ if (NOT BUILD_LAYERS)
     return()
 endif()
 
-add_library(VkLayer_khronos_validation SHARED)
+add_library(VkLayer_khronos_validation MODULE)
 
 target_sources(VkLayer_khronos_validation PRIVATE
     ${CHASSIS_LIBRARY_FILES}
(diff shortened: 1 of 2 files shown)


juan-lunarg:
Now on WIN32 only the dll is installed, and backcompat has been preserved for MacOS

```
# WIN32
-- Installing: D:/a/.../install/lib/VkLayer_khronos_validation.json
-- Installing: D:/a/.../install/lib/VkLayer_khronos_validation.dll

# MACOS
-- Installing: /Users/.../install/share/vulkan/explicit_layer.d/VkLayer_khronos_validation.json
-- Installing: /Users/.../install/lib/libVkLayer_khronos_validation.dylib
```

spencer-lunarg (approved):

PR #4839, opened 2022-11-19
layers: Fix image nullptr crash

ziga-lunarg:
This crash happened when `vkQueuePresentKHR ` returned `VK_SUBOPTIMAL_KHR`, not sure if I can write a test for this

### DIFF
--- a/layers/buffer_validation.cpp
+++ b/layers/buffer_validation.cpp
@@ -4773,6 +4773,9 @@ bool CoreChecks::ValidateCmdBufImageLayouts(const Location &loc, const CMD_BUFFE
     for (const auto &layout_map_entry : pCB->image_layout_map) {
         const auto *image_state = layout_map_entry.first;
         const auto &subres_map = layout_map_entry.second;
+        if (!subres_map) {
+            continue;
+        }
         const auto &layout_map = subres_map->GetLayoutMap();
         // Validate the initial_uses for each subresource referenced
         if (layout_map.empty()) continue;
@@ -4846,6 +4849,9 @@ void CoreChecks::UpdateCmdBufImageLayouts(CMD_BUFFER_STATE *pCB) {
     for (const auto &layout_map_entry : pCB->image_layout_map) {
         const auto *image_state = layout_map_entry.first;
         const auto &subres_map = layout_map_entry.second;
+        if (!subres_map) {
+            continue;
+        }
         auto guard = image_state->layout_range_map->WriteLock();
         sparse_container::splice(*image_state->layout_range_map, subres_map->GetLayoutMap(), GlobalLayoutUpdater());
     }
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -12832,9 +12832,11 @@ bool CoreChecks::ValidateFramebufferCreateInfo(const VkFramebufferCreateInfo *pC
                         }
 
                         // Verify that image memory is valid
-                        auto image_data = Get<IMAGE_STATE>(ivci.image);
-                        skip |= ValidateMemoryIsBoundToImage(image_data.get(), "vkCreateFramebuffer()",
-                                                             kVUID_Core_Bound_Resource_FreedMemoryAccess);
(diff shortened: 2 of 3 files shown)


layers/buffer_validation.cpp
```diff
@@ -4773,6 +4773,9 @@ bool CoreChecks::ValidateCmdBufImageLayouts(const Location &loc, const CMD_BUFFE
     for (const auto &layout_map_entry : pCB->image_layout_map) {
         const auto *image_state = layout_map_entry.first;
         const auto &subres_map = layout_map_entry.second;
+        if (!subres_map) {
```
spencer-lunarg, line 4776: I am not sure where `pCB->image_layout_map` is actually being populated, but not sure why the `ImageSubresourceLayoutMap` would be null

spencer-lunarg: So this check is not being done in the `ExecuteCommands` calls for secondary buffer?

tests/vklayertests_wsi.cpp
```diff
@@ -2450,3 +2450,125 @@ TEST_F(VkLayerTest, PhysicalDeviceSurfaceCapabilities) {
...
+    fbci.pAttachments = &image_view_handle;
+    fbci.width = 32;
+    fbci.height = 32;
+    fbci.layers = 1;
+    vk_testing::Framebuffer fb(*m_device, fbci);
+
+    VkShaderObj vs(this, bindStateVertShaderText, VK_SHADER_STAGE_VERTEX_BIT);
+    VkShaderObj fs(this, bindStateFragShaderText, VK_SHADER_STAGE_FRAGMENT_BIT);
+    const VkPipelineLayoutObj pipeline_layout(m_device);
+    VkPipelineColorBlendAttachmentState att_state = {}, att_state2 = {};
```
juan-lunarg, line 2544: att_state2 isn't being used

```suggestion
    VkPipelineColorBlendAttachmentState att_state = {};
```

tests/vklayertests_wsi.cpp
```diff
@@ -2450,3 +2450,125 @@ TEST_F(VkLayerTest, PhysicalDeviceSurfaceCapabilities) {
     m_errorMonitor->VerifyFound();
 }
 #endif  // VK_USE_PLATFORM_WIN32_KHR
+
+TEST_F(VkLayerTest, UseDestroyedSwapchain) {
```
juan-lunarg, line 2454: I don't see any usage of `m_errorMonitor`. If that's the case then it belongs in the `tests/positive` directory

ziga-lunarg: This is WIP, this test should fail, but it doesn't.

layers/core_validation.cpp
```diff
@@ -13051,8 +13053,8 @@ bool CoreChecks::ValidateFramebufferCreateInfo(const VkFramebufferCreateInfo *pC
                                 string_VkComponentSwizzle(ivci.components.b), string_VkComponentSwizzle(ivci.components.a));
                         }
                         if ((ivci.viewType == VK_IMAGE_VIEW_TYPE_2D) || (ivci.viewType == VK_IMAGE_VIEW_TYPE_2D)) {
-                            auto image_state = Get<IMAGE_STATE>(ivci.image);
-                            if (image_state->createInfo.imageType == VK_IMAGE_TYPE_3D) {
+                            const auto &image_state = Get<IMAGE_STATE>(ivci.image);
```
jeremyg-lunarg, line 13056: This shouldn't be an auto &. We need to copy to ensure a reference is held on the IMAGE_STATE, otherwise it could be destroyed in a different thread. (This would be invalid vulkan but we need to guard against it)

jeremyg-lunarg:
> This crash happened when `vkQueuePresentKHR ` returned `VK_SUBOPTIMAL_KHR`, not sure if I can write a test for this

It almost seems like we need a "Mock Swapchain" to test things like this. Not sure how hard that would be.

jeremyg-lunarg (changes requested):
-1, the shared_ptr & needs to be fixed.

PR #4842, opened 2022-11-21
manifest: improve warn robust oob setting dependency

christophe-lunarg:
After discussion with @Tony-LunarG, this is a proposed manifest improvement to better expose to the user the layer settings dependence.


Tony-LunarG (approved):
LGTM - thanks!

PR #4844, opened 2022-11-21
cmake: Fix incorrect install location

### LINKED ISSUE #4805: Fix library install paths
Currently `VkLayer_khronos_validation` is installed to the incorrect location. It's currently installed to `CMAKE_INSTALL_LIBDIR` instead of `CMAKE_INSTALL_BINDIR`

The change is easy to fix. The main problem is preserving back-compat.

See https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/2861 for more details.

juan-lunarg:
After looking into this for a while. CMake doesn't have a standard location for `MODULE` libraries. Because there is no standard for plugin type libraries.

As a result just match our SDK creation process. This doesn't change UNIX platforms, it only matches our WIN32 SDK installation.

closes #4805

Example install ouput

```
# WIN32
-- Installing: D:/a/.../install/bin/VkLayer_khronos_validation.json
-- Installing: D:/a/.../install/bin/VkLayer_khronos_validation.dll
```

### DIFF
--- a/layers/CMakeLists.txt
+++ b/layers/CMakeLists.txt
@@ -306,7 +306,7 @@ set(INTERMEDIATE_FILE "${CMAKE_CURRENT_BINARY_DIR}/json/validation.json")
 set(OUTPUT_FILE_FINAL_NAME "VkLayer_khronos_validation.json")
 set(LAYER_INSTALL_DIR ${CMAKE_INSTALL_LIBDIR})
 if (WIN32)
-    set(LAYER_INSTALL_DIR ${CMAKE_INSTALL_BINDIR})
+    set(LAYER_INSTALL_DIR ${CMAKE_INSTALL_BINDIR}) # WIN32/MINGW expect the dll in the `bin` dir, this matches our WIN32 SDK process
 endif()
 
 if (WIN32)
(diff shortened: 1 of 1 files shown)


juan-lunarg:
While the SDK scripts would need a bit of tweaking to account for the new behavior. I don't see the harm in this PR at least for SDK creation.

And for users who install VVL directly I'd imagine using a standard location would be welcomed.

I could use symbolic links for back-compat but that ultimately seems like it could complicate things even more.

Thoughts?

juan-lunarg:
Example install ouput

```
# WIN32
-- Installing: D:/a/.../install/bin/VkLayer_khronos_validation.json
-- Installing: D:/a/.../install/bin/VkLayer_khronos_validation.dll
```

```
# Apple
-- Installing: /Users/.../install/share/vulkan/explicit_layer.d/VkLayer_khronos_validation.json
-- Installing: /Users/.../bin/libVkLayer_khronos_validation.dylib
```

```
# Linux
-- Installing: /home/.../install/share/vulkan/explicit_layer.d/VkLayer_khronos_validation.json
-- Installing: /home/.../install/bin/libVkLayer_khronos_validation.so
```

spencer-lunarg (approved):

PR #4845, opened 2022-11-21
layers: Fix GetDeviceProcAddr querying instance functions

### LINKED ISSUE #4814: Bug in `ApiParentExtensionEnabled` for device level debug utils functions
**Describe the Issue**
ApiParentExtensionEnabled() is called during vkGetDeviceProcAddr to see if the relevant extensions are enabled. Debug utils is an *instance* level extension with numerous device level functions, like `vkSetDebugUtilsObjectNameEXT`. Since VK_EXT_debug_utils is not a device extension, `ApiParentExtensionEnabled` will never return true since it doesn't handle this corner case.

```c++

const layer_data::unordered_map<std::string, std::string> api_extension_map {
    {"vkCmdBeginDebugUtilsLabelEXT", "VK_EXT_debug_utils"},
    {"vkCmdEndDebugUtilsLabelEXT", "VK_EXT_debug_utils"},
    {"vkCmdInsertDebugUtilsLabelEXT", "VK_EXT_debug_utils"},
    {"vkQueueBeginDebugUtilsLabelEXT", "VK_EXT_debug_utils"},
    {"vkQueueEndDebugUtilsLabelEXT", "VK_EXT_debug_utils"},
    {"vkQueueInsertDebugUtilsLabelEXT", "VK_EXT_debug_utils"},
    {"vkSetDebugUtilsObjectNameEXT", "VK_EXT_debug_utils"},
    {"vkSetDebugUtilsObjectTagEXT", "VK_EXT_debug_utils"},
...
};
// Using the above code-generated map of APINames-to-parent extension names, this function will:
//   o  Determine if the API has an associated extension
//   o  If it does, determine if that extension name is present in the passed-in set of enabled_ext_names 
//   If the APIname has no parent extension, OR its parent extension name is IN the set, return TRUE, else FALSE
static inline bool ApiParentExtensionEnabled(const std::string api_name, const DeviceExtensions *device_extension_info) {
    auto has_ext = api_extension_map.find(api_name);
    // Is this API part of an extension or feature group?
    if (has_ext != api_extension_map.end()) {
        // Was the extension for this API enabled in the CreateDevice call?
        auto info = device_extension_info->get_info(has_ext->second.c_str());
        if ((!info.state) || (device_extension_info->*(info.state) != kEnabledByCreateinfo)) {
            return false;
        }
    }
    return true;
}
```

**Additional context**
This issue was discovered by baldurk when investigating the situation where `vkSetDebugUtilsObjectNameEXT` names were not being used in renderdoc while validation layers were active. The conversations about this are on the renderdoc discord server. Poke me if you want links to the conversations.

A recent loader change was to use `vkGetDeviceProcAddr` for debug utils functions, instead of `vkGetInstanceProcAddr`. That means there is a good change that these debug utils functions may break unless this is fixed. https://github.com/KhronosGroup/Vulkan-Loader/commit/d1db6c5bce9ed474ca124f0bffed9aa2468f3973

charles-lunarg:
Certain device level functions originate from instance level extensions such as vkSetDebugUtilsObjectName. The code to validate that functions queried with vkGetDeviceProcAddr did not properly handle this case, leading to nullptr being returned. This is hidden because the loader uses vkGetInstanceProcAddr to query debug utils functions, but layers such as renderdoc use vkGetDeviceProcAddr leading to calls to these functions never reaching the validation layers.

Fixes #4814


layers/generated/vk_dispatch_table_helper.h
```diff
@@ -1027,17 +1027,23 @@ const layer_data::unordered_map<std::string, std::string> api_extension_map {
...
     // Is this API part of an extension or feature group?
     if (has_ext != api_extension_map.end()) {
         // Was the extension for this API enabled in the CreateDevice call?
         auto info = device_extension_info->get_info(has_ext->second.c_str());
-        if ((!info.state) || (device_extension_info->*(info.state) != kEnabledByCreateinfo)) {
-            return false;
+        if (info.state) {
+            return device_extension_info->*(info.state) == kEnabledByCreateinfo;
         }
+        auto instance_extension_info = static_cast<const InstanceExtensions*>(device_extension_info);
```
mikes-lunarg, line 1041: You shouldn't have to cast to `InstanceExtensions`, the `DeviceExtensions` struct contains both the enabled device extensions and enabled instance extensions.

layers/generated/vk_dispatch_table_helper.h
```diff
@@ -1027,17 +1027,23 @@ const layer_data::unordered_map<std::string, std::string> api_extension_map {
...
+        auto inst_info = instance_extension_info->get_info(has_ext->second.c_str());
+        if (inst_info.state) {
+            return instance_extension_info->*(inst_info.state) == kEnabledByCreateinfo;
```
mikes-lunarg, line 1044: This is an improvement but it still doesn't match the table here: https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#vkGetDeviceProcAddr

I don't think we are looking at the whether the entry points are device-level dispatchable. The spec specifically excludes "physical-device-level commands" from being queried by `vkGetDeviceProcAddr`.

layers/generated/vk_dispatch_table_helper.h
charles-lunarg: While `DeviceExtensions` does effectively contain all the instance extensions, it doesn't check them in `get_info()`. This leads to any device level functions that are provided by instance extensions being missed.

layers/generated/vk_dispatch_table_helper.h
charles-lunarg: I do agree that this improvement doesn't match the table. The issue is 'knowing' whether the instance provided extension function is a device level function. 

I have half a mind to special case debug utils because I do not know any other case of an instance extension providing device level functions.

mikes-lunarg: Actually on second look I think we might be OK. The only function that calls this one has a map of name->dispatch-type and will still return `nullptr` for anything that isn't `kFuncTypeDev`
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/fc1c6286054e5772a506e95d67503dfc9bb2182b/layers/generated/chassis.cpp#L192-L208

mikes-lunarg (approved):
After looking at this more, I think it is fine.

I did attempt to write a [test](https://github.com/mikes-lunarg/Vulkan-ValidationLayers/commit/1d775a5c219f5b9ad090d774e9f0b6a1203aa2d3) but I couldn't get it to fail without modifiy these lines in the loader: https://github.com/KhronosGroup/Vulkan-Loader/blob/master/loader/generated/vk_loader_extensions.c#L783-L790. It seems a little strange to me that the dispatch table there is still populated with `gipa` instead of `gdpa`. Can you double check if that is correct?

charles-lunarg:
I noticed the use of gipa there and I suspect it’s due to some misconception about querying the function early on. It should be possible to recreate the issue with any layer that supports debug util functions and uses gdpa to get then, like renderdoc. I’m hesitant to change the loader here because who knows which layers were “depending” on the loader to query using GIPA. Though that’s a weak argument because anytime two layers have to work together that assumption goes out the window.

juan-lunarg (approved):
LGTM minor doc issues

PR #4850, opened 2022-11-22
cmake: Fix incorrect else() argument warning

karl-lunarg:
CMake requires the same argument in if(), else(), endif() sequences.

"elseif(LINUX)" may have been intended here, but simply removing the LINUX from the else() retains current behavior and is arguably clearer than replacing it with "APPLE" to match the else with the if.


juan-lunarg (approved):

juan-lunarg:
CI Vulkan-ValidationLayers test failures are completely unrelated

Issue #4714, opened 2022-10-26
Wrong reports of invalid buffers when using device address


llatta:
**Describe the Issue**
I've been digging into validation error message like this:

`Validation Error: [ VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03802 ] Object 0: handle = 0xc5ccdfb0, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x2ef78eeb | vkBuildAccelerationStructuresKHR(): The buffer associated with pInfos[0].scratchData.deviceAddress 188c2d00 is not a valid device address. The Vulkan spec states: For each element of pInfos, its scratchData.deviceAddress member must be a valid device address obtained from vkGetBufferDeviceAddress (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03802)`

I've tracked this down into a bug in the ValidationStateTracker when maintaining the buffer_address_map_ .
The steps leading to this condition are:

1. Buffer A gets allocated with vkCreateBuffer
2. Buffer A's device address gets fetched with vkGetDeviceAddress, which adds it to the ranges in the buffer_address_map_
3. Buffer B gets allocated with vkCreateBuffer. It does not have a device address usage flag
4. Buffer B gets deleted with vkDestroyBuffer

The code in the destroy buffer handling is this:
```
void ValidationStateTracker::PreCallRecordDestroyBuffer(VkDevice device, VkBuffer buffer, const VkAllocationCallbacks *pAllocator) {
    auto buffer_state = Get<BUFFER_STATE>(buffer);
    if (buffer_state) {
        WriteLockGuard guard(buffer_address_lock_);
        buffer_address_map_.erase_range(buffer_state->DeviceAddressRange());
    }
...
```

Since buffer B has never queried a device address, the deviceAddress stored in buffer_state is 0, and the returned DeviceAddressRange() is { 0, bufferSize }.
Buffer B is relatively large (>500 MB in my case), and this erase_range operation will remove all buffers from the buffer_address_map_ that have addresses lower than this value, which buffer A falls under.

I think the code in PreCallRecordDestroyBuffer probably needs to be changed to check for deviceAddress != 0 before erasing from the map.

**Valid Usage ID**
See above: VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03802

**Environment:**
 - OS: Windows 10
[...]

Issue #4556, opened 2022-09-15
Missing VUIDs for vkCmdTraceRays(Indirect)KHR


arno-lunarg:
**Valid Usage IDs requested**
A bunch of VUIDs related to the Khronos trace rays commands need to be implemented:

VUID-vkCmdTraceRaysKHR-pRayGenShaderBindingTable-03680
VUID-vkCmdTraceRaysKHR-pRayGenShaderBindingTable-03681
VUID-vkCmdTraceRaysKHR-pMissShaderBindingTable-03683 
VUID-vkCmdTraceRaysKHR-pMissShaderBindingTable-03684 
VUID-vkCmdTraceRaysKHR-pHitShaderBindingTable-03687 
VUID-vkCmdTraceRaysKHR-pHitShaderBindingTable-03688 
VUID-vkCmdTraceRaysKHR-pCallableShaderBindingTable-03691 
VUID-vkCmdTraceRaysKHR-pCallableShaderBindingTable-03692 

VUID-VkStridedDeviceAddressRegionKHR-size-04631 
VUID-VkStridedDeviceAddressRegionKHR-size-04632

VUID-vkCmdTraceRaysIndirectKHR-pRayGenShaderBindingTable-03680
VUID-vkCmdTraceRaysIndirectKHR-pRayGenShaderBindingTable-03681
VUID-vkCmdTraceRaysIndirectKHR-pMissShaderBindingTable-03683
VUID-vkCmdTraceRaysIndirectKHR-pMissShaderBindingTable-03684
VUID-vkCmdTraceRaysIndirectKHR-pHitShaderBindingTable-03687 
VUID-vkCmdTraceRaysIndirectKHR-pHitShaderBindingTable-03688 
VUID-vkCmdTraceRaysIndirectKHR-pCallableShaderBindingTable-03691 
VUID-vkCmdTraceRaysIndirectKHR-pCallableShaderBindingTable-03692

And some log messages for existing VUIDs in the same validation functions need to be updated, the ones that just copied the VUID message.

PR #4861, opened 2022-11-24
Add several RTX VUIDs, Update `buffer` <-> `device address` mapping

arno-lunarg:
Objectives:
=
- Add VUIDs listed in https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4556
- Still some TODOs to properly handle new `GetBuffersByAddress` calls introduced by rebasing on master
- Fix for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4714

Summary:
=
Aside adding some VUIDS, how the mapping between `VkBuffer` and `VkDeviceAddress` has been updated. This mapping is not 1:1, multiple `VkBuffer` can return the same starting `VkDeviceAddress`. It can happen if for instance 2 `VkBuffer`s are bound to the same memory region. See 118ec423c61f9c0afc28e6be9f717a081743dc8b for how the mapping has been updated.

Tests:
`BuffersAndBufferDeviceAddressesMapping`: Test how that adding and removing elements in the buffer <-> address map is correctly handled.
`ValidateCmdTraceRaysKHR`: Added new test cases

Where reviewers are especially needed:
=

- I need to make sure update to the aforementioned mapping were done where needed, especially I am not sure updates made to `RecordDeviceAccelerationStructureBuildInfo` and how it now stores multiple buffers instead of one are correctly propagated to code portions that consume that data.

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -10648,17 +10648,24 @@ bool CoreChecks::ValidateAccelerationBuffers(uint32_t info_index, const VkAccele
 
     auto buffer_check = [this, info_index, func_name](uint32_t gi, const VkDeviceOrHostAddressConstKHR address,
                                                       const char *field) -> bool {
-        // being inside the lambda confuses the compiler about our constness
-        const auto buffer_state = const_cast<const CoreChecks *>(this)->GetBufferByAddress(address.deviceAddress);
-        if (buffer_state &&
-            !(buffer_state->createInfo.usage & VK_BUFFER_USAGE_ACCELERATION_STRUCTURE_BUILD_INPUT_READ_ONLY_BIT_KHR)) {
+        const auto buffer_states = GetBuffersByAddress(address.deviceAddress);
+        const bool no_valid_buffer_found =
+            !buffer_states.empty() &&
+            std::none_of(
+                buffer_states.begin(), buffer_states.end(), [](const ValidationStateTracker::BUFFER_STATE_PTR &buffer_state) {
+                    return buffer_state->createInfo.usage & VK_BUFFER_USAGE_ACCELERATION_STRUCTURE_BUILD_INPUT_READ_ONLY_BIT_KHR;
+                });
+        if (no_valid_buffer_found) {
             LogObjectList objlist(device);
-            objlist.add(buffer_state->Handle());
+            for (const auto &buffer_state : buffer_states) {
+                objlist.add(buffer_state->Handle());
+            }
             return LogError(objlist, "VUID-vkCmdBuildAccelerationStructuresKHR-geometry-03673",
-                            "%s(): The buffer associated with pInfos[%" PRIu32 "].pGeometries[%" PRIu32
-                            "].%s was not created with VK_BUFFER_USAGE_ACCELERATION_STRUCTURE_BUILD_INPUT_READ_ONLY_BIT_KHR.",
+                            "%s(): No buffer associated with pInfos[%" PRIu32 "].pGeometries[%" PRIu32
+                            "].%s was created with VK_BUFFER_USAGE_ACCELERATION_STRUCTURE_BUILD_INPUT_READ_ONLY_BIT_KHR.",
                             func_name, info_index, gi, field);
         }
+
         return false;
     };
 
@@ -10697,18 +10704,26 @@ bool CoreChecks::ValidateAccelerationBuffers(uint32_t info_index, const VkAccele
         }
     }
 
-    const auto buffer_state = GetBufferByAddress(info.scratchData.deviceAddress);
-    if (!buffer_state) {
+    const auto buffer_states = GetBuffersByAddress(info.scratchData.deviceAddress);
+    if (buffer_states.empty()) {
         skip |= LogError(device, "VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03802",
                 "vkBuildAccelerationStructuresKHR(): The buffer associated with pInfos[%" PRIu32
                 "].scratchData.deviceAddress %" PRIx64 " is not a valid device address.",
                 info_index, info.scratchData.deviceAddress);
     } else {
-        if (!(buffer_state->createInfo.usage & VK_BUFFER_USAGE_STORAGE_BUFFER_BIT)) {
-            skip |= LogError(device, "VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03674",
-                    "vkBuildAccelerationStructuresKHR(): The buffer associated with pInfos[%" PRIu32
-                    "].scratchData.deviceAddress was not created with VK_BUFFER_USAGE_STORAGE_BUFFER_BIT bit.",
-                    info_index);
+        const bool no_valid_buffer_found = std::none_of(
+            buffer_states.begin(), buffer_states.end(), [](const ValidationStateTracker::BUFFER_STATE_PTR &buffer_state) {
+                return buffer_state->createInfo.usage & VK_BUFFER_USAGE_STORAGE_BUFFER_BIT;
+            });
+        if (no_valid_buffer_found) {
+            LogObjectList objlist(device);
+            for (const auto &buffer_state : buffer_states) {
+                objlist.add(buffer_state->Handle());
+            }
+            skip |= LogError(objlist, "VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03674",
+                             "vkBuildAccelerationStructuresKHR(): No buffer associated with pInfos[%" PRIu32
+                             "].scratchData.deviceAddress was created with VK_BUFFER_USAGE_STORAGE_BUFFER_BIT bit.",
+                             info_index);
         }
     }
 
--- a/layers/drawdispatch.cpp
+++ b/layers/drawdispatch.cpp
@@ -2228,61 +2228,261 @@ bool CoreChecks::PreCallValidateCmdTraceRaysNV(VkCommandBuffer commandBuffer, Vk
     return skip;
 }
 
+bool CoreChecks::ValidateRaytracingShaderBindingTable(VkCommandBuffer commandBuffer, const char *rt_func_name,
+                                                      const char *vuid_single_device_memory, const char *vuid_binding_table_flag,
+                                                      const VkStridedDeviceAddressRegionKHR &binding_table,
+                                                      const char *binding_table_name) const {
+    bool skip = false;
+
+    const auto buffer_states = GetBuffersByAddress(binding_table.deviceAddress);
+    if (buffer_states.empty()) {
+        // We can get here if for instance deviceAddress is 0
+        skip |= LogError(device, vuid_single_device_memory, "%s: no buffer is associated with %s->deviceAddress (0x%" PRIx64 ").",
+                         rt_func_name, binding_table_name, binding_table.deviceAddress);
+    } else {
+        // Try to find a buffer satisfying all VUIDs
+        const bool no_valid_buffer_found = std::none_of(
+            buffer_states.begin(), buffer_states.end(),
+            [&binding_table](const ValidationStateTracker::BUFFER_STATE_PTR &buffer_state) {
+                assert(buffer_state);
+
+                if (!buffer_state) {
+                    return false;
+                }
+
+                if (!(static_cast<uint32_t>(buffer_state->createInfo.usage) & VK_BUFFER_USAGE_SHADER_BINDING_TABLE_BIT_KHR)) {
+                    return false;
+                }
+                if (const auto mem_state = buffer_state->MemState(); !mem_state || mem_state->Destroyed()) {
+                    return false;
+                }
+                if (binding_table.size != 0) {
+                    const auto device_address_range = buffer_state->DeviceAddressRange();
+                    const sparse_container::range<VkDeviceSize> requested_range(
+                        binding_table.deviceAddress, binding_table.deviceAddress + binding_table.size - 1);
+                    if (!device_address_range.includes(requested_range)) {
+                        return false;
+                    }
+
+                    if (binding_table.size != 0 && binding_table.stride > buffer_state->createInfo.size) {
+                        return false;
+                    }
+                }
+
+                return true;
+            });
+
+        // If no valid buffer was found, for each violated VUID,
+        // output the list of buffers (associated to binding_table.deviceAddress) that violates it,
+        // alongside relevant info.
+        if (no_valid_buffer_found) {
+            struct InvalidBuffers {
+                LogObjectList buffers;
+                std::string error_msg;
+            };
+
+            InvalidBuffers vuid_binding_table_flag_invalid_buffers{{commandBuffer}};
+            InvalidBuffers vuid_address_range_invalid_buffers{{commandBuffer}};
+            InvalidBuffers vuid_stride_invalid_buffers{{commandBuffer}};
+            const std::string address_string = std::to_string(binding_table.deviceAddress);
+            const sparse_container::range<VkDeviceSize> requested_range(binding_table.deviceAddress,
+                                                                        binding_table.deviceAddress + binding_table.size - 1);
+            std::string requested_range_string =
+                '[' + std::to_string(requested_range.begin) + ", " + std::to_string(requested_range.end) + ')';
+            std::stringstream error_msg_prefix_ss;
+            error_msg_prefix_ss << rt_func_name << ": No buffer associated to" << binding_table_name << "->deviceAddress (0x"
+                                << address_string
+                                << ") was found such that valid usage passes. "
+                                   "At least one buffer associated to this device address must be valid. The following buffers ";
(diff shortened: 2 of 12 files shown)


layers/state_tracker.h
```diff
@@ -443,25 +443,23 @@ class ValidationStateTracker : public ValidationObject {
             return nullptr;
         }
         return found_it->second;
-    }
-
-    std::shared_ptr<BUFFER_STATE> GetBufferByAddress(VkDeviceAddress address) {
+    };
+    using BUFFER_STATE_PTR = std::shared_ptr<BUFFER_STATE>;
+    layer_data::span<BUFFER_STATE_PTR> GetBuffersByAddress(VkDeviceAddress address) {
```
spencer-lunarg, line 448: I know there was some spec language I think added to clarify this, can you add a comment here explaining why we can have multiple buffers, or quote the spec in the comment

layers/core_validation.cpp
```diff
@@ -10548,16 +10548,24 @@ bool CoreChecks::ValidateAccelerationBuffers(uint32_t info_index, const VkAccele
...
+        if (no_valid_buffer_found) {
             LogObjectList objlist(device);
-            objlist.add(buffer_state->Handle());
+            for (const auto &buffer_state : buffer_states) {
+                objlist.add(buffer_state->Handle());
+            }
             return LogError(objlist, "VUID-vkCmdBuildAccelerationStructuresKHR-geometry-03673",
-                            "%s(): The buffer associated with pInfos[%" PRIu32 "].pGeometries[%" PRIu32
-                            "].%s was not created with VK_BUFFER_USAGE_ACCELERATION_STRUCTURE_BUILD_INPUT_READ_ONLY_BIT_KHR.",
+                            "%s(): No buffer associated with pInfos[%" PRIu32 "].pGeometries[%" PRIu32
```
spencer-lunarg, line 10564: > VUID-vkCmdBuildAccelerationStructuresKHR-geometry-03673
The buffers from which the buffer device addresses for all of the ...  must have been created with the VK_BUFFER_USAGE_ACCELERATION_STRUCTURE_BUILD_INPUT_READ_ONLY_BIT_KHR usage flag

But seems if only 1 of many `buffer_state` use the usage this will not be triggered

layers/drawdispatch.cpp
```diff
@@ -2069,119 +2069,261 @@ bool CoreChecks::PreCallValidateCmdTraceRaysNV(VkCommandBuffer commandBuffer, Vk
...
+            error_msg_prefix += ": No buffer associated to";
+            error_msg_prefix += binding_table_name;
+            error_msg_prefix += "->deviceAddress (0x";
+            error_msg_prefix += address_string;
+            error_msg_prefix += ") was found such that valid usage passes. The following buffers ";
+            for (const auto &buffer_state : buffer_states) {
+                skip |= ValidateMemoryIsBoundToBuffer(buffer_state.get(), rt_func_name, vuid_single_device_memory);
+
+                if (!(static_cast<uint32_t>(buffer_state->createInfo.usage) & VK_BUFFER_USAGE_SHADER_BINDING_TABLE_BIT_KHR)) {
+                    vuid_binding_table_flag_invalid_buffers.buffers.add(buffer_state->Handle());
```
spencer-lunarg, line 2129: I guess why not just print a `LogError` for each buffer state instead of needing all this "error string concatenation" logic

layers/drawdispatch.cpp
```diff
@@ -2069,119 +2069,261 @@ bool CoreChecks::PreCallValidateCmdTraceRaysNV(VkCommandBuffer commandBuffer, Vk
...
+
+    return skip;
+}
+
+bool CoreChecks::ValidateCmdTraceRaysKHR(bool isIndirect, VkCommandBuffer commandBuffer,
+                                         const VkStridedDeviceAddressRegionKHR *pRaygenShaderBindingTable,
+                                         const VkStridedDeviceAddressRegionKHR *pMissShaderBindingTable,
+                                         const VkStridedDeviceAddressRegionKHR *pHitShaderBindingTable,
+                                         const VkStridedDeviceAddressRegionKHR *pCallableShaderBindingTable) const {
+    assert(pRaygenShaderBindingTable);
```
spencer-lunarg, line 2209: these should not be needed, there should be a VUID like ` VUID-vkCmdTraceRaysKHR-pRaygenShaderBindingTable-parameter` that happens in the Stateless layer and if an error is caught in stateless it never makes it to `CoreChecks`

layers/drawdispatch.cpp
```diff
@@ -2069,119 +2069,261 @@ bool CoreChecks::PreCallValidateCmdTraceRaysNV(VkCommandBuffer commandBuffer, Vk
...
+        skip |= LogError(device, vuid,
                          "vkCmdTraceRaysKHR: A valid pipeline must be bound to the pipeline bind point used by this command.");
     } else {  // bound to valid RT pipeline
-        if (pipeline_state->GetPipelineCreateFlags() & VK_PIPELINE_CREATE_RAY_TRACING_NO_NULL_INTERSECTION_SHADERS_BIT_KHR) {
-            if (!pHitShaderBindingTable->deviceAddress) {
-                skip |= LogError(device, "VUID-vkCmdTraceRaysKHR-flags-03697",
-                                 "vkCmdTraceRaysKHR: If the currently bound ray tracing pipeline was created with flags "
-                                 "that included VK_PIPELINE_CREATE_RAY_TRACING_NO_NULL_INTERSECTION_SHADERS_BIT_KHR, the "
-                                 "deviceAddress member of pHitShaderBindingTable must not be zero.");
+        std::string pHitShaderBindingTable_error_msg;
```
spencer-lunarg, line 2224: Instead of doing this, just have the error message say something like

```c++
 skip |= LogError(device, vuid, "vkCmdTraceRaysKHR: ...  pHitShaderBindingTable->size (%" PRIu32 ") and pHitShaderBindingTable->stride (%" PRIu32 ")", pHitShaderBindingTable->size, pHitShaderBindingTable->stride;
```

the idea is to not waste time with strings unless there is an actual error and printing the value out is always helpful/good practice for any `LogError`

layers/core_validation.cpp
arno-lunarg: This vuid was written prior to the spec update on buffer device address mapping. So from my understanding, a way to rephrase this vuid with the spec change in mind would be:
A buffer associated to `geometry.triangles.vertexData` must have been created with BUILD_INPUT_READ_ONLY flag.
A buffer associated to `geometry.triangles.indexData` must been created with BUILD_INPUT_READ_ONLY flag.
etc...

The spec change:
>If multiple VkBuffer objects are bound to overlapping ranges of VkDeviceMemory, implementations may return address ranges which overlap. In this case, it is ambiguous which VkBuffer is associated with any given device address. For purposes of valid usage, if multiple VkBuffer objects can be attributed to a device address, a VkBuffer is selected such that valid usage passes, if it exists.

layers/drawdispatch.cpp
arno-lunarg: I think it makes the end result way clearer, and also this way of logging is in the spirit of the spec change for buffer <-> device address mapping. The update mentions "multiple VkBuffer objects [that] can be attributed to a device address", so I think grouping them in the same error message makes sense.

layers/core_validation.cpp
arno-lunarg: I do not think that in "The buffers from which the buffer device addresses", "the buffers" refers to all buffers retrieved from a device address. It must refer to the set {geometry.triangles.vertexData, geometry.triangles.indexData, ...} with the assumption that 1 address == 1 buffer.

layers/state_tracker.cpp
```diff
@@ -368,7 +368,18 @@ void ValidationStateTracker::PostCallRecordCreateBuffer(VkDevice device, const V
             WriteLockGuard guard(buffer_address_lock_);
             // address is used for GPU-AV and ray tracing buffer validation
             buffer_state->deviceAddress = opaque_capture_address->opaqueCaptureAddress;
-            buffer_address_map_.insert({buffer_state->DeviceAddressRange(), buffer_state});
+            const auto address_range = buffer_state->DeviceAddressRange();
```
jeremyg-lunarg, line 371: I *think* you're using a version of std::map::insert that returns std::pair<iterator, bool> with see: https://en.cppreference.com/w/cpp/container/map/insert
Could you use the bool 'inserted or not' value to make this simpler?

layers/state_tracker.h
```diff
@@ -443,25 +443,29 @@ class ValidationStateTracker : public ValidationObject {
...
         if (found_it == buffer_address_map_.end()) {
-            return nullptr;
+            return layer_data::make_span<BUFFER_STATE_PTR>(nullptr, static_cast<size_t>(0));
         }
-        // NOTE: for the address map found_it is the actual map entry rather than a copy so we cannot std::move
         return found_it->second;
     }
 
-    std::shared_ptr<const BUFFER_STATE> GetBufferByAddress(VkDeviceAddress address) const {
+    layer_data::span<const BUFFER_STATE_PTR> GetBuffersByAddress(VkDeviceAddress address) const {
```
jeremyg-lunarg, line 463: Does this code handle the case where buffers don't start at the same address but partially overlap? For example, what if there are buffers with address ranges (0x1000 - 0x5000), (0x2000 - 0x6000), (0x3000 - 0x4000) and I try to look up with addresses 0x3500 and then 0x4500?

layers/state_tracker.cpp
arno-lunarg: Oh yes, I did not look at the return value of insert. And now that with have c++ 17 I can do this:
```cpp
if (auto [it, was_inserted] = buffer_address_map_.insert({address_range, {buffer_state}}); !was_inserted) {
    auto &buffer_list = it->second;
    const auto search = std::find(buffer_list.begin(), buffer_list.end(), buffer_state);
    // Do not add the same buffer twice
    if (search == buffer_list.end()) {
        buffer_list.emplace_back(buffer_state);
    }
}
```

layers/state_tracker.h
arno-lunarg: Oof no I think this case is not handled. Will fix it. At first glance, it does not look so trivial to do. I need to look further into `buffer_address_map_`'s type, `sparse_container::range_map<VkDeviceAddress, std::vector<std::shared_ptr<BUFFER_STATE>>>`

layers/drawdispatch.cpp
```diff
@@ -2069,119 +2069,256 @@ bool CoreChecks::PreCallValidateCmdTraceRaysNV(VkCommandBuffer commandBuffer, Vk
...
+                            error_msg += "do not include ";
+                            error_msg += binding_table_name;
+                            error_msg += " buffer device address range ";
+                            error_msg += requested_range_string;
+                            error_msg += ":\n";
+                        }
+                        const auto obj_index = vuid_address_range_invalid_buffers.buffers.object_list.size() - 1;
+                        char buffer_address_range_string[64];
+                        snprintf(buffer_address_range_string, 64, "[0x%" PRIx64 ", 0x%" PRIx64 ")", buffer_address_range.begin,
+                                 buffer_address_range.end);
```
ncesario-lunarg, line 2159: I would again use `std::to_string` here.

layers/drawdispatch.cpp
ncesario-lunarg: After an offline discussion with Arno and Jeremy, I think the current approach makes sense given our current logging. If our formatting changes and/or machine readable output gets added it might make sense to change this, but I don't think so.

ncesario-lunarg (commented):
Some minor nits/comments, but this LGTM once all comments are resolved.

layers/drawdispatch.cpp
jeremyg-lunarg: I added https://github.com/KhronosGroup/Vulkan-ValidationLayers/projects/2#card-87156609 which was something I mentioned in the discussion.

layers/state_tracker.cpp
arno-lunarg: This problem should be addressed now, using new methods I defined in the range map: `split_and_merge_insert` and `erase_range_or_touch`

layers/state_tracker.h
arno-lunarg: This problem should be addressed now, using new methods I defined in the range map: `split_and_merge_insert` and `erase_range_or_touch`

layers/core_validation.cpp
```diff
@@ -10548,16 +10548,24 @@ bool CoreChecks::ValidateAccelerationBuffers(uint32_t info_index, const VkAccele
     auto buffer_check = [this, info_index, func_name](uint32_t gi, const VkDeviceOrHostAddressConstKHR address,
                                                       const char *field) -> bool {
         // being inside the lambda confuses the compiler about our constness
-        const auto buffer_state = const_cast<const CoreChecks *>(this)->GetBufferByAddress(address.deviceAddress);
-        if (buffer_state &&
-            !(buffer_state->createInfo.usage & VK_BUFFER_USAGE_ACCELERATION_STRUCTURE_BUILD_INPUT_READ_ONLY_BIT_KHR)) {
+        const auto buffer_states = const_cast<const CoreChecks *>(this)->GetBuffersByAddress(address.deviceAddress);
```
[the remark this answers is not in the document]
arno-lunarg: removing it works locally, will see what CI tells

layers/drawdispatch.cpp
```diff
@@ -2069,119 +2069,256 @@ bool CoreChecks::PreCallValidateCmdTraceRaysNV(VkCommandBuffer commandBuffer, Vk
...
+            struct InvalidBuffers {
+                LogObjectList buffers;
+                std::string error_msg;
+            };
+
+            InvalidBuffers vuid_binding_table_flag_invalid_buffers;
+            InvalidBuffers vuid_address_range_invalid_buffers;
+            InvalidBuffers vuid_stride_invalid_buffers;
+            char address_string[32];
+            snprintf(address_string, 32, "%" PRIx64, binding_table->deviceAddress);
```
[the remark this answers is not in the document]
arno-lunarg: oh yes for whatever reason I was thinking that only `int` were working with `std::to_string`
Also, I am gonna change strings where I do a lot of `+=` to `std::stringstream` and use `<<` instead, less lines

layers/core_validation.cpp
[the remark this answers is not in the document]
ncesario-lunarg: Not necessary to mess around with in this PR, more of a question/note since this reminded me of this, and was hoping we could remove it if possible.

jeremyg-lunarg (commented):
This looks pretty good to me. I want to re-review the range_map parts again after a few hours break.  Everything about that class is super tricky!

layers/range_vector.h
```diff
@@ -634,6 +659,45 @@ class range_map {
...
+    // Range intersection is merged using merge_op.
+    // Ranges before and after this intersection are recursively inserted.
+    // merge_pos should have this signature: (mapped_type& current_value, const mapped_type& new_value) -> void
+    template <typename MergeOp>
+    iterator split_and_merge_insert(const value_type &value, const MergeOp &merge_op) {
+        if (!value.first.non_empty()) {
+            return end();
+        }
+
+        if (auto [it, was_inserted] = insert(value); !was_inserted) {
```
jeremyg-lunarg, line 673: I forgot that this syntax was part of C++17. Nice!

jeremyg-lunarg (approved):
LGTM

layers/range_vector.h
mark-lunarg: Aw, sheesh.  What's it called so I can look it up?

ncesario-lunarg (approved):

layers/core_validation.cpp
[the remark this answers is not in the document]
ncesario-lunarg: Thanks for trying this out @arno-lunarg, looks like it's works now!

layers/range_vector.h
ncesario-lunarg: https://en.cppreference.com/w/cpp/language/structured_binding

jeremyg-lunarg: https://en.cppreference.com/w/cpp/language/structured_binding  but I only knew about it because of python

layers/range_vector.h
```diff
@@ -411,12 +412,27 @@ class range_map {
...
+                current = impl_erase(current);
+            } else {
+                ++current;
+            }
         }
 
         if (!at_impl_end(current) && current->first.includes(bounds.end)) {
             // last entry extends past the end of the bounds range, snip to only erase the bounded section
-            current = split_impl(current, bounds.end, split_op_keep_upper());
+            current = split_impl(current, bounds.end, split_op_keep_both());
```
arno-lunarg, line 424: @jeremyg-lunarg  oh yes the range_map class is tricky, and I decided to do some refactoring to add my new functions without relying on a copy pasta. The one part that could break existing behaviour is those changes from here line 424 -> 435

PR #4863, opened 2022-11-25
Python raise requires an Exception class

dneto0:


spencer-lunarg:
@dneto0 

1. can you please re commit as `scripts: Python raise requires an Exception class` to make the CI happy
2. How did you come across this issue, want to prevent the next `raise` added by someone to not regress this PR

juan-lunarg (approved):

PR #4865, opened 2022-11-26
Improve CONTRIBUTING.md  doc

spencer-lunarg:
- Clean up file to be as simple as possible to make easier to read for new users
- Add "TODO" and "Commenting out Code" section
- Describe review system

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,22 +1,20 @@
-# How to Contribute to Vulkan Source Repositories
-
-## **The Repository**
+# How to Contribute to The Vulkan Validation Layer
 
 The source code for The Vulkan-ValidationLayer components is sponsored by Khronos and LunarG.
-* [Khronos Vulkan-ValidationLayers](https://github.com/KhronosGroup/Vulkan-ValidationLayers)
-
-## **The Vulkan Ecosystem Needs Your Help**
-
 The Vulkan validation layers make up a significant part of the Vulkan ecosystem.
 While there are often active and organized development efforts underway to improve their coverage,
 opportunities always exist for anyone to help by contributing additional validation layer checks
 and tests.
 
-There are a couple of methods to identify areas of need:
-* Examine the [issues list](https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues)
-in this repository and look for issues that are of interest
-* Examine the Validation Layer 'Coverage - html' page at [the Vulkan SDK
-documentation page](https://vulkan.lunarg.com/doc/sdk/latest/windows/validation_error_database.html) -- it lists all published Vulkan VUIDs and their status.
+It is the maintainers goal for all issues to be assigned or triaged within one business day of their submission. If you choose
+to work on an issue that is assigned, simply coordinate with the current assignee.
+
+## Incomplete VUIDs
+
+There are some VUID that are incomplete and need to be added. The following can be used to find them
+* [Incomplete tagged issues](https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues?q=is%3Aopen+is%3Aissue+label%3AIncomplete)
+* The 'Coverage - html' page at [the Vulkan SDK documentation page](https://vulkan.lunarg.com/doc/sdk/latest/windows/validation_error_database.html)
+  * it lists all published Vulkan VUIDs and their status.
 * Run `scripts/vk_validation_stats.py` with `-todo` to see a list of as-yet unimplemented validation checks.
   * ```bash
     # Get summary report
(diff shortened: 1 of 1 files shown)


ziga-lunarg (approved):

CONTRIBUTING.md
```diff
@@ -29,49 +27,46 @@ documentation page](https://vulkan.lunarg.com/doc/sdk/latest/windows/validation_
...
+* Source code must follow the repo coding style guidelines, including a pass through a clang-format utility
+* Test your change with [internal test suite](./tests)
 * The resulting Pull Request will be assigned to a repository maintainer. It is the maintainer's responsibility to ensure the Pull Request
   passes the Google/LunarG internal CI processes. Once the Pull Request has been approved and is passing internal CI, a repository maintainer
   will merge the PR.
+* Please base your fixes on the master branch. SDK branches are generally not updated except for critical fixes needed to repair an SDK release.
+
+If implementing a new VUID some extra considerations:
+* New VUID checks must be accompanied by a [relevant tests](./docs/creating_tests.md)
+* Validation source code should be in a separate commit from the tests, unless there are interdependencies. The repo should compile and pass all tests after each commit.
```
arno-lunarg, line 43: Compile and pass all tests after each commit? It seems drastic

arno-lunarg (commented):
LGTM! One comment about compilation/tests succeeding

CONTRIBUTING.md
spencer-lunarg: I would expect any commit to at least compile

I guess the goal of this is to say that any "partial" commit should still in theory work

CONTRIBUTING.md
```diff
@@ -1,22 +1,20 @@
...
 While there are often active and organized development efforts underway to improve their coverage,
 opportunities always exist for anyone to help by contributing additional validation layer checks
 and tests.
 
-There are a couple of methods to identify areas of need:
-* Examine the [issues list](https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues)
-in this repository and look for issues that are of interest
-* Examine the Validation Layer 'Coverage - html' page at [the Vulkan SDK
-documentation page](https://vulkan.lunarg.com/doc/sdk/latest/windows/validation_error_database.html) -- it lists all published Vulkan VUIDs and their status.
+It is the maintainers goal for all issues to be assigned or triaged within one business day of their submission. If you choose
```
aitor-lunarg, line 9: While I do think it would be ideal to have issues either assigned or triaged in a single day of their submission, I think it may put expectations too high for the public and adding extra pressure on having new issues processed when the team may be focused on something else

CONTRIBUTING.md
```diff
@@ -1,22 +1,20 @@
...
+to work on an issue that is assigned, simply coordinate with the current assignee.
```
aitor-lunarg, line 10: I do agree with this, but we also need to assign issues to the pool if we are not working on them (I'm looking at myself, but there's probably more people in the same boat). I think this should be stated somewhere, also, before unassigning to have the issue back in the pool, the assigned should leave some comments with findings or something about the current progress.

CONTRIBUTING.md
aitor-lunarg: I do also think that it should compile at least. Testing every single test when adding a new VUID seems excessive since there's CI to check for that, plus the user may not have all the configurations tests. On the other hand, we don't want to drown CI with changes that fail...

aitor-lunarg (commented):
I left a few comments on parts where I think there should be some discussion before proceeding with the merge since we would be committing to those. Other than that, looks good to me!

CONTRIBUTING.md
spencer-lunarg: So for my "triaged" means

- First decide if actually a real issue (often people raise spec bugs and we need to tell them to move it to the correct repo)
- Next decide what to label it as (bug, incomplete VUID, nice-to-have, we are breaking ANGLE's build, etc)
- Decide how important it is (somethings are in nature going to take priority over othere)
- Label it with a GitHub tag and then assign to someone

The goal is not to have no action at all on an issue, want to at least show "we are aware of the issue" which is NOT the same as "we are going to fix it NOW"

You raised a good concern.... I will update what `triage` means in this context in the doc

CONTRIBUTING.md
```diff
@@ -146,12 +135,32 @@ corresponding generated output to keep the repository self-consistent. [Here for
...
+- `-2` | code must not merge
+    - GitHub's `Request changes` option
+- `-1` | something seems wrong, but won't block merging
+    - Normal review comment
+- `+1` | code seems good, but not confident to give a `+2`
+    - Add `LGTM` comment (looks good to me)
+- `+2` | code should merge
+    - GitHub's `Approve` option
+
+All PR's must have been marked as `Approve` by 1 maintainer and have no outstanding request changes.
```
jeremyg-lunarg, line 151: So we should be adding +1/+2 in comments when reviewing? Who is allowed to +2?  I like this idea, I wish github had more of gerrit's good features.

jeremyg-lunarg (approved):
+1

CONTRIBUTING.md
mark-lunarg: This requirement was chosen for a few reasons but one of the big ones is for `git bisect`.  Each commit should run and pass all tests, otherwise the needless test failures will make bisecting ineffective or impossible.  And, the discipline entailed is good for organizing changes anyway (but it does take more thought and effort -- interactive rebase is your friend).

CONTRIBUTING.md
jeremyg-lunarg: Yeah we need to set expectations on who can approve, and maybe point to a github group that defines the 'who'.  Making +2 = 'Approve button clicked by maintainer' works for me. -1/-2 seems harder. If something has an identifiable bug in it, I really want to put 'Changes requested' on it.  -2 seems like it happens fairly rarely and usually results in off line discussion with relevant parties.

CONTRIBUTING.md
jzulauf-lunarg: @mark-lunarg is exactly correct.  Changes must be bisectable.  There are times this couple test/validation changes (especially for a correction to validation logic) awkwardly, but bisectability is very important for regression debugging etc.

juan-lunarg (changes requested):
I don't see where in our docs we tell contributors to run `check_code_format.py` before making a PR. We have instructions for running clang-format directly, but not the python script which also checks the commit

Matraqueeheh (approved):

spencer-lunarg:
Closing - going to redo in smaller PRs

Issue #4014, opened 2022-04-14
Remove references to devsim


ncesario-lunarg:
Devsim has been removed and effectively replaced by profiles. As such, all references to devsim should be removed from the validation layers repository.

See https://gitlab.khronos.org/vulkan/Vulkan-SDK-Packaging/-/issues/1030 for more details.

PR #4866, opened 2022-11-28
tests: Replace devsim with Profile layer

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4014

The old DevSim layer has been replaced in favor of the new Profile Layer. 

- Removes all logic/reference to devsim
- Adds to doc how to use profile layer instead with MockICD
- replaces files in `tests/device_profiles` (but need to still re-enable testing with them)


docs/portability_validation.md
```diff
@@ -10,21 +10,18 @@
 ## Requirements
 
 Either
-- The [VK_LAYER_LUNARG_device_simulation](https://vulkan.lunarg.com/doc/sdk/1.2.154.1/windows/device_simulation_layer.html) layer with portability enabled _or_
+- The [VK_LAYER_KHRONOS_profiles](https://github.com/KhronosGroup/Vulkan-Profiles) layer with portability enabled _or_
 - A driver that supports `VK_KHR_portability_subset`.
 
 ## Running the tests
```
juan-lunarg, line 16:
```suggestion
## Running the portability tests
```

layers/subresource_adapter.cpp
```diff
@@ -282,7 +282,7 @@ ImageRangeEncoder::ImageRangeEncoder(const IMAGE_STATE& image, const AspectParam
     VkImageSubresourceLayers subres_layers = {limits_.aspectMask, 0, 0, limits_.arrayLayer};
     linear_image_ = false;
 
-    // WORKAROUND for dev_sim and mock_icd not containing valid VkSubresourceLayout yet. Treat it as optimal image.
+    // WORKAROUND for dev_sim/profile and mock_icd not containing valid VkSubresourceLayout yet. Treat it as optimal image.
```
juan-lunarg, line 285: Looking at the issue it says "all references to devsim should be removed from the validation layers repository."

Is there a reason to leave in the comment about dev_sim?

```suggestion
    // WORKAROUND for profile and mock_icd not containing valid VkSubresourceLayout yet. Treat it as optimal image.
```

tests/README.md
```diff
@@ -94,13 +94,17 @@ You will have to direct the loader from Vulkan-Loader to the MoltenVK ICD:
 export VK_ICD_FILENAMES=<path to MoltenVK repository>/Package/Latest/MoltenVK/macOS/MoltenVK_icd.json
 ```
 
-## Running Tests on DevSim and MockICD
+## Running Tests on MockICD and Profiles layer
```
juan-lunarg, line 97: Question. Do we run the MockICD with profile layers on CI?

tests/README.md
```diff
@@ -109,22 +113,32 @@ export VVL=/path/to/Vulkan-ValidationLayers
...
 
 # This step can be skipped if the Vulkan SDK is properly installed
-# Add path for device simulation layer
+# Add path for Profile Layer
 export VK_LAYER_PATH=$VK_LAYER_PATH:$VULKAN_SDK/etc/vulkan/explicit_layer.d/
 export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$VULKAN_SDK/lib/
 
 # Set MockICD to be driver
 export VK_ICD_FILENAMES=/path/to/Vulkan-Tools/build/icd/VkICD_mock_icd.json
+export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/path/to/Vulkan-Tools/build/icd/
```
juan-lunarg, line 122: I don't believe you need to modify the LD_LIBRARY_PATH in order to set the Mock ICD (at least not on my machine).

tests/README.md
```diff
@@ -109,22 +113,32 @@ export VVL=/path/to/Vulkan-ValidationLayers
...
+
+# Set Layers, the order here is VERY IMPORTANT otherwise the test will see the
+# profile layer but the layer won't see it and have a different value
+export VK_INSTANCE_LAYERS=VK_LAYER_KHRONOS_validation:VK_LAYER_KHRONOS_profiles
 
-# Set device simulation profile to a valid json file
+# Set device profile to a valid json file
 # There are a set of profiles used in CI in the device_profiles folder
-export VK_DEVSIM_FILENAME=$VVL/tests/device_profiles/mobile_chip.json
+export VK_KHRONOS_PROFILES_PROFILE_FILE=$VVL/tests/device_profiles/mobile_chip.json
```
juan-lunarg, line 130: mobile_chip.json was removed

tests/README.md
```diff
@@ -109,22 +113,32 @@ export VVL=/path/to/Vulkan-ValidationLayers
...
 # Set MockICD to be driver
 export VK_ICD_FILENAMES=/path/to/Vulkan-Tools/build/icd/VkICD_mock_icd.json
+export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/path/to/Vulkan-Tools/build/icd/
+
+# Set Layers, the order here is VERY IMPORTANT otherwise the test will see the
+# profile layer but the layer won't see it and have a different value
+export VK_INSTANCE_LAYERS=VK_LAYER_KHRONOS_validation:VK_LAYER_KHRONOS_profiles
 
-# Set device simulation profile to a valid json file
+# Set device profile to a valid json file
```
juan-lunarg, line 128: How are these profiles generated? May be useful to document.

tests/README.md
juan-lunarg: https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4867

tests/README.md
spencer-lunarg: where does it load the library (`.so`) from then?

juan-lunarg: The json produced by CMake uses a relative path since they are in the same directory. This only applies if you are using the CMake build (the installed json doesn't use a relative path). So your documentation is more generic/applicable.

layers/subresource_adapter.cpp
spencer-lunarg: thinking about it more I guess not, will remove them all where I did this `/profile` thing

tests/README.md
spencer-lunarg: good point, added

juan-lunarg (approved):
LGTM

docs/portability_validation.md
```diff
@@ -10,21 +10,18 @@
...
-# This environment variable must be set to ensure portability is enabled as there is currently no programatic way to enable portability in the devsim layer
-$env:VK_DEVSIM_EMULATE_PORTABILITY_SUBSET_EXTENSION="1"
+The [./tests README](../tests/README.md) explains how to use the `VK_LAYER_KHRONOS_profiles` to test the Validation Layers
 
-/path/to/Vulkan-ValidationLayers/build/tests/Debug/vk_layer_validation_tests.exe --gtest_filter=VkPortability*
+**NOTE** that for portability tests to make sure `VK_KHR_portability_subset` is exposed from the MockICD
+
+```patch
+-export VK_KHRONOS_PROFILES_EMULATE_PORTABILITY=true
++export VK_KHRONOS_PROFILES_EMULATE_PORTABILITY=false
```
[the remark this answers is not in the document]
spencer-lunarg: I wrote the other section which sets it to `true` but you are right, no real reason, will fix

Issue #4857, opened 2022-11-24
Move RTX tests into separate test file


spencer-lunarg:
we should have both a `tests/vklayertests_ray_tracing.cpp` and `tests/positive/ray_tracing.cpp` file for anything that is ray tracing related (kind of like https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4758/commits/f8b1aa08017a4b85131cd145bd9391c726872ad3)

1. Anything using the following extensions

- `VK_KHR_ACCELERATION_STRUCTURE_EXTENSION_NAME`
- `VK_KHR_RAY_TRACING_PIPELINE_EXTENSION_NAME`
- `VK_KHR_RAY_QUERY_EXTENSION_NAME`
- `VK_KHR_DEFERRED_HOST_OPERATIONS_EXTENSION_NAME`

should be moved there. (unless it is already part of something specific like `VkGpuAssistedLayerTest`)

2. There is a `InitFrameworkForRayTracingTest` function, it would be nice for all Ray Tracing tests to use the same setup, makes updating all the tests in the future easier

3. The tests should all have `RayTracing` in the name (regardless of which file they are in) so we can test with `--gtest_filter=*RayTracing*`

4. Anything else that can be unified across all the tests is a positive!

PR #4871, opened 2022-11-28
Move RTX tests to new files, some cleaning along the way

arno-lunarg:
- Issue https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4857

Summary
=
- Move ray tracing tests to their own files, depending on their nature
- Prefix them all by `RayTracing`
- Prefix NVIDIA specific tests by `NVRayTracing`, and group them
- Unify usage of `Get{Device, Instance}ProcAddr`
- Use `InitFrameworkForRayTracingTest` when relevant
- Use `VkLayerTest::GetPhysicalDeviceProperties2` instead of manual pointer acquisition & dispatch


tests/layer_validation_tests.h
```diff
@@ -270,6 +270,24 @@ class VkLayerTest : public VkRenderFramework {
...
     }
 
+    template <typename Proc>
+    const Proc GetInstanceProcAddr(const char *proc_name, bool assert_proc = true) {
+        auto proc = reinterpret_cast<Proc>(vk::GetInstanceProcAddr(instance(), proc_name));
+        if (assert_proc) {
+            assert(proc);
+        }
+        return proc;
+    }
```
juan-lunarg, line 280: Instead of having `assert_proc` be a runtime variable, what about a template parameter?

```suggestion
    template <typename Proc, bool assert_proc = true>
    const Proc GetInstanceProcAddr(const char *proc_name) {
        auto proc = reinterpret_cast<Proc>(vk::GetInstanceProcAddr(instance(), proc_name));
        if constexpr (assert_proc) {
            assert(proc);
        }
        return proc;
    }
```

tests/vklayertests_ray_tracing.cpp
```diff
@@ -0,0 +1,2416 @@
...
+    auto features2 = GetPhysicalDeviceFeatures2(acc_structure_features);
+    if (acc_structure_features.accelerationStructureHostCommands == VK_FALSE) {
+        GTEST_SKIP() << "accelerationStructureHostCommands feature not supported";
+    }
+
+    ASSERT_NO_FATAL_FAILURE(InitState(nullptr, &features2));
+
+    PFN_vkBuildAccelerationStructuresKHR vkBuildAccelerationStructuresKHR =
+        (PFN_vkBuildAccelerationStructuresKHR)vk::GetInstanceProcAddr(instance(), "vkBuildAccelerationStructuresKHR");
+    assert(vkBuildAccelerationStructuresKHR);
```
juan-lunarg, line 482: Use new helper function

tests/layer_validation_tests.h
```diff
@@ -270,6 +270,24 @@ class VkLayerTest : public VkRenderFramework {
...
+
+    template <typename Proc>
+    const Proc GetDeviceProcAddr(const char *proc_name, bool assert_proc = true) {
+        auto proc = reinterpret_cast<Proc>(vk::GetDeviceProcAddr(device(), proc_name));
```
juan-lunarg, line 284: Ensure the template parameter is a function.

```suggestion
        static_assert(std::is_function_v<Proc>);
        auto proc = reinterpret_cast<Proc>(vk::GetDeviceProcAddr(device(), proc_name));
```

tests/vklayertests_ray_tracing.cpp
```diff
@@ -0,0 +1,2416 @@
...
+    valid_geometry_triangles.geometry.triangles.vertexStride = 8;
+    valid_geometry_triangles.geometry.triangles.indexType = VK_INDEX_TYPE_UINT32;
+    valid_geometry_triangles.geometry.triangles.indexData.deviceAddress = indexAddress + kBufferOffset;
+    valid_geometry_triangles.geometry.triangles.transformData.deviceAddress = 0;
+    valid_geometry_triangles.geometry.triangles.maxVertex = 1;
+    valid_geometry_triangles.flags = 0;
+
+    auto vkGetPhysicalDeviceProperties2KHR = reinterpret_cast<PFN_vkGetPhysicalDeviceProperties2KHR>(
+        vk::GetInstanceProcAddr(instance(), "vkGetPhysicalDeviceProperties2KHR"));
+    ASSERT_TRUE(vkGetPhysicalDeviceProperties2KHR != nullptr);
```
juan-lunarg, line 1369: Should the new helper function use ASSERT_TRUE instead?

tests/layer_validation_tests.h
arno-lunarg: Yep this is better!

tests/layer_validation_tests.h
arno-lunarg: Oh I did not know about that, will do

tests/vklayertests_ray_tracing.cpp
arno-lunarg: Yes it seems better

arno-lunarg: Oh but I cannot call ASSERT_TRUE in a function called in a test, it is a weird macro, it returns a value...

juan-lunarg: If that's the case then regular assert is fine for the helper function.

tests/layer_validation_tests.h
```diff
@@ -270,6 +270,24 @@ class VkLayerTest : public VkRenderFramework {
         return GetPhysicalDeviceProperties2(props2);
     }
 
+    template <typename Proc>
+    const Proc GetInstanceProcAddr(const char *proc_name, bool assert_proc = true) {
```
juan-lunarg, line 274: Use `[[nodiscard]]` for these new helper functions

```suggestion
     [[nodiscard]]  const Proc GetInstanceProcAddr(const char *proc_name, bool assert_proc = true) {
```

juan-lunarg (changes requested):
Squash and rebase this PR into 1/2 commits. 12 commits is too much.

mark-lunarg:
Number of commits is fine, the smaller the better as long as they're grouped by functionality and are atomic in that they'll run and pass CI.

juan-lunarg:
> Number of commits is fine, the smaller the better as long as they're grouped by functionality and are atomic in that they'll run and pass CI.

I agree. I just looked at the recent commits and saw not each one is atomic and passes CI. Figured it'd be easier to just squash into 1 or 2.

tests/vklayertests_ray_tracing.cpp
spencer-lunarg: For now this is fine IMO, we do this all over, it is on my radar to fix across everywhere with a better system

spencer-lunarg: nvm, already added the logic I see, expands the scope of this PR a bit, but I think it is not too big of a change so ok with it

spencer-lunarg:
> Number of commits is fine, the smaller the better as long as they're grouped by functionality and are atomic in that they'll run and pass CI.

I agree, but also guess I feel this is on the verge of being too small IMO where you have to do extra bisect for overly small changes, also no reason IMO to add 2 commit for the function implementation and also the usage, when I look at a git blame of when a function is added, I want to see where it was initially used right away too

For now I am ok with it, but for example I would have grouped this PR as

- `test: Add new test files for ray tracing` (creating the file and just doing a `move` on the tests and the "grouping")
- `tests: Use InitFrameworkForRayTracingTest`
- `tests: Add VkLayerTest::Get***ProcAddr` (**both** creating and using it)
- `tests: Make RTX Tests consistent` (adding the spaces, etx)

spencer-lunarg (approved):
Nice cleanup of these tests into their own spot (will also make future RTX PR much nicer!). Still want @juan-lunarg code/git comments resolved first, but otherwise the LGTM!

arno-lunarg:
I consolidated the commits, there are 6 now. 
Trying to fixup the "group NV tests" into the first "add new RT tests files" commit gives me horrible merge issues because of the following "Use  InitFrameworkForRayTracingTest" commit. I could merge this commit into the first one, but I'd rather not, they solve different problems. 
There is still a commit for the off by one error, because again it is unrelated to the other ones.

Still, each commit should compile, and pass the tests now.

PR #4878, opened 2022-11-29
Fix VkPositiveLayerTest.ExportMetalObjects

### LINKED ISSUE #4873: Validation layers null out values returned from vkExportMetalObjectsEXT.
**Describe the Issue**
Not sure if this is the correct place for this bug, but if you enable the validation layers and try to use VK_EXT_metal_objects to get the metal objects, the value for the native metal object is always null. Here is an example segment of code that will reproduce this issue. If I enable validation layers, non of the layers are triggered, but the value in deviceInfo.mtlDevice is always null. If I turn off the validation layers the deviceInfo.mtlDevice value is populated with the correct device.

```c++
if (mVkExportMetalObjectsEXT)
{
    VkExportMetalObjectsInfoEXT objectsInfo{};
    VkExportMetalDeviceInfoEXT deviceInfo{};
    
    deviceInfo.sType = VK_STRUCTURE_TYPE_EXPORT_METAL_DEVICE_INFO_EXT;
    
    objectsInfo.sType = VK_STRUCTURE_TYPE_EXPORT_METAL_OBJECTS_INFO_EXT;
    objectsInfo.pNext = &deviceInfo;
    
    mVkExportMetalObjectsEXT(mDevice, &objectsInfo);
    
    return deviceInfo.mtlDevice;
}
```

**Valid Usage ID**
No validation messages encountered leading up to the issue.

**Environment:**
 - OS: MacOS 13.0.1
 - GPU: Radeon Pro 560X/Intel UHD Graphics 630 1536 MB/
 - SDK or header version if building from repo: 1.3.231.1
 - Options enabled (synchronization, best practices, etc.): VK_LAYER_KHRONOS_validation

**Additional context**
https://github.com/KhronosGroup/MoltenVK/issues/1787

juan-lunarg:


layers/generated/layer_chassis_dispatch.cpp
```diff
@@ -9534,27 +9545,7 @@ void DispatchDestroyPrivateDataSlotEXT(
...
-    safe_VkExportMetalObjectsInfoEXT var_local_pMetalObjectsInfo;
-    safe_VkExportMetalObjectsInfoEXT *local_pMetalObjectsInfo = NULL;
-    {
-        if (pMetalObjectsInfo) {
-            local_pMetalObjectsInfo = &var_local_pMetalObjectsInfo;
-            local_pMetalObjectsInfo->initialize(pMetalObjectsInfo);
-            WrapPnextChainHandles(layer_data, local_pMetalObjectsInfo->pNext);
-        }
-    }
-    layer_data->device_dispatch_table.ExportMetalObjectsEXT(device, (VkExportMetalObjectsInfoEXT*)local_pMetalObjectsInfo);
```
juan-lunarg, line 9554: Currently marked as a draft PR, since I'm not sure my new DispatchExportMetalObjectsEXT is safe.

tests/positive/other.cpp
```diff
@@ -776,3 +776,39 @@ TEST_F(VkPositiveLayerTest, EnumeratePhysicalDeviceGroups) {
     debug_reporter.Destroy(test_instance);
     vk::DestroyInstance(test_instance, nullptr);
 }
+
+#ifdef VK_USE_PLATFORM_METAL_EXT
+TEST_F(VkPositiveLayerTest, ExportMetalObjectsNotNull) {
```
juan-lunarg, line 781: I've verified the regression with this test. And verified this fixes the issue on a MacBook Pro that was lent to me.

layers/generated/layer_chassis_dispatch.cpp
```diff
@@ -1601,6 +1601,17 @@ void DispatchGetAccelerationStructureBuildSizesKHR(
...
 }
 
+#ifdef VK_USE_PLATFORM_METAL_EXT
+void DispatchExportMetalObjectsEXT(
+    VkDevice                     device,
+    VkExportMetalObjectsInfoEXT* pMetalObjectsInfo)
+{
+    auto layer_data = GetLayerDataPtr(get_dispatch_key(device), layer_data_map);
+    if (pMetalObjectsInfo != nullptr) {
+        layer_data->device_dispatch_table.ExportMetalObjectsEXT(device, pMetalObjectsInfo);
```
mark-lunarg, line 1611: I think this function needs to contain the generated code that the original contained -- if any structure in the pNext chain of `vkExportMetalObjectsEXT()` has a handle in it, the display driver won't know what it is and might crash.  I think the old code needs to be copied back in, and then after the dispatch call is made, you'd need to copy the appropriate `MTLXxx_id` from the temporary structure _back_ into the original pMetalObjectsInfo structure.  This thing is so crazy overloaded (16 possible pNext structure types!) that hopefully there's a clever way to do this, but it might just be a huge loop/switch statement.

tests/positive/other.cpp
```diff
@@ -776,3 +776,39 @@ TEST_F(VkPositiveLayerTest, EnumeratePhysicalDeviceGroups) {
...
+    auto portability_features = LvlInitStruct<VkPhysicalDevicePortabilitySubsetFeaturesKHR>();
+    auto features2 = GetPhysicalDeviceFeatures2(portability_features);
+
+    ASSERT_NO_FATAL_FAILURE(InitState(nullptr, &features2));
+
+    auto vkExportMetalObjectsEXT =
+        reinterpret_cast<PFN_vkExportMetalObjectsEXT>(vk::GetDeviceProcAddr(m_device->device(), "vkExportMetalObjectsEXT"));
+    ASSERT_TRUE(vkExportMetalObjectsEXT != nullptr);
+
+    auto deviceInfo = LvlInitStruct<VkExportMetalDeviceInfoEXT>();
```
mark-lunarg, line 807: I think you might've picked the one `pNext` struct for this guy that doesn't have a VkObject handle in it.  For instance, the `VkExportMetalCommandQueueInfoEXT` structure has a `VkQueue` that would need to be unwrapped before calling down to the driver.  Thanks Apple!

mark-lunarg (changes requested):
Sorry I was so confusing with my advice yesterday -- let me know what additional information I can supply.

layers/generated/layer_chassis_dispatch.cpp
Tony-LunarG, line 1611: I think you are going to run into handle wrapping issues this way.  For instance, if pMetalObjectsInfo.pNext points to a VkExportMetalTextureInfoEXT, the image or imageview or bufferview parameter will have been wrapped when it was created and needs to be unwrapped before sending it to the driver.

Tony-LunarG: Or what Mark said...

tests/positive/other.cpp
```diff
@@ -776,3 +776,39 @@ TEST_F(VkPositiveLayerTest, EnumeratePhysicalDeviceGroups) {
     debug_reporter.Destroy(test_instance);
     vk::DestroyInstance(test_instance, nullptr);
 }
+
+#ifdef VK_USE_PLATFORM_METAL_EXT
+TEST_F(VkPositiveLayerTest, ExportMetalObjects) {
```
juan-lunarg, line 781: This test will currently fail to pass. But that's fine for now.

layers/generated/layer_chassis_dispatch.cpp
juan-lunarg: I'll handle this in another PR. I'm trying to figure out a good way to handle this robustly.

tests/positive/other.cpp
juan-lunarg: In my next PR I'll add more to the ExportMetalObjects test

mark-lunarg (changes requested):
As an outside observer it looks like this is two unrelated issues?  Let's split this into separate PRs, one for the MacOS build fix (prolly GTG as-is) and another for the ExportMetalObjects stuff, which we can leave WIP until that's covered.

PR #4876, opened 2022-11-28
mimalloc test

jeremyg-lunarg:

PR #4879, opened 2022-11-29
build: Make USE_ROBIN_HOOD_HASHING more sensible

jeremyg-lunarg:
Make robin_hood a non-REQUIRED package and use it if found.


jeremyg-lunarg:
@juan-lunarg is there a better way than this to deal with optional dependencies?  I'm about to add another one in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4876

CMakeLists.txt
```diff
@@ -220,9 +220,11 @@ endif()
 option(INSTALL_TESTS "Install tests" OFF)
 option(BUILD_LAYERS "Build layers" ON)
 option(BUILD_LAYER_SUPPORT_FILES "Generate layer files" OFF) # For generating files when not building layers
-option(USE_ROBIN_HOOD_HASHING "Use robin-hood-hashing" ON)
+
+find_package(robin_hood CONFIG)
+option(USE_ROBIN_HOOD_HASHING "Use robin-hood-hashing" ${robin_hood_FOUND})
 if (USE_ROBIN_HOOD_HASHING)
```
jeremyg-lunarg, line 226: I'm also going to change the help text to something like:
`"robin_hood provides faster versions of std::unordered_map and std::unordered_set"`

CMakeLists.txt
```diff
@@ -220,9 +220,11 @@ endif()
 option(INSTALL_TESTS "Install tests" OFF)
 option(BUILD_LAYERS "Build layers" ON)
 option(BUILD_LAYER_SUPPORT_FILES "Generate layer files" OFF) # For generating files when not building layers
-option(USE_ROBIN_HOOD_HASHING "Use robin-hood-hashing" ON)
+
+find_package(robin_hood CONFIG)
+option(USE_ROBIN_HOOD_HASHING "Use robin-hood-hashing" ${robin_hood_FOUND})
```
juan-lunarg, line 225: Why is this approach better? Seems more error prone.

Say I am a user who sets `USE_ROBIN_HOOD_HASHING` to `ON` but `find_package` fails. I'll get compilation failures instead of an early CMake failure.

Also we have users who build and package VVL themselves (arch folks). This makes it easier for them to not include robin_hood at all since it's not required anymore. Which could result in distros having a slower version of VVL

jeremyg-lunarg: The idea is that if you don't set USE_ROBIN_HOOD_HASHING to anything, it only turns on if find_package() finds it.
- For standalone update_deps.py + helper.cmake will cause it be found and ON, unless you explicitly set it to off.
- For package builds that expect everything to come from system dirs, find_package() would fail if they don't have it installed and they just wouldn't use it. This makes it more likely that they get working builds w/o worrying about USE_ROBIN_HOOD_HASHING

jeremyg-lunarg: But.. worrying about packagers is a good point:
- arch uses update_deps.py https://github.com/archlinux/svntogit-packages/blob/packages/vulkan-validation-layers/trunk/PKGBUILD
- fedora turns it off https://src.fedoraproject.org/rpms/vulkan-validation-layers/blob/f36/f/vulkan-validation-layers.spec
- debian turns it off https://salsa.debian.org/xorg-team/vulkan/vulkan-validationlayers/-/blob/debian-unstable/debian/rules

CMakeLists.txt
```diff
@@ -238,9 +238,11 @@ endif()
 option(INSTALL_TESTS "Install tests" OFF)
 option(BUILD_LAYERS "Build layers" ON)
 option(BUILD_LAYER_SUPPORT_FILES "Generate layer files" OFF) # For generating files when not building layers
-option(USE_ROBIN_HOOD_HASHING "Use robin-hood-hashing" ON)
+
+find_package(robin_hood CONFIG)
+option(USE_ROBIN_HOOD_HASHING "robin_hood provides faster versions of std::unordered_map and std::unordered_set" ${robin_hood_FOUND})
 if (USE_ROBIN_HOOD_HASHING)
-    find_package(robin_hood REQUIRED CONFIG)
+    add_definitions(-DUSE_ROBIN_HOOD_HASHING)
```
juan-lunarg, line 245: `add_definitions` isn't recommended anymore with modern CMake. Use `target_compile_definitions`

```suggestion
    target_compile_definitions(robin_hood::robin_hood INTERFACE USE_ROBIN_HOOD_HASHING)
```

This will apply the macro to any libraries that link against `robin_hood::robin_hood`

CMakeLists.txt
```diff
@@ -238,9 +238,11 @@ endif()
 option(INSTALL_TESTS "Install tests" OFF)
 option(BUILD_LAYERS "Build layers" ON)
 option(BUILD_LAYER_SUPPORT_FILES "Generate layer files" OFF) # For generating files when not building layers
-option(USE_ROBIN_HOOD_HASHING "Use robin-hood-hashing" ON)
+
+find_package(robin_hood CONFIG)
```
juan-lunarg, line 242: Move this find_package call to `layer/CMakeLists.txt` testing doesn't need/use robinhood.

layers/CMakeLists.txt
```diff
@@ -62,7 +62,6 @@ target_include_directories(VkLayer_utils PUBLIC
 
 if (USE_ROBIN_HOOD_HASHING)
     target_link_libraries(VkLayer_utils PUBLIC robin_hood::robin_hood)
```
juan-lunarg, line 64:
```suggestion
    target_link_libraries(VkLayer_utils PRIVATE robin_hood::robin_hood)
```

juan-lunarg:
I've been thinking about this for a while. And I have a couple thoughts.

- In the future should we prefer libraries that are provided by system package managers? [Like lib-format](https://github.com/fmtlib/fmt) seems the trouble arises when using less popular libraries.

- Is there an environment variable the SDK build scripts set so we can add extra error checking? In this case ensure that robin_hood is found. (cc @johnzupin / @richard-lunarg )

EX:
```cmake
if (DEFINED ENV{VULKAN_SDK_SCRIPT})
    # Make robin_hood required ...
endif()
```

jeremyg-lunarg:
> ```cmake
> if (DEFINED ENV{VULKAN_SDK_SCRIPT})
>     # Make robin_hood required ...
> endif()
> ```

Can't the SDK already do this (including with this PR merged) by explicitly setting -DUSE_ROBIN_HOOD_HASHING=ON? Then cmake will try to enable it for the build, which will fail if it isn't found.

CMakeLists.txt
jeremyg-lunarg: This doesn't seem to work, because tests use vk_layer_data.h, which includes robin_hood.h

jeremyg-lunarg: nevermind, forgot to remove something else

juan-lunarg: Grepping the test directory it looks like the only file using `vk_layer_data.h` is `tests/layers/device_profile_api.cpp`. Also I tried moving the find_package call and it worked without issue. What error are you seeing?

juan-lunarg:
> Can't the SDK already do this (including with this PR merged) by explicitly setting -DUSE_ROBIN_HOOD_HASHING=ON?

True because then they will see a compiler error. I was hoping for a more generic solution. For example ensuring `ANNOTATED_SPEC_LINK` is defined during SDK builds

CMakeLists.txt
jeremyg-lunarg: I have it working now. I hadn't yet put in target_compile_definitions() correctly.

layers/shader_instruction.h
```diff
@@ -16,13 +16,13 @@
...
-#ifndef VULKAN_SHADER_INSTRUCTION_H
-#define VULKAN_SHADER_INSTRUCTION_H
+#pragma once
 
-#include <stddef.h>  // size_t for gcc
-#include <stdint.h>
+#include <cassert>
+#include <cstddef>
+#include <cstdint>
+#include <string>
```
jeremyg-lunarg, line 24: This file was relying on picking up <string> from robin_hood.h via vk_layer_data.h

jeremyg-lunarg:
> > Can't the SDK already do this (including with this PR merged) by explicitly setting -DUSE_ROBIN_HOOD_HASHING=ON?
> 
> True because then they will see a compiler error. I was hoping for a more generic solution. For example ensuring `ANNOTATED_SPEC_LINK` is defined during SDK builds

Make an SDK preset?

layers/shader_instruction.h
juan-lunarg: Using `PRIVATE` tends to catch these implicit dependencies :)

`PRIVATE` is the `const` if CMake land.

juan-lunarg (approved):
LGTM

juan-lunarg:
> Make an SDK preset?

I think I have an idea that should work well. For now this LGTM and should help the package managers.

PR #4852, opened 2022-11-23
Improve scripts to use C++ features

spencer-lunarg:
better use of `std::array` and also `NULL` -> `nullptr`

PR #4880, opened 2022-11-30
layers: Use new C++ 17 [[maybe_unused]]

spencer-lunarg:
in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4852 failing CI because of unused variables, before using the `[[maybe_unused]]` there, want to just update it everywhere else first and make sure CI is happy with it


juan-lunarg (approved):
:+1:

PR #4882, opened 2022-11-30
Add draw time Unknown format texel count

spencer-lunarg:
Adds `VUID-vkCmdDispatch-None-04115` and `VUID-vkCmdDispatch-OpImageWrite-04469`

These are the draw time version of `VUID-RuntimeSpirv-OpImageWrite-07112`

It maps the `OpImageWrite` -> `OpLoad`/`OpAccessChain` -> Descriptor `OpVariable` . From there the descriptor can see all the `OpImageWrite` used on it.

------

I think there is improvement in how we are having to copy the data from `InterfaceVariable` to `DescriptorRequirement`, but realize that becomes messy due to `UPDATE_AFTER_BIND` descriptors being a thing. I decided to defer that to a follow-up PR

### DIFF
--- a/layers/core_validation.h
+++ b/layers/core_validation.h
@@ -114,6 +114,8 @@ struct DrawDispatchVuid {
     const char* storage_image_write_without_format = kVUIDUndefined;
     const char* storage_texel_buffer_read_without_format = kVUIDUndefined;
     const char* storage_texel_buffer_write_without_format = kVUIDUndefined;
+    const char* storage_image_write_texel_count = kVUIDUndefined;
+    const char* storage_texel_buffer_write_texel_count = kVUIDUndefined;
     const char* depth_compare_sample = kVUIDUndefined;
     const char* depth_read_only = kVUIDUndefined;
     const char* stencil_read_only = kVUIDUndefined;
@@ -595,9 +597,8 @@ class CoreChecks : public ValidationStateTracker {
 
     // helper for the common parts of ImageSamplerDescriptor and SamplerDescriptor validation
     bool ValidateSamplerDescriptor(const char* caller, const DrawDispatchVuid& vuids, const CMD_BUFFER_STATE* cb_node,
-                                   const cvdescriptorset::DescriptorSet* descriptor_set,
-                                   const std::pair<const uint32_t, DescriptorRequirement>& binding_info, uint32_t index,
-                                   VkSampler sampler, bool is_immutable, const SAMPLER_STATE* sampler_state) const;
+                                   const cvdescriptorset::DescriptorSet* descriptor_set, const DescriptorBindingInfo& binding_info,
+                                   uint32_t index, VkSampler sampler, bool is_immutable, const SAMPLER_STATE* sampler_state) const;
 
     // Validate contents of a CopyUpdate
     using DescriptorSet = cvdescriptorset::DescriptorSet;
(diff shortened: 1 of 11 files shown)


mark-lunarg (approved):

layers/core_validation.h
```diff
@@ -592,9 +594,8 @@ class CoreChecks : public ValidationStateTracker {
 
     // helper for the common parts of ImageSamplerDescriptor and SamplerDescriptor validation
     bool ValidateSamplerDescriptor(const char* caller, const DrawDispatchVuid& vuids, const CMD_BUFFER_STATE* cb_node,
-                                   const cvdescriptorset::DescriptorSet* descriptor_set,
-                                   const std::pair<const uint32_t, DescriptorRequirement>& binding_info, uint32_t index,
-                                   VkSampler sampler, bool is_immutable, const SAMPLER_STATE* sampler_state) const;
+                                   const cvdescriptorset::DescriptorSet* descriptor_set, const DescriptorBindingInfo& binding_info,
```
jeremyg-lunarg, line 597: can descriptor_set be nullptr? If not this could be a good time to convert it to a reference. Same for sampler_state.

jeremyg-lunarg: also passing in both sampler and sampler_state seems very redundant to me.

layers/pipeline_state.h
```diff
@@ -89,13 +89,16 @@ struct DescriptorRequirement {
     bool is_writable;
     // Copy from StageState.InterfaceVariable. It combines from plural shader stages. The index of array is index of image.
     std::vector<layer_data::unordered_set<SamplerUsedByImage>> samplers_used_by_image;
+    // For storage images - list of < OpImageWrite : Texel component length >
```
jeremyg-lunarg, line 92: These comments explaining the maps are very helpful.

jeremyg-lunarg (commented):
This LGTM but it will slightly conflict with the VK_EXT_descriptor_buffer PR. I think at this point it would be good not to churn that one again and rebase this instead.

layers/core_validation.h
spencer-lunarg: > If not this could be a good time to convert it to a reference

yep, already planned to go through the code and make a "Make pointers into const ref" change

layers/shader_module.cpp
```diff
@@ -1732,6 +1747,22 @@ uint32_t SHADER_MODULE_STATE::GetTypeId(uint32_t id) const {
     return type ? type->Word(type->TypeId()) : 0;
 }
 
+// Return zero if nothing is found
+uint32_t SHADER_MODULE_STATE::GetTexelComponentCount(const Instruction& insn) const {
+    uint32_t texel_component_count = 0;
+    switch (insn.Opcode()) {
+        case spv::OpImageWrite: {
+            const Instruction* texel_def = FindDef(insn.Word(3));
+            const Instruction* texel_type = FindDef(texel_def->Word(1));
```
arno-lunarg, line 1756: Would an `assert` on those 2 pointers make sense?

jeremyg-lunarg (approved):
LGTM once rebased through the conflict

Tony-LunarG (approved):
LGTM - one typo

layers/shader_module.cpp
spencer-lunarg: valid question, but Nate and I discussed this a while ago, it becomes super easy to start to put `assert` at every case of `FindDef`. I haven't had time to go through and clean them up and spell out the new rules but for this case, if these pointers are ever null, then `spirv-val` failed it job badly

Issue #4589, opened 2022-09-23
vkCmdBuildAccelerationStructuresKHR update mode crashes the validation layer if srcAccelerationStructure=VK_NULL_HANDLE


gurkiapplied:
**Describe the Issue**
The layer crashes in `CoreChecks::PreCallValidateCmdBuildAccelerationStructuresKHR()` when `pInfos->mode == VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR` and `srcAccelerationStructure == VK_NULL_HANDLE`. These params are not valid as per vulkan spec and the layer should display a validation error without crashing.

**Valid Usage ID**
Related VUIDs:
VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03708

**Environment:**
 - OS: Linux
 - GPU: NVIDIA RTX 3090
 - SDK or header version if building from repo: master
 - Options enabled (synchronization, best practices, etc.):

**Additional context**

ncesario-lunarg:
I think something like the following could fix this:
```diff
diff --git a/layers/core_validation.cpp b/layers/core_validation.cpp
index 0f8c0dd68..14c03d771 100644
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -9899,8 +9899,16 @@ bool CoreChecks::PreCallValidateCmdBuildAccelerationStructuresKHR(
             skip |= ValidateMemoryIsBoundToBuffer(dst_as_state->buffer_state.get(), "vkCmdBuildAccelerationStructuresKHR",
                                                   "VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03707");
             if (pInfos[info_index].mode == VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR) {
-                skip |= ValidateMemoryIsBoundToBuffer(src_as_state->buffer_state.get(), "vkCmdBuildAccelerationStructuresKHR",
-                                                      "VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03708");
+                if (src_as_state->buffer_state) {
+                    skip |= ValidateMemoryIsBoundToBuffer(src_as_state->buffer_state.get(), "vkCmdBuildAccelerationStructuresKHR",
+                                                          "VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03708");
+                } else {
+                    LogObjectList objs;
+                    objs.add(device);
[...]

PR #4883, opened 2022-11-30
Fix for PreCallValidateCmdBuildAccelerationStructuresKHR nullptr handling + new VUID

arno-lunarg:
- Fix for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4589

Adds VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-04630

Also some cleanups in ray tracing tests:
- Make `AccelerationStructure{KHR}::create_scratch_buffer` return a buffer
- Use `device()` instead of `m_device->handle()`
- Remove unhelpful usage of `std::make_unique`

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -10472,47 +10472,75 @@ bool CoreChecks::PreCallValidateCmdBuildAccelerationStructuresKHR(
         for (uint32_t info_index = 0; info_index < infoCount; ++info_index) {
             auto src_as_state = Get<ACCELERATION_STRUCTURE_STATE_KHR>(pInfos[info_index].srcAccelerationStructure);
             auto dst_as_state = Get<ACCELERATION_STRUCTURE_STATE_KHR>(pInfos[info_index].dstAccelerationStructure);
-            skip |= ValidateMemoryIsBoundToBuffer(dst_as_state->buffer_state.get(), "vkCmdBuildAccelerationStructuresKHR",
-                                                  "VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03707");
+
+            if (dst_as_state != nullptr) {
+                skip |= ValidateMemoryIsBoundToBuffer(dst_as_state->buffer_state.get(), "vkCmdBuildAccelerationStructuresKHR",
+                                                      "VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03707");
+            }
+
             if (pInfos[info_index].mode == VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR) {
-                skip |= ValidateMemoryIsBoundToBuffer(src_as_state->buffer_state.get(), "vkCmdBuildAccelerationStructuresKHR",
-                                                      "VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03708");
-                if (src_as_state == nullptr || !src_as_state->built ||
-                    !(src_as_state->build_info_khr.flags & VK_BUILD_ACCELERATION_STRUCTURE_ALLOW_UPDATE_BIT_KHR)) {
-                    skip |= LogError(device, "VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03667",
+                if (pInfos[info_index].srcAccelerationStructure == VK_NULL_HANDLE) {
+                    LogObjectList objs(device, commandBuffer);
+                    skip |= LogError(objs, "VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-04630",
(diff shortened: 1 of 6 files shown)


layers/core_validation.cpp
```diff
@@ -10472,47 +10472,71 @@ bool CoreChecks::PreCallValidateCmdBuildAccelerationStructuresKHR(
...
             auto dst_as_state = Get<ACCELERATION_STRUCTURE_STATE_KHR>(pInfos[info_index].dstAccelerationStructure);
-            skip |= ValidateMemoryIsBoundToBuffer(dst_as_state->buffer_state.get(), "vkCmdBuildAccelerationStructuresKHR",
-                                                  "VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03707");
+
+            if (dst_as_state != nullptr) {
+                skip |= ValidateMemoryIsBoundToBuffer(dst_as_state->buffer_state.get(), "vkCmdBuildAccelerationStructuresKHR",
+                                                      "VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03707");
+            }
+
+            LogObjectList objs(device, commandBuffer);
```
spencer-lunarg, line 10481: in theory, the majority of time there are no errors/warning and Validation Performance is an ongoing issue... please move this to above the `LogError` of each call

spencer-lunarg (approved):
LGTM besides single comment I left

layers/core_validation.cpp
arno-lunarg: `LogObjectList` uses a `small_vector<VulkanTypedHandle, 4, uint32_t>` to store its element. Here, at most 3 elements will be stored, so no dynamic allocation, everything is on the stack. So I do not think there is a perceptible perf loss with the current code?

spencer-lunarg: ok that is fair and this is all speculation without any real testing.

I guess I was thinking also we do this everywhere else so keep it consistent. Also I like it by the `LogError` because in the future when someone ends up copy/pasting a `LogError` code snippet, I want them to try and think explicitly what objects should be logged and not blindly using what was above it

arno-lunarg:
Investigating a crash on AMD with the new test

layers/vk_layer_logging.h
```diff
@@ -102,6 +102,11 @@ struct LogObjectList {
         add(object);
     }
 
+    template <typename... HANDLE_T>
+    LogObjectList(HANDLE_T... objects) {
```
jeremyg-lunarg, line 106: Nice!

jeremyg-lunarg (approved):
LGTM

tests/vklayertests_ray_tracing.cpp
```diff
@@ -1492,7 +1484,105 @@ TEST_F(VkLayerTest, RayTracingCmdCopyAccelerationStructureToMemoryKHR) {
...
+    ASSERT_NO_FATAL_FAILURE(InitState(nullptr, &features2));
+
+    auto vkCmdBuildAccelerationStructuresKHR =
+        GetDeviceProcAddr<PFN_vkCmdBuildAccelerationStructuresKHR>("vkCmdBuildAccelerationStructuresKHR");
+    auto vkGetBufferDeviceAddressKHR = GetDeviceProcAddr<PFN_vkGetBufferDeviceAddress>("vkGetBufferDeviceAddressKHR");
+
+    // Create bottom level acceleration structure buffer
+    VkBufferObj as_buffer;
+    as_buffer.init(*m_device, 4096, VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT, VK_BUFFER_USAGE_ACCELERATION_STRUCTURE_STORAGE_BIT_KHR);
+    std::cout << "as_buffer created" << std::endl;
```
Tony-LunarG, line 1523: Good for debugging, but do we want this printed every time the test is run?  Most other tests don't print status in this detail. Same for the prints below

Tony-LunarG (approved):
LGTM - Question about logging in test

tests/vklayertests_ray_tracing.cpp
jeremyg-lunarg: yeah take this out. We should start using SCOPED_TRACE() or something similar that can print diagnostics like this only when tests fail.

arno-lunarg: The test is crashing right now on AMD in the CI, this is just to help me figure things out. I am going to remove all of this!

arno-lunarg: The issue was disabled extensions, switching to `InitFrameworkForRayTracingTest` enables the correct ones. I removed the debug logs.

arno-lunarg:
Failing test is `VkPositiveLayerTest.GetTimelineSemThreadRace` on linux + NV

PR #4887, opened 2022-11-30
Add GPU-AV and debug printf allocations for EXT mesh_shader

Tony-LunarG:
I didn't add the entrypoints to state_tracker, or calls to those entrypoints from GPU-AV or DbPf because I don't think anything is checking them, but I could be wrong and will happily add them if so.

### DIFF
--- a/layers/gpu_validation.cpp
+++ b/layers/gpu_validation.cpp
@@ -1630,6 +1630,22 @@ void GpuAssisted::PreCallRecordCmdDrawMeshTasksIndirectCountNV(VkCommandBuffer c
     AllocateValidationResources(commandBuffer, VK_PIPELINE_BIND_POINT_GRAPHICS, CMD_DRAWMESHTASKSINDIRECTCOUNTNV);
 }
 
+void GpuAssisted::PreCallRecordCmdDrawMeshTasksEXT(VkCommandBuffer commandBuffer, uint32_t groupCountX, uint32_t groupCountY,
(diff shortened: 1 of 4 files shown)


layers/gpu_validation.cpp
```diff
@@ -1630,6 +1630,22 @@ void GpuAssisted::PreCallRecordCmdDrawMeshTasksIndirectCountNV(VkCommandBuffer c
     AllocateValidationResources(commandBuffer, VK_PIPELINE_BIND_POINT_GRAPHICS, CMD_DRAWMESHTASKSINDIRECTCOUNTNV);
 }
 
+void GpuAssisted::PreCallRecordCmdDrawMeshTasksEXT(VkCommandBuffer commandBuffer, uint32_t groupCountX, uint32_t groupCountY,
+                                                   uint32_t groupCountZ) {
+    AllocateValidationResources(commandBuffer, VK_PIPELINE_BIND_POINT_GRAPHICS, CMD_DRAWMESHTASKSNV);
```
Tony-LunarG, line 1635: Oops - wrong CMD_*

Tony-LunarG: Fixed

jeremyg-lunarg (approved):
LGTM. Odds of state tracker ever doing PreCallRecord for these seems low to me.

spencer-lunarg (approved):

PR #4888, opened 2022-11-30
ci: Build against latest MacOS SDK

juan-lunarg:
Prevents build failures against newer MacOS SDKs

### DIFF
--- a/.github/workflows/build_macos.yml
+++ b/.github/workflows/build_macos.yml
@@ -22,7 +22,7 @@ on:
             - master
 
 jobs:
-  macos:
+  macos-min-sdk:
     runs-on: macos-latest
     strategy:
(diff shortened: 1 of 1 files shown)


juan-lunarg:
Having this is in CI would have prevented https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4886

spencer-lunarg (approved):

jeremyg-lunarg (approved):
LGTM. It would be nice if the commit message and issue were more clear that this is about the macos SDK and not the vulkan SDK.

juan-lunarg:
> LGTM. It would be nice if the commit message and issue were more clear that this is about the macos SDK and not the vulkan SDK.

Updated commit message

PR #4889, opened 2022-12-01
layers: Track input assembly state at pre_raster_state

zzoon:
Currently the validation layer can't judge if 00736 is being kept correctly because pre_raster_state doesn't have the input assembly state.

### DIFF
--- a/layers/pipeline_state.h
+++ b/layers/pipeline_state.h
@@ -393,6 +393,8 @@ class PIPELINE_STATE : public BASE_NODE {
     const safe_VkPipelineInputAssemblyStateCreateInfo *InputAssemblyState() const {
         if (vertex_input_state) {
             return vertex_input_state->input_assembly_state;
+        } else if (pre_raster_state) {
+            return pre_raster_state->input_assembly_state;
         }
         return nullptr;
     }
--- a/layers/pipeline_sub_state.cpp
+++ b/layers/pipeline_sub_state.cpp
@@ -63,6 +63,7 @@ PreRasterState::PreRasterState(const PIPELINE_STATE &p, const ValidationStateTra
 
     rp_state = dev_data.Get<RENDER_PASS_STATE>(create_info.renderPass);
 
+    input_assembly_state = create_info.pInputAssemblyState;
     raster_state = create_info.pRasterizationState;
 
     tess_create_info = create_info.pTessellationState;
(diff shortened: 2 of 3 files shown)


spencer-lunarg:
For record keeping - This is an internal issue https://gitlab.khronos.org/vulkan/vulkan/-/issues/3300

spencer-lunarg (approved):
This looks good to me, but before merging is there a CTS tests or something related that this fixes?

zzoon:
> This looks good to me, but before merging is there a CTS tests or something related that this fixes?
Yeah. https://gitlab.khronos.org/Tracker/vk-gl-cts/-/issues/4146 

> For record keeping - This is an internal issue https://gitlab.khronos.org/vulkan/vulkan/-/issues/3300
Thanks. I didn't notice when I make this PR.
We probably need to wait for the conclusion of the issue?

ncesario-lunarg (changes requested):
This does not seem correct. My understanding is that each sub-state is disjoint from every other sub-state. If 00736 requires pre-raster _and_ vertex-input state, then we should be checking that each of those states are present. It may be the case that check isn't being made for a linked pipeline, and that would need to be fixed differently. We also need a test for 00736 when using graphics pipeline libraries (I don't think one currently exists).

If you supply `pInputAssemblyState` when creating a pre-raster library, I think it should be ignored (or maybe an error?). Is that correct @Tobski?

ncesario-lunarg:
I think a simple fix for this is to update the 00736 condition to:
```cpp
if (has_control && has_eval && pipeline.vertex_input_state && pipeline.pre_raster_state && (!ia_state || ia_state->topology != VK_PRIMITIVE_TOPOLOGY_PATCH_LIST)) {
```
I'll create a test to try this out.

ncesario-lunarg:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4936 should address the problem and adds a test that reproduced the original issue.

ncesario-lunarg:
@zzoon Would you be able to try out #4936 and see if that resolves your issue?

zzoon:
> @zzoon Would you be able to try out #4936 and see if that resolves your issue?

Sorry about late reply, I got back from holidays.

Yeah I think your approach makes sense. 

Closing this issue.

PR #4890, opened 2022-12-01
Turn PIPELINE_STATE and CMD_BUFFER_STATE into const refs

spencer-lunarg:
Both `PIPELINE_STATE` and `CMD_BUFFER_STATE` were unnecessarily being passed around as pointers. This change while large, goes through and both unify the names of the variables and passes the state all around as a const ref


layers/state_tracker.h
```diff
@@ -482,9 +482,9 @@ class ValidationStateTracker : public ValidationObject {
         set_image_view_initial_layout_callback.reset(new SetImageViewInitialLayoutCallback(std::forward<Fn>(fn)));
     }
 
-    void CallSetImageViewInitialLayoutCallback(CMD_BUFFER_STATE* cb_node, const IMAGE_VIEW_STATE& iv_state, VkImageLayout layout) {
+    void CallSetImageViewInitialLayoutCallback(CMD_BUFFER_STATE* cb_state, const IMAGE_VIEW_STATE& iv_state, VkImageLayout layout) {
         if (set_image_view_initial_layout_callback) {
-            (*set_image_view_initial_layout_callback)(cb_node, iv_state, layout);
+            (*set_image_view_initial_layout_callback)(cb_state, iv_state, layout);
```
jeremyg-lunarg, line 487: I'm guessing this is why the CMD_BUFFER_STATE's cannot be de-pointered in the image layout paths. The correct way to clean this up is to move functionality to CORE_CMD_BUFFER_STATE virtual methods.

layers/descriptor_validation.cpp
```diff
@@ -181,7 +181,7 @@ bool CoreChecks::PreCallValidateCmdBindDescriptorSets(VkCommandBuffer commandBuf
     auto cb_state = GetRead<CMD_BUFFER_STATE>(commandBuffer);
     assert(cb_state);
     bool skip = false;
-    skip |= ValidateCmd(cb_state.get(), CMD_BINDDESCRIPTORSETS);
+    skip |= ValidateCmd(*cb_state.get(), CMD_BINDDESCRIPTORSETS);
```
jeremyg-lunarg, line 184: this can be just *cb_state instead of *cb_state.get()

layers/buffer_validation.cpp
```diff
@@ -2375,24 +2372,25 @@ bool CoreChecks::PreCallValidateCmdClearColorImage(VkCommandBuffer commandBuffer
                                                    const VkImageSubresourceRange *pRanges) const {
     bool skip = false;
     // TODO : Verify memory is in VK_IMAGE_STATE_CLEAR state
-    auto cb_node = GetRead<CMD_BUFFER_STATE>(commandBuffer);
+    auto cb_state_ptr = GetRead<CMD_BUFFER_STATE>(commandBuffer);
     auto image_state = Get<IMAGE_STATE>(image);
-    if (cb_node && image_state) {
+    if (cb_state_ptr && image_state) {
+        const CMD_BUFFER_STATE &cb_state = *cb_state_ptr.get();
```
jeremyg-lunarg, line 2378: This can be *cb_state_ptr;  I've found 2 of these, probably there are more.

layers/core_validation.cpp
```diff
@@ -5007,25 +5003,29 @@ bool CoreChecks::ValidatePrimaryCommandBufferState(
...
-        skip |= LogError(pCB->commandBuffer(), vuid, "%s Command buffer %s must be allocated with VK_COMMAND_BUFFER_LEVEL_PRIMARY.",
-                         loc.Message().c_str(), report_data->FormatHandle(pCB->commandBuffer()).c_str());
+        skip |=
+            LogError(cb_state.commandBuffer(), vuid, "%s Command buffer %s must be allocated with VK_COMMAND_BUFFER_LEVEL_PRIMARY.",
+                     loc.Message().c_str(), report_data->FormatHandle(cb_state.commandBuffer()).c_str());
     } else {
-        for (const auto *sub_cb : pCB->linkedCommandBuffers) {
-            skip |= ValidateQueuedQFOTransfers(sub_cb, qfo_image_scoreboards, qfo_buffer_scoreboards);
+        for (const auto *sub_cb : cb_state.linkedCommandBuffers) {
+            if (!sub_cb) {
```
jeremyg-lunarg, line 5013: can this really happen? It seems like if we're inserting nullptr into this unordered_set, that's a bug.

layers/core_validation.cpp
```diff
@@ -5580,10 +5580,8 @@ struct CommandBufferSubmitState {
                 function += CommandTypeString(cmd_info.cmd_type);
                 CoreChecks::DescriptorContext context{function.c_str(),
                                                       core->GetDrawDispatchVuid(cmd_info.cmd_type),
-                                                      &cb_node,
-                                                      set_node.get(),
-                                                      cmd_info.attachments.get(),
-                                                      cmd_info.subpasses.get(),
+                                                      cb_state,
+                                                      *set_node.get(),
```
jeremyg-lunarg, line 5584: *set_node,

mark-lunarg (approved):
I like it a lot.

layers/core_validation.cpp
```diff
@@ -10108,18 +10118,19 @@ bool CoreChecks::PreCallValidateCmdBindPipeline(VkCommandBuffer commandBuffer, V
...
         std::make_pair(VK_PIPELINE_BIND_POINT_COMPUTE, "VUID-vkCmdBindPipeline-pipelineBindPoint-00778"),
         std::make_pair(VK_PIPELINE_BIND_POINT_RAY_TRACING_KHR, "VUID-vkCmdBindPipeline-pipelineBindPoint-02391")};
 
     skip |= ValidatePipelineBindPoint(cb_state.get(), pipelineBindPoint, "vkCmdBindPipeline()", bindpoint_errors);
 
-    auto pipeline_state = Get<PIPELINE_STATE>(pipeline);
-    assert(pipeline_state);
+    auto pPipeline = Get<PIPELINE_STATE>(pipeline);
+    assert(pPipeline);
+    const PIPELINE_STATE &pipeline_state = *pPipeline.get();
```
jeremyg-lunarg, line 10131: should this pPipeline get cleaned up? (otherwise it will be lonely). also extra get().

jeremyg-lunarg (commented):
This is a great cleanup! Just a few problems found.

layers/state_tracker.h
jeremyg-lunarg: To be clear this is a separate PR thing.

layers/core_validation.cpp
spencer-lunarg: agree

PR #4896, opened 2022-12-02
layers: Relabel copy cmd VUID from 1.3.236

spencer-lunarg:
These VUIDs were shuffled around in the last 1.3.236 header update

`VUID-VkImageCopy-extent-00140` was not replaced correctly, will be making another internal spec fix for that


jeremyg-lunarg (approved):
LGTM

PR #4897, opened 2022-12-02
layers: Label 1.3.236 descriptor VUs

spencer-lunarg:
adds `VUID-vkCmdDrawIndexed-viewType-07752` and `VUID-vkCmdDrawIndexed-format-07753` in effort to reduce how many checks `vuids.descriptor_valid` has

### DIFF
--- a/layers/descriptor_validation.cpp
+++ b/layers/descriptor_validation.cpp
@@ -913,10 +913,7 @@ bool CoreChecks::ValidateDescriptor(const DescriptorContext &context, const Desc
         if (reqs & DESCRIPTOR_REQ_ALL_VIEW_TYPE_BITS) {
             if (~reqs & (1 << image_view_ci.viewType)) {
                 auto set = context.descriptor_set.GetSet();
-                auto vuid_text = enabled_features.descriptor_buffer_features.descriptorBuffer
-                                     ? context.vuids.descriptor_buffer_bit_set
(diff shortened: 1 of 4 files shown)


layers/descriptor_validation.cpp
```diff
@@ -915,10 +915,7 @@ bool CoreChecks::ValidateDescriptor(const DescriptorContext &context, const Desc
         if (reqs & DESCRIPTOR_REQ_ALL_VIEW_TYPE_BITS) {
             if (~reqs & (1 << image_view_ci.viewType)) {
                 auto set = context.descriptor_set->GetSet();
-                auto vuid_text = enabled_features.descriptor_buffer_features.descriptorBuffer
-                                     ? context.vuids.descriptor_buffer_bit_set
-                                     : context.vuids.descriptor_valid;
-                return LogError(set, vuid_text,
+                return LogError(set, "VUID-vkCmdDrawIndexed-viewType-07752",
```
jeremyg-lunarg, line 918: It looks like 07752 and 07752 are defined for every draw call variant. Shouldn't this be handled with an entry in the vuids struct like in the old code?

spencer-lunarg: ... ya, idk how I messed that up, will fix

ncesario-lunarg (approved):
LGTM!

PR #4900, opened 2022-12-02
Misc false positive and crash fixes when running against vkd3d-proton

HansKristian-Work:
Found some minor problems that I've fixed.

- Fix dynamic rendering with no attachment. Multisample state may be NULL.
- In descriptor buffer validation, only validate that bindings that are non-variable count are considered when checking for range. Fixes lots of false positive spam in vkd3d-proton.


jeremyg-lunarg:
@stu-s and/or @archimedus, can you please review the descriptor buffer portion of this?

stu-s:
LGTM

jeremyg-lunarg (approved):
LGTM. Thanks for the fix! And thanks for reviewing @stu-s

PR #4901, opened 2022-12-02
build, docs: try to remove repo-specific incantations

Keenuts:
This project uses cmake, and maybe it would be useful to remove friction? -> run cmake, it works.

By setting the UPDATE_DEPS as a enabled-by-default option, we can remove the manual call to update deps script.
Updated the doc to reflect that. Added Ninja as supported code generator since the default build instruction were already using Ninja.

Kept the windows part mostly unchanged, as Windows is a bit more specific, just added the VS env call in the short ones as ninja/cmake/msbuild might not be available on the default cmd by default.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,7 @@ project(VVL LANGUAGES CXX C)
 option(VVL_ENABLE_ASAN "Use address sanitization (specifically -fsanitize=address)" OFF)
 
 option(BUILD_TESTS "Build the tests" OFF)
+option(UPDATE_DEPS "Clone & build dependencies" ON)
 
 add_definitions(-DVK_ENABLE_BETA_EXTENSIONS) # Enable beta Vulkan extensions
 
--- a/BUILD.md
+++ b/BUILD.md
@@ -24,14 +24,13 @@ The following will be enough for most people, for platform-specific instructions
 git clone https://github.com/KhronosGroup/Vulkan-ValidationLayers.git
 cd Vulkan-ValidationLayers
 
-mkdir build
-cd build
-
 # Linux
-python3 ../scripts/update_deps.py --dir ../external --arch x64 --config debug
-cmake -G Ninja -C ../external/helper.cmake -DCMAKE_BUILD_TYPE=Debug ..
+cmake -G Ninja -B build -DCMAKE_BUILD_TYPE=Debug -DUPDATE_DEPS=TRUE
+ninja -C build
 
 # Windows
+mkdir build
+cd build
 python3 ..\scripts\update_deps.py --dir ..\external --arch x64 --config debug
 cmake -A x64 -C ..\external\helper.cmake -DCMAKE_BUILD_TYPE=Debug ..
 
--- a/BUILD.md
+++ b/BUILD.md
@@ -20,22 +20,38 @@
 
 The following will be enough for most people, for platform-specific instructions, see below.
 
+### Linux
+
 ```bash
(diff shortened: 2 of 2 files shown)


juan-lunarg:
> This project uses cmake, and maybe it would be useful to remove friction? -> run cmake, it works.

I agree it would reduce friction for developers of the project. But not necessarily for package developers who try to package the project for their own platform.

Fedora, Ubuntu, and Arch developers also try and package their own SDK aside from Lunarg. (Those are the ones I'm aware of)

It's what led to us disabling warnings by default (despite not being convenient for us):
https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4780

An attempt to make this easier was the addition of CMakePresets.json since it applies `UPDATE_DEPS=ON` and `BUILD_WERROR=ON` for users.

https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/BUILD.md#cmakepresetsjson-321

Keenuts:
Thanks for the feedback!

> It's what led to us disabling warnings by default (despite not being convenient for us):
Sure, not having -Werror by default makes sense, since maintainer toolchain might not be the most recent, or some warnings can be a surprise (like when the fallthrough warning was added). Having it enabled on the CI only could still provide the "PRs are clean" feature, without hindering people on other environments.

> But not necessarily for package developers who try to package the project for their own platform
You mean if Arch package maintainer wants to provide their own Vulkan-Header hash? Or that the update-deps step, which requires network would be done outside of the build-step, possibly without network (like openbsd maintainers likes to do it) ?

juan-lunarg:
> Having it enabled on the CI only could still provide the "PRs are clean" feature, without hindering people on other environments.

Correct warnings are treated as errors on CI.

> You mean if Arch package maintainer wants to provide their own Vulkan-Header hash? Or that the update-deps step, which requires network would be done outside of the build-step, possibly without network (like openbsd maintainers likes to do it) ?

As an example we can look at conan (they aren't a system package manager, but a general C/C++ package manager)

https://github.com/conan-io/conan-center-index/tree/master/recipes/vulkan-validationlayers

Currently because `UPDATE_DEPS` is disabled they find their dependencies via their package management system. Which is what they want for numerous reasons (applying patches to their builds, dependency tracking to get everything working with conan, etc). Enabling `UPDATE_DEPS` by default would throw off this workflow. And this workflow isn't specific to just conan.

The other problem with enabling `UPDATE_DEPS` by default is that it would be very inconsistent. The validation layers isn't the only repo that uses this script. Numerous Vulkan repos use this script and only having the validation layers enabling it by default would be confusing.

johnzupin:
> Or that the update-deps step, which requires network would be done outside of the build-step, possibly without network (like openbsd maintainers likes to do it) ?

This sounds like my experience with building Debian/Ubuntu packages, there isn't any network access with the build step. This could break some other package builders, not that it would be hard to fix, but would cause an interrupt. I agree with Juan about being consistent with the other repos if this is the desired behavior. Also, ensuring that the other repos have a flag to turn this off as well.

CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ project(VVL LANGUAGES CXX C)
 option(VVL_ENABLE_ASAN "Use address sanitization (specifically -fsanitize=address)" OFF)
 
 option(BUILD_TESTS "Build the tests" OFF)
+option(UPDATE_DEPS "Clone & build dependencies" ON)
```
ncesario-lunarg, line 25: FWIW, I would very much like to make this the default, but there are probably some gremlins to workout first :)

Issue #4884, opened 2022-11-30
Global variables might be used when invalid in some cases.


Keenuts:
Hi!

This issue was found when running validation layers with ubsan on our internal CI. We have a multi-threaded application and build validation layers differently (bazel, custom rules), but cannot share logs. (I know, horrible bug report :upside_down_face: )
What we observe is failure when creating vulkan instance because of some SIGILL during what I believe is global variable initialization by the dynamic linker.

After some debugging, I noticed global variables were declared in the generated files, and those globals were not trivially constructible, like `std::map`. This issue could be "solved" by changing the compilation order and some other magic incantations, but never fixed all cases.

However, changing:
```
const std::map<> something = { [...] }
```
to:
```
const std::map<>& getSomething() {
     static const std::map<> map = { [...] };
     return map;
}
```

Fixed it in all cases (we found other issues, but don't have concrete solution/more context yet).
This is not a perfect fix either, just masked the problem since at least this function-local static initialization order is defined.
An alternative would be to follow this: https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables and only have constexpr globals for those cases.

**Environment:**
 - OS: linux
 - GPU: n/a
 - seen at commit af45e1ee442f5071521f0fdf4dc33a866a383e75

Keenuts:
Looked into that a bit today, and I don't see a straightforward way to make it all constexpr right now:
 -> having those maps converted to constexpr arrays requires the contained items to be constexpr.
 -> base type can be std::bitset<128>, which allow some constexpr operation, but not enough to set high bits.
 -> extended constexpr support is coming, but cpp23, which is a bit too modern...

So in the meantime, I'd suggest the following:
 - convert map/unoderered_maps to arrays, which are trivially destructible
 - wrap those as function-local static, to make sure initialization is thread-safe.
 - since they are read-only consts, no need to lock on read.

Keenuts:
@juan-lunarg I see you mentioned working on the code generation more broadly on other issues. Do you plan on changing the actual code, or just tooling/generation process?
(to know if I should work on the more long-term solution, or stop after the band-aid PR in flight)

juan-lunarg:
I don't really have big plans for the current code generation. The only thing on my roadmap is currently the code generation is causing a bug for `vkExportMetalObjectsExt` when handle wrapping is enabled. Which I'm currently working on.

juan-lunarg:
[...]

PR #4905, opened 2022-12-02
layers: Force thread-safe global initialization

Keenuts:
Synchronization checks are relying on a set of global constants to map some flags to values. Those constants are not constexpr, and thus they can require a runtime initialization. This initialization is not thread-safe, and can cause issue in multithreaded environments.

Making those globals as function-local static enforces a static initialization.

This is not an optimal solution, a better option could be to make those constexpr, but AFAIK I cannot do that yet, since required std::bitset operations are only available starting cpp23.

Next step would be to make those variables trivially destructible, hence replacing those maps with arrays. Large maps have O(100) elements, so doing a linear sweep shouldn't be too bad on those.

related to #4884 but not fixing it yet (map -> array change)
Signed-off-by: Nathan Gauër <brioche@google.com>


jeremyg-lunarg:
> Next step would be to make those variables trivially destructible, hence replacing those maps with arrays. Large maps have O(100) elements, so doing a linear sweep shouldn't be too bad on those.

You could also use std::search() although for 100 elements or less it might not matter.

jeremyg-lunarg (approved):
LGTM. If you make this non-draft I will merge it. If you want to continue on with the follow on patch, that'd be great too. Thanks!

Keenuts:
> LGTM. If you make this non-draft I will merge it. If you want to continue on with the follow on patch, that'd be great too. Thanks!

Thanks!
Making it non-draft as it fixes some issues for now and could unblock our CI, will follow up next week with the rest :)

PR #4916, opened 2022-12-05
layers: Ensure ExportMetalObjects is valid

juan-lunarg:
Currently validation for vkExportMetalObjectsEXT is a mixture of code-gen / manual code. This static assert ensures the existing code is fine.


Tony-LunarG (approved):
LGTM

PR #4918, opened 2022-12-05
layers: Fix descriptor buffer VUIDs 08115 and 08117

### LINKED ISSUE #4913: VK_EXT_descriptor_buffer: false positive when mixing buffer/set binding models
`VK_EXT_descriptor_buffer` does not allow to mix descriptor buffer and descriptor set within a single draw call but it's valid to do this between draw calls/dispatches.

In my setup I render 3D object using `VK_EXT_descriptor_buffer` to manage descriptors, and after that ImGUI 2D does `VkDescriptorSet`-based rendering. This leads to this validation error:

```Validation Error: [ VUID-vkCmdDrawIndexed-None-08117 ] Object 0: handle = 0x2e2cd000000002b, type = VK_OBJECT_TYPE_PIPELINE; | MessageID = 0x9dd4889e | vkCmdDrawIndexed: If the descriptors used by the VkPipeline bound to the pipeline bind point were specified via vkCmdSetDescriptorBufferOffsetsEXT, the bound VkPipeline must have been created with VK_PIPELINE_CREATE_DESCRIPTOR_BUFFER_BIT_EXT The Vulkan spec states: If the descriptors used by the VkPipeline bound to the pipeline bind point were specified via vkCmdSetDescriptorBufferOffsetsEXT, the bound VkPipeline must have been created with VK_PIPELINE_CREATE_DESCRIPTOR_BUFFER_BIT_EXT (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdDrawIndexed-None-08117)```

Environment:
* Window 10
* NVidia 2080
* Vulkan Beta Driver 517.83
* VK_HEADER_VERSION 236
* Valiation layers built from source from December 3, d5824670d4d0d52b3749632ad8913a46ad23a3f6

I have a sample app that can be built with VS 2022 (earlier versions should also work but platform toolset should be updated in VS project settings): https://github.com/kennyalive/vulkan-base/tree/VK_EXT_descriptor_buffer
After repo is cloned please switch to VK_EXT_descriptor_buffer branch.
Running with validation enabled (--validation-layers on the command line) demonstrates the issue.

With validation disabled the app works as expected (renders both 3D model and ImGui UI).
Also by disabling ImGui, the 3D model is rendered without validation errors, it looks the VK_EXT_descriptor_buffer setup is correct and the problem is with mixing different approaches.

### PR DESCRIPTION
**Pull request by jeremyg-lunarg**
These 2 VUIDs are about which pipelines are compatible with vkCmdBindDescriptorSets() or vkCmdSetDescriptorBufferOffsetsEXT(). The old code was looking at state set by vkCmdBindDescriptorBuffersEXT(). It should be valid to have descriptor buffers bound but not used to allow switching back and forth between pipelines that use descriptor sets and descriptor buffers.

Fixes #4913

### DIFF
--- a/layers/cmd_buffer_state.cpp
+++ b/layers/cmd_buffer_state.cpp
@@ -808,17 +808,17 @@ void CMD_BUFFER_STATE::ExecuteCommands(uint32_t commandBuffersCount, const VkCom
     }
 }
 
-void CMD_BUFFER_STATE::PushDescriptorSetState(VkPipelineBindPoint pipelineBindPoint, PIPELINE_LAYOUT_STATE *pipeline_layout,
+void CMD_BUFFER_STATE::PushDescriptorSetState(VkPipelineBindPoint pipelineBindPoint, const PIPELINE_LAYOUT_STATE &pipeline_layout,
                                               uint32_t set, uint32_t descriptorWriteCount,
                                               const VkWriteDescriptorSet *pDescriptorWrites) {
     // Short circuit invalid updates
-    if (!pipeline_layout || (set >= pipeline_layout->set_layouts.size()) || !pipeline_layout->set_layouts[set] ||
-        !pipeline_layout->set_layouts[set]->IsPushDescriptor()) {
+    if ((set >= pipeline_layout.set_layouts.size()) || !pipeline_layout.set_layouts[set] ||
+        !pipeline_layout.set_layouts[set]->IsPushDescriptor()) {
         return;
     }
 
     // We need a descriptor set to update the bindings with, compatible with the passed layout
-    const auto &dsl = pipeline_layout->set_layouts[set];
+    const auto &dsl = pipeline_layout.set_layouts[set];
     const auto lv_bind_point = ConvertToLvlBindPoint(pipelineBindPoint);
     auto &last_bound = lastBound[lv_bind_point];
     auto &push_descriptor_set = last_bound.push_descriptor_set;
@@ -829,7 +829,7 @@ void CMD_BUFFER_STATE::PushDescriptorSetState(VkPipelineBindPoint pipelineBindPo
     }
 
     UpdateLastBoundDescriptorSets(pipelineBindPoint, pipeline_layout, set, 1, nullptr, push_descriptor_set, 0, nullptr);
-    last_bound.pipeline_layout = pipeline_layout->layout();
+    last_bound.pipeline_layout = pipeline_layout.layout();
 
     // Now that we have either the new or extant push_descriptor set ... do the write updates against it
     push_descriptor_set->PerformPushDescriptorsUpdate(dev_data, descriptorWriteCount, pDescriptorWrites);
@@ -944,51 +944,53 @@ void CMD_BUFFER_STATE::UpdatePipelineState(CMD_TYPE cmd_type, const VkPipelineBi
     }
 }
 
+// Helper for descriptor set (and buffer) updates.
+static bool PushDescriptorCleanup(LAST_BOUND_STATE &last_bound, uint32_t set_idx) {
+    // All uses are from loops over per_set, but just in case..
+    assert(set_idx < last_bound.per_set.size());
+
+    auto ds = last_bound.per_set[set_idx].bound_descriptor_set.get();
+    if (ds && ds->IsPushDescriptor()) {
+        assert(ds == last_bound.push_descriptor_set.get());
+        last_bound.push_descriptor_set = nullptr;
+        return true;
+    }
+    return true;
+}
+
 // Update pipeline_layout bind points applying the "Pipeline Layout Compatibility" rules.
 // One of pDescriptorSets or push_descriptor_set should be nullptr, indicating whether this
 // is called for CmdBindDescriptorSets or CmdPushDescriptorSet.
 void CMD_BUFFER_STATE::UpdateLastBoundDescriptorSets(VkPipelineBindPoint pipeline_bind_point,
-                                                     const PIPELINE_LAYOUT_STATE *pipeline_layout, uint32_t first_set,
+                                                     const PIPELINE_LAYOUT_STATE &pipeline_layout, uint32_t first_set,
                                                      uint32_t set_count, const VkDescriptorSet *pDescriptorSets,
                                                      std::shared_ptr<cvdescriptorset::DescriptorSet> &push_descriptor_set,
                                                      uint32_t dynamic_offset_count, const uint32_t *p_dynamic_offsets) {
     assert((pDescriptorSets == nullptr) ^ (push_descriptor_set == nullptr));
-    // Defensive
-    assert(pipeline_layout);
-    if (!pipeline_layout) return;
 
     uint32_t required_size = first_set + set_count;
     const uint32_t last_binding_index = required_size - 1;
-    assert(last_binding_index < pipeline_layout->set_compat_ids.size());
+    assert(last_binding_index < pipeline_layout.set_compat_ids.size());
 
     // Some useful shorthand
     const auto lv_bind_point = ConvertToLvlBindPoint(pipeline_bind_point);
     auto &last_bound = lastBound[lv_bind_point];
-    last_bound.pipeline_layout = pipeline_layout->layout();
-    auto &pipe_compat_ids = pipeline_layout->set_compat_ids;
+    last_bound.pipeline_layout = pipeline_layout.layout();
+    auto &pipe_compat_ids = pipeline_layout.set_compat_ids;
     // Resize binding arrays
-    uint32_t last_set_index = first_set + set_count - 1;
-    if (last_set_index >= last_bound.per_set.size()) {
-        last_bound.per_set.resize(last_set_index + 1);
+    if (last_binding_index >= last_bound.per_set.size()) {
+        last_bound.per_set.resize(required_size);
     }
     const uint32_t current_size = static_cast<uint32_t>(last_bound.per_set.size());
-
-    // We need this three times in this function, but nowhere else
-    auto push_descriptor_cleanup = [&last_bound](const std::shared_ptr<cvdescriptorset::DescriptorSet> &ds) -> bool {
-        if (ds && ds->IsPushDescriptor()) {
-            assert(ds == last_bound.push_descriptor_set);
-            last_bound.push_descriptor_set = nullptr;
-            return true;
-        }
-        return false;
-    };
 
     // Clean up the "disturbed" before and after the range to be set
     if (required_size < current_size) {
         if (last_bound.per_set[last_binding_index].compat_id_for_set != pipe_compat_ids[last_binding_index]) {
             // We're disturbing those after last, we'll shrink below, but first need to check for and cleanup the push_descriptor
             for (auto set_idx = required_size; set_idx < current_size; ++set_idx) {
-                if (push_descriptor_cleanup(last_bound.per_set[set_idx].bound_descriptor_set)) break;
+                if (PushDescriptorCleanup(last_bound, set_idx)) {
+                    break;
+                }
             }
         } else {
             // We're not disturbing past last, so leave the upper binding data alone.
@@ -1005,9 +1007,8 @@ void CMD_BUFFER_STATE::UpdateLastBoundDescriptorSets(VkPipelineBindPoint pipelin
     for (uint32_t set_idx = 0; set_idx < first_set; ++set_idx) {
         auto &set_info = last_bound.per_set[set_idx];
         if (set_info.compat_id_for_set != pipe_compat_ids[set_idx]) {
-            push_descriptor_cleanup(set_info.bound_descriptor_set);
-            set_info.bound_descriptor_set = nullptr;
-            set_info.dynamicOffsets.clear();
+            PushDescriptorCleanup(last_bound, set_idx);
+            set_info.Reset();
             set_info.compat_id_for_set = pipe_compat_ids[set_idx];
         }
     }
@@ -1020,10 +1021,11 @@ void CMD_BUFFER_STATE::UpdateLastBoundDescriptorSets(VkPipelineBindPoint pipelin
         auto descriptor_set =
             push_descriptor_set ? push_descriptor_set : dev_data->Get<cvdescriptorset::DescriptorSet>(pDescriptorSets[input_idx]);
 
+        set_info.Reset();
         // Record binding (or push)
         if (descriptor_set != last_bound.push_descriptor_set) {
             // Only cleanup the push descriptors if they aren't the currently used set.
-            push_descriptor_cleanup(set_info.bound_descriptor_set);
+            PushDescriptorCleanup(last_bound, set_idx);
         }
         set_info.bound_descriptor_set = descriptor_set;
         set_info.compat_id_for_set = pipe_compat_ids[set_idx];  // compat ids are canonical *per* set index
@@ -1040,6 +1042,63 @@ void CMD_BUFFER_STATE::UpdateLastBoundDescriptorSets(VkPipelineBindPoint pipelin
                 set_info.dynamicOffsets.clear();
             }
         }
+    }
+}
+
+void CMD_BUFFER_STATE::UpdateLastBoundDescriptorBuffers(VkPipelineBindPoint pipeline_bind_point,
+                                                        const PIPELINE_LAYOUT_STATE &pipeline_layout, uint32_t first_set,
+                                                        uint32_t set_count, const uint32_t *buffer_indicies,
+                                                        const VkDeviceSize *buffer_offsets) {
+    uint32_t required_size = first_set + set_count;
+    const uint32_t last_binding_index = required_size - 1;
+    assert(last_binding_index < pipeline_layout.set_compat_ids.size());
+
+    // Some useful shorthand
+    const auto lv_bind_point = ConvertToLvlBindPoint(pipeline_bind_point);
+    auto &last_bound = lastBound[lv_bind_point];
+    last_bound.pipeline_layout = pipeline_layout.layout();
+    auto &pipe_compat_ids = pipeline_layout.set_compat_ids;
+    // Resize binding arrays
+    if (last_binding_index >= last_bound.per_set.size()) {
+        last_bound.per_set.resize(required_size);
+    }
+    const uint32_t current_size = static_cast<uint32_t>(last_bound.per_set.size());
+
+    // Clean up the "disturbed" before and after the range to be set
+    if (required_size < current_size) {
+        if (last_bound.per_set[last_binding_index].compat_id_for_set != pipe_compat_ids[last_binding_index]) {
+            // We're disturbing those after last, we'll shrink below, but first need to check for and cleanup the push_descriptor
+            for (auto set_idx = required_size; set_idx < current_size; ++set_idx) {
+                if (PushDescriptorCleanup(last_bound, set_idx)) {
+                    break;
+                }
+            }
+        } else {
+            // We're not disturbing past last, so leave the upper binding data alone.
+            required_size = current_size;
+        }
+    }
+
+    // We resize if we need more set entries or if those past "last" are disturbed
+    if (required_size != current_size) {
+        last_bound.per_set.resize(required_size);
+    }
+
+    // For any previously bound sets, need to set them to "invalid" if they were disturbed by this update
+    for (uint32_t set_idx = 0; set_idx < first_set; ++set_idx) {
+        PushDescriptorCleanup(last_bound, set_idx);
+        last_bound.per_set[set_idx].Reset();
+    }
+
+    // Now update the bound sets with the input sets
+    for (uint32_t input_idx = 0; input_idx < set_count; input_idx++) {
+        auto set_idx = input_idx + first_set;  // set_idx is index within layout, input_idx is index within input descriptor sets
+        auto &set_info = last_bound.per_set[set_idx];
+        set_info.Reset();
+
+        // Record binding
+        set_info.bound_descriptor_buffer = {buffer_indicies[input_idx], buffer_offsets[input_idx]};
+        set_info.compat_id_for_set = pipe_compat_ids[set_idx];  // compat ids are canonical *per* set index
     }
 }
 
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -1603,30 +1603,28 @@ bool CoreChecks::ValidateCmdBufDrawState(const CMD_BUFFER_STATE &cb_state, CMD_T
 
     bool result = false;
 
-    if (pipeline.descriptor_buffer_mode) {
-        if (!last_bound.per_set.empty()) {
-            bool non_push = false;
-            for (const auto &ds : last_bound.per_set) {
-                non_push = ds.bound_descriptor_set && !ds.bound_descriptor_set->IsPushDescriptor();
-
-                if (non_push) break;
+    for (const auto &ds : last_bound.per_set) {
+        if (pipeline.descriptor_buffer_mode) {
+            if (ds.bound_descriptor_set && !ds.bound_descriptor_set->IsPushDescriptor()) {
+                LogObjectList obj_list(cb_state.Handle(), pipeline.Handle(), ds.bound_descriptor_set->Handle());
+                result |= LogError(pipeline.pipeline(), vuid.descriptor_buffer_set_offset_missing,
+                                   "%s: pipeline bound to %s requires a descriptor buffer but has a bound descriptor set (%s)",
+                                   function, string_VkPipelineBindPoint(bind_point),
+                                   report_data->FormatHandle(ds.bound_descriptor_set->Handle()).c_str());
+                break;
             }
 
-            if (non_push)
-                result |=
-                    LogError(pipeline.pipeline(), vuid.descriptor_buffer_bit_not_set,
-                             "%s: If the descriptors used by the VkPipeline bound to the pipeline bind point were specified via "
-                             "vkCmdBindDescriptorSets, the bound VkPipeline must have been created without "
-                             "VK_PIPELINE_CREATE_DESCRIPTOR_BUFFER_BIT_EXT",
-                             function);
+        } else {
+            if (ds.bound_descriptor_buffer.has_value()) {
+                LogObjectList obj_list(cb_state.Handle(), pipeline.Handle());
+                result |= LogError(obj_list, vuid.descriptor_buffer_bit_not_set,
+                                   "%s: pipeline bound to %s requires a descriptor set but has a bound descriptor buffer"
+                                   " (index=%" PRIu32 " offset=%" PRIu64 ")",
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

**Comment by jeremyg-lunarg** [comment]
This is a draft because I still need to make a unit test for these 2 VUIDs, even if this PR isn't correct.

**Comment by archimedus** [comment]
> @archimedus I think you added the descriptor_buffer state stuff - are you able to look at this?

I am on it.

**Comment by archimedus** [comment]
@jeremyg-lunarg 

The code you propose is definitely better and fixes the issue. I would write following tests for these VUIDs:

// Test for 08115
pdb = createpipeline(flags | VK_PIPELINE_CREATE_DESCRIPTOR_BUFFER_BIT_EXT)
begincmdbuffer
vkCmdBindDescriptorSets(pdb)
vkCmdDraw()
endcmdbuffer

// Test for 08117
pds = createpipeline(flags & ~VK_PIPELINE_CREATE_DESCRIPTOR_BUFFER_BIT_EXT)
begincmdbuffer
vkCmdSetDescriptorBufferOffsetsEXT(pds)
vkCmdBindDescriptorBuffersEXT(pds)
vkCmdDraw()
endcmdbuffer

If you didn't start writing the tests I can do that.

**Comment by jeremyg-lunarg** [comment]
> If you didn't start writing the tests I can do that.

I haven't been able to start yet and that would be greatly appreciated.  Thanks!

File: layers/state_tracker.cpp [hunk 1]
```diff
@@ -3038,7 +3045,30 @@ void ValidationStateTracker::PreCallRecordCmdPushDescriptorSetKHR(VkCommandBuffe
...
+void ValidationStateTracker::PreCallRecordCmdBindDescriptorBuffersEXT(VkCommandBuffer commandBuffer, uint32_t bufferCount,
+                                                                      const VkDescriptorBufferBindingInfoEXT *pBindingInfos) {
+    auto cb_state = Get<CMD_BUFFER_STATE>(commandBuffer);
+
+    cb_state->descriptor_buffer_binding_info.clear();
+    cb_state->descriptor_buffer_binding_info.reserve(bufferCount);
+
+    for (uint32_t i = 0; i < bufferCount; i++) {
+        cb_state->descriptor_buffer_binding_info.push_back(pBindingInfos[i]);
+    }
```
juan-lunarg, line 3060: Use [insert](https://en.cppreference.com/w/cpp/container/vector/insert) instead for succinctness.

layers/state_tracker.cpp
```diff
@@ -3038,7 +3045,30 @@ void ValidationStateTracker::PreCallRecordCmdPushDescriptorSetKHR(VkCommandBuffe
...
-    cb_state->PushDescriptorSetState(pipelineBindPoint, pipeline_layout.get(), set, descriptorWriteCount, pDescriptorWrites);
+    cb_state->PushDescriptorSetState(pipelineBindPoint, *pipeline_layout, set, descriptorWriteCount, pDescriptorWrites);
+}
+
+void ValidationStateTracker::PreCallRecordCmdBindDescriptorBuffersEXT(VkCommandBuffer commandBuffer, uint32_t bufferCount,
+                                                                      const VkDescriptorBufferBindingInfoEXT *pBindingInfos) {
+    auto cb_state = Get<CMD_BUFFER_STATE>(commandBuffer);
+
+    cb_state->descriptor_buffer_binding_info.clear();
+    cb_state->descriptor_buffer_binding_info.reserve(bufferCount);
```
juan-lunarg, line 3056: Since clear doesn't change the capacity of the vector, the reserve call isn't needed.

```suggestion
```

juan-lunarg (changes requested):

layers/cmd_buffer_state.cpp
```diff
@@ -806,17 +806,17 @@ void CMD_BUFFER_STATE::ExecuteCommands(uint32_t commandBuffersCount, const VkCom
     }
 }
 
-void CMD_BUFFER_STATE::PushDescriptorSetState(VkPipelineBindPoint pipelineBindPoint, PIPELINE_LAYOUT_STATE *pipeline_layout,
+void CMD_BUFFER_STATE::PushDescriptorSetState(VkPipelineBindPoint pipelineBindPoint, PIPELINE_LAYOUT_STATE &pipeline_layout,
```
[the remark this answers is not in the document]
jeremyg-lunarg: wow I can't believe this compiles as-is.

layers/state_tracker.cpp
jeremyg-lunarg: Wouldn't this be less succinct since I need to supply an iterator to insert()?

layers/cmd_buffer_state.cpp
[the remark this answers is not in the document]
jeremyg-lunarg: looks like it wasn't const in the header but it can and should be.

layers/state_tracker.cpp
jeremyg-lunarg: switched to use resize() + indexing, which is much more concise

archimedus:
> I haven't been able to start yet and that would be greatly appreciated. Thanks!

I prepared tests in https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/b57df5b92392c9cf9f20fb331549f5fd72228662 , though it seems it needs to be retargeted (will do tomorrow)  . Feel free to retarget or just copy-paste tests from vklayertests_descriptor_renderpass_framebuffer.cpp.

jeremyg-lunarg:
> > I haven't been able to start yet and that would be greatly appreciated. Thanks!
> 
> I prepared tests in [b57df5b](https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/b57df5b92392c9cf9f20fb331549f5fd72228662) , though it seems it needs to be retargeted (will do tomorrow) . Feel free to retarget or just copy-paste tests from vklayertests_descriptor_renderpass_framebuffer.cpp.

Thanks, I'll pull them over to this PR today

layers/cmd_buffer_state.cpp
```diff
@@ -1041,6 +1038,73 @@ void CMD_BUFFER_STATE::UpdateLastBoundDescriptorSets(VkPipelineBindPoint pipelin
...
+    // Some useful shorthand
+    const auto lv_bind_point = ConvertToLvlBindPoint(pipeline_bind_point);
+    auto &last_bound = lastBound[lv_bind_point];
+    last_bound.pipeline_layout = pipeline_layout.layout();
+    auto &pipe_compat_ids = pipeline_layout.set_compat_ids;
+    // Resize binding arrays
+    uint32_t last_set_index = first_set + set_count - 1;
+    if (last_set_index >= last_bound.per_set.size()) {
+        last_bound.per_set.resize(last_set_index + 1);
+    }
```
juan-lunarg, line 1058: Why is `last_set_index` needed?

```suggestion
    // Resize binding arrays
    if (last_binding_index >= last_bound.per_set.size()) {
        last_bound.per_set.resize(required_size);
    }
```

layers/core_validation.cpp
```diff
@@ -1603,30 +1603,29 @@ bool CoreChecks::ValidateCmdBufDrawState(const CMD_BUFFER_STATE &cb_state, CMD_T
...
 
-    if (pipeline.descriptor_buffer_mode) {
-        if (!last_bound.per_set.empty()) {
-            bool non_push = false;
-            for (const auto &ds : last_bound.per_set) {
-                non_push = ds.bound_descriptor_set && !ds.bound_descriptor_set->IsPushDescriptor();
-
-                if (non_push) break;
+    for (uint32_t i = 0; i < last_bound.per_set.size(); i++) {
+        const auto &ds = last_bound.per_set[i];
```
juan-lunarg, line 1607: range-based for loop

```suggestion
    for (const auto &ds : last_bound.per_set) {
```

layers/core_validation.cpp
```diff
@@ -1668,7 +1667,7 @@ bool CoreChecks::ValidateCmdBufDrawState(const CMD_BUFFER_STATE &cb_state, CMD_T
 
     // Check if the current pipeline is compatible for the maximum used set with the bound sets.
     if (!pipeline.descriptor_buffer_mode) {
-        if (pipeline.active_slots.size() > 0 && !IsBoundSetCompat(pipeline.max_active_slot, last_bound, pipeline_layout.get())) {
+        if (pipeline.active_slots.size() > 0 && !IsBoundSetCompat(pipeline.max_active_slot, last_bound, *pipeline_layout)) {
```
juan-lunarg, line 1670:
```suggestion
        if (!pipeline.active_slots.empty() && !IsBoundSetCompat(pipeline.max_active_slot, last_bound, *pipeline_layout)) {
```

layers/pipeline_state.h
```diff
@@ -681,9 +693,9 @@ struct LAST_BOUND_STATE {
 };
 
 static inline bool IsBoundSetCompat(uint32_t set, const LAST_BOUND_STATE &last_bound,
-                                    const PIPELINE_LAYOUT_STATE *pipeline_layout) {
-    bool result = (set < last_bound.per_set.size()) && (set < pipeline_layout->set_compat_ids.size()) &&
-                  (*(last_bound.per_set[set].compat_id_for_set) == *(pipeline_layout->set_compat_ids[set]));
+                                    const PIPELINE_LAYOUT_STATE &pipeline_layout) {
+    bool result = (set < last_bound.per_set.size()) && (set < pipeline_layout.set_compat_ids.size()) &&
+                  (*(last_bound.per_set[set].compat_id_for_set) == *(pipeline_layout.set_compat_ids[set]));
```
juan-lunarg, line 698: The bool isn't needed, return the expression directly.

layers/state_tracker.cpp
```diff
@@ -4241,7 +4270,7 @@ void ValidationStateTracker::PreCallRecordCmdPushDescriptorSetWithTemplateKHR(Vk
         // Decode the template into a set of write updates
         cvdescriptorset::DecodedTemplateUpdate decoded_template(this, VK_NULL_HANDLE, template_state.get(), pData,
                                                                 dsl->GetDescriptorSetLayout());
-        cb_state->PushDescriptorSetState(template_ci.pipelineBindPoint, layout_data.get(), set,
+        cb_state->PushDescriptorSetState(template_ci.pipelineBindPoint, *layout_data, set,
```
juan-lunarg, line 4273: Looks like layout_data could be nullptr based on line 4268

juan-lunarg (changes requested):

layers/pipeline_state.h
jeremyg-lunarg: Based on the comment in IsPipelineLayoutSetCompat() about wanting the bool to debug this expression, I'm going to leave this one that way too (and copy the comment)

juan-lunarg: It may be worth breaking up this function into multiple bools for readability/debugability.

jeremyg-lunarg: Yeah I might as well since I'm touching it.

layers/pipeline_state.h
```diff
@@ -681,16 +693,23 @@ struct LAST_BOUND_STATE {
...
+    }
     return result;
 }
 
 static inline bool IsPipelineLayoutSetCompat(uint32_t set, const PIPELINE_LAYOUT_STATE *a, const PIPELINE_LAYOUT_STATE *b) {
     // Intentionally have a result variable to simplify debugging
-    bool result = a && b && (set < a->set_compat_ids.size()) && (set < b->set_compat_ids.size()) &&
-                  (a->set_compat_ids[set] == b->set_compat_ids[set]);
+    bool result = false;
+    if (a && b) {
```
juan-lunarg, line 708: Avoid nested if statement.

```suggestion
    if (a == nullptr || b == nullptr) { return false; }
```

layers/state_tracker.cpp
```diff
@@ -4234,14 +4263,14 @@ void ValidationStateTracker::PreCallRecordCmdPushDescriptorSetWithTemplateKHR(Vk
 
     cb_state->RecordCmd(CMD_PUSHDESCRIPTORSETWITHTEMPLATEKHR);
     auto template_state = Get<UPDATE_TEMPLATE_STATE>(descriptorUpdateTemplate);
-    if (template_state) {
-        auto layout_data = Get<PIPELINE_LAYOUT_STATE>(layout);
-        auto dsl = layout_data ? layout_data->GetDsl(set) : nullptr;
+    auto layout_data = Get<PIPELINE_LAYOUT_STATE>(layout);
+    if (template_state && layout_data) {
```
juan-lunarg, line 4267: Avoid code indentation for readability.

```suggestion
    if (template_state == nullptr || layout_data == nullptr) { return; }
```

layers/state_tracker.cpp
```diff
@@ -3038,7 +3045,29 @@ void ValidationStateTracker::PreCallRecordCmdPushDescriptorSetKHR(VkCommandBuffe
...
+
+void ValidationStateTracker::PreCallRecordCmdBindDescriptorBuffersEXT(VkCommandBuffer commandBuffer, uint32_t bufferCount,
+                                                                      const VkDescriptorBufferBindingInfoEXT *pBindingInfos) {
+    auto cb_state = Get<CMD_BUFFER_STATE>(commandBuffer);
+
+    cb_state->descriptor_buffer_binding_info.resize(bufferCount);
+
+    for (uint32_t i = 0; i < bufferCount; i++) {
+        cb_state->descriptor_buffer_binding_info[i] = pBindingInfos[i];
+    }
```
juan-lunarg, line 3059:
```suggestion
    std::copy(pBindingInfos, pBindingInfos + bufferCount, cb_state->descriptor_buffer_binding_info.data());
```

layers/cmd_buffer_state.cpp
```diff
@@ -1043,6 +1039,72 @@ void CMD_BUFFER_STATE::UpdateLastBoundDescriptorSets(VkPipelineBindPoint pipelin
...
+
+    // We resize if we need more set entries or if those past "last" are disturbed
+    if (required_size != current_size) {
+        last_bound.per_set.resize(required_size);
+    }
+
+    // For any previously bound sets, need to set them to "invalid" if they were disturbed by this update
+    for (uint32_t set_idx = 0; set_idx < first_set; ++set_idx) {
+        auto &set_info = last_bound.per_set[set_idx];
+        push_descriptor_cleanup(set_info.bound_descriptor_set);  // TODO
```
juan-lunarg, line 1092: TODO?

layers/cmd_buffer_state.cpp
```diff
@@ -1043,6 +1039,72 @@ void CMD_BUFFER_STATE::UpdateLastBoundDescriptorSets(VkPipelineBindPoint pipelin
...
+    auto &last_bound = lastBound[lv_bind_point];
+    last_bound.pipeline_layout = pipeline_layout.layout();
+    auto &pipe_compat_ids = pipeline_layout.set_compat_ids;
+    // Resize binding arrays
+    if (last_binding_index >= last_bound.per_set.size()) {
+        last_bound.per_set.resize(required_size);
+    }
+    const uint32_t current_size = static_cast<uint32_t>(last_bound.per_set.size());
+
+    // We need this three times in this function, but nowhere else
```
juan-lunarg, line 1061:
```suggestion
    // We need this two times in this function, but nowhere else
```

layers/state_tracker.cpp
```diff
@@ -3028,8 +3035,8 @@ void ValidationStateTracker::PreCallRecordCmdBindDescriptorSets(VkCommandBuffer
     auto pipeline_layout = Get<PIPELINE_LAYOUT_STATE>(layout);
     std::shared_ptr<cvdescriptorset::DescriptorSet> no_push_desc;
 
-    cb_state->UpdateLastBoundDescriptorSets(pipelineBindPoint, pipeline_layout.get(), firstSet, setCount, pDescriptorSets,
-                                            no_push_desc, dynamicOffsetCount, pDynamicOffsets);
+    cb_state->UpdateLastBoundDescriptorSets(pipelineBindPoint, *pipeline_layout, firstSet, setCount, pDescriptorSets, no_push_desc,
```
juan-lunarg, line 3038: Potential nullptr dereference

```suggestion
    if (pipeline_layout == nullptr) { return; }
    
    cb_state->UpdateLastBoundDescriptorSets(pipelineBindPoint, *pipeline_layout, firstSet, setCount, pDescriptorSets, no_push_desc,
```

tests/vklayertests_descriptor_renderpass_framebuffer.cpp
```diff
@@ -13937,6 +13937,207 @@ TEST_F(VkLayerTest, DescriptorBufferBindingAndOffsets) {
...
+    plci.setLayoutCount = 1;
+    plci.pSetLayouts = &dsl;
+
+    err = vk::CreatePipelineLayout(device(), &plci, NULL, &pipeline_layout);
+    ASSERT_VK_SUCCESS(err);
+
+    uint32_t qfi = 0;
+    VkBufferCreateInfo buffCI = LvlInitStruct<VkBufferCreateInfo>();
+    buffCI.size = 4096;
+    buffCI.usage = VK_BUFFER_USAGE_SHADER_DEVICE_ADDRESS_BIT | VK_BUFFER_USAGE_SAMPLER_DESCRIPTOR_BUFFER_BIT_EXT;
```
juan-lunarg, line 13995: `VK_BUFFER_USAGE_SHADER_DEVICE_ADDRESS_BIT` is used but `VK_KHR_BUFFER_DEVICE_ADDRESS_EXTENSION_NAME` was listed as an optional extension.

juan-lunarg (changes requested):

layers/cmd_buffer_state.cpp
juan-lunarg: Considering this lambda is used in 2 different functions now, it may be worth making into it's own static function.

tests/vklayertests_descriptor_renderpass_framebuffer.cpp
```diff
@@ -13937,6 +13937,207 @@ TEST_F(VkLayerTest, DescriptorBufferBindingAndOffsets) {
     vk::DestroyPipelineLayout(m_device->device(), pipeline_layout, NULL);
 }
 
+TEST_F(VkLayerTest, DescriptorBufferIncosistentBuffer) {
```
juan-lunarg, line 13940:
```suggestion
TEST_F(VkLayerTest, DescriptorBufferInconsistentBuffer) {
```

tests/vklayertests_descriptor_renderpass_framebuffer.cpp
```diff
@@ -13937,6 +13937,207 @@ TEST_F(VkLayerTest, DescriptorBufferBindingAndOffsets) {
...
+
+    m_commandBuffer->end();
+
+    vk::DestroyBuffer(m_device->device(), buffer, NULL);
+    vk::FreeMemory(m_device->device(), mem, NULL);
+    vk::DestroyDescriptorSetLayout(m_device->device(), dsl, NULL);
+    vk::DestroyPipelineLayout(m_device->device(), pipeline_layout, NULL);
+}
+
+TEST_F(VkLayerTest, DescriptorBufferIncosistentSet) {
```
juan-lunarg, line 14052:
```suggestion
TEST_F(VkLayerTest, DescriptorBufferInconsistentSet) {
```

tests/vklayertests_descriptor_renderpass_framebuffer.cpp
```diff
@@ -13937,6 +13937,207 @@ TEST_F(VkLayerTest, DescriptorBufferBindingAndOffsets) {
...
+    auto vkCmdSetDescriptorBufferOffsetsEXT = reinterpret_cast<PFN_vkCmdSetDescriptorBufferOffsetsEXT>(
+        vk::GetDeviceProcAddr(m_device->device(), "vkCmdSetDescriptorBufferOffsetsEXT"));
+
+    VkResult err;
+    VkDescriptorSetLayout dsl;
+    VkPipelineLayout pipeline_layout;
+
+    const VkDescriptorSetLayoutBinding bindings[] = {
+        {0, VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER, 1, VK_SHADER_STAGE_COMPUTE_BIT, nullptr},
+    };
```
juan-lunarg, line 13978: Use std::array

tests/vklayertests_descriptor_renderpass_framebuffer.cpp
```diff
@@ -13937,6 +13937,207 @@ TEST_F(VkLayerTest, DescriptorBufferBindingAndOffsets) {
...
+    if (IsPlatform(kMockICD)) {
+        GTEST_SKIP() << "Test not supported by MockICD";
+    }
+
+    ASSERT_NO_FATAL_FAILURE(InitState(nullptr, &features2, VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT));
+
+    auto vkCmdBindDescriptorBuffersEXT = reinterpret_cast<PFN_vkCmdBindDescriptorBuffersEXT>(
+        vk::GetDeviceProcAddr(m_device->device(), "vkCmdBindDescriptorBuffersEXT"));
+    auto vkCmdSetDescriptorBufferOffsetsEXT = reinterpret_cast<PFN_vkCmdSetDescriptorBufferOffsetsEXT>(
+        vk::GetDeviceProcAddr(m_device->device(), "vkCmdSetDescriptorBufferOffsetsEXT"));
```
juan-lunarg, line 13970: ASSERT_TRUE that the function pointers aren't nullptr

tests/vklayertests_descriptor_renderpass_framebuffer.cpp
```diff
@@ -13937,6 +13937,207 @@ TEST_F(VkLayerTest, DescriptorBufferBindingAndOffsets) {
...
+TEST_F(VkLayerTest, DescriptorBufferIncosistentSet) {
+    TEST_DESCRIPTION("Dispatch pipeline with descriptor buffer bound while of descriptor set expected");
+    SetTargetApiVersion(VK_API_VERSION_1_2);
+
+    AddRequiredExtensions(VK_EXT_DESCRIPTOR_BUFFER_EXTENSION_NAME);
+    AddOptionalExtensions(VK_KHR_BUFFER_DEVICE_ADDRESS_EXTENSION_NAME);
+
+    auto descriptor_buffer_features = LvlInitStruct<VkPhysicalDeviceDescriptorBufferFeaturesEXT>();
+    auto buffer_device_address_features = LvlInitStruct<VkPhysicalDeviceBufferDeviceAddressFeatures>(&descriptor_buffer_features);
+    auto features2 = LvlInitStruct<VkPhysicalDeviceFeatures2KHR>(&buffer_device_address_features);
```
juan-lunarg, line 14061: Is this extension needed? I don't see it used at all, unlike the prior test.

tests/vklayertests_descriptor_renderpass_framebuffer.cpp
```diff
@@ -13937,6 +13937,207 @@ TEST_F(VkLayerTest, DescriptorBufferBindingAndOffsets) {
...
+
+    ASSERT_NO_FATAL_FAILURE(InitState(nullptr, &features2, VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT));
+
+    VkResult err;
+    VkDescriptorSetLayout dsl;
+    VkPipelineLayout pipeline_layout;
+
+    const VkDescriptorSetLayoutBinding bindings[] = {
+        {0, VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER, 1, VK_SHADER_STAGE_COMPUTE_BIT, nullptr},
+    };
```
juan-lunarg, line 14085: use std::array

juan-lunarg (changes requested):

layers/pipeline_state.h
ncesario-lunarg: In general I think we avoid the `== nullptr` as we use a lot of smart pointers, and this way we can take advantage of `operator bool` if the type changes.

jeremyg-lunarg: I was under the impression that we were trying to *not* use `== nullptr` for checks like this.

juan-lunarg: I updated the suggestion to reflect that.

layers/cmd_buffer_state.cpp
jeremyg-lunarg: Yeah. done.

tests/vklayertests_descriptor_renderpass_framebuffer.cpp
jeremyg-lunarg: VK_EXT_descriptor_buffer requires VK_EXT_buffer_device_address, which is why these tests can get away with being inconsistent.

layers/cmd_buffer_state.cpp
```diff
@@ -1043,6 +1045,64 @@ void CMD_BUFFER_STATE::UpdateLastBoundDescriptorSets(VkPipelineBindPoint pipelin
...
+    const auto lv_bind_point = ConvertToLvlBindPoint(pipeline_bind_point);
+    auto &last_bound = lastBound[lv_bind_point];
+    last_bound.pipeline_layout = pipeline_layout.layout();
+    auto &pipe_compat_ids = pipeline_layout.set_compat_ids;
+    // Resize binding arrays
+    if (last_binding_index >= last_bound.per_set.size()) {
+        last_bound.per_set.resize(required_size);
+    }
+    const uint32_t current_size = static_cast<uint32_t>(last_bound.per_set.size());
+
```
juan-lunarg, line 1066:
```suggestion
```

layers/pipeline_state.h
```diff
@@ -681,17 +693,21 @@ struct LAST_BOUND_STATE {
...
 
 static inline bool IsPipelineLayoutSetCompat(uint32_t set, const PIPELINE_LAYOUT_STATE *a, const PIPELINE_LAYOUT_STATE *b) {
-    // Intentionally have a result variable to simplify debugging
-    bool result = a && b && (set < a->set_compat_ids.size()) && (set < b->set_compat_ids.size()) &&
-                  (a->set_compat_ids[set] == b->set_compat_ids[set]);
-    return result;
+    if (!a || !b) {
+        return false;
+    }
+    if ((set >= a->set_compat_ids.size()) && (set >= b->set_compat_ids.size())) {
```
juan-lunarg, line 707: Minor logic issue.

```suggestion
    if ((set >= a->set_compat_ids.size()) || (set >= b->set_compat_ids.size())) {
```

juan-lunarg (changes requested):

layers/pipeline_state.h
jeremyg-lunarg: that's more than minor!

juan-lunarg (approved):
LGTM

jeremyg-lunarg:
> LGTM

Thanks for the thorough reviewing. I'm making a separate PR to clean up the other tests.

juan-lunarg:
> Thanks for the thorough reviewing. I'm making a separate PR to clean up the other tests.

No worries. I tried to run the tests as a double check but my local system doesn't support the extension you are testing.

PR #4919, opened 2022-12-05
layers: Use std::array for best practice codes

juan-lunarg:


layers/best_practices_validation.h
```diff
@@ -694,8 +694,8 @@ class BestPractices : public ValidationStateTracker {
     bool PreCallValidateCmdClearAttachments(VkCommandBuffer commandBuffer, uint32_t attachmentCount,
                                             const VkClearAttachment* pAttachments, uint32_t rectCount,
                                             const VkClearRect* pRects) const override;
-    void ValidateReturnCodes(const char* api_name, VkResult result, const std::vector<VkResult>& error_codes,
-                             const std::vector<VkResult>& success_codes) const;
+    void ValidateReturnCodes(const char* api_name, VkResult result, layer_data::span<const VkResult> error_codes,
+                             layer_data::span<const VkResult> success_codes) const;
```
juan-lunarg, line 698: Note a span<T> object does not own its elements and is so small that it can be passed by value.

jeremyg-lunarg (approved):
LGTM

Issue #4414, opened 2022-08-13
Validation error regarding VkPipelineLayout lifetime


Cyphall:
**Describe the Issue**
The Vulkan specification says :
> A VkRenderPass or VkPipelineLayout object passed as a parameter to create another object is not further accessed by that object after the duration of the command it is passed into.

If I understand this correctly, it means that when I create a `VkPipeline` using a `VkPipelineLayout`, once the call to `vkCreateGraphicsPipelines()` returns, the `VkPipelineLayout` can be safely destroyed.
However, the validation message I recieved tells me that I am supposed to keep the `VkPipelineLayout` object alive for the lifetime of the `VkPipeline`.

**Valid Usage ID**
```
ERROR (UNASSIGNED-GPU-Assisted-Validation): Validation Error: [ UNASSIGNED-GPU-Assisted-Validation ] Object 0: handle = 0x1765c3483b0, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0xea8d5c05 | Setup Error. Detail: (Pipeline layout has been destroyed, aborting GPU-AV)
```

**Environment:**
 - OS: Windows 10 21H2
 - GPU: RTX 3070 (driver version 512.95)
 - SDK version: 1.3.216.0
 - Options enabled:
     - `vk::ValidationFeatureEnableEXT::eGpuAssisted`
     - `vk::ValidationFeatureEnableEXT::eSynchronizationValidation`

Tony-LunarG:
Addressed in #4922

PR #4922, opened 2022-12-06
Better handle unavailable pipeline layout in GPU-AV

Tony-LunarG:
Also fixes VMA memory leak
Also addresses #4414

### DIFF
--- a/layers/gpu_validation.cpp
+++ b/layers/gpu_validation.cpp
@@ -1288,6 +1288,7 @@ void GpuAssisted::PostCallRecordCmdBindDescriptorSets(VkCommandBuffer commandBuf
     ValidationStateTracker::PostCallRecordCmdBindDescriptorSets(commandBuffer, pipelineBindPoint, layout, firstSet,
                                                                 descriptorSetCount, pDescriptorSets, dynamicOffsetCount,
                                                                 pDynamicOffsets);
+    if (aborted) return;
     auto cb_node = GetWrite<gpuav_state::CommandBuffer>(commandBuffer);
     if (!cb_node) {
         ReportSetupProblem(device, "Unrecognized command buffer");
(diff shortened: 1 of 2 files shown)


jeremyg-lunarg (approved):
LGTM

layers/gpu_validation.cpp
```diff
@@ -2224,14 +2225,21 @@ void GpuAssisted::AllocateValidationResources(const VkCommandBuffer cmd_buffer,
...
         // may be a "pseudo layout" used to represent the union of pre-raster and fragment shader layouts, and therefore have a
         // null handle.
-        const auto pipeline_layout_handle =
-            (last_bound.pipeline_layout) ? last_bound.pipeline_layout : pipeline_state->PreRasterPipelineLayoutState()->layout();
-        if ((pipeline_layout->set_layouts.size() <= desc_set_bind_index) && !pipeline_layout->Destroyed()) {
+        VkPipelineLayout pipeline_layout_handle = VK_NULL_HANDLE;
+        if (last_bound.pipeline_layout)
+            pipeline_layout_handle = last_bound.pipeline_layout;
+        else if (enabled_features.graphics_pipeline_library_features.graphicsPipelineLibrary)
+            pipeline_layout_handle = pipeline_state->PreRasterPipelineLayoutState()->layout();
```
ncesario-lunarg, line 2232: I _think_ you only need this and the previous case. IOW:
```cpp
        if (last_bound.pipeline_layout)
            pipeline_layout_handle = last_bound.pipeline_layout;
        else if (!pipeline_state->PreRasterPipelineLayoutState()->Destroyed())
            pipeline_layout_handle = pipeline_state->PreRasterPipelineLayoutState()->layout();
```
If GPL is not used, `pipeline_state->PreRasterPipelineLayoutState() == pipeline_state->PipelineLayoutState()`, so the last 2 conditions seem somewhat redundant. It might be slightly clearer with 3 branches; not sure.

Also some `{}` would be greatly appreciated, for my clumsiness, and to adhere to the [style guide](https://google.github.io/styleguide/cppguide.html#Conditionals) :)

ncesario-lunarg (commented):
One comment, but LGTM!

layers/gpu_validation.cpp
Tony-LunarG: I didn't know that and I like your suggestion much better.  And clang-format failed me, but now I know the {} rules. :-)

ncesario-lunarg: > And clang-format failed me

Yeah, I think clang-format just handles white space; we need to get clang-tidy integrated to automatically handle this (I think).

ncesario-lunarg (approved):

PR #4923, opened 2022-12-06
state: Simplify CMD_BUFFER_STATE object tracking

jeremyg-lunarg:
Remove the `framebuffers` map since it is redundant with `object_bindings`.

Work around a possible race condition in NotifyInvalidate. Due to interactions between the `BASE_NODE::tree_lock_` and `CMD_BUFFER_STATE::lock` it was possible for a (not really) 'bad' binding to get added to `broken_bindings`. Try to fix this by only adding a node to `broken_bindings` if it was still found in `object_bindings`.

### DIFF
--- a/layers/cmd_buffer_state.cpp
+++ b/layers/cmd_buffer_state.cpp
@@ -117,6 +117,13 @@ void CMD_BUFFER_STATE::RemoveChild(std::shared_ptr<BASE_NODE> &child_node) {
 // Reset the command buffer state
 //  Maintain the createInfo and set state to CB_NEW, but clear all other state
 void CMD_BUFFER_STATE::Reset() {
+    // Remove object bindings
+    for (const auto &obj : object_bindings) {
+        obj->RemoveParent(this);
+    }
+    object_bindings.clear();
+    broken_bindings.clear();
+
     // Reset CB state (note that createInfo is not cleared)
     memset(&beginInfo, 0, sizeof(VkCommandBufferBeginInfo));
     memset(&inheritanceInfo, 0, sizeof(VkCommandBufferInheritanceInfo));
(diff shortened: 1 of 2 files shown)


ncesario-lunarg (approved):

jeremyg-lunarg:
This is a test change to help debug an internally reported problem.

layers/cmd_buffer_state.cpp
```diff
@@ -298,25 +293,32 @@ void CMD_BUFFER_STATE::NotifyInvalidate(const BASE_NODE::NodeList &invalid_nodes
...
-            for (auto &obj : invalid_nodes) {
-                object_bindings.erase(obj);
-                switch (obj->Type()) {
-                    case kVulkanObjectTypeCommandBuffer:
+        for (auto &obj : invalid_nodes) {
+            // Only record a broken binding if one of the nodes in the invalid chain is still
+            // being tracked by the command buffer. This is to try to avoid race conditions
+            // caused by separate CMD_BUFFER_STATE and BASE_NODE::parent_nodes locking.
+            if (object_bindings.erase(obj)) {
+                broken_bindings.emplace(invalid_nodes[0]->Handle(), log_list);
```
danginsburg, line 307: Should state be changed only if a broken_binding is found?  Otherwise what happens if vkBeginCommandBuffer was called and this kicks in - it will assign the state to CB_INVALID_INCOMPLETE in the check at the top of this function.

I'm not sure if this is a real problem because I haven't seen it happen when testing this fix, but theoretically it seems plausible.

jeremyg-lunarg: yes I'm working on an update to do state changes only if this condition is true.

Tony-LunarG (approved):
LGTM

PR #4925, opened 2022-12-06
gn: Disable all warnings for generated code

### LINKED ISSUE #4924: Chrome builds are failing on "layers: Modernize generated code"
After commit b4539c542c35d1643bdf73b426fc951ea035b65e, Chrome builds seem to be failing with:

Step _compile (with patch)_ failed. Error logs are shown below:
[2131/24370] CXX obj/third_party/vulkan-deps/vulkan-validation-layers/src/VkLayer_khronos_validation/parameter_validation.o
FAILED: obj/third_party/vulkan-deps/vulkan-validation-layers/src/VkLayer_khronos_validation/parameter_validation.o
../../buildtools/reclient/rewrapper -cfg=../../buildtools/reclient_cfgs/chromium-browser-clang/rewra...(too long)
../../third_party/vulkan-deps/vulkan-validation-layers/src/layers/generated/parameter_validation.cpp:5336:376: error: integer value 3000300003 is outside the valid range of values [0, 2147483647] for this enumeration type [-Wenum-constexpr-conversion]

juan-lunarg:
closes #4924

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -46,14 +46,7 @@ action("vulkan_clean_old_validation_layer_objects") {
 config("generated_layers_config") {
   if (is_clang) {
     cflags = [
-      "-Wno-conversion",
-      "-Wno-deprecated-copy",
-      "-Wno-extra-semi",
-      "-Wno-implicit-fallthrough",
-      "-Wno-missing-field-initializers",
-      "-Wno-newline-eof",
-      "-Wno-sign-compare",
-      "-Wno-unused-const-variable",
+      "-Wno-everything"
     ]
   }
 }


jeremyg-lunarg (approved):
LGTM. Are we doing this permanently or just to unblock the issue reporter?

juan-lunarg:
> LGTM. Are we doing this permanently or just to unblock the issue reporter?

Originally I was going to just change the value of `VK_STRUCTURE_TYPE_INSTANCE_LAYER_SETTINGS_EXT` to be in the valid range described in the issue [0, INT_MAX].

The problem is I have no idea if someone is depending on the existing value of `VK_STRUCTURE_TYPE_INSTANCE_LAYER_SETTINGS_EXT`. 

As far as I can tell VVL is the only one that relies on the value but I'm not sure.

juan-lunarg:
After checking our SDK scripts, conan, chromium build, github, and general googling. I think I'm safe to change the value of VK_STRUCTURE_TYPE_INSTANCE_LAYER_SETTINGS_EXT

PR #4933, opened 2022-12-08
layers: Remove legacy lvl_* vk_typemap_helper

spencer-lunarg:
`lvl_find_in_chain` -> `LvlFindInChain`
`lvl_find_mod_in_chain` -> `LvlFindModInChain`
`lvl_init_struct` -> `LvlInitStruct`

might cause some merge conflicts, but should be small enough now to just fully remove (82 spots replaced)


jeremyg-lunarg (approved):
LGTM. Yay!

mark-lunarg (approved):
LGTM!  Thanks, I have wanted to do this for the longest time.

mark-lunarg:
On thing, I'm too lazy to check but I think `Vulkan-Tools` uses these old functions and I'm not sure if it relies on this repo for the defs or not.

jeremyg-lunarg:
> On thing, I'm too lazy to check but I think `Vulkan-Tools` uses these old functions and I'm not sure if it relies on this repo for the defs or not.

I have a VulkanTools repo locally and `src/LunarG/VulkanTools>git grep lvl_` returns nothing.

mark-lunarg:
> > On thing, I'm too lazy to check but I think `Vulkan-Tools` uses these old functions and I'm not sure if it relies on this repo for the defs or not.
> 
> I have a VulkanTools repo locally and `src/LunarG/VulkanTools>git grep lvl_` returns nothing.

VulkanTools doesn't but it looks like Vulkan-Tools still does?

layers/generated/vk_typemap_helper.h
```diff
@@ -6964,50 +6964,3 @@ template <typename T> T LvlInitStruct() {
     return out;
 }
 
-
-// Find an entry of the given type in the const pNext chain
-template <typename T> const T *lvl_find_in_chain(const void *next) {
```
juan-lunarg, line 6969: Looks like these functions are being used by the MockICD.

https://github.com/KhronosGroup/Vulkan-Tools/blob/ce45337c51ed53e6f1b337aa746a5a96c5018f86/icd/generated/mock_icd.cpp#L2936

jeremyg-lunarg: @charles-lunarg it looks to me like Vulkan-Tools uses a separate copy of this codegen rather than having a dependency on VVL. Is that correct?

charles-lunarg: I believe so.

PR #4934, opened 2022-12-08
build: Update to header 1.3.237

mikes-lunarg:
- Update known-good
- Generate source


mikes-lunarg:
FYI @mark-lunarg this will bring in the new maintenance extensions

ziga-lunarg (approved):

PR #4938, opened 2022-12-09
layers: Fix mem leak

juan-lunarg:


layers/core_validation.cpp
```diff
@@ -4809,6 +4809,7 @@ void CoreChecks::PreCallRecordDestroyDevice(VkDevice device, const VkAllocationC
 
         if (result != VK_SUCCESS) {
             LogInfo(device, "UNASSIGNED-cache-retrieval-error", "Validation Cache Retrieval Error");
+            free(validation_cache_data);
```
juan-lunarg, line 4812: clang-tidy found this minor issue. Where we return early on an error and don't free the memory.

Tony-LunarG (approved):
LGTM

jeremyg-lunarg (approved):
Fix is correct but could we use std::vector<uint8_t> or some other RAII object to avoid the need for freeing at all?

juan-lunarg:
> Fix is correct but could we use std::vector<uint8_t> or some other RAII object to avoid the need for freeing at all?

My plan is to address all the malloc/free calls in VVL in another PR.

PR #4944, opened 2022-12-10
layers: Add helper function for counting bits set

spencer-lunarg:
Create a `GetBitSetCount` util for all the spots doing 32-bit mask bit counting


mark-lunarg (approved):
This is nice!  Just did a bunch of these the old way...

PR #4956, opened 2022-12-13
layers: Make bool const where possible

spencer-lunarg:
`^((?!const).)*bool (?!skip)\w* = ((?!false)(?!true))` seemed to work well finding cases where `bool` were able to be `const`


juan-lunarg (approved):
:+1:

PR #4959, opened 2022-12-13
tests: Use vk_testing:: objects in descriptor tests

jeremyg-lunarg:
Add some stuff to make the vk_testing:: objects easier to use:

- vk_testing::*::destroy() methods 
- vk_testing::Buffer::address() and another constructor that makes it easy to set VK_MEMORY_ALLOCATE_DEVICE_ADDRESS_BIT

Then convert a bunch of descriptor tests to use these objects instead of raw vulkan handles.

### DIFF
--- a/tests/vklayertests_descriptor_renderpass_framebuffer.cpp
+++ b/tests/vklayertests_descriptor_renderpass_framebuffer.cpp
@@ -32,8 +32,6 @@
 #include "layer_validation_tests.h"
 
 TEST_F(VkLayerTest, InvalidDescriptorPoolConsistency) {
-    VkResult err;
-
     TEST_DESCRIPTION("Allocate descriptor sets from one DS pool and attempt to delete them from another.");
 
     m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-vkFreeDescriptorSets-pDescriptorSets-parent");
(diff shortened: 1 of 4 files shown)


spencer-lunarg (approved):
LGTM, really reduces a lot of code in a nice way!

tests/vktestbinding.cpp
```diff
@@ -46,12 +47,14 @@ namespace {
     } while (0)
 
 #define NON_DISPATCHABLE_HANDLE_DTOR(cls, destroy_func) \
-    cls::~cls() noexcept {                              \
-        if (initialized()) {                            \
-            destroy_func(device(), handle(), NULL);     \
-            handle_ = VK_NULL_HANDLE;                   \
+    void cls::destroy() {                               \
+        if (!initialized()) {                           \
```
juan-lunarg, line 51: Shouldn't `destroy` also be marked an `noexcept`?

tests/vktestbinding.h
```diff
@@ -519,6 +532,8 @@ class Buffer : public internal::NonDispHandle<VkBuffer> {
         return barrier;
     }
 
+    VkDeviceAddress address() const;
```
juan-lunarg, line 535: Use ` [[nodiscard]] ` attribute.

tests/vktestbinding.cpp
jeremyg-lunarg: It could be.

tests/vktestbinding.h
jeremyg-lunarg: I'm not sure about this. vkGetBufferDeviceAddress() doesn't allocate the address, just returns it.  Does it really matter if the caller discards it?

juan-lunarg: Why would you call the function if you intend to discard the result?

jeremyg-lunarg: Ok. I don't have a good reason for to not tell people not to do that. :)

arno-lunarg: Pls do not add `[[nodiscard]]`! For some tests I wrote, I called `vk::GetBufferDeviceAddress` without storing the address, because calling this function will register the buffer and its address in the address -> buffers map.

arno-lunarg:
Nice, I am waiting for this to be merged so that I use those changes in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4861

tests/vktestbinding.h
arno-lunarg: Or maybe it is up to me to add [[maybe_unused]], since what I describe is a corner case...

Tony-LunarG (approved):
LGTM! Nice Cleanup

tests/vklayertests_descriptor_renderpass_framebuffer.cpp
```diff
@@ -3462,51 +3337,39 @@ TEST_F(VkLayerTest, FramebufferCreateErrors) {
...
-    {
-        if (!multiviewSupported) {
-            printf("VK_KHR_Multiview Extension not supported, skipping tests\n");
-        } else {
-            // Test multiview renderpass with more than 1 layer
-            uint32_t viewMasks[] = {0x3u};
-            auto rpmvci = LvlInitStruct<VkRenderPassMultiviewCreateInfo>(nullptr, 1u, viewMasks, 0u, nullptr, 0u, nullptr);
-            auto rpci_mv = LvlInitStruct<VkRenderPassCreateInfo>(&rpmvci, 0u, 0u, nullptr, 1u, &subpass, 0u, nullptr);
+    if (!multiviewSupported) {
+        printf("VK_KHR_Multiview Extension not supported, skipping tests\n");
```
arno-lunarg, line 3353: GTEST_SKIP this?

jeremyg-lunarg: there's additional checking the test that happens after this statement that should run even without multiview extension

juan-lunarg (approved):

PR #4963, opened 2022-12-14
Use DS dynamic state for draw time checks

spencer-lunarg:
- First move all the dynamic state in `ValidationStateTracker` to `PostCallRecord`
- collapse the `EXT` functions to use a single helper function
- For all "at draw time call" check if DepthStencil state was set dynamically (currently we are checking only the static value for most values)
- Note: I want to fix this with other pipeline states as well, but this is a good first step


jeremyg-lunarg (approved):
LGTM / +2

PR #4964, opened 2022-12-14
layers: Remove layer_data::optional typedef

juan-lunarg:


layers/vk_layer_data.h
```diff
@@ -873,14 +870,11 @@ class TlsGuard {
     T *operator->() { return &(*payload_); }
 
   private:
-    thread_local static optional<T> payload_;
+    inline thread_local static std::optional<T> payload_{};
```
juan-lunarg, line 873: C++17 allows you to inline static member variables to avoid defining them in the source file.

https://github.com/AnthonyCalandra/modern-cpp-features/blob/master/CPP17.md#inline-variables

jeremyg-lunarg (approved):
LGTM / +2

ncesario-lunarg (approved):
I'm good with this if others are. Presumably we won't be going backwards in standard versions, and using a non-standard version of `optional` seems unlikely.

Tony-LunarG (approved):
LGTM

Issue #4859, opened 2022-11-24
Adding missing Ray Tracing Access Mask VUs


spencer-lunarg:
**Describe the situation in which you encountered the missing validation**
Missing Ray Tracing access mask VUs

**Valid Usage IDs requested**
// related 06254 06255 06257

- [ ]  VUID-VkMemoryBarrier2-srcAccessMask-06256
- [ ]  VUID-VkMemoryBarrier2-dstAccessMask-06256
- [ ]  VUID-VkBufferMemoryBarrier2-srcAccessMask-06256
- [ ]  VUID-VkBufferMemoryBarrier2-dstAccessMask-06256
- [ ]  VUID-VkImageMemoryBarrier2-srcAccessMask-06256
- [ ]  VUID-VkImageMemoryBarrier2-dstAccessMask-06256

The same 6 permutation for

- VUID-VkMemoryBarrier2-srcAccessMask-06254
- VUID-VkMemoryBarrier2-srcAccessMask-06255
- VUID-VkMemoryBarrier2-srcAccessMask-06257 

**Additional Context**

These are all related, just depending on the extensions enabled (search for the VUID in https://raw.githubusercontent.com/KhronosGroup/Vulkan-Docs/f4eb3e9a7acdef9ab62ac9af954a50409895ac6d/chapters/commonvalidity/access_mask_2_common.adoc)

A simple logic view is

```
if not VK_KHR_ray_query
if (VK_KHR_ray_tracing_pipelin OR VK_NV_ray_tracing)
  * VUID-VkMemoryBarrier2-srcAccessMask-06254

if not VK_KHR_ray_query
if not (VK_KHR_ray_tracing_pipelin OR VK_NV_ray_tracing)
* VUID-VkMemoryBarrier2-srcAccessMask-06255

if not VK_KHR_ray_query
if (VK_KHR_ray_tracing_pipelin OR VK_NV_ray_tracing)
* VUID-VkMemoryBarrier2-srcAccessMask-06256

if not VK_KHR_ray_query
if not (VK_KHR_ray_tracing_pipelin OR VK_NV_ray_tracing)
* VUID-VkMemoryBarrier2-srcAccessMask-06257
```

A related VUID to look at for how to implement is `VUID-VkMemoryBarrier2-dstAccessMask-03927`

arno-lunarg:
- VUID-VkMemoryBarrier2-srcAccessMask-06254
[...]

Issue #2832, opened 2021-05-23
Wrong validation of VK_PIPELINE_STAGE_RAY_TRACING_SHADER_BIT_KHR vs VK_ACCESS_ACCELERATION_STRUCTURE_READ_BIT_KHR


danielmaciel:
This validation is wrong:

Validation Error: [ VUID-vkCmdPipelineBarrier-dstAccessMask-02816 ] Object 0: handle = 0x2113620, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x69c8467d | vkCmdPipelineBarrier(): .pMemoryBarriers[0].dstAccessMask bit VK_ACCESS_ACCELERATION_STRUCTURE_READ_BIT_KHR is not supported by stage mask (VK_PIPELINE_STAGE_RAY_TRACING_SHADER_BIT_KHR). The Vulkan spec states: The dstAccessMask member of each element of pMemoryBarriers must only include access flags that are supported by one or more of the pipeline stages in dstStageMask, as specified in the table of supported access types (https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VUID-vkCmdPipelineBarrier-dstAccessMask-02816)

According to the spec, VK_ACCESS_ACCELERATION_STRUCTURE_READ_BIT_KHR is supported by VK_PIPELINE_STAGE_RAY_TRACING_SHADER_BIT_KHR

See https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#synchronization-access-types-supported

![image](https://user-images.githubusercontent.com/2924812/119255020-c269f580-bbb9-11eb-8987-0a9b2a7dabff.png)

jeremyg-lunarg:
This should be fixed by https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/2824 once it merges.

jeremyg-lunarg:
#2824 is now merged

JoseEmilio-ARM:
Hi @jeremy-lunarg , I'm using the layers from SDK 1.2.182.0 which should include this fix, however I am seeing a similar issue, in my case with the combination of `VK_ACCESS_ACCELERATION_STRUCTURE_READ_BIT_KHR` and `VK_PIPELINE_STAGE_FRAGMENT_SHADER_BIT` which seems to be supported in the spec as shown above:

[...]

PR #4404, opened 2022-08-11
Add AS Read access flag VUs

sjfricke:
adds 
- `VUID-VkMemoryBarrier2-*-06254` 
- `VUID-VkMemoryBarrier2-*-06255`
- `VUID-VkMemoryBarrier2-*-06256`
- `VUID-VkMemoryBarrier2-*-06257`

These are all making use of the `VK_KHR_acceleration_structure`/`VK_KHR_ray_tracing_pipeline`/`VK_KHR_ray_query`

----

Spec for the [fun 4 way VUID](https://github.com/KhronosGroup/Vulkan-Docs/blob/8dcc7469b01529600b712596a5a48ec8c710e228/chapters/commonvalidity/access_mask_2_common.txt#L215-L250) but here is the cheat sheet

```
With `VK_KHR_ray_tracing_pipeline` and Without `VK_KHR_ray_query`
[[VUID-{refpage}-{accessMaskName}-06254]]

Without `VK_KHR_ray_tracing_pipeline` and Without `VK_KHR_ray_query`
[[VUID-{refpage}-{accessMaskName}-06255]]

With `VK_KHR_ray_tracing_pipeline` and With `VK_KHR_ray_query`
[[VUID-{refpage}-{accessMaskName}-06256]]

Without `VK_KHR_ray_tracing_pipeline` and With `VK_KHR_ray_query`
[[VUID-{refpage}-{accessMaskName}-06257]]
```

PR #4976, opened 2022-12-19
Add new barrier access masks VUs

arno-lunarg:
# Issue:
- Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4859
- Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/2832 

# Summary:
## 3 x 2 x 4 VUIDs

This PR adds 24 VUIDs. They essentially check for the same thing, but the check can be performed on different struct members (3 structs, and 2 struct members), and given the same same struct member, 4 different VUIDs are defined, the one that must be used depends on whether or not `VK_KHR_ray_query`, `VK_KHR_ray_tracing_pipeline` or `VK_NV_ray_tracing` are enabled. The blueprint that has been used can be found in the doc repo, in `commonvalidity/access_mask_2_common.adoc`:
```
ifndef::VK_KHR_ray_query[]
ifdef::VK_KHR_ray_tracing_pipeline,VK_NV_ray_tracing[]
  * [[VUID-{refpage}-{accessMaskName}-06254]]
    If pname:{accessMaskName} includes
    ename:VK_ACCESS_2_ACCELERATION_STRUCTURE_READ_BIT_KHR,
    pname:{stageMaskName} must: not include any of the
    etext:VK_PIPELINE_STAGE_*_SHADER_BIT stages except
    ename:VK_PIPELINE_STAGE_2_RAY_TRACING_SHADER_BIT_KHR
endif::VK_KHR_ray_tracing_pipeline,VK_NV_ray_tracing[]
ifndef::VK_KHR_ray_tracing_pipeline,VK_NV_ray_tracing[]
  * [[VUID-{refpage}-{accessMaskName}-06255]]
    If pname:{accessMaskName} includes
    ename:VK_ACCESS_2_ACCELERATION_STRUCTURE_READ_BIT_KHR,
    pname:{stageMaskName} must: not include any of the
    etext:VK_PIPELINE_STAGE_*_SHADER_BIT stages
endif::VK_KHR_ray_tracing_pipeline,VK_NV_ray_tracing[]
endif::VK_KHR_ray_query[]
ifdef::VK_KHR_ray_query[]
ifdef::VK_KHR_ray_tracing_pipeline,VK_NV_ray_tracing[]
  * [[VUID-{refpage}-{accessMaskName}-06256]]
    If the <<features-rayQuery, pname:rayQuery>> feature is not enabled and
    pname:{accessMaskName} includes
    ename:VK_ACCESS_2_ACCELERATION_STRUCTURE_READ_BIT_KHR,
    pname:{stageMaskName} must: not include any of the
    etext:VK_PIPELINE_STAGE_*_SHADER_BIT stages except
    ename:VK_PIPELINE_STAGE_2_RAY_TRACING_SHADER_BIT_KHR
endif::VK_KHR_ray_tracing_pipeline,VK_NV_ray_tracing[]
ifndef::VK_KHR_ray_tracing_pipeline,VK_NV_ray_tracing[]
  * [[VUID-{refpage}-{accessMaskName}-06257]]
    If the <<features-rayQuery, pname:rayQuery>> feature is not enabled and
    pname:{accessMaskName} includes
    ename:VK_ACCESS_2_ACCELERATION_STRUCTURE_READ_BIT_KHR,
    pname:{stageMaskName} must: not include any of the
    etext:VK_PIPELINE_STAGE_*_SHADER_BIT stages
endif::VK_KHR_ray_tracing_pipeline,VK_NV_ray_tracing[]
endif::VK_KHR_ray_query[]
```

## My attempt at keeping things structured

~~Since the validation logic is similar for those VUIDs, there is an opportunity for factorisation. I first looked at how VUIDs from the same family are validated, but unfortunately, I do not think using present functions and update data structures found in `sync_vuid_maps.cpp` is an option, for various reasons: present validation code for VUIDS of this family lives in `CoreChecks::ValidateAccessMask`. The validation logic being the same for all of them, they have been regrouped. They are accessed based on the `srcAccessMask` and `dstAccessMask` members of structures such as `VkMemoryBarrier2`, `VkImageMemoryBarrier2`, `VkBufferMemoryBarrier2` and are stored in a `std::map` with `VK_ACCESS_*` values used as keys.~~
~~This does not let me add new VUIDs for `VK_ACCESS_2_ACCELERATION_STRUCTURE_READ_BIT_KHR`, and besides, the new logic I want to add does not map to what is already here.~~

~~So I added a new `std::map Location2VUID`, mapping a `Location` (e.g. `(Struct::VkMemoryBarrier2, Field::srcAccessMask)`) to a vuid selection functor. Why not directly map to a vuid? Because the actual vuid that will be used, according to the blueprint above, depends on which extensions are enabled. The retrieved functor is used to get the correct vuid and its corresponding validation function given a set of enabled extensions. The code for this is in `sync_vuid_maps.h/cpp`. It is used in `synchronization_validation.cpp`~~
~~This logic allows good factorisation, and should be extensible without it being painful to do.~~

Edit:
Spencer demonstrated me that I could go with a simpler approach, closer to the existing logic (see discussion between Spencer and I below for more details).
To synthesise, a new map as been added, that for the particular validation at hand, will return the correct VUID based on the supplied location and enabled device extensions.


layers/core_validation.cpp
```diff
@@ -11676,29 +11676,127 @@ bool CoreChecks::ValidatePipelineStage(const LogObjectList &objects, const Locat
 bool CoreChecks::ValidateAccessMask(const LogObjectList &objects, const Location &loc, VkQueueFlags queue_flags,
                                     VkAccessFlags2KHR access_mask, VkPipelineStageFlags2KHR stage_mask) const {
     bool skip = false;
+
+    const auto expanded_pipeline_stages = sync_utils::ExpandPipelineStages(stage_mask, queue_flags);  // TODO:
```
arno-lunarg, line 11680: I traced this empty TODO to an old (almost 2 yrs) commit from Jeremy: e60d417b5e8b557af67288753e1e6b6385ad5d52

layers/sync_vuid_maps.cpp
```diff
@@ -661,6 +665,90 @@ const std::string &GetBadAccessFlagsVUID(const Location &loc, VkAccessFlags2KHR
...
+        }
+    }
+}
+
+// commonvalidity/access_mask_2_common.adoc
+static const auto &GetLocation2VUIDMap() {
+    static const std::map<Key, VUIDSelector> Location2VUID{
+        {Key(Struct::VkMemoryBarrier2, Field::srcAccessMask),
+         [](const DeviceExtensions &device_extensions) {
+             using namespace std::string_view_literals;  // for ""sv
```
spencer-lunarg, line 690: what does `// for ""sv` mean?

arno-lunarg: It is an operator, used to create a `std::string_view` over its operand. You can find more here: https://en.cppreference.com/w/cpp/string/basic_string_view/operator%22%22sv

layers/vk_layer_utils.h
```diff
@@ -323,6 +324,16 @@ static inline uint32_t GetBitSetCount(uint32_t field) {
     return static_cast<uint32_t>(view_bits.count());
 }
 
+namespace std {
+
+// A short utility function to handle going from an enumeration to its underlying integral value for safety and ease of use
+template <class T>
+constexpr std::underlying_type_t<T> to_underlying(T value) noexcept {
```
spencer-lunarg, line 331: Is this being used?

arno-lunarg: At first I used it, but not anymore. Still, I think this function could be useful, so I left it. They are a number of spots where we cast an enum to some int type, it is worth considering changing those cast to a call to this function instead. 
FYI `std::to_underlying` will be part of the STL starting with c++ 23.

layers/sync_vuid_maps.cpp
arno-lunarg: Oh ok. Well using `std::string` would mean heap allocations, and why use `const char*` when we can use `std::string_view`?
Overall, we use `std::string` extensively to store vuids, when those strings are never changed once they are initialised, meaning we probably do a lot of useless heap allocations (not sure small string optimisation applies here). Changing `LogError`  to use a `std::string_view` instead of a `std::string` for the vuid is probably worth considering.

juan-lunarg: Agree std::string_view is definitely worth considering as it can avoid copies/heap allocations. That's why it was introduced to the STL.

spencer-lunarg: I'm less opposed to using `string_view` and more of starting it here in this PR. If we want to do this, it would be nice to have a single commit/PR that goes and updates it everywhere, otherwise it becomes lost information and later someone goes

"wait the other things aren't `string_view`? is there is subtle reason why now?"

juan-lunarg: I agree. Introducing `string_view` in this PR isn't the best way to go about it.

arno-lunarg: A "move to `std::string_view`" PR would probably be gigenormous, one would need to check every `char*` and `std::string` usage to see if a `std::string_view` can be used instead ＼（〇_ｏ）／
So I do not see this happening anytime soon...
My opinion is that onward, whenever thinking about using a `char *` or a `std::string`, we should think about possibly using `std::string_view` instead. And tolerate to live with old code that does not do it. For now. 
Probably a topic worth bringing up during our next team meeting!

layers/core_validation.h
```diff
@@ -343,8 +343,25 @@ class CoreChecks : public ValidationStateTracker {
                                              VkPipelineStageFlags2KHR stage_mask) const;
     bool ValidatePipelineStage(const LogObjectList& objects, const Location& loc, VkQueueFlags queue_flags,
                                VkPipelineStageFlags2KHR stage_mask) const;
+    static bool AccessMaskRayQueryEnabledRTXEnabled(const CoreChecks& validator, std::string_view vuid,
```
spencer-lunarg, line 346: What would be the issue of having a single function with 2 boolean passed in instead of 4 functions, these functions are doing mostly the same thing.

layers/sync_vuid_maps.cpp
```diff
@@ -661,6 +665,90 @@ const std::string &GetBadAccessFlagsVUID(const Location &loc, VkAccessFlags2KHR
...
+                 "VUID-VkImageMemoryBarrier2-dstAccessMask-06254"sv, "VUID-VkImageMemoryBarrier2-dstAccessMask-06255"sv,
+                 "VUID-VkImageMemoryBarrier2-dstAccessMask-06256"sv, "VUID-VkImageMemoryBarrier2-dstAccessMask-06257"sv};
+             return WhichAccessMaskRayQueryVuid(device_extensions, vuids);
+         }},
+    };
+    assert(Location2VUID.size() == 6);
+    return Location2VUID;
+}
+
+sync_vuid_maps::VUIDSelector GetAccessMaskVUIDSelector(const Location &loc) {
```
spencer-lunarg, line 743: instead of creating all this extra framework code and having to do things like re-pass in the `CoreCheck::validator` via lamba, why could this not be simplified to something like

```cpp
static const std::vector<Entry> kRayQueryEnabledRTXEnabled = {
    {Key(Struct::VkMemoryBarrier2, Field::srcAccessMask), "VUID-VkMemoryBarrier2-srcAccessMask-06256"},
    {Key(Struct::VkMemoryBarrier2, Field::dstAccessMask), "VUID-VkMemoryBarrier2-dstAccessMask-06256"},
    {Key(Struct::VkBufferMemoryBarrier2, Field::srcAccessMask), "VUID-VkBufferMemoryBarrier2-srcAccessMask-06256"},
    {Key(Struct::VkBufferMemoryBarrier2, Field::dstAccessMask), "VUID-VkBufferMemoryBarrier2-dstAccessMask-06256"},
    {Key(Struct::VkImageMemoryBarrier2, Field::srcAccessMask), "VUID-VkImageMemoryBarrier2-srcAccessMask-06256"},
    {Key(Struct::VkImageMemoryBarrier2, Field::dstAccessMask), "VUID-VkImageMemoryBarrier2-dstAccessMask-06256"},
}
static const std::vector<Entry> kRayQueryEnabledRTXDisabled = {
    {Key(Struct::VkMemoryBarrier2, Field::srcAccessMask), "VUID-VkMemoryBarrier2-srcAccessMask-06257"},
    {Key(Struct::VkMemoryBarrier2, Field::dstAccessMask), "VUID-VkMemoryBarrier2-dstAccessMask-06257"},
    {Key(Struct::VkBufferMemoryBarrier2, Field::srcAccessMask), "VUID-VkBufferMemoryBarrier2-srcAccessMask-06257"},
    {Key(Struct::VkBufferMemoryBarrier2, Field::dstAccessMask), "VUID-VkBufferMemoryBarrier2-dstAccessMask-06257"},
    {Key(Struct::VkImageMemoryBarrier2, Field::srcAccessMask), "VUID-VkImageMemoryBarrier2-srcAccessMask-06257"},
    {Key(Struct::VkImageMemoryBarrier2, Field::dstAccessMask), "VUID-VkImageMemoryBarrier2-dstAccessMask-06257"},
}
static const std::vector<Entry> kRayQueryDisabledRTXEnabled = { /**/}
static const std::vector<Entry> kRayQueryDisabledRTXDisabled = { /**/ }

const std::string &GetThisVUID(const Location &loc) {
    const bool ray_query = IsExtEnabled(device_extensions.vk_khr_ray_query);
    const bool ray_tracing = IsExtEnabled(device_extensions.vk_khr_ray_tracing_pipeline) || IsExtEnabled(device_extensions.vk_nv_ray_tracing);
    
    const Table& table = 
    (ray_query & ray_tracing) ? kRayQueryEnabledRTXEnabled :
    (ray_query & !ray_tracing) ? kRayQueryEnabledRTXDisabled :
    (!ray_query & ray_tracing) ? kRayQueryDisabledRTXEnabled :
    kRayQueryDisabledRTXDisabled;

    const auto &result = FindVUID(loc, table);
    if (result.empty()) {
        // ...
    }
    return result;
}
```

arno-lunarg: I like your approach because it fits better to the system we already have.
But what about the actual "vuid check" code? In my case, I tie a vuid to its validation logic, so when I get the vuid, I just need to call its associated check function without adding any logic. 
Also with you example, I am not sure it does scale well: there are already 4 maps, with the `Key`s being the same each time. If more vuids are to come, this could be problematic. And the code to select the table would need to be revisited.

I went for a more functional approach, it should scale better. I definitely agree on the fact that what I propose is unusual, but I think the possibilities of this new approach offers are worth it. Or at least worth considering!

layers/core_validation.h
arno-lunarg: I thought about that too, but writing it, I found my code to be messy, so I went for 4 functions instead. But I will try to do a better job and go for a single function.

ncesario-lunarg (commented):
A lot to think about here! Let's wait for after the holidays to get Jeremy's input on this (if he's got some spare cycles, it might be worth getting some input from Zulauf for further design perspective).

layers/sync_vuid_maps.cpp
```diff
@@ -661,6 +665,90 @@ const std::string &GetBadAccessFlagsVUID(const Location &loc, VkAccessFlags2KHR
     return result2;
 }
 
+static auto WhichAccessMaskRayQueryVuid(const DeviceExtensions &device_extensions, std::array<std::string_view, 4> vuids) {
+    // This logic is based on VUIDs 06254, 06255, 06256 and 06257 from commonvalidity/access_mask_2_common.adoc
+    if (!IsExtEnabled(device_extensions.vk_khr_ray_query)) {
+        if (IsExtEnabled(device_extensions.vk_khr_ray_tracing_pipeline) || IsExtEnabled(device_extensions.vk_nv_ray_tracing)) {
+            return std::make_pair(vuids[0], CoreChecks::AccessMaskRayQueryDisabledRTXEnabled);
```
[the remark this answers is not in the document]
arno-lunarg: yes yes using a struct would be better

layers/core_validation.cpp
jeremyg-lunarg: No..  😵‍💫

tests/vklayertests_ray_tracing.cpp
```diff
@@ -80,6 +80,322 @@ TEST_F(VkLayerTest, RayTracingTestBarrierAccessAccelerationStructure) {
...
+    image.Init(128, 128, 1, VK_FORMAT_B8G8R8A8_UNORM, VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT, VK_IMAGE_TILING_OPTIMAL, 0);
+    ASSERT_TRUE(image.initialized());
+
+    auto image_barrier = LvlInitStruct<VkImageMemoryBarrier2>();
+    image_barrier.dstAccessMask = VK_ACCESS_2_SHADER_READ_BIT;
+    image_barrier.dstStageMask = VK_PIPELINE_STAGE_2_FRAGMENT_SHADER_BIT;
+    image_barrier.image = image.handle();
+    image_barrier.subresourceRange = {VK_IMAGE_ASPECT_COLOR_BIT, 0, 1, 0, 1};
+
+    VkDependencyInfo dependency_info = LvlInitStruct<VkDependencyInfo>();
```
jeremyg-lunarg, line 203: auto dependency_info =

tests/vklayertests_ray_tracing.cpp
```diff
@@ -80,6 +80,322 @@ TEST_F(VkLayerTest, RayTracingTestBarrierAccessAccelerationStructure) {
...
```
jeremyg-lunarg, line 126: auto dependency_info =

tests/vklayertests_ray_tracing.cpp
```diff
@@ -80,6 +80,322 @@ TEST_F(VkLayerTest, RayTracingTestBarrierAccessAccelerationStructure) {
...
```
jeremyg-lunarg, line 284: here too

tests/vklayertests_ray_tracing.cpp
```diff
@@ -80,6 +80,322 @@ TEST_F(VkLayerTest, RayTracingTestBarrierAccessAccelerationStructure) {
...
```
jeremyg-lunarg, line 364: and here

tests/positive/ray_tracing.cpp
```diff
@@ -204,3 +204,163 @@ TEST_F(VkPositiveLayerTest, RayTracingAccelerationStructureReference) {
...
+    image.Init(128, 128, 1, VK_FORMAT_B8G8R8A8_UNORM, VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT, VK_IMAGE_TILING_OPTIMAL, 0);
+    ASSERT_TRUE(image.initialized());
+
+    auto image_barrier = LvlInitStruct<VkImageMemoryBarrier2>();
+    image_barrier.dstAccessMask = VK_ACCESS_2_SHADER_READ_BIT;
+    image_barrier.dstStageMask = VK_PIPELINE_STAGE_2_FRAGMENT_SHADER_BIT;
+    image_barrier.image = image.handle();
+    image_barrier.subresourceRange = {VK_IMAGE_ASPECT_COLOR_BIT, 0, 1, 0, 1};
+
+    VkDependencyInfo dependency_info = LvlInitStruct<VkDependencyInfo>();
```
jeremyg-lunarg, line 338: auto

tests/positive/ray_tracing.cpp
```diff
@@ -204,3 +204,163 @@ TEST_F(VkPositiveLayerTest, RayTracingAccelerationStructureReference) {
...
```
jeremyg-lunarg, line 258: auto

layers/vk_layer_utils.h
jeremyg-lunarg: works for me. Should it be in a namespace though?

layers/core_validation.cpp
```diff
@@ -11676,29 +11676,127 @@ bool CoreChecks::ValidatePipelineStage(const LogObjectList &objects, const Locat
...
+    bool skip = false;
+
+    if (!validator.enabled_features.ray_query_features.rayQuery &&
+        (access_mask & VK_ACCESS_2_ACCELERATION_STRUCTURE_READ_BIT_KHR)) {
+        const auto illegal_pipeline_stages = allVkPipelineShaderStageBits2;
+        if (stage_mask & illegal_pipeline_stages) {
+            const auto stages_string = sync_utils::StringPipelineStageFlags(stage_mask);
+            std::stringstream msg;
+            msg << loc.Message() << " contains pipeline stages " << stages_string
+                << ", when only VK_PIPELINE_STAGE_2_RAY_TRACING_SHADER_BIT_KHR is expected.";
```
jeremyg-lunarg, line 11753: Is this error message correct? You're not masking off RAY_TRACING_SHADER_BIT like in the other functions.

jeremyg-lunarg (commented):
This is a great approach and long overdue. The RTX features and state/access bits have been improperly handled for quite some time.

jeremyg-lunarg (commented):
For the future, I definitely think it is worth looking at using std::string_view with VUIDs if possible. Also, we really need to figure out how to consolidate all of our VUID lookup code somehow and hopefully automate more of it, if possible.

layers/core_validation.cpp
arno-lunarg: No it is not, failed copy/paste... ty!

layers/vk_layer_utils.h
arno-lunarg: I think putting it in the namespace `std` is ok because this function will be declared there eventually. And seeing such a function, I would go "oh this function is declared in the `std` namespace, so if need be I would probably be able to search doc on it online".

jeremyg-lunarg: If you're going to put it in std:: you should guard it with the feature test macro [__cpp_lib_to_underlying](https://en.cppreference.com/w/cpp/feature_test#Library_features)

layers/core_checks/core_validation.h
```diff
@@ -473,8 +473,25 @@ class CoreChecks : public ValidationStateTracker {
                                              VkPipelineStageFlags2KHR stage_mask) const;
     bool ValidatePipelineStage(const LogObjectList& objlist, const Location& loc, VkQueueFlags queue_flags,
                                VkPipelineStageFlags2KHR stage_mask) const;
+    static bool AccessMaskRayQueryEnabledRTXEnabled(const CoreChecks& validator, std::string_view vuid,
```
spencer-lunarg, line 476: I still don't see why we need these CoreChecks functions, we have been doing the VUID logic all in the `sync_vuid_maps`

I would expect something along the lines of https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4404/files

arno-lunarg: Yes it would be better to have them elsewhere since they are only referenced in `sync_vuid_maps`. Those functions represent the validation logic, and there is no validation logic in `sync_vuid_maps`, only a mapping, that's why I ended up with those functions in `core_validation.h` : /

arno-lunarg: I could create a new file and put them here, would it be ok ? I do not want to break the current architecture

spencer-lunarg: Actually, what you have here I think is the exact problem @artem-lunarg is doing in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5721 (The whole, if-else extension for which VUID)

arno-lunarg: Thinking about it, for now I think leaving those functions in `sync_vuid_maps.cpp` is ok, as they are only referenced here. I pushed the change, I will see how people react to that.

arno-lunarg: In the end I decided to create new files, `cc_access_mask.h/cpp`, to put those new functions

layers/core_checks/cc_access_mask.cpp
```diff
@@ -0,0 +1,101 @@
...
+            msg << loc.Message() << " contains pipeline stages " << stages_string
+                << ", when only VK_PIPELINE_STAGE_2_RAY_TRACING_SHADER_BIT_KHR is expected.";
+            skip |= validator.LogError(objects, vuid.data(), "%s", msg.str().c_str());
+        }
+    }
+
+    return skip;
+}
+
+bool AccessMaskRayQueryEnabledRTXDisabled(const CoreChecks &validator, std::string_view vuid, const LogObjectList &objects,
```
spencer-lunarg, line 45: I still not sure why we need 4 functions here, this can just be a single check

We have stateless validation to make sure `VK_PIPELINE_STAGE_2_RAY_TRACING_SHADER_BIT_KHR` is never used if the extension is enabled, so for all of these it can be

```cpp
const auto illegal_pipeline_stages = allVkPipelineShaderStageBits2 & ~VK_PIPELINE_STAGE_2_RAY_TRACING_SHADER_BIT_KHR;
```

Also all 4 of these can be

```
if (!rayQuery) { 

}
```

so with that this is a single check

Basically you are not making use of the fact features/flags/enums will not used if the extension is not enabled

arno-lunarg: Ah yes I had not thought about this! But now with 1 function having a new file could be overkill... But if do not do that, I do not know where to put the definition. `core_validation.h` and `sync_validation.h` do not seem to be great spots, so having new tiny files does not seem that bad?

spencer-lunarg: well I feel the

```
WhichAccessMaskRayQueryVuid(device_extensions, vuids0625X);
```

only job should be to get the VUID, not validate, so that can just live in `sync_vuid_maps.cpp`

The real fundamental issue is your call to `sync_vuid_maps::GetAccessMaskVUIDSelector(loc);` is trying to get the VUID **and** validate

I would just do your validation logic in `cc_synchronization.cpp` and then  if invalid, call `GetAccessMaskVUIDSelector` to figure out the VUID

arno-lunarg: So I killed the 4 validation functions, now there is only one.

Regarding your last comment:
If I do not have a way to tie a VUID and its validation function, I am going to end up with something like this:
```cpp
// Do my validation...
if (!validator.enabled_features.ray_query_features.rayQuery &&
        (access_mask & VK_ACCESS_2_ACCELERATION_STRUCTURE_READ_BIT_KHR)) {
    const auto illegal_pipeline_stages = allVkPipelineShaderStageBits2 & ~VK_PIPELINE_STAGE_2_RAY_TRACING_SHADER_BIT_KHR;
    if (stage_mask & illegal_pipeline_stages) {


        // ... then get a vuid based on the location to check and enabled extensions
        const char *vuid = sync_vuid_maps::GetAccessMaskVUID(location, device_extensions);
        const auto stages_string = sync_utils::StringPipelineStageFlags(stage_mask);            
        std::stringstream msg;
        msg << location.Message() << " contains pipeline stages " << stages_string << '.';
        skip |= validator.LogError(objlist, vuid, "%s", msg.str().c_str());
    }
}
```

The problem here is that the call to `sync_vuid_maps::GetAccessMaskVUID(loc, device_extensions)` implicitly is tied to the specific validation here. If a new VUID is added, for the same `location`, extending the "get the right VUID for this location" logic will be a mess.
I think the way I propose to do things, tying a VUID and its validation, is a good solution to this problem.

(Am I clear ^^" ?)

spencer-lunarg: >  is tied to this specific validation

Yes, all of these `sync_vuid_maps` are tied to a single Validation check

I guess what is wrong with how we do things with `sync_vuid_maps::GetBadFeatureVUID(`

-----

> I think the way I propose to do things, tying a VUID and its validation, is a good solution to this problem.

We have not done it anywhere else, not necessarily saying this idea is bad, but it does basically force us to break EVERY check into its own small function (basically how the codified VUs logic would work too)

The main issue with this is in the future, if a new/separate VU is added that looks like

> If srcAccessMask includes VK_ACCESS_2_ACCELERATION_STRUCTURE_READ_BIT_KHR, **and dstAccessMask include VK_ACCESS_2_ACCELERATION_STRUCTURE_WRITE_BIT_KHR**,  srcStageMask must not include any of the VK_PIPELINE_STAGE_*_SHADER_BIT stages except VK_PIPELINE_STAGE_2_RAY_TRACING_SHADER_BIT_KHR,

We have no way to reuse the logic, and would have to create another function to do the same base check as this one

arno-lunarg: > Yes, all of these sync_vuid_maps are tied to a single Validation check

Ah ok! Validation of those access masks is definitely something... special ^^

I made the modifications, now validation is basically this: [`0d07d33` (#4976)](https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4976/commits/0d07d335aa4793d861e79163f5317b256479155b#diff-94676d5278f6629f3235f8d5d5a989b701bd6d384b79cea74631be1bd6fc8fc5R991)

layers/utils/vk_layer_utils.h
```diff
@@ -521,6 +522,17 @@ constexpr uint32_t ResolveRemainingLevels(const VkImageCreateInfo &ci, VkImageSu
 constexpr uint32_t ResolveRemainingLayers(const VkImageCreateInfo &ci, VkImageSubresourceRange const &range) {
     return (range.layerCount == VK_REMAINING_ARRAY_LAYERS) ? (ci.arrayLayers - range.baseArrayLayer) : range.layerCount;
 }
+#ifndef __cpp_lib_to_underlying
+namespace std {
+
+// A short utility function to handle going from an enumeration to its underlying integral value for safety and ease of use
+template <class T>
+constexpr std::underlying_type_t<T> to_underlying(T value) noexcept {
```
spencer-lunarg, line 530: is anyone using this?

spencer-lunarg (approved):
sorry this keeps getting forgotten, I think this looks good now (pending `to_underlying` question)

just need to rebase

spencer-lunarg (approved):
sorry this keeps getting forgotten, I think this looks good now (pending `to_underlying` question)

just need to rebase

layers/utils/vk_layer_utils.h
arno-lunarg: no, I was thinking about adding it anyway, but I am going to remove it. 
I remember seeing some spots where this could be used instead of manually casting some enum value to an integer type, but I cannot find those anymore. Will add it back only if it proves useful

juan-lunarg (approved):

PR #4978, opened 2022-12-19
layers: Validate framebuffer when using gl_Layer

### LINKED ISSUE #3019: Writing to gl_Layer with invalid framebuffer layers
Issue description:
I am writing to gl_Layer from within a geometry shader to draw points in a cubemap, while my framebuffers were configured with 1 layer instead of 6. 

It was working flawlessly on an AMD GPU, but was only outputting on a single layer (0) on an NVIDIA GPU. 
Changing the framebuffer to use 6 layers fixed it on the NVIDIA. 

I think this should have given off a validation error.

juan-lunarg:
closes #3019

### DIFF
--- a/layers/drawdispatch.cpp
+++ b/layers/drawdispatch.cpp
@@ -1801,6 +1801,61 @@ const DrawDispatchVuid &CoreChecks::GetDrawDispatchVuid(CMD_TYPE cmd_type) const
     }
 }
 
+bool CoreChecks::ValidateCmdDrawFramebuffer(const CMD_BUFFER_STATE &cb_state, const char *caller) const {
+    const LvlBindPoint lv_bind_point = ConvertToLvlBindPoint(VK_PIPELINE_BIND_POINT_GRAPHICS);
+    const PIPELINE_STATE *pipeline_state = cb_state.lastBound[lv_bind_point].pipeline_state;
+    const std::shared_ptr<RENDER_PASS_STATE> &render_pass_state = cb_state.activeRenderPass;
+    if (!pipeline_state || !render_pass_state) {
+        return false;
+    }
+
+    // Nothing to validate if there is no geometry shader
+    // Only the geometry shader can write to gl_Layer
+    if (!pipeline_state->HasShaderStage(VK_SHADER_STAGE_GEOMETRY_BIT)) {
+        return false;
+    }
+
+    // Nothing to validate if we don't write to gl_Layer
+    if (!pipeline_state->WritesToGLLayer()) {
+        return false;
+    }
+
+    bool skip = false;
+
+    const safe_VkSubpassDescription2 *active_subpass_desc = &render_pass_state->createInfo.pSubpasses[cb_state.activeSubpass];
+    const safe_VkAttachmentReference2 *p_color_attachments = active_subpass_desc->pColorAttachments;
+    const uint32_t color_attachment_count = active_subpass_desc->colorAttachmentCount;
+    const std::shared_ptr<FRAMEBUFFER_STATE> &active_frame_buffer = cb_state.activeFramebuffer;
+
+    // Iterate over the output locations in the fragment shader, and find the framebuffer we are rendering to.
+    // Then check if that framebuffer is a multi-layer framebuffer. If not, then log the error.
+    for (const uint32_t output_location : pipeline_state->fragmentShader_writable_output_location_list) {
+        if (output_location >= color_attachment_count) {
+            continue;
+        }
+
+        // attachment is either an integer value identifying an attachment at the corresponding index in
+        // VkRenderPassCreateInfo::pAttachments, or VK_ATTACHMENT_UNUSED to signify that this attachment is not used.
+        const uint32_t attachment = p_color_attachments[output_location].attachment;
+
+        if (attachment == VK_ATTACHMENT_UNUSED) {
+            continue;
+        }
+
+        const bool is_multi_layer = active_frame_buffer->IsMultiLayerAttachment(attachment);
+
+        // "gl_Layer is used to select a specific layer (or face and layer of cube map) in a multi-layer framebuffer attachment."
+        if (!is_multi_layer) {
+            skip |= LogError(cb_state.Handle(), "UNASSIGNED-CoreValidation-Shader-BuiltInLayerIncompatibleFramebuffer",
+                             "%s: bound pipeline has a geometry shader that write to gl_Layer, but currently bound framebuffer "
+                             "is not a multi-layer framebuffer attachment.",
+                             caller);
+        }
+    }
+
+    return skip;
+}
+
 // Generic function to handle validation for all CmdDraw* type functions
 bool CoreChecks::ValidateCmdDrawType(const CMD_BUFFER_STATE &cb_state, bool indexed, VkPipelineBindPoint bind_point,
                                      CMD_TYPE cmd_type) const {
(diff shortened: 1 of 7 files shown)


juan-lunarg:
Here is the official doc for [gl_Layer](https://registry.khronos.org/OpenGL-Refpages/gl4/html/gl_Layer.xhtml) for reference.

As stated in the doc: "In the geometry language, gl_Layer is used to select a specific layer (or face and layer of cube map) in a multi-layer framebuffer attachment."

Which implies it shouldn't be legal to use gl_Layer when using a single-layer framebuffer attachment. Which aligns with what was originally reported in https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/3019

layers/pipeline_state.h
```diff
@@ -193,12 +193,14 @@ class PIPELINE_STATE : public BASE_NODE {
...
-                return true;
-            }
-        }
-        return false;
+        return std::any_of(stage_state.begin(), stage_state.end(),
+                           [&stage_flag](const PipelineStageState &stage) { return (stage.stage_flag == stage_flag); });
+    }
+
+    // Returns true if any of the shaders writes to `gl_layer`
+    bool WritesToGLLayer() const {
```
juan-lunarg, line 201: Wasn't sure how to handle the function name. Would `WriteTo_gl_layer()` be preferred?

spencer-lunarg: I don't see a need for this to be a function, I can't imagine it ever being reused

layers/drawdispatch.cpp
```diff
@@ -1801,6 +1801,61 @@ const DrawDispatchVuid &CoreChecks::GetDrawDispatchVuid(CMD_TYPE cmd_type) const
...
+bool CoreChecks::ValidateCmdDrawFramebuffer(const CMD_BUFFER_STATE &cb_state, const char *caller) const {
+    const LvlBindPoint lv_bind_point = ConvertToLvlBindPoint(VK_PIPELINE_BIND_POINT_GRAPHICS);
+    const PIPELINE_STATE *pipeline_state = cb_state.lastBound[lv_bind_point].pipeline_state;
+    const std::shared_ptr<RENDER_PASS_STATE> &render_pass_state = cb_state.activeRenderPass;
+    if (!pipeline_state || !render_pass_state) {
+        return false;
+    }
+
+    // Nothing to validate if there is no geometry shader
+    // Only the geometry shader can write to gl_Layer
```
spencer-lunarg, line 1813: This isn't true

> VUID-Layer-Layer-04272
> 
> The Layer decoration must be used only within the MeshEXT, MeshNV, Vertex, TessellationEvaluation, Geometry, or Fragment Execution Model

I think a proper VUID is needed here, curious if this is actually limited to Geometry

juan-lunarg: Well for fragment shaders it's a read operation not a write operation (If I'm reading the docs correctly). As for the Mesh/Vertex/Tesselation I had no idea based off the documentation I found (https://registry.khronos.org/OpenGL-Refpages/gl4/html/gl_Layer.xhtml).

ncesario-lunarg: I think SPIR-v allows layer to be written in any "pre-raster stage," and only GLSL restricts writing to the geometry stage, but I'm not 100% sure.

ncesario-lunarg: https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_builtin

Looks like vertex, tessellation, or geometry stages can write the layer; I don't see any mention of mesh shaders, but the documentation does say "See the client API specification for more detail."

layers/drawdispatch.cpp
```diff
@@ -1801,6 +1801,61 @@ const DrawDispatchVuid &CoreChecks::GetDrawDispatchVuid(CMD_TYPE cmd_type) const
...
+    if (!pipeline_state->HasShaderStage(VK_SHADER_STAGE_GEOMETRY_BIT)) {
```
ncesario-lunarg, line 1814: Based on Spencer's comment this needs to be updated to include vertex, tess, and mesh stages.

layers/drawdispatch.cpp
```diff
@@ -1801,6 +1801,61 @@ const DrawDispatchVuid &CoreChecks::GetDrawDispatchVuid(CMD_TYPE cmd_type) const
...
+
+    // Nothing to validate if we don't write to gl_Layer
+    if (!pipeline_state->WritesToGLLayer()) {
+        return false;
+    }
+
+    bool skip = false;
+
+    const safe_VkSubpassDescription2 *active_subpass_desc = &render_pass_state->createInfo.pSubpasses[cb_state.activeSubpass];
+    const safe_VkAttachmentReference2 *p_color_attachments = active_subpass_desc->pColorAttachments;
```
ncesario-lunarg, line 1826: I don't think the `p_` "hungarian-like notation" is consistent with code elsewhere. These prefixes are usually only used when referencing variable names coming directly from the vulkan API.

layers/pipeline_state.h
ncesario-lunarg: `WritesToGLLayer` is clearer than `std::any_of(stage_state.begin(), stage_state.end(), [](const PipelineStageState &stage) { return stage.writes_to_gl_layer; })` IMHO, but I don't have a particularly strong feeling either way.

layers/drawdispatch.cpp
```diff
@@ -1801,6 +1801,61 @@ const DrawDispatchVuid &CoreChecks::GetDrawDispatchVuid(CMD_TYPE cmd_type) const
...
+
+        // attachment is either an integer value identifying an attachment at the corresponding index in
+        // VkRenderPassCreateInfo::pAttachments, or VK_ATTACHMENT_UNUSED to signify that this attachment is not used.
+        const uint32_t attachment = p_color_attachments[output_location].attachment;
+
+        if (attachment == VK_ATTACHMENT_UNUSED) {
+            continue;
+        }
+
+        const bool is_multi_layer = active_frame_buffer->IsMultiLayerAttachment(attachment);
```
ncesario-lunarg, line 1845: So if I understand correctly, this isn't _really_ catching the error. i.e., a shader stage could be write the value `x` to `gl_Layer`, but the attachment only has `y` layers with `x > y`? If so, we need GPUAV to do this properly.

tests/vklayertests_pipeline_shader.cpp
```diff
@@ -16810,3 +16810,51 @@ TEST_F(VkLayerTest, MeshShaderNVRuntimeSpirv) {
...
 }
+
+TEST_F(VkLayerTest, WritingToGLLayerWithIncompatibleFramebuffer) {
+    TEST_DESCRIPTION("Validate gl_Layer is being used only when framebuffer attachment is multi-layer");
+
+    ASSERT_NO_FATAL_FAILURE(Init());
+
+    // Because we are using gl_Layer without a multi-layer framebuffer, there is no telling how an actual
+    // GPU driver will respond. It could handle it gracefully but there is no guarantee.
+    if (!IsPlatform(kMockICD)) {
```
ncesario-lunarg, line 16821: If the error is triggered properly, the draw should never get submitted to the driver. Seems like this is unnecessarily limiting.

ncesario-lunarg (changes requested):
Looks good! I think the feedback from Spencer is necessary, but I'm also curious if my understanding of the problem is correct (see GPUAV related comment). If it is, we should make it clear where the VUID is triggered and in the issue that this is only an approximation (maybe that was made clear and I just missed it).

juan-lunarg:
"actually I am going to raise an issue with the WG on this one... I have found we have wasted a lot of time in the past trying to add UNASSIGNED VUID support to only have a VU change and be different then what it should be

also VK_EXT_shader_viewport_index_layer and VK_NV_geometry_shader_passthrough is going to add wrinkles to this"
- @spencer-lunarg 

I agree. I didn't consider the other shader stages when I made this PR, I also am unsure how robust my solution is altogether.

Given the small length of this PR. It wouldn't be hard to redo in the future except incorporating the feedback here.

PR #4982, opened 2022-12-20
Add best practices VU for useless secondary command buffer

aqnuep:
This change adds a new VU to the best practices layer that detects `vkAllocateCommandBuffers` calls trying to allocate secondary command buffers for queue families that do not support the `vkCmdExecuteCommands` command (queues that do not support graphics, compute, or transfer) and reports a warning about the allocated command buffer not being submittable.

### DIFF
--- a/layers/best_practices_utils.cpp
+++ b/layers/best_practices_utils.cpp
@@ -1486,6 +1486,28 @@ bool BestPractices::PreCallValidateCreateCommandPool(VkDevice device, const VkCo
     return skip;
 }
 
+bool BestPractices::PreCallValidateAllocateCommandBuffers(VkDevice device, const VkCommandBufferAllocateInfo* pAllocateInfo,
+                                                          VkCommandBuffer* pCommandBuffers) const {
+    bool skip = false;
+
+    auto cp_state = Get<COMMAND_POOL_STATE>(pAllocateInfo->commandPool);
+    if (!cp_state) return false;
+
+    const VkQueueFlags queue_flags = physical_device_state->queue_family_properties[cp_state->queueFamilyIndex].queueFlags;
+    const VkQueueFlags sec_cmd_buf_queue_flags = VK_QUEUE_GRAPHICS_BIT | VK_QUEUE_COMPUTE_BIT | VK_QUEUE_TRANSFER_BIT;
+
+    if (pAllocateInfo->level == VK_COMMAND_BUFFER_LEVEL_SECONDARY && (queue_flags & sec_cmd_buf_queue_flags) == 0) {
+        skip |= LogWarning(device, kVUID_BestPractices_AllocateCommandBuffers_UnusableSecondary,
+                           "vkAllocateCommandBuffer(): Allocating secondary level command buffer from command pool "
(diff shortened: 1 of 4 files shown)


layers/best_practices_utils.cpp
```diff
@@ -1486,6 +1486,28 @@ bool BestPractices::PreCallValidateCreateCommandPool(VkDevice device, const VkCo
...
+
+    const VkQueueFlags queue_flags = physical_device_state->queue_family_properties[cp_state->queueFamilyIndex].queueFlags;
+    const VkQueueFlags sec_cmd_buf_queue_flags = VK_QUEUE_GRAPHICS_BIT | VK_QUEUE_COMPUTE_BIT | VK_QUEUE_TRANSFER_BIT;
+
+    if (pAllocateInfo->level == VK_COMMAND_BUFFER_LEVEL_SECONDARY && (queue_flags & sec_cmd_buf_queue_flags) == 0) {
+        skip |= LogWarning(device, kVUID_BestPractices_AllocateCommandBuffers_UnusableSecondary,
+                           "vkAllocateCommandBuffer(): Allocating secondary level command buffer from command pool "
+                           "created against queue family #%u (queue flags: %s), but vkCmdExecuteCommands() is only "
+                           "supported on queue families supporting VK_QUEUE_GRAPHICS_BIT, VK_QUEUE_COMPUTE_BIT, or "
+                           "VK_QUEUE_TRANSFER_BIT. The allocated command buffer will not be submittable.",
```
ncesario-lunarg, line 1504: This seems like an error. Is there no VUID for this?

tests/vklayertests_best_practices.cpp
```diff
@@ -569,6 +569,52 @@ TEST_F(VkBestPracticesLayerTest, CommandBufferReset) {
     m_errorMonitor->VerifyFound();
 }
 
+TEST_F(VkBestPracticesLayerTest, SecondaryCommandBuffer) {
+    TEST_DESCRIPTION("Test for validating usage of vkCreateCommandPool with VK_COMMAND_BUFFER_LEVEL_SECONDARY");
+
+    InitBestPracticesFramework();
+    InitState();
+
+    m_errorMonitor->SetDesiredFailureMsg(kWarningBit, "UNASSIGNED-BestPractices-vkAllocateCommandBuffers-unusable-secondary");
```
ncesario-lunarg, line 578: This should appear right before the triggering call (i.e., `vk::AllocateCommandBuffers` on line 610) to 1) help clarify where the error occurs and 2) avoid errors occurring at unexpected/incorrect times.

ncesario-lunarg (commented):
One question and a minor nit, but otherwise LGTM!

layers/best_practices_utils.cpp
aqnuep: There is no VUID for this and adding one would have been a backward incompatible change, hence it was decided that this is best handled as a best practices VU, especially as implementations already handle this gracefully.

tests/vklayertests_best_practices.cpp
aqnuep: Fair point. I fixed this.

ncesario-lunarg (approved):

Issue #4768, opened 2022-11-08
VK_EXT_graphics_pipeline_library zero binding descriptor set layouts - possible bug


danginsburg:
**Describe the Issue**
I have been debugging a VK_ERROR_DEVICE_LOST on NVIDIA 526.47 GTX 1080 with VK_EXT_graphics_pipeline_library.  I hit a case that causes a device lost and worked with NVIDIA to find a solution.  The specific case is as follows - I have a VS pipeline library and FS pipeline library both created with:

* A pipeline layout containing one descriptor set layout 
* Descriptor set layout 0 has many descriptors in it (UBOs, Samplers, and Sampled Images)
* The VS/FS don't actually consume any descriptors.

At vkCmdBindDescriptorSets time I bind the pipeline layout that was used to create the linked pipeline library along with a descriptor set that has zero bindings.  This triggers the hang.  It does not trigger a validation error because of this code in validation:

```
   if (layout_node && (descriptor_set.GetBindingCount() > 0) && (layout_node->GetBindingCount() > 0)) {
        return VerifySetLayoutCompatibility(*layout_node, *descriptor_set.GetLayout(), errorMsg);
    } else {
        // It's possible the DSL is null when creating a graphics pipeline library, in which case we can't verify compatibility
        // here.
        return true;
    }
```

If instead of binding the descriptor set with zero bindings I bind VK_NULL_HANDLE, it works around the hang.  I am not actually sure whether this is an NVIDIA driver bug or validation bug.  The request here is to investigate and tighten validation to require VK_NULL_HANDLE descriptor set instead of accepting one with zero bindings.

**Valid Usage ID**
VUID-vkCmdBindDescriptorSets-pDescriptorSets-00358
Each element of pDescriptorSets must have been allocated with a VkDescriptorSetLayout that matches (is the same as, or identically defined as) the VkDescriptorSetLayout at set n in layout, where n is the sum of firstSet and the index into pDescriptorSets

**Environment:**
 - OS: Windows
 - GPU: GTX 1080
 - SDK or header version if building from repo: 1.3.231.1
 - Options enabled (synchronization, best practices, etc.): Regular validation

**Additional context**

ncesario-lunarg:
@danginsburg are you seeing 00358 when the app hangs, or this is the VUID you are _expecting_ to receive to avoid the hang?

ncesario-lunarg:
[...]

PR #4990, opened 2022-12-22
layers: Do not allow binding empty DS

ncesario-lunarg:
This effectively reverts https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/c82d75fe22da9a1115e3a5b5f98cd220fc8653bb, which allowed binding empty
descriptor sets.

Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4768.

### DIFF
--- a/layers/descriptor_validation.cpp
+++ b/layers/descriptor_validation.cpp
@@ -165,8 +165,14 @@ bool CoreChecks::VerifySetLayoutCompatibility(const cvdescriptorset::DescriptorS
     }
     if (descriptor_set.IsPushDescriptor()) return true;
     const auto *layout_node = pipeline_layout.set_layouts[layoutIndex].get();
-    if (layout_node && (descriptor_set.GetBindingCount() > 0) && (layout_node->GetBindingCount() > 0)) {
-        return VerifySetLayoutCompatibility(*layout_node, *descriptor_set.GetLayout(), errorMsg);
+    if (layout_node) {
+        if ((descriptor_set.GetBindingCount() > 0) && (layout_node->GetBindingCount() > 0)) {
+            return VerifySetLayoutCompatibility(*layout_node, *descriptor_set.GetLayout(), errorMsg);
+        } else {
+            // Only a null DSL can be used to "skip" a descriptor set a bind time, not an empty one.
+            errorMsg = "Descriptor set " + report_data->FormatHandle(descriptor_set.Handle()) + " is empty (has no bindings). Use VK_NULL_HANDLE to indicate this set is unused if using VK_PIPELINE_LAYOUT_CREATE_INDEPENDENT_SETS_BIT_EXT.";
+            return false;
+        }
     } else {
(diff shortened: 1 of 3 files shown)


jeremyg-lunarg (approved):
LGTM

tests/vklayertests_graphics_library.cpp
```diff
@@ -1568,3 +1568,151 @@ TEST_F(VkGraphicsLibraryLayerTest, CreatePipelineTessErrors) {
...
+        frag_shader_lib.gp_ci_.layout = pipeline_layout_lib.handle();
+        frag_shader_lib.CreateGraphicsPipeline(true, false);
+    }
+
+    CreatePipelineHelper frag_out_lib(*this);
+    frag_out_lib.InitFragmentOutputLibInfo();
+    frag_out_lib.InitState();
+    ASSERT_VK_SUCCESS(frag_out_lib.CreateGraphicsPipeline(true, false));
+
+    VkPipeline libraries[4] = {
```
arno-lunarg, line 1686: nit: use `std::array libraries = { /*...*/ };` ?

arno-lunarg (commented):
LGTM!

layers/descriptor_validation.cpp
```diff
@@ -165,8 +165,14 @@ bool CoreChecks::VerifySetLayoutCompatibility(const cvdescriptorset::DescriptorS
     }
     if (descriptor_set.IsPushDescriptor()) return true;
     const auto *layout_node = pipeline_layout.set_layouts[layoutIndex].get();
-    if (layout_node && (descriptor_set.GetBindingCount() > 0) && (layout_node->GetBindingCount() > 0)) {
-        return VerifySetLayoutCompatibility(*layout_node, *descriptor_set.GetLayout(), errorMsg);
+    if (layout_node) {
+        if ((descriptor_set.GetBindingCount() > 0) && (layout_node->GetBindingCount() > 0)) {
+            return VerifySetLayoutCompatibility(*layout_node, *descriptor_set.GetLayout(), errorMsg);
+        } else {
+            // Only a null DSL can be used to "skip" a descriptor set a bind time, not an empty one.
```
Kangz, line 172: Isn't it valid to have both be empty? I don't see anything in the spec that would prevent it. Yes it is a degenerate case but still valid?

So this would add a 

```cpp
} else if (descriptor_set.GetBindingCount() == 0 && layout_node->GetBindingCount() == 0) {
  // Optional warning
  return true;
```

Kangz: Context: https://bugs.chromium.org/p/dawn/issues/detail?id=1627

ncesario-lunarg: @Kangz you are 100% correct. See https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5059.

Kangz: Thank you for the timely fix!

PR #4991, opened 2022-12-22
tests: Check developer environment variables

juan-lunarg:
Checks VK_ICD_FILENAMES and VK_LAYER_PATH before running tests.

### DIFF
--- a/tests/vktestframework.cpp
+++ b/tests/vktestframework.cpp
@@ -22,6 +22,7 @@
 
 #include "vktestframework.h"
 #include "vkrenderframework.h"
+#include <filesystem>
 
 // For versions prior to VS 2015, suppress the warning
 // caused by the inconsistent redefinition of snprintf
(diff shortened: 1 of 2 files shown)


juan-lunarg:
Accidentally uninstalled my Nvidia driver and wasted a lot of time pointing to a `nvidia_icd.json` that didn't exist. This will prevent that in the future with a nice clean error message:

```
[----------] Global test environment set-up.
Invalid VK_ICD_FILENAMES! File doesn't exist!

VK_ICD_FILENAMES = /usr/share/vulkan/icd.d/foobar.json
```

These checks are also consistent with what our `common_ci.py` script does.

tests/vktestframework.cpp
```diff
@@ -108,9 +109,60 @@ int fopen_s(FILE **pFile, const char *filename, const char *mode) {
 
 #endif
 
+static void CheckEnvironmentVariables() {
+    if (const std::string vk_icd_filenames = GetEnvironment("VK_ICD_FILENAMES"); !vk_icd_filenames.empty()) {
```
ncesario-lunarg, line 113: Can't `VK_ICD_FILENAMES` be a list?

jeremyg-lunarg (approved):
LGTM

tests/vktestframework.cpp
```diff
@@ -108,9 +109,60 @@ int fopen_s(FILE **pFile, const char *filename, const char *mode) {
...
+        }
+
+        if (!std::filesystem::is_directory(layer_dir)) {
+            std::cerr << "Invalid VK_LAYER_PATH! VK_LAYER_PATH must be a directory!" << user_provided << std::endl;
+            std::exit(EXIT_FAILURE);
+        }
+
+        bool found_json = false;
+        for (auto const &dir_entry : std::filesystem::directory_iterator{layer_dir}) {
+            if (dir_entry.path().extension() == ".json") {
```
ncesario-lunarg, line 146: Do we want to check specifically for `VkLayer_khronos_validation.json`?

ncesario-lunarg (commented):
Couple questions, but otherwise LGTM!

tests/vktestframework.cpp
juan-lunarg: Looks like both environment variables can be lists. I'll handle that case.

tests/vktestframework.cpp
juan-lunarg: The reason I didn't do this is because we have a CMake option to install our tests `INSTALL_TESTS`. I'm not sure who is installing our tests and/or what they are doing. So I didn't want to possibly interrupt someones workflow.

tests/vktestframework.cpp
juan-lunarg: Also it seems like `VK_ICD_FILENAMES is deprecated but still working. It's been superceded by `VK_DRIVER_FILES`
https://github.com/KhronosGroup/Vulkan-Loader/blob/master/docs/LoaderInterfaceArchitecture.md

tests/vktestframework.cpp
ncesario-lunarg: Would that change the name of the file, though?

juan-lunarg: From the loader docs: "Override the loader's standard Layer library search folders and use the provided delimited file and/or folders to locate explicit layer manifest files."  It looks like this environment variable isn't only intended for the validation layers.

arno-lunarg (commented):
LGTM!
Just one question: in the case of the env variables being lists, they are not checked? Is that so because checking a list is not feasible, or too tedious?

juan-lunarg:
> Just one question: in the case of the env variables being lists, they are not checked? Is that so because checking a list is not feasible, or too tedious?

I just added support for lists. The non-trivial part is verifying relative paths, since the loader has implicit search paths which are platform dependent.

tests/vktestframework.cpp
juan-lunarg: I added support for lists.

tests/vktestframework.cpp
juan-lunarg: I specifically check for `VkLayer_khronos_validation.json` now since it's much more useful. If someone has an issue I'll fix it afterwards given the limited amount of people who would run the tests.

PR #4996, opened 2022-12-26
tests: Fix VK_LAYER_PATH check

spencer-lunarg:
My `VK_LAYER_PATH` for using the Profiles layer is

`VK_LAYER_PATH=/path/to/profile/:/path/to/validation_layer_repo:/path/to/sync2/layer/in/sdk`

the current check had 2 issues

1. It was not able to handle 3 or more paths due to using the `pos` instead of `length` with `std::substr`
2. Checking each path for the validation layers json file


juan-lunarg (approved):

juan-lunarg (commented):
FWIW in the future `GetVkEnvironmentVariable` can be replaced entirely with `std::views::split`. Sadly it requires C++20.

https://en.cppreference.com/w/cpp/ranges/split_view

PR #4997, opened 2022-12-26
layers: Fix discard rectangle draw VU

spencer-lunarg:
Have internal request to fix the draw time check for dynamic discard rectangle so it checks all indexes are set

https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5625

### DIFF
--- a/layers/core_validation.cpp
+++ b/layers/core_validation.cpp
@@ -403,7 +403,6 @@ bool CoreChecks::ValidateDrawDynamicState(const CMD_BUFFER_STATE &cb_state, cons
     skip |= ValidateCBDynamicStatus(cb_state, CB_DYNAMIC_PRIMITIVE_RESTART_ENABLE_SET, cmd_type, vuid.primitive_restart_enable);
     skip |= ValidateCBDynamicStatus(cb_state, CB_DYNAMIC_VERTEX_INPUT_BINDING_STRIDE_SET, cmd_type, vuid.vertex_input_binding_stride);
     skip |= ValidateCBDynamicStatus(cb_state, CB_DYNAMIC_VERTEX_INPUT_EXT_SET, cmd_type, vuid.vertex_input);
-    skip |= ValidateCBDynamicStatus(cb_state, CB_DYNAMIC_DISCARD_RECTANGLE_EXT_SET, cmd_type, vuid.dynamic_discard_rectangle);
     skip |= ValidateCBDynamicStatus(cb_state, CB_DYNAMIC_COLOR_WRITE_ENABLE_EXT_SET, cmd_type, vuid.dynamic_color_write_enable);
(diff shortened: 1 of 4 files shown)


tests/vklayertests_others.cpp
```diff
@@ -11674,11 +11674,12 @@ TEST_F(VkLayerTest, ValidateDiscardRectanglesDynamicStateNotSet) {
         GTEST_SKIP() << "Test not supported by MockICD";
     }
 
-    const VkRect2D rect = {{0, 0}, {16, 16}};
+    auto vkCmdSetDiscardRectangleEXT =
```
arno-lunarg, line 11677:
```suggestion
    auto vkCmdSetDiscardRectangleEXT = GetDeviceProcAddr<PFN_vkCmdSetDiscardRectangleEXT>("vkCmdSetDiscardRectangleEXT")
```

layers/core_validation.cpp
```diff
@@ -490,6 +489,22 @@ bool CoreChecks::ValidateDrawDynamicState(const CMD_BUFFER_STATE &cb_state, cons
         }
     }
 
+    // vkCmdSetDiscardRectangleEXT needs to be set on each rectangle
+    const auto *discard_rectangle_state = LvlFindInChain<VkPipelineDiscardRectangleStateCreateInfoEXT>(pipeline.PNext());
+    if (discard_rectangle_state) {
+        for (uint32_t i = 0; i < discard_rectangle_state->discardRectangleCount; i++) {
+            if (!cb_state.dynamic_state_value.discard_rectangles.test(i)) {
+                LogObjectList objlist(cb_state.commandBuffer());
```
arno-lunarg, line 497:
```suggestion
                const LogObjectList objlist(cb_state.commandBuffer(), pipeline.pipeline());
```

arno-lunarg (approved):
LGTM! Just 2 code suggestions

tests/vklayertests_others.cpp
spencer-lunarg: thanks, I need to scrub all these old style ones out still I guess

arno-lunarg: The improvement you suggested when we last talked about this problem of getting function pointers would be the best answer to this code flow. I think it was something like:
```
GET_DEVICE_PROC_ADDR(vkCmdSetDiscardRectangleEXT);
```
with the macro handling the get_proc_addr call and variable definition.

jeremyg-lunarg: Is this going to get fixed in a re-push or are you planning a global scrub PR?

jeremyg-lunarg (approved):
LGTM

tests/vklayertests_others.cpp
spencer-lunarg: We talked about this internally, I forgot if we created an issue, but I will not get to it anytime soon, but I did a global scrub on the `LogObjectList` (the other comment Arno pointed out)

spencer-lunarg:
The WG agreed this was correct

PR #4998, opened 2022-12-26
layers: New wsi_validation for CoreChecks

spencer-lunarg:
Creates a new `wsi_validation.cpp` file to move WSI related `CoreChecks` into

### DIFF
--- a/layers/core_validation.h
+++ b/layers/core_validation.h
@@ -38,6 +38,7 @@
 #include "cmd_buffer_state.h"
 #include "render_pass_state.h"
 #include "video_session_state.h"
+#include "sync_vuid_maps.h"
 
 // Set of VUID that need to go between core_validation.cpp and drawdispatch.cpp
 struct DrawDispatchVuid {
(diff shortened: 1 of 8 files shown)


layers/core_validation.h
```diff
@@ -222,6 +223,107 @@ class CORE_CMD_BUFFER_STATE : public CMD_BUFFER_STATE {
                           VkPipelineStageFlags2KHR src_stage_mask) override;
 };
 
+struct TimelineMaxDiffCheck {
```
ncesario-lunarg, line 226: Seems like this (and _maybe_ SemaphoreSubmitState below) belongs in a utility header (e.g., `vk_layer_utils.h`/`vk_layer_data.h`)?

ncesario-lunarg (approved):
One question, but if this is again just moving code without other refactoring, LGTM!

layers/core_validation.h
jeremyg-lunarg: Yes to the possiblity of another header. No to either of those.  SemaphoreSubmitState is heavily tied to core validation and it wouldn't be a good fit for either of those files.

spencer-lunarg: I thought about making a new header, but figured it might be better for the short term to keep here near were it was

spencer-lunarg:
> just moving code without other refactoring, LGTM!

The only "code change" was changing the 3 functions from using a template to using the new `LogObjectList` system

```patch
- template <typename T1>
- bool ValidateDeviceMaskToPhysicalDeviceCount(uint32_t deviceMask, const T1 object, const char* VUID) const;
+ bool ValidateDeviceMaskToPhysicalDeviceCount(uint32_t deviceMask, const LogObjectList& objlist, const char* VUID) const;
```

this was done because I was getting 

> undefined symbol: _ZNK10CoreChecks39ValidateDeviceMaskToPhysicalDeviceCountIP16VkSwapchainKHR_TEEbjT_PKc

and also it was on my radar to use `LogObjectList` everywhere instead anyway

Issue #4983, opened 2022-12-20
VkGpuAssistedLayerTest.GpuBufferDeviceAddressOOB hangs on VK_ERROR_DEVICE_LOST


lunarpapillo:
Sometimes the `VkGpuAssistedLayerTest.GpuBufferDeviceAddressOOB` provokes an error:
```
[ RUN      ] VkGpuAssistedLayerTest.GpuBufferDeviceAddressOOB
C:\j\vz3\Release32\Vulkan-ValidationLayers\tests\vklayertests_gpu.cpp(1151): error: Expected equality of these values:
  VK_SUCCESS
    Which is: 0
  resolved_err
    Which is: -4
VK_ERROR_DEVICE_LOST
```
I have only seen this error appear on suspect drivers, but I believe that the driver can throw `VK_ERROR_DEVICE_LOST` at any time and the test should be able to manage it gracefully.

At a minimum, the test should fail gracefully when the error is returned; this would be far better than hanging forever.

Ideally, the test case should attempt to recover, re-initializing the device and attempting the test again, a limited number of times before giving up and failing gracefully.

spencer-lunarg:
How was this different than https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4967 ?

lunarpapillo:
#4967 is a driver issue.  This issue is intended to be a test enhancement issue.

jeremyg-lunarg:
It looks like we're ready consider this fixed and reenable the test in CI.

lunarpapillo:
It's not fixed.

lunarpapillo:
Out of 16 runs since the blacklist was "fixed", four of the `Release32` runs aborted on `VkGpuAssistedLayerTest.GpuBufferDeviceAddressOOB`, e.g.: http://erusea:8080/job/Vulkan-ValidationLayers/10668/BITS=32,BUILD_MODE=Release,USE_ROBIN_HOOD_HASHING=OFF,label=tcwinnvi2/artifact/vulkantest-results/execution-logs/008-vk_layer_validation_tests-stdout.txt
```
        [ RUN      ] VkGpuAssistedLayerTest.GpuBufferDeviceAddressOOB
C:\j\vz2\Release32\Vulkan-ValidationLayers\tests\vklayertests_gpu.cpp(1136): error: Expected equality of these values:
          VK_SUCCESS
            Which is: 0
          resolved_err
            Which is: -4
[...]

PR #4999, opened 2022-12-26
vkQueueWaitIdle with timeout

arno-lunarg:
- Fix for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4983

The targeted `vkQueueWaitIdle` may cause a `VK_ERROR_DEVICE_LOST` after hanging for too long, and crash other processes. So add a timeout on the wait. It is currently set to 10 seconds.

### DIFF
--- a/tests/vklayertests_gpu.cpp
+++ b/tests/vklayertests_gpu.cpp
@@ -16,6 +16,8 @@
 
 #include "layer_validation_tests.h"
 
+#include <future>
+
 static VkValidationFeatureEnableEXT gpu_av_enables[] = {VK_VALIDATION_FEATURE_ENABLE_GPU_ASSISTED_EXT};
 static VkValidationFeatureDisableEXT gpu_av_disables[] = {
     VK_VALIDATION_FEATURE_DISABLE_THREAD_SAFETY_EXT, VK_VALIDATION_FEATURE_DISABLE_API_PARAMETERS_EXT,
@@ -1001,6 +1003,31 @@ TEST_F(VkGpuAssistedLayerTest, DrawTimeShaderUniformBufferTooSmallNestedStruct) 
                          "Descriptor size is 8 and highest byte accessed was 19");
 }
 
+static bool QueueWaitIdleWithTimeout(std::chrono::nanoseconds wait_duration, VkQueue queue) {
+    auto err_future = std::async(
+        std::launch::async, [](VkQueue queue) { return vk::QueueWaitIdle(queue); }, queue);
+
+    switch (err_future.wait_for(wait_duration)) {
+        case std::future_status::ready:
+            if (const auto err = err_future.get(); err != VK_SUCCESS) {
+                ADD_FAILURE() << "vk::QueueWaitIdle failed with error " << string_VkResult(err) << '.';
+                return false;
+            }
+            return true;
+        case std::future_status::timeout:
+            ADD_FAILURE() << "vk::QueueWaitIdle timeout. Waited " << wait_duration.count() << " seconds.";
+            return false;
+        case std::future_status::deferred:
+            assert(false);  // no reason to land here
+            ADD_FAILURE() << "unexpected deferred function using lazy evaluation.";
+            return false;
+        default:
+            assert(false);  // no reason to land here
+            ADD_FAILURE() << "Unexpected error in QueueWaitIdleWithTimeout.";
+            return false;
+    }
+};
+
 TEST_F(VkGpuAssistedLayerTest, GpuBufferDeviceAddressOOB) {
     SetTargetApiVersion(VK_API_VERSION_1_2);
     AddRequiredExtensions(VK_KHR_BUFFER_DEVICE_ADDRESS_EXTENSION_NAME);
@@ -1147,8 +1174,7 @@ TEST_F(VkGpuAssistedLayerTest, GpuBufferDeviceAddressOOB) {
     m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "access out of bounds");
     err = vk::QueueSubmit(m_device->m_queue, 1, &submit_info, VK_NULL_HANDLE);
     ASSERT_VK_SUCCESS(err);
-    err = vk::QueueWaitIdle(m_device->m_queue);
-    ASSERT_VK_SUCCESS(err);
+    ASSERT_TRUE(QueueWaitIdleWithTimeout(std::chrono::nanoseconds(kWaitTimeout), m_device->m_queue));
     m_errorMonitor->VerifyFound();
 
     // Run past the end


arno-lunarg:
The failure on nvidia is still happening but does not cause a test suite crash:
```
[==========] Running 1564 tests from 15 test suites.
[----------] Global test environment set-up.
[----------] 1 test from VkGpuAssistedLayerDeathTest
[ RUN      ] VkGpuAssistedLayerDeathTest.GpuBufferDeviceAddressOOB
Driver Name = NVIDIA
Driver Info = 515.65.01
/home/lunarg/.jenkins/vz4/Release64/Vulkan-ValidationLayers/tests/vklayertests_gpu.cpp:1185: Failure
Death test: QueueWaitIdleWithTimeout(std::chrono::seconds(8), m_errorMonitor, m_device->m_queue)
    Result: died but not with expected exit code:
            Terminated by signal 11 (core dumped)
Actual msg:
[  DEATH   ] vk::QueueWaitIdle timeout. Waited 8 seconds.
[  FAILED  ] VkGpuAssistedLayerDeathTest.GpuBufferDeviceAddressOOB (10490 ms)
[----------] 1 test from VkGpuAssistedLayerDeathTest (10490 ms total)
```

tests/vklayertests_gpu.cpp
```diff
@@ -1001,7 +1003,39 @@ TEST_F(VkGpuAssistedLayerTest, DrawTimeShaderUniformBufferTooSmallNestedStruct)
...
+static void QueueWaitIdleWithTimeout(const std::chrono::seconds wait_duration, ErrorMonitor *errorMonitor, VkQueue queue) {
+    auto err_future = std::async(
+        std::launch::async, [](VkQueue queue) { return vk::QueueWaitIdle(queue); }, queue);
+
+    switch (err_future.wait_for(wait_duration)) {
+        case std::future_status::ready:
+            if (const auto err = err_future.get(); err == VK_SUCCESS) {
+                if (const bool found = errorMonitor->VerifyFound(); found) {
+                    std::cerr << "Success";
+                    std::exit(0);
```
lunarpapillo, line 1015: Everywhere else I can find, we exit with either `EXIT_FAILURE` or `EXIT_SUCCESS`.  Here we use integers 0 through -5.  Help me understand why?

tests/vklayertests_gpu.cpp
```diff
@@ -1001,7 +1003,39 @@ TEST_F(VkGpuAssistedLayerTest, DrawTimeShaderUniformBufferTooSmallNestedStruct)
...
 
-TEST_F(VkGpuAssistedLayerTest, GpuBufferDeviceAddressOOB) {
+static void QueueWaitIdleWithTimeout(const std::chrono::seconds wait_duration, ErrorMonitor *errorMonitor, VkQueue queue) {
+    auto err_future = std::async(
+        std::launch::async, [](VkQueue queue) { return vk::QueueWaitIdle(queue); }, queue);
+
+    switch (err_future.wait_for(wait_duration)) {
+        case std::future_status::ready:
+            if (const auto err = err_future.get(); err == VK_SUCCESS) {
+                if (const bool found = errorMonitor->VerifyFound(); found) {
```
lunarpapillo, line 1013: Nit: I don't think the `; found` is necessary.  (Is it?)

tests/vkerrormonitor.cpp
```diff
@@ -152,7 +152,8 @@ bool ErrorMonitor::ExpectingSuccess() const {
 
 bool ErrorMonitor::NeedCheckSuccess() const { return ExpectingSuccess(); }
 
-void ErrorMonitor::VerifyFound() {
+bool ErrorMonitor::VerifyFound() {
+    bool found = true;
     {
         // The lock must be released before the ExpectSuccess call at the end
```
lunarpapillo, line 158: Not your code.... but is this comment correct?

lunarpapillo (commented):
Thank you for doing this.

I looked at the code as an interested party, though I should not be considered an official reviewer.

lunarpapillo:
> The failure on nvidia is still happening but does not cause a test suite crash:

That's a **Linux-NVIDIA** failure, which might be a concern since this test used to run correctly: http://erusea:8080/job/Vulkan-ValidationLayers/9910/BITS=64,BUILD_MODE=Debug,USE_ROBIN_HOOD_HASHING=ON,label=Linux-NVIVIA-new/artifact/vulkantest-results/execution-logs/009-vk_layer_validation_tests-stdout.txt
```
Driver Name = NVIDIA
Driver Info = 515.65.01
...
[ RUN      ] VkGpuAssistedLayerTest.GpuBufferDeviceAddressOOB
[       OK ] VkGpuAssistedLayerTest.GpuBufferDeviceAddressOOB (193 ms)
```
I noticed that the test section changed from `VkGpuAssistedLayerTest` to `VkGpuAssistedLayerDeathTest` - does this mean that you intentionally modified the test to die (in order to check this code)?  If so, I'm fine...

lunarpapillo:
> The targeted vkQueueWaitIdle may cause a crash, so perform wait in its own process using a death test to avoid crashing the test suite.

Ah, then, that's a problem.

For what it's worth, I don't mind if a failure crashes the test suite.  In the original issue, the problem was that the test hanged the test suite, stopping CI throughput until an hour had passed and potentially leaving behind a running process that would cause subsequent runs to fail.

tests/vklayertests_gpu.cpp
arno-lunarg: oh yes indeed!

tests/vklayertests_gpu.cpp
arno-lunarg: I wanted to use a different error code for every handled error. I really do not use `exit` often, so maybe I just should not do that, and rely on the message printed to `stderr`.

tests/vkerrormonitor.cpp
arno-lunarg: ExpectSuccess is locking the mutex, so I think this comment is valid

arno-lunarg:
> I noticed that the test section changed from VkGpuAssistedLayerTest to VkGpuAssistedLayerDeathTest - does this mean that you intentionally modified the test to die (in order to check this code)? If so, I'm fine...

Yes, the google test doc advises to use a test class ending with `DeathTest` so that death test are detected and ran first. But the test should succeed if there is no timeout or a driver crash. The test will now just die properly if there is a crash.

arno-lunarg:
> That's a Linux-NVIDIA failure, which might be a concern since this test used to run correctly
Maybe the crash happens _because_ of my changes then, unforeseen consequences of the fork call on linux maybe?

arno-lunarg:
> For what it's worth, I don't mind if a failure crashes the test suite. In the original issue, the problem was that the test hanged the test suite, stopping CI throughput until an hour had passed and potentially leaving behind a running process that would cause subsequent runs to fail.

Oh ok! Doing this should be easy, basically I just have to call `QueueWaitIdleWithTimeout` without putting it in a death test macro.

arno-lunarg:
I switched to a simple timeout on the wait idle without new process/death test logic to see if it fixes the test crashing. If it does, then probably let's go for this way of doing things, since it is what @lunarpapillo asked for.

jeremyg-lunarg (changes requested):
I'm worried what happens when the QueueWaitIdle() doesn't finish on time and this code handles the timeout but there's still that wait hanging out in the background. What does that do to the rest of the test run? Why not just wait on a fence, which can time out?

tests/vklayertests_gpu.cpp
```diff
@@ -1147,8 +1173,7 @@ TEST_F(VkGpuAssistedLayerTest, GpuBufferDeviceAddressOOB) {
     m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "access out of bounds");
     err = vk::QueueSubmit(m_device->m_queue, 1, &submit_info, VK_NULL_HANDLE);
     ASSERT_VK_SUCCESS(err);
-    err = vk::QueueWaitIdle(m_device->m_queue);
-    ASSERT_VK_SUCCESS(err);
+    QueueWaitIdleWithTimeout(std::chrono::seconds(10), m_device->m_queue);
```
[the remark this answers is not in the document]
arno-lunarg: ah I did not think about that, it should be more simple (and also simply the correct Vulkan way)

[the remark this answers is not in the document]
arno-lunarg: Thinking about this overnight, since the error is a device lost, and since the fence is created using the device, isn't there a risk for the fence to be in a bad state and _unwaitable_?

arno-lunarg:
Closing this PR, it does not fix the problem

PR #5000, opened 2022-12-26
Add support for VK_EXT_surface_maintenance1 and VK_EXT_swapchain_maintenance1

### LINKED ISSUE #4985: Update swapchain VUs changed from VK_EXT_swapchain_maintenance1
With `VK_EXT_swapchain_maintenance1` added in the 1.3.237 headers the `VUID-vkAcquireNextImage2KHR-swapchain-01803` / `VUID-vkAcquireNextImageKHR-swapchain-01802` were slightly altered to `VUID-vkAcquireNextImage2KHR-surface-07784` / `VUID-vkAcquireNextImageKHR-surface-07783` with the new [swapchain forward progress](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#swapchain-acquire-forward-progress) language

This VU checked should be updated correctly

mark-lunarg:
Most if not all of the existing review feedback has been addressed. Should be in fully reviewable form -- see the top two commits for the newer, more questionable changes.

replaces `VUID-vkAcquireNextImage2KHR-swapchain-01803` and `VUID-vkAcquireNextImageKHR-swapchain-01802` for `VUID-vkAcquireNextImage2KHR-surface-07784` and `VUID-vkAcquireNextImageKHR-surface-07783`

Fixes #4985.

When final, update tracking issue [here](https://github.com/KhronosGroup/Vulkan-Docs/issues/2006).

### DIFF
--- a/layers/image_state.cpp
+++ b/layers/image_state.cpp
@@ -704,17 +704,24 @@ bool SURFACE_STATE::GetQueueSupport(VkPhysicalDevice phys_dev, uint32_t qfi) con
 void SURFACE_STATE::SetPresentModes(VkPhysicalDevice phys_dev, std::vector<VkPresentModeKHR> &&modes) {
     auto guard = Lock();
     assert(phys_dev);
-    present_modes_[phys_dev] = std::move(modes);
+    for (auto new_present_mode : modes) {
+        if ((present_modes_data_.find(phys_dev) == present_modes_data_.end()) ||
+            (present_modes_data_[phys_dev].find(new_present_mode) == present_modes_data_[phys_dev].end())) {
+            present_modes_data_[phys_dev][new_present_mode] = std::nullopt;
+        }
+    }
 }
 
 std::vector<VkPresentModeKHR> SURFACE_STATE::GetPresentModes(VkPhysicalDevice phys_dev) const {
     auto guard = Lock();
     assert(phys_dev);
-    auto iter = present_modes_.find(phys_dev);
-    if (iter != present_modes_.end()) {
-        return iter->second;
-    }
     std::vector<VkPresentModeKHR> result;
+    if (present_modes_data_.find(phys_dev) != present_modes_data_.end()) {
+        for (auto mode = present_modes_data_[phys_dev].begin(); mode != present_modes_data_[phys_dev].end(); mode++) {
+            result.push_back(mode->first);
+        }
+        return result;
+    }
     uint32_t count = 0;
     DispatchGetPhysicalDeviceSurfacePresentModesKHR(phys_dev, surface(), &count, nullptr);
     result.resize(count);
@@ -769,14 +776,14 @@ void SURFACE_STATE::SetCompatibleModes(VkPhysicalDevice phys_dev, const VkPresen
     assert(phys_dev);
 
     // If this surface or the present_mode is not in the map, create and add the present_mode state
-    auto surface_map = maint1_present_modes_.find(phys_dev);
-    if ((surface_map == maint1_present_modes_.end()) || (surface_map->second.find(present_mode) == surface_map->second.end())) {
+    auto surface_map = present_modes_data_.find(phys_dev);
+    if ((surface_map == present_modes_data_.end()) || (surface_map->second.find(present_mode) == surface_map->second.end())) {
         auto present_mode_state = std::make_shared<PresentModeState>();
         present_mode_state->compatible_present_modes_ = compatible_modes;
 
-        // For every present mode in compatible modes, add present_mode_state for it in maint1_present_modes_
+        // For every present mode in compatible modes, add present_mode_state for it in present_modes_data_
         for (auto mode : compatible_modes) {
-                maint1_present_modes_[phys_dev][mode] = present_mode_state;
+                present_modes_data_[phys_dev][mode] = present_mode_state;
         }
     }
 }
@@ -784,9 +791,14 @@ void SURFACE_STATE::SetCompatibleModes(VkPhysicalDevice phys_dev, const VkPresen
 std::vector<VkPresentModeKHR> SURFACE_STATE::GetCompatibleModes(VkPhysicalDevice phys_dev, const VkPresentModeKHR present_mode) const {
     auto guard = Lock();
     assert(phys_dev);
-    auto iter = maint1_present_modes_.find(phys_dev);
-    if ((iter != maint1_present_modes_.end()) && (iter->second.find(present_mode) != iter->second.end())) {
-        return (iter->second)[present_mode]->compatible_present_modes_;
+    auto iter = present_modes_data_.find(phys_dev);
+    if ((iter != present_modes_data_.end()) && (iter->second.find(present_mode) != iter->second.end())) {
+        if (((iter->second)[present_mode]).has_value()) {
+            auto &compatible_modes = *(iter->second)[present_mode];
+            if (compatible_modes->compatible_present_modes_.size() != 0) {
+                return compatible_modes->compatible_present_modes_;
+            }
+        }
     }
 
     // Compatible modes not in state tracker, call to get compatible modes
@@ -812,16 +824,25 @@ void SURFACE_STATE::SetPresentModeCapabilities(VkPhysicalDevice phys_dev, const 
                                                const VkSurfacePresentScalingCapabilitiesEXT &scaling_caps) {
     auto guard = Lock();
     assert(phys_dev);
-    maint1_present_modes_[phys_dev][present_mode]->scaling_capabilities_ = scaling_caps;
-    maint1_present_modes_[phys_dev][present_mode]->surface_capabilities_ = caps;
-};
+    if (!present_modes_data_[phys_dev][present_mode].has_value()) {
+        present_modes_data_[phys_dev][present_mode] = std::make_shared<PresentModeState>();
+    }
+    // Old MacOS doesn't support Value()
+    auto &present_mode_state = *(present_modes_data_[phys_dev][present_mode]);
+    present_mode_state->scaling_capabilities_ = scaling_caps;
+    present_mode_state->surface_capabilities_ = caps;
+}
 
 // Get the surface caps this particular present mode
 VkSurfaceCapabilitiesKHR SURFACE_STATE::GetPresentModeSurfaceCapabilities(VkPhysicalDevice phys_dev,
                                                                           const VkPresentModeKHR present_mode) const {
-    auto iter = maint1_present_modes_.find(phys_dev);
-    if ((iter != maint1_present_modes_.end()) && (iter->second.find(present_mode) != iter->second.end())) {
-        return (iter->second)[present_mode]->surface_capabilities_;
+    auto iter = present_modes_data_.find(phys_dev);
+    if ((iter != present_modes_data_.end()) && (iter->second.find(present_mode) != iter->second.end())) {
+        auto const caps = (iter->second)[present_mode];
+        if (caps.has_value()) {
+            auto &surface_caps = *caps;
+            return surface_caps->surface_capabilities_;
+        }
     }
 
     // Present mode surface capabilties not in state tracker, call to get surface capabilities
@@ -838,9 +859,13 @@ VkSurfaceCapabilitiesKHR SURFACE_STATE::GetPresentModeSurfaceCapabilities(VkPhys
 // Get the scaling capabilities for this particular present mode
 VkSurfacePresentScalingCapabilitiesEXT SURFACE_STATE::GetPresentModeScalingCapabilities(VkPhysicalDevice phys_dev,
                                                                                         const VkPresentModeKHR present_mode) const {
-    auto iter = maint1_present_modes_.find(phys_dev);
-    if ((iter != maint1_present_modes_.end()) && (iter->second.find(present_mode) != iter->second.end())) {
-        return (iter->second)[present_mode]->scaling_capabilities_;
+    auto iter = present_modes_data_.find(phys_dev);
+    if ((iter != present_modes_data_.end()) && (iter->second.find(present_mode) != iter->second.end())) {
+        auto const &caps = (iter->second)[present_mode];
+        if (caps.has_value()) {
+            auto &scaling_caps = *caps;
+            return scaling_caps->scaling_capabilities_;
+        }
     }
 
     // Present mode scaling capabilties not in state tracker, call to get scaling capabilities
@@ -855,8 +880,3 @@ VkSurfacePresentScalingCapabilitiesEXT SURFACE_STATE::GetPresentModeScalingCapab
     DispatchGetPhysicalDeviceSurfaceCapabilities2KHR(phys_dev, &surface_info, &surface_capabilities);
     return scaling_caps;
 }
-
-
-
-
-
--- a/layers/image_state.cpp
+++ b/layers/image_state.cpp
@@ -701,7 +701,9 @@ bool SURFACE_STATE::GetQueueSupport(VkPhysicalDevice phys_dev, uint32_t qfi) con
     return supported == VK_TRUE;
 }
 
-void SURFACE_STATE::SetPresentModes(VkPhysicalDevice phys_dev, std::vector<VkPresentModeKHR> &&modes) {
+// Save data from vkGetPhysicalDeviceSurfacePresentModes
+void SURFACE_STATE::SetPresentModes(VkPhysicalDevice phys_dev, layer_data::span<const VkPresentModeKHR> modes) {
+
     auto guard = Lock();
     assert(phys_dev);
     for (auto new_present_mode : modes) {
@@ -712,6 +714,7 @@ void SURFACE_STATE::SetPresentModes(VkPhysicalDevice phys_dev, std::vector<VkPre
     }
 }
 
+// Helper for data obtained from vkGetPhysicalDeviceSurfacePresentModesKHR
 std::vector<VkPresentModeKHR> SURFACE_STATE::GetPresentModes(VkPhysicalDevice phys_dev) const {
     auto guard = Lock();
     assert(phys_dev);
@@ -771,19 +774,21 @@ VkSurfaceCapabilitiesKHR SURFACE_STATE::GetCapabilities(VkPhysicalDevice phys_de
 }
 
 void SURFACE_STATE::SetCompatibleModes(VkPhysicalDevice phys_dev, const VkPresentModeKHR present_mode,
-                                       std::vector<VkPresentModeKHR> &&compatible_modes) {
+                                       layer_data::span<const VkPresentModeKHR> compatible_modes) {
     auto guard = Lock();
     assert(phys_dev);
 
-    // If this surface or the present_mode is not in the map, create and add the present_mode state
+    // If this surface or the present_mode is not in the map, or if the state structure has no value,
+    // create and add the new present_mode state structure for each of the compatible modes
     auto surface_map = present_modes_data_.find(phys_dev);
-    if ((surface_map == present_modes_data_.end()) || (surface_map->second.find(present_mode) == surface_map->second.end())) {
+    if ((surface_map == present_modes_data_.end()) || (surface_map->second.find(present_mode) == surface_map->second.end()) ||
+        (surface_map->second.find(present_mode)->second.has_value() == false)) {
         auto present_mode_state = std::make_shared<PresentModeState>();
-        present_mode_state->compatible_present_modes_ = compatible_modes;
+        present_mode_state->compatible_present_modes_.assign(compatible_modes.begin(), compatible_modes.end());
 
         // For every present mode in compatible modes, add present_mode_state for it in present_modes_data_
         for (auto mode : compatible_modes) {
-                present_modes_data_[phys_dev][mode] = present_mode_state;
+            present_modes_data_[phys_dev][mode] = present_mode_state;
         }
     }
 }
@@ -795,7 +800,7 @@ std::vector<VkPresentModeKHR> SURFACE_STATE::GetCompatibleModes(VkPhysicalDevice
     if ((iter != present_modes_data_.end()) && (iter->second.find(present_mode) != iter->second.end())) {
         if (((iter->second)[present_mode]).has_value()) {
             auto &compatible_modes = *(iter->second)[present_mode];
-            if (compatible_modes->compatible_present_modes_.size() != 0) {
+            if (compatible_modes->compatible_present_modes_.empty()) {
                 return compatible_modes->compatible_present_modes_;
             }
         }
@@ -803,13 +808,13 @@ std::vector<VkPresentModeKHR> SURFACE_STATE::GetCompatibleModes(VkPhysicalDevice
 
     // Compatible modes not in state tracker, call to get compatible modes
     std::vector<VkPresentModeKHR> result;
-    VkPhysicalDeviceSurfaceInfo2KHR surface_info = LvlInitStruct<VkPhysicalDeviceSurfaceInfo2KHR>();
+    auto surface_info = LvlInitStruct<VkPhysicalDeviceSurfaceInfo2KHR>();
     surface_info.surface = surface();
-    VkSurfacePresentModeEXT surface_present_mode = LvlInitStruct<VkSurfacePresentModeEXT>();
+    auto surface_present_mode = LvlInitStruct<VkSurfacePresentModeEXT>();
     surface_present_mode.presentMode = present_mode;
     surface_info.pNext = &surface_present_mode;
-    VkSurfacePresentModeCompatibilityEXT present_mode_compatibility = LvlInitStruct<VkSurfacePresentModeCompatibilityEXT>();
-    VkSurfaceCapabilities2KHR surface_capabilities = LvlInitStruct<VkSurfaceCapabilities2KHR>();
+    auto present_mode_compatibility = LvlInitStruct<VkSurfacePresentModeCompatibilityEXT>();
+    auto surface_capabilities = LvlInitStruct<VkSurfaceCapabilities2KHR>();
     surface_capabilities.pNext = &present_mode_compatibility;
     DispatchGetPhysicalDeviceSurfaceCapabilities2KHR(phys_dev, &surface_info, &surface_capabilities);
     result.resize(present_mode_compatibility.presentModeCount);
@@ -846,12 +851,12 @@ VkSurfaceCapabilitiesKHR SURFACE_STATE::GetPresentModeSurfaceCapabilities(VkPhys
     }
 
     // Present mode surface capabilties not in state tracker, call to get surface capabilities
-    VkPhysicalDeviceSurfaceInfo2KHR surface_info = LvlInitStruct<VkPhysicalDeviceSurfaceInfo2KHR>();
+    auto surface_info = LvlInitStruct<VkPhysicalDeviceSurfaceInfo2KHR>();
     surface_info.surface = surface();
-    VkSurfacePresentModeEXT surface_present_mode = LvlInitStruct<VkSurfacePresentModeEXT>();
+    auto surface_present_mode = LvlInitStruct<VkSurfacePresentModeEXT>();
     surface_present_mode.presentMode = present_mode;
     surface_info.pNext = &surface_present_mode;
-    VkSurfaceCapabilities2KHR surface_capabilities = LvlInitStruct<VkSurfaceCapabilities2KHR>();
+    auto surface_capabilities = LvlInitStruct<VkSurfaceCapabilities2KHR>();
     DispatchGetPhysicalDeviceSurfaceCapabilities2KHR(phys_dev, &surface_info, &surface_capabilities);
     return surface_capabilities.surfaceCapabilities;
 }
@@ -869,13 +874,13 @@ VkSurfacePresentScalingCapabilitiesEXT SURFACE_STATE::GetPresentModeScalingCapab
     }
 
     // Present mode scaling capabilties not in state tracker, call to get scaling capabilities
-    VkPhysicalDeviceSurfaceInfo2KHR surface_info = LvlInitStruct<VkPhysicalDeviceSurfaceInfo2KHR>();
+    auto surface_info = LvlInitStruct<VkPhysicalDeviceSurfaceInfo2KHR>();
     surface_info.surface = surface();
-    VkSurfacePresentModeEXT surface_present_mode = LvlInitStruct<VkSurfacePresentModeEXT>();
+    auto surface_present_mode = LvlInitStruct<VkSurfacePresentModeEXT>();
     surface_present_mode.presentMode = present_mode;
     surface_info.pNext = &surface_present_mode;
-    VkSurfacePresentScalingCapabilitiesEXT scaling_caps = LvlInitStruct<VkSurfacePresentScalingCapabilitiesEXT>();
-    VkSurfaceCapabilities2KHR surface_capabilities = LvlInitStruct<VkSurfaceCapabilities2KHR>();
+    auto scaling_caps = LvlInitStruct<VkSurfacePresentScalingCapabilitiesEXT>();
+    auto surface_capabilities = LvlInitStruct<VkSurfaceCapabilities2KHR>();
     surface_capabilities.pNext = &scaling_caps;
     DispatchGetPhysicalDeviceSurfaceCapabilities2KHR(phys_dev, &surface_info, &surface_capabilities);
     return scaling_caps;
--- a/layers/image_state.cpp
+++ b/layers/image_state.cpp
@@ -1,6 +1,6 @@
-/* Copyright (c) 2015-2022 The Khronos Group Inc.
- * Copyright (c) 2015-2022 Valve Corporation
- * Copyright (c) 2015-2022 LunarG, Inc.
+/* Copyright (c) 2015-2023 The Khronos Group Inc.
+ * Copyright (c) 2015-2023 Valve Corporation
+ * Copyright (c) 2015-2023 LunarG, Inc.
  * Copyright (C) 2015-2022 Google Inc.
  * Modifications Copyright (C) 2020 Advanced Micro Devices, Inc. All rights reserved.
  * Modifications Copyright (C) 2022 RasterGrid Kft.
@@ -762,3 +762,101 @@ VkSurfaceCapabilitiesKHR SURFACE_STATE::GetCapabilities(VkPhysicalDevice phys_de
     capabilities_[phys_dev] = result;
     return result;
 }
+
+void SURFACE_STATE::SetCompatibleModes(VkPhysicalDevice phys_dev, const VkPresentModeKHR present_mode,
+                                       std::vector<VkPresentModeKHR> &&compatible_modes) {
+    auto guard = Lock();
+    assert(phys_dev);
+
+    // If this surface or the present_mode is not in the map, create and add the present_mode state
+    auto surface_map = maint1_present_modes_.find(phys_dev);
+    if ((surface_map == maint1_present_modes_.end()) || (surface_map->second.find(present_mode) == surface_map->second.end())) {
+        auto present_mode_state = std::make_shared<PresentModeState>();
+        present_mode_state->compatible_present_modes_ = compatible_modes;
+
+        // For every present mode in compatible modes, add present_mode_state for it in maint1_present_modes_
+        for (auto mode : compatible_modes) {
+                maint1_present_modes_[phys_dev][mode] = present_mode_state;
+        }
+    }
+}
+
+std::vector<VkPresentModeKHR> SURFACE_STATE::GetCompatibleModes(VkPhysicalDevice phys_dev, const VkPresentModeKHR present_mode) const {
+    auto guard = Lock();
+    assert(phys_dev);
+    auto iter = maint1_present_modes_.find(phys_dev);
+    if ((iter != maint1_present_modes_.end()) && (iter->second.find(present_mode) != iter->second.end())) {
+        return (iter->second)[present_mode]->compatible_present_modes_;
+    }
+
+    // Compatible modes not in state tracker, call to get compatible modes
+    std::vector<VkPresentModeKHR> result;
+    VkPhysicalDeviceSurfaceInfo2KHR surface_info = LvlInitStruct<VkPhysicalDeviceSurfaceInfo2KHR>();
+    surface_info.surface = surface();
+    VkSurfacePresentModeEXT surface_present_mode = LvlInitStruct<VkSurfacePresentModeEXT>();
+    surface_present_mode.presentMode = present_mode;
+    surface_info.pNext = &surface_present_mode;
+    VkSurfacePresentModeCompatibilityEXT present_mode_compatibility = LvlInitStruct<VkSurfacePresentModeCompatibilityEXT>();
+    VkSurfaceCapabilities2KHR surface_capabilities = LvlInitStruct<VkSurfaceCapabilities2KHR>();
+    surface_capabilities.pNext = &present_mode_compatibility;
+    DispatchGetPhysicalDeviceSurfaceCapabilities2KHR(phys_dev, &surface_info, &surface_capabilities);
+    result.resize(present_mode_compatibility.presentModeCount);
+    present_mode_compatibility.pPresentModes = result.data();
+    DispatchGetPhysicalDeviceSurfaceCapabilities2KHR(phys_dev, &surface_info, &surface_capabilities);
+    return result;
+}
+
+// Set the surface and scaling caps for this present mode
+void SURFACE_STATE::SetPresentModeCapabilities(VkPhysicalDevice phys_dev, const VkPresentModeKHR present_mode,
+                                               const VkSurfaceCapabilitiesKHR &caps,
+                                               const VkSurfacePresentScalingCapabilitiesEXT &scaling_caps) {
+    auto guard = Lock();
+    assert(phys_dev);
+    maint1_present_modes_[phys_dev][present_mode]->scaling_capabilities_ = scaling_caps;
+    maint1_present_modes_[phys_dev][present_mode]->surface_capabilities_ = caps;
+};
+
+// Get the surface caps this particular present mode
+VkSurfaceCapabilitiesKHR SURFACE_STATE::GetPresentModeSurfaceCapabilities(VkPhysicalDevice phys_dev,
+                                                                          const VkPresentModeKHR present_mode) const {
+    auto iter = maint1_present_modes_.find(phys_dev);
+    if ((iter != maint1_present_modes_.end()) && (iter->second.find(present_mode) != iter->second.end())) {
+        return (iter->second)[present_mode]->surface_capabilities_;
+    }
+
+    // Present mode surface capabilties not in state tracker, call to get surface capabilities
+    VkPhysicalDeviceSurfaceInfo2KHR surface_info = LvlInitStruct<VkPhysicalDeviceSurfaceInfo2KHR>();
+    surface_info.surface = surface();
+    VkSurfacePresentModeEXT surface_present_mode = LvlInitStruct<VkSurfacePresentModeEXT>();
+    surface_present_mode.presentMode = present_mode;
+    surface_info.pNext = &surface_present_mode;
+    VkSurfaceCapabilities2KHR surface_capabilities = LvlInitStruct<VkSurfaceCapabilities2KHR>();
+    DispatchGetPhysicalDeviceSurfaceCapabilities2KHR(phys_dev, &surface_info, &surface_capabilities);
+    return surface_capabilities.surfaceCapabilities;
+}
+
+// Get the scaling capabilities for this particular present mode
+VkSurfacePresentScalingCapabilitiesEXT SURFACE_STATE::GetPresentModeScalingCapabilities(VkPhysicalDevice phys_dev,
+                                                                                        const VkPresentModeKHR present_mode) const {
+    auto iter = maint1_present_modes_.find(phys_dev);
+    if ((iter != maint1_present_modes_.end()) && (iter->second.find(present_mode) != iter->second.end())) {
+        return (iter->second)[present_mode]->scaling_capabilities_;
+    }
+
+    // Present mode scaling capabilties not in state tracker, call to get scaling capabilities
+    VkPhysicalDeviceSurfaceInfo2KHR surface_info = LvlInitStruct<VkPhysicalDeviceSurfaceInfo2KHR>();
+    surface_info.surface = surface();
+    VkSurfacePresentModeEXT surface_present_mode = LvlInitStruct<VkSurfacePresentModeEXT>();
+    surface_present_mode.presentMode = present_mode;
+    surface_info.pNext = &surface_present_mode;
+    VkSurfacePresentScalingCapabilitiesEXT scaling_caps = LvlInitStruct<VkSurfacePresentScalingCapabilitiesEXT>();
+    VkSurfaceCapabilities2KHR surface_capabilities = LvlInitStruct<VkSurfaceCapabilities2KHR>();
+    surface_capabilities.pNext = &scaling_caps;
+    DispatchGetPhysicalDeviceSurfaceCapabilities2KHR(phys_dev, &surface_info, &surface_capabilities);
+    return scaling_caps;
+}
+
+
+
+
+
--- a/layers/image_state.h
+++ b/layers/image_state.h
@@ -1,6 +1,6 @@
-/* Copyright (c) 2015-2022 The Khronos Group Inc.
- * Copyright (c) 2015-2022 Valve Corporation
- * Copyright (c) 2015-2022 LunarG, Inc.
+/* Copyright (c) 2015-2023 The Khronos Group Inc.
+ * Copyright (c) 2015-2023 Valve Corporation
+ * Copyright (c) 2015-2023 LunarG, Inc.
  * Copyright (C) 2015-2022 Google Inc.
  * Modifications Copyright (C) 2020 Advanced Micro Devices, Inc. All rights reserved.
  * Modifications Copyright (C) 2022 RasterGrid Kft.
@@ -288,6 +288,7 @@ struct SWAPCHAIN_IMAGE {
 class SWAPCHAIN_NODE : public BASE_NODE {
   public:
     const safe_VkSwapchainCreateInfoKHR createInfo;
+    std::vector<VkPresentModeKHR> present_modes;
     std::vector<SWAPCHAIN_IMAGE> images;
     bool retired = false;
     bool exclusive_full_screen_access;
@@ -343,6 +344,12 @@ struct hash<GpuQueue> {
 }  // namespace std
 
 // State for VkSurfaceKHR objects.
+struct PresentModeState {
+    VkSurfaceCapabilitiesKHR surface_capabilities_;
+    VkSurfacePresentScalingCapabilitiesEXT scaling_capabilities_;
+    std::vector<VkPresentModeKHR> compatible_present_modes_;
+};
+
 // Parent -> child relationships in the object usage tree:
 //    SURFACE_STATE -> nothing
 class SURFACE_STATE : public BASE_NODE {
@@ -375,13 +382,26 @@ class SURFACE_STATE : public BASE_NODE {
     void SetCapabilities(VkPhysicalDevice phys_dev, const VkSurfaceCapabilitiesKHR &caps);
     VkSurfaceCapabilitiesKHR GetCapabilities(VkPhysicalDevice phys_dev) const;
 
+    void SetCompatibleModes(VkPhysicalDevice phys_dev, const VkPresentModeKHR present_mode,
+                            std::vector<VkPresentModeKHR> &&compatible_modes);
+    std::vector<VkPresentModeKHR> GetCompatibleModes(VkPhysicalDevice phys_dev, const VkPresentModeKHR present_mode) const;
+
+    void SetPresentModeCapabilities(VkPhysicalDevice phys_dev, const VkPresentModeKHR present_mode,
+                                    const VkSurfaceCapabilitiesKHR &caps, const VkSurfacePresentScalingCapabilitiesEXT &scaling_caps);
+    VkSurfaceCapabilitiesKHR GetPresentModeSurfaceCapabilities(VkPhysicalDevice phys_dev,
+                                                                const VkPresentModeKHR present_mode) const;
+    VkSurfacePresentScalingCapabilitiesEXT GetPresentModeScalingCapabilities(VkPhysicalDevice phys_dev,
+                                                                             const VkPresentModeKHR present_mode) const;
     SWAPCHAIN_NODE *swapchain{nullptr};
 
   private:
     std::unique_lock<std::mutex> Lock() const { return std::unique_lock<std::mutex>(lock_); }
     mutable std::mutex lock_;
     mutable layer_data::unordered_map<GpuQueue, bool> gpu_queue_support_;
-    mutable layer_data::unordered_map<VkPhysicalDevice, std::vector<VkPresentModeKHR>> present_modes_;
     mutable layer_data::unordered_map<VkPhysicalDevice, std::vector<VkSurfaceFormatKHR>> formats_;
     mutable layer_data::unordered_map<VkPhysicalDevice, VkSurfaceCapabilitiesKHR> capabilities_;
+    mutable layer_data::unordered_map<VkPhysicalDevice, std::vector<VkPresentModeKHR>> present_modes_;
+    mutable layer_data::unordered_map<VkPhysicalDevice,
+        layer_data::unordered_map<VkPresentModeKHR, std::shared_ptr<PresentModeState>>> maint1_present_modes_;
+
 };
--- a/layers/image_state.h
+++ b/layers/image_state.h
@@ -373,7 +373,7 @@ class SURFACE_STATE : public BASE_NODE {
     void SetQueueSupport(VkPhysicalDevice phys_dev, uint32_t qfi, bool supported);
     bool GetQueueSupport(VkPhysicalDevice phys_dev, uint32_t qfi) const;
 
-    void SetPresentModes(VkPhysicalDevice phys_dev, std::vector<VkPresentModeKHR> &&modes);
+    void SetPresentModes(VkPhysicalDevice phys_dev, layer_data::span<const VkPresentModeKHR> modes);
     std::vector<VkPresentModeKHR> GetPresentModes(VkPhysicalDevice phys_dev) const;
 
     void SetFormats(VkPhysicalDevice phys_dev, std::vector<VkSurfaceFormatKHR> &&fmts);
@@ -383,9 +383,8 @@ class SURFACE_STATE : public BASE_NODE {
     VkSurfaceCapabilitiesKHR GetCapabilities(VkPhysicalDevice phys_dev) const;
 
     void SetCompatibleModes(VkPhysicalDevice phys_dev, const VkPresentModeKHR present_mode,
-                            std::vector<VkPresentModeKHR> &&compatible_modes);
+                            layer_data::span<const VkPresentModeKHR> compatible_modes);
     std::vector<VkPresentModeKHR> GetCompatibleModes(VkPhysicalDevice phys_dev, const VkPresentModeKHR present_mode) const;
-
     void SetPresentModeCapabilities(VkPhysicalDevice phys_dev, const VkPresentModeKHR present_mode,
                                     const VkSurfaceCapabilitiesKHR &caps, const VkSurfacePresentScalingCapabilitiesEXT &scaling_caps);
     VkSurfaceCapabilitiesKHR GetPresentModeSurfaceCapabilities(VkPhysicalDevice phys_dev,
@@ -401,5 +400,6 @@ class SURFACE_STATE : public BASE_NODE {
     mutable layer_data::unordered_map<VkPhysicalDevice, std::vector<VkSurfaceFormatKHR>> formats_;
     mutable layer_data::unordered_map<VkPhysicalDevice, VkSurfaceCapabilitiesKHR> capabilities_;
     mutable layer_data::unordered_map<VkPhysicalDevice,
-        layer_data::unordered_map<VkPresentModeKHR, std::optional<std::shared_ptr<PresentModeState>>>> present_modes_data_;
+                                      layer_data::unordered_map<VkPresentModeKHR, std::optional<std::shared_ptr<PresentModeState>>>>
+        present_modes_data_;
 };
--- a/layers/image_state.h
+++ b/layers/image_state.h
@@ -400,8 +400,6 @@ class SURFACE_STATE : public BASE_NODE {
     mutable layer_data::unordered_map<GpuQueue, bool> gpu_queue_support_;
     mutable layer_data::unordered_map<VkPhysicalDevice, std::vector<VkSurfaceFormatKHR>> formats_;
     mutable layer_data::unordered_map<VkPhysicalDevice, VkSurfaceCapabilitiesKHR> capabilities_;
-    mutable layer_data::unordered_map<VkPhysicalDevice, std::vector<VkPresentModeKHR>> present_modes_;
     mutable layer_data::unordered_map<VkPhysicalDevice,
-        layer_data::unordered_map<VkPresentModeKHR, std::shared_ptr<PresentModeState>>> maint1_present_modes_;
-
+        layer_data::unordered_map<VkPresentModeKHR, std::optional<std::shared_ptr<PresentModeState>>>> present_modes_data_;
 };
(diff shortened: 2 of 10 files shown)


layers/parameter_validation_utils.cpp
```diff
@@ -9268,6 +9268,45 @@ bool StatelessValidation::manual_PreCallValidateGetPhysicalDeviceSurfaceCapabili
         }
     }
 #endif
+
+    if (instance_extensions.vk_ext_surface_maintenance1) {
```
jeremyg-lunarg, line 9272: Shouldn't this use IsExtEnabled()?

layers/parameter_validation_utils.cpp
```diff
@@ -9268,6 +9268,45 @@ bool StatelessValidation::manual_PreCallValidateGetPhysicalDeviceSurfaceCapabili
...
+
+        if (surface_present_scaling_compatibilities && !(LvlFindInChain<VkSurfacePresentModeEXT>(pSurfaceInfo->pNext))) {
+            skip |= LogError(device, "VUID-vkGetPhysicalDeviceSurfaceCapabilities2KHR-pNext-07777",
+                             "vkGetPhysicalDeviceSurfaceCapabilities2KHR(): VK_EXT_surface_maintenance1 is enabled and "
+                             "pSurfaceCapabilities->pNext contains "
+                             "VkSurfacePresentScalingCapabilitiesEXT, but pSurfaceInfo->pNext does not contain a "
+                             "VkSurfacePresentModeEXT structure.");
+        }
+
+        if (instance_extensions.vk_google_surfaceless_query) {
```
jeremyg-lunarg, line 9294: Also IsExtEnabled()?

layers/state_tracker.cpp
```diff
@@ -3762,6 +3762,12 @@ void ValidationStateTracker::RecordCreateSwapchainState(VkResult result, const V
         surface_state->AddParent(swapchain.get());
         surface_state->swapchain = swapchain.get();
         swapchain->surface = std::move(surface_state);
+        auto swapchain_present_modes_ci = LvlFindInChain<VkSwapchainPresentModesCreateInfoEXT>(pCreateInfo->pNext);
+        if (swapchain_present_modes_ci) {
+            for (uint32_t i = 0; i < swapchain_present_modes_ci->presentModeCount; i++) {
```
jeremyg-lunarg, line 3767: You should call swapchain->present_modes.reserve(swapchain_present_modes_ci->presentModeCount); before the loop. Or use std::copy() instead of the loop.

layers/state_tracker.cpp
```diff
@@ -3845,6 +3851,43 @@ void ValidationStateTracker::PostCallRecordQueuePresentKHR(VkQueue queue, const
     }
 }
 
+void ValidationStateTracker::PostCallRecordReleaseSwapchainImagesEXT(VkDevice device, const VkReleaseSwapchainImagesInfoEXT *pReleaseInfo, VkResult result) {
+
+    if (result == VK_ERROR_DEVICE_LOST) {
```
jeremyg-lunarg, line 3856: Should this be result != VK_SUCCESS for consistency with other code?

layers/state_tracker.cpp
```diff
@@ -4085,7 +4128,30 @@ void ValidationStateTracker::PostCallRecordGetPhysicalDeviceSurfaceCapabilities2
 
     if (pSurfaceInfo->surface) {
         auto surface_state = Get<SURFACE_STATE>(pSurfaceInfo->surface);
-        surface_state->SetCapabilities(physicalDevice, pSurfaceCapabilities->surfaceCapabilities);
+
+        const VkSurfacePresentModeEXT *surface_present_mode = LvlFindInChain<VkSurfacePresentModeEXT>(pSurfaceInfo->pNext);
+        if ((!device_extensions.vk_ext_surface_maintenance1) || (!surface_present_mode)) {
```
jeremyg-lunarg, line 4133: IsExtEnabled()

layers/state_tracker.cpp
```diff
@@ -3845,6 +3851,43 @@ void ValidationStateTracker::PostCallRecordQueuePresentKHR(VkQueue queue, const
     }
 }
 
+void ValidationStateTracker::PostCallRecordReleaseSwapchainImagesEXT(VkDevice device, const VkReleaseSwapchainImagesInfoEXT *pReleaseInfo, VkResult result) {
+
+    if (result == VK_ERROR_DEVICE_LOST) {
+        return;
+    }
+    /*
```
jeremyg-lunarg, line 3859: We don't really have a way to handle pReleaseInfo correctly here at the present. To do it right, I think we need to store the fences and semaphores associated with each swapchain image in either SWAPCHAIN_NODE or IMAGE_STATE and then use this call as another way to call Retire() on the fences and semaphores when it succeeds.

layers/state_tracker.cpp
mark-lunarg: *sweats*. Is it right that I can either do a reserve + copy here, or do a copy with  `back_inserter()`?

layers/state_tracker.cpp
mark-lunarg: I had forgotten all about this function, sigh.  I will give it a go but look out.

layers/state_tracker.cpp
jeremyg-lunarg: yep either of those should work

layers/image_state.h
```diff
@@ -370,13 +377,26 @@ class SURFACE_STATE : public BASE_NODE {
...
 
   private:
     std::unique_lock<std::mutex> Lock() const { return std::unique_lock<std::mutex>(lock_); }
     mutable std::mutex lock_;
     mutable layer_data::unordered_map<GpuQueue, bool> gpu_queue_support_;
-    mutable layer_data::unordered_map<VkPhysicalDevice, std::vector<VkPresentModeKHR>> present_modes_;
     mutable layer_data::unordered_map<VkPhysicalDevice, std::vector<VkSurfaceFormatKHR>> formats_;
     mutable layer_data::unordered_map<VkPhysicalDevice, VkSurfaceCapabilitiesKHR> capabilities_;
+    mutable layer_data::unordered_map<VkPhysicalDevice, std::vector<VkPresentModeKHR>> present_modes_;
+    mutable layer_data::unordered_map<VkPhysicalDevice,
```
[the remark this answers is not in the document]
mark-lunarg: I think there is a catch -- differing `min/maxImageCounts` can be returned if the extension structs are chained.  So for any given `[pdev][presentMode]`, even if the extension is enabled those values will be different depending on how it gets called.  We could unify the maps and add some state indicating which values should be returned, but at the cost of some additional complexity. In practice how many apps are going to simultaneously create multiple swapchains with and without using the extension, but I guess they could?  I'm happy to go either way!

mark-lunarg:
@spencer-lunarg, @juan-lunarg, @ncesario-lunarg, Jeremy thinks this is far enough along to take off the WIP.  I'd appreciate any attention you can afford.  We did decide to handle the swapchain state_tracker/synchronization updates in a separate PR though the validation is ostensibly covered. Dang, copyright dates!  Somebody should redo that script.

jeremyg-lunarg:
If you rebase this on to tip of master, you should get the new copyright script and clear the errors without changing them all...

layers/state_tracker.cpp
mark-lunarg: This will be handled in a follow-on PR.

layers/core_validation.cpp
```diff
@@ -18314,12 +18303,263 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
...
+                if (LogError(device, "VUID-VkSwapchainCreateInfoKHR-imageExtent-01274",
+                             "%s called with imageExtent = (%" PRIu32 ",%" PRIu32 "), which is outside the bounds returned by "
+                             "vkGetPhysicalDeviceSurfaceCapabilitiesKHR(): currentExtent = (%" PRIu32 ",%" PRIu32
+                             "), minImageExtent = (%" PRIu32 ",%" PRIu32 "), "
+                             "maxImageExtent = (%" PRIu32 ",%" PRIu32 ").",
+                             func_name, pCreateInfo->imageExtent.width, pCreateInfo->imageExtent.height,
+                             capabilities.currentExtent.width, capabilities.currentExtent.height, capabilities.minImageExtent.width,
+                             capabilities.minImageExtent.height, capabilities.maxImageExtent.width,
+                             capabilities.maxImageExtent.height)) {
+                    return true;
```
ncesario-lunarg, line 18337: Probably not necessary due to precedent in this file, but this is kind of abnormal and might be nice to have a comment indicating why no other validation in this function should run. Same comment for similar instances below.

layers/core_validation.cpp
```diff
@@ -18314,12 +18303,263 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
...
+        if (!IsExtentInsideBounds(pCreateInfo->imageExtent, capabilities.minImageExtent, capabilities.maxImageExtent)) {
+            VkSurfaceCapabilitiesKHR cached_capabilities{};
+            if (surface_state) {
+                cached_capabilities = surface_state->GetCapabilities(physical_device);
+            } else if (IsExtEnabled(instance_extensions.vk_google_surfaceless_query)) {
+                cached_capabilities = physical_device_state->surfaceless_query_state.capabilities;
+            }
+            if (!IsExtentInsideBounds(pCreateInfo->imageExtent, cached_capabilities.minImageExtent,
+                                      cached_capabilities.maxImageExtent)) {
+                if (LogError(device, "VUID-VkSwapchainCreateInfoKHR-imageExtent-01274",
```
ncesario-lunarg, line 18328: I'm probably just being ignorant here, but why is the usual `skip |= LogError(...` not used here and below?

layers/core_validation.cpp
```diff
@@ -18314,12 +18303,263 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
...
+                                "VkSurfacePresentScalingCapabilitiesEXT::supportedPresentGravityY for pPresentModes[%d]: (%s).",
+                                func_name, string_VkPresentGravityFlagsEXT(pres_scale_ci->presentGravityY).c_str(), i,
+                                string_VkPresentGravityFlagsEXT(scaling_caps.supportedPresentGravityY).c_str())) {
+                            skip |= true;
+                        }
+                    }
+                }
+            }
         }
+        if (skip) return true;
```
ncesario-lunarg, line 18562: Shouldn't this return skip? If it shouldn't (??), don't we need an `else`, or something to indicate what is returned? As it stands now, if `!skip`, seems like the return value is undefined.

layers/core_validation.cpp
```diff
@@ -18712,11 +18952,102 @@ bool CoreChecks::PreCallValidateQueuePresentKHR(VkQueue queue, const VkPresentIn
...
+                        pPresentInfo->pSwapchains[0], "VUID-VkSwapchainPresentFenceInfoEXT-swapchainCount-07757",
+                        "vkQueuePresentKHR(): VkSwapchainPresentFenceInfoEXT.swapchainCount is %i but pPresentInfo->swapchainCount "
+                        "is %i. For VkSwapchainPresentFenceInfoEXT in pNext chain of VkPresentInfoKHR, "
+                        "VkSwapchainPresentFenceInfoEXT.swapchainCount must equal VkPresentInfoKHR.swapchainCount.",
+                        swapchain_present_fence_info->swapchainCount, pPresentInfo->swapchainCount);
+                }
+
+                for (uint32_t i = 0; i < swapchain_present_fence_info->swapchainCount; i++) {
+                    if (swapchain_present_fence_info->pFences[i]) {
+                        auto fence_state = Get<FENCE_STATE>(swapchain_present_fence_info->pFences[i]);
```
ncesario-lunarg, line 18970: This can probably be `const`. I only noticed since you were much more thorough than I and used `const` in your function parameter definition, which I always forget to do :)

layers/parameter_validation_utils.cpp
```diff
@@ -9268,6 +9268,46 @@ bool StatelessValidation::manual_PreCallValidateGetPhysicalDeviceSurfaceCapabili
         }
     }
 #endif
+
+    //if (instance_extensions.vk_ext_surface_maintenance1) {
```
ncesario-lunarg, line 9272: ?

tests/vklayertests_wsi.cpp
```diff
@@ -2122,6 +2123,661 @@ TEST_F(VkLayerTest, TestSurfaceSupportByPhysicalDevice) {
...
+
+    AddRequiredExtensions(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
+    AddRequiredExtensions(VK_EXT_SURFACE_MAINTENANCE_1_EXTENSION_NAME);
+    AddRequiredExtensions(VK_KHR_GET_SURFACE_CAPABILITIES_2_EXTENSION_NAME);
+    AddRequiredExtensions(VK_EXT_SWAPCHAIN_MAINTENANCE_1_EXTENSION_NAME);
+    AddSurfaceExtension();
+
+    ASSERT_NO_FATAL_FAILURE(InitFramework(m_errorMonitor));
+    if (DeviceValidationVersion() < VK_API_VERSION_1_0) {
+        GTEST_SKIP() << "At least Vulkan version 1.1 is required";
```
ncesario-lunarg, line 2138: `VK_API_VERSION_1_0` -> `VK_API_VERSION_1_1` above?

ncesario-lunarg (commented):
Some nits and comments, but LGTM!

layers/core_validation.cpp
```diff
@@ -18314,12 +18303,263 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
...
+        // Validate VkSwapchainPresentModesCreateInfoEXT data
+        auto swapchain_present_modes_ci = LvlFindInChain<VkSwapchainPresentModesCreateInfoEXT>(pCreateInfo->pNext);
+        if (swapchain_present_modes_ci) {
+            bool found_swapchain_modes_ci_present_mode = false;
+            for (uint32_t i = 0; i < swapchain_present_modes_ci->presentModeCount; i++) {
+                VkPresentModeKHR swapchain_present_mode = swapchain_present_modes_ci->pPresentModes[i];
+
+                bool found_match =
+                    std::find(present_modes.begin(), present_modes.end(), swapchain_present_mode) != present_modes.end();
+                if (!found_match) {
```
juan-lunarg, line 18351: C++17 if statement. Reads better and keeps tight scope. See above example.

layers/core_validation.cpp
```diff
@@ -18314,12 +18303,263 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
...
+                                 func_name, i, string_VkPresentModeKHR(swapchain_present_mode))) {
+                        skip |= true;
+                    }
+                }
+
+                std::vector<VkPresentModeKHR> compatible_present_modes =
+                    surface_state->GetCompatibleModes(physical_device, present_mode);
+                found_match = std::find(compatible_present_modes.begin(), compatible_present_modes.end(), swapchain_present_mode) !=
+                              compatible_present_modes.end();
+                if (!found_match) {
```
juan-lunarg, line 18364: C++17 if statement. Reads better and keeps tight scope. See above example.

layers/core_validation.cpp
```diff
@@ -18314,12 +18303,263 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
...
+            bool found_swapchain_modes_ci_present_mode = false;
+            for (uint32_t i = 0; i < swapchain_present_modes_ci->presentModeCount; i++) {
+                VkPresentModeKHR swapchain_present_mode = swapchain_present_modes_ci->pPresentModes[i];
+
+                bool found_match =
+                    std::find(present_modes.begin(), present_modes.end(), swapchain_present_mode) != present_modes.end();
+                if (!found_match) {
+                    if (LogError(device, "VUID-VkSwapchainPresentModesCreateInfoEXT-None-07762",
+                                 "%s was called with vkSwapchainPresentModesCreateInfoEXT structure in its pNext chain that "
+                                 "contains a non-supported presentMode in pPresentModes[%d]: (%s).",
```
juan-lunarg, line 18354: Use PRIu32 format specifier

layers/core_validation.cpp
```diff
@@ -18314,12 +18303,263 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
...
+
+                std::vector<VkPresentModeKHR> compatible_present_modes =
+                    surface_state->GetCompatibleModes(physical_device, present_mode);
+                found_match = std::find(compatible_present_modes.begin(), compatible_present_modes.end(), swapchain_present_mode) !=
+                              compatible_present_modes.end();
+                if (!found_match) {
+                    if (LogError(device, "VUID-VkSwapchainPresentModesCreateInfoEXT-pPresentModes-07763",
+                                 "%s was called with vkSwapchainPresentModesCreateInfoEXT structure in its pNext chain that "
+                                 "contains a non-compatible presentMode in pPresentModes[%d]: (%s).",
+                                 func_name, i, string_VkPresentModeKHR(swapchain_present_mode))) {
```
juan-lunarg, line 18368: Use PRIu32 format specifier

layers/core_validation.cpp
```diff
@@ -18314,12 +18303,263 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
...
+                        skip |= true;
+                    }
+                }
+
+                if (swapchain_present_modes_ci->pPresentModes[i] == present_mode) {
+                    found_swapchain_modes_ci_present_mode = true;
+                }
```
juan-lunarg, line 18375: Avoid if statement with `|=`

```suggestion
                const bool has_present_mode = (swapchain_present_modes_ci->pPresentModes[i] == present_mode);
                found_swapchain_modes_ci_present_mode |= has_present_mode;
```

layers/core_validation.cpp
mark-lunarg: I will verify, but IIRC any accumulated errors cause bad behavior below, so if there are errors we return otherwise we keep validating.

layers/core_validation.cpp
mark-lunarg: Sure, I'll audit the error handling.

layers/core_validation.cpp
mark-lunarg: I noticed that the default for the WSI-related checks was mostly one-and-done -- any single validation error results in the validation routine exiting immediately.  My theory was that if just about any of this stuff fails your app is hosed anyway.  I did play some games because the new extension validation checks were so interrelated -- switching to all skips or all returns would mean a slew of `SetUnexpectedError()` calls in the tests, so now they're kinda grouped by related failure groups.

layers/core_validation.cpp
```diff
@@ -18314,12 +18303,263 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
...
+                        "the vkSwapchainPresentModesCreateInfoEXT structure included in its pNext chain.",
+                        func_name, string_VkPresentModeKHR(present_mode))) {
+                    skip |= true;
+                }
+            }
+            if (skip) return true;
+        }
+
+        // Validate VkSwapchainPresentScalingCreateInfoEXT data
+        auto pres_scale_ci = LvlFindInChain<VkSwapchainPresentScalingCreateInfoEXT>(pCreateInfo->pNext);
```
juan-lunarg, line 18390: Given the amount of code a helper function for validating `VkSwapchainPresentScalingCreateInfoEXT` would be nice (and reduce indentation).

juan-lunarg (changes requested):

layers/core_validation.cpp
```diff
@@ -17320,11 +17320,23 @@ bool CoreChecks::ValidateBindImageMemory(uint32_t bindInfoCount, const VkBindIma
...
+                                         swapchain_info->imageIndex, report_data->FormatHandle(swapchain_info->swapchain).c_str(),
+                                         static_cast<int>(swapchain_state->images.size()));
+                    }
+                    if (device_extensions.vk_ext_swapchain_maintenance1 &&
+                        (swapchain_state->createInfo.flags & VK_SWAPCHAIN_CREATE_DEFERRED_MEMORY_ALLOCATION_BIT_EXT)) {
+                        if (swapchain_state->images[swapchain_info->imageIndex].acquired == false) {
+                            skip |= LogError(
+                                bind_info.image, "VUID-VkBindImageMemorySwapchainInfoKHR-swapchain-07756",
+                                "%s: The swapchain was created with VK_SWAPCHAIN_CREATE_DEFERRED_MEMORY_ALLOCATION_BIT_EXT but "
+                                "imageIndex (%i) has not been acquired",
```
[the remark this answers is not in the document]
juan-lunarg: removing suggestion since it could become out of date with a spec update (didn't think about that).

layers/core_validation.cpp
```diff
@@ -17320,11 +17320,23 @@ bool CoreChecks::ValidateBindImageMemory(uint32_t bindInfoCount, const VkBindIma
...
-                                     swapchain_info->imageIndex, report_data->FormatHandle(swapchain_info->swapchain).c_str(),
-                                     static_cast<int>(swapchain_state->images.size()));
+                if (swapchain_state) {
+                    if (swapchain_state->images.size() <= swapchain_info->imageIndex) {
+                        skip |= LogError(bind_info.image, "VUID-VkBindImageMemorySwapchainInfoKHR-imageIndex-01644",
+                                         "%s: imageIndex (%i) is out of bounds of %s images (size: %i)", error_prefix,
+                                         swapchain_info->imageIndex, report_data->FormatHandle(swapchain_info->swapchain).c_str(),
+                                         static_cast<int>(swapchain_state->images.size()));
+                    }
+                    if (device_extensions.vk_ext_swapchain_maintenance1 &&
```
spencer-lunarg, line 17330: for all the `device_extensions.vk_*` in this PR, please use the`IsExtEnabled` wrapper for it

layers/core_validation.cpp
```diff
@@ -18314,12 +18303,263 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
...
+
+            if (GetBitSetCount(pres_scale_ci->presentGravityY) > 1) {
+                if (LogError(device, "VUID-VkSwapchainPresentScalingCreateInfoEXT-presentGravityY-07769",
+                             "%s: VkSwapchainPresentScalingCreateInfoEXT is present in the pNext chain and presentGravityY (%d) "
+                             "must not have more than one bit set.",
+                             func_name, pres_scale_ci->presentGravityY)) {
+                    skip |= true;
+                }
+            }
+            if (skip) return true;
```
spencer-lunarg, line 18454: There are a few early returns in this function, was this on purpose, is there a reason we shouldn't keep checking (if so not 100% obvious and maybe worth a comment)

mark-lunarg:
Thanks to all the great review input -- there was a regression running with swiftshader (CI doesn't really touch any of this) and I'm almost done combining the `present_modes_` maps in state_tracker.  I'll be OOO 4 of the next 5 or 6 days but plan to address all of the feedback ASAP.

layers/core_validation.cpp
juan-lunarg, line 18351:
```suggestion
                if (std::find(present_modes.begin(), present_modes.end(), swapchain_present_mode) == present_modes.end()) {
```

layers/core_validation.cpp
juan-lunarg, line 18364:
```suggestion
                if (std::find(compatible_present_modes.begin(), compatible_present_modes.end(), swapchain_present_mode) == compatible_present_modes.end()) {
```

layers/core_validation.cpp
```diff
@@ -18314,12 +18303,263 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
...
+                    if (LogError(device, "VUID-VkSwapchainPresentModesCreateInfoEXT-None-07762",
+                                 "%s was called with vkSwapchainPresentModesCreateInfoEXT structure in its pNext chain that "
+                                 "contains a non-supported presentMode in pPresentModes[%d]: (%s).",
+                                 func_name, i, string_VkPresentModeKHR(swapchain_present_mode))) {
+                        skip |= true;
+                    }
+                }
+
+                std::vector<VkPresentModeKHR> compatible_present_modes =
+                    surface_state->GetCompatibleModes(physical_device, present_mode);
```
juan-lunarg, line 18361:
```suggestion
                const std::vector<VkPresentModeKHR> compatible_present_modes =
                    surface_state->GetCompatibleModes(physical_device, present_mode);
```

juan-lunarg (changes requested):

layers/core_validation.cpp
```diff
@@ -17320,11 +17320,23 @@ bool CoreChecks::ValidateBindImageMemory(uint32_t bindInfoCount, const VkBindIma
...
-                    skip |= LogError(bind_info.image, "VUID-VkBindImageMemorySwapchainInfoKHR-imageIndex-01644",
-                                     "%s: imageIndex (%i) is out of bounds of %s images (size: %i)", error_prefix,
-                                     swapchain_info->imageIndex, report_data->FormatHandle(swapchain_info->swapchain).c_str(),
-                                     static_cast<int>(swapchain_state->images.size()));
+                if (swapchain_state) {
+                    if (swapchain_state->images.size() <= swapchain_info->imageIndex) {
+                        skip |= LogError(bind_info.image, "VUID-VkBindImageMemorySwapchainInfoKHR-imageIndex-01644",
+                                         "%s: imageIndex (%i) is out of bounds of %s images (size: %i)", error_prefix,
+                                         swapchain_info->imageIndex, report_data->FormatHandle(swapchain_info->swapchain).c_str(),
+                                         static_cast<int>(swapchain_state->images.size()));
```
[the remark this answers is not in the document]
ncesario-lunarg: `%z` == `size_t`? If that's true, I don't think there is a `PRIz` to use.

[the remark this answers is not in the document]
juan-lunarg: Agreed

layers/image_state.h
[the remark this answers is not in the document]
mark-lunarg: Done.

layers/core_validation.cpp
```diff
@@ -17320,11 +17320,23 @@ bool CoreChecks::ValidateBindImageMemory(uint32_t bindInfoCount, const VkBindIma
...
+                    }
+                    if (device_extensions.vk_ext_swapchain_maintenance1 &&
+                        (swapchain_state->createInfo.flags & VK_SWAPCHAIN_CREATE_DEFERRED_MEMORY_ALLOCATION_BIT_EXT)) {
+                        if (swapchain_state->images[swapchain_info->imageIndex].acquired == false) {
+                            skip |= LogError(
+                                bind_info.image, "VUID-VkBindImageMemorySwapchainInfoKHR-swapchain-07756",
+                                "%s: The swapchain was created with VK_SWAPCHAIN_CREATE_DEFERRED_MEMORY_ALLOCATION_BIT_EXT but "
+                                "imageIndex (%i) has not been acquired",
+                                error_prefix, swapchain_info->imageIndex);
```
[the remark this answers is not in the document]
mark-lunarg: Done

layers/core_validation.cpp
mark-lunarg: Done

layers/core_validation.cpp
mark-lunarg: Done

layers/core_validation.cpp
mark-lunarg: Done

layers/core_validation.cpp
mark-lunarg: Done

layers/core_validation.cpp
mark-lunarg: Done.  Uncomfortable, but done.

layers/core_validation.cpp
mark-lunarg: Done

layers/parameter_validation_utils.cpp
mark-lunarg: Done

tests/vklayertests_wsi.cpp
mark-lunarg: Done

layers/core_validation.cpp
ncesario-lunarg: Sounds like we should `assert(skip)` at this point then, and return `skip`? If `!skip` here, that sill seems like it's going to be undefined behavior, but maybe I'm missing something.

ncesario-lunarg:
> ... Jeremy thinks this is far enough along to take off the WIP. I'd appreciate any attention you can afford. We did decide to handle the swapchain state_tracker/synchronization updates in a separate PR though the validation is ostensibly covered.

Sounds good to me! Though I think there are still some outstanding comments that need to be resolved one way or the other.

> Dang, copyright dates! Somebody should redo that script.

I believe @jeremyg-lunarg already did?

layers/core_validation.cpp
ncesario-lunarg: Just ignore me. For some silly reason the diff I was looking at made it look as though the `}` on the next line was the end of the function.

layers/core_validation.cpp
mark-lunarg: Done.

layers/core_validation.cpp
```diff
@@ -18314,12 +18303,263 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
...
+        }
+    }
+
+    if (!device_extensions.vk_ext_swapchain_maintenance1) {
+        // Validate pCreateInfo->imageExtent against VkSurfaceCapabilitiesKHR::{current|min|max}ImageExtent:
+        if (!IsExtentInsideBounds(pCreateInfo->imageExtent, capabilities.minImageExtent, capabilities.maxImageExtent)) {
+            VkSurfaceCapabilitiesKHR cached_capabilities{};
+            if (surface_state) {
+                cached_capabilities = surface_state->GetCapabilities(physical_device);
+            } else if (IsExtEnabled(instance_extensions.vk_google_surfaceless_query)) {
```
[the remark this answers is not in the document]
mark-lunarg: I believe so, this was not changed other than to wrap it in a `!IsExtEnabled(swapchain_ext)` conditional.

layers/core_validation.cpp
mark-lunarg: Got rid of all the early returns.

layers/core_validation.cpp
mark-lunarg: Switched to normal skip error returns.

layers/core_validation.cpp
mark-lunarg: Removed early returns.

mark-lunarg:
@jeremyg-lunarg, you might take a look at the changes for combining the `present_mode` maps using `std::optional` (in a separate commit)-- CI choked on the 'old' MacOS build saying it didn't support `value()`.  My workarounds for this are in the final commit but I'm not confident there isn't a better way to fix things.

layers/image_state.cpp
```diff
@@ -762,3 +769,117 @@ VkSurfaceCapabilitiesKHR SURFACE_STATE::GetCapabilities(VkPhysicalDevice phys_de
     capabilities_[phys_dev] = result;
     return result;
 }
+
+void SURFACE_STATE::SetCompatibleModes(VkPhysicalDevice phys_dev, const VkPresentModeKHR present_mode,
+                                       std::vector<VkPresentModeKHR> &&compatible_modes) {
```
juan-lunarg, line 774: This should also use a span. Avoids creating temp std::vector at call site.

```suggestion
void SURFACE_STATE::SetCompatibleModes(VkPhysicalDevice phys_dev, const VkPresentModeKHR present_mode,
                                       layer_data::span<const VkPresentModeKHR> compatible_modes) {
```

layers/image_state.cpp
```diff
@@ -762,3 +769,117 @@ VkSurfaceCapabilitiesKHR SURFACE_STATE::GetCapabilities(VkPhysicalDevice phys_de
...
+    auto guard = Lock();
+    assert(phys_dev);
+
+    // If this surface or the present_mode is not in the map, create and add the present_mode state
+    auto surface_map = present_modes_data_.find(phys_dev);
+    if ((surface_map == present_modes_data_.end()) || (surface_map->second.find(present_mode) == surface_map->second.end())) {
+        auto present_mode_state = std::make_shared<PresentModeState>();
+        present_mode_state->compatible_present_modes_ = compatible_modes;
```
juan-lunarg, line 782: NOTE: compatible_modes won't be moved it will be copied. And if it was moved it the following for loop would be invalid.

layers/image_state.cpp
```diff
@@ -704,17 +704,24 @@ bool SURFACE_STATE::GetQueueSupport(VkPhysicalDevice phys_dev, uint32_t qfi) con
...
 
 std::vector<VkPresentModeKHR> SURFACE_STATE::GetPresentModes(VkPhysicalDevice phys_dev) const {
     auto guard = Lock();
     assert(phys_dev);
-    auto iter = present_modes_.find(phys_dev);
-    if (iter != present_modes_.end()) {
-        return iter->second;
-    }
     std::vector<VkPresentModeKHR> result;
+    if (present_modes_data_.find(phys_dev) != present_modes_data_.end()) {
```
juan-lunarg, line 719: Add 1 line comment. Since I would assume by default this function is just a helper for `DispatchGetPhysicalDeviceSurfacePresentModesKHR`

layers/image_state.cpp
```diff
@@ -762,3 +769,117 @@ VkSurfaceCapabilitiesKHR SURFACE_STATE::GetCapabilities(VkPhysicalDevice phys_de
...
+void SURFACE_STATE::SetPresentModeCapabilities(VkPhysicalDevice phys_dev, const VkPresentModeKHR present_mode,
+                                               const VkSurfaceCapabilitiesKHR &caps,
+                                               const VkSurfacePresentScalingCapabilitiesEXT &scaling_caps) {
+    auto guard = Lock();
+    assert(phys_dev);
+    if (!present_modes_data_[phys_dev][present_mode].has_value()) {
+        present_modes_data_[phys_dev][present_mode] = std::make_shared<PresentModeState>();
+    }
+    // Old MacOS doesn't support Value()
+    auto &present_mode_state = *(present_modes_data_[phys_dev][present_mode]);
```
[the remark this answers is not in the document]
mark-lunarg: Used temp var to make clang-tidy happy.

layers/image_state.cpp
```diff
@@ -762,3 +769,117 @@ VkSurfaceCapabilitiesKHR SURFACE_STATE::GetCapabilities(VkPhysicalDevice phys_de
...
+    return surface_capabilities.surfaceCapabilities;
+}
+
+// Get the scaling capabilities for this particular present mode
+VkSurfacePresentScalingCapabilitiesEXT SURFACE_STATE::GetPresentModeScalingCapabilities(VkPhysicalDevice phys_dev,
+                                                                                        const VkPresentModeKHR present_mode) const {
+    auto iter = present_modes_data_.find(phys_dev);
+    if ((iter != present_modes_data_.end()) && (iter->second.find(present_mode) != iter->second.end())) {
+        if ((iter->second)[present_mode].has_value()) {
+            auto &scaling_caps = *((iter->second)[present_mode]);
```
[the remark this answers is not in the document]
mark-lunarg: Fixed.

layers/image_state.cpp
```diff
@@ -762,3 +769,117 @@ VkSurfaceCapabilitiesKHR SURFACE_STATE::GetCapabilities(VkPhysicalDevice phys_de
...
+    VkSurfaceCapabilities2KHR surface_capabilities = LvlInitStruct<VkSurfaceCapabilities2KHR>();
+    surface_capabilities.pNext = &scaling_caps;
+    DispatchGetPhysicalDeviceSurfaceCapabilities2KHR(phys_dev, &surface_info, &surface_capabilities);
+    return scaling_caps;
+}
+
+
+
+
+
```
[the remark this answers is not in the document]
mark-lunarg: Fixed.

layers/image_state.cpp
```diff
@@ -704,17 +704,24 @@ bool SURFACE_STATE::GetQueueSupport(VkPhysicalDevice phys_dev, uint32_t qfi) con
 void SURFACE_STATE::SetPresentModes(VkPhysicalDevice phys_dev, std::vector<VkPresentModeKHR> &&modes) {
```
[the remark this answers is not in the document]
mark-lunarg: Done

layers/image_state.cpp
```diff
@@ -704,17 +704,24 @@ bool SURFACE_STATE::GetQueueSupport(VkPhysicalDevice phys_dev, uint32_t qfi) con
...
     auto guard = Lock();
     assert(phys_dev);
-    present_modes_[phys_dev] = std::move(modes);
+    for (auto new_present_mode : modes) {
+        if ((present_modes_data_.find(phys_dev) == present_modes_data_.end()) ||
+            (present_modes_data_[phys_dev].find(new_present_mode) == present_modes_data_[phys_dev].end())) {
+            present_modes_data_[phys_dev][new_present_mode] = std::nullopt;
```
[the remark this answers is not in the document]
mark-lunarg: I believe so, yes.  It just sets the present modes for the legacy API and does nothing else.

layers/image_state.cpp
mark-lunarg: Done

layers/image_state.cpp
mark-lunarg: Done.

layers/image_state.cpp
```diff
@@ -762,3 +769,117 @@ VkSurfaceCapabilitiesKHR SURFACE_STATE::GetCapabilities(VkPhysicalDevice phys_de
...
+            auto &compatible_modes = *(iter->second)[present_mode];
+            if (compatible_modes->compatible_present_modes_.size() != 0) {
+                return compatible_modes->compatible_present_modes_;
+            }
+        }
+    }
+
+    // Compatible modes not in state tracker, call to get compatible modes
+    std::vector<VkPresentModeKHR> result;
+    VkPhysicalDeviceSurfaceInfo2KHR surface_info = LvlInitStruct<VkPhysicalDeviceSurfaceInfo2KHR>();
```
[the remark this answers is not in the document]
mark-lunarg: Fixed all of these guys, thanks!

layers/parameter_validation_utils.cpp
```diff
@@ -9294,6 +9294,45 @@ bool StatelessValidation::manual_PreCallValidateGetPhysicalDeviceSurfaceCapabili
...
     }
 #endif
+
+    if (IsExtEnabled(device_extensions.vk_ext_surface_maintenance1)) {
+            const auto *surface_present_mode_compatibility =
+            LvlFindInChain<VkSurfacePresentModeCompatibilityEXT>(pSurfaceCapabilities->pNext);
+        const auto *surface_present_scaling_compatibilities =
+            LvlFindInChain<VkSurfacePresentScalingCapabilitiesEXT>(pSurfaceCapabilities->pNext);
+
+        if (surface_present_mode_compatibility && !(LvlFindInChain<VkSurfacePresentModeEXT>(pSurfaceInfo->pNext))) {
```
spencer-lunarg, line 9304: nit: Can group both skips together for both

- `if !(LvlFindInChain<VkSurfacePresentModeEXT>(pSurfaceInfo->pNext))`
- `if (pSurfaceInfo->surface == VK_NULL_HANDLE)`

mark-lunarg: Done.

mark-lunarg:
@juan-lunarg, what else can I update for y'all?

layers/parameter_validation_utils.cpp
juan-lunarg, line 9304: Minor code duplication regarding `!(LvlFindInChain<VkSurfacePresentModeEXT>(pSurfaceInfo->pNext)))`  lines 9304 and 9312

layers/parameter_validation_utils.cpp
```diff
@@ -9294,6 +9294,45 @@ bool StatelessValidation::manual_PreCallValidateGetPhysicalDeviceSurfaceCapabili
...
+        if (surface_present_scaling_compatibilities && !(LvlFindInChain<VkSurfacePresentModeEXT>(pSurfaceInfo->pNext))) {
+            skip |= LogError(device, "VUID-vkGetPhysicalDeviceSurfaceCapabilities2KHR-pNext-07777",
+                             "vkGetPhysicalDeviceSurfaceCapabilities2KHR(): VK_EXT_surface_maintenance1 is enabled and "
+                             "pSurfaceCapabilities->pNext contains "
+                             "VkSurfacePresentScalingCapabilitiesEXT, but pSurfaceInfo->pNext does not contain a "
+                             "VkSurfacePresentModeEXT structure.");
+        }
+
+        if (IsExtEnabled(instance_extensions.vk_google_surfaceless_query)) {
+            if (surface_present_mode_compatibility && (pSurfaceInfo->surface == VK_NULL_HANDLE)) {
```
juan-lunarg, line 9321: Minor code duplication regarding ```(pSurfaceInfo->surface == VK_NULL_HANDLE))``` lines 9321 and 9329

layers/parameter_validation_utils.cpp
```diff
@@ -9294,6 +9294,45 @@ bool StatelessValidation::manual_PreCallValidateGetPhysicalDeviceSurfaceCapabili
         }
     }
 #endif
+
+    if (IsExtEnabled(device_extensions.vk_ext_surface_maintenance1)) {
+            const auto *surface_present_mode_compatibility =
```
juan-lunarg, line 9299: Minor indentation issue? Surprised clang-format didn't complain.
```suggestion
        const auto *surface_present_mode_compatibility =
```

layers/state_tracker.cpp
```diff
@@ -3872,6 +3872,13 @@ void ValidationStateTracker::RecordCreateSwapchainState(VkResult result, const V
         surface_state->AddParent(swapchain.get());
         surface_state->swapchain = swapchain.get();
         swapchain->surface = std::move(surface_state);
+        auto swapchain_present_modes_ci = LvlFindInChain<VkSwapchainPresentModesCreateInfoEXT>(pCreateInfo->pNext);
+        if (swapchain_present_modes_ci) {
+            uint32_t present_mode_count = swapchain_present_modes_ci->presentModeCount;
```
juan-lunarg, line 3877:
```suggestion
            const uint32_t present_mode_count = swapchain_present_modes_ci->presentModeCount;
```

layers/wsi_validation.cpp
```diff
@@ -85,6 +85,236 @@ static VkImageCreateInfo GetSwapchainImpliedImageCreateInfo(VkSwapchainCreateInf
...
+                if (LogError(device, "VUID-VkSwapchainPresentModesCreateInfoEXT-None-07762",
+                             "%s was called with vkSwapchainPresentModesCreateInfoEXT structure in its pNext chain that "
+                             "contains a non-supported presentMode in pPresentModes[%" PRIu32 "]: (%s).",
+                             func_name, i, string_VkPresentModeKHR(swapchain_present_mode))) {
+                    skip |= true;
+                }
+            }
+
+            const std::vector<VkPresentModeKHR> compatible_present_modes =
+                surface_state->GetCompatibleModes(physical_device, present_mode);
```
juan-lunarg, line 121: This vector shouldn't be constructed in the for loop. And surface_state needs to be checked nullptr.

layers/wsi_validation.cpp
```diff
@@ -85,6 +85,236 @@ static VkImageCreateInfo GetSwapchainImpliedImageCreateInfo(VkSwapchainCreateInf
...
+                             capabilities->minImageExtent.width, capabilities->minImageExtent.height,
+                             capabilities->maxImageExtent.width, capabilities->maxImageExtent.height);
+        }
+    }
+
+    if (pres_scale_ci) {
+        if ((pres_scale_ci->presentGravityX == 0) && (pres_scale_ci->presentGravityY != 0)) {
+            if (LogError(device, "VUID-VkSwapchainPresentScalingCreateInfoEXT-presentGravityX-07765",
+                         "%s: VkSwapchainPresentScalingCreateInfoEXT  is present in the pNext chain and if presentGravityX is zero "
+                         "(%d) then presentGravityY (%d) must also be zero.",
```
juan-lunarg, line 171: Use PRIu32 (VkPresentGravityFlagsEXT is really just a VkFlag which is a uint32_t)

layers/wsi_validation.cpp
```diff
@@ -85,6 +85,236 @@ static VkImageCreateInfo GetSwapchainImpliedImageCreateInfo(VkSwapchainCreateInf
...
+                         func_name, pres_scale_ci->presentGravityX, pres_scale_ci->presentGravityY)) {
+                skip |= true;
+            }
+        }
+
+        if ((pres_scale_ci->presentGravityX != 0) && (pres_scale_ci->presentGravityY == 0)) {
+            if (LogError(device, "VUID-VkSwapchainPresentScalingCreateInfoEXT-presentGravityX-07766",
+                         "%s: VkSwapchainPresentScalingCreateInfoEXT is present in the pNext chain and if presentGravityX is "
+                         "non-zero (%d) then presentGravityY (%d) must also be non-zero.",
+                         func_name, pres_scale_ci->presentGravityX, pres_scale_ci->presentGravityY)) {
```
juan-lunarg, line 181: Use PRIu32 (VkPresentGravityFlagsEXT is really just a VkFlag which is a uint32_t)

layers/wsi_validation.cpp
```diff
@@ -85,6 +85,236 @@ static VkImageCreateInfo GetSwapchainImpliedImageCreateInfo(VkSwapchainCreateInf
...
+                skip |= true;
+            }
+        }
+
+        if (GetBitSetCount(pres_scale_ci->scalingBehavior) > 1) {
+            if (LogError(device, "VUID-VkSwapchainPresentScalingCreateInfoEXT-scalingBehavior-07767",
+                         "%s: VkSwapchainPresentScalingCreateInfoEXT is present in the pNext chain and scalingBehavior (%d) "
```
juan-lunarg, line 188: Use PRIu32 (VkPresentGravityFlagsEXT is really just a VkFlag which is a uint32_t)

layers/wsi_validation.cpp
```diff
@@ -85,6 +85,236 @@ static VkImageCreateInfo GetSwapchainImpliedImageCreateInfo(VkSwapchainCreateInf
...
+                         "must not have more than one bit set.",
+                         func_name, pres_scale_ci->presentGravityX)) {
+                skip |= true;
+            }
+        }
+
+        if (GetBitSetCount(pres_scale_ci->presentGravityX) > 1) {
+            if (LogError(device, "VUID-VkSwapchainPresentScalingCreateInfoEXT-presentGravityX-07768",
+                         "%s: VkSwapchainPresentScalingCreateInfoEXT is present in the pNext chain and presentGravityX (%d) "
```
juan-lunarg, line 197: Use PRIu32 (VkPresentGravityFlagsEXT is really just a VkFlag which is a uint32_t)

layers/wsi_validation.cpp
```diff
@@ -85,6 +85,236 @@ static VkImageCreateInfo GetSwapchainImpliedImageCreateInfo(VkSwapchainCreateInf
...
+                         "must not have more than one bit set.",
+                         func_name, pres_scale_ci->presentGravityX)) {
+                skip |= true;
+            }
+        }
+
+        if (GetBitSetCount(pres_scale_ci->presentGravityY) > 1) {
+            if (LogError(device, "VUID-VkSwapchainPresentScalingCreateInfoEXT-presentGravityY-07769",
+                         "%s: VkSwapchainPresentScalingCreateInfoEXT is present in the pNext chain and presentGravityY (%d) "
```
juan-lunarg, line 206: Use PRIu32 (VkPresentGravityFlagsEXT is really just a VkFlag which is a uint32_t)

layers/wsi_validation.cpp
```diff
@@ -85,6 +85,236 @@ static VkImageCreateInfo GetSwapchainImpliedImageCreateInfo(VkSwapchainCreateInf
...
+                scaling_caps =
+                    surface_state->GetPresentModeScalingCapabilities(physical_device, present_modes_ci->pPresentModes[i]);
+
+                if ((scaling_caps.supportedPresentScaling != 0) &&
+                    (scaling_caps.supportedPresentScaling & pres_scale_ci->scalingBehavior) == 0) {
+                    if (LogError(device, "VUID-VkSwapchainPresentScalingCreateInfoEXT-scalingBehavior-07771",
+                                 "%s: VkSwapchainPresentModesCreateInfoEXT and VkSwapchainPresentScalingCreateInfoEXT are "
+                                 "present in the pNext chain, "
+                                 "but scalingBehavior (%s) is not a valid present scaling benavior as returned in "
+                                 "VkSurfacePresentScalingCapabilitiesEXT::supportedPresentScaling for pPresentModes[%d]: (%s).",
```
juan-lunarg, line 280: Use PRIu32

layers/wsi_validation.cpp
```diff
@@ -85,6 +85,236 @@ static VkImageCreateInfo GetSwapchainImpliedImageCreateInfo(VkSwapchainCreateInf
...
+                }
+
+                if ((scaling_caps.supportedPresentGravityX != 0) &&
+                    (scaling_caps.supportedPresentGravityX & pres_scale_ci->presentGravityX) == 0) {
+                    if (LogError(device, "VUID-VkSwapchainPresentScalingCreateInfoEXT-presentGravityX-07773",
+                                 "%s: VkSwapchainPresentModesCreateInfoEXT and VkSwapchainPresentScalingCreateInfoEXT are present "
+                                 "in the pNext chain, "
+                                 "but presentGravityX (%s) is not a valid x-axis present gravity as returned in "
+                                 "VkSurfacePresentScalingCapabilitiesEXT::supportedPresentGravityX for pPresentModes[%d]: (%s).",
+                                 func_name, string_VkPresentGravityFlagsEXT(pres_scale_ci->presentGravityX).c_str(), i,
```
juan-lunarg, line 294: Use PRIu32

layers/wsi_validation.cpp
```diff
@@ -85,6 +85,236 @@ static VkImageCreateInfo GetSwapchainImpliedImageCreateInfo(VkSwapchainCreateInf
...
+                    }
+                }
+
+                if ((scaling_caps.supportedPresentGravityY != 0) &&
+                    (scaling_caps.supportedPresentGravityY & pres_scale_ci->presentGravityY) == 0) {
+                    if (LogError(
+                            device, "VUID-VkSwapchainPresentScalingCreateInfoEXT-presentGravityY-07775",
+                            "%s: VkSwapchainPresentModesCreateInfoEXT and VkSwapchainPresentScalingCreateInfoEXT are present "
+                            "in the pNext chain, but presentGravityY (%s) is not a valid y-axis present gravity as returned in "
+                            "VkSurfacePresentScalingCapabilitiesEXT::supportedPresentGravityY for pPresentModes[%d]: (%s).",
```
juan-lunarg, line 306: Use PRIu32

layers/wsi_validation.cpp
```diff
@@ -85,6 +85,236 @@ static VkImageCreateInfo GetSwapchainImpliedImageCreateInfo(VkSwapchainCreateInf
...
+            if (LogError(device, "VUID-VkSwapchainPresentScalingCreateInfoEXT-presentGravityY-07769",
+                         "%s: VkSwapchainPresentScalingCreateInfoEXT is present in the pNext chain and presentGravityY (%d) "
+                         "must not have more than one bit set.",
+                         func_name, pres_scale_ci->presentGravityY)) {
+                skip |= true;
+            }
+        }
+
+        VkSurfacePresentScalingCapabilitiesEXT scaling_caps =
+            surface_state->GetPresentModeScalingCapabilities(physical_device, present_mode);
```
juan-lunarg, line 214: surface_state could potentially be nullptr

layers/wsi_validation.cpp
```diff
@@ -85,6 +85,236 @@ static VkImageCreateInfo GetSwapchainImpliedImageCreateInfo(VkSwapchainCreateInf
...
+            }
+        }
+
+        // Further validation for when a VkSwapchainPresentModesCreateInfoEXT struct is *also* in the pNext chain
+        const VkSwapchainPresentModesCreateInfoEXT *present_modes_ci =
+            LvlFindInChain<VkSwapchainPresentModesCreateInfoEXT>(create_info->pNext);
+        if (present_modes_ci) {
+            for (uint32_t i = 0; i < present_modes_ci->presentModeCount; i++) {
+                scaling_caps =
+                    surface_state->GetPresentModeScalingCapabilities(physical_device, present_modes_ci->pPresentModes[i]);
```
juan-lunarg, line 272: surface_state could potentially be nullptr

layers/wsi_validation.cpp
```diff
@@ -735,11 +952,102 @@ bool CoreChecks::PreCallValidateQueuePresentKHR(VkQueue queue, const VkPresentIn
...
         }
+
+        if (IsExtEnabled(device_extensions.vk_ext_swapchain_maintenance1)) {
+            const auto *swapchain_present_fence_info = LvlFindInChain<VkSwapchainPresentFenceInfoEXT>(pPresentInfo->pNext);
+            if (swapchain_present_fence_info) {
+                if (pPresentInfo->swapchainCount != swapchain_present_fence_info->swapchainCount) {
+                    skip |= LogError(
+                        pPresentInfo->pSwapchains[0], "VUID-VkSwapchainPresentFenceInfoEXT-swapchainCount-07757",
+                        "vkQueuePresentKHR(): VkSwapchainPresentFenceInfoEXT.swapchainCount is %i but pPresentInfo->swapchainCount "
+                        "is %i. For VkSwapchainPresentFenceInfoEXT in pNext chain of VkPresentInfoKHR, "
```
juan-lunarg, line 963: Use PRIU32

layers/wsi_validation.cpp
```diff
@@ -735,11 +952,102 @@ bool CoreChecks::PreCallValidateQueuePresentKHR(VkQueue queue, const VkPresentIn
...
+                }
+            }
+
+            const auto *swapchain_present_mode_info = LvlFindInChain<VkSwapchainPresentModeInfoEXT>(pPresentInfo->pNext);
+            if (swapchain_present_mode_info) {
+                if (pPresentInfo->swapchainCount != swapchain_present_mode_info->swapchainCount) {
+                    skip |= LogError(
+                        pPresentInfo->pSwapchains[0], "VUID-VkSwapchainPresentModeInfoEXT-swapchainCount-07760",
+                        "vkQueuePresentKHR(): VkSwapchainPresentModeInfoEXT.swapchainCount is %i but pPresentInfo->swapchainCount "
+                        "is %i. For VkSwapchainPresentModeInfoEXT in pNext chain of VkPresentInfoKHR, "
```
juan-lunarg, line 983: Use PRIU32

layers/wsi_validation.cpp
```diff
@@ -735,11 +952,102 @@ bool CoreChecks::PreCallValidateQueuePresentKHR(VkQueue queue, const VkPresentIn
...
+                        "VkSwapchainPresentModeInfoEXT.swapchainCount must equal VkPresentInfoKHR.swapchainCount.",
+                        swapchain_present_mode_info->swapchainCount, pPresentInfo->swapchainCount);
+                }
+
+                for (uint32_t i = 0; i < swapchain_present_mode_info->swapchainCount; i++) {
+                    const VkPresentModeKHR present_mode = swapchain_present_mode_info->pPresentModes[i];
+                    const auto swapchain_state = Get<SWAPCHAIN_NODE>(pPresentInfo->pSwapchains[i]);
+                    if (swapchain_state->present_modes.size()) {
```
juan-lunarg, line 991:
```suggestion
                    if (!swapchain_state->present_modes.empty()) {
```

layers/wsi_validation.cpp
```diff
@@ -735,11 +952,102 @@ bool CoreChecks::PreCallValidateQueuePresentKHR(VkQueue queue, const VkPresentIn
...
+bool CoreChecks::PreCallValidateReleaseSwapchainImagesEXT(VkDevice device,
+                                                          const VkReleaseSwapchainImagesInfoEXT *pReleaseInfo) const {
+    bool skip = false;
+    bool image_in_use = false;
+    auto swapchain_state = Get<SWAPCHAIN_NODE>(pReleaseInfo->swapchain);
+    if (swapchain_state) {
+        for (uint32_t i = 0; i < pReleaseInfo->imageIndexCount; i++) {
+            if (pReleaseInfo->pImageIndices[i] >= swapchain_state->images.size()) {
+                skip |= LogError(pReleaseInfo->swapchain, "VUID-VkReleaseSwapchainImagesInfoEXT-pImageIndices-07785",
+                                 "vkReleaseSwapchainImagesEXT: swapchain image index is too large (%u). There are only %u images "
```
juan-lunarg, line 1027: Use PRIu32

juan-lunarg (changes requested):
Mostly just PRIu32 issues

tests/vklayertests_wsi.cpp
```diff
@@ -2122,6 +2123,664 @@ TEST_F(VkLayerTest, TestSurfaceSupportByPhysicalDevice) {
...
+    swapchain_create_info.preTransform = preTransform;
+    swapchain_create_info.compositeAlpha = m_surface_composite_alpha;
+    swapchain_create_info.presentMode = m_surface_non_shared_present_mode;
+    swapchain_create_info.clipped = VK_FALSE;
+    swapchain_create_info.oldSwapchain = 0;
+
+    // Query present mode data
+    auto vkGetPhysicalDeviceSurfaceCapabilities2KHR = reinterpret_cast<PFN_vkGetPhysicalDeviceSurfaceCapabilities2KHR>(
+        vk::GetInstanceProcAddr(instance(), "vkGetPhysicalDeviceSurfaceCapabilities2KHR"));
+    const std::vector<VkPresentModeKHR> defined_present_modes{
```
juan-lunarg, line 2179: This doesn't need to be a std::vector, use std::array with CTAD

```suggestion
    const std::array defined_present_modes{
```

tests/vklayertests_wsi.cpp
```diff
@@ -2122,6 +2123,664 @@ TEST_F(VkLayerTest, TestSurfaceSupportByPhysicalDevice) {
...
+    present_scaling_info.presentGravityX = VK_PRESENT_SCALING_ONE_TO_ONE_BIT_EXT;
+    present_scaling_info.presentGravityY = VK_PRESENT_SCALING_ONE_TO_ONE_BIT_EXT | VK_PRESENT_SCALING_ASPECT_RATIO_STRETCH_BIT_EXT;
+    vk::CreateSwapchainKHR(device(), &swapchain_create_info, nullptr, &m_swapchain);
+    m_errorMonitor->VerifyFound();
+
+    present_scaling_info.presentGravityX = 0;
+    present_scaling_info.presentGravityY = 0;
+    // Find scaling cap not in scaling_capabilities.supportedPresentScaling and create a swapchain using that
+    if (scaling_capabilities.supportedPresentScaling != 0) {
+        std::vector<VkPresentScalingFlagBitsEXT> scaling_flag_bits = {VK_PRESENT_SCALING_ONE_TO_ONE_BIT_EXT,
```
juan-lunarg, line 2320:
```suggestion
        const std::array scaling_flag_bits = {VK_PRESENT_SCALING_ONE_TO_ONE_BIT_EXT,
```

tests/vklayertests_wsi.cpp
```diff
@@ -2122,6 +2123,664 @@ TEST_F(VkLayerTest, TestSurfaceSupportByPhysicalDevice) {
...
+                m_errorMonitor->SetDesiredFailureMsg(kErrorBit,
+                                                     "VUID-VkSwapchainPresentScalingCreateInfoEXT-scalingBehavior-07770");
+                vk::CreateSwapchainKHR(device(), &swapchain_create_info, nullptr, &m_swapchain);
+                m_errorMonitor->VerifyFound();
+                break;
+            }
+        }
+    }
+
+    std::vector<VkPresentGravityFlagBitsEXT> gravity_flag_bits = {VK_PRESENT_GRAVITY_MIN_BIT_EXT, VK_PRESENT_GRAVITY_MAX_BIT_EXT,
```
juan-lunarg, line 2335:
```suggestion
    const std::array gravity_flag_bits = {VK_PRESENT_GRAVITY_MIN_BIT_EXT, VK_PRESENT_GRAVITY_MAX_BIT_EXT,
```

tests/vklayertests_wsi.cpp
```diff
@@ -2122,6 +2123,664 @@ TEST_F(VkLayerTest, TestSurfaceSupportByPhysicalDevice) {
...
+                m_errorMonitor->SetDesiredFailureMsg(kErrorBit,
+                                                     "VUID-VkSwapchainPresentScalingCreateInfoEXT-scalingBehavior-07771");
+                vk::CreateSwapchainKHR(device(), &swapchain_create_info, nullptr, &m_swapchain);
+                m_errorMonitor->VerifyFound();
+                break;
+            }
+        }
+    }
+
+    std::vector<VkPresentGravityFlagBitsEXT> defined_gravity_flag_bits = {
```
juan-lunarg, line 2386:
```suggestion
    const std::array defined_gravity_flag_bits = {
```

tests/vklayertests_wsi.cpp
```diff
@@ -2122,6 +2123,664 @@ TEST_F(VkLayerTest, TestSurfaceSupportByPhysicalDevice) {
...
+    VkSwapchainPresentFenceInfoEXT fence_info = LvlInitStruct<VkSwapchainPresentFenceInfoEXT>();
+    fence_info.swapchainCount = present.swapchainCount + 1;
+    fence_info.pFences = &present_fence.handle();
+    present.pNext = &fence_info;
+
+    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkSwapchainPresentFenceInfoEXT-swapchainCount-07757");
+    vk::QueuePresentKHR(m_device->m_queue, &present);
+    m_errorMonitor->VerifyFound();
+
+    const std::vector<VkPresentModeKHR> defined_present_modes{
```
juan-lunarg, line 2731:
```suggestion
    const std::array defined_present_modes{
```

juan-lunarg (changes requested):
Testing looks good overall. I'd like to see a follow up PR though that adds positive testing though. As that can help catch bugs in the VVL implementation as well.

layers/parameter_validation_utils.cpp
mark-lunarg: Yeah, I thought that was weird too.  I fixed it while working Spencer's comments, seems OK now.

layers/parameter_validation_utils.cpp
mark-lunarg: Spencer also pointed these out, fixed.

layers/wsi_validation.cpp
mark-lunarg: Done.

tests/vklayertests_wsi.cpp
mark-lunarg: That's a cool trick! Fixed all three of these.

juan-lunarg: Agreed class template argument deduction really helps usability of `std::array` in particular.

layers/wsi_validation.cpp
```diff
@@ -85,6 +85,241 @@ static VkImageCreateInfo GetSwapchainImpliedImageCreateInfo(VkSwapchainCreateInf
...
+                         "%s called with a non-supported presentMode (i.e. %s).", func_name, string_VkPresentModeKHR(present_mode));
+    }
+
+    // Validate VkSwapchainPresentModesCreateInfoEXT data
+    auto swapchain_present_modes_ci = LvlFindInChain<VkSwapchainPresentModesCreateInfoEXT>(create_info->pNext);
+    if (swapchain_present_modes_ci) {
+        assert(surface_state);
+        bool found_swapchain_modes_ci_present_mode = false;
+        const std::vector<VkPresentModeKHR> compatible_present_modes =
+            surface_state->GetCompatibleModes(physical_device, present_mode);
```
juan-lunarg, line 110: Why do we check for surface_state on line 93 but assert on line 107?

juan-lunarg (commented):
Just 1 last question

layers/wsi_validation.cpp
mark-lunarg: In my understanding, if surfaceless_query is enabled then surface_state _can_ be null here.  Below, guarded by the `if (swapchain_present_modes_ci) {` it has to be non-null there.

juan-lunarg: Got it that makes sense. A comment explaining the assert would be helpful.

mark-lunarg: 👍

juan-lunarg (approved):
LGTM

mark-lunarg:
@juan-lunarg, @jeremyg-lunarg, thanks for all your help. When more hardware coverage arrives I'm sure there'll be some deficiencies and I'll be glad to hop back in if it'd be helpful.

PR #5001, opened 2022-12-27
tests: Use new GetProcAddr helper

spencer-lunarg:
Cleans up a lot (not all) places to use the new `GetInstanceProcAddr` and `GetDeviceProcAddr` helper functions


arno-lunarg:
The improvement over those utility functions you suggested when we last talked about this problem of getting function pointers would probably be the best answer to this code flow. I think it was something like:
```
GET_DEVICE_PROC_ADDR(vkCmdSetDiscardRectangleEXT);
```
with the macro handling the get_proc_addr call and variable definition.
And using a macro would also allow to use `ASSERT_TRUE(func_ptr != nullptr)`

spencer-lunarg:
@arno-lunarg yes, I agree, I guess I was thinking to quickly fix some of these to help prevent copy-and-pasting the old style... but probably just better to do it the fully correct way we discussed first instead of 2-stepping it

PR #5002, opened 2022-12-27
layers: Label VUID vkCmdCopyImage 07744

spencer-lunarg:
`VUID-vkCmdCopyImage-srcImage-07744`/`VUID-VkCopyImageInfo2-srcImage-07744` was added to the Vulkan Spec and covers the previous UNASSIGNED VU


arno-lunarg (approved):
LGTM!

PR #5003, opened 2022-12-27
layers: Handle SAMPLED_READ in sync2 validation

HansKristian-Work:
Ran into this when moving over to sync2 in Granite. A COMBINED_IMAGE_SAMPLER was trying to access STORAGE_READ rather than SAMPLED_READ, causing false positive. This fixed the issue for me, but I don't know the sync validation code enough to know if this fix is complete.


jeremyg-lunarg (approved):
LGTM. Thanks for the fix!

PR #5006, opened 2022-12-27
layers: Use ValidationObject pointer instead of auto

juan-lunarg:
The generated file is so large my vscode intellisense gives up and can't deduce the auto. Provide the type for simplicity.

### DIFF
--- a/scripts/layer_chassis_generator.py
+++ b/scripts/layer_chassis_generator.py
@@ -207,6 +207,7 @@ class LayerChassisOutputGenerator(OutputGenerator):
         'vkDestroyDebugUtilsMessengerEXT' : 'layer_destroy_callback(layer_data->report_data, messenger, pAllocator);',
         }
 
+    # Avoid using auto in generated code. Intellisense has been known to have issues with large files.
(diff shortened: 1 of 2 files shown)


scripts/layer_chassis_generator.py
```diff
@@ -207,9 +207,9 @@ class LayerChassisOutputGenerator(OutputGenerator):
         'vkDestroyDebugUtilsMessengerEXT' : 'layer_destroy_callback(layer_data->report_data, messenger, pAllocator);',
         }
 
-    precallvalidate_loop = "for (auto intercept : layer_data->object_dispatch) {"
+    precallvalidate_loop = "for (ValidationObject* intercept : layer_data->object_dispatch) {"
```
spencer-lunarg, line 210: I would just add a comment here and maybe around line 973 that we don't want `auto` here for this reason

spencer-lunarg (approved):
I have hit this same issue in the past too, LGTM!

scripts/layer_chassis_generator.py
jeremyg-lunarg: agree

jeremyg-lunarg (approved):

PR #5009, opened 2022-12-28
tests: Update extension enabling

spencer-lunarg:
remove `m_device_extension_names` to `AddRequiredExtensions`/`AddOptionalExtensions` system for a lot more of tests

### DIFF
--- a/tests/vklayertests_wsi.cpp
+++ b/tests/vklayertests_wsi.cpp
@@ -552,7 +552,6 @@ TEST_F(VkLayerTest, SwapchainAcquireImageNoSync2KHR) {
     TEST_DESCRIPTION("Test vkAcquireNextImage2KHR with VK_NULL_HANDLE semaphore and fence");
     SetTargetApiVersion(VK_API_VERSION_1_1);
 
-    AddOptionalExtensions(VK_KHR_DEVICE_GROUP_CREATION_EXTENSION_NAME);
     AddSurfaceExtension();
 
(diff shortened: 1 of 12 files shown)


tests/vklayertests_wsi.cpp
```diff
@@ -552,11 +552,6 @@ TEST_F(VkLayerTest, SwapchainAcquireImageNoSync2KHR) {
     TEST_DESCRIPTION("Test vkAcquireNextImage2KHR with VK_NULL_HANDLE semaphore and fence");
     SetTargetApiVersion(VK_API_VERSION_1_1);
 
-    bool extension_dependency_satisfied = false;
```
jeremyg-lunarg, line 555: this part (and more) is also in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5005

jeremyg-lunarg (approved):
LGTM

tests/vklayertests_wsi.cpp
spencer-lunarg: Some of these I feel the extensions were being enabled and had no real reason to be enabled and mostly likely just left there from some previous change (or some copy and pasting when the test was made)

jeremyg-lunarg: I don't know if I removed all the unneeded extensions but I did convert them to use the new API. Either way, one of us will end up rebasing.

PR #5010, opened 2022-12-28
build: Fix windows build when the source path includes spaces

jeremyg-lunarg:


tests/layers/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ add_library(VkLayer_device_profile_api MODULE
 target_link_libraries(VkLayer_device_profile_api PRIVATE VkLayer_utils)
 
 if (WIN32)
-    set_target_properties(VkLayer_device_profile_api PROPERTIES LINK_FLAGS "/DEF:${CMAKE_CURRENT_SOURCE_DIR}/VkLayer_device_profile_api.def")
+    set_target_properties(VkLayer_device_profile_api PROPERTIES LINK_FLAGS "/DEF:\"${CMAKE_CURRENT_SOURCE_DIR}/VkLayer_device_profile_api.def\"")
```
juan-lunarg, line 26: Out of curiosity is this still a problem if you use `target_link_options`?

jeremyg-lunarg: It looks like that works, I'll push again.

layers/CMakeLists.txt
```diff
@@ -239,7 +239,7 @@ if (VVL_ENABLE_ASAN)
 endif()
 
 if(WIN32)
-    set_target_properties(VkLayer_khronos_validation PROPERTIES LINK_FLAGS "/DEF:${CMAKE_CURRENT_SOURCE_DIR}/VkLayer_khronos_validation.def")
+    target_link_options(VkLayer_khronos_validation INTERFACE "/DEF:${CMAKE_CURRENT_SOURCE_DIR}/VkLayer_khronos_validation.def")
```
jeremyg-lunarg, line 242: I am just guessing with INTERFACE, should it be PUBLIC or PRIVATE instead?

juan-lunarg: It should be `PRIVATE`. Note, we can't use `target_link_options` yet due to our cmake minimum being `3.10`.

juan-lunarg (approved):

PR #5016, opened 2022-12-29
layers: Clean up vk_layer_logging.h

jeremyg-lunarg:
Add vk_layer_logging.cpp, so that most of the logging functions can be de-inlined. This appears to reduce compiled code size for the validation layer by 11% to 33% (on linux).

Change LogMsgLocked() to LogMsg() and restructure it to avoid duplicated code in class ValidationObject.

Remove use of vasprintf() and use stack buffers and snprintf() instead. vasprintf() calls malloc internally and VVL must call free. When overriding malloc, this gets complicated because libc and VVL can end up using different malloc and free implementations.

### DIFF
--- a/layers/vk_layer_logging.cpp
+++ b/layers/vk_layer_logging.cpp
@@ -1,0 +1,572 @@
+/* Copyright (c) 2015-2023 The Khronos Group Inc.
+ * Copyright (c) 2015-2023 Valve Corporation
+ * Copyright (c) 2015-2023 LunarG, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Author: Courtney Goeltzenleuchter <courtney@LunarG.com>
+ * Author: Tobin Ehlis <tobin@lunarg.com>
+ * Author: Mark Young <marky@lunarg.com>
+ * Author: Dave Houlton <daveh@lunarg.com>
+ *
+ */
+#include "vk_layer_logging.h"
+
+#include <csignal>
+#include <cstring>
+#ifdef VK_USE_PLATFORM_WIN32_KHR
+#include <winsock2.h>
+#include <debugapi.h>
+#endif
+
+#include "vk_enum_string_helper.h"
+#include "vk_safe_struct.h"
+#include "vk_validation_error_messages.h"
+#include "xxhash.h"
+
+VKAPI_ATTR void SetDebugUtilsSeverityFlags(std::vector<VkLayerDbgFunctionState> &callbacks, debug_report_data *debug_data) {
+    // For all callback in list, return their complete set of severities and modes
+    for (const auto &item : callbacks) {
+        if (item.IsUtils()) {
+            debug_data->active_severities |= item.debug_utils_msg_flags;
+            debug_data->active_types |= item.debug_utils_msg_type;
+        } else {
+            VkFlags severities = 0;
+            VkFlags types = 0;
+            DebugReportFlagsToAnnotFlags(item.debug_report_msg_flags, true, &severities, &types);
+            debug_data->active_severities |= severities;
+            debug_data->active_types |= types;
+        }
+    }
+}
+
+VKAPI_ATTR void RemoveDebugUtilsCallback(debug_report_data *debug_data, std::vector<VkLayerDbgFunctionState> &callbacks,
+                                            uint64_t callback) {
+    auto item = callbacks.begin();
+    for (item = callbacks.begin(); item != callbacks.end(); item++) {
+        if (item->IsUtils()) {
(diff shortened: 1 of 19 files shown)


jeremyg-lunarg:
This is needed to allow https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4876 to work on linux. But it is worth doing on its own.

The size of the VVL .so on linux / GCC 12 goes from 412Mb to 276Mb for Debug and 208Mb to 184Mb for Release

mark-lunarg (approved):
This is great!

layers/vk_layer_logging.cpp
```diff
@@ -0,0 +1,559 @@
...
+    if (item != callbacks.end()) {
+        callbacks.erase(item);
+    }
+    SetDebugUtilsSeverityFlags(callbacks, debug_data);
+}
+
+// Deletes all debug callback function structs
+static inline void RemoveAllMessageCallbacks(debug_report_data *debug_data, std::vector<VkLayerDbgFunctionState> &callbacks) {
+    callbacks.clear();
+}
```
juan-lunarg, line 72: This one line function is only used once. Better to just remove IMO. Especially since it only used once, and doesn't even use the first parameter it's passed.

layers/vk_layer_logging.cpp
```diff
@@ -0,0 +1,559 @@
...
+    VkDebugUtilsMessageTypeFlagsEXT type;
+
+    DebugReportFlagsToAnnotFlags(msg_flags, false, &severity, &type);
+    std::unique_lock<std::mutex> lock(debug_data->debug_output_mutex);
+    // Avoid logging cost if msg is to be ignored
+    if (!LogMsgEnabled(debug_data, vuid_text, severity, type)) {
+        return false;
+    }
+
+    char buf[1024];
```
juan-lunarg, line 373: Is 1024 enough? Should we assert if it's not enough?

jeremyg-lunarg: I *think* it is enough, but probably we should assert. There's another buffer in one of the callback functions which is 2048, but that's for this part of the message and the VUID text.

layers/vk_layer_logging.cpp
jeremyg-lunarg: it's actually worse than that, because the call site is immediately before `delete debug_data;` which will clear the callback list.

layers/vk_layer_logging.cpp
jeremyg-lunarg: I forgot to add the assert, but I think I figured out a way to do it without the possibility of overrun failure.

spencer-lunarg (approved):

juan-lunarg (changes requested):
Looks like just 1 test is failing.

```
[  FAILED  ] VkLayerTest.SpecLinks

 1 FAILED TEST
  YOU HAVE 1 DISABLED TEST
```

layers/vk_layer_logging.cpp
```diff
@@ -0,0 +1,567 @@
...
+        result = vsnprintf(str_plus_spec_text.data(), str_plus_spec_text.size(), format, arg_copy);
+        va_end(arg_copy);
+
+        assert(result >= 0);
+        if (result < 0) {
+            str_plus_spec_text = "Message generation failure";
+            break;
+        }
+        // Shrink or grow the string to match the message size
+        str_plus_spec_text.resize(result);
```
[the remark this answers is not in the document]
jeremyg-lunarg: Ugh, it still isn't right. I need the +1 in the overflow case because:
```
The functions snprintf() and vsnprintf() write at most size bytes (including the terminating null byte ('\0')) to str.
```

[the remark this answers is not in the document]
jeremyg-lunarg: BUT:
```
RETURN VALUE
       Upon successful return, these functions return the number of characters printed (excluding the null byte used to end output to strings).
```

tests/vklayertests_others.cpp
```diff
@@ -375,7 +375,10 @@ TEST_F(VkLayerTest, DuplicateMessageLimit) {
     m_errorMonitor->SetDesiredFailureMsg((kErrorBit | kWarningBit), "VUID-VkPhysicalDeviceProperties2-pNext-pNext");
     vkGetPhysicalDeviceProperties2KHR(gpu(), &properties2);
     m_errorMonitor->VerifyFound();
-    m_errorMonitor->SetDesiredFailureMsg((kErrorBit | kWarningBit), "VUID-VkPhysicalDeviceProperties2-pNext-pNext");
+    // VUID-VkPhysicalDeviceProperties2-pNext-pNext produces a massive ~3600 character log message, which hits a
```
ncesario-lunarg, line 378: 😬

ncesario-lunarg (approved):
LGTM once the gremlins are worked out :)

The overflow case is more motivation to refactor the base logging mechanisms, but if that's working again, we'll probably kick that down the road again...

jeremyg-lunarg:
> The overflow case is more motivation to refactor the base logging mechanisms, but if that's working again, we'll probably kick that down the road again...

I agree we should refactor more.

tests/vklayertests_others.cpp
juan-lunarg: I think this PR is fine once the logging code is fixed to address this.

jeremyg-lunarg: Yes it is working now. With earlier versions of the code I was either dropping a character in this log message or putting a '\0' in the middle of the message which would cause the SpecLinks test to fail. By changing the DesiredFailure here, both of those problems are covered by a test case.

juan-lunarg (approved):

PR #5018, opened 2022-12-30
cmake: Build a shared library instead of module

sarnex:
If we use MODULE, cmake doesn't set SONAME which we should set. MODULE is meant only for runtime explicit dynamic linking with dlopen

Signed-off-by: Nick Sarnie <sarnex@gentoo.org>

### DIFF
--- a/layers/CMakeLists.txt
+++ b/layers/CMakeLists.txt
@@ -213,7 +213,7 @@ if (NOT BUILD_LAYERS)
     return()
 endif()
 
-add_library(VkLayer_khronos_validation MODULE)
+add_library(VkLayer_khronos_validation SHARED)
 
 target_sources(VkLayer_khronos_validation PRIVATE
     ${CHASSIS_LIBRARY_FILES}


layers/CMakeLists.txt
```diff
@@ -213,7 +213,7 @@ if (NOT BUILD_LAYERS)
     return()
 endif()
 
-add_library(VkLayer_khronos_validation MODULE)
+add_library(VkLayer_khronos_validation SHARED)
```
juan-lunarg, line 216: 2 main things.

1.) This will cause problems for [Windows](https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4804). By specifying the library as a `MODULE` CMake won't install the pointless `.lib`, which is important for LunarG SDK script generation

2.) "MODULE is meant only for runtime explicit dynamic linking with dlopen". Yes that is correct. That is the way layers are opened by the Vulkan Loader. Users aren't meant to link against the loader.

juan-lunarg (changes requested):
-1

juan-lunarg:
> If we use MODULE, cmake doesn't set SONAME which we should set

What is the consequence? Because testing didn't catch anything.

sarnex:
Sorry, I didn't look deep enough into the vulkan loader usage to find the dlopen

layers/CMakeLists.txt
sarnex: > > If we use MODULE, cmake doesn't set SONAME which we should set
> 
> What is the consequence? Testing didn't catch anything.

I found this from some automated QA testing in Gentoo

juan-lunarg:
> Sorry, I didn't look deep enough into the vulkan loader usage to find the dlopen

No worries. Out of curiosity what does the automated QA testing look for?

sarnex:
> > Sorry, I didn't look deep enough into the vulkan loader usage to find the dlopen
> 
> No worries. Out of curiosity what does Gentoo's automated QA testing look for?

We look for SONAME in all shared libraries, the implementation of that check is here

https://gitweb.gentoo.org/proj/portage.git/tree/bin/install-qa-check.d/80libraries

juan-lunarg:
> We look for SONAME in all shared libraries, the implementation of that check is here
> 
> https://gitweb.gentoo.org/proj/portage.git/tree/bin/install-qa-check.d/80libraries

Very cool to see this proactive approach to fixing libraries upstream :+1:

PR #5019, opened 2022-12-30
layers: Remove extra SHADER_MODULE_STATE constructor

spencer-lunarg:
Reduce the number of `SHADER_MODULE_STATE` constructors to only what is needed

### DIFF
--- a/layers/gpu_validation/debug_printf.cpp
+++ b/layers/gpu_validation/debug_printf.cpp
@@ -238,7 +238,7 @@ std::vector<DPFSubstring> DebugPrintf::ParseFormatString(const std::string &form
     return parsed_strings;
 }
 
-std::string DebugPrintf::FindFormatString(std::vector<uint32_t> pgm, uint32_t string_id) {
+std::string DebugPrintf::FindFormatString(layer_data::span<const uint32_t> pgm, uint32_t string_id) {
     std::string format_string;
     SHADER_MODULE_STATE module_state(pgm);
     if (module_state.words_.empty()) {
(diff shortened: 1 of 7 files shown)


layers/debug_printf.cpp
```diff
@@ -240,7 +240,7 @@ std::vector<DPFSubstring> DebugPrintf::ParseFormatString(const std::string &form
 
 std::string DebugPrintf::FindFormatString(std::vector<uint32_t> pgm, uint32_t string_id) {
     std::string format_string;
-    SHADER_MODULE_STATE module_state(pgm);
+    SHADER_MODULE_STATE module_state(pgm.data(), pgm.size() * sizeof(uint32_t));
```
juan-lunarg, line 243: The constructor would be more robust if it took a `layer_data::span` instead of an individual pointer / size.

jeremyg-lunarg (dismissed):
LGTM

ncesario-lunarg (dismissed):

layers/gpu_validation/gpu_validation.cpp
```diff
@@ -1149,7 +1149,7 @@ void GpuAssisted::AnalyzeAndGenerateMessages(VkCommandBuffer command_buffer, VkQ
     std::string vuid_msg;
     VkShaderModule shader_module_handle = VK_NULL_HANDLE;
     VkPipeline pipeline_handle = VK_NULL_HANDLE;
-    std::vector<uint32_t> pgm;
+    layer_data::span<const uint32_t> pgm;
```
juan-lunarg, line 1152: Avoids copy on line 1161

layers/gpu_validation/debug_printf.cpp
```diff
@@ -313,7 +313,7 @@ void DebugPrintf::AnalyzeAndGenerateMessages(VkCommandBuffer command_buffer, VkQ
         std::stringstream shader_message;
         VkShaderModule shader_module_handle = VK_NULL_HANDLE;
         VkPipeline pipeline_handle = VK_NULL_HANDLE;
-        std::vector<uint32_t> pgm;
+        layer_data::span<uint32_t> pgm;
```
juan-lunarg, line 316: avoids copy of line 325

layers/debug_printf.cpp
juan-lunarg: Went ahead and update this PR to use `layer_data::span` which led me to find std::vector copies that could be eliminated with span usage

juan-lunarg (approved):

PR #5020, opened 2022-12-30
layers: Add GetBitWidth SPIR-V helper

spencer-lunarg:
- Move the `GetBitWidth` logic into a per-`Instruction` function to be reused
- cleaned up `GetNumComponentsInBaseType` and `GetTypeBitsSize` a bit to be consistent with other similar helper functions

### DIFF
--- a/layers/shader_instruction.cpp
+++ b/layers/shader_instruction.cpp
@@ -76,6 +76,31 @@ uint32_t Instruction::GetConstantValue() const {
     return Word(3);
 }
 
+// The idea of this function is to not have to constantly lookup which operand for the width
+// inst.Word(2) -> inst.GetBitWidth()
+uint32_t Instruction::GetBitWidth() const {
+    const uint32_t opcode = Opcode();
+    uint32_t bit_width = 0;
+    switch (opcode) {
+        case spv::Op::OpTypeFloat:
+        case spv::Op::OpTypeInt:
+            bit_width = Word(2);
+            break;
+        case spv::Op::OpTypeBool:
+            // The Spec states:
+            // "Boolean values considered as 32-bit integer values for the purpose of this calculation"
+            // when getting the size for the limits
+            bit_width = 32;
+            break;
+        default:
+            // Most likely the caller is not checking for this being a matrix/array/struct/etc
+            // This class only knows a single instruction's information
+            assert(0);
+            break;
+    }
+    return bit_width;
+}
+
 AtomicInstructionInfo Instruction::GetAtomicInfo(const SHADER_MODULE_STATE& module_state) const {
     AtomicInstructionInfo info;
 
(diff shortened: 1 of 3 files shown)


layers/shader_instruction.cpp
```diff
@@ -76,6 +76,23 @@ uint32_t Instruction::GetConstantValue() const {
     return Word(3);
 }
 
+uint32_t Instruction::GetBitWidth() const {
+    const uint32_t opcode = Opcode();
+    uint32_t bit_width = 0;
+    if (opcode == spv::Op::OpTypeFloat || opcode == spv::Op::OpTypeInt) {
```
juan-lunarg, line 82: Would a switch statement be more appropriate?

layers/shader_instruction.cpp
```diff
@@ -76,6 +76,23 @@ uint32_t Instruction::GetConstantValue() const {
...
+        bit_width = Word(2);
+    } else if (opcode == spv::Op::OpTypeBool) {
+        // The Spec states:
+        // "Boolean values considered as 32-bit integer values for the purpose of this calculation"
+        // when getting the size for the limits
+        bit_width = 32;
+    } else {
+        // Most likely the caller is not checking for this being a matrix/array/struct/etc
+        assert(0);
```
juan-lunarg, line 91: Would it be better to fill this out now? Since this is currently another TODO?

layers/shader_instruction.cpp
```diff
@@ -76,6 +76,23 @@ uint32_t Instruction::GetConstantValue() const {
     return Word(3);
 }
 
+uint32_t Instruction::GetBitWidth() const {
```
juan-lunarg, line 79: Do the Spriv-Headers not provide helper functionality?

layers/shader_instruction.cpp
spencer-lunarg: fill out with what? The `Instruction` class only knows what the instruction itself, so it should never be a Array/Matrix here

spencer-lunarg: basically this is not a TODO, but a helper message in the future if someone uses this function incorrectly

layers/shader_instruction.cpp
spencer-lunarg: kinda, it provides where the `Width` is for Floats/Ints are, but we also need to manually add in `Bool`

I thought about generating this, but I can't imagine what other "base type" would ever be added that also provide a bit width

The idea of this function is to not have to constantly lookup which operand in things like `OpTypeFloat` is needed to get the width

`inst.Word(2)` vs `inst.GetBitWidth()`

layers/shader_instruction.cpp
juan-lunarg: Gotcha that makes sense.

layers/shader_instruction.cpp
spencer-lunarg: Added some better comments as well to this function

juan-lunarg (approved):
LGTM

PR #5021, opened 2022-12-30
layers: Remove old spirv_inst_iter system

spencer-lunarg:
`spirv_inst_iter` was only used in GPU AV and DebugPrintf, this removes it in favor for the new dedicated `Instruction` class

### DIFF
--- a/layers/debug_printf.cpp
+++ b/layers/debug_printf.cpp
@@ -241,14 +241,17 @@ std::vector<DPFSubstring> DebugPrintf::ParseFormatString(const std::string &form
 std::string DebugPrintf::FindFormatString(std::vector<uint32_t> pgm, uint32_t string_id) {
     std::string format_string;
     SHADER_MODULE_STATE module_state(pgm);
-    if (module_state.words_.size() > 0) {
-        for (const auto &insn : module_state) {
-            if (insn.opcode() == spv::OpString) {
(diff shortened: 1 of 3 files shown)


layers/debug_printf.cpp
```diff
@@ -242,11 +242,13 @@ std::string DebugPrintf::FindFormatString(std::vector<uint32_t> pgm, uint32_t st
     std::string format_string;
     SHADER_MODULE_STATE module_state(pgm);
     if (module_state.words_.size() > 0) {
```
juan-lunarg, line 244: Is the size check needed? It seems strange to check the size before iterating through the loop, since normally the loop logic would handle that.

If it is needed return early.

```suggestion
    if (module_state.words_.empty()) { return {}; }
```

layers/gpu_utils.cpp
```diff
@@ -1162,11 +1169,11 @@ void UtilGenerateSourceMessages(const std::vector<uint32_t> &pgm, const uint32_t
     uint32_t reported_line_number = 0;
     uint32_t reported_column_number = 0;
     if (module_state.words_.size() > 0) {
```
juan-lunarg, line 1171: ditto to this size check

juan-lunarg (changes requested):

juan-lunarg (approved):
LGTM

PR #5024, opened 2022-12-30
layers: Mark ValidationObject::ReadLock method as const 

juan-lunarg:


scripts/layer_chassis_generator.py
```diff
@@ -393,8 +393,8 @@ class ValidationObject {
 
         void InitObjectDispatchVectors();
 
-        std::shared_mutex validation_object_mutex;
-        virtual ReadLockGuard ReadLock() {
+        mutable std::shared_mutex validation_object_mutex;
```
juan-lunarg, line 396: Currently this is the crux. `validation_object_mutex` wasn't marked as `mutable`.

Marking a mutex as mutable is perfectly acceptable:
https://stackoverflow.com/a/25521702

And we already do this in other places in the codebase like `CMD_BUFFER_STATE`

layers/generated/chassis.cpp
```diff
@@ -318,9 +318,9 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateInstance(const VkInstanceCreateInfo *pCreat
 
     // Init dispatch array and call registration functions
     bool skip = false;
-    for (ValidationObject* intercept : local_object_dispatch) {
+    for (const ValidationObject* intercept : local_object_dispatch) {
```
juan-lunarg, line 321: This is essentially the main objective of this PR. Allowing the PreCallValidate loop to use a const pointer and removing the const_cast from being necessary.

spencer-lunarg (approved):

PR #5030, opened 2023-01-02
scripts: Run code format checks commit-by-commit

jeremyg-lunarg:
Run each part of the code format check on each commit in a PR so that we ensure that each commit is correct rather than the whole set ends up correct even if there are errors in the middle.

### DIFF
--- a/scripts/check_code_format.py
+++ b/scripts/check_code_format.py
@@ -59,8 +59,8 @@ def CPrint(msg_type, msg_string):
 #
 #
 # Check clang-formatting of source code diff
-def VerifyClangFormatSource(target_refspec, target_files):
-    CPrint('', "\nChecking PR source code for clang-format errors:")
+def VerifyClangFormatSource(commit, target_files):
+    target_refspec = f'{commit}^...{commit}'
     retval = 0
     good_file_pattern = re.compile('.*\\.(cpp|cc|c\+\+|cxx|c|h|hpp)$')
     diff_files_list = [item for item in target_files if good_file_pattern.search(item)]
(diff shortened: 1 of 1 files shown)


scripts/check_code_format.py
```diff
@@ -265,6 +250,10 @@ def main():
         print("This script requires Python 3. Run script with [-h] option for more details.")
         exit()
 
+    if os.path.isfile('check_code_format.py'):
+        os.chdir('..')
+
+
```
juan-lunarg, line 256:
```suggestion
```

mark-lunarg (approved):

scripts/check_code_format.py
```diff
@@ -282,29 +271,48 @@ def main():
...
-    #
-    #
-    # Get list of files involved in this branch
-    target_files_data = subprocess.check_output(['git', 'diff', '--name-only', diff_range])
-    target_files = target_files_data.decode('utf-8')
-    target_files = target_files.split("\n")
+    clang_format_failure = 0
+    copyright_failure = 0
+    commit_msg_failure = 0
+    stype_assign_failure = 0
```
juan-lunarg, line 278: What's the rational for having separate variables instead of 1 failure variable?

scripts/check_code_format.py
```diff
@@ -282,29 +271,48 @@ def main():
...
-    stype_assign_failure = VerifyTypeAssign(diff_range, target_files)
+    subprocess.run(['git', 'checkout', '-q', orig_branch])
 
     if clang_format_failure or copyright_failure or commit_msg_failure or stype_assign_failure:
-        CPrint('ERR_MSG', "\nOne or more format checks failed.\n\n")
+        CPrint('ERR_MSG', "One or more format checks failed.\n")
         exit(1)
     else:
-        CPrint('SUCCESS_MSG', "\nAll format checks passed.\n\n")
+        CPrint('SUCCESS_MSG', "All format checks passed.\n")
```
juan-lunarg, line 315: The `else:` and the `exit(0)` aren't needed.

scripts/check_code_format.py
```diff
@@ -282,29 +271,48 @@ def main():
...
 
-    if os.path.isfile('check_code_format.py'):
-        os.chdir('..')
+    orig_branch = check_output(['git', 'rev-parse', '--abbrev-ref', 'HEAD']).decode('utf-8').splitlines()[0]
+    if orig_branch == 'HEAD':
+        orig_branch = check_output(['git', 'rev-parse', 'HEAD']).decode('utf-8').splitlines()[0]
+
+    commits = check_output(['git', 'log', '--format=%h', f'{base_refspec}...{target_refspec}']).split(b'\n')
+    commits.reverse()
+    for c in commits:
```
juan-lunarg, line 286:
```suggestion
    # Run code format check on each commit in a PR so that we ensure that each commit is correct.
    for c in commits:
```

juan-lunarg (approved):
Minor nitpicks but it LGTM

scripts/check_code_format.py
jeremyg-lunarg: good question!

Tony-LunarG (approved):
LGTM

ncesario-lunarg (approved):

juan-lunarg (approved):
:+1:

PR #5032, opened 2023-01-03
cmake: Raise minimum to 3.17.2

juan-lunarg:
spirv-tools updated their cmake minimum to 3.17.2

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -12,7 +12,7 @@
 
 ## Requirements
 
-1. CMake >= 3.10.2
+1. CMake >= 3.17.2
 2. C++ >= c++17 compiler. See platform-specific sections below for supported compiler versions.
 3. Python >= 3.7 (3.6 may work, 3.5 and earlier is not supported)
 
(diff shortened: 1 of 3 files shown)


ncesario-lunarg (approved):
LGTM!

SpaceIm:
I don't understand the rational. Why updating min CMake version based on min version of external dependencies? It's orthogonal. CMake min version should be based on CMake policies you want and features used in CMakeLists.

juan-lunarg:
> I don't understand the rational. Why updating min CMake version based on min version of external dependencies (and spirv-tools rational is also very strange)? It's orthogonal. CMake min version should be based on CMake policies you want and features used in CMakeLists.

- It would break one of our CI processes for github actions, since we test against the cmake minimum we support
- Ensures consistency across the Vulkan SDK which is much easier for releases
- Less context switching for developers who often work on more than 1 repo
- We have dramatically cleaned up our CMake code after raising our minimum to 3.17
- etc.

The stated rational in this PR is more like the smoking gun instead of the complete rational for the change.

PR #5036, opened 2023-01-04
Using InterfaceVariable in EntryPoint 

spencer-lunarg:
Currently the shader validation code gets the same information from many different ways resulting in a tangled mess.

This attempts to move more information over to `InterfaceVariable` and then have it at a per-EntryPoint level (which will be needed to support multiple entry points correctly)

This is just a first step PR, the biggest issue is an "Interface Variable" being used for a descriptor is not being treated the same as the Input/Output shader stage interface variables, but yet are sharing some of the same logic. The Input/Output interface is a slight more tricky due to the fact it can have `OpMemberDecorate`

In the process of trying to refactor this, had some generic cleanup as well

### DIFF
--- a/layers/best_practices_utils.cpp
+++ b/layers/best_practices_utils.cpp
@@ -1371,31 +1371,32 @@ bool BestPractices::ValidateCreateComputePipelineArm(const VkComputePipelineCrea
     }
 
     auto interface_variables = module_state->GetInterfaceVariable(entrypoint);
+    if (interface_variables) {
+        unsigned dimensions = 0;
+        if (x > 1) dimensions++;
+        if (y > 1) dimensions++;
+        if (z > 1) dimensions++;
+        // Here the dimension will really depend on the dispatch grid, but assume it's 1D.
+        dimensions = std::max(dimensions, 1u);
 
-    unsigned dimensions = 0;
-    if (x > 1) dimensions++;
-    if (y > 1) dimensions++;
-    if (z > 1) dimensions++;
-    // Here the dimension will really depend on the dispatch grid, but assume it's 1D.
-    dimensions = std::max(dimensions, 1u);
+        // If we're accessing images, we almost certainly want to have a 2D workgroup for cache reasons.
+        // There are some false positives here. We could simply have a shader that does this within a 1D grid,
+        // or we may have a linearly tiled image, but these cases are quite unlikely in practice.
+        bool accesses_2d = false;
+        for (const InterfaceVariable& variable : *interface_variables) {
+            auto dim = module_state->GetShaderResourceDimensionality(variable);
+            if (dim < 0) continue;
+            auto spvdim = spv::Dim(dim);
+            if (spvdim != spv::Dim1D && spvdim != spv::DimBuffer) accesses_2d = true;
+        }
 
-    // If we're accessing images, we almost certainly want to have a 2D workgroup for cache reasons.
-    // There are some false positives here. We could simply have a shader that does this within a 1D grid,
-    // or we may have a linearly tiled image, but these cases are quite unlikely in practice.
-    bool accesses_2d = false;
-    for (const InterfaceVariable& variable : *interface_variables) {
-        auto dim = module_state->GetShaderResourceDimensionality(variable);
-        if (dim < 0) continue;
-        auto spvdim = spv::Dim(dim);
-        if (spvdim != spv::Dim1D && spvdim != spv::DimBuffer) accesses_2d = true;
-    }
-
-    if (accesses_2d && dimensions < 2) {
-        LogPerformanceWarning(device, kVUID_BestPractices_CreateComputePipelines_ComputeSpatialLocality,
-                              "%s vkCreateComputePipelines(): compute shader has work group dimensions (%u, %u, %u), which "
-                              "suggests a 1D dispatch, but the shader is accessing 2D or 3D images. The shader may be "
-                              "exhibiting poor spatial locality with respect to one or more shader resources.",
-                              VendorSpecificTag(kBPVendorArm), x, y, z);
+        if (accesses_2d && dimensions < 2) {
+            LogPerformanceWarning(device, kVUID_BestPractices_CreateComputePipelines_ComputeSpatialLocality,
+                                  "%s vkCreateComputePipelines(): compute shader has work group dimensions (%u, %u, %u), which "
+                                  "suggests a 1D dispatch, but the shader is accessing 2D or 3D images. The shader may be "
+                                  "exhibiting poor spatial locality with respect to one or more shader resources.",
+                                  VendorSpecificTag(kBPVendorArm), x, y, z);
+        }
     }
 
     return skip;
--- a/layers/best_practices_utils.cpp
+++ b/layers/best_practices_utils.cpp
@@ -1,6 +1,6 @@
-/* Copyright (c) 2015-2022 The Khronos Group Inc.
- * Copyright (c) 2015-2022 Valve Corporation
- * Copyright (c) 2015-2022 LunarG, Inc.
+/* Copyright (c) 2015-2023 The Khronos Group Inc.
+ * Copyright (c) 2015-2023 Valve Corporation
+ * Copyright (c) 2015-2023 LunarG, Inc.
  * Modifications Copyright (C) 2020 Advanced Micro Devices, Inc. All rights reserved.
  * Modifications Copyright (C) 2022 RasterGrid Kft.
  *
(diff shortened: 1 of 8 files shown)


ncesario-lunarg (approved):
LGTM once internal CI passes!

layers/best_practices_utils.cpp
```diff
@@ -1370,7 +1370,7 @@ bool BestPractices::ValidateCreateComputePipelineArm(const VkComputePipelineCrea
                                       kThreadGroupDispatchCountAlignmentArm);
     }
 
-    auto descriptor_uses = module_state->CollectInterfaceByDescriptorSlot(entrypoint_optional);
+    auto interface_variables = module_state->GetInterfaceVariable(entrypoint);
```
juan-lunarg, line 1373: Current code doesn't check if `module_state->GetInterfaceVariable(` returns a nullptr.

Also IMO the type returned by `GetInterfaceVariable` should be specified.

```suggestion
    const std::vector<InterfaceVariable>* interface_variables = module_state->GetInterfaceVariable(entrypoint);
    if (interface_variables) {
        ... // check for kVUID_BestPractices_CreateComputePipelines_ComputeSpatialLocality
    }
```

layers/pipeline_state.cpp
```diff
@@ -32,24 +32,14 @@
...
-
-static bool HasAtomicDescriptor(const std::vector<PipelineStageState::DescriptorUse> &descriptor_uses) {
-    return std::any_of(descriptor_uses.begin(), descriptor_uses.end(),
-                       [](const PipelineStageState::DescriptorUse &use) { return use.second.is_atomic_operation; });
-}
-
-static bool WrotePrimitiveShadingRate(VkShaderStageFlagBits stage_flag, std::optional<Instruction> entrypoint,
+static bool WrotePrimitiveShadingRate(VkShaderStageFlagBits stage_flag, const Instruction &entrypoint,
                                       const SHADER_MODULE_STATE *module_state) {
     bool primitiverate_written = false;
```
juan-lunarg, line 37: nit

```suggestion
    bool primitive_rate_written = false;
```

layers/pipeline_state.cpp
```diff
@@ -65,12 +55,20 @@ PipelineStageState::PipelineStageState(const safe_VkPipelineShaderStageCreateInf
...
       stage_flag(stage->stage),
       entrypoint(module_state->FindEntrypoint(stage->pName, stage->stage)),
-      descriptor_uses(module_state->CollectInterfaceByDescriptorSlot(entrypoint)),
-      has_writable_descriptor(HasWriteableDescriptor(descriptor_uses)),
-      has_atomic_descriptor(HasAtomicDescriptor(descriptor_uses)),
-      wrote_primitive_shading_rate(WrotePrimitiveShadingRate(stage_flag, entrypoint, module_state.get())),
       writes_to_gl_layer(module_state->WritesToGlLayer()),
-      has_input_attachment_capability(module_state->HasInputAttachmentCapability()) {}
+      has_input_attachment_capability(module_state->HasInputAttachmentCapability()) {
+    if (entrypoint) {
```
juan-lunarg, line 60:
```suggestion
    if (!entrypoint) {
        return;
    }
```

layers/pipeline_state.cpp
```diff
@@ -156,38 +154,38 @@ PIPELINE_STATE::StageStateVec PIPELINE_STATE::GetStageStates(const ValidationSta
 PIPELINE_STATE::ActiveSlotMap PIPELINE_STATE::GetActiveSlots(const StageStateVec &stage_states) {
     PIPELINE_STATE::ActiveSlotMap active_slots;
     for (const auto &stage : stage_states) {
-        if (!stage.entrypoint) {
+        if (!stage.entrypoint || !stage.descriptor_variables) {
             continue;
         }
         // Capture descriptor uses for the pipeline
-        for (const auto &use : stage.descriptor_uses) {
+        for (const InterfaceVariable &variable : *stage.descriptor_variables) {
```
juan-lunarg, line 161:
```suggestion
        for (const InterfaceVariable &descriptor : *stage.descriptor_variables) {
```

layers/pipeline_state.cpp
```diff
@@ -156,38 +154,38 @@ PIPELINE_STATE::StageStateVec PIPELINE_STATE::GetStageStates(const ValidationSta
...
+            reqs |= stage.module_state->DescriptorTypeToReqs(variable.type_id);
+            if (variable.is_atomic_operation) reqs |= DESCRIPTOR_REQ_VIEW_ATOMIC_OPERATION;
+            if (variable.is_sampler_sampled) reqs |= DESCRIPTOR_REQ_SAMPLER_SAMPLED;
+            if (variable.is_sampler_implicitLod_dref_proj) reqs |= DESCRIPTOR_REQ_SAMPLER_IMPLICITLOD_DREF_PROJ;
+            if (variable.is_sampler_bias_offset) reqs |= DESCRIPTOR_REQ_SAMPLER_BIAS_OFFSET;
+            if (variable.is_read_without_format) reqs |= DESCRIPTOR_REQ_IMAGE_READ_WITHOUT_FORMAT;
+            if (variable.is_write_without_format) reqs |= DESCRIPTOR_REQ_IMAGE_WRITE_WITHOUT_FORMAT;
+            if (variable.is_dref_operation) reqs |= DESCRIPTOR_REQ_IMAGE_DREF;
+
+            if (variable.samplers_used_by_image.size()) {
```
juan-lunarg, line 176:
```suggestion
            if (!variable.samplers_used_by_image.empty()) {
```

layers/shader_module.cpp
```diff
@@ -1333,7 +1364,8 @@ void SHADER_MODULE_STATE::FindVariableDescriptorType(bool is_storage_buffer, Int
 
         case spv::OpTypeStruct: {
             layer_data::unordered_set<uint32_t> nonwritable_members;
-            if (GetDecorationSet(type->Word(1)).flags & DecorationSet::buffer_block_bit) {
+            bool is_storage_buffer = interface_var.storage_class == spv::StorageClassStorageBuffer;
```
juan-lunarg, line 1367:
```suggestion
            const bool is_storage_buffer = interface_var.storage_class == spv::StorageClassStorageBuffer;
```

juan-lunarg (changes requested):

layers/best_practices_utils.cpp
ncesario-lunarg: We should probably have this discussion in the next team sync. After using `auto` _liberally_ for the past 6+ years, I haven't found a benefit to being verbose with types in cases like this, but that is my subjective opinion. If we want to reduce or remove the use of `auto`, that's fine, but we'll need to come to a consensus, and then going forward we can make that official. For now, this lines up with current usage of `auto` elsewhere in the codebase, so I don't see a need for that to change here.

juan-lunarg: I made a case for not using auto here because to me it might have made it more clear to check for `nullptr`. My only indication that this was a pointer was it being dereferenced in the for loop. In general I don't mind `auto` (assuming my intellisense works).

juan-lunarg: Updated suggestion.

```suggestion
    auto interface_variables = module_state->GetInterfaceVariable(entrypoint);
    if (interface_variables) {
        ... // check for kVUID_BestPractices_CreateComputePipelines_ComputeSpatialLocality
    }
```

ncesario-lunarg: That's a good point. We've been putting off the "auto discussion" for a long time, and I'd like to get it nailed down on what the team would like to see going forward.

In this particular instance I personally don't see the benefit to even adding `*` as the type could change to `std::shared_ptr`, `std::optional`, or something else with `operator bool` and it does not change how the code below is interpreted by a human. Again, though, this is my lazy and subjective point of view, and I assume I'm in the minority here :)

layers/pipeline_state.cpp
spencer-lunarg: Thought about this, but in future someone might want to add logic to the constructor non related to `entrypoint` and this clearly groups all the member variables effected by `entrypoint`

layers/pipeline_state.cpp
spencer-lunarg: `descriptor` is a misleading variable name here. I made an effort to use `variable` consistently everywhere because `InterfaceVariable` is describing the `OpVariable`

layers/best_practices_utils.cpp
spencer-lunarg: applied change, was going to add my two cents about `auto` but rather defer that to a proper discussion time/place

layers/pipeline_state.cpp
juan-lunarg: Makes sense :+1:

juan-lunarg (approved):
:+1:

PR #5038, opened 2023-01-04
layers: Make RENDER_PASS_STATE const ref

spencer-lunarg:
There was no reason not to use `RENDER_PASS_STATE` as a const ref as we are dereferencing without checking already


jeremyg-lunarg (approved):
LGTM

ncesario-lunarg (approved):
LGTM!

juan-lunarg (approved):

Issue #5035, opened 2023-01-04
Issue in validation of trace rays cmds


arno-lunarg:
**Describe the Issue**
Shader binding tables may be specified as unused if the `deviceAddress` member is 0. Currently, validation does not account for this.

**Valid Usage ID**
Those VUIDs need to be fixed:
`VUID-vkCmdTraceRaysKHR-pRayGenShaderBindingTable-03680`
`VUID-vkCmdTraceRaysKHR-pMissShaderBindingTable-03683`
`VUID-vkCmdTraceRaysKHR-pHitShaderBindingTable-03687`
`VUID-vkCmdTraceRaysKHR-pCallableShaderBindingTable-03691`

As well as the `vkCmdTraceRaysIndirectKHR` equivalents.

PR #5039, opened 2023-01-04
Fix shader binding table with device address == 0

arno-lunarg:
- Fix for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5035
(The issue was spotted in the CTS)

When VkStridedDeviceAddressRegionKHR.deviceAddress is 0, the region is considered unused, and this is a valid behaviour, [as per the spec](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkStridedDeviceAddressRegionKHR).

Also, the case where a user provided device address cannot be associated to a buffer using the internal `buffer_address_map_` does not fall under the umbrella of any VUID it seems. So I added `UNASSIGNED-CoreValidation-VkDeviceAddress-NoAssociatedBuffer` to output a proper error for that case. Using an invalid address (or one that became invalid following a buffer destruction) seems likely to me, so better have a custom error.

### DIFF
--- a/layers/core_validation_error_enums.h
+++ b/layers/core_validation_error_enums.h
@@ -1,7 +1,7 @@
-/* Copyright (c) 2015-2022 The Khronos Group Inc.
- * Copyright (c) 2015-2022 Valve Corporation
- * Copyright (c) 2015-2022 LunarG, Inc.
- * Copyright (C) 2015-2022 Google Inc.
+/* Copyright (c) 2015-2023 The Khronos Group Inc.
+ * Copyright (c) 2015-2023 Valve Corporation
+ * Copyright (c) 2015-2023 LunarG, Inc.
+ * Copyright (C) 2015-2023 Google Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


layers/core_validation_error_enums.h
```diff
@@ -84,6 +84,8 @@
 
 [[maybe_unused]] static const char *kVUID_Core_invalidDepthStencilFormat = "UNASSIGNED-CoreValidation-depthStencil-format";
 
+[[maybe_unused]] static const char *kVUID_Core_DeviceAddress_NoAssociatedBuffer = "UNASSIGNED-CoreValidation-VkDeviceAddress-NoAssociatedBuffer";
```
ncesario-lunarg, line 87: @spencer-lunarg is/should this be on WG radar?

ncesario-lunarg (approved):
LGTM!

Tony-LunarG (approved):
LGTM, fixes errors in CTS

layers/core_validation_error_enums.h
spencer-lunarg: 100% should be, will raise the internal issue

spencer-lunarg: https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5635

arno-lunarg: Thank you Spencer!

PR #5046, opened 2023-01-05
layers: Collapse ValidateShaderModuleId logic

spencer-lunarg:
Noticed this function was reusing the same logic in checks instead of just grouping it together


arno-lunarg (approved):
LGTM!

PR #5047, opened 2023-01-05
Fix for sparse shader binding table buffer validation

arno-lunarg:
(I forgot to add that to https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5039)
For the shader binding table buffers mentioned in trace rays commands, their memory binding needs to be validated if and only if the buffer is not sparse.

The concerned VUID family is the one mentioned in https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5035
Their definition is the same, only the `deviceAddress` member changes. e.g.:
> VUID-vkCmdTraceRaysKHR-pRayGenShaderBindingTable-03680
If the buffer from which pRayGenShaderBindingTable->deviceAddress was queried is non-sparse then it must be bound completely and contiguously to a single VkDeviceMemory object

### DIFF
--- a/layers/drawdispatch.cpp
+++ b/layers/drawdispatch.cpp
@@ -2256,8 +2256,10 @@ bool CoreChecks::ValidateRaytracingShaderBindingTable(VkCommandBuffer commandBuf
                 if (!(static_cast<uint32_t>(buffer_state->createInfo.usage) & VK_BUFFER_USAGE_SHADER_BINDING_TABLE_BIT_KHR)) {
                     return false;
                 }
-                if (const auto mem_state = buffer_state->MemState(); !mem_state || mem_state->Destroyed()) {
-                    return false;
+                if (!buffer_state->sparse) {
+                    if (const auto mem_state = buffer_state->MemState(); !mem_state || mem_state->Destroyed()) {
+                        return false;
+                    }
                 }
                 if (binding_table.size != 0) {
                     const auto device_address_range = buffer_state->DeviceAddressRange();


layers/drawdispatch.cpp
```diff
@@ -2256,8 +2256,10 @@ bool CoreChecks::ValidateRaytracingShaderBindingTable(VkCommandBuffer commandBuf
                 if (!(static_cast<uint32_t>(buffer_state->createInfo.usage) & VK_BUFFER_USAGE_SHADER_BINDING_TABLE_BIT_KHR)) {
                     return false;
                 }
-                if (const auto mem_state = buffer_state->MemState(); !mem_state || mem_state->Destroyed()) {
-                    return false;
+                if (!(buffer_state->createInfo.flags & VK_BUFFER_CREATE_SPARSE_RESIDENCY_BIT)) {
```
spencer-lunarg, line 2259: what about `VK_BUFFER_CREATE_SPARSE_BINDING_BIT` and `VK_BUFFER_CREATE_SPARSE_ALIASED_BIT` ?

arno-lunarg: My knowledge on sparse resources is today old, so I do not know the answer to that : /
[From the spec](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#sparsememory-sparseresourcefeatures):
> Requested via the VK_IMAGE_CREATE_SPARSE_RESIDENCY_BIT and VK_BUFFER_CREATE_SPARSE_RESIDENCY_BIT bits.

`CoreChecks::ValidateMemoryIsBoundToBuffer` only checks for `VK_BUFFER_CREATE_SPARSE_BINDING_BIT`

So I am lost!

spencer-lunarg: So `sparse` is just a concept and represented by `SPARSE_BINDING_BIT`

Then there are 2 additional options on top of that (`SPARSE_RESIDENCY_BIT` and `SPARSE_ALIASED_BIT`)

The spec points this out also in `VUID-VkBufferCreateInfo-flags-00918` and here

> Buffers created with this flag must also be created with the VK_BUFFER_CREATE_SPARSE_BINDING_BIT flag.

so really we only ever need to check for `VK_BUFFER_CREATE_SPARSE_BINDING_BIT`

-----

for now, change this to `VK_BUFFER_CREATE_SPARSE_BINDING_BIT` as that will cover all 3 bits

as a follow-up PR, it would be nice to just make a `bool sparse` in the `BUFFER_STATE` and query that... then in the code where we set `bool sparse` we leave a comment with what I described above for the next engineer who hits this

arno-lunarg: Will do so, thank you for the explanation!!

spencer-lunarg: So I just looked, `BINDABLE` has a `sparse` bool already so just use `buffer_state->sparse`

spencer-lunarg: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5063

jeremyg-lunarg: So does this PR go away now or does it rebase on top of 5063?

spencer-lunarg: on top, I didn't touch this code

spencer-lunarg (approved):

layers/drawdispatch.cpp
arno-lunarg: done!

PR #5050, opened 2023-01-05
layers: Fix mislabled Video VUID

spencer-lunarg:
Just a trivial fixing of the VUID being used

cc @aqnuep


aqnuep:
Looks good. Not sure how you spotted those, but if you have some tools for that I'd be happy to use them too. It's too easy to make these copy-paste errors. Thanks @spencer-lunarg.

ncesario-lunarg (approved):

ncesario-lunarg:
@aqnuep I think the following might work in this case:
```bash
python ./scripts/vk_validation_stats.py external/Vulkan-Headers/registry/validusage.json -c
```

spencer-lunarg:
what @ncesario-lunarg said, on my todo to add to the documentation already (or honestly CI again)

PR #5051, opened 2023-01-05
Add GPU-AV debug option to validate instrumented shaders

Tony-LunarG:
This is for debugging GPU-AV, and not really valuable to an end user.  It does catch the current problem with VkGpuAssistedLayerTest.GpuBufferDeviceAddressOOB.

### DIFF
--- a/layers/gpu_validation.cpp
+++ b/layers/gpu_validation.cpp
@@ -1,6 +1,6 @@
-/* Copyright (c) 2018-2022 The Khronos Group Inc.
- * Copyright (c) 2018-2022 Valve Corporation
- * Copyright (c) 2018-2022 LunarG, Inc.
+/* Copyright (c) 2018-2023 The Khronos Group Inc.
+ * Copyright (c) 2018-2023 Valve Corporation
+ * Copyright (c) 2018-2023 LunarG, Inc.
  *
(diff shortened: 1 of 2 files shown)


jeremyg-lunarg (approved):
LGTM

layers/gpu_validation.cpp
```diff
@@ -895,6 +897,20 @@ void GpuAssisted::PreCallRecordDestroyRenderPass(VkDevice device, VkRenderPass r
     ValidationStateTracker::PreCallRecordDestroyRenderPass(device, renderPass, pAllocator);
 }
 
+bool GpuValidateShader(uint32_t *code, const size_t code_size, bool SetRelaxBlockLayout, bool SetScalerBlockLayout, std::string &error) {
```
ncesario-lunarg, line 900: _Could_ use `layer_data::span` to encapsulate `code` and `code_size` and then just pass in `new_pgm` below to avoid the "pointer + size dual parameters," but this is kind of a more general comment/question about all the "pointer + size" parameters we have, and wondering if `span` helps with readability or not.

ncesario-lunarg (approved):
LGTM!

ncesario-lunarg (approved):

PR #5054, opened 2023-01-05
Apply render pass final transitions even if attachment is not used

Rua:
I noticed this mistake, perhaps reflecting an older state of the spec. The current spec says this:
> If an attachment is not used by any subpass, `loadOp`, `storeOp`, `stencilStoreOp`, and `stencilLoadOp` will be ignored for that attachment, and no load or store ops will be performed. However, any transition specified by `initialLayout` and `finalLayout` will still be executed.

The code previously added the transition to `finalLayout` only if the attachment was used at least one subpass (via the `last` variable) but that is wrong.

### DIFF
--- a/layers/render_pass_state.cpp
+++ b/layers/render_pass_state.cpp
@@ -213,8 +213,8 @@ struct AttachmentTracker {  // This is really only of local interest, but a bit 
 
         for (uint32_t attachment = 0; attachment < attachment_count; ++attachment) {
             const auto final_layout = rp->createInfo.pAttachments[attachment].finalLayout;
-            // Add final transitions for attachments that were used and change layout.
-            if ((last[attachment] != VK_SUBPASS_EXTERNAL) && final_layout != attachment_layout[attachment]) {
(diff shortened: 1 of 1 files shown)


juan-lunarg:
Ideally in these scenarios we also add a positive test to prevent a regression in the future.

Here is a good minimal:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4779

Let me know if you need help writing a test.

Rua:
> Let me know if you need help writing a test.

I don't really feel confident with the code to write a test like that, sorry.

juan-lunarg:
> I don't really feel confident with the code to write a test like that, sorry. C++ is not my normal language and it's been many years since I used it.

If you can describe the valid usage scenario (in a Github Issue) which prompted this PR I can write the test for it in another PR.

Rua:
I didn't encounter it while running the validation layers, I literally spotted it while looking at the source code itself and then conferring with the spec.

PR #5055, opened 2023-01-05
docs: Update repo contacts

ncesario-lunarg:


Tony-LunarG (approved):
LGTM

README.md
```diff
@@ -22,9 +22,8 @@ to assist development by enabling developers to verify their applications correc
 This repository contains both the [*Validation Layers*](layers/) source as well as [*Tests*](tests/) for them.
 
 ## Contact Information
-* [Mark Lobodzinski](mailto:mark@lunarg.com)
-* [Nate Cesario](mailto:nathaniel@lunarg.com)
-* Can also be found on the [Khronos Slack](https://khr.io/slack)
+* @KhronosGroup/VVL-CODEOWNERS
```
juan-lunarg, line 25: I don't think regular users can see KhronosGroup teams. I just tried in a private Window. Most likely for security/privacy reasons.

juan-lunarg: To clarify this is something in the KhronosGroup organization settings. Generally teams are visible on GitHub.

README.md
```diff
@@ -22,9 +22,8 @@ to assist development by enabling developers to verify their applications correc
...
+* VVL-CODEOWNERS members can also be found on the [Khronos Slack](https://khr.io/slack)
```
juan-lunarg, line 26: Should we mention the Vulkan discord?
https://discord.com/invite/vulkan

README.md
ncesario-lunarg: Ok, is there some other way we should/could reference it here?

README.md
ncesario-lunarg: We can, but I'm not sure how many of the VVL developers are on the vulkan discord (e.g., I am not).

juan-lunarg: Yeah even though I'm on it I'm not on there too often. What about https://github.com/KhronosGroup/Vulkan-ValidationLayers/discussions?

README.md
juan-lunarg: I think you can still message the vvl-codeowners by just pinging them at @KhronosGroup/vvl-codeowners either in a discussion/issue.

README.md
ncesario-lunarg: We talked a bit about discussions when they first became available on github, but haven't used them. If we did mention them, then I think we would also need to clarify when they should be used vs creating an issue. For this (VVL) repo, issues seem to be the most natural "entry point" to communication, and that's probably a longer "internal team discussion" we need to have.

README.md
ncesario-lunarg: > I think you can still message the vvl-codeowners by just pinging them at @KhronosGroup/vvl-codeowners either in a discussion/issue.

This was my understanding. Is VVL-CODEOWNERS not public?

README.md
juan-lunarg: I thought we left discussions open because we wanted them. I have them disabled on Vulkan-Headers (https://github.com/KhronosGroup/Vulkan-Headers) since I'd rather only have issues.

ncesario-lunarg: Nope! Didn't know you could disable them :)

I'd vote for disabling them unless someone has argument for keeping them.

juan-lunarg: I'd vote for disabling as well.

ncesario-lunarg: Done!

ncesario-lunarg:
Going to merge this for now; we can always update later.

PR #5059, opened 2023-01-05
layers: Allow binding empty DS with empty layout

ncesario-lunarg:
Allow an empty descriptor set to be bound if the current layout also
defines an empty descriptor set.

### DIFF
--- a/layers/descriptor_validation.cpp
+++ b/layers/descriptor_validation.cpp
@@ -166,13 +166,7 @@ bool CoreChecks::VerifySetLayoutCompatibility(const cvdescriptorset::DescriptorS
     if (descriptor_set.IsPushDescriptor()) return true;
     const auto *layout_node = pipeline_layout.set_layouts[layoutIndex].get();
     if (layout_node) {
-        if ((descriptor_set.GetBindingCount() > 0) && (layout_node->GetBindingCount() > 0)) {
-            return VerifySetLayoutCompatibility(*layout_node, *descriptor_set.GetLayout(), errorMsg);
-        } else {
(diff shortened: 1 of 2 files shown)


layers/descriptor_validation.cpp
```diff
@@ -166,7 +166,8 @@ bool CoreChecks::VerifySetLayoutCompatibility(const cvdescriptorset::DescriptorS
     if (descriptor_set.IsPushDescriptor()) return true;
     const auto *layout_node = pipeline_layout.set_layouts[layoutIndex].get();
     if (layout_node) {
-        if ((descriptor_set.GetBindingCount() > 0) && (layout_node->GetBindingCount() > 0)) {
+        if (((descriptor_set.GetBindingCount() > 0) && (layout_node->GetBindingCount() > 0)) ||
+            (descriptor_set.GetBindingCount() == layout_node->GetBindingCount())) {
```
Tony-LunarG, line 170: Is this the same as saying they are both 0?

ncesario-lunarg: Yeah, I should have added a comment to make that clear. I'm still not 100% clear if this is correct for graphics pipeline libraries + independent sets, but I'm fairly confident this is right w.r.t. the spec.

That said, this whole condition is now really just here so that a GPL-specific "note" can be printed out, and I don't think it's all that helpful. I'm going to try getting rid of this whole condition which should catch all errors and simplify this a bit.

juan-lunarg:
FYI to @Kangz this should address https://bugs.chromium.org/p/dawn/issues/detail?id=1627

Kangz:
Thank you!

PR #5068, opened 2023-01-06
cmake: Replace Visual Studio user file with cmake properties

artem-lunarg:


tests/CMakeLists.txt
```diff
@@ -107,6 +103,10 @@ if (MSVC AND CMAKE_SIZEOF_VOID_P EQUAL 4)
     target_link_options(vk_layer_validation_tests PRIVATE /LARGEADDRESSAWARE)
 endif()
 
+if(MSVC_IDE)
+    set_target_properties(vk_layer_validation_tests PROPERTIES VS_DEBUGGER_ENVIRONMENT "VK_LAYER_PATH=$(SolutionDir)\\layers\\$(Configuration)")
```
juan-lunarg, line 107: More robust to use generator expressions in this situation since `VS_DEBUGGER_ENVIRONMENT` [supports](https://cmake.org/cmake/help/latest/prop_tgt/VS_DEBUGGER_ENVIRONMENT.html?highlight=vs_debugger_environment) it.

```suggestion
    set_target_properties(vk_layer_validation_tests PROPERTIES VS_DEBUGGER_ENVIRONMENT "VK_LAYER_PATH=$<TARGET_FILE_DIR:VkLayer_khronos_validation>")
```

juan-lunarg (changes requested):

tests/CMakeLists.txt
artem-lunarg: Thanks for the suggestion!

juan-lunarg (approved):
Minor nitpick. I'd squash the 2 commits into 1 commit, and force push. The 2nd commit doesn't have a lot of value.

PR #5070, opened 2023-01-06
layers: Move Video CoreCheck to own file

spencer-lunarg:
Creates a new video_validation.cpp file to move Vulkan Video related CoreChecks into

@aqnuep we decided recently to try and break up what became a 25k line `core_validation.cpp` and took your video code and moved it into its own file (git is acting strange, but I just `ctrl+x ctrl+v` it

There is a second commit where I made use of the new BitSet count helper we have now as well


aqnuep:
Sounds good. I thought about splitting the video related CoreChecks as well, but didn't want to do it preemptively unless requested. Next time, please feel to indicate these on the reviews as I would have been happy to do that split myself.

layers/video_validation.cpp
```diff
@@ -0,0 +1,1774 @@
...
+    }
+
+    if (profile->chromaSubsampling != VK_VIDEO_CHROMA_SUBSAMPLING_MONOCHROME_BIT_KHR) {
+        if (GetBitSetCount(profile->chromaBitDepth) != 1) {
+            skip |= LogError(object, "VUID-VkVideoProfileInfoKHR-chromaSubsampling-07015",
+                             "%s(): chromaBitDepth in %s must have a single bit set", api_name, where);
+        }
+    }
+
+    switch (profile->videoCodecOperation) {
```
Tony-LunarG, line 161: Should the beta _EXT enums also be considered in this switch, or is the assert when using them really intended behavior?

aqnuep: Yes, the encode codec operations will eventually land in this switch-case as well, as part of the video encode validation efforts.

Tony-LunarG: And old code that uses, for instance, VK_VIDEO_CODEC_OPERATION_DECODE_H264_BIT_EXT will assert here. Is that desired?

aqnuep: No, that won't be the case as the KHR extensions use identical enum values to the old EXT ones. But it doesn't matter in the end because the provisional EXT extensions changed multiple times in a non-backwards-compatible way. Nonetheless, the KHR enums/bits/etc. are all identical to the corresponding APIs of the last revision of the provisional EXT.

Tony-LunarG: Thanks

charles-lunarg (approved):

jeremyg-lunarg (approved):
LGTM

PR #5071, opened 2023-01-06
tests: Fix jump to null function pointer

karl-lunarg:
Found on lavapipe 22.3.2.

Lavapipe doesn't support the VK_KHR_SAMPLER_YCBCR_CONVERSION_EXTENSION.

This test uses this extension by calling one of its functions to test the case where one of its dependent extensions isn't enabled.  The trouble with this is that if the sampler extension isn't supported, the test should not call the extension function and should skip the rest of the test.

This commit modifies the check for the extension to skip the test in this situation.

### DIFF
--- a/tests/vklayertests_buffer_image_memory_sampler.cpp
+++ b/tests/vklayertests_buffer_image_memory_sampler.cpp
@@ -1,8 +1,8 @@
 /*
- * Copyright (c) 2015-2022 The Khronos Group Inc.
- * Copyright (c) 2015-2022 Valve Corporation
- * Copyright (c) 2015-2022 LunarG, Inc.
- * Copyright (c) 2015-2022 Google, Inc.
+ * Copyright (c) 2015-2023 The Khronos Group Inc.
+ * Copyright (c) 2015-2023 Valve Corporation
+ * Copyright (c) 2015-2023 LunarG, Inc.
+ * Copyright (c) 2015-2023 Google, Inc.
  * Modifications Copyright (C) 2020 Advanced Micro Devices, Inc. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -10103,8 +10103,8 @@ TEST_F(VkLayerTest, ExtensionNotEnabled) {
         if (DeviceExtensionSupported(dev_ext)) {
             m_device_extension_names.push_back(dev_ext);
         } else {
-            printf("Did not find required device extension %s; skipped.\n", dev_ext);
-            break;
+            // Need to get out of the test now so that the subsequent code doesn't try to use an extension that isn't enabled.
+            GTEST_SKIP() << "Did not find required device extension: " << dev_ext;
         }
     }
 


tests/vklayertests_buffer_image_memory_sampler.cpp
```diff
@@ -10128,8 +10128,13 @@ TEST_F(VkLayerTest, ExtensionNotEnabled) {
...
     VkSamplerYcbcrConversion conversion;
-    vkCreateSamplerYcbcrConversionKHR(m_device->handle(), &ycbcr_create_info, nullptr, &conversion);
-    m_errorMonitor->VerifyFound();
+    if (vkCreateSamplerYcbcrConversionKHR) {
+        vkCreateSamplerYcbcrConversionKHR(m_device->handle(), &ycbcr_create_info, nullptr, &conversion);
+        m_errorMonitor->VerifyFound();
+    } else {
+        // This shouldn't happen because the above checks should cause an early skip if this extension isn't there.
+        printf("Unexpected nullptr for PFN_vkCreateSamplerYcbcrConversionKHR\n");
+    }
```
juan-lunarg, line 10137: `VkLayerTest::GetDeviceProcAddr` asserts on `vk::GetDeviceProcAddr` returning nullptr.

```suggestion
    auto vkCreateSamplerYcbcrConversionKHR =
GetDeviceProcAddr<PFN_vkCreateSamplerYcbcrConversionKHR>("vkCreateSamplerYcbcrConversionKHR");
    vkCreateSamplerYcbcrConversionKHR(m_device->handle(), &ycbcr_create_info, nullptr, &conversion);
    m_errorMonitor->VerifyFound();
```

juan-lunarg (approved):
Minor nitpick

tests/vklayertests_buffer_image_memory_sampler.cpp
```diff
@@ -10128,8 +10128,13 @@ TEST_F(VkLayerTest, ExtensionNotEnabled) {
     ycbcr_create_info.chromaFilter = VK_FILTER_NEAREST;
     ycbcr_create_info.forceExplicitReconstruction = false;
     VkSamplerYcbcrConversion conversion;
-    vkCreateSamplerYcbcrConversionKHR(m_device->handle(), &ycbcr_create_info, nullptr, &conversion);
-    m_errorMonitor->VerifyFound();
+    if (vkCreateSamplerYcbcrConversionKHR) {
```
ncesario-lunarg, line 10131: Should this check occur right after the `GetDeviceProcAddr` call and just `GTEST_SKIP` if null? Seems like there's no point in doing any of the test if that function is null. Also, having a `SetDesiredFailureMsg` without a `VerifyFound` seems wrong/confusing; i.e., in the case where the function pointer is null, we "expect" an error, but then never verify it's presence...so probably no point in running the code?

ncesario-lunarg (commented):
One question, but I'd be good with merging as is if it makes sense to others.

tests/vklayertests_buffer_image_memory_sampler.cpp
arno-lunarg: `assert` in only defined in debug builds I think

tests/vklayertests_buffer_image_memory_sampler.cpp
karl-lunarg: I've removed this part of the change, which is actually unnecessary.  The real fix is changing the `break` in the extensions checking loop just above to a `GTEST_SKIP()`, which I consider to be a typo.  I think the intent in that loop was really to not try running the test instead of just leaving the loop which didn't accomplish anything but print a message.

juan-lunarg (approved):

ncesario-lunarg (approved):

PR #5072, opened 2023-01-06
layers: Remove optick instrumentation 

juan-lunarg:
As far as I can tell optick used to be a dependency of VVL? Since it was referenced as being in the `external/` folder.

Thoughts?


layers/CMakeLists.txt
```diff
@@ -237,36 +236,6 @@ set(SYNC_VALIDATION_LIBRARY_FILES
     synchronization_validation.cpp
     synchronization_validation.h)
 
-# Validation Layer performance instrumentation support using Optick.
-# https://optick.dev/ https://github.com/bombomby/optick
-# To include Optick instrumentation:
-# 1) Download the Optick 1.3.1 release: https://github.com/bombomby/optick/releases/tag/1.3.1.0
-# 2) Copy the files from the src directory to external/optick.
-# 3) Run CMake configure with -DINSTRUMENT_OPTICK.
-set(OPTICK_SOURCE_DIR "${PROJECT_SOURCE_DIR}/external/optick")
```
juan-lunarg, line 246: This is what gives me the indication we don't use it anymore. Since it isn't listed in `known_good.json` which AFAIK are the only things that are moved into the `external/` directory.

juan-lunarg:
For clarification my PR is more about finding out who is using the optick developer functionality.

If someone is using it now, I'd like to clean up the CMake code around it.

If no one is using it I'd like to remove it.

ncesario-lunarg:
From what I can recall, this was added by @karl-lunarg (and used?) for performance profiling. While it's probably not used often, I'm not sure we want to remove it entirely: if we _do_ ever want the level of detail optick profiling provides, it would be unfortunate to have to resurrect Karl's changes down the road.

That said, @jeremyg-lunarg probably has the most insight as to what was useful during performance profiling, so probably good to get his input here first.

jeremyg-lunarg:
> That said, @jeremyg-lunarg probably has the most insight as to what was useful during performance profiling, so probably good to get his input here first.

I haven't used this often, most of the time it ends up being better to use an external CPU profiler, which can see more detail about what happens in our code for each vulkan call.

ncesario-lunarg:
> I haven't used this often, most of the time it ends up being better to use an external CPU profiler, which can see more detail about what happens in our code for each vulkan call.

In that case, sounds like it's ok to remove it. I'm still a little hesitant to just remove this as it hasn't cost us any maintenance time since it is completely auto-generated. Maybe @karl-lunarg can weigh in, as there might have been a specific reason he added this that I'm forgetting?

juan-lunarg:
> In that case, sounds like it's ok to remove it. I'm still a little hesitant to just remove this as it hasn't cost us any maintenance time since it is completely auto-generated.

Agreed it's not much of a maintenance burden. Which is why I'm also offering to clean it up. This PR is more of question about the state of this developer feature.

layers/CMakeLists.txt
karl-lunarg: Don't over-associate the update_deps/known_good mechanism with the `external` directory.   The targets of update_deps can be placed anywhere and then the `helper.cmake` written to the same place tells CMake where the targets are.  The `external` directory was just a convenient place to put the Optick code.

karl-lunarg:
I'd be OK with removing it.  I don't think we're using it much and it was sort of experimental to see if it would "stick".  And I suppose that if we ever wanted to resurrect it, we could fish out the commits that added or removed it and figure out how to put it back pretty easily.

karl-lunarg:
And you may want to refer to 5e5a05fabea2560c5e6f41c9e6e051face0de6f5 for complete removal.  I tried a revert and had a lot of conflicts, but they weren't too bad.

layers/CMakeLists.txt
juan-lunarg: That makes more sense. I was confused why I didn't see it in the history for `known_good.json` :+1:

juan-lunarg:
All CI is passing now. Just a minor blip in the azure services cause a github action to fail.

Does this look good to you @karl-lunarg ?

karl-lunarg (approved):
Looks good.  I compared it to the original commit.  Thanks for doing this!

jeremyg-lunarg (approved):
LGTM

PR #5077, opened 2023-01-08
layers: Move Query CoreChecks to own file

spencer-lunarg:
Move all Queries related CoreChecks to `query_validation.cpp`


arno-lunarg (approved):
LGTM!

PR #5081, opened 2023-01-09
tests: Add note about implicit validation layers

artem-lunarg:
Add section to tests README.

~~Update VS debugging environment to exclude known implicit layers that affect validation output.~~

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -20,6 +20,15 @@ The tests are grouped into different categories. Some of the main test categorie
 - `VkGpuAssistedLayerTest` - Test [GPU-Assisted Validation](../docs/gpu_validation.md)
 - `VkPortabilitySubsetTest` - Test [VK_KHR_portability_subset validation](../docs/portability_validation.md)
 
+## Implicit Layers note
+
+In general, the implicit validation layers may change the expected message output, which may cause some tests to fail. One solution is to disable the implicit layer during the test session using the `VK_LOADER_LAYERS_DISABLE` environment variable introduced in Vulkan Loader v.1.3.234.
+
+Here is an example with OBS implicit layer:
+```bash
+export VK_LOADER_LAYERS_DISABLE=VK_LAYER_OBS_HOOK
+```
+
 ## Running Test on Linux
 
 **IMPORTANT** Make sure you have the correct `VK_LAYER_PATH` set on Linux


tests/CMakeLists.txt
```diff
@@ -104,7 +104,12 @@ if (MSVC AND CMAKE_SIZEOF_VOID_P EQUAL 4)
 endif()
 
 if(MSVC_IDE)
-    set_target_properties(vk_layer_validation_tests PROPERTIES VS_DEBUGGER_ENVIRONMENT "VK_LAYER_PATH=$<TARGET_FILE_DIR:VkLayer_khronos_validation>")
+    string(JOIN "\n" CUSTOM_VS_ENVIRONMENT
+        "VK_LAYER_PATH=$<TARGET_FILE_DIR:VkLayer_khronos_validation>"
+        # Disable implicit layers that affect the validation message output
+        "VK_LOADER_LAYERS_DISABLE=VK_LAYER_OBS_HOOK"
```
juan-lunarg, line 110: I'm not a fan of this approach. Modifying the CMake to be aware of a specific application is strange to read and ages poorly.

juan-lunarg (changes requested):
We have had implicit layers cause problems in the past but AFAIK never came up with a one size fits all approach.

FWIW:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/4b5a19722073afa2394e51092d82a4a655912279

The added documentation looks fine though.

tests/CMakeLists.txt
ncesario-lunarg: I personally agree with Juan here. I would prefer leaving the cmake as is. Yes folks will have to update their VS projects every time they re-generate cmake (although maybe VS has something similar to VSCode's "launch.json" that can be used to persist these settings?), but I think that is preferable to adding workarounds in cmake.

ncesario-lunarg (commented):
The documentation LGTM, but I would also prefer leaving the workaround out of cmake.

tests/CMakeLists.txt
artem-lunarg: I agree, documentation should give enough information how handle this issue.

ncesario-lunarg (approved):

tests/CMakeLists.txt
artem-lunarg: > Yes folks will have to update their VS projects every time they re-generate cmake.

It should be fine; manually entered debug environment settings are stored in the `*.user` file and preserved when build system is re-generated.

PR #5083, opened 2023-01-09
build: Use mimalloc to improve win64 performance

jeremyg-lunarg:
When instrumenting shaders for GPU-AV, SPIRV-Tools spends a large amount of time allocating and freeing small buffers. Using [mimalloc  ](https://github.com/microsoft/mimalloc) instead of the standard system malloc can improve performance for this use case by about 20%. Other validation use cases appear to have performance improvement of 5-10% from this change.

Currently this is only enabled for 64-bit MSVC builds. Testing on other OSes shows stability problems when statically linking mimalloc into the layer. For these platforms, mimalloc can be used by including mimalloc.so in `LD_PRELOAD`.


layers/CMakeLists.txt
```diff
@@ -100,6 +100,12 @@ if (USE_ROBIN_HOOD_HASHING)
     target_compile_definitions(robin_hood::robin_hood INTERFACE USE_ROBIN_HOOD_HASHING)
 endif()
 
+find_package(mimalloc CONFIG)
```
juan-lunarg, line 103: FWIW mimalloc is available as an [Ubuntu package](https://packages.ubuntu.com/search?suite=kinetic&section=all&arch=any&keywords=mimalloc&searchon=names). So I think we can get away with making it required.

Thoughts?

jeremyg-lunarg: hmm. I guess I lean against making it required because you won't  strictly NEED it since there's always the default malloc/free.  But its also in fedora so, 🤷

juan-lunarg:
> Using mimalloc instead of the standard system malloc can improve performanc for this use case by about 20%

Was this true across multiple platforms?

layers/CMakeLists.txt
```diff
@@ -93,6 +93,17 @@ if (USE_ROBIN_HOOD_HASHING)
     target_compile_definitions(robin_hood::robin_hood INTERFACE USE_ROBIN_HOOD_HASHING)
 endif()
 
+# Using mimalloc on non-Windows OSes currently results in unit test instability with some
+# OS version / driver combinations.
```
juan-lunarg, line 97: Does `mimalloc` work well with MSVC's address sanitizer?
https://learn.microsoft.com/en-us/cpp/build/reference/fsanitize?view=msvc-170

scripts/known_good.json
```diff
@@ -64,6 +64,20 @@
...
+            "sub_dir": "mimalloc",
+            "build_dir": "mimalloc/build",
+            "install_dir": "mimalloc/build/install",
+            "cmake_options": [
+                "-DMI_BUILD_STATIC=ON",
+                "-DMI_BUILD_OBJECT=OFF",
+                "-DMI_BUILD_SHARED=OFF",
+                "-DMI_BUILD_TESTS=OFF"
+            ],
+            "commit": "v2.0.7"
```
juan-lunarg, line 80: Specify `build_platforms` to avoid downloads for non-win32 platforms.

```suggestion
            "commit": "v2.0.7"
            "build_platforms" : [ "windows" ]
```

layers/state_tracker.cpp
```diff
@@ -1,7 +1,7 @@
-/* Copyright (c) 2015-2022 The Khronos Group Inc.
- * Copyright (c) 2015-2022 Valve Corporation
- * Copyright (c) 2015-2022 LunarG, Inc.
- * Copyright (C) 2015-2022 Google Inc.
+/* Copyright (c) 2015-2023 The Khronos Group Inc.
```
juan-lunarg, line 1: Why is this file being updated?

layers/vk_layer_utils.cpp
```diff
@@ -28,6 +28,12 @@
 #include "vulkan/vulkan.h"
 #include "vk_layer_config.h"
 
+// Include new / delete overrides for windows if using mimalloc.  This needs to be include exactly once in a file that is
+// part of the layer utils library.
```
juan-lunarg, line 32: How does this propagate to `VkLayer_khronos_validation`?

layers/state_tracker.cpp
jeremyg-lunarg: I moved stuff around and it no longer should be changed.

layers/CMakeLists.txt
jeremyg-lunarg: I haven't tried that yet. It does work with asan on linux.

layers/vk_layer_utils.cpp
jeremyg-lunarg: they're all part of the same .so so I *think* it propagates. I'll double check I guess.

jeremyg-lunarg: yup. If I put back the #include in state_tracker.cpp I get:
`4>VkLayer_utils.lib(vk_layer_utils.obj) : error LNK2005: "void * __cdecl operator new[](unsigned __int64,enum std::align_val_t,struct std::nothrow_t const &)" (??_U@YAPEAX_KW4align_val_t@std@@AEBUnothrow_t@1@@Z) already defined in state_tracker.obj`

layers/vk_layer_utils.cpp
```diff
@@ -28,6 +28,12 @@
 #include "vulkan/vulkan.h"
 #include "vk_layer_config.h"
 
+// Include new / delete overrides if using mimalloc. This needs to be include exactly once in a file that is
+// part of the layer utils library.
+#if defined(USE_MIMALLOC)
```
juan-lunarg, line 33: Ensure MIMALLOC is only used on WIN64 builds.

```suggestion
#if defined(USE_MIMALLOC) && defined(_WIN64)
```

layers/CMakeLists.txt
```diff
@@ -332,7 +343,8 @@ if (USE_ROBIN_HOOD_HASHING)
     # This warning produces what look like false positives in robin_hood.h with Visual Studio 2015
     target_compile_options(VkLayer_khronos_validation PRIVATE "$<$<AND:$<CXX_COMPILER_ID:MSVC>,$<VERSION_LESS:$<CXX_COMPILER_VERSION>,19.1>>:/wd4996>")
 endif()
-target_link_libraries(VkLayer_khronos_validation PRIVATE VVL-SPIRV-LIBS)
+# Order matters here. VkLayer_utils should be the last link library to ensure mimalloc overrides are picked up correctly.
+target_link_libraries(VkLayer_khronos_validation PRIVATE VVL-SPIRV-LIBS VkLayer_utils)
```
juan-lunarg, line 347: What will happen if it isn't the last library? Will it fail silently? Or will the build fail?

BUILD.md
```diff
@@ -101,6 +101,11 @@ version of that dependency.
 - [robin-hood-hashing repository](https://github.com/martinus/robin-hood-hashing)
     - This is a header-only reimplementation of `std::unordered_map` and `std::unordered_set` which provides substantial performance improvements on all platforms.
     - You must clone this repository and build its `install` target
+- [mimalloc repository](https://github.com/microsoft/mimalloc)
+    - This is a reimplementation of malloc()/free() and their c++ equivalents.
+    - It is currently only used for windows 64 bit builds, where it is statically linked into the layer.
+    - For Linux and MacOS, mimalloc can be used by installing it as package and then including it `LD_PRELOAD` as described in the mimalloc documentation.
```
juan-lunarg, line 107: Are we recommending this to folks? I think it's best we don't since we don't test it on our CI.

juan-lunarg (approved):
LGTM just some minor comments

layers/CMakeLists.txt
jeremyg-lunarg: the libraries after it end up not using the overrides and we get slow again.

BUILD.md
jeremyg-lunarg: kinda? the LD_PRELOAD way works well on linux, trying to do it the way this PR does for windows gives you a mess.  I can take this line out.

juan-lunarg: I think it's best to just take out this 1 line for now.

jeremyg-lunarg:
I cancelled the last internal CI build because I pushed comment-only changes.

PR #5085, opened 2023-01-10
layers: Move RenderPass CoreCheck to own file

spencer-lunarg:
Moves all Render Pass (and Dynamic rendering) to a `render_pass_validation.cpp` file


juan-lunarg (approved):
LGTM

Issue #4705, opened 2022-10-25
Handle Wrapping will wrap NULL handles when it shouldn't


charles-lunarg:
**Describe the Issue**
A recent regression in the loader caused a null handle to be returned by terminator_CreateSwapchainKHR. The handle wrapping code dutifully took the null handle, wrapped it, and returned a non-null handle to the application. This hides the underlying issue, that the handle is NULL.

I propose asserting/aborting on trying to wrap a NULL. This should immediately make it clear when the NULL handle is being returned, rather than silently letting the code continue until that null handle is attempted to be used (thus likely crashing in an entirely different function).

PR #5089, opened 2023-01-10
layers: Ensure object handle is not null

artem-lunarg:
It's a safeguard against potential bugs in other layers or the loader.

According to specification, successfully created non-dispatchable object can not be VK_NULL_HANDLE.

closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4705

### DIFF
--- a/layers/generated/chassis.h
+++ b/layers/generated/chassis.h
@@ -4086,6 +4086,9 @@ class ValidationObject {
         // Wrap a newly created handle with a new unique ID, and return the new ID.
         template <typename HandleType>
         HandleType WrapNew(HandleType newlyCreatedHandle) {
+            // The assert is a safeguard against potential bugs in other layers or the loader.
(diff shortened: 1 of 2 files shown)


ncesario-lunarg (approved):
LGTM!

layers/generated/chassis.h
```diff
@@ -4086,6 +4086,7 @@ class ValidationObject {
         // Wrap a newly created handle with a new unique ID, and return the new ID.
         template <typename HandleType>
         HandleType WrapNew(HandleType newlyCreatedHandle) {
+            assert(newlyCreatedHandle != VK_NULL_HANDLE);
```
spencer-lunarg, line 4089: There was a lot of discussion and thought behind this... I agree 100% with it, but see it easily being forgotten later. Can you leave a comment (what you summarized in the PR summary) here to first

artem-lunarg: Updated.

spencer-lunarg (approved):

PR #5091, opened 2023-01-10
cmake: Fix LunarG/VulkanTools build

juan-lunarg:
Regression caused by 4050be7c1960ae0aec00b421bb47be8d67cabe63

Currently LunarG/VulkanTools will fail to build.
This is due to disabling BUILD_LAYERS and BUILD_TESTS, which causes the SPIRV-Headers::SPIRV-Headers dependency to not exist.

### DIFF
--- a/layers/CMakeLists.txt
+++ b/layers/CMakeLists.txt
@@ -15,32 +15,6 @@
 # limitations under the License.
 # ~~~
 
-find_package(PythonInterp 3 QUIET)
-
-if (PYTHONINTERP_FOUND)
-    # Get the include directory of the SPIRV-Headers
-    get_target_property(SPIRV_HEADERS_INCLUDE_DIR SPIRV-Headers::SPIRV-Headers INTERFACE_INCLUDE_DIRECTORIES)
-
-    set(spirv_unified_include_dir "${SPIRV_HEADERS_INCLUDE_DIR}/spirv/unified1/")
-    if (NOT IS_DIRECTORY ${spirv_unified_include_dir})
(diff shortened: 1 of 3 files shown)


layers/CMakeLists.txt
```diff
@@ -15,32 +15,6 @@
 # limitations under the License.
 # ~~~
 
-find_package(PythonInterp 3 QUIET)
-
-if (PYTHONINTERP_FOUND)
-    # Get the include directory of the SPIRV-Headers
-    get_target_property(SPIRV_HEADERS_INCLUDE_DIR SPIRV-Headers::SPIRV-Headers INTERFACE_INCLUDE_DIRECTORIES)
```
juan-lunarg, line 22: This is the cause of the current build failure:

```
This is the current build failure caused by VVL

CMake Error at layers/CMakeLists.txt:22 (get_target_property):
  get_target_property() called with non-existent target
  "SPIRV-Headers::SPIRV-Headers".


CMake Error at layers/CMakeLists.txt:26 (message):
  Unable to find spirv/unified1
```

spencer-lunarg (approved):

jeremyg-lunarg (approved):
LGTM. I always wondered why it would be good for both BUILD_LAYERS and BUILD_TESTS to be false.

juan-lunarg:
> LGTM. I always wondered why it would be good for both BUILD_LAYERS and BUILD_TESTS to be false.

Yep. I just learned the hard way.

ncesario-lunarg (approved):

PR #5092, opened 2023-01-11
Move more code out of core_validation

spencer-lunarg:
- Moved some functions and subpass functions to `render_pass_validation.cpp`
- Moved some functions to `descriptor_validation.cpp`
- Created a `cmd_buffer_dynamic_validation.cpp` to hold all the `CoreChecks` that check `vkCmdSet*` dynamic state


layers/buffer_validation.cpp
```diff
@@ -4370,6 +4370,18 @@ bool CoreChecks::PreCallValidateCmdResolveImage2(VkCommandBuffer commandBuffer,
                                    pResolveImageInfo->pRegions, CMD_RESOLVEIMAGE2);
 }
 
+// Validate that an image's sampleCount matches the requirement for a specific API call
+bool CoreChecks::ValidateImageSampleCount(const IMAGE_STATE *image_state, VkSampleCountFlagBits sample_count, const char *location,
```
arno-lunarg, line 4374: What about havind `image_validation.cpp` and putting this function here?

arno-lunarg (approved):
LGMT! One comment, could probably be resolved in another PR

layers/buffer_validation.cpp
spencer-lunarg: Was thinking the same thing, but currently `buffer_validation` has both the "buffer" and "image" stuff inside, I want to break that up, but figured first goal was to first pick apart `core_validation` as much as possible!

arno-lunarg: yep agreed!

ncesario-lunarg (approved):
🎉

PR #5093, opened 2023-01-11
Cleanup tests for Profiles in CI

spencer-lunarg:
Before making the change to get profiles working in CI, first wanted to 

- Clean up some tests I know locally were failing when using the Profiles (exposed next corner cases)
- Clean up the `./tests/device_profiles` 
   - The current profiles were there because when using devsim we were using them
   - They are not a good spread of what current GPUs are doing
   - Want to limit how many device profiles we use as each one will make CI run longer
   - Picked 3 android related ones as they covered the more nasty edge cases
   - Plan to add a Intel/AMD/NVIDIA desktop profile as well once things are stable (those will cause more tests to be fixed)


ncesario-lunarg (approved):
LGTM! @spencer-lunarg is the plan to do a follow up that runs these profiles in GH actions?

spencer-lunarg:
>  we could even gate running the "official" Android CI on whether or not those tests pass?

So this would run in GitHub CI, so not sure if we can chain the real Android CI to use that pass/fail... first will need to see how long it takes to actually run a single Profile through everything

> is the plan to do a follow up that runs these profiles in GH actions?

Yes! I wanted to get these tests fixes in ASAP in case someone else hits them and doesn't redo the fun work of fixing the tests 😆

Issue #5022, opened 2022-12-30
SecondaryCommandBufferRerecordedExplicitReset | Potential Deadlock


juan-lunarg:
**Describe the Issue**
Run `VkLayerTest.SecondaryCommandBufferRerecordedExplicitReset` with `-fsanitize=thread`.

<details>

<summary>Additional context</summary>

```
Note: Google Test filter = VkLayerTest.SecondaryCommandBufferRerecordedExplicitReset
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from VkLayerTest
[ RUN      ] VkLayerTest.SecondaryCommandBufferRerecordedExplicitReset
==================
WARNING: ThreadSanitizer: lock-order-inversion (potential deadlock) (pid=31730)
  Cycle in lock order graph: M0 (0x7b740001c180) => M1 (0x7b740001cb80) => M0

  Mutex M1 acquired here while holding mutex M0 in main thread:
    #0 pthread_rwlock_rdlock <null> (vk_layer_validation_tests+0x316c5a) (BuildId: 0ce6a4768a5129e569c0e0f4a2aa660e8b816120)
    #1 std::__glibcxx_rwlock_rdlock(pthread_rwlock_t*) /usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/shared_mutex:78:3 (libVkLayer_khronos_validation.so+0x1899173) (BuildId: d027a4abb0643a335e1493658cf25b6315497c33)
    #2 std::__shared_mutex_pthread::lock_shared() /usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/shared_mutex:229:10 (libVkLayer_khronos_validation.so+0x1912bbd) (BuildId: d027a4abb0643a335e1493658cf25b6315497c33)
    #3 std::shared_mutex::lock_shared() /usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/shared_mutex:426:34 (libVkLayer_khronos_validation.so+0x1912b65) (BuildId: d027a4abb0643a335e1493658cf25b6315497c33)
[...]

PR #5094, opened 2023-01-11
cmake: Add VVL_ENABLE_TSAN

juan-lunarg:
Adds thread sanitization to the build.


juan-lunarg:
Helps work on thread sanitization issues like https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5022

And will be needed anyway when I add thread sanitization into CI.

jeremyg-lunarg (approved):
LGTM

PR #5096, opened 2023-01-12
tests: More tests fixes for profiles

spencer-lunarg:
Fixed more tests using profiles from things such as Lavapipe/desktop profiles

cc @karl-lunarg

### DIFF
--- a/tests/positive/image_buffer.cpp
+++ b/tests/positive/image_buffer.cpp
@@ -1,8 +1,8 @@
 /*
- * Copyright (c) 2015-2022 The Khronos Group Inc.
- * Copyright (c) 2015-2022 Valve Corporation
- * Copyright (c) 2015-2022 LunarG, Inc.
- * Copyright (c) 2015-2022 Google, Inc.
+ * Copyright (c) 2015-2023 The Khronos Group Inc.
+ * Copyright (c) 2015-2023 Valve Corporation
+ * Copyright (c) 2015-2023 LunarG, Inc.
+ * Copyright (c) 2015-2023 Google, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -954,14 +954,16 @@ TEST_F(VkPositiveLayerTest, ExternalMemory) {
 #endif
 
     AddRequiredExtensions(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
-    AddRequiredExtensions(VK_KHR_EXTERNAL_MEMORY_CAPABILITIES_EXTENSION_NAME);
+    AddRequiredExtensions(VK_KHR_EXTERNAL_MEMORY_EXTENSION_NAME);
     AddRequiredExtensions(VK_KHR_DEDICATED_ALLOCATION_EXTENSION_NAME);
     AddOptionalExtensions(ext_mem_extension_name);
     ASSERT_NO_FATAL_FAILURE(InitFramework(m_errorMonitor));
     if (!AreRequiredExtensionsEnabled()) {
         GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported";
     }
-
+    if (IsPlatform(kMockICD)) {
+        GTEST_SKIP() << "External tests are not supported by MockICD, skipping tests";
+    }
     // Check for import/export capability
     VkPhysicalDeviceExternalBufferInfoKHR ebi = {VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_EXTERNAL_BUFFER_INFO_KHR, nullptr, 0,
                                                  VK_BUFFER_USAGE_TRANSFER_SRC_BIT | VK_BUFFER_USAGE_TRANSFER_DST_BIT, handle_type};
@@ -1301,6 +1303,11 @@ TEST_F(VkPositiveLayerTest, MultiplaneImageTests) {
     }
     ASSERT_NO_FATAL_FAILURE(InitState(nullptr, &mp_features, VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT));
     ASSERT_NO_FATAL_FAILURE(InitRenderTarget());
+
+    if (!ImageFormatAndFeaturesSupported(gpu(), VK_FORMAT_G8_B8_R8_3PLANE_444_UNORM_KHR, VK_IMAGE_TILING_OPTIMAL,
+                                         VK_FORMAT_FEATURE_COSITED_CHROMA_SAMPLES_BIT)) {
+        GTEST_SKIP() << "Required formats/features not supported";
+    }
 
     // Create aliased function pointers for 1.0 and 1.1 contexts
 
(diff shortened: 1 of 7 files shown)


tests/positive/image_buffer.cpp
```diff
@@ -954,14 +954,16 @@ TEST_F(VkPositiveLayerTest, ExternalMemory) {
...
+    AddRequiredExtensions(VK_KHR_EXTERNAL_MEMORY_EXTENSION_NAME);
     AddRequiredExtensions(VK_KHR_DEDICATED_ALLOCATION_EXTENSION_NAME);
     AddOptionalExtensions(ext_mem_extension_name);
     ASSERT_NO_FATAL_FAILURE(InitFramework(m_errorMonitor));
     if (!AreRequiredExtensionsEnabled()) {
         GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported";
     }
-
+    if (IsPlatform(kMockICD)) {
+        GTEST_SKIP() << "External tests are supported by MockICD, skipping tests";
```
artem-lunarg, line 965: Should it be "not supported by MockICD" ?

tests/vklayertests_others.cpp
```diff
@@ -7433,6 +7433,9 @@ TEST_F(VkLayerTest, ValidateImportMemoryHandleType) {
     if (!AreRequiredExtensionsEnabled()) {
         GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported";
     }
+    if (IsPlatform(kMockICD)) {
+        GTEST_SKIP() << "External tests are supported by MockICD, skipping tests";
```
artem-lunarg, line 7437: The same remark as above, "not supported" ?

tests/vklayertests_dynamic_rendering.cpp
```diff
@@ -1891,6 +1891,9 @@ TEST_F(VkLayerTest, DynamicRenderingInfoMismatchedSamples) {
     image_ci.samples = VK_SAMPLE_COUNT_2_BIT;
     image_ci.tiling = VK_IMAGE_TILING_OPTIMAL;
     image_ci.usage = VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT;
+    if (!(ImageSampleCountSupported(gpu(), image_ci))) {
```
karl-lunarg, line 1894: It seems a shame to toss out some tests when an alternate supported choice could be made.  In this case, we could change the sample count to 4 at line 1891 to provoke the same error and pass without skipping.  It just needs to be different than 1.  I think in a lot of cases, the test authors just needed a different sample count and picked 2 when 1 or 4 would have worked.

I suppose that there may be some tests that really need 2 sample counts and a third to provoke the error where this won't be possible.  Or a test may require 2 for some reason, or a test may be too time-consuming to rewrite to not use 2.  In these cases, then we should skip if not supported.

Can you go through all the "2_BIT" changes and see if you can change each test like this one, instead of skipping?

tests/layer_validation_tests.h
```diff
@@ -205,6 +205,9 @@ VkFormat FindSupportedDepthStencilFormat(VkPhysicalDevice phy);
 bool ImageFormatIsSupported(VkPhysicalDevice phy, VkFormat format, VkImageTiling tiling = VK_IMAGE_TILING_OPTIMAL,
                             VkFormatFeatureFlags features = ~VK_FORMAT_FEATURE_VERTEX_BUFFER_BIT);
 
+// Only VK_SAMPLE_COUNT_1_BIT and VK_SAMPLE_COUNT_4_BIT are required to be supported
+bool ImageSampleCountSupported(VkPhysicalDevice phy, const VkImageCreateInfo info);
```
juan-lunarg, line 209: Nitpick (const ref)

```suggestion
bool ImageSampleCountSupported(VkPhysicalDevice phy, const VkImageCreateInfo& info);
```

tests/positive/image_buffer.cpp
```diff
@@ -1302,6 +1304,11 @@ TEST_F(VkPositiveLayerTest, MultiplaneImageTests) {
     ASSERT_NO_FATAL_FAILURE(InitState(nullptr, &mp_features, VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT));
     ASSERT_NO_FATAL_FAILURE(InitRenderTarget());
 
+    if (!ImageFormatAndFeaturesSupported(gpu(), VK_FORMAT_G8_B8_R8_3PLANE_444_UNORM_KHR, VK_IMAGE_TILING_OPTIMAL,
```
juan-lunarg, line 1307: Considering the importance for writing tests perhaps `ImageFormatAndFeaturesSupported` and `ImageSampleCountSupported` should be mentioned in `creating_tests.md`? (This can be done in another PR of course).

tests/vklayertests_dynamic_rendering.cpp
spencer-lunarg: good idea, this was just to get the tests suppressed, will see if any of these `2_BIT` are "needed"... I rather not have to add yet-another-image-creation-helper-function as well

spencer-lunarg: seems going `2_BIT` -> `4_BIT` was enough to fix this!

tests/positive/image_buffer.cpp
spencer-lunarg: agree, on my "things to improve in the docs" TODO list

tests/vklayertests_command.cpp
```diff
@@ -10094,7 +10094,7 @@ TEST_F(VkLayerTest, ResolveInvalidUsage) {
     image_create_info.extent.depth = 1;
     image_create_info.mipLevels = 1;
     image_create_info.arrayLayers = 1;
-    image_create_info.samples = VK_SAMPLE_COUNT_2_BIT;
```
mark-lunarg, line 10097: I ran across this too but my fix was to query the supported sample count caps and checked to see if `VK_SAMPLE_COUNT_2_BIT` was supported.  Is `VK_SAMPLE_COUNT_2_BIT` always supported (it is on Swiftshader)?

karl-lunarg: The spec requires only sample counts of 1 and 4.  See this [discussion](https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5096#discussion_r1068395555).  I think, where possible if you need a "different" sample count to provoke an error, use a supported value.

mark-lunarg: Perfect, should have known.  Thanks!

mark-lunarg: Actually, it'll depend on the image usage.  The only **guaranteed** sample count is 1.

mark-lunarg (approved):
LGTM.  Failures look to be maybe internet related?

tests/vklayertests_command.cpp
spencer-lunarg: So there were a few tests failing on this, they were all the in the same test group, except `ResolveInvalidUsage` so agree this one in particular needs the fix in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5110

Will remove my change here and let Mark's change fix it

juan-lunarg (approved):

Issue #4809, opened 2022-11-16
Use fallthrough and remove unfinished switch cases


spencer-lunarg:
with C++ 17 the `[[fallthrough]]` syntax can be used to ensure the code is suppose to be missing the `break`

Also some `switch` cases have `TODO` to add the rest of the statements in `default`

juan-lunarg:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/fc1c6286054e5772a506e95d67503dfc9bb2182b/CMakeLists.txt#L192-L196

To enforce this we just need to remove the following CMake. The main problem is the warnings produce by VMA.

PR #5097, opened 2023-01-12
layers: Use fallthrough attribute and enable warning

artem-lunarg:
Enables `-Wimplicit-fallthrough` in both gcc and clang. 

`[[fallthrough]]`s and `break`s are added in a few places.

There are no compiler version checks regarding `-Wimplicit-fallthrough` support. The assumption that this warning is supported by GCC/Clang versions that support C++17/`[[fallthrough]]`.

closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4809

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -194,6 +194,7 @@ endif()
 # Platform-specific compiler switches
 if(${CMAKE_C_COMPILER_ID} STREQUAL "Clang")
     add_compile_options(-Wconversion
+                        -Wimplicit-fallthrough
 
                         # TODO These warnings also get turned on with -Wconversion in some versions of clang.
                         #      Leave off until further investigation.
@@ -213,12 +214,6 @@ if(${CMAKE_C_COMPILER_ID} MATCHES "(GNU|Clang)")
                         -fno-builtin-memcmp)
 
     set(CMAKE_C_STANDARD 99)
-
-    # For GCC version 7.1 or greater, we need to disable the implicit fallthrough warning since there's no consistent way to satisfy
-    # all compilers until they all accept the C++17 standard.
-    if(${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 7.1)
-        add_compile_options(-Wimplicit-fallthrough=0)
-    endif()
 elseif(MSVC)
     # Warn about nested declarations
     add_compile_options("/w34456")
--- a/layers/best_practices_utils.cpp
+++ b/layers/best_practices_utils.cpp
@@ -5130,6 +5130,7 @@ bool BestPractices::PreCallValidateCreatePipelineLayout(VkDevice device, const V
                             break;
                         case VK_DESCRIPTOR_TYPE_INLINE_UNIFORM_BLOCK:
                             descriptor_type_size = 1;
+                            break;
                         default:
                             // Unknown type.
                             break;
--- a/layers/vk_mem_alloc.h
+++ b/layers/vk_mem_alloc.h
@@ -29,6 +29,7 @@
 //    6/05/19 - Make changes to suppress warnings from clang 3.8.0
 //    6/05/19 - Make changes to suppress more warnings from GCC
 //   11/18/20 - Make changes to suppress warnings from clang
+//    1/12/23 - Make changes to suppress -Wimplicit-fallthrough warnings from GCC and clang
 // 
 //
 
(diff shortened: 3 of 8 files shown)


CMakeLists.txt
```diff
@@ -194,6 +194,7 @@ endif()
 # Platform-specific compiler switches
 if(${CMAKE_C_COMPILER_ID} STREQUAL "Clang")
     add_compile_options(-Wconversion
+                        -Wimplicit-fallthrough
```
artem-lunarg, line 197: For clang should be enabled explicitly. For GCC it's included in `-Wextra`

layers/vk_mem_alloc.h
```diff
@@ -29,6 +29,7 @@
 //    6/05/19 - Make changes to suppress warnings from clang 3.8.0
 //    6/05/19 - Make changes to suppress more warnings from GCC
 //   11/18/20 - Make changes to suppress warnings from clang
+//    1/12/23 - Make changes to suppress -Wimplicit-fallthrough warnings from GCC and clang
```
artem-lunarg, line 32: VMA does not guarantee warning free-compilation according to its [documentation](https://gpuopen-librariesandsdks.github.io/VulkanMemoryAllocator/html/general_considerations.html#general_considerations_validation_layer_warnings). The following lines handle implicit fallthrough warnings. The current decision is to disable warnings in one place instead of adding multiple fallthrough attributes to minimize changes to the library.

juan-lunarg: FWIW https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/commit/14469a5eea389f91d64b19198cf2b27732544b01 should have addressed these warnings (I think).

jeremyg-lunarg: We should update again then (separate PR) or even better move this dependency to known_good.json

artem-lunarg: Yes, checked the latest yesterday, Adam fixed many of them. Still there are places like this:
https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/blob/8b87b6cbf765d2b4d6cf66a06cc9004e11c096c4/include/vk_mem_alloc.h#L13777

juan-lunarg: Doesn't moving VMA to known_good.json complicate things for system package managers? AFAIK VMA isn't a toggle-able library like `robin_hood`.

jeremyg-lunarg: Good point.  AFAICT ubuntu, arch and fedora don't provide a package for it.  The SDK does make [packages ](https://packages.lunarg.com/vulkan/1.3.236/pool/main/v/vma/)for it.

layers/generated/best_practices.cpp
```diff
@@ -4,9 +4,9 @@
 
 /***************************************************************************
  *
- * Copyright (c) 2015-2022 The Khronos Group Inc.
- * Copyright (c) 2015-2022 Valve Corporation
- * Copyright (c) 2015-2022 LunarG, Inc.
+ * Copyright (c) 2015-2023 The Khronos Group Inc.
```
juan-lunarg, line 7: Why is this file being updated?

layers/generated/best_practices.h
```diff
@@ -4,9 +4,9 @@
 
 /***************************************************************************
  *
- * Copyright (c) 2015-2022 The Khronos Group Inc.
- * Copyright (c) 2015-2022 Valve Corporation
- * Copyright (c) 2015-2022 LunarG, Inc.
+ * Copyright (c) 2015-2023 The Khronos Group Inc.
+ * Copyright (c) 2015-2023 Valve Corporation
+ * Copyright (c) 2015-2023 LunarG, Inc.
```
juan-lunarg, line 9: Why is this file being updated?

juan-lunarg (changes requested):
Minor nitpicks

layers/best_practices_utils.cpp
```diff
@@ -5130,6 +5130,7 @@ bool BestPractices::PreCallValidateCreatePipelineLayout(VkDevice device, const V
                             break;
                         case VK_DESCRIPTOR_TYPE_INLINE_UNIFORM_BLOCK:
                             descriptor_type_size = 1;
+                            break;
```
ncesario-lunarg, line 5133: Well that was a bug waiting to happen :grimacing:

ncesario-lunarg (commented):
LGTM once the copyright date comments are addressed!

layers/generated/best_practices.h
artem-lunarg: Thanks for noticing this. Going to revert this. At one point there was fail on github CI due to outdated copyright, I guess I had other changes here and then reverted them...

juan-lunarg (approved):

PR #5099, opened 2023-01-12
cmake: Use target_compile_definitions & add_compile_definitions

juan-lunarg:
CMake docs don't recommend using `add_definitions` anymore.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,8 +21,6 @@ project(VVL LANGUAGES CXX C)
 
 option(BUILD_TESTS "Build the tests" OFF)
 
-add_definitions(-DVK_ENABLE_BETA_EXTENSIONS) # Enable beta Vulkan extensions
-
 if (UPDATE_DEPS)
     find_package(PythonInterp 3 REQUIRED)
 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -153,22 +152,22 @@ else()
...
     if(BUILD_WSI_WAYLAND_SUPPORT)
-        add_definitions(-DVK_USE_PLATFORM_WAYLAND_KHR)
+        target_compile_definitions(Vulkan::Headers INTERFACE VK_USE_PLATFORM_WAYLAND_KHR)
     endif()
 endif()
 
 if (ANNOTATED_SPEC_LINK)
     message("ANNOTATED_SPEC_LINK is ${ANNOTATED_SPEC_LINK}")
-    add_definitions(-DANNOTATED_SPEC_LINK=${ANNOTATED_SPEC_LINK})
+    add_compile_definitions(ANNOTATED_SPEC_LINK=${ANNOTATED_SPEC_LINK})
```
ncesario-lunarg, line 170: This will still apply this to everything going forward, even projects outside of VVL, no? Should this be converted to `target_compile_definitions`, or is `add_compile_definitions` the "correct" behavior here?

ncesario-lunarg (approved):
One question, but LGTM!

CMakeLists.txt
juan-lunarg: It's used by `vk_layer_validation_tests` and `VkLayer_khronos_validation` so I wasn't quite sure how to handle it elegantly with `target_compile_definitions` without code duplication.

> even projects outside of VVL, no?

`add_compile_definitions` only affects the current directory and below.

juan-lunarg:
> CI Vulkan-ValidationLayers build # 10203 failed.

Failure looks spurious

> [ RUN      ] VkLayerTest.InvalidBarriers
ERROR: -------------------------------------------------------------
ERROR: FAIL during test after 11.52 minutes (691.16 seconds)

spencer-lunarg (approved):

CMakeLists.txt
ncesario-lunarg: > `add_compile_definitions` only affects the current directory and below.

Ah, good to know!

PR #5100, opened 2023-01-13
build: Set external dir based on build type

ncesario-lunarg:
Set the --dir value passed to update_deps.py via cmake variable (-DUPDATE_DEPS_DIR), or automatically based on build type.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,22 +59,28 @@ if (UPDATE_DEPS)
         set(optional_args ${optional_args} "CMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE}")
     endif()
 
+    set(_update_deps_dir ${CMAKE_CURRENT_LIST_DIR}/external/${_build_type})
+    if (UPDATE_DEPS_DIR AND EXISTS "${UPDATE_DEPS_DIR}")
+        set(_update_deps_dir ${UPDATE_DEPS_DIR})
+    endif()
+    set(_helper_file "${_update_deps_dir}/helper.cmake")
+
     # Add a target so that update_deps.py will run when necessary
     # NOTE: This is triggered off of the timestamps of known_good.json and helper.cmake
     add_custom_command(
-        OUTPUT ${CMAKE_CURRENT_LIST_DIR}/external/helper.cmake
+        OUTPUT ${_helper_file}
         COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_LIST_DIR}/scripts/update_deps.py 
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -59,30 +59,36 @@ if (UPDATE_DEPS)
         set(optional_args ${optional_args} "CMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE}")
     endif()
 
+    set(_update_deps_dir ${CMAKE_CURRENT_LIST_DIR}/external-${_build_type})
```
juan-lunarg, line 62: This will cause problems for GitHub actions caching.

If we want to improve the update_deps.py script it'd be good to mimic how vcpkg handles this scenario. I believe they build debug/release binaries in 1 go. That way it works for Ninja / Visual Studio builds elegantly.

juan-lunarg (changes requested):

CMakeLists.txt
ncesario-lunarg: Ah, yep, good catch!

juan-lunarg, line 62: Keep all installed dependencies in 1 folder.

```suggestion
    set(_update_deps_dir ${CMAKE_CURRENT_LIST_DIR}/external/${_build_type})
```

CMakeLists.txt
```diff
@@ -59,30 +59,36 @@ if (UPDATE_DEPS)
...
+    if (UPDATE_DEPS_DIR AND EXISTS "${UPDATE_DEPS_DIR}")
+        set(_update_deps_dir ${UPDATE_DEPS_DIR})
+    endif()
```
juan-lunarg, line 65: I don't think this code does anything.

```suggestion
```

juan-lunarg (changes requested):
GitHub actions will need to be updated as well

CMakeLists.txt
juan-lunarg: Also removes the need to update the `.gitignore`

juan-lunarg:
This will also require updating multiple CI scripts due to changing the `helper.cmake` invocation.

```
 loading initial cache file ../external/helper.cmake
CMake Error: Error processing file: /home/runner/work/Vulkan-ValidationLayers/Vulkan-ValidationLayers/external/helper.cmake
-- The CXX compiler identification is GNU 9.4.0
```

CMakeLists.txt
ncesario-lunarg: Yep, much better, thanks!

CMakeLists.txt
ncesario-lunarg: This is to allow for setting the "external directory" directly, rather than having it derived from the build type.

juan-lunarg (approved):
LGTM once the merge conflict is resolved

PR #5101, opened 2023-01-13
layers: Fix GCC release build with -Werror

ncesario-lunarg:
Fixes a warning on gcc 12.2 when building release with -Werror.


arno-lunarg (approved):

jeremyg-lunarg (approved):
LGTM. The cloud appears to be sad for unrelated reasons.

Issue #3805, opened 2022-03-05
Invalid dynamic rendering validation...


unit-a-user:
**Describe the Issue**
Despite of error, `pDepthStencilState` and `pColorBlendState` has and valid, or there is no additional validation details... 

**Valid Usage ID**
Look to log...

**Environment:**
 - OS: Windows 11
 - GPU: NVIDIA RTX 2070 
 - SDK or header version if building from repo: 1.3.204.1
 - Options enabled (synchronization, best practices, etc.):

**Additional context**
```
Vulkan Debug Error: Validation Error: [ VUID-VkGraphicsPipelineCreateInfo-renderPass-06053 ] Object 0: handle = 0x1b736479ed0, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x247dedb2 | vkCreateGraphicsPipelines(): Pipeline 0 has fragment state and a depth format (VK_FORMAT_D32_SFLOAT_S8_UINT) or stencil format (VK_FORMAT_D32_SFLOAT_S8_UINT) and an invalid pDepthStencilState structure The Vulkan spec states: If renderPass is VK_NULL_HANDLE, the pipeline is being created with fragment shader state, and either of VkPipelineRenderingCreateInfo::depthAttachmentFormat or VkPipelineRenderingCreateInfo::stencilAttachmentFormat are not VK_FORMAT_UNDEFINED, pDepthStencilState must be a valid pointer to a valid VkPipelineDepthStencilStateCreateInfo structure (https://vulkan.lunarg.com/doc/view/1.3.204.1/windows/1.3-extensions/vkspec.html#VUID-VkGraphicsPipelineCreateInfo-renderPass-06053)
[...]

PR #5102, opened 2023-01-13
layers: Check pipeline state for 06053 and 06054

zzoon:
The spec says:

```
VUID-VkGraphicsPipelineCreateInfo-renderPass-06053
If renderPass is VK_NULL_HANDLE, the pipeline is being created with fragment shader state and fragment output interface state, and either of VkPipelineRenderingCreateInfo::depthAttachmentFormat or VkPipelineRenderingCreateInfo::stencilAttachmentFormat are not VK_FORMAT_UNDEFINED, pDepthStencilState must be....
```
```
VUID-VkGraphicsPipelineCreateInfo-renderPass-06054
If renderPass is VK_NULL_HANDLE, the pipeline is being created with fragment output interface state, and VkPipelineRenderingCreateInfo::colorAttachmentCount is not equal to 0, pColorBlendState must be .....
```

### DIFF
--- a/layers/core_checks/pipeline_validation.cpp
+++ b/layers/core_checks/pipeline_validation.cpp
@@ -2862,7 +2862,8 @@ bool CoreChecks::ValidateGraphicsPipelineDynamicRendering(const PIPELINE_STATE &
         const auto raster_state = pipeline.RasterizationState();
         const bool has_rasterization = raster_state && (raster_state->rasterizerDiscardEnable == VK_FALSE);
         if (has_rasterization) {
-            if (((rendering_struct->depthAttachmentFormat != VK_FORMAT_UNDEFINED) ||
+            if (pipeline.fragment_shader_state && pipeline.fragment_output_state &&
+                ((rendering_struct->depthAttachmentFormat != VK_FORMAT_UNDEFINED) ||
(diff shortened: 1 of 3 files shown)


spencer-lunarg:
This is related to https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/3805

ncesario-lunarg (commented):
This LGTM, but is it possible to get a test for the original issue? This is probably relying on something like `rasterizerDiscardEnabled == VK_TRUE => !fragment_output_state` (or something along those lines), but I'm not sure that's true.

zzoon:
> This LGTM, but is it possible to get a test for the original issue? This is probably relying on something like `rasterizerDiscardEnabled == VK_TRUE => !fragment_output_state` (or something along those lines), but I'm not sure that's true.

This issue can be found with both gpl and dynamic rendering enabled. If I understand correctly, a new test for this can be in `VkGraphicsLibraryLayerTest`.

ncesario-lunarg:
Thanks for the test @zzoon! I went ahead and pushed a change to your branch that split the test code and layer code into separate commits and ran clang-format (which is required).

ncesario-lunarg (approved):

PR #5104, opened 2023-01-13
scripts: Remove false positive for sType

spencer-lunarg:
Currently tests are showing false positive

```
[ RUN      ] VkLayerTest.DescriptorUpdateTemplate

Validation Error: [ VUID-VkDescriptorSetLayoutCreateInfo-pNext-pNext ] Object 0: handle = 0x55cc55a46570, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0x2b747b43 | vkCreateDescriptorSetLayout: Includes a pNext pointer (pCreateInfo->pNext) to a VkStructureType (VK_STRUCTURE_TYPE_MUTABLE_DESCRIPTOR_TYPE_CREATE_INFO_EXT), but its parent extension VK_VALVE_mutable_descriptor_type has not been enabled. The Vulkan spec states: Each pNext member of any structure (including this one) in the pNext chain must be either NULL or a pointer to a valid instance of VkDescriptorSetLayoutBindingFlagsCreateInfo or VkMutableDescriptorTypeCreateInfoEXT (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkDescriptorSetLayoutCreateInfo-pNext-pNext)
```

### DIFF
--- a/scripts/parameter_validation_generator.py
+++ b/scripts/parameter_validation_generator.py
@@ -1,9 +1,9 @@
 #!/usr/bin/python3 -i
 #
-# Copyright (c) 2015-2022 The Khronos Group Inc.
-# Copyright (c) 2015-2022 Valve Corporation
-# Copyright (c) 2015-2022 LunarG, Inc.
-# Copyright (c) 2015-2022 Google Inc.
+# Copyright (c) 2015-2023 The Khronos Group Inc.
+# Copyright (c) 2015-2023 Valve Corporation
+# Copyright (c) 2015-2023 LunarG, Inc.
+# Copyright (c) 2015-2023 Google Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


jeremyg-lunarg:
generate_source.py needs to be run to update the SPIRV-Tools commit id.

scripts/parameter_validation_generator.py
```diff
@@ -436,13 +436,20 @@ def beginFile(self, genOpts):
             for enum_item in node.iter('enum'):
                 if enum_item.get('extends') == "VkStructureType":
                     struct_type_id = enum_item.get('name')
+                    # This caputres all sType in core (or promoted)
```
ncesario-lunarg, line 439: `caputres` -> `captures`

ncesario-lunarg (approved):
I think this looks good, but TBH the promotion code is rather confusing to me. I'll try to see if this can be utilized in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4228 somehow.

Issue #4891, opened 2022-12-01
Add VUID-vkCmdDraw-None-07748


spencer-lunarg:
**Describe the situation in which you encountered the missing validation**

This was added as `VUID-vkCmdDraw-None-02686`, then an [issue of it being wrong](https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/2243) was created and it was then [removed](https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/2244). Recently was [added again](https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4519) but an [internal change](https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5543) to solve https://github.com/KhronosGroup/Vulkan-Docs/issues/1979

**Valid Usage IDs requested**

`VUID-vkCmdDraw-None-07748`

**Additional Context**

N/A

Issue #4604, opened 2022-09-29
Add a regression test for null descriptor_set pointer in CMD_BUFFER_STATE::lastBound::per_set


jeremyg-lunarg:
We need test cases to cover https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4597 and  https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4576

Issue #4597, opened 2022-09-27
ANGLE segfaults in BestPractices::ValidateBoundDescriptorSets


romanl-g:
```
std::Cr::vector<std::Cr::unique_ptr<cvdescriptorset::DescriptorBinding, cvdescriptorset::DescriptorSet::BindingDeleter>, std::Cr::allocator<std::Cr::unique_ptr<cvdescriptorset::DescriptorBinding, cvdescriptorset::DescriptorSet::BindingDeleter> > >::begin[abi:v160000]() at vector:1374
cvdescriptorset::DescriptorSet::begin() at descriptor_sets.h:890
BestPractices::ValidateBoundDescriptorSets(bp_state::CommandBuffer&, VkPipelineBindPoint, char const*) at best_practices_utils.cpp:3780
BestPractices::PreCallRecordCmdDraw(VkCommandBuffer_T*, unsigned int, unsigned int, unsigned int, unsigned int) at best_practices_utils.cpp:3827
vulkan_layer_chassis::CmdDraw(VkCommandBuffer_T*, unsigned int, unsigned int, unsigned int, unsigned int) at chassis.cpp:3037
rx::vk::priv::SecondaryCommandBuffer::executeCommands(rx::vk::priv::CommandBuffer*) at SecondaryCommandBuffer.cpp:391
```

Blocking ANGLE roll: https://ci.chromium.org/ui/p/angle/builders/try/linux-test/8805/overview

Appears to be introduced here (@paul-lunarg @jeremyg-lunarg):
https://chromium.googlesource.com/external/github.com/KhronosGroup/Vulkan-ValidationLayers/+/5b86357111eedb8bcb700afeed84f65aa4ac6ab5

`*descriptor_set.bound_descriptor_set` wasn't accessed before this change - could it be nullptr or invalid?
Thanks!

romanl-g:
Local repro in ANGLE checkout:

```
autoninja -C out/Debug angle_deqp_khr_gles31_tests && out/Debug/angle_deqp_khr_gles31_tests --gtest_filter='KHR-GLES31.core.texture_storage_multisample.FunctionalTests.blitting_multisampled_depth_attachment'
```

romanl-g:
Hi, just checking if this got lost in the shuffle?

paul-lunarg:
Hi, no, I will get to this ASAP

romanl-g:
[...]

PR #5116, opened 2023-01-16
tests: Add regression test for null set crash in best practices

artem-lunarg:
This PR adds regression test for the first issue (https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4597) mentioned in: https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4604

I'm going to investigate the second issue (https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4576) separately. That section of code was reverted at some point, but it still looks it's a good idea to add regression test because something similar might be implemented again (for example, as part of https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4891).


tests/vklayertests_best_practices.cpp
```diff
@@ -2347,3 +2347,40 @@ TEST_F(VkBestPracticesLayerTest, ImageMemoryBarrierAccessLayoutCombinations) {
         m_commandBuffer->end();
     }
 }
+
+// This is a positive test. No failures are expected.
+TEST_F(VkBestPracticesLayerTest, DrawingWithUnboundUnusedSet) {
```
spencer-lunarg, line 2352: I would just make a `tests/positive/best_practice.cpp` and move this there, I assume we will have more like this in the future

artem-lunarg: Yes, actually it's the second positive test in this file.

spencer-lunarg (approved):

tests/positive/best_practices.cpp
```diff
@@ -0,0 +1,152 @@
...
+TEST_F(VkPositiveBestPracticesLayerTest, TestDestroyFreeNullHandles) {
+    VkResult err;
+
+    TEST_DESCRIPTION("Call all applicable destroy and free routines with NULL handles, expecting no validation errors");
+
+    InitBestPracticesFramework();
+    InitState();
+
+    ASSERT_NO_FATAL_FAILURE(InitViewport());
+    ASSERT_NO_FATAL_FAILURE(InitRenderTarget());
```
arno-lunarg, line 40: Nit (to match your new code, which I think should be the standard)
```suggestion
    ASSERT_NO_FATAL_FAILURE(InitBestPracticesFramework());
    ASSERT_NO_FATAL_FAILURE(InitState());
    ASSERT_NO_FATAL_FAILURE(InitViewport());
    ASSERT_NO_FATAL_FAILURE(InitRenderTarget());
```

arno-lunarg (commented):
LGTM!

arno-lunarg (approved):

tests/positive/best_practices.cpp
artem-lunarg: Updated.

PR #5117, opened 2023-01-16
ci: Test CMake min with GitHub action

juan-lunarg:
Removes need for custom cmake scripts


juan-lunarg:
`get-cmake` also caches the CMake version specified so it also helps out the CI a bit.

juan-lunarg:
Verified CI is using CMake 3.17.2

```
Run python3 scripts/github_ci_win_linux.py --config release
cmake version 3.17.2

CMake suite maintained and supported by Kitware (kitware.com/cmake).
-- The CXX compiler identification is GNU 9.4.0
-- The C compiler identification is GNU 9.4.0
-- Check for working CXX compiler: /usr/bin/g++
```

ncesario-lunarg (approved):
IIRC, the reason for this script's existence was due to needing a newer cmake to work with newer versions of VS, but there was no need for newer features and so ultimately keeping the 3.10.2 minimum. I'm pretty sure that is no longer a requirement, so I'm happy to say "good riddance" to these :)

PR #5119, opened 2023-01-17
Make separate Interface Variables 

spencer-lunarg:
Split `InterfaceVariables` into both `UserDefinedInterfaceVariable` and `ResourceInterfaceVariable` as they are used differently. Moved more logic into `ResourceInterfaceVariable` constructor which is done at `Entrypoint` creation time now once

### DIFF
--- a/layers/shader_module.h
+++ b/layers/shader_module.h
@@ -68,18 +68,38 @@ struct DecorationSet {
     bool Has(FlagBit flag_bit) const { return (flags & flag_bit) != 0; }
 };
 
-// Information about a OpVariable used as an interface in the shader
-struct InterfaceVariable {
+// vkspec.html#interfaces-iointerfaces-user describes 'User-defined Variable Interface'
+// These are Input/Output OpVariable that go in-between stages
+// (also for example the input to a Vertex and output of the Fragment).
+// These are always ints/floats (not images or samplers).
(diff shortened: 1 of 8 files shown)


layers/shader_module.h
```diff
@@ -68,26 +68,45 @@ struct DecorationSet {
...
+
+// vkspec.html#interfaces-resources describes 'Shader Resource Interface'
+// These are the OpVariable attached to descriptors.
+// The slots are known at Pipeline creation time, but the type images/sampler/etc is
+// not known until the descriptors are bound.
+// The main purpose of this struct is to track what operations are statically done so
+// at draw/submit time we can cross reference with the last bound descriptor.
+struct ResourceInterfaceVariable {
+    uint32_t id;
+    uint32_t type_id;
```
ncesario-lunarg, line 98: Wondering if it would make sense to make a common "Instruction base class," but at least at this point it probably doesn't make sense.

ncesario-lunarg (approved):
LGTM!

layers/shader_module.h
spencer-lunarg: I looked into this, was going to see first if the uses of `id`/`type_id` here need to be a `ResourceInterfaceVariable` or if it could just be a `Instruction`

Issue #4860, opened 2022-11-24
Missing Ray Tracing parameter validation 


spencer-lunarg:
**Describe the situation in which you encountered the missing validation**
Some low hanging fruit to add that is missing

**Valid Usage IDs requested**

- [ ] VUID-VkAccelerationStructureGeometryAabbsDataKHR-stride-03545
- [ ] VUID-VkAccelerationStructureGeometryAabbsDataKHR-stride-03820
- [ ] VUID-VkAccelerationStructureBuildGeometryInfoKHR-type-03794
- [ ] VUID-VkAccelerationStructureBuildGeometryInfoKHR-type-03795
- [ ] VUID-VkRayTracingPipelineCreateInfoKHR-flags-07403
- [ ] VUID-VkRayTracingPipelineCreateInfoKHR-stage-06899

PR #5120, opened 2023-01-17
Add 6 RT VUIDs

arno-lunarg:
Add VUIDs mentioned in https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4860
` VUID-VkAccelerationStructureGeometryAabbsDataKHR-stride-03820` was already checked, but was lacking a test.

### DIFF
--- a/layers/parameter_validation_utils.cpp
+++ b/layers/parameter_validation_utils.cpp
@@ -1,7 +1,7 @@
-/* Copyright (c) 2015-2022 The Khronos Group Inc.
- * Copyright (c) 2015-2022 Valve Corporation
- * Copyright (c) 2015-2022 LunarG, Inc.
- * Copyright (C) 2015-2022 Google Inc.
+/* Copyright (c) 2015-2023 The Khronos Group Inc.
+ * Copyright (c) 2015-2023 Valve Corporation
+ * Copyright (c) 2015-2023 LunarG, Inc.
+ * Copyright (C) 2015-2023 Google Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


layers/parameter_validation_utils.cpp
```diff
@@ -8331,7 +8346,8 @@ bool StatelessValidation::manual_PreCallValidateCmdBindVertexBuffers2(VkCommandB
 }
 
 bool StatelessValidation::ValidateAccelerationStructureBuildGeometryInfoKHR(
-    const VkAccelerationStructureBuildGeometryInfoKHR *pInfos, uint32_t infoCount, const char *api_name) const {
+    const VkAccelerationStructureBuildGeometryInfoKHR *pInfos, uint32_t infoCount, std::optional<uint64_t> total_primitive_count,
```
spencer-lunarg, line 8349: what is the advantage of using `std::optional<uint64_t>` instead of just `uint64_t` since if it is `zero` it will never be larger than `maxPrimitiveCount`

layers/parameter_validation_utils.cpp
```diff
@@ -8412,6 +8428,19 @@ bool StatelessValidation::ValidateAccelerationStructureBuildGeometryInfoKHR(
                                      "(%s):indexType must be VK_INDEX_TYPE_UINT16, VK_INDEX_TYPE_UINT32, or VK_INDEX_TYPE_NONE_KHR",
                                      api_name);
                     }
+                    // Only check once
+                    if (j == 0) {
```
spencer-lunarg, line 8432: why not just move outside the `for` loop since it already checks `pInfos[i].pGeometries` is not null there

layers/parameter_validation_utils.cpp
```diff
@@ -8453,24 +8482,42 @@ bool StatelessValidation::ValidateAccelerationStructureBuildGeometryInfoKHR(
                         ParameterName("pInfos[%i].pGeometries[%i].geometry.aabbs.pNext", ParameterName::IndexVector{i, j}), NULL,
                         pInfos[i].pGeometries[j].geometry.aabbs.pNext, 0, NULL, GeneratedVulkanHeaderVersion,
                         "VUID-VkAccelerationStructureGeometryAabbsDataKHR-pNext-pNext", kVUIDUndefined);
+                    if (pInfos[i].pGeometries[j].geometry.aabbs.stride % 8) {
```
spencer-lunarg, line 8485: we have a `SafeModulo` to use here

tests/vklayertests_ray_tracing_pipeline.cpp
```diff
@@ -219,18 +219,25 @@ TEST_F(VkLayerTest, RayTracingPipelineShaderGroupsKHR) {
 
     const VkPipelineLayoutObj empty_pipeline_layout(m_device, {});
 
-    const char *empty_shader = R"glsl(
+    const char *empty_rt_shader = R"glsl(
```
spencer-lunarg, line 222: we have a `bindStateMinimalShaderText` in `layer_validation_tests.h`, I could see this being used more and might be worth making a a "minimal ray tracing shader" constant there

layers/parameter_validation_utils.cpp
arno-lunarg: I think that 0 is a legal value for `maxPrimitiveCount`?

spencer-lunarg: > the total number of triangles in all geometries must be less than or equal to [VkPhysicalDeviceAccelerationStructurePropertiesKHR](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkPhysicalDeviceAccelerationStructurePropertiesKHR)::maxPrimitiveCount

so if they are both `0` this won't error out anyway

layers/parameter_validation_utils.cpp
arno-lunarg: I wanted to avoid writing the following if again:
```cpp
if (pInfos[i].pGeometries[j].geometryType == VK_GEOMETRY_TYPE_TRIANGLES_KHR) 
```
again ^^"

layers/parameter_validation_utils.cpp
arno-lunarg: Is it needed ? The divisor is 8, so no need to check it?

tests/positive/ray_tracing_pipeline.cpp
```diff
@@ -1,8 +1,8 @@
 /*
- * Copyright (c) 2015-2022 The Khronos Group Inc.
- * Copyright (c) 2015-2022 Valve Corporation
- * Copyright (c) 2015-2022 LunarG, Inc.
- * Copyright (c) 2015-2022 Google, Inc.
+ * Copyright (c) 2015-2023 The Khronos Group Inc.
```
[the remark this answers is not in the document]
arno-lunarg: I went on a copyright change spree and this file got swallowed x)

layers/parameter_validation_utils.cpp
spencer-lunarg: ya, good point, we do this the same else where too

layers/parameter_validation_utils.cpp
arno-lunarg: Oh yes it is less or equal, my check is wrong then O_O

tests/layer_validation_tests.cpp
```diff
@@ -2573,16 +2602,7 @@ void VkLayerTest::OOBRayTracingShadersTestBody(bool gpu_assisted) {
...
-        float min_x;
-        float min_y;
-        float min_z;
-        float max_x;
-        float max_y;
-        float max_z;
-    };
-
-    const std::vector<AABB> aabbs = {{-1.0f, -1.0f, -1.0f, +1.0f, +1.0f, +1.0f}};
+    const std::array<VkAabbPositionsKHR, 1> aabbs = {{{-1.0f, -1.0f, -1.0f, +1.0f, +1.0f, +1.0f}}};
```
spencer-lunarg, line 2605: constexpr

layers/parameter_validation_utils.cpp
arno-lunarg: I changed this logic, to I think what you advised? It avoids VUID duplication too

spencer-lunarg: think the new logic looks really clean! thanks for re-working it

layers/parameter_validation_utils.cpp
```diff
@@ -8361,6 +8375,33 @@ bool StatelessValidation::ValidateAccelerationStructureBuildGeometryInfoKHR(
...
+
+        const VkAccelerationStructureGeometryKHR *pGeometry = pInfos[i].pGeometries    ? pInfos[i].pGeometries
+                                                              : pInfos[i].ppGeometries ? pInfos[i].ppGeometries[0]
+                                                                                       : nullptr;
+        if (pGeometry) {
+            if (total_primitive_count > phys_dev_ext_props.acc_structure_props.maxPrimitiveCount) {
+                switch (pGeometry->geometryType) {
+                    case VK_GEOMETRY_TYPE_TRIANGLES_KHR:
+                        skip |= LogError(device, "VUID-VkAccelerationStructureBuildGeometryInfoKHR-type-03795",
+                                         "(%s): total number of triangles in all geometries (%" PRIu64
```
spencer-lunarg, line 8387: (realized there are a few other places this is wrong from before the PR)

```patch
- (%s):
+ %s():
```

spencer-lunarg (approved):
LGTM, nice to get these VUs in with proper tests!

layers/parameter_validation_utils.cpp
arno-lunarg: oh yes I forgot to change that, will do with a fixup

PR #5122, opened 2023-01-17
macos: Target 10.15

juan-lunarg:


ncesario-lunarg (approved):

layers/image_state.cpp
```diff
@@ -832,8 +832,7 @@ void SURFACE_STATE::SetPresentModeCapabilities(VkPhysicalDevice phys_dev, const
     if (!present_modes_data_[phys_dev][present_mode].has_value()) {
         present_modes_data_[phys_dev][present_mode] = std::make_shared<PresentModeState>();
     }
-    // Old MacOS doesn't support Value()
```
juan-lunarg, line 835: FYI to @mark-lunarg since you ran into this.

richard-lunarg (commented):
Looks good!

juan-lunarg:
SDK script has been updated to use 10.15, so there should be no issues when making the  SDK.

Issue #5115, opened 2023-01-16
"Ensure object handle is not null" fails on vkCreateDebugUtilsMessengerEXT returning null


romanl-g:
**Describe the Issue**
This commit causing assertion fail
https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/e4d0a215586dcc42fba6ce54e51b2cc5d507849a

in ANGLE with VVL enabled:
```
target_os = "android"
target_cpu = "arm64"
is_component_build = false
is_debug = true
```

Driver doesn't expose `VK_EXT_debug_utils`, VVL adds it so ANGLE calls `vkCreateDebugUtilsMessengerEXT`

`third_party/vulkan-deps/vulkan-validation-layers/src/layers/generated/layer_chassis_dispatch.cpp`
calls driver's `vkCreateDebugUtilsMessengerEXT`
`VkResult result = layer_data->instance_dispatch_table.CreateDebugUtilsMessengerEXT(instance, pCreateInfo, pAllocator, pMessenger);`

which rightfully returns VK_NULL_HANDLE and assertion fails

**Environment:**
 - OS: Android
 - GPU:
 - SDK or header version if building from repo:
 - Options enabled (synchronization, best practices, etc.):

**Additional context**

romanl-g:
@artem-lunarg

artem-lunarg:
@romanl-g I'm working on the solution.

In addition to assert fix one more change in behavior can happen. In the previous implementation, when VK_NULL_HANDLE was returned from `CreateDebugUtilsMessengerEXT`, it was wrapped with `WrapNew`. `CreateDebugUtilsMessengerEXT` was returning non-null value even if the object was not created. I'm going to change that behavior, so `*pMessenger` will stay VK_NULL_HANDLE. If there's evidence that ANGLE relied on `*pMessenger` being not null, please let us know.

romanl-g:
Driver doesn't have VK_EXT_debug_utils. Wouldn't that mean that VVL shouldn't be calling CreateDebugUtilsMessengerEXT at all regardless of whether that would fail or return null?

romanl-g:
Despite driver not having VK_EXT_debug_utils, isn't VK_EXT_debug_utils provided by VVL itself? We're expecting this to keep working the same way as it was working before the assert was added.

artem-lunarg:
[...]

PR #5123, opened 2023-01-17
layers: Add special case for debug utils on Android

artem-lunarg:
It's a workaround for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5115

### DIFF
--- a/scripts/layer_chassis_dispatch_generator.py
+++ b/scripts/layer_chassis_dispatch_generator.py
@@ -1,9 +1,9 @@
 #!/usr/bin/python3 -i
 #
-# Copyright (c) 2015-2022 The Khronos Group Inc.
-# Copyright (c) 2015-2022 Valve Corporation
-# Copyright (c) 2015-2022 LunarG, Inc.
-# Copyright (c) 2015-2022 Google Inc.
+# Copyright (c) 2015-2023 The Khronos Group Inc.
+# Copyright (c) 2015-2023 Valve Corporation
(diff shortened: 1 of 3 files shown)


mark-lunarg (approved):
LGTM!

scripts/layer_chassis_dispatch_generator.py
```diff
@@ -2015,6 +2015,10 @@ def generate_create_ndo_code(self, indent, proto, params, cmd_info):
                 if is_create_pipelines:
                     create_ndo_code += '%sif (%s != VK_NULL_HANDLE) {\n' % (indent, ndo_dest)
                     indent = self.incIndent(indent)
+            if 'CreateDebugUtilsMessengerEXT' in proto.text:
+                create_ndo_code += '#ifdef VK_USE_PLATFORM_ANDROID_KHR\n'
+                create_ndo_code += '%sif (%s)\n' % (indent, ndo_dest)
+                create_ndo_code += '#endif\n'
```
ncesario-lunarg, line 2021: I realize this is generated code, but would it be possible to add `{`/`}` and indent the line below? Makes it a bit confusing when debugging given the current output.

ncesario-lunarg (commented):
LGTM, though admittedly I do not fully understand the original issue given that VVL implements this extension. If this is correct, should we be doing this on all platforms? Does it make sense to return `VK_NULL_HANDLE` in general (i.e., not just for `CreateDebugUtilsMessengerEXT`)?

PR #5125, opened 2023-01-18
tests: Fix descriptor buffer tests

ncesario-lunarg:
Also fix a compilation error seen on VS 2022 + release + -DBUILD_WERROR=ON.

NOTE: some of these fixes are for #4228.

### DIFF
--- a/tests/vklayertests_descriptor_renderpass_framebuffer.cpp
+++ b/tests/vklayertests_descriptor_renderpass_framebuffer.cpp
@@ -12202,7 +12202,7 @@ TEST_F(VkLayerTest, DescriptorBufferSetLayout) {
     }
     {
         VkSampler samplers[2] = {sampler.handle(), sampler.handle()};
-        const VkDescriptorSetLayoutBinding binding{0, VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE, 1, VK_SHADER_STAGE_FRAGMENT_BIT, samplers};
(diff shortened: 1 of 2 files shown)


tests/vklayertests_descriptor_renderpass_framebuffer.cpp
```diff
@@ -13149,6 +13152,9 @@ TEST_F(VkLayerTest, DescriptorBufferDescriptorGetInfo) {
         GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported";
     }
 
+    const auto acceleration_structure = IsExtensionsEnabled(VK_KHR_ACCELERATION_STRUCTURE_EXTENSION_NAME);
```
spencer-lunarg, line 13155: (warning, `auto` discussion)

`const bool` to match the reset of the uses of `IsExtensionsEnabled`... and because `bool` is a keyword 😃

spencer-lunarg (approved):

tests/vklayertests_descriptor_renderpass_framebuffer.cpp
ncesario-lunarg: Hmm, I would argue that I really only care whether or not the result of `IsExtensionsEnabled` is comparable to bool (e.g., could be an object with an `operator bool`). But that's not a hill I'm willing to die on; will change to `bool` ;)

PR #5126, opened 2023-01-18
layers: Move files into dedicated folders

spencer-lunarg:
Takes all files and move them into a more structured folder scheme. In the process some notable modifications were made:
- synchronization_* renamed to sync_* to match other files
- drawdispatch.cpp -> drawdispatch_validation.cpp
- clang-format was ran on some files not ran in a while


CMakeLists.txt
```diff
@@ -63,7 +63,7 @@ if (UPDATE_DEPS)
     # NOTE: This is triggered off of the timestamps of known_good.json and helper.cmake
     add_custom_command(
         OUTPUT ${CMAKE_CURRENT_LIST_DIR}/external/helper.cmake
-        COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_LIST_DIR}/scripts/update_deps.py 
+        COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_LIST_DIR}/scripts/update_deps.py
```
juan-lunarg, line 66: Revert change.

layers/CMakeLists.txt
```diff
@@ -272,6 +275,15 @@ else()
     target_link_options(VkLayer_khronos_validation PRIVATE LINKER:--version-script=${CMAKE_CURRENT_SOURCE_DIR}/libVkLayer_khronos_validation.map,-Bsymbolic,--exclude-libs,ALL)
 endif()
 
+target_include_directories(VkLayer_khronos_validation PUBLIC
```
juan-lunarg, line 278: I disagree with adding these include directories, projects should strive to minimize their include directories as much as possible.

It also forces users to specify the directory, which I'd argue is a good thing. Since it highlights the project structure.

EX:

```
#include "best_practices/best_practices_validation.h"
```

CMakeLists.txt
spencer-lunarg: fixed on rebase

layers/CMakeLists.txt
```diff
@@ -57,7 +57,6 @@ target_link_libraries(VkLayer_utils PUBLIC Vulkan::Headers)
 set_target_properties(VkLayer_utils PROPERTIES LINKER_LANGUAGE CXX)
 target_include_directories(VkLayer_utils PUBLIC
     .
-    generated
```
juan-lunarg, line 60: We need to keep the generated include directory. Otherwise `VkLayer_utils` won't install properly.

juan-lunarg (changes requested):

layers/CMakeLists.txt
spencer-lunarg: ok, will revert... with a comment 😆

juan-lunarg (approved):

PR #5127, opened 2023-01-18
layers: Do not wrap/unwrap VK_NULL_HANDLE

artem-lunarg:
This should help with https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5115.

Put the checks inside WrapNew/Unwrap to minimize code changes (even if generated). There are over 1000 usages of Unwrap. Far fewer use cases of WrapNew, but it was changed to be symmetrical to Unwrap.

### DIFF
--- a/layers/generated/chassis.h
+++ b/layers/generated/chassis.h
@@ -4077,6 +4077,7 @@ class ValidationObject {
         // Unwrap a handle.
         template <typename HandleType>
         HandleType Unwrap(HandleType wrappedHandle) {
+            if (wrappedHandle == (HandleType)VK_NULL_HANDLE) return wrappedHandle;
             auto iter = unique_id_mapping.find(reinterpret_cast<uint64_t const &>(wrappedHandle));
             if (iter == unique_id_mapping.end())
                 return (HandleType)0;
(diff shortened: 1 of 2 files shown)


layers/generated/chassis.h
```diff
@@ -4086,11 +4087,10 @@ class ValidationObject {
         // Wrap a newly created handle with a new unique ID, and return the new ID.
         template <typename HandleType>
         HandleType WrapNew(HandleType newlyCreatedHandle) {
-            // The assert is a safeguard against potential bugs in other layers or the loader.
-            // According to specification, a successfully created non-dispatchable object can not be VK_NULL_HANDLE.
-            assert(newlyCreatedHandle != VK_NULL_HANDLE);
+            if (newlyCreatedHandle == (HandleType)VK_NULL_HANDLE) return newlyCreatedHandle;
             auto unique_id = global_unique_id++;
             unique_id = HashedUint64::hash(unique_id);
+            assert(unique_id != 0); // can't be 0, otherwise unwrap will apply special rule for VK_NULL_HANDLE
```
artem-lunarg, line 4093: It is true for current implementation until we reach 40-bit counter values. After that point depending on the hash function there could be a) non-zero and low probability to get 0 or b) zero probability to get 0.

arno-lunarg (approved):
LGTM!

layers/generated/chassis.h
artem-lunarg, line 4093: It is true in our implementation until we wrap around uint64_t.

This is more of a theoretical interest but for completeness. After uint64_t wrap around, the assert will be triggered with current robin hood hash implementation, which computes 0 from 0 input.

scripts/layer_chassis_generator.py
```diff
@@ -499,6 +499,7 @@ class ValidationObject {
         // Unwrap a handle.
         template <typename HandleType>
         HandleType Unwrap(HandleType wrappedHandle) {
+            if (wrappedHandle == (HandleType)VK_NULL_HANDLE) return wrappedHandle;
```
ncesario-lunarg, line 502: Using `CastToHandle` might make sense here (or one of the helpers in `cast_utils.h`).

mark-lunarg (approved):
I think this looks fine, and certainly addresses my worries. Hopefully the next couple of weeks won't bring any unintended consequences!

scripts/layer_chassis_generator.py
artem-lunarg: CastToHandle in Unwrap results in compilation error on Win32 and only with VkDisplayKHR handles. Will be investigated separately.

PR #5128, opened 2023-01-18
Log VUIDs using a string_view

arno-lunarg:
Most (if not all) VUIDs are defined as string literals, so this PR should prevent a bunch of string constructions.

### DIFF
--- a/layers/error_message/logging.cpp
+++ b/layers/error_message/logging.cpp
@@ -317,8 +317,8 @@ VKAPI_ATTR void DeactivateInstanceDebugCallbacks(debug_report_data *debug_data) 
 
 // helper for VUID based filtering. This needs to be separate so it can be called before incurring
 // the cost of sprintf()-ing the err_msg needed by LogMsgLocked().
-static bool LogMsgEnabled(const debug_report_data *debug_data, const std::string &vuid_text,
-                                 VkDebugUtilsMessageSeverityFlagsEXT severity, VkDebugUtilsMessageTypeFlagsEXT type) {
+static bool LogMsgEnabled(const debug_report_data *debug_data, std::string_view vuid_text,
+                          VkDebugUtilsMessageSeverityFlagsEXT severity, VkDebugUtilsMessageTypeFlagsEXT type) {
     if (!(debug_data->active_severities & severity) || !(debug_data->active_types & type)) {
         return false;
     }
@@ -335,7 +335,10 @@ static bool LogMsgEnabled(const debug_report_data *debug_data, const std::string
     return true;
 }
 
-VKAPI_ATTR bool LogMsg(const debug_report_data *debug_data, VkFlags msg_flags, const LogObjectList &objects, const std::string &vuid_text, const char *format, va_list argptr) {
+VKAPI_ATTR bool LogMsg(const debug_report_data *debug_data, VkFlags msg_flags, const LogObjectList &objects,
+                       std::string_view vuid_text, const char *format, va_list argptr) {
+    assert(*(vuid_text.data() + vuid_text.size()) == '\0');
+
     VkDebugUtilsMessageSeverityFlagsEXT severity;
     VkDebugUtilsMessageTypeFlagsEXT type;
 
@@ -387,7 +390,7 @@ VKAPI_ATTR bool LogMsg(const debug_report_data *debug_data, VkFlags msg_flags, c
         const char *spec_text = nullptr;
         std::string spec_type;
         for (uint32_t i = 0; i < num_vuids; i++) {
-            if (0 == strcmp(vuid_text.c_str(), vuid_spec_text[i].vuid)) {
+            if (0 == strncmp(vuid_text.data(), vuid_spec_text[i].vuid, vuid_text.size())) {
                 spec_text = vuid_spec_text[i].spec_text;
                 spec_type = vuid_spec_text[i].url_id;
                 break;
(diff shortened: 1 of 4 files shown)


layers/vk_layer_logging.cpp
```diff
@@ -398,7 +399,7 @@ VKAPI_ATTR bool LogMsg(const debug_report_data *debug_data, VkFlags msg_flags, c
         const char *spec_text = nullptr;
         std::string spec_type;
         for (uint32_t i = 0; i < num_vuids; i++) {
-            if (0 == strcmp(vuid_text.c_str(), vuid_spec_text[i].vuid)) {
+            if (0 == strcmp(vuid_text.data(), vuid_spec_text[i].vuid)) {
```
artem-lunarg, line 402: Should we use strncmp or any other comparison alternative that allows to specify the number of characters? It's mostly about if we have a guarantee that string view is null terminated or does not point to a longer string.

arno-lunarg: yep `strncmp` should be better

layers/vk_layer_logging.cpp
```diff
@@ -442,7 +443,7 @@ VKAPI_ATTR bool LogMsg(const debug_report_data *debug_data, VkFlags msg_flags, c
         }
     }
 
-    return debug_log_msg(debug_data, msg_flags, objects, "Validation", str_plus_spec_text.c_str(), vuid_text.c_str());
+    return debug_log_msg(debug_data, msg_flags, objects, "Validation", str_plus_spec_text.c_str(), vuid_text.data());
```
artem-lunarg, line 446: Hm.. Here too.

`debug_log_msg` computes hash and uses strlen.

Also `const char*` is passed to `VkDebugUtilsMessengerCallbackDataEXT` where there is no way to control string length.

It's good to have a convention (did we have discussions about this?) how to use string_view in a safe way. One option is not to use string_view when it ends up in the interfaces that assumes null terminated string. Or we can have a convention that for VUIDs we don't create string_view from substrings, ideally enforced in some way.

jeremyg-lunarg: I'm pretty sure this is the first time we'll be using std::string_view in the validation code

juan-lunarg: For spec reference:

> Unlike [std::basic_string::data()](https://en.cppreference.com/w/cpp/string/basic_string/data) and string literals, std::basic_string_view::data() returns a pointer to a buffer that is not necessarily null-terminated, for example a substring view (e.g. from [remove_suffix](https://en.cppreference.com/w/cpp/string/basic_string_view/remove_suffix)). Therefore, it is typically a mistake to pass data() to a routine that takes just a const CharT* and expects a null-terminated string.

I understand the concern now. But it should be fine for this function since nullptr isn't a valid argument for vuid_text in the first place.

arno-lunarg: I think those are strong arguments to use `const char*` instead of `string_view`. I will do the change unless someone has a counter argument!

layers/vk_layer_logging.cpp
ncesario-lunarg: Is there really no `operator==` overload that takes a string_view and a `const char*`? If there is, maybe we should use that?

layers/vk_layer_logging.cpp
ncesario-lunarg: I wonder if something like `using Vuid = const char*;` would make sense. In any case, I yes, I think `const char*` probably makes the most sense.

layers/vk_layer_logging.cpp
jeremyg-lunarg: Another option would be to add the hash of the vuid text into the vuid_spec_text[] table and then use hashes here

jeremyg-lunarg: Or turn vuid_spec_text into a map or unordered_map. There are now over 14,000 entries in this table and linear search is not great for that many.

layers/vk_layer_logging.cpp
arno-lunarg: What would be the best spot to put this `using Vuid = const char *`?

ncesario-lunarg: Maybe `vk_validation_error_messages.h` (which is generated)?

Using a type alias to refer to VUID strings might be a pretty pervasive change. Maybe stick with `const char*` for now and see if that gets a perf bump in the "error case"?

jeremyg-lunarg: I'd say stick with const char * for now and skip the `using` alias

arno-lunarg: Since there is no implicit conversion from `std::string` to `const char*`, using `const char *` is a way more painful change, I suggest sticking with `std::string_view`

spencer-lunarg (approved):

layers/error_message/logging.cpp
```diff
@@ -317,8 +317,8 @@ VKAPI_ATTR void DeactivateInstanceDebugCallbacks(debug_report_data *debug_data)
 
 // helper for VUID based filtering. This needs to be separate so it can be called before incurring
 // the cost of sprintf()-ing the err_msg needed by LogMsgLocked().
-static bool LogMsgEnabled(const debug_report_data *debug_data, const std::string &vuid_text,
-                                 VkDebugUtilsMessageSeverityFlagsEXT severity, VkDebugUtilsMessageTypeFlagsEXT type) {
+static bool LogMsgEnabled(const debug_report_data *debug_data, std::string_view vuid_text,
```
ncesario-lunarg, line 320: While I guess inconsistent with other code in the repo, I think the `const` could be left (though not the `&`).

ncesario-lunarg (approved):
LGTM!

layers/error_message/logging.cpp
arno-lunarg: I am not opposed to it, but what is the benefit?

jeremyg-lunarg (approved):

PR #5138, opened 2023-01-20
git: Add git-blame-ignore-revs file

spencer-lunarg:
git 2.23 added a `--ignore-revs-file <file>` option
GitHub picks up this file name by default

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,98 @@
+# Move Device CoreCheck into own file
+4d196a6c74b7ec9a6f805b7b2959c827a0628cd8
+
+# Move ReleaseProfilingLock in query_validation.cpp
+61314fb36166f07e4f903969721135cbb606a670
+
+# Move External object CoreCheck into own file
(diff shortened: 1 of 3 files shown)


.git-blame-ignore-revs
```diff
@@ -0,0 +1,98 @@
+# Move Device CoreCheck into own file
```
artem-lunarg, line 1: I can find some entries by the message but not by hash. For example, for this entry git gives this hash 4d196a6c74b7ec9a6f805b7b2959c827a0628cd8

juan-lunarg (approved):

.git-blame-ignore-revs
ncesario-lunarg: ^ yeah, this is strange

spencer-lunarg: ok, I'm stupid, I went through the PRs and took the commits from there, realized when merged in, these change for what it is in the current main branch... will update date

spencer-lunarg: fixed! Thanks for catching this

PR #5142, opened 2023-01-20
macos: Require VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME

juan-lunarg:
With this the majority of the testing on MacOS passes.

### DIFF
--- a/tests/positive/other.cpp
+++ b/tests/positive/other.cpp
@@ -1,7 +1,7 @@
 /*
  * Copyright (c) 2015-2021 The Khronos Group Inc.
  * Copyright (c) 2015-2022 Valve Corporation
- * Copyright (c) 2015-2022 LunarG, Inc.
+ * Copyright (c) 2015-2023 LunarG, Inc.
  * Copyright (c) 2015-2022 Google, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -783,11 +783,6 @@ TEST_F(VkPositiveLayerTest, ExportMetalObjects) {
 
     AddRequiredExtensions(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
     AddRequiredExtensions(VK_EXT_METAL_OBJECTS_EXTENSION_NAME);
-    AddRequiredExtensions(VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME);
-    AddRequiredExtensions(VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME);
-    if (!InstanceExtensionSupported(VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME)) {
-        GTEST_SKIP() << VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME << " not supported";
-    }
 
     // Initialize framework
     {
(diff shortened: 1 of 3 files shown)


tests/vkrenderframework.cpp
```diff
@@ -315,7 +315,9 @@ void VkRenderFramework::InitFramework(void * /*unused compatibility parameter*/,
     }
 
 #ifdef VK_USE_PLATFORM_METAL_EXT
-    instance_extensions_.push_back(VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME);
+    AddRequiredExtensions(VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME);
+    // Beginning with the 1.3.216 Vulkan SDK, the VK_KHR_PORTABILITY_subset extension is mandatory.
+    AddRequiredExtensions(VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME);
```
charles-lunarg, line 320: I'm very confused by this. Portability Enumeration is an *instance* extension while Portability Subset is a *device* extension. does `AddRequiredExtensions` somehow differentiate the two?)

tests/positive/other.cpp
```diff
@@ -783,11 +783,6 @@ TEST_F(VkPositiveLayerTest, ExportMetalObjects) {
 
     AddRequiredExtensions(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
     AddRequiredExtensions(VK_EXT_METAL_OBJECTS_EXTENSION_NAME);
```
juan-lunarg, line 785: I debated just adding this extension as required for the test framework since I want to make sure these are tested on CI. Thoughts?

Since there is really only 1 target for Mac (MoltenVK) it should be fine?

tests/vkrenderframework.cpp
juan-lunarg: `AddRequiredExtension` figures out which one it belongs to and applies it correctly. If I'm reading the code correctly.

tests/positive/other.cpp
charles-lunarg: `VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME` is only necessary if you are using Vulkan 1.0.
Otherwise the necessary functions are available.
However, I'm fine with just always enabling the extension. I *think* the loader handles the case where the app enabled 1.1 and the extension just fine.

juan-lunarg: I was referring to `VK_EXT_METAL_OBJECTS_EXTENSION_NAME` although it would be nice to just add `VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME` as well for simplicity.

tests/vkrenderframework.cpp
ncesario-lunarg: > `AddRequiredExtension` figures out which one it belongs to and applies it correctly. If I'm reading the code correctly.

:heavy_plus_sign:

ncesario-lunarg (approved):

tests/positive/other.cpp
```diff
@@ -1,7 +1,7 @@
 /*
  * Copyright (c) 2015-2021 The Khronos Group Inc.
  * Copyright (c) 2015-2022 Valve Corporation
- * Copyright (c) 2015-2022 LunarG, Inc.
+ * Copyright (c) 2015-2023 LunarG, Inc.
```
lunarpapillo, line 4: I think that most of our changes are Valve-sponsored, so the Valve copyright should also be updated.

juan-lunarg: Sounds right, I'll update the script to error check this.

PR #5145, opened 2023-01-21
tests: More cleanup to tests using profiles

spencer-lunarg:
The following tests were not working with local max profile

### DIFF
--- a/tests/vklayertests_others.cpp
+++ b/tests/vklayertests_others.cpp
@@ -4591,13 +4591,8 @@ TEST_F(VkLayerTest, QueryPerformanceCreation) {
         GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported";
     }
 
-    PFN_vkGetPhysicalDeviceFeatures2KHR vkGetPhysicalDeviceFeatures2KHR =
-        (PFN_vkGetPhysicalDeviceFeatures2KHR)vk::GetInstanceProcAddr(instance(), "vkGetPhysicalDeviceFeatures2KHR");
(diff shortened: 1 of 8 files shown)


juan-lunarg (approved):
:+1:

tests/vklayertests_others.cpp
```diff
@@ -4587,18 +4587,13 @@ TEST_F(VkLayerTest, QueryPerformanceCreation) {
...
     auto performance_features = LvlInitStruct<VkPhysicalDevicePerformanceQueryFeaturesKHR>();
-    features2 = LvlInitStruct<VkPhysicalDeviceFeatures2KHR>(&performance_features);
-    vkGetPhysicalDeviceFeatures2KHR(gpu(), &features2);
+    auto features2 = GetPhysicalDeviceFeatures2(performance_features);
     if (!performance_features.performanceCounterQueryPools) {
         GTEST_SKIP() << "Performance query pools are not supported.";
     }
 
-    ASSERT_NO_FATAL_FAILURE(InitState(nullptr, &performance_features));
+    ASSERT_NO_FATAL_FAILURE(InitState(nullptr, &features2));
```
ncesario-lunarg, line 4596: Are `features2` necessary, or are only the features we care about present in `performance_features`? Similar comments for below instances.

ncesario-lunarg (approved):
One question, but LGTM!

tests/vklayertests_others.cpp
spencer-lunarg: nope you are right, changing back, good catch

PR #5146, opened 2023-01-21
layers: Move some CmdBeginRendering VUs to stateless

spencer-lunarg:
The `ValidateCmdBeginRendering` was growing quite large and there are some new VUs that need to be added, this first moves more checks to statelesss


jeremyg-lunarg (approved):
LGTM. The copyright-only changes to generated files are because all those files share codegen and you changed the copyright header in that codegen. Right?

spencer-lunarg:
> The copyright-only changes to generated files are because all those files share codegen 

yep!

PR #5147, opened 2023-01-22
Use CMD_TYPE and consistent promotion variable names in stateless

spencer-lunarg:
Cleanup from working in the stateless files


arno-lunarg (approved):
LGTM!
I did not know about `CMD_TYPE` :0

PR #5148, opened 2023-01-22
layers: Use consistent function names in stateless

spencer-lunarg:
Takes all the `validate_*()` functions in stateless and follows the function naming convention used elsewhere (plan to add to git-blame-ignore after)


arno-lunarg (approved):
LGTM!
(I only really paid attention to modifications made in `stateless_validation.h`, since I guess any unmodified function call in source files will just not compile anymore)

Issue #5149, opened 2023-01-22
Missing validation in VkSubmitInfo (pWaitDstStageMask)


BeastLe9enD:
If `VkSubmitInfo::waitSemaphoreCount > 0` and `VkSubmitInfo::pWaitDstStageMask` is `nullptr`, it results in an access violation.
 
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/layers/best_practices_utils.cpp#L1477
In the line listed above, we have a loop from `0` to `waitSemaphoreCount`, and we access `pWaitDstStageMask[semaphore]` without checking if its `nullptr`, which results in the access violation. We should log `VUID-VkSubmitInfo-pWaitDstStageMask-parameter` here instead in my opinion. 
I can make a pull request with this fix if you want.

spencer-lunarg:
@BeastLe9enD thanks for bringing this up, a few things going on here

This is currently being accessed in the Best Practice layers when `pWaitDstStageMask` is null, **but** the normal validation layers "should" have caught `VUID-VkSubmitInfo-pWaitDstStageMask-parameter`

VUs such as `VUID-VkSubmitInfo-pWaitDstStageMask-parameter` are generated, but seems there is no generation going on for this

What is needed is

1. Fix the scripts so this is caught before getting into the Best Practice layers
2. A tests to prevent this from happening again 

> I can make a pull request with this fix if you want.

I will admit the generated code logic is not the shiniest part of this code base, I am happy to take this as there might be other hidden issues going on

spencer-lunarg:
ok, so some more information after looking into this more

The `vk.xml` has this as

```xml
<member optional="false,true" len="waitSemaphoreCount">const <type>VkPipelineStageFlags</type>* <name>pWaitDstStageMask</name></member>
```

`optional="false,true"` is suppose to represent that `pWaitDstStageMask` must not be `null`, but the values inside the array **can** still be `null` (in this case `VK_PIPELINE_STAGE_NONE_KHR`)

Seems this is not currently handled correctly in the parameter validation generation script

PR #5150, opened 2023-01-23
Add missing Flag Array parameter VU

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5149

Adds proper support for

- `VUID-VkDescriptorSetLayoutBindingFlagsCreateInfo-pBindingFlags-parameter`
- `VUID-VkSubmitInfo-pWaitDstStageMask-parameter`
- `VUID-vkCmdSetColorWriteMaskEXT-pColorWriteMasks-parameter`

These are the only 3 things that use `ValidateFlagsArray`

### DIFF
--- a/layers/stateless_validation.h
+++ b/layers/stateless_validation.h
@@ -863,28 +863,22 @@ class StatelessValidation : public ValidationObject {
      * @param count Number of VkFlags values in the array.
      * @param array Array of VkFlags value to validate.
      * @param count_required The 'count' parameter may not be 0 when true.
-     * @param array_required The 'array' parameter may not be NULL when true.
+     * @param array_required_vuid The VUID for the 'array' parameter.
      * @return Boolean value indicating that the call should be skipped.
      */
     bool ValidateFlagsArray(const char *api_name, const ParameterName &count_name, const ParameterName &array_name,
                             const char *flag_bits_name, VkFlags all_flags, uint32_t count, const VkFlags *array,
-                            bool count_required, bool array_required) const {
+                            bool count_required, const char *array_required_vuid) const {
         bool skip_call = false;
 
-        if ((count == 0) || (array == nullptr)) {
-            skip_call |= ValidateArray(api_name, count_name, array_name, count, &array, count_required, array_required,
-                                       kVUIDUndefined, kVUIDUndefined);
+        if (array == nullptr) {
+            // Flag arrays always need to have a valid array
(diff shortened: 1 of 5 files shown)


layers/stateless_validation.h
```diff
@@ -863,28 +863,22 @@ class StatelessValidation : public ValidationObject {
...
-                                       kVUIDUndefined, kVUIDUndefined);
+        if (array == nullptr) {
+            // Flag arrays always need to have a valid array
+            skip_call |= ValidateArray(api_name, count_name, array_name, count, &array, count_required, true, kVUIDUndefined,
+                                       array_required_vuid);
         } else {
             // Verify that all VkFlags values in the array
             for (uint32_t i = 0; i < count; ++i) {
-                if (array[i] == 0) {
-                    // Current XML registry logic for validity generation uses the array parameter's optional tag to determine if
```
spencer-lunarg, line 881: This was actually a mistake, was corrected a while ago in
https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5376

but all Flag Array should be `optional="false,true"`

tests/vklayertests_others.cpp
```diff
@@ -9886,6 +9892,18 @@ TEST_F(VkLayerTest, ValidateExtendedDynamicState3Enabled) {
 
         commandBuffer.end();
     }
+
+    if (extended_dynamic_state3_features.extendedDynamicState3ColorWriteMask) {
+        auto vkCmdSetColorWriteMaskEXT =
+            (PFN_vkCmdSetColorWriteMaskEXT)vk::GetDeviceProcAddr(m_device->device(), "vkCmdSetColorWriteMaskEXT");
```
arno-lunarg, line 9898:
```suggestion
        auto vkCmdSetColorWriteMaskEXT =
            GetDeviceProcAddr<PFN_vkCmdSetColorWriteMaskEXT>("vkCmdSetColorWriteMaskEXT");
```

arno-lunarg (commented):
LGTM, not feeling to approve as I am not exactly sure about what is going on with the script modifications and the behaviour change in `ValidateFlagsArray`

jeremyg-lunarg (approved):
LGTM

tests/vklayertests_others.cpp
spencer-lunarg: good catch

PR #5153, opened 2023-01-23
Cleanup github templates and BUILD.md

juan-lunarg:
- Add code sections to github issue templates, these code sections are collapsible which helps with readability of issues.
- Remove unnecessary detail from the BUILD.md and highlight new CMake functionality
- Point to official CMake docs for installation


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -7,16 +7,33 @@ assignees: ''
...
+**Valid Usage ID**
+If applicable, please include the validation messages encountered leading up to the issue
+
 **Additional context**
+
+<details>
+<summary>code or terminal output</summary>
+
+```cpp
+int main()
```
spencer-lunarg, line 33: I expect most people to not have spent time digging through the code, I would make this example of the output like

```
Validation Error: [ VUID-vkCmdDrawMultiEXT-colorAttachmentCount-06188 ] Object 0: handle = 0x3d47e60 ...
```

spencer-lunarg: I guess if I saw this, I would be curious if they (the repo owners) wanted code to reproduce it here

PR #5158, opened 2023-01-24
build: Fix duplicate xxhash object files

juan-lunarg:
Currently both VkLayer_utils and VVL build xxhash.c which results in duplicate identical object files. Now VkLayer_utils is responsible for the xxhash object file.

Also turn xxhash.c to xxhash.cpp to avoid mixed language usage in a single library. This helps avoid minor confusions in the CMake code.


juan-lunarg:
Example ninja build output from before this change:

```sh
Building C object layers/CMakeFiles/VkLayer_utils.dir/xxhash.c.o
...
Building C object layers/CMakeFiles/VkLayer_khronos_validation.dir/xxhash.c.o
```

Now it looks like this:

```sh
Building CXX object layers/CMakeFiles/VkLayer_utils.dir/xxhash.cpp.o
```

ncesario-lunarg (approved):
Would creating a static library/target containing xxhash.c[pp] and linking against that work? If we have multiple targets that depend on xxhash, seems like they should each "depend" on xxhash. That's _assuming_ cmake has the smarts to not bring in the xxhash target/lib twice when linking the .so.

juan-lunarg:
> Would creating a static library/target containing xxhash.c[pp] and linking against that work?

I tried that approach but it doesn't work cleanly due to the fact we have to install `VkLayer_utils` which `LunarG/VulkanTools` depends on.

ncesario-lunarg:
> > Would creating a static library/target containing xxhash.c[pp] and linking against that work?
> 
> I tried that approach but it doesn't work cleanly due to the fact we have to install `VkLayer_utils` which `LunarG/VulkanTools` depends on.

Ah yes, another "TODO" :/

PR #5159, opened 2023-01-24
cmake: Remove custom install code

juan-lunarg:


juan-lunarg:
FWIW I have already removed this functionality from `Vulkan-Headers` and have had no problems/complaints.

If this code was really valuable it would have been upstreamed to CMake by now either as native or helper module support.

ncesario-lunarg (approved):
LGTM; you have better insight as to what is actually using the install targets, so I'll assume they all work as expected :)

PR #5166, opened 2023-01-25
gn: Remove deprecated vulkan_core.h parsing

juan-lunarg:
Should have been part of https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/ac7c75bb519ede9ee65e43778da2274002c525b0

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -268,13 +268,8 @@ if (!is_android) {
       deps += [ ":vulkan_clean_old_validation_layer_objects" ]
     }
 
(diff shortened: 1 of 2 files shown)


ncesario-lunarg:
> > I completely forgot we had a `build-gn` directory cry
> 
> Is `build-gn` an arbitrary name/location? Could it be moved into the scripts folder?

I don't think there is anythnig special about `build-gn`, but I also don't know what's currently referencing that directory. At first glance looks like it's just BUILD.gn and scripts/github_ci_gn.py.

Issue #5136, opened 2023-01-19
VUID-vkCmdDrawMultiEXT-colorAttachmentCount-06188


zmike:
I'm getting this error when using dynamic rasterization samples from EDS3:

```
VUID-vkCmdDrawMultiEXT-colorAttachmentCount-06188(ERROR / SPEC): msgNum: 1026350275 - Validation Error: [ VUID-vkCmdDrawMultiEXT-colorAttachmentCount-06188 ] Object 0: handle = 0x3d47e60, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x3d2cdcc3 | vkCmdDrawMultiEXT: Color attachment (0) sample count (VK_SAMPLE_COUNT_4_BIT) must match corresponding VkPipelineMultisampleStateCreateInfo sample count (VK_SAMPLE_COUNT_1_BIT) The Vulkan spec states: If the currently bound pipeline was created without a VkAttachmentSampleCountInfoAMD or VkAttachmentSampleCountInfoNV structure, and the current render pass instance was begun with vkCmdBeginRendering with a VkRenderingInfo::colorAttachmentCount parameter greater than 0, then each element of the VkRenderingInfo::pColorAttachments array with a imageView not equal to VK_NULL_HANDLE must have been created with a sample count equal to the value of VkPipelineMultisampleStateCreateInfo::rasterizationSamples used to create the currently bound graphics pipeline (https://github.com/KhronosGroup/Vulkan-Docs/search?q=)VUID-vkCmdDrawMultiEXT-colorAttachmentCount-06188)
```

This is no longer an actual VU though.

Using VVL 57fe3f50cde464dbfd790cdb035f5247e38027b2.

zmike:
Can be reproduced with this caselist:

```
dEQP-GLES31.functional.shaders.helper_invocation.value.triangles
dEQP-GLES31.functional.shaders.helper_invocation.value.wide_points_4_samples
```

using zink on any vulkan driver that supports EDS3:

`VK_INSTANCE_LAYERS=VK_LAYER_KHRONOS_validation MESA_LOADER_DRIVER_OVERRIDE=zink ./glcts --deqp-caselist-file l`

spencer-lunarg:
> This is no longer an actual VU though.

[...]

PR #5168, opened 2023-01-26
Fix dynamic rasterizationSamples

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5136

There is [future spec language](https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5678) for this as well

The 3 main things done in the 2 layer commits

- Group all the checks in `ValidateGraphicsPipelineMultisampleState` that were dependent of the pipeline using `VK_DYNAMIC_STATE_RASTERIZATION_SAMPLES_EXT`
- Move the `GetNumSamples` from `PIPELINE_STATE` to `CMD_BUFFER_STATE` (and rename to `GetRasterizationSamples`)
- Track `vkCmdSetRasterizationSamplesEXT` value being set

### DIFF
--- a/layers/core_checks/pipeline_validation.cpp
+++ b/layers/core_checks/pipeline_validation.cpp
@@ -253,7 +253,8 @@ bool CoreChecks::ValidatePipeline(std::vector<std::shared_ptr<PIPELINE_STATE>> c
     }
 
     if (rp_state && rp_state->renderPass() != VK_NULL_HANDLE &&
-        IsExtEnabled(device_extensions.vk_ext_multisampled_render_to_single_sampled)) {
+        IsExtEnabled(device_extensions.vk_ext_multisampled_render_to_single_sampled) &&
+        !pipeline.IsDynamic(VK_DYNAMIC_STATE_RASTERIZATION_SAMPLES_EXT)) {
         const auto msrtss_info = LvlFindInChain<VkMultisampledRenderToSingleSampledInfoEXT>(subpass_desc->pNext);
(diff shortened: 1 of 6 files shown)


juan-lunarg (approved):
LGTM :+1:

layers/core_checks/pipeline_validation.cpp
```diff
@@ -1994,6 +1993,7 @@ bool CoreChecks::ValidateGraphicsPipelineMultisampleState(const PIPELINE_STATE &
             }
         };
 
+        const uint32_t raster_samples = SampleCountSize(multisample_state->rasterizationSamples);
```
artem-lunarg, line 1996: Not related directly to this PR, minor observation. `SampleCountSize` contract allows 0 as return value. Valid `VkSampleCountFlagBits` values can't be zero and also standard guarantees (due to compatibility) that cast of enum to uint will produce correct sample count. In this regard current `SampleCountSize` implementation looks less robust comparing to `static_cast<uint>(enum_value)`, which might just replace entire switch. Except the case when we indeed have a code that allows on 0 for VkSampleCountFlagBits variable.

artem-lunarg:
LGTM!

artem-lunarg (approved):

layers/core_checks/pipeline_validation.cpp
spencer-lunarg: agree, going to make another PR to fix that and look for other functions doing this while at it

Issue #4539, opened 2022-09-11
Missing checks for three handle export VUIDs


Rua:
Three similar VUIDs are currently not validated:

VUID-VkExportMemoryAllocateInfo-handleTypes-00656
VUID-VkExportFenceCreateInfo-handleTypes-01446
VUID-VkExportSemaphoreCreateInfo-handleTypes-01124

Rua:
I noticed two more VUIDs related to external memory that aren't validated yet:
VUID-VkBufferCreateInfo-pNext-00920
VUID-VkImageCreateInfo-pNext-00990

PR #5169, opened 2023-01-26
VUIDs for external memory compatibility when creating resource

artem-lunarg:
VUIDs: 00920, 00990, 00991. (fixes part of https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4539)

External memory types specified in buffer/image create info structures should be compatible.
The compatibility property also includes supported status. If a handle type is not supported, it is not reported as compatible (even with itself).

Fixes to existing tests where new VUs detected issues.

### DIFF
--- a/layers/core_checks/buffer_validation.cpp
+++ b/layers/core_checks/buffer_validation.cpp
@@ -2190,21 +2190,81 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
                          "vkCreateImage(): VkImageCreateInfo struct has both VkExternalMemoryImageCreateInfoNV and "
                          "VkExternalMemoryImageCreateInfo chained structs.");
     }
-    if (external_memory_create_info) {
-        if (external_memory_create_info->handleTypes != 0 && pCreateInfo->initialLayout != VK_IMAGE_LAYOUT_UNDEFINED) {
+    if (external_memory_create_info && external_memory_create_info->handleTypes != 0) {
+        if (pCreateInfo->initialLayout != VK_IMAGE_LAYOUT_UNDEFINED) {
             skip |= LogError(
                 device, "VUID-VkImageCreateInfo-pNext-01443",
                 "vkCreateImage: VkImageCreateInfo pNext chain includes VkExternalMemoryImageCreateInfo with handleTypes %" PRIu32
                 " but pCreateInfo->initialLayout is %s.",
                 external_memory_create_info->handleTypes, string_VkImageLayout(pCreateInfo->initialLayout));
         }
-    } else if (external_memory_create_info_nv) {
-        if (external_memory_create_info_nv->handleTypes != 0 && pCreateInfo->initialLayout != VK_IMAGE_LAYOUT_UNDEFINED) {
+        // Check external memory handle types compatibility
+        const uint32_t any_type = 1u << MostSignificantBit(external_memory_create_info->handleTypes);
+        auto external_image_info = LvlInitStruct<VkPhysicalDeviceExternalImageFormatInfo>();
+        external_image_info.handleType = static_cast<VkExternalMemoryHandleTypeFlagBits>(any_type);
+        auto image_info = LvlInitStruct<VkPhysicalDeviceImageFormatInfo2>(&external_image_info);
+        image_info.format = pCreateInfo->format;
+        image_info.type = pCreateInfo->imageType;
+        image_info.tiling = pCreateInfo->tiling;
+        image_info.usage = pCreateInfo->usage;
+        image_info.flags = pCreateInfo->flags;
+
+        auto external_image_properties = LvlInitStruct<VkExternalImageFormatProperties>();
+        auto image_properties = LvlInitStruct<VkImageFormatProperties2>(&external_image_properties);
+        result = DispatchGetPhysicalDeviceImageFormatProperties2(physical_device, &image_info, &image_properties);
+        const auto compatible_types = external_image_properties.externalMemoryProperties.compatibleHandleTypes;
+
+        if (result != VK_SUCCESS) {
+            skip |= LogError(
+                device, "VUID-VkImageCreateInfo-pNext-00990",
+                "vkCreateImage(): The handle type (%s), format (%s), type (%s), tiling (%s), usage (%s), flags (%s) "
+                "is not supported combination of parameters and vkGetPhysicalDeviceImageFormatProperties2 returned back %s.",
+                string_VkExternalMemoryHandleTypeFlagBits(external_image_info.handleType), string_VkFormat(image_info.format),
+                string_VkImageType(image_info.type), string_VkImageTiling(image_info.tiling),
+                string_VkImageUsageFlags(image_info.usage).c_str(), string_VkImageCreateFlags(image_info.flags).c_str(),
+                string_VkResult(result));
+        } else if ((external_memory_create_info->handleTypes & compatible_types) != external_memory_create_info->handleTypes) {
+            const bool single_flag = GetBitSetCount(external_memory_create_info->handleTypes) == 1;
+            skip |= LogError(device, "VUID-VkImageCreateInfo-pNext-00990",
+                             "vkCreateImage(): VkImageCreateInfo pNext chain contains VkExternalMemoryImageCreateInfo with "
+                             "%s (%s).",
+                             single_flag ? "unsupported flag" : "incompatible flags",
+                             string_VkExternalMemoryHandleTypeFlags(external_memory_create_info->handleTypes).c_str());
+        }
+    } else if (external_memory_create_info_nv && external_memory_create_info_nv->handleTypes != 0) {
+        if (pCreateInfo->initialLayout != VK_IMAGE_LAYOUT_UNDEFINED) {
             skip |= LogError(
                 device, "VUID-VkImageCreateInfo-pNext-01443",
                 "vkCreateImage: VkImageCreateInfo pNext chain includes VkExternalMemoryImageCreateInfoNV with handleTypes %" PRIu32
                 " but pCreateInfo->initialLayout is %s.",
                 external_memory_create_info_nv->handleTypes, string_VkImageLayout(pCreateInfo->initialLayout));
+        }
+        // Check external memory handle types compatibility
+        const uint32_t any_type = 1u << MostSignificantBit(external_memory_create_info_nv->handleTypes);
+        auto handle_type = static_cast<VkExternalMemoryHandleTypeFlagBitsNV>(any_type);
+        VkExternalImageFormatPropertiesNV external_image_properties = {};
+        result = DispatchGetPhysicalDeviceExternalImageFormatPropertiesNV(
+            physical_device, pCreateInfo->format, pCreateInfo->imageType, pCreateInfo->tiling, pCreateInfo->usage,
+            pCreateInfo->flags, handle_type, &external_image_properties);
+        const auto compatible_types = external_image_properties.compatibleHandleTypes;
+
+        if (result != VK_SUCCESS) {
+            skip |= LogError(device, "VUID-VkImageCreateInfo-pNext-00991",
+                             "vkCreateImage(): The handle type (%s), format (%s), type (%s), tiling (%s), usage (%s), flags (%s) "
+                             "is not supported combination of parameters and vkGetPhysicalDeviceExternalImageFormatPropertiesNV "
+                             "returned back %s.",
+                             string_VkExternalMemoryHandleTypeFlagBitsNV(handle_type), string_VkFormat(pCreateInfo->format),
+                             string_VkImageType(pCreateInfo->imageType), string_VkImageTiling(pCreateInfo->tiling),
+                             string_VkImageUsageFlags(pCreateInfo->usage).c_str(),
+                             string_VkImageCreateFlags(pCreateInfo->flags).c_str(), string_VkResult(result));
+        } else if ((external_memory_create_info_nv->handleTypes & compatible_types) !=
+                   external_memory_create_info_nv->handleTypes) {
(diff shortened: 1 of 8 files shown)


tests/layer_validation_tests.cpp
```diff
@@ -25,6 +25,8 @@
  */
 #include "cast_utils.h"
 #include "layer_validation_tests.h"
+#include "enum_flag_bits.h"
```
artem-lunarg, line 28: `enum_flag_bits.h` depends on vulkan definitions from `layer_validaton_tests.h`, that's why not in lexicographical order.

jeremyg-lunarg: Can enum_flag_bits.h include the header it needs directly rather than relying on include order like this?

tests/layer_validation_tests.cpp
```diff
@@ -497,6 +499,58 @@ VkFormat FindFormatWithoutFeatures(VkPhysicalDevice gpu, VkImageTiling tiling, V
...
+    auto image_info = LvlInitStruct<VkPhysicalDeviceImageFormatInfo2>(&external_image_info);
+    image_info.format = image_create_info.format;
+    image_info.type = image_create_info.imageType;
+    image_info.tiling = image_create_info.tiling;
+    image_info.usage = image_create_info.usage;
+    image_info.flags = image_create_info.flags;
+
+    VkExternalMemoryHandleTypeFlags supported_handle_type = 0;
+    IterateFlags<VkExternalMemoryHandleTypeFlagBits>(
+        AllVkExternalMemoryHandleTypeFlagBits, [&](VkExternalMemoryHandleTypeFlagBits flag) {
```
artem-lunarg, line 515: Capture all variables. Most locals are used inside lambda anyway, and listing them resulted in less readable code.

tests/layer_validation_tests.cpp
artem-lunarg: It's generated code. Currently I'm not sure what are the restrictions and requirements for it. It also does not have any `#pragma once` or include guards.

artem-lunarg: According to @mikes-lunarg it's fine to fix enum_flag_bits.h. I can do this as a separate change, since it will touch other files not related to this PR.

ncesario-lunarg: I'm in the process of making changes to generated enum code, so I can probably make the necessary changes to `enum_flag_bits.h`.

artem-lunarg:
Crash on Jenkins CI on some android devices (not all). It happens in the test that was not touched (at least directly) by this commit.

Test name: `ValidateImportMemoryHandleType`

[logcat.txt](https://github.com/KhronosGroup/Vulkan-ValidationLayers/files/10513120/logcat.txt)

layers/core_checks/buffer_validation.cpp
```diff
@@ -2198,6 +2198,30 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
                 " but pCreateInfo->initialLayout is %s.",
                 external_memory_create_info->handleTypes, string_VkImageLayout(pCreateInfo->initialLayout));
         }
+        if (external_memory_create_info->handleTypes != 0) {
```
spencer-lunarg, line 2201: would just restructure this as

```c++
if (external_memory_create_info && external_memory_create_info->handleTypes != 0) {
    if (pCreateInfo->initialLayout != VK_IMAGE_LAYOUT_UNDEFINED) {
        // VUID-VkImageCreateInfo-pNext-01443"
    }

    // all new code added in this PR
```

layers/core_checks/buffer_validation.cpp
```diff
@@ -2198,6 +2198,30 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
...
+            image_info.tiling = pCreateInfo->tiling;
+            image_info.usage = pCreateInfo->usage;
+            image_info.flags = pCreateInfo->flags;
+
+            auto external_image_properties = LvlInitStruct<VkExternalImageFormatProperties>();
+            auto image_properties = LvlInitStruct<VkImageFormatProperties2>(&external_image_properties);
+            VkResult result = DispatchGetPhysicalDeviceImageFormatProperties2(physical_device, &image_info, &image_properties);
+            const auto compatible_types = external_image_properties.externalMemoryProperties.compatibleHandleTypes;
+
+            if (result != VK_SUCCESS ||
```
spencer-lunarg, line 2217: if result is bad, should either return the `string_VkResult(result)` for it, or I am still in favor of just skipping this check if it is not `VK_SUCCESS` (for now until we finish making a more unify way of handling this)

so this could become `result == VK_SUCCESS &&` instead

If the call is not successful, throwing this Validation error will just developers searching down the wrong path

layers/core_checks/buffer_validation.cpp
artem-lunarg: Yes, that's better.

layers/core_checks/buffer_validation.cpp
artem-lunarg: The `result  != VK_SUCCESS` was intended to  deal with `VK_ERROR_FORMAT_NOT_SUPPORTED`,  one option is to use this condition:
 `if (result == VK_SUCCESS && !flags_compatible || result == VK_ERROR_FORMAT_NOT_SUPPORTED) `.

With current drivers, `vkGetPhysicalDeviceImageFormatProperties2` reports `VK_ERROR_FORMAT_NOT_SUPPORTED` when the platform does not support a handle type. 

There might be some benefits to checking _support_ status in 00990 (during vkCreateImage).

a) There is VU 00656 that checks for type _support_. Still, it happens later (vkBindImageMemory is the earliest point where validation can be done), and VU 00656 checks a different set of flags (VkExportMemoryAllocateInfo::handleTypes vs VkExternalMemoryImageCreateInfo::handleTypes ). Some handle types specified in vkCreateImage might never be checked for _support_ by 00656.

b) Checking support in VU 00990 also matches buffer behavior. `vkGetPhysicalDeviceExternalBufferProperties` does not have a return code. During buffer's 00920 validation for unsupported handle type, it feels `compatibleHandleTypes` with zeros, and the compatibility check does not pass. So, for a buffer, _compatibility_ check automatically includes _support_ check, and unsupported types are detected early during vkCreateBuffer. That was the motivation for 00990 to react to the error code (but instead of any code, we can check VK_ERROR_FORMAT_NOT_SUPPORTED) to have behavior consistent with buffer validation.

spencer-lunarg: all good points and now thinking about it more, I think lets keep it, but just do

```c++
if (result != VK_SUCCESS) {
    LogError("VUID-VkImageCreateInfo-pNext-00990", "The format (%s), type (%s), .... is not supported" , string_VkFormat(image_info.format), /*...*/); 
else if (external_memory_create_info->handleTypes & compatible_types) != external_memory_create_info->handleTypes) {
    LogError("VUID-VkImageCreateInfo-pNext-00990", /*what you have already here*/);
}
```

layers/vk_layer_utils.h
```diff
@@ -178,6 +178,21 @@ static inline int MostSignificantBit(uint32_t mask) {
 #endif
 }
 
+// Iterates over all set bits and calls the callback with a bit mask corresponding to each flag.
+// FlagBits and Flags follow Vulkan naming convensions for flag types.
+// An example of a more efficient implementation: https://lemire.me/blog/2018/02/21/iterating-over-set-bits-quickly/
+template <typename FlagBits, typename Flags>
+void IterateFlags(Flags flags, std::function<bool(FlagBits flag)> callback) {
```
arno-lunarg, line 185:
```suggestion
template <typename FlagBits, typename Flags, typename Callback>
void IterateFlags(Flags flags, Callback callback) {
```

I think it is safe to let the compiler decide the best type to choose for `callback`?

layers/core_checks/buffer_validation.cpp
```diff
@@ -2206,6 +2230,23 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
                 " but pCreateInfo->initialLayout is %s.",
                 external_memory_create_info_nv->handleTypes, string_VkImageLayout(pCreateInfo->initialLayout));
         }
+        if (external_memory_create_info_nv->handleTypes != 0) {
+            const uint32_t any_type = 1u << MostSignificantBit(external_memory_create_info_nv->handleTypes);
+            VkExternalMemoryHandleTypeFlagsNV handle_type = static_cast<VkExternalMemoryHandleTypeFlagBitsNV>(any_type);
```
arno-lunarg, line 2235: There is a type mismatch between `handle_type` and the `static_cast`, probably a typo?
Probably should use `auto handle_type`

layers/vk_layer_utils.h
artem-lunarg: That's great suggestion. I wanted something like this :)

layers/core_checks/buffer_validation.cpp
artem-lunarg: Yes, it's better with auto. I think initial code also makes sense, nvidia extension is different that it accepts Flags and not a FlagBits as in core functionality. static cast is concerned only with a single flag conversion uint -> FlagBits and then the type in the assignment is whatever expected by the API (FlagsBits in core or Flags in NV).

arno-lunarg: Oh yes indeed the current code better transcribes the data flow

layers/core_checks/buffer_validation.cpp
```diff
@@ -2190,22 +2190,78 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
...
+        image_info.flags = pCreateInfo->flags;
+
+        auto external_image_properties = LvlInitStruct<VkExternalImageFormatProperties>();
+        auto image_properties = LvlInitStruct<VkImageFormatProperties2>(&external_image_properties);
+        VkResult result = DispatchGetPhysicalDeviceImageFormatProperties2(physical_device, &image_info, &image_properties);
+        const auto compatible_types = external_image_properties.externalMemoryProperties.compatibleHandleTypes;
+
+        if (result != VK_SUCCESS) {
+            skip |= LogError(device, "VUID-VkImageCreateInfo-pNext-00990",
+                             "The handle type (%s), format (%s), type (%s), tiling (%s), usage (%s), flags (%s) is not supported.",
```
spencer-lunarg, line 2219: nit: Would also print

- this was during a `vkCreateImage(): ` call
- What failed (`vkGetPhysicalDeviceExternalImageFormatProperties`)

tests/vklayertests_buffer_image_memory_sampler.cpp
```diff
@@ -11194,6 +11196,162 @@ TEST_F(VkLayerTest, BufferDeviceAddressKHRDisabled) {
...
+    image_create_info.samples = VK_SAMPLE_COUNT_1_BIT;
+    image_create_info.tiling = VK_IMAGE_TILING_OPTIMAL;
+    image_create_info.usage = VK_IMAGE_USAGE_SAMPLED_BIT;
+    image_create_info.sharingMode = VK_SHARING_MODE_EXCLUSIVE;
+
+    // Get all exportable handle types supported by the platform.
+    VkExternalMemoryHandleTypeFlagsNV supported_handle_types = 0;
+    VkExternalMemoryHandleTypeFlagsNV any_compatible_group = 0;
+
+    PFN_vkGetPhysicalDeviceExternalImageFormatPropertiesNV vkGetPhysicalDeviceExternalImageFormatPropertiesNV =
```
arno-lunarg, line 11330:
```suggestion
   auto vkGetPhysicalDeviceExternalImageFormatPropertiesNV = GetInstanceProcAddr<PFN_vkGetPhysicalDeviceExternalImageFormatPropertiesNV >("vkGetPhysicalDeviceExternalImageFormatPropertiesNV ");
```

tests/vklayertests_others.cpp
```diff
@@ -7455,7 +7455,15 @@ TEST_F(VkLayerTest, ValidateImportMemoryHandleType) {
...
     buffer_export.init_no_mem(*m_device, buffer_info);
-    external_buffer_info.handleTypes = wrong_handle_type;
+
+    auto importable_buffer_types =
+        FindSupportedExternalMemoryHandleTypes(gpu(), buffer_info, VK_EXTERNAL_MEMORY_FEATURE_IMPORTABLE_BIT);
+    importable_buffer_types &= ~handle_type;  // we need to find a flag that is different from handle_type
+    if (importable_buffer_types == 0) GTEST_SKIP() << "Cannot find two different buffer handle types, skipping test";
+    auto wrong_buffer_handle_type =
+        static_cast<VkExternalMemoryHandleTypeFlagBits>(1 << MostSignificantBit(importable_buffer_types));
+    external_buffer_info.handleTypes = wrong_buffer_handle_type;
```
arno-lunarg, line 7465: Isn't this supposed to trigger `00920` ?? Or do I get this wrong ^^"
(I did not find the `m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkBufferCreateInfo-pNext-00920");` call)

artem-lunarg: No, this test should not trigger 00920. New validation started to trigger 00920 here and resource creation failed. This was preventing validation of VUIDs that test is responsible for. The fixes to this test ensure that we select handle type in such way that 00920 is not generated, but at the same time test pre-conditions are true (import handle type != export handle type)

artem-lunarg: ...and `wrong` terminology here  is related to another VUID, that the test checks below, not about 00920!

arno-lunarg: ok, thank you for the explanation!

spencer-lunarg (approved):

PR #5174, opened 2023-01-26
gn: Fix GN GitHub action

juan-lunarg:
- Fix failure caused by clang-format DEP
- Use GitHub caching


build-gn/DEPS
```diff
@@ -44,19 +44,6 @@ deps = {
 }
 
 hooks = [
-  # Pull clang-format binaries using checked-in hashes.
-  {
```
juan-lunarg, line 48: This is the main cause of the recent CI failures and removing it has no impact to the build.

mikes-lunarg: I wonder if it could also be fixed by updating the toolchain SHAs in DEPS. We periodically sync to what ANGLE uses, but it has been a while

juan-lunarg: Yes but is there a need to download clang-format given we don't need it?

mikes-lunarg: Nope! These are still worthwhile improvements. I will try to roll the dependencies after it gets merged.

ncesario-lunarg (approved):

Issue #5135, opened 2023-01-19
wrong dynamic state printed for missing blend constants


zmike:
I get this error:

```
VUID-vkCmdDrawMultiIndexedEXT-commandBuffer-02701(ERROR / SPEC): msgNum: -277019797 - Validation Error: [ VUID-vkCmdDrawMultiIndexedEXT-commandBuffer-02701 ] Object 0: handle = 0x46a67b0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xef7d036b | vkCmdDrawMultiIndexedEXT: VK_DYNAMIC_STATE_SCISSOR state not set for this command buffer. The Vulkan spec states: If the VkPipeline object bound to the pipeline bind point used by this command requires any dynamic state, that state must have been set or inherited (if the VK_NV_inherited_viewport_scissor extension is enabled) for commandBuffer, and done so after any previously bound pipeline with the corresponding state not specified as dynamic (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdDrawMultiIndexedEXT-commandBuffer-02701)
```

with this backtrace

```
#0  zink_error (msg=0x51040c0 "Validation Error: [ VUID-vkCmdDrawMultiIndexedEXT-commandBuffer-02701 ] Object 0: handle = 0x46a67b0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xef7d036b | vkCmdDrawMultiIndexedEXT: VK_DYNA"...) at ../src/gallium/drivers/zink/zink_screen.c:1754
#1  0x00007ffff3a0b767 in zink_debug_util_callback (messageSeverity=VK_DEBUG_UTILS_MESSAGE_SEVERITY_ERROR_BIT_EXT, messageType=2, pCallbackData=0x7fffffff91e0, pUserData=0x0) at ../src/gallium/drivers/zink/zink_screen.c:1781
[...]

PR #5176, opened 2023-01-27
layers: Relabel new Dynamic State VUIDs

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5135

These VUID were added in Vulkan 1.3.240 spec, they replace the old, overloaded, `VUID-vkCmdDraw-commandBuffer-02701` VUID

cc @mikes-lunarg

### DIFF
--- a/layers/core_checks/drawdispatch_validation.cpp
+++ b/layers/core_checks/drawdispatch_validation.cpp
@@ -46,7 +46,6 @@
 struct DispatchVuidsCmdDraw : DrawDispatchVuid {
     DispatchVuidsCmdDraw() : DrawDispatchVuid() {
         pipeline_bound                     = "VUID-vkCmdDraw-None-02700";
-        dynamic_state                      = "VUID-vkCmdDraw-commandBuffer-02701";
         vertex_binding                     = "VUID-vkCmdDraw-None-04007";
         vertex_binding_null                = "VUID-vkCmdDraw-None-04008";
         compatible_pipeline                = "VUID-vkCmdDraw-None-02697";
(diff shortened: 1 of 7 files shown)


layers/core_checks/drawdispatch_validation.cpp
```diff
@@ -150,6 +149,17 @@ struct DispatchVuidsCmdDraw : DrawDispatchVuid {
         dynamic_coverage_reduction_mode    = "VUID-vkCmdDraw-None-07647";
         dynamic_representative_fragment_test_enable  = "VUID-vkCmdDraw-None-07648";
         dynamic_shading_rate_image_enable  = "VUID-vkCmdDraw-None-07649";
+        dynamic_viewport                   = "VUID-vkCmdDraw-None-07832";
```
artem-lunarg, line 152: Should  be VUID-vkCmdDraw-None-07831

layers/core_checks/drawdispatch_validation.cpp
```diff
@@ -150,6 +149,17 @@ struct DispatchVuidsCmdDraw : DrawDispatchVuid {
...
+        dynamic_scissor                    = "VUID-vkCmdDraw-None-07831";
```
artem-lunarg, line 153: And here VUID-vkCmdDraw-None-07832

layers/core_checks/drawdispatch_validation.cpp
spencer-lunarg: yes, good call, mixed the scissor/viewport 🤦, will fix

artem-lunarg (approved):

juan-lunarg:
> These VUID were added in Vulkan 1.3.240 spec

Does this not require a header update?

juan-lunarg (approved):

PR #5181, opened 2023-01-27
build: Update to header 1.3.240

mikes-lunarg:
- Update known-good
- Generate source

All of the VUID updates are already in #5176


juan-lunarg:
Rebasing on main should address the CI failures now

juan-lunarg (approved):

ncesario-lunarg (approved):

PR #5182, opened 2023-01-27
CMake Cleanup

juan-lunarg:


CMakeLists.txt
```diff
@@ -32,10 +32,7 @@ set(CMAKE_VISIBILITY_INLINES_HIDDEN "YES")
 
 include(GNUInstallDirs)
 
-# Enable GUI folders
-set_property(GLOBAL PROPERTY USE_FOLDERS ON)
-# "Helper" targets that don't have interesting source code should set their FOLDER property to this
-set(LAYERS_HELPER_FOLDER "Helper Targets")
+set_property(GLOBAL PROPERTY USE_FOLDERS ON) # Remove when min is 3.26, see CMP0143
```
juan-lunarg, line 35: Been meaning to address since I was the one who added this policy to CMake.
https://cmake.org/cmake/help/git-stage/policy/CMP0143.html?highlight=cmp0143

BUILD.md
```diff
@@ -160,6 +160,18 @@ directories and place them in any location (see option `--dir` in the
...
 
+### Warnings as errors off by default!
+
+By default `BUILD_WERROR` is `OFF`
+
+The idiom for open source projects is to NOT enable warnings as errors.
+
+System package managers, and language package managers have to build on multiple different platforms and compilers.
+
+By defaulting to `ON` we cause issues for package managers since there is no standard way to disable warnings until CMake 3.24
```
juan-lunarg, line 171: Reference to 3.24 changes:

- https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-compile-no-warning-as-error
- https://cmake.org/cmake/help/latest/prop_tgt/COMPILE_WARNING_AS_ERROR.html#prop_tgt:COMPILE_WARNING_AS_ERROR
- https://cmake.org/cmake/help/latest/variable/CMAKE_COMPILE_WARNING_AS_ERROR.html#variable:CMAKE_COMPILE_WARNING_AS_ERROR

CMakeLists.txt
```diff
@@ -112,6 +107,8 @@ if(${CMAKE_CXX_COMPILER_ID} MATCHES "(GNU|Clang)")
                         -Wno-unused-parameter
                         -Wno-missing-field-initializers)
 elseif(MSVC)
+    # TODO: Update to /W4
+    add_compile_options("/W3")
```
juan-lunarg, line 111: This is a regression caused by updating the cmake min to 3.17 since CMake used to add this compiler flag

scripts/update_deps.py
```diff
@@ -487,16 +487,14 @@ def CMakeConfig(self, repos):
...
     def CMakeBuild(self):
         """Build CMake command for the build phase and execute it"""
-        cmake_cmd = ['cmake', '--build', self.build_dir, '--target', 'install']
+        cmake_cmd = ['cmake', '--build', self.build_dir, '--target', 'install', '--config', CONFIG_MAP[self._args.config]]
         if self._args.do_clean:
             cmake_cmd.append('--clean-first')
 
-        # Multi-config generators need this specified, it's a no-op on single-config generators
-        cmake_cmd.append('--config {}'.format(CONFIG_MAP[self._args.config]))
-
```
juan-lunarg, line 496: IE this fine with newer cmake version, but not 3.17 (Not entirely sure why)

ncesario-lunarg (approved):

CMakeLists.txt
ncesario-lunarg: From a quick test it looks like `/W4` will require a non-trivial amount of changes.

juan-lunarg: It does look non-trivial but there may be valuable warnings worth adding to the build. Also we don't have to enable all the /W4 warnings, we can disable some.

PR #5183, opened 2023-01-27
syncval Revert "Apply render pass final transitions always"

juan-lunarg:
The change as written causes CTS crashes, as it doesn't make changes to the consumers of the altered data structures.

This reverts commit 41709ca1e50273aa7452fdb04f0c03cde0e53a01.


juan-lunarg:
Makes this annoying UI go away

image

juan-lunarg:
https://stackoverflow.com/questions/71162762/how-to-get-rid-of-githubs-compare-pull-request-suggestion

PR #5188, opened 2023-01-30
build: Add missing std header

artem-lunarg:


spencer-lunarg (approved):
How is/was this building?

artem-lunarg:
> How is/was this building?

It looks so far this header was treated in a special way by manually putting it in _compilable_ place.

juan-lunarg (approved):

PR #5191, opened 2023-01-30
layers: Fix formatting of zero flags

artem-lunarg:
The `string_***FlagBits()` family of functions format zero _***FlagBits_  variables as `"Unhandled FlagBits"` because zero is not a valid flag.
But zero is a valid value of _***Flags_ variable.

Previous Flags formatting:
"flags (Unhandled VkImageCreateFlags)"
New Flags formatting:
"flags (VkImageCreateFlags(0))"

The formatting of FlagBits remains unchanged.


juan-lunarg (approved):
:100:

spencer-lunarg (approved):

PR #5192, opened 2023-01-30
gn: Add Ninja to DEPS

juan-lunarg:


mikes-lunarg (approved):
:ninja:

juan-lunarg:
Fixes this build error:

>  depot_tools/ninja.py: Could not find Ninja in the third_party of the current project, nor in your PATH.
Please take a following action to install Ninja.

In the past the chromium build implicitly used an old ninja binary. Not anymore. Kinda my fault since the chromium builds did warn about this. I just ignored it...

> depot_tools/ninja.py: Fallback to a deprecated legacy ninja binary. Note that this ninja binary will be removed soon.
> Please install ninja to your project using DEPS. If your project does not have DEPS, Please install ninja in your PATH.
> See also https://crbug.com/1340825
> Cloning Chromium depot_tools

artem-lunarg (approved):

Issue #5167, opened 2023-01-25
VVL to check descriptor set allocation requesting types missed from pool creation


zhangyiwei:
The context is in https://gitlab.khronos.org/vulkan/vulkan/-/issues/3347. I wonder if we can do something better than best practice, since this is indeed causing app bailing on MESA Virtio-GPU Venus implementation.

spencer-lunarg:
Reopen to look into making this part of normal validation if `VK_ERROR_OUT_OF_POOL_MEMORY` is returned

spencer-lunarg:
@zhangyiwei sorry this got lost for so long, but finally fixed this so now by default it will give a warning if we detect `VK_ERROR_OUT_OF_POOL_MEMORY` was returned **and** the user used a type that wasn't in the pool

PR #5196, opened 2023-01-31
bp: Add DescriptorTypeNotInPool support

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5167 (cc @zhangyiwei)

This will check, for example, if the `VkDescriptorPool` was created `VkDescriptorType` of `{SAMPLER, STORAGE_BUFFER}` but then when calling `vkAllocateDescriptorSets` with a `VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER` it will throw a best practice warning (if using `VK_KHR_maintenance1`/Vulkan 1.1)

### DIFF
--- a/layers/best_practices/best_practices_utils.cpp
+++ b/layers/best_practices/best_practices_utils.cpp
@@ -678,6 +678,24 @@ bool BestPractices::PreCallValidateAllocateDescriptorSets(VkDevice device, const
                                    pAllocateInfo->descriptorSetCount, report_data->FormatHandle(pool_state->Handle()).c_str(),
                                    pool_state->GetAvailableSets());
             }
+
+            for (uint32_t i = 0; i < pAllocateInfo->descriptorSetCount; i++) {
+                auto layout = Get<cvdescriptorset::DescriptorSetLayout>(pAllocateInfo->pSetLayouts[i]);
+                if (layout) {  // if null, this is validated/logged in object_tracker
+                    const uint32_t binding_count = layout->GetBindingCount();
+                    for (uint32_t j = 0; j < binding_count; ++j) {
+                        const VkDescriptorType type = layout->GetTypeFromIndex(j);
+                        if (!pool_state->IsAvailableType(type)) {
+                            // This check would be caught by validation if VK_KHR_maintenance1 was not enabled
+                            skip |= LogWarning(pool_state->Handle(), kVUID_BestPractices_DescriptorTypeNotInPool,
+                                               "vkAllocateDescriptorSets(): pSetLayouts[%" PRIu32 "] binding %" PRIu32
+                                               " was created with %s but the "
+                                               "Descriptor Pool was not created with this type",
+                                               i, j, string_VkDescriptorType(type));
+                        }
+                    }
+                }
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


layers/best_practices/best_practices_utils.cpp
```diff
@@ -678,6 +678,24 @@ bool BestPractices::PreCallValidateAllocateDescriptorSets(VkDevice device, const
...
             }
+
+            for (uint32_t i = 0; i < pAllocateInfo->descriptorSetCount; i++) {
+                auto layout = Get<cvdescriptorset::DescriptorSetLayout>(pAllocateInfo->pSetLayouts[i]);
+                if (layout) {  // if null, this is validated/logged in object_tracker
+                    const uint32_t binding_count = layout->GetBindingCount();
+                    for (uint32_t j = 0; j < binding_count; ++j) {
+                        const VkDescriptorType type = layout->GetTypeFromIndex(j);
+                        if (!pool_state->IsAvailableType(type)) {
+                            // This check would be caught by validation if VK_KHR_maintenance1 was not enabled
```
artem-lunarg, line 689: Question to this comment. Is the mentioned validation Vulkan 1.0 only?

layers/state_tracker/descriptor_sets.h
```diff
@@ -70,6 +70,11 @@ class DESCRIPTOR_POOL_STATE : public BASE_NODE {
         return iter != available_counts_.end() ? iter->second : 0;
     }
 
+    bool IsAvailableType(uint32_t type) const {
+        auto guard = ReadLock();
+        return available_counts_.find(type) != available_counts_.end();
```
artem-lunarg, line 75: Additional safeguard might be to check if available count is actually not zero. If there's a guarantee that there can't be map entries with zero count then please ignore this comment.

layers/best_practices/best_practices_utils.cpp
spencer-lunarg: yes, from the spec source (`Vulkan-Docs/chapters/descriptorsets.adoc`)

```
ifndef::VK_VERSION_1_1,VK_KHR_maintenance1[]
  * [[VUID-VkDescriptorSetAllocateInfo-descriptorSetCount-00306]]
    pname:descriptorSetCount must: not be greater than the number of sets
    that are currently available for allocation in pname:descriptorPool
  * [[VUID-VkDescriptorSetAllocateInfo-descriptorPool-00307]]
    pname:descriptorPool must: have enough free descriptor capacity
    remaining to allocate the descriptor sets of the specified layouts
endif::VK_VERSION_1_1,VK_KHR_maintenance1[]
```

so in this case, the descriptor type not in the descriptor pool would hit one of these VUs

(side note: `ifndef::VK_VERSION_1_1,VK_KHR_maintenance1[]` ==  `VK_VERSION_1_1 OR VK_KHR_maintenance1`)

layers/state_tracker/descriptor_sets.h
spencer-lunarg: good question, `available_counts_` is a `layer_data::unordered_map<uint32_t, uint32_t>;` mapping (key == type, value == count)

The map is only created once at creation time, so it should have no be zero

above this function there is a `GetAvailableCount` that would return zero if not found, but I was afraid that would get confused with it just being out of valid slots

... after writing this, I realize I need to just add a comment, this isn't obvious

artem-lunarg (approved):

layers/best_practices/best_practices_utils.cpp
artem-lunarg: Thanks for explanation!

Issue #3797, opened 2022-03-02
use std instead of macro for limits


sfricke-samsung:
as pointed out by @ziga-lunarg in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/3758#discussion_r818054735 we should have a PR to update macro such as `UINT32_MAX` to use `std::numeric_limits`

jeremyg-lunarg:
Note that a while ago @ncesario-lunarg  added these helpers to vk_layer_data.h:

```
template <typename T>
constexpr T MaxTypeValue(T) {
    return std::numeric_limits<T>::max();
}

template <typename T>
constexpr T MaxTypeValue() {
    return std::numeric_limits<T>::max();
}

template <typename T>
constexpr T MinTypeValue(T) {
    return std::numeric_limits<T>::min();
}

template <typename T>
constexpr T MinTypeValue() {
    return std::numeric_limits<T>::min();
}
```

PR #5201, opened 2023-02-01
layers: Rename namespace layer_data to namespace vvl

juan-lunarg:
Been thinking about doing this for a while since `layer_data::` is very annoying to type.

This also helps make address https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/3797 more palatable for developers to use.

`layer_data::MaxTypeValue` vs `vvl::MaxTypeValue`
`layer_data::span` vs `vvl::span`


ncesario-lunarg (approved):
Makes sense to me!

PR #5205, opened 2023-02-01
layers: Remove macro constants for numeric limits

### LINKED ISSUE #3797: use std instead of macro for limits

juan-lunarg:
Originally I was going to use `std::numeric_limits` or `vvl::MaxTypeValue` everywhere. But this made the code less readable in certain areas.

Instead I created `constexpr` constants in the vvl namespace for the most common use cases in vvl:
```c++
// Typesafe UINT32_MAX
constexpr auto kU32Max = std::numeric_limits<uint32_t>::max();
// Typesafe UINT64_MAX
constexpr auto kU64Max = std::numeric_limits<uint64_t>::max();
// Typesafe INT32_MAX
constexpr auto kI32Max = std::numeric_limits<int32_t>::max();
// Typesafe INT64_MAX
constexpr auto kI64Max = std::numeric_limits<int64_t>::max();
```

Note how `vvl::kU64Max` is the same amount of characters as `UINT64_MAX`

Also these constants are available from the vvl namespace
image

closes #3797

### DIFF
--- a/layers/best_practices/best_practices_utils.cpp
+++ b/layers/best_practices/best_practices_utils.cpp
@@ -1709,7 +1709,7 @@ bool BestPractices::ValidateAccessLayoutCombination(const std::string& api_name,
                                                     VkImageAspectFlags aspect) const {
     bool skip = false;
 
-    const VkAccessFlags2 all = UINT64_MAX;  // core validation is responsible for detecting undefined flags.
+    const VkAccessFlags2 all = vvl::kU64Max;  // core validation is responsible for detecting undefined flags.
     VkAccessFlags2 allowed = 0;
 
     // Combinations taken from https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/2918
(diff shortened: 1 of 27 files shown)


artem-lunarg (approved):
LGTM!

`trashedViewportMask = vvl::MaxTypeValue(trashedViewportMask)` is a bit  unusual to read, maybe because it's the first time I found this and had to check details. I guess the intent is to make this more robust to changes.

spencer-lunarg (approved):

arno-lunarg:
What about using template variables:
```cpp
template <typename T>
constexpr auto kMax = std::numeric_limits<T>::max();

void foo() {
   auto bar = kMax<uint32_t>;
}
```
It is a bit longer though... It makes me wish we had `u32`, `u64`, etc... (★‿★)

juan-lunarg:
> trashedViewportMask = vvl::MaxTypeValue(trashedViewportMask) is a bit unusual to read, maybe because it's the first time I found this and had to check details. I guess the intent is to make this more robust to changes.

Agreed it's a bit unusual at first. But inferring the type from the parameter passed is powerful.

> It is a bit longer though... But then we have it for more types? It makes me wish we had u32, u64, etc... (★‿★)

I singled out U32_MAX, U64_MAX, I32_MAX, I64_MAX because they are often used when writing validation. So I wanted them as short as possible.

The rest of the types can use std::numeric_limits, MaxTypeValue, or MinTypeValue.

For `UINTPTR_MAX` I just used `std::numeric_limits<uintptr_t>::max()` given how there was only 1 usage of `UINTPTR_MAX`

Issue #5197, opened 2023-01-31
debugPrintfEXT: callback not invoked when using EXT mesh shaders


filnet:
**Environment:**
 - OS: Windows 10
 - GPU: NVIDIA GeForce RTX 2060
 - SDK or header version if building from repo: 1.3.239.0
 - Options enabled (synchronization, best practices, etc.):

**Describe the Issue**

Possibly after migrating to dynamic rendering (VK_KHR_dynamic_rendering), I noticed that turning off all but mesh shaders would stop `debugPrintfEXT` from producing output. Prior to that `debugPrintfEXT` was working fine.

Re-enabling the non mesh shaders solves the issue.
I also noticed that simply adding an utterly invalid `cmd_draw(command_buffer, 0, 0, 0, 0)` call causes the debug output to be restored. 

So it seems that mesh shaders only pipelines (possibly in addition to using dynamic rendering) causes debugPrintfEXT to stop working.

**Expected behavior**

`debugPrintfEXT` should work even if only mesh shaders are used.

filnet:
I am not 100% sure the issue is related to the switch to dynamic rendering, or, I forgot to mention, to the migration from the `VK_NV_mesh_shader` to `VK_NV_mesh_shader`.

I noticed the issue a few weeks after the mentioned changes...

Tony-LunarG:
We have a test that tests debug printf in a mesh shader called VkDebugPrintfTest.MeshTaskShadersPrintf.  As written, it does debug printfs from a task shader and a mesh shader.  If I comment out use of the task shader, I still get the message back from the mesh shader, so it seems there's more to it than using only a mesh shader.  It's hard to imagine how dynamic rendering would affect debug  printfs.  What draw call are you using to execute the mesh shader?  Are you certain that your mesh shader is actually executing when you don't get the debug printf?  Could you provide a small program that reproduces the issue, or modify MeshTaskShadersPrintf to demonstrate what you're seeing?

filnet:
I believe my description of the issue was misleading...
Yes, I am pretty sure too that debugPrintfEXT in the shader works well and produces output.

 The issue, if there is one, is on the validation side: the buffer is apparently not read on the CPU and the debug callback is not invoked. 
In that light, dynamic rendering or VK_EXT_mesh_shader could play a role. I don't know what trigger the CPU to read the buffer.

Tony-LunarG:
[...]

PR #5206, opened 2023-02-01
gpu: Add state tracking to GPU-AV and debug printf for EXT mesh_shader

filnet:
fixes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5197


ncesario-lunarg (commented):
LGTM, but needs a signoff from @Tony-LunarG.

PR #5211, opened 2023-02-02
layers: Fix printing dynamic states

spencer-lunarg:
I realized the REAL issue with https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5135 was the printing logic was wrong, before this PR it was

>  vkCmdDraw: VK_DYNAMIC_STATE_VIEWPORT|VK_DYNAMIC_STATE_SCISSOR|VK_DYNAMIC_STATE_LINE_WIDTH|VK_DYNAMIC_STATE_BLEND_CONSTANTS state not set for this command buffer

but not correctly is

> vkCmdDraw: VK_DYNAMIC_STATE_COLOR_BLEND_EQUATION_EXT state not set for this command buffer

### DIFF
--- a/layers/core_checks/cmd_buffer_dynamic_validation.cpp
+++ b/layers/core_checks/cmd_buffer_dynamic_validation.cpp
@@ -50,7 +50,7 @@ bool CoreChecks::ValidateCBDynamicStatus(const CMD_BUFFER_STATE &cb_state, CBDyn
                                          const char *msg_code) const {
     if (!(cb_state.status[status])) {
         return LogError(cb_state.commandBuffer(), msg_code, "%s: %s state not set for this command buffer.",
-                        CommandTypeString(cmd_type), DynamicStateString(status).c_str());
+                        CommandTypeString(cmd_type), DynamicStateToString(status));
     }
     return false;
 }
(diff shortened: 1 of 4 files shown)


layers/core_checks/cmd_buffer_dynamic_validation.cpp
```diff
@@ -50,7 +50,7 @@ bool CoreChecks::ValidateCBDynamicStatus(const CMD_BUFFER_STATE &cb_state, CBDyn
                                          const char *msg_code) const {
     if (!(cb_state.status[status])) {
         return LogError(cb_state.commandBuffer(), msg_code, "%s: %s state not set for this command buffer.",
-                        CommandTypeString(cmd_type), DynamicStateString(status).c_str());
+                        CommandTypeString(cmd_type), DynamicStateToString(status));
```
artem-lunarg, line 53: You had to read signatures carefully to see whether c_str() is needed. Maybe tools can help to write this correctly, plus compiler warnings (but not on Windows! CI caught few bugs in the last few days in my changes because of missed c_str()). But from a review or reading code standpoint, if `c_str()` were missing, I would not notice with high probability. Would it be a good idea to name `const char*` functions in a way that gives an explicit hint it's a `const char*` version? For example, we can have `DynamicStateToChar` (or `DynamicStateToConstChar` or `DynamicStateToCString`) for `const char*` and `DynamicStatesToString` for `std::string`.

artem-lunarg (approved):

layers/core_checks/cmd_buffer_dynamic_validation.cpp
arno-lunarg: +1 for this comment!

spencer-lunarg: the one `s` is because one takes only one flag and the other takes a flag mask

I didn't like the `const char*` vs `std::string` but having them same would be better

spencer-lunarg: for sake of getting this fixed, will make a follow up PR with other related `const char*` vs `std::string` fixes I wanted to do as well

PR #5214, opened 2023-02-02
Fix no longer expected error in DescriptorBufferNotEnabled

Tony-LunarG:


tests/vklayertests_descriptor_renderpass_framebuffer.cpp
```diff
@@ -12375,7 +12375,6 @@ TEST_F(VkLayerTest, DescriptorBufferNotEnabled) {
 
             if (descriptor_buffer_properties.bufferlessPushDescriptors) {
                 m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkBufferCreateInfo-usage-08102");
-                m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkBufferCreateInfo-usage-08103");
```
Tony-LunarG, line 12378: 08103 says that if you use VK_BUFFER_USAGE_PUSH_DESCRIPTORS_DESCRIPTOR_BUFFER_BIT_EXT, you have to also have VK_BUFFER_USAGE_RESOURCE_DESCRIPTOR_BUFFER_BIT_EXT or VK_BUFFER_USAGE_SAMPLER_DESCRIPTOR_BUFFER_BIT_EXT  which is clearly the case on line 12380

juan-lunarg (approved):

PR #5215, opened 2023-02-03
tests: Load libvulkan.so.1 instead of libvulkan.so

ncesario-lunarg:


layers/generated/lvt_function_pointers.cpp
```diff
@@ -349,7 +349,7 @@ void InitDispatchTable() {
 #elif(__APPLE__)
     const char filename[] = "libvulkan.dylib";
 #else
-    const char filename[] = "libvulkan.so";
+    const char filename[] = "libvulkan.so.1";
```
juan-lunarg, line 352: I ended up checking what vulkan.hpp does. Seems this is the most robust approach:

```
#  if defined( __unix__ ) || defined( __QNXNTO__ ) || defined( __Fuchsia__ )
        m_library = dlopen( "libvulkan.so", RTLD_NOW | RTLD_LOCAL );
        if ( m_library == nullptr )
        {
          m_library = dlopen( "libvulkan.so.1", RTLD_NOW | RTLD_LOCAL );
        }
```

Or we could just use Vulkan.hpp's DynamicLoader

juan-lunarg (changes requested):

juan-lunarg (approved):

Issue #5213, opened 2023-02-02
Remove unnecessary null pointer checks


elfring:
[An extra null pointer check is not needed in functions](https://isocpp.org/wiki/faq/freestore-mgmt#delete-handles-null "Do I need to check for null before delete p?") like the following.
- [SetCode](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/1f8c41e0c006b8efd8c43719103f322689d3368d/scripts/helper_file_generator.py#L577 "SetCode function")
- [VkGpuAssistedLayerTest](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/1f8c41e0c006b8efd8c43719103f322689d3368d/tests/vklayertests_gpu.cpp#L466 "Test case")

Issue #5212, opened 2023-02-02
Completion of error handling


elfring:
:eyes: It can be determined by [the means of static source code analysis](https://en.wikipedia.org/wiki/List_of_tools_for_static_code_analysis#C.2C_C.2B.2B "Tools for code analysis") that some return values from function calls are not used for desirable condition checks.

:thought_balloon: Would you like to add more error handling at places like the following?
* [fseek](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fseek.html "Reposition a file-position indicator") ⇒ [VkTestFramework::ReadFileData](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/efc0e5693abfef9ac7eff397240a194a3cd3b82b/tests/vktestframework.cpp#L693)
* [malloc](https://pubs.opengroup.org/onlinepubs/9699919799/functions/malloc.html "Memory allocation") ⇒ [DummyAlloc](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/efc0e5693abfef9ac7eff397240a194a3cd3b82b/tests/vklayertests_instanceless.cpp#L189)

elfring:
[Further update candidates were left over](https://github.com/spencer-lunarg/Vulkan-ValidationLayers/blob/480fa7c89d569263a09626c1862dfddc3c8d5fb7/tests/vktestframework.cpp#L719 "Update candidate: VkTestFramework::ReadFileData()").

:crystal_ball: Will code adjustments follow accordingly?

elfring:
* :thinking: How do you think about to improve [static source code analysis](https://en.wikipedia.org/wiki/List_of_tools_for_static_code_analysis#C.2C_C.2B.2B "Tools for code analysis") also for this software?
* :crystal_ball: How will chances evolve to benefit any more also from [the means of aspect-oriented software development](https://en.wikipedia.org/wiki/Aspect-oriented_programming "Documentation")?

PR #5217, opened 2023-02-03
layers: Improve C/C++ std error checking

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5213
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5212

### DIFF
--- a/tests/vklayertests_instanceless.cpp
+++ b/tests/vklayertests_instanceless.cpp
@@ -189,6 +189,10 @@ TEST_F(VkLayerTest, InstanceBadValidationFlags) {
 void* VKAPI_PTR DummyAlloc(void*, size_t size, size_t alignment, VkSystemAllocationScope) {
     size_t space = size + alignment - 1;
     void* mem_ptr = std::malloc(space);
+    if (!mem_ptr) {
+        printf("Dummy Alloc failed\n");
+        return nullptr;
+    }
     return std::align(alignment, size, mem_ptr, space);
 }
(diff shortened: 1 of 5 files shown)


artem-lunarg (approved):

juan-lunarg (approved):

tests/vklayertests_instanceless.cpp
```diff
@@ -189,6 +189,10 @@ TEST_F(VkLayerTest, InstanceBadValidationFlags) {
 void* VKAPI_PTR DummyAlloc(void*, size_t size, size_t alignment, VkSystemAllocationScope) {
     size_t space = size + alignment - 1;
     void* mem_ptr = std::malloc(space);
+    if (!mem_ptr) {
+        printf("Dummy Alloc failed\n");
```
elfring, line 193: :thought_balloon: It can be nicer to use the function “perror”, can't it?

jeremyg-lunarg: We do accept pull requests.

tests/vktestframework.cpp
```diff
@@ -690,7 +690,10 @@ char **VkTestFramework::ReadFileData(const char *fileName) {
 
     while (fgetc(in) != EOF) count++;
```
elfring, line 691: :thought_balloon: Is the error detection and corresponding still improvable also according to this source code?

elfring (commented):
Why were changes for different issues mixed in such a commit?

PR #5220, opened 2023-02-03
layers: Fix Dynamic State for Color attachments

spencer-lunarg:
`VK_DYNAMIC_STATE_COLOR_BLEND_ENABLE_EXT`, `VK_DYNAMIC_STATE_COLOR_BLEND_EQUATION_EXT`, `VK_DYNAMIC_STATE_COLOR_WRITE_MASK_EXT`, and `VK_DYNAMIC_STATE_COLOR_BLEND_ADVANCED_EXT`

all have a `firstAttachment` and `attachmentCount` in the `vkCmdSet*()` command

There were two VUs, going to get a spec fix to remove the one that doesn't make sure all attachments are checked

Same issue as https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4997/files had


layers/state_tracker/state_tracker.cpp
```diff
@@ -5388,20 +5389,29 @@ void ValidationStateTracker::PostCallRecordCmdSetColorBlendEnableEXT(VkCommandBu
                                                                      uint32_t attachmentCount, const VkBool32 *pColorBlendEnables) {
     auto cb_state = GetWrite<CMD_BUFFER_STATE>(commandBuffer);
     cb_state->RecordStateCmd(CMD_SETCOLORBLENDENABLEEXT, CB_DYNAMIC_COLOR_BLEND_ENABLE_EXT_SET);
+    for (uint32_t i = 0; i < attachmentCount; i++) {
+        cb_state->dynamic_state_value.color_blend_enable_attachments.set(firstAttachment + i);
```
juan-lunarg, line 5393: Where do the bits get `reset()`? I only see them get `set` and `tested`

spencer-lunarg: ok, so thanks for pointing this out, reading the spec

> dynamic state can be modified at any time and persists for the lifetime of the command buffer,

So I realized a `vkResetCommandBuffer` or `vkResetCommandPool` should reset this

spencer-lunarg: ok, so figured out the true issue here, wrote tests and added comment where we reset these `bitset`

juan-lunarg (approved):

PR #5221, opened 2023-02-03
layers: Add vuid hash testing

juan-lunarg:
Ensure we don't have any VUID hash collisions
Ensure we don't accidentally break hash stability

NOTE: I had to fix the Android build to make the testing work. Android fix is in a separate commit.


juan-lunarg:
@johnkslang / @dneto0

mark-lunarg (approved):
LGTM.

PR #5228, opened 2023-02-06
build: Fix BUILD.gn

y-novikov:
Missing valid_param_values.h source

Fixes #4228

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -242,6 +242,7 @@ chassis_sources = [
   "$vulkan_headers_dir/include/vulkan/vulkan.h",
   "layers/generated/chassis.cpp",
   "layers/generated/valid_param_values.cpp",
+  "layers/generated/valid_param_values.h",
   "layers/generated/chassis.h",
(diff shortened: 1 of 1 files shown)


juan-lunarg (approved):
We run the gn build as part of CI. Any suggestions to prevent this from happening again?

y-novikov:
> We run the gn build as part of CI. Any suggestions to prevent this from happening again?

To clarify - this isn't a problem for GN build.
There is "gn check" command, but it doesn't report any error, so I don't think you can do something on your part.

ANGLE has additional tests in
https://source.chromium.org/chromium/chromium/src/+/main:third_party/angle/scripts/export_targets.py
I think they are for exporting ANGLE into Firefox.
So, I think it's OK for ANGLE folks to fix your BUILD.gn to pass our tests once in a while.

juan-lunarg:
> So, I think it's OK for ANGLE folks to fix your BUILD.gn to pass our tests once in a while.

:+1:

PR #5229, opened 2023-02-06
cmake: Apply -Wno-shorten-64-to-32 to tests only

juan-lunarg:
Layer code doesn't need this warning disabled.


arno-lunarg (approved):
LGTM!

artem-lunarg (approved):

PR #5231, opened 2023-02-07
Vuid exernal memory 00656

artem-lunarg:
Commit 1. Clean up and fixes of existing tests to make sure they will work with VU 00656.
Commit 2 and 3. Implement and test VUID 00656.

Fixes one item from https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4539

VU 00656 is about external memory handle types associated with a _memory object_. The type of the handles should be supported and compatible as defined in the spec. Previously similar VUs 00920/00990 were implemented and they are concerned with the handle types associated with the _resource_ itself (buffer or image).

### DIFF
--- a/layers/core_checks/device_memory_validation.cpp
+++ b/layers/core_checks/device_memory_validation.cpp
@@ -545,18 +545,63 @@ bool CoreChecks::ValidateBindBufferMemory(VkBuffer buffer, VkDeviceMemory mem, V
             }
 
             // Validate export memory handles
-            if ((mem_info->export_handle_type_flags != 0) &&
-                ((mem_info->export_handle_type_flags & buffer_state->external_memory_handle) == 0)) {
-                const char *vuid =
-                    bind_buffer_mem_2 ? "VUID-VkBindBufferMemoryInfo-memory-02726" : "VUID-vkBindBufferMemory-memory-02726";
-                const LogObjectList objlist(buffer, mem);
-                skip |= LogError(objlist, vuid,
+            if (mem_info->export_handle_type_flags != 0) {
+                auto external_buffer_info = LvlInitStruct<VkPhysicalDeviceExternalBufferInfo>();
+                external_buffer_info.flags = buffer_state->createInfo.flags;
+                external_buffer_info.usage = buffer_state->createInfo.usage;
+                auto external_buffer_properties = LvlInitStruct<VkExternalBufferProperties>();
+                bool export_supported = true;
+
+                // Check export operation support
+                auto check_export_support = [&](VkExternalMemoryHandleTypeFlagBits flag) {
+                    external_buffer_info.handleType = flag;
+                    DispatchGetPhysicalDeviceExternalBufferProperties(physical_device, &external_buffer_info,
+                                                                      &external_buffer_properties);
+                    if ((external_buffer_properties.externalMemoryProperties.externalMemoryFeatures &
+                         VK_EXTERNAL_MEMORY_FEATURE_EXPORTABLE_BIT) == 0) {
+                        export_supported = false;
+                        const LogObjectList objlist(buffer, mem);
+                        skip |= LogError(objlist, "VUID-VkExportMemoryAllocateInfo-handleTypes-00656",
+                                         "%s: The VkDeviceMemory (%s) has VkExportMemoryAllocateInfo::handleTypes with the %s flag "
+                                         "set, which does not support VK_EXTERNAL_MEMORY_FEATURE_EXPORTABLE_BIT with the buffer "
+                                         "create flags (%s) and usage flags (%s).",
+                                         api_name, report_data->FormatHandle(mem).c_str(),
+                                         string_VkExternalMemoryHandleTypeFlagBits(flag),
+                                         string_VkBufferCreateFlags(external_buffer_info.flags).c_str(),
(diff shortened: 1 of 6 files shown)


layers/core_checks/device_memory_validation.cpp
```diff
@@ -567,18 +567,60 @@ bool CoreChecks::ValidateBindBufferMemory(VkBuffer buffer, VkDeviceMemory mem, V
...
+                    external_buffer_info.handleType = flag;
+                    DispatchGetPhysicalDeviceExternalBufferProperties(physical_device, &external_buffer_info,
+                                                                      &external_buffer_properties);
+                    if ((external_buffer_properties.externalMemoryProperties.externalMemoryFeatures &
+                         VK_EXTERNAL_MEMORY_FEATURE_EXPORTABLE_BIT) == 0) {
+                        export_supported = false;
+                        const LogObjectList objlist(buffer, mem);
+                        skip |= LogError(objlist, "VUID-VkExportMemoryAllocateInfo-handleTypes-00656",
+                                         "%s: The VkDeviceMemory (%s) has an external handleType flag %s which does not support "
+                                         "export operation. The handleTypes flags are specified in the VkExportMemoryAllocateInfo "
```
spencer-lunarg, line 588: > The handleTypes flags are specified in the VkExportMemoryAllocateInfo structure included in the pNext chain of VkMemoryAllocateInfo.

If they are getting this error, they already are using `VkExportMemoryAllocateInfo` in the pNext correctly. I would maybe do

> The VkDeviceMemory (%s) has an VkExportMemoryAllocateInfo::handleType flag %s which does not support  VK_EXTERNAL_MEMORY_FEATURE_EXPORTABLE_BIT with the buffer create flags (%s) and usage flag (%s)

spencer-lunarg (approved):
Just some error message cleanup suggestions, otherwise looks good to me

layers/core_checks/device_memory_validation.cpp
artem-lunarg: >The handleTypes flags are specified in the VkExportMemoryAllocateInfo structure included in the pNext chain of VkMemoryAllocateInfo.

Validation happens during `BindMemory` call, but wrong flags are defined during the `AllocateMemory` call. The idea was that for the user, it might be a bit confusing where we get `handleTypes` during the `BindMemory` cal, that's why the additional information. Maybe I overthought it.

Anyway, I like your version more. It's concise and provides all information.

arno-lunarg (commented):
LGTM!

Rua:
I'm not sure if this is correct? The VUID is for the `VkExportMemoryAllocateInfo` struct, and applies when allocating memory, but the code here seems to have been applied to `vkBindBufferMemory` and `vkBindImageMemory`.

spencer-lunarg:
@Rua you are right that something seems wrong here... Maybe this is a Spec bug, the `00656` VU needs to know the Image/Buffer details before validating, so the VU can only be changed at AllocateMemory time IF there is a dedicated memory to it

artem-lunarg:
Yes, validation is possible only at the binding point. On the other hand the application writer might have all the information when `VkExportMemoryAllocateInfo` is initialized and from API usage standpoint this VU makes sense. Is the some conventions about the point when validation is performed?

Rua:
> @Rua you are right that something seems wrong here... Maybe this is a Spec bug, the `00656` VU needs to know the Image/Buffer details before validating, so the VU can only be changed at AllocateMemory time IF there is a dedicated memory to it

This VUID always seemed strange to me. I'll admit that's partially why I asked for it to be added to the validation, to see how people much smarter than me would deal with it! :D It could be an oversight in the spec, but it wouldn't be the first time that the spec is deliberately weird about stuff like that.

spencer-lunarg:
(taking notes for myself as I am assigned to fix the Spec for this)

`VkLayerTest.BufferMemoryWithUnsupportedExternalHandleType` and `VkLayerTest.ImageMemoryWithUnsupportedExternalHandleType` fail on SwiftShader wtih

> [ RUN      ] VkLayerTest.BufferMemoryWithUnsupportedExternalHandleType
/swiftshader/src/Vulkan/VkDeviceMemory.cpp:218 WARNING: UNSUPPORTED: extendedAllocationInfo->exportMemoryAllocateInfo->handleTypes 3
Vulkan-ValidationLayers/tests/vktestbinding.cpp:469: Failure
Failed
Assertion: `(result == VK_SUCCESS) || (result == VK_ERROR_VALIDATION_FAILED_EXT) || (result == VK_ERROR_OUT_OF_DEVICE_MEMORY) || (result == VK_ERROR_OUT_OF_HOST_MEMORY)'
Vulkan-ValidationLayers/tests/vkerrormonitor.cpp:103: Failure
Failed
Validation Error: [ UNASSIGNED-GeneralParameterError-RequiredParameter ] Object 0: handle = 0x55f1a717e200, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x99fb7dfd | vkBindBufferMemory: required parameter memory specified as VK_NULL_HANDLE
/home/fricke/github/Vulkan-ValidationLayers/tests/vkerrormonitor.cpp:144: Failure
Failed
Did not receive expected error 'VUID-VkExportMemoryAllocateInfo-handleTypes-00656'

PR #5236, opened 2023-02-07
tests: Return std::optional for QueueFamilyIndex

juan-lunarg:
Cleaner than checking UINT32_MAX, and less error prone due to stricter type checking.

### DIFF
--- a/tests/positive/image_buffer.cpp
+++ b/tests/positive/image_buffer.cpp
@@ -91,13 +91,13 @@ TEST_F(VkPositiveLayerTest, OwnershipTranfersImage) {
     TEST_DESCRIPTION("Valid image ownership transfers that shouldn't create errors");
     ASSERT_NO_FATAL_FAILURE(Init(nullptr, nullptr, VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT));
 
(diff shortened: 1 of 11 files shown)


tests/positive/image_buffer.cpp
```diff
@@ -2466,8 +2466,11 @@ TEST_F(VkPositiveLayerTest, CorrectSparseBufferOverlapCopy) {
     bind_info.signalSemaphoreCount = 1;
     bind_info.pSignalSemaphores = &semaphore;
 
-    uint32_t sparse_index = m_device->QueueFamilyMatching(VK_QUEUE_SPARSE_BINDING_BIT, 0u);
-    VkQueue sparse_queue = m_device->graphics_queues()[sparse_index]->handle();
```
artem-lunarg, line 2470: It's interesting how it worked before? or is sparse queue so ubiquitous that we never got uint32 max here...

artem-lunarg (approved):

tests/positive/image_buffer.cpp
juan-lunarg: This test does do a feature check for `sparseBinding` functionality at the top.

Which should be good enough. But still I'd rather be safe than sorry when it comes to vulkan drivers.

PR #5237, opened 2023-02-07
docs: Remove author section

juan-lunarg:
Arguments for removing the author section at the top of files is:
1.) It's not getting updated
2.) Has emails of devs who no longer work on vvl
3.) Takes up room
4.) If you need to figure out who wrote something you use git


spencer-lunarg (approved):
100% agree to all arguments and only left them because "I thought we had to" fear

PR #5244, opened 2023-02-08
layers: Remove VK_LAYER_EXPORT

### LINKED ISSUE #5243: Build broken with ToT Vulkan-Headers, due to removal of `VK_LAYER_EXPORT`
This commit, https://github.com/KhronosGroup/Vulkan-Headers/commit/e8b8e06d092ab406b097907ecaae1a8aae9c7d53, removed `VK_LAYER_EXPORT` from Vulkan-Headers, but VVL still contains references to it, so trying to use ToT of these repos together fails to build.

This was detected by the Chromium bots that roll these dependencies, https://ci.chromium.org/ui/p/chromium/builders/try/chromeos-arm-generic-rel/1136175/overview.
I don't think it is a blocking issue at the moment, so this is mostly an FYI.

juan-lunarg:
closes #5243

See https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5243#issuecomment-1422957585 & https://github.com/KhronosGroup/Vulkan-Headers/pull/377 for the full rational.

But the long and short of it is that the linux binary was only exporting 5 symbols while other platforms were exporting more than was necessary.

Here are the files that controls exports for linux/windows:
- https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/layers/VkLayer_khronos_validation.def
- https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/layers/libVkLayer_khronos_validation.map


layers/generated/chassis.cpp
```diff
@@ -15838,3 +15829,24 @@ VK_LAYER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL vkNegotiateLoaderLayerInterfaceVe
 
     return VK_SUCCESS;
 }
+
+#if defined(VK_USE_PLATFORM_ANDROID_KHR)
+#define VVL_ANDROID_EXPORT __attribute__((visibility("default")))
```
juan-lunarg, line 15834: `vkEnumerateDeviceLayerProperties` and `vkEnumerateDeviceExtensionProperties` only need to be exported for the Android build.

Otherwise CI will fail.

juan-lunarg: + @charles-lunarg for insight into Android loader? (IE I'd like to write a good comment here)

charles-lunarg: I would of thought that `vkEnumerateDeviceLayerProperties` and `vkEnumerateDeviceExtensionProperties` would just need to be added to the version script, since android behaves like linux most of the time.

https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/layers/libVkLayer_khronos_validation.map

However, it appears that the android loader uses an 'older' version of the loader-layer interface which requires that those functions are exported, so it may be wise to have 2 version script files, one for desktop and one for android.

juan-lunarg: Originally had 2 version scripts but it got complicated to manage with the android/gn builds. So I decided on `VVL_ANDROID_EXPORT`.

ncesario-lunarg (commented):
Once comment/question, but this looks like some nice cleanup :)

layers/generated/chassis.cpp
juan-lunarg: They are used in multiple other sections of the code, but it's non-trivial without modifying a decent amount of python scripts.

juan-lunarg: How about this approach?
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5244/commits/dbc14f3ef4f57df4cd994c5ff304f7c3ed7cb05a

juan-lunarg: That is cleaner :+1:

spencer-lunarg (approved):

scripts/known_good.json
```diff
@@ -17,7 +17,7 @@
             "sub_dir": "Vulkan-Headers",
             "build_dir": "Vulkan-Headers/build",
             "install_dir": "Vulkan-Headers/build/install",
-            "commit": "v1.3.240"
+            "commit": "e8b8e06d092ab406b097907ecaae1a8aae9c7d53"
```
ncesario-lunarg, line 20: Necessary for this change?

ncesario-lunarg (approved):
LGTM!

scripts/known_good.json
juan-lunarg: This is the commit that removes VK_LAYER_EXPORT from vulkan-headers. So it matches what the GN folks are doing.

PR #5245, opened 2023-02-08
docs: Automate creation of CONTRIBUTORS.md

juan-lunarg:


ncesario-lunarg (approved):
LGTM! Do we have an example PR to look at?

juan-lunarg:
> LGTM! Do we have an example PR to look at?

Looking at the documentation once this is merged it should make a PR. If not I can force the action to run via github cli.

Here is the default template it will use unless we do something custom:

```
# Top Contributors: {{ repo }}
Shout out to our top contributors!

{{ content }}

_Last updated: {{ updated }}_
```

PR #5248, opened 2023-02-08
docs: Fix contributors.yml

juan-lunarg:
- Updates CONTRIBUTORS.md when needed
- Works with existing action workflow
- Shows all contributors/collaborators


juan-lunarg:
Also it allows more flexibility with how the contributing.md is laid out

juan-lunarg:
Here is an example of what it will look like:
https://github.com/akhilmhdh/contributors-readme-action/blob/master/contributors.md

ncesario-lunarg (approved):

PR #5261, opened 2023-02-09
ci: Use concurrency to prevent duplicate jobs for workflows

juan-lunarg:
GitHub docs for reference:
https://docs.github.com/en/actions/using-jobs/using-concurrency

By default GitHub actions doesn't cancel old runs of workflows.

With the new `concurrency` support you can cancel in progress runs now.

This makes it so you don't have to manually cancel old runs when you force push. IE saving CI/developer time.

### DIFF
--- a/.github/workflows/build_android.yml
+++ b/.github/workflows/build_android.yml
@@ -15,6 +15,11 @@
 
 name: Android
 
+# https://docs.github.com/en/actions/using-jobs/using-concurrency
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 on:
     push:
     pull_request:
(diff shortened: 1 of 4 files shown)


.github/workflows/build_android.yml
```diff
@@ -15,6 +15,11 @@
 
 name: Android
 
+# https://docs.github.com/en/actions/using-jobs/using-concurrency
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
juan-lunarg, line 21: Example:

github.workflow = `Android`
github.ref = `juaramos/actions`

If I make a new push/commit, cancel-in-progress runs so the new ones don't have to wait.

juan-lunarg: For testing I used my fork and force pushed quite a bit. Verifying that older runs were cancelled by the newer runs.

spencer-lunarg (approved):

PR #5262, opened 2023-02-10
layers: Add remaining VUIDs for wayland surface creation

### LINKED ISSUE #4790: Missing checks for VkWaylandSurfaceCreateInfoKHR's display and surface members
Was helping a user on the Vulkan Discord when they reported issues creating a swapchain. Turns out they were NOT specifying the `VkWaylandSurfaceCreateInfoKHR::surface` member in their creation of the wayland surface.

This is because there is no validation of either the wl_display and wl_surface members of VkWaylandSurfaceCreateInfoKHR, likely because they are *not* core Vulkan types, but are "wayland" types.

**Valid Usage IDs requested**
VUID-VkWaylandSurfaceCreateInfoKHR-display-01304
VUID-VkWaylandSurfaceCreateInfoKHR-surface-01305

**Additional Context**
This definitely should be an autogenerated check in the parameter validation checks, but seems that the code-gen doesn't deal with the WSI related types at all, which is a shame.

juan-lunarg:
- VUID-VkWaylandSurfaceCreateInfoKHR-display-01304
- VUID-VkWaylandSurfaceCreateInfoKHR-surface-01305

closes #4790

### DIFF
--- a/layers/stateless/parameter_validation_utils.cpp
+++ b/layers/stateless/parameter_validation_utils.cpp
@@ -5770,6 +5770,36 @@ bool StatelessValidation::manual_PreCallValidateCreateWin32SurfaceKHR(VkInstance
 }
 #endif  // VK_USE_PLATFORM_WIN32_KHR
 
+#ifdef VK_USE_PLATFORM_WAYLAND_KHR
+
+bool StatelessValidation::manual_PreCallValidateCreateWaylandSurfaceKHR(VkInstance instance,
+                                                                        const VkWaylandSurfaceCreateInfoKHR *pCreateInfo,
+                                                                        const VkAllocationCallbacks *pAllocator,
+                                                                        VkSurfaceKHR *pSurface) const {
+    bool skip = false;
+
+    const auto display = pCreateInfo->display;
+    const auto surface = pCreateInfo->surface;
+
+    // Beyond checking for nullptr not much we can do.
+    // EX: Calling wl_display_get_registry on an invalid display will crash the application.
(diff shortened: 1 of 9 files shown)


layers/stateless/parameter_validation_utils.cpp
```diff
@@ -5770,6 +5770,36 @@ bool StatelessValidation::manual_PreCallValidateCreateWin32SurfaceKHR(VkInstance
...
+        skip |= LogError(device, "VUID-VkWaylandSurfaceCreateInfoKHR-display-01304",
+                         "vkCreateWaylandSurfaceKHR: display must be valid Wayland wl_display but display is NULL!");
+    }
+
+    // Beyond checking for nullptr not much we can do.
+    // EX: Calling wl_surface_get_version on an invalid surface will crash the application.
+    if (surface == nullptr) {
+        skip |= LogError(device, "VUID-VkWaylandSurfaceCreateInfoKHR-surface-01305",
+                         "vkCreateWaylandSurfaceKHR: surface must be valid Wayland wl_surface but surface is NULL!");
+    }
```
juan-lunarg, line 5796: Any advice besides just checking if the user passed in `nullptr`?

spencer-lunarg (approved):

layers/stateless/parameter_validation_utils.cpp
spencer-lunarg: My only suggestion was to check it with a simple getter function, but you already listed that will not work, so I am satisfied with a null check

PR #5263, opened 2023-02-10
Add dynamic VK_EXT_line_rasterization

spencer-lunarg:
Adds the 4 draw time VUs

- `VUID-vkCmdDraw-stippledLineEnable-07495`
- `VUID-vkCmdDraw-stippledLineEnable-07496`
- `VUID-vkCmdDraw-stippledLineEnable-07497`
- `VUID-vkCmdDraw-stippledLineEnable-07498`

Fixes the Pipeline creation VUs (`VkGraphicsPipelineCreateInfo` and `VkPipelineRasterizationLineStateCreateInfoEXT`) to not check `lineRasterizationMode` or `stippledLineEnable` if they are dynamic state

### DIFF
--- a/layers/core_checks/drawdispatch_validation.cpp
+++ b/layers/core_checks/drawdispatch_validation.cpp
@@ -148,6 +148,10 @@ struct DispatchVuidsCmdDraw : DrawDispatchVuid {
         descriptor_buffer_set_offset_missing = "VUID-vkCmdDraw-None-08117";
         image_view_dim                     = "VUID-vkCmdDraw-viewType-07752";
         image_view_numeric_format          = "VUID-vkCmdDraw-format-07753";
+        stippled_rectangular_lines         = "VUID-vkCmdDraw-stippledLineEnable-07495";
+        stippled_bresenham_lines           = "VUID-vkCmdDraw-stippledLineEnable-07496";
+        stippled_smooth_lines              = "VUID-vkCmdDraw-stippledLineEnable-07497";
+        stippled_default_strict            = "VUID-vkCmdDraw-stippledLineEnable-07498";
     }
 };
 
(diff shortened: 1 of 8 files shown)


layers/core_checks/drawdispatch_validation.cpp
```diff
@@ -1535,6 +1571,10 @@ struct DispatchVuidsCmdDrawMeshTasksIndirectNV: DrawDispatchVuid {
         descriptor_buffer_set_offset_missing = "VUID-vkCmdDrawMeshTasksIndirectNV-None-08117";
         image_view_dim                     = "VUID-vkCmdDrawMeshTasksIndirectNV-viewType-07752";
         image_view_numeric_format          = "VUID-vkCmdDrawMeshTasksIndirectNV-format-07753";
+        stippled_rectangular_lines         = "VUID-vkCmdDraw-stippledLineEnable-07495";
```
artem-lunarg, line 1574: `VUID-vkCmdDrawMeshTasksIndirectNV` in the next four lines?

layers/stateless/parameter_validation_utils.cpp
```diff
@@ -3093,112 +3093,116 @@ bool StatelessValidation::manual_PreCallValidateCreateGraphicsPipelines(VkDevice
...
-                                             i);
+                            if (line_state->stippledLineEnable && !dynamic_line_stipple_enable) {
+                                if (line_state->lineRasterizationMode == VK_LINE_RASTERIZATION_MODE_RECTANGULAR_EXT &&
+                                    (!line_features || !line_features->stippledRectangularLines)) {
+                                    skip |= LogError(
+                                        device, "VUID-VkPipelineRasterizationLineStateCreateInfoEXT-stippledLineEnable-02771",
+                                        "vkCreateGraphicsPipelines(): pCreateInfos[%" PRIu32
+                                        "] lineRasterizationMode = "
+                                        "VK_LINE_RASTERIZATION_MODE_RECTANGULAR_EXT with stippledLineEnable requires the "
+                                        "stippledRectangularLines feature is not enabled.",
```
artem-lunarg, line 3173: Trying to undersand what's going on here. As I understood from the spec `strippledRectangularLines` must be enabled in this context. Then `is not enabled` part does not look correct or is it a typo and should be `which is not enabled`? The same question to other places below.

layers/core_checks/drawdispatch_validation.cpp
spencer-lunarg: yes... good catch

layers/stateless/parameter_validation_utils.cpp
spencer-lunarg: so the error message is wrong... should say "this is an error because stippledRectangularLines is not enabled"

will fix wording for all messages

artem-lunarg (approved):

ncesario-lunarg (approved):

Issue #2629, opened 2021-03-17
Missing runtime atomic feature validation check


sfricke-samsung:
**Describe the situation in which you encountered the missing validation**
Issue description: Currently there is no coverage of `VK_EXT_shader_atomic_float`, `VK_EXT_shader_atomic_float2`, `VK_EXT_shader_image_atomic_int64`, or `VK_KHR_shader_atomic_int64` feature bits to see if the SPIR-V is allowed to use the various forms of atomics

**Valid Usage IDs requested**
Please include the valid usage IDs for the checks you are requesting: (currently no VUID for runtime SPIR-V validation)

**Additional Context**
Any additional information: I have been slowly working on this with the Working Group to clarify various parts of the specifications around these atomic feature bits. I have some basic tests is written up and implementation ready. Creating issue to mark that this is being worked on and will be up when spec language is ironed out

Adding for search-ability
- [x] VK_KHR_shader_atomic_int64 (#3094)
    - [x]       shaderBufferInt64Atomics
    - [x]      shaderSharedInt64Atomics
- [x] VK_EXT_shader_image_atomic_int64
    - [x]      shaderImageInt64Atomics (#3125)
    - [x]       sparseImageInt64Atomics
- [x] VK_EXT_shader_atomic_float (#3209)
    - [x]       shaderBufferFloat32Atomics
    - [x]       shaderBufferFloat32AtomicAdd
    - [x]       shaderBufferFloat64Atomics
    - [x]       shaderBufferFloat64AtomicAdd
    - [x]       shaderSharedFloat32Atomics
    - [x]       shaderSharedFloat32AtomicAdd
    - [x]       shaderSharedFloat64Atomics
    - [x]       shaderSharedFloat64AtomicAdd
    - [x]       shaderImageFloat32Atomics
    - [x]       shaderImageFloat32AtomicAdd
    - [x]       sparseImageFloat32Atomics
    - [x]       sparseImageFloat32AtomicAdd
- [x] VK_EXT_shader_atomic_float2 (#3209)
    - [x]       shaderBufferFloat16Atomics
    - [x]       shaderBufferFloat16AtomicAdd
    - [x]       shaderBufferFloat16AtomicMinMax
    - [x]       shaderBufferFloat32AtomicMinMax
    - [x]       shaderBufferFloat64AtomicMinMax
    - [x]       shaderSharedFloat16Atomics
    - [x]       shaderSharedFloat16AtomicAdd
[...]

PR #5264, opened 2023-02-10
Add VK_EXT_shader_image_atomic_int64 draw VUs

spencer-lunarg:
Makes sure if there is a 64-bit `VkImageView` or 64-bit `OpTypeImage` that both are either 64-bit or 32-bit (only other valid option) 

(related issue https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/2629)


arno-lunarg (approved):
LGTM!

tests/vklayertests_pipeline_shader.cpp
```diff
@@ -11508,6 +11508,125 @@ TEST_F(VkLayerTest, ShaderImageAtomicInt64) {
...
+    SetTargetApiVersion(VK_API_VERSION_1_1);
+    AddRequiredExtensions(VK_EXT_SHADER_IMAGE_ATOMIC_INT64_EXTENSION_NAME);
+    ASSERT_NO_FATAL_FAILURE(InitFramework());
+    if (!AreRequiredExtensionsEnabled()) {
+        GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported";
+    }
+
+    auto image_atomic_int64_features = LvlInitStruct<VkPhysicalDeviceShaderImageAtomicInt64FeaturesEXT>();
+    auto features2 = GetPhysicalDeviceFeatures2(image_atomic_int64_features);
+    if (features2.features.shaderInt64 == VK_FALSE) {
```
artem-lunarg, line 11522: Is it a good idea to skip checking for features if the feature support is guaranteed by the extension?

tests/vklayertests_pipeline_shader.cpp
```diff
@@ -11508,6 +11508,125 @@ TEST_F(VkLayerTest, ShaderImageAtomicInt64) {
     }
 }
 
+TEST_F(VkLayerTest, ShaderImageAtomicInt64Drawtime64) {
```
artem-lunarg, line 11511: typo: ShaderImageAtomicInt32Drawtime64

spencer-lunarg: the extension is `VK_EXT_shader_image_atomic_int64` so that is why it is `ShaderImageAtomicInt64`

artem-lunarg: Ah, makes sense! I came up with some imaginary encoding...

PR #5266, opened 2023-02-10
VUIDs for external fence and semaphore

artem-lunarg:
That's almost identical to external memory support and compatibility for images/buffers, but now for fences and semaphores.
Some variable naming clean up.

Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4539

### DIFF
--- a/layers/core_checks/core_validation.h
+++ b/layers/core_checks/core_validation.h
@@ -1385,6 +1385,8 @@ class CoreChecks : public ValidationStateTracker {
     bool PreCallValidateAllocateMemory(VkDevice device, const VkMemoryAllocateInfo* pAllocateInfo,
                                        const VkAllocationCallbacks* pAllocator, VkDeviceMemory* pMemory) const override;
     bool PreCallValidateFreeMemory(VkDevice device, VkDeviceMemory mem, const VkAllocationCallbacks* pAllocator) const override;
+    bool PreCallValidateCreateFence(VkDevice device, const VkFenceCreateInfo* pCreateInfo, const VkAllocationCallbacks* pAllocator,
+                                    VkFence* pFence) const override;
     bool PreCallValidateCreateSemaphore(VkDevice device, const VkSemaphoreCreateInfo* pCreateInfo,
                                         const VkAllocationCallbacks* pAllocator, VkSemaphore* pSemaphore) const override;
     bool PreCallValidateWaitSemaphores(VkDevice device, const VkSemaphoreWaitInfo* pWaitInfo, uint64_t timeout) const override;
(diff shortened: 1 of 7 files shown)


layers/core_checks/core_validation.h
```diff
@@ -1395,7 +1397,6 @@ class CoreChecks : public ValidationStateTracker {
     bool PreCallValidateDestroyEvent(VkDevice device, VkEvent event, const VkAllocationCallbacks* pAllocator) const override;
     bool PreCallValidateDestroyQueryPool(VkDevice device, VkQueryPool queryPool,
                                          const VkAllocationCallbacks* pAllocator) const override;
-    bool ValidateGetQueryPoolResultsQueries(VkQueryPool queryPool, uint32_t firstQuery, uint32_t queryCount) const;
```
juan-lunarg, line 1398: Strange the compiler didn't complain this function has no implementation.

artem-lunarg: Yes, compiler did not complain. VS editor highlights such signatures, that's the way how I discover this.

layers/core_checks/synchronization_validation.cpp
```diff
@@ -397,10 +397,43 @@ bool CoreChecks::ValidateSemaphoresForSubmit(SemaphoreSubmitState &state, const
...
+        auto check_export_support = [&](VkExternalFenceHandleTypeFlagBits flag) {
+            auto external_info = LvlInitStruct<VkPhysicalDeviceExternalFenceInfo>();
+            external_info.handleType = flag;
+            DispatchGetPhysicalDeviceExternalFenceProperties(physical_device, &external_info, &external_properties);
+            if ((external_properties.externalFenceFeatures & VK_EXTERNAL_FENCE_FEATURE_EXPORTABLE_BIT) == 0) {
+                export_supported = false;
+                skip |= LogError(device, "VUID-VkExportFenceCreateInfo-handleTypes-01446",
+                                 "vkCreateFence(): VkFenceCreateInfo pNext chain contains VkExportFenceCreateInfo with the %s flag "
+                                 "set, which does not support VK_EXTERNAL_FENCE_FEATURE_EXPORTABLE_BIT.",
+                                 string_VkExternalFenceHandleTypeFlagBits(flag));
```
juan-lunarg, line 417: `VUID-VkExportFenceCreateInfo-handleTypes-01446`

> The bits in handleTypes must be supported and compatible, as reported by [VkExternalFenceProperties](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkExternalFenceProperties.html)

Why the check for VK_EXTERNAL_FENCE_FEATURE_EXPORTABLE_BIT?

tests/layer_validation_tests.cpp
```diff
@@ -582,6 +614,23 @@ VkExternalMemoryHandleTypeFlags GetCompatibleHandleTypes(const VkLayerTest &test
     return external_properties.externalMemoryProperties.compatibleHandleTypes;
 }
 
+VkExternalFenceHandleTypeFlags GetCompatibleHandleTypes(const VkLayerTest &test, VkExternalFenceHandleTypeFlagBits handle_type) {
```
juan-lunarg, line 617: Instead of passing in the entire test. Just pass the physical device.

juan-lunarg (commented):
Looks good overall. Just 1 question and 1 nitpick.

layers/core_checks/synchronization_validation.cpp
artem-lunarg: That's that part with 'supported' requirement. If EXPORTABLE feature bit is present then the fence can be exported.

tests/layer_validation_tests.cpp
artem-lunarg: Good point. Need to update other place.

spencer-lunarg (approved):
lgtm pending the Physical Device change

juan-lunarg (approved):

PR #5267, opened 2023-02-10
 tests: Add VkRenderFramework::GetRenderTargetFormat 

juan-lunarg:


tests/vktestframework.cpp
```diff
@@ -270,23 +258,6 @@ void VkTestFramework::InitArgs(int *argc, char *argv[]) {
     }
 }
 
-VkFormat VkTestFramework::GetFormat(VkInstance instance, vk_testing::Device *device) {
```
artem-lunarg, line 273: It was a good property of these functions that they assumed that format is supported, so the caller could assume result is always valid format. According to VulkanCapsViewer rgba8/bgra8 with optimal tiling + color_attachment is supported on 99.45%  across all platforms and the rest of that percent it's probably some anomaly.

juan-lunarg: That's fair my main ration was trying to avoid usage of std::exit since it has concurrency issues:
https://releases.llvm.org/15.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/concurrency/mt-unsafe.html

I'll make the function simpler.

juan-lunarg: I've revised the function to just assert when the format isn't supported.

tests/vkrenderframework.cpp
```diff
@@ -583,6 +583,24 @@ bool VkRenderFramework::IsPlatform(PlatformType platform) {
 
 void VkRenderFramework::GetPhysicalDeviceProperties(VkPhysicalDeviceProperties *props) { *props = physDevProps_; }
 
+VkFormat VkRenderFramework::GetRenderTargetFormat() {
+    VkFormat format = VK_FORMAT_UNDEFINED;
+    VkFormatProperties format_props = {};
+    vk::GetPhysicalDeviceFormatProperties(gpu_, VK_FORMAT_B8G8R8A8_UNORM, &format_props);
+    if (format_props.linearTilingFeatures & VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT ||
+        format_props.optimalTilingFeatures & VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT) {
+        format = VK_FORMAT_B8G8R8A8_UNORM;
```
artem-lunarg, line 592: There is one difference with original version - it returned format immediately, so format was from the first 'if' and now it's the from the second one. Also if format was found in the first part then GetphysicalDeviceFormatProperties was called only once. I'm writing this as a comment, up to the author whether this is important!

artem-lunarg (approved):

tests/vkrenderframework.cpp
juan-lunarg: Good catch. Tests are currently relying on the early return.

spencer-lunarg (approved):
good cleanup!

PR #5273, opened 2023-02-13
Pipeline Validation error message cleanup

spencer-lunarg:
The main thing this PR became was adding the index of `pCreateInfos`  of the pipeline to the `PIPELINE_STATE` and then updating and improving the error messages to help locate **where** the error occurred

### DIFF
--- a/layers/core_checks/shader_validation.cpp
+++ b/layers/core_checks/shader_validation.cpp
@@ -122,10 +122,10 @@ static uint32_t GetShaderStageId(VkShaderStageFlagBits stage) {
     return bit_pos - 1;
 }
 
-bool CoreChecks::ValidateViAgainstVsInputs(safe_VkPipelineVertexInputStateCreateInfo const *vi,
-                                           const SHADER_MODULE_STATE &module_state, const Instruction &entrypoint) const {
+bool CoreChecks::ValidateViAgainstVsInputs(const PIPELINE_STATE &pipeline, const SHADER_MODULE_STATE &module_state,
+                                           const Instruction &entrypoint) const {
     bool skip = false;
-
+    safe_VkPipelineVertexInputStateCreateInfo const *vi = pipeline.vertex_input_state->input_state;
     const auto inputs = module_state.CollectInterfaceByLocation(entrypoint, spv::StorageClassInput, false);
 
     // Build index by location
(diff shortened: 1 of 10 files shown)


juan-lunarg (approved):

layers/core_checks/shader_validation.cpp
```diff
@@ -343,9 +356,10 @@ bool CoreChecks::ValidateFsOutputsAgainstRenderPass(const SHADER_MODULE_STATE &m
                 if (!(output_type & attachment_type)) {
                     skip |= LogWarning(
                         module_state.vk_shader_module(), kVUID_Core_Shader_InterfaceTypeMismatch,
-                        "Attachment %" PRIu32
+                        "vkCreateGraphicsPipelines(): pCreateInfos[%" PRIu32 "] Attachment %" PRIu32
```
arno-lunarg, line 359: Question: is it normal to have `vkCreateGraphicsPipelines` hard coded instead of using `pipe.GetCreateFunctionName()`?

layers/state_tracker/pipeline_state.h
```diff
@@ -261,6 +261,22 @@ class PIPELINE_STATE : public BASE_NODE {
...
+                return "vkCreateGraphicsPipelines";
+            case VK_STRUCTURE_TYPE_COMPUTE_PIPELINE_CREATE_INFO:
+                return "vkCreateComputePipelines";
+            case VK_STRUCTURE_TYPE_RAY_TRACING_PIPELINE_CREATE_INFO_NV:
+                return "vkCreateRayTracingPipelinesNV";
+            case VK_STRUCTURE_TYPE_RAY_TRACING_PIPELINE_CREATE_INFO_KHR:
+                return "vkCreateRayTracingPipelinesKHR";
+            default:
+                assert(false);
+                return nullptr;
```
arno-lunarg, line 276: I would return `""` instead of `nullptr` to avoid an eventual crash

arno-lunarg (dismissed):
LGTM, 1 change requested

layers/state_tracker/pipeline_state.h
spencer-lunarg: Good catch

layers/core_checks/shader_validation.cpp
spencer-lunarg: well since it is `ValidateFsOutputsAgainstRenderPass` and deals with fragment shaders and render passes it will never be in a Ray Tracing or Compute pipeline

>  is it normal

To be fair, normally we pass in the `caller` as an argument everywhere, this is all very much "testing to see if this works well and then expand it more" type thing

PR #5274, opened 2023-02-13
layers: Combine vkGetImageSubresourceLayout logic

spencer-lunarg:
combine `vkGetImageSubresourceLayout` and `vkGetImageSubresourceLayout2EXT` into a common function. Took what each was doing well into single function


layers/core_checks/core_validation.h
```diff
@@ -2190,6 +2188,10 @@ class CoreChecks : public ValidationStateTracker {
     void PostCallRecordGetQueryPoolResults(VkDevice device, VkQueryPool queryPool, uint32_t firstQuery, uint32_t queryCount,
                                            size_t dataSize, void* pData, VkDeviceSize stride, VkQueryResultFlags flags,
                                            VkResult result) override;
+    bool ValidateGetImageSubresourceLayout(VkDevice device, const IMAGE_STATE& image_state, const VkImageSubresource subresource,
+                                           VkSubresourceLayout layout, bool is_ext) const;
```
artem-lunarg, line 2192: `VkSubresourceLayout` is 40 bytes on 64-bit. The suggestion is to pass by reference.

layers/core_checks/image_validation.cpp
```diff
@@ -2123,128 +2123,117 @@ bool CoreChecks::PreCallValidateDestroyImageView(VkDevice device, VkImageView im
...
+            if ((aspect_mask != VK_IMAGE_ASPECT_PLANE_0_BIT) && (aspect_mask != VK_IMAGE_ASPECT_PLANE_1_BIT)) {
+                const char *vuid =
+                    is_ext ? "VUID-vkGetImageSubresourceLayout2EXT-format-01581" : "VUID-vkGetImageSubresourceLayout-format-01581";
+                skip |= LogError(image_state.image(), vuid,
+                                 "%s: plane count of image format (%s) is 2 but VkImageSubresource.aspectMask is %s", caller,
+                                 string_VkFormat(image_format), string_VkImageAspectFlags(aspect_mask).c_str());
+            }
+        }
+
+        if (FormatPlaneCount(image_format) == 3) {
```
artem-lunarg, line 2207: That's more clear and specific comparing to `> 2u` before, also other similar changes. Nice readability improvements for functions with a lot of conditions.

spencer-lunarg: The `vkGetImageSubresourceLayout` seemed to have been written first and "duct taped" together as new VUs were added and `vkGetImageSubresourceLayout2`

layers/core_checks/image_validation.cpp
```diff
@@ -2123,128 +2123,117 @@ bool CoreChecks::PreCallValidateDestroyImageView(VkDevice device, VkImageView im
...
+            is_ext ? "VUID-vkGetImageSubresourceLayout2EXT-format-04462" : "VUID-vkGetImageSubresourceLayout-format-04462";
+        skip |= LogError(image_state.image(), vuid,
+                         "%s: format of image is %s which has depth component "
+                         "but VkImageSubresource.aspectMask is %s",
+                         caller, string_VkFormat(image_format), string_VkImageAspectFlags(aspect_mask).c_str());
     }
 
-    // subresource's aspect must be compatible with image's format.
-    const VkFormat img_format = image_entry->createInfo.format;
-    if (image_entry->createInfo.tiling == VK_IMAGE_TILING_LINEAR) {
```
artem-lunarg, line 2175: I have question about `== VK_IMAGE_TILING_LINEAR` from the line above from the old version. Is it possible to have tiling `VK_IMAGE_TILING_DRM_FORMAT_MODIFIER_EXT` here? In the old version all validations were skipped here, for example, `VUID-vkGetImageSubresourceLayout-format-04461`, but it looks in the new version 04461 will always be checked, so the question if it's expected behavior?

spencer-lunarg: so the current ` VUID-vkGetImageSubresourceLayout-format-04461` says ([and always said](https://github.com/KhronosGroup/Vulkan-Docs/commit/5c8259673b722a86525f5a9dbee1ffd92371656b#diff-3d38098a20cbc7e6fea6b6ef5f07338ec4afeae0d355ac0c805029e2d12e76ddR2278))

>  If format is a color format, the aspectMask member of pSubresource must be VK_IMAGE_ASPECT_COLOR_BIT

The `tiling` should only matter when it comes to Multi-planar as that is when the DRM Format Modifier is slightly different... so I think this was a bug before

artem-lunarg (approved):
It's a bit hard to check diff, but LGTM!

layers/core_checks/image_validation.cpp
```diff
@@ -2262,27 +2251,31 @@ bool CoreChecks::PreCallValidateGetImageSubresourceLayout(VkDevice device, VkIma
...
-                                 string_VkFormat(image_entry->createInfo.format), drm_format_properties.drmFormatModifier);
+                                 caller, string_VkImageAspectFlags(aspect_mask).c_str(), max_plane_count,
+                                 string_VkFormat(image_state.createInfo.format), drm_format_properties.drmFormatModifier);
             }
         }
     }
 
-    if (image_entry->IsExternalAHB() && (0 == image_entry->GetBoundMemoryStates().size())) {
-        skip |= LogError(image, "VUID-vkGetImageSubresourceLayout-image-01895",
+    if (image_state.IsExternalAHB() && (0 == image_state.GetBoundMemoryStates().size())) {
```
juan-lunarg, line 2275: 1 minor nitpick

```suggestion
    if (image_state.IsExternalAHB() && image_state.GetBoundMemoryStates().empty()) {
```

juan-lunarg (approved):

PR #5276, opened 2023-02-13
Add new ray tracing tests and helper classes

arno-lunarg:
Add tests for 03432, 06731, 06733 and 06742

Also add 3 new ray tracing helper classes, to ease the burden of creating ray tracing objects, and make ray tracing tests more simple and readable. Using those classes, I removed about 250 lines of codes to existing ray tracing tests.
They are defined in `ray_tracing_objects.h`, and targeted at helping creating `VkAccelerationStructureGeometryKHR`, `VkAccelerationStructureCreateInfoKHR` and `VkAccelerationStructureBuildGeometryInfoKHR`. 

Typical usage probably is:
```cpp
    // Get a valid object template. Calling right after BuildCmdBuffer() will result in a valid object.
    rt::as::BuildGeometryInfoKHR as_build_info = BuildGeometryInfoSimpleOnDeviceBottomLevel(*m_device);

    // Use the various setters to invalidate the template as needed
    as_build_info.GetDstAS().SetBufferMemoryPropertyFlags(VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT);
    as_build_info.SetFlags(VK_BUILD_ACCELERATION_STRUCTURE_ALLOW_COMPACTION_BIT_KHR);

    // Build the object, triggering VUIDs
    m_commandBuffer->begin();
    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-FooBar");
    as_build_info.BuildCmdBuffer(*m_device, m_commandBuffer->handle());
    m_commandBuffer->end();
```

As of now, creating all the structs needed to be able to call `vkCmdBuildAccelerationStructuresKHR` is painful, and is done by copy pasting some big chunk of code from some test, finding where the structs where made invalid in the original test, make them valid again, and finally invalidate them as you want for your new test. With those classes, creating a valid object is a one liner, right after you invalidate it, then you try to build it. I think it makes things clearer.

I already updated some tests, mostly to make sure those new classes are relevant and working as intended. If I am green lighted, I will modify other tests where it is possible.

I added comments in `ray_tracing_objects.h` to explain usage and code flow, just tell me if they do not make sense or if I am missing some!


tests/CMakeLists.txt
```diff
@@ -77,6 +77,8 @@ target_sources(vk_layer_validation_tests PRIVATE
     vksyncvaltests.cpp
     vktestbinding.cpp
     vktestframework.cpp
+    ray_tracing_objects.h
```
spencer-lunarg, line 80: the name currently is inconsistent with other test file names... but honestly this is something I wanted to clean up anyway, happy for now to keep it as `ray_tracing_objects` for now

tests/ray_tracing_objects.h
```diff
@@ -0,0 +1,195 @@
...
+//
+//    // for instance:
+//    as_build_info.GetDstAS().SetBufferMemoryPropertyFlags(VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT);
+//    as_build_info.SetFlags(VK_BUILD_ACCELERATION_STRUCTURE_ALLOW_COMPACTION_BIT_KHR);
+//
+//    m_commandBuffer->begin();
+//    as_build_info.BuildCmdBuffer(*m_device, m_commandBuffer->handle());
+//    m_commandBuffer->end();
+// }
+namespace blueprint {
```
spencer-lunarg, line 183: why `blueprint`, this a reference to something?

tests/ray_tracing_objects.h
```diff
@@ -0,0 +1,195 @@
...
+
+    GeometryKHR& Build();
+
+    const auto& GetVkObj() const { return vk_obj_; }
+    VkAccelerationStructureBuildRangeInfoKHR GetFullBuildRange() const;
+    const auto& GetTriangles() const { return triangles_; }
+    const auto& GetAABBs() const { return aabbs_; }
+
+  private:
+    uint32_t vk_api_version_;
```
spencer-lunarg, line 72: not for the scope of this PR, but I realize in the test framework, we have to constantly pass around lots of things like API Version which are only set once at `TEST` level, not sure if this is something that can be static for everyone to use

tests/ray_tracing_objects.h
```diff
@@ -0,0 +1,195 @@
...
+  private:
+    void BuildCommon(const vk_testing::Device& device, bool use_ppGeometries = true);
+
+    uint32_t vk_api_version_;
+    uint32_t vk_info_count_ = 1;
+    bool use_null_infos_ = false;
+    bool use_null_build_range_infos_ = false;
+    VkAccelerationStructureBuildGeometryInfoKHR vk_info_;
+    std::vector<GeometryKHR> geometries_;
+    std::shared_ptr<AccelerationStructureKHR> srcAS_, dstAS_;
```
spencer-lunarg, line 165: super nit: `SetSrcAS` has these as `src_as` (for consistency make the same)

tests/ray_tracing_objects.h
```diff
@@ -0,0 +1,195 @@
...
+ */
+
+#pragma once
+
+#include "vktestbinding.h"
+
+#include <memory>
+
+// ray tracing
+namespace rt {
```
spencer-lunarg, line 22: I guess curious if for now it makes sense to have both a `rt` and `as` namespace

Could argue for now just a `as` namespace is really needed. Not opposed to either way

tests/ray_tracing_objects.cpp
```diff
@@ -0,0 +1,593 @@
...
+    return *this;
+}
+
+GeometryKHR &GeometryKHR::SetPrimitiveCount(uint32_t primitiveCount) {
+    primitiveCount_ = primitiveCount;
+    return *this;
+}
+
+GeometryKHR &GeometryKHR::SetStride(VkDeviceSize stride) {
+    static_assert(int(Type::_INTERNAL_COUNT) == 3, "Unhandled Geometry type in GeometryKHR::SetStride");
```
spencer-lunarg, line 53: I don't see a real use for `_INTERNAL_COUNT` in this code

I would just remove it, the `static_assert` and replace the `default` with `_INTERNAL_UNSPECIFIED` so if a new enum is added, we will hit an error for not covering all the switch cases then

tests/ray_tracing_objects.cpp
```diff
@@ -0,0 +1,593 @@
...
+    }
+    vk_info_.buffer = buffer_.handle();
+
+    // Create acceleration structure
+    auto vkCreateAccelerationStructureKHR = reinterpret_cast<PFN_vkCreateAccelerationStructureKHR>(
+        vk::GetDeviceProcAddr(device.handle(), "vkCreateAccelerationStructureKHR"));
+    assert(vkCreateAccelerationStructureKHR);
+    VkAccelerationStructureKHR handle;
+    VkResult result = vkCreateAccelerationStructureKHR(device.handle(), &vk_info_, nullptr, &handle);
+    assert(result == VK_SUCCESS || result == VK_ERROR_VALIDATION_FAILED_EXT || result == VK_ERROR_OUT_OF_HOST_MEMORY);
```
spencer-lunarg, line 180: `assert(result == VK_SUCCESS)`?

tests/ray_tracing_objects.cpp
```diff
@@ -0,0 +1,593 @@
...
+        if (buffer_usage_flags_ != 0) {
+            auto alloc_flags = LvlInitStruct<VkMemoryAllocateFlagsInfo>();
+            alloc_flags.flags = buffer_memory_allocate_flags_;
+            buffer_.init(device, vk_info_.size, buffer_memory_property_flags_, buffer_usage_flags_, &alloc_flags);
+        }
+    }
+    vk_info_.buffer = buffer_.handle();
+
+    // Create acceleration structure
+    auto vkCreateAccelerationStructureKHR = reinterpret_cast<PFN_vkCreateAccelerationStructureKHR>(
```
spencer-lunarg, line 175:
```patch
-  auto vkCreateAccelerationStructureKHR = reinterpret_cast<PFN_vkCreateAccelerationStructureKHR>(vk::GetDeviceProcAddr(device.handle(), "vkCreateAccelerationStructureKHR"));
-  assert(vkCreateAccelerationStructureKHR);
+ auto vkCreateAccelerationStructureKHR = GetDeviceProcAddr<PFN_vkCreateAccelerationStructureKHR>("vkCreateAccelerationStructureKHR")
```

tests/ray_tracing_objects.h
```diff
@@ -0,0 +1,195 @@
...
+
+class BuildGeometryInfoKHR {
+  public:
+    ~BuildGeometryInfoKHR() = default;
+    BuildGeometryInfoKHR(uint32_t vk_api_version);
+    BuildGeometryInfoKHR(BuildGeometryInfoKHR&&) = default;
+    BuildGeometryInfoKHR& operator=(BuildGeometryInfoKHR&& rhs) = default;
+    BuildGeometryInfoKHR& operator=(const BuildGeometryInfoKHR&) = delete;
+
+    BuildGeometryInfoKHR& SetType(VkAccelerationStructureTypeKHR type);
```
spencer-lunarg, line 125: Why are these `Set*()` all returning the object, I don't see any code using it? should they just return `void`?

tests/ray_tracing_objects.cpp
arno-lunarg: `GetDeviceProcAddr` is a method of `VkLayerTest`, so I think I need to keep this crap 😓

tests/ray_tracing_objects.h
arno-lunarg: I was thinking that it would be cool to be able to chain `Set` calls, like so:
```cpp
foo.SetAAA(1).SetBBB("bar").Set(VK_FOO).Build();
```
But I ended up not doing it, probably scared of a "sorry we do not do that here" comment. But maybe it would be ok to do it? If not, I am going to change them all to return `void`

tests/ray_tracing_objects.h
arno-lunarg: So to me `blueprint` captures the intent of the functions in this namespace: provide the data necessary to successfully build the corresponding object, but not directly build objects, so that the user can modify those blueprints before actually building them.
I wanted to have a dedicated namespace to help find them using completion. At least on VS, typing `rt::as::blueprint::` will list all of the available blueprint functions, and _only them_, no clutter from the object classes.

tests/ray_tracing_objects.h
arno-lunarg: I really like the idea!

tests/ray_tracing_objects.h
arno-lunarg: I was thinking about the future, eventually I am going to add helper functions that manage ray tracing aspects, but are not necessarily related to acceleration structures.

tests/ray_tracing_objects.cpp
arno-lunarg: If I understand correctly, then we would only be able to detect defect at run time if a new type is added? I added this so that we can detect problems at compile time, and at every spots where I `switch`, I think it is safer? But yes more code, and I am not sure that this is something we do elsewhere.

tests/ray_tracing_objects.h
spencer-lunarg: I guess for testing I would not be opposed to this type of chaining

tests/CMakeLists.txt
arno-lunarg: A cleanup would be welcome ^^"
Maybe not a topic to discuss here, but I would not mind clarifications behind naming conventions in the tests folder, eg why do we have both names like `vktestfoo` and `vklayertests_bar`

tests/ray_tracing_objects.cpp
arno-lunarg: ah I had not noticed that, I will do as you advise!

spencer-lunarg (approved):

tests/ray_tracing_objects.cpp
```diff
@@ -0,0 +1,583 @@
...
+    vk_testing::Buffer index_buffer(device, 1024, VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT,
+                                    buffer_usage, &alloc_flags);
+
+    // Fill vertex and index buffers with one triangle
+    triangle_geometry.SetPrimitiveCount(1);
+    constexpr std::array vertices = {1.0f, 0.0f, 0.0f, 0.0f, 1.0f, 0.0f, -1.0f, 0.0f, 0.0f};
+    constexpr std::array<uint32_t, 3> indicies = {{0, 1, 2}};
+
+    auto mapped_vbo_buffer_data = vertex_buffer.memory().map();
+    std::memcpy(mapped_vbo_buffer_data, vertices.data(), sizeof(vertices[0]) * vertices.size());
```
ncesario-lunarg, line 484: If it compiles, it might be a tad bit cleaner to cast to the appropriate pointer type and use `std::copy`, which avoids the need for element size calculations.

tests/ray_tracing_objects.h
```diff
@@ -0,0 +1,196 @@
...
+  private:
+    void BuildCommon(const vk_testing::Device& device, bool use_ppGeometries = true);
+
+    uint32_t vk_api_version_;
+    uint32_t vk_info_count_ = 1;
+    bool use_null_infos_ = false;
+    bool use_null_build_range_infos_ = false;
+    VkAccelerationStructureBuildGeometryInfoKHR vk_info_;
+    std::vector<GeometryKHR> geometries_;
+    std::shared_ptr<AccelerationStructureKHR> src_as_, dst_as_;
```
ncesario-lunarg, line 166: Why are these `shared_ptr` and not `unique_ptr`? Are there really multiple owners of this data? If so, why?

ncesario-lunarg (approved):
Couple questions, but LGTM! Nice to see the testing framework get some TLC :)

tests/ray_tracing_objects.h
arno-lunarg: I wanted the user to be able to use the same `AccelerationStructureKHR` as both a source and a destination, so I think I have to use `shared_ptr`?

tests/ray_tracing_objects.cpp
arno-lunarg: much cleaner!

tests/ray_tracing_objects.h
```diff
@@ -0,0 +1,196 @@
...
+    GeometryKHR& SetTrianglesVertexBuffer(vk_testing::Buffer&& vertex_buffer, uint32_t max_vertex,
+                                          VkFormat vertex_format = VK_FORMAT_R32G32B32_SFLOAT,
+                                          VkDeviceSize stride = 3 * sizeof(float));
+    GeometryKHR& SetTrianglesIndexBuffer(vk_testing::Buffer&& index_buffer, VkIndexType index_type = VK_INDEX_TYPE_UINT32);
+    GeometryKHR& SetTrianglesIndexType(VkIndexType index_type);
+    GeometryKHR& SetAABBsBuffer(vk_testing::Buffer&& buffer, VkDeviceSize stride = sizeof(VkAabbPositionsKHR));
+
+    GeometryKHR& Build();
+
+    const auto& GetVkObj() const { return vk_obj_; }
```
artem-lunarg, line 66: I find this `auto` usage is a bit harder to understand comparing to other auto use cases. Probably the reason that I see the interface as the documentation of the system and explicit types here are more useful comparing to implementation code.  It looks it's the first such usage in the codebase (or the regex I used is wrong!). Not sure, it could be just a habit, I'm fine with this if it works for other developers.

tests/ray_tracing_objects.cpp
artem-lunarg: I checked that at least vs implementation generates `memmove` for this `std::copy`, so it's also fast.

tests/ray_tracing_objects.h
artem-lunarg: Is it required to make allocation of `AccelerationStructureKHR` at all? I did not check most of the code, it might be the answer will be clear if I read more.

PR #5282, opened 2023-02-14
cmake: Disable warnings on a per target basis

juan-lunarg:
Makes fixing warnings more feasible. Since you can tackle the problem one target at a time, instead of having to fix the entire project


spencer-lunarg (approved):

artem-lunarg (approved):
LGTM ✔

ncesario-lunarg (approved):

arno-lunarg (approved):

PR #5284, opened 2023-02-14
layers: Add missing promoted extensions

ncesario-lunarg:
Use 'promotedto' attribute for identifying promoted extensions in vk.xml.

FYI @Tony-LunarG.


Tony-LunarG (approved):
LGTM and much better than the manually maintained list - Thanks!

juan-lunarg (approved):

PR #5285, opened 2023-02-14
android: Fix layerlib/CMakeLists.txt

### LINKED ISSUE #5271: Delete build-android/cmake/ ?
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/6e62e504d21c633e1a12cb698bfe83d839125f4d/build-android/cmake/layerlib/CMakeLists.txt#L15

Is this CMake code even used/tested? Currently it uses C++11 which our code won't compile with.

It looks like the Android.mk script is what gets tested/used.

+ @spencer-lunarg & @ncesario-lunarg for input

juan-lunarg:
closes #5271


spencer-lunarg (approved):

ncesario-lunarg:
Is it confirmed that `build-android/cmake` is being actively used? The fact that nobody has complained about it being broken would seem to indicate it should be removed.

PR #5286, opened 2023-02-15
Fix bug with vkCmdEndQuery and move tests

spencer-lunarg:
Using Profiles/MockICD I found we have not been testing with a `VK_KHR_performance_query` device as there was a bug where we were off by one

To make future query-related testing easier, moved query tests to own file


layers/state_tracker/cmd_buffer_state.h
```diff
@@ -167,7 +167,7 @@ class CMD_BUFFER_STATE : public REFCOUNTED_NODE {
     bool has_build_as_cmd;
 
     CB_STATE state;               // Track cmd buffer update state
-    uint64_t commandCount;        // Number of commands recorded. Currently only used with VK_KHR_performance_query
+    uint64_t command_count;       // Number of commands recorded. Currently only used with VK_KHR_performance_query
```
ncesario-lunarg, line 170: Might be good to do variable renaming in a separate commit...or just wait until we do that impending/"some day" clang-tidy refactor.

tests/vklayertests_query.cpp
```diff
@@ -0,0 +1,2385 @@
...
+
+#include "cast_utils.h"
+#include "core_validation_error_enums.h"
+#include "enum_flag_bits.h"
+#include "layer_validation_tests.h"
+#include "vk_layer_utils.h"
+#include "generated/vk_validation_error_messages.h"
+#include "vk_layer_logging.h"
+
+TEST_F(VkLayerTest, QueryPerformanceCreation) {
```
ncesario-lunarg, line 23: Would it make sense to namespace these accordingly (e.g., `VkLayerQueryTest, ...`)? Maybe not for this change, but asking more in general if having a namespace per-file would make sense.

ncesario-lunarg (approved):

tests/vklayertests_query.cpp
spencer-lunarg: Had Exact thought, I don't see why not to start now

layers/state_tracker/cmd_buffer_state.h
spencer-lunarg: agree, can make a separate commit here

PR #5291, opened 2023-02-15
tests: Fix tests for use with Profile layer

spencer-lunarg:
Clean up some tests in order for Profile layer and MockICD to run with them

`MultisampledRenderToSingleSampled` and `RenderPassCreateFragmentDensityMapReferenceToInvalidAttachment` were not valid tests (not being tested in CI it seems)

### DIFF
--- a/tests/vklayertests_imageless_framebuffer.cpp
+++ b/tests/vklayertests_imageless_framebuffer.cpp
@@ -1494,8 +1494,8 @@ TEST_F(VkLayerTest, RenderPassCreateFragmentDensityMapReferenceToInvalidAttachme
         GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported";
     }
 
-    if (DeviceValidationVersion() < VK_API_VERSION_1_1) {
-        GTEST_SKIP() << "At least Vulkan version 1.1 is required";
+    if (DeviceValidationVersion() >= VK_API_VERSION_1_1) {
+        GTEST_SKIP() << "Test requires Vulkan version 1.0";
     }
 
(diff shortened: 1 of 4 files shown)


tests/vklayertests_imageless_framebuffer.cpp
```diff
@@ -1564,4 +1565,6 @@ TEST_F(VkLayerTest, RenderPassCreateFragmentDensityMapReferenceToInvalidAttachme
     m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkFramebufferCreateInfo-pAttachments-02744");
     vk::CreateFramebuffer(device(), &fb_info, nullptr, &framebuffer);
     m_errorMonitor->VerifyFound();
+
+    vk::DestroyRenderPass(device(), renderPass, nullptr);
```
ncesario-lunarg, line 1569: Might as well convert this to a `vk_testing::RenderPass`. The bigger question, though, is why is this not failing? This "dangling object" should have triggered an error.

ncesario-lunarg (approved):
LGTM, but we need to find out why un-destroyed objects are not triggering an error (or the test framework is ignoring those errors). That has been the cause of sporadic test failures in the past.

juan-lunarg (approved):

spencer-lunarg:
> LGTM, but we need to find out why un-destroyed objects are not triggering an error 

I don't think this test is ran anywhere, we seem to lack a `VK_EXT_fragment_density_map` CI device (would probably be only a select set of Android device) ... but why getting the Profile Layer setup fully is my next side goal

PR #5294, opened 2023-02-15
cmake: Fix unused parameters warnings for VkLayer_utils

juan-lunarg:


layers/generated/vk_safe_struct.cpp
```diff
@@ -50664,6 +50664,7 @@ void safe_VkPipelineFragmentShadingRateEnumStateCreateInfoNV::initialize(const s
 
 safe_VkDeviceOrHostAddressConstKHR::safe_VkDeviceOrHostAddressConstKHR(const VkDeviceOrHostAddressConstKHR* in_struct)
 {
+    initialize(in_struct);
```
juan-lunarg, line 50667: Turns out https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4828 caused a regression which I spotted thanks to enabling with warning.

See `scripts/helper_file_generator.py` for the fix.

spencer-lunarg (approved):

PR #5298, opened 2023-02-16
tests: Fix minor video test framework issues

aqnuep:
This PR fixes some minor issues with the test framework for Vulkan Video tests found by valgrind and further review.

### DIFF
--- a/tests/vklayertests_video.cpp
+++ b/tests/vklayertests_video.cpp
@@ -286,9 +286,9 @@ TEST_F(VkVideoLayerTest, CreateSessionProtectedContentUnsupported) {
         GTEST_SKIP() << "Test requires protectedMemory support";
     }
 
-    VideoConfig config = GetConfigWithProtectedContent(GetConfigs());
+    VideoConfig config = GetConfigWithoutProtectedContent(GetConfigs());
     if (!config) {
-        GTEST_SKIP() << "Test requires a video profile with protected content support";
+        GTEST_SKIP() << "Test requires a video profile with no protected content support";
     }
 
     VideoContext context(DeviceObj(), config);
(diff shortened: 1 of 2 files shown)


juan-lunarg (commented):
While I have you here a have a question about the video code.

```
class VkVideoLayerTest : public VkLayerTest {
  protected:
    struct {
        PFN_vkGetPhysicalDeviceVideoCapabilitiesKHR GetPhysicalDeviceVideoCapabilitiesKHR;
        PFN_vkGetPhysicalDeviceVideoFormatPropertiesKHR GetPhysicalDeviceVideoFormatPropertiesKHR;
    } vk{};

    void Init(bool enable_protected_memory = false) {
```

enable_protected_memory is never used in the init function and it doesn't seem intentional.

juan-lunarg:
> This PR fixes some minor issues with the test framework for Vulkan Video tests found by valgrind and further review.

Have you tried using address sanitizer as well? I know it's a bit more robust than valgrind in certain areas.

`-D VVL_ENABLE_ASAN`

aqnuep:
> enable_protected_memory is never used in the init function and it doesn't seem intentional.

Indeed, you're right. That's just a left-over of an (in the end) unnecessary parameter.

> Have you tried using address sanitizer as well? I know it's a bit more robust than valgrind in certain areas.

No, but I'll give it a try.

ncesario-lunarg (approved):

juan-lunarg (changes requested):
Will approve once the `enable_protected_memory` issue is fixed.

aqnuep:
> Will approve once the enable_protected_memory issue is fixed.

In the end I went back to using the flag to explicitly not enable protected memory, as it was originally designed to be handled.

It's actually good that you noted this one as I found a bug in one of the protected memory related test cases, which I also fixed.

Unfortunately, currently no Vulkan Video implementation supports protected content, so these test cases don't have the necessary coverage due to the tests being skipped.

juan-lunarg (approved):

juan-lunarg (approved):

PR #5301, opened 2023-02-16
ci: Fix Vulkan-Profiles build failure

juan-lunarg:


juan-lunarg:
FYI to @christophe-lunarg

PR #5303, opened 2023-02-16
build: Update to header v1.3.241

ncesario-lunarg:
This also includes script updates for to accomodate:
- https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5542
- https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5700


juan-lunarg:
rebase on main to address linux CI issues

spencer-lunarg (approved):

juan-lunarg (approved):

.github/workflows/build_linux.yml
```diff
@@ -91,8 +92,12 @@ jobs:
             external/${{ matrix.config }}/Vulkan-Headers/registry
             external/${{ matrix.config }}/SPIRV-Headers/include
           key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ matrix.cc }}-${{ matrix.cxx }}-${{ matrix.config }}-${{ hashfiles('scripts/known_good.json') }}
+      - name: Set PYTHONPATH
```
mikes-lunarg, line 95: I'm surprised you have to mess with `PYTHONPATH`, previous versions of lvl_genvk.py adjusted the search path automatically based on the `-registry` argument

ncesario-lunarg: Right, but the usage of the registry python modules is becoming more prevalent and passing around the command line arguments to allow `import` statements doesn't seem ideal. If this is considered "bad python," I can change it; just seemed cleaner to me.

mikes-lunarg: The old way is definitely more magic than correct. It was convenient though, because you had to pass the registry to do anything anyway. Long-term we should probably use something like pipenv to lock down and manage python dependencies.

mikes-lunarg (approved):

.github/workflows/build_linux.yml
ncesario-lunarg: > ...because you had to pass the registry to do anything anyway...

Yeah, it might be nice if there were accessors in the registry module; e.g.:
```python
import reg
...
reg.registry_directory # the directory we currently rely on, or...
reg.vulkan_xml_path # path to vk.xml
reg.validusage_json_path # path to validusage.json
...
```
Then there would be no need to pass in the registry directory as an argument.

PR #5305, opened 2023-02-17
Add a Max Profile to CI

spencer-lunarg:
Adds a custom `max_profile.json` that should start allowing much larger test coverage in GitHub actions via the profiles layers + MockICD

Added  documentation with more info about this file. Tried to make it as small, a "full" generated profile json is about 4mb


.github/workflows/build_linux.yml
```diff
@@ -162,6 +166,10 @@ jobs:
         run: python3 scripts/github_ci_win_linux.py --test --config release --cmake='-DVVL_CPP_STANDARD=20'
         env:
           VK_KHRONOS_PROFILES_PROFILE_FILE: tests/device_profiles/nvidia_tegra_x1.json
+      - name: Test Vulkan-ValidationLayers - Max Profile
+        run: python3 scripts/github_ci_win_linux.py --test --config release --cmake='-DVVL_CPP_STANDARD=20'
```
arno-lunarg, line 170: `VVL_CPP_STANDARD=20`? Not `=17`?

tests/README.md
```diff
@@ -153,3 +153,11 @@ export VK_KHRONOS_PROFILES_DEBUG_REPORTS=DEBUG_REPORT_ERROR_BIT
 # Running tests just as normal
 $VVL/build/tests/vk_layer_validation_tests --gtest_filter=TestName
 ```
+
+### Max Profile
+
+The file in `tests/device_profiles/max_profile.json` is a custom file designed to allow as much testing as possible
```
arno-lunarg, line 159: I suggest :
```suggestion
`tests/device_profiles/max_profile.json` is a custom file designed to allow as much testing as possible.
```

tests/README.md
```diff
@@ -153,3 +153,11 @@ export VK_KHRONOS_PROFILES_DEBUG_REPORTS=DEBUG_REPORT_ERROR_BIT
...
+
+It would be ideal to generate this similar to the layer generated in [the Profile Layer repo](https://github.com/KhronosGroup/Vulkan-Profiles/blob/main/scripts/gen_profiles_layer.py), but the goal is to make sure as many tests are running in CI as possible.
```
arno-lunarg, line 161: I do not remember my grammar lessons too well, but should not it be `similarly` instead of `similar`?

tests/README.md
```diff
@@ -153,3 +153,11 @@ export VK_KHRONOS_PROFILES_DEBUG_REPORTS=DEBUG_REPORT_ERROR_BIT
...
+
+This file will **never** fully cover all tests because some tests will tests for certain properties/extension to be **both** enabled and disabled. The goal is to get as close to 100% coverage as possible
```
arno-lunarg, line 163: I suggest:
```suggestion
This file will **never** fully cover all tests because some require certain properties/extension to be **both** enabled and disabled. The goal is to get as close to 100% coverage as possible.
```

arno-lunarg (approved):
LGTM! 
One question about `VVL_CPP_STANDARD` value.
And as for my English suggestions, I might be totally wrong, so do as you like ^^

.github/workflows/build_linux.yml
ncesario-lunarg: Probably best to remove that value entirely to use the default. That value is used in a test to see if we're forward compatible with the standard (which should probably change to 23 soon).

Since the other tests are also using this value, I'd vote for leaving it alone here and updating everything in a smaller, subsequent PR.

ncesario-lunarg (approved):
💯

.github/workflows/build_linux.yml
ncesario-lunarg: Ah, I missed that this is under the linux-_cpp20_ yml entry. This shouldn't matter, but all of these profiles tests should probably move to the _linux_ entry above. _linux-cpp20_ was intended to only test forward standard compatibility.

juan-lunarg (approved):

.github/workflows/build_linux.yml
spencer-lunarg: > Since the other tests are also using this value, I'd vote for leaving it alone here and updating everything in a smaller, subsequent PR.

My same thoughts as well, should be addressed

tests/README.md
spencer-lunarg: I never did well in English class, you are right, will improve my grammar :disappointed:

PR #5309, opened 2023-02-19
tests: Create more test groups

spencer-lunarg:
Makes use of the `TEST_F` grouping for more reable test names

### DIFF
--- a/tests/positive/android_hardware_buffer.cpp
+++ b/tests/positive/android_hardware_buffer.cpp
@@ -16,7 +16,8 @@
 
 #include "android_ndk_types.h"
 #ifdef AHB_VALIDATION_SUPPORT
-TEST_F(VkPositiveLayerTest, AndroidHardwareBufferMemoryRequirements) {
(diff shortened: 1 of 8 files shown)


arno-lunarg:
Just my 0.02$: 
What about having names composed akin to namespaces, and have for instance:
`VkLayerTestRayTracing` instead of `VkLayerRayTracingTest`
`VkLayerTestPositiveDynamicRendering` instead of `VkPositiveDynamicRenderingLayerTest` (As if we had `VkLayerTest::Positive::DynamicRendering`)
etc... so that they all start with `VkLayerTest` or `VkLayerTestPositive`?

mark-lunarg:
While we're in this space, onsideration might also be given to dumping `Vk`, `Layer`, and/or `Test`.  At least 'Vulkan'and 'Test' are redundant for sure.

spencer-lunarg:
closing, going to re-do this but with all the tests in the future to minimize churn with CI blacklist

Issue #5302, opened 2023-02-16
VUID-VkGraphicsPipelineCreateInfo-renderPass-06038 does not correctly check if attachment is read


aitor-lunarg:
**Environment:**
 - OS: Any
 - GPU: Any
 - SDK or header version if building from repo: ToT
 - Options enabled (synchronization, best practices, etc.):

**Describe the Issue**

Issue can be replicated with CTS tests:
 - dEQP-VK.descriptor_indexing.input_attachment
 - dEQP-VK.descriptor_indexing.input_attachment_in_loop

`VUID-VkGraphicsPipelineCreateInfo-renderPass-06038` does not correctly check if the attachment is actually read in the fragment shader. The current check does not take into account what inputs the shader requires, it just checks if there's an input attachment present, if the index is greater or equal than input attachment count for subpass, and that the attachment is not `VK_ATTACHMENT_UNUSED`.

The ambiguity could be due to the `read` part in the VUID definition, but I do understand that the `read` refers to actually accessing that input attachment at run time.

There are 2 scenarios to take into account for this VUID:
 - Statically accessed attachments in the shader. These can be checked at pipeline creation time.
 - Dynamically accessed attachments in the shader. These require to be checked at draw time.

**Expected behavior**

The VUID should not be reported even if there's an input attachment with `VK_ATTACHMENT_UNUSED` as long as it's not actually read by the shader.

spencer-lunarg:
(just adding information) I added this in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5026 

I was going to add ` VUID-VkGraphicsPipelineCreateInfo-renderPass-06056` next, but will fix the fact `read` is not statically reading what the shader is doing

aitor-lunarg:
I see, thank you for the information.

I was not aware of `VUID-VkGraphicsPipelineCreateInfo-renderPass-06056`. Maybe it would be good to have that VUID state it's dynamic rendering related. Mainly to avoid confussion between `06038` and `06056` since the former would just be a subset of the latter (there's no mention to dynamic rendering in `06056`).

So I think there should at least be some mention to dynamic rendering for `06056` or have it stated that if renderPass is NULL then what the VUID states at the moment.

Let me know your thoughts, and I can open an issue in the spec to tackle this.

spencer-lunarg:
@aitor-lunarg looking at `dEQP-VK.descriptor_indexing.input_attachment`

[...]

PR #5310, opened 2023-02-20
Cleanup OpVariable access status code

spencer-lunarg:
In the attempt to fix https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5302 I ran into some spec clarification issues that need to be resolved first. This PR pulls out the "clean up" part of the change I was planning to make. 

The goal is nothing functional should have change in this PR and make the next PR have less noise to review

### DIFF
--- a/layers/state_tracker/pipeline_state.h
+++ b/layers/state_tracker/pipeline_state.h
@@ -77,14 +77,14 @@ extern DescriptorReqFlags DescriptorRequirementsBitsFromFormat(VkFormat fmt);
 
 struct DescriptorRequirement {
     DescriptorReqFlags reqs;
-    bool is_writable;
+    bool is_written_to;
     // Copy from StageState.ResourceInterfaceVariable. It combines from plural shader stages. The index of array is index of image.
     std::vector<vvl::unordered_set<SamplerUsedByImage>> samplers_used_by_image;
     // For storage images - list of < OpImageWrite : Texel component length >
     std::vector<std::pair<Instruction, uint32_t>> write_without_formats_component_count_list;
     // OpTypeImage the variable points to, null if doesn't use the image
     uint32_t image_sampled_type_width = 0;
-    DescriptorRequirement() : reqs(0), is_writable(false) {}
+    DescriptorRequirement() : reqs(0), is_written_to(false) {}
 };
 
 inline bool operator==(const DescriptorRequirement &a, const DescriptorRequirement &b) noexcept { return a.reqs == b.reqs; }
@@ -100,7 +100,7 @@ struct PipelineStageState {
     VkShaderStageFlagBits stage_flag;
     std::optional<Instruction> entrypoint;
     const std::vector<ResourceInterfaceVariable> *descriptor_variables = {};
-    bool has_writable_descriptor;
+    bool has_descriptor_written_to;
     bool has_atomic_descriptor;
     bool wrote_primitive_shading_rate;
     bool writes_to_gl_layer;
--- a/layers/state_tracker/shader_module.cpp
+++ b/layers/state_tracker/shader_module.cpp
@@ -41,9 +41,6 @@ void DecorationSet::Add(uint32_t decoration, uint32_t value) {
         case spv::DecorationComponent:
             component = value;
             break;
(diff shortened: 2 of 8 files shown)


layers/state_tracker/pipeline_state.h
```diff
@@ -77,14 +77,14 @@ extern DescriptorReqFlags DescriptorRequirementsBitsFromFormat(VkFormat fmt);
 
 struct DescriptorRequirement {
     DescriptorReqFlags reqs;
-    bool is_writable;
+    bool is_written_to;
```
artem-lunarg, line 80: Question about the semantics of the news names (asking about all names in this change, not this specific variable). In the new version it says descriptor write has already happened. Is it correct? Is it guaranteed in all contexts? (quite different meaning from the old name which allows the writes).

spencer-lunarg: so I guess when I saw "writable" I thought "Ok, so it must be a UNIFORM and not a STORAGE descriptor type"

But looking at the code, it was being used as "we are doing a write operation (`OpImageWrite`, `OpAtomicStore`, etc)
 
So it is very misleading of a name

this variable is being used to check VUs like

> VUID-vkCmdDraw-commandBuffer-02712
If commandBuffer is a protected command buffer and protectedNoFault is not supported, **any resource written** to by the VkPipeline object bound to the pipeline bind point used by this command must not be an unprotected resource

and

> VUID-vkCmdDrawIndexed-None-06539
If any recorded command in the current subpass will read from an image subresource used as an attachment in any way other than as an attachment, this command **must not write** to that image subresource as an attachment

------

>  In the new version it says descriptor write has already happened. Is it correct?

yep! that is the goal here :smile:

artem-lunarg: Thanks for the explanation. Better names it's great, almost like a bug fix.

layers/state_tracker/shader_module.cpp
```diff
@@ -1217,6 +1211,30 @@ static uint32_t CheckObjectIDFromOpLoad(
     return 0;
 }
 
+// Returns the id from Atomic access that matched the object_id, otherwise returns zero
+uint32_t SHADER_MODULE_STATE::StaticData::FindIDFromAtomic(uint32_t object_id,
```
artem-lunarg, line 1215: nitpick, but also I don't know most of the details of this code, please ignore if it does not makes sense.

`FindIDFromAtomic` correctly translates to the previous behavior, but comparing to `FindIDFromLoad` it looks a bit more artificial mostly because of `image_texel_pointer_members`. Is `image_texel_pointer_members` related to atomic operations or it's here mostly because we had that in the original code?

Another option is it to have a generic method (similar to original function but a method now).  Instead of `FindIDFromAtomic` this generic method can be called with `image_texel_pointer_members` as an argument. And then `FindIDFromLoad` helper method could be implemented by delegating to generic method with `load_members` as an argument. This assume that `FindIDFromAtomic` is not fundamental operation and is the result of code extraction.

artem-lunarg (approved):

layers/state_tracker/shader_module.cpp
spencer-lunarg: > Is image_texel_pointer_members related to atomic operations or it's here mostly because we had that in the original code?

So a `OpImageTexelPointer` is the way to touch atomic images in SPIR-V... this isn't clear anywhere, that should be added

I think there is room for a larger refactor then trying to duct tape these functions more. My main goal was to separate the fact things are accessed via atomic and things are just read as normal more

I will probably add another commit to help try to clarify this more (finally got some answers about the spec questions I had)

artem-lunarg: Okay, now it’s clear. Please ignore my comment :)

layers/state_tracker/pipeline_state.h
spencer-lunarg: ok (for sake of leaving a note somewhere), so seems the SPIR-V `NonWritable` and `NonReadable` was what this code was "trying" to convey, about to make fix, but has lost in meaning in the many iterations/refactoring of the Shader Module code over the last year

about to push a fix to clarify this

Issue #5172, opened 2023-01-26
Random segfaults with validation layers and debug util labels


andmoos:
**Environment:**
 - OS: Linux 6.1.7-1-MANJARO
 - GPU: Nvidia GTX 1080
 - Arch package version: 1.3.231.1-1

**Describe the Issue**
Random segfaults with validation layers active and debug util labels in use.

**Expected behavior**
-

**Valid Usage ID**
-

**Additional context**
- Application is threaded, so this could very well be a race condition. But access to the VkCommandPool  is serialized when e.g. passed to `vkAllocateCommandBuffers()`.

Segfault trace when run with ASAN:

```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==150433==ERROR: AddressSanitizer: SEGV on unknown address 0x0003000000c6 (pc 0x7f2a35fa6869 bp 0x607000958650 sp 0x7f2a0b281d00 T3)
==150433==The signal is caused by a READ memory access.
    #0 0x7f2a35fa6869 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>::_M_data() const /usr/include/c++/12.2.0/bits/basic_string.h:234
    #1 0x7f2a35fa6869 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>::_M_is_local() const /usr/include/c++/12.2.0/bits/basic_string.h:274
    #2 0x7f2a35fa6869 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>::_M_dispose() /usr/include/c++/12.2.0/bits/basic_string.h:285
    #3 0x7f2a35fa6869 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>::~basic_string() /usr/include/c++/12.2.0/bits/basic_string.h:795
    #4 0x7f2a35fa6869 in LoggingLabel::~LoggingLabel() /build/vulkan-validation-layers/src/Vulkan-ValidationLayers-sdk-1.3.231.1/layers/vk_layer_logging.h:138
    #5 0x7f2a35fa6869 in void std::_Destroy<LoggingLabel>(LoggingLabel*) /usr/include/c++/12.2.0/bits/stl_construct.h:151
[...]

PR #5312, opened 2023-02-20
Fix debug label segfault

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5172

The attached regression test causes a crash on most runs both on linux and windows when the fix is not applied.

### DIFF
--- a/tests/positive/command.cpp
+++ b/tests/positive/command.cpp
@@ -1258,3 +1258,51 @@ TEST_F(VkPositiveLayerTest, EventsInSecondaryCommandBuffers) {
     vk::QueueSubmit(m_device->m_queue, 1, &submit_info, VK_NULL_HANDLE);
     vk::QueueWaitIdle(m_device->m_queue);
 }
+
+TEST_F(VkPositiveLayerTest, ThreadedCommandBuffersWithLabels) {
+    AddRequiredExtensions(VK_EXT_DEBUG_UTILS_EXTENSION_NAME);
+    ASSERT_NO_FATAL_FAILURE(InitFramework());
+    if (!AreRequiredExtensionsEnabled()) {
+        GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported";
+    }
+    ASSERT_NO_FATAL_FAILURE(InitState());
+    auto vkCmdInsertDebugUtilsLabelEXT = GetInstanceProcAddr<PFN_vkCmdInsertDebugUtilsLabelEXT>("vkCmdInsertDebugUtilsLabelEXT");
+
+    constexpr int worker_count = 8;
+    ThreadTimeoutHelper timeout_helper(worker_count);
+
+    auto worker_thread = [&](int worker_index) {
+        auto timeout_guard = timeout_helper.ThreadGuard();
+        // Command pool per worker thread
+        VkCommandPoolObj pool(m_device, m_device->graphics_queue_node_index_, VK_COMMAND_POOL_CREATE_TRANSIENT_BIT);
+
+        constexpr int command_buffers_per_pool = 4;
+        auto commands_allocate_info = LvlInitStruct<VkCommandBufferAllocateInfo>();
+        commands_allocate_info.commandPool = pool.handle();
+        commands_allocate_info.commandBufferCount = command_buffers_per_pool;
+        commands_allocate_info.level = VK_COMMAND_BUFFER_LEVEL_PRIMARY;
+        const auto begin_info = LvlInitStruct<VkCommandBufferBeginInfo>();
+
+        auto label = LvlInitStruct<VkDebugUtilsLabelEXT>();
+        label.pLabelName = "Test label";
+
+        constexpr int iteration_count = 1000;
+        for (int frame = 0; frame < iteration_count; frame++) {
+            std::array<VkCommandBuffer, command_buffers_per_pool> command_buffers;
+            ASSERT_VK_SUCCESS(vk::AllocateCommandBuffers(m_device->device(), &commands_allocate_info, command_buffers.data()));
+            for (int i = 0; i < command_buffers_per_pool; i++) {
+                ASSERT_VK_SUCCESS(vk::BeginCommandBuffer(command_buffers[i], &begin_info));
+                // Record debug label. It's a required step to reproduce the original issue
+                vkCmdInsertDebugUtilsLabelEXT(command_buffers[i], &label);
+                ASSERT_VK_SUCCESS(vk::EndCommandBuffer(command_buffers[i]));
+            }
+            vk::FreeCommandBuffers(m_device->device(), pool.handle(), command_buffers_per_pool, command_buffers.data());
+        }
+    };
+    std::vector<std::thread> workers;
+    for (int i = 0; i < worker_count; i++) workers.emplace_back(worker_thread, i);
+    constexpr int wait_time = 60;
+    if (!timeout_helper.WaitForThreads(wait_time))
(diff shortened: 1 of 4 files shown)


artem-lunarg:
@jeremy-lunarg, @jzulauf-lunarg, do you know if we had a specific reason not to add mutex to `EraseCmdDebugUtilsLabel` as we do in all other places where `debugUtilsCmdBufLabels` can be modified?

juan-lunarg (commented):
Does this run cleanly if you enable thread sanitizer?

artem-lunarg:
@juan-lunarg, I have just checked thread sanitizer, it does not run cleanly. It detects data race during framework shutdown. Do you see how this test can cause this or is it a known issue?

```
WARNING: ThreadSanitizer: data race (pid=15648)
  Write of size 8 at 0x7fffffffd2c0 by main thread:
    #0 __gnu_cxx::__normal_iterator<std::pair<VkQueue_T* const, std::shared_ptr<QUEUE_STATE> >*, std::vector<std::pair<VkQueue_T* const, std::shared_ptr<QUEUE_STATE> >, std::allocator<std::pair<VkQueue_T* const, std::shared_ptr<QUEUE_STATE> > > > >::__normal_iterator(std::pair<VkQueue_T* const, std::shared_ptr<QUEUE_STATE> >* const&) /usr/include/c++/11/bits/stl_iterator.h:1011 (libVkLayer_khronos_validation.so+0x7ca031)
    #1 std::vector<std::pair<VkQueue_T* const, std::shared_ptr<QUEUE_STATE> >, std::allocator<std::pair<VkQueue_T* const, std::shared_ptr<QUEUE_STATE> > > >::begin() /usr/include/c++/11/bits/stl_vector.h:812 (libVkLayer_khronos_validation.so+0x7c7920)
    #2 ValidationStateTracker::PostCallRecordDeviceWaitIdle(VkDevice_T*, VkResult) /home/lunarg/vvl-artem-tsan/layers/state_tracker/state_tracker.cpp:2095 (libVkLayer_khronos_validation.so+0x1496ee1)
    #3 vulkan_layer_chassis::DeviceWaitIdle(VkDevice_T*) /home/lunarg/vvl-artem-tsan/layers/generated/chassis.cpp:1187 (libVkLayer_khronos_validation.so+0x830b94)
    #4 VkRenderFramework::ShutdownFramework() /home/lunarg/vvl-artem-tsan/tests/vkrenderframework.cpp:526 (vk_layer_validation_tests+0x10ab71b)
    #5 VkRenderFramework::~VkRenderFramework() /home/lunarg/vvl-artem-tsan/tests/vkrenderframework.cpp:124 (vk_layer_validation_tests+0x10a7afd)
    #6 VkLayerTest::~VkLayerTest() /home/lunarg/vvl-artem-tsan/tests/layer_validation_tests.h:246 (vk_layer_validation_tests+0x6b8311)
    #7 VkPositiveLayerTest::~VkPositiveLayerTest() /home/lunarg/vvl-artem-tsan/tests/layer_validation_tests.h:327 (vk_layer_validation_tests+0x6e16ad)
    #8 VkPositiveLayerTest_ThreadedCommandBuffersWithLabels_Test::~VkPositiveLayerTest_ThreadedCommandBuffersWithLabels_Test() /home/lunarg/vvl-artem-tsan/tests/positive/command.cpp:1262 (vk_layer_validation_tests+0x6e173f)
    #9 VkPositiveLayerTest_ThreadedCommandBuffersWithLabels_Test::~VkPositiveLayerTest_ThreadedCommandBuffersWithLabels_Test() /home/lunarg/vvl-artem-tsan/tests/positive/command.cpp:1262 (vk_layer_validation_tests+0x6e1773)
    #10 testing::Test::DeleteSelf_() /home/lunarg/vvl-artem-tsan/external/Debug/googletest/googletest/include/gtest/gtest.h:318 (vk_layer_validation_tests+0x1469a35)
    #11 main /home/lunarg/vvl-artem-tsan/tests/layer_validation_tests.cpp:3944 (vk_layer_validation_tests+0x9370a8)
```

juan-lunarg:
>  Do you see how this test can cause this or is it a known issue?

Currently it's one of my goals to have CI pass with both ASAN/TSAN. Currently it doesn't though. I mainly suggested using tsan to try and check your work since it helped me with a race condition in the past.

If it's not passing with tsan there is likely something still wrong with vvl.

artem-lunarg:
To clarify this for other reviewers. Thread sanitizer issue is not related to this specific test, it happens for other tests too.

The original issue was detected by the Address sanitizer (and also without it as a crash) and with the current fix the address sanitizer runs cleanly.

The test itself does not introduce _new_ synchronizations, the threads run independently and do not communicate. The only synchronization that happens is the one that is used inside vvl.

mark-lunarg (approved):
LGTM

tests/positive/command.cpp
```diff
@@ -1257,4 +1257,48 @@ TEST_F(VkPositiveLayerTest, EventsInSecondaryCommandBuffers) {
...
+        commands_allocate_info.level = VK_COMMAND_BUFFER_LEVEL_PRIMARY;
+        const auto begin_info = LvlInitStruct<VkCommandBufferBeginInfo>();
+
+        auto label = LvlInitStruct<VkDebugUtilsLabelEXT>();
+        label.pLabelName = "Test label";
+
+        constexpr int iteration_count = 1000;
+        for (int frame = 0; frame < iteration_count; frame++) {
+            // Allocate command buffers
+            std::array<VkCommandBuffer, command_buffers_per_pool> command_buffers;
```
ncesario-lunarg, line 1288: Might be a bit cleaner to make these `vk_testing::CommandBuffer`s so cleanup is implicit.

tests/positive/command.cpp
```diff
@@ -1257,4 +1257,48 @@ TEST_F(VkPositiveLayerTest, EventsInSecondaryCommandBuffers) {
...
+                ASSERT_VK_SUCCESS(vk::EndCommandBuffer(command_buffers[i]));
+            }
+            // Free command buffers
+            vk::FreeCommandBuffers(m_device->device(), pool.handle(), command_buffers_per_pool, command_buffers.data());
+        }
+    };
+    std::vector<std::thread> workers;
+    constexpr int worker_count = 8;
+    for (int i = 0; i < worker_count; i++) workers.emplace_back(worker_thread, i);
+    for (auto &worker : workers) worker.join();
```
ncesario-lunarg, line 1303: I think this should be good for this test, though might be slightly better to use a conditional variable with `wait_for` to ensure the test finishes, failing the test if it doesn't finish after a certain timeout.

artem-lunarg: Yes, that makes sense. I remember those ghost test instances on CI 👻 . 

It could be useful even for this test. The test does not use explicit synchronization between the threads, but there is still synchronization inside vvl (that's why we had a bug).

tests/positive/command.cpp
artem-lunarg: This test allocates multiple command buffers at once with `vk::AllocateCommandBuffers` and I can't find a helper for this. I prefer to keep this version, mostly because this was tested on two platforms and reproduces issue well. It might be tricky to predict all effects in threading environment without extensive re-resting 🧪🧪🧪

tests/positive/command.cpp
artem-lunarg: Introduced `ThreadTimeoutHelper` class to fail the tests with stuck worker threads.

ncesario-lunarg (approved):
LGTM!

PR #5315, opened 2023-02-21
tests: Add MockICD support for Ray Tracing tests

spencer-lunarg:
waiting on https://github.com/KhronosGroup/Vulkan-Tools/pull/747

Adds about 50+ tests and allows for testing most Ray Tracing without needing a supported physical device

### DIFF
--- a/tests/vklayertests_ray_tracing.cpp
+++ b/tests/vklayertests_ray_tracing.cpp
@@ -138,7 +138,7 @@ TEST_F(VkLayerTest, RayTracingAccelerationStructureBindings) {
     dslb.descriptorCount = 1;
     dslb.stageFlags = VK_SHADER_STAGE_FRAGMENT_BIT;
 
-    for (uint32_t i = 0; i < maxBlocks; ++i) {
+    for (uint32_t i = 0; i < maxBlocks + 1; ++i) {
         dslb.binding = i;
         dslb_vec.push_back(dslb);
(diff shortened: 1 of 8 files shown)


ncesario-lunarg (approved):
Wow, this is awesome!

tests/vklayertests_ray_tracing.cpp
```diff
@@ -138,7 +138,7 @@ TEST_F(VkLayerTest, RayTracingAccelerationStructureBindings) {
     dslb.descriptorCount = 1;
     dslb.stageFlags = VK_SHADER_STAGE_FRAGMENT_BIT;
 
-    for (uint32_t i = 0; i < maxBlocks; ++i) {
+    for (uint32_t i = 0; i < maxBlocks + 1; ++i) {
```
juan-lunarg, line 141: +1?

juan-lunarg (approved):

tests/vklayertests_ray_tracing.cpp
spencer-lunarg: this test is for `VUID-VkPipelineLayoutCreateInfo-descriptorType-03572`, it needs to make this larger then `maxBlocks` (`maxPerStageDescriptorUpdateAfterBindAccelerationStructures`)

I realized this test was not being ran in CI before because it was off by 1

Issue #2661, opened 2021-03-26
False-positive with validation of srcStageMask on VkSubpassDependency2


KolyaNaichuk:
**Describe the Issue**
I am using VkSubpassDependency2 with VkMemoryBarrier2KHR extension. I specify srcStageMask, srcAccessMask, dstStageMask, dstAccessMask on VkMemoryBarrier2KHR and keep srcStageMask, srcAccessMask, dstStageMask, dstAccessMask uninitialised on VkSubpassDependency2. According to the [specification](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkSubpassDependency2.html) if we include the extension VkMemoryBarrier2KHR, stage masks and access masks should be ignored on VkSubpassDependency2. This does not seem  to be the case for the validation layer as it throws this error while calling vkCreateRenderPass2.
 
"[ VUID-VkSubpassDependency2-srcStageMask-parameter ] Object 0: handle = 0x19a3847a268, name = Renderer::m_DeviceHandle, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0xe5ee6532 | vkCreateRenderPass2: value of pCreateInfo->pDependencies[0].srcStageMask contains flag bits that are not recognized members of VkPipelineStageFlagBits The Vulkan spec states: srcStageMask must be a valid combination of VkPipelineStageFlagBits values"

**Valid Usage ID**

**Environment:**
 - OS: Windows 10 Home (19042.867)
 - GPU: GTX 1070 (458.06)
 - SDK or header version if building from repo: 1.2.170.0
 - Options enabled (synchronization, best practices, etc.):

**Additional context**

ncesario-lunarg:
Thanks for pointing this out @KolyaNaichuk, it looks like we don't currently have a way of ignoring these as this validation is currently generated (see `layers/generated/parameter_validation.cpp` around line 7051). It might take a bit of effort in the generation scripts to get this resolved.

KolyaNaichuk:
Alright! No worries I am able to workaround the issue by initialising the masks to 0 on VkSubpassDependency2.

PR #5324, opened 2023-02-23
Ignore fields with VkMemoryBarrier2 pNext

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/2661

The spec says for `VkSubpassDependency2`

> If a [VkMemoryBarrier2](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkMemoryBarrier2) is included in the pNext chain, **srcStageMask**, **dstStageMask**, **srcAccessMask**, and **dstAccessMask** parameters are ignored. The synchronization and access scopes instead are defined by the parameters of [VkMemoryBarrier2](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkMemoryBarrier2).

We are already doing this in the Sync validation, but were not doing it in the parameter checking

some other notes about the PR/

- the `UNASSIGNED` were bogus, so removed
- `ConvertVkRenderPassCreateInfoToV2KHR` was not being used in the stateless code (but used 2 other places)
- the old `self.structMemberValidationConditions` was not being used anymore (the VU was updated awhile ago and code was never removed) so I just used that path to implement this "ignore" conditional logic

### DIFF
--- a/layers/core_checks/render_pass_validation.cpp
+++ b/layers/core_checks/render_pass_validation.cpp
@@ -2388,8 +2388,7 @@ bool CoreChecks::PreCallValidateCreateRenderPass(VkDevice device, const VkRender
     }
 
     if (!skip) {
-        safe_VkRenderPassCreateInfo2 create_info_2;
-        ConvertVkRenderPassCreateInfoToV2KHR(*pCreateInfo, &create_info_2);
+        safe_VkRenderPassCreateInfo2 create_info_2 = ConvertVkRenderPassCreateInfoToV2KHR(*pCreateInfo);
         skip |= ValidateCreateRenderPass(device, RENDER_PASS_VERSION_1, create_info_2.ptr(), "vkCreateRenderPass()");
     }
 
--- a/layers/core_checks/render_pass_validation.cpp
+++ b/layers/core_checks/render_pass_validation.cpp
@@ -189,6 +189,28 @@ bool CoreChecks::ValidateDependencyCompatibility(const char *type1_string, const
     const auto &primary_dep = rp1_state.createInfo.pDependencies[dependency];
     const auto &secondary_dep = rp2_state.createInfo.pDependencies[dependency];
 
+    VkPipelineStageFlags2 primary_src_stage_mask = primary_dep.srcStageMask;
+    VkPipelineStageFlags2 primary_dst_stage_mask = primary_dep.dstStageMask;
+    VkAccessFlags2 primary_src_access_mask = primary_dep.srcAccessMask;
+    VkAccessFlags2 primary_dst_access_mask = primary_dep.dstAccessMask;
+    VkPipelineStageFlags2 secondary_src_stage_mask = secondary_dep.srcStageMask;
+    VkPipelineStageFlags2 secondary_dst_stage_mask = secondary_dep.dstStageMask;
+    VkAccessFlags2 secondary_src_access_mask = secondary_dep.srcAccessMask;
+    VkAccessFlags2 secondary_dst_access_mask = secondary_dep.dstAccessMask;
+
+    if (const auto primary_barrier = LvlFindInChain<VkMemoryBarrier2KHR>(rp1_state.createInfo.pNext); primary_barrier) {
+        primary_src_stage_mask = primary_barrier->srcStageMask;
+        primary_dst_stage_mask = primary_barrier->dstStageMask;
+        primary_src_access_mask = primary_barrier->srcAccessMask;
+        primary_dst_access_mask = primary_barrier->dstAccessMask;
+    }
+    if (const auto secondary_barrier = LvlFindInChain<VkMemoryBarrier2KHR>(rp2_state.createInfo.pNext); secondary_barrier) {
+        secondary_src_stage_mask = secondary_barrier->srcStageMask;
+        secondary_dst_stage_mask = secondary_barrier->dstStageMask;
+        secondary_src_access_mask = secondary_barrier->srcAccessMask;
+        secondary_dst_access_mask = secondary_barrier->dstAccessMask;
+    }
+
     if (primary_dep.srcSubpass != secondary_dep.srcSubpass) {
         std::stringstream ss;
         ss << "First srcSubpass is " << primary_dep.srcSubpass << ", but second srcSubpass is " << secondary_dep.srcSubpass << ".";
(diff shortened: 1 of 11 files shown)


layers/stateless/parameter_validation_utils.cpp
```diff
@@ -24,6 +24,7 @@
 #include "core_validation_error_enums.h"
 #include "enum_flag_bits.h"
 #include "vk_layer_data.h"
+#include "convert_to_renderpass2.h"
```
artem-lunarg, line 27: nit: include order

artem-lunarg (approved):

layers/core_checks/render_pass_validation.cpp
```diff
@@ -189,6 +189,30 @@ bool CoreChecks::ValidateDependencyCompatibility(const char *type1_string, const
...
+    VkPipelineStageFlags2 primary_dst_stage_mask = primary_dep.dstStageMask;
+    VkAccessFlags2 primary_src_access_mask = primary_dep.srcAccessMask;
+    VkAccessFlags2 primary_dst_access_mask = primary_dep.dstAccessMask;
+    VkPipelineStageFlags2 secondary_src_stage_mask = secondary_dep.srcStageMask;
+    VkPipelineStageFlags2 secondary_dst_stage_mask = secondary_dep.dstStageMask;
+    VkAccessFlags2 secondary_src_access_mask = secondary_dep.srcAccessMask;
+    VkAccessFlags2 secondary_dst_access_mask = secondary_dep.dstAccessMask;
+
+    const auto primary_barrier = LvlFindInChain<VkMemoryBarrier2KHR>(rp1_state.createInfo.pNext);
+    if (primary_barrier) {
```
juan-lunarg, line 202:
```suggestion
    if ( const auto primary_barrier = LvlFindInChain<VkMemoryBarrier2KHR>(rp1_state.createInfo.pNext); primary_barrier) {
```

layers/core_checks/render_pass_validation.cpp
```diff
@@ -189,6 +189,30 @@ bool CoreChecks::ValidateDependencyCompatibility(const char *type1_string, const
...
+        primary_src_stage_mask = primary_barrier->srcStageMask;
+        primary_dst_stage_mask = primary_barrier->dstStageMask;
+        primary_src_access_mask = primary_barrier->srcAccessMask;
+        primary_dst_access_mask = primary_barrier->dstAccessMask;
+    }
+    const auto secondary_barrier = LvlFindInChain<VkMemoryBarrier2KHR>(rp2_state.createInfo.pNext);
+    if (secondary_barrier) {
```
juan-lunarg, line 209:
```suggestion
    if (const auto secondary_barrier = LvlFindInChain<VkMemoryBarrier2KHR>(rp2_state.createInfo.pNext); secondary_barrier) {
```

layers/stateless/parameter_validation_utils.cpp
```diff
@@ -4984,14 +4984,22 @@ bool StatelessValidation::CreateRenderPassGeneric(VkDevice device, const RenderP
                              func_name, i, dependency.dstSubpass, pCreateInfo->subpassCount);
         }
 
+        VkPipelineStageFlags2 srcStageMask = dependency.srcStageMask;
+        VkPipelineStageFlags2 dstStageMask = dependency.dstStageMask;
+        const auto barrier = LvlFindInChain<VkMemoryBarrier2KHR>(pCreateInfo->pDependencies[i].pNext);
+        if (barrier) {
```
juan-lunarg, line 4990:
```suggestion
        if (const auto barrier = LvlFindInChain<VkMemoryBarrier2KHR>(pCreateInfo->pDependencies[i].pNext); barrier) {
```

layers/stateless/parameter_validation_utils.cpp
```diff
@@ -5000,19 +5008,23 @@ bool StatelessValidation::CreateRenderPassGeneric(VkDevice device, const RenderP
 bool StatelessValidation::manual_PreCallValidateCreateRenderPass(VkDevice device, const VkRenderPassCreateInfo *pCreateInfo,
                                                                  const VkAllocationCallbacks *pAllocator,
                                                                  VkRenderPass *pRenderPass) const {
-    return CreateRenderPassGeneric(device, pCreateInfo, pAllocator, pRenderPass, RENDER_PASS_VERSION_1);
+    safe_VkRenderPassCreateInfo2 create_info_2;
+    ConvertVkRenderPassCreateInfoToV2KHR(*pCreateInfo, &create_info_2);
```
juan-lunarg, line 5012: Complete nit FFTI. Have `ConvertVkRenderPassCreateInfoToV2KHR` return a `safe_VkRenderPassCreateInfo2` instead of modifying an existing one. It would allow for better const correctness and IMO read better.

tests/vklayertests_descriptor_renderpass_framebuffer.cpp
```diff
@@ -13606,3 +13606,83 @@ TEST_F(VkLayerTest, DescriptorBufferInvalidExtensionCombination) {
     vk::CreateDevice(gpu(), &device_ci, nullptr, &testDevice);
     m_errorMonitor->VerifyFound();
 }
+
+TEST_F(VkLayerTest, IgnoreSubpassDependencyMasksSync2) {
+    // Testing from the spec:
+    // If a VkMemoryBarrier2 is included in the pNext chain,
+    // srcStageMask, dstStageMask, srcAccessMask, and dstAccessMask parameters are ignored.
+    // The synchronization and access scopes instead are defined by the parameters of VkMemoryBarrier2.
```
juan-lunarg, line 13614: I really appreciate comments on the tests :+1:

juan-lunarg (approved):

layers/stateless/parameter_validation_utils.cpp
ncesario-lunarg: :heavy_plus_sign:

ncesario-lunarg (commented):
LGTM, but will wait for other comments to get resolved.

layers/stateless/parameter_validation_utils.cpp
spencer-lunarg: We have not included this 
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/.clang-format#L7

it was discussed awhile ago, but I guess not sure why it needs to be in alphabetical order

layers/core_checks/render_pass_validation.cpp
spencer-lunarg: can change (because indifference either way), but I guess what is the reason/advantage to inline this into a single line

layers/stateless/parameter_validation_utils.cpp
spencer-lunarg: agree, made an extra commit on top to have it return the `safe_VkRenderPassCreateInfo2`

layers/core_checks/render_pass_validation.cpp
juan-lunarg: The main advantage is it reduces the scope of the variable to the if statement. Which can have slight perf advantages in some situations. Or reduce the number of variables on the stack when debugging a large function. Not a big deal though.

Issue #5080, opened 2023-01-08
`AttachmentTracker::subpass_attachment_layout` never gets updated


Rua:
In render_pass_state.cpp, in `AttachmentTracker::Update`, the variable `subpass_attachment_layout` is used to determine whether to insert a layout transition from a previous subpass to the current one, but only if the stored layout is not `kInvalidLayout`. In the constructor, this variable is initialised to contain `kInvalidLayout` for every subpass+attachment combination. But the stored value is then never updated, so the value stays at `kInvalidLayout`, which means that no layout transition ever gets added between the subpasses.

arno-lunarg:
Hi @Rua, thank you for spotting this! To potentially have some more details, did you notice something break because of this bug?

PR #5327, opened 2023-02-23
several syncval attachment and test  fixes

jzulauf-lunarg:
Addresses #5080, and test and syncval issues discovered by fixing it.

### DIFF
--- a/layers/state_tracker/render_pass_state.cpp
+++ b/layers/state_tracker/render_pass_state.cpp
@@ -170,6 +170,25 @@ struct AttachmentTracker {  // This is really only of local interest, but a bit 
         }
     }
 
+    void Update(uint32_t subpass, const uint32_t *preserved, uint32_t count) {
+        // for preserved attachment, preserve the layout from the most recent (max subpass) dependency
+        // or initial, if none
+        for (const auto attachment : vvl::make_span(preserved, count)) {
+            uint32_t max_prev = VK_SUBPASS_EXTERNAL;
+            for (const auto &prev : rp->subpass_dependencies[subpass].prev) {
+                const auto prev_pass = prev.first->pass;
+                max_prev = (max_prev == VK_SUBPASS_EXTERNAL) ? prev_pass : std::max(prev_pass, max_prev);
+            }
+
+            if (max_prev == VK_SUBPASS_EXTERNAL) {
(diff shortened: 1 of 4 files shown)


ncesario-lunarg:
Maybe convert to a draft until it's ready for review?

jzulauf-lunarg:
@jeremyg-lunarg @ncesario-lunarg -- moved to ready for review

layers/sync/sync_validation.h
```diff
@@ -1185,12 +1186,20 @@ class AccessContext {
...
     void ClearAsyncContext(const AccessContext *context) { async_.clear(); }
 
-    struct AsyncReference {
-        const AccessContext *context;
+    class AsyncReference {
+      public:
+        AsyncReference(const AccessContext &async_context, ResourceUsageTag async_tag)
+            : context_(&async_context), tag_(async_tag) {}
+        // AsyncReference(const AsyncReference &) = default;
+        // AsyncReference &operator =(const AsyncReference &) = default;
```
ncesario-lunarg, line 1194: Remove?

ncesario-lunarg (approved):
Minor comment nit, but otherwise LGTM!

layers/state_tracker/render_pass_state.cpp
```diff
@@ -170,6 +170,25 @@ struct AttachmentTracker {  // This is really only of local interest, but a bit
         }
     }
 
+    void Update(uint32_t subpass, const uint32_t *preserved, uint32_t count) {
+        // for preserved attachment, preserve the layout from the most recent (max subpass) dependency
+        // or initial, if none
+        for (const auto attachment : vvl::make_span(preserved, count)) {
+            uint32_t max_prev = VK_SUBPASS_EXTERNAL;
```
artem-lunarg, line 177: The suggestion is to compute `max_prev` outside of the loop. It looks like it does not depend on the loop variable.

jzulauf-lunarg: It's used per attachment to assess the result of the next for loop.  moving it would break that logic.

layers/sync/sync_validation.h
jzulauf-lunarg: Catch that in a cleanup.

PR #5328, opened 2023-02-23
Fix check for rasterizerDiscardEnable-00751

ChristianF-ARM:
Correct inverted logic to detect pMultisampleState == nullptr when not creating a graphics library.

This problem was hit by @sebbbi recently: https://twitter.com/SebAaltonen/status/1621144550067830784

### DIFF
--- a/layers/stateless/parameter_validation_utils.cpp
+++ b/layers/stateless/parameter_validation_utils.cpp
@@ -3092,7 +3092,7 @@ bool StatelessValidation::manual_PreCallValidateCreateGraphicsPipelines(VkDevice
                 const bool is_frag_out_graphics_lib =
                     graphics_lib_info &&
                     ((graphics_lib_info->flags & VK_GRAPHICS_PIPELINE_LIBRARY_FRAGMENT_OUTPUT_INTERFACE_BIT_EXT) != 0);
-                if (is_frag_out_graphics_lib && (create_info.pMultisampleState == nullptr)) {
+                if (!is_frag_out_graphics_lib && (create_info.pMultisampleState == nullptr)) {
                     skip |= LogError(device, "VUID-VkGraphicsPipelineCreateInfo-rasterizerDiscardEnable-00751",
                                      "vkCreateGraphicsPipelines: if pCreateInfos[%" PRIu32
                                      "].pRasterizationState->rasterizerDiscardEnable "


ChristianF-ARM:
FYI @ncesario-lunarg -- seems this was introduced by 81257cb32fc7c5ac257aaabf8881f0bd2477eef0 about a year ago.

layers/stateless/parameter_validation_utils.cpp
```diff
@@ -3092,7 +3092,7 @@ bool StatelessValidation::manual_PreCallValidateCreateGraphicsPipelines(VkDevice
                 const bool is_frag_out_graphics_lib =
                     graphics_lib_info &&
                     ((graphics_lib_info->flags & VK_GRAPHICS_PIPELINE_LIBRARY_FRAGMENT_OUTPUT_INTERFACE_BIT_EXT) != 0);
-                if (is_frag_out_graphics_lib && (create_info.pMultisampleState == nullptr)) {
+                if (!is_frag_out_graphics_lib && (create_info.pMultisampleState == nullptr)) {
```
HansKristian-Work, line 3095: Isn't this wrong?

```
VUID-VkGraphicsPipelineCreateInfo-rasterizerDiscardEnable-00751
If the pipeline is being created with [fragment output interface state](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#pipelines-graphics-subsets-fragment-output), pMultisampleState must be a valid pointer to a valid [VkPipelineMultisampleStateCreateInfo](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkPipelineMultisampleStateCreateInfo) structure
```

I'd maybe expect:

```
const bool is_frag_out_graphics_lib = !graphics_lib_info || (create_info->flags & VK_PIPELINE_CREATE_LIBRARY_BIT_KHR) == 0 || (graphics_lib_info->flags & OUTPUT_BIT) != 0;
```

HansKristian-Work: It's misleading to check if "graphics_lib" is used. The bool should reflect if there is a fragment output interface.

HansKristian-Work: rasterizerDiscardEnable will also play into this. If that's true and not dynamic, fragment output interface is ignored.

ncesario-lunarg: We really need a unit test reproducing the issue first. I can add that (assuming I understand the original issue correctly).

It looks like the original issue (from the Twitter post) was regarding fragment _shader_ state (not fragment _output_ state), so if this fixes that issue, it's not immediately clear to me how or why (without a test).

Ideally the check for this particular VUID would/should be:
```cpp
if (pipeline_state.fragment_output_state && !pipeline_state.MultisampleState()) {
    // Error 00751...
}
```
as the existence `pipeline_state.fragment_output_state` is supposed to already account for rasterizerDiscardEnable, whether or not this is a graphics library, etc. However, since 00751 is currently implemented as a stateless check (because it technically does not require state to validate), we do not have access to the pipeline state object. For clarity, it might be worth moving this check to core checks (i.e., "stateful" checking), as the potential gain in performance here is probably not worth the confusion.

ncesario-lunarg: Pretty sure 00751 is just wrong as currently implemented. I'm working on a fix, but it would still be good to have a test that encapsulates the original issue.

ncesario-lunarg: @ChristianF-ARM can you please see if https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5337 resolves the original issue?

ChristianF-ARM: Thanks for looking into this! Yes, I'll have a look :)

ChristianF-ARM: Can confirm your PR is flagging the problem
> VUID-VkGraphicsPipelineCreateInfo-rasterizerDiscardEnable-00751(ERROR / SPEC): msgNum: 433274771 - Validation Error: [ VUID-VkGraphicsPipelineCreateInfo-rasterizerDiscardEnable-00751 ] Object 0: handle = 0x55b6f310c2b0, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x19d33f93 | vkCreateGraphicsPipelines(): pCreateInfos[0] is being created without multisample state (VkGraphicsPipelineCreateInfo::pMultisampleState == null). The Vulkan spec states: If the pipeline is being created with fragment output interface state, pMultisampleState must be a valid pointer to a valid VkPipelineMultisampleStateCreateInfo structure (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkGraphicsPipelineCreateInfo-rasterizerDiscardEnable-00751)

ncesario-lunarg: Thanks for confirming @ChristianF-ARM!

ChristianF-ARM:
Closing as this is superseded by 5337 :)

Issue #5076, opened 2023-01-08
Memory leaks when call down vkCreateInstance fails or other Vulkan API calls fail inside of vkCreateInstance


charles-lunarg:
**Describe the Issue**
There are two leaks happening, one causing VVL to leak memory directly, and another causing memory leaks in the loader.

1. When VK_ERROR_OUT_OF_HOST_MEMORY is returned from the call to vkEnumeratePhysicalDevices that happens inside of VVL's `vkCreateInstance`, the validation layers cause the loader to leak memory. These Vulkan API calls are made inside `PreCallRecordCreateInstance` so require modification to the generated code. The problem is that when OOM is returned from `vkEnumeratePhysicalDevices`, the loader has successfully completed the vkCreateInstance terminator. OOM from one of the subsequent Vulkan API calls requires that VVL cleanup all vulkan objects before existing, which should only require calling vkDestroyInstance from inside VVL.

A description of the sequence of events
* Loader's vkCreateInstance (trampoline), calls into layer's vkCreateInstance
* VVL's vkCreateInstance is called, initializing local objects and setting up dispatch tables.
* VVL calls down vkCreateInstance
* Loader's terminator_CreateInstance is called and completes successfully
* Control flow returns to VVL, continues setting up objects
* `PreCallRecordCreateInstance` is called on the validation layer objects
* A vulkan API call is made, such as `vkEnumeratePhysicalDevices`, which fails with OOHM (error_out_of_host_memory)
* What happens now is: Ignore the error condition and continue as if nothing happened. 

What should happen is that VVL should catch that OOHM, delete the instance by calling vkDestroyInstance, then return immediately.

2. The structures directly leaked by VVL are the validation layer objects allocated here.
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/d5426ea20669a4d8de34dd3b89685fba73a21c4f/layers/generated/chassis.cpp#L290

The solution would be to put them in std::unique_ptr<> wrappers immediately instead of having naked pointers laying about.

AbdullahAmrSobh:
I noticed that this also happens with `vkCreateShaderModule` when `VkShaderModuleCreateInfo::pCode` is misaligned. The application's memory usage keeps growing until it the system crashes.

charles-lunarg:
@AbdullahAmrSobh Please create a separate issue so we can better track it. This issue is for memory leaks during the vkCreateInstance call and not anywhere else.

PR #5333, opened 2023-02-24
layers: Fix memory leak from early return in vkCreateInstance

charles-lunarg:
If VkResult returned something other than VK_SUCCESS, the ValidationObjects that were allocated would be leaked. This commit adds cleanup in that path.

This fixes the second memory leak in vkCreateInstance from issue #5076

### DIFF
--- a/layers/generated/chassis.cpp
+++ b/layers/generated/chassis.cpp
@@ -320,8 +320,18 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateInstance(const VkInstanceCreateInfo *pCreat
     }
 
     VkResult result = fpCreateInstance(pCreateInfo, pAllocator, pInstance);
-    if (result != VK_SUCCESS) return result;
-
+    if (result != VK_SUCCESS) {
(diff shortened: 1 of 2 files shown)


charles-lunarg:
Turns out the fix for the other issue makes this PR redundant, closing since the next PR I make will also implement this functionality (and do it a lot cleaner!)

layers/generated/chassis.cpp
```diff
@@ -320,8 +320,18 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateInstance(const VkInstanceCreateInfo *pCreat
     }
 
     VkResult result = fpCreateInstance(pCreateInfo, pAllocator, pInstance);
-    if (result != VK_SUCCESS) return result;
-
+    if (result != VK_SUCCESS) {
+        // Delete unused validation objects to avoid memory leak.
+        std::vector<ValidationObject*> local_objs = {
```
jeremyg-lunarg, line 325: Why not just move the local_objs declaration at line 360 up above here?

Issue #4322, opened 2022-07-18
Check for VUID 00017 should use the render pass instance layercount


janharaldfredriksen-arm:
**Describe the Issue**
The VU says:

```
VUID-vkCmdClearAttachments-pRects-00017
The layers specified by each element of pRects must be contained within every attachment that pAttachments refers to
```

The "contained within" statement here is intended to refer to the number of layers in the _attachment_ (i.e. the render pass instance). Currently, the validation layers check against the number of layers in the attachment's _image view_.

The Vulkan WG will clarify this (internal issue 3157).

**Valid Usage ID**
00017 (potentially it gets a new number after spec clarification)

**Environment:**
 - Any

**Additional context**

[internal MR](https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5296/diffs)
[PR VUID updated](https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4595)

mikes-lunarg:
New VUID is `VUID-vkCmdClearAttachments-pRects-06937` in #4341

PR #5335, opened 2023-02-24
Fix 06937 validation

arno-lunarg:
- Fix for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4322

Updated tests:
`VkPositiveLayerTest.ClearRectWith2DArray`: Update frame buffer layers count so that clear call is valid
`VkLayerTest.SecondaryCommandBufferClearColorAttachmentsRenderArea`: Use `vk_testing::CommandBuffer`
`VkLayerTest.InvalidClearColorAttachmentsWithMultiview`: This test now (correctly) triggers 06937 , so set it set as a desired error
`VkLayerTest.DynamicRenderingCmdClearAttachmentTests` and `VkLayerTest.CmdClearAttachmentTests`: Update tests to account for validation change, and add test cases for secondary command buffers. Adding test cases for secondary command buffers lead to the modification in `cmd_buffer_validation.cpp` to properly validate them when using dynamic rendering.

### DIFF
--- a/tests/positive/command.cpp
+++ b/tests/positive/command.cpp
@@ -1121,7 +1121,7 @@ TEST_F(VkPositiveLayerTest, ClearRectWith2DArray) {
         fbci.pAttachments = &image_view;
         fbci.width = image_ci.extent.width;
         fbci.height = image_ci.extent.height;
-        fbci.layers = 1;
+        fbci.layers = image_ci.extent.depth;
 
         vk_testing::Framebuffer framebuffer;
         framebuffer.init(*m_device, fbci);
--- a/tests/vklayertests_command.cpp
+++ b/tests/vklayertests_command.cpp
@@ -5008,18 +5008,18 @@ TEST_F(VkLayerTest, SecondaryCommandBufferClearColorAttachmentsRenderArea) {
     command_buffer_allocate_info.level = VK_COMMAND_BUFFER_LEVEL_SECONDARY;
     command_buffer_allocate_info.commandBufferCount = 1;
 
-    VkCommandBuffer secondary_command_buffer;
-    ASSERT_VK_SUCCESS(vk::AllocateCommandBuffers(m_device->device(), &command_buffer_allocate_info, &secondary_command_buffer));
-    VkCommandBufferBeginInfo command_buffer_begin_info = LvlInitStruct<VkCommandBufferBeginInfo>();
+    vk_testing::CommandBuffer secondary_command_buffer(*m_device, command_buffer_allocate_info);
     VkCommandBufferInheritanceInfo command_buffer_inheritance_info = LvlInitStruct<VkCommandBufferInheritanceInfo>();
     command_buffer_inheritance_info.renderPass = m_renderPass;
     command_buffer_inheritance_info.framebuffer = m_framebuffer;
 
+    VkCommandBufferBeginInfo command_buffer_begin_info = LvlInitStruct<VkCommandBufferBeginInfo>();
     command_buffer_begin_info.flags =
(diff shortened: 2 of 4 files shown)


layers/core_checks/image_validation.cpp
```diff
@@ -852,30 +852,28 @@ static inline bool ContainsRect(VkRect2D rect, VkRect2D sub_rect) {
...
-                (clear_rects[j].baseArrayLayer + clear_rects[j].layerCount > attachment_layer_count)) {
-                skip |= LogError(cb_state.Handle(), "VUID-vkCmdClearAttachments-pRects-06937",
-                                 "vkCmdClearAttachments(): The layers defined in pRects[%d] are not contained in the layers "
-                                 "of pAttachment[%d].",
-                                 j, attachment_index);
-            }
+        const uint32_t rect_base_layer = clear_rects[i].baseArrayLayer;
+        const uint32_t rect_layer_count = clear_rects[i].layerCount;
+        // The layer indices specified by elements of pRects must be inferior to render pass layer count
+        if ((rect_base_layer >= render_pass_layer_count) || (rect_base_layer + rect_layer_count > render_pass_layer_count)) {
```
ncesario-lunarg, line 871: If `rect_base_layer` and `rect_layer_count` are always `>= 0`, then is the `rect_base_layer >= render_pass_layer_count` check needed, or is `rect_base_layer + rect_layer_count > render_pass_layer_count` sufficient?

ncesario-lunarg (commented):
One questions, but this LGTM!

layers/core_checks/cmd_buffer_validation.cpp
```diff
@@ -867,22 +867,22 @@ bool CoreChecks::PreCallValidateCmdExecuteCommands(VkCommandBuffer commandBuffer
...
-                    //  If framebuffer for secondary CB is not NULL, then it must match active FB from primaryCB
-                    skip |= ValidateInheritanceInfoFramebuffer(commandBuffer, *cb_state, pCommandBuffers[i], *sub_cb_state,
-                                                               "vkCmdExecuteCommands()");
-                    if (!sub_cb_state->cmd_execute_commands_functions.empty()) {
-                        //  Inherit primary's activeFramebuffer and while running validate functions
-                        for (auto &function : sub_cb_state->cmd_execute_commands_functions) {
-                            skip |= function(*sub_cb_state, cb_state.get(), cb_state->activeFramebuffer.get());
+                } else if (sub_cb_state->beginInfo.flags & VK_COMMAND_BUFFER_USAGE_RENDER_PASS_CONTINUE_BIT) {
+                    if (!cb_state->activeRenderPass->UsesDynamicRendering()) {
+                        // Make sure render pass is compatible with parent command buffer pass if has continue
```
juan-lunarg, line 872: if has continue?

juan-lunarg (changes requested):
Failing Android CI

layers/core_checks/cmd_buffer_validation.cpp
arno-lunarg: yes the comment is confusing, I updated it to 
> Make sure render pass is compatible with parent command buffer pass if secondary command buffer has "continue" flag

ncesario-lunarg (approved):

arno-lunarg:
Still have what looks like a crash in `VkPositiveLayerTest.ClearRectWith2DArray` on S10 and Pixel 6...

spencer-lunarg:
> Still have what looks like a crash in VkPositiveLayerTest.ClearRectWith2DArray

I looked into this, it is crashing on `vkCmdBeginRenderPass` and I made sure all things related to the framebuffer/renderPass are not violating any VUs

I am convinced this is a driver bug and that the correct course of action is to blacklist the test on Mali Android devices (the S10 and Pixel 6) and then for me to get an issue with CTS to add a test

PR #5336, opened 2023-02-24
ci: Fix issue of workflows being cancelled on main

juan-lunarg:
Currently if 2 PRs are merged in near the same time. The new job cancels the older one on main.


juan-lunarg:
> Currently if 2 PRs are merged in near the same time. The new job cancels the older one on main.

This is what I am referring to:

image

ncesario-lunarg (approved):

PR #5337, opened 2023-02-24
layers: Fix 00751 check

ncesario-lunarg:
This change also fixes GPL sub-state tracking where some sub-states are
allowed to be missing. Specifically:
- Vertex input state is not necessary if there is no vertex shader
  present
- Fragment shader/output state is not necessary if rasterizer discard
  enabled

### DIFF
--- a/layers/core_checks/pipeline_validation.cpp
+++ b/layers/core_checks/pipeline_validation.cpp
@@ -427,12 +427,12 @@ bool CoreChecks::ValidateGraphicsPipelineLibrary(const PIPELINE_STATE &pipeline)
             if ((gpl_info->flags & VK_GRAPHICS_PIPELINE_LIBRARY_PRE_RASTERIZATION_SHADERS_BIT_EXT) &&
                 !(gpl_info->flags & VK_GRAPHICS_PIPELINE_LIBRARY_FRAGMENT_SHADER_BIT_EXT)) {
                 pre_raster_flags.first =
-                    (pipeline.pre_raster_state->pipeline_layout) ? pipeline.pre_raster_state->pipeline_layout->CreateFlags() : 0;
+                    (pipeline.PreRasterPipelineLayoutState()) ? pipeline.PreRasterPipelineLayoutState()->CreateFlags() : 0;
                 pre_raster_flags.second = {GPLInitInfo::from_gpl_info, pipeline.PreRasterPipelineLayoutState().get()};
             } else if ((gpl_info->flags & VK_GRAPHICS_PIPELINE_LIBRARY_FRAGMENT_SHADER_BIT_EXT) &&
                        !(gpl_info->flags & VK_GRAPHICS_PIPELINE_LIBRARY_PRE_RASTERIZATION_SHADERS_BIT_EXT)) {
-                fs_flags.first = (pipeline.fragment_shader_state->pipeline_layout)
-                                     ? pipeline.fragment_shader_state->pipeline_layout->CreateFlags()
+                fs_flags.first = (pipeline.FragmentShaderPipelineLayoutState())
+                                     ? pipeline.FragmentShaderPipelineLayoutState()->CreateFlags()
                                      : 0;
                 fs_flags.second = {GPLInitInfo::from_gpl_info, pipeline.FragmentShaderPipelineLayoutState().get()};
             }
@@ -452,11 +452,11 @@ bool CoreChecks::ValidateGraphicsPipelineLibrary(const PIPELINE_STATE &pipeline)
                     const VkPipelineCreateFlags lib_pipeline_flags = lib->create_flags;
                     if (lib->PipelineLayoutState()) {
                         if (lib->graphics_lib_type == VK_GRAPHICS_PIPELINE_LIBRARY_PRE_RASTERIZATION_SHADERS_BIT_EXT) {
-                            pre_raster_flags.first = lib->PipelineLayoutState()->CreateFlags();
-                            pre_raster_flags.second = {GPLInitInfo::from_link_info, lib->PipelineLayoutState().get()};
+                            pre_raster_flags.first = lib->PreRasterPipelineLayoutState()->CreateFlags();
+                            pre_raster_flags.second = {GPLInitInfo::from_link_info, lib->PreRasterPipelineLayoutState().get()};
                         } else if (lib->graphics_lib_type == VK_GRAPHICS_PIPELINE_LIBRARY_FRAGMENT_SHADER_BIT_EXT) {
-                            fs_flags.first = lib->PipelineLayoutState()->CreateFlags();
-                            fs_flags.second = {GPLInitInfo::from_link_info, lib->PipelineLayoutState().get()};
+                            fs_flags.first = lib->FragmentShaderPipelineLayoutState()->CreateFlags();
+                            fs_flags.second = {GPLInitInfo::from_link_info, lib->FragmentShaderPipelineLayoutState().get()};
                         }
                     }
 
@@ -2319,6 +2319,14 @@ bool CoreChecks::ValidateGraphicsPipelineMultisampleState(const PIPELINE_STATE &
             }
         }
     }
+
+    if (!multisample_state && pipeline.fragment_output_state) {
+        skip |= LogError(device, "VUID-VkGraphicsPipelineCreateInfo-rasterizerDiscardEnable-00751",
+                         "vkCreateGraphicsPipelines(): pCreateInfos[%" PRIu32
+                         "] is being created without multisample state (VkGraphicsPipelineCreateInfo::pMultisampleState == null).",
+                         pipeline.create_index);
+    }
+
     return skip;
 }
 
@@ -3169,8 +3177,7 @@ bool CoreChecks::PreCallValidateCmdBindPipeline(VkCommandBuffer commandBuffer, V
 
             if (cb_state->activeRenderPass && phys_dev_ext_props.sample_locations_props.variableSampleLocations == VK_FALSE) {
                 const auto *multisample_state = pipeline_state.MultisampleState();
-                const auto *sample_locations =
-                    LvlFindInChain<VkPipelineSampleLocationsStateCreateInfoEXT>(multisample_state->pNext);
+                const auto *sample_locations = LvlFindInChain<VkPipelineSampleLocationsStateCreateInfoEXT>(multisample_state);
                 if (sample_locations && sample_locations->sampleLocationsEnable == VK_TRUE &&
                     !pipeline_state.IsDynamic(VK_DYNAMIC_STATE_SAMPLE_LOCATIONS_EXT)) {
                     const VkRenderPassSampleLocationsBeginInfoEXT *sample_locations_begin_info =
@@ -3843,7 +3850,7 @@ bool CoreChecks::ValidatePipelineDrawtimeState(const LAST_BOUND_STATE &state, co
         }
     }
 
-    if (pipeline.fragment_output_state->dual_source_blending && cb_state.activeRenderPass) {
+    if (pipeline.fragment_output_state && pipeline.fragment_output_state->dual_source_blending && cb_state.activeRenderPass) {
         uint32_t count = cb_state.activeRenderPass->UsesDynamicRendering()
                              ? cb_state.activeRenderPass->dynamic_rendering_begin_rendering_info.colorAttachmentCount
                              : cb_state.activeRenderPass->createInfo.pSubpasses[cb_state.activeSubpass].colorAttachmentCount;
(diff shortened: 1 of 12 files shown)


layers/state_tracker/pipeline_state.h
```diff
@@ -273,7 +273,16 @@ class PIPELINE_STATE : public BASE_NODE {
     }
 
     bool IsGraphicsLibrary() const { return !HasFullState(); }
-    bool HasFullState() const { return vertex_input_state && pre_raster_state && fragment_shader_state && fragment_output_state; }
+    bool HasFullState() const {
+        // If Pre-raster state does contain a vertex shader, vertex input state is not required
+        const bool vi_satisfied = (pre_raster_state->vertex_shader) ? static_cast<bool>(vertex_input_state) : true;
```
artem-lunarg, line 278: 1. if `pre_raster_state` can be null then check for null is needed before dereferencing.
2. Do we need to swap operands in ternary selection?

ncesario-lunarg: Yep, good catch, thanks! I don't think the ternary operands need to be swapped, but I will expand that check to make it clearer.

layers/state_tracker/pipeline_state.h
```diff
@@ -273,7 +273,24 @@ class PIPELINE_STATE : public BASE_NODE {
     }
 
     bool IsGraphicsLibrary() const { return !HasFullState(); }
-    bool HasFullState() const { return vertex_input_state && pre_raster_state && fragment_shader_state && fragment_output_state; }
+    bool HasFullState() const {
+        // If Pre-raster state does contain a vertex shader, vertex input state is not required
+        const bool vi_satisfied = [this]() -> bool {
```
ncesario-lunarg, line 278: I vaguely remember @arno-lunarg running some perf numbers on this approach and there being no measurable difference. That said, I could not recall if it was agreed this was "readably acceptable." If not, I'll change it to not use the lambda (though I do like the fact that I can keep `vi_satisfied` `const`).

layers/state_tracker/pipeline_state.h
artem-lunarg: I probably was confused by the comment: `// If Pre-raster state does contain ...`. Should it be "does not contain" ?

layers/state_tracker/pipeline_state.h
juan-lunarg: I personally like the const correctness.

layers/state_tracker/pipeline_state.h
```diff
@@ -273,7 +273,24 @@ class PIPELINE_STATE : public BASE_NODE {
...
+                return static_cast<bool>(vertex_input_state);
+            } else {
+                // there is no vertex shader, so no vertex input state is required
+                return true;
+            }
+        }();
+
+        // Fragment output/shader state is not required if rasterization is disabled.
+        const bool frag_shader_satisfied = RasterizationDisabled() ? true : static_cast<bool>(fragment_shader_state);
+        const bool frag_out_satisfied = RasterizationDisabled() ? true : static_cast<bool>(fragment_output_state);
```
juan-lunarg, line 290: minor code duplication

layers/state_tracker/pipeline_state.h
```diff
@@ -273,7 +273,24 @@ class PIPELINE_STATE : public BASE_NODE {
...
+
+        return vi_satisfied && pre_raster_state && frag_shader_satisfied && frag_out_satisfied;
```
juan-lunarg, line 292: Nitpick FTTI this should be sorted from cheapest to evaluate to most expensive. Due to short circuit evaluation.

pre_raster_state being the cheapest, and vi_satisfied being the most expensive

layers/stateless/parameter_validation_utils.cpp
```diff
@@ -3089,152 +3089,140 @@ bool StatelessValidation::manual_PreCallValidateCreateGraphicsPipelines(VkDevice
...
+                            skip |= LogError(device, "VUID-VkPipelineMultisampleStateCreateInfo-sampleShadingEnable-00784",
+                                             "vkCreateGraphicsPipelines(): parameter "
+                                             "pCreateInfos[%" PRIu32 "].pMultisampleState->sampleShadingEnable.",
+                                             i);
+                        }
+                        // TODO Add documentation issue about when minSampleShading must be in range and when it is ignored
+                        // For now a "least noise" test *only* when sampleShadingEnable is VK_TRUE.
+                        if (!in_inclusive_range(create_info.pMultisampleState->minSampleShading, 0.F, 1.0F)) {
+                            skip |= LogError(device,
+
```
juan-lunarg, line 3107:
```suggestion
```

juan-lunarg (approved):

layers/state_tracker/pipeline_state.h
juan-lunarg: Although this one can still be written via a ternary operator can't it?

layers/core_checks/pipeline_validation.cpp
```diff
@@ -368,6 +368,13 @@ bool CoreChecks::ValidatePipeline(std::vector<std::shared_ptr<PIPELINE_STATE>> c
             pipeline.create_index);
     }
 
+    if (!pipeline.MultisampleState() && pipeline.fragment_output_state) {
```
spencer-lunarg, line 371: I would move this to `ValidateGraphicsPipelineMultisampleState`

layers/state_tracker/pipeline_state.cpp
```diff
@@ -340,7 +340,21 @@ std::shared_ptr<PreRasterState> PIPELINE_STATE::CreatePreRasterState(const PIPEL
 std::shared_ptr<FragmentShaderState> PIPELINE_STATE::CreateFragmentShaderState(
     const PIPELINE_STATE &p, const ValidationStateTracker &state, const VkGraphicsPipelineCreateInfo &create_info,
     const safe_VkGraphicsPipelineCreateInfo &safe_create_info, const std::shared_ptr<const RENDER_PASS_STATE> &rp) {
+    // There is no fragment shader state if there is no fragment shader
+    bool contains_frag_shader = false;
```
spencer-lunarg, line 344: can we just do

```patch
       library_create_info(LvlFindInChain<VkPipelineLibraryCreateInfoKHR>(PNext())),
+      stage_state(GetStageStates(*state_data, *this, csm_states)),
+      active_shaders(GetActiveShaders(stage_state)),
+      linking_shaders(GetLinkingShaders(library_create_info, *state_data)),
       graphics_lib_type(GetGraphicsLibType(create_info.graphics)),
       vertex_input_state(CreateVertexInputState(*this, *state_data, create_info.graphics)),
       pre_raster_state(CreatePreRasterState(*this, *state_data, create_info.graphics, rpstate)),
       fragment_shader_state(CreateFragmentShaderState(*this, *state_data, *pCreateInfo, create_info.graphics, rpstate)),
       fragment_output_state(CreateFragmentOutputState(*this, *state_data, *pCreateInfo, create_info.graphics, rpstate)),
-      stage_state(GetStageStates(*state_data, *this, csm_states)),
-      active_shaders(GetActiveShaders(stage_state)),
-      linking_shaders(GetLinkingShaders(library_create_info, *state_data)),
```

and this can just be a `if (active_shaders & VK_SHADER_STAGE_FRAGMENT_BIT)` check

layers/state_tracker/pipeline_state.h
spencer-lunarg: if you do this, please add a comment explaining the order was given with purpose for the next person who has to touch it :smile:

layers/state_tracker/pipeline_state.h
ncesario-lunarg: Yes, it can, but it was not very clear as a ternary.

layers/state_tracker/pipeline_state.h
ncesario-lunarg: I agree, but not sure how to do it without getting rid of the const and/or making it less clear.

juan-lunarg:
```
const bool rasterization_disabled = RasterizationDisabled();
const bool frag_shader_satisfied = rasterization_disabled ? true : static_cast<bool>(fragment_shader_state);
const bool frag_out_satisfied = rasterization_disabled ? true : static_cast<bool>(fragment_output_state);
```

layers/state_tracker/pipeline_state.h
juan-lunarg: What about:
```
const bool foobar = (pre_raster_state && pre_raster_state->vertex_shader);
const bool vi_satisfied = foobar ? return static_cast<bool>(vertex_input_state) : true;
```

Not sure what the best name for foobar would be though.

layers/state_tracker/pipeline_state.h
ncesario-lunarg: :+1: yep, that looks good. I forgot that `RasterizationDisabled()` was not trivial.

layers/state_tracker/pipeline_state.h
ncesario-lunarg: Is that better, or ordering based on which is most likely to be false (in theory avoiding some evaluations)? Asking because I've never done any runtime analysis on it.

I'm kind of curious if the order here matters much at all when branch prediction is involved?

layers/state_tracker/pipeline_state.cpp
ncesario-lunarg: `GetStageStates` (and transitively `GetActiveShaders`) currently relies on the logic that sets the various sub-state pointers, so that won't work unfortunately. This could be re-worked, but leaving that for a future exercise.

layers/state_tracker/pipeline_state.h
ncesario-lunarg: Looks like the assembly output is pretty similar between both approaches (when optimized). I think expanding the code with the lambda is a bit clearer/more readable, so I'd lean towards that approach personally for this particular case, but I don't feel particularly strongly about it if you think ternary route is markedly better.

juan-lunarg: Either way works for me. I was just having fun thinking about the possibilities :)

layers/state_tracker/pipeline_state.cpp
```diff
@@ -333,7 +333,21 @@ std::shared_ptr<PreRasterState> PIPELINE_STATE::CreatePreRasterState(const PIPEL
...
```
ncesario-lunarg, line 337: I forgot about https://gitlab.khronos.org/vulkan/vulkan/-/issues/3178, which led me to think this change would cause a bunch of new errors in CTS. Oddly, that appears to not be the case (at least not in the [tests](https://gitlab.khronos.org/vulkan/vulkan/-/issues/3178#note_370520) where @Tony-LunarG originally identified problems). I'm not sure if this is due to all the validation fixes that have gone into CTS recently or some other reason. I have a feeling this is still going to be a problem because I had to fixup some VVL tests that only had a VS, but rasterization was enabled. I'll run a few real world tests before moving ahead with merging this.

layers/state_tracker/pipeline_state.cpp
spencer-lunarg: Ok, I will most likely want to clean this up myself after this gets in then because I hate how we have more than one way to get which shader stages are used

spencer-lunarg (approved):

layers/state_tracker/pipeline_state.cpp
```diff
@@ -334,6 +334,7 @@ std::shared_ptr<FragmentShaderState> PIPELINE_STATE::CreateFragmentShaderState(
     const PIPELINE_STATE &p, const ValidationStateTracker &state, const VkGraphicsPipelineCreateInfo &create_info,
     const safe_VkGraphicsPipelineCreateInfo &safe_create_info, const std::shared_ptr<const RENDER_PASS_STATE> &rp) {
     const auto lib_type = GetGraphicsLibType(create_info);
+
```
juan-lunarg, line 337: nitpick

```suggestion
```

tests/vkrenderframework.h
```diff
@@ -787,6 +787,7 @@ class VkPipelineObj : public vk_testing::Pipeline {
     void SetViewport(const std::vector<VkViewport> &viewports);
     void SetScissor(const std::vector<VkRect2D> &scissors);
     void SetLineState(const VkPipelineRasterizationLineStateCreateInfoEXT *line_state);
+    constexpr void DisableRasterization() { m_rs_state.rasterizerDiscardEnable = VK_TRUE; }
```
juan-lunarg, line 790: constexpr?

layers/state_tracker/pipeline_state.cpp
```diff
@@ -345,6 +346,24 @@ std::shared_ptr<FragmentShaderState> PIPELINE_STATE::CreateFragmentShaderState(
...
+            }
+
+            // No fragment shader _should_ imply no fragment shader state, however, for historical (GL) reasons, a pipeline _can_
+            // be created without a VS but no FS and still have valid fragment shader state. Here we try to infer if this is the
+            // case based on whether or not depth is written to.
+            // See https://gitlab.khronos.org/vulkan/vulkan/-/issues/3178 for more details.
+            if (!contains_frag_shader && !(create_info.pDepthStencilState && create_info.pDepthStencilState->depthTestEnable &&
+                                           create_info.pDepthStencilState->depthWriteEnable)) {
+                return {};
+            }
```
juan-lunarg, line 365: :+1:

juan-lunarg (approved):
LGTM

tests/vkrenderframework.h
ncesario-lunarg: Yeah, I'm not sure why I added that, nor why there wasn't a compiler error.

juan-lunarg: Most likely because we allow the compiler to be permissive

PR #5338, opened 2023-02-25
chassis: Fix memory leak in vkCreateInstance if an error occured

### LINKED ISSUE #5076: Memory leaks when call down vkCreateInstance fails or other Vulkan API calls fail inside of vkCreateInstance

charles-lunarg:
If an error occurs duing the call down the CreateInstance chain, VVL must
clean up any allocated objects. This is made easier by only allocating
Validation Objects for which the filters allow, rather than the previous
method of allocating everything and removing the unused ones. While from a
first glance std::unique_ptr<> seems like the obvious choice, because of
how intrusive such a change would be it was not chosen.

The second place an error can occur is during PostCallRecordCreateInstance.
This results in faileding a vulkan function call after a successful call
down the vkCreateInstance chain, leaving Vulkan in a weird state. The
previous commit would not pass the error code upwards, so to the application
everything appeared correct since the return value was VK_SUCCESS. However,
this causes CTS Out Of Memory tests to fail due to OOM not being returned.
The fix for this is to make the PostCallRecordCreateInstance pass the
VkResult value by reference instead of value, allowing the error code to
propagate upwards. This did require special casing the PostCallRecord
in the various code generators.

Lastly, the commit does a general cleanup of the ValidationObject code by
removing the functions only used during initialization. Since they are used
only during vkCreateInstance and vkCreateDevice, moving them into those
functions cleans up the overal interface of the functions.

This commit makes the following tests have clean output when run with a
leak sanitizer:
* VkLayerTest.InstanceExtensionDependencies
* VkLayerTest.InstanceBadStype
* VkLayerTest.InstanceDuplicatePnextStype
* VkLayerTest.InstanceAppInfoBadStype
* VkLayerTest.InstanceValidationFeaturesBadFlags
* VkLayerTest.InstanceBadValidationFlags

Note that to get the clean output, the loader must have been updated to
contain the commit 3db19f3 (or just run with a loader built against
1.3.243 or higher).

Fixes #5076

### DIFF
--- a/layers/generated/chassis.cpp
+++ b/layers/generated/chassis.cpp
@@ -278,29 +278,43 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateInstance(const VkInstanceCreateInfo *pCreat
     std::vector<ValidationObject*> local_object_dispatch;
 
     // Add VOs to dispatch vector. Order here will be the validation dispatch order!
-    auto thread_checker_obj = new ThreadSafety(nullptr);
-    thread_checker_obj->RegisterValidationObject(!local_disables[thread_safety], api_version, report_data, local_object_dispatch);
+    if (!local_disables[thread_safety]) {
+        local_object_dispatch.emplace_back(new ThreadSafety(nullptr));
+    }
 
-    auto parameter_validation_obj = new StatelessValidation;
-    parameter_validation_obj->RegisterValidationObject(!local_disables[stateless_checks], api_version, report_data, local_object_dispatch);
+    if (!local_disables[stateless_checks]) {
+        local_object_dispatch.emplace_back(new StatelessValidation);
+    }
 
-    auto object_tracker_obj = new ObjectLifetimes;
-    object_tracker_obj->RegisterValidationObject(!local_disables[object_tracking], api_version, report_data, local_object_dispatch);
+    if (!local_disables[object_tracking]) {
+        local_object_dispatch.emplace_back(new ObjectLifetimes);
+    }
 
-    auto core_checks_obj = new CoreChecks;
-    core_checks_obj->RegisterValidationObject(!local_disables[core_checks], api_version, report_data, local_object_dispatch);
+    if (!local_disables[core_checks]) {
+        local_object_dispatch.emplace_back(new CoreChecks);
+    }
 
-    auto best_practices_obj = new BestPractices;
-    best_practices_obj->RegisterValidationObject(local_enables[best_practices], api_version, report_data, local_object_dispatch);
+    if (local_enables[best_practices]) {
+        local_object_dispatch.emplace_back(new BestPractices);
+    }
 
-    auto gpu_assisted_obj = new GpuAssisted;
-    gpu_assisted_obj->RegisterValidationObject(local_enables[gpu_validation], api_version, report_data, local_object_dispatch);
+    if (local_enables[gpu_validation]) {
+        local_object_dispatch.emplace_back(new GpuAssisted);
+    }
 
-    auto debug_printf_obj = new DebugPrintf;
-    debug_printf_obj->RegisterValidationObject(local_enables[debug_printf], api_version, report_data, local_object_dispatch);
+    if (local_enables[debug_printf]) {
+        local_object_dispatch.emplace_back(new DebugPrintf);
+    }
 
-    auto sync_validation_obj = new SyncValidator;
-    sync_validation_obj->RegisterValidationObject(local_enables[sync_validation], api_version, report_data, local_object_dispatch);
+    if (local_enables[sync_validation]) {
+        local_object_dispatch.emplace_back(new SyncValidator);
+    }
+
+    // Initialize the validation objects
+    for (ValidationObject* intercept : local_object_dispatch) {
+        intercept->api_version = api_version;
+        intercept->report_data = report_data;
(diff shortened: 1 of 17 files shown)


layers/best_practices/best_practices_validation.h
```diff
@@ -904,10 +904,6 @@ class BestPractices : public ValidationStateTracker {
         return std::static_pointer_cast<SWAPCHAIN_NODE>(std::make_shared<bp_state::Swapchain>(this, create_info, swapchain));
     }
 
-    std::shared_ptr<PHYSICAL_DEVICE_STATE> CreatePhysicalDeviceState(VkPhysicalDevice phys_dev) final {
```
ncesario-lunarg, line 907: Why was this removed? If I understand correctly, best practices will not get the correct derived state without this function.

layers/state_tracker/state_tracker.cpp
```diff
@@ -4048,7 +4044,7 @@ void ValidationStateTracker::PostCallRecordCreateInstance(const VkInstanceCreate
     }
 
     for (auto physdev : physdev_handles) {
-        Add(CreatePhysicalDeviceState(physdev));
+        Add(std::make_shared<PHYSICAL_DEVICE_STATE>(physdev));
```
ncesario-lunarg, line 4047: I might be missing something, but I don't think this is correct. See comment above.

layers/best_practices/best_practices_validation.h
charles-lunarg: I misunderstood the code, yes this shouldn't be removed. I did this before realizing that the `ValidationStateTracker` object is a base class for others, rather than a leaf object.

layers/state_tracker/state_tracker.cpp
charles-lunarg: Yep, will do the same here.

charles-lunarg:
Okay it appears I can't re-request a review from both reviewers. Apologies for the spam.

Regardless, the code now passes tests. I went ahead and ran the code with leak-sanitizer and found several test cases which triggered a different leak, whose fix was also added.

Additionally, I fixed a couple of very low hanging fruit memory leaks, let me know if those should be in their own PR.

layers/generated/chassis.cpp
```diff
@@ -312,16 +326,32 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateInstance(const VkInstanceCreateInfo *pCreat
...
         auto lock = intercept->ReadLock();
         skip |= intercept->PreCallValidateCreateInstance(pCreateInfo, pAllocator, pInstance);
-        if (skip) return VK_ERROR_VALIDATION_FAILED_EXT;
+        if (skip) {
+            for (ValidationObject* object : local_object_dispatch) {
+                delete object;
+            }
+            DeactivateInstanceDebugCallbacks(report_data);
+            FreePnextChain(report_data->instance_pnext_chain);
+            LayerDebugUtilsDestroyInstance(report_data);
```
juan-lunarg, line 335: Code duplication issue regarding freeing memory. Ideally this would be a function.

charles-lunarg:
I am on the fence about this PR.
Its... big, and not because it has to be. There are 3 separate things being fixed here, VkCreateInstance leaking when errors occur, vkCreateInstance not propagating errors up from PostCallRecordCreateInstance, and misc memory leak fixes in some tests. I'm of the mind of splitting this into three separate PR's, and I'll probably do it 'right' as well, since I'm not sold on the fix for PostCallRecordCreateInstance.

So, I will close this PR and submit smaller ones to replace it.

PR #5339, opened 2023-02-25
tests: Add Video to Max Profile CI

spencer-lunarg:
With https://github.com/KhronosGroup/Vulkan-Tools/pull/750 adds support for most Video tests to run on the Max Profile CI

brings testing to **1571** pass, **158** skipped

### DIFF
--- a/tests/device_profiles/max_profile.json
+++ b/tests/device_profiles/max_profile.json
@@ -2814,7 +2814,11 @@
                             "VK_FORMAT_FEATURE_2_STORAGE_WRITE_WITHOUT_FORMAT_BIT",
                             "VK_FORMAT_FEATURE_2_SAMPLED_IMAGE_DEPTH_COMPARISON_BIT",
                             "VK_FORMAT_FEATURE_2_FRAGMENT_DENSITY_MAP_BIT_EXT",
-                            "VK_FORMAT_FEATURE_2_FRAGMENT_SHADING_RATE_ATTACHMENT_BIT_KHR"
+                            "VK_FORMAT_FEATURE_2_FRAGMENT_SHADING_RATE_ATTACHMENT_BIT_KHR",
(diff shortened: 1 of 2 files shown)


tests/device_profiles/max_profile.json
```diff
@@ -2814,7 +2814,11 @@
                             "VK_FORMAT_FEATURE_2_STORAGE_WRITE_WITHOUT_FORMAT_BIT",
                             "VK_FORMAT_FEATURE_2_SAMPLED_IMAGE_DEPTH_COMPARISON_BIT",
                             "VK_FORMAT_FEATURE_2_FRAGMENT_DENSITY_MAP_BIT_EXT",
-                            "VK_FORMAT_FEATURE_2_FRAGMENT_SHADING_RATE_ATTACHMENT_BIT_KHR"
```
artem-lunarg, line 2817: thanks json for this line!

tests/device_profiles/max_profile.json
```diff
@@ -2860,7 +2868,11 @@
...
+                            "VK_FORMAT_FEATURE_2_FRAGMENT_SHADING_RATE_ATTACHMENT_BIT_KHR",
```
artem-lunarg, line 2871: This feature was added in one of the previous commits,  but it is still worth to mention. The spec does not allow using this feature with buffers. Not sure about R8G8B8A8_UNORM images, they mention about UINT.
> An implementation must not set this feature for formats with numeric type other than *UINT, or set it as a buffer feature.

spencer-lunarg: so yes you are correct

- Some tests we just look for the first format feature and if it is "possible" don't change how we test it
- we only really need `optimalTilingFeatures` support here for Video tests
- I had this strange "worry" it would be better to just add it for all of them

I will remove these, I realize a lot of these format features didn't need to be duplicated across all 3 features, but it was easier to try and get all the tests running with the profile

artem-lunarg (approved):

PR #5344, opened 2023-02-27
layers: Label Buffer View Atomic VUID

spencer-lunarg:
The 1.3.242 spec added `VUID-vkCmdDraw-None-07888` (and other draw/dispatches)


artem-lunarg (approved):
LGTM

PR #5347, opened 2023-02-27
layers: Cleanup variable passed around shader module val

spencer-lunarg:
`pStage->stage` and `safe_VkPipelineShaderStageCreateInfo const* pStage` were being passed around but I realized most places just need one part of the struct


layers/core_checks/shader_validation.cpp
```diff
@@ -948,11 +943,9 @@ bool CoreChecks::ValidateShaderStageInputOutputLimits(const SHADER_MODULE_STATE
                          pipeline.create_index);
     }
 
-    bool strip_output_array_level =
-        (pStage->stage == VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT || pStage->stage == VK_SHADER_STAGE_MESH_BIT_NV);
-    bool strip_input_array_level =
-        (pStage->stage == VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT ||
-         pStage->stage == VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT || pStage->stage == VK_SHADER_STAGE_GEOMETRY_BIT);
+    bool strip_output_array_level = (stage == VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT || stage == VK_SHADER_STAGE_MESH_BIT_NV);
```
ncesario-lunarg, line 946: Since this line is already modified, would it make sense to use the `EXT` version of `VK_SHADER_STAGE_MESH_BIT`?

ncesario-lunarg (approved):
LGTM!

juan-lunarg (approved):
:+1:

layers/core_checks/shader_validation.cpp
spencer-lunarg: Let me do this as a quick follow-up commit now with all MESH things, that way easy to add to the `.git-blame` ignore

PR #5349, opened 2023-02-27
layers: Suppress incorrect validation error for buffer_reference

piofra01-arm:
SHADER_MODULE_STATE::GetBaseType returns zero when passed a physical storage buffer to a struct, which later results in a validation error, despite no check whether structs passed between shader stages are compatible. This change suppresses the validation error.

Conceptually, this change addresses a wider spectrum of cases, where in the `TypesMatch` function we failed to identify both of the compared types, and would return `false` (types don't match). Currently the only clear path for this to happen is dealing with a physical storage buffer to a struct, in which case a potentially unhelpful validation error is produced:

```
Validation Error: [ UNASSIGNED-CoreValidation-Shader-InterfaceTypeMismatch ] Object 0: handle = 0x9294aa0000036c76, type = VK_OBJECT_TYPE_SHADER_MODULE; | MessageID = 0xb6cf33fe | Type mismatch on location 0.0, between vertex shader and fragment shader: 'ptr to Output ptr to PhysicalStorageBuffer struct of (sint32, sint32)' vs 'ptr to Input ptr to PhysicalStorageBuffer struct of (sint32, sint32)'
```

This logic seems to hold for a general case, where both types are unknown - since we failed to identify parameter types, we cannot validate their compatibility.

Returning `true` in this case leads to some counter-intuitive logic, where `true` doesn't mean we verified the types match, rather, that we're not certain that they don't match. If this is not acceptable, I can refactor this to also be able to express the case where we were unable to compare the types.

Otherwise, `TypesMatch` is only used in one place, and is static within the `shader_validation.cpp` file, so this workaround doesn't introduce issues elsewhere.

Conceptually, failing to identify types in `TypesMatch` likely warrants its own validation warnings, but handling this is beyond the scope of this pull request.

An ideal solution would be to implement a mechanism to compare this type of parameters being passed between shader stages, at least for cases where they can be resolved, and validate struct compatibility. The above validation error shows the types are known, and [BaseTypesMatch](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/v1.3.235/layers/shader_validation.cpp#L57) already contains the code for comparing structs. This is also beyond the scope of this pull request.

### DIFF
--- a/layers/core_checks/shader_validation.cpp
+++ b/layers/core_checks/shader_validation.cpp
@@ -90,6 +90,8 @@ static bool TypesMatch(const SHADER_MODULE_STATE &a, const SHADER_MODULE_STATE &
     const Instruction *a_base_insn = GetBaseTypeInstruction(a, a_type);
     const Instruction *b_base_insn = GetBaseTypeInstruction(b, b_type);
 
+    if (nullptr == a_base_insn && nullptr == b_base_insn) return true;
+
     return BaseTypesMatch(a, b, a_base_insn, b_base_insn);
 }
 


juan-lunarg:
Minor issue with the formatting of the commit see action for feedback.

spencer-lunarg:
I really want a test to prevent regression of this. Do you have a SPIR-V example that was hitting this, I am willing to add the test if you don't want to

piofra01-arm:
Here are the shaders used in the test where the issue was spotted:

[5349.vert.txt](https://github.com/KhronosGroup/Vulkan-ValidationLayers/files/10870114/5349.vert.txt)

[5349.frag.txt](https://github.com/KhronosGroup/Vulkan-ValidationLayers/files/10870113/5349.frag.txt)

The test mostly just draws a quad, the only twist is that it uses data from the shared buffer in the fragment shader. It doesn't do anything special, besides setting up basic scaffolding to create the buffer and pass it to the vertex shader.

Specifically:

- The shared buffer is created with usage flags: `VK_BUFFER_USAGE_SHADER_DEVICE_ADDRESS_BIT_KHR | VK_BUFFER_USAGE_STORAGE_BUFFER_BIT` and alloc flag: `VK_MEMORY_ALLOCATE_DEVICE_ADDRESS_BIT_KHR`. 
- We instantiate some test data and upload it to the buffer. 
- Then the shared buffer is referenced with a `VkDeviceAddress`, whose value is then stored in yet another buffer that is passed to the vertex shader. 

The input description/state looks as follows:
```
	/* Vertex input descriptions */
	VkVertexInputBindingDescription binding_desc[2] = {};
	binding_desc[0].binding = 0;
	binding_desc[0].stride = sizeof(float) * 4;
	binding_desc[0].inputRate = VK_VERTEX_INPUT_RATE_VERTEX;
	binding_desc[1].binding = 1;
	binding_desc[1].stride = sizeof(VkDeviceAddress);
	binding_desc[1].inputRate = VK_VERTEX_INPUT_RATE_INSTANCE;

	/* Vertex input attribute descriptions */
	VkVertexInputAttributeDescription attributes[2] = {};
	attributes[0].location = 0;
	attributes[0].binding = binding_desc[0].binding;
	attributes[0].format = VK_FORMAT_R32G32B32A32_SFLOAT;
	attributes[0].offset = 0;
	attributes[1].location = 1;
	attributes[1].binding = binding_desc[1].binding;
	attributes[1].format = VK_FORMAT_R32G32_UINT;
	attributes[1].offset = 0;

	/* Vertex input state */
	VkPipelineVertexInputStateCreateInfo pipe_vert_state_create_info = {};
	pipe_vert_state_create_info.sType = VK_STRUCTURE_TYPE_PIPELINE_VERTEX_INPUT_STATE_CREATE_INFO;
	pipe_vert_state_create_info.pNext = nullptr;
	pipe_vert_state_create_info.vertexBindingDescriptionCount = 2;
	pipe_vert_state_create_info.pVertexBindingDescriptions = binding_desc;
	pipe_vert_state_create_info.vertexAttributeDescriptionCount = 2;
	pipe_vert_state_create_info.pVertexAttributeDescriptions = attributes;
```

spencer-lunarg:
@piofra01-arm thanks for the descriptive example. I am going to quick write up a negative test to 

1. Prevent future regressions
2. Make sure this is the actual correct fix and that there isn't something more subtle that needs to be fixed

spencer-lunarg (approved):
So looking into this, the real issue is I wasn't aware that you could use a `PhysicalStorageBuffer` struct as a shader interface type

I wrote up some test to make sure to test this correctly

Because currently we are falsely reporting any `PhysicalStorageBuffer` interface, I am ok for taking this change now until I finish up my change to correctly check these types of interfaces (Will create an issue once this is merged to track it)

Issue #5121, opened 2023-01-17
need unit test(s) for "VUID-vkUpdateDescriptorSets-None-03047


jeremyg-lunarg:
We need a unit test for this VUID, which makes sure that the descriptorset isn't in use by a command buffer when vkUpdateDescriptorSets() is called.  It should include a positive sub-test which replicates what happened in #4981, which is roughly:

1. Make 2 descriptorsets from the same pool, `a` and `b`
2. Bind `a` to a command buffer
3. submit command buffer
4. wait for idle
5. Bind `b` to a command buffer
6. submit command buffer
7. call vkUpdateDescriptorSets on `a`  - This should not cause 03047 but at one point it did.

spencer-lunarg:
Related https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5090

artem-lunarg:
The negative test is provided in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/3708

PR #5354, opened 2023-02-28
tests: Add regress for VU 03047

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5121

Please note that reverting https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4994 does not fail this test (@jeremyg-lunarg ). There are earlier checks in the code that ensure regression does not happen:
* `QueueWaitIdle` marks command buffer as not in use.
* Tried fence wait instance of `QueueWaitIdle` - also does not trigger the issue.
* When `vkBeginCommandBuffer` is called the second time (with the same command buffer), it resets the command buffer, and `CMD_BUFFER_STATE::ResetCBState` removes the bindings' command buffer reference.


jeremyg-lunarg (approved):
LGTM, thank you

PR #5355, opened 2023-02-28
layers: Move input attachment to ResourceInterfaceVariable

spencer-lunarg:
Instead of keeping a list of Input attachments per Entrypoint, this just moves it to the Interface Variable itself. This is nice because future change needs to be able to map which index is which variable, and currently this was an extra step

### DIFF
--- a/layers/core_checks/shader_validation.cpp
+++ b/layers/core_checks/shader_validation.cpp
@@ -242,44 +242,41 @@ bool CoreChecks::ValidateFsOutputsAgainstDynamicRenderingRenderPass(const SHADER
 }
 
 // Validate use of input attachments against subpass structure
-bool CoreChecks::ValidateShaderInputAttachment(const SHADER_MODULE_STATE &module_state, const Instruction &entrypoint,
-                                               const PIPELINE_STATE &pipeline) const {
+bool CoreChecks::ValidateShaderInputAttachment(const SHADER_MODULE_STATE &module_state, const PIPELINE_STATE &pipeline,
+                                               const ResourceInterfaceVariable &variable) const {
     bool skip = false;
+    assert(variable.is_input_attachment);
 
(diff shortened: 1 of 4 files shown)


layers/core_checks/shader_validation.cpp
```diff
@@ -242,44 +242,41 @@ bool CoreChecks::ValidateFsOutputsAgainstDynamicRenderingRenderPass(const SHADER
...
 // Validate use of input attachments against subpass structure
-bool CoreChecks::ValidateShaderInputAttachment(const SHADER_MODULE_STATE &module_state, const Instruction &entrypoint,
-                                               const PIPELINE_STATE &pipeline) const {
+bool CoreChecks::ValidateShaderInputAttachment(const SHADER_MODULE_STATE &module_state, const PIPELINE_STATE &pipeline,
+                                               const ResourceInterfaceVariable &variable) const {
     bool skip = false;
+    assert(variable.is_input_attachment);
 
     const auto &rp_state = pipeline.RenderPassState();
     if (rp_state && !rp_state->UsesDynamicRendering()) {
```
juan-lunarg, line 251: Avoid nested logic.

```suggestion
    if (!rp_state) {
        return skip;
    }
    if (rp_state->UsesDynamicRendering()) {
        return skip;
    }
```

juan-lunarg (approved):
LGTM aside from nitpick

artem-lunarg (approved):
LGTM! Nice simplification.

layers/core_checks/shader_validation.cpp
spencer-lunarg: `if (rp_state && !rp_state->UsesDynamicRendering()` was there from before and I also there are 2 other spots that do this, was going to combine their logic together in the next PR

PR #5357, opened 2023-02-28
ci: Move LunarG CI blacklist.json to VVL repo

ncesario-lunarg:


.github/CODEOWNERS
```diff
@@ -15,3 +15,6 @@
 # ~~~
 
 * @KhronosGroup/VVL-CODEOWNERS
+
+# @lunarpapillo must be notified prior to any blacklist.json changes being merged
+blacklist.json @lunarpapillo
```
ncesario-lunarg, line 20: Looks like @lunarpapillo would need to be given write access for this to work?

juan-lunarg: Yeah `CODEOWNERS` must have write-access.

blacklist.json
```diff
@@ -0,0 +1,394 @@
+{
+    "comments": [
+        "This is an internal CI filter for LunarG"
+    ],
+    "Vulkan-ValidationLayers": {
```
lunarpapillo, line 5: Probably need to remove the whole `"Vulkan-ValidationLayers"` level here.

lunarpapillo (changes requested):
This change should also alter the **GitHub Actions** CI, right?

It will have to alter **VulkanTests** as well, but I imagine that will happen in a separate PR.

I'll be curious to know what the defined process becomes when updating drivers or adding devices.  Do we add the disabled tests directly to the blacklist, and still create an issue?

I'm not fond of having two separate blacklist mechanisms (i.e. this external blacklist file in addition to `IsPlatform()`.  If we're moving to using a blacklist file, we should move away from `IsPlatform()`, and all those entries should enter the blacklist, right?

.github/CODEOWNERS
lunarpapillo: I'm not sure I need or want such access...?  If `blacklist.json` becomes owned by VVL, what's the relationship between `blacklist.json` and `IsPlatform()`?  What's the expected process for when a new device or driver update occurs (when we historically have had to push blacklist changes without review)?

ncesario-lunarg:
> This change should also alter the **GitHub Actions** CI, right?

No, not yet. The "skip by code" mechanism still exists and I'm not changing that as part of this PR (unnecessary additional moving part). When (or if) all the code skips get converted to a file, those tests will be defined in a _separate_ file from the "CI blacklist," and then I would imagine further **VulkanTests** modifications will be necessary to pick up both lists. This PR _just_ consolidates the blacklist information into a single repo. Consolidating the blacklist _method_ will be in a future change.

> It will have to alter **VulkanTests** as well, but I imagine that will happen in a separate PR.

https://github.com/LunarG/VulkanTests/tree/local-blacklist
 
> I'll be curious to know what the defined process becomes when updating drivers or adding devices. Do we add the disabled tests directly to the blacklist, and still create an issue?

I would imagine the process would be the same as before, no? Only the change would be to this repo rather than **VulkanTests**. Are there usually CI script changes necessary when updating the blacklist?

> I'm not fond of having two separate blacklist mechanisms (i.e. this external blacklist file in addition to `IsPlatform()`. If we're moving to using a blacklist file, we should move away from `IsPlatform()`, and all those entries should enter the blacklist, right?

Agreed. See :point_up:

blacklist.json
```diff
@@ -0,0 +1,394 @@
+{
+    "comments": [
+        "This is an internal CI filter for LunarG"
```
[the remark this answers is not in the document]
ncesario-lunarg: I actually added this comment to make it clear these "filters" are primarily for inside LunarG.

.github/CODEOWNERS
ncesario-lunarg: > I'm not sure I need or want such access...?

Sure you do :)

> If blacklist.json becomes owned by VVL, what's the relationship between blacklist.json and IsPlatform()?

Same as before. i.e., the `IsPlatform` (and related) checks in code filter tests that should not run in general...anywhere. `blacklist.json` filters out tests that shouldn't run a particular CI configuration.

> What's the expected process for when a new device or driver update occurs (when we historically have had to push blacklist changes without review)?

Again, this should be the same as before, you're just pushing to this repo instead of VulkanTests.

spencer-lunarg: ideally, once this is merged here, I will be happy to remove the `IsPlatform` in favor of updating this, will be nice to do as a single commit to make sure the transfer (`IsPlatform` to `json`) works

blacklist.json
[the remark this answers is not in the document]
spencer-lunarg: I still think having a `tests/.blacklist_lunarg.json` would also help

blacklist.json
```diff
@@ -0,0 +1,394 @@
+{
```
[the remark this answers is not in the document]
juan-lunarg: ?

blacklist.json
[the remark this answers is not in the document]
juan-lunarg: I'd just remove the comment fwiw, since the plan is to eventually have OS contributors be able to see the results of LunarG CI. And I can't come up with a good comment that isn't confusing or would become dated.

.github/CODEOWNERS
juan-lunarg: @lunarpapillo the main idea behind making you a codeowner of this file is so you are aware of changes to the `blacklist.json` and that those changes also `REQUIRE` your approval.

ncesario-lunarg:
Closing this for now, though will hopefully revisit later.

PR #5358, opened 2023-03-01
tests: Fix memory leak in Free callback

charles-lunarg:


spencer-lunarg (approved):
looking at the history of this

https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/810/files#diff-bbf41d0c011e79c0608f6b87085bc5fc4fe8cf949974ac7a478d2ccb8348a730R209

This seems correct as there wasn't a good reason to just leak it from what I can tell

juan-lunarg (approved):

jeremyg-lunarg:
> looking at the history of this
> 
> https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/810/files#diff-bbf41d0c011e79c0608f6b87085bc5fc4fe8cf949974ac7a478d2ccb8348a730R209
> 
> This seems correct as there wasn't a good reason to just leak it from what I can tell

Is the right solution to leave DummyFree() as a leaker and then REALLY free after the VerifyFound?

charles-lunarg:
I am able to reproduce the test failure locally. The crash is occurring in the NVidia driver, presumably because it is using the allocator callback to free memory that did not come from that allocator. This test failing on nvidia means that this PR is blocked until either a fix is devised or this test is blacklisted on NVidia drivers. I am investigating it to see if there is something the loader can do but I don't think thats likely, because this is 'invalid' application behavior and so shouldn't be designed around.

charles-lunarg:
Correction, while the crash does occur in the driver, if I disable calling into the driver the test still crashes due to the MSVC debug runtime checks detecting that the free call is being made from the application rather than the loader/driver.

It would be possible for the loader to stash away the VkAllocationCallbacks used to create the type and use the pAllocator passed into vkDestroyInstance if and only if they match. This seems like working around a test which causes pain no matter how you look at it.

juan-lunarg (approved):

juan-lunarg:
You may need to force push or run github cli since the actions aren't running

charles-lunarg:
Woo! Github actions finally ran. Would love to have this merged in, if anyone is available.

PR #5360, opened 2023-03-01
layers: Verify input attachment is only validated if read from

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5302

Makes sure `VUID-VkGraphicsPipelineCreateInfo-renderPass-06038` only cares if there is a `read` access to the input attachment


artem-lunarg (approved):
LGTM!

PR #5362, opened 2023-03-02
layers: Fix assumption that VK_KHR_format_feature_flags2 is always supported

aqnuep:
When support for VK_KHR_format_feature_flags2 was added to the validation layers it assumed it's always supported when vkGetPhysicalDeviceFormatProperties2[KHR] is supported.

This fixes this assumption and will make the validation layer function properly again on implementations supporting Vulkan 1.1 or GPDP2, but not VK_KHR_format_feature_flags2.


ncesario-lunarg (approved):
LGTM!

PR #5365, opened 2023-03-02
ci: Use ccache on the platforms that support it

karl-lunarg:


juan-lunarg:
Out of curiosity how does the performance compare?

.github/workflows/vvl.yml
```diff
@@ -53,33 +53,21 @@ jobs:
...
-            external/${{ matrix.config }}/Vulkan-Headers/build/install
-            external/${{ matrix.config }}/SPIRV-Headers/build/install
-            external/${{ matrix.config }}/SPIRV-Tools/build/install
-            external/${{ matrix.config }}/robin-hood-hashing/build/install
-            external/${{ matrix.config }}/googletest/build/install
-            external/${{ matrix.config }}/Vulkan-Headers/registry
-            external/${{ matrix.config }}/SPIRV-Headers/include
-          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ matrix.cc }}-${{ matrix.cxx }}-${{ matrix.config }}-${{ hashfiles('scripts/known_good.json') }}
+      # Turn on ccache this way, which is easier than figuring out how to turn it on for all deps
+      # PATH changes need to be done in an action before any action that needs it
```
juan-lunarg, line 68: FWIW we could also set the environment variables CMAKE_CXX_COMPILER_LAUNCHER and CMAKE_C_COMPILER_LAUNCHER. But I agree this method is simpler.

karl-lunarg:
You can look at:

Last build before this change:  https://github.com/karl-lunarg/Vulkan-ValidationLayers/actions/runs/4316433334

2nd build after this change (caches available): https://github.com/karl-lunarg/Vulkan-ValidationLayers/actions/runs/4317179193

linux-gcc-g++-release went from 25:57 mins to 8:35, for example.   The debug variant isn't caching very well.  There must be something in one of the components that is "compiling in" some sort of date timestamp or something.  I didn't chase it.  Linux cpp20 improved by about 3x as well.

Overall duration improved by about 9 minutes (37 down to 28), but that will vary depending on how long the long pole needs to run and job scheduling.

.github/workflows/vvl.yml
juan-lunarg: For those who are curious: https://cmake.org/cmake/help/latest/envvar/CMAKE_LANG_COMPILER_LAUNCHER.html

karl-lunarg: That's a good point.  I tried setting those as CMake parameters and realized that it wouldn't work for our dependencies unless we propagated it down through update_deps, etc.  I didn't think of the env vars.

juan-lunarg:
Considering the perf benefits maybe we should have both Mac jobs use Ninja. And have the Jenkins CI test XCode instead.

.github/workflows/vvl.yml
juan-lunarg: It could be cleaner since CMake will search for ccache if it's not provided an absolute path

```
# EX
export CMAKE_CXX_COMPILER_LAUNCHER="ccache"
export CMAKE_C_COMPILER_LAUNCHER="ccache"
```

However, this is 2 lines of code compared to the existing 1 line of code. So either way is good to me.

karl-lunarg: Yes, either way.  The PATH approach is pretty widely-accepted.  And maybe there's a tiny chance that using PATH would help cache anything that doesn't get compiled by a CMake-generated build.  I know that's unlikely with our current dependencies.  I'll leave it as is for now.

karl-lunarg:
> Considering the perf benefits maybe we should have both Mac jobs use Ninja. And have the Jenkins CI test XCode instead.

It might be good to check in on the build times occasionally to see if the macos-12 times consistently are too long.  I don't have enough data points and times do vary.  Today's results are not that bad.  Besides, you'd want to have Jenkins doing it before switching it here.

juan-lunarg:
> Edit: And there is a way to do ccache with XCode, but I think it involves adding a "ccache toolchain" to the XCode project and then selecting it at build time.

Yeah it's not very elegant with the XCode generator. I've seen what is required and it really does feel a bit hacky.

Regardless thanks for improving our CI time and simplifying our workflow :+1:

PR #5369, opened 2023-03-04
chassis: Cleanup ValidationObject creation

charles-lunarg:
This commit makes it so that only validation objects that will be used will be created. This prevents the need of manually removing unused objects.

In addition, cleanup code was added when an error is detected, thus preventing the leaking of ValidationObjects, debug_report_data, and copied pNext chains.

Lastly, the various 'init' functions of the ValidationObject have been deleted by having their contents moved into the vkCreateInstance/vkCreateDevice functions instead of having a separate function.

### DIFF
--- a/layers/generated/chassis.cpp
+++ b/layers/generated/chassis.cpp
@@ -278,41 +278,68 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateInstance(const VkInstanceCreateInfo *pCreat
     std::vector<ValidationObject*> local_object_dispatch;
 
     // Add VOs to dispatch vector. Order here will be the validation dispatch order!
-    auto thread_checker_obj = new ThreadSafety(nullptr);
-    thread_checker_obj->RegisterValidationObject(!local_disables[thread_safety], api_version, report_data, local_object_dispatch);
+    if (!local_disables[thread_safety]) {
+        local_object_dispatch.emplace_back(new ThreadSafety(nullptr));
+    }
 
-    auto parameter_validation_obj = new StatelessValidation;
-    parameter_validation_obj->RegisterValidationObject(!local_disables[stateless_checks], api_version, report_data, local_object_dispatch);
+    if (!local_disables[stateless_checks]) {
+        local_object_dispatch.emplace_back(new StatelessValidation);
+    }
 
-    auto object_tracker_obj = new ObjectLifetimes;
-    object_tracker_obj->RegisterValidationObject(!local_disables[object_tracking], api_version, report_data, local_object_dispatch);
+    if (!local_disables[object_tracking]) {
+        local_object_dispatch.emplace_back(new ObjectLifetimes);
+    }
 
-    auto core_checks_obj = new CoreChecks;
-    core_checks_obj->RegisterValidationObject(!local_disables[core_checks], api_version, report_data, local_object_dispatch);
+    if (!local_disables[core_checks]) {
+        local_object_dispatch.emplace_back(new CoreChecks);
+    }
 
-    auto best_practices_obj = new BestPractices;
-    best_practices_obj->RegisterValidationObject(local_enables[best_practices], api_version, report_data, local_object_dispatch);
+    if (local_enables[best_practices]) {
+        local_object_dispatch.emplace_back(new BestPractices);
+    }
 
-    auto gpu_assisted_obj = new GpuAssisted;
-    gpu_assisted_obj->RegisterValidationObject(local_enables[gpu_validation], api_version, report_data, local_object_dispatch);
+    if (local_enables[gpu_validation]) {
+        local_object_dispatch.emplace_back(new GpuAssisted);
+    }
 
-    auto debug_printf_obj = new DebugPrintf;
-    debug_printf_obj->RegisterValidationObject(local_enables[debug_printf], api_version, report_data, local_object_dispatch);
+    if (local_enables[debug_printf]) {
+        local_object_dispatch.emplace_back(new DebugPrintf);
+    }
 
-    auto sync_validation_obj = new SyncValidator;
-    sync_validation_obj->RegisterValidationObject(local_enables[sync_validation], api_version, report_data, local_object_dispatch);
+    if (local_enables[sync_validation]) {
+        local_object_dispatch.emplace_back(new SyncValidator);
+    }
 
     // If handle wrapping is disabled via the ValidationFeatures extension, override build flag
     if (local_disables[handle_wrapping]) {
         wrap_handles = false;
     }
 
+    // Initialize the validation objects
+    for (auto* intercept : local_object_dispatch) {
+        intercept->api_version = api_version;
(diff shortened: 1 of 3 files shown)


layers/generated/chassis.cpp
```diff
@@ -470,41 +488,53 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateDevice(VkPhysicalDevice gpu, const VkDevice
     auto disables = instance_interceptor->disabled;
     auto enables = instance_interceptor->enabled;
 
-    auto thread_safety_obj = new ThreadSafety(reinterpret_cast<ThreadSafety *>(instance_interceptor->GetValidationObject(instance_interceptor->object_dispatch, LayerObjectTypeThreading)));
-    thread_safety_obj->InitDeviceValidationObject(!disables[thread_safety], instance_interceptor, device_interceptor);
+    if (!disables[thread_safety]) {
+        device_interceptor->object_dispatch.emplace_back(new ThreadSafety(reinterpret_cast<ThreadSafety *>(
```
artem-lunarg, line 492: Because this code is already touched, the suggestion is to use `static_cast` for base->derived cast, since it's of the use cases that `static_cast` handles.

scripts/layer_chassis_generator.py
```diff
@@ -821,50 +786,81 @@ class ValidationObject {
...
         auto lock = intercept->WriteLock();
         intercept->PreCallRecordCreateInstance(pCreateInfo, pAllocator, pInstance);
     }
 
     VkResult result = fpCreateInstance(pCreateInfo, pAllocator, pInstance);
-    if (result != VK_SUCCESS) return result;
-
+    if (result != VK_SUCCESS) {
+        // Destroy all of the created validation objects if there is an error
+        DeactivateInstanceDebugCallbacks(report_data);
```
arno-lunarg, line 856: Since this is the same destruction logic as the one you previously added, maybe factorise it in a lambda?

artem-lunarg (approved):
LGTM. It's up to the author whether to apply the suggestion or not.

layers/generated/chassis.cpp
```diff
@@ -342,32 +373,19 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateInstance(const VkInstanceCreateInfo *pCreat
...
-    object_tracker_obj->FinalizeInstanceValidationObject(framework, *pInstance);
-    parameter_validation_obj->FinalizeInstanceValidationObject(framework, *pInstance);
-    core_checks_obj->FinalizeInstanceValidationObject(framework, *pInstance);
-    best_practices_obj->FinalizeInstanceValidationObject(framework, *pInstance);
-    gpu_assisted_obj->FinalizeInstanceValidationObject(framework, *pInstance);
-    debug_printf_obj->FinalizeInstanceValidationObject(framework, *pInstance);
-    sync_validation_obj->FinalizeInstanceValidationObject(framework, *pInstance);
+    for (auto* intercept : framework->object_dispatch) {
+        intercept->instance_dispatch_table = framework->instance_dispatch_table;
+        intercept->enabled = framework->enabled;
```
arno-lunarg, line 378: Question: why do we have both `enabled` and `disabled` attributes?

artem-lunarg: +1 to @arno-lunarg's question

mark-lunarg: It was originally because some options defaulted to enabled and some to disabled. In order to not require changing options for the default case we ended up with a mix of options with off or on defaults.  Many of the options pre-existed the settings and we did consider using exclusively disables or enables, but chose to avoid confusing things like 'Enable_disabling_shader_validation'.  There was a long-standing plan to have a fine-grained settings scheme where there would have been scores of options and flags which ontrolled almost every kind of validation check, but this would have required publishing these flags in the spec which is problematic for a number of reasons (and it turned out to be mostly unnecessary since we were able to get @jeremyg-lunarg who makes things fast without turning them off!).  It's a whole, long, sad story really.

jeremyg-lunarg (approved):

layers/generated/chassis.cpp
artem-lunarg: I like current solution better than `Enable_disabling_shader_validation`!  Still having two arrays with opposite meanings (enables/disables) it's some additional mental load. @mark-lunarg, as you explain there were reasons related to compatibility (and probably not only them). But with a purpose just to write the idea somewhere (probably quite naive, because I missed the discussions related to this topic, but anyway), here's the design I find it's easier to reason about. On the lowest API level everything is disabled by default and there are only `enables`. Standard validation by default initializes `enables` to whatever defaults we have now, but the user can specify custom `enables` that overwrite standard defaults.

mark-lunarg: Yeah the current solution is far from ideal, but once something is in the spec it is forever (which is why it's best to keep individual settings out of the spec IMO).  Some additional context is that there are many things which must be enabled by default but that it might still be useful to disable. If we could do it all over again I think we'd end up with something much more sensible like what you're suggesting.

scripts/layer_chassis_generator.py
charles-lunarg: I'm not against that persay, but I find it to be a bandaid to a deeper problem - that is use of manually managed memory in the ValidationObject class. Because the 'framework' is also the base class, I can't just swap out all the pointers with unique_ptr and call it a day. I could modify the destructor to check if this is a derived class or the base ValidationObject and only then destroy all the memory. This would still require a manual cleanup call inside vkCreateInstance and vkDestroyInstance, because there is no 'stack' in which to key the lifetime off of, but that is something I can live with.

One reason I didn't go with the lambda approach is that there is actually a third place where cleanup needs to happen that isn't in this PR (it was in the original but that was dropped). That third place requires more code because more stuff had been created by that point, which would require a different lambda.

layers/generated/chassis.cpp
charles-lunarg: That's a good suggestions. I was blindly copying the code over because I wasn't sure why it needed what it did, but static_cast sure is nicer to look at. 

I honestly thought about making a PR that replaced `GetValidationObject` with a type-correct version so the caller didn't have to cast, but that was a bit of a 'is that really needed?', unlike these memory leak cleanups.

ncesario-lunarg (approved):

layers/generated/chassis.cpp
```diff
@@ -278,50 +278,79 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateInstance(const VkInstanceCreateInfo *pCreat
...
     }
 
+    // Initialize the validation objects
+    for (auto* intercept : local_object_dispatch) {
+        intercept->api_version = api_version;
+        intercept->report_data = report_data;
+    }
+
+    // Define logic to cleanup everything in case of an error
+    auto cleanup_allocations = [report_data, &local_object_dispatch](){
```
juan-lunarg, line 325: :100:

PR #5370, opened 2023-03-04
tests: Cleanup swapchain & surface creation

charles-lunarg:
Most tests only need 1 swapchain & 1 surface, however a few tests create multiple surfaces and swapchains. This commit cleans up the code which handles that, separating the logic to initialize the swapchain, surface, and related variables that live inside of the VkRenderFramework from the actual logic necessary to do so. The result is a host of 'Create' functions that take variables by reference and initialize them with the appropriate calls. Then the 'Init' functions are used by tests to call the corresponding 'Create' call with the VkRenderFramework's variables as the destination of the functions.

This allows test which want to create additional swapchains and surface the ability to do so without overwriting the VkRenderFramework variables and later on clean up the created objects.

There are two new structures added.
* SurfaceContext: Holds the WSI specific types associated with a surface, such as a window or display. This struct allows tests to create a seperate 'context' which is cleaned up individually from the VkRenderFramework's context
* SurfaceInformation: Holds the capabilities, present modes, and other info about a given surface. This is to allow tests to query info about a surface that is distinct from the VkRenderFramework's surface.

### DIFF
--- a/tests/vklayertests_wsi.cpp
+++ b/tests/vklayertests_wsi.cpp
@@ -1740,17 +1740,17 @@ TEST_F(VkLayerTest, LeakASwapchain) {
     }
     ASSERT_NO_FATAL_FAILURE(InitState());
 
-    ASSERT_TRUE(InitSwapchain());
-
-    VkSurfaceKHR surface = VK_NULL_HANDLE;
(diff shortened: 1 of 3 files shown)


spencer-lunarg (approved):

jeremyg-lunarg (approved):
Yay! This is a great cleanup.

PR #5373, opened 2023-03-05
tests: Clean up instance extension names

spencer-lunarg:
`std::vector<const char *> &m_instance_extension_names = instance_extensions_;  // compatibility alias name` seems like a useless thing to keep around

there are more instances of `m_instance_extension_names` so decided to keep it (also it matches `m_device_extension_names`)

Happy to rename the member variable to be consistent with each other later, but having 2 variables alias make it hard to search for anything

(also `InstanceExtensionEnabled` is not used so removed)

### DIFF
--- a/tests/vkrenderframework.cpp
+++ b/tests/vkrenderframework.cpp
@@ -197,21 +197,6 @@ bool VkRenderFramework::DeviceExtensionSupported(const char *extension_name, con
     return std::any_of(extensions.begin(), extensions.end(), IsTheQueriedExtension);
 }
 
-// Return true if device is created and extension name is found in the list
-bool VkRenderFramework::DeviceExtensionEnabled(const char *ext_name) {
-    if (NULL == m_device) return false;
-
-    bool ext_found = false;
-    for (auto ext : m_device_extension_names) {
-        if (!strncmp(ext, ext_name, VK_MAX_EXTENSION_NAME_SIZE)) {
-            ext_found = true;
-            break;
-        }
-    }
-    return ext_found;
-}
-
-
 VkInstanceCreateInfo VkRenderFramework::GetInstanceCreateInfo() const {
 #ifdef VK_USE_PLATFORM_METAL_EXT
     return {
--- a/tests/vkrenderframework.cpp
+++ b/tests/vkrenderframework.cpp
@@ -171,13 +171,6 @@ bool VkRenderFramework::InstanceExtensionSupported(const char *const extension_n
     return std::any_of(available_extensions_.begin(), available_extensions_.end(), IsTheQueriedExtension);
 }
 
-// Return true if instance exists and extension name is in the list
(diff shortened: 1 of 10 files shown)


arno-lunarg (approved):
LGTM! I strongly agree with your sentiment, this is a weird thing ^^

tests/vkrenderframework.cpp
```diff
@@ -171,13 +171,6 @@ bool VkRenderFramework::InstanceExtensionSupported(const char *const extension_n
     return std::any_of(available_extensions_.begin(), available_extensions_.end(), IsTheQueriedExtension);
 }
 
-// Return true if instance exists and extension name is in the list
-bool VkRenderFramework::InstanceExtensionEnabled(const char *ext_name) {
```
artem-lunarg, line 175: Can we imaging scenarios when this method is needed? When unused function is very specialized it's more clear it's not needed. On the other it can always be re-added.

artem-lunarg (approved):
LGTM

tests/vkrenderframework.cpp
arno-lunarg: If I remember correctly, this:
```cpp
AddRequiredExtensions(VK_FOO_INSTANCE_EXT_NAME);
AddRequiredExtensions(VK_BAR_DEVICE_EXT_NAME);
ASSERT_NO_FATAL_FAILURE(InitFramework());
}
```
is the idiomatic way of adding and checking for extension support, handling both instance and device extensions. Last time I tried to use `InstanceExtensionEnabled`, I was told to use the idiomatic way instead, and do the change in my reference test.

Also in `VkRenderFramework::AreRequiredExtensionsEnabled` we use `VkRenderFramework::CanEnableInstanceExtension`, and this method does the same job as `InstanceExtensionEnabled`, so one more reason to remove it :p

spencer-lunarg: So I find the `InstanceExtensionSupported` helpful as it shows what is supported

```cpp
if (!InstanceExtensionSupported(VK_KHR_DEVICE_GROUP_CREATION_EXTENSION_NAME)) {
    GTEST_SKIP() << "Did not find the required instance extensions";
}
m_instance_extension_names.push_back(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
ASSERT_NO_FATAL_FAILURE(InitFramework());
```

but we spent a lot of time moving to the `AreRequiredExtensionsEnabled` instead of these manual `is extension enabled for each extension` system

artem-lunarg: We also have `DeviceExtensionEnabled` method. I wonder if this means we should remove it too (not in this PR because it is still in use), or there are different considerations for it.

spencer-lunarg: actually good point! I just made a second commit in this PR to remove `DeviceExtensionEnabled` as we have `IsExtensionsEnabled` now (aka, don't need to look up if instance or device extension)

Issue #4075, opened 2022-05-03
Validation for VUID-VkGraphicsPipelineCreateInfo-subpass-00758


Rua:
**Valid Usage IDs requested**
VUID-VkGraphicsPipelineCreateInfo-subpass-00758

PR #5377, opened 2023-03-06
Add VUID 00758

arno-lunarg:
Add VUID-VkGraphicsPipelineCreateInfo-subpass-00758 after request in https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4075

## Some context

VUID-VkGraphicsPipelineCreateInfo-subpass-00758
> If the pipeline is being created with fragment output interface state, and subpass does not use any color and/or depth/stencil attachments, then the rasterizationSamples member of pMultisampleState must follow the rules for a zero-attachment subpass

zero-attachment subpass definition:
> It is legal for a subpass to use no color or depth/stencil attachments, either because it has no attachment references or because all of them are VK_ATTACHMENT_UNUSED. This kind of subpass can use shader side effects such as image stores and atomics to produce an output. In this case, the subpass continues to use the width, height, and layers of the framebuffer to define the dimensions of the rendering area, and the rasterizationSamples from each pipeline’s VkPipelineMultisampleStateCreateInfo to define the number of samples used in rasterization; however, **if VkPhysicalDeviceFeatures::variableMultisampleRate is VK_FALSE, then all pipelines to be bound with the subpass must have the same value for VkPipelineMultisampleStateCreateInfo::rasterizationSamples**.

VkPhysicalDeviceFeatures::variableMultisampleRate definition:
> VkPhysicalDeviceFeatures::variableMultisampleRate specifies whether all pipelines that will be bound to a command buffer during a subpass which uses no attachments must have the same value for VkPipelineMultisampleStateCreateInfo::rasterizationSamples. If set to VK_TRUE, the implementation supports variable multisample rates in a subpass which uses no attachments. **If set to VK_FALSE, then all pipelines bound in such a subpass must have the same multisample rate**. This has no effect in situations where a subpass uses any attachments.

## Some questions

1) In the error message, I mention a subpass but not its corresponding render pass. I mention the render pass in the `LogObjectList` though. Is it ok?
2) For the test, I just want to create a default render pass, only thing is that I want every attachment reference in the subpass to be UNUSED, and then I want to create 2 pipelines with different samples count.
So I think the test is needlessly verbose right now. I do not know `OneShotTest` much, could it help simplify things?

### DIFF
--- a/layers/state_tracker/render_pass_state.cpp
+++ b/layers/state_tracker/render_pass_state.cpp
@@ -20,6 +20,7 @@
 #include "state_tracker/render_pass_state.h"
 #include "convert_to_renderpass2.h"
 #include "state_tracker/image_state.h"
+#include "vk_layer_utils.h"
 
 static const VkImageLayout kInvalidLayout = VK_IMAGE_LAYOUT_MAX_ENUM;
 
@@ -325,6 +326,26 @@ uint32_t RENDER_PASS_STATE::GetViewMaskBits(uint32_t subpass) const {
         }
     }
     return 0;
+}
+
+void RENDER_PASS_STATE::SetSubpassRasterizationSamples(uint32_t subpass, VkSampleCountFlagBits sample_count) {
+    WriteLockGuard lock(subpass_rasterization_samples_mutex_);
+
+    if (subpass_rasterization_samples_.size() < (subpass + 1)) {
+        subpass_rasterization_samples_.resize(subpass + 1);
+    }
+    if (!subpass_rasterization_samples_[subpass]) {
+        subpass_rasterization_samples_[subpass] = sample_count;
+    }
+}
+
+std::optional<VkSampleCountFlagBits> RENDER_PASS_STATE::GetSubpassRasterizationSamples(uint32_t subpass) const {
+    ReadLockGuard lock(subpass_rasterization_samples_mutex_);
+
+    if (subpass >= subpass_rasterization_samples_.size()) {
+        return std::nullopt;
+    }
+    return subpass_rasterization_samples_[subpass];
 }
 
 RENDER_PASS_STATE::RENDER_PASS_STATE(VkRenderingInfo const *pRenderingInfo)
--- a/layers/state_tracker/render_pass_state.h
+++ b/layers/state_tracker/render_pass_state.h
@@ -17,6 +17,8 @@
  * limitations under the License.
  */
 #pragma once
+#include <mutex>
+
 #include "state_tracker/base_node.h"
 #include "vk_safe_struct.h"
 
@@ -119,6 +121,20 @@ class RENDER_PASS_STATE : public BASE_NODE {
     uint32_t GetDynamicRenderingColorAttachmentCount() const;
     uint32_t GetDynamicRenderingViewMask() const;
     uint32_t GetViewMaskBits(uint32_t subpass) const;
+
+    // If there is already a value stored for the subpass, nothing will be updated.
+    void SetSubpassRasterizationSamples(uint32_t subpass, VkSampleCountFlagBits sample_count);
+    // Returns stored rasterization samples count for given subpass if there is one, else std::nullopt
+    std::optional<VkSampleCountFlagBits> GetSubpassRasterizationSamples(uint32_t subpass) const;
+
+  private:
+    // Used to conditionally maps a subpass to the rasterization samples count that was specified in the FIRST pipeline referencing
+    // this subpass in a call to vkCreateGraphicsPipelines.
+    // It is conditional because:
+    // - mapping is NOT maintained if VkPhysicalDeviceFeatures::variableMultisampleRate is VK_FALSE
+    // - pipelines do not necessarily use VkGraphicsPipelineCreateInfo->pMultisampleState
+    std::vector<std::optional<VkSampleCountFlagBits>> subpass_rasterization_samples_{};
+    mutable std::shared_mutex subpass_rasterization_samples_mutex_;
 };
 
 class FRAMEBUFFER_STATE : public BASE_NODE {
--- a/layers/state_tracker/state_tracker.cpp
+++ b/layers/state_tracker/state_tracker.cpp
@@ -2383,8 +2383,34 @@ bool ValidationStateTracker::PreCallValidateCreateGraphicsPipelines(VkDevice dev
         auto layout_state = Get<PIPELINE_LAYOUT_STATE>(create_info.layout);
         std::shared_ptr<const RENDER_PASS_STATE> render_pass;
 
-        if (pCreateInfos[i].renderPass != VK_NULL_HANDLE) {
+        if (create_info.renderPass != VK_NULL_HANDLE) {
             render_pass = Get<RENDER_PASS_STATE>(create_info.renderPass);
+            if (enabled_features.core.variableMultisampleRate == VK_FALSE && create_info.pMultisampleState) {
+                // if render pass uses no attachment, verify that all pipelines referencing the same subpass have the same
+                // pMultisampleState->rasterizationSamples
+                if (!render_pass->UsesColorAttachment(create_info.subpass) &&
+                    !render_pass->UsesDepthStencilAttachment(create_info.subpass)) {
+                    // If execution ends up here, subpass_rasterization_samples can still be empty if this is the first pipeline
+                    // with a pMultisampleState referencing create_info.subpass.
+                    // Rasterization samples count for create_info.subpass will be updated in PostCallRecordCreateGraphicsPipelines,
+                    // iff subpass_rasterization_samples is empty.
+                    if (std::optional<VkSampleCountFlagBits> subpass_rasterization_samples =
+                            render_pass->GetSubpassRasterizationSamples(create_info.subpass);
+                        subpass_rasterization_samples &&
+                        *subpass_rasterization_samples != create_info.pMultisampleState->rasterizationSamples) {
+                        const LogObjectList objlist(device, create_info.renderPass);
+                        skip |= LogError(
+                            objlist, "VUID-VkGraphicsPipelineCreateInfo-subpass-00758",
+                            "vkCreateGraphicsPipelines(): VkPhysicalDeviceFeatures::variableMultisampleRate is VK_FALSE and "
+                            "pCreateInfos[%" PRIu32 "] references subpass %" PRIu32
+                            " with pMultisampleState->rasterizationSamples equal to %s, while a previously created pipeline used "
+                            "pMultisampleState->rasterizationSamples equal to %s.",
+                            i, create_info.subpass,
+                            string_VkSampleCountFlagBits(create_info.pMultisampleState->rasterizationSamples),
+                            string_VkSampleCountFlagBits(*subpass_rasterization_samples));
+                    }
+                }
+            }
         } else if (enabled_features.core13.dynamicRendering) {
             auto dynamic_rendering = LvlFindInChain<VkPipelineRenderingCreateInfo>(create_info.pNext);
             render_pass = std::make_shared<RENDER_PASS_STATE>(dynamic_rendering);
@@ -2412,6 +2438,18 @@ void ValidationStateTracker::PostCallRecordCreateGraphicsPipelines(VkDevice devi
         if (pPipelines[i] != VK_NULL_HANDLE) {
             (cgpl_state->pipe_state)[i]->SetHandle(pPipelines[i]);
             Add(std::move((cgpl_state->pipe_state)[i]));
+        }
+
+        if (enabled_features.core.variableMultisampleRate == VK_FALSE) {
+            const auto &create_info = pCreateInfos[i];
+            if (create_info.renderPass != VK_NULL_HANDLE && create_info.pMultisampleState) {
+                if (auto render_pass = Get<RENDER_PASS_STATE>(create_info.renderPass);
+                    !render_pass->UsesColorAttachment(create_info.subpass) &&
+                    !render_pass->UsesDepthStencilAttachment(create_info.subpass)) {
+                    render_pass->SetSubpassRasterizationSamples(create_info.subpass,
+                                                                create_info.pMultisampleState->rasterizationSamples);
+                }
+            }
         }
     }
     cgpl_state->pipe_state.clear();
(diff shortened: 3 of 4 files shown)


layers/best_practices/best_practices_validation.h
```diff
@@ -831,7 +831,7 @@ class BestPractices : public ValidationStateTracker {
     void PostCallRecordCmdPipelineBarrier2KHR(VkCommandBuffer commandBuffer, const VkDependencyInfo* pDependencyInfo) override;
     template <typename ImageMemoryBarrier>
     void RecordCmdPipelineBarrierImageBarrier(VkCommandBuffer commandBuffer, const ImageMemoryBarrier& barrier);
-
+    using ValidationStateTracker::PreCallRecordCreateGraphicsPipelines;
```
arno-lunarg, line 834: To prevent `hides overloaded virtual function` coming from `-Woverloaded-virtual` (only clang and gcc)

artem-lunarg: Not sure if I understand how the following (line 835) `PreCallRecordCreateGraphicsPipelines` signature can override something? It has `void* cgpl_state` parameter. Do we have base classes that define virtual method with this signature?

layers/state_tracker/state_tracker.cpp
```diff
@@ -2402,6 +2424,24 @@ bool ValidationStateTracker::PreCallValidateCreateGraphicsPipelines(VkDevice dev
     return skip;
 }
 
+void ValidationStateTracker::PreCallRecordCreateGraphicsPipelines(VkDevice device, VkPipelineCache pipelineCache,
```
spencer-lunarg, line 2427: So why not `PostCallRecordCreateGraphicsPipelines`? If there is a reason we need to record this prior I would leave a comment in the code

layers/state_tracker/render_pass_state.cpp
```diff
@@ -327,6 +327,22 @@ uint32_t RENDER_PASS_STATE::GetViewMaskBits(uint32_t subpass) const {
     return 0;
 }
 
+void RENDER_PASS_STATE::TrySetSubpassRasterizationSamples(uint32_t subpass, VkSampleCountFlagBits sample_count) {
```
spencer-lunarg, line 330: I think `SetSubpassRasterizationSamples` is valid, no real reason to have `Try` in the name, especially if there is no value of know if it was set or not

layers/state_tracker/state_tracker.cpp
```diff
@@ -2383,8 +2383,30 @@ bool ValidationStateTracker::PreCallValidateCreateGraphicsPipelines(VkDevice dev
...
+            if (enabled_features.core.variableMultisampleRate == VK_FALSE && create_info.pMultisampleState) {
+                // if render pass uses no attachment, verify that all pipelines referencing the same subpass have the same
+                // pMultisampleState->rasterizationSamples
+                if (!render_pass->UsesColorAttachment(create_info.subpass) &&
+                    !render_pass->UsesDepthStencilAttachment(create_info.subpass)) {
+                    if (std::optional<VkSampleCountFlagBits> subpass_rasterization_samples =
+                            render_pass->GetSubpassRasterizationSamples(create_info.subpass);
+                        subpass_rasterization_samples &&
+                        *subpass_rasterization_samples != create_info.pMultisampleState->rasterizationSamples) {
+                        const LogObjectList objlist(device, create_info.renderPass);
```
spencer-lunarg, line 2397: I think passing the render pass object here is 100% ok!

tests/vklayertests_pipeline_shader.cpp
```diff
@@ -2884,6 +2884,66 @@ TEST_F(VkLayerTest, InvalidPipelineSamplePNext) {
...
+    VkPipelineObj pipeline_2(m_device);
+    pipeline_2.AddShader(&vs);
+    ms_state.rasterizationSamples = VK_SAMPLE_COUNT_4_BIT;
+    pipeline_2.SetMSAA(&ms_state);
+
+    pipeline_1.CreateVKPipeline(descriptorSet.GetPipelineLayout(), renderpass.handle());
+    // VkPhysicalDeviceFeatures::variableMultisampleRate is false and 2nd pipeline has not the sample samples count as 1st
+    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkGraphicsPipelineCreateInfo-subpass-00758");
+    pipeline_2.CreateVKPipeline(descriptorSet.GetPipelineLayout(), renderpass.handle());
+    m_errorMonitor->VerifyFound();
```
spencer-lunarg, line 2944: the way to use the `OneshotTest` would be

```cpp

auto ms_state_ci = LvlInitStruct<VkPipelineMultisampleStateCreateInfo>();
ms_state_ci.rasterizationSamples = VK_SAMPLE_COUNT_1_BIT;
ms_state_ci.sampleShadingEnable = 0; 
ms_state_ci.minSampleShading = 1.0;
ms_state_ci.pSampleMask = NULL;

{
    auto set_info = [&](CreatePipelineHelper &helper) {
        helper.shader_stages_ = {vs.GetStageCreateInfo(), helper.fs_->GetStageCreateInfo()};
        helper.pipe_ms_state_ci_ = ms_state_ci;
    };
    CreatePipelineHelper::OneshotTest(*this, set_feedback, kErrorBit);
}   
{
    ms_state_ci.rasterizationSamples = VK_SAMPLE_COUNT_4_BIT;
    auto set_info = [&](CreatePipelineHelper &helper) {
        helper.shader_stages_ = {vs.GetStageCreateInfo(), helper.fs_->GetStageCreateInfo()};
        helper.pipe_ms_state_ci_ = ms_state_ci;
    };
    CreatePipelineHelper::OneshotTest(*this, set_feedback, kErrorBit, "VUID-VkGraphicsPipelineCreateInfo-subpass-00758"));
}   
```

then you don't need the `VkPipelineObj`

spencer-lunarg: but honestly, this isn't that verbose and ok with it

layers/state_tracker/render_pass_state.cpp
arno-lunarg: I used `Try` because nothing will be written if there is already a value registered for the subpass. I did not return a `bool` to indicate success because right now I do not it. 
I feel that if I remove the `Try`, it will not be clear that this function does not necessarily writes a value. But maybe it is just me? 
I can change the signature to return a `bool` too, but I am still not going to read it.

juan-lunarg: Is it worth it to have this function? It's only used in 1 location. It may just be better to inline the logic.

Actually both of these helper functions are only used in 1 location.

arno-lunarg: Yes having functions is totally debatable. I would like them to not "bloat" (well you can argue it is like 3 lines of code anyway ^^") validation code with the vector size management logic when setting/getting a sample count. I think code is clearer this way?
Also, other methods like `RENDER_PASS_STATE::UsesColorAttachment` and `RENDER_PASS_STATE::UsesDepthStencilAttachment` are used in only 1 location as well.

juan-lunarg: Either way is fine to me. My main nitpick would just be that `Try` makes me think of exceptions in `C++`. Regardless FFTI.

arno-lunarg: Ok, that's the final nail in the coffin, I will remove it!

layers/state_tracker/state_tracker.cpp
```diff
@@ -2383,8 +2383,30 @@ bool ValidationStateTracker::PreCallValidateCreateGraphicsPipelines(VkDevice dev
...
-        if (pCreateInfos[i].renderPass != VK_NULL_HANDLE) {
+        if (create_info.renderPass != VK_NULL_HANDLE) {
             render_pass = Get<RENDER_PASS_STATE>(create_info.renderPass);
+            if (enabled_features.core.variableMultisampleRate == VK_FALSE && create_info.pMultisampleState) {
+                // if render pass uses no attachment, verify that all pipelines referencing the same subpass have the same
+                // pMultisampleState->rasterizationSamples
+                if (!render_pass->UsesColorAttachment(create_info.subpass) &&
+                    !render_pass->UsesDepthStencilAttachment(create_info.subpass)) {
+                    if (std::optional<VkSampleCountFlagBits> subpass_rasterization_samples =
+                            render_pass->GetSubpassRasterizationSamples(create_info.subpass);
```
juan-lunarg, line 2394: Won't this always return an empty std::optional the first time? Since `SetSubpassRasterizationSamples` is called in the `PostCallRecordCreateGraphicsPipelines` step?

tests/vklayertests_pipeline_shader.cpp
juan-lunarg, line 2944: Switching the order the pipeline is created will cause a problem:

```
    // VUID won't trigger
    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkGraphicsPipelineCreateInfo-subpass-00758");
    pipeline_2.CreateVKPipeline(descriptorSet.GetPipelineLayout(), renderpass.handle());

    pipeline_1.CreateVKPipeline(descriptorSet.GetPipelineLayout(), renderpass.handle()); // vuid will trigger
```

ncesario-lunarg (commented):
Some questions, but I think this will correctly trigger the VUID. I don't have any good alternatives, though :/

layers/state_tracker/state_tracker.cpp
```diff
@@ -2383,8 +2383,30 @@ bool ValidationStateTracker::PreCallValidateCreateGraphicsPipelines(VkDevice dev
...
+                        subpass_rasterization_samples &&
+                        *subpass_rasterization_samples != create_info.pMultisampleState->rasterizationSamples) {
```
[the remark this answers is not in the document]
arno-lunarg: I share your concerns, I too am not very happy with how I do track this state. Especially because the methods I added in RenderPass state should almost certainly not be used outside of this context. What their names *suggest* you can get/set is misleading, and the relation to pipeline creation is all but clear.

My goal is the following: At `vkCreateGraphicsPipeline` time, you provide a render pass and one of its subpass. This create info can also have to a `pMultisampleState`. If `variableMultisampleRate` is disabled, I want to make sure that all the `pMultisampleState->rasterizationSamples` specified for the same (render pass, subpass) combo specified in a pipeline create info are equal. 
So I somehow need to maintain a `rasterizationSamples` per (render pass, subpass combo). Said like this, to me RenderPass state sounds like the best spot to do this. But I end up with some highly specific please-do-not-touch-or-look code...

I added a bunch of comments to explain what I do:
- [render_pass_state.h](https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5377/commits/a13aaedfb85189ed34afa7e208d35010daf0bfa0#diff-96c48eb35d0a9486e99663386bd3814b10779e965de7eda21968a58c10db8e7bR129)
- [state_tracker.cpp](https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5377/commits/a13aaedfb85189ed34afa7e208d35010daf0bfa0#diff-0d90097147115e14fcfd561a56050886c383859f2cf15c05f11b86dfb10ea5c9R2393)
- [in the new test](https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5377/commits/0fc506b69099f827ae1aa951e17ba1655de19e85#diff-d9d7c83f4b4ab6c14eb9381ab5ee353d3d5d8e1d0449da064b7597dd5136f8e8R2941)

@juan-lunarg those comments should answer your questions, if not well they suck and I need to modify them (or the code), but TL;DR is the behaviours you pointed are intended.

layers/state_tracker/render_pass_state.h
```diff
@@ -119,6 +119,12 @@ class RENDER_PASS_STATE : public BASE_NODE {
     uint32_t GetDynamicRenderingColorAttachmentCount() const;
     uint32_t GetDynamicRenderingViewMask() const;
     uint32_t GetViewMaskBits(uint32_t subpass) const;
+
+    void SetSubpassRasterizationSamples(uint32_t subpass, VkSampleCountFlagBits sample_count);
+    std::optional<VkSampleCountFlagBits> GetSubpassRasterizationSamples(uint32_t subpass) const;
+
+  private:
+    std::vector<std::optional<VkSampleCountFlagBits>> subpass_rasterization_samples{};
```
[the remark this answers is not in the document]
jeremyg-lunarg: I agree with this concern.

layers/state_tracker/render_pass_state.cpp
```diff
@@ -327,6 +327,22 @@ uint32_t RENDER_PASS_STATE::GetViewMaskBits(uint32_t subpass) const {
     return 0;
 }
 
+void RENDER_PASS_STATE::SetSubpassRasterizationSamples(uint32_t subpass, VkSampleCountFlagBits sample_count) {
+    if (subpass_rasterization_samples.size() < (subpass + 1)) {
```
jeremyg-lunarg, line 331: If you want this to be 'close enough' to thread safe, you'd need to avoid calling resize() here and instead set it up in the constructor. As long as you're not changing the size of the vector the worst case is stale data, but resize() could reallocate memory and cause a crash.

layers/state_tracker/render_pass_state.h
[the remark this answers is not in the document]
ncesario-lunarg: I _think_ I've convinced myself this is an acceptable approach.

Is pipeline creation the only way to "bind" a pipeline to a subpass? If pipeline creation is the only place a subpass is bound to a pipeline (I think it is, but not 100% sure....e.g., is there some dynamic way to change this?), then pipeline creation seems to be the correct place to check this VUID and the RP seems like as good a place as any to store the state. I don't like it, but I don't have a good alternative.

ncesario-lunarg (commented):
I think I've convinced myself this works and is an acceptable approach, assuming my understanding of "binding a graphics pipeline to a render pass" is correct and

layers/state_tracker/render_pass_state.cpp
arno-lunarg: I just added locking to those new Get/Set methods

juan-lunarg: Would be ideal to add testing to ensure this new validation is thread safe. We have a couple positive tests that could be useful reference. And to double check with TSAN would be pragmatic.

arno-lunarg:
> Would this still be true if the Get/Set were guarded by a Read/Write lock respectively?

@ncesario-lunarg  (cannot respond inline): yes, vector is initially empty, and resized accordingly when trying to add a new element. No need to allocate upfront, I think it is better to do it just when needed.

PR #5380, opened 2023-03-06
Remove crash caused by use of VK_EXT_image_2d_view_of_3d

### LINKED ISSUE #4536: Validation incompatible with VK_EXT_image_2d_view_of_3d
The validation layer is not compatible with VK_EXT_image_2d_view_of_3d, which allows creating 2d depth-sliced views of 3D images under some circumstances. If these are accessed via descriptor then the validation layer asserts:
```
synchronization_validation.cpp:2065: bool CommandBufferAccessContext::ValidateDispatchDrawDescriptorSet(VkPipelineBindPoint, CMD_TYPE) const: Assertion `!img_view_state->IsDepthSliced()' failed.
```
The comment preceding the assert in the code says:
```
// NOTE: 2D ImageViews of VK_IMAGE_CREATE_2D_ARRAY_COMPATIBLE_BIT Images are not allowed in
// Descriptors, so we do not have to worry about depth slicing here.
// See: VUID 00343
assert(!img_view_state->IsDepthSliced());
```
so this needs updating to deal with the extension.

There are no validation messages encountered before the assert, and I believe that the code is valid vulkan.

The assert looks like it can be triggered by any code using the extension, but the CTS test `dEQP-VK.image.image_size.3d.readonly_writeonly_2d_view_7x1x1` shows the issue.

jzulauf-lunarg:
VK_EXT_image_2d_view_of_3d is not supported by syncval at this time, but this changed causes depth sliced views to be ignored inside of asserting on their presence.  Fixes #4536.

### DIFF
--- a/layers/sync/sync_validation.cpp
+++ b/layers/sync/sync_validation.cpp
@@ -2153,11 +2153,14 @@ bool CommandBufferAccessContext::ValidateDispatchDrawDescriptorSet(VkPipelineBin
                         const auto *img_view_state = image_descriptor->GetImageViewState();
                         VkImageLayout image_layout = image_descriptor->GetImageLayout();
 
+                        if (img_view_state->IsDepthSliced()) {
+                            // NOTE: 2D ImageViews of VK_IMAGE_CREATE_2D_ARRAY_COMPATIBLE_BIT Images are not allowed in
+                            // Descriptors, unless VK_EXT_image_2d_view_of_3d is supported, which it isn't at the moment.
+                            // See: VUID 00343
+                            continue;
+                        }
+
                         HazardResult hazard;
-                        // NOTE: 2D ImageViews of VK_IMAGE_CREATE_2D_ARRAY_COMPATIBLE_BIT Images are not allowed in
-                        // Descriptors, so we do not have to worry about depth slicing here.
-                        // See: VUID 00343
-                        assert(!img_view_state->IsDepthSliced());
                         const IMAGE_STATE *img_state = img_view_state->image_state.get();
                         const auto &subresource_range = img_view_state->normalized_subresource_range;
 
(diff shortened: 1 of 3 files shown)


layers/sync/sync_validation.cpp
```diff
@@ -2153,11 +2153,14 @@ bool CommandBufferAccessContext::ValidateDispatchDrawDescriptorSet(VkPipelineBin
                         const auto *img_view_state = image_descriptor->GetImageViewState();
                         VkImageLayout image_layout = image_descriptor->GetImageLayout();
 
+                        if (img_view_state->IsDepthSliced()) {
```
ncesario-lunarg, line 2156: Does whether or not `VK_EXT_image_2d_view_of_3d` is enabled matter at this point (and similarly for below)?

ncesario-lunarg (approved):
LGTM, but @juan-lunarg should probably take a look (as I think he did the original validation for this extension).

juan-lunarg:
> LGTM, but @juan-lunarg should probably take a look (as I think he did the original validation for this extension).

I did the validation for `VkImageViewSlicedCreateInfoEXT` which is similar but different.

Tony-LunarG (approved):
LGTM and fixes a spot check of previously asserting CTS tests

juan-lunarg (commented):
LGTM would be ideal to add a positive test to prevent a regression however.

layers/sync/sync_validation.cpp
jzulauf-lunarg: Not for this change.  We are replacing an assert with a continue.  When we add support for the extension this will need to be removed.  However, since a precondition for running syncval is an otherwise *clean* validation (meaning that all other validation passes), we don't tend to double check things in syncval, if it's being used, we assume it's valid to be used.  If not, the developer is ignoring best practice so YMMV applies.

PR #5383, opened 2023-03-06
sync: Call Destroy instead of Reset

charles-lunarg:
The Destroy call was calling ResetCBState() which results in Reset() being called. Reset adds in a self reference which causes a cycle in the shared_ptr's. Calling Destroy instead cleans everything up properly and doesn't leak any memory.

### DIFF
--- a/layers/sync/sync_validation.cpp
+++ b/layers/sync/sync_validation.cpp
@@ -4276,19 +4276,15 @@ syncval_state::CommandBuffer::CommandBuffer(SyncValidator *dev, VkCommandBuffer 
                                             const COMMAND_POOL_STATE *pool)
     : CMD_BUFFER_STATE(dev, cb, pCreateInfo, pool), access_context(*dev, this) {}
 
-syncval_state::CommandBuffer::~CommandBuffer() { Destroy(); }
-
 void syncval_state::CommandBuffer::Destroy() {
-    ResetCBState();  // must be first to clean up self references correctly.
+    access_context.Destroy();  // must be first to clean up self references correctly.
     CMD_BUFFER_STATE::Destroy();
 }
 
 void syncval_state::CommandBuffer::Reset() {
     CMD_BUFFER_STATE::Reset();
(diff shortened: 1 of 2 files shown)


layers/sync/sync_validation.cpp
```diff
@@ -4279,7 +4279,7 @@ syncval_state::CommandBuffer::CommandBuffer(SyncValidator *dev, VkCommandBuffer
 syncval_state::CommandBuffer::~CommandBuffer() { Destroy(); }
 
 void syncval_state::CommandBuffer::Destroy() {
-    ResetCBState();  // must be first to clean up self references correctly.
+    access_context.Destroy();  // must be first to clean up self references correctly.
```
juan-lunarg, line 4282: Looks like a regression caused by me:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/3a7ab5ca48fb9171c22f22fcb832bf4263f6f610

I'd revert the changes made to layers/synchronization_validation.* to make the fix more clear.

charles-lunarg: I can do that, though it does look like some of the changes are needed, like moving `CommandBufferAccessContext access_context;` to be public.

juan-lunarg (approved):
LGTM

FYI @jeremyg-lunarg for a double check

layers/sync/sync_validation.cpp
```diff
@@ -4276,20 +4276,16 @@ syncval_state::CommandBuffer::CommandBuffer(SyncValidator *dev, VkCommandBuffer
...
     CMD_BUFFER_STATE::Destroy();
 }
 
 void syncval_state::CommandBuffer::Reset() {
     CMD_BUFFER_STATE::Reset();
-    ResetCBState();
+    access_context.Reset();
```
jeremyg-lunarg, line 4286: 👍 for this

jeremyg-lunarg (approved):

PR #5388, opened 2023-03-07
layers: Move more to PIPELINE_STATE init

spencer-lunarg:
This change does 2 main things

1. Moves logic to member variables of `PIPELINE_STATE` that are part of the Create Info struct. These are constantly being queried and checking which sType, but we know this information at init time
2. Moves more away from using `stage_states` to indirectly get the information already known in create info (main idea is `stage_states` is re-doing a lot of work and future goal is to trim it down more)

### DIFF
--- a/layers/state_tracker/pipeline_state.h
+++ b/layers/state_tracker/pipeline_state.h
@@ -167,9 +167,16 @@ class PIPELINE_STATE : public BASE_NODE {
 
   public:
     const uint32_t create_index;  // which index in pCreateInfos, used for error messages
+
+    // Create Info values saved for fast access later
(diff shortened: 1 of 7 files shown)


layers/state_tracker/pipeline_state.h
```diff
@@ -167,9 +167,16 @@ class PIPELINE_STATE : public BASE_NODE {
 
   public:
     const uint32_t create_index;  // which index in pCreateInfos, used for error messages
+
+    // Create Info values saved for fast access later
```
juan-lunarg, line 171: Did you notice a performance gain?

juan-lunarg (approved):

PR #5389, opened 2023-03-07
tests: Make width and height uint32

artem-lunarg:
In most places the code works with uint32_t width and height. This change replaces many float->uint32 casts with only a few uint32->float casts.


spencer-lunarg (approved):
:100:

juan-lunarg (approved):

PR #5392, opened 2023-03-07
ci: Warnings as errors for MinGW build

juan-lunarg:
The main intent here is to prevent niche MinGW warnings like 08ce2e77f92c6e09a75a1d686e0e2438ed783053

And given MinGW uses GCC there isn't much work to maintain this

### DIFF
--- a/.github/workflows/vvl.yml
+++ b/.github/workflows/vvl.yml
@@ -339,9 +339,10 @@ jobs:
       - name: Set up MinGW
         uses: egor-tensin/setup-mingw@v2
       - name: Configure
-        run: cmake -S. -B build -D UPDATE_DEPS=ON -D CMAKE_BUILD_TYPE=Debug -D UPDATE_DEPS_SKIP_EXISTING_INSTALL=ON -G "Ninja"
+        run: cmake -S. -B build -D BUILD_WERROR=ON -D UPDATE_DEPS=ON -D CMAKE_BUILD_TYPE=Debug -D UPDATE_DEPS_SKIP_EXISTING_INSTALL=ON -G "Ninja"
         env:
           LDFLAGS: "-fuse-ld=lld" # MINGW linking is very slow. Use llvm linker instead.
+          CXXFLAGS: "-Wno-format" # MINGW doesn't recognize %z as a format specifier
       - name: Build
         run: cmake --build build
       - name: Install


.github/workflows/vvl.yml
```diff
@@ -339,9 +339,10 @@ jobs:
       - name: Set up MinGW
         uses: egor-tensin/setup-mingw@v2
       - name: Configure
-        run: cmake -S. -B build -D UPDATE_DEPS=ON -D CMAKE_BUILD_TYPE=Debug -D UPDATE_DEPS_SKIP_EXISTING_INSTALL=ON -G "Ninja"
+        run: cmake -S. -B build -D BUILD_WERROR=ON -D UPDATE_DEPS=ON -D CMAKE_BUILD_TYPE=Debug -D UPDATE_DEPS_SKIP_EXISTING_INSTALL=ON -G "Ninja"
         env:
           LDFLAGS: "-fuse-ld=lld" # MINGW linking is very slow. Use llvm linker instead.
+          CXXFLAGS: "-Wno-format" # MINGW doesn't recognize %z as a format specifier
```
juan-lunarg, line 345: @Biswa96 when vvl is built as a package for MinGW how does it handle %z as a format specifier?

spencer-lunarg (approved):

Biswa96:
> when vvl is built as a package for MinGW how does it handle %z as a format specifier?

Apology for late reply. Would you like to check if the following change fixes those compiler warnings?

```diff
--- a/layers/generated/chassis.h
+++ b/layers/generated/chassis.h
@@ -3958,8 +3958,10 @@ typedef enum EnableFlags {
 typedef std::array<bool, kMaxDisableFlags> CHECK_DISABLED;
 typedef std::array<bool, kMaxEnableFlags> CHECK_ENABLED;
 
-#if defined(__GNUC__) || defined(__clang__)
+#if defined(__clang__)
 #define DECORATE_PRINTF(_fmt_argnum, _first_param_num)  __attribute__((format (printf, _fmt_argnum, _first_param_num)))
+#elif defined(__GNUC__)
+#define DECORATE_PRINTF(_fmt_argnum, _first_param_num)  __attribute__((format (gnu_printf, _fmt_argnum, _first_param_num)))
 #else
 #define DECORATE_PRINTF(_fmt_num, _first_param_num)
 #endif
```

clang also defines `__GNUC__` macro. So, the order in that diff is important. I do not know how that header file was generated and did that change manually.

juan-lunarg:
I'll test this change asap to see if this fixes the problem:
https://github.com/juan-lunarg/Vulkan-ValidationLayers/actions/runs/4367450524/jobs/7638776962

juan-lunarg:
Mainly asking since this seems like a common problem users may run into when supporting mingw:
https://stackoverflow.com/questions/59147875/how-do-i-work-around-the-unknown-conversion-type-character-z-in-format-compi/75689271#75689271

Biswa96:
I can not explain the reason in depth because it is related to something with C or GNU [standards](https://xkcd.com/927/). The actual answer was suggested by @lhmouse, knows everything btw :) Then I looked into the mingw-w64 code which I am little familiar with. The logic is same as this https://github.com/mingw-w64/mingw-w64/commit/015e637b4b24d9915162ab877ed539ad0e657951

lhmouse:
The issue is not about which standard that the compiler is conforming to.

libstdc++ always defines `__USE_MINGW_ANSI_STDIO` in os_defines.h, so printf functions are always `gnu_printf`. The non-GNU `printf` format attribute denotes the legacy MSVCRT which still has no C99 support other than the `ll` specifier, and expects the MS `long double` format (same as `double`). The necessity of the dedicated check for Clang is because Clang doesn't support `gnu_printf`.

juan-lunarg:
Gotcha, that makes sense. Thanks for the explanation 👍🏾

PR #5395, opened 2023-03-08
ci: Enable Address Sanitizer

charles-lunarg:
This commit enabled Address Sanitizer in github actions CI runs. 
Currently, only Linux will have ASAN enabled with the MockICD.
ASAN requires enabling it in the loader with the option to disable unloading of dynamic libraries which enables useful stack traces.

Because leaks are unavoidable in certain circumstances, a mechanism has been setup to allow tests to disable leak detection if need be. The current tests which require this are:
* VkSyncValTest.SyncQSPresentAcquire - known leak with fix incoming
* VkLayerTest.LeakASwapchain - intentionally leaks memory to trigger test case

To disable leak checking in a test, add the following to the start of a test.

```
#if defined(VVL_ENABLE_ASAN)
    auto leak_sanitizer_disabler = __lsan::ScopedDisabler();
#endif
```

__lsan::ScopedDisabler() is a RAII type which calls the necessary leak sanitizer interface functions to prevent leaks that occur during the execution of the test from failing the test.

### DIFF
--- a/scripts/common_ci.py
+++ b/scripts/common_ci.py
@@ -125,6 +125,8 @@ def BuildLoader(args):
     make_dirs(LOADER_BUILD_DIR)
     cmake_cmd = f'cmake -C ../external/helper.cmake -DCMAKE_BUILD_TYPE={args.configuration} {args.cmake} ..'
     if IsWindows(): cmake_cmd = cmake_cmd + f' -A {args.arch}'
+    # This enables better stack traces from leak sanitizer by using the loader feature which prevents unloading of libraries at shutdown.
+    if not IsWindows(): cmake_cmd = cmake_cmd + ' -D LOADER_ENABLE_ADDRESS_SANITIZER=ON -D LOADER_DISABLE_DYNAMIC_LIBRARY_UNLOADING=ON'
     RunShellCmd(cmake_cmd, LOADER_BUILD_DIR)
 
     print("Build Loader")
--- a/tests/layer_validation_tests.h
+++ b/tests/layer_validation_tests.h
@@ -47,6 +47,14 @@
 using std::string;
 using std::vector;
 
+#if defined(VVL_ENABLE_ASAN)
+#if __has_include(<sanitizer/lsan_interface.h>)
+#include <sanitizer/lsan_interface.h>
+#else
+#error The lsan_interface.h header was not found!
+#endif
+#endif
+
 //--------------------------------------------------------------------------------------
 // Mesh and VertexFormat Data
 //--------------------------------------------------------------------------------------
--- a/tests/vklayertests_wsi.cpp
+++ b/tests/vklayertests_wsi.cpp
@@ -1728,6 +1728,11 @@ TEST_F(VkLayerTest, DisplayPresentInfoSrcRect) {
 
 TEST_F(VkLayerTest, LeakASwapchain) {
     TEST_DESCRIPTION("Leak a VkSwapchainKHR.");
+    // Because this test intentionally leaks swapchains & surfaces, we need to disable leak checking because drivers may leak memory
+    // that cannot be cleaned up from this test.
+#if defined(VVL_ENABLE_ASAN)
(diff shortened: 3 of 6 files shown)


scripts/common_ci.py
```diff
@@ -125,6 +126,7 @@ def BuildLoader(args):
     make_dirs(LOADER_BUILD_DIR)
     cmake_cmd = f'cmake -C ../external/helper.cmake -DCMAKE_BUILD_TYPE={args.configuration} {args.cmake} ..'
     if IsWindows(): cmake_cmd = cmake_cmd + f' -A {args.arch}'
+    if not IsWindows(): cmake_cmd = cmake_cmd + ' -D LOADER_ENABLE_ADDRESS_SANITIZER=ON -D LOADER_DISABLE_DYNAMIC_LIBRARY_UNLOADING=ON'
```
juan-lunarg, line 129: Add comment explaining what this accomplishes. I believe you mentioned this gives better stacktraces for users when an ASAN issue occurs.

tests/layer_validation_tests.h
```diff
@@ -47,6 +47,14 @@
 using std::string;
 using std::vector;
 
+#if defined(VVL_ENABLE_ASAN) && defined(__has_include)
```
juan-lunarg, line 50: What does `defined(__has_include)` accomplish? Our C++ minimum is 17

tests/vklayertests_wsi.cpp
```diff
@@ -1728,6 +1728,11 @@ TEST_F(VkLayerTest, DisplayPresentInfoSrcRect) {
 
 TEST_F(VkLayerTest, LeakASwapchain) {
     TEST_DESCRIPTION("Leak a VkSwapchainKHR.");
+// Because this test intentionally leaks swapchains & surfaces, we need to disable leak checking because drivers may leak memory
+// that cannot be cleaned up from this test.
```
juan-lunarg, line 1732: Minor inconsistency with `VkSyncValTest.SyncQSPresentAcquire`

```suggestion
    // Because this test intentionally leaks swapchains & surfaces, we need to disable leak checking because drivers may leak memory
    // that cannot be cleaned up from this test.
```

juan-lunarg (commented):
Overall looks good. Just some comments.

scripts/common_ci.py
```diff
@@ -87,6 +87,7 @@ def BuildVVL(args, build_tests=False):
     cmake_cmd = cmake_cmd + ' -DBUILD_WERROR=ON'
     if IsWindows(): cmake_cmd = cmake_cmd + f' -A {args.arch}'
     if build_tests: cmake_cmd = cmake_cmd + ' -DBUILD_TESTS=ON'
+    if not IsWindows(): cmake_cmd = cmake_cmd + ' -D VVL_ENABLE_ASAN=ON'
```
juan-lunarg, line 90: Instead of having this here. Modify the vvl.yml.

This will also make it easier when we add TSAN to CI.

```
      - name: Test Vulkan-ValidationLayers - Max Profile
        run: python3 scripts/github_ci_win_linux.py --test --config release --cmake='-DVVL_CPP_STANDARD=20'
        env:
          VK_KHRONOS_PROFILES_PROFILE_FILE: tests/device_profiles/max_profile.json
```

tests/layer_validation_tests.h
```diff
@@ -47,6 +47,14 @@
 using std::string;
 using std::vector;
 
+#if defined(VVL_ENABLE_ASAN)
+#if __has_include(<sanitizer/lsan_interface.h>)
+#include <sanitizer/lsan_interface.h>
+#else
+#error The lsan_interface.h header was not found!
+#endif
+#endif
```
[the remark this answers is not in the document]
charles-lunarg: This is actually wrong, because if VVL_ENABLE_ASAN is not defined, this causes compilation to fail with the #error.

juan-lunarg (approved):

spencer-lunarg (approved):

PR #5396, opened 2023-03-08
layers: Move Pipeline Derivatives to own function

spencer-lunarg:
creates a `ValidatePipelineDerivatives` function to do all pipeline derivatives logic

### DIFF
--- a/layers/core_checks/pipeline_validation.cpp
+++ b/layers/core_checks/pipeline_validation.cpp
@@ -84,11 +84,10 @@ bool CoreChecks::ValidatePipelineLibraryFlags(const VkGraphicsPipelineLibraryFla
     return skip;
 }
 
-bool CoreChecks::ValidatePipeline(std::vector<std::shared_ptr<PIPELINE_STATE>> const &pipelines, uint32_t pipe_index) const {
+bool CoreChecks::ValidatePipelineDerivatives(std::vector<std::shared_ptr<PIPELINE_STATE>> const &pipelines,
+                                             uint32_t pipe_index) const {
     bool skip = false;
-    safe_VkSubpassDescription2 *subpass_desc = nullptr;
     const auto &pipeline = *pipelines[pipe_index].get();
-
     // If create derivative bit is set, check that we've specified a base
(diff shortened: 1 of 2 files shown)


layers/core_checks/pipeline_validation.cpp
```diff
@@ -97,9 +96,7 @@ bool CoreChecks::ValidatePipeline(std::vector<std::shared_ptr<PIPELINE_STATE>> c
         const VkPipeline base_handle = pipeline.BasePipeline<VkGraphicsPipelineCreateInfo>();
         const int32_t base_index = pipeline.BasePipelineIndex<VkGraphicsPipelineCreateInfo>();
         if (!((base_handle != VK_NULL_HANDLE) ^ (base_index != -1))) {
-            // TODO: This check is a superset of VUID-VkGraphicsPipelineCreateInfo-flags-00724 and
-            // TODO: VUID-VkGraphicsPipelineCreateInfo-flags-00725
```
spencer-lunarg, line 101: https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5750 will both reduce this to the first VUID (00724) and then make it so this function can just be used across the various pipeline creations (not just graphics) with same VUID

artem-lunarg: Question about https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5750 MR.
The common section replaces the previous definitions, but what will happen with VU IDs? I can't see VUIDs in the new version.

spencer-lunarg: > I can't see VU IDs in the extracted section, how does it work? 

totally a valid question! So the MR doesn't have the new VUID, they will be generated when it gets merged and before being released into the new headers/spec version

from there, these VUID might change, if so, Mike will catch them easily on the header update MR

artem-lunarg: Ah, that's how this spec bakery works. Thanks.

artem-lunarg (approved):

PR #5397, opened 2023-03-08
layers: Move pipeline check into correct function

spencer-lunarg:
Tries to move more things out of `ValidatePipeline` and into better isolated functions

also adds a `ValidateGraphicsPipelineDepthStencilState`

### DIFF
--- a/layers/core_checks/pipeline_validation.cpp
+++ b/layers/core_checks/pipeline_validation.cpp
@@ -211,47 +211,12 @@ bool CoreChecks::ValidatePipeline(std::vector<std::shared_ptr<PIPELINE_STATE>> c
         }
     }
 
-    const auto ds_state = pipeline.DepthStencilState();
-    if (ds_state) {
-        if ((((ds_state->flags & VK_PIPELINE_DEPTH_STENCIL_STATE_CREATE_RASTERIZATION_ORDER_ATTACHMENT_DEPTH_ACCESS_BIT_ARM) !=
-              0) ||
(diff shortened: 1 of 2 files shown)


layers/core_checks/pipeline_validation.cpp
```diff
@@ -818,19 +761,6 @@ bool CoreChecks::ValidateGraphicsPipelineLibrary(const PIPELINE_STATE &pipeline)
                 }
             }
         }
-
-        const auto &rp_state = pipeline.RenderPassState();
-        if ((!rp_state || !rp_state->renderPass()) && pipeline.fragment_shader_state && !pipeline.fragment_output_state) {
-            if (!pipeline.DepthStencilState() ||
-                (pipeline.DepthStencilState()->sType != VK_STRUCTURE_TYPE_PIPELINE_DEPTH_STENCIL_STATE_CREATE_INFO)) {
```
artem-lunarg, line 825: Is the assumption that `sType` check is done somewhere else?

spencer-lunarg: > VUID-VkPipelineDepthStencilStateCreateInfo-sType-sType
sType must be VK_STRUCTURE_TYPE_PIPELINE_DEPTH_STENCIL_STATE_CREATE_INFO

So if there is ever something in `pDepthStencilState`, it will be checked in StatelessValidation::ValidatePipelineDepthStencilStateCreateInfo before this... basically we should never have to worry about 

>  does not point to a valid VkPipelineDepthStencilStateCreateInfo struct

This is in other places, but this is the only spot we were checking for sType

artem-lunarg (approved):

PR #5399, opened 2023-03-08
layers: Add LogObjectList to copy blit resolve

spencer-lunarg:
went trough `copy_blit_resolve_validation.cpp` and added `LogObjectList` wherever it was missing objects that would be valuable to print out


jeremyg-lunarg (approved):
LGTM

juan-lunarg (approved):

PR #5403, opened 2023-03-08
layers: Cleanup effective extent logic

juan-lunarg:
All the logic for calculating the effective subresource extent is now in GetEffectiveExtent.


layers/state_tracker/image_state.cpp
```diff
@@ -394,45 +394,6 @@ VkDeviceSize IMAGE_STATE::GetFakeBaseAddress() const {
...
-
-    if (createInfo.flags & VK_IMAGE_CREATE_CORNER_SAMPLED_BIT_NV) {
-        extent.width = (0 == extent.width ? 0 : std::max(2U, 1 + ((extent.width - 1) >> mip_level)));
-        extent.height = (0 == extent.height ? 0 : std::max(2U, 1 + ((extent.height - 1) >> mip_level)));
-        extent.depth = (0 == extent.depth ? 0 : std::max(2U, 1 + ((extent.depth - 1) >> mip_level)));
-    } else {
-        extent.width = (0 == extent.width ? 0 : std::max(1U, extent.width >> mip_level));
-        extent.height = (0 == extent.height ? 0 : std::max(1U, extent.height >> mip_level));
-        extent.depth = (0 == extent.depth ? 0 : std::max(1U, extent.depth >> mip_level));
-    }
```
juan-lunarg, line 421: I rewrote this logic to not require an if statement since I think it reads a bit better.

layers/vk_layer_utils.h
```diff
@@ -384,14 +384,58 @@ static inline uint32_t FullMipChainLevels(VkExtent3D extent) {
...
+    const VkFormat format = ci.format;
+    if (FormatIsMultiplane(format)) {
+        VkExtent2D divisors = FindMultiplaneExtentDivisors(format, aspect_mask);
+        extent.width /= divisors.width;
+        extent.height /= divisors.height;
+    }
+
+    // Don't allow mip adjustment to create 0 dim, but pass along a 0 if that's what subresource specified
+    {
+        const uint32_t corner = (ci.flags & VK_IMAGE_CREATE_CORNER_SAMPLED_BIT_NV);
```
spencer-lunarg, line 407: sorry,  can we have a comment here, how is a flag with the value `0x00002000` being used to set the extent?

spencer-lunarg (approved):
nice cleanup, just a comment otherwise LGTM

layers/vk_layer_utils.h
juan-lunarg: That is a bug. Thank you for pointing this out.

juan-lunarg: Corner should evaluate to 1 if we corner sample. 0 otherwise.

juan-lunarg: I added testing to ensure that corner sampled images are tested properly. Since there didn't seem to be any testing that caught the original incorrect code I had.

tests/positive/image_buffer.cpp
```diff
@@ -2798,3 +2798,75 @@ TEST_F(VkPositiveLayerTest, SlicedCreateInfo) {
         vk_testing::ImageView image_view(*m_device, ivci);
     }
 }
+
+TEST_F(VkPositiveLayerTest, GetEffectiveExtent) {
+    TEST_DESCRIPTION("Test unlikely GetEffectiveExtent edge cases");
```
juan-lunarg, line 2803: Helps test edge cases are current testing doesn't cover explicitly

Issue #5382, opened 2023-03-06
Memory leaks in Sync Val triggered by test VkSyncValTest.SyncQSPresentAcquire


charles-lunarg:
**Environment:**
 - OS: Linux
 - GPU: MockICD

**Describe the Issue**

After enabling leak-sanitizer, all SyncVal tests failed due to memory leaks caused by cyclic dependencies of shared_ptr. This is proven by the lack of 'direct' leaks. All leaks detected are indirect, and when looking at the callsite, they are all due to the creation of shared_ptr's.

Good starting test cases are `VkSyncValTest.Sync2BufferCopyHazards` which only has 5 leaks and `VkSyncValTest.TestCopyingToCompressedImage` which only has 4. I suspect many of the leaks are from the same underlying issue(s) so focusing on the tests with few leaks are likely to be more fruitful than investigating tests with 40+ leaks.

EDIT: I have removed the log file, since it is 5mb and doesn't need to be preserved for history. Turns out 99% of the leaks were from one source, so the log is not very helpful.

charles-lunarg:
All but one test case running the Mock ICD are resolved by PR #5383 

The only test case that still leaks memory with that PR is VkSyncValTest.SyncQSPresentAcquire.

charles-lunarg:
I spoke too soon.
This is not a fixed memory leak, and I feel quite silly for how I missed that it didn't fix the leak.
When I ran the changes sent to me, I forgot to remove the 'disable-leak-checking' code from the test. Of course the test didn't produce any leaks, I intentionally disabled leak checking on it!

Anyhow, I will reopen the issue to log it, but also will re-add the leak-check-disabling for the offending test so ASAN can be enabled in CI. The flip side of this is that the PR to *fix* the leak can simply remove the leak-check-disabling logic and run CI to verify that it succeeds, instead of relying on me to run it correctly.

PR #5406, opened 2023-03-09
sync: Remove circular shared_ptr reference between log and queue batch context

jzulauf-lunarg:
Record only the needed information in the log, without the reference to the queue batch context, by storing the base class information only.


charles-lunarg:
I confirm that this change fixes the leak I detected. So it fixes #5382, which is very nice.

charles-lunarg (approved):

artem-lunarg (approved):

juan-lunarg (approved):

PR #5407, opened 2023-03-09
tests: Revert change to DescriptorPoolInUseResetSignaled

juan-lunarg:
Accidentally submitted in d1a51896a8922

Not sure if this is actually needed since CI passed on d1a51896a8922


spencer-lunarg:
> Not sure if this is actually needed since CI passed on https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/d1a51896a89220e1ad646766ecda5ce578ad8b37

agree, why revert if not needed?

juan-lunarg:
> agree, why revert if not needed?

👍🏾

PR #5408, opened 2023-03-09
XCB vuids, remove broken XCB usage, add positive WSI testing

juan-lunarg:
- Adds VUIDs for [VkXcbSurfaceCreateInfoKHR](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkXcbSurfaceCreateInfoKHR.html)
- Removes broken XCB usage from vkrenderframework.h
- Add positive WSI testing


juan-lunarg:
> Removes broken XCB usage from vkrenderframework.h

Note: In this PR I'm removing XCB usage since it's apparently never been an issue since it will simplify adding wayland support to the vkrenderframework.h (since there will be less clutter)

tests/vkrenderframework.cpp
```diff
@@ -681,16 +681,6 @@ bool VkRenderFramework::CreateSurface(SurfaceContext &surface_context, VkSurface
...
-#if defined(VK_USE_PLATFORM_XCB_KHR)
-    surface_context.m_surface_xcb_conn = xcb_connect(NULL, NULL);
-    if (surface_context.m_surface_xcb_conn) {
-        xcb_window_t window = xcb_generate_id(surface_context.m_surface_xcb_conn);
-        VkXcbSurfaceCreateInfoKHR surface_create_info = LvlInitStruct<VkXcbSurfaceCreateInfoKHR>();
-        surface_create_info.connection = surface_context.m_surface_xcb_conn;
-        surface_create_info.window = window;
-        return VK_SUCCESS == vk::CreateXcbSurfaceKHR(instance(), &surface_create_info, nullptr, &surface);
-    }
-#endif
```
juan-lunarg, line 693: NOTE: This xcb_window_t is invalid (no width, no height, etc)

Instead of fixing it. My plan is currently to just properly implement Wayland into `vkrenderframework.*` which is a more valuable WSI to have in `vkrenderframework`

juan-lunarg: I spoke too soon. Looks like this is needed for CI. I'll revert this commit.

spencer-lunarg (approved):

PR #5410, opened 2023-03-10
layers: Fix if InputAttachmentIndex is non-zero

spencer-lunarg:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5366 didn't have a test with `InputAttachmentIndex` set to a non-zero value. @Tony-LunarG pointed out from a CTS test there was a bug which was because I added the offset where not needed

### DIFF
--- a/layers/state_tracker/shader_module.cpp
+++ b/layers/state_tracker/shader_module.cpp
@@ -1333,8 +1333,7 @@ ResourceInterfaceVariable::ResourceInterfaceVariable(const SHADER_MODULE_STATE& 
                     for (const Instruction* insn : image_read_loads) {
                         if (insn->Opcode() == spv::OpAccessChain) {
                             const uint32_t access_index = module_state.GetConstantValueById(insn->Word(4));
-                            const uint32_t index = access_index + decorations.input_attachment_index_start;
-                            input_attachment_index_read.at(index) = true;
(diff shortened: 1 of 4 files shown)


Tony-LunarG (approved):
LGTM and fixes CTS tests

layers/state_tracker/shader_module.cpp
```diff
@@ -1333,8 +1333,7 @@ ResourceInterfaceVariable::ResourceInterfaceVariable(const SHADER_MODULE_STATE&
                     for (const Instruction* insn : image_read_loads) {
                         if (insn->Opcode() == spv::OpAccessChain) {
                             const uint32_t access_index = module_state.GetConstantValueById(insn->Word(4));
-                            const uint32_t index = access_index + decorations.input_attachment_index_start;
-                            input_attachment_index_read.at(index) = true;
+                            input_attachment_index_read.at(access_index) = true;
```
juan-lunarg, line 1336: complete nit FFTI is there a reason to use the `at` instead of `[]` here?

`at` does add more overhead since it does bounds checking

juan-lunarg (approved):

layers/state_tracker/shader_module.cpp
spencer-lunarg: no reason, happy to use `[]`

Issue #5372, opened 2023-03-05
Not handling XML extension depend logic correctly


spencer-lunarg:
The recent XML introduced the `+` and `,` to mean both `and` and `or` for what an extension depends on

For example the XML has

```xml
<extension name="VK_KHR_buffer_device_address" depends="(VK_KHR_get_physical_device_properties2+VK_KHR_device_group),VK_VERSION_1_1">
```

but the following test shows that the `or` logic is not being observed correctly

```cpp
TEST_F(VkPositiveLayerTest, ExtensionXmlDependsLogic) {
    TEST_DESCRIPTION("Make sure the OR in 'depends' from XML is observed correctly");
    // VK_KHR_buffer_device_address requires
    // (VK_KHR_get_physical_device_properties2 AND VK_KHR_device_group) OR VK_VERSION_1_1
    // If Vulkan 1.1 is not supported, should still be valid
    SetTargetApiVersion(VK_API_VERSION_1_0);
    if (!InstanceExtensionSupported(VK_KHR_DEVICE_GROUP_CREATION_EXTENSION_NAME) ||
        !InstanceExtensionSupported(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME)) {
        GTEST_SKIP() << "Did not find the required instance extensions";
    }
    m_instance_extension_names.push_back(VK_KHR_DEVICE_GROUP_CREATION_EXTENSION_NAME);
    m_instance_extension_names.push_back(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
    ASSERT_NO_FATAL_FAILURE(InitFramework());

    if (!DeviceExtensionSupported(VK_KHR_BUFFER_DEVICE_ADDRESS_EXTENSION_NAME) ||
        !DeviceExtensionSupported(VK_KHR_DEVICE_GROUP_EXTENSION_NAME)) {
        GTEST_SKIP() << "Did not find the required device extensions";
    }

    m_device_extension_names.push_back(VK_KHR_BUFFER_DEVICE_ADDRESS_EXTENSION_NAME);
    m_device_extension_names.push_back(VK_KHR_DEVICE_GROUP_EXTENSION_NAME);
    ASSERT_NO_FATAL_FAILURE(InitState());
}
```

error:

[...]

PR #5411, opened 2023-03-10
layers: Workaround extension dependency expressions

spencer-lunarg:
(related to https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5372)

This doesn't fix the issue, but provides a temporary fix for giving any false positive while removing only a very small amount of valid checks out until a better fix is worked out

cc @ncesario-lunarg

### DIFF
--- a/tests/vklayertests_command.cpp
+++ b/tests/vklayertests_command.cpp
@@ -8017,12 +8017,15 @@ TEST_F(VkLayerTest, VerifyMaxMultiviewInstanceIndex) {
 
 TEST_F(VkLayerTest, InvalidSetFragmentShadingRateValues) {
     TEST_DESCRIPTION("Specify invalid fragment shading rate values");
-
+    SetTargetApiVersion(VK_API_VERSION_1_2);
     AddRequiredExtensions(VK_KHR_MAINTENANCE_1_EXTENSION_NAME);
     AddRequiredExtensions(VK_KHR_FRAGMENT_SHADING_RATE_EXTENSION_NAME);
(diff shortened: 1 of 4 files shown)


tests/vklayertests_command.cpp
```diff
@@ -8017,13 +8017,16 @@ TEST_F(VkLayerTest, VerifyMaxMultiviewInstanceIndex) {
 
 TEST_F(VkLayerTest, InvalidSetFragmentShadingRateValues) {
     TEST_DESCRIPTION("Specify invalid fragment shading rate values");
-
+    SetTargetApiVersion(VK_API_VERSION_1_2);
```
spencer-lunarg, line 8020: this doesn't fail now, but with a future XML change of

`<extension name="VK_KHR_fragment_shading_rate" depends="(VK_KHR_create_renderpass2,VK_VERSION_1_2)+(VK_KHR_get_physical_device_properties2,VK_VERSION_1_1)">` this will give a false positive of

> Validation Error: [ UNASSIGNED-GeneralParameterError-ExtensionNotEnabled ] Attempted to call vkCmdSetFragmentShadingRateKHR() but its required extension VK_VERSION_1_1 has not been enabled
>
> Validation Error: [ UNASSIGNED-GeneralParameterError-ExtensionNotEnabled ]  Attempted to call vkCmdSetFragmentShadingRateKHR() but its required extension VK_VERSION_1_2 has not been enabled

mikes-lunarg:
Merged in #5424

PR #5413, opened 2023-03-10
Improve error message in Image and Command Buffer validation

spencer-lunarg:
Adds the `VkFormat` and `LogOjectList` to more error messages


artem-lunarg (approved):
🤪👍

juan-lunarg (approved):

PR #5415, opened 2023-03-10
layers: Add validation for XCB surface creation

juan-lunarg:
- Add VUIDs/testing for X11 surface creation
- Fix typo in Xcb validation

### DIFF
--- a/layers/stateless/parameter_validation_utils.cpp
+++ b/layers/stateless/parameter_validation_utils.cpp
@@ -5796,8 +5796,7 @@ bool StatelessValidation::manual_PreCallValidateCreateWin32SurfaceKHR(VkInstance
     bool skip = false;
 
     if (pCreateInfo->hwnd == nullptr) {
-        skip |= LogError(device, "VUID-VkWin32SurfaceCreateInfoKHR-hwnd-01308",
-                         "vkCreateWin32SurfaceKHR(): hwnd must be a valid Win32 HWND but hwnd is NULL.");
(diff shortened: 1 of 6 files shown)


layers/stateless/parameter_validation_utils.cpp
```diff
@@ -5848,20 +5848,43 @@ bool StatelessValidation::manual_PreCallValidateCreateXcbSurfaceKHR(VkInstance i
...
+                                                                     const VkXlibSurfaceCreateInfoKHR *pCreateInfo,
+                                                                     const VkAllocationCallbacks *pAllocator,
+                                                                     VkSurfaceKHR *pSurface) const {
+    bool skip = false;
+
+    const auto display = pCreateInfo->dpy;
+    const auto window = pCreateInfo->window;
+
+    if (display == nullptr) {
+        skip |= LogError(device, "VUID-VkXlibSurfaceCreateInfoKHR-dpy-01313", "vkCreateXlibSurfaceKHR: connection is NULL!");
```
arno-lunarg, line 5877: I think `connection` should be `display` instead in the error message?

arno-lunarg (approved):
LGTM! One comment on error message

layers/stateless/parameter_validation_utils.cpp
juan-lunarg: Good catch 👍🏾

ncesario-lunarg (approved):

PR #5421, opened 2023-03-13
Add VUID 00758 (2nd PR)

arno-lunarg:
Add VUID-VkGraphicsPipelineCreateInfo-subpass-00758 after request in https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4075

## Previous PR

https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5377
In the previous PR, validation was performed at pipeline creation time, using state stored in the supplied render pass. This is incorrect, as the supplied render pass only serves as a template in this context.
Validation is now done at pipeline bind time, using a new attribute `active_subpass_sample_count_` stored in `CMD_BUFFER_STATE`

## Some context for 00758 

VUID-VkGraphicsPipelineCreateInfo-subpass-00758
> If the pipeline is being created with fragment output interface state, and subpass does not use any color and/or depth/stencil attachments, then the rasterizationSamples member of pMultisampleState must follow the rules for a zero-attachment subpass

zero-attachment subpass definition:
> It is legal for a subpass to use no color or depth/stencil attachments, either because it has no attachment references or because all of them are VK_ATTACHMENT_UNUSED. This kind of subpass can use shader side effects such as image stores and atomics to produce an output. In this case, the subpass continues to use the width, height, and layers of the framebuffer to define the dimensions of the rendering area, and the rasterizationSamples from each pipeline’s VkPipelineMultisampleStateCreateInfo to define the number of samples used in rasterization; however, **if VkPhysicalDeviceFeatures::variableMultisampleRate is VK_FALSE, then all pipelines to be bound with the subpass must have the same value for VkPipelineMultisampleStateCreateInfo::rasterizationSamples**.

VkPhysicalDeviceFeatures::variableMultisampleRate definition:
> VkPhysicalDeviceFeatures::variableMultisampleRate specifies whether all pipelines that will be bound to a command buffer during a subpass which uses no attachments must have the same value for VkPipelineMultisampleStateCreateInfo::rasterizationSamples. If set to VK_TRUE, the implementation supports variable multisample rates in a subpass which uses no attachments. **If set to VK_FALSE, then all pipelines bound in such a subpass must have the same multisample rate**. This has no effect in situations where a subpass uses any attachments.

### DIFF
--- a/layers/state_tracker/cmd_buffer_state.h
+++ b/layers/state_tracker/cmd_buffer_state.h
@@ -296,6 +296,10 @@ class CMD_BUFFER_STATE : public REFCOUNTED_NODE {
     uint32_t active_render_pass_device_mask;
     uint32_t GetActiveSubpass() const { return active_subpass_; }
     void SetActiveSubpass(uint32_t subpass);
+    std::optional<VkSampleCountFlagBits> GetActiveSubpassRasterizationSampleCount() const { return active_subpass_sample_count_; }
+    void SetActiveSubpassRasterizationSampleCount(VkSampleCountFlagBits rasterization_sample_count) {
+        active_subpass_sample_count_ = rasterization_sample_count;
+    }
     std::shared_ptr<FRAMEBUFFER_STATE> activeFramebuffer;
     // Unified data structs to track objects bound to this command buffer as well as object
     //  dependencies that have been broken : either destroyed objects, or updated descriptor sets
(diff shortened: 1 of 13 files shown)


layers/state_tracker/cmd_buffer_state.h
```diff
@@ -294,7 +294,12 @@ class CMD_BUFFER_STATE : public REFCOUNTED_NODE {
 
     VkSubpassContents activeSubpassContents;
     uint32_t active_render_pass_device_mask;
-    uint32_t activeSubpass;
+    uint32_t GetActiveSubpass() const { return active_subpass_; }
```
spencer-lunarg, line 297: nit: We don't normally use a getter for just grabbing a `public` member

spencer-lunarg (approved):

PR #5425, opened 2023-03-13
style: Use #pragma once everywhere

charles-lunarg:
The code is using #pragma once in most headers, but some still had include guards, and most perplexingly some had both.

### DIFF
--- a/layers/android_ndk_types.h
+++ b/layers/android_ndk_types.h
@@ -32,8 +32,7 @@
 //     - r20 YCbCr was added
 //     - r23 (current) no differences from r20 being used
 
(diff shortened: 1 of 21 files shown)


juan-lunarg (approved):

juan-lunarg (changes requested):
vk_extension_helper is generated

```
--- temp/vk_extension_helper.h
+++ repo/vk_extension_helper.h
@@ -22,8 +22,8 @@
  ****************************************************************************/
 
-
 #pragma once
+
 #include <string>
 #include <utility>
 #include <set>
```

juan-lunarg (approved):

PR #5433, opened 2023-03-14
layers: Fix which GPL shader stages are checked

spencer-lunarg:
There are VUs that say

> If the pipeline is being created with [pre-rasterization shader state](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#pipelines-graphics-subsets-pre-rasterization)

But currently we were checking all stages, including those being linked in

Adding the `HasSameParent` function allows us to know only to check if the pipeline was created with the state (for GPL)

### DIFF
--- a/layers/core_checks/pipeline_validation.cpp
+++ b/layers/core_checks/pipeline_validation.cpp
@@ -226,7 +226,6 @@ bool CoreChecks::ValidatePipeline(const PIPELINE_STATE &pipeline) const {
     skip |= ValidateGraphicsPipelineShaderState(pipeline);
     skip |= ValidateGraphicsPipelineBlendEnable(pipeline);
 
-    const uint32_t active_shaders = pipeline.active_shaders;
     if (pipeline.pre_raster_state || pipeline.fragment_shader_state) {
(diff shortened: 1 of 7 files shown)


layers/core_checks/pipeline_validation.cpp
```diff
@@ -1553,6 +1531,32 @@ bool CoreChecks::ValidateGraphicsPipelinePreRasterState(const PIPELINE_STATE &pi
             }
         }
     }
+
+    // pre_raster has Tessellation shaders, but vertex input has topology
+    if (pipeline.pre_raster_state && pipeline.vertex_input_state) {
+        // Either both or neither TC/TE shaders should be defined
+        const bool has_control = (pipeline.active_shaders & VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT) != 0;
+        const bool has_eval = (pipeline.active_shaders & VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT) != 0;
```
artem-lunarg, line 1539: If `active_shaders` is correct choice here, could you please clarify why?

artem-lunarg: Thanks, it's clear now.

PR #5436, opened 2023-03-14
Spencer lunarg clean2

spencer-lunarg:
Remove unused and dead values/variables/functions/etc


layers/state_tracker/cmd_buffer_state.cpp
```diff
@@ -1576,16 +1575,3 @@ void CMD_BUFFER_STATE::UnbindResources() {
     // Pipeline and descriptor sets
     lastBound[BindPoint_Graphics].Reset();
 }
-
-bool CMD_BUFFER_STATE::RasterizationDisabled() const {
```
arno-lunarg, line 1580: I feel that this function could be useful one day, and since it is not that trivial to implement (you have to think about correctly handling dynamic state), maybe it is ok to keep it around?

layers/state_tracker/image_layout_map.cpp
```diff
@@ -164,17 +164,6 @@ void ImageSubresourceLayoutMap::SetSubresourceRangeInitialLayout(const CMD_BUFFE
     }
 }
 
-// Saves an encode to fetch both in the same call
-const ImageSubresourceLayoutMap::LayoutEntry* ImageSubresourceLayoutMap::GetSubresourceLayouts(
```
arno-lunarg, line 168: Again here it feels like something worth keeping? I do not know the context enough to have a firm position

arno-lunarg (commented):
LGTM! 2 comments
I do not feel legitimate to approve this because of the changes in `sync_validation.cpp`

layers/state_tracker/cmd_buffer_state.cpp
jeremyg-lunarg: I'd be ok with leaving this around but maybe add a comment about why you're leaving it

layers/state_tracker/image_layout_map.cpp
jeremyg-lunarg: I don't think this one is worth keeping

layers/vk_mem_alloc.h
```diff
@@ -900,7 +900,7 @@ Use it as a unique identifier to virtual allocation within the single block.
 
 Use value `VK_NULL_HANDLE` to represent a null/invalid allocation.
 */
-VK_DEFINE_NON_DISPATCHABLE_HANDLE(VmaVirtualAllocation);
+VK_DEFINE_NON_DISPATCHABLE_HANDLE(VmaVirtualAllocation)
```
jeremyg-lunarg, line 903: You shouldn't change vk_mem_alloc.h because it is a 3rd party header. (really it should be in known_good.json)

jeremyg-lunarg (dismissed):
I don't think you should change vk_mem_alloc.h  the rest looks fine to me.

layers/vk_mem_alloc.h
juan-lunarg: Can't be due in `known_good.json` due to system package managers not deliver vma

layers/state_tracker/cmd_buffer_state.cpp
spencer-lunarg: sounds good

PR #5440, opened 2023-03-14
tests: Make GoogleTest assertions throw

charles-lunarg:
This change adds a TestEventListener which causes failed assertions to
throw, forcing a test to end early rather than continuing on. This is
beneficial due to making any assertions inside of constructors and
subroutines cause the test to stop execution rather than continue.

For example, if the test framework fails to create a device, subsequent
code is going to hard crash, potentially halting execution or worse
causing subsequent tests to fail in unpredictable ways. While it is
possible to halt execution with ASSERT_NO_FATAL_FAILURE in most cases,
it requires all tests to be rewritten using it, and doesn't work where
the failed assertion is inside of a constructor.

The tests on android use the android_main function to execute. Thus
we can platform-guard main to only exist on not-android. This is
helpful if there are constructs that can be used in not-android but
would cause compilation failure if they appeared anywhere in the
source, like exceptions.

### DIFF
--- a/tests/framework/layer_validation_tests.cpp
+++ b/tests/framework/layer_validation_tests.cpp
@@ -3823,6 +3823,7 @@ void android_main(struct android_app *app) {
 }
 #endif
 
+#if !defined(ANDROID) && !defined(VALIDATION_APK)
 #if defined(_WIN32) && !defined(NDEBUG)
 #include <crtdbg.h>
 #endif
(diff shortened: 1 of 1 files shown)


charles-lunarg:
Yes, [this PR](https://github.com/LunarG/VulkanTests/pull/347) these changes to work to their fullest. Without them, the program will just die whenever an assert fails. (as it'll result in an uncaught exception)

juan-lunarg:
Closing due to inactivity

PR #5441, opened 2023-03-14
build: Move mimalloc dependency to VVL from VkLayer_utils

jeremyg-lunarg:
VkLayer_utils is used by the LunarG/VulkanTools repository and having it require mimalloc causes build problems.


jeremyg-lunarg:
This change should fix the SDK build problem. Performance looks similar to previous code see vvltraceperf jobs 1771 and 1772 on internal CI.

juan-lunarg (approved):

PR #5447, opened 2023-03-15
Fixes access violation on dispatch with unbound descriptor set

vladislavbelov:
I caught access violation inside `vkCmdDispatch` (with unbound but active descriptor set) in `VkLayer_khronos_validation.dll` during implementing compute for [0 A.D.](https://play0ad.com/). I've investigated further and it seems `max_active_slot` was always `0` for compute pipeline states. Because of that `IsBoundSetCompat` was returning `true` even with a missing active descriptor set:

https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/8e4ba9f4c19c962c87db09db2d46b7b66d361a45/layers/core_checks/drawdispatch_validation.cpp#L2852

And that was causing out of bounds in (because `set_index` was `1`):

https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/8e4ba9f4c19c962c87db09db2d46b7b66d361a45/layers/core_checks/drawdispatch_validation.cpp#L2876

I wrote a test which reproduces the problem without the fix.


juan-lunarg (approved):
LGTM!

I tested locally on my machine to verify the access violation/fix

juan-lunarg:
Thanks for the fix @vladislavbelov 👍🏾

PR #5449, opened 2023-03-15
android: Update BUILD.md for manual library installation

mark-lunarg:
It is not possible to manually copy/install layer libraries on a development device if they are built with c++_shared. As this remains common practice, added a note to enable installing layer binaries without using an APK.

Change-Id: I4e315204afd1541a4eb50e73784a0958725b46f1

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -391,6 +391,13 @@ demo APKs can be installed on production devices with:
 
     ./install_all.sh [-s <serial number>]
 
+> **NOTE:** Manual installation of Android layer libraries on development devices will require modifying [build-android/jni/Application.mk](build-android/jni/Application.mk) to use `libc++_static.so` instead of `libc++_shared.so`, like so:
+```patch
+-APP_STL := c++_shared
++APP_STL := c++_static
+```
+
+
 Note that there are no equivalent scripts on Windows yet, that work needs to
 be completed. The following per platform commands can be used for layer only
 builds:


juan-lunarg (commented):
Should we statically link by default?

juan-lunarg:
> Change-Id: I4e315204afd1541a4eb50e73784a0958725b46f1

?

BUILD.md
```diff
@@ -391,6 +391,13 @@ demo APKs can be installed on production devices with:
 
     ./install_all.sh [-s <serial number>]
 
+> **NOTE:** Manual installation of Android layer libraries on development devices will require modifying [build-android/jni/Application.mk](build-android/jni/Application.mk) to use `libc++_static.so` instead of `libc++_shared.so`, like so:
+
+        -APP_STL := c++_shared
```
spencer-lunarg, line 396: using

\`\`\`patch
   
\`\`\`

provides a better view:

```patch
-APP_STL := c++_shared
+APP_STL := c++_static
```

spencer-lunarg:
> Should we statically link by default?

https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/a64eec06b8a155c8deb3f27411c4a322faa95e9f

BUILD.md
mark-lunarg: Done.

BUILD.md
```diff
@@ -391,6 +391,13 @@ demo APKs can be installed on production devices with:
 
     ./install_all.sh [-s <serial number>]
 
+> **NOTE:** Manual installation of Android layer libraries on development devices will require modifying [build-android/jni/Application.mk](build-android/jni/Application.mk) to use `libc++_static.so` instead of `libc++_shared.so`, like so:
+```
```
ncesario-lunarg, line 395: I think you need to suffix this with `patch` to do what Spencer was talking about.

juan-lunarg (approved):

ncesario-lunarg (approved):
LGTM, but I don't think Spencer's comment is technically resolved.

BUILD.md
mark-lunarg: Ah, thanks @ncesario-lunarg, I was completely unaware of this markdown thingy.

PR #5450, opened 2023-03-15
Fix issues related to pipeline library sub state and shader module identifiers

ncesario-lunarg:
This PR does the following:
- `layers: Fix HasFullState`: Fixes issues where HasFullState() was incorrectly returning true
- `layers: Add shader identifier tracking`, `layers: User shader identifiers`: Adds state tracking for associating shader identifiers with shader state objects

### DIFF
--- a/layers/state_tracker/pipeline_state.h
+++ b/layers/state_tracker/pipeline_state.h
@@ -253,6 +253,15 @@ class PIPELINE_STATE : public BASE_NODE {
 
     bool IsGraphicsLibrary() const { return !HasFullState(); }
     bool HasFullState() const {
+        // First make sure that this pipeline is a "classic" pipeline, or is linked together with the appropriate sub-state
+        // libraries
+        if (graphics_lib_type && (graphics_lib_type != (VK_GRAPHICS_PIPELINE_LIBRARY_VERTEX_INPUT_INTERFACE_BIT_EXT |
+                                                        VK_GRAPHICS_PIPELINE_LIBRARY_PRE_RASTERIZATION_SHADERS_BIT_EXT |
+                                                        VK_GRAPHICS_PIPELINE_LIBRARY_FRAGMENT_SHADER_BIT_EXT |
+                                                        VK_GRAPHICS_PIPELINE_LIBRARY_FRAGMENT_OUTPUT_INTERFACE_BIT_EXT))) {
+            return false;
+        }
+
         // If Pre-raster state does contain a vertex shader, vertex input state is not required
         const bool vi_satisfied = [this]() -> bool {
             if (pre_raster_state && pre_raster_state->vertex_shader) {
--- a/layers/state_tracker/state_tracker.h
+++ b/layers/state_tracker/state_tracker.h
@@ -1455,6 +1455,29 @@ class ValidationStateTracker : public ValidationObject {
         }
     }
 
+    inline std::shared_ptr<SHADER_MODULE_STATE> GetShaderModuleStateFromIdentifier(const VkShaderModuleIdentifierEXT& ident) {
+        ReadLockGuard guard(shader_identifier_map_lock_);
+        if (const auto itr = shader_identifier_map_.find(ident); itr != shader_identifier_map_.cend()) {
+            return itr->second;
+        }
+        return {};
+    }
+
+    inline std::shared_ptr<SHADER_MODULE_STATE> GetShaderModuleStateFromIdentifier(
+        const VkPipelineShaderStageModuleIdentifierCreateInfoEXT& shader_stage_id) const {
+        if (shader_stage_id.pIdentifier) {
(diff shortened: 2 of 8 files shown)


layers/state_tracker/pipeline_state.h
```diff
@@ -253,6 +253,15 @@ class PIPELINE_STATE : public BASE_NODE {
 
     bool IsGraphicsLibrary() const { return !HasFullState(); }
```
spencer-lunarg, line 254: (related)  curious if this could just check `graphics_lib_type == static_cast<VkGraphicsPipelineLibraryFlagsEXT>(0)`>

layers/state_tracker/pipeline_state.h
```diff
@@ -253,6 +253,15 @@ class PIPELINE_STATE : public BASE_NODE {
...
     bool HasFullState() const {
+        // First make sure that this pipeline is a "classic" pipeline, or is linked together with the appropriate sub-state
+        // libraries
+        if (graphics_lib_type && (graphics_lib_type != (VK_GRAPHICS_PIPELINE_LIBRARY_VERTEX_INPUT_INTERFACE_BIT_EXT |
```
spencer-lunarg, line 258: where is a case where this would have returned `true` but this prevent?

spencer-lunarg (approved):

layers/state_tracker/pipeline_state.h
ncesario-lunarg: Yeah, I almost made that change, but can't remember exactly why I didn't...probably just forgot.

layers/state_tracker/pipeline_state.h
ncesario-lunarg: I think this _shouldn't_ be necessary now. I added it while fixing a false positive, and left it in as an "early out."

layers/state_tracker/pipeline_state.h
ncesario-lunarg: I think I remember the motivation here now. `IsGraphicsLibrary` shouldn't return true if the pipeline contains a complete set of state (and is therefore an "executable pipeline," not a library). It is possible `graphics_lib_type != 0` in the case where a pipeline is created entirely from "linked" state (at least the way I currently have it implemented). It is also possible that `graphics_lib_type` only contains a subset of `VK_GRAPHICS_PIPELINE_LIBRARY*` bits, but the pipeline still contains a full set of state and is therefore not a library (e.g., some sub-state is linked in, the rest is provided in the create info). Going to leave this for now as it seems to be working.

juan-lunarg: > (related) curious if this could just check graphics_lib_type == static_cast<VkGraphicsPipelineLibraryFlagsEXT>(0)>

Should that functionality be in a helper function?
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5452/files

ncesario-lunarg: I'm not sure I understand the connection to #5452? I think the "helper function" is `HasFullState`, at least as I currently understand things.

juan-lunarg: I see the pattern `== static_cast<VkGraphicsPipelineLibraryFlagsEXT>(0)` repeated a bit in the codebase. And #5425 is an example of that.

ncesario-lunarg:
The internal CI failures seem to be sporadic; in the process of verifying that.

layers/state_tracker/pipeline_state.h
ncesario-lunarg: The context is slightly different in 5452, as that is when the sub-state is created, and checking for full/incomplete state does not make sense at that point. The name `IsGraphicsLibrary` may be overloaded, and that is the problem.

layers/state_tracker/state_tracker.h
```diff
@@ -1449,6 +1455,29 @@ class ValidationStateTracker : public ValidationObject {
...
+        }
+        return {};
+    }
+
+    inline std::shared_ptr<SHADER_MODULE_STATE> GetShaderModuleStateFromIdentifier(
+        const VkPipelineShaderStageModuleIdentifierCreateInfoEXT& shader_stage_id) const {
+        if (shader_stage_id.pIdentifier) {
+            auto shader_id = LvlInitStruct<VkShaderModuleIdentifierEXT>();
+            shader_id.identifierSize = shader_stage_id.identifierSize;
+            const uint32_t copy_size = std::min(VK_MAX_SHADER_MODULE_IDENTIFIER_SIZE_EXT, shader_stage_id.identifierSize);
```
ncesario-lunarg, line 1471: The nvidia driver on my system is returning `VK_MAX_SHADER_MODULE_IDENTIFIER_SIZE_EXT + 1` for `identifierSize`. That doesn't make sense to me, but I'm not that familiar with the extension. In any case, I should have added this to start with.

PR #5451, opened 2023-03-16
build: Compile vma/xxhash as system include directories

juan-lunarg:
- This is consistent with how we treat our other dependencies.
- Fixes clang-tidy issue (tidy ignores sys includes)
- Allow higher warnings for vvl


layers/CMakeLists.txt
```diff
@@ -54,6 +54,7 @@ target_sources(VkLayer_utils PRIVATE
 target_compile_definitions(VkLayer_utils PUBLIC XXH_NO_LONG_LONG)
 
 target_link_libraries(VkLayer_utils PUBLIC Vulkan::Headers)
+target_include_directories(VkLayer_utils SYSTEM PRIVATE external)
```
juan-lunarg, line 57: On GNU/GCC `SYSTEM` translate to `-isystem` (https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html#Directory-Options)

This is what CMake already does to dependencies found via `find_package` and is a useful tool to avoid analyzing third party code you don't control.

spencer-lunarg (approved):

PR #5452, opened 2023-03-16
layers: Remove all fragment shader requirements

ncesario-lunarg:
Make a fragment shader unconditionally optional.


juan-lunarg (commented):
Should a new test be added to cover this?

spencer-lunarg (approved):
LGTM - as Juan said, if there a test already great, if not it would be nice to have as this stuff can be fragile with upcoming extensions

ncesario-lunarg:
I likely need to follow up with an issue to the WG. This check has changed 3 times since the GPL extension was released...

Issue #4946, opened 2022-12-10
`vkCmdClearAttachments` is missing from sync validation


Rua:
I noticed that the `vkCmdClearAttachments` command is completely missing from the sync validation code. Neither the validation nor the recording of the command are processed at all. I don't know if it was missed, or if it's particularly hard to implement.

spencer-lunarg:
cc @jzulauf-lunarg @jeremyg-lunarg - either of you  know the answer to this?

jeremyg-lunarg:
> cc @jzulauf-lunarg @jeremyg-lunarg - either of you know the answer to this?

This is almost certainly an oversight.

jzulauf-lunarg:
@spencer-lunarg  -- should be straightforward.  Like clear, but with ordering rules.  The resources to clear will be looked up through the current renderpass.  There is a pre-baked range generator for each attachment.  You can look at other operations that affect attachments for examples.

PR #5455, opened 2023-03-16
vkCmdClearAttachments sync validation

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4946

### DIFF
--- a/layers/sync/sync_validation.cpp
+++ b/layers/sync/sync_validation.cpp
@@ -2776,6 +2776,174 @@ void RenderPassAccessContext::RecordDrawSubpassAttachment(const CMD_BUFFER_STATE
     }
 }
 
+static constexpr VkImageAspectFlags kColorAspects =
+    VK_IMAGE_ASPECT_COLOR_BIT | VK_IMAGE_ASPECT_PLANE_0_BIT | VK_IMAGE_ASPECT_PLANE_1_BIT | VK_IMAGE_ASPECT_PLANE_2_BIT;
+static constexpr VkImageAspectFlags kDepthStencilAspects = VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT;
+
+static std::optional<uint32_t> GetAttachmentIndex(const RenderPassAccessContext &rp_context,
+                                                  const VkClearAttachment &clear_attachment) {
+    const auto &rpci = rp_context.GetRenderPassState()->createInfo;
+    const auto &subpass = rpci.pSubpasses[rp_context.GetCurrentSubpass()];
+    uint32_t attachment_index = VK_ATTACHMENT_UNUSED;
+
+    if (clear_attachment.aspectMask & VK_IMAGE_ASPECT_COLOR_BIT) {
+        if (clear_attachment.colorAttachment < subpass.colorAttachmentCount) {
+            attachment_index = subpass.pColorAttachments[clear_attachment.colorAttachment].attachment;
+        }
+    } else if (clear_attachment.aspectMask & kDepthStencilAspects) {
+        if (subpass.pDepthStencilAttachment) {
+            attachment_index = subpass.pDepthStencilAttachment->attachment;
+        }
+    }
+    const bool invalid_index = (attachment_index == VK_ATTACHMENT_UNUSED || attachment_index >= rpci.attachmentCount);
+    return invalid_index ? std::optional<uint32_t>() : attachment_index;
+}
+
+static VkImageAspectFlags GetAspectsToClear(VkImageAspectFlags clear_aspect_mask, VkImageAspectFlags view_aspect_mask) {
+    // Check if clear request is valid.
+    const bool clear_color = (clear_aspect_mask & VK_IMAGE_ASPECT_COLOR_BIT) != 0;
+    const bool clear_depth = (clear_aspect_mask & VK_IMAGE_ASPECT_DEPTH_BIT) != 0;
+    const bool clear_stencil = (clear_aspect_mask & VK_IMAGE_ASPECT_STENCIL_BIT) != 0;
+    if (!clear_color && !clear_depth && !clear_stencil) {
+        return 0;  // nothing to clear
+    }
+    if (clear_color && (clear_depth || clear_stencil)) {
+        return 0;  // according to spec it's not allowed
+    }
+
+    // Collect aspects that should be cleared.
+    VkImageAspectFlags aspects_to_clear = 0;
+    if (clear_color && (view_aspect_mask & kColorAspects) != 0) {
+        assert(GetBitSetCount(view_aspect_mask) == 1);
+        aspects_to_clear |= view_aspect_mask;
+    }
+    if (clear_depth && (view_aspect_mask & VK_IMAGE_ASPECT_DEPTH_BIT) != 0) {
+        aspects_to_clear |= VK_IMAGE_ASPECT_DEPTH_BIT;
+    }
+    if (clear_stencil && (view_aspect_mask & VK_IMAGE_ASPECT_STENCIL_BIT) != 0) {
+        aspects_to_clear |= VK_IMAGE_ASPECT_STENCIL_BIT;
+    }
+    return aspects_to_clear;
+}
+
+static std::optional<VkImageSubresourceRange> RestrictSubresourceRange(const VkImageSubresourceRange &normalized_subresource_range,
+                                                                       const VkClearRect &clear_rect) {
+    assert(normalized_subresource_range.layerCount != VK_REMAINING_ARRAY_LAYERS);  // contract of this function
+    assert(clear_rect.layerCount != VK_REMAINING_ARRAY_LAYERS);                    // according to spec
+    const uint32_t first = std::max(normalized_subresource_range.baseArrayLayer, clear_rect.baseArrayLayer);
+    const uint32_t last_range = normalized_subresource_range.baseArrayLayer + normalized_subresource_range.layerCount;
+    const uint32_t last_clear = clear_rect.baseArrayLayer + clear_rect.layerCount;
+    const uint32_t last = std::min(last_range, last_clear);
+    std::optional<VkImageSubresourceRange> result;
+    if (first < last) {
+        result = normalized_subresource_range;
+        result->baseArrayLayer = first;
+        result->layerCount = last - first;
+    }
+    return result;
+}
+
+std::optional<RenderPassAccessContext::ClearAttachmentInfo> RenderPassAccessContext::GetClearAttachmentInfo(
+    const VkClearAttachment &clear_attachment, const VkClearRect &rect) const {
+    const auto attachment_index = GetAttachmentIndex(*this, clear_attachment);
+    if (!attachment_index) {
+        return {};
+    }
+    const auto &view_subresource_range = attachment_views_[attachment_index.value()].GetViewState()->normalized_subresource_range;
+    const auto aspects = GetAspectsToClear(clear_attachment.aspectMask, view_subresource_range.aspectMask);
+    if (!aspects) {
+        return {};
+    }
+    const auto subresource_range = RestrictSubresourceRange(view_subresource_range, rect);
+    if (!subresource_range) {
+        return {};
+    }
+    return ClearAttachmentInfo{attachment_index.value(), aspects, subresource_range.value()};
+}
+
+bool RenderPassAccessContext::ValidateClearAttachment(const CommandExecutionContext &exec_context,
+                                                      const CMD_BUFFER_STATE &cmd_buffer, CMD_TYPE cmd_type,
+                                                      const VkClearAttachment &clear_attachment, const VkClearRect &rect,
+                                                      uint32_t rect_index) const {
+    const auto info = GetClearAttachmentInfo(clear_attachment, rect);
+    if (!info) {
+        return false;
+    }
+    const auto &view_state = *attachment_views_[info->attachment_index].GetViewState();
+    const VkOffset3D offset = CastTo3D(rect.rect.offset);
+    const VkExtent3D extent = CastTo3D(rect.rect.extent);
+    auto subresource_range = info->subresource_range;
+    bool skip = false;
+
+    if (info->aspects_to_clear & kColorAspects) {
+        assert(GetBitSetCount(info->aspects_to_clear) == 1);
+        subresource_range.aspectMask = info->aspects_to_clear;
+
+        HazardResult hazard = CurrentContext().DetectHazard(
+            *view_state.image_state, SYNC_COLOR_ATTACHMENT_OUTPUT_COLOR_ATTACHMENT_WRITE, subresource_range,
+            SyncOrdering::kColorAttachment, offset, extent, view_state.IsDepthSliced());
+        if (hazard.hazard) {
+            const LogObjectList objlist(cmd_buffer.commandBuffer(), view_state.image_view());
+            skip |= exec_context.GetSyncState().LogError(
+                objlist, string_SyncHazardVUID(hazard.hazard),
+                "%s: Hazard %s when clearing pRects[%" PRIu32 "] region of color attachment %" PRIu32 " in subpass %" PRIu32
+                ". Access info %s.",
+                CommandTypeString(cmd_type), string_SyncHazard(hazard.hazard), rect_index, info->attachment_index,
+                cmd_buffer.GetActiveSubpass(), exec_context.FormatHazard(hazard).c_str());
+        }
+    }
+
+    constexpr VkImageAspectFlagBits depth_stencil_aspects[2] = {VK_IMAGE_ASPECT_DEPTH_BIT, VK_IMAGE_ASPECT_STENCIL_BIT};
+    for (const auto aspect : depth_stencil_aspects) {
+        if (info->aspects_to_clear & aspect) {
+            // Original aspect mask can contain both stencil and depth but here we track each aspect separately
+            subresource_range.aspectMask = aspect;
+
+            // vkCmdClearAttachments depth/stencil writes are executed by the EARLY_FRAGMENT_TESTS_BIT and LATE_FRAGMENT_TESTS_BIT
+            // stages. The implementation tracks the most recent access, which happens in the LATE_FRAGMENT_TESTS_BIT stage.
+            HazardResult hazard = CurrentContext().DetectHazard(
+                *view_state.image_state, SYNC_LATE_FRAGMENT_TESTS_DEPTH_STENCIL_ATTACHMENT_WRITE, subresource_range,
+                SyncOrdering::kDepthStencilAttachment, offset, extent, view_state.IsDepthSliced());
+
+            if (hazard.hazard) {
+                const LogObjectList objlist(cmd_buffer.commandBuffer(), view_state.image_view());
+                skip |= exec_context.GetSyncState().LogError(
+                    objlist, string_SyncHazardVUID(hazard.hazard),
+                    "%s: Hazard %s when clearing pRects[%" PRIu32 "] region of %s aspect of depth-stencil attachment %" PRIu32
+                    " in subpass %" PRIu32 ". Access info %s.",
+                    CommandTypeString(cmd_type), string_SyncHazard(hazard.hazard), rect_index, string_VkImageAspectFlagBits(aspect),
+                    info->attachment_index, cmd_buffer.GetActiveSubpass(), exec_context.FormatHazard(hazard).c_str());
+            }
+        }
+    }
+    return skip;
+}
+
+void RenderPassAccessContext::RecordClearAttachment(const CMD_BUFFER_STATE &cmd_buffer, ResourceUsageTag tag,
+                                                    const VkClearAttachment &clear_attachment, const VkClearRect &rect) {
+    const auto info = GetClearAttachmentInfo(clear_attachment, rect);
+    if (!info) {
+        return;
+    }
+    const auto &view_state = *attachment_views_[info->attachment_index].GetViewState();
+    const VkOffset3D offset = CastTo3D(rect.rect.offset);
+    const VkExtent3D extent = CastTo3D(rect.rect.extent);
+    auto subresource_range = info->subresource_range;
+
+    // Original subresource range can include aspects that are not cleared, they should not be tracked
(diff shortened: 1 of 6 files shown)


layers/sync/sync_validation.cpp
```diff
@@ -2774,6 +2774,161 @@ void RenderPassAccessContext::RecordDrawSubpassAttachment(const CMD_BUFFER_STATE
...
+            attachment_index = subpass.pColorAttachments[clear_attachment.colorAttachment].attachment;
+        }
+    } else if (clear_attachment.aspectMask & (VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT)) {
+        if (subpass.pDepthStencilAttachment) {
+            attachment_index = subpass.pDepthStencilAttachment->attachment;
+        }
+    }
+
+    if (attachment_index == VK_ATTACHMENT_UNUSED || attachment_index >= rpci.attachmentCount) return {};
+    return attachment_index;
```
juan-lunarg, line 2795: FFTI

```suggestion
    const bool invalid_index = (attachment_index == VK_ATTACHMENT_UNUSED || attachment_index >= rpci.attachmentCount);
    return invalid_index ? {} : attachment_index;
```

layers/sync/sync_validation.cpp
```diff
@@ -2774,6 +2774,161 @@ void RenderPassAccessContext::RecordDrawSubpassAttachment(const CMD_BUFFER_STATE
...
+    }
+    if (clear_depth_stencil && !can_clear_depth_stencil) {
+        return false;  // view does not share depth or stencil aspect with clear mask
+    }
+    return true;
+}
+
+static std::optional<VkImageSubresourceRange> RestrictSubresourceRange(const VkImageSubresourceRange &subresource_range,
+                                                                       const VkClearRect &clear_rect) {
+    const uint32_t first = std::max(subresource_range.baseArrayLayer, subresource_range.baseArrayLayer);
```
juan-lunarg, line 2824: ?

This looks like a mistake. Which might mean more testing is needed.

layers/sync/sync_validation.cpp
```diff
@@ -2774,6 +2774,161 @@ void RenderPassAccessContext::RecordDrawSubpassAttachment(const CMD_BUFFER_STATE
...
+    const uint32_t last_range = subresource_range.baseArrayLayer + subresource_range.layerCount;
```
juan-lunarg, line 2825: What if `layerCount == VK_REMAINING_ARRAY_LAYERS`?

See `ResolveRemainingLayers` helper function.

juan-lunarg (changes requested):

layers/sync/sync_validation.cpp
artem-lunarg: Nice catch, thanks!

tests/negative/sync_val.cpp
```diff
@@ -298,6 +298,241 @@ TEST_F(VkSyncValTest, Sync2BufferCopyHazards) {
...
+        // Read
+        vk::CmdCopyImage(*m_commandBuffer, rt, VK_IMAGE_LAYOUT_GENERAL, image, VK_IMAGE_LAYOUT_GENERAL, 1, &copy_region);
+        m_errorMonitor->VerifyFound();
+
+        m_commandBuffer->end();
+        m_commandBuffer->QueueCommandBuffer();
+        vk::QueueWaitIdle(m_device->m_queue);
+    }
+
+    // Positive test: regions touch but do not overlap. Copy to the first region, clear the second region.
```
juan-lunarg, line 505: Minor nitpick. Positive testing should go in the positive testing folder.

tests/framework/render.h
```diff
@@ -518,17 +518,17 @@ class VkImageObj : public vk_testing::Image {
         return ci;
     }
 
-    VkImageView targetView(VkImageViewCreateInfo ci) {
+    const VkImageView &targetView(VkImageViewCreateInfo ci) {
```
juan-lunarg, line 521: What's the advantage of const reference here? VkImageView is just a handle (IE a pointer).

layers/sync/sync_validation.cpp
artem-lunarg: It can't be in this context. Normalized subresource range is passed as the first parameter (`view_state.normalized_subresource_range`). And according VkClearRect documentation it does not use special numbers. But I will add assert to check anyway, since bugs like these are likely.

tests/framework/render.h
artem-lunarg: So, it's similar to what we have for some other types. It helps with a situation when Vulkan API accepts pointer to handle. Then it's possible to write
```
vk_struct.pView = &targetView();
```

Otherwise we need local variable.
Initially I used this in this PR but then needed array, so it's unused now.
One example is `Handle` type:
```
template <typename T>
class Handle {
  public:
    const T &handle() const noexcept { return handle_; }
```

juan-lunarg: Gotcha that makes sense :+1:

layers/sync/sync_validation.cpp
artem-lunarg: Added positive test that caught that issue.

tests/negative/sync_val.cpp
artem-lunarg: New positive test now in the new positive folder. The old positive test is part of  a single scenario, so I keep it, but removed the comment that it's a positive test. FWIW, we sometimes do that in the negative tests - show that some parameters do not cause errors before/after going with the negative test scenario.

layers/sync/sync_validation.cpp
```diff
@@ -2776,6 +2776,163 @@ void RenderPassAccessContext::RecordDrawSubpassAttachment(const CMD_BUFFER_STATE
...
+
+static std::optional<VkImageSubresourceRange> RestrictSubresourceRange(const VkImageSubresourceRange &normalized_subresource_range,
+                                                                       const VkClearRect &clear_rect) {
+    assert(normalized_subresource_range.layerCount != VK_REMAINING_ARRAY_LAYERS);  // contract of this function
+    assert(clear_rect.layerCount != VK_REMAINING_ARRAY_LAYERS);                    // according to spec
+    const uint32_t first = std::max(normalized_subresource_range.baseArrayLayer, clear_rect.baseArrayLayer);
+    const uint32_t last_range = normalized_subresource_range.baseArrayLayer + normalized_subresource_range.layerCount;
+    const uint32_t last_clear = clear_rect.baseArrayLayer + clear_rect.layerCount;
+    const uint32_t last = std::min(last_range, last_clear);
+    if (first >= last) return {};
```
spencer-lunarg, line 2832: so is the `{}` here the same as returning `std::optional<VkImageSubresourceRange>()`?

Or is this returning an empty struct?

spencer-lunarg (approved):

layers/sync/sync_validation.cpp
artem-lunarg: Good question. I checked and this constructs empty `std::optional`. It's also possible to return `std::nullopt` to make it more clear.

juan-lunarg: I personally find returning `{};` readable here. Since the function is small enough you can see the return type.

ncesario-lunarg (approved):
LGTM, but it would be good if one of the sync wizards could also sign off.

layers/sync/sync_validation.cpp
```diff
@@ -2776,6 +2776,163 @@ void RenderPassAccessContext::RecordDrawSubpassAttachment(const CMD_BUFFER_STATE
...
+}
+
+static bool CanClear(VkImageAspectFlags clear_aspect_mask, VkImageAspectFlags view_aspect_mask) {
+    const bool clear_color = (clear_aspect_mask & VK_IMAGE_ASPECT_COLOR_BIT) != 0;
+    const bool clear_depth_stencil = (clear_aspect_mask & (VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT)) != 0;
+
+    const bool can_clear_color = (view_aspect_mask & (VK_IMAGE_ASPECT_COLOR_BIT | VK_IMAGE_ASPECT_PLANE_0_BIT |
+                                                      VK_IMAGE_ASPECT_PLANE_1_BIT | VK_IMAGE_ASPECT_PLANE_2_BIT)) != 0;
+    const bool can_clear_depth_stencil =
+        ((clear_aspect_mask & view_aspect_mask) & (VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT)) != 0;
```
jzulauf-lunarg, line 2807: why is the `clear_aspect_mask` in this?

layers/sync/sync_validation.cpp
```diff
@@ -2776,6 +2776,163 @@ void RenderPassAccessContext::RecordDrawSubpassAttachment(const CMD_BUFFER_STATE
...
+    assert(clear_rect.layerCount != VK_REMAINING_ARRAY_LAYERS);                    // according to spec
+    const uint32_t first = std::max(normalized_subresource_range.baseArrayLayer, clear_rect.baseArrayLayer);
+    const uint32_t last_range = normalized_subresource_range.baseArrayLayer + normalized_subresource_range.layerCount;
+    const uint32_t last_clear = clear_rect.baseArrayLayer + clear_rect.layerCount;
+    const uint32_t last = std::min(last_range, last_clear);
+    if (first >= last) return {};
+    auto result = normalized_subresource_range;
+    result.baseArrayLayer = first;
+    result.layerCount = last - first;
+    return result;
```
jzulauf-lunarg, line 2836: Not a fan of this.  `result` is auto, so the type in non-obvious.  The construction of the returned optional is implicit... so it smells a bit magic.

Would think something like the below more clear/maintainable...  or is that too old school?

```
std::optional<VkImageSubresourceRange> result;
if (first >= last) {
    result.emplace(normalized_ ...
    result->baseArray ... // touch up result
    ...
}
return result;
```

layers/sync/sync_validation.cpp
```diff
@@ -2776,6 +2776,163 @@ void RenderPassAccessContext::RecordDrawSubpassAttachment(const CMD_BUFFER_STATE
     }
 }
 
+namespace ClearAttachmentHelper {
```
jzulauf-lunarg, line 2779: Not that we've been consistent about namespace uses, but I've been placing localized functionality like this a class and struct member of the the class needing the functionality, in this case `RenderPassAccessContext`.

Eventually, too support dynamic rendering they'll likely be some base RenderingAccessContext call for RP and dynamic rendering subclasses this could be moved into (or specialized between).

Also saves needing to namespace qualifiying in the RenderpassAccessContext methods below.

Style thoughts @ncesario-lunarg @juan-lunarg @jeremyg-lunarg ?

layers/sync/sync_validation.cpp
```diff
@@ -2776,6 +2776,163 @@ void RenderPassAccessContext::RecordDrawSubpassAttachment(const CMD_BUFFER_STATE
...
+            skip |= exec_context.GetSyncState().LogError(
+                objlist, string_SyncHazardVUID(hazard.hazard),
+                "%s: Hazard %s when clearing color attachment %" PRIu32 " in subpass %" PRIu32 ". Access info %s.",
+                CommandTypeString(cmd_type), string_SyncHazard(hazard.hazard), attachment_index.value(),
+                cmd_buffer.GetActiveSubpass(), exec_context.FormatHazard(hazard).c_str());
+        }
+    }
+    if (clear_attachment.aspectMask & VK_IMAGE_ASPECT_DEPTH_BIT) {
+        subresource_range->aspectMask = VK_IMAGE_ASPECT_DEPTH_BIT;
+        HazardResult hazard = CurrentContext().DetectHazard(image_state, SYNC_LATE_FRAGMENT_TESTS_DEPTH_STENCIL_ATTACHMENT_WRITE,
```
jzulauf-lunarg, line 2876: Probably needs a comment describing why LATE_FRAGMENT... is being used (because it's after EARLY and thus the "most recent access" per the design approach)

layers/sync/sync_validation.cpp
```diff
@@ -2776,6 +2776,163 @@ void RenderPassAccessContext::RecordDrawSubpassAttachment(const CMD_BUFFER_STATE
...
+    const uint32_t last = std::min(last_range, last_clear);
+    if (first >= last) return {};
+    auto result = normalized_subresource_range;
+    result.baseArrayLayer = first;
+    result.layerCount = last - first;
+    return result;
+}
+}  // namespace ClearAttachmentHelper
+
+bool RenderPassAccessContext::ValidateClearAttachment(const CommandExecutionContext &exec_context,
```
jzulauf-lunarg, line 2840: Probably want to report which rectangle from pRect's is hazarding, by passing in the rect index.

layers/sync/sync_validation.cpp
```diff
@@ -2776,6 +2776,163 @@ void RenderPassAccessContext::RecordDrawSubpassAttachment(const CMD_BUFFER_STATE
...
+    auto subresource_range = ClearAttachmentHelper::RestrictSubresourceRange(view_state.normalized_subresource_range, rect);
+    if (!subresource_range) {
+        return false;
+    }
+    const auto &image_state = *view_state.image_state;
+    const VkOffset3D offset = CastTo3D(rect.rect.offset);
+    const VkExtent3D extent = CastTo3D(rect.rect.extent);
+    bool skip = false;
+
+    if (clear_attachment.aspectMask & VK_IMAGE_ASPECT_COLOR_BIT) {
```
jzulauf-lunarg, line 2860: Hazard detection and reporting could be DRY'r.  See ValidateLoadOperation for ideas.

layers/sync/sync_validation.cpp
```diff
@@ -2776,6 +2776,163 @@ void RenderPassAccessContext::RecordDrawSubpassAttachment(const CMD_BUFFER_STATE
...
+            skip |= exec_context.GetSyncState().LogError(
+                objlist, string_SyncHazardVUID(hazard.hazard),
+                "%s: Hazard %s when clearing VK_IMAGE_ASPECT_DEPTH_BIT aspect of depth-stencil attachment %" PRIu32
+                " in subpass %" PRIu32 ". Access info %s.",
+                CommandTypeString(cmd_type), string_SyncHazard(hazard.hazard), attachment_index.value(),
+                cmd_buffer.GetActiveSubpass(), exec_context.FormatHazard(hazard).c_str());
+        }
+    }
+    if (clear_attachment.aspectMask & VK_IMAGE_ASPECT_STENCIL_BIT) {
+        subresource_range->aspectMask = VK_IMAGE_ASPECT_STENCIL_BIT;
```
jzulauf-lunarg, line 2890: we shouldn't be changing the subresource_range, because we need to check if the aspect is *both* in the view and the ClearAttachment correct?  What if the clear_attachment.aspectMask has both D/S and the view only has one... is that a noop, or just ignore the aspect not in the view?

layers/sync/sync_validation.cpp
```diff
@@ -2776,6 +2776,163 @@ void RenderPassAccessContext::RecordDrawSubpassAttachment(const CMD_BUFFER_STATE
...
+    const auto &image_state = *view_state.image_state;
+    const VkOffset3D offset = CastTo3D(rect.rect.offset);
+    const VkExtent3D extent = CastTo3D(rect.rect.extent);
+
+    if (clear_attachment.aspectMask & VK_IMAGE_ASPECT_COLOR_BIT) {
+        subresource_range->aspectMask = VK_IMAGE_ASPECT_COLOR_BIT;
+        CurrentContext().UpdateAccessState(image_state, SYNC_COLOR_ATTACHMENT_OUTPUT_COLOR_ATTACHMENT_WRITE,
+                                           SyncOrdering::kColorAttachment, subresource_range.value(), offset, extent, tag);
+    } else {
+        subresource_range->aspectMask &= (VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT);
```
jzulauf-lunarg, line 2930: The subresource range should only include the aspects in both the view and clear masks.

jzulauf-lunarg (changes requested):
A couple of nits, a couple correctness issues.

layers/sync/sync_validation.cpp
juan-lunarg: Personally since the functions are local to this cpp file I think a nameless namespace is most appropriate. Nameless namespaces are intended to replace the `static` keyword for functions (Which is pretty overloaded).

When I see a namespace I assume these functions are exposed in the public interface (IE the header file).

But these are just my opinions.

artem-lunarg: >  Nameless namespaces are intended to replace the static keyword for functions (Which is pretty overloaded)
There is one drawback in nameless namespaces, in my opinion. One example, we have a function in a cpp file:
```
void f() {}
```
And just by looking into it it's unclear if it's function with linkage or not because nameless namespace can be far away and static keyword is local to the function.

layers/sync/sync_validation.cpp
artem-lunarg: 💯

layers/sync/sync_validation.cpp
artem-lunarg: One example. `clear_aspect_mask` specifies `VK_IMAGE_ASPECT_DEPTH_BIT` and `view_aspect_mask` speficies `VK_IMAGE_ASPECT_STENCIL_BIT` then there is not depth aspect in the view to clear, according to spec nothing should happen. For color aspect we can have only single aspect and that's the reason why it have simple `can_clear_color` condition.

layers/sync/sync_validation.cpp
artem-lunarg: Initially it was struct and later I replaced it with a namespace. The idea was that these functions are related only to `clear attachment` code and not very useful in other parts. So `struct` might be an option if you think it's better.

juan-lunarg: I personally think the code is readable and straightforward.

layers/sync/sync_validation.cpp
artem-lunarg: This version looks good. One thing I personally prefer more, if we have opportunity to return specific value, for example `false` and then I like to return just that constant instead of relying on `result` variable's initial value. By returning result you still need to apply some min amount of mental effort to check that `result` was not changed after initialization. But it looks different people have different tastes here and I don't think this is something fundamental. Will go with suggested change.

layers/sync/sync_validation.cpp
artem-lunarg: Good catch, thanks!

In your example, I assumed it's noop. According to spec:
> If any attachment’s aspectMask to be cleared is not backed by an image view, the clear has no effect on that aspect.

artem-lunarg: Replaced `CanClear` helper with `GetAspectsToClear` that returns the exact set of aspects that should be cleared, thus it removes the need for more complex checks later that _view_ and _clear_ masks have to match.

View subresource's aspect still should be updated, added comments to clarify this.

layers/sync/sync_validation.cpp
```diff
@@ -2776,6 +2776,168 @@ void RenderPassAccessContext::RecordDrawSubpassAttachment(const CMD_BUFFER_STATE
...
+    if (!info) return false;
+
+    const auto &view_state = info->view_state;
+    const auto &image_state = *view_state.image_state;
+    const VkOffset3D offset = CastTo3D(rect.rect.offset);
+    const VkExtent3D extent = CastTo3D(rect.rect.extent);
+    auto subresource_range = info->subresource_range;
+    bool skip = false;
+
+    if (info->aspects_to_clear & VK_IMAGE_ASPECT_COLOR_BIT) {
```
jzulauf-lunarg, line 2873: If we change aspect_to_clear above, we'll need to change this test too.

layers/sync/sync_validation.cpp
```diff
@@ -2776,6 +2776,168 @@ void RenderPassAccessContext::RecordDrawSubpassAttachment(const CMD_BUFFER_STATE
...
+
+    const auto &image_state = *info->view_state.image_state;
+    const VkOffset3D offset = CastTo3D(rect.rect.offset);
+    const VkExtent3D extent = CastTo3D(rect.rect.extent);
+    auto subresource_range = info->subresource_range;
+
+    // Original subresource range can include aspects that won't be cleared, they should not be tracked
+    subresource_range.aspectMask = info->aspects_to_clear;
+
+    if (info->aspects_to_clear & VK_IMAGE_ASPECT_COLOR_BIT) {
```
jzulauf-lunarg, line 2930: and again here.

jzulauf-lunarg (changes requested):
Aspect mask issues, I believe.  Check on my assertion that only a single bit can be set, but in any case clear mask COLOR -> view aspect mask (either COLOR or the plane (planes)) in terms of the memory affected.

layers/sync/sync_validation.cpp
artem-lunarg: Merged some similar depth and stencil code and also added helper method to extract identical initialization in Validate and Record.

layers/sync/sync_validation.cpp
```diff
@@ -2776,6 +2776,186 @@ void RenderPassAccessContext::RecordDrawSubpassAttachment(const CMD_BUFFER_STATE
...
+    }
+    if (clear_color && (clear_depth || clear_stencil)) {
+        return 0;  // according to spec it's not allowed
+    }
+
+    // Collect aspects that should be cleared.
+    constexpr auto color_aspects =
+        VK_IMAGE_ASPECT_COLOR_BIT | VK_IMAGE_ASPECT_PLANE_0_BIT | VK_IMAGE_ASPECT_PLANE_1_BIT | VK_IMAGE_ASPECT_PLANE_2_BIT;
+    VkImageAspectFlags aspects_to_clear = 0;
+    if (clear_color && (view_aspect_mask & color_aspects) != 0) {
```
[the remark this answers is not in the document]
artem-lunarg: Thanks, that's valuable clarification. I discussed this some time ago with Spencer and there were open questions about this part, but this explanation makes sense and fits the spec.

artem-lunarg:
@jzulauf-lunarg the requested changes are implemented. Please fill free to check them when you have opportunity.

jzulauf-lunarg (approved):
LGTM

PR #5456, opened 2023-03-17
Tests cleanup: Remove old AccelerationStructureKHR in favor of RT helper classes defined in rt::as

arno-lunarg:
Bulk of the additions (~500 for a total of ~730) are in `tests\framework\ray_tracing_objects.h/.cpp`. They mostly consist in new methods to handle top level acceleration structures, and host acceleration structure operations.

The 1st commit fixes for `safe_VkAccelerationStructureGeometryKHR`, a memcpy was incorrect. The way `array_size` is computed is dubious, I created an issue to investigate that: https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5458

The 2nd removes `AccelerationStructureKHR` defined in `tests\framework\binding.h` in favor of classes from the `rt::as` namespace. It also adds the positive test `RayTracingAccelerationStructureReference`: it creates a bottom level acceleration structure on the device, then a top level acceleration structure that references the bottom level one.

The 3rd one removes old `AccelerationStructureKHR`


layers/utils/vk_layer_utils.h
```diff
@@ -185,6 +190,15 @@ static inline int u_ffs(int val) {
 #endif
 }
 
+// Given p2 a power of two, returns smallest multiple of p2 greater than or equal to x
+template <typename T>
+constexpr T align(T x, T p2) {
```
spencer-lunarg, line 195: 1. Something wrong with `std::align`?
2. should be `Align`

spencer-lunarg: also is this used anywhere?

arno-lunarg: `std::align` is a more complex beast: https://en.cppreference.com/w/cpp/memory/align
I use it in `BuildGeometryInfoKHR::BuildCommon` to align scratch buffer address
Will rename it! ( But why are names in this file all over the place x) )

spencer-lunarg: > std::align is a more complex beast:

Just please add a comment that this is suppose to be a simpler version of it so it doesn't get accidentally removed in the future

layers/generated/vk_safe_struct.cpp
```diff
@@ -60098,7 +60098,8 @@ safe_VkAccelerationStructureGeometryKHR::safe_VkAccelerationStructureGeometryKHR
         } else {
             size_t array_size = build_range_info->primitiveOffset + build_range_info->primitiveCount * sizeof(VkAccelerationStructureInstanceKHR);
```
artem-lunarg, line 60099: Would it be enough to allocate memory just for `build_range_info->primitiveCount` structures? Do we need part of allocation of the size  `build_range_info->primitiveOffset` bytes?

layers/utils/vk_layer_utils.h
```diff
@@ -153,7 +153,12 @@ static inline dispatch_key get_dispatch_key(const void *object) { return (dispat
 VkLayerInstanceCreateInfo *get_chain_info(const VkInstanceCreateInfo *pCreateInfo, VkLayerFunction func);
 VkLayerDeviceCreateInfo *get_chain_info(const VkDeviceCreateInfo *pCreateInfo, VkLayerFunction func);
 
-static inline bool IsPowerOfTwo(unsigned x) { return x && !(x & (x - 1)); }
+template <typename T>
+constexpr bool IsPowerOfTwo(T x) {
+    static_assert(std::numeric_limits<T>::is_integer, "Unsigned integer required.");
+    static_assert(!std::numeric_limits<T>::is_signed, "Unsigned integer required.");
```
artem-lunarg, line 159: super nit 🦸: `std::is_unsigned` might be a little bit easier for brain because no negation

layers/generated/vk_safe_struct.cpp
arno-lunarg: I think so too, but I need to explore how `safe_VkAccelerationStructureGeometryKHR` are used to make sure it will not break anything. For this PR, I am just going to fix the memcpy that is just wrong. I plan to have a deeper look at `safe_VkAccelerationStructureGeometryKHR` next

layers/utils/vk_layer_utils.h
arno-lunarg: Ah yes definitely, I was thinking that just `is_signed` existed!

layers/generated/vk_safe_struct.cpp
arno-lunarg: You make me realise that to keep the current weird logic, I also need to offset the destination address in the memcpy

arno-lunarg: I created an issue for the dubious `array_size` computation: https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5458

artem-lunarg: thanks! yes, it would be nice to figure out if some code relies on that, especially if other computations need to be adjusted to match current logic.

tests/framework/ray_tracing_objects.h
```diff
@@ -89,30 +113,33 @@ class AccelerationStructureKHR : public vk_testing::internal::NonDispHandle<VkAc
     AccelerationStructureKHR& SetOffset(VkDeviceSize offset);
     AccelerationStructureKHR& SetType(VkAccelerationStructureTypeKHR type);
     AccelerationStructureKHR& SetFlags(VkAccelerationStructureCreateFlagsKHR flags);
-    AccelerationStructureKHR& SetBuffer(vk_testing::Buffer&& buffer);
-    AccelerationStructureKHR& SetBufferMemoryAllocateFlags(VkMemoryAllocateFlags memory_allocate_flags);
-    AccelerationStructureKHR& SetBufferMemoryPropertyFlags(VkMemoryAllocateFlags memory_property_flags_);
+    AccelerationStructureKHR& SetDeviceBuffer(vk_testing::Buffer&& buffer);
+    AccelerationStructureKHR& SetDeviceBufferMemoryAllocateFlags(VkMemoryAllocateFlags memory_allocate_flags);
+    AccelerationStructureKHR& SetDeviceBufferMemoryPropertyFlags(VkMemoryAllocateFlags memory_property_flags_);
```
artem-lunarg, line 118: nit: last underscore

tests/framework/ray_tracing_objects.cpp
```diff
@@ -86,13 +116,57 @@ GeometryKHR &GeometryKHR::SetTrianglesIndexType(VkIndexType index_type) {
...
-    aabbs_.buffer = std::move(buffer);
-    vk_obj_.geometry.aabbs.data.deviceAddress = aabbs_.buffer.address(vk_api_version_);
+GeometryKHR &GeometryKHR::SetAABBsDeviceBuffer(vk_testing::Buffer &&buffer, VkDeviceSize stride /*= sizeof(VkAabbPositionsKHR)*/) {
+    aabbs_.device_buffer = std::move(buffer);
+    vk_obj_.geometry.aabbs.data.deviceAddress = aabbs_.device_buffer.address(vk_api_version_);
     vk_obj_.geometry.aabbs.stride = stride;
     return *this;
 }
 
+GeometryKHR &GeometryKHR::SetAABBsHostBuffer(std::unique_ptr<VkAabbPositionsKHR[]> &&buffer,
```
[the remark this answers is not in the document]
arno-lunarg: I was thinking that only an rvalue reference was possible here, thank you for showing me this. Clarity is the priority in those classes, so I am going to make this change!

PR #5457, opened 2023-03-17
Move more logic to Shader Module creation time

spencer-lunarg:
This change does

- Lower where "strip array" interface check is done
- Make `DecorationSet` a const ref
- Move `BuiltIn` checks to creation time when it is known

### DIFF
--- a/layers/best_practices/best_practices_utils.cpp
+++ b/layers/best_practices/best_practices_utils.cpp
@@ -1343,16 +1343,13 @@ bool BestPractices::PreCallValidateCreateComputePipelines(VkDevice device, VkPip
 
         if (IsExtEnabled(device_extensions.vk_khr_maintenance4)) {
             auto module_state = Get<SHADER_MODULE_STATE>(createInfo.stage.module);
-            if (module_state) {  // No module if creating from module identifier
-                for (const Instruction* inst : module_state->GetBuiltinDecorationList()) {
-                    if (inst->GetBuiltIn() == spv::BuiltInWorkgroupSize) {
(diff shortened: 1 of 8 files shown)


layers/best_practices/best_practices_utils.cpp
```diff
@@ -1343,16 +1343,13 @@ bool BestPractices::PreCallValidateCreateComputePipelines(VkDevice device, VkPip
...
-                    if (inst->GetBuiltIn() == spv::BuiltInWorkgroupSize) {
-                        skip |= LogWarning(device, kVUID_BestPractices_SpirvDeprecated_WorkgroupSize,
-                                           "vkCreateComputePipelines(): pCreateInfos[ %" PRIu32
-                                           "] is using the Workgroup built-in which SPIR-V 1.6 deprecated. The VK_KHR_maintenance4 "
-                                           "extension exposes a new LocalSizeId execution mode that should be used instead.",
-                                           i);
-                    }
-                }
+            if (module_state &&
+                module_state->static_data_.has_builtin_workgroup_size) {  // No module if creating from module identifier
```
ncesario-lunarg, line 1347: It might be more relevant to move this comment to the line above (though I realize this was probably clang-format's doing).

ncesario-lunarg (approved):
Looking good!

layers/best_practices/best_practices_utils.cpp
spencer-lunarg: ya, clang-format, can try and adjust it

PR #5464, opened 2023-03-19
layers: Label new pipeline layout consistency VUID

spencer-lunarg:
Adds new VUID for 1.3.244

Note, there was a bug in the Spec and currently being fixed (https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5795) but these are the correct VUID they should be


layers/core_checks/shader_validation.cpp
```diff
@@ -2699,9 +2718,43 @@ bool CoreChecks::ValidateVariables(const SHADER_MODULE_STATE &module_state) cons
...
                                                   const PIPELINE_STATE &pipeline,
-                                                  const std::vector<ResourceInterfaceVariable> &descriptor_variables,
-                                                  const std::string &vuid_layout_mismatch) const {
+                                                  const std::vector<ResourceInterfaceVariable> &descriptor_variables) const {
     bool skip = false;
+
+    std::string vuid_07988;
+    std::string vuid_07989;
+    std::string vuid_07990;
+    std::string vuid_07991;
```
ncesario-lunarg, line 2727: I wonder if something like the `DrawDispatch*` classes approach would make sense for VUID name lookup depending on how common this scenario is.

ncesario-lunarg (approved):
LGTM!

Heads up @lunarpapillo, as I'm guessing this will effect nightly trace runs when it goes in.

layers/core_checks/shader_validation.cpp
spencer-lunarg: So I just wanted to get this labeled for the sake of the header update, but once the spec bug gets fixed, I want to just make a `DrawDispatch`-ish class for pipelines, and move these checks all to a single area

arno-lunarg (approved):
LGTM!

mikes-lunarg:
> Note, there was a bug in the Spec and currently being fixed (https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5795) but these are the correct VUID they should be

Awesome, that explains the missing `VUID-VkRayTracingPipelineCreateInfoKHR-*` VUs I see when running the checker

PR #5469, opened 2023-03-20
Ensure pipeline layout is always present

ncesario-lunarg:
- Move 07826 check to pipeline creation time
- Add 07827

Marking as a draft for now as there may be other missing VUIDs that should be added as part of this change.


juan-lunarg (approved):
:+1:

spencer-lunarg (approved):

PR #5479, opened 2023-03-21
layers: Workaround VkStructureType collision

juan-lunarg:
Currently there is an sType collision causing problems with the intel driver.

### DIFF
--- a/layers/vk_layer_settings_ext.h
+++ b/layers/vk_layer_settings_ext.h
@@ -75,8 +75,10 @@
 // This extension is exclusively used by VVL, and is NOT intended as a deliverable.
 // The value of the VK_STRUCTURE_TYPE is arbitrary. The only requirement,
 // is that it must not conflict with existing sTypes.
+//
+// NOTE: VK_STRUCTURE_TYPE_MAX_ENUM - 1 is used by the intel driver.
 constexpr VkStructureType VK_STRUCTURE_TYPE_INSTANCE_LAYER_SETTINGS_EXT =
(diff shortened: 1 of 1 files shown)


mark-lunarg (approved):
Thus the magic number used previously!

charles-lunarg (approved):
This struct is the answer to life, the universe, and everything!

artem-lunarg (approved):
LGTM. The constant is correct.

layers/vk_layer_settings_ext.h
```diff
@@ -75,8 +75,10 @@
 // This extension is exclusively used by VVL, and is NOT intended as a deliverable.
 // The value of the VK_STRUCTURE_TYPE is arbitrary. The only requirement,
 // is that it must not conflict with existing sTypes.
+//
+// NOTE: VK_STRUCTURE_TYPE_MAX_ENUM - 1 is used by the intel driver.
 constexpr VkStructureType VK_STRUCTURE_TYPE_INSTANCE_LAYER_SETTINGS_EXT =
-    static_cast<VkStructureType>(VK_STRUCTURE_TYPE_MAX_ENUM - 1);
+    static_cast<VkStructureType>(VK_STRUCTURE_TYPE_MAX_ENUM - 42);
```
ncesario-lunarg, line 81: Hmm...does this help if another driver implementer is a fan of HG2G?

ncesario-lunarg (approved):

layers/vk_layer_settings_ext.h
juan-lunarg: It could potentially cause another collision in the future. AFAIK the only proper solution is to make this an official extension.

PR #5481, opened 2023-03-21
Refactoring how Variables are handles in Shader Module

spencer-lunarg:
Main goal - Group `OpVariable` into 3 types

- Stage interface (input/output)
  - Built Ins
  - User Defined
- Resource Variables

From here the goal is to do as much as possible during the init prasing of the shader

### DIFF
--- a/layers/error_message/validation_error_enums.h
+++ b/layers/error_message/validation_error_enums.h
@@ -39,9 +39,12 @@
 [[maybe_unused]] static const char *kVUID_Core_DrawState_InvalidImageView = "UNASSIGNED-CoreValidation-DrawState-InvalidImageView";
 
 [[maybe_unused]] static const char *kVUID_Core_Shader_InputNotProduced = "UNASSIGNED-CoreValidation-Shader-InputNotProduced";
-[[maybe_unused]] static const char *kVUID_Core_Shader_InterfaceTypeMismatch = "UNASSIGNED-CoreValidation-Shader-InterfaceTypeMismatch";
+[[maybe_unused]] static const char *kVUID_Core_Shader_VertexInputMismatch = "UNASSIGNED-CoreValidation-Shader-VertexInputMismatch";
+[[maybe_unused]] static const char *kVUID_Core_Shader_FragmentOutputMismatch = "UNASSIGNED-CoreValidation-Shader-FragmentOutputMismatch";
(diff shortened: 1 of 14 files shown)


layers/error_message/validation_error_enums.h
```diff
@@ -39,9 +39,12 @@
 [[maybe_unused]] static const char *kVUID_Core_DrawState_InvalidImageView = "UNASSIGNED-CoreValidation-DrawState-InvalidImageView";
 
 [[maybe_unused]] static const char *kVUID_Core_Shader_InputNotProduced = "UNASSIGNED-CoreValidation-Shader-InputNotProduced";
-[[maybe_unused]] static const char *kVUID_Core_Shader_InterfaceTypeMismatch = "UNASSIGNED-CoreValidation-Shader-InterfaceTypeMismatch";
+[[maybe_unused]] static const char *kVUID_Core_Shader_VertexInputMismatch = "UNASSIGNED-CoreValidation-Shader-VertexInputMismatch";
```
spencer-lunarg, line 42: broke up this VUID so when a real one gets assigned, easier to know which is which

tests/negative/geometry_tessellation.cpp
```diff
@@ -492,9 +492,8 @@ TEST_F(VkLayerTest, CreatePipelineExceedMaxTessellationControlInputOutputCompone
                 break;
             }
             case 1: {
-                // (in and out component limit) and (in and out location limit)
-                constexpr std::array vuids = {"VUID-RuntimeSpirv-Location-06272", "VUID-RuntimeSpirv-Location-06272",
-                                              "VUID-RuntimeSpirv-Location-06272", "VUID-RuntimeSpirv-Location-06272"};
+                // in and out component limit
```
spencer-lunarg, line 495: we were tracking the component count that it could trigger `VUID-RuntimeSpirv-Location-06272` in 2 different ways. After making a single way to track it, the duplicate VUs are now not triggered

tests/positive/shaderval.cpp
```diff
@@ -1982,8 +1982,7 @@ TEST_F(VkPositiveLayerTest, TestShaderInputAndOutputComponents) {
                 layout(location = 4, component = 0) in vec2 xy;
                 layout(location = 4, component = 2) in vec2 zw;
 
-                layout(location = 5, component = 0) in vec2 st;
```
spencer-lunarg, line 1985: was not valid, has `vec3` vertex Output and vector must match or be longer in Fragment input

This still does the test job of being different/mismatched between shaders

ncesario-lunarg (approved):
Great--and much needed--cleanup!

PR #5488, opened 2023-03-22
build: Fix LunarG/VulkanTools build

juan-lunarg:
Currently Lunarg/VulkanTools depends on VkLayer_utils, and recent changes broke the expected interface.


layers/vk_layer_logging.cpp
```diff
@@ -335,7 +335,8 @@ static bool LogMsgEnabled(const debug_report_data *debug_data, const std::string
     return true;
 }
 
-VKAPI_ATTR bool LogMsg(const debug_report_data *debug_data, VkFlags msg_flags, const LogObjectList &objects, const std::string &vuid_text, const char *format, va_list argptr) {
```
mark-lunarg, line 338: Hard to tell but some of these seem to be whitespace/formatting changes -- is this a generated file?  If not, it'd be better for such changes to be in a separate commit/PR maybe?

layers/vk_layer_logging.cpp
```diff
@@ -44,7 +44,7 @@ VKAPI_ATTR void SetDebugUtilsSeverityFlags(std::vector<VkLayerDbgFunctionState>
 }
 
 VKAPI_ATTR void RemoveDebugUtilsCallback(debug_report_data *debug_data, std::vector<VkLayerDbgFunctionState> &callbacks,
-                                            uint64_t callback) {
+                                         uint64_t callback) {
```
lunarpapillo: The changes in this file seem to be whitespace changes only.  Why are they here?

layers/vk_layer_logging.cpp
juan-lunarg: It's not a generated file the formatting was done on accident.

layers/xxhash.cpp
```diff
@@ -0,0 +1,25 @@
+/* Copyright (c) 2023 The Khronos Group Inc.
```
lunarpapillo, line 1: This file looks like a functional duplicate of the `layers/external/xhash.cpp` file that was marked deleted above - but the copyright notice is different.  The last time I checked, we weren't allowed to modify copyright notices in that way.

layers/vk_layer_logging.cpp
juan-lunarg: Fixed accidental formatting

lunarpapillo (changes requested):
I'm trying to understand the overview here..  With extensive movement of files in `BUILD.gn`, `CMakeLists.txt`, and `Android.mk`, how was the VVL build not completely broken before this commit?

BUILD.gn
```diff
@@ -139,8 +139,8 @@ core_validation_sources = [
   "layers/error_message/core_error_location.h",
   "layers/error_message/validation_error_enums.h",
   "layers/external/vma/vk_mem_alloc.h",
```
charles-lunarg, line 141: Why does vma/vk_mem_alloc.h live in external while vma.cpp doesn't?

layers/xxhash.cpp
juan-lunarg: This file wasn't created by xxhash devs. This is a file we created. It should have been using our copyright in the first place.

layers/vk_layer_logging.cpp
juan-lunarg: This has already been fixed

juan-lunarg:
> I'm trying to understand the overview here.. With extensive movement of files in `BUILD.gn`, `CMakeLists.txt`, and `Android.mk`, how was the VVL build not completely broken before this commit?

Because this doesn't affect the vvl build. It affects the Lunarg/VulkanTools build due to `VkLayer_utils` being installed. Which isn't tested anywhere

BUILD.gn
juan-lunarg: vma.cpp doesn't need to live there it's a file we own/created similar to xxhash.cpp

charles-lunarg: Gotcha, its a vendored file and lives in `layers/external`.

spencer-lunarg (approved):

spencer-lunarg:
For the short term I understand why this is needed, but we really should move away from being constrained in this repo to do things for other repos

ncesario-lunarg (changes requested):
I do not understand this change. If VulkanTools is referencing _source_ files in VVL (which shouldn't be the case, but I understand that's the current state of things), then why can VulkanTools not update? Either that, or we need to bite the bullet and fix the longstanding dependency issues we have here. This is undoing organizational changes I would rather not see undone.

juan-lunarg:
> This is undoing organizational changes I would rather not see undone.

There are multiple ways to fix this build breakage. I'll try another tempt that doesn't undo the organizational changes. I just did what I found easiest.

PR #5493, opened 2023-03-23
Change back to using libc++_static.so for Android

### LINKED ISSUE #5491: Android builds switching to use libc++_shared.so breaks some workflows
**Environment:**
 - OS: Android

**Describe the Issue**

Android allows layers to be injected into an application from outside of the APK, which used to work on older binary releases where the C++ library was statically linked with the layer. The recent move for binary releases on GitHub  to use a separate `libc++_shared.so` breaks this workflow as the Android dynamic linker cannot resolve the extra library even though the Vulkan loader finds the layer itself.

This workflow is particularly important for developer tools where we want to inject the layer at runtime without modifying the application APK, and for IHV/ODM workflows where we are often working with pre-existing APKs that cannot be rebuilt. 

**Expected behavior**

The Android workflow for installing and enabling a global layer is:

```
adb push /host/path/to/libVkLayer_khronos_validation.so /data/local/tmp
adb shell run-as <com.example.app> cp /data/local/tmp/libVkLayer_khronos_validation.so .
adb shell settings put global enable_gpu_debug_layers 1
adb shell settings put global gpu_debug_app <com.example.app>
adb shell settings put global gpu_debug_layers VK_LAYER_KHRONOS_validation
```

This works fine with the older statically linked layers from earlier releases. I've tried pushing `libc++_shared.so` to `/data/data/<com.example.app>/`, `/data/data/<com.example.app/lib/`, and `/data/data/<com.example.app/lib/arm64-v8a/` and none of them seem to resolve the library so the layer load fails. Error is:

```
failed to load layer library '/data/user/0/com.example.app/libVkLayer_khronos_validation.so': dlopen failed: library "libc++_shared.so" not found: 
```

Please can Android prebuilts revert back to statically linked C++ libraries, or at least provide both variants?

mark-lunarg:
Using the c++ shared library (libc++_shared.so) causes non-trivial install complications on Android, so changing back to using libc++_static.so.

Fixes #5491.

### DIFF
--- a/build-android/jni/Application.mk
+++ b/build-android/jni/Application.mk
@@ -1,5 +1,6 @@
-# Copyright 2015 The Android Open Source Project
-# Copyright (C) 2015 Valve Corporation
+# Copyright 2023 The Android Open Source Project
+# Copyright (C) 2023 Valve Corporation
+# Copyright (C) 2023 LunarG, Inc.
 
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -16,6 +17,6 @@
 APP_ABI := armeabi-v7a arm64-v8a x86 x86_64
 # APP_ABI := arm64-v8a   # just build for pixel2  (don't check in)
 APP_PLATFORM := android-26
-APP_STL := c++_shared
+APP_STL := c++_static
 NDK_TOOLCHAIN_VERSION := clang
 NDK_MODULE_PATH := .
--- a/BUILD.md
+++ b/BUILD.md
@@ -391,13 +391,6 @@ demo APKs can be installed on production devices with:
 
     ./install_all.sh [-s <serial number>]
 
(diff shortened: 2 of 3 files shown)


ncesario-lunarg (approved):
Fun!

build-android/jni/Application.mk
```diff
@@ -16,6 +17,6 @@
 APP_ABI := armeabi-v7a arm64-v8a x86 x86_64
 # APP_ABI := arm64-v8a   # just build for pixel2  (don't check in)
 APP_PLATFORM := android-26
-APP_STL := c++_shared
+APP_STL := c++_static
```
spencer-lunarg, line 20: @mark-lunarg could we write up a "why we decided static" comment here... 100% chance this question of using static or shared will come up again

mark-lunarg: Sure, justa minute.

PR #5494, opened 2023-03-23
build: Refactor VkLayer_util install files

ncesario-lunarg:
- Refactor install destination to match source
- Remove layers/generated from include directories

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,7 @@ build-android/third_party
 .editorconfig
 scripts/.vs
 .*.sw[op]
+*~
 # BUILD.gn artifacts
 .cipd/
 .gclient
--- a/layers/CMakeLists.txt
+++ b/layers/CMakeLists.txt
@@ -55,10 +55,7 @@ target_compile_definitions(VkLayer_utils PUBLIC XXH_NO_LONG_LONG)
 
 target_link_libraries(VkLayer_utils PUBLIC Vulkan::Headers)
 target_include_directories(VkLayer_utils SYSTEM PRIVATE external)
-target_include_directories(VkLayer_utils PUBLIC
-    .
-    generated # Needed otherwise VkLayer_utils won't install properly.
-)
+target_include_directories(VkLayer_utils PUBLIC .)
 
 find_package(robin_hood CONFIG)
 option(USE_ROBIN_HOOD_HASHING "robin_hood provides faster versions of std::unordered_map and std::unordered_set" ${robin_hood_FOUND})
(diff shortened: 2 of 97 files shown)


layers/CMakeLists.txt
```diff
@@ -73,7 +73,40 @@ endif()
 if(BUILD_LAYER_SUPPORT_FILES)
     # NOTE: We need to install header/source files for Android
     get_target_property(LAYER_UTIL_FILES VkLayer_utils SOURCES)
-    install(FILES ${LAYER_UTIL_FILES} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/vulkan)
```
ncesario-lunarg, line 76: This apparently flattens the directory structure, so we can't use this. If there's a way to preserve the directory structure, we could possibly do that, but then we would need to do something custom for the `generated` files. In general, being explicit with install files seems like the safest thing to do.

juan-lunarg (commented):
This doesn't take into account fixing the Android builds. See the code in the layer factory. They also reference the directory structure of our project. Not just our install directory.

layers/CMakeLists.txt
```diff
@@ -67,13 +64,42 @@ if (USE_ROBIN_HOOD_HASHING)
     target_compile_definitions(robin_hood::robin_hood INTERFACE USE_ROBIN_HOOD_HASHING)
 endif()
 
-# Currently repos like LunarG/VulkanTools require source code from VVL
-# This results in VkLayer_utils needing to be installed. Ideally this is
-# cleaned up in the future since this is a result of history involving the mono-repo.
+# NOTE: We need to install header/source files for Android
+# TODO: This should be removed once the official "Vulkan Layer Utils" library is available for consumption
```
juan-lunarg, line 68:
```suggestion
# TODO: This should be removed once the official "Vulkan-Utility-Libraries" library is available for consumption
```

.gitignore
```diff
@@ -39,6 +39,7 @@ build-android/third_party
 .editorconfig
 scripts/.vs
 .*.sw[op]
+*~
```
juan-lunarg, line 42: ?

```suggestion
```

juan-lunarg (approved):
LGTM

juan-lunarg:
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/KhronosGroup/Vulkan-ValidationLayers?pullRequest=5494) before we can accept your contribution.

It looks like someone using the GitHub username nces has committed to this PR and would need to sign also.

.gitignore
ncesario-lunarg: Vim temporary files on some old versions of vim.

PR #5499, opened 2023-03-24
layers: Fix 245 rtx pipeline VUs

spencer-lunarg:
Language was changed in 245 headers

new VUIDs

- VUID-VkRayTracingPipelineCreateInfoKHR-pLibraryInfo-07999
- VUID-VkRayTracingPipelineCreateInfoKHR-flags-08700

### DIFF
--- a/layers/stateless/parameter_validation_utils.cpp
+++ b/layers/stateless/parameter_validation_utils.cpp
@@ -7114,24 +7114,22 @@ bool StatelessValidation::manual_PreCallValidateCreateRayTracingPipelinesKHR(
                              "pLibraryInfo and pLibraryInterface must be NULL.");
         }
         if (pCreateInfos[i].pLibraryInfo) {
-            if (pCreateInfos[i].pLibraryInfo->libraryCount == 0) {
-                if (pCreateInfos[i].stageCount == 0) {
-                    skip |= LogError(device, "VUID-VkRayTracingPipelineCreateInfoKHR-pLibraryInfo-03600",
-                                     "vkCreateRayTracingPipelinesKHR(): If pLibraryInfo is not NULL and its libraryCount is 0, "
-                                     "stageCount must not be 0.");
-                }
-                if (pCreateInfos[i].groupCount == 0) {
-                    skip |= LogError(device, "VUID-VkRayTracingPipelineCreateInfoKHR-pLibraryInfo-03601",
-                                     "vkCreateRayTracingPipelinesKHR(): If pLibraryInfo is not NULL and its libraryCount is 0, "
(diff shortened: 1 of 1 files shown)


layers/stateless/parameter_validation_utils.cpp
```diff
@@ -7114,24 +7114,22 @@ bool StatelessValidation::manual_PreCallValidateCreateRayTracingPipelinesKHR(
...
+            if ((pCreateInfos[i].pLibraryInfo->libraryCount > 0) && (pCreateInfos[i].pLibraryInterface == nullptr)) {
+                skip |= LogError(device, "VUID-VkRayTracingPipelineCreateInfoKHR-pLibraryInfo-03590",
+                                 "vkCreateRayTracingPipelinesKHR(): If pLibraryInfo is not NULL and its libraryCount member "
+                                 "is greater than 0, its pLibraryInterface member must not be NULL.");
+            }
+        }
+        if ((pCreateInfos[i].pLibraryInfo == nullptr) || (pCreateInfos[i].pLibraryInfo->libraryCount == 0)) {
+            if (pCreateInfos[i].stageCount == 0) {
+                skip |= LogError(device, "VUID-VkRayTracingPipelineCreateInfoKHR-pLibraryInfo-07999",
+                                 "vkCreateRayTracingPipelinesKHR(): If pLibraryInfo is NULL or libraryCount is 0, "
```
artem-lunarg, line 7126: The suggestion is to use "or _its_ libraryCount" for consistency with the previous message and with  the spec.

layers/stateless/parameter_validation_utils.cpp
```diff
@@ -7114,24 +7114,22 @@ bool StatelessValidation::manual_PreCallValidateCreateRayTracingPipelinesKHR(
...
+                                 "stageCount must not be 0.");
+            }
+            if (pCreateInfos[i].groupCount == 0) {
+                skip |= LogError(device, "VUID-VkRayTracingPipelineCreateInfoKHR-flags-08700",
```
artem-lunarg, line 7130: There is this part of the vuid: `If flags does not include VK_PIPELINE_CREATE_LIBRARY_BIT_KHR` that probably should be added here. At least I can't see that the above checks can guarantee this condition is always true here.

spencer-lunarg: fixed it up... I plan to honestly just fix all these messages as they don't even list which index into the pipeline and all could be improved after this PR

artem-lunarg: I see, it could be some remarks not in the scope of this PR, so thanks for the additional fixes.

PR #5500, opened 2023-03-24
layers: PointSize not needed if not emit vertex

spencer-lunarg:
Original issue https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4657

Fix for ` VUID-VkGraphicsPipelineCreateInfo-Geometry-07725`

New 245 spec clarify that when not emitting any vertex, Geometry shaders don't need a PointSize

### DIFF
--- a/tests/positive/shaderval.cpp
+++ b/tests/positive/shaderval.cpp
@@ -951,6 +951,38 @@ TEST_F(VkPositiveLayerTest, PointSizeGeomShaderSuccess) {
     pipe.CreateGraphicsPipeline();
 }
 
+TEST_F(VkPositiveLayerTest, PointSizeGeomShaderDontEmit) {
+    TEST_DESCRIPTION("If vertex is not emitted, don't need Point Size in Geometry shader");
+
+    ASSERT_NO_FATAL_FAILURE(Init());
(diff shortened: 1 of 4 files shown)


tests/positive/shaderval.cpp
```diff
@@ -951,6 +951,38 @@ TEST_F(VkPositiveLayerTest, PointSizeGeomShaderSuccess) {
     pipe.CreateGraphicsPipeline();
 }
 
+TEST_F(VkLayerTest, PointSizeGeomShaderDontEmit) {
```
artem-lunarg, line 954: VkPositiveLayerTest

artem-lunarg (approved):

ncesario-lunarg (approved):
Trying to think of when you wouldn't want to produce geometry in a GS :thinking: ...

spencer-lunarg:
> Trying to think of when you wouldn't want to produce geometry in a GS thinking ...

https://gitlab.khronos.org/vulkan/vulkan/-/issues/3359

Sometime translation layers have an empty geometry shader and it is easier to leave it than have them detect the SPIR-V at runtime

PR #5501, opened 2023-03-24
Revert static lib changes to fix CI.

mark-lunarg:
Need to temporarily replace static libs with shared libs to fix CI though static is preferred long-term.

Tracking issue for static/shared lib issue is here: https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5508

### DIFF
--- a/build-android/jni/Application.mk
+++ b/build-android/jni/Application.mk
@@ -17,6 +17,8 @@
 APP_ABI := armeabi-v7a arm64-v8a x86 x86_64
 # APP_ABI := arm64-v8a   # just build for pixel2  (don't check in)
 APP_PLATFORM := android-26
-APP_STL := c++_static
+# Change back to using shared libs until the Android Tests issues
+# can be resolved
+APP_STL := c++_shared
 NDK_TOOLCHAIN_VERSION := clang
 NDK_MODULE_PATH := .
--- a/build-android/jni/shaderc/Application.mk
+++ b/build-android/jni/shaderc/Application.mk
@@ -1,4 +1,5 @@
 APP_ABI := all
 APP_BUILD_SCRIPT := Android.mk
-APP_STL := c++_static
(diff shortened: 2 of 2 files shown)


build-android/jni/Application.mk
```diff
@@ -1,4 +1,5 @@
 # Copyright 2023 The Android Open Source Project
+
```
juan-lunarg, line 2:
```suggestion
```

mark-lunarg: Ah, thanks.

juan-lunarg (approved):

build-android/jni/Application.mk
```diff
@@ -17,6 +17,10 @@
 APP_ABI := armeabi-v7a arm64-v8a x86 x86_64
 # APP_ABI := arm64-v8a   # just build for pixel2  (don't check in)
 APP_PLATFORM := android-26
-APP_STL := c++_static
+# There are advantages to building with c++_shared including fixing some
+# memory leaks but shared *requires* that layer libraries be installed
+# via an APK. However it is common practice to manually configure/install
+# layer libs which requires that c++_static be used.
+APP_STL := c++_shared
```
spencer-lunarg, line 24: wait... we are changing it back to shared? or is this a typo

mark-lunarg: @spencer-lunarg, yeah, I need to change it back to shared because static breaks CI.  I'm running through a 'final-ish' change one more time and if it passes I plan to push it until we can resolve these testing issues.

spencer-lunarg: sounds good, for the new "Until Android Issues are fixed" comment, do you have some tracking issue for that?

mark-lunarg: No, I'll make one.  I think this change looks good, the Pixel3 run failed so I'll run it through one more time but it does fix the crashes on all the devices that ran.

PR #5505, opened 2023-03-25
layers: Fix invalid RenderPass VU logic

spencer-lunarg:
Breaking down the logic

- 06574
  - **renderPass** must be a valid render pass object
- 06575
  - **renderPass** must be **VK_NULL_HANDLE** or a valid render pass object
- 06576
  - if **dynamicRendering** is not enabled, **renderPass** must not be **VK_NULL_HANDLE**
  
 This change
 
 - Makes logic easier to see
 - Gives dedicated message for each case
 - Removes check for API version (not needed)


spencer-lunarg:
Closing because realize the current logic takes in consideration if there is GPL linking of all the stages

PR #5506, opened 2023-03-25
 Add dynamic state viewport/scissors count

spencer-lunarg:
Adds `VUID-VkGraphicsPipelineCreateInfo-pDynamicStates-04130` and `VUID-VkGraphicsPipelineCreateInfo-pDynamicStates-04131` which were already basically implemented minus checking for the new Viewport/Scissor dynamic state added in `VK_EXT_extended_dynamic_state`


artem-lunarg:
These VUs also mention `If the pipeline is being created with pre-rasterization shader state`. Is it taken into account somewhere?

artem-lunarg (approved):

Issue #5431, opened 2023-03-14
Missing External memory/sync VUs


spencer-lunarg:
Some remaining external memory/sync VU that shouldn't be too hard

memory allocate time
- [x] VUID-VkMemoryAllocateInfo-pNext-00641
- [x] VUID-VkMemoryAllocateInfo-allocationSize-01742
- [x] VUID-VkMemoryAllocateInfo-allocationSize-01743
- [x] VUID-VkMemoryAllocateInfo-allocationSize-01745
- [x] VUID-VkMemoryAllocateInfo-allocationSize-07897
- [ ] VUID-VkMemoryAllocateInfo-memoryTypeIndex-00648
- [ ] VUID-VkMemoryAllocateInfo-None-00643
- [ ] VUID-VkMemoryAllocateInfo-None-00644
- [x] VUID-VkMemoryAllocateInfo-pNext-02806
- [ ] VUID-VkMemoryAllocateInfo-memoryTypeIndex-00645
- [x] VUID-VkMemoryAllocateInfo-memoryTypeIndex-01744

dedicated memory
- [x] VUID-VkMemoryAllocateInfo-pNext-00639
- [ ] VUID-VkMemoryDedicatedAllocateInfo-image-01876
- [ ] VUID-VkMemoryDedicatedAllocateInfo-buffer-01877
- [x] VUID-VkMemoryDedicatedAllocateInfo-image-01878
- [x] VUID-VkMemoryDedicatedAllocateInfo-buffer-01879

VK_KHR_external_memory_fd
- [x] VUID-vkGetMemoryFdPropertiesKHR-fd-00673
- [x] VUID-vkGetMemoryFdPropertiesKHR-handleType-00674
Some decisions behind design: https://gitlab.khronos.org/vulkan/vulkan/-/issues/3423#note_405999
- [x] VUID-VkMemoryGetFdInfoKHR-handleType-00671
- [x] VUID-VkMemoryGetFdInfoKHR-handleType-00672

VkImportMemoryFdInfoKHR
- [ ] VUID-VkImportMemoryFdInfoKHR-handleType-00667 (https://gitlab.khronos.org/vulkan/vulkan/-/issues/3667)
- [ ] VUID-VkImportMemoryFdInfoKHR-fd-00668
- [x] VUID-VkImportMemoryFdInfoKHR-handleType-00669
- [x] VUID-VkImportMemoryFdInfoKHR-handleType-00670
[...]

PR #5509, opened 2023-03-25
layers: Add VkMemoryAllocateInfo import VUs

artem-lunarg:
Part of https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5431

This PR validates `VkMemoryAllocateInfo`'s allocation size and memory type index during memory import operation:
`VUID-VkMemoryAllocateInfo-allocationSize-01742`
`VUID-VkMemoryAllocateInfo-allocationSize-01743`

Updated `ValidateImportMemoryHandleType` to skip the test if there is no memory type that supports import and export that use different handle types. This happens on AMD drivers and is a valid situation. There is no guarantee the same memory type is supported by different handle types.

One lesson from this PR is that in some cases external memory requires dedicated allocation. It's the case on Android Pixel 3. 
We have scheduled work for dedicated memory VUs to improve detection of such cases.

### DIFF
--- a/layers/core_checks/device_memory_validation.cpp
+++ b/layers/core_checks/device_memory_validation.cpp
@@ -357,7 +357,56 @@ bool CoreChecks::PreCallValidateAllocateMemory(VkDevice device, const VkMemoryAl
         }
     }
 
-    // TODO: VUIDs ending in 00643, 00644, 00646, 00647, 01742, 01743, 01745, 00645, 00648, 01744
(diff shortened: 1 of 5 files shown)


ncesario-lunarg (approved):
LGTM once the CI gremlins have been dealt with :)

artem-lunarg:
Yes, desktop is mostly fine, but fight with Android is ongoing!

spencer-lunarg (approved):

PR #5510, opened 2023-03-26
Enable static C++ std library use for non-CI Android builds

### LINKED ISSUE #5508: Using `c++_static` in Android builds causes random test crashes on all devices
Changing the android build to use `c++_static` instead of `c++_shared` for the layer and shaderc libs causes random crashes in the LVTs in CI.  The static lib used to work, was sensibly changed by JeremyG and those changes were reverted to fix some usability/install problems.  SDK testing showed that the reversion caused test crashes and this also became the case on Main. 

The change was removed from the SDK branch and the removal from main is in progress, but the goal is to build VVL with the static libs so an investigation/fix is needed.

Jeremy's original PR changing from static to shared: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4408

PR that reverted the shared changes: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5493

PR changing to shared again (at least temporarily) to get CI running cleanly: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5501

mark-lunarg:
Android builds targeted at CI testing need to use the shared C++ standard library, but Android binary releases and manually built layers need to use the static C++ standard library.

This PR modifies the `build-android\build_all.sh` script to produce builds using the static library.  Github Actions and LunarG internal CI do not use this script and will continue to use the shared version of the library.

See tracking issue #5508 for background and related issues.

Also added some descriptive text in the Android build scripts to server as guidance when the Android build is converted to use CMake.

Fixes #5508.

### DIFF
--- a/build-android/build_all.sh
+++ b/build-android/build_all.sh
@@ -45,9 +45,7 @@ findtool apksigner
 set -ev
 
 LAYER_BUILD_DIR=$PWD
-DEMO_BUILD_DIR=$PWD/../demos/android
 echo LAYER_BUILD_DIR="${LAYER_BUILD_DIR}"
-echo DEMO_BUILD_DIR="${DEMO_BUILD_DIR}"
 
 #
 # Android builds and c++ libraries:
--- a/build-android/build_all.sh
+++ b/build-android/build_all.sh
@@ -49,6 +49,31 @@ DEMO_BUILD_DIR=$PWD/../demos/android
 echo LAYER_BUILD_DIR="${LAYER_BUILD_DIR}"
 echo DEMO_BUILD_DIR="${DEMO_BUILD_DIR}"
 
+#
+# Android builds and c++ libraries:
+#
+#   https://developer.android.com/ndk/guides/cpp-support recommends using
+#   c++_shared for applications that use more than one shared library.
+#   If multiple libraries using c++_static are loaded several copies of
+#   the globals will be present in the C++ runtime. This also happens
+#   if the same library is dlopen/dlclosed several times, as when running
+#   the Layer Validation Tests. Some of the c++ runtime globals are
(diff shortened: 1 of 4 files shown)


mark-lunarg:
@jeremyg-lunarg, @ncesario-lunarg, @juan-lunarg, PTAL.

build-android/build_all.sh
```diff
@@ -45,9 +45,30 @@ findtool apksigner
...
+#   repeatedly loading and unloading VVL.
+#
+#   The drawback to using c++shared is that the layer library can no longer
+#   be installed manually, but must be installed in an APK. It is still
+#   common practice to load layer libraries manually, so this script will
+#   build the layer library using c++static by default.
+#
+# To build the layer libaries for running the layer validation tests on Android, c++_shared needs to be used for APP_STL. To do this set
+# C_LIB_TYPE to 'SHARED' on the following line:
+export C_LIB_TYPE=STATIC
```
juan-lunarg, line 70: Allow the user to override C_LIB_TYPE

```suggestion
if [ -z "${C_LIB_TYPE}" ]; then
    export C_LIB_TYPE=STATIC
fi
```

juan-lunarg (commented):
What about naming the environment variable `ANDROID_STL_TYPE`?

Since many users are familiar with `CMAKE_ANDROID_STL_TYPE` which I believe have the same intent.

https://cmake.org/cmake/help/latest/variable/CMAKE_ANDROID_STL_TYPE.html?highlight=cmake_android_stl_type

build-android/build_all.sh
mark-lunarg: Great, done.

juan-lunarg (approved):

PR #5511, opened 2023-03-26
layers: Fix and Label MaxFramebuffer extent

spencer-lunarg:
When using `VK_KHR_dynamic_rendering` the `VUID-VkRenderingInfo-pNext-07815` and `VUID-VkRenderingInfo-pNext-07816` VUs were missing

Fixed logic as the VUs were gated by the `VK_KHR_device_group` being enabled which is wrong, only the VUID changes

### DIFF
--- a/layers/core_checks/render_pass_validation.cpp
+++ b/layers/core_checks/render_pass_validation.cpp
@@ -2855,7 +2855,7 @@ bool CoreChecks::ValidateRenderingInfoAttachment(const std::shared_ptr<const IMA
         if (!device_group_render_pass_begin_info || device_group_render_pass_begin_info->deviceRenderAreaCount == 0) {
             if (!x_extent_valid) {
                 skip |= LogError(image_view->Handle(), "VUID-VkRenderingInfo-pNext-06079",
-                                 "%s(): %s width (%" PRIu32 ") is less than pRenderingInfo->renderArea.offset.x (%" PRIu32
+                                 "%s(): %s width (%" PRIu32 ") is less than pRenderingInfo->renderArea.offset.x (%" PRIi32
                                  ") + pRenderingInfo->renderArea.extent.width (%" PRIu32 ").",
                                  func_name, attachment, image_view->image_state->createInfo.extent.width,
                                  pRenderingInfo->renderArea.offset.x, pRenderingInfo->renderArea.extent.width);
@@ -2863,7 +2863,7 @@ bool CoreChecks::ValidateRenderingInfoAttachment(const std::shared_ptr<const IMA
 
             if (!y_extent_valid) {
                 skip |= LogError(image_view->Handle(), "VUID-VkRenderingInfo-pNext-06080",
-                                 "%s(): %s height (%" PRIu32 ") is less than pRenderingInfo->renderArea.offset.y (%" PRIu32
+                                 "%s(): %s height (%" PRIu32 ") is less than pRenderingInfo->renderArea.offset.y (%" PRIi32
                                  ") + pRenderingInfo->renderArea.extent.width (%" PRIu32 ").",
                                  func_name, attachment, image_view->image_state->createInfo.extent.height,
                                  pRenderingInfo->renderArea.offset.y, pRenderingInfo->renderArea.extent.height);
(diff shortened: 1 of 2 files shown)


layers/core_checks/render_pass_validation.cpp
```diff
@@ -3326,18 +3326,37 @@ bool CoreChecks::ValidateCmdBeginRendering(VkCommandBuffer commandBuffer, const
...
-
-            if (pRenderingInfo->renderArea.offset.y < 0) {
-                skip |= LogError(commandBuffer, "VUID-VkRenderingInfo-pNext-06078",
-                                 "%s(): renderArea.offset.y is %d and must be greater than 0.", func_name,
-                                 pRenderingInfo->renderArea.offset.y);
-            }
+        if (pRenderingInfo->renderArea.offset.x < 0) {
+            const char *vuid = IsExtEnabled(device_extensions.vk_khr_device_group) ? "VUID-VkRenderingInfo-pNext-06077"
+                                                                                   : "VUID-VkRenderingInfo-renderArea-06071";
+            skip |= LogError(commandBuffer, vuid, "%s(): pRenderingInfo->renderArea.offset.x (%" PRIu32 ") must not be negative.",
```
artem-lunarg, line 3332: PRIi32

layers/core_checks/render_pass_validation.cpp
```diff
@@ -3326,18 +3326,37 @@ bool CoreChecks::ValidateCmdBeginRendering(VkCommandBuffer commandBuffer, const
...
+            skip |= LogError(commandBuffer, vuid, "%s(): pRenderingInfo->renderArea.offset.y (%" PRIu32 ") must not be negative.",
+                             func_name, pRenderingInfo->renderArea.offset.y);
+        }
+        if (x_adjusted_extent > phys_dev_props.limits.maxFramebufferWidth) {
+            const char *vuid = IsExtEnabled(device_extensions.vk_khr_device_group) ? "VUID-VkRenderingInfo-pNext-07815"
+                                                                                   : "VUID-VkRenderingInfo-renderArea-06073";
+            skip |= LogError(commandBuffer, vuid,
+                             "%s(): pRenderingInfo->renderArea.offset.x (%" PRIu32
+                             ") + pRenderingInfo->renderArea.extent.width (%" PRIu32
+                             ") is not less than maxFramebufferWidth (%" PRIu32 ").",
```
artem-lunarg, line 3347: One more comment not directly related to this PR but still looks like a small change to improve correctness.

"is not less than": the correct usage is less than or _equal_.

layers/core_checks/render_pass_validation.cpp
```diff
@@ -3326,18 +3326,37 @@ bool CoreChecks::ValidateCmdBeginRendering(VkCommandBuffer commandBuffer, const
...
+            const char *vuid = IsExtEnabled(device_extensions.vk_khr_device_group) ? "VUID-VkRenderingInfo-pNext-06078"
+                                                                                   : "VUID-VkRenderingInfo-renderArea-06072";
+            skip |= LogError(commandBuffer, vuid, "%s(): pRenderingInfo->renderArea.offset.y (%" PRIi32 ") must not be negative.",
+                             func_name, pRenderingInfo->renderArea.offset.y);
+        }
+        if (x_adjusted_extent > phys_dev_props.limits.maxFramebufferWidth) {
+            const char *vuid = IsExtEnabled(device_extensions.vk_khr_device_group) ? "VUID-VkRenderingInfo-pNext-07815"
+                                                                                   : "VUID-VkRenderingInfo-renderArea-06073";
+            skip |= LogError(commandBuffer, vuid,
+                             "%s(): pRenderingInfo->renderArea.offset.x (%" PRIu32
```
artem-lunarg, line 3345: One more place for PRIi32. I'm not sure if we have some singed/unsigned warnings disabled or that's the compiler does not detect this, but ideally would be nice to get compiler notifications.

artem-lunarg (approved):
LGTM! One survived PRIu32 to update.

tests/negative/dynamic_rendering.cpp
```diff
@@ -2412,12 +2412,14 @@ TEST_F(VkLayerTest, DynamicRenderingTestFragmentDensityMapRenderArea) {
 
     begin_rendering_info.renderArea.offset.x = 1;
     begin_rendering_info.renderArea.extent.width = vvl::MaxTypeValue(begin_rendering_info.renderArea.extent.width) - 1;
+    m_errorMonitor->SetUnexpectedError("VUID-VkRenderingInfo-pNext-07815");  // if over max
```
juan-lunarg, line 2415: > The long-term goal is to remove the use of this function and its definition.

Why use `SetUnexpectedError`?

spencer-lunarg: Because this VU is for `maxFramebufferWidth`

It is useful to use `SetUnexpectedError` in this case IMO because keeps to test simple not worrying about every device's limits as it still hits the targeted VU

> The long-term goal is to remove the use of this function and its definition.

Didn't know that was in the comment. I do believe we sometime abuse this function sometimes really badly

PR #5512, opened 2023-03-26
layers: Dynamic Rendering maxFramebufferLayers

spencer-lunarg:
Adds `VUID-VkRenderingInfo-layerCount-07817` which is `VUID-VkFramebufferCreateInfo-layers-00890` but for Dynamic Rendering

### DIFF
--- a/tests/negative/dynamic_rendering.cpp
+++ b/tests/negative/dynamic_rendering.cpp
@@ -6818,6 +6818,43 @@ TEST_F(VkLayerTest, DynamicRenderingInvalidDeviceGroupRenderArea) {
     m_commandBuffer->end();
 }
 
+TEST_F(VkLayerTest, DynamicRenderingMaxFramebufferLayers) {
+    TEST_DESCRIPTION("Go over maxFramebufferLayers");
+    SetTargetApiVersion(VK_API_VERSION_1_1);
+    AddRequiredExtensions(VK_KHR_DYNAMIC_RENDERING_EXTENSION_NAME);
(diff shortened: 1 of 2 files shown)


tests/negative/dynamic_rendering.cpp
```diff
@@ -6818,6 +6818,43 @@ TEST_F(VkLayerTest, DynamicRenderingInvalidDeviceGroupRenderArea) {
...
+        GTEST_SKIP() << "Test requires (unsupported) dynamicRendering";
+    }
+
+    ASSERT_NO_FATAL_FAILURE(InitState(nullptr, &features2));
+
+    VkRenderingAttachmentInfoKHR color_attachment = LvlInitStruct<VkRenderingAttachmentInfoKHR>();
+    color_attachment.imageLayout = VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL;
+
+    auto begin_rendering_info = LvlInitStruct<VkRenderingInfoKHR>();
+    begin_rendering_info.layerCount = m_device->props.limits.maxFramebufferLayers + 1;
```
artem-lunarg, line 6846: One paranoid thought what if some driver reports `maxFramebufferLayers` as max uint32. In that case `std::min(m_device->props.limits.maxFramebufferLayers, kU32Max - 1) + 1` will help. But I'm not sure if this brings more confusion to the code than actual improvement. But will put it here for documentation purposes.

artem-lunarg (approved):

PR #5516, opened 2023-03-26
layers: Label non Sync2 STATE_NONE VUIDs

spencer-lunarg:
Add missing "StageMask must not be 0" when not using Sync2

Before there use to be a `VUID-vkCmdPipelineBarrier-srcStageMask-requiredbitmask` but it was fixed in an XML change a while ago and now became `VUID-vkCmdPipelineBarrier-srcStageMask-04996`

also fixed `VUID-vkCmdWriteTimestamp-pipelineStage-04074` should have been `VUID-vkCmdWriteTimestamp-synchronization2-06489`

### DIFF
--- a/layers/sync/sync_vuid_maps.cpp
+++ b/layers/sync/sync_vuid_maps.cpp
@@ -266,26 +266,43 @@ static const std::map<VkPipelineStageFlags2KHR, std::vector<Entry>> kStageMaskEr
          {Key(Struct::VkSubpassDependency2, Field::srcStageMask), "VUID-VkSubpassDependency2-srcStageMask-04094"},
          {Key(Struct::VkSubpassDependency2, Field::dstStageMask), "VUID-VkSubpassDependency2-dstStageMask-04094"},
      }},
-    // special case for the NONE stage. This entry omits the synchronization2
-    // commands because they shouldn't be called unless synchronization2 is enabled.
-    {VK_PIPELINE_STAGE_2_NONE_KHR,
-     {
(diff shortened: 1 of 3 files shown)


artem-lunarg (approved):

layers/sync/sync_vuid_maps.cpp
```diff
@@ -266,26 +266,43 @@ static const std::map<VkPipelineStageFlags2KHR, std::vector<Entry>> kStageMaskEr
...
-         {Key(Struct::VkSubmitInfo, Field::pWaitDstStageMask), "VUID-VkSubmitInfo-pWaitDstStageMask-03937"},
-         {Key(Struct::VkSubpassDependency, Field::srcStageMask), "VUID-VkSubpassDependency-srcStageMask-03937"},
-         {Key(Struct::VkSubpassDependency, Field::dstStageMask), "VUID-VkSubpassDependency-dstStageMask-03937"},
-         {Key(Struct::VkSubpassDependency2, Field::srcStageMask), "VUID-VkSubpassDependency2-srcStageMask-03937"},
-         {Key(Struct::VkSubpassDependency2, Field::dstStageMask), "VUID-VkSubpassDependency2-dstStageMask-03937"},
-     }},
 };
 
-const std::string &GetBadFeatureVUID(const Location &loc, VkPipelineStageFlags2KHR bit) {
+static const std::vector<Entry> kStageMaskErrorsNoneWithSync2{
```
juan-lunarg, line 271: Could these new globals be converted to `std::array` instead?

PR #5517, opened 2023-03-27
tests: Fix more tests for Max Profiles

spencer-lunarg:
Adds a few more tests to run properly on Max Profiles while still being able to run on my local machine

This the last of the low-hanging fruit for Max Profiles, the rest of the tests will need more work in order to work

### DIFF
--- a/layers/stateless/parameter_validation_utils.cpp
+++ b/layers/stateless/parameter_validation_utils.cpp
@@ -3450,6 +3450,7 @@ bool StatelessValidation::manual_PreCallValidateCreateGraphicsPipelines(VkDevice
                                             i, attachment_index, string_VkBlendOp(attachment_state.colorBlendOp),
                                             color_attachment_count,
                                             phys_dev_ext_props.blend_operation_advanced_props.advancedBlendMaxColorAttachments);
+                                        break;  // if this fails once, will fail every iteration
                                     }
                                 }
                             }
--- a/tests/device_profiles/max_profile.json
+++ b/tests/device_profiles/max_profile.json
@@ -1055,11 +1055,11 @@
                     ]
                 },
                 "VkPhysicalDeviceBlendOperationAdvancedPropertiesEXT": {
-                    "advancedBlendMaxColorAttachments": 4294967000,
-                    "advancedBlendIndependentBlend": true,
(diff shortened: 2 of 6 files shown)


tests/device_profiles/max_profile.json
```diff
@@ -2074,8 +2072,7 @@
                             "VK_FORMAT_FEATURE_2_STORAGE_READ_WITHOUT_FORMAT_BIT",
                             "VK_FORMAT_FEATURE_2_STORAGE_WRITE_WITHOUT_FORMAT_BIT",
                             "VK_FORMAT_FEATURE_2_SAMPLED_IMAGE_DEPTH_COMPARISON_BIT",
-                            "VK_FORMAT_FEATURE_2_FRAGMENT_DENSITY_MAP_BIT_EXT",
-                            "VK_FORMAT_FEATURE_2_FRAGMENT_SHADING_RATE_ATTACHMENT_BIT_KHR"
+                            "VK_FORMAT_FEATURE_2_FRAGMENT_DENSITY_MAP_BIT_EXT"
```
artem-lunarg, line 2075: This feature is probably only for images, should still be here? I remember I made similar comment some time ago, but don't remember the resolution now.

spencer-lunarg: I think it could honestly be removed from much more... I guess since it got accidentally added at the start, wasn't sure the real benefits of taking them all out now

artem-lunarg (approved):

layers/stateless/parameter_validation_utils.cpp
```diff
@@ -3450,6 +3450,7 @@ bool StatelessValidation::manual_PreCallValidateCreateGraphicsPipelines(VkDevice
                                             i, attachment_index, string_VkBlendOp(attachment_state.colorBlendOp),
                                             color_attachment_count,
                                             phys_dev_ext_props.blend_operation_advanced_props.advancedBlendMaxColorAttachments);
+                                        break;  // if this fails once, will fail every iteration
```
juan-lunarg, line 3453: My lord this function is huge. If it wasn't for this `break` I wouldn't have been aware of the loop.

juan-lunarg (approved):

PR #5523, opened 2023-03-27
cmake: Add VVL_CODEGEN

juan-lunarg:
Code generation should be OFF by default. It's not needed to build the vulkan validation layers.

- It's causing issues for add_subdirectory users.
- Package managers have no good way to disable this aside from patching.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -15,6 +15,7 @@
         "BUILD_TESTS": "ON",
         "BUILD_WERROR": "ON",
         "UPDATE_DEPS": "ON",
+        "VVL_CODEGEN": "ON",
         "CMAKE_BUILD_TYPE": "Debug"
       }
     }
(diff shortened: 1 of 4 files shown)


ncesario-lunarg (approved):
LGTM, though I don't see any harm in adding the target unconditionally.

juan-lunarg:
> LGTM, though I don't see any harm in adding the target unconditionally.

It's a VERY similar rational to `BUILD_WERROR` being `OFF` by default. You have to click on those links for the full context.

The more I learn about  open source development is that you should do the absolute minimum in your build system to avoid problems with the community building your stuff.

mikes-lunarg (approved):
Makes sense, I assume the `FATAL_ERROR`s are what is messing with downstream builds. Originally this target was better about gracefully degrading.

PR #5524, opened 2023-03-27
layers: Add vkGetMemoryFdKHR VUs

artem-lunarg:
Part of https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5431

This PR validates `VkMemoryGetFdInfoKHR` passed to `vkGetMemoryFdKHR`:
`VUID-VkMemoryGetFdInfoKHR-handleType-00671`
`VUID-VkMemoryGetFdInfoKHR-handleType-00672`

### DIFF
--- a/layers/core_checks/external_object_validation.cpp
+++ b/layers/core_checks/external_object_validation.cpp
@@ -21,6 +21,24 @@
 #include "chassis.h"
 #include "core_checks/core_validation.h"
 
+bool CoreChecks::PreCallValidateGetMemoryFdKHR(VkDevice device, const VkMemoryGetFdInfoKHR *pGetFdInfo, int *pFd) const {
+    bool skip = false;
+    if (const auto memory_state = Get<DEVICE_MEMORY_STATE>(pGetFdInfo->memory)) {
+        const auto export_info = LvlFindInChain<VkExportMemoryAllocateInfo>(memory_state->alloc_info.pNext);
+        if (!export_info) {
+            skip |= LogError(pGetFdInfo->memory, "VUID-VkMemoryGetFdInfoKHR-handleType-00671",
+                             "vkGetMemoryFdKHR(): memory's pNext chain does not include a VkExportMemoryAllocateInfo structure.");
+        } else if ((export_info->handleTypes & pGetFdInfo->handleType) == 0) {
+            skip |= LogError(pGetFdInfo->memory, "VUID-VkMemoryGetFdInfoKHR-handleType-00671",
+                             "vkGetMemoryFdKHR(): the requested handle type (%s) is not included in the memory's "
+                             "VkExportMemoryAllocateInfo::handleTypes (%s).",
+                             string_VkExternalMemoryHandleTypeFlagBits(pGetFdInfo->handleType),
+                             string_VkExternalMemoryHandleTypeFlags(export_info->handleTypes).c_str());
+        }
+    }
+    return skip;
+}
+
 bool CoreChecks::PreCallValidateImportSemaphoreFdKHR(VkDevice device, const VkImportSemaphoreFdInfoKHR *info) const {
     bool skip = false;
     const char *func_name = "vkImportSemaphoreFdKHR";
--- a/layers/stateless/parameter_validation_utils.cpp
+++ b/layers/stateless/parameter_validation_utils.cpp
@@ -7621,6 +7621,18 @@ bool StatelessValidation::manual_PreCallValidateCreateSamplerYcbcrConversionKHR(
                                                 "vkCreateSamplerYcbcrConversionKHR");
 }
 
+bool StatelessValidation::manual_PreCallValidateGetMemoryFdKHR(VkDevice device, const VkMemoryGetFdInfoKHR *pGetFdInfo,
+                                                               int *pFd) const {
+    constexpr auto allowed_types = VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD_BIT | VK_EXTERNAL_MEMORY_HANDLE_TYPE_DMA_BUF_BIT_EXT;
+    bool skip = false;
+    if (0 == (pGetFdInfo->handleType & allowed_types)) {
+        skip |= LogError(pGetFdInfo->memory, "VUID-VkMemoryGetFdInfoKHR-handleType-00672",
+                         "vkGetMemoryFdKHR(): handle type %s is not one of the supported handle types.",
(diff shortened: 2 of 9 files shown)


layers/core_checks/external_object_validation.cpp
```diff
@@ -21,6 +21,24 @@
 #include "chassis.h"
 #include "core_checks/core_validation.h"
 
+bool CoreChecks::PreCallValidateGetMemoryFdKHR(VkDevice device, const VkMemoryGetFdInfoKHR *pGetFdInfo, int *pFd) const {
+    bool skip = false;
+    if (const auto memory_state = Get<DEVICE_MEMORY_STATE>(pGetFdInfo->memory)) {
```
ncesario-lunarg, line 26: I vaguely recall this not working without an extra set of `()`; good to know this works :)

ncesario-lunarg (approved):
LGTM!

layers/core_checks/external_object_validation.cpp
artem-lunarg: I also was not 100% sure about this, but at least it passed CI. Hopefully it's according to spec.

artem-lunarg: Checked cpp reference, this use case should covered by the second rule:
> condition	-	one of
> * expression which is contextually convertible to bool
> * declaration of a single non-array variable with a brace-or-equals initializer.

tests/negative/external_memory_sync.cpp
juan-lunarg, line 1: Add a positive test as well.

layers/core_checks/external_object_validation.cpp
juan-lunarg, line 26: Return early if `memory_state` is false. Avoids indenting most of the logic.

ncesario-lunarg: Yeah, I think I'm thinking of pre-c++17, where the "declaration" could not be in the condition.

tests/negative/external_memory_sync.cpp
artem-lunarg: Is the suggestion to add one positive scenario to sequence of steps in GetMemoryFdHandle or a separate positive test?

layers/stateless/parameter_validation_utils.cpp
```diff
@@ -7621,6 +7621,19 @@ bool StatelessValidation::manual_PreCallValidateCreateSamplerYcbcrConversionKHR(
                                                 "vkCreateSamplerYcbcrConversionKHR");
 }
 
+bool StatelessValidation::manual_PreCallValidateGetMemoryFdKHR(VkDevice device, const VkMemoryGetFdInfoKHR *pGetFdInfo,
+                                                               int *pFd) const {
+    constexpr auto allowed_types = VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD_BIT | VK_EXTERNAL_MEMORY_HANDLE_TYPE_DMA_BUF_BIT_EXT;
+    bool skip = false;
+    if (0 == (pGetFdInfo->handleType & allowed_types)) {
+        skip |= LogError(pGetFdInfo->memory, "VUID-VkMemoryGetFdInfoKHR-handleType-00672",
+                         "vkGetMemoryFdKHR(): handle type %s is not one of the supported handleTypes (%s).",
```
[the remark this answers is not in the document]
artem-lunarg: Lets discuss this, since the error messages topic was during the meeting. I will mention @ncesario-lunarg and @spencer-lunarg here too in case they have opinions and to share the ideas.

My personal preference is that the part of the output that we write manually (lets call it the main part) should be a self sufficient error message, and VUID part can indeed be used to remove some information from the main part to make it more simple and focused, but the main part nevertheless should be a self sufficient error message. One idea for the future it to provide configurable forms of the output, and, for example, short form might be only the main part. Or even in the full message I think often people as the first step try to extract the main part, and it's good if it provides summary of what's going on. In this regard, it might be similar to conventions about the first sentence of the git message.

The error message of the form "Variable is equal X" is not an error message by itself. But I like the suggestion to remove the constants since they will be listed in the VUID text and to leave `vkGetMemoryFdKHR(): handle type %s is not one of the supported handle types`.

layers/core_checks/external_object_validation.cpp
artem-lunarg: The idea here is that this block is self sufficient block of validation, which is good when new parts of validation are added to the same API function, I think it gets more easy to add and remove parts of validation and makes such validation block more independent. I have impression that this style is more common in the code or maybe not?

tests/negative/external_memory_sync.cpp
juan-lunarg: Ideally a separate positive test.

artem-lunarg: This helped to detect android issue!

layers/stateless/parameter_validation_utils.cpp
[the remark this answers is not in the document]
juan-lunarg: The main thing I was trying to point out was the repetition of the supported handleTypes. Which you have already handled.

So I'd say the fix is good IMO.

juan-lunarg (changes requested):
Positive test is getting skipped on CI:

```
[ RUN      ] VkPositiveLayerTest.GetMemoryFdHandle
../tests/positive/external_memory_sync.cpp:28: Skipped
vkGetMemoryFdKHR not supported by MockICD
[  SKIPPED ] VkPositiveLayerTest.GetMemoryFdHandle (48 ms)
```

tests/negative/external_memory_sync.cpp
juan-lunarg: Does adding the extension to profiles not help?

juan-lunarg: Ah it's already a part of `max_profiles.json`

How tricky would it be to add the support to the mock driver?

artem-lunarg: The extension is there (`VK_KHR_external_memory_fd`). The problem we rely on the result from the API function (not VkResult, but some computed value), so this requires either driver or modification of the MockICD.

If the intention for the positive test is to ensure that we don't have false-positive error reporting (in contrast to testing API call), one option is to remove checking the value returned by `vkGetMemoryFdKHR`, actually I think it's more correct approach.

artem-lunarg: > How tricky would it be to add the support to the mock driver?
That shouldn't be very complex, incrementing counter might work. But I changed the test without checking results of the API call, it's enough to ensure we don't have false-positive validation errors.

Issue #5520, opened 2023-03-27
Array output causing UNASSIGNED-CoreValidation-Shader-NoAlphaAtLocation0WithAlphaToCoverage


romanl-g:
Hi,

We hit an issue in ANGLE where `glEnable(GL_SAMPLE_ALPHA_TO_COVERAGE);` together with `gl_FragData[0]` results in a validation failure:

```
UNASSIGNED-CoreValidation-Shader-NoAlphaAtLocation0WithAlphaToCoverage(ERROR / SPEC): msgNum: 621927094 - Validation Error: [ UNASSIGNED-CoreValidation-Shader-NoAlphaAtLocation0WithAlphaToCoverage ] Object 0: handle = 0x8190000000819, type = VK_OBJECT_TYPE_SHADER_MODULE; | MessageID = 0x2511dab6 | fragment shader doesn't declare alpha output at location 0 even though alpha to coverage is enabled.
```

ANGLE repro: https://chromium-review.googlesource.com/c/angle/angle/+/4374098

The relevant part of the SPIR-V:
```
%5 = OpTypeFloat 32
%10 = OpTypeVector %5 4
%11 = OpConstant %2 8
%12 = OpTypeArray %10 %11
%13 = OpTypePointer Output %12
%14 = OpVariable %13 Output
```

The "== 4" check here is probably checking for the alpha channel:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/9124f788b7de42a4b2589b7ede812dd887447f0c/layers/core_checks/shader_validation.cpp#L408

But we're actually getting 8*4=32 components there because gl_FragData is an array[8]
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/9124f788b7de42a4b2589b7ede812dd887447f0c/layers/state_tracker/shader_module.cpp#L799

Unsure if it's just `strip_array_level=false` that should be `true` or something else

spencer-lunarg:
Side note: Currently trying to fix this VU so it has a proper VU in the spec as not sure what it is suppose to be https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5811

The `GetComponentsConsumedByType` has some issue which I am trying to work on/fix in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5481

I would be happy removing this check for now until it is properly added in the spec AND the `GetComponentsConsumedByType` logic is correctly implemented

[...]

PR #5526, opened 2023-03-28
layers: Temporarily remove UNASSIGNED alphaToCoverage

spencer-lunarg:
As noted in https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5520

- `UNASSIGNED-CoreValidation-Shader-NoAlphaAtLocation0WithAlphaToCoverage` needs a proper VU
- The `GetComponentsConsumedByType` currently is being refactored

Instead of leaving this in broken or adding more duct tape to fix it, I would rather just remove this now. I will keep the #5520 issue open as a task to come back and add it once the two above issues are fixed


tests/negative/pipeline_shader.cpp
```diff
@@ -5708,7 +5708,8 @@ TEST_F(VkLayerTest, CreatePipelineFragmentOutputNotConsumed) {
     CreatePipelineHelper::OneshotTest(*this, set_info, kWarningBit, "UNASSIGNED-CoreValidation-Shader-OutputNotConsumed");
 }
 
-TEST_F(VkLayerTest, CreatePipelineFragmentNoOutputLocation0ButAlphaToCoverageEnabled) {
+// Currently need to clarify the VU - https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5520
+TEST_F(VkLayerTest, DISABLED_CreatePipelineFragmentNoOutputLocation0ButAlphaToCoverageEnabled) {
```
juan-lunarg, line 5712: Is prefixing a test with "DISABLED_" proper?

spencer-lunarg: Yes, we have 4 other tests with it, I think better to leave in, makes adding it easier if we change some fundamental way to do tests, or want to move to another file, etc

The tests (probably 95% of it) WILL be re-enabled

PR #5530, opened 2023-03-29
tests: Forbid copy/move of timeout guard

artem-lunarg:
Based on this comment: https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/0f6867787052243586afeadeb9c3b6b8cff6638a#r106221779

Plus added a comment why original code does not cause issues when returning local object with non-trivial destructor.


spencer-lunarg (approved):

ben-clayton (approved):
Thanks!

juan-lunarg (approved):

PR #5531, opened 2023-03-29
layers: Fix VkLayerTest.ViewportSwizzleNV

juan-lunarg:
Update validation for viewportCount-01215

Instead of checking for `not equal` it should be `less than`

### DIFF
--- a/layers/stateless/parameter_validation_utils.cpp
+++ b/layers/stateless/parameter_validation_utils.cpp
@@ -2983,12 +2983,13 @@ bool StatelessValidation::manual_PreCallValidateCreateGraphicsPipelines(VkDevice
                     }
 
                     if (vp_swizzle_struct) {
-                        if (vp_swizzle_struct->viewportCount != viewport_state.viewportCount) {
+                        const auto swizzle_viewport_count = vp_swizzle_struct->viewportCount;
+                        const auto viewport_count = viewport_state.viewportCount;
+                        if (swizzle_viewport_count < viewport_count) {
                             skip |= LogError(device, "VUID-VkPipelineViewportSwizzleStateCreateInfoNV-viewportCount-01215",
-                                             "vkCreateGraphicsPipelines: The viewport swizzle state vieport count of %" PRIu32
-                                             " does "
-                                             "not match the viewport count of %" PRIu32 " in VkPipelineViewportStateCreateInfo.",
-                                             vp_swizzle_struct->viewportCount, viewport_state.viewportCount);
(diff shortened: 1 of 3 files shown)


tests/negative/pipeline_shader.cpp
```diff
@@ -5535,60 +5535,56 @@ TEST_F(VkLayerTest, ViewportSwizzleNV) {
...
         };
 
-        if (!test_case.positive) {
-            CreatePipelineHelper::OneshotTest(*this, break_vp_count, kErrorBit,
-                                              "VUID-VkPipelineViewportSwizzleStateCreateInfoNV-viewportCount-01215");
-        } else {
-            CreatePipelineHelper::OneshotTest(*this, break_vp_count, kErrorBit);
-        }
+        CreatePipelineHelper::OneshotTest(*this, break_vp_count, kErrorBit,
+                                          "VUID-VkPipelineViewportSwizzleStateCreateInfoNV-viewportCount-01215");
```
juan-lunarg, line 5587: Error message provided to user:

> Validation Error: [ VUID-VkPipelineViewportSwizzleStateCreateInfoNV-viewportCount-01215 ] Object 0: handle = 0x555557c08a80, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x1be71841 | vkCreateGraphicsPipelines: SwizzleState viewportCount (1) less than ViewportState viewportCount (2)! The Vulkan spec states: viewportCount must be greater than or equal to the viewportCount set in VkPipelineViewportStateCreateInfo (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkPipelineViewportSwizzleStateCreateInfoNV-viewportCount-01215)

layers/stateless/parameter_validation_utils.cpp
```diff
@@ -2983,12 +2983,13 @@ bool StatelessValidation::manual_PreCallValidateCreateGraphicsPipelines(VkDevice
                     }
 
                     if (vp_swizzle_struct) {
-                        if (vp_swizzle_struct->viewportCount != viewport_state.viewportCount) {
+                        const auto swizzle_viewport_count = vp_swizzle_struct->viewportCount;
+                        const auto viewport_count = viewport_state.viewportCount;
+                        if (swizzle_viewport_count < viewport_count) {
```
juan-lunarg, line 2988: `viewportCount must be greater than or equal to the viewportCount set in VkPipelineViewportStateCreateInfo`

https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPipelineViewportSwizzleStateCreateInfoNV.html

spencer-lunarg (approved):

PR #5537, opened 2023-03-30
tests: Add Profiles VkFormatProperties3 regression

spencer-lunarg:
Found bug in [Profiles layer](https://github.com/KhronosGroup/Vulkan-Profiles/pull/392) that was causing a second call with `VkFormatProperties3` to return `VUID-VkFormatProperties2-pNext-pNext` because the `sType` was being overwritten

### DIFF
--- a/tests/positive/other.cpp
+++ b/tests/positive/other.cpp
@@ -998,3 +998,19 @@ TEST_F(VkPositiveLayerTest, SafeVoidPointerCopies) {
         ASSERT_TRUE(copied_bytes[10] == std::byte{0b01001001});
     }
 }
+
+TEST_F(VkPositiveLayerTest, FormatProperties3FromProfiles) {
+    // https://github.com/KhronosGroup/Vulkan-Profiles/pull/392
+    TEST_DESCRIPTION("Make sure VkFormatProperties3KHR is overwritten correctly in Profiles layer");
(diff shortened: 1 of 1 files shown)


tests/positive/other.cpp
arno-lunarg, line 1: LGTM! I just think the test class should be `VkPositiveLayerTest`

arno-lunarg (approved):

tests/positive/other.cpp
```diff
@@ -997,4 +997,20 @@ TEST_F(VkPositiveLayerTest, SafeVoidPointerCopies) {
         auto copied_bytes = reinterpret_cast<const std::byte *>(safe.pData);
         ASSERT_TRUE(copied_bytes[10] == std::byte{0b01001001});
     }
+}
+
+TEST_F(VkPositiveLayerTest, FormatProperties3FromProfiles) {
+    // https://github.com/KhronosGroup/Vulkan-Profiles/pull/392
+    TEST_DESCRIPTION("Makes VkFormatProperties3KHR is overwritten correctly in Profiles");
```
juan-lunarg, line 1004:
```suggestion
    TEST_DESCRIPTION("Make sure VkFormatProperties3KHR is overwritten correctly in Profiles layer");
```

juan-lunarg (approved):

PR #5544, opened 2023-03-31
tests: Use Binding Handle to cleanup itself

spencer-lunarg:
Was inspired to just fix all the tests (where possible) to use the `vk_testing::` helper that way we don't need to worry about `vk::Destroy*`

A lot of tests are just copy-pastes of each other, so good to fix it to prevent the bad practice from spreading

### DIFF
--- a/tests/negative/atomics.cpp
+++ b/tests/negative/atomics.cpp
@@ -1060,9 +1060,7 @@ TEST_F(VkLayerTest, InvalidStorageAtomicOperation) {
     image.Init(image_ci);
     VkImageView image_view = image.targetView(image_format);
 
-    VkSampler sampler = VK_NULL_HANDLE;
-    VkSamplerCreateInfo sampler_info = SafeSaneSamplerCreateInfo();
-    vk::CreateSampler(m_device->device(), &sampler_info, NULL, &sampler);
+    vk_testing::Sampler sampler(*m_device, SafeSaneSamplerCreateInfo());
 
     VkBufferObj buffer;
     buffer.init(*m_device, 64, 0, VK_BUFFER_USAGE_STORAGE_TEXEL_BUFFER_BIT);
@@ -1071,8 +1069,7 @@ TEST_F(VkLayerTest, InvalidStorageAtomicOperation) {
     bvci.buffer = buffer.handle();
     bvci.format = buffer_view_format;
     bvci.range = VK_WHOLE_SIZE;
-    VkBufferView buffer_view;
-    vk::CreateBufferView(m_device->device(), &bvci, NULL, &buffer_view);
+    vk_testing::BufferView buffer_view(*m_device, bvci);
 
     char const *fsSource = R"glsl(
         #version 450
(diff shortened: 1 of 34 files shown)


tests/negative/atomics.cpp
```diff
@@ -1103,12 +1100,12 @@ TEST_F(VkLayerTest, InvalidStorageAtomicOperation) {
...
 
-    g_pipe.descriptor_set_->WriteDescriptorImageInfo(3, image_view, sampler, VK_DESCRIPTOR_TYPE_STORAGE_IMAGE,
+    g_pipe.descriptor_set_->WriteDescriptorImageInfo(3, image_view, sampler.handle(), VK_DESCRIPTOR_TYPE_STORAGE_IMAGE,
                                                      VK_IMAGE_LAYOUT_GENERAL);
-    g_pipe.descriptor_set_->WriteDescriptorImageInfo(2, image_view, sampler, VK_DESCRIPTOR_TYPE_STORAGE_IMAGE,
+    g_pipe.descriptor_set_->WriteDescriptorImageInfo(2, image_view, sampler.handle(), VK_DESCRIPTOR_TYPE_STORAGE_IMAGE,
                                                      VK_IMAGE_LAYOUT_GENERAL, 0, 2);
-    g_pipe.descriptor_set_->WriteDescriptorBufferView(1, buffer_view);
-    g_pipe.descriptor_set_->WriteDescriptorBufferView(0, buffer_view, VK_DESCRIPTOR_TYPE_STORAGE_TEXEL_BUFFER, 0, 2);
+    g_pipe.descriptor_set_->WriteDescriptorBufferView(1, buffer_view.handle());
```
artem-lunarg, line 1107: Can we rely on implicit conversion instead of calling `handle()` ?

spencer-lunarg: .... seems we can... if others agree, will go through and remove them all

arno-lunarg: _They call him the VVL janitor_

spencer-lunarg: well I guess what is the point of the `handle()` if we can do it implicitly

artem-lunarg: I don't have a good example, but maybe some C++ rules that require exact types  (templates?) and where implicit conversion is not happening.

juan-lunarg: I'm not a fan of implicit conversions personally and prefer using `handle()`

juan-lunarg (approved):

tests/negative/atomics.cpp
artem-lunarg: Added my arguments and motivation (for implicit `handle` in VVL project). But the most important is probably to come up with a single style. I'm fine with explicit `handle()` if we decide on it. There might be ping-pong game when people with different styles change the same code (and we haves tons of code with handles). It's even worse than having different styles in the same codebase because the code gets less stable due to constant modifications. 

Personally for me, _making code more explicit_ is one of the most important principles to produce more maintainable and robust software. I'm quite extreme here in my personal style (rather distinct function names than overloading, avoid constructs with unclear codegen, minimize usage of non trivial ctors/dtors due their implicit nature, avoid many constructors in the same type because they do not differentiate by name, and in my personal code for some Vulkan wrappers I also have a  `handle` struct field and _do not_ provide implicit type conversion). 

Having that said, I realize that my personal style is usually considered not a modern C++ and VVL codebase is a better example of it, there are areas in C++ where implicit conversions are considered a good tradeoff between being being explicit on one side and convenience or automatic enforcement on another side. Here are some C++ examples with implicit aspects that are considered a norm (intermixed with some thoughts):

* non-explicit constructors to create `std::string_view` from `std::string` or from `const char*`.
* implicit invocation of destructors
* function overloading (the choice is made by the compiler, sometimes not what you expected). One example, is that without function overloading it's hard to implement custom operators. Without custom operators for some people it's inconvenient to use linear math libraries. One example is expression `Matrix m = m1 + m2  + m3 + m4` which will be a chain of nested function calls. This problem arises in `Zig` programming language, which is very nice and even more explicit, but people complain about ugly math code. There are some point targeted proposals to relax conventions only for math-like scenario by introducing an implicit element with restrict area of influence.
* More similar to our use case, some types are designed to be wrappers for other types (infinite arithmetic, objects that monitor other objects, etc.)  that represent similar or identical semantics. With non-explicit constructors and operator Type(),  C++ allows to use them mostly interchangeably, if that's desired. It might be something that is not desired and you just don't go with that approach but if that approach it already chosen and some aspects of the object are handled in an implicit way (construction/destruction) then type conversion (without perf hit in our case, the same codegen) might be change in the same direction.
* One example where implicit is not good. Implicit conversion between builtin numeric types. That's source of bugs because types have different representation and data ranges and loss of data might happen, that's why compile warnings. It's not our scenario, we have same the data (handle value) and similar or the same semantics.

jzulauf-lunarg:
My preference would be to not have the implicit conversion to the handle type.

PR #5545, opened 2023-03-31
docs: What is a good error message

spencer-lunarg:
Add example of writing a good error message

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -95,6 +95,26 @@ That said, please ensure that the repository compiles and passes tests without e
 that to be accepted into the repository, the pull request must [pass all tests](#testing your changes) on all supported platforms
 -- the continuous integration features will assist in enforcing this requirement.
 
+### **Writing good error messages**
+
+When writing an error message for `LogError` it is important to
+
+1. Print values related to the error message
+2. Explain the logic that got to that error
+
+Example of a good error message
+
+```cpp
+if (render_pass == VK_NULL_HANDLE) {
+    // ...
+} else if (value != 0 && HasDepthFlag(flag)) {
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -95,6 +95,26 @@ That said, please ensure that the repository compiles and passes tests without e
...
+
+1. Print values related to the error message
+2. Explain the logic that got to that error
+
+Example of a good error message
+
+```cpp
+if (render_pass == VK_NULL_HANDLE) {
+    // ...
+} else if (value != 0 && HasDepthFlag(flag)) {
```
arno-lunarg, line 110:
```suggestion
} else if (value == 0 && HasDepthFlag(flag)) {
```

CONTRIBUTING.md
```diff
@@ -95,6 +95,26 @@ That said, please ensure that the repository compiles and passes tests without e
...
+    // print Render Pass object
+    // Value is given value, no need to bring
+    // list flag users used
+    skip |= LogError(render_pass, "value is 0 but flag (%s) is missing VK_FLAG_DEPTH.", string_VkFlag(flag));
```
arno-lunarg, line 114:
```suggestion
    // Value is not the expected one, log it
    // List flag users used
    skip |= LogError(render_pass, "value is %" PRIu32 " but flag (%s) is missing VK_FLAG_DEPTH.", value, string_VkFlag(flag));
```
To have an example of using PRIu32, I personally did not know about that before working on the VVL

arno-lunarg (approved):
LGTM! And a nice addition I think : )
One suggestion

juan-lunarg (approved):

PR #5556, opened 2023-04-01
layers: Breakup Stateless validation

spencer-lunarg:
Takes the 10,000 line `parameter_validation_utils.cpp` and breaks up into files to better represent the validation


layers/stateless/image_buffer_validation.cpp
```diff
@@ -0,0 +1,794 @@
+/* Copyright (c) 2015-2023 The Khronos Group Inc.
```
arno-lunarg, line 1: Would it make sense to have both `stateless/image_validation.cpp` and `stateless/buffer_validation.cpp`?

layers/stateless/instance_device_validation.cpp
```diff
@@ -0,0 +1,866 @@
+/* Copyright (c) 2015-2023 The Khronos Group Inc.
```
arno-lunarg, line 1: Same question here with `stateless/instance_validation.cpp` and `stateless_device_validation.cpp`

layers/stateless/render_pass_validation.cpp
```diff
@@ -0,0 +1,521 @@
...
+}
+
+void StatelessValidation::PostCallRecordDestroyRenderPass(VkDevice device, VkRenderPass renderPass,
+                                                          const VkAllocationCallbacks *pAllocator) {
+    // Track the state necessary for checking vkCreateGraphicsPipeline (subpass usage of depth and color attachments)
+    std::unique_lock<std::mutex> lock(renderpass_map_mutex);
+    renderpasses_states.erase(renderPass);
+}
+
+bool StatelessValidation::manual_PreCallValidateCreateFramebuffer(VkDevice device, const VkFramebufferCreateInfo *pCreateInfo,
```
arno-lunarg, line 511: What about a dedicated `stateless/framebuffer_validation.cpp`?

layers/stateless/image_buffer_validation.cpp
spencer-lunarg: I guess I saw the `buffer_validation` as 2 functions and only like 80 lines of code... but honestly that is how we got into these 10k line files in the first place... will split

layers/stateless/instance_device_validation.cpp
spencer-lunarg: this one is harder because some of these helper functions start to span both

ncesario-lunarg (approved):

PR #5557, opened 2023-04-02
IWYU: add cstdio for snprintf in vk_mem_alloc.h

stha09:
Discovered with GCC 13.

### DIFF
--- a/layers/external/vma/vk_mem_alloc.h
+++ b/layers/external/vma/vk_mem_alloc.h
@@ -2579,6 +2579,7 @@ VMA_CALL_PRE void VMA_CALL_POST vmaFreeStatsString(
 #undef VMA_IMPLEMENTATION
 
 #include <cstdint>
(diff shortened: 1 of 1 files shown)


spencer-lunarg:
@stha09 so this is an external file we use, if you want to make the fix, it should be done https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/blob/master/include/vk_mem_alloc.h#L2604

then we will absorb it when we update

stha09:
> @stha09 so this is an external file we use, if you want to make the fix, it should be done https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/blob/master/include/vk_mem_alloc.h#L2604
> 
> then we will absorb it when we update

I opened a PR there: https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/pull/326

PR #5562, opened 2023-04-03
docs: Remove implication of 18.04 build support

### LINKED ISSUE #5473: Build fails because of a compilation error.
**Environment:**
 - OS: Ubuntu 18.04
 - SDK or header version if building from repo: main

**Describe the Issue**

Build fails with the message:

```
Vulkan-ValidationLayers/layers/./utils/vk_layer_utils.h: In function ‘constexpr VkExtent3D GetEffectiveExtent(const VkImageCreateInfo&, VkImageAspectFlags, uint32_t)’:
Vulkan-ValidationLayers/layers/./utils/vk_layer_utils.h:415:27: error: call to non-constexpr function ‘bool FormatIsMultiplane(VkFormat)’
     if (FormatIsMultiplane(format)) {
```

juan-lunarg:
Our code doesn't compile with the version of GCC that comes with Ubuntu 18.04 by default.

closes #5473

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -234,11 +234,7 @@ cmake --build . --config Release
 
 ### Linux Build Requirements
 
-This repository has been built and tested on the two most recent Ubuntu LTS
-versions. Currently, the oldest supported version is Ubuntu 18.04, meaning
-that the minimum officially supported C++17 compiler version is GCC 7.3.0,
-although earlier versions may work. It should be straightforward to adapt this
-repository to other Linux distributions.
+This repository is regularly built and tested on the two most recent Ubuntu LTS versions.
 
 [CMake 3.17.2](https://cmake.org/files/v3.17/cmake-3.17.2-Linux-x86_64.tar.gz) is recommended.
 


juan-lunarg:
After looking into this and fixing various issues I hit a dead end fixing support for `18.04` builds due to incomplete C++17 support from GCC [gcc-7](https://packages.ubuntu.com/bionic/gcc-7).

Users can still build on `18.04` however, they can't use the default packages the come with the distro. They would have to setup their own build environment.

BUILD.md
```diff
@@ -234,11 +234,7 @@ cmake --build . --config Release
 
 ### Linux Build Requirements
 
-This repository has been built and tested on the two most recent Ubuntu LTS
-versions. Currently, the oldest supported version is Ubuntu 18.04, meaning
-that the minimum officially supported C++17 compiler version is GCC 7.3.0,
-although earlier versions may work. It should be straightforward to adapt this
-repository to other Linux distributions.
+This repository has been built and tested on the two most recent Ubuntu LTS versions.
```
spencer-lunarg, line 237: > Users can still build on 18.04 however, they can't use the default packages the come with the distro. They would have to setup their own build environment.

This information could be added in some extent

spencer-lunarg:
```suggestion
This repository is built and tested on the two most recent Ubuntu LTS versions.
```

to make it sound like we will on-going keep doing this

juan-lunarg: > This information could be added in some extent

I can't think of way to add that information that wouldn't age badly or require a lot of explanation.

ncesario-lunarg (approved):

juan-lunarg:
It's also worth noting that the `18.04` runner provided by github is officially deprecated as of yesterday:

https://github.com/actions/runner-images/issues/6002

PR #5565, opened 2023-04-03
bp: Add best practice for AMD workgroup size

juan-lunarg:
https://gpuopen.com/learn/rdna-performance-guide/#shaders
"Make the workgroup size a multiple of 64 to obtain best performance across all GPU generations."

### DIFF
--- a/layers/best_practices/best_practices_utils.cpp
+++ b/layers/best_practices/best_practices_utils.cpp
@@ -1368,10 +1368,12 @@ bool BestPractices::ValidateCreateComputePipelineArm(const VkComputePipelineCrea
     if (!entrypoint_optional) return false;
 
     const Instruction& entrypoint = *entrypoint_optional;
-    uint32_t x = 1, y = 1, z = 1;
-    module_state->FindLocalSize(entrypoint, x, y, z);
+    uint32_t x = {}, y = {}, z = {};
+    if (!module_state->FindLocalSize(entrypoint, x, y, z)) {
+        return false;
+    }
 
-    uint32_t thread_count = x * y * z;
+    const uint32_t thread_count = x * y * z;
 
     // Generate a priori warnings about work group sizes.
     if (thread_count > kMaxEfficientWorkGroupThreadCountArm) {
--- a/layers/best_practices/best_practices_utils.cpp
+++ b/layers/best_practices/best_practices_utils.cpp
@@ -1341,6 +1341,10 @@ bool BestPractices::PreCallValidateCreateComputePipelines(VkDevice device, VkPip
             skip |= ValidateCreateComputePipelineArm(createInfo);
         }
 
+        if (VendorCheckEnabled(kBPVendorAMD)) {
+            skip |= ValidateCreateComputePipelineAmd(createInfo);
+        }
+
         if (IsExtEnabled(device_extensions.vk_khr_maintenance4)) {
             auto module_state = Get<SHADER_MODULE_STATE>(createInfo.stage.module);
             if (module_state &&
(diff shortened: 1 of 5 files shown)


tests/negative/amd_best_practices.cpp
```diff
@@ -789,3 +789,57 @@ TEST_F(VkAmdBestPracticesLayerTest, SecondaryCmdBuffer) {
 
     m_errorMonitor->VerifyFound();
 }
+
+TEST_F(VkArmBestPracticesLayerTest, ComputeWorkgroupSize) {
```
artem-lunarg, line 793: Arm -> Amd

juan-lunarg: Good catch.

artem-lunarg (approved):
LGTM!

artem-lunarg:
One comment that sometimes people use with such small workgroup sizes,  (1,1,1) is a typical use case to do some initialization or other scalar operation when there is no other good choice. But I assume that's what is expected from the best practices. Mostly thinking how critical false-positives are. For big AAA game it's high chance there are such small compute dispatches.

juan-lunarg:
IMO best practices are similar to compiler warnings. Often times there isn't actually an issue. However, when there is an issue you are very grateful for the warning.

I'd say the same applies here.

juan-lunarg:
Note: I've reverted the new BP check due to confusion around the performance suggestion. I'm currently trying to get clarification from an AMD engineer.

juan-lunarg:
The main confusion is which workgroup the performance guide is referring to.

In the RDNA performance guide it says to "Make the workgroup size a multiple of 64 to obtain best performance across all GPU generations." for compute shaders.

What is workgroup size referring to here?

Is it referring to the local workgroup defined by the shader or the global workgroup defined by vkCmdDispatch and vkCmdDispatchIndirect commands?

It makes most sense to me that it's talking about the local workgroup. But I want to double check.

ncesario-lunarg (approved):

artem-lunarg:
> It makes most sense to me that it's talking about the local workgroup.
That's also my understanding, workgroup size is usually used in this context. Shader defined workgroup size correlates with wavefront size (32 or 64). Dispatch grid size is not affected by wavefront size.

juan-lunarg:
> That's also my understanding, workgroup size is usually used in this context. Shader defined workgroup size correlates with wavefront size (32 or 64). Dispatch grid size is not affected by wavefront size.

After double checking everything I've added back the BP check. The performance guide is most definitely refering to the `local workgroup`.

- https://computergraphics.stackexchange.com/questions/12462/compute-shader-workgroups-execution-and-size
- https://en.wikipedia.org/wiki/RDNA_(microarchitecture)#Differences_between_GCN_and_RDNA
- https://registry.khronos.org/vulkan/specs/1.3/html/chap9.html#shaders-compute

layers/best_practices/best_practices_utils.cpp
```diff
@@ -1429,6 +1435,39 @@ bool BestPractices::ValidateCreateComputePipelineArm(const VkComputePipelineCrea
...
+        return false;
+    }
+
+    const Instruction& entrypoint = *entrypoint_optional;
+    uint32_t x = {}, y = {}, z = {};
+    if (!module_state->FindLocalSize(entrypoint, x, y, z)) {
+        return false;
+    }
+
+    const uint32_t thread_count = x * y * z;
```
spencer-lunarg, line 1455: It would be nice to move this and`ValidateCreateComputePipelineArm` into a common function to grab this info

Also if `module_state.HasSpecConstants()` is true, this might produce a false result, but that is something outside the scope of this PR to fix

PR #5567, opened 2023-04-03
syncval: Clean up and correct bound size code.

jzulauf-lunarg:
Get rid of magic UINT32 max usage, and correctly set size for bind vertex buffer2

### DIFF
--- a/layers/state_tracker/buffer_state.h
+++ b/layers/state_tracker/buffer_state.h
@@ -40,6 +40,17 @@ class BUFFER_STATE : public BINDABLE {
     BUFFER_STATE(BUFFER_STATE const &rh_obj) = delete;
 
     VkBuffer buffer() const { return handle_.Cast<VkBuffer>(); }
+    std::optional<VkDeviceSize> ComputeValidSize(VkDeviceSize offset, VkDeviceSize size) const {
+        return ::ComputeValidSize(offset, size, createInfo.size);
+    }
+    VkDeviceSize ComputeSize(VkDeviceSize offset, VkDeviceSize size) const {
+        std::optional<VkDeviceSize> valid_size = ComputeValidSize(offset, size);
+        return valid_size.has_value() ? *valid_size : VkDeviceSize(0);
+    }
+    static VkDeviceSize ComputeSize(const std::shared_ptr<const BUFFER_STATE> &buffer_state, VkDeviceSize offset,
+                                    VkDeviceSize size) {
+        return buffer_state ? buffer_state->ComputeSize(offset, size) : VkDeviceSize(0);
+    }
 
     sparse_container::range<VkDeviceAddress> DeviceAddressRange() const { return {deviceAddress, deviceAddress + createInfo.size}; }
 };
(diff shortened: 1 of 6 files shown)


ncesario-lunarg (approved):

layers/state_tracker/buffer_state.h
```diff
@@ -40,6 +40,20 @@ class BUFFER_STATE : public BINDABLE {
     BUFFER_STATE(BUFFER_STATE const &rh_obj) = delete;
 
     VkBuffer buffer() const { return handle_.Cast<VkBuffer>(); }
+    std::optional<VkDeviceSize> ComputeValidSize(VkDeviceSize offset, VkDeviceSize size) const {
```
spencer-lunarg, line 43: why is this function needed and not just directly call the `::ComputeValidSize`

layers/state_tracker/buffer_state.h
```diff
@@ -40,6 +40,20 @@ class BUFFER_STATE : public BINDABLE {
...
+    }
+    VkDeviceSize ComputeSize(VkDeviceSize offset, VkDeviceSize size) const {
+        std::optional<VkDeviceSize> valid_size = ComputeValidSize(offset, size);
+        return valid_size.has_value() ? *valid_size : VkDeviceSize(0);
+    }
+    static VkDeviceSize ComputeSize(const std::shared_ptr<const BUFFER_STATE> &buffer_state, VkDeviceSize offset,
+                                    VkDeviceSize size) {
+        return buffer_state ? buffer_state->ComputeSize(offset, size) : VkDeviceSize(0);
+    }
+    static VkDeviceSize ComputeWholeSize(const std::shared_ptr<const BUFFER_STATE> &buffer_state, VkDeviceSize offset) {
```
spencer-lunarg, line 54: I think just passing in `VK_WHOLE_SIZE` into `ComputeSize` would be more intuitive

Without looking, trying to figure out the difference between`ComputeWholeSize` vs `ComputeSize` was a bit misleading... "Aren't they both getting the whole size of what I pass in?"

layers/state_tracker/buffer_state.h
jzulauf-lunarg: Assuming someone might want to track success as this gets adopted.  SyncVal assumes the rest of val is happy.

layers/sync/sync_validation.cpp
```diff
@@ -421,13 +410,27 @@ static ResourceAccessRange MakeRange(const T &has_offset_and_size) {
...
 static inline ResourceAccessRange MakeRange(const BUFFER_STATE &buffer, VkDeviceSize offset, VkDeviceSize size) {
-    return MakeRange(offset, GetBufferWholeSize(buffer, offset, size));
+    return MakeRange(offset, buffer.ComputeSize(offset, size));
 }
 
 static inline ResourceAccessRange MakeRange(const BUFFER_VIEW_STATE &buf_view_state) {
     return MakeRange(*buf_view_state.buffer_state.get(), buf_view_state.create_info.offset, buf_view_state.create_info.range);
 }
 
+ResourceAccessRange MakeRange_stride_impl(VkDeviceSize offset, uint32_t first_index, uint32_t count, uint32_t stride) {
```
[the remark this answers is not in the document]
jzulauf-lunarg: Cleaned.

jzulauf-lunarg (commented):
good input

PR #5570, opened 2023-04-05
Compatible validation for VK_EXT_shader_object

daniel-story:
This PR introduces a number of changes needed to allow validation of applications that use VK_EXT_shader_object without the layer crashing or generating false positives:

ValidateExtendedDynamicState() and its callers are modified to check for either the specific EDS1/2/3/etc. feature originally associated with each function, or the shaderObject feature (which also makes it legal to call them, even when the corresponding dynamic state feature is disabled). This allows these commands to be called with only the shaderObject feature enabled, as expected by the spec.

ValidateCmdBufDrawState() is temporarily modified to early out if the shaderObject feature is enabled but there isn't a pipeline bound. This avoids a false positive error when shader objects have been bound instead of pipelines.

The autogeneration code for the vk_extension_helper.h and vk_dispatch_table_helper.h files is modified to support a new ExtEnabled enum value: kEnabledByInteraction. This is a workaround for a limitation in the autogeneration architecture where types (including functions) that are introduced into the API by more than one extension can only be attributed to only one of the extensions at a time. Without this fix, vkGetInstanceProcAddr()/vkGetDeviceProcAddr() incorrectly return nullptr whenever applications query for EDS1/2/3/etc. entrypoints with only VK_EXT_shader_object enabled, which is intended to work. Also, many false positives would be generated when calling those entrypoints.

A missing null check in CMD_BUFFER_STATE::UpdatePipelineState() is added in order to avoid a crash when no pipeline is bound.

Finally, the logic in thread_safety_generator.py is modified to prevent inappropriate calls to StartReadObject() and CreateObject() in the autogenerated code for vkCreateShadersEXT().

With all these changes, shader object applications can run and make use of the functionality of the extension without the layer crashing or reporting false positives.

Full validation will come in the future.

### DIFF
--- a/layers/core_checks/cmd_buffer_dynamic_validation.cpp
+++ b/layers/core_checks/cmd_buffer_dynamic_validation.cpp
@@ -406,7 +406,7 @@ bool CoreChecks::ForbidInheritedViewportScissor(const CMD_BUFFER_STATE &cb_state
 
 // Used for all vkCmdSet* functions
 // Some calls are behind a feature bit that needs to be enabled
-bool CoreChecks::ValidateExtendedDynamicState(const CMD_BUFFER_STATE &cb_state, const CMD_TYPE cmd_type, VkBool32 feature,
+bool CoreChecks::ValidateExtendedDynamicState(const CMD_BUFFER_STATE &cb_state, const CMD_TYPE cmd_type, bool feature,
                                               const char *vuid, const char *feature_name) const {
     bool skip = false;
     skip |= ValidateCmd(cb_state, cmd_type);
@@ -597,8 +597,9 @@ bool CoreChecks::PreCallValidateCmdSetCheckpointNV(VkCommandBuffer commandBuffer
 bool CoreChecks::PreCallValidateCmdSetLogicOpEXT(VkCommandBuffer commandBuffer, VkLogicOp logicOp) const {
     auto cb_state = GetRead<CMD_BUFFER_STATE>(commandBuffer);
     return ValidateExtendedDynamicState(*cb_state, CMD_SETLOGICOPEXT,
-                                        enabled_features.extended_dynamic_state2_features.extendedDynamicState2LogicOp,
-                                        "VUID-vkCmdSetLogicOpEXT-None-04867", "extendedDynamicState2LogicOp");
+                                        enabled_features.extended_dynamic_state2_features.extendedDynamicState2LogicOp ||
+                                            enabled_features.shader_object_features.shaderObject,
+                                        "VUID-vkCmdSetLogicOpEXT-None-08544", "extendedDynamicState2LogicOp or shaderObject");
 }
 
 bool CoreChecks::PreCallValidateCmdSetPatchControlPointsEXT(VkCommandBuffer commandBuffer, uint32_t patchControlPoints) const {
@@ -606,8 +607,10 @@ bool CoreChecks::PreCallValidateCmdSetPatchControlPointsEXT(VkCommandBuffer comm
     bool skip = false;
     skip |=
         ValidateExtendedDynamicState(*cb_state, CMD_SETPATCHCONTROLPOINTSEXT,
-                                     enabled_features.extended_dynamic_state2_features.extendedDynamicState2PatchControlPoints,
-                                     "VUID-vkCmdSetPatchControlPointsEXT-None-04873", "extendedDynamicState2PatchControlPoints");
+                                     enabled_features.extended_dynamic_state2_features.extendedDynamicState2PatchControlPoints ||
+                                         enabled_features.shader_object_features.shaderObject,
+                                     "VUID-vkCmdSetPatchControlPointsEXT-None-08574",
+                                     "extendedDynamicState2PatchControlPoints or shaderObject");
 
     if (patchControlPoints > phys_dev_props.limits.maxTessellationPatchSize) {
         skip |= LogError(commandBuffer, "VUID-vkCmdSetPatchControlPointsEXT-patchControlPoints-04874",
@@ -623,8 +626,9 @@ bool CoreChecks::PreCallValidateCmdSetRasterizerDiscardEnableEXT(VkCommandBuffer
                                                                  VkBool32 rasterizerDiscardEnable) const {
     auto cb_state = GetRead<CMD_BUFFER_STATE>(commandBuffer);
     return ValidateExtendedDynamicState(*cb_state, CMD_SETRASTERIZERDISCARDENABLEEXT,
-                                        enabled_features.extended_dynamic_state2_features.extendedDynamicState2,
-                                        "VUID-vkCmdSetRasterizerDiscardEnable-None-04871", "extendedDynamicState2");
+                                        enabled_features.extended_dynamic_state2_features.extendedDynamicState2 ||
+                                            enabled_features.shader_object_features.shaderObject,
+                                        "VUID-vkCmdSetRasterizerDiscardEnable-None-08548", "extendedDynamicState2 or shaderObject");
 }
 
 bool CoreChecks::PreCallValidateCmdSetRasterizerDiscardEnable(VkCommandBuffer commandBuffer,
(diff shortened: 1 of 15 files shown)


layers/core_checks/cmd_buffer_dynamic_validation.cpp
```diff
@@ -597,17 +597,20 @@ bool CoreChecks::PreCallValidateCmdSetCheckpointNV(VkCommandBuffer commandBuffer
...
+                                            enabled_features.shader_object_features.shaderObject,
+                                        "VUID-vkCmdSetLogicOpEXT-None-08544", "extendedDynamicState2LogicOp or shaderObject");
 }
 
 bool CoreChecks::PreCallValidateCmdSetPatchControlPointsEXT(VkCommandBuffer commandBuffer, uint32_t patchControlPoints) const {
     auto cb_state = GetRead<CMD_BUFFER_STATE>(commandBuffer);
     bool skip = false;
     skip |=
         ValidateExtendedDynamicState(*cb_state, CMD_SETPATCHCONTROLPOINTSEXT,
-                                     enabled_features.extended_dynamic_state2_features.extendedDynamicState2PatchControlPoints,
```
spencer-lunarg, line 609: traditionally in the past for things like

```
ifdef::VK_EXT_shader_object[]
ifdef::VK_EXT_extended_dynamic_state2[]
  * [[VUID-vkCmdSetPatchControlPointsEXT-None-08574]]
    Either the <<features-extendedDynamicState2PatchControlPoints,
    pname:extendedDynamicState2PatchControlPoints>> feature or the
    <<features-shaderObject, pname:shaderObject>> feature or both must: be
    enabled
endif::VK_EXT_extended_dynamic_state2[]
ifndef::VK_EXT_extended_dynamic_state2[]
  * [[VUID-vkCmdSetPatchControlPointsEXT-None-08575]]
    The <<features-shaderObject, pname:shaderObject>> feature must: be
    enabled
endif::VK_EXT_extended_dynamic_state2[]
endif::VK_EXT_shader_object[]
ifndef::VK_EXT_shader_object[]
  * [[VUID-vkCmdSetPatchControlPointsEXT-None-04873]]
    The <<features-extendedDynamicState2PatchControlPoints,
    pname:extendedDynamicState2PatchControlPoints>> feature must: be enabled
endif::VK_EXT_shader_object[]
```

We would do some ternary operation to select the VUID depending on which extension the user picked. (One reason is so the script that scans which VUID are done catches them)

For sake of this PR's goal, not worried about it, but should be addressed after in the "full validation" PR

scripts/helper_file_generator.py
```diff
@@ -799,7 +800,31 @@ def format_info(ext_name, info):
                 '                auto info = get_info(pCreateInfo->ppEnabledExtensionNames[i]);',
                 '                if (info.state) this->*(info.state) = kEnabledByCreateinfo;',
                 '            }',
-                '        }',
+                '        }' ])
+            if type == 'Device':
+                struct.extend([
+                    '        // Workaround for functions being introducted by multiple extensions, until the layer is fixed to handle this correctly',
```
spencer-lunarg, line 806: @ncesario-lunarg we have a working issue for this so this doesn't just get lost?

layers/state_tracker/state_tracker.cpp
```diff
@@ -5619,6 +5624,16 @@ void ValidationStateTracker::PostCallRecordGetShaderModuleCreateInfoIdentifierEX
     shader_identifier_map_.emplace(*pIdentifier, CreateShaderModuleState(*pCreateInfo, 0, VK_NULL_HANDLE));
 }
 
+void ValidationStateTracker::PreCallRecordCmdBindShadersEXT(VkCommandBuffer commandBuffer, uint32_t stageCount,
+                                                            const VkShaderStageFlagBits* pStages, const VkShaderEXT* pShaders) {
+    auto cb_state = GetWrite<CMD_BUFFER_STATE>(commandBuffer);
+    if (pStages) {
```
spencer-lunarg, line 5630: if `pStages` is false, I would imagine it would get trigged first by `VUID-vkCmdBindShadersEXT-pStages-parameter`/`VUID-vkCmdBindShadersEXT-stageCount-arraylength` and would never make it to `ValidationStateTracker`

spencer-lunarg (approved):

spencer-lunarg:
> Author daniel-story not on autobuild list. Waiting for curator authorization before starting CI build.

@daniel-story I added you to the list and kicked off the CI for you

layers/core_checks/cmd_buffer_dynamic_validation.cpp
ncesario-lunarg: ➕

scripts/helper_file_generator.py
juan-lunarg: Is the issue with our layer or the extension layer that was just added? (https://github.com/KhronosGroup/Vulkan-ExtensionLayer)

scripts/helper_file_generator.py
```diff
@@ -799,7 +800,31 @@ def format_info(ext_name, info):
...
+                    '                VK_EXT_VERTEX_INPUT_DYNAMIC_STATE_EXTENSION_NAME,',
+                    '            };',
+                    '            auto info = get_info(VK_EXT_SHADER_OBJECT_EXTENSION_NAME);',
+                    '            if (info.state) {',
+                    '                if (this->*(info.state) != kNotEnabled) {',
+                    '                    for (auto interaction_ext : shader_object_interactions) {',
+                    '                        info = get_info(interaction_ext);',
+                    '                        assert(info.state);',
+                    '                        if (this->*(info.state) != kEnabledByCreateinfo) {',
+                    '                            this->*(info.state) = kEnabledByInteraction;',
```
ncesario-lunarg, line 821: Did you happen to try out using `kEnabledByCreateInfo` to avoid adding another `ExtEnabled` state?

ncesario-lunarg (commented):
A couple questions/comments, but this LGTM.

layers/generated/vk_extension_helper.h
```diff
@@ -1422,6 +1423,27 @@ struct DeviceExtensions : public InstanceExtensions {
...
+                    for (auto interaction_ext : shader_object_interactions) {
+                        info = get_info(interaction_ext);
+                        assert(info.state);
+                        if (this->*(info.state) != kEnabledByCreateinfo) {
+                            this->*(info.state) = kEnabledByInteraction;
+                        }
+                    }
+                }
+            }
+        }
```
ncesario-lunarg, line 1446: Looking at this a bit more, I actually do not understand why this is needed at all. The purpose of the code in this function is to ensure that dependent extensions have been enabled, and my initial understanding was that these extensions were _dependencies_ of VK_EXT_shader_object. However, that does not appear to be the case (this extension appears to only depend on `VK_KHR_get_physical_device_properties2` and `VK_KHR_dynamic_rendering`). Even if these extensions were dependencies, this would actually still be incorrect.

@daniel-story can you please elaborate on what this is meant to solve?

ncesario-lunarg (changes requested):

scripts/helper_file_generator.py
spencer-lunarg: @daniel-story can you add a link to 5579 in the comment?

layers/generated/vk_extension_helper.h
ncesario-lunarg: Thank you for the clarification (on gitlab) @daniel-story. Per my comments over on gitlab, can you please try using https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5590? If that works, that should be a general solution for this particular problem (i.e., GetDeviceProcAddr), and then `kEnabledByInteraction` should not be necessary.

layers/generated/vk_extension_helper.h
```diff
@@ -1422,6 +1423,27 @@ struct DeviceExtensions : public InstanceExtensions {
                 if (info.state) this->*(info.state) = kEnabledByCreateinfo;
             }
         }
+        // Workaround for functions being introducted by multiple extensions, until the layer is fixed to handle this correctly
```
juan-lunarg, line 1426: `introduceted`?

layers/generated/vk_extension_helper.h
```diff
@@ -1422,6 +1423,27 @@ struct DeviceExtensions : public InstanceExtensions {
...
+        {
+            constexpr std::array<const char*, 4> shader_object_interactions = {
```
juan-lunarg, line 1428: Use CTAD

```suggestion
            constexpr std::array shader_object_interactions = {
```

layers/state_tracker/cmd_buffer_state.cpp
```diff
@@ -1078,7 +1078,7 @@ void CMD_BUFFER_STATE::UpdatePipelineState(CMD_TYPE cmd_type, const VkPipelineBi
     const auto lv_bind_point = ConvertToLvlBindPoint(bind_point);
     auto &last_bound = lastBound[lv_bind_point];
     PIPELINE_STATE *pipe = last_bound.pipeline_state;
-    if (VK_NULL_HANDLE != last_bound.pipeline_layout) {
+    if (pipe && VK_NULL_HANDLE != last_bound.pipeline_layout) {
```
juan-lunarg, line 1081: Return early if pipe is false.

Also avoids a check on line 1153

layers/generated/vk_extension_helper.h
ncesario-lunarg: Daniel pointed out another issue: https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5600. "Extension interactions" apparently have always been very broken and we've just gotten lucky about not running into any issues in the past.

In light of this, I'm good with merging this as is to get things unblocked for the upcoming SDK.

scripts/helper_file_generator.py
ncesario-lunarg: https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5600 is also related.

ncesario-lunarg (approved):

Issue #5492, opened 2023-03-23
Missing check when dynamic rendering's `vkCmdEndRendering` is called without beginning with `vkCmdBeginRendering`.


nikitablack:
**Describe the situation in which you encountered the missing validation**

- Begin first rendering with `vkCmdBeginRendering()`.
- Begin second rendering with `vkCmdBeginRendering()`.
- End second rendering with `vkCmdEndRendering()`.
- End first rendering with `vkCmdEndRendering()`. <- Segmentation fault happens here.

I accidentally started a rendering while already in a rendering. Instead of a warning/error, I'm receiving a frustrating segfault.

**Valid Usage IDs requested**

I believe `VUID-vkCmdEndRendering-None-06161` is the valid usage in this situation.

spencer-lunarg:
Asking the Working Group, the spec doesn't seem to cover a case if you call `vkCmdBeginRendering` twice

Once a clarification is made, will fix the Validation Layers accordingly (and write tests)

spencer-lunarg:
ok, so I realize ` VUID-vkCmdBeginRendering-renderpass` ( or normally`VUID-vkCmdBeginRenderPass-renderpass`) should have caught the 2nd `vkCmdBeginRendering` 

So my question is did you not get that error? (or ` VUID-vkCmdEndRendering-renderpass` on the 2nd ending?)

nikitablack:
No, the only thing I get is a segfault on the 2nd ending.

spencer-lunarg:
opened a PR, we were missing the check, wrote some regression tests

Thanks for raising the issue!

PR #5574, opened 2023-04-05
Fix DynamicRendering calling command cmd helper

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5492

We were not calling `ValidateCmd` for `vkCmdBeginRendering`/`vkCmdEndRendering` and missing the checks from it

### DIFF
--- a/layers/core_checks/render_pass_validation.cpp
+++ b/layers/core_checks/render_pass_validation.cpp
@@ -3067,6 +3067,7 @@ bool CoreChecks::ValidateCmdBeginRendering(VkCommandBuffer commandBuffer, const 
     if (!cb_state) return false;
     bool skip = false;
     const char *func_name = CommandTypeString(cmd_type);
+    skip |= ValidateCmd(*cb_state, cmd_type);
 
     const auto chained_device_group_struct = LvlFindInChain<VkDeviceGroupRenderPassBeginInfo>(pRenderingInfo->pNext);
     const bool non_zero_device_render_area = chained_device_group_struct && chained_device_group_struct->deviceRenderAreaCount != 0;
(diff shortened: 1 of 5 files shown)


layers/core_checks/render_pass_validation.cpp
```diff
@@ -3785,7 +3776,7 @@ bool CoreChecks::PreCallValidateCmdBeginRendering(VkCommandBuffer commandBuffer,
 
 // If a renderpass is active, verify that the given command type is appropriate for current subpass state
 bool CoreChecks::ValidateCmdSubpassState(const CMD_BUFFER_STATE &cb_state, const CMD_TYPE cmd_type) const {
-    if (!cb_state.activeRenderPass) return false;
+    if (!cb_state.activeRenderPass || cb_state.activeRenderPass->UsesDynamicRendering()) return false;
```
artem-lunarg, line 3779: We have `CMD_EXECUTECOMMANDS` validation for regular render pass here,  but for dynamic rendering we check `VK_RENDERING_CONTENTS_SECONDARY_COMMAND_BUFFERS_BIT`  in `PreCallValidateCmdExecuteCommands`. I wonder if it's possible/makes sense to make these two validation in a single place (ValidateCmdSubpassState or PreCallValidateCmdExecuteCommands) ?

juan-lunarg (approved):

artem-lunarg (approved):
LGTM!

ncesario-lunarg (approved):

layers/core_checks/render_pass_validation.cpp
spencer-lunarg: agree, actually I have a lot of things I want to fix about this "command draw code" after looking at for this PR

PR #5577, opened 2023-04-05
layers: Fix index buffer size treatment

### LINKED ISSUE #5575: CoreChecks::ValidateCmdDrawIndexedBufferSize is double-counting the offset
Seeing failures in ANGLE, I took a look and this code is double-counting the index offset:

```
        VkDeviceSize end_offset = static_cast<VkDeviceSize>(index_size * (firstIndex + indexCount)) + index_buffer_binding.offset;
        if (end_offset > index_buffer_binding.size) {
```

I see that `index_buffer_binding.size + index_buffer_binding.offset` is the create info size, so clearly `index_buffer_binding.size` has already taken offset into account. `end_offset` can just drop `+ index_buffer_binding.offset`.

jzulauf-lunarg:
Fixing earlier change that make bound index buffer size *consistent* with bound vertex buffers, but didn't account for it's usage a CoreChecks validation.

Fixes #5575

### DIFF
--- a/layers/core_checks/drawdispatch_validation.cpp
+++ b/layers/core_checks/drawdispatch_validation.cpp
@@ -2211,14 +2211,17 @@ bool CoreChecks::ValidateCmdDrawIndexedBufferSize(const CMD_BUFFER_STATE &cb_sta
     if (!enabled_features.robustness2_features.robustBufferAccess2 && cb_state.index_buffer_binding.bound()) {
         const auto &index_buffer_binding = cb_state.index_buffer_binding;
         const uint32_t index_size = GetIndexAlignment(index_buffer_binding.index_type);
-        VkDeviceSize end_offset = static_cast<VkDeviceSize>(index_size * (firstIndex + indexCount)) + index_buffer_binding.offset;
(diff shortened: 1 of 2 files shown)


jzulauf-lunarg:
@Tony-LunarG -- can you check against your case?

ncesario-lunarg (approved):

ncesario-lunarg:
Should we have a test case for this? e.g., if the check for 07825 in `VkLayerTest.MultiDrawTests` used an offset other than 0, would we have caught this earlier?

jzulauf-lunarg:
Would have needed a positive test with a non-zero offset.

ncesario-lunarg:
> Would have needed a positive test with a non-zero offset.

Yes, that should probably be added.

juan-lunarg (dismissed):
Add a positive test

jzulauf-lunarg:
Add two cases to existing test, the first of which functions as a positive test, since we don't allow any unexpected at this point.

jzulauf-lunarg:
Just noted I need to update the error message.

juan-lunarg (approved):
LGTM thanks for the added testing :+1:

PR #5581, opened 2023-04-05
test: Remove static from Video function pointers

charles-lunarg:
Various functions are being loaded using vkGetDeviceProcAddr then stored inside a static local variable. This means subsequent tests that use those functions will contain a *stale* function pointer, since the device is different and the dynamic libraries would have been unloaded.


ncesario-lunarg (approved):

spencer-lunarg (approved):

arno-lunarg (approved):

arno-lunarg:
I would have trusted ASAN to catch that (。﹏。*)

PR #5584, opened 2023-04-06
layers: Account for `VkSurfaceFullScreenExclusiveInfoEXT` when validating swapchain creation

arno-lunarg:
Originally was https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4703, but since rebasing this old PR lead to terrible merge issues, I'd rather leave the old one untouched just in case.

- Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4594

The main thing about the fix is an update to `SURFACE_STATE` to use `VkSurfaceFormat2KHR`, `VkSurfaceCapabilities2KHR` and `VkPhysicalDeviceSurfaceInfo2KHR`. It now relies on `DispatchGetPhysicalDeviceSurfaceFormats2KHR` and `DispatchGetPhysicalDeviceSurfaceCapabilities2KHR` when `vk_khr_get_surface_capabilities2` is enabled. Otherwise it fall backs to the original functions.

This PR defines a `LogInternalError` used to log errors we get from our internal `DispatchVkFoo` calls.

### DIFF
--- a/layers/core_checks/wsi_validation.cpp
+++ b/layers/core_checks/wsi_validation.cpp
@@ -69,7 +69,7 @@ bool CoreChecks::ValidateSwapchainPresentModesCreateInfo(VkPresentModeKHR presen
     bool skip = false;
     std::vector<VkPresentModeKHR> present_modes{};
     if (surface_state) {
-        present_modes = surface_state->GetPresentModes(physical_device);
+        present_modes = surface_state->GetPresentModes(physical_device, this);
     } else if (IsExtEnabled(instance_extensions.vk_google_surfaceless_query)) {
         present_modes = physical_device_state->surfaceless_query_state.present_modes;
     }
@@ -125,7 +125,7 @@ bool CoreChecks::ValidateSwapchainPresentModesCreateInfo(VkPresentModeKHR presen
 }
 
 bool CoreChecks::ValidateSwapchainPresentScalingCreateInfo(VkPresentModeKHR present_mode, const char *func_name,
-                                                           VkSurfaceCapabilitiesKHR *capabilities,
+                                                           const VkSurfaceCapabilitiesKHR *capabilities,
                                                            VkSwapchainCreateInfoKHR const *create_info,
                                                            const SURFACE_STATE *surface_state) const {
     bool skip = false;
@@ -341,10 +341,42 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
         }
     }
 
-    VkSurfaceCapabilitiesKHR capabilities{};
-    DispatchGetPhysicalDeviceSurfaceCapabilitiesKHR(physical_device_state->PhysDev(), pCreateInfo->surface, &capabilities);
+    // Will be assigned &full_screen_info_copy if a VkSurfaceFullScreenExclusiveInfoEXT is found in pCreateInfo->pNext chain
+    const void *full_screen_info_copy_addr = nullptr;
+#if defined(VK_USE_PLATFORM_WIN32_KHR)
+    auto full_screen_info_copy = LvlInitStruct<VkSurfaceFullScreenExclusiveInfoEXT>();
+    auto win32_full_screen_info_copy = LvlInitStruct<VkSurfaceFullScreenExclusiveWin32InfoEXT>();
+    if (IsExtEnabled(device_extensions.vk_ext_full_screen_exclusive)) {
+        const auto *full_screen_info = LvlFindInChain<VkSurfaceFullScreenExclusiveInfoEXT>(pCreateInfo->pNext);
+        if (full_screen_info) {
+            full_screen_info_copy_addr = &full_screen_info_copy;
+            full_screen_info_copy = *full_screen_info;
+            full_screen_info_copy.pNext = nullptr;
+
+            if (IsExtEnabled(device_extensions.vk_khr_win32_surface)) {
+                const auto *win32_full_screen_info = LvlFindInChain<VkSurfaceFullScreenExclusiveWin32InfoEXT>(pCreateInfo->pNext);
+                if (!win32_full_screen_info) {
+                    if (LogError(device, "VUID-VkSwapchainCreateInfoKHR-pNext-02679",
+                                 "%s: pCreateInfo->pNext contains "
+                                 "VkSurfaceFullScreenExclusiveInfoEXT, but pCreateInfo->pNext does not contain "
+                                 "VkSurfaceFullScreenExclusiveWin32InfoEXT",
+                                 func_name)) {
+                        return true;
+                    }
+                } else {
+                    win32_full_screen_info_copy = *win32_full_screen_info;
+                    win32_full_screen_info_copy.pNext = nullptr;
+                    full_screen_info_copy.pNext = &win32_full_screen_info_copy;
+                }
+            }
+        }
+    }
+#endif
+    const auto surface_caps2 = surface_state->GetCapabilities(IsExtEnabled(instance_extensions.vk_khr_get_surface_capabilities2),
+                                                              physical_device_state->PhysDev(), full_screen_info_copy_addr, this);
+
     bool skip = false;
-    VkSurfaceTransformFlagBitsKHR current_transform = capabilities.currentTransform;
+    VkSurfaceTransformFlagBitsKHR current_transform = surface_caps2.surfaceCapabilities.currentTransform;
     if ((pCreateInfo->preTransform & current_transform) != pCreateInfo->preTransform) {
         skip |= LogPerformanceWarning(physical_device, kVUID_Core_Swapchain_PreTransform,
                                       "%s: pCreateInfo->preTransform (%s) doesn't match the currentTransform (%s) returned by "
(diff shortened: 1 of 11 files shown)


layers/core_checks/wsi_validation.cpp
```diff
@@ -305,13 +305,14 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
     // application to have established support on any other platform.
     if (!instance_extensions.vk_khr_android_surface) {
         // restrict search only to queue families of VkDeviceQueueCreateInfos, not the whole physical device
-        const bool is_supported = AnyOf<QUEUE_STATE>([this, surface_state](const QUEUE_STATE &queue_state) {
+        bool is_supported = AnyOf<QUEUE_STATE>([this, surface_state](const QUEUE_STATE &queue_state) {
```
spencer-lunarg, line 308: why did this have to lose it's `const`?

layers/core_checks/wsi_validation.cpp
```diff
@@ -305,13 +305,14 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
...
             return surface_state->GetQueueSupport(physical_device, queue_state.queueFamilyIndex);
         });
 
         if (!is_supported) {
-            const LogObjectList objlist(device, surface_state->Handle());
-            if (LogError(objlist, "VUID-VkSwapchainCreateInfoKHR-surface-01270",
+            LogObjectList objs(device);
```
spencer-lunarg, line 313: I think you can go ` LogObjectList objlist(device, surface_state->Handle());` now

also `objlist` so it is consistent how we name it else where

layers/generated/chassis.h
```diff
@@ -4091,6 +4092,15 @@ class ValidationObject {
             return result;
         }
 
+        void LogInternalError(std::string_view failure_location, const LogObjectList &obj_list,
+                              std::string_view entrypoint, VkResult err) const {
+            const std::string_view err_string = string_VkResult(err);
+            std::string vuid = "INTERNAL-ERROR-";
+            vuid += entrypoint;
+            LogError(obj_list, vuid, "In %s: %s() failed with result = %s.", failure_location.data(), entrypoint.data(),
```
spencer-lunarg, line 4100: I would maybe say

```
"In %s: %s() was called in the Validation Layer state tracking and failed with result = %s."
```

so end users 100% know we made the dispatch call

layers/state_tracker/image_state.cpp
```diff
@@ -686,62 +687,159 @@ void SURFACE_STATE::SetPresentModes(VkPhysicalDevice phys_dev, vvl::span<const V
...
+        if (validation_obj) {
+            LogObjectList obj_list(std::forward<decltype(objects)>(objects)...);
+            validation_obj->LogInternalError(__PRETTY_FUNCTION__, obj_list, "vkGetPhysicalDeviceSurfacePresentModesKHR", err);
+        }
+    };
+
     uint32_t count = 0;
-    DispatchGetPhysicalDeviceSurfacePresentModesKHR(phys_dev, surface(), &count, nullptr);
+    if (const VkResult err = DispatchGetPhysicalDeviceSurfacePresentModesKHR(phys_dev, surface(), &count, nullptr);
+        !IsValueIn(err, {VK_SUCCESS, VK_INCOMPLETE})) {
```
spencer-lunarg, line 711: Something we can probably do (after this PR) is grab the `successcodes` and generate this, might not be worth it

layers/state_tracker/image_state.cpp
```diff
@@ -686,62 +687,159 @@ void SURFACE_STATE::SetPresentModes(VkPhysicalDevice phys_dev, vvl::span<const V
...
 
-void SURFACE_STATE::SetCapabilities(VkPhysicalDevice phys_dev, const VkSurfaceCapabilitiesKHR &caps) {
+void SURFACE_STATE::SetCapabilities(VkPhysicalDevice phys_dev, const safe_VkSurfaceCapabilities2KHR &caps) {
     auto guard = Lock();
     assert(phys_dev);
     capabilities_[phys_dev] = caps;
 }
 
-VkSurfaceCapabilitiesKHR SURFACE_STATE::GetCapabilities(VkPhysicalDevice phys_dev) const {
+safe_VkSurfaceCapabilities2KHR SURFACE_STATE::GetCapabilities(bool get_surface_capabilities2, VkPhysicalDevice phys_dev,
```
spencer-lunarg, line 809: curious if there a good way to get this info passed in, I know in `IMAGE_STATE::MemoryReqs GetMemoryRequirements(` we pass the `ValidationStateTracker` object... `IsExtEnabled(instance_extensions.vk_khr_get_surface_capabilities2)` I think is fine for now, but future idea to look into

spencer-lunarg (approved):
outside my comments, I think this overall is good and like got a `LogInternalError` out of this to start using around!

layers/core_checks/wsi_validation.cpp
arno-lunarg: Argh, merging issues : /

layers/generated/chassis.h
arno-lunarg: yep better!

layers/core_checks/wsi_validation.cpp
```diff
@@ -341,10 +341,40 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
...
+            if (!win32_full_screen_info) {
+                if (LogError(device, "VUID-VkSwapchainCreateInfoKHR-pNext-02679",
+                             "%s: pCreateInfo->pNext contains "
+                             "VkSurfaceFullScreenExclusiveInfoEXT, but pCreateInfo->pNext does not contain "
+                             "VkSurfaceFullScreenExclusiveWin32InfoEXT",
+                             func_name)) {
+                    return true;
+                }
+            }
+            win32_full_screen_info_copy = *win32_full_screen_info;
```
artem-lunarg, line 367: Is it possible to get here so `win32_full_screen_info` is null? Mostly the question if LogError above can return false.

layers/core_checks/wsi_validation.cpp
```diff
@@ -341,10 +341,40 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
...
+    if (IsExtEnabled(device_extensions.vk_ext_full_screen_exclusive)) {
+        const auto *full_screen_info = LvlFindInChain<VkSurfaceFullScreenExclusiveInfoEXT>(pCreateInfo->pNext);
+        full_screen_info_copy = LvlInitStruct<VkSurfaceFullScreenExclusiveInfoEXT>();
+        if (full_screen_info) {
+            full_screen_info_copy_addr = &full_screen_info_copy;
+            full_screen_info_copy = *full_screen_info;
+            full_screen_info_copy.pNext = nullptr;
+
+            const auto *win32_full_screen_info = LvlFindInChain<VkSurfaceFullScreenExclusiveWin32InfoEXT>(pCreateInfo->pNext);
+            if (!win32_full_screen_info) {
```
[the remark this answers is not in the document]
arno-lunarg: Yep it is, I completely forgot!

layers/core_checks/wsi_validation.cpp
arno-lunarg: Ah yes it could be null, I am going to guard it

layers/state_tracker/image_state.cpp
arno-lunarg: I do not have a strong opinion on this. I think that using a `ValidationStateTracker` object would obfuscate things a bit, since we would have to look at the function implementation to know what data is needed from the object. Directly passing the needed data make things clearer to me.

layers/core_checks/wsi_validation.cpp
```diff
@@ -125,7 +125,7 @@ bool CoreChecks::ValidateSwapchainPresentModesCreateInfo(VkPresentModeKHR presen
 }
 
 bool CoreChecks::ValidateSwapchainPresentScalingCreateInfo(VkPresentModeKHR present_mode, const char *func_name,
-                                                           VkSurfaceCapabilitiesKHR *capabilities,
+                                                           const VkSurfaceCapabilitiesKHR *capabilities,
```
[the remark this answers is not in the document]
arno-lunarg: How so 🤔? This variable is used to compute valid bounds

layers/utils/vk_layer_utils.h
```diff
@@ -39,7 +39,17 @@
 #define STRINGIFY(s) STRINGIFY_HELPER(s)
 #define STRINGIFY_HELPER(s) #s
 
+// For MSVC
+#if !defined(__PRETTY_FUNCTION__)
+#if defined(__FUNCSIG__)
+#define __PRETTY_FUNCTION__ __FUNCSIG__
```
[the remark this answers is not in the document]
arno-lunarg: Ah yes I need to come up with something else!

layers/core_checks/wsi_validation.cpp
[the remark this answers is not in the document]
artem-lunarg: Sorry, my mistake if it is used! I could not see usage examples by looking at github code and even did some text search... maybe it was hidden in collapsed section.

layers/core_checks/wsi_validation.cpp
```diff
@@ -341,10 +341,42 @@ bool CoreChecks::ValidateCreateSwapchain(const char *func_name, VkSwapchainCreat
         }
     }
 
-    VkSurfaceCapabilitiesKHR capabilities{};
-    DispatchGetPhysicalDeviceSurfaceCapabilitiesKHR(physical_device_state->PhysDev(), pCreateInfo->surface, &capabilities);
+    // Will be assigned &full_screen_info_copy if a VkSurfaceFullScreenExclusiveInfoEXT is found in pCreateInfo->pNext chain
+    const void *full_screen_info_copy_addr = nullptr;
+#if defined(VK_USE_PLATFORM_WIN32_KHR)
```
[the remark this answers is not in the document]
arno-lunarg: I have nothing to validate in the non WIN32 env yet!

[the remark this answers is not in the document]
arno-lunarg: Though you make me realize that I could implement `VUID-VkPhysicalDeviceSurfaceInfo2KHR-pNext-02672`, which is similar to the VUID added here. Created https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5595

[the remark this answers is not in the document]
artem-lunarg: Isn't it possible to have `VkSurfaceFullScreenExclusiveInfoEXT` in the pNext chain on non-windows setup? I thought that `VkSurfaceFullScreenExclusiveWin32InfoEXT` is only required on win32 to accompany `VkSurfaceFullScreenExclusiveInfoEXT` but on other platforms `VkSurfaceFullScreenExclusiveInfoEXT` should work too, but maybe I missed something.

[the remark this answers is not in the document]
arno-lunarg: It is possible, but I did not find any `VUID-VkPhysicalDeviceSurfaceInfo2KHR` that needs to be checked in this case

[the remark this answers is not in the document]
artem-lunarg: Thanks for clarification. I missed that validation inside in win32 specific.

artem-lunarg (approved):
LGTM!

tests/negative/wsi.cpp
juan-lunarg, line 1: This test isn't getting tested on CI.

```
#ifdef VVL_TESTS_ENABLE_EXCLUSIVE_FULLSCREEN
TEST_F(VkLayerTest, TestvkAcquireFullScreenExclusiveModeEXT) {
```

We are not defining `VVL_TESTS_ENABLE_EXCLUSIVE_FULLSCREEN` anywhere in our build and/or CI process.

Is there a reason we can't add this test to regular CI? Our just run this test only on a mock driver?

arno-lunarg: @ncesario-lunarg disabled this test in 1c9bacb239863028189050737cfa2056d738409f
This test is only enabled on Windows, but at least locally I can run it with both my nvidia driver and the mock driver.
I am going to add a new commit with this macro removed and see how things go

layers/state_tracker/image_state.cpp
ncesario-lunarg: Maybe store the result of `IsExtEnabled(device_extensions.vk_khr_get_surface_capabilities2)` in `SURFACE_STATE` at surface creation time?

tests/negative/wsi.cpp
ncesario-lunarg: I do not think we can request "exclusive full screen" in CI due to shared resources, at least I think that's the gist of why this was disabled (@lunarpapillo's archive probably has better details). If that's the case I don't think we can enable this without some sort of change to CI.

ncesario-lunarg (approved):
LGTM!

PR #5585, opened 2023-04-06
Add ExecutionModeSet

spencer-lunarg:
Adds `ExecutionModeSet` to act like the `DecorationSet` we have

Instead of having to look up and re-find each `OpExecutionMode` value, this just stores them reducing many loops

While at it, I also moved all the `const Instruction& entrypoint` to `const EntryPoint& entrypoint` as it allows for grabbing more information as it is passed around

Also added a `transform_feedback_stream_inst` lookup for `ValidateTransformFeedback` instead of re-looping the module

Lastly, in attempt to make `SHADER_MODULE_STATE` simpler, I removed the getters in favor of just grabbing `static_data_` as it was not being done consistently and `StaticData` is const so can't write it anyway

### DIFF
--- a/layers/state_tracker/shader_module.h
+++ b/layers/state_tracker/shader_module.h
@@ -315,7 +315,7 @@ struct SHADER_MODULE_STATE : public BASE_NODE {
 
         bool has_passthrough{false};
 
-        EntryPoint(const SHADER_MODULE_STATE &module_state, const Instruction &entrypoint);
+        EntryPoint(const SHADER_MODULE_STATE &module_state, const Instruction &entrypoint_insn);
     };
 
     // Static/const data extracted from a SPIRV module at initialization time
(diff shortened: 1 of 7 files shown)


layers/state_tracker/shader_module.h
```diff
@@ -80,6 +80,55 @@ struct DecorationSet : public DecorationBase {
...
+        point_mode_bit = 1 << 1,
+        post_depth_coverage_bit = 1 << 2,
+        local_size_bit = 1 << 3,
+        local_size_id_bit = 1 << 4,
+        iso_lines_bit = 1 << 5,
+        xfb_bit = 1 << 6,
+        early_fragment_test_bit = 1 << 7,
+        subgroup_uniform_control_flow_bit = 1 << 8,
+
+        signed_zero_inf_nan_preserve_width_16 = 1 << 8,
```
arno-lunarg, line 96: those 2 have the same bit, is it a typo?

ncesario-lunarg (approved):
:tada:

PR #5587, opened 2023-04-06
layers: Cleanup false sharing code

juan-lunarg:
Instead of using an array of bytes to pad structs/classes it's better to use `alignas` as shown in this `cppreference` example:
https://en.cppreference.com/w/cpp/thread/hardware_destructive_interference_size

I did initially try using `std::hardware_destructive_interference_size` and it worked on Windows/MacOS, but had issues on Linux/Android.

Even though `std::hardware_destructive_interference_size` is in C++17 it's not fully supported on all platforms and the only way to properly check for it is to use C++20 feature testing...

### DIFF
--- a/layers/utils/vk_layer_utils.h
+++ b/layers/utils/vk_layer_utils.h
@@ -497,6 +497,11 @@ class LockedSharedPtr : public std::shared_ptr<T> {
     Guard guard_;
 };
 
+// https://en.cppreference.com/w/cpp/thread/hardware_destructive_interference_size
+// https://en.wikipedia.org/wiki/False_sharing
+// TODO use C++20 to check for std::hardware_destructive_interference_size feature support.
+constexpr std::size_t get_hardware_destructive_interference_size() { return 64; }
+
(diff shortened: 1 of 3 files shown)


layers/utils/vk_layer_utils.h
```diff
@@ -497,6 +497,11 @@ class LockedSharedPtr : public std::shared_ptr<T> {
     Guard guard_;
 };
 
+// https://en.cppreference.com/w/cpp/thread/hardware_destructive_interference_size
+// https://en.wikipedia.org/wiki/False_sharing
+// TODO use C++20 to check for std::hardware_destructive_interference_size feature support.
+constexpr std::size_t get_hardware_destructive_interference_size() { return 64; }
```
artem-lunarg, line 503: Apple M1 has 128 cache line size. 
Such type of comments probably does not add to karma, but I will start using 'hardware destructive interference size' term instead of cache line to redeem myself.

artem-lunarg (approved):

PR #5588, opened 2023-04-06
Initial support for Android in main CMake build

juan-lunarg:
- Minor C++ code fixes
- Minor fix to CMake to address finding things when cross-compiling
- Add minimal CI


juan-lunarg:
Currently testing doesn't compile. That will be the next thing to address.

spencer-lunarg (approved):

.github/workflows/vvl.yml
```diff
@@ -215,6 +215,31 @@ jobs:
...
+        - uses: lukka/get-cmake@latest
+        - name: Setup ccache
+          uses: hendrikmuhs/ccache-action@v1.2
+          with:
+            key: ${{ runner.os }}-android-cmake
+        - name: Add ccache to PATH
+          run: echo "/usr/lib/ccache" >> $GITHUB_PATH
+        - name: Configure
+          run: |
+            cmake -S . -B build/ --toolchain $ANDROID_NDK_ROOT/build/cmake/android.toolchain.cmake \
```
juan-lunarg, line 234: https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#android

1.) The Android NDK comes pre-installed
2.) `ANDROID_NDK_ROOT` is set by the docker image for us

ncesario-lunarg (approved):
:scream: are these really the only changes that are necessary? If so, that would be truly awesome!

juan-lunarg:
> scream are these really the only changes that are necessary? If so, that would be truly awesome!

The hard part is going to be untangling the old systems. Actually adding the android support to the main CMake build isn't that hard.

PR #5590, opened 2023-04-07
layers: Fix GDPA for spec "interaction" 

ncesario-lunarg:
The Vulkan spec allows commands to be defined by multiple extensions,
which breaks the current GetDeviceProcAddr logic. Mapping commands to a
list of extensions rather than a single extension fixes this issue.


tests/positive/other.cpp
```diff
@@ -963,4 +963,17 @@ TEST_F(VkPositiveLayerTest, FormatProperties3FromProfiles) {
     auto fmt_props = LvlInitStruct<VkFormatProperties2>(&fmt_props_3);
     vk::GetPhysicalDeviceFormatProperties2(gpu(), VK_FORMAT_R8_UNORM, &fmt_props);
     vk::GetPhysicalDeviceFormatProperties2(gpu(), VK_FORMAT_R8G8B8A8_UNORM, &fmt_props);
-}
\ No newline at end of file
+}
+
+TEST_F(VkPositiveLayerTest, GDPAWithMultiCmdExt) {
+    TEST_DESCRIPTION("Use GetDeviceProcAddr on a function which is provided by multiple extensions");
+    AddRequiredExtensions(VK_EXT_SHADER_OBJECT_EXTENSION_NAME);
```
ncesario-lunarg, line 970: Note that I don't think this is actually running anywhere yet...

layers/generated/vk_dispatch_table_helper.h
```diff
@@ -620,416 +620,417 @@ static VKAPI_ATTR void VKAPI_CALL StubCmdDrawMeshTasksIndirectCountEXT(VkCommand
...
-    {"vkUnmapMemory2KHR", "VK_KHR_map_memory2"},
-    {"vkUpdateDescriptorSetWithTemplate", "VK_VERSION_1_1"},
-    {"vkUpdateDescriptorSetWithTemplateKHR", "VK_KHR_descriptor_update_template"},
-    {"vkUpdateVideoSessionParametersKHR", "VK_KHR_video_queue"},
-    {"vkWaitForPresentKHR", "VK_KHR_present_wait"},
-    {"vkWaitSemaphores", "VK_VERSION_1_2"},
-    {"vkWaitSemaphoresKHR", "VK_KHR_timeline_semaphore"},
-    {"vkWriteAccelerationStructuresPropertiesKHR", "VK_KHR_acceleration_structure"},
-    {"vkWriteMicromapsPropertiesEXT", "VK_EXT_opacity_micromap"},
+const vvl::unordered_map<std::string, std::vector<std::string>> api_extension_map {
```
juan-lunarg, line 623: Looks like a good use case for `small_vector`

scripts/dispatch_table_helper_generator.py
```diff
@@ -226,7 +226,7 @@ def getTypeNameTuple(self, param):
     def OutputExtEnabledFunction(self):
         ext_fcn = ''
         # First, write out our static data structure -- map of all APIs that are part of extensions to their extension.
-        ext_fcn += 'const vvl::unordered_map<std::string, std::string> api_extension_map {\n'
+        ext_fcn += 'const vvl::unordered_map<std::string, small_vector<std::string, 4>> api_extension_map {\n'
```
arno-lunarg, line 229:
```suggestion
        ext_fcn += 'const vvl::unordered_map<std::string, small_vector<std::string, 2>> api_extension_map {\n'
```
Suggesting 2 because right now I see at most 2 elements in the list

arno-lunarg (approved):
LGTM!

artem-lunarg (approved):

scripts/dispatch_table_helper_generator.py
ncesario-lunarg: Yes, I believe there are only 2. The reason I went with 4 is an attempt to "future proof" this a bit, but maybe that's not having the intended affect.

ncesario-lunarg: Actually, I'll just grab the max list size during codegen; that's probably the "right" thing to do here.

artem-lunarg: Is max size for all entries? In that case might look less pretty if we have outliers, so extra space should be paid by each entry. Not that we lose a lot of memory but if we expect that reasonable threshold (2-4) will work for most entries, then few allocations might be not an issue. Tradeoffs everywhere!

PR #5592, opened 2023-04-07
layers: Add new TypeStructInfo class

spencer-lunarg:
adds a `TypeStructInfo` class

This is used to help manage `OpTypeStruct` information

Currently there is a `StructInfo` but that was used only to manage the Push Constants (and has to deal with structs of struct, currently the only user for `TypeStructInfo` is `StageInterfaceVariable` which is not allowed to have nested Structs)

The next goal is to migrate all of `StructInfo` into `OpTypeStruct`


arno-lunarg (approved):
LGTM!

PR #5599, opened 2023-04-07
android: Remove VK_PROTOTYPES define

juan-lunarg:
Not used anywhere


spencer-lunarg (approved):
Background behind this:

The NDK had/(has?) a [vulkan_wrapper.c/h](https://android.googlesource.com/platform/external/vulkan-validation-layers/+/refs/heads/oreo-m4-s3-release/common/vulkan_wrapper.h) file for indirectly linking

PR #5601, opened 2023-04-08
layers: Hotfix for nested struct interface

spencer-lunarg:
I miss read (as commented below) the spec and forgot about nested structs inbetween the interface

For now just skip checking nested struct User Defined Variables until proper fix (and VVL tests) are added

@Tony-LunarG  this fixes CTS on my end

### DIFF
--- a/layers/core_checks/shader_validation.cpp
+++ b/layers/core_checks/shader_validation.cpp
@@ -2954,12 +2954,18 @@ bool CoreChecks::ValidateInterfaceBetweenStages(const SHADER_MODULE_STATE &produ
 
     for (const auto &interface_slot : producer_entrypoint.output_interface_slots) {
         auto &slot = slot_map[interface_slot.first.Location()][interface_slot.first.Component()];
+        if (interface_slot.second->nested_struct) {
+            return skip;  // TODO workaround
+        }
         slot.output = interface_slot.second;
         slot.output_type = interface_slot.first.type;
         slot.output_width = interface_slot.first.bit_width;
     }
     for (const auto &interface_slot : consumer_entrypoint.input_interface_slots) {
         auto &slot = slot_map[interface_slot.first.Location()][interface_slot.first.Component()];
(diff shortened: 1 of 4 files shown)


layers/core_checks/shader_validation.cpp
```diff
@@ -2954,12 +2954,18 @@ bool CoreChecks::ValidateInterfaceBetweenStages(const SHADER_MODULE_STATE &produ
 
     for (const auto &interface_slot : producer_entrypoint.output_interface_slots) {
         auto &slot = slot_map[interface_slot.first.Location()][interface_slot.first.Component()];
+        if (interface_slot.second->nested_struct) {
+            return skip;  // TODO workaround
```
spencer-lunarg, line 2958: Actively fixing now, so don't need issue tracker

layers/state_tracker/shader_module.cpp
```diff
@@ -1493,47 +1493,59 @@ bool StageInteraceVariable::IsBuiltin(const StageInteraceVariable& variable, con
...
-            // Option 2
-            for (const auto& member_decoration : variable.type_struct_info->decorations.member_decorations) {
-                member_locations[member_decoration.first] = member_decoration.second.location;
-                member_components[member_decoration.first] = member_decoration.second.component;
-            }
-        }
-        // In case of option 1, need to keep track as we go
-        uint32_t base_location = variable.decorations.location;
-
-        // Location/Components cant be decorated in nested structs, so no need to keep checking further
```
spencer-lunarg, line 1521: This is only for `Option 2`, for `Option 1` you can have a shader like

```glsl
        #version 450
        struct TestStruct {
            vec4 variableInStruct;
            vec2 dummy;
        };
        layout(location = 0) out block {
            vec2 dummy;
            TestStruct structInBlock[2];
        } testBlock;
        void main(void) {}
 ```

Tony-LunarG (approved):
LGTM and fixes the assert I saw in CTS

Issue #4881, opened 2022-11-30
Handle multiple entry point shader modules


spencer-lunarg:
# Issue

Currently a lot of Shader Validation logic is done at a whole Shader Module level which causes a shader module with multiple entry points to throw false positives.

Some VUIDs such as `VUID-vkCmdDraw-None-04115` say

> as a result of this command

which means it needs to take in consideration if it is a draw/dispatch/rayTrace/etc

https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4816 moved the logic for checking the accessible IDs into the EntryPoint scope. A proper fix will require making more use of the per-EntryPoint logic

## Related issues / PR:

https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/2450
https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4350
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/2658

Issue #3518, opened 2021-11-09
Validate bound push constant size.


rebelroad-reinhart:
Issue description: If you bind a total size of push constants less than what the pipeline expects, undefined behavior can happen at shader execution time. It would be valuable to have a validation check that the total bytes bound to push constants is what is expected by the pipeline.

Example, a fragment shader contains this push_constant declaration:

```
layout (push_constant) uniform push_constants
{
	uint resolution;
	uint count;
	uint texture_id;
	uint color; 
};
```

At run time, I can bind only 2 ints worth of push constants:

```
        uint32_t push_constants[] = { resolution, count };
        cmd.pushConstants( render_layout, vk::ShaderStageFlagBits::eFragment, 0, 
            sizeof( push_constants ), push_constants );
```
Subsequent draw calls using this pipeline will execute, but fail to warn that insufficient push constants were provided.

In this example, texture_id is an index:

```
	vec4 x = texture( textures[nonuniformEXT(texture_id)], in_uv );
```
Which will result in arbitrary memory access and undefined behavior as the contents of texture_id is undefined.

Took me several hours to find this dumb ass bug, so I figure validation would help here.

sfricke-samsung:
So the spec does say

> When a command buffer begins recording, all push constant values are undefined. Reads of undefined push constant values by the executing shader return undefined values.

So what you are doing is "valid" (the app **behaviour** is define, the **values** are undefined) but 100% agree a best-practice layer warning would be good to add 

@ncesario-lunarg since I am already in Push Constant land with https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/2711 I would be happy to take this on unless someone plans to beat me to it.

ncesario-lunarg:
Sounds good to me @sfricke-samsung! Please go ahead and unassign (or just assign to me) if that changes. Thanks!

Issue #2450, opened 2021-01-05
Validator errors that "Push constant is used in VK_SHADER_STAGE_FRAGMENT_BIT" when it's not used in the fragment shader.


seanbaxter:
In a multiple entry-point module, if the vertex shader uses a push constant, Validator errors if the fragment shader isn't marked to use a push constant in the pipeline layout, even if the fragment shader doesn't use a push constant.

```
ERROR: UNASSIGNED-CoreValidation-Shader-PushConstantOutOfRange 
 --> Validation Error: [ UNASSIGNED-CoreValidation-Shader-PushConstantOutOfRange ] Object 0: handle = 0x510000000051, type = VK_OBJECT_TYPE_SHADER_MODULE; Object 1: handle = 0x260000000026, type = VK_OBJECT_TYPE_PIPELINE_LAYOUT; | MessageID = 0x650ba96a | Push constant is used in VK_SHADER_STAGE_FRAGMENT_BIT of VkShaderModule 0x510000000051[]. But VkPipelineLayout 0x260000000026[] doesn't set VK_SHADER_STAGE_FRAGMENT_BIT.
```

[bbox_shaders.cxx.spv.zip](https://github.com/KhronosGroup/Vulkan-ValidationLayers/files/5771466/bbox_shaders.cxx.spv.zip)

```
               OpEntryPoint Vertex %_Z9bbox_vertv "_Z9bbox_vertv" %glvert_VertexIndex %_Z11shader_pushIDv4_jE %meshletDescs %object %_Z10shader_outILi0E6VertexE %scene
               OpEntryPoint Fragment %_Z9bbox_fragv "_Z9bbox_fragv" %_Z9shader_inILi0EjE %_Z10shader_outILi0EDv4_fE
```

You can see the interface of Fragment only has its in and out variables. This is the gl_vk_cadscene_meshlet sample that I ported to Circle. I'll post the application tonight, probably, if a specific repro is necessary. 

 - Linux
 - RTX 2060
 - 1.2.154

ncesario-lunarg:
Thanks for the issue @seanbaxter! A repro case might make it a bit easier to understand what's going on.

[...]

PR #5605, opened 2023-04-10
Move Push Constants to TypeStructInfo

spencer-lunarg:
This moves the old `StructInfo` to the new `TypeStructInfo`

Also creates a `PushConstantVariable` to work like the other variables

The previous code was much more complex as it was trying (and not fully correct) doing what https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/3518 is asking, but as noticed, that is for Best Practice layers as it is undefined **value** and not an actual VU

This change also caught a few tests that were breaking validation

Extra: I **think** this would be enough for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4881 and https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/2450 but might want to add more tests to confirm after this merges


tests/positive/pipeline.cpp
juan-lunarg, line 1: I'd say for now let's add this positive test. Since it seems useful. Positive testing can be added without validation.

spencer-lunarg: The tests fail right now if ran, I wrote them because I realize we weren't handling it right to start with

juan-lunarg (approved):

layers/state_tracker/shader_module.cpp
```diff
@@ -1885,11 +1685,48 @@ ResourceInterfaceVariable::ResourceInterfaceVariable(const SHADER_MODULE_STATE&
...
+        const auto& member = type_struct_info->members[i];
+        // all struct elements are required to have offset decorations in Block
+        const uint32_t memeber_offset = member.decorations->offset;
+        offset = std::min(offset, memeber_offset);
+        if (memeber_offset > highest_element_offset) {
+            highest_element_index = i;
+            highest_element_offset = memeber_offset;
+        }
+    }
+    const auto& highest_member = type_struct_info->members[highest_element_index];
```
ncesario-lunarg, line 1712: Does it matter if `highest_element_index` was never set? Should we at least assert that it was set and is >= to `type_struct_info.size()`?

layers/state_tracker/shader_module.cpp
```diff
@@ -1885,11 +1685,48 @@ ResourceInterfaceVariable::ResourceInterfaceVariable(const SHADER_MODULE_STATE&
...
 TypeStructInfo::TypeStructInfo(const SHADER_MODULE_STATE& module_state, const Instruction& struct_insn)
     : id(struct_insn.Word(1)), length(struct_insn.Length() - 2), decorations(module_state.GetDecorationSet(id)) {
-    member_ids.resize(length);
+    members.resize(length);
     for (uint32_t i = 0; i < length; i++) {
-        member_ids[i] = struct_insn.Word(2 + i);
+        Member& member = members[i];
+        member.id = struct_insn.Word(2 + i);
+        member.insn = module_state.FindDef(member.id);
+        member.type_struct_info = module_state.GetTypeStructInfo(member.insn);
```
ncesario-lunarg, line 1724: Minor nit: `resize` -> `reserve` and use `emplace_back` avoid multiple constructions.

layers/state_tracker/shader_module.cpp
```diff
@@ -1943,8 +1780,14 @@ uint32_t SHADER_MODULE_STATE::GetTypeBitsSize(const Instruction* insn) const {
             bit_size += GetTypeBitsSize(FindDef(insn->Word(i)));
         }
     } else if (opcode == spv::OpTypePointer) {
-        const Instruction* type = FindDef(insn->Word(3));
-        bit_size = GetTypeBitsSize(type);
+        if (insn->StorageClass() == spv::StorageClassPhysicalStorageBuffer) {
+            // All PhysicalStorageBuffer are just 64-bit pointers
+            // We don't need to go chasing it to find the size, as it is not calculated for any VUs
```
ncesario-lunarg, line 1785: :+1:

layers/state_tracker/shader_module.h
```diff
@@ -297,63 +307,16 @@ struct ResourceInterfaceVariable : public VariableBase {
...
-        if (!root) return false;
-        return !root->used_bytes.empty();
-    }
-
-    std::vector<uint8_t> *GetUsedbytes() const {
-        if (!root) return nullptr;
-        return &root->used_bytes;
-    }
+// vkspec.html#interfaces-resources-pushconst
+// need to be statically used in shader
```
ncesario-lunarg, line 311: "[Push constants] need to be statically used in shader"?
"[Push constants] are always OpTypeStruct and Block decorated"?

tests/CMakeLists.txt
```diff
@@ -66,6 +66,7 @@ target_sources(vk_layer_validation_tests PRIVATE
     positive/shader_compute.cpp
     positive/shader_interface.cpp
     positive/shader_limits.cpp
+    positive/shader_push_constants.cpp
```
ncesario-lunarg, line 69: So this file is already checked in, but hasn't been building :thinking:?

ncesario-lunarg (approved):
More great cleanup!

One question that I _think_ is irrelevant due to spirv-val, but I'm not 100% sure.

layers/state_tracker/shader_module.cpp
spencer-lunarg: so there is a ` assert(type_struct_info != nullptr); ` above, and `type_struct_info` has an internal check (in VVL and spirv-val) that it can never be empty

spencer-lunarg: will write a test regardless to make sure!

tests/CMakeLists.txt
spencer-lunarg: so when I split up the shader tests, I seem to forgot to add this, so for about a week we have not been testing these 6 or 7 tests

PR #5607, opened 2023-04-10
Use new Shader Module info in Shader Validation

spencer-lunarg:
With all the refactoring in Shader Module state tracking, this tries to reduce the runtime logic that was being done in Shader Validation

### DIFF
--- a/layers/state_tracker/shader_module.cpp
+++ b/layers/state_tracker/shader_module.cpp
@@ -1629,12 +1629,11 @@ StageInteraceVariable::StageInteraceVariable(const SHADER_MODULE_STATE& module_s
       builtin_block(GetBuiltinBlock(*this, module_state)),
       total_builtin_components(GetBuiltinComponents(*this, module_state)) {}
 
-ResourceInterfaceVariable::ResourceInterfaceVariable(const SHADER_MODULE_STATE& module_state, const Instruction& insn,
-                                                     VkShaderStageFlagBits stage)
-    : VariableBase(module_state, insn, stage) {
+const Instruction& ResourceInterfaceVariable::FindBaseType(ResourceInterfaceVariable& variable,
+                                                           const SHADER_MODULE_STATE& module_state) {
     // Takes a OpVariable and looks at the the descriptor type it uses. This will find things such as if the variable is writable,
     // image atomic operation, matching images to samplers, etc
-    const Instruction* type = module_state.FindDef(type_id);
+    const Instruction* type = module_state.FindDef(variable.type_id);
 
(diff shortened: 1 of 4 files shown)


layers/state_tracker/shader_module.h
```diff
@@ -247,15 +247,15 @@ struct ResourceInterfaceVariable : public VariableBase {
...
+    const Instruction &base_type;
 
     bool is_read_from{false};   // has operation to reads from the variable
     bool is_written_to{false};  // has operation to writes to the variable
 
     // Type of resource type (vkspec.html#interfaces-resources-storage-class-correspondence)
     bool is_storage_image{false};
     bool is_storage_texel_buffer{false};
-    bool is_storage_buffer{false};
+    const bool is_storage_buffer;
```
juan-lunarg, line 258: Note: This is a general comment about our codebase which is particularly notable in this file.

`bool` takes up an entire byte of space in a struct. We should be using bitfields.

https://en.cppreference.com/w/cpp/language/bit_field

This doesn't need to be fixed in this PR. I just wanted to point this out since it's particularly problematic with this struct.

layers/state_tracker/shader_module.cpp
```diff
@@ -1613,53 +1620,56 @@ StageInteraceVariable::StageInteraceVariable(const SHADER_MODULE_STATE& module_s
...
+    const bool uniform = variable.storage_class == spv::StorageClassUniform;
+    const bool buffer_block = variable.decorations.Has(DecorationSet::buffer_block_bit);
+    const bool block = variable.decorations.Has(DecorationSet::block_bit);
+    return ((uniform && buffer_block) || ((storage_buffer || physical_storage_buffer) && block));
+}
 
+ResourceInterfaceVariable::ResourceInterfaceVariable(const SHADER_MODULE_STATE& module_state, const Instruction& insn,
+                                                     VkShaderStageFlagBits stage)
+    : VariableBase(module_state, insn, stage),
+      base_type(FindBaseType(*this, module_state)),
```
juan-lunarg, line 1663: Is this safe? Concerned about having an invalid reference.

We are assigning a reference to memory owned by another class (module_state).

spencer-lunarg: yes, every `OpVariable` needs to point to something `spirv-val` would catch it first

layers/state_tracker/shader_module.h
spencer-lunarg: agree, can clean this up

juan-lunarg (approved):

ncesario-lunarg (approved):

Issue #5620, opened 2023-04-11
Replace GetDeviceProcAddr with a generated system


artem-lunarg:
Instead of calling 

```c++
 PFN_vkCreateSamplerYcbcrConversionKHR vkCreateSamplerYcbcrConversionKHR =
                reinterpret_cast<PFN_vkCreateSamplerYcbcrConversionKHR>(
                    vk::GetDeviceProcAddr(m_device->device(), "vkCreateSamplerYcbcrConversionKHR"));`
```

when a user enables the extension such as `
    AddOptionalExtensions(VK_KHR_SAMPLER_YCBCR_CONVERSION_EXTENSION_NAME);` it will automatically create these for us so that one can just call `vkCreateSamplerYcbcrConversionKHR`

To do this, the `vk.xml` looks like

```xml
<extension name="VK_KHR_sampler_ycbcr_conversion" ... supported="vulkan" promotedto="VK_VERSION_1_1">
    <require>
        <other_objects/>
        <command name="vkCreateSamplerYcbcrConversionKHR"/>
        <command name="vkDestroySamplerYcbcrConversionKHR"/>
    </require>
</extension>
```

the python will look like

```python
for ext_elem in tree.findall('extensions/extension'):
    ext_name = ext_elem.get('name')
    for type_elem in ext_elem.findall('require/command'):
        func_name = type_elem.get('name')
```

PR #5621, opened 2023-04-11
tests: Add codegen to initialize extension functions

artem-lunarg:
This PR implements codegen part.  The intention it to have a separate PR to update all tests and remove function pointer initialization (or to do this incrementally over time). Still two tests were changed to show how it works, one test for instance extension and another one for device extension.

Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5620


scripts/lvt_file_generator.py
```diff
@@ -259,9 +261,20 @@ def GenerateFunctionPointerSource(self):
             if item[1] is not None:
                 table += '#endif // %s\n' % item[1]
 
+        table += '\n// Extension function pointers\n'
+        for name in sorted(self.extensionInfo.keys()):
+            table += '// {}\n'.format(name)
```
ncesario-lunarg, line 266: General nit: I personally find the `'...'.format(...)` method much less readable than `f'// {name}'`. `.format` is used throughout the scripts for historical reasons (namely, f-strings were not supported in the past).

ncesario-lunarg (approved):
Aside from a general python nit, this looks good!

Am I correct in my understanding that these codegen additions are only used in the tests?

scripts/lvt_file_generator.py
artem-lunarg: Yes, it is only for the tests.
I'll check alternative formatting. With python it's like a Groundhog day, I faster forget it than learn.

spencer-lunarg (approved):
LGTM, all the nasty extensions I know of seem to be handled, excited to start removing more code from the tests!

PR #5626, opened 2023-04-11
cmake: Only run update_deps.py during configure step

juan-lunarg:
Remove need for `vvl_update_deps` custom target.

This fixes the bug where update_deps.py is run twice on CI for no reason. Which will help performance.

CMake will re-run the configure step when either known_good.json or update_deps.py are modified.

### DIFF
--- a/scripts/CMakeLists.txt
+++ b/scripts/CMakeLists.txt
@@ -34,11 +34,6 @@ if (UPDATE_DEPS)
         set(_build_type ${CMAKE_BUILD_TYPE})
     endif()
 
-    message("********************************************************************************")
-    message("* NOTE: Adding target vvl_update_deps to run as needed for updating            *")
-    message("*       dependencies.                                                          *")
-    message("********************************************************************************")
-
     set(optional_args)
     if (NOT BUILD_TESTS)
         set(optional_args "--optional=tests")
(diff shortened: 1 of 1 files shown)


juan-lunarg:
Note. This also fixes other minor annoyances like how running the `clean` or `clean-first` targets would also re-run `update_deps.py` for no reason. 

```
$ cmake --build build --clean-first
[1/1] Cleaning all built files...
Cleaning... 156 files.
[158/158] Linking CXX executable tests/vk_layer_validation_tests
```

scripts/CMakeLists.txt
```diff
@@ -65,46 +60,36 @@ if (UPDATE_DEPS)
...
 
-    # Add a target so that update_deps.py will run when necessary
-    # NOTE: This is triggered off of the timestamps of known_good.json and helper.cmake
-    add_custom_command(
-        OUTPUT ${_helper_file}
-        COMMAND ${PYTHON_EXECUTABLE} ${update_dep_py}
-            --dir ${_update_deps_dir} --arch ${_target_arch} --config ${_build_type} --generator "${CMAKE_GENERATOR}" ${optional_args}
-        DEPENDS ${known_good_json}
-    )
+    set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS ${update_dep_py} ${known_good_json})
```
juan-lunarg, line 66: This is the key to maintaining parity with the existing setup.

https://cmake.org/cmake/help/latest/prop_dir/CMAKE_CONFIGURE_DEPENDS.html?highlight=cmake_configure_depends

ncesario-lunarg (approved):
Neat! Although, I'm now even more afraid to touch our cmake code than before 😬

juan-lunarg:
> Neat! Although, I'm now even more afraid to touch our cmake code than before grimacing

In all fairness this kind of advanced CMake is intended for tools like `vcpkg` or `conan`. Most users should never have to know about this. `update_deps.py` is essentially a custom language package manager we made.

PR #5629, opened 2023-04-12
Split up draw time functions

spencer-lunarg:
Two spots I realize we could easily split things up

This is only just moving code (into new functions)

### DIFF
--- a/layers/core_checks/drawdispatch_validation.cpp
+++ b/layers/core_checks/drawdispatch_validation.cpp
@@ -2104,12 +2104,21 @@ const DrawDispatchVuid &CoreChecks::GetDrawDispatchVuid(CMD_TYPE cmd_type) const
 
 // Generic function to handle validation for all action commands
 // Action command == vkCmdDraw*, vkCmdDispatch*, vkCmdTraceRays*, etc.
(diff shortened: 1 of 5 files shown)


layers/core_checks/drawdispatch_validation.cpp
```diff
@@ -2178,7 +2187,7 @@ bool CoreChecks::PreCallValidateCmdDraw(VkCommandBuffer commandBuffer, uint32_t
     bool skip = false;
     auto cb_state = GetRead<CMD_BUFFER_STATE>(commandBuffer);
     skip |= ValidateCmdDrawInstance(*cb_state, instanceCount, firstInstance, CMD_DRAW);
-    skip |= ValidateActionCmd(*cb_state, false, VK_PIPELINE_BIND_POINT_GRAPHICS, CMD_DRAW);
+    skip |= ValidateActionCmd(*cb_state, VK_PIPELINE_BIND_POINT_GRAPHICS, CMD_DRAW);
```
juan-lunarg, line 2190: That reads a lot better :+1:

juan-lunarg (approved):

ncesario-lunarg (approved):

PR #5633, opened 2023-04-12
layers: Create common Plane Aspect Mask helper

spencer-lunarg:
recently I got a new section [added in the spec](https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5816) to unify the logic which Aspect Masks should be with which multi-planar formats. Sometimes the spec says "a single valid" and sometimes it says "at least one"

These new helper functions clear that up

Also before there were 2 VUs (one for 2 plane, one for 3 plane) but now it is a single VU


layers/utils/vk_layer_utils.h
```diff
@@ -351,6 +351,27 @@ static inline uint32_t GetPlaneIndex(VkImageAspectFlags aspect) {
     }
 }
 
+// vkspec.html#formats-planes-image-aspect
+static inline bool IsAnyValidPlaneAspect(VkFormat format, VkImageAspectFlags aspect_mask) {
```
artem-lunarg, line 355: I'm curious about `static inline` syntax. One explanation I heard, that in case of inlining failure, `static` is needed to prevent multiple definitions. Not sure if it's correct. It assumes that two aspects of `inline` keyword, a) code inlinining and b) allowing multiple definitions, work in sync. My thesis is that `inline` keyword allows multiple definitions regardless whether code inlining is successful or not. If that's true, then `static` is not needed to prevent multiple definitions. My impression it's true in C++ (it might be `static inline` is needed in C). I don't have some solid proofs or  or C++ spec knowledge to claim it's true. Personally never had issue with just `inline` and also saw suggestion on google board to replace static inline with just inline. Maybe we have some platform/compiler where `static inline` makes a diference?

layers/utils/vk_layer_utils.h
```diff
@@ -351,6 +351,27 @@ static inline uint32_t GetPlaneIndex(VkImageAspectFlags aspect) {
...
+    } else if ((planes == 3) && ((aspect_mask & valid_mask_3_planes) != 0)) {
+        return true;
+    } else {
+        return false;  // Expects calls to make sure it is a multi-planar format
+    }
+}
+
+static inline bool IsOnlyOneValidPlaneAspect(VkFormat format, VkImageAspectFlags aspect_mask) {
+    // Check if more then one bit is set
+    const bool multiple_bits = (aspect_mask & (aspect_mask - 1)) != 0;
```
artem-lunarg, line 371: nit: `const bool multiple_bits = aspect_mask != 0 && !IsPowerOfTwo(aspect_mask)` - might be simpler? or not?

layers/core_checks/descriptor_validation.cpp
```diff
@@ -3325,17 +3325,15 @@ bool CoreChecks::VerifyWriteUpdateContents(const DescriptorSet *dest_set, const
...
                             return false;
                         }
-                        // image view need aspect mask for only the planes supported of format
-                        VkImageAspectFlags legal_aspect_flags = (VK_IMAGE_ASPECT_PLANE_0_BIT | VK_IMAGE_ASPECT_PLANE_1_BIT);
-                        legal_aspect_flags |=
-                            (FormatPlaneCount(image_state->createInfo.format) == 3) ? VK_IMAGE_ASPECT_PLANE_2_BIT : 0;
-                        if (0 != (iv_state->create_info.subresourceRange.aspectMask & (~legal_aspect_flags))) {
+                        const VkFormat image_format = image_state->createInfo.format;
+                        const VkImageAspectFlags image_aspect = iv_state->create_info.subresourceRange.aspectMask;
+                        if (!IsAnyValidPlaneAspect(image_format, image_aspect)) {
```
artem-lunarg, line 3330: It looks that for 2-plane format (PLANE_0 + PLANE_1) and aspect mask (PLANE_0 | PLANE_2) the original code triggered error (PLANE_2 is not part of format). With this change `IsAnyValidPlaneAspect` returns true because PLANE_0 is part of format, so it won't trigger the error.

layers/utils/vk_layer_utils.h
ncesario-lunarg: Perhaps https://en.cppreference.com/w/cpp/language/inline helps?

My understanding is that _without_ `inline` here one could not reliably have the definition of the static function here in the header (i.e., it would need to live in a single translation unit).

layers/core_checks/descriptor_validation.cpp
spencer-lunarg: good point!!... let me get a test for this and then fix

layers/utils/vk_layer_utils.h
spencer-lunarg: I took this from https://graphics.stanford.edu/~seander/bithacks.html#DetermineIfPowerOf2

But honestly happy to make it more human readable with the `IsPowerOfTwo` function

artem-lunarg: I guess my suggestion was that proposed expression might be easier to understand even for people who did not dig into the underlying mechanism of (x & (x-1)) operation. IsPowerOfTwo() expression requires only knowledge that powerof2 value has only one bit set, but with (x & (x-1)) I still had to think (to ensure that edge condition x != 0 should indeed be removed), even I know how the underlying powerof2 arithmetic works.

layers/core_checks/synchronization_validation.cpp
```diff
@@ -1665,25 +1665,15 @@ bool CoreChecks::ValidateBarriersToImages(const Location &outer_loc, const CMD_B
...
-        }
-
-        if ((FormatPlaneCount(image_format) == 2) && ((aspect_mask & VK_IMAGE_ASPECT_PLANE_2_BIT) != 0)) {
-            const auto &vuid = GetImageBarrierVUID(image_loc, ImageError::kBadMultiplanarAspect);
-            skip |= LogError(img_barrier.image, vuid,
-                             "%s references %s of format %s has only two planes but included "
-                             "VK_IMAGE_ASPECT_PLANE_2_BIT in its aspectMask (0x%" PRIx32 ").",
-                             image_loc.Message().c_str(), report_data->FormatHandle(img_barrier.image).c_str(),
-                             string_VkFormat(image_format), aspect_mask);
+        if ((FormatIsMultiplane(image_format)) && (image_state->disjoint == true)) {
```
artem-lunarg, line 1668: It's not obvious that new condition is equivalent to the previous two. One of them was without `disjoint` check. Could you please confirm/check this.

spencer-lunarg: these are both for `VUID-VkImageMemoryBarrier-image-01672`

> If image has a multi-planar format and the image is disjoint, then the aspectMask member of subresourceRange must include at least one [multi-planar aspect mask](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#formats-planes-image-aspect) or VK_IMAGE_ASPECT_COLOR_BIT

so seems that this was a bug before and it should have been checking `disjoint`

layers/utils/vk_layer_utils.h
spencer-lunarg: took your suggestion for this

layers/core_checks/descriptor_validation.cpp
spencer-lunarg: created tests (and file for them) and re-worked logic to be correct

tests/positive/layer_utils.cpp
```diff
@@ -0,0 +1,111 @@
+
+/*
+ * Copyright (c) 2015-2023 The Khronos Group Inc.
+ * Copyright (c) 2015-2023 Valve Corporation
+ * Copyright (c) 2015-2023 LunarG, Inc.
+ * Copyright (c) 2015-2023 Google, Inc.
```
juan-lunarg, line 6:
```suggestion
 * Copyright (c) 2023 Valve Corporation
 * Copyright (c) 2023 LunarG, Inc.
```

juan-lunarg (approved):

artem-lunarg (approved):

PR #5636, opened 2023-04-12
cmake: Fix vk_enum_string_helper.h installation

juan-lunarg:
- cmake: Fix `vk_enum_string_helper.h` installation
- ci: Add integration tests for installation

### DIFF
--- a/layers/CMakeLists.txt
+++ b/layers/CMakeLists.txt
@@ -28,8 +28,6 @@ target_sources(VkLayer_utils PRIVATE
     generated/vk_dispatch_table_helper.h
     generated/vk_safe_struct.h
     generated/vk_safe_struct.cpp
-    # NOTE: This header file is installed by the LunarG SDK and treated as part of VulkanHeaders
-    # IE: This header file is also a deliverable of VVL.
     generated/vk_enum_string_helper.h
     generated/vk_object_types.h
     generated/vk_extension_helper.h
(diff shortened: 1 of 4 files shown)


layers/CMakeLists.txt
```diff
@@ -82,9 +79,6 @@ if(BUILD_LAYER_SUPPORT_FILES)
                   ${CMAKE_SOURCE_DIR}/layers/generated/vk_dispatch_table_helper.h
                   ${CMAKE_SOURCE_DIR}/layers/generated/vk_safe_struct.h
                   ${CMAKE_SOURCE_DIR}/layers/generated/vk_safe_struct.cpp
-                  # NOTE: This header file is installed by the LunarG SDK and treated as part of VulkanHeaders
-                  # IE: This header file is also a deliverable of VVL.
-                  ${CMAKE_SOURCE_DIR}/layers/generated/vk_enum_string_helper.h
```
juan-lunarg, line 87: Note by removing this I'll have to update LunarG/VulkanTools and Vulkan-Profiles.

Technically we could just install the file twice.

Thoughts?

@lunarpapillo / @ncesario-lunarg

lunarpapillo: The plan is to eventually (hopefully, soon) move `vk_enum_string_helper.h` to **Vulkan-Utility-Library**, correct?  Until that can be done, whatever "hacks" that are the easiest to keep things working are fine.

juan-lunarg: Sounds good. Installing it twice to avoid fixing up tools/profiles again.

tests/integration/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+target_include_directories(foobar PRIVATE
+    ${VVL_INSTALL_DIR}/include/
+    ${VVL_INSTALL_DIR}/include/vulkan/
+)
+
+target_sources(foobar PRIVATE main.cpp)
+
+# LunarG/VulkanTools expects source files in specific places for Android/Layer-Factory usage.
+if (NOT EXISTS "${VVL_SOURCE_DIR}/layers/external/xxhash.cpp")
+    message(FATAL_ERROR "Update LunarG/VulkanTools")
```
lunarpapillo, line 62: I'd like each of these messages to name the missing file, e.g. something like:
```
    message(FATAL_ERROR "xxhash.cpp is missing - update LunarG/VulkanTools")
```
I'd also like _all_ the tests to be exercised each build, rather than stop at the first error, so the full scope of which files are missing/moved is known before the update effort occurs.

But if either of those is untenable or difficult, what is here is acceptable and far better than what we had before.

lunarpapillo (approved):
I'm not a "real" reviewer, but this looks pretty great to me.  It's quite thorough and provides checking that we didn't have before.

artem-lunarg (approved):
LGTM

PR #5637, opened 2023-04-12
android: Remove usage of VALIDATION_APK

juan-lunarg:


juan-lunarg:
```
jni/../../tests/framework/render.h:32:10: fatal error: 'android_native_app_glue.h' file not found
#include <android_native_app_glue.h>
```

mark-lunarg:
There's no description, can a summary of the reasoning behind the VALIDATION_APK changes be provided please?

juan-lunarg:
> There's no description, can a summary of the reasoning behind the VALIDATION_APK changes be provided please?

I talked to @spencer-lunarg about it. I couldn't determine the rational for the VALIDATION_APK define. So I would prefer to remove it. Since I don't want things in the main CMake build I don't understand.

mark-lunarg:
In `e88fb5310f` @locke-lunarg indicated that this was used to keep some tests from running outside of an APK when a valid Android window surface can be created.  Perhaps this isn't necessary any more?

juan-lunarg:
Looking at the build code I didn't see a way to toggle `VALIDATION_APK` so I figured it was best to just remove it entirely.

spencer-lunarg:
ok, I understand what is going on here now @juan-lunarg 

So when it comes to Android you have two options

1. Run a binary linux style `adb shell '/path/to/binary'`
2. Build and run an APK (an app)

So the issue with `1.` is you can't get the `ANativeWindow` handle needed for `vkCreateAndroidSurfaceKHR`. Android gives it to you when the app starts (because you are now the foreground app running on the phone) via the `android_native_app_glue.h` helper file (you can see this if you search for `APP_CMD_INIT_WINDOW`

So I this was added 4 years ago, but to my knowledge, we only build the Tests as an APK now, which means we shouldn't have to worry about number `1.` anymore... so we **should** be able to remove this now

As for the missing file error, here is some cmake from a personal project I have how I built it

```cmake
# build native_app_glue as a static lib
set(APP_GLUE_DIR ${ANDROID_NDK}/sources/android/native_app_glue)
include_directories(${APP_GLUE_DIR})
add_library(native_glue STATIC ${APP_GLUE_DIR}/android_native_app_glue.c)

target_link_libraries(project_name  native_glue)
```

juan-lunarg:
I found a CMake module in the NDK called `AndroidNdkModules.cmake`

/home/juan/Android/Sdk/cmake/3.22.1/share/cmake-3.22/Modules/AndroidNdkModules.cmake

```
macro(android_ndk_import_module_cpufeatures)
    if(ANDROID)
        include_directories(${ANDROID_NDK}/sources/android/cpufeatures)
        add_library(cpufeatures ${ANDROID_NDK}/sources/android/cpufeatures/cpu-features.c)
        target_link_libraries(cpufeatures dl)
    endif()
endmacro()

macro(android_ndk_import_module_native_app_glue)
    if(ANDROID)
        include_directories(${ANDROID_NDK}/sources/android/native_app_glue)
        add_library(native_app_glue ${ANDROID_NDK}/sources/android/native_app_glue/android_native_app_glue.c)
        target_link_libraries(native_app_glue log)
    endif()
endmacro()

macro(android_ndk_import_module_ndk_helper)
    if(ANDROID)
        android_ndk_import_module_cpufeatures()
        android_ndk_import_module_native_app_glue()
        
        include_directories(${ANDROID_NDK}/sources/android/ndk_helper)
        file(GLOB _NDK_HELPER_SRCS ${ANDROID_NDK}/sources/android/ndk_helper/*.cpp ${ANDROID_NDK}/sources/android/ndk_helper/gl3stub.c)
        add_library(ndk_helper ${_NDK_HELPER_SRCS})
        target_link_libraries(ndk_helper log android EGL GLESv2 cpufeatures native_app_glue)
        
        unset(_NDK_HELPER_SRCS)
    endif()
endmacro()
```

Looks like this is an unofficial CMake module created by the Android NDK:

Thoughts, @spencer-lunarg ?

spencer-lunarg:
I am in favor of using `android_ndk_import_module_native_app_glue` and still think we can remove `VALIDATION_APK`

PR #5643, opened 2023-04-13
gpu: Add descriptor set and binding numbers to GPU-AV errors

### LINKED ISSUE #5495: Request for clearer GPU assisted validation messages
**Environment:**
 - OS: Windows 11
 - GPU: Nvidia RTX 2080 Super
 - SDK or header version if building from repo: 1.3.216.0
 - Options enabled (synchronization, best practices, etc.): GPU assisted validation 

**Describe the Issue**

The GPU assisted validation's error message says I have an error at "descriptor index 0"

**Expected behavior**

The GPU assisted validation layer should say that I have an error at "descriptor set=0 binding=1"

**Valid Usage ID**
```
[2023-03-23 09:12:39.171] [error] [ERROR: Validation](user defined)
Validation Error: [ VUID-vkCmdDispatch-None-02706 ] Object 0: handle = 0x228199a3510, type = VK_OBJECT_TYPE_QUEUE; | MessageID = 0x9d0dde06 | Descriptor index 0 access out of bounds. Descriptor size is 524288 and highest byte accessed was 527363 Command buffer (0x2285f6c4600). Compute Index 0x34. Pipeline (Fine binning)(0xe9b2ee0000000094). Shader Module (0xd68cbb0000000091). Shader Instruction Index = 433.  Stage = Compute.  Global invocation ID (x, y, z) = (768, 0, 8 ) Shader validation error occurred in file D:/Source/SahRenderer/Windows/../RenderCore/shaders/voxelizer/binning_fine.comp at line 247.Unable to find suitable #line directive in SPIR-V OpSource. The Vulkan spec states: If the robust buffer access feature is not enabled, and if the VkPipeline object bound to the pipeline bind point used by this command accesses a storage buffer, it must not access values outside of the range of the buffer as specified in the descriptor set bound to the same pipeline bind point (https://vulkan.lunarg.com/doc/view/1.3.216.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdDispatch-None-02706)
```

**Additional context**

I enabled GPU assisted validation on my project and encountered the above error. The error states "descriptor index 0 access out of bounds". This was very confusing - I only have one descriptor set for this dispatch, and descriptor index 0 in this set has a size of 1179648! It appeared that the GPU assisted validation layer was simply wrong

After some debugging I realized that it was referring to descriptor index _1_ inside of descriptor set 0. This was not clear at all from the error. The error message should clearly state which descriptor in which set had the out-of-bounds access, or at least be clear that it's talking about descriptor _sets_ and not descriptors themselves

Additional context in the Vulkan Discord server: https://discord.com/channels/427551838099996672/427558253275250718/1088511714477232129

jeremyg-lunarg:
The descriptor set and binding numbers are now part of the error record produced by the GPU-AV SPIR-V instrumentation. Add it to the error messages so that it is easier to find which descriptor needs to be fixed.
    
Fixes #5495
    
NOTE: This change requires SPIRV-Tools to be updated

### DIFF
--- a/build-android/known_good.json
+++ b/build-android/known_good.json
@@ -22,7 +22,7 @@
             "name": "SPIRV-Tools",
             "url": "https://github.com/KhronosGroup/SPIRV-Tools.git",
             "sub_dir": "shaderc/third_party/spirv-tools",
-            "commit": "d4c0abdcad60325a2ab3c00a81847e2dbdc927a2"
+            "commit": "0ce36ad7856e15b545bc7b9fc9a7c49671b40f96"
         },
         {
             "name": "SPIRV-Headers",
--- a/scripts/known_good.json
+++ b/scripts/known_good.json
@@ -37,7 +37,7 @@
                 "-DSPIRV-Headers_SOURCE_DIR={repo_dir}/../SPIRV-Headers",
                 "-DSPIRV_WERROR=OFF"
             ],
-            "commit": "d4c0abdcad60325a2ab3c00a81847e2dbdc927a2"
+            "commit": "0ce36ad7856e15b545bc7b9fc9a7c49671b40f96"
         },
         {
             "name": "robin-hood-hashing",
--- a/layers/generated/spirv_tools_commit_id.h
+++ b/layers/generated/spirv_tools_commit_id.h
@@ -22,4 +22,4 @@
  ****************************************************************************/
 #pragma once
 
-#define SPIRV_TOOLS_COMMIT_ID "d4c0abdcad60325a2ab3c00a81847e2dbdc927a2"
+#define SPIRV_TOOLS_COMMIT_ID "0ce36ad7856e15b545bc7b9fc9a7c49671b40f96"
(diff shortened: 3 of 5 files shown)


scripts/known_good.json
juan-lunarg, line 1: Should also update spirv-headers along with this.

https://github.com/KhronosGroup/SPIRV-Tools/blob/main/DEPS

Also I fixed up SPIRV-Headers so we can remove the fix I added:

```patch
+            "commit": "cfbe4feef20c3c0628712c2792624f0221e378ac",
-            "cmake_options": [
-                "-D SPIRV_HEADERS_SKIP_EXAMPLES=ON"
-            ]
```

jeremyg-lunarg: ok.  This change has to wait until my SPIRV-Tools changes merge and then I'll have to touch this part again.

build-android/known_good.json
juan-lunarg, line 1: Does this PR have new enough version of the SPIRV-* repos for your PR?
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5750

mikes-lunarg: Related question: When I updated SPIRV-Tools to ToT, the GPUAV tests started hanging/failing and so I had to back off to a commit before https://github.com/KhronosGroup/SPIRV-Tools/commit/d4c0abdcad60325a2ab3c00a81847e2dbdc927a2. Is that expected?

jeremyg-lunarg: This change has a hard dependency on the latest SPIRV-Tools. There's not backwards compatibility so it would break if ran latest SPIRV-Tools with VVL without this change.

Tony-LunarG (approved):
Very nice!  LGTM

PR #5644, opened 2023-04-13
bp: Optimize best practices for return codes

juan-lunarg:
Utilize the fact positive return codes are greater than 0. And error codes are less than 0.

### DIFF
--- a/layers/best_practices/best_practices_utils.cpp
+++ b/layers/best_practices/best_practices_utils.cpp
@@ -816,31 +816,36 @@ void BestPractices::ManualPostCallRecordAllocateMemory(VkDevice device, const Vk
                                                        const VkAllocationCallbacks* pAllocator, VkDeviceMemory* pMemory,
                                                        VkResult result) {
     if (result != VK_SUCCESS) {
-        constexpr std::array error_codes = {VK_ERROR_OUT_OF_HOST_MEMORY, VK_ERROR_OUT_OF_DEVICE_MEMORY, VK_ERROR_TOO_MANY_OBJECTS,
-                                            VK_ERROR_INVALID_EXTERNAL_HANDLE, VK_ERROR_INVALID_OPAQUE_CAPTURE_ADDRESS};
-        ValidateReturnCodes("vkAllocateMemory", result, error_codes, {});
+        ValidateReturnCodes("vkAllocateMemory", result);
         return;
     }
 }
 
-void BestPractices::ValidateReturnCodes(const char* api_name, VkResult result, vvl::span<const VkResult> error_codes,
-                                        vvl::span<const VkResult> success_codes) const {
-    auto error = std::find(error_codes.begin(), error_codes.end(), result);
-    if (error != error_codes.end()) {
-        constexpr std::array common_failure_codes = {VK_ERROR_OUT_OF_DATE_KHR, VK_ERROR_FULL_SCREEN_EXCLUSIVE_MODE_LOST_EXT};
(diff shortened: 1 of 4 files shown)


scripts/best_practices_generator.py
```diff
@@ -275,6 +275,7 @@ def genCmd(self, cmdinfo, cmdname, alias):
             # Treat empty string as 'None' for consistency with 'error_codes'
             if success_codes == '':
                 success_codes = None
+            # This function only returns VK_SUCCESS
```
juan-lunarg, line 278: Example: `vkResetDescriptorPool` only returns `VK_SUCCESS` currently.

https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkResetDescriptorPool.html

juan-lunarg:
> Utilize the fact positive return codes are greater than 0. And error codes are less than 0.

Credit to @charles-lunarg for pointing this out during our latest meeting.

layers/best_practices/best_practices_utils.cpp
```diff
@@ -816,31 +816,34 @@ void BestPractices::ManualPostCallRecordAllocateMemory(VkDevice device, const Vk
...
     }
-    auto success = std::find(success_codes.begin(), success_codes.end(), result);
-    if (success != success_codes.end()) {
-        LogInfo(instance, kVUID_BestPractices_NonSuccess_Result, "%s(): Returned non-success return code %s.", api_name,
-                string_VkResult(result));
+
+    assert(result < VK_SUCCESS);  // At this point we have an error
+
+    // Despite being error codes log these results as informational.
+    constexpr std::array common_failure_codes = {VK_ERROR_OUT_OF_DATE_KHR, VK_ERROR_FULL_SCREEN_EXCLUSIVE_MODE_LOST_EXT};
```
juan-lunarg, line 840: Why do we treat these 2 error codes special btw? I'd like to add a detailed comment or reference a link.

charles-lunarg: That is because they are returned *frequently* during window resizing. While they are errors, as once you get this error code you must recreate the swapchain, they are not considered errors by BP because they are expected to occur during the normal application lifecycle.

charles-lunarg (approved):

layers/best_practices/best_practices_utils.cpp
```diff
@@ -812,35 +812,31 @@ bool BestPractices::PreCallValidateAllocateMemory(VkDevice device, const VkMemor
...
-void BestPractices::ManualPostCallRecordAllocateMemory(VkDevice device, const VkMemoryAllocateInfo* pAllocateInfo,
-                                                       const VkAllocationCallbacks* pAllocator, VkDeviceMemory* pMemory,
-                                                       VkResult result) {
-    if (result != VK_SUCCESS) {
-        constexpr std::array error_codes = {VK_ERROR_OUT_OF_HOST_MEMORY, VK_ERROR_OUT_OF_DEVICE_MEMORY, VK_ERROR_TOO_MANY_OBJECTS,
-                                            VK_ERROR_INVALID_EXTERNAL_HANDLE, VK_ERROR_INVALID_OPAQUE_CAPTURE_ADDRESS};
-        ValidateReturnCodes("vkAllocateMemory", result, error_codes, {});
-        return;
-    }
-}
```
juan-lunarg, line 824: Second commit removes this function. No reason have a manual post call for vkAllocateMemory anymore.

It used to track `num_mem_objects` but it doesn't do that anymore.

jeremyg-lunarg (approved):
LGTM

PR #5646, opened 2023-04-14
Descriptors use ShaderModule Variable object

spencer-lunarg:
Before we were going:

1. Parse shader object
2. Copy info about the Shader to the Pipeline object
3. Have the Descriptors us the Pipeline object

This just skips the step 2

This can be seen in `PIPELINE_STATE::GetActiveSlots` where we not just give it the `ResourceInterfaceVariable` and not do all the "re-copy what found"

I also redefined `NUMERIC_TYPE` - There is an internal spec PR I have that will properly spec that

>  "Numeric Type" == Float vs SInt vs UInt

### DIFF
--- a/layers/state_tracker/pipeline_state.cpp
+++ b/layers/state_tracker/pipeline_state.cpp
@@ -135,33 +135,7 @@ PIPELINE_STATE::ActiveSlotMap PIPELINE_STATE::GetActiveSlots(const StageStateVec
         // Capture descriptor uses for the pipeline
         for (const auto &variable : stage.entrypoint->resource_interface_variables) {
             // While validating shaders capture which slots are used by the pipeline
-            auto &entry = active_slots[variable.decorations.set][variable.decorations.binding];
-            entry.is_written_to |= variable.is_written_to;
-
-            auto &reqs = entry.reqs;
-            reqs |= stage.module_state->DescriptorTypeToReqs(variable.type_id);
-            if (variable.is_atomic_operation) reqs |= DESCRIPTOR_REQ_VIEW_ATOMIC_OPERATION;
-            if (variable.is_sampler_sampled) reqs |= DESCRIPTOR_REQ_SAMPLER_SAMPLED;
-            if (variable.is_sampler_implicitLod_dref_proj) reqs |= DESCRIPTOR_REQ_SAMPLER_IMPLICITLOD_DREF_PROJ;
-            if (variable.is_sampler_bias_offset) reqs |= DESCRIPTOR_REQ_SAMPLER_BIAS_OFFSET;
-            if (variable.is_read_without_format) reqs |= DESCRIPTOR_REQ_IMAGE_READ_WITHOUT_FORMAT;
-            if (variable.is_write_without_format) reqs |= DESCRIPTOR_REQ_IMAGE_WRITE_WITHOUT_FORMAT;
-            if (variable.is_dref_operation) reqs |= DESCRIPTOR_REQ_IMAGE_DREF;
-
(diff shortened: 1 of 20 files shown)


layers/core_checks/descriptor_validation.cpp
```diff
@@ -924,50 +894,80 @@ bool CoreChecks::ValidateDescriptor(const DescriptorContext &context, const Desc
                         report_data->FormatHandle(image_view).c_str());
     }
     if (image_view) {
-        const auto reqs = binding_info.second.reqs;
+        const auto &variable = *binding_info.second;
         const auto &image_view_ci = image_view_state->create_info;
 
-        if (reqs & DESCRIPTOR_REQ_ALL_VIEW_TYPE_BITS) {
```
spencer-lunarg, line 930: Before we were using the bit logic, but it made it not easy to understand we were trying to avoid `SubpassData`

the goal of this VU is to check 

image

layers/state_tracker/pipeline_state.cpp
```diff
@@ -24,11 +24,7 @@
 PipelineStageState::PipelineStageState(const safe_VkPipelineShaderStageCreateInfo *create_info,
                                        std::shared_ptr<const SHADER_MODULE_STATE> &module_state,
                                        std::shared_ptr<const SHADER_MODULE_STATE::EntryPoint> &entrypoint)
-    : module_state(module_state), create_info(create_info), entrypoint(entrypoint) {
-    if (entrypoint) {
-        descriptor_variables = &(*entrypoint).resource_interface_variables;
```
spencer-lunarg, line 29: this was the redundant copy we had before

layers/state_tracker/pipeline_state.h
```diff
@@ -201,9 +150,9 @@ class PIPELINE_STATE : public BASE_NODE {
     const vvl::unordered_set<uint32_t> fragmentShader_writable_output_location_list;
 
     // Capture which slots (set#->bindings) are actually used by the shaders of this pipeline
-    using ActiveSlotMap = vvl::unordered_map<uint32_t, BindingReqMap>;
+    using ActiveSlotMap = vvl::unordered_map<uint32_t, BindingVariableMap>;
     // NOTE: this map is 'almost' const and used in performance critical code paths.
-    // The values of existing entries in the DescriptorRequirement::samplers_used_by_image map
+    // The values of existing entries in the samplers_used_by_image map
```
spencer-lunarg, line 155: `samplers_used_by_image` should just be const in `PIPELINE_STATE` now, follow up PR can look into making the map const

layers/state_tracker/shader_module.h
```diff
@@ -175,6 +175,16 @@ struct InterfaceSlot {
     uint32_t GetSlotValue(uint32_t location, uint32_t component) { return (location * 4) + component; }
 };
 
+// Represents the Image formats that can map to a SPIR-V format
+enum NUMERIC_TYPE {
+    NUMERIC_TYPE_UNKNOWN = 0,  // In case image is not used
+    NUMERIC_TYPE_FLOAT = 1,    // UNORM, SNORM, FLOAT, USCALED, SSCALED, SRGB -- anything we consider float in the shader
+    NUMERIC_TYPE_SINT = 2,
+    NUMERIC_TYPE_UINT = 4,
+};
```
ncesario-lunarg, line 184: These should probably be [CamelCase](https://google.github.io/styleguide/cppguide.html#Enumerator_Names).

ncesario-lunarg (approved):
Minor style nit, but LGTM!

PR #5653, opened 2023-04-14
scripts: Remove the need to set PYTHONPATH everywhere

jeremyg-lunarg:
The scripts all need to know the registry directory already, so they can update sys.path to include it.

### DIFF
--- a/docs/generated_code.md
+++ b/docs/generated_code.md
@@ -7,23 +7,21 @@ How to generate the code:
 
 - Linux:
 ```bash
-PYTHONPATH=$PYTHONPATH:$PWD/external/Vulkan-Headers/registry python3 scripts/generate_source.py external/Vulkan-Headers/registry/ external/SPIRV-Headers/include/spirv/unified1/
+scripts/generate_source.py external/Vulkan-Headers/registry/ external/SPIRV-Headers/include/spirv/unified1/
 ```
 
 - Windows Powershell:
 ```powershell
-pwsh -Command { $env:PYTHONPATH+=";$PWD/external/Vulkan-Headers/registry"; python3 scripts/generate_source.py external/Vulkan-Headers/registry/ external/SPIRV-Headers/include/spirv/unified1/ }
+pwsh -Command { python3 scripts/generate_source.py external/Vulkan-Headers/registry/ external/SPIRV-Headers/include/spirv/unified1/ }
 ```
 
 - Windows Command:
 ```cmd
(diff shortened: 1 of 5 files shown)


scripts/generate_source.py
```diff
@@ -41,6 +40,10 @@ def main(argv):
     group.add_argument('-v', '--verify', action='store_true', help='verify repo files match generator output')
     args = parser.parse_args(argv)
 
+    # We need modules from the registry directory, add it here so no one has to set it in PYTHONPATH
+    sys.path.insert(0, args.registry)
+    import common_codegen
```
ncesario-lunarg, line 45: I understand the convenience of this change, but this is what I was trying to avoid. i.e., this import statement is now a "magic `#include`" that must occur in a certain order. I agree this "should be fine" for us going forward, but it's a bit of a hack I was trying to avoid. I'm good with it if this is what others prefer.

FYI @aqnuep in case any of your refactoring or downstream scripts are affected.

juan-lunarg: Perhaps @lunarpapillo has an idea?

docs/generated_code.md
```diff
@@ -7,24 +7,22 @@ How to generate the code:
...
 - Linux:
 ```bash
-PYTHONPATH=$PYTHONPATH:$PWD/external/Vulkan-Headers/registry python3 scripts/generate_source.py external/Vulkan-Headers/registry/ external/SPIRV-Headers/include/spirv/unified1/
+scripts/generate_source.py external/Vulkan-Headers/registry/ external/SPIRV-Headers/include/spirv/unified1/
 ```
 
 - Windows Powershell:
 ```powershell
-pwsh -Command { $env:PYTHONPATH+=";$PWD/external/Vulkan-Headers/registry"; python3 scripts/generate_source.py external/Vulkan-Headers/registry/ external/SPIRV-Headers/include/spirv/unified1/ }
+pwsh -Command { python3 scripts/generate_source.py external/Vulkan-Headers/registry/ external/SPIRV-Headers/include/spirv/unified1/ }
```
arno-lunarg, line 15:
```suggestion
python3 scripts/generate_source.py external/Vulkan-Headers/registry/ external/SPIRV-Headers/include/spirv/unified1/
```

artem-lunarg: The suggestion is to use `python` command instead of `python3` on Windows.
https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/b4e93d11f6940a1fb8cc0c0ce432db6182ba9bae

juan-lunarg (approved):
LGTM less environment variables the better

ncesario-lunarg (approved):

Issue #5658, opened 2023-04-16
VUID-VkGraphicsPipelineCreateInfo-pStages-06896 is validated incorrectly


Rua:
This VUID says:
> If the pipeline is being created with pre-rasterization shader state, all elements of `pStages` **must** have a stage set to a shader stage which participates in fragment shader state or pre-rasterization shader state

In layman's terms, this is saying that all the provided shaders must be graphics shaders, they can't be e.g. compute or raytracing shaders. But the actual code is testing something else:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/ddd3f7ee9d6d87307e3e85e8860d085880998068/layers/core_checks/pipeline_cc_validation.cpp#L1527-L1532

This is testing whether at least one pre-raster shader has been provided, which is very different. This would not catch the user providing a compute shader, for example. Fragment shaders are also not included in the check.

spencer-lunarg:
Before I write a test to see if this works, I have to believe there is another VU that would blow up first

Did you hit this VU in some code or are you just bringing up this from eye-ing the code (both is valid)

Rua:
I got it from looking at the code itself.

spencer-lunarg:
Tried to write a test, it is not possible to hit this, we are already guarding against things like a compute stage in the `VkGraphicsPipelineCreateInfo::pStages`

Also we have tests for this VU so seems the coverage is working as expected

spencer-lunarg:
> Fragment shaders are also not included in the check.

this is a spec bug, getting it fixed now

PR #5661, opened 2023-04-17
layers: Fix logic checking SubStage stages

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5658

As noted in issue, the logic before was checking if "at least one" stage was valid, but really we should be checking if "all stages" are valid


juan-lunarg (changes requested):
Additional testing required.

ncesario-lunarg (commented):
LGTM once comments are resolved.

spencer-lunarg:
this has got stale, closing for now, will try again once have time for GPL again

PR #5663, opened 2023-04-17
layers: Remove duplicated if in FragmentOutputState

ziga-lunarg:
Remove redundant if


artem-lunarg (approved):
LGTM. Good catch, this might also free one branch prediction slot 🔀

juan-lunarg (approved):

spencer-lunarg (approved):

Issue #5624, opened 2023-04-11
Add validation for VUIDs mentioning VkAttachmentReferenceStencilLayout 


arno-lunarg:
**Valid Usage IDs requested**
VUID-VkRenderPassCreateInfo2-attachment-06244
VUID-VkRenderPassCreateInfo2-attachment-06245
VUID-VkRenderPassCreateInfo2-attachment-06246
VUID-vkCmdBeginRenderPass-stencilInitialLayout-02843
VUID-vkCmdBeginRenderPass2-stencilInitialLayout-02845
VUID-vkCmdBeginRenderPass-initialLayout-02842
VUID-vkCmdBeginRenderPass2-initialLayout-02844

**Additional Context**
Follow-up for issue https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5608

PR #5664, opened 2023-04-17
layers: Add `VK_KHR_separate_depth_stencil_layouts` related VUIDs

arno-lunarg:
- Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5624

Add:
VUID-VkRenderPassCreateInfo2-attachment-06244
VUID-VkRenderPassCreateInfo2-attachment-06245
VUID-VkRenderPassCreateInfo2-attachment-06246
VUID-vkCmdBeginRenderPass-stencilInitialLayout-02843
VUID-vkCmdBeginRenderPass2-stencilInitialLayout-02845
VUID-vkCmdBeginRenderPass-initialLayout-02842
VUID-vkCmdBeginRenderPass2-initialLayout-02844

Two new tests: `RenderPassBeginLayoutsStencilBufferImageUsageMismatches` and `RenderPassBeginInvalidStencilFormat`
Split up `RenderPassCreateAttachmentReferenceInvalidLayout` into two tests because the `if (separate_depth_stencil_layouts_features.separateDepthStencilLayouts) { ... }` logic was not working as intended anymore, and one branch was dead. There is now one test for each of the previous branches.

### DIFF
--- a/layers/core_checks/image_layout_cc_validation.cpp
+++ b/layers/core_checks/image_layout_cc_validation.cpp
@@ -477,7 +477,6 @@ bool CoreChecks::ValidateRenderPassLayoutAgainstFramebufferImageUsage(RenderPass
     const auto &image_view = image_view_state.Handle();
     const auto *image_state = image_view_state.image_state.get();
     const auto &image = image_state->Handle();
-    const char *vuid;
     const bool use_rp2 = (rp_version == RENDER_PASS_VERSION_2);
     const char *function_name = use_rp2 ? "vkCmdBeginRenderPass2()" : "vkCmdBeginRenderPass()";
 
@@ -485,7 +484,7 @@ bool CoreChecks::ValidateRenderPassLayoutAgainstFramebufferImageUsage(RenderPass
         const LogObjectList objlist(image, renderpass, framebuffer, image_view);
         skip |=
             LogError(objlist, "VUID-VkRenderPassBeginInfo-framebuffer-parameter",
-                     "%s: RenderPass %s uses %s where pAttachments[%" PRIu32 "] = %s, which refers to an invalid image",
+                     "%s: RenderPass %s uses %s where pAttachments[%" PRIu32 "] = %s, which refers to an invalid image.",
                      function_name, report_data->FormatHandle(renderpass).c_str(), report_data->FormatHandle(framebuffer).c_str(),
                      attachment_index, report_data->FormatHandle(image_view).c_str());
         return skip;
@@ -499,13 +498,14 @@ bool CoreChecks::ValidateRenderPassLayoutAgainstFramebufferImageUsage(RenderPass
 
     // Check for layouts that mismatch image usages in the framebuffer
     if (layout == VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL && !(image_usage & VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT)) {
-        vuid = use_rp2 ? "VUID-vkCmdBeginRenderPass2-initialLayout-03094" : "VUID-vkCmdBeginRenderPass-initialLayout-00895";
+        const char *vuid =
+            use_rp2 ? "VUID-vkCmdBeginRenderPass2-initialLayout-03094" : "VUID-vkCmdBeginRenderPass-initialLayout-00895";
         const LogObjectList objlist(image, renderpass, framebuffer, image_view);
         skip |= LogError(objlist, vuid,
                          "%s: Layout/usage mismatch for attachment %" PRIu32
                          " in %s"
                          " - the %s is %s but the image attached to %s via %s"
-                         " was not created with VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT",
+                         " was not created with VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT.",
                          function_name, attachment_index, report_data->FormatHandle(renderpass).c_str(), variable_name,
                          string_VkImageLayout(layout), report_data->FormatHandle(framebuffer).c_str(),
                          report_data->FormatHandle(image_view).c_str());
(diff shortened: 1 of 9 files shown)


tests/negative/subpass.cpp
juan-lunarg, line 1: Create `tests/positive/subpass.cpp` then add the corresponding positive tests.

tests/negative/renderpass.cpp
juan-lunarg, line 1: Add corresponding positive testing to `tests/positive/render_pass.cpp`

tests/negative/multiview.cpp
juan-lunarg, line 1: Create `tests/positive/multiview.cpp` then add the corresponding positive test(s).

tests/framework/layer_validation_tests.cpp
```diff
@@ -307,8 +303,9 @@ void TestRenderPassBegin(ErrorMonitor *error_monitor, const VkDevice device, con
         vk::ResetCommandBuffer(command_buffer, 0);
     }
     if (rp2Supported && rp2_vuid) {
-        PFN_vkCmdBeginRenderPass2KHR vkCmdBeginRenderPass2KHR =
-            (PFN_vkCmdBeginRenderPass2KHR)vk::GetDeviceProcAddr(device, "vkCmdBeginRenderPass2KHR");
+        auto vkCmdBeginRenderPass2KHR =
+            reinterpret_cast<PFN_vkCmdBeginRenderPass2KHR>(vk::GetDeviceProcAddr(device, "vkCmdBeginRenderPass2KHR"));
+        assert(vkCmdBeginRenderPass2KHR);
```
juan-lunarg, line 308: I don't think we need to do this anymore now that Artem's change is merged in?

juan-lunarg (changes requested):

tests/negative/subpass.cpp
spencer-lunarg: for a test like `RenderPassCreate2SubpassInvalidInputAttachmentParameters` would it not just be easier and more maintainable to add a 

```c++
// Valid
reference.aspectMask = VK_IMAGE_ASPECT_COLOR_BIT;
PositiveTestRenderPass2KHRCreate(*m_errorMonitor, *m_device, rpci2);
```

otherwise I feel we are writing an very basic positive test with little coverage

---

I agree having positive tests is good, but it is hard to determine what a good positive test would be because there infinite positive ways to write Vulkan and a set number of VUs how not to write it

The positive test that are the best are for the more complex VUs

layers/core_checks/image_layout_cc_validation.cpp
```diff
@@ -609,12 +615,66 @@ bool CoreChecks::ValidateRenderPassLayoutAgainstFramebufferImageUsage(RenderPass
...
                              " - the %s is %s but the image attached to %s via %s"
-                             " was not created with VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT",
+                             " was not created with VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT.",
                              function_name, attachment_index, report_data->FormatHandle(renderpass).c_str(), variable_name,
                              string_VkImageLayout(layout), report_data->FormatHandle(framebuffer).c_str(),
                              report_data->FormatHandle(image_view).c_str());
         }
     }
+
+    if (IsValueIn(layout, {VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_OPTIMAL,
```
spencer-lunarg, line 625: ## future cleanup PR idea after this

We have the `IsImageLayoutReadOnly`/`IsImageLayoutStencilReadOnly`/`IsImageLayoutDepthReadOnly` helper functions already

Here (and other places using depth and/or stencil layouts) could make use of a `IsSeperateDepthStencilLayout` - it would make this logic more easy to understand right away

layers/core_checks/render_pass_cc_validation.cpp
```diff
@@ -1759,6 +1764,37 @@ bool CoreChecks::ValidateRenderpassAttachmentUsage(RenderPassCreateVersion rp_ve
                             }
                         }
                     }
+
+                    if (IsExtEnabled(device_extensions.vk_khr_separate_depth_stencil_layouts)) {
```
spencer-lunarg, line 1768: this `if` is not needed, we will throw an error earlier if you use `VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_OPTIMAL` without the extension enabled

spencer-lunarg (approved):
Nits, otherwise happy with this

tests/negative/multiview.cpp
arno-lunarg: I am going to add a call to `PositiveTestRenderPass2KHRCreate` in this test

tests/negative/subpass.cpp
arno-lunarg: I share Spencer's view, here to me it makes more sense to add a `PositiveTestRenderPass2KHRCreate` call with valid parameters, and then invalidate them and do the negative testing.

tests/negative/renderpass.cpp
arno-lunarg: Given this PR's scope the only positive test I would consider is adding one where a `VkAttachmentDescriptionStencilLayoutKHR` is used, but there is already `VkPositiveLayerTest::BeginRenderPassDedicatedStencilLayout`

layers/core_checks/image_layout_cc_validation.cpp
arno-lunarg: I defined `IsImageLayoutStencilOnly` and `IsImageLayoutDepthOnly`, and replace the `IsValueIn` code by calls to those. I will do the cleanups in a following PR.

spencer-lunarg: awesome, it already looks much better to read with those functions in this PR

PR #5666, opened 2023-04-17
gpu: Use HOST_CACHED memory for descriptor set and bda state

jeremyg-lunarg:
The memory used for storing descriptor set state can be very large, (32 bits for every descriptor, some apps have 1m+ descriptors). Changing this memory from HOST_COHERENT to HOST_CACHED and manually flushing it saves about 1/3rd of the time spent updating this memory.

The buffer device address state could also grow quite large, so do the same for it.

### DIFF
--- a/layers/gpu_validation/gpu_validation.cpp
+++ b/layers/gpu_validation/gpu_validation.cpp
@@ -1299,6 +1299,10 @@ void GpuAssisted::UpdateInstrumentationBuffer(gpuav_state::CommandBuffer *cb_nod
                 for (const auto &update : buffer_info.update_at_submit) {
                     SetBindingState(data, update.first, update.second);
                 }
+                // Flush the descriptor state buffer before unmapping so that the new state is visible to the GPU
+                result = vmaFlushAllocation(vmaAllocator, buffer_info.allocation, 0, VK_WHOLE_SIZE);
+                // No good way to handle this error, we should still try to unmap.
+                assert(result == VK_SUCCESS);
                 vmaUnmapMemory(vmaAllocator, buffer_info.allocation);
             }
         }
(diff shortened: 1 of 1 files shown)


spencer-lunarg (approved):

layers/gpu_validation/gpu_validation.cpp
```diff
@@ -1299,6 +1299,7 @@ void GpuAssisted::UpdateInstrumentationBuffer(gpuav_state::CommandBuffer *cb_nod
                 for (const auto &update : buffer_info.update_at_submit) {
                     SetBindingState(data, update.first, update.second);
                 }
+                vmaFlushAllocation(vmaAllocator, buffer_info.allocation, 0, VK_WHOLE_SIZE);
```
juan-lunarg, line 1302: Return value isn't being used/checked.

Also comment add a comment explaining the flushing.

layers/gpu_validation/gpu_validation.cpp
```diff
@@ -1379,7 +1380,7 @@ void GpuAssisted::PostCallRecordCmdBindDescriptorSets(VkCommandBuffer commandBuf
             buffer_info.size = words_needed * 4;
             buffer_info.usage = VK_BUFFER_USAGE_STORAGE_BUFFER_BIT;
             VmaAllocationCreateInfo alloc_info = {};
-            alloc_info.requiredFlags = VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT;
+            alloc_info.requiredFlags = VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_CACHED_BIT;
```
juan-lunarg, line 1383: Comment

layers/gpu_validation/gpu_validation.cpp
```diff
@@ -1497,6 +1498,7 @@ void GpuAssisted::PostCallRecordCmdBindDescriptorSets(VkCommandBuffer commandBuf
                     }
                 }
             }
+            vmaFlushAllocation(vmaAllocator, di_input_block.allocation, 0, VK_WHOLE_SIZE);
```
juan-lunarg, line 1501: Return value isn't being used/checked.

Also comment add a comment explaining the flushing.

layers/gpu_validation/gpu_validation.cpp
jeremyg-lunarg: I'm not sure what we should do in case of an error with this call. We'd still want to try to unmap.

jeremyg-lunarg: should I just explicitly cast away the return value and comment?

juan-lunarg: I think asserting the return value is equal to VK_SUCCESS is good enough IMO

ncesario-lunarg (approved):

juan-lunarg (approved):

Tony-LunarG (approved):
LGTM

juan-lunarg:
How much does this improve performance btw? Like for doom eternal?

jeremyg-lunarg:
> How much does this improve performance btw? Like for doom eternal?

For that trace it drops the total CPU time from 70s to 64s. So a little bit less than 10%.

PR #5670, opened 2023-04-18
Label UNASSIGNED Atomic VUID

spencer-lunarg:
I while ago I got the WG to reword the 3 VUs

-  VUID-RuntimeSpirv-None-06337
-  VUID-RuntimeSpirv-None-06338
-  VUID-RuntimeSpirv-None-06339

so it would cover the case you were using `VK_EXT_shader_atomic_float2` to do `add`/`min`/`max` 

wrote a test to show how to hit this from having support for one with (16t-bit) but trying to use another (32-bit)


artem-lunarg (approved):
LGTM

PR #5674, opened 2023-04-18
layers: Clarify VUID error message

arno-lunarg:
Clarify that the "logical or" of the mentioned flags is not what is expected, rather it is a set of possible values

### DIFF
--- a/layers/core_checks/cc_render_pass.cpp
+++ b/layers/core_checks/cc_render_pass.cpp
@@ -1481,20 +1481,13 @@ bool CoreChecks::ValidateRenderpassAttachmentUsage(RenderPassCreateVersion rp_ve
                         if (!enabled_features.linear_color_attachment_features.linearColorAttachment) {
                             const char *vuid = use_rp2 ? "VUID-VkSubpassDescription2-pInputAttachments-02897"
                                                        : "VUID-VkSubpassDescription-pInputAttachments-02647";
-                            skip |= LogError(
-                                device, vuid,
-                                "%s: Input attachment %s format (%s) does not contain VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT "
-                                "| VK_FORMAT_FEATURE_DEPTH_STENCIL_ATTACHMENT_BIT.",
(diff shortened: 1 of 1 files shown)


layers/core_checks/render_pass_cc_validation.cpp
```diff
@@ -1478,20 +1478,20 @@ bool CoreChecks::ValidateRenderpassAttachmentUsage(RenderPassCreateVersion rp_ve
...
                                            : "VUID-VkSubpassDescription-pInputAttachments-02647";
-                            skip |= LogError(
-                                device, vuid,
-                                "%s: Input attachment %s format (%s) does not contain VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT "
-                                "| VK_FORMAT_FEATURE_DEPTH_STENCIL_ATTACHMENT_BIT.",
-                                function_name, error_type.c_str(), string_VkFormat(attachment_format));
+                            skip |= LogError(device, vuid,
+                                             "%s: Input attachment %s format (%s) does not contain any of "
+                                             "VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT "
+                                             "or VK_FORMAT_FEATURE_DEPTH_STENCIL_ATTACHMENT_BIT.",
```
juan-lunarg, line 1484: Just my 2 cents. I think this would be easier to read. Since the VUID description gets amended to the error message.

```suggestion
                                             "%s: Input attachment %s format (%s) lacks sufficient format features!",
```

> %s: Input attachment %s format (%s) lacks sufficient format features! All attachments in pInputAttachments that are not VK_ATTACHMENT_UNUSED must have image formats whose [potential format features](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#potential-format-features) contain at least VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT or VK_FORMAT_FEATURE_DEPTH_STENCIL_ATTACHMENT_BIT

arno-lunarg: Yes this is better. Not sure about the `!` though?

juan-lunarg (approved):

PR #5680, opened 2023-04-18
Sub state cts

ncesario-lunarg:


spencer-lunarg (approved):

layers/state_tracker/pipeline_state.h
```diff
@@ -463,7 +463,28 @@ class PIPELINE_STATE : public BASE_NODE {
                                                                           const safe_VkGraphicsPipelineCreateInfo &safe_create_info,
                                                                           const std::shared_ptr<const RENDER_PASS_STATE> &rp);
 
-    static bool EnablesRasterizationStates(const safe_VkGraphicsPipelineCreateInfo &create_info);
+    template <typename PRState>
+    static bool EnablesRasterizationStates(const PRState pr_state) {
```
artem-lunarg, line 467: nit: pr_state -> pre_raster_state is easier to understand personally for me, but if 'pr' is widely used then it's fine.

artem-lunarg, line 467: Is the expectation this function will be used outside pipeline_state.cpp ?

ncesario-lunarg: :+1: I agree

ncesario-lunarg: Yeah, I have an upcoming change. Sorry, I should have mentioned that explicitly :grimacing:

Issue #5604, opened 2023-04-09
Unexpected validation message with vkGetFenceStatus()


Cyphall:
**Environment:**
 - OS: Windows 11
 - GPU: RTX 3070
 - SDK version: 1.3.243.0
 - Options enabled: eBestPractices, eSynchronizationValidation

**Describe the Issue**

When calling `vkGetFenceStatus()`, if the return value is `VK_NOT_READY`, the validation will emit the following message (with a severity of "INFO"):
```
Validation Information: [ UNASSIGNED-BestPractices-NonSuccess-Result ] Object 0: handle = 0x21e67a84f80, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0x8928392f | vkGetFenceStatus(): Returned non-success return code VK_NOT_READY.
```

**Expected behavior**

As per the [Vulkan specs](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetFenceStatus.html), `VK_NOT_READY` is considered a "success" return value for `vkGetFenceStatus()`.
However, the validation layers tells me `Returned non-success return code VK_NOT_READY`, which is incorrect.

Also, I would expect no message at all since there is no error value being returned there.

**Valid Usage ID**
`UNASSIGNED-BestPractices-NonSuccess-Result`

mark-lunarg:
The default configuration of the validation layers is responsible for enforcing the Vulkan specification and it does treat `VK_NOT_READY` as a success code -- no error is emitted.  However, the intent of enabling the optional Best Practices layer ([see the description here](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/docs/best_practices.md))  is to point out API usage which is may be technically correct but might lead to bad assumptions or programming practices. As you noted this layer will output 'info' messages for any return code that is not `VK_SUCCESS`.  This is because it is a best practice to check for non-failure return codes so that a well-written application can handle those cases correctly.

Cyphall:
That makes sense.

However, I still wonder if this message should be sent for `vkGetFenceStatus()` specificly, because this function is basically a "const" function (in C++ terms), with officially no other effect than returning a value, which means that calling this function and discarding the result can already be considered as a bug in itself.
I am not sure if the scope of the Best Practices layer corresponds to that.

spencer-lunarg:
>  it does treat VK_NOT_READY as a success code

[...]

Issue #4654, opened 2022-10-13
UNASSIGNED-BestPractices-SemaphoreCount and similar are spammy and not useful


ShabbyX:
**TL;DR please remove these messages**

A class of best practices messages like `UNASSIGNED-BestPractices-SemaphoreCount` say something like "count is 0, but pointer is not NULL". This message doesn't seem useful; it's pretty common to have code like this, and it's perfectly legitimate code:

```
data.pData = dataStorage;
data.dataCount = count;
```

To silence this message, the app would have to do this:

```
data.pData = count == 0 ? NULL : dataStorage;
data.dataCount = count;
```

which is arguably not an improvement. However, this by itself is not a big deal, just a minor inconvenience.

The reason I brought this up is this however: https://github.com/baldurk/renderdoc/pull/2739

When running an app that enables best practices on RenderDoc, due to RenderDoc's capture and reconstruction of the apps structs, this message spams the error output in RenderDoc even if the app doesn't originally trigger it. While this could be worked around in RenderDoc, it may be better to just stop spamming apps with this message.

PR #5682, opened 2023-04-18
layers: Add LogDebug

juan-lunarg:
The intent is to address concerns of #4654 and #5604

VK_DEBUG_UTILS_MESSAGE_SEVERITY_VERBOSE_BIT_EXT (LogDebug):
`Diagnostic message`

VK_DEBUG_UTILS_MESSAGE_SEVERITY_INFO_BIT_EXT (LogInfo):
`Informational message like the creation of a resource`

### DIFF
--- a/layers/best_practices/best_practices_utils.cpp
+++ b/layers/best_practices/best_practices_utils.cpp
@@ -820,8 +820,7 @@ void BestPractices::ValidateReturnCodes(const char* api_name, VkResult result) c
 
     // While the result is successful, users may still want this information. Since it may be unexpected.
     if (result > VK_SUCCESS) {
-        LogInfo(instance, kVUID_BestPractices_NonSuccess_Result, "%s(): Returned non-success return code %s.", api_name,
-                result_string);
+        LogDebug(instance, kVUID_BestPractices_NonSuccess_Result, "%s(): Returned successful value %s.", api_name, result_string);
         return;
     }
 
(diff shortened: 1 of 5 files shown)


juan-lunarg:
> The intent is to address concerns of https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4654 and https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5604

Basically the idea is to use `LogDebug` for cases where we are being extra verbose about vulkan behavior. In both instances we are reporting on valid behavior. And in both cases we have users complaining about not wanting the message.

As far as I can tell we don't use `kDebugBit` and I think this would be a good case for it. Also IMO it should be the `kVerboseBit` since it would line up nicely with `VK_DEBUG_UTILS_MESSAGE_SEVERITY_VERBOSE_BIT_EXT`

jeremyg-lunarg:
> 


> > The intent is to address concerns of #4654 and #5604
> 
> Basically the idea is to use `LogDebug` for cases where we are being extra verbose about vulkan behavior. In both instances we are reporting on valid behavior. And in both cases we have users complaining about not wanting the message.
> 
> As far as I can tell we don't use `kDebugBit` and I think this would be a good case for it. Also IMO it should be the `kVerboseBit` since it would line up nicely with `VK_DEBUG_UTILS_MESSAGE_SEVERITY_VERBOSE_BIT_EXT`

I really like this idea. However, I disagree with the messages you moved to LogDebug(). I think they should be LogInfo(). 

I think we should conform to recommendations of [syslog](https://en.wikipedia.org/wiki/Syslog#Severity_level) as far is reasonable. That would imply that Info is for things that a user might want to see. And Debug/Verbose would be for things that a VVL developer might want to see while investigating a bug.

jeremyg-lunarg:
> > 
> 
> > > The intent is to address concerns of #4654 and #5604
> > 
> > 
> > Basically the idea is to use `LogDebug` for cases where we are being extra verbose about vulkan behavior. In both instances we are reporting on valid behavior. And in both cases we have users complaining about not wanting the message.
> > As far as I can tell we don't use `kDebugBit` and I think this would be a good case for it. Also IMO it should be the `kVerboseBit` since it would line up nicely with `VK_DEBUG_UTILS_MESSAGE_SEVERITY_VERBOSE_BIT_EXT`
> 
> I really like this idea. However, I disagree with the messages you moved to LogDebug(). I think they should be LogInfo().
> 
> I think we should conform to recommendations of [syslog](https://en.wikipedia.org/wiki/Syslog#Severity_level) as far is reasonable. That would imply that Info is for things that a user might want to see. And Debug/Verbose would be for things that a VVL developer might want to see while investigating a bug.

Also you should update one of the documentation files with our recommendations, no matter which way we go on Debug vs Info

juan-lunarg:
> I really like this idea. However, I disagree with the messages you moved to LogDebug(). I think they should be LogInfo().

I think `kVUID_BestPractices_NonSuccess_Result` should be `LogDebug`. Since the function returned a successful value. And would only be useful in niche debugging scenarios.

However, I do agree that `kVUID_BestPractices_SemaphoreCount` should be `LogInfo` instead.

juan-lunarg:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5699

https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5700

PR #5685, opened 2023-04-19
Multiplane ImageView aspect mask check

spencer-lunarg:
Adds `VUID-VkImageViewCreateInfo-subresourceRange-07818` using new `IsOnlyOneValidPlaneAspect` helper

### DIFF
--- a/layers/core_checks/cc_image.cpp
+++ b/layers/core_checks/cc_image.cpp
@@ -1710,8 +1710,8 @@ bool CoreChecks::PreCallValidateCreateImageView(VkDevice device, const VkImageVi
 
             if (!IsOnlyOneValidPlaneAspect(image_format, aspect_mask)) {
                 skip |= LogError(pCreateInfo->image, "VUID-VkImageViewCreateInfo-subresourceRange-07818",
-                                 "vkCreateImageView(): subresourceRange.aspectMask (0x%x) is invalid for %s.", aspect_mask,
-                                 string_VkFormat(image_format));
+                                 "vkCreateImageView(): subresourceRange.aspectMask (%s) is invalid for %s.",
+                                 string_VkImageAspectFlags(aspect_mask).c_str(), string_VkFormat(image_format));
             } else if ((image_class != view_class) || (image_class == FORMAT_COMPATIBILITY_CLASS::NONE)) {
                 // Need to only check if one is NONE to handle edge case both are NONE
                 // View format must match the multiplane compatible format
--- a/layers/core_checks/cc_image.cpp
+++ b/layers/core_checks/cc_image.cpp
@@ -1701,18 +1701,22 @@ bool CoreChecks::PreCallValidateCreateImageView(VkDevice device, const VkImageVi
     }
 
     const bool multiplane_image = FormatIsMultiplane(image_format);
+    if (multiplane_image && IsMultiplePlaneAspect(aspect_mask)) {
+        skip |= LogError(pCreateInfo->image, "VUID-VkImageViewCreateInfo-subresourceRange-07818",
+                         "vkCreateImageView(): subresourceRange.aspectMask (%s) is invalid for %s.",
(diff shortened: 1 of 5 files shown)


juan-lunarg:
Fails on MacOS
```
[  FAILED  ] VkLayerTest.MultiplaneImageViewAspectMasks
```

tests/negative/ycbcr.cpp
```diff
@@ -1456,6 +1455,67 @@ TEST_F(VkLayerTest, MultiplaneIncompatibleViewFormat) {
...
+    ycbcr_create_info.components = {VK_COMPONENT_SWIZZLE_IDENTITY, VK_COMPONENT_SWIZZLE_IDENTITY, VK_COMPONENT_SWIZZLE_IDENTITY,
+                                    VK_COMPONENT_SWIZZLE_IDENTITY};
+    ycbcr_create_info.xChromaOffset = VK_CHROMA_LOCATION_MIDPOINT;
+    ycbcr_create_info.yChromaOffset = VK_CHROMA_LOCATION_MIDPOINT;
+    ycbcr_create_info.chromaFilter = VK_FILTER_NEAREST;
+    ycbcr_create_info.forceExplicitReconstruction = false;
+
+    vk_testing::SamplerYcbcrConversion conversion(*m_device, ycbcr_create_info);
+
+    VkSamplerYcbcrConversionInfo ycbcr_info = LvlInitStruct<VkSamplerYcbcrConversionInfo>();
```
artem-lunarg: Can `ycbcr_info` and `conversion` be removed?

layers/core_checks/cc_image.cpp
```diff
@@ -1701,18 +1701,22 @@ bool CoreChecks::PreCallValidateCreateImageView(VkDevice device, const VkImageVi
...
         if (multiplane_image) {
             const VkFormat compat_format = FindMultiplaneCompatibleFormat(image_format, aspect_mask);
             const auto image_class = FormatCompatibilityClass(compat_format);
-
-            if (!IsOnlyOneValidPlaneAspect(image_format, aspect_mask)) {
-                skip |= LogError(pCreateInfo->image, "VUID-VkImageViewCreateInfo-subresourceRange-07818",
-                                 "vkCreateImageView(): subresourceRange.aspectMask (%s) is invalid for %s.",
-                                 string_VkImageAspectFlags(aspect_mask).c_str(), string_VkFormat(image_format));
-            } else if ((image_class != view_class) || (image_class == FORMAT_COMPATIBILITY_CLASS::NONE)) {
+            // Need valid aspect mask otherwise will throw extra error when getting compatible format
```
artem-lunarg, line 1716: Just trying to understand, where the extra error can happen? Unexpected return value in  `GetPlaneIndex`?

spencer-lunarg: So there are some cases where you **can** use `ASPECT_COLOR`

> VUID-VkImageViewCreateInfo-image-01762
If image was not created with the VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT flag, or if the format of the image is a **multi-planar format** and if subresourceRange.aspectMask is **VK_IMAGE_ASPECT_COLOR_BIT**, format must be identical to the format used to create image

So if that is the case, `FindMultiplaneCompatibleFormat` returns `FORMAT_COMPATIBILITY_CLASS::NONE`, which is ok as the VU says

> VUID-VkImageViewCreateInfo-subresourceRange-07818
subresourceRange.aspectMask must only have **at most 1** valid [multi-planar aspect mask](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#formats-planes-image-aspect)

We have tests that were hitting this, thats how I found out

artem-lunarg: Thanks, this makes it clear.

artem-lunarg (approved):
LGTM!

PR #5686, opened 2023-04-19
VkDeviceImageMemoryRequirementsKHR plane aspect

spencer-lunarg:
Adds VUID-VkDeviceImageMemoryRequirementsKHR-pCreateInfo-06419 using new `IsOnlyOneValidPlaneAspect` helper

### DIFF
--- a/layers/stateless/sl_device_memory.cpp
+++ b/layers/stateless/sl_device_memory.cpp
@@ -180,24 +180,26 @@ bool StatelessValidation::ValidateDeviceImageMemoryRequirements(VkDevice device,
     bool skip = false;
 
     if (pInfo && pInfo->pCreateInfo) {
-        const auto *image_swapchain_create_info = LvlFindInChain<VkImageSwapchainCreateInfoKHR>(pInfo->pCreateInfo);
-        if (image_swapchain_create_info) {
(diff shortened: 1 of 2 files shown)


layers/stateless/device_memory_sl_validation.cpp
```diff
@@ -180,24 +180,26 @@ bool StatelessValidation::ValidateDeviceImageMemoryRequirements(VkDevice device,
     bool skip = false;
 
     if (pInfo && pInfo->pCreateInfo) {
-        const auto *image_swapchain_create_info = LvlFindInChain<VkImageSwapchainCreateInfoKHR>(pInfo->pCreateInfo);
-        if (image_swapchain_create_info) {
+        const auto create_info = *(pInfo->pCreateInfo);
```
artem-lunarg, line 183: nit: reference is not deduced by `auto` in case we need it here

artem-lunarg (approved):
LGTM!

layers/stateless/device_memory_sl_validation.cpp
spencer-lunarg: good call, let me add that here quick

arno-lunarg (approved):

Issue #5687, opened 2023-04-19
Spec violation when validating DMA-BUF import in vkCreateImage


emersion:
**Environment:**
 - OS: Linux
 - GPU: Intel
 - SDK or header version if building from repo: latest commit

**Describe the Issue**

Call `vkCreateImage` with an `VkExternalMemoryImageCreateInfo` chained with the `VK_EXTERNAL_MEMORY_HANDLE_TYPE_DMA_BUF_BIT_EXT` handle type.

The validation layers will perform a `GetPhysicalDeviceImageFormatProperties2` under the hood with a `VK_IMAGE_TILING_DRM_FORMAT_MODIFIER_EXT` tiling, but will not chain any `VkPhysicalDeviceImageDrmFormatModifierInfoEXT`.

This is a violation of the spec: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceImageFormatInfo2.html#VUID-VkPhysicalDeviceImageFormatInfo2-tiling-02249

The ANV driver will crash with this stack trace:

```
    #0 0x7f8a11d8c2df in anv_get_image_format_properties ../src/intel/vulkan/anv_formats.c:1017
    #1 0x7f8a11d9098d in anv_GetPhysicalDeviceImageFormatProperties2 ../src/intel/vulkan/anv_formats.c:1431
    #2 0x7f8a0acdc41b in DispatchGetPhysicalDeviceImageFormatProperties2(VkPhysicalDevice_T*, VkPhysicalDeviceImageFormatInfo2 const*, VkImageFormatProperties2*) /home/simon/src/Vulkan-ValidationLayers/layers/generated/layer_chassis_dispatch.cpp:4170
    #3 0x7f8a0a510e66 in CoreChecks::PreCallValidateCreateImage(VkDevice_T*, VkImageCreateInfo const*, VkAllocationCallbacks const*, VkImage_T**) const /home/simon/src/Vulkan-ValidationLayers/layers/core_checks/image_cc_validation.cpp:471
    #4 0x7f8a0a9f4b68 in vulkan_layer_chassis::CreateImage(VkDevice_T*, VkImageCreateInfo const*, VkAllocationCallbacks const*, VkImage_T**) /home/simon/src/Vulkan-ValidationLayers/layers/generated/chassis.cpp:1972
    #5 0x7f8a1e14bb3c in vulkan_import_dmabuf ../render/vulkan/texture.c:531
[...]

PR #5688, opened 2023-04-19
Chain VkPhysicalDeviceImageDrmFormatModifierInfoEXT in PreCallValidateCreateImage

emersion:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5687

### DIFF
--- a/layers/core_checks/image_cc_validation.cpp
+++ b/layers/core_checks/image_cc_validation.cpp
@@ -466,6 +466,14 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
         image_info.usage = pCreateInfo->usage;
         image_info.flags = pCreateInfo->flags;
 
+        const auto image_drm_format_modifier_explicit_create_info_ext = LvlFindInChain<VkImageDrmFormatModifierExplicitCreateInfoEXT>(pCreateInfo->pNext);
+        auto drm_format_modifier_info_ext = LvlInitStruct<VkPhysicalDeviceImageDrmFormatModifierInfoEXT>();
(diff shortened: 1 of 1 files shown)


layers/core_checks/image_cc_validation.cpp
```diff
@@ -466,6 +466,14 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
         image_info.usage = pCreateInfo->usage;
         image_info.flags = pCreateInfo->flags;
 
+        const auto image_drm_format_modifier_explicit_create_info_ext = LvlFindInChain<VkImageDrmFormatModifierExplicitCreateInfoEXT>(pCreateInfo->pNext);
+        auto drm_format_modifier_info_ext = LvlInitStruct<VkPhysicalDeviceImageDrmFormatModifierInfoEXT>();
+        if (pCreateInfo->tiling == VK_IMAGE_TILING_DRM_FORMAT_MODIFIER_EXT && image_drm_format_modifier_explicit_create_info_ext != nullptr) {
```
emersion, line 471: Maybe we should just assert that the `VkImageDrmFormatModifierExplicitCreateInfoEXT` is non-null here, because the parameter validation would catch it.

spencer-lunarg (approved):
This looks good and assume this fixes your problem

We will want a positive test to prevent regression, if you want to add that, otherwise I can do it 

It would be in `tests/positive/external_memory_sync.cpp`

emersion:
> We will want a positive test to prevent regression, if you want to add that, otherwise I can do it

Yeah, sorry, I have no clue how to do this, I'll leave it up to you.

emersion:
Closing in favor of #5706

PR #5690, opened 2023-04-19
layers: Use constexpr for some vk_format_helpers

gnl21:
These can be used as constexpr and are called from other constexpr functions, so not using the qualifier is needlessly restrictive. Some older compilers also complain about calling non-constexpr functions from within constexpr functions.


spencer-lunarg (approved):

spencer-lunarg:
> Are we OK to move these functions into the header file so that they can be used as constexpr?

I am personally ok with it, I know how much faster `constexpr` can turns things into

layers/generated/vk_format_utils.h
```diff
@@ -181,9 +181,63 @@ bool FormatIsXChromaSubsampled(VkFormat format);
...
+        case VK_FORMAT_G8_B8_R8_3PLANE_420_UNORM:
+        case VK_FORMAT_G8_B8_R8_3PLANE_422_UNORM:
+        case VK_FORMAT_G8_B8_R8_3PLANE_444_UNORM:
+            return 3;
+        default:
+            return 1;
+     }
+}
+
+static inline constexpr bool FormatIsMultiplane(VkFormat format) { return ((FormatPlaneCount(format)) > 1u); }
```
juan-lunarg, line 240:
```suggestion
constexpr bool FormatIsMultiplane(VkFormat format) { return ((FormatPlaneCount(format)) > 1u); }
```

layers/generated/vk_format_utils.h
```diff
@@ -181,9 +181,63 @@ bool FormatIsXChromaSubsampled(VkFormat format);
...
+        case VK_FORMAT_B10X6G10X6R10X6G10X6_422_UNORM_4PACK16:
+        case VK_FORMAT_B12X4G12X4R12X4G12X4_422_UNORM_4PACK16:
+        case VK_FORMAT_B16G16R16G16_422_UNORM:
+        case VK_FORMAT_B8G8R8G8_422_UNORM:
+        case VK_FORMAT_G10X6B10X6G10X6R10X6_422_UNORM_4PACK16:
+        case VK_FORMAT_G12X4B12X4G12X4R12X4_422_UNORM_4PACK16:
+        case VK_FORMAT_G16B16G16R16_422_UNORM:
+        case VK_FORMAT_G8B8G8R8_422_UNORM:
+        found = true;
+            break;
```
juan-lunarg, line 199:
```suggestion
            found = true;
            break;
```

layers/generated/vk_format_utils.h
```diff
@@ -181,9 +181,63 @@ bool FormatIsXChromaSubsampled(VkFormat format);
 bool FormatIsYChromaSubsampled(VkFormat format);
 
 // Multiplane
-bool FormatIsSinglePlane_422(VkFormat format);
-uint32_t FormatPlaneCount(VkFormat format);
-static inline bool FormatIsMultiplane(VkFormat format) { return ((FormatPlaneCount(format)) > 1u); }
+
+// Single-plane "_422" formats are treated as 2x1 compressed (for copies)
+
```
juan-lunarg, line 186:
```suggestion
// Single-plane "_422" formats are treated as 2x1 compressed (for copies)
```

juan-lunarg (changes requested):
Minor nitpicks. LGTM overall though.

layers/generated/vk_format_utils.h
gnl21: This is generated from `commonBoolSwitch`, which bakes in an indentation. It was like this in the original, so I'd prefer to leave it alone, but if you'd like me to change the generating script then let me know what I should do.

layers/generated/vk_format_utils.h
gnl21: Done

layers/generated/vk_format_utils.h
gnl21: Done

juan-lunarg (approved):

Issue #5694, opened 2023-04-19
False positive VUID-vkResetCommandBuffer-commandBuffer-00045 error after exporting submitted fence handle


i-nazarov-ukraine:
**Environment:**
 - OS: Android
 - GPU: Any
 - SDK or header version if building from repo: 1.3.247
 - Options enabled: Default

**Describe the Issue**

- Create `VkFence` with `VkExportFenceCreateInfoKHR::handleTypes = VK_EXTERNAL_FENCE_HANDLE_TYPE_SYNC_FD_BIT_KHR`;
- Submit empty `VkCommandBuffer` with created Fence.
- Call `vkGetFenceFdKHR()` on the Fence.
- Wait on the exported FD using `poll()` function.
- Call `vkResetCommandBuffer()` on the Command Buffer.
- The "VUID-vkResetCommandBuffer-commandBuffer-00045" error is generated.

**Expected behavior**

VVL should not generate the error, because it has no means to verify how exported FD was used by the application.
It should behave similarly to `vkQueueSubmit()` when submitting `VkFence` with scope other than `kSyncScopeInternal`.
In other words, submission should be retired early after `vkGetFenceFdKHR()` is called.

**Additional context**

<details>
<summary>terminal output</summary>

```sh
Validation Error: [ VUID-vkResetCommandBuffer-commandBuffer-00045 ] Object 0: handle = 0x6e09023af0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x1e7883ea | Attempt to reset VkCommandBuffer 0x6e09023af0[] which is in use. The Vulkan spec states: commandBuffer must not be in the pending state (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkResetCommandBuffer-commandBuffer-00045)
```

</details>

artem-lunarg:
Fixed by https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5695

PR #5695, opened 2023-04-19
Fix false positive VUID-vkResetCommandBuffer-commandBuffer-00045 error after exporting submitted fence handle

i-nazarov-samsung:
This request fixes the issue: https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5694

### DIFF
--- a/layers/state_tracker/queue_state.cpp
+++ b/layers/state_tracker/queue_state.cpp
@@ -285,9 +285,7 @@ void FENCE_STATE::Reset() {
     if (scope_ == kSyncScopeExternalTemporary) {
         scope_ = kSyncScopeInternal;
     }
-    if (scope_ == kSyncScopeInternal) {
-        state_ = FENCE_UNSIGNALED;
-    }
+    state_ = FENCE_UNSIGNALED;
     completed_ = std::promise<void>();
     waiter_ = std::shared_future<void>(completed_.get_future());
 }
@@ -295,11 +293,10 @@ void FENCE_STATE::Reset() {
 void FENCE_STATE::Import(VkExternalFenceHandleTypeFlagBits handle_type, VkFenceImportFlags flags) {
     auto guard = WriteLock();
     if (scope_ != kSyncScopeExternalPermanent) {
-        if ((handle_type == VK_EXTERNAL_FENCE_HANDLE_TYPE_SYNC_FD_BIT || flags & VK_FENCE_IMPORT_TEMPORARY_BIT) &&
-            scope_ == kSyncScopeInternal) {
-            scope_ = kSyncScopeExternalTemporary;
-        } else {
+        if (handle_type != VK_EXTERNAL_FENCE_HANDLE_TYPE_SYNC_FD_BIT && (flags & VK_FENCE_IMPORT_TEMPORARY_BIT) == 0) {
             scope_ = kSyncScopeExternalPermanent;
+        } else if (scope_ == kSyncScopeInternal) {
+            scope_ = kSyncScopeExternalTemporary;
         }
     }
 }
@@ -309,9 +306,14 @@ void FENCE_STATE::Export(VkExternalFenceHandleTypeFlagBits handle_type) {
     if (handle_type != VK_EXTERNAL_FENCE_HANDLE_TYPE_SYNC_FD_BIT) {
         // Export with reference transference becomes external
         scope_ = kSyncScopeExternalPermanent;
-    } else if (scope_ == kSyncScopeInternal) {
+    } else {
         // Export with copy transference has a side effect of resetting the fence
+        if (scope_ == kSyncScopeExternalTemporary) {
+            scope_ = kSyncScopeInternal;
+        }
         state_ = FENCE_UNSIGNALED;
+        completed_ = std::promise<void>();
+        waiter_ = std::shared_future<void>(completed_.get_future());
     }
 }
 
--- a/tests/positive/sync.cpp
+++ b/tests/positive/sync.cpp
@@ -19,6 +19,10 @@
 #include <deque>
 #include <thread>
 
+#ifndef VK_USE_PLATFORM_WIN32_KHR
+#include <poll.h>
+#endif
+
 //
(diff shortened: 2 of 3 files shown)


juan-lunarg (changes requested):
Currently CI is failing on `VkPositiveLayerTest.ExternalFenceSubmitCmdBuffer`

```
[  FAILED  ] VkPositiveLayerTest.ExternalFenceSubmitCmdBuffer

 1 FAILED TEST
  YOU HAVE 10 DISABLED TESTS
```

jeremyg-lunarg:
> Currently CI is failing on `VkPositiveLayerTest.ExternalFenceSubmitCmdBuffer`

Could this be related to the use of MockICD in our github actions for testing on Linux?  I doubt it exports a valid fence file descriptor.

The code changes seem reasonable to me.

juan-lunarg (changes requested):
Test runs fine on my AMD APU. Doesn't work on our Mock driver. Looking at the other tests this is a known limitation.

tests/positive/sync.cpp
```diff
@@ -1622,6 +1626,84 @@ TEST_F(VkPositiveLayerTest, ExternalFenceSyncFdLoop) {
...
+    const auto handle_type = VK_EXTERNAL_FENCE_HANDLE_TYPE_SYNC_FD_BIT_KHR;
+    AddRequiredExtensions(VK_KHR_EXTERNAL_FENCE_CAPABILITIES_EXTENSION_NAME);
+    AddRequiredExtensions(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
+    AddRequiredExtensions(VK_KHR_EXTERNAL_FENCE_EXTENSION_NAME);
+    AddRequiredExtensions(extension_name);
+    ASSERT_NO_FATAL_FAILURE(InitFramework(m_errorMonitor));
+    ASSERT_NO_FATAL_FAILURE(InitState(nullptr, nullptr, VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT));
+    if (!AreRequiredExtensionsEnabled()) {
+        GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported";
+    }
```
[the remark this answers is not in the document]
i-nazarov-samsung: Thanks!

On my end test runs OK on Android: S906B (Samsung S22 AMD), G996B (S21+ Mali), G991U (S21 Adreno), and Pixel 4a.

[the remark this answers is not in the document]
i-nazarov-samsung: Done.

tests/positive/sync.cpp
```diff
@@ -1622,6 +1626,84 @@ TEST_F(VkPositiveLayerTest, ExternalFenceSyncFdLoop) {
...
+
+    // Check for external fence export capability
+    auto efi = LvlInitStruct<VkPhysicalDeviceExternalFenceInfoKHR>();
+    efi.handleType = handle_type;
+    auto efp = LvlInitStruct<VkExternalFencePropertiesKHR>();
+    auto vkGetPhysicalDeviceExternalFencePropertiesKHR = (PFN_vkGetPhysicalDeviceExternalFencePropertiesKHR)vk::GetInstanceProcAddr(
+        instance(), "vkGetPhysicalDeviceExternalFencePropertiesKHR");
+    vkGetPhysicalDeviceExternalFencePropertiesKHR(gpu(), &efi, &efp);
```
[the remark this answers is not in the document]
i-nazarov-samsung: Done.

juan-lunarg (approved):

ShabbyX:
I can confirm the test runs fine on Pixel 6

spencer-lunarg (approved):

layers/state_tracker/queue_state.cpp
```diff
@@ -309,9 +308,16 @@ void FENCE_STATE::Export(VkExternalFenceHandleTypeFlagBits handle_type) {
...
         // Export with reference transference becomes external
         scope_ = kSyncScopeExternalPermanent;
-    } else if (scope_ == kSyncScopeInternal) {
+    } else {
         // Export with copy transference has a side effect of resetting the fence
-        state_ = FENCE_UNSIGNALED;
+        if (scope_ == kSyncScopeExternalTemporary) {
+            scope_ = kSyncScopeInternal;
+        }
+        if (scope_ == kSyncScopeInternal) {
```
artem-lunarg, line 316: Do we need this condition?
According to spec:
> Additionally, exporting a fence payload to a handle with copy transference has the same side effects on the source fence’s payload as executing a fence reset operation.

i-nazarov-samsung: > According to spec it sounds like the reset happens unconditionally:

I guess the question was for the `if (scope_ == kSyncScopeInternal)` condition and unconditional `state_ = FENCE_UNSIGNALED` assignment. I'm not 100% sure either. I took logic from `FENCE_STATE::Reset()`.

I do not see strong reasons why we need this condition in both methods. I guess the original intent was that if scope is not `kSyncScopeInternal`, other parts of the code ignore what is in the `state_`, therefore - no need to change it.

However, `completed_` is still reset regardless of `scope_ `. IMO, it is better to also reset `state_` unconditionally for consistency.

i-nazarov-samsung: Removed `if (scope_ == kSyncScopeInternal)` from both methods.

artem-lunarg: Thanks for the explanation. Yeah, if no good reason then a check before assignment is not an (micro)optimization either, condition if not predicted is much heavier. But IMO the main reason that it can cost developer time in an attempt to understand the whys. If it breaks something then we will revert this part back with a good comment.

artem-lunarg (approved):

artem-lunarg (approved):
LGTM, still need to wait for CI to confirm that suggestion to reset unconditionally does not cause issues.

juan-lunarg:
Thanks for the fix/test @i-nazarov-samsung :+1:

PR #5699, opened 2023-04-19
bp: Log BestPractices-SemaphoreCount as informational

### LINKED ISSUE #4654: UNASSIGNED-BestPractices-SemaphoreCount and similar are spammy and not useful

juan-lunarg:
As highlighted in #4654 it isn't ideal to log this as a warning.

Downgrading it to VK_DEBUG_UTILS_MESSAGE_SEVERITY_INFO_BIT_EXT seems like a good compromise since the logging was requested by an engineer working on Vulkan CTS originally.

closes #4654

### DIFF
--- a/layers/best_practices/bp_instance_device.cpp
+++ b/layers/best_practices/bp_instance_device.cpp
@@ -381,15 +381,14 @@ bool BestPractices::PreCallValidateQueueSubmit(VkQueue queue, uint32_t submitCou
             skip |= CheckPipelineStageFlags("vkQueueSubmit", pSubmits[submit].pWaitDstStageMask[semaphore]);
         }
         if (pSubmits[submit].signalSemaphoreCount == 0 && pSubmits[submit].pSignalSemaphores != nullptr) {
-            skip |=
-                LogWarning(device, kVUID_BestPractices_SemaphoreCount,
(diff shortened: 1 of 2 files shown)


juan-lunarg:
@ShabbyX would this approach work for you?

ShabbyX:
> @ShabbyX would this approach work for you?

We have disabled best practices entirely (it never produced anything useful for ANGLE), so I'm afraid I no longer have any stakes in this.

spencer-lunarg (approved):

arno-lunarg (approved):

PR #5700, opened 2023-04-19
bp: Log successful return values with LogVerbose

### LINKED ISSUE #5604: Unexpected validation message with vkGetFenceStatus()

juan-lunarg:
Log successful return values with LogVerbose instead of LogInfo, but only if the vulkan call can return several different success values. EX: vkGetFenceStatus

closes #5604


spencer-lunarg (approved):

tests/negative/best_practices.cpp
```diff
@@ -86,8 +86,9 @@ TEST_F(VkBestPracticesLayerTest, ValidateReturnCodes) {
     format_count -= 1;
     formats.resize(format_count);
 
-    m_errorMonitor->SetDesiredFailureMsg(kInformationBit, "UNASSIGNED-BestPractices-NonSuccess-Result");
+    m_errorMonitor->SetDesiredFailureMsg(kVerboseBit, "UNASSIGNED-BestPractices-Verbose-Success-Logging");
     result = vk::GetPhysicalDeviceSurfaceFormatsKHR(gpu(), m_surface, &format_count, formats.data());
+    assert(result > VK_SUCCESS);
```
arno-lunarg, line 91:
```suggestion
    ASSERT_TRUE(result > VK_SUCCESS);
```
(it it makes sense?)

layers/best_practices/best_practices_validation.h
```diff
@@ -697,7 +697,8 @@ class BestPractices : public ValidationStateTracker {
     bool PreCallValidateCmdClearAttachments(VkCommandBuffer commandBuffer, uint32_t attachmentCount,
                                             const VkClearAttachment* pAttachments, uint32_t rectCount,
                                             const VkClearRect* pRects) const override;
-    void ValidateReturnCodes(const char* api_name, VkResult result) const;
+    void LogSuccessCodes(const char* api_name, VkResult result) const;
```
arno-lunarg, line 700: I do not get the plural for these functions? From my understanding, only one result is logged?

arno-lunarg (approved):
LGTM! 2 questions
This is a welcome feature I think!

layers/best_practices/best_practices_validation.h
juan-lunarg: I just carried over the plural from `ValidateReturnCodes` you are right it doesn't make much sense.

tests/negative/best_practices.cpp
juan-lunarg: It makes sense.

layers/best_practices/best_practices_utils.cpp
```diff
@@ -101,25 +101,21 @@ const char* BestPractices::VendorSpecificTag(BPVendorFlags vendors) const {
     return res->second.c_str();
 }
 
-void BestPractices::ValidateReturnCodes(const char* api_name, VkResult result) const {
-    // This function shouldn't be called if result is equal to VK_SUCCESS.
-    assert(result != VK_SUCCESS);
+// Despite the return code being successful this can be a useful utility for some developers in niche debugging situation.
+void BestPractices::LogSuccess(const char* api_name, VkResult result) const {
+    assert(result > VK_SUCCESS);
```
artem-lunarg, line 106: Sorry, I'm a bit late to this review. Would it be a problem if this function handles VK_SUCCESS too and the caller can filter with `> VK_SUCCESS` if necessary? (it looks this PR already does this). Was thinking mostly about element of surprise, `LogSuccess` name does not suggest some success values are excluded. But if we really want non-VK_SUCCESS values then maybe more detailed name will work too (`LogPositiveSuccessCode` or similar)?

juan-lunarg: > Was thinking mostly about element of surprise, LogSuccess name does not suggest some success values are excluded. 

The layers have been logging success code not equal to VK_SUCCESS for a while now. And the testing is explicit about what they should log.

juan-lunarg: You are right thought the code is a bit confusing:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5719

juan-lunarg: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5720

PR #5702, opened 2023-04-20
layers: Fix ColorBlendState dynamic check

spencer-lunarg:
This will be fixed in the [248 spec](https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5862), but `VK_DYNAMIC_STATE_COLOR_BLEND_ADVANCED_EXT` was missing from the list of dynamic states that effect the Color Blend State


juan-lunarg (approved):
LGTM.

Nitpick uses std::array for tests.

spencer-lunarg:
> Nitpick uses std::array for tests.

There are more places this is like that, will make a "fix them all" PR quick

PR #5703, opened 2023-04-20
layers: Hash ResourceInterfaceVariable descriptor members

spencer-lunarg:
@Tony-LunarG pointed out https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/898721a7a9e98689e0b8ec75ebc77896a295cdf4 gave Doom a 20% decrease. At first I thought it might be of doing a `copy` instead of a `move` but then realize we use to check if descriptors were different from the `DescriptorReqFlags` we had. In that commit I used the Variable ID instead. I think what is happening is a lot of variables have the same properties (Image type, is_atomic, sampled operation type), and so I created a hash to check if worth re-validating

Will wait for Tony to confirm this was the issue for the performance regression before debating how to possibly improve the design


Tony-LunarG:
Unfortunately, this change did not improve performance.

PR #5705, opened 2023-04-20
Label new 248 spec VUIDs

spencer-lunarg:
@mikes-lunarg - Here are the new VUIDs from the new spec, most the churn was mind trying to reduce the dynamic state VUs that were the same

Also 2 more `UNASSIGNED` VUs removed

### DIFF
--- a/layers/core_checks/cc_shader.cpp
+++ b/layers/core_checks/cc_shader.cpp
@@ -2987,7 +2987,7 @@ bool CoreChecks::ValidateInterfaceBetweenStages(const SHADER_MODULE_STATE &produ
                     const LogObjectList objlist(producer.vk_shader_module(), consumer.vk_shader_module());
                     skip |= LogError(objlist, "VUID-RuntimeSpirv-OpVariable-08746",
                                      "vkCreateGraphicsPipelines(): pCreateInfos[%" PRIu32 "] at Location %" PRIu32
-                                     " Comonent %" PRIu32 " Tessellation Control is %s while Tessellation Evalutaion is %s",
+                                     " Component %" PRIu32 " Tessellation Control is %s while Tessellation Evaluation is %s",
                                      pipe_index, location, component, input_var->is_patch ? "patch" : "vertex",
                                      output_var->is_patch ? "patch" : "vertex");
                 }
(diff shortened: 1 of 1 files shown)


layers/core_checks/cc_shader.cpp
```diff
@@ -2994,7 +2994,7 @@ bool CoreChecks::ValidateInterfaceBetweenStages(const SHADER_MODULE_STATE &produ
                     (consumer_stage == VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT) &&
                     (input_var->is_patch != output_var->is_patch)) {
                     const LogObjectList objlist(producer.vk_shader_module(), consumer.vk_shader_module());
-                    skip |= LogError(objlist, kVUID_Core_Shader_InterfacePatchVertex,
+                    skip |= LogError(objlist, "VUID-RuntimeSpirv-OpVariable-08746",
                                      "vkCreateGraphicsPipelines(): pCreateInfos[%" PRIu32 "] at Location %" PRIu32
                                      " Comonent %" PRIu32 " Tessellation Control is %s while Tessellation Evalutaion is %s",
```
arno-lunarg, line 2999:
```suggestion
                                     " Component %" PRIu32 " Tessellation Control is %s while Tessellation Evaluation is %s",
```
For a future PR!

arno-lunarg (approved):
LGTM, one question

layers/core_checks/cc_shader.cpp
```diff
@@ -2994,7 +2994,7 @@ bool CoreChecks::ValidateInterfaceBetweenStages(const SHADER_MODULE_STATE &produ
                     (consumer_stage == VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT) &&
                     (input_var->is_patch != output_var->is_patch)) {
                     const LogObjectList objlist(producer.vk_shader_module(), consumer.vk_shader_module());
-                    skip |= LogError(objlist, kVUID_Core_Shader_InterfacePatchVertex,
+                    skip |= LogError(objlist, "VUID-RuntimeSpirv-OpVariable-08746",
```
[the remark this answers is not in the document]
spencer-lunarg: Yes, good question, so there is a overall "how `Decorations` need to be matched" and we can break that into 2 checks

1. For Builtins
2. For User Defined Variables

the first is done below, the 2nd still needs to be done, but for tessellation we had this already

layers/core_checks/cc_shader.cpp
spencer-lunarg: have to rebase, will fix now

artem-lunarg (approved):

Issue #5649, opened 2023-04-14
Segmentation fault on video playback with libplacebo's plplay, 1.3.243


quietvoid:
**Environment:**
 - OS: Arch Linux, Linux 6.1.23
 - GPU: Advanced Micro Devices, Inc. [AMD/ATI] Navi 21 [Radeon RX 6900 XT] [1002:73af] (rev c0)
 - SDK or header version if building from repo: 1.3.243
 - Options enabled (synchronization, best practices, etc.): Not sure.

**Describe the Issue**

Trying to open a file for playback with `plplay`, I get a segmentation fault seemingly caused by Vulkan Validation Layers.

Part of the trace:
```
#0  radv_get_image_format_properties() () at ../mesa-23.0.1/src/amd/vulkan/radv_formats.c:1471
#1  0x00007fffe40d4f2f in radv_GetPhysicalDeviceImageFormatProperties2() ()
    at ../mesa-23.0.1/src/amd/vulkan/radv_formats.c:1755
#2  0x00007fffdaf7ed46 in DispatchGetPhysicalDeviceImageFormatProperties2(VkPhysicalDevice_T*, VkPhysicalDeviceImageFormatInfo2 const*, VkImageFormatProperties2*)
    (pImageFormatProperties=0x7ffffff58510, pImageFormatInfo=0x7ffffff584e0, physicalDevice=0x555556224a60)
    at /usr/src/debug/vulkan-validation-layers/Vulkan-ValidationLayers-sdk-1.3.243.0/layers/generated/layer_chassis_dispatch.cpp:4161
```

**Steps to reproduce**
Sample file: https://0x0.st/H8ZK.mkv

1. Install Vulkan-Validation-Layers 1.3.243
2. Install [libplacebo v5.264.1](https://code.videolan.org/videolan/libplacebo/-/tags/v5.264.1) with demos for `plplay`
  ```
  meson setup --buildtype=debugoptimized build \
    -D vulkan=enabled \
    -D glslang=enabled \
    -D shaderc=enabled \
    -D lcms=enabled \
    -D d3d11=disabled \
    -D tests=true \
    -D demos=true

  ninja -C build
  ```
3. Run `build/demos/plplay -v sample.mkv`
4. The app segfaults before the window/output shows.

**Expected behavior**
The `plplay` window should open and the video should be rendered.

**Valid Usage ID**
Not sure.

**Additional context**

[...]

PR #5706, opened 2023-04-20
layers: Fix UB for VK_IMAGE_TILING_DRM_FORMAT_MODIFIER_EXT

haasn:
Several codepaths in the layers called into
vkGetPhysicalDeviceImageFormatProperties2KHR without including a valid VkPhysicalDeviceImageDrmFormatModifierInfoEXT, despite this being a requirement for images with DRM format modifier tiling.

For the vkBindImageMemory check, there is a hypothetical possibility of the driver returning VK_ERROR_OUT_OF_HOST_MEMORY when querying the image's DRM format. In this case, for lack of anything better to do, we preserve the current behavior.

It might make more sense to query the DRM format modifier immediately after image creation and cache it inside the image state, but I didn't want to make that (more invasive) change myself.

Fixes: https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5649
Supersedes: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5688

### DIFF
--- a/layers/core_checks/cc_image.cpp
+++ b/layers/core_checks/cc_image.cpp
@@ -168,30 +168,35 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
         }
     }
 
-    VkImageFormatProperties format_limits = {};
+    auto image_format_properties = LvlInitStruct<VkImageFormatProperties2>();
+    auto image_format_info = LvlInitStruct<VkPhysicalDeviceImageFormatInfo2>();
+    image_format_info.type = pCreateInfo->imageType;
+    image_format_info.format = pCreateInfo->format;
+    image_format_info.tiling = pCreateInfo->tiling;
+    image_format_info.usage = pCreateInfo->usage;
+    image_format_info.flags = pCreateInfo->flags;
     VkResult result = VK_SUCCESS;
     if (pCreateInfo->tiling != VK_IMAGE_TILING_DRM_FORMAT_MODIFIER_EXT) {
-        result = DispatchGetPhysicalDeviceImageFormatProperties(physical_device, pCreateInfo->format, pCreateInfo->imageType,
-                                                                pCreateInfo->tiling, pCreateInfo->usage, pCreateInfo->flags,
-                                                                &format_limits);
+        if (IsExtEnabled(device_extensions.vk_khr_get_physical_device_properties2)) {
+            result = DispatchGetPhysicalDeviceImageFormatProperties2(physical_device, &image_format_info, &image_format_properties);
+        } else {
+            result = DispatchGetPhysicalDeviceImageFormatProperties(physical_device, pCreateInfo->format, pCreateInfo->imageType,
+                                                                    pCreateInfo->tiling, pCreateInfo->usage, pCreateInfo->flags,
+                                                                    &image_format_properties.imageFormatProperties);
+        }
     } else {
         auto modifier_list = LvlFindInChain<VkImageDrmFormatModifierListCreateInfoEXT>(pCreateInfo->pNext);
         auto explicit_modifier = LvlFindInChain<VkImageDrmFormatModifierExplicitCreateInfoEXT>(pCreateInfo->pNext);
+        auto drm_format_modifier = LvlInitStruct<VkPhysicalDeviceImageDrmFormatModifierInfoEXT>();
+        image_format_info.pNext = &drm_format_modifier;
+        drm_format_modifier.sharingMode = pCreateInfo->sharingMode;
+        drm_format_modifier.queueFamilyIndexCount = pCreateInfo->queueFamilyIndexCount;
+        drm_format_modifier.pQueueFamilyIndices = pCreateInfo->pQueueFamilyIndices;
         if (modifier_list) {
             for (uint32_t i = 0; i < modifier_list->drmFormatModifierCount; i++) {
-                auto drm_format_modifier = LvlInitStruct<VkPhysicalDeviceImageDrmFormatModifierInfoEXT>();
                 drm_format_modifier.drmFormatModifier = modifier_list->pDrmFormatModifiers[i];
-                auto image_format_info = LvlInitStruct<VkPhysicalDeviceImageFormatInfo2>(&drm_format_modifier);
-                image_format_info.type = pCreateInfo->imageType;
-                image_format_info.format = pCreateInfo->format;
-                image_format_info.tiling = pCreateInfo->tiling;
-                image_format_info.usage = pCreateInfo->usage;
-                image_format_info.flags = pCreateInfo->flags;
-                auto image_format_properties = LvlInitStruct<VkImageFormatProperties2>();
-
                 result =
                     DispatchGetPhysicalDeviceImageFormatProperties2(physical_device, &image_format_info, &image_format_properties);
-                format_limits = image_format_properties.imageFormatProperties;
 
                 /* The application gives a list of modifier and the driver
                  * selects one. If one is wrong, stop there.
@@ -199,18 +204,8 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
                 if (result != VK_SUCCESS) break;
             }
         } else if (explicit_modifier) {
-            auto drm_format_modifier = LvlInitStruct<VkPhysicalDeviceImageDrmFormatModifierInfoEXT>();
             drm_format_modifier.drmFormatModifier = explicit_modifier->drmFormatModifier;
-            auto image_format_info = LvlInitStruct<VkPhysicalDeviceImageFormatInfo2>(&drm_format_modifier);
-            image_format_info.type = pCreateInfo->imageType;
-            image_format_info.format = pCreateInfo->format;
-            image_format_info.tiling = pCreateInfo->tiling;
-            image_format_info.usage = pCreateInfo->usage;
-            image_format_info.flags = pCreateInfo->flags;
-            auto image_format_properties = LvlInitStruct<VkImageFormatProperties2>();
-
             result = DispatchGetPhysicalDeviceImageFormatProperties2(physical_device, &image_format_info, &image_format_properties);
-            format_limits = image_format_properties.imageFormatProperties;
         }
     }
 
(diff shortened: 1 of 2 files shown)


layers/core_checks/cc_image.cpp
```diff
@@ -468,8 +468,29 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
 
         auto external_image_properties = LvlInitStruct<VkExternalImageFormatProperties>();
         auto image_properties = LvlInitStruct<VkImageFormatProperties2>(&external_image_properties);
-        result = DispatchGetPhysicalDeviceImageFormatProperties2(physical_device, &image_info, &image_properties);
-        const auto compatible_types = external_image_properties.externalMemoryProperties.compatibleHandleTypes;
+        VkExternalMemoryHandleTypeFlags compatible_types = 0;
```
spencer-lunarg, line 471: this is the same idea as https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5688 , but I seem to like this better @emersion can you confirm this solves your problem

I am happy to get this in and I can write a test (for regression) after to help unblock people

layers/core_checks/cc_image.cpp
```diff
@@ -468,8 +468,29 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
...
+        if (pCreateInfo->tiling != VK_IMAGE_TILING_DRM_FORMAT_MODIFIER_EXT) {
+            result = DispatchGetPhysicalDeviceImageFormatProperties2(physical_device, &image_info, &image_properties);
+            compatible_types = external_image_properties.externalMemoryProperties.compatibleHandleTypes;
+        } else {
+            auto modifier_list = LvlFindInChain<VkImageDrmFormatModifierListCreateInfoEXT>(pCreateInfo->pNext);
```
emersion, line 476: Since `external_memory_create_info != nullptr`, it's illegal to pass a modifier list. IOW, when importing a DMA-BUF, a single modifier must be specified.

layers/core_checks/cc_image.cpp
```diff
@@ -468,8 +468,29 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
...
+            if (modifier_list) {
+                for (uint32_t i = 0; i < modifier_list->drmFormatModifierCount; i++) {
+                    drm_format_modifier.drmFormatModifier = modifier_list->pDrmFormatModifiers[i];
+                    result = DispatchGetPhysicalDeviceImageFormatProperties2(physical_device, &image_info, &image_properties);
+                    compatible_types |= external_image_properties.externalMemoryProperties.compatibleHandleTypes;
+                    if (result != VK_SUCCESS)
+                        break;
+                }
+            } else if (explicit_modifier) {
+                drm_format_modifier.drmFormatModifier = explicit_modifier->drmFormatModifier;
```
emersion, line 489: Should maybe set `sharingMode` here?

layers/core_checks/cc_image.cpp
emersion: > I seem to like this better

Can you elaborate?

haasn:
Oh, I totally didn't realize me and @emersion worked on the same fix for the same bug (tracked in separate issues) at the same time. My bad.

layers/core_checks/cc_image.cpp
emersion: This patch fixes both export and import, while https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5688 fixes only import.

emersion:
Yeah, sorry, I completely missed https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5649 as well!

haasn:
For the missing positive check, what I do in libplacebo is use the vulkan API to create a DRM tiled image and export it as a DMABUF, then re-import that back into the vulkan device as a second VkImage. This does not require any non-Vulkan calls to accomplish.

My test case can be found [here](https://code.videolan.org/videolan/libplacebo/-/blob/master/src/tests/gpu_tests.h#L1548) but it's entirely written around my own high-level wrappers.

The relevant bits are the `pl_tex_create` calls, which ultimately resolve to [vk_tex_create](https://code.videolan.org/videolan/libplacebo/-/blob/master/src/vulkan/gpu_tex.c#L230), [slab_alloc](https://code.videolan.org/videolan/libplacebo/-/blob/master/src/vulkan/malloc.c#L365) and [vk_malloc_import](https://code.videolan.org/videolan/libplacebo/-/blob/master/src/vulkan/malloc.c#L746)

It is quite a lot of code but copy/pasting it should at least be somewhat straightforward.

layers/core_checks/cc_image.cpp
haasn: Correct me if I'm wrong, but `VkExternalMemoryImageCreateInfo` is also passed when *exporting* a VkImage to a DMA-BUF, in which case a modifier list is perfectly legal. In fact, this is what my code does. (But maybe my code is wrong)

layers/core_checks/cc_image.cpp
haasn: I can add this, though note that the existing code for the DRM format validation also does not set `sharingMode` for this call. Should we also fix that code path?

layers/core_checks/cc_image.cpp
emersion: Hm right… You are correct!

haasn:
Updated to set `sharingMode` correctly, including for the existing calls to this function.

haasn:
Updated to also set `sharingMode` in the `vkBindImageMemory` check, for completeness' sake, and tagged @emersion as co-author.

spencer-lunarg (dismissed):

juan-lunarg (commented):
LGTM overall. Main concern is a test should be added to prevent a regression.

haasn:
> CI Vulkan-ValidationLayers build # 11742 failed.

Can I get details on this failure somehow?

arno-lunarg:
It looks like a CI problem. I restarted a test run, we will see how it goes

juan-lunarg:
> Can I get details on this failure somehow?

Currently LunarG has internal jenkins testing to test on real GPU hardware. Sadly OS contributors don't have an easy way to access these logs. We are working on improving the process for OS contributors.

ncesario-lunarg (dismissed):

juan-lunarg:
> CI Vulkan-ValidationLayers build # 11754 failed.

Failure doesn't look sporadic.

It's failing on `VkPositiveLayerTest.SecondaryCommandBufferBarrier` for `Linux-Mesa-6800` and `VkPositiveLayerTest.OwnershipTranfersImage` Windows-AMD.

@haasn if you have an AMD GPU could you try running those tests locally? If not I'll look into it.

juan-lunarg (changes requested):
Passes MockICD but fails on real hardware.

juan-lunarg:
> It's failing on VkPositiveLayerTest.SecondaryCommandBufferBarrier for Linux-Mesa-6800

I rebased the PR via the github UI.

I wasn't able to repro the test failure on my local system. Running jenkins CI again.

juan-lunarg:
> CI Vulkan-ValidationLayers build # 11789 failed.

It's failing the following tests on AMD/Android GPUs with access violations.

- `VkPositiveLayerTest.OwnershipTranfersImage` (AMD)
- `VkPositiveLayerTest.SecondaryCommandBufferBarrier` (AMD)
- `VkLayerTest.PSOPolygonModeInvalid` (Android)

Passes fine on Nvidia / MacOS... not sure why yet.

juan-lunarg:
Rebased on main to address merge conflict. Trying to figure out testing failure to see if we can get this in for the upcoming SDK release.

juan-lunarg:
Also FWIW @haasn `LunarG CI Checkrun` now shows the results of internal CI to external contributors. So if you are planning on future contributions it will be a smoother experience.

layers/core_checks/cc_image.cpp
```diff
@@ -168,49 +168,40 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
...
+    image_format_info.format = pCreateInfo->format;
+    image_format_info.tiling = pCreateInfo->tiling;
+    image_format_info.usage = pCreateInfo->usage;
+    image_format_info.flags = pCreateInfo->flags;
     VkResult result = VK_SUCCESS;
     if (pCreateInfo->tiling != VK_IMAGE_TILING_DRM_FORMAT_MODIFIER_EXT) {
-        result = DispatchGetPhysicalDeviceImageFormatProperties(physical_device, pCreateInfo->format, pCreateInfo->imageType,
-                                                                pCreateInfo->tiling, pCreateInfo->usage, pCreateInfo->flags,
-                                                                &format_limits);
+        if IsExtEnabled(device_extensions.vk_khr_get_physical_device_properties2) { // I believe this is the cause of the crashes
```
juan-lunarg, line 180: FYI @jeremyg-lunarg

ncesario-lunarg: Considering how ubiquitous the query APIs behind `VK_KHR_get_physical_device_properties2` are, I wonder if it makes sense to create a dispatch wrapper similar to what we use in the tests?

jeremyg-lunarg: I would love to have magic dispatch wrappers for pretty much every function that exists with multiple names (EXT, KHR, plain)

layers/core_checks/cc_image.cpp
```diff
@@ -168,49 +168,38 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
...
+    image_format_info.format = pCreateInfo->format;
+    image_format_info.tiling = pCreateInfo->tiling;
+    image_format_info.usage = pCreateInfo->usage;
+    image_format_info.flags = pCreateInfo->flags;
     VkResult result = VK_SUCCESS;
     if (pCreateInfo->tiling != VK_IMAGE_TILING_DRM_FORMAT_MODIFIER_EXT) {
-        result = DispatchGetPhysicalDeviceImageFormatProperties(physical_device, pCreateInfo->format, pCreateInfo->imageType,
-                                                                pCreateInfo->tiling, pCreateInfo->usage, pCreateInfo->flags,
-                                                                &format_limits);
+        result = DispatchGetPhysicalDeviceImageFormatProperties2(physical_device, &image_format_info, &image_format_properties);
```
juan-lunarg, line 180: This is the cause of the CI failures. `GetPhysicalDeviceImageFormatProperties2` isn't supported on all  devices.

Credit to @jeremyg-lunarg for pointing that out.

juan-lunarg (changes requested):

juan-lunarg (approved):
I amended the original commit to check for `vk_khr_get_physical_device_properties2` support.

With that and the new test to prevent a regression I think this PR is good to go.

juan-lunarg:
Reverted my test since it has a minor issue on Android. I'll add the test in another PR so we can get this PR in time for the SDK release.

https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5841

PR #3088, opened 2021-07-30
[WIP] build: Refactor Android build

ncesario-lunarg:
Android build refactor:
- Removes dependency on ndk-build/Android.mk build tools
- Remove shaderc dependency in the Android build for tests
- Use single `known_good.json` for desktop and Android

Still WIP:
- Adding an "easy" way to build and package multiple architectures (may or may not be needed depending on CI)
- Still investigating some test failures

Creating this now to get early feedback on the build process. In short, setting up an Android build would look something like:
```bash
cmake -GNinja -Bbuild-android-cmake \
  -DCMAKE_BUILD_TYPE=Debug \
  -DUPDATE_DEPS=ON \
  -DCMAKE_TOOLCHAIN_FILE=<ndk-path>/build/cmake/android.toolchain.cmake \
  -DANDROID_NDK_HOME=<ndk-path> \
  -DANDROID_SDK_HOME=<sdk-path> \
  -DANDROID_PLATFORM=26 \
  -DANDROID_BUILD_TOOLS=31.0.0-rc1 \
  -DANDROID_ABI=arm64-v8a \
  -DANDROID_STL=c++_static
```
I removed the need for using environment variables to set `ANDROID_{S,N}DK_HOME` as I'm not personally a fan of this (though you can still do it that way if you prefer with this change). Also, using `UPDATE_DEPS=ON` this should be a similar "one-time-setup" as with desktop.

FYI @Tony-LunarG @jeremyg-lunarg @ziga-lunarg @mikes-lunarg.

FYI @lunarpapillo @johnzupin as this would eventually require some changes to CI. The big one (I'm guessing) is that each arch would need its own cmake build (instead of specifying all archs in `Application.mk`).

PR #5709, opened 2023-04-20
Android CMake Draft

juan-lunarg:
Tests compile cleanly with CMake Android build which should mean it's equivalent to our current Android.mk system.

Haven't run anything yet, but I don't see any reason why things shouldn't just work. Fingers crossed.

Looking for help on how to proceed given we have existing processes with our current Android situation.

CI: https://github.com/juan-lunarg/Vulkan-ValidationLayers/actions/runs/4759672005


juan-lunarg:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/3088/files

juan-lunarg:
Talked with @ncesario-lunarg currently have a plan to tackle this incrementally

PR #5713, opened 2023-04-21
layers: Revert ActiveSlot mapping type

spencer-lunarg:
Attempt to fix 898721a7a9e98689e0b8ec75ebc77896a295cdf4 perf issues

This shows the perf issues resolved, if I can't figure out another way, I would like to get this in

### DIFF
--- a/layers/state_tracker/pipeline_state.cpp
+++ b/layers/state_tracker/pipeline_state.cpp
@@ -135,7 +135,39 @@ PIPELINE_STATE::ActiveSlotMap PIPELINE_STATE::GetActiveSlots(const StageStateVec
         // Capture descriptor uses for the pipeline
         for (const auto &variable : stage.entrypoint->resource_interface_variables) {
             // While validating shaders capture which slots are used by the pipeline
-            active_slots[variable.decorations.set][variable.decorations.binding] = &variable;
+            auto &entry = active_slots[variable.decorations.set][variable.decorations.binding];
+            entry.variable = &variable;
+
+            auto &reqs = entry.reqs;
+            if (variable.is_atomic_operation) reqs |= DESCRIPTOR_REQ_VIEW_ATOMIC_OPERATION;
+            if (variable.is_sampler_sampled) reqs |= DESCRIPTOR_REQ_SAMPLER_SAMPLED;
+            if (variable.is_sampler_implicitLod_dref_proj) reqs |= DESCRIPTOR_REQ_SAMPLER_IMPLICITLOD_DREF_PROJ;
+            if (variable.is_sampler_bias_offset) reqs |= DESCRIPTOR_REQ_SAMPLER_BIAS_OFFSET;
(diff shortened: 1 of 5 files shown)


Tony-LunarG:
This does restore the trace playback performance to the FPS seen before https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/898721a7a9e98689e0b8ec75ebc77896a295cdf4

arno-lunarg (approved):

layers/state_tracker/pipeline_state.cpp
```diff
@@ -135,7 +135,39 @@ PIPELINE_STATE::ActiveSlotMap PIPELINE_STATE::GetActiveSlots(const StageStateVec
         // Capture descriptor uses for the pipeline
         for (const auto &variable : stage.entrypoint->resource_interface_variables) {
             // While validating shaders capture which slots are used by the pipeline
-            active_slots[variable.decorations.set][variable.decorations.binding] = &variable;
+            auto &entry = active_slots[variable.decorations.set][variable.decorations.binding];
+            entry.variable = &variable;
+
+            auto &reqs = entry.reqs;
+            if (variable.is_atomic_operation) reqs |= DESCRIPTOR_REQ_VIEW_ATOMIC_OPERATION;
```
ncesario-lunarg, line 142: Are all of these `if`s suspected to be the perf impact? Just curious as it doesn't look like this is reverting very much.

ncesario-lunarg (approved):
LGTM if it restores the perf regression.

layers/state_tracker/pipeline_state.cpp
spencer-lunarg: No, its a bit more complex than that...

So it is more clear to see in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5703/files

Basically there are 2 things happening here

1.  The mapping that the descriptor logic uses

```patch
- typedef std::map<uint32_t, const ResourceInterfaceVariable *> BindingVariableMap;
+ typedef std::map<uint32_t, DescriptorRequirement> BindingVariableMap;
```

2. The logic to know if we need to re-validate (because it saves a LOT of time it seems to not check if the `OpVariable` offers nothing new

```patch
+ inline bool operator==(const DescriptorRequirement &a, const DescriptorRequirement &b) noexcept { return a.reqs == b.reqs; }
+ inline bool operator<(const DescriptorRequirement &a, const DescriptorRequirement &b) noexcept { return a.reqs < b.reqs; }
```

For this we need to know what values are using what. In the `5703 PR` I attempted to that to a "Hash" in the `ResourceInterfaceVariable` object. It is very error prone to not have a way to know if you need a new variable in the Descriptor validation, it has to be part of the "re-validate check"

spencer-lunarg:
I am merging this in for now, it solves the perf regression, and still makes improvements in the code from what it use to be

I am doing on-going testing of the issue and planning to have a follow up PR

PR #5716, opened 2023-04-21
build: Update to header 1.3.248

mikes-lunarg:
- Update known-good
- Generate source

Tiny update, VUID changes are here: #5705


ncesario-lunarg:
@mikes-lunarg, assuming #5708 passes review and CI, can we get that in first so that is included in the v1.3.248 tag?

mikes-lunarg:
> @mikes-lunarg, assuming #5708 passes review and CI, can we get that in first so that is included in the v1.3.248 tag?

Sure!

ncesario-lunarg (approved):
LGTM!

PR #5718, opened 2023-04-21
Android cleanup. Minor C++ inconsistency and CMake updates

juan-lunarg:
- android: Use VK_USE_PLATFORM_ANDROID_KHR consistently
- Android CMake build fixes/cleanup


.github/workflows/vvl.yml
juan-lunarg, line 1: For those interested you can check ccache utilization via the github actions:
image

Which is how I know ccache isn't working on android

spencer-lunarg (approved):
sorry you had to learn a lot of why we do stuff in Android here, but this looks great

Issue #5662, opened 2023-04-17
Incorrect sync validation feature check


rg3igalia:
**Environment:**
 - OS: Linux
 - GPU: RX 6700 XT
 - SDK or header version if building from repo: ddd3f7ee9d6d87307e3e85e8860d085880998068
 - Options enabled (synchronization, best practices, etc.): defaults+synchronization

**Describe the Issue**

Some CTS attachment rate tests (e.g. `dEQP-VK.fragment_shading_rate.renderpass2.monolithic.attachment_rate.setup_with_copying_using_transfer_queue_concurent.r8_uint.rate_1x1`) are hitting the following VU:

```
Validation Error: [ VUID-vkCmdPipelineBarrier-dstStageMask-04097 ] Object 0: handle = 0x555562be7aa0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x6d2c788e | vkCmdPipelineBarrier(): .dstStageMask includes VK_PIPELINE_STAGE_FRAGMENT_SHADING_RATE_ATTACHMENT_BIT_KHR when the device does not have shadingRate feature enabled. The Vulkan spec states: If the shadingRateImage feature is not enabled, pname:dstStageMask must not contain VK_PIPELINE_STAGE_SHADING_RATE_IMAGE_BIT_NV (https://github.com/KhronosGroup/Vulkan-Docs/search?q=)VUID-vkCmdPipelineBarrier-dstStageMask-04097)
```

The VUID in question no longer exists in the current Vulkan spec but it's been replaced with `VUID-vkCmdPipelineBarrier-dstStageMask-07318` (and other related similar VUs), which says:

> If neither the [shadingRateImage](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#features-shadingRateImage) or [attachmentFragmentShadingRate](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#features-attachmentFragmentShadingRate) are enabled, dstStageMask must not contain VK_PIPELINE_STAGE_FRAGMENT_SHADING_RATE_ATTACHMENT_BIT_KHR

Synchronization validation is checking the wrong feature in this case, using `pipelineFragmentShadingRate` instead of `attachmentFragmentShadingRate` as the VU text mentions.

The patch below fixes the issue for me.

```diff
diff --git a/layers/sync/sync_utils.cpp b/layers/sync/sync_utils.cpp
[...]

PR #5721, opened 2023-04-23
layers: Shading rate stage fixes

artem-lunarg:
The first commit fixes `VK_PIPELINE_STAGE_FRAGMENT_SHADING_RATE_ATTACHMENT_BIT_KHR` detection and closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5662

UPDATE: @ziga-lunarg independently and simultaneously implemented fix for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5662 in a different PR https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5726 . This PR integrates the test from that PR.

The second commit adds shading rate VUIDs that vary only based on enabled extensions.

### DIFF
--- a/layers/sync/sync_vuid_maps.cpp
+++ b/layers/sync/sync_vuid_maps.cpp
@@ -96,30 +96,6 @@ static const std::map<VkPipelineStageFlags2KHR, std::vector<Entry>> kStageMaskEr
          {Key(Struct::VkSubpassDependency2, Field::srcStageMask), "VUID-VkSubpassDependency2-srcStageMask-04093"},
          {Key(Struct::VkSubpassDependency2, Field::dstStageMask), "VUID-VkSubpassDependency2-dstStageMask-04093"},
      }},
-    {VK_PIPELINE_STAGE_2_FRAGMENT_SHADING_RATE_ATTACHMENT_BIT_KHR,
-     {
-         {Key(Struct::VkBufferMemoryBarrier2, Field::dstStageMask), "VUID-VkBufferMemoryBarrier2-dstStageMask-04956"},
-         {Key(Struct::VkBufferMemoryBarrier2, Field::srcStageMask), "VUID-VkBufferMemoryBarrier2-srcStageMask-04956"},
-         {Key(Func::vkCmdPipelineBarrier, Field::dstStageMask), "VUID-vkCmdPipelineBarrier-dstStageMask-04097"},
-         {Key(Func::vkCmdPipelineBarrier, Field::srcStageMask), "VUID-vkCmdPipelineBarrier-srcStageMask-04097"},
-         {Key(Func::vkCmdResetEvent2, Field::stageMask), "VUID-vkCmdResetEvent2-stageMask-04956"},
-         {Key(Func::vkCmdResetEvent, Field::stageMask), "VUID-vkCmdResetEvent-stageMask-04097"},
-         {Key(Func::vkCmdSetEvent, Field::stageMask), "VUID-vkCmdSetEvent-stageMask-04097"},
-         {Key(Func::vkCmdWaitEvents, Field::dstStageMask), "VUID-vkCmdWaitEvents-dstStageMask-04097"},
-         {Key(Func::vkCmdWaitEvents, Field::srcStageMask), "VUID-vkCmdWaitEvents-srcStageMask-04097"},
-         {Key(Func::vkCmdWriteTimestamp2, Field::stage), "VUID-vkCmdWriteTimestamp2-stage-04956"},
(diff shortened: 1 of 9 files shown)


spencer-lunarg (approved):

arno-lunarg:
I had a similar problem to solve in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4976 with the "choose the right VUIDs based on enabled extensions". I also had the problem of validation slightly changing between the different VUIDs, not sure this is the case here. 
I went for a more data oriented approach, to avoid code duplication. The arrays introduced in this PR are essentially the same, but for the VUID strings, so I wonder if the approach I took could also have some value here?

artem-lunarg:
> The arrays introduced in this PR are essentially the same, but for the VUID strings

That's right, it's very similar arrays. Initially I implemented the logic to minimize data duplication: I had array of _shared prefixes_ and then a separate array of the part that varies - _numerical suffixes_ (as a side note, later I found it won't work because there is exception for vkCmdWriteTimestamp). I changed that approach because we had internal conversation (VVL space) and it was recommended to keep VUIDs in the full form, so coverage statistics tools can find it and also regular code search.

I did not investigate your approach in details (sorry!), do it require encoding VUIDs somehow?

artem-lunarg:
From what I briefly checked, [your ](https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4976) PR specifies full identifiers (?), so it should not be a problem. Specifically for this change I think it's fine as it is, _for now_. There is no per-VUID variation of logic. And the logic for selecting VUID is one condition: `return nv_enabled ? vuid1 : vuid2`. Still it's probably a good idea if this somehow can be autogenerated, especially if we have more similar use-case.

tests/positive/fragment_shading_rate.cpp
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright (c) 2015-2023 The Khronos Group Inc.
+ * Copyright (c) 2015-2023 Valve Corporation
+ * Copyright (c) 2015-2023 LunarG, Inc.
+ * Copyright (c) 2015-2023 Google, Inc.
```
juan-lunarg, line 5:
```suggestion
 * Copyright (c) 2023 Valve Corporation
 * Copyright (c) 2023 LunarG, Inc.
```

arno-lunarg:
Yes just like you one solution for me would have been to encode VUIDs, but as you said the trade off is not worth it.
I prefer the single array with `{key, vuid1, vuid2, vuid3}`! Less copy paste and probably easier to extend

artem-lunarg:
@arno-lunarg, sorry, I just found the reason why I did not do this initially. The code uses existing `Entry` type and `FindVUID` which operates on arrays of entries, so with current design that's how it works...

layers/sync/sync_vuid_maps.cpp
```diff
@@ -298,11 +274,95 @@ static const std::array<Entry, 12> kStageMaskErrorsNoneWithoutSync2{{
...
     }
+
+    // special case for FRAGMENT_SHADING_RATE_ATTACHMENT because it depends on KHR and/or NV extensions being enabled.
+    // NOTE: the kStageMaskErrorsShadingRateNV table above is only for the VUID coverage tooling and corresponding vuids
+    // are not used for reporting. Those vuids can only be found in the NV-but-not-KHR version of the specification, which
+    // is not one of the published documents. Instead the following approach is used: if NV extension is enabled then report
+    // KHR+NV vuid (all extensions spec), otherwise report KHR vuid (khr extensions spec).
+    if (bit == VK_PIPELINE_STAGE_2_FRAGMENT_SHADING_RATE_ATTACHMENT_BIT_KHR) {
+        const bool nv_enabled = IsExtEnabled(device_extensions.vk_nv_shading_rate_image);
+        return nv_enabled ? FindVUID(loc, kStageMaskErrorsShadingRateKHRAndNV) : FindVUID(loc, kStageMaskErrorsShadingRateKHR);
```
ziga-lunarg, line 363: Shouldn't this be something like:
```
if (khr && nv)
    FindVUID(loc, kStageMaskErrorsShadingRateKHRAndNV)
else if (khr)
    FindVUID(loc, kStageMaskErrorsShadingRateKHR)
else if (nv)
    FindVUID(loc, kStageMaskErrorsShadingRateNV)
```

artem-lunarg: The comment explains the logic. But I can rephrase in this form: user enables NV extension and gets VUID that can be found only in spec that is complied with NV extension but without KHR, which is not a  standard document. The decision here was to provide KHR+NV vuid which can be found in published document and contains all information to fix the issue.

Issue #2361, opened 2020-12-01
layers: Add finer VkFormat size information for VUID 04466


ncesario-lunarg:
**Add more detailed size information to vk_format_utils**
VUID 04466 requires bit-per-channel information which is currently not present in the `VULKAN_FORMAT_INFO` struct. This information should probably be generated by the python scripts.

**Valid Usage ID**
- VUID-VkImageViewCreateInfo-imageViewFormatReinterpretation-04466

This is part of #2208.

FYI @mark-lunarg.

ncesario-lunarg:
@sfricke-samsung has [started work](https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/4787/) to get this information into `vk.xml`; once that change gets merged, automation should be updated to generate `vk_fomat_utils.cpp`.

Rua:
What would be the number of bits per component in the case of block-compressed formats? Is that even well-defined?

sfricke-samsung:
I "**think**" with block-compressed formats its not, but at same time, the issue about is where Image and ImageView have different formats, so in that case, if block-compressed, it would have to just be the same format

Rua:
What if the image was created with `VK_IMAGE_CREATE_BLOCK_TEXEL_VIEW_COMPATIBLE_BIT`, and the image view has an uncompressed format?

PR #5722, opened 2023-04-24
layers: Use FormatUtils for imageViewFormatReinterpretation

spencer-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/2361

Before (when issue was opened 2+ years ago) we didn't have a generated Format Util that could check against each component bit size

Even if it is compressed formats, it will still match correctly

### DIFF
--- a/layers/generated/vk_format_utils.cpp
+++ b/layers/generated/vk_format_utils.cpp
@@ -1880,3 +1880,30 @@ bool FormatHasAlpha(VkFormat format) {
     return FormatHasComponentType(format, COMPONENT_TYPE::A);
 }
 
+bool FormatsSameComponentBits(VkFormat format_a, VkFormat format_b) {
+    const auto item_a = kVkFormatTable.find(format_a);
+    const auto item_b = kVkFormatTable.find(format_b);
+    if (item_a == kVkFormatTable.end() || item_b == kVkFormatTable.end()) {
+        return false;
+    } else if (item_a->second.component_count != item_b->second.component_count) {
+        return false;
+    }
+    // Need to loop match each component type is found in both formats
+    // formats are maxed at 4 components, so the double loop is not going to scale
+    for (uint32_t i = 0; i < item_a->second.component_count; i++) {
+        const auto& component_a = item_a->second.components[i];
+        bool component_match = false;
+        for (uint32_t j = 0; j < item_b->second.component_count; j++) {
+            const auto& component_b = item_b->second.components[j];
+            if ((component_a.type == component_b.type) && (component_a.size == component_b.size)) {
+                component_match = true;
+                break;
+            }
+        }
+        if (!component_match) {
+            return false;
+        }
+    }
+    return true;
+}
+
--- a/scripts/format_utils_generator.py
+++ b/scripts/format_utils_generator.py
@@ -703,6 +703,7 @@ bool FormatHasRed(VkFormat format);
 bool FormatHasGreen(VkFormat format);
 bool FormatHasBlue(VkFormat format);
 bool FormatHasAlpha(VkFormat format);
+bool FormatsSameComponentBits(VkFormat format_a, VkFormat format_b);
 '''
         elif self.sourceFile:
             output += '''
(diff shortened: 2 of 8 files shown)


layers/generated/vk_format_utils.cpp
```diff
@@ -1935,3 +1935,19 @@ bool FormatHasAlpha(VkFormat format) {
     return FormatHasComponentType(format, COMPONENT_TYPE::A);
 }
 
+bool FormatsSameComponentBits(VkFormat format_a, VkFormat format_b) {
+    auto item_a = kVkFormatTable.find(format_a);
+    auto item_b = kVkFormatTable.find(format_b);
+    if (item_a == kVkFormatTable.end() || item_b == kVkFormatTable.end()) {
+        return false;
+    } else if (item_a->second.component_count != item_a->second.component_count) {
```
arno-lunarg, line 1943:
```suggestion
    } else if (item_a->second.component_count != item_b->second.component_count) {
```

scripts/format_utils_generator.py
```diff
@@ -807,6 +808,22 @@ def sizeFunctions(self):
...
 bool FormatHasAlpha(VkFormat format) {
     return FormatHasComponentType(format, COMPONENT_TYPE::A);
+}
+
+bool FormatsSameComponentBits(VkFormat format_a, VkFormat format_b) {
+    auto item_a = kVkFormatTable.find(format_a);
+    auto item_b = kVkFormatTable.find(format_b);
+    if (item_a == kVkFormatTable.end() || item_b == kVkFormatTable.end()) {
+        return false;
+    } else if (item_a->second.component_count != item_a->second.component_count) {
```
arno-lunarg, line 818:
```suggestion
    } else if (item_a->second.component_count != item_b->second.component_count) {
```

arno-lunarg (approved):
LGTM! One needed change

scripts/format_utils_generator.py
spencer-lunarg: :sob: thanks for catching this

arno-lunarg: I have done it so much time in the past, and will do it countless times in the future, so be wary of my PRs ( ͡• ͜ʖ ͡• )

layers/generated/vk_format_utils.cpp
```diff
@@ -1935,3 +1935,19 @@ bool FormatHasAlpha(VkFormat format) {
     return FormatHasComponentType(format, COMPONENT_TYPE::A);
 }
 
+bool FormatsSameComponentBits(VkFormat format_a, VkFormat format_b) {
+    auto item_a = kVkFormatTable.find(format_a);
+    auto item_b = kVkFormatTable.find(format_b);
```
juan-lunarg, line 1940:
```suggestion
    const auto item_a = kVkFormatTable.find(format_a);
    const auto item_b = kVkFormatTable.find(format_b);
```

layers/generated/vk_format_utils.cpp
```diff
@@ -1935,3 +1935,19 @@ bool FormatHasAlpha(VkFormat format) {
...
+        return false;
+    } else if (item_a->second.component_count != item_b->second.component_count) {
+        return false;
+    }
+    for (uint32_t i = 0; i < item_a->second.component_count; i++) {
+        if (item_a->second.components[i].size != item_b->second.components[i].size) {
+            return false;
+        }
+    }
+    return true;
```
juan-lunarg, line 1951: When adding testing for this function test formats that would hit each of these return statements.

layers/generated/vk_format_utils.cpp
```diff
@@ -1935,3 +1935,19 @@ bool FormatHasAlpha(VkFormat format) {
     return FormatHasComponentType(format, COMPONENT_TYPE::A);
 }
 
+bool FormatsSameComponentBits(VkFormat format_a, VkFormat format_b) {
```
juan-lunarg, line 1938: Add testing for `FormatsSameComponentBits` in `tests/positive/layer_utils.cpp`

juan-lunarg (changes requested):
:+1:

layers/generated/vk_format_utils.cpp
spencer-lunarg: added tests, good call, there were more nasty edge cases that were failing

juan-lunarg (approved):

tests/negative/portability_subset.cpp
```diff
@@ -174,18 +167,52 @@ TEST_F(VkPortabilitySubsetTest, CreateImageView) {
...
+    ci.format = VK_FORMAT_B10G11R11_UFLOAT_PACK32;
+    // Incorrect swizzling due to portability
     ci.components.r = VK_COMPONENT_SWIZZLE_IDENTITY;
     ci.components.g = VK_COMPONENT_SWIZZLE_IDENTITY;
     ci.components.b = VK_COMPONENT_SWIZZLE_IDENTITY;
     ci.components.a = VK_COMPONENT_SWIZZLE_IDENTITY;
-    ci.format = VK_FORMAT_R5G6B5_UNORM_PACK16;  // Wrong number of components
+    ci.subresourceRange = {VK_IMAGE_ASPECT_COLOR_BIT, 0, 1, 0, 1};
+    // Format might not be supported
+    m_errorMonitor->SetUnexpectedError("VUID-VkImageViewCreateInfo-None-02273");
```
juan-lunarg, line 214: Can we detect the ImageView format isn't supported and skip instead?

ncesario-lunarg: :heavy_plus_sign:

ncesario-lunarg (approved):
I completely forgot about this issue :)

tests/negative/portability_subset.cpp
spencer-lunarg: If we do, this will never be tested, I am not sure what format is supported on MacOS that was designed to hit `VUID-VkImageViewCreateInfo-imageViewFormatReinterpretation-04466` this way

Added a comment to look into ti

PR #5726, opened 2023-04-24
Check correct feature for 04097 VUID

ziga-lunarg:
A while ago the spec was clarified that pipeline stage `VK_PIPELINE_STAGE_FRAGMENT_SHADING_RATE_ATTACHMENT_BIT_KHR` requires the feature `attachmentFragmentShadingRate`

Validation here is not completely correct yet, because VUIDs 07318 still need to be implemented.


artem-lunarg (approved):

ziga-lunarg:
@artem-lunarg Didn't see yours, odd timing for this fix, I'll close it as yours fixes this anyway and I don't want to cause conflicts.

PR #5727, opened 2023-04-24
Fix BP illegal access

Tony-LunarG:
Fixes a crash running dEQP-VK.pipeline.monolithic.extended_dynamic_state.cmd_buffer_start.null_color_blend_att_ptr with BP enabled

Spec:
pAttachments is a pointer to an array of [VkPipelineColorBlendAttachmentState] structures defining blend state for each color attachment. It is ignored if the pipeline is created with VK_DYNAMIC_STATE_COLOR_BLEND_ENABLE_EXT, VK_DYNAMIC_STATE_COLOR_BLEND_EQUATION_EXT, and VK_DYNAMIC_STATE_COLOR_WRITE_MASK_EXT dynamic states set.

### DIFF
--- a/layers/best_practices/bp_pipeline.cpp
+++ b/layers/best_practices/bp_pipeline.cpp
@@ -194,25 +194,33 @@ bool BestPractices::PreCallValidateCreateGraphicsPipelines(VkDevice device, VkPi
     return skip;
 }
 
-static std::vector<bp_state::AttachmentInfo> GetAttachmentAccess(const safe_VkGraphicsPipelineCreateInfo& create_info,
-                                                                 std::shared_ptr<const RENDER_PASS_STATE>& rp) {
+static std::vector<bp_state::AttachmentInfo> GetAttachmentAccess(bp_state::Pipeline& pipe_state) {
     std::vector<bp_state::AttachmentInfo> result;
+    auto rp = pipe_state.RenderPassState();
     if (!rp || rp->UsesDynamicRendering()) {
         return result;
     }
-
+    auto& create_info = pipe_state.GetCreateInfo<VkGraphicsPipelineCreateInfo>();
     const auto& subpass = rp->createInfo.pSubpasses[create_info.subpass];
 
     // NOTE: see PIPELINE_LAYOUT and safe_VkGraphicsPipelineCreateInfo constructors. pColorBlendState and pDepthStencilState
     // are only non-null if they are enabled.
     if (create_info.pColorBlendState) {
-        // According to spec, pColorBlendState must be ignored if subpass does not have color attachments.
-        uint32_t num_color_attachments = std::min(subpass.colorAttachmentCount, create_info.pColorBlendState->attachmentCount);
-        for (uint32_t j = 0; j < num_color_attachments; j++) {
-            if (create_info.pColorBlendState->pAttachments[j].colorWriteMask != 0) {
-                uint32_t attachment = subpass.pColorAttachments[j].attachment;
-                if (attachment != VK_ATTACHMENT_UNUSED) {
-                    result.push_back({attachment, VK_IMAGE_ASPECT_COLOR_BIT});
+        // According to the spec, pAttachments is to be ignored if if the pipeline is created with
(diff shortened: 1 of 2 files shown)


juan-lunarg (approved):

artem-lunarg (approved):

layers/best_practices/bp_pipeline.cpp
```diff
@@ -206,13 +213,21 @@ static std::vector<bp_state::AttachmentInfo> GetAttachmentAccess(const safe_VkGr
...
-        for (uint32_t j = 0; j < num_color_attachments; j++) {
-            if (create_info.pColorBlendState->pAttachments[j].colorWriteMask != 0) {
-                uint32_t attachment = subpass.pColorAttachments[j].attachment;
-                if (attachment != VK_ATTACHMENT_UNUSED) {
-                    result.push_back({attachment, VK_IMAGE_ASPECT_COLOR_BIT});
+        // According to the spec, pAttachments is to be ignored if if the pipeline is created with
+        // VK_DYNAMIC_STATE_COLOR_BLEND_ENABLE_EXT, VK_DYNAMIC_STATE_COLOR_BLEND_EQUATION_EXT, and
+        // VK_DYNAMIC_STATE_COLOR_WRITE_MASK_EXT dynamic states set
+        auto p_dynamic = create_info.pDynamicState;
+        if (!(p_dynamic && (IsDynamic(p_dynamic, VK_DYNAMIC_STATE_COLOR_BLEND_ENABLE_EXT) &&
```
ncesario-lunarg, line 220: I wonder if `PIPELINE_STATE` could be passed in to `GetAttachmentAccess` (since it looks like the base is constructed before this function is called) and then make use of the existing `PIPELINE_STATE::IsDynamic`?

ncesario-lunarg (approved):
One question regarding possible simplification, but LGTM!

layers/best_practices/bp_pipeline.cpp
Tony-LunarG: I thought about something like that, but didn't persist.  I'll take another look.

layers/best_practices/bp_pipeline.cpp
```diff
@@ -195,24 +195,32 @@ bool BestPractices::PreCallValidateCreateGraphicsPipelines(VkDevice device, VkPi
 }
 
 static std::vector<bp_state::AttachmentInfo> GetAttachmentAccess(const safe_VkGraphicsPipelineCreateInfo& create_info,
-                                                                 std::shared_ptr<const RENDER_PASS_STATE>& rp) {
+                                                                 std::shared_ptr<const RENDER_PASS_STATE>& rp,
+                                                                 bp_state::Pipeline* pipe_state) {
```
ncesario-lunarg, line 199: Looks like `pipe_state` should be a reference as it is assumed to be non-null.

layers/best_practices/bp_pipeline.cpp
```diff
@@ -195,24 +195,32 @@ bool BestPractices::PreCallValidateCreateGraphicsPipelines(VkDevice device, VkPi
 }
 
 static std::vector<bp_state::AttachmentInfo> GetAttachmentAccess(const safe_VkGraphicsPipelineCreateInfo& create_info,
-                                                                 std::shared_ptr<const RENDER_PASS_STATE>& rp) {
+                                                                 std::shared_ptr<const RENDER_PASS_STATE>& rp,
```
ncesario-lunarg, line 198: `create_info` and `rp` could be removed as parameters and derived from `pipe_state`.

ncesario-lunarg (approved):
Minor nits/suggestions, but LGTM!

layers/best_practices/bp_pipeline.cpp
Tony-LunarG: Done

layers/best_practices/bp_pipeline.cpp
Tony-LunarG: Done

layers/best_practices/bp_pipeline.cpp
Tony-LunarG: Done

Issue #5732, opened 2023-04-25
False positive on descriptor buffer creation


dustContributor:
**Environment:**
 - OS: Debian 64bit
 - GPU: Radeon W6600 (RX 6600), Mesa RADV 23.0.2
 - SDK or header version if building from repo:  1.3.239.0-2 from Debian repositories
 - Options enabled (synchronization, best practices, etc.): Just validation.

**Describe the Issue**

I'm making a buffer for descriptor buffer usage with SHADER_DEVICE_ADDRESS, RESOURCE_DESCRIPTOR_BUFFER_EXT and SAMPLER_DESCRIPTOR_BUFFER_EXT flags, and I get a validation error telling me I'm exceeding the descriptor buffer address space. I'm making a 256 byte allocation, and it's telling me I have 0 address space available.

> ERROR:
> VALIDATION [VUID-VkBufferCreateInfo-usage-08098 (-501398844)] : Validation Error: [ VUID-VkBufferCreateInfo-usage-08098 ] Object 0: handle = 0x555556508220, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0xe21d42c4 | vkCreateBuffer(): Requested buffer size (256) plus current total (0) is greater than specified in properties field descriptorBufferAddressSpaceSize (0). The Vulkan spec states: If usage includes VK_BUFFER_USAGE_RESOURCE_DESCRIPTOR_BUFFER_BIT_EXT, creating this VkBuffer must not cause the total required space for all currently valid buffers using this flag on the device to exceed VkPhysicalDeviceDescriptorBufferPropertiesEXT::resourceDescriptorBufferAddressSpaceSize or VkPhysicalDeviceDescriptorBufferPropertiesEXT::descriptorBufferAddressSpaceSize (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkBufferCreateInfo-usage-08098)

However, my device reports around 4GB for this address space, not zero, as it can be seen in the screenshot. Regardless there are reports in gpuinfo.org of roughly the same device with the same driver here https://vulkan.gpuinfo.org/displayreport.php?id=20065#properties_extensions if anybody wants to take a look at other values.

![device-props-desc-buffer](https://user-images.githubusercontent.com/8086687/234161167-503765ce-4236-4030-b5aa-933421d28dfc.png)

spencer-lunarg:
[...]

PR #5734, opened 2023-04-25
tests: Add descriptorBufferAddressSpaceSize

spencer-lunarg:
tests to check https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5732


layers/core_checks/cc_buffer.cpp
```diff
@@ -211,10 +211,8 @@ bool CoreChecks::PreCallValidateCreateBuffer(VkDevice device, const VkBufferCrea
                 ") is greater than specified in properties field samplerDescriptorBufferAddressSpaceSize (%" PRIuLEAST64 ").",
                 pCreateInfo->size, samplerDescriptorBufferAddressSpaceSize.load(),
                 phys_dev_ext_props.descriptor_buffer_props.samplerDescriptorBufferAddressSpaceSize);
-        }
-
-        if (pCreateInfo->size + descriptorBufferAddressSpaceSize >
-            phys_dev_ext_props.descriptor_buffer_props.descriptorBufferAddressSpaceSize) {
+        } else if (pCreateInfo->size + descriptorBufferAddressSpaceSize >
```
spencer-lunarg, line 214: there was no reason to print 2 errors for same VUID here

arno-lunarg (approved):
LGTM, but is this just a draft then?

tests/negative/descriptor_buffer.cpp
```diff
@@ -1499,3 +1499,51 @@ TEST_F(VkLayerTest, DescriptorBufferInvalidExtensionCombination) {
...
+    ASSERT_NO_FATAL_FAILURE(InitState(nullptr, &descriptor_buffer_features));
+
+    auto descriptor_buffer_properties = LvlInitStruct<VkPhysicalDeviceDescriptorBufferPropertiesEXT>();
+    GetPhysicalDeviceProperties2(descriptor_buffer_properties);
+    // After a few GB, can have memory issues running these tests
+    // descriptorBufferAddressSpaceSize is always the largest of the 3 buffer address size limits
+    const VkDeviceSize max_limit = 1 << 31;
+    if (descriptor_buffer_properties.descriptorBufferAddressSpaceSize > max_limit) {
+        GTEST_SKIP() << "descriptorBufferAddressSpaceSize are too large";
+    }
```
juan-lunarg, line 1527: Should we run this test on Jenkins CI? Is the MockICD enough?

juan-lunarg (commented):
I ran `VkPositiveLayerTest.DescriptorBufferSetCreateBuffer` without your changes to repro the false positive. But the test passes fine.

Looks like this doesn't address the original false positive? Does the user have a false positive?

tests/negative/descriptor_buffer.cpp
spencer-lunarg: MockICD and local GPU both worked

spencer-lunarg:
This doesn't "fix" the issue, but the issue is because `descriptorBufferAddressSpaceSize` is returning `0` which not sure, but we have tests now regardless to prove it "should" work

dustContributor:
I closed my issue #5732 It wasn't a false positive, it was PEBCAK. I wasn't passing the VK_EXT_descriptor_buffer extension to the device creation. Thank you for your time.

Issue #4167, opened 2022-06-01
vkFreeCommandBuffers reports threading error against VkQueue


wolfpld:
In my program it is possible that the following conditions will occur. One of the threads is executing a `vkQueueSubmit` call, which properly protects access to the `VkQueue` in use. The other thread at the same time is executing `vkFreeCommandBuffers` call, which is required to protect access to command buffers and their pool, but nothing else.

In such conditions, the validation layers report an issue that is either invalid or not covered by the documentation:
`Validation Error: [ UNASSIGNED-Threading-MultipleThreads ] Object 0: handle = 0x1fe1d658a10, type = VK_OBJECT_TYPE_QUEUE; | MessageID = 0x141cb623 | THREADING ERROR : vkQueueSubmit(): object of type VkQueue is simultaneously used in thread 0xe140 and thread 0x16670`

This happens with Vulkan SDK 1.3.211.

ncesario-lunarg:
FYI @jeremyg-lunarg @jzulauf-lunarg.

I believe this error is in reference to https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/chap3.html#fundamentals-threadingbehavior. Specifically the fact that the `queue` object used when calling `vkQueueSubmit` can only be used by one thread at a time.

The error you're getting indicates that the `queue` object used in your `vkQueueSubmit` is not **externally synchronized**, and the `vkFreeCommandBuffers` could be a red herring. i.e., `queue` is in use at the same time by another API call in a different thread (possibly the same API, but in a different thread). If you are certain this is not possible, would you be able to provide an example app we could run to reproduce the behavior you're seeing?

wolfpld:
Only the two threads I have specified are touching the Vulkan API at the moment of the validation error. The validation error only occurs in the specified conditions, i.e. there are no other validation errors reported.

To be more specific, the call stack of the thread triggering the error is:

```
VulkanDebugCallback(unsigned int flags, VkDebugReportObjectTypeEXT objType, unsigned __int64 obj, unsigned __int64 location, int code, const char * layerPrefix, const char * msg, void * userData) Line 28	C++
VkLayer_khronos_validation.dll!debug_log_msg(const _debug_report_data * debug_data, unsigned int msg_flags, const LogObjectList & objects, const char * layer_prefix, const char * message, const char * text_vuid) Line 485	C++
[...]

PR #5738, opened 2023-04-25
layers: Report api names for both race participants

artem-lunarg:
- Adds additional parameter (one more api name) to the thread-safety error message.
- Limited clean up: two identical places for write reporting are moved to `HandleErrorOnWrite`. Three identical error messages are replaced with `GetErrorMessage` helper. These helpers can be further parameterized in case of future changes (or removed).

<sup> This PR tries to preserve original control flow structure and minimize changes to functionality (there are still repetitive control blocks, question about updates based on logging return value, etc). The plan is to have a separate PR to take into account history since 2015 and update some parts with potentially junk DNA.</sub>

Currently, when the race happens, VVL reports api name from the current thread but does not say which call was on the other side of the race. Call stack might be misleading: another thread is not blocked and can progress a bit further during short period of time that current thread takes to reach the breakpoint (can be reproduced).

Updated error message:
`THREADING ERROR : vkQueueSubmit(): object of type VkQueue is simultaneously used in thread 17440 (vkQueueWaitIdle) and current reporting thread 22392`

The idea to report more information about the race context is from this issue: https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4167

### DIFF
--- a/layers/generated/thread_safety.h
+++ b/layers/generated/thread_safety.h
@@ -93,7 +93,25 @@ public:
         }
     }
 
+    void Store(std::thread::id thread, const char* api_name) {
+        this->api_name.store(api_name, std::memory_order_relaxed);
+
+        // release ordering to store 'thread' variable acts like a barrier
+        // for previous stores, so they can be relaxed.
+        this->thread.store(thread, std::memory_order_release);
+    }
+    std::thread::id LoadThreadId() const {
+        // acquire ordering to load 'thread' variable acts like a barrier
+        // for subsequent loads, so they can be relaxed.
+        return this->thread.load(std::memory_order_acquire);
+    }
+    // This function is called in scenarios when 'thread' variable is already load-acquired.
+    void LoadRelaxed(std::thread::id& thread, const char*& api_name) const {
+        thread = this->thread.load(std::memory_order_relaxed);
+        api_name = this->api_name.load(std::memory_order_relaxed);
+    }
     std::atomic<std::thread::id> thread{};
+    std::atomic<const char*> api_name{nullptr};
 
 private:
     // Need to update write and read counts atomically. Writer in high 32 bits, reader in low 32 bits.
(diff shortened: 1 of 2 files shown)


ncesario-lunarg (commented):
I think this looks good, but should get a sign-off from @jzulauf-lunarg first.

layers/generated/thread_safety.h
juan-lunarg, line 1: The `ObjectUseData` and `counter` class should be moved to a seperate file.

Having it be generated doesn't make a lot of sense.

juan-lunarg (commented):
Lots of the generated code in this file shouldn't be generated.

layers/generated/thread_safety.h
artem-lunarg: > Lots of the generated code in this file shouldn't be generated.
Exactly my thoughts. Modifications is hard.
Going to create separate issue for this task, and include in the next PR.

juan-lunarg: Also FWIW perhaps if they are moved to another file you could write explicit unit tests for those classes? Just a thought. Not sure if it makes a lot of sense. This is just based off of my narrow reading of the code.

layers/generated/thread_safety.h
```diff
@@ -137,58 +155,31 @@ class counter {
...
         if (!use_data) {
             return;
         }
+        const std::thread::id tid = std::this_thread::get_id();
         const ObjectUseData::WriteReadCount prevCount = use_data->AddWriter();
 
         if (prevCount.GetReadCount() == 0 && prevCount.GetWriteCount() == 0) {
             // There is no current use of the object.  Record writer thread.
-            use_data->thread = tid;
+            use_data->Store(tid, api_name);
```
jzulauf-lunarg, line 168: I don't see how this isn't racy.  Another thread could come in and dive down the else path somewhere in the middle of the Store.  It's racy on tid before, but now we could be halfway in the middle of a multiword update... seem like a bunch of the logic needs to be mutex'd -- with some meaningful return code to report invalid threading...

artem-lunarg: the spinlock `use_data->WaitForObjectIdle(true)` ensures other updates won't proceed.

This code was originally mutexed but later updated to atomics, could be due to performance reasons. But in general I like idea to investigate in the future option with mutex and how much it affects performance.

jzulauf-lunarg: Not sure what prevents the Store on 168 from racing with the LoadThreadId on 174 or 181.

Also the WaitForObjectIdle in HandleErrorOnWrite only protects one path.

artem-lunarg: That's a good question! That's not a problem for this validation.

Lets distinguish between `data race`,  when it's defined as data corruption due to write-write or write-read without atomic guarantees. It's obviously not happening here because operations are atomic.

What is happening here is `race condition` when ordering is not guaranteed and non-deterministic (so I can read either new tid or old tid). It's not a problem for this algorithm. Thread safe validation does not guarantee it catches problematic code, only if we are lucky and `BeginWrite/EndWrite` sections in two threads temporarily overlap. It's easy miss that overlap by thousands-millions instructions depending on scheduling in which case we don't have validation error. Attempt to enforce some ordering of how thread id atomics are updated does not improve detection rate. The lenth of `BeginWrite/EndWrite` code sequence is much larger comparing single (or few) instructrions around tid load, so even if ordering is enforced with mutex it's might not bring even theoretical improvement of detection rate because we can be with equal probability before and after tid load at the moment when tid update locks the mutex. I hope this explanation makes at least some sense but we can discuss further. I was checking this question during implementation but it's always good to discuss more deeply.

artem-lunarg: > Also the WaitForObjectIdle in HandleErrorOnWrite only protects one path.

Good observation as always. That's correct. That else part will be removed and Store will be done unconditionally and with spin lock protection. I decided not to touch general control flow in this PR and only introduce logging of new variable, but looking through the history of modifications it looks like some parts had more sense in the old versions.

artem-lunarg:
I will start with this task in the reversed direction, I would be more reliable.

1. Make part of the code non-generated, as suggested by Juan.
2. Clean up and fixes of existing code.
3. Finally functionality from this PR as suggested here or in a different form, if we come up with a solution that is simpler to understand and validate.

PR #5739, opened 2023-04-26
tests: Remove check for zero maxPushDescriptors

spencer-lunarg:
These checks for a zero `maxPushDescriptors` seems like an artifact of the old DevSim/MockICD that didn't fill in all the properties, I assume this will run find on CI now

### DIFF
--- a/tests/negative/descriptors.cpp
+++ b/tests/negative/descriptors.cpp
@@ -1950,21 +1950,6 @@ TEST_F(VkLayerTest, DSBufferInfoErrors) {
     if (!AreRequiredExtensionsEnabled()) {
         GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported.";
     }
-    const bool update_template_support =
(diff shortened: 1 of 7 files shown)


tests/negative/descriptors.cpp
```diff
@@ -1950,21 +1950,6 @@ TEST_F(VkLayerTest, DSBufferInfoErrors) {
     if (!AreRequiredExtensionsEnabled()) {
         GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported.";
     }
-    const bool update_template_support =
-        IsExtensionsEnabled(VK_KHR_DESCRIPTOR_UPDATE_TEMPLATE_EXTENSION_NAME) && !IsDriver(VK_DRIVER_ID_AMD_PROPRIETARY);
```
spencer-lunarg, line 1954: these are marked as `AddRequiredExtensions`

If this is still failing on the AMD driver, will add back or we can add to blacklist

artem-lunarg (approved):

PR #5741, opened 2023-04-26
layers: Split vk_safe_struct.cpp

arno-lunarg:
Split `vk_safe_struct.cpp`:
- `vk_safe_struct_core.cpp`: 21k lines
- `vk_safe_stuct_khr.cpp`: 11k lines
- `vk_safe_struct_ext.cpp`: 18k lines
- `vk_safe_struct_vendor.cpp`: 13k lines
- `vk_safe_struct_utils.cpp`: 3k lines


scripts/helper_file_generator.py
```diff
@@ -1261,11 +1248,49 @@ def qfi_construct(item, member):
...
+        extra_code = ''
+
+        # Find what types of safe structs need to be generated based on output file name
+        safe_struct_type_re = r'.*';
+        if self.genOpts.filename.endswith('_core.cpp'):
+            safe_struct_type_re = r'.*[a-z0-9]$'
+            extra_code = ''
+        elif self.genOpts.filename.endswith('_khr.cpp'):
+            safe_struct_type_re = r'.*(KHR|KHX)$'
+            extra_code = """
```
spencer-lunarg, line 1262: can you add a comment why this is needed

scripts/helper_file_generator.py
```diff
@@ -1261,11 +1248,49 @@ def qfi_construct(item, member):
...
+};
+vl_concurrent_unordered_map<const safe_VkAccelerationStructureGeometryKHR*, ASGeomKHRExtraData*, 4> as_geom_khr_host_alloc;
+
+"""
+        elif self.genOpts.filename.endswith('_ext.cpp'):
+            safe_struct_type_re = r'.*EXT$'
+            extra_code = ''
+        elif self.genOpts.filename.endswith('_vendor.cpp'):
+            safe_struct_type_re = r'^(?!.*(KHR|KHX|EXT)$).*[A-Z]$' # Matches all words finishing with an upper case letters, but not ending with KHR, KHX or EXT
+
```
spencer-lunarg, line 1285: there is no `extra_code = ''` here, but I think this is right, and `extra_code = ''` should removed above in other `if` cases

scripts/helper_file_generator.py
```diff
@@ -1261,11 +1248,49 @@ def qfi_construct(item, member):
...
+        # source code prepended to output
+        extra_code = ''
+
+        # Find what types of safe structs need to be generated based on output file name
+        safe_struct_type_re = r'.*';
+        if self.genOpts.filename.endswith('_core.cpp'):
+            safe_struct_type_re = r'.*[a-z0-9]$'
+            extra_code = ''
+        elif self.genOpts.filename.endswith('_khr.cpp'):
+            safe_struct_type_re = r'.*(KHR|KHX)$'
```
spencer-lunarg, line 1261: there should be no `KHX`

scripts/helper_file_generator.py
```diff
@@ -1261,11 +1248,49 @@ def qfi_construct(item, member):
                           'ANativeWindow',
                           'CAMetalLayer'
                          ]
+
+        # source code prepended to output
+        extra_code = ''
+
+        # Find what types of safe structs need to be generated based on output file name
+        safe_struct_type_re = r'.*';
+        if self.genOpts.filename.endswith('_core.cpp'):
```
spencer-lunarg, line 1257: This should just be the `else` on the bottom I would assume

scripts/helper_file_generator.py
arno-lunarg: Well that comment would probably say "there is no clean way to add custom code in this file, so this as been dumped here" ^^"

scripts/helper_file_generator.py
arno-lunarg: Ok, I added this because of this in `vk.xml`:
```xml
        <tag name="KHR"         author="Khronos"                       contact="Tom Olson @tomolson"/>
        <tag name="KHX"         author="Khronos"                       contact="Tom Olson @tomolson"/>
```

scripts/helper_file_generator.py
spencer-lunarg: sorry not 

"why is there `extra_code`"

but

"why is there a `ASGeomKHRExtraData`"

I don't understand why only this things has an Extra data struct

scripts/helper_file_generator.py
spencer-lunarg: fair, but `KHX` is long dead, it was an attempt to make things "more important than EXT" but without the ratification that `KHR` provides

there is no `KHX` anyway in the repo, except for this line in the XML

scripts/helper_file_generator.py
```diff
@@ -1261,11 +1248,48 @@ def qfi_construct(item, member):
...
+    uint32_t primitiveOffset;
+    uint32_t primitiveCount;
+};
+vl_concurrent_unordered_map<const safe_VkAccelerationStructureGeometryKHR*, ASGeomKHRExtraData*, 4> as_geom_khr_host_alloc;
+
+"""
+        elif self.genOpts.filename.endswith('_ext.cpp'):
+            safe_struct_type_re = r'.*EXT$'
+        elif self.genOpts.filename.endswith('_vendor.cpp'):
+            safe_struct_type_re = r'^(?!.*(KHR|EXT)$).*[A-Z]$' # Matches all words finishing with an upper case letters, but not ending with KHR, KHX or EXT
```
artem-lunarg, line 1281: KHX: regex and comment mismatch

artem-lunarg (approved):
LGTM! One note about comment

scripts/helper_file_generator.py
artem-lunarg, line 1281: nit: KHX in the comment (not sure if another comment was about this usage of KHX, so just in case)

scripts/helper_file_generator.py
artem-lunarg: Initially this code was in the `GenerateSafeStructHelperSource`, is it possible to keep it there with additional check for for KHR case?

scripts/helper_file_generator.py
```diff
@@ -1261,11 +1248,48 @@ def qfi_construct(item, member):
...
+        # elif self.genOpts.filename.endswith('_core.cpp'):
```
spencer-lunarg, line 1282: nit: I would do

`else: # self.genOpts.filename.endswith('_core.cpp'):`

this looks like some random code was commented out

spencer-lunarg (approved):
LGTM with nit fixes

scripts/helper_file_generator.py
arno-lunarg: The programmer who added `safe_VkAccelerationStructureGeometryKHR` wanted to have some custom logic to manage a `as_geom_khr_host_alloc` map, mapping a `safe_VkAccelerationStructureGeometryKHR` to the extra data it needs to allocate to do a proper copy of the original struct. The logic is something unique to `VkAccelerationStructureGeometryKHR`, and the code generation of `vk_safe_struct.cpp` has no good way to account for that, so he originally dumped it in `GenerateSafeStructHelperSource`.

There is already custom code for `safe_VkAccelerationStructureGeometryKHR` in `GenerateSafeStructSource`, so I am going to move this `ASGeomKHRExtraData` logic just above the `safe_VkAccelerationStructureGeometryKHR` methods, to group things together. With this change I do not need `extra_code` anymore. 
Overall it will still look bad I think, but it will be closer to the existing logic.

ncesario-lunarg (approved):
LGTM!

PR #5742, opened 2023-04-26
tests: Get rid of VkVerticesObj

spencer-lunarg:
This has not be used and seems of little value to keep (it didn't really save any lines of code and made it more complex to know what it was doing)

Also the code was not in the style the rest of the tests are

There are much better ways to improve the pipeline helper logic IMO


arno-lunarg (approved):
LGTM! Makes tests more readable IMO
image

arno-lunarg:
(why is the image so big!)

artem-lunarg:
> (why is the image so big!)
computers!

spencer-lunarg:
> the content of the vertex buffer is not important for affected tests

Nope, it just needs garbage to send, 99% of our tests don't care about the values we pass in, such a luxury IMO

PR #5743, opened 2023-04-26
tests: Give better names for test groups

spencer-lunarg:
This is the first step to give better naming for tests, this is easy since it can just match the file name

These tests didn't have any exception with CI, so changing them should be no issue and a good "do we like this convention" test


arno-lunarg (approved):
LGTM! Better names yeah (o゜▽゜)o☆

ncesario-lunarg (approved):
Nice!

Issue #5733, opened 2023-04-25
vkCmdSetFragmentShadingRateKHR pre-call validation incorrect


tycho:
**Environment:**
 - OS: Windows 11, build 25346
 - GPU: NVIDIA RTX 4090
 - SDK or header version if building from repo: 1.3.243.0
 - Options enabled (synchronization, best practices, etc.): "Standard" preset in vkconfig

**Describe the Issue**

The [specification](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_fragment_shading_rate.html) for `VK_KHR_fragment_shading_rate` says it depends on:

>>    [VK_KHR_create_renderpass2](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_create_renderpass2.html)
>>    or
>>    [Version 1.2](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_fragment_shading_rate.html#versions-1.2)
>
>and
>
>>    [VK_KHR_get_physical_device_properties2](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_get_physical_device_properties2.html)
>>    or
>>    [Version 1.1](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_fragment_shading_rate.html#versions-1.1)

So basically: `(VK_KHR_create_renderpass2 || VK_VERSION_1_2) && (VK_KHR_get_physical_device_properties2 || VK_VERSION_1_1)`

But the code for the validation does this:

```c++
bool StatelessValidation::PreCallValidateCmdSetFragmentShadingRateKHR(
    VkCommandBuffer                             commandBuffer,
    const VkExtent2D*                           pFragmentSize,
    const VkFragmentShadingRateCombinerOpKHR    combinerOps[2]) const {
    bool skip = false;
    if (!instance_extensions.vk_feature_version_1_1) skip |= OutputExtensionError("vkCmdSetFragmentShadingRateKHR", "VK_VERSION_1_1");
    if (!IsExtEnabled(device_extensions.vk_khr_get_physical_device_properties2)) skip |= OutputExtensionError("vkCmdSetFragmentShadingRateKHR", "VK_KHR_get_physical_device_properties2");
[...]

PR #5745, opened 2023-04-26
layers: Add XML workaround for param validation

ncesario-lunarg:
Parameter validation was missing the workaround required for recent
upstream vulkan XML changes.

Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5733.

### DIFF
--- a/scripts/parameter_validation_generator.py
+++ b/scripts/parameter_validation_generator.py
@@ -728,7 +728,9 @@ class ParameterValidationOutputGenerator(OutputGenerator):
         if requires is not None:
             base_req_exts.append(f'({requires})')
         if len(base_req_exts) > 0:
-            base_required_extension_expressions.append(parseExpr('+'.join(base_req_exts)))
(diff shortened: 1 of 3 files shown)


scripts/parameter_validation_generator.py
```diff
@@ -728,7 +728,9 @@ def beginFeature(self, interface, emit):
         if requires is not None:
             base_req_exts.append(f'({requires})')
         if len(base_req_exts) > 0:
-            base_required_extension_expressions.append(parseExpr('+'.join(base_req_exts)))
+            # This is a work around for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5372
+            expr_str = re.sub(r',VK_VERSION_1_\d+', '', '+'.join(base_req_exts))
+            base_required_extension_expressions.append(parseExpr(expr_str))
```
ncesario-lunarg, line 733: Looks like this workaround was added to the helper file generator, but not parameter validation.

spencer-lunarg (approved):

PR #5746, opened 2023-04-26
Android discoveries / cleanup

juan-lunarg:


tests/android/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+- https://github.com/android/ndk-samples/blob/main/camera/basic/src/main/cpp/CMakeLists.txt <- In this example the symbol exported is `ANativeActivity_onCreate`
+- https://github.com/android/ndk-samples/blob/main/hello-vulkan/app/src/main/cpp/CMakeLists.txt <- In this example the symbol exported is `Java_com_google_androidgamesdk_GameActivity_initializeNativeCode`
+
+However you can run executables on Android:
+https://github.com/android/ndk/discussions/1726
+
+From one of the NDK maintainers: "code executed with "adb shell" runs in a different, more permissive environment than an Android app."
+
+Which seems perfect for us. Since we aren't really shipping an Android app. We just want to test on Android.
```
juan-lunarg, line 15: @spencer-lunarg based on this link it seems running an executable via adb shell should be fine.

But you mentioned something about WSI functionality. Which I couldn't find any documentation about.

spencer-lunarg (approved):

tests/android/README.md
spencer-lunarg: so yes, for WSI, we need to call `vkCreateAndroidSurfaceKHR` which takes a `ANativeWindow` handle

Android has a life-cycle system for apps (start, background, etc), With the `ANativeActivity_onCreate` we are getting hooks into it

I wrote this a LONG time ago, but this should help illustrate things
https://github.com/sjfricke/Vulkan-NDK-Template/blob/master/app/src/main/cpp/AndroidMain.cpp

1. We need to get the mapping from the Java events so the C++ code can see it

```
    // Set the callback to process system events
    app->onAppCmd = handle_cmd;
```

2.  From here we can use `NativeAppGlueAppCmd` (https://developer.android.com/reference/games/game-activity/group/android-native-app-glue) (or just look in `android_native_app_glue.c`)

3. For `APP_CMD_INIT_WINDOW` we can go `android_app->window` to get the `ANativeWindow` handle

-----

Android has many apps running, but only the ones in the foreground on the device get access to the `ANativeWindow` as that is what decides what is being displayed by SurfaceFlinger (in the AOSP)

When doing a command line executable, we have no proper way to get that handle, and therefor can't create a `VkSurface`

------

More info: There is actually 2 `ANativeWindow`, one in the NDK, and the "real" one in the AOSP (Android itself)

I know it has been done with debug AOSP builds, you can just build the AOSP `ANativeWindow` libs and link it with the program and it it bypass the NDK to get direct access to the `ANativeWindow` as a command line tool.... I highly recommend not trying this as having a debug AOSP build on CI machines is a rabit hole in itself IMO

PR #5748, opened 2023-04-27
tests: Use new Negative test name

spencer-lunarg:
This adds more tests with the `Negative*` schema

I didn't touch the GPU-AV or Best Practices here

While doing this, I remove some redundant keywords such as

- `Tests`
- `Invalid`
- `Validate`
- `Validation`
- `Violation`
- `Bad`
- `Verify`
- `Error`
- `Illegal`

### DIFF
--- a/tests/framework/layer_validation_tests.h
+++ b/tests/framework/layer_validation_tests.h
@@ -384,7 +384,7 @@ class VkGpuAssistedLayerTest : public VkLayerTest {
     bool CanEnableGpuAV();
 };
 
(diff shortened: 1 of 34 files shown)


arno-lunarg (approved):
LGTM! Regarding test names cleanup, it would be worth mentioning it somewhere so that everyone is one the same page regarding new test names?

spencer-lunarg:
> it would be worth mentioning it somewhere so that everyone is one the same page regarding new test names?

Yes, I plan to, don't worry

ncesario-lunarg (approved):

PR #5478, opened 2023-03-21
layers: Mask fragment shader state with rasterizerDiscard

HansKristian-Work:
It is possible to encounter VUID 06590 as a false positive if rasterizer discard is used. Fragment output state was masked out (incorrectly, since it did not consider dynamic state), and fragment shader state was not masked at all.

Ran into this in vkd3d-proton test suite.

PR #5751, opened 2023-04-27
layers: Default to enabled rasterization

ncesario-lunarg:
Assume rasterization is enabled for non-pre-raster graphics libraries
and linked libraries that do not contain pre-raster state.

### DIFF
--- a/layers/state_tracker/pipeline_state.h
+++ b/layers/state_tracker/pipeline_state.h
@@ -484,21 +484,37 @@ class PIPELINE_STATE : public BASE_NODE {
         return false;
     }
 
-    template <typename CreateInfo>
-    static bool EnablesRasterizationStates(const CreateInfo &create_info) {
-        if (create_info.pDynamicState && create_info.pDynamicState->pDynamicStates) {
-            for (uint32_t i = 0; i < create_info.pDynamicState->dynamicStateCount; ++i) {
-                if (create_info.pDynamicState->pDynamicStates[i] == VK_DYNAMIC_STATE_RASTERIZER_DISCARD_ENABLE_EXT) {
-                    return create_info.pRasterizationState != nullptr;
+    template <typename ValidationObject, typename CreateInfo>
+    static bool EnablesRasterizationStates(const ValidationObject &vo, const CreateInfo &create_info) {
+        // If this is an executable pipeline created from linking graphics libraries, we need to find the pre-raster library to
+        // check if rasterization is enabled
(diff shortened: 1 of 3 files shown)


layers/state_tracker/pipeline_state.h
```diff
@@ -517,6 +533,23 @@ class PIPELINE_STATE : public BASE_NODE {
                                                                           const safe_VkGraphicsPipelineCreateInfo &safe_create_info,
                                                                           const std::shared_ptr<const RENDER_PASS_STATE> &rp);
 
+    template <typename CreateInfo>
+    static bool EnablesRasterizationStates(const CreateInfo &create_info) {
+        if (create_info.pDynamicState && create_info.pDynamicState->pDynamicStates) {
+            for (uint32_t i = 0; i < create_info.pDynamicState->dynamicStateCount; ++i) {
+                if (create_info.pDynamicState->pDynamicStates[i] == VK_DYNAMIC_STATE_RASTERIZER_DISCARD_ENABLE_EXT) {
+                    return create_info.pRasterizationState != nullptr;
```
artem-lunarg, line 541: Can we have a comment here, similar to the above ones about what we assume? I believe there was explanation in one of the previous reviews.

artem-lunarg (approved):
LGTM! Opportunity in one place to add a comment.

Tony-LunarG (approved):
This fixes the errors I was seeing running CTS with validation

layers/state_tracker/pipeline_state.h
ncesario-lunarg: This was copied over from #5478. Looking at this again, I think this should return true. `pRasterizationState` is only valid if `create_info` is a complete set of state or is a graphics library containing pre-raster state.

@HansKristian-Workif this sounds accurate, I'll add a comment to this effect.

HansKristian-Work: I think the idea behind that check was for pipelines where you have dynamic state that is just ignored. IIRC, there is nothing preventing fragment output to contain RASTERIZER_DISCARD_ENABLE_EXT, it will be ignored.

Maybe the check has to consider if the pipeline is supposed to contain pRasterizationState at all?

ncesario-lunarg: > Maybe the check has to consider if the pipeline is supposed to contain pRasterizationState at all?

Yeah, I think this is what should be done, and then I will return true (i.e., rasterization is _enabled_) as a default.

Thanks for the quick response!

PR #5752, opened 2023-04-27
DebugReporter cleanup

juan-lunarg:
Move debug reporter code into error_monitor.h/error_monitor.cpp respectively.


spencer-lunarg (approved):
agree, not sure why we had this in `render`

ncesario-lunarg (approved):

arno-lunarg (approved):

PR #5754, opened 2023-04-27
Get tests compiling on Android with CMake

juan-lunarg:


build-android/jni/Android.mk
```diff
@@ -243,9 +243,6 @@ LOCAL_CPPFLAGS += -DVK_ENABLE_BETA_EXTENSIONS -DVK_USE_PLATFORM_ANDROID_KHR -DUS
 LOCAL_WHOLE_STATIC_LIBRARIES += android_native_app_glue
 LOCAL_LDLIBS := -llog -landroid -ldl
 
-# Building tests as a shared library (.so) with the intent of being tested as an APK instead of an executable.
-LOCAL_CPPFLAGS += -DVVL_TESTS_APK_GLUE
-
```
juan-lunarg, line 248: It may be valuable in the future to test the standalone executable despite lack of WSI support. But given limited resources, I'd say it's not a priority. So tests can only be built with the intention of being used in a APK.

ncesario-lunarg (approved):

PR #5756, opened 2023-04-28
build: Add arch to deps dir

ncesario-lunarg:
Add the architecture type to the external dependency path when using UPDATE_DEPS=ON.


scripts/CMakeLists.txt
```diff
@@ -49,7 +49,11 @@ if (UPDATE_DEPS)
 
     set(UPDATE_DEPS_DIR_SUFFIX "${_build_type}")
     if (ANDROID)
-        set(UPDATE_DEPS_DIR_SUFFIX "android/${UPDATE_DEPS_DIR_SUFFIX}")
+        set(UPDATE_DEPS_DIR_SUFFIX "android/${UPDATE_DEPS_DIR_SUFFIX}/${CMAKE_ANDROID_ARCH_ABI}")
+    elseif(CMAKE_SIZEOF_VOID_P EQUAL 8)
+        set(UPDATE_DEPS_DIR_SUFFIX "${UPDATE_DEPS_DIR_SUFFIX}/x64")
+    elseif(CMAKE_SIZEOF_VOID_P EQUAL 4)
+        set(UPDATE_DEPS_DIR_SUFFIX "${UPDATE_DEPS_DIR_SUFFIX}/x86")
```
juan-lunarg, line 56: x64/x86 is misleading. You can build on arm for Windows.
```suggestion
    elseif(CMAKE_SIZEOF_VOID_P EQUAL 8)
        set(UPDATE_DEPS_DIR_SUFFIX "${UPDATE_DEPS_DIR_SUFFIX}/64")
    elseif(CMAKE_SIZEOF_VOID_P EQUAL 4)
        set(UPDATE_DEPS_DIR_SUFFIX "${UPDATE_DEPS_DIR_SUFFIX}/32")
```

juan-lunarg (changes requested):
This will break `vvl.yml` caching.

juan-lunarg:
Closing old PR due to lack of activity

scripts/CMakeLists.txt
ncesario-lunarg: > This will break `vvl.yml` caching.

@juan-lunarg Any ideas how to get this working with vvl.yml caching? Seems like we need to declare a map or something to that effect to map the various cmake generators to 32/64, but I'm not sure if that's possible in github yml. I'd like to get this merged before Android cmake is working, but it would also be nice to have for Windows 32/64 bit builds.

ncesario-lunarg:
"Reopen PR" didn't seem to work as intended. This is now moved #5875.

Issue #3476, opened 2021-10-26
missing 64bit vertex attribute validation


zmike:
The Vulkan specification states:

`The input variable in the shader must be declared as a 64-bit data type if and only if format is a 64-bit data type.`

This is not a VU (that I can see), but it is a restrictive statement that has no accompanying validation, e.g., binding an attribute of format `VK_FORMAT_R64_SFLOAT` and then using a vertex shader that loads a `vec4` from it is illegal but does not trigger errors.

I think there should be validation for this, but probably that needs a new VU? I'm happy to file spec bugs to get a new VU if needed, but I thought I'd post here first to see if there were any other opinions.

spencer-lunarg:
I have https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5869 now, also have a local PR ready with tests, just waiting to confirm that all 4 ways to mix-and-match 64-bit vertex input are invalid

spencer-lunarg:
@zmike this was a long issue finally closed, the VUID are still being added, but if you hit issues with the 64-bit vertex input let us know here

PR #5757, opened 2023-04-28
Add 64-bit vertex input validation

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/3476

I have a spec change to add proper VUIDs, so most likely by next spec update we will have them and I will come back and label these

A few Quality-of-life improvement while adding this


tests/positive/vertex_input.cpp
```diff
@@ -0,0 +1,289 @@
...
+    pipe.vi_ci_.vertexBindingDescriptionCount = 1;
+    pipe.vi_ci_.pVertexAttributeDescriptions = input_attribs;
+    pipe.vi_ci_.vertexAttributeDescriptionCount = 2;
+    pipe.shader_stages_ = {vs.GetStageCreateInfo(), fs.GetStageCreateInfo()};
+    pipe.InitState();
+    pipe.CreateGraphicsPipeline();
+    /* expect success */
+}
+
+TEST_F(VkPositiveLayerTest, CreatePipelineAttribArrayType) {
```
arno-lunarg, line 57: Missing description?

arno-lunarg (approved):
LGTM! Maybe one missing test description

layers/generated/vk_format_utils.h
```diff
@@ -247,6 +247,29 @@ FORMAT_COMPATIBILITY_CLASS FormatCompatibilityClass(VkFormat format);
 bool FormatElementIsTexel(VkFormat format);
 uint32_t FormatElementSize(VkFormat format, VkImageAspectFlags aspectMask = VK_IMAGE_ASPECT_COLOR_BIT);
 double FormatTexelSize(VkFormat format, VkImageAspectFlags aspectMask = VK_IMAGE_ASPECT_COLOR_BIT);
+// True if Format contains a 64-bit component
+constexpr bool FormatIs64bit(VkFormat format) {
```
juan-lunarg, line 251: While I realize this function is very straightforward please add testing. It's easy for `codegen` mistakes to happen in the future.

layers/state_tracker/shader_instruction.h
```diff
@@ -93,6 +93,6 @@ class Instruction {
 
     // Max capacity needs to be uint32_t because an instruction can have a string operand that is (2^16)-1 bytes long
     small_vector<uint32_t, word_vector_length, uint32_t> words_;
-    uint32_t result_id_;
-    uint32_t type_id_;
+    uint32_t result_id_index_;
+    uint32_t type_id_index_;
```
juan-lunarg, line 97:
```suggestion
    uint32_t result_id_index_ = 0;
    uint32_t type_id_index_ = 0;
```

layers/state_tracker/shader_instruction.h
```diff
@@ -50,10 +50,10 @@ class Instruction {
 
     uint32_t Opcode() const { return words_[0] & 0x0ffffu; }
 
-    // operand id index, return 0 if no result
-    uint32_t ResultId() const { return result_id_; }
-    // operand id index, return 0 if no type
-    uint32_t TypeId() const { return type_id_; }
+    // operand id, return 0 if no result
+    uint32_t ResultId() const { return (result_id_index_ == 0) ? 0 : words_[result_id_index_]; }
```
juan-lunarg, line 54: Would it make sense to have `result_id_index_` be a std::optional?

Just looks like 0 means that this value wasn't filled out. Which looks like the pattern for where std::optional should be used.

layers/state_tracker/shader_instruction.cpp
```diff
@@ -17,7 +17,7 @@
 #include "state_tracker/shader_module.h"
 #include "generated/spirv_grammar_helper.h"
 
-Instruction::Instruction(std::vector<uint32_t>::const_iterator it) : result_id_(0), type_id_(0) {
+Instruction::Instruction(std::vector<uint32_t>::const_iterator it) : result_id_index_(0), type_id_index_(0) {
```
juan-lunarg, line 20: This can be handled in the header file instead.

```suggestion
Instruction::Instruction(std::vector<uint32_t>::const_iterator it) {
```

spencer-lunarg: I guess we have been doing these init at construction time everywhere else in the code, wasn't sure "when" it should be done in the header vs here

layers/state_tracker/shader_instruction.h
spencer-lunarg: Most common use case of `ResultId()` is seeing if it matches some other ID, so returning `0` is a perfectly valid as no other ID can be `0`

layers/state_tracker/shader_instruction.cpp
juan-lunarg: Less chance of bugs if someone makes another constructor when doing it in the header file in this instance.

juan-lunarg (approved):

layers/state_tracker/shader_instruction.h
ncesario-lunarg: I think the SPIR-V spec says somewhere that a valid ID is an integer greater than zero, but I can't find the .

scripts/format_utils_generator.py
```diff
@@ -696,7 +703,17 @@ def sizeFunctions(self):
...
+             # Could loop the components, but faster to just list these
+            output += '// True if Format contains a 64-bit component\n'
+            output += 'constexpr bool FormatIs64bit(VkFormat format) {\n'
+            output += '    bool found = false;\n'
+            output += '    switch (format) {\n'
+            for f in sorted(self.formats64bit):
+                output += '        case {}:\n'.format(f)
+            output += self.commonBoolSwitch
 
+            output += '''
```
ncesario-lunarg, line 716:
```suggestion
            output += '''// True if Format contains a 64-bit component\n'
constexpr bool FormatIs64bit(VkFormat format) {'
    bool found = false;'
    switch (format) {'
'''
            for f in sorted(self.formats64bit):
                output += f'        case {f}:\n'
            output += self.commonBoolSwitch

            output += '''
```

ncesario-lunarg (approved):
Some non-essential nits, but LGTM!

layers/core_checks/cc_shader.cpp
```diff
@@ -32,20 +32,6 @@
 #include "generated/spirv_grammar_helper.h"
 #include "external/xxhash.h"
 
-static uint32_t GetLocationsConsumedByFormat(VkFormat format) {
```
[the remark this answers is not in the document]
spencer-lunarg: So the "logic" was not the problem, it was that it was hand-maintained, in case in the future someone added something like `R128G128` or `R64G32B32A32` format, that would also need to return `2` but doubt we would know to update it here

PR #5762, opened 2023-04-29
tests: Breakup negative shader tests

spencer-lunarg:
create dedicated shader files for tests that are focused on testing the shader (not the pipeline itself)

This also provides a nice `NegativeShader*` test filter

Tests with `VkLayerTest` need to be updated in the blacklist in sync


tests/negative/shader_interface.cpp
```diff
@@ -0,0 +1,822 @@
...
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+#include "../framework/layer_validation_tests.h"
+
+class NegativeShaderInterface : public VkLayerTest {};
+
+TEST_F(VkLayerTest, CreatePipelineExceedVertexMaxComponentsWithBuiltins) {
```
ncesario-lunarg, line 19: Should these be using the new namespace `NegativeShaderInterace`?

ncesario-lunarg:
> Tests with `VkLayerTest` need to be updated in the blacklist in sync

Is this why the namespaces have been "created" but the tests aren't using them?

tests/negative/shader_interface.cpp
spencer-lunarg: these are the tests that need to be updated in the Blacklist

So after this, it will be even easier to spot which ones they are, and we can do the whole lockstep Blackist->VVL->Blacklist change

> Is this why the namespaces have been "created" but the tests aren't using them?

yes

ziga-lunarg (approved):

juan-lunarg (approved):

PR #5763, opened 2023-04-29
tests: Give most Positive tests new naming

spencer-lunarg:
I decide to go and add the new `Positive*` test name/class where possible

Tests with `VkPositiveLayerTest` still just need to be updated in the blacklist


tests/positive/sync_val.cpp
```diff
@@ -13,9 +13,9 @@
 
 #include "../framework/layer_validation_tests.h"
 
-class VkPositiveSyncValTest : public VkSyncValTest {};
+class kPositiveSyncVal : public VkSyncValTest {};
```
ncesario-lunarg, line 16:
```suggestion
class PositiveSyncVal : public VkSyncValTest {};
```
I'm guessing that was the intention?

ncesario-lunarg (dismissed):
I think one "search/replace" went awry. Otherwise LGTM!

tests/positive/sync_val.cpp
spencer-lunarg: ........yes

arno-lunarg (approved):

juan-lunarg (approved):

PR #5767, opened 2023-05-01
Add Mesh/Task shader workgroup memory limits

spencer-lunarg:
Adds `VUID-RuntimeSpirv-maxMeshSharedMemorySize-08754` and `VUID-RuntimeSpirv-maxTaskSharedMemorySize-08759` which check the Workgroup Shader memory size against Mesh and Task shaders

### DIFF
--- a/layers/core_checks/cc_shader.cpp
+++ b/layers/core_checks/cc_shader.cpp
@@ -1875,21 +1875,49 @@ bool CoreChecks::ValidateTransformFeedbackDecorations(const SHADER_MODULE_STATE 
     return skip;
 }
 
-bool CoreChecks::ValidateComputeSharedMemory(const SHADER_MODULE_STATE &module_state, uint32_t total_shared_size) const {
+bool CoreChecks::ValidateWorkgroupSharedMemory(const SHADER_MODULE_STATE &module_state, VkShaderStageFlagBits stage,
+                                               uint32_t total_workgroup_shared_memory) const {
     bool skip = false;
 
     // If not found before with spec constants, find here
-    if (total_shared_size == 0) {
-        total_shared_size = module_state.CalculateComputeSharedMemory();
+    if (total_workgroup_shared_memory == 0) {
+        total_workgroup_shared_memory = module_state.CalculateWorkgroupSharedMemory();
     }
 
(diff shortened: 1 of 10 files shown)


artem-lunarg (approved):
LGTM!

tests/device_profiles/max_profile.json
```diff
@@ -1179,65 +1179,65 @@
                     "shadingRateMaxCoarseSamples": 32
                 },
                 "VkPhysicalDeviceMeshShaderPropertiesNV": {
```
juan-lunarg, line 1181: What about `VkPhysicalDeviceMeshShaderPropertiesEXT`? Does this just alias it? If it does please use the EXT version instead of the NV version.

layers/core_checks/cc_shader.cpp
```diff
@@ -1875,21 +1875,48 @@ bool CoreChecks::ValidateTransformFeedbackDecorations(const SHADER_MODULE_STATE
...
+            if (total_workgroup_shared_memory > phys_dev_ext_props.mesh_shader_props_ext.maxTaskSharedMemorySize) {
+                skip |= LogError(module_state.vk_shader_module(), "VUID-RuntimeSpirv-maxTaskSharedMemorySize-08759",
+                                 "Shader uses %" PRIu32
+                                 " bytes of shared memory, more than allowed by maxTaskSharedMemorySize (%" PRIu32 ")",
+                                 total_workgroup_shared_memory, phys_dev_ext_props.mesh_shader_props_ext.maxTaskSharedMemorySize);
+            }
+            break;
+        }
+        default:
+            break;
```
juan-lunarg, line 1917: `assert` on `default`.

```suggestion
        default:
            assert(false);
            break;
```

tests/device_profiles/max_profile.json
spencer-lunarg: I edited `VkPhysicalDeviceMeshShaderPropertiesEXT` below

These are not alias, the NV and EXT extension are a good amount of different for Mesh Shaders

juan-lunarg (approved):

ncesario-lunarg (approved):
LGTM!

PR #5768, opened 2023-05-01
layers: Fix naming and type selection of external handles

artem-lunarg:
Use correct type (Flags vs FlagBits).
Use correct singular or plural form for handle type variables.

### DIFF
--- a/layers/state_tracker/device_memory_state.cpp
+++ b/layers/state_tracker/device_memory_state.cpp
@@ -19,12 +19,16 @@
 #include "state_tracker/device_memory_state.h"
 #include "state_tracker/image_state.h"
 
-static VkExternalMemoryHandleTypeFlags GetExportHandleType(const VkMemoryAllocateInfo *p_alloc_info) {
+// It is allowed to export memory into the handles of different types,
+// that's why we use set of flags (VkExternalMemoryHandleTypeFlags)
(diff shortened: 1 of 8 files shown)


layers/state_tracker/device_memory_state.cpp
```diff
@@ -19,12 +19,16 @@
 #include "state_tracker/device_memory_state.h"
 #include "state_tracker/image_state.h"
 
-static VkExternalMemoryHandleTypeFlags GetExportHandleType(const VkMemoryAllocateInfo *p_alloc_info) {
+// It is allowed to export memory into the handles of different types,
+// that's why we use set of flags (VkExternalMemoryHandleTypeFlags)
+static VkExternalMemoryHandleTypeFlags GetExportHandleTypes(const VkMemoryAllocateInfo *p_alloc_info) {
```
artem-lunarg, line 24: These two functions are the main motivation for this PR. The other changes are mostly mechanical rename.

spencer-lunarg (approved):
:100: like these long-needed cleanups!

artem-lunarg:
Probably far from complete clean up, I did not try to find everything, only the parts that directly interface with my current VUID task.

arno-lunarg (approved):

artem-lunarg:
Removed the build from internal CI for the last change, which is a comment update. Github CI should be enough for this.

Issue #5672, opened 2023-04-18
Cleanup: Use `IsImageLayoutStencilOnly` and `IsImageLayoutDepthOnly` where it makes sense


arno-lunarg:
Needs https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5664 to be merged

PR #5770, opened 2023-05-01
cleanup: Use IsImageLayout{Depth, Stencil}Only

arno-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5672

Use IsImageLayoutDepthOnly and IsImageLayoutStencilOnly where it makes sense. Also add some random missing comas.


juan-lunarg (approved):

juan-lunarg (commented):
It would be a bit overkill but adding some tests for these 2 helper functions in `tests/positive/layer_utils.cpp` would be appreciated.

spencer-lunarg (approved):

ncesario-lunarg (approved):
LGTM!

PR #5773, opened 2023-05-02
layers: Prepare ValidateBindBufferMemory for new changes

artem-lunarg:
I want to keep since change separately from the following changes,
since it's impossible to diff.

Clean up is about reordering validation sections and  indentation
changes. It's better to keep vkAllocateInfo-related checks before
vkBindBufferMemory checks to have chronological sequence of
events.

And the reason why we have vkAllocateInfo checks here in the first
place is because they depend on buffer/image properties.

### DIFF
--- a/layers/core_checks/cc_device_memory.cpp
+++ b/layers/core_checks/cc_device_memory.cpp
@@ -480,215 +480,209 @@ bool CoreChecks::ValidateMemoryTypes(const DEVICE_MEMORY_STATE *mem_info, const 
 
 bool CoreChecks::ValidateBindBufferMemory(VkBuffer buffer, VkDeviceMemory mem, VkDeviceSize memoryOffset, const void *pNext,
                                           const char *api_name) const {
+    bool skip = false;
+
+    // Validate device group information
+    if (const auto *bind_buffer_memory_device_group_info = LvlFindInChain<VkBindBufferMemoryDeviceGroupInfo>(pNext)) {
+        if (bind_buffer_memory_device_group_info->deviceIndexCount != 0 &&
+            bind_buffer_memory_device_group_info->deviceIndexCount != device_group_create_info.physicalDeviceCount &&
+            device_group_create_info.physicalDeviceCount > 0) {
+            const LogObjectList objlist(buffer, mem);
+            skip |= LogError(objlist, "VUID-VkBindBufferMemoryDeviceGroupInfo-deviceIndexCount-01606",
+                             "%s: The number of physical devices in the logical device is %" PRIu32
+                             ", but VkBindBufferMemoryDeviceGroupInfo::deviceIndexCount is %" PRIu32 ".",
+                             api_name, device_group_create_info.physicalDeviceCount,
+                             bind_buffer_memory_device_group_info->deviceIndexCount);
+        } else {
+            for (uint32_t i = 0; i < bind_buffer_memory_device_group_info->deviceIndexCount; ++i) {
+                if (bind_buffer_memory_device_group_info->pDeviceIndices[i] >= device_group_create_info.physicalDeviceCount) {
+                    const LogObjectList objlist(buffer, mem);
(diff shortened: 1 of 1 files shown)


spencer-lunarg (approved):

layers/core_checks/cc_device_memory.cpp
```diff
@@ -480,215 +480,209 @@ bool CoreChecks::ValidateMemoryTypes(const DEVICE_MEMORY_STATE *mem_info, const
...
+                             "%s: The VkDeviceMemory (%s) was created with an import operation with handleType of %s which "
+                             "is not set in the VkBuffer (%s) VkExternalMemoryBufferCreateInfo::handleType (%s)",
+                             api_name, report_data->FormatHandle(mem).c_str(),
+                             string_VkExternalMemoryHandleTypeFlagBits(mem_info->import_handle_type.value()),
+                             report_data->FormatHandle(buffer).c_str(),
+                             string_VkExternalMemoryHandleTypeFlags(buffer_state->external_memory_handle_types).c_str());
+        }
+
+        // Validate mix of protected buffer and memory
+        if ((buffer_state->unprotected == false) && (mem_info->unprotected == true)) {
```
juan-lunarg, line 672: It should just be `if (buffer_state->unprotected != mem_info->unprotected)`.
```suggestion
            // Validate mix of protected buffer and memory
            if (mem_info->unprotected != buffer_state->unprotected) {
                const char *vuid =
                    bind_buffer_mem_2 ? "VUID-VkBindBufferMemoryInfo-None-01898" : "VUID-vkBindBufferMemory-None-01898";

                const char *mem_msg = mem_info->unprotected ? "unprotected" : "protected";
                const char *buf_msg = buffer_state->unprotected ? "unprotected" : "protected";

                const LogObjectList objlist(buffer, mem);
                skip |= LogError(objlist, vuid,
                                 "%s: The VkDeviceMemory (%s) was created with %s memory but the VkBuffer (%s) was set "
                                 "to use %s memory.",
                                 api_name, report_data->FormatHandle(mem).c_str(), mem_msg,
                                 report_data->FormatHandle(buffer).c_str(), buf_msg);
            }
```

juan-lunarg (approved):
Minor nitpick

layers/core_checks/cc_device_memory.cpp
artem-lunarg: It does not take into account different VUIDs. Here's my attempt to add vuids:
```
        if (mem_info->unprotected != buffer_state->unprotected) {
            const char *vuid = nullptr;
            if (buffer_state->unprotected) {
                vuid = bind_buffer_mem_2 ? "VUID-VkBindBufferMemoryInfo-None-01899" : "VUID-vkBindBufferMemory-None-01899";
            } else {
                vuid = bind_buffer_mem_2 ? "VUID-VkBindBufferMemoryInfo-None-01898" : "VUID-vkBindBufferMemory-None-01898";
            }
            auto mem_msg = mem_info->unprotected ? "unprotected" : "protected";
            auto buf_msg = buffer_state->unprotected ? "unprotected" : "protected";
            const LogObjectList objlist(buffer, mem);
            skip |= LogError(objlist, vuid,
                             "%s: The VkDeviceMemory (%s) was created with %s memory but the VkBuffer (%s) was set "
                             "to use %s memory.",
                             api_name, report_data->FormatHandle(mem).c_str(), mem_msg, report_data->FormatHandle(buffer).c_str(),
                             buf_msg);
        }
```
My personal preference is that having separate paths for different vuids is easier to read. Single code path maybe shorter (still, not much with vuid selection), but I have to start thinking how the final message will look like, instead of just reading it. I agree that for single VUID that proposed method would be better.

juan-lunarg: > It does not take into account that we have 2 pairs of VUIDs.

:facepalm:

PR #5775, opened 2023-05-02
scripts: Add dedicated dynamic state helper script

spencer-lunarg:
This does 3 things

1. Creates a separate Dynamic State helper script, will be useful in future when we want to add more to it
2. Fix the name of `CB_DYNAMIC_*_SET` to `CB_DYNAMIC_STATE_*` otherwise it is **REALLY** annoying to find anything as you forget to remove `STATE`, also the `SET` sufix is redundant, the only reason we have these is to create an bitset with them
3. Rename `Status` to `State`, because the "status" part is using the bitset, not the individual enum values


ncesario-lunarg (approved):
LGTM!

PR #5778, opened 2023-05-02
gpu: Split descriptor set status into separate buffers

jeremyg-lunarg:
Split per-DescriptorSet state into separate memory blocks which are accessed via an array of buffer device addresses. This is being done to make it easier to update state for a single DescriptorSet without rebuilding the old giant flat buffer.

NOTE: This change requires SPIRV-Tools to be updated

### DIFF
--- a/layers/gpu_validation/gpu_utils.cpp
+++ b/layers/gpu_validation/gpu_utils.cpp
@@ -192,12 +192,16 @@ static VKAPI_ATTR void VKAPI_CALL gpuVkCmdCopyBuffer(VkCommandBuffer commandBuff
     DispatchCmdCopyBuffer(commandBuffer, srcBuffer, dstBuffer, regionCount, pRegions);
 }
 
-VkResult UtilInitializeVma(VkInstance instance, VkPhysicalDevice physical_device, VkDevice device, VmaAllocator *pAllocator) {
+VkResult UtilInitializeVma(VkInstance instance, VkPhysicalDevice physical_device, VkDevice device, bool use_buffer_device_address, VmaAllocator *pAllocator) {
     VmaVulkanFunctions functions;
     VmaAllocatorCreateInfo allocator_info = {};
     allocator_info.instance = instance;
     allocator_info.device = device;
     allocator_info.physicalDevice = physical_device;
+
+    if (use_buffer_device_address) {
+        allocator_info.flags |= VMA_ALLOCATOR_CREATE_BUFFER_DEVICE_ADDRESS_BIT;
+    }
 
     functions.vkGetInstanceProcAddr = static_cast<PFN_vkGetInstanceProcAddr>(gpuVkGetInstanceProcAddr);
     functions.vkGetDeviceProcAddr = static_cast<PFN_vkGetDeviceProcAddr>(gpuVkGetDeviceProcAddr);
@@ -254,13 +258,18 @@ void GpuAssistedBase::PreCallRecordCreateDevice(VkPhysicalDevice gpu, const VkDe
     // See CreateDevice() below.
     VkPhysicalDeviceFeatures gpu_supported_features;
     DispatchGetPhysicalDeviceFeatures(gpu, &gpu_supported_features);
-    auto modified_create_info = static_cast<VkDeviceCreateInfo *>(modified_ci);
+
+    // See CreateDevice() in chassis.cpp. modified_ci is a pointer to a safe struct stored on the stack.
+    // This code follows the safe struct memory memory management scheme. That is, we must delete any memory
+    // remove from the safe struct, and any additions must be allocated in a way that is compatible with
+    // the safe struct destructor.
+    auto *modified_create_info = static_cast<safe_VkDeviceCreateInfo *>(modified_ci);
     if (modified_create_info->pEnabledFeatures) {
         // If pEnabledFeatures, VkPhysicalDeviceFeatures2 in pNext chain is not allowed
         features = const_cast<VkPhysicalDeviceFeatures *>(modified_create_info->pEnabledFeatures);
     } else {
-        VkPhysicalDeviceFeatures2 *features2 = nullptr;
-        features2 = const_cast<VkPhysicalDeviceFeatures2 *>(LvlFindInChain<VkPhysicalDeviceFeatures2>(modified_create_info->pNext));
+        auto *features2 =
+            const_cast<VkPhysicalDeviceFeatures2 *>(LvlFindInChain<VkPhysicalDeviceFeatures2>(modified_create_info->pNext));
         if (features2) features = &features2->features;
     }
     VkPhysicalDeviceFeatures new_features = {};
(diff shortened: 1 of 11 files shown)


layers/gpu_validation/gpu_utils.cpp
```diff
@@ -284,6 +288,57 @@ void GpuAssistedBase::PreCallRecordCreateDevice(VkPhysicalDevice gpu, const VkDe
         delete modified_create_info->pEnabledFeatures;
         modified_create_info->pEnabledFeatures = new VkPhysicalDeviceFeatures(new_features);
     }
+    if (force_buffer_device_address) {
+        // TODO How to handle multi-device
```
ncesario-lunarg, line 292: Is there (or _should_ we) have an issue tracking this?

layers/gpu_validation/gpu_utils.cpp
```diff
@@ -284,6 +288,57 @@ void GpuAssistedBase::PreCallRecordCreateDevice(VkPhysicalDevice gpu, const VkDe
...
+                VkPhysicalDeviceBufferDeviceAddressFeatures *bda_features =
+                    const_cast<VkPhysicalDeviceBufferDeviceAddressFeatures *>(
+                        LvlFindInChain<VkPhysicalDeviceBufferDeviceAddressFeatures>(modified_create_info));
+                if (bda_features) {
+                    bda_features->bufferDeviceAddress = VK_TRUE;
+                } else {
+                    auto new_bda_features = LvlInitStruct<VkPhysicalDeviceBufferDeviceAddressFeatures>();
+                    new_bda_features.bufferDeviceAddress = VK_TRUE;
+                    new_bda_features.pNext = const_cast<void *>(modified_create_info->pNext);
+                    modified_create_info->pNext = new VkPhysicalDeviceBufferDeviceAddressFeatures(new_bda_features);
```
ncesario-lunarg, line 308: Is this leaked? I guess in the real world case, it doesn't matter much, but might not be great for testing. Maybe the `GpuAssistedBase` instance could own the memory?

layers/gpu_validation/gpu_utils.cpp
```diff
@@ -284,6 +288,57 @@ void GpuAssistedBase::PreCallRecordCreateDevice(VkPhysicalDevice gpu, const VkDe
...
+                }
+            }
+        } else if (api_version == VK_API_VERSION_1_1) {
+            bool found_ext = false;
+            for (uint32_t i = 0; i < modified_create_info->enabledExtensionCount; i++) {
+                if (strcmp(modified_create_info->ppEnabledExtensionNames[i], "VK_KHR_buffer_device_address") == 0) {
```
ncesario-lunarg, line 314: `strncmp` or `std::string{...} == modified_create_info...`?

layers/gpu_validation/gpu_utils.cpp
```diff
@@ -284,6 +288,57 @@ void GpuAssistedBase::PreCallRecordCreateDevice(VkPhysicalDevice gpu, const VkDe
...
+                    found_ext = true;
+                    break;
+                }
+            }
+            if (!found_ext) {
+                char ** ext_names = new char*[modified_create_info->enabledExtensionCount + 1];
```
ncesario-lunarg, line 320: Similar question as with `VkPhysicalDeviceBufferDeviceAddressFeatures` above.

layers/gpu_validation/gpu_utils.cpp
```diff
@@ -284,6 +288,57 @@ void GpuAssistedBase::PreCallRecordCreateDevice(VkPhysicalDevice gpu, const VkDe
...
+                memcpy(ext_names, modified_create_info->ppEnabledExtensionNames, sizeof(char*)*modified_create_info->enabledExtensionCount);
```
ncesario-lunarg, line 321: Something that takes the type into account (e.g., `std::copy`) might be clearer (i.e., than having the size calculation).

layers/gpu_validation/gpu_validation.cpp
```diff
@@ -1323,23 +1333,37 @@ void GpuAssisted::UpdateBoundDescriptors(VkCommandBuffer commandBuffer, VkPipeli
...
-        // code is going to dereference Input[ i + Input[ b + Input[ s + Input[ Input[0] ] ] ] ] to
-        // see if descriptors have been written. In gpu_validation.md, we note this and advise
-        // using densely packed bindings as a best practice when using gpu-av with descriptor indexing
-        bool has_buffers = false;
-        uint32_t descriptor_count = 0;  // Number of descriptors, including all array elements
-        uint32_t binding_count = 0;     // Number of bindings based on the max binding number used
-
-        // Figure out how much memory we need for the input block based on how many sets and bindings there are
-        // and how big each of the bindings is
+    if (number_of_sets > 0 && (descriptor_indexing || buffer_oob_enabled) && force_buffer_device_address) {
```
ncesario-lunarg, line 1336: Does this mean that BDA will be _required_ for GPUAV after this change?

layers/gpu_validation/gpu_validation.cpp
```diff
@@ -1360,150 +1384,81 @@ void GpuAssisted::UpdateBoundDescriptors(VkCommandBuffer commandBuffer, VkPipeli
...
-                        bindings_to_written += last + 1;
-                        bind_counter += last + 1;
-                        sizes += last + 1;
-                    } else {
-                        *sets_to_sizes++ = 0;
-                        *sets_to_bindings++ = 0;
+
+            // For each set, allocate an input buffer that describes the descriptor set and its update status as follows
+            // Word 0 = the number of bindings in the descriptor set - note that the bindings can be sparse and this is the largest
+            // binding number + 1 which we'll refer to as N Words 1 through Word N  = the number of descriptors in each binding
```
ncesario-lunarg, line 1390: Having some difficulty parsing this. Maybe some additional punctuation?
```suggestion
            // binding number + 1 which we'll refer to as N. Words 1 through Word N = the number of descriptors in each binding.
```

ncesario-lunarg (approved):
Some nits here and there, but LGTM!

layers/gpu_validation/gpu_utils.cpp
```diff
@@ -284,6 +288,57 @@ void GpuAssistedBase::PreCallRecordCreateDevice(VkPhysicalDevice gpu, const VkDe
         delete modified_create_info->pEnabledFeatures;
         modified_create_info->pEnabledFeatures = new VkPhysicalDeviceFeatures(new_features);
     }
+    if (force_buffer_device_address) {
+        // TODO How to handle multi-device
+        if (api_version > VK_API_VERSION_1_1) {
+            VkPhysicalDeviceVulkan12Features *features12 = const_cast<VkPhysicalDeviceVulkan12Features *>(
+                LvlFindInChain<VkPhysicalDeviceVulkan12Features>(modified_create_info->pNext));
+            if (features12) {
+                features12->bufferDeviceAddress = VK_TRUE;
```
spencer-lunarg, line 297: in the doc, there is a section `GpuPreCallRecordCreateDevice` that should be updated with this new logic

layers/gpu_validation/gpu_utils.cpp
spencer-lunarg: is this something we could find in `ValidationStateTracker::PreCallRecordCreateDevice` suprised we don't have a `bool` we can set somewhere to find this, but I guess its at init time and not the worst

layers/gpu_validation/gpu_utils.cpp
```diff
@@ -284,6 +288,57 @@ void GpuAssistedBase::PreCallRecordCreateDevice(VkPhysicalDevice gpu, const VkDe
...
+            } else {
+                VkPhysicalDeviceBufferDeviceAddressFeatures *bda_features =
+                    const_cast<VkPhysicalDeviceBufferDeviceAddressFeatures *>(
+                        LvlFindInChain<VkPhysicalDeviceBufferDeviceAddressFeatures>(modified_create_info));
```
artem-lunarg, line 301: nit: `modified_create_info->pNext`
`modified_create_info`  itself cannot be `VkPhysicalDeviceBufferDeviceAddressFeatures`.

layers/gpu_validation/gpu_validation.cpp
```diff
@@ -1323,23 +1333,37 @@ void GpuAssisted::UpdateBoundDescriptors(VkCommandBuffer commandBuffer, VkPipeli
...
+            vmaCreateBuffer(vmaAllocator, &buffer_info, &alloc_info, &di_buffers.address_buffer, &di_buffers.address_buffer_allocation, nullptr);
+        if (result != VK_SUCCESS) {
+            ReportSetupProblem(device, "Unable to allocate device memory.  Device could become unstable.", true);
+            aborted = true;
+            return;
+        }
+        // Allocate buffer for device addresses of the input buffer for each descriptor set.  This is the buffer written to each
+        // draw's descriptor set.
+        uint64_t *address_data_ptr;
+        result = vmaMapMemory(vmaAllocator, di_buffers.address_buffer_allocation, reinterpret_cast<void **>(&address_data_ptr));
```
artem-lunarg, line 1355: if the result is not checked, then the suggestion is to initialize `address_data_ptr` with null, to have some early alarm in the case of fire.

layers/gpu_validation/gpu_utils.cpp
jeremyg-lunarg: ext_names is leaked, found with ASan in github CI, so this part is going to change

layers/gpu_validation/gpu_utils.cpp
jeremyg-lunarg: I don't think it is leaked but this part of the chassis is not very clear so I will double check & comment

layers/gpu_validation/gpu_validation.cpp
jeremyg-lunarg: yes

jeremyg-lunarg: I'm adding some text in the documentation for this new requirement.

layers/gpu_validation/gpu_utils.cpp
jeremyg-lunarg: modified_create_info is actually a safe_VkDeviceCreateInfo stored on the stack in the chassis code, so this isn't a leak. Adding a comment.

layers/gpu_validation/gpu_utils.cpp
jeremyg-lunarg: modified_create_info is actually a safe_VkDeviceCreateInfo stored on the stack in the chassis code, so this isn't a leak. Adding a comment.

layers/gpu_validation/gpu_utils.cpp
jeremyg-lunarg: std::string{} would malloc. string_view maybe...

jeremyg-lunarg: also, there is stuff in state tracker that should sets bools for the enabled extensions. But this code runs before that gets set up by ValidationStateTracker::CreateDevice()

jeremyg-lunarg:
> LGTM - Thanks for fixing everything up

There's still some sort of problem on AMD gpus that I have to fix.

layers/gpu_validation/gpu_utils.cpp
jeremyg-lunarg: resolved by storing the extension name literal in a static const std::string

layers/gpu_validation/gpu_utils.cpp
juan-lunarg: Do we even test multi device? I have not seen any issues related to multi-device.

Might be more pragmatic to just add an assert or fail if used with multi-device.

jeremyg-lunarg: @Tony-LunarG  what do you think about this proposal?

layers/gpu_validation/gpu_validation.cpp
```diff
@@ -1360,150 +1384,90 @@ void GpuAssisted::UpdateBoundDescriptors(VkCommandBuffer commandBuffer, VkPipeli
...
+                        SetBindingState(descriptor_data_ptr, written_index, binding.get());
                     }
+                    written_index += binding->count;
                 }
+                auto buffer_device_address_info = LvlInitStruct<VkBufferDeviceAddressInfo>();
+                buffer_device_address_info.buffer = di_input_buffer.buffer;
+
+                // We cannot rely on device_extensions here, since we may be enabling BDA support even
+                // though the application has not requested it.
+                if (api_version >= VK_API_VERSION_1_2) {
```
jeremyg-lunarg, line 1453: This was the problem with AMD & Mesa drivers. We need to call the correct function depending on what we did in UtilInitializeVma()

layers/gpu_validation/gpu_utils.cpp
Tony-LunarG: I couldn't think of a good way to handle it at the time.  I think an abort should be ok

layers/gpu_validation/gpu_validation.cpp
Tony-LunarG: That makes sense.  I suppose the Nvidia driver doesn't care.

PR #5779, opened 2023-05-03
tests: Remove DebugReporter

juan-lunarg:
Move all functionality into ErrorMonitor.


tests/framework/error_monitor.cpp
```diff
@@ -17,9 +17,100 @@
  */
 #include "error_monitor.h"
 
+#ifdef VK_USE_PLATFORM_ANDROID_KHR
+// Note. VK_EXT_debug_report is deprecated by the VK_EXT_debug_utils extension.
+// However, we still support this old extension due to CI running old Android devices.
```
juan-lunarg, line 22: I believe this is the only rational. Because I tried only using `VK_EXT_debug_utils` and Jenkins CI failed on some Android devices. If there is more rational for using `VK_EXT_debug_report` I didn't see it.

spencer-lunarg: I don't know details, but some of debug extensions are not supported in the Android Vulkan Loader so it can't be used

spencer-lunarg (approved):

juan-lunarg:
> CI Vulkan-ValidationLayers build # 11919 failed.

CI failed due to compilation warning on Windows testing. Which GHA didn't find because GHA doesn't build tests on Windows.

juan-lunarg:
> CI failed due to compilation warning on Windows testing. Which GHA didn't find because GHA doesn't build tests on Windows.

I'm currently working on getting Windows testing on GHA.

PR #5780, opened 2023-05-03
docs: Remove outdated GetDeviceProcAddr example usage

juan-lunarg:


docs/creating_tests.md
```diff
@@ -138,7 +132,7 @@ PFN_vkBindImageMemory2KHR vkBindImageMemory2Function = nullptr;
 if (DeviceValidationVersion() >= VK_API_VERSION_1_1) {
     vkBindImageMemory2Function = vk::BindImageMemory2;
 } else {
-    vkBindImageMemory2Function = reinterpret_cast<PFN_vkBindImageMemory2KHR>(vk::GetDeviceProcAddr(m_device->handle(), "vkBindImageMemory2KHR"));
+    vkBindImageMemory2Function = vk::BindImageMemory2KHR;
```
spencer-lunarg, line 135: we should just remove this all and rewrite this `Getting Function Pointers` section, idea:

----

When using a version that has promoted the function, one can just directly use the call.

In the case of enabling the extensions, all the functions for those extensions will call `vkGetDeviceProcAddr` automatically 

```c++
// If VK_API_VERSION_1_1 or later is set
vk::BindImageMemory2(...);

// If VK_KHR_bind_memory2 is enabled
vk::BindImageMemory2KHR(...);
```

artem-lunarg:
@juan-lunarg, why am I assigned to this PR? you are doing the work! :)

juan-lunarg:
> @juan-lunarg, why am I assigned to this PR? you are doing the work! :)

I usually use assignment as an indicator I want a specific reviewer to take a look. In this case I'm changing the docs because of a change you made so I figured you might have feedback.

docs/creating_tests.md
juan-lunarg: updated based on feedback

artem-lunarg: Just thoughts. Reading the paragraphs above I'm not  quite sure if we can just go with `vk::` or there can be some corner cases when we need to get function pointers manually. If we don't expect such situations, or it's a niche cases, then one option to simplify even more by saying (maybe with improved grammar) that "all core and extension function pointers are available based on requested Vulkan API version and enabled extensions".

arno-lunarg (approved):

artem-lunarg (approved):

PR #5788, opened 2023-05-04
Add InitBasicTransformFeedback

spencer-lunarg:
This creates a new `m_attempted_api_version` to track what `SetTargetApiVersion` was trying to use so we can go both

```c++
InitBasicTransformFeedback();

// and

InitBasicTransformFeedback();
```

### DIFF
--- a/tests/negative/transform_feedback.cpp
+++ b/tests/negative/transform_feedback.cpp
@@ -15,7 +15,29 @@
 #include "utils/cast_utils.h"
 #include "../framework/layer_validation_tests.h"
 
-class NegativeTransformFeedback : public VkLayerTest {};
+class NegativeTransformFeedback : public VkLayerTest {
+  public:
+    void InitBasicTransformFeedback();
+};
+
+void NegativeTransformFeedback::InitBasicTransformFeedback() {
+    AddRequiredExtensions(VK_EXT_TRANSFORM_FEEDBACK_EXTENSION_NAME);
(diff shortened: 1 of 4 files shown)


tests/negative/transform_feedback.cpp
```diff
@@ -70,25 +92,10 @@ TEST_F(NegativeTransformFeedback, FeatureEnabled) {
...
-    ASSERT_NO_FATAL_FAILURE(InitFramework(m_errorMonitor));
-
-    if (DeviceValidationVersion() < VK_API_VERSION_1_1) {
-        GTEST_SKIP() << "At least Vulkan version 1.1 is required.";
-    }
-    if (!AreRequiredExtensionsEnabled()) {
-        GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported.";
-    }
+    InitBasicTransformFeedback();
+    if (::testing::Test::IsSkipped()) return;
```
arno-lunarg, line 97: I did not know about that, this is cool!

arno-lunarg (approved):

tests/negative/transform_feedback.cpp
spencer-lunarg: So how GTest works is everything is a `void` function so when you call `return` or `GTEST_SKIP()` it will return out, so if we skip in a helper function we have to make sure to do the actual `return`

PR #5789, opened 2023-05-04
layers: Add an unimplementable_validation.h file

spencer-lunarg:
from https://gitlab.khronos.org/vulkan/vulkan/-/issues/3475

This is in effort to mark VUs that are just not possible with the scope of a Vulkan Layer

Each item added in future should have an explanation as well

### DIFF
--- a/layers/error_message/unimplementable_validation.h
+++ b/layers/error_message/unimplementable_validation.h
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2023 LunarG, Inc.
+ * Copyright (c) 2023 Valve Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+// This file list all VUID that are no possible to validate.
+// This file should never be included, but here for searchability and statistics
+
+const char* unimplementable_validation[] = {
+    // TODO - This might be able to be rewritten as how VUID-VkWaylandSurfaceCreateInfoKHR-display-01304 is
+    //
+    // Not possible to validate a "valid" handle for fd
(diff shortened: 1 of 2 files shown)


layers/error_message/unimplementable_validation.h
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright (c) 2023 LunarG, Inc.
```
artem-lunarg, line 2: The last time when I added copyright the recommendation was to add Valve too. Not sure if it's always needed though.

artem-lunarg (approved):
LGTM! copyright comment

layers/error_message/unimplementable_validation.h
```diff
@@ -0,0 +1,28 @@
...
+ */
+#pragma once
+
+// This file list all VUID that are no possible to validate.
+// This file should never be included, but here for searchability and statistics
+
+const char* unimplementable_validation[] = {
+    // Not possible to validate a "valid" handle for fd
+    // https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5431#issuecomment-1484167442
+    "VUID-vkGetMemoryFdPropertiesKHR-fd-00673",
```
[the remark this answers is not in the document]
spencer-lunarg: Agree, might be worth trying to rephrase it as

>  fd  must point to a valid POSIX file descriptor memory handle

Added a `TODO` and will bring this up for not WG meeting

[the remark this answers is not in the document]
spencer-lunarg: https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5895

layers/error_message/unimplementable_validation.h
```diff
@@ -0,0 +1,31 @@
...
+
+// This file list all VUID that are no possible to validate.
+// This file should never be included, but here for searchability and statistics
+
+const char* unimplementable_validation[] = {
+    // TODO - This might be able to be rewritten as how VUID-VkWaylandSurfaceCreateInfoKHR-display-01304 is
+    //
+    // Not possible to validate a "valid" handle for fd
+    // https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5431#issuecomment-1484167442
+    "VUID-vkGetMemoryFdPropertiesKHR-fd-00673",
```
ncesario-lunarg, line 27: Are these for the benefit of vk_validation_stats.py?

spencer-lunarg: yes, but also

- To prevent new attempts to validate such VUIDs
- Valuable for the Vulkan API users to know that Validation Layer does not assist with a specific VUID

Issue #5783, opened 2023-05-03
False positive with read-only layout validation with dynamic state that is ineffective


ShabbyX:
Take the following scenario:

- Render pass uses DEPTH_STENCIL_READ_ONLY_OPTIMAL
- It starts and sets `vkCmdSetDepthTestEnable(false)`. Depth test and write are dynamic state
- It does _not_ set `vkCmdSetDepthWriteEnable`
- Draw
- End render pass

In the above, VVL complains that:

> vkCmdDraw: depthWriteEnable is VK_TRUE, while the layout (VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL) of the depth aspect of the depth/stencil attachment in the render pass is read only. The Vulkan spec states: If the current render pass instance uses a depth/stencil attachment with a read-only layout for the depth aspect, depth writes must be disabled (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdDraw-None-06886)

It seems that VVL sees depth write enabled because of a previous render pass which set it enabled. Vulkan says:

> Dynamic state that does not affect the result of operations can be left undefined.

In the above, since depth-test is disabled, depth-write doesn't matter and so it doesn't have to be set (as I understand it).

ShabbyX:
This is very similar to https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4921 actually. That issue is closed, but the false positive still happens in ANGLE (last I tried after it was closed): https://chromium-review.googlesource.com/c/angle/angle/+/4103721

spencer-lunarg:
Agree, I can easily fix this case, but it annoying we don't have a good way to see which state effect other state as a dependency (`depthWriteEnable` depends on `depthTestEnable`)

ShabbyX:
> edit: I went through spec, built a table

In the xml I hope?

spencer-lunarg:
> In the xml I hope?

For now, just manually did it to get the job done, when I get done with the Sync XML, can look into these tables

PR #5790, opened 2023-05-04
DepthTestEnable overrides DepthWriteEnable

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5783

There is a lot more work to follow-up with the PR after, but for now, it adds a proper check to `LAST_BOUND_STATE` to check if `DepthWriteEnable` is really set or not

### DIFF
--- a/layers/state_tracker/pipeline_state.cpp
+++ b/layers/state_tracker/pipeline_state.cpp
@@ -860,3 +860,17 @@ void LAST_BOUND_STATE::Reset() {
     push_descriptor_set.reset();
     per_set.clear();
 }
+
+bool LAST_BOUND_STATE::IsDepthTestEnable() const {
+    return pipeline_state->IsDynamic(VK_DYNAMIC_STATE_DEPTH_TEST_ENABLE) ? cb_state.dynamic_state_value.depth_test_enable
+                                                                         : pipeline_state->DepthStencilState()->depthTestEnable;
+}
+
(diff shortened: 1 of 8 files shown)


layers/state_tracker/pipeline_state.cpp
```diff
@@ -707,6 +707,17 @@ PIPELINE_STATE::PIPELINE_STATE(const ValidationStateTracker *state_data, const V
                    VK_SHADER_STAGE_MISS_BIT_KHR | VK_SHADER_STAGE_INTERSECTION_BIT_KHR | VK_SHADER_STAGE_CALLABLE_BIT_KHR)));
 }
 
+// Return true if for a given PSO, the given state enum is dynamic, else return false
+bool PIPELINE_STATE::IsDynamic(const VkDynamicState state) const {
+    const auto *dynamic_state = DynamicState();
+    if ((pipeline_type == VK_PIPELINE_BIND_POINT_GRAPHICS) && dynamic_state) {
```
artem-lunarg, line 713: Do we need `pipeline_type == VK_PIPELINE_BIND_POINT_GRAPHICS` for correctness?

artem-lunarg (approved):
LGTM

layers/state_tracker/pipeline_state.cpp
spencer-lunarg: this was remove, https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5792 fixed this

Issue #5288, opened 2023-02-15
Add VkSubpassDescription VUs for VkImageLayout


spencer-lunarg:
The following were all added recently, they all are all for checking the `VkImageLayout` is correct

note there is a `VkSubpassDescription2` version for each of these

// Core Vulkan
- [ ] VUID-VkSubpassDescription-attachment-06912
- [ ] VUID-VkSubpassDescription-attachment-06913
- [ ] VUID-VkSubpassDescription-attachment-06914
- [ ] VUID-VkSubpassDescription-attachment-06915

// VK_KHR_maintenance2
- [ ] VUID-VkSubpassDescription-attachment-06916
- [ ] VUID-VkSubpassDescription-attachment-06917

// VK_KHR_separate_depth_stencil_layouts
- [ ] VUID-VkSubpassDescription-attachment-06918
- [ ] VUID-VkSubpassDescription-attachment-06919
- [ ] VUID-VkSubpassDescription-attachment-06920
- [ ] VUID-VkSubpassDescription2-attachment-06251 (only for `VkSubpassDescription2`)

// VK_KHR_synchronization2
- [ ] VUID-VkSubpassDescription-attachment-06921
- [ ] VUID-VkSubpassDescription-attachment-06922
- [ ] VUID-VkSubpassDescription-attachment-06923

PR #5791, opened 2023-05-04
Add subpass attachment layout VUIDS 

arno-lunarg:
- Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5288

Add VUIDs listed from this one:
https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkSubpassDescription2.html#VUID-VkSubpassDescription2-attachment-06912
to this one:
https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkSubpassDescription2.html#VUID-VkSubpassDescription2-attachment-06251

No new positive test, it does not make much sense here

### DIFF
--- a/layers/core_checks/cc_render_pass.cpp
+++ b/layers/core_checks/cc_render_pass.cpp
@@ -1434,6 +1434,28 @@ bool CoreChecks::ValidateRenderpassAttachmentUsage(RenderPassCreateVersion rp_ve
                                      function_name, j, i);
                 }
 
+                const VkImageLayout attachment_layout = attachment_ref.layout;
+                if (IsValueIn(attachment_layout,
+                              {VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL})) {
+                    const char *vuid =
+                        use_rp2 ? "VUID-VkSubpassDescription2-attachment-06912" : "VUID-VkSubpassDescription-attachment-06912";
+                    skip |= LogError(device, vuid, "%s: Input attachment %s layout (%s) is invalid.", function_name,
+                                     error_type.c_str(), string_VkImageLayout(attachment_layout));
+                }
+                if (IsValueIn(attachment_layout,
+                              {VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_STENCIL_ATTACHMENT_OPTIMAL})) {
+                    const char *vuid =
+                        use_rp2 ? "VUID-VkSubpassDescription2-attachment-06918" : "VUID-VkSubpassDescription-attachment-06918";
+                    skip |= LogError(device, vuid, "%s: Input attachment %s layout (%s) is invalid.", function_name,
+                                     error_type.c_str(), string_VkImageLayout(attachment_layout));
+                }
+                if (attachment_layout == VK_IMAGE_LAYOUT_ATTACHMENT_OPTIMAL_KHR) {
+                    const char *vuid =
+                        use_rp2 ? "VUID-VkSubpassDescription2-attachment-06921" : "VUID-VkSubpassDescription-attachment-06921";
+                    skip |= LogError(device, vuid, "%s: Input attachment %s layout (%s) is invalid.", function_name,
+                                     error_type.c_str(), string_VkImageLayout(attachment_layout));
+                }
+
                 // safe to dereference pCreateInfo->pAttachments[]
                 if (attachment_index < pCreateInfo->attachmentCount) {
                     const VkFormat attachment_format = pCreateInfo->pAttachments[attachment_index].format;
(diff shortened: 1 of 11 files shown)


layers/core_checks/cc_render_pass.cpp
```diff
@@ -1545,6 +1569,39 @@ bool CoreChecks::ValidateRenderpassAttachmentUsage(RenderPassCreateVersion rp_ve
...
 
+                    const VkImageLayout attachment_layout = attachment_ref.layout;
+                    if (IsValueIn(attachment_layout,
+                                  {VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL})) {
+                        const char *vuid =
+                            use_rp2 ? "VUID-VkSubpassDescription2-attachment-06914" : "VUID-VkSubpassDescription-attachment-06914";
+                        skip |= LogError(device, vuid, "%s: Input attachment %s layout (%s) is invalid.", function_name,
+                                         error_type.c_str(), string_VkImageLayout(attachment_layout));
+                    }
+                    if (IsExtEnabled(device_extensions.vk_khr_maintenance2) &&
```
spencer-lunarg, line 1580: `std::vector<VkImageLayout> ValidationObject::ValidParamValues()` will catch this if `IsExtEnabled(device_extensions.vk_khr_maintenance2)` is false

so you can remove these `IsExtEnabled` for these checks

tests/negative/renderpass.cpp
```diff
@@ -4263,3 +4262,381 @@ TEST_F(NegativeRenderPass, IncompatibleRenderPass2) {
...
+        ref.layout = VK_IMAGE_LAYOUT_STENCIL_ATTACHMENT_OPTIMAL;
+        TestRenderPassCreate(m_errorMonitor, *m_device, rpci, rp2_supported, "VUID-VkSubpassDescription-attachment-06920",
+                             "VUID-VkSubpassDescription2-attachment-06920");
+
+        ref.layout = VK_IMAGE_LAYOUT_STENCIL_READ_ONLY_OPTIMAL;
+        TestRenderPassCreate(m_errorMonitor, *m_device, rpci, rp2_supported, "VUID-VkSubpassDescription-attachment-06920",
+                             "VUID-VkSubpassDescription2-attachment-06920");
+    }
+
+    if (rp2_supported) {
```
spencer-lunarg, line 4608: that fact we have `auto rpci2 = LvlInitStruct<VkRenderPassCreateInfo2KHR>();` can we just make this a separate test... easier to separate in future if we care about moving `VK_KHR_separate_depth_stencil_layouts` tests out to own file

tests/positive/render_pass.cpp
```diff
@@ -1340,7 +1340,7 @@ TEST_F(PositiveRenderPass, FramebufferCreateDepthStencilLayoutTransitionForDepth
     m_commandBuffer->QueueCommandBuffer(false);
 }
 
-TEST_F(VkPositiveLayerTest, FramebufferWithAttachmentsTo3DImageMultipleSubpasses) {
+TEST_F(PositiveRenderPass, FramebufferWithAttachmentsTo3DImageMultipleSubpasses) {
```
spencer-lunarg, line 1343: sorry, this wasn't a mistake, this test was on the CI blacklist

I still needed to go through them... but if this passes CI, then horray! we can mark test as fixed

tests/negative/renderpass.cpp
```diff
@@ -4263,3 +4262,381 @@ TEST_F(NegativeRenderPass, IncompatibleRenderPass2) {
...
+    const bool rp2_supported = IsExtensionsEnabled(VK_KHR_CREATE_RENDERPASS_2_EXTENSION_NAME);
+
+    if (!AreRequiredExtensionsEnabled()) {
+        GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported";
+    }
+
+    ASSERT_NO_FATAL_FAILURE(InitState());
+
+    constexpr std::array<VkAttachmentDescription, 3> attachments = {{
+        {0, VK_FORMAT_R8G8B8A8_UNORM, VK_SAMPLE_COUNT_1_BIT, VK_ATTACHMENT_LOAD_OP_DONT_CARE, VK_ATTACHMENT_STORE_OP_STORE,
```
spencer-lunarg, line 4280: something for post-PR, but I have never liked these large `VkAttachmentDescription` we use everywhere as doesn't ever describe anything

Since this file is where the most of them are, we could create some `constexpr VkAttachmentDescription` in `NegativeRenderPass` that can be used so this can become something more readable such as

```c++
constexpr std::array<VkAttachmentDescription, 3> attachments = {{basicColorAttachment, multisampleColorAttachment, multisampleDepthStencilAttachment}}
```

tests/negative/renderpass.cpp
arno-lunarg: All the test cases here rely on `VK_KHR_separate_depth_stencil_layouts` being enabled, including before this `if (rp2_supported)`. It is not clear because tests cases before that just need the extension to be enabled to work!

tests/negative/renderpass.cpp
arno-lunarg: I agree your approach is much more readable!

artem-lunarg (approved):
LGTM. One suggestion about code compression if that makes sense.

layers/core_checks/cc_render_pass.cpp
```diff
@@ -1434,6 +1434,30 @@ bool CoreChecks::ValidateRenderpassAttachmentUsage(RenderPassCreateVersion rp_ve
                                      function_name, j, i);
                 }
 
+                const VkImageLayout attachment_layout = attachment_ref.layout;
```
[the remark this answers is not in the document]
arno-lunarg: I thought about that too, I deemed the validation code small enough to not deserve a lambda, even though the error message is duplicated. I feel that code is clearer without the lambda? But ask me tomorrow and I will say otherwise ^^

tests/negative/renderpass.cpp
spencer-lunarg: for now it is ok, agree other tests need to be split up as well

layers/core_checks/cc_render_pass.cpp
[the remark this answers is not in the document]
spencer-lunarg: I think it looks good here as in

layers/core_checks/cc_render_pass.cpp
```diff
@@ -1434,6 +1434,30 @@ bool CoreChecks::ValidateRenderpassAttachmentUsage(RenderPassCreateVersion rp_ve
...
+                if (IsValueIn(attachment_layout,
+                              {VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL})) {
+                    const char *vuid =
+                        use_rp2 ? "VUID-VkSubpassDescription2-attachment-06912" : "VUID-VkSubpassDescription-attachment-06912";
+                    skip |= LogError(device, vuid, "%s: Input attachment %s layout (%s) is invalid.", function_name,
+                                     error_type.c_str(), string_VkImageLayout(attachment_layout));
+                }
+                if (IsExtEnabled(device_extensions.vk_khr_separate_depth_stencil_layouts) &&
```
spencer-lunarg, line 1445: `IsExtEnabled` not needed (or below for `vk_khr_synchronization2`) for all of these checks in this PR, same idea with the other other `VK_KHR_maintenance2` comment I had

spencer-lunarg (approved):
LGTM - once the redundant `IsExtEnabled` are removed

PR #5794, opened 2023-05-04
tests: Allow fd posix handle testing on Windows

artem-lunarg:
Now we can run fd posix handle tests on Windows through MockICD.

Vulkan API for posix handles are defined in a cross platform way, so it's possible to write related code without platform #ifdefs. This change removes restriction in the test framework that vk_testing::Fence/Semaphore supports posix handle only in non-Windows environments.

If the test is built around `ExternalHandle` abstraction, then that abstraction if moved to corresponding test suite.

### DIFF
--- a/tests/negative/external_memory_sync.cpp
+++ b/tests/negative/external_memory_sync.cpp
@@ -17,7 +17,14 @@
 #include "../framework/layer_validation_tests.h"
 #include "utils/vk_layer_utils.h"
 
-class NegativeExternalMemorySync : public VkLayerTest {};
+class NegativeExternalMemorySync : public VkLayerTest {
+  protected:
+#ifdef VK_USE_PLATFORM_WIN32_KHR
+    using ExternalHandle = HANDLE;
+#else
+    using ExternalHandle = int;
+#endif
+};
 
 TEST_F(NegativeExternalMemorySync, CreateBufferIncompatibleHandleTypes) {
     TEST_DESCRIPTION("Creating buffer with incompatible external memory handle types");
@@ -518,7 +525,7 @@ TEST_F(NegativeExternalMemorySync, TimelineSemaphore) {
     stci.pNext = nullptr;
     vk_testing::Semaphore import_semaphore(*m_device, sci);
 
-    vk_testing::Semaphore::ExternalHandle ext_handle{};
+    ExternalHandle ext_handle{};
     err = export_semaphore.export_handle(ext_handle, handle_type);
     ASSERT_VK_SUCCESS(err);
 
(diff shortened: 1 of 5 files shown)


tests/negative/external_memory_sync.cpp
```diff
@@ -827,23 +834,23 @@ TEST_F(NegativeExternalMemorySync, SyncFdFence) {
     fci.pNext = nullptr;
     vk_testing::Fence import_fence(*m_device, fci);
 
-    vk_testing::Fence::ExternalHandle ext_handle{};
+    int fd_handle{};
```
artem-lunarg, line 837: `NegativeExternalMemorySync.SyncFdFence`: That's example of the test that previously crashed and now works on Windows with MockICD.

tests/positive/sync_object.cpp
```diff
@@ -1652,15 +1659,15 @@ TEST_F(PositiveSyncObject, ExternalFenceSubmitCmdBuffer) {
         err = vk::QueueSubmit(m_device->m_queue, 1, &submit_info, export_fence.handle());
         ASSERT_VK_SUCCESS(err);
 
-        vk_testing::Fence::ExternalHandle ext_handle{};
-        err = export_fence.export_handle(ext_handle, handle_type);
+        int fd_handle{};
```
jeremyg-lunarg, line 1662: fd_handle should be initialized to -1, since 0 is a valid file descriptor.

tests/negative/external_memory_sync.cpp
jeremyg-lunarg, line 837: fd_handle should be initialized to -1, since 0 is a valid file descriptor.

tests/negative/external_memory_sync.cpp
```diff
@@ -592,17 +599,17 @@ TEST_F(NegativeExternalMemorySync, SyncFdSemaphore) {
     // Create a semaphore to import payload into
     vk_testing::Semaphore import_semaphore(*m_device);
 
-    vk_testing::Semaphore::ExternalHandle ext_handle{};
+    int fd_handle{};
```
jeremyg-lunarg, line 602: fd_handle should be initialized to -1, since 0 is a valid file descriptor.

tests/positive/sync_object.cpp
```diff
@@ -1672,7 +1679,7 @@ TEST_F(PositiveSyncObject, ExternalFenceSubmitCmdBuffer) {
                 ASSERT_FALSE(ret == 0);                                         // Timeout.
                 ASSERT_TRUE(ret == -1 && (errno == EINTR || errno == EAGAIN));  // Retry...
             }
-            close(ext_handle);
+            close(fd_handle);
         }
 #else
         err = vk::QueueWaitIdle(m_device->m_queue);
```
jeremyg-lunarg, line 1685: Does anything need to be done with the fd_handle on Windows? close() it?

tests/positive/sync_object.cpp
artem-lunarg: 👍 I have to remember this.

tests/positive/sync_object.cpp
artem-lunarg: I guess not? when it's only for MockICD.
If there will be _real_ POSIX compatible implementation on Windows, then the entire #ifdef block should be made accessible.

jeremyg-lunarg: ok, would be good to have a comment about this at least.

artem-lunarg: Done.

juan-lunarg (approved):

PR #5799, opened 2023-05-04
layers: Only read relevant settings

christophe-lunarg:
Layers settings using environment variables on Android is slow. This is causing some C.I. performance issues.

To resolve this, we are now reading only settings which dependences are met.

Also, if legacy 'enables' and 'disables' settings are set, then we don't read the new settings.

### DIFF
--- a/layers/layer_options.cpp
+++ b/layers/layer_options.cpp
@@ -475,35 +475,53 @@ void ProcessConfigAndEnvSettings(ConfigAndEnvSettings *settings_data) {
 #else
     std::string env_delimiter = ":";
 #endif
-    // Process layer enable settings
-    SetLocalEnableSetting(GetConfigValue(SETTING_ENABLES), ",", settings_data->enables);
-    SetLocalEnableSetting(GetEnvVarValue(SETTING_ENABLES), env_delimiter, settings_data->enables);
 
-    SetValidationSetting(settings_data->enables, best_practices, SETTING_VALIDATE_BEST_PRACTICES);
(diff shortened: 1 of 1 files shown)


tests/positive/layer_settings_utils.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+#include "../framework/layer_validation_tests.h"
+#include "layer_settings_utils.h"
+
+class PositiveLayerSettingUtils : public VkPositiveLayerTest {};
+
+// These test check utils in the layer without needing to create a full Vulkan instance
+
+TEST_F(PositiveLayerSettingUtils, GetNextToken_list) {
+	TEST_DESCRIPTION("Test GetNextToken");
```
juan-lunarg, line 21: Indentation issue

juan-lunarg (approved):
Minor nitpick about indentation issue

juan-lunarg:
Does merging this also close https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5777?

christophe-lunarg:
> Does merging this also close #5777?

I am more trying to understand why #5777 is failing...

tests/positive/layer_settings_utils.cpp
christophe-lunarg: Fixed, this said others tests follow the same approach.

spencer-lunarg (approved):
LGTM, CI is only failing because warnings being marked as error, but should be easy to fix

PR #5800, opened 2023-05-04
macOS portability fixes

richard-lunarg:
A growing collection of VVL fixes for macOS CI tests. So far it is mostly adding support for the portability subset extension, or bailing if the test will never work on macOS.

### DIFF
--- a/tests/negative/arm_best_practices.cpp
+++ b/tests/negative/arm_best_practices.cpp
@@ -901,8 +901,9 @@ TEST_F(VkArmBestPracticesLayerTest, RedundantRenderPassStore) {
         GTEST_SKIP() << "Test temporarily disabled on S10 device";
     }
     
-    if (IsExtensionsEnabled(VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME))
+    if (IsExtensionsEnabled(VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME)) {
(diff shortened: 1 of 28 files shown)


tests/negative/arm_best_practices.cpp
```diff
@@ -900,6 +900,9 @@ TEST_F(VkArmBestPracticesLayerTest, RedundantRenderPassStore) {
     if (IsPlatform(kGalaxyS10)) {
         GTEST_SKIP() << "Test temporarily disabled on S10 device";
     }
+    
+    if (IsExtensionsEnabled(VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME))
+        GTEST_SKIP() << "VK_KHR_portability_subset enabled, skipping.\n";
```
spencer-lunarg, line 905: nit: I really have zero preference on HOW we do single line `if` statements, but because the one above all matches the same use of `{ }` I would like to keep things consistent as its been a lot of work to finally get all these tests consistent

richard-lunarg: Fair enough. Easier to add it to mine than remove it from everywhere else ;-)

ncesario-lunarg: :heavy_plus_sign: less error prone IMHO.

richard-lunarg:
Merge errors are becoming overwhelming due to the constant refactoring. I'm going to reboot this project, doing a handful of tests at a time rather than try to do them all at once.

PR #5801, opened 2023-05-04
syncval: aliasing design update

jzulauf-lunarg:
Design for changes needed to support accurate hazard detection of consistent and inconsistent memory aliases.

### DIFF
--- a/docs/synchronization.md
+++ b/docs/synchronization.md
@@ -1427,3 +1427,192 @@ If depthTestEnable is TRUE, otherwise, any, since with depth testing disabled th
 ### Host Synchronization commands
 
 **TODO/KNOWN LIMITATION:** Host synchronization not supported in phases 1 and 2.
+
+
+
+
+
+## Aliasing Support WIP Design
+
+Vulkan does not restrict device memory to be bound to only a single resource (buffer or image), but has the following restrictions as to when memory multiply bound and be interpreted in consistent ways.
+
+> Buffers, and linear image subresources in either the `VK_IMAGE_LAYOUT_PREINITIALIZED` or `VK_IMAGE_LAYOUT_GENERAL` layouts, are host-accessible subresources.
+>
+> ...
+>
+> If two aliases are both host-accessible, then they interpret the contents of the memory in consistent ways, and data written to one alias can be read by the other alias.
+
+> If two aliases are both images that were created with identical creation parameters, both were created with the `VK_IMAGE_CREATE_ALIAS_BIT` flag set, and both are bound identically to memory, ..., then they interpret the contents of the memory in consistent ways, and data written to one alias can be read by the other alias.
+
+When memory can be interpreted a consistent way, hazard detection, and state update can be done equivalently from all consistent aliases.  However, when consistent interpretation is not possible, accesses to any address in one bound resource must be interpreted as impacting *every* address in the aliased region.  If one does not know the organization of a resource, the effect must be treated as if they apply to the entire bound region.
+
+This violates some assumptions of the non-aliased model in the following ways:
+
+1) that a single *opaque range* mapping of a tiled image (or other opaque layout resources) is sufficient to store needed state
+2) that only a single, most recent write and barrier state is sufficient to characterize the `ResourceAccessState`
+5) that only a single, most recent read per stage is sufficient to characterize the `ResourceAccessState`
+
+These all arise from the need to integrate the read, write, and synchronization operations over the entire bound range into a single, pessimistic `ResourceAccessState` reflecting the accesses and barriers defining the potential hazards.
+
+
+
+### Memory Model
+
+To support hazard detection between Linear and opaque accesses, opaque resources (specifically tiled images) will have two distinct ranges in the `ValidationStateTracker::fake_memory` space, a Linear range and an opaque range.  The Linear range corresponds to the range of addresses needed for `VkMemoryRequirements::size` bytes at `VkBindImageMemoryInfo::memoryOffset`, offset by the FBA of `VkBindImageMemoryInfo::memory`. The opaque ranges for tiled images that have a consistent memory interpretation (a Consistent Alias Group) remains unchanged from the opaque range used in the pre-alias design.  All tiled images belong to a group of at least the image itself, with a unique opaque range. Note that since swapchain images are both always tiled, and do not map to any `VkDeviceMemory`, they do not have a valid Linear range, but only an opaque range shared by all images referencing the same swapchain/swapchain index pair.
+
+The effect of operations for the various resources:
+
+| Range  | Buffers/Linear Images                   | Tiled Images (or other opaque resources)                     |
+| ------ | --------------------------------------- | ------------------------------------------------------------ |
+| Linear | accesses affect byte-accurate locations | access at any location touches entire bound region (N/A swapchain images) |
+| Opaque | N/A                                     | accesses affect byte accurate location within the Synchronization Validation internal representation, limited to consistent aliases |
+
+Tiled images will have opaque ranges reserved which are uniquely mapped to all consistent aliases of a given `VkDeviceMemory` range (or swapchain image).  It is possible that multiple inconsistent tiled images may be mapped to the same `VkDeviceMemory` Thus, a given Linear range may map in a 1-N relationship to opaque ranges.  Hazard detection between *inconsistent* memory aliases is performed based on the Linear range, comprising the entire bound `VkDeviceMemory` range of an opaque resource.
+
+Memory Example:
+
+To understand this better, consider a single device memory allocation ***M***  of size 512 with the following allocations bound to it:
+
+| Type             | Name     | Bound Range | Notes                    |
+| ---------------- | -------- | ----------- | ------------------------ |
+| Buffer (liinear) | ***B***  | [0, 512)    |                          |
+| Image (Tiled)    | ***I0*** | [128, 256)  |                          |
+| Image (Tiled)    | ***I1*** | [128, 256)  | Consistent with ***I1*** |
+| Image (Linear)   | ***I2*** | [192 384)   |                          |
+
+The resources can be thought mapping to `fake_memory` in three ways --  in relation to other consistent aliases within the opaque range, and with inconsistent aliases within the allocation to which they are bound (if any), and within `fake_memory` as a whole.  Note that no accesses should span across different `VkDeviceMemory` ranges or opaque ranges. 
+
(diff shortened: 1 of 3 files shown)


docs/synchronization.md
```diff
@@ -1429,3 +1430,170 @@ If depthTestEnable is TRUE, otherwise, any, since with depth testing disabled th
...
+
+These all arise from the need to integrate the read, write, and synchronization operations over the entire bound range into a single, pessimistic `ResourceAccessState` reflecting the accesses and barriers defining the potential hazards.
+
+
+
+### Memory Model
+
+To support hazard detection between Linear and Idealized accesses, the current segregation of the access to Linear and Idealized `ResourceAccessRangeMap` objects needs to be abandoned. (Note the term Idealized refers to the mapping of tiled images to byte addresses using a simplified interpretation, not reflecting what any real implementation necessarily would use, but valid for the image size and format.)  Instead, resources (specifically tiled images) will have two distinct base addresses in the `ValidationStateTracker::fake_memory` space, a Linear binding and an Idealized binding.  The Linear binding is shared across all resource with the same `VkDeviceMemory`. The Idealized binding is shared across all tiled images that have a consistent memory interpretation (a Consistent Alias Group).  All tiled images belong to a group of at least the image itself, with a unique fake base address for the Idealized binding.
```
artem-lunarg, line 1466: One suggestion is to integrate terminology from the spec when possible. For example, the specification defines the term _host-accessible subresource_. The word _consistent_ is also used in the spec, but it's not part of the glossary and it is used as an additional characteristic. I think we can use them both, since _consistent_ is pretty good in describing the concept, and the use of _host-accessible subresource_ term itself is a from of re-use, where we can use all information that is related to that term in our document without providing additional definitions.

docs/synchronization.md
```diff
@@ -1429,3 +1430,170 @@ If depthTestEnable is TRUE, otherwise, any, since with depth testing disabled th
...
+
+The `AliasID` of `AliasID::kLinear` is reserved for all resources that are linearly mapped (buffers, non-tiled images).
+
+***Note:*** *One could make the `AliasID` and `idealized_base_address` synonymous, at the cost of require the `AliasID` to be the size of `VkDeviceMemory`.   (doubling the impact on the `ResourceAccessState` size).*
+
+#### `AliasID`Creation
+
+`AliasID` values are allocated during `BindImageMemory` and `CreateSwapchain` calls.  The `AliasID` is assign to an Image at `BindImageMemory`  and`CreatImage` calls depending on whether the image is bound to a`VkDeviceMemory` or `VkSwapchain` respectively.
```
artem-lunarg, line 1491: `CreatImage` -> `CreateSwapchain` ?

docs/synchronization.md
artem-lunarg, line 1466: From internal meeting. Clarify why both `linear` and `idealized` bindings are needed. A specific use case and/or a diagram would be helpful. 

For me it was helpful for understanding to realize that fake linear address is a mirror to real device space within the same VkDeviceMemory, so it's basically what you will see on the GPU. Also It was helpful to realize that memory range for `idealized binding` could be very different from full linear object representation, examples here would be useful how such idealized representation can be very different from linear range.

docs/synchronization.md
```diff
@@ -1429,3 +1430,202 @@ If depthTestEnable is TRUE, otherwise, any, since with depth testing disabled th
...
+
+We can see the usage of the various representation for hazard detection between the above operations.
+
+| Access pair <br />(before, after) | Idealized | Linear | Comments |
+| -------- | ---- | -------- | -------- |
+| w<sub>a</sub>, r<sub>b</sub> | No hazard | N/A | Since a consistent interpretation is possible (***A0***), only that representation need be tested. |
+| w<sub>a</sub>, w<sub>c</sub> | N/A | WAW | w<sub>a</sub> must be considered to affect the entire bound range, thus the portion of w<sub>c</sub> overlapping I1 constitutes a hazard |
+| w<sub>a</sub>, r<sub>d</sub> | N/A | RAW | the portion of r<sub>d</sub> overlapping constitutes a hazard for the same reason |
+| r<sub>b</sub>, r<sub>d</sub> | N/A | No hazard | r<sub>b</sub> and r<sub>d</sub> pose no hazard as they are both writes, regardless of representation |
```
artem-lunarg, line 1505: they are both writes -> they are both reads

docs/synchronization.md
```diff
@@ -1429,3 +1430,202 @@ If depthTestEnable is TRUE, otherwise, any, since with depth testing disabled th
...
+
+We can see the usage of the various representation for hazard detection between the above operations.
+
+| Access pair <br />(before, after) | Idealized | Linear | Comments |
```
artem-lunarg, line 1500: Excellent examples!

docs/synchronization.md
```diff
@@ -1429,3 +1430,202 @@ If depthTestEnable is TRUE, otherwise, any, since with depth testing disabled th
...
+| w<sub>a</sub>, r<sub>d</sub> | N/A | RAW | the portion of r<sub>d</sub> overlapping constitutes a hazard for the same reason |
+| r<sub>b</sub>, r<sub>d</sub> | N/A | No hazard | r<sub>b</sub> and r<sub>d</sub> pose no hazard as they are both writes, regardless of representation |
+| w<sub>c</sub>, r<sub>d</sub> | N/A | No hazard | As w<sub>c</sub> and r<sub>d</sub> are known at byte accuracy within *M* and do not overlap they do not hazard. |
+| w<sub>c</sub>, r<sub>b</sub> | N/A | RAW | r<sub>b</sub> must be considered to affect the entire bound range, thus the portion of w<sub>c</sub> overlapping I1 constitutes a hazard |
+
+Were ***I2*** to be tiled instead of linear, an additional representation *A1* would be needed to provided a consistent interpretation for operations involving ***I2*** with itself. Also in that case the  (w<sub>c</sub>, r<sub>d</sub>) access pair *would* constitute a hazard, given the same "any access affects all locations" logic applied to operations in **A0**.
+
+### Consistent Alias Group
+
+#### Access tagging
```
artem-lunarg, line 1513: One potential concern is that Aliasing is becoming too much first class citizen in the sense that everything has some aliasing ID comparing to the situation when non-aliased resources have AliasID set to zero/none/null, and only the aliased resources have non-default value  (linear resources have kLinear special value, but Vulkan production usage it's mostly tiled images and we always assign alias ID to them.

This is mostly about debugging experience. It would be nice to have a scheme that exposes brain pattern matching, so when I see aliasing ID 123, then I know it is an aliased resource. And, if in that situation we don't expect aliasing, this will immediately trigger an alarm, even if I was working on something else and did not think much about aliasing. Ironically having more alias ID data will not trigger such alarm (you know all tiled resources have alias ID and the visuall inspection does not bring new information). I guess here kicks in information theory, more data but less surprise and less information.

docs/synchronization.md
```diff
@@ -1429,3 +1430,202 @@ If depthTestEnable is TRUE, otherwise, any, since with depth testing disabled th
...
+
+
+
+### Hazard Detection
+
+For buffers and linear images, hazard detection is unchanged relative to pre-aliasing support.  Accesses are validated against the Linear address `VkDeviceMemory` range affected.  The effect of inconsistent memory alias access, if any, is recorded in the Linear binding by the update mechanism noted below.
+
+For tiled images every hazard detection and update operation now must be performed in against both Linear and Idealized bindings. Hazard detection is first performed against the image itself (including the consistent aliases). The hazard check is done on the exact Idealized memory location of the access or barrier operation. The second hazard check applies the same access or barrier, but against the entire Linear binding. Since the mapping of tiled image to Linear addresses is unknown, every access is considered to conflict with the entire bound Linear range.  During this check the Linear `ResourceAccessState` information reflecting accesses *consistent* with the tile image must be ignored.  This requirement means that accesses must be tagged with an the `AliasID`. (Each read and write state.)
```
artem-lunarg, line 1547: > During this check the Linear ResourceAccessState information reflecting accesses consistent with the tile image must be ignored.

Could you please clarify this sentence? Is it bout ignoring accesses that previously were integrated from tiled image into linear range because the linear projection is not-independent and just reflects tiled accesses? But I don't have confidence in this explanation, it is just my best guess.

docs/synchronization.md
```diff
@@ -1427,3 +1427,192 @@ If depthTestEnable is TRUE, otherwise, any, since with depth testing disabled th
...
+Memory Example:
+
+To understand this better, consider a single device memory allocation ***M***  of size 512 with the following allocations bound to it:
+
+| Type             | Name     | Bound Range | Notes                    |
+| ---------------- | -------- | ----------- | ------------------------ |
+| Buffer (liinear) | ***B***  | [0, 512)    |                          |
+| Image (Tiled)    | ***I0*** | [128, 256)  |                          |
+| Image (Tiled)    | ***I1*** | [128, 256)  | Consistent with ***I1*** |
+| Image (Linear)   | ***I2*** | [128, 384)  |                          |
```
artem-lunarg, line 1481: Image (Linear): [192, 384) to match the image?

docs/synchronization.md
```diff
@@ -1427,3 +1427,192 @@ If depthTestEnable is TRUE, otherwise, any, since with depth testing disabled th
...
+##### Recorded Usage
+
+The `DetectHazard` taking recorded usage (in the `DetectFirstUseHazard` path) will also need modification to 
+
+* Accept argument to specific the scope rule used for Linear or opaque ranges (***Note:*** *likely need to tag all `ResourceAccessStates` with Linear/opaque boolean*)
+* Pass extracted recorded `AliasID` and scope rule information from first use information to underlying `DetectHazard` overrides. 
+
+### AccessContext Updates
+
+There are several `AccessContext::DetectHazard` member that apply to images.  Theses differ by hazard detection functors use, and the parameter to the range generators.  Eventually all of these call into the generic AccessContext::DetectHazard operating on a single range.  In order to avoid duplication of the aliasing logic, each of these need to be modified to supply a range generation object factory callable encapsulating the member specific range generation function arguments and creation, and taking the object base address as an argument.
```
artem-lunarg, line 1611: nit: typo: member that apply -> members that apply, Theses -> These.

artem-lunarg (approved):

artem-lunarg:
Moved markdown document to internal location: https://drive.google.com/drive/folders/1ZQS99hzybnsnEytMDiLjdFA0nUgWjaJY?usp=drive_link

When we start working on the actual task then the official syncval design document will be updated with details of the final solution.

PR #5804, opened 2023-05-05
Move rest of dynamic depth/stencil to lastBoundState

spencer-lunarg:
This does

- Use `LAST_BOUND_STATE` to pass around draw time validation that need both
- Move dynamic state checks for Depth/Stencil into `LAST_BOUND_STATE`

### DIFF
--- a/layers/sync/sync_validation.cpp
+++ b/layers/sync/sync_validation.cpp
@@ -2593,7 +2593,9 @@ bool RenderPassAccessContext::ValidateDrawSubpassAttachment(const CommandExecuti
                                                             const CMD_BUFFER_STATE &cmd_buffer, CMD_TYPE cmd_type) const {
     bool skip = false;
     const auto &sync_state = exec_context.GetSyncState();
-    const auto *pipe = cmd_buffer.GetCurrentPipeline(VK_PIPELINE_BIND_POINT_GRAPHICS);
+    const auto lv_bind_point = ConvertToLvlBindPoint(VK_PIPELINE_BIND_POINT_GRAPHICS);
+    const auto last_bound_state = cmd_buffer.lastBound[lv_bind_point];
+    const auto *pipe = last_bound_state.pipeline_state;
     if (!pipe) {
         return skip;
     }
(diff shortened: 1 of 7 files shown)


layers/sync/sync_validation.cpp
```diff
@@ -2737,18 +2734,11 @@ void RenderPassAccessContext::RecordDrawSubpassAttachment(const CMD_BUFFER_STATE
...
-                                           : ds_state->depthTestEnable;
-        const bool stencil_test_enable = pipe->IsDynamic(VK_DYNAMIC_STATE_DEPTH_WRITE_ENABLE)
-                                             ? cmd_buffer.dynamic_state_value.stencil_test_enable
-                                             : ds_state->stencilTestEnable;
+        const bool depth_write_enable = last_bound_state.IsDepthWriteEnable();  // implicitly means DepthTestEnable is set
+        const bool stencil_test_enable = last_bound_state.IsStencilTestEnable();
 
         // PHASE1 TODO: These validation should be in core_checks.
-        if (has_depth && !FormatIsStencilOnly(view_state.create_info.format) && depth_test_enable && depth_write_enable &&
+        if (has_depth && !FormatIsStencilOnly(view_state.create_info.format) && depth_write_enable &&
```
arno-lunarg, line 2741: `depth_test_enable` is not used anymore, is it intended?

arno-lunarg (approved):
LGTM! One question

artem-lunarg (approved):

layers/sync/sync_validation.cpp
artem-lunarg: I'll take a liberty to answer. `IsDepthWriteEnable()` now includes `IsDepthTestEnable()`, it was part of recent fixes.

This raises another question. Can this create some confusion, because this can be interpreted in two ways 'function that check write enable state' and 'function that check if write enable functionality is active'. The situation to imagine that someone has a VUID that mentions only write-enable state and then it's easy to use `IsDepthWriteEnable()`. Not sure if it's realistic example though.

arno-lunarg: Thank you for the explanation, it makes sense! 
According to the comment in `IsDepthWriteEnable` (https://github.com/spencer-lunarg/Vulkan-ValidationLayers/blob/ff2cda67bea9755ad05df378cd036bc3f2c2f6b7/layers/state_tracker/pipeline_state.cpp#L877), not sure this example is possible. But with Vukan being Vulkan...

PR #5807, opened 2023-05-05
 WIP  test new CI

Tony-LunarG:


juan-lunarg (commented):
Is it possible to do the following with the new CI workflow?

1.) Only run vvl.yml when formatting.yml passes

2.) Only run lunarg-ci when vvl.yml passes

Tony-LunarG:
> Is it possible to do the following with the new CI workflow?
> 
> 1.) Only run vvl.yml when formatting.yml passes
> 
> 2.) Only run lunarg-ci when vvl.yml passes

1 - I don't think so - The server doesn't have anything to do with either of those running
2 - Maybe - it would take some investigation, but I think the server could check the status of other check runs before starting lunarg-ci

PR #5808, opened 2023-05-05
v1.3.250 update

mikes-lunarg:


ncesario-lunarg (approved):

layers/stateless/sl_image.cpp
```diff
@@ -82,13 +82,13 @@ bool StatelessValidation::manual_PreCallValidateCreateImage(VkDevice device, con
                                  string_VkImageType(type));
             } else {
```
juan-lunarg, line 83: Remove the else statement as well. It's no longer needed.

juan-lunarg (changes requested):

layers/stateless/sl_image.cpp
juan-lunarg: I believe this will also require updating `NegativeImage.CubeCompatibleMustBeImageType2D` since `VUID-VkImageCreateInfo-flags-08866` will now trigger.

juan-lunarg (approved):

Issue #5551, opened 2023-03-31
Dynamic states are not tracked correctly with mesh pipelines


ziga-lunarg:
The CTS test
dEQP-VK.pipeline.pipeline_library.extended_dynamic_state.cmd_buffer_start.prim_restart_enable_bind_unused_ms
reveals a bug in tracking dynamic states.

The tests sets the dynamic state VK_DYNAMIC_STATE_PRIMITIVE_RESTART_ENABLE, then binds a mesh pipeline (which can't have this state at all, as per VUID-VkGraphicsPipelineCreateInfo-pDynamicStates-07066), then binds a graphics pipeline which uses the dynamic state VK_DYNAMIC_STATE_PRIMITIVE_RESTART_ENABLE and issues a draw call, but a VUID is reported that the dynamic state was not set. 

Binding a mesh pipeline should not disturb the dynamic states which it can't have.

spencer-lunarg:
So looking into this, this is a bug, but not for mesh pipelines

the following pattern forces the issue

```c++
    VkCommandBufferObj m_commandBuffer(m_device, m_commandPool);
    m_commandBuffer.begin();
    vkCmdSetPrimitiveRestartEnableEXT(m_commandBuffer.handle(), VK_TRUE);
    m_commandBuffer.BeginRenderPass(m_renderPassBeginInfo);
    vk::CmdBindPipeline(m_commandBuffer.handle(), VK_PIPELINE_BIND_POINT_GRAPHICS, pipe_no_dynamic.pipeline_);
    vk::CmdBindPipeline(m_commandBuffer.handle(), VK_PIPELINE_BIND_POINT_GRAPHICS, pipe.pipeline_);
    vk::CmdDraw(m_commandBuffer.handle(), 1, 1, 0, 0);
```

going

- Setting state
- Binding pipeline that uses it
- Binding pipeline that doesn't use it

causes the issues... will fix the dynamic tracking state

ziga-lunarg:
I think the behaviour in your example is correct, if you bind a pipeline that has a static state you have to set the dynamic state again if you plan to use it after. The problem with mesh pipelines is that it doesn't have some states at all, so it shouldn't unset those states.

spencer-lunarg:
Adding test here to help reproduce in future

```cpp

TEST_F(VkPositiveLayerTest, DynamicStateDoublePipelineBind) {
[...]

Issue #5677, opened 2023-04-18
Potential bug in depthWriteEnabled traccking with dynamicState?


changanlao:
I am working on ANGLE, and seeing a strange VVL error that might possibly a VVL bug. 
Steps to reproduce:
1) Setup ANGLE and able to run end2end tests
2) Revert this CL which I try to workaround the VVL error
     https://chromium-review.googlesource.com/c/angle/angle/+/4442446
3) Run ClearTestES3.RepeatedStencilClearWithBlitInBetween/ES3_Vulkan on windows intel GPU, and it some times report this VVL error:
[ RUN      ] ClearTestES3.RepeatedStencilClearWithBlitInBetween/ES3_Vulkan
..\..\src\tests\test_utils\ANGLETest.cpp(65): error: RendererVk.cpp:777 (DebugUtilsMessenger): [ VUID-vkCmdDraw-None-06886 ] Validation Error: [ VUID-vkCmdDraw-None-06886 ] Object 0: handle = 0x2660000000266, type = VK_OBJECT_TYPE_PIPELINE; Object 1: handle = 0x26b000000026b, type = VK_OBJECT_TYPE_RENDER_PASS; Object 2: handle = 0x27654d828a8, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xf157442b | vkCmdDraw: depthWriteEnable is VK_TRUE, while the layout (VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL) of the depth aspect of the depth/stencil attachment in the render pass is read only. The Vulkan spec states: If the current render pass instance uses a depth/stencil attachment with a read-only layout for the depth aspect, depth writes must be disabled (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdDraw-None-06886)
                            Object: 0x2660000000266 (type = Pipeline(19))
                            Object: 0x26b000000026b (type = Render Pass(18))
                            Object: 0x27654d828a8 (type = Command Buffer(6))

One example can be seen in the ANGLE bots (if you have access):  
      https://chromium-swarm.appspot.com/task?id=61aa817c7663ef10

I have debugged this a bit and here is what I have found so far:

intel GPU has vulkan dynamic state feature enabled:
    supportsExtendedDynamicState is true 
    supportsExtendedDynamicState2 is true.

[...]

PR #5811, opened 2023-05-06
Fix how Dynamic State is tracked

spencer-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5551 (maybe even https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5677)

Refactor how we track dynamic state, the previous `status` had bugs, the way  around it was simply to us the same `CBDynamicFlags` in both the Command Buffer and Pipeline. Before, we were trying to track Pipeline state in `CMB_BUFFER_STATE`

### DIFF
--- a/layers/core_checks/cc_cmd_buffer_dynamic.cpp
+++ b/layers/core_checks/cc_cmd_buffer_dynamic.cpp
@@ -361,14 +361,15 @@ bool CoreChecks::ValidateDrawDynamicState(const LAST_BOUND_STATE &last_bound_sta
         const auto color_blend_state = cb_state.GetCurrentPipeline(VK_PIPELINE_BIND_POINT_GRAPHICS)->ColorBlendState();
         if (color_blend_state) {
             uint32_t blend_attachment_count = color_blend_state->attachmentCount;
-            if (cb_state.dynamicColorWriteEnableAttachmentCount < blend_attachment_count) {
+            uint32_t dynamic_attachment_count = cb_state.dynamic_state_value.color_write_enable_attachment_count;
+            if (dynamic_attachment_count < blend_attachment_count) {
                 const LogObjectList objlist(cb_state.commandBuffer(), pipeline.pipeline());
                 skip |= LogError(
                     objlist, vuid.dynamic_color_write_enable_count_07750,
                     "%s(): Currently bound pipeline was created with VkPipelineColorBlendStateCreateInfo::attachmentCount %" PRIu32
                     " and VK_DYNAMIC_STATE_COLOR_WRITE_ENABLE_EXT, but the number of attachments written by "
                     "vkCmdSetColorWriteEnableEXT() is %" PRIu32 ".",
-                    CommandTypeString(cmd_type), blend_attachment_count, cb_state.dynamicColorWriteEnableAttachmentCount);
+                    CommandTypeString(cmd_type), blend_attachment_count, dynamic_attachment_count);
             }
         }
     }
--- a/layers/core_checks/cc_cmd_buffer_dynamic.cpp
+++ b/layers/core_checks/cc_cmd_buffer_dynamic.cpp
@@ -23,12 +23,11 @@
 #include "generated/chassis.h"
 #include "core_validation.h"
 
-// Check object status for selected flag state
-bool CoreChecks::ValidateCBDynamicStatus(const CMD_BUFFER_STATE &cb_state, CBDynamicState dynamic_state, CMD_TYPE cmd_type,
-                                         const char *msg_code) const {
-    if (!(cb_state.status[dynamic_state])) {
-        return LogError(cb_state.commandBuffer(), msg_code, "%s: %s state not set for this command buffer.",
-                        CommandTypeString(cmd_type), DynamicStateToString(dynamic_state));
+bool CoreChecks::ValidateDynamicStateIsSet(CBDynamicFlags state_status_cb, CBDynamicState dynamic_state,
+                                           const LogObjectList &objlist, CMD_TYPE cmd_type, const char *msg_code) const {
+    if (!state_status_cb[dynamic_state]) {
+        return LogError(objlist, msg_code, "%s: %s state not set for this command buffer.", CommandTypeString(cmd_type),
+                        DynamicStateToString(dynamic_state));
     }
     return false;
 }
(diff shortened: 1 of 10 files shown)


layers/state_tracker/state_tracker.cpp
```diff
@@ -2729,18 +2729,14 @@ void ValidationStateTracker::PreCallRecordCmdBindPipeline(VkCommandBuffer comman
         const auto *raster_state = pipe_state->RasterizationState();
         const bool rasterization_enabled = raster_state && !raster_state->rasterizerDiscardEnable;
         const auto *viewport_state = pipe_state->ViewportState();
-        const auto *dynamic_state = pipe_state->DynamicState();
-        cb_state->status &= ~cb_state->static_status;
-        cb_state->static_status = MakeStaticStateMask(dynamic_state ? dynamic_state->ptr() : nullptr);
```
spencer-lunarg, line 2734: This logic was a hack to basically track the `PIPELINE_STATE` in the `CMD_BUFFER_STATE` but now just can call into `pipeline.dynamic_state` when needed

layers/state_tracker/cmd_buffer_state.cpp
```diff
@@ -1078,7 +1070,20 @@ void CMD_BUFFER_STATE::UpdatePipelineState(CMD_TYPE cmd_type, const VkPipelineBi
...
-    if (pipe && VK_NULL_HANDLE != last_bound.pipeline_layout) {
+    if (!pipe) {
+        return;
+    }
+
+    // Update the consumed viewport/scissor count.
+    {
+        uint32_t &used = usedViewportScissorCount;
+        used = std::max(used, pipelineStaticViewportCount);
+        used = std::max(used, pipelineStaticScissorCount);
```
juan-lunarg, line 1081: std::max can take a initializer_list

```suggestion
        usedViewportScissorCount = std::max({usedViewportScissorCount, pipelineStaticViewportCount, pipelineStaticScissorCount});
```

layers/state_tracker/cmd_buffer_state.cpp
```diff
@@ -1150,7 +1155,7 @@ void CMD_BUFFER_STATE::UpdatePipelineState(CMD_TYPE cmd_type, const VkPipelineBi
             }
         }
     }
-    if (pipe && pipe->vertex_input_state && !pipe->vertex_input_state->binding_descriptions.empty()) {
+    if (pipe->vertex_input_state && !pipe->vertex_input_state->binding_descriptions.empty()) {
         vertex_buffer_used = true;
     }
```
juan-lunarg, line 1160:
```suggestion
```

layers/state_tracker/cmd_buffer_state.cpp
```diff
@@ -1078,7 +1070,20 @@ void CMD_BUFFER_STATE::UpdatePipelineState(CMD_TYPE cmd_type, const VkPipelineBi
...
+    // Update the consumed viewport/scissor count.
+    {
+        uint32_t &used = usedViewportScissorCount;
+        used = std::max(used, pipelineStaticViewportCount);
+        used = std::max(used, pipelineStaticScissorCount);
+        usedDynamicViewportCount |= pipe->IsDynamic(VK_DYNAMIC_STATE_VIEWPORT_WITH_COUNT);
+        usedDynamicScissorCount |= pipe->IsDynamic(VK_DYNAMIC_STATE_SCISSOR_WITH_COUNT);
+    }
+
+    if (VK_NULL_HANDLE != last_bound.pipeline_layout) {
```
juan-lunarg, line 1086:
```suggestion
    if (last_bound.pipeline_layout == VK_NULL_HANDLE) {
        return;
    }
```

layers/state_tracker/state_tracker.cpp
```diff
@@ -2752,14 +2748,14 @@ void ValidationStateTracker::PreCallRecordCmdBindPipeline(VkCommandBuffer comman
         // I am taking the latter interpretation based on the implementation details of NVIDIA's Vulkan driver.
         if (!has_dynamic_viewport_count) {
             cb_state->trashedViewportCount = true;
-            if (rasterization_enabled && (cb_state->static_status[CB_DYNAMIC_STATE_VIEWPORT])) {
+            if (rasterization_enabled && (!pipe_state->IsDynamic(VK_DYNAMIC_STATE_VIEWPORT))) {
                 cb_state->trashedViewportMask |= (uint32_t(1) << viewport_state->viewportCount) - 1u;
                 // should become = ~uint32_t(0) if the other interpretation is correct.
```
juan-lunarg, line 2753: Should this be an assert?

layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -38,154 +37,179 @@ bool CoreChecks::ValidateDynamicStateSetStatus(const CMD_BUFFER_STATE &cb_state,
...
+        (cb_state.cmd_set_status_pipeline ^ pipeline.dynamic_state) & cb_state.cmd_set_status_pipeline;
+    if (not_set_pipeline_status.any()) {
+        skip |= LogError(objlist, vuid.dynamic_state_setting_commands_02859,
+                         "%s: %s doesn't set up %s, but it calls the related dynamic state setting commands.",
+                         CommandTypeString(cmd_type), report_data->FormatHandle(pipeline.pipeline()).c_str(),
+                         DynamicStatesToString(not_set_pipeline_status).c_str());
+    }
 
+    // build the mask of what has been set in the Pipeline, but yet to be set in the Command Buffer
+    const CBDynamicFlags not_set_cb_status = (cb_state.cmd_set_status_cb ^ pipeline.dynamic_state) & pipeline.dynamic_state;
```
juan-lunarg, line 53: FFTI

This variable name makes the code confusing to read. Same with `not_set_pipeline_status`.

The comments make sense. The code is just hard to follow.

In general the `not_` variables make the code harder to read.

layers/state_tracker/cmd_buffer_state.h
```diff
@@ -180,12 +180,9 @@ class CMD_BUFFER_STATE : public REFCOUNTED_NODE {
...
 
-    // Dynamic State
-    CBDynamicFlags status;          // Track status of various bindings on cmd buffer
-    CBDynamicFlags static_status;   // All state bits provided by current graphics pipeline
-                                    // rather than dynamic state
-    CBDynamicFlags dynamic_status;  // dynamic state set up in pipeline
-
+    // Track status of all vkCmdSet* calls, if 1, means it was set
+    CBDynamicFlags cmd_set_status_cb;        // for lifetime of CommandBuffer
+    CBDynamicFlags cmd_set_status_pipeline;  // for lifetime since last bound pipeline
```
juan-lunarg, line 185: FFTI

What about something like this? Having them together in a struct IMO helps readability.

```suggestion
    struct Status {
        CBDynamicFlags cb;
        CBDynamicFlags pipeline;
    } set_status;
```

layers/state_tracker/state_tracker.cpp
spencer-lunarg: The Viewport/Scissor logic needs refactoring, but for this PR, just trying to "not break it" and going to followup after and unify this all

layers/state_tracker/cmd_buffer_state.h
spencer-lunarg: :clap: agree, like this

layers/state_tracker/cmd_buffer_state.cpp
spencer-lunarg: thanks, I just moved this, but like fixing it while here

layers/core_checks/cc_cmd_buffer_dynamic.cpp
spencer-lunarg: agree

```patch
- not_set_cb_status
+ unset_status_cb
```

artem-lunarg: Personally for me, `unset_status_cb` still have semantics of inversion but is masked by another word. Sometimes the advice is to invert the semantics, so it will end up with `set_cb_status` or just `cb_status`, not sure how it can be done here. Also, sometimes inversed term can be more natural to the problem area. Unfortunately don't have ideas about other options.

layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -39,154 +38,179 @@ bool CoreChecks::ValidateDynamicStateSetStatus(const LAST_BOUND_STATE &last_boun
...
-            ValidateCBDynamicStatus(cb_state, CB_DYNAMIC_STATE_DEPTH_WRITE_ENABLE, cmd_type, vuid.dynamic_depth_write_enable_07844);
-        skip |= ValidateCBDynamicStatus(cb_state, CB_DYNAMIC_STATE_DEPTH_COMPARE_OP, cmd_type, vuid.dynamic_depth_compare_op_07845);
-        skip |= ValidateCBDynamicStatus(cb_state, CB_DYNAMIC_STATE_DEPTH_BOUNDS_TEST_ENABLE, cmd_type,
-                                        vuid.dynamic_depth_bound_test_enable_07846);
-        skip |= ValidateCBDynamicStatus(cb_state, CB_DYNAMIC_STATE_STENCIL_TEST_ENABLE, cmd_type,
-                                        vuid.dynamic_stencil_test_enable_07847);
-        skip |= ValidateCBDynamicStatus(cb_state, CB_DYNAMIC_STATE_STENCIL_OP, cmd_type, vuid.dynamic_stencil_op_07848);
-        skip |= ValidateCBDynamicStatus(cb_state, CB_DYNAMIC_STATE_VERTEX_INPUT_BINDING_STRIDE, cmd_type,
-                                        vuid.vertex_input_binding_stride_04884);
+        skip |= ValidateDynamicStateIsSet(not_set_cb_status, CB_DYNAMIC_STATE_CULL_MODE, objlist, cmd_type,
```
artem-lunarg: Question. Each `ValidateDynamicStateIsSet` call validations a single flag. Is it allowed to have some flag set in `not_set_cb_status` that according to logic should not be checked, or does `ValidateDynamicStateIsSet` have to check all possible flags?

layers/state_tracker/cmd_buffer_state.cpp
```diff
@@ -1078,7 +1070,21 @@ void CMD_BUFFER_STATE::UpdatePipelineState(CMD_TYPE cmd_type, const VkPipelineBi
...
+    }
+
+    // Update the consumed viewport/scissor count.
+    {
+        usedViewportScissorCount = std::max({usedViewportScissorCount, pipelineStaticViewportCount, pipelineStaticScissorCount});
+        usedDynamicViewportCount |= pipe->IsDynamic(VK_DYNAMIC_STATE_VIEWPORT_WITH_COUNT);
+        usedDynamicScissorCount |= pipe->IsDynamic(VK_DYNAMIC_STATE_SCISSOR_WITH_COUNT);
+    }
+
+    if (last_bound.pipeline_layout != VK_NULL_HANDLE) {
```
juan-lunarg, line 1087: You can return early. This avoids double nested logic.

```suggestion
    if (last_bound.pipeline_layout == VK_NULL_HANDLE) {
        return;
    }
```

layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -23,12 +23,11 @@
 #include "generated/chassis.h"
 #include "core_validation.h"
 
-// Check object status for selected flag state
-bool CoreChecks::ValidateCBDynamicStatus(const CMD_BUFFER_STATE &cb_state, CBDynamicState dynamic_state, CMD_TYPE cmd_type,
-                                         const char *msg_code) const {
-    if (!(cb_state.status[dynamic_state])) {
-        return LogError(cb_state.commandBuffer(), msg_code, "%s: %s state not set for this command buffer.",
-                        CommandTypeString(cmd_type), DynamicStateToString(dynamic_state));
```
juan-lunarg, line 31: This is much more readable to me then the new version.

If the intent behind the `not_set_status` is to remove the `!`

What about just returning early?

```
if (cb_state.status[dynamic_state]) {
    return false;
}

return LogError(...);
```

layers/state_tracker/cmd_buffer_state.h
```diff
@@ -216,14 +216,16 @@ class CMD_BUFFER_STATE : public REFCOUNTED_NODE {
         // VK_DYNAMIC_STATE_LINE_STIPPLE_ENABLE_EXT
         bool stippled_line_enable;
 
+        uint32_t color_write_enable_attachment_count;
```
juan-lunarg, line 219: Initialize this in the header file. 


```suggestion
        uint32_t color_write_enable_attachment_count = 0;
```

Also does this value need to be set to 0 when reseting? I only see this value being assigned a value in 1 location.

IE should this value be set to 0 in the reset helper function.

layers/state_tracker/state_tracker.cpp
```diff
@@ -5441,24 +5437,17 @@ void ValidationStateTracker::PostCallRecordCmdSetVertexInputEXT(
...
     }
     cb_state->RecordStateCmd(CMD_SETVERTEXINPUTEXT, status_flags);
 }
 
 void ValidationStateTracker::PostCallRecordCmdSetColorWriteEnableEXT(VkCommandBuffer commandBuffer, uint32_t attachmentCount,
                                                                      const VkBool32 *pColorWriteEnables) {
     auto cb_state = GetWrite<CMD_BUFFER_STATE>(commandBuffer);
-    cb_state->RecordColorWriteEnableStateCmd(CMD_SETCOLORWRITEENABLEEXT, CB_DYNAMIC_STATE_COLOR_WRITE_ENABLE_EXT, attachmentCount);
+    cb_state->RecordStateCmd(CMD_SETCOLORWRITEENABLEEXT, CB_DYNAMIC_STATE_COLOR_WRITE_ENABLE_EXT);
+    cb_state->dynamic_state_value.color_write_enable_attachment_count = attachmentCount;
```
juan-lunarg, line 5450: This is what I'm referencing I see this value only being assigned here.

So on the next frame if the command buffer is used again and doesn't call `CmdSetColorWriteEnableEXT` what is the value of `color_write_enable_attachment_count`?

layers/state_tracker/cmd_buffer_state.cpp
spencer-lunarg: so the issue is when we decide to add more logic under to add to `UpdatePipelineState` we now have to be careful, either make sure to add it above, or risk a bug someone didn't see this sudden `return` case in the middle of the function (I have hit this a few times before in the VVL with early returns mid-functions)

I think if we are nesting to much, it should be pulled out to another function

layers/state_tracker/cmd_buffer_state.h
spencer-lunarg: right below

```
 // setting the state, it will fail in ValidateDynamicStateIsSet() for us
```

layers/core_checks/cc_cmd_buffer_dynamic.cpp
spencer-lunarg: ok, so `unset` was suppose to mean "Dynamic State non set", which means "static state set" so can rename this to `CBDynamicFlags static_state_cb`

layers/state_tracker/state_tracker.cpp
spencer-lunarg: this is `dynamic_state_value.color_write_enable_attachment_count`

so only used when things are dynamic

>  and doesn't call CmdSetColorWriteEnableEXT

You hit this VU

> VUID-vkCmdDraw-None-07749
> If the bound graphics pipeline state was created with the VK_DYNAMIC_STATE_COLOR_WRITE_ENABLE_EXT dynamic state enabled then [vkCmdSetColorWriteEnableEXT](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#vkCmdSetColorWriteEnableEXT) must have been called in the current command buffer prior to this drawing command

Basically the values are "undefined" in the spec so "resetting" them to some arbitrary value like zero, means nothing

layers/core_checks/cc_cmd_buffer_dynamic.cpp
spencer-lunarg: yep renamed, agree was not well worded

layers/core_checks/cc_cmd_buffer_dynamic.cpp
spencer-lunarg: >  Is it allowed to have some flag set in not_set_cb_status that according to logic should not be checked

yes, below you will find there are some more complex checks that are made, these are the "simple" ones

layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -39,154 +38,180 @@ bool CoreChecks::ValidateDynamicStateSetStatus(const LAST_BOUND_STATE &last_boun
...
+        (cb_state.dynamic_state_status.pipeline ^ pipeline.dynamic_state) & cb_state.dynamic_state_status.pipeline;
+    if (unset_status_pipeline.any()) {
+        skip |= LogError(objlist, vuid.dynamic_state_setting_commands_02859,
+                         "%s: %s doesn't set up %s, but it calls the related dynamic state setting commands.",
+                         CommandTypeString(cmd_type), report_data->FormatHandle(pipeline.pipeline()).c_str(),
+                         DynamicStatesToString(unset_status_pipeline).c_str());
+    }
+
+    // build the mask of what has been set in the Pipeline, but yet to be set in the Command Buffer
+    const CBDynamicFlags state_status_cb = ~((cb_state.dynamic_state_status.cb ^ pipeline.dynamic_state) & pipeline.dynamic_state);
```
spencer-lunarg, line 54: @juan-lunarg @artem-lunarg sorry, I just flipped the logic here, remove the `un`/`non` in the name, and keeps the logic easier in `ValidateDynamicStateIsSet` as well

This whole Dynamic Code will most likely need to be fix when we add `VK_EXT_shader_object`

juan-lunarg (commented):
LGTM

layers/state_tracker/cmd_buffer_state.cpp
juan-lunarg: > or risk a bug someone didn't see this sudden return case in the middle of the function

I see your point. But also that is exactly why we try to require tests when writing validation.

PR #5813, opened 2023-05-06
layers: Add external dedicated memory VU 00639

artem-lunarg:


tests/framework/layer_validation_tests.h
```diff
@@ -1067,6 +1067,9 @@ VkExternalMemoryHandleTypeFlags FindSupportedExternalMemoryHandleTypes(VkPhysica
                                                                        const VkBufferCreateInfo &buffer_create_info,
                                                                        VkExternalMemoryFeatureFlags requested_features);
 
+bool HandleTypeNeedsDedicatedAllocation(VkPhysicalDevice gpu, const VkBufferCreateInfo &buffer_create_info,
```
spencer-lunarg, line 1070: is this something we could just move to `tests/negative/external_memory_sync.cpp`? Not worry about it for this PR as I see other related functions that could go there as well

spencer-lunarg (approved):
LGTM  - good us of the `IsExport()`

tests/framework/layer_validation_tests.h
artem-lunarg: Yes, I was not quite comfortable with adding all other functions in the layer_validation_tests.h in the first place. Some clean up would be good. Positive tests might need that too.

layers/core_checks/cc_device_memory.cpp
```diff
@@ -1276,8 +1291,10 @@ bool CoreChecks::ValidateBindImageMemory(uint32_t bindInfoCount, const VkBindIma
                                 string_VkImageType(image_info.type), string_VkImageTiling(image_info.tiling),
                                 string_VkImageUsageFlags(image_info.usage).c_str(),
                                 string_VkImageCreateFlags(image_info.flags).c_str(), string_VkResult(result));
-                        } else if ((external_properties.externalMemoryProperties.externalMemoryFeatures &
-                                    VK_EXTERNAL_MEMORY_FEATURE_EXPORTABLE_BIT) == 0) {
+                            return;  // this exits lambda, not parent function
```
ncesario-lunarg, line 1294: Yeah, this "lambda" is getting pretty lengthy. Maybe it's worth looking into whether or not a separate function makes sense for clarity and utilizing `std::bind`?

ncesario-lunarg (approved):
One question about the "lambda that continues to grow," but LGTM!

layers/core_checks/cc_device_memory.cpp
artem-lunarg: Experimented a bit with std::bind a helper member-function, it gets a bit complicated. I think one problem with a piece of code under lambda is that it does not have well defined semantics (abstracted body of a loop over flags) and also touches a lot of local variables. Probably I need to extract more related code (not only lambda), so it's more or less self-sufficient. Going to postpone this for later (other external VUID work is planned, could be more obvious when more functionality is available).

ncesario-lunarg: For sure, not saying a resolution to the "long lambda" needs to be in this PR, but it's something we might want to think about in general.

PR #5819, opened 2023-05-08
RFC: Test initialization streamlining

ncesario-lunarg:
First attempt at streamlining test initialization checks, specifically the
```cpp
    }
```
`DeviceValidationVersion() < ...`, etc. checks.

### DIFF
--- a/tests/negative/others.cpp
+++ b/tests/negative/others.cpp
@@ -150,26 +150,22 @@ TEST_F(VkLayerTest, UnsupportedPnextApiVersion) {
 TEST_F(VkLayerTest, PrivateDataExtTest) {
     TEST_DESCRIPTION("Test private data extension use.");
 
-    AddRequiredExtensions(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
-    AddRequiredExtensions(VK_EXT_PRIVATE_DATA_EXTENSION_NAME);
-
-    ASSERT_NO_FATAL_FAILURE(InitFramework());
-
-    if (IsPlatform(kMockICD)) {
-        GTEST_SKIP() << "Test not supported by MockICD";
-    }
-
-    if (!AreRequiredExtensionsEnabled()) {
-        GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported";
-    }
-
     auto private_data_features = LvlInitStruct<VkPhysicalDevicePrivateDataFeaturesEXT>();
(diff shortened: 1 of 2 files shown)


tests/negative/others.cpp
```diff
@@ -150,27 +150,23 @@ TEST_F(VkLayerTest, UnsupportedPnextApiVersion) {
...
-
-    if (IsPlatform(kMockICD)) {
-        GTEST_SKIP() << "Test not supported by MockICD";
-    }
+    auto private_data_features = LvlInitStruct<VkPhysicalDevicePrivateDataFeaturesEXT>();
 
-    if (!AreRequiredExtensionsEnabled()) {
-        GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported";
-    }
+    InitInfo init_info = {{// Required extensions
```
spencer-lunarg, line 155: Where does a minimum version go?

tests/negative/others.cpp
```diff
@@ -150,27 +150,23 @@ TEST_F(VkLayerTest, UnsupportedPnextApiVersion) {
...
+                           VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME, VK_EXT_PRIVATE_DATA_EXTENSION_NAME},
+                          {
+                              // Feature checks
+                              {private_data_features, &VkPhysicalDevicePrivateDataFeaturesEXT::privateData},
```
spencer-lunarg, line 159: how do I say "Don't enable some feature bit"?

tests/negative/others.cpp
ncesario-lunarg: This is done by setting the `api_version` field in `InitInfo`.

tests/negative/others.cpp
ncesario-lunarg: There are a few different ways this could be achieved. The way things are currently setup, you would do something like:
```cpp
init_info.required_feature_checks.emplace_back([&private_data_features]() -> bool {
    private_data_features.privateData = VK_FALSE; // This will get passed to CreateDevice
    return true;
});
```
But a specific list could be added for this as well (e.g., `disable_feature_checks`).

This PR is more to get input on whether or not this method makes sense for removing test initialization boilerplate. e.g., is this flexible enough to cover most if not all testing pre-conditions.

juan-lunarg:
closing due to staleness. Also there has been some work on improving code sharing in testing code by @spencer-lunarg

PR #5822, opened 2023-05-09
layers: Fix dynamic GPL bug

spencer-lunarg:
Re-reading the spec, things like `VkPipelineMultisampleStateCreateInfo` dynamic state can be set in **either** fragment output OR fragment shader state

Added Positive tests as well


juan-lunarg:
> Re-reading the spec, things like `VkPipelineMultisampleStateCreateInfo` dynamic state can be set in **either** fragment output OR fragment shader state

Could you link to the part of the spec you read?

juan-lunarg (approved):

spencer-lunarg:
> Could you link to the part of the spec you read?

https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#pipelines-graphics-subsets-vertex-input

PR #5823, opened 2023-05-09
tests: Remove redundant renderpass in GPL tests

spencer-lunarg:
cc @ncesario-lunarg


ncesario-lunarg (commented):
LGTM! Still a lot of holes with GPL testing framework setup :(

Issue #5817, opened 2023-05-08
VK_KHR_deferred_host_operations - VK_OPERATION_NOT_DEFERRED_KHR crash


dorian-apanel-intel:
**Environment:**
 - OS: Windows
 - GPU: Intel A770
 - SDK or header version if building from repo: VVL: ea79898d1372bd9ca46f186d8d4febbe282d762e
 - Options enabled (synchronization, best practices, etc.): Standard checks (with handle wrapping)

**Describe the Issue**

App is calling:
1. vkCreateRayTracingPipelinesKHR (driver returns VK_OPERATION_NOT_DEFERRED_KHR)
2. vkDestroyDeferredOperationKHR
3. vkGetRayTracingShaderGroupHandlesKHR (with pipeline created in pt1.) - CRASH when unwrapping pipeline handle.

**Expected behavior**

It should not crash.

**Additional context**

I see no VK_OPERATION_NOT_DEFERRED_KHR handling in [layers\generated\layer_chassis_dispatch.cpp](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/43c4cd3504645bb9fd91383f45a867090264d993/layers/generated/layer_chassis_dispatch.cpp#L1456) in function `DispatchCreateRayTracingPipelinesKHR`

I have a commit that seems to workaround the issue (for this single function).

spencer-lunarg:
Was added in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/3793 @aitor-lunarg any idea (cc @arno-lunarg since I'm no RTX expert)

aitor-lunarg:
I had to dig a little bit to refresh. The reason behind the crash is that the handle is not wrapped, so when it "unwraps" it, it'll be null and therefore crash when we try to use that null handle.

I didn't correctly check if the operation was deferred or not for the object tracker part (thread safety is correctly handled) which leads to the post completion functions never being triggered because there's no call to `vkGetDeferredOperationResultKHR` (where the post completion functions are exectued for deferred operations, and therefore, the wrapping of the handle).

I couldn't find any CTS nor VVL test that cover this case, so that's probably one of the reasons this went unnoticed. I'll double check and open a CTS issue if I find none. All CTS tests did call `vkGetDeferredOperationResultKHR` even if the operation was not deferred. It would also be good to have some VVL test to be extra safe.

I'll provide a PR to fix the issue.

dorian-apanel-intel:
[...]

PR #5795, opened 2023-05-04
layers: Add condition to validate clear attachments

matkow01-arm:
Issue with `VkCmdClearAttachments` was spotted after changing VL version to 1.3.243 and higher.
This is very similar to issue #657.

While using `VkCmdClearAttachments` in secondary command buffer with null in framebuffer pointer at inheritance structure I see that application is crashing in `CoreChecks::PreCallValidateCmdClearAttachments`. VL are trying to operate on image view's that comes from framebuffer pointer which is NULL.

I propose to add additional conditions to if statement for covering that case (skip checking framebuffer related objects when framebuffer pointer is NULL).

PR #5825, opened 2023-05-09
Fix condition for deferred operations' clean up

aitor-lunarg:
Extra fixes pointed out in #5817 needed to correctly handle non deferred operations


juan-lunarg (approved):
Add a comment in the code you feel is most appropriate. If testing isn't being added the best way to prevent a regression is with comments.

spencer-lunarg (approved):
LGTM - With Juan's comments

ncesario-lunarg (approved):
LGTM, thanks @aitor-lunarg!

juan-lunarg (approved):
I added the comment and made the bool const consistently since I want to get this into the upcoming SDK.

aitor-lunarg:
@juan-lunarg Thank you for adding the comment to the change! I was unsure what type of comment we wanted to avoid future regressions (due to being a small change to fix to what the spec states). I'll take this as a reference for next time!

juan-lunarg:
> @juan-lunarg Thank you for adding the comment to the change! I was unsure what type of comment we wanted to avoid future regressions (due to being a small change to fix to what the spec states). I'll take this as a reference for next time!

No worries.

Just as an example in this recent PR from the arm folks (https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5795) I didn't ask for a comment on the validation layer code because they added a test.

IMO testing is the ultimate comment. However, if there is no corresponding test then a comment is the best way to go. Ideally a comment that just points to the original issue, since if someone does want to add testing they have context to go off of. And also potential contacts.

Issue #5233, opened 2023-02-07
broken dynamic render + GPL validation


zmike:
I'm seeing spurious validation errors for (at least):
* VUID-VkGraphicsPipelineCreateInfo-renderPass-06585
* VUID-VkGraphicsPipelineCreateInfo-renderPass-06589
* VUID-VkGraphicsPipelineCreateInfo-renderPass-06586
* VUID-VkGraphicsPipelineCreateInfo-renderPass-06581

when creating a graphics pipeline library with `VK_GRAPHICS_PIPELINE_LIBRARY_PRE_RASTERIZATION_SHADERS_BIT_EXT | VK_GRAPHICS_PIPELINE_LIBRARY_FRAGMENT_SHADER_BIT_EXT`. According to the spec, no member of `VkPipelineRenderingCreateInfo` should be read except `viewMask` in this scenario, therefore attempting to validate any formats is broken.

Using VVL 96f49d713eeb33cf87d8e56d1a8f3e05948002ed

zmike:
This has regressed, and VVL is once again reading fragment output members from the fragment shader stage.

zmike:
Still broken. I've uploaded a single frame trace that should exhibit all the issues: https://mega.nz/file/dRciBJrJ#QH73c9NHWJMn8nA1ySrENRSa9KEuElwg18B0yrhrDEM

Issue #5650, opened 2023-04-14
Spurious error when using graphics pipeline libraries and dynamic rendering


rg3igalia:
**Environment:**
 - OS: Linux
 - GPU: RTX 2060
 - SDK or header version if building from repo: 801211cb3ce6b37a76c7e14661644653e203295a
 - Options enabled (synchronization, best practices, etc.): defaults+synchronization

**Describe the Issue**

When using graphics pipeline libraries and dynamic rendering, the spec states that VkPipelineRenderingCreateInfo, which can be passed as part of the pNext chain in VkGraphicsPipelineCreateInfo, is only fully relevant for the fragment output state. Fragment shader state and pre-rasterization shader state can look at the viewMask member of the structure but nothing else.

An upcoming CTS test is trying to test this and hits `VUID-VkGraphicsPipelineCreateInfo-layout-01688` when creating the fragment shader state library, because the layers are [taking a look at the colorAttachmentCount member](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/801211cb3ce6b37a76c7e14661644653e203295a/layers/core_checks/shader_validation.cpp#L1021) in those cases as well, but it should be ignored because it may contain garbage.

**Expected behavior**

No validation errors.

**Valid Usage ID**
VUID-VkGraphicsPipelineCreateInfo-layout-01688

**Additional context**

<details>
<summary>code or terminal output</summary>

```sh
VUID-VkGraphicsPipelineCreateInfo-layout-01688(ERROR / SPEC): msgNum: 1143872976 - Validation Error: [ VUID-VkGraphicsPipelineCreateInfo-layout-01688 ] Object 0: handle = 0xe88693000000000c, type = VK_OBJECT_TYPE_SHADER_MODULE; | MessageID = 0x442e1dd0 | vkCreateGraphicsPipelines(): pCreateInfos[0] Shader Stage VK_SHADER_STAGE_FRAGMENT_BIT exceeds component limit VkPhysicalDeviceLimits::maxPerStageResources (128) The Vulkan spec states: The number of resources in layout accessible to each shader stage that is used by the pipeline must be less than or equal to VkPhysicalDeviceLimits::maxPerStageResources (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkGraphicsPipelineCreateInfo-layout-01688)
```

</details>

ncesario-lunarg:
[...]

PR #5829, opened 2023-05-09
layers: Ignore pNext structs in GPLs

ncesario-lunarg:
VkPipelineRenderingCreateInfo is part of fragment output state, and thus
should be ignored when a pipeline is being created without fragment
output state.


layers/generated/layer_chassis_dispatch.cpp
```diff
@@ -422,7 +423,14 @@ VkResult DispatchCreateGraphicsPipelines(VkDevice device, VkPipelineCache pipeli
                                                 dynamic_rendering->stencilAttachmentFormat != VK_FORMAT_UNDEFINED);
             }
 
-            local_pCreateInfos[idx0].initialize(&pCreateInfos[idx0], uses_color_attachment, uses_depthstencil_attachment);
+            auto& graphics_info = pCreateInfos[idx0];
+            auto state_info = dynamic_cast<ValidationStateTracker*>(layer_data);
```
ncesario-lunarg, line 427: Note that we already rely on RTTI (see `BUILD_DISPATCH_VECTOR` in chassis_dispatch_helper.h).

layers/state_tracker/cmd_buffer_state.cpp
```diff
@@ -250,10 +250,8 @@ void CMD_BUFFER_STATE::ResetPushConstantDataIfIncompatible(const PIPELINE_LAYOUT
         size_needed = std::max(size_needed, size);
 
         auto stage_flags = push_constant_range.stageFlags;
-        uint32_t bit_shift = 0;
```
ncesario-lunarg, line 253: I'm hitting this warning locally on main; I'm not sure why it's not getting hit in CI.

layers/state_tracker/pipeline_state.h
```diff
@@ -98,7 +98,9 @@ class PIPELINE_STATE : public BASE_NODE {
         template <typename CI>
         struct Traits {};
 
-        CreateInfo(const VkGraphicsPipelineCreateInfo *ci, std::shared_ptr<const RENDER_PASS_STATE> rpstate) : graphics() {
+        CreateInfo(const VkGraphicsPipelineCreateInfo *ci, std::shared_ptr<const RENDER_PASS_STATE> rpstate,
+                   const ValidationStateTracker *state_data)
```
spencer-lunarg, line 102: nit: The other helpers in `PIPELINE_STATE` use `const ValidationStateTracker& state_data`

layers/state_tracker/render_pass_state.h
```diff
@@ -77,6 +77,7 @@ class RENDER_PASS_STATE : public BASE_NODE {
     const bool use_dynamic_rendering;
     const bool use_dynamic_rendering_inherited;
     const bool has_multiview_enabled;
+    const bool rasterization_enabled{true};
```
spencer-lunarg, line 80: This will override `rasterization_enabled(rasterization_enabled)` in the constructor

layers/state_tracker/pipeline_state.h
```diff
@@ -513,6 +519,88 @@ class PIPELINE_STATE : public BASE_NODE {
...
+    }
+
+    // This is a helper that is meant to be used during safe_VkPipelineRenderingCreateInfo construction to determine whether or not
+    // certain fields should be ignored based on graphics pipeline state
+    static bool PnextRenderingInfoCustomCopy(const ValidationStateTracker *state_data,
+                                             const VkGraphicsPipelineCreateInfo &graphics_info, VkBaseOutStructure *safe_struct,
+                                             const VkBaseOutStructure *in_struct) {
+        // "safe_struct" is assumed to be non-null as it should be the "this" member of calling class instance
+        assert(safe_struct);
+        if (safe_struct->sType == VK_STRUCTURE_TYPE_PIPELINE_RENDERING_CREATE_INFO) {
```
spencer-lunarg, line 585: Is the idea `PNextCopyState` could have more than just `PnextRenderingInfoCustomCopy` and see we will chekc the `sType` in each helper Custom Copy function?

spencer-lunarg (approved):
LGTM - Few questions

layers/state_tracker/pipeline_state.h
ncesario-lunarg: :+1: yeah, I'll take this opportunity to update this.

layers/state_tracker/render_pass_state.h
ncesario-lunarg: I'm not sure I understand. This is to convey that the default value of this variable is true, and it will be set to true in constructors that do not set it. That's the intent at least, is that not happening?

juan-lunarg:
Will this PR close https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5233 and https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5650 ?

layers/state_tracker/cmd_buffer_state.cpp
juan-lunarg: Likely you are using a newer compiler than what CI is using.

layers/state_tracker/pipeline_state.h
ncesario-lunarg: Yes, the idea is that we may have other cases like this, and I didn't want to bake specific structs into `PNextCopyState`. e.g., I'm not sure if we need to ignore other structs that are only mentioned https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap10.html#pipelines-graphics-subsets-pre-rasterization in one particular sub-state. That said, I am very much open to changing this method for passing data down the pNext chain; I spun my wheels on this implementation way too much, and I'm not super happy with the end result, but I think it's better than some of the previous versions I had of it.

scripts/generators/helper_file_generator.py
```diff
@@ -447,16 +448,23 @@ def DeclareCounter(self, string_var, declare_flag):
...
-        safe_struct_helper_header += '#include <algorithm>\n'
-        safe_struct_helper_header += '\n'
-        safe_struct_helper_header += 'void *SafePnextCopy(const void *pNext);\n'
-        safe_struct_helper_header += 'void FreePnextChain(const void *pNext);\n'
-        safe_struct_helper_header += 'char *SafeStringCopy(const char *in_string);\n'
-        safe_struct_helper_header += '\n'
+        safe_struct_helper_header = '''
+#pragma once
+#include <vulkan/vulkan.h>
+#include <stdlib.h>
```
juan-lunarg, line 454: Use cstdlib instead of stdlib.h

```suggestion
#include <cstdlib>
```

layers/state_tracker/render_pass_state.h
juan-lunarg: I believe @ncesario-lunarg is correct. I don't think setting the default value overrides the constructor.

scripts/generators/helper_file_generator.py
```diff
@@ -1011,33 +1019,34 @@ def build_safe_struct_utility_funcs(self):
...
-        build_pnext_proc += '            break;\n'
-        build_pnext_proc += '        }\n'
+        build_pnext_proc = '''
+void *SafePnextCopy(const void *pNext, PNextCopyState* copy_state) {
+    if (!pNext) return nullptr;
+
+    void *safe_pNext{};
+    const VkBaseOutStructure *header = reinterpret_cast<const VkBaseOutStructure *>(pNext);
+
+    switch (header->sType) {
```
juan-lunarg, line 1029: We are adding extra overhead to all safe copies/initialization just for GPLs?

This doesn't seem like the best approach.

If GPLs safe structs are so specialized maybe they need their own explicit code generation. This current approach adds extra complexity / overhead to structs that don't need it.

I'm also not sure why we want the safe struct to have a different pnext chain from the original struct. To make the validation simpler?

juan-lunarg (changes requested):
This approach adds extra complexity the majority of safe_* structs don't need just for GPL.

scripts/generators/helper_file_generator.py
ncesario-lunarg: > We are adding extra overhead to all safe copies/initialization just for GPLs?

Yes, though there are other potential places this could be used. e.g., this could replace all of the "manual construction" we currently specify in the code generation script, which extends beyond GPLs and is potentially useful in the future.

> This doesn't seem like the best approach.

I'm very happy to try other approaches, but I didn't see a "clean" way of doing this with our current infrastructure. `SafePNextCopy` is a recursive call, and we need state passed down the chain; having "ad-hoc" `SafePNextCopy`s for individula structs is also an added complication. If performance is an issue, then that's something we can explore, but I'd rather not if possible.

> If GPLs safe structs are so specialized maybe they need their own explicit code generation. This current approach adds extra complexity / overhead to structs that don't need it.

I would rather not wait for a solution such as this if performance is not affected. This fixes a crash and is required based on the spec (see second commit message), but--as stated earlier--this method _could_ be used to remove manual construction that is currently specified in the scripts (which would also be future work if done at all). Also, I assume separate code generation would imply different copy methods for individual structs, which, also as stated before, seems like unnecessary complexity.

> I'm also not sure why we want the safe struct to have a different pnext chain from the original struct. To make the validation simpler?

No, the spec requires that we must ignore certain fields of certain pNext structs based on graphics pipeline state. See, e.g., the comment about `viewMask` in https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap10.html#pipelines-graphics-subsets-pre-rasterization. I'm happy to explore other suggestions anyone has to approach this differently; there is probably a much nicer way to get to a solution here that I didn't see.

juan-lunarg: Gotcha I think I see the tradeoffs and understand the current approach.

juan-lunarg (changes requested):
Add a test in `other.cpp` directly testing the new functionality added to safe structs.

See for an example of the safe structs being tested. `VkPositiveLayerTest.SafeVoidPointerCopies`

It may be prudent to have a test file for generated code.

The test should also highlight why the new functionality is needed.

layers/state_tracker/render_pass_state.h
spencer-lunarg: ok, my C++ judo is always mixed, maybe it was

`const bool rasterization_enabled = true;`

or

`bool rasterization_enabled{true};`

but I recently had a bug because I thought I was setting it in the constructor and then it was re-initializing it here in the header

juan-lunarg: > If a non-static data member has a default member initializer  and also appears in a member initializer list, then the member  initializer is used and the default member initializer is ignored (since C++11):

```c++
struct S
{
    int n = 42;   // default member initializer
    S() : n(7) {} // will set n to 7, not 42
};
```

https://en.cppreference.com/w/cpp/language/constructor

tests/positive/other.cpp
```diff
@@ -1069,3 +1069,76 @@ TEST_F(VkPositiveLayerTest, ExtensionsInCreateInstance) {
 
     ASSERT_NO_FATAL_FAILURE(InitFramework(m_errorMonitor));
 }
+
+TEST_F(VkPositiveLayerTest, CustomSafePNextCopy) {
+    TEST_DESCRIPTION("Check passing custom data down the pNext chain for safe struct construction");
```
juan-lunarg, line 1074: Complete nitpick. Please add a brief explanation of why this functionality was added to safe structs.

juan-lunarg (approved):

juan-lunarg (changes requested):
`VkPositiveLayerTest.CustomSafePNextCopy` fails ASAN

ncesario-lunarg:
> `VkPositiveLayerTest.CustomSafePNextCopy` fails ASAN

Yeah, scoping mistake on my part. Should be fixed now.

juan-lunarg (approved):

ncesario-lunarg:
I want to get this through the perf suite before merging, but having some difficulties there with our CI machines. I will hopefully merge this once that's figured out.

layers/state_tracker/cmd_buffer_state.cpp
```diff
@@ -250,10 +250,8 @@ void CMD_BUFFER_STATE::ResetPushConstantDataIfIncompatible(const PIPELINE_LAYOUT
...
         while (stage_flags) {
```
artem-lunarg, line 253: nit: since this code is touched,  it should be possible to remove `stage_flags` and the `while` loop which sets the flags to zero.

ncesario-lunarg: :+1: good point. I just removed this since my compiler complained about it.

PR #5830, opened 2023-05-10
scripts: Fix tests listed in vk_validation_stats

spencer-lunarg:
Currently the tests were not being listed correctly, now it looks like

image

cc @mikes-lunarg

### DIFF
--- a/scripts/vk_validation_stats.py
+++ b/scripts/vk_validation_stats.py
@@ -240,12 +240,10 @@ class ValidationSource:
 
 # Class to parse the validation layer test source and store testnames
 class ValidationTests:
-    def __init__(self, test_file_list, unassigned_vuid_files, test_group_name=['VkLayerTest', 'VkPositiveLayerTest', 'VkBestPracticesLayerTest']):
+    def __init__(self, test_file_list, unassigned_vuid_files):
         self.test_files = test_file_list
         self.unassigned_vuid_files = unassigned_vuid_files
-        self.test_trigger_txt_list = []
-        for tg in test_group_name:
-            self.test_trigger_txt_list.append('TEST_F(%s' % tg)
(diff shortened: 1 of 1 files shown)


scripts/vk_validation_stats.py
```diff
@@ -290,6 +288,8 @@ def parse(self, spirv_val):
                         if ('' == testname):
                             grab_next_line = True
                             continue
+                        testgroup = line.split(',')[0][line.index('(') + 1:]
+                        testname = testgroup + '.' + testname
```
juan-lunarg, line 292: Just as a thought. We were looking for a way to enforce good test names. IE not having `Test` in the name.

Perhaps we could do that in this script? IE error out on CI that you have a bad test name?

juan-lunarg (approved):
LGTM with @mikes-lunarg's approval

scripts/vk_validation_stats.py
```diff
@@ -290,6 +288,8 @@ def parse(self, spirv_val):
                         if ('' == testname):
                             grab_next_line = True
                             continue
+                        testgroup = line.split(',')[0][line.index('(') + 1:]
```
mikes-lunarg, line 291: I would rather read one regex than a pile of splits and slices, but 1) I'm not sure that feeling is universal and 2) You are just matching the existing code style, so I get it.

mikes-lunarg (approved):
LGTM!

scripts/vk_validation_stats.py
spencer-lunarg: To be fair, it is not hard to just write another script that does

```python
for line in file:
    if line has 'TEST_F':
        validate_name(line)
```

in a separate file... this is still on my radar to do soon

PR #5832, opened 2023-05-10
tests: Remove remaining explicit GetProcAddr

spencer-lunarg:
mainly removes the `GetInstanceProcAddr` that were not removed, also a few extra `GetDeviceProcAddr`

### DIFF
--- a/tests/negative/ray_tracing_pipeline.cpp
+++ b/tests/negative/ray_tracing_pipeline.cpp
@@ -38,8 +38,6 @@ TEST_F(NegativeRayTracing, BasicUsage) {
     VkShaderObj miss_shader(this, bindStateRTNVShaderText, VK_SHADER_STAGE_MISS_BIT_KHR);
     VkShaderObj intr_shader(this, bindStateRTNVShaderText, VK_SHADER_STAGE_INTERSECTION_BIT_KHR);
     VkShaderObj call_shader(this, bindStateRTNVShaderText, VK_SHADER_STAGE_CALLABLE_BIT_KHR);
-    const auto vkCreateRayTracingPipelinesKHR =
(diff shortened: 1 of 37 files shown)


tests/negative/ray_tracing_pipeline.cpp
```diff
@@ -1269,17 +1230,14 @@ TEST_F(NegativeRayTracing, LibraryGroupHandlesEXT) {
     handle_buffer.resize(ray_tracing_properties.shaderGroupHandleSize);
 
     m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-vkGetRayTracingShaderGroupHandlesKHR-pipeline-07828");
-    vkGetRayTracingShaderGroupHandlesKHR(m_device->handle(), rt_pipe.pipeline_, 0, 1, ray_tracing_properties.shaderGroupHandleSize,
-                                         handle_buffer.data());
+    vk::GetRayTracingShaderGroupHandlesKHR(m_device->handle(), rt_pipe.pipeline_, 0, 1,
```
spencer-lunarg, line 1233: @artem-lunarg seems there is a bug in the code to grab this function pointers is in `VK_NV_ray_tracing`  and not in `VK_KHR_ray_tracing_pipeline` and it is crashing here

spencer-lunarg: only one that seems broken, so going to just revert this one

artem-lunarg: I'll check it.

While we are here, I'll mention about another potential issue, did not investigate yet. My impression that previously we could use mockidc driver either by loading it through regular loader, or just renaming it to vulkan-1.dll and replacing our loader. I had one crash that suggest that the latter case does not work anymore, one of the global instance functions is null. Will create a task to fix all these.

artem-lunarg (approved):

PR #5835, opened 2023-05-10
layers: DescriptorSet state cleanup

jeremyg-lunarg:
Remove some unnecessary method arguments and convert pointers into references.


ncesario-lunarg (approved):
Nice cleanup!

PR #5836, opened 2023-05-10
cmake: Fix invalid CMAKE_GENERATOR_PLATFORM warning 

juan-lunarg:
Noticed this while building on Windows.


scripts/CMakeLists.txt
```diff
@@ -27,16 +27,15 @@ if (UPDATE_DEPS)
     list(APPEND update_dep_command "--generator")
     list(APPEND update_dep_command "${CMAKE_GENERATOR}")
 
+    if (${CMAKE_VS_VERSION_BUILD_NUMBER} VERSION_LESS "16")
+        message(FATAL_ERROR "Visual Studio 16 2019 is the oldest visual studio generator we support.")
+    endif()
```
juan-lunarg, line 32: Per the `BUILD.md` we only support 2019 and higher.

scripts/CMakeLists.txt
```diff
@@ -27,16 +27,15 @@ if (UPDATE_DEPS)
...
+
     if (CMAKE_GENERATOR_PLATFORM)
         set(_target_arch ${CMAKE_GENERATOR_PLATFORM})
-    else()
-        if (MSVC_IDE)
-            message(WARNING "CMAKE_GENERATOR_PLATFORM not set. Using x64 as target architecture.")
-        endif()
-        set(_target_arch x64)
```
juan-lunarg, line 36: In the past this used to be a good warning since it defaulted to Win32. Now the default matches your host machine. IE x64 for most users.

tests/CMakeLists.txt
```diff
@@ -168,9 +168,7 @@ elseif(MSVC)
     # Disable some signed/unsigned mismatch warnings.
     target_compile_options(vk_layer_validation_tests PRIVATE /wd4267)
 
-    if(MSVC_IDE)
```
juan-lunarg, line 171: The MSVC_IDE check isn't needed. CMake will just ignore this property other generators.

spencer-lunarg (approved):

scripts/CMakeLists.txt
```diff
@@ -28,15 +28,13 @@ if (UPDATE_DEPS)
     list(APPEND update_dep_command "${CMAKE_GENERATOR}")
 
     if (CMAKE_GENERATOR_PLATFORM)
-        set(_target_arch ${CMAKE_GENERATOR_PLATFORM})
-    else()
-        if (MSVC_IDE)
-            message(WARNING "CMAKE_GENERATOR_PLATFORM not set. Using x64 as target architecture.")
+        if (${CMAKE_VS_VERSION_BUILD_NUMBER} VERSION_LESS "16")
+            message(FATAL_ERROR "Visual Studio 16 2019 is the oldest visual studio generator we support.")
```
arno-lunarg, line 32: Is the `FATAL_ERROR` necessary? What about just issuing a `WARNING` and eventually let cmake fail?

arno-lunarg (approved):
LGTM! One question

scripts/CMakeLists.txt
juan-lunarg: 2 main reasons. We don't test with anything earlier than 2019. People ignore warnings.

PR #5837, opened 2023-05-10
bp: Remove best practice warnings generating false positives

### LINKED ISSUE #4730: Incorrect best practice warning when using `vkGetDevice{Buffer/Image}MemoryRequirements`
**Describe the Issue**

Validation layers report `UNASSIGNED-BestPractices-vkBindBufferMemory-requirements-not-retrieved` and `UNASSIGNED-BestPractices-vkBindImageMemory-requirements-not-retrieved` if you bind device memory with memory requirements queried using `vkGetDeviceBufferMemoryRequirements` or `vkGetDeviceImageMemoryRequirements`. These functions are provided by Vulkan 1.3 as alternatives to `vkGetBufferMemoryRequirements` and `vkGetImageMemoryRequirements`.

**Valid Usage ID**
`
UNASSIGNED-BestPractices-vkBindImageMemory-requirements-not-retrieved(WARN / SPEC): msgNum: 1185611228 - Validation Warning: [ UNASSIGNED-BestPractices-vkBindImageMemory-requirements-not-retrieved ] Object 0: handle = 0x1ed20d4d200, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x46aafddc | vkBindImageMemory(): Binding memory to VkImage 0x1ed23c955c0[] but vkGetImageMemoryRequirements() has not been called on that image.
`

juan-lunarg:
closes #4730

Currently I'd just like to remove this false positive before our SDK release.

Here is my rational.

1.) It generates false positives. See issue for details.
2.) As discussed in the issue this way of checking for mem requirements is questionable. Just checking you called the function isn't very robust.
3.) It doesn't have any testing

My current plan of action is remove this invalid logging for the upcoming SDK release, then implement essentially the same thing except more robustly based on the advice here:

> Yeah it may end up being better to check the inputs for the memory binding function and emitting a warning based on that than to track which functions were called where + associated data. Looking for specific function calls probably really isn't that useful.

Thoughts?

### DIFF
--- a/layers/best_practices/bp_device_memory.cpp
+++ b/layers/best_practices/bp_device_memory.cpp
@@ -147,13 +147,6 @@ bool BestPractices::PreCallValidateFreeMemory(VkDevice device, VkDeviceMemory me
 bool BestPractices::ValidateBindBufferMemory(VkBuffer buffer, VkDeviceMemory memory, const char* api_name) const {
     bool skip = false;
     auto buffer_state = Get<BUFFER_STATE>(buffer);
-
-    if (!buffer_state->memory_requirements_checked && !buffer_state->external_memory_handle_types) {
-        skip |= LogWarning(device, kVUID_BestPractices_BufferMemReqNotCalled,
-                           "%s: Binding memory to %s but vkGetBufferMemoryRequirements() has not been called on that buffer.",
-                           api_name, report_data->FormatHandle(buffer).c_str());
-    }
-
     auto mem_state = Get<DEVICE_MEMORY_STATE>(memory);
 
     if (mem_state && mem_state->alloc_info.allocationSize == buffer_state->createInfo.size &&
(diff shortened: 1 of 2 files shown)


layers/best_practices/bp_device_memory.cpp
```diff
@@ -147,13 +147,6 @@ bool BestPractices::PreCallValidateFreeMemory(VkDevice device, VkDeviceMemory me
 bool BestPractices::ValidateBindBufferMemory(VkBuffer buffer, VkDeviceMemory memory, const char* api_name) const {
     bool skip = false;
     auto buffer_state = Get<BUFFER_STATE>(buffer);
-
-    if (!buffer_state->memory_requirements_checked && !buffer_state->external_memory_handle_types) {
```
juan-lunarg, line 151: Note if we are fine with my proposed plan of action I'll remove the rest of the dangling code. EX: `memory_requirements_checked` is only used for this 1 best practice check.

artem-lunarg: I have vague memory that this (or similar) warning was reported like 5 years ago... but it was part of the core. I like the idea to remove checks that we can't validate reliably, the less false-positives the better chances that BP will be used during development.

juan-lunarg: It did use to be part of standard validation based on the git history. It got moved to bp.

artem-lunarg (approved):

spencer-lunarg (approved):
> Thoughts?

This is great, agree :100:

arno-lunarg (approved):

layers/best_practices/bp_device_memory.cpp
juan-lunarg: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5839

PR #5843, opened 2023-05-12
layers: Cleanup error msg in ValidateBeginCmdBuff

arno-lunarg:
I was adding validation in this function in a previous PR, and saw that some error messages could be cleaned up.
Feel free to make additional suggestions!


layers/core_checks/cc_cmd_buffer.cpp
```diff
@@ -158,28 +158,27 @@ bool CoreChecks::PreCallValidateBeginCommandBuffer(VkCommandBuffer commandBuffer
                 if (p_attachment_sample_count_info_amd &&
                     p_attachment_sample_count_info_amd->colorAttachmentCount != p_inherited_rendering_info->colorAttachmentCount) {
                     skip |= LogError(commandBuffer, "VUID-VkCommandBufferBeginInfo-flags-06003",
-                                     "vkBeginCommandBuffer(): VkAttachmentSampleCountInfo{AMD,NV}->colorAttachmentCount[%u] must "
-                                     "equal VkCommandBufferInheritanceRenderingInfoKHR->colorAttachmentCount[%u]",
+                                     "vkBeginCommandBuffer(): colorAttachmentCount (%" PRIu32
```
artem-lunarg, line 161: nit: this string might be formatted as two lines

layers/core_checks/cc_cmd_buffer.cpp
```diff
@@ -158,28 +158,27 @@ bool CoreChecks::PreCallValidateBeginCommandBuffer(VkCommandBuffer commandBuffer
...
                     (p_inherited_rendering_info->rasterizationSamples & AllVkSampleCountFlagBits) == 0) {
-                    skip |=
-                        LogError(commandBuffer, "VUID-VkCommandBufferInheritanceRenderingInfo-colorAttachmentCount-06004",
-                                 "vkBeginCommandBuffer(): VkCommandBufferInheritanceRenderingInfo->colorAttachmentCount (%" PRIu32
-                                 ") is not 0, rasterizationSamples (%s) must be valid VkSampleCountFlagBits value",
-                                 p_inherited_rendering_info->colorAttachmentCount,
-                                 string_VkSampleCountFlagBits(p_inherited_rendering_info->rasterizationSamples));
+                    skip |= LogError(
+                        commandBuffer, "VUID-VkCommandBufferInheritanceRenderingInfo-colorAttachmentCount-06004",
+                        "vkBeginCommandBuffer(): colorAttachmentCount (%" PRIu32 ") is not 0, so rasterizationSamples (%" PRIx32
```
artem-lunarg, line 172: Why PRIx32 instead of  `string_VkSampleCountFlagBits` ?

arno-lunarg: In this case `p_inherited_rendering_info->rasterizationSamples` will have an undefined value, so trying to print a string out of it will just result in something like `un handled value`. So I think printing the hex value is more informative.
Actually I think this something we should always do when trying to print an undefined value: have something like `Unhandled enum value 0xFA`

ncesario-lunarg (approved):
Nice cleanup!

PR #4246, opened 2022-06-20
layers: Relax sync VUs in 1.3.218

ncesario-lunarg:
Rename 00867 to 06809 and 03087 to 06810. This also removes the ordering
restrictions on these VUs.

Closes #4236.

PR #5848, opened 2023-05-12
layers: Remove global stage ordering functions

artem-lunarg:
These functions can produce unexpected result. The problem is that `GetLogicallyEarLiest(Latest)GraphicsPipelineStage` tries to create global ordering of __all__ pipeline stages (not only graphics as the name suggests). So, if the mask specifies `fragment+raytracing` stages then the earliest stage will always be evaluated to a `fragment` stage (because it has lower index), independenty if the application logic is concerned with the fragment or with the raytracing stage.

Also added a comment here:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/4246#discussion_r1190329610

This PR is a side effect from a parallel work to update current syncval codegen to use xml input instead of `validusage.json` and adhoc text snippets.


ncesario-lunarg (approved):
Nice!

spencer-lunarg (approved):

arno-lunarg (approved):

PR #5850, opened 2023-05-12
layers: Replace shader accesses codegen with cpp code

artem-lunarg:
Python codegen for this function does not use external data to drive code generation. It's not much benefits to push this through code generation pipeline.

This can be moved back if we need more complex mapping that requires data driven approach.

This PR is a side effect from a parallel work to update current syncval codegen to use xml input instead of validusage.json and adhoc text snippets.


spencer-lunarg (approved):
Agree with this approach, can also data-drive it from the scripts if needed in the future

ncesario-lunarg (approved):

PR #5851, opened 2023-05-13
Fix imageOpaque sparse binding crash

nyorain:
The sparse-binding check has a copy-paste issue, it uses `bind_info.bufferBindCount` when iterating over the opaque image binds, causing it to possibly read-out-of-bounds or access garbage pointers. I had a crash in the validation layers during a submission of sparse-buffer-binds where `pImageOpaqueBinds` was set to garbage since `imageOpaqueBindCount` was 0 (in which case the pointer is ignored per spec).


spencer-lunarg (approved):
Thanks for the fix

1. The CI broken due to one our dependency, will be fixed and reran for you soon
2. This should definitely have a regression test, happy to write one

lunarpapillo:
FYI, the LunarG CI is having a machine configuration problem.  I'll restart this test once we get it resolved.

PR #5854, opened 2023-05-15
scripts: Create a BaseGeneratorOptions object and cleanup

spencer-lunarg:
As someone who has personally added a few of these generator files, I know instead of looking up the `Vulkan-Headers` generator options, these are just copied and pasted. The real issue is there is only so much that can be done globally because each file does complete different things

After removing all the unused values, I realized a `BaseGeneratorOptions` would be easy to have because most times we just want to to have everything so we can sort it out as we need.

Lastly I assume no one is actually using all these args at command line as these are not tested and if someone actually needs to make changes for developing, they can just do it in the Python itself


juan-lunarg (approved):

ncesario-lunarg (approved):
Nice!

PR #5856, opened 2023-05-15
layers: Relabel maintenance4 VUID

spencer-lunarg:
add `VUID-RuntimeSpirv-OpTypeVector-06816` VUID

Add a negative test also (`PositiveShaderInterface.RelaxedTypeMatch` is already testing this in for a Positive test)

### DIFF
--- a/layers/core_checks/cc_shader.cpp
+++ b/layers/core_checks/cc_shader.cpp
@@ -2883,14 +2883,18 @@ bool CoreChecks::ValidateInterfaceBetweenStages(const SHADER_MODULE_STATE &produ
                     const uint32_t input_vec_size = input_var->base_type.Word(3);
                     if (output_vec_size > input_vec_size) {
                         const LogObjectList objlist(producer.vk_shader_module(), consumer.vk_shader_module());
+                        const char *vuid = IsExtEnabled(device_extensions.vk_khr_maintenance4)
+                                               ? "VUID-RuntimeSpirv-maintenance4-06817"
+                                               : "VUID-RuntimeSpirv-OpTypeVector-06816";
                         skip |=
-                            LogError(objlist, "VUID-RuntimeSpirv-maintenance4-06817",
+                            LogError(objlist, vuid,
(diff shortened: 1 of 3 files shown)


layers/core_checks/cc_shader.cpp
```diff
@@ -2883,14 +2883,18 @@ bool CoreChecks::ValidateInterfaceBetweenStages(const SHADER_MODULE_STATE &produ
...
+                                               : "VUID-RuntimeSpirv-OpTypeVector-06816";
                         skip |=
-                            LogError(objlist, "VUID-RuntimeSpirv-maintenance4-06817",
+                            LogError(objlist, vuid,
                                      "vkCreateGraphicsPipelines(): pCreateInfos[%" PRIu32 "] starting at Location %" PRIu32
                                      " Component %" PRIu32 " the Output (%s) has a Vec%" PRIu32 " while Input (%s) as a Vec%" PRIu32
-                                     " Enable VK_KHR_maintenance4 device extension to allow relaxed interface matching "
-                                     "between input and output vectors",
+                                     ". Enable VK_KHR_maintenance4 device extension to allow relaxed interface matching "
+                                     "between input and output vectors.",
```
juan-lunarg, line 2894: This last half of the error message doesn't make sense if `IsExtEnabled(device_extensions.vk_khr_maintenance4)` is true.

juan-lunarg (changes requested):

arno-lunarg (approved):

layers/core_checks/cc_shader.cpp
spencer-lunarg: It is still a valid message, honestly these "If VUID" things are kinda nasty

You will ONLY get the first VUID if you enabled the extension, but not the feature bit.

I think adding information that a `maintenance` extension is around that makes a VU valid is helpful for a developer

ncesario-lunarg (approved):

juan-lunarg (approved):

PR #5859, opened 2023-05-16
layers: Prevent crash on Core_DrawState_NoActiveRenderpass

kearwood:
cb_state.activeRenderPass will always be null when logging kVUID_Core_DrawState_NoActiveRenderpass.  Layers would always crash due to dereferencing cb_state.activeRenderPass for inclusion in the LogObjectList.


spencer-lunarg (approved):
looking, we had no test for `NoActiveRenderpass`... but this should have a proper VU in the spec

Will get this in and take action to both get a proper VU and test for this

Thanks for the PR!

PR #5860, opened 2023-05-16
layers: Parameterization changes for Vulkan SC

aqnuep:


layers/CMakeLists.txt
```diff
@@ -14,30 +15,34 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ~~~
+
+set(API_TYPE "vulkan")
+set(LAYER_NAME "VkLayer_khronos_validation")
+
```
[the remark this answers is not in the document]
juan-lunarg: Perhaps something like this:

```suggestion

# Variables needed for SC
set(API_TYPE "vulkan")
set(LAYER_NAME "VkLayer_khronos_validation")

```

layers/CMakeLists.txt
```diff
@@ -154,14 +159,14 @@ if (VVL_CODEGEN)
...
             ${VULKAN_HEADERS_REGISTRY_DIRECTORY} "${SPIRV_HEADERS_INSTALL_DIR}/include/spirv/unified1"
-            --incremental --generated-version ${VulkanHeaders_VERSION}
-        WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/generated
+            --incremental --generated-version ${VulkanHeaders_VERSION} --api ${API_TYPE}
+        WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/${API_TYPE}/generated
     )
 endif()
 
-add_library(VkLayer_khronos_validation MODULE)
+add_library(vvl MODULE)
```
juan-lunarg, line 167: 🥇

layers/CMakeLists.txt
```diff
@@ -311,30 +316,32 @@ target_sources(VkLayer_khronos_validation PRIVATE
...
+    target_compile_options(vvl PRIVATE -Wa,-mbig-obj)
 elseif(APPLE)
     message(STATUS "Functions are exported via VVL_EXPORT")
-    set_target_properties(VkLayer_khronos_validation PROPERTIES SUFFIX ".dylib")
+    set_target_properties(vvl PROPERTIES SUFFIX ".dylib")
 elseif(ANDROID)
     message(STATUS "Functions are exported via VVL_EXPORT")
 else()
-    target_link_options(VkLayer_khronos_validation PRIVATE LINKER:--version-script=${CMAKE_CURRENT_SOURCE_DIR}/libVkLayer_khronos_validation.map,-Bsymbolic,--exclude-libs,ALL)
+    target_link_options(vvl PRIVATE LINKER:--version-script=${CMAKE_CURRENT_SOURCE_DIR}/lib${LAYER_NAME}.map,-Bsymbolic,--exclude-libs,ALL)
```
juan-lunarg, line 336: Rename `libVkLayer_khronos_validation.map` to `VkLayer_khronos_validation.map`.

layers/vulkan/generated/lvt_function_pointers.cpp
```diff
@@ -901,25 +901,25 @@ PFN_vkGetDescriptorSetLayoutHostMappingInfoVALVE GetDescriptorSetLayoutHostMappi
 PFN_vkGetDescriptorSetHostMappingVALVE GetDescriptorSetHostMappingVALVE;
 
 
-void InitCore() {
+void InitCore(const char *api_name) {
```
juan-lunarg, line 904: Does this need to be passed as a parameter? Couldn't this just be a constant in the function?

layers/vulkan/generated/lvt_function_pointers.cpp
```diff
@@ -50,7 +50,7 @@ static void *get_proc_address(dl_handle library, const char *name) {
     assert(name);
     return (void *)GetProcAddress(library, name);
 }
-#elif defined(__linux__) || defined(__APPLE__) || defined(__FreeBSD__) || defined(__OpenBSD__)
+#elif defined(__linux__) || defined(__APPLE__) || defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__QNX__)
```
juan-lunarg, line 53: `QNX`? Is this something affecting SC?

layers/CMakeLists.txt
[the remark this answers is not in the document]
aqnuep: I've added some comments about the variables, although slightly more detailed ones than the one proposed here.

layers/CMakeLists.txt
aqnuep: Done.

layers/vulkan/generated/lvt_function_pointers.cpp
aqnuep: It could, for now. However, this seemed like a more flexible approach, especially if we consider that in the future we might want to run the validation layer tests in a similar two-pass fashion for Vulkan SC, as it's done in the CTS, where first the JSON files used as input for the pipeline cache compiler are collected using a regular Vulkan driver, and then the built pipeline caches are used when running the tests again against a Vulkan SC driver, which requires changing the used API at run-time, hence the parameter.

layers/vulkan/generated/lvt_function_pointers.cpp
aqnuep: Yes, Vulkan SC is used on QNX. This change could have been left to leave in the downstream repository, but as it shouldn't have an effect on VVL, I thought that there should be no harm in upstreaming this small change.

juan-lunarg (approved):
LGTM just needs to be rebased on main to address merge conflict

BUILD.gn
```diff
@@ -101,7 +101,7 @@ config("vulkan_layer_config") {
   include_dirs = [
     "layers",
     "layers/external",
-    "layers/generated",
```
spencer-lunarg, line 104: When I see this, I am expecting there to be a 

`layers/vulkan/sync` / `layers/vulkansc/sync` too?

Why not have this be `layers/generated/vulkan`?

layers/vulkan/generated/vk_extension_helper.h
```diff
@@ -52,6 +52,48 @@ enum ExtEnabled : unsigned char {
 #define VK_VERSION_1_2_NAME "VK_VERSION_1_2"
 #define VK_VERSION_1_3_NAME "VK_VERSION_1_3"
 
+
+#define VVL_UNRECOGNIZED_API_VERSION 0xFFFFFFFF
+
+class APIVersion {
```
spencer-lunarg, line 58: :+1: 

Could argue this could just become another file like `vk_api_version.h` but happy to do that later, for now really like this util

layers/VkLayer_khronos_validation.map
ncesario-lunarg, line 1: Makes sense, though I am curious if this affected SC in some way?

ncesario-lunarg (approved):
LGTM!

I agree that `layers/generated/(vulkan|vulkansc)` seems a little bit cleaner, but I don't have a strong opinion on this.

BUILD.gn
aqnuep: I'm not sure I understand the question.

The generated directory have been moved from `layers/generated` to `layers/vulkan/generated`, hence the include path needs to be `layers/vulkan`.

This build config file was already incorrect, because at the time you've changed all includes to refer to the generated folder explicitly (`#include "generated/..."`) that `layers/generated` include directory should have been removed in the first place. I just fixed that per our new directory structure.

layers/vulkan/generated/vk_extension_helper.h
aqnuep: Yes, this was discussed internally. We decided that it's still best to keep this class generated, as we expect some parts of it being dependent on definitions. Now on whether the generated file itself should be separated out, good question. I can do it, if needed, or you can do it later if you'd like to.

layers/VkLayer_khronos_validation.map
aqnuep: That could work too, but then you'd have to remove all the hard-coded `#include "generated/..."` paths in the source files, as I originally suggested, but that idea was rejected.

aqnuep:
> I agree that `layers/generated/(vulkan|vulkansc)` seems a little bit cleaner, but I don't have a strong opinion on this.

That could work, but that would require removing the hard-coded paths `#include "generated/..."` in the source files as I originally suggested, but that idea was rejected before.

aqnuep:
> LGTM just needs to be rebased on main to address merge conflict

Yeah, unfortunately @spencer-lunarg's change to refactor the generator script had significant impact on this PR, as it removed the possibility to parameterize the generator options needed for handling different API types.

He did give me a heads-up about that change, but I didn't expect that to be merged before this one, so I'll probably need a day to add back the removed capabilities. Stay tuned...

layers/vulkan/generated/vk_extension_helper.h
spencer-lunarg: Feel free to separate it out, this PR or future PR

BUILD.gn
spencer-lunarg: I guess my concern is someone new looking at the source will probably be confused why they would need to look in `layers/vulkan/` or `layers/vulkansc/` for the generated code

I realize the other around this is you would have to go `#include "generated/vulkan/` for everything... I guess is the idea that the VulkanSC are going to have different interface, or will using either `vulkan/generated` or `vulkansc/generated` file build?

spencer-lunarg:
> He did give me a heads-up about that change,

That was for https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5866 because that is a really invasive change

https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5854 is not doing that much but removing redundant code

The way to easily solve it is how I did

```python
def SetOutputDirectory(directory):
    global outputDirectory
    outputDirectory = directory
    
    # ...
    GeneratorOptions.__init__(self,
        directory = outputDirectory,
```

This allows you at `lvl_genvk.py` to set the value globally (in this case `mergeApiNames` and `apiname`) and have it just go to every generator

I found no use case where you would want to have half the scripts use different `mergeApiNames`... you set them all at once, this reduces the copy and pasting between EVERY file in `lvl_genvk.py`

BUILD.gn
spencer-lunarg: Honestly, for now it is fine, rather do this now then go around change every `include` also in the PR

aqnuep: > I realize the other around this is you would have to go #include "generated/vulkan/ for everything... I guess is the idea that the VulkanSC are going to have different interface, or will using either vulkan/generated or vulkansc/generated file build?

Yeah, that's not an option, because it would hard-code the API into the source files, so we couldn't switch out the generated files with the Vulkan SC ones without having to modify every source file.

The includes are using `layers/vulkan` vs `layers/vulkansc` per the current design as, once again, the `generated/...` part of the path was chosen to be hard-coded in the source files recently. No build file should use the `generated` directory itself as an include directory per the new policy, it's just that apparently some build files weren't updated.

aqnuep:
> This allows you at `lvl_genvk.py` to set the value globally (in this case `mergeApiNames` and `apiname`) and have it just go to every generator
> 
> I found no use case where you would want to have half the scripts use different `mergeApiNames`... you set them all at once, this reduces the copy and pasting between EVERY file in `lvl_genvk.py`

That is not the case though as while some generated files would need to use `mergeApiNames`, because they should include both Vulkan and Vulkan SC definitions (like `vk_safe_struct.h`), while others need to be specialized to only include the specific API, hence no `mergeApiNames` (like `parameter_validation.cpp`). So I'll have to re-add those parameters to make things work.

spencer-lunarg:
> That is not the case though as while some generated files would need to use mergeApiNames, because they should include both Vulkan and Vulkan SC definitions

You are still able to pass things one-by-one such as we do for `valid_usage_path`

```python
    genOpts['parameter_validation.h'] = [
          ParameterValidationOutputGenerator,
          BaseGeneratorOptions(
            filename          = 'parameter_validation.h',
            valid_usage_path  = args.scripts)
          ]
```

scripts/lvl_genvk.py
```diff
@@ -298,6 +370,7 @@ def makeGenOpts(args):
           SpirvGrammarHelperOutputGenerator,
           BaseGeneratorOptions(
             filename          = 'spirv_grammar_helper.cpp',
+            apiname           = apiname,
```
spencer-lunarg, line 373: so `spirv_grammar_helper` doesn't use the Vulkan API at all, just the SPIR-V grammar file

So curious if maybe it would be better to set `apiname` globally like `dictionary` then? Rather not have setting here that aren't actually being used

spencer-lunarg: For the record, I am not asking for a change here, it is still valid and good to merge as is

>  It does not mean that it won't ever use it.

The purpose of that file is purely to **only** be SPIR-V while the `spirv_validation_helper` is the cross Vulkan/SPIR-V helper

>  I think there should be no harm in specifying all parameters that may affect the code gen.

I would argue that is like saying we should keep all `#include` headers around incase things might need it... I would like people to know "why" they are adding these generator params, otherwise we just ended up with years of copy-and-pasting things around blindly (I personally did that the first generation file I added)

aqnuep: No, I get what you mean and I understand that this particular generator will not use the APIs defined in `vk.xml`.

But technically `BaseGeneratorOptions` is passing down an `apiname` to its base class `GeneratorOptions`, hence we shouldn't be passing down the incorrect default API name just because we know that the particular generator's behavior won't be affected by that generator option.

In case of the unlikely event that the `GeneratorOptions` class in the Vulkan spec/headers repository is changed, having an incorrect `apiname` parameter might actually have side effects that VVL itself may not have anticipated.

So IMO it's just safer to set such parameters to the correct values, even when we may know that the _current_ implementation isn't affected by them.

aqnuep: That being said, if you'd like me to remove these parameters, I have no issues with that, we can always add them back later if the incorrect parameters break anything.

spencer-lunarg: For now, I rather just get something working in first TBH... there will probably be a lot of `./scripts` changes in the near future, good to have a "this was working" reference

spencer-lunarg (approved):

scripts/lvl_genvk.py
spencer-lunarg: but, if you have to rebase anyway, I would suggest taking the logic

```python
def SetOutputDirectory(directory):
    global outputDirectory
    outputDirectory = directory
    
    # ...
    GeneratorOptions.__init__(self,
        directory = outputDirectory,
```

does an apply it to `apiname`

aqnuep:
Thank you all for the reviews!

PR #5862, opened 2023-05-16
android: Fix and test exported Android functions

juan-lunarg:


tests/android/CMakeLists.txt
```diff
@@ -38,31 +38,56 @@ if ("${CMAKE_ANDROID_STL_TYPE}" MATCHES "static")
...
 
-# TODO, also based on https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/3088/files testing should be a SHARED library instead of an executable.
-# Needs investigation.
+find_program(GNU_NM NAMES nm)
+if (GNU_NM)
+    # Ensure ANativeActivity_onCreate is being exported
+    add_test(NAME ANativeActivity_onCreate COMMAND ${GNU_NM} --dynamic $<TARGET_FILE:vk_layer_validation_tests>)
+    set_tests_properties(ANativeActivity_onCreate
+        PROPERTIES PASS_REGULAR_EXPRESSION "T ANativeActivity_onCreate"
+    )
```
juan-lunarg, line 77: This is equivalent to running the following linux command to check the symbol is properly exported.

Except you can't use pipe with CMake's `add_test`. CMake doesn't use bash.

So instead we replace grep with `PASS_REGULAR_EXPRESSION`.

```
juan@pop-os:~/projects/vvl/build$ nm --dynamic tests/VulkanLayerValidationTests.so | grep "T ANativeActivity_onCreate"
000000000172c824 T ANativeActivity_onCreate
```

No we can ensure our tests our properly exporting `ANativeActivity_onCreate` which was quite painful to get right, and ensure we are properly exporting `vkEnumerateDeviceExtensionProperties` and `vkEnumerateDeviceLayerProperties` to keep compatibility with older Android loaders.

spencer-lunarg (approved):
LGTM, very good documentation of the "hidden magic" that is the Android Vulkan Loader!

Issue #2287, opened 2020-11-03
AHardwareBuffer_describe symbol missing in latest release


Izhido:
Attempting to include the validation layers in the latest release, in a project that targets Android 25, generates the following message:

E/vulkan: failed to load layer library '/data/app/com.blablabla.myapp/lib/arm64/libVkLayer_khronos_validation.so': dlopen failed: cannot locate symbol "AHardwareBuffer_describe" referenced by "/data/app/com.blablabla.myapp/lib/arm64/libVkLayer_khronos_validation.so"...

Interestingly enough, this error also happens with the validation layers that are included in the latest Android SDK - the main reason I went to look for the latest release of the layers was precisely this error message.

mark-lunarg:
@Izhido, in the past, issues like this have been because a version of Android was used that was not current enough (I believe this symbol was added in Android 'O', or Oreo).  Android 25 should be fine however.  Can you enumerate the NDK version you are using?  Thanks!

Izhido:
Sure. I'm currently on NDK 21.3.6528147, which incidentally is the version (I believe) the latest release was compiled against.

Being stuck in Android 25 in my device has given me no small grief - Android Studio offers a fully featured native code profiler which I'd love to use for my app - problem is, the "native code" part of it is enabled for Android 26 and later :(

mark-lunarg:
@Izhido, i'm certainly not an Android expert, but am confident that the validation layers (and the latest Android and Android SDK versions) support this symbol.  The only other thing I can think of might be the setting of APP_PLATFORM in your Application.mk file -- what might that be?

Izhido:
Interesting point. My project **does not** include APP_PLATFORM in its own Application.mk. However, there is one Application.mk at root level in the Oculus Mobile SDK, which includes the following:

```
# NOTE: These properties are ignored for any libraries or applications
# using the externalNativeBuild path. This file can go away, once
# remaining usage can be stripped from test libraries and applications.

# This needs to be defined to get the right header directories for egl / etc
# NOTE: this is ignored from here now, and must be specified in build.gradle!
APP_PLATFORM := android-24
```

mark-lunarg:
[...]

PR #5864, opened 2023-05-17
layers: Improve GetEnvironment() for android

### LINKED ISSUE #4605: Android: CheckAndroidVersion spawns subprocesses when it doesn't have to

jeremyg-lunarg:
Use system_properties.h functions instead of spawning a process to run the getprop command.

NOTE: These functions are only supported in Android API version 26 or higher.

Fixes #4605

### DIFF
--- a/layers/vk_layer_config.cpp
+++ b/layers/vk_layer_config.cpp
@@ -33,6 +33,10 @@
 #include <windows.h>
 #include <direct.h>
 #define GetCurrentDir _getcwd
+#elif defined(__ANDROID__)
+#include <sys/system_properties.h>
+#include <unistd.h>
+#define GetCurrentDir getcwd
 #else
 #include <unistd.h>
 #define GetCurrentDir getcwd
@@ -60,6 +64,13 @@ class ConfigFile {
 
 static ConfigFile layer_config;
 
+#if defined(__ANDROID__)
+static void PropCallback(void *cookie, [[maybe_unused]] const char *name, const char *value, [[maybe_unused]] uint32_t serial) {
+    std::string *property = static_cast<std::string *>(cookie);
+    *property = value;
+}
+#endif
+
 std::string GetEnvironment(const char *variable) {
 #if !defined(__ANDROID__) && !defined(_WIN32)
     const char *output = getenv(variable);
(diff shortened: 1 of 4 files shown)


layers/vk_layer_config.cpp
```diff
@@ -75,21 +86,12 @@ std::string GetEnvironment(const char *variable) {
...
-            return "";
-        } else {
-            value[str_len] = '\0';
-            return string(value);
-        }
+    const prop_info *prop_info = __system_property_find(variable);
+
+    if (prop_info) {
+        std::string property;
+        __system_property_read_callback(prop_info, PropCallback, &property);
```
juan-lunarg, line 93: `__system_property_read_callback` isn't available until __ANDROID_API__ 26 or higher.

Which correlates to passing `ANDROID_PLATFORM` via CMake.

IE the validation layers wouldn't work on older versions of Android. Which I'm fine with. But we should update our code to reflect that.

```
#if __ANDROID_API__ >= 26
void __system_property_read_callback(const prop_info* __pi,
    void (*__callback)(void* __cookie, const char* __name, const char* __value, uint32_t __serial),
    void* __cookie) __INTRODUCED_IN(26);
#endif /* __ANDROID_API__ >= 26 */
```

juan-lunarg (changes requested):
If we decide to raise our android api minimum we will also need to update the following files:

- layers/android/CMakeLists.txt
- layers/utils/android_ndk_types.h

layers/vk_layer_config.cpp
jeremyg-lunarg: what version of android api do we support currently?

juan-lunarg:
> Note that Android API 26 was released in 2017. https://developer.android.com/tools/releases/platforms#8.0

Agreed I think there is no harm in updating our minimum. Just saying to update the relevant files so the user gets good error messages.

juan-lunarg (approved):
We will need to mention this on our release notes somewhere.

jeremyg-lunarg:
> We will need to mention this on our release notes somewhere.

Probably here: https://github.com/KhronosGroup/Vulkan-ValidationLayers/releases -> where we put the android binaries

ncesario-lunarg (approved):
:tada:

juan-lunarg:
FYI @johnzupin

juan-lunarg:
I made an issue on the SDK packaging to make sure we don't forget to mention this.

jeremyg-lunarg:
> I made an issue on the SDK packaging to make sure we don't forget to mention this.

Thanks

spencer-lunarg:
I have no true knowledge who is still developing on 24/25, but for future reference, this is why we added the "edit by hand" support https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/2287

PR #5865, opened 2023-05-17
tests: Add tests for Multi-Entrypoint

spencer-lunarg:
After much refactoring of the Shader Validation over last year, these tests show that both https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4881 and https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/2450 can be closed as we are now properly handling things per-Entrypoint!


ncesario-lunarg (approved):
💯

Issue #2646, opened 2021-03-24
Segfault in RunUsedStruct with two push constants


khyperia:
Possible related issue: https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/2583 (duplicate of https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/2450 )

When validating this SPIR-V, a segfault occurs (`Program received signal SIGSEGV, Segmentation fault.`). The SPIR-V is a bit unusual in that it holds two OpVariable PushConstants (one of which is unused), because it's minimized down from a (rather large) SPIR-V file with multiple entry points - the other OpVariable PushConstant is used in another entry point.

It appears to only reproduce on linux, not windows, which is weird?

<details>
  <summary>SPIR-V, click to expand</summary>

```spirv
; SPIR-V
; Version: 1.0
; Generator: Khronos SPIR-V Tools Assembler; 0
; Bound: 18
; Schema: 0
               OpCapability Shader
               OpCapability VulkanMemoryModel
               OpExtension "SPV_KHR_vulkan_memory_model"
               OpMemoryModel Logical Vulkan
               OpEntryPoint Vertex %1 "main"
               OpName %StructVec4 "StructVec4"
               OpName %VarVec4 "VarVec4"
               OpName %StructFloat "StructFloat"
               OpName %VarFloat "VarFloat"
               OpDecorate %StructVec4 Block
               OpMemberDecorate %StructVec4 0 Offset 0
               OpDecorate %StructFloat Block
               OpMemberDecorate %StructFloat 0 Offset 0
       %void = OpTypeVoid
          %7 = OpTypeFunction %void
       %uint = OpTypeInt 32 0
      %float = OpTypeFloat 32
    %v4float = OpTypeVector %float 4
     %uint_0 = OpConstant %uint 0
     %uint_2 = OpConstant %uint 2
%_ptr_PushConstant_float = OpTypePointer PushConstant %float
 %StructVec4 = OpTypeStruct %v4float
%_ptr_PushConstant_StructVec4 = OpTypePointer PushConstant %StructVec4
    %VarVec4 = OpVariable %_ptr_PushConstant_StructVec4 PushConstant
%StructFloat = OpTypeStruct %float
%_ptr_PushConstant_StructFloat = OpTypePointer PushConstant %StructFloat
   %VarFloat = OpVariable %_ptr_PushConstant_StructFloat PushConstant
[...]

PR #5867, opened 2023-05-17
tests: Add test for Multiple Push Constants

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/2646

We now properly handle checking Push Constants only if accessed from the Entrypoint, so having an unused Push Constant block now will just be ignored

### DIFF
--- a/tests/positive/shader_push_constants.cpp
+++ b/tests/positive/shader_push_constants.cpp
@@ -675,3 +675,73 @@ TEST_F(PositiveShaderPushConstants, PhysicalStorageBufferVertFrag) {
     pipe.InitState();
     pipe.CreateGraphicsPipeline();
 }
+
+TEST_F(PositiveShaderPushConstants, MultipleStructs) {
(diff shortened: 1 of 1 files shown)


ncesario-lunarg (approved):

juan-lunarg (approved):

tests/positive/shader_push_constants.cpp
```diff
@@ -675,3 +675,73 @@ TEST_F(PositiveShaderPushConstants, PhysicalStorageBufferVertFrag) {
     pipe.InitState();
     pipe.CreateGraphicsPipeline();
 }
+
+TEST_F(PositiveShaderPushConstants, MultipleStructs) {
+    TEST_DESCRIPTION("Test having mutlpile structs Push Constant structs, but only one is used.");
```
juan-lunarg, line 680: spelling issue

```suggestion
    TEST_DESCRIPTION("Test having multiple structs Push Constant structs, but only one is used.");
```

PR #5874, opened 2023-05-18
cmake: Fix invalid VS version check

juan-lunarg:
Remove error check entirely. Not very useful and easy to become outdated.


jzulauf-lunarg:
Fix works for me.

spencer-lunarg (approved):

PR #5875, opened 2023-05-18
build: Add arch to deps dir

ncesario-lunarg:
Add the architecture type to the external dependency path when using UPDATE_DEPS=ON.


.github/workflows/vvl.yml
```diff
@@ -165,15 +169,15 @@ jobs:
...
-            external/${{ matrix.config }}/glslang/build/install
-            external/${{ matrix.config }}/Vulkan-Headers/build/install
-            external/${{ matrix.config }}/SPIRV-Headers/build/install
-            external/${{ matrix.config }}/SPIRV-Tools/build/install
-            external/${{ matrix.config }}/robin-hood-hashing/build/install
-            external/${{ matrix.config }}/googletest/build/install
-            external/${{ matrix.config }}/Vulkan-Headers/registry
-            external/${{ matrix.config }}/SPIRV-Headers/include
-            external/${{ matrix.config }}/mimalloc/build/install
+            external/${{ matrix.config }}/${{ win_arch_map[matrix.arch] }}/glslang/build/install
```
juan-lunarg, line 172: Update MinGW as well.

juan-lunarg (changes requested):

.github/workflows/vvl.yml
ncesario-lunarg: This approach doesn't work (see https://github.com/KhronosGroup/Vulkan-ValidationLayers/actions/runs/5017157429).

.github/workflows/vvl.yml
```diff
@@ -22,6 +22,10 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id  }}
   cancel-in-progress: true
 
+win_arch_map:
+  x64: 64
+  Win32: 32
```
juan-lunarg, line 27: Never seen this before. Can you link to a doc/reference?

ncesario-lunarg: That's because it's an experiment to see if I could add generic "maps" to the yml file, but this doesn't work.

.github/workflows/vvl.yml
juan-lunarg: See `android_cmake` for an example of using if statements in GHA.

`if: matrix.build_tests == 'ON'`

.github/workflows/vvl.yml
```diff
@@ -176,7 +176,7 @@ jobs:
             external/${{ matrix.config }}/mimalloc/build/install
           key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ matrix.arch }}-${{ matrix.os }}-${{ matrix.config }}-${{ hashfiles('scripts/known_good.json') }}
       - name: Build
-        run: python3 scripts/github_ci_win_linux.py --build --config ${{ matrix.config }} --cmake='-DUPDATE_DEPS_SKIP_EXISTING_INSTALL=ON'
+        run: python3 scripts/github_ci_win_linux.py --build --config ${{ matrix.config }} --cmake="-DUPDATE_DEPS_SKIP_EXISTING_INSTALL=ON -DUPDATE_DEPS_DIR=external/${{ matrix.config }}"
```
ncesario-lunarg, line 179: Seems like this will be less verbose than specifying different external directories based on an if condition, but I can try the `if` condition route if preferred.

scripts/CMakeLists.txt
```diff
@@ -47,7 +47,11 @@ if (UPDATE_DEPS)
 
     set(UPDATE_DEPS_DIR_SUFFIX "${_build_type}")
     if (ANDROID)
-        set(UPDATE_DEPS_DIR_SUFFIX "android/${UPDATE_DEPS_DIR_SUFFIX}")
+        set(UPDATE_DEPS_DIR_SUFFIX "android/${UPDATE_DEPS_DIR_SUFFIX}/${CMAKE_ANDROID_ARCH_ABI}")
+    elseif(CMAKE_SIZEOF_VOID_P EQUAL 8)
+        set(UPDATE_DEPS_DIR_SUFFIX "${UPDATE_DEPS_DIR_SUFFIX}/64")
```
juan-lunarg, line 52: `common_ci.py` hard codes the external directory so our scripts will now break. Except for WIN32 on CI. But WIN32 locally would break.

ncesario-lunarg: I forgot that was hard-coded. I'll see if I can fix that as part of this PR.

juan-lunarg:
See https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5888 for an alternative approach.

Since you need to reconfigure your build anyway when switching architectures, I just run --clean-build and --clean-install on initial configuration in the scripts/CMakeLists.txt

It also works nicely with expectation of deleting the cache

juan-lunarg:
Closing due to staleness and plans I have to speed up update_deps.py

Issue #5870, opened 2023-05-17
Crash on  ImageAccess::ImageAccess()


HuYuxin:
**Environment:**
 - OS: Linux
 - GPU: Intel
 - SDK or header version if building from repo:
 - Options enabled (synchronization, best practices, etc.):

**Describe the Issue**

After integrating below changes:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/e46a55b7019e8934fea4d574ec34bbbb2bb3811d
https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/1348f53bfa5b4527d2d067baaa5e4caa4c7505ef
https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/f536871709794dd69485c54fa48991623be006d2

Two of the tests in [Dawn](https://github.com/hexops/dawn) are crashing with below callstacks:

```
#0 0x55eb5f0e6922 base::debug::CollectStackTrace()
#1 0x55eb5f0dbb03 base::debug::StackTrace::StackTrace()
#2 0x55eb5f0e63c1 base::debug::(anonymous namespace)::StackDumpSignalHandler()
#3 0x7f15e48e2980 (/lib/x86_64-linux-gnu/libpthread-2.27.so+0x1297f)
#4 0x7f15df96f321 ImageAccess::ImageAccess()
#5 0x7f15df97128d SHADER_MODULE_STATE::StaticData::StaticData()
#6 0x7f15df9dce6d std::__Cr::__shared_ptr_emplace<>::__shared_ptr_emplace<>()
#7 0x7f15df9dcd02 std::__Cr::allocate_shared<>()
#8 0x7f15df9cdce9 ValidationStateTracker::CreateShaderModuleState()
#9 0x7f15df9cdf35 ValidationStateTracker::PostCallRecordCreateShaderModule()
#10 0x7f15e03c878d vulkan_layer_chassis::CreateShaderModule()
#11 0x55eb5ed54802 dawn::native::vulkan::ShaderModule::GetHandleAndSpirv()
#12 0x55eb5ed4fdc2 dawn::native::vulkan::RenderPipeline::Initialize()::$_0::operator()()
#13 0x55eb5ed4ee3f dawn::native::vulkan::RenderPipeline::Initialize()
#14 0x55eb5ecaf680 dawn::native::DeviceBase::CreateRenderPipeline()
[...]

Issue #5614, opened 2023-04-10
Non supported texture sampling method used as function parameter.


Paltoquet:
**Environment:**
 - OS:  Android
 - GPU: Adreno (TM) 630
 - Vulkan Version: 1.1
 - Options enabled (synchronization, best practices, etc.):

**Describe the Issue**

Using a nonsupported texture sampling method will not raise a validation error if texture and sampler are passed through function parameters.

```
#define SampleLvlTex2D(NAME, SAMPLER, COORD, LEVEL) \
textureLod(sampler2D(NAME, SAMPLER), COORD, LEVEL)

vec4 SampleLvlTex2DFunc( texture2D NAME, sampler SAMPLER, vec2 COORD, float LEVEL )
{ return textureLod(sampler2D(NAME, SAMPLER), COORD, LEVEL); }

float3 EnvironmentBRDF(float3 N, float3 V, float3 albedo, float roughness, float metalness)
{
	// this throws the validation error
	float3 specular = SampleLvlTex2D(myTexture, mySampler, uv, roughness);
	// this do not throw :)
	float3 specular = SampleLvlTex2DFunc(myTexture, mySampler, uv, roughness);
}
```
The texture format: VK_FORMAT_R32G32_SFLOAT, VK_FORMAT_D32_SFLOAT or VK_FORMAT_R32G32B32A32_SFLOAT

**Expected behavior**
throws validation error:
[...]

PR #5878, opened 2023-05-19
layers: Prevent crash for Image Access through function calls

spencer-lunarg:
should now close https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5870

This provides a workaround fix to not have apps crash if the `ImageAccess` hits a Function call

The https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5614 issue already is tracking adding proper support. I added many tests, but want to fix the crash as adding a proper fix will take more time and testing

The way GLSL and Tint (Dawn's WGL compiler) handle function calls of Images is different so added both variation tests as well


spencer-lunarg:
confirmed with http://erusea:8080/view/Manual/job/manual-vvltraces/149/ that this fixes the nightly crash (@Tony-LunarG )

jeremyg-lunarg (approved):
LGTM

PR #5881, opened 2023-05-19
tests: Improve testing via vvl_utils

juan-lunarg:
Move the majority of validation layer code into the vvl_utils static library. This allows the tests to link against vvl_utils and test as much code as possible.

The actual vvl target should be as minimal as possible.

The only thing our vvl target requires is exporting specific functions for the vulkan loader.


juan-lunarg:
Closing for now. While the core idea is fine chassis.cpp does far more than define the needed symbol exports.

Issue #3925, opened 2022-03-28
`CmdCopyImage` erroneously allows copies between incompatible compressed formats


Rua:
The code here is meant to check whether the formats of the source and destination image in `vkCmdCopyImage` are compatible:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/70835ff23b3f513d86fcb918ab8352a8f9ee8c1f/layers/buffer_validation.cpp#L3178-L3184

If neither format is a depth/stencil format, then it performs this check based on the texel block sizes of each format. But this is not correct. In the [description section of `vkCmdCopyImage`](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkCmdCopyImage.html#_description) the spec states two things:
> The formats of `srcImage` and `dstImage` **must** be compatible. Formats are compatible if they share the same class, as shown in the [Compatible Formats](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#formats-compatibility) table. Depth/stencil formats **must** match exactly.

> `vkCmdCopyImage` allows copying between _size-compatible_ compressed and uncompressed internal formats. Formats are size-compatible if the texel block size of the uncompressed format is equal to the texel block size of the compressed format.

From this I gather that if _both_ formats are compressed/blocked, the size compatibility rule does not hold, and compatibility is determined by the Compatible Formats table.

sfricke-samsung:
I think doing something a `vkCmdCopyImage` between a `VK_FORMAT_BC3_UNORM_BLOCK` and `VK_FORMAT_BC5_UNORM_BLOCK` is legal... checking on this

Rua:
If I'm wrong, then `GetAdjustedDestImageExtent` may need changing as it currently doesn't account for the case you mentioned.

Rua:
Also may be important: https://github.com/KhronosGroup/Vulkan-Docs/issues/1807

sfricke-samsung:
So I think the right course of action is to make a issue against Vulkan Docs to clarify it, because if it allowed would want to make sure there is CTS coverage.

I guess also not sure the exact use case of wanting to do on-the-fly recompressing of a texture in a first place. Most people will have spent time offline deciding which BC or ASTC variation they want to my knowledge

Issue #3497, opened 2021-11-02
Block texture dimension validation error


frustum:
There is a wrong vkCmdCopyImage-dstOffset-00150|151 error reporting during last mipmaps copying when the source texture is RGBAu32, and the destination texture is block format texture.

The source extent size is 1x1 with RGBAu32 format that represents the minimal single compressed block. The error is reported when the destination mipmap is lower than 4x4 in size. But this data is valid for 2x2 and 1x1 mipmaps.

buffer_validation.cpp:3092 is not checking formats.
// extent_check = ExceedsBounds(&(region.dstOffset), &dst_copy_extent, &subresource_extent);

ziga-lunarg:
I believe this is the same error as with #1908 and #1946.

I am currently waiting for clarification from Vulkan-Docs here: https://github.com/KhronosGroup/Vulkan-Docs/issues/1637

frustum:
Yes, it is the same. Both Nvidia and AMD drivers handle such copying well.
Direct3D12/11 and Metal do not report errors.
Thank you.

ziga-lunarg:
Closing as a duplicate of #1946

Issue #1946, opened 2020-06-16
VUID-VkImageCopy-dstOffset and copying form UINT format to DXT (BC)


mark-lunarg:
[From LunarXchange]
Operating System | Win 10     Graphics Card | any     Driver | any     SDK Version 2.141.0

There could be an issue with VL when copying from uncompressed (UINT) formats to comrpessed (DXT), or I'm doing something wrong there. :/ (There are some problems with lower mipmaps, size less than 4x4). Here...

> Validation Error: [ VUID-VkImageCopy-dstOffset-00150 ] Object 0: handle = 0x22931790, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x5c912440 | vkCmdCopyImage(): Dest image pRegion 0 x-dimension offset [0] + extent [4] exceeds subResource width [2]. The Vulkan spec states: dstOffset.x and (extent.width + dstOffset.x) must both be greater than or equal to 0 and less than or equal to the destination image subresource width (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkImageCopy-dstOffset-00150)

> Validation Error: [ VUID-VkImageCopy-dstOffset-00151 ] Object 0: handle = 0x22931790, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xb73b73ad | vkCmdCopyImage(): Dest image pRegion 0 y-dimension offset [0] + extent [4] exceeds subResource height [2]. The Vulkan spec states: dstOffset.y and (extent.height + dstOffset.y) must both be greater than or equal to 0 and less than or equal to the destination image subresource height (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkImageCopy-dstOffset-00151)

`pvkCmdCopyImage( vkCmdBuf, vkSrcImage, VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL, vkDstImage, VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL, 1, &vkImgCopy);`

- vkImgCopy {srcSubresource={aspectMask=1 mipLevel=0 baseArrayLayer=0 ...} srcOffset={x=0 y=0 z=0 } dstSubresource=...} const VkImageCopy
[...]

Issue #1908, opened 2020-06-04
vkCmdCopyImage validation reports error when copying to compressed image with dimensions < block size


aellem:
The only way to copy from a VK_FORMAT_R16G16B16A16_UINT to a VK_FORMAT_BC1_RGB_UNORM_BLOCK whose size < 4 is by specifying extents of 1 block.

However the validation layer interprets that as dimensions of 4x4 and complains the sizes don't match:

VUID-VkImageCopy-dstOffset-00150(ERROR / SPEC): msgNum: 1553015872 - Validation Error: [ VUID-VkImageCopy-dstOffset-00150 ] Object 0: handle = 0xc556330, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x5c912440 | vkCmdCopyImage(): Dest image pRegion 0 x-dimension offset [0] + extent [4] exceeds subResource width [2]. The Vulkan spec states: dstOffset.x and (extent.width + dstOffset.x) must both be greater than or equal to 0 and less than or equal to the destination image subresource width (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkImageCopy-dstOffset-00150)

aellem:
VUID-vkCmdCopyImage-pRegions-00123 has a similar issue.

If you copy to a < blocksize compressed image the validation layer will complain:  that extent [ 4, 4, 1 ] exceeds the
destination image dimensions.

ziga-lunarg:
Closing as a duplicate of #1946

PR #5882, opened 2023-05-19
Refactoring and infrastructure for syncval aliasing support

jzulauf-lunarg:
1. DRY up the IMAGE_STATE subclass implementation.
2. Add sync val derived class for image
3. Add fake space allocation for opaque representations
4. Convert AccessContext to single flat range map for both linear bindings and opaque representations.

### DIFF
--- a/docs/synchronization.md
+++ b/docs/synchronization.md
@@ -1,5 +1,5 @@
 <!-- markdownlint-disable MD041 -->
-<!-- Copyright 2015-2022 LunarG, Inc. -->
+<!-- Copyright 2015-2023 LunarG, Inc. -->
 [![Khronos Vulkan][1]][2]
 
 [1]: https://vulkan.lunarg.com/img/Vulkan_100px_Dec16.png "https://www.khronos.org/vulkan/"
@@ -138,19 +138,18 @@ Memory access is characterized by the _memory extent_ and _memory usage_.  Memor
 
 ### Memory Address Space
 
-To simplify tracking and updating the state of memory extents as listed below, state is stored in a consistent way for all resource types. Also, to eliminate needing to define separate address spaces for each device memory allocations, VkMemory are placed at “fictitious” base addresses (FBA) within a global address space using a simple “next available” allocation scheme. (Which is robust for all rational Vulkan usage models for months of runtime.) However, because the actual memory organization of certain resources is opaque, the tracking is divided into Linear and Ideal extents which exist in independent spaces.  
+To simplify tracking and updating the state of memory extents as listed below, state is stored in a consistent way for all resource types. Also, to eliminate needing to define separate address spaces for each device memory allocations, VkMemory are placed at “fake” base addresses (FBA) within a global address space using a simple “next available” allocation scheme. (Which is robust for all rational Vulkan usage models for months of runtime.) Additionally, because the actual memory organization of certain resources is opaque, an "opaque range" is allocated with an independent FBA to the extent needed to hold the Synchronization Validation internal representation of these opaque resources. 
 
 
 ### Memory Extent Definitions and Comparisons
 
 Memory extent can be specified in a variety of ways:
 
-
 <table>
   <tr>
    <td>Memory range
    </td>
-   <td>A <code>VkMemory </code>handle referencing a specific memory allocation, and a  contiguous range within the Vulkan memory allocation (offset, size) further offset in the Linear address space by the  FBA
+   <td>A range of addresses within either <code>VkDeviceMemory</code> allocation offset within the "fake" address space, referencing a specific device memory range (i.e. a "linear range"), or a range of "fake" address space represention a portion of the Synchronization Validation representation of a tiled image or other opaquely represented device resource (i.e. an "opaque range").  All accesses to a given resource are contained within the linear range(s) of the `VkDeviceMemory`(s) to which it is bound, or within the opaque range representing the resource.
    </td>
   </tr>
   <tr>
@@ -162,7 +161,7 @@ Memory extent can be specified in a variety of ways:
   <tr>
    <td>Tiled (opaque) image subresource range
    </td>
-   <td>A VkImage handle to an opaque (optimal tiling) image with a VkSubresourceRange which cannot be mapped to a specific subset of the the <code>VkMemory </code>bound to the <code>VkImage</code>.  The subresources are encoded to the  “Idealized” memory range, specific to the image creation parameters which can only be compared relative to “compatible” aliases.  Otherwise all subresources within the range are assumed to alias to *every* device memory address bound to the image.
+   <td>A VkImage handle to an opaque (optimal tiling) image with a VkSubresourceRange which cannot be mapped to a specific subset of the the <code>VkMemory </code>bound to the <code>VkImage</code>.  The subresources are encoded to the opaque range reserved for the resource, specific to the image creation parameters which can only be compared relative to “compatible” aliases.  Otherwise all subresources within the range are assumed to alias to *every* device memory address bound to the image.
    </td>
   </tr>
   <tr>
@@ -175,7 +174,6 @@ Memory extent can be specified in a variety of ways:
 
 
 These distinctions affect the ability of memory references to be compared w.r.t. “overlap” --
-
 
 <table>
   <tr>
@@ -209,19 +207,21 @@ These distinctions affect the ability of memory references to be compared w.r.t.
    </td>
   </tr>
   <tr>
-   <td>Tiled image
+   <td>Tiled image (Opaque Resources)
    </td>
    <td>
    </td>
    <td>
    </td>
-   <td>For “compatible*” tiled images, compare encoded “idealized” memory ranges.  For incompatible images, every subresource is assumed to alias to any subresource in the other image.
+   <td>For “compatible*” tiled images (or other opaque resources), compare encoded opaque ranges.  For incompatible images, every subresource is assumed to alias to all subresources of the other image.
    </td>
   </tr>
 </table>
 
 
-* Compatible tiled images are those with same image resource handle, or with the same image create info with `VK_IMAGE_CREATE_ALIAS_BIT `set and the same memory handle binding offset.
+
+
+* Compatible tiled images are those with same image resource handle, or with the same image create info with `VK_IMAGE_CREATE_ALIAS_BIT ` set and the same `VkDeviceMemory` and binding offset.
 
 For incompatible memory extent types any overlap of the bound range must be assumed to be an overlap of the entire bound range (and thus all subresource ranges).  Note: looking for binding range overlap is a likely important first step before doing more detailed overlap analysis.  Also, as a lower precision option, any comparison of non-similar resources (offset or create info) could revert to the opaque “tiled” case for linear images.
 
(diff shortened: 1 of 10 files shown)


jzulauf-lunarg:
@jeremyg-lunarg @artem-lunarg -- this is where I'm going to be doing the prework for the aliasing support.

jeremyg-lunarg (approved):
LGTM. The IMAGE_STATE cleanup is awesome. The address type stuff as well. Are new unit tests coming for the new functionality?

layers/state_tracker/image_state.h
```diff
@@ -221,6 +224,34 @@ class IMAGE_STATE : public BINDABLE {
 
   protected:
     void NotifyInvalidate(const BASE_NODE::NodeList &invalid_nodes, bool unlink) override;
+
+    template <typename UnaryPredicate>
+    bool AnyAliasBindingOf(BASE_NODE::NodeMap &&bindings, UnaryPredicate pred) {
```
artem-lunarg, line 229: rvalues can be bound to const reference which has the benefits of additional constness.

Suggestion:
`bool AnyAliasBindingOf(const BASE_NODE::NodeMap &bindings, UnaryPredicate pred)`

(I tried, it compiles)

layers/state_tracker/image_state.h
```diff
@@ -221,6 +224,34 @@ class IMAGE_STATE : public BINDABLE {
...
+                        if (pred(*other_image)) return true;
+                    }
+                }
+            }
+        }
+        return false;
+    }
+
+    template <typename UnaryPredicate>
+    bool AnyImageAliasOf(UnaryPredicate &&pred) {
```
artem-lunarg, line 245: `UnaryPredicate &&pred` is a universal reference in this context (can be bound both to lvalue reference and rvalue reference depending on what caller passed). But  `AnyAliasBindingOf` accepts `pred` by value.

Either `AnyAliasBindingOf` should make `pred` parameter to  be a universal reference: `UnaryPredicate &&pred` or lvalue reference should be used `const UnaryPredicate &pred)`. I prefer the latter because in the current implementation in all places where `AnyAliasBindingOf` is called the predicate argument is lvalue reference.

layers/state_tracker/image_state.h
```diff
@@ -142,7 +144,8 @@ class IMAGE_STATE : public BINDABLE {
     bool IsCompatibleAliasing(IMAGE_STATE *other_image_state) const;
 
     // returns true if this image could be using the same memory as another image
-    bool CanAlias() const { return ((createInfo.flags & VK_IMAGE_CREATE_ALIAS_BIT) != 0) || bind_swapchain; }
+    bool CanAliasNonSwapchain() const { return 0 != (createInfo.flags & VK_IMAGE_CREATE_ALIAS_BIT); }
```
artem-lunarg, line 147: The property we check is not related to swapchain but the name makes association with swapchain-related API area (even if it's a non prefix). Can we chose the name that does not mention swapchain? Something like `HasAliasFlag`. 

We have other places in VVL that check for ALIAS_BIT flag. `HasAliasFlag` can be re-used then and `CanAliasNonSwapchain` can not because it has more special semantics.

layers/state_tracker/state_tracker.cpp
```diff
@@ -3803,14 +3803,19 @@ void ValidationStateTracker::UpdateBindImageMemoryState(const VkBindImageMemoryI
     }
 }
 
-void ValidationStateTracker::PostCallRecordBindImageMemory(VkDevice device, VkImage image, VkDeviceMemory mem,
-                                                           VkDeviceSize memoryOffset, VkResult result) {
-    if (VK_SUCCESS != result) return;
+VkBindImageMemoryInfo ValidationStateTracker::ConvertImageMemoryInfo(VkDevice device, VkImage image, VkDeviceMemory mem,
```
artem-lunarg, line 3806: nit: `Convert` in the name is often used to create object-to-object mapping where objects have similar semantics but of a different types in the programming language. In our codebase examples are `ConvertVkRenderPassCreateInfoToV2KHR` or `ConvertToLvlBindPoint`. Personally for me `Create`, `Init` or in some cases `Get` would indicate intent more precisely.

layers/sync/sync_validation.cpp
```diff
@@ -8915,3 +8973,23 @@ FenceSyncState::FenceSyncState(const std::shared_ptr<const FENCE_STATE> &fence_,
 // unsynchronized tag for the Queue being tested against (max synchrononous + 1, perhaps)
 
 ResourceUsageTag AccessContext::AsyncReference::StartTag() const { return (tag_ == kInvalidTag) ? context_->StartTag() : tag_; }
+
+void syncval_state::ImageState::SetOpaqueBaseAddress(ValidationStateTracker &dev_data) {
+    // This is safe to call even for linear resources, or if already called to simplify caller logic
```
artem-lunarg, line 8924: nit: calling this function for linear resource can create assumption that OpaqueBaseAddress is a concept that is applied to all images and should be set for all images, you need to check implementation details to create precise mental model. My personal preference is to do this check at the call site and maybe to have assert here, but up to the author!

layers/sync/sync_validation.cpp
```diff
@@ -8915,3 +8973,23 @@ FenceSyncState::FenceSyncState(const std::shared_ptr<const FENCE_STATE> &fence_,
...
+    if (opaque_base_address_ || IsLinear()) return;
+
+    VkDeviceSize opaque_base = 0U;  // Fakespace Allocator starts > 0
+    auto get_opaque_base = [&opaque_base](const IMAGE_STATE &other) {
+        const ImageState &other_sync = static_cast<const ImageState &>(other);
+        opaque_base = other_sync.opaque_base_address_;
+        return true;
+    };
+    if (IsSwapchainImage()) {
+        AnyAliasBindingOf(bind_swapchain->ObjectBindings(), get_opaque_base);
```
artem-lunarg, line 8934: `AnyAliasBindingOf` is implemented so it returns the first found alias. Do we have a guarantee that we don't have a case when there are multiple swapchain images in ObjectBindings() and at least one has opaque address assigned but the others do not. If `AnyAliasBindingOf` returns image without opaque address assigned then we assign one more opaque address and will have two aliases with two different opaque addresses.

It might be it's not an issue due to how the calls are ordered, so we never have more than one unassigned image in `ObjectBindings()` but it's not obvious, so it's better to ask.

docs/synchronization.md
```diff
@@ -138,19 +138,18 @@ Memory access is characterized by the _memory extent_ and _memory usage_.  Memor
...
+To simplify tracking and updating the state of memory extents as listed below, state is stored in a consistent way for all resource types. Also, to eliminate needing to define separate address spaces for each device memory allocations, VkMemory are placed at “fake” base addresses (FBA) within a global address space using a simple “next available” allocation scheme. (Which is robust for all rational Vulkan usage models for months of runtime.) Additionally, because the actual memory organization of certain resources is opaque, an "opaque range" is allocated with an independent FBA to the extent needed to hold the Synchronization Validation internal representation of these opaque resources.
```
artem-lunarg, line 141: 👍Nice adjustment

docs/synchronization.md
```diff
@@ -524,22 +524,20 @@ Queue Id and Sync stages are used for semaphore, queue, and fence wait operation
...
 
 #### AccessContext
 
 ResourceAccessState and dependency information for a given “context” (for example within a given command buffer, but outside a renderpass, or within a given subpass of a renderpass instance)
 
 Stores:
 
-*   Idealized address space ResourceAccessRangeMap
-*   Linear address space ResourceAccessRangeMap
+*   "Fake"address space ResourceAccessRangeMap
```
artem-lunarg, line 535: Just juggling possible alternatives to the "fake" term, something more transparent that has less chance to alarm readers (is fake in quotes is even more fake or it's two fakes that cancel each other?). Is virtual address space can be used? The downside it's also quite overloaded term.

docs/synchronization.md
```diff
@@ -1315,11 +1313,11 @@ The stage/access for each bound descriptor referenced by a draw or dispatch call
...
+> Designer’s note.  Given the “inmost loop” nature of component read/write enable, it is likely the performance and memory use will be unacceptable unless the opaque range encoding changes the apparent memory organization of components to be at least just above the x, y, and z dimensions, s.t. the effective “run-length-compression” of the range maps is not lost. Storing access state byte-by-byte is likely not going to be acceptable.  Whether this encoding should be adaptive (only present for images with component level information or globally present in the opaque range Synchronization Validation specific encoding) is a design detail TBD. For broadest application this may mean that non-aliased linear tiling images should also be given an an validation specific opaque encoding w.r.t. component granularity information.
```
artem-lunarg, line 1320: I like this `opaque range encoding` wording! It's clear what it means and it also does not mention any memory addresses so the reader doesn't start thinking about memory representation details. So with this terminology I can think in terms "the _sync validation address mapping_ (fake/virtual mapping) is used to represent both _linear memory ranges_ that correspond directly to VkDeviceMemory ranges and also _opaque range encoding_ to model tiled and other resources with opaque memory layout .

artem-lunarg (approved):
LGTM! I added some comments/thoughts/questions, feel free to comment/discuss if anything of interest.

docs/synchronization.md
jzulauf-lunarg: "validation memory space"?  @jeremyg-lunarg @artem-lunarg ?

layers/sync/sync_validation.h
```diff
@@ -39,10 +39,33 @@ struct ResourceFirstAccess;
 class SyncEventsContext;
 struct SyncEventState;
 class SyncValidator;
+
 namespace syncval_state {
 class CommandBuffer;
 class Swapchain;
+
+class ImageState : public IMAGE_STATE {
```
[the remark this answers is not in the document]
jzulauf-lunarg: Each validation object has it's own state tracker, so this is the validation object specific derived class support (@jeremyg-lunarg  is there a design doc for this?)

layers/state_tracker/state_tracker.cpp
artem-lunarg: I don't see it this way. The V1 inputs are not converted. They are used to initialize VkBindImageMemoryInfo structures. In my understanding conversion it's one object -> another object. If we had VkBindImageMemoryInfoV1 and used it to initialize VkBindImageMemoryInfoV2 then I usually call it a conversion. But it might be just different point of views, so I'm fine with how it works now.

Kangz:
FYI we are seeing test failures in Dawn after the VVL roll containing this commit. The tests pass for BC formats but not ASTC formats (when Dawn's code is mostly agnostic to them). Does this commit have code paths that depend on the block size of the compressed formats? CC @jrprice

This is the Dawn issue for context: https://bugs.chromium.org/p/dawn/issues/detail?id=1879

jzulauf-lunarg:
@kangz @jrprice -- I don't see anything that does, however it does reorganized the layout in syncval's internal memory layout it represents memory with, s.t. if the internal representation were incorrect, there may be bounds violations (or it may that previously violated bounds have been corrected in the relayout).  

If the issue isn't a valid syncval error, needed to find a way to reproduce and fix.  

@spencer-lunarg @artem-lunarg  -- we may need to look at way to validate that our ImageRangeEncoder parameterization for *opaque* formats is fully correct across all formats, when addressing within offset and extent windows.

spencer-lunarg:
> we may need to look at way to validate that our ImageRangeEncoder parameterization for opaque image resources is fully correct across all formats

Ok, so this has been on my looooong todo list, but when it comes to these formats, there is a quite a back log of related issues

https://github.com/KhronosGroup/Vulkan-Docs/issues/1005
https://github.com/KhronosGroup/Vulkan-Docs/issues/593
https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/1908
https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/1946
https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/3497
https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/3925

PR #5883, opened 2023-05-20
android: Cleanup Android CMake code

juan-lunarg:
Use CMAKE_SYSTEM_VERSION to check Android API level

Use CMAKE_ANDROID_NDK instead of ANDROID_NDK

Print Android API / NDK version.

Install tests/layer to the same location.

Remove INSTALL_TESTS for simplicity.

### DIFF
--- a/layers/android/CMakeLists.txt
+++ b/layers/android/CMakeLists.txt
@@ -14,18 +14,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ~~~
-if (ANDROID_USE_LEGACY_TOOLCHAIN_FILE)
-    # https://github.com/android/ndk/issues/1693
-    message(WARNING "Using legacy android toolchain!")
-endif()
 
-if (ANDROID_PLATFORM LESS "26")
-    message(FATAL_ERROR "Vulkan-ValidationLayers is not supported on Android 25 and below!")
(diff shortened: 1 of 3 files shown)


layers/android/CMakeLists.txt
```diff
@@ -14,18 +14,12 @@
...
 # ~~~
-if (ANDROID_USE_LEGACY_TOOLCHAIN_FILE)
-    # https://github.com/android/ndk/issues/1693
-    message(WARNING "Using legacy android toolchain!")
-endif()
 
-if (ANDROID_PLATFORM LESS "26")
-    message(FATAL_ERROR "Vulkan-ValidationLayers is not supported on Android 25 and below!")
-endif()
+message(STATUS "Targeting Android API Level ${CMAKE_SYSTEM_VERSION}")
```
ncesario-lunarg, line 18: Does this handle the "less than 26" case?

ncesario-lunarg (approved):
One question, but LGTM!

layers/android/CMakeLists.txt
juan-lunarg: No. I decided to not error in the CMake code. Leaving it up to the C++ code (see layers/utils/android_ndk_types.h) to properly fail.

PR #5886, opened 2023-05-22
layers: Use xml for syncval generation

artem-lunarg:
The latest 1.3.253 headers provide `vk.xml` with new `<sync>` section at the end of the file. 

This PR introduces `sync_validation_generator.py` which implements syncval codegen using sync data from `vk.xml` and removes `sync_val_gen.py` which relied on ad-hoc approaches to get sync data.

New implementation works the same way as the old one, this PR does not introduce new fixes/improvements. Some efforts were made to minimize difference with the original output. The diff in `sync_validation_types.h/cpp` for the commit that introduces new generator ([layers: Use XML for sync val generation](https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5886/commits/41ab2ab9d9e43512900f44d0c5613df0fec5e4b7)) is minimal and a few differences is mostly due to the newly added stages. Later commits add more difference due to the changes to XML data to add more spec details.


spencer-lunarg (approved):

spencer-lunarg:
cc @jeremyg-lunarg and @jzulauf-lunarg as you are probably the 2 who probably know the most about the output of these generated files

scripts/generators/sync_validation_generator.py
```diff
@@ -0,0 +1,726 @@
...
+        file_comment = '// *** THIS FILE IS GENERATED - DO NOT EDIT ***\n'
+        file_comment += '// See sync_validation_generator.py for modifications\n'
+        write(file_comment, file=self.outFile)
+        # Copyright Statement
+        copyright = ''
+        copyright += '\n'
+        copyright += '/***************************************************************************\n'
+        copyright += ' *\n'
+        copyright += ' * Copyright (c) 2015-2023 Valve Corporation\n'
+        copyright += ' * Copyright (c) 2015-2023 LunarG, Inc.\n'
```
arno-lunarg, line 144:
```suggestion
        copyright += ' * Copyright (c) 2023 Valve Corporation\n'
        copyright += ' * Copyright (c) 2023 LunarG, Inc.\n'
```

artem-lunarg: I guess the idea that we do not generate new file - `sync_validation_types.h/cpp` existed before, so it keeps copyright part unchanged. That's something that sounded logical but I don't know much about about copyright nuances and if generated code is somewhat different. So far I've assumed it's a regular code since we're commit it to the repository.

spencer-lunarg: I agree, I think this is fine as is

scripts/generators/sync_validation_generator.py
```diff
@@ -0,0 +1,726 @@
...
+    'graphics' : 'VK_QUEUE_GRAPHICS_BIT',
+    'compute' : 'VK_QUEUE_COMPUTE_BIT',
+    'decode' : 'VK_QUEUE_VIDEO_DECODE_BIT_KHR',
+    'encode'  : 'VK_QUEUE_VIDEO_ENCODE_BIT_KHR',
+    'opticalflow' : 'VK_QUEUE_OPTICAL_FLOW_BIT_NV',
+    'sparse_binding' : 'VK_QUEUE_SPARSE_BINDING_BIT',
+}
+
+transferExpansion = [
+    ('transfer copy', 'VK_PIPELINE_STAGE_2_COPY_BIT'),
```
jzulauf-lunarg, line 49: why is this one a space while the others are _

artem-lunarg: That's a typo. Does not change the generated code, the only requirement it should be unique names. I think good opportunity to fix this when we start updating this code for new code generation framework.

spencer-lunarg: noted

PR #5887, opened 2023-05-22
Various GHA improvements

juan-lunarg:
- ci: Use fail-fast judiciously
- ci: Fix permissions
- ci: Cleanup compiler usage for linux testing
- ci: Use 64 bit compiler when targeting x86

### DIFF
--- a/.github/workflows/vvl.yml
+++ b/.github/workflows/vvl.yml
@@ -98,8 +98,6 @@ jobs:
 
   gn:
     runs-on: ubuntu-22.04
-    strategy:
-      fail-fast: false
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
@@ -181,7 +179,6 @@ jobs:
   android_cmake:
       runs-on: ubuntu-22.04
       strategy:
-        fail-fast: false
         matrix:
           abi: [ arm64-v8a ]
           build_tests: [ "ON" ]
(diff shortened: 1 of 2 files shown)


.github/workflows/vvl.yml
```diff
@@ -64,7 +64,7 @@ jobs:
       - name: Setup ccache
         uses: hendrikmuhs/ccache-action@v1.2
         with:
-          key: ${{ runner.os }}-${{ matrix.config }}-${{ matrix.cc }}-${{ matrix.cpp_std }}
+          key: ${{ runner.os }}-${{ matrix.config }}-${{ matrix.compiler }}-${{ matrix.cpp_std }}
```
artem-lunarg, line 67: Previously it was C `cc` compiler and now we pass configuration for both `cc` and `cxx`, is it the meaning of this line change?

artem-lunarg (approved):

.github/workflows/vvl.yml
juan-lunarg: Thanks for pointing that out. The intent is to hash based on the compiler ID.

Looking at the output from github actions this translates to `Save cache using key "ccache-Linux-release-Object-17-2023-05-22T19:45:38.322Z".`

So that's wrong... I don't know why `${{ matrix.compiler }}` translates to `Object. I'll fix this.

PR #5888, opened 2023-05-22
scripts: Clean fetched dependencies on new configuration

juan-lunarg:
Makes switching architectures easier since CMake will delete old build artifacts.

### DIFF
--- a/scripts/CMakeLists.txt
+++ b/scripts/CMakeLists.txt
@@ -82,12 +82,16 @@ if (UPDATE_DEPS)
     set(UPDATE_DEPS_HASH "0" CACHE STRING "Default value until we run update_deps.py")
     mark_as_advanced(UPDATE_DEPS_HASH)
 
+    if ("${UPDATE_DEPS_HASH}" STREQUAL "0")
+        list(APPEND update_dep_command "--clean-build")
+        list(APPEND update_dep_command "--clean-install")
+    endif()
+
     if ("${md5_hash}" STREQUAL $CACHE{UPDATE_DEPS_HASH})
         message(DEBUG "update_deps.py: no work to do.")
     else()
         execute_process(
             COMMAND ${Python3_EXECUTABLE} ${update_dep_command}
-            WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
             RESULT_VARIABLE _update_deps_result
         )
         if (NOT (${_update_deps_result} EQUAL 0))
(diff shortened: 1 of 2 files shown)


scripts/CMakeLists.txt
```diff
@@ -82,12 +82,16 @@ if (UPDATE_DEPS)
     set(UPDATE_DEPS_HASH "0" CACHE STRING "Default value until we run update_deps.py")
     mark_as_advanced(UPDATE_DEPS_HASH)
 
+    if ("${UPDATE_DEPS_HASH}" STREQUAL "0")
+        list(APPEND update_dep_command "--clean-build")
+        list(APPEND update_dep_command "--clean-install")
+    endif()
```
juan-lunarg, line 88: Essentially during a fresh configuration we delete the old build/install folders.

This removes the problem of using 32 bit binaries on 64 bit since they will be deleted on reconfiguration.

scripts/CMakeLists.txt
```diff
@@ -82,12 +82,16 @@ if (UPDATE_DEPS)
...
+
     if ("${md5_hash}" STREQUAL $CACHE{UPDATE_DEPS_HASH})
         message(DEBUG "update_deps.py: no work to do.")
     else()
         execute_process(
             COMMAND ${Python3_EXECUTABLE} ${update_dep_command}
-            WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
```
juan-lunarg, line 90: Isn't needed.

scripts/update_deps.py
```diff
@@ -32,11 +32,6 @@
 repository at a "known-good" commit in order to provide stability in
 the dependent repositories.
 
-Python Compatibility
---------------------
-
-This program can be used with Python 2.7 and Python 3.
-
```
juan-lunarg, line 39: Hasn't been true for a while

scripts/update_deps.py
```diff
@@ -431,9 +424,11 @@ def CustomBuild(self, repo_dict):
     def CMakeConfig(self, repos):
         """Build CMake command for the configuration phase and execute it"""
         if self._args.do_clean_build:
-            shutil.rmtree(self.build_dir)
+            if os.path.isdir(self.build_dir):
+                shutil.rmtree(self.build_dir, onerror=on_rm_error)
         if self._args.do_clean_install:
-            shutil.rmtree(self.install_dir)
+            if os.path.isdir(self.install_dir):
+                shutil.rmtree(self.install_dir, onerror=on_rm_error)
```
juan-lunarg, line 431: Fixed to be consistent with other uses of `shutil.rmtree`

ncesario-lunarg (approved):
LGTM!

Issue #3543, opened 2021-11-17
feature request: vendor best practices via programmatic interface


samwalls-arm:
Is there currently any way to enable vendor-specific best practice checks via the programmatic interface?

If there is, it would be good to document it (i.e. the programmatic interface section [here](https://vulkan.lunarg.com/doc/view/1.2.189.0/linux/best_practices.html)). Otherwise, are there any plans to allow this?

ncesario-lunarg:
I think we're currently doing this for the best practices tests? https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/21286f81b8ecc60aa2ad0a7dcb17201bfaed675d/tests/vklayertests_best_practices.cpp#L29-L36

mark-lunarg:
This control is achieved through a private layer extension defined in [vk_layer_settings_ext.h](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/layers/vk_layer_settings_ext.h).

TomAtkinsonArm:
Hey @mark-lunarg,

Does `private layer extension` mean that this method can't be used by Vulkan applications and can only be used within the SDK tests? It would be useful for developers to be able to enable different checks pragmatically with the public API

mark-lunarg:
@TomAtkinsonArm, it has been our long-term goal to publish this private extension as an official EXT or LUNARG extension through Khronos.  It should be safe for use by an application by pulling in the necessary header file(s), and if resources can be gathered we'd like to expand the extension to support any/all settings used in the validation layers. Please note that while the interface has not changed in the last 18 months, we can't guarantee that changes can be avoided in the future.

juan-lunarg:
We have made some progress on this front. See https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6245

christophe-lunarg:
@samwalls-arm 

VK_EXT_layer_settings remains a private extension but we are working with the Vulkan working group to get it in the spec.
It's probably not resonnable to use a private extension as it can change but the following code will allow you to configure ARM best practices.

```
    const VK_BOOL32 value = VK_TRUE;
    const VkLayerSettingEXT settings[] = {
        {"VK_LAYER_KHRONOS_validation", "validate_best_practices_arm", VK_LAYER_SETTING_TYPE_BOOL_EXT, 1, &value}
    };

    const VkLayerSettingsCreateInfoEXT layer_settings_create_info{
[...]

PR #5890, opened 2023-05-22
docs: Show usage of Vendor Specific Best Practices

### LINKED ISSUE #3495: Add Mention of Vendor-Specific Best Practice Flags to Example Settings
In this file https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/layers/vk_layer_settings.txt it would be helpful to also have some mention of flags for vendor-specific best practice checks.

What would be the most appropriate way to include these, another section after `<LayerIdentifier>.enables`? Are there any other settings which are missing from the example?

juan-lunarg:
closes #3495

### DIFF
--- a/layers/vk_layer_settings.txt
+++ b/layers/vk_layer_settings.txt
@@ -111,3 +111,32 @@ khronos_validation.enables =
 # performance in multithreaded applications.
 khronos_validation.fine_grained_locking = true
 
+# Best Practices
+# =====================
+# Enable best practices layer
+khronos_validation.enables=VK_VALIDATION_FEATURE_ENABLE_BEST_PRACTICES_EXT
+
+# Best Practices with ARM best practices
+# =====================
+# Enable best practices layer
+khronos_validation.enables=VK_VALIDATION_FEATURE_ENABLE_BEST_PRACTICES_EXT,VALIDATION_CHECK_ENABLE_VENDOR_SPECIFIC_ARM
+
+# Best Practices with AMD best practices
+# =====================
+# Enable best practices layer
+khronos_validation.enables=VK_VALIDATION_FEATURE_ENABLE_BEST_PRACTICES_EXT,VALIDATION_CHECK_ENABLE_VENDOR_SPECIFIC_AMD
+
+# Best Practices with Imagination Tech best practices
+# =====================
+# Enable best practices layer
+khronos_validation.enables=VK_VALIDATION_FEATURE_ENABLE_BEST_PRACTICES_EXT,VALIDATION_CHECK_ENABLE_VENDOR_SPECIFIC_IMG
+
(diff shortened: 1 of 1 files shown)


docs/best_practices.md
```diff
@@ -25,3 +25,24 @@ Best Practices Validation settings are managed by configuring the Validation Lay
...
+features.enabledValidationFeatureCount = 1;
+features.pEnabledValidationFeatures = enables;
+
+VkInstanceCreateInfo info = {};
+info.pNext = &features;
+```
+
+Note you currently cannot enable vendor specific (ARM, AMD, Nvidia, etc) best practices via a programmatic interface.
+
+This is a [known limitation](https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/3543) which is being worked on.
```
juan-lunarg, line 48: This used to be documented but no longer is.

@davidlunarg can provide more context but essentially we wanted to generate more documentation. So we removed documentation.

However, there is now no documentation on how to programmatically set the best practice layer.

I also linked to https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/3543 since if I was a user I'd try and enable the best practices programmatically.

layers/vk_layer_settings.txt
juan-lunarg, line 1: Should this file be moved to the `docs` directory? Doesn't feel appropriate having it here since this essentially serves as documentation for the SDK.

This file gets concatenated along with `VulkanTools/layersvt/vk_layer_settings.txt`, `gfxreconstruct/layer/vk_layer_settings.txt`, and `Vulkan-ExtensionLayer/layers/vk_layer_settings.tx` and then resides in the SDK as an example for users.

KarenGhavam-lunarG:
> [@davidlunarg](https://github.com/davidlunarg) can provide more context but essentially we wanted to generate more documentation. So we removed documentation.

I am not sure I agree with this comment. Documentation common to any Vulkan layer and how to enable it was consolidated here: https://vulkan.lunarg.com/doc/sdk/1.3.246.1/windows/layer_configuration.html.

juan-lunarg:
> I am not sure I agree with this comment. Documentation common to any Vulkan layer and how to enable it was consolidated here: https://vulkan.lunarg.com/doc/sdk/1.3.246.1/windows/layer_configuration.html.

Updated comment. That link doesn't cover how to set the layer programmatically.

KarenGhavam-lunarG:
> Updated comment. That link doesn't cover how to set the layer programmatically.

I would agree with that and would recommend updating the referenced document with how to set layers programmatically. Don't just do that in best practices documentation.

juan-lunarg:
PR only modifies vk_layer_settings.txt now

spencer-lunarg (approved):

ncesario-lunarg (approved):

PR #5891, opened 2023-05-23
android: Remove usage of AHB_VALIDATION_SUPPORT

juan-lunarg:
Removes usage of AHB_VALIDATION_SUPPORT and cleans up surrounding Android code.

Use C++17 std::from_chars to convert string to integer. It's also much safer to use considering it's non-allocating and non-throwing.

### DIFF
--- a/layers/core_checks/cc_android.cpp
+++ b/layers/core_checks/cc_android.cpp
@@ -21,7 +21,7 @@
 #include "generated/vk_enum_string_helper.h"
 #include "core_validation.h"
 
-#ifdef AHB_VALIDATION_SUPPORT
+#if defined(VK_USE_PLATFORM_ANDROID_KHR)
 // Android-specific validation that uses types defined only on Android and only for NDK versions
 // that support the VK_ANDROID_external_memory_android_hardware_buffer extension.
 // This chunk could move into a seperate core_validation_android.cpp file... ?
@@ -471,7 +471,6 @@ bool CoreChecks::ValidateImageImportedHandleANDROID(const char *func_name, VkExt
 }
 
 // Validate creating an image with an external format
-// This could be wrapped with VK_USE_PLATFORM_ANDROID_KHR instead of AHB_VALIDATION_SUPPORT, but this check is only for AHB
 bool CoreChecks::ValidateCreateImageANDROID(const VkImageCreateInfo *create_info) const {
     bool skip = false;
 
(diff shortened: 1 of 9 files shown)


layers/vk_layer_config.cpp
```diff
@@ -413,23 +413,32 @@ void PrintMessageType(VkFlags vk_flags, char *msg_flags) {
...
+#include <charconv>
+
 // This catches before dlopen fails if the default Android-26 layers are being used and attempted to be ran on Android 25 or below
-#if defined(__ANDROID__)
-#include "utils/android_ndk_types.h"  // get AHB_VALIDATION_SUPPORT macro
-void __attribute__((constructor)) CheckAndroidVersion();
-void CheckAndroidVersion() {
-#ifdef AHB_VALIDATION_SUPPORT
-    string version_env = GetEnvironment("ro.build.version.sdk");
-    int target_version = atoi(version_env.c_str());
```
juan-lunarg, line 423: See issues with using atoi (https://releases.llvm.org/16.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/cert/err34-c.html).

Which is why I decided to use C++17 std::from_chars

https://en.cppreference.com/w/cpp/utility/from_chars

layers/core_checks/cc_android.cpp
```diff
@@ -21,7 +21,7 @@
 #include "generated/vk_enum_string_helper.h"
 #include "core_validation.h"
 
-#ifdef AHB_VALIDATION_SUPPORT
```
juan-lunarg, line 24: AHB_VALIDATION_SUPPORT isn't needed as of https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5864

layers/vk_layer_config.cpp
```diff
@@ -413,23 +413,32 @@ void PrintMessageType(VkFlags vk_flags, char *msg_flags) {
...
+    uint32_t target_version{};
+
+    if (std::from_chars(version.data(), version.data() + version.size(), target_version).ec != std::errc()) {
+        return;
+    }
+
+    if (target_version < 26) {
         LOGCONSOLE(
             "ERROR - Targeted Android version is %d and needs to be 26 or above. Please read "
             "https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/BUILD.md for how to build the Validation Layers "
```
spencer-lunarg, line 438: So this is not a valid thing now right? We are removing support for 24/25 with the removal of AHB_VALIDATION_SUPPORT

spencer-lunarg (approved):
LGTM - Just want to make sure it is documented well that we did support 24/25 but now aren't because of <insert reasons>

ncesario-lunarg (approved):
LGTM once Spencer's comment is resolved.

tests/CMakeLists.txt
```diff
@@ -151,8 +151,6 @@ target_compile_definitions(vk_layer_validation_tests PRIVATE VVL_TESTS_USE_CMAKE
 
 add_dependencies(vk_layer_validation_tests vvl)
 
-target_include_directories(vk_layer_validation_tests PRIVATE .)
-
```
juan-lunarg, line 155: This was the main source of my CI troubles. The NDK build doesn't have this include directory which I was relying on. So I'm removing it for compatibility.

Issue #5842, opened 2023-05-12
Missing Negative Tests


spencer-lunarg:
I went through all the validation we are checking but missing a test to see how bad it really was

a LOT of "missing tests" are really just because we have things like `vkCmdDraw` or `VK_KHR_copy_commands2` that would require writing the same test, but with/without extensions (which I think is not high priority at all)

I also removed some sections such as RTX, Vendor EXT, and WSI, and things where it depended on NOT having an extension enabled from the list

There is about 100 tests here that I feel should be added as they are not that hard (might be wrong) and are worth to catch future changes to the layers

- [x] `VUID-VkAttachmentDescription-format-06243` | core 
- [x] `VUID-VkAttachmentDescription-format-06242` | core 
- [x] `VUID-VkAttachmentDescription-format-06487` | (VK_VERSION_1_1,VK_KHR_maintenance2) 
- [x] `VUID-VkAttachmentDescription-format-06488` | (VK_VERSION_1_1,VK_KHR_maintenance2) 
- [x] `VUID-VkBufferCreateInfo-flags-00915`   | core 
- [x] `VUID-VkBufferCreateInfo-flags-00916`   | core 
- [x] `VUID-VkBufferCreateInfo-flags-00917`   | core 
- [x] `VUID-VkBufferCreateInfo-size-06409`    | (VK_VERSION_1_3,VK_KHR_maintenance4) 
- [x] `VUID-VkBufferCreateInfo-usage-08103`   | (VK_EXT_descriptor_buffer) 
- [x] `VUID-VkClearDepthStencilValue-depth-00022` | (VK_EXT_depth_range_unrestricted) 
- [x] `VUID-VkColorBlendEquationEXT-constantAlphaColorBlendFactors-07362` | (VK_EXT_extended_dynamic_state3,VK_EXT_shader_object)+(VK_KHR_portability_subset) 
- [x] `VUID-VkColorBlendEquationEXT-constantAlphaColorBlendFactors-07363` | (VK_EXT_extended_dynamic_state3,VK_EXT_shader_object)+(VK_KHR_portability_subset) 
- [x] `VUID-VkCommandBufferBeginInfo-flags-00055` | core 
- [x] `VUID-VkCopyDescriptorSet-dstArrayElement-00348`    | core 
- [x] `VUID-VkCopyDescriptorSet-dstBinding-00347` | core 
- [x] `VUID-VkCopyDescriptorSet-dstBinding-02632` | core 
[...]

PR #5893, opened 2023-05-23
Remove ObjectLifetime check from stateless

spencer-lunarg:
Adds 2 tests for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5842

Currently we validate `VUID-VkDescriptorSetLayoutBinding-descriptorType-00282` in  the `ObjectLifetimes::PreCallValidateCreateDescriptorSetLayout`

If you turn off `ObjectLifetimes` you can crash as the `VkSampler` is bad, but the stateless check was only checking if it was `VK_NULL_HANDLE`


artem-lunarg (approved):
LGTM!

PR #5897, opened 2023-05-23
TEST ONLY: Try main branch on Profiles repository

christophe-lunarg:
TEST ONLY!


juan-lunarg:
image

Note because you are using a branch and not a fork the same testing is running twice.

juan-lunarg:
Also worth noting you can run this script locally which would be faster.

```
python3 scripts/github_ci_win_linux.py --build --config debug --cmake='-DVVL_CPP_STANDARD=17 -DVVL_ENABLE_ASAN=ON'

python3 scripts/github_ci_win_linux.py --test
```

Note the test script only works on Linux at the moment.

PR #5899, opened 2023-05-23
layers: Remove stateless validation from config

ncesario-lunarg:
Remove the stateless/parameter validation option from config settings
and leave it enabled by default.

Validation makes the assumption that stateless validation has run. In
addition, stateless validation is usually innocuous, having a negligible
affect on performance. For these reasons, it was determined that
stateless validation should always be enabled.

### DIFF
--- a/tests/negative/video.cpp
+++ b/tests/negative/video.cpp
@@ -4443,14 +4443,32 @@ TEST_F(VkVideoBestPracticesLayerTest, BindVideoSessionMemory) {
 
     VideoContext context(DeviceObj(), config);
 
+    // Create a buffer to get non-video-related memory requirements
+    VkBufferCreateInfo buffer_create_info =
+        LvlInitStruct<VkBufferCreateInfo>(nullptr, static_cast<VkBufferCreateFlags>(0), static_cast<VkDeviceSize>(4096),
+                                          static_cast<VkBufferUsageFlags>(VK_BUFFER_USAGE_TRANSFER_SRC_BIT));
+    vk_testing::Buffer buffer(*m_device, buffer_create_info);
+    VkMemoryRequirements buf_mem_reqs;
+    vk::GetBufferMemoryRequirements(device(), buffer, &buf_mem_reqs);
+
+    // Create non-video-related DeviceMemory
+    VkMemoryAllocateInfo alloc_info = LvlInitStruct<VkMemoryAllocateInfo>();
+    alloc_info.allocationSize = buf_mem_reqs.size;
+    ASSERT_TRUE(m_device->phy().set_memory_type(buf_mem_reqs.memoryTypeBits, &alloc_info, VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT));
+    vk_testing::DeviceMemory memory(*m_device, alloc_info);
+
+    // Set VkBindVideoSessionMemoryInfoKHR::memory to an allocation created before GetVideoSessionMemoryRequirementsKHR was called
     auto bind_info = LvlInitStruct<VkBindVideoSessionMemoryInfoKHR>();
+    bind_info.memory = memory;
+    bind_info.memoryOffset = 0;
+    bind_info.memorySize = alloc_info.allocationSize;
 
     m_errorMonitor->SetDesiredFailureMsg(kWarningBit,
                                          "UNASSIGNED-BestPractices-vkBindVideoSessionMemoryKHR-requirements-count-not-retrieved");
     context.vk.BindVideoSessionMemoryKHR(m_device->device(), context.Session(), 1, &bind_info);
     m_errorMonitor->VerifyFound();
 
-    uint32_t mem_req_count;
+    uint32_t mem_req_count = 0;
     context.vk.GetVideoSessionMemoryRequirementsKHR(m_device->device(), context.Session(), &mem_req_count, nullptr);
 
     if (mem_req_count > 0) {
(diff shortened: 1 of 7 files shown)


tests/negative/video.cpp
```diff
@@ -4467,17 +4467,15 @@ TEST_F(VkVideoBestPracticesLayerTest, BindVideoSessionMemory) {
 
     auto bind_info = LvlInitStruct<VkBindVideoSessionMemoryInfoKHR>();
 
-    m_errorMonitor->SetDesiredFailureMsg(kWarningBit,
-                                         "UNASSIGNED-BestPractices-vkBindVideoSessionMemoryKHR-requirements-count-not-retrieved");
```
ncesario-lunarg, line 4471: @aqnuep Are these VUIDs no longer necessary if parameter validation is enabled?

layers/layer_options.cpp
```diff
@@ -523,6 +522,9 @@ void ProcessConfigAndEnvSettings(ConfigAndEnvSettings *settings_data) {
         SetValidationSetting(settings_data->disables, shader_validation_caching, SETTING_CHECK_SHADERS_CACHING);
     }
 
+    // Stateless validation must always be enabled
+    settings_data->disables[stateless_checks] = true;
```
Tony-LunarG, line 526: Does this contradict VK_VALIDATION_FEATURE_DISABLE_ALL_EXT?  I think I would expect DISABLE_ALL to disable stateless_checks too.  But I could be wrong

spencer-lunarg (approved):
Few questions still left, but other LGTM

I would add this information somewhere in the `docs` folder as well

layers/layer_options.cpp
Tony-LunarG: And the spec says that VK_VALIDATION_FEATURE_DISABLE_API_PARAMETERS_EXT disables stateless parameter validation - Do we need a spec change to go with this?

ncesario-lunarg: > Does this contradict VK_VALIDATION_FEATURE_DISABLE_ALL_EXT? I think I would expect DISABLE_ALL to disable stateless_checks too. But I could be wrong

Yes, DISABLE_ALL should probably disable all. Though, I would think one would just not enable VVL altogether in that case?

> And the spec says that VK_VALIDATION_FEATURE_DISABLE_API_PARAMETERS_EXT disables stateless parameter validation - Do we need a spec change to go with this?

That's a good question. I'm not sure if we want a spec change, but if we continue to allow `VK_VALIDATION_FEATURE_DISABLE_API_PARAMETERS_EXT`, we should probably warn that it is not supported?

Tony-LunarG: > Yes, DISABLE_ALL should probably disable all. Though, I would think one would just not enable VVL altogether in that case?
My use case is to disable all, and then enable GPU-AV, BP, or SyncVal.  But I imagine most folks are using vkconfig for feature selection.  Will vkconfig need updates?
> 

> That's a good question. I'm not sure if we want a spec change, but if we continue to allow `VK_VALIDATION_FEATURE_DISABLE_API_PARAMETERS_EXT`, we should probably warn that it is not supported?

A warning with an explanation that some layers rely on stateless validation to not crash would be appropriate.  Especially in vkconfig.

tests/negative/video.cpp
aqnuep: This (and the one below) is a best-practices warning VU, similar to the ones used in other memory requirement queries to check that the requirements of a video session were retrieved before binding memory to it, so it should stay AFAICT.

I'm not sure what is the new "UNASSIGNED-GeneralParameterError-RequiredParameter" VU error, but I don't quite get why these changes are proposed to be made to this test case.

aqnuep: Ah, okay, then I think the test should be updated to specify a valid handle for `memory` (by allocating one), as this was meant to test the best practices warnings indicated.

I guess I just missed adding valid values to `VkBindVideoSessionMemoryInfoKHR` simply because I didn't get an error as is. Honestly, I didn't know that stateless validation wasn't enabled by default, I thought it is, as I got some stateless validation errors elsewhere. Was this limitation specific to the best practices VU tests?

In any case, the proposed modifications to the test case are incorrect as it would no longer test those best practices VUs. Let me know if you need any help with getting the right changes in to this test case.

ncesario-lunarg: > I guess I just missed adding valid values to VkBindVideoSessionMemoryInfoKHR simply because I didn't get an error as is. Honestly, I didn't know that stateless validation wasn't enabled by default, I thought it is, as I got some stateless validation errors elsewhere. Was this limitation specific to the best practices VU tests?

No, stateless was disabled for best practices previously. We disable pretty much everything when running best practices for performance reasons, but the perf hit on stateless is nominal, and I think it makes sense to enable during testing at least (though I am advocating for enabling by default in vkconfig as well with this change).

> In any case, the proposed modifications to the test case are incorrect as it would no longer test those best practices VUs. Let me know if you need any help with getting the right changes in to this test case.

:+1:  this is what I was unsure of; thank you for clarifying! I'll see if I can make the appropriate change correctly and ping you if I run into issues (I'm not very familiar with the video extension, so hopefully copy/pasting from your other tests will suffice :)).

ncesario-lunarg: Let me know if something is wrong with the these new changes.

aqnuep: This seems reasonable, although you could have just created a memory from the first memory type and be done with it. But anything works for me if the tests pass fine and the best practices warnings are properly covered.

juan-lunarg (commented):
Is there any documentation here or in the SDK docs that would need to be updated?

juan-lunarg (commented):
This change LGTM any reason to not merge this in? I know this was discussed a bit ago, but I forgot the details.

arno-lunarg:
This is a link to an (the?) internal discussion on this: https://chat.google.com/room/AAAAOXVAYGg/XZFD0TPpKgw

ncesario-lunarg:
I have removed the config changes and just left in the changes to enable stateless for all tests.

juan-lunarg (approved):

PR #5900, opened 2023-05-24
tests: Add tests if Sampler/SampledImage share same slot

spencer-lunarg:
Make sure we are testing correctly that if both a `SAMPLER` and `SAMPLED_IMAGE` are sharing the same descriptor slot, it really is a `COMBINED_IMAGE_SAMPLER`

Also improved the VU wording to be more clear what is going on

### DIFF
--- a/layers/core_checks/cc_shader.cpp
+++ b/layers/core_checks/cc_shader.cpp
@@ -459,7 +459,7 @@ static std::set<uint32_t> TypeToDescriptorTypeSet(const SHADER_MODULE_STATE &mod
 static std::string string_descriptorTypeSet(const std::set<uint32_t> &descriptor_type_set) {
     std::stringstream ss;
     for (auto it = descriptor_type_set.begin(); it != descriptor_type_set.end(); ++it) {
-        if (ss.tellp()) ss << ", ";
+        if (ss.tellp()) ss << " or ";
         ss << string_VkDescriptorType(VkDescriptorType(*it));
     }
     return ss.str();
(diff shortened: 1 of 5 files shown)


layers/core_checks/cc_shader.cpp
```diff
@@ -2174,8 +2174,8 @@ bool CoreChecks::ValidateShaderDescriptorVariable(const SHADER_MODULE_STATE &mod
             const LogObjectList objlist(module_state.vk_shader_module(), pipeline.PipelineLayoutState()->layout());
             skip |=
                 LogError(objlist, vuid,
-                         "%s(): pCreateInfos[%" PRIu32 "] Set %" PRIu32 " Binding %" PRIu32
-                         " type mismatch on descriptor slot in shader (%s), uses type %s but expected %s",
+                         "%s(): pCreateInfos[%" PRIu32 "] has a type mismatch for descriptor slot [Set %" PRIu32 " Binding %" PRIu32
+                         "] for shader (%s), uses type %s but expected (%s).",
```
arno-lunarg, line 2178:
```suggestion
                         "] for shader (%s), uses type (%s) but expected (%s).",
```
For consistency?

arno-lunarg (approved):
LGTM! One comment

layers/core_checks/cc_shader.cpp
spencer-lunarg: Agree,

The main goal was to add the tests... I want to re-do all the error messages in this function, will have a follow-up PR soon for all of these error messages

PR #5903, opened 2023-05-24
Fix some tests failing on AMD

Tony-LunarG:
I noticed these during my SDK testing.


juan-lunarg (approved):
Changes look good. I wasn't able to run the Mesh/DescriptorBuffer test though with my current Mesa driver.

For the FragmentShadingRate it passed fine on Mesa, so I couldn't confirm the fix. But the fix still makes sense.

ncesario-lunarg (approved):

juan-lunarg:
Rebased via github UI to fix integration test failure

spencer-lunarg (approved):

PR #5905, opened 2023-05-24
scripts: Enable testing script on Windows

juan-lunarg:
Currently can't run on GHA due to security permissions.

It works locally on my Windows machine now though.

### DIFF
--- a/scripts/common_ci.py
+++ b/scripts/common_ci.py
@@ -210,21 +210,28 @@ def BuildProfileLayer():
 #
 # Run the Layer Validation Tests
 def RunVVLTests():
-    print("Run Vulkan-ValidationLayer Tests using Mock ICD")
-
-    if IsWindows():
-        print("Not implemented yet")
-        exit(-1)
-
-    lvt_cmd = os.path.join(PROJECT_ROOT, BUILD_DIR_NAME, 'tests', 'vk_layer_validation_tests')
+    print("Run VVL Tests using Mock ICD")
 
     lvt_env = dict(os.environ)
 
     # Because we installed everything to TEST_INSTALL_DIR all the libraries/json files are in pre-determined locations
-    # defined by GNUInstallDirs. This makes adding the LD_LIBRARY_PATH and VK_LAYER_PATH trivial/robust.
-    lvt_env['LD_LIBRARY_PATH'] = os.path.join(TEST_INSTALL_DIR, 'lib')
-    lvt_env['VK_LAYER_PATH'] = os.path.join(TEST_INSTALL_DIR, 'share/vulkan/explicit_layer.d')
-    lvt_env['VK_DRIVER_FILES'] = os.path.join(TEST_INSTALL_DIR, 'share/vulkan/icd.d/VkICD_mock_icd.json')
+    # defined by GNUInstallDirs. This makes setting VK_LAYER_PATH and other environment variables trivial/robust.
+    if IsWindows():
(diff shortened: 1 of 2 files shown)


scripts/github_ci_win_linux.py
```diff
@@ -16,14 +16,21 @@
...
 def Build(args):
     config = args.configuration
 
+    # Since this script uses Ninja to build Windows users need to be in a developer command prompt.
+    if common_ci.IsWindows():
+        # This environment variable is arbitrary. I just picked one set by the developer command prompt.
+        if "VSCMD_ARG_TGT_ARCH" not in os.environ:
+            print("This script must be invoked in a developer command prompt!")
+            sys.exit(1)
+
```
juan-lunarg, line 33: This is my attempt at checking for the developer command prompt being present. Not super happy with this approach but it works.

Tested with a regular powershell prompt and it gave the proper error message.

scripts/common_ci.py
```diff
@@ -210,21 +210,28 @@ def BuildProfileLayer():
...
-    lvt_env['LD_LIBRARY_PATH'] = os.path.join(TEST_INSTALL_DIR, 'lib')
-    lvt_env['VK_LAYER_PATH'] = os.path.join(TEST_INSTALL_DIR, 'share/vulkan/explicit_layer.d')
-    lvt_env['VK_DRIVER_FILES'] = os.path.join(TEST_INSTALL_DIR, 'share/vulkan/icd.d/VkICD_mock_icd.json')
+    # defined by GNUInstallDirs. This makes setting VK_LAYER_PATH and other environment variables trivial/robust.
+    if IsWindows():
+        # TODO: Enable Windows testing on GHA. GHA run the test with elevated permissions which causes the loader
+        # to ignore VK_LAYER_PATH for security reasons.
+        #
+        # Perhaps this can be fixed with GHA or more likely with registry editing.
+        # Ideally the registry editing is only done during GHA to avoid impacting local users who run this script.
```
juan-lunarg, line 224: Sadly I'm running into a security issue with running this script on github actions.

`Loader is running with elevated permissions. Environment variable VK_LAYER_PATH will be ignored`

@charles-lunarg has suggested editing the registry. Which may be what I have to in another PR unless anyone else has any ideas.

juan-lunarg (commented):
I'm now able to run the script locally on my Windows machine.

scripts/common_ci.py
charles-lunarg: USE_UNSAFE_FILE_SEARCH is a build option you can use to allow VK_LAYER_PATH to be used while in admin mode.

juan-lunarg: I'll try that in a follow up PR.

scripts/github_ci_win_linux.py
artem-lunarg: Yep, detecting anything related to VS was always a pain. Not sure if people do this but one option is to check if you can run cl.exe process but also can not recommend this as a good option, just something that came to mind.

artem-lunarg (approved):
LGTM 🙏

artem-lunarg (approved):
LGTM 🙏

scripts/github_ci_win_linux.py
juan-lunarg: Yeah I also thought about checking for cl.exe proccess in the path. But I didn't see too much more merit to that approach over checking an environment variable. I guess we will see how this approach goes for now. I can change if it ever becomes a problem.

PR #5910, opened 2023-05-25
Fixed rtti settings for chromium builds.

gaaclarke:
I believe this is the proper thing to do in chromium builds.  In chromiums buildroot the default is `no_rtti`:https://chromium.googlesource.com/chromium/src/+/master/build/config/BUILDCONFIG.gn#347

I think this may only work for some people because it appears when having `rtti` and `no_rtti`; the ordering of the flags may result in `rtti`.

I'm not sure what the criteria are for changing this behavior now though.

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -445,6 +445,7 @@ foreach(layer_info, layers) {
     } else {
       configs -= [ "//build/config/compiler:chromium_code" ]
       configs += [ "//build/config/compiler:no_chromium_code" ]
+      configs -= [ "//build/config/compiler:no_rtti" ]
       configs += [ "//build/config/compiler:rtti" ]
     }
     configs -= vulkan_undefine_configs


dnfield:
Chromium doesn't run into this because of https://source.chromium.org/chromium/chromium/src/+/main:build/config/compiler/BUILD.gn;l=1972;bpv=1;bpt=0

That line was added to support compilation with `gcc` which has a different opinion from `clang` about whether RTTI is enabled by default.


Flutter would like to avoid relying on clang's behavior when you specify both `-fno-rtti` and `-frtti` on the command line if possible. Right now clang seems to take the last specified option (or maybe the one closest to the source file?).

spencer-lunarg:
to fix CI, please start commit message with `build:`

gaaclarke:
> to fix CI, please start commit message with `build:`

done

BUILD.gn
```diff
@@ -445,6 +445,7 @@ foreach(layer_info, layers) {
     } else {
       configs -= [ "//build/config/compiler:chromium_code" ]
       configs += [ "//build/config/compiler:no_chromium_code" ]
+      configs -= [ "//build/config/compiler:no_rtti" ]
```
juan-lunarg, line 448: I'm confused why this is needed? Doesn't the line below enable `rtti`? Do we also need to do this for the fuschia build?

Could we add static asserts in our code to ensure RTTI is enabled?

I know MSVC defines `_CPPRTTI` as 1 if the [/GR (Enable Run-Time Type Information)](https://learn.microsoft.com/en-us/cpp/build/reference/gr-enable-run-time-type-information?view=msvc-170) compiler option is set. Otherwise, undefined.

I'm not sure off the top of my head if GCC/Clang has something similar we could leverage though.

juan-lunarg: > I'm confused why this is needed? Doesn't the line below enable rtti? Do we also need to do this for the fuschia build?

To be clear I believe your change is needed (otherwise you wouldn't make it). My main point would be that a comment explaining this would be helpful in case I need to modify the chromium build and don't want to break anything.

gaaclarke: This is required because the validation_layers don't compile in `flutter` without it.

The `BUILDCONFIG.gn` files for flutter and chromium add `"//build/config/compiler:no_rtti"` to all projects by default.  So, when we add `"//build/config/compiler:rtti"` we end up with both `"//build/config/compiler:no_rtti"` and `"//build/config/compiler:rtti"`.

It just so happens in chromium that compiling something like `clang++ -fno-rtti -frtti foo.cc` will result in `rtti`, which is what we want.  It's risky relying on: 1) clang evaluating conflicting flags like that 2) the order in which GN adds flags.

This happy accident doesn't work for flutter because `no_rtti` is defined to rely on the fact that the default option for clang is `-frtti` so it doesn't specify any cflags.  So if you end up with `"//build/config/compiler:no_rtti"` and `"//build/config/compiler:rtti"` you just end up with `-frtti`, not `-fno-rtti -frtti`.

It's confusing, hope that clears it up.

juan-lunarg: > The BUILDCONFIG.gn files for flutter and chromium add "//build/config/compiler:no_rtti" to all projects by default.

Modern CMake uses `target_` based functionality to avoid that exact problem.

Regardless I understand the problem now.

juan-lunarg (approved):

BUILD.gn
juan-lunarg: To be clear I do understand the desire to set compiler flags globally. But it's always complicated as to what can be set globally safely.

RTTI is one such example. Exceptions are another. Even LTO can be dangerous to set globally. We had a really bad LTO crash caused by a library not following ODR. Etc.

Issue #5760, opened 2023-04-28
Move thread safety code that should not be generated


artem-lunarg:
Some code form thread safety validation can live in a regular h/cpp files.
Suggested here: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5738#pullrequestreview-1406535544

PR #5912, opened 2023-05-26
layers: Split thread-safety into regular and generated code

artem-lunarg:
This moves parts of thread-safety code to regular cpp/h files.
The moved code lived in script files but otherwise did not use codegen facilities.

Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5760

This PR does not do any other changes to the thread safety code other than moves pieces around. The goal of this PR is to simplify further modifications (that we want to make!)


artem-lunarg:
gn failure:
```
ERROR at //build/config/BUILDCONFIG.gn:521:5: Source file not found.
    target(_target_type, _target_name) {
    ^-----------------------------------
The target:
  //:VkLayer_khronos_validation
has a source file:
  //layers/vulkan/generated/thread_safety.h
which was not found.
```

How to make it work?

layers/vulkan/generated/thread_safety_counter_bodies.h
```diff
@@ -0,0 +1,67 @@
+
+// This file is ***GENERATED***.  Do Not Edit.
+// See thread_safety_generator.py for modifications.
+
+/* Copyright (c) 2015-2023 The Khronos Group Inc.
```
artem-lunarg, line 5: All new files that have code from the old file  (either manual written or generated) use the license from the old files in unchanged form.

layers/thread_tracker/thread_safety_validation.cpp
```diff
@@ -0,0 +1,735 @@
+/* Copyright (c) 2015-2023 The Khronos Group Inc.
```
artem-lunarg, line 1: All new files that have code from the old file (either manual written or generated) use the license from the old files in unchanged form.

arno-lunarg (approved):

PR #5913, opened 2023-05-26
tests: Add SPIRV-Hopper

spencer-lunarg:
Adds a new `SPIRV-Hopper` test harness

I have ran this on a lot of shaders generated from various places, and it now passes them all

As more shaders are added, this can be updated, but this is in good enough shape to start using for CI once merged

### DIFF
--- a/tests/device_profiles/spirv_hopper_profile.json
+++ b/tests/device_profiles/spirv_hopper_profile.json
@@ -1,0 +1,793 @@
+{
+    "$schema": "https://schema.khronos.org/vulkan/profiles-0.8.0-latest.json#",
+    "profiles": {
+        "SPIRV_HOPPER_PROFILE": {
+            "version": 1,
+            "api-version": "1.3.250",
+            "label": "SPIRV-Hopper Profile",
+            "description": "Profile that supports as much for SPIR-V shaders",
+            "contributors": {},
+            "history": [
+                {
+                    "revision": 1,
+                    "date": "2023-05-31",
+                    "author": "SPIRV-Hopper",
+                    "comment": "handed-groomed from what is only needed for SPIRV-Hopper"
+                }
+            ],
+            "capabilities": [
+                "device"
+            ]
+        }
+    },
+    "capabilities": {
+        "device": {
+            "features": {
+                "VkPhysicalDeviceFeatures": {
+                    "robustBufferAccess": true,
+                    "geometryShader": true,
+                    "tessellationShader": true,
+                    "sampleRateShading": true,
+                    "alphaToOne": true,
+                    "multiViewport": true,
+                    "vertexPipelineStoresAndAtomics": true,
+                    "fragmentStoresAndAtomics": true,
+                    "shaderTessellationAndGeometryPointSize": true,
+                    "shaderImageGatherExtended": true,
+                    "shaderStorageImageExtendedFormats": true,
+                    "shaderStorageImageMultisample": true,
+                    "shaderStorageImageReadWithoutFormat": true,
+                    "shaderStorageImageWriteWithoutFormat": true,
+                    "shaderUniformBufferArrayDynamicIndexing": true,
+                    "shaderSampledImageArrayDynamicIndexing": true,
+                    "shaderStorageBufferArrayDynamicIndexing": true,
+                    "shaderStorageImageArrayDynamicIndexing": true,
+                    "shaderClipDistance": true,
+                    "shaderCullDistance": true,
+                    "shaderFloat64": true,
+                    "shaderInt64": true,
+                    "shaderInt16": true,
+                    "shaderResourceResidency": true,
+                    "shaderResourceMinLod": true,
+                    "sparseBinding": true,
+                    "sparseResidencyBuffer": true,
+                    "sparseResidencyImage2D": true,
+                    "sparseResidencyImage3D": true,
+                    "sparseResidency2Samples": true,
+                    "sparseResidency4Samples": true,
+                    "sparseResidency8Samples": true,
+                    "sparseResidency16Samples": true,
+                    "sparseResidencyAliased": true,
+                    "variableMultisampleRate": true
+                },
+                "VkPhysicalDeviceVariablePointersFeatures": {
+                    "variablePointersStorageBuffer": true,
+                    "variablePointers": true
+                },
+                "VkPhysicalDeviceMultiviewFeatures": {
+                    "multiview": true,
+                    "multiviewGeometryShader": true,
+                    "multiviewTessellationShader": true
+                },
+                "VkPhysicalDevice16BitStorageFeatures": {
+                    "storageBuffer16BitAccess": true,
+                    "uniformAndStorageBuffer16BitAccess": true,
+                    "storagePushConstant16": true,
+                    "storageInputOutput16": true
+                },
+                "VkPhysicalDeviceShaderSubgroupExtendedTypesFeatures": {
+                    "shaderSubgroupExtendedTypes": true
+                },
+                "VkPhysicalDeviceSubgroupSizeControlFeatures": {
+                    "subgroupSizeControl": true,
+                    "computeFullSubgroups": true
+                },
+                "VkPhysicalDeviceMaintenance4Features": {
+                    "maintenance4": true
+                },
+                "VkPhysicalDeviceShaderDrawParametersFeatures": {
+                    "shaderDrawParameters": true
+                },
+                "VkPhysicalDeviceShaderFloat16Int8Features": {
+                    "shaderFloat16": true,
+                    "shaderInt8": true
+                },
+                "VkPhysicalDeviceDescriptorIndexingFeatures": {
+                    "shaderInputAttachmentArrayDynamicIndexing": true,
+                    "shaderUniformTexelBufferArrayDynamicIndexing": true,
+                    "shaderStorageTexelBufferArrayDynamicIndexing": true,
+                    "shaderUniformBufferArrayNonUniformIndexing": true,
+                    "shaderSampledImageArrayNonUniformIndexing": true,
+                    "shaderStorageBufferArrayNonUniformIndexing": true,
+                    "shaderStorageImageArrayNonUniformIndexing": true,
+                    "shaderInputAttachmentArrayNonUniformIndexing": true,
+                    "shaderUniformTexelBufferArrayNonUniformIndexing": true,
+                    "shaderStorageTexelBufferArrayNonUniformIndexing": true,
+                    "descriptorBindingUniformBufferUpdateAfterBind": true,
+                    "descriptorBindingSampledImageUpdateAfterBind": true,
+                    "descriptorBindingStorageImageUpdateAfterBind": true,
+                    "descriptorBindingStorageBufferUpdateAfterBind": true,
+                    "descriptorBindingUniformTexelBufferUpdateAfterBind": true,
+                    "descriptorBindingStorageTexelBufferUpdateAfterBind": true,
+                    "descriptorBindingUpdateUnusedWhilePending": true,
+                    "descriptorBindingPartiallyBound": true,
+                    "descriptorBindingVariableDescriptorCount": true,
+                    "runtimeDescriptorArray": true
+                },
+                "VkPhysicalDevice8BitStorageFeatures": {
+                    "storageBuffer8BitAccess": true,
+                    "uniformAndStorageBuffer8BitAccess": true,
+                    "storagePushConstant8": true
+                },
+                "VkPhysicalDeviceVulkanMemoryModelFeatures": {
+                    "vulkanMemoryModel": true,
+                    "vulkanMemoryModelDeviceScope": true,
+                    "vulkanMemoryModelAvailabilityVisibilityChains": true
+                },
+                "VkPhysicalDeviceShaderAtomicInt64Features": {
+                    "shaderBufferInt64Atomics": true,
+                    "shaderSharedInt64Atomics": true
+                },
+                "VkPhysicalDeviceShaderAtomicFloatFeaturesEXT": {
+                    "shaderBufferFloat32Atomics": true,
+                    "shaderBufferFloat32AtomicAdd": true,
+                    "shaderBufferFloat64Atomics": true,
+                    "shaderBufferFloat64AtomicAdd": true,
+                    "shaderSharedFloat32Atomics": true,
+                    "shaderSharedFloat32AtomicAdd": true,
+                    "shaderSharedFloat64Atomics": true,
+                    "shaderSharedFloat64AtomicAdd": true,
+                    "shaderImageFloat32Atomics": true,
+                    "shaderImageFloat32AtomicAdd": true,
+                    "sparseImageFloat32Atomics": true,
+                    "sparseImageFloat32AtomicAdd": true
+                },
+                "VkPhysicalDeviceShaderAtomicFloat2FeaturesEXT": {
+                    "shaderBufferFloat16Atomics": true,
+                    "shaderBufferFloat16AtomicAdd": true,
+                    "shaderBufferFloat16AtomicMinMax": true,
+                    "shaderBufferFloat32AtomicMinMax": true,
+                    "shaderBufferFloat64AtomicMinMax": true,
+                    "shaderSharedFloat16Atomics": true,
+                    "shaderSharedFloat16AtomicAdd": true,
+                    "shaderSharedFloat16AtomicMinMax": true,
+                    "shaderSharedFloat32AtomicMinMax": true,
+                    "shaderSharedFloat64AtomicMinMax": true,
+                    "shaderImageFloat32AtomicMinMax": true,
+                    "sparseImageFloat32AtomicMinMax": true
+                },
+                "VkPhysicalDeviceTransformFeedbackFeaturesEXT": {
+                    "transformFeedback": true,
+                    "geometryStreams": true
+                },
+                "VkPhysicalDeviceComputeShaderDerivativesFeaturesNV": {
+                    "computeDerivativeGroupQuads": true,
+                    "computeDerivativeGroupLinear": true
+                },
+                "VkPhysicalDeviceShaderImageFootprintFeaturesNV": {
+                    "imageFootprint": true
+                },
+                "VkPhysicalDeviceShadingRateImageFeaturesNV": {
+                    "shadingRateImage": true,
+                    "shadingRateCoarseSampleOrder": true
+                },
+                "VkPhysicalDeviceInvocationMaskFeaturesHUAWEI": {
+                    "invocationMask": true
+                },
+                "VkPhysicalDeviceMeshShaderFeaturesNV": {
+                    "taskShader": true,
+                    "meshShader": true
+                },
+                "VkPhysicalDeviceMeshShaderFeaturesEXT": {
+                    "taskShader": true,
+                    "meshShader": true,
+                    "multiviewMeshShader": true,
+                    "primitiveFragmentShadingRateMeshShader": true,
+                    "meshShaderQueries": true
+                },
+                "VkPhysicalDeviceRayTracingPipelineFeaturesKHR": {
+                    "rayTracingPipeline": true,
+                    "rayTracingPipelineShaderGroupHandleCaptureReplay": true,
+                    "rayTracingPipelineShaderGroupHandleCaptureReplayMixed": true,
+                    "rayTracingPipelineTraceRaysIndirect": true,
+                    "rayTraversalPrimitiveCulling": true
+                },
+                "VkPhysicalDeviceRayQueryFeaturesKHR": {
+                    "rayQuery": true
+                },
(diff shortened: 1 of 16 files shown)


tests/spirv_hopper/pass_through_shaders.cpp
```diff
@@ -0,0 +1,401 @@
...
+    shader->setEnvClient(glslang::EShClientVulkan, glslang::EShTargetVulkan_1_3);
+
+    const char* shader_strings = source.c_str();
+    const int shader_length = static_cast<int>(source.length());
+    shader->setStringsWithLengths(&shader_strings, &shader_length, 1);
+    EShMessages messages = EShMsgDefault;
+    if (!shader->parse(&DefaultTBuiltInResource, 110, false, messages)) {
+        std::cout << shader->getInfoLog() << std::endl;
+        std::cout << source << std::endl;
+        return false;
```
artem-lunarg, line 386: `shader` probably leaks here

tests/spirv_hopper/main.cpp
```diff
@@ -0,0 +1,120 @@
...
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+#include <fstream>
+#include <string>
+#include <iostream>
+#include <filesystem>
+namespace fs = std::filesystem;
```
juan-lunarg, line 16:
```suggestion
```

tests/spirv_hopper/external/spirv_reflect.cpp
```diff
@@ -0,0 +1,4 @@
+//
+// This file exists to force compiling spirv_reflect.c as C++.
+//
+#include "spirv_reflect.c"
```
juan-lunarg, line 4: Change the file extension of spirv_reflect.cpp. That's what we did with xxhash.c -> xxhash.cpp

tests/spirv_hopper/spirv_hopper.cpp
```diff
@@ -0,0 +1,491 @@
...
+        std::vector<VkAttachmentReference> input_attachment_reference;
+        std::vector<VkAttachmentDescription> input_attachment_description;
+        for (uint32_t i = 0; i < input_attachments.size(); i++) {
+            // input_attachment_reference.push_back({input_attachments[i].input_attachment_index, VK_IMAGE_LAYOUT_GENERAL});
+            input_attachment_reference.push_back({0, VK_IMAGE_LAYOUT_GENERAL});
+        }
+        subpass_dscription.inputAttachmentCount = static_cast<uint32_t>(input_attachment_reference.size());
+        subpass_dscription.pInputAttachments = input_attachment_reference.data();
+
+        VkSubpassDependency dependency = {};
```
artem-lunarg, line 391: This dependency looks fine but I wonder if there was something that forced to add it? It looks the render pass specifies layout transition from UNDEFINED to GENERAL and in that case we have implicit EXTERNAL dependency.

tests/spirv_hopper/external/CMakeLists.txt
juan-lunarg, line 1: Please add a README.md in this directory where these files came from since there isn't a known_good.json

tests/spirv_hopper/main.cpp
```diff
@@ -0,0 +1,120 @@
...
+            spirv_file.close();
+
+            if (!IsValidSPIRV(file_size, static_cast<const char*>(spirv_data))) {
+                std::cout << "Warning: File " << file << " is not a valid SPIR-V File - skipping" << std::endl;
+                failed_files.push_back(file);
+            } else {
+                vk.is_valid = true;
+                Hopper hopper(vk, file_size, spirv_data);
+
+                std::cout << "[" << files_ran_count++ << "|" << files_list_count << "] Running " << file << std::endl;
```
artem-lunarg, line 97: nit: Just to raise awareness. The difference between '\n' and std::endl is that the latter additionally flushes the stream. When a lot of output is performed on slow output devices (like Windows console or write to file) this can have measurable performance impact. Hard to say for this specific case.

juan-lunarg: Agree it's best practice to only use std::endl when you actually need it.

tests/spirv_hopper/pass_through_shaders.cpp
```diff
@@ -0,0 +1,401 @@
...
+        shader += "var_" + std::to_string(variable->location) + "_" + std::to_string(component);
+        shader += "[]";
+        shader += ";\n";
+    }
+    shader += "void main() { }";
+    return BuildPassThroughShader(shader, VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT);
+}
+
+static const TBuiltInResource DefaultTBuiltInResource = {
+    /* .MaxLights = */ 32,
```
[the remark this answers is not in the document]
spencer-lunarg: went the `static TBuiltInResource InitResources()` approuch

tests/spirv_hopper/pass_through_shaders.cpp
spencer-lunarg: 1. We should never be getting here, this would mean there is an error in `spirv-hopper`

2. I don't know if it leaks, I think it is all scoped on this stack in this function anyway

tests/device_profiles/spirv_hopper_profile.json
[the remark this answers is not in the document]
spencer-lunarg: No (I can explain more in the README.md)

1. But there are limits we don't actually want set to Max UINT32_T in `max_profile.json` (because we need to test above the limit)
2. Some features we turn to `false` in  `max_profile.json` in order to make the tests fail because it doesn't have it
3. If we decide to move this out of VVL, its a small and portable profile

tests/spirv_hopper/external/spirv_reflect.cpp
spencer-lunarg: ya, I guess that is an advantage of checking it in

tests/spirv_hopper/main.cpp
spencer-lunarg: #TheMoreYouKnow - Will update

tests/spirv_hopper/spirv_hopper.cpp
spencer-lunarg: ok, good catch, this is not needed at all to still be valid, removed

tests/spirv_hopper/main.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+    std::vector<std::string> file_list;
+    std::vector<std::string> failed_files;
+
+    if (!std::filesystem::is_directory(argv[1])) {
+        file_list.push_back(argv[1]);
+    } else {
+        for (auto const& dir_entry : std::filesystem::recursive_directory_iterator(argv[1])) {
+            if (std::filesystem::is_regular_file(dir_entry)) {
+                file_list.push_back(dir_entry.path().c_str());
```
arno-lunarg, line 52:
```suggestion
                file_list.push_back(dir_entry.path().string());
```

tests/spirv_hopper/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+    spirv_hopper.cpp
+    pass_through_shaders.cpp
+)
+
+target_compile_definitions(spirv-hopper PRIVATE SPIRV_REFLECT_USE_SYSTEM_SPIRV_H=1)
+
+target_link_libraries(spirv-hopper PUBLIC
+    Vulkan::Headers
+    glslang::SPIRV
+    dl
```
arno-lunarg, line 33: `dl` Not found on Windows

arno-lunarg:
Regarding the `dl` lib: since it seems needed for the ubuntu 20 + gcc 9 build, maybe continue to link it on linux?

tests/spirv_hopper/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+target_sources(spirv-hopper PRIVATE
+    main.cpp
+    vulkan_object.h
+    vulkan_object.cpp
+    spirv_hopper.h
+    spirv_hopper.cpp
+    pass_through_shaders.cpp
+)
+
+target_compile_definitions(spirv-hopper PRIVATE SPIRV_REFLECT_USE_SYSTEM_SPIRV_H=1)
```
juan-lunarg, line 28: Remove this line entirely.

```suggestion
```

tests/spirv_hopper/README.md
```diff
@@ -0,0 +1,58 @@
...
+- Uses [SPIRV-Reflect](https://github.com/KhronosGroup/SPIRV-Reflect) to get the basic interface information
+- Creates everything needed to make a valid `VkPipeline`
+    - `VkPipelineLayout`
+    - Descriptor Sets
+    - Passthrough shaders
+- `vkCreate*Pipelines()`
+
+## Goals
+
+This tool is designed to be simple, its main job is to help catch issues in Shader Validation by allowing us to run against MANY shaders quickly. Paired with a database of SPIR-V binaries from a various SPIR-V generated tools (not just `glslang`), we can catch bugs/crashes at a per-commit testing level
```
juan-lunarg, line 27:
```suggestion
This tool is designed to be simple, its main job is to help catch issues in Shader Validation by allowing us to run against MANY shaders quickly. Paired with a database of SPIR-V binaries from various SPIR-V generated tools (not just `glslang`), we can catch bugs/crashes at a per-commit testing level
```

tests/spirv_hopper/external/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+    spirv_reflect.cpp
+)
+
+if(${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
+    target_compile_options(spirv-reflect PUBLIC
+        -Wno-sign-conversion
+    )
+endif()
+
+target_compile_definitions(spirv-reflect PUBLIC SPIRV_REFLECT_USE_SYSTEM_SPIRV_H=1)
```
juan-lunarg, line 30:
```suggestion
target_compile_definitions(spirv-reflect PUBLIC SPIRV_REFLECT_USE_SYSTEM_SPIRV_H)
```

juan-lunarg:
https://kubasejdak.com/modern-cmake-is-like-inheritance

Please review the above documentation for understanding PRIVATE vs PUBLIC.

tests/spirv_hopper/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+    spirv_hopper.h
+    spirv_hopper.cpp
+    pass_through_shaders.cpp
+)
+
+target_link_libraries(spirv-hopper PRIVATE
+    Vulkan::Headers
+    glslang::SPIRV
+    VkLayer_utils
+    ${CMAKE_DL_LIBS}
```
juan-lunarg, line 32: I'd prefer `VkLayer_utils` link against ${CMAKE_DL_LIBS} since `lvt_function_pointers.cpp` belongs to `VkLayer_utils` and is what actually requires the ${CMAKE_DL_LIBS} functionality.

That way we also only link ${CMAKE_DL_LIBS} once in the CMake code as well.

tests/spirv_hopper/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+add_subdirectory(external)
+add_executable(spirv-hopper)
+
+target_sources(spirv-hopper PRIVATE
+    main.cpp
+    vulkan_object.h
+    vulkan_object.cpp
+    spirv_hopper.h
+    spirv_hopper.cpp
+    pass_through_shaders.cpp
```
juan-lunarg, line 25: Sort the source files listed here.

tests/device_profiles/spirv_hopper_profile.json
```diff
@@ -0,0 +1,1085 @@
+{
+    "$schema": "https://schema.khronos.org/vulkan/profiles-0.8.0-latest.json#",
+    "profiles": {
+        "MAX_PROFILE": {
```
juan-lunarg, line 4: Why does this say `MAX_PROFILE`?

tests/device_profiles/spirv_hopper_profile.json
juan-lunarg, line 1: Can we inherit from `max_profile.json`? IE override what we need to, but defer to `max_profile.json` when we don't specify? My intent being to reduce the size of this json and avoid duplicate efforts.

tests/spirv_hopper/main.cpp
```diff
@@ -0,0 +1,119 @@
...
+    } else if (0x07 == spirv_data[0] && 0x23 == spirv_data[1] && 0x02 == spirv_data[2] && 0x03 == spirv_data[3]) {
+        return true;  // Big Endianness
+    } else {
+        return false;
+    }
+}
+
+int main(int argc, char** argv) {
+    int result = 0;
+
```
juan-lunarg, line 35:
```suggestion
```

tests/spirv_hopper/main.cpp
```diff
@@ -0,0 +1,119 @@
...
+    if (argc != 2) {
+        std::cout << "Error:\n\tUsage: ./spirv-hopper [file | directory]\n";
+        return -1;
```
juan-lunarg, line 38:
```suggestion
        return EXIT_FAILURE;
```

tests/spirv_hopper/main.cpp
```diff
@@ -0,0 +1,119 @@
...
+    } else if (!std::filesystem::exists(argv[1])) {
+        std::cout << "Error: " << argv[1] << " Does not exists\n";
+        return -1;
```
juan-lunarg, line 41:
```suggestion
        return EXIT_FAILURE;
```

tests/spirv_hopper/main.cpp
```diff
@@ -0,0 +1,119 @@
...
+        std::cout << "\nRESULT: Success!\n";
+    } else {
+        result = -1;
+        std::cout << "\nRESULT: Failure!\n";
+        std::cout << "\nThe following tests failed:\n";
+        for (const auto& file : failed_files) {
+            std::cout << "\t" << file << "\n";
+        }
+    }
+    return result;
```
juan-lunarg, line 118:
```suggestion
    if (!failed_files.empty()) {
        std::cout << "\nRESULT: Failure!\n";
        std::cout << "\nThe following tests failed:\n";
        for (const auto& file : failed_files) {
            std::cout << "\t" << file << "\n";
        }
        return EXIT_FAILURE;
    }
    
    std::cout << "\nRESULT: Success!\n";
```

tests/spirv_hopper/main.cpp
```diff
@@ -0,0 +1,119 @@
...
+                std::cout << "Error: Unable to open the file " << file << "\n";
+                failed_files.push_back(file);
+                continue;
+            }
+            spirv_file.seekg(0, spirv_file.end);
+            const size_t file_size = static_cast<size_t>(spirv_file.tellg());
+            spirv_file.seekg(0, spirv_file.beg);
+
+            void* spirv_data = malloc(file_size);
+            if (spirv_data == nullptr) {
```
juan-lunarg, line 79: Use std::vector to avoid manually calling malloc/free

tests/spirv_hopper/pass_through_shaders.cpp
```diff
@@ -0,0 +1,405 @@
...
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+#include "spirv_hopper.h"
+#include "glslang/SPIRV/GlslangToSpv.h"
+#include <iostream>
+
+static constexpr bool IsFloatFormat(SpvReflectFormat format) {
```
juan-lunarg, line 17: Do we not already have this functionality in vvl or vll testing?

juan-lunarg (changes requested):

tests/spirv_hopper/pass_through_shaders.cpp
spencer-lunarg: This is `SPV_REFLECT_FORMAT` not `VkFormat`, its a subset the library decided

Honestly, my goal was to replace this with a fix in SPIRV-Reflect to have this, as this function depends on SPIRV-Reflect,  unlike the other VVL util

tests/device_profiles/spirv_hopper_profile.json
spencer-lunarg: > Can we inherit from max_profile.json

We don't want to inherit anything from `max_profile`, the point of this profile is to have different values

> but defer to max_profile.json when we don't specify? 

If we leave it blank, we defer it to false (or whatever default Profile layers give)

> My intent being to reduce the size of this json and avoid duplicate efforts.

This json file in theory could just be generated from the `vk.xml` using the `<spirvextensions>` and `<spirvcapabilities>` tags (Which would be more ideal and what I originally wanted)

As for size, this is by far the smallest profile we have now (after I just went in now and trimmed more stuff out)

```
475K max_profile.json
216K pixel_6_mali.json
179K nvidia_tegra_x1.json
171K pixel_6_adreno.json
 41K spirv_hopper_profile.json
```

arno-lunarg (approved):

tests/spirv_hopper/main.cpp
```diff
@@ -0,0 +1,108 @@
...
+        std::cout << "\nRESULT: Failure!\n";
+        std::cout << "\nThe following tests failed:\n";
+        for (const auto& file : failed_files) {
+            std::cout << "\t" << file << "\n";
+        }
+        return EXIT_FAILURE;
+    }
+
+    std::cout << "\nRESULT: Success!\n";
+    return 0;
```
juan-lunarg, line 107: 4) The body of the main function does not need to contain the [return statement](https://en.cppreference.com/w/cpp/language/return): if control reaches the end of main without encountering a return statement, the effect is that of executing return 0;.

```suggestion
```

tests/spirv_hopper/pass_through_shaders.cpp
```diff
@@ -0,0 +1,405 @@
...
+            glsl_stage = EShLangTessControl;
+            break;
+        case VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT:
+            glsl_stage = EShLangTessEvaluation;
+            break;
+        default:
+            break;
+    }
+
+    glslang::TShader* shader = new glslang::TShader(glsl_stage);
```
juan-lunarg, line 379: Use a smart pointer. From a cursory glance I don't see where this memory is freed.

juan-lunarg (approved):

tests/spirv_hopper/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+### Only handles single Entry Point
+
+For getting things up and running, only single Entry Points are tested.
+
+A future goal is to take each entry point and run each as there own "sub-run"
+
+### Single threaded
+
+Looked into threading the work, but seems to be more investigation how to properly thread the glslang instance at runtime.
```
ncesario-lunarg, line 47: "...threading the work..." is a bit vague. Maybe just indicate that `SPIRV-Hopper` is single-threaded, but multi-threading is left for future work?

tests/spirv_hopper/external/README.md
```diff
@@ -0,0 +1,17 @@
+# SPIRV-Reflect
+
+These files are taken from https://github.com/KhronosGroup/SPIRV-Reflect
+
+The reasons to check these in instead of adding to `known_good.json` are
+
+- This repo is not updated often
+- It is designed to be a files people include in their build
```
ncesario-lunarg, line 8:
```suggestion
- These files are designed for people to include in their build
```

ncesario-lunarg (approved):
Minor nits, but LGTM!

tests/spirv_hopper/pass_through_shaders.cpp
spencer-lunarg: there is a `delete shader;` this code was copied from `VkTestFramework::GLSLtoSPV(`, but agree smart pointer would be better (For both)

tests/spirv_hopper/external/README.md
```diff
@@ -0,0 +1,17 @@
+# SPIRV-Reflect
+
+These files are taken from https://github.com/KhronosGroup/SPIRV-Reflect
```
dneto0, line 3: Can we not do this please?

Bugfixes and potentially security fixes get stranded.

spencer-lunarg: Would it be better for the short term to hide this `SPIRV-Hopper` behind an extra Opt-In flag

juan-lunarg: The best solution is to upstream good CMake support IMO. I'd rather not add an extra Opt-In flag.

jeremyg-lunarg: > The best solution is to upstream good CMake support IMO. I'd rather not add an extra Opt-In flag.

Should we do the same for VMA?

jeremyg-lunarg: Also, this needs to be a new issue, not a discussion on a closed PR

PR #5914, opened 2023-05-26
Update LICENSE.txt

ben-clayton:
Include BSD license used in `layers/external/vma/vk_mem_alloc.h`

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -206,3 +206,25 @@ Any file without a license in it's source defaults to the repository Apache 2.0 
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+
+
+===========================================================================================
+File: layers/external/vma/vk_mem_alloc.h
+
+
+Copyright (c) 2017-2022 Advanced Micro Devices, Inc. All rights reserved.
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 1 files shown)


ben-clayton:
Required to make Google's license checking bots happy

juan-lunarg:
I don't know enough about licensing to be useful here.

Pinging @KarenGhavam-lunarG and @KhronosWebservices

dneto0 (commented):
Right. This will be necessary, since AMD's VMA code was copied into this repo.

(I was surprised to see that, but I don't know all the reasons, and it's not up to me.)

juan-lunarg:
> (I was surprised to see that, but I don't know all the reasons, and it's not up to me.)

TLDR: The chromium build, system package managers, and the current android build.

juan-lunarg (dismissed):

KarenGhavam-lunarG:
> Pinging @KarenGhavam-lunarG and @KhronosWebservices

No concerns here. This fix is correct. We need to detail the licenses in use in this repository.

outofcontrol:
The only concern I can think of is that we don't have a CLA for the BSD, so accepting changes to the file is problematic. Is it possible for AMD to accept a license change to Apache 2.0? Might be good also to get reuse setup to ensure we are tracking the copyright and license across all files in this repository.

KarenGhavam-lunarG:
>  Is it possible for AMD to accept a license change to Apache 2.0?

This would be problematic. And most likely this is not the only file/license beyond Apache 2.0 in the repository

> Might be good also to get reuse setup to ensure we are tracking the copyright and license across all files in this repository.

I am unfamiliar with reuse and what the implications of this would be.

KarenGhavam-lunarG:
IMHO, the questions about the CLA and reuse should not hold up merging of this PR. Rather a separate issue can be logged regarding these questions. This is due to the fact that there are other 3rd party files in this repository with licenses that are not Apache 2.0 (as noted in the license.txt file) and the questions around the CLA and reuse are not new or introduced by this PR.

outofcontrol:
> IMHO, the questions about the CLA and reuse should not hold up merging of this PR.

Agree.

juan-lunarg (approved):

ben-clayton:
Many thanks

PR #5916, opened 2023-05-26
ci: Fix linux performance

juan-lunarg:
1.) Fix ccache misses due to not using matrix.os
2.) Only run 1 job for ubuntu-20.04

### DIFF
--- a/.github/workflows/vvl.yml
+++ b/.github/workflows/vvl.yml
@@ -37,11 +37,16 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-20.04, ubuntu-22.04]
+        os: [ubuntu-22.04]
         compiler: [{cc: gcc, cxx: g++}]
         config: [debug, release]
         cpp_std: [17]
         include:
+          # Test Ubuntu-20.04 release build works.
+          - os: ubuntu-20.04
+            compiler: {cc: gcc, cxx: g++}
+            config: release
+            cpp_std: 17
           # Test C++ 20 support
           - os: ubuntu-22.04
             compiler: {cc: gcc, cxx: g++}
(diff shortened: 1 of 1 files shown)


.github/workflows/vvl.yml
```diff
@@ -172,7 +176,7 @@ jobs:
             external/${{ matrix.config }}/Vulkan-Headers/registry
             external/${{ matrix.config }}/SPIRV-Headers/include
             external/${{ matrix.config }}/mimalloc/build/install
-          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ matrix.arch }}-${{ matrix.os }}-${{ matrix.config }}-${{ hashfiles('scripts/known_good.json') }}
+          key: ${{ matrix.os }}-build-${{ env.cache-name }}-${{ matrix.arch }}-${{ matrix.os }}-${{ matrix.config }}-${{ hashfiles('scripts/known_good.json') }}
```
juan-lunarg, line 179: This minor issue adds up tremendously. Adding an additional 30-50 minutes for building dependencies.

juan-lunarg: `runner.os` is equal to linux. Not the current runner image. Which meant the ubuntu-20 and ubuntu-22 caches were fighting each other.

ncesario-lunarg: :grimacing:

ncesario-lunarg (approved):

lunarpapillo:
Shouldn't line 72 also be modified?

juan-lunarg:
> Shouldn't line 72 also be modified?

Yes it should. I've removed all usage of `runner.os`.

spencer-lunarg (approved):

PR #5917, opened 2023-05-26
build: Fix Android apk build

aitor-lunarg:
Adds a step to add libraries in as Android likes them

### DIFF
--- a/build-android/build_all.sh
+++ b/build-android/build_all.sh
@@ -73,7 +73,8 @@ fi
 echo "Building Layer using ${ANDROID_STL_TYPE}";
 
 function create_APK() {
-    aapt package -f -M AndroidManifest.xml -I "$ANDROID_SDK_HOME/platforms/android-26/android.jar" -S res -F bin/$1-unaligned.apk bin/libs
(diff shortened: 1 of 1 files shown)


build-android/build_all.sh
```diff
@@ -73,7 +73,8 @@ fi
 echo "Building Layer using ${ANDROID_STL_TYPE}";
 
 function create_APK() {
-    aapt package -f -M AndroidManifest.xml -I "$ANDROID_SDK_HOME/platforms/android-26/android.jar" -S res -F bin/$1-unaligned.apk bin/libs
+    aapt package -f -M AndroidManifest.xml -I "$ANDROID_SDK_HOME/platforms/android-26/android.jar" -S res -F bin/$1-unaligned.apk
+    aapt add -f bin/$1-unaligned.apk bin/libs/*.so
```
juan-lunarg, line 77: Add comments for lines 76 and 77 similar to the code below

juan-lunarg (changes requested):
LGTM just needs comments

build-android/build_all.sh
aitor-lunarg: Yes, I'll wait for confirmation that this works and I'll update with comments

juan-lunarg: Please also revise the commit message to be more specific as well.

aitor-lunarg:
Won't solve issue

Issue #5911, opened 2023-05-26
Assert firing in `Instruction::GetConstantValue()`


ben-clayton:
This assertion is firing for a google-internal project:

https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/4155717b71a3c627a64c360de6f45b6fa6d072c9/layers/state_tracker/shader_instruction.cpp#L70-L73

The backtrace is:

```
Instruction::GetConstantValue()
SHADER_MODULE_STATE::GetTypeBitsSize()
PushConstantVariable::PushConstantVariable()
EntryPoint::GetResourceInterfaceVariables()
EntryPoint::EntryPoint()
SHADER_MODULE_STATE::StaticData::StaticData()
std::__u::__shared_ptr_emplace<>::__shared_ptr_emplace<>()
ValidationStateTracker::CreateShaderModuleState()
ValidationStateTracker::PostCallRecordCreateShaderModule()
vulkan_layer_chassis::CreateShaderModule()
```

`Opcode()` is a `OpSpecConstant`.

I'm currently investigating, and will attempt to update this issue with any additional details I can find. Any assistance here would be greatly appreciated.

dneto0:
@spencer-lunarg has been looking at shader validation in VVL.

dneto0:
Looks like this is the culprit:

In SHADER_MODULE_STATE::GetTypeBitsSize, the length of an array is assumed to be an OpConstant.
But array lengths can be spec constants.

ben-clayton:
I suspect the change that caused this is: https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/05f5efa06976e7255acbddaa6efe3d21488e5771#diff-1177be4184e687396ca5956a0670bb9f46b235ef1e5a10ca0d7e1704dd43124b

This change added the lines:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/4155717b71a3c627a64c360de6f45b6fa6d072c9/layers/state_tracker/shader_module.cpp#L456-L458

The assert assumes that `CreateFoldSpecConstantOpAndCompositePass` has been run - but I haven't found anything in the call stack to assume that specialization constants have been folded.

ben-clayton:
Looking at the offending shader (I'd have to cut through red-tape to share) the issue seems to be related to:

```glsl
layout (constant_id = 2) const int my_array_size = 1;

layout (push_constant) uniform my_buf {
[...]

PR #5919, opened 2023-05-27
layers: Fix crash with push constants using spec const

spencer-lunarg:
This is a temporary fix for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5911

It might still produce false positive  for `VUID-VkComputePipelineCreateInfo-layout-07987`, but decide that was **much better** then crashing from an assert.

I plan to put together a proper fix, but that will require a more work as we need more tests as it will be invasive, but want to get things unblocked for now

### DIFF
--- a/layers/core_checks/cc_shader.cpp
+++ b/layers/core_checks/cc_shader.cpp
@@ -224,6 +224,11 @@ bool CoreChecks::ValidatePushConstantUsage(const PIPELINE_STATE &pipeline, const
                                            const EntryPoint &entrypoint) const {
     bool skip = false;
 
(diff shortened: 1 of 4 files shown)


ben-clayton:
Do you think false-negative would be a better temporary solution than false-positive?
I'm concerned projects hitting the assert will still not be able to roll VVL as we'll then be blocked by the false-positive.

spencer-lunarg:
> Do you think false-negative would be a better temporary solution than false-positive?

Agree :100: - I updated PR to just skip checking for push constants ranges if it detects the use of spec constants

ben-clayton:
Thank you!

arno-lunarg (approved):

PR #5920, opened 2023-05-29
layers: Define when to use IsExtEnabled

spencer-lunarg:
This was my first pass attempt to try and unify how we use `IsExtEnabled`

Decided to write the "documentation" in `IsExtEnabled` itself

### DIFF
--- a/layers/core_checks/cc_descriptor.cpp
+++ b/layers/core_checks/cc_descriptor.cpp
@@ -452,7 +452,7 @@ bool CoreChecks::PreCallValidateCreateDescriptorSetLayout(VkDevice device, const
 
         if ((binding_info.descriptorType == VK_DESCRIPTOR_TYPE_SAMPLER ||
              binding_info.descriptorType == VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER) &&
-            binding_info.pImmutableSamplers && IsExtEnabled(device_extensions.vk_ext_custom_border_color)) {
+            binding_info.pImmutableSamplers) {
             for (uint32_t j = 0; j < binding_info.descriptorCount; j++) {
                 auto sampler_state = Get<SAMPLER_STATE>(binding_info.pImmutableSamplers[j]);
                 if (sampler_state && (sampler_state->createInfo.borderColor == VK_BORDER_COLOR_INT_CUSTOM_EXT ||
@@ -5210,6 +5210,7 @@ bool CoreChecks::PreCallValidateCreatePipelineLayout(VkDevice device, const VkPi
                          phys_dev_ext_props.ray_tracing_props_nv.maxDescriptorSetAccelerationStructures);
     }
 
+    // Extension exposes new properties limits
     if (IsExtEnabled(device_extensions.vk_ext_descriptor_indexing)) {
         // XXX TODO: replace with correct VU messages
 
@@ -5404,6 +5405,7 @@ bool CoreChecks::PreCallValidateCreatePipelineLayout(VkDevice device, const VkPi
         }
     }
 
+    // Extension exposes new properties limits
     if (IsExtEnabled(device_extensions.vk_ext_fragment_density_map2)) {
         uint32_t sum_subsampled_samplers = 0;
         for (const auto &dsl : set_layouts) {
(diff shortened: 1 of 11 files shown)


layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -2610,7 +2602,6 @@ bool CoreChecks::ValidatePipelineDrawtimeState(const LAST_BOUND_STATE &last_boun
         }
     }
 
-    if (IsExtEnabled(device_extensions.vk_ext_primitives_generated_query)) {
```
artem-lunarg, line 2613: I explained this to myself like this: we do not need to check extension enable status when checking extension's feature enable status because, if feature is enabled, then we already validated that extension is enabled. 

Can this be a candidate to `Times to NOT use it` to mention feature checks?

layers/core_checks/cc_render_pass.cpp
```diff
@@ -4544,16 +4541,31 @@ bool CoreChecks::PreCallValidateCreateFramebuffer(VkDevice device, const VkFrame
...
-                    if ((rpci->subpassCount > 0) && (rpci->pSubpasses[0].viewMask != 0)) {
-                        for (uint32_t i = 0; i < rpci->subpassCount; ++i) {
-                            const VkSubpassDescriptionDepthStencilResolve *depth_stencil_resolve =
-                                LvlFindInChain<VkSubpassDescriptionDepthStencilResolve>(rpci->pSubpasses[i].pNext);
-                            uint32_t view_bits = rpci->pSubpasses[i].viewMask;
-                            int highest_view_bit = MostSignificantBit(view_bits);
-
-                            for (uint32_t j = 0; j < rpci->pSubpasses[i].colorAttachmentCount; ++j) {
-                                attachment_index = rpci->pSubpasses[i].pColorAttachments[j].attachment;
+                if ((rpci->subpassCount > 0) && (rpci->pSubpasses[0].viewMask != 0)) {
```
artem-lunarg, line 4544: This looks like one more use case. We check variable `viewMask` that was validated earlier against multiview support, if I got it right. Should we mention this in the 'when not to' section?

layers/core_checks/cc_pipeline_graphics.cpp
spencer-lunarg: yes, the feature can't be enabled if the extension is not

Can add that

layers/core_checks/cc_render_pass.cpp
spencer-lunarg: 1. Yes, can add
2. I will also add a comment here that `VUID-VkSubpassDescription2-multiview-06558` forces `viewMask` to be zero if not using multiView

artem-lunarg (approved):
LGTM! Few questions but otherwise it's good to have this specified in the documentation.

Some feedback. I like this approach more now, compared to having more extensions checks. At the first sight extension check creates some modularity and without this you need to do some mental work which relies on the understanding of how the entire system works. 

I had some not very nice past experiences where in a similar way we had checks at the beginning of the function or block of logic, for example, if some value is not null. Then even in cases when system provided a guarantee the value is not null you still had to do that check because all other code did that and you were not sure anymore. That resulted in worse understanding of the system which had reinforcement effect that you started to check for nulls even more, because you weren't sure about how things work. In the end the mental load was much higher and you were less confident comparing to situation where you know how the system works in more wider context and explicitly take this into account (it was not possible to compare both approaches on the same project, so it's experiences from different projects).

layers/core_checks/cc_descriptor.cpp
```diff
@@ -5210,6 +5210,7 @@ bool CoreChecks::PreCallValidateCreatePipelineLayout(VkDevice device, const VkPi
                          phys_dev_ext_props.ray_tracing_props_nv.maxDescriptorSetAccelerationStructures);
     }
 
+    // Extension exposese new properties limits
```
juan-lunarg, line 5213:
```suggestion
    // Extension exposes new properties limits
```

layers/core_checks/cc_descriptor.cpp
```diff
@@ -5404,6 +5405,7 @@ bool CoreChecks::PreCallValidateCreatePipelineLayout(VkDevice device, const VkPi
         }
     }
 
+    // Extension exposese new properties limits
```
juan-lunarg, line 5408:
```suggestion
    // Extension exposes new properties limits
```

juan-lunarg (approved):

arno-lunarg (approved):
LGTM, thank you for writing down the reasoning behind the use of this function!

PR #5923, opened 2023-05-30
layers: Fix printing wrong SPIR-V type for builtin

spencer-lunarg:
`BuiltIns` are a type of `Decoration` - I realize when I added this I provided the wrong `string_Spv` helper and was not printing out correct, useful values

------

Before

```
[ VUID-RuntimeSpirv-OpVariable-08746 ] ... Mistmatch in BuiltIn blocks:
 VK_SHADER_STAGE_VERTEX_BIT Output Block {
	0: RelaxedPrecision
	1: SpecId
}
VK_SHADER_STAGE_GEOMETRY_BIT Input Block {
	0: RelaxedPrecision
	1: SpecId
	2: BufferBlock
}
```

now

```
[ VUID-RuntimeSpirv-OpVariable-08746 ] ... Mistmatch in BuiltIn blocks:
 VK_SHADER_STAGE_VERTEX_BIT Output Block {
	0: Position
	1: PointSize
}
VK_SHADER_STAGE_GEOMETRY_BIT Input Block {
	0: Position
	1: PointSize
	2: ClipDistance
}
```


juan-lunarg (approved):

layers/vulkan/generated/spirv_grammar_helper.h
```diff
@@ -52,5 +52,6 @@ const char* string_SpvOpcode(uint32_t opcode);
 const char* string_SpvStorageClass(uint32_t storage_class);
 const char* string_SpvExecutionModel(uint32_t execution_model);
 const char* string_SpvDecoration(uint32_t decoration);
+const char* string_SpvBuiltIn(uint32_t built_in);
```
juan-lunarg, line 55: Minor nitpick is the lack of type safety. IE this function takes a uint32_t.

But that is seems to be consistent with the other `string_Spv` functions.

spencer-lunarg: SPIR-V is just a stream of `uint32_t` words by definition, not really sure what other type to use

spencer-lunarg: so I guess all of this SPIR-V code is based off what is parsed from the binary, we are never just going to go `string_SpvBuiltIn(89)`, it will always be some result of the `std::vector<uint32_t> SHADER_MODULE_STATE::words_` object that was found

I guess from my experience with other SPIR-V related projects, wrapping the `uint32_t` to something like `SpirvToken` makes me assume there will be more then just the `uint32_t` (which I guess is what our `Instruction` is trying to do)

PR #5929, opened 2023-05-30
Enable Win32 testing on CI

juan-lunarg:
- Fix `NegativeQuery.PerformanceCreation`
- Enables FullScreen testing on MockICD only. Also made naming consistent Fullscreen -> FullScreen to make it easier to run tests.
- Enable CI testing

### DIFF
--- a/tests/negative/query.cpp
+++ b/tests/negative/query.cpp
@@ -79,6 +79,7 @@ TEST_F(NegativeQuery, PerformanceCreation) {
     // Invalid counter indices
     counterIndices.push_back(counters.size());
     perf_query_pool_ci.counterIndexCount++;
+    perf_query_pool_ci.pCounterIndices = counterIndices.data();
     m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkQueryPoolPerformanceCreateInfoKHR-pCounterIndices-03321");
     query_pool.init(*m_device, query_pool_ci);
     m_errorMonitor->VerifyFound();
--- a/tests/negative/wsi.cpp
+++ b/tests/negative/wsi.cpp
@@ -2772,7 +2772,7 @@ TEST_F(NegativeWsi, SwapchainMaintenance1ExtensionRelease) {
     vk::QueueWaitIdle(m_device->m_queue);
 }
 
-#if defined(VVL_TESTS_ENABLE_EXCLUSIVE_FULLSCREEN) && defined(VK_USE_PLATFORM_WIN32_KHR)
+#if defined(VK_USE_PLATFORM_WIN32_KHR)
 TEST_F(NegativeWsi, AcquireFullScreenExclusiveModeEXT) {
     TEST_DESCRIPTION("Test vkAcquireFullScreenExclusiveModeEXT.");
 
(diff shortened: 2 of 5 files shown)


tests/negative/query.cpp
```diff
@@ -79,6 +79,7 @@ TEST_F(NegativeQuery, PerformanceCreation) {
     // Invalid counter indices
     counterIndices.push_back(counters.size());
     perf_query_pool_ci.counterIndexCount++;
+    perf_query_pool_ci.pCounterIndices = counterIndices.data();
```
juan-lunarg, line 82: On line 65 we assign `pCounterIndicies` to `&counterIndices[0]` however on line 80 we push_back.

On MSVC this resulted in references to std::vector being invalidated. Causing testing to fail on GHA.

tests/negative/wsi.cpp
```diff
@@ -2772,7 +2772,7 @@ TEST_F(NegativeWsi, SwapchainMaintenance1ExtensionRelease) {
     vk::QueueWaitIdle(m_device->m_queue);
 }
 
-#if defined(VVL_TESTS_ENABLE_EXCLUSIVE_FULLSCREEN) && defined(VK_USE_PLATFORM_WIN32_KHR)
```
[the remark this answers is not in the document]
arno-lunarg: Nice, more coverage!!!

tests/negative/query.cpp
arno-lunarg: Nice catch!

tests/positive/wsi.cpp
```diff
@@ -188,6 +188,10 @@ TEST_F(PositiveWsi, GetPhysicalDeviceSurfaceCapabilities2KHRWithFullscreenEXT) {
         GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported.";
     }
 
+    if (!IsPlatform(kMockICD)) {
+        GTEST_SKIP() << "Only run test MockICD due to CI stability";
+    }
+
```
arno-lunarg, line 194:
```suggestion
#ifndef VVL_TESTS_ENABLE_EXCLUSIVE_FULLSCREEN
    if (!IsPlatform(kMockICD)) {
        GTEST_SKIP() << "Only run test MockICD due to CI stability";
    }
#endif
```
So that running this test locally is simpler than having to remove this `if IsPlatform`?
Same for other test

arno-lunarg (approved):
LGTM! One query

tests/positive/wsi.cpp
arno-lunarg: nothing, the dev has to explicitly supply it. I personally do so using the Visual Studio UI

juan-lunarg: I agree with the intent. But not with the solution. That just solves the problem for fullscreen testing.

See `VkAmdBestPracticesLayerTest.TooManyPipelines` as an example.

We need a general purpose solution which would be better done in another PR.

When I see `VVL_TESTS_ENABLE_EXCLUSIVE_FULLSCREEN` I assume there is a CMake option to go along with it. Having to manually define the pre-processor define is non-obvious. Especially to contributors not familiar with our codebase.

arno-lunarg: That's a good point
Then if a CMake option is ok for you, I am going to add it

juan-lunarg: > Then if a CMake option is ok for you, I am going to add it

A CMake option would work. Ideally something like a CLI parameter or an environment variable.

As shown in this PR special build configurations like `LOADER_DISABLE_DYNAMIC_LIBRARY_UNLOADING` and `LOADER_USE_UNSAFE_FILE_SEARCH` make things more complicated.

juan-lunarg:
@christophe-lunarg you can now run CI testing locally on Windows.

Open a visual studio developer command prompt and then run the following
```
rmdir build
rmdir external
python scripts/github_ci_win_linux.py --build --config debug
python scripts/github_ci_win_linux.py --test
```

PR #5930, opened 2023-05-30
Inherit ignoring color attachments from libraries

Tony-LunarG:
Fixes a crash running CTS with Best Practices.  Those familiar with GPLs (not I) are encouraged to suggest a better solution, if needed.


layers/state_tracker/pipeline_state.cpp
```diff
@@ -396,8 +396,14 @@ static bool UsesPipelineRobustness(const void *pNext, const PIPELINE_STATE &pipe
     return result;
 }
 
-static bool IgnoreColorAttachments(PIPELINE_STATE &pipe_state) {
+static bool IgnoreColorAttachments(const ValidationStateTracker *state_data, PIPELINE_STATE &pipe_state) {
     bool ignore = false;
+    if (pipe_state.library_create_info) {
+        for (int i = 0; i < pipe_state.library_create_info->libraryCount; i++) {
```
juan-lunarg, line 402: nitpick

```suggestion
        for (uint32_t i = 0; i < pipe_state.library_create_info->libraryCount; i++) {
```

juan-lunarg (approved):

layers/state_tracker/pipeline_state.cpp
```diff
@@ -396,8 +396,14 @@ static bool UsesPipelineRobustness(const void *pNext, const PIPELINE_STATE &pipe
     return result;
 }
 
-static bool IgnoreColorAttachments(PIPELINE_STATE &pipe_state) {
+static bool IgnoreColorAttachments(const ValidationStateTracker *state_data, PIPELINE_STATE &pipe_state) {
     bool ignore = false;
+    if (pipe_state.library_create_info) {
```
spencer-lunarg, line 401: Can we add a comment here, from first look, I have no idea why we would be crashing if linking against another library also ignoring it

Tony-LunarG: The crash is when we don't ignore the color attachments as pAttachments is null.  If the libraries are ignoring attachments, the pipeline created from them should as well.  At least that's my thinking.  I'll add a comment.

spencer-lunarg (approved):
This is good for now, there is a lot needed for both Best Practice and GPL that can be done

PR #5934, opened 2023-05-31
Label framebuffer VUID correctly for non Imageless Framebuffers

spencer-lunarg:
Adds ` VUID-VkFramebufferCreateInfo-flags-02778` for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5842

In adding the test, noticed we were actually getting a `VUID_Undefined`. I realize both Stateless and Object Tracker were actually doing the validation and it never made it to core

### DIFF
--- a/layers/core_checks/cc_render_pass.cpp
+++ b/layers/core_checks/cc_render_pass.cpp
@@ -4117,108 +4117,100 @@ bool CoreChecks::PreCallValidateCreateFramebuffer(VkDevice device, const VkFrame
                 const VkImageView *image_views = pCreateInfo->pAttachments;
                 for (uint32_t i = 0; i < pCreateInfo->attachmentCount; ++i) {
                     auto view_state = Get<IMAGE_VIEW_STATE>(image_views[i]);
-                    if (view_state == nullptr) {
-                        skip |= LogError(
-                            image_views[i], "VUID-VkFramebufferCreateInfo-flags-02778",
-                            "vkCreateFramebuffer(): VkFramebufferCreateInfo attachment #%u is not a valid VkImageView.", i);
-                    } else {
-                        auto &ivci = view_state->create_info;
-                        auto &subresource_range = view_state->normalized_subresource_range;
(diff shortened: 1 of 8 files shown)


layers/core_checks/cc_render_pass.cpp
```diff
@@ -4117,108 +4117,100 @@ bool CoreChecks::PreCallValidateCreateFramebuffer(VkDevice device, const VkFrame
                 const VkImageView *image_views = pCreateInfo->pAttachments;
                 for (uint32_t i = 0; i < pCreateInfo->attachmentCount; ++i) {
                     auto view_state = Get<IMAGE_VIEW_STATE>(image_views[i]);
-                    if (view_state == nullptr) {
```
spencer-lunarg, line 4120: I just removed this `if` (and everything is indented back)

We don't pointer check this anywhere else in the file, tests show it will never be null here (unless something more crazy is broken)

tests/negative/renderpass.cpp
```diff
@@ -2336,6 +2336,53 @@ TEST_F(NegativeRenderPass, Framebuffer) {
...
+TEST_F(NegativeRenderPass, FramebufferAttachmentPointers) {
+    TEST_DESCRIPTION("pAttachments points to valid objects for the Framebuffer creation");
+
+    ASSERT_NO_FATAL_FAILURE(Init());
+    ASSERT_NO_FATAL_FAILURE(InitRenderTarget(2));
+
+    // Create a renderPass with a single color attachment
+    VkAttachmentReference attachment_ref = {};
+    attachment_ref.layout = VK_IMAGE_LAYOUT_GENERAL;
+    std::array<VkAttachmentReference, 2> attachment_refs = {attachment_ref, attachment_ref};
```
arno-lunarg, line 2348: Nit: could simply be:
```suggestion
    std::array attachment_refs = {attachment_ref, attachment_ref};
```

spencer-lunarg: agree, that is nicer

Issue #5898, opened 2023-05-23
Improve Object Tracker error messages


zmike:
I updated today and now I'm getting a bunch of these errors:

```
VUID-vkCmdBindPipeline-commonparent(ERROR / SPEC): msgNum: -1497306210 - Validation Error: [ VUID-vkCmdBindPipeline-commonparent ] Object 0: handle = 0x5e50a50, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0xa6c0eb9e | Object 0xba7514000000002a of type VkPipeline was not created, allocated or retrieved from the correct device. The Vulkan spec states: Both of commandBuffer, and pipeline must have been created, allocated, or retrieved from the same VkDevice (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdBindPipeline-commonparent)
```

**Environment:**
 - OS: linux
 - GPU: any
 - SDK or header version if building from repo: 8f4fc4c98acf7a6a2c77524a389969a651b7a8c9
 - Options enabled (synchronization, best practices, etc.):

Reproduce with `ZINK_DEBUG=validation MESA_LOADER_DRIVER_OVERRIDE=zink ./glcts -n KHR-GLES31.core.compute_shader.simple-compute-shared_context`

zmike:
I also get a bonus VVL crash if I add this VUID to my ignore list.

spencer-lunarg:
The issue is `vkCreateComputePipelines` used `VkDevice = 0x55cbf7f4c4d0` to create the `VkPipeline`

The `VkCommandBuffer = 0x55cbf93f0040` in `vkCmdBindPipeline` used a `vkAllocateCommandBuffers` with  `VkDevice = 0x55cbf9721ef0`
     
So the issue is the `VkDevice` are not the same for the Pipeline and CommandBuffer in this bind

zmike:
Ohh okay, that error message is a bit misleading. It'd be great if it could be improved to:
* explicitly state that the cmdbuf and pipeline were allocated from different devices
* print the pointers of the cmdbuf, pipeline, device(s)

spencer-lunarg:
This is a generated error check (and message)

I agree the error message is not great. I will alter this issue to fix the error message (assuming it was actually catching the error)

PR #5935, opened 2023-05-31
Improve CommonParent ObjectTracker VU message

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5898

before error message

> Validation Error: [ VUID-vkCmdFillBuffer-commonparent ] Object 0: handle = 0x564839cb0ed0, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0x40480f7a | **Object 0xfa21a40000000003 of type VkBuffer was not created, allocated or retrieved from the correct device**

and now

> Validation Error: [ VUID-vkCmdFillBuffer-commonparent ] Object 0: handle = 0x55bf1cd25ed0, type = VK_OBJECT_TYPE_INSTANCE; Object 1: handle = 0x55bf1e4ffaf0, type = VK_OBJECT_TYPE_DEVICE; Object 2: handle = 0x55bf1e5805f0, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x40480f7a | **The VkBuffer (0xfa21a40000000003) was created, allocated or retrieved from the VkDevice 0x55bf1e5805f0 but the other Dispatchable Handle used VkDevice 0x55bf1e4ffaf0.**


layers/object_tracker/object_lifetime_validation.h
```diff
@@ -145,11 +145,13 @@ class ObjectLifetimes : public ValidationObject {
...
                                     if ((wrong_device_code != kVUIDUndefined) && (object_type != kVulkanObjectTypeSurfaceKHR)) {
-                                        return LogError(instance, wrong_device_code,
-                                                        "Object 0x%" PRIxLEAST64
-                                                        " of type %s"
-                                                        " was not created, allocated or retrieved from the correct device.",
-                                                        object_handle, object_string[object_type]);
+                                        const LogObjectList objlist(instance, device, layer_object_data->device);
+                                        return LogError(objlist, wrong_device_code,
+                                                        "The %s (0x%" PRIxLEAST64
+                                                        ") was created, allocated or retrieved from the VkDevice %p but the other "
```
arno-lunarg, line 151: IMO what "the other Dispatchable Handle" refers to is not clear. Could it be more specific?

layers/object_tracker/object_lifetime_validation.h
```diff
@@ -145,11 +145,13 @@ class ObjectLifetimes : public ValidationObject {
...
+                                                        "Dispatchable Handle used VkDevice %p.",
+                                                        object_string[object_type], object_handle, layer_object_data->device,
+                                                        device);
```
arno-lunarg, line 154:
```suggestion
                                        return LogError(objlist, wrong_device_code,
                                                        "The %s (0x%" PRIxLEAST64
                                                        ") was created, allocated or retrieved from %s but the other "
                                                        "Dispatchable Handle used %s.",
                                                        object_string[object_type], object_handle, report_data->FormatHandle(layer_object_data->device).c_str(),
                                                        report_data->FormatHandle(device).c_str());
```

layers/object_tracker/object_lifetime_validation.h
spencer-lunarg: So the next part of the error message says

> The Vulkan spec states: Both of commandBuffer, and dstBuffer must have been created, allocated, or retrieved from the same VkDevice

I 100% agree with you, the issue is this is calling

```
// command_buffer = device_a
// buffer = device_b
vkCmdFillBuffer(command_buffer, buffer)
```

and the "Other Dispatchable Handle" is always going to refer to the `command_buffer` here because the object tracking uses the first handle to base everything off of

What really needs to happen is improve the generation code to pass down more information, which requires a change in the code generation logic which adds 20x the time to fix (in the current state of generation :cry: )

I feel this is at least providing a **better** message then we had which is good for the short term

arno-lunarg: Without speaking code gen, you comment addresses my issue: It seems the error message could be `The %s was created, allocated or retrieved from %s but commandBuffer was created, allocated, or retrieved from %s.` ?

spencer-lunarg: how does one know this if from `commandBuffer` and not something like `VUID-vkQueueBindSparse-commonparent` where it would be the `VkQueue` in this case

Again, this will require code gen to pass down this information.... I am gonna just do it right now, no point to have this intermediate state since the SDK already just shipped

spencer-lunarg: This now says

> vkCmdFillBuffer(): Expected all Dispatchable Handles to use VkDevice 0x557b103efaf0[], but the VkBuffer (0xfa21a40000000003) was created, allocated or retrieved from VkDevice 0x557b104703b0[].

arno-lunarg (approved):

artem-lunarg (approved):

PR #5936, opened 2023-05-31
Fix VUID for CommandBuffers and DescriptorSets

spencer-lunarg:
Currently we have `kVUIDUndefined` where we should be using the proper VUID

Also `vk::FreeCommandBuffers(m_device->device(), m_commandPool->handle(), 2, nullptr);` was crashing because we were not actually calling into `manual_PreCallValidateFreeCommandBuffers`


jeremyg-lunarg (approved):
LGTM

PR #5937, opened 2023-05-31
layers: Add missing vkImportSemaphore{Type} VUIDs

artem-lunarg:
Adds VUIDs (fd and FUCHSIA handles) that it's illegal to import into the semaphore that is in use.


jeremyg-lunarg (approved):
LGTM

spencer-lunarg (approved):

PR #5941, opened 2023-05-31
layers: Fix debug label validation

### LINKED ISSUE #5671: Validation for VUID-vkCmdEndDebugUtilsLabelEXT-commandBuffer-01912 is not correct according to the Vulkan spec.
**Environment:**
 - OS: Windows 10
 - GPU: RTX 3070ti
 - SDK or header version if building from repo: Vulkan SDK 1.3.243.0

**Describe the Issue**
VVL error introduced in this [PR](https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5277/files#diff-4311fff80ea2686e702e6fb7fc1a33f8963ca268ea3212fe9ccae3a2434e60d9R590) is not correct according to the Vulkan spec. The spec says ([source](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCmdEndDebugUtilsLabelEXT.html)): 

> An application may open a debug label region in one command buffer and close it in another, or otherwise split debug label regions across multiple command buffers or multiple queue submissions. When viewed from the linear series of submissions to a single queue, the calls to [vkCmdBeginDebugUtilsLabelEXT](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCmdBeginDebugUtilsLabelEXT.html) and [vkCmdEndDebugUtilsLabelEXT](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCmdEndDebugUtilsLabelEXT.html) must be matched and balanced.

This validation check implemented [here](https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5277/files#diff-8c4c0adae5e8a48e96ed1ad2569f5c065155cfca223e50cac4c13af0c7e0b5f2R1687) checks for matching debug labels *per command buffer* while it should check *per queue*. There's an exception for secondary command buffers ([source](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCmdEndDebugUtilsLabelEXT.html)):

> VUID-vkCmdEndDebugUtilsLabelEXT-commandBuffer-01913
If commandBuffer is a secondary command buffer, there must be an outstanding vkCmdBeginDebugUtilsLabelEXT command recorded to commandBuffer that has not previously been ended by a call to vkCmdEndDebugUtilsLabelEXT

In that case checking for matching debug labels makes sense _per secondary command buffer_.

**Steps to reproduce**
1. Download and install Unity 2022.2.4f1 or above (e.g. https://unity.com/releases/editor/whats-new/2022.2.15)
2. Install VVL that has this [PR](https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5277/files#diff-4311fff80ea2686e702e6fb7fc1a33f8963ca268ea3212fe9ccae3a2434e60d9R590) landed
3. Launch Unity with any project
4. Switch Graphics API to Vulkan (https://docs.unity3d.com/Manual/GraphicsAPIs.html)
5. Launch Unity again with `--force-vulkan-layers` argument from CLI.
6. Observe `VUID-vkCmdEndDebugUtilsLabelEXT-commandBuffer-01912` get logged in the Unity console.

**Expected behavior**
VUID-vkCmdEndDebugUtilsLabelEXT-commandBuffer-01912 gets validated per queue
VUID-vkCmdEndDebugUtilsLabelEXT-commandBuffer-01913 gets validated per secondary command buffer

**Valid Usage ID**
VUID-vkCmdEndDebugUtilsLabelEXT-commandBuffer-01912

juan-lunarg:
This VUID cannot be validated in vkCmdEndDebugUtilsLabelEXT

So I'm adding it to the unimplementable_validation.h file.

My plan is to help fix/clear up the specification.

Meanwhile fixing up this false positive and adding positive
testing to prevent a regression.

closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5671

### DIFF
--- a/layers/core_checks/cc_cmd_buffer.cpp
+++ b/layers/core_checks/cc_cmd_buffer.cpp
@@ -1743,10 +1743,13 @@ bool CoreChecks::PreCallValidateCmdEndDebugUtilsLabelEXT(VkCommandBuffer command
     auto cb_state = GetRead<CMD_BUFFER_STATE>(commandBuffer);
     assert(cb_state);
     bool skip = false;
+
+    if (cb_state->IsPrimary()) {
+        return skip;
+    }
+
     if (cb_state->LabelStackDepth() < 1) {
-        const auto vuid = cb_state->IsPrimary() ? "VUID-vkCmdEndDebugUtilsLabelEXT-commandBuffer-01912"
-                                                : "VUID-vkCmdEndDebugUtilsLabelEXT-commandBuffer-01913";
-        skip |= LogError(commandBuffer, vuid,
+        skip |= LogError(commandBuffer, "VUID-vkCmdEndDebugUtilsLabelEXT-commandBuffer-01913",
                          "vkCmdEndDebugUtilsLabelEXT() called without a corresponding vkCmdBeginDebugUtilsLabelEXT first");
     }
     return skip;
--- a/tests/positive/command.cpp
+++ b/tests/positive/command.cpp
@@ -938,3 +938,89 @@ TEST_F(PositiveCommand, FillBufferCmdPoolTransferQueue) {
     cb.FillBuffer(buffer.handle(), 0, 12, 0x11111111);
     cb.end();
 }
+
+TEST_F(PositiveCommand, DebugLabelPrimaryCommandBuffer) {
+    TEST_DESCRIPTION("Create a debug label on a primary command buffer");
+
+    AddRequiredExtensions(VK_EXT_DEBUG_UTILS_EXTENSION_NAME);
+    ASSERT_NO_FATAL_FAILURE(InitFramework());
+    if (!AreRequiredExtensionsEnabled()) {
+        GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported";
+    }
+    ASSERT_NO_FATAL_FAILURE(InitState());
+    m_commandBuffer->begin();
+
+    auto label = LvlInitStruct<VkDebugUtilsLabelEXT>();
+    label.pLabelName = "test";
+    vk::CmdBeginDebugUtilsLabelEXT(*m_commandBuffer, &label);
+    vk::CmdEndDebugUtilsLabelEXT(*m_commandBuffer);
+
+    m_commandBuffer->end();
+
+    const std::array command_buffers = {m_commandBuffer->handle()};
+
+    auto submit_info = LvlInitStruct<VkSubmitInfo>();
+    submit_info.commandBufferCount = size32(command_buffers);
+    submit_info.pCommandBuffers = command_buffers.data();
+
+    vk::QueueSubmit(m_device->m_queue, 1, &submit_info, VK_NULL_HANDLE);
+    vk::QueueWaitIdle(m_device->m_queue);
+}
+
+TEST_F(PositiveCommand, DebugLabelPrimaryCommandBuffers) {
+    TEST_DESCRIPTION("Begin a debug label on 1 command buffer, then end it on another command buffer");
+
+    AddRequiredExtensions(VK_EXT_DEBUG_UTILS_EXTENSION_NAME);
+    ASSERT_NO_FATAL_FAILURE(InitFramework());
+    if (!AreRequiredExtensionsEnabled()) {
+        GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported";
+    }
+    ASSERT_NO_FATAL_FAILURE(InitState());
+
+    VkCommandBufferObj command_buffer_start(m_device, m_commandPool);
+    VkCommandBufferObj command_buffer_end(m_device, m_commandPool);
+
+    // Start DebugLabel on 1 command buffer
+    {
+        auto label = LvlInitStruct<VkDebugUtilsLabelEXT>();
+        label.pLabelName = "test";
+        command_buffer_start.begin();
+        vk::CmdBeginDebugUtilsLabelEXT(command_buffer_start.handle(), &label);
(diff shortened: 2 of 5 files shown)


tests/positive/command.cpp
```diff
@@ -938,3 +938,85 @@ TEST_F(PositiveCommand, FillBufferCmdPoolTransferQueue) {
...
+
+    auto submit_info = LvlInitStruct<VkSubmitInfo>();
+    submit_info.commandBufferCount = size32(command_buffers);
+    submit_info.pCommandBuffers = command_buffers.data();
+
+    vk::QueueSubmit(m_device->m_queue, 1, &submit_info, VK_NULL_HANDLE);
+    vk::QueueWaitIdle(m_device->m_queue);
+}
+
+TEST_F(PositiveCommand, DebugLabelPrimaryCommandBuffers) {
```
juan-lunarg, line 968: This test highlights reproduces the issue described in https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5671

spencer-lunarg:
> The main issue is if Queue refers to the logical vulkan queue the work is being submitted on. Or the actual hardware queue.

This is 100% the `VkQueue` object. The hardware implementation of how it maps the `VkQueue` is outside the scope of any Vulkan Tool

layers/core_checks/cc_queue.cpp
```diff
@@ -212,28 +214,41 @@ bool CoreChecks::PreCallValidateQueueSubmit(VkQueue queue, uint32_t submitCount,
...
+                                 "vkQueueSubmit(): command buffer %s is protected while queue %s pSubmits[%u] has %s",
+                                 report_data->FormatHandle(cb_state->commandBuffer()).c_str(),
+                                 report_data->FormatHandle(queue).c_str(), submit_idx,
+                                 protected_submit_info ? "VkProtectedSubmitInfo:protectedSubmit set to VK_FALSE"
+                                                       : "no VkProtectedSubmitInfo in the pNext chain");
             }
         }
     }
 
+    // What about QueueSubmit2
```
spencer-lunarg, line 246: also `vkQueueBindSparse` (https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#devsandqueues-submission)

layers/core_checks/cc_cmd_buffer.cpp
```diff
@@ -346,13 +346,25 @@ bool CoreChecks::PreCallValidateEndCommandBuffer(VkCommandBuffer commandBuffer)
...
+    if (is_primary || !(cb_state.beginInfo.flags & VK_COMMAND_BUFFER_USAGE_RENDER_PASS_CONTINUE_BIT)) {
         // This needs spec clarification to update valid usage, see comments in PR:
         // https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/165
         skip |= InsideRenderPass(cb_state, "vkEndCommandBuffer()", "VUID-vkEndCommandBuffer-commandBuffer-00060");
     }
 
+    if (!is_primary) {
+        const int32_t label_stack = cb_state.LabelStack();
+
+        if (label_stack != 0) {
```
jimblandy, line 361: The problem with checking the depth only at the end is that it permits command buffers that end a range first, and then begin a new range. The checks really do need to happen at every end.

For the primary command buffer, the validator needs to track the *signed* minimum and maximum depth reached over the course of the command buffer. Then, validation at submission time needs to check that such a range of motion starting at the queue's current depth is acceptable.

jimblandy: Hmm, I guess if we're tracking signed minimum and maximum depth, then you could indeed check for 01913 when the secondary command buffer is ended: just see if the minimum is negative.

juan-lunarg:
> This is 100% the `VkQueue` object. The hardware implementation of how it maps the `VkQueue` is outside the scope of any Vulkan Tool

My main concern was this 1 line here from the description. 

> `An application may open a debug label region in one command buffer and close it in another, or otherwise split debug label regions across multiple command buffers or multiple queue submissions.`

Which makes it sounds like you can open a debug label region on 1 queue. Submit. Then end the debug label, then submit.

And when talking to @MarkY-LunarG (The author of this extension) they said

> I was working with the AMD people who were using this, and we really can't validate it.  I thought we added a note that we might not be able to validate it because of all this.

MarkY-LunarG:
> This is 100% the VkQueue object. The hardware implementation of how it maps the VkQueue is outside the scope of any Vulkan Tool

This logic goes back because we were asked to support this across multiple QueueSubmits.  Because of this, we really can only validate this on the GPU using something akin to GPU AV where we have a check for every `BeginLabel` and increment a counter and then check every `EndLabel` on the GPU to make sure the value is > 0 before we decrement and throw the error then if the value is <= 0.  It might just be something we can't test properly at this point.  [NOTE: There are other valid usage statements that also fall into this bin].

layers/core_checks/cc_cmd_buffer.cpp
juan-lunarg: I've reverted the error checking for `VUID-vkCmdEndDebugUtilsLabelEXT-commandBuffer-01913` so this should be addressed now.

tests/positive/command.cpp
```diff
@@ -938,3 +938,89 @@ TEST_F(PositiveCommand, FillBufferCmdPoolTransferQueue) {
...
```
juan-lunarg, line 970: This positive test will at the very least prevent a regression on what users expect from the validation layer.

spencer-lunarg (approved):

PR #5944, opened 2023-06-01
Add new PrimitiveTopology VUIDs

spencer-lunarg:
[change](https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5916) from the 1.3.251 spec

This adds a VUID (and tests) for 

- VUID-VkGraphicsPipelineCreateInfo-pStages-08888
- VUID-VkGraphicsPipelineCreateInfo-topology-08889
- VUID-VkGraphicsPipelineCreateInfo-topology-08890

### DIFF
--- a/layers/core_checks/cc_pipeline_graphics.cpp
+++ b/layers/core_checks/cc_pipeline_graphics.cpp
@@ -899,9 +899,11 @@ bool CoreChecks::ValidateGraphicsPipelineInputAssemblyState(const PIPELINE_STATE
         }
     }
 
+    const bool ignore_topology = pipeline.IsDynamic(VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY) &&
+                                 phys_dev_ext_props.extended_dynamic_state3_props.dynamicPrimitiveTopologyUnrestricted;
     // pre_raster has tessellation stage, vertex input has topology
-    // Both are needed for  these checks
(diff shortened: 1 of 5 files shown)


layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -899,24 +899,32 @@ bool CoreChecks::ValidateGraphicsPipelineInputAssemblyState(const PIPELINE_STATE
         }
     }
 
+    const bool ignore_topology = pipeline.IsDynamic(VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY) &&
+                                 phys_dev_ext_props.extended_dynamic_state3_props.dynamicPrimitiveTopologyUnrestricted;
```
juan-lunarg, line 903: Minor code repetition here and in cc_shader.cpp

juan-lunarg (approved):
LGTM Jenkins failure looks unrelated

layers/core_checks/cc_pipeline_graphics.cpp
spencer-lunarg: There is a 3rd spot, as well, was going to fix it, but lot of other things that need fixing around this dynamic state / GPL stuff (mainly for GPL) ... I will make a follow-up PR that fixes this when get back to GPL stuff soon

PR #5945, opened 2023-06-01
Remove unused stateless for VkPresentRegionsKHR

spencer-lunarg:
There were 2 `ValidateArray` that were not correct, this removes them and adds tests

### DIFF
--- a/layers/stateless/sl_cmd_buffer.cpp
+++ b/layers/stateless/sl_cmd_buffer.cpp
@@ -227,9 +227,8 @@ bool StatelessValidation::ValidateCmdBindVertexBuffers2(VkCommandBuffer commandB
                     not_null_msg = "pSizes is not NULL";
                 else
                     not_null_msg = "pStrides is not NULL";
-                const char *vuid_breach_msg = "%s: %s, so bindingCount must be greater than 0.";
-                skip |= LogError(commandBuffer, "VUID-vkCmdBindVertexBuffers2-bindingCount-arraylength", vuid_breach_msg, api_call,
-                                 not_null_msg);
+                skip |= LogError(commandBuffer, "VUID-vkCmdBindVertexBuffers2-bindingCount-arraylength",
+                                 "%s: %s, so bindingCount must be greater than 0.", api_call, not_null_msg);
             }
         }
     }
--- a/layers/stateless/sl_cmd_buffer.cpp
+++ b/layers/stateless/sl_cmd_buffer.cpp
@@ -227,7 +227,7 @@ bool StatelessValidation::ValidateCmdBindVertexBuffers2(VkCommandBuffer commandB
                     not_null_msg = "pSizes is not NULL";
                 else
                     not_null_msg = "pStrides is not NULL";
-                const char *vuid_breach_msg = "%s: %s, so bindingCount must be greater that 0.";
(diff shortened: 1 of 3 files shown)


layers/stateless/sl_wsi.cpp
```diff
@@ -186,14 +186,6 @@ bool StatelessValidation::manual_PreCallValidateQueuePresentKHR(VkQueue queue, c
             skip |= ValidateStructPnext("QueuePresentKHR", "pCreateInfo->pNext->pNext", NULL, present_regions->pNext, 0, NULL,
                                         GeneratedVulkanHeaderVersion, "VUID-VkPresentInfoKHR-pNext-pNext",
                                         "VUID-VkPresentInfoKHR-sType-unique");
-            skip |= ValidateArray("QueuePresentKHR", "pCreateInfo->pNext->swapchainCount", "pCreateInfo->pNext->pRegions",
```
spencer-lunarg, line 189: already being done in generated code, tests hits this, makes sense why this had no VUID

layers/stateless/sl_wsi.cpp
```diff
@@ -186,14 +186,6 @@ bool StatelessValidation::manual_PreCallValidateQueuePresentKHR(VkQueue queue, c
...
-                                  present_regions->swapchainCount, &present_regions->pRegions, true, false, kVUIDUndefined,
-                                  kVUIDUndefined);
-            for (uint32_t i = 0; i < present_regions->swapchainCount; ++i) {
-                skip |= ValidateArray("QueuePresentKHR", "pCreateInfo->pNext->pRegions[].rectangleCount",
```
spencer-lunarg, line 193: Spec:

> rectangleCount is the number of rectangles in pRectangles, or zero if the entire image has changed and should be presented.

This was checking if `rectangleCount` was zero, which is produced a false positive

layers/stateless/sl_cmd_buffer.cpp
```diff
@@ -227,7 +227,7 @@ bool StatelessValidation::ValidateCmdBindVertexBuffers2(VkCommandBuffer commandB
                     not_null_msg = "pSizes is not NULL";
                 else
                     not_null_msg = "pStrides is not NULL";
-                const char *vuid_breach_msg = "%s: %s, so bindingCount must be greater that 0.";
+                const char *vuid_breach_msg = "%s: %s, so bindingCount must be greater than 0.";
                 skip |= LogError(commandBuffer, "VUID-vkCmdBindVertexBuffers2-bindingCount-arraylength", vuid_breach_msg, api_call,
```
juan-lunarg, line 231: Don't see much point to the `vuid_breach_msg` variable.

```suggestion
                skip |= LogError(commandBuffer, "VUID-vkCmdBindVertexBuffers2-bindingCount-arraylength", "%s: %s, so bindingCount must be greater than 0.", api_call,
```

juan-lunarg (approved):

layers/stateless/sl_cmd_buffer.cpp
spencer-lunarg: agree, let me fix this while here

spencer-lunarg: I only found this because I mis-spelled `than` and noticed this code lol

Issue #5295, opened 2023-02-16
False positive waiting on exported timeline semaphore that has had its win32 handle retrieved?


rygo6:
I seem to have run into the same bug highlighted in #1502 and #1999 that was then supposedly fixed in #2196 where it erroneously reports that you cannot wait on only a timeline semaphore submitted to a queue, but that you also need a fence.

It appears like #2196 did fix it except for when that timeline semaphore is also created with VkExportSemaphoreCreateInfo chained on pNext. Then you call vkGetSemaphoreWin32HandleKHR on the semaphore to get the WinNT handle.

So to be clear. This timeline semaphore works perfectly fine:
```
    const VkExportSemaphoreCreateInfo exportSemaphoreCreateInfo = {
            .sType = VK_STRUCTURE_TYPE_EXPORT_SEMAPHORE_CREATE_INFO,
            .handleTypes = externalHandleType,
    };
    const VkSemaphoreTypeCreateInfo timelineSemaphoreTypeCreateInfo = {
            .sType = VK_STRUCTURE_TYPE_SEMAPHORE_TYPE_CREATE_INFO,
            .pNext = &exportSemaphoreCreateInfo,
            .semaphoreType = VK_SEMAPHORE_TYPE_TIMELINE,
            .initialValue = 0,
    };
    const VkSemaphoreCreateInfo timelineSemaphoreCreateInfo = {
            .sType = VK_STRUCTURE_TYPE_SEMAPHORE_CREATE_INFO,
            .pNext = &timelineSemaphoreTypeCreateInfo,
            .flags = 0,
    };
    FBR_VK_CHECK_RETURN(vkCreateSemaphore(pVulkan->device, &timelineSemaphoreCreateInfo, NULL, &pVulkan->timelineSemaphore));
```

If I also make the call to retrieve it's WinNT handle:
```
    VkSemaphoreGetWin32HandleInfoKHR semaphoreGetWin32HandleInfo = {
            .sType = VK_STRUCTURE_TYPE_SEMAPHORE_GET_WIN32_HANDLE_INFO_KHR,
            .pNext = VK_NULL_HANDLE,
            .handleType = externalHandleType,
            .semaphore = pVulkan->timelineSemaphore,
    };

    // TODO move to preloaded refs
[...]

PR #5947, opened 2023-06-01
layers: Fix queue progression with imported semaphore

artem-lunarg:
Might also fix https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5295


artem-lunarg:
Potential issue with AMD driver. It fails to create semaphore with __both__ timeline type and export support. Separately I can create timeline semaphores and export semaphores, only combination of `VkSemaphoreTypeCreateInfo` + `VkExportSemaphoreCreateInfo` fails. It works on NVIDIA.

jeremyg-lunarg (approved):
LGTM. I wish timeline semaphores could be simpler

PR #5957, opened 2023-06-02
cmake: Fix MSVC warning 4130

juan-lunarg:


layers/sync/sync_validation.h
```diff
@@ -1380,10 +1380,19 @@ class CommandExecutionContext {
 
     virtual HazardResult DetectFirstUseHazard(const ResourceUsageRange &tag_range) = 0;
     virtual void BeginRenderPassReplay(const SyncOpBeginRenderPass &begin_op, ResourceUsageTag tag) {
-        assert("Must override if use by derived type is valid" == nullptr);
```
artem-lunarg, line 1383: In case of failures such asserts print informative messages, comments do not provide runtime functionality. Can we keep this warning suppressed?

artem-lunarg: On the other hand I see we have only few cases of this, then it's fine.

juan-lunarg: Agreed it's nice to have the informative messages but this is not the way to do it.

I like the assertm show on ccpreference:
`#define assertm(exp, msg) assert(((void)msg, exp))`

But didn't want to complicate this PR

PR #5958, opened 2023-06-02
VVL for VK_EXT_dynamic_rendering_unused_attachments

pdaniell-nv:
Validation implementation for the [VK_EXT_dynamic_rendering_unused_attachments](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_EXT_dynamic_rendering_unused_attachments.html) extension.


tests/negative/dynamic_rendering.cpp
```diff
@@ -1356,6 +1356,185 @@ TEST_F(NegativeDynamicRendering, MistmatchingAttachmentFormats2) {
     m_commandBuffer->end();
 }
 
+TEST_F(NegativeDynamicRendering, MistmatchingAttachmentFormats3) {
```
juan-lunarg, line 1359: Currently this test won't run on our Mock driver + profiles testing. Please add the necessary functionality to `max_profiles.json`.

This commit should be useful reference:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/e83daa920061718ec958ab396f2bee2b131a34b4

pdaniell-nv: Thanks for the reference. I've added to max_profiles.json as requested.

tests/negative/dynamic_rendering.cpp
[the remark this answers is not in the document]
pdaniell-nv: Isn't that what CTS is for?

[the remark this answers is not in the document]
pdaniell-nv: I've added positive tests as requested.

tests/negative/dynamic_rendering.cpp
spencer-lunarg: This isn't obvious, I will add this to the "how to add validation for new extension" docs i am creating

tests/negative/dynamic_rendering.cpp
```diff
@@ -1356,6 +1356,185 @@ TEST_F(NegativeDynamicRendering, MistmatchingAttachmentFormats2) {
...
+    GetPhysicalDeviceFeatures2(dynamic_rendering_features);
+    if (!dynamic_rendering_features.dynamicRendering) {
+        GTEST_SKIP() << "Test requires (unsupported) dynamicRendering , skipping.";
+    }
+    if (!dynamic_rendering_unused_attachments_features.dynamicRenderingUnusedAttachments) {
+        GTEST_SKIP() << "Test requires (unsupported) dynamicRenderingUnusedAttachments , skipping.";
+    }
+
+    ASSERT_NO_FATAL_FAILURE(InitState(nullptr, &dynamic_rendering_features, VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT));
+    if (::testing::Test::IsSkipped()) return;
```
spencer-lunarg, line 1385: this shouldn't be needed now (in other tests as well)

pdaniell-nv: I just removed this from my own tests. I didn't want to risk removing it from all the others. It appears ~146 times.

spencer-lunarg: awww, sorry, we use it for `InitBasicDynamicRendering`

Basically GTest has issue where it needs to return void, so if we skip in the setup, we need a way to actually end the tests

spencer-lunarg: for now these tests can't make use of `InitBasicDynamicRendering` it seems, but we are slowly trying to reduce the amount of code/hassle to setup all the ext/features per test!

layers/core_checks/cc_drawdispatch.cpp
```diff
@@ -80,11 +80,14 @@ struct DispatchVuidsCmdDraw : DrawDispatchVuid {
         dynamic_rendering_view_mask_06178        = "VUID-vkCmdDraw-viewMask-06178";
         dynamic_rendering_color_count_06179      = "VUID-vkCmdDraw-colorAttachmentCount-06179";
         dynamic_rendering_color_formats_06180    = "VUID-vkCmdDraw-colorAttachmentCount-06180";
+        dynamic_rendering_unused_attachments_08911 = "VUID-vkCmdDraw-dynamicRenderingUnusedAttachments-08911";
```
spencer-lunarg, line 83: so `VUID-vkCmdDraw-dynamicRenderingUnusedAttachments-08910`, `08912`, `08913`, `08914`, `08916`, and `08917` I realize are new, but just the same "logic" with

> the dynamicRenderingUnusedAttachments feature is not enabled

added

Something for outside the scope of this PR, but will probably need to start creating a better way then creating 

90 more of these copy-and-paste lines

In the short there, the "support" for `VK_EXT_dynamic_rendering_unused_attachments` will not look 100% but I realize it actually is now 100% done with this PR

spencer-lunarg (approved):

layers/core_checks/cc_drawdispatch.cpp
pdaniell-nv: I wondered about this general problem while working on this PR. I guess the improvements to simplify the VU statements to be less ifdefy will improve this situation.

tests/negative/dynamic_rendering.cpp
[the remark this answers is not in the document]
juan-lunarg: > Isn't that what CTS is for?

Our testing goes through address / thread sanitizer now to catch hard to find bugs. So writing positive tests here adds quite a bit. Also it makes it easier to add new positive tests in the future in case of false positives. Also CTS isn't yet part of our CI process we just manually run CTS currently.

PR #5962, opened 2023-06-02
cmake: Add VVL_SPIRV_HOPPER option

juan-lunarg:


spencer-lunarg (approved):
cc @dneto0

tests/spirv_hopper/README.md
```diff
@@ -2,6 +2,12 @@
 
 The goal of `SPIRV-Hopper` is to feed SPIR-V binaries into the Validation Layers
 
+## How To Build
+
+By default `SPIRV-Hopper` is turned off in the build until SPIRV-Reflect is part of `known_good.json` which requires upstream changes.
```
juan-lunarg, line 7: Note I'd actually like to upstream CMake fixes to SPIRV-Tools as well since it also has a couple issues causing us problems and will be necessary to make sure everything works nicely together.

Issue #5971, opened 2023-06-09
VUID-VkImageCompressionControlEXT-flags-06747 - VK_IMAGE_COMPRESSION_DEFAULT_EXT flag causing error.


sewbro01-ARM:
**Describe the Issue**

When using **VK_IMAGE_COMPRESSION_DEFAULT_EXT** in **VkImageCompressionControlEXT** we are getting Validation Layers error about using flag **0**. Problem is that **VK_IMAGE_COMPRESSION_DEFAULT_EXT** value is **0** which according to specification is valid flag for usage with **VkImageCompressionControlEXT** but it is being handled as incorrect by VL.

**Expected behavior**

VK_IMAGE_COMPRESSION_DEFAULT_EXT flag not causing VL error.

**Valid Usage ID**
VUID-VkImageCompressionControlEXT-flags-06747

**Additional context**
I debugged code and found possible error in logic - in Vulkan-ValidationLayers/layers/stateless/sl_image.cpp around line 583 where we have :

        const auto image_compression_control = LvlFindInChain<VkImageCompressionControlEXT>(pCreateInfo->pNext);
        if (image_compression_control) {
            constexpr VkImageCompressionFlagsEXT AllVkImageCompressionFlagBitsEXT =
                (VK_IMAGE_COMPRESSION_DEFAULT_EXT | VK_IMAGE_COMPRESSION_FIXED_RATE_DEFAULT_EXT |
                 VK_IMAGE_COMPRESSION_FIXED_RATE_EXPLICIT_EXT | VK_IMAGE_COMPRESSION_DISABLED_EXT);
            skip |= ValidateFlags("vkCreateImage", "VkImageCompressionControlEXT::flags", "VkImageCompressionFlagsEXT",
                                  AllVkImageCompressionFlagBitsEXT, image_compression_control->flags, kRequiredSingleBit,
                                  "VUID-VkImageCompressionControlEXT-flags-06747"); 

Most probable fix is just changing flag_type in: **ValidateFlags()** from kRequiredSingleBit to kOptionalSingleBit.

<details>
<summary>VL Error</summary>

```sh
[...]

PR #5972, opened 2023-06-09
layers: VK_IMAGE_COMPRESSION_DEFAULT_EXT flag fix

sewbro01-ARM:
VL falsely marks flag VK_IMAGE_COMPRESSION_DEFAULT_EXT as invalid. The value of this enum is 0 and ValidateFlags function used to check if flags are properly set has 'flag_type' parameter set to kRequiredSingleBit which doesnt allows 0 as flag value.

Changing flag_type to kOptionalSingleBit resolves issue.

This fixes issue #5971

### DIFF
--- a/layers/stateless/sl_image.cpp
+++ b/layers/stateless/sl_image.cpp
@@ -585,7 +585,7 @@ bool StatelessValidation::manual_PreCallValidateCreateImage(VkDevice device, con
                 (VK_IMAGE_COMPRESSION_DEFAULT_EXT | VK_IMAGE_COMPRESSION_FIXED_RATE_DEFAULT_EXT |
                  VK_IMAGE_COMPRESSION_FIXED_RATE_EXPLICIT_EXT | VK_IMAGE_COMPRESSION_DISABLED_EXT);
             skip |= ValidateFlags("vkCreateImage", "VkImageCompressionControlEXT::flags", "VkImageCompressionFlagsEXT",
-                                  AllVkImageCompressionFlagBitsEXT, image_compression_control->flags, kRequiredSingleBit,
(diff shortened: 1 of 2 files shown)


spencer-lunarg (changes requested):
This looks good, but we really should have a positive test that uses `VK_IMAGE_COMPRESSION_DEFAULT_EXT` to prevent regressions as this is an easy thing to mess up if we would ever refactor `ValidateFlags` in the future (which is on my todo list)

spencer-lunarg:
> had some trouble with setting environment correctly for the required compression extension to work properly with VL tests.

Not sure what you had troubles with, but is it anything we failed to document somewhere? I hate when people have to waste time getting dev environment setup instead of just doing the actual work

sewbro01-ARM:
I used MockICD and Profiles layer, maybe I made some silly mistake but I couldn't make those tests to be not skipped at first, then I used our internal method for running tests with our device from ARM that unfortunately I can't share with you.

juan-lunarg:
> I used MockICD and Profiles layer, maybe I made some silly mistake but I couldn't make those tests to be not skipped at first, then I used our internal method for running tests with our device from ARM that unfortunately I can't share with you.

Was there something wrong/unclear with the documentation on how to setup the Mock + Profiles testing?

https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/tests/README.md#running-tests-on-mockicd-and-profiles-layer

juan-lunarg:
I ended up resetting my linux dev computer. And tried to run mock + profiles testing and it didn't work at all.

I updated my vulkan loader and it works now. So our testing currently relies on an up to date loader.

PR #5978, opened 2023-06-10
layers: Add Const Ref for ImageLayout CoreChecks

spencer-lunarg:
Move more items in Image Layout to use const ref where we know we can

also removed a `FRAMEBUFFER_STATE` that was never being used

There seems to be some code here that really belongs in state tracker, I can't remove `CMD_BUFFER_STATE *cb_state` because we are writing to it in this file, which seems strange

### DIFF
--- a/layers/core_checks/cc_render_pass.cpp
+++ b/layers/core_checks/cc_render_pass.cpp
@@ -462,13 +462,17 @@ bool CoreChecks::ValidateCmdBeginRenderPass(VkCommandBuffer commandBuffer, Rende
             }
             skip |= VerifyFramebufferAndRenderPassImageViews(pRenderPassBegin, function_name);
             skip |= VerifyRenderAreaBounds(pRenderPassBegin, function_name);
-            skip |= VerifyFramebufferAndRenderPassLayouts(rp_version, *cb_state, pRenderPassBegin, fb_state.get());
-            if (fb_state->rp_state->renderPass() != rp_state->renderPass()) {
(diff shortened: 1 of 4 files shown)


layers/core_checks/cc_render_pass.cpp
```diff
@@ -3934,8 +3940,9 @@ bool CoreChecks::PreCallValidateCmdNextSubpass2(VkCommandBuffer commandBuffer, c
 
 void CoreChecks::RecordCmdNextSubpassLayouts(VkCommandBuffer commandBuffer, VkSubpassContents contents) {
     auto cb_state = GetWrite<CMD_BUFFER_STATE>(commandBuffer);
-    auto framebuffer = Get<FRAMEBUFFER_STATE>(cb_state->activeRenderPassBeginInfo.framebuffer);
-    TransitionSubpassLayouts(cb_state.get(), cb_state->activeRenderPass.get(), cb_state->GetActiveSubpass(), framebuffer.get());
+    if (cb_state && cb_state->activeRenderPass) {
```
artem-lunarg, line 3943: Suggestion not to check for null if can't be null according to current logic. I'm not sure if there can be case when Get<> functions return null, if that's not possible that corresponding signatures

artem-lunarg (approved):
LGTM. One comment about null checks.

layers/core_checks/cc_render_pass.cpp
spencer-lunarg: I think what we really need is to replace `GetWrite<>` and `Get<>` accesses with a "guaranteed to not be null" call

There are times where there might not be a valid `cb_state`... but I agree I myself have no way of knowing

PR #5982, opened 2023-06-12
layers: Move SPIRV-Reflect to known_goods

spencer-lunarg:
Move to using `SPIRV-Reflect` as a proper known_good dep now that all the bugs are resolved

I left the `-DVVL_SPIRV_HOPPER` option incase people have issues with this as we migrate it in, there is a way not be blocked by it


tests/spirv_hopper/CMakeLists.txt
```diff
@@ -13,7 +13,25 @@
...
 # ~~~
-add_subdirectory(external)
+add_library(spirv-reflect STATIC)
+target_sources(spirv-reflect PRIVATE
+    ${SPIRV_REFLECT_INSTALL_DIR}/spirv_reflect.h
+    ${SPIRV_REFLECT_INSTALL_DIR}/spirv_reflect.cpp
+)
+
+if(${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
+    target_compile_options(spirv-reflect PUBLIC
```
juan-lunarg, line 23:
```suggestion
    target_compile_options(spirv-reflect PRIVATE
```

tests/spirv_hopper/CMakeLists.txt
```diff
@@ -13,7 +13,25 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ~~~
-add_subdirectory(external)
+add_library(spirv-reflect STATIC)
+target_sources(spirv-reflect PRIVATE
+    ${SPIRV_REFLECT_INSTALL_DIR}/spirv_reflect.h
+    ${SPIRV_REFLECT_INSTALL_DIR}/spirv_reflect.cpp
+)
```
juan-lunarg, line 20: Since the include directory belongs to spirv-reflect it should be a PUBLIC include of spirv-reflect.

```suggestion
)

target_include_directories(spirv-reflect SYSTEM PUBLIC ${SPIRV_REFLECT_INSTALL_DIR})
```

tests/spirv_hopper/CMakeLists.txt
```diff
@@ -31,3 +49,5 @@ target_link_libraries(spirv-hopper PRIVATE
     VkLayer_utils
     spirv-reflect
 )
+
+target_include_directories(spirv-hopper SYSTEM PRIVATE ${SPIRV_REFLECT_INSTALL_DIR})
```
juan-lunarg, line 53:
```suggestion
```

tests/spirv_hopper/README.md
```diff
@@ -4,9 +4,9 @@ The goal of `SPIRV-Hopper` is to feed SPIR-V binaries into the Validation Layers
 
 ## How To Build
 
-By default `SPIRV-Hopper` is turned off in the build until SPIRV-Reflect is part of `known_good.json` which requires upstream changes.
+By default `SPIRV-Hopper` is turned on when building with the rest of the Validation Layers tests.
 
-Enable it with `-D VVL_SPIRV_HOPPER` when building the validation layers.
+This can be turned off with `-DVVL_SPIRV_HOPPER=OFF` when building the tests.
```
juan-lunarg, line 9: This is just a comment. But you can have a space when passing a variable to CMake

`-D VVL_SPIRV_HOPPER=OFF` is valid and is the recommended approach for scripts/documentation since it reads better.

`-DVVL_SPIRV_HOPPER=OFF` is just slightly better for CLI.

juan-lunarg (changes requested):

tests/spirv_hopper/README.md
spencer-lunarg: aww, agree, added space back

tests/spirv_hopper/CMakeLists.txt
```diff
@@ -13,7 +13,27 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ~~~
-add_subdirectory(external)
+add_library(spirv-reflect STATIC)
+target_sources(spirv-reflect PRIVATE
+    ${SPIRV_REFLECT_INSTALL_DIR}/spirv_reflect.h
+    ${SPIRV_REFLECT_INSTALL_DIR}/spirv_reflect.cpp
```
juan-lunarg, line 19: Minor point that doesn't matter because this is test only code. IE this is just a informative comment.

Package managers we support wouldn't play well with this. The proper thing to do would be to use `find_file`. However, since this is test only code it doesn't matter. Since package manager don't try and build our tests.

juan-lunarg (approved):

PR #5983, opened 2023-06-12
tests: Combine Dynamic Rendering tests

spencer-lunarg:
This is the first attempt to combine files for tests

I realized `VkPositiveLayerTest` was empty (since all tests need to be positive outside the dedicated errors) so now this file looks like

```c++
class DynamicRendering : public VkLayerTest {
  public:
    void InitBasicDynamicRendering();
};
class NegativeDynamicRendering : public DynamicRendering {};
class PositiveDynamicRendering : public DynamicRendering {};
```

It is now 9k lines, but I guess common code could still be found and shared between tests in the future

I went with `./tests/tests/dynamic_rendering.cpp` and open to another name for the sub directory other than `tests`

Note: This is just moving code, nothing has been altered to the tests


arno-lunarg:
I personally prefer the `tests/positive/dynamic_rendering.cpp` - `tests/negative/dynamic_rendering.cpp` split. But I guess one problem is having to have to put definitions common to those files, like `class DynamicRendering` here, in some header file.
If we are to keep only one file, why not just name it `tests/dynamic_rendering.cpp`?

artem-lunarg:
That's right. It's Visual Studio specific. There is VS [bug](https://developercommunity.visualstudio.com/t/vs2017-cc-multi-processor-compilation-does-not-wor/586584) that disables multithreaded compilation if a) there are files with the same name and b) VS solution is generated in a specific way (and CMake generates it that way). 

I think it's not great if such issues drive how we organize the project but if there are other considerations to use single file name then a side effect that compilation speed is fixed would be a great bonus.

artem-lunarg:
@arno-lunarg positive/negative prefixes as you suggested also will fix the VS issue. Also +1 for not using nested `tests/tests`.

One idea is to remove some negativity from negative tests and assume that, for example, `tests\dynamic_rendering.cpp` is negative the tests (and also do not use word Negative as part of test suite name) and to keep positive prefix for the positive tests.

PR #5984, opened 2023-06-12
tests: Add missing max_profile ext/features

spencer-lunarg:
Adds `VK_ARM_rasterization_order_attachment_access` features (goes with https://github.com/KhronosGroup/Vulkan-Profiles/pull/447)

Adds `VK_EXT_shader_object` as the one test that uses it was not running (and probably will be used in the near future more)


spencer-lunarg:
> Does the test that uses VK_EXT_shader_object now runs?

Yes, there is only a single `VkPositiveLayerTest.GDPAWithMultiCmdExt` test that uses it, but see in CI it ran

PR #5988, opened 2023-06-12
layers: Move Viewport and Scissor to DynamicState

spencer-lunarg:
There seems to be a need to clean up the Viewport/Scissor dynamic state Secondary Command Buffer Inheritance logic, but for now, the goal is to move all dynamic state value data into the `DynamicStateValue` struct so it is in a single location

### DIFF
--- a/layers/state_tracker/state_tracker.cpp
+++ b/layers/state_tracker/state_tracker.cpp
@@ -2840,9 +2840,9 @@ void ValidationStateTracker::PostCallRecordCmdSetViewport(VkCommandBuffer comman
     cb_state->viewportMask |= bits;
     cb_state->trashedViewportMask &= ~bits;
 
-    cb_state->dynamicViewports.resize(std::max(size_t(firstViewport + viewportCount), cb_state->dynamicViewports.size()));
+    cb_state->dynamic_state_value.viewports.resize(firstViewport + viewportCount);
     for (size_t i = 0; i < viewportCount; ++i) {
-        cb_state->dynamicViewports[firstViewport + i] = pViewports[i];
+        cb_state->dynamic_state_value.viewports[firstViewport + i] = pViewports[i];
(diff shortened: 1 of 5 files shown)


layers/state_tracker/state_tracker.cpp
```diff
@@ -2840,9 +2840,9 @@ void ValidationStateTracker::PostCallRecordCmdSetViewport(VkCommandBuffer comman
     cb_state->viewportMask |= bits;
     cb_state->trashedViewportMask &= ~bits;
 
-    cb_state->dynamicViewports.resize(std::max(size_t(firstViewport + viewportCount), cb_state->dynamicViewports.size()));
+    cb_state->dynamic_state_value.viewports.resize(firstViewport + viewportCount);
```
arno-lunarg, line 2843: The resizing logic is not the same, is it ok?

layers/state_tracker/state_tracker.cpp
```diff
@@ -5193,12 +5193,12 @@ void ValidationStateTracker::RecordCmdSetViewportWithCount(VkCommandBuffer comma
     uint32_t bits = (1u << viewportCount) - 1u;
     cb_state->viewportWithCountMask |= bits;
     cb_state->trashedViewportMask &= ~bits;
-    cb_state->viewportWithCountCount = viewportCount;
+    cb_state->dynamic_state_value.viewport_count = viewportCount;
     cb_state->trashedViewportCount = false;
 
-    cb_state->dynamicViewports.resize(std::max(size_t(viewportCount), cb_state->dynamicViewports.size()));
+    cb_state->dynamic_state_value.viewports.resize(viewportCount);
```
arno-lunarg, line 5199: Same here

arno-lunarg (approved):
LGTM, 1 question

PR #5989, opened 2023-06-12
tests: Add missing new lines

arno-lunarg:
Saw that when running ray tracing tests


tests/negative/ray_tracing.cpp
```diff
@@ -460,11 +460,11 @@ TEST_F(NegativeRayTracing, AccelerationStructureBindings) {
             printf(
```
juan-lunarg, line 460: Shouldn't this be using `GTEST_SKIP` instead of printf?

arno-lunarg: Those messages serve to tell that _parts_ of the global test may be skipped based on some conditions. To my knowledge, `GTEST_SKIP` only allows to skip the entire test?

juan-lunarg (approved):

PR #5993, opened 2023-06-12
cmake: Fix MSVC warning 4754

juan-lunarg:
This warning is actually very useful and pointed out a real issue with the code.

See the compiler docs for a detailed walkthrough:
https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4754?view=msvc-170

### DIFF
--- a/layers/core_checks/cc_shader.cpp
+++ b/layers/core_checks/cc_shader.cpp
@@ -3168,7 +3168,7 @@ bool CoreChecks::ValidateComputeWorkGroupSizes(const SHADER_MODULE_STATE &module
     }
 
     uint32_t limit = phys_dev_props.limits.maxComputeWorkGroupInvocations;
-    uint64_t invocations = local_size_x * local_size_y;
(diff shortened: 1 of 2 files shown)


layers/core_checks/cc_shader.cpp
```diff
@@ -3168,7 +3168,7 @@ bool CoreChecks::ValidateComputeWorkGroupSizes(const SHADER_MODULE_STATE &module
     }
 
     uint32_t limit = phys_dev_props.limits.maxComputeWorkGroupInvocations;
-    uint64_t invocations = local_size_x * local_size_y;
+    uint64_t invocations = static_cast<uint64_t>(local_size_x) * static_cast<uint64_t>(local_size_y);
     // Prevent overflow.
     bool fail = false;
     if (invocations > vvl::kU32Max || invocations > limit) {
```
juan-lunarg, line 3174: `invocations > vvl::kU32Max` was never going to be hit due to the aforementioned bug.

arno-lunarg (approved):

Issue #4258, opened 2022-06-26
VUID-vkCmdPipelineBarrier2-pDependencies-02285 reported for vkCmdWaitEvents2KHR


pierremoreau:
**Describe the Issue**
VUID-vkCmdPipelineBarrier2-pDependencies-02285 is being reported for a call to `vkCmdWaitEvents2KHR()`.
- This is confusing as `vkCmdPipelineBarrier2()` is not involved here.
- AFAICT, there is no requirement for there to be a self-dependency for `vkCmdWaitEvents2KHR()`, unlike `vkCmdPipelineBarrier2()`, when it is being called within a render pass.

**Valid Usage ID**
If applicable, please include the validation messages encountered leading up to the issue

> [error] { Validation } (VUID-vkCmdPipelineBarrier2-pDependencies-02285) Validation Error: [ VUID-vkCmdPipelineBarrier2-pDependencies-02285 ] Object 0: handle = 0xe88693000000000c, name = Deferred render pass, type = VK_OBJECT_TYPE_RENDER_PASS; | MessageID = 0x678997c4 | vkCmdWaitEvents2: Barriers cannot be set during subpass 1 of VkRenderPass 0xe88693000000000c[Deferred render pass] with no self-dependency specified. The Vulkan spec states: If fname:vkCmdPipelineBarrier2 is called within a render pass instance, the render pass must have been created with at least one VkSubpassDependency instance in VkRenderPassCreateInfo::pDependencies that expresses a dependency from the current subpass to itself, with synchronization scopes and access scopes that are all supersets of the scopes defined in this command (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdPipelineBarrier2-pDependencies-02285)

**Environment:**
 - OS: Linux
 - GPU: NVIDIA GTX 1080 Ti
 - SDK or header version if building from repo:
 - Options enabled (synchronization, best practices, etc.): `VK_LAYER_KHRONOS_validation`, `VK_KHR_synchronization2`

**Additional context**
Workflow:
```
┌ BeginCommandBuffer
│┌ CmdResetEvent2KHR
│├ BeginRenderPass // Render pass 1
││╶ Draw
│├ EndRenderPass
│├ CmdSetEvent2KHR
│├ BeginRenderPass // Render pass 2
││┌ Draw
││├ NextSubpass
││├ CmdWaitEvents2KHR
││└ Draw
[...]

PR #5995, opened 2023-06-12
layers: Do not check self-dependency in ValidateDependencyInfo

artem-lunarg:
* Removes duplicated call to `ValidateRenderPassPipelineBarriers`.
* Fixes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4258
* Renames `convert_to_renderpass2` files to `convert_utils` to use it as a general place for conversion routines.

DETAILS:

`ValidateRenderPassPipelineBarriers` is used to validate subpass self-dependencies. It can be called through these code paths:
1. By ValidateCmdPipelineBarrier2
2. By ValidateCmdSetEvent2/ValidateCmdWaitEvents2

In the first case we have two duplicate calls to `ValidateRenderPassPipelineBarriers`. It's called directly by `ValidateCmdPipelineBarrier2`, but also as part of `ValidateDependencyInfo`.

```mermaid
flowchart LR
    ValidateCmdPipelineBarrier2-->ValidateRenderPassPipelineBarriers
    ValidateDependencyInfo-->ValidateRenderPassPipelineBarriers
    ValidateCmdPipelineBarrier2 --> ValidateDependencyInfo
    ValidateCmdSetEvent2 --> ValidateDependencyInfo
    ValidateCmdWaitEvents2 --> ValidateDependencyInfo
```

In the second case a `VkEvent` is not allowed to form a self-dependency inside a subpass (according to spec). Because the code did that anyway it resulted in false-positive error messages (https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4258).

The solution is to remove `ValidateRenderPassPipelineBarriers` call from `ValidateDependencyInfo`.

With this fix the self-dependency validation is done by `PreCallValidateCmdPipelineBarrier` and `ValidateCmdPipelineBarrier2` that directly invoke `ValidateRenderPassPipelineBarriers`.

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -110,8 +110,8 @@ config("vulkan_layer_config") {
 
 core_validation_sources = [
   "layers/utils/android_ndk_types.h",
-  "layers/utils/convert_to_renderpass2.cpp",
-  "layers/utils/convert_to_renderpass2.h",
+  "layers/utils/convert_utils.cpp",
+  "layers/utils/convert_utils.h",
   "layers/core_checks/cc_android.cpp",
   "layers/core_checks/cc_buffer.cpp",
   "layers/core_checks/cc_buffer_address.h",
--- a/tests/positive/subpass.cpp
+++ b/tests/positive/subpass.cpp
@@ -1,0 +1,119 @@
+/*
+ * Copyright (c) 2023 Valve Corporation
+ * Copyright (c) 2023 LunarG, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "../framework/layer_validation_tests.h"
+
+TEST_F(PositiveSubpass, SubpassImageBarrier) {
+    TEST_DESCRIPTION("Subpass with image barrier (self-dependency)");
+    SetTargetApiVersion(VK_API_VERSION_1_3);
+    ASSERT_NO_FATAL_FAILURE(InitFramework());
+    if (DeviceValidationVersion() < VK_API_VERSION_1_3) {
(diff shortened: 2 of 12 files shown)


BUILD.gn
```diff
@@ -110,8 +110,8 @@ config("vulkan_layer_config") {
 
 core_validation_sources = [
   "layers/utils/android_ndk_types.h",
-  "layers/utils/convert_to_renderpass2.cpp",
-  "layers/utils/convert_to_renderpass2.h",
+  "layers/utils/convert_utils.cpp",
```
spencer-lunarg, line 113: I would love to call this `covert_to_version_2_of_a_vulkan_object_that_just_has_the_pNext.cpp`.... but for now happy with `convert_utils` as I can't think of a better name

spencer-lunarg (approved):

tests/positive/subpass.cpp
```diff
@@ -0,0 +1,121 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "../framework/layer_validation_tests.h"
+
+class PositiveSubpass : public VkLayerTest {};
```
spencer-lunarg, line 20: until we finalize the new naming scheme, can you rebase and move this to `layer_validation_tests.h` as

```
class SubpassTest : public VkLayerTest {};
class NegativeSubpass : public SubpassTest {};
class PositiveSubpass : public SubpassTest {};
``

BUILD.gn
artem-lunarg: yeah, pNext might be the weak link in the conversion algorithms. Regular field you can convert exactly in most cases. But what if old version supported some pNext types that new version does not support? It looks renderpass conversion selects only types that are supported in both v1 and v2 and skips pNext types that are only in v1. But it's hard to say how it changes the semantics, what the effect of the skipped types. It feels like regular fields should be converted just as we do now and maybe some additional control for pNext? Like do not copy when not requested explicitly or manually specify types that should be detect in pNext, hard to say what's the right approach or if it's an issue at all, just does not feel robust enough, but no intention to make any changes here :)

ncesario-lunarg (approved):

artem-lunarg:
I forgot to add a test that reproduces (https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4258) to demonstrate it is fixed. Will push one more change with a test (the test already added in this PR is to debug self-dependency but it does not work with events).

PR #5996, opened 2023-06-13
cmake: Fix MSVC warning 4459

juan-lunarg:
The main issue is the globals instance_extensions and device_extensions defined in chassis.cpp

This name is already used by ValidationObject which is what is causing the warning.
https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4459?view=msvc-170

This commit also removes the usage of ARRAY_SIZE in favor of std::array and removes extern "C" usage that didn't need to exist.

### DIFF
--- a/layers/utils/vk_layer_extension_utils.h
+++ b/layers/utils/vk_layer_extension_utils.h
@@ -18,17 +18,8 @@
 
 #include "vulkan/vulkan_core.h"
 
-#define ARRAY_SIZE(a) (sizeof(a) / sizeof(a[0]))
-
-/*
- * This file contains static functions for the generated layers
- */
-extern "C" {
-
 VkResult util_GetExtensionProperties(const uint32_t count, const VkExtensionProperties *layer_extensions, uint32_t *pCount,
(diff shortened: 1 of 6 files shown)


layers/utils/vk_layer_extension_utils.h
```diff
@@ -18,17 +18,8 @@
 
 #include "vulkan/vulkan_core.h"
 
-#define ARRAY_SIZE(a) (sizeof(a) / sizeof(a[0]))
-
-/*
- * This file contains static functions for the generated layers
- */
-extern "C" {
```
spencer-lunarg, line 26: totally happy with this, but question for future VUL repo, we plan to make that C or C++ based, I know at least 1 person in the world will complain

spencer-lunarg (approved):

arno-lunarg (approved):

layers/utils/vk_layer_extension_utils.h
juan-lunarg: VUL has a C interface

Issue #5938, opened 2023-05-31
VUID-VkSwapchainCreateInfoKHR-pNext-02679 validation too strict/incomplete


kasper93:
**Describe the Issue**

After update to 1.3.250, validation started complaining about `VUID-VkSwapchainCreateInfoKHR-pNext-02679`, the check has been added by https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/1baa836737b234dccb0d30f6c9afc84f474f35a6

```
Validation Error: [ VUID-VkSwapchainCreateInfoKHR-pNext-02679 ] Object 0: handle = 0x192ebca8160, type = VK_OBJECT_TYPE_DEVICE; Object 1: handle = 0xfab64d0000000002, type = VK_OBJECT_TYPE_SURFACE_KHR; | MessageID = 0x769aa5a9 | vkCreateSwapchainKHR(): pCreateInfo->pNext chain contains VkSurfaceFullScreenExclusiveInfoEXT, but does not contain VkSurfaceFullScreenExclusiveWin32InfoEXT. The Vulkan spec states: If the pNext chain includes a VkSurfaceFullScreenExclusiveInfoEXT structure with its fullScreenExclusive member set to VK_FULL_SCREEN_EXCLUSIVE_APPLICATION_CONTROLLED_EXT, and surface was created using vkCreateWin32SurfaceKHR, a VkSurfaceFullScreenExclusiveWin32InfoEXT structure must be included in the pNext chain (https://vulkan.lunarg.com/doc/view/1.3.250.0/windows/1.3-extensions/vkspec.html#VUID-VkSwapchainCreateInfoKHR-pNext-02679)
```

But it is clearly a false-positive, because we are setting it to `VK_FULL_SCREEN_EXCLUSIVE_DISALLOWED_EXT` in which case we do not have to link `VkSurfaceFullScreenExclusiveWin32InfoEXT` as the VUID clearly states. As we can see in the current VVL HEAD it only checks if both struct has been linked together without additional constraints.

Would you please review and improve this check to be compatible with the specification?

Thanks,
Kacper

PR #5998, opened 2023-06-13
layers: Fix 02679 validation

arno-lunarg:
- Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5938

Also add a regression test

### DIFF
--- a/tests/negative/wsi.cpp
+++ b/tests/negative/wsi.cpp
@@ -2856,6 +2856,56 @@ TEST_F(NegativeWsi, AcquireFullScreenExclusiveModeEXT) {
 #endif
 
 #if defined(VK_USE_PLATFORM_WIN32_KHR)
+TEST_F(NegativeWsi, CreateSwapchainFullscreenExclusive) {
+    TEST_DESCRIPTION("Test creating a swapchain with VkSurfaceFullScreenExclusiveWin32InfoEXT");
+
+    SetTargetApiVersion(VK_API_VERSION_1_2);
+
+    AddSurfaceExtension();
+    AddRequiredExtensions(VK_EXT_FULL_SCREEN_EXCLUSIVE_EXTENSION_NAME);
+
+    ASSERT_NO_FATAL_FAILURE(InitFramework());
+    if (!AreRequiredExtensionsEnabled()) {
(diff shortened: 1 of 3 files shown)


spencer-lunarg (approved):

juan-lunarg (changes requested):
The issue (https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5938) in question points to a false positive. Shouldn't a positive test be added?

arno-lunarg:
@juan-lunarg  The last few lines of the new test is the positive part, but I can just add a proper positive test if you want

tests/negative/wsi.cpp
```diff
@@ -2855,6 +2855,60 @@ TEST_F(NegativeWsi, AcquireFullScreenExclusiveModeEXT) {
...
+    VkSwapchainKHR swapchain = VK_NULL_HANDLE;
+
+    surface_full_screen_exlusive_info.fullScreenExclusive = VK_FULL_SCREEN_EXCLUSIVE_APPLICATION_CONTROLLED_EXT;
+    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkSwapchainCreateInfoKHR-pNext-02679");
+    vk::CreateSwapchainKHR(device(), &swapchain_create_info, nullptr, &swapchain);
+    m_errorMonitor->VerifyFound();
+
+    surface_full_screen_exlusive_info.fullScreenExclusive = VK_FULL_SCREEN_EXCLUSIVE_DEFAULT_EXT;
+    vk::CreateSwapchainKHR(device(), &swapchain_create_info, nullptr, &swapchain);
+    vk::DestroySwapchainKHR(device(), swapchain, nullptr);
```
juan-lunarg, line 2908: > @juan-lunarg The last few lines of the new test is the positive part, but I can just add a proper positive test if you want

A separate positive test would be my preference. Or adding a comment. My main nitpick is it's not immediately obvious that positive behavior is being tested.

It's obvious from the context of this PR. But it's not as obvious for a future test maintainer.

arno-lunarg: True, I am going to add a proper positive test

juan-lunarg (approved):

PR #6002, opened 2023-06-13
tests: Combine Test class helpers

spencer-lunarg:
Basically an attempt 2 for https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5983/

This attempt tries to just create a structure like 

```
class DynamicRenderingTest : public VkLayerTest {
  public:
    void InitBasicDynamicRendering();
};
class NegativeDynamicRendering : public DynamicRenderingTest {};
class PositiveDynamicRendering : public DynamicRenderingTest {};
```

Where `DynamicRenderingTest::InitBasicDynamicRendering` is just implemented in one of the test files

This way these tests can share the same common helpers


juan-lunarg (approved):

artem-lunarg (approved):

tests/framework/layer_validation_tests.h
```diff
@@ -571,6 +574,223 @@ class VkSyncValTest : public VkLayerTest {
...
+
+class FragmentShadingRateTest : public VkLayerTest {};
+class NegativeFragmentShadingRate : public FragmentShadingRateTest {};
+class PositiveFragmentShadingRate : public FragmentShadingRateTest {};
+
+class NegativeGeometryTessellation : public VkLayerTest {};
+
+class GraphicsLibraryTest : public VkLayerTest {
+  public:
+    void InitBasicGraphicsLibrary(void *pNextFeatures = nullptr);
```
juan-lunarg, line 643: The only issue I see with this approach is the ambiguity of where the function is implemented.

Seems like you chose to have the shared logic implemented in the negative cpp files. I think this is good enough for now, and we will see how this works out.

juan-lunarg: Although IMO it makes slightly more sense for helper logic to live in the positive cpp files. Since often shared functionality isn't meant to trigger invalid behavior.

spencer-lunarg: I guess my initial worry was "how do I run just the negative tests" but realized 

```
// Negative only
 --gtest_filter=Subpass.*
  
// Positive only
 --gtest_filter=PositiveSubpass.*
 
// Both
 --gtest_filter=*Subpass.*
```

So I agree, we should move towards that

spencer-lunarg: > Although IMO it makes slightly more sense for helper logic to live in the positive cpp files

Agree, let me just move it there for now

PR #6007, opened 2023-06-14
build: Update SPIRV-Tools to fix issue 5691

### LINKED ISSUE #5691: False error on out of bounds buffer device address read
**Environment:**
 - OS: Windows 10
 - GPU: RTX 4080
 - SDK or header version if building from repo: 1.3.236.0
 - Options enabled (synchronization, best practices, etc.): only gpu assisted validation is set on. everything else (including gpu assisted validation sub options) are turned off.

**Describe the Issue**

Gpu assisted validation false claimed an out of bounds buffer device address read.

**How to reproduce**

Download the daxa repository: https://github.com/Ipotrick/Daxa/tree/task_list_dev
commit: b20ac64bea9180e6a8603c78ace06ac88bb6366a
build it and run the sample daxa_test_4_hello_daxa_2_triangle with vulkan configurator open and gpu assisted validation turned on.

**Expected behavior**

I expect the validation to not give me any error as this is not an error from my perspective.

**Additional context**

When enabling gpu assisted validation i get a message claiming a read out of bounds. The given addresss is at an 48 byte offset to a 72 byte array of 3 structs, each 24 bytes large. It loads one of these 24 byute structs at that offset. So from my perspective this is not out of bounds, as 48 + 24 = 72, so within the bounds of the buffer size.

This works just fine with validation turned off!

```
UNASSIGNED-Device address out of bounds(ERROR / SPEC): msgNum: 445220389 - Validation Error: [ UNASSIGNED-Device address out of bounds ] Object 0: handle = 0x141278dac60, name = my device, type = VK_OBJECT_TYPE_QUEUE; | MessageID = 0x1a898625 | Device address 0xd080030 access out of bounds.  Command buffer (0x14128af7060). Draw Index 0. Pipeline (my pipeline)(0x1412aa1b7c0). Shader Module (0x141281561d0). Shader Instruction Index = 715.  Stage = Vertex. Vertex Index = 2 Instance Index = 0.  Shader validation error occurred in file ./tests/4_hello_daxa/2_triangle/main.glsl at line 18.Unable to find suitable #line directive in SPIR-V OpSource.
    Objects: 1
        [0] 0x141278dac60, type: 4, name: my device
```

jeremyg-lunarg:
This issue was caused by problems in the instrumentation code.

Fixes #5691


juan-lunarg (changes requested):
Looks like there is a new dependency?

```
CMake Error at external/CMakeLists.txt:106 (add_subdirectory):
  add_subdirectory given source "abseil_cpp" which is not an existing
  directory.
```

scripts/known_good.json
```diff
@@ -38,7 +38,7 @@
                 "-DSPIRV-Headers_SOURCE_DIR={repo_dir}/../SPIRV-Headers",
                 "-DSPIRV_WERROR=OFF"
```
juan-lunarg, line 39: Skipping tests appears to fix the build issue.

```suggestion
                "-DSPIRV_WERROR=OFF",
                "-DSPIRV_SKIP_TESTS=ON"
```

spencer-lunarg:
> Looks like there is a new dependency?

https://github.com/KhronosGroup/SPIRV-Tools/blob/main/CMakeLists.txt#L299

I realize we should be really skipping the test building `-D SPIRV_SKIP_TESTS` ... I didn't realize until SPIRV-Hopper issue that building the tests were the default for spirv-tools

scripts/known_good.json
juan-lunarg: Actually please skip building the tests and skip building the executables. We only need the libraries.

```
        {
            "name": "SPIRV-Tools",
            "url": "https://github.com/KhronosGroup/SPIRV-Tools.git",
            "sub_dir": "SPIRV-Tools",
            "build_dir": "SPIRV-Tools/build",
            "install_dir": "SPIRV-Tools/build/install",
            "cmake_options": [
                "-DSPIRV-Headers_SOURCE_DIR={repo_dir}/../SPIRV-Headers",
                "-DSPIRV_WERROR=OFF",
                "-DSPIRV_SKIP_TESTS=ON",
                "-DSPIRV_SKIP_EXECUTABLES=ON"
            ],
            "commit": "d33bea584707c3509065010c45d9b79c54c8376b"
        },
```

spencer-lunarg: aww, beat me to finding this... yes, do this :100:

scripts/known_good.json
```diff
@@ -36,9 +36,11 @@
             "install_dir": "SPIRV-Tools/build/install",
             "cmake_options": [
                 "-DSPIRV-Headers_SOURCE_DIR={repo_dir}/../SPIRV-Headers",
-                "-DSPIRV_WERROR=OFF"
+                "-DSPIRV_WERROR=OFF",
+                "-DSPIRV_SKIP_TESTS=ON",
+                "-DSPIRV_SKIP_EXECUTABLES=ON"
```
juan-lunarg, line 41: Hope you don't mind but I tested locally and this worked fine. So I just pushed to your branch.

spencer-lunarg (approved):

scripts/known_good.json
jeremyg-lunarg: Thank you. We were warned about this new dependency by the SPIRV-Tools people. I just happened to pick up something on top of the change that brought it in. But better to find it now than during SDK building.

juan-lunarg: https://gitlab.khronos.org/vulkan/Vulkan-SDK-Packaging/-/issues/1313

juan-lunarg (approved):

juan-lunarg:
I should have asked initially @jeremyg-lunarg but do we have a way to prevent a regression for this issue? Either a unit test in VVL or in SPIRV-Tools?

jeremyg-lunarg:
> I should have asked initially @jeremyg-lunarg but do we have a way to prevent a regression for this issue? Either a unit test in VVL or in SPIRV-Tools?

I added unit tests in SPIRV-Tools.

juan-lunarg:
> I added unit tests in SPIRV-Tools.

💯

Issue #4968, opened 2022-12-15
Timeout in `vkGetSemaphoreCounterValue` and `vkWaitSemaphores`


glebov-andrey:
Hi, while testing the fixes for #4308 and #4733, I've encountered an internal timeout when calling `vkGetSemaphoreCounterValue` and `vkWaitSemaphores`.
~~As far as I can tell, it only occurs with GPU-Assisted validation enabled.~~
UPD: On Windows I'm getting the error only in `vkGetSemaphoreCounterValue` and only with GPU-Assisted validation enabled. On Linux I'm also getting it in `vkWaitSemaphores` with Synchronization Validation enabled.

Here's the error message and stack trace:
```
[ UNASSIGNED-VkSemaphore-state-timeout ] Object 0: handle = 0xee647e0000000009, name = Transfer queue semaphore, type = VK_OBJECT_TYPE_SEMAPHORE; | MessageID = 0x57e65a33 | Timeout waiting for timeline semaphore state to update. This is most likely a validation bug. completed_.payload=0 wait_payload=1
```
```
VkLayer_khronos_validation.dll!SEMAPHORE_STATE::NotifyAndWait(unsigned __int64 payload) Line 515
	at E:\dev\sandbox\Vulkan-ValidationLayers\layers\queue_state.cpp(515)
VkLayer_khronos_validation.dll!ValidationStateTracker::RecordGetSemaphoreCounterValue(VkDevice_T * device, VkSemaphore_T * semaphore, unsigned __int64 * pValue, VkResult result) Line 1976
	at E:\dev\sandbox\Vulkan-ValidationLayers\layers\state_tracker.cpp(1976)
VkLayer_khronos_validation.dll!ValidationStateTracker::PostCallRecordGetSemaphoreCounterValue(VkDevice_T * device, VkSemaphore_T * semaphore, unsigned __int64 * pValue, VkResult result) Line 1983
	at E:\dev\sandbox\Vulkan-ValidationLayers\layers\state_tracker.cpp(1983)
VkLayer_khronos_validation.dll!vulkan_layer_chassis::GetSemaphoreCounterValue(VkDevice_T * device, VkSemaphore_T * semaphore, unsigned __int64 * pValue) Line 4553
	at E:\dev\sandbox\Vulkan-ValidationLayers\layers\generated\chassis.cpp(4553)
```
The stack trace for `vkWaitSemaphores` is pretty much the same:
```
[...]

PR #6010, opened 2023-06-15
wip: Add tests from issue 4968

artem-lunarg:
That's the tests provided by @glebov-andrey in https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4968 and adjusted to run under VVL test framework.

The tests were further simplified. Currently both of them have 100% repro case on the development windows machine and also fail on the internal CI.

`PositiveGpuAssistedLayer.GetCounterFromSignaledSemaphore`
Fails under **GPU AV** validation.
Fails on **Windows**.
**Passes mock icd** run
No failures detected when it was modified to run under Regular or SyncVal validation.
Comparing to the original test no iterations are needed to reproduce the failure.

`PositiveSyncVal, SignalAndWaitSemaphoreFromHost`
Fails under **SyncVal** validation.
Fails  both on **Linux and Windows**.
**Does NOT pass mock icd** run (with timeout error as on a regular driver).
No failures detected when it was modified to run under Regular or GPU AV validation.
This test still needs iterations. 100'000 runs looks like a minimum limit, in some cases it managed to pass 10'000 runs.

### DIFF
--- a/layers/state_tracker/queue_state.cpp
+++ b/layers/state_tracker/queue_state.cpp
@@ -25,7 +25,7 @@ using SemOp = SEMAPHORE_STATE::SemOp;
 // (via being in a PostRecord call) that a fence, semaphore or wait for idle has
 // completed. Hitting it is almost a certainly a bug in this code.
 static std::chrono::time_point<std::chrono::steady_clock> GetCondWaitTimeout() {
-    return std::chrono::steady_clock::now() + std::chrono::seconds(10);
+    return std::chrono::steady_clock::now() + std::chrono::seconds(2);
 }
 
 void CB_SUBMISSION::BeginUse() {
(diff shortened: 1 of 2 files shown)


tests/positive/gpu_av.cpp
```diff
@@ -226,3 +226,42 @@ TEST_F(PositiveGpuAssistedLayer, SetSSBOPushDescriptor) {
 
     vk::DestroyPipeline(m_device->device(), pipeline, nullptr);
 }
+
+TEST_F(PositiveGpuAssistedLayer, Issue_4968_vkGetSemaphoreCounterValue) {
```
spencer-lunarg, line 230: I would just add a comment above with link to issue, try to give a name that somewhat describes what the test is doing

glebov-andrey:
I should note that Vulkan 1.3 is most likely not required here. After all, timeline semaphores have been around since before 1.2 as an extension.
The only reason my reproducers use 1.3 features like `synchronization2` is because they are based on code which itself uses Vulkan 1.3.

artem-lunarg:
Updated Vulkan version the tests use:
SignalAndWaitSemaphoreFromHost: 1.2 is enough
GetCounterFromSignaledSemaphore: needs 1.3 due to vkQueueSubmit2

artem-lunarg:
`PositiveGpuAssistedLayer.GetCounterFromSignaledSemaphore` was added as part of https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6079

artem-lunarg:
This PR was a development helper, and the tests are integrated into production PRs now.

PR #6011, opened 2023-06-15
layers: Avoid warnings from std::tolower

jeremyg-lunarg:
Using this function with std::transform on a std::string can result in truncation from int to char warnings with some compilers.


juan-lunarg (approved):

jeremyg-lunarg:
I was going to ask if anyone prefers a different .h file for the new functions. I just guessed.

juan-lunarg (approved):

PR #6015, opened 2023-06-15
Android script + BUILD.md cleanup

juan-lunarg:


scripts/android.py
```diff
@@ -0,0 +1,149 @@
...
+import argparse
+import subprocess
+import sys
+import common_ci
+
+# TODO: Use CMake from Android SDK?
+
+# Android APKs can contain binaries for multiple ABIs (armeabi-v7a, arm64-v8a, x86, x86_64).
+# CMake will need to be run multiple times to create a complete test APK that can run on any Android device.
+def main():
```
juan-lunarg, line 27: Note. This script is pretty rough right now but I am able to successfully create Android binaries with CMake and create the test `VulkanLayerValidationTests.apk`

image

scripts/android.py
```diff
@@ -0,0 +1,155 @@
...
+
+    print(f"ANDROID_NDK_HOME = {android_ndk_home}")
+    print(f"ANDROID_SDK_ROOT = {android_sdk_root}")
+    print(f'Android cmake toolchain file = {android_toolchain}')
+
+    # TODO: Verify aapt, ndk-build, cmake, ninja are in the path.
+
+    # TODO: Allow user to specify abis to save time.
+    # android_abis = ["armeabi-v7a", "arm64-v8a", "x86", "x86_64"]
+    android_abis = [ "armeabi-v7a", "arm64-v8a" ]
```
spencer-lunarg, line 46: honestly, I only ever build `arm64-v8a` locally, 32-bit is something, like `x86` I think people are opting into now

scripts/android.py
```diff
@@ -0,0 +1,155 @@
...
+    # Align APK
+    common_ci.RunShellCmd(f'zipalign -f 4 {unaligned_apk} {aligned_apk}')
+
+    
+
+    # If zipalign was run after signing, it won't be a valid signature
+
+    # TODO: Figure out how to sign APK. Looks like jarsigner is deprecated.
+    #
+    # jarsigner used to be used until it was removed from the Android SDK
```
spencer-lunarg, line 119: this seems left in by accident

spencer-lunarg (approved):
few questions, but looks good to me

scripts/android.py
juan-lunarg: I currently am just making sure the APK has multiple binaries. I just picked the 2 arms ABIs arbitrarily for now to save time compiling.

scripts/android.py
juan-lunarg: I'm going to do a follow up with the signing logic. I want this script to be flexible enough for vvl and jenkins.

scripts/android.py
juan-lunarg: IE by default I think they should all be on. However, I want to give users the ability to specify how many they want in another PR.

scripts/android.py
juan-lunarg: I've removed the comments.

scripts/android.py
juan-lunarg: Or maybe the default should just be `arm64-v8a`... I'm not sure yet to be honest. This is still WIP

spencer-lunarg: I guess its like why we don't build both 32 and 64 by default for Linux/Windows... you should only pick what you need, one is enough for most developers who want to test on their local device

agree, should make sure we build 32-bit ARM in CI though

PR #6016, opened 2023-06-16
scripts: Initial version of VulkanObject in script

spencer-lunarg:
Successor to https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5866

This tries to wrap the `reg.py` with the new `VulkanObject` class found in `scripts/

- `base_generator.py` does all the ugly "xml to Python class" logic
- `vulkan_object.py` is designed to look like a C header file
    - When debugging, you can now even print all the values, no more getting random XML elem blocks
- Took the simple file (`dynamic_state_generator.py`) as a test, it now has a single `def generate(self):` entry point

There is future work, but the first step is to move everything over to use `VulkanObject` (see what is missing) and then from there, build a better "common logic" around `VulkanObject`


scripts/generators/vulkan_object.py
```diff
@@ -0,0 +1,259 @@
...
+class VulkanObject():
+    extensions: Dict[str, Extension] = field(default_factory=dict, init=False)
+    versions:   Dict[str, Version]   = field(default_factory=dict, init=False)
+
+    commands: Dict[str, Command]     = field(default_factory=dict, init=False)
+    enums:    Dict[str, Enum]        = field(default_factory=dict, init=False)
+    structs:  Dict[str, Struct]      = field(default_factory=dict, init=False)
+    unions:   Dict[str, Union]       = field(default_factory=dict, init=False)
+
+    formats: List[Format]            = field(default_factory=list, init=False)
```
charles-lunarg, line 246: Why not have this be a dict where the key is the 'format enum'? (just thinking outloud about how I might want to access this list.) Very much just thinking out loud.

charles-lunarg (commented):
Looks great overall. Have some thoughts about what I see, all of which are nitpicks or spur off the moment thoughts.

scripts/generators/base_generator.py
```diff
@@ -65,8 +80,302 @@ def __init__(self,
...
+        platforms = self.registry.tree.findall('platforms/platform')
+        for platform in platforms:
+            self.vk.platforms[platform.get('name')] = platform.get('protect')
+
+        tags = self.registry.tree.findall('tags')
+        for tag in tags:
+            self.vk.vendorTags.append(tag.get('name'))
+
+    # This function should be overloaded
+    def generate(self):
```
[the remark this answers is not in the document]
spencer-lunarg: ya, can print an error saying "ummm, don't call this actually"

scripts/generators/base_generator.py
```diff
@@ -65,8 +80,302 @@ def __init__(self,
...
+            device = not instance
+            depends = interface.get('depends')
+            vendorTag = interface.get('author')
+            platform = interface.get('platform')
+            provisional = boolGet(interface, 'provisional')
+            promotedto = interface.get('promotedto')
+            deprecatedby = interface.get('deprecatedby')
+            obsoletedby = interface.get('obsoletedby')
+            specialuse = splitIfGet(interface, 'specialuse')
+            self.currentFeature = Extension(name, instance, device, depends, vendorTag, platform,
```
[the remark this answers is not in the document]
spencer-lunarg: I think this was left over from refactoring, agree, can just be local

scripts/generators/vulkan_object.py
```diff
@@ -0,0 +1,259 @@
...
+# not supported until Python 3.9+
+from typing import *
+
+@dataclass(frozen=True)
+class Extension:
+    """<extension>"""
+    name: str
+    # Only one will be True, the other is False
+    instance: bool
+    device: bool
```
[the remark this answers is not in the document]
spencer-lunarg: agree, enum would be better

scripts/generators/vulkan_object.py
spencer-lunarg: thinking out loud to... I see not real advantage keeping this as a List, can just iterate the `keys` to do the same

docs/generated_code.md
```diff
@@ -21,7 +21,7 @@ cmd /C "python3 scripts/generate_source.py external/Vulkan-Headers/registry/ ext
 ```
 
 When making change to the `scripts/` folder, make sure to run `generate_source.py` and check in both the changes to
-`scripts/` and `layers/vulkan/generated/` in any PR.
+`scripts/` and `layers/vulkan/generated/` in any PR. (Code generation does **not** happen automatically at build time.)
```
juan-lunarg, line 24: Do we need to state this?

juan-lunarg (commented):
LGTM I'll defer to @charles-lunarg and @mikes-lunarg though since they are a lot more familiar with these scripts.

docs/generated_code.md
spencer-lunarg: we have had a handful of external contributor PR fixing the scripts, just to ask them to then run `generate_source.py`

I know various projects that generate at build if it detects changes in script file

scripts/generators/vulkan_object.py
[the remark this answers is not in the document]
spencer-lunarg: actually I lied, rather have some "sanity check" logic... my issue here is what would this variable be called that could only take an enum of instance/device... plus its an extra indirect-lookup for the off chance the one spot in the code that creates this `Extension` is doing it wrong

scripts/generators/vulkan_object.py
```diff
@@ -0,0 +1,259 @@
...
+    #   VkInstance* pInstance);'
+    cPrototype: str
+
+    # function pointer typedef  - ex:
+    # typedef VkResult (VKAPI_PTR *PFN_vkCreateInstance)
+    #   (const VkInstanceCreateInfo* pCreateInfo, const VkAllocationCallbacks* pAllocator, VkInstance* pInstance);'
+    cFunctionPointer: str
+
+@dataclass(frozen=True)
+class Enum:
```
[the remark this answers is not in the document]
spencer-lunarg: done!

mikes-lunarg (approved):
I like this a lot. I wish we would have done something like this years ago. Obviously it will need to grow over time to support the different generators, but this seems like a very functional first step.

scripts/generators/vulkan_object.py
```diff
@@ -0,0 +1,274 @@
...
+    name: str # VK_VERSION_1_0
+    number: str # 1.0
+
+@dataclass(frozen=True)
+class CommandParam:
+    """<command/param>"""
+    name: str
+    type: str
+    alias: str
+    externSync: str
```
artem-lunarg, line 53: can `externSync` be `bool`?

scripts/generators/vulkan_object.py
```diff
@@ -0,0 +1,274 @@
...
+    feature: object
+
+    returnType: str # 'void', 'VkResult', etc
+
+    api: List[str]            # ex. [ 'vulkan' ]
+    tasks: List[str]          # ex. [ 'action', 'state', 'synchronization' ]
+    queues: List[str]         # ex. [ 'graphics', 'compute' ]
+    successCodes: List[str]   # ex. [ 'VK_SUCCESS', 'VK_INCOMPLETE' ]
+    errorCodes: List[str]     # ex. [ 'VK_ERROR_OUT_OF_HOST_MEMORY' ]
+    cmdBufferLevel: List[str] # ex. [ 'primary', 'secondary' ]
```
artem-lunarg, line 74: `cmdBufLevel`: two bool values (for primary and secondary support) might be easier to work with

scripts/generators/vulkan_object.py
```diff
@@ -0,0 +1,274 @@
...
+    bitWidth: int # 32 or 64
+    fields: List[Enum]
+
+@dataclass(frozen=True)
+class FlagBit:
+    """<enum> of type bitmask"""
+    name: str
+    # one of 'bitPosition' or 'value' will be None, the other is valid value
+    bitPosition: int # Which bit in flag
+    value: int # If multi-bit field, will be value for it (ex. VK_SHADER_STAGE_ALL_GRAPHICS)
```
artem-lunarg, line 113: Is `value` only for multi-bit fields? If yes, should this be reflected in the name somehow? Without checking implementation details `value` name sounds general enough, so I assume that for single bit field it's also corresponds to numeric values of the bit pattern (basically the same as for multi-bit). But if it's not the case then such general name might be confusing.

scripts/generators/vulkan_object.py
spencer-lunarg: yes!

scripts/generators/vulkan_object.py
spencer-lunarg: agree

scripts/generators/vulkan_object.py
spencer-lunarg: it is only for multi-bit

maybe better just having a single "value" and a `bool` for `multi-bit`

artem-lunarg (approved):

scripts/generators/dynamic_state_generator.py
```diff
@@ -15,34 +15,33 @@
...
-    #
-    # Called at beginning of processing as file is opened
-    def beginFile(self, genOpts):
-        OutputGenerator.beginFile(self, genOpts)
-        self.headerFile = (genOpts.filename == 'dynamic_state_helper.h')
-        self.sourceFile = (genOpts.filename == 'dynamic_state_helper.cpp')
+    def generate(self):
+        self.headerFile = (self.filename == 'dynamic_state_helper.h')
+        self.sourceFile = (self.filename == 'dynamic_state_helper.cpp')
+        for field in self.vk.enums['VkDynamicState'].fields:
```
jzulauf-lunarg, line 38: More pythonic is:

`self.dynamic_states = [ field.name for field in self.vk.enums['VkDynamicState'].fields]`

scripts/generators/dynamic_state_generator.py
```diff
@@ -15,34 +15,33 @@
...
+            self.dynamic_states.append(field.name)
 
         # File Comment
         file_comment = '// *** THIS FILE IS GENERATED - DO NOT EDIT ***\n'
         file_comment += '// See {} for modifications\n'.format(os.path.basename(__file__))
-        write(file_comment, file=self.outFile)
+        self.write(file_comment)
         # Copyright Statement
```
jzulauf-lunarg, line 45: Should migrate copyright generation into base generator -- we'll be happier we did sometime in approximately January.

jzulauf-lunarg: Even if we have to add a "starting year" parameter

scripts/generators/dynamic_state_generator.py
spencer-lunarg: can fix, I plan to do another pass through the `dynamic_state_generator`... I just needed "a" file that invoked VulkanObject and this was the simplest

PR #6018, opened 2023-06-16
layers: Fix null image view state dereferencing

arno-lunarg:
The test introduced here leads to null image view states that would cause null pointers dereferencing in `CoreChecks::PreCallValidateCreateFramebuffer`, `FRAMEBUFFER_STATE::LinkChildNodes()` and `FRAMEBUFFER_STATE::Destroy()`. 
Before adding null checks, just bumping the Vulkan API version for this test from 1.0 to 1.1 would cause image view states to not be null.
Null checks are added to prevent crashes.

Internal discussion: https://chat.google.com/room/AAAAOXVAYGg/cGoDLJK09hI

### DIFF
--- a/tests/negative/renderpass.cpp
+++ b/tests/negative/renderpass.cpp
@@ -4677,3 +4677,81 @@ TEST_F(NegativeRenderPass, SubpassAttachmentImageLayoutSeparateDepthStencil) {
         }
     }
 }
+
+TEST_F(NegativeRenderPass, NullImageViewState) {
+    TEST_DESCRIPTION(
+        "This test lead to null image view states that would cause null pointers dereferencing in "
+        "CoreChecks::PreCallValidateCreateFramebuffer, FRAMEBUFFER_STATE::LinkChildNodes() and FRAMEBUFFER_STATE::Destroy(). "
+        "Before adding null checks, just bumping the Vulkan API version for this test from 1.0 to 1.1 would cause image view "
+        "states to not be null");
+
+    ASSERT_NO_FATAL_FAILURE(Init());
+
+    // A renderpass with one color attachment.
+    VkAttachmentDescription attachment = {0,
+                                          VK_FORMAT_R8G8B8A8_UNORM,
+                                          VK_SAMPLE_COUNT_1_BIT,
+                                          VK_ATTACHMENT_LOAD_OP_DONT_CARE,
+                                          VK_ATTACHMENT_STORE_OP_STORE,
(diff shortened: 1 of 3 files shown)


tests/negative/renderpass.cpp
```diff
@@ -4677,3 +4677,81 @@ TEST_F(NegativeRenderPass, SubpassAttachmentImageLayoutSeparateDepthStencil) {
...
+    m_errorMonitor->VerifyFound();
+
+    VkFramebufferCreateInfo fci = LvlInitStruct<VkFramebufferCreateInfo>();
+    fci.renderPass = rp;
+    fci.attachmentCount = 1;
+    fci.pAttachments = &array_2d_view_layer_0.handle();
+    fci.width = 32;
+    fci.height = 32;
+    fci.layers = 1;
+    vk_testing::Framebuffer fb(*m_device, fci);  // Crash when validation framebuffer creation
```
spencer-lunarg, line 4756: I don't want a crash to be expected, I want a VUID to be caught

arno-lunarg: I need to re-write this: a more accurate comment would be `Used to crash because of null pointers to image view states`

But actually this is still crashing, but for other reasons it seems. I do not know them yet

artem-lunarg:
```
void FRAMEBUFFER_STATE::LinkChildNodes() {
    for (auto &view : attachments_view_state) {
        if (view) {
            view->AddParent(this);
        }
    }
}
```

Sorry, I did not dig into details of this PR but just a comment about aspect that I personally found can make life much harder or much simpler, and if it is not what this PR is about then please ignore.

The above function is quite central to the system architecture and that's the reason why it's important that the programmer should be able to understand it very well. If nulls are allowed it probably should be part of another design document or system-wide convention that describes this, or at least if it's a special case it should be a good comment. IMO `if (something)` can be the most complex part of some non-trivial code.

So the idea is to have good answers to the necessity of each `if (someObjectPtr)` check (general conventions + well documented special cases). Trying to write more defensive code for such unclear places, not necessarily without downsides, since new code can have its own issues.

arno-lunarg: You are right, and anyway thinking more about the problem adding nulls checks in `FRAMEBUFFER_STATE` is not the solution. Issue seems deeper: no validation error is triggered in this test for the framebuffer being created with a NULL image view...

PR #6028, opened 2023-06-19
tests: Move and rename test shader templates

spencer-lunarg:
1. the name `bindStateVertShaderText` is not helpful at all (or others related)
2. the `layer_validation_tests.h` is starting to become a "throw everything in it" file which I rather try and sort it out little by little


tests/framework/shader_templates.h
```diff
@@ -0,0 +1,291 @@
...
+        float gl_PointSize;
+    };
+    void main() {
+       gl_Position = vec4(1);
+       gl_PointSize = 1.0;
+       EmitVertex();
+    }
+)glsl";
+
+static const char kTscMinimalGlsl[] = R"glsl(
```
arno-lunarg, line 66:
```suggestion
static const char kTesselationControlMinimalGlsl[] = R"glsl(
```
Given the other shader template names, if I was to intuitively try find a tessellation shader template, I would look for this name.
But yeah it is longer...

tests/framework/shader_templates.h
```diff
@@ -0,0 +1,291 @@
...
+    #version 460
+    layout(vertices=3) out;
+    void main() {
+       gl_TessLevelOuter[0] = gl_TessLevelOuter[1] = gl_TessLevelOuter[2] = 1;
+       gl_TessLevelInner[0] = 1;
+    }
+)glsl";
+
+static const char kTesMinimalGlsl[] = R"glsl(
```
arno-lunarg, line 75:
```suggestion
static const char kTessellationEvaluationMinimalGlsl[] = R"glsl(
```
or
```suggestion
static const char kTessellationEvalMinimalGlsl[] = R"glsl(
```

arno-lunarg (commented):
LGTM, 2 comments!

tests/framework/shader_templates.h
spencer-lunarg: I would be lying if 90% of this decision was because I didn't want to look up how to spell `Tessellation`... will fix though

tests/negative/arm_best_practices.cpp
```diff
@@ -919,10 +919,8 @@ TEST_F(VkArmBestPracticesLayerTest, RedundantRenderPassStore) {
 
     CreatePipelineHelper graphics_pipeline(*this);
 
-    graphics_pipeline.vs_ =
-        std::unique_ptr<VkShaderObj>(new VkShaderObj(this, bindStateVertShaderText, VK_SHADER_STAGE_VERTEX_BIT));
-    graphics_pipeline.fs_ =
-        std::unique_ptr<VkShaderObj>(new VkShaderObj(this, bindStateFragSamplerShaderText, VK_SHADER_STAGE_FRAGMENT_BIT));
+    graphics_pipeline.vs_ = std::unique_ptr<VkShaderObj>(new VkShaderObj(this, kVertexMinimalGlsl, VK_SHADER_STAGE_VERTEX_BIT));
```
juan-lunarg, line 922: Use `std::make_unique` instead.

https://clang.llvm.org/extra/clang-tidy/checks/modernize/make-unique.html

juan-lunarg (approved):

tests/negative/arm_best_practices.cpp
spencer-lunarg: I see we do this all over... will make a PR

PR #6032, opened 2023-06-19
Use a single buffer for buffer device address GPU-AV

Tony-LunarG:
Should be better than 1 buffer per draw, and now accounts for buffers created after draw is recorded

### DIFF
--- a/layers/gpu_validation/gpu_validation.cpp
+++ b/layers/gpu_validation/gpu_validation.cpp
@@ -132,6 +132,29 @@ void GpuAssisted::CreateDevice(const VkDeviceCreateInfo *pCreateInfo) {
                               IsExtEnabled(device_extensions.vk_khr_buffer_device_address)) &&
                              shaderInt64 && enabled_features.core12.bufferDeviceAddress);
 
+    if (buffer_device_address) {
+        const char *size_string = getLayerOption("khronos_validation.gpuav_max_buffer_device_addresses");
+        app_bda_max_addresses = *size_string ? atoi(size_string) : 10000;
+        VkBufferCreateInfo buffer_info = LvlInitStruct<VkBufferCreateInfo>();
+        buffer_info.usage = VK_BUFFER_USAGE_STORAGE_BUFFER_BIT;
+        VmaAllocationCreateInfo alloc_info = {};
+        // We need 2 words per address (address and size), 1 word for the start of sizes index, 2 words for the address section
+        // bounds, and 2 more words for the size section bounds
+        app_bda_buffer_size = (1 + (app_bda_max_addresses + 2) + (app_bda_max_addresses + 2)) * 8;  // 64 bit words
+        buffer_info.size = app_bda_buffer_size;
+        // This buffer could be very large if an application uses many buffers. Allocating it as HOST_CACHED
+        // and manually flushing it at the end of the state updates is faster than using HOST_COHERENT.
+        alloc_info.requiredFlags = VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_CACHED_BIT;
+        VkResult result = vmaCreateBuffer(vmaAllocator, &buffer_info, &alloc_info, &app_buffer_device_addresses.buffer,
+                                          &app_buffer_device_addresses.allocation, nullptr);
+        if (result != VK_SUCCESS) {
+            ReportSetupProblem(
+                device, "Unable to allocate device memory for buffer device address data.  Device could become unstable.", true);
+            aborted = true;
+            return;
+        }
+    }
+
     if (IsExtEnabled(device_extensions.vk_ext_shader_object)) {
         LogWarning(device, "UNASSIGNED-GPU-Assisted Validation Warning",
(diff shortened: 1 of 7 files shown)


layers/gpu_validation/gpu_validation.cpp
```diff
@@ -133,6 +133,26 @@ void GpuAssisted::CreateDevice(const VkDeviceCreateInfo *pCreateInfo) {
                               IsExtEnabled(device_extensions.vk_khr_buffer_device_address)) &&
                              shaderInt64 && enabled_features.core12.bufferDeviceAddress);
 
+    if (buffer_device_address) {
+        const char *size_string = getLayerOption("khronos_validation.gpuav_max_buffer_device_addresses");
+        app_bda_max_addresses = *size_string ? atoi(size_string) : 1024;
+        VkBufferCreateInfo buffer_info = LvlInitStruct<VkBufferCreateInfo>();
+        buffer_info.usage = VK_BUFFER_USAGE_STORAGE_BUFFER_BIT;
+        VmaAllocationCreateInfo alloc_info = {};
+        app_bda_buffer_size = ((app_bda_max_addresses + 3) + (app_bda_max_addresses + 2)) * 8; // 64 bit words
```
jeremyg-lunarg, line 142: the magic  3 and 2 here could use a comment.

jeremyg-lunarg (commented):
this looks pretty good to me! Any performance data?

Tony-LunarG:
> this looks pretty good to me! Any performance data?

Sadly, none of our performance workloads use BDA.  The single allocation has to be faster than multiple allocations, so command buffer recording should be faster.  The question is whether the BDA address table changes often, which would cause overhead in QueueSubmit.

Tony-LunarG:
Adding @christophe-lunarg for new setting

tests/negative/gpu_av.cpp
```diff
@@ -1204,6 +1204,15 @@ TEST_F(VkGpuAssistedLayerTest, GpuBufferDeviceAddressOOB) {
                 m_errorMonitor->VerifyFound();
             }
         }
+        // Positive test using a buffer created after the draw was recorded
```
juan-lunarg, line 1207: `VkGpuAssistedLayerTest.GpuBufferDeviceAddressOOB` is currently blacklisted by Jenkins for 2 CI machines.

Please break up the positive test into a new unit test.

layers/gpu_validation/gpu_validation.cpp
Tony-LunarG: Added a comment

tests/negative/gpu_av.cpp
Tony-LunarG: Created a new positive unit test

layers/gpu_validation/gpu_validation.cpp
```diff
@@ -1255,6 +1278,58 @@ void GpuAssisted::UpdateInstrumentationBuffer(gpuav_state::CommandBuffer *cb_nod
     }
 }
 
+void GpuAssisted::UpdateBDABuffer(GpuAssistedDeviceMemoryBlock device_address_buffer) {
+    if (gpuav_bda_buffer_version == buffer_device_address_ranges_version) {
+        return;
+    }
+    auto address_ranges = GetBufferAddressRanges();
+    auto address_ranges_num_addresses = address_ranges.size();
+    if (address_ranges_num_addresses > 0) {
```
juan-lunarg, line 1287: Complete nitpick FFTI.

Early return to avoid indenting most of the logic in this function.

```suggestion
    if (address_ranges_num_addresses == 0) { // Assuming address_ranges_num_addresses is unsigned
        return;
    }
```

juan-lunarg (approved):
LGTM

layers/vk_layer_settings.txt
```diff
@@ -104,6 +104,12 @@ khronos_validation.enables =
 # Use VMA linear memory allocations for GPU-AV output buffers
 #khronos_validation.vma_linear_output = true
 
+# Specify the maximum number of buffer device addresses in simultaneous use
+# =====================
+# <LayerIdentifier>.gpuav_max_buffer_device_addresses
+# Specify the maximum number of buffer device addresses to allow GPU-AV allocate resources
+#khronos_validation.gpuav_max_buffer_device_addresses = 1024
```
jeremyg-lunarg, line 111: based on the test app we got recently this might not be large enough. I'm not sure what a good value would be. 
https://vulkan.gpuinfo.org/displaydevicelimit.php?name=maxMemoryAllocationCount&platform=all is one idea but it can be extremely large on some devices.

jeremyg-lunarg: @arno-lunarg what do you think?

arno-lunarg: I share your opinion, could be too small

jeremyg-lunarg: what would a good value be?

christophe-lunarg (approved):

layers/vk_layer_settings.txt
christophe-lunarg: In the code the actual default values is 10000 I believe

Tony-LunarG: Yes, without much more evidence to go on, I picked 10000

PR #6034, opened 2023-06-20
scripts: Move SpirvValidationHelper to VulkanObject

spencer-lunarg:
Most of the logic replaced was

- The old way we had to gather up information that is just now in `VulkanObject::SpirvEnable`
- use f-string and array of strings
- Remove the functions as they aren't helpful to bounce around IMO from editing this file in the past

The generated code has a lot of changes, but it is just the ordering, it is now just ordered the way it is seen in the `vk.xml` instead of the arbitrary `sorted()` order from before

### DIFF
--- a/scripts/generators/spirv_validation_generator.py
+++ b/scripts/generators/spirv_validation_generator.py
@@ -15,21 +15,18 @@
 # limitations under the License.
 
 import sys,os
-from generator import *
 from common_codegen import *
+from generators.generator_utils import *
+from generators.base_generator import BaseGenerator
 
 #
-# SpirvValidationHelperOutputGenerator - Generate SPIR-V validation
-# for SPIR-V extensions and capabilities
-class SpirvValidationHelperOutputGenerator(OutputGenerator):
+# Generate SPIR-V validation for SPIR-V extensions and capabilities
+class SpirvValidationHelperOutputGenerator(BaseGenerator):
     def __init__(self,
                  errFile = sys.stderr,
                  warnFile = sys.stderr,
                  diagFile = sys.stdout):
-        OutputGenerator.__init__(self, errFile, warnFile, diagFile)
-        self.extensions = dict()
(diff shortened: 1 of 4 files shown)


scripts/generators/generator_utils.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/python3 -i
```
spencer-lunarg, line 1: copied this from https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6026 for helping with merge conflicts (since I needed `fileIsGeneratedWarning`)

ncesario-lunarg (approved):
LGTM!

scripts/generators/spirv_validation_generator.py
```diff
@@ -324,43 +150,145 @@ def createMapValue(self, name, enable, isExtension):
...
+    FeaturePointer feature;
+    ExtEnabled DeviceExtensions::*extension;
+    const char* property; // For human readability and make some capabilities unique
+};
+''')
+
+        #
+        # Build the struct with all the requirments for the spirv capabilities
+        out = []
+        out.append('// clang-format off\n')
```
juan-lunarg, line 225: Shouldn't we just ignore the whole file?

Can you also the ignore checks for clang-tidy?

juan-lunarg (approved):
Looks good overall to me. Just minor nitpick about clang-tidy / clang-format.

scripts/generators/spirv_validation_generator.py
spencer-lunarg: will add the clang-tidy comments

as for clang-format... I guess I always thought that we should be running clang-format on the generated code to make sure changes to the generator script didn't produce ugly looking stuff by accident

I guess for these, I had the `clang-format off` so it would be 1-per-line ... like that was a deliberate choice.... thinking out loud, not sure what the right answer really is

Issue #6020, opened 2023-06-16
`NegativeDynamicState.RasterizationLine` sporadic x86 Win32 failure 


juan-lunarg:
**Describe the Issue**

This test randomly fails during 32 bit testing:

![image](https://github.com/KhronosGroup/Vulkan-ValidationLayers/assets/114601453/e5140425-4a05-409c-9760-07d6a7b2097c)

spencer-lunarg:
ok... I see what is happening, the first VU is `VUID-vkCmdDraw-None-07637`

> VUID-vkCmdDrawIndexed-None-07637
If the bound graphics pipeline state was created with the VK_DYNAMIC_STATE_LINE_RASTERIZATION_MODE_EXT dynamic state enabled then [vkCmdSetLineRasterizationModeEXT](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#vkCmdSetLineRasterizationModeEXT) must have been called in the current command buffer prior to this drawing command

We have a 

```
m_errorMonitor->SetUnexpectedError("VUID-vkCmdDraw-stippledLineEnable-07498");  // default values undefined
```

in the test, I guess we should have `VUID-vkCmdDrawIndexed-stippledLineEnable-07496` check as well

... also this can be 4 tests, not 1 mega test

spencer-lunarg:
If this fails again, please ping me, will look into it more

juan-lunarg:
Still sporadically failing on 32 bit Windows build:

![image](https://github.com/KhronosGroup/Vulkan-ValidationLayers/assets/114601453/b4608174-cf3e-4d2b-9587-5d95f003c34b)

spencer-lunarg:
ok, but at least I know which of the previous subtests is failing... will try again

PR #6036, opened 2023-06-20
tests: Fix RasterizationLineModeDefault

spencer-lunarg:
**actually** fixes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6020

Realize the test was trying to test 2 things and overloaded

`VUID-vkCmdDraw-None-07637` is already being tested in another test

wrote a separate test for `VUID-vkCmdDraw-None-07638`


juan-lunarg (approved):

juan-lunarg (approved):
On passing CI. You messaged me locally that you know what the fix is.

PR #6039, opened 2023-06-20
scripts: Move CommandValidation to VulkanObject

spencer-lunarg:
This makes use of the VulkanObject to make the command_validation code gen much nicer... comments on code with details of use cases

The generated code was mainly just re-ordered with new `VulkanObject`


scripts/generators/base_generator.py
```diff
@@ -50,18 +51,33 @@ def intIfGet(elem, name):
 def boolGet(elem, name):
     return elem.get(name) is None or elem.get(name) != "true"
 
+#
+# Walk the JSON-derived dict and find all "vuid" key values
+def ExtractVUIDs(vuid_dict):
```
spencer-lunarg, line 56: this, `self.valid_usage_path`. and `self.valid_vuids` are used by few other generators, so will be nice to have here so others can reuse

scripts/generators/vulkan_object.py
```diff
@@ -54,6 +55,22 @@ class CommandParam:
     optional: bool
     noAutoValidity: bool
 
+class Queues(IntFlag):
```
spencer-lunarg, line 58: Made use of Flags/Enum in python as everyone else was just going to do the same string parsing anyway

scripts/generators/command_validation_generator.py
```diff
@@ -14,343 +14,229 @@
...
 // Used as key for maps of all vkCmd* calls
 // Does not include vkBeginCommandBuffer/vkEndCommandBuffer
 typedef enum CMD_TYPE {
-    CMD_NONE = 0,\n'''
-
+    CMD_NONE = 0,\n''')
         counter = 1
-        for name, cmdinfo in sorted(self.commands.items()):
-            output += '    CMD_' + name[5:].upper() + ' = ' + str(counter) + ',\n'
+        for name in filter(lambda x: x.startswith('vkCmd'), self.vk.commands.keys()):
```
spencer-lunarg, line 76: Instead of having a single pass to filter all the commands and have yet-another-container, decided to just have all the loops do a `filter`

scripts/generators/vulkan_object.py
```diff
@@ -68,16 +85,16 @@ class Command:
...
-    queues: List[str]         # ex. [ 'graphics', 'compute' ]
+    queues: Queues            # zero == No Queues found
     successCodes: List[str]   # ex. [ 'VK_SUCCESS', 'VK_INCOMPLETE' ]
     errorCodes: List[str]     # ex. [ 'VK_ERROR_OUT_OF_HOST_MEMORY' ]
 
     # The command can be inside either a primary or secondary command buffer
-    cmdBufferPrimary: bool
-    cmdBufferSecondary: bool
+    primary: bool
+    secondary: bool
```
charles-lunarg, line 94: Just to be clear, does that imply exclusiveness to the two bools? IE can a command be inside both primary and secondary? (The comment kinda seems like its exclusive-or rather than just logical-or.)

charles-lunarg (approved):
Looks good, only one nit about a comment.

juan-lunarg (approved):
Approving in place of @charles-lunarg

scripts/generators/vulkan_object.py
spencer-lunarg: sorry, good catch, it can be both

PR #6042, opened 2023-06-20
ci: Test with robin hood disabled

juan-lunarg:
The chromium folks and some package managers disable robin hood.

### DIFF
--- a/.github/workflows/vvl.yml
+++ b/.github/workflows/vvl.yml
@@ -42,22 +42,34 @@ jobs:
         flags: [{c: "-fsanitize=address", ld: "-fsanitize=address"}, {c: "-fsanitize=thread", ld: "-fsanitize=thread"}]
         config: [debug, release]
         cpp_std: [17]
+        robin_hood: ["ON"]
         include:
           # Test Ubuntu-20.04 release build works.
           - os: ubuntu-20.04
             compiler: {cc: gcc, cxx: g++}
             config: release
             cpp_std: 17
+            robin_hood: "ON"
           # Test C++ 20 support
           - os: ubuntu-22.04
             compiler: {cc: gcc, cxx: g++}
             config: release
             cpp_std: 20
+            robin_hood: "ON"
           # Test clang support
           - os: ubuntu-22.04
             compiler: {cc: clang, cxx: clang++}
             config: release
             cpp_std: 17
+            robin_hood: "ON"
+          # Test with Robin Hood disabled
+          # Chromium build, and some package managers don't use it.
+          - os: ubuntu-22.04
+            compiler: {cc: clang, cxx: clang++}
+            config: release
+            cpp_std: 17
+            robin_hood: "OFF"
+            flags: [{c: "-fsanitize=address", ld: "-fsanitize=address"}]
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


juan-lunarg:
I talked with @lunarpapillo and we are planning to remove the robin hood on/off from the Jenkins matrix. Instead handling testing the build configuration here in Github Actions.

.github/workflows/vvl.yml
```diff
@@ -42,22 +42,34 @@ jobs:
...
             cpp_std: 17
+            robin_hood: "ON"
+          # Test with Robin Hood disabled
+          # Chromium build, and some package managers don't use it.
+          - os: ubuntu-22.04
+            compiler: {cc: clang, cxx: clang++}
+            config: release
+            cpp_std: 17
+            robin_hood: "OFF"
+            flags: [{c: "-fsanitize=address", ld: "-fsanitize=address"}]
```
juan-lunarg, line 72: Note: Testing it here also offers us the opportunity for address sanitizer to potentially catch something.

.github/workflows/vvl.yml
```diff
@@ -42,22 +42,34 @@ jobs:
         flags: [{c: "-fsanitize=address", ld: "-fsanitize=address"}, {c: "-fsanitize=thread", ld: "-fsanitize=thread"}]
         config: [debug, release]
         cpp_std: [17]
+        robin_hood: ["ON"]
```
lunarpapillo, line 45: I'm a little confused by the presence of `robin_hood` as a matrix axis with the only defined value being `"ON"`, and its use below in one configuration with value `"OFF"`.  Shouldn't the axis line include both valid values?

I may just not be familiar enough with GHA.  Feel free to ignore if I"m completely off base.

.github/workflows/vvl.yml
lunarpapillo, line 72: `flags` also appears in the `matrix`, with its list of valid values.  For the other axis values, the matrix definition is an array, and the corresponding configuration value is an element of that array.  Here `flags` is still an array value.

Again, this may be perfectly correct, but it looks weird to me, so I'm pointing it out and letting you make the decision.

lunarpapillo (commented):
I'm providing commentary, but I don't know enough to do a real review.

.github/workflows/vvl.yml
juan-lunarg: It's intentional. I agree though it's not pretty.

.github/workflows/vvl.yml
juan-lunarg: > Shouldn't the axis line include both valid values?

This is valid and intentional.

Here is the job with robin hood disabled:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/actions/runs/5327006154/jobs/9649844774?pr=6042

.github/workflows/vvl.yml
juan-lunarg: Note, since this is primarily for the chromium folks. I'm also intentionally using clang to try and have parity.

.github/workflows/vvl.yml
```diff
@@ -70,14 +82,14 @@ jobs:
       - name: Setup ccache
         uses: hendrikmuhs/ccache-action@v1.2
         with:
-          key: ${{ matrix.os }}-${{ matrix.config }}-${{ matrix.compiler.cc }}-${{ matrix.compiler.cxx }}-${{ matrix.cpp_std }}-${{ matrix.flags.c }}-${{ matrix.flags.ld }}
+          key: ${{ matrix.os }}-${{ matrix.config }}-${{ matrix.compiler.cc }}-${{ matrix.compiler.cxx }}-${{ matrix.cpp_std }}-${{ matrix.flags.c }}-${{ matrix.flags.ld }}-${{matrix.robin_hood}}
```
juan-lunarg, line 85: I'm very open if anyone knows a better way to handle this. Since this isn't very robust.

Ideally there was a matrix hash / ID we could use instead.

juan-lunarg:
CI passes 🎉 

More context for this PR. Currently Jenkins tests this build permutation to ensure we don't break the chromium folks who don't use robin hood hashing. There are also several package managers that don't use robin hood hashing as well.

I want to move that build permutation testing to github actions instead.
- Since there is little benefit to testing this permutation in general
- We run with address sanitizer on Github actions
- I'm planning on removing robin hood and updating it with it's successor project (robin hood isn't maintained anymore)
- Reduces complexity of Jenkins jobs and makes sure we always test with robin hood enabled. Which should help perf slightly in Jenkins.

spencer-lunarg (approved):

PR #6043, opened 2023-06-21
android: Use debug.vvl. for settings

ncesario-lunarg:
Android API 26 allows property keys longer than 32 characters. Since API level 26 is the minimum version required for VVL, allow system properties to be used for modifying any setting that is controlled via an environment variable on desktop.

### DIFF
--- a/layers/vk_layer_config.cpp
+++ b/layers/vk_layer_config.cpp
@@ -88,7 +88,16 @@ std::string GetEnvironment(const char *variable) {
     delete[] buffer;
     return output;
 #elif defined(__ANDROID__)
-    const prop_info *prop_info = __system_property_find(variable);
+    std::string var = variable;
+
+    if (std::string_view{variable} != kForceDefaultCallbackKey) {
+        // kForceDefaultCallbackKey is a special key that needs to be recognized for backwards compatibilty.
+        // For all other strings, prefix the requested variable with "debug.vvl." so that desktop environment settings can be used
+        // on Android.
+        var = "debug.vvl." + var;
+    }
+
+    const prop_info *prop_info = __system_property_find(var.data());
 
     if (prop_info) {
         std::string property;


juan-lunarg (commented):
LGTM

You mentioned that @lunarpapillo can use this new functionality in VulkanTests. Is there code and/or a PR we could look at that takes advantage of this new functionality?

ncesario-lunarg:
> You mentioned that @lunarpapillo can use this new functionality in VulkanTests. Is there code and/or a PR we could look at that takes advantage of this new functionality?

Yes, this would--in theory--allow the "driver update process" (or whatever the official name is, for which Bob will correct me ;)) to work on Android in a similar fashion as it does on desktop. I can test this locally, but was having some difficulties with the Android build. Onece I've tested this out I'll mark this PR "ready for review."

lunarpapillo:
> Yes, this would--in theory--allow the "driver update process" (or whatever the official name is, for which Bob will correct me ;)) to work on Android in a similar fashion as it does on desktop.

Yes, this is helpful for the VVL **Blacklist exercise**, which we do whenever we deploy a new GPU or Android device, or update a driver or Android OS.  The blacklist exercise queries the list of available tests, then executes them one at a time, looking for failures and crashes.

 On desktop, when doing a new exercise for an established GPU, we disable the VVL internal blacklist implemented by `IsPlatform()` and `IsDriver()` by setting `VK_LAYER_TESTS_IGNORE_DISABLE_CHECKS`.  This ensures that we don't skip those tests, so we can verify whether the blacklist still needs to contain any tests that used to be skipped on the older driver.

There's been no equivalent functionality on Android.  So whenever we've done an Android blacklist exercise after an OS upgrade, the tests that were skipped due to the VVL internal blacklist on the older device are still skipped, so we get no information on whether any of those tests no longer need to be skipped.

Nate's changes will let us get complete and correct blacklist information on Android updates, like we can currently do for desktop GPU driver upates.

The process document, if you're curious, is currently under review, but will be checked in this week in all likelihood: https://github.com/LunarG/VulkanTests/blob/bob-doc-driver-upgrade/doc/administrator/new-device-new-driver.md

juan-lunarg:
> The process document, if you're curious, is currently under review, but will be checked in this week in all likelihood: https://github.com/LunarG/VulkanTests/blob/bob-doc-driver-upgrade/doc/administrator/new-device-new-driver.md

Sounds good to me. I'll approve on based on @lunarpapillo's approval.

juan-lunarg (approved):

Issue #5994, opened 2023-06-12
Spurious synchronization validation errors on Linux


martin-ejdestig:
**Environment:**
 - OS: Arch/Linux
 - GPU: Able to reproduce on Nvidia (proprietary driver) and Intel GPU (Mesa).
 - SDK or header version if building from repo: Using system provided package: https://archlinux.org/packages/extra/x86_64/vulkan-validation-layers/
 - Options enabled (synchronization, best practices, etc.): VK_VALIDATION_FEATURE_ENABLE_SYNCHRONIZATION_VALIDATION_EXT

**Describe the Issue**

Started noticing spurious synchronization validation errors when Arch upgraded from 1.3.231 to 1.3.236.0. Upgraded to 1.3.250.0 today and problem still persists.

See https://gitlab.archlinux.org/archlinux/packaging/packages/vulkan-validation-layers/-/blob/main/PKGBUILD for how Arch builds the validation layer.

Do not see same problem on Windows with Vulkan SDK on AMD GPU. (Regardless of SDK version.)

**Expected behavior**

No spurious synchronization validation errors on Linux.

**Additional context**

To verify in available code, that should be easy for you to build, I enabled validation in https://github.com/ocornut/imgui/blob/master/examples/example_sdl2_vulkan/main.cpp by applying the following changes:

```diff
diff --git a/examples/example_sdl2_vulkan/main.cpp b/examples/example_sdl2_vulkan/main.cpp
index 03a952c0..ea83b162 100644
--- a/examples/example_sdl2_vulkan/main.cpp
+++ b/examples/example_sdl2_vulkan/main.cpp
@@ -20,9 +20,9 @@
 //#include <vulkan/vulkan_beta.h>
 
 //#define IMGUI_UNLIMITED_FRAME_RATE
-#ifdef _DEBUG
+//#ifdef _DEBUG
 #define IMGUI_VULKAN_DEBUG_REPORT
-#endif
+//#endif
 
 // Data
 static VkAllocationCallbacks*   g_Allocator = nullptr;
@@ -128,6 +128,12 @@ static void SetupVulkan(ImVector<const char*> instance_extensions)
         create_info.enabledLayerCount = 1;
         create_info.ppEnabledLayerNames = layers;
         instance_extensions.push_back("VK_EXT_debug_report");
+
[...]

Issue #5318, opened 2023-02-21
Strange synchronization validation error when uploading initial data to an image


Pilzschaf:
**Environment:**
 - OS: Manjaro
 - GPU: AMD RX 7900XT
 - Options enabled: Validation, Best Practices, GPU assisted, GPU reserved binding slot and synchronization validation

**Describe the Issue**

Synchronization validation reports a write after write hazard for my initial data upload to an image. I simply create a new image, and a new scratch buffer. Fill the scratch buffer with the image data and try to copy the buffer to the image with vkCmdCopyBufferToImage. I have an image barrier before the copy to ensure the image is in the correct layout.

**Expected behavior**

I would not expect a sync validation error for the image as the layout is transitioned and there is no prior write to the image

**Valid Usage ID**
Vulkan Error: Validation Error: [ SYNC-HAZARD-WRITE-AFTER-WRITE ] Object 0: handle = 0xe7f79a0000000005, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x5c0ec5d6 | vkCmdCopyBufferToImage: Hazard WRITE_AFTER_WRITE for dstImage VkImage 0xe7f79a0000000005[], region 0. Access info (usage: SYNC_COPY_TRANSFER_WRITE, prior_usage: SYNC_IMAGE_LAYOUT_TRANSITION, write_barriers: 0, command: vkCmdPipelineBarrier2, seq_no: 1, reset_no: 2).
Thread 0, Frame 0:

**Additional context**

I have included two files with the API dump of my repro case. One contains all calls. The other is reduced to the parts which I think are relevant

<details>

This is the code I use for the copy of the buffer to the image
```c
VkImageMemoryBarrier2 imageBarrier = {VK_STRUCTURE_TYPE_IMAGE_MEMORY_BARRIER_2};
imageBarrier.srcStageMask = VK_PIPELINE_STAGE_TRANSFER_BIT;
imageBarrier.srcAccessMask = VK_ACCESS_TRANSFER_WRITE_BIT;
imageBarrier.dstStageMask = VK_PIPELINE_STAGE_TRANSFER_BIT;
imageBarrier.dstAccessMask = VK_ACCESS_TRANSFER_WRITE_BIT;
imageBarrier.oldLayout = VK_IMAGE_LAYOUT_UNDEFINED;
imageBarrier.newLayout = VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL;
[...]

PR #6049, opened 2023-06-21
layers: Fix syncval undefined initialization order

artem-lunarg:
This PR fixes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5994. Some linux distributions provide VVL packages that report false-positive syncval errors and this does not happen when VVL is built according to official build configuration.

The issue is reported on Arch Linux but the problem is also reproducible on Ubuntu (and probably other distributions) both in debug and release configurations and also with different versions of GCC compiler (I tested on 11.xx and 13.1.1).

The issue is observable when using link time optimization `-flto`. Some linux packages are built by specifying it through `CXXFLAGS` environment variable that is consumed by cmake.

UPDATE: after thinking more about this it looks it's not a compiler problem but rather our code goes into the territory that initialization of globals across compilation units is undefined: we try to use global value from another compilation unit (by calling `syncStageAccessMaskByStageBit()` which accesses `static const SyncStageAccessFlags`) to initialize global (`kPresentValidAccesses`) in the current compilation unit. In this perspective the issue is not related to LTO and is undefined behavior in general. LTO exposes it and maybe it's luck that without LTO it works.

This undefined initialization of globals across units is more obvious now when the globals are identified but I spent the last 2 days on this issue and did not notice this. One more reason to be very careful with using global variables. It's easy to cross the boundary of compilation unit during initialization, especially when calling non-trivial functions that can access globals from another compilation units (or such functions can be modified in the future to read new global variables,  and suddenly this introduces undefined behavior).

The solution is still the same (do not initialize `kPresentValidAccesses` as part of global init sequence) but the initial reasoning (about compiler bug) was wrong. Thanks everyone for comments that helped to clarify this, rather than complaining about the compiler.

~~This PR re-orders code in a way that LTO initialization re-ordering does not affect us. So far it looks like a compiler bug where optimization affects correctness of the program. More specifically it violates the rule that the order of globals initialization within single translation unit is well defined. As always in such cases, it worth to mention it might be something wrong with out code too.~~

Was also reported in https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5318

### Details of the issue
UPDATE: removed this section since it was incorrect.

### DIFF
--- a/layers/sync/sync_validation.cpp
+++ b/layers/sync/sync_validation.cpp
@@ -8274,10 +8274,8 @@ class ApplySemaphoreBarrierAction {
 
 class ApplyAcquireNextSemaphoreAction {
   public:
-    static const SyncStageAccessFlags kPresentValidAccesses;
-    static const SyncExecScope kPresentSrcScope;
     ApplyAcquireNextSemaphoreAction(const SyncExecScope &wait_scope, ResourceUsageTag acquire_tag)
-        : barrier_(1, SyncBarrier(kPresentSrcScope, kPresentValidAccesses, wait_scope, SyncStageAccessFlags())),
+        : barrier_(1, SyncBarrier(getPresentSrcScope(), getPresentValidAccesses(), wait_scope, SyncStageAccessFlags())),
           acq_tag_(acquire_tag) {}
     void operator()(ResourceAccessState *access) const {
         // Note that the present operations may or may not be present, given that the fence wait may have cleared them out.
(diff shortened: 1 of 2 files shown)


juan-lunarg (commented):
Good job catching an LTO bug 👍🏾 , those are particularly nasty. We currently enable it on Windows but MSVC is a lot more forgiving.

artem-lunarg:
"Windows but MSVC is a lot more forgiving." In a way that it does not have this issue? ;)

artem-lunarg:
Yes, I assume in MSVC it's different implementation. Also it might be not an issue on clang, I tested only gcc.

jeremyg-lunarg:
Great description of the problem in this PR!

juan-lunarg:
> What happens with GCC's LTO is that it detects that we have two identical const variables and decides to use only one of them

That sounds like an ODR violation to me. I don't think this is a compiler bug.

There are methods to detect ODR violation we should employ in CI:
https://maskray.me/blog/2022-11-13-odr-violation-detection

artem-lunarg:
It should not be ODR violation because each compilation unit uses it’s own ‘static’ variable which us private to compilation unit (internal linkage). But I’ll check the link.

juan-lunarg:
From the GCC documentation:

`at link time, the current unit is the union of all objects compiled with LTO`

https://gcc.gnu.org/onlinedocs/gccint/WHOPR.html

artem-lunarg:
Okay, but it goes on addition to C++ spec, not replaces it. They could do any optimization until they can prove correctness with regard to spec.

charles-lunarg:
I have a hunch that the variable should be inline constexpr but am running my own build to confirm if I can 1. reproduce the issue and 2. fix it with inline.

artem-lunarg:
@charles-lunarg there probably could be different ways to fix this but it’s not like we have wrong c++ code here so it “should” be fixed due to the reason it’s incorrect. static global variables it’s pretty old stuff and boring stuff. New element here is LTO.

artem-lunarg:
> Great description of the problem in this PR!

Thanks, unfortunately it was wrong conclusions. Updated description.

layers/sync/sync_validation.cpp
```diff
@@ -8287,19 +8285,26 @@ class ApplyAcquireNextSemaphoreAction {
         }
     }
 
+  private:
+    const SyncExecScope &getPresentSrcScope() const {
```
spencer-lunarg, line 8289: A note (even just pointing to your PR summary) might be worth adding here in case someone tries touching this in the future, they are aware

spencer-lunarg (approved):
:100: :tada:

layers/sync/sync_validation.cpp
artem-lunarg: Added a comment. Also I agree with @juan-lunarg that if there is a tool that can help with this it's the best solution (that helps to detect cross-file initialization dependency). It's easy to make such mistakes. An example when you update an existing function by reading a new global variable (no initialization here) might introduce undefined behavior in some cases. That's very non-obvious.

jeremyg-lunarg (approved):

PR #6051, opened 2023-06-21
syncval: Restore byte accurate linear images

jzulauf-lunarg:
All images were being treated as opaque (non-linear) accidentally. Restored implementation derived parameterizations.

### DIFF
--- a/layers/containers/subresource_adapter.cpp
+++ b/layers/containers/subresource_adapter.cpp
@@ -282,6 +282,15 @@ ImageRangeEncoder::ImageRangeEncoder(const IMAGE_STATE& image, const AspectParam
     VkImageSubresourceLayers subres_layers = {limits_.aspectMask, 0, 0, limits_.arrayLayer};
     linear_image_ = false;
 
+    // WORKAROUND for profile and mock_icd not containing valid VkSubresourceLayout yet. Treat it as optimal image.
+    if (image.createInfo.tiling == VK_IMAGE_TILING_LINEAR) {
+        subres = {static_cast<VkImageAspectFlags>(AspectBit(0)), 0, 0};
+        DispatchGetImageSubresourceLayout(image.store_device_as_workaround, image.image(), &subres, &layout);
+        if (layout.size > 0) {
+            linear_image_ = true;
+        }
+    }
+
     is_compressed_ = FormatIsCompressed(image.createInfo.format);
     texel_extent_ = FormatTexelBlockExtent(image.createInfo.format);
 


artem-lunarg (approved):

layers/containers/subresource_adapter.cpp
```diff
@@ -280,7 +280,7 @@ ImageRangeEncoder::ImageRangeEncoder(const IMAGE_STATE& image, const AspectParam
     VkSubresourceLayout layout = {};
     VkImageSubresource subres = {};
     VkImageSubresourceLayers subres_layers = {limits_.aspectMask, 0, 0, limits_.arrayLayer};
-    linear_image_ = false;
+    linear_image_ = image.createInfo.tiling == VK_IMAGE_TILING_LINEAR;
```
juan-lunarg, line 283: Is there a test / app failing without this? Should we add a new test for sync_val?

jzulauf-lunarg: It's an internal representation choice that most would affect a "not completely supported" feature "aliasing of linear resources" --  probably needs a test, but with the pending aliasing rewrite, will change (and need expanded testing).

jzulauf-lunarg: @juan-lunarg -- what I'm more worried over is the *rest* of the encoder/generator test coverage.  On the non-linear path, we create our own encoding, and it doesn't get a wide range of formats covered in test usage.

spencer-lunarg (commented):
Can we please get a positive test for this

jzulauf-lunarg:
Mock ICD still doesn't support `GetImageSubresourceLayout`

layers/containers/subresource_adapter.cpp
```diff
@@ -282,6 +282,15 @@ ImageRangeEncoder::ImageRangeEncoder(const IMAGE_STATE& image, const AspectParam
     VkImageSubresourceLayers subres_layers = {limits_.aspectMask, 0, 0, limits_.arrayLayer};
     linear_image_ = false;
 
+    // WORKAROUND for profile and mock_icd not containing valid VkSubresourceLayout yet. Treat it as optimal image.
```
juan-lunarg, line 285: There isn't a developer dedicated to mock icd. Generally it's expected the user who needs the support should add it.

@charles-lunarg recently did some work on the Mock ICD to make it easier to work on for all developers.

layers/containers/subresource_adapter.cpp
```diff
@@ -282,6 +282,15 @@ ImageRangeEncoder::ImageRangeEncoder(const IMAGE_STATE& image, const AspectParam
...
+    if (image.createInfo.tiling == VK_IMAGE_TILING_LINEAR) {
+        subres = {static_cast<VkImageAspectFlags>(AspectBit(0)), 0, 0};
+        DispatchGetImageSubresourceLayout(image.store_device_as_workaround, image.image(), &subres, &layout);
```
artem-lunarg, line 288: How does this workaround work in general? Isn't zero aspect flags is invalid value for `vkGetImageSubresourceLayout` and raises `VUID-vkGetImageSubresourceLayout-aspectMask-00997` ?

jzulauf-lunarg: It works by seeing if the driver gives an empty result (indicating MockICD). As for the the aspectMask... I was just looking at that when I back ported it.  In testing on my NVidia the driver wasn't choking on it... but trusting a generous driver doesn't seem prudent.  It's been around quite a while, but if I'm going to reinstate it, it should be correct... so I'll check.

artem-lunarg: It's a good point about the driver, I did not think about that and more about VVL side. Does dispatch calls like `DispatchGetImageSubresourceLayout` skip layers chain? (@charles-lunarg for visibility)

jzulauf-lunarg: Doesn't bypass layers, but does bypass any other ValidationLayerObjects enabled.

artem-lunarg: Ah, now it's clear!

jzulauf-lunarg: To simplify aspect bit handling, the aspectMask stored as an array of the set bits (one bit at a time). `RangeEncoder::AspectBit(N)` returns the Nth valid bit (for N < number of bit set in aspectMask).

So AspectBit(0) returns the first valid aspect bit from the creation mask, and every image must have at least one.

So I think we're good, unless you think this needs some commentary to clarify for future maintainers.

artem-lunarg: I was not careful enough when read the code, somehow thought it's cast of zero to `VkImageAspectFlags`. If I check this code in IDE then it's usually less of a problem since I can quickly check definitions, but if I check github page it's more tricky (one of the reason for more complicated reviews sometimes I checkout the code). In the latter case intermediate local variable for the result of `AspectBit(0)` might help to avoid confusion.

PR #6055, opened 2023-06-22
scripts: Move enum_string_helper to own generator

spencer-lunarg:
Main goal: Get rid of `helper_file_generator.py` and split it's own generator (we were only sharing because of the issue solved with `VulkanObject`)


layers/core_checks/cc_descriptor.cpp
```diff
@@ -135,9 +135,8 @@ bool CoreChecks::VerifySetLayoutCompatibility(const DescriptorSetLayout &layout_
         }
         error_str << report_data->FormatHandle(layout_dsl_handle)
                   << " from pipeline layout does not have the same binding flags at binding " << i << " ( "
-                  << string_VkDescriptorBindingFlagsEXT(ds_layout_flags[i]) << " ) as "
-                  << report_data->FormatHandle(bound_dsl_handle) << " ( "
-                  << string_VkDescriptorBindingFlagsEXT(bound_layout_flags[i]) << " ), which is bound";
+                  << string_VkDescriptorBindingFlags(ds_layout_flags[i]) << " ) as " << report_data->FormatHandle(bound_dsl_handle)
```
spencer-lunarg, line 138: the old `vk_enum_string_helper.h` had duplicate functions for promoted extensions, this change caught a few places still using the old ones

scripts/generators/base_generator.py
```diff
@@ -276,10 +281,19 @@ def genGroup(self, groupinfo, name, alias):
                     continue
                 fieldName = elem.get('name')
                 negative = elem.get('dir') != None
-                extension = elem.get('extname')
-                fields.append(Enum(fieldName, negative, extension))
+                extensions = None if elem.get('extname') is None else [elem.get('extname')]
+                protect = elem.get('protect')
+
+                # Some values have multiple extensions (ex VK_DESCRIPTOR_UPDATE_TEMPLATE_TYPE_PUSH_DESCRIPTORS_KHR)
```
spencer-lunarg, line 287: This is a  little ugly, but decided that this is something that could be fixed in the `reg.py` scripts, but for now, rather keep all the "ugly" logic in `base_generator` where most people will never have to look now

scripts/generators/enum_string_helper_generator.py
```diff
@@ -0,0 +1,139 @@
...
+                out.append(f'        case {flag.name}:\n')
+                out.append(f'            return "{flag.name}";\n')
+                out.extend([f'#endif //{flag.protect}\n'] if flag.protect else [])
+            out.append('        default:\n')
+            out.append(f'            return "Unhandled {bitmask.name}";\n')
+            out.append('    }\n')
+            out.append('}\n')
+
+            mulitBitChecks = ''
+            for flag in filter(lambda x: x.multiBit, bitmask.flags):
```
spencer-lunarg, line 117: this actually caught a lot of places we were not printing out the multi bit flag (like `VK_CULL_MODE_FRONT_AND_BACK`) because we had a manual list :tada:

scripts/generators/vulkan_object.py
```diff
@@ -111,34 +111,41 @@ class Command:
...
     """<enums> of type enum"""
     name: str
     bitWidth: int # 32 or 64
-    fields: List[Enum]
+    protect: str  # ex. 'VK_ENABLE_BETA_EXTENSIONS'
+    fields: List[EnumField]
 
 @dataclass(frozen=True)
-class FlagBit:
+class Flag:
```
spencer-lunarg, line 131: After actually using these, I found a `Bitmask`/`Flag` and `Enum`/`EnumField` was easier to understand... open to other name suggestions as well

scripts/generators/base_generator.py
```diff
@@ -267,6 +267,11 @@ def genCmd(self, cmdinfo, name, alias):
     #
     # List the enum for the commands
     def genGroup(self, groupinfo, name, alias):
+        if alias is not None:
+            return
+        # There can be case where the Enum/Bitmask is in a protect, but the individual
+        # fields also have their own protect
+        groupProtect = self.currentFeature.protect if hasattr(self.currentFeature, 'protect') and self.currentFeature.protect is not None else None
```
artem-lunarg, line 274: probably can be simplified to this:
`groupProtect = self.currentFeature.protect if hasattr(self.currentFeature, 'protect') else None`

scripts/generators/base_generator.py
```diff
@@ -276,10 +281,19 @@ def genGroup(self, groupinfo, name, alias):
...
-                extension = elem.get('extname')
-                fields.append(Enum(fieldName, negative, extension))
+                extensions = None if elem.get('extname') is None else [elem.get('extname')]
+                protect = elem.get('protect')
+
+                # Some values have multiple extensions (ex VK_DESCRIPTOR_UPDATE_TEMPLATE_TYPE_PUSH_DESCRIPTORS_KHR)
+                # genGroup() lists them twice, so need to just remove, update, re-add if we find a duplicate
+                for field in fields:
+                    if field.name == fieldName:
+                        extensions.append(field.extensions)
```
artem-lunarg, line 291: extend?

scripts/generators/base_generator.py
spencer-lunarg: ...yes, thanks!

scripts/generators/base_generator.py
```diff
@@ -276,10 +281,19 @@ def genGroup(self, groupinfo, name, alias):
...
+                        fields.remove(field)
```
artem-lunarg, line 292: From what I read on the internet, removing from the list during iteration iis not safe. Here's one program from the same source:
```
lst = [1, 2, 3, 4, 5, 6, 7, 8]
for x in lst:
    if x < 6:
        lst.remove(x)
>>> lst
[2, 4, 6, 7, 8]
```

I also run this one and for some reason it works, so not sure if there are special cases when it works or luck:
```
lst = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
for x in lst:
    if x % 2 == 0:
        lst.remove(x)

>> lst
[1, 3, 5, 7, 9]
```

layers/vulkan/generated/vk_object_types.h
```diff
@@ -511,195 +436,211 @@ template <> struct VkHandleInfo<VkBuffer> {
...
-    static const VulkanObjectType kVulkanObjectType = kVulkanObjectTypeDeviceMemory;
-    static const VkDebugReportObjectTypeEXT kDebugReportObjectType = VK_DEBUG_REPORT_OBJECT_TYPE_DEVICE_MEMORY_EXT;
-    static const VkObjectType kVkObjectType = VK_OBJECT_TYPE_DEVICE_MEMORY;
+
+template <> struct VkHandleInfo<VkRenderPass> {
+    static const VulkanObjectType kVulkanObjectType = kVulkanObjectTypeRenderPass;
+    static const VkDebugReportObjectTypeEXT kDebugReportObjectType = VK_DEBUG_REPORT_OBJECT_TYPE_RENDER_PASS_EXT;
+    static const VkObjectType kVkObjectType = VK_OBJECT_TYPE_RENDER_PASS;
     static const char* Typename() {
-        return "VkDeviceMemory";
```
artem-lunarg, line 653: General question, why order changes (VkDeviceMemory -> VkRenderPass)? Is it specific for this run or it might happen each time?

scripts/generators/vulkan_object.py
artem-lunarg: +1 for Bitmask/Flag
Not sure about calling enum element a field (but maybe it is!), `EnumValue` just because I used it before, but not sure if it's better (EnumValue to indicate that enum's elements are numerical values in contrast to fields in structs that are more complex).

scripts/generators/enum_string_helper_generator.py
```diff
@@ -0,0 +1,139 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+****************************************************************************/\n'''
+        self.write(copyright)
+        self.write('// NOLINTBEGIN') # Wrap for clang-tidy to ignore
+
+        self.write('''
+#pragma once
+#ifdef _MSC_VER
+#pragma warning( disable : 4065 )
```
juan-lunarg, line 61: Do we still need to disable this warning?

layers/vulkan/generated/vk_object_types.h
artem-lunarg: thanks, that's great

juan-lunarg (approved):

scripts/generators/enum_string_helper_generator.py
spencer-lunarg: no idea, just matching code... happy to make a PR after to remove it

PR #6057, opened 2023-06-22
cmake: Fix build_platform usage when dealing with Android

juan-lunarg:
Currently update_deps.py breaks if a dependency doesn't work on Android like the Vulkan Loader.

This is because we detect build platforms based on the host platform.

Current fix is to check for Android toolchain explicitly.

### DIFF
--- a/scripts/update_deps.py
+++ b/scripts/update_deps.py
@@ -220,6 +220,7 @@ Legal options include:
 "windows"
 "linux"
 "darwin"
+"android"
 
 Builds on all platforms by default.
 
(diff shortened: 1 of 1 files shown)


scripts/update_deps.py
juan-lunarg, line 1: This issue only arrises when dealing with a dependency like the Vulkan-Loader.

```
        {
            "name": "Vulkan-Loader",
            "url": "https://github.com/KhronosGroup/Vulkan-Loader.git",
            "sub_dir": "Vulkan-Loader",
            "build_dir": "Vulkan-Loader/build",
            "install_dir": "Vulkan-Loader/build/install",
            "commit": "v1.3.254",
            "deps": [
                {
                    "var_name": "VULKAN_HEADERS_INSTALL_DIR",
                    "repo_name": "Vulkan-Headers"
                }
            ],
            "build_platforms": [
                "linux",
                "darwin"
            ],
            "optional": [
                "tests"
            ]
        }
```

This is an example taken from Vulkan-Profiles. Which is what revealed the issue to me. As you can see the build_platforms are specified to avoid causing a build failure on Android.

spencer-lunarg (approved):

PR #6058, opened 2023-06-22
Zulauf new split merge

jzulauf-lunarg:
Created a non recurrent update and infill .  Tests out identical with the one that has been in syncval for a while, though it's from scratch.


jzulauf-lunarg:
All needed code cherry picked elsewhere

PR #6060, opened 2023-06-23
layers: Remove Warning from Enum String helper

spencer-lunarg:
Try and remove the `#pragma warning( disable : 4065 )` warning from `vk_enum_string_helper.h` as not sure if needed anymore

### DIFF
--- a/scripts/generators/enum_string_helper_generator.py
+++ b/scripts/generators/enum_string_helper_generator.py
@@ -57,10 +57,6 @@ class EnumStringHelperOutputGenerator(BaseGenerator):
 
         self.write('''
 #pragma once
-#ifdef _MSC_VER
-#pragma warning( disable : 4065 )
-#endif
-
 #include <string>
 #include <vulkan/vulkan.h>''')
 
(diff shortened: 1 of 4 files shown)


spencer-lunarg:
seems things like `VkAcquireProfilingLockFlagBitsKHR` are empty (because they were added for the future)

Will update the scripts to just not create a `string_` for these (and still try and remove the warning)

scripts/generators/enum_string_helper_generator.py
```diff
@@ -79,7 +75,8 @@ def generate(self):
         self.write("".join(out))
 
         out = []
-        for enum in self.vk.enums.values():
+        # If there are no fields (empty enum) ignore
+        for enum in filter(lambda x: len(x.fields) > 0, self.vk.enums.values()):
```
arno-lunarg, line 79: Could this use the syntax Bob mentioned?
```python
for enum in [x in self.vk.enums.values() if len(x.fields) > 0]:
```

arno-lunarg (approved):
LGTM, 1 suggestion

scripts/generators/enum_string_helper_generator.py
spencer-lunarg: yes, I need to go back through and replace all the `filter` I added

spencer-lunarg:
seems these warnings were hidding issues in `spirv_validation_helper` too... :cry:

juan-lunarg (approved):

PR #6062, opened 2023-06-23
scripts: Move SafeStruct to VulkanObject

spencer-lunarg:
Started to move SafeStruct out, plan to do the Utils and Source as well (either in this PR or follow up)

### DIFF
--- a/layers/vulkan/generated/vk_safe_struct.h
+++ b/layers/vulkan/generated/vk_safe_struct.h
@@ -1,27 +1,27 @@
 // *** THIS FILE IS GENERATED - DO NOT EDIT ***
-// See helper_file_generator.py for modifications
-
+// See safe_struct_generator.py for modifications
 
 /***************************************************************************
- *
- * Copyright (c) 2015-2023 The Khronos Group Inc.
- * Copyright (c) 2015-2023 Valve Corporation
- * Copyright (c) 2015-2023 LunarG, Inc.
- * Copyright (c) 2015-2023 Google Inc.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


layers/vulkan/generated/vk_safe_struct.h
```diff
@@ -194,17 +195,17 @@ struct safe_VkDeviceCreateInfo {
...
     uint32_t waitSemaphoreCount;
-    VkSemaphore* pWaitSemaphores{};
+    VkSemaphore* pWaitSemaphores;
     const VkPipelineStageFlags* pWaitDstStageMask{};
     uint32_t commandBufferCount;
-    VkCommandBuffer* pCommandBuffers{};
+    VkCommandBuffer* pCommandBuffers;
     uint32_t signalSemaphoreCount;
-    VkSemaphore* pSignalSemaphores{};
+    VkSemaphore* pSignalSemaphores;
```
juan-lunarg, line 207: Minor point I've noticed. Some pointers used to default initialize to nullptr. Now they don't. Most likely not an issue.

However, my preference would be that all variables are default initialized via `{}`

juan-lunarg: I realize that this shouldn't matter if the codegen works properly. But in the case it doesn't work properly it's nice to have a stable default like nullptr to default to.

spencer-lunarg: agree, will fix

spencer-lunarg: fixed, and was able to re-write the code cleaner now taking a second look at why this was happening

layers/vulkan/generated/vk_safe_struct.h
```diff
@@ -71,6 +71,7 @@ struct safe_VkImageMemoryBarrier {
     uint32_t dstQueueFamilyIndex;
     VkImage image;
     VkImageSubresourceRange subresourceRange;
```
juan-lunarg, line 73: FFTI I think all members of the struct should be initialized with {}. Not just the pointers.

juan-lunarg (approved):

layers/vulkan/generated/vk_safe_struct.h
spencer-lunarg: well the nice thing now, anyone should be easy to fix this now

PR #6067, opened 2023-06-23
Potential short-term fix for https://github.com/KhronosGroup/SPIRV-Tools/issues/5270

ncesario-lunarg:


tests/unit/gpu_av_positive.cpp
ncesario-lunarg, line 1: @jeremyg-lunarg This doesn't seem to reproduce the crash I see when running `spirv-opt` on the command line manually. I can see that the optimizer is running, but it's not crashing with or without my changes to `gpu_validation.cpp`.

ncesario-lunarg:
This was resolved with https://github.com/KhronosGroup/SPIRV-Tools/pull/5316.

PR #6069, opened 2023-06-25
layers: Change message filter container to unordered_set

### LINKED ISSUE #5902: Message filtering inaccuracy and inefficiency
I just noticed that `filter_message_ids`, which is used to filter VUIDs is a vector and it stores hashes of VUID strings.

First, this will not always work correctly, as if two VUIDs have the same hash then we might get unexpected filtering.

Second, if the filter list is large, then looking up for a matching VUID hash may be slow due to the linear lookup.

Instead, I suggest `filter_message_ids` to be converted to a `std::unordered_set` and use `vvl_vuid_hash` as its hash function. This should solve both the performance and correctness concerns.

aqnuep:
Fixes #5902.

### DIFF
--- a/layers/layer_options.h
+++ b/layers/layer_options.h
@@ -27,7 +27,7 @@ typedef struct {
     const void *pnext_chain;
     CHECK_ENABLED &enables;
     CHECK_DISABLED &disables;
-    std::vector<uint32_t> &message_filter_list;
+    std::unordered_set<uint32_t> &message_filter_list;
     int32_t *duplicate_message_limit;
     bool *fine_grained_locking;
 } ConfigAndEnvSettings;
--- a/layers/error_message/logging.cpp
+++ b/layers/error_message/logging.cpp
@@ -324,8 +324,7 @@ static bool LogMsgEnabled(const debug_report_data *debug_data, std::string_view 
     }
     // If message is in filter list, bail out very early
     const uint32_t message_id = vvl_vuid_hash(vuid_text);
-    if (std::find(debug_data->filter_message_ids.begin(), debug_data->filter_message_ids.end(), message_id)
-        != debug_data->filter_message_ids.end()) {
+    if (debug_data->filter_message_ids.find(message_id) != debug_data->filter_message_ids.end()) {
         return false;
     }
     if ((debug_data->duplicate_message_limit > 0) && UpdateLogMsgCounts(debug_data, static_cast<int32_t>(message_id))) {
--- a/layers/error_message/logging.h
+++ b/layers/error_message/logging.h
@@ -24,6 +24,7 @@
 #include <string>
 #include <string_view>
 #include <vector>
+#include <unordered_set>
 
 #include "vk_layer_config.h"
 #include "containers/custom_containers.h"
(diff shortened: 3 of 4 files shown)


layers/layer_options.h
```diff
@@ -27,7 +27,7 @@ typedef struct {
     const void *pnext_chain;
     CHECK_ENABLED &enables;
     CHECK_DISABLED &disables;
-    std::vector<uint32_t> &message_filter_list;
+    std::unordered_set<uint32_t> &message_filter_list;
```
juan-lunarg, line 30: Did you try `vvl::unordered_set`? It uses the robin_hood hashing container instead of the STL one.

juan-lunarg (commented):
Out of curiosity did you notice a measurable performance impact in any scenarios?

layers/layer_options.h
jeremyg-lunarg: Yes if std:: is needed here we need to have a documented reason why

aqnuep: Measured saving is minimal, to be fair, simply the cost of the robin_hood hashing, as STL uses identity hash for uint32_t. But it just made sense to avoid even that small cost, when it's unnecessary, especially as the keys are already hashed.

The reason is documented in `logging.h`. This `message_filter_list` variable will only ever be a reference to `filter_message_ids` in `logging.h`, hence I didn't duplicate the comment here. But if you'd be happier if I'd add a comment here too, that's fine.

I'm also happy to change this to `vvl::unordered_set`, if that's preferred, but as the comment in `logging.h` explains, that would just unnecessarily re-hash an already perfect hash, as `vvl_vuid_hash` is tested to be perfect for the set of valid usage messages coming from the spec at least.

That's another question whether testing that `vvl_vuid_hash` is collision free is enough, considering that there are valid usage messages (internal messages, warnings, best practices, etc.) that don't come from the spec, so the current hash based solution may not be entirely correct. However, as lots of mechanisms depend on the use of `vvl_vuid_hash` (including the ability to filter based on decimal and hexadecimal hash values), I did not want to disturb that.

jeremyg-lunarg: The comment in logging.h is good enough for me for the std:: vs vvl:: set.   I'll defer to @juan-lunarg about hash collisions, he's pondered that problem quite a bit in the past.

juan-lunarg (approved):

layers/layer_options.h
juan-lunarg: > That's another question whether testing that vvl_vuid_hash is collision free is enough, considering that there are valid usage messages (internal messages, warnings, best practices, etc.) that don't come from the spec, so the current hash based solution may not be entirely correct. However, as lots of mechanisms depend on the use of vvl_vuid_hash (including the ability to filter based on decimal and hexadecimal hash values), I did not want to disturb that.

Yeah that's basically the same result I came to when I tried to look at revisiting the hashing we were using.

jeremyg-lunarg (approved):

layers/layer_options.h
mark-lunarg: At implementation time (albeit with fewer VUIDs) there were zero collisions.

aqnuep: There are still no collisions for spec defined VUIDs, thanks to the dedicated test cases @juan-lunarg added a while ago.

The only possible source of problem is the set of VUIDs internally defined by the validation layer implementation, let those be best practices VUs, or anything else that doesn't directly come from the spec and thus is not present in the `vuid_spec_text` generated into `vk_validation_error_messages.h`, because @juan-lunarg's test cases could only check those for collisions without enumerating all of those internal VUIDs in the tests.

Maybe something could be done about that too by including all these custom VUID names into the `vuid_spec_text` array, but I didn't look into that.

Issue #6056, opened 2023-06-22
Assert/Crash related to vkGetBufferDeviceAddress.


dorian-apanel-intel:
**Environment:**
 - OS: Windows
 - GPU: Intel A770 (NVidia too)
 - SDK or header version if building from repo: 886d29ee27e1aaf359ddc55e9b97353a97563842
 - Options enabled (synchronization, best practices, etc.): api_dump (closer to app), validation (closer to driver):  Standard validation preset from vkconfig.

**Describe the Issue**

Sporadic assert in DestroyBuffer if vkGetBufferDeviceAddress is used. Sometimes it crashes later in some following vkGetBufferDeviceAddress.
Maybe it could be reproduced on some repeatable unit test, but I see it in game.

**Additional context**

Assert: `assert(buffer_found_it != buffers.end());`

<details>

<summary>Callstack:</summary>

```
 	VkLayer_khronos_validation.dll!`ValidationStateTracker::PreCallRecordDestroyBuffer'::`15'::<lambda_1>::operator()<...>(small_vector<...> &) Line 480	C++	Symbols loaded.
 	VkLayer_khronos_validation.dll!sparse_container::range_map<...>::impl_erase_range<...>(const sparse_container::range<...> &, std::_Tree_iterator<...>, const ValidationStateTracker::PreCallRecordDestroyBuffer::__l15::<lambda_1> &) Line 432	C++	Symbols loaded.
 	VkLayer_khronos_validation.dll!sparse_container::range_map<...>::erase_range_or_touch<...>(const sparse_container::range<...> &, const ValidationStateTracker::PreCallRecordDestroyBuffer::__l15::<lambda_1> &) Line 586	C++	Symbols loaded.
 	VkLayer_khronos_validation.dll!ValidationStateTracker::PreCallRecordDestroyBuffer(VkDevice_T *, VkBuffer_T *, const VkAllocationCallbacks *) Line 477	C++	Symbols loaded.
>	VkLayer_khronos_validation.dll!vulkan_layer_chassis::DestroyBuffer(VkDevice_T *, VkBuffer_T *, const VkAllocationCallbacks *) Line 1911	C++	Symbols loaded.
 	VkLayer_api_dump.dll!00007ff89f9acb73()	Unknown	Non-user code. Cannot find or open the PDB file.
```

</details>

jeremyg-lunarg:
@arno-lunarg, do you think this is related to the problem you're already working on?

[...]

PR #6072, opened 2023-06-26
layers: Fix `vkGetBufferDeviceAddress` state recording bug

arno-lunarg:
- Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6056
Also should fix Dan G issue

Adding a buffer to `buffer_address_map_` should now work as expected (in some cases adding a buffer could fail), and be faster.

### DIFF
--- a/layers/state_tracker/state_tracker.h
+++ b/layers/state_tracker/state_tracker.h
@@ -1607,6 +1607,9 @@ class ValidationStateTracker : public ValidationObject {
 
     mutable VideoProfileDesc::Cache video_profile_cache_;
 
+    using BufferAddressMapStore = small_vector<std::shared_ptr<BUFFER_STATE>, 1, size_t>;
+    using BufferAddressRangeMap = sparse_container::range_map<VkDeviceAddress, BufferAddressMapStore>;
+
   protected:
     // tracks which queue family index were used when creating the device for quick lookup
     vvl::unordered_set<uint32_t> queue_family_index_set;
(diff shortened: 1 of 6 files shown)


tests/containers/small_vector.cpp
```diff
@@ -266,4 +267,3378 @@ TEST(CustomContainer, SmallVectorNotDefaultInsertableDefaultValue) {
         std::vector<NoDefaultCons> ref = {};
         ASSERT_TRUE(HaveSameElements(v9, ref));
     }
-}
\ No newline at end of file
+}
+
+// Test taken from https://github.com/dorian-apanel-intel/Vulkan-ValidationLayers/commit/b3b96320cd75f2ea4f72d6e23893d6f1b10c948b
+TEST(CustomContainer, RangeTest) {
```
jeremyg-lunarg, line 273: Do we really want this test here or somewhere under tests/?

layers/state_tracker/state_tracker.h
```diff
@@ -1607,6 +1607,9 @@ class ValidationStateTracker : public ValidationObject {
 
     mutable VideoProfileDesc::Cache video_profile_cache_;
 
+    using BufferAddressMapStore = small_vector<std::shared_ptr<BUFFER_STATE>, 1, size_t>;
```
jeremyg-lunarg, line 1610: Did you ever try using raw BUFFER_STATE pointers here? No perf difference?

tests/containers/small_vector.cpp
arno-lunarg: I don't know, do we have a better spot for this kind of thing?

juan-lunarg: I think this test should be under `tests/containers/` but not in the small_vector.cpp class.

arno-lunarg: Test has been removed: it will not show any regression

layers/state_tracker/state_tracker.h
arno-lunarg: Done with latest commit

PR #6073, opened 2023-06-26
cmake: Update googletest / glslang to fix CMake warnings

juan-lunarg:
Relevant PRs I made to fix these warnings in these repos:

https://github.com/KhronosGroup/glslang/pull/3241
https://github.com/google/googletest/pull/4288
https://github.com/google/googletest/pull/4290
https://github.com/google/googletest/pull/4293

### DIFF
--- a/scripts/known_good.json
+++ b/scripts/known_good.json
@@ -6,7 +6,7 @@
             "sub_dir": "glslang",
             "build_dir": "glslang/build",
             "install_dir": "glslang/build/install",
-            "commit": "d1517d64cfca91f573af1bf7341dc3a5113349c0",
+            "commit": "9575e33186c74a68831c469f7271edf386ea43a5",
             "optional": [
                 "tests"
             ]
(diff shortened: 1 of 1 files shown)


scripts/known_good.json
```diff
@@ -75,7 +75,7 @@
                 "-Dgtest_force_shared_crt=ON",
                 "-DBUILD_SHARED_LIBS=OFF"
             ],
-            "commit": "v1.13.0",
+            "commit": "ec4fed93217bc2830959bb8e86798c1d86956949",
```
artem-lunarg, line 78: We used tagged released previously, do they continue to create new releases? just to double check if it's fine. I heard they use live at the HEAD internally, so I guess it should be fine.

juan-lunarg: Google test very infrequently creates tagged releases. The have a live at head philosophy so it's fine to use specific commits.

> GoogleTest now follows the [Abseil Live at Head philosophy](https://abseil.io/about/philosophy#upgrade-support). We recommend [updating to the latest commit in the main branch as often as possible](https://github.com/abseil/abseil-cpp/blob/master/FAQ.md#what-is-live-at-head-and-how-do-i-do-it). We do publish occasional semantic versions, tagged with v${major}.${minor}.${patch} (e.g. v1.13.0).

artem-lunarg (approved):
I don't follow all details but it looks like a good clean up in other projects and also each one went through a review, so LGTM!

PR #6078, opened 2023-06-28
scripts: Refactoring codegen

spencer-lunarg:
The `parameter_validation` code gen is a lot... this is me getting in the "simple" stuff to make the 2nd phase review bearable 

-  renamed to `stateless_validation_helper`
- made dedicated generator for `enum_flag_bits.h` and `valid_enum_usage` (was `valid_param_usage`)
    - plan to just move the `valid_enum_usage` logic into the chassis or something, as it seemed to be created as a file due to the issue of not having a `VulkanObject` to work off of
- moved `stateless_validation_helper.h` to new `VulkanObject` (second phase is move the source)
- This accidentally fixed some missing items we were not validating from the old system :tada: 

Others:

- Moved `Thread Saftey`, `dispatch table`,`Function Pointers`, `spirv grammar` and `Layer Dispatch Table` code gen over as well


layers/vulkan/generated/vk_enum_string_helper.h
```diff
@@ -585,6 +585,18 @@ static inline const char* string_VkStructureType(VkStructureType input_value) {
             return "VK_STRUCTURE_TYPE_SWAPCHAIN_CREATE_INFO_KHR";
         case VK_STRUCTURE_TYPE_PRESENT_INFO_KHR:
             return "VK_STRUCTURE_TYPE_PRESENT_INFO_KHR";
+        case VK_STRUCTURE_TYPE_DEVICE_GROUP_PRESENT_CAPABILITIES_KHR:
```
spencer-lunarg, line 588: This was reordered because of the new code in `base_generator` to handle the duplicates there, shouldn't change again after

scripts/generators/vulkan_object.py
```diff
@@ -21,7 +21,7 @@
 # not supported until Python 3.9+
 from typing import List, Dict
 
-@dataclass(frozen=True)
```
spencer-lunarg, line 24: so I originally thought to make these frozen

- It makes the `base_generator` code way uglier when it needs to do a 2nd pass post-processing step
- the main `VulkanObject` wasn't frozen anyway (also class members in classes don't inherit the `frozen` I found :cry`)
- If a future use of `base_generator` wants to make a small tweak, I rather they be able to do it, then start re-creating yet-antoher-data-structure
- each generator gets a fresh copy anyway

arno-lunarg (approved):

PR #6079, opened 2023-06-28
sync: Fix semaphore timeout with gpu av

artem-lunarg:
Addresses this test case: https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4968#issuecomment-1354342939

### DIFF
--- a/layers/gpu_validation/gpu_validation.cpp
+++ b/layers/gpu_validation/gpu_validation.cpp
@@ -1453,6 +1453,7 @@ void GpuAssisted::PreCallRecordQueueSubmit2KHR(VkQueue queue, uint32_t submitCou
 }
 
 void GpuAssisted::PreCallRecordQueueSubmit2(VkQueue queue, uint32_t submitCount, const VkSubmitInfo2 *pSubmits, VkFence fence) {
+    ValidationStateTracker::PreCallRecordQueueSubmit2(queue, submitCount, pSubmits, fence);
     for (uint32_t submit_idx = 0; submit_idx < submitCount; submit_idx++) {
         const VkSubmitInfo2 *submit = &pSubmits[submit_idx];
         for (uint32_t i = 0; i < submit->commandBufferInfoCount; i++) {
--- a/tests/unit/gpu_av_positive.cpp
+++ b/tests/unit/gpu_av_positive.cpp
@@ -353,3 +353,43 @@ TEST_F(PositiveGpuAssistedLayer, GpuBufferDeviceAddress) {
     subcase_err = vk::QueueWaitIdle(m_device->m_queue);
     ASSERT_VK_SUCCESS(subcase_err);
 }
+
+// Regression test for semaphore timeout with GPU-AV enabled:
+// https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4968
+TEST_F(PositiveGpuAssistedLayer, GetCounterFromSignaledSemaphoreAfterSubmit) {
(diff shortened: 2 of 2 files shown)


layers/gpu_validation/gpu_validation.h
```diff
@@ -237,6 +237,19 @@ class GpuAssisted : public GpuAssistedBase {
     void PreCallRecordQueueSubmit2KHR(VkQueue queue, uint32_t submitCount, const VkSubmitInfo2KHR* pSubmits,
                                       VkFence fence) override;
     void PreCallRecordQueueSubmit2(VkQueue queue, uint32_t submitCount, const VkSubmitInfo2* pSubmits, VkFence fence) override;
+
+    // Inherited ValidationStateTracker::PostCallRecordGetSemaphoreCounterValue() works in combination
+    // with ValidationStateTracker::PreCallRecordQueueSubmit2() (or other submit commands) - without submit
+    // tracking PostCallRecordGetSemaphoreCounterValue will timeout.
+    //
+    // Currently, GPU AV does not use these state tracking endpoints. ValidationStateTracker::PreCallRecordQueueSubmit2()
```
jeremyg-lunarg, line 245: Should GPU-AV be calling the state tracker queue submit functions?

artem-lunarg: That's another option, which also works. I'm not sure why to add this specific state tracking if GPU-AV does not need it, except if we have a policy to include state tracking rather than not.

jeremyg-lunarg: I think this approach is fine for now because the GPU-AV QueueSubmit methods wait for idle so that they can process results.

jeremyg-lunarg (approved):

layers/gpu_validation/gpu_validation.h
Tony-LunarG: I think the real problem is that PreCallRecordQueueSubmit2 doesn't call state tracker. Or I'm not understanding the problem.

artem-lunarg: I discussed this with @jeremyg-lunarg, state tracking is tricky. From one side it feels more safe and easier just to track all state but you might track things that you don't need or expensive to track, on the other hand if you need to decide is it safe not to track something you have to check underlying state tracking details, which might be fine also Jeremy has some ideas to different ways to track state in general... Yeah, that's definitely not an easy life of functional language programmer.

artem-lunarg: @Tony-LunarG I think you are right. The bug was with QueueSubmit2 and I checked only it, but now I see that `QueueSubmit2KHR`, `QueueSubmit` call parent method, so even if state is not needed I lean more toward to do the same for `QueueSubmit2` for consistency.

jeremyg-lunarg: Either solution works for me. I wish we had a better way to make sure derived classes call the base class methods consistently.

Tony-LunarG: Changing GpuAssisted::PreCallRecordQueueSubmit2 to call ValidationStateTracker::PreCallRecordQueueSubmit2 fixes the test in this PR, is consistent with all of the other PreCallRecordQueueSubmits, and seems more straightforward than special casing PostCallRecordGetSemaphoreCounterValue.  Is there a reason that wouldn't work?

Tony-LunarG: Nevermind, I see our messages passed each other.

Tony-LunarG (approved):
LGTM

PR #6081, opened 2023-06-28
gpu: Handle OOB errors on mutable descriptors

jeremyg-lunarg:
Use the current active class of the mutable descriptor to decide how to report an error.


Tony-LunarG (approved):
LGTM and fixes asserting CTS tests

spencer-lunarg (approved):

layers/state_tracker/buffer_state.h
```diff
@@ -117,4 +117,12 @@ class BUFFER_VIEW_STATE : public BASE_NODE {
         BASE_NODE::Destroy();
     }
     bool Invalid() const override { return Destroyed() || !buffer_state || buffer_state->Invalid(); }
+
+    VkDeviceSize Size() const {
+        VkDeviceSize size = create_info.range;
+        if (size == VK_WHOLE_SIZE) {
+            size = buffer_state->createInfo.size - create_info.offset;
```
arno-lunarg, line 124: Reading the doc on `VkBufferViewCreateInfo::range`:
> If VK_WHOLE_SIZE is used and the remaining size of the buffer is not a multiple of the [texel block size](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#texel-block-size) of format, the nearest smaller multiple is used.

So should this code account for that?

arno-lunarg (approved):
LGTM! One question

PR #6083, opened 2023-06-28
Minor cccache CI fixes

juan-lunarg:
Adds ccache to CMake Android builds using CMAKE_\<LANG\>_COMPILER_LAUNCHER.

Also makes the macos build consistent with the other builds.

### DIFF
--- a/.github/workflows/vvl.yml
+++ b/.github/workflows/vvl.yml
@@ -184,6 +184,10 @@ jobs:
       steps:
         - uses: actions/checkout@v3
         - uses: lukka/get-cmake@latest
+        - name: Setup ccache
+          uses: hendrikmuhs/ccache-action@v1.2
+          with:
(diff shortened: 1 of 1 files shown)


.github/workflows/vvl.yml
```diff
@@ -230,12 +237,13 @@ jobs:
         uses: hendrikmuhs/ccache-action@v1.2
         with:
           key: macos-${{ matrix.macos_version }}
-      - name: Add ccache to PATH
-        run: echo "/usr/lib/ccache:/usr/local/opt/ccache/libexec" >> $GITHUB_PATH
```
juan-lunarg, line 234: Note: This approach of manipulating the PATH doesn't work for Android NDK builds. While the COMPILER_LAUNCHER approach does work without issue.

jeremyg-lunarg (approved):

spencer-lunarg (approved):

PR #6084, opened 2023-06-28
tests: Fix hiding class member warnings

juan-lunarg:


tests/unit/sync_val.cpp
```diff
@@ -2687,7 +2687,6 @@ struct SyncTestPipeline {
           sampler_info(SafeSaneSamplerCreateInfo()),
           sampler() {}
     void InitState() {
-        VkSamplerCreateInfo sampler_info = SafeSaneSamplerCreateInfo();
```
juan-lunarg, line 2690: Note. This is a pretty good example of the warning being useful.

sample_info was already initialized in the constructor with `SafeSaneSamplerCreateInfo()` then we define another `sample_info` the same way again and use that one instead. Do to what looks like a simple mistake.

https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4458?view=msvc-170

tests/framework/render.h
```diff
@@ -279,7 +279,7 @@ class VkRenderFramework : public VkTestFramework {
     // Device extensions to enable
     std::vector<const char *> m_device_extension_names;
 
-    VkValidationFeaturesEXT validation_features;
+    VkValidationFeaturesEXT m_validation_features;
```
jeremyg-lunarg, line 282: You are being consistent with the existing members but it is a shame that this class doesn't use trailing `_` like almost everything else. That's best left for a separate cleanup at some point.

jeremyg-lunarg (approved):
LGTM. The m_command_buffer non-member was especially in need of cleanup.

PR #6086, opened 2023-06-29
layers: Store buffers in `buffer_address_map_` using raw pointers

arno-lunarg:
Doing so improve memory accesses, it can make a good difference when performing a linear search. 
Improves upon https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6072

### DIFF
--- a/layers/state_tracker/cmd_buffer_state.h
+++ b/layers/state_tracker/cmd_buffer_state.h
@@ -405,10 +405,11 @@ class CMD_BUFFER_STATE : public REFCOUNTED_NODE {
         auto base = std::static_pointer_cast<BASE_NODE>(child_node);
         AddChild(base);
     }
-    template <typename StateObject>
-    void AddChildren(vvl::span<std::shared_ptr<StateObject>> &child_nodes) {
+
+    void AddChildren(vvl::span<BUFFER_STATE *> &child_nodes) {
         for (auto &child_node : child_nodes) {
-            AddChild(child_node);
+            auto child_node_shared_ptr = child_node->shared_from_this();
+            AddChild(child_node_shared_ptr);
         }
     }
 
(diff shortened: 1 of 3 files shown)


layers/state_tracker/state_tracker.h
```diff
@@ -1607,7 +1607,7 @@ class ValidationStateTracker : public ValidationObject {
 
     mutable VideoProfileDesc::Cache video_profile_cache_;
 
-    using BufferAddressMapStore = small_vector<std::shared_ptr<BUFFER_STATE>, 1, size_t>;
+    using BufferAddressMapStore = small_vector<BUFFER_STATE*, 1, size_t>;
```
jeremyg-lunarg, line 1610: It would be nice to have a comment somewhere explaining how it is safe to use a raw pointer in this case. Which is because we already have a reference count on the buffer when it is created and the address map entries for each buffer are removed when it is destroyed. Right?

artem-lunarg: +1 Was going to write similar comment, my understanding that pointer is safe because DestroyBuffer cleans up all raw pointers.

layers/state_tracker/cmd_buffer_state.h
```diff
@@ -406,9 +406,10 @@ class CMD_BUFFER_STATE : public REFCOUNTED_NODE {
         AddChild(base);
     }
     template <typename StateObject>
-    void AddChildren(vvl::span<std::shared_ptr<StateObject>> &child_nodes) {
+    void AddChildren(vvl::span<StateObject *> &child_nodes) {
```
artem-lunarg, line 409: Would it be a good approach to make `AddChildren` a helper function private to cpp file? Now that we use raw pointer instead of shared_ptr I guess we can have larger set of potential failure cases when someone decides to use it for other object types than buffers (the signature suggest it's general function for different object types). As I understand for buffers we have guarantee that pointers are safe due to DestroyBuffer handling, but it might not be the case for other object type.

layers/state_tracker/state_tracker.h
arno-lunarg: > It would be nice to have a comment somewhere explaining how it is safe to use a raw pointer in this case. Which is because we already have a reference count on the buffer when it is created and the address map entries for each buffer are removed when it is destroyed.
Yes that is the reason, I am going to write that down

layers/state_tracker/cmd_buffer_state.h
arno-lunarg: I went for Jeremy's suggestion, indeed no need for a template here

artem-lunarg (approved):

PR #6087, opened 2023-06-29
layers: Fix shader components counting

matkow01-arm:
I'm suspecting that in function `ValidateShaderStageInputOutputLimits` in `cc_shader.cpp` determining total input/output components are not correct.

Values are taken from `Entrypoint` variable which contain data about last input and last output variable declared in shader.
That values are assigned in `EntryPoint` constructor (shader_module.cpp) by searching through `StageInteraceVariable` struct for `InterfaceSlot` with highest value of slot (Location * 4 + Component)
That method works good when variables declared in shader occupy whole location space (4 components). The issue arises when in shader variable declaration occupy less than 4 components (e.g. float).
Example:
part of shader code:
```
layout(location=0) in float vin0; // occupy location 0, component 0
layout(location=1) in vec4 vin1; // occupy location 1, components 0-3
```
max slot index = 7 (**amount of components = 8**)
After swapping locations for this variables:
```
layout(location=0) in vec4 vin0; // occupy location 0, components 0-3
layout(location=1) in float vin1; // occupy location 1, component 0
```
max slot index = 4 (**amount of components = 5**)
Difference comes from equation `slot = Location * 4 + Component`
For same variables but layered differently max slot index is changed, so I suppose that can't be used as total input/output component value.

Shaders which are used by tests inside Validation Layers have always declare `vec4` variable (and sometimes last variable as `float` to intentionally exceed limits). This cases cannot reproduce issue that I was presented.

I propose solution which is similar to count buildin components.

### DIFF
--- a/layers/core_checks/cc_shader.cpp
+++ b/layers/core_checks/cc_shader.cpp
@@ -607,7 +607,7 @@ bool CoreChecks::ValidateShaderStageInputOutputLimits(const SHADER_MODULE_STATE 
                          pipeline.create_index);
     }
 
-    // The max is a combiniation of both the user defined variables largest values
+    // The max is a combination of both the user defined variables largest values
     // and
     // The total components used by built ins
     const auto max_input_slot =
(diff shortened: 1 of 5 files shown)


spencer-lunarg:
@matkow01-arm thanks for the PR, I noticed this 2 months ago and raised https://gitlab.khronos.org/vulkan/vulkan/-/issues/3448

I am happy to re-write the validation code, but we need actual spec language fixes to explicitly say "how" the max component is counted

matkow01-arm:
@spencer-lunarg Thanks for your response, I agree that in spec currently component counting isn't directly described. I search only issues in VL github so I didn't meet with your issue on Gitlab.

How can we deal with this PR? Do we want to reject it for now and someone from your side will re-make code when new spec will be released? Or we will leave this PR on hold and wait for spec update?

spencer-lunarg:
@matkow01-arm I wrote this code and more than happy to fix it... let me try and get some answers from the Gitlab issue, give me a week and I will take action on this... your code looks good to me (pending spec clarification)

matkow01-arm:
Yeah I totally understand your motivation for fixing it :) 
I will wait for your response in this PR and for now I will not take any additional actions for it.

matkow01-arm:
@spencer-lunarg Do you have any updates related to this topic?

spencer-lunarg:
@matkow01-arm so 2 weeks ago we [discussed this](https://gitlab.khronos.org/vulkan/vulkan/-/issues/3448#note_421252) and decided that what is needed is a spec clarification (and probably CTS cleanup), but just lacks anyone one to move that forward (I know @janharaldfredriksen-arm wasn't on that call and not sure saw that update)

I am happy to adjust the code in the VVL to whatever the Working Group decides

matkow01-arm:
Okay, so probably we can close/reject my PR. You will handle this components counting after spec update etc.
I don't see any advantage to hold this PR in open state :)

spencer-lunarg:
sounds good, I agree

PR #6091, opened 2023-06-29
tests: Fixes for macos

richard-lunarg:
Removed errnoneous default constructors.

Fixed macOS failures for:
VkAmdBestPracticesLayerTest.NumberOfSubmissions
NegativeGeometryTessellation.StageMaskGsTsEnabled
ClearAttachmentsCalledWithoutFbInSecondaryCB          
PushDescriptorTemplateDestroyDescriptorSetLayout     
PositiveCommand.ClearRectWith2DArray


tests/unit/amd_best_practices.cpp
```diff
@@ -638,6 +638,11 @@ TEST_F(VkAmdBestPracticesLayerTest, NumberOfSubmissions) {
     }
 
     InitState();
+    
+    if (IsExtensionsEnabled(VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME)) {
```
spencer-lunarg, line 642: so I feel all this has done is moved the "we skip this" from the blacklist.json to embedded in the file

What happens when this runs? There should be more information **why** this needs to be skipped

richard-lunarg: Fair enough. I think my initial guidance, or understanding of the situation (clean up VVL for macOS) was a bit off. I started over with a small PR to see really what you (maintainers) really cared about. I think the strategy should be to fix only the tests that will pass when the portability extension is properly applied, and leave everything else blacklisted... and the blacklisted tests should be audited periodically, say after every new MoltenVK release to see if any of them now pass, and can be removed from the blacklist. The list of failing tests can also be used as guidance to finding bugs in MoltenVK. Meanwhile, new tests written that fail on macOS can be evaluated at the time they are created to see if it's the fault of the test, or a driver bug we can blacklist/report as the test author sees fit.

PR #6093, opened 2023-06-30
Vk khr cooperative matrix vvl

archimedus:
Implemented test NegativeSubgroup.CooperativeMatrixKHR


layers/core_checks/cc_shader.cpp
```diff
@@ -1176,6 +1176,292 @@ bool CoreChecks::ValidateCooperativeMatrix(const SHADER_MODULE_STATE &module_sta
     return skip;
 }
 
+// Map SPIR-V type to VK_COMPONENT_TYPE enum
+VkComponentTypeKHR GetComponentTypeKHR(const Instruction *insn) {
```
spencer-lunarg, line 1180: nit: Can we move this into the `Instruction` class

spencer-lunarg, line 1180: Lets make this either a static function, lambda, or just inline this, I can't imagine this will be used elsewhere

spencer-lunarg: ok, nvm, I see you are just matching the `ValidateCooperativeMatrix` way we are doing it for the NV version... I guess for now this is fine

tests/unit/subgroups.cpp
```diff
@@ -841,8 +917,8 @@ TEST_F(NegativeSubgroup, CooperativeMatrixNV) {
         std::make_unique<VkShaderObj>(this, csSource, VK_SHADER_STAGE_COMPUTE_BIT, SPV_ENV_VULKAN_1_0, SPV_SOURCE_GLSL, &specInfo);
     pipe.InitState();
     pipe.pipeline_layout_ = VkPipelineLayoutObj(m_device, {});
-    m_errorMonitor->SetUnexpectedError("VUID-RuntimeSpirv-OpTypeCooperativeMatrixNV-06316");
-    m_errorMonitor->SetUnexpectedError("VUID-RuntimeSpirv-OpCooperativeMatrixMulAddNV-06317");
+    m_errorMonitor->SetUnexpectedError("VUID-RuntimeSpirv-OpTypeCooperativeMatrixNV-08974");
+    m_errorMonitor->SetUnexpectedError("VUID-RuntimeSpirv-MSize-08975");
```
spencer-lunarg, line 921: This test is for `VK_NV_cooperative_matrix`, why are we now hitting ` VUID-RuntimeSpirv-MSize-08975` which is for KHR only

spencer-lunarg (changes requested):
The layer code looks fine, we have a PR to get the 1.3.256 headers in, so will need to rebase on that (hopefully that will just be merged in the next 24 hours)

But this PR is missing actual testing

1. We need a test for every VU that was added
- [ ] VUID-RuntimeSpirv-cooperativeMatrixSupportedStages-08985
- [ ] VUID-RuntimeSpirv-OpTypeCooperativeMatrixKHR-08974
- [ ] VUID-RuntimeSpirv-OpCooperativeMatrixMulAddKHR-08976
- [ ] VUID-RuntimeSpirv-OpCooperativeMatrixMulAddKHR-08978
- [ ] VUID-RuntimeSpirv-OpCooperativeMatrixMulAddKHR-08980
- [ ] VUID-RuntimeSpirv-OpCooperativeMatrixMulAddKHR-08982
- [ ] VUID-VkPipelineShaderStageCreateInfo-module-08987
- [ ] VUID-RuntimeSpirv-scope-08984
- [ ] VUID-RuntimeSpirv-saturatingAccumulation-08983
- [ ] VUID-RuntimeSpirv-MSize-08975
- [ ] VUID-RuntimeSpirv-KSize-08977
- [ ] VUID-RuntimeSpirv-MSize-08979
- [ ] VUID-RuntimeSpirv-MSize-08981

2. We need at least 1 positive test (in `tests/unit/subgroups_positive.cpp`) that uses `VK_KHR_cooperative_matrix` but doesn't hit any Validation errors

archimedus:
> 1.  We need a test for every VU that was added

Well, I was basing on NV version, this one did not have any of such.

Will start implementing.


> 2. We need at least 1 positive test (in tests/unit/subgroups_positive.cpp) that uses VK_KHR_cooperative_matrix but doesn't hit any Validation errors

I am sorry I do not see `subgroups_positive.cpp` file in the project, do you mean I should add it?

kpet:
cc'ing @jeffbolznv since he took an action to start looking into this in the last Vulkan ML teleconference.

jeffbolznv:
> cc'ing @jeffbolznv since he took an action to start looking into this in the last Vulkan ML teleconference.

I hadn't done any work yet, and I don't object to somebody else doing it.

archimedus:
It looks like the only left 

* VUID-RuntimeSpirv-saturatingAccumulation-08983

archimedus:
> It looks like the only left
>
> *    VUID-RuntimeSpirv-saturatingAccumulation-08983

I don't know how to implement it in a sensible way.

It looks like it can force a lot of iterations: it looks we should find a pair of properties where saturatingAccumulation is true in one record and false in another. Number of properties returned by implementation can be as large as 11*11*11*2=2662 (i.e. 11 types for matrix A, B and C and saturatingAccumulation set to true/false), and number of comparisons in this case 2662<sup>2</sup>/2 = 3.5M.

spencer-lunarg:
> I don't know how to implement it in a sensible way.

Actually reading that VU, I am confused by it as well, 

> For OpCooperativeMatrixMulAddKHR, the SaturatingAccumulation cooperative matrix operand must be present if and only if VkCooperativeMatrixPropertiesKHR::saturatingAccumulation is VK_TRUE.

@jeffbolznv was this just suppose to mean if "any" `VkCooperativeMatrixPropertiesKHR` has `saturatingAccumulation` , or is there some implicit way we map a certain index of `VkCooperativeMatrixPropertiesKHR`  to each instruction?

tests/unit/shader_cooperative_matrix.cpp
```diff
@@ -0,0 +1,637 @@
...
+    // Ignore messages that types and sizes are unsupported by implementation
+    m_errorMonitor->SetAllowedFailureMsg("VUID-RuntimeSpirv-MSize-08975");
+    m_errorMonitor->SetAllowedFailureMsg("VUID-RuntimeSpirv-KSize-08977");
+    m_errorMonitor->SetAllowedFailureMsg("VUID-RuntimeSpirv-MSize-08979");
+    m_errorMonitor->SetAllowedFailureMsg("VUID-RuntimeSpirv-MSize-08981");
+    m_errorMonitor->SetAllowedFailureMsg("VUID-RuntimeSpirv-OpCooperativeMatrixMulAddKHR-08976");
+    m_errorMonitor->SetAllowedFailureMsg("VUID-RuntimeSpirv-OpCooperativeMatrixMulAddKHR-08978");
+    m_errorMonitor->SetAllowedFailureMsg("VUID-RuntimeSpirv-OpCooperativeMatrixMulAddKHR-08980");
+    m_errorMonitor->SetAllowedFailureMsg("VUID-RuntimeSpirv-OpCooperativeMatrixMulAddKHR-08982");
+    m_errorMonitor->SetUnexpectedError("VUID-RuntimeSpirv-OpTypeCooperativeMatrixKHR-06316");
```
spencer-lunarg, line 102: just fyi, this is not an actual VU

I try to avoid using`SetUnexpectedError` unless adding a comment why

tests/unit/shader_cooperative_matrix.cpp
```diff
@@ -0,0 +1,637 @@
...
+    };
+    auto replace = [](std::string &str, const std::string &from, const std::string &to) {
+        size_t pos;
+        while ((pos = str.find(from)) != std::string::npos) str.replace(pos, from.length(), to);
+    };
+
+    for (const auto &x : subtests) {
+        std::string css(csSourceTemplate);
+
+        for (const auto &y : subtests) {
```
spencer-lunarg, line 498: while this seems "clever" to do, we have repeated by experiences with these types of test logic around the VUID error messages

I would perfer just unrolling the loop, while its more verbose, its MUCH easier to maintain in the long run for us

... a better question, why are we having to ignore 8 VUs each time? why are we not able to query the properties and write a test that doesn't fail in `VUID-RuntimeSpirv-MSize-08975`

tests/unit/shader_cooperative_matrix_positive.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+    CreateComputePipelineHelper pipe(*this);
+    pipe.InitInfo();
+    pipe.cs_ =
+        std::make_unique<VkShaderObj>(this, csSource, VK_SHADER_STAGE_COMPUTE_BIT, SPV_ENV_VULKAN_1_0, SPV_SOURCE_GLSL, &specInfo);
+    pipe.InitState();
+    pipe.pipeline_layout_ = VkPipelineLayoutObj(m_device, {});
+    m_errorMonitor->SetUnexpectedError("VUID-RuntimeSpirv-OpTypeCooperativeMatrixNV-06316");
+    m_errorMonitor->SetUnexpectedError("VUID-RuntimeSpirv-OpCooperativeMatrixMulAddNV-06317");
+    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkPipelineShaderStageCreateInfo-pSpecializationInfo-06849");
```
spencer-lunarg, line 89: positive tests are not allowed to have any errors, they are designed to make sure when people use the API correctly, it won't produce false positive validation errors

tests/unit/shader_cooperative_matrix.cpp
```diff
@@ -0,0 +1,637 @@
...
+    };
+
+    CreateComputePipelineHelper pipe(*this);
+    pipe.InitInfo();
+    pipe.cs_ =
+        std::make_unique<VkShaderObj>(this, csSource, VK_SHADER_STAGE_COMPUTE_BIT, SPV_ENV_VULKAN_1_0, SPV_SOURCE_GLSL, &specInfo);
+    pipe.InitState();
+    pipe.pipeline_layout_ = VkPipelineLayoutObj(m_device, {});
+    // Ignore messages that types and sizes are unsupported by implementation
+    m_errorMonitor->SetAllowedFailureMsg("VUID-RuntimeSpirv-MSize-08975");
```
spencer-lunarg, line 383: why are we having to ignore 8 VUs each time? why are we not able to query the properties and write a test that doesn't fail in `VUID-RuntimeSpirv-MSize-08975`

jeffbolznv:
> > For OpCooperativeMatrixMulAddKHR, the SaturatingAccumulation cooperative matrix operand must be present if and only if VkCooperativeMatrixPropertiesKHR::saturatingAccumulation is VK_TRUE.
> 
> @jeffbolznv was this just suppose to mean if "any" `VkCooperativeMatrixPropertiesKHR` has `saturatingAccumulation` , or is there some implicit way we map a certain index of `VkCooperativeMatrixPropertiesKHR` to each instruction?

I didn't write it, but I interpret it to mean that if an instruction uses SaturatingAccumulation then there must be a property matching the instruction's types and dimensions that has saturatingAccumulation==TRUE.

I don't understand the previous comment. Realistically there aren't going to be thousands of cooperative matrix properties, and the complexity of this check could just be a hash map lookup.

archimedus:
> I don't understand the previous comment. Realistically there aren't going to be thousands of cooperative matrix properties

Upper limit properties reported: 11<sup>3</sup>*2. I agree that it will look weird to have A=i32, B=i32 and C=i8, though possible and I have no information what implementations will export.

> and the complexity of this check could just be a hash map lookup.

Could not agree with this. For example: assume we have only one property: A,B,C=i32, saturatingAccumulation=TRUE.
The test can easily be implemented: OpCooperativeMatrixMulAddKHR without SaturatingAccumulation. 

Now imagine two such properties reported:

A,B,C=i32, saturatingAccumulation=TRUE
A,B,C=i32, saturatingAccumulation=FALSE

Test CANNOT be implemented, because due to operation is allowed with SaturatingAccumulation (allowed by 1st properties) and without SaturatingAccumulation (allowed by 2nd properties). I.e. test CANNOT cause this VUID.

Thus test should take first record (with saturatingAccumulation=TRUE) and enumerate all left looking for 2nd record (with saturatingAccumulation=FALSE). Thus complexity is (11<sup>3</sup>*2)<sup>2</sup>/2.

tests/unit/shader_cooperative_matrix_positive.cpp
archimedus: This is NV code, I just moved it as you suggested from one file to another. 
I don't know why it was added, but it is optional within my environment, I will remove these two SetUnexpectedError. 

@jeffbolznv 
If you are not agree, please give me to know.

spencer-lunarg:
> test CANNOT cause this VUID

So for these types of things in the past we have used Device Profile API... honestly for the sake of getting this in... also we don't expect tests to be ran on EVERY device, just usually one

I am happy to get a test for `VUID-RuntimeSpirv-saturatingAccumulation-08983` after this PR, I rather not have this be blocking for this PR

tests/unit/shader_cooperative_matrix.cpp
spencer-lunarg: ok, fair, again for now I rather just get something in, can always be smoothed out in a follow up PR, but yes, as long as there is "a test" for "each VUID" that is the real important part

archimedus:
Thank you @spencer-lunarg for reviews.

> So for these types of things in the past we have used Device Profile API... honestly for the sake of getting this in...

I suppose I have seen something similar in CTS besides xgl and win32, there is null device. I was thinking is there something similar in VVL. Now I know there is :)

> I am happy to get a test for VUID-RuntimeSpirv-saturatingAccumulation-08983 after this PR, I rather not have this be blocking for this PR

It looks reasonable for me.

spencer-lunarg (approved):

scripts/known_good.json
```diff
@@ -2,11 +2,11 @@
     "repos": [
         {
             "name": "glslang",
-            "url": "https://github.com/KhronosGroup/glslang.git",
+            "url": "https://github.com/archimedus/glslang.git",
```
juan-lunarg, line 5: This should have not been accepted until this was merged into `KhronosGroup/glslang`.

We will need to revert these changes.

spencer-lunarg: sorry @juan-lunarg , I missed this

PR #6095, opened 2023-06-30
v1.3.256 update

mikes-lunarg:


layers/vulkan/generated/sync_validation_types.cpp
```diff
@@ -24,8 +24,8 @@
 
 
 
-const std::array<SyncStageAccessInfoType, 128>& syncStageAccessInfoByStageAccessIndex() {
```
spencer-lunarg, line 27: can confirm that replacing these generated files fix the failing sync tests, not sure why this is breaking it

artem-lunarg: By debugging before and after versions the issue points to UB with global initialization order, similar to what we had recently.  Continue looking into this.

Some global constants (`static const SyncStageAccessFlags SYNC_IMAGE_LAYOUT_TRANSITION_BIT = ...`)
contain zero values, instead of expected non-zero value, which suggests they are used too early, probably during initialization of the another global from different compilation unit

mikes-lunarg: Thanks to both of you, I rebased and CI is looking much happier.

juan-lunarg (approved):

artem-lunarg (approved):

PR #6099, opened 2023-07-04
scripts: Move SyncVal to VulkanObject

spencer-lunarg:
Got most of the code over to just using the new VulkanObjects, there is still some stuff that could be cleaned up, but for now this gets use 95% of the way there

This change also moves `layer_chassis_dispatch` over to use VulkanObject


layers/vulkan/generated/sync_validation_types.cpp
```diff
@@ -1548,29 +1542,22 @@ const std::map<VkPipelineStageFlags2, VkAccessFlags2>& syncDirectStageToAccessMa
...
         VK_PIPELINE_STAGE_2_RESOLVE_BIT |
         VK_PIPELINE_STAGE_2_BLIT_BIT |
-        VK_PIPELINE_STAGE_2_CLEAR_BIT
+        VK_PIPELINE_STAGE_2_CLEAR_BIT |
+        VK_PIPELINE_STAGE_2_ACCELERATION_STRUCTURE_COPY_BIT_KHR
     )},
     { VK_QUEUE_GRAPHICS_BIT, (
         VK_PIPELINE_STAGE_2_TOP_OF_PIPE_BIT |
         VK_PIPELINE_STAGE_2_DRAW_INDIRECT_BIT |
+        VK_PIPELINE_STAGE_2_VERTEX_INPUT_BIT |
```
artem-lunarg, line 1560: VERTEX_INPUT should not be here, it's a composite stage

spencer-lunarg:
```xml
        <syncstage name="VK_PIPELINE_STAGE_2_VERTEX_INPUT_BIT" alias="VK_PIPELINE_STAGE_VERTEX_INPUT_BIT">
            <syncsupport queues="graphics"/>
            <syncequivalent stage="VK_PIPELINE_STAGE_2_INDEX_INPUT_BIT,VK_PIPELINE_STAGE_2_VERTEX_ATTRIBUTE_INPUT_BIT"/>
        </syncstage>
 ```
 
 ok, I see, good catch, I check for not having a `equivalent`, must of missed here

File: layers/vulkan/generated/sync_validation_types.cpp [hunk 2]
```diff
@@ -1581,39 +1568,56 @@ const std::map<VkQueueFlagBits, VkPipelineStageFlags2>& syncAllCommandStagesByQu
...
+        VK_PIPELINE_STAGE_2_ACCELERATION_STRUCTURE_COPY_BIT_KHR |
+        VK_PIPELINE_STAGE_2_MICROMAP_BUILD_BIT_EXT
     )},
-    { VK_QUEUE_TRANSFER_BIT, (
+    { VK_QUEUE_PROTECTED_BIT, (
+        VK_PIPELINE_STAGE_2_TOP_OF_PIPE_BIT |
+        VK_PIPELINE_STAGE_2_BOTTOM_OF_PIPE_BIT |
+        VK_PIPELINE_STAGE_2_HOST_BIT
+    )},
+    { VK_QUEUE_SPARSE_BINDING_BIT, (
```
artem-lunarg, line 1611: A note for documentation purposes:
In previous the previous script version we intentionally skipped sparse queue:
`if queue_name == 'sparse_binding': continue`
but it could be it was done only to have the same output as in the very first implementation.

+1 for keeping it here in case everything works as expected (and also PROTECTED queue).

scripts/generators/delete.py
artem-lunarg, line 1: why this file name?

artem-lunarg (dismissed):
LGTM! assuming that VK_PIPELINE_STAGE_2_VERTEX_INPUT_BIT is removed

scripts/generators/sync_validation_generator.py
```diff
@@ -105,163 +71,302 @@ def __init__(self,
...
+''')
+
+        out.append('// Unique number for each  stage/access combination\n')
+        out.append('enum SyncStageAccessIndex {\n')
+        for access in self.stageAccessCombo:
+            out.append(f'    {access["stage_access"]} = {access["index"]},\n')
+        out.append('};\n')
+
+        out.append('\n')
+        out.append('using SyncStageAccessFlags = std::bitset<128>;\n')
```
artem-lunarg, line 170: Could you increase this to 192 (+64, one machine word) to fix https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6095#discussion_r1252941607.

More proper solution is to calculate precise value based on the number of enum members, but if takes additional time then it's not critical here and can be done later.

spencer-lunarg: oooo, this is the bug, ok will get that going

artem-lunarg: thanks!

artem-lunarg (approved):

PR #6104, opened 2023-07-06
scripts: Move ObjectTracker to VulkanObjects

spencer-lunarg:
ObjectTracker scripts now use `VulkanObject`


layers/vulkan/generated/object_tracker.cpp
```diff
@@ -2045,7 +2042,7 @@ bool ObjectLifetimes::PreCallValidateBindImageMemory2(
     if (pBindInfos) {
         for (uint32_t index0 = 0; index0 < bindInfoCount; ++index0) {
             skip |= ValidateObject(pBindInfos[index0].image, kVulkanObjectTypeImage, false, "VUID-VkBindImageMemoryInfo-image-parameter", "VUID-VkBindImageMemoryInfo-commonparent", "VkBindImageMemoryInfo");
-            skip |= ValidateObject(pBindInfos[index0].memory, kVulkanObjectTypeDeviceMemory, true, kVUIDUndefined, "VUID-VkBindImageMemoryInfo-commonparent", "VkBindImageMemoryInfo");
```
spencer-lunarg, line 2048: I confirmed this should be correct, not sure why this was `true` before

layers/vulkan/generated/object_tracker.cpp
```diff
@@ -5096,7 +5075,7 @@ bool ObjectLifetimes::PreCallValidateCmdBeginTransformFeedbackEXT(
     skip |= ValidateObject(commandBuffer, kVulkanObjectTypeCommandBuffer, false, "VUID-vkCmdBeginTransformFeedbackEXT-commandBuffer-parameter", "VUID-vkCmdBeginTransformFeedbackEXT-commonparent", "vkCmdBeginTransformFeedbackEXT");
     if ((counterBufferCount > 0) && (pCounterBuffers)) {
         for (uint32_t index0 = 0; index0 < counterBufferCount; ++index0) {
-            skip |= ValidateObject(pCounterBuffers[index0], kVulkanObjectTypeBuffer, true, kVUIDUndefined, "VUID-vkCmdBeginTransformFeedbackEXT-commonparent", "vkCmdBeginTransformFeedbackEXT");
```
artem-lunarg, line 5099: It looks this one allowed to be null:
```
VUID-vkCmdEndTransformFeedbackEXT-counterBufferCount-02608
If counterBufferCount is not 0, and pCounterBuffers is not NULL, pCounterBuffers must be a valid
pointer to an array of counterBufferCount VkBuffer handles that are either valid or [VK_NULL_HANDLE]
(https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VK_NULL_HANDLE)
```

artem-lunarg, line 5099: This one is allowed to be null:
```
VUID-vkCmdBeginTransformFeedbackEXT-counterBufferCount-02607
If counterBufferCount is not 0, and pCounterBuffers is not NULL, pCounterBuffers must be a valid
pointer to an array of counterBufferCount VkBuffer handles that are either valid or [VK_NULL_HANDLE]
(https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VK_NULL_HANDLE)
```

spencer-lunarg: hmmm, seems something is off with how I am reading this, luckily `VkPositiveLayerTest.DestroySwapchainWithBoundImages` caught this as well... will look more again

spencer-lunarg: ok, figured it out, these can be `null`, but the issue was only checking the `optional` field and not the `noAutoValidity` also

spencer-lunarg: ok, fixed (and reverted)

artem-lunarg (approved):

PR #6105, opened 2023-07-06
layers: Add write lock to protect timeline modification

artem-lunarg:
Found accidentally while working on a different sync issue. Not sure if we had actual issues because of this.


jeremyg-lunarg (approved):
LGTM, good find.

spencer-lunarg (approved):

PR #6110, opened 2023-07-07
scripts: Move Stateless source to VulkanObject

spencer-lunarg:
This moves the Stateless validation to the new `VulkanObject`

this is easily the most complex script we have, I didn't "fully" covert it, but got 90% of the way... there is a **LOT** to do in that code (Split it up, better naming, testing it!!)

For now, it is my main goal to get things to use the `VulkanObject`


layers/vulkan/generated/stateless_validation_helper.cpp
```diff
@@ -2360,8 +2363,10 @@ bool StatelessValidation::ValidatePnextStructContents(const char *api_name, cons
             }
         } break;
 #endif // VK_USE_PLATFORM_WIN32_KHR
+#ifdef VK_USE_PLATFORM_WIN32_KHR
```
spencer-lunarg, line 2366: these will get worked out in a future update

layers/vulkan/generated/stateless_validation_helper.cpp
```diff
@@ -10935,9 +10989,9 @@ bool StatelessValidation::PreCallValidateGetDeviceGroupPresentCapabilitiesKHR(
     VkDevice                                    device,
     VkDeviceGroupPresentCapabilitiesKHR*        pDeviceGroupPresentCapabilities) const {
     bool skip = false;
-    if (!instance_extensions.vk_feature_version_1_1) skip |= OutputExtensionError("vkGetDeviceGroupPresentCapabilitiesKHR", "VK_VERSION_1_1");
     if (!IsExtEnabled(device_extensions.vk_khr_surface)) skip |= OutputExtensionError("vkGetDeviceGroupPresentCapabilitiesKHR", "VK_KHR_surface");
-    if (!IsExtEnabled(device_extensions.vk_khr_swapchain)) skip |= OutputExtensionError("vkGetDeviceGroupPresentCapabilitiesKHR", "VK_KHR_swapchain");
+    if (!IsExtEnabled(device_extensions.vk_khr_device_group_creation)) skip |= OutputExtensionError("vkGetDeviceGroupPresentCapabilitiesKHR", "VK_KHR_device_group_creation");
```
spencer-lunarg, line 10993: there is a long going issue to cleanup the whole "fun" that is extension dependency, I looked and they all "correct" but might not be "the most correct" (some missing edge cases)

layers/vulkan/generated/stateless_validation_helper.cpp
```diff
@@ -13258,14 +13235,6 @@ bool StatelessValidation::PreCallValidateGetDisplayPlaneCapabilities2KHR(
     return skip;
 }
 
-
```
spencer-lunarg, line 13261: so many lost `\n` :vulcan_salute:

layers/vulkan/generated/stateless_validation_helper.cpp
```diff
@@ -16601,10 +16322,6 @@ bool StatelessValidation::PreCallValidateCreateAccelerationStructureNV(
 
         skip |= ValidateStructPnext("vkCreateAccelerationStructureNV", "pCreateInfo->info.pNext", nullptr, pCreateInfo->info.pNext, 0, nullptr, GeneratedVulkanHeaderVersion, "VUID-VkAccelerationStructureInfoNV-pNext-pNext", kVUIDUndefined, false, true);
 
-        skip |= ValidateRangedEnum("vkCreateAccelerationStructureNV", "pCreateInfo->info.type", "VkAccelerationStructureTypeNV", pCreateInfo->info.type, "VUID-VkAccelerationStructureInfoNV-type-parameter");
```
spencer-lunarg, line 16604: there is some nasty issues with how the NV an KHR Ray Tracing aliased each other, for now, rather have no false positives

layers/vulkan/generated/stateless_validation_helper.cpp
spencer-lunarg: ... the `NegativeDynamicState` tests beg to differ...

layers/vulkan/generated/stateless_validation_helper.cpp
```diff
@@ -10783,15 +10839,12 @@ bool StatelessValidation::PreCallValidateGetPhysicalDeviceSurfacePresentModesKHR
...
 
-
-
 bool StatelessValidation::PreCallValidateCreateSwapchainKHR(
     VkDevice                                    device,
     const VkSwapchainCreateInfoKHR*             pCreateInfo,
     const VkAllocationCallbacks*                pAllocator,
     VkSwapchainKHR*                             pSwapchain) const {
     bool skip = false;
-    if (!IsExtEnabled(device_extensions.vk_khr_surface)) skip |= OutputExtensionError("vkCreateSwapchainKHR", "VK_KHR_surface");
```
spencer-lunarg, line 10794: We have checks elsewhere that `VK_KHR_swapchain` can't be enabled without `VK_KHR_surface` so these have been redundant

layers/vulkan/generated/stateless_validation_helper.cpp
```diff
@@ -10935,9 +10984,7 @@ bool StatelessValidation::PreCallValidateGetDeviceGroupPresentCapabilitiesKHR(
     VkDevice                                    device,
     VkDeviceGroupPresentCapabilitiesKHR*        pDeviceGroupPresentCapabilities) const {
     bool skip = false;
-    if (!instance_extensions.vk_feature_version_1_1) skip |= OutputExtensionError("vkGetDeviceGroupPresentCapabilitiesKHR", "VK_VERSION_1_1");
```
spencer-lunarg, line 10938: for version, one of 2 case

1. The version will enable the extension for us before we get here
2. For cases like this, you still need the extensions regardless of the extension

layers/vulkan/generated/stateless_validation_helper.cpp
```diff
@@ -18427,11 +17673,6 @@ bool StatelessValidation::PreCallValidateGetAccelerationStructureOpaqueCaptureDe
     const VkAccelerationStructureCaptureDescriptorDataInfoEXT* pInfo,
     void*                                       pData) const {
     bool skip = false;
-    if (!(IsExtEnabled(device_extensions.vk_khr_acceleration_structure) || IsExtEnabled(device_extensions.vk_nv_ray_tracing))) skip |= OutputExtensionError("vkGetAccelerationStructureOpaqueCaptureDescriptorDataEXT", "VK_KHR_acceleration_structure || VK_NV_ray_tracing");
```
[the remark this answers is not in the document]
spencer-lunarg: image

image

Yes... this is the 1% of edge cases that still need to be solved, but lack tests and I have a list of these to look into after this PR (so it is ready by SDK time), the real goal was to remove the hack in the old code gen we used to get this (as it added all the redundant checks). Luckily these `false negatives` are very hard edge cases not likely

scripts/generators/safe_struct_generator.py
```diff
@@ -155,12 +167,13 @@ def generateHeader(self):
                             out.append(f'    safe_{member.type} {member.name};\n')
                         continue
 
-                initialize = '{}' if member.pointer and canInitialize else ''
-                if member.pointer and canInitialize:
-                    canInitialize = not struct.union # Prevents union from initializing agian
+                explicitInitialize = member.pointer  and 'PFN_' not in member.type and canInitialize
```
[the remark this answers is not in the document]
spencer-lunarg: yes, that needs to be cleaned up, I was actually curious if other things checking for `pointer` ment go have these or not

Was trying to decide if having a `pointer` and `functionPointer` field made sense or not

artem-lunarg (approved):

scripts/generators/stateless_validation_helper_generator.py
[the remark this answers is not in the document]
spencer-lunarg: Thanks for giving this a good look, there is a point where you almost just have to go "well its definitely not worse then it was before"... this script will get some improvements after, but want to unlock other from using `VulkanObject`

PR #6111, opened 2023-07-07
layers: Make CB_STATE an enum class

juan-lunarg:
Helps readability, and make it easier to search for all instances of the state being used since I can search for `CB_STATE::`


layers/state_tracker/cmd_buffer_state.h
```diff
@@ -108,12 +108,12 @@ class COMMAND_POOL_STATE : public BASE_NODE {
 // Autogenerated as part of the command_validation.h codegen
 const char *CommandTypeString(CMD_TYPE type);
 
-enum CB_STATE {
-    CB_NEW,                 // Newly created CB w/o any cmds
-    CB_RECORDING,           // BeginCB has been called on this CB
-    CB_RECORDED,            // EndCB has been called on this CB
-    CB_INVALID_COMPLETE,    // had a complete recording, but was since invalidated
-    CB_INVALID_INCOMPLETE,  // fouled before recording was completed
+enum class CB_STATE {
```
jeremyg-lunarg, line 111: Since you are touching every line using this enum I highly recommend also making this a non-shouty enum conforming to the style guid.

jeremyg-lunarg (approved):
Ahh much better

PR #6113, opened 2023-07-07
layers: Fix descriptor copies with a mutable source

jeremyg-lunarg:
We need to use the current type of each mutable descriptor in the binding, rather than the type of the binding (which would always be mutable).

### DIFF
--- a/layers/state_tracker/descriptor_sets.cpp
+++ b/layers/state_tracker/descriptor_sets.cpp
@@ -531,7 +531,12 @@ void cvdescriptorset::DescriptorSet::PerformCopyUpdate(const VkCopyDescriptorSet
         auto &src = *src_iter;
         auto &dst = *dst_iter;
         if (src_iter.updated()) {
-            dst.CopyUpdate(*this, *state_data_, src, src_iter.CurrentBinding().IsBindless(), src_iter.CurrentBinding().type);
+            auto type = src_iter.CurrentBinding().type;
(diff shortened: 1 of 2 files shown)


layers/state_tracker/descriptor_sets.cpp
```diff
@@ -531,7 +531,12 @@ void cvdescriptorset::DescriptorSet::PerformCopyUpdate(const VkCopyDescriptorSet
         auto &src = *src_iter;
         auto &dst = *dst_iter;
         if (src_iter.updated()) {
-            dst.CopyUpdate(*this, *state_data_, src, src_iter.CurrentBinding().IsBindless(), src_iter.CurrentBinding().type);
+            auto type = src_iter.CurrentBinding().type;
+            if (type == VK_DESCRIPTOR_TYPE_MUTABLE_EXT) {
+                const auto &mutable_src = static_cast<const MutableDescriptor &>(src);
+                type = mutable_src.ActiveType();
+            }
+            dst.CopyUpdate(*this, *state_data_, src, src_iter.CurrentBinding().IsBindless(), type);
```
juan-lunarg, line 539: Should their be a test added? Or does this fix an existing problematic test?

jeremyg-lunarg: this is a fix for CTS. but yeah we probably do need a unit test.

Tony-LunarG (approved):
LGTM

PR #6117, opened 2023-07-09
Add computeFullSubgroups stage VU

spencer-lunarg:
Adds `VUID-VkPipelineShaderStageCreateInfo-flags-08988` (added in 1.3.256 header)

### DIFF
--- a/tests/unit/subgroups.cpp
+++ b/tests/unit/subgroups.cpp
@@ -567,10 +567,6 @@ TEST_F(NegativeSubgroup, SubgroupSizeControlFeaturesNotEnabled) {
     VkPhysicalDeviceVulkan11Properties props11 = LvlInitStruct<VkPhysicalDeviceVulkan11Properties>();
     GetPhysicalDeviceProperties2(props11);
 
-    if (props11.subgroupSize == 0) {
-        GTEST_SKIP() << "subgroupSize is 0, skipping test.";
-    }
-
     std::stringstream csSource;
     // Make sure compute pipeline has a compute shader stage set
     csSource << R"(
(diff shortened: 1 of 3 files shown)


tests/unit/subgroups.cpp
```diff
@@ -599,6 +599,55 @@ TEST_F(NegativeSubgroup, SubgroupSizeControlFeaturesNotEnabled) {
...
+    if (sscf.subgroupSizeControl == VK_FALSE || sscf.computeFullSubgroups == VK_FALSE || sscf.subgroupSizeControl == VK_FALSE) {
+        GTEST_SKIP() << "Required features are not supported";
+    }
+    ASSERT_NO_FATAL_FAILURE(InitState(nullptr, &features2));
+    ASSERT_NO_FATAL_FAILURE(InitRenderTarget());
+
+    VkPhysicalDeviceVulkan11Properties props11 = LvlInitStruct<VkPhysicalDeviceVulkan11Properties>();
+    GetPhysicalDeviceProperties2(props11);
+    if (props11.subgroupSize == 0) {
+        GTEST_SKIP() << "subgroupSize is 0.";
```
artem-lunarg, line 626: According to spec this can't be 0 (shoud be power of two). Does this happen on the mock driver? In that case we can skip the mock driver instead. Checking for 0 gives this value some legitimacy.

artem-lunarg (approved):
LGTM. One question.

tests/unit/subgroups.cpp
spencer-lunarg: aww, this was copied and pasted, 99.9% chance this is left over from when we didn't have proper profile/mockICD support and it would return zero sometimes... let me just remove these all everywhere

PR #6118, opened 2023-07-10
docs: Add design philosophy for new code gen flow

spencer-lunarg:
Small write up of things I observed when refactoring all the python scripts

### DIFF
--- a/docs/generated_code.md
+++ b/docs/generated_code.md
@@ -3,7 +3,11 @@
 There is a lot of code generated in `layers/vulkan/generated/`. This is done to prevent errors forgetting to add support for new
 values when the Vulkan Headers or SPIR-V Grammer is updated.
 
-How to generate the code:
+- [How to generate the code](#how-to-generate-the-code)
+- [Adding and Editing code generation](#adding-and-editing-code-generation)
+- [How it works](#how-it-works)
+
+# How to generate the code
 
 - Linux:
 ```bash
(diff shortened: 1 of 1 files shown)


docs/generated_code.md
```diff
@@ -34,18 +38,57 @@ cmake --build build --target vvl_codegen
...
+    print(struct.name)
+```
+
+## Design philosophy
+
+> Written by someone who has written bad Vulkan code gen, debugged other's bad code gen, and rewrote all the scripts.
+
+### Avoid functions when possible
+
+All code gen has a single function that outputs one large string, there is zero dynamic control flow that occurs.
```
artem-lunarg, line 72: > there is zero dynamic control flow that occurs
This can also mean to avoid branching. Is it mostly about function calls?

spencer-lunarg: There is "branching" as in 

```python
if extension.instance:
    print("Instance")
else:
    print("Device")
```

But we have no runtime parameters for these scripts is what I ment

I guess what I want to prove that wrapping 5 lines of code to generate a string that used in 4 different places is better in the long run to just inline it. I find the biggest issue maintaining these scripts is not being able to look at the script as a linear string concatenation as the file output is... not sure how to word this the best

artem-lunarg: if that's about "zero dynamic control flow based on external parameters" or similar that maybe to say this directly? but if not, please ignore.

juan-lunarg (approved):

artem-lunarg (approved):

PR #6124, opened 2023-07-11
tests: Fix json in max profile

artem-lunarg:
`vkconfig` complains about invalid json and this allowed to detect this minor issue


juan-lunarg (approved):
LGTM

FYI @christophe-lunarg

PR #6125, opened 2023-07-11
sync: Clarify aspect mask specification

jzulauf-lunarg:
@artem-lunarg -- here's the small clarification you requested...

### DIFF
--- a/layers/containers/subresource_adapter.cpp
+++ b/layers/containers/subresource_adapter.cpp
@@ -284,7 +284,8 @@ ImageRangeEncoder::ImageRangeEncoder(const IMAGE_STATE& image, const AspectParam
 
     // WORKAROUND for profile and mock_icd not containing valid VkSubresourceLayout yet. Treat it as optimal image.
     if (image.createInfo.tiling == VK_IMAGE_TILING_LINEAR) {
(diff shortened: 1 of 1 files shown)


layers/containers/subresource_adapter.cpp
```diff
@@ -284,7 +284,8 @@ ImageRangeEncoder::ImageRangeEncoder(const IMAGE_STATE& image, const AspectParam
 
     // WORKAROUND for profile and mock_icd not containing valid VkSubresourceLayout yet. Treat it as optimal image.
     if (image.createInfo.tiling == VK_IMAGE_TILING_LINEAR) {
-        subres = {static_cast<VkImageAspectFlags>(AspectBit(0)), 0, 0};
+        const auto first_aspect = static_cast<VkImageAspectFlags>(AspectBit(0));  // AspectBit returns aspects by index
```
artem-lunarg, line 287: Thanks! Yes, I remember I was a bit confused by all these bits/flags flow and made incorrect assumption.
We probably can be even more clear by avoiding `static_cast`?
```
const VkImageAspectFlags first_aspect = AspectBit(0);
subres = {first_aspect, 0, 0};
```

artem-lunarg (approved):

PR #6126, opened 2023-07-12
build: Bump python version to 3.10

spencer-lunarg:
Python 3.10 provide some helpful features around making `VulkanObject` much easier to use

Main two features

1. Can use `list` and `dict` directly without `typing` library
2. Can make fields unions, which is nice to show that some fields might be `None`


spencer-lunarg:
seems CI is happy, will be applying to another change that changes all things that required 3.10

PR #6127, opened 2023-07-12
Final VulkanObject cleanup

spencer-lunarg:
These are the final "cleanup" for `VulkanObject` and after we can call the code generation cleanup/refactor as over :tada:


scripts/generators/base_generator.py
```diff
@@ -616,7 +551,7 @@ def genType(self, typeInfo, typeName, alias):
             device = not instance
             dispatchable = typeElem.find('type').text == 'VK_DEFINE_HANDLE'
 
-            self.vk.handles[typeName] = Handle(typeName, type, parent, protect, instance, device, dispatchable)
+            self.vk.handles[typeName] = Handle(typeName, type, protect, parent, instance, device, dispatchable)
```
artem-lunarg, line 554: Do linters/IDEs warn about wrong order arguments based on datatypes? Mostly wondering how easy is to make mistake in such constructor expressions.

scripts/generators/format_utils_generator.py
```diff
@@ -26,30 +26,18 @@ def getClassName(className: str) -> str:
...
-            return True
-    return False
+def formatHasDepth(format: Format) -> bool:
+    return any(x.type == 'D' for x in format.components)
+
+def formatHasStencil(format: Format) -> bool:
+    return any(x.type == 'S' for x in format.components)
+
+def formatHas64Bit(format: Format) -> bool:
+    return any(x.bits == '64' for x in format.components)
```
artem-lunarg, line 36: that's much nicer to read and to validate

scripts/generators/safe_struct_generator.py
```diff
@@ -168,8 +156,7 @@ def generateHeader(self):
                 # Prevents union from initializing agian
                 canInitialize = not struct.union if explicitInitialize else canInitialize
 
-                # TOOD - don't need formatted helper
-                if (getFormatedLength(member.length) is not None) and self.containsObjectHandle(member) and not member.staticArray:
```
artem-lunarg, line 172: All those special cases in `getFormatedLength`, is it something we don't need anymore ? It was quite complicated so good riddance assuming we don't need those special cases.

scripts/run_generator.py
```diff
@@ -56,356 +56,264 @@ def RunGenerator(api: str, registry: str, grammar: str, directory: str, target:
...
+    genOpts = {
+        'thread_safety_counter_definitions.h' : {
+            'generator' : ThreadSafetyOutputGenerator,
+            'options' : [],
+            'baseGenOptions' : []
+        },
+        'thread_safety_counter_instances.h' : {
+            'generator' : ThreadSafetyOutputGenerator,
+            'options' : [],
+            'baseGenOptions' : []
```
artem-lunarg, line 79: `baseGenOptions` - is it something we might want to customize in some future scenario?

scripts/generators/vulkan_object.py
```diff
@@ -85,17 +83,17 @@ class Param:
...
-    const: bool # type contains 'const'
-    length: str # the known length of pointer, will never be 'null-terminated'
-    nullTerminated: bool # If a UTF-8 string that will be null-terminated
-    pointer: bool # type contains a pointer
-    staticArray: List[str]
+    const: bool            # type contains 'const'
+    length:  (str | None)  # the known length of pointer, will never be 'null-terminated'
+    nullTerminated: bool   # If a UTF-8 string that will be null-terminated
+    pointer: bool          # type contains a pointer
+    staticArray: list[str] # for VkTransformMatrixKHR:matrix this is [3, 4]
```
artem-lunarg, line 90: nit and maybe just me. `static` might be considered overloaded term, one example, you might need to check if it's not related to `static` declaration in C language. `fixedSizeArray` might be more precise. I guess `static` is also used in this context and could be just different habits.

scripts/generators/safe_struct_generator.py
spencer-lunarg: ya, so there is basically `length`, `nullTerminated` and `staticArray` fields which hold all the information this function was trying to do

scripts/run_generator.py
spencer-lunarg: maybe, main candidate will be Vulkan SC if anything... I guess if a year goes by and nothing has changed, would be worth removing

scripts/generators/vulkan_object.py
spencer-lunarg: I like `fixedSizeArray` a lot better, easier to understand, let me update it

Issue #6101, opened 2023-07-05
Debug Printf just before VK_DEVICE_LOST


dorian-apanel-intel:
Hi, Debug Printf is a very useful feature. In [documentation](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/docs/debug_printf.md) there is no mention on behavior in case of VK_ERROR_DEVICE_LOST.

I think that having opportunity do insert Printf in failing shader would be an even better debug tool.
Maybe this could be enabled if `Printf buffer` mentioned [here](https://www.lunarg.com/wp-content/uploads/2021/08/Using-Debug-Printf-02August2021.pdf) was allocated optionally with [VK_MEMORY_PROPERTY_DEVICE_UNCACHED_BIT_AMD ](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_AMD_device_coherent_memory.html)?

Issue #5895, opened 2023-05-23
layer_configuration.md does not describe programmatic configuration of layers


KarenGhavam-lunarG:
The file https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/LAYER_CONFIGURATION.md does not explain how to enable layers programmatically. I think it should. This file is used by the SDK and referenced by all other layers in the SDK as the general guide on how to configure layers.

I know we are working on defining an API for programmatically configuring layers, should we wait to update this referenced documentation until that API is defined and made public? @christophe-lunarg

mark-lunarg:
The existing `vk_layer_settings_ext.h` allows setting layer options programmatically.  When the updated version arrives with the rework the method will be very similar with the exception of having to include the header file.

KarenGhavam-lunarG:
See [#5895](https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5895). layer_configuration.md describes how to configure layers and is used in the SDK documentation and referenced by all SDK layers. I believe this document should be updated with how to configure layers programmatically.

christophe-lunarg:
Unless there is an explicit request from a user, I think we should way until the updated extenion is ready as this will be relatively soon.

juan-lunarg:
Assigning @christophe-lunarg due to their recent work in VVL.

christophe-lunarg:
This issue is resolved by https://github.com/KhronosGroup/Vulkan-Utility-Libraries/pull/85

spencer-lunarg:
as @christophe-lunarg said, this seems resolved and actually just missing someone closing the issue

PR #6128, opened 2023-07-12
layers: Allow using uncached buffer for Debug Printf 

dorian-apanel-intel:
Adds option to force using AMD_DEVICE_COHERENT_MEMORY for debug printf buffer to print messages even if VK_ERROR_DEVICE_LOST is encountered.

Forcing extension and device feature if not enabled by application.

Added workaround for atomic operations in
uncached memory being in cache anyway.

Added workaround to failing MapMemory after DEVICE_LOST (occurs on AMD): When using uncached buffer, do not unmap buffer until messages are analyzed.

This PR attempts to implement #6101 Tested on AMD and Intel (extension not public yet) using this modified vkcube: https://github.com/dorian-apanel-intel/Vulkan-Tools/commit/e0226930a7958bb4964428691dd1376c3447bca3

### DIFF
--- a/layers/gpu_validation/debug_printf.cpp
+++ b/layers/gpu_validation/debug_printf.cpp
@@ -21,6 +21,87 @@
 #include <iostream>
 #include "generated/layer_chassis_dispatch.h"
 
+void DebugPrintf::PreCallRecordCreateDevice(VkPhysicalDevice gpu, const VkDeviceCreateInfo *pCreateInfo,
+                                            const VkAllocationCallbacks *pAllocator, VkDevice *pDevice, void *modified_ci) {
+    GpuAssistedBase::PreCallRecordCreateDevice(gpu, pCreateInfo, pAllocator, pDevice, modified_ci);
+
+    std::string use_uncached_buffer_string = getLayerOption("khronos_validation.printf_uncached_buffer");
+    vvl::ToLower(use_uncached_buffer_string);
+    use_uncached_buffer = !use_uncached_buffer_string.compare("true");
+
+    if (use_uncached_buffer)
+    {
+        static const std::string dcm_ext{"VK_AMD_device_coherent_memory"};
+        {
+            bool dcm_supported = false;
+            uint32_t property_count = 0;
+            if (DispatchEnumerateDeviceExtensionProperties(gpu, nullptr, &property_count, nullptr) == VK_SUCCESS) {
+                std::vector<VkExtensionProperties> property_list(property_count);
+                if (DispatchEnumerateDeviceExtensionProperties(gpu, nullptr, &property_count, property_list.data()) ==
+                    VK_SUCCESS) {
+                    for (const VkExtensionProperties &properties : property_list) {
+                        if (dcm_ext == properties.extensionName) {
+                            dcm_supported = true;
+                        }
+                    }
+                }
+            }
+            if (!dcm_supported) {
+                ReportSetupProblem(
+                    device, "Debug Printf with uncached buffer requires VK_AMD_device_coherent_memory which is not supported");
+                aborted = true;
+                return;
+            }
+        }
+
+        // See CreateDevice() in chassis.cpp. modified_ci is a pointer to a safe struct stored on the stack.
+        // This code follows the safe struct memory memory management scheme. That is, we must delete any memory
+        // remove from the safe struct, and any additions must be allocated in a way that is compatible with
+        // the safe struct destructor.
+        auto *modified_create_info = static_cast<safe_VkDeviceCreateInfo *>(modified_ci);
+
+        bool found_ext = false;
+        for (uint32_t i = 0; i < modified_create_info->enabledExtensionCount; i++) {
+            if (dcm_ext == modified_create_info->ppEnabledExtensionNames[i]) {
+                found_ext = true;
+                break;
+            }
+        }
+        if (!found_ext) {
+            LogInfo(gpu, "UNASSIGNED-DEBUG-PRINTF", "VK_AMD_device_coherent_memory extension not enabled but use_uncached_buffer is true. Forcing extension.");
+            const char **ext_names = new const char *[modified_create_info->enabledExtensionCount + 1];
+            // Copy the existing pointer table
+            std::copy(modified_create_info->ppEnabledExtensionNames,
+                      modified_create_info->ppEnabledExtensionNames + modified_create_info->enabledExtensionCount, ext_names);
+            // Add our new extension
+            char *dcm_ext_copy = new char[dcm_ext.size() + 1]{};
+            dcm_ext.copy(dcm_ext_copy, dcm_ext.size());
+            dcm_ext_copy[dcm_ext.size()] = '\0';
+            ext_names[modified_create_info->enabledExtensionCount] = dcm_ext_copy;
+            // Patch up the safe struct
+            delete[] modified_create_info->ppEnabledExtensionNames;
+            modified_create_info->ppEnabledExtensionNames = ext_names;
+            modified_create_info->enabledExtensionCount++;
+        }
+        auto *dcm_features = const_cast<VkPhysicalDeviceCoherentMemoryFeaturesAMD *>(
+            LvlFindInChain<VkPhysicalDeviceCoherentMemoryFeaturesAMD>(modified_create_info));
+        if (dcm_features) {
+            if (dcm_features->deviceCoherentMemory != VK_TRUE) {
+                LogInfo(gpu, "UNASSIGNED-DEBUG-PRINTF",
+                        "use_uncached_buffer is true, but deviceCoherentMemory feature is not enabled. Force enabling feature.");
+                dcm_features->deviceCoherentMemory = VK_TRUE;
+            }
+        } else {
+            LogInfo(gpu, "UNASSIGNED-DEBUG-PRINTF",
+                    "use_uncached_buffer is true, but deviceCoherentMemory feature is not enabled. Force enabling feature.");
+            auto new_dcm_features = LvlInitStruct<VkPhysicalDeviceCoherentMemoryFeaturesAMD>();
+            new_dcm_features.deviceCoherentMemory = VK_TRUE;
+            new_dcm_features.pNext = const_cast<void *>(modified_create_info->pNext);
+            modified_create_info->pNext = new VkPhysicalDeviceCoherentMemoryFeaturesAMD(new_dcm_features);
+        }
+    }
+}
+
 // Perform initializations that can be done at Create Device time.
 void DebugPrintf::CreateDevice(const VkDeviceCreateInfo *pCreateInfo) {
     if (enabled[gpu_validation]) {
@@ -42,7 +123,16 @@ void DebugPrintf::CreateDevice(const VkDeviceCreateInfo *pCreateInfo) {
     use_stdout = !stdout_string.compare("true");
     if (getenv("DEBUG_PRINTF_TO_STDOUT")) use_stdout = true;
 
-    // GpuAssistedBase::CreateDevice will set up bindings
+    // Need to get this option again, because PreCallRecordCreateDevice was done
+    // in separate DebugPrintf instance (during VkInstance creation).
+    std::string use_uncached_buffer_string = getLayerOption("khronos_validation.printf_uncached_buffer");
+    vvl::ToLower(use_uncached_buffer_string);
+    use_uncached_buffer = !use_uncached_buffer_string.compare("true");
+    if (use_uncached_buffer) {
+        force_device_coherent_memory = true; // vma needs to know it.
+    }
+
+    // GpuAssistedBase::CreateDevice will set up bindings.
     VkDescriptorSetLayoutBinding binding = {3, VK_DESCRIPTOR_TYPE_STORAGE_BUFFER, 1,
                                             VK_SHADER_STAGE_ALL_GRAPHICS | VK_SHADER_STAGE_MESH_BIT_EXT |
                                                 VK_SHADER_STAGE_TASK_BIT_EXT | VK_SHADER_STAGE_COMPUTE_BIT |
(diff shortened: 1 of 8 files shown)


layers/gpu_validation/debug_printf.cpp
```diff
@@ -680,11 +819,16 @@ void DebugPrintf::AllocateDebugPrintfResources(const VkCommandBuffer cmd_buffer,
...
+    result = vmaMapMemory(vmaAllocator, output_block.allocation, reinterpret_cast<void **>(&output_block.data));
     if (result == VK_SUCCESS) {
-        memset(data, 0, output_buffer_size);
-        vmaUnmapMemory(vmaAllocator, output_block.allocation);
+        memset(output_block.data, 0, output_buffer_size);
+        if (!use_uncached_buffer) {
+            vmaUnmapMemory(vmaAllocator, output_block.allocation);
+            output_block.data = nullptr;
+        } else {
+            // Mapping may fail after DEVICE_LOST. Keep it mapped for now.
```
jeremyg-lunarg, line 829: I'd prefer just the comment above the if block rather than this empty else block.

layers/gpu_validation/debug_printf.cpp
```diff
@@ -138,6 +228,42 @@ void DebugPrintf::PreCallRecordCreateShaderModule(VkDevice device, const VkShade
...
 
+// Override GpuAssistedBase version to allow processing in case of VK_ERROR_DEVICE_LOST when using uncached buffer.
+void DebugPrintf::PostCallRecordQueueSubmit(VkQueue queue, uint32_t submitCount, const VkSubmitInfo *pSubmits, VkFence fence,
+                                            VkResult result) {
+    ValidationStateTracker::PostCallRecordQueueSubmit(queue, submitCount, pSubmits, fence, result);
+
+    bool device_lost = (result == VK_ERROR_DEVICE_LOST);
+
+    if (aborted) return;
+    if (!((result == VK_SUCCESS) || (device_lost && use_uncached_buffer))) return;
```
jeremyg-lunarg, line 239: Should this be `...(device_lost && !use_uncached_bufffer))) return;`

It seems to me that if we're using uncached buffers to get data after a device lost we'd want to process them here. Or am I missing something?

Tony-LunarG (commented):
Has this been tested?  Is there a test that can go with this?

juan-lunarg (commented):
I don't see any testing being added. Is the plan to add them later?

layers/gpu_validation/debug_printf.cpp
dorian-apanel-intel: Maybe you missed that negation at start `!`
I will try to write it cleaner.

dorian-apanel-intel:
@juan-lunarg @Tony-LunarG 
> I don't see any testing being added. Is the plan to add them later?
> Has this been tested? Is there a test that can go with this?

I've only tested it manually so far.
Would it be enough to copy one test from `tests\negative\debug_printf.cpp` and modify it to use uncached buffer and hang?

layers/gpu_validation/debug_printf.cpp
dorian-apanel-intel: Done

layers/gpu_validation/debug_printf.cpp
jeremyg-lunarg: yeah I did miss that.

Tony-LunarG:
> @juan-lunarg @Tony-LunarG
> 
> > I don't see any testing being added. Is the plan to add them later?
> > Has this been tested? Is there a test that can go with this?
> 
> I've only tested it manually so far. Would it be enough to copy one test from `tests\negative\debug_printf.cpp` and modify it to use uncached buffer and hang?

I think so, as long as the test gets a DEVICE_LOST and then successfully parses a debug printf

juan-lunarg:
>  Would it be enough to copy one test from tests\negative\debug_printf.cpp and modify it to use uncached buffer and hang?

This sounds like it would cause problems for Jenkins CI (Although if we switched to ctest maybe not... since the test would be in it's own process).

Because currently CI invokes the test executable and tries running all tests.

dorian-apanel-intel:
@Tony-LunarG @juan-lunarg @jeremyg-lunarg
* I've added unit test, had to pass a value using push constants, because hardcoded infinite loop was not hanging on AMD.
* I modified layer_options.cpp file to move option from test to layer. It is not perfect, but after looking at #5895, I think it will need rewrite soon anyway.
* After updating AMD drivers, Atomics WA is no longer needed there, but leaving it for Intel for now.
* NVidia does not support required extension yet.
* Force push was required due to printf test file rename.

I plan to add also unit test for Page Fault, but preferably in another pull request.

juan-lunarg:
Galaxy S22 testing failed

```
[ RUN      ] NegativeDebugPrintf.UncachedBuffer
jni/../../tests/framework/error_monitor.cpp:200: Failure
Failed
Validation Error: [ UNASSIGNED-DEBUG-PRINTF ] Object 0: handle = 0x6c1d064220, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x92394c89 | Setup Error. Detail: (Unable to allocate device memory.  Device could become unstable.)
jni/../../tests/framework/error_monitor.cpp:241: Failure
Failed
Did not receive expected error 'Here are three float values 1.000000, 3.141500, 0.000000'
[  FAILED  ] NegativeDebugPrintf.UncachedBuffer (5836 ms)
```

juan-lunarg (commented):
LGTM. Thanks for the test 👍🏾 

I'll defer to @Tony-LunarG and @jeremyg-lunarg for the actual review.

Tony-LunarG:
> Galaxy S22 testing failed
> 
> ```
> [ RUN      ] NegativeDebugPrintf.UncachedBuffer
> jni/../../tests/framework/error_monitor.cpp:200: Failure
> Failed
> Validation Error: [ UNASSIGNED-DEBUG-PRINTF ] Object 0: handle = 0x6c1d064220, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x92394c89 | Setup Error. Detail: (Unable to allocate device memory.  Device could become unstable.)
> jni/../../tests/framework/error_monitor.cpp:241: Failure
> Failed
> Did not receive expected error 'Here are three float values 1.000000, 3.141500, 0.000000'
> [  FAILED  ] NegativeDebugPrintf.UncachedBuffer (5836 ms)
> ```

I don't think there's much the test can do about that.  I'll just add the new test to the S22 blacklist.

layers/gpu_validation/debug_printf.cpp
```diff
@@ -308,10 +441,14 @@ void DebugPrintf::AnalyzeAndGenerateMessages(VkCommandBuffer command_buffer, VkQ
...
-    if (!expect) return;
+    // Total size of all messages are written by AtomicAdd. Atomics in uncached memory seems to be working in caches anyway
+    // and are not flushed to uncached memory at the end. In that case, expect will contain zero.
+    // As a WA just parse messages using individual sizes (written correctly).
+    // Can the messages be overridden because of those atomics?
+    if (!expect && !use_uncached_buffer) return;
 
     uint32_t index = spvtools::kDebugOutputDataOffset;
-    while (debug_output_buffer[index]) {
+    while ((index < output_buffer_size) && debug_output_buffer[index]) {
```
Tony-LunarG, line 451: I don't think this loop will work if the output buffer has one message but isn't full.  To see this, use uncached buffer memory in the test NegativeDebugPrintf.BasicUsage and run it. This loop doesn't stop when it's supposed to.

Tony-LunarG (changes requested):
The message processing loop won't exit correctly

layers/gpu_validation/debug_printf.cpp
dorian-apanel-intel: Could you check in which vkQueueSubmit/AnalyzeAndGenerateMessages you observe it? BasicUsage does it 15 times.

Here is how I understand it:

In my first one, `debug_output_buffer == 0x0000028b8e520000`
Memory of that buffer is:

```
0x0000028B8E520000  00 00 00 00  MBZ
0x0000028B8E520004  0a 00 00 00  expect (size of all messages, written by atomic in shader)
0x0000028B8E520008  0a 00 00 00  DPFOutputRecord.size (size in dwords of this message)
0x0000028B8E52000C  00 00 00 00  DPFOutputRecord.shader_id
0x0000028B8E520010  70 00 00 00  DPFOutputRecord.instruction_position
0x0000028B8E520014  00 00 00 00  DPFOutputRecord.stage
0x0000028B8E520018  00 00 00 00  DPFOutputRecord.stage_word_1
0x0000028B8E52001C  00 00 00 00  DPFOutputRecord.stage_word_2
0x0000028B8E520020  00 00 00 00  DPFOutputRecord.stage_word_3
0x0000028B8E520024  29 00 00 00  DPFOutputRecord.format_string_id ("Here are two float values %f, %f")
0x0000028B8E520028  00 00 80 3f  DPFOutputRecord.values (float 1.0)
0x0000028B8E52002C  56 0e 49 40  DPFOutputRecord.values (float 3.14150000)
0x0000028B8E520030  00 00 00 00  DPFOutputRecord.size  (no more messages)
0x0000028B8E520034  00 00 00 00  DPFOutputRecord.
0x0000028B8E520038  00 00 00 00  DPFOutputRecord.
```

Index starts at 0x2. At the end of first loop iteration:
`index += debug_record->size; // 0x2 + 0xa = 0xc`
At second loop beginning 
`debug_output_buffer[index]` == 0x0000028b8e520030 {0x0}
So the loop ends.

`(index < output_buffer_size)` means "while we are still in the buffer"
`&& debug_output_buffer[index])` means "while next message is not empty"

Tony-LunarG: I don't have access to the system I was using to test this today - I'll get back to it tomorrow.  The system I do have is getting VK_TIMEOUT instead of VK_ERROR_DEVICE_LOST, and I had to add that to the test's expected results. You should add that to your test.

Tony-LunarG: I see the problem.  Because debug_output_buffer[1] comes back as 0, the memset at the end of message processing doesn't clear the buffer:
```
memset(debug_output_buffer, 0, 4 * (debug_output_buffer[spvtools::kDebugOutputSizeOffset] + spvtools::kDebugOutputDataOffset));
```
And the memset clears 4 * (0 + 2) bytes which doesn't cover all of the records.  Then on the next printf, the loop depends on ```debug_output_buffer[index]``` being zero when the loop is supposed to stop, but previous printfs may have left values that didn't get memset to 0.  I suppose one answer would be to memset the whole buffer.

But I'm seeing another problem.  If I modify your test to do two debug printfs:
```
          if (gl_VertexIndex == 0) {
                debugPrintfEXT("Here are three float values %f, %f, %f", 1.0, myfloat, gl_Position.x);
                debugPrintfEXT("Here's another debug printf");
                float x = constants.x[0];
                while(x > -1.f) { // infinite loop
                    x += 0.000001f;
                }
                debugPrintfEXT("Here is a value that should not be printed %f", x);
            }
```
I only get the second printf back, as if the first one was overwritten in the shader.  Could that be a result of the atomic operation problem and debug_output_buffer[1] not getting written?
(This is all on an AMD RX6600 with driver 23.7.1)

dorian-apanel-intel: Thanks for finding that memset issue, I've added whole buffer clear in case of uncached buffer.

The only second printf working I cannot reproduce on AMD RX 5700 XT with the same driver you have. Maybe I will ask AMD devs for feedback.

Tony-LunarG: My working theory was that the shader instrumentation was using data_buffer[1] as an offset to write the next debug printf result, so if it's zero, all printfs would overwrite each other.  But it sounds like you've tested a case where [1] is not written, but you still get 2 printf results?

lunarpapillo:
Last LunarG CI failure was due to a system issue.  I've restarted.

layers/gpu_validation/debug_printf.cpp
```diff
@@ -672,6 +821,9 @@ void DebugPrintf::AllocateDebugPrintfResources(const VkCommandBuffer cmd_buffer,
     buffer_info.usage = VK_BUFFER_USAGE_STORAGE_BUFFER_BIT;
     VmaAllocationCreateInfo alloc_info = {};
     alloc_info.requiredFlags = VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT;
+    if (use_uncached_buffer) {
+        alloc_info.requiredFlags |= VK_MEMORY_PROPERTY_DEVICE_COHERENT_BIT_AMD | VK_MEMORY_PROPERTY_DEVICE_UNCACHED_BIT_AMD;
```
Tobski, line 825: You probably don't want to require UNCACHED here, since _theoretically_ in future you might get COHERENT _and_ CACHED memory. Uncached is meant to be informative, not something you generally rely upon unless you're doing some very specific performance tuning.

layers/gpu_validation/debug_printf.cpp
Tobski: How is the problematic AtomicAdd performed? I don't know where to look for the code to try to figure this out.

Tobski: I may have uncovered some driver weirdness here so I've started a discussion internally to see whether we've messed something up here, I'll let you know if we confirm a bug.

Tony-LunarG: The atomic add is performed by the instrumentation added to the shader via spirv-tools code, which I'm not very familiar with.

jeremyg-lunarg: https://github.com/KhronosGroup/SPIRV-Tools/blob/main/source/opt/instrument_pass.cpp#L823

Tobski: Ok so after a bit of digging, this extension was enabled before we had the MALL cache in hardware, and nobody ever went back and changed the interface to bypass MALL, so there IS caching enabled on this memory type for newer hardware. Whoops. I've filed an internal bug to fix this and to add a more comprehensive test to make sure this actually works on the off chance we make another big change to the caching hierarchy...

You can go look at the public driver source if for some reason you want to see where this is happening 🙃

I'll let you know when it's fixed and subsequently released.

dorian-apanel-intel: @Tobski I assume the fix is released now?

layers/gpu_validation/debug_printf.cpp
dorian-apanel-intel: I don't quite understand. Here I want to get device uncached to read it after device is lost.

layers/gpu_validation/debug_printf.cpp
Tobski: A fix has been implemented but I'm not sure on release status, I'll take a look.

spencer-lunarg:
this has got stale - closing, but tracking here (https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9211) so can reference this in future

PR #6130, opened 2023-07-12
layers: Move VulkanMemoryAllocator to known_good.json

juan-lunarg:
Note the point of this draft PR is to show what it looks like to add vulkan memory allocator to known_good.json

There are currently 2 main issues.

The android ndk-build and the chromium build.

The ndk-build will soon be removed. But I don't know how to address the chromium build (Since I don't really understand it)


juan-lunarg:
> But I don't know how to address the chromium build (Since I don't really understand it)

Pinging @mikes-lunarg who understand the chromium build.

Talking to @mikes-lunarg they mentioned that VMA would have to be added to https://chromium.googlesource.com/vulkan-deps/

I don't really understand what this vulkan-deps repo is. Also I'm not sure this address dependencies like `xxhash`. Which is in a similar situation to VulkanMemoryAllocator.

mikes-lunarg:
I did a little more digging and both chromium and ANGLE already have VMA in their DEPS file, but unfortunately it is an extremely old fork: https://chromium.googlesource.com/external/github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/

juan-lunarg:
Who would we ask to update this?

What about other dependencies like `xxhash`?

Tony-LunarG:
Also, GPU-AV performance has heavily depended on VMA performance.  Will the updated known good be GPU-AV performance tested?

juan-lunarg:
> Also, GPU-AV performance has heavily depended on VMA performance. Will the updated known good be GPU-AV performance tested?

Yes it would be. I've closed this PR though since I'm not planning on updating the version of VMA yet.

Issue #6122, opened 2023-07-10
broken handling for nullptr IMAGE_STATE in PostCallRecordCreateImageView


zmike:
Not exactly sure what to file this as, but `ValidationStateTracker::PostCallRecordCreateImageView` seems broken if `auto image_state = Get<IMAGE_STATE>(pCreateInfo->image);` returns a nullptr. I'm hitting this when trying to use VVL with zink on steam, and I'm not entirely sure what to do about it.

spencer-lunarg:
1. Did you have `VkImageViewCreateInfo::image` set to null by accident?
2. Did you not hit ` VUID-VkImageViewCreateInfo-image-parameter` first in stateless validation?

zmike:
1. No
2. There are no errors printed

spencer-lunarg:
is this a reproducible issue or sometimes happens?

zmike:
100% when running steam on zink

spencer-lunarg:
@zmike still not sure where this would have gone wrong, but even if we go

```patch
   auto image_state = Get<IMAGE_STATE>(pCreateInfo->image);
+   if (!image_state) return;
```

it will probably just fail later as we won't have the ImageView state

@zmike  Is there a no-zero chance the image and imageview have different `VkDevice`?

spencer-lunarg:
as always @zmike, feel free to reopen if this doesn't solve the issue

gents83:
@spencer-lunarg & @zmike I'm getting the exact same error while trying to use NVIDIA NSight Graphics to do a frame debug on my test app in wgpu (on Vulkan), but it's not really clear to me reading here how to solve the issue :)

spencer-lunarg:
@gents83 to help understand, as of the latest spec (1.3.260) there is now a dedicated VU

> VUID-vkCreateImageView-image-09179
> VkImageViewCreateInfo::image must have been created from device

as the working group agreed this is not valid Vulkan that works everywhere.

This was not being caught before the change in #6132 was added 2 weeks ago, so anything that was violating this, would most likely crash (but now we will not crash **at Image View creation time** and return the error message)

So I am curious if the issue is only when you enabled NVIDIA NSight Graphics? Is your WGPU app doing anything that would involve multiple Device instances (multi-GPU)?

gents83:
Tnx a lot for the explanation 🤗
Nope, it only happens with nsight graphics, but not when launching the app alone

PR #6132, opened 2023-07-13
Add check for Image and ImageView having different Device

spencer-lunarg:
Should close https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6122

@zmike I now 99.9% think your issue is the fact the `ImageView` and `Image` have different `VkDevice` handles (my test failed in the same spot you did)

I will need to figure out in https://gitlab.khronos.org/vulkan/vulkan/-/issues/3553 why it never generated the VU for this and inturn prevented the code from generating the check for it


tests/framework/layer_validation_tests.h
```diff
@@ -534,6 +534,12 @@ class CooperativeMatrixTest : public VkLayerTest {};
 class NegativeShaderCooperativeMatrix : public CooperativeMatrixTest {};
 class PositiveShaderCooperativeMatrix : public CooperativeMatrixTest {};
 
+class MultiDeviceTest : public VkLayerTest {
+  public:
+    ~MultiDeviceTest();
+    VkDeviceObj *m_second_device = nullptr;
```
artem-lunarg, line 540: nit: suggestion to consider `unique_ptr`

spencer-lunarg: this is what `VkRenderFramework` does... and there is a LOT of these raw pointers everywhere... let me do a followup PR to try and clean them all at once (to make sure CI is happy with it)

artem-lunarg (approved):

Issue #6134, opened 2023-07-13
VUID-VkImageCreateInfo-imageCreateMaxMipLevels-02251 could be more detailed


zmike:
If I get this error, e.g.,

```
Validation Error: [ VUID-VkImageCreateInfo-imageCreateMaxMipLevels-02251 ] Object 0: handle = 0x28df55bb2000, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0xbebcae79 | vkCreateImage(): Format VK_FORMAT_B8G8R8A8_UNORM is not supported for this combination of parameters and VkGetPhysicalDeviceImageFormatProperties returned back VK_ERROR_FORMAT_NOT_SUPPORTED. The Vulkan spec states: Each of the following values (as described in Image Creation Limits) must not be undefined : imageCreateMaxMipLevels, imageCreateMaxArrayLayers, imageCreateMaxExtent, and imageCreateSampleCounts (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkImageCreateInfo-imageCreateMaxMipLevels-02251)
```

it would be nice if VVL could be more detailed if it knows which params are broken, or in the case where it isn't, just print all the params.

spencer-lunarg:
agree... this VU (because of DMR format modifier and Android External formats) is easily one of the nastiest VUs... I agree the message is less than desirable

spencer-lunarg:
actually looking more into this, the whole way we are checking this seems strange and lacks any real test, thanks for bringing this up, happy (kinda :upside_down_face: ) to get this quickly fixed

PR #6135, opened 2023-07-13
Improve message for 02251

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6134

If you take a deep look you find `02251` VU indeed is just a VERY complex way of saying `vkGetPhysicalDeviceImageFormatProperties` must be `VK_SUCCESS`

This improves it and adds a test (didn't have one because its hard to guarantee the result to be non-success)

> Validation Error: [ VUID-VkImageCreateInfo-imageCreateMaxMipLevels-02251 ] Object 0: handle = 0x56229a852780, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0xbebcae79 | vkCreateImage(): The following parameters -
format (VK_FORMAT_E5B9G9R9_UFLOAT_PACK32)
type (VK_IMAGE_TYPE_2D)
tiling (VK_IMAGE_TILING_OPTIMAL)
usage (VK_IMAGE_USAGE_STORAGE_BIT)
flags (VkImageCreateFlags(0))
returned a non-success result (VK_ERROR_FORMAT_NOT_SUPPORTED) when calling VkGetPhysicalDeviceImageFormatProperties (which makes all things undefined). The Vulkan spec states: Each of the following values (as described in Image Creation Limits) must not be undefined : imageCreateMaxMipLevels, imageCreateMaxArrayLayers, imageCreateMaxExtent, and imageCreateSampleCounts (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkImageCreateInfo-imageCreateMaxMipLevels-02251)`

### DIFF
--- a/layers/core_checks/cc_image.cpp
+++ b/layers/core_checks/cc_image.cpp
@@ -218,9 +218,17 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
         if (!LvlFindInChain<VkExternalFormatANDROID>(pCreateInfo->pNext)) {
 #endif  // VK_USE_PLATFORM_ANDROID_KHR
             skip |= LogError(device, "VUID-VkImageCreateInfo-imageCreateMaxMipLevels-02251",
-                             "vkCreateImage(): Format %s is not supported for this combination of parameters and "
-                             "VkGetPhysicalDeviceImageFormatProperties returned back %s.",
-                             string_VkFormat(pCreateInfo->format), string_VkResult(result));
+                             "vkCreateImage(): The following parameters -\n"
+                             "format (%s)\n"
+                             "type (%s)\n"
+                             "tiling (%s)\n"
+                             "usage (%s)\n"
+                             "flags (%s)\n"
+                             "returned a non-success result (%s) when calling VkGetPhysicalDeviceImageFormatProperties (which "
+                             "makes all things undefined).",
+                             string_VkFormat(pCreateInfo->format), string_VkImageType(pCreateInfo->imageType),
+                             string_VkImageTiling(pCreateInfo->tiling), string_VkImageUsageFlags(pCreateInfo->usage).c_str(),
+                             string_VkImageCreateFlags(pCreateInfo->flags).c_str(), string_VkResult(result));
 #ifdef VK_USE_PLATFORM_ANDROID_KHR
         }
 #endif  // VK_USE_PLATFORM_ANDROID_KHR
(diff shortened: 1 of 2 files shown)


layers/core_checks/cc_image.cpp
```diff
@@ -218,9 +218,17 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
...
-                             "VkGetPhysicalDeviceImageFormatProperties returned back %s.",
-                             string_VkFormat(pCreateInfo->format), string_VkResult(result));
+                             "vkCreateImage(): The following parameters -\n"
+                             "format (%s)\n"
+                             "type (%s)\n"
+                             "tiling (%s)\n"
+                             "usage (%s)\n"
+                             "flags (%s)\n"
+                             "returned a non-success result (%s) when calling VkGetPhysicalDeviceImageFormatProperties (which "
+                             "makes all things undefined).",
```
juan-lunarg, line 228: IMO this reads better. Since if you are getting a validation error it's kinda implied the result isn't successful. Also the `which makes all things undefined` isn't really needed IMO. Since the VUID already says that. Making it kinda redundant.

```suggestion
                             "returned (%s) when calling VkGetPhysicalDeviceImageFormatProperties.",
```

juan-lunarg (approved):

layers/core_checks/cc_image.cpp
```diff
@@ -218,9 +218,17 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
...
-                             "vkCreateImage(): Format %s is not supported for this combination of parameters and "
-                             "VkGetPhysicalDeviceImageFormatProperties returned back %s.",
-                             string_VkFormat(pCreateInfo->format), string_VkResult(result));
+                             "vkCreateImage(): The following parameters -\n"
+                             "format (%s)\n"
+                             "type (%s)\n"
+                             "tiling (%s)\n"
+                             "usage (%s)\n"
+                             "flags (%s)\n"
+                             "returned a non-success result (%s) when calling VkGetPhysicalDeviceImageFormatProperties (which "
```
juan-lunarg, line 227: Note also looking at the above code it could be using `VkGetPhysicalDeviceImageFormatProperties2` so this error message could be somewhat misleading.

spencer-lunarg: ok, can track and print which we use actually

layers/core_checks/cc_image.cpp
spencer-lunarg: agree, I guess this VU is just aweful to understand, but happy with your suggestions

juan-lunarg: Yes I also think this VU is awful to read/understand.

PR #6136, opened 2023-07-13
scripts: Fix path to tests for vk_validation_stats.py

spencer-lunarg:
this hopefully (as I see no reason why) will not change again and can be happy with this

### DIFF
--- a/scripts/vk_validation_stats.py
+++ b/scripts/vk_validation_stats.py
@@ -753,7 +753,7 @@ def main(argv):
     layer_source_files.extend(glob.glob(os.path.join(repo_relative('layers/sync/'), '*.cpp')))
     layer_source_files.extend(glob.glob(os.path.join(repo_relative('layers/object_tracker/'), '*.cpp')))
 
-    test_source_files = glob.glob(os.path.join(repo_relative('tests/negative'), '*.cpp'))
+    test_source_files = glob.glob(os.path.join(repo_relative('tests/unit'), '*.cpp'))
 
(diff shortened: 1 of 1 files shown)


scripts/vk_validation_stats.py
```diff
@@ -753,7 +753,7 @@ def main(argv):
     layer_source_files.extend(glob.glob(os.path.join(repo_relative('layers/sync/'), '*.cpp')))
     layer_source_files.extend(glob.glob(os.path.join(repo_relative('layers/object_tracker/'), '*.cpp')))
 
-    test_source_files = glob.glob(os.path.join(repo_relative('tests/negative'), '*.cpp'))
+    test_source_files = glob.glob(os.path.join(repo_relative('tests/unit'), '*.cpp'))
```
juan-lunarg, line 756: Should we add this script to our github actions CI? Seems important.

juan-lunarg: Wait this is already part of our CI process. I guess the script continued to work without issue.
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/c3065a7aed0445a5ae905d373410fef16f2e9a8d/scripts/common_ci.py#L106

spencer-lunarg: it works, just the result showed we had no test coverage, that is only read by humans (Mike normally maps it in an excel sheet for releases)

juan-lunarg (approved):

PR #6138, opened 2023-07-13
sync: Fix semaphore wait timeout

artem-lunarg:
Fixes `vkWaitSemaphores` timeouts while waiting for `vkSignalSemaphore` https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4968#issuecomment-1613006858

Also fixes the same issue but for `vkGetSemaphoreCounterValue` <-> `vkSignalSemaphore` combination.

Potentially fixes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4968

The root cause is that some commands perform blocking operations inside `Record` (ex. vkWaitSemaphores can't return until a signal is processed). `Record` calls themselves are protected by a mutex (except the case of fine grained locking). If the waiting condition of the blocked thread depends on another thread executing `Record` to update some state, this creates a deadlock. During a blocking operation, there's no need to keep the validation object locked (blocked thread can't modify state), so we can pass exclusive ownership to other threads. After the wait is completed, the lock is re-acquired again.

Validation objects that use fine grained locking do not have this issue. That's why repro case is implemented for SyncVal which does not use fine grained locking.

### DIFF
--- a/layers/vulkan/generated/chassis.h
+++ b/layers/vulkan/generated/chassis.h
@@ -3785,6 +3785,33 @@ class ValidationObject {
             return WriteLockGuard(validation_object_mutex);
         }
 
+        // If the Record phase calls a function that blocks, we might need to release
+        // the lock that protects Record itself in order to avoid mutual waiting.
+        WriteLockGuard* record_guard = nullptr;
+
+        // Should be used instead of WriteLock() if the Record phase wants to release
+        // its lock during the blocking operation.
+        void GetWriteLockForBlockingOperation(WriteLockGuard& write_lock) {
+            write_lock = WriteLock();
+            // Initialize record_guard only when Record is actually protected by the
+            // mutex. It's not the case when fine grained locking is enabled.
+            record_guard = write_lock.owns_lock() ? &write_lock : nullptr;
+        }
+
+        // The following Begin/End methods should be called during the Record phase
+        // around blocking operation that causes mutual waiting (deadlock).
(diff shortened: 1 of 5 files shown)


layers/vulkan/generated/chassis.h
```diff
@@ -3785,6 +3785,23 @@ class ValidationObject {
             return WriteLockGuard(validation_object_mutex);
         }
 
+        WriteLockGuard* write_lock_for_blocking_call = nullptr;
+        void GetWriteLockForBlockingCall(WriteLockGuard& write_lock) {
+            write_lock = WriteLock();
+            write_lock_for_blocking_call = write_lock.owns_lock() ? &write_lock : nullptr;
```
jeremyg-lunarg, line 3791: these methods need some comments explaining how they work.  Is the assumption is that if  `write_lock.owns_lock() == false` then there's an attempt at recursive locking?

jeremyg-lunarg: or is it a way to handle fine grained locking being configured off or on?

artem-lunarg: I'll add comments. This solution is needed if fine grained locking is not used. So yes, basically it checks if fine grained locking is disabled or not implemented.

artem-lunarg: Added comments and hopefully better naming.

spencer-lunarg (approved):

PR #6139, opened 2023-07-13
ci: Remove ndk-build from github actions CI

juan-lunarg:
ndk-build will be removed soon.


juan-lunarg:
In the meantime Jenkins will continue to test the ndk-build until it's fully removed.

spencer-lunarg (approved):
:tada:

PR #6146, opened 2023-07-14
android: Require CMake 3.21+ and ndk r24+

juan-lunarg:
Avoids legacy issues with older versions of these tools.

### DIFF
--- a/scripts/android.py
+++ b/scripts/android.py
@@ -156,8 +156,6 @@ def main():
         cmake_cmd += f' -D CMAKE_ANDROID_STL_TYPE={android_stl}'
 
         cmake_cmd += ' -D ANDROID_PLATFORM=26'
-        cmake_cmd += ' -D CMAKE_ANDROID_RTTI=YES'
-        cmake_cmd += ' -D CMAKE_ANDROID_EXCEPTIONS=YES'
         cmake_cmd += ' -D ANDROID_USE_LEGACY_TOOLCHAIN_FILE=NO'
 
         common_ci.RunShellCmd(cmake_cmd)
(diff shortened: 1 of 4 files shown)


lunarpapillo (changes requested):
These changes to `CMakeLists.txt` look great - clear, good commenting, good error messages.

Shouldn't there also be corresponding changes to the **Building On Android** section of `BUILD.md`?

scripts/android.py
```diff
@@ -156,8 +156,6 @@ def main():
         cmake_cmd += f' -D CMAKE_ANDROID_STL_TYPE={android_stl}'
 
         cmake_cmd += ' -D ANDROID_PLATFORM=26'
-        cmake_cmd += ' -D CMAKE_ANDROID_RTTI=YES'
```
lunarpapillo, line 159: Just for my illumination... do you have any insights into why these two lines were there in the first place?

(Don't research it if you don't "just know" the answer - it's not that important.)

juan-lunarg:
> Shouldn't there also be corresponding changes to the **Building On Android** section of `BUILD.md`?

There will be. Currently BUILD.md still uses instructions for the ndk-build because that is what CI is using.

scripts/android.py
juan-lunarg: CMAKE_ANDROID_RTTI was never needed as far as I can tell. I added it since it was in the CMake book for the sake of verbosity.

`CMAKE_ANDROID_EXCEPTIONS`  is buggy in the r23 NDK release. But now that we explicitly require a newer NDK we don't have to explicitly set it anymore.

spencer-lunarg (approved):

PR #6147, opened 2023-07-15
layers: Move cc_shader.h to proper util

spencer-lunarg:
currently we have things from `gpu_av`(and BP will to in future) trying to access code in `core_checks/cc_shader.cpp` which is just structurally wrong

Created a proper util file to hold the various shader related things not specific to core checks

### DIFF
--- a/layers/CMakeLists.txt
+++ b/layers/CMakeLists.txt
@@ -231,7 +231,6 @@ target_sources(vvl PRIVATE
     core_checks/cc_ray_tracing.cpp
     core_checks/cc_render_pass.cpp
     core_checks/cc_shader.cpp
-    core_checks/cc_shader.h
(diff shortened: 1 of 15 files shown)


layers/CMakeLists.txt
```diff
@@ -341,6 +340,8 @@ target_sources(vvl PRIVATE
     sync/sync_vuid_maps.h
     thread_tracker/thread_safety_validation.cpp
     thread_tracker/thread_safety_validation.h
+    utils/shader_utils.cpp
+    utils/shader_utils.h
```
spencer-lunarg, line 344: @juan-lunarg kept these here, while this is a util, rather not require `VkLayer_utils` to bring in all the SPIR-V libs

artem-lunarg (approved):

PR #6150, opened 2023-07-17
build: Redo API parameterization

aqnuep:
This re-does the API-specific parameterization needed to support downstream API variants like Vulkan SC after the generator script rewrites.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,10 @@ cmake_minimum_required(VERSION 3.17.2)
 
 project(VVL LANGUAGES CXX)
 
+# This variable enables downstream users to customize the target API
+# variant (e.g. Vulkan SC)
+set(API_TYPE "vulkan")
+
 add_subdirectory(scripts)
 
 set(VVL_CPP_STANDARD 17 CACHE STRING "Set the C++ standard to build against.")
--- a/layers/CMakeLists.txt
+++ b/layers/CMakeLists.txt
@@ -16,9 +16,8 @@
 # limitations under the License.
 # ~~~
 
-# These variables enable downstream users to customize the CMake targets
+# This variable enables downstream users to customize the CMake targets
 # based on the target API variant (e.g. Vulkan SC)
-set(API_TYPE "vulkan")
 set(LAYER_NAME "VkLayer_khronos_validation")
 
 add_library(VkLayer_utils STATIC)
--- a/scripts/CMakeLists.txt
+++ b/scripts/CMakeLists.txt
@@ -40,6 +40,8 @@ if (UPDATE_DEPS)
     endif()
     list(APPEND update_dep_command "--config")
     list(APPEND update_dep_command "${_build_type}")
+    list(APPEND update_dep_command "--api")
+    list(APPEND update_dep_command "${API_TYPE}")
 
     set(UPDATE_DEPS_DIR_SUFFIX "${_build_type}")
     if (ANDROID)
--- a/scripts/generate_source.py
+++ b/scripts/generate_source.py
@@ -66,7 +66,7 @@ def RunGenerators(api: str, registry: str, grammar: str, directory: str, targetF
 
     # These set fields that are needed by both OutputGenerator and BaseGenerator,
     # but are uniform and don't need to be set at a per-generated file level
-    from generators.base_generator import (SetOutputDirectory, SetTargetApiName)
+    from generators.base_generator import (SetOutputDirectory, SetTargetApiName, SetMergedApiNames)
     SetOutputDirectory(directory)
     SetTargetApiName(api)
 
@@ -77,146 +77,191 @@ def RunGenerators(api: str, registry: str, grammar: str, directory: str, targetF
     generators = {
         'thread_safety_counter_definitions.h' : {
             'generator' : ThreadSafetyOutputGenerator,
+            'genCombined': True,
         },
         'thread_safety_counter_instances.h' : {
             'generator' : ThreadSafetyOutputGenerator,
+            'genCombined': True,
         },
         'thread_safety_counter_bodies.h' : {
             'generator' : ThreadSafetyOutputGenerator,
+            'genCombined': True,
         },
         'thread_safety_commands.h' : {
             'generator' : ThreadSafetyOutputGenerator,
+            'genCombined': True,
         },
         'thread_safety.cpp' : {
             'generator' : ThreadSafetyOutputGenerator,
+            'genCombined': True,
         },
         'stateless_validation_helper.h' : {
             'generator' : StatelessValidationHelperOutputGenerator,
+            'genCombined': False,
             'options' : [valid_usage_file],
         },
         'stateless_validation_helper.cpp' : {
             'generator' : StatelessValidationHelperOutputGenerator,
+            'genCombined': False,
             'options' : [valid_usage_file],
         },
         'enum_flag_bits.h' : {
             'generator' : EnumFlagBitsOutputGenerator,
+            'genCombined': False,
         },
         'valid_enum_values.h' : {
             'generator' : ValidEnumValuesOutputGenerator,
+            'genCombined': False,
         },
         'valid_enum_values.cpp' : {
             'generator' : ValidEnumValuesOutputGenerator,
+            'genCombined': False,
         },
         'object_tracker.h' : {
             'generator' : ObjectTrackerOutputGenerator,
+            'genCombined': True,
             'options' : [valid_usage_file],
         },
         'object_tracker.cpp' : {
             'generator' : ObjectTrackerOutputGenerator,
+            'genCombined': True,
             'options' : [valid_usage_file],
         },
         'vk_dispatch_table_helper.h' : {
             'generator' : DispatchTableHelperOutputGenerator,
+            'genCombined': True,
         },
         'vk_function_pointers.h' : {
             'generator' : FunctionPointersOutputGenerator,
+            'genCombined': True,
         },
         'vk_function_pointers.cpp' : {
             'generator' : FunctionPointersOutputGenerator,
+            'genCombined': True,
         },
         'vk_layer_dispatch_table.h' : {
             'generator' : LayerDispatchTableOutputGenerator,
+            'genCombined': True,
         },
         'vk_enum_string_helper.h' : {
             'generator' : EnumStringHelperOutputGenerator,
+            'genCombined': True,
         },
         'vk_safe_struct.h' : {
             'generator' : SafeStructOutputGenerator,
+            'genCombined': True,
         },
         'vk_safe_struct_utils.cpp' : {
             'generator' : SafeStructOutputGenerator,
+            'genCombined': True,
         },
         'vk_safe_struct_core.cpp' : {
             'generator' : SafeStructOutputGenerator,
+            'genCombined': True,
         },
         'vk_safe_struct_khr.cpp' : {
             'generator' : SafeStructOutputGenerator,
+            'genCombined': True,
         },
         'vk_safe_struct_ext.cpp' : {
             'generator' : SafeStructOutputGenerator,
+            'genCombined': True,
         },
         'vk_safe_struct_vendor.cpp' : {
             'generator' : SafeStructOutputGenerator,
+            'genCombined': True,
         },
         'vk_object_types.h' : {
             'generator' : ObjectTypesOutputGenerator,
+            'genCombined': True,
         },
         'vk_extension_helper.h' : {
             'generator' : ExtensionHelperOutputGenerator,
+            'genCombined': True,
         },
         'vk_typemap_helper.h' : {
             'generator' : TypemapHelperOutputGenerator,
+            'genCombined': True,
         },
         'chassis.h' : {
             'generator' : LayerChassisOutputGenerator,
+            'genCombined': True,
         },
         'chassis.cpp' : {
             'generator' : LayerChassisOutputGenerator,
+            'genCombined': False,
         },
         'chassis_dispatch_helper.h' : {
             'generator' : LayerChassisOutputGenerator,
+            'genCombined': True,
         },
         'layer_chassis_dispatch.h' : {
             'generator' : LayerChassisDispatchOutputGenerator,
+            'genCombined': True,
         },
         'layer_chassis_dispatch.cpp' : {
             'generator' : LayerChassisDispatchOutputGenerator,
+            'genCombined': True,
         },
         'best_practices.h' : {
             'generator' : BestPracticesOutputGenerator,
+            'genCombined': True,
         },
         'best_practices.cpp' : {
             'generator' : BestPracticesOutputGenerator,
+            'genCombined': True,
         },
         'sync_validation_types.h' : {
             'generator' : SyncValidationOutputGenerator,
+            'genCombined': False,
         },
         'sync_validation_types.cpp' : {
             'generator' : SyncValidationOutputGenerator,
+            'genCombined': False,
         },
         'spirv_validation_helper.cpp' : {
             'generator' : SpirvValidationHelperOutputGenerator,
+            'genCombined': False,
         },
         'spirv_grammar_helper.h' : {
             'generator' : SpirvGrammarHelperOutputGenerator,
+            'genCombined': False,
             'options' : [grammar],
         },
         'spirv_grammar_helper.cpp' : {
             'generator' : SpirvGrammarHelperOutputGenerator,
+            'genCombined': False,
             'options' : [grammar],
         },
         'spirv_tools_commit_id.h' : {
+            'genCombined': False,
             'generator' : SpirvToolCommitIdOutputGenerator,
         },
         'command_validation.h' : {
             'generator' : CommandValidationOutputGenerator,
+            'genCombined': True,
             'options' : [valid_usage_file],
         },
         'command_validation.cpp' : {
             'generator' : CommandValidationOutputGenerator,
+            'genCombined': True,
             'options' : [valid_usage_file],
         },
         'dynamic_state_helper.h' : {
             'generator' : DynamicStateOutputGenerator,
+            'genCombined': False,
         },
         'dynamic_state_helper.cpp' : {
             'generator' : DynamicStateOutputGenerator,
+            'genCombined': False,
         },
         'vk_format_utils.h' : {
             'generator' : FormatUtilsOutputGenerator,
+            'genCombined': True,
         },
         'vk_format_utils.cpp' : {
             'generator' : FormatUtilsOutputGenerator,
+            'genCombined': True,
         },
     }
 
@@ -235,6 +280,17 @@ def RunGenerators(api: str, registry: str, grammar: str, directory: str, targetF
         generatorOptions = generators[target]['options'] if 'options' in generators[target] else []
         generator = targetGenerator(*generatorOptions)
 
+        # This code and the 'genCombined' generator metadata is used by downstream
+        # users to generate code with all Vulkan APIs merged into the target API variant
+        # (e.g. Vulkan SC) when needed. The constructed apiList is also used to filter
+        # out non-applicable extensions later below.
+        apiList = [api]
+        if api != 'vulkan' and generators[target]['genCombined']:
+            SetMergedApiNames('vulkan')
+            apiList.append('vulkan')
+        else:
+            SetMergedApiNames(None)
+
         baseOptions = BaseGeneratorOptions(customFileName = target)
 
         # Create the registry object with the specified generator and generator
@@ -244,9 +300,8 @@ def RunGenerators(api: str, registry: str, grammar: str, directory: str, targetF
         # Parse the specified registry XML into an ElementTree object
         tree = ElementTree.parse(registry)
 
-        # Filter out non-Vulkan extensions
-        if api == 'vulkan':
-            [exts.remove(e) for exts in tree.findall('extensions') for e in exts.findall('extension') if (sup := e.get('supported')) is not None and baseOptions.apiname not in sup.split(',')]
+        # Filter out extensions that are not on the API list
+        [exts.remove(e) for exts in tree.findall('extensions') for e in exts.findall('extension') if (sup := e.get('supported')) is not None and all(api not in sup.split(',') for api in apiList)]
 
         # Load the XML tree into the registry object
         reg.loadElementTree(tree)
--- a/scripts/generators/base_generator.py
+++ b/scripts/generators/base_generator.py
@@ -75,6 +75,30 @@ def SetTargetApiName(apiname: str) -> None:
     global globalApiName
     globalApiName = apiname
 
+def SetMergedApiNames(names: str) -> None:
+    global mergedApiNames
+    mergedApiNames = names
+
+
+# This class is a container for any source code, data, or other behavior that is necessary to
+# customize the generator script for a specific target API variant (e.g. Vulkan SC). As such,
+# all of these API-specific interfaces and their use in the generator script are part of the
+# contract between this repository and its downstream users. Changing or removing any of these
+# interfaces or their use in the generator script will have downstream effects and thus
+# should be avoided unless absolutely necessary.
+class APISpecific:
+    # Version object factory method
+    @staticmethod
+    def createApiVersion(targetApiName: str, name: str, number: str) -> Version:
+        match targetApiName:
+
+            # Vulkan specific API version creation
+            case 'vulkan':
+                nameApi = name.replace('VK_', 'VK_API_')
+                nameString = f'"{name}"'
+                return Version(name, nameString, nameApi, number)
+
+
 # This Generator Option is used across all generators.
 # After years of use, it has shown that most the options are unified across each generator (file)
 # as it is easier to modifiy things per-file that need the difference
@@ -88,7 +112,7 @@ class BaseGeneratorOptions(GeneratorOptions):
                 filename = customFileName if customFileName else globalFileName,
                 directory = customDirectory if customDirectory else globalDirectory,
                 apiname = customApiName if customApiName else globalApiName,
-                mergeApiNames = None,
+                mergeApiNames = mergedApiNames,
                 defaultExtensions = customApiName if customApiName else globalApiName,
                 emitExtensions = '.*',
                 emitSpirv = '.*',
@@ -106,6 +130,7 @@ class BaseGenerator(OutputGenerator):
     def __init__(self):
         OutputGenerator.__init__(self, None, None, None)
         self.vk = VulkanObject()
+        self.targetApiName = globalApiName
 
         # reg.py has a `self.featureName` but this is nicer because
         # it will be either the Version or Extension object
@@ -171,6 +196,15 @@ class BaseGenerator(OutputGenerator):
             #  one or more extension and/or core version names
             for required in dict:
                 for commandName in dict[required]:
+                    # Skip commands removed in the target API
+                    # This check is needed because parts of the base generator code bypass the
+                    # dependency resolution logic in the registry tooling and thus the generator
+                    # may attempt to generate code for commands which are not supported in the
+                    # target API variant, thus this check needs to happen even if any specific
+                    # target API variant may not specifically need it
+                    if not commandName in self.vk.commands:
+                        continue
+
                     command = self.vk.commands[commandName]
                     # Make sure list is unique
                     command.extensions.extend([extension] if extension not in command.extensions else [])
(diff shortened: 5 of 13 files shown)


scripts/generate_source.py
```diff
@@ -235,6 +280,19 @@ def RunGenerators(api: str, registry: str, grammar: str, directory: str, targetF
         generatorOptions = generators[target]['options'] if 'options' in generators[target] else []
         generator = targetGenerator(*generatorOptions)
 
+        # IMPORTANT!
```
spencer-lunarg, line 283: instead of just putting "important" and "don't remove" it would be more helpful to the future user to know why

Can we just label these as "for SC"

aqnuep: I'd say this could have a wider impact, as it's a way to support Vulkan API variants in general, and some of these may apply to Vulkan too, going forward (in particular the solutions for the bypassed Vulkan registry dependency tools).

The reason I'm trying to very explicitly mark these from here on to not have to redo this effort a 3rd time.

juan-lunarg: > a way to support Vulkan API variants in general

Is there going to be another Vulkan API variant?

tests/framework/layer_validation_tests.cpp
```diff
@@ -3699,6 +3699,9 @@ void android_main(struct android_app *app) {
 #include <crtdbg.h>
 #endif
 
+// IMPORTANT! DO NOT REMOVE!
```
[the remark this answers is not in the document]
aqnuep: There's a common set of language I used in the comments for the different instances of the `APISpecific` utility class across the generator scripts.

If you could propose a change to those comments that you'd be happy with and that would still convey the key message that they and their use must be kept as is, unless otherwise agreed with downstream users, and shouldn't be merged back to the general parts of the generator script as part of refactoring, as it happened recently, then I'm happy to update the PR with it.

My goal, and why I used such strong words in these places, is to avoid accidental changes to these, as my previous attempt, even though I moved these pieces into their own variables and functions (which I naively thought should be sufficient), was thrown away as part of refactoring.

scripts/generate_source.py
aqnuep: That's not really what I said, but rather that there's nothing inherently Vulkan SC specific to the changes I'm upstreaming, but rather it's a set of injection points where a Vulkan variant like Vulkan SC can add its own specializations.

tests/framework/layer_validation_tests.cpp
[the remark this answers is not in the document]
spencer-lunarg: IMO, so test would be MUCH better at preventing regressions then a comment, as that actually tests the logic is not broken

I guess I would argue ALL code in the VVL is "Important" or else we really should get rid of it (same with `# DO NOT REMOVE THIS CHECK!`... makes it sound like everything else we could just remove)

A simple

```
// Keep for non-Vulkan API (ex. Vulkan SC)
```

is going to be way more useful to the future developer who is not currently reviewing this PR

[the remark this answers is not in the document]
aqnuep: I've updated the comments.

Please merge this change if you're happy with it.

If we can agree offline to adding some sort of tests to prevent regressions, then I'll follow up with a PR for that ASAP.

aqnuep:
It looks like there's some general instability in one of the CI builds, as the same failure seems to appear after my comment-only updates in this PR, as in some of the latest main builds and in other PRs. Could you please rerun that manually, if possible?

aqnuep:
I've made some further minor changes, just to get everything out of the way, and also rebased now that I see that main was red only because of CI instabilities.

I'll also come back next week with a PR with a github actions proposal for testing the stability of the hooks.

scripts/generators/extension_helper_generator.py
```diff
@@ -66,6 +66,89 @@ def exprToCpp(pr: ParseResults, opt = lambda x: x) -> str:
...
+    else if (specified_version < VK_API_VERSION_1_3)
+        return VK_API_VERSION_1_2;
+    else
+        return VK_API_VERSION_1_3;
+}
+'''
+
+    # Returns dictionary of version field names
+    @staticmethod
+    def getVersionFieldNameDict(targetApiName: str) -> dict[str, str]:
```
spencer-lunarg, line 126: I REALLY REALLY wanted to move away from these "random small containers of data everywhere" as it had made maintainability a nightmare before by 1000 small cuts

I didn't like the fact we had this manually here to begin with... I realize this is either a chance for us to

1. Make things more complex again
2. Fix this properly at the source

I opt 10000% to do `2` and created a PR that should remove the need for these methods in this file (since `self.vk.versions` should be empty for SC)

https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6161

scripts/generators/extension_helper_generator.py
```diff
@@ -66,6 +66,89 @@ def exprToCpp(pr: ParseResults, opt = lambda x: x) -> str:
...
+class APISpecific:
+    # Generates source code for APIVersion class and related utilities
+    @staticmethod
+    def genAPIVersionSource(targetApiName: str) -> str:
+        match targetApiName:
+
+            # Vulkan specific APIVersion class and related utilities
+            case 'vulkan':
+                return '''
+class APIVersion {
```
spencer-lunarg, line 84: I think this class should not be generated, its should be pulled out into its own non-generated header things can include

scripts/generators/layer_chassis_generator.py
```diff
@@ -25,6 +25,165 @@
...
+            ]
+
+
+    # Generates source code for InitObjectDispatchVector
+    @staticmethod
+    def genInitObjectDispatchVectorSource(targetApiName: str) -> str:
+        match targetApiName:
+
+            # Vulkan specific InitObjectDispatchVector
+            case 'vulkan':
```
spencer-lunarg, line 120: instead of having to do case statement on what is basically a C function written in a python string, would it be easier to have `void ValidationObject::InitObjectDispatchVectors() {` just call into a function that then could be `ifdef` out for SC?

scripts/generators/object_tracker_generator.py
```diff
@@ -23,6 +23,54 @@
...
+class APISpecific:
+    # Returns VUIDs to report when detecting undestroyed objects
+    @staticmethod
+    def getUndestroyedObjectVUID(targetApiName: str, scope: str) -> str:
+        match targetApiName:
+
+            # Vulkan specific undestroyed object VUIDs
+            case 'vulkan':
+                per_scope = {
+                    'instance': 'VUID-vkDestroyInstance-instance-00629',
```
spencer-lunarg, line 41: SC has different VUID?!?

spencer-lunarg (changes requested):

scripts/generators/extension_helper_generator.py
spencer-lunarg: or it would be easy to just generate it as it's own file, no reason I see it needs to live in this file, happy to do that if wanted

scripts/generate_source.py
```diff
@@ -235,6 +280,17 @@ def RunGenerators(api: str, registry: str, grammar: str, directory: str, targetF
         generatorOptions = generators[target]['options'] if 'options' in generators[target] else []
         generator = targetGenerator(*generatorOptions)
 
+        # This code and the 'genCombined' generator metadata is used by downstream
+        # users to generate code with all Vulkan APIs merged into the target API variant
+        # (e.g. Vulkan SC) when needed. The constructed apiList is also used to filter
+        # out non-applicable extensions later below.
+        if api != 'vulkan' and generators[target]['genCombined']:
```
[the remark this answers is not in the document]
aqnuep: Not really, if e.g. `api` is `vulkansc` and `genCombined` is `False`, then `apiList` must only contain `vulkansc`.

It's the other way, so maybe we can rewrite it like this:

```python
apiList = api

if api != 'vulkan' and generators[target]['genCombined']:
    SetMergedApiNames('vulkan')
    apiList.append('vulkan)
```

scripts/generators/base_generator.py
```diff
@@ -106,6 +130,7 @@ class BaseGenerator(OutputGenerator):
     def __init__(self):
         OutputGenerator.__init__(self, None, None, None)
         self.vk = VulkanObject()
+        self.targetApiName = globalApiName
```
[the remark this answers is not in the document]
aqnuep: > to be honest, still not 100% why there is a apiname, mergeApiNames and defaultExtensions field in the reg.py generator options

They control completely different things:

  * `apiname` is the target API name to use
  * `mergeApiNames` allows merging other APIs into the target API, as used by Vulkan SC ecosystem components to merge also basic Vulkan API definitions with the SC ones to produce combined headers that enable ecosystem components like the VVL build, which obviously rely on all Vulkan definitions to be there
  * `defaultExtensions` allows specifying a filter for which extensions to include in the current spec/header/etc. build

All, except `mergeApiNames` have been always there in the registry tooling and they have their purpose. I think you'd need to talk with Jon if you don't know what some of the generator parameters mean (although they have documentation in their source).

I also don't see why that would be better. First of all, the code you listed here is about the `GeneratorOptions` class, but we need the target API in the `Generator` class.

What is exactly the problem with having a member variable for the target API? Why is that a problem in the first place?

scripts/generators/base_generator.py
```diff
@@ -316,9 +350,7 @@ def beginFeature(self, interface, emit):
         else: # version
             number = interface.get('number')
             if number != '1.0':
-                nameApi = name.replace('VK_', 'VK_API_')
-                nameString = f'"{name}"'
```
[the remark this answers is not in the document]
aqnuep: Yes, the purpose is to support API variant versions like `VKSC_VERSION_*`.

There is no `VKSC_VERSION_1_1` **yet**. But there's also a big difference between `VKSC_VERSION_1_0` and `VK_VERSION_1_0` in that Vulkan SC 1.0 includes almost all functionality of Vulkan 1.2, so it's not as trivial as you'd think.

Please don't worry about what's needed for SC specifically, that's something I'm taking care of.

scripts/generators/extension_helper_generator.py
aqnuep: This has been there in the generator script, and wasn't factored out. So why now?

Also, over time this class will likely be partially auto-generated. I mean anywhere where a version number is used should really be auto-generated.

Sure, currently the generator scripts mostly hard-coded these, but that has been a bad practice anyway.

scripts/generators/extension_helper_generator.py
aqnuep: No `self.vk.versions` would not be empty for SC. You're making assumptions about SC here, but you really shouldn't.

Vulkan SC 1.0 requires Vulkan 1.0, Vulkan 1.1, and Vulkan 1.2 with some changes.

Now whether these variable names should be manually listed here (or hard-coded at source-level as they were before), is another question, but see my reply above.

scripts/generators/layer_chassis_generator.py
aqnuep: We're not adding `#ifdef`'d code for SC upstream, because that was something LunarG explicitly didn't want.

We're not adding `#ifdef`'d code for SC downstream into files that come from upstream, because it's a guaranteed merge conflict and subject to refactoring breaking it.

What is really the problem for different API variants to generate different implementations of this function?

This was always generated upstream, I just made it so that downstream can customize it.

scripts/generators/object_tracker_generator.py
aqnuep: Yes.

tests/framework/layer_validation_tests.cpp
```diff
@@ -3556,6 +3556,13 @@ void android_main(struct android_app *app) {
 #include <crtdbg.h>
 #endif
 
+// Defining VVL_TESTS_USE_CUSTOM_TEST_FRAMEWORK allows downstream users
+// to inject custom test framework changes. This includes the ability
+// to override the main entry point of the test executable in order to
+// add custom command line arguments and use a custom test environment
+// class. This #ifndef thus makes sure that when the definition is
+// present we do not include the default main entry point.
+#ifndef VVL_TESTS_USE_CUSTOM_TEST_FRAMEWORK
```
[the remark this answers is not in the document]
aqnuep: Why? This is literally less than 10 lines of effective code with no impact on VVL.

The reason for their existence is also explained in detail in the comments.

scripts/generate_source.py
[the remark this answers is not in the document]
spencer-lunarg: ok, happy with this, cleaner to read

scripts/generators/base_generator.py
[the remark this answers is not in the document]
spencer-lunarg: I want to move `base_generator.py` to be something the whole ecosystem uses, I left the `customApiName` only there for you, if you aren't going to use it, the code slowly just gets populated with random variables trying to solve the same problem

scripts/generators/extension_helper_generator.py
spencer-lunarg: before we had the bad habits of merging multiple things into a single generator because the overhead of making a new one was so high (having to do all the `gen*()` calls and such)

I guess this was the logic moves from having 2 strings hardcoded, to just adjusting the `#include` statement

scripts/generators/base_generator.py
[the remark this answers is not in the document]
aqnuep: Why `customApiName` is needed in the first place? It's not custom API name, it's **the** API name, which may vary depending on the target API variant.

aqnuep:
CC @dgkoch.

@spencer-lunarg, in order to try to move this forward, can you please explain what is exactly so objectionable with having the `targetApiName` a member of the generator classes and with having dedicated `APISpecific` classes alongside the common generator code, which I separated out exactly to make it clear what may need to be API-specific vs what is common, as I think that is the number one contention about this PR.

Regarding the API version specific customizations, as we discussed offline, the main reason why those need customization, because the dependencies between versions (`<feature>` tags) is not linear, as it is assumed in some of the suggestions, because Vulkan SC, for example, has `VKSC_VERSION_1_0` depend on `VK_VERSION_1_2`.

tests/framework/layer_validation_tests.cpp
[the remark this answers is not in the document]
ncesario-lunarg: So the usage here is that a "vulkan offshoot" like SC defines `VVL_TESTS_USE_CUSTOM_TEST_FRAMEWORK`, and then adds their own framework file(s) in cmake? If so, I would vote for a cmake change along the lines of the following. I'm guessing @juan-lunarg has better ideas, though.
```diff
diff --git a/tests/CMakeLists.txt b/tests/CMakeLists.txt
index 92166d549..959d97614 100644
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -21,8 +21,6 @@ else()
 endif()
 target_sources(vk_layer_validation_tests PRIVATE
     framework/android_hardware_buffer.h
-    framework/layer_validation_tests.h
-    framework/layer_validation_tests.cpp
     framework/test_common.h
     framework/shader_templates.h
     framework/error_monitor.cpp
@@ -144,6 +142,17 @@ target_sources(vk_layer_validation_tests PRIVATE
     unit/ycbcr_positive.cpp
     containers/small_vector.cpp
 )
+if ("${API_TYPE}" STREQUAL "vulkan")
+    target_sources(vk_layer_validation_tests PRIVATE
+        framework/layer_validation_tests.h
+        framework/layer_validation_tests.cpp
+    )
+elseif("${API_STYLE}" STREQUAL "vulkansc")
+    target_sources(vk_layer_validation_tests PRIVATE
+        framework/layer_validation_tests_sc.h
+        framework/layer_validation_tests_sc.cpp
+    )
+endif()
 get_target_property(TEST_SOURCES vk_layer_validation_tests SOURCES)
 source_group(TREE "${CMAKE_CURRENT_SOURCE_DIR}" FILES ${TEST_SOURCES})
```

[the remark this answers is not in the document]
ncesario-lunarg: I _think_ you can call `target_sources` multiple times for a single target, but not 100% sure.

scripts/generators/base_generator.py
```diff
@@ -75,6 +75,30 @@ def SetTargetApiName(apiname: str) -> None:
...
+# interfaces or their use in the generator script will have downstream effects and thus
+# should be avoided unless absolutely necessary.
+class APISpecific:
+    # Version object factory method
+    @staticmethod
+    def createApiVersion(targetApiName: str, name: str, number: str) -> Version:
+        match targetApiName:
+
+            # Vulkan specific API version creation
+            case 'vulkan':
```
ncesario-lunarg, line 96: How do downstream vulkan variants utilize this? Do they modify this file, adding cases here?

scripts/generators/enum_flag_bits_generator.py
```diff
@@ -21,6 +21,23 @@
 import os
 from generators.base_generator import BaseGenerator
 
+# This class is a container for any source code, data, or other behavior that is necessary to
+# customize the generator script for a specific target API variant (e.g. Vulkan SC). As such,
+# all of these API-specific interfaces and their use in the generator script are part of the
+# contract between this repository and its downstream users. Changing or removing any of these
+# interfaces or their use in the generator script will have downstream effects and thus
+# should be avoided unless absolutely necessary.
+class APISpecific:
```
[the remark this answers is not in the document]
ncesario-lunarg: @aqnuep do you have downstream SC VVL changes that utilize these changes? I'm curious how downstream variants are meant to consume this. If it's modifying _this_ class in _this_ file, that seems problematic. There are a number of possible approaches, like passing in a different _instance_ of this class when the generator is created in `generate_source.py` based on the api type, that I would personally prefer, but I'm probably missing a larger part of the picture here.

@spencer-lunarg
> Also I would like the make the assumption we will only have 1 non-Vulkan API downstream, so it will either be vulkan or not

I would like to make that assumption, too, but I also would have liked to assume there was just Vulkan...but here we are :) Since we have to add the plumbing for SC, might as well make it flexible enough to handle other variants "just in case."

scripts/generate_source.py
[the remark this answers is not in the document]
aqnuep: I've redid this, but it had to look as follows:

```python
        apiList = [api]
        if api != 'vulkan' and generators[target]['genCombined']:
            SetMergedApiNames('vulkan')
            apiList.append('vulkan')
        else:
            SetMergedApiNames(None)
```

The `SetMergedApiNames(None)` call was necessary because I followed the design you established with these global variables, and unless you actually call the function the global won't be defined.

tests/framework/layer_validation_tests.cpp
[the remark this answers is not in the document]
aqnuep: The problem is that `layer_validation_tests.h` and `layer_validation_tests.cpp` contains 100 different, unrelated things (got worse since the recent refactoring with the dozens of new subclasses of `VkLayerTest`).

So replacing the entire header and source would require forking those files and keeping them in sync, which is a maintenance nightmare.

The only thing we need is to be able to change the base class of `VkLayerTest`, which is exactly the change tries to achieve. Nothing more, nothing less.

Of course, if things would be better structured, or if the test framework would support dependency injection or dependency inversion, then things could be done differently too, but that would require very intrusive and time consuming refactoring of the entire test framework of VVL.

In contrast, I tried to propose the minimal sufficient changes in order for us to make things work downstream, which is after all the goal we agreed to.

scripts/generators/extension_helper_generator.py
ncesario-lunarg: Also for context: https://gitlab.khronos.org/vulkan/Vulkan-ValidationLayers/-/merge_requests/91#note_411319

tests/framework/layer_validation_tests.cpp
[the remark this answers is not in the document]
aqnuep: But, to be fair, if these `#ifdef`s are not accepted, I'll just write a script downstream to generate a patched versions of these files.

Regarding the cmake changes, we're not using the upstream cmake file for the tests at all, as we run our own version due to having very different needs on that front.

tests/framework/layer_validation_tests.h
```diff
@@ -41,6 +41,7 @@
 #include <string>
 #include <unordered_set>
 #include <vector>
+#include <condition_variable>
```
ncesario-lunarg, line 44: Used?

scripts/generators/enum_flag_bits_generator.py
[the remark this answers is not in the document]
aqnuep: But then we'd have to run our own copy of `generate_source.py`, and it's the same issue like with `layer_validation_tests.*`.

Also, if you look at the other generator files, it's not always the case that SC needs to **add** something in addition, but rather **replace** something. In fact this is the only exception where it's just about adding something.

This is exactly the reason why the recent refactoring "broke" downstream, as it moved around, inlined, or otherwise removed the places that downstream could inject its own version to.

There simply have to be hooks for which there exists a default Vulkan implementation and downstream users can provide their own custom implementation.

tests/framework/layer_validation_tests.cpp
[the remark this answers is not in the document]
ncesario-lunarg: Fair enough. The portion of code guarded in the ifdef is quite small, and looking at it a bit more I would agree this is probably the best approach without more intrusive changes as you pointed out.

tests/framework/layer_validation_tests.h
aqnuep: Yes, condition variables are used in your upstream code, and actually upstream currently fails to build with some compilers due to the missing include.

tests/framework/layer_validation_tests.cpp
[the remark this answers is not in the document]
juan-lunarg: > But, to be fair, if these #ifdefs are not accepted, I'll just write a script downstream to generate a patched versions of these files.

I'd prefer that solution.

tests/framework/layer_validation_tests.h
juan-lunarg: What compilers? Could we add it to our github actions testing?

CMakeLists.txt
```diff
@@ -18,6 +18,10 @@ cmake_minimum_required(VERSION 3.17.2)
 
 project(VVL LANGUAGES CXX)
 
+# This variable enables downstream users to customize the target API
+# variant (e.g. Vulkan SC)
```
juan-lunarg, line 22: nitpick: make this a 1 line comment

layers/CMakeLists.txt
```diff
@@ -16,9 +16,8 @@
 # limitations under the License.
 # ~~~
 
-# These variables enable downstream users to customize the CMake targets
+# This variable enables downstream users to customize the CMake targets
```
juan-lunarg, line 19: nitpick: make this a 1 line comment

scripts/CMakeLists.txt
```diff
@@ -40,6 +40,8 @@ if (UPDATE_DEPS)
     endif()
     list(APPEND update_dep_command "--config")
     list(APPEND update_dep_command "${_build_type}")
+    list(APPEND update_dep_command "--api")
+    list(APPEND update_dep_command "${API_TYPE}")
```
juan-lunarg, line 44: 👍🏾

tests/framework/layer_validation_tests.cpp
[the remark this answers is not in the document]
aqnuep: > I'd prefer that solution.

Fair enough, and I'm okay with that (probably wouldn't need a script either, as I could just modify the file in-place downstream, conflicts or breakage would still happen at the same rate).

It does make downstream's life more difficult by being more susceptible to conflicting changes, and I'm sure @dgkoch won't be happy about it, but these small pieces are the least of my concern.

tests/framework/layer_validation_tests.h
aqnuep: I wish I would recall (it's been a few months now since I've added this), but I got an error of `std::conditional_variable` not being defined on one of them, as this file uses conditional variables and never included its header. It just so happens that most STL implementations implicitly include a lot of things in many headers, so it doesn't fail everywhere.

scripts/generators/object_tracker_generator.py
dgkoch: (in some cases - we re-use as much as possible).

scripts/generators/enum_flag_bits_generator.py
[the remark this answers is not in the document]
dgkoch: Could it be acceptable if we put some of the `vulkansc` labels and minimal script logic into the upstream repository? 
Then it might be clearer why things are structured the way they are, that might otherwise look silly when there's only one case statement.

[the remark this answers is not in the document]
KarenGhavam-lunarG: > Could it be acceptable if we put some of the vulkansc labels and minimal script logic into the upstream repository?
Then it might be clearer why things are structured the way they are, that might otherwise look silly when there's only one case statement.

Aren't we doing a fork to avoid having SC specific content in VVL?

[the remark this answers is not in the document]
dgkoch: If the compliant is that these changes look pointless on their own, and someone might want to "clean them up" - the best way to give them a point is to make it more obvious what the hooks are for, and having a few Vulkan SC switches or branches shouldn't be an onerous ask. We AREN'T asking to land the whole VKSC VVL code here, just identifiers of places that are important for downstream consumers.

spencer-lunarg:
> @spencer-lunarg, in order to try to move this forward

sorry, spending a night thinking about this, these `APISpecific` right now are fine, what I don't understand is the future process to make any change to our scripts now

Ex. #6161 and #6162  are both valid changes we could want to make (with having to insight in SC's downstream). If we merge this PR, and then i want to make those changes, what is the process to do that? I want to be able to make changes in the future to these scripts, but if it involves needing to break an area where a `APISpecific` function is, how do we go about this?

scripts/generators/base_generator.py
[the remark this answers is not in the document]
spencer-lunarg: > Why customApiName is needed in the first place? 

I thought you might use it. It can be removed later if no one ends up using it

scripts/generators/extension_helper_generator.py
spencer-lunarg: > You're making assumptions about SC here, but you really shouldn't.

sorry, I was unaware the whole concept of a "version" I had is now broken because of SC.

I finally understand what you mean about the `<features>` not being linear, but the fact SC 1.0 is not "core" like Vulkan 1.0 is now just-yet-another-quirk that will need to be accounted for

... it was it is, will just note it now and move on

scripts/generators/object_tracker_generator.py
spencer-lunarg: ok, noted

tests/framework/layer_validation_tests.cpp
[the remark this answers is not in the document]
spencer-lunarg: the reason I wanted a different PR was the tests framework has its own issues compared to the scripts (not a fan how `layer_validation_tests` has become such a garbage bin of stuff, but not the time to clean it up)

spencer-lunarg (approved):
@aqnuep @dgkoch I was unaware by the time constraints you had for SC, I think there is a LOT to be figured out for how we move forward making changes, but these **can** be addressed after

I am going to merge this to allow Rakos life to be easier, but I still intend to figure out post PR to come up with a more long term sustainable way to do this

CMakeLists.txt
spencer-lunarg: can be cleaned up after

aqnuep:
> > @spencer-lunarg, in order to try to move this forward
> 
> sorry, spending a night thinking about this, these `APISpecific` right now are fine, what I don't understand is the future process to make any change to our scripts now
> 
> Ex. #6161 and #6162 are both valid changes we could want to make (with having no insight in SC's downstream). If we merge this PR, and then i want to make those changes, what is the process to do that?
> 
> I'm happy to now know that these area will cause issues for you, but I want to be able to make changes in the future to these scripts, but if it involves needing to break an area where a `APISpecific` function is, how do we go about this?

@spencer-lunarg, if you need to refactor code, like moving the APIVersion class to its own header and generator, that should be fine, as long as you still keep the portions of the code that are marked as API specific under a similar `APISpecific` function.

Sure, it would cause a conflict when we downstream in the next round, but one that's trivially resolvable by finding where that interface moved.

The main problem with the refactoring that happened recently is that it remerged code or strings that were intentionally separated into separate python methods or variables, and that's what we really should be avoiding.

Also, I can imagine another corner case, when the code or data generated by one of the `APISpecific` functions would need to be split out or otherwise refactored. That could be fine too, if the interface would be kept around and call the new sub-pieces.

If none of these options would work I think it's still fine to refactor if given a heads-up about the change.

The important part really would be to not have to reverse engineer where and how things were changed when it comes to these API-specific interfaces and the reason I tried to separate them out this time around so that there's a clean distinction between the common and API-specific pieces.

PR #6154, opened 2023-07-17
docs: Update Android build documentation to use CMake

juan-lunarg:


spencer-lunarg (approved):

BUILD.md
```diff
@@ -183,7 +183,12 @@ cmake --open build
 
 See the [CMake documentation](https://cmake.org/cmake/help/latest/generator/Xcode.html) for further information on the Xcode generator.
 
-## Building On Android
+## Building For Android
+
+- CMake 3.21+
+- NDK r25+
+- Ninja 1.10+
```
lunarpapillo, line 190: I'm not certain, so correct me when I'm wrong... both the `cmake` and the `ninja` used by Android builds comes with the Android SDK when you install the components as listed below, yes?

Rather than a minimum CMake and Ninja version, the minimum SDK version should be listed (as the user could install a newer CMake and/or Ninja and still not get builds working correctly).

BUILD.md
```diff
@@ -232,52 +243,60 @@ sudo apt install default-jre
...
+```sh
+cmake -S . -B build \
+  -D CMAKE_TOOLCHAIN_FILE=$ANDROID_NDK_HOME/build/cmake/android.toolchain.cmake \
+  -D ANDROID_PLATFORM=26 \
+  -D CMAKE_ANDROID_ARCH_ABI=arm64-v8a \
+  -D CMAKE_ANDROID_STL_TYPE=c++_static \
+  -D ANDROID_USE_LEGACY_TOOLCHAIN_FILE=NO \
+  -D CMAKE_BUILD_TYPE=Release \
+  -D UPDATE_DEPS=ON \
+  -G "Ninja"
```
lunarpapillo, line 261: The quotes around "Ninja" are not necessary.  I'd suggest they be removed, or quotes be put around all the other arguments in a similar unnecessary way, for consistency.

Is the `-G Ninja` actually necessary at all?  Is that implied by the toolchain?  I'm struggling to remember the details.

BUILD.md
```diff
@@ -232,52 +243,60 @@ sudo apt install default-jre
...
-> **NOTE:** By default, the `build_all.sh` script will build for all Android ABI variations. To **speed up the build time** if you know your target(s), set `APP_ABI` in both [build-android/jni/Application.mk](build-android/jni/Application.mk) and [build-android/jni/shaderc/Application.mk](build-android/jni/shaderc/Application.mk) to the desired [Android ABI](https://developer.android.com/ndk/guides/application_mk#app_abi)
+Alternatively users can also use `scripts/android.py` to build the binaries.
 
-Resulting validation layer binaries will be in `build-android/libs`. Test and demo APKs can be installed on production devices with:
+Note: `scripts/android.py` will place the binaries in the `build-android/libs` directory.
 
 ```sh
-./install_all.sh [-s <serial number>]
+# Build release binaries for all ABIs
+python scripts/android.py --config Release --app-abi 'armeabi-v7a arm64-v8a x86 x86_64' --app-stl c++_static
```
lunarpapillo, line 277: For most Unix users, `python` is still Python 2; you'll need `python3` to get to Python 3.

lunarpapillo, line 277: It's strange to list all 4 Android ABIs here, when the command line above at line 256 named only `arm64-v8a`.

BUILD.md
```diff
@@ -232,52 +243,60 @@ sudo apt install default-jre
...
+
+# Build debug binary for arm64-v8a
+python scripts/android.py --config Debug --app-abi 'arm64-v8a' --app-stl c++_static
```
lunarpapillo, line 280: Probably `python` => `python3`.

The quotes around `arm64-v8a` are not necessary.  For consistency, either remove them, or quote all arguments.

BUILD.md
```diff
@@ -232,52 +243,60 @@ sudo apt install default-jre
...
 
-    cd build-android
-    ./update_external_sources_android.sh --no-build
-    ndk-build -j4
+This script handles wrapping CMake and various Android CLI tools to create the APK for you.
 
-#### Windows
+```sh
+# Build a complete test APK with debug binaries for all ABIS
+python scripts/android.py --config Debug --app-abi 'armeabi-v7a arm64-v8a x86 x86_64' --app-stl c++_shared --apk --tests
```
lunarpapillo, line 293: Probably `python` => `python3`

BUILD.md
```diff
@@ -232,52 +243,60 @@ sudo apt install default-jre
...
 
-Follow the setup steps for Windows above, then from the Developer Command Prompt:
+# Build a test APK with release binaries for arm64-v8a
+python scripts/android.py --config Release --app-abi 'arm64-v8a' --app-stl c++_shared --apk --tests
```
lunarpapillo, line 296: Probably `python` => `python3`.

Probably remove quotes from `arm64-v8a`.

lunarpapillo (changes requested):
Thank you for this - it's nice to have the documentation self-consistent with our practices.

BUILD.md
artem-lunarg: Not sure if any of this should run on Windows, but if it is, then on Windows `python3` often does not exist, but all distributions provide `python`. One use case I'm aware of, is Windows app alias that you get after OS installation and usually you want to disable it as soon as possible:
image

BUILD.md
juan-lunarg: image

I'm not sure what you mean here. As you can see you just pick the CMake version you want.

BUILD.md
juan-lunarg: You can build for Android using a Windows host.

BUILD.md
juan-lunarg: Users can also build for Android on MacOS.

juan-lunarg: Also in the NDK documentation there are multiple examples of using CMake + Ninja. Not just using the default generator.

juan-lunarg:
> I address the feedback

addressed*

juan-lunarg:
Note Jenkins doesn't have to use Ninja if it doesn't want to. But it's much simpler for cross-platform build instructions for Android for our BUILD.md file

And does seem to be what the Android developers want you to use based on the NDK documentation.

PR #6155, opened 2023-07-18
layers: Create dedicated shader interface file

spencer-lunarg:
Help separate logic for all the shader code dealing only with interactions between 2 shader interfaces


artem-lunarg:
LGTM. That's very nice, there is a chapter for shader interfaces, and if I'd like to check something from that area it's much easier now to find.

Issue #5863, opened 2023-05-17
test and label kVUID_Core_DrawState_NoActiveRenderpass


spencer-lunarg:
As pointed out in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5859

- We have no test for `kVUID_Core_DrawState_NoActiveRenderpass`
- We need to get this into a proper VU

spencer-lunarg:
Actually there is a 

> VUID-vkCmdDraw-renderpass
> 
> This command must only be called inside of a render pass instance

not sure how we are getting here now

PR #6158, opened 2023-07-18
layers: Guarantee there is an active renderpass

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5863

We have all these `if (activeRenderPass) {` which don't make sense because there is a bunch of implicit VUs in `ValidateCmd` that check if we are in a renderpass or not.

I went through and all the command that require to be in renderPass, I wrote a test (to confirm it won't crash) and then removed the redundant checks so now it is clear if we actually need to check for renderPass

some small house cleaning was done along the way

### DIFF
--- a/layers/core_checks/cc_drawdispatch.cpp
+++ b/layers/core_checks/cc_drawdispatch.cpp
@@ -2676,7 +2676,6 @@ const DrawDispatchVuid &CoreChecks::GetDrawDispatchVuid(CMD_TYPE cmd_type) const
 // Action command == vkCmdDraw*, vkCmdDispatch*, vkCmdTraceRays*, etc.
 bool CoreChecks::ValidateActionCmd(const CMD_BUFFER_STATE &cb_state, VkPipelineBindPoint bind_point, CMD_TYPE cmd_type) const {
     bool skip = false;
-    skip |= ValidateCmd(cb_state, cmd_type);
     skip |= ValidateCmdBufDrawState(cb_state, cmd_type, bind_point);
     skip |= ValidateCmdRayQueryState(cb_state, cmd_type, bind_point);
     return skip;
@@ -2699,7 +2698,7 @@ bool CoreChecks::ValidateCmdDrawInstance(const CMD_BUFFER_STATE &cb_state, uint3
     const char *caller = CommandTypeString(cmd_type);
 
     // Verify maxMultiviewInstanceIndex
-    if (cb_state.activeRenderPass && cb_state.activeRenderPass->renderPass() && enabled_features.multiview_features.multiview &&
+    if (cb_state.activeRenderPass->renderPass() && enabled_features.multiview_features.multiview &&
         ((instanceCount + firstInstance) > phys_dev_ext_props.multiview_props.maxMultiviewInstanceIndex)) {
         const LogObjectList objlist(cb_state.Handle(), cb_state.activeRenderPass->Handle());
         skip |= LogError(objlist, vuid.max_multiview_instance_index_02688,
@@ -2756,6 +2755,9 @@ bool CoreChecks::PreCallValidateCmdDraw(VkCommandBuffer commandBuffer, uint32_t 
                                         uint32_t firstVertex, uint32_t firstInstance) const {
     bool skip = false;
     auto cb_state = GetRead<CMD_BUFFER_STATE>(commandBuffer);
+    skip |= ValidateCmd(*cb_state, CMD_DRAW);
+    if (skip) return skip;  // basic validation failed, might have null pointers
+
     skip |= ValidateCmdDrawInstance(*cb_state, instanceCount, firstInstance, CMD_DRAW);
     skip |= ValidateActionCmd(*cb_state, VK_PIPELINE_BIND_POINT_GRAPHICS, CMD_DRAW);
     skip |= ValidateVTGShaderStages(*cb_state, CMD_DRAW);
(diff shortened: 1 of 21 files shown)


layers/core_checks/cc_drawdispatch.cpp
```diff
@@ -2676,7 +2676,6 @@ const DrawDispatchVuid &CoreChecks::GetDrawDispatchVuid(CMD_TYPE cmd_type) const
 // Action command == vkCmdDraw*, vkCmdDispatch*, vkCmdTraceRays*, etc.
 bool CoreChecks::ValidateActionCmd(const CMD_BUFFER_STATE &cb_state, VkPipelineBindPoint bind_point, CMD_TYPE cmd_type) const {
     bool skip = false;
-    skip |= ValidateCmd(cb_state, cmd_type);
```
spencer-lunarg, line 2679: `ValidateCmd` already makes use of a lot of generated code and is called for all commands.

We should really make use of this function more in future to do an "init check" on all commands ... basically its like a stateless validation, but requires the `CMD_BUFFER_STATE`

artem-lunarg: I don't understand why we remove `ValidateCmd` here. Could you please elaborate.

The first search example for me it's `CoreChecks::PreCallValidateCmdDraw` calls `ValidateActionCmd`, so is it fine not to call `ValidateCmd` here or it is called somewhere else?

layers/core_checks/cc_drawdispatch.cpp
```diff
@@ -2676,7 +2676,6 @@ const DrawDispatchVuid &CoreChecks::GetDrawDispatchVuid(CMD_TYPE cmd_type) const
 // Action command == vkCmdDraw*, vkCmdDispatch*, vkCmdTraceRays*, etc.
 bool CoreChecks::ValidateActionCmd(const CMD_BUFFER_STATE &cb_state, VkPipelineBindPoint bind_point, CMD_TYPE cmd_type) const {
```
artem-lunarg, line 2677: Makes me wonder how universal is `ValidateActionCmd` and whether we can also consider to get rid of it (when time is right). Currently it calls `ValidateCmdBufDrawState` + `ValidateCmdRayQueryState`. And `ValidateActionCmd` is called for draws, dispatches, traces and meshdraws. Not sure if all these actions require the above two commands but even if it is, then flattening call hierarchy a bit might simplify things?

layers/core_checks/cc_drawdispatch.cpp
artem-lunarg: Thanks for clarification (I deleted the question after asked..)

layers/state_tracker/state_tracker.cpp
```diff
@@ -3732,11 +3726,13 @@ void ValidationStateTracker::PreCallRecordCmdBeginRendering(VkCommandBuffer comm
 }
 
 void ValidationStateTracker::PreCallRecordCmdEndRenderingKHR(VkCommandBuffer commandBuffer) {
-    RecordCmdEndRenderingRenderPassState(commandBuffer);
+    auto cb_state = GetWrite<CMD_BUFFER_STATE>(commandBuffer);
+    cb_state->EndRendering(CMD_ENDRENDERING);
```
artem-lunarg, line 3730: typo: enums should be swapped in this and the next command

tests/framework/render.cpp
```diff
@@ -2266,19 +2266,19 @@ void VkCommandBufferObj::NextSubpass(VkSubpassContents contents) { vk::CmdNextSu
 void VkCommandBufferObj::EndRenderPass() { vk::CmdEndRenderPass(handle()); }
 
 void VkCommandBufferObj::BeginRendering(const VkRenderingInfoKHR &renderingInfo) {
-    PFN_vkCmdBeginRenderingKHR vkCmdBeginRenderingKHR =
-        (PFN_vkCmdBeginRenderingKHR)vk::GetDeviceProcAddr(m_device->device(), "vkCmdBeginRenderingKHR");
-    assert(vkCmdBeginRenderingKHR != nullptr);
-
-    vkCmdBeginRenderingKHR(handle(), &renderingInfo);
+    if (vk::CmdBeginRenderingKHR) {
```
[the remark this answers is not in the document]
spencer-lunarg: yes, will add comment

tests/unit/renderpass.cpp
```diff
@@ -4653,3 +4653,26 @@ TEST_F(NegativeRenderPass, SubpassAttachmentImageLayoutSeparateDepthStencil) {
...
     }
 }
+
+TEST_F(NegativeRenderPass, BeginInfoWithoutRenderPass) {
+    TEST_DESCRIPTION("call VkRenderPassBeginInfo with invalid renderpass");
+    ASSERT_NO_FATAL_FAILURE(Init());
+    ASSERT_NO_FATAL_FAILURE(InitRenderTarget());
+    m_commandBuffer->begin();
+    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkRenderPassBeginInfo-renderPass-parameter");
+    m_renderPassBeginInfo.renderPass = CastFromUint64<VkRenderPass>(0xFFFFEEEE);
```
artem-lunarg, line 4663: Is it the intention here to start with invalid render pass?

juan-lunarg (approved):

PR #6159, opened 2023-07-18
android: Remove unused scripts

juan-lunarg:


spencer-lunarg (approved):
never thought I would live to see the day

PR #6161, opened 2023-07-19
WIP scripts: Remove hardcode API string in extension helper

spencer-lunarg:
to help reduce the workaround needed for the code in #6150 

Will also make a Vulkan 1.4 much easier as it will be generated automatically for us (which is what we want!)

### DIFF
--- a/layers/vulkan/generated/vk_extension_helper.h
+++ b/layers/vulkan/generated/vk_extension_helper.h
@@ -285,24 +285,25 @@ struct InstanceExtensions {
     }
 
     APIVersion InitFromInstanceCreateInfo(APIVersion requested_api_version, const VkInstanceCreateInfo *pCreateInfo) {
-        constexpr std::array<const char*, 5> V_1_1_promoted_instance_apis = {
+        constexpr std::array<const char*, 5> vk_version_1_1_promoted_instance_apis = {
             VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME,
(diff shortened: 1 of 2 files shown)


juan-lunarg (approved):

spencer-lunarg:
cc @aqnuep

aqnuep:
PR #6150 provides a better solution for this already. Also, if really needed, we can further refine that instead to generate the variable names for `vk_feature_*` and `v_*_promoted` based on the version string instead. But this PR is just a partial solution for the overall problem with still many of the things left hard-coded.

spencer-lunarg:
> PR #6150 provides a better solution for this already.

I think hard-coding version like we were was bad, we should just be using the new `self.vk.versions` to make Vulkan 1.4 intergration for use easier... this also should remove the need for you to need logic to decide which hardcoded array  to use

aqnuep:
> I think hard-coding version like we were was bad, we should just be using the new self.vk.versions to make Vulkan 1.4 intergration for use easier... this also should remove the need for you to need logic to decide which hardcoded array to use

Not quite, as unlike in Vulkan, where versions are strictly linear, with SC we have cross-dependencies, like Vulkan SC 1.0 requiring Vulkan 1.2 with modifications. The dependency possibilities enabled by `vk.xml` for versions is almost like that for extensions. Any version can depend on any other version, just like with extensions. Ignoring that would only cause problems down the road.

spencer-lunarg:
this logic is rarely ever touched, would be more churn to push this forward then just leave it

PR #6162, opened 2023-07-19
WIP scripts: Move APIVersion class to own file

spencer-lunarg:
cc @aqnuep

### DIFF
--- a/layers/vulkan/generated/vk_api_version.h
+++ b/layers/vulkan/generated/vk_api_version.h
@@ -1,0 +1,70 @@
+// *** THIS FILE IS GENERATED - DO NOT EDIT ***
+// See api_version_generator.py for modifications
+
+/***************************************************************************
+*
+* Copyright (c) 2015-2023 The Khronos Group Inc.
+* Copyright (c) 2015-2023 Valve Corporation
+* Copyright (c) 2015-2023 LunarG, Inc.
+* Copyright (c) 2015-2023 Google Inc.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


aqnuep:
Please don't merge it. In the future parts of the `APIVersion` class could very well be auto-generated based on the version information in the `vk.xml`, and this just breaks that option.

We already agreed with @juan-lunarg in PR #5860 that it's better to keep this class generated.

juan-lunarg:
> We already agreed with @juan-lunarg in PR #5860 that it's better to keep this class generated.

I think you are referring to this discussion on the gitlab (https://gitlab.khronos.org/vulkan/Vulkan-ValidationLayers/-/merge_requests/91#note_411319).

>  In the future parts of the APIVersion class could very well be auto-generated based on the version information in the vk.xml, and this just breaks that option.

Wouldn't this PR make that easier? Maybe I'm reading the codegen wrong though.

aqnuep:
> > We already agreed with @juan-lunarg in PR #5860 that it's better to keep this class generated.
> 
> I think you are referring to this discussion on the gitlab (https://gitlab.khronos.org/vulkan/Vulkan-ValidationLayers/-/merge_requests/91#note_411319).
> 
> > In the future parts of the APIVersion class could very well be auto-generated based on the version information in the vk.xml, and this just breaks that option.
> 
> Wouldn't this PR make that easier? Maybe I'm reading the codegen wrong though.

It would only work if we assume that downstream would completely replace the script itself, and we assume that in the future no refactoring or additional code added to this generator that is incompatible with downstream would be added, as there are no hooks to change behavior of on a per-API-basis, so as is it's an all-or-nothing kind of script.

Fair enough though, honestly, I didn't review the entire PR, as I was immediately annoyed by the fact that a PR was posted to cause conflicts with my PR, when, if requested, I would have happily do it in my PR with appropriate API-specific hooks.

I think the key thing is still to try to separate out the parts that are supposed to be specialized per API from the rest of the code that's supposed to be common, which this PR does not achieve in any fashion.

I'm not objecting against separating out the generator for the APIVersion class to a separate header, just would like to have the API-specific hook instead of inlining everything in the main generator script, as it is proposed here.

PR #6164, opened 2023-07-19
layers: Reorganize draw/dispatch/rayTracing common checks

spencer-lunarg:
- remove `ValidateActionCmd`
- move `ValidateCmdRayQueryState` into what was `ValidateCmdBufDrawState`
- renamed this common check to `ValidateActionState`


layers/core_checks/cc_drawdispatch.cpp
```diff
@@ -3426,9 +3417,11 @@ bool CoreChecks::PreCallValidateCmdDrawMeshTasksIndirectCountEXT(VkCommandBuffer
     return skip;
 }
 
-// Validate overall state at the time of a draw call
-bool CoreChecks::ValidateCmdBufDrawState(const CMD_BUFFER_STATE &cb_state, CMD_TYPE cmd_type,
-                                         const VkPipelineBindPoint bind_point) const {
+// Action command == vkCmdDraw*, vkCmdDispatch*, vkCmdTraceRays*
+// Does not include commands like vkCmdBuildAccelerationStructure*
```
artem-lunarg, line 3421: Is this exclusion due to how we structure the code, or is it related to the specification terminology (although I can't find in the specification a special case for the acceleration structure build commands)?

scripts/generators/command_validation_generator.py
```diff
@@ -223,7 +223,9 @@ def generateSource(self):
         # The main function to validate all the commands
         # TODO - Remove C++ code from being a single python string
         out.append('''
-// Used to handle all the implicit VUs that are autogenerated from the registry
+// Ran on all vkCmd* commands
+// Because it validate the implicit VUs that stateless can't, if this fails, it is likely
```
artem-lunarg, line 227: Nice clarification how this relates to stateless 👍

layers/core_checks/cc_drawdispatch.cpp
spencer-lunarg: > related to the specification terminology

sadly there is no terminology in the spec lol https://github.com/KhronosGroup/Vulkan-Docs/issues/1908

I guess thinking about it, it doesn't have to have a pipeline bound (ignoring VK_EXT_shader_object for the moment). So I guess this would in theory be the same as `vkCmdCopyImage`

... let me look more first actually, might wait for Arno to come back

spencer-lunarg: also might rename `Action` to `Provoke` now

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -1090,7 +1090,7 @@ bool CoreChecks::PreCallValidateCmdBuildAccelerationStructuresIndirectKHR(VkComm
     assert(cb_state);
     bool skip = false;
     skip |= ValidateCmd(*cb_state, CMD_BUILDACCELERATIONSTRUCTURESINDIRECTKHR);
-    skip |= ValidateCmdRayQueryState(*cb_state, CMD_BUILDACCELERATIONSTRUCTURESINDIRECTKHR, VK_PIPELINE_BIND_POINT_RAY_TRACING_KHR);
+    skip |= ValidateCmdRayQueryState(*cb_state, VK_PIPELINE_BIND_POINT_RAY_TRACING_KHR, CMD_BUILDACCELERATIONSTRUCTURESINDIRECTKHR);
```
artem-lunarg, line 1093: I think that `ValidateCmdRayQueryState` is ill-suited to be used for `vkCmdBuildAccelerationStructuresIndirectKHR` validation. Even thought it correctly detects violation of:
> VUID-vkCmdBuildAccelerationStructuresIndirectKHR-commandBuffer-03649
commandBuffer must not be a protected command buffer

it does this by using functionality written for RayQuery operations (shader stuff). It also mentions RayQuery as part of the error message. 

Alternative solution could be to make the check, like the one provided below, directly in the function. It's simpler than  `ValidateCmdRayQueryState` (and `ValidateCmdRayQueryState` will be even more complex in the future since it has to do shader analysis):
```
    if (cb_state.unprotected == false) {
        skip |= LogError(cb_state.commandBuffer(), vuid.ray_query_protected_cb_03635,
                         "%s(): can't use in protected command buffers for acceleration structure builds.", CommandTypeString(cmd_type));
    }
```

Then we can `ValidateCmdRayQueryState` only for situations when shader issues RayQuery.

I agree, that it's not part of this PR and could be a separate issue.

layers/core_checks/cc_drawdispatch.cpp
artem-lunarg: > sadly there is no terminology in the spec lol https://github.com/KhronosGroup/Vulkan-Docs/issues/1908

This is only this vague description:

> In general, action commands are those commands that alter framebuffer attachments, read/write buffer or image memory, or write to query pools.

artem-lunarg: The above sentence from the spec sounds like it's about functionality that initiates some GPU execution and has side effects (draws, copies, dispatches, traces, clears) in contrast to state update operations or scheduling operations (barriers).

artem-lunarg (approved):
LGTM. Some suggestions, not necessarily part of this PR.

layers/core_checks/cc_ray_tracing.cpp
spencer-lunarg: There is a long-overdue TODO I have in `ValidateCmdRayQueryState`, I think for now, leave it here, remove the comment above, and do another cleanup post-PR for just the Ray Query stuff.... I will make a separate commit on the PR to make what I mean more visual

PR #6165, opened 2023-07-19
scripts: Restore the -o option on generate_source

Keenuts:
The old script had a way to specify a custom output directory for generated files. The recent refactoring removed this option.

This option is important if your CI runs on a read-only version of the repository, and generated files/intermediates are written to a different folder before being processed (Bazel-like build).

### DIFF
--- a/scripts/generate_source.py
+++ b/scripts/generate_source.py
@@ -328,6 +328,7 @@ def main(argv):
     parser.add_argument('--generated-version', help='sets the header version used to generate the repo')
     group = parser.add_mutually_exclusive_group()
     group.add_argument('--target', nargs='+', help='only generate file names passed in')
+    group.add_argument('-o', help='Create target and related files in specified directory.', dest='output_directory')
     group.add_argument('-i', '--incremental', action='store_true', help='only update repo files that change')
     group.add_argument('-v', '--verify', action='store_true', help='verify repo files match generator output')
     args = parser.parse_args(argv)
@@ -357,6 +358,9 @@ def main(argv):
     else:
         # generate directly in the repo
         gen_dir = repo_dir
+
+    if args.output_directory is not None:
+      gen_dir = args.output_directory;
 
     registry = os.path.abspath(os.path.join(args.registry,  'vk.xml'))
     grammar = os.path.abspath(os.path.join(args.grammar, 'spirv.core.grammar.json'))


spencer-lunarg (approved):

scripts/generate_source.py
```diff
@@ -302,6 +303,9 @@ def main(argv):
         # generate directly in the repo
         gen_dir = repo_dir
 
+    if args.output_directory is not None:
+      gen_dir = args.output_directory;
```
juan-lunarg, line 307: > The old script had a way to specify a custom output directory for generated files. The recent refactoring removed this option, preventing us to build this.

Who/what did we break exactly? I'd like that to be commented at least.

juan-lunarg (changes requested):

scripts/generate_source.py
Keenuts: Reworded the commit message to explain that.
This is useful when your build-system does the following:
 - source code is read-only
 - generated files are written to a writable intermediate folder
 - then, code is compiled, using those intermediates.
 
Also, before this refactoring, we were able to generate each file individually using the fine-grained python script. This means the build-system is responsible from parallelization. Now, we can only call this, and not parallelize this file generation step easily.

Another note: this script calls another python script using subprocess.check_call. This assumes the machine can run python scripts as-is. What if we require python scripts to be bundled first to run in an hermetic env?
(Also, same issue with the output directory)

spencer-lunarg: >  we were able to generate each file individually using the fine-grained python script

you can in theory run `generate_source.py --target file_name` to do one file at a time

... but honestly, I would want to make parallelize generation the default, not something that is added afterwards

> this script calls another python script using subprocess.check_call

so before, ALL files were being called with `check_call`, now only the `vk_validation_stats.py` file is being called with it (it generates the `vk_validation_error_messages.h` and something that **can** be cleaned up in the future as well) ... so not sure why this is a problem now

Keenuts: Thanks for the answer!

> you can in theory run `generate_source.py --target file_name` to do one file at a time
Oh, missed this one, thanks, that will be handy!


> so before, ALL files were being called with `check_call`
It was? I must have missed that in the old `run_generator.py` file (we called this one directly)

But those are 2 other problems, we stil remain with the issue that we cannot write in the repo directory, hence need a way to say "go output the files there", which was the output option.

juan-lunarg (approved):

Keenuts:
thanks for the review! force-pushed to rebase

Issue #6107, opened 2023-07-06
Queries with multiview + wait causes spurious UNASSIGNED-CoreValidation-DrawState-InvalidQuery 


juston-li:
**Environment:**
 - OS: Linux
 - GPU: anv
 - SDK or header version if building from repo: 1.3.243
 - Options enabled (synchronization, best practices, etc.)

Per spec 
> If queries are used while executing a render pass instance that has
> multiview enabled, the query uses N consecutive query indices in the
> query pool (starting at query) where N is the number of bits set in
> the view mask in the subpass the query is used in.

When `vkCmdCopyQueryPoolResults` is used with `VK_QUERY_RESULT_WAIT_BIT` and multiview enabled, VVL logs these errors.

> Validation Error: [ UNASSIGNED-CoreValidation-DrawState-InvalidQuery ]
> Object 0: handle = 0x7b39e4902ed0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x46ab3a6e |
> vkCmdCopyQueryPoolResults(): Requesting a copy from query to buffer on VkQueryPool 0xd897d90000000016[]
> query 1: waiting on a query that has been reset and not issued yet
> 
> Validation Error: [ UNASSIGNED-CoreValidation-DrawState-InvalidQuery ]
> Object 0: handle = 0x7b39e4902ed0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x46ab3a6e |
> vkCmdCopyQueryPoolResults(): Requesting a copy from query to buffer on VkQueryPool 0xd897d90000000016[]
> query 2: waiting on a query that has been reset and not issued yet
> 
> Validation Error: [ UNASSIGNED-CoreValidation-DrawState-InvalidQuery ]
> Object 0: handle = 0x7b39e4902ed0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x46ab3a6e |
> vkCmdCopyQueryPoolResults(): Requesting a copy from query to buffer on VkQueryPool 0xd897d90000000016[]
> query 3: waiting on a query that has been reset and not issued yet

In this instance, the BeginQuery was called on query 0 inside a renderpass with multiview enabled, generating 3 additional queries.

VVL thinks queries 1-3 have not been started because BeginQuery wasn't explicitly called for them.

**Expected behavior**
VVL should recognize multiview generates additional queries and include them when tracking Begin/End Query

[...]

PR #6178, opened 2023-07-21
Cleanup usage of query_pool_state

spencer-lunarg:
While looking into https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6107 I realized the `cc_query.cpp` file needed some cleanup

I added tests to make sure a valid `VkQueryPool` is guaranteed


layers/core_checks/cc_query.cpp
```diff
@@ -28,21 +28,21 @@
 bool CoreChecks::PreCallValidateDestroyQueryPool(VkDevice device, VkQueryPool queryPool,
                                                  const VkAllocationCallbacks *pAllocator) const {
     if (disabled[query_validation]) return false;
-    auto qp_state = Get<QUERY_POOL_STATE>(queryPool);
+    auto query_pool_state = Get<QUERY_POOL_STATE>(queryPool);
+    if (!query_pool_state) return false;
```
artem-lunarg, line 32: This probably was discussed in the past but then it's a refresher. In which cases can there can be a null pointer here? Is it for use case when Stateless checks are disabled?

spencer-lunarg: All destroy calls can have null, (see `VkPositiveLayerTest.TestDestroyFreeNullHandles`) Luckily we have tests to catch this

layers/core_checks/cc_query.cpp
```diff
@@ -88,128 +88,112 @@ bool CoreChecks::ValidatePerformanceQueryResults(const char *cmd_name, const QUE
...
+                if (query_size > stride) {
+                    skip |= LogError(queryPool, "VUID-vkGetQueryPoolResults-queryType-04519",
+                                     "vkGetQueryPoolResults() on querypool %s specified stride %" PRIu64
+                                     " which must be at least counterIndexCount (%d) "
+                                     "multiplied by sizeof(VkPerformanceCounterResultKHR) (%zu).",
+                                     report_data->FormatHandle(queryPool).c_str(), stride, query_items,
+                                     sizeof(VkPerformanceCounterResultKHR));
+                }
+
+                if (((((uintptr_t)pData) % sizeof(VkPerformanceCounterResultKHR)) != 0 ||
```
artem-lunarg, line 171: nit: it looks one outer `( )` can be safely removed

layers/core_checks/cc_query.cpp
artem-lunarg: Okay, one option could be to check `queryPool` directly against VK_NULL_HANDLE to separate this case, so `Get<QUERY_POOL_STATE>` will self-document it can't be zero, or if it also can be zero then due to different reason.

spencer-lunarg: ya, I agree, checking `queryPool` for `VK_NULL_HANDLE` is better in every way, will update

jeremyg-lunarg (approved):

PR #6179, opened 2023-07-21
scripts: Removes exclusivity constraint on -o

Keenuts:
PR 6165 re-introduced the `-o` flag for the generator script, but I made a mistake and added it to the group instead of the parser itself. This means we cannot use both `--target` and `-o`.


Keenuts:
There is 1 workflow failure: github returned an error 500 when getting vulkan profiles. Seems unrelated.

juan-lunarg (approved):

artem-lunarg (approved):

PR #6181, opened 2023-07-21
android: Support legacy toolchain file

juan-lunarg:
Dependent upon recent fixes to glslang:
KhronosGroup/glslang/pull/3264


juan-lunarg:
APK generated by this build failed to run on my Pixel 3. Using the legacy toolchain isn't supported for now.

PR #6182, opened 2023-07-21
build: Update to header 1.3.259

mikes-lunarg:
- Update known-good, ~include shader toolchain for upcoming SDK~ EDIT: no shader toolchain updates for now
- Generate source

PR #6185, opened 2023-07-24
Fix 1.3.258 VUID Churn

spencer-lunarg:
VUID churn for https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6182

cc @mikes-lunarg

### DIFF
--- a/layers/core_checks/cc_synchronization.cpp
+++ b/layers/core_checks/cc_synchronization.cpp
@@ -2072,8 +2072,7 @@ class ValidatorState {
           loc_(location),
           barrier_handle_(barrier_handle),
           sharing_mode_(sharing_mode),
-          limit_(static_cast<uint32_t>(device_data->physical_device_state->queue_family_properties.size())),
-          mem_ext_(IsExtEnabled(device_data->device_extensions.vk_khr_external_memory)) {}
+          limit_(static_cast<uint32_t>(device_data->physical_device_state->queue_family_properties.size())) {}
 
     // Log the messages using boilerplate from object state, and Vu specific information from the template arg
     // One and two family versions, in the single family version, Vu holds the name of the passed parameter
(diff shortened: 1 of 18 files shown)


juan-lunarg (approved):

tests/unit/sync_object.cpp
```diff
@@ -1607,41 +1616,10 @@ TEST_F(NegativeSyncObject, Sync2BarrierQueueFamily) {
     BarrierQueueFamilyTestHelper::Context test_context(this, qf_indices);
     Barrier2QueueFamilyTestHelper::Context test_context2(this, qf_indices);
 
-    if (only_one_family) {
```
spencer-lunarg, line 1610: note why this was removed 

the old VUs were for both Sync1 and Sync2, but from https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5979 it was realized it should have just been Sync1 and the new VUs reflect that

> VUID-VkImageMemoryBarrier-None-09052
If the synchronization2 feature is not enabled, and image was created with a sharing mode of VK_SHARING_MODE_CONCURRENT, at least one of srcQueueFamilyIndex and dstQueueFamilyIndex must be VK_QUEUE_FAMILY_IGNORED

layers/core_checks/cc_synchronization.cpp
```diff
@@ -2160,53 +2157,56 @@ class ValidatorState {
...
     const VkSharingMode sharing_mode_;
     const uint32_t limit_;
-    const bool mem_ext_;
 };
 
 bool Validate(const CoreChecks *device_data, const CMD_BUFFER_STATE *cb_state, const ValidatorState &val,
               const uint32_t src_queue_family, const uint32_t dst_queue_family) {
     bool skip = false;
 
-    const bool mode_concurrent = val.GetSharingMode() == VK_SHARING_MODE_CONCURRENT;
```
spencer-lunarg, line 2170: this function was basically just re-written, from `sync_vuid_maps.h` you see all the VUID were changed in https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5979 and made the logic a lot simpler to understand what is going on (also broke things into a VU for `src` and `dst`)

arno-lunarg (approved):

PR #6187, opened 2023-07-24
layers: Rename SHADER_MODULE_STATE to SPIRV_MODULE_STATE

ziga-lunarg:
In preparation for VK_EXT_shader_object validation rename SHADER_MODULE_STATE to SPIRV_MODULE_STATE and the file cc_shader.cpp to cc_spirv.cpp


spencer-lunarg (approved):
So the current issue with this is there is still a `VkShaderModule` handle in `SPIRV_MODULE_STATE`, but the only 2 places I think that "need" that is

1. `VK_EXT_validation_cache` (which I still need to find out how we were suppose to do that with `VK_EXT_shader_object`)
2. Somewhere after `vkCreateShaderModule` were we separate the two

This is a task for after, because we should start thinking of the `module_state` as just the SPIR-V, not the `VkShaderModule` object

layers/state_tracker/state_tracker.cpp
```diff
@@ -2272,7 +2272,7 @@ void ValidationStateTracker::PostCallRecordGetImageSparseMemoryRequirements2KHR(
 
 void ValidationStateTracker::PreCallRecordDestroyShaderModule(VkDevice device, VkShaderModule shaderModule,
                                                               const VkAllocationCallbacks *pAllocator) {
-    Destroy<SHADER_MODULE_STATE>(shaderModule);
```
spencer-lunarg, line 2275: This will be one of the spots we will need to update post-PR to separate the `VkShaderModule` with `SHADER_MODULE_STATE`

Issue #6112, opened 2023-07-07
Descriptor buffer CTS tests crash running with synchronization validation


Tony-LunarG:
**Environment:**
 - OS:  Windows 10
 - GPU: RTX 2080
 - SDK or header version if building from repo:  1.3.253
 - Options enabled (synchronization, best practices, etc.): synchronization

With synchronization validation enabled, and all other validation disabled, these tests crash:
dEQP-VK.binding_model.descriptor_buffer.push_template.graphics_vert_sets2_push_set0
dEQP-VK.binding_model.descriptor_buffer.push_descriptor.graphics_vert_sets2_push_set0

The crash is vector subscript out of range in sync_validation.cpp line 2048:
```const auto *descriptor_set = (*per_sets)[variable.decorations.set].bound_descriptor_set.get();```
per_sets is size of 1, variable.decorations.set is 1

**Expected behavior:**

The tests should not crash

artem-lunarg:
This started happening after https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5420
It's not clear if that change is the root cause but it enabled this behavior

artem-lunarg:
@Tony-LunarG, the latest code has fix for this crash

Tony-LunarG:
Verified fixed - Thanks!

PR #6188, opened 2023-07-24
layers: Check out of bounds access in syncval

artem-lunarg:
This can happen when core checks are disabled or when corresponding checks not implemented.

There are core validation checks for set-not-bound error for `vkCmdBindDescriptorSets` API.
That functionality is not implemented yet for `vkCmdBindDescriptorBuffers` + `vkCmdSetDescriptorBufferOffsets` API.
That is why, even with core checks enabled, we still need to do preventing checks in SyncVal.

Addresses https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6112


arno-lunarg (commented):
LGTM

spencer-lunarg (approved):

jzulauf-lunarg (commented):
LGTM

PR #6192, opened 2023-07-25
tests: Move nv ray tracing framework

arno-lunarg:


tests/framework/layer_validation_tests.h
```diff
@@ -461,14 +460,24 @@ class QueryTest : public VkLayerTest {};
...
+};
 class NegativeRayTracing : public RayTracingTest {};
 class PositiveRayTracing : public RayTracingTest {};
 
-class RayTracingPipelineTest : public VkLayerTest {};
+class RayTracingPipelineTest : public RayTracingTest {};
 class NegativeRayTracingPipeline : public RayTracingPipelineTest {};
 class PositiveRayTracingPipeline : public RayTracingPipelineTest {};
 
+class GpuAssistedRayTracingTest : public VkGpuAssistedLayerTest, public RayTracingTest {};
```
spencer-lunarg, line 478: if not obvious for others at first glance, the issue here is we want `GpuAssistedRayTracingTest` tests to both have

`RayTracingTest::InitFrameworkForRayTracingTest` like all ray tracing tests want
and
`VkGpuAssistedLayerTest::CanEnableGpuAV()`

tests/unit/ray_tracing.cpp
```diff
@@ -14,8 +14,995 @@
 
 #include "../framework/layer_validation_tests.h"
 #include "../framework/ray_tracing_objects.h"
+#include "../framework/ray_tracing_nv.h"
```
spencer-lunarg, line 17: is it worth just having a 

`unit/ray_tracing.cpp` and `unit/ray_tracing_nv.cpp` file at this point

arno-lunarg: having this split only for `ray_tracing.cpp` feels strange, but doing this split for every ray tracing test files seems overkill. Instead, to still cement the split that exists with the legacy nvidia ray tracing tests, I created dedicated test classes ending with `NV` (`NegativeRayTracingNV`, `PositiveRayTracingNV`...)

spencer-lunarg (approved):

PR #6193, opened 2023-07-25
layers: Use constant for ObjectUseData alignment

artem-lunarg:
`get_hardware_destructive_interference_size()` value is specific to build machine and at least in theory no guarantee it's the same as on the user machine.

For threading/sync code it's good to be as specific and unambiguous as possible to simplify the reasoning and validation of the code.

There could be two options here.
1. Use maximum cache line size that we have on the market. Currently it's 128 bytes.
2. Use the most common cache line size of 64 bytes.

The choice here is the 2nd option, it decreases cache pressure which is helpful for Apple devices too.

In case of potential false sharing on devices with 128 cache line size we can use conditional compilation and specify 128 alignment there, but until evidence, prefer a unified solution. Also for regular workloads we don't expect a lot of false sharing (it's a rare event in the CPU instruction stream - twice per high level Vulkan API call), but heavy threaded applications with a lot of Vulkan API calls increase the chance of collisions.

### DIFF
--- a/layers/thread_tracker/thread_safety_validation.h
+++ b/layers/thread_tracker/thread_safety_validation.h
@@ -46,7 +46,13 @@ static_assert(std::is_same<uint64_t, DISTINCT_NONDISPATCHABLE_PHONY_HANDLE>::val
 [[maybe_unused]] static const char *kVUID_Threading_MultipleThreads = "UNASSIGNED-Threading-MultipleThreads";
 [[maybe_unused]] static const char *kVUID_Threading_SingleThreadReuse = "UNASSIGNED-Threading-SingleThreadReuse";
 
-class alignas(get_hardware_destructive_interference_size()) ObjectUseData {
+// Modern CPUs have 64 or 128-byte cache line sizes (Apple M1 has 128-byte cache line size).
+// Use alignment of 64 bytes (instead of 128) to prioritize using less memory and decrease
+// cache pressure.
+inline constexpr size_t kObjectUserDataAlignment = 64;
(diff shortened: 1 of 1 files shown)


layers/thread_tracker/thread_safety_validation.h
```diff
@@ -46,7 +46,13 @@ static_assert(std::is_same<uint64_t, DISTINCT_NONDISPATCHABLE_PHONY_HANDLE>::val
 [[maybe_unused]] static const char *kVUID_Threading_MultipleThreads = "UNASSIGNED-Threading-MultipleThreads";
 [[maybe_unused]] static const char *kVUID_Threading_SingleThreadReuse = "UNASSIGNED-Threading-SingleThreadReuse";
 
-class alignas(get_hardware_destructive_interference_size()) ObjectUseData {
+// Modern CPUs have 64 or 128-byte cache line sizes (Apple M1 has 128-byte cache line size).
+// Use alignment of 64 bytes (instead of 128) to prioritize using less memory and decrease
+// cache pressure.
+inline constexpr size_t kObjectUserDataAlignment = 64;
```
spencer-lunarg, line 52: I'm for this, if we find it really worth wild, it could always be an option in the config in the future, but maybe worth making sure we don't have a perf dip before running this

artem-lunarg: Currently we use our custom implementation of `get_hardware_destructive_interference_size()` which returns 64, so functionally it's the same as before. It's mostly about the future change when we decide to replace `get_hardware_destructive_interference_size` with an implementation from std.

juan-lunarg (approved):

PR #6194, opened 2023-07-25
Add downstream interface tests

aqnuep:
This is a proposal to add tests for the downstream interfaces of VVL used to be able to implement and maintain downstream forks of VVL targeting other Vulkan API variants.

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -1,0 +1,55 @@
+# Copyright (c) 2023-2023 The Khronos Group Inc.
+# Copyright (c) 2023-2023 RasterGrid Kft.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: VVL Downstream Interfaces
+
+# https://docs.github.com/en/actions/using-jobs/using-concurrency
+concurrency:
+  # github.head_ref is only defined on pull_request
+  # Fallback to the run ID, which is guaranteed to be both unique and defined for the run.
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id  }}
+  cancel-in-progress: true
+
+on:
+    push:
+    pull_request:
+        branches:
+            - main
+
+permissions:
+  contents: read
+
+jobs:
+  linux:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - uses: lukka/get-cmake@latest
+      - name: Setup ccache
+        uses: hendrikmuhs/ccache-action@v1.2
+        with:
+          key: downstream_test_ccache
+      - name: Configure Downstream Interface Test
+        run: |
+          cmake -S tests/downstream -B tests/downstream/build
+        env:
+          CMAKE_C_COMPILER_LAUNCHER: ccache
+          CMAKE_CXX_COMPILER_LAUNCHER: ccache
+          CMAKE_GENERATOR: Ninja
+      - name: Build Downstream Interface Test
+        run: cmake --build tests/downstream/build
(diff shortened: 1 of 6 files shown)


aqnuep:
@spencer-lunarg, @juan-lunarg, @ncesario-lunarg, @dgkoch, FYI, this is the draft proposal for testing the downstream interfaces.

I'm still fighting with github actions, and I expect changes based on your feedback, so keeping it draft until all those things get out of the way.

aqnuep:
Just in case you're waiting for fixing the CI issue on this PR (which is apparently some environment only issue), I thought I'd first like to hear what you think about this approach to test the stability of these downstream interfaces before spending more time on on fixing the github action environment issues. So please let me know what you think, and whether you'd be okay with having such a CI action upstream.

spencer-lunarg:
sorry, finally this week got back to doing code-gen stuff

I have 2 main concerns with this

1. If a test is failing, and you are not around to answer the question, I for-see tests just getting blacklisted, I am happy to have these run as a "warning" so we know not to just blindly merge a change and actually take a second-look, but seems strange to fully block a PR from going in

2. https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6322 is an example where we might want to move things around, my issue with these tests here, is now if I want to make a change, I know need to adjust these tests that I don't know anything about

aqnuep:
Just to understand the extent of your feedback...

Does that mean you are fine with the way how the interfaces are tested, or you don't really care? Or you simply did not yet get the chance to think about the way the interfaces are tested in general?

> 1. If a test is failing, and you are not around to answer the question, I for-see tests just getting blacklisted, I am happy to have these run as a "warning" so we know not to just blindly merge a change and actually take a second-look, but seems strange to fully block a PR from going in

Sure, I hear you. Although I do have concerns about it being a "warning" that people will just not care and approve any breaking change. Maybe if there would be an email hook at least where the downstream maintainer list is notified about the commit that breaks things, to give a chance to look into it, that could be fine. But the problem is that once things are in, especially if, for example, an outside contributor broke things in ways that reveal that the original PR itself was incorrect, then what will we do about it? Just back it out? With such restrictions, I'm afraid we defeat the purpose of the tests.

Also, I'm not entirely sure what's the worry about me or any maintainer to be around. How is that different when something breaks VVL due to a dependency, e.g. due to a Vulkan-Headers change? What if the original author is not around? In addition, I could argue the same thing can happen if a LunarG dev breaks something in VVL that your tests don't catch but are otherwise general Vulkan (not Vulkan SC) VVL bugs and the original author is not around.

> 2. https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6322 is an example where we might want to move things around, my issue with these tests here, is now if I want to make a change, I know need to adjust these tests that I don't know anything about

That also seems like an odd argument. I'm sure you don't know everything about all the tests that you refactor or change. It's just how things are with a large project with many contributors: one has to work with limited understanding and learn from there as needed. Do you not accept outside contributions that add tests, which inevitably may need adjustments later on?

I apologize, but I don't completely understand the motivation behind some of these concerns, and how they are specific to these tests (or the API-specific interfaces for that matter) vs all other contributions from non-LunarG devs.

juan-lunarg:
Sorry I was on vacation for a bit.

IMO I think the PR looks fine overall. The downstream testing seems pretty minimal.

I have nitpicks about how it's done (The CMake / github action in code in particular) but overall it looks fine to me.

aqnuep:
Thank you both, for looking into this.

> I have nitpicks about how it's done (The CMake / github action in code in particular) but overall it looks fine to me.

Please, feel free to share those nitpicks, as I'd like to know what you'd like to be implemented before finalizing this PR.

juan-lunarg:
Looking into this more. I think a good idea would be to have this CI outside this repo.

IE having a nightly (or per commit) github action bot try to run this CI code in the corresponding safety critical repo.

That way SC is still informed of a breakage and we address the issues having this SC integration code in our repo brings up.

I think this solution addresses both parties needs fairly well. And I don't see too much of an issue with it (other than requiring more github actions knowledge potentially).

ncesario-lunarg:
> The SC repo doesn't downstream every commit from VVL, only at a certain cadence, as specs are downstreamed, which means there could be potentially months between subsequent downstreaming, hence it's not really possible to run this CI action downstream without accumulating hundreds of potentially breaking changes. This is why this PR has been proposed.

I find it hard to believe there is not a number of different ways this could be done. A cursory search yields https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule which could be used to setup a nightly job in VVL-SC that pulls down ToT VVL and run its scripts against VVL-SC. AFAICT that gets everyone what they need. Is there a specific reason you are opposed to that approach @aqnuep?

aqnuep:
Closing this, as the preference is to detect breaking changes downstream, i.e. regressions from the perspective of the downstream interfaces will have to be fixed downstream post facto and re-upstreamed afterwards.

PR #6196, opened 2023-07-25
Fix extension version checks

ziga-lunarg:


spencer-lunarg (approved):
good catch, this tests is nice

layers/stateless/stateless_validation.h
```diff
@@ -44,8 +44,8 @@ class StatelessValidation : public ValidationObject {
 
     // This was a special case where it was decided to use the extension version for validation
     // https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5671
-    uint32_t discard_rectangles_extension_version = 0;
-    uint32_t scissor_exclusive_extension_version = 0;
+    inline static uint32_t discard_rectangles_extension_version = 0;
```
artem-lunarg, line 47: Is this a fix or clean up?

ziga-lunarg: A fix, multiple objects of StatelessValidation are created. What I found was that one was recording the extension version, but another was used to test them.

artem-lunarg: Okay, I see. According to code `PostCallRecordEnumeratePhysicalDevices` initializes those variables, so it was called on one StatelessValidation object but the validation that variables was called on a different one (repeating out loud to ensure that's the case). Do you think it's a valid situation that `PostCallRecordEnumeratePhysicalDevices` is called only on one stateless object and what are the base types of those objects? Thanks!

PR #6197, opened 2023-07-26
layers: Separate SHADER_MODULE_STATE and SPIRV_MODULE_STATE

spencer-lunarg:
Follow up for https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6187 (cc @ziga-lunarg)

This breaks up the logic between a `SPIRV_MODULE_STATE` (which parses and holds the SPIR-V) and a `SHADER_MODULE_STATE` (and future `SHADER_OBJECT_STATE`) struct which is a thin wrapper around the SPIR-V

There is a lot of "fun" quirks around `SHADER_MODULE_STATE` and `VK_EXT_shader_module_identifier` / `VK_EXT_graphics_pipeline_library`, but luckily these extensions don't interact with `VK_EXT_shader_object`

This PR just separates the StateTracking objects, a follow up PR will adjust the GPU-AV to properly use `SPIRV_MODULE_STATE`

### DIFF
--- a/layers/state_tracker/shader_module.h
+++ b/layers/state_tracker/shader_module.h
@@ -411,7 +411,9 @@ struct EntryPoint {
                                                                                 const ImageAccessMap &image_access_map);
 };
 
-struct SPIRV_MODULE_STATE : public BASE_NODE {
+// Represents a SPIR-V Module
+// This holds the SPIR-V source and parse it
+struct SPIRV_MODULE_STATE {
     // Static/const data extracted from a SPIRV module at initialization time
     // The goal of this struct is to move everything that is ready only into here
     struct StaticData {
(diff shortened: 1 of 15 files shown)


layers/state_tracker/shader_module.h
```diff
@@ -487,24 +489,17 @@ struct SPIRV_MODULE_STATE : public BASE_NODE {
...
-    SPIRV_MODULE_STATE(const VkShaderModuleCreateInfo &create_info, VkShaderModule shaderModule, uint32_t unique_shader_id)
-        : BASE_NODE(shaderModule, kVulkanObjectTypeShaderModule),
-          words_(create_info.pCode, create_info.pCode + create_info.codeSize / sizeof(uint32_t)),
-          has_valid_spirv(true),
-          static_data_(*this),
-          gpu_validation_shader_id(unique_shader_id) {}
-
-    SPIRV_MODULE_STATE() : BASE_NODE(static_cast<VkShaderModule>(VK_NULL_HANDLE), kVulkanObjectTypeShaderModule) {}
+    SPIRV_MODULE_STATE(size_t codeSize, const uint32_t *pCode, VulkanTypedHandle handle)
+        : words_(pCode, pCode + codeSize / sizeof(uint32_t)), static_data_(*this), handle_(handle) {}
```
artem-lunarg, line 502: Would it be useful to have `codeSize % 4 == 0` assert in this constructor?

layers/state_tracker/shader_module.h
```diff
@@ -587,3 +580,34 @@ struct SPIRV_MODULE_STATE : public BASE_NODE {
...
+    SHADER_MODULE_STATE(const VkShaderModuleCreateInfo &create_info, VkShaderModule shader_module, uint32_t unique_shader_id = 0)
+        : BASE_NODE(shader_module, kVulkanObjectTypeShaderModule), gpu_validation_shader_id(unique_shader_id) {
+        // Sometime an empty SHADER_MODULE_STATE is needed with no actual SPIR-V backing it for GPL
+        if (create_info.pCode[0] == spv::MagicNumber) {
+            spirv = std::make_shared<SPIRV_MODULE_STATE>(create_info.codeSize, create_info.pCode, handle_);
+        }
+    }
+
+    // If null, means this is a empty object and no shader backing it
+    std::shared_ptr<SPIRV_MODULE_STATE> spirv;
```
artem-lunarg, line 595: Can `SPIRV_MODULE_STATE` referenced here outlive parent `SHADER_MODULE_STATE`? Mostly thinking if it can work with `unique_ptr`.

layers/state_tracker/shader_module.h
spencer-lunarg: We fail much earlier in  

> VUID-VkShaderModuleCreateInfo-codeSize-08735
>
>  If pCode is a pointer to SPIR-V code, codeSize must be a multiple of 4

layers/state_tracker/shader_module.h
spencer-lunarg: yes, because you can go

```
vkCreateShaderModule()
vkCreateGraphicsPipeline()
vkDestroyShaderModule()
// everything else
vkCmdDraw()
```

and we need to track this at draw time, which we do with the `std::shared_ptr<const SPIRV_MODULE_STATE> module_state;` in `PipelineStageState`

artem-lunarg: Thanks for providing good example

artem-lunarg: It looks `PipelineStageState` uses now `SHADER_MODULE_STATE` :
` std::shared_ptr<const SHADER_MODULE_STATE> module_state;`

spencer-lunarg: sorry, was looking at my code on different branch

So currently we have been (before this PR) holding onto the `SHADER_MODULE_STATE` because of Graphics Pipeline Library where we need to grab the `SHADER_MODULE_STATE` later

I guess going back to your original question, I think this could be `unique_ptr` now, because thinking about it now, we are separating the classes in this PR, but the `spirv` inside `SHADER_MODULE_STATE` is very much tied together and will never outlive it... let me update this to be a unique pointer

layers/state_tracker/shader_module.h
```diff
@@ -587,3 +580,34 @@ struct SPIRV_MODULE_STATE : public BASE_NODE {
...
+        if (create_info.pCode[0] == spv::MagicNumber) {
+            spirv = std::make_unique<SPIRV_MODULE_STATE>(create_info.codeSize, create_info.pCode, handle_);
+        }
+    }
+
+    // If null, means this is a empty object and no shader backing it
+    std::unique_ptr<SPIRV_MODULE_STATE> spirv;
+
+    // Used as way to match instrumented GPU-AV shader to a VkShaderModule handle
+    uint32_t gpu_validation_shader_id = 0;
```
arno-lunarg, line 598: Maybe use some enum/static variable instead of 0?

layers/state_tracker/shader_module.h
artem-lunarg: Thanks, it that's possible it would be nice. Dodges some atomics.

artem-lunarg (approved):

arno-lunarg (approved):

layers/state_tracker/shader_module.h
spencer-lunarg: yep, discussed offline that will do in next PR as that will be fixing how GPU-AV interacts with both these structs, this was just to keep things working as expected while refactoring these classes out

Issue #5552, opened 2023-03-31
Issue with CoreChecks::TransitionSubpassLayouts when working with 3D or 2D array images


mbechard:
**Environment:**
 - OS: Windows 10
 - GPU: RTX 6000 Ada
 - SDK or header version if building from repo: 1.3.243.0 (downloaded from Web)
 - Options enabled (synchronization, best practices, etc.): 

**Describe the Issue**

When doing a renderpass with an attachment for 3D or 2D array image with an imageview created using a SubresourceRange of a subset of the image's layers (say, only 1 layer), the layout transition tracking in CoreChecks::TransitionSubpassLayouts is only applied to the layers listed by the SubresourceRange. But, the Vulkan spec states:

`When the VkImageSubresourceRange structure is used to select a subset of the slices of a 3D image’s mip level in order to create a 2D or 2D array image view of a 3D image created with VK_IMAGE_CREATE_2D_ARRAY_COMPATIBLE_BIT, baseArrayLayer and layerCount specify the first slice index and the number of slices to include in the created image view. Such an image view can be used as a framebuffer attachment that refers only to the specified range of slices of the selected mip level. However, any layout transitions performed on such an attachment view during a render pass instance still apply to the entire subresource referenced which includes all the slices of the selected mip level.`

**Expected behavior**

All of the layers of the image should be treated as transitioned, even if they aren't part of the SubresourceRange.

**Valid Usage ID**

UNASSIGNED-CoreValidation-DrawState-InvalidRenderpass

spencer-lunarg:
@mbechard do you have the full error message the validation layers printed out?

mbechard:
The error message is a step later, with:
`Validation Error: [ UNASSIGNED-CoreValidation-DrawState-InvalidRenderpass ] Object 0: handle = 0x2184a27d0d0, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0xa1662b9 | You cannot start a render pass using attachment 0 where the render pass initial layout is VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL and the previous known layout of the attachment is VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL. The layouts must match, or the render pass initial layout for the attachment must be VK_IMAGE_LAYOUT_UNDEFINED`

[...]

PR #6198, opened 2023-07-26
tests: ImageLayoutTransitionOf3dImageWith2dViews

arno-lunarg:
Add positive test `ImageLayoutTransitionOf3dImageWith2dViews` to try to reproduce issue https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5552
But this new test does produce any validation error.
It does the following:
- Create a 3D image of depth 2
- Create 2D views, one for each depth level
- Create a render pass using the first image view to perform a layout transition from `VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL` to `VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL`
- Create a second render pass using the second image view with an initial layout of `VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL` => this is where I was expecting `UNASSIGNED-CoreValidation-DrawState-InvalidRenderpass` to be reported. But its seems layout transition is correctly tracked for the second image view.

### DIFF
--- a/tests/unit/render_pass_positive.cpp
+++ b/tests/unit/render_pass_positive.cpp
@@ -1432,6 +1432,148 @@ TEST_F(PositiveRenderPass, FramebufferWithAttachmentsTo3DImageMultipleSubpasses)
     m_commandBuffer->end();
 }
 
+TEST_F(PositiveRenderPass, ImageLayoutTransitionOf3dImageWith2dViews) {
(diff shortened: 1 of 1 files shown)


spencer-lunarg (approved):
LGTM, happy to have extra test

PR #6199, opened 2023-07-26
Add Not Active query check and label unassigned Query VUIDs

spencer-lunarg:
As a first step for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6107

This labels the VUIDs for the 2 UNASSIGNED VUs for queries

Also adds `VUID-vkCmdResetQueryPool-None-02841` and `VUID-vkCmdCopyQueryPoolResults-None-07429`

### DIFF
--- a/layers/core_checks/cc_query.cpp
+++ b/layers/core_checks/cc_query.cpp
@@ -848,6 +848,20 @@ bool CoreChecks::ValidateQueryPoolIndex(const QUERY_POOL_STATE &query_pool_state
     return skip;
 }
 
+bool CoreChecks::ValidateQueryPoolNotActive(const CMD_BUFFER_STATE &cb_state, VkQueryPool queryPool, uint32_t firstQuery,
+                                            uint32_t queryCount, const char *func_name, const char *vuid) const {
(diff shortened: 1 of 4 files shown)


layers/core_checks/cc_query.cpp
```diff
@@ -844,6 +848,20 @@ bool CoreChecks::ValidateQueryPoolIndex(const QUERY_POOL_STATE &query_pool_state
     return skip;
 }
 
+bool CoreChecks::ValidateQueryPoolNotActive(const CMD_BUFFER_STATE &cb_state, VkQueryPool queryPool, uint32_t firstQuery,
```
artem-lunarg, line 851: Name suggestion: `ValidateQueriesNotActive`, since it checks _subset_ of query pool queries

artem-lunarg (approved):
LGTM. One naming suggestion.

layers/core_checks/cc_query.cpp
spencer-lunarg: agree, naming is hard, will fix

Issue #5465, opened 2023-03-20
Handle Descriptor Indexing and getting Constant value by Id in Shader Module


spencer-lunarg:
if using `VK_EXT_descriptor_indexing` we can't know the static size of indexing into an array. The `GetConstantValueById` currently returns `1` which is not a useful value.

An example of this being an issue is CTS test `dEQP-VK.binding_model.descriptorset_random.sets4.unifindexed.noubo.sbolimitlow.sampledimglow.outimgtexlow.noiub.nouab.frag.ialimitlow.1` which produces the following SPIR-V (warpped in .txt for GitHub to let me upload)
[input_bug.spv.txt](https://github.com/KhronosGroup/Vulkan-ValidationLayers/files/11013444/input_bug.spv.txt)

![image](https://user-images.githubusercontent.com/115671160/226240800-1bee8b1c-c0b9-4131-8587-c899afdf6a9a.png)

The `OpAccessChain`'s Index `0` is another `OpLoad` which can't be resolved

PR #6200, opened 2023-07-26
tests: Dedicated file for Descriptor Indexing

spencer-lunarg:
I have future tests for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5465 and decided to just make this its own PR as might be a bit before I get the issue done


tests/framework/layer_validation_tests.h
```diff
@@ -339,6 +339,14 @@ class DescriptorBufferTest : public VkLayerTest {
 class NegativeDescriptorBuffer : public DescriptorBufferTest {};
 class PositiveDescriptorBuffer : public DescriptorBufferTest {};
 
+class DescriptorIndexingTest : public VkLayerTest {
+  public:
+    void InitBasicDescriptorIndexing(void *pNextFeatures = nullptr);
+    VkPhysicalDeviceDescriptorIndexingFeatures descriptor_indexing_features;
```
spencer-lunarg, line 345: there is no "main feature" for descriptor indexing, so each test will need to check which of the many features it wants to make sure are enabled

artem-lunarg (approved):

PR #6203, opened 2023-07-26
layers: Add Zero RenderArea checks

spencer-lunarg:
Adds `VUID-VkRenderingInfo-None-08994` and `VUID-VkRenderingInfo-None-08995` which were added in the 1.3.257 spec


layers/core_checks/cc_render_pass.cpp
```diff
@@ -399,88 +399,86 @@ bool CoreChecks::ValidateCmdBeginRenderPass(VkCommandBuffer commandBuffer, Rende
     auto cb_state = GetRead<CMD_BUFFER_STATE>(commandBuffer);
     const char *function_name = CommandTypeString(cmd_type);
     assert(cb_state);
-    if (pRenderPassBegin) {
```
spencer-lunarg, line 402: 1. We were dereferencing  this right after with `LvlFindInChain<VkDeviceGroupRenderPassBeginInfo>(pRenderPassBegin->pNext);`
2. I added tests, this can't be null, it is caught before

layers/core_checks/cc_render_pass.cpp
```diff
@@ -399,88 +399,86 @@ bool CoreChecks::ValidateCmdBeginRenderPass(VkCommandBuffer commandBuffer, Rende
...
+        // Handle extension struct from EXT_sample_locations
+        const VkRenderPassSampleLocationsBeginInfoEXT *sample_locations_begin_info =
+            LvlFindInChain<VkRenderPassSampleLocationsBeginInfoEXT>(pRenderPassBegin->pNext);
+        if (sample_locations_begin_info) {
+            for (uint32_t i = 0; i < sample_locations_begin_info->attachmentInitialSampleLocationsCount; ++i) {
+                const VkAttachmentSampleLocationsEXT &sample_location =
+                    sample_locations_begin_info->pAttachmentInitialSampleLocations[i];
+                skip |= ValidateSampleLocationsInfo(&sample_location.sampleLocationsInfo, function_name);
+                if (sample_location.attachmentIndex >= rp_state->createInfo.attachmentCount) {
+                    const LogObjectList objlist(commandBuffer, pRenderPassBegin->renderPass);
```
spencer-lunarg, line 417: while here, updated these to use `LogObjectList`

layers/core_checks/cc_render_pass.cpp
```diff
@@ -495,6 +493,17 @@ bool CoreChecks::ValidateCmdBeginRenderPass(VkCommandBuffer commandBuffer, Rende
                              function_name, chained_device_group_struct->deviceRenderAreaCount, physical_device_count);
         }
     }
+
+    if (!chained_device_group_struct || chained_device_group_struct->deviceRenderAreaCount == 0) {
+        if (pRenderPassBegin->renderArea.extent.width == 0) {
+            skip |= LogError(commandBuffer, "VUID-VkRenderPassBeginInfo-None-08996", "%s: renderArea.extent.width is zero.",
+                             function_name);
+        } else if (pRenderPassBegin->renderArea.extent.height == 0) {
```
artem-lunarg, line 501: Can it be just `if` here?

layers/core_checks/cc_render_pass.cpp
```diff
@@ -495,6 +493,17 @@ bool CoreChecks::ValidateCmdBeginRenderPass(VkCommandBuffer commandBuffer, Rende
                              function_name, chained_device_group_struct->deviceRenderAreaCount, physical_device_count);
         }
     }
+
+    if (!chained_device_group_struct || chained_device_group_struct->deviceRenderAreaCount == 0) {
```
artem-lunarg, line 497: It's interesting that some other vuids have this part `If the pNext chain does not contain VkDeviceGroupRenderPassBeginInfo or..` but not 08996 and 08997. I assume it's inconsistency in the spec.

spencer-lunarg: its because `08996` were written with the new VUID flow, before we had to make 2 VUID, one with and without `VK_KHR_device_group` which made it worded this way, future changes might make them all worded the same as `08996`

tests/unit/imageless_framebuffer_positive.cpp
```diff
@@ -78,3 +78,103 @@ TEST_F(PositiveImagelessFramebuffer, BasicUsage) {
...
+    clear_value.color = {{0u, 0u, 0u, 0u}};
+
+    auto render_pass_attachment_bi = LvlInitStruct<VkRenderPassAttachmentBeginInfo>();
+    render_pass_attachment_bi.attachmentCount = 1;
+    render_pass_attachment_bi.pAttachments = &imageView;
+
+    auto render_pass_bi = LvlInitStruct<VkRenderPassBeginInfo>(&render_pass_attachment_bi);
+    render_pass_bi.renderPass = render_pass.handle();
+    render_pass_bi.framebuffer = framebuffer.handle();
+    render_pass_bi.renderArea.extent = {1, 1};
```
spencer-lunarg, line 172: this test was failing because it was missing the `renderArea.extent` but then realized this was a positive test actually and just moved it while here

arno-lunarg (approved):

PR #6205, opened 2023-07-26
tests: Use helper to get swapchain images

artem-lunarg:


spencer-lunarg (approved):
nice use of a helper we have already!

PR #6207, opened 2023-07-26
layers: Fix record_guard misses clean up

artem-lunarg:
If control flow takes path that does not call `EndBlockingOperation` then `record_guard` stays initialized which triggers assert.

Reset record guard in destructor to ensure clean up always happens.

External semaphores do not call `{Begin/End}BlockingOperation`, and it triggered assert in CTS tests.


Tony-LunarG (approved):
LGTM and fixes CTS asserts

PR #6208, opened 2023-07-26
Allow 32/64 bit external dependencies in the same tree

### LINKED ISSUE #6121: build is broken with 32bit and 64bit in same tree
If I clone VVL and configure both a 32bit and 64bit build in the same tree (with the same config but different build dirs), the second one will fail because all the deps are built into "Debug" and "Release" directories without specifying arch, which leads to trying to link the wrong arch's intermediate objects into the final library.

juan-lunarg:
Utilize CMAKE_SYSTEM_PROCESSOR, CMAKE_CROSSCOMPILING, and
CMAKE_SYSTEM_NAME to set default external directory when
cross-compiling.

Utilize CMAKE_SIZEOF_VOID_P otherwise.

This allows for 32/64 bit binaries to be built in the same tree.

This required updating the CI scripts to account for this new behavior. I ended up just doing general cleanup of the CI scripts while I was there.

- `build-ci/` directory will house all CI artifacts
- naming fixes/cleanup
- Drastically cleanup common_ci.py

closes #6121


scripts/CMakeLists.txt
```diff
@@ -44,9 +44,12 @@ if (UPDATE_DEPS)
...
     set(UPDATE_DEPS_DIR_SUFFIX "${_build_type}")
-    if (ANDROID)
-        set(UPDATE_DEPS_DIR_SUFFIX "android/${UPDATE_DEPS_DIR_SUFFIX}")
-    endif()
+    if (CMAKE_CROSSCOMPILING)
+        set(UPDATE_DEPS_DIR_SUFFIX "${CMAKE_SYSTEM_NAME}/${UPDATE_DEPS_DIR_SUFFIX}/${CMAKE_SYSTEM_PROCESSOR}")
+    else()
+        math(EXPR bitness "8 * ${CMAKE_SIZEOF_VOID_P}")
+        set(UPDATE_DEPS_DIR_SUFFIX "${UPDATE_DEPS_DIR_SUFFIX}/${bitness}")
+    endif()
```
juan-lunarg, line 52: Tested on Windows:

image

scripts/CMakeLists.txt
```diff
@@ -44,9 +44,12 @@ if (UPDATE_DEPS)
     list(APPEND update_dep_command "${API_TYPE}")
 
     set(UPDATE_DEPS_DIR_SUFFIX "${_build_type}")
-    if (ANDROID)
-        set(UPDATE_DEPS_DIR_SUFFIX "android/${UPDATE_DEPS_DIR_SUFFIX}")
-    endif()
+    if (CMAKE_CROSSCOMPILING)
+        set(UPDATE_DEPS_DIR_SUFFIX "${CMAKE_SYSTEM_NAME}/${UPDATE_DEPS_DIR_SUFFIX}/${CMAKE_SYSTEM_PROCESSOR}")
```
juan-lunarg, line 48: Verified this works properly for Android builds.

image

.github/workflows/vvl.yml
```diff
@@ -133,27 +131,24 @@ jobs:
...
-            external/${{ matrix.config }}/Vulkan-Headers/build/install
-            external/${{ matrix.config }}/SPIRV-Headers/build/install
-            external/${{ matrix.config }}/SPIRV-Tools/build/install
-            external/${{ matrix.config }}/robin-hood-hashing/build/install
-            external/${{ matrix.config }}/googletest/build/install
-            external/${{ matrix.config }}/Vulkan-Headers/registry
-            external/${{ matrix.config }}/SPIRV-Headers/include
-            external/${{ matrix.config }}/mimalloc/build/install
-          key: ${{ matrix.os }}-build-${{ matrix.arch }}-${{ matrix.os }}-${{ matrix.config }}-${{ hashfiles('scripts/known_good.json') }}
+            build-ci/external/glslang/build/install
```
artem-lunarg, line 140: This removes mimalloc and adds SPIRV-Reflect. Is the latter for spirv hopper?

ncesario-lunarg (approved):
This is awesome. Thanks @juan-lunarg!

artem-lunarg (approved):
LGTM. Nice update!

.github/workflows/vvl.yml
juan-lunarg: Thanks for catching that. When I was cleaning this on my linux machine I was using alphabetical ordering.

juan-lunarg: Also yes I added SPIRV-Reflect since it's part of known_good.json

juan-lunarg: Added mimalloc

spencer-lunarg (approved):

PR #6218, opened 2023-07-27
ci: Update Vulkan-Loader library unloading option

charles-lunarg:
The loader recently updated how the dynamic library unloading behavior is enabled, from a build option to an environment variable set at startup. this commit updates common_ci.py to respect that change.

### DIFF
--- a/scripts/common_ci.py
+++ b/scripts/common_ci.py
@@ -140,9 +140,6 @@ def BuildLoader():
     cmake_cmd = f'cmake -S {SRC_DIR} -B {BUILD_DIR}'
     cmake_cmd += ' -D UPDATE_DEPS=ON -D CMAKE_BUILD_TYPE=Release'
 
-    # This enables better stack traces from tools like leak sanitizer by using the loader feature which prevents unloading of libraries at shutdown.
-    cmake_cmd += ' -D LOADER_DISABLE_DYNAMIC_LIBRARY_UNLOADING=ON'
-
     # GitHub actions runs our test as admin on Windows
     if IsGHA() and IsWindows():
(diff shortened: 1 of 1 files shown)


scripts/common_ci.py
```diff
@@ -225,6 +222,9 @@ def RunVVLTests():
         lvt_env['VK_LAYER_PATH'] = os.path.join(CI_INSTALL_DIR, 'share/vulkan/explicit_layer.d')
         lvt_env['VK_DRIVER_FILES'] = os.path.join(CI_INSTALL_DIR, 'share/vulkan/icd.d/VkICD_mock_icd.json')
 
+    # This enables better stack traces from tools like leak sanitizer by using the loader feature which prevents unloading of libraries at shutdown.
+    lvt_env['VK_LOADER_DISABLE_DYNAMIC_LIBRARY_UNLOADING'] = '1'
```
juan-lunarg, line 226: This scripts check out ToT main for the loader currently. You need to modify the git checkout command for the loader.

charles-lunarg: I knew I was missing something.

Granted, this PR *is* correct because the behavior will be in ToT main when the loader PR merges, so thats nice.

juan-lunarg: Yep this commit will be correct.

I'd just add another commit to point to your branch. Then revert that commit once it's merged into main.

charles-lunarg: We've merged in the Loader PR, so because this PR's CI hasn't started for the most part, when it *does* run it will run with the loader changes, thus testing it.

juan-lunarg (approved):

PR #6221, opened 2023-07-28
Eliminate two FormatHandle verbosities that have annoyed

jzulauf-lunarg:
DRY up some repeated code.  With SFINAE fun.

### DIFF
--- a/layers/core_checks/cc_query.cpp
+++ b/layers/core_checks/cc_query.cpp
@@ -65,7 +65,7 @@ bool CoreChecks::ValidatePerformanceQueryResults(const char *cmd_name, const QUE
                                                                         : "VUID-vkCmdCopyQueryPoolResults-queryType-03233",
                          "%s: QueryPool %s was created with a queryType of"
                          "VK_QUERY_TYPE_PERFORMANCE_QUERY_KHR but flags contains %s.",
-                         cmd_name, report_data->FormatHandle(query_pool_state.pool()).c_str(), invalid_flags_string.c_str());
+                         cmd_name, FormatHandle(query_pool_state.pool()).c_str(), invalid_flags_string.c_str());
     }
 
     for (uint32_t query_index = firstQuery; query_index < queryCount; query_index++) {
@@ -77,11 +77,11 @@ bool CoreChecks::ValidatePerformanceQueryResults(const char *cmd_name, const QUE
             }
         }
         if (submitted < query_pool_state.n_performance_passes) {
-            skip |= LogError(query_pool_state.pool(), "VUID-vkGetQueryPoolResults-queryType-03231",
-                             "%s: QueryPool %s has %u performance query passes, but the query has only been "
-                             "submitted for %u of the passes.",
-                             cmd_name, report_data->FormatHandle(query_pool_state.pool()).c_str(),
-                             query_pool_state.n_performance_passes, submitted);
+            skip |=
+                LogError(query_pool_state.pool(), "VUID-vkGetQueryPoolResults-queryType-03231",
+                         "%s: QueryPool %s has %u performance query passes, but the query has only been "
+                         "submitted for %u of the passes.",
+                         cmd_name, FormatHandle(query_pool_state.pool()).c_str(), query_pool_state.n_performance_passes, submitted);
         }
     }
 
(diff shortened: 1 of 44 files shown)


layers/error_message/logging.h
```diff
@@ -229,9 +241,11 @@ typedef struct _debug_report_data {
         return FormatHandle(object_string[handle.type], handle.handle);
     }
 
-    template <typename HANDLE_T>
-    std::string FormatHandle(HANDLE_T handle) const {
-        return FormatHandle(VkHandleInfo<HANDLE_T>::Typename(), HandleToUint64(handle));
+    std::string FormatHandle(const TypedHandleWrapper &wrapper) const { return FormatHandle(wrapper.Handle()); }
+
+    template <typename T, typename std::enable_if_t<!std::is_base_of<TypedHandleWrapper, T>::value, void *> = nullptr>
```
jzulauf-lunarg, line 246: Note the SFINAE fun to ensure the the TypedHandleWrapper will be selected for derived classes of the mixin

layers/error_message/logging.h
```diff
@@ -160,6 +160,18 @@ struct LoggingLabelState {
     }
 };
 
+class TypedHandleWrapper {
```
jzulauf-lunarg, line 163: Use this mixin for any class you want to auto support Handle fetches.  There are a couple of other places in the code it could be done.

layers/vulkan/generated/chassis.h
```diff
@@ -3776,6 +3776,9 @@ class ValidationObject {
     public:
         APIVersion api_version;
         debug_report_data* report_data = nullptr;
+        template <typename T>
```
jzulauf-lunarg, line 3779: It's a pure forwarding convenience function, which steers itself to the appropriate FormatHandle overload.

jzulauf-lunarg (commented):
Two changes:

1. Eliminating the need for report_data-> everywhere
2. Eliminate the need for BASE_NODE derived object to extract the VulkanTypedHandle everywhere

spencer-lunarg (approved):
LGTM, happy to clean up error messages

arno-lunarg (approved):

layers/core_checks/cc_query.cpp
```diff
@@ -728,21 +719,19 @@ void CoreChecks::PreCallRecordCmdBeginQuery(VkCommandBuffer commandBuffer, VkQue
...
-                                                   VkQueryPool &firstPerfQueryPool, uint32_t perfPass,
-                                                   QueryMap *localQueryToStateMap) {
+    cb_state.queryUpdates.emplace_back([this, query_obj](CMD_BUFFER_STATE &cb_state_arg, bool do_validate,
+                                                         VkQueryPool &firstPerfQueryPool, uint32_t perfPass,
+                                                         QueryMap *localQueryToStateMap) {
         if (!do_validate) return false;
         bool skip = false;
-        auto device_data = cb_state_arg.dev_data;
-        auto query_pool_state = device_data->Get<QUERY_POOL_STATE>(query_obj.pool);
+        auto query_pool_state = Get<QUERY_POOL_STATE>(query_obj.pool);
```
spencer-lunarg, line 727: internal 
CI is failing here on Window builds

> cc_query.cpp(727,1): error C2668: 'ValidationStateTracker::Get': ambiguous call to overloaded function 
cc_query.cpp(728,1): error C3536: 'query_pool_state': cannot be used before it is initialized

layers/vulkan/generated/chassis.h
```diff
@@ -3776,6 +3776,9 @@ class ValidationObject {
...
+        std::string FormatHandle(T&& h) const { return report_data->FormatHandle(std::forward<T>(h)); }
```
artem-lunarg, line 3780: Not sure if rvalue and forwarding is needed here. The signature is `std::string FormatHandle(const TypedHandleWrapper &wrapper)`.

`TypedHandleWrapper` is also a mixin, so often part of the base object. We probably don't expect that reporting routing will move something out of the base object, if I understand this correctly.

artem-lunarg (approved):
LGTM. Nice clean up, `report_data` always looked for me out of place.
One comment about rvalue parameter.

layers/vulkan/generated/chassis.h
jzulauf-lunarg: @artem-lunarg -- there are several underlying overloads including a templated one... so the && universal reference is needed. -- https://levelup.gitconnected.com/perfect-forwarding-647e1caaf879

artem-lunarg: Universal reference can be evaluated either to rvalue or lvalue reference. Do we have two classes of calls that use either lvalue or rvalue references?

jzulauf-lunarg: Or pass by value where the argument is a literal (and thus an rvalue?). In any case, Vk handles passed directly failed until I made it universal. The other choice would have been having three forwarding functions matching the overloads.

` std::string FormatHandle(const VulkanTypedHandle &handle)`
`std::string FormatHandle(const TypedHandleWrapper &wrapper)`
` template <typename T, typename std::enable_if_t<!std::is_base_of<TypedHandleWrapper, T>::value, void *> = nullptr>
    std::string FormatHandle(T handle) const `

PR #6222, opened 2023-07-28
GPU-AV Filename cleanup

spencer-lunarg:
The `gpu_utils` file is not helpful for understanding what was actually in it

This change breaks it into a `gpu_error_message` and `gpu_state_tracker` file

A follow up would to create a better namespace/naming scheme for the contents inside rather then just `Util`


Tony-LunarG (approved):
Pulling out the error message code into it's own file is a nice cleanup.  It seems like state_tracker.cpp does a lot more than track state as it allocates resources, initializes VMA, etc, but I don't have a better suggestion.

Issue #6224, opened 2023-07-28
Non SPIR-V dependent validation no longer executed when SPIR-V module data is not available since recent refactoring


aqnuep:
**Environment:**
 - OS: All
 - GPU: All
 - SDK or header version if building from repo: latest main
 - Options enabled (synchronization, best practices, etc.):

**Describe the Issue**

SPIR-V module data has been moved to a new `SPIRV_MODULE_STATE` object, but that object is not created for cases when a SPIR-V module is not available, e.g. if shader module identifier is used (see `PIPELINE_STATE::GetStageStates()` code creating placeholder module state for this case).

Also, validation code previously accepting `SHADER_MODULE_STATE` objects now take `SPRIV_MODULE_STATE` objects, even when they don't rely on any SPIR-V state (e.g. `ValidateSpecializations` or `ValidateShaderSubgroupSizeControl`).

As a result, `ValidatePipelineShaderStage` will not run certain validation for cases that could previously be validated due to the reliance on the existence of a `SPIRV_MODULE_STATE` object:

```c++
    if (pipeline.uses_shader_module_id || !stage_state.module_state->spirv) {
        return skip;  // these edge cases should be validated already
    }
```

Thus validation coverage has been lost in cases where SPIR-V data is unavailable (like in case of the use of shader module indentifiers), and non-SPIR-V validation code was made SPIR-V dependent.

**Expected behavior**

Validation that does not require SPIR-V data should execute even when such data is not present, like it happened before the recent refactoring. Thus technically this is a regression that isn't caught by any test case.

**Valid Usage ID**

The following VUIDs (not a complete list) which do not depend on SPIR-V data are no longer generated when SPIR-V data is not available:

  * `VUID-VkSpecializationInfo-offset-00773`
  * `VUID-VkSpecializationInfo-pMapEntries-00774`
  * `VUID-VkSpecializationInfo-constantID-04911`
  * `VUID-VkPipelineShaderStageCreateInfo-flags-02784`
  * `VUID-VkPipelineShaderStageCreateInfo-flags-02785`
  * `VUID-VkPipelineShaderStageCreateInfo-flags-08988`

There could be more.

spencer-lunarg:
>  Thus technically this is a regression that isn't caught by any test case.

... that's unfortunate, did you hit this from a real world app or just from reviewing the code

I will get some tests together and fix this (since I made the change so we can get `VK_EXT_shader_object` hooked up)

[...]

PR #6226, opened 2023-07-31
Fix VK_EXT_shader_module_identifier checks for VkPipelineShaderStageCreateInfo

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6224 (cc @aqnuep)

When we validate `VkPipelineShaderStageCreateInfo` struct, we need to first do all the validation that doesn't require any actual SPIR-V. I went through each function and found the ones that were not actually using the `SPIRV_MODULE_STATE` object (or the old `SHADER_MODULE_STATE` for that matter)

There are now tests that hit these with a `VK_EXT_shader_module_identifier` so this won't regress again


aqnuep:
Thanks, @spencer-lunarg, this approach looks reasonable.

Maybe one thing I'd note is that it may worth considering to separate out the parts of `ValidatePipelineShaderStage` that do not depend on SPIR-V vs that do beyond just putting them before/after the check for an existing SPIR-V state object. It's a nit, but may also help preventing future refactoring to accidentally move certain checks around (although, sure, for these cases the new tests you've added would already enable detecting a regression).

spencer-lunarg:
> worth considering to separate out the parts of ValidatePipelineShaderStage

I was actually going to, but the whole `ValidatePipelineShaderStage` is going to soon be broken up into "need pipeline info" and "need shader info" and some of these so figured that will be when I move these. Also while `VK_EXT_shader_object` doesn't interact with `VK_EXT_shader_module_identifier` to my knowledge, there is some overlap with `ValidateSpecializations`, but not with `ValidatePipelineRobustnessCreateInfo` so will need to see if it makes sense later to group that or not as well

aqnuep:
Makes sense, thanks again for the quick fix!

layers/core_checks/cc_spirv.cpp
```diff
@@ -2398,11 +2172,24 @@ bool CoreChecks::ValidateShaderTileImage(const SPIRV_MODULE_STATE &module_state,
...
 
+// Function to get the VkPipelineShaderStageCreateInfo from the various pipeline types
 bool CoreChecks::ValidatePipelineShaderStage(const PIPELINE_STATE &pipeline, const PipelineStageState &stage_state) const {
     bool skip = false;
     const auto *create_info = stage_state.create_info;
     const VkShaderStageFlagBits stage = create_info->stage;
 
+    // First validate all things that don't require valid SPIR-V
+    // this is found when using VK_EXT_shader_module_identifier
+    skip |= ValidateShaderSubgroupSizeControl(pipeline, stage, create_info->flags);
```
artem-lunarg, line 2183: Previously we checked for `if (IsExtEnabled(device_extensions.vk_ext_subgroup_size_control))`. Do we need it here?

spencer-lunarg: We had a lot of `IsExtEnabled` I don't think we really needed, you could argue that it might be faster, but we don't guard around extensions be enabled, the function is not that complex, and it was misleading to me thinking we "needed" it (or else things would crash)... so I rather just have it removed (plus it was promoted to core, so its not a random extension)

artem-lunarg (approved):

Issue #6204, opened 2023-07-26
Unexpected VUID-vkCmdPipelineBarrier-None-07889


rg3igalia:
**Environment:**
 - OS: Linux
 - GPU: RTX 4060 Ti
 - SDK or header version if building from repo: 3b7995d6ba828baac79755fcf9b5ab4489731209
 - Options enabled (synchronization, best practices, etc.): defaults + synchronization

**Describe the Issue**

When using a pipeline barrier inside a render pass, one must provide a subpass self-dependency with a superset of pipeline stages and accesses. The layers verify this in [RenderPassDepState::ValidateStage](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/3b7995d6ba828baac79755fcf9b5ab4489731209/layers/core_checks/cc_synchronization.cpp#L705).

If one of the stages is VK_PIPELINE_STAGE_VERTEX_INPUT_BIT, the calls to [sync_utils::ExpandPipelineStages](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/3b7995d6ba828baac79755fcf9b5ab4489731209/layers/sync/sync_utils.cpp#L81) remove that bit from the mask and replace it with (VK_PIPELINE_STAGE_2_INDEX_INPUT_BIT_KHR | VK_PIPELINE_STAGE_2_VERTEX_ATTRIBUTE_INPUT_BIT_KHR), but they only do that with the subpass masks. If the same bit is used in the pipeline barrier, this is not expanded and the layers detect an apparent mismatch, generating VUID-vkCmdPipelineBarrier-None-07889 erroneously.

Possible fix below. Not sure if that's the right approach.

```diff
diff --git a/layers/core_checks/cc_synchronization.cpp b/layers/core_checks/cc_synchronization.cpp
index 63c089e2c..1be4c2c08 100644
--- a/layers/core_checks/cc_synchronization.cpp
+++ b/layers/core_checks/cc_synchronization.cpp
@@ -707,14 +707,18 @@ struct RenderPassDepState {
         bool match = false;
         for (const auto self_dep_index : self_dependencies) {
             const auto sub_dep = GetSubPassDepBarrier(dependencies[self_dep_index]);
-            auto sub_src_stage_mask =
+            const auto sub_src_stage_mask =
                 sync_utils::ExpandPipelineStages(sub_dep.srcStageMask, sync_utils::kAllQueueTypes, disabled_features);
[...]

PR #6228, opened 2023-07-31
layers: Fix RenderPassDepState stage subset test and cleanup

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6204

To review the fix check a specific commit, the PR makes other changes so general diff is not helpful.

Other changes:
* Clean up of `RenderPassDepState`
* Reworked messages: 
** Print flag names instead of hex constants
** Single message instead of two with the same information and hopefully easier to understand
** Do not print *dot* at the beginning of the variable name `.srcStageMask` if there is no parent
** Do not print _candidate_ information. That's prediction and it might be wrong, but it's easy to check during debugging session. On the other hand makes message more complicated.


layers/core_checks/cc_synchronization.cpp
```diff
@@ -657,11 +657,8 @@ bool CoreChecks::PreCallValidateCmdResetEvent2(VkCommandBuffer commandBuffer, Vk
     return ValidateCmdResetEvent2(commandBuffer, event, stageMask, CMD_RESETEVENT2);
 }
 
-// transient helper struct for checking parts of VUID 02285
```
artem-lunarg, line 660: this VUID is not anymore

layers/core_checks/cc_synchronization.cpp
```diff
@@ -685,103 +682,77 @@ struct RenderPassDepState {
...
-        const auto *barrier = LvlFindInChain<VkMemoryBarrier2KHR>(dep.pNext);
-        if (barrier) {
-            result = *barrier;
-        } else {
-            result.srcStageMask = dep.srcStageMask;
-            result.dstStageMask = dep.dstStageMask;
-            result.srcAccessMask = dep.srcAccessMask;
-            result.dstAccessMask = dep.dstAccessMask;
+    VkMemoryBarrier2 GetSubPassDepBarrier(const safe_VkSubpassDependency2 &dep) const {
+        // "If a VkMemoryBarrier2 is included in the pNext chain, srcStageMask, dstStageMask,
```
spencer-lunarg, line 686: we have this logic all over the code, its a nasty thing not obvious at first, not sure if good way in future to try and centralize this logic

spencer-lunarg (approved):

Issue #2835, opened 2021-05-25
Validate PSOs follow maxFragmentInputComponents et. al.


darksylinc:
**Describe the situation in which you encountered the missing validation**
Issue description:

I [found this issue](https://gitlab.freedesktop.org/mesa/mesa/-/issues/4524) by accident:

Overflowing number of varyings passed to pixel shader results in GPU crash

The following VS and PS will cause a GPU hang when using RADV. In OpenGL it correctly refuses to link this shader citing that the 'array is too large':

```
// VS
layout( location = 0 ) in float4 inputVsPos;

layout( location = 0 ) out block
{
float4 a0;float4 a1;float4 a2;float4 a3;float4 a4;float4 a5;float4 a6;float4 a7;float4 a8;float4 a9;float4 a10;
float4 a11;float4 a12;float4 a13;float4 a14;float4 a15;float4 a16;float4 a17;float4 a18;float4 a19;float4 a20;
float4 a21;float4 a22;float4 a23;float4 a24;float4 a25;float4 a26;float4 a27;float4 a28;float4 a29;float4 a30;
float4 a31;float4 a32;float4 a33;float4 a34;float4 a35;float4 a36;float4 a37;float4 a38;float4 a39;float4 a40;
float4 a41;float4 a42;float4 a43;float4 a44;float4 a45;float4 a46;float4 a47;float4 a48;float4 a49;
} outVs;

void main()
{
	gl_Position.xyzw = inputVsPos;
	outVs.a0 = inputVsPos; outVs.a1 = inputVsPos; outVs.a2 = inputVsPos; outVs.a3 = inputVsPos; outVs.a4 = inputVsPos; outVs.a5 = inputVsPos;
	outVs.a6 = inputVsPos; outVs.a7 = inputVsPos; outVs.a8 = inputVsPos; outVs.a9 = inputVsPos; outVs.a10 = inputVsPos;
	outVs.a11 = inputVsPos; outVs.a12 = inputVsPos; outVs.a13 = inputVsPos; outVs.a14 = inputVsPos; outVs.a15 = inputVsPos;
	outVs.a16 = inputVsPos; outVs.a17 = inputVsPos; outVs.a18 = inputVsPos; outVs.a19 = inputVsPos; outVs.a20 = inputVsPos;
[...]

PR #6230, opened 2023-08-01
layers: Move SPIRV_MODULE_STATE creation to PreRecord

spencer-lunarg:
This reworks the chassis for `vkCreateShaderModule` and `vkCreateShadersEXT` so it can after

1. Validate the SPIR-V without needing pipeline state (to solve issues like https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/2835)
2. Easier to add support for `VK_EXT_shader_object` in both Validation and GPU-AV (making separate PR as needs many tests added)

Overview of flow

spirv

### DIFF
--- a/docs/shader_validation.md
+++ b/docs/shader_validation.md
@@ -23,7 +23,7 @@ Currently these are
   - This `spirv-opt` pass is used to inject the constants from the pipeline layout.
   - Some checks require the runtime spec constant values
 - Flatten OpGroupDecorations
-- Detects if group decorations were used; however, group decorations were [deprecated](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpGroupDecorate) early on in the development of the SPIR-v specification.
+  - Detects if group decorations were used; however, group decorations were [deprecated](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpGroupDecorate) early on in the development of the SPIR-v specification.
 - Debug Printf
   - instruments the shaders - see [GPU-Assisted Validation](gpu_validation.md).
 - GPU-VA
@@ -48,7 +48,7 @@ The code is currently split up into the following main sections
 
 All Shader Validation can be broken into 4 types of checks
 
-- Standalone with runtime properties
+- SPIR-V with runtime properties
   - Things like features and limits
 - Shader interface
   - Ex. going between a Vertex and Fragment shader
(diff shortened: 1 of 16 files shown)


docs/shader_validation.md
```diff
@@ -65,32 +65,36 @@ When dealing with shader validation there are a few concepts to understand and n
   - Tied to a shader stage (fragment, vertex, etc)
   - Knows which variables and instructions are touched in stage
     - There might be things in a `ShaderModule` not related to shader stage validation
-- `Shader Module`
-  - a `VkShaderModule` object
-  - contains all information about the SPIR-V module
+- `SPIR-V Module`
+  - `SPIRV_MODULE_STATE`
+  - This object takes in SPIR-V, parses it, creates `EntryPoint` objects, validates what we can
```
artem-lunarg, line 70: Currently it's formatted like below, is it the intention? If yes, then transition from `what we can` to nested statements is a bit unclear.
  - This object takes in SPIR-V, parses it, creates `EntryPoint` objects, validates what we can
    - We do validation first as sometimes a bad SPIR-V can [crash a driver](https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/2835)
    - can contain [multiple EntryPoints](https://github.com/KhronosGroup/SPIRV-Guide/blob/master/chapters/entry_execution.md#instructions-with-multiple-execution-modes)

layers/state_tracker/state_tracker.h
```diff
@@ -79,7 +80,14 @@ enum AcquireVersion { ACQUIRE_VERSION_1 = 0, ACQUIRE_VERSION_2 = 1 };
 
 // This structure is used modify and pass parameters for the CreateShaderModule down-chain API call
 struct create_shader_module_api_state {
-    uint32_t unique_shader_id;
+    // We build a SPIRV_MODULE_STATE at PreCallValidate time were we can do basic validation of the SPIR-V (which can crash drivers
```
artem-lunarg, line 83: Great comment. Adds a lot to understanding how different parts are connected.

layers/core_checks/cc_spirv.cpp
```diff
@@ -2699,6 +2699,26 @@ static ValidationCache *GetValidationCacheInfo(VkShaderModuleCreateInfo const *p
     return nullptr;
 }
 
+void CoreChecks::PreCallRecordCreateShaderModule(VkDevice device, const VkShaderModuleCreateInfo *pCreateInfo,
+                                                 const VkAllocationCallbacks *pAllocator, VkShaderModule *pShaderModule,
+                                                 void *csm_state_data) {
+    // Normally would validate in PreCallValidate, but need a non-const function to update csm_state
+    ValidationStateTracker::PreCallRecordCreateShaderModule(device, pCreateInfo, pAllocator, pShaderModule, csm_state_data);
+    create_shader_module_api_state *csm_state = reinterpret_cast<create_shader_module_api_state *>(csm_state_data);
```
artem-lunarg, line 2707: `static_cast` is designed to convert from `void*` to `Type*`

spencer-lunarg: ok, happy to change this, I just copied and pasted how we were doing other places, but agree they could all be `static_cast`

layers/state_tracker/state_tracker.h
artem-lunarg, line 83: `We build a SPIRV_MODULE_STATE at PreCallValidate time` - could you confirm if this still true?

spencer-lunarg: in `ValidationStateTracker::PreCallRecordCreateShaderModule`

we call `std::make_unique<SPIRV_MODULE_STATE>` to build it

spencer-lunarg: awww... sorry, yes, this is now `PreCallRecord`.. .will update

layers/core_checks/cc_spirv.cpp
```diff
@@ -2699,6 +2699,26 @@ static ValidationCache *GetValidationCacheInfo(VkShaderModuleCreateInfo const *p
     return nullptr;
 }
 
+void CoreChecks::PreCallRecordCreateShaderModule(VkDevice device, const VkShaderModuleCreateInfo *pCreateInfo,
+                                                 const VkAllocationCallbacks *pAllocator, VkShaderModule *pShaderModule,
+                                                 void *csm_state_data) {
+    // Normally would validate in PreCallValidate, but need a non-const function to update csm_state
```
artem-lunarg, line 2705: It looks that `create_shader_module_api_state` objects are created on the stack, and in that case they are not used by multiple threads. If that's true and no other data transfer that I missed,  then it might be fine to use `PreCallValidate` to modify this cms state (with `const_cast`).

The point that there is a chance we don't have a threading hazard here. I'm totally fine with how it's done currently, but if for some reason we need a change in the future to move things to Validate it might work from correctness point of view.

artem-lunarg (approved):
LGTM. Some implementation details nits.

arno-lunarg (approved):

layers/core_checks/cc_spirv.cpp
spencer-lunarg: added a comment for this

Issue #6225, opened 2023-07-28
Compile broken by SPIRV-Headers commit


romanl-g:
https://github.com/KhronosGroup/SPIRV-Headers/issues/365

romanl-g:
CooperativeMatrixOperandsMatrixASignedComponentsMask no longer a valid constant

https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/c320ff46b5a914dafe7e0512a6395ecd643de457/layers/core_checks/cc_spirv.cpp#L1000

spencer-lunarg:
cc @HuYuxin

Ok, so I was confused before when I saw this issue, I didn't realize there was a random change to the Header to change the names. I updated to ToT `SPIRV-Headers` and `SPIRV-Tools`/`SPIRV-Reflect` for completeness and seems to be good now

HuYuxin:
Thank you @spencer-lunarg for taking care of the fix!

spencer-lunarg:
Should be fixed now with https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6235

PR #6229, opened 2023-07-31
Change the name of spv cooperative matrix operands

HuYuxin:
Change the name of spv cooperative matrix operands to match with the new naming conventions in SPIRV-Headers: https://github.com/KhronosGroup/SPIRV-Headers/commit/272be321ad6ccb8bc7533c325a5faf5806e6c06f

PR #6231, opened 2023-08-01
build: Fix SPIRV-Header VK_KHR_cooperative_matrix

spencer-lunarg:
fixes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6225


build-android/known_good.json
```diff
@@ -22,13 +22,13 @@
             "name": "SPIRV-Tools",
             "url": "https://github.com/KhronosGroup/SPIRV-Tools.git",
             "sub_dir": "shaderc/third_party/spirv-tools",
-            "commit": "04cdb2d344706052c7a2d359294e830ebac63e74"
+            "commit": "e68fe9be4e6ca63097ac4305d7552ad29afd5004"
```
spencer-lunarg, line 25: this is failing in internal CI because the NDK version (known issue from the 158/159 header update)

artem-lunarg (approved):
LGTM. The change itself makes sense, but not sure about Android CI.

spencer-lunarg:
@HuYuxin @romanl-g

So our internal CI is failing because in order to bump the SPIRV-Headers, we need to bump the version of SPIRV-Tools (otherwise will get the SPIRV-Tools build errors as seen in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6229#issuecomment-1659469988)

The CI uses NDK 25 and now SPIRV-Tools fails to parse the SPIRV-Headers file and NDK 25 completely fails... we were planning soon (like end of week hopfully) to either fix the CI or better, just swap over to our CMake build completely

For the short term, I think I am going to just completey remove the `VK_KHR_cooperative_matrix` code in order to unblock everyone

dneto0:
SPIRV-Tools is now working correctly with latest SPIRV-Headers.
Two things happened recently:
- the change in the enum names, as you already saw.
- the coop matrix KHR capability was removed from some enums. That made those enums invisible to the SPIRV-Tools parser: the enums were not enabled by any version, extension, or capability.  In Google's CI, the symptom was a spv.coopmatrixKHR.comp Glslang test to fail validation.
     -  I've patched things up by landing a change in SPIRV-Headers to make the enums depend on the extension instead. 

Relevant changes:
- https://github.com/KhronosGroup/SPIRV-Headers/pull/367 
- https://github.com/KhronosGroup/SPIRV-Tools/pull/5361 also adds a test to make sure some of those enums are parsed and validated


So, @spencer-lunarg I think you can keep the coop matrix code if you update both SPIRV-Headers and SPIRV-Tools, but update the enum names in the VVL code.

dneto0:
I have a PR for SPIRV-Headers to avoid the "invisible" enums problem in the future.

spencer-lunarg:
Thanks @dneto0  for the info

> I think you can keep the coop matrix code if you update both SPIRV-Headers and SPIRV-Tools, but update the enum names in the VVL code.

This was (is?) blocked for a bit as we were fighting internal Android build issues... but will get everything updated and re-enable `VK_KHR_cooperative_matrix`

jeremyg-lunarg (approved):

jeremyg-lunarg:
I did a maintainer push to fix the SPIRV_TOOLS_COMMIT_ID

PR #6234, opened 2023-08-01
Tests RenderFramework and Image cleanup

spencer-lunarg:
Broken into 4 commits, but goal is to simplify the framework from things that are only used once, or have many ways to do the same thing


tests/unit/command.cpp
```diff
@@ -203,9 +203,13 @@ TEST_F(NegativeCommand, Sync2SecondaryCommandbufferAsPrimary) {
     if (!AreRequiredExtensionsEnabled()) {
         GTEST_SKIP() << RequiredExtensionsNotSupported() << " not supported";
     }
-    if (!CheckSynchronization2SupportAndInitState(this)) {
-        GTEST_SKIP() << "Synchronization2 not supported";
+    auto sync2_features = LvlInitStruct<VkPhysicalDeviceSynchronization2FeaturesKHR>();
+    GetPhysicalDeviceFeatures2(sync2_features);
+    if (!sync2_features.synchronization2) {
```
artem-lunarg, line 208: I learned recently that support of `VK_KHR_synchronization2`  extension guarantees support of `syncrhonization2` feature. The suggestion to `sync2_features.synchronization2 = VK_TRUE` without querying feature support. 

```
45.1. Feature Requirements
All Vulkan graphics implementations must support the following features:
...
synchronization2 if Vulkan 1.3 or the VK_KHR_synchronization2 extension is supported.
```

artem-lunarg: Thanks! Did not think about relation between extension and only single feature. Good insight.

tests/unit/descriptors.cpp
```diff
@@ -1021,7 +1021,7 @@ TEST_F(NegativeDescriptors, ImageDescriptorLayoutMismatch) {
     ASSERT_TRUE(image.initialized());
 
     vk_testing::ImageView view;
-    auto image_view_create_info = SafeSaneImageViewCreateInfo(image, format, VK_IMAGE_ASPECT_COLOR_BIT);
+    auto image_view_create_info = image.BasicTargetViewCreatInfo();
```
artem-lunarg, line 1024: nit: will it be more clear with target `BasicViewCreatInfo`. Not sure what `target` originally meant, `basic` sounds good though.

artem-lunarg (approved):
LGTM. sync2 suggestion which can be a separate PR too.

PR #6237, opened 2023-08-01
layers: Fix skip in shader tile image validation

ziga-lunarg:
VK_EXT_shader_tile_image validation should be skip if stage is not fragment OR if the extension is not enabled


spencer-lunarg (approved):
agree with this, but we are seriously lacking a positive test if this was happening. Can you please add a positive test with this

ziga-lunarg:
> agree with this, but we are seriously lacking a positive test if this was happening. Can you please add a positive test with this

I think you might have misunderstood the fix here, this just skip unnecessary validation that can't find anything. I can't think of a positive test that would fail before and not with the skip now.

spencer-lunarg:
> this just skip unnecessary validation that can't find anything

Awww, sorry, lets get this in

PR #6239, opened 2023-08-02
scripts: Fixes for parent and commonparent VUIDs

spencer-lunarg:
Cleanup how we generate the implicit VUs for `-parent` and `-commonparent`

notes:

- `parent` = There is a `VkDevice` (`VkInstance` is a todo) and another Handle in the same call
- `commonparent` = there are 2 Handles, that aren't `VkDevice`

### DIFF
--- a/layers/vulkan/generated/object_tracker.cpp
+++ b/layers/vulkan/generated/object_tracker.cpp
@@ -232,7 +232,7 @@ bool ObjectLifetimes::PreCallValidateQueueSubmit(
     const VkSubmitInfo*                         pSubmits,
     VkFence                                     fence) const {
     bool skip = false;
-    skip |= ValidateObject(queue, kVulkanObjectTypeQueue, false, "VUID-vkQueueSubmit-queue-parameter", "VUID-vkQueueSubmit-commonparent", "vkQueueSubmit");
+    skip |= ValidateObject(queue, kVulkanObjectTypeQueue, false, "VUID-vkQueueSubmit-queue-parameter", kVUIDUndefined, "vkQueueSubmit");
     if (pSubmits) {
         for (uint32_t index0 = 0; index0 < submitCount; ++index0) {
             if ((pSubmits[index0].waitSemaphoreCount > 0) && (pSubmits[index0].pWaitSemaphores)) {
@@ -461,7 +461,7 @@ bool ObjectLifetimes::PreCallValidateQueueBindSparse(
     const VkBindSparseInfo*                     pBindInfo,
     VkFence                                     fence) const {
     bool skip = false;
-    skip |= ValidateObject(queue, kVulkanObjectTypeQueue, false, "VUID-vkQueueBindSparse-queue-parameter", "VUID-vkQueueBindSparse-commonparent", "vkQueueBindSparse");
+    skip |= ValidateObject(queue, kVulkanObjectTypeQueue, false, "VUID-vkQueueBindSparse-queue-parameter", kVUIDUndefined, "vkQueueBindSparse");
     if (pBindInfo) {
         for (uint32_t index0 = 0; index0 < bindInfoCount; ++index0) {
             if ((pBindInfo[index0].waitSemaphoreCount > 0) && (pBindInfo[index0].pWaitSemaphores)) {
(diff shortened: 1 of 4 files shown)


layers/vulkan/generated/object_tracker.cpp
```diff
@@ -232,7 +232,7 @@ bool ObjectLifetimes::PreCallValidateQueueSubmit(
     const VkSubmitInfo*                         pSubmits,
     VkFence                                     fence) const {
     bool skip = false;
-    skip |= ValidateObject(queue, kVulkanObjectTypeQueue, false, "VUID-vkQueueSubmit-queue-parameter", "VUID-vkQueueSubmit-commonparent", "vkQueueSubmit");
```
spencer-lunarg, line 235: note, these are currently validating that the `queue` in `vkQueueSubmit` was from the same device as `queue` .... below you will find `ValidateObject(fence, ...., "VUID-vkQueueSubmit-commonparent", ...)` which is the check we wanted

layers/vulkan/generated/object_tracker.cpp
```diff
@@ -3565,7 +3565,7 @@ bool ObjectLifetimes::PreCallValidateCreateVideoSessionParametersKHR(
     bool skip = false;
     skip |= ValidateObject(device, kVulkanObjectTypeDevice, false, "VUID-vkCreateVideoSessionParametersKHR-device-parameter", kVUIDUndefined, "vkCreateVideoSessionParametersKHR");
     if (pCreateInfo) {
-        skip |= ValidateObject(pCreateInfo->videoSessionParametersTemplate, kVulkanObjectTypeVideoSessionParametersKHR, true, "VUID-VkVideoSessionParametersCreateInfoKHR-videoSessionParametersTemplate-parameter", "VUID-VkVideoSessionParametersCreateInfoKHR-videoSessionParametersTemplate-parent", "VkVideoSessionParametersCreateInfoKHR");
```
spencer-lunarg, line 3568: this is an edge case needed to fix in the spec were we really don't need a ` VUID-VkVideoSessionParametersCreateInfoKHR-videoSessionParametersTemplate-parent` because if both of these are correct, that is the same as ` VUID-VkVideoSessionParametersCreateInfoKHR-commonparent`

scripts/generators/object_tracker_generator.py
```diff
@@ -435,12 +423,25 @@ def validateObjects(self, members: list[Member], indent: str, prefix: str, array
...
-                    if parentVUID == 'kVUIDUndefined':
-                        parentVUID = self.GetVuid(parentName, 'commonparent')
+                # always skip the first member, its the dispatch handle and has not parent VUs
+                if member.type not in parent_exception_list and member != members[0]:
+                    # Replace with alias if one
+                    alias = self.vk.commands[parentName].alias if parentName in self.vk.commands else None
+                    parent = alias if alias else parentName
+
+                    if members[0].type == 'VkDevice':
+                        parentVUID = getVUID(self.valid_vuids, f'VUID-{parent}-{member.name}-parent')
```
spencer-lunarg, line 433: main reason to use the util `getVUID` is it gives warnings if the VUID doesn't exists and in returns allows me to find issues with the spec instead of just suppressing it with a `kVUIDUndefined` forever as we were doing

layers/vulkan/generated/object_tracker.cpp
artem-lunarg: `kVUIDUndefined` (wrong_device_code) for the queue should never be signaled, is it correct?

layers/vulkan/generated/object_tracker.cpp
```diff
@@ -1432,7 +1432,7 @@ bool ObjectLifetimes::PreCallValidateCmdBindPipeline(
     VkPipelineBindPoint                         pipelineBindPoint,
     VkPipeline                                  pipeline) const {
     bool skip = false;
-    skip |= ValidateObject(commandBuffer, kVulkanObjectTypeCommandBuffer, false, "VUID-vkCmdBindPipeline-commandBuffer-parameter", "VUID-vkCmdBindPipeline-commonparent", "vkCmdBindPipeline");
+    skip |= ValidateObject(commandBuffer, kVulkanObjectTypeCommandBuffer, false, "VUID-vkCmdBindPipeline-commandBuffer-parameter", kVUIDUndefined, "vkCmdBindPipeline");
     skip |= ValidateObject(pipeline, kVulkanObjectTypePipeline, false, "VUID-vkCmdBindPipeline-pipeline-parameter", "VUID-vkCmdBindPipeline-commonparent", "vkCmdBindPipeline");
```
artem-lunarg, line 1436: Question about second call that validates `pipeline`. It accepts `VUID-vkCmdBindPipeline-commonparent`. Looking into (old) `ValidateObject` implementation not sure that original code can detect what this VU asks for. Is it something that is addressed by this PR?

layers/vulkan/generated/object_tracker.cpp
spencer-lunarg: correct, how would you determine if the `vkQueueSubmit` came from the "correct" device, as long as everything is valid

layers/vulkan/generated/object_tracker.cpp
spencer-lunarg: first, let me write a test to prove my assumption as well

but the idea is we know what `VkDevice` was used for `vkCreateGraphicsPipelines` and we are making sure it is the same as the `VkDevice` to allocate the `VkCommandBuffer`

spencer-lunarg: added `CommonParentBindPipeline` to confirm this works as intended

artem-lunarg: thanks for explanation, I had to check implementation more carefully

artem-lunarg (approved):

PR #6240, opened 2023-08-02
tests: Remove features checks not needed

spencer-lunarg:
There is a feature requirement section in the spec with many statements such as

image

I went through a large amount of these and removed the useless check for the feature bit support

1. Some tests are already forgetting to do this and work fine
2. before Profile/MockICD, the old DevSim layer couldn't enable extension feature bits so I think this is a source why they were always checked

### DIFF
--- a/tests/unit/buffer_positive.cpp
+++ b/tests/unit/buffer_positive.cpp
@@ -108,11 +108,8 @@ TEST_F(PositiveBuffer, DISABLED_PerfGetBufferAddressWorstCase) {
     }
 
     auto buffer_addr_features = LvlInitStruct<VkPhysicalDeviceBufferDeviceAddressFeaturesKHR>();
-    auto features2 = GetPhysicalDeviceFeatures2(buffer_addr_features);
-    if (!buffer_addr_features.bufferDeviceAddress) {
-        GTEST_SKIP() << "bufferDeviceAddress not supported";
-    }
-    ASSERT_NO_FATAL_FAILURE(InitState(nullptr, &features2, VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT));
+    GetPhysicalDeviceFeatures2(buffer_addr_features);
+    ASSERT_NO_FATAL_FAILURE(InitState(nullptr, &buffer_addr_features, VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT));
 
     // Allocate common buffer memory, all buffers will be bound to it so that they have the same starting address
     auto alloc_flags = LvlInitStruct<VkMemoryAllocateFlagsInfo>();
@@ -163,11 +160,8 @@ TEST_F(PositiveBuffer, DISABLED_PerfGetBufferAddressGoodCase) {
     }
 
     auto buffer_addr_features = LvlInitStruct<VkPhysicalDeviceBufferDeviceAddressFeaturesKHR>();
-    auto features2 = GetPhysicalDeviceFeatures2(buffer_addr_features);
-    if (!buffer_addr_features.bufferDeviceAddress) {
-        GTEST_SKIP() << "bufferDeviceAddress not supported";
-    }
-    ASSERT_NO_FATAL_FAILURE(InitState(nullptr, &features2, VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT));
+    GetPhysicalDeviceFeatures2(buffer_addr_features);
+    ASSERT_NO_FATAL_FAILURE(InitState(nullptr, &buffer_addr_features, VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT));
 
     // Allocate common buffer memory, all buffers will be bound to it so that they have the same starting address
     auto alloc_flags = LvlInitStruct<VkMemoryAllocateFlagsInfo>();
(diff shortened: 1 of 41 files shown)


tests/unit/buffer_positive.cpp
```diff
@@ -108,11 +108,8 @@ TEST_F(PositiveBuffer, DISABLED_PerfGetBufferAddressWorstCase) {
     }
 
     auto buffer_addr_features = LvlInitStruct<VkPhysicalDeviceBufferDeviceAddressFeaturesKHR>();
-    auto features2 = GetPhysicalDeviceFeatures2(buffer_addr_features);
-    if (!buffer_addr_features.bufferDeviceAddress) {
-        GTEST_SKIP() << "bufferDeviceAddress not supported";
-    }
-    ASSERT_NO_FATAL_FAILURE(InitState(nullptr, &features2, VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT));
+    GetPhysicalDeviceFeatures2(buffer_addr_features);
```
artem-lunarg, line 111: It's a lot of changes and I don't want to suggest a change which might be a personal preference.

Still with regard to future changes, I guess that requesting feature directly should be fine too (I find it more explicit): 
```
buffer_addr_features.bufferDeviceAddress = VK_TRUE
```

tests/unit/memory.cpp
```diff
@@ -1622,9 +1622,6 @@ TEST_F(NegativeMemory, BufferDeviceAddressEXT) {
 
     auto buffer_device_address_features = LvlInitStruct<VkPhysicalDeviceBufferAddressFeaturesEXT>();
     GetPhysicalDeviceFeatures2(buffer_device_address_features);
-    if (buffer_device_address_features.bufferDeviceAddressCaptureReplay != VK_TRUE) {
```
artem-lunarg, line 1625: `bufferDeviceAddressCaptureReplay` - did not found guarantees for this feature

tests/unit/other_positive.cpp
```diff
@@ -433,10 +430,6 @@ TEST_F(VkPositiveLayerTest, Vulkan12Features) {
 
     auto bda_features = LvlInitStruct<VkPhysicalDeviceBufferDeviceAddressFeatures>();
     VkPhysicalDeviceFeatures2 features2 = GetPhysicalDeviceFeatures2(bda_features);
-    if (!bda_features.bufferDeviceAddress) {
```
artem-lunarg, line 436: `bufferDeviceAddress` - Vulkan 1.3 or  VK_KHR_buffer_device_address

tests/unit/ray_tracing.cpp
```diff
@@ -789,9 +762,6 @@ TEST_F(NegativeRayTracing, CmdCopyUnboundAccelerationStructure) {
     auto accel_features = LvlInitStruct<VkPhysicalDeviceAccelerationStructureFeaturesKHR>(&bda_features);
     auto features2 = GetPhysicalDeviceFeatures2(accel_features);
 
-    if (bda_features.bufferDeviceAddress == VK_FALSE) {
```
artem-lunarg, line 792: `bufferDeviceAddress` - with the presence of acceleration structure extension the requirements are: Vulkan 1.2 or the VK_KHR_buffer_device_address extension.

tests/unit/shader_cooperative_matrix.cpp
```diff
@@ -110,13 +103,7 @@ TEST_F(NegativeShaderCooperativeMatrix, KHRUnsupportedStage) {
 
     auto cooperative_matrix_features = LvlInitStruct<VkPhysicalDeviceCooperativeMatrixFeaturesKHR>();
     auto memory_model_features = LvlInitStruct<VkPhysicalDeviceVulkanMemoryModelFeaturesKHR>(&cooperative_matrix_features);
-    auto features2 = GetPhysicalDeviceFeatures2(memory_model_features);
-    if (memory_model_features.vulkanMemoryModel == VK_FALSE) {
-        GTEST_SKIP() << "vulkanMemoryModel feature not supported";
-    }
-    if (!cooperative_matrix_features.cooperativeMatrix) {
```
artem-lunarg, line 117: `cooperativeMatrix` - same here

tests/unit/shader_spirv_positive.cpp
```diff
@@ -435,12 +430,8 @@ TEST_F(PositiveShaderSpirv, Std430SpirvOptFlags12) {
     }
 
     auto features12 = LvlInitStruct<VkPhysicalDeviceVulkan12Features>();
-    auto features2 = GetPhysicalDeviceFeatures2(features12);
-    if (features12.scalarBlockLayout == VK_FALSE || features12.uniformBufferStandardLayout == VK_FALSE) {
```
artem-lunarg, line 439: `scalarBlockLayout` - 1.2 is sufficient only for `uniformBufferStandardLayout`

artem-lunarg:
Thanks for this clean up! Comments in a few places where we might not have a guarantee.

tests/unit/buffer_positive.cpp
spencer-lunarg: agree, I guess was going for consistency, also if you have a test that needs to chain them, it is ready to go

tests/unit/memory.cpp
spencer-lunarg: ugh, good catch

tests/unit/other_positive.cpp
spencer-lunarg: good catch

tests/unit/shader_spirv_positive.cpp
spencer-lunarg: once again, glad your eyes are working

tests/unit/memory.cpp
spencer-lunarg: sorry, I removed this because we are setting it false under, so we don't care if it is supported or not

so, actually just removing the `GetPhysicalDeviceFeatures2(buffer_device_address_features);` now as well

artem-lunarg (approved):

PR #6242, opened 2023-08-02
layers: Add initial VK_KHR_maintenance5

spencer-lunarg:
Adds some initial support for `VK_KHR_maintenance5`

### DIFF
--- a/layers/core_checks/cc_cmd_buffer.cpp
+++ b/layers/core_checks/cc_cmd_buffer.cpp
@@ -388,33 +388,65 @@ bool CoreChecks::PreCallValidateResetCommandBuffer(VkCommandBuffer commandBuffer
     return skip;
 }
 
+bool CoreChecks::ValidateCmdBindIndexBuffer(const CMD_BUFFER_STATE &cb_state, const BUFFER_STATE &buffer_state, VkDeviceSize offset,
+                                            VkIndexType indexType, CMD_TYPE cmd_type) const {
+    bool skip = false;
+    const char *api_call = CommandTypeString(cmd_type);
+    const bool is_2 = cmd_type != CMD_BINDINDEXBUFFER;
+
+    skip |= ValidateCmd(cb_state, cmd_type);
+
+    const char *vuid = is_2 ? "VUID-vkCmdBindIndexBuffer2KHR-buffer-08784" : "VUID-vkCmdBindIndexBuffer-buffer-08784";
+    skip |= ValidateBufferUsageFlags(cb_state.commandBuffer(), buffer_state, VK_BUFFER_USAGE_INDEX_BUFFER_BIT, true, vuid, api_call,
+                                     "VK_BUFFER_USAGE_INDEX_BUFFER_BIT");
+    vuid = is_2 ? "VUID-vkCmdBindIndexBuffer2KHR-buffer-08785" : "VUID-vkCmdBindIndexBuffer-buffer-08785";
+    skip |= ValidateMemoryIsBoundToBuffer(cb_state.commandBuffer(), buffer_state, api_call, vuid);
+
+    const VkDeviceSize offset_align = static_cast<VkDeviceSize>(GetIndexAlignment(indexType));
+    if (!IsIntegerMultipleOf(offset, offset_align)) {
+        const LogObjectList objlist(cb_state.commandBuffer(), buffer_state.buffer());
+        vuid = is_2 ? "VUID-vkCmdBindIndexBuffer2KHR-offset-08783" : "VUID-vkCmdBindIndexBuffer-offset-08783";
+        skip |= LogError(objlist, vuid, "%s() offset (0x%" PRIxLEAST64 ") does not fall on alignment (%s) boundary.", api_call,
+                         offset, string_VkIndexType(indexType));
+    }
+    if (offset >= buffer_state.requirements.size) {
+        const LogObjectList objlist(cb_state.commandBuffer(), buffer_state.buffer());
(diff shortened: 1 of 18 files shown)


layers/stateless/sl_cmd_buffer.cpp
```diff
@@ -45,6 +45,26 @@ bool StatelessValidation::manual_PreCallValidateCmdBindIndexBuffer(VkCommandBuff
...
 }
 
+bool StatelessValidation::manual_PreCallValidateCmdBindIndexBuffer2KHR(VkCommandBuffer commandBuffer, VkBuffer buffer,
+                                                                       VkDeviceSize offset, VkDeviceSize size,
+                                                                       VkIndexType indexType) const {
+    bool skip = false;
+
+    if (indexType == VK_INDEX_TYPE_NONE_NV) {
+        skip |= LogError(commandBuffer, "VUID-vkCmdBindIndexBuffer2KHR-indexType-08786",
+                         "vkCmdBindIndexBuffer2KHR() indexType must not be VK_INDEX_TYPE_NONE_NV.");
```
artem-lunarg, line 55: The latest spec uses `VK_INDEX_TYPE_NONE_KHR` name

layers/stateless/sl_cmd_buffer.cpp
```diff
@@ -45,6 +45,26 @@ bool StatelessValidation::manual_PreCallValidateCmdBindIndexBuffer(VkCommandBuff
...
+    }
+
+    const auto *index_type_uint8_features = LvlFindInChain<VkPhysicalDeviceIndexTypeUint8FeaturesEXT>(device_createinfo_pnext);
```
artem-lunarg, line 58: perf nit: if indexType is not UINT8_EXT then `index_type_uint8_features` is computed unnecessary. The suggestion is to move `index_type_uint8_features` initialization inside block.

layers/core_checks/cc_cmd_buffer.cpp
```diff
@@ -388,33 +388,65 @@ bool CoreChecks::PreCallValidateResetCommandBuffer(VkCommandBuffer commandBuffer
...
+    bool skip = false;
+    skip |= ValidateCmdBindIndexBuffer(*cb_state, *buffer_state, offset, indexType, CMD_BINDINDEXBUFFER2KHR);
+    if (size != VK_WHOLE_SIZE) {
+        const VkDeviceSize offset_align = static_cast<VkDeviceSize>(GetIndexAlignment(indexType));
+        if (!IsIntegerMultipleOf(size, offset_align)) {
+            skip |= LogError(commandBuffer, "VUID-vkCmdBindIndexBuffer2KHR-size-08767",
+                             "vkCmdBindIndexBuffer2KHR() size (0x%" PRIxLEAST64 ") does not fall on alignment (%s) boundary.", size,
+                             string_VkIndexType(indexType));
+        }
+        if ((offset + size) > buffer_state->requirements.size) {
```
artem-lunarg, line 442: I wonder if we should check against `createInfo.size`. Driver may request more memory in `requirements.size` than `createInfo.size` but it might be not safe to write in that additional tail region (driver could use for internal purposes) unless spec says it's fine.

layers/core_checks/cc_descriptor.cpp
```diff
@@ -1184,7 +1184,8 @@ bool CoreChecks::ValidateDescriptor(const DescriptorContext &context, const Desc
                  sampler_state->createInfo.borderColor == VK_BORDER_COLOR_FLOAT_CUSTOM_EXT) &&
                 (sampler_state->customCreateInfo.format == VK_FORMAT_UNDEFINED)) {
                 if (image_view_format == VK_FORMAT_B4G4R4A4_UNORM_PACK16 || image_view_format == VK_FORMAT_B5G6R5_UNORM_PACK16 ||
-                    image_view_format == VK_FORMAT_B5G5R5A1_UNORM_PACK16) {
+                    image_view_format == VK_FORMAT_B5G5R5A1_UNORM_PACK16 ||
+                    image_view_format == VK_FORMAT_A1B5G5R5_UNORM_PACK16_KHR) {
```
artem-lunarg, line 1188: The latest spec requires separate `VUID-VkSamplerCustomBorderColorCreateInfoEXT-format-08807` for new format.

spencer-lunarg: sorry, was not obvious at all, but I have https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6032 to go and clean up some of these VUID that should (and now can!) be the same

layers/stateless/sl_cmd_buffer.cpp
spencer-lunarg: yep, this was all updated in the 1.3.261 headers, when I rebase it all fails so will update the new enums accordingly

spencer-lunarg: aww, this was aliased, regardless, I cross referenced https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6050

spencer-lunarg: good catch

layers/core_checks/cc_cmd_buffer.cpp
spencer-lunarg: good question, raised a spec question, for now using `requirements` over `createInfo` would prevent false positives if they decide it should be `requirements`

if they decide it should be `createInfo`, that is a quick fix after

spencer-lunarg: since we have had `08782` covered for a LONG time and it also says

> size of buffer

I think its safe to assume it did mean `requirements`

artem-lunarg (approved):

PR #6243, opened 2023-08-02
cmake: Add utility libraries repository dependence

christophe-lunarg:
Splitting https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6088, the PR being too big and hard to resolve cause of test failures


jeremyg-lunarg (approved):
LGTM

layers/CMakeLists.txt
```diff
@@ -392,7 +392,7 @@ if (USE_ROBIN_HOOD_HASHING)
 endif()
 # Order matters here. VkLayer_utils should be the last link library to ensure mimalloc overrides are picked up correctly.
```
artem-lunarg, line 393: > **Order matters here. VkLayer_utils should be the last link library to ensure mimalloc overrides are picked up correctly.**

scripts/known_good.json
```diff
@@ -20,6 +20,20 @@
...
             "commit": "v1.3.260"
         },
+        {
+            "name": "Vulkan-Utility-Libraries",
+            "url": "https://github.com/KhronosGroup/Vulkan-Utility-Libraries.git",
+            "sub_dir": "Vulkan-Utility-Libraries",
+            "build_dir": "Vulkan-Utility-Libraries/build",
+            "install_dir": "Vulkan-Utility-Libraries/build/install",
+            "commit": "v1.3.260",
+            "deps": [
```
artem-lunarg, line 30: I don't know much how know_good works. Why do we have `deps` here but don't specify dependencies in other sections?

layers/CMakeLists.txt
```diff
@@ -392,7 +392,7 @@ if (USE_ROBIN_HOOD_HASHING)
...
 # Otherwise, libraries after VkLayer_utils will not benefit from this performance improvement.
-target_link_libraries(vvl PRIVATE VVL-SPIRV-LIBS VkLayer_utils)
+target_link_libraries(vvl PRIVATE VVL-SPIRV-LIBS VkLayer_utils Vulkan::LayerSettings)
```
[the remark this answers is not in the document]
jeremyg-lunarg: doh! very good point!

jeremyg-lunarg (changes requested):
@artem-lunarg 's comment about ordering needs to be addressed.

layers/CMakeLists.txt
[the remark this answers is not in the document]
christophe-lunarg: Fixed

scripts/known_good.json
artem-lunarg: Okay, sounds good.

I'll ping @juan-lunarg here. It would be nice to get comment later. I see that in the profiles repository we use `deps` in a few places but did not use in vvl before.

artem-lunarg (approved):

jeremyg-lunarg (approved):

spencer-lunarg (approved):

scripts/known_good.json
juan-lunarg: By specifying VULKAN_HEADERS_INSTALL_DIR the utility libraries will use the headers from VVL instead of it's own. That's the intent at least.

IE this situation arises due to shared dependencies which isn't an issue with other repos we consume.

PR #6245, opened 2023-08-02
Using VK_EXT_layer_settings  from the layer settings lib

christophe-lunarg:
Use layer settings library, replacing VK_EXT_layer_settings from the validation layer by the version from the layer settings library, shared across layers.


juan-lunarg:
I'd very much like this issue to be addressed first:
https://github.com/KhronosGroup/Vulkan-Utility-Libraries/issues/13

It should fix the dependency issue we have been having.

layers/layer_options.cpp
```diff
@@ -417,54 +404,65 @@ static void SetValidationGPUBasedSetting(CHECK_ENABLED &enable_data, const char
...
-        }
+    VlLayerSettingSet layerSettingSet = VK_NULL_HANDLE;
+    vlCreateLayerSettingSet(OBJECT_LAYER_NAME, vlFindLayerSettingsCreateInfo(settings_data->create_info), nullptr, nullptr,
+                            &layerSettingSet);
+   
+    std::vector<std::string> enables;
+    if (vlHasLayerSetting(layerSettingSet, SETTING_ENABLES)) {
+        vlGetLayerSettingValues(layerSettingSet, SETTING_ENABLES, enables);
+    }
+    const std::string& string_enables = Merge(enables);
```
charles-lunarg, line 433: This looks very suspicious - Merge returns a value, but then we store it as a reference, does this dangle the string?

juan-lunarg: Agreed the code looks problematic.

artem-lunarg: That's fine according to c++ rules, lifetime of rvalues is extended if they are assigned to `const ref`. The problem here that this code looks like optimization but in fact can be slower (or maybe the same speed). According to `Merge` implementation compiler might do RVO or if it can then IIRC it **must** to use move constructor. RVO is likely so result can be put into result variable without copying. And if `const ref` it can be additional copy to store somewhere temp object so you can assign const ref to.

juan-lunarg: Gotcha. Although the code is legal I'd rather `string_enables` just be a `const std::string`. I'd rather initialization code be as simple as possible.

scripts/known_good.json
```diff
@@ -20,6 +20,20 @@
             "install_dir": "Vulkan-Headers/build/install",
             "commit": "v1.3.261"
         },
+        {
+            "name": "Vulkan-Utility-Libraries",
```
juan-lunarg, line 24: Note we will want to document this new dependency somewhere for system package managers when this PR gets merged in.

Otherwise they will get build errors and make issues. Not sure where the best place for that would be.

tests/framework/test_framework.h
```diff
@@ -37,6 +37,8 @@
 // Can be used by tests to record additional details / description of test
 #define TEST_DESCRIPTION(desc) RecordProperty("description", desc)
 
+std::string format(const char *message, ...);
```
juan-lunarg, line 40: Given this helper function is only used in `others.cpp` please move the function to `others.cpp` It doesn't need to be in test_framework.h

tests/framework/layer_validation_tests.h
```diff
@@ -43,6 +42,8 @@
 #include <vector>
 #include <condition_variable>
 
+#define OBJECT_LAYER_NAME "VK_LAYER_KHRONOS_validation"
```
juan-lunarg, line 45: Can't the test code just include `layer_options.h`?

layers/layer_options.h
```diff
@@ -16,19 +16,22 @@
  * limitations under the License.
  */
 
+#pragma once
 #include "generated/chassis.h"
 
+#define OBJECT_LAYER_NAME "VK_LAYER_KHRONOS_validation"
```
juan-lunarg, line 22: Avoid macros.

```suggestion
inline const char * GetObjectLayerName() { return "VK_LAYER_KHRONOS_validation"; }
```

tests/framework/layer_validation_tests.cpp
```diff
@@ -1858,13 +1858,10 @@ void VkSyncValTest::InitSyncValFramework(bool enable_queue_submit_validation) {
...
     // Optional feature definition, add if requested (but they can't be defined at the conditional scope)
-    const char *kEnableQueuSubmitSyncValidation = "VALIDATION_CHECK_ENABLE_SYNCHRONIZATION_VALIDATION_QUEUE_SUBMIT";
-    VkLayerSettingValueDataEXT qs_setting_string_value{};
-    qs_setting_string_value.arrayString.pCharArray = kEnableQueuSubmitSyncValidation;
-    qs_setting_string_value.arrayString.count = strlen(qs_setting_string_value.arrayString.pCharArray);
-    VkLayerSettingValueEXT qs_enable_setting_val = {"enables", VK_LAYER_SETTING_VALUE_TYPE_STRING_ARRAY_EXT,
-                                                    qs_setting_string_value};
-    VkLayerSettingsEXT qs_settings{VK_STRUCTURE_TYPE_INSTANCE_LAYER_SETTINGS_EXT, nullptr, 1, &qs_enable_setting_val};
+    const char *kEnableQueuSubmitSyncValidation[] = {"VALIDATION_CHECK_ENABLE_SYNCHRONIZATION_VALIDATION_QUEUE_SUBMIT"};
+    std::vector<VkLayerSettingEXT> settings{
```
juan-lunarg, line 1862: I don't see a need for std::vector. There is only 1 `VkLayerSettingEXT`.

tests/unit/best_practices.cpp
```diff
@@ -23,13 +23,18 @@ void VkBestPracticesLayerTest::InitBestPracticesFramework() {
...
-    bp_setting_string_value.arrayString.pCharArray = vendor_checks_to_enable;
-    bp_setting_string_value.arrayString.count = sizeof(bp_setting_string_value.arrayString.pCharArray);
-    VkLayerSettingValueEXT bp_vendor_all_setting_val = {"enables", VK_LAYER_SETTING_VALUE_TYPE_STRING_ARRAY_EXT,
-                                                        bp_setting_string_value};
-    VkLayerSettingsEXT bp_settings{VK_STRUCTURE_TYPE_INSTANCE_LAYER_SETTINGS_EXT, nullptr, 1, &bp_vendor_all_setting_val};
-    features_.pNext = &bp_settings;
+    std::vector<const char *> input_values{vendor_checks_to_enable};
+    std::vector<VkLayerSettingEXT> settings{
+        {
+            OBJECT_LAYER_NAME, "enables",
```
juan-lunarg, line 29: I don't see a need for std::vector. There is only 1 VkLayerSettingEXT.

tests/unit/others.cpp
juan-lunarg, line 1: Please write 2 positives tests that enables all the validation layer functionality using the new programmatic interface. Just creating a device with a command buffer is good enough. Nothing too complex.

- Best practice + synchronization + gpuav
- Best practice + synchronization + debug printf

This is something developers will try to do with the new programmatic interface.

And given that `debug Printf` functionality and `GPU-Assisted validation` cannot be run at the same time. We should have a negative test that produces a clear error message to the user as well. (I don't believe we have a test for this AFAIK).

juan-lunarg (changes requested):
https://github.com/KhronosGroup/Vulkan-Utility-Libraries/issues/13

I'd like this issue to be fixed before we merge this change. Avoiding the circular dependency. FYI @charles-lunarg

layers/layer_options.cpp
christophe-lunarg: https://herbsutter.com/2008/01/01/gotw-88-a-candidate-for-the-most-important-const/

tests/unit/best_practices.cpp
christophe-lunarg: Done

scripts/known_good.json
juan-lunarg: I'll just make an issue once this gets in.

juan-lunarg:
@charles-lunarg has a WIP PR to fix the circular dependency issue.

https://github.com/KhronosGroup/Vulkan-Profiles/pull/487

layers/layer_options.h
christophe-lunarg: I do find this pointless but I changed this.

juan-lunarg:
Circular dependency has been fixed. Please see updates to Vulkan-Profiles and the utility libraries.

tests/unit/others.cpp
juan-lunarg: Feel free to ignore the `GetLayerName` suggestion. I didn't consider that aspect.

juan-lunarg (approved):

tests/unit/others.cpp
christophe-lunarg: Ok!

PR #6248, opened 2023-08-03
tests: Remove some VkBufferObj overloads

spencer-lunarg:
Currently we have so many random ways to create a `VkBufferObj` that it becomes hard to follow which path does what

This also leads to a lot of times we need to do things like

```
VkMemoryPropertyFlags reqs = 0
VkBufferObj buffer_one(*m_device, 2048, reqs);
```

Because `VkBufferObj buffer_one(*m_device, 2048, 0);` is ambiguous with the various overloads

This just a small change to get things in cleaner direction


artem-lunarg (approved):
LGTM. Very nice change. IMO win in both directions. Interfaces are cleaner and code also easier to understand. New concepts have its own cost, a lot of concepts for non-fundamental parts quickly add to complexity.

PR #6249, opened 2023-08-03
tests: Confirm ShaderVal handles descriptorIndexing

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5465

Finally looking into this, I feel confident that we handle all cases of descriptor indexing correctly now in Shader Validation. I added some tests to confirm this

### DIFF
--- a/layers/state_tracker/shader_module.cpp
+++ b/layers/state_tracker/shader_module.cpp
@@ -589,8 +589,7 @@ ImageAccess::ImageAccess(const SPIRV_MODULE_STATE& module_state, const Instructi
     if (image_load_pointer->Opcode() == spv::OpVariable) {
         variable_image_insn = image_load_pointer;
     } else if (image_load_pointer->Opcode() == spv::OpAccessChain || image_load_pointer->Opcode() == spv::OpInBoundsAccessChain) {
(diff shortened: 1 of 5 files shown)


layers/state_tracker/shader_module.cpp
```diff
@@ -1132,15 +1131,16 @@ const Instruction* SPIRV_MODULE_STATE::GetConstantDef(uint32_t id) const {
     return nullptr;
 }
 
-// Either returns the constant value described by the instruction at id, or 1
+// Returns the constant value described by the instruction at |id|
+// Caller ensures there can't be a runtime array or specialization constants
 uint32_t SPIRV_MODULE_STATE::GetConstantValueById(uint32_t id) const {
     const Instruction* value = GetConstantDef(id);
 
     if (!value) {
```
artem-lunarg, line 1139: nit: entire block can be: `assert(value);` + comment

artem-lunarg (approved):
LGTM

layers/state_tracker/shader_module.cpp
spencer-lunarg: good call, fixed

PR #6252, opened 2023-08-03
layers: Fix and test VUID 03230 and 03233

aqnuep:
This PR is the validation and test fixes for https://github.com/KhronosGroup/Vulkan-Docs/issues/2185.

I've also added some tests for VUID 03233 which didn't have any tests before.

@spencer-lunarg, I guess you'd like to take a look at it.

@llandwerlin-intel, I think you might want to look at this, considering you have an implementation of `VK_KHR_performance_query`.

### DIFF
--- a/tests/unit/query.cpp
+++ b/tests/unit/query.cpp
@@ -529,6 +529,11 @@ TEST_F(NegativeQuery, PerformanceIncompletePasses) {
     AddRequiredExtensions(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
     AddRequiredExtensions(VK_KHR_PERFORMANCE_QUERY_EXTENSION_NAME);
     AddRequiredExtensions(VK_EXT_HOST_QUERY_RESET_EXTENSION_NAME);
+    AddOptionalExtensions(VK_KHR_VIDEO_QUEUE_EXTENSION_NAME);
+
+    // Vulkan 1.1 is a dependency of VK_KHR_video_queue, but both the version and the extension
+    // is optional from the point of view of this test case
+    SetTargetApiVersion(VK_API_VERSION_1_1);
 
     ASSERT_NO_FATAL_FAILURE(InitFramework(m_errorMonitor));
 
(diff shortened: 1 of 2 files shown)


spencer-lunarg (approved):
LGTM, thanks for getting this up quickly

aqnuep:
> LGTM, thanks for getting this up quickly

Sure thing. However, I did not manage to get the test case tested because I don't have a driver at hand actually supporting `VK_KHR_performance_query` and the test case is incompatible with the MockICD.

Let's see if the CI passes.

aqnuep:
@spencer-lunarg, could you let me know what fails in your internal CI with this PR? I see a public failure related to thread sanitizer, but I doubt it's actually related to this PR.

aqnuep:
Ah, looks like `VK_KHR_performance_query` has another capability that indicates whether `vkCmdCopyQueryPoolResults` is supported for performance queries in the first place.

I guess Mesa Intel maybe? (there are many of them)

I don't want to start refactoring someone else's extension's validation tests. I can either just remove the additional coverage I've added for `vkCmdCopyQueryPoolResults`, which the original tests for performance query lacked, or I can add the capability check, per above.

tests/unit/query.cpp
```diff
@@ -529,6 +529,7 @@ TEST_F(NegativeQuery, PerformanceIncompletePasses) {
     AddRequiredExtensions(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
     AddRequiredExtensions(VK_KHR_PERFORMANCE_QUERY_EXTENSION_NAME);
     AddRequiredExtensions(VK_EXT_HOST_QUERY_RESET_EXTENSION_NAME);
+    AddOptionalExtensions(VK_KHR_VIDEO_QUEUE_EXTENSION_NAME);
```
spencer-lunarg, line 532: trying to run locally, this requires version 1.1

aqnuep: In my run I only see the test skipped when running against MockICD for other reasons, but you're right that `VK_KHR_video_queue` requires Vulkan 1.1. I've added appropriate code to make sure that's handled.

spencer-lunarg: we have an explicit skip, I was just trying to mess around to see if I could get the test running, only to finally get on internet and find out the CI then passed which ment you fixed it and I just ended my investigation there

aqnuep: I see. Anyway, it was very useful feedback that was worth addressing. Thanks!

PR #6253, opened 2023-08-03
WIP - layers: Add ErrorObject

spencer-lunarg:
This adds a `ErrorObject` that is designed to be a replacement for passing various parameters through functions for the sake of logging them.

The main idea is that in the same way we have a `VkResult` in all `PostCallRecord`, we would add a `ErrorObject` on the stack when the chassis calls all `PreCallValidate` functions

Currently this is WIP as want feedback on the design, things I want to still add is

- Integrate the `Location`/`Loc` system used in the Synchronization validation 
- Have a way so you can add to things like `LogObjectList` and it resets when you leave scope

### DIFF
--- a/layers/error_message/core_error_location.h
+++ b/layers/error_message/core_error_location.h
@@ -22,6 +22,7 @@
 #include <sstream>
 #include <limits>
 
+#include "logging.h"
 #include "containers/custom_containers.h"
 namespace core_error {
 // structure to track where a validation error occurs, and capture enough information
@@ -259,3 +260,14 @@ static const std::string& FindVUID(OuterKey key, const Location& loc, const Tabl
 }
 
 }  // namespace core_error
+
+// Contains all information needed for errors to be logged out
+// This allows only a single object to be passed around for error messages
+struct ErrorObject {
+    std::string command;
+    VulkanTypedHandle handle;  // dispatchable handle is always first parameter of the function call
+    LogObjectList objlist;
+
+    ErrorObject(std::string command_, VulkanTypedHandle handle_) : command(command_), handle(handle_), objlist(handle) {}
+    std::string Location() const { return command; }
+};
--- a/scripts/generators/generator_utils.py
+++ b/scripts/generators/generator_utils.py
@@ -18,6 +18,17 @@ import os
 import sys
 import json
 
+# TODO - This is a temporary design to allow us to slowly increase functions to use
+#  ErrorObject until we feel confident with the design to roll it out everywhere
+# (this allows use to see what works well before spending time refactoring each call)
+error_object_functions = [
+    'vkCreateRenderPass',
+    'vkCreateRenderPass2',
+    'vkCreateRenderPass2KHR',
+    'vkCmdPushConstants',
+    'vkQueueBindSparse',
+]
+
 # Build a set of all vuid text strings found in validusage.json
 def buildListVUID(valid_usage_file: str) -> set:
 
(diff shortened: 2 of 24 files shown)


layers/error_message/core_error_location.h
```diff
@@ -259,3 +260,14 @@ static const std::string& FindVUID(OuterKey key, const Location& loc, const Tabl
 }
 
 }  // namespace core_error
+
+// Contains all information needed for errors to be logged out
+// This allows only a single object to be passed around for error messages
+struct ErrorObject {
+    std::string command;
```
jeremyg-lunarg, line 267: my long term goal with the other stuff in this file was to codegen all of the enums needed for the Location object so that they could apply to any command or data structure in the spec.  If that got done you could just shove one of them in here. That's a big lift though. 

Also, the first version of the location object was string based and copying all the strings around caused perf problems.  I don't think you're doing as much copying as that code did though.

scripts/generators/generator_utils.py
```diff
@@ -18,6 +18,17 @@
...
+# TODO - This is a temporary design to allow us to slowly increase functions to use
+#  ErrorObject until we feel confident with the design to roll it out everywhere
+# (this allows use to see what works well before spending time refactoring each call)
+error_object_functions = [
+    'vkCreateRenderPass',
+    'vkCreateRenderPass2',
+    'vkCreateRenderPass2KHR',
+    'vkCmdPushConstants',
+    'vkQueueBindSparse',
+]
```
jeremyg-lunarg, line 30: It would be good to do 1 function that has EXT, KHR and plain versions now.

jeremyg-lunarg (commented):
I'm really glad you're working on this!

layers/error_message/core_error_location.h
```diff
@@ -259,3 +260,14 @@ static const std::string& FindVUID(OuterKey key, const Location& loc, const Tabl
...
+    VulkanTypedHandle handle;  // dispatchable handle is always first parameter of the function call
+    LogObjectList objlist;
+
+    ErrorObject(std::string command_, VulkanTypedHandle handle_) : command(command_), handle(handle_), objlist(handle) {}
```
jzulauf-lunarg, line 271: I'd store the command as an enum.  We try to avoid any string manipulation until we have something to report.  std::string is notorious cumulative performance hits (death by 1000's cuts)

artem-lunarg: Death by 25000 cuts: https://groups.google.com/a/chromium.org/g/chromium-dev/c/EUqoIz2iFU4/m/kPZ5ZK0K3gEJ

jzulauf-lunarg: If we want to capture strings from char *'s would rather capture them similar to SimpleErrorLocation....

But we need consolidate Location, SimpleErrorLocation, etc into something that doesn't evoke [XKCD 927: Standards](https://www.explainxkcd.com/wiki/index.php/927:_Standards)

scripts/generators/generator_utils.py
spencer-lunarg: added `vkGetBufferDeviceAddress` (has EXT and KHR and core)

layers/error_message/core_error_location.h
spencer-lunarg: we have https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6259 now :tada: so will replace with `enum class Func` now

spencer-lunarg:
closing, refactored and have better design to show

Issue #6256, opened 2023-08-03
Pass pNext struct down to Dispatch calls in the Layers


zmike:
It seems like some part of the `CreateImage` chain in VVL is stripping out `VkImageFormatListCreateInfo` / `VK_STRUCTURE_TYPE_IMAGE_FORMAT_LIST_CREATE_INFO` structs, which prevents drivers from validating mutable format chains in the implicit `vkGetPhysicalDeviceImageFormatProperties2` call, which then guarantees a `NOT_SUPPORTED` error.

This struct must be included.

spencer-lunarg:
(everything this VU comes up :cry: )

So looking, I see there are a few pNext structs that are allowed in both `vkGetPhysicalDeviceImageFormatProperties2` and `vkCreateImage`

- VkImageFormatListCreateInfo
- VkImageStencilUsageCreateInfo
- VkImageCompressionControlEXT
- VkVideoProfileListInfoKHR
- VkOpticalFlowImageFormatInfoNV

Should be easy to add the logic, much harder to write a test with these `vkGetPhysicalDevice` as can't guarantee return values (until we make better improvements to MockICD for testing)

zmike:
fwiw I've filed this same issue in a variety of places (possibly even this same exact one) at least once before, so it seems like something has to be done to ensure all the pnext structs are passed to functions that need them.

spencer-lunarg:
Yes, the issue is knowing which pNext structs are allowed between the `caller` (ex `vkCreateImage`) and our internal dispatch calls (ex. `vkGetPhysicalDeviceImageFormatProperties2`)

(noting this for @arno-lunarg to take a stab at)
Luckily with the new `VulkanObject` we can generate this easily now, the `Struct` object has a `extends` and `extendedBy` member that can be used, so before would dispatch a call `vkGetPhysicalDeviceImageFormatProperties2`, it would see what `extendedBy` is for `VkPhysicalDeviceImageFormatInfo2` and if the `caller` has any, it would append to the pNext chain.

spencer-lunarg:
@zmike we hopefully got a temp fix for the `VUID-VkImageCreateInfo-imageCreateMaxMipLevels-02251` you were having (if not please shout again!) but keeping issue open for Arno to track work on providing a proper solution to this problem

zmike:
I just built 5546d5e82e0a365c4c3a5393c469ae3156fee85c and I'm still seeing the same issue.

[...]

PR #6258, opened 2023-08-04
layers: Pass pNext from vkCreatImage to dispatch

spencer-lunarg:
This "solves" https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6256 ... but I really hate how this code looks

We probably need a more elegant solution for the future as other calls might need this

The goal is to take a copy of the `pNext` structs from the incoming call that might be needed when making `DisaptchvkGetPhysicalDevice*` queries

### DIFF
--- a/layers/core_checks/cc_image.cpp
+++ b/layers/core_checks/cc_image.cpp
@@ -175,6 +175,18 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
     image_format_info.tiling = pCreateInfo->tiling;
     image_format_info.usage = pCreateInfo->usage;
     image_format_info.flags = pCreateInfo->flags;
+
+    // These are pNext structs that can be in both vkGetPhysicalDeviceImageFormatProperties2 and vkCreateImage
+    // need to pass in or else driver might return NOT_SUPPORTED falsly
+    // TODO - Generate the full list
+    auto image_format_list = LvlInitStruct<VkImageFormatListCreateInfo>();
+    auto image_format_list_in = LvlFindInChain<VkImageFormatListCreateInfo>(pCreateInfo->pNext);
+    if (image_format_list_in) {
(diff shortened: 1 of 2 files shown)


layers/core_checks/cc_image.cpp
```diff
@@ -175,6 +175,66 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
...
+    auto image_compression_control = LvlInitStruct<VkImageCompressionControlEXT>();
+    auto video_profile_list_info = LvlInitStruct<VkVideoProfileListInfoKHR>();
+    auto optical_flow_image_format = LvlInitStruct<VkOpticalFlowImageFormatInfoNV>();
+    VkBaseInStructure *pNext_end = nullptr;
+
+    if (auto image_format_list_in = LvlFindInChain<VkImageFormatListCreateInfo>(pCreateInfo->pNext)) {
+        image_format_list.pViewFormats = image_format_list_in->pViewFormats;
+        image_format_list.viewFormatCount = image_format_list_in->viewFormatCount;
+        if (image_format_info.pNext) {
+            pNext_end->pNext = reinterpret_cast<VkBaseInStructure *>(&image_format_list);
```
arno-lunarg, line 192: When landing on this line, `pNext_end` can be null, so `pNext_end->pNext` is not valid

arno-lunarg (changes requested):
I think the way the pNext chain is managed has some issues

layers/core_checks/cc_image.cpp
spencer-lunarg: I decided to solve the short term problem and pass in `VkImageFormatListCreateInfo` to unblock Mesa, a follow up can be added to provide proper support to all `Dispatch` calls (not just here)

arno-lunarg (approved):

PR #6259, opened 2023-08-04
layers: Add generated Error Location Helper

spencer-lunarg:
as a prequel to https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6253

Moves the `core_error`'s `Func`/`Struct`/`Field` to a generated file to allow it to be used outside of just sync


layers/vulkan/generated/error_location_helper.cpp
```diff
@@ -0,0 +1,4171 @@
...
+****************************************************************************/
+
+// NOLINTBEGIN
+
+#include "error_location_helper.h"
+#include "containers/custom_containers.h"
+#include <assert.h>
+
+namespace core_error {
+static const vvl::unordered_map<Func, std::string_view> funcTable{
```
artem-lunarg, line 29: If we want to avoid allocations of map nodes at loading time then these options can be considered:
* move map declaration inside function (if it's used only in a single function)
* global array that stores (enum + const char*) entries (point here is not to use C++ data structure that allocates).
* it might also might be small performance impact (< 1ms of loading time), so not that important, should be measured to know.
* something else

spencer-lunarg: > move map declaration inside function (if it's used only in a single function)

I opt in for this, simple, and only a perf issue if the app is having lots of errors (which at that point, this is the least of their worries)

artem-lunarg: 100% agree. If app starts hitting errors all bets are off.

artem-lunarg (approved):
LGTM! Comment about the place of map initialization.

layers/vulkan/generated/error_location_helper.h
```diff
@@ -0,0 +1,4170 @@
...
+//        auto field_level = struct_level.dot(Field::srcAccessMask);
+//        std::cout << field_level.Message() << std::endl;
+// will print:
+//        vkCmdPipelineBarrier(): pImageMemoryBarriers[42].srcAccessMask
+// VUIDs can be found for an error in generic code using a combination of the
+// function, structure, and fieldmembers.
+
+namespace core_error {
+enum class Func {
+    Empty = 0,
```
jzulauf-lunarg, line 43: From https://en.cppreference.com/w/cpp/language/enum : 
 
 > Each enumerator is associated with a value of the underlying type. When initializers are provided in the enumerator-list, the values of enumerators are defined by those initializers. If the first enumerator does not have an initializer, the associated value is zero. For any other enumerator whose definition does not have an initializer, the associated value is the value of the previous enumerator plus one.
 
 So I'm torn, do we include all the initializers to explicit values, not even include the =0, or leave as is.  For debug reasons, I often find being able to look up a value in an enum declaration useful.  Also, anything I feel like I have to look up in the spec strikes me as non-obvious.
 
Do we need a comment (in the generator and/or generated code) that references the spec and lets people know that elsewhere we are depending on "zero-based and contiguous" as as not to break stuff?

jzulauf-lunarg: Would alpha-sort help people here?

layers/vulkan/generated/error_location_helper.cpp
```diff
@@ -0,0 +1,4166 @@
...
+// NOLINTBEGIN
+
+#include "error_location_helper.h"
+#include "containers/custom_containers.h"
+#include <assert.h>
+
+namespace core_error {
+
+const char* String(Func func) {
+    static const vvl::unordered_map<Func, std::string_view> table {
```
jzulauf-lunarg, line 31: We've found map setup's can impact VVL startup time.  Also string_view(const char *) does a strlen while string_view(of const char *, size_t) just does a store.

Array indexed by Func would be lighter weight.

layers/vulkan/generated/error_location_helper.cpp
```diff
@@ -0,0 +1,4166 @@
...
+    {Func::vkCmdSetRayTracingPipelineStackSizeKHR, "vkCmdSetRayTracingPipelineStackSizeKHR"},
+    {Func::vkCmdDrawMeshTasksEXT, "vkCmdDrawMeshTasksEXT"},
+    {Func::vkCmdDrawMeshTasksIndirectEXT, "vkCmdDrawMeshTasksIndirectEXT"},
+    {Func::vkCmdDrawMeshTasksIndirectCountEXT, "vkCmdDrawMeshTasksIndirectCountEXT"},
+    };
+    const auto entry = table.find(func);
+    return entry->second.data();
+}
+
+const char* String(Struct structure) {
```
jzulauf-lunarg, line 696: Same comments re maps, array, and string_view constructors

layers/vulkan/generated/error_location_helper.cpp
```diff
@@ -0,0 +1,4166 @@
...
+    {Struct::VkPhysicalDeviceRayQueryFeaturesKHR, "VkPhysicalDeviceRayQueryFeaturesKHR"},
+    {Struct::VkPhysicalDeviceMeshShaderFeaturesEXT, "VkPhysicalDeviceMeshShaderFeaturesEXT"},
+    {Struct::VkPhysicalDeviceMeshShaderPropertiesEXT, "VkPhysicalDeviceMeshShaderPropertiesEXT"},
+    {Struct::VkDrawMeshTasksIndirectCommandEXT, "VkDrawMeshTasksIndirectCommandEXT"},
+    };
+    const auto entry = table.find(structure);
+    return entry->second.data();
+}
+
+const char* String(Field field) {
```
jzulauf-lunarg, line 1665: And again here.

jzulauf-lunarg (commented):
Not 100% sold on string_view, but having these enum's fully fleshed out is good.  See comments on enum -> info lookups.

layers/vulkan/generated/error_location_helper.cpp
jeremyg-lunarg: good idea, since this enum is under our control and has contiguous values.  for vulkan enums, extensions have weird values that would make the array be very sparse.

jeremyg-lunarg (commented):
This is awesome! John has a very good point about array vs map though.

layers/vulkan/generated/error_location_helper.h
spencer-lunarg: > for debug reasons

I feel most Enums we have are not explicit values, and any good IDE/Editor should be able to help you with this

image

I guess I thought we had the `Empty = 0` so if people do an initialize (memset) and forgot to set a value, it defaults to something that is clearly wrong

> Would alpha-sort help people here?

I already have the `Field` sorted, can just do the same with `Func` and `Struct`

spencer-lunarg: decided to just give an explicit value since going to the array index style

layers/vulkan/generated/error_location_helper.cpp
spencer-lunarg: updated to be an Array index with `string_view(const char *, size_t)`

spencer-lunarg:
@jzulauf-lunarg @jeremyg-lunarg any other suggestion otherwise if this is good enough for you 2 like to merge this and slowly keep making progress towards the new error object

PR #6262, opened 2023-08-04
Queue Submit time replay rework, including render pass operations

jzulauf-lunarg:


layers/utils/vk_layer_utils.h
```diff
@@ -810,6 +810,18 @@ static inline void ToUpper(std::string &str) {
               [](char c) { return static_cast<char>(std::toupper(c)); });
 }
 
+// The standard does not specify the value of data() for zero-sized contatiners as being null or non-null,
+// only that it is not dereferenceable.
+//
+// Vulkan VUID's OTOH frequently require NULLs for zero-sized entries, or for option entries with non-zero counts
+template <typename T>
+const typename T::value_type *DataOrNull(const T &container) {
```
spencer-lunarg, line 818: what is the benefit of this over `std::optional`?

I am slightly against just throwing more stuff people aren't aware of in the utils headers if it will only be used in 1 test class. If the goal is to use this everywhere, then it should be added everywhere (or close to it) to start

layers/sync/sync_validation.cpp
```diff
@@ -8948,3 +8951,22 @@ void syncval_state::ImageState::SetOpaqueBaseAddress(ValidationStateTracker &dev
...
+      func_name(func_name_),
+      index(index_),
+      base_tag(exec_context.GetTagLimit()) {
+    exec_context_.BeginCommandBufferReplay(*this);
+}
+
+ReplayGuard::~ReplayGuard() { exec_context.EndCommandBufferReplay(); }
+
+const AccessContext *ReplayGuard::GetRecordedAccessContext() const {
+    if (rp_replay) {
```
artem-lunarg, line 8968: Q: In the old implementation we checked `rp_replay` only in `QueueBatchContext` but now we have a unified implementation which is also used by the `CommandBufferAccessContext`. Could you please elaborate on the reasons behind this change?

layers/sync/sync_validation.cpp
```diff
@@ -6748,7 +6755,7 @@ ResourceUsageTag SyncOpPipelineBarrier::Record(CommandBufferAccessContext *cb_co
     return tag;
 }
 
-void SyncOpPipelineBarrier::ReplayRecord(CommandExecutionContext &exec_context, const ResourceUsageTag tag) const {
+void SyncOpPipelineBarrier::ReplayRecord(CommandExecutionContext &exec_context, const ResourceUsageTag exec_tag) const {
```
artem-lunarg, line 6758: This naming clarifications (`tag` -> `exec_tag`, `tag_range` -> `first_use_range`, etc.) is very helpful to understand this non-trivial code. Might look like a minor change but IMO that's very valuable.

layers/sync/sync_validation.cpp
```diff
@@ -2365,53 +2371,58 @@ ResourceUsageTag CommandBufferAccessContext::RecordEndRenderPass(const CMD_TYPE
...
-                                     string_SyncHazard(hazard.hazard), index, sync_state_->FormatHandle(recorded_handle).c_str(),
-                                     FormatUsage(*hazard.recorded_access).c_str(), exec_context.FormatHazard(hazard).c_str());
-    };
     for (const auto &sync_op : sync_ops_) {
-        // we update the range to any include layout transition first use writes,
-        // as they are stored along with the source scope (as effective barrier) when recorded
-        tag_range.end = sync_op.tag + 1;
-        skip |= sync_op.sync_op->ReplayValidate(sync_op.tag, *this, base_tag, exec_context);
+        // Set the range to cover all accesses until the next sync_op, and validate
+        first_use_range.end = sync_op.tag;
```
artem-lunarg, line 2412: The range does not include `sync_op` (previously we did `sync_op.tag + 1`). Is it intentional?

layers/sync/sync_validation.cpp
jzulauf-lunarg: There were too many variants of DetectFirstUse, and too easy to get lost.  Moving the rp_replay to the base class allowed simplification, and CommandBufferAccessContext will never have it set (as the SyncOp derived classes that set this during replay are only legal at queue submit time.

artem-lunarg: Thanks. Useful information. Will keep this as not resolved to make it more visible.

artem-lunarg:
Quick update on the review status. My review questions was answered. There was one offline question related to events that should be verified, but otherwise LGTM.

layers/utils/vk_layer_utils.h
```diff
@@ -810,6 +810,18 @@ static inline void ToUpper(std::string &str) {
...
+    if (container.size()) {
```
arno-lunarg, line 819: Is there a reason to use `size()` over `!empty()`?

jzulauf-lunarg: Because I didn't think about `size()` performance on non-vector/array containers?  Fixed.

layers/sync/sync_validation.cpp
```diff
@@ -367,6 +367,10 @@ bool CommandExecutionContext::ValidForSyncOps() const {
     return valid;
 }
 
+ResourceUsageTag CommandExecutionContext::GetReplayBaseTag() const {
+    return current_replay_ ? current_replay_->base_tag : ResourceUsageRecord::kMaxIndex;
```
artem-lunarg, line 371: Is it possible to get here with null `current_replay_`? If not then it might be more clear to assert this.
This function is used in constructions like `exec_context.GetReplayBaseTag() + recorded_tag`. In the assumed case of null current replay adding `recorded_tag` to `ResourceUsageRecord::kMaxIndex` will overflow and will result in plausible small `ResourceUsageTag` value.

artem-lunarg (approved):
LGTM. One question about `GetReplayBaseTag` implementation. 

One refactor/cleanup suggestion for future PRs. Previously responsibility of `ReplayGuard` were more or less RAII house cleaning (and `current_replay_` was recorded context). Now `ReplayGuard` has additional functionality and it might look unusual to see `Guard`-named type  in other contexts (`current_replay_` also points to a guard variable). It would be nice to clarify this somehow, maybe just a rename to show it's more than a Guard wrapper?

jzulauf-lunarg:
@artem-lunarg -- your question about the replay base tag pushed me over the edge for a refactor to cleanup the replay state and flow.  Formerly replay has ping-ponged between the recorded and executing contexts (because it never really belong to either) and now it's cleaner, it think. And the "what it I'm not replaying" is mooted.

artem-lunarg:
> Formerly replay has ping-ponged between the recorded and executing contexts (because it never really belong to either) and now it's cleaner, it think

Yes, my notes say that `ReplayGuard` is owned by recorded context (on the stack) and exec context gets reference to it.

layers/sync/sync_validation.h
```diff
@@ -1577,7 +1565,8 @@ class CommandBufferAccessContext : public CommandExecutionContext {
 };
 
 // Allow keep track of the exec contexts replay state
-struct ReplayGuard {
+class ReplayState {
```
artem-lunarg, line 1568: Thanks, that's great. Much easier to understand how participants of replay communicate.

PR #6266, opened 2023-08-06
layers: Fix typo in logging output

ashemedai:
As the subject states.


spencer-lunarg (approved):
thanks, luckily it is the only instance of `likley` we had

ashemedai:
> thanks, luckily it is the only instance of `likley` we had

I did a full repo check for other occurences when I saw this. Thanks for the merge!

PR #6278, opened 2023-08-08
layers: Only print VK_OBJECT_TYPE_DEVICE if multi-device

spencer-lunarg:
Currently our error messages look like

> Validation Error: [ VUID-VkImageCreateInfo-mipLevels-00947 ] Object 0: handle = 0x561abf1affe0, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x4ca5e939 | vkCreateImage: parameter pCreateInfo->mipLevels is zero

but when using 1 `VkDevice` its quite obvious which device it is and this `VK_OBJECT_TYPE_DEVICE` serves no purpose and just adds noise

with this change it now just looks like

> Validation Error: [ VUID-VkImageCreateInfo-mipLevels-00947 ] | MessageID = 0x4ca5e939 | vkCreateImage: parameter pCreateInfo->mipLevels is zero

and when dealing with multi-device VUs it still works as normal

> Validation Error: [ VUID-vkCmdBindPipeline-commonparent ] Object 0: handle = 0x5567d5eae030, type = VK_OBJECT_TYPE_INSTANCE; Object 1: handle = 0x5567d6c32c70, type = VK_OBJECT_TYPE_DEVICE; Object 2: handle = 0x5567d6cc5c80, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0xa6c0eb9e | vkCmdBindPipeline(): Expected all Dispatchable Handles to use VkDevice 0x5567d6c32c70[], but the VkPipeline (0xf443490000000006) was created, allocated or retrieved from VkDevice 0x5567d6cc5c80[].

### DIFF
--- a/layers/error_message/logging.cpp
+++ b/layers/error_message/logging.cpp
@@ -91,13 +91,18 @@ static bool debug_log_msg(const debug_report_data *debug_data, VkFlags msg_flags
     // stability for pushed strings.
     object_labels.reserve(objects.object_list.size());
 
-    std::vector<VkDebugUtilsObjectNameInfoEXT> object_name_info;
-    object_name_info.resize(objects.object_list.size());
+    std::vector<VkDebugUtilsObjectNameInfoEXT> object_name_infos;
+    object_name_infos.reserve(objects.object_list.size());
     for (uint32_t i = 0; i < objects.object_list.size(); i++) {
-        object_name_info[i] = LvlInitStruct<VkDebugUtilsObjectNameInfoEXT>();
-        object_name_info[i].objectType = ConvertVulkanObjectToCoreObject(objects.object_list[i].type);
-        object_name_info[i].objectHandle = objects.object_list[i].handle;
-        object_name_info[i].pObjectName = NULL;
+        // If only one VkDevice was created, it is just noise to print it out in the error message
+        if (objects.object_list[i].type == kVulkanObjectTypeDevice && debug_data->device_created <= 1) {
+            continue;
+        }
+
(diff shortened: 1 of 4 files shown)


layers/error_message/logging.cpp
```diff
@@ -91,13 +91,18 @@ static bool debug_log_msg(const debug_report_data *debug_data, VkFlags msg_flags
     // stability for pushed strings.
     object_labels.reserve(objects.object_list.size());
 
-    std::vector<VkDebugUtilsObjectNameInfoEXT> object_name_info;
-    object_name_info.resize(objects.object_list.size());
+    std::vector<VkDebugUtilsObjectNameInfoEXT> object_name_infos;
+    object_name_infos.reserve(objects.object_list.size());
```
arno-lunarg, line 95: Perf nit: If I remember correctly @artem-lunarg made a comment on some PR I made not so long ago saying that when numbers of objects to be allocated is known, it is more efficient to use `std::vector::resize` and then reference elements using `vector[i]` than to `std::vector::reserve` and push back elements

arno-lunarg (approved):
LGTM! 1 suggestion

layers/error_message/logging.cpp
spencer-lunarg: > when numbers of objects to be allocated is known

We don't know the size anymore, we look and skip all `VkDevice` objects, if you leave as `resize` it will have garbage in the end of the vector when we skip

also this is during error logging and these are normally at most 3 or 4 items in this array (most time its just 1)

Issue #6274, opened 2023-08-07
VUID-vkCmdPipelineBarrier-shaderTileImageColorReadAccess-08718 does not check if VK_EXT_shader_tile_image is enabled


ziga-lunarg:
The VUID is reported even if the extension is not enabled.

spencer-lunarg:
So this is checked as 

```c++
    // Check shader tile image features
    const auto tile_image_features = enabled_features.shader_tile_image_features;
    const bool features_enabled = tile_image_features.shaderTileImageColorReadAccess ||
                                  tile_image_features.shaderTileImageDepthReadAccess ||
                                  tile_image_features.shaderTileImageStencilReadAccess;

    if (!features_enabled) {
        logError(08718);
    }
 ```
 
 so the only way these features are set in state tracker is if you set `VkPhysicalDeviceShaderTileImageFeaturesEXT` which you need `VK_EXT_shader_tile_image` as well... but trying to write a test, I see no error at `VkCreateDevice` time like I would expect

spencer-lunarg:
ok, so this looks like a spec bug

we shouldn't have to be checking extensions, we should be checking features bits for the VUs

so I will try and re-work this VU, it should have 2 VUs saying something like something like

> If vkCmdPipelineBarrier2 is called within a render pass instance the render pass must not have been started with vkCmdBeginRendering 
>
> If these {{any features}} are enabled, it must adhere to the restrictions in
Explicit Render Pass Tile Image Access Synchronization

PR #6280, opened 2023-08-08
layers: Improve VK_EXT_shader_tile_image feature VU message

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6274

If you are looking at the current extension-less spec there is a `VUID-vkCmdPipelineBarrier2-None-06191` that says

> If vkCmdPipelineBarrier2 is called within a render pass instance, the render pass must not have been started with vkCmdBeginRendering

so not enabling the feature bits in `VK_EXT_shader_tile_image` still should produce the `VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718` error

----

The error use to say

> vkCmdPipelineBarrier2 can not be used if none of the features of tile image read is enabled.

but now says

> vkCmdPipelineBarrier2 can not be called inside a dynamic rendering instance. This can be fixed by enabling the VK_EXT_shader_tile_image features.

to help explain what is going on better

-----
I also have [spec language](https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6065) PR collaspe `06191` and `08718` to just be `08718` to avoid future confusion

### DIFF
--- a/layers/core_checks/cc_synchronization.cpp
+++ b/layers/core_checks/cc_synchronization.cpp
@@ -2440,7 +2440,9 @@ bool CoreChecks::ValidateBarriersForShaderTileImage(const LogObjectList &objlist
 
     if (!features_enabled) {
         const auto &vuid = GetShaderTileImageVUID(outer_loc, ShaderTileImageError::kShaderTileImageFeatureError);
-        skip |= LogError(objlist, vuid, "%s can not be used if none of the features of tile image read is enabled.",
+        skip |= LogError(objlist, vuid,
+                         "%s can not be called inside a dynamic rendering instance. This can be fixed by enabling the "
+                         "VK_EXT_shader_tile_image features.",
                          outer_loc.StringFunc());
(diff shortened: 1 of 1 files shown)


artem-lunarg (approved):

artem-lunarg:
Hm.. I read this that barriers can not be used inside dynamic render pass.

We have also in the spec:
`If vkCmdPipelineBarrier2 is recorded within a render pass instance, the synchronization scopes are limited to operations within the same subpass .`

Could you please clarify what am I missing here?

ziga-lunarg (approved):

spencer-lunarg:
@artem-lunarg so to answer your questions

https://github.com/KhronosGroup/Vulkan-Docs/blob/main/proposals/VK_KHR_dynamic_rendering.adoc#54-should-pipeline-barriers-work-inside-these-limited-render-passes

image

so this is as expected

PR #6281, opened 2023-08-08
layers: Add 09200 09201 09202

arno-lunarg:
Also update related error messages

### DIFF
--- a/layers/core_checks/cc_device_memory.cpp
+++ b/layers/core_checks/cc_device_memory.cpp
@@ -617,18 +617,44 @@ bool CoreChecks::ValidateBindBufferMemory(VkBuffer buffer, VkDeviceMemory mem, V
                              "memory must have been allocated with the VK_MEMORY_ALLOCATE_DEVICE_ADDRESS_BIT bit set.",
                              api_name);
         }
+        const VkMemoryAllocateFlags memory_allocate_flags = chained_flags_struct ? chained_flags_struct->flags : 0;
+        if (buffer_state->createInfo.flags & VK_BUFFER_CREATE_DESCRIPTOR_BUFFER_CAPTURE_REPLAY_BIT_EXT) {
+            if (!(memory_allocate_flags & VK_MEMORY_ALLOCATE_DEVICE_ADDRESS_BIT)) {
+                const char *vuid = bind_buffer_mem_2 ? "VUID-VkBindBufferMemoryInfo-descriptorBufferCaptureReplay-08112"
+                                                     : "VUID-vkBindBufferMemory-descriptorBufferCaptureReplay-08112";
+                const LogObjectList objlist(buffer, mem);
+                skip |=
+                    LogError(objlist, vuid,
+                             "%s: Buffer was created with the VK_BUFFER_CREATE_DESCRIPTOR_BUFFER_CAPTURE_REPLAY_BIT_EXT bit set,"
+                             "but the bound memory was allocated with %s and needs VK_MEMORY_ALLOCATE_DEVICE_ADDRESS_BIT.",
+                             api_name, string_VkMemoryAllocateFlags(memory_allocate_flags).c_str());
+            }
 
-        if (enabled_features.descriptor_buffer_features.descriptorBufferCaptureReplay &&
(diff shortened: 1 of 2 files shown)


layers/core_checks/cc_device_memory.cpp
```diff
@@ -618,17 +618,48 @@ bool CoreChecks::ValidateBindBufferMemory(VkBuffer buffer, VkDeviceMemory mem, V
...
-                                                 : "VUID-vkBindBufferMemory-descriptorBufferCaptureReplay-08112";
-            const LogObjectList objlist(buffer, mem);
-            skip |=
-                LogError(objlist, vuid,
-                         "%s: If buffer was created with the VK_BUFFER_CREATE_DESCRIPTOR_BUFFER_CAPTURE_REPLAY_BIT_EXT bit set, "
-                         "memory must have been allocated with the VK_MEMORY_ALLOCATE_DEVICE_ADDRESS_BIT bit set.",
-                         api_name);
+        if (enabled_features.descriptor_buffer_features.descriptorBufferCaptureReplay) {
+            if (buffer_state->createInfo.flags & VK_BUFFER_CREATE_DESCRIPTOR_BUFFER_CAPTURE_REPLAY_BIT_EXT) {
+                if ((!chained_flags_struct || !(chained_flags_struct->flags & VK_MEMORY_ALLOCATE_DEVICE_ADDRESS_BIT))) {
```
spencer-lunarg, line 623: we could just do a

`VkMemoryAllocateFlags flags = chained_flags_struct ? chained_flags_struct->flags : 0;`

that way we don't have to check each time for `!chained_flags_struct` and the error message if you use `string_VkMemoryAllocateFlags(0)` will print `VkMemoryAllocateFlags(0)` which I find it more useful then printing "None"

layers/core_checks/cc_device_memory.cpp
```diff
@@ -618,17 +618,48 @@ bool CoreChecks::ValidateBindBufferMemory(VkBuffer buffer, VkDeviceMemory mem, V
...
+                    const char *vuid = bind_buffer_mem_2 ? "VUID-VkBindBufferMemoryInfo-descriptorBufferCaptureReplay-08112"
+                                                         : "VUID-vkBindBufferMemory-descriptorBufferCaptureReplay-08112";
+                    const LogObjectList objlist(buffer, mem);
+                    skip |= LogError(
+                        objlist, vuid,
+                        "%s: Buffer was created with the VK_BUFFER_CREATE_DESCRIPTOR_BUFFER_CAPTURE_REPLAY_BIT_EXT bit set, but"
+                        "bound memory was not allocated with the VK_MEMORY_ALLOCATE_DEVICE_ADDRESS_BIT bit set. It was allocated "
+                        "with the following bit(s) set: %s.",
```
spencer-lunarg, line 631: I think these could all be a little simpler to read if we did something like

```
"%s: Buffer was created with the VK_BUFFER_CREATE_DESCRIPTOR_BUFFER_CAPTURE_REPLAY_BIT_EXT bit set,"
"but the bound memory was allocated with %s and needs VK_MEMORY_ALLOCATE_DEVICE_ADDRESS_BIT.",
```

arno-lunarg: Yes, wording is better!

layers/core_checks/cc_device_memory.cpp
```diff
@@ -617,18 +617,45 @@ bool CoreChecks::ValidateBindBufferMemory(VkBuffer buffer, VkDeviceMemory mem, V
                              "memory must have been allocated with the VK_MEMORY_ALLOCATE_DEVICE_ADDRESS_BIT bit set.",
                              api_name);
         }
+        const VkMemoryAllocateFlags memory_allocate_flags = chained_flags_struct ? chained_flags_struct->flags : 0;
+        if (buffer_state->createInfo.flags & VK_BUFFER_CREATE_DESCRIPTOR_BUFFER_CAPTURE_REPLAY_BIT_EXT) {
+            if ((!chained_flags_struct || !(chained_flags_struct->flags & VK_MEMORY_ALLOCATE_DEVICE_ADDRESS_BIT))) {
```
spencer-lunarg, line 622:
```suggestion
            if (!(memory_allocate_flags & VK_MEMORY_ALLOCATE_DEVICE_ADDRESS_BIT)) {
```

spencer-lunarg (approved):

PR #6284, opened 2023-08-09
Add VkBufferUsageFlags2CreateInfoKHR support (VK_KHR_maintenance5)

spencer-lunarg:
Part 3 of `VK_KHR_maintenance5`

Adds `VkBufferUsageFlags2CreateInfoKHR` support which is designed for now to be used at `vkCreateBuffer` and `vkCreateBufferView` time 

The spec says for `vkCreateBuffer`

> If a [VkBufferUsageFlags2CreateInfoKHR](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkBufferUsageFlags2CreateInfoKHR) structure is present in the pNext chain, [VkBufferUsageFlags2CreateInfoKHR](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkBufferUsageFlags2CreateInfoKHR)::usage from that structure is used **instead** of usage from this structure.

### DIFF
--- a/layers/core_checks/cc_device_memory.cpp
+++ b/layers/core_checks/cc_device_memory.cpp
@@ -524,7 +524,8 @@ bool CoreChecks::ValidateBindBufferMemory(VkBuffer buffer, VkDeviceMemory mem, V
         if (mem_info->IsExport()) {
             auto external_info = LvlInitStruct<VkPhysicalDeviceExternalBufferInfo>();
             external_info.flags = buffer_state->createInfo.flags;
-            external_info.usage = buffer_state->createInfo.usage;
+            // for now no VkBufferUsageFlags2KHR flag can be used, so safe to pass in as 32-bit version
+            external_info.usage = VkBufferUsageFlags(buffer_state->usage);
             auto external_properties = LvlInitStruct<VkExternalBufferProperties>();
             bool export_supported = true;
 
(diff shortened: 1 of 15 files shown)


layers/core_checks/cc_device_memory.cpp
```diff
@@ -524,7 +524,8 @@ bool CoreChecks::ValidateBindBufferMemory(VkBuffer buffer, VkDeviceMemory mem, V
         if (mem_info->IsExport()) {
             auto external_info = LvlInitStruct<VkPhysicalDeviceExternalBufferInfo>();
             external_info.flags = buffer_state->createInfo.flags;
-            external_info.usage = buffer_state->createInfo.usage;
+            // for now no VkBufferUsageFlags2KHR flag can be used, so safe to pass in as 32-bit version
+            external_info.usage = VkBufferUsageFlags(buffer_state->usage);
```
artem-lunarg, line 528: if there is a possibility that in the future 64 bit value will be used (I don't know!) then assert can do runtime detection of values > 32 bit.

layers/state_tracker/buffer_state.cpp
```diff
@@ -32,11 +32,49 @@ static VkMemoryRequirements GetMemoryRequirements(ValidationStateTracker *dev_da
...
+        }
+        export_metal_object_info = LvlFindInChain<VkExportMetalObjectCreateInfoEXT>(export_metal_object_info->pNext);
+    }
+    return retval;
+}
+#endif
+
+static VkBufferUsageFlags2KHR GetBufferUsageFlags(const VkBufferViewCreateInfo &create_info) {
+    const auto *usage_flags2 = LvlFindInChain<VkBufferUsageFlags2CreateInfoKHR>(create_info.pNext);
+    return usage_flags2 ? usage_flags2->usage : 0;
```
artem-lunarg, line 67: A remark to to confirm that 0 is fine and we don't to extract usage from the `buffer`

artem-lunarg (approved):
LGTM. One question.

layers/state_tracker/buffer_state.cpp
spencer-lunarg: ... ok, so I realize we never have any use for keeping the state of `VkBufferUsageFlags2KHR` for the `BUFFER_VIEW_STATE`, just removed it (including this function)

PR #6287, opened 2023-08-10
build: Don't pass -std= in BUILD.gn

nico:
All embedders use C++17 or newer now, and
language version should generally be set by the
embedder anyways.

This also happens to work around a bug in libc++
which currently doesn't work correctly in mixed
language version modes (https://crbug.com/1463881).

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -64,12 +64,6 @@ config("vulkan_internal_config") {
     "VK_ENABLE_BETA_EXTENSIONS",
   ]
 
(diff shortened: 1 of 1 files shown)


ncesario-lunarg (approved):

jeremyg-lunarg (approved):

jeremyg-lunarg:
maintainer push to rebase

nico:
> What is an embedder? Is there a doc commenting this?

A program using this library.

PR #6288, opened 2023-08-10
layers: Add ErrorObject

spencer-lunarg:
This adds a `ErrorObject` that is designed to be a replacement for passing various parameters through functions for the sake of logging them.

The main idea is that in the same way we have a `VkResult` in all `PostCallRecord`, we would add a `ErrorObject` on the stack when the chassis calls all `PreCallValidate` functions

details in new `error_object.md` doc file

- renamed `core_error` to just `vvl`
    - it is used in Best Practice and found no value in pratice having the namespace
    - renamed `core_error_location.cpp` to just `error_location.cpp`
- Reorder new `LogError` arguments
    - It makes more logcial sense when writting the errors to do
        - VUID
        - LogObjectList
        - Location
        - Error message
    - This is both how the error message is generated, anad makes it easier to see how `Location` and the error message connects


layers/core_checks/cc_device_memory.cpp
```diff
@@ -1613,159 +1612,160 @@ void CoreChecks::PostCallRecordBindImageMemory2KHR(VkDevice device, uint32_t bin
...
 
-bool CoreChecks::ValidateSparseMemoryBind(const VkSparseMemoryBind &bind, VkDeviceSize resource_size, const char *func_name,
-                                          const char *parameter_name) const {
+bool CoreChecks::ValidateSparseMemoryBind(const VkSparseMemoryBind &bind, VkDeviceSize resource_size, const Location &loc) const {
     bool skip = false;
     auto mem_info = Get<DEVICE_MEMORY_STATE>(bind.memory);
     if (mem_info) {
         if (phys_dev_mem_props.memoryTypes[mem_info->alloc_info.memoryTypeIndex].propertyFlags &
             VK_MEMORY_PROPERTY_LAZILY_ALLOCATED_BIT) {
-            skip |= LogError(bind.memory, "VUID-VkSparseMemoryBind-memory-01097",
```
spencer-lunarg, line 1623: new error 

> Validation Error: [ VUID-VkSparseMemoryBind-memory-01097 ] | MessageID = 0x8c1f718 | vkQueueBindSparse(): pBindInfo[0].pBufferBinds[0].pBinds[0] memory type has VK_MEMORY_PROPERTY_LAZILY_ALLOCATED_BIT bit set.

layers/core_checks/cc_render_pass.cpp
```diff
@@ -2540,77 +2488,64 @@ bool CoreChecks::ValidateDepthStencilResolve(const VkRenderPassCreateInfo2 *pCre
...
         // Same VU, but better error message if one of the resolves are ignored
         if (resolve_has_depth && !resolve_has_stencil && resolve->depthResolveMode == VK_RESOLVE_MODE_NONE) {
-            skip |= LogError(device, "VUID-VkSubpassDescriptionDepthStencilResolve-pDepthStencilResolveAttachment-03178",
-                             "%s: Subpass %" PRIu32
-                             " includes a VkSubpassDescriptionDepthStencilResolve structure with resolve "
-                             "attachment %" PRIu32
-                             ", but the depth resolve mode is VK_RESOLVE_MODE_NONE (stencil resolve mode is "
-                             "ignored due to format not having stencil component).",
-                             function_name, i, resolve_attachment);
+            skip |= LogError("VUID-VkSubpassDescriptionDepthStencilResolve-pDepthStencilResolveAttachment-03178", device,
```
spencer-lunarg, line 2536: new error

> Validation Error: [ VUID-VkSubpassDescriptionDepthStencilResolve-pDepthStencilResolveAttachment-03178 ] | MessageID = 0xd3523b67 | vkCreateRenderPass2KHR(): pSubpasses[0].pNext<VkSubpassDescriptionDepthStencilResolve>.pDepthStencilResolveAttachment[1] is not NULL, but both depth and stencil resolve modes are VK_RESOLVE_MODE_NONE.

layers/core_checks/cc_render_pass.cpp
```diff
@@ -1386,55 +1375,44 @@ bool CoreChecks::ValidateRenderpassAttachmentUsage(RenderPassCreateVersion rp_ve
...
+                skip |= ValidateAttachmentIndex(rp_version, attachment_index, pCreateInfo->attachmentCount, input_loc);
 
                 if (aspect_mask & VK_IMAGE_ASPECT_METADATA_BIT) {
                     const char *vuid = use_rp2 ? "VUID-VkSubpassDescription2-attachment-02801"
                                                : "VUID-VkInputAttachmentAspectReference-aspectMask-01964";
-                    skip |= LogError(
-                        device, vuid,
-                        "%s: Aspect mask for input attachment reference %d in subpass %d includes VK_IMAGE_ASPECT_METADATA_BIT.",
-                        function_name, j, i);
+                    skip |= LogError(vuid, device, input_loc, "aspect mask includes VK_IMAGE_ASPECT_METADATA_BIT.");
```
spencer-lunarg, line 1385: > Validation Error: [ VUID-VkInputAttachmentAspectReference-aspectMask-01964 ] | MessageID = 0x85eb158d | vkCreateRenderPass(): pSubpasses[0].pInputAttachments[0] aspect mask includes VK_IMAGE_ASPECT_METADATA_BIT.

layers/core_checks/cc_render_pass.cpp
```diff
@@ -1880,12 +1835,11 @@ bool CoreChecks::ValidateRenderpassAttachmentUsage(RenderPassCreateVersion rp_ve
                         ms_render_to_single_sample && ms_render_to_single_sample->multisampledRenderToSingleSampledEnable) {
                         if (current_sample_count != VK_SAMPLE_COUNT_1_BIT &&
                             current_sample_count != ms_render_to_single_sample->rasterizationSamples) {
-                            skip |= LogError(device, "VUID-VkSubpassDescription2-pNext-06870",
-                                             "%s:  Subpass %u has a VkMultisampledRenderToSingleSampledInfoEXT struct in its "
-                                             "VkSubpassDescription2 pNext chain with multisampledRenderToSingleSampled set to "
-                                             "VK_TRUE and rasterizationSamples set to %s "
+                            skip |= LogError("VUID-VkSubpassDescription2-pNext-06870", device,
```
spencer-lunarg, line 1838: > Validation Error: [ VUID-VkSubpassDescription2-pNext-06870 ] | MessageID = 0x992b9193 | vkCreateRenderPass2(): pSubpasses[0].pNext<VkMultisampledRenderToSingleSampledInfoEXT>.multisampledRenderToSingleSampled is VK_TRUE and rasterizationSamples set to VK_SAMPLE_COUNT_2_BIT but color attachment ref 0 has a sample count of VK_SAMPLE_COUNT_4_BIT.

layers/core_checks/cc_device_memory.cpp
```diff
@@ -1613,159 +1612,160 @@ void CoreChecks::PostCallRecordBindImageMemory2KHR(VkDevice device, uint32_t bin
...
-                                 "(VkSparseImageFormatProperties::imageGranularity.height (%" PRIu32 ")) of the image",
-                                 bind_idx, image_idx, bind.offset.y, granularity.height);
+                                 "(VkSparseImageFormatProperties::imageGranularity.height (%" PRIu32 ")) of the image.",
+                                 bind.offset.y, granularity.height);
             }
 
             if (SafeModulo(bind.offset.z, granularity.depth) != 0) {
-                skip |= LogError(image_state->Handle(), "VUID-VkSparseImageMemoryBind-offset-01111",
-                                 "vkQueueBindSparse(): pImageBinds[%" PRIu32 "].pBindInfo[%" PRIu32 "]: offset.z (%" PRIi32
+                skip |= LogError("VUID-VkSparseImageMemoryBind-offset-01111", image_state->Handle(),
```
spencer-lunarg, line 1725: > Validation Error: [ VUID-VkSparseImageMemoryBind-offset-01111 ] Object 0: handle = 0xfa21a40000000003, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x25173401 | vkQueueBindSparse(): pBindInfo[0].pImageBinds[0].offset.z (3) must be a multiple of the sparse image block depth (VkSparseImageFormatProperties::imageGranularity.depth (4)) of the image

layers/core_checks/cc_render_pass.cpp
```diff
@@ -2452,37 +2396,41 @@ bool CoreChecks::PreCallValidateCreateRenderPass(VkDevice device, const VkRender
...
+                        .dot(Field::subpass),
+                    "is %" PRIu32 " which is not less than pCreateInfo->subpassCount (%" PRIu32 ").", subpass,
+                    pCreateInfo->subpassCount);
             } else if (pCreateInfo->pSubpasses && attachment >= pCreateInfo->pSubpasses[subpass].inputAttachmentCount) {
-                skip |= LogError(device, "VUID-VkRenderPassCreateInfo-pNext-01927",
-                                 "vkCreateRenderPass(): Input attachment index %u specified by input attachment aspect info %u is "
-                                 "greater than the "
-                                 "input attachment count of %u for this subpass.",
-                                 attachment, i, pCreateInfo->pSubpasses[subpass].inputAttachmentCount);
+                skip |= LogError(
```
spencer-lunarg, line 2406: > Validation Error: [ VUID-VkRenderPassCreateInfo-pNext-01927 ] | MessageID = 0x8e1c8f13 | vkCreateRenderPass(): pNext<VkRenderPassInputAttachmentAspectCreateInfo>.pAspectReferences[0].inputAttachmentIndex is 1 which not less then pCreateInfo->pSubpasses[0].inputAttachmentCount (1).

jeremyg-lunarg (commented):
This looks really good to me! But I'm very biased about the error location code. :)

docs/error_object.md
```diff
@@ -0,0 +1,68 @@
...
+```cpp
+// example
+
+bool PreCallValidateQueueBindSparse(/*..*/ ErrorObject &errorObj) const {
+    // automatically has Func::vkQueueBindSparse
+    LogError("VUID-A", errorObj.location);
+
+    for (uint32_t i = 0; i < bindInfoCount; ++i) {
+        const Location bind_info_loc(Field::pBindInfo, i);
+        LogError("VUID-B", bind_info_loc); // i == 3
```
artem-lunarg, line 26: Do these examples need `.dot` calls or similar to chain with a parent?

layers/core_checks/cc_descriptor.cpp
```diff
@@ -5306,11 +5306,10 @@ bool CoreChecks::PreCallValidateCmdPushConstants(VkCommandBuffer commandBuffer,
         }
         if (found_stages != stageFlags) {
             uint32_t missing_stages = ~found_stages & stageFlags;
-            skip |= LogError(
-                commandBuffer, "VUID-vkCmdPushConstants-offset-01795",
-                "vkCmdPushConstants(): %s, VkPushConstantRange in %s overlapping offset = %d and size = %d, do not contain %s.",
-                string_VkShaderStageFlags(stageFlags).c_str(), FormatHandle(layout).c_str(), offset, size,
-                string_VkShaderStageFlags(missing_stages).c_str());
+            skip |= LogError("VUID-vkCmdPushConstants-offset-01795", commandBuffer, errorObj.location,
```
artem-lunarg, line 5309: Comment about order. `errorObj.location` is a static code and it is between VUID and error message which are dynamic parts (text changes for each error report) . When reading it might be more convenient to group dynamic elements together. That's the logic. But it's hard to say if it's good in practice. I'll write down one alternative that has different grouping for visibility:  `LogError(errorObj.location, objList, VUID, message);`

layers/error_message/error_location.h
```diff
@@ -23,57 +23,67 @@
...
+    const vvl::Func function;
 
-    Struct structure;
-    Field field;
-    // optional index if checking an array.
-    uint32_t index;
+    const vvl::Struct structure;
+    const vvl::Field field;
+    const uint32_t index;  // optional index if checking an array.
+    const bool pNext;      // will print out it is from a pNext chain
```
artem-lunarg, line 41: Is it for Vulkan's pNext chain or terminology for linking Location objects?

docs/error_object.md
```diff
@@ -0,0 +1,68 @@
...
+[VUID-B] vkQueueBindSparse(): pBindInfo[3]
+[VUID-C] vkQueueBindSparse(): pBindInfo[3].pBufferBinds[2]
+[VUID-D] vkQueueBindSparse(): pBindInfo[3].pNext<VkTimelineSemaphoreSubmitInfo>.waitSemaphoreValueCount
+```
+
+> We generate the `Func`/`Struct`/`Field` for all possible items from the XML
+
+### Using stack and making copies
+
+The original design was to not make copies of `Location` and just have it modify the `ErrorObject::Location` but this would require you to remove items after a function call
```
artem-lunarg, line 55: +1. Modification of single Location object also involves copying the same amount of information (new data should be stored somewhere), but also adds additional bookkeeping overhead (if fixed size array is used then index should be incremented).

Automatic removal can be less of a problem comparing to efficiency, can be done with raiii. But current approach looks very efficient, and simpler comparing to that alternatives. Also looks quite robust because as I understand you can use this immutable style - create const object, then other object can reference it and also be const.

docs/error_object.md
spencer-lunarg: ya yes, good catch, I have other improvements for this doc in the work

layers/core_checks/cc_descriptor.cpp
spencer-lunarg: to be fair, the `objList` and `Location` can change between each error message

my goal was the match how things are printed. Also since the `Location` and what gets printed before the error message, it makes sense when writing the error message what is about to be printed in front of it

layers/error_message/error_location.h
spencer-lunarg: I wanted a way to know if we need to print `pNext<SomeStruct>`... open to better names (or comment)

layers/core_checks/cc_descriptor.cpp
artem-lunarg: > to be fair, the objList and Location can change between each error message
Is it about the change of the variable's value? If yes, then I meant a bit different thing, the text of the code itself, that we might type static text like "errorObj.location" and it's similar, in contrast to VUID and message. In this alternative the ordering is how closely the  semantics of parameters are associated with the message. But I see the point of ordering according to message structure, did not realize this initially.

layers/error_message/error_location.h
artem-lunarg: Missed the example you highlighted above, but after checking it it's clear how it works: `pSubpasses[0].pNext<VkMultisampledRenderToSingleSampledInfoEXT>.multisampledRenderToSingleSampled`
Can we add example like this in the comment?

docs/error_object.md
```diff
@@ -0,0 +1,107 @@
...
+```cpp
+const Location loc = errorObj.location.dot(Field::pBindInfo, i); // vkQueueBindSparse(): pBindInfo[3]
+
+// prints "pBindInfo[3]"
+LogError(/*..*/, "%s". loc.Fields().c_str());
+```
+
+### Limitations
+
+When using the `.dot()` operation, it returns a new copy of `Location`. If you chain two `.dot().dot()` you need to be mindful.
```
artem-lunarg, line 67: > If you chain two `.dot().dot()` you need to be mindful.

Good point. Are there use cases when it's valid? Even if there are valid use cases still one option to consider is to restrict the flexibility and to declare method in the form `return_value dot(params) &&` that is called only for `rvalues` and do not provide implementation, so it causes compile time error. Did not try if it works for this use case though.

layers/stateless/sl_device_memory.cpp
```diff
@@ -90,44 +92,48 @@ bool StatelessValidation::manual_PreCallValidateAllocateMemory(VkDevice device,
 
         auto opaque_alloc_info = LvlFindInChain<VkMemoryOpaqueCaptureAddressAllocateInfo>(pAllocateInfo->pNext);
         if (opaque_alloc_info && opaque_alloc_info->opaqueCaptureAddress != 0) {
+            const Location address_loc =
+                loc.dot(Struct::VkMemoryOpaqueCaptureAddressAllocateInfo, Field::opaqueCaptureAddress, true);
```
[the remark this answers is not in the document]
jeremyg-lunarg: or just a pNext() method for the true case.

layers/error_message/error_location.h
```diff
@@ -148,12 +159,30 @@ static const std::string& FindVUID(const Location& loc, const Table& table) {
 // different VUIDs for different values of an enum or bitfield
 template <typename OuterKey, typename Table>
 static const std::string& FindVUID(OuterKey key, const Location& loc, const Table& table) {
+    // Currently need to squash all the KHR versions to find in the table
+    // Only need for functions because they are now generated in the chassis
```
jeremyg-lunarg, line 163: seems like this process could be codegened in a future PR

docs/error_object.md
jeremyg-lunarg: It would be handy in cases where you have to immediately drill down far into a rats nest of structures to check just 1 VUID. Usually you're checking lots of things in sequence and having a variable for each level of dot() just happens naturally.

We were still using C++11 when I wrote this so if there are more modern ways to write this function that remove this limitation I'm all for them.

artem-lunarg (approved):
LGTM! Checked only subset of files. Infrastructure code and some of the updated files.

docs/error_object.md
spencer-lunarg: tbh, adding all these caes, I found almost never did I need to do this

I would always just go `LogError(/*..*/, loc.dot(Field::A).dot(Field::B).dot(Field::C), ... );`

there is rarely a case you need 2 or 3 deep location to be a variable on the stack I found

layers/stateless/sl_device_memory.cpp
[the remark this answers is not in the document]
spencer-lunarg: ya, `.pNext(Struct::VkMemoryOpaqueCaptureAddressAllocateInfo, Field::opaqueCaptureAddress);` is smart

will update

arno-lunarg (approved):

spencer-lunarg:
another (of many I feel) cool side effect of this is how we alias the promoted functions

before

```cpp
bool CoreChekc::ValidateFunction2(VkDevice device, VkInfo info, const char* api_name) const {
    LogError();
}
bool CoreChekc::PreCallValidateFunction2(VkDevice device, VkInfo info) const {
    ValidateFunction2(device, info, "Function2");
}
bool CoreChekc::PreCallValidateFunction2KHR(VkDevice device, VkInfo info) const {
    ValidateFunction2(device, info, "Function2KHR");
}
```

is now just

```cpp
bool CoreChekc::PreCallValidateFunction2(VkDevice device, VkInfo info, const ErrorObject &errorObj) const {
    LogError();
}
bool CoreChekc::PreCallValidateFunction2KHR(VkDevice device, VkInfo info, const ErrorObject &errorObj) const {
    return PreCallValidateFunction2(commandBuffer, pCopyBufferInfo, errorObj);
}
```

we could even go as far as doing this in the chassis and not even need the `PreCallValidateFunction2KHR` at all

(also this `ErrorObject` is going to kill a lot of redundant half-attempts like `ParameterName` and `CMD_TYPE`)

PR #6289, opened 2023-08-10
Fix VUID 02688

ziga-lunarg:
This VUID has to be checked with dynamic rendering too

And `enabled_features.core11` should be checked instead of `enabled_features.multiview_features`

### DIFF
--- a/layers/core_checks/cc_drawdispatch.cpp
+++ b/layers/core_checks/cc_drawdispatch.cpp
@@ -2711,14 +2711,14 @@ bool CoreChecks::ValidateCmdDrawInstance(const CMD_BUFFER_STATE &cb_state, uint3
     const char *caller = CommandTypeString(cmd_type);
 
     // Verify maxMultiviewInstanceIndex
-    if (cb_state.activeRenderPass->renderPass() && enabled_features.multiview_features.multiview &&
-        ((instanceCount + firstInstance) > phys_dev_ext_props.multiview_props.maxMultiviewInstanceIndex)) {
+    if (cb_state.activeRenderPass && enabled_features.core11.multiview &&
+        ((static_cast<uint64_t>(instanceCount) + static_cast<uint64_t>(firstInstance)) >
+         static_cast<uint64_t>(phys_dev_ext_props.multiview_props.maxMultiviewInstanceIndex))) {
         const LogObjectList objlist(cb_state.Handle(), cb_state.activeRenderPass->Handle());
         skip |= LogError(objlist, vuid.max_multiview_instance_index_02688,
-                         "%s: renderpass %s multiview is enabled, and maxMultiviewInstanceIndex: %" PRIu32
+                         "%s: renderpass instance has multiview enabled, and maxMultiviewInstanceIndex: %" PRIu32
                          ", but instanceCount: %" PRIu32 "and firstInstance: %" PRIu32 ".",
-                         caller, FormatHandle(cb_state.activeRenderPass->Handle()).c_str(),
-                         phys_dev_ext_props.multiview_props.maxMultiviewInstanceIndex, instanceCount, firstInstance);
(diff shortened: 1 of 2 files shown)


layers/core_checks/cc_drawdispatch.cpp
```diff
@@ -2711,11 +2711,12 @@ bool CoreChecks::ValidateCmdDrawInstance(const CMD_BUFFER_STATE &cb_state, uint3
     const char *caller = CommandTypeString(cmd_type);
 
     // Verify maxMultiviewInstanceIndex
-    if (cb_state.activeRenderPass->renderPass() && enabled_features.multiview_features.multiview &&
-        ((instanceCount + firstInstance) > phys_dev_ext_props.multiview_props.maxMultiviewInstanceIndex)) {
+    if (cb_state.activeRenderPass && enabled_features.core11.multiview &&
+        ((static_cast<uint64_t>(instanceCount) + static_cast<uint64_t>(firstInstance)) >
+         static_cast<uint64_t>(phys_dev_ext_props.multiview_props.maxMultiviewInstanceIndex))) {
```
spencer-lunarg, line 2716: (thinking out loud) I am wondering if we should have some helper for checking against max limits for things that likely to have a max near UINT32_MAX

tests/unit/multiview.cpp
```diff
@@ -1378,3 +1378,66 @@ TEST_F(NegativeMultiview, FeaturesDisabled) {
...
+    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-vkCmdDraw-maxMultiviewInstanceIndex-02688");
+
+    AddRequiredExtensions(VK_KHR_DYNAMIC_RENDERING_EXTENSION_NAME);
+    SetTargetApiVersion(VK_API_VERSION_1_2);
+    ASSERT_NO_FATAL_FAILURE(InitFramework(m_errorMonitor));
+    if (DeviceValidationVersion() < VK_API_VERSION_1_2) {
+        GTEST_SKIP() << "At least Vulkan version 1.2 is required";
+    }
+
+    VkPhysicalDeviceDynamicRenderingFeaturesKHR dynamic_rendering_features =
```
spencer-lunarg, line 1394: nit: we have started using `auto` for the `LvlInitStruct`

tests/unit/multiview.cpp
```diff
@@ -1378,3 +1378,66 @@ TEST_F(NegativeMultiview, FeaturesDisabled) {
...
+        LvlInitStruct<VkPhysicalDeviceDynamicRenderingFeaturesKHR>();
+    VkPhysicalDeviceVulkan11Features features11 = LvlInitStruct<VkPhysicalDeviceVulkan11Features>(&dynamic_rendering_features);
+    auto features2 = GetPhysicalDeviceFeatures2(features11);
```
spencer-lunarg, line 1397: nit: This can just be  `GetPhysicalDeviceFeatures2(features11);`

And then you go `ASSERT_NO_FATAL_FAILURE(InitState(nullptr, &features11));`

tests/unit/multiview.cpp
```diff
@@ -1378,3 +1378,66 @@ TEST_F(NegativeMultiview, FeaturesDisabled) {
...
+    if (!dynamic_rendering_features.dynamicRendering) {
```
spencer-lunarg, line 1398: we recently decided there is no point to do a 

`if (!dynamic_rendering_features.dynamicRendering) {` check because the spec guarantee it is going to be supported

image

spencer-lunarg (approved):
LGTM - just some "nit how we have been doing tests now" comments

layers/core_checks/cc_drawdispatch.cpp
ziga-lunarg: At first I thought this was not legal to do, have max instance higher than uint32_t::max, but I couldn't find anything in the spec.

PR #6290, opened 2023-08-11
WIP Add validation for VK_EXT_host_image_copy

Tony-LunarG:
No positive test yet - coming soon
Known missing VUIDs:
VUID-VkCopyImageToImageInfoEXT-dstImage-09109
VUID-VkCopyImageToImageInfoEXT-srcImage-09109
VUID-VkCopyImageToMemoryInfoEXT-srcImage-09109
VUID-VkCopyMemoryToImageInfoEXT-srcImage-09109
VUID-VkImageToMemoryCopyEXT-pRegions-09067
VUID-VkImageToMemoryCopyEXT-pHostPointer-09066
VUID-VkMemoryToImageCopyEXT-pRegions-09062
VUID-VkMemoryToImageCopyEXT-pHostPointer-09061

I'll add an issue listing them as todo after this PR is merged


layers/core_checks/cc_device.cpp
```diff
@@ -600,12 +600,32 @@ bool CoreChecks::PreCallValidateGetDeviceQueue2(VkDevice device, const VkDeviceQ
     }
     return skip;
 }
+bool CoreChecks::ValidateGetPhysicalDeviceImageFormatProperties2(const VkPhysicalDeviceImageFormatInfo2 *pImageFormatInfo,
+                                                                 VkImageFormatProperties2 *pImageFormatProperties) const {
+    bool skip = false;
+    const auto *copy_perf_query = LvlFindInChain<VkHostImageCopyDevicePerformanceQueryEXT>(pImageFormatProperties->pNext);
+    if (nullptr != copy_perf_query) {
```
spencer-lunarg, line 607: nit: for consistency, we have just gone either

`if (copy_perf_query) {`

or even

`if (const auto *copy_perf_query = LvlFindInChain<VkHostImageCopyDevicePerformanceQueryEXT>(pImageFormatProperties->pNext) {`

layers/core_checks/cc_device.cpp
```diff
@@ -600,12 +600,32 @@ bool CoreChecks::PreCallValidateGetDeviceQueue2(VkDevice device, const VkDeviceQ
...
+        if (!(enabled_features.host_image_copy_features.hostImageCopy)) {
```
spencer-lunarg, line 608: nit: extra `( )` around this

layers/core_checks/cc_image.cpp
```diff
@@ -2323,6 +2299,139 @@ bool CoreChecks::PreCallValidateGetImageSubresourceLayout2EXT(VkDevice device, V
...
+
+bool CoreChecks::PreCallValidateTransitionImageLayoutEXT(VkDevice device, uint32_t transitionCount,
+                                                         const VkHostImageLayoutTransitionInfoEXT *pTransitions) const {
+    bool skip = false;
+    const char *func_name = "vkTransitionImageLayoutEXT()";
+
+    for (uint32_t i = 0; i < transitionCount; ++i) {
+        auto &transition = pTransitions[i];
+        auto image_state = Get<IMAGE_STATE>(transition.image);
+        auto image_format = image_state->createInfo.format;
```
spencer-lunarg, line 2317: nit: we have `const bool has_depth_mask` and I saw there were not `const` and thought it ment something... can we just label these `const auto` where possible

layers/core_checks/cc_device.cpp
artem-lunarg: Is it from google style guide or internal style? I remember I made opposite comments because thought additional parentheses in  `if ((expr))` was a typo.

tests/framework/layer_validation_tests.cpp
```diff
@@ -394,6 +394,27 @@ VkFormat FindFormatWithoutFeatures(VkPhysicalDevice gpu, VkImageTiling tiling, V
     return return_format;
 }
 
+VkFormat FindFormatWithoutFeatures2(VkPhysicalDevice gpu, VkImageTiling tiling, VkFormatFeatureFlags2 undesired_features) {
+    const VkFormat first_vk_format = static_cast<VkFormat>(1);
+    const VkFormat last_vk_format = static_cast<VkFormat>(130);  // avoid compressed/feature protected, otherwise 184
```
artem-lunarg, line 399: Raw constants are undesirable, you have to jump to the header to validate the code. It might be better to have `vkFormat first_compressed_format = VK_FORMAT_BC1_RGB_UNORM_BLOCK` and then `less` in the below loop for comparison.

tests/framework/layer_validation_tests.cpp
```diff
@@ -394,6 +394,27 @@ VkFormat FindFormatWithoutFeatures(VkPhysicalDevice gpu, VkImageTiling tiling, V
...
+    VkFormat return_format = VK_FORMAT_UNDEFINED;
+    for (VkFormat format = first_vk_format; format <= last_vk_format; format = static_cast<VkFormat>(format + 1)) {
+        auto fmt_props_3 = LvlInitStruct<VkFormatProperties3KHR>();
+        auto fmt_props_2 = LvlInitStruct<VkFormatProperties2>(&fmt_props_3);
+        vk::GetPhysicalDeviceFormatProperties2(gpu, format, &fmt_props_2);
+        fmt_props_3.linearTilingFeatures |= fmt_props_2.formatProperties.linearTilingFeatures;
```
artem-lunarg, line 405: These `or`s are not needed (except we workaround something). The spec says about `VkFormatProperties3`:

> The bits reported in linearTilingFeatures, optimalTilingFeatures and bufferFeatures must include the bits reported in the corresponding fields of VkFormatProperties2::formatProperties.

tests/framework/layer_validation_tests.h
```diff
@@ -1023,6 +1030,9 @@ VkResult GPDIFPHelper(VkPhysicalDevice dev, const VkImageCreateInfo *ci, VkImage
 VkFormat FindFormatWithoutFeatures(VkPhysicalDevice gpu, VkImageTiling tiling,
                                    VkFormatFeatureFlags undesired_features = vvl::kU32Max);
 
+VkFormat FindFormatWithoutFeatures2(VkPhysicalDevice gpu, VkImageTiling tiling,
+                                   VkFormatFeatureFlags2 undesired_features = vvl::kU32Max);
```
artem-lunarg, line 1034: Should it be `vvl:kU64Max` with `VkFormatFeatureFlags2`?

layers/core_checks/cc_device.cpp
```diff
@@ -600,12 +600,32 @@ bool CoreChecks::PreCallValidateGetDeviceQueue2(VkDevice device, const VkDeviceQ
...
+            skip |= LogError(device, "VUID-VkImageFormatProperties2-pNext-09005",
+                             "vkGetPhysicalDeviceImageFormatProperties2(): pImageFormatProperties includes a chained "
+                             "VkHostImageCopyDevicePerformanceQueryEXT struct when the hostImageCopy feature was not enabled");
+        }
+        if ((pImageFormatInfo->usage & VK_IMAGE_USAGE_HOST_TRANSFER_BIT_EXT) == 0) {
+            skip |= LogError(physical_device, "VUID-vkGetPhysicalDeviceImageFormatProperties2-pNext-09004",
+                             "vkGetPhysicalDeviceImageFormatProperties2: pImageFormatProperties includes a chained "
```
artem-lunarg, line 615: nit: `vkGetPhysicalDeviceImageFormatProperties2()` to be consistent with the previous message

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -874,6 +1049,39 @@ bool CoreChecks::ValidateCopyBufferImageTransferGranularityRequirements(const CM
     return skip;
 }
 
+template <typename HandleT>
+bool CoreChecks::ValidateImageSubresourceLayers(HandleT handle, const VkImageSubresourceLayers *subresource_layers,
```
artem-lunarg, line 1053: This PR makes some methods (`ValidateImageSubresourceLayers` is one example)  templated to pass vulkan handle type, which in the end reaches logging routines. Personally, I'm not fan of this direction mostly because many copies of compiled functions in the binary but also because the same can be done without templates.  `LogObjectList` supports `VulkanTypedHandle` arguments and the same functionality can be achieved by using parameter of type `VulkanTypedHandle` instead of templated parameter.

layers/core_checks/cc_image.cpp
```diff
@@ -2323,6 +2299,139 @@ bool CoreChecks::PreCallValidateGetImageSubresourceLayout2EXT(VkDevice device, V
...
+        auto aspect_mask = transition.subresourceRange.aspectMask;
+        const bool has_depth_mask = (aspect_mask & VK_IMAGE_ASPECT_DEPTH_BIT) != 0;
+        const bool has_stencil_mask = (aspect_mask & VK_IMAGE_ASPECT_STENCIL_BIT) != 0;
+
+        if ((image_state->createInfo.usage & VK_IMAGE_USAGE_HOST_TRANSFER_BIT_EXT) == 0) {
+            const LogObjectList objlist(device, image_state->Handle());
+            skip |= LogError(objlist, "VUID-VkHostImageLayoutTransitionInfoEXT-image-09055",
+                             "%s: image in transition %d was not created with VK_IMAGE_USAGE_HOST_TRANSFER_BIT_EXT", func_name, i);
```
artem-lunarg, line 2325: signed type format specified is used (%d). Suggestion: PRIu32

layers/core_checks/cc_image.cpp
```diff
@@ -2323,6 +2299,139 @@ bool CoreChecks::PreCallValidateGetImageSubresourceLayout2EXT(VkDevice device, V
...
+        }
+
+        skip |= ValidateImageSubresourceRange(image_state->createInfo.mipLevels, image_state->createInfo.arrayLayers,
+                                              transition.subresourceRange, func_name, "subresourceRange", "arrayLayers",
+                                              image_state->image(), TransitionImageLayoutVUIDs);
+        skip |=
+            ValidateMemoryIsBoundToImage(device, *image_state, func_name, "VUID-VkHostImageLayoutTransitionInfoEXT-image-01932");
+
+        if (((FormatIsColor(image_format) == true) && (FormatIsMultiplane(image_format) == false)) ||
```
artem-lunarg, line 2334: nit: perf: if `disjoint ==false` is checked first  (which might often evaluate to true) that will be more efficient, FormatIsColor+FormatIsMultiplane can be enough more instructions comparing to single comparision.

layers/core_checks/cc_image.cpp
```diff
@@ -2323,6 +2299,139 @@ bool CoreChecks::PreCallValidateGetImageSubresourceLayout2EXT(VkDevice device, V
...
+            if (aspect_mask != VK_IMAGE_ASPECT_COLOR_BIT) {
+                const LogObjectList objlist(device, image_state->Handle());
+                skip |= LogError(
+                    objlist, "VUID-VkHostImageLayoutTransitionInfoEXT-image-01671",
+                    "%s: image in transition %d requires subresourceRange.aspectMask to be VK_IMAGE_ASPECT_COLOR_BIT, but it is %s",
+                    func_name, i, string_VkImageAspectFlags(aspect_mask).c_str());
+            }
+        }
+        if ((FormatIsMultiplane(image_format)) && (image_state->disjoint == true)) {
+            if (!IsValidPlaneAspect(image_format, aspect_mask) && ((aspect_mask & VK_IMAGE_ASPECT_COLOR_BIT) == 0)) {
```
artem-lunarg, line 2345: Does `IsValidPlaneAspect` work correctly here? The VUIDs asks _"include at least one"_ multi-planar aspect. If we have 2-plane 'format' (so it has plan0+plane1 aspects) and aspect mask with PLANE_0 + PLANE_2, then the above condition is true because `PLANE_0` is included, but it looks `IsValidPlaneAspect` will result `false` due to PLANE_2, which will trigger error.

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -271,147 +271,258 @@ bool CoreChecks::ValidateImageArrayLayerRange(const CMD_BUFFER_STATE &cb_state,
...
+static const char *GetBufferMemoryImageCopyCommandVUID(const std::string &id, bool from_image, bool copy2, bool is_memory = false) {
     // clang-format off
-    static const std::map<std::string, std::array<const char *, 4>> copy_imagebuffer_vuid = {
+    static const std::map<std::string, std::array<const char *, 6>> copy_imagebuffermemory_vuid = {
+        {"07974", {
+            "VUID-vkCmdCopyBufferToImage-dstImage-07974",    // !copy2 & !from_image
+            "VUID-vkCmdCopyImageToBuffer-srcImage-07974",    // !copy2 &  from_image
+            "VUID-VkCopyBufferToImageInfo2-dstImage-07974",  //  copy2 & !from_image
+            "VUID-VkCopyImageToBufferInfo2-srcImage-07974",  //  copy2 &  from_image
+            "DOES-NOT-APPLY",                                //  memory & !from_image
```
spencer-lunarg, line 283: like just use `kVUIDUndefined`

tests/unit/host_image_copy.cpp
```diff
@@ -0,0 +1,1482 @@
...
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+#include "../framework/layer_validation_tests.h"
+#include "utils/vk_layer_utils.h"
+#include "generated/enum_flag_bits.h"
+
+bool copy_layout_supported(std::vector<VkImageLayout> &copy_src_layouts, std::vector<VkImageLayout> &copy_dst_layouts,
```
spencer-lunarg, line 18: nit: `CopyLayoutSupported`

also can you move to `NegativeHostImageCopy` for a future `host_image_copy_positive.cpp` to use

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -2241,6 +2681,441 @@ void CoreChecks::PreCallRecordCmdCopyBufferToImage2(VkCommandBuffer commandBuffe
...
+        const char *resource_name = is_src ? "srcSubresouce" : "dstSubresource";
+        LogObjectList objlist(device, image_state.image());
+        skip |= LogError(objlist, vuid,
+                         "vkCopyMemoryToImageEXT(): "
+                         "%s has a format with 2 planes (%s) and pRegion[%d].%s.aspectMask (%s) must be "
+                         "VK_IMAGE_ASPECT_PLANE_0_BIT or VK_IMAGE_ASPECT_PLANE_1_BIT",
+                         image_name, string_VkFormat(image_state.createInfo.format), i, resource_name,
+                         string_VkImageAspectFlags(aspect_mask).c_str());
+    }
+    if (FormatPlaneCount(image_state.createInfo.format) == 3 &&
```
spencer-lunarg, line 3020: (noting) so we have a `IsOnlyOneValidPlaneAspect` and I rewrote the VUs in the spec so we didn't need to aweful extra logic just to check the same thing but with the plane size being `2` or `3`

for now this is good, but going to make a spec PR to change these VUs to just be a single thing

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -1263,77 +1626,262 @@ bool CoreChecks::ValidateCmdCopyImage(VkCommandBuffer commandBuffer, VkImage src
...
+        skip |= ValidateImageMipLevel(handle, dst_image_state, region.dstSubresource.mipLevel, i, func_name, "dstSubresource",
+                                      GetImageCopyVUID("07967dst", is_2, is_host));
+        skip |= ValidateImageArrayLayerRange(handle, src_image_state, region.srcSubresource.baseArrayLayer,
+                                             region.srcSubresource.layerCount, i, func_name, "srcSubresource",
+                                             GetImageCopyVUID("08790src", is_2, is_host));
+        skip |= ValidateImageArrayLayerRange(handle, dst_image_state, region.dstSubresource.baseArrayLayer,
+                                             region.dstSubresource.layerCount, i, func_name, "dstSubresource",
+                                             GetImageCopyVUID("08790dst", is_2, is_host));
+
+        if (api_version < VK_API_VERSION_1_1) {
```
spencer-lunarg, line 1643: we shouldn't need this, the `!IsExtEnabled` will handle this already

layers/core_checks/cc_image_layout.cpp
```diff
@@ -1114,3 +1114,86 @@ template void CoreChecks::TransitionImageLayouts(CMD_BUFFER_STATE *cb_state, uin
                                                  const VkImageMemoryBarrier *barrier);
 template void CoreChecks::TransitionImageLayouts(CMD_BUFFER_STATE *cb_state, uint32_t barrier_count,
                                                  const VkImageMemoryBarrier2KHR *barrier);
+
+bool CoreChecks::IsCompliantSubresourceRange(const VkImageSubresourceRange &subres_range, const IMAGE_STATE &image_state) const {
+    if (subres_range.baseMipLevel >= image_state.createInfo.mipLevels) return false;
+    if (subres_range.baseMipLevel + subres_range.levelCount > image_state.createInfo.mipLevels) return false;
+    if (subres_range.baseArrayLayer >= image_state.createInfo.arrayLayers ||
+        subres_range.layerCount > image_state.createInfo.arrayLayers ||
```
artem-lunarg, line 1122: nit:  `subres_range.layerCount > image_state.createInfo.arrayLayers` check is not needed.
It will be covered by `(subres_range.baseArrayLayer + subres_range.layerCount) > image_state.createInfo.arrayLayers`.
That's similar to the above check with mip levels where we have 2 checks (instead of 3 here).

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -271,147 +271,258 @@ bool CoreChecks::ValidateImageArrayLayerRange(const CMD_BUFFER_STATE &cb_state,
...
 }
 
-// All VUID from copy_bufferimage_to_imagebuffer_common.txt
-static const char *GetBufferImageCopyCommandVUID(const std::string &id, bool image_to_buffer, bool copy2) {
+// All VUID from copy_bufferimage_to_imagebuffer_common.txt with more as a result of host_image_copy
+static const char *GetBufferMemoryImageCopyCommandVUID(const std::string &id, bool from_image, bool copy2, bool is_memory = false) {
     // clang-format off
-    static const std::map<std::string, std::array<const char *, 4>> copy_imagebuffer_vuid = {
+    static const std::map<std::string, std::array<const char *, 6>> copy_imagebuffermemory_vuid = {
+        {"07974", {
```
artem-lunarg, line 278: can't find 07974 in the 261 spec, not sure if it's an old vuid or a future one.

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -271,147 +271,258 @@ bool CoreChecks::ValidateImageArrayLayerRange(const CMD_BUFFER_STATE &cb_state,
...
+        index = 4 + (from_image ? 1 : 0);
+    } else {
+        index |= uint8_t((from_image) ? 0x1 : 0);
+        index |= uint8_t((copy2) ? 0x2 : 0);
+    }
+    return copy_imagebuffermemory_vuid.at(id).at(index);
 }
 
-static bool VerifyAspectsPresent(VkImageAspectFlags aspect_mask, VkFormat format) {
+bool CoreChecks::VerifyAspectsPresent(VkImageAspectFlags aspect_mask, VkFormat format) const {
```
artem-lunarg, line 525: nit: it looks this function does not depend on `this`. If the reason to add to it `CoreChecks` interface was to make it visible in multiple files, then the alternative that does not add new members to `CoreChecks` is just to remove `static` from `VerifyAspectsPresent` and declare its signature. In the future we can put such helpers into one of the utility/helper files.

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -666,39 +773,107 @@ bool CoreChecks::ValidateBufferImageCopyData(const CMD_BUFFER_STATE &cb_state, u
...
                              function, i, region.imageExtent.depth, block_size.depth, region.imageOffset.z, mip_extent.depth);
         }
 
+        // *RowLength divided by the texel block extent width and then multiplied by the texel block size of the image must be
+        // less than or equal to 2^31-1
+        const uint32_t element_size =
+            FormatIsDepthOrStencil(image_format) ? 0 : FormatElementSize(image_format, region_aspect_mask);
+        double test_value = row_length / block_size.width;
+        test_value = test_value * element_size;
+        const auto two_to_31_minus_1 = pow(2, 31) - 1;
```
artem-lunarg, line 789: nit: double can represent all `int`s or `uint`s precisely, so we could also use somehting like `static_cast<double>(1u << 31 - 1)`. I'm not sure if `pow` gives any guarantees that the result will be as good (i.e. exact value without any fraction).

artem-lunarg: One observation that might help to simplify. Because we perform `09106` earlier we can go here without floating point calculations and assume `bufferRowLength must be a multiple of the texel block extent width`. For this we need skip this validation if `09106` failed. It should be fine because if `09106` happens then this validation is not useful (if it passes it just a luck, and when it fails it's not representable until `09106` is fixed).

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -666,39 +773,107 @@ bool CoreChecks::ValidateBufferImageCopyData(const CMD_BUFFER_STATE &cb_state, u
...
+    const bool is_2 = (cmd_type == CMD_COPYBUFFERTOIMAGE2KHR || cmd_type == CMD_COPYBUFFERTOIMAGE2);
+    const char *vuid;
+
+    const VkFormat image_format = image_state.createInfo.format;
+
+    skip |= ValidateBufferOrMemoryToOrFromImageCopyData(cb_state.commandBuffer(), regionCount, pRegions, image_state, function,
+                                                        cmd_type, image_to_buffer, false);
+
+    for (uint32_t i = 0; i < regionCount; i++) {
+        const RegionType region = pRegions[i];
```
artem-lunarg, line 836: nit: one option is to use reference (`const RegionType&`) (unless it was intention to go with values).

layers/core_checks/cc_device.cpp
Tony-LunarG: Done

layers/core_checks/cc_device.cpp
Tony-LunarG: This code was removed because it was checking an invalid VUID

layers/core_checks/cc_image.cpp
Tony-LunarG: All marked const auto

tests/framework/layer_validation_tests.cpp
Tony-LunarG: Enums used in place of raw constants

tests/framework/layer_validation_tests.cpp
Tony-LunarG: or s removed

layers/core_checks/cc_device.cpp
Tony-LunarG: Changed

layers/core_checks/cc_copy_blit_resolve.cpp
Tony-LunarG: I started with typed handles, in fact there's a commit where I switched from typed handles to templates, because that seemed more prevalent as a precedent

layers/core_checks/cc_image.cpp
Tony-LunarG: Fixed

layers/core_checks/cc_image.cpp
Tony-LunarG: Reordered

layers/core_checks/cc_image.cpp
Tony-LunarG: Added and used HasAtLeastOneMultiplane

tests/unit/host_image_copy.cpp
Tony-LunarG: Done

layers/core_checks/cc_image_layout.cpp
```diff
@@ -1114,3 +1114,86 @@ template void CoreChecks::TransitionImageLayouts(CMD_BUFFER_STATE *cb_state, uin
...
+        subres_range.aspectMask & VK_IMAGE_ASPECT_MEMORY_PLANE_0_BIT_EXT ||
+        subres_range.aspectMask & VK_IMAGE_ASPECT_MEMORY_PLANE_1_BIT_EXT ||
+        subres_range.aspectMask & VK_IMAGE_ASPECT_MEMORY_PLANE_2_BIT_EXT ||
+        subres_range.aspectMask & VK_IMAGE_ASPECT_MEMORY_PLANE_3_BIT_EXT) {
+        return false;
+    }
+    return true;
+}
+
+bool CoreChecks::ValidateHostCopyCurrentLayout(VkDevice device, const VkImageLayout expected_layout,
```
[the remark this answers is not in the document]
Tony-LunarG: That would cause most of the callers to have to call RangeFromLayers rather than one call from here.  Is that really better?

layers/core_checks/cc_image_layout.cpp
Tony-LunarG: Reworked

layers/core_checks/cc_image_layout.cpp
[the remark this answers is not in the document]
artem-lunarg: I'm not sure what is the direction of asymmetry (I thought that RangeFromLayers is called in fewer cases, but maybe overlooked, still not that important).  The idea is that top level entities (functions, types) have a tendency to add more to the complexity of software comparing to similar construction inside implementation (e.g. function call). I noticed that sometimes we add new methods for relatively small conveniences and interfaces grow in complexity (instead of adding a one more step to implementation details) which sometimes makes it harder to reason about the system.

But as I mentioned that might be subjective and that's not a request for change. I consider my comments just as comments (not a request for change), except where it's a bug or mentioned explicitly, and it's up to the author who usually has the most context what to with feedback!

layers/core_checks/cc_copy_blit_resolve.cpp
Tony-LunarG: Do you mean that we don't need the checks guarded by the api_version check?  This is code common to both the host image copy ImageToImage as well as CmdCopyImage which does still have 07940 and 07941.

layers/core_checks/cc_copy_blit_resolve.cpp
artem-lunarg: Okay, maybe that's something we can discuss with a team.

Tony-LunarG: I can do either, but maybe in a future PR.

layers/core_checks/cc_image.cpp
artem-lunarg: Hm... I read the VUID requirement a bit differently. The spec has definition of [multi-planar aspect mask](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#formats-planes-image-aspect) term .

Original version that called `IsValidPlaneAspect`) checked that **all** aspect bits satisfy condition of multi-planar aspect mask. 
My understanding we need something similar to `IsValidPlaneAspect` but that returns true if at least **one** aspect bit satisfies condition. This can be emulated by running `IsValidPlaneAspect` in a loop for each bit of `aspect_mask` and `or`-ing results. Maybe that helpers should be polished even more that will allow to distinguish such differences more easily. 

I remember @spencer-lunarg worked with multi plane aspects. Feel free to correct me if I misunderstood something.

artem-lunarg: Okay, I found `VUID-VkImageMemoryBarrier2-image-01672` and `VUID-VkImageMemoryBarrier-image-01672` that have similar formulations and also use `IsValidPlaneAspect`. If that's correct implementation then sorry for the confusion.

I still don't understand how `IsValidPlaneAspect` is enough. One explanation that it relies on some earlier check that ensures that aspect mask is valid for the given multiplane format. But if that's the case it still might be a good idea not to rely on that check in order no to have false positives due to the previous errors.

In case it's a real issue, I'm totally fine if it will be fixed outside this PR since this solution is not used in other places too.

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -666,39 +773,107 @@ bool CoreChecks::ValidateBufferImageCopyData(const CMD_BUFFER_STATE &cb_state, u
...
+            FormatIsDepthOrStencil(image_format) ? 0 : FormatElementSize(image_format, region_aspect_mask);
+        double test_value = row_length / block_size.width;
+        test_value = test_value * element_size;
+        const auto two_to_31_minus_1 = pow(2, 31) - 1;
+        if (test_value > two_to_31_minus_1) {
+            const LogObjectList objlist(handle, image_state.image());
+            skip |= LogError(objlist, GetBufferMemoryImageCopyCommandVUID("09108", from_image, is_2, is_memory),
+                             "%s: pRegion[%d] *RowLength (%" PRIu32
+                             ") divided by the texel block extent width (%d) then multiplied by the "
+                             "texel block size of image (%d) is (%" PRIu64 ") which is greater than 2^31 - 1",
```
arno-lunarg, line 795:
```suggestion
                             ") divided by the texel block extent width (%" PRIu32 ") then multiplied by the "
                             "texel block size of image (%" PRIu32 ") is (%" PRIu64 ") which is greater than 2^31 - 1",
```

layers/core_checks/cc_image.cpp
artem-lunarg: @Tony-LunarG I think I misunderstood the vuid, sorry for this. I was confused by the word `one` in `include at least one` (still don't understand if it is used there correctly, could be my english). But it looks the VUID does not refer to a single aspect bit but to the entire mask (which always a **single** mask, that's why that `one` forced me to think in the wrong direction). 

So, it looks the original implementation of `IsValidPlaneAspect` was correct (and that's how it is also implemented in other similar VUIDs like `VUID-VkImageMemoryBarrier2-image-01672`). Again, sorry for the waste of time.

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -2241,6 +2681,441 @@ void CoreChecks::PreCallRecordCmdCopyBufferToImage2(VkCommandBuffer commandBuffe
...
+bool CoreChecks::ValidateMemcpyExtents(VkDevice device, const VkImageCopy2 region, uint32_t i, const IMAGE_STATE &image_state, bool is_src) const{
+    bool skip = false;
+    if (region.srcOffset.x != 0 || region.srcOffset.y != 0 || region.srcOffset.z != 0) {
+        const char *vuid =
+            is_src ? "VUID-VkCopyImageToImageInfoEXT-srcOffset-09114" : "VUID-VkCopyImageToImageInfoEXT-dstOffset-09114";
+        const char *type_name = is_src ? "srcOffset" : "dstOffset";
+        LogObjectList objlist(device);
+        skip |= LogError(objlist, vuid,
+                 "vkCopyMemoryToImageEXT(): VkCopyImageToImageInfoEXT.flags contains VK_HOST_IMAGE_COPY_MEMCPY_EXT which "
+                 "means that pRegions[%d].%s x(%u), y(%u) and z(%u) must all be zero",
```
arno-lunarg, line 2986:
```suggestion
                 "means that pRegions[%" PRIu32 "].%s x(%" PRId32 "), y(%" PRId32 ") and z(%" PRId32 ") must all be zero",
```

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -2241,6 +2681,441 @@ void CoreChecks::PreCallRecordCmdCopyBufferToImage2(VkCommandBuffer commandBuffe
...
+                 i, type_name, region.srcOffset.x, region.srcOffset.y, region.srcOffset.z);
+    }
+    if (!IsExtentEqual(region.extent, image_state.createInfo.extent)) {
+        const char *vuid = is_src ? "VUID-VkCopyImageToImageInfoEXT-extent-09115" : "VUID-VkCopyImageToImageInfoEXT-extent-09115";
+        LogObjectList objlist(device, image_state.image());
+        skip |= LogError(objlist, vuid,
+                            "vkCopyMemoryToImageEXT(): pRegion[%d].imageExtent (w=%d, h=%d, d=%d) must match the image's subresource "
+                            "extents (w=%d, h=%d, d=%d) when VkCopyImageToImageInfoEXT->flags contains VK_HOST_IMAGE_COPY_MEMCPY_EXT",
```
arno-lunarg, line 2994:
```suggestion
                            "vkCopyMemoryToImageEXT(): pRegion[%" PRIu32 "].imageExtent (w=%" PRIu32 ", h=%" PRIu32 ", d=%" PRIu32 ") must match the image's subresource "
                            "extents (w=%" PRIu32 ", h=%" PRIu32 ", d=%" PRIu32 ") when VkCopyImageToImageInfoEXT->flags contains VK_HOST_IMAGE_COPY_MEMCPY_EXT",
```

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -2241,6 +2681,441 @@ void CoreChecks::PreCallRecordCmdCopyBufferToImage2(VkCommandBuffer commandBuffe
...
+    if (FormatPlaneCount(image_state.createInfo.format) == 2 &&
+        (aspect_mask != VK_IMAGE_ASPECT_PLANE_0_BIT && aspect_mask != VK_IMAGE_ASPECT_PLANE_1_BIT)) {
+        const char *vuid =
+            is_src ? "VUID-VkCopyImageToImageInfoEXT-srcImage-07981" : "VUID-VkCopyImageToImageInfoEXT-dstImage-07981";
+        const char *image_name = is_src ? "srcImage" : "dstImage";
+        const char *resource_name = is_src ? "srcSubresouce" : "dstSubresource";
+        LogObjectList objlist(device, image_state.image());
+        skip |= LogError(objlist, vuid,
+                         "vkCopyMemoryToImageEXT(): "
+                         "%s has a format with 2 planes (%s) and pRegion[%d].%s.aspectMask (%s) must be "
```
arno-lunarg, line 3015:
```suggestion
                         "%s has a format with 2 planes (%s) and pRegion[%" PRIu32 "].%s.aspectMask (%s) must be "
```

arno-lunarg: I make this suggestion here (could be done at other similar spots) because `i` is a `uint32_t`, but maybe we can safely assume that it will always be a small value, hence using `%d` could be ok?

tests/framework/layer_validation_tests.h
Tony-LunarG: I removed the default parameter

layers/core_checks/cc_image.cpp
Tony-LunarG: I think you're right - I missed the one _mask_ and read one _plane_ - I'll go back to using IsValidPlaneAspect

layers/core_checks/cc_copy_blit_resolve.cpp
Tony-LunarG: Done

layers/core_checks/cc_copy_blit_resolve.cpp
Tony-LunarG: Fixed

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -427,23 +538,45 @@ static bool VerifyAspectsPresent(VkImageAspectFlags aspect_mask, VkFormat format
...
+template <>
+uint32_t GetImageHeight<VkMemoryToImageCopyEXT>(VkMemoryToImageCopyEXT data) {
+    return data.memoryImageHeight;
+}
+template <>
+uint32_t GetImageHeight<VkImageToMemoryCopyEXT>(VkImageToMemoryCopyEXT data) {
+    return data.memoryImageHeight;
+}
+template <typename HandleT, typename RegionType>
+bool CoreChecks::ValidateBufferOrMemoryToOrFromImageCopyData(const HandleT handle, uint32_t regionCount, const RegionType *pRegions,
```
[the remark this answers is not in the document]
Tony-LunarG: Changed to ValidateHeterogeneousCopyData.  Heterogeneous to signify that the copy is between two different types of things

layers/state_tracker/state_tracker.h
```diff
@@ -1479,10 +1479,10 @@ class ValidationStateTracker : public ValidationObject {
                                         const VkShaderEXT* pShaders) override;
 
     template <typename ExtProp>
-    void GetPhysicalDeviceExtProperties(VkPhysicalDevice gpu, ExtEnabled enabled, ExtProp* ext_prop) {
+    void GetPhysicalDeviceExtProperties(VkPhysicalDevice gpu, ExtEnabled enabled, ExtProp* ext_prop, bool init = true) {
```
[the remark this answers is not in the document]
juan-lunarg: Here is a patch to show what I mean

```
diff --git a/layers/state_tracker/state_tracker.cpp b/layers/state_tracker/state_tracker.cpp
index 08e5dbb82..2d821ca6b 100644
--- a/layers/state_tracker/state_tracker.cpp
+++ b/layers/state_tracker/state_tracker.cpp
@@ -1639,8 +1639,8 @@ void ValidationStateTracker::CreateDevice(const VkDeviceCreateInfo *pCreateInfo)
         phys_dev_props->host_image_copy_properties.pCopySrcLayouts = host_image_copy_src_layouts.data();
         phys_dev_props->host_image_copy_properties.pCopyDstLayouts = host_image_copy_dst_layouts.data();
         // Call again (without init) to fill in lists
-        GetPhysicalDeviceExtProperties(physical_device, dev_ext.vk_ext_host_image_copy,
-                                       &phys_dev_props->host_image_copy_properties, false);
+        GetPhysicalDeviceExtProperties<false>(physical_device, dev_ext.vk_ext_host_image_copy,
+                                              &phys_dev_props->host_image_copy_properties);
     }
     if (api_version >= VK_API_VERSION_1_1) {
         GetPhysicalDeviceExtProperties(physical_device, &phys_dev_props->subgroup_props);
diff --git a/layers/state_tracker/state_tracker.h b/layers/state_tracker/state_tracker.h
index 953401eb6..05a246a00 100644
--- a/layers/state_tracker/state_tracker.h
+++ b/layers/state_tracker/state_tracker.h
@@ -1478,11 +1478,13 @@ class ValidationStateTracker : public ValidationObject {
     void PreCallRecordCmdBindShadersEXT(VkCommandBuffer commandBuffer, uint32_t stageCount, const VkShaderStageFlagBits* pStages,
                                         const VkShaderEXT* pShaders) override;
 
-    template <typename ExtProp>
-    void GetPhysicalDeviceExtProperties(VkPhysicalDevice gpu, ExtEnabled enabled, ExtProp* ext_prop, bool init = true) {
+    template <bool init = true, typename ExtProp>
+    void GetPhysicalDeviceExtProperties(VkPhysicalDevice gpu, ExtEnabled enabled, ExtProp* ext_prop) {
         assert(ext_prop);
         if (IsExtEnabled(enabled)) {
-            if (init) *ext_prop = LvlInitStruct<ExtProp>();
+            if constexpr (init) {
+                *ext_prop = LvlInitStruct<ExtProp>();
+            }
             if (api_version < VK_API_VERSION_1_1) {
                 auto prop2 = LvlInitStruct<VkPhysicalDeviceProperties2>(ext_prop);
                 DispatchGetPhysicalDeviceProperties2KHR(gpu, &prop2);
```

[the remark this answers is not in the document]
juan-lunarg: I tested this patch locally and it compiled without issue.

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -230,28 +230,28 @@ bool CoreChecks::CheckItgExtent(const LogObjectList &objlist, const VkExtent3D &
     }
     return skip;
 }
-
-bool CoreChecks::ValidateImageMipLevel(const CMD_BUFFER_STATE &cb_state, const IMAGE_STATE &img, uint32_t mip_level,
+template <typename HandleT>
```
spencer-lunarg, line 233: (sorry if this was discussed before)


but can we just pass in a `const LogObjectList& objlist` here?

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -914,10 +1124,195 @@ bool CoreChecks::ValidateCopyImageTransferGranularityRequirements(const CMD_BUFF
...
+            "VUID-vkCmdCopyImage-dstImage-01733",
+            "VUID-VkCopyImageInfo2-dstImage-01733",
+            "VUID-VkCopyImageToImageInfoEXT-imageExtent-00208",
+       }},
+       {"01734", {
+            "VUID-vkCmdCopyImage-dstImage-01734",
+            "VUID-VkCopyImageInfo2-dstImage-01734",
+            "VUID-VkCopyImageToImageInfoEXT-imageExtent-00209",
+       }},
+       {"07967src", {
```
spencer-lunarg, line 1220: for now this is fine... I nice future enhancement will just to use the new `Location` and do it like we do for `sync_vuid_map` in future

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -1655,29 +2070,24 @@ bool CoreChecks::ValidateCmdCopyImage(VkCommandBuffer commandBuffer, VkImage src
         }
     }
 
+    // Source and dest image sample counts must match
+    if (src_image_state->createInfo.samples != dst_image_state->createInfo.samples) {
+        const LogObjectList objlist(commandBuffer, srcImage, dstImage);
+        std::stringstream ss;
```
spencer-lunarg, line 2076: `ss` is not used in the message

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -1826,18 +2236,47 @@ void CoreChecks::PreCallRecordCmdCopyBuffer2(VkCommandBuffer commandBuffer, cons
...
-template <typename RegionType>
-bool CoreChecks::ValidateImageBounds(VkCommandBuffer cb, const IMAGE_STATE &image_state, const uint32_t regionCount,
-                                     const RegionType *pRegions, const char *func_name, const char *msg_code) const {
+template <typename T>
+VkImageSubresourceLayers GetImageSubresource(T data, bool is_src) {
+    return data.imageSubresource;
+}
+template <>
+VkImageSubresourceLayers GetImageSubresource<VkImageCopy2>(VkImageCopy2 data, bool is_src) {
+    if (is_src)
```
spencer-lunarg, line 2245: nit: just do a ternary? `return is_src ? data.srcSubresource : data.dstSubresource;`  (same with `GetOffset`)

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -2241,6 +2683,428 @@ void CoreChecks::PreCallRecordCmdCopyBufferToImage2(VkCommandBuffer commandBuffe
...
+        mismatch_stream << "srcImage sharingMode = " << string_VkSharingMode(src_info.sharingMode)
+                        << "and dstImage sharingMode = " << string_VkSharingMode(dst_info.sharingMode) << "\n";
+    }
+    if (src_info.initialLayout != dst_info.initialLayout) {
+        mismatch_stream << "srcImage initialLayout = " << string_VkImageLayout(src_info.initialLayout)
+                        << "and dstImage initialLayout = " << string_VkImageLayout(dst_info.initialLayout) << "\n";
+    }
+
+    if (mismatch_stream.str().length() > 0) {
+        std::stringstream ss;
```
spencer-lunarg, line 2912: can just do this inside the `LogError`, no need for this `stringstream` here IMO

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -2241,6 +2683,428 @@ void CoreChecks::PreCallRecordCmdCopyBufferToImage2(VkCommandBuffer commandBuffe
...
+                                         const VkImageLayout *supported_image_layouts, const VkImageLayout image_layout,
+                                         const char *func_name, const char *field_name, const char *supported_name,
+                                         const char *vuid) const {
+    bool skip = false;
+
+    bool found = false;
+    for (uint32_t i = 0; i < layout_count; ++i) {
+        if (supported_image_layouts[i] == image_layout) {
+            found = true;
+            break;
```
spencer-lunarg, line 2930: if this function is self-contained, could just `return` here and then no need for a `found` bool at all

tests/unit/host_image_copy.cpp
```diff
@@ -0,0 +1,1563 @@
...
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+#include "../framework/layer_validation_tests.h"
+#include "utils/vk_layer_utils.h"
+#include "generated/enum_flag_bits.h"
+
+bool HostImageCopyTest::CopyLayoutSupported(const std::vector<VkImageLayout> &copy_src_layouts,
```
spencer-lunarg, line 18: nit: we have been putting the helper in the `_positive.cpp` file

layers/core_checks/cc_copy_blit_resolve.cpp
artem-lunarg: saw this few times in other places, it might be a style to prefer single return (or maybe not). Personally I prefer approach without local variable.

layers/core_checks/cc_copy_blit_resolve.cpp
Tony-LunarG: Used the static_cast

layers/core_checks/cc_copy_blit_resolve.cpp
Tony-LunarG: Done

layers/core_checks/cc_copy_blit_resolve.cpp
Tony-LunarG: We could, but then every place it is called would have to build a LogObjectList to pass to it.  The template seems cleaner.

layers/core_checks/cc_copy_blit_resolve.cpp
Tony-LunarG: I'm using the stringstream to accumulate what is mismatched.  Without the stringstream how would I know what to pass to LogError in terms of what is mismatched? Or one LogError per mismatched CreateInfo field?

layers/utils/vk_layer_utils.h
```diff
@@ -415,6 +415,11 @@ static inline bool IsMultiplePlaneAspect(VkImageAspectFlags aspect_mask) {
     return planes != 0 && !IsPowerOfTwo(planes);
 }
 
+static inline bool HasAtLeastOneMultiplane(VkImageAspectFlags aspect_mask) {
+    return ((aspect_mask & VK_IMAGE_ASPECT_PLANE_0_BIT) || (aspect_mask & VK_IMAGE_ASPECT_PLANE_1_BIT) ||
+            (aspect_mask & VK_IMAGE_ASPECT_PLANE_2_BIT));
+}
```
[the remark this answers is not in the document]
Tony-LunarG: Removed

layers/core_checks/cc_copy_blit_resolve.cpp
Tony-LunarG: Fixed

layers/core_checks/cc_copy_blit_resolve.cpp
Tony-LunarG: Fixed

layers/core_checks/cc_copy_blit_resolve.cpp
Tony-LunarG: Fixed

layers/core_checks/cc_copy_blit_resolve.cpp
Tony-LunarG: Fixed.  I think it's worth doing for consistency

layers/core_checks/cc_copy_blit_resolve.cpp
Tony-LunarG: Removed

layers/core_checks/cc_copy_blit_resolve.cpp
Tony-LunarG: Done

layers/core_checks/cc_copy_blit_resolve.cpp
Tony-LunarG: Done

tests/unit/host_image_copy.cpp
Tony-LunarG: Moved

layers/core_checks/cc_copy_blit_resolve.cpp
spencer-lunarg: I guess my goal with the new "improving error message" campaign was how we passed around handles, but I guess for this case right now this is as good as we can do

layers/state_tracker/state_tracker.h
```diff
@@ -1478,11 +1478,13 @@ class ValidationStateTracker : public ValidationObject {
     void PreCallRecordCmdBindShadersEXT(VkCommandBuffer commandBuffer, uint32_t stageCount, const VkShaderStageFlagBits* pStages,
                                         const VkShaderEXT* pShaders) override;
 
-    template <typename ExtProp>
+    template <bool init = true, typename ExtProp>
     void GetPhysicalDeviceExtProperties(VkPhysicalDevice gpu, ExtEnabled enabled, ExtProp* ext_prop) {
         assert(ext_prop);
         if (IsExtEnabled(enabled)) {
-            *ext_prop = LvlInitStruct<ExtProp>();
+            if constexpr (init) {
```
spencer-lunarg, line 1485: maybe a comment about how this is for extensions that require having to query its property struct twice to get values

layers/core_checks/cc_device.cpp
```diff
@@ -601,11 +601,27 @@ bool CoreChecks::PreCallValidateGetDeviceQueue2(VkDevice device, const VkDeviceQ
...
 
+bool CoreChecks::ValidateGetPhysicalDeviceImageFormatProperties2(const VkPhysicalDeviceImageFormatInfo2 *pImageFormatInfo,
+                                                                 VkImageFormatProperties2 *pImageFormatProperties) const {
+    bool skip = false;
+    const auto *copy_perf_query = LvlFindInChain<VkHostImageCopyDevicePerformanceQueryEXT>(pImageFormatProperties->pNext);
+    if (copy_perf_query) {
+        if ((pImageFormatInfo->usage & VK_IMAGE_USAGE_HOST_TRANSFER_BIT_EXT) == 0) {
+            skip |= LogError(physical_device, "VUID-vkGetPhysicalDeviceImageFormatProperties2-pNext-09004",
+                             "vkGetPhysicalDeviceImageFormatProperties2(): pImageFormatProperties includes a chained "
+                             "VkHostImageCopyDevicePerformanceQueryEXT struct, but pImageFormatInfo->usage does not contain "
```
spencer-lunarg, line 612: print `pImageFormatInfo->usage` values in error message (with `string_` helper)

layers/core_checks/cc_image.cpp
```diff
@@ -2323,6 +2299,139 @@ bool CoreChecks::PreCallValidateGetImageSubresourceLayout2EXT(VkDevice device, V
...
+        const auto &transition = pTransitions[i];
+        const auto image_state = Get<IMAGE_STATE>(transition.image);
+        const auto image_format = image_state->createInfo.format;
+        const auto aspect_mask = transition.subresourceRange.aspectMask;
+        const bool has_depth_mask = (aspect_mask & VK_IMAGE_ASPECT_DEPTH_BIT) != 0;
+        const bool has_stencil_mask = (aspect_mask & VK_IMAGE_ASPECT_STENCIL_BIT) != 0;
+
+        if ((image_state->createInfo.usage & VK_IMAGE_USAGE_HOST_TRANSFER_BIT_EXT) == 0) {
+            const LogObjectList objlist(device, image_state->Handle());
+            skip |= LogError(objlist, "VUID-VkHostImageLayoutTransitionInfoEXT-image-09055",
```
spencer-lunarg, line 2324: print `image_state->createInfo.usage ` in error

tests/unit/host_image_copy_positive.cpp
```diff
@@ -0,0 +1,220 @@
...
+        !CopyLayoutSupported(copy_src_layouts, copy_dst_layouts, VK_IMAGE_LAYOUT_GENERAL)) {
+        GTEST_SKIP() << "Required formats/features not supported";
+    }
+}
+
+
+TEST_F(PositiveHostImageCopy, BasicUsage) {
+    TEST_DESCRIPTION("Use VK_EXT_host_image_copy to copy to and from host memory");
+
+    VkFormat compressed_format = VK_FORMAT_UNDEFINED;
```
spencer-lunarg, line 79: instead of passing these (`compressed_format`, `copy_src_layouts`) as reference in the init, lets just make them member variables of `HostImageCopyTest`

spencer-lunarg (approved):
few nits left, but otherwise happy with this

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -2241,6 +2677,426 @@ void CoreChecks::PreCallRecordCmdCopyBufferToImage2(VkCommandBuffer commandBuffe
...
+
+bool CoreChecks::ValidateHostCopyImageLayout(const VkDevice device, const VkImage image, const uint32_t layout_count,
+                                         const VkImageLayout *supported_image_layouts, const VkImageLayout image_layout,
+                                         const char *func_name, const char *field_name, const char *supported_name,
+                                         const char *vuid) const {
+    bool skip = false;
+
+    for (uint32_t i = 0; i < layout_count; ++i) {
+        if (supported_image_layouts[i] == image_layout) {
+            return false;
```
artem-lunarg, line 2922: nit: we can return here skip or skip can be initialized later

layers/state_tracker/state_tracker.cpp
```diff
@@ -1625,6 +1630,18 @@ void ValidationStateTracker::CreateDevice(const VkDeviceCreateInfo *pCreateInfo)
     GetPhysicalDeviceExtProperties(physical_device, dev_ext.vk_qcom_image_processing, &phys_dev_props->image_processing_props);
     GetPhysicalDeviceExtProperties(physical_device, dev_ext.vk_ext_descriptor_buffer, &phys_dev_props->descriptor_buffer_props);
     GetPhysicalDeviceExtProperties(physical_device, dev_ext.vk_ext_descriptor_buffer, &phys_dev_props->descriptor_buffer_density_props);
+    GetPhysicalDeviceExtProperties(physical_device, dev_ext.vk_ext_host_image_copy, &phys_dev_props->host_image_copy_properties);
+    if ((phys_dev_props->host_image_copy_properties.copySrcLayoutCount > 0) ||
```
artem-lunarg, line 1634: nit: if only one of `copySrcLayoutCount ` or `copyDstLayoutCount ` is 0 then the following code will initialize corresponding `pCopySrcLayouts` or `pCopyDstLayouts` with non-zero pointer. According to c++ spec everything is valid (you can call data() on empty vector). But we will store a non-null pointer in the structure that we are not allowed to dereference and that is not used. It would be cleaner and safer to keep it then as nil. Could be for a separate maintenance PR.

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -2241,6 +2677,426 @@ void CoreChecks::PreCallRecordCmdCopyBufferToImage2(VkCommandBuffer commandBuffe
...
+                                                 : "VUID-VkCopyMemoryToImageInfoEXT-imageSubresource-07967");
+        skip |= ValidateImageArrayLayerRange(device, *image_state, region.imageSubresource.baseArrayLayer,
+                                             region.imageSubresource.layerCount, i, func_name, "imageSubresource",
+                                             from_image ? "VUID-VkCopyImageToMemoryInfoEXT-imageSubresource-08790"
+                                                        : "VUID-VkCopyMemoryToImageInfoEXT-imageSubresource-08790");
+        skip |= ValidateImageSubresourceLayers(device, &region.imageSubresource, func_name, "imageSubresource", i);
+        if (region.imageSubresource.aspectMask & VK_IMAGE_ASPECT_STENCIL_BIT) has_stencil = true;
+        if (region.imageSubresource.aspectMask & ~VK_IMAGE_ASPECT_STENCIL_BIT) has_non_stencil = true;
+
+        if (memcpy) {
```
artem-lunarg, line 2811: nit: Should we avoid using names that match global  function names? If someone decides to rename variable this will compile without changes and the condition will always evaluate to true. This snippet is valid in c++ even without introducing `memcpy` variable:
```
if (memcpy) {
        printf("copyman");
}
```

artem-lunarg (approved):
LGTM. The nits I added are not critical and can be addressed later, if necessary. This PR looks pretty impressive for me, a lot of non-trivial validation and the implementation has good structure to follow.

Although I personally introduced tons of nits, could it be not a bad idea, for such big changes, to get them in as soon as there are no critical issues, and then do an additional maintenance? I imagine it might be painful to keep such large PRs in flight.

layers/state_tracker/state_tracker.h
Tony-LunarG: Done

layers/core_checks/cc_device.cpp
Tony-LunarG: Done

layers/core_checks/cc_image.cpp
Tony-LunarG: Done

tests/unit/host_image_copy_positive.cpp
Tony-LunarG: Great idea!  Done

layers/core_checks/cc_copy_blit_resolve.cpp
Tony-LunarG: Done

layers/state_tracker/state_tracker.cpp
Tony-LunarG: Fixed

layers/core_checks/cc_copy_blit_resolve.cpp
Tony-LunarG: Renamed

Tony-LunarG:
Thanks for all your feedback - Very useful

PR #6285, opened 2023-08-09
layers: Add GPU-AV support for VK_EXT_shader_object

spencer-lunarg:
This adds all the plumbing support so that `VK_EXT_shader_object` can be used in GPU-AV

It uses the new flow from https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6230

The core idea is the idea of `SPIRV` and `ShaderModule`/`ShaderEXT` are completely separated 

( related https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6265 )

PR #6291, opened 2023-08-11
Add VK_EXT_shader_object validation

ziga-lunarg:


layers/error_message/validation_error_enums.h
```diff
@@ -33,6 +33,14 @@
 [[maybe_unused]] static const char *kVUID_Core_DrawState_NoEndCommandBuffer = "UNASSIGNED-CoreValidation-DrawState-NoEndCommandBuffer";
 [[maybe_unused]] static const char *kVUID_Core_DrawState_QueueForwardProgress = "UNASSIGNED-CoreValidation-DrawState-QueueForwardProgress";
 [[maybe_unused]] static const char *kVUID_Core_DrawState_InvalidImageView = "UNASSIGNED-CoreValidation-DrawState-InvalidImageView";
+[[maybe_unused]] static const char *kVUID_Core_DrawState_TessellationDomainOrigin = "UNASSIGNED-CoreValidatoin-DrawState-TessellationDomainOrigin";
```
spencer-lunarg, line 36: I can take the action to get these into the spec

ziga-lunarg: I already have something in progress, there are a lot of other dynamic state VUIDs that dont match what the other part of the spec says

layers/state_tracker/cmd_buffer_state.h
```diff
@@ -209,51 +211,99 @@ class CMD_BUFFER_STATE : public REFCOUNTED_NODE {
...
+        uint32_t exclusive_scissor_enable_count;
+        std::vector<VkBool32> exclusive_scissor_enables;
+        // VK_DYNAMIC_STATE_EXCLUSIVE_SCISSOR_NV
+        uint32_t exclusive_scissor_first;
+        uint32_t exclusive_scissor_count;
+        std::vector<VkRect2D> exclusive_scissors;
         // When the Command Buffer resets, the value most things in this struct don't matter because if they are read without
         // setting the state, it will fail in ValidateDynamicStateIsSet() for us. Some values (ex. the bitset) are tracking in
         // replacement for static_status/dynamic_status so this needs to reset along with those
         void reset() {
```
spencer-lunarg, line 292: you need to clear all the `vector`

color_blend_equations 
color_write_masks
vertex_attribute_descriptions
viewport_w_scalings
exclusive_scissor_enables
exclusive_scissors

layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -225,6 +225,89 @@ bool CoreChecks::ValidateDynamicStateSetStatus(const LAST_BOUND_STATE &last_boun
...
+                        cb_state.commandBuffer(), vuid.color_write_mask_09116,
+                        "%s: Render pass attachment %" PRIu32
+                        " has format VK_FORMAT_E5B9G9R9_UFLOAT_PACK32, but the corresponding element of pColorWriteMasks is %s.",
+                        CommandTypeString(cmd_type), i, string_VkColorComponentFlags(color_write_mask).c_str());
+                }
+            }
+        }
+    }
+
+    std::shared_ptr<const SPIRV_MODULE_STATE> spirv_state;
```
spencer-lunarg, line 256: (thinking out loud) for future, we probably want a helper in `LAST_BOUND_STATE` that can get use either (or both) the `SPIRV_MODULE_STATE` and `EntryPoint` (but that is outside the scope of this PR I think)

layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -504,6 +587,460 @@ bool CoreChecks::ValidateDrawDynamicState(const LAST_BOUND_STATE &last_bound_sta
...
+    graphics_shader_bound |= last_bound_state.IsValidShaderBound(SHADER_OBJECT_GEOMETRY);
+    graphics_shader_bound |= last_bound_state.IsValidShaderBound(SHADER_OBJECT_FRAGMENT);
+    graphics_shader_bound |= last_bound_state.IsValidShaderBound(SHADER_OBJECT_TASK);
+    graphics_shader_bound |= last_bound_state.IsValidShaderBound(SHADER_OBJECT_MESH);
+    bool vertex_shader_bound = last_bound_state.IsValidShaderBound(SHADER_OBJECT_VERTEX);
+    bool tessev_shader_bound = last_bound_state.IsValidShaderBound(SHADER_OBJECT_TESSELLATION_EVALUATION);
+    bool geom_shader_bound = last_bound_state.IsValidShaderBound(SHADER_OBJECT_GEOMETRY);
+    bool fragment_shader_bound = last_bound_state.IsValidShaderBound(SHADER_OBJECT_FRAGMENT);
+
+    const CMD_TYPE mesh_cmd_types[] = {CMD_DRAWMESHTASKSNV,  CMD_DRAWMESHTASKSINDIRECTNV,  CMD_DRAWMESHTASKSINDIRECTCOUNTNV,
```
spencer-lunarg, line 609: not used?

layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -504,6 +587,460 @@ bool CoreChecks::ValidateDrawDynamicState(const LAST_BOUND_STATE &last_bound_sta
...
+                                       CMD_DRAWMESHTASKSEXT, CMD_DRAWMESHTASKSINDIRECTEXT, CMD_DRAWMESHTASKSINDIRECTCOUNTEXT};
+
+    if (graphics_shader_bound) {
```
spencer-lunarg, line 612: I'm actually surprised (and happy) that this whole function is much more readable than I was expecting it to be!

layers/core_checks/cc_drawdispatch.cpp
```diff
@@ -3441,54 +4796,63 @@ bool CoreChecks::PreCallValidateCmdDrawMeshTasksIndirectCountEXT(VkCommandBuffer
...
-        if (pipeline.descriptor_buffer_mode) {
-            if (ds.bound_descriptor_set && !ds.bound_descriptor_set->IsPushDescriptor()) {
-                const LogObjectList objlist(cb_state.Handle(), pipeline.Handle(), ds.bound_descriptor_set->Handle());
-                skip |=
-                    LogError(objlist, vuid.descriptor_buffer_set_offset_missing_08117,
-                             "%s: pipeline bound to %s requires a descriptor buffer but has a bound descriptor set (%s)", function,
-                             string_VkPipelineBindPoint(bind_point), FormatHandle(ds.bound_descriptor_set->Handle()).c_str());
-                break;
-            }
+    if (pipeline) {
```
spencer-lunarg, line 4823: I would just combine all this logic to the other `if (pipeline) {` below

layers/core_checks/cc_drawdispatch.cpp
```diff
@@ -3441,54 +4796,63 @@ bool CoreChecks::PreCallValidateCmdDrawMeshTasksIndirectCountEXT(VkCommandBuffer
...
+                    break;
+                }
             }
         }
     }
 
     if (VK_PIPELINE_BIND_POINT_GRAPHICS == bind_point) {
         skip |= ValidateDrawDynamicState(last_bound_state, cmd_type);
         skip |= ValidatePipelineDrawtimeState(last_bound_state, cmd_type);
+        if ((!last_bound_state.pipeline_state || !last_bound_state.pipeline_state->pipeline()) &&
```
spencer-lunarg, line 4852: lets go

```
// use with pipeline_bound_08606 also
const bool has_last_pipeline = last_pipeline & last_pipeline->pipeline() != nullptr;
```

then here use the feature bit to be consistent
```
if (enabled_features.shader_object_features.shaderObject and !has_last_pipeline)
```

layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -2841,6 +2868,248 @@ bool CoreChecks::ValidatePipelineDrawtimeState(const LAST_BOUND_STATE &last_boun
...
+        } else {
+            msg << "Mesh shader is bound.";
+        }
+        skip |= LogError(objlist, vuid.draw_shaders_no_task_mesh_08885, "%s(): %s", caller, msg.str().c_str());
+    }
+
+    return skip;
+}
+
+bool CoreChecks::ValidateShaderObjectMeshDrawtimeState(const LAST_BOUND_STATE &last_bound_state, CMD_TYPE cmd_type) const {
```
spencer-lunarg, line 3107: there a todo still for mesh?

layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -2937,39 +3206,42 @@ bool CoreChecks::ValidatePipelineRenderpassDraw(const LAST_BOUND_STATE &last_bou
 bool CoreChecks::ValidatePipelineDynamicRenderpassDraw(const LAST_BOUND_STATE &last_bound_state, CMD_TYPE cmd_type) const {
     bool skip = false;
     const CMD_BUFFER_STATE &cb_state = last_bound_state.cb_state;
-    const PIPELINE_STATE &pipeline = *last_bound_state.pipeline_state;
```
spencer-lunarg, line 2940: we do 

```c++
if (pipeline) {
    ValidatePipelineRenderpassDraw(last_bound_state, cmd_type);
}
```

can we do the same for `ValidatePipelineDynamicRenderpassDraw` to keep this reference

otherwise for these function also happy to do something like

```c++
    if (!last_bound_state.pipeline_state) {
        return skip;
    }
    const PIPELINE_STATE &pipeline = *last_bound_state.pipeline_state;
```

layers/core_checks/cc_shader_object.cpp
```diff
@@ -15,12 +15,70 @@
  */
 
 #include "core_validation.h"
+#include "state_tracker/shader_object_state.h"
+
+VkShaderStageFlags findNextStage(uint32_t createInfoCount, const VkShaderCreateInfoEXT* pCreateInfos, VkShaderStageFlagBits stage) {
```
spencer-lunarg, line 20: nit: `FindNextStage`

layers/core_checks/cc_shader_object.cpp
```diff
@@ -15,12 +15,70 @@
...
                                                  const VkShaderCreateInfoEXT* pCreateInfos, const VkAllocationCallbacks* pAllocator,
                                                  VkShaderEXT* pShaders) const {
     bool skip = false;
 
+    if (enabled_features.shader_object_features.shaderObject == VK_FALSE) {
+        skip |=
+            LogError(device, "VUID-vkCreateShadersEXT-None-08400", "vkCreateShadersEXT(): shaderObject feature is not enabled.");
+    }
+
+    uint32_t linked_stage = createInfoCount;
```
spencer-lunarg, line 73: might read nicer with something like

```c++
const uint32_t invalid = createInfoCount;
uint32_t linked_stage = invalid;

// ... 
if (linked_stage != invalid)
```

layers/core_checks/cc_shader_object.cpp
```diff
@@ -41,17 +99,184 @@ bool CoreChecks::PreCallValidateCreateShadersEXT(VkDevice device, uint32_t creat
...
+                         "vkCreateShadersEXT(): createInfo[%" PRIu32
+                         "] is a linked shader with codeType VK_SHADER_CODE_TYPE_SPIRV_EXT, but createInfo[%" PRIu32
+                         "] is a linked shader with code type VK_SHADER_CODE_TYPE_BINARY_EXT.",
+                         linked_spirv_index, linked_binary_index);
+    }
+
+    for (uint32_t i = 0; i < createInfoCount; ++i) {
+        if (pCreateInfos[i].codeType == VK_SHADER_CODE_TYPE_SPIRV_EXT) {
+            const StageCreateInfo stage_create_info("vkCreateShadersEXT", i, pCreateInfos[i]);
+            const auto spirv =
```
spencer-lunarg, line 275: I realize this will conflict with https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6285 and you said the GPU-AV is not working still with that change... so happy to go this route first and look into merging PR 6285 after

layers/state_tracker/pipeline_state.cpp
```diff
@@ -119,68 +128,12 @@ PIPELINE_STATE::StageStateVec PIPELINE_STATE::GetStageStates(const ValidationSta
                 continue;
             }
 
-            stage_states.emplace_back(stage_ci, module_state);
+            // stage_states.emplace_back(stage_ci, nullptr, module_state, module_state->spirv);
```
spencer-lunarg, line 131: (haven't looked at CI failures) but I assume this would be failing a lot of non-shader object tests

spencer-lunarg (commented):
overall looks good to me, large PR, but for an init PR of `VK_EXT_shader_object` I am happy. Did a great job fitting it in a very similar fashion as the pipeline stuff!

I think there are ways we might be able to clean up things (instead of having all these `if (pipeline)` conditions) but that will involve more investigation

Tests look good, few nit/questions

I see a LOT of compiler warnings while building (assuming that is also failing CI)

layers/state_tracker/pipeline_state.h
```diff
@@ -916,3 +882,18 @@ static LvlBindPoint inline ConvertToLvlBindPoint(VkShaderStageFlagBits stage) {
     }
     return BindPoint_Count;
 }
+
+static ShaderObjectStage inline ConvertToShaderObjectStage(VkShaderStageFlagBits stage) {
```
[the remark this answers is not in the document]
ziga-lunarg: Doesn't work because I'm looping through the enums.

layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -504,6 +587,459 @@ bool CoreChecks::ValidateDrawDynamicState(const LAST_BOUND_STATE &last_bound_sta
...
 }
 
+bool CoreChecks::ValidateDrawDynamicStateShaderObject(const LAST_BOUND_STATE &last_bound_state, CMD_TYPE cmd_type) const {
+    bool skip = false;
+    const CMD_BUFFER_STATE &cb_state = last_bound_state.cb_state;
+    const DrawDispatchVuid &vuid = GetDrawDispatchVuid(cmd_type);
+    const LogObjectList objlist(cb_state.commandBuffer());
+
+    bool graphics_shader_bound = false;
+    graphics_shader_bound |= last_bound_state.IsValidShaderBound(ShaderObjectStage::SHADER_OBJECT_VERTEX);
```
spencer-lunarg, line 597: (just noting) this can be `ShaderObjectStage::Vertex` now

layers/state_tracker/pipeline_state.cpp
```diff
@@ -923,4 +876,62 @@ bool LAST_BOUND_STATE::IsRasterizationDisabled() const {
...
+}
+
+bool LAST_BOUND_STATE::IsValidShaderBound(ShaderObjectStage stage) const {
+    if (!shader_object_bound[static_cast<uint32_t>(stage)]) {
+        return false;
+    }
+    return shader_object_states[static_cast<uint32_t>(stage)] != nullptr;
+}
+
+bool LAST_BOUND_STATE::IsValidShaderOrNullBound(ShaderObjectStage stage) const {
```
spencer-lunarg, line 935: we already size the array to be the max enum

`bool shader_object_bound[SHADER_OBJECT_STAGE_COUNT]{false};`

I guess not sure why we need a null bound check version... when would we actually need to use the `IsValidShaderBound`? We are always passing in a valid enum right?

ziga-lunarg: You can also pass VK_NULL_HANDLE.

spencer-lunarg (approved):
Pending final comments - but overall looks good and happy with it

layers/state_tracker/pipeline_state.h
```diff
@@ -789,6 +747,8 @@ struct LAST_BOUND_STATE {
 
     CMD_BUFFER_STATE &cb_state;
     PIPELINE_STATE *pipeline_state{nullptr};
+    bool shader_object_bound[SHADER_OBJECT_STAGE_COUNT]{false};
```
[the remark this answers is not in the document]
ziga-lunarg: No, because VK_NULL_HANDLE can and must be bound if a shader stage wont be used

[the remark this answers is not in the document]
spencer-lunarg: ok, this is something not obvious unless you read the spec. I would just add a comment about this here expressing that was the decision, the next person might also read this and thinks "can't I just use `nullptr` as a `false`?"

layers/state_tracker/pipeline_state.cpp
spencer-lunarg: ok, can we just add a comment about when/why we have two calls... I guess haven't fully grasped when to use each

PR #6295, opened 2023-08-14
layers: Add pnext chain extraction logic

arno-lunarg:
- Should close https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6256

Add utilities to make a selective copy of a pNext chain. Selected elements are the structs extending some reference struct.

### DIFF
--- a/layers/vulkan/generated/pnext_chain_extraction.cpp
+++ b/layers/vulkan/generated/pnext_chain_extraction.cpp
@@ -1,0 +1,135 @@
+// *** THIS FILE IS GENERATED - DO NOT EDIT ***
+// See pnext_chain_extraction_generator.py for modifications
+
+/***************************************************************************
+*
+* Copyright (c) 2023 The Khronos Group Inc.
+* Copyright (c) 2023 Valve Corporation
+* Copyright (c) 2023 LunarG, Inc.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+****************************************************************************/
+
+// NOLINTBEGIN
+
+
+#include "pnext_chain_extraction.h"
+
+#include "vk_typemap_helper.h"
+
+#include <cassert>
+
+namespace vvl {
+       
+void* PnextChainAdd(void *chain_end, void *new_struct) {
+    assert(new_struct);
+    if (!chain_end) {
+    	return new_struct;
+    }
+    auto *vk_base_struct = reinterpret_cast<VkBaseOutStructure*>(chain_end);
+    assert(!vk_base_struct->pNext);
+    vk_base_struct->pNext = reinterpret_cast<VkBaseOutStructure*>(new_struct);
+    return new_struct;
+};
+
+void PnextChainRemoveLast(void *chain) {
+    if (!chain) {
+        return;
+    }
+    auto *current = reinterpret_cast<VkBaseOutStructure *>(chain);
+    auto *prev = current;
+    while (current) {
+        prev = current;
+        current = reinterpret_cast<VkBaseOutStructure *>(current->pNext);
+    }
+    prev->pNext = nullptr;
+}
+
+
+template <>
+PnextChainVkPhysicalDeviceImageFormatInfo2 ExtractPnextChain(const void *in_pnext_chain) {
+
+	std::tuple<void *,
+	VkImageCompressionControlEXT,
+	VkImageFormatListCreateInfo,
+	VkImageStencilUsageCreateInfo,
+	VkOpticalFlowImageFormatInfoNV,
+	VkPhysicalDeviceExternalImageFormatInfo,
+	VkPhysicalDeviceImageDrmFormatModifierInfoEXT,
+	VkPhysicalDeviceImageViewImageFormatInfoEXT,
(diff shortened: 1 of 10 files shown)


scripts/generators/pnext_chain_extraction_generator.py
```diff
@@ -0,0 +1,171 @@
...
+            self.generateHeader()
+        elif self.filename == 'pnext_chain_extraction.cpp':
+            self.generateSource()
+        else:
+            self.write(f'\nFile name {self.filename} has no code to generate\n')
+
+        self.write('\n// NOLINTEND') # Wrap for clang-tidy to ignore
+
+    def generateHeader(self):
+        self.write(f'''#pragma once
```
spencer-lunarg, line 69: nit: use `out = []` and then add with `out.append(` so there is a single File I/O write at end with `self.write("".join(out))`

layers/vulkan/generated/pnext_chain_extraction.cpp
```diff
@@ -0,0 +1,134 @@
...
+#include "pnext_chain_extraction.h"
+
+#include "vk_typemap_helper.h"
+
+#include <cassert>
+
+namespace vvl {
+
+
+void* PnextChainAdd(void *chain_end, void *new_elt) {
```
spencer-lunarg, line 35: `elt`? I guess not sure if you mean `element`? can we just call this `new_struct` or `new_element`

layers/vulkan/generated/pnext_chain_extraction.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+    std::tuple<VkBaseOutStructure,
+	VkImageCompressionControlEXT,
+	VkImageFormatListCreateInfo,
+	VkImageStencilUsageCreateInfo,
+	VkOpticalFlowImageFormatInfoNV,
+	VkPhysicalDeviceExternalImageFormatInfo,
+	VkPhysicalDeviceImageDrmFormatModifierInfoEXT,
+	VkPhysicalDeviceImageViewImageFormatInfoEXT,
+	VkVideoProfileListInfoKHR> out_structs;
```
spencer-lunarg, line 69: won't all the `pNext` be pointing the pointer where these are on the stack in `ExtractPnextChain`? so when we call the dispatch call these will have been released and just garbage `pNext` pointers?

scripts/generators/pnext_chain_extraction_generator.py
```diff
@@ -0,0 +1,171 @@
...
+// If the returned tuple is stack allocated, each struct is NOT a deep copy of the corresponding struct in in_pnext_chain.
+// So be mindful of pointers copies.
+template <typename T>
+T ExtractPnextChain(const void *in_pnext_chain) {{}}
+
+''')
+        # Declare functions
+        for struct_name in self.target_structs:
+            struct = self.vk.structs[struct_name]
+            self.write(f'using PnextChain{struct_name} = {self.getTupleDefinition(struct)};')
```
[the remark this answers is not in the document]
arno-lunarg: Sorry, I do not get what you want to do?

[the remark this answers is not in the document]
arno-lunarg: Ah ok, you want to get rid of the helpers `getTupleDefinition`, `getFuncPrototype`...?
They allow me to factorise code, and while developing this it ended up saving me some time

layers/vulkan/generated/pnext_chain_extraction.cpp
arno-lunarg: I think my preceding answer completely misses your question: lifetimes are fine, as the tuple storing the pnext chain is returned from the call to `vvl::ExtractPnextChain`, it is allocated on the stack at the call site, like so:
```cpp
auto image_format_info_pnext_chain =
        vvl::ExtractPnextChain<vvl::PnextChainVkPhysicalDeviceImageFormatInfo2>(pCreateInfo->pNext);
```

ASAN should catch any issues, but in this PR, dispatch calls live in the same scope as the returned tuple and I do not expect memory accesses issues

scripts/generators/pnext_chain_extraction_generator.py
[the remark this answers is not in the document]
spencer-lunarg: so I understand this idea of in-lining 3 lines of logic in multiple spaces seems like the most-counterintuitive way of thinking, but I firmly believe after all my time trying to "clean up" and "debug" the old code gen that the Number 1 issue was that it slowly become hard to know where code in the `.cpp` file was being created from as I started having to jump through many functions

I argue while it might take some more time in development, it in the long runs will save us time maintaining the code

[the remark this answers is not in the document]
arno-lunarg: Ok, as you say you have more experience than me, so I trust you on this and will do the change
(But yes this is not intuitive ^.^)

layers/vulkan/generated/pnext_chain_extraction.h
```diff
@@ -0,0 +1,63 @@
+// *** THIS FILE IS GENERATED - DO NOT EDIT ***
+// See pnext_chain_extraction_generator.py for modifications
+
+/***************************************************************************
+*
+* Copyright (c) 2015-2023 The Khronos Group Inc.
+* Copyright (c) 2015-2023 Valve Corporation
+* Copyright (c) 2015-2023 LunarG, Inc.
```
juan-lunarg, line 8:
```suggestion
* Copyright (c) 2023 The Khronos Group Inc.
* Copyright (c) 2023 Valve Corporation
* Copyright (c) 2023 LunarG, Inc.
```

juan-lunarg (commented):
Could a test be added to prevent a regression apologies if I missed some context around this.

spencer-lunarg:
> Could a test be added to prevent a regression apologies if I missed some context around this.

So I have been trying to reproduce the issue with Zink locally (no success yet)... One way is to add yet-another hard-coded path in MockICD to return the status code we want... but at that point we are really just making MockICD a "general purpose null driver" into a VVL only testing library

PR #6296, opened 2023-08-14
android: Remove ndk-build

juan-lunarg:
Jenkins CI no longer uses ndk-build for vvl


build-android/known_good.json
juan-lunarg, line 1: One less known_good.json to worry about.

FYI @mikes-lunarg

spencer-lunarg (approved):
:100:

PR #6299, opened 2023-08-15
android: Move APK XML files

juan-lunarg:
Allows Jenkins CI to clean Android builds

### DIFF
--- a/tests/android/AndroidManifest.xml
+++ b/tests/android/AndroidManifest.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.example.VulkanLayerValidationTests" android:versionCode="1" android:versionName="1.0">
+
+    <uses-sdk android:minSdkVersion="26" android:targetSdkVersion="26"/>
+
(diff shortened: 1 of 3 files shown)


juan-lunarg:
Note once this PR and the Jenkins PR get merged in I will fully delete the build-android directory once and for all.

This needs to be done in lockstep due to the CI code being in another repo.

tests/android/AndroidManifest.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.example.VulkanLayerValidationTests" android:versionCode="1" android:versionName="1.0">
+
+    <!-- This is the platform API where NativeActivity was introduced. -->
```
artem-lunarg, line 4: nit: this comment might be out dated, in case information that NativeActivity was introduced in SDK 9 is correct.

artem-lunarg (approved):
LGTM

tests/android/AndroidManifest.xml
juan-lunarg: Good catch I'll fix that.

PR #6300, opened 2023-08-15
tests: Remove VVL_TESTS_USE_CMAKE

juan-lunarg:
Not needed now that Android uses CMake


tests/framework/test_common.h
```diff
@@ -47,13 +48,6 @@
 #undef Bool
 #endif
 
-// Use the NDK's header on Android
-#ifndef __ANDROID__
-#include "gtest/gtest.h"
-#else
-#include "gtest/gtest.h"
-#endif
-
```
juan-lunarg, line 56: For some context. The Android NDK comes bundled with google test. The ndk-build was using that one. Now the android build acts like the desktop build. Meaning we can always guarantee the version of gtest used.

spencer-lunarg (approved):

PR #6302, opened 2023-08-16
layers: Fix VK_EXT_debug_report callback wiith no device

spencer-lunarg:
@Tony-LunarG pointed out some CTS was crashing on errors when using `VK_EXT_debug_report` (which is deprecated now)

This adds a fix to prevent the crash


spencer-lunarg:
closing, @Tony-LunarG can reproduce so will help get a proper fix in tomorrow

PR #6307, opened 2023-08-16
tests: Remove vk_result_string

juan-lunarg:
Use string_VkResult instead


spencer-lunarg (approved):
always happy to see the test framework unify to the same way of doing things!

PR #6311, opened 2023-08-17
tests: Remove xxhash.cpp dependency

juan-lunarg:
Now that the layer factor is removed from LunarG/VulkanTools there is 1 less dependency on the validation layers.


juan-lunarg:
FYI @charles-lunarg who removed VLF

spencer-lunarg (approved):

PR #6312, opened 2023-08-17
layers: fix typo in 06080 error description

brakhane:
The error checks for the height of the render area, but mentioned `pRenderingInfo->renderArea.extent.width` instead.


juan-lunarg (approved):

juan-lunarg:
Thanks for the fix @ brakhane 👍🏾

PR #6319, opened 2023-08-19
layers: Fix error message with push descriptors

ziga-lunarg:


layers/core_checks/cc_descriptor.cpp
```diff
@@ -968,11 +968,17 @@ bool CoreChecks::ValidateDescriptor(const DescriptorContext &context, const Desc
...
                     auto set = context.descriptor_set.GetSet();
-                    return LogError(
-                        set, context.vuids.descriptor_buffer_bit_set_08114,
-                        "%s: Descriptor set %s Image layout specified at vkCmdBindDescriptorSets time doesn't match actual image "
-                        "layout at time descriptor is used. See previous error callback for specific details.",
-                        context.caller, FormatHandle(set).c_str());
+                    std::stringstream msg;
+                    if (!context.descriptor_set.IsPushDescriptor()) {
+                        msg << "Descriptor set " << FormatHandle(set)
+                            << " Image layout specified at vkCmdBindDescriptorSets time doesn't "
```
artem-lunarg, line 974: nit: I see that *"at vkCmdBindDescriptorSets time"* was used in the original version, but does it a bit unusual usage of "time"? I heard terms "submit time" or "command buffer recording time", but not with regard to specific command. Maybe it's used and I did not pay attention. One alternative, if you like this better is *"specified by vkCmdBindDescriptorSets"*.

artem-lunarg (approved):
LGTM

layers/core_checks/cc_descriptor.cpp
ziga-lunarg: Fixed

PR #6321, opened 2023-08-19
layers: Remove CMD_TYPE for Location

spencer-lunarg:
Remove `CMD_TYPE` in favor of the new `Location` (or `Func`)

the old `CMD_TYPE` enum was hard to read, has been a source of bug, and is really a duplicate of what we now have with the `Func` enum


layers/core_checks/cc_cmd_buffer.cpp
```diff
@@ -424,7 +423,7 @@ bool CoreChecks::PreCallValidateCmdBindIndexBuffer(VkCommandBuffer commandBuffer
                                                    VkIndexType indexType, const ErrorObject &errorObj) const {
     auto buffer_state = Get<BUFFER_STATE>(buffer);
     auto cb_state = GetRead<CMD_BUFFER_STATE>(commandBuffer);
-    return ValidateCmdBindIndexBuffer(*cb_state, *buffer_state, offset, indexType, CMD_BINDINDEXBUFFER);
+    return ValidateCmdBindIndexBuffer(*cb_state, *buffer_state, offset, indexType, errorObj.location);
```
spencer-lunarg, line 426: with this approach, when people need to reuse the code, there is a less chance of we send in the wrong function name

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -3135,7 +3126,6 @@ bool CoreChecks::ValidateCmdBlitImage(VkCommandBuffer commandBuffer, VkImage src
         skip |= ValidateImageUsageFlags(commandBuffer, *src_image_state, VK_IMAGE_USAGE_TRANSFER_SRC_BIT, true, vuid, func_name);
         vuid = is_2 ? "VUID-VkBlitImageInfo2-dstImage-00224" : "VUID-vkCmdBlitImage-dstImage-00224";
         skip |= ValidateImageUsageFlags(commandBuffer, *dst_image_state, VK_IMAGE_USAGE_TRANSFER_DST_BIT, true, vuid, func_name);
-        skip |= ValidateCmd(cb_state, cmd_type);
```
spencer-lunarg, line 3138: this was being called twice in this function

layers/core_checks/core_validation.h
```diff
@@ -343,7 +343,7 @@ class CORE_CMD_BUFFER_STATE : public CMD_BUFFER_STATE {
                           const COMMAND_POOL_STATE* cmd_pool)
         : CMD_BUFFER_STATE(dev_data, cb, pCreateInfo, cmd_pool) {}
 
-    void RecordWaitEvents(CMD_TYPE cmd_type, uint32_t eventCount, const VkEvent* pEvents,
+    void RecordWaitEvents(vvl::Func command, uint32_t eventCount, const VkEvent* pEvents,
```
arno-lunarg, line 346: What is the reason behind sometimes just passing a `vvl::Func` and other times a `Location`?

arno-lunarg (approved):
LGTM! One question

layers/core_checks/core_validation.h
spencer-lunarg: `vvl::Func` is just a enum (basically what `CMD_TYPE` was too)

`Location`/`ErrorObject` is a struct and not sure if the ownership would be correctly maintained as we store alot of these records until queue submit time (need to investigate more)

spencer-lunarg: another things, `Location` is designed to know the `Struct` and `Field` while a lot of the places being replaced here really just need the `Func` only because it is checking the function name only

The `Location` is used in the `LogError` but this PR is mainly touching state-tracking which doesn't have any error to print

arno-lunarg: Ok, thank you for those details!

PR #6322, opened 2023-08-20
scripts: Move ApiVersion to own file

spencer-lunarg:
I have found that the `ApiVersion` will be a desired class for other repos and envision this just going into the Vulkan-Utility-Libraries as well. The only reason it was ever in `vk_extension_helper.h` was it was originally only used there to start, but has since got used all over the code base now

@aqnuep this is a good first attempt to see how we can work with the `APISpecific` hooks, open to adjusting the scripts in whatever way is easier for you. I wasn't sure if would be easier to have the `vk_api_version.h` file just be generated differently, or in `generate_source.py` to have a check and generate a `vk_sc_api_version.h` file instead (but I realize that would cause a different `include` so assume this is the path you would want to take)


scripts/generators/api_version_generator.py
```diff
@@ -0,0 +1,134 @@
...
+static inline std::string StringAPIVersion(APIVersion version) {
+    std::stringstream version_name;
+    if (!version.Valid()) {
+        return "<unrecognized>";
+    }
+    version_name << version.Major() << "." << version.Minor() << "." << version.Patch() << " (0x" << std::setfill('0')
+                 << std::setw(8) << std::hex << version.Value() << ")";
+    return version_name.str();
+}
+''')
```
aqnuep, line 131: Really, this is only special in the sense that 80% of the generated header's content is API specific. If you move back the APIVersion class and the NormalizeApiVersion function to the APISpecific interface (called the same way as before, or really, any way you wish), then this should be fine.

However, the PR, as is, is a great example how **not** to do things, because this change moves the APIVersion class and the NormalizeApiVersion from the APISpecific class into an inlined string in the common code, which is exactly what we should avoid, as it would require us to re-create the removed APISpecific interface and re-split the source as it was before at the next time we downstream, as the current form of the PR undoes the code I've previously upstreamed.

aqnuep:
Could you, please review #6194 and let me know whether the proposed approach for testing the downstream interface looks fine to you, because I'd rather not spend time on updating that PR and kicking it through CI until I actually get some sort of confirmation that the proposed downstream interface tests would be fine with LunarG.

scripts/generators/api_version_generator.py
```diff
@@ -0,0 +1,133 @@
...
+****************************************************************************/\n''')
+        out.append('// NOLINTBEGIN') # Wrap for clang-tidy to ignore
+
+        out.append('''
+#pragma once
+#include <vulkan/vulkan.h>
+#include <sstream>
+#include <iomanip>
+
+#define VK_VERSION_1_1_NAME "VK_VERSION_1_1"
```
charles-lunarg, line 110: why not generate the versions?

arno-lunarg (approved):

scripts/generators/api_version_generator.py
charles-lunarg: Works for me! I was mostly being persnickety - minor version updates require a lot more than just re-running codegen even in the best case. Loader explicitly doesn't generate code for minor versions so this tracks.

PR #6326, opened 2023-08-21
layers: Avoid allocation for single semaphore wait

artem-lunarg:
`TimePoint` objects are regularly created and destroyed.  When such
object registers semaphore wait operation it allocates memory.
This PR avoids allocation when semaphore has a single waiter.

*First commit:*
Reduces size of `TimePoint` structure on 64 bit platforms.
The numbers is for the Release build, Debug build can add additional
debug state.

SemOp    : 40  -> 32 bytes
TimePoint: 104 -> 96 bytes

That's preparation for the second commit which increases size of
`TimePoint` in order to use build-in storage of `small_vector` to avoid
allocation for the case when semaphore has a single waiter.

We try to reduce size enough, in order not to go over 2 cache lines
(128 bytes) when `std::set` is replaced with `small_vector`. If for some
compiler settings or std library implementation we start using 3rd cache
line that should be fine too (avoid allocation is the main priority). Additional 
cache line can optionally be used to add more static slots to a small_vector.

*Second commit:*
Replaces `std::set` with a `small_vector`.

The first commit allowed to reduce sizes of `SemOp`/`TimPoint` structures
and here we borrow that size back. `TimePoint` structure in Release build
on x64 grows to 128 bytes (perfectly fits 2 cache lines).

The size of `SemOp` remains 32 bytes. In the cases with multiple waiters we'll
be able to nicely pack them in cache lines.

The original code did not use properties of set container, only iteration over
all elements, so replacing with `small_vector` is straightforward.

*Third commit:*
Added asserts to validate assumptions.

### DIFF
--- a/layers/core_checks/core_validation.h
+++ b/layers/core_checks/core_validation.h
@@ -397,17 +397,18 @@ struct SemaphoreSubmitState {
     bool ValidateWaitSemaphore(const Location& loc, VkSemaphore semaphore, uint64_t value);
     bool ValidateSignalSemaphore(const Location& loc, VkSemaphore semaphore, uint64_t value);
 
-    bool CannotSignal(const SEMAPHORE_STATE& semaphore_state, VkQueue& other_queue, const char*& other_func) const {
+    bool CannotSignal(const SEMAPHORE_STATE& semaphore_state, VkQueue& other_queue, vvl::Func& other_command) const {
         const auto semaphore = semaphore_state.semaphore();
         if (signaled_semaphores.count(semaphore)) {
             other_queue = queue;
+            other_command = vvl::Func::Empty;
             return true;
         }
         if (!unsignaled_semaphores.count(semaphore)) {
             const auto last_op = semaphore_state.LastOp();
             if (last_op && !last_op->CanBeSignaled()) {
                 other_queue = last_op->queue ? last_op->queue->Queue() : VK_NULL_HANDLE;
-                other_func = last_op->func_name;
+                other_command = last_op->command;
                 return true;
             }
         }
--- a/layers/state_tracker/queue_state.h
+++ b/layers/state_tracker/queue_state.h
@@ -21,9 +21,9 @@
 #include <condition_variable>
 #include <deque>
 #include <future>
-#include <set>
 #include <thread>
 #include <vector>
+#include "containers/custom_containers.h"
 #include "error_message/error_location.h"
 #include "utils/vk_layer_utils.h"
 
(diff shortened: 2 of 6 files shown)


layers/core_checks/core_validation.h
```diff
@@ -407,7 +407,7 @@ struct SemaphoreSubmitState {
             const auto last_op = semaphore_state.LastOp();
             if (last_op && !last_op->CanBeSignaled()) {
                 other_queue = last_op->queue ? last_op->queue->Queue() : VK_NULL_HANDLE;
-                other_func = last_op->func_name;
+                other_func = String(last_op->command);
```
jeremyg-lunarg, line 410: you could do even more cleanup in this file, such as returning the enum from CannotSignal() and only converting back to a string if there's an error.

jeremyg-lunarg (commented):
LGTM, but the commit message should say small_vector instead of array. And the PR message should mention both of the things happening here.

layers/core_checks/core_validation.h
artem-lunarg: Good observation. Fixed.

artem-lunarg: Updated PR message.

layers/state_tracker/queue_state.h
```diff
@@ -141,13 +142,13 @@ class SEMAPHORE_STATE : public REFCOUNTED_NODE {
...
             if (op.op_type == kWait) {
-                wait_ops.emplace(op);
+                wait_ops.emplace_back(op);
             } else {
                 signal_op.emplace(op);
             }
         }
         std::optional<SemOp> signal_op;
-        std::set<SemOp> wait_ops;
+        small_vector<SemOp, 1, uint32_t> wait_ops;
```
jeremyg-lunarg, line 151: Thanks to the new PR description I have another question. :) There are various loops in queue_state.cpp over wait_ops that I think assume the iteration is in order.  Do those still work?

artem-lunarg: Oh, I did not think about that. Let me check.

artem-lunarg: If ordering is broken, then I still would prefer vector and do inserts, my observation so far that apps usually don't have many waits (I believe I saw 5 once). Please let me know what do you think. In case something wild happens we might consider using some pool scheme later to avoid real heap allocs with many waits (but I hope good apps don't do this).

jeremyg-lunarg: yeah your way is worth a try. I think you're right that > 1 waiters is rare. You might try running some of the CTS tests to confirm. Another option would be to make small_set<>

artem-lunarg: According to code wait ops are sorted based on payload value, and specific timepoint collects wait ops with the same payload. With this setup `std::set` does not provide additional ordering (sorts the same payload value). Added asserts to validation this assumption.

If the assumption is correct then it feels like an opportunity for another PR to try reduce data duplication (in case it does not increase cost of accessing that data).

artem-lunarg: Will run tomorrow CTS and maybe will try some games. DOOM? Is it the idea that CTS implements more real-app scenarios comparing to VVL tests?

jeremyg-lunarg: CTS has quite a few tests for timeline semaphores that have caught many bugs for me in the past.

spencer-lunarg (approved):
pending Jeremy questions, but happy to see use now removing all these string literals!

jeremyg-lunarg (approved):
LGTM!

PR #6327, opened 2023-08-21
docs: Correct SPIR-V compilation instructions

Eearslya:
The documentation for debug printf includes a couple example shaders and a short example command for how to compile them. These commands include the `-x` flag which exports the code in text-based hexadecimal format, rather than the binary SPIR-V representation, which might be confusing for someone looking to learn from the document.

### DIFF
--- a/docs/debug_printf.md
+++ b/docs/debug_printf.md
@@ -40,7 +40,7 @@ debugPrintfEXT("My float is %f", myfloat);
 }
 ```
 Then use glslangValidator to generate SPIR-V to use in vkCreateShaderModule.
-"glslangvalidator --target-env vulkan1.2 -x -e main -o shader.vert.spv shader.vert" would be one
+"glslangvalidator --target-env vulkan1.2 -e main -o shader.vert.spv shader.vert" would be one
 example of compiling shader.vert
 
 Note that every time this shader is executed, "My float is 3.141500" will be printed. If this were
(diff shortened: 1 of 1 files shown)


docs/debug_printf.md
```diff
@@ -40,7 +40,7 @@ debugPrintfEXT("My float is %f", myfloat);
 }
 ```
 Then use glslangValidator to generate SPIR-V to use in vkCreateShaderModule.
-"glslangvalidator --target-env vulkan1.2 -x -e main -o shader.vert.spv shader.vert" would be one
+"glslangvalidator --target-env vulkan1.2 -e main -o shader.vert.spv shader.vert" would be one
```
juan-lunarg, line 43: I believe this can now just be glslang instead of glslangvalidator.

https://github.com/KhronosGroup/glslang/commit/856e280502d762aa81cb7653bd522271b53a4f40

Thoughts

Eearslya: I wonder if that might be a bit premature, since that commit hasn't actually been released in an SDK yet.

It would be accurate to what is currently in the repositories, but not accurate for users currently looking up documentation. Not sure what the priority should be.

ncesario-lunarg: I would say whatever is current should be used here, which I think is `glslang`. In order to see documentation that "matches" one's SDK binaries, one has to look at the documentation associated with the specific SDK release they have. So either the documentation that comes packaged with the SDK, or looking at the source at the tagged commit in VVL history is required. In general, ToT VVL documentation will not be correct for any particular SDK X.Y.Z in the past.

That said, `glslangValidator` is still a glslang build artifact and packaged in the SDK, and probably not going anywhere any time soon, so it would also be valid to use that.

It is ultimately up to @KhronosGroup/vvl-codeowners on what should be used here.

Tony-LunarG (commented):
Removing the -x LGTM.  I'd lean slightly toward going with glslang so we don't forget to change it later, but I have no strong opinion

juan-lunarg (approved):

spencer-lunarg (approved):

spencer-lunarg:
Seems only thing failing is the commit message

I think the commit message body is over the 72-char per-line limit and just needs to add some breaks in it

Eearslya:
That should fix the commit length issue, I hope.

juan-lunarg:
Thanks for noticing / fixing this @Eearslya 👍🏾

PR #6329, opened 2023-08-22
layers: Combine alias functions calls

spencer-lunarg:
This reduces a lot of duplication code with alias/promoted function hooks,

With the new `ErrorObject` we always can know what function the call came from and there is not reason to keep 2 paths now.

A future goal would be maybe just having the chassis calling the non-KHR/EXT function right away saving another function jump


layers/stateless/stateless_validation.h
```diff
@@ -838,6 +826,9 @@ class StatelessValidation : public ValidationObject {
     bool manual_PreCallValidateCmdDispatchIndirect(VkCommandBuffer commandBuffer, VkBuffer buffer, VkDeviceSize offset,
                                                    const ErrorObject &errorObj) const;
 
+    bool manual_PreCallValidateCmdDispatchBase(VkCommandBuffer commandBuffer, uint32_t baseGroupX, uint32_t baseGroupY,
```
spencer-lunarg, line 829: example case where we were not validating things from the promoted `vkCmdDispatchBase` but were only for `vkCmdDispatchBaseKHR`

I think its not realistic to test every promotion, but instead to generate it in future as that is only one point of failure then

layers/sync/sync_validation.cpp
```diff
@@ -6376,6 +6315,10 @@ bool SyncValidator::PreCallValidateCmdSetEvent2(VkCommandBuffer commandBuffer, V
     const auto *cb_context = &cb_state->access_context;
     if (!pDependencyInfo) return skip;
 
+    const auto *access_context = cb_context->GetCurrentAccessContext();
```
spencer-lunarg, line 6318: notice we had these 3 lines in `PreCallValidateCmdSetEvent2KHR` but not `PreCallValidateCmdSetEvent2`

layers/core_checks/cc_device.cpp
```diff
@@ -634,10 +634,9 @@ bool CoreChecks::PreCallValidateGetPhysicalDeviceImageFormatProperties2KHR(VkPhy
                                                                            const VkPhysicalDeviceImageFormatInfo2 *pImageFormatInfo,
                                                                            VkImageFormatProperties2 *pImageFormatProperties,
                                                                            const ErrorObject &errorObj) const {
-    // Can't wrap AHB-specific validation in a device extension check here, but no harm
-    bool skip = ValidateGetPhysicalDeviceImageFormatProperties2ANDROID(pImageFormatInfo, pImageFormatProperties, errorObj);
-    skip |= ValidateGetPhysicalDeviceImageFormatProperties2(pImageFormatInfo, pImageFormatProperties, errorObj);
-    return skip;
+    return PreCallValidateGetPhysicalDeviceImageFormatProperties2(physicalDevice, pImageFormatInfo, pImageFormatProperties,
+                                                                  errorObj);
+    ;
```
artem-lunarg, line 639: nit: ;

layers/sync/sync_validation.cpp
artem-lunarg: looks safe for me

artem-lunarg (approved):
LGTM!

PR #6330, opened 2023-08-22
Add vkGetImageSubresourceLayout2KHR

spencer-lunarg:
image

All the VUIDs are old as they have been changed to the KHR version

Added test to make sure this works with `VK_KHR_maintenance5` as expected

### DIFF
--- a/layers/core_checks/cc_image.cpp
+++ b/layers/core_checks/cc_image.cpp
@@ -2068,34 +2068,34 @@ bool CoreChecks::PreCallValidateDestroyImageView(VkDevice device, VkImageView im
     }
     return skip;
 }
-bool CoreChecks::ValidateGetImageSubresourceLayout(VkDevice device, const IMAGE_STATE &image_state,
-                                                   const VkImageSubresource &subresource, VkSubresourceLayout &layout,
-                                                   bool is_ext) const {
+
+bool CoreChecks::ValidateGetImageSubresourceLayout(const IMAGE_STATE &image_state, const VkImageSubresource &subresource,
+                                                   const Location &loc) const {
     bool skip = false;
-    const char *caller = is_ext ? "vkGetImageSubresourceLayout2EXT()" : "vkGetImageSubresourceLayout()";
+    const bool is_2 = loc.function != Func::vkGetImageSubresourceLayout;
     const VkImageAspectFlags aspect_mask = subresource.aspectMask;
 
     // The aspectMask member of pSubresource must only have a single bit set
     if (GetBitSetCount(aspect_mask) != 1) {
         const char *vuid =
-            is_ext ? "VUID-vkGetImageSubresourceLayout2EXT-aspectMask-00997" : "VUID-vkGetImageSubresourceLayout-aspectMask-00997";
-        skip |= LogError(image_state.image(), vuid, "%s: VkImageSubresource.aspectMask must have exactly 1 bit set (currently %s).",
-                         caller, string_VkImageAspectFlags(aspect_mask).c_str());
+            is_2 ? "VUID-vkGetImageSubresourceLayout2KHR-aspectMask-00997" : "VUID-vkGetImageSubresourceLayout-aspectMask-00997";
+        skip |= LogError(vuid, image_state.image(), loc.dot(Field::aspectMask), "(%s) must have exactly 1 bit set.",
+                         string_VkImageAspectFlags(aspect_mask).c_str());
     }
 
     // mipLevel must be less than the mipLevels specified in VkImageCreateInfo when the image was created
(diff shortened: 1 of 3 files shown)


layers/core_checks/core_validation.h
```diff
@@ -2608,15 +2608,17 @@ class CoreChecks : public ValidationStateTracker {
...
+                                           const Location& loc) const;
     bool PreCallValidateTransitionImageLayoutEXT(VkDevice device, uint32_t transitionCount,
                                                  const VkHostImageLayoutTransitionInfoEXT* pTransitions,
                                                  const ErrorObject& errorObj) const override;
     void PostCallRecordTransitionImageLayoutEXT(VkDevice device, uint32_t transitionCount,
                                                 const VkHostImageLayoutTransitionInfoEXT* pTransitions, VkResult result) override;
     bool PreCallValidateGetImageSubresourceLayout(VkDevice device, VkImage image, const VkImageSubresource* pSubresource,
                                                   VkSubresourceLayout* pLayout, const ErrorObject& errorObj) const override;
+    bool PreCallValidateGetImageSubresourceLayout2KHR(VkDevice device, VkImage image, const VkImageSubresource2EXT* pSubresource,
+                                                      VkSubresourceLayout2EXT* pLayout, const ErrorObject& errorObj) const override;
```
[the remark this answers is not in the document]
spencer-lunarg: awww, good catch, should be `KHR`

layers/core_checks/cc_image.cpp
```diff
@@ -2068,34 +2068,34 @@ bool CoreChecks::PreCallValidateDestroyImageView(VkDevice device, VkImageView im
     }
     return skip;
 }
-bool CoreChecks::ValidateGetImageSubresourceLayout(VkDevice device, const IMAGE_STATE &image_state,
-                                                   const VkImageSubresource &subresource, VkSubresourceLayout &layout,
-                                                   bool is_ext) const {
+
+bool CoreChecks::ValidateGetImageSubresourceLayout(const IMAGE_STATE &image_state, const VkImageSubresource &subresource,
+                                                   const Location &loc) const {
```
artem-lunarg, line 2073: By reading implementation I was not sure what does loc mean here. Is it a function or a structure was appended too, I checked caller site to be sure. Do you think it can be helpful to use more detailed names for Location parameters in such case? (e.g. subresource_loc)

artem-lunarg (approved):
LGTM. One question.

layers/core_checks/cc_image.cpp
```diff
@@ -2193,24 +2187,22 @@ bool CoreChecks::ValidateGetImageSubresourceLayout(VkDevice device, const IMAGE_
...
-                                 "VK_IMAGE_ASPECT_MEMORY_PLANE_i_BIT_EXT, with i being less than the "
-                                 "VkDrmFormatModifierPropertiesEXT::drmFormatModifierPlaneCount (%" PRIu32
-                                 ") associated with the image's format (%s) and "
-                                 "VkImageDrmFormatModifierPropertiesEXT::drmFormatModifier (%" PRIu64 ").",
-                                 caller, string_VkImageAspectFlags(aspect_mask).c_str(), max_plane_count,
+                    is_2 ? "VUID-vkGetImageSubresourceLayout2KHR-tiling-02271" : "VUID-vkGetImageSubresourceLayout-tiling-02271";
+                skip |= LogError(vuid, image_state.image(), loc.dot(Field::aspectMask),
+                                 "is %s and drmFormatModifierPlaneCount (%" PRIu32
+                                 ") which associated with the image's format (%s) and "
+                                 "drmFormatModifier (%" PRIu64 ").",
```
arno-lunarg: Wording looks weird to me, I suggest something along those lines:
```suggestion
                skip |= LogError(vuid, image_state.image(), loc.dot(Field::aspectMask),
                                 "is %s, hence the plane index is either unspecified or superior or equal to the drmFormatModifierPlaneCount (%" PRIu32
                                 ") associated with the image's format (%s) and "
                                 "drmFormatModifier (%" PRIu64 ").",
```

layers/core_checks/cc_image.cpp
spencer-lunarg: I think you brought up what I forsee being the biggest issue with the new `Location` system, I agree we should be more descriptive with the names (like `subresource_loc` here)

I think `loc` should be reserved for the top level items (ex, the`pCreateInfo` for a creation call as its the only "real" parameter  to have passed in) 

will update here

arno-lunarg (approved):
LGTM! One comment, now tagged as `outdated` but is is not

layers/core_checks/cc_image.cpp
spencer-lunarg: agree, re-worded it

PR #6331, opened 2023-08-22
layers: Fix getting shader stage spirv

ziga-lunarg:


spencer-lunarg (approved):
for the moment happy with this

Not sure if would be better to have a helper in `PipelineStageState` to get the `SPIRV_MODULE_STATE`, but its definitely not obvious what the "pipeline" class has a way to get `spirv_state` (without knowing `VK_EXT_shader_object` exists).... marking this up as slight code debt to redesign/renamae things

Tony-LunarG (approved):
Fixes the CTS tests that were asserting

PR #6332, opened 2023-08-22
layers: Add Location for ObjectTracker (part 1)

spencer-lunarg:
This change is just replacing the old `const char*` with `Location` for the `ObjectTracker` helpers (While adding some `Location` to errors that didn't print where it came from

My next goal would be to provide a nested location where possible

### DIFF
--- a/tests/unit/sparse.cpp
+++ b/tests/unit/sparse.cpp
@@ -1079,3 +1079,43 @@ TEST_F(NegativeSparse, BufferFlagsFeature) {
     m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkBufferCreateInfo-flags-00917");
     CreateBufferTest(*this, &buffer_create_info, "VUID-VkBufferCreateInfo-flags-00918");
 }
+
+TEST_F(NegativeSparse, VkSparseMemoryBindMemory) {
+    TEST_DESCRIPTION("test VkSparseMemoryBind::memory is valid");
+
+    ASSERT_NO_FATAL_FAILURE(Init());
+
+    VkBufferCreateInfo buffer_create_info = LvlInitStruct<VkBufferCreateInfo>();
+    buffer_create_info.usage = VK_BUFFER_USAGE_TRANSFER_SRC_BIT;
+    buffer_create_info.size = 1024;
+    buffer_create_info.queueFamilyIndexCount = 0;
+
+    if (m_device->phy().features().sparseResidencyBuffer) {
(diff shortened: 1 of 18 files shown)


tests/unit/sparse.cpp
```diff
@@ -892,4 +892,44 @@ TEST_F(NegativeSparse, BufferFlagsFeature) {
...
+    VkSparseBufferMemoryBindInfo buffer_memory_bind_info = {};
+    buffer_memory_bind_info.buffer = buffer.handle();
+    buffer_memory_bind_info.bindCount = 1;
+    buffer_memory_bind_info.pBinds = &buffer_memory_bind;
+
+    VkBindSparseInfo bind_info = LvlInitStruct<VkBindSparseInfo>();
+    bind_info.bufferBindCount = 1;
+    bind_info.pBufferBinds = &buffer_memory_bind_info;
+
+    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkSparseMemoryBind-memory-parameter");
```
spencer-lunarg, line 932: Currently is (and was before this PR)

> vkQueueBindSparse(): Invalid VkDeviceMemory Object 0xbaadbeef.

but after getting this first PR in (to make code review easier) the final desired error would be

> vkQueueBindSparse(): pBindInfo[0].pBufferBinds[0].pBinds[0].memory Invalid VkDeviceMemory Object 0xbaadbeef

tests/unit/sparse.cpp
```diff
@@ -892,4 +892,44 @@ TEST_F(NegativeSparse, BufferFlagsFeature) {
     buffer_create_info.flags = VK_BUFFER_CREATE_SPARSE_ALIASED_BIT;
     m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkBufferCreateInfo-flags-00917");
     CreateBufferTest(*this, &buffer_create_info, "VUID-VkBufferCreateInfo-flags-00918");
-}
\ No newline at end of file
+}
+
+TEST_F(PositiveSparse, VkSparseMemoryBindMemory) {
```
juan-lunarg, line 897: Shouldn't this be `NegativeSparse` since it's triggering an error message?

juan-lunarg (approved):

tests/unit/sparse.cpp
arno-lunarg: Sooo much better

arno-lunarg (approved):

tests/unit/sparse.cpp
spencer-lunarg: ....yes, good catch

tests/unit/sparse.cpp
spencer-lunarg: I guess what would be your "prefect error message" (without caring how we do the `LogError` under the hood)

PR #6342, opened 2023-08-23
tests: Cleanup, move and rename sparse tests

arno-lunarg:
Cleanups mainly consist in using `VkXXXObj`, using `auto` where it makes sense, moving variable declaration to the line before a variable is actually needed.
Also create dedicated test files for sparse buffers and sparse images, they were bundled together.
Renamed some tests.


arno-lunarg:
`PositiveSparse.BindFreeMemory` is on CI blacklist, I will see how it goes

juan-lunarg (approved):

tests/unit/sparse_buffer_positive.cpp
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    const std::optional<uint32_t> sparse_index = m_device->QueueFamilyMatching(VK_QUEUE_SPARSE_BINDING_BIT, 0u);
+    if (!sparse_index) {
+        GTEST_SKIP() << "Required queue families not present";
+    }
+
+    // 2 semaphores needed since we need to bind twice before copying
+    auto s_info = LvlInitStruct<VkSemaphoreCreateInfo>();
+    vk_testing::Semaphore semaphore(*m_device, s_info);
```
spencer-lunarg, line 32: (thinking out loud) you think it for `VkSemaphore` it is worth having an overload constructor so you can just go

`vk_testing::Semaphore semaphore(*m_device)` as nothing to ever add to `VkSemaphoreCreateInfo` (unless doing import/export stuff)

spencer-lunarg (approved):
Much nicer, thanks for giving Sparse some :heart: it needed

arno-lunarg:
Hopefully I come with more improvements to sparse validation/testing soon

tests/unit/sparse_buffer_positive.cpp
arno-lunarg: Yep it makes sense, and it would be in line with other `vk_testing::*` objects

PR #6038, opened 2023-06-20
ci: Add SPIRV-Hopper to CI

spencer-lunarg:
Run SPIRV-Hopper on shaders for CI

PR #6349, opened 2023-08-24
tests: Remove SPIRV-Hopper

spencer-lunarg:
I created a new repo for https://github.com/LunarG/SPIRV-Hopper

After spending time I realize having this in VVL was a mistake, reasons:

- The tool is a binary that I want built in Release mode, but want to test the VVL in Debug mode, this forces me to have to build the VVL twice.
- I find times I have a bug in the hopper code AND the VVL code, I need to have 2 branches going to make the two changes.
- https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6038 showed that adding this for CI is not practical per commit and will most likely run it nightly instead.
- Takes up CI resources to add a change that doesn't need to run in CI


tests/README.md
```diff
@@ -18,8 +18,6 @@ The tests are grouped into different categories. Some of the main test categorie
 - Positive testing
     - Make sure Validation isn't accidentally triggering an error
     - Commonly created to prevent bug regressions
-- SPIR-V testing with [SPIRV-Hopper](./spirv_hopper/)
```
arno-lunarg, line 21: Maybe it is still worth mentioning it in this README, with a link to the repo?

arno-lunarg (approved):

juan-lunarg (approved):

tests/README.md
spencer-lunarg: I guess for now, it is not being used and I expect no outside user to worry about it for the time being

PR #6351, opened 2023-08-24
layers: Support per-present-mode capabilities for VK_EXT_surface_maintenance1

HildarTheDorf:
When using VK_EXT_surface_maintenance1 and providing a VkSurfacePresentModeEXT, the values in VkSurfaceCapabilities can change. This can result in a false positive for VUID-VkSwapchainCreateInfoKHR-presentMode-02839. (e.g. on my machine without VkSurfacePresentModeEXT the minImageCount is 4, with a VkSurfacePresentModeEXT specifying FIFO the minImageCount is 2).

When the extension is enabled, always pass a VkSurfacePresentModeEXT to get the most accurate capabilities of the surface.

Question for maintainer: Why is vk_ext_surface_maintenance1 stored in device_extensions when it is an instance_extension?

### DIFF
--- a/layers/core_checks/cc_wsi.cpp
+++ b/layers/core_checks/cc_wsi.cpp
@@ -335,15 +335,15 @@ bool CoreChecks::ValidateCreateSwapchain(VkSwapchainCreateInfoKHR const *pCreate
     }
 
     // Will be assigned &full_screen_info_copy if a VkSurfaceFullScreenExclusiveInfoEXT is found in pCreateInfo->pNext chain
-    const void *full_screen_info_copy_addr = nullptr;
+    void *pnext_copy_addr = nullptr;
 #if defined(VK_USE_PLATFORM_WIN32_KHR)
     auto full_screen_info_copy = LvlInitStruct<VkSurfaceFullScreenExclusiveInfoEXT>();
     auto win32_full_screen_info_copy = LvlInitStruct<VkSurfaceFullScreenExclusiveWin32InfoEXT>();
     const auto *full_screen_info = LvlFindInChain<VkSurfaceFullScreenExclusiveInfoEXT>(pCreateInfo->pNext);
     if (full_screen_info && full_screen_info->fullScreenExclusive == VK_FULL_SCREEN_EXCLUSIVE_APPLICATION_CONTROLLED_EXT) {
-        full_screen_info_copy_addr = &full_screen_info_copy;
         full_screen_info_copy = *full_screen_info;
-        full_screen_info_copy.pNext = nullptr;
+        full_screen_info_copy.pNext = pnext_copy_addr;
+        pnext_copy_addr = &full_screen_info_copy;
 
         if (IsExtEnabled(device_extensions.vk_khr_win32_surface)) {
             const auto *win32_full_screen_info = LvlFindInChain<VkSurfaceFullScreenExclusiveWin32InfoEXT>(pCreateInfo->pNext);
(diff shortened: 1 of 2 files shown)


juan-lunarg (commented):
Looks good overall.

To prevent a regression though a test should be added to `tests/unit/wsi_positive.cpp`

HildarTheDorf:
Excellent. I will write a test tomorrow when I have time.

layers/core_checks/cc_wsi.cpp
```diff
@@ -363,8 +363,16 @@ bool CoreChecks::ValidateCreateSwapchain(VkSwapchainCreateInfoKHR const *pCreate
         }
     }
 #endif
+
+    if (IsExtEnabled(device_extensions.vk_ext_surface_maintenance1)) {
+        auto present_mode_info = LvlInitStruct<VkSurfacePresentModeEXT>();
+        present_mode_info.presentMode = pCreateInfo->presentMode;
+        present_mode_info.pNext = pnext_copy_addr;
+        pnext_copy_addr = &present_mode_info;
```
spencer-lunarg, line 371: `present_mode_info` will fall out of scope after this and the `pNext` will point to garbage, change to

```c++
        auto present_mode_info = LvlInitStruct<VkSurfacePresentModeEXT>();
            if (IsExtEnabled(device_extensions.vk_ext_surface_maintenance1)) {
```

tests/unit/wsi_positive.cpp
```diff
@@ -1187,3 +1187,54 @@ TEST_F(PositiveWsi, ProtectedSwapchainImageColorAttachment) {
...
+    swapchain_create_info.imageArrayLayers = 1;
+    swapchain_create_info.imageUsage = VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT;  // implementations must support
+    swapchain_create_info.imageSharingMode = VK_SHARING_MODE_EXCLUSIVE;
+    swapchain_create_info.preTransform = VK_SURFACE_TRANSFORM_IDENTITY_BIT_KHR;
+    swapchain_create_info.compositeAlpha = m_surface_composite_alpha;
+    swapchain_create_info.presentMode = present_mode;
+    swapchain_create_info.clipped = VK_FALSE;
+    swapchain_create_info.oldSwapchain = 0;
+
+    vk::CreateSwapchainKHR(device(), &swapchain_create_info, nullptr, &m_swapchain);
```
juan-lunarg, line 1239: CI is failing because the CPU mock driver we use is lacking functionality. Given none of the hardware on CI currently support `VK_EXT_SWAPCHAIN_MAINTENANCE_1_EXTENSION_NAME` currently we should fix the mock driver.

```
[ RUN      ] PositiveWsi.CreateSwapchainWithPresentModeInfo
../../tests/framework/error_monitor.cpp:200: Failure
Failed
Validation Error: [ VUID-VkSwapchainPresentModesCreateInfoEXT-pPresentModes-07763 ] | MessageID = 0x76cf26e9 | vkCreateSwapchainKHR(): pCreateInfo.pNext<VkSwapchainPresentModesCreateInfoEXT>.pPresentModes[0] VK_PRESENT_MODE_FIFO_KHR is a non-compatible presentMode. The Vulkan spec states: The entries in pPresentModes must be a subset of the present modes returned in VkSurfacePresentModeCompatibilityEXT::pPresentModes, given VkSwapchainCreateInfoKHR::presentMode in VkSurfacePresentModeEXT (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkSwapchainPresentModesCreateInfoEXT-pPresentModes-07763)
```

spencer-lunarg: I can add the fix to Mock Driver, then re-start CI

spencer-lunarg: Made https://github.com/KhronosGroup/Vulkan-Tools/pull/849 - this test passes with the change, so will restart CI when that PR gets merged!

layers/core_checks/cc_wsi.cpp
```diff
@@ -363,8 +363,15 @@ bool CoreChecks::ValidateCreateSwapchain(VkSwapchainCreateInfoKHR const *pCreate
         }
     }
 #endif
+    auto present_mode_info = LvlInitStruct<VkSurfacePresentModeEXT>();
+    if (IsExtEnabled(device_extensions.vk_ext_surface_maintenance1)) {
+        present_mode_info.presentMode = pCreateInfo->presentMode;
+        present_mode_info.pNext = pnext_copy_addr;
+        pnext_copy_addr = &present_mode_info;
```
spencer-lunarg, line 370: So final concern (but I know nothing about the Win32 Full Screen extension)

But if we would want both the `VkSurfaceFullScreenExclusiveWin32InfoEXT` and `VkSurfacePresentModeEXT` this currently will overwrite the prior one

@arno-lunarg has been working on this dispatch pNext chaining area, might have ideas/opinions

spencer-lunarg (approved):
Not 100% about the `VkSurfaceFullScreenExclusiveWin32InfoEXT`/`VkSurfacePresentModeEXT` interactions, but it lacks a tests and not sure it needs to be solved here

layers/core_checks/cc_wsi.cpp
HildarTheDorf: Is that right? I thought I'd written the code to chain the pnext correctly (so if swapchain_maint1 and win32 full screen are supported, pnext points to the present mode info, and pnext->pnext points to win32 fullscreen info).

Regardless, I have no way to test it myself as my development environment is linux.

arno-lunarg (approved):

PR #6355, opened 2023-08-24
tests: Add vkQueuePresentKHR syncval submit time tests

artem-lunarg:
Negative tests check for PRESENT-AFTER-WRITE hazard when present does not wait on submit.

Positive tests check that presentation engine accesses do not generate any hazards (especially with regard to how swapchain image writes are made visible to the presentation engine).


artem-lunarg:
Going to get this in. I discussed this test setup with John before making PR. If any additional feedback, will be happy to add it later.

PR #6359, opened 2023-08-25
layers: Add remaining RenderArea VUs

spencer-lunarg:
https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5972/diffs adds that Render Area need not to be zero


artem-lunarg (approved):
LGTM!

PR #6364, opened 2023-08-25
layers: Add new 06257 VUIDs from 1.3.262

artem-lunarg:
This adds sync1 version of VUIDs that check acceleration structure reads when ray query is not enabled.
The sync2 version was implemented earlier.

Duplicated functionality for existing sync2 06256 tests was removed (this makes final diff a bit messy, it's easier to review separate commits).


spencer-lunarg (approved):
also nice test cleanup

PR #6365, opened 2023-08-25
tests: Profiles no longer depends on VVL

juan-lunarg:
https://github.com/KhronosGroup/Vulkan-Profiles/pull/487


juan-lunarg:
Thank you @charles-lunarg

spencer-lunarg (approved):

PR #6369, opened 2023-08-25
Syncval Performance Set 1

jzulauf-lunarg:
A series of changes that improved various traces between 1% and 10% reduction in frametime (cumulatively) in local testing.  Each change is in a separate commit.  Relative performance is batched, and relative to the previous batch reported (SB == Strange Brigade  TS == TheSurge)  

This branch is built on the QueueSubmit rework branch (which had a 5%-67% reduction in frametime on one trace all by itself).  

Some expansion on each 

syncval: Cleanup of ResourceAccessState for perf 

Mostly housekeeping change 0's to applicable constants, except for using `reset` for the bitsets vs. `= 0` which was more efficient in the profiling

syncval: Convert usage index to info part 1

Changed from passing around usage information as an index to passing around a pointer to an entry to a const (static) array of usage information structs.  Prevented each ResourceAccessState operation from having to do the array lookup multiple times. During range_map traversal, the info struct is likely staying in memory, as part of the perf improvement.

syncval: Convert read/write to bit lookups 

When we converted from integers to bits sets we kept the bitwise operations paradigm for working with the Stage/Access masks.  When looking for whether a single bit is set, `bitset::hoperator[]` is faster that (bitset_a & bitset_b).any().  So do it everywhere wi

syncval: Convert last InfoType  (7% SB 10% TS)
Storage of the infotype pointer is larger than the enum index, but avoids look ups. 

syncval: Isolate last_write info to class

Something a of a straight clean-up/reorg, but keeps the next step isolated for performance testing

syncval: Convert last_write to optional (1.4% SB 2.8% TS)

It turns out that a lot of ResourceAccessState entries contain read entries (about 98%).  Avoiding the construction/destruction of the (now encapsulated) write state proved a win.

syncval: Convert First Access to InfoType (1% SB 1.5 TS)

Small perf win, maintenance cleanup of the resolve operation

syncval: Small refinements in ApplyBarrier 6.6% SB 6.6 TS)

The code was checking conditions it could know the answers to, and the optimizer wasn't finding it.  Small improvement.

syncval: Cached Insert for tag usage (1% SB 1.2% TS)

Set insertion showed up as a hot spot.  This simple 1-way direct-mapped cache gave a 40% hit rate, saving nearly half of the insertions.  Larger, multi-way, LRU caches had better hit rates, but no better performance.  Probably due to a balance of the (physical) cache residency of the insert cache array.

syncval: Encapsulate pending_write in write state (1% SB)

there are a set of pending_ fields only used when a write is present, this eliminates creation/destruction when no write is present

### DIFF
--- a/layers/sync/sync_validation.cpp
+++ b/layers/sync/sync_validation.cpp
@@ -1013,7 +1013,7 @@ void AccessContext::ResolveAccessRange(const ResourceAccessRange &range, Barrier
         const auto current_range = current->range & range;
         if (current->pos_B->valid) {
             const auto &src_pos = current->pos_B->lower_bound;
-            auto access = src_pos->second;  // intentional copy
+            ResourceAccessState access(src_pos->second);  // intentional copy
             barrier_action(&access);
             if (current->pos_A->valid) {
                 const auto trimmed = sparse_container::split(current->pos_A->lower_bound, *resolve_map, current_range);
@@ -1589,7 +1589,7 @@ HazardResult AccessContext::DetectImageBarrierHazard(const SyncImageMemoryBarrie
 
 template <typename Flags, typename Map>
 SyncStageAccessFlags AccessScopeImpl(Flags flag_mask, const Map &map) {
-    SyncStageAccessFlags scope = 0;
+    SyncStageAccessFlags scope;
     for (const auto &bit_scope : map) {
         if (flag_mask < bit_scope.first) break;
 
@@ -2731,7 +2731,7 @@ static VkImageAspectFlags GetAspectsToClear(VkImageAspectFlags clear_aspect_mask
     }
 
     // Collect aspects that should be cleared.
-    VkImageAspectFlags aspects_to_clear = 0;
+    VkImageAspectFlags aspects_to_clear = VK_IMAGE_ASPECT_NONE;
     if (clear_color && (view_aspect_mask & kColorAspects) != 0) {
         assert(GetBitSetCount(view_aspect_mask) == 1);
         aspects_to_clear |= view_aspect_mask;
@@ -3209,8 +3209,8 @@ HazardResult ResourceAccessState::DetectHazard(SyncStageAccessIndex usage_index,
     HazardResult hazard;
     const auto usage_bit = FlagBit(usage_index);
     const auto usage_stage = PipelineStageBit(usage_index);
-    const bool input_attachment_ordering = (ordering.access_scope & SYNC_FRAGMENT_SHADER_INPUT_ATTACHMENT_READ_BIT).any();
-    const bool last_write_is_ordered = (last_write & ordering.access_scope).any() && (write_queue == queue_id);
+    const bool input_attachment_ordering = ordering.access_scope[SYNC_FRAGMENT_SHADER_INPUT_ATTACHMENT_READ];
+    const bool last_write_is_ordered = (write_queue == queue_id) && (last_write & ordering.access_scope).any();
     if (IsRead(usage_bit)) {
         // Exclude RAW if no write, or write not most "most recent" operation w.r.t. usage;
         bool is_raw_hazard = IsRAWHazard(usage_stage, usage_bit);
@@ -3238,7 +3238,7 @@ HazardResult ResourceAccessState::DetectHazard(SyncStageAccessIndex usage_index,
         const bool usage_write_is_ordered = (usage_bit & ordering.access_scope).any();
         if (last_reads.size()) {
             // Look for any WAR hazards outside the ordered set of stages
-            VkPipelineStageFlags2KHR ordered_stages = 0;
+            VkPipelineStageFlags2KHR ordered_stages = VK_PIPELINE_STAGE_2_NONE;
             if (usage_write_is_ordered) {
                 // If the usage is ordered, we can ignore all ordered read stages w.r.t. WAR)
                 ordered_stages = GetOrderedStages(queue_id, ordering);
@@ -3691,8 +3691,8 @@ void ResourceAccessState::ApplyPendingBarriers(const ResourceUsageTag tag) {
     // We OR in the accumulated write chain and barriers even in the case of a layout transition as SetWrite zeros them.
     write_dependency_chain |= pending_write_dep_chain;
     write_barriers |= pending_write_barriers;
-    pending_write_dep_chain = 0;
-    pending_write_barriers = 0;
+    pending_write_dep_chain = VK_PIPELINE_STAGE_2_NONE;
+    pending_write_barriers.reset();
 }
 
 // Assumes signal queue != wait queue
@@ -3824,26 +3824,28 @@ void ResourceAccessState::OffsetTag(ResourceUsageTag offset) {
     }
 }
 
+static const SyncStageAccessFlags kAllSyncStageAccessBits = ~SyncStageAccessFlags(0);
 ResourceAccessState::ResourceAccessState()
-    : write_barriers(~SyncStageAccessFlags(0)),
-      write_dependency_chain(0),
+    : write_barriers(kAllSyncStageAccessBits),
+      write_dependency_chain(VK_PIPELINE_STAGE_2_NONE),
       write_tag(),
       write_queue(QueueSyncState::kQueueIdInvalid),
-      last_write(0),
+      last_write(),
       input_attachment_read(false),
       last_read_stages(0),
-      read_execution_barriers(0),
-      pending_write_dep_chain(0),
+      read_execution_barriers(VK_PIPELINE_STAGE_2_NONE),
+      last_reads(),
+      pending_write_dep_chain(VK_PIPELINE_STAGE_2_NONE),
       pending_layout_transition(false),
-      pending_write_barriers(0),
+      pending_write_barriers(),
       pending_layout_ordering_(),
       first_accesses_(),
-      first_read_stages_(0U),
+      first_read_stages_(VK_PIPELINE_STAGE_2_NONE),
       first_write_layout_ordering_() {}
 
 // This should be just Bits or Index, but we don't have an invalid state for Index
 VkPipelineStageFlags2KHR ResourceAccessState::GetReadBarriers(const SyncStageAccessFlags &usage_bit) const {
-    VkPipelineStageFlags2KHR barriers = 0U;
+    VkPipelineStageFlags2KHR barriers = VK_PIPELINE_STAGE_2_NONE;
 
     for (const auto &read_access : last_reads) {
         if ((read_access.access & usage_bit).any()) {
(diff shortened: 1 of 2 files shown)


layers/sync/sync_validation.h
```diff
@@ -266,7 +266,41 @@ struct ResourceUsageRecord : public ResourceCmdUsageRecord {
...
+    static constexpr Index kSize = 1 << LogSize;
+    static constexpr key_type kMask = static_cast<key_type>(kSize) - 1;
+    using Entry = std::optional<IntegralKey>;
+
+    void CachedInsert(const key_type &key) {
+        // 1-way direct replacement
+        const Index index = static_cast<Index>(key & kMask);  // Simplest
+        Entry &entry = entries_[index];
+
+        if (!entry || (entry.value() != key)) {
```
artem-lunarg, line 293: nit: perf:  it looks `!entry` condition can be avoid if `entries_` are preinitialized in the constructor with default integral values. Might work without `optional` too.

layers/sync/sync_validation.h
```diff
@@ -687,7 +692,7 @@ class ResourceAccessState : public SyncStageAccess {
     VkPipelineStageFlags2KHR write_dependency_chain;  // intiially zero, but accumulating the dstStages of barriers if they chain.
     ResourceUsageTag write_tag;
     QueueId write_queue;
-    SyncStageAccessFlags last_write;  // only the most recent write
+    const SyncStageAccessInfoType *last_write;  // only the most recent write
```
artem-lunarg: That's more clear then the previous version. `SyncStageAccessFlags` has ability to include multiple bits and it's not obvious just from the definition it's a single access. Previously I had this idea whether we need `SyncStageAccessFlagBit` type and did not realize that `SyncStageAccessInfoType` can be used for this purpose.

layers/sync/sync_validation.h
```diff
@@ -561,8 +565,9 @@ class ResourceAccessState : public SyncStageAccess {
     bool HasPendingState() const {
         return (0 != pending_layout_transition) || pending_write_barriers.any() || (0 != pending_write_dep_chain);
     }
-    bool HasWriteOp() const { return last_write != 0; }
-    const SyncStageAccessFlags &LastWriteOp() const { return last_write; }
+    bool HasWriteOp() const { return last_write != nullptr; }
+    SyncStageAccessIndex LastWriteOp() const { return last_write ? last_write->stage_access_index : SYNC_ACCESS_INDEX_NONE; }
```
artem-lunarg: suggestion (since this PR is about optimization): With this change we have few places that add additional `last_write` (!= nullptr) condition. If `last_write` can regularly be null and not-null this create chances for branch misprediction and a chance to spend a dozen of cycles or so during that unexciting event. One option is to have additional constant that is the address of `syncStageAccessInfoByStageAccessIndex()`'s entry that corresponds to `VK_ACCESS_2_NONE_KHR` and use it in scenarios where nullptr is used now.

layers/sync/sync_validation.h
jzulauf-lunarg: As I don't have an invalid value defined, I'd still need to know if the cache entry was tristate (uninitialized) with an init bool (it could be a bit...),. I'm using optional to do that for me.

artem-lunarg: Ah, I see, unused value is needed to make it work.

jzulauf-lunarg: Turns out I lied.  Since this is a very special purpose cache... I could init to an invalid value.  Hmmm. 
Also I think the insert is more awkward than it needs to be.


```
    // Started the global tag count at 1 s.t. zero are invalid and ResourceUsageTag normalization can just zero them.
    mutable std::atomic<ResourceUsageTag> tag_limit_{1};  // This is reserved in Validation phase, thus mutable and atomic
 ```


`

File: layers/sync/sync_validation.h (hunk 3 above)
Comment (jzulauf-lunarg) [reply layers/sync/sync_validation.h]: The goal is to never initalize the rest of the write related members.  Profiling showed that there are *many* ResourceAccessState created that never are written.

File: layers/sync/sync_validation.cpp [hunk 4]
```diff
@@ -1357,28 +1357,29 @@ void AccessContext::AddAsyncContext(const AccessContext *context, ResourceUsageT
 }
 
 class HazardDetector {
-    SyncStageAccessIndex usage_index_;
+    const SyncStageAccessInfoType &usage_info_;
```
artem-lunarg, line 1358: Due to curiosity checked why it's slow to access array. In a nutshell we started to use pointer (reference) to the required element instead of doing `elem_ptr = base_ptr + index * elem_size`.

Addition is fast.
It turns out `base_ptr` computation is expensive. In release build msvc does not inline call to `syncStageAccessInfoByStageAccessIndex()` but that's not the worst part. Beause this function returns reference to local `const static` array (which is probably good because we had bug in syncval with initialization order of globals) the compiler generates code to make it thread safe + initialization checks, in total it's ~10 instruction including condition and jump (which will be predicted but still). 

Then because sizeof(syncStageAccessInfoByStageAccessIndex) is not power of two `index * elem_size` generates `imul` instead of shift. In total this add around 15 instruction on each access which can take some cycles.

layers/sync/sync_validation.h
artem-lunarg: In case I poorly explained above, here's one example:
```
 const bool last_write_is_ordered =
      last_write && (write_queue == queue_id) && ordering.access_scope[last_write->stage_access_index];
```

`last_write &&` was added to check for null, previously we did not have this. Assuming that `last_write` regularly flips between null/not-null this creates chances for branch misprediction.

layers/sync/sync_validation.cpp
```diff
@@ -3229,12 +3232,13 @@ HazardResult ResourceAccessState::DetectHazard(const SyncStageAccessInfoType &us
                 (input_attachment_ordering && usage_is_input_attachment) || (0 != (usage_stage & ordering.exec_scope));
             if (usage_is_ordered) {
                 // Now see of the most recent write (or a subsequent read) are ordered
-                const bool most_recent_is_ordered = last_write_is_ordered || (0 != GetOrderedStages(queue_id, ordering));
+                const bool most_recent_is_ordered =
+                    last_write.IsOrdered(ordering, queue_id) || (0 != GetOrderedStages(queue_id, ordering));
```
artem-lunarg: should this also check `last_write.HasValue()`?

layers/sync/sync_validation.cpp
jzulauf-lunarg: I remember seeing a function named "thread_saftey_check" or something like that in the profile... and couldn't figure out why it was there. Thanks for finding that.

layers/sync/sync_validation.h
```diff
@@ -424,24 +427,78 @@ struct ResourceFirstAccess {
...
+        exec_scope |= rhs.exec_scope;
+        access_scope |= rhs.access_scope;
+        return *this;
+    }
+    bool operator==(const OrderingBarrier &rhs) const {
+        return (exec_scope == rhs.exec_scope) && (access_scope == rhs.access_scope);
+    }
+};
+
+class ResourceAccessWriteState {
```
artem-lunarg, line 481: One last comment as a follow up to offline discussion. It's a bit surprising that construction/destruction of this class makes effect. What are the performance measurement numbers for improvement from using optional?

layers/sync/sync_validation.cpp
artem-lunarg: In that place I have function `_Init_thread_header`, not sure if it's the same. It's not called from what I see, probably only once during initialization, but it requires to run condition to check that it should not be called.

artem-lunarg (approved):
LGTM. Some nice clean up in addition to performance improvements.

jzulauf-lunarg:
note that [syncval: Remove redunancy in WriteState member names](https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6369/commits/b3a05201f577095b868996f579b461f05f6663c7) is a renaming/rearranging only change.

layers/sync/sync_validation.h
jzulauf-lunarg: Added the relative performance numbers to the description.  The distribution of ReadAccessState contents for the two traces used during development were

```
StrangeBrigade 
Read: 0.985905
ReadWrite: 0.00449582
Write: 0.010181
```
```

TheSurge
Read: 0.986375
ReadWrite: 0.00448118
Write: 0.0136236
```

jzulauf-lunarg:
@artem-lunarg the removal of the aggressive import of the external context in renderpass validation/replay.  Turns out lazy import was more efficient than I had thought.

`current_access_context_->ResolvePreviousAccesses();  `

https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/230249779c315dd9f7f37c7c13a03e158e01fda5#diff-e27ceb3d90bf3f6b578c8e62a52fb3bdc4cd61ec6514c3af52461307b9892861L8159

layers/sync/sync_validation.h
```diff
@@ -514,40 +514,32 @@ class ResourceAccessWriteState {
...
-    }
+    void OffsetTag(ResourceUsageTag offset) { tag_ += offset; }
 
-    void UpdatePendingLayoutOrdering(const SyncBarrier &barrier) {
-        pending_layout_ordering_ |= OrderingBarrier(barrier.src_exec_scope.exec_scope, barrier.src_access_scope);
-    }
+    bool HasPendingState() const { return pending_barriers_.any() || (0 != pending_dep_chain_); }
+    void ClearPending();
+    void UpdatePendingBarriers(const SyncBarrier &barrier);
+    ;
```
artem-lunarg, line 526: nit: ;

artem-lunarg (approved):
LGTM. One note on performance gains. The gains due to complex accesses to array (7%-10%) or helping compiler with branches (6.5%) is something that correlates with typical performance bottlenecks in modern cpu architectures.

It is less clear about performance gains of the order of 1%. My personal opinion such changes are good to have if they are also good even if it's not a performance improvement change but a general refactoring (better code structure or both versions are equally good or addition is modular is easily deletable in case of ungains).

jzulauf-lunarg:
Artem -- the 1% gains... I ran and reran to ensure they persisted outside the noise floor.  They're there.  I even tried making the `pending_*` fields in WriteState optional and saw a persistent 1% decrease (and abandoned it).  Most of the small stuff appears to be avoiding initialization in fields that are only used 1% of the time (the write related fields) -- and mostly when more than one of the large bitsets is involved.  Don't know if it's a cache or instructions effect.  Each of the bitsets is about half a cacheline, and we're suppressing 3 per  ResourceAccessState 99% of the time -- and there are a lot of Construct, copy, and moves of ResourceAccessState.

artem-lunarg:
Thanks for confirmation. Yeah, it's probably me, trying to be cautious if 1% change can make harder to make 10% change. But I don't have objectives to specific commits, looks more cleaner code in most cases (like introduction of combined write state).

Issue #6368, opened 2023-08-25
`VUID-vkCmdDispatch-groupCountX-00386` Should contain shader pipeline name


Jasper-Bekkers:
Currently the error message for this validation layere message only outputs the handle to the command buffer which is relatively useless for pinpointing where this error is coming from. 

We annotate all of our PSOs with names, and to find where in our code this validation layer message triggers it would be extremely helpful to know the name of the PSO when it's available.

<hr>

I tried patching this myself but it seems rather involved: right now this check lives in `manual_PreCallValidateCmdDispatch` in the `StatelessValidation` object which presumably doesn't carry any state (and thus can't locate the PSO easily).

spencer-lunarg:
Love this feedback, the last 2 weeks I have been working on improving the error messages, stuff like this is good to know

I can just move these draw commands to non-stateless and I agree, for draw/dispatch calls having the `VkPipeline` (or now `VkShaderEXT` too) would be helpful

Jasper-Bekkers:
@spencer-lunarg Agreed; and thanks for picking this up. Having the right debugging tools is key!

PR #6373, opened 2023-08-28
layers: Move all Action commands to CoreChecks

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6368

This adds a `GetObjectList` to the command buffer state object. The goal is for "Action" commands (draw, dispatch, traceRays) that we actually mark the `VkPipeline`/`VkShaderEXT` it came from

This PR moves all the checks from stateless to `CoreCheck`, it also organizes all the calls to be unified in the order of things


layers/state_tracker/cmd_buffer_state.cpp
```diff
@@ -1541,3 +1541,72 @@ void CMD_BUFFER_STATE::UnbindResources() {
...
+void CMD_BUFFER_STATE::GetCurrentPipelineAndDesriptorSets(VkPipelineBindPoint pipelineBindPoint, const PIPELINE_STATE **rtn_pipe,
+                                                          const std::vector<LAST_BOUND_STATE::PER_SET> **rtn_sets) const {
+    const auto lv_bind_point = ConvertToLvlBindPoint(pipelineBindPoint);
+    const auto &last_bound = lastBound[lv_bind_point];
+    if (!last_bound.IsUsing()) {
+        return;
+    }
+    *rtn_pipe = last_bound.pipeline_state;
+    *rtn_sets = &(last_bound.per_set);
+}
```
arno-lunarg, line 1612: For consistency, I would return a `std::pair` from this function:
```suggestion
std::pair<const PIPELINE_STATE *, const std::vector<LAST_BOUND_STATE::PER_SET> *> CMD_BUFFER_STATE::GetCurrentPipelineAndDesriptorSets(VkPipelineBindPoint pipelineBindPoint) const {
    const auto lv_bind_point = ConvertToLvlBindPoint(pipelineBindPoint);
    const auto &last_bound = lastBound[lv_bind_point];
    if (!last_bound.IsUsing()) {
        return {nullptr, nullptr};
    }
	return {last_bound.pipeline_state, &last_bound.per_set};
}
```

tests/unit/mesh.cpp
```diff
@@ -683,12 +683,17 @@ TEST_F(NegativeMesh, BasicUsageNV) {
     ASSERT_VK_SUCCESS(result);
 
     m_commandBuffer->begin();
+    m_commandBuffer->BeginRenderPass(m_renderPassBeginInfo);
 
+    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-vkCmdDrawMeshTasksIndirectNV-None-08606");
```
arno-lunarg, line 688: Just to know, why do we have those new VUIDs being emitted in the tests?

arno-lunarg (approved):
LGTM!
Not sure about the motivation behind moving a lot of things from stateless to core checks? Is it mandatory because of the introduced changes?

spencer-lunarg:
> Not sure about the motivation behind moving a lot of things from stateless to core checks?

We need to be able to print the last bound `VkPipeline`/`VkShaderEXT` which we don't have with stateless

tests/unit/mesh.cpp
spencer-lunarg: If we hit a VU in Stateless it stops and never goes into CoreChecks

Before, it hit the VU below this line and stopped, now we hit multiple VUs, the "real" fix is improving the tests in the long term

layers/state_tracker/cmd_buffer_state.cpp
spencer-lunarg: I just moved this from the Header to Source file, didn't look at it, but agree, can fix it up

arno-lunarg: Ah I was thinking this was new!

PR #6374, opened 2023-08-28
layers: Simplify BINDABLE class hierarchy

jeremyg-lunarg:
Make BindableMemoryTracker be a member of each state object instead of the 'leaf' of the inheritance hierarchy.

All of the trackers are now less than 64 bytes in size so they can be stored directly in a variant without excessive overhead.

I'm about to add subclasses of BUFFER_STATE and I don't want to go through the whole ValidationStateTracker::CreateImageStateImpl() mess again.

### DIFF
--- a/layers/state_tracker/buffer_state.cpp
+++ b/layers/state_tracker/buffer_state.cpp
@@ -44,8 +44,16 @@ BUFFER_STATE::BUFFER_STATE(ValidationStateTracker *dev_data, VkBuffer buff, cons
       createInfo(*safe_create_info.ptr()),
       requirements(GetMemoryRequirements(dev_data, buff)),
       usage(GetBufferUsageFlags(createInfo)),
-      supported_video_profiles(
-          dev_data->video_profile_cache_.Get(dev_data, LvlFindInChain<VkVideoProfileListInfoKHR>(pCreateInfo->pNext))) {}
+      supported_video_profiles(dev_data->video_profile_cache_.Get(dev_data, LvlFindInChain<VkVideoProfileListInfoKHR>(pCreateInfo->pNext))) {
+    if (pCreateInfo->flags & VK_BUFFER_CREATE_SPARSE_BINDING_BIT) {
+        tracker_.emplace<BindableSparseMemoryTracker>(&requirements,
+                                                      (pCreateInfo->flags & VK_BUFFER_CREATE_SPARSE_RESIDENCY_BIT) != 0);
+        SetMemoryTracker(&std::get<BindableSparseMemoryTracker>(tracker_));
+    } else {
+        tracker_.emplace<BindableLinearMemoryTracker>(&requirements);
+        SetMemoryTracker(&std::get<BindableLinearMemoryTracker>(tracker_));
+    }
+}
 
 #ifdef VK_USE_PLATFORM_METAL_EXT
 static bool GetMetalExport(const VkBufferViewCreateInfo *info) {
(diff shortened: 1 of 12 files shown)


layers/state_tracker/buffer_state.cpp
```diff
@@ -44,8 +44,21 @@ BUFFER_STATE::BUFFER_STATE(ValidationStateTracker *dev_data, VkBuffer buff, cons
       createInfo(*safe_create_info.ptr()),
       requirements(GetMemoryRequirements(dev_data, buff)),
       usage(GetBufferUsageFlags(createInfo)),
-      supported_video_profiles(
-          dev_data->video_profile_cache_.Get(dev_data, LvlFindInChain<VkVideoProfileListInfoKHR>(pCreateInfo->pNext))) {}
+      supported_video_profiles(dev_data->video_profile_cache_.Get(dev_data, LvlFindInChain<VkVideoProfileListInfoKHR>(pCreateInfo->pNext))) {
+    if (pCreateInfo->flags & VK_BUFFER_CREATE_SPARSE_BINDING_BIT) {
+        std::unique_ptr<BindableMemoryTracker> ptr;
+        if (pCreateInfo->flags & VK_BUFFER_CREATE_SPARSE_RESIDENCY_BIT) {
+            ptr.reset(new BindableSparseMemoryTracker<true>(&requirements));
```
artem-lunarg, line 51: suggestion to use `std::make_unique`

layers/state_tracker/buffer_state.h
```diff
@@ -54,11 +54,12 @@ class BUFFER_STATE : public BINDABLE {
...
     sparse_container::range<VkDeviceAddress> DeviceAddressRange() const { return {deviceAddress, deviceAddress + createInfo.size}; }
-};
 
-using BUFFER_STATE_LINEAR = MEMORY_TRACKED_RESOURCE_STATE<BUFFER_STATE, BindableLinearMemoryTracker>;
-template <bool IS_RESIDENT>
-using BUFFER_STATE_SPARSE = MEMORY_TRACKED_RESOURCE_STATE<BUFFER_STATE, BindableSparseMemoryTracker<IS_RESIDENT>>;
+  private:
+    // Store the linear tracker inline since it is small and the common case.
+    // The sparse are allocated separately behind a unique_ptr<> to avoid bloating.
+    std::variant<std::unique_ptr<BindableMemoryTracker>, BindableLinearMemoryTracker> tracker_;
```
artem-lunarg, line 61: For msvc debug build sparse tracker is 48 bytes and linear tracker is 40 bytes.
In release build both linear and sparse trackers are 40 bytes.

One option is to use sparse type directly. One bonus that code is more specific, base class provides more options. Not sure how it's on other compilers, also tried to put these to types into variant and this resulted in compilation error, not sure what's going on :)

For images multiplane tracker with 3 planes is much bigger (128 bytes).

jeremyg-lunarg: good to know. thanks for looking into that.

layers/state_tracker/buffer_state.cpp
jeremyg-lunarg: just using inline objects now

layers/state_tracker/buffer_state.h
jeremyg-lunarg: your compile error was because you need something in the variant that is default constructable. std:monostate can be used for that.  I restructured multiplanar so that it is 56 bytes and now it is reasonable to have everything be inline.

artem-lunarg: thanks, great clean up

artem-lunarg (approved):
LGTM

PR #6379, opened 2023-08-29
layers: Add RecordObj for PostCallRecord

spencer-lunarg:
Adds a `RecordObj` to all `PostCallRecord` to allow a way to pass information around in an automated way in the future (such as function name things came from in order to track better error messages)

We already had the `VkResult` passed in some `PostCallRecord` so figured this was a good first move, a future PR would add this as well to `PreCallRecord`


layers/error_message/record_object.h
```diff
@@ -0,0 +1,31 @@
...
+
+// Contains all information for each PostCallRecord function
+struct RecordObject {
+    const Location location; // starting location (Always the function entrypoint)
+    const VkResult result = VK_RESULT_MAX_ENUM; // Not all items return a VkResult
+    const VkDeviceAddress device_address = 0;
+
+    RecordObject(vvl::Func command_) : location(Location(command_)) {}
+    RecordObject(vvl::Func command_, VkResult result_) : location(Location(command_)), result(result_) {}
+    RecordObject(vvl::Func command_, VkDeviceAddress device_address_) : location(Location(command_)), device_address(device_address_) {}
```
arno-lunarg, line 28: I wonder if calls like:
```cpp
RecordObject foo(Func::vkBeginCommandBuffer, 42);
```
are ambiguous, not sure which constructor overload would be selected here, C enums being named integers maybe the `VkResult` would be picked, when I think we expect the `VkDeviceAddress` one?

spencer-lunarg: I guess part of this assumes no one is ever going to construct the `RecordObject` outside the generated chassis code (with very minimal exceptions)

It is a fair concern, but also I hope we aren't using magic number literals lol

PR #6380, opened 2023-08-29
layers: Add barrier queue family VUs 

artem-lunarg:
This adds support for 03851 03854 VUs.
Some minor clean up.


layers/core_checks/core_validation.h
```diff
@@ -567,11 +567,13 @@ class CoreChecks : public ValidationStateTracker {
                               const ErrorObject& errorObj) const;
     template <typename Barrier>
     bool ValidateBufferBarrier(const LogObjectList& objlist, const Location& loc, const CMD_BUFFER_STATE* cb_state,
```
artem-lunarg, line 569: Can't say I like these templated methods. The problem it works as a lowest common denominator. One example, if we need some field that exists in Sync2 but not in Sync1, then that field should be added as parameter or we need to introduce overload for that case. Personally I prefer "new is good, old survives" strategy (not universally, but for this case). Example of this can be non-templated method that accepts Sync2 specific type, and Sync1 calls convert to Sync2 type. I can try that change in a separate PR if we want to try this direction.

layers/core_checks/cc_synchronization.cpp
```diff
@@ -2147,41 +2169,56 @@ bool CoreChecks::ValidateConcurrentBarrierAtSubmit(const Location &loc, const Va
 
 // Type specific wrapper for image barriers
 template <typename ImgBarrier>
-bool CoreChecks::ValidateBarrierQueueFamilies(const Location &loc, const CMD_BUFFER_STATE *cb_state, const ImgBarrier &barrier,
+bool CoreChecks::ValidateBarrierQueueFamilies(const LogObjectList &objects, const Location &barrier_loc, const ImgBarrier &barrier,
+                                              VkPipelineStageFlags2 src_stage_mask, VkPipelineStageFlags2 dst_stage_mask,
                                               const IMAGE_STATE *state_data) const {
-    // State data is required
     if (!state_data) {
```
arno-lunarg, line 2175: Comment was removed, is it ok?

layers/core_checks/cc_synchronization.cpp
```diff
@@ -2134,6 +2134,28 @@ bool Validate(const CoreChecks *device_data, const CMD_BUFFER_STATE *cb_state, c
 
     return skip;
 }
+
+bool ValidateHostStage(const ValidationObject *validation_obj, const LogObjectList &objects, const Location &barrier_loc,
```
arno-lunarg, line 2138:
```suggestion
static bool ValidateHostStage(const ValidationObject *validation_obj, const LogObjectList &objects, const Location &barrier_loc,
```

arno-lunarg (approved):
LGTM! 1 nit and 1 question

layers/core_checks/cc_synchronization.cpp
artem-lunarg: I think that's no different from other null checks, later in the code we need to access this pointer unconditionally, so it should be valid.

layers/core_checks/cc_synchronization.cpp
artem-lunarg: Fixed

layers/core_checks/cc_synchronization.cpp
jeremyg-lunarg: could the parameter be changed to & instead of *?

layers/core_checks/core_validation.h
jeremyg-lunarg: yeah if you have a way to make this stuff cleaner, give it a try.

layers/core_checks/cc_synchronization.cpp
artem-lunarg: Good point. Caller also does the null check.

PR #6383, opened 2023-08-29
layers: Fix pnext chain handling for GetSurfaceCapabilities in swapchain creation validation

arno-lunarg:
Follow up to https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6351

Fix a case where part of the pnext chain passed down to `GetSurfaceCapabilities` could be lost.
Add a new positive test that takes the problematic path.

### DIFF
--- a/layers/core_checks/cc_wsi.cpp
+++ b/layers/core_checks/cc_wsi.cpp
@@ -334,16 +334,15 @@ bool CoreChecks::ValidateCreateSwapchain(VkSwapchainCreateInfoKHR const *pCreate
         }
     }
 
-    // Will be assigned &full_screen_info_copy if a VkSurfaceFullScreenExclusiveInfoEXT is found in pCreateInfo->pNext chain
-    void *pnext_copy_addr = nullptr;
+    void *surface_caps_query_pnext = nullptr;
 #if defined(VK_USE_PLATFORM_WIN32_KHR)
     auto full_screen_info_copy = LvlInitStruct<VkSurfaceFullScreenExclusiveInfoEXT>();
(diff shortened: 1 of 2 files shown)


layers/core_checks/cc_wsi.cpp
```diff
@@ -357,21 +356,21 @@ bool CoreChecks::ValidateCreateSwapchain(VkSwapchainCreateInfoKHR const *pCreate
                 }
             } else {
                 win32_full_screen_info_copy = *win32_full_screen_info;
-                win32_full_screen_info_copy.pNext = nullptr;
-                full_screen_info_copy.pNext = &win32_full_screen_info_copy;
+                win32_full_screen_info_copy.pNext = surface_caps_query_pnext;
+                surface_caps_query_pnext = &win32_full_screen_info_copy;
```
juan-lunarg, line 360: Took me a sec to spot the fix here.

juan-lunarg (approved):

tests/unit/wsi_positive.cpp
```diff
@@ -810,6 +811,61 @@ TEST_F(PositiveWsi, CreateSwapchainFullscreenExclusive) {
 }
 #endif
 
+#if defined(VK_USE_PLATFORM_WIN32_KHR)
```
spencer-lunarg, line 814: (for whoever ends up doing larger WSI cleanup) we should just make a `class PositiveWsiWin32 : PositiveWsi` and try splitting our WSI tests up

spencer-lunarg (approved):

layers/core_checks/cc_wsi.cpp
arno-lunarg: Sorry should have made a github comment, I did not realise that those 2 lines were buried in the name change...

PR #6387, opened 2023-08-30
layers: First move towards Stateless using Location

spencer-lunarg:
My goal is to remove `ParameterName` for `Location` but there is a lot of deep code to workout, so this first PR is to just get the `Location` inside all the functions alongside and remove it after

So with this, all the generated Stateless checks only use `Location` to pass the function name, but will soon just become a full replacement


scripts/generators/stateless_validation_helper_generator.py
```diff
@@ -433,17 +433,19 @@ def generateSource(self):
...
                 prototype = command.cPrototype[:-1].split('\n')
                 prototype = '\n'.join(prototype)
                 prototype += ' const {\n'
                 prototype = prototype.split('VKAPI_CALL vk')[1]
-                prototype = prototype.replace(')', ',\n    const ErrorObject&                         error_obj)')
+                prototype = prototype.replace(')', ',\n    const ErrorObject&                          error_obj)')
                 out.append('bool StatelessValidation::PreCallValidate' + prototype)
                 out.append(f'{indent}bool skip = false;\n')
+                # Create a copy here to make the logic simpler passing into ValidatePnextStructContents
+                out.append(f'{indent}const Location& loc = error_obj.location;\n')
```
arno-lunarg, line 446:
```suggestion
                out.append(f'{indent}[[maybe_unused]] const Location& loc = error_obj.location;\n')
```

arno-lunarg (approved):

scripts/generators/stateless_validation_helper_generator.py
spencer-lunarg: much better idea then adding logic to detect if it will ever be used in the scripts (that would not be fun python)

PR #6389, opened 2023-08-30
layers: Use Location for VK_EXT_shader_object

spencer-lunarg:
adds `Location` to errors for `VK_EXT_shader_object` (cc @ziga-lunarg)

### DIFF
--- a/layers/core_checks/cc_shader_object.cpp
+++ b/layers/core_checks/cc_shader_object.cpp
@@ -67,7 +67,7 @@ bool CoreChecks::PreCallValidateCreateShadersEXT(VkDevice device, uint32_t creat
 
     if (enabled_features.shader_object_features.shaderObject == VK_FALSE) {
         skip |=
-            LogError(device, "VUID-vkCreateShadersEXT-None-08400", "vkCreateShadersEXT(): shaderObject feature is not enabled.");
+            LogError("VUID-vkCreateShadersEXT-None-08400", device, error_obj.location, "the shaderObject feature was not enabled.");
     }
 
     const uint32_t invalid = createInfoCount;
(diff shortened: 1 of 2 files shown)


layers/core_checks/cc_shader_object.cpp
```diff
@@ -81,75 +81,65 @@ bool CoreChecks::PreCallValidateCreateShadersEXT(VkDevice device, uint32_t creat
     uint32_t linked_spirv_index = invalid;
     uint32_t linked_binary_index = invalid;
     for (uint32_t i = 0; i < createInfoCount; ++i) {
+        const Location loc = error_obj.location.dot(Field::pCreateInfos, i);
```
arno-lunarg, line 84: Personal nit: would like a rename, `loc` is a bit vague. Something like `create_info_i` or `create_info_i_loc`?

layers/stateless/sl_shader_object.cpp
```diff
@@ -23,29 +23,23 @@ bool StatelessValidation::manual_PreCallValidateCreateShadersEXT(VkDevice device
     bool skip = false;
 
     for (uint32_t i = 0; i < createInfoCount; ++i) {
+        const Location loc = error_obj.location.dot(Field::pCreateInfos, i);
```
arno-lunarg, line 26: Same renaming here

arno-lunarg (approved):
LGTM! 1 nit

layers/core_checks/cc_shader_object.cpp
spencer-lunarg: I have done `info_loc` in the other spots, but `create_info_loc` seems good to me (main thing to have `_loc` as the prefix for consistency`

I guess I don't see the reason for the `_i_` because the idea is people don't need to worry if it is or is not an indexed field while printing, just what field they are in

spencer-lunarg: ok, there are 34 other instances of this, let me just make a PR right now fixing this

PR #6390, opened 2023-08-30
Remove more const char for Func

spencer-lunarg:
Some misc spots where we use `const char*` to pass the function that can be done with the new `vvl::Func`

currently can use `RecordObject` for the `PostCallRecord`, but for the `PreCallRecord`, still use the `vvl::Func` by hand


layers/vulkan/generated/thread_safety.cpp
```diff
@@ -53,7 +53,7 @@ void ThreadSafety::PostCallRecordDestroyInstance(
     VkInstance                                  instance,
     const VkAllocationCallbacks*                pAllocator,
     const RecordObject&                         record_obj) {
-    FinishWriteObjectParentInstance(instance, "vkDestroyInstance");
+    FinishWriteObjectParentInstance(instance, record_obj.location.function);
```
artem-lunarg, line 56: I could miss details from the previous PRs, but why do we use variable instead of constant? Can it be different than `vvl::Func::vkDestroyInstance`?

artem-lunarg: Actually I have no strong opinion about this. Using variable will probably touch memory, constant can be optimized (be part of instruction if lucky), but that's pessimization only under other equal conditions. General design is more important, but at least we can have a mental note that it's possible to do for specific places if some performance work will need this.

artem-lunarg (approved):
LGTM. Left some thoughts about constants.

layers/vulkan/generated/thread_safety.cpp
spencer-lunarg: ok, so myself thinking about it, the reason I wanted to use the variable was to avoid human errors (I found a few `const char* api_name` that were mis-copy-and-pasted)

but since this is generated, I see no reason to just use the constant, going to update

PR #6393, opened 2023-08-30
corechecks: Fix perf regression from shader object validation

jeremyg-lunarg:
FormatHandle() creates a string which adds malloc() overhead in per-draw call code paths. Don't do this unless an error is detected.

This fixes a perf regression from https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6291

### DIFF
--- a/layers/core_checks/cc_descriptor.cpp
+++ b/layers/core_checks/cc_descriptor.cpp
@@ -139,12 +139,12 @@ bool CoreChecks::VerifySetLayoutCompatibility(const DescriptorSetLayout &layout_
 // pipelineLayout[layoutIndex]
 bool CoreChecks::VerifySetLayoutCompatibility(
     const cvdescriptorset::DescriptorSet &descriptor_set,
-    const std::vector<std::shared_ptr<cvdescriptorset::DescriptorSetLayout const>> &set_layouts, const std::string &handle,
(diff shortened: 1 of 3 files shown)


ziga-lunarg (approved):

juan-lunarg (approved):

Tony-LunarG (approved):

jeremyg-lunarg:
nv win32 error in internal CI:
```
[ RUN      ] PositiveShaderObject.DrawMinimalDynamicStatesRasterizationDisabled
C:\j\vz2\Release32\Vulkan-ValidationLayers\tests\framework\error_monitor.cpp(200): error: Failed
Validation Error: [ VUID-vkCmdDraw-None-08617 ] Object 0: handle = 0x1f6f9ca0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x4c85677 | vkCmdDraw():  VK_DYNAMIC_STATE_LINE_WIDTH state not set for this command buffer. The Vulkan spec states: If a shader object is bound to any graphics stage, and the most recent call to vkCmdSetPolygonModeEXT in the current command buffer set polygonMode to VK_POLYGON_MODE_LINE, vkCmdSetLineWidth must have been called in the current command buffer prior to this drawing command (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdDraw-None-08617)
```

spencer-lunarg (approved):

PR #6395, opened 2023-08-30
vscode: Don't use CMakePresets.json workflow

juan-lunarg:
Our repo doesn't work well with the CMakePresets.json workflow vscode encourages, and can be actively confusing/hindering to new contributors.

We actually discussed this a while ago internally.

Provide this default settings.json so vscode uses the default CMake workflow.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "cmake.useCMakePresets": "never"
+}


juan-lunarg:
> How will this affect repo's that already _have_ .vscode/settings.json's?

It will overwrite the existing settings.json currently as is.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "cmake.useCMakePresets": "never"
+}
```
juan-lunarg, line 3: My goal is to share this setting among users. Not sure how to do this elegantly / idiomatically.

charles-lunarg:
> It will overwrite the existing settings.json currently as is.

Thats quite annoying, since I have a lot of settings set per-repo.

juan-lunarg:
> Thats quite annoying, since I have a lot of settings set per-repo.

Agreed. Closing for now. Might be better for everyone to just remove CMakePresets.json from this repo.

juan-lunarg:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6396

PR #6396, opened 2023-08-30
cmake: Remove CMakePresets.json

juan-lunarg:
- Makes VVL consistent with the other repos we maintain
- Fixes vscode issue for new contributors
- Simplifies build instructions

Note: Users who like this workflow are encourage to use CMakeUserPresets.json


BUILD.md
```diff
@@ -151,7 +138,7 @@ Run CMake to generate [Visual Studio project files](https://cmake.org/cmake/help
 
 ```bash
 # NOTE: By default CMake picks the latest version of Visual Studio as the default generator.
-cmake -S . -B build --preset dev
+cmake -S . -B build ...
```
charles-lunarg: why not remove the `...`?

juan-lunarg: 👍🏾

charles-lunarg (approved):

PR #6397, opened 2023-08-30
android: Remove dependency on strings.xml

juan-lunarg:
I'll remove strings.xml in a follow up PR since it will require a corresponding change in VulkanTests

### DIFF
--- a/scripts/android.py
+++ b/scripts/android.py
@@ -35,14 +35,6 @@ def get_android_manifest() -> str:
         sys.exit(-1)
     return manifest
 
-# Resources for our test application.
-def get_android_resources() -> str:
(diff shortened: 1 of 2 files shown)


juan-lunarg:
FWIW I already fixed/tested this in the vulkan extension layers:
https://github.com/KhronosGroup/Vulkan-ExtensionLayer/commit/851c147e1e87339e08706ea0c4a0298d7be02ad0

jeremyg-lunarg (approved):

spencer-lunarg (approved):
the `strings.xml` file is just the default way to build APKs so "normal" app devs have good practices, agree it was an unnecessary file we just had

PR #6399, opened 2023-08-31
layers: Use Location for Update Descriptor checks

spencer-lunarg:
The Update Descriptor logic is quite old and is nothing like any other part of the code base, it used this `std::string *error_code, std::string *error_msg` system which was awful.

This change doesn't change any of the logic (except a few places it was wrong actually) and mainly just redid all the `LogError` to use the new `Location` 

The new error message provide better Object handle and info then before, **but** I do plan to come back through as there is a LOT that could still be cleaned up about this code

### DIFF
--- a/layers/core_checks/cc_descriptor.cpp
+++ b/layers/core_checks/cc_descriptor.cpp
@@ -1635,14 +1635,13 @@ bool CoreChecks::ValidateDescriptor(const DescriptorContext &context, const Desc
 // Starting at offset descriptor of given binding, parse over update_count
 //  descriptor updates and verify that for any binding boundaries that are crossed, the next binding(s) are all consistent
 //  Consistency means that their type, stage flags, and whether or not they use immutable samplers matches
-//  If so, return true. If not, fill in error_msg and return false
-static bool VerifyUpdateConsistency(debug_report_data *report_data, const DescriptorSet &set, uint32_t binding, uint32_t offset,
-                                    uint32_t update_count, const char *type, std::string *error_msg) {
+bool CoreChecks::VerifyUpdateConsistency(const DescriptorSet &set, uint32_t binding, uint32_t offset, uint32_t update_count,
+                                         const char *vuid, const Location &set_loc) const {
+    bool skip = false;
     auto current_iter = set.FindBinding(binding);
-    bool pass = true;
     // Verify consecutive bindings match (if needed)
     auto &orig_binding = **current_iter;
-    while (pass && update_count) {
+    while (!skip && update_count) {
         // First, it's legal to offset beyond your own binding so handle that case
         if (offset > 0) {
             // index_range.start + offset is which descriptor is needed to update. If it > index_range.end, it means the descriptor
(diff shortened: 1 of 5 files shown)


layers/core_checks/cc_descriptor.cpp
```diff
@@ -3009,127 +2715,98 @@ bool CoreChecks::ValidateWriteUpdate(const DescriptorSet *dest_set, const VkWrit
...
-                                        << "Attempted write update to combined image sampler and image view and sampler ycbcr "
-                                           "conversions are not identical, sampler: "
-                                        << FormatHandle(desc.GetSampler())
-                                        << " image view: " << FormatHandle(iv_state->image_view()) << ".";
-                                    *error_msg = error_str.str();
-                                    return false;
+                                    const LogObjectList objlist(update.dstSet, desc.GetSampler(), iv_state->image_view());
+                                    skip |= LogError("VUID-VkWriteDescriptorSet-descriptorType-01948", objlist, write_loc,
+                                                     "Attempted write update to combined image sampler and image view and sampler "
+                                                     "ycbcr conversions are not identical.");
```
arno-lunarg, line 2782:
```suggestion
                                                     "YCbCr conversions are not identical.");
```

layers/core_checks/cc_descriptor.cpp
```diff
@@ -3009,127 +2715,98 @@ bool CoreChecks::ValidateWriteUpdate(const DescriptorSet *dest_set, const VkWrit
...
-                                          << ") that includes a YCBCR conversion, it must have been allocated with a layout that "
-                                             "includes an immutable sampler.";
-                                *error_msg = error_str.str();
-                                return false;
-                            }
+                        } else if (iv_state && (iv_state->samplerConversion != VK_NULL_HANDLE)) {
+                            const LogObjectList objlist(update.dstSet, iv_state->image_view());
+                            skip |=
+                                LogError("VUID-VkWriteDescriptorSet-descriptorType-02738", objlist, write_loc.dot(Field::dstSet),
+                                         "is bound to image view that includes a YCBCR conversion, it must have been allocated "
```
arno-lunarg, line 2789:
```suggestion
                                         "is bound to image view that includes a YCbCr conversion, it must have been allocated "
```

arno-lunarg (approved):
LGTM!

layers/core_checks/cc_descriptor.cpp
```diff
@@ -2016,219 +1955,158 @@ bool CoreChecks::ValidateImageUpdate(VkImageView image_view, VkImageLayout image
...
-                          << FormatHandle(image) << " in imageView " << FormatHandle(image_view)
-                          << ". Allowed layouts are: VK_IMAGE_LAYOUT_GENERAL";
-                if (IsExtEnabled(device_extensions.vk_khr_shared_presentable_image)) {
-                    error_str << " or VK_IMAGE_LAYOUT_SHARED_PRESENT_KHR";
-                }
-                *error_msg = error_str.str();
-                return false;
+                skip |= LogError("VUID-VkWriteDescriptorSet-descriptorType-04152", objlist, image_info_loc.dot(Field::imageView),
+                                 "image layout is %s, but descriptorType is VK_DESCRIPTOR_TYPE_STORAGE_IMAGE. (allowed layouts are "
+                                 "GENERAL or SHARED_PRESENT).",
```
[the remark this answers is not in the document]
spencer-lunarg: agree... honestly this whole section needs a fix as these VUs point to other parts of the spec that list the valid items so we should probably list them... for now will just take your suggestion here

Issue #4388, opened 2022-08-08
Missing validation for simultaneous use of parameters for vkQueuePresentKHR


martty:
**Describe the situation in which you encountered the missing validation**
Issue description: vkQueuePresentKHR has some externally synchronized parameters, most importantly the swapchain, which is not validated.

**Valid Usage IDs requested**
Please include the valid usage IDs for the checks you are requesting: external synchronization

**Additional Context**
Any additional information:

sjfricke:
all the externally synchronized parameters should be done in `layers/generated/thread_safty.cpp` but looking at the spec there are a few parameters that are arrays (`pWaitSemaphores` and `pSwapchains` for `vkQueuePresentKHR`)

`vkResetFences`, `vkFreeDescriptorSets`, `vkFreeCommandBuffers`, and `vkCreateSharedSwapchainsKHR` seem to be doing it correctly for their array parameters, but there is nothing for `vkQueuePresentKHR` at all as you pointed out... instresting...

So the fix will be to add a special case for `vkQueuePresentKHR` into `thread_safty_generator.py`

----

Spec language for reference

> Each element of the pWaitSemaphores member of the pPresentInfo parameter in vkQueuePresentKHR
> 
> Each element of the pSwapchains member of the pPresentInfo parameter in vkQueuePresentKHR

artem-lunarg:
Fixed in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6400

PR #6400, opened 2023-08-31
layers: Add thread safety checks for vkQueuePresentKHR

AlinaKalyakina:
Add manual Pre- and Post- for thread safety checks of `vkQueuePresentKHR`. Regenerated files.

### DIFF
--- a/layers/thread_tracker/thread_safety_validation.cpp
+++ b/layers/thread_tracker/thread_safety_validation.cpp
@@ -740,3 +740,34 @@ void ThreadSafety::PostCallRecordCreateRayTracingPipelinesKHR(VkDevice device, V
         }
     }
 }
+
+void ThreadSafety::PreCallRecordQueuePresentKHR(VkQueue queue, const VkPresentInfoKHR* pPresentInfo) {
+    StartWriteObject(queue, vvl::Func::vkQueuePresentKHR);
+    uint32_t waitSemaphoreCount = pPresentInfo->waitSemaphoreCount;
+    if (pPresentInfo->pWaitSemaphores != nullptr) {
+        for (uint32_t index = 0; index < waitSemaphoreCount; index++) {
+            StartReadObject(pPresentInfo->pWaitSemaphores[index], vvl::Func::vkQueuePresentKHR);
+        }
+    }
+    if (pPresentInfo->pSwapchains != nullptr) {
+        for (uint32_t index = 0; index < pPresentInfo->swapchainCount; ++index) {
+            StartWriteObjectParentInstance(pPresentInfo->pSwapchains[index], vvl::Func::vkQueuePresentKHR);
+        }
+    }
+}
+
+void ThreadSafety::PostCallRecordQueuePresentKHR(VkQueue queue, const VkPresentInfoKHR* pPresentInfo, 
+                                                 const RecordObject& record_obj) {
+    FinishWriteObject(queue, record_obj.location.function);
+    uint32_t waitSemaphoreCount = pPresentInfo->waitSemaphoreCount;
+    if (pPresentInfo->pWaitSemaphores != nullptr) {
+        for (uint32_t index = 0; index < waitSemaphoreCount; index++) {
+            FinishReadObject(pPresentInfo->pWaitSemaphores[index], record_obj.location.function);
+        }
+    }
+    if (pPresentInfo->pSwapchains != nullptr) {
+        for (uint32_t index = 0; index < pPresentInfo->swapchainCount; ++index) {
+            FinishWriteObjectParentInstance(pPresentInfo->pSwapchains[index], record_obj.location.function);
+        }
+    }
+}
--- a/layers/vulkan/generated/thread_safety_commands.h
+++ b/layers/vulkan/generated/thread_safety_commands.h
@@ -2479,6 +2479,15 @@ void PostCallRecordAcquireNextImageKHR(
     uint32_t*                                   pImageIndex,
     const RecordObject&                         record_obj) override;
 
+void PreCallRecordQueuePresentKHR(
+    VkQueue                                     queue,
+    const VkPresentInfoKHR*                     pPresentInfo) override;
+
+void PostCallRecordQueuePresentKHR(
+    VkQueue                                     queue,
+    const VkPresentInfoKHR*                     pPresentInfo,
+    const RecordObject&                         record_obj) override;
+
 void PreCallRecordGetDeviceGroupPresentCapabilitiesKHR(
     VkDevice                                    device,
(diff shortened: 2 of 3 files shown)


AlinaKalyakina:
Related #[4388](https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4388)

spencer-lunarg (approved):
looking back at my comment in https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4388#issuecomment-1207807524

this seems correct, the `thread_safety_generator.py` was recently refactored and I guess a little surprised  the fix was removing from the blacklist, but also happy to know things "just work" now in the script

This looks good to me, but nice if either @jeremyg-lunarg @artem-lunarg or @jzulauf-lunarg took a second look as I'm not an expert with the thread tracker overall picture

layers/thread_tracker/thread_safety_validation.cpp
```diff
@@ -740,3 +740,31 @@ void ThreadSafety::PostCallRecordCreateRayTracingPipelinesKHR(VkDevice device, V
...
+
+void ThreadSafety::PreCallRecordQueuePresentKHR(VkQueue queue, const VkPresentInfoKHR* pPresentInfo) {
+    StartWriteObject(queue, vvl::Func::vkQueuePresentKHR);
+    uint32_t waitSemaphoreCount = pPresentInfo->waitSemaphoreCount;
+    if (pPresentInfo->pWaitSemaphores != nullptr) {
+        for (uint32_t index = 0; index < waitSemaphoreCount; index++) {
+            StartReadObject(pPresentInfo->pWaitSemaphores[index], vvl::Func::vkQueuePresentKHR);
+        }
+    }
+    for (uint32_t index = 0; index < pPresentInfo->swapchainCount; ++index) {
```
artem-lunarg, line 752: Is there a reason not to check `pSwapchains != nullptr` similar to `pWaitSemaphores` above?

layers/thread_tracker/thread_safety_validation.cpp
```diff
@@ -740,3 +740,31 @@ void ThreadSafety::PostCallRecordCreateRayTracingPipelinesKHR(VkDevice device, V
...
+        StartWriteObjectParentInstance(pPresentInfo->pSwapchains[index], vvl::Func::vkQueuePresentKHR);
```
artem-lunarg, line 753: @jeremy-lunarg @jzulauf-lunarg I'm not sure why swapchain was declared as `WRAPPER_PARENT_INSTANCE(VkSwapchainKHR)`. Swapchain it's device functionality. Instance level `VkSurfaceKHR` and `VkDisplayKHR` are declared with `WRAPPER_PARENT_INSTANCE` and it that case it looks logical.

If that's a bug then it should be fixed first in order not to use `StartWriteObjectParentInstance` here. But I also see this "parent instance" functionality the first time, could be missing something.

layers/thread_tracker/thread_safety_validation.cpp
AlinaKalyakina: From [spec](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPresentInfoKHR.html#VUID-VkPresentInfoKHR-pWaitSemaphores-parameter):
 - `VUID-VkPresentInfoKHR-pWaitSemaphores-parameter`
If waitSemaphoreCount is not 0, pWaitSemaphores must be a valid pointer to an array of waitSemaphoreCount valid [VkSemaphore](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkSemaphore.html) handles
- `VUID-VkPresentInfoKHR-pSwapchains-parameter`
pSwapchains must be a valid pointer to an array of swapchainCount valid [VkSwapchainKHR](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkSwapchainKHR.html) handles

So, `VkPresentInfoKHR` correctness was checked before and my check `pPresentInfo->pWaitSemaphores != nullptr` isn't needed also - worth removing I think.

artem-lunarg: Thanks for the clarification, that makes sense. I'd like also to get confirmation from @spencer-lunarg, do we usually assume that Stateless validation is enabled to avoid additional parameter checks?

spencer-lunarg: so looking at `CreateObjectDispatch` it seems we actually have `ThreadSafety` before `StatelessValidation` so it would not be able to assume valid null parameter checks at the moment (just from looking at the code, didn't write any tests to validate this idea yet)

AlinaKalyakina: Makes sense, cause it helps to catch more concurrency issues, for some reason I thought the other way around. Thanks a lot!

juan-lunarg:
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/KhronosGroup/Vulkan-ValidationLayers?pullRequest=6400) before we can accept your contribution.

Looks like you haven't signed the CLA. Are you having any issues with the CLA tool? We can't merge the PR until you have.

layers/thread_tracker/thread_safety_validation.cpp
AlinaKalyakina: I added checking if `pSwapchains` is null as well not to have unexpected crash in future

juan-lunarg:
> All committers have signed the CLA.

👍🏾

layers/thread_tracker/thread_safety_validation.cpp
artem-lunarg: I'll check parent instance for swapchain question tomorrow, and if it turns out it's an issue, then will make a separate PR. Then this fix can be updated based on that.

artem-lunarg: The use of `WRAPPER_PARENT_INSTANCE` is intentional, some time ago it was changed from device-level `WRAPPER` macro. The reason is not exactly clear for me. I opened a separate issue to clarify this https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6429 but this should not affect current PR.

layers/thread_tracker/thread_safety_validation.cpp
artem-lunarg: > ThreadSafety before StatelessValidation so it would not be able to assume valid null parameter checks at the moment

In a few places where I run debugger it looks that stateless validation was before thread safety. One pattern is that ThreadSafety is implemented as `Record` methods and at least part (all?) of the Stateless validation is in the `Validate` methods, so even if the same method is implemented by both validation objects, stateless is still called first because thread safety does not provide validate. Not sure about possible exceptions where Stateless does actual validation in `Record`.

artem-lunarg (approved):

PR #6402, opened 2023-08-31
layers: Revert "Simplify BINDABLE class hierarchy"

jeremyg-lunarg:
This reverts commit 1d1317e5761d926560a9ce11815483c6d73efeaa.

It is causing crashes in some internal tests.


charles-lunarg (approved):
Approving so the revert can happen. Just waiting on CI to finish.

PR #3650, opened 2022-01-04
tests: Add tests for VUIDs 01779 and 01781

jeremyg-lunarg:
WIP. This is a test case that needs working validation code to go with it...

PR #6407, opened 2023-09-01
Issue 5532 - Add WSI semaphore pending VUs

### LINKED ISSUE #5532: Add WSI semaphore pending VUs
Add 

- ` VUID-vkAcquireNextImageKHR-semaphore-01779`
- ` VUID-VkAcquireNextImageInfoKHR-semaphore-01781`

tests can be found in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/3650

`CoreChecks::ValidateAcquireNextImage` checks `VUID-VkAcquireNextImageInfoKHR-semaphore-01288`/`VUID-vkAcquireNextImageKHR-semaphore-01286` which is related

HildarTheDorf:
Resolves #5532 
Test originally authored @jeremyg-lunarg in #3650

### DIFF
--- a/layers/core_checks/cc_wsi.cpp
+++ b/layers/core_checks/cc_wsi.cpp
@@ -1082,6 +1082,11 @@ bool CoreChecks::ValidateAcquireNextImage(VkDevice device, VkSwapchainKHR swapch
                     version_2 ? "VUID-VkAcquireNextImageInfoKHR-semaphore-01288" : "VUID-vkAcquireNextImageKHR-semaphore-01286";
                 skip |= LogError(vuid, semaphore, loc, "Semaphore must not be currently signaled.");
             }
+            if (semaphore_state->InUse()) {
+                const char *vuid =
+                    version_2 ? "VUID-VkAcquireNextImageInfoKHR-semaphore-01781" : "VUID-vkAcquireNextImageKHR-semaphore-01779";
+                skip |= LogError(vuid, semaphore, loc, "Semaphore must not have any pending operations.");
+            }
         }
     }
 
--- a/tests/unit/wsi.cpp
+++ b/tests/unit/wsi.cpp
@@ -1654,7 +1654,6 @@ TEST_F(NegativeWsi, DeviceGroupSubmitInfoSemaphoreCount) {
 TEST_F(NegativeWsi, SwapchainAcquireImageWithSignaledSemaphore) {
     TEST_DESCRIPTION("Test vkAcquireNextImageKHR with signaled semaphore");
     SetTargetApiVersion(VK_API_VERSION_1_1);
-    AddRequiredExtensions(VK_KHR_DEVICE_GROUP_EXTENSION_NAME);
     AddSurfaceExtension();
 
     ASSERT_NO_FATAL_FAILURE(InitFramework(m_errorMonitor));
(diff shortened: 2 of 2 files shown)


tests/unit/wsi.cpp
```diff
@@ -1692,6 +1692,29 @@ TEST_F(NegativeWsi, SwapchainAcquireImageWithSignaledSemaphore) {
     m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkAcquireNextImageInfoKHR-semaphore-01288");
     vk::AcquireNextImage2KHR(device(), &acquire_info, &dummy);
     m_errorMonitor->VerifyFound();
+
+    // Add a wait, but don't let it finish.
```
artem-lunarg, line 1696: This test is for the specific case when "semaphore is signaled".  Is it possible to move new scenario to a separate "semaphore with wait pending" test?

HildarTheDorf: It's possible sure. I tried both and found that adding a separate test basically requires copying the entire SwapchainAcquireImageWithSignaledSemaphore test into the new 'SwapchainAcquireImageWithSemaphorePendingWait' as you need to first submit a semaphore to be able to add a pending wait.

spencer-lunarg: so this is a constant topic discussion we have of when its good to "combine tests"

- It is MUCH nicer to have smaller tests when debugging things
- There are sometimes we just need to except test coverage consumes some "repeated setup" as that is the nature of the beast of testing
- For this tests, I think it would be probably better to just make a 2nd test as there isn't that much to repeat
- We probably don't need `AddRequiredExtensions(VK_KHR_DEVICE_GROUP_EXTENSION_NAME);` in these tests looking at it

HildarTheDorf: It needs 1.1 OR VK_KHR_DEVICE_GROUP_EXTENSION_NAME. The current test enables both.

spencer-lunarg: yep, just going

`SetTargetApiVersion(VK_API_VERSION_1_1);` and `if (DeviceValidationVersion() < VK_API_VERSION_1_1) {` is sufficient, can remove the `AreRequiredExtensionsEnabled` check then

spencer-lunarg (approved):

tests/unit/wsi.cpp
artem-lunarg: For me, the main point here that the test name should be valid (and completely agree with having small tests). Without good names it's harder to find tests that cover specific area. In the rare case when test should be updated it's a good idea to ensure that the test name reflects current functionality.

layers/core_checks/cc_wsi.cpp
```diff
@@ -1082,6 +1082,11 @@ bool CoreChecks::ValidateAcquireNextImage(VkDevice device, VkSwapchainKHR swapch
                     version_2 ? "VUID-VkAcquireNextImageInfoKHR-semaphore-01288" : "VUID-vkAcquireNextImageKHR-semaphore-01286";
                 skip |= LogError(vuid, semaphore, loc, "Semaphore must not be currently signaled.");
             }
+            if (semaphore_state->InUse()) {
```
artem-lunarg, line 1085: @jeremyg-lunarg my thinking that using `InUse` here is according to design. Not 100% in the absence of any issues since swapchain sync can be tricky, but if this reveals any problem then it's the other code that should be fixed.

artem-lunarg (approved):

HildarTheDorf:
MacOS build failed as the `VkLayerTests::AddSurfaceExtension()` doesn't support Mac at all.

spencer-lunarg:
> MacOS build failed as the VkLayerTests::AddSurfaceExtension() doesn't support Mac at all.

Yes, I see we need to add proper support for WSI for MacOS yet ... Will make an issue for that, but for now, just going

```c++
    if (!InitSwapchain()) {
        GTEST_SKIP() << "Cannot create surface or swapchain";
    }
```

should get you around this

juan-lunarg (approved):
LGTM just needs the workaround for Mac

juan-lunarg:
> Yes, I see we need to add proper support for WSI for MacOS yet ... Will make an issue for that, but for now, just going

We now have WSI support for MacOS.

spencer-lunarg:
rebased https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7137

PR #6409, opened 2023-09-01
layers: Hash ResourceInterfaceVariable descriptor members

spencer-lunarg:
This is a rehash of https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5703 for removing the `DescriptorReqFlags` as it is currently a trap to not realize if some `ResourceInterfaceVariable` are changed they may not be updated and when being validated at draw time

### DIFF
--- a/layers/state_tracker/shader_module.cpp
+++ b/layers/state_tracker/shader_module.cpp
@@ -1767,6 +1767,11 @@ const Instruction& ResourceInterfaceVariable::FindBaseType(ResourceInterfaceVari
     return *type;
 }
 
+uint32_t ResourceInterfaceVariable::FindImageSampledTypeWidth(const SPIRV_MODULE_STATE& module_state,
+                                                              const Instruction& base_type) {
+    return (base_type.Opcode() == spv::OpTypeImage) ? module_state.GetTypeBitsSize(&base_type) : 0;
+}
+
 NumericType ResourceInterfaceVariable::FindImageFormatType(const SPIRV_MODULE_STATE& module_state, const Instruction& base_type) {
     return (base_type.Opcode() == spv::OpTypeImage) ? module_state.GetNumericType(base_type.Word(2)) : NumericTypeUnknown;
 }
@@ -1781,6 +1786,11 @@ bool ResourceInterfaceVariable::IsStorageBuffer(const ResourceInterfaceVariable&
     return ((uniform && buffer_block) || ((storage_buffer || physical_storage_buffer) && block));
 }
 
+bool ResourceInterfaceVariable::IsAtomicOperation(const SPIRV_MODULE_STATE& module_state,
+                                                  const ResourceInterfaceVariable& variable) {
+    return !module_state.FindVariableAccesses(variable.id, module_state.static_data_.atomic_pointer_ids, true).empty();
+}
+
 ResourceInterfaceVariable::ResourceInterfaceVariable(const SPIRV_MODULE_STATE& module_state, const EntryPoint& entrypoint,
                                                      const Instruction& insn, const ImageAccessMap& image_access_map)
     : VariableBase(module_state, insn, entrypoint.stage),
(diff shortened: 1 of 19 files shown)


arno-lunarg (approved):

layers/state_tracker/shader_module.cpp
```diff
@@ -1844,8 +1853,10 @@ ResourceInterfaceVariable::ResourceInterfaceVariable(const SPIRV_MODULE_STATE& m
 
     // Type independent checks
     if (!module_state.FindVariableAccesses(id, static_data_.atomic_pointer_ids, true).empty()) {
-        is_atomic_operation = true;
+        info.is_atomic_operation = true;
     }
+
+    descriptor_hash = XXH32(&info, sizeof(info), 0);
```
artem-lunarg, line 1859: Can be a problem if `Info` structure has gaps/paddings. In that case it's random data. Did not check if we explicitly mem-zero entire structure. If yes, that's not a problem. If no, one solution is to zero entire memory footprint of the structure before any other initialization. Or another solution is to ensure fields of guaranteed sizes are used, specify padding explicitly, so all memory covered by variables and can be zero explicitly, and also to add static assert that checks that structure size is what you expect. The `memset` method looks easier but both are the options.

juan-lunarg, line 1859: Any particular reason to use the XXH32 hash?

xxhash offers faster hashes.

https://github.com/Cyan4973/xxHash

layers/state_tracker/shader_module.h
```diff
@@ -301,13 +301,28 @@ struct ResourceInterfaceVariable : public VariableBase {
...
-    const spv::Dim image_dim;
-    const bool is_image_array;
-    const bool is_multisampled;
     // Sampled Type width of the OpTypeImage the variable points to, 0 if doesn't use the image
-    uint32_t image_sampled_type_width = 0;
+    const uint32_t image_sampled_type_width;
+
+    // All info regarding what will be validated from requirements imposed by the pipeline on a descriptor. These
+    // can't be checked at pipeline creation time as they depend on the Image or ImageView bound.
+    // That is perf-critical code and hashing if 2 variables have same info provides a 20% perf bonus
```
juan-lunarg, line 309: If this is perf critical all the more reason to use a faster hash. At the bare minimum XXH64 will easily outperform XXH32

artem-lunarg: Might be a good idea additionally to confirm it's true for small input sizes. The benchmark results can be from testing MBs or GBs of continuous data chunks which can make initial warm up/set up (if any) insignificant, and it might be still important for us.

juan-lunarg: > It looks faster for small datasets too.

Yep. I did a fair amount of hash perf testing a bit ago. XXH64 uses 64 bit registers and has less work to do to reduce collisions.

layers/utils/shader_utils.h
```diff
@@ -107,12 +79,16 @@ inline ShaderObjectStage VkShaderStageToShaderObjectStage(VkShaderStageFlagBits
...
+}
 
-inline bool operator<(const DescriptorRequirement &a, const DescriptorRequirement &b) noexcept { return a.reqs < b.reqs; }
+inline bool operator<(const DescriptorRequirement &a, const DescriptorRequirement &b) noexcept {
+    return a.revalidate_hash < b.revalidate_hash;
+}
 
 // < binding index (of descriptor set) : meta data >
-typedef std::map<uint32_t, DescriptorRequirement> BindingVariableMap;
+typedef std::unordered_map<uint32_t, DescriptorRequirement> BindingVariableMap;
```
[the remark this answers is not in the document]
jeremyg-lunarg: yep, any use of std::unordered_map<> should be accompanied by a comment explaining why  it cannot be the vvl:: version.

layers/state_tracker/shader_module.cpp
spencer-lunarg: seems we have `XXH_NO_LONG_LONG` currently in, once I get this in, plan to create an issue and another hash and benchmark the perf diff

spencer-lunarg: agree, going to memset

layers/state_tracker/shader_module.h
spencer-lunarg: @juan-lunarg I see we have `XXH_NO_LONG_LONG` which seems to prevent any other hash from being used at the moment

layers/utils/shader_utils.h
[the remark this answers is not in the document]
spencer-lunarg: I was having issues because I was getting

> could not match 'std::pair' against 'robin_hood::pair'

due to

> using DescriptorBindingInfo = std::pair<const uint32_t, DescriptorRequirement>;

but going to re-work things so can use `vvl::unordered_map`

layers/state_tracker/shader_module.h
juan-lunarg: That was to fix a chromium bug. You'll need to update the xxhash dependency and remove that XXH_NO_LONG_LONG code.

juan-lunarg: I'll make a PR to address this.

juan-lunarg: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6804

spencer-lunarg: updated to `XXH64` and added to `hash_util`

layers/utils/hash_util.cpp
```diff
@@ -32,14 +32,19 @@ static_assert(XXH_VERSION_RELEASE == 2);
...
 
-uint32_t shader_hash(const void *pCode, const size_t codeSize) {
+uint32_t ShaderHash(const void *pCode, const size_t codeSize) {
     constexpr uint32_t seed = 0;
     return XXH32(pCode, codeSize, seed);
 }
 
+uint64_t DescriptorVariableHash(const void *info, const size_t info_size) {
+    constexpr uint64_t seed = 0;
+    return XXH64(info, info_size, seed);
```
juan-lunarg, line 47: 💯

juan-lunarg (approved):

PR #6411, opened 2023-09-01
layers: Redo "Simplify BINDABLE class hierarchy""

jeremyg-lunarg:
The earlier version of this change needed to add BUFFER_STATE::Destroy() because of interactions between the memory tracker and the BINDABLE base class.

### DIFF
--- a/layers/state_tracker/buffer_state.h
+++ b/layers/state_tracker/buffer_state.h
@@ -18,6 +18,7 @@
  * limitations under the License.
  */
 #pragma once
+#include <variant>
 #include "state_tracker/device_memory_state.h"
 #include "containers/range_vector.h"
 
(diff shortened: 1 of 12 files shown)


layers/state_tracker/buffer_state.h
```diff
@@ -40,6 +40,14 @@ class BUFFER_STATE : public BINDABLE {
 
     BUFFER_STATE(BUFFER_STATE const &rh_obj) = delete;
 
+    // This destructor is needed because BINDABLE depends on the tracker_ variant defined in this
```
jeremyg-lunarg, line 43: This is the part that I added to fix the crash.

arno-lunarg (approved):

artem-lunarg (approved):

PR #6415, opened 2023-09-04
layers: Untemplate some barrier methods

artem-lunarg:
The problem with templated methods is that parameterized barrier types have to be the most common denominator for sync1/sync2, so they do not include stage masks. Stage masks parameters have to be added where needed or new overloads are introduced.

This PR introduces helper barrier structures and converts some existing methods to untemplated form (more to follow). I started with the methods I worked with recently and had troubles with extending them due to the fact that stage mask is not included. Code got more compact in most cases too.

New types are similar to sync2 types with a benefit of hierarchical data organization. For example, `BufferBarrier can` also be used by the apis that accept `QueueFamilyBarrier` or `MemoryBarrier` types. New barrier types can be initialized with both sync2 and sync1 types.  For sync1 types, stage masks should be additionally provided.

`sType/pNext` are not included. Currently it's not needed for validation and if there is a need the idea to pass that information using higher level concepts or separately.

Initialization of new structures is efficient. Comparing to `Location` structure, initialization of barrier structures on average takes 1x-2x more instructions, but barrier structures are less common. GCC/Clang compilers generates 16-byte moves for initialization code, and for example, initialization of `BufferBarrier` from `VkBufferMemoryBarrier2` it's 4 16-byte moves. MSVC compiler does not do this. There is a way to do this with `memcpy` but I decided not to include this because it complicates code and CGG/Clang already do this, which limits the benefits (example implementation with optimized  wide moves for all compilers is here https://github.com/artem-lunarg/Vulkan-ValidationLayers/commit/09ccc39932f1bc2c86883fdb77996b3724a4a0eb)

### DIFF
--- a/layers/core_checks/cc_synchronization.cpp
+++ b/layers/core_checks/cc_synchronization.cpp
@@ -27,6 +27,11 @@
 #include "sync/sync_utils.h"
 #include "generated/enum_flag_bits.h"
 
+using sync_utils::BufferBarrier;
+using sync_utils::ImageBarrier;
+using sync_utils::MemoryBarrier;
+using sync_utils::QueueFamilyBarrier;
+
 ReadLockGuard CoreChecks::ReadLock() const {
     if (fine_grained_locking) {
         return ReadLockGuard(validation_object_mutex, std::defer_lock);
@@ -785,7 +790,7 @@ bool CoreChecks::ValidateRenderPassPipelineBarriers(const Location &outer_loc, c
     }
 
     for (uint32_t i = 0; i < image_mem_barrier_count; ++i) {
-        const auto &img_barrier = image_barriers[i];
+        const auto img_barrier = ImageBarrier(image_barriers[i], src_stage_mask, dst_stage_mask);
         Location loc(outer_loc.function, Struct::VkImageMemoryBarrier, Field::pImageMemoryBarriers, i);
         skip |= state.ValidateAccess(loc, img_barrier.srcAccessMask, img_barrier.dstAccessMask);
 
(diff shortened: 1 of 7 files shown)


layers/core_checks/cc_synchronization.cpp
```diff
@@ -27,6 +27,14 @@
 #include "sync/sync_utils.h"
 #include "generated/enum_flag_bits.h"
 
+using sync_utils::BufferBarrier;
+using sync_utils::ImageBarrier;
+using sync_utils::MemoryBarrier;
+using sync_utils::QueueFamilyBarrier;
+using vvl::Field;
```
spencer-lunarg, line 34: we do this in the `CoreChecks` header, so not sure we need the `using` for the Location items here (I could be completely wrong though)

layers/core_checks/cc_synchronization.cpp
```diff
@@ -1502,9 +1510,9 @@ bool CoreChecks::ValidateBarrierLayoutToImageUsage(const Location &loc, VkImage
 }
 
 // Verify image barriers are compatible with the images they reference.
-template <typename ImageBarrier>
+template <typename ImgBarrier>
 bool CoreChecks::ValidateBarriersToImages(const Location &outer_loc, const CMD_BUFFER_STATE *cb_state,
-                                          uint32_t imageMemoryBarrierCount, const ImageBarrier *pImageMemoryBarriers) const {
+                                          uint32_t imageMemoryBarrierCount, const ImgBarrier *pImageMemoryBarriers) const {
```
spencer-lunarg, line 1515: nit: seems all the other changes in this PR are `ImageBarrier` but aren't here

spencer-lunarg (approved):

layers/core_checks/cc_synchronization.cpp
artem-lunarg: Used that for free functions. I see that in other places we use `vvl::` in free functions, which looks good, but I find `sync_utils::` a bit ugly... the idea was that for sync code new barrier structures are directly accessible but `sync_utils.h` looked like a good place where to define them. 

Going to remove `vvl::` usings and to keep `sync_utils::` ones.

layers/core_checks/cc_synchronization.cpp
artem-lunarg: It is still a templated function. I have to use different name than `ImageBarrier` for now. 

If it make sense to untemplate this method then it will be fixed, and if it makes sense to keep a template (for some functions it might be a good solution) then we can rename this to just Barrier to avoid confusion. Going to do a rename pass later. I left some ugly names in other places too and also use `srcStageMask` camel case instead of `src_stage_mask` in new structures to avoid noise due to renames.

artem-lunarg: `ValidateBarriersToImages` will be untemplated as part of the next batch (https://github.com/artem-lunarg/Vulkan-ValidationLayers/commit/521b0afc69dcb19adb04170ebc56a5c99f134c8b)

arno-lunarg (approved):
LGTM!

jeremyg-lunarg (approved):
nice work!

PR #6416, opened 2023-09-04
Improve the Pipeline testing

spencer-lunarg:
This aims to

- Making the `CreatePipelineHelper` the only system we use (not also `VkPipelineObj` sometimes, but this is the first part of this)
- Having dedicated files to find things easier in our ever-expending test helper file
- Reduce redundant work and try to simplify the pipeline code a bit


juan-lunarg:
PositiveShaderObject.DrawMinimalDynamicStatesRasterizationDisabled failed on 32 bit release mode. Not sure if it's related to your PR.

```
[ RUN      ] PositiveShaderObject.DrawMinimalDynamicStatesRasterizationDisabled
D:\a\Vulkan-ValidationLayers\Vulkan-ValidationLayers\tests\framework\error_monitor.cpp(200): error: Failed
Validation Error: [ VUID-vkCmdDraw-None-08617 ] Object 0: handle = 0x46dd668, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x4c85677 | vkCmdDraw():  VK_DYNAMIC_STATE_LINE_WIDTH state not set for this command buffer. The Vulkan spec states: If a shader object is bound to any graphics stage, and the most recent call to vkCmdSetPolygonModeEXT in the current command buffer set polygonMode to VK_POLYGON_MODE_LINE, vkCmdSetLineWidth must have been called in the current command buffer prior to this drawing command (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdDraw-None-08617)

[  FAILED  ] PositiveShaderObject.DrawMinimalDynamicStatesRasterizationDisabled (34 ms)
```

tests/framework/pipeline_helper.h
```diff
@@ -0,0 +1,263 @@
...
+
+#pragma once
+
+#include "layer_validation_tests.h"
+
+// Helper class for tersely creating create pipeline tests
+//
+// Designed with minimal error checking to ensure easy error state creation
+// See OneshotTest for typical usage
+struct CreatePipelineHelper {
```
juan-lunarg, line 22: nitpick FFTI

If you add a methods to a struct. It's no longer a struct. It's a class.

```suggestion
class CreatePipelineHelper {
```

Just make the necessary members `public`

tests/framework/pipeline_helper.h
```diff
@@ -0,0 +1,263 @@
...
+    template <typename Test, typename OverrideFunc>
+    static void OneshotTest(Test &test, const OverrideFunc &info_override, const VkFlags flags) {
+        OneshotTest(test, info_override, flags, std::vector<std::string>{});
+    }
+};
+
+// Helper class for to create ray tracing pipeline tests
+// Designed with minimal error checking to ensure easy error state creation
+// See OneshotTest for typical usage
+struct RayTracingPipelineHelper {
```
juan-lunarg, line 198: similar class complaint here

tests/unit/amd_best_practices.cpp
```diff
@@ -265,7 +264,6 @@ TEST_F(VkAmdBestPracticesLayerTest, NumDynamicStates) {
     dynamic_state_info.pDynamicStates = dynamic_states_array;
 
     CreatePipelineHelper pipe(*this);
-    pipe.InitInfo();
     pipe.InitState();
```
juan-lunarg, line 267: Good cleanup 👍🏾 

The double init was weird.

juan-lunarg (approved):

tests/framework/pipeline_helper.h
spencer-lunarg: agree, it was a struct with `public` which was more strange, but agree, just make it a class with `public` this a crazy struct doing too much

PR #6418, opened 2023-09-05
layers: Add arrow to pointers in Location error message

spencer-lunarg:
@aqnuep pointed out a good idea of just generating the `Fields` that are pointers so now the error message will use an `->` when a pointer (and no index)

```patch
- pCreateInfo.pAttachments[0].format
+ pCreateInfo->pAttachments[0].format
```


arno-lunarg (approved):

aqnuep:
This seems fine. In the meantime, I realized I will have to do some further modifications to allow passing locations of fields vs pNext structs as a whole as parameter, I'll post a PR of that after you get this in.

juan-lunarg (approved):

PR #6421, opened 2023-09-05
layers: Check correct flag for SetColorBlendEnable

Darianopolis:
The SetColorBlendEnable check VUID-08643 should check for the relevant `pColorBlendEnables[i]` bit set, which is tracked by `cb_state->dynamic_state_value.color_blend_enabled` in [state_tracker.cpp:5830-5843](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/bea5beb302dc1add1bb9b52c8de37c0b92687159/layers/state_tracker/state_tracker.cpp#L5830-L5843)
Instead, it checks `cb_state->dynamic_state_value.color_blend_enable_attachments` which is set unconditionally.

Thus, with a color attachment that doesn't support `VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BLEND_BIT`, this validation check always fails even when `VK_FALSE` is correctly passed to `vkCmdSetColorBlendEnableEXT` for the relevant attachment.

### DIFF
--- a/layers/core_checks/cc_cmd_buffer_dynamic.cpp
+++ b/layers/core_checks/cc_cmd_buffer_dynamic.cpp
@@ -666,7 +666,7 @@ bool CoreChecks::ValidateDrawDynamicStateShaderObject(const LAST_BOUND_STATE& la
                     const auto attachment = (*cb_state.active_attachments)[i];
                     if (attachment && FormatIsColor(attachment->create_info.format) &&
                         (attachment->format_features & VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BLEND_BIT) == 0 &&
-                        cb_state.dynamic_state_value.color_blend_enable_attachments[i] == VK_TRUE) {
+                        cb_state.dynamic_state_value.color_blend_enabled[i] == VK_TRUE) {
                         skip |= LogError(vuid.set_color_blend_enable_08643, cb_state.commandBuffer(), loc,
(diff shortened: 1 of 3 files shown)


juan-lunarg (commented):
A test needs to be added to `tests/unit/shader_object.cpp` to prevent a regression. We have documentation explaining the process for adding a test. Although you should be able to figure it out from the tests in that file.

ziga-lunarg (approved):

juan-lunarg:
Ah looks like there is an existing test case that needs modification with this change `NegativeShaderObject.MissingCmdSetColorBlendEnable` is failing now. So fixing that test should be good enough to get this PR merged in.

Darianopolis:
@juan-lunarg Thanks, I've updated the tests now.

I had a look at `MissingCmdSetColorBlendEnable` and realized that it was indeed trying to test the negative case for this issue, however it was missing some setup and the name seemed to be copied from `MissingCmdSetColorBlendEnableEXT`.
I renamed it to `BlendEnabledWithNonBlendableFormat` and fixed the setup to match the required test case.

I also added a positive test case `DrawWithNonBlendableFormat` to catch any regressions (this new test correctly catches the original issue with `color_blend_enable_attachments`).

juan-lunarg (approved):
🥇 

Thanks for adding the additional testing. Fixes are nice, but fixes with tests are fantastic 👍🏾

PR #6422, opened 2023-09-05
Fix TransitionImageLayout VUID now that there is one

Tony-LunarG:


layers/core_checks/cc_image.cpp
```diff
@@ -2366,7 +2366,7 @@ bool CoreChecks::PreCallValidateTransitionImageLayoutEXT(VkDevice device, uint32
         if (transition.oldLayout != VK_IMAGE_LAYOUT_UNDEFINED) {
```
juan-lunarg, line 2366: > oldLayout must be either VK_IMAGE_LAYOUT_UNDEFINED or the current layout of the image subresources as specified in subresourceRange

I might be missing something but it looks like this handles 1/2 of the VUID as is.

Tony-LunarG: I read it as "if old layout is UNDEFINED, no need to see what the current layout is.  If old layout is not UNDEFINED (the if check), it has to match the current layout (the Validate call)".  But I've confused these sorts of things before.

juan-lunarg (approved):

jeremyg-lunarg (approved):

PR #6424, opened 2023-09-05
layers: Error location improvements

aqnuep:
This PR allows a Location object to point to an entire structure within the pNext chain, which is necessary in order to allow a function to accept structure locations indifferent of whether they are specified as a member of another structure or chained to another structure.


spencer-lunarg (approved):
Awesome, this is a really nice enhancement for `Location`

PR #6425, opened 2023-09-05
corechecks: Make QNX buffer import validation according to spec

mgorchak-blackberry:
Add QNX buffer extension support for allocated memory size validation.

Validation message has mention of QNX, but not in the code:

ERROR : VALIDATION - Message Id Number: 1700101245 | Message Id Name: VUID-VkMemoryDedicatedAllocateInfo-image-02964
        Validation Error: [ VUID-VkMemoryDedicatedAllocateInfo-image-02964 ] | MessageID = 0x65557c7d |
vkAllocateMemory: Allocation Size (8355840) needs to be equal to VkImage VkImage 0x2cfba2000000001c[]
VkMemoryRequirements::size (8896880) The Vulkan spec states: If image is not VK_NULL_HANDLE and the
memory is not an imported Android Hardware Buffer or an imported QNX Screen buffer,
VkMemoryAllocateInfo::allocationSize must equal the VkMemoryRequirements::size of the image
(https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkMemoryDedicatedAllocateInfo-image-02964)

With this patch this message will be suppressed.

Thank you!


juan-lunarg (approved):

juan-lunarg (changes requested):
Run scripts/check_code_format.py 

Your commit message doesn't fit our style guidelines.

PR #6426, opened 2023-09-05
build: Update to header 1.3.263

mikes-lunarg:
- Update known-good
- Generate source


juan-lunarg (approved):

mikes-lunarg:
Opened https://github.com/KhronosGroup/Vulkan-Utility-Libraries/pull/94 to fix the Chromium integration CI failure

PR #1557, opened 2020-02-17
Plumb thread safety API names through to error messages

mark-lunarg:
Threading object would generate errors, but with no mention of the API which caused the problem.  Modified the codegen to include the api name as part of the validation error messages.

Fixes #259.

PR #6428, opened 2023-09-06
layers: Fix double destruction in thread safety

artem-lunarg:
Probably an editing error from https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/1557

Does not affect correctness, for the second `DestroyObject` call the hash table's `erase` can't find the key and returns. Also, one fewer `WriteLockGuard` lock.


spencer-lunarg (approved):
LGTM, never seen this before

PR #6436, opened 2023-09-07
layers: Fix VUIDs for image copies

spencer-lunarg:
The 1.3.263 spec removed a lot of bad VUs and this is just cleaning them up


charles-lunarg (approved):
Cleanup is good!

Issue #6440, opened 2023-09-07
Check for VUID 06706 only checks multiview extension property but not Vulkan 1.1 property


aqnuep:
**Environment:**
 - OS: All
 - GPU: Using a profile that sets `VkPhysicalDeviceVulkan11Properties::maxMultiviewViewCount` but not `VkPhysicalDeviceMultiviewProperties::maxMultiviewViewCount`
 - SDK or header version if building from repo: any recent
 - Options enabled (synchronization, best practices, etc.): none

**Describe the Issue**

It seems certain validation code for multiview wasn't updated to check the `VkPhysicalDeviceVulkan11Properties` values instead of the `VkPhysicalDeviceMultiviewProperties`. This causes the `NegativeSubpass.SubpassDescriptionViewMask` test case to fail on an implementation that supports Vulkan 1.2 (and thus multiview and the `VkPhysicalDeviceVulkan11Properties` structure), but does not expose the `VK_KHR_multiview` extension or expose its properties also in the `VkPhysicalDeviceMultiviewProperties` structure.

A quick search seems to indicate that the only 2 validation cases affected by issue are VUID 06706 and VUID 02688, as all other cases seem to have been updated appropriately at the time.

**Expected behavior**

The test case `NegativeSubpass.SubpassDescriptionViewMask` should pass on an implementation that supports Vulkan 1.2 (and thus multiview and the `VkPhysicalDeviceVulkan11Properties` structure), but does not expose the `VK_KHR_multiview` extension or expose its properties also in the `VkPhysicalDeviceMultiviewProperties` structure.

**Valid Usage ID**
VUID 06706
VUID 02688

**Additional context**

<details>
<summary>Test case output</summary>

```sh
[ RUN      ] NegativeQuery.MultiviewBeginQuery
error_monitor.cpp: Failure
Failed
[...]

PR #6442, opened 2023-09-07
layers: Various feature pre-requisite layer and test fixes

aqnuep:
Fixes issue #6440, among other feature pre-requisite related issues in various test cases.

In general, there're a large number of test cases that do not properly check their prerequisites and assume the extension to be supported, the KHR extension being supported for an otherwise promoted feature, etc.


arno-lunarg (approved):
LGTM!

PR #6443, opened 2023-09-07
layer: Clean up layer settings code

christophe-lunarg:
Layer settings where read both by the old and the new system as a first step, removing the old system which is entirely covered by the new system and more use of the `Vulkan-Layer-Settings` library capabilities

### DIFF
--- a/layers/layer_options.cpp
+++ b/layers/layer_options.cpp
@@ -291,120 +291,25 @@ void CreateFilterMessageIdList(std::string raw_id_list, const std::string &delim
     }
 }
 
-void SetCustomStypeInfo(std::string raw_id_list, const std::string &delimiter) {
-    size_t pos = 0;
-    std::string token;
-    // List format is a list of integer pairs
-    while (raw_id_list.length() != 0) {
-        token = GetNextToken(&raw_id_list, delimiter, &pos);
-        uint32_t stype_id = TokenToUint(token);
-        token = GetNextToken(&raw_id_list, delimiter, &pos);
-        uint32_t struct_size_in_bytes = TokenToUint(token);
-        if ((stype_id != 0) && (struct_size_in_bytes != 0)) {
-            bool found = false;
-            // Prevent duplicate entries
-            for (const auto &item : custom_stype_info) {
-                if (item.first == stype_id) {
-                    found = true;
-                    break;
-                }
-            }
-            if (!found) custom_stype_info.push_back(std::make_pair(stype_id, struct_size_in_bytes));
-        }
+static void SetValidationSetting(VlLayerSettingSet layer_setting_set, CHECK_DISABLED &disable_data,
+                                 const DisableFlags feature_disable, const char *setting) {
+    if (vlHasLayerSetting(layer_setting_set, setting)) {
+        bool enabled = true;
+        vlGetLayerSettingValue(layer_setting_set, setting, enabled);
+        disable_data[feature_disable] = !enabled;
     }
 }
 
-uint32_t SetMessageDuplicateLimit(const std::string &config_message_limit, const std::string &env_message_limit) {
-    uint32_t limit = 0;
-    auto get_num = [](const std::string &source_string) {
-        uint32_t limit = 0;
-        int radix = ((source_string.find("0x") == 0) ? 16 : 10);
-        limit = static_cast<uint32_t>(std::strtoul(source_string.c_str(), nullptr, radix));
-        return limit;
-    };
-    // ENV var takes precedence over settings file
-    limit = get_num(env_message_limit);
-    if (limit == 0) {
-        limit = get_num(config_message_limit);
-    }
-    return limit;
-}
-
-static bool SetBool(const std::string &config_string, const std::string &env_string, bool default_val) {
-    bool result = default_val;
-
-    std::string setting;
-    if (!env_string.empty()) {
-        setting = env_string;
-    } else if (!config_string.empty()) {
-        setting = config_string;
-    }
-    if (!setting.empty()) {
-        vvl::ToLower(setting);
-        if (setting == "true") {
-            result = true;
-        } else {
-            result = std::atoi(setting.c_str()) != 0;
-        }
-    }
-    return result;
-}
-
-static std::string GetSettingKey(const char *setting) {
-    const std::string prefix("khronos_validation.");
-    return prefix + setting;
-}
-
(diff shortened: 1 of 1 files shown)


layers/layer_options.cpp
```diff
@@ -425,38 +330,49 @@ void ProcessConfigAndEnvSettings(ConfigAndEnvSettings *settings_data) {
...
     std::vector<std::string> disables;
     if (vlHasLayerSetting(layerSettingSet, SETTING_DISABLES)) {
         vlGetLayerSettingValues(layerSettingSet, SETTING_DISABLES, disables);
     }
-    const std::string& string_disables = Merge(disables);
+    const std::string &string_disables = Merge(disables);
     SetLocalDisableSetting(string_disables, ",", settings_data->disables);
 
+    // Fine Grained Locking
+    *settings_data->fine_grained_locking = true;
```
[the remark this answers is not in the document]
juan-lunarg: Worth bringing up in the VVL meeting next week.

layers/layer_options.cpp
```diff
@@ -467,77 +383,47 @@ void ProcessConfigAndEnvSettings(ConfigAndEnvSettings *settings_data) {
...
-    const bool use_disables_fine_gain_settings = config_value_disables.empty() && envvar_value_disables.empty();
-    const std::string &config_value_enables = GetConfigValue(SETTING_ENABLES);
-    const std::string &envvar_value_enables = GetEnvVarValue(SETTING_ENABLES);
-    const bool use_enables_fine_gain_settings = config_value_enables.empty() && envvar_value_enables.empty();
-    const bool use_fine_gain_settings = use_disables_fine_gain_settings && use_enables_fine_gain_settings;
-
-    // Process layer enable settings
-    SetLocalEnableSetting(config_value_enables, ",", settings_data->enables);
-    SetLocalEnableSetting(envvar_value_enables, env_delimiter, settings_data->enables);
+    const bool use_fine_gain_settings = disables.empty() && enables.empty();
```
artem-lunarg, line 386: nit: It's the name from the original code, but probably a typo in `fine_grained`

juan-lunarg (approved):

layers/layer_options.cpp
```diff
@@ -425,38 +330,49 @@ void ProcessConfigAndEnvSettings(ConfigAndEnvSettings *settings_data) {
...
                             &layerSettingSet);
-   
+
+    // Read legacy "enables" flags for backward compatibility
     std::vector<std::string> enables;
     if (vlHasLayerSetting(layerSettingSet, SETTING_ENABLES)) {
         vlGetLayerSettingValues(layerSettingSet, SETTING_ENABLES, enables);
     }
-    const std::string& string_enables = Merge(enables);
+    const std::string &string_enables = Merge(enables);
```
artem-lunarg, line 339: I remember we discussed in the previous review that it's technically correct in C++. Still, reference in this expression looks unconventional. Assigned by value might be more efficient, because of NRVO compiler optimization which compilers don't have much reasons not to do if local variable is returned.  Having a reference will disable NRVO, and not sure about this, but could be additional move/copy construction.

artem-lunarg (approved):
LGTM. Some suggestions.

layers/layer_options.cpp
artem-lunarg: I would argue that with reference it's slower or in the best case the same amount of instructions. Please check the first version of the above answer (in the edited).

I guess the best way to be sure 100% is to check generated code for main compilers.

layers/layer_options.cpp
```diff
@@ -291,120 +291,25 @@ void CreateFilterMessageIdList(std::string raw_id_list, const std::string &delim
...
-        vvl::ToLower(setting);
-        if (setting == "true") {
-            result = true;
-        } else {
-            result = std::atoi(setting.c_str()) != 0;
-        }
+static void SetValidationSetting(VlLayerSettingSet layerSettingSet, CHECK_DISABLED &disable_data,
+                                 const DisableFlags feature_disable, const char *setting) {
+    if (vlHasLayerSetting(layerSettingSet, setting)) {
+        bool setting_value = true;
```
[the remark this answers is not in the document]
christophe-lunarg: That's a good idea, fixed !

layers/layer_options.cpp
```diff
@@ -291,120 +291,25 @@ void CreateFilterMessageIdList(std::string raw_id_list, const std::string &delim
...
-        setting = config_string;
-    }
-    if (!setting.empty()) {
-        vvl::ToLower(setting);
-        if (setting == "true") {
-            result = true;
-        } else {
-            result = std::atoi(setting.c_str()) != 0;
-        }
+static void SetValidationSetting(VlLayerSettingSet layerSettingSet, CHECK_DISABLED &disable_data,
```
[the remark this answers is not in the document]
christophe-lunarg: fixed

layers/layer_options.cpp
christophe-lunarg: https://herbsutter.com/2008/01/01/gotw-88-a-candidate-for-the-most-important-const/
I am not seeing how that's unconventional.

A reference slower than calling a constructor... That's a stretch. For a class that implement r-value reference maybe but that's precisely because the contructor is not actually called.

That has been used to have the nice form "type func()" instead of func(type& ) and avoid copies and avoid having to hope the compiler can/will optimize all this correctly.

layers/layer_options.cpp
christophe-lunarg: Fixed

layers/layer_options.cpp
artem-lunarg: RNVO optimization does not call a constructor but constructs object in destination place.

> A reference slower than calling a constructor... That's a stretch.

Lets do a mental experiment how it works with a reference. At first the object is constructed on the stack in the stack frame of the called function. When the function returns in order `const std::string&` to point to a valid object the compiler has to generate a copy, because stack space gets overriden by other locals of the caller (and other calls of the caller). So returning const ref in the classical c++ model suggets a copy or a move. In practice I can imagine compiler can do optimization similar to RNVO to construct object directly in the destination space in the stack frame of the caller - in that case performance will be as with RNVO.

I believe that post it's the place where I learnt this the first time. My impression this approach is less common in the codebases, that's why I said unconventional.  But I'm fine with references if you prefer that approach!

PR #6444, opened 2023-09-07
scripts: Fix extra whitespace in VU messages

spencer-lunarg:
Removes multiple whitespace artifacts from the `validusage.json` file


artem-lunarg (approved):
LGTM

PR #6446, opened 2023-09-07
layers: Improve nesting by returning back sooner

spencer-lunarg:
There are some functions that could really make use of an early return at top (or some `continue`) to prevent some deep nesting we have. This commit should have no semantic changes and just ending the function/loop sooner

also plan to add this to `.git-blame-ignore-revs` after

### DIFF
--- a/layers/core_checks/cc_descriptor.cpp
+++ b/layers/core_checks/cc_descriptor.cpp
@@ -777,7 +777,7 @@ bool CoreChecks::ValidateDescriptorSetBindingData(const DescriptorContext &conte
 bool CoreChecks::ValidateDescriptor(const DescriptorContext &context, const DescriptorBindingInfo &binding_info, uint32_t index,
                                     VkDescriptorType descriptor_type, const cvdescriptorset::BufferDescriptor &descriptor) const {
     // Verify that buffers are valid
-    auto buffer = descriptor.GetBuffer();
+    const VkBuffer buffer = descriptor.GetBuffer();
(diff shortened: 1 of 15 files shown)


layers/core_checks/cc_descriptor.cpp
```diff
@@ -777,7 +777,11 @@ bool CoreChecks::ValidateDescriptorSetBindingData(const DescriptorContext &conte
 bool CoreChecks::ValidateDescriptor(const DescriptorContext &context, const DescriptorBindingInfo &binding_info, uint32_t index,
                                     VkDescriptorType descriptor_type, const cvdescriptorset::BufferDescriptor &descriptor) const {
     // Verify that buffers are valid
-    auto buffer = descriptor.GetBuffer();
+    const VkBuffer buffer = descriptor.GetBuffer();
+    if (buffer == VK_NULL_HANDLE) {
+        return false;
+    }
+
     auto buffer_node = descriptor.GetBufferState();
```
artem-lunarg, line 785: Just to confirm it's intentional. Previously the first validation was performed on the code path where `buffer` could be zero.

artem-lunarg (approved):
LGTM. One question.

layers/core_checks/cc_descriptor.cpp
spencer-lunarg: yes intentional, but looking back at this, I disagree with this (and the other) one, going to move the check below

PR #6449, opened 2023-09-08
scripts: Fix struct members not in IsFieldPointer

spencer-lunarg:
Fix typo in script and not populating the `IsFieldPointer` correctly

### DIFF
--- a/layers/vulkan/generated/error_location_helper.cpp
+++ b/layers/vulkan/generated/error_location_helper.cpp
@@ -4183,26 +4183,46 @@ const char* String(Field field) {
 bool IsFieldPointer(Field field) {
     switch (field) {
     case Field::connection:
+    case Field::context:
     case Field::dfb:
+    case Field::displayName:
     case Field::dpy:
+    case Field::hostAddress:
+    case Field::opaqueCaptureDescriptorData:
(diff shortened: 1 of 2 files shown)


aqnuep:
LGTM.

arno-lunarg (approved):
LGTM

I wonder about how efficient it is to have this `IsFieldPointer` logic in such a long case list?
I think that doing a lookup in a sorted `Field` array would be more efficient at this point

spencer-lunarg:
> I wonder about how efficient it is to have this IsFieldPointer logic implemented using a huge case list?

Well this code is only every executed at error logging time, which we have historically put on the back of any performance bottle neck (since in theory it should never be hit if apps are good)

I agree this probably could be done more efficient, but argue that without measuring, the simplicity here out-weighs any real world benefits I imagine we would see in perf

arno-lunarg:
Yes some quick perf test show that the case list is still completely ok

artem-lunarg:
Was curious and checked codegen. It's efficient. For enum value `Field::pDependencyInfo` == 1328 it runs a dozen of instructions in total to get the result (lets see if github supports animated gifs).

is_field_pointer

PR #6463, opened 2023-09-11
Thread safety cleanup and tests

artem-lunarg:
Thread safety cleanup:
* Conventions: `prevCount` -> `prev_count`
* Remove one indentation level in `StartWrite`
* Use boolean variables instead of checking counters for 0/not zero.
* Update comments. Remove comments that do not bring new information.

Add new test suite for multi-threading: thread safety checks and the tests that implement non-trivial threading scenarios.

Add asserts that check that counters do not go negative. Negative value is a bug. In addition to indication it's a mismatch
between Start/Finish it's also leads to broken subsequent validation: the next "Start" operation will not produce positive value and corresponding read/write won't be taken into account against other parallel reads/writes. The plan is to use unsigned values for counters after testing this code for a bit.


artem-lunarg:
Internal CI fail is due to the test rename that is blacklisted . The plan is to push blacklist update (already approved, thanks @spencer-lunarg ) just before pushing this PR to not affect other PRs.

juan-lunarg (approved):
LGTM 👍🏾 

FWIW it would be nice to run tests that contain things like `Threading` multiples times in GHA. Since I have observed that TSAN will sporadically catch issues with certain tests.

See common_ci.py

artem-lunarg:
@juan-lunarg Added to todo list, also remember about the assigned to me issue with TSAN failures.

PR #6464, opened 2023-09-11
Remove ParameterName for Location

spencer-lunarg:
This removes the `ParameterName` in favor for `Location` which was being used for the generated stateless validation

We now get the nicer error messages such as 

> [ VUID-VkSparseMemoryBind-flags-parameter ] | vkQueueBindSparse(): pBindInfo[0].pBufferBinds[0].pBinds[0].flags contains flag bits that are not recognized members of VkSparseMemoryBindFlagBits.

with the cleaner code of `Location`


layers/error_message/error_location.h
```diff
@@ -61,6 +61,10 @@ struct Location {
         Location result(*this, this->structure, sub_field, sub_index, false);
         return result;
     }
+    Location dot(uint32_t sub_index) const {
```
spencer-lunarg, line 64: we have cases were we need to tack on the index when checking the values in the stateless helpers

artem-lunarg: This `dot` looks similar to other overloads that do "append". Can something like `replace_index` better indicate intention?

spencer-lunarg: I have thought a lot of about this

so the usecase for this code is

```c++
    bool ValidateStructTypeArray(const Location &count_loc, const Location &array_loc, /*...*/) const {
        // ...
        for (uint32_t i = 0; i < count; ++i) {
            if (array[i].sType != sType) {
                skip |= LogError(stype_vuid, device, array_loc.dot(i).dot(Field::sType), "must be %s", sTypeName);
            }
        }
    }
```

Parts of me looks at the other `dot` and see it does the same thing to add a `Field` to a `Struct` that was passed along

Part of me is curious if making `uint32_t index` a `const` was a mistake, It might make sense while looping to just update the `index` with a non-const `Location` as you are looping through (instead of having to re-create a new object each iteration)

@jeremyg-lunarg as someone who also used `Location` a lot, any thoughts?

artem-lunarg: With the above example the name makes sense for me now.

artem-lunarg: It looks this `dot` also constructs bigger path by *appending* index to array location, initially I thought I replaces existing index. Revoking my question about naming.

scripts/generators/stateless_validation_helper_generator.py
```diff
@@ -354,10 +349,9 @@ def generateSource(self):
             if struct.sType and struct.version and all(not x.promotedTo for x in struct.extensions):
                 pnext_check += f'''            if (api_version < {struct.version.nameApi}) {{
                 skip |= LogError(
-                           pnext_vuid, instance, loc,
-                           "Includes a pNext pointer (%s) to a VkStructureType ({struct.sType}) which was added in {struct.version.nameApi} but the "
-                           "current effective API version is %s.",
-                           parameter_name.get_name().c_str(), StringAPIVersion(api_version).c_str());
+                           pnext_vuid, instance, loc.dot(Field::pNext),
+                           "Includes a pointer to a VkStructureType ({struct.sType}) which was added in {struct.version.nameApi} but the "
```
arno-lunarg, line 353:
```suggestion
                           "includes a pointer to a VkStructureType ({struct.sType}) which was added in {struct.version.nameApi} but the "
```

scripts/generators/stateless_validation_helper_generator.py
```diff
@@ -382,17 +376,17 @@ def generateSource(self):
                 pnext_check += f'''            if (is_const_param) {{
                 {extension_check}
                         skip |= LogError(
-                               pnext_vuid, instance, loc,
-                               "Includes a pNext pointer (%s) to a VkStructureType ({struct.sType}), but its parent extension "
-                               "{extension.name} has not been enabled.",
-                               parameter_name.get_name().c_str());
+                               pnext_vuid, instance, loc.dot(Field::pNext),
+                               "Includes a pointer to a VkStructureType ({struct.sType}), but its parent extension "
```
arno-lunarg, line 380:
```suggestion
                               "includes a pointer to a VkStructureType ({struct.sType}), but its parent extension "
```

scripts/generators/stateless_validation_helper_generator.py
```diff
@@ -412,12 +406,17 @@ def generateSource(self):
...
-            startIndex = 0 if command.name == 'vkCreateInstance' else 1
-            lines = self.genFuncBody(command.params[startIndex:], command.name, '', '', None, isPhysDevice = command.params[0].type == 'VkPhysicalDevice')
+
+            prototype = (command.cPrototype.split('VKAPI_CALL ')[1])[2:-1]
+            prototype = prototype.replace(')', ',\n    const ErrorObject&                          error_obj)')
+            out.append(f'bool StatelessValidation::PreCallValidate{prototype} const {{\n')
+            out.append('    bool skip = false;\n')
+
+            # Create a copy here to make the logic simpler passing into ValidatePnextStructContents
+            out.append('    [[maybe_unused]] const Location& loc = error_obj.location;\n')
```
arno-lunarg, line 418: Just to make sure, your comment mentions a copy, but `loc` is a reference. Everything fine?

layers/stateless/sl_ray_tracing.cpp
```diff
@@ -913,40 +915,33 @@ bool StatelessValidation::ValidateAccelerationStructureBuildGeometryInfoKHR(
...
-                        kVUIDUndefined);
-                    skip |= ValidateRangedEnum(loc,
-                                               ParameterName("pInfos[%i].pGeometries[%i].geometry.triangles.vertexFormat",
-                                                             ParameterName::IndexVector{i, j}),
+                    skip |= ValidateStructType(geometry_loc.dot(Field::geometry).dot(Field::triangles),
+                                               "VK_STRUCTURE_TYPE_ACCELERATION_STRUCTURE_GEOMETRY_TRIANGLES_DATA_KHR",
+                                               &(pInfos[i].pGeometries[j].geometry.triangles),
+                                               VK_STRUCTURE_TYPE_ACCELERATION_STRUCTURE_GEOMETRY_TRIANGLES_DATA_KHR, false,
+                                               kVUIDUndefined, "VUID-VkAccelerationStructureGeometryTrianglesDataKHR-sType-sType");
+                    skip |= ValidateStructPnext(geometry_loc.dot(Field::geometry).dot(Field::triangles), nullptr,
```
arno-lunarg, line 930: Should it be:
```suggestion
                    skip |= ValidateStructPnext(geometry_loc.dot(Field::geometry).dot(Field::triangles).dot(Field::pNext), nullptr,
```
Or is it handled it `ValidateStructPnext`?

arno-lunarg (approved):
LGTM, mainly looked at (and tried to make sense of) the changes in the script

PR #6468, opened 2023-09-11
cmake: Remove BUILD_LAYER_SUPPORT_FILES

### LINKED ISSUE #6246: Move to ../layers/vulkan/generated/ broke includes.
As you know an #include ".." with double quotes, as opposed to using angle brackets, first searches relative to the directory that the file, that has the `#include`, resides in.

When `vk_format_utils.cpp` was still in `.../layers/vulkan/vk_format_utils.cpp` its `#include "utils/vk_layer_utils.h"` worked as intended because `utils` is in `.../layers/vulkan/`.

However, now that this file was moved into `.../layers/vulkan/generated/` this no longer works.
To correct this, the include should now read: ` #include "../utils/vk_layer_utils.h"` (note the prepended `../`).

See
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/layers/vulkan/generated/vk_format_utils.cpp

Note that this file is "generated", so something else needs to be changed somewhere. Also, this might not be the only place where this change needs to be made (in theory every include using double quotes and some subdirectory that wasn't moved into generated/ as well needs to have added `../` in front of it).

I realize it will still compile if an include path is added to the compilation of projects that use `generated/vk_format_utils.cpp` to where `utils/` sits (aka, where `generated` sits), which is likely something like `/usr/include/vulkan/`, but it would be HARMFUL to demand vulkan users to add that include path. The idea is that one uses `#include <vulkan/....>` in the source code for namespace reasons, and having to add that `vulkan/` path as include directory might break other things.

juan-lunarg:
LunarG/VulkanTools no longer depends on the validation layers.

See https://github.com/LunarG/VulkanTools/pull/1883

BUILD_LAYERS and BUILD_LAYER_SUPPORT_FILES are removed as they are no longer needed.

The corresponding integration tests are also no longer needed.

closes #6246


CMakeLists.txt
```diff
@@ -134,13 +134,12 @@ elseif(MSVC)
     add_compile_options($<$<BOOL:${MSVC_IDE}>:/MP>) # Speed up Visual Studio builds
 endif()
 
-option(BUILD_LAYERS "Build validation layers" ON)
-option(BUILD_LAYER_SUPPORT_FILES "Install VkLayer_utils")
-
-if(BUILD_LAYERS OR BUILD_LAYER_SUPPORT_FILES)
-    add_subdirectory(layers)
+if (BUILD_LAYER_SUPPORT_FILES)
+    message(WARNING "BUILD_LAYER_SUPPORT_FILES is deprecated! See https://github.com/KhronosGroup/Vulkan-Utility-Libraries/issues/18")
```
juan-lunarg, line 138: This warning is for package managers.

spencer-lunarg (approved):

PR #6472, opened 2023-09-12
layers: Run ObjectTracker on pNext structs

spencer-lunarg:
I found if `VkPipelineLibraryCreateInfoKHR::pLibraries` was a bad `VkPipeline` we were crashing inside the layers

I realized we never ran ObjectTracker on any of the extended `pNext` objects

### DIFF
--- a/scripts/generators/object_tracker_generator.py
+++ b/scripts/generators/object_tracker_generator.py
@@ -464,46 +464,73 @@ f'''if (({countName} > 0) && ({prefix}{member.name})) {{
 
             # Handle Structs that contain objects at some level
             elif member.type in self.vk.structs:
+                nested_struct = []
                 struct = self.vk.structs[member.type]
+
                 # Structs at first level will have an object
-                if self.structContainsObject(struct):
-                    # Struct Array
-                    if member.length is not None:
-                        # Update struct prefix
-                        pre_call_validate += f'{indent}if ({prefix}{member.name}) {{\n'
+                contains_object = self.structContainsObject(struct)
+
+                # Struct Array
+                if member.length is not None:
+                    # Update struct prefix
+                    nested_struct.append(f'{indent}if ({prefix}{member.name}) {{\n')
+                    indent = incIndent(indent)
+                    nested_struct.append(f'{indent}for (uint32_t {index} = 0; {index} < {prefix}{member.length}; ++{index}) {{\n')
+                    indent = incIndent(indent)
+                    new_error_loc = f'{index}_loc'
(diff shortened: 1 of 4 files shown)


tests/unit/graphics_library.cpp
```diff
@@ -1771,3 +1771,84 @@ TEST_F(NegativeGraphicsLibrary, IncompatibleLayouts) {
...
+        frag_out_lib.pipeline_,
+    };
+    auto link_info = LvlInitStruct<VkPipelineLibraryCreateInfoKHR>();
+    link_info.libraryCount = size(libraries);
+    link_info.pLibraries = libraries;
+
+    auto exe_pipe_ci = LvlInitStruct<VkGraphicsPipelineCreateInfo>(&link_info);
+    exe_pipe_ci.layout = pre_raster_lib.gp_ci_.layout;
+    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkPipelineLibraryCreateInfoKHR-pLibraries-parameter");
+    vk_testing::Pipeline exe_pipe(*m_device, exe_pipe_ci);
```
spencer-lunarg, line 1811: > [ VUID-VkPipelineLibraryCreateInfoKHR-pLibraries-parameter ] Object 0: handle = 0x555b284aafe0, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0xb36aee50 | vkCreateGraphicsPipelines(): pCreateInfos[0].pNext\<VkPipelineLibraryCreateInfoKHR\>.pLibraries[2] Invalid VkPipeline Object 0x0.

which is what we want!

scripts/generators/object_tracker_generator.py
```diff
@@ -464,46 +464,73 @@ def validateObjects(self, members: list[Member], indent: str, prefix: str, array
...
+                    new_prefix = f'{prefix}{member.name}.'
+                    new_error_loc = f'{member.name}_loc'
+                    nested_struct.append(f'{indent}[[maybe_unused]] const Location {new_error_loc} = {errorLoc}.dot(Field::{member.name});\n')
+                    # Process sub-structs
+
+                if contains_object:
+                    nested_struct.append(self.validateObjects(struct.members, indent, new_prefix, arrayIndex, member.type, topCommand, new_error_loc, False))
+
+                contains_pNext = False
+                if struct.extendedBy:
```
spencer-lunarg, line 504: this is the new logic added, I had to rework this part of the script because currently this code is setup were we need to setup the `if (pCreateInfo != nullptr)` THEN we can call into the recursive `validateObjects` call

artem-lunarg (approved):

PR #6474, opened 2023-09-12
layers: Combine and cleanup CooperativeMatrix

spencer-lunarg:
- Combines the NV and KHR logic together
- Use more of the logic already done in `SPIRV_MODULE_STATE`
- Improve error messages to provide more info and use `Location`
- Rework tests so we can test with MockICD
- Add Max Profile support so we run these tests on GitHub CI

### DIFF
--- a/tests/unit/shader_cooperative_matrix_positive.cpp
+++ b/tests/unit/shader_cooperative_matrix_positive.cpp
@@ -16,6 +16,35 @@
 #include "../framework/layer_validation_tests.h"
 #include "../framework/pipeline_helper.h"
 
+const char *vkComponentTypeToGLSL(VkComponentTypeKHR type) {
+    switch (type) {
+        case VK_COMPONENT_TYPE_FLOAT16_KHR:
+            return "float16_t";
+        case VK_COMPONENT_TYPE_FLOAT32_KHR:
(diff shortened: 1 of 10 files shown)


tests/unit/shader_cooperative_matrix_positive.cpp
```diff
@@ -16,6 +16,35 @@
 #include "../framework/layer_validation_tests.h"
 #include "../framework/pipeline_helper.h"
 
+const char *vkComponentTypeToGLSL(VkComponentTypeKHR type) {
```
juan-lunarg, line 19: I wonder if this type of functionality would be useful in `vk_format_utils.h`...

juan-lunarg (approved):

tests/unit/shader_cooperative_matrix_positive.cpp
spencer-lunarg: no, `vk_format_utils` should be focused on `VK_FORMAT`

This is for generating GLSL, which I am still not a fan we are doing in this test, but didn't want to spend too much time with these tests at the moment

juan-lunarg: 👍🏾

Issue #6461, opened 2023-09-09
Validation layer crash in vkCmdBuildAccelerationStructuresKHR


Onhi:
**Environment:**
 - OS: Windows 11 (21H2 - 22000.2295)
 - GPU: nVidia 2080 Super
 - SDK or header version if building from repo: 1.3.241
 - Options enabled (synchronization, best practices, etc.): VK_LAYER_KHRONOS_validation (default settings)

It looks like there is an issue related to the tracking of scratch data memory.
I mean this data specifically: `VkAccelerationStructureBuildGeometryInfoKHR::scratchData`.

To build acceleration structures I create vkbuffers to hold scratch data.
The buffers are temporary so I reuse the underliying memory for several of them.

Proper synchronization and resource life time management is respected yet, the validation layer triggers buffer lifetime errors and eventually crash on build acceleration structure call.

Typically it takes this form:

> frame 0:
> - wait fence -2
> - clean up frame -2
> - create buffer 0 at address A
> - build acceleration structure X (using buffer 0's device memroy address A)
> - a barrier ensuring acceleration structure X build is complete
> - ...
> - signal fence 0
> 
> frame 1:
> - wait fence -1
> - clean up frame -1
> - create buffer 1 at address A
> - build acceleration structure Y (using buffer 1's device memroy address A)
> - a barrier ensuring acceleration structure Y build is complete
> - ...
> - signal fence 1
> 
> frame 2:
> - wait fence 0
> - clean up frame 0
> - - release buffer 0 -> **This fires up the validation layer! Probaly because buffer 0 & 1 shares the same memory address A** (**message 1**)
> - create buffer 2 at address A
> - build acceleration structure Z (using buffer 1) -> **Validation layer crash here** (**message 2**) (**callstack**)

_Important note_: if I give all the buffers their own memory region, there is no validation message and no crash.

**Expected behavior**

Proper resource lifetime tracking and no validation layer crashes.

**Valid Usage ID**

[...]

PR #6475, opened 2023-09-12
layers: Fix issue 6461

arno-lunarg:
- Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6461

The situation described in the issue showed that it is incorrect to add buffers obtained through a call do `GetBuffersByAddress` as children to a command buffer, given that there is no 1 to 1 mapping between a buffer and a device address. It can lead to false positives when tracking resources life span.

Add a regression test, `PositiveRayTracing.AccelerationStructuresReuseScratchMemory`:
It simulates 3 frames, where in each frame a new acceleration structure is built. Building an acceleration structure requires a _scratch_ buffer. Each frames has its own buffer, but they are all bound to the same memory.
When starting the last frame, the fence from the first frame is waited upon, then resources from frame 0 are released. It used to trigger `VUID-vkDestroyBuffer-buffer-00922`.
The issue was that when recording a acceleration structure build command, any buffer indirectly mentioned through a device address was added using a call to `GetBuffersByAddress`. When recording the build happening on frame 1, given that all scratch buffers have the same base device address, `scratch_buffer_frame_0` will *also* be added as a child to `cmd_buffer_frame_1`. So when destroying it, at the beginning of frame 2, since frame 1 is still in flight, `scratch_buffer_frame_0` is still considered in use, so 00922 is triggered.

### DIFF
--- a/layers/state_tracker/cmd_buffer_state.h
+++ b/layers/state_tracker/cmd_buffer_state.h
@@ -469,13 +469,6 @@ class CMD_BUFFER_STATE : public REFCOUNTED_NODE {
         AddChild(base);
     }
 
-    void AddChildren(vvl::span<BUFFER_STATE *> &child_nodes) {
-        for (auto &child_node : child_nodes) {
-            auto child_node_shared_ptr = child_node->shared_from_this();
-            AddChild(child_node_shared_ptr);
-        }
-    }
-
     void RemoveChild(std::shared_ptr<BASE_NODE> &base_node);
     template <typename StateObject>
     void RemoveChild(std::shared_ptr<StateObject> &child_node) {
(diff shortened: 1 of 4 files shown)


artem-lunarg (approved):
LGTM! Nice fix. I understood this in a way that we can reliably track resources when their handles are part of the api, and we might have troubles when resources are accessed indirectly, i.e. by just accessing memory.

spencer-lunarg (approved):

arno-lunarg:
> LGTM! Nice fix. I understood this in a way that we can reliably track resources when their handles are part of the api, and we might have troubles when resources are accessed indirectly, i.e. by just accessing memory.

Yes, it makes me wonder why you need to have a buffer to get an address, when what you really get is just a virtual memory address. We truly need a `vkGetMemoryAddress` instead of `vkGetBufferDeviceAddress` I think. 
Also, even though I basically remove some buffers life span tracking, memory life span tracking has not been touched, so you will still get an error when destroying memory used by an in flight command buffer, which is what you are truly interested in.

Issue #6473, opened 2023-09-12
Missing check for VUID-VkGraphicsPipelineCreateInfo-renderPass-06055


lhiuming:
**Describe the situation in which you encountered the missing validation**
Issue description: VLL didn't report any error when create graphics pipeline with `pColorBlendState->attachmentCount` equal to 1 and `VkPipelineRenderingCreateInfo::colorAttachmentCount` equal to 0, while the spec say they must be equal. 

**Valid Usage IDs requested**
VUID-VkGraphicsPipelineCreateInfo-renderPass-06055

**Additional Context**
The result is that everything looks fine in the rendering except that color blending is not working. Environment:
Vulkan SDK: 1.3.261.1
nVIDIA driver: 537.13

spencer-lunarg:
I just recently edited this code as I noticed it looked potentially wrong.

Let me write up some better tests to confirm, but this should be fixed as of now

(also seems related to https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4856)

Issue #4856, opened 2022-11-24
No validation message for mismatched attachment count


baldurk:
**Describe the Issue**
When creating a graphics pipeline state with `VkPipelineRenderingCreateInfo::colorAttachmentCount` being 1 but `VkPipelineColorBlendStateCreateInfo::attachmentCount` being 0 I encountered a driver crash but validation did not catch this error.

Looking into it the check for `VUID-VkGraphicsPipelineCreateInfo-renderPass-06055` requiring that color blend attachment count == dynamic rendering attachment count is inside a loop over the min() of the number of attachments, so if `color_blend_state->attachmentCount` is zero and `numberColorAttachments` is not then it will never do the check. Since this check doesn't depend on the individual attachments I think it should happen outside

See core_validation.cpp line 1872 or so, inside the loop at line 1869.

As a side note there is `VUID-VkGraphicsPipelineCreateInfo-renderPass-07609` which is the equivalent non-dynamic rendering VUID that should be checked here, but it doesn't seem to be yet - I see it's mentioned in #4727.

**Valid Usage ID**
No validation messages but a driver crash, when looking into it I see that `VUID-VkGraphicsPipelineCreateInfo-renderPass-06055` was violated but the validation layers didn't catch it.

**Environment:**
 - OS: Windows 10
 - GPU: nVidia 2080 Ti
 - SDK or header version if building from repo: built from commit a978a7ce4e50b41a617655af6fa2f756bc045459 from header 1.3.234
 - Options enabled (synchronization, best practices, etc.): None

**Additional context**
None that I can think of.

PR #6476, opened 2023-09-12
layers: Fix DynamicRendering AttachmentCount check

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6473
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4856

as stated in https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4856 the issue was in `safe_VkGraphicsPipelineCreateInfo::initialize` we go 

```cpp
    if (in_struct->pColorBlendState && ((has_rasterization && uses_color_attachment) || is_graphics_library))
        pColorBlendState = new safe_VkPipelineColorBlendStateCreateInfo(in_struct->pColorBlendState);
    else
        pColorBlendState = nullptr; // original pColorBlendState pointer ignored
```

so it was never checking the `pColorBlendState` , so now it was moved to Stateless where it gets the original information to check

### DIFF
--- a/tests/unit/dynamic_rendering.cpp
+++ b/tests/unit/dynamic_rendering.cpp
@@ -131,7 +131,6 @@ TEST_F(NegativeDynamicRendering, CommandDraw) {
     VkPipelineObj pipe(m_device);
     pipe.AddShader(&vs);
     pipe.AddShader(&fs);
-    pipe.AddDefaultColorAttachment();
     pipe.SetDepthStencil(&ds_state);
 
     VkDescriptorSetLayoutBinding dslb = {0, VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT, 1, VK_SHADER_STAGE_FRAGMENT_BIT, nullptr};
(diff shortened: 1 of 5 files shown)


tests/unit/dynamic_rendering.cpp
```diff
@@ -834,11 +833,11 @@ TEST_F(NegativeDynamicRendering, GraphicsPipelineCreateInfo) {
     const VkPipelineLayoutObj pl(m_device);
     VkPipelineObj pipe(m_device);
 
-    VkPipelineColorBlendAttachmentState color_blend_attachment_state = {};
+    std::vector<VkPipelineColorBlendAttachmentState> color_blend_attachment_state(2);
```
juan-lunarg, line 836: Why std::vector? Looks like std::array would work fine.

juan-lunarg (approved):

tests/unit/dynamic_rendering.cpp
spencer-lunarg: I copied this from some another test, I think we need a cleanup of using `std::array` for our tests (can add that to my to-do list)

Issue #5580, opened 2023-04-05
Add validation for VK_EXT_shader_object


ncesario-lunarg:
Add missing VUIDs related to `VK_EXT_shader_object`:

- [x] CPU runtime validation
- [ ] GPU-AV validation

https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5579 should be completed before this work is done.

PR #6478, opened 2023-09-12
Add gpu assisted validation for shader objects

ziga-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5580

### DIFF
--- a/tests/unit/debug_printf.cpp
+++ b/tests/unit/debug_printf.cpp
@@ -330,6 +330,7 @@ TEST_F(NegativeDebugPrintf, BasicUsage) {
         m_errorMonitor->VerifyFound();
     }
 }
+
 TEST_F(NegativeDebugPrintf, MeshTaskShaders) {
     TEST_DESCRIPTION("Test debug printf in mesh and task shaders.");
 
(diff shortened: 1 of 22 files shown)


tests/unit/debug_printf.cpp
```diff
@@ -1063,3 +1064,379 @@ TEST_F(NegativeDebugPrintf, GPLFragmentIndependentSets) {
     vk::QueueWaitIdle(m_device->m_queue);
     m_errorMonitor->VerifyFound();
 }
+
+static void SetDefaultDynamicStates(VkCommandBuffer cmdBuffer) {
```
spencer-lunarg, line 1068: this might be worth a some helper in the common test framework, but should be obvious it for `VkShaderEXT`

tests/unit/gpu_av.cpp
```diff
@@ -798,12 +798,67 @@ TEST_F(VkGpuAssistedLayerTest, GpuBufferOOB) {
...
+}
+
 void VkGpuAssistedLayerTest::ShaderBufferSizeTest(VkDeviceSize buffer_size, VkDeviceSize binding_offset, VkDeviceSize binding_range,
                                                   VkDescriptorType descriptor_type, const char *fragment_shader,
-                                                  const char *expected_error) {
+                                                  const char *expected_error, bool shader_objects) {
     SetTargetApiVersion(VK_API_VERSION_1_1);
 
     VkValidationFeaturesEXT validation_features = GetValidationFeatures();
+    if (shader_objects||1) {
```
spencer-lunarg, line 858: was this `|| 1` left in from testing?

tests/unit/gpu_av.cpp
```diff
@@ -860,6 +926,15 @@ void VkGpuAssistedLayerTest::ShaderBufferSizeTest(VkDeviceSize buffer_size, VkDe
         "      gl_Position = vec4(vertices[gl_VertexIndex % 3], 0.0, 1.0);\n"
         "}\n";
 
+    vk_testing::Shader *vso = nullptr;
```
spencer-lunarg, line 929: why not just create these regardless, then you won't need the `delete` at the end

spencer-lunarg (approved):
some small questions/ideas, but thanks for taking the old PR and actually making it work!

tests/unit/gpu_av.cpp
ziga-lunarg: Yes

tests/unit/gpu_av.cpp
ziga-lunarg: VK_EXT_shader_object isn't enabled in all cases

spencer-lunarg: ok, for now it is fine

jeremyg-lunarg (approved):

Tony-LunarG (approved):
LGTM

PR #6480, opened 2023-09-13
layers: Run Clang-Format on generated files

spencer-lunarg:
Applied `clang-format` to all the generated files

Some spots look better line-by-line and inserted `// clang-format off` / `on` around there

The goal is to remove the python logic in the more complex code (stateless validation, object tracker, etc) that is used purely to present well formatted code

This commit will be put in `.git-blame-ignore-revs`


artem-lunarg (approved):
LGTM. It looks the failures are due to new formatting.

spencer-lunarg:
> LGTM. It looks the failures are due to new formatting.

Yes, seems I forgot to push the code up that ran `clang-format` after the code gen. Our CI runs `generate_source.py` and checks if there is a difference (mainly to catch people updating the generated file and not the scripts)

arno-lunarg:
image
Is this your record?

PR #6484, opened 2023-09-14
layers: Add Location to Descriptor Draw time

spencer-lunarg:
Pass the `Location` into the `ValidateDescriptor` calls

There are 2 ways to hit there, first with `vkCmdDraw` type calls, but also with `vkQueueSubmit` if using `UPDATE_AFTER_BIND` descriptors

before we just printed `vkQueueSubmit: ...` but now we list which command buffer in the queue submit called it

> Validation Error: [ VUID-vkCmdDraw-viewType-07752 ]  vkQueueSubmit(): pSubmits[0].pCommandBuffers[0] Descriptor set VkDescriptorSet 0xd897d90000000016[] in binding #0 index 0 ImageView type is VK_IMAGE_VIEW_TYPE_2D but the OpTypeImage has (Dim = 3D) and (Arrrayed = 0).


artem-lunarg:
> VkDescriptorSet 0xd897d90000000016[] in binding #0 index 0

My personal association of "binding" is more strong with a descriptor binding within a descriptor set (like in `VkDescriptorSetLayoutBinding::binding`), and descriptor sets sounds more usual when it's referenced by index (`set 2`). Thoughts? Is this more common or not?

spencer-lunarg:
I though `Index` was the array inside the binding (`VkDescriptorSetLayoutBinding::descriptorCount`)

I was not a fan of the wording we had, but it was consistent at least (so easy to replace/fix all them all)

I guess what is your ideal error message for describing the location of the descriptor?

artem-lunarg:
My confusion is mostly that descriptor set and descriptor are close together in the same sentence and and the meaning of this is a bit unclear, at least for me: `Descriptor set VkDescriptorSet 0xd897d90000000016[] in binding #0 index 0`.

The original formulation `Descriptor set %s encountered the following validation error at %s time: Descriptor in "binding #%" PRIu32 " index %" PRIu32 "` separates concepts. At first it mentions descriptor set in question and then descriptor in that descriptor set.

I guess the style of original formulation works for me. I agree, that *index* is better to use for descriptor location and *binding* as the index of the descriptor set, as the general graphics terminology. But in Vulkan for some reason word *Binding* is often part of types that define location of specific descriptors within a set.

spencer-lunarg:
ok, I see 

> Descriptor set in binding #0 index 0.

is wrong, it could say

> The descriptor in VkDescriptorSet 0xd897d90000000016[], binding 0, index 0, ....

layers/core_checks/cc_descriptor.cpp
```diff
@@ -1276,13 +1264,12 @@ bool CoreChecks::ValidateDescriptor(const DescriptorContext &context, const Desc
...
                     const LogObjectList objlist(set, sampler_state->sampler(), image_view_state->image_view());
-                    return LogError(
-                        objlist, context.vuids.img_filter_cubic_02693,
-                        "%s: Descriptor set %s Sampler(%s)is set to use VK_FILTER_CUBIC_EXT while the VK_IMG_filter_cubic "
-                        "extension is enabled, but image view (%s) has an invalid imageViewType (%s).",
-                        String(context.command), FormatHandle(set).c_str(), FormatHandle(sampler_state->sampler()).c_str(),
-                        FormatHandle(image_view_state->image_view()).c_str(),
-                        string_VkImageViewType(image_view_state->create_info.viewType));
+                    return LogError(context.vuids.img_filter_cubic_02693, objlist, context.loc,
+                                    "Descriptor set %s Sampler(%s)is set to use VK_FILTER_CUBIC_EXT while the VK_IMG_filter_cubic "
```
arno-lunarg, line 1268:
```suggestion
                                    "Descriptor set %s Sampler(%s) is set to use VK_FILTER_CUBIC_EXT while the VK_IMG_filter_cubic "
```

arno-lunarg (approved):

layers/core_checks/cc_descriptor.cpp
```diff
@@ -1361,19 +1353,21 @@ bool CoreChecks::ValidateDescriptor(const DescriptorContext &context, const Desc
...
+                return LogError(context.vuids.storage_image_write_texel_count_08796, objlist, context.loc,
+                                "the descriptor (%s, binding %" PRIu32 ", index %" PRIu32
+                                ") VkImageView is mapped to a OpImage format of VK_FORMAT_A8_UNORM_KHR, but the OpImageWrite Texel "
+                                "operand only contains %" PRIu32 " components.",
+                                FormatHandle(set).c_str(), binding, index, texel_component_count);
             }
         } else if (texel_component_count < format_component_count) {
             auto set = context.descriptor_set.GetSet();
             const LogObjectList objlist(set, image_view);
-            return LogError(device, context.vuids.storage_image_write_texel_count_08795,
```
spencer-lunarg, line 1372: from

> vkCmdDispatch(): OpImageWrite Texel operand only contains 3 components, but the VkImageView is mapped to a OpImage format of VK_FORMAT_R8G8B8A8_UINT has 4 components.

to

> vkCmdDispatch(): the descriptor (VkDescriptorSet 0xe7f79a0000000005[], binding 2, index 0) VkImageView is mapped to a OpImage format of VK_FORMAT_R8G8B8A8_UINT which has 4 components, but the OpImageWrite Texel operand only contains 3 components.

which is a nice upgrade, next is to add pipeline and shader stage info... (thinking out loud) when it comes to these VUs that interact with 3 or 4 objects, we need a clean way to print that information out

to

> vkCmdDispatch(): the descriptor (VkDescriptorSet 0xe7f79a0000000005[], binding 2, index 0) VkImageView is mapped to a OpImage format of VK_FORMAT_R8G8B8A8_UINT which has 4 components, but the OpImageWrite Texel operand only contains 3 components.

PR #6485, opened 2023-09-14
layers: Add Location to remaining Ray Tracing

spencer-lunarg:
Improves the rest of the Ray Tracing messages to use `Location`

PR #6489, opened 2023-09-14
layers: Cleanup RT error messages

arno-lunarg:
- Some cleanups following https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6485

### DIFF
--- a/layers/stateless/sl_ray_tracing.cpp
+++ b/layers/stateless/sl_ray_tracing.cpp
@@ -307,11 +307,11 @@ bool StatelessValidation::manual_PreCallValidateCreateRayTracingPipelinesNV(
         auto feedback_struct = LvlFindInChain<VkPipelineCreationFeedbackCreateInfoEXT>(pCreateInfos[i].pNext);
         if ((feedback_struct != nullptr) && (feedback_struct->pipelineStageCreationFeedbackCount != 0) &&
             (feedback_struct->pipelineStageCreationFeedbackCount != pCreateInfos[i].stageCount)) {
-            skip |=
-                LogError("VUID-VkRayTracingPipelineCreateInfoNV-pipelineStageCreationFeedbackCount-06651", device, create_info_loc,
-                         "VkPipelineCreationFeedbackEXT::pipelineStageCreationFeedbackCount"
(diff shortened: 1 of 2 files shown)


layers/stateless/sl_ray_tracing.cpp
```diff
@@ -321,102 +321,95 @@ bool StatelessValidation::manual_PreCallValidateCreateRayTracingPipelinesNV(
...
             skip |= LogError("VUID-VkRayTracingPipelineCreateInfoNV-flags-03459", device, create_info_loc.dot(Field::flags),
-                             "must not include "
-                             "VK_PIPELINE_CREATE_RAY_TRACING_NO_NULL_CLOSEST_HIT_SHADERS_BIT_KHR.");
+                             "is %s.", string_VkPipelineCreateFlags(pCreateInfos[i].flags).c_str());
         }
         if (pCreateInfos[i].flags & VK_PIPELINE_CREATE_RAY_TRACING_NO_NULL_MISS_SHADERS_BIT_KHR) {
             skip |= LogError("VUID-VkRayTracingPipelineCreateInfoNV-flags-03460", device, create_info_loc.dot(Field::flags),
-                             "must not include "
-                             "VK_PIPELINE_CREATE_RAY_TRACING_NO_NULL_MISS_SHADERS_BIT_KHR.");
+                             "is %s", string_VkPipelineCreateFlags(pCreateInfos[i].flags).c_str());
```
artem-lunarg, line 388: nit: missing dot in the end of the message (and few other vuids)

artem-lunarg (approved):
LGTM. One nit.

layers/stateless/sl_ray_tracing.cpp
spencer-lunarg: honestly we should just do this in the `LogError` logic, check if the last item is a period, if not add it.... I like having it because it clearly separates the error message from the Spec text following... but easily we are missing these all over

(so happy to do that logic and save you the nit fixing here Arno :smile: )

spencer-lunarg (approved):

layers/stateless/sl_ray_tracing.cpp
arno-lunarg: Thank you Sensei!

PR #6490, opened 2023-09-14
Fix crash when no stage entrypoint

Tony-LunarG:
Fixes crashes in some of the shader_object CTS tests


layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -1023,7 +1023,7 @@ bool CoreChecks::ValidateDrawDynamicStateShaderObject(const LAST_BOUND_STATE& la
     if (cb_state.dynamic_state_status.cb[CB_DYNAMIC_STATE_ALPHA_TO_COVERAGE_ENABLE_EXT] &&
         cb_state.dynamic_state_value.alpha_to_coverage_enable) {
         const auto fragment_shader_stage = last_bound_state.GetShaderState(ShaderObjectStage::FRAGMENT);
-        if (!fragment_shader_stage->entrypoint->has_alpha_to_coverage_variable) {
+        if (fragment_shader_stage->entrypoint && !fragment_shader_stage->entrypoint->has_alpha_to_coverage_variable) {
```
artem-lunarg, line 1026: nit: if shader stage can be null check for it too, similar condition on line 1008 does this

artem-lunarg (approved):
LGTM. One nit.

layers/core_checks/cc_cmd_buffer_dynamic.cpp
Tony-LunarG: Yep, good point.  Done

PR #6491, opened 2023-09-14
layers: Remove an extra whitespace in the message for VUID 00171

absolutelynothelix:


charles-lunarg (approved):

spencer-lunarg (approved):
Thanks for making the quick PR, this made me realize we have other spots like this as well (which I plan to fix after)

PR #6493, opened 2023-09-15
Make whitespace logic simpler

spencer-lunarg:
Part 2 of https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6480

The goal is to help improve the readability of our scripts now that most places don't need to worry about indenting and whitespace

### DIFF
--- a/scripts/generators/best_practices_generator.py
+++ b/scripts/generators/best_practices_generator.py
@@ -126,10 +126,10 @@ class BestPracticesOutputGenerator(BaseGenerator):
             out.extend([f'#ifdef {command.protect}\n'] if command.protect else [])
             prototype = command.cPrototype.split("VKAPI_CALL ")[1]
             prototype = f'void PostCallRecord{prototype[2:]}'
-            prototype = prototype.replace(');', ',\n    const RecordObject&                         record_obj) {\n')
+            prototype = prototype.replace(');', ', const RecordObject& record_obj) {\n')
             prototype = prototype.replace(') {', ') override;\n')
             if command.name in self.extra_parameter_list:
-                prototype = prototype.replace(')', ',\n    void*                                       state_data)')
+                prototype = prototype.replace(')', ', void* state_data)')
             out.append(prototype)
             out.extend([f'#endif // {command.protect}\n'] if command.protect else [])
 
@@ -176,25 +176,27 @@ class BestPracticesOutputGenerator(BaseGenerator):
             out.extend([f'#ifdef {command.protect}\n'] if command.protect else [])
             prototype = command.cPrototype.split("VKAPI_CALL ")[1]
             prototype = f'void BestPractices::PostCallRecord{prototype[2:]}'
-            prototype = prototype.replace(');', ',\n    const RecordObject&                         record_obj) {\n')
+            prototype = prototype.replace(');', ', const RecordObject& record_obj) {\n')
             if command.name in self.extra_parameter_list:
-                prototype = prototype.replace(')', ',\n    void*                                       state_data)')
+                prototype = prototype.replace(')', ', void* state_data)')
             out.append(prototype)
 
-            out.append(f'    ValidationStateTracker::PostCallRecord{command.name[2:]}({params});\n')
+            out.append(f'ValidationStateTracker::PostCallRecord{command.name[2:]}({params});\n')
             if command.name in self.manual_postcallrecord_list:
-                out.append(f'    ManualPostCallRecord{command.name[2:]}({params});\n')
+                out.append(f'ManualPostCallRecord{command.name[2:]}({params});\n')
 
             if hasNonVkSuccess(command.successCodes):
-                out.append('    if (record_obj.result > VK_SUCCESS) {\n')
-                out.append('        LogPositiveSuccessCode(record_obj);\n')
-                out.append('        return;\n')
-                out.append('    }\n')
+                out.append(
+                    '''if (record_obj.result > VK_SUCCESS) {
+                        LogPositiveSuccessCode(record_obj);
+                        return;
+                    }''')
 
             if command.errorCodes is not None:
-                out.append('    if (record_obj.result < VK_SUCCESS) {\n')
-                out.append('        LogErrorCode(record_obj);\n')
-                out.append('    }\n')
+                out.append(
+                    '''if (record_obj.result < VK_SUCCESS) {
+                        LogErrorCode(record_obj);
+                    }''')
 
             out.append('}\n')
             out.extend([f'#endif // {command.protect}\n'] if command.protect else [])
--- a/scripts/generators/best_practices_generator.py
+++ b/scripts/generators/best_practices_generator.py
@@ -73,26 +73,26 @@ class BestPracticesOutputGenerator(BaseGenerator):
 
     def generate(self):
         self.write(f'''// *** THIS FILE IS GENERATED - DO NOT EDIT ***
-// See {os.path.basename(__file__)} for modifications
+            // See {os.path.basename(__file__)} for modifications
 
-/***************************************************************************
-*
-* Copyright (c) 2015-2023 The Khronos Group Inc.
-* Copyright (c) 2015-2023 Valve Corporation
-* Copyright (c) 2015-2023 LunarG, Inc.
-*
-* Licensed under the Apache License, Version 2.0 (the "License");
-* you may not use this file except in compliance with the License.
-* You may obtain a copy of the License at
-*
-*     http://www.apache.org/licenses/LICENSE-2.0
-*
-* Unless required by applicable law or agreed to in writing, software
-* distributed under the License is distributed on an "AS IS" BASIS,
-* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-* See the License for the specific language governing permissions and
-* limitations under the License.
-****************************************************************************/\n''')
+            /***************************************************************************
+            *
+            * Copyright (c) 2015-2023 The Khronos Group Inc.
+            * Copyright (c) 2015-2023 Valve Corporation
+            * Copyright (c) 2015-2023 LunarG, Inc.
+            *
+            * Licensed under the Apache License, Version 2.0 (the "License");
+            * you may not use this file except in compliance with the License.
+            * You may obtain a copy of the License at
+            *
+            *     http://www.apache.org/licenses/LICENSE-2.0
+            *
+            * Unless required by applicable law or agreed to in writing, software
+            * distributed under the License is distributed on an "AS IS" BASIS,
+            * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+            * See the License for the specific language governing permissions and
+            * limitations under the License.
+            ****************************************************************************/\n''')
         self.write('// NOLINTBEGIN') # Wrap for clang-tidy to ignore
 
         # Build additional pass of functions to ignore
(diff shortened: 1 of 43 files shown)


scripts/generators/command_validation_generator.py
```diff
@@ -64,11 +64,7 @@ def generateSource(self):
 #include "error_message/logging.h"
 #include "core_checks/core_validation.h"
 
-enum CMD_SCOPE_TYPE {
```
spencer-lunarg, line 67: spots like this are me just porting the new clang-formatted output back to the scripts

I expect this to never be a 1-for-1 match, but honestly the issue with a lot of these spots is that we are still having large amount of C++ code in a python string to start

scripts/generators/thread_safety_generator.py
```diff
@@ -127,9 +127,8 @@ def makeThreadUseBlock(self, command: Command, start: bool = False, finish: bool
             if param.externSyncPointer:
                 if param.length:
                     # Externsync can list pointers to arrays of members to synchronize
-                    out.append(f'    if ({param.name}) {{\n')
-                    out.append(f'        for (uint32_t index = 0; index < {param.length}; index++) {{\n')
-                    second_indent = '    '
+                    out.append(f'if ({param.name}) {{\n')
+                    out.append(f'for (uint32_t index = 0; index < {param.length}; index++) {{\n')
```
artem-lunarg, line 131: nit: one level of indentation for loop still might be helpful

artem-lunarg (approved):

scripts/generators/thread_safety_generator.py
arno-lunarg: clang-format will add it back I think?

scripts/generators/layer_chassis_dispatch_generator.py
```diff
@@ -195,36 +191,34 @@ def generateSource(self):
...
-            out.append(f'                    safe_{struct.name} *safe_struct = reinterpret_cast<safe_{struct.name} *>(cur_pnext);\n')
+            out.append(f'case {struct.sType}: {{\n')
+            out.append(f'    safe_{struct.name} *safe_struct = reinterpret_cast<safe_{struct.name} *>(cur_pnext);\n')
             out.append(api_pre)
-            out.append('                } break;\n')
+            out.append('} break;\n')
             out.extend([f'#endif // {struct.protect}\n'] if struct.protect else [])
             out.append('\n')
 
         out.append('''            default:
```
arno-lunarg, line 206:
```suggestion
        out.append('''default:
```

arno-lunarg (approved):
Nice to see better generated code readability. And moving away from manual indentation in the scripts is VERY welcome

scripts/generators/thread_safety_generator.py
artem-lunarg: yes, the point of the comment that it’s still good to have idea of c++ code structure. Other parts still do relative indentation, this one looked like typo

scripts/generators/extension_helper_generator.py
```diff
@@ -205,73 +201,73 @@ def generate(self):
...
-       InstanceInfo(ExtEnabled InstanceExtensions::* state_, const InstanceReqVec requirements_): state(state_), requirements(requirements_) {}
-       ExtEnabled InstanceExtensions::* state;
-       InstanceReqVec requirements;
-    };
-
-    typedef vvl::unordered_map<std::string,InstanceInfo> InstanceInfoMap;
-    static const InstanceInfoMap &get_info_map() {
-        static const InstanceInfoMap info_map = {
-''')
+        struct InstanceReq {
```
jzulauf-lunarg, line 204: I'm torn.  Not over indenting means manageable line lengths... but I think it would be clearer indented from `out.append('''`

Thoughts?

scripts/generators/pnext_chain_extraction_generator.py
```diff
@@ -126,35 +126,35 @@ class PnextChainScopedAdd {
 
     def generateSource(self):
         out = []
-        out.append(f'''
+        out.append('''
```
jzulauf-lunarg, line 129: Why is this left outdented?  Later change, or ... ?

scripts/generators/extension_helper_generator.py
spencer-lunarg: I have no preference (other then I fix it consistently everywhere)

it is currently

```python
    out.append('''
    struct InstanceReq {
        ....
    }'''
```

What did you want instead exactly?

scripts/generators/pnext_chain_extraction_generator.py
spencer-lunarg: I think I just missed some, mainly at the time of each file... I guess no reason and better off just using the new system here as well

scripts/generators/layer_chassis_dispatch_generator.py
```diff
@@ -195,36 +191,34 @@ def generateSource(self):
...
-            indent = '                '
-            (api_decls, api_pre, api_post) = self.uniquifyMembers(struct.members, indent, 'safe_struct->', 0, False, False, False)
+            (api_decls, api_pre, api_post) = self.uniquifyMembers(struct.members, 'safe_struct->', 0, False, False, False)
             # Only process extension structs containing handles
             if not api_pre:
                 continue
             out.extend([f'#ifdef {struct.protect}\n'] if struct.protect else [])
-            out.append(f'            case {struct.sType}: {{\n')
-            out.append(f'                    safe_{struct.name} *safe_struct = reinterpret_cast<safe_{struct.name} *>(cur_pnext);\n')
+            out.append(f'case {struct.sType}: {{\n')
```
lunarpapillo, line 199: This sequence of `out.append()` could be replaced with a single `out.extend([...])`, or with a multiline string, to match what is done elsewhere.

scripts/generators/safe_struct_generator.py
```diff
@@ -916,52 +916,60 @@ def qfi_construct(item, member):
                             if m_type == 'char':
                                 # Create deep copies of strings
                                 if member.length:
-                                    copy_strings += f'    char **tmp_{member.name} = new char *[in_struct->{member.length}];\n'
-                                    copy_strings += f'    for (uint32_t i = 0; i < {member.length}; ++i) {{\n'
-                                    copy_strings += f'        tmp_{member.name}[i] = SafeStringCopy(in_struct->{member.name}[i]);\n'
-                                    copy_strings += '    }\n'
-                                    copy_strings += f'    {member.name} = tmp_{member.name};\n'
+                                    copy_strings += (
```
lunarpapillo, line 919: You shouldn't need the surrounding parentheses if you're using a single '''-quoted string.

scripts/generators/safe_struct_generator.py
```diff
@@ -971,65 +979,67 @@ def qfi_construct(item, member):
...
+                        init_func_txt += f'{member.name} = nullptr;\n'
                         array_element = f'in_struct->{member.name}[i]'
                         if member.type in self.vk.structs and needSafeStruct(self.vk.structs[member.type]):
                             array_element = f'{member.type}(&in_struct->safe_{member.name}[i])'
-                        construct_txt += f'    if ({member.length} && in_struct->{member.name}) {{\n'
-                        construct_txt += f'        {member.name} = new {m_type}[{member.length}];\n'
-                        destruct_txt += f'    if ({member.name})\n'
-                        destruct_txt += f'        delete[] {member.name};\n'
-                        construct_txt += f'        for (uint32_t i = 0; i < {member.length}; ++i) {{\n'
+                        construct_txt += f'if ({member.length} && in_struct->{member.name}) {{\n'
```
lunarpapillo, line 1007: Generally not a fan of repeated string concatenations - Python has the same issues as languages that use `StringBuilder`.  But if performance is not an issue, I recognize that it would be a huge change to make code consistent across all these generators.

scripts/generators/spirv_grammar_generator.py
```diff
@@ -260,15 +260,15 @@ def generateSource(self):
 // All information related to each SPIR-V opcode instruction
 struct InstructionInfo {
     const char* name;
-    bool has_type; // always operand 0 if present
-    bool has_result; // always operand 1 if present
+    bool has_type;    // always operand 0 if present
```
lunarpapillo, line 263: why add whitespace at all here, if `clang-format` is going to reformat it anyway?

scripts/generators/stateless_validation_helper_generator.py
```diff
@@ -20,7 +20,7 @@
 
 import os
 import re
-from generators.generator_utils import (buildListVUID, incIndent, decIndent)
+from generators.generator_utils import (buildListVUID)
```
lunarpapillo, line 23: I don't think the parentheses are necessary (or recommended)

lunarpapillo (commented):
A whole lot of work here, kudos to your attention to detail...

As a complete aside, I do recommend use of `pylint` and a `.pylintrc` to:
- catch Python errors early
- provide recommendations on Python best practices
- help keep Python code style consistent across different scripts

scripts/generators/extension_helper_generator.py
jzulauf-lunarg:
```
    out.append('''
        struct InstanceReq {
            ....
        }'''
```

jzulauf-lunarg: s.t. the next line of python lines up with `out.` visually bracketing the quoted text

scripts/generators/best_practices_generator.py
```diff
@@ -175,25 +175,27 @@ def generateSource(self):
...
             if command.name in self.manual_postcallrecord_list:
-                out.append(f'    ManualPostCallRecord{command.name[2:]}({params});\n')
+                out.append(f'ManualPostCallRecord{command.name[2:]}({params});\n')
 
             if hasNonVkSuccess(command.successCodes):
-                out.append('    if (record_obj.result > VK_SUCCESS) {\n')
-                out.append('        LogPositiveSuccessCode(record_obj);\n')
-                out.append('        return;\n')
-                out.append('    }\n')
+                out.append(
```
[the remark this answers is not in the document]
jzulauf-lunarg: I think that loses readability of the code that will be generated.

scripts/generators/best_practices_generator.py
```diff
@@ -175,25 +175,27 @@ def generateSource(self):
...
+                        LogPositiveSuccessCode(record_obj);
+                        return;
+                    }''')
 
             if command.errorCodes is not None:
-                out.append('    if (record_obj.result < VK_SUCCESS) {\n')
-                out.append('        LogErrorCode(record_obj);\n')
-                out.append('    }\n')
+                out.append(
+                    '''if (record_obj.result < VK_SUCCESS) {
```
jzulauf-lunarg, line 196: Leaving the `'''` with the '(' would preserve relative indent of the quoted snippet.

scripts/generators/safe_struct_generator.py
spencer-lunarg: So the issue here is slightly stupid IMO

My goal of this change was not to make any change to the generated output, but If I go

```python
copy_strings += '''if () {
    for (...) {
    '''
```

it doesn't align well, but if I go

```python
copy_strings += '''
    if () {
        for (...) {
    '''
```

the new generated code has this extra new line above it...

scripts/generators/safe_struct_generator.py
spencer-lunarg: when I started the refactor most the code was string concatenations and we got it over to `append()`

I still have a few spots left, but it is a TODO to remove all these concatenations soon

scripts/generators/spirv_grammar_generator.py
spencer-lunarg: this is how clang-format reformatted it... I guess goal was to sync things up now (in case you wanted to search).... but also torn because this is not maintainable and probably not worth it

scripts/generators/best_practices_generator.py
spencer-lunarg: ya, @lunarpapillo had similar concerns... I guess maybe just worth doing it, and checking in the extra `\n` that will come from it

spencer-lunarg: applied and will just check in the new `\n` and that way things are just "how they should be" in the scripts for the next person copying and pasting it

Issue #5507, opened 2023-03-25
Add missing Dynamic State draw time VUs


spencer-lunarg:
The following are draw time VUs missing for Dynamic State

`VK_DYNAMIC_STATE_RASTERIZATION_SAMPLES_EXT`
- [x] VUID-vkCmdDraw-pNext-07935 (simliar to VUID-VkGraphicsPipelineCreateInfo-renderPass-06854)
- [x] VUID-vkCmdDraw-sampleLocationsEnable-07936 (simliar to VUID-VkGraphicsPipelineCreateInfo-pDynamicStates-07610)
- [x] VUID-vkCmdDraw-sampleLocationsEnable-07937 (simliar to VUID-VkGraphicsPipelineCreateInfo-pDynamicStates-07611)
- [x] VUID-vkCmdDraw-sampleLocationsEnable-07938 (simliar to VUID-VkGraphicsPipelineCreateInfo-pDynamicStates-07612)
 
`VK_DYNAMIC_STATE_SAMPLE_LOCATIONS_EXT`
- [x] VUID-vkCmdDraw-sampleLocationsPerPixel-07482
- [x] VUID-vkCmdDraw-sampleLocationsPerPixel-07483
- [x] VUID-vkCmdDraw-rasterizationSamples-07471
- [x] VUID-vkCmdDraw-samples-07472 (`VK_DYNAMIC_STATE_SAMPLE_MASK_EXT`)
- [x] VUID-vkCmdDraw-samples-07473
- [x] VUID-vkCmdDraw-rasterizationSamples-07474 (without `VK_EXT_multisampled_render_to_single_sampled`)
- [x] VUID-vkCmdDraw-multisampledRenderToSingleSampled-07475 (with `VK_EXT_multisampled_render_to_single_sampled`)

`VK_DYNAMIC_STATE_COLOR_BLEND_ENABLE_EXT`
- [x] VUID-vkCmdDraw-pColorBlendEnables-07470

`VK_DYNAMIC_STATE_COLOR_BLEND_ENABLE_EXT` and `VK_DYNAMIC_STATE_COLOR_BLEND_ADVANCED_EXT`
- [x] VUID-vkCmdDraw-advancedBlendMaxColorAttachments-07480

`VK_DYNAMIC_STATE_SAMPLE_LOCATIONS_ENABLE_EXT`
- [x] VUID-vkCmdDraw-sampleLocationsEnable-07484
[...]

PR #6500, opened 2023-09-15
Add vuid 07939, validate all required vertex shader inputs are provided with vkCmdSetVertexInputEXT()

ziga-lunarg:
Part of #5507

### DIFF
--- a/layers/core_checks/cc_cmd_buffer_dynamic.cpp
+++ b/layers/core_checks/cc_cmd_buffer_dynamic.cpp
@@ -269,17 +269,22 @@ bool CoreChecks::ValidateDrawDynamicState(const LAST_BOUND_STATE& last_bound_sta
         }
     }
     if ((!pipeline_state || pipeline_state->IsDynamic(VK_DYNAMIC_STATE_VERTEX_INPUT_EXT)) && entrypoint) {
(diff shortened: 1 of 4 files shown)


layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -269,17 +269,19 @@ bool CoreChecks::ValidateDrawDynamicState(const LAST_BOUND_STATE& last_bound_sta
         }
     }
     if ((!pipeline_state || pipeline_state->IsDynamic(VK_DYNAMIC_STATE_VERTEX_INPUT_EXT)) && entrypoint) {
-        for (uint32_t i = 0; i < cb_state.dynamic_state_value.vertex_attribute_descriptions.size(); ++i) {
-            const auto& description = cb_state.dynamic_state_value.vertex_attribute_descriptions[i];
-            bool format64 = vkuFormatIs64bit(description.format);
-            for (const auto* variable_ptr : entrypoint->user_defined_interface_variables) {
+        for (const auto* variable_ptr : entrypoint->user_defined_interface_variables) {
+            bool location_provided = false;
```
spencer-lunarg, line 273: maybe lets just do a 

```cpp
if (variable_ptr->storage_class != spv::StorageClass::StorageClassInput) {
    continue;
}
```

here as it will simplify the logic and save having to do the 2nd for loop on Output variables

PR #6504, opened 2023-09-16
tests: Fix mesh shader object debug printf test

ziga-lunarg:


spencer-lunarg (approved):
LGTM (once ran with clang-format to please CI)

PR #6507, opened 2023-09-18
Misc cleanup to shader logic

spencer-lunarg:
things I wanted to fix, but didn't have a place to put them

### DIFF
--- a/layers/core_checks/cc_spirv.cpp
+++ b/layers/core_checks/cc_spirv.cpp
@@ -174,13 +174,11 @@ bool CoreChecks::ValidatePushConstantUsage(const StageCreateInfo &create_info, c
     return skip;
 }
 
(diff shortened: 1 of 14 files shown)


layers/core_checks/cc_spirv.cpp
```diff
@@ -1815,81 +1805,87 @@ bool CoreChecks::ValidateTexelOffsetLimits(const SPIRV_MODULE_STATE &module_stat
     bool skip = false;
 
     const uint32_t opcode = insn.Opcode();
-    if (ImageGatherOperation(opcode) || ImageSampleOperation(opcode) || ImageFetchOperation(opcode)) {
```
spencer-lunarg, line 1818: this is just returning early to not have to get so deeply nested

juan-lunarg (approved):

Issue #6467, opened 2023-09-11
Missing check for VUID-VkBindImagePlaneMemoryInfo-planeAspect-02284


dkorkmazturk:
**Describe the situation in which you encountered the missing validation**
Issue description:

The Vulkan spec specifies that when an image's tiling is set to VK_IMAGE_TILING_DRM_FORMAT_MODIFIER_EXT, planeAspect field of VkBindImagePlaneMemoryInfo struct should be a single valid **memory plane** for the image. However, VLL does not produce any warnings when an image plane (VK_IMAGE_ASPECT_PLANE_i_BIT) is used instead of a memory plane (VK_IMAGE_ASPECT_MEMORY_PLANE_i_BIT_EXT) for VkBindImagePlaneMemoryInfo::planeAspect.

Using memory planes here actually causes a segmentation fault in VLL while binding the planes of a disjoint image using vkBindImageMemory2().

**Valid Usage IDs requested**
Please include the valid usage IDs for the checks you are requesting:

VUID-VkBindImagePlaneMemoryInfo-planeAspect-02284

**Additional Context**

Issue #5632, opened 2023-04-12
Add missing VK_EXT_image_drm_format_modifier


spencer-lunarg:
- [x] VUID-VkImageCreateInfo-tiling-02353
  - related VUID-VkImageCreateInfo-tiling-02261
- [x] VUID-VkImageCreateInfo-pNext-06746
- [x] VUID-VkDeviceImageMemoryRequirementsKHR-pCreateInfo-06420 https://gitlab.khronos.org/vulkan/vulkan/-/issues/3620
- [ ] VUID-VkBindImagePlaneMemoryInfo-planeAspect-02284
- [x] VUID-vkGetImageDrmFormatModifierPropertiesEXT-image-02272
- [x] VUID-VkPhysicalDeviceImageDrmFormatModifierInfoEXT-sharingMode-02314
- [x] VUID-VkPhysicalDeviceImageDrmFormatModifierInfoEXT-sharingMode-02316
- [x] VUID-VkImageSubresourceRange-aspectMask-02278
  - related VUID-VkImageSubresourceLayers-aspectMask-02247
- [ ] VUID-VkImagePlaneMemoryRequirementsInfo-planeAspect-02282

PR #6511, opened 2023-09-19
Validate DRM Disjoint MultiPlanar PlaneAspect

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5632
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6467

This will catch the most likely error case of using `VK_IMAGE_ASPECT_PLANE_0_BIT` and not  `VK_IMAGE_ASPECT_MEMORY_PLANE_0_BIT_EXT`

Plan to add proper support for `drmFormatModifierPlaneCount` but need to first confirm spec related questions around it


arno-lunarg (approved):
LGTM!

PR #6512, opened 2023-09-19
layers: Cleanup VerifyImageLayout code

spencer-lunarg:
We had 2 different paths for `VerifyImageLayout` and not obvious which was being checked with the extra VUs so renamed to `VerifyImageLayoutSubresource`

### DIFF
--- a/layers/core_checks/cc_image_layout.cpp
+++ b/layers/core_checks/cc_image_layout.cpp
@@ -143,22 +143,22 @@ struct LayoutUseCheckAndMessage {
 template <typename RangeFactory>
 bool CoreChecks::VerifyImageLayoutRange(const CMD_BUFFER_STATE &cb_state, const IMAGE_STATE &image_state,
                                         VkImageAspectFlags aspect_mask, VkImageLayout explicit_layout,
-                                        const RangeFactory &range_factory, const Location &loc,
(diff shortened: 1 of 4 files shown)


layers/core_checks/cc_image_layout.cpp
```diff
@@ -172,17 +172,20 @@ bool CoreChecks::VerifyImageLayoutRange(const CMD_BUFFER_STATE &cb_state, const
...
     if (disabled[image_layout_validation]) return false;
     bool skip = false;
 
+    const VkImageSubresourceRange range = RangeFromLayers(subresource_layers);
+
     VkImageSubresourceRange normalized_isr = image_state.NormalizeSubresourceRange(range);
     auto range_factory = [&normalized_isr](const ImageSubresourceLayoutMap &map) { return map.RangeGen(normalized_isr); };
-    skip |= VerifyImageLayoutRange(cb_state, image_state, aspect_mask, explicit_layout, range_factory, loc,
-                                   layout_mismatch_msg_code, error);
+    bool unused_error = false;
```
spencer-lunarg, line 186: TODO to remove this next when I remove it in `VerifyImageLayoutRange`

arno-lunarg (approved):

Issue #4158, opened 2022-05-26
Missing VUID checking for xfb pipeline in vkCmdBeginTransformFeedbackEXT


baldurk:
**Describe the situation in which you encountered the missing validation**
I encountered a crash on a driver when calling `vkCmdBeginTransformFeedbackEXT`, after investigating I found the problem was that the pipeline bound didn't have the `Xfb` execution mode properly set, but validation did not catch this.

**Valid Usage IDs requested**
* [ ] VUID-vkCmdBeginTransformFeedbackEXT-None-04128

**Additional Context**
There may be some others in this area that are missing, I didn't check for related ones.

(related  VUID-VkGraphicsPipelineCreateInfo-pStages-02318)

PR #6513, opened 2023-09-19
layers: Add Xfb Execution Mode for PreRaster

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4158

adds support for `VUID-vkCmdBeginTransformFeedbackEXT-None-04128` and `VUID-VkGraphicsPipelineCreateInfo-pStages-02318`

### DIFF
--- a/layers/core_checks/cc_spirv.cpp
+++ b/layers/core_checks/cc_spirv.cpp
@@ -1758,6 +1758,25 @@ bool CoreChecks::ValidateTransformFeedback(const SPIRV_MODULE_STATE &module_stat
                                            const StageCreateInfo &create_info, const Location &loc) const {
     bool skip = false;
 
+    if (create_info.pipeline) {
+        const bool is_xfb_execution_mode = entrypoint.execution_mode.Has(ExecutionModeSet::xfb_bit);
+        if (is_xfb_execution_mode) {
+            if ((create_info.pipeline->create_info_shaders & (VK_SHADER_STAGE_MESH_BIT_EXT | VK_SHADER_STAGE_TASK_BIT_EXT)) != 0) {
+                skip |= LogError("VUID-VkGraphicsPipelineCreateInfo-None-02322", module_state.handle(), loc,
+                                 "SPIR-V has OpExecutionMode of Xfb and using mesh shaders (%s).",
+                                 string_VkShaderStageFlags(create_info.pipeline->create_info_shaders).c_str());
+            }
+
+            if (create_info.pipeline->pre_raster_state &&
+                (entrypoint.stage != create_info.pipeline->pre_raster_state->last_stage)) {
+                skip |= LogError("VUID-VkGraphicsPipelineCreateInfo-pStages-02318", module_state.handle(), loc,
+                                 "SPIR-V has OpExecutionMode of Xfb in %s, but %s is the last last pre-rasterization shader stage.",
+                                 string_VkShaderStageFlagBits(entrypoint.stage),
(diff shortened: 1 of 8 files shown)


layers/state_tracker/pipeline_sub_state.cpp
```diff
@@ -69,64 +69,86 @@ PreRasterState::PreRasterState(const PIPELINE_STATE &p, const ValidationStateTra
...
+        last_stage = VK_SHADER_STAGE_GEOMETRY_BIT;
+    } else if (all_stages & VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT) {
+        last_stage = VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT;
+    } else if (all_stages & VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT) {
+        last_stage = VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT;
+    } else if (all_stages & VK_SHADER_STAGE_VERTEX_BIT) {
+        last_stage = VK_SHADER_STAGE_VERTEX_BIT;
+    } else {
+        // TODO is this an error?
+        last_stage = VK_SHADER_STAGE_FLAG_BITS_MAX_ENUM;
```
artem-lunarg, line 150: assert this place?

layers/state_tracker/pipeline_sub_state.h
```diff
@@ -81,6 +81,8 @@ struct PreRasterState : public PipelineSubState {
     std::shared_ptr<const RENDER_PASS_STATE> rp_state;
     uint32_t subpass = 0;
 
+    VkShaderStageFlagBits last_stage;
```
artem-lunarg, line 84: the style of this class is to use in-class member initialization for primitive types (even constructor does more complex job). The suggestion is to initialize `last_stage`

layers/core_checks/cc_spirv.cpp
```diff
@@ -1758,6 +1758,25 @@ bool CoreChecks::ValidateTransformFeedback(const SPIRV_MODULE_STATE &module_stat
                                            const StageCreateInfo &create_info, const Location &loc) const {
     bool skip = false;
 
+    if (create_info.pipeline) {
+        const bool is_xfb_execution_mode = entrypoint.execution_mode.Has(ExecutionModeSet::xfb_bit);
+        if (is_xfb_execution_mode) {
+            if ((create_info.pipeline->create_info_shaders & (VK_SHADER_STAGE_MESH_BIT_EXT | VK_SHADER_STAGE_TASK_BIT_EXT)) != 0) {
```
artem-lunarg, line 1764: How was this VU split into two conditions? Can't see this directly from VU description.

layers/state_tracker/pipeline_sub_state.cpp
spencer-lunarg: there is this same TODO above, I have a very near-future todo to prove these are indeed assert-able

layers/state_tracker/pipeline_sub_state.h
spencer-lunarg: aww, ya good call

layers/core_checks/cc_spirv.cpp
spencer-lunarg: `02322` was in `ValidateShaderStageInputOutputLimits` for some reason (but it is really checking for Transform feedback, so moved it here and then add `02318`... which I typo'ed on, good catch, will update the VUID

artem-lunarg (approved):

PR #6516, opened 2023-09-19
build: Use vk_struct_helper.hpp instead

charles-lunarg:
Replaces usage of vk_typemap_helper.h with vk_struct_helper.hpp from the Vulkan-Utility-Libraries. It has the same functionality but since this helper library lives in multiple places, it was moved to Vulkan-Utility-Libraries to reduce code duplication.

Many places are able to use the vku::InitStructHelper class to deduce the type to be initialized. This helper class does not work with auto nor with a variadic argument list.

Note - this PR breaks basically every other PR in flight. I suggest inserting a execution barrier in the PR queue since this changes every instance of LvlInitStruct, which has 6000+ instances.


layers/vulkan/generated/vk_typemap_helper.h
juan-lunarg, line 1: 💯

juan-lunarg (approved):
On passing CI

PR #6519, opened 2023-09-20
Change vk_testing to vkt

spencer-lunarg:
As agreed on, goal is to consolidate all the Testing RAII objects to a single format


artem-lunarg (approved):
LGTM. Shorter name also removed a bunch of lines.

arno-lunarg (approved):

Issue #4667, opened 2022-10-15
LAYER_CONFIGURATION.md is entirely useless and provides no info


Joshua-Ashton:
This has been bugging me for a while.
This file has *a lot* of text that tells me *absolutely nothing*.

I want to find the env var/property to filter some warnings/errors, its not here.

The actual layer configuration information is at https://vulkan.lunarg.com/doc/view/latest/linux/khronos_validation_layer.html or in the `VkLayer_khronos_validation.json` which is not referenced by this file or the README.md at all.

So what is this file for? It provides no information whatsoever to me as a user of the validation layers aside from wasting my time thinking that what I want is here, and not pointing me to the right places...

It would be better to just have a link to https://vulkan.lunarg.com/doc/view/latest/linux/khronos_validation_layer.html in the README.md

chloekek:
I disagree that this document is completely useless. I frequently visit the document to copy the name of the `VK_INSTANCE_LAYERS` environment variable and the name of the `VK_LAYER_KHRONOS_validation` layer to my clipboard. The documents you linked do not tell my the name of the environment variables.

PR #6523, opened 2023-09-20
docs: Link validation layer settings on the front page

christophe-lunarg:
Address https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4667
When https://github.com/KhronosGroup/Vulkan-Utility-Libraries/pull/85 will be merged we can change the link of the layer configurations to point the new shared and common documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,19 @@
-# Vulkan Validation Layers (VVL)
+# Vulkan Validation Layer (VVL)
 
 ## Introduction
 
 Vulkan is an Explicit API, enabling direct control over how GPUs actually work. By design, minimal error checking is done inside
 a Vulkan driver. Applications have full control and responsibility for correct operation. Any errors in
-how Vulkan is used can result in a crash. This project provides Vulkan validation layers that can be enabled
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,15 +1,19 @@
-# Vulkan Validation Layers (VVL)
+# Vulkan Validation Layer (VVL)
```
juan-lunarg, line 1: The repo is called the `Vulkan-ValidationLayers`. I'd prefer not to change that in this PR.

I know it's technically 1 layer. But I'd prefer to leave it alone for now.

juan-lunarg (changes requested):

README.md
christophe-lunarg: Why?

christophe-lunarg: Closing, I let you resolve https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4667 the way you want.

juan-lunarg: ?

Most of this PR looks good. It's just that there are multiple places where we reference this as the validation `layers`. So we can't just change it in 1 place.

Issue #6282, opened 2023-08-08
Layers not taking into account ray tracing pipeline libraries in capture/replay handle verification


rg3igalia:
**Environment:**
 - OS: Linux
 - GPU: RX 6700XT
 - SDK or header version if building from repo: 195b6d8b16d5494af686c73306767e54647fb352
 - Options enabled (synchronization, best practices, etc.): defaults + synchronization

**Describe the Issue**

When running some pipeline library CTS tests from the `dEQP-VK.ray_tracing_pipeline.pipeline_library.*` subgroup the layers are emitting `VUID-vkGetRayTracingCaptureReplayShaderGroupHandlesKHR-firstGroup-03483`, but the errors seem spurious because the layers are not taking into account that some of the pipelines the replay handles are being requested from include other pipeline libraries.

E.g. `dEQP-VK.ray_tracing_pipeline.pipeline_library.configurations.singlethreaded_compilation.s0_l1_check_capture_replay_handles` results in:

> VUID-vkGetRayTracingCaptureReplayShaderGroupHandlesKHR-firstGroup-03483(ERROR / SPEC): msgNum: 164308374 - Validation Error: [ VUID-vkGetRayTracingCaptureReplayShaderGroupHandlesKHR-firstGroup-03483 ] Object 0: handle = 0x556d412820e0, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x9cb2596 | vkGetRayTracingCaptureReplayShaderGroupHandlesKHR: The sum of firstGroup (0) and groupCount (3) must be less than or equal to the number of shader groups in pipeline (2). The Vulkan spec states: The sum of firstGroup and groupCount must be less than or equal to the number of shader groups in pipeline (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkGetRayTracingCaptureReplayShaderGroupHandlesKHR-firstGroup-03483)

This does not always happen. For example, `dEQP-VK.ray_tracing_pipeline.pipeline_library.configurations.singlethreaded_compilation.s0_l1_check_group_handles` does not emit any error because, for `vkGetRayTracingShaderGroupHandlesKHR`, the layers are already calculating the number of groups recursively. The same logic should probably be applied to `vkGetRayTracingCaptureReplayShaderGroupHandlesKHR`.

**Expected behavior**

No validation errors.

**Valid Usage ID**

[...]

PR #6525, opened 2023-09-20
layers: Fix group count calculation

arno-lunarg:
- Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6282

Add a regression test, `PositiveRayTracingPipeline.GetCaptureReplayShaderGroupHandlesKHR`

### DIFF
--- a/tests/framework/shader_templates.h
+++ b/tests/framework/shader_templates.h
@@ -143,6 +143,16 @@ static char const kFragmentColorOutputGlsl[] = R"glsl(
     void main() {}
 )glsl";
 
+[[maybe_unused]] static char const kRayTracingPayloadMinimalGlsl[] = R"glsl(
+        #version 460
+        #extension GL_EXT_ray_tracing : enable
+        layout(location = 0) rayPayloadInEXT float hitValue;
+
+        void main() {
+            hitValue = 1.0;
+        }
+    )glsl";
+
 [[maybe_unused]] static const char *kRayTracingNVMinimalGlsl = R"glsl(
     #version 460
     #extension GL_NV_ray_tracing : require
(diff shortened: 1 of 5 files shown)


tests/framework/shader_templates.h
```diff
@@ -297,4 +297,48 @@ static char const kShaderTileImageDepthStencilReadSpv[] = R"(
                OpStore %uFragColor %22
                OpReturn
                OpFunctionEnd
-        )";
\ No newline at end of file
+        )";
+
+[[maybe_unused]] static const char kRayGenShaderText[] = R"glsl(
```
spencer-lunarg, line 302: The goal of `shader_templates.h` was suppose to be for things used in various tests

if this will only be used inside the ray tracing tests, might be better to keep in it that file

arno-lunarg: yes the scope is more limited, I will move it back

spencer-lunarg: aww, I see it is used in the `pipeline_helper.h` so maybe ok being here ... I guess can we give this a more descriptive name,  like we already have `kRayTracingMinimalGlsl`, what is the goal when to use this one?

tests/framework/shader_templates.h
```diff
@@ -297,4 +297,48 @@ static char const kShaderTileImageDepthStencilReadSpv[] = R"(
...
+           payload = 0.5;
+           traceRayEXT(as, gl_RayFlagsCullBackFacingTrianglesEXT, 0xff, 0, 1, 0, origin, 0.0, dir, 1000.0, 0);
+
+           col.y = payload;
+
+           imageStore(image, ivec2(gl_LaunchIDEXT.xy), col);
+        }
+    )glsl";
+
+[[maybe_unused]] static char const kClosestHitShaderText[] = R"glsl(
```
spencer-lunarg, line 326: this is the same shader as `kMissShaderText` and seems to really be like `kRayTracingMinimalGlsl` but with payload... so maybe lets just call this something like `kRayTracingPayloadMinimalGlsl` (and move it next to the `kRayTracingMinimalGlsl` code as well)

arno-lunarg: ok!

tests/framework/shader_templates.h
```diff
@@ -297,4 +307,38 @@ static char const kShaderTileImageDepthStencilReadSpv[] = R"(
...
+           payload = 0.5;
+           traceRayEXT(as, gl_RayFlagsCullBackFacingTrianglesEXT, 0xff, 0, 1, 0, origin, 0.0, dir, 1000.0, 0);
+
+           col.y = payload;
+
+           imageStore(image, ivec2(gl_LaunchIDEXT.xy), col);
+        }
+    )glsl";
+
+[[maybe_unused]] static char const kMissShaderText[] = R"glsl(
```
spencer-lunarg, line 336: this the same as `kRayTracingPayloadMinimalGlsl`

tests/framework/shader_templates.h
```diff
@@ -297,4 +307,38 @@ static char const kShaderTileImageDepthStencilReadSpv[] = R"(
                OpStore %uFragColor %22
                OpReturn
                OpFunctionEnd
-        )";
\ No newline at end of file
+        )";
+
+[[maybe_unused]] static const char kRayGenShaderText[] = R"glsl(
```
spencer-lunarg, line 312: nit, end name with `Glsl` like the others

PR #6529, opened 2023-09-20
gpu: Remove SPIR-V disassembly from generated headers

jeremyg-lunarg:
These comments aren't of any value, they just add churn.


juan-lunarg (approved):

Tony-LunarG (approved):
LGTM

PR #6531, opened 2023-09-21
Remove VkObj for vkt classes

spencer-lunarg:
This is another step to move towards only 1 RAII object

### DIFF
--- a/tests/unit/arm_best_practices.cpp
+++ b/tests/unit/arm_best_practices.cpp
@@ -17,6 +17,19 @@
 
 const char *kEnableArmValidation = "VALIDATION_CHECK_ENABLE_VENDOR_SPECIFIC_ARM";
 
+class VkConstantBufferObj : public vkt::Buffer {
+  public:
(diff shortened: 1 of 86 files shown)


tests/unit/arm_best_practices.cpp
```diff
@@ -17,6 +17,19 @@
 
 const char *kEnableArmValidation = "VALIDATION_CHECK_ENABLE_VENDOR_SPECIFIC_ARM";
 
+class VkConstantBufferObj : public vkt::Buffer {
```
spencer-lunarg, line 20: There are 2 tests in this file that need the values in the buffer, for now just moved what is still needed of `VkConstantBufferObj` into this file for future cleanup

charles-lunarg:
_In the Land of Testing where the Objects lie._
_One RAII to rule them all, One RAII to construct them,_
_One RAII to bring them all and in the destructor bind them_
_In the Land of Testing where the Objects lie._

juan-lunarg (approved):

PR #6532, opened 2023-09-21
build: Fix asan option for MSVC

arno-lunarg:


CMakeLists.txt
```diff
@@ -82,8 +82,12 @@ endif()
 
 option(VVL_ENABLE_ASAN "Use address sanitization")
 if (VVL_ENABLE_ASAN)
-    add_compile_options(-fsanitize=address -fno-omit-frame-pointer)
-    add_link_options(-fsanitize=address)
+    if (MSVC)
+        add_compile_options(/fsanitize=address /fno-omit-frame-pointer)
```
juan-lunarg, line 86: ` /fno-omit-frame-pointer` isn't a MSVC compiler flag as far as I can see.
https://learn.microsoft.com/en-us/cpp/build/reference/compiler-options-listed-by-category?view=msvc-170

```suggestion
        add_compile_options(/fsanitize=address)
```

PR #6534, opened 2023-09-21
Ray tracing cleanup: Add _NV suffix to ACCELERATION_STRUCTURE_STATE, use vkt namespace instead of rt in tests

arno-lunarg:


spencer-lunarg (approved):

arno-lunarg:
Merging `ACCELERATION_STRUCTURE_KHR` and `ACCELERATION_STRUCTURE_NV` is not feasible, they share a name but not the same usage philosophy, better to keep them well separated.

jeremyg-lunarg (approved):

PR #6542, opened 2023-09-22
layers: Fix broken VUIDs

spencer-lunarg:
We have some VUID left that aren't in the spec so this fixes them up

### DIFF
--- a/layers/core_checks/cc_pipeline_graphics.cpp
+++ b/layers/core_checks/cc_pipeline_graphics.cpp
@@ -274,21 +274,6 @@ bool CoreChecks::ValidateGraphicsPipelinePortability(const PIPELINE_STATE &pipel
 bool CoreChecks::ValidateGraphicsPipelineLibrary(const PIPELINE_STATE &pipeline, const Location &create_info_loc) const {
     bool skip = false;
 
-    // If VK_EXT_graphics_pipeline_library is not enabled, a complete set of state must be defined at this point
-    std::string full_pipeline_state_msg;
-    if (!pipeline.vertex_input_state) {
-        full_pipeline_state_msg += "<vertex input> ";
-    }
-    if (!pipeline.pre_raster_state) {
-        full_pipeline_state_msg += "<pre-raster> ";
-    }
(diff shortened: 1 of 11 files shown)


layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -298,12 +283,7 @@ bool CoreChecks::ValidateGraphicsPipelineLibrary(const PIPELINE_STATE &pipeline,
     }
 
     if (!IsExtEnabled(device_extensions.vk_ext_graphics_pipeline_library)) {
-        if (!pipeline.HasFullState()) {
```
spencer-lunarg, line 301: these VUs were removed because no way to actually hit them without GPL

layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -807,35 +777,28 @@ bool CoreChecks::ValidateGraphicsPipelineInputAssemblyState(const PIPELINE_STATE
...
+                                 VK_PRIMITIVE_TOPOLOGY_TRIANGLE_LIST_WITH_ADJACENCY, VK_PRIMITIVE_TOPOLOGY_PATCH_LIST})) {
+            if (topology == VK_PRIMITIVE_TOPOLOGY_PATCH_LIST) {
+                if (!enabled_features.primitive_topology_list_restart_features.primitiveTopologyPatchListRestart) {
+                    skip |= LogError("VUID-VkPipelineInputAssemblyStateCreateInfo-topology-06253", device, ia_loc,
+                                     "topology is %s and primitiveRestartEnable is VK_TRUE and the "
+                                     "primitiveTopologyPatchListRestart feature was not enabled.",
+                                     string_VkPrimitiveTopology(topology));
+                }
+            } else if (!enabled_features.primitive_topology_list_restart_features.primitiveTopologyListRestart) {
+                skip |=
```
spencer-lunarg, line 791: `00428` was the old VU, now can just use `06252`

juan-lunarg (approved):

PR #6546, opened 2023-09-22
Fix resetting performance queries

### LINKED ISSUE #6304: Unexpected validation errors with performance queries with multiple passes
**Environment:**
 - OS: Linux
 - GPU: RX 6700 XT (using the RADV Mesa driver)
 - SDK or header version if building from repo: d966a60f13e21ebb2bdd3112861d50c18f1e0620
 - Options enabled (synchronization, best practices, etc.): defaults + synchronization

**Describe the Issue**

When an app uses performance queries that need multiple passes, it's supposed to reset the query either on the host or a separate command buffer, and then submit a command buffer containing a `vkCmdBeginQuery` + stuff + `vkCmdEndQuery` sequence multiple times. In each submission, it has to include a `VkPerformanceQuerySubmitInfoKHR` indicating the pass index in the `pNext` chain of `VkSubmitInfo`. However, when doing that the layers generate `VUID-vkCmdBeginQuery-None-00807` erroneously.

I reproduced this problem with the performance query CTS tests in `dEQP-VK.query_pool.performance_query.*` tests. However, note that, before my changes for Khronos internal CTS issue 4615, the tests contain multiple validation errors. After my changes, `VUID-vkCmdBeginQuery-None-00807` appears to be the only remaining validation issue in those tests.

**Expected behavior**

No validation errors.

**Valid Usage ID**

`VUID-vkCmdBeginQuery-None-00807`

> VUID-vkCmdBeginQuery-None-00807(ERROR / SPEC): msgNum: 680962600 - Validation Error: [ VUID-vkCmdBeginQuery-None-00807 ] Object 0: handle = 0x55ac26bc6d90, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0x967dd1000000000e, type = VK_OBJECT_TYPE_QUERY_POOL; | MessageID = 0x2896aa28 | vkCmdBeginQuery: VkQueryPool 0x967dd1000000000e[] and query 0: query not reset. After query pool creation, each query must be reset before it is used. Queries must also be reset between uses. The Vulkan spec states: All queries used by the command must be unavailable (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdBeginQuery-None-00807)

**Additional context**

Possibly related to issue #1795, but I prefered to create a new one with more context.

ziga-lunarg:
Closes #6304

### DIFF
--- a/tests/unit/query_positive.cpp
+++ b/tests/unit/query_positive.cpp
@@ -447,3 +447,101 @@ TEST_F(PositiveQuery, CommandBufferInheritanceFlags) {
     vk::CmdEndQuery(m_commandBuffer->handle(), query_pool.handle(), 0);
     m_commandBuffer->end();
 }
+
+TEST_F(PositiveQuery, PerformanceQueries) {
+    TEST_DESCRIPTION("Test performance queries.");
+
+    SetTargetApiVersion(VK_API_VERSION_1_1);
+    AddRequiredExtensions(VK_KHR_PERFORMANCE_QUERY_EXTENSION_NAME);
+    ASSERT_NO_FATAL_FAILURE(InitFramework());
+
+    if (DeviceValidationVersion() < VK_API_VERSION_1_1) {
+        GTEST_SKIP() << "At least Vulkan version 1.1 is required";
+    }
+
+    auto performance_query_features = vku::InitStruct<VkPhysicalDevicePerformanceQueryFeaturesKHR>();
+    auto features2 = GetPhysicalDeviceFeatures2(performance_query_features);
+
+    if (!performance_query_features.performanceCounterQueryPools) {
+        GTEST_SKIP() << "Test requires (unsupported) performanceCounterQueryPools";
+    }
+
+    ASSERT_NO_FATAL_FAILURE(InitState(nullptr, &features2));
+
+    uint32_t counterCount = 0u;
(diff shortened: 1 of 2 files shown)


tests/unit/query_positive.cpp
```diff
@@ -447,3 +447,102 @@ TEST_F(PositiveQuery, CommandBufferInheritanceFlags) {
...
+        m_commandBuffer->end();
+
+        auto submit_info = vku::InitStruct<VkSubmitInfo>();
+        submit_info.commandBufferCount = 1;
+        submit_info.pCommandBuffers = &m_commandBuffer->handle();
+        vk::QueueSubmit(m_device->m_queue, 1, &submit_info, VK_NULL_HANDLE);
+        vk::DeviceWaitIdle(*m_device);
+    }
+
+    vkt::CommandPool command_pool(*m_device, m_device->graphics_queue_node_index_);
```
spencer-lunarg, line 512: was there a reason to create a new command pool/buffer here?

spencer-lunarg (approved):

tests/unit/query_positive.cpp
ziga-lunarg: No, removed.

PR #6552, opened 2023-09-25
tests: Fix some dependency issues

aqnuep:
This PR fixes miscellaneous feature/version/extension dependency issues in test cases.

We've identified over 150 other test cases that have dependency issues. In particular, many test cases that test extensions or features that were promoted to core in Vulkan 1.1, 1.2 (and maybe 1.3) get skipped if only the core version is available but not the extensions. However, as exposing the extensions corresponding to features promoted to core is not mandatory, it would be preferable to fix these.

One way this could be detected is by creating a test device profile that only supports the core features and see how many test cases (that would otherwise apply) get skipped.

Furthermore, there are also many test cases that do correctly use the core API version as a dependency (or alternative dependency alongside of the core version) yet they don't always use the correct entry point (core vs extension suffixed ones). But this MR does not try to attempt to fix those, or fix all incorrect dependencies in general.


tests/unit/command.cpp
```diff
@@ -1230,7 +1230,8 @@ TEST_F(NegativeCommand, CopyImageLayerCountMismatch) {
         "Try to copy between images with the source subresource having a different layerCount than the destination subresource");
     AddOptionalExtensions(VK_KHR_MAINTENANCE_1_EXTENSION_NAME);
     ASSERT_NO_FATAL_FAILURE(InitFramework(m_errorMonitor));
-    const bool maintenance1 = IsExtensionsEnabled(VK_KHR_MAINTENANCE_1_EXTENSION_NAME);
+    const bool maintenance1 =
```
artem-lunarg, line 1233: I did not check dependencies code closely (that's why more eyes are welcomed) but it looks this test was written against Vulkan 1.0 and that's why it has to check for extension. My mental model that `DeviceValidationVersion` will return `VK_API_VERSION_1_0` because `SetTargetApiVersion` was not called to request higher baseline.

spencer-lunarg: >  My mental model that DeviceValidationVersion always returns here VK_API_VERSION_1_0 because SetTargetApiVersion was not called to request a higher baseline.

that is how it should be working, so agree not sure the device setup that this fixes

tests/unit/multiview.cpp
```diff
@@ -1239,6 +1239,11 @@ TEST_F(NegativeMultiview, FeaturesDisabled) {
 
     VkPhysicalDeviceMultiviewFeatures multiview_features = vku::InitStructHelper();
     auto features2 = GetPhysicalDeviceFeatures2(multiview_features);
+
+    if (!multiview_features.multiview) {
+        GTEST_SKIP() << "Multiview support is required";
```
spencer-lunarg, line 1244: if we query for Multiview here, and the device is 1.2, when is this returning false because this is a required feature bit

image

tests/unit/command.cpp
aqnuep: Indeed, you're right. I forgot to also call `SetTargetApiVersion`.

In my excuse I've tested these downstream in the SC VVL where versions 1.0/1.1/1.2 are effectively equivalent as Vulkan SC 1.0 is based on Vulkan 1.2, hence I overlooked that part of the equation. I'll update the PR.

tests/unit/multiview.cpp
aqnuep: Yes, I'm aware, I guess I was somewhat "sneaky" with this one.

In Vulkan SC, multiview optional, despite being in core, so I thought I upstream this change because it wouldn't make a difference for Vulkan (per the above), yet would allow us to not have to replace this test with an SC-specific one.

spencer-lunarg: that is fair, I guess we still have plenty of other places where we still check feature bits even though they are added in core versions, so not against keeping this, just wanted to know how

aqnuep:
I've addressed the comments and rebased, so in case you don't have further concerns I believe this PR should be ready to go.

tests/unit/command.cpp
```diff
@@ -1251,9 +1251,11 @@ TEST_F(NegativeCommand, DrawTimeImageComponentTypeMismatchWithPipeline) {
 TEST_F(NegativeCommand, CopyImageLayerCountMismatch) {
     TEST_DESCRIPTION(
         "Try to copy between images with the source subresource having a different layerCount than the destination subresource");
+    SetTargetApiVersion(VK_API_VERSION_1_1);
     AddOptionalExtensions(VK_KHR_MAINTENANCE_1_EXTENSION_NAME);
     ASSERT_NO_FATAL_FAILURE(InitFramework(m_errorMonitor));
-    const bool maintenance1 = IsExtensionsEnabled(VK_KHR_MAINTENANCE_1_EXTENSION_NAME);
+    const bool maintenance1 =
+        IsExtensionsEnabled(VK_KHR_MAINTENANCE_1_EXTENSION_NAME) || DeviceValidationVersion() >= VK_API_VERSION_1_1;
```
artem-lunarg, line 1258: It took me some time but now I see why we need to do this check - the case when core version does not enable extension but functionality is still available :)

I'd like make one more note. In some cases, in order to simplify such conditions and make tests simpler(since we have a lot of them, simpler tests is easier to quickly validate for correctness) we enforce minimum API version when after `SetTargetApiVersion` we do:
```
if (DeviceValidationVersion() < VK_API_VERSION_1_1 {
        GTEST_SKIP() << "At least Vulkan version 1.1 is required";
}
```
With that setup we often don't need to check for extension and the above check looks the same in all tests, so it's easy validate it. In more complex expressions you can't do just pattern matching and need to explicitly _think_, so for the tests when we don't have additional requirements I think enforcement of minimum version is a good alternative. And disclaimer that I don't know much about Vulkan SC so there's a chance the comment misses important context.

tests/unit/query.cpp
```diff
@@ -970,14 +970,15 @@ TEST_F(NegativeQuery, HostResetFirstQuery) {
...
     TEST_DESCRIPTION("Bad range in vkResetQueryPoolEXT");
 
+    SetTargetApiVersion(VK_API_VERSION_1_1);
     AddRequiredExtensions(VK_EXT_HOST_QUERY_RESET_EXTENSION_NAME);
-
-    m_instance_extension_names.push_back(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
+    AddOptionalExtensions(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
     ASSERT_NO_FATAL_FAILURE(InitFramework(m_errorMonitor));
-    if (!AreRequiredExtensionsEnabled()) {
+    if (!AreRequiredExtensionsEnabled() || (!InstanceExtensionSupported(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME) &&
```
artem-lunarg, line 980: The above comment is mostly above places like this. It takes some time to analyze boolean logic and then 1.0 path probably won't be tested :(
Is there something with Vulkan 1.0 that is important for Vulkan SC? Can we enforce Vulkan 1.1 here?

tests/unit/command.cpp
aqnuep: Yeah, I get that, but here we don't necessarily require Vulkan 1.1 for this test. In fact we don't even need `VK_KHR_maintenance1`. Instead, some test cases expect a different VUID to be triggered based on whether `VK_KHR_maintenance1` is supported (directly, or implicitly through Vulkan 1.1), or maybe the test case checks additional things in case it is supported.

So the conditions I've added are really the correct ones: the test case does not need to be skipped in case the condition is not fulfilled.

tests/unit/query.cpp
aqnuep: No, Vulkan SC doesn't care about Vulkan 1.0 as it is based on Vulkan 1.2, but Vulkan does care, as you'd still want such test cases to be compatible with Vulkan 1.0 implementations.

artem-lunarg: My impression that we have limited ability to test on Vulkan 1.0 devices and untested code in general it's a buggy code. Also it was a common direction how the tests were designed recently by selecting reasonably minimum Vulkan version, so the extension checks are not getting too complicated. 

One example with the above conditions, I checked they are correct, but it took some non trivial time to be sure (a minute or so), and that can be a source of bugs, especially when dealing with a lot of tests and without running CI on real 1.0 devices.

I think that's the area that @spencer-lunarg is more familiar with and he has more experience here to make this decision.

aqnuep: We talked about this with @spencer-lunarg, and what's really missing is a set of profiles to test core 1.0-only functionality with absolute minimum requirements and core 1.3 functionality with all core features. This could be tested with appropriate device profile files on the CI without any actual driver implementations exposing exactly that. So there's a plan in place to make sure these are tested on CI AFAICT.

spencer-lunarg (approved):
This is good as is, got the new profiles from @aqnuep and will add those and hopefully fix anymore spots 

(...`VK_KHR_get_physical_device_properties2`, the extension that will hunt Vulkan 1.0 forever)

PR #6553, opened 2023-09-25
layers: Do not validate dispatchable handles in object tracker

artem-lunarg:
This is similar to https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6509 but applied more broadly.

Do not generate validation code for the function's dispatchable parameter (the first one). Validation of such parameters is always successful based on the model of how the chassis dispactches the calls to `ValidationObject`. If invalid handle is used it will cause crash/corruption on the chassis level (in `get_dispatch_key` or later). And if correct handle is passed, then due to the mapping done by `get_dispatch_key()` the handle will belong to the retrieved validation object, which guarantees positive result of the parenting test. 

Added VVL's project item with idea to improve dispatchable handle validation: https://github.com/orgs/KhronosGroup/projects/22/views/1?pane=issue&itemId=39673708

This is still a prequel to a larger re-work and fixes of object tracking that adds a bunch of new VUIDs. It creates a noise there for the changes that I'd like to keep clearly visible. Also this has importance of its own.

### DIFF
--- a/scripts/generators/object_tracker_generator.py
+++ b/scripts/generators/object_tracker_generator.py
@@ -184,6 +184,10 @@ class ObjectTrackerOutputGenerator(BaseGenerator):
             handle = handle.parent
         return False
 
+    def allComments(self, lines: list[str]) -> bool:
+        not_empty_lines = [line for line in lines.splitlines() if line.strip()]
+        return all(line.startswith('//') for line in not_empty_lines)
+
     def generate(self):
         self.write(f'''// *** THIS FILE IS GENERATED - DO NOT EDIT ***
             // See {os.path.basename(__file__)} for modifications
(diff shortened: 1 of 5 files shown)


arno-lunarg (approved):
LGTM! Less code yeah!

artem-lunarg:
Going to add one more commit here and will re-request the review.
* Need to do the same for a few manual functions, 
* Need to mention removed VUIDs in the comments in order not to change coverage statistics (the convention was that the chassis crash model marks the vuids as validated): https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/9fe2f727998ccac7b107b5d62edc3a71b2867240/layers/object_tracker/object_tracker_utils.cpp#L302)

scripts/generators/object_tracker_generator.py
```diff
@@ -417,6 +420,16 @@ def validateObjects(self, members: list[Member], prefix: str, arrayIndex: int, p
...
                 paramVUID = f'"{vuid_string}"' if vuid_string in self.valid_vuids else "kVUIDUndefined"
 
+                # Do not generate validation code for the function's dispatchable parameter (the first one).
+                # Validation of such parameters is always successful based on the model of how the chassis
+                # dispactches the calls to ValidationObject. If invalid handle is used it will cause
+                # crash/corruption on the chassis level (in get_dispatch_key or later). And if correct handle
+                # is passed, then due to the mapping done by get_dispatch_key() the handle will belong to the
+                # retrieved validation object, which guarantees positive result of the parenting test.
+                # Consider TODO: add validation of function's dispatchable handles at the chassis level during
+                # `get_dispatch_key`, to replace crash/undefined behavior with actual VUID.
```
jeremyg-lunarg, line 430: Thanks for this comment. I think at some point we should change the object model in the chassis so that there's a separate class for each type of dispatchable handle. Not exactly related to what you're discussing here but something to keep in mind if you're going to work in this area.

artem-lunarg:
Addressed the above list.  Coverage was not changed according to `vk_validation_stats.py`.

jeremyg-lunarg (approved):

juan-lunarg (approved):

PR #6555, opened 2023-09-25
layers: Use vku::InitStructHelper

arno-lunarg:
Replace vku::InitStruct calls by vku::InitStructHelper where it makes sense

image


artem-lunarg:
By coincidence tried `InitStructHelper` the first time today. If we'd like to use it as the main solution, should the winner use `vku::InitStruct` name or are there any issues with such renames?

juan-lunarg (approved):

arno-lunarg:
`vku::InitStruct` can be used to fill the fields of the Vulkan struct, just by passing additional arguments, something `vku::InitStructHelper` does not try to do.
Other than that usage, I think `vku::InitStructHelper` is to be prefered now, but that is still up for debate!

PR #6557, opened 2023-09-25
scripts: Make html/csv stats output deterministic

artem-lunarg:
This helps with tracking stats diff.
Text mode was already deterministic.

### DIFF
--- a/scripts/vk_validation_stats.py
+++ b/scripts/vk_validation_stats.py
@@ -497,7 +497,7 @@ static const vuid_spec_text_pair vuid_spec_text[] = {
                     test = 'None'
                     if vuid in self.vt.vuid_to_tests:
                         sep = ', '
-                        test = sep.join(self.vt.vuid_to_tests[vuid])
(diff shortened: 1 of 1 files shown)


scripts/vk_validation_stats.py
```diff
@@ -497,7 +497,7 @@ def dump_csv(self, filename, only_unimplemented=False):
                     test = 'None'
                     if vuid in self.vt.vuid_to_tests:
                         sep = ', '
-                        test = sep.join(self.vt.vuid_to_tests[vuid])
+                        test = sep.join(sorted(list(self.vt.vuid_to_tests[vuid])))
```
mikes-lunarg, line 500: nit: `list(` conversion is unnecessary here. `sorted` can take any iterable type

mikes-lunarg (approved):

juan-lunarg (approved):

scripts/vk_validation_stats.py
artem-lunarg: Fixed

spencer-lunarg (approved):

Issue #4082, opened 2022-05-05
VkSparseMemoryBind missing validation


aitor-lunarg:
Missing VUIDs:
- [ ] VUID-VkSparseMemoryBind-memory-01096
- [ ] VUID-VkSparseMemoryBind-memory-02730
- [ ] VUID-VkSparseMemoryBind-memory-02731

PR #6566, opened 2023-09-28
layers: Add 01096 02730 02731 

arno-lunarg:
- Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4082

Also move the external handle util functions so they can be used elsewhere. I use some of them in the new tests.


tests/framework/layer_validation_tests.h
```diff
@@ -881,3 +857,31 @@ void CreateBufferViewTest(VkLayerTest &test, const VkBufferViewCreateInfo *pCrea
 void CreateImageViewTest(VkLayerTest &test, const VkImageViewCreateInfo *pCreateInfo, const std::string &code = "");
 
 void print_android(const char *c);
+
```
artem-lunarg, line 860: Ironically these helpers previously were in the `layer_validation_tests.h\cpp`. @arno-lunarg @spencer-lunarg @jeremyg-lunarg what do you think about appoarch to keep helper functions in associated test files, and if external test needs that functionality then, maybe, we can introduce a header file with declarations (`external_memory_sync.h`), so `layer_validation_tests.h\cpp` does not grow. I guess in some cases helper functions need to be moved outside of class scope to be visible.

arno-lunarg: Moving them to a dedicated file sounds good to me

artem-lunarg: Yeah, also it might be fine to keep them in the existing cpp file and only to add header if it's not too dirty...

artem-lunarg (approved):
LGTM. Comment about organization of test helpers.

tests/framework/layer_validation_tests.h
spencer-lunarg: I think this is the same issue as the old `positive_tests.cpp` and `core_checks.cpp`... lets just stop putting things in a generic util/garbage bin and make a dedicated file and include it where needed

@arno-lunarg can you just make a `tests/framework/external_memory_sync.h/cpp` file and add them there, I think that is the best solution here

(also happy if you think of something other then `external_memory_sync` you like)

spencer-lunarg (approved):
LGTM - but lets please move things out of `layer_validation_tests`

PR #6572, opened 2023-09-28
Relax descriptor checking disable in GPU-AV

Tony-LunarG:
Only disable descriptor checking if the app actually uses descriptor buffers


jeremyg-lunarg (approved):
LGTM. Glad this was an easy change.

PR #6573, opened 2023-09-28
fix VkExternalMemoryHandleTypeFlags handling

zmike:
`VkExternalMemoryImageCreateInfo` allows passing a bitmask of types, but `VkExternalImageFormatProperties` stupidly allows checking only a single handle type at a time. This should fix all cases where all the handle types passed to image creation aren't correctly passed through validation.

### DIFF
--- a/layers/core_checks/cc_device_memory.cpp
+++ b/layers/core_checks/cc_device_memory.cpp
@@ -1230,6 +1230,7 @@ bool CoreChecks::ValidateBindImageMemory(uint32_t bindInfoCount, const VkBindIma
                     image_info.tiling = image_state->createInfo.tiling;
                     image_info.usage = image_state->createInfo.usage;
                     image_info.flags = image_state->createInfo.flags;
+                    VkExternalMemoryHandleTypeFlags compatible_types = 0;
                     VkExternalImageFormatProperties external_properties = vku::InitStructHelper();
                     VkImageFormatProperties2 image_properties = vku::InitStructHelper(&external_properties);
                     bool export_supported = true;
@@ -1248,7 +1249,9 @@ bool CoreChecks::ValidateBindImageMemory(uint32_t bindInfoCount, const VkBindIma
                         }
                         auto result =
                             DispatchGetPhysicalDeviceImageFormatProperties2(physical_device, &image_info, &image_properties);
-                        if (result != VK_SUCCESS) {
+                        if (result == VK_SUCCESS) {
+                            compatible_types |= flag;
+                        } else {
                             export_supported = false;
                             const LogObjectList objlist(bind_info.image, bind_info.memory);
                             skip |= LogError(
@@ -1300,7 +1303,6 @@ bool CoreChecks::ValidateBindImageMemory(uint32_t bindInfoCount, const VkBindIma
                     IterateFlags<VkExternalMemoryHandleTypeFlagBits>(mem_info->export_handle_types, validate_export_handle_types);
 
                     // The types of external memory handles must be compatible
-                    const auto compatible_types = external_properties.externalMemoryProperties.compatibleHandleTypes;
                     if (export_supported && (mem_info->export_handle_types & compatible_types) != mem_info->export_handle_types) {
                         const LogObjectList objlist(bind_info.image, bind_info.memory);
                         skip |=
--- a/layers/core_checks/cc_image.cpp
+++ b/layers/core_checks/cc_image.cpp
@@ -232,7 +232,6 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
         drm_format_modifier.queueFamilyIndexCount = pCreateInfo->queueFamilyIndexCount;
         drm_format_modifier.pQueueFamilyIndices = pCreateInfo->pQueueFamilyIndices;
         vvl::PnextChainScopedAdd scoped_add_drm_fmt_mod(&image_format_info, &drm_format_modifier);
-        uint32_t bad_index = 0;
 
         if (modifier_list) {
             for (uint32_t i = 0; i < modifier_list->drmFormatModifierCount; i++) {
@@ -240,9 +239,8 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
                 result =
                     DispatchGetPhysicalDeviceImageFormatProperties2(physical_device, &image_format_info, &image_format_properties);
 
-                // The application gives a list of modifier and the driver selects one. If one is wrong, stop there.
-                if (result != VK_SUCCESS) {
-                    bad_index = i;
+                // The application gives a list of modifier and the driver selects one. If one is valid, stop there.
+                if (result == VK_SUCCESS) {
                     break;
                 }
             }
@@ -253,8 +251,7 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
 
         if (result != VK_SUCCESS) {
             // Will not have to worry about VkExternalFormatANDROID if using DRM format modifier
-            std::string drm_source = modifier_list ? ("pDrmFormatModifiers[" + std::to_string(bad_index) + "]")
-                                                   : "VkImageDrmFormatModifierExplicitCreateInfoEXT";
+            std::string drm_source = modifier_list ? "pDrmFormatModifiers[]" : "VkImageDrmFormatModifierExplicitCreateInfoEXT";
             skip |= LogError("VUID-VkImageCreateInfo-imageCreateMaxMipLevels-02251", device, create_info_loc,
                              "The following parameters -\n"
                              "format (%s)\n"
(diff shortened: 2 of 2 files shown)


juan-lunarg (commented):
>  This should fix all cases where all the handle types passed to image creation aren't correctly passed through validation.

Add test(s) to prevent a regression.

zmike:
Probably don't have time to figure out how tests work in VVL. I'm happy to turn this over to you if you want to finish things up.

layers/core_checks/cc_device_memory.cpp
```diff
@@ -1248,7 +1249,9 @@ bool CoreChecks::ValidateBindImageMemory(uint32_t bindInfoCount, const VkBindIma
                         }
                         auto result =
                             DispatchGetPhysicalDeviceImageFormatProperties2(physical_device, &image_info, &image_properties);
-                        if (result != VK_SUCCESS) {
+                        if (result == VK_SUCCESS) {
+                            compatible_types |= flag;
```
artem-lunarg, line 1253: Thanks for the PR! 

I think the reason why `NegativeExternalMemorySync.ImageMemoryWithIncompatibleHandleTypes` test fails is because the results of compatibility queries should be intersected. There is also this part of spec that says the same in a bit convoluted form:
> compatibleHandleTypes must include at least handleType. Inclusion of a handle type in compatibleHandleTypes does not imply the values returned in [VkImageFormatProperties2](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkImageFormatProperties2) will be the same when [VkPhysicalDeviceExternalImageFormatInfo](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkPhysicalDeviceExternalImageFormatInfo)::handleType is set to that type. The application is responsible for querying the capabilities of all handle types intended for concurrent use in a single image and intersecting them to obtain the compatible set of capabilities.

Locally I did: 
```
VkExternalMemoryHandleTypeFlags compatible_types = vvl::kU32Max;

if (result == VK_SUCCESS) {
      compatible_types &= flag;
 }
```

and it helped with at least one test failure, did not check if it fixes everything.

artem-lunarg:
I worked on this part of the code I can try to figure out why internal CI is failing (I had initial idea, but it's probably wrong, need to dig deeper and refresh memory). `NegativeExternalMemorySync.ImageMemoryWithIncompatibleHandleTypes` also fails on Windows.

zmike:
Oh, I didn't realize there were tests to be run. You're correct about that intersection bit in the spec. I'd expect applying the same logic everywhere should resolve the fails?

artem-lunarg:
Sorry, I removed that message, because I think I misunderstood some things.

Basically in the successful scenario without failures in `DispatchGetPhysicalDeviceImageFormatProperties2` the `compatible_types |= flag;` part reconstructs exactly the first parameter of this call (lamda iterates over all flags):
```
IterateFlags<VkExternalMemoryHandleTypeFlagBits>(mem_info->export_handle_types, validate_export_handle_types);
``` 

So in the modified solution
```
compatible_types is mem_info->export_handle_types
```
Which is different from the original code where:
```
compatible_types is external_properties.externalMemoryProperties.compatibleHandleTypes
```

Insight for the reason of the test failure on some configurations. For example, on my windows machine ` mem_info->export_handle_types` is 134 (128 + 4 + 2).
But `external_properties.externalMemoryProperties.compatibleHandleTypes`, which is the compatibility flags from the last iteration (as you mentioned!) is 128. It means that the handleType 128 is compatible only with itself. Not sure if flags 2 and 4 form a compatibility group or each of them also compatible with only itself. Anyway 134 is not a compatible set of flags, if it was then 4 and 2 were in the compatibility set of 128, but they are not.

Can dig further tomorrow since it's getting late here. But hopefully this shows that ORs is not a good option. Also now I'm not sure about ANDing too  - if we AND compatible sets `external_properties.externalMemoryProperties.compatibleHandleTypes` on each iteration we probably get zeroL: (128 & 4 & 2) == 0).

Need to think about the original meaning of the code. One option is that the original code was correct (only what is related to the first commit, did not check second yet), due to some guarantees for compatibility. Again, will check that probably tomorrow.

artem-lunarg:
Okay, I'm pretty sure the original code was correct:

```
// Here we got compatibleHandleTypes from the last iteration but it does not matter, it can be 
// any iteration. Kind of mathematical proof further.
const auto compatible_types = external_properties.externalMemoryProperties.compatibleHandleTypes;

if (export_supported && (mem_info->export_handle_types & compatible_types) != mem_info->export_handle_types)   {
 ....
```

The task here is to check if `mem_info->export_handle_types` flags form a compatibility group.
Lets consider two possibilities.

Option A. `mem_info->export_handle_types` form a compatibility group. In this case if we query `compatibleHandleTypes` using arbitrary flag from `mem_info->export_handle_types` the returned `compatibleHandleTypes` should include all `mem_info->export_handle_types` bits (and maybe some others), otherwise `mem_info->export_handle_types` do not form compatibility group which violates assumption of Option A. It  means it does not matter which flag to choose.

Option B. `mem_info->export_handle_types` do not form a compatibility group. The same reasoning. If it would be possible for at least one flag from `mem_info->export_handle_types` to return `compatibleHandleTypes` that includes all bits from  `mem_info->export_handle_types` then it means `mem_info->export_handle_types` forms a compatibility group which is violation of assumption of option B. So compatibilities queries for any bit of `mem_info->export_handle_types` return the same result that `mem_info->export_handle_types` flags are not compatibile. Again it does not matter which flag to choose.

artem-lunarg:
This part of spec:
> compatibleHandleTypes must include at least handleType. Inclusion of a handle type in compatibleHandleTypes does not imply the values returned in [VkImageFormatProperties2](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkImageFormatProperties2) will be the same when [VkPhysicalDeviceExternalImageFormatInfo](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkPhysicalDeviceExternalImageFormatInfo)::handleType is set to that type. The application is responsible for querying the capabilities of all handle types intended for concurrent use in a single image and intersecting them to obtain the compatible set of capabilities.

I think it's mostly what to expect from returned properties.

One example.
Flag 2 reports compatibility with flags 4, 32, 64.
I think the above part of spec says that using 64 to query compatibility does not necessarily will return exactly the same properties as querying 2. `compatibleHandleTypes` can be the same (kind of have to be for commutative/transitive logic) but other fields of `VkExternalMemoryProperties ` or `VkImageFormatProperties2` can contain different values for such queries, in that case some form for intersection or filtering should be done.

zmike:
Hm interesting. So it seems we agree that the issue is effectively what I've done in this PR, it's just a question of whether the bug is in VVL or the driver.

I could believe that this is a driver issue (i.e., the driver must report all supported `VkExternalMemoryHandleTypeFlags` for a given query and one of the bits must be the one that was queried), but this seems like an insane set of mental gymnastics required by the spec. Effectively:
* you query support for `X`
* the driver then returns `X|Y|Z`
* you must then query separately `Y` and `Z` to determine if they are truly supported, even though the driver already returned support for them

I think I'm going to raise a spec issue to confirm that this is intended, as it seems to me this goes against the explicit nature of the rest of the vulkan API.

artem-lunarg:
> but this seems like an insane set of mental gymnastics required by the spec

That also was my conclusion. It's non-trivial API and I would be happy if it's possible to make it simpler only with minor adjustments without re-working it completely. Could be the price for the generalization.

The part with different properties and finding intersection is probably one of the most confused places, but maybe that's expected. Compatibility says you can use all these handle types, but they can have different nature, so please don't expect that, for example, they will support the same image properties. That's how I understand this.

With all these complexity the drivers we tested (desktop windows and linux and android) respected the rules for compatibility. One thing that can be helpful. During implementing tests I had few occasions when a driver reported results that violated the spec. Often that was on android (could be on other platforms too, don't remember exactly) and in that case I made conclusion that driver is broken. For example, after querying compatibility set for a given handle, the same handle was not in the returned compatibility set and it should be the case according to spec. In most such cases I later found that I also violated some other spec rule but we did not have validation for it so we did not know. After fixing another broken part everything worked correctly. One such case was with dedicated allocations. IIRC, some handle types require dedicated allocations and we ignored that check at first. So when we did queries for handles with such requirements driver returned seemingly bad data. By checking for dedicated flags we rejected those handle types early (or ensured we have dedicated allocation), so in the end the there were not issue with the driver. 

So yeah, the impression so far that drivers worked more or less correctly and it's hard to implement well, required separate iterations, and that's only on the VVL side, I never used this functionality on the application side.

zmike:
I'm not so sure drivers are working correctly even if this is the expected behavior. For example, I'm creating images with `VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD_BIT|VK_EXTERNAL_MEMORY_HANDLE_TYPE_DMA_BUF_BIT_EXT`, which is supported by the driver for a given set of parameters, but when querying for support only the queried bit is returned, triggering every validation error in existence.

artem-lunarg:
> when querying for support only the queried bit is returned

Could you please elaborate on this. Terminology is a bit confusing since there are supported handle types but supported handle types are not necessarily form a compatibility group and can't be used together, only separately. I also had use cases when handle type is compatible only with itself.  Then such handle type should be specified alone, otherwise yes it causes validation error. Is it your scenario or something different?

PR #6574, opened 2023-09-29
layers: Fix modifier selection validation

zmike:
when a list of modifiers is passed, only one of them must be valid, not all of them

successor to https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6573

### DIFF
--- a/layers/core_checks/cc_image.cpp
+++ b/layers/core_checks/cc_image.cpp
@@ -232,7 +232,6 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
         drm_format_modifier.queueFamilyIndexCount = pCreateInfo->queueFamilyIndexCount;
         drm_format_modifier.pQueueFamilyIndices = pCreateInfo->pQueueFamilyIndices;
         vvl::PnextChainScopedAdd scoped_add_drm_fmt_mod(&image_format_info, &drm_format_modifier);
-        uint32_t bad_index = 0;
 
         if (modifier_list) {
             for (uint32_t i = 0; i < modifier_list->drmFormatModifierCount; i++) {
@@ -240,9 +239,8 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
                 result =
                     DispatchGetPhysicalDeviceImageFormatProperties2(physical_device, &image_format_info, &image_format_properties);
 
-                // The application gives a list of modifier and the driver selects one. If one is wrong, stop there.
-                if (result != VK_SUCCESS) {
-                    bad_index = i;
+                // The application gives a list of modifier and the driver selects one. If one is valid, stop there.
+                if (result == VK_SUCCESS) {
                     break;
                 }
             }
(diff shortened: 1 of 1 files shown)


layers/core_checks/cc_image.cpp
```diff
@@ -502,10 +499,10 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
                     drm_format_modifier.drmFormatModifier = modifier_list->pDrmFormatModifiers[i];
                     result =
                         DispatchGetPhysicalDeviceImageFormatProperties2(physical_device, &image_format_info, &image_properties);
-                    compatible_types |= external_image_properties.externalMemoryProperties.compatibleHandleTypes;
-                    if (result != VK_SUCCESS)
-                        break;
+                    if (result == VK_SUCCESS)
+                        compatible_types |= external_image_properties.externalMemoryProperties.compatibleHandleTypes;
```
artem-lunarg, line 503: Can we merge `compatibleHandleTypes` flags from different queries?
If 1 compatible with 2, and 2 compatible with 4, then after ORing we have 1 is compatible with 4.

Is it invalid logic here to break the loop after the first success?

zmike: breaking there would be fine, sure. And I suppose probably it should just be `=`

zmike: Ah, actually I did that to aggregate the handletypes in case there was a modifier which only supported some handle types while another modifier supported more. This is not necessarily a realistic example, but I think I'll need to do more refactoring to handle it better.

zmike: Or maybe this is enough...

artem-lunarg: Thanks for the explanation. If you'd like to keep aggregation with `|=` could you please add the above explanation as a comment, so it's clear it's intentional, since in a basic usage of this API (without DRM features) such concatenation of compatibility flags can result of unsupported combination.

layers/core_checks/cc_image.cpp
```diff
@@ -502,10 +499,13 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
                     drm_format_modifier.drmFormatModifier = modifier_list->pDrmFormatModifiers[i];
                     result =
                         DispatchGetPhysicalDeviceImageFormatProperties2(physical_device, &image_format_info, &image_properties);
-                    compatible_types |= external_image_properties.externalMemoryProperties.compatibleHandleTypes;
-                    if (result != VK_SUCCESS)
-                        break;
+                    if (result == VK_SUCCESS) {
+                        compatible_types = external_image_properties.externalMemoryProperties.compatibleHandleTypes;
+                        if ((external_memory_create_info->handleTypes & compatible_types) != external_memory_create_info->handleTypes)
```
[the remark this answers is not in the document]
zmike: yeah good catch :face_with_head_bandage:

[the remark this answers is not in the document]
artem-lunarg: one more nit, it looks there was formatting error on CI, if you have opportunity could you format with clang-format, otherwise we can do this.

artem-lunarg (approved):
LGTM!

PR #6577, opened 2023-09-29
cmake: Begin to allow clang on Windows

juan-lunarg:
Forwards the CMAKE_LANG_COMPILER variables in order to build everything properly

Sets WIN32 defines properly. Since you can use clang on Windows.

Fix warning found by clang on Windows regarding unused variable.


juan-lunarg:
I've been stuck on Windows and have been dealing with some compiler warnings/issues that MSVC isn't catching. As a result I figured out an easy way to run clang on Windows via vscode.

There are still some linker issues to sort out but overall it should be a nice quality of life feature for Windows devs looking to replicate linux issues.

image

scripts/CMakeLists.txt
```diff
@@ -67,6 +67,9 @@ if (UPDATE_DEPS)
     endif()
 
     list(APPEND cmake_vars "CMAKE_TOOLCHAIN_FILE")
+    if (NOT MSVC_IDE)
+        list(APPEND cmake_vars "CMAKE_CXX_COMPILER" "CMAKE_C_COMPILER" "CMAKE_ASM_COMPILER")
+    endif()
```
juan-lunarg, line 72: If you don't forward these variables then the dependencies will be built with msvc. clang doesn't like this and will error out.

Which is kinda weird since mixing/matching binaries compiled by gcc/clang works fine. But Windows is a different story.

scripts/generators/spirv_grammar_generator.py
```diff
@@ -550,13 +550,11 @@ def generateSource(self):
 
             std::string string_SpvCooperativeMatrixOperands(uint32_t mask) {{
                 std::string ret;
-                int index = 0;
```
juan-lunarg, line 553: Not sure why CI didn't catch this. Most likely my version of clang is newer than what CI is using.

scripts/CMakeLists.txt
```diff
@@ -67,6 +67,9 @@ if (UPDATE_DEPS)
     endif()
 
     list(APPEND cmake_vars "CMAKE_TOOLCHAIN_FILE")
```
artem-lunarg, line 69: is `cmake_vars` our custom variable?

artem-lunarg (approved):

scripts/CMakeLists.txt
juan-lunarg: > is cmake_vars our custom variable?

Yes.

PR #6583, opened 2023-09-29
Add vuids 07299 07300 07301 07302, validate mesh work groups

ziga-lunarg:


layers/core_checks/cc_spirv.cpp
```diff
@@ -2706,3 +2709,63 @@ bool CoreChecks::ValidateTaskMeshWorkGroupSizes(const SPIRV_MODULE_STATE &module
...
     return skip;
 }
+
+bool CoreChecks::ValidateEmitMeshTasksSize(const SPIRV_MODULE_STATE &module_state, const EntryPoint &entrypoint,
+                                           const Location &loc) const {
+    bool skip = false;
+
+    for (const Instruction &insn : module_state.static_data_.instructions) {
+        if (insn.Opcode() == spv::OpEmitMeshTasksEXT) {
+            uint32_t x = module_state.FindDef(insn.Word(1))->Word(3);
```
artem-lunarg, line 2719: Just learning how spirv validation works, do we have a guarantee that Word(1) references to `OpConstant`? We also have more general `GetIntConstantValue` can it be used here?

layers/core_checks/cc_spirv.cpp
```diff
@@ -2706,3 +2709,63 @@ bool CoreChecks::ValidateTaskMeshWorkGroupSizes(const SPIRV_MODULE_STATE &module
...
+                                 "SPIR-V (%s) is emitting %" PRIu32
+                                 " mesh work groups in Z dimension, which is greater than max mesh "
+                                 "workgroup count (%" PRIu32 ").",
+                                 string_SpvExecutionModel(entrypoint.execution_model), z,
+                                 phys_dev_ext_props.mesh_shader_props_ext.maxMeshWorkGroupCount[2]);
+            }
+            uint64_t invocations = static_cast<uint64_t>(x) * static_cast<uint64_t>(y);
+            // Prevent overflow.
+            bool fail = false;
+            if (invocations > vvl::kU32Max || invocations > phys_dev_ext_props.mesh_shader_props_ext.maxMeshWorkGroupTotalCount) {
```
artem-lunarg, line 2749: `invocations > vvl::kU32Max` check is not needed. 
If `invocations > vvl::kU32Max` then we automatically have that `invocations > phys_dev_ext_props.mesh_shader_props_ext.maxMeshWorkGroupTotalCount` so only the second check is enough.

layers/core_checks/cc_spirv.cpp
ziga-lunarg: You are correct, fixed

artem-lunarg (approved):

PR #6588, opened 2023-09-30
Add VUID 07480, validate color attachments with advanced blend

ziga-lunarg:
Part of #5507

### DIFF
--- a/layers/core_checks/cc_cmd_buffer_dynamic.cpp
+++ b/layers/core_checks/cc_cmd_buffer_dynamic.cpp
@@ -389,6 +389,30 @@ bool CoreChecks::ValidateDrawDynamicStatePipeline(const LAST_BOUND_STATE& last_b
                              color_index);
         }
     }
+    if (pipeline.IsDynamic(VK_DYNAMIC_STATE_COLOR_BLEND_ADVANCED_EXT) &&
+        pipeline.IsDynamic(VK_DYNAMIC_STATE_COLOR_BLEND_ENABLE_EXT)) {
+        const uint32_t attachment_count =
+            cb_state.activeRenderPass->UsesDynamicRendering()
+                ? cb_state.activeRenderPass->GetDynamicRenderingColorAttachmentCount()
+                : cb_state.activeRenderPass->createInfo.pSubpasses[cb_state.GetActiveSubpass()].colorAttachmentCount;
+
+        bool advanced_blend = false;
+        for (uint32_t i = 0; i < attachment_count; ++i) {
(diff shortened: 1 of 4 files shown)


arno-lunarg (approved):

layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -389,6 +389,30 @@ bool CoreChecks::ValidateDrawDynamicStatePipeline(const LAST_BOUND_STATE& last_b
                              color_index);
         }
     }
+    if (pipeline.IsDynamic(VK_DYNAMIC_STATE_COLOR_BLEND_ADVANCED_EXT) &&
+        pipeline.IsDynamic(VK_DYNAMIC_STATE_COLOR_BLEND_ENABLE_EXT)) {
+        uint32_t attachment_count =
```
arno-lunarg, line 394:
```suggestion
        const uint32_t attachment_count =
```

arno-lunarg (approved):

spencer-lunarg (approved):

layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -389,6 +389,30 @@ bool CoreChecks::ValidateDrawDynamicStatePipeline(const LAST_BOUND_STATE& last_b
                              color_index);
         }
     }
+    if (pipeline.IsDynamic(VK_DYNAMIC_STATE_COLOR_BLEND_ADVANCED_EXT) &&
+        pipeline.IsDynamic(VK_DYNAMIC_STATE_COLOR_BLEND_ENABLE_EXT)) {
+        const uint32_t attachment_count =
+            cb_state.activeRenderPass->UsesDynamicRendering()
+                ? cb_state.activeRenderPass->GetDynamicRenderingColorAttachmentCount()
+                : cb_state.activeRenderPass->createInfo.pSubpasses[cb_state.GetActiveSubpass()].colorAttachmentCount;
```
spencer-lunarg, line 397: > the number of active color attachments in the current subpass 

so I am curious if you copied this logic from somewhere else or just created it here... I think we might need to cleanup/unify how we get these values now that there is dynamic rendering... worried we might be silently missing dynamic or non-dynamic rendering edge cases throughout the code

ziga-lunarg: I created this here. `GetDynamicRenderingColorAttachmentCount()` is a nice function, we could add `GetColorAttachmentCount` that would handle both.

PR #6591, opened 2023-10-01
Add VUID 07073, validate mesh shader queries

ziga-lunarg:


layers/core_checks/cc_drawdispatch.cpp
```diff
@@ -4101,6 +4115,7 @@ struct DispatchVuidsCmdDrawIndirectByteCountEXT: DrawDispatchVuid {
         set_exclusive_scissor_enable_09235       = "VUID-vkCmdDrawIndirectByteCountEXT-exclusiveScissor-09235";
         set_fragment_shading_rate_09238          = "VUID-vkCmdDrawIndirectByteCountEXT-pipelineFragmentShadingRate-09238";
         set_tessellation_domain_origin_09237     = "VUID-vkCmdDrawIndirectByteCountEXT-None-09237";
+        mesh_shader_queries_07073                = "VUID-vkCmdDrawIndirectByteCountEXT-stage-07073";
```
artem-lunarg, line 4118: It looks we don't have VUID sections for some NV and HUAWEI commands. In case it's not intentional, can we add a future TODO comment to add missing sections, so the following VUIDs will be addressed ultimately:

VUID-vkCmdDrawClusterHUAWEI-stage-07073
VUID-vkCmdDrawClusterIndirectHUAWEI-stage-07073
VUID-vkCmdExecuteGeneratedCommandsNV-stage-07073

layers/core_checks/cc_drawdispatch.cpp
```diff
@@ -5883,6 +5898,27 @@ bool CoreChecks::ValidateActionState(const CMD_BUFFER_STATE &cb_state, const VkP
 
     skip |= ValidateCmdRayQueryState(cb_state, bind_point, loc);
 
+    if (pipeline) {
+        bool rasterization_stage = false;
```
artem-lunarg, line 5902: naming nit: mesh shading pipeline also performs rasterization, maybe something like `primitive_shading` to match the spec language:

> Drawing can be achieved in two modes:
[Programmable Mesh Shading](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#drawing-mesh-shading), the mesh shader assembles primitives, or
[Programmable Primitive Shading](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#drawing-primitive-shading), the input primitives are assembled as follows.

layers/core_checks/cc_drawdispatch.cpp
```diff
@@ -5883,6 +5898,27 @@ bool CoreChecks::ValidateActionState(const CMD_BUFFER_STATE &cb_state, const VkP
...
+            if ((stage.GetStage() & (VK_SHADER_STAGE_VERTEX_BIT | VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT |
+                                     VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT | VK_SHADER_STAGE_GEOMETRY_BIT)) != 0) {
+                rasterization_stage = true;
+                break;
+            }
+        }
+        if (rasterization_stage) {
+            for (const auto &query : cb_state.activeQueries) {
+                const auto query_pool_state = Get<QUERY_POOL_STATE>(query.pool);
+                if (query_pool_state->createInfo.queryType == VK_QUERY_TYPE_MESH_PRIMITIVES_GENERATED_EXT) {
```
artem-lunarg, line 5913: This validates something else than is asked in `07073`

layers/core_checks/cc_drawdispatch.cpp
```diff
@@ -5883,6 +5898,27 @@ bool CoreChecks::ValidateActionState(const CMD_BUFFER_STATE &cb_state, const VkP
...
+                    const LogObjectList objlist(cb_state.commandBuffer(), query.pool);
+                    skip |= LogError(vuid.mesh_shader_queries_07073, objlist, loc,
```
artem-lunarg, line 5915: `query.slot` in the message can be useful to identify specific query.

artem-lunarg (approved):
LGTM, some suggestions

layers/core_checks/cc_drawdispatch.cpp
ziga-lunarg: We don't have any validation for those 3 commands, so the other 100 or more draw VUIDs are also missing for these.

layers/core_checks/cc_drawdispatch.cpp
```diff
@@ -5883,6 +5898,28 @@ bool CoreChecks::ValidateActionState(const CMD_BUFFER_STATE &cb_state, const VkP
 
     skip |= ValidateCmdRayQueryState(cb_state, bind_point, loc);
 
+    if (pipeline) {
+        bool primitive_shading = false;
+        for (const auto &stage : pipeline->stage_states) {
+            if ((stage.GetStage() & (VK_SHADER_STAGE_VERTEX_BIT | VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT |
```
spencer-lunarg, line 5904: there is a `pipeline->create_info_shaders` that can be used instead of having to loop for these stages

spencer-lunarg (approved):

PR #6592, opened 2023-10-01
Validate query was reset before getting result and best practice that it is available

### LINKED ISSUE #6129: Detect that no query has been sent with BeginQuery/EndQuery before when calling GetQueryPoolResults
When implementing a video encoder, I encountered a dead lock in the driver when querying the encoding result without calling Begin/End combo before.
I see that there is a check in the [EndQuery](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/2ce44f7f1928edb981adc587d8d3e2fb0f41e975/layers/core_checks/cc_query.cpp#L761), could we add this check in [GstQueryPoolResult](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/2ce44f7f1928edb981adc587d8d3e2fb0f41e975/layers/core_checks/cc_query.cpp#L113) ?

**Valid Usage IDs requested**
Check that a query has been sent in GstQueryPoolResult

**Additional Context**

ziga-lunarg:
Closes #6129


arno-lunarg (approved):
LGTM!

layers/core_checks/cc_query.cpp
```diff
@@ -91,6 +101,28 @@ bool CoreChecks::ValidatePerformanceQueryResults(const QUERY_POOL_STATE &query_p
...
+    bool skip = false;
+
+    for (uint32_t i = firstQuery; i < firstQuery + queryCount; ++i) {
+        // static QueryState GetLocalQueryState(const QueryMap *localQueryToStateMap, VkQueryPool queryPool, uint32_t queryIndex, uint32_t perfPass) {
+        if (localQueryToStateMap &&
+            GetLocalQueryState(localQueryToStateMap, query_pool_state.pool(), i, perfPass) != QUERYSTATE_UNKNOWN) {
+            continue;
+        }
+        if (query_pool_state.GetQueryState(i, 0u) == QUERYSTATE_UNKNOWN) {
+            skip |= LogError(kVUID_Core_QueryPool_NotReset, query_pool_state.pool(), loc.dot(Field::queryPool),
```
spencer-lunarg, line 115: I haven't looked through the spec if this is a VU already, but we need to get a MR in GitLab to add this VU or else it will sit here as "unassigned" forever

layers/core_checks/cc_query.cpp
```diff
@@ -91,6 +101,28 @@ bool CoreChecks::ValidatePerformanceQueryResults(const QUERY_POOL_STATE &query_p
     return skip;
 }
 
+bool CoreChecks::ValidateQueryPoolWasReset(const QUERY_POOL_STATE &query_pool_state, uint32_t firstQuery, uint32_t queryCount,
+                                           const Location &loc, QueryMap *localQueryToStateMap, uint32_t perfPass) const {
+    bool skip = false;
+
+    for (uint32_t i = firstQuery; i < firstQuery + queryCount; ++i) {
+        // static QueryState GetLocalQueryState(const QueryMap *localQueryToStateMap, VkQueryPool queryPool, uint32_t queryIndex, uint32_t perfPass) {
```
spencer-lunarg, line 109: think this can be removed?

tests/unit/query.cpp
```diff
@@ -2780,6 +2801,41 @@ TEST_F(NegativeQuery, PerformanceQueryReset) {
...
+    VkQueryPoolCreateInfo query_pool_create_info = vku::InitStructHelper();
+    query_pool_create_info.queryType = VK_QUERY_TYPE_OCCLUSION;
+    query_pool_create_info.queryCount = 1;
+    vkt::QueryPool query_pool(*m_device, query_pool_create_info);
+
+    vkt::Buffer buffer(*m_device, 128, VK_BUFFER_USAGE_TRANSFER_DST_BIT, VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT);
+
+    uint32_t data = 0u;
+    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "UNASSIGNED-CoreValidation-QueryPool-NotReset");
+    vk::GetQueryPoolResults(*m_device, query_pool.handle(), 0u, 1u, sizeof(uint32_t), &data, sizeof(uint32_t), 0u);
```
spencer-lunarg, line 2823: > If VK_QUERY_RESULT_WAIT_BIT is not set, vkGetQueryPoolResults may return VK_NOT_READY if there are queries in the unavailable state.

also the comment https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6129#issuecomment-1632859513

I want to confirm in this weeks WG meeting - https://gitlab.khronos.org/vulkan/vulkan/-/issues/3638

ziga-lunarg: Can you clarify what you mean with the first part of the comment? Your quote talks about `unavailable state` which is not the case here.

I will wait for the spec clarification before merging.

spencer-lunarg: We agreed this is invalid and your PR is correct, as for the VUID, will get that after

spencer-lunarg (approved):

PR #6595, opened 2023-10-02
cmake: Allow add_subdirectory usage

### LINKED ISSUE #6587: build: CMake targets of dependencies are not respected
Available targets of dependencies in a CMake build are currently not respected. VVL only integrates them via find_package in config mode. 

See top level CMakeLists.txt @ line 39
```
 find_package(VulkanHeaders REQUIRED CONFIG QUIET)
```
This prevents building VVL in one CMake project together with the Loader and Headers in a  convenient way. If VVL would first check for available targets in the overarching project like so:

```
if(NOT TARGET Vulkan::Headers)
    find_package(VulkanHeaders REQUIRED CONFIG QUIET)
endif()
```
this could be possible. 
In fact that was the case in older versions e.g. v1.2.154. Currently one has to install the Vulkan-Headers and point CMake to the install directory before configuring the VVL CMake project so the package configs can be found by VVL. This is also the case for other dependencies like SPIRV-Headers and -Tools. 

Is there a reason this behavior got changed from older versions and no longer makes sense to support? And what is the current intended way of building parts of the Vulkan SDK from source?

edit:
A first step towards that functionality was made in 3bf912e

juan-lunarg:
Basically the simple fix it to not use the REQUIRED keyword.

Similar to BUILD_WERROR being OFF by default this makes our CMake build more permissive for users to do what they want.

closes #6587


layers/CMakeLists.txt
```diff
@@ -102,47 +102,6 @@ elseif(MSVC)
     )
 endif()
 
-# Represents all SPIRV libraries we need
-add_library(VVL-SPIRV-LIBS INTERFACE)
```
juan-lunarg, line 106: This target was created to deal with the BUILD_LAYER_SUPPORT_FILES option we had. It's no longer needed and actually complicates the code.

spencer-lunarg (approved):

scripts/CMakeLists.txt
juan-lunarg, line 1: In order to have better error checking for us I've added the usage of this relatively new CMake feature:
https://cmake.org/cmake/help/latest/variable/CMAKE_REQUIRE_FIND_PACKAGE_PackageName.html

arno-lunarg (approved):

PR #6596, opened 2023-10-02
tests: OneOffDescriptorSet update

artem-lunarg:
Some cleanup. Removes the need to manually adjust descriptor write thresholds in `OneOffDescriptorSet` (or theoretical possibility to go over default limits, which was 10). When write count goes over the limit then due to reallocation `vkUpdateDescriptorSets` works with dangled pointers and that's not always manifest itself in an immediate crash.

The second commit could be more subjective. The semantics of `OneOffDescriptorSet` methods that accept element count was a bit confusing. In Vulkan, element count means that multiple resources are consumed, and in our helpers, array count means a single resource will be duplicated this number of times.

Updated helpers remove element count parameter which duplicated writes. Single element is a majority of cases

If we need to specify multiple elements, then new overloads can be added, with semantics more similar to Vulkan - element count and pointer to array of resources.


spencer-lunarg (approved):
LGTM! (hope CI is happy too)

PR #6599, opened 2023-10-02
Cache instrumented shaders when running with GPU-AV

Tony-LunarG:
Cache instrumented shaders to disk to keep from having to instrument them on every run.

### DIFF
--- a/layers/VkLayer_khronos_validation.json.in
+++ b/layers/VkLayer_khronos_validation.json.in
@@ -930,6 +930,26 @@
                                             }
                                         },
                                         {
+                                            "key": "use_instrumented_shader_cache",
+                                            "label": "Cache instrumented shaders rather than instrumenting them on every run",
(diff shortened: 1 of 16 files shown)


jeremyg-lunarg (approved):
LGTM! Have you run this on any of the traces yet or do you need me to do that on the perf machine?

Tony-LunarG:
> LGTM! Have you run this on any of the traces yet or do you need me to do that on the perf machine?

I've only tried Doom Eternal 370 frames (which is trimmed and wouldn't reflect savings in FPS) and the surge, which went from 15.6 to 16.2 fps.  The Jenkins job didn't see anything. If you have the cycles to try it out, that would be helpful.

jeremyg-lunarg:
I ran it last night and there wasn't a huge perf difference so go for it if you think this is ready to merge

Tony-LunarG:
I'm going to run CTS on it tonight, then merge if nothing found

PR #6604, opened 2023-10-03
Second set of performance optimizations for synchronization

jzulauf-lunarg:
A series of small hotspot removals.

### DIFF
--- a/layers/state_tracker/descriptor_sets.h
+++ b/layers/state_tracker/descriptor_sets.h
@@ -339,6 +339,10 @@ class DescriptorSet;
 
 class Descriptor {
   public:
+    static bool SupportsNotifyInvalidate() { return false; }
+    static bool IsNotifyInvalidateType(VulkanObjectType) { return false; }
+    virtual void InvalidateNode(const std::shared_ptr<BASE_NODE> &, bool) {}  // Most descriptor types will not call
+
     Descriptor() {}
     virtual ~Descriptor() {}
     virtual void WriteUpdate(DescriptorSet &set_state, const ValidationStateTracker &dev_data, const VkWriteDescriptorSet &,
@@ -407,6 +411,10 @@ class SamplerDescriptor : public Descriptor {
 
 class ImageDescriptor : public Descriptor {
   public:
+    static bool SupportsNotifyInvalidate() { return true; }
+    static bool IsNotifyInvalidateType(const VulkanObjectType node_type) {
+        return node_type == VulkanObjectType::kVulkanObjectTypeImageView;
+    }
     ImageDescriptor() = default;
     DescriptorClass GetClass() const override { return Image; }
     void WriteUpdate(DescriptorSet &set_state, const ValidationStateTracker &dev_data, const VkWriteDescriptorSet &, const uint32_t,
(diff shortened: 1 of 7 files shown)


layers/state_tracker/descriptor_sets.h
```diff
@@ -407,6 +411,10 @@ class SamplerDescriptor : public Descriptor {
 
 class ImageDescriptor : public Descriptor {
   public:
+    static bool SupportsNotifyInvalidate() { return false; }
```
artem-lunarg, line 414: Should it be `return true` here?

layers/state_tracker/descriptor_sets.h
```diff
@@ -432,12 +440,24 @@ class ImageDescriptor : public Descriptor {
...
+            }
+        }
+    }
 
-    bool Invalid() const override { return !image_view_state_ || image_view_state_->Invalid(); }
+    bool Invalid() const override { return !known_valid_view_ && ComputeInvalid(); }
 
   protected:
+    bool ComputeInvalid() const { return !image_view_state_ || image_view_state_->Invalid(); }
+    void UpdateKnownValidView(bool is_bindless) { known_valid_view_ = !is_bindless && !ComputeInvalid(); }
```
artem-lunarg, line 456: Comparing to the old implementation it looks that there could be a change in behavior related to `is_bindless` processing. The following are the details how I understood this part  and which part differs in behavior.

In the new implementation when we Write or Copy image descriptor for which `is_bindless == true`, we evaluate `known_valid_view_` to be `false` (when `UpdateKnownValidView` is called at the end of  Update/Copy).


In the old implementation which did not call `UpdateKnownValidView` we had only `ReplaceStatePtr` call, which does something like this:
```
    if (dst && !is_bindless) {
        dst->RemoveParent(&set_state);
    }
    dst = src;
    if (dst && !is_bindless) {
        dst->AddParent(&set_state);
    }
```

So,  when `ImageDescriptor::WriteUpdate` called `ReplaceStatePtr(set_state, image_view_state_..., is_bindless` then for the bindless case we did not update parent but we still upate `image_view_state_` in assignment ` dst = src;`

Then calling ImageDescriptor::Invalid() in the old implementation could return true in case the updated image state `image_view_state_->Invalid()` returns `true`, but in the new implementation `known_valid_view_` was compute as false (bindless case), so `Invalid()` return `false` which is a change in behavior (not sure if it's fine or not).

layers/sync/sync_validation.cpp
```diff
@@ -3260,9 +3260,12 @@ HazardResult ResourceAccessState::DetectHazard(const ResourceAccessState &record
     const auto &recorded_accesses = recorded_use.first_accesses_;
     Size count = recorded_accesses.size();
     if (count) {
-        const auto &last_access = recorded_accesses.back();
-        bool do_write_last = IsWrite(*last_access.usage_info);
-        if (do_write_last) --count;
+        // First access is only closed if the last is a write
+        bool do_write_last = recorded_use.first_access_closed_;
+        if (do_write_last) {
+            // Note: We know count > 0 so this is alway safe.
```
artem-lunarg, line 3248: Nice comment to have! We have other places in the codebase that subtract unsigned types, without additional remark this does not look alarming enough.

layers/state_tracker/descriptor_sets.h
jzulauf-lunarg: when `known_valid_view_` is `false` then `!known_valid_view_ && ComputeInvalid()` returns `ComputeInvalid`'s evaluation of `!image_view_state_ || image_view_state_->Invalid();` which is exactly what Invalid did.

The only time that Invalid differs from the original is when `known_valid_view_` is `true`

layers/state_tracker/descriptor_sets.h
jzulauf-lunarg: Yes. Checking performance implications

jzulauf-lunarg: No impact on perf AFAICT.

artem-lunarg: Further focusing the question about performance.This part introduces new API (invalidate-style) which adds some complexity. Does it provide measurable performance improvement to pay for its existence?

layers/state_tracker/descriptor_sets.h
artem-lunarg: I agree, I made a mistake in the boolean logic.

artem-lunarg (approved):

Issue #3450, opened 2021-10-18
Bindless + Sync Validation


Eichenherz:
Sync Validation incorrectly reports hazards for bound, but not used resources.

>>> VK_ERROR <<<
SYNC-HAZARD-READ_AFTER_WRITE
Validation Error: [ SYNC-HAZARD-READ_AFTER_WRITE ] Object 0: handle = 0xcc25c90000000082, type = VK_OBJECT_TYPE_IMAGE_VIEW; | MessageID = 0x4cb75b2d | vkCmdDispatch: Hazard READ_AFTER_WRITE for VkImageView 0xcc25c90000000082[], in VkCommandBuffer 0x2c31d47aef0[], and VkPipeline 0x59ffe0000000003d[Pipeline_Comp_DrawCull], VkDescriptorSet 0x980b0000000002e[], type: VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE, imageLayout: VK_IMAGE_LAYOUT_READ_ONLY_OPTIMAL_KHR, binding #2, index 0. Access info (usage: SYNC_COMPUTE_SHADER_SHADER_STORAGE_READ, prior_usage: SYNC_IMAGE_LAYOUT_TRANSITION, write_barriers: SYNC_EARLY_FRAGMENT_TESTS_DEPTH_STENCIL_ATTACHMENT_READ|SYNC_EARLY_FRAGMENT_TESTS_DEPTH_STENCIL_ATTACHMENT_WRITE|SYNC_LATE_FRAGMENT_TESTS_DEPTH_STENCIL_ATTACHMENT_READ|SYNC_LATE_FRAGMENT_TESTS_DEPTH_STENCIL_ATTACHMENT_WRITE, command: vkCmdPipelineBarrier2KHR, seq_no: 52, reset_no: 1).

>>> VK_ERROR <<<
SYNC-HAZARD-READ_AFTER_WRITE
[...]

PR #6605, opened 2023-10-03
layers: Workaround sync validation of descriptor arrays

artem-lunarg:
Addresses https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/3450

Currently, validation of memory accesses based on declared descriptors can produce false-positives. The shader can decide not to do such accesses, it can perform accesses with more narrow scope (e.g. read access, when both reads and writes are allowed) or for an array of descriptors, not all elements are accessed in the general case.

This workaround disables validation for the runtime descriptor array case (`OpCapability RuntimeDescriptorArray`).

### DIFF
--- a/layers/state_tracker/shader_module.cpp
+++ b/layers/state_tracker/shader_module.cpp
@@ -839,6 +839,10 @@ SPIRV_MODULE_STATE::StaticData::StaticData(const SPIRV_MODULE_STATE& module_stat
 
             case spv::OpCapability:
                 capability_list.push_back(static_cast<spv::Capability>(insn.Word(1)));
+                // Cache frequently checked capabilities
+                if (capability_list.back() == spv::CapabilityRuntimeDescriptorArray) {
+                    has_capability_runtime_descriptor_array = true;
+                }
                 break;
 
             case spv::OpVariable:
--- a/layers/sync/sync_validation.cpp
+++ b/layers/sync/sync_validation.cpp
@@ -1980,6 +1980,16 @@ bool CommandBufferAccessContext::ValidateDispatchDrawDescriptorSet(VkPipelineBin
             SyncStageAccessIndex sync_index =
                 GetSyncStageAccessIndexsByDescriptorSet(descriptor_type, variable, stage_state.GetStage());
 
+            // Currently, validation of memory accesses based on declared descriptors can produce false-positives.
+            // The shader can decide not to do such accesses, it can perform accesses with more narrow scope
+            // (e.g. read access, when both reads and writes are allowed) or for an array of descriptors, not all
+            // elements are accessed in the general case.
+            //
+            // This workaround disables validation for the runtime descriptor array case.
+            if (binding->count > 1 && stage_state.spirv_state->static_data_.has_capability_runtime_descriptor_array) {
+                continue;
+            }
+
             for (uint32_t index = 0; index < binding->count; index++) {
                 const auto *descriptor = binding->GetDescriptor(index);
                 switch (descriptor->GetClass()) {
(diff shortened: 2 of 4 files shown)


layers/sync/sync_validation.cpp
```diff
@@ -2141,6 +2141,16 @@ void CommandBufferAccessContext::RecordDispatchDrawDescriptorSet(VkPipelineBindP
             SyncStageAccessIndex sync_index =
                 GetSyncStageAccessIndexsByDescriptorSet(descriptor_type, variable, stage_state.GetStage());
 
+            // Currently, validation of memory accesses based on declared descriptors can produce false-positives.
```
jzulauf-lunarg, line 2144: This disables *all* descriptor access validation... needs to be more narrowly defined for the specific false positive cases affecting users.

jzulauf-lunarg (changes requested):
Too broad, effectively eliminates all syncval for descriptor based accesses -- and only stops validation, not record...  see the other instance of `for (const auto &variable : stage_state.entrypoint->resource_interface_variables)` ...

Any eventual fix will have to match in both `CommandBufferAccessContext::ValidateDispatchDrawDescriptorSet` and `CommandBufferAccessContext::RecordDispatchDrawDescriptorSet`

layers/sync/sync_validation.cpp
artem-lunarg: Isn't this a specific case? Any descriptor array can produce false-positives, and that's a typical case for modern graphical apps that increasingly use more descriptor indexing. My impression that keeping this makes syncval unusable for most big engines, and the proper solution should be gpu-av framework when it's ready.

artem-lunarg: > and only stops validation, not record

I was trying to skip in record function, and then validation automatically does not have this access. Is `RecordDispatchDrawDescriptorSet` a wrong place to change?

artem-lunarg: ah, I see. I saw both places but decided it's enough to skip update. I agree, validation should also do the same.

artem-lunarg: Fixed to have check both in Record and Validate.

> Too broad, effectively eliminates all syncval for descriptor based accesses 

It does not eliminate *all* validation. We still validate a single descriptor. We can not validate array because the code assumes that every access propagates to all elements but the shader is not required to access all array elements. 

That to some degree if true even for a single descriptor but that's part is kept for now: the code's assumption about arrays is more often than not is false in practice, but with a single descriptor it's often true the shader accesses declared single descriptor, so at least it's different category of probabilities.

martin-ejdestig: If disabling checking on all arrays is not wanted, could it perhaps be limited to bindings with `VK_DESCRIPTOR_BINDING_UPDATE_AFTER_BIND_BIT`and `VK_DESCRIPTOR_BINDING_PARTIALLY_BOUND_BIT` ?

(Not sure it is a good idea. I have only implemented bindless for textures in my engine at the moment, and do not really know what other engines do. But I assume most bindless "designs" have those flags set. I also really have no idea how the validation layer works here. Just a thought.)

artem-lunarg:
@jzulauf-lunarg Updated implementation to narrow scope to runtime  descriptor arrays (`OpCapability RuntimeDescriptorArray`).

jzulauf-lunarg (approved):
LGTM

layers/sync/sync_validation.cpp
```diff
@@ -1994,6 +1994,8 @@ bool CommandBufferAccessContext::ValidateDispatchDrawDescriptorSet(VkPipelineBin
         } else if (!stage_state.entrypoint) {
             continue;
         }
+        const bool has_runtime_descriptor_array_capability =
```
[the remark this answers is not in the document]
artem-lunarg: Updated code to cache runtime array capability.

layers/state_tracker/shader_module.cpp
```diff
@@ -839,6 +839,10 @@ SPIRV_MODULE_STATE::StaticData::StaticData(const SPIRV_MODULE_STATE& module_stat
 
             case spv::OpCapability:
                 capability_list.push_back(static_cast<spv::Capability>(insn.Word(1)));
+                // Cache frequently checked capabilities
+                if (capability_list.back() == spv::CapabilityRuntimeDescriptorArray) {
+                    has_capability_runtime_descriptor_array = true;
```
jzulauf-lunarg, line 844: 👍

jzulauf-lunarg (approved):
LGTM

jeremyg-lunarg (approved):

PR #6615, opened 2023-10-03
tests: Test tessellation point size

ziga-lunarg:
Part of #5842 

VUID-VkGraphicsPipelineCreateInfo-TessellationEvaluation-07723
VUID-VkGraphicsPipelineCreateInfo-TessellationEvaluation-07724


tests/unit/pipeline.cpp
```diff
@@ -3405,3 +3405,78 @@ TEST_F(NegativePipeline, MismatchedRasterizationSamples) {
     m_commandBuffer->EndRendering();
     m_commandBuffer->end();
 }
+
+TEST_F(NegativePipeline, PipelineTessellationMissingPointSize) {
```
spencer-lunarg, line 3409: would just trow these in `geometry_tessellation.cpp`

PR #6616, opened 2023-10-03
tests: Test invalid pipeline depth bias clamp

ziga-lunarg:
Part of #5842 

VUID-VkGraphicsPipelineCreateInfo-pDynamicStates-00754

### DIFF
--- a/tests/unit/pipeline.cpp
+++ b/tests/unit/pipeline.cpp
@@ -3326,6 +3326,23 @@ TEST_F(NegativePipeline, RasterStateWithDepthBiasRepresentationInfo) {
     m_errorMonitor->VerifyFound();
 }
 
+TEST_F(NegativePipeline, InvalidPipelineDepthBias) {
+    TEST_DESCRIPTION("Create pipeline with invalid depth bias");
+
+    ASSERT_NO_FATAL_FAILURE(InitFramework());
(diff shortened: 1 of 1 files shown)


tests/unit/pipeline.cpp
```diff
@@ -3326,6 +3326,23 @@ TEST_F(NegativePipeline, RasterStateWithDepthBiasRepresentationInfo) {
     m_errorMonitor->VerifyFound();
 }
 
+TEST_F(NegativePipeline, InvalidPipelineDepthBias) {
+    TEST_DESCRIPTION("Create pipeline with invalid depth bias");
+
+    ASSERT_NO_FATAL_FAILURE(InitFramework());
+    VkPhysicalDeviceFeatures features = {};
```
spencer-lunarg, line 3333: this is already done, should be able to just go 

```
ASSERT_NO_FATAL_FAILURE(Init());
```

spencer-lunarg (approved):

tests/unit/pipeline.cpp
ziga-lunarg: That does not work, because if you use `Init()` then `VkPhysicalDeviceFeatures` members are not all VK_FALSE, but are queries from the device.

Issue #6629, opened 2023-10-05
Check VkSurfacePresentModeEXT is chained for VkSurfacePresentModeCompatibilityEXT


emersion:
**Describe the situation in which you encountered the missing validation**
Issue description: https://github.com/KhronosGroup/Vulkan-Tools/issues/846 was not detected by the validation layers.

**Valid Usage IDs requested**
Please include the valid usage IDs for the checks you are requesting:
VUID-vkGetPhysicalDeviceSurfaceCapabilities2KHR-pNext-07776

**Additional Context**

PR #6630, opened 2023-10-05
layers: Fix vkGetPhysicalDeviceSurfaceCapabilities2KHR

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6629

### DIFF
--- a/layers/stateless/sl_wsi.cpp
+++ b/layers/stateless/sl_wsi.cpp
@@ -252,46 +252,36 @@ bool StatelessValidation::manual_PreCallValidateGetPhysicalDeviceSurfaceCapabili
     }
 #endif
 
-    if (IsExtEnabled(device_extensions.vk_ext_surface_maintenance1)) {
-        const auto *surface_present_mode_compatibility =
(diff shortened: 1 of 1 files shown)


ziga-lunarg (approved):

artem-lunarg (approved):

layers/stateless/sl_wsi.cpp
```diff
@@ -252,46 +252,36 @@ bool StatelessValidation::manual_PreCallValidateGetPhysicalDeviceSurfaceCapabili
     }
 #endif
 
-    if (IsExtEnabled(device_extensions.vk_ext_surface_maintenance1)) {
```
artem-lunarg, line 255: Is the use case here that it's possible to provide pNext structures without vk_ext_surface_maintenance1 enabled?

spencer-lunarg: we should have an automatic check that if you are using an pNext struct without the extension it will have an error, have tracked in another issue to confirm this (not just this issue, this is a long time generation code todo)

spencer-lunarg: clean up

PR #6632, opened 2023-10-05
Explicitly exclude shaders from codegen verification

Tony-LunarG:
Also allows verification of gpu_inst_shader_hash.h

### DIFF
--- a/scripts/generate_source.py
+++ b/scripts/generate_source.py
@@ -330,7 +330,8 @@ def repo_relative(path):
 
 def main(argv):
     # files to exclude from --verify check
-    verify_exclude = ['.clang-format']
+    # The shaders requires glslangvalidator, so they are updated manually with generate_spirv when needed
+    verify_exclude = ['.clang-format', 'gpu_as_inspection_comp.h', 'gpu_pre_dispatch_comp.h', 'gpu_pre_draw_vert.h', 'inst_functions_comp.h']
 
     parser = argparse.ArgumentParser(description='Generate source code for this repository')
     parser.add_argument('--api',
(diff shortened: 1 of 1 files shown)


scripts/generate_source.py
```diff
@@ -330,7 +330,8 @@ def repo_relative(path):
 
 def main(argv):
     # files to exclude from --verify check
-    verify_exclude = ['.clang-format']
+    # The shaders requires glslangvalidator, so they are updated manually with generate_spirv when needed
+    verify_exclude = ['.clang-format', 'gpu_as_inspection_comp.h', 'gpu_pre_dispatch_comp.h', 'gpu_pre_draw_vert.h', 'inst_functions_comp.h']
```
spencer-lunarg, line 334: not sure if it would be better to just have a `generated/gpu_shader` folder and then could just glob that

spencer-lunarg: @Tony-LunarG side question, why does `inst_functions_comp` not start with `gpu_`?

Tony-LunarG: @jeremyg-lunarg added that one.  Maybe that's what they were called in spirv-tools?

jeremyg-lunarg:
@spencer-lunarg inst_functions.comp has to be compiled with different flags to glslang than the gpu_* files, see generate_spirv.py.  and it is related to the instrumentation passes in SPIRV-Tools. Those are the reasons why, but both of them could be solved differently.

spencer-lunarg:
sounds good, nothing urgent, maybe a dedicated `generated/gpu_shader` folder would be helpful, but something to look into while cleaning up GPU-AV

PR #6640, opened 2023-10-06
tests: Rename functions checking format support

### LINKED ISSUE #5861: ImageFormatAndFeaturesSupported is overloaded for no good reason
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/f72e63f21d02ee7e1e81115bb91f940e90d4bb34/tests/framework/layer_validation_tests.cpp#L76

The overloads of ImageFormatAndFeaturesSupported will either call `vkvGetPhysicalDeviceFormatProperties` then `vkGetPhysicalDeviceImageFormatProperties`, or just call `vkGetPhysicalDeviceFormatProperties`, which is confusing.

This came up due to the `PositiveFragmentShadingRate.StageWithPipelineBarrier` test failing with `VUID-VkImageCreateInfo-imageCreateMaxMipLevels-02251` because `vkGetPhysicalDeviceFormatProperties` reported the format as supported while `vkGetPhysicalDeviceImageFormatProperties` does not.

ziga-lunarg:
Make it clear which functions are calling vkGetPhysicalDeviceFormatProperties and which vkGetPhysicalDeviceImageFormatProperties from the name

Closes #5861

### DIFF
--- a/tests/framework/layer_validation_tests.cpp
+++ b/tests/framework/layer_validation_tests.cpp
@@ -64,7 +64,7 @@ VkFormat FindSupportedDepthStencilFormat(VkPhysicalDevice phy) {
     return VK_FORMAT_UNDEFINED;
 }
 
-bool ImageFormatIsSupported(VkPhysicalDevice phy, VkFormat format, VkImageTiling tiling, VkFormatFeatureFlags features) {
+bool FormatIsSupported(VkPhysicalDevice phy, VkFormat format, VkImageTiling tiling, VkFormatFeatureFlags features) {
     VkFormatProperties format_props;
     vk::GetPhysicalDeviceFormatProperties(phy, format, &format_props);
     VkFormatFeatureFlags phy_features =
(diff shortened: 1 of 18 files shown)


tests/framework/layer_validation_tests.cpp
```diff
@@ -64,26 +64,26 @@ VkFormat FindSupportedDepthStencilFormat(VkPhysicalDevice phy) {
...
     VkFormatProperties format_props;
     vk::GetPhysicalDeviceFormatProperties(phy, format, &format_props);
     VkFormatFeatureFlags phy_features =
         (VK_IMAGE_TILING_OPTIMAL == tiling ? format_props.optimalTilingFeatures : format_props.linearTilingFeatures);
     return (features == (phy_features & features));
 }
 
-bool ImageFormatAndFeaturesSupported(const VkInstance inst, const VkPhysicalDevice phy, const VkImageCreateInfo info,
-                                     const VkFormatFeatureFlags features) {
+bool ImageFormatIsSupported(const VkInstance inst, const VkPhysicalDevice phy, const VkImageCreateInfo info,
```
spencer-lunarg, line 83: (just want to say I hated how confusing these were and thanks for taking this issue)

I feel having a 3rd function for `ImageFormatIsSupported` is not useful, the caller can just pass in the format/tiling

If you want to keep this, then lets just make it `FormatFeaturesAreSupported` and have it be an overload, not an another named function

ziga-lunarg: This third `ImageFormatIsSupported()` function calls `vkGetPhysicalDeviceImageFormatProperties` and checks if it returns `VK_SUCCESS`, while the other 2 only check format features, so if anything I would get rid of `FormatFeaturesAreSupported()`

spencer-lunarg: fair, could you try getting rid of `FormatFeaturesAreSupported`,  I can't really see where we actually need it over `FormatIsSupported` and it is only there because it has been copy-and-pasted for years

ziga-lunarg: `FormatFeaturesAreSupported` checks for specific format features so I believe it is needed. And I can see `FormatIsSupported` is used to check for overall format support in cases where we add format features with device profile.

spencer-lunarg (approved):

PR #6641, opened 2023-10-06
layers: Improve ValidateMemoryTypes error message

spencer-lunarg:
before

> vkBindImageMemory2(): pBindInfos[0] MemoryRequirements->memoryTypeBits (0x8) for this object type are not compatible with the memory type (2) of VkDeviceMemory 0xf56c9b0000000004[].

and now

> vkBindImageMemory2(): pBindInfos[0].image require memoryTypeBits (0x8) but VkDeviceMemory 0xf56c9b0000000004[] was allocated with memoryTypeIndex (2).

### DIFF
--- a/layers/core_checks/cc_device_memory.cpp
+++ b/layers/core_checks/cc_device_memory.cpp
@@ -429,15 +429,13 @@ bool CoreChecks::ValidateInsertBufferMemoryRange(VkBuffer buffer, const DEVICE_M
     return ValidateInsertMemoryRange(VulkanTypedHandle(buffer, kVulkanObjectTypeBuffer), mem_info, mem_offset, loc);
 }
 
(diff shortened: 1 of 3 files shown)


artem-lunarg (approved):
LGTM. Simpler message with the same information.

One more idea for the future. When we print both memory type index and flags in the same message user has to do some mental math to compare flags to index. In addition to flags hex value we can also print memory type indices encoded in the flags. If we have flags 0x6  = 2 + 4 -> we also can print that supported indices are (1, 2).  
Sometimes that's not obvious just from looking into numbers. In the description the flags is 0x8 which denotes index = 3

spencer-lunarg:
agree, would be nice to help represent the values better... probably a lot around memory type that need the same improvement.. added to my todo list

PR #6647, opened 2023-10-06
tests: Validate warnings in best practice tests

ziga-lunarg:


tests/unit/best_practices_positive.cpp
```diff
@@ -25,6 +25,8 @@ TEST_F(VkPositiveBestPracticesLayerTest, TestDestroyFreeNullHandles) {
     ASSERT_NO_FATAL_FAILURE(InitState());
     ASSERT_NO_FATAL_FAILURE(InitRenderTarget());
 
+    m_errorMonitor->ExpectSuccess(kErrorBit | kWarningBit);
```
spencer-lunarg, line 28: what we really should do is make this the default for all BestPractices tests

also I think `kPerformanceWarningBit` should be added to this

ziga-lunarg: The problem I ran into trying to make it default is that InitRenderTarget() already produces warnings. That might need a bit more work

spencer-lunarg: aww, for now I think this is an improvement, so happy to apply this PR, but long term would like a more default solution

spencer-lunarg (approved):

tests/unit/best_practices_positive.cpp
ziga-lunarg: I agree

ziga-lunarg: `kPerformanceWarningBit` also seems tricky, for example when we create a `vkt::Buffer`, we only allocate memory that we need, but then we get a small allocation performance warning

PR #6658, opened 2023-10-08
tests: Add support to Mock AHB 

spencer-lunarg:
Problem: No way to test `VK_ANDROID_external_format_resolve` (or `VK_ANDROID_external_memory_android_hardware_buffer`) with out an Android device that supports it (even then it is not fun). This means no coverage is done in Github CI and development is hard

Solution: Add a "Mock AHB" so we can simulate the Android Hardware Buffer to allow the tests to run on local desktop devices

This change

- Adds a bunch headers to allow things to compile
- CI job to test `--gtest_filter=*AndroidHardwareBuffer*` tests
- CMake support for this with a `-DVVL_MOCK_ANDROID=ON` flag to toggle it on

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -237,7 +237,14 @@ else()
     message(FATAL_ERROR "Cannot determine SPIRV-Tools target name")
 endif()
 
-if (ANDROID)
+# More details in tests/android/mock/README.md
+option(VVL_MOCK_ANDROID "Enable building for Android on desktop for testing with MockICD setup")
+if(VVL_MOCK_ANDROID)
+    # We don't want to build the android APK, so we just set the minimum Android settings
+    target_sources(VkLayer_utils PRIVATE android/mock/android/hardware_buffer.cpp)
+    target_compile_definitions(VkLayer_utils PUBLIC VK_USE_PLATFORM_ANDROID_KHR VVL_MOCK_ANDROID __ANDROID__)
+    target_include_directories(VkLayer_utils SYSTEM PUBLIC android/mock)
+elseif (ANDROID)
     add_subdirectory(android)
     return()
 endif()
--- a/tests/device_profiles/max_profile.json
+++ b/tests/device_profiles/max_profile.json
@@ -1996,6 +1996,29 @@
                 "VK_EXT_dynamic_rendering_unused_attachments": 1
             },
             "formats": {
+                "VK_FORMAT_UNDEFINED": {
+                    "VkFormatProperties3": {
+                        "linearTilingFeatures": [
+                            "VK_FORMAT_FEATURE_2_SAMPLED_IMAGE_BIT",
+                            "VK_FORMAT_FEATURE_2_STORAGE_IMAGE_BIT",
+                            "VK_FORMAT_FEATURE_2_UNIFORM_TEXEL_BUFFER_BIT",
+                            "VK_FORMAT_FEATURE_2_STORAGE_TEXEL_BUFFER_BIT",
+                            "VK_FORMAT_FEATURE_2_COLOR_ATTACHMENT_BIT",
+                            "VK_FORMAT_FEATURE_2_TRANSFER_SRC_BIT",
+                            "VK_FORMAT_FEATURE_2_TRANSFER_DST_BIT"
+                        ],
+                        "optimalTilingFeatures": [
+                            "VK_FORMAT_FEATURE_2_SAMPLED_IMAGE_BIT",
+                            "VK_FORMAT_FEATURE_2_STORAGE_IMAGE_BIT",
+                            "VK_FORMAT_FEATURE_2_UNIFORM_TEXEL_BUFFER_BIT",
+                            "VK_FORMAT_FEATURE_2_STORAGE_TEXEL_BUFFER_BIT",
+                            "VK_FORMAT_FEATURE_2_COLOR_ATTACHMENT_BIT",
+                            "VK_FORMAT_FEATURE_2_TRANSFER_SRC_BIT",
+                            "VK_FORMAT_FEATURE_2_TRANSFER_DST_BIT"
+                        ],
+                        "bufferFeatures": []
+                    }
+                },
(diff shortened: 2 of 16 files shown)


layers/CMakeLists.txt
```diff
@@ -352,6 +356,10 @@ if (ANDROID)
     return()
 endif()
 
+if(VVL_MOCK_ANDROID)
+    target_include_directories(VkLayer_utils SYSTEM PUBLIC ../tests/android/mock)
```
spencer-lunarg, line 360: @juan-lunarg is there a better way to do this then `../`?

tests/device_profiles/max_profile.json
```diff
@@ -1996,6 +1996,29 @@
                 "VK_EXT_dynamic_rendering_unused_attachments": 1
             },
             "formats": {
+                "VK_FORMAT_UNDEFINED": {
```
spencer-lunarg, line 1999: when using external formats, you pass in `VK_FORMAT_UNDEFINED`

tests/framework/layer_validation_tests.cpp
```diff
@@ -1257,7 +1257,7 @@ void print_android(const char *c) {
 #endif  // VK_USE_PLATFORM_ANDROID_KHR
 }
 
-#if defined(VK_USE_PLATFORM_ANDROID_KHR)
+#if defined(VK_USE_PLATFORM_ANDROID_KHR) && !defined(VVL_MOCK_ANDROID)
```
spencer-lunarg, line 1260: we don't need to build the APK `main` for tests as we are using the "standard desktop" gtest still

tests/android/CMakeLists.txt
juan-lunarg, line 1: Example how you could move all the `VVL_MOCK_ANDROID` CMake logic in here. Which would be much better IMO.

```
# More details in tests/android/mock
option(VVL_MOCK_ANDROID "Enable building for Android on desktop for testing with MockICD setup")

if (VVL_MOCK_ANDROID)
    target_sources(VkLayer_utils PRIVATE mock/android/hardware_buffer.cpp)
    target_compile_definitions(VkLayer_utils PUBLIC VK_USE_PLATFORM_ANDROID_KHR VVL_MOCK_ANDROID __ANDROID__)
    target_include_directories(VkLayer_utils SYSTEM PUBLIC ../tests/android/mock)
endif()
```

juan-lunarg (changes requested):
Mainly just feedback on how to cleanup the CMake

tests/android/CMakeLists.txt
juan-lunarg: NOTE: The use of PUBLIC with `VkLayer_utils`. Since everything links against `VkLayer_utils` everything will be propagated properly.

tests/unit/android_hardware_buffer.cpp
```diff
@@ -246,6 +246,7 @@ TEST_F(NegativeAndroidHardwareBuffer, AllocationSize) {
     {
         memory_allocate_info.allocationSize = ahb_props.allocationSize;
         memory_allocate_info.memoryTypeIndex++;
+        m_errorMonitor->SetUnexpectedError("VUID-vkAllocateMemory-pAllocateInfo-01714");  // incase at last index
```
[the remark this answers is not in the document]
juan-lunarg: gotcha

tests/android/CMakeLists.txt
spencer-lunarg: yep, this works much cleaner having the `VVL_MOCK_ANDROID` all together and not scatter all around

layers/CMakeLists.txt
```diff
@@ -352,6 +352,14 @@ if (ANDROID)
     return()
 endif()
 
+# More details in tests/android/mock
+option(VVL_MOCK_ANDROID "Enable building for Android on desktop for testing with MockICD setup")
+if(VVL_MOCK_ANDROID)
+    target_sources(VkLayer_utils PRIVATE ../tests/android/mock/android/hardware_buffer.cpp)
+    target_compile_definitions(VkLayer_utils PUBLIC VK_USE_PLATFORM_ANDROID_KHR VVL_MOCK_ANDROID __ANDROID__)
+    target_include_directories(VkLayer_utils SYSTEM PUBLIC ../tests/android/mock)
+endif()
```
juan-lunarg, line 361: This code can be moved to `tests/android/CMakeLists.txt` instead.

Which IMO is better since this code is only relevant for testing.

spencer-lunarg: moved to `tests/CMakelist.txt` (added comment why we don't plan to build the APK so it shouldn't be in the `tests/android/CMakeLists.txt`

tests/android/CMakeLists.txt
```diff
@@ -27,7 +27,7 @@
 # https://github.com/android/ndk/issues/360#issuecomment-339107521
 # https://github.com/android/ndk/releases/tag/r26-beta1
 if ("${CMAKE_ANDROID_STL_TYPE}" MATCHES "static")
-    # nodelete : Marks that vvl shouldn't be unloaded at runtime. 
+    # nodelete : Marks that vvl shouldn't be unloaded at runtime.
```
juan-lunarg, line 30: Revert change here.

tests/CMakeLists.txt
```diff
@@ -237,11 +237,19 @@ else()
...
+    # We don't want to build the android APK, so we just set the minimum Android settings
+    target_sources(VkLayer_utils PRIVATE android/mock/android/hardware_buffer.cpp)
+    target_compile_definitions(VkLayer_utils PUBLIC VK_USE_PLATFORM_ANDROID_KHR VVL_MOCK_ANDROID __ANDROID__)
+    target_include_directories(VkLayer_utils SYSTEM PUBLIC android/mock)
+elseif (ANDROID)
     add_subdirectory(android)
     return()
 endif()
 
+
```
juan-lunarg, line 252: Revert change here.

juan-lunarg (approved):

arno-lunarg (approved):

PR #6659, opened 2023-10-08
tests: Cleanup CommandBuffer binding class

spencer-lunarg:
Replace `VkCommandBufferObj` with `vkt::CommandBuffer`

Removed the helper functions that really didn't add much value (were just passing in the same arguments and provided no defaults)


arno-lunarg (approved):
LGTM!

PR #6660, opened 2023-10-08
Update VUID 02756, validate required subgroup size in task and mesh shaders

### LINKED ISSUE #6655: Fix 02756 to account for Mesh Shading
` VUID-VkPipelineShaderStageCreateInfo-pNext-02756` [was updated](https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5977/diffs) in 1.3.257 to add "mesh and task" but currently only being validated with Compute (needs a test as well)

ziga-lunarg:
Closes #6655

### DIFF
--- a/layers/core_checks/cc_spirv.cpp
+++ b/layers/core_checks/cc_spirv.cpp
@@ -2071,7 +2071,8 @@ bool CoreChecks::ValidatePipelineShaderStage(const StageCreateInfo &stage_create
     skip |= ValidateShaderSubgroupSizeControl(stage_create_info, stage, stage_state, loc);
     skip |= ValidateSpecializations(stage_state.GetSpecializationInfo(), stage_create_info, loc.dot(Field::pSpecializationInfo));
     skip |= ValidateShaderStageMaxResources(stage, stage_create_info, loc);
-    if (const auto *pipeline_robustness_info = vku::FindStructInPNextChain<VkPipelineRobustnessCreateInfoEXT>(stage_state.GetPNext());
+    if (const auto *pipeline_robustness_info =
+            vku::FindStructInPNextChain<VkPipelineRobustnessCreateInfoEXT>(stage_state.GetPNext());
         pipeline_robustness_info) {
         skip |= ValidatePipelineRobustnessCreateInfo(*stage_create_info.pipeline, *pipeline_robustness_info, loc);
     }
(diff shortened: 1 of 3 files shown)


layers/core_checks/cc_spirv.cpp
```diff
@@ -2289,6 +2289,13 @@ bool CoreChecks::ValidatePipelineShaderStage(const StageCreateInfo &stage_create
         }
     }
 
+    if (stage == VK_SHADER_STAGE_COMPUTE_BIT || stage == VK_SHADER_STAGE_TASK_BIT_EXT || stage == VK_SHADER_STAGE_MESH_BIT_EXT) {
```
spencer-lunarg, line 2292: I think it makes more sense to put `ValidateRequiredSubgroupSize` and `GetInvocations` both in here... nice to have an area that is 

- "compute only"
- "mesh task only"
- all three

 (the `VUID-RuntimeSpirv-x-06432` check can just check if stage is `COMPUTE_BIT`

layers/core_checks/cc_spirv.cpp
```diff
@@ -2466,15 +2489,91 @@ bool CoreChecks::PreCallValidateGetShaderModuleCreateInfoIdentifierEXT(VkDevice
     return skip;
 }
 
+bool CoreChecks::GetInvocations(uint32_t local_size_x, uint32_t local_size_y, uint32_t local_size_z, uint64_t *invocations) const {
+    bool fail = false;
+
+    uint32_t limit = phys_dev_props.limits.maxComputeWorkGroupInvocations;
+    uint64_t total = static_cast<uint64_t>(local_size_x) * static_cast<uint64_t>(local_size_y);
+    // Prevent overflow.
+    if (total > vvl::kU32Max || total > limit) {
```
artem-lunarg, line 2498: In all such places, kU32Max >= limit, therefore `(total > kU32max || total > limit)` is equivalent to `(total > limit)`. Not sure if the check with kU32Max is something to emphasize the spec/vuid wording, but from correctness point of view it's not needed.

ziga-lunarg: This was already in before, and it was where I copied it from the last time. I'll remove it

artem-lunarg: Ah, sorry, did not pay attention it's existence code.

layers/core_checks/cc_spirv.cpp
```diff
@@ -2466,15 +2489,84 @@ bool CoreChecks::PreCallValidateGetShaderModuleCreateInfoIdentifierEXT(VkDevice
     return skip;
 }
 
+bool CoreChecks::GetInvocations(uint32_t local_size_x, uint32_t local_size_y, uint32_t local_size_z, uint64_t *invocations) const {
```
spencer-lunarg, line 2492: I feel this function now would just make more sense inlined where it used (then don't need to pass in all the local_size and understand why a `Get*()` function returns a bool

spencer-lunarg (approved):
LGTM with change to GetInvocations

Issue #4696, opened 2022-10-22
Depth/stencil attachments are not transitioned correctly in the render pass


ziga-lunarg:
If a framebuffer has a depth/stencil attachment, that has only one of `VK_IMAGE_ASPECT_DEPTH_BIT` and `VK_IMAGE_ASPECT_STENCIL_BIT`, the validation layers will not correctly track the layout of the other. 

Here is a test to reproduce the issue, and changing `depthImage.targetView(format, VK_IMAGE_ASPECT_DEPTH_BIT);` to `depthImage.targetView(format, VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT);` makes the test pass.

```
TEST_F(VkPositiveLayerTest, TestDepthStencilRenderPassTransition) {
    TEST_DESCRIPTION("Create framebuffer with a depth/stencil attachment that has only depth aspect and transition it in render pass.");
    SetTargetApiVersion(VK_API_VERSION_1_1);
    ASSERT_NO_FATAL_FAILURE(InitFramework());
    if (DeviceValidationVersion() < VK_API_VERSION_1_1) {
        GTEST_SKIP() << "At least Vulkan version 1.1 is required";
    }
    ASSERT_NO_FATAL_FAILURE(InitState());

    VkFormat format = VK_FORMAT_D24_UNORM_S8_UINT;
    VkImageObj depthImage(m_device);
    depthImage.Init(32, 32, 1, format, VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT | VK_IMAGE_USAGE_TRANSFER_SRC_BIT, VK_IMAGE_TILING_OPTIMAL, 0);

    VkImageView depthView = depthImage.targetView(format, VK_IMAGE_ASPECT_DEPTH_BIT);

    VkAttachmentReference depthAttachment = {};
    depthAttachment.attachment = 0;
    depthAttachment.layout = VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL;
    VkSubpassDescription subpass = {};
    subpass.pDepthStencilAttachment = &depthAttachment;
    VkAttachmentDescription depth_attach_desc = {};
    depth_attach_desc.format = format;
    depth_attach_desc.samples = VK_SAMPLE_COUNT_1_BIT;
[...]

PR #6661, opened 2023-10-09
layers: Fix image layout transition for depth stencil image when only one aspect is supplied

arno-lunarg:
- Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4696

Fix image layout transition for depth stencil image when only one aspect is supplied

### DIFF
--- a/layers/core_checks/cc_image_layout.cpp
+++ b/layers/core_checks/cc_image_layout.cpp
@@ -852,7 +852,16 @@ void CoreChecks::TransitionBeginRenderPassLayouts(CMD_BUFFER_STATE *cb_state, co
                 sub_range.aspectMask = VK_IMAGE_ASPECT_STENCIL_BIT;
                 cb_state->SetImageInitialLayout(*image_state, sub_range, stencil_initial_layout);
             } else {
-                cb_state->SetImageInitialLayout(*image_state, view_state->normalized_subresource_range, initial_layout);
+                // If layoutStencil is kInvalidLayout (meaning no separate depth/stencil layout), image view format has both depth
+                // and stencil aspects, and subresource has only one of aspect out of depth or stencil, then the missing aspect will
+                // also be transitioned and thus must be included explicitly
+                auto subresource_range = view_state->normalized_subresource_range;
+                if (const VkFormat format = view_state->create_info.format; vkuFormatIsDepthAndStencil(format)) {
+                    if (subresource_range.aspectMask & (VK_IMAGE_ASPECT_STENCIL_BIT | VK_IMAGE_ASPECT_DEPTH_BIT)) {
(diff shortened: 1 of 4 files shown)


layers/utils/vk_layer_utils.h
```diff
@@ -330,6 +345,12 @@ static inline bool IsImageLayoutStencilReadOnly(VkImageLayout layout) {
                        [layout](const VkImageLayout read_only_layout) { return layout == read_only_layout; });
 }
 
+static inline bool IsFormatDepthOnly(VkFormat format) {
```
spencer-lunarg, line 348: we have this in VUL already `vkuFormatIsDepthOnly`

arno-lunarg: Ah I was pretty sure we had it already, but just could not find. My 0.02$ is that `IsFormatDepthOnly` is a better name though, I prefer "question" names when it comes to return a bool

layers/core_checks/cc_image_layout.cpp
```diff
@@ -853,7 +853,16 @@ void CoreChecks::TransitionBeginRenderPassLayouts(CMD_BUFFER_STATE *cb_state, co
                 sub_range.aspectMask = VK_IMAGE_ASPECT_STENCIL_BIT;
                 cb_state->SetImageInitialLayout(*image_state, sub_range, stencil_initial_layout);
             } else {
-                cb_state->SetImageInitialLayout(*image_state, view_state->normalized_subresource_range, initial_layout);
+                // If layoutStencil is kInvalidLayout (meaning no separate depth/stencil layout), image view format has both depth
+                // and stencil aspects, and subresource has only one of aspect out of depth or stencil, then the missing aspect will
+                // also be transitioned and thus must be included explicitly
+                auto subresource_range = view_state->normalized_subresource_range;
+                if (const VkFormat format = view_state->create_info.format; vkuFormatIsDepthAndStencil(format)) {
```
artem-lunarg, line 860: Not sure if format should be tested (here and other function) for depth/stencil because `ValidateImageAspectMask` does validation during image view creation that format matches aspect mask. It looks like it should be fine to have only `aspectMask` logic.

artem-lunarg:
Did not work closely with this code but it looks like a good fix.

layers/core_checks/cc_image_layout.cpp
arno-lunarg: Some tests were failing without this check. Cannot find which one anymore, the failing build logs have been erased...

artem-lunarg: > This vkuFormatIsDepthAndStencil check was added so that images that were created with a format holding only one aspect do not get their aspectMask modified by this new logic.

I agree, missed this part

artem-lunarg (approved):
LGTM from my current understanding of this code.

PR #6668, opened 2023-10-10
Validate task and mesh shader memory limits

### LINKED ISSUE #6656: Missing Runtime Mesh Shading limits
- [ ] VUID-RuntimeSpirv-maxMeshPayloadAndSharedMemorySize-08755
- [ ] VUID-RuntimeSpirv-maxMeshOutputMemorySize-08756
- [ ] VUID-RuntimeSpirv-maxMeshPayloadAndOutputMemorySize-08757
- [ ] VUID-RuntimeSpirv-maxTaskPayloadSize-08758
- [ ] VUID-RuntimeSpirv-maxTaskPayloadAndSharedMemorySize-08760

ziga-lunarg:
Closes #6656

### DIFF
--- a/layers/state_tracker/shader_module.cpp
+++ b/layers/state_tracker/shader_module.cpp
@@ -64,10 +64,14 @@ void DecorationBase::Add(uint32_t decoration, uint32_t value) {
         case spv::DecorationPerPrimitiveEXT:  // VK_EXT_mesh_shader
             flags |= per_primitive_ext;
             break;
+        case spv::DecorationPerViewNV:
+            flags |= per_view_bit;
+            break;
         case spv::DecorationOffset:
             offset |= value;
             break;
         default:
+            location = location;
             break;
     }
 }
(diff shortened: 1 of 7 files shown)


tests/unit/shader_mesh.cpp
```diff
@@ -241,8 +241,8 @@ TEST_F(NegativeShaderMesh, SharedMemorySpecConstantSet) {
     CreatePipelineHelper::OneshotTest(*this, set_info, kErrorBit, "VUID-RuntimeSpirv-maxMeshSharedMemorySize-08754");
 }
 
-TEST_F(NegativeShaderMesh, TaskSharedMemoryOverLimit) {
-    TEST_DESCRIPTION("Validate Task shader shared memory does not exceed maxTaskSharedMemorySize");
+TEST_F(NegativeShaderMesh, TaskMemoryLimits) {
```
spencer-lunarg, line 244: Can we make a `InitBasicMeshShader` helper (like `InitBasicDynamicRendering`) and then break these into 1 shader per test... especially with shader stuff, I found having one per test really improves debugging time

tests/unit/shader_mesh.cpp
```diff
@@ -241,8 +241,8 @@ TEST_F(NegativeShaderMesh, SharedMemorySpecConstantSet) {
...
+    TEST_DESCRIPTION("Validate Task shader memory limits");
```
spencer-lunarg, line 245: It would good to have at least one Positive test using Spec Constants to set the limits (where the spec constant default is over the limit, but we pass in values that are lower)

This way we don't get an issue in the future of a false positive from not accounting the spec constant

layers/state_tracker/shader_module.cpp
```diff
@@ -1160,6 +1164,92 @@ uint32_t SPIRV_MODULE_STATE::CalculateWorkgroupSharedMemory() const {
...
+            total_size += variable_shared_size;
+        }
+    }
+    return total_size;
+}
+
+uint32_t SPIRV_MODULE_STATE::CalculateMeshOutputMemory(const VkPhysicalDeviceMeshShaderPropertiesEXT& mesh_shader_properties,
+                                                       const EntryPoint& entrypoint,
+                                                       std::vector<const Instruction*> builtin_decoration_inst) const { 
+    uint32_t n_v = 0u;    // number of effective scalar per vertex attributes not dependent on ViewIndex
```
spencer-lunarg, line 1187: between the inconsistent use of single letter variables and the formatting of the comments (would put above in this case) this function is very hard to understand and not going to be friendly at all to the future person who as to debug it. Can names like

```cpp
uint32_t num_vertex
uint32_t num_vertex_per_view
uint32_t views
```

the `v` is REALLY hard to know if it means `vertex` or `view` when looking at the logic

layers/state_tracker/shader_module.cpp
```diff
@@ -64,10 +64,14 @@ void DecorationBase::Add(uint32_t decoration, uint32_t value) {
         case spv::DecorationPerPrimitiveEXT:  // VK_EXT_mesh_shader
             flags |= per_primitive_ext;
             break;
+        case spv::DecorationPerViewNV:
```
spencer-lunarg, line 67: there are only found in the `SPV_NV_mesh_shader` extension, not the EXT version

Would it be better to check the `Entrypoint::execution_model` if `ExecutionModelMeshNV` or `ExecutionModelMeshEXT` and then have two different versions of `
CalculateMeshOutputMemory` t(or at least a `bool is_ext` argument) that are called as it seems they are slightly different

layers/state_tracker/shader_module.cpp
ziga-lunarg: This was copied from the equation in the spec, but yes I'll improve the names

layers/state_tracker/shader_module.cpp
ziga-lunarg: Shouldn't there be a an enum for the EXT as well? Or do ext mesh shaders not have per view attributes?

spencer-lunarg: so while the vulkan enum for shader stage is the same, the SPIR-V version of MeshNV and MeshEXT are slightly different (which is why there is no enum for EXT in spirv-headers)

Rua: Related: https://github.com/KhronosGroup/Vulkan-Docs/issues/2287

layers/state_tracker/shader_module.cpp
```diff
@@ -1160,6 +1164,100 @@ uint32_t SPIRV_MODULE_STATE::CalculateWorkgroupSharedMemory() const {
...
+
+    for (const Instruction* decoration_inst : builtin_decoration_inst) {
+        if (decoration_inst->GetBuiltIn() == spv::BuiltInMeshViewCountNV) {
+            views = decoration_inst->Word(1);
+        }
+    }
+
+    for (const auto& variable : entrypoint.stage_interface_variables) {
+        for (const auto& slot : variable.interface_slots) {
+            uint32_t size = slot.bit_width / 8u;
```
Rua, line 1212: The spec talks about the number of attributes (components?) but doesn't say anything about their size. So I think this should just be 1. Maybe 2 if the size is 64 bits, but that depends on whether `interface_slots` has one or two entries per 64-bit value.

PR #6676, opened 2023-10-11
test: Replace usage of ASSERT_VK_SUCCESS

charles-lunarg:
Replaced with `ASSERT_EQ(VK_SUCCESS, ...);` and added the necessary logic to allow gtest to pretty-print VkResult values.

This allows every place where VkResult is compared to inside of a gtest macro to get pretty-printing, instead of just ASSERT_VK_SUCCESS.

### DIFF
--- a/tests/unit/debug_printf.cpp
+++ b/tests/unit/debug_printf.cpp
@@ -496,7 +496,7 @@ TEST_F(NegativeDebugPrintf, GPL) {
     CreatePipelineHelper vi(*this);
     vi.InitVertexInputLibInfo();
     vi.InitState();
-    ASSERT_VK_SUCCESS(vi.CreateGraphicsPipeline(false));
+    ASSERT_EQ(VK_SUCCESS, vi.CreateGraphicsPipeline(false));
 
     const auto vs_spv = GLSLToSPV(VK_SHADER_STAGE_VERTEX_BIT, shader_source);
     vkt::GraphicsPipelineLibraryStage pre_raster_stage(vs_spv, VK_SHADER_STAGE_VERTEX_BIT);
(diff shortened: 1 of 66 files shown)


jeremyg-lunarg (approved):
LGTM

spencer-lunarg (changes requested):
So I have talked a lot about this with @artem-lunarg 

I honestly think most of these `ASSERT` are useless and should just be removed

We use to have these mainly because we didn't default to each test needing to not have any errors, so there were cases where we had a invalid `VkCreateImage` that trigger another validation, but was not seen

if it doesn't return VK_SUCCESS but has **not** Validation error that notifies you means

1. Your driver has a bug
2. We are missing some validation case (which I feel almost never happens now with testing since our coverage is higher).... also MockICD always returns success 
3. The call has another "successful" return which is when I would justify this

so instead of having a random subset of `vkCreateImage` check, I argue it should just be removed unless one ACTUALLY cares about the error code, which 99% of these don't

tests/unit/debug_printf.cpp
```diff
@@ -496,7 +496,7 @@ TEST_F(NegativeDebugPrintf, GPL) {
     CreatePipelineHelper vi(*this);
     vi.InitVertexInputLibInfo();
     vi.InitState();
-    ASSERT_VK_SUCCESS(vi.CreateGraphicsPipeline(false));
```
spencer-lunarg, line 499: this could just be asserted inside `CreateGraphicsPipeline`

tests/unit/dynamic_rendering.cpp
```diff
@@ -739,7 +739,7 @@ TEST_F(NegativeDynamicRendering, ClearAttachments) {
                 cmd_buffer_inheritance_info.pNext = &inheritance_rendering_info;
             } else {
                 const VkResult err = vk::CreateFramebuffer(m_device->handle(), &framebuffer_ci, nullptr, &framebuffers[3]);
-                ASSERT_VK_SUCCESS(err);
```
spencer-lunarg, line 742: cases like this are really just places we should replace with our framework helpers

for example, could declare the object with`vkt::Framebuffer` and then go `init` here and make sure the `VK_SUCCESS` is working inside the binding

spencer-lunarg (approved):
my issues are orthogonal, this removes `ASSERT_VK_SUCCESS` which is a valid step forward

PR #6677, opened 2023-10-11
tests: Throw exception upon ASSERT_* failure

charles-lunarg:
This change adds a 'listener' to the googletest framework so that whenever a fatal assertion is made, an exception is thrown forcing the test to stop executing at the point the assertion was made. This is important to ensure that any failed assertions in nested function calls are caught.

Without the change, assertions deep within the test framework's init code will not cause execution to halt, potentially causing hard crashes which reduce debuggability.

The googletest framework sets up exception handlers so that the exception thrown by the failed assertion doesn't escape up the callstack, causing the process to abort.


jeremyg-lunarg:
Before merging this, I think it would be good to do a test run with exceptions reenabled in VulkanTests

charles-lunarg:
> Before merging this, I think it would be good to do a test run with exceptions reenabled in VulkanTests

Yes, that is a good idea. I will kick off a custom build then.

spencer-lunarg (approved):
looks good from what I can see, but as @jeremyg-lunarg said, nice to see what CI shows with exceptions on

Issue #6508, opened 2023-09-18
Synchronization - false positive on image layout transition with split barriers


vmilea:
**Describe the Issue**

The validation layer reports WAW hazard during an image layout transition, even if memory writes were made available prior to the transition through a dependency chain. This can happen for initial layout transitions during a render pass, if it assumes memory availability through some previous synchronization mechanism.

**Expected behavior**

No false positives.

**Additional context**

I'll start with a toy example involving buffers only:

```
// [1] copy from stagingBuffer to deviceBuffer
vkCmdCopyBuffer(commandBuffer, stagingBuffer, deviceBuffer, 1, &region);

VkBufferMemoryBarrier bufferMemoryBarrierA = {
    .srcAccessMask = VK_ACCESS_TRANSFER_WRITE_BIT,
    .dstAccessMask = VK_ACCESS_NONE,
    .buffer = deviceBuffer,
    ...
};
// [2a] make writes available
vkCmdPipelineBarrier(commandBuffer,
    VK_PIPELINE_STAGE_TRANSFER_BIT,
    VK_PIPELINE_STAGE_TRANSFER_BIT,
    0,
    0, nullptr,
    1, &bufferMemoryBarrierA,
    0, nullptr);

VkBufferMemoryBarrier bufferMemoryBarrierB = {
    .srcAccessMask = VK_ACCESS_NONE, // already available
    .dstAccessMask = VK_ACCESS_TRANSFER_WRITE_BIT,
    .buffer = deviceBuffer,
    ...
};
// [2b] make writes visible
vkCmdPipelineBarrier(commandBuffer,
    VK_PIPELINE_STAGE_TRANSFER_BIT,
    VK_PIPELINE_STAGE_TRANSFER_BIT,
    0,
    0, nullptr,
    1, &bufferMemoryBarrierB,
    0, nullptr);

// [3] copy again from stagingBuffer to deviceBuffer
vkCmdCopyBuffer(commandBuffer, stagingBuffer, deviceBuffer, 1, &region);
```

[...]

PR #6684, opened 2023-10-12
layers: Ensure available accesses are not hazardous during ILT

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6508

`PositiveSyncVal.LayoutTransitionWithAlreadyAvailableImage` recreates scenario described in the issue. When synchronize with *available* write it should not cause WAW hazard during ILT which makes this write visible.

Other tests are added to illustrate additional scenarios but they worked before.

### DIFF
--- a/layers/sync/sync_validation.cpp
+++ b/layers/sync/sync_validation.cpp
@@ -8957,12 +8957,6 @@ bool ResourceAccessWriteState::IsOrdered(const OrderingBarrier &ordering, QueueI
     return (queue_ == queue_id) && ordering.access_scope[access_->stage_access_index];
 }
 
-bool ResourceAccessWriteState::IsOrderedWriteHazard(VkPipelineStageFlags2KHR src_exec_scope,
-                                                    const SyncStageAccessFlags &src_access_scope) const {
-    // Must be neither in the access scope, nor in the chained access scope
-    return !WriteInScope(src_access_scope) && !WriteInChainedScope(src_exec_scope, src_access_scope);
-}
-
 bool ResourceAccessWriteState::IsWriteBarrierHazard(QueueId queue_id, VkPipelineStageFlags2KHR src_exec_scope,
                                                     const SyncStageAccessFlags &src_access_scope) const {
     // Special rules for sequential ILT's
@@ -8980,8 +8974,8 @@ bool ResourceAccessWriteState::IsWriteBarrierHazard(QueueId queue_id, VkPipeline
     if (WriteInChain(src_exec_scope)) {
         return false;
     }
-    // Otherwise treat as an ordinary write hazard check with ordering rules.
-    return IsOrderedWriteHazard(src_exec_scope, src_access_scope);
+    // The write is not in chain (previous call), so need only to check if the write is in access scope.
+    return !WriteInScope(src_access_scope);
 }
 
 void ResourceAccessWriteState::Set(const SyncStageAccessInfoType &usage_info, ResourceUsageTag tag) {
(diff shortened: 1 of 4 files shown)


spencer-lunarg (approved):
LGTM

layers/sync/sync_validation.cpp
```diff
@@ -8971,6 +8971,11 @@ bool ResourceAccessWriteState::IsWriteBarrierHazard(QueueId queue_id, VkPipeline
             return !WriteInChain(src_exec_scope);
         }
     }
+    // In dep chain means that the write is *available*.
+    // Available writes are automatically made visible and can't cause hazards during transition.
+    if (WriteInChain(src_exec_scope)) {
+        return false;
+    }
```
jzulauf-lunarg, line 8978: DRY'r to unfold the WriteInChain case to a single common check ?
```
    // Special rules for sequential ILT's
    if ( (queue_id == queue_) &&  IsIndex(SYNC_IMAGE_LAYOUT_TRANSITION)) {
        // In queue, they are implicitly ordered
        return false;
     }
        
    // In dep chain means that this write is *available*, and automatically made visible to an ILT
    if (WriteInChain(src_exec_scope)) {
        return false;
    }
```

jzulauf-lunarg (commented):
Good catch on the ILT special rules.  Better without duplicating the WriteInChain...

layers/sync/sync_validation.cpp
artem-lunarg: I wanted to make this change minimal, but I already prepared a separate small cleanup. It turns with added `WriteInChain` we can greatly simplify `IsOrderedWriteHazard` since it includes `WriteInChain` down the callstack:

https://github.com/artem-lunarg/Vulkan-ValidationLayers/commit/ee158f1cf80ace4f76527932b05386986ecb624e

I can add the above suggestion there too to collapse even more.

artem-lunarg: Actually I also have subsequent cleanup here https://github.com/artem-lunarg/Vulkan-ValidationLayers/commit/ee158f1cf80ace4f76527932b05386986ecb624e so I'll add it as separate commits with the above suggestion.

artem-lunarg: Checked the above suggestion one more time and it does not look like equivalent.

Consider that current state such that:
 * `IsIndex(SYNC_IMAGE_LAYOUT_TRANSITION) == true`
 * `queue_id != queue_`
* `WriteInChain(src_exec_scope)` returns **false**


The PR's code:
```
if (IsIndex(SYNC_IMAGE_LAYOUT_TRANSITION)) {
        if (queue_id == queue_) {
            // In queue, they are implicitly ordered
            return false;
        } else {
            // In dep chain means that the ILT is *available*
            return !WriteInChain(src_exec_scope);
        }
    }

  if (WriteInChain(src_exec_scope)) {
        return false;
   }

   return IsOrderedWriteHazard(src_exec_scope, src_access_scope);
```
will return:  `!WriteInChain(src_exec_scope)` == **true**

Suggested version:
```
if ( (queue_id == queue_) &&  IsIndex(SYNC_IMAGE_LAYOUT_TRANSITION)) {
     // In queue, they are implicitly ordered
     return false;
}

  if (WriteInChain(src_exec_scope)) {
        return false;
   }

   return IsOrderedWriteHazard(src_exec_scope, src_access_scope);
```
will return: `IsOrderedWriteHazard(src_exec_scope, src_access_scope)` which can be either **true** or **false**.

arno-lunarg (approved):
LGTM!

layers/sync/sync_validation.cpp
artem-lunarg: Will go with a merge. If there are other clean up/DRY ideas I'll be happy to implement them in a separate PR.

PR #6685, opened 2023-10-12
scripts: Update gpu_inst_shader_hash.h from generate_spirv.py

jeremyg-lunarg:
It is far less error prone to update the hash when recompiling the instrumentation shader functions.


spencer-lunarg (approved):

Tony-LunarG (approved):
Cool! LGTM

PR #6686, opened 2023-10-12
tests: Add min/max core profile and test cleanup

spencer-lunarg:
This adds a `max_core` and `min_core` profile in favor of a random device (ex. pixel 6 profile) to allow for the best test coverage with MockICD on CI

This test also cleanup a lot of test and improve checking for proper support


tests/unit/subgroups.cpp
```diff
@@ -265,43 +266,6 @@ TEST_F(NegativeSubgroup, Properties) {
     }
 }
 
-TEST_F(NegativeSubgroup, Features) {
```
spencer-lunarg, line 268: This is not an actual Validation test, this is trying to do the job of CTS so removed it

arno-lunarg (approved):
LGTM!
Less code :D

PR #6692, opened 2023-10-13
tests: Breakup layer_validation_tests into more files

spencer-lunarg:
removes some dedicated helpers from `layer_validation_tests` into own files to help reduce the bloat in that single file


artem-lunarg (approved):

tests/framework/pipeline_helper.cpp
```diff
@@ -11,6 +11,7 @@
  */
 
 #include "pipeline_helper.h"
+#include "descriptor_helper.h"
```
arno-lunarg, line 14: Can be removed

arno-lunarg (approved):
LBTM

PR #6693, opened 2023-10-13
tests: Remove unused VkBufferTest class

spencer-lunarg:
`VkBufferTest` was only being used for a single test and seems it wasn't even doing what it was expected

### DIFF
--- a/tests/unit/buffer.cpp
+++ b/tests/unit/buffer.cpp
@@ -535,77 +535,34 @@ TEST_F(NegativeBuffer, IdxBufferAlignmentError) {
     m_errorMonitor->VerifyFound();
 }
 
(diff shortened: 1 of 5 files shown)


tests/unit/buffer.cpp
```diff
@@ -535,77 +535,34 @@ TEST_F(NegativeBuffer, IdxBufferAlignmentError) {
     m_errorMonitor->VerifyFound();
 }
 
-TEST_F(NegativeBuffer, VertexBuffer) {
```
spencer-lunarg, line 538: created 2 tests, but the other VUs here already had a test somewhere else doing the same thing

arno-lunarg (approved):

PR #6694, opened 2023-10-13
scripts: Change the outfilename parameter to a directory

s-perron:
Now that generate_spirv.py will always generate gpu_inst_shader_hash.h,
I want to change the parameter to give a directory that will be the base
directory for the generated outputs. This makes the script work better
when there are multiple outputs.

### DIFF
--- a/scripts/generate_spirv.py
+++ b/scripts/generate_spirv.py
@@ -73,7 +73,7 @@ def compile(filename, glslang_validator):
 
     return words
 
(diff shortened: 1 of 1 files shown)


spencer-lunarg (approved):
LGTM - @jeremyg-lunarg can approve and merge if seems good to him too

jeremyg-lunarg (approved):
LGTM

PR #6697, opened 2023-10-14
mac: Use exported_symbols_list

juan-lunarg:
Saves 1,398,896 bytes for the dylib in a Debug build

Similar to --version-script for Linux/Android

Sources:
- https://www.yrom.net/blog/2023/04/19/how-to-explicitly-control-exported-symbols-of-dyamic-shared-libraries/
- https://developer.apple.com/library/archive/documentation/Performance/Conceptual/CodeFootprint/Articles/ReducingExports.html


juan-lunarg:
From the `ld` manual

`-exported_symbols_list filename`

> The specified filename contains a list of global symbol names that will remain as global symbols in the output file.  All other global symbols will be treated as if they were marked as __private_extern__ (aka visibility=hidden) and will not be global in the output file. The symbol names listed in filename must be one per line.  Leading and trailing white space are not part of the symbol name.  Lines starting with # are ignored, as are lines with only white space.  Some wildcards (similar to shell file matching) are supported.  The * matches zero or more characters.  The ? matches one character.  [abc] matches one character which must be an 'a', 'b', or 'c'.  [a-z] matches any single lower case letter from 'a' to 'z'.

layers/CMakeLists.txt
```diff
@@ -283,8 +283,8 @@ elseif(MINGW)
     target_sources(vvl PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/${LAYER_NAME}.def)
     target_compile_options(vvl PRIVATE -Wa,-mbig-obj)
 elseif(APPLE)
-    message(STATUS "Functions are exported via VVL_EXPORT")
     set_target_properties(vvl PROPERTIES SUFFIX ".dylib")
+    target_link_options(vvl PRIVATE -exported_symbols_list ${CMAKE_CURRENT_SOURCE_DIR}/${LAYER_NAME}.exp)
```
juan-lunarg, line 287: Note Apple's version of `ld` is different than on all other platforms.

richard-lunarg (approved):
Nice catch and fix.

spencer-lunarg (approved):

Issue #6009, opened 2023-06-15
Feature Request: Selective shader check in GPU-AV


dorian-apanel-intel:
In current app that I'm debugging, enabling GPU-AV causes drivers to use much more RAM at shader compilation, and total pipelines compile time is increased from 2 minutes to 2 hours.

I know the shader where issue occurs and would like to enable GPU-AV only to a selected shaders.
Ability to specify list in file which shaders are to be validated would help here a lot.

List could be a file of VK_EXT_shader_module_identifier in hex ascii. Alternatively it could it be a DebugUtilsLabel/DebugMarkerObjectName (but those can be set after module creation).

spencer-lunarg:
I agree with this, the biggest question is "how" to select

I think the "cleanest" way would have some `pNext` struct passed down that lets you opt-in/out... but maybe not

dorian-apanel-intel:
For app/game developers pNext would be enough. For driver developers a file or list that could be selected from vkconfig.

jeremyg-lunarg:
@Tony-LunarG is also going to be looking at caching the instrumented shaders, which will help when doing repeated runs and only changing a few shaders. But I agree that this is a good feature to investigate.

@dorian-apanel-intel in the meantime you should make sure you have mimalloc enabled, it should improve the speed of the instrumentation code. For windows you can enable it with the `USE_MIMALLOC` cmake option (it should be on by default), but for Linux you need to use `LD_PRELOAD` as describe [here](https://github.com/microsoft/mimalloc).

Tony-LunarG:
Feature implemented in #6711

PR #6711, opened 2023-10-16
Add GPU-AV option to select which shaders are instrumented

Tony-LunarG:
Addresses #6009
Will also require a spec change to allow a VkValidationFeaturesEXT in the pNext chain of VkShaderModuleCreateInfo (06904)
Updated GPU-AV settings code to use the new utilities which made writing the test for this code much easier
The test does not get a 06904 error, because 06904 isn't implemented, presumably due to pNext being marked noautovalidity

### DIFF
--- a/layers/gpu_validation/gpu_state_tracker.cpp
+++ b/layers/gpu_validation/gpu_state_tracker.cpp
@@ -993,11 +993,11 @@ void GpuAssistedBase::PreCallRecordPipelineCreations(uint32_t count, const Creat
                         auto sm_ci =
                             const_cast<safe_VkShaderModuleCreateInfo *>(reinterpret_cast<const safe_VkShaderModuleCreateInfo *>(
                                 vku::FindStructInPNextChain<VkShaderModuleCreateInfo>(stage_ci.pNext)));
-                        if (select_instrumented_shaders && sm_ci && !CheckForGpuAvEnabled(sm_ci->pNext)) continue;
+                        if (gpuav_settings.select_instrumented_shaders && sm_ci && !CheckForGpuAvEnabled(sm_ci->pNext)) continue;
                         auto &csm_state = cgpl_state.shader_states[pipeline][stage];
                         bool cached = false;
                         bool pass = false;
-                        if (cache_instrumented_shaders) {
+                        if (gpuav_settings.cache_instrumented_shaders) {
                             csm_state.unique_shader_id = ValidationCache::MakeShaderHash(module_state->spirv->words_.data(),
                                                                                          module_state->spirv->words_.size());
                             auto it = instrumented_shaders.find(csm_state.unique_shader_id);
@@ -1019,7 +1019,7 @@ void GpuAssistedBase::PreCallRecordPipelineCreations(uint32_t count, const Creat
                             if (sm_ci) {
                                 sm_ci->SetCode(csm_state.instrumented_spirv);
                             }
-                            if (cache_instrumented_shaders && !cached) {
+                            if (gpuav_settings.cache_instrumented_shaders && !cached) {
                                 instrumented_shaders.emplace(
                                     csm_state.unique_shader_id,
                                     std::make_pair(csm_state.instrumented_spirv.size(), csm_state.instrumented_spirv));
--- a/layers/gpu_validation/gpu_state_tracker.cpp
+++ b/layers/gpu_validation/gpu_state_tracker.cpp
@@ -911,6 +911,18 @@ safe_VkPipelineShaderStageCreateInfo &GetShaderStageCI(safe_VkComputePipelineCre
     return ci.stage;
 }
 
+bool GpuAssistedBase::CheckForGpuAvEnabled(const void *pNext) {
+    auto features = vku::FindStructInPNextChain<VkValidationFeaturesEXT>(pNext);
+    if (features) {
+        for (uint32_t i = 0; i < features->enabledValidationFeatureCount; i++) {
+            if (features->pEnabledValidationFeatures[i] == VK_VALIDATION_FEATURE_ENABLE_GPU_ASSISTED_EXT) {
+                return true;
+            }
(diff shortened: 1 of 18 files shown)


tests/unit/gpu_av.cpp
```diff
@@ -2929,3 +2929,94 @@ TEST_F(VkGpuAssistedLayerTest, DispatchIndirectWorkgroupSizeShaderObjects) {
     ASSERT_EQ(VK_SUCCESS, vk::QueueWaitIdle(m_default_queue));
     m_errorMonitor->VerifyFound();
 }
+
+TEST_F(VkGpuAssistedLayerTest, SelectInstrumentedShaders) {
```
spencer-lunarg, line 2933: it would be nice to have a positive test where there are 2 shaders, one that triggers a GPU-AV error and one that is good

then select only the good one and make sure the selecting is correctly not validating everything by accident (for any future changes)

spencer-lunarg: (I mean one is vertex and one is fragment, so sharing same pipeline)

tests/framework/render.h
```diff
@@ -413,9 +414,11 @@ class VkShaderObj : public vkt::ShaderModule {
                                                       const char *entry_point = "main");
 
   protected:
+    VkShaderModuleCreateInfo m_module_info;
     VkPipelineShaderStageCreateInfo m_stage_info;
     VkRenderFramework &m_framework;
     vkt::Device &m_device;
     const char *m_source;
     spv_target_env m_spv_env;
+    std::vector<uint32_t> spv;
```
spencer-lunarg, line 423: I rather not have to keep the spir-v data around after we create the shader module for every test now. Needing to pass a `pNext` is rare

instead of going

```cpp
    auto inst_module_ci = vs.GetModuleCreateInfo();
    inst_module_ci.pNext = &features;
    vkt::ShaderModule instrumented_module;
    instrumented_module.init(*m_device, inst_module_ci);
```

could we just add another argument to `VkShaderObj::VkShaderObj` and `InitFromGLSL` to pass in the `const void* pNext`

this also doesn't produce this half-use-`vkt::ShaderModule` and half-use-`VkShaderObj` for this test (as I still plan to try and combine these in the future)

layers/gpu_validation/gpu_validation.cpp
```diff
@@ -154,12 +154,6 @@ void GpuAssisted::CreateDevice(const VkDeviceCreateInfo *pCreateInfo) {
     GpuAssistedBase::CreateDevice(pCreateInfo);
     Location loc(vvl::Func::vkCreateDevice);
 
-    validate_descriptors = GpuGetOption("khronos_validation.gpuav_descriptor_checks", true);
-    validate_draw_indirect = GpuGetOption("khronos_validation.validate_draw_indirect", true);
-    validate_dispatch_indirect = GpuGetOption("khronos_validation.validate_dispatch_indirect", true);
-    warn_on_robust_oob = GpuGetOption("khronos_validation.warn_on_robust_oob", true);
-    cache_instrumented_shaders = GpuGetOption("khronos_validation.use_instrumented_shader_cache", true);
-    select_instrumented_shaders = GpuGetOption("khronos_validation.select_instrumented_shaders", false);
     validate_instrumented_shaders = (GetEnvironment("VK_LAYER_GPUAV_VALIDATE_INSTRUMENTED_SHADERS").size() > 0);
```
spencer-lunarg, line 157: why is this not in the new `gpuav_settings` as well?

tests/unit/gpu_av.cpp
Tony-LunarG: Added

tests/framework/render.h
Tony-LunarG: Reworked as suggested

layers/gpu_validation/gpu_validation.cpp
Tony-LunarG: The thought at the time that we added the environment variable was that it was for GPU-AV developers more than end users, and that end users wouldn't be able to do much with the information.

layers/gpu_validation/gpu_state_tracker.cpp
```diff
@@ -911,6 +911,20 @@ safe_VkPipelineShaderStageCreateInfo &GetShaderStageCI(safe_VkComputePipelineCre
     return ci.stage;
 }
 
+bool GpuAssistedBase::CheckForSelectFeature(const void *pNext) {
```
spencer-lunarg, line 914: could this not be simplified to 

```c++
bool GpuAssistedBase::CheckForSelectFeature(const void *pNext) {
    auto features = vku::FindStructInPNextChain<VkValidationFeaturesEXT>(pNext);
    if (features) {
        for (uint32_t i = 0; i < features->enabledValidationFeatureCount; i++) {
            if (features->pEnabledValidationFeatures[i] == VK_VALIDATION_FEATURE_ENABLE_GPU_ASSISTED_EXT) {
                return true;
            }
        }    
    } 
    return false;
}
```

spencer-lunarg: also the function name seems misleading, not how `VK_VALIDATION_FEATURE_ENABLE_GPU_ASSISTED_EXT` == `SelectFeature`

maybe a better name is `CheckForGpuAvEnabled` or `CheckForEnabledFeature`.... #NamingIsHard

spencer-lunarg (approved):
LGTM (with last nits)

Future work will be adding to the XML to allow for the struct to be a valid `pNext` type (this is related the work I need for https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6089).. so I will take care of that

layers/gpu_validation/gpu_state_tracker.cpp
Tony-LunarG: Simplified the code renamed the routine

PR #6713, opened 2023-10-16
ci: Prevent duplicate CI runs

juan-lunarg:
Currently if a user doesn't developer in a fork they run twice the amount of CI.

This fixes that issue.

See:
https://github.com/orgs/community/discussions/26276


juan-lunarg:
This removes automatic testing on forks. @charles-lunarg pointed this out to me.

PR #6715, opened 2023-10-17
layers: Autogen enabled feature detection 

ShabbyX:
This change adds autogeneration for the DeviceFeatures struct as well as
a helper to populate it.

The DeviceFeatures struct is now a flat list of features; i.e. members
of VkPhysicalDevice*Features* classes.  This facilitates looking up
features everywhere else as it is irrelevant through which struct the
feature was enabled.  It will also reduce churn the next time a struct
such as VkPhysicalDeviceVulkan13Features is introduced.

Paired with this struct is a helper that populates the DeviceFeatures
struct with the contents of VkPhysicalDeviceFeatures as well as whatever
feature struct is found in the pNext chain of the device create info.


ShabbyX:
This needs to land as one commit, but I've kept it in 4 different commits for easier review.

**Please consider as RFC right now, I haven't verified it builds/works**

layers/state_tracker/state_tracker.cpp
```diff
@@ -677,282 +678,7 @@ std::shared_ptr<QUEUE_STATE> ValidationStateTracker::CreateQueue(VkQueue q, uint
...
-        }
-    }
-
-    if (nullptr == enabled_features_found) {
-        enabled_features.core = {};
-    } else {
-        enabled_features.core = *enabled_features_found;
-    }
-
-    const auto *vulkan_13_features = vku::FindStructInPNextChain<VkPhysicalDeviceVulkan13Features>(pCreateInfo->pNext);
```
spencer-lunarg, line 694: Something I don't see in the new script is catching that some features are in `VkPhysicalDeviceVulkan13Features` and sometimes they can pass in the EXT/KHR struct instead

layers/state_tracker/state_tracker.cpp
```diff
@@ -969,489 +695,6 @@ void ValidationStateTracker::CreateDevice(const VkDeviceCreateInfo *pCreateInfo)
...
-
-        const auto *texel_buffer_alignment_features =
-            vku::FindStructInPNextChain<VkPhysicalDeviceTexelBufferAlignmentFeaturesEXT>(pCreateInfo->pNext);
-        if (texel_buffer_alignment_features) {
-            enabled_features.texel_buffer_alignment_features = *texel_buffer_alignment_features;
-        }
-        // texelBufferAlignment was not promoted to VkPhysicalDeviceVulkan13Features
-        // but the feature is automatically enabled.
-        // Setting the feature explicitly to 'false' doesn't change that
-        if (api_version >= VK_API_VERSION_1_3) {
```
spencer-lunarg, line 1078: some of these strange edge cases might be good to just inline in the python code

layers/state_tracker/state_tracker.cpp
```diff
@@ -677,282 +678,7 @@ std::shared_ptr<QUEUE_STATE> ValidationStateTracker::CreateQueue(VkQueue q, uint
...
-            enabled_features.core13.maintenance4 = maintenance4_features->maintenance4;
-        }
-    }
-
-    const auto *vulkan_12_features = vku::FindStructInPNextChain<VkPhysicalDeviceVulkan12Features>(pCreateInfo->pNext);
-    if (vulkan_12_features) {
-        enabled_features.core12 = *vulkan_12_features;
-    } else {
-        // Set Extension Feature Aliases to false as there is no struct to check
-        enabled_features.core12.drawIndirectCount = VK_FALSE;
```
spencer-lunarg, line 783: be aware of this nasty edge case in the spec too

image

ShabbyX: Oh nice, thanks, I should definitely fix that up.

layers/state_tracker/state_tracker.cpp
ShabbyX: That's handled appropriately if you look at generated/state_tracker_helper.cpp. The way it's done is that for each feature struct that is encountered, its features are `||`ed with the corresponding entry in DeviceFeatures. If a feature is present in multiple structs, it can get set to true when encountering it in either of the structs.

For example, search for `synchronization2` or `pipelineCreationCacheControl` in that file.

layers/state_tracker/state_tracker.cpp
ShabbyX: Good catch. I didn't know this exists; is this documented in the spec like the other table you showed me so I can work based on that?

ShabbyX: Done

layers/state_tracker/state_tracker.cpp
ShabbyX: Done

layers/state_tracker/state_tracker.cpp
spencer-lunarg: there is nothing machine readable I think, just was in the [Promotion to Vulkan 1.3](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#_promotion_to_vulkan_1_3_21) notes

layers/vulkan/generated/state_tracker_helper.cpp
```diff
@@ -0,0 +1,1710 @@
...
+         pNext = pNext->pNext) {
+        switch (pNext->sType) {
+            case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_16BIT_STORAGE_FEATURES: {
+                const VkPhysicalDevice16BitStorageFeatures *enabled =
+                    reinterpret_cast<const VkPhysicalDevice16BitStorageFeatures *>(pNext);
+                features->storageBuffer16BitAccess =
+                    features->storageBuffer16BitAccess || enabled->storageBuffer16BitAccess == VK_TRUE;
+                features->uniformAndStorageBuffer16BitAccess =
+                    features->uniformAndStorageBuffer16BitAccess || enabled->uniformAndStorageBuffer16BitAccess == VK_TRUE;
+                features->storagePushConstant16 = features->storagePushConstant16 || enabled->storagePushConstant16 == VK_TRUE;
```
spencer-lunarg, line 107: not a deal breaker, but would nice to generate this as 

```suggestion
                features->storagePushConstant16 |= enabled->storagePushConstant16 == VK_TRUE;
```

layers/vulkan/generated/state_tracker_helper.cpp
```diff
@@ -0,0 +1,1710 @@
...
+                features->meshShaderQueries = features->meshShaderQueries || enabled->meshShaderQueries == VK_TRUE;
+                break;
+            }
+
+            default:
+                break;
+        }
+    }
+
+    for (uint32_t i = 0; i < pCreateInfo->enabledExtensionCount; i++) {
```
spencer-lunarg, line 1679: having a comment here in the generated code would be nice (since this is where people will look when their IDE directs them here)

scripts/generators/state_tracker_helper_generator.py
```diff
@@ -0,0 +1,215 @@
...
+                    if (features2 != nullptr) {
+                        core_features = &features2->features;
+                    }
+                }
+                if (core_features != nullptr) {
+            ''')
+        for member in self.vk.structs['VkPhysicalDeviceFeatures'].members:
+            if member.type == 'VkBool32':
+                out.append(f'features->{member.name} = core_features->{member.name} == VK_TRUE;\n')
+        out.append('''}
```
spencer-lunarg, line 138: nit:

```suggestion
        out.append('}\n')
```

should do the same thing and easier to read I feel

layers/vulkan/generated/state_tracker_helper.cpp
ShabbyX: It's technically not the same thing (logical vs bitwas or), which is why I avoided it, but I see VVL does this all over so fine, done.

layers/vulkan/generated/state_tracker_helper.cpp
ShabbyX: Done.

scripts/generators/state_tracker_helper_generator.py
ShabbyX: Done

PR #6718, opened 2023-10-17
tests: Improve feature requirements handling

arno-lunarg:
Improve how feature requirements are handled in tests. See updates to `docs/creating_tests.md` for details.

I started using it in some tests, and plan to expand usage if everyone is happy about this change.

For a quick example, here is the new flow I propose to specify feature requirements:

```cpp
AddRequiredExtensions(VK_EXT_EXTENDED_DYNAMIC_STATE_3_EXTENSION_NAME);
RETURN_IF_SKIP(InitFramework())

auto feature_requirements = MakeFeatureRequirementsCollection(
    FEATURE_REQUIREMENTS(VkPhysicalDeviceFeatures2,
       FEATURE_MANDATORY(features.logicOp), FEATURE_DISABLED(features.fillModeNonSolid)),
    FEATURE_REQUIREMENTS(VkPhysicalDeviceExtendedDynamicState3FeaturesEXT, FEATURE_MANDATORY(extendedDynamicState3PolygonMode)));

RETURN_IF_SKIP(InitStateWithRequirements(feature_requirements))
```

### DIFF
--- a/docs/creating_tests.md
+++ b/docs/creating_tests.md
@@ -96,6 +96,26 @@ if (copy_commands2) {
 }
 ```
 
+### Pattern for feature requirements
+
+A set of helper functions and macro are available to help defining feature requirements. Use them to specify that you want some feature to be enabled or disabled. Test will be skipped if the required features are not supported by the device. Typical flow looks like this:
+
+```cpp
+AddRequiredExtensions(VK_EXT_EXTENDED_DYNAMIC_STATE_3_EXTENSION_NAME);
+RETURN_IF_SKIP(InitFramework())
+
+auto feature_requirements = MakeFeatureRequirementsCollection(
+    FEATURE_REQUIREMENTS(VkPhysicalDeviceFeatures2,
+       FEATURE_MANDATORY(features.logicOp), FEATURE_DISABLED(features.fillModeNonSolid)),
+    FEATURE_REQUIREMENTS(VkPhysicalDeviceExtendedDynamicState3FeaturesEXT, FEATURE_MANDATORY(extendedDynamicState3PolygonMode)));
+
+RETURN_IF_SKIP(InitStateWithRequirements(feature_requirements))
+```
+`MakeFeatureRequirementsCollection` is a variadic function taking a collection of feature requirements as an input.
+`FEATURE_REQUIREMENTS` is a variadic macro. Start by specifying the name of your Vulkan physical device features struct, then use either `FEATURE_MANDATORY` to specify that you want a feature to be enabled, or `FEATURE_DISABLED` to specify that you want a feature to be disabled.
+
+Note: Every Vulkan physical device features must appear at most once in the collection. 
+
 ### Vulkan Version
 
 As [raised in a previous issue](https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/1553), when a Vulkan version is enabled, all extensions that are required are exposed. (For example, if the test is created as a Vulkan 1.1 application, then the `VK_KHR_16bit_storage` extension would be supported regardless as it was promoted to Vulkan 1.1 core)
--- a/tests/framework/feature_requirements.h
+++ b/tests/framework/feature_requirements.h
@@ -1,0 +1,147 @@
+/*
+ * Copyright (c) 2023 Valve Corporation
+ * Copyright (c) 2023 LunarG, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+#pragma once
+
+#include "test_common.h"
+
+#include <memory>
+#include <string>
+#include <vector>
+#include <tuple>
+#include <utility>
+
+template <typename VkFeatureA, typename VkFeatureB>
+void pnext_connect(VkFeatureA &feature_a, VkFeatureB &feature_b) {
+    feature_a.feature_struct.pNext = &feature_b.feature_struct;
+}
+
+// Adapted from https://stackoverflow.com/questions/4832949/c-iterating-over-a-tuple-resolution-of-type-vs-constant-parameters
+
+template <std::size_t I = 0, typename... Tp>
+inline typename std::enable_if<(I + 1) >= sizeof...(Tp), void>::type pnext_connect_vk_tuple(std::tuple<Tp...> &) {}
+
+template <std::size_t I = 0, typename... Tp>
+inline typename std::enable_if<(I + 1) < sizeof...(Tp), void>::type pnext_connect_vk_tuple(std::tuple<Tp...> &t) {
+    pnext_connect(*std::get<I>(t), *std::get<I + 1>(t));
+    pnext_connect_vk_tuple<I + 1, Tp...>(t);
+}
+
+template <std::size_t I = 0, typename... Tp>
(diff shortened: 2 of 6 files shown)


tests/framework/layer_validation_tests.h
```diff
@@ -153,6 +156,13 @@ class VkLayerTest : public VkLayerTestBase {
     const char *kValidationLayerName = "VK_LAYER_KHRONOS_validation";
     const char *kSynchronization2LayerName = "VK_LAYER_KHRONOS_synchronization2";
 
+    template <typename... T>
+    void InitStateWithRequirements(FeatureRequirementsCollection<T...> &feature_requirements) {
```
spencer-lunarg, line 160: eThis makes it seem like `InitState` can't be used with requirements

I guess I would like a future where we have

`InitState(); // no features` or `InitState(features);` so why not just make an overloaded `InitState`?

arno-lunarg: I would like this too in the future, for now I just wanted to make the separation between `InitState` and this new variant clearer

tests/framework/feature_requirements.h
```diff
@@ -0,0 +1,147 @@
...
+  private:
+    std::tuple<T...> feature_structs;
+};
+
+// Use this to specify that a feature needs to be enabled
+// eg: FEATURE_REQUIREMENTS(VkPhysicalDeviceFeatures2, FEATURE_MANDATORY(features.fillModeNonSolid))
+#define FEATURE_MANDATORY(field) feature_req->AddMandatoryFeature(&feature_req->feature_struct.field, #field);
+// Use this to specify that a feature needs to be disabled
+// eg: FEATURE_REQUIREMENTS(VkPhysicalDeviceFeatures2, FEATURE_DISABLED(features.fillModeNonSolid))
+#define FEATURE_DISABLED(field) feature_req->AddDisabledFeature(&feature_req->feature_struct.field, #field);
```
artem-lunarg, line 109: Is by default all features are enabled?

docs/creating_tests.md
```diff
@@ -96,6 +96,26 @@ if (copy_commands2) {
...
+
+A set of helper functions and macro are available to help defining feature requirements. Use them to specify that you want some feature to be enabled or disabled. Test will be skipped if the required features are not supported by the device. Typical flow looks like this:
+
+```cpp
+AddRequiredExtensions(VK_EXT_EXTENDED_DYNAMIC_STATE_3_EXTENSION_NAME);
+RETURN_IF_SKIP(InitFramework())
+
+auto feature_requirements = MakeFeatureRequirementsCollection(
+    FEATURE_REQUIREMENTS(VkPhysicalDeviceFeatures2,
+       FEATURE_MANDATORY(features.logicOp), FEATURE_DISABLED(features.fillModeNonSolid)),
```
artem-lunarg, line 109: `FEATURE_MANDATORY(features.logicOp)` - is _fearures_ name required? another FEATURE_MANDATORY does not use it.

artem-lunarg:
I like the idea since we have a lot of tests and features is one of the areas that are not concise enough yet. I usually spend some time to figure out how to properly enable feature.

Not sure about solution with metaprogramming. My main concern is compile time if a lot of tests use this. I'm less worrying about implementation if API is good but yes, implementation is advanced.

Can we make usage less macro look like? Will it help with API usage ergonomic? `FeatureRequirements`, `NeedFeature`/ `EnableFeature`, `DisableFeature`. Might be challenging because macros are visible everywhere.

One alternative that might provide better API is to generate code using scripts instead of metaprogramming. Allows to design any form of the API (regular C++ code), probably less affects compilation time and might be easier to read/maintain implementation if there is a need.

Did not dig into this topic, not 100% sure if there are no obstacles in script codegen approach, but idea that scripting approach might allow to do something like this (random example, many ways to design api like this):
```
Features features;
features.nullDescriptor().syncrhonization2();
features.shaderFloat16().shaderInt8Optional();

RETURN_IF_SKIP(InitState(...))

assert(features.isShaderFloat16());
if (features.isShaderInt8()) // check optional

```

I like idea with script generation instead of metaprogramming generation (some new C++ features to make templated approach simpler?). But current solution in terms of API in my opinion looks better than before.

ncesario-lunarg (commented):
I forgot I took a look at this problem a while back, but never had time to flesh it out. [This branch](https://github.com/ncesario-lunarg/Vulkan-ValidationLayers/commit/bdf4695c2f163b19bee7e8b1478c39b78f755dea#diff-be80f7891b2babdd8c11e0144d5d7fab6589ef8104b3db0848351d41d88a5db1R155) is incomplete and probably doesn't compile, but sort of shows the end usage I was going for at the time, namely an "init object". One of the ideas behind the "init object" was that a "test class" could define a "base" configuration from which individual tests can start with and adjust. I can't recall the specifics now, but I think that was desirable at the time (that may not be the case anymore :)).

It's a bit hard to not be wary of the maintainability of this solution (and it's hard to understate my distaste for macros), but given this is purely for making tests easier to write, the implementation is probably fine. I think it's more a question of whether or not the usage in individual tests makes sense, and I think the usage looks pretty nice!

docs/creating_tests.md
arno-lunarg: Unfortunately no, for `VkPhysicalDeviceFeatures2` you have to access `VkPhysicalDeviceFeatures` using its `features` member

tests/framework/feature_requirements.h
arno-lunarg: Right now by default every supported feature for some Feature struct will be enabled, except for the one tagged with `FEATURE_DISABLED`. But yes it is better to just enable the ones the user asked for, at least so that the set of enabled features is always the same for every device.
Discussing with Spencer `FEATURE_OPTIONAL` should not be a thing, when you find yourself in this situation it is better to write 2 tests, one with the feature enabled and another one with it disabled

arno-lunarg:
> I like the idea since we have a lot of tests and features is one of the areas that are not concise enough yet. I usually spend some time to figure out how to properly enable feature.
> 
> Not sure about solution with metaprogramming. My main concern is compile time if a lot of tests use this. I'm less worrying about implementation if API is good but yes, implementation is advanced.
> 
> Can we make usage less macro look like? Will it help with API usage ergonomic? `FeatureRequirements`, `NeedFeature`/ `EnableFeature`, `DisableFeature`. Might be challenging because macros are visible everywhere.
> 
> One alternative that might provide better API is to generate code using scripts instead of metaprogramming. Allows to design any form of the API (regular C++ code), probably less affects compilation time and might be easier to read/maintain implementation if there is a need.
> 
> Not 100% sure if there are no obstacles in script codegen approach, but idea that scripting approach might allow to do something like this (random example):
> 
> ```
> Features features;
> features.nullDescriptor().syncrhonization2();
> features.shaderFloat16().shaderInt8Optional();
> 
> RETURN_IF_SKIP(InitState(...))
> 
> assert(features.isShaderFloat16());
> if (features.isShaderInt8()) // check optional
> ```
> 
> I like idea with script generation more, but current solution in terms of API looks better than before.

I need to think more about a code gen solution. Since the goal is to use this system for every tests, compilation time could become an issue

tests/framework/feature_requirements.h
artem-lunarg: > Discussing with Spencer FEATURE_OPTIONAL should not be a thing, when you find yourself in this situation it is better to write 2 tests

+1

docs/creating_tests.md
artem-lunarg: Okay, thanks for clarifying this

arno-lunarg:
closing in favor of https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7126

PR #6722, opened 2023-10-17
Skip pipeline topology validation at pipeline create time if it's dynamic

### LINKED ISSUE #6717: Not checking for dynamic topology
We are not accounting for `vkCmdSetPrimitiveTopology` for `VUID-VkGraphicsPipelineCreateInfo-pStages-00738` or `VUID-VkGraphicsPipelineCreateInfo-pStages-00739`

Would want t a test to confirm and prevent regression

ziga-lunarg:
Closes #6717

### DIFF
--- a/layers/core_checks/cc_shader_interface.cpp
+++ b/layers/core_checks/cc_shader_interface.cpp
@@ -210,7 +210,7 @@ bool CoreChecks::ValidatePrimitiveTopology(const SPIRV_MODULE_STATE &module_stat
     bool skip = false;
 
     if (!create_info.pipeline || !create_info.pipeline->pre_raster_state || !create_info.pipeline->InputAssemblyState() ||
-        entrypoint.stage != VK_SHADER_STAGE_GEOMETRY_BIT) {
+        entrypoint.stage != VK_SHADER_STAGE_GEOMETRY_BIT || create_info.pipeline->IsDynamic(VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY)) {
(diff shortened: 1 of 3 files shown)


layers/core_checks/cc_shader_interface.cpp
```diff
@@ -210,7 +210,7 @@ bool CoreChecks::ValidatePrimitiveTopology(const SPIRV_MODULE_STATE &module_stat
     bool skip = false;
 
     if (!create_info.pipeline || !create_info.pipeline->pre_raster_state || !create_info.pipeline->InputAssemblyState() ||
-        entrypoint.stage != VK_SHADER_STAGE_GEOMETRY_BIT) {
+        entrypoint.stage != VK_SHADER_STAGE_GEOMETRY_BIT || create_info.pipeline->IsDynamic(VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY)) {
```
spencer-lunarg, line 213: I was expecting us to use the `dynamic_state_value.primitive_topology`, where in the spec does it say we skip checking?

ziga-lunarg: It's a VkGraphicsPipelineCreateInfo VUID, I don't think I'm supposed to validate this at draw time. I think a separate VUID would be required.

spencer-lunarg: wow I totally misread this, but yes, this is good then

spencer-lunarg (approved):

PR #6725, opened 2023-10-18
layers: Fix ExternalFormatResolve UNDEFINED format

spencer-lunarg:
Fix the logic for `VUID-VkAttachmentDescription2-format-09332` and `VUID-VkAttachmentDescription2-format-09334`

image


artem-lunarg (approved):
GLTM The logic looks correct.

PR #6739, opened 2023-10-18
WIP layers: Remove unnecessary of IsExtEnabled

spencer-lunarg:
Few places `IsExtEnabled` is not needed

### DIFF
--- a/layers/core_checks/cc_spirv.cpp
+++ b/layers/core_checks/cc_spirv.cpp
@@ -2275,8 +2275,7 @@ bool CoreChecks::ValidatePipelineShaderStage(const StageCreateInfo &stage_create
     if (IsExtEnabled(device_extensions.vk_qcom_render_pass_shader_resolve)) {
         skip |= ValidateShaderResolveQCOM(module_state, stage, stage_create_info, loc);
     }
-    if (stage_create_info.pipeline && IsExtEnabled(device_extensions.vk_khr_dynamic_rendering) &&
-        IsExtEnabled(device_extensions.vk_khr_multiview)) {
+    if (stage_create_info.pipeline) {
         if (stage == VK_SHADER_STAGE_FRAGMENT_BIT &&
             stage_create_info.pipeline->GetCreateInfo<VkGraphicsPipelineCreateInfo>().renderPass == VK_NULL_HANDLE &&
             module_state.HasCapability(spv::CapabilityInputAttachment)) {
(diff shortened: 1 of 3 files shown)


layers/core_checks/cc_spirv.cpp
```diff
@@ -2279,8 +2279,7 @@ bool CoreChecks::ValidatePipelineShaderStage(const StageCreateInfo &stage_create
     if (IsExtEnabled(device_extensions.vk_qcom_render_pass_shader_resolve)) {
         skip |= ValidateShaderResolveQCOM(module_state, stage, stage_create_info, loc);
     }
-    if (stage_create_info.pipeline && IsExtEnabled(device_extensions.vk_khr_dynamic_rendering) &&
```
spencer-lunarg, line 2282: the `vk_khr_dynamic_rendering` might be here because it checks for `renderPass == null`, but I have no idea why multiview was here... will let CI confirm these were not needed

arno-lunarg (approved):

layers/gpu_validation/gpu_validation.cpp
```diff
@@ -60,27 +60,27 @@ struct BindlessStateBuffer {
...
-          enabled_features.core12.descriptorBindingSampledImageUpdateAfterBind ||
-          enabled_features.core12.descriptorBindingStorageImageUpdateAfterBind ||
-          enabled_features.core12.descriptorBindingStorageBufferUpdateAfterBind ||
-          enabled_features.core12.descriptorBindingUniformTexelBufferUpdateAfterBind ||
-          enabled_features.core12.descriptorBindingStorageTexelBufferUpdateAfterBind ||
-          enabled_features.core12.descriptorBindingUpdateUnusedWhilePending ||
-          enabled_features.core12.descriptorBindingPartiallyBound ||
-          enabled_features.core12.descriptorBindingVariableDescriptorCount || enabled_features.core12.runtimeDescriptorArray));
+    // If no features are enabled, descriptor indexing is not being used actually
+    bool result = enabled_features.core12.descriptorIndexing || enabled_features.core12.shaderInputAttachmentArrayDynamicIndexing ||
```
ShabbyX, line 64: IIUC, for older spec versions it should do `IsExtEnabled(device_extensions.vk_ext_descriptor_indexing) ||` because that's the same thing as `descriptorIndexing` (because `descriptorIndexing` is set to false currently, fixed in my PR). If you land this after mine, then the change is ok.

ShabbyX (commented):
nit: typo in the PR title

layers/gpu_validation/gpu_validation.cpp
spencer-lunarg: ok, happy to just wait, also you can just make the change for this function as well in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6715

Issue #6738, opened 2023-10-18
Misc VUs missing coverage


spencer-lunarg:
These are found during my searching for things not validated yet, these didn't match any group and making a single issue as they are all relatively simple to add

- [x] VUID-VkBindBufferMemoryInfo-bufferDeviceAddress-03339
- [x] VUID-VkCommandBufferBeginInfo-flags-09240
- [x] VUID-VkPipelineLayoutCreateInfo-descriptorType-06531 (related VUID-VkPipelineLayoutCreateInfo-descriptorType-02217)
- [x] VUID-VkRenderPassMultiviewCreateInfo-multiview-06555
- [x] VUID-vkCmdBindPipeline-pipelineBindPoint-06653 (Spencer to see why it is a duplicate)
- [x] VUID-vkCmdBindPipeline-pipelineBindPoint-06721

ziga-lunarg:
Is `VUID-vkCmdBindPipeline-pipelineBindPoint-06653` also new? Seems like a copy of `VUID-vkCmdBindPipeline-pipeline-03382`

spencer-lunarg:
aww, seems like it, will take care of that one

PR #6749, opened 2023-10-18
layers: Validate multiview feature is enabled when render pass view masks are used

ziga-lunarg:
Part of #6738

### DIFF
--- a/layers/core_checks/cc_render_pass.cpp
+++ b/layers/core_checks/cc_render_pass.cpp
@@ -2436,6 +2436,12 @@ bool CoreChecks::PreCallValidateCreateRenderPass(VkDevice device, const VkRender
         bool all_zero = true;
         bool all_not_zero = true;
         for (uint32_t i = 0; i < multiview_info->subpassCount; ++i) {
+            if (!enabled_features.multiview && multiview_info->pViewMasks[i] != 0) {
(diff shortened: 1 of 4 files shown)


layers/core_checks/cc_render_pass.cpp
```diff
@@ -2441,6 +2441,12 @@ bool CoreChecks::PreCallValidateCreateRenderPass(VkDevice device, const VkRender
         bool all_zero = true;
         bool all_not_zero = true;
         for (uint32_t i = 0; i < multiview_info->subpassCount; ++i) {
+            if (!enabled_features.multiview_features.multiview && multiview_info->pViewMasks[i] != 0) {
+                skip |= LogError("VUID-VkRenderPassMultiviewCreateInfo-multiview-06555", device, error_obj.location,
+                                 "multiview feature is not enabled, but "
+                                 "VkRenderPassMultiviewCreateInfo->pViewMask[%" PRIu32 "] is %" PRIu32 ".",
```
spencer-lunarg, line 2447: nit: for view mask I print it out as hex `0x%" PRIx32 "`  since it is a mask

(I know I still have a todo to fix that everywhere)

spencer-lunarg (approved):

PR #6755, opened 2023-10-19
tests: Remove VkDescriptorSetObj

spencer-lunarg:
Replace `VkDescriptorSetObj` with the `OneOffDescriptorSet` helper class

This way we only have **one** way to allocate descriptors


tests/unit/pipeline.cpp
```diff
@@ -146,7 +142,6 @@ TEST_F(NegativePipeline, DisabledIndependentBlend) {
 
     CreatePipelineHelper pipe(*this, 2);
     pipe.InitState();
-    pipe.gp_ci_.layout = descriptorSet.GetPipelineLayout();
```
spencer-lunarg, line 149: since most tests will possibly want a descriptor by default, `CreatePipelineHelper` has an equivalent of `AppendDummy` already

artem-lunarg (approved):

Issue #3792, opened 2022-02-28
VK_KHR_acceleration_structure: missing buffer & deviceAddress VUIDs


jeremyg-lunarg:
**Describe the situation in which you encountered the missing validation**
Issue description:
A recent PR (https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/3782) added tracking for buffers used by deviceAddress in several ray tracing commands. However, there are many VUIDs that still need to be implemented in this area.

**Valid Usage IDs requested**
Please include the valid usage IDs for the checks you are requesting:

(NOTE: most of the below also occur in `vkCmdBuildAccelerationStructuresIndirectKHR` too)

For the ones that do need GPU-AV to be validated, see https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5858

- [x] VUID-vkCmdBuildAccelerationStructuresKHR-dstAccelerationStructure-03701
- [x] VUID-vkCmdBuildAccelerationStructuresKHR-dstAccelerationStructure-03702
- [x] VUID-vkCmdBuildAccelerationStructuresKHR-dstAccelerationStructure-03703 (Bug, see https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6040)
- [x] VUID-vkCmdBuildAccelerationStructuresKHR-firstVertex-03770
- [x] VUID-vkCmdBuildAccelerationStructuresKHR-mode-04628
- [x] VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03668
- [x] VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03671
- [x] VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03672
- [x] VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03675
- [x] VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03707
- [x] VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03708
- [x] VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03711
- [x] VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03712
- [x] VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03761
- [x] VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03762
- [x] VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03763
[...]

PR #6756, opened 2023-10-19
layers: Add `vkCmdBuildAccelerationStructuresKHR` VUIDs related to acceleration structures builds constraints

arno-lunarg:
- Partially addresses https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/3792
Also add `04628`, that simply checks that the build mode is a valid value

Add a new utility similar to `vvl::span`, `vvl::enumerate`, that is basically a span where but with the addition of a loop variable holding the current iteration index. `vvl::make_span` and `vvl::enumerate` should cover the needs of the majority of our `for` loops, and probably should be preferred to the usual basic `for` loop. Usage example:
```cpp
std::vector<int> v = {1, 2, 3, 4};
for (auto [x, i] : vvl::enumerate(s.data(), s.size())) {// x is an int*
     std::cout << "Value at iteration " << i << ": " << *x << '\n';
}
```

### DIFF
--- a/layers/core_checks/cc_ray_tracing.cpp
+++ b/layers/core_checks/cc_ray_tracing.cpp
@@ -164,10 +164,10 @@ bool CoreChecks::PreCallValidateCmdBuildAccelerationStructuresKHR(
         return skip;
     }
 
-    for (uint32_t info_i = 0; info_i < infoCount; ++info_i) {
+    for (const auto [info, info_i] : vvl::enumerate(pInfos, infoCount)) {
         const Location info_loc = error_obj.location.dot(Field::pInfos, info_i);
-        const auto src_as_state = Get<ACCELERATION_STRUCTURE_STATE_KHR>(pInfos[info_i].srcAccelerationStructure);
-        const auto dst_as_state = Get<ACCELERATION_STRUCTURE_STATE_KHR>(pInfos[info_i].dstAccelerationStructure);
+        const auto src_as_state = Get<ACCELERATION_STRUCTURE_STATE_KHR>(info->srcAccelerationStructure);
+        const auto dst_as_state = Get<ACCELERATION_STRUCTURE_STATE_KHR>(info->dstAccelerationStructure);
 
         if (dst_as_state != nullptr) {
             skip |= ValidateMemoryIsBoundToBuffer(commandBuffer, *dst_as_state->buffer_state,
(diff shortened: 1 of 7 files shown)


layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -200,36 +200,150 @@ bool CoreChecks::PreCallValidateCmdBuildAccelerationStructuresKHR(
...
+                                                  uint32_t geometry_i) -> const safe_VkAccelerationStructureGeometryKHR & {
+                        if (info.pGeometries) {
+                            return info.pGeometries[geometry_i];
+                        } else {
+                            return *info.ppGeometries[geometry_i];
+                        }
+                    };
+
+                    for (uint32_t geom_i = 0; geom_i < info->geometryCount; ++geom_i) {
+                        const VkAccelerationStructureGeometryKHR &updated_geometry = deref_geometry(*info, geom_i);
```
spencer-lunarg, line 249: what is the advantage of using a lambda here if its only used once and could just got

```c++
const VkAccelerationStructureGeometryKHR &updated_geometry = info->pGeometries ? info.pGeometries[geom_i] : *info.ppGeometries[geom_i];
```

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -200,36 +200,150 @@ bool CoreChecks::PreCallValidateCmdBuildAccelerationStructuresKHR(
...
+                        const safe_VkAccelerationStructureGeometryKHR &last_geometry =
+                            deref_safe_geometry(src_as_state->build_info_khr, geom_i);
+
+                        if (updated_geometry.geometryType != last_geometry.geometryType) {
+                            const LogObjectList objlist(device, commandBuffer, info->srcAccelerationStructure);
+                            skip |= LogError("VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03761", objlist,
+                                             info_loc.dot(info->pGeometries ? Field::pGeometries : Field::ppGeometries, geom_i)
```
spencer-lunarg, line 256: maybe be nicer to just have a single

`const Location geom_loc = info_loc.dot(info->pGeometries ? Field::pGeometries : Field::ppGeometries, geom_i);`

above

layers/stateless/sl_ray_tracing.cpp
```diff
@@ -1139,39 +1139,41 @@ bool StatelessValidation::manual_PreCallValidateCmdBuildAccelerationStructuresKH
     const VkAccelerationStructureBuildRangeInfoKHR *const *ppBuildRangeInfos, const ErrorObject &error_obj) const {
     bool skip = false;
     skip |= ValidateAccelerationStructureBuildGeometryInfoKHR(pInfos, infoCount, 0, error_obj.handle, error_obj.location);
-    for (uint32_t i = 0; i < infoCount; ++i) {
-        const Location info_loc = error_obj.location.dot(Field::pInfos, i);
-        if (SafeModulo(pInfos[i].scratchData.deviceAddress,
+    for (const auto [info, info_i] : vvl::enumerate(pInfos, infoCount)) {
```
spencer-lunarg, line 1142: :100:

tests/unit/ray_tracing.cpp
```diff
@@ -3339,6 +3339,189 @@ TEST_F(NegativeRayTracing, WriteAccelerationStructuresPropertiesMaintenance1) {
...
+    m_errorMonitor->VerifyFound();
+    m_commandBuffer->end();
+}
+
+TEST_F(NegativeRayTracing, BuildAccelerationStructuresInvalidUpdatesToGeometry) {
+    TEST_DESCRIPTION(
+        "Build a list of destination acceleration structures, then do an update build on that same list but with invalid updates "
+        "to the geometries");
+
+    SetTargetApiVersion(VK_API_VERSION_1_1);
```
spencer-lunarg, line 3371: move all this init setup to a init function (like `InitBasicDynamicRendering`) and then it would be nice to just make a separate test for each command buffer you have as they are already isolated

layers/core_checks/cc_ray_tracing.cpp
arno-lunarg: Wow yes indeed much simpler, why did I go for that x_x

tests/unit/ray_tracing.cpp
arno-lunarg: I will not touch the Init part for now, I will do it when my feature requirements gets merged. I will split the tests though!

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -200,36 +200,137 @@ bool CoreChecks::PreCallValidateCmdBuildAccelerationStructuresKHR(
...
+                    auto deref_geometry = [](const VkAccelerationStructureBuildGeometryInfoKHR &info,
+                                             uint32_t geometry_i) -> const VkAccelerationStructureGeometryKHR & {
+                        if (info.pGeometries) {
+                            return info.pGeometries[geometry_i];
+                        } else {
+                            return *info.ppGeometries[geometry_i];
+                        }
+                    };
+
+                    auto deref_safe_geometry = [](const safe_VkAccelerationStructureBuildGeometryInfoKHR &info,
```
spencer-lunarg, line 239: this and `deref_geometry` are unused now

spencer-lunarg (approved):
LGTM with nits

PR #6759, opened 2023-10-19
Validate binding ray tracing pipeline in protected command buffer

ziga-lunarg:
Part of #6738

### DIFF
--- a/tests/unit/protected_memory.cpp
+++ b/tests/unit/protected_memory.cpp
@@ -1091,3 +1091,75 @@ TEST_F(NegativeProtectedMemory, MixingProtectedResources) {
     vk::QueueSubmit(m_default_queue, 1, &submit_info, VK_NULL_HANDLE);
     m_errorMonitor->VerifyFound();
 }
+
+TEST_F(NegativeProtectedMemory, RayTracingPipeline) {
+    TEST_DESCRIPTION("Bind ray tracing pipeline in a protected command buffer");
+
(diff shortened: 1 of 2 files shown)


tests/unit/protected_memory.cpp
```diff
@@ -1091,3 +1091,75 @@ TEST_F(NegativeProtectedMemory, MixingProtectedResources) {
...
+    vk::CreateRayTracingPipelinesKHR(m_device->handle(), VK_NULL_HANDLE, VK_NULL_HANDLE, 1, &raytracing_pipeline_ci, nullptr,
+                                     &raytracing_pipeline);
+
+    m_commandBuffer->begin();
+    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-vkCmdBindPipeline-pipelineBindPoint-06721");
+    vk::CmdBindPipeline(m_commandBuffer->handle(), VK_PIPELINE_BIND_POINT_RAY_TRACING_KHR, raytracing_pipeline);
+    m_errorMonitor->VerifyFound();
+    m_commandBuffer->end();
+
+    vk::DestroyPipeline(device(), raytracing_pipeline, nullptr);
```
spencer-lunarg, line 1164: not something I think needs to be fixed here, but we have a `RayTracingPipelineHelper` helper class, I have never used it, but my long term goal is not to have these `vkDestroy*` calls where not needed

spencer-lunarg (approved):

PR #6769, opened 2023-10-20
layers: Fix Clear Attachment 07271

spencer-lunarg:
we currently were printing

> vkCmdClearAttachments(): pAttachments[0].colorAttachment (1) is larger than colorAttachmentCount (1)

it looking, it seems our logic was messed up

### DIFF
--- a/tests/unit/command_positive.cpp
+++ b/tests/unit/command_positive.cpp
@@ -155,13 +155,7 @@ TEST_F(PositiveCommand, ClearAttachmentsCalledWithoutFbInSecondaryCB) {
     info.pInheritanceInfo = &hinfo;
 
     secondary.begin(&info);
-    VkClearAttachment color_attachment = {};
-    color_attachment.aspectMask = VK_IMAGE_ASPECT_COLOR_BIT;
-    color_attachment.clearValue.color.float32[0] = 0.0;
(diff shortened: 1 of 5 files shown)


tests/unit/command_positive.cpp
```diff
@@ -1207,3 +1201,23 @@ TEST_F(PositiveCommand, CopyBufferToRemaingImageLayers) {
     vk::CmdCopyBufferToImage2KHR(m_commandBuffer->handle(), &copy_buffer_to_image);
     m_commandBuffer->end();
 }
+
+TEST_F(PositiveCommand, ClearAttachmentBasicUsage) {
+    TEST_DESCRIPTION("Points to a wrong colorAttachment index in a VkClearAttachment structure passed to vkCmdClearAttachments");
```
arno-lunarg, line 1206:
```suggestion
    TEST_DESCRIPTION("Points to a correct colorAttachment index in a VkClearAttachment structure passed to vkCmdClearAttachments");
```
?

arno-lunarg (approved):
LGTM, One question

tests/unit/command_positive.cpp
spencer-lunarg: copy and paste error ™️ - will fix

PR #6788, opened 2023-10-22
Validate viewport state in graphics pipeline create, add vuid 04892

### LINKED ISSUE #4366: Missing VUID-VkGraphicsPipelineCreateInfo-pViewportState-04892
Found by code inspection while debugging a Mesa driver.

**Valid Usage IDs requested**
`VUID-VkGraphicsPipelineCreateInfo-pViewportState-04892`

**Additional Context**
The missing VUID is a sibling of `VUID-VkGraphicsPipelineCreateInfo-rasterizerDiscardEnable-00750`, which is handled correctly.

ziga-lunarg:
Closes #4366

### DIFF
--- a/layers/stateless/sl_pipeline.cpp
+++ b/layers/stateless/sl_pipeline.cpp
@@ -1096,8 +1096,11 @@ bool StatelessValidation::manual_PreCallValidateCreateGraphicsPipelines(
                                 "is VK_TRUE but the depthClipControl feature was not enabled.");
                         }
                     }
-                } else if (!has_dynamic_viewport_with_count || !has_dynamic_scissor_with_count ||
-                           !IsExtEnabled(device_extensions.vk_ext_extended_dynamic_state3)) {
+                } else if ((!has_dynamic_viewport_with_count || !has_dynamic_scissor_with_count ||
+                            !IsExtEnabled(device_extensions.vk_ext_extended_dynamic_state3)) &&
+                           (vvl::Contains(dynamic_state_map, VK_DYNAMIC_STATE_RASTERIZER_DISCARD_ENABLE) ||
+                            (!vvl::Contains(dynamic_state_map, VK_DYNAMIC_STATE_RASTERIZER_DISCARD_ENABLE) &&
+                             !create_info.pRasterizationState->rasterizerDiscardEnable))) {
                     skip |= LogError("VUID-VkGraphicsPipelineCreateInfo-rasterizerDiscardEnable-09024", device, create_info_loc,
                                      "Rasterization is enabled (pCreateInfos[%" PRIu32
                                      "].pRasterizationState->rasterizerDiscardEnable is VK_FALSE), but pCreateInfos[%" PRIu32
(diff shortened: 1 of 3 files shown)


layers/stateless/sl_pipeline.cpp
```diff
@@ -1384,6 +1384,15 @@ bool StatelessValidation::manual_PreCallValidateCreateGraphicsPipelines(
                 }
             }
 
+            if (create_info.pRasterizationState != nullptr &&
+                vvl::Contains(dynamic_state_map, VK_DYNAMIC_STATE_RASTERIZER_DISCARD_ENABLE) && !create_info.pViewportState) {
+                skip |= LogError("VUID-VkGraphicsPipelineCreateInfo-pViewportState-04892", device, create_info_loc,
```
spencer-lunarg, line 1389: this will become `09024` in the future when https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6151 is merged

so for now, would just use `09024` instead of `04892`

and then we want a positive test where 
- `VK_DYNAMIC_STATE_VIEWPORT_WITH_COUNT` is used
- `VK_DYNAMIC_STATE_SCISSOR_WITH_COUNT` is used
- `VK_EXT_extended_dynamic_state3` is enabled

and then `pViewportState` is allowed to be null as well

tests/unit/pipeline.cpp
```diff
@@ -3546,3 +3546,33 @@ TEST_F(NegativePipeline, MissingPipelineFormat) {
...
+    }
+    if (!eds2_features.extendedDynamicState2) {
+        GTEST_SKIP() << "extendedDynamicState2 not supported";
+    }
+    RETURN_IF_SKIP(InitState(nullptr, &eds2_features));
+
+    CreatePipelineHelper pipe(*this);
+    pipe.AddDynamicState(VK_DYNAMIC_STATE_SCISSOR_WITH_COUNT);
+    pipe.AddDynamicState(VK_DYNAMIC_STATE_VIEWPORT_WITH_COUNT_EXT);
+    pipe.AddDynamicState(VK_DYNAMIC_STATE_RASTERIZER_DISCARD_ENABLE);
```
spencer-lunarg, line 3571: having these 3 dynamic state with `VK_EXT_extended_dynamic_state3` would make this legal to have `pViewportState` null

a negative test would be not having one of these

layers/stateless/sl_pipeline.cpp
ziga-lunarg: `09024` is already implemented, should I just close this PR?

spencer-lunarg: ok, wow this just gets more entangled... 

Ya, I think the issue is closed, we have added `09024` (and it accounts for `rasterizerDiscardEnable`)  and `04892` willl soon be removed for being redundant

... It might be worth keeping the tests from this PR, mainly because I was shocked what looked like a positive test was a negative test (unless I missed something subtle about it)

ziga-lunarg: I have updated the 09024 VUID to do what the spec MR says it will do. I have also added negative and positive tests for it

spencer-lunarg (approved):

Issue #6792, opened 2023-10-23
v1.3.269 cosolidated/refactored VUIDs


mikes-lunarg:
The 1.3.269 header came with a larger-than-usual list of refactored VUIDs. I assume that most of these will be simple re-numberings, but it is a long enough list that I'm punting to VVL team.

```
Following VUIDs found in layer code are not defined in validusage.json (46):
    VUID-VkDescriptorDataEXT-type
    VUID-VkFramebufferCreateInfo-renderPass-02747
    VUID-VkGraphicsPipelineCreateInfo-renderPass-06575
    VUID-vkCmdDraw-None-08882
    VUID-vkCmdDrawIndexed-None-08882
    VUID-vkCmdDrawIndexedIndirect-None-08882
    VUID-vkCmdDrawIndexedIndirectCount-None-08882
    VUID-vkCmdDrawIndirect-None-08882
    VUID-vkCmdDrawIndirectByteCountEXT-None-08882
    VUID-vkCmdDrawIndirectCount-None-08882
    VUID-vkCmdDrawMultiEXT-None-08882
    VUID-vkCmdDrawMultiIndexedEXT-None-08882
    VUID-vkCmdSetAlphaToCoverageEnableEXT-None-08506
    VUID-vkCmdSetAlphaToOneEnableEXT-None-08508
    VUID-vkCmdSetColorBlendAdvancedEXT-None-08592
    VUID-vkCmdSetColorBlendEnableEXT-None-08536
    VUID-vkCmdSetColorBlendEquationEXT-None-08538
    VUID-vkCmdSetColorWriteMaskEXT-None-08540
    VUID-vkCmdSetConservativeRasterizationModeEXT-None-08570
    VUID-vkCmdSetCoverageModulationModeNV-None-08530
    VUID-vkCmdSetCoverageModulationTableEnableNV-None-08532
    VUID-vkCmdSetCoverageModulationTableNV-None-08534
    VUID-vkCmdSetCoverageReductionModeNV-None-08528
    VUID-vkCmdSetCoverageToColorEnableNV-None-08524
    VUID-vkCmdSetCoverageToColorLocationNV-None-08526
    VUID-vkCmdSetDepthClampEnableEXT-None-08582
    VUID-vkCmdSetDepthClipEnableEXT-None-08584
    VUID-vkCmdSetDepthClipNegativeOneToOneEXT-None-08586
[...]

PR #6793, opened 2023-10-23
build: Update to header 1.3.269

mikes-lunarg:
- Update known-good
- Generate source


mikes-lunarg:
Opened #6792 to deal with the VUID churn separately

spencer-lunarg:
@mikes-lunarg should be good now if you rebase

spencer-lunarg (approved):

PR #6804, opened 2023-10-23
layers: Update to xxhash v0.8.2

juan-lunarg:
v0.8.2 fixes the chromium build failure which was the reason XXH_NO_LONG_LONG was added to the build.

This allows us to use the faster hashes provide by xxhash

Furthermore I've removed xxhash.c and created hash_util.cpp

xxhash.c is NOT needed for projects using xxhash

STB style libraries supply this C file for 3 primary reasons.
- package management
- examples
- convenience

None of which matter in our case.

This allows us to better track what exactly we are using from xxhash and ensure we properly set PREPROCESSOR definitions when using the library to avoid the pitfalls of STB style headers.


spencer-lunarg (approved):
Awesome, thanks @juan-lunarg this is way better then my attempt to update `xxhash` would have been

PR #6810, opened 2023-10-24
layers: Cleanup pipeline state generic helpers

spencer-lunarg:
Does 3 things at the same time

- Cleanup the `BasePipeline`, `BasePipelineIndex`, and `GetShaderModuleByCIIndex` helper not needed
- Cleanup use of Derivatives (only placed that used it)
- Added a early return to `manual_PreCallValidateCreateGraphicsPipelines` as it got really deeply nested

### DIFF
--- a/layers/stateless/sl_pipeline.cpp
+++ b/layers/stateless/sl_pipeline.cpp
@@ -394,1207 +394,1182 @@ bool StatelessValidation::manual_PreCallValidateCreateGraphicsPipelines(
     const VkAllocationCallbacks *pAllocator, VkPipeline *pPipelines, const ErrorObject &error_obj) const {
     bool skip = false;
 
-    if (pCreateInfos != nullptr) {
-        for (uint32_t i = 0; i < createInfoCount; ++i) {
-            const Location create_info_loc = error_obj.location.dot(Field::pCreateInfos, i);
-            bool has_pre_raster_state = true;
-            // Create a copy of create_info and set non-included sub-state to null
-            auto create_info = pCreateInfos[i];
-            const auto *graphics_lib_info = vku::FindStructInPNextChain<VkGraphicsPipelineLibraryCreateInfoEXT>(create_info.pNext);
-            if (graphics_lib_info) {
-                if (!(graphics_lib_info->flags & VK_GRAPHICS_PIPELINE_LIBRARY_VERTEX_INPUT_INTERFACE_BIT_EXT)) {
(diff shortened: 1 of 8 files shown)


layers/stateless/sl_pipeline.cpp
```diff
@@ -394,1208 +394,1183 @@ bool StatelessValidation::manual_PreCallValidateCreateGraphicsPipelines(
...
-                                 "(%s) must not include VK_PIPELINE_CREATE_RAY_TRACING_ALLOW_MOTION_BIT_NV.",
-                                 string_VkPipelineCreateFlags(flags).c_str());
+        if (flags & VK_PIPELINE_CREATE_DERIVATIVE_BIT) {
+            if (create_info.basePipelineHandle != VK_NULL_HANDLE) {
+                if (create_info.basePipelineIndex != -1) {
+                    skip |= LogError(
+                        "VUID-VkGraphicsPipelineCreateInfo-flags-07986", device, create_info_loc.dot(Field::basePipelineIndex),
+                        "%" PRIu32 " and basePipelineHandle is not VK_NULL_HANDLE.", pCreateInfos[i].basePipelineIndex);
+                }
+            } else if (static_cast<uint32_t>(create_info.basePipelineIndex) >= createInfoCount) {
```
artem-lunarg, line 1519: I see that it's existing code, but might be good opportunity to fix potential issues with negative values. `basePipelineIndex` is signed and should not be casted to uint32. The same reason to check for lower bound (not only -1): `index < 0 || index >= count`.

artem-lunarg (approved):
LGTM. One suggestion to update signed integer code in one place.

layers/stateless/sl_pipeline.cpp
spencer-lunarg: yep, good catch

spencer-lunarg: actually I forgot how awful the `07984`/`07985`/`07986` VUs are, so if `basePipelineIndex` is negative it must be `-1` and if it is, that case is handled already

artem-lunarg: Two observations. The above checks are in the `object lifetimes` which runs later. But I realized that cast of negative value to uint32_t results in a very big number (need to re-check, but probably the case), so (index >= count) will be true, basically handles both large positive and negative values. I wish such places had a comment...

PR #6811, opened 2023-10-24
git: Use less minutes for GitHub Actions

spencer-lunarg:
An attempt to use less minutes for our normal github actions

https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions

Currently we are waiting hours for our GitHub CI to run

### DIFF
--- a/.github/workflows/vvl.yml
+++ b/.github/workflows/vvl.yml
@@ -35,7 +35,7 @@ permissions:
   contents: read
 
 jobs:
-  linux:
+  linuxBuild:
     env:
       CMAKE_C_COMPILER_LAUNCHER: ccache
       CMAKE_CXX_COMPILER_LAUNCHER: ccache
@@ -43,12 +43,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-22.04]
-        compiler: [{cc: gcc, cxx: g++}]
-        flags: [{c: "-fsanitize=address", ld: "-fsanitize=address"}, {c: "-fsanitize=thread", ld: "-fsanitize=thread"}]
         config: [debug, release]
-        cpp_std: [17]
-        robin_hood: ["ON"]
         include:
           # Test Ubuntu-20.04 release build works.
           - os: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


.github/workflows/vvl.yml
```diff
@@ -35,20 +35,15 @@ permissions:
   contents: read
 
 jobs:
-  linux:
+  linuxBuild:
```
spencer-lunarg, line 38: From what I found, the gcc and ubuntu 20 type CI is great for catching build errors, not so much for testing

.github/workflows/vvl.yml
```diff
@@ -151,8 +183,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        arch: [ amd64, amd64_x86 ]
```
spencer-lunarg, line 154: have yet to see not all or none of the Window clients fail

.github/workflows/vvl.yml
```diff
@@ -225,7 +257,7 @@ jobs:
     runs-on: ${{matrix.os}}
     strategy:
       matrix:
-        os: [ macos-11, macos-12 ]
+        os: [ macos-12 ]
```
spencer-lunarg, line 260: at a 10x cost for minutes, this might be better served for internal CI (maybe the other way around if internal CI is Macos 11)

.github/workflows/vvl.yml
juan-lunarg: When I added amd64_x86 it caught a few 32 bit issues. I'm fine with not testing release builds here.

.github/workflows/vvl.yml
juan-lunarg: I'd rather leave this alone for now. I agree this could be better served for internal CI potentially. But since it isn't being done I'd rather not change this right now.

.github/workflows/vvl.yml
juan-lunarg: agreed

.github/workflows/vvl.yml
```diff
@@ -62,6 +57,43 @@ jobs:
...
+      - name: Install WSI dependencies
+        run: sudo apt-get -qq update && sudo apt-get install -y libwayland-dev xorg-dev
+      - name: Build
+        run: python scripts/tests.py --build --config ${{ matrix.config }} --cmake='-DVVL_CPP_STANDARD=${{ matrix.cpp_std }} -DUSE_ROBIN_HOOD_HASHING=${{matrix.robin_hood}}'
+        env:
+          CC: ${{ matrix.compiler.cc }}
+          CXX: ${{ matrix.compiler.cxx }}
+          CFLAGS: ${{ matrix.flags.c }}
+          CXXFLAGS: ${{ matrix.flags.c }}
+          LDFLAGS: ${{ matrix.flags.ld }}
```
juan-lunarg, line 83: None of these apply anymore.

.github/workflows/vvl.yml
```diff
@@ -62,6 +57,43 @@ jobs:
...
+      - uses: lukka/get-cmake@latest
+        with:
+          cmakeVersion: 3.17.2
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - name: Setup ccache
+        uses: hendrikmuhs/ccache-action@v1.2
+        with:
+          key: ${{ matrix.os }}-${{ matrix.config }}-${{ matrix.compiler.cc }}-${{ matrix.compiler.cxx }}-${{ matrix.cpp_std }}-${{ matrix.flags.c }}-${{ matrix.flags.ld }}-${{matrix.robin_hood}}
```
juan-lunarg, line 71: This key uses matrix values that don't exist

juan-lunarg (changes requested):

mikes-lunarg:
I just noticed there is a usage page for each run. Are you referencing this to trim down the matrix?
https://github.com/KhronosGroup/Vulkan-ValidationLayers/actions/runs/6627852969/usage

spencer-lunarg:
@mikes-lunarg the term "matrix" I ment is like we do Windows for `[32-bit, 64-bit]` and `[debug, release]` so instead of doing all 4 matrix permutation, just to do what we need to save minutes used

.github/workflows/vvl.yml
spencer-lunarg: I guess do we know of any MacOS 11 users, just as we stop supporting older Windows/Linux/C++ version, I feel MacOS should be same (especially for a platform that prides itself having a high adoption rate for updates)

spencer-lunarg:
closing for https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6818

couldn't update the branch easily because it is in this repo and not my fork

.github/workflows/vvl.yml
juan-lunarg: In our SDK documentation we say we support the latest 2 MacOS releases

Same rational as supporting Ubuntu 20.04

PR #6814, opened 2023-10-24
Cleanup linker options and copy pasted defines

juan-lunarg:
FYI @charles-lunarg / @mikes-lunarg

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -466,9 +466,6 @@ foreach(layer_info, layers) {
       defines += [ "NOMINMAX" ]
       sources += [ "layers/VkLayer_$name.def" ]
     }
-    if (is_linux || is_android || is_fuchsia) {
-      ldflags += [ "-Wl,-Bsymbolic,--exclude-libs,ALL" ]
-    }
     if (defined(ozone_platform_x11) && ozone_platform_x11) {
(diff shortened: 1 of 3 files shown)


BUILD.gn
```diff
@@ -68,13 +68,6 @@ config("vulkan_internal_config") {
   if (is_clang && is_mac) {
     cflags += [ "-Wno-unguarded-availability-new" ]
   }
-  if (is_linux) {
-    defines += [
-      "SYSCONFDIR=\"/etc\"",
-      "FALLBACK_CONFIG_DIRS=\"/etc/xdg\"",
-      "FALLBACK_DATA_DIRS=\"/usr/local/share:/usr/share\"",
```
charles-lunarg, line 75: Is this because the vulkan-loader uses these as default values (and so don't need to be set manually)?

charles-lunarg: So because the loader's build.gn sets these, they do not need to be set here.

juan-lunarg: Yep.

https://github.com/KhronosGroup/Vulkan-Loader/blob/7028f47e870bf8885c24f39df3c193d4a7f6b55e/BUILD.gn#L38

spencer-lunarg (approved):

PR #6818, opened 2023-10-25
git: Use less minutes for GitHub Actions

spencer-lunarg:
An attempt to use less minutes for our normal github actions

https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions

Currently we are waiting hours for our GitHub CI to run

(redo of https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6811)

### DIFF
--- a/.github/workflows/vvl.yml
+++ b/.github/workflows/vvl.yml
@@ -35,7 +35,52 @@ permissions:
   contents: read
 
 jobs:
-  linux:
+  # CI we just care builds, but no need to spend Github Action minutes running the tests
+  linuxBuild:
+    env:
+      CMAKE_C_COMPILER_LAUNCHER: ccache
+      CMAKE_CXX_COMPILER_LAUNCHER: ccache
+    runs-on: ${{ matrix.os }}
+    strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/vvl.yml
```diff
@@ -225,7 +254,9 @@ jobs:
     runs-on: ${{matrix.os}}
     strategy:
       matrix:
-        os: [ macos-11, macos-12 ]
+        # Dev Survey shows little to no users still use MacOS 11 or earlier
+        # iOS run uses macos-11 already also
+        os: [ macos-12 ]
```
juan-lunarg, line 259: We should test for maco-13 now though as well.

Our docs say we support the 2 latest releases

juan-lunarg (approved):

.github/workflows/vvl.yml
juan-lunarg: Currently this runner is in beta however. So I wouldn't add it in this PR.

arno-lunarg (approved):

PR #6820, opened 2023-10-25
layers: Add 03668

arno-lunarg:
Also refacto the ray tracing memory aliasing validation code

### DIFF
--- a/layers/core_checks/cc_ray_tracing.cpp
+++ b/layers/core_checks/cc_ray_tracing.cpp
@@ -637,8 +637,9 @@ bool CoreChecks::PreCallValidateCmdBuildAccelerationStructuresKHR(
                     const LogObjectList objlist(device, commandBuffer, info->srcAccelerationStructure);
                     skip |=
                         LogError("VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03759", objlist, info_loc.dot(Field::mode),
-                                 "is VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR, but flags (%s) must have the same value which"
+                                 "is VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR, but %s (%s) must have the same value which"
                                  " was specified when srcAccelerationStructure was last built (%s).",
+                                 info_loc.dot(Field::flags).Fields().c_str(),
                                  string_VkBuildAccelerationStructureFlagsKHR(info->flags).c_str(),
                                  string_VkBuildAccelerationStructureFlagsKHR(src_as_state->build_info_khr.flags).c_str());
                 }
@@ -814,7 +815,8 @@ bool CoreChecks::PreCallValidateBuildAccelerationStructuresKHR(
                     skip |= LogError("VUID-vkBuildAccelerationStructuresKHR-pInfos-03759", pInfos[i].srcAccelerationStructure,
                                      info_loc.dot(Field::mode),
                                      "is VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR,"
-                                     " but flags (%s) is different then srcAccelerationStructure value (%s).",
+                                     " but %s (%s) is different then srcAccelerationStructure value (%s).",
+                                     info_loc.dot(Field::flags).Fields().c_str(),
                                      string_VkBuildAccelerationStructureFlagsKHR(pInfos[i].flags).c_str(),
                                      string_VkBuildAccelerationStructureFlagsKHR(src_as_state->build_info_khr.flags).c_str());
                 }
@@ -1236,7 +1238,8 @@ bool CoreChecks::PreCallValidateCmdBuildAccelerationStructuresIndirectKHR(VkComm
                 skip |= LogError("VUID-vkCmdBuildAccelerationStructuresIndirectKHR-pInfos-03759",
                                  pInfos[i].srcAccelerationStructure, info_loc.dot(Field::mode),
                                  "is VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR,"
-                                 " but flags (%s) is different than srcAccelerationStructure value (%s).",
+                                 " but %s (%s) is different than srcAccelerationStructure value (%s).",
+                                 info_loc.dot(Field::flags).Fields().c_str(),
                                  string_VkBuildAccelerationStructureFlagsKHR(pInfos[i].flags).c_str(),
                                  string_VkBuildAccelerationStructureFlagsKHR(src_as_state->build_info_khr.flags).c_str());
             }
--- a/layers/core_checks/cc_ray_tracing.cpp
+++ b/layers/core_checks/cc_ray_tracing.cpp
@@ -158,14 +158,15 @@ bool CoreChecks::ValidateAccelerationStructuresMemoryAlisasing(VkCommandBuffer c
 
     bool skip = false;
     const VkAccelerationStructureBuildGeometryInfoKHR &info = pInfos[info_i];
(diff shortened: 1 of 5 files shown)


tests/unit/ray_tracing.cpp
```diff
@@ -2659,8 +2659,8 @@ TEST_F(NegativeRayTracing, CmdBuildAccelerationStructuresKHR) {
     m_commandBuffer->end();
 }
 
-// https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6040
-TEST_F(NegativeRayTracing, DISABLED_AccelerationStructuresOverlappingMemory) {
+// Partially disabled, see https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6040
+TEST_F(NegativeRayTracing, AccelerationStructuresOverlappingMemory) {
```
spencer-lunarg, line 2663: I guess why partially disable it? 

can we not just break the test up so the working parts are in a "ready to go" test and the "not ready stuff" is behind a single test `DISABLED` instead of playing the `if 0` game

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -773,6 +600,199 @@ bool CoreChecks::ValidateAccelerationBuffers(uint32_t info_index, const VkAccele
...
+                skip |= LogError(
+                    "VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03699", objlist, info_loc.dot(Field::type),
+                    "is VK_ACCELERATION_STRUCTURE_TYPE_TOP_LEVEL_KHR, but its dstAccelerationStructure was created with %s.",
+                    string_VkAccelerationStructureTypeKHR(dst_as_state->create_infoKHR.type));
+            }
+        }
+
+        skip |= ValidateAccelerationBuffers(info_i, *info, info_loc);
+
+        skip |= ValidateAccelerationStructuresMemoryAlisasing(commandBuffer, infoCount, pInfos, info_i, error_obj);
```
spencer-lunarg, line 790: why are we not passing in `info_loc` instead of `error_obj`, we lose the index knowledge

spencer-lunarg (changes requested):

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -151,239 +151,69 @@ bool CoreChecks::PreCallValidateGetAccelerationStructureHandleNV(VkDevice device
...
-        for (const auto [other_info, other_info_j] : vvl::enumerate(pInfos, infoCount)) {
-            // Validate that scratch buffer's memory does not overlap destination acceleration structure's memory, or source
-            // acceleration structure's memory if build mode is update, or other scratch buffers' memory.
-            // Here validation is pessimistic: if one buffer associated to pInfos[other_info_j].scratchData.deviceAddress has an
-            // overlap, an error will be logged.
+            std::stringstream dst_as_log_info;
+            dst_as_log_info << "pInfos[" << info_i << "].dstAccelerationStructure (" << FormatHandle(info.dstAccelerationStructure)
+                            << ')';
+
+            validate_no_as_buffer_memory_overlap(*src_as_state, src_as_log_info.str().c_str(), *src_as_buffer, src_as_range,
```
spencer-lunarg, line 204: Why not have the lambda only do the 

```
 if (const auto [memory, overlap_range] = buffer_a.GetResourceMemoryOverlap(range_a, &buffer_b, range_b);
                memory != VK_NULL_HANDLE) {
```

logic, the reason is I like not having to do string concatenation every time when this has no error

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -773,6 +600,199 @@ bool CoreChecks::ValidateAccelerationBuffers(uint32_t info_index, const VkAccele
...
+        }
+        if (info->type == VK_ACCELERATION_STRUCTURE_TYPE_BOTTOM_LEVEL_KHR) {
+            if (!dst_as_state ||
+                (dst_as_state && dst_as_state->create_infoKHR.type != VK_ACCELERATION_STRUCTURE_TYPE_BOTTOM_LEVEL_KHR &&
+                 dst_as_state->create_infoKHR.type != VK_ACCELERATION_STRUCTURE_TYPE_GENERIC_KHR)) {
+                const LogObjectList objlist(device, commandBuffer);
+                skip |= LogError(
+                    "VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03700", objlist, info_loc.dot(Field::type),
+                    "is VK_ACCELERATION_STRUCTURE_TYPE_BOTTOM_LEVEL_KHR, but its dstAccelerationStructure was created with %s.",
+                    string_VkAccelerationStructureTypeKHR(dst_as_state->create_infoKHR.type));
```
spencer-lunarg, line 773: you have `if (!dst_as_state` so this will be a null pointer reference

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -773,6 +600,199 @@ bool CoreChecks::ValidateAccelerationBuffers(uint32_t info_index, const VkAccele
...
+                skip |=
+                    LogError("VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-04630", objlist, info_loc.dot(Field::mode),
+                             "is VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR but srcAccelerationStructure is VK_NULL_HANDLE.");
+            } else if (!src_as_state->built ||
+                       !(src_as_state->build_info_khr.flags & VK_BUILD_ACCELERATION_STRUCTURE_ALLOW_UPDATE_BIT_KHR)) {
+                const LogObjectList objlist(device, commandBuffer, info->srcAccelerationStructure);
+                skip |= LogError("VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03667", objlist, info_loc.dot(Field::mode),
+                                 "is VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR, srcAccelerationStructure has been previously "
+                                 "constructed with flags %s.",
+                                 string_VkBuildAccelerationStructureFlagsKHR(src_as_state->build_info_khr.flags).c_str());
```
spencer-lunarg, line 640: if we hit `!src_as_state->built`  won't these flags value be undefined

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -151,239 +151,69 @@ bool CoreChecks::PreCallValidateGetAccelerationStructureHandleNV(VkDevice device
...
-        error_msg_ss << "memory backing buffer (" << FormatHandle(buffer_handle_a) << ") used as storage for " << variable_name_a
-                     << " overlaps memory backing buffer (" << FormatHandle(buffer_handle_b) << ") used as storage for "
-                     << variable_name_b << ". Overlapped memory is " << FormatHandle(memory_handle) << " on range "
-                     << string_range(overlap_range) << '.';
-
-        return error_msg_ss.str();
-    };
+    bool skip = false;
+    const VkAccelerationStructureBuildGeometryInfoKHR &info = pInfos[info_i];
+    const auto src_as_state = Get<ACCELERATION_STRUCTURE_STATE_KHR>(pInfos[info_i].srcAccelerationStructure);
```
spencer-lunarg, line 161:
```suggestion
    const auto src_as_state = Get<ACCELERATION_STRUCTURE_STATE_KHR>(info.srcAccelerationStructure);
```

layers/core_checks/cc_ray_tracing.cpp
arno-lunarg: nice catch ty

layers/core_checks/cc_ray_tracing.cpp
arno-lunarg: nice catch ty

spencer-lunarg (approved):

PR #6821, opened 2023-10-25
layers: Add more LogError with Location

spencer-lunarg:
In effort to remove the old `LogError`, adds `Location` in some places in order to use the new `LogError`


jeremyg-lunarg (approved):
LGTM

Issue #4847, opened 2022-11-22
Use scripts to help generate code for noautovalidity structs


spencer-lunarg:
For some structs with `noautovalidity` we are currently doing the validation manually in `parameter_validation_utils.cpp` but this runs the risks of not adding new extensions correctly.

Good example are those found in `VkGraphicsPipelineCreateInfo` such as
- `VkPipelineTessellationStateCreateInfo`
- `VkPipelineVertexInputStateCreateInfo`
- `VkPipelineViewportStateCreateInfo`
- `VkPipelineMultisampleStateCreateInfo`
- `VkPipelineColorBlendStateCreateInfo`
- `VkPipelineDepthStencilStateCreateInfo`
- `VkPipelineRasterizationStateCreateInfo`
- `VkPipelineInputAssemblyStateCreateInfo`

These structs are only `noautovalidity` because there is some logic that is needed to decide to ignore them or not, but otherwise the normal generated logic for the struct itself would be the same

PR #6824, opened 2023-10-25
scripts: Generate noautovalidity pipeline structs

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4847

Generate the helper struct functions for Pipeline states that we do the logic to call by hand (since the logic to ignore them is complex)


artem-lunarg (approved):
LGTM

Issue #6537, opened 2023-09-21
NegativeSyncVal.QSOBarrierHazard sporadic failures on NVIDIA


lunarpapillo:
@juan-lunarg and @artem-lunarg report that this test is sporadically failing on the **Windows-NVIDIA** platform using:
```
INFO: deviceName: NVIDIA GeForce RTX 2080 Ti
INFO: deviceType: VK_PHYSICAL_DEVICE_TYPE_DISCRETE_GPU
INFO: vendorID: 4318 (0x10de)
INFO: deviceID: 7687 (0x1e07)
INFO: apiVersion: 4206851 (0x403103) (1.3.259)
INFO: driverVersion: 2231926784 (0x85088000) (532.136.0)
```
(the `driverVersion` uses a slightly different version encoding - this corresponds to NVIDIA driver 532.34)

Investigation into one instance showed that the test provoked a GPU crash and recovery, with events like this showing up in **Event Viewer** at the same time as the failure:
```
Warning 9/21/2023 11:02:14 AM Display 4101
    Display driver nvlddmkm stopped responding and has successfully recovered
Warning 9/21/2023 11:02:13 AM NVIDIA OpenGL Driver 2
    The NVIDIA OpenGL driver has not been able to initialize
    a connection with the GPU.  This might be due to out of memory error,
    an exhaustion of system resources or too many graphical applications running.
    The application will not render correctly.
    (pid=27$
Error 9/21/2023 11:02:13 AM nvlddmkm 0
    \Device\Video3
    Error occurred on GPUID: 100 
```

Subsequent tests passed, so I don't suspect exhaustion of resources is the problem.  Since the test provoked an unexpected GPU reset, there may very well be an issue with the test.

The test is being blacklisted for now until investigation can determine the root cause.

PR #6825, opened 2023-10-25
tests: Fix test that breaks nvidia driver

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6537
Reverts https://github.com/LunarG/VulkanTests/pull/454

The bug was 100% reproducible on the local machine with regular nvidia drivers. Run `QSBufferEvents` test with `gtest_repeat=100` and run 3 or 4 such sessions in parallel. It fails like this:

Corruption due to `QSBufferEvents` affected `QSOBarrierHazard` that runs after it, but `QSOBarrierHazard` never fails when run alone. 

The issue was not detected by the core validation because it's disabled for syncval tests. There are other failures of core validation in syncval tests. The plan to fix them and also to have a separate run of syncval tests with core validation enabled.

### DIFF
--- a/tests/unit/sync_val.cpp
+++ b/tests/unit/sync_val.cpp
@@ -5001,6 +5001,7 @@ TEST_F(NegativeSyncVal, QSBufferEvents) {
     m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, "SYNC-HAZARD-WRITE-AFTER-READ");
     test.Submit1Wait(test.cbb, VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT);
     m_errorMonitor->VerifyFound();
+    m_device->wait();
 }
 
 TEST_F(NegativeSyncVal, QSOBarrierHazard) {


tests/unit/sync_val.cpp
```diff
@@ -5001,6 +5001,7 @@ TEST_F(NegativeSyncVal, QSBufferEvents) {
     m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, "SYNC-HAZARD-WRITE-AFTER-READ");
     test.Submit1Wait(test.cbb, VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT);
     m_errorMonitor->VerifyFound();
+    m_device->wait();
```
spencer-lunarg, line 5004: :open_mouth:

spencer-lunarg (approved):

tests/unit/sync_val.cpp
spencer-lunarg: From a scale from OpenGL to Vulkan Synchronization, how hard was this to solve?

artem-lunarg: I guess Metal level, not terrible.

PR #6831, opened 2023-10-25
layers: Fix pipeline rendering info validation

ziga-lunarg:
Move `VkPipelineRenderingCreateInfo` out of the fragment shader state condition and into a separate function

### DIFF
--- a/layers/stateless/sl_pipeline.cpp
+++ b/layers/stateless/sl_pipeline.cpp
@@ -116,6 +116,52 @@ bool StatelessValidation::ValidatePipelineShaderStageCreateInfo(const VkPipeline
     return skip;
 }
 
+bool StatelessValidation::ValidatePipelineRenderingCreateInfo(const VkPipelineRenderingCreateInfo &rendering_struct,
+                                                              const Location &loc) const {
+    bool skip = false;
+
+    if ((rendering_struct.depthAttachmentFormat != VK_FORMAT_UNDEFINED)) {
+        skip |= ValidateRangedEnum(loc.pNext(Struct::VkPipelineRenderingCreateInfo, Field::depthAttachmentFormat), "VkFormat",
+                                   rendering_struct.depthAttachmentFormat, "VUID-VkGraphicsPipelineCreateInfo-renderPass-06583");
+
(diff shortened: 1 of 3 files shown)


layers/stateless/sl_pipeline.cpp
```diff
@@ -441,58 +487,14 @@ bool StatelessValidation::manual_PreCallValidateCreateGraphicsPipelines(
...
-                    }
-                }
+            const auto rendering_struct = vku::FindStructInPNextChain<VkPipelineRenderingCreateInfo>(create_info.pNext);
+            if (rendering_struct) {
+                color_attachment_count = rendering_struct->colorAttachmentCount;
+                skip |= ValidatePipelineRenderingCreateInfo(*rendering_struct, create_info_loc);
+            }
 
+            // Pipeline has fragment output state
+            if (create_info.pColorBlendState && create_info.pMultisampleState) {
```
spencer-lunarg, line 497: looked at spec, `VkAttachmentSampleCountInfoAMD` itself works like `VkPipelineRenderingCreateInfo`

so basically asking to remove

```patch
-             // Pipeline has fragment output state
-            if (create_info.pColorBlendState && create_info.pMultisampleState) {
```

as this is not a requirement for that struct neither

spencer-lunarg (approved):
LGTM with one nit

layers/stateless/sl_pipeline.cpp
ziga-lunarg: Turns out this might be needed for libraries

layers/stateless/sl_pipeline.cpp
```diff
@@ -168,56 +214,13 @@ bool StatelessValidation::manual_PreCallValidateCreateGraphicsPipelines(
         uint32_t color_attachment_count = 0;
 
         if (!create_info.renderPass) {
-            if (create_info.pColorBlendState && create_info.pMultisampleState) {
+            // Pipeline has fragment output state
+            if ((create_info.flags & VK_PIPELINE_CREATE_LIBRARY_BIT_KHR) == 0 ||
```
spencer-lunarg, line 218: aww, the issue is with GPL, you do need the "full" version of the state... this is all still a mess, but glad to have a test and working case for now

PR #6832, opened 2023-10-25
ci: Cleanup vvl.yml

juan-lunarg:
- Enable linker warnings as errors on all major platforms
- Cleanup matrix config usage
- Only test CMake 3.17.2 on `linuxBuild`
- Remove names that aren't that helpful

### DIFF
--- a/.github/workflows/vvl.yml
+++ b/.github/workflows/vvl.yml
@@ -37,9 +37,6 @@ permissions:
 jobs:
   # CI we just care builds, but no need to spend Github Action minutes running the tests
   linuxBuild:
-    env:
-      CMAKE_C_COMPILER_LAUNCHER: ccache
-      CMAKE_CXX_COMPILER_LAUNCHER: ccache
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/vvl.yml
```diff
@@ -37,100 +37,82 @@ permissions:
...
-      CMAKE_C_COMPILER_LAUNCHER: ccache
-      CMAKE_CXX_COMPILER_LAUNCHER: ccache
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         include:
           # Test Ubuntu-20.04 release build works.
           - os: ubuntu-20.04
-            compiler: {cc: gcc, cxx: g++}
```
juan-lunarg, line 50: The default compiler for linux is already gcc so this doesn't need to be specified.

.github/workflows/vvl.yml
```diff
@@ -37,100 +37,82 @@ permissions:
...
             config: release
             cpp_std: 17
-            robin_hood: "ON"
```
juan-lunarg, line 53: We were only testing "ON" so this can be removed.

.github/workflows/vvl.yml
```diff
@@ -37,100 +37,82 @@ permissions:
...
             config: debug
             cpp_std: 20
-            robin_hood: "ON"
     steps:
       - uses: actions/checkout@v3
-      - uses: lukka/get-cmake@latest
+      - name: Test Minimum CMake Version
+        uses: lukka/get-cmake@latest
         with:
           cmakeVersion: 3.17.2
```
juan-lunarg, line 59: Note all the other builds can use cmake latest. We just need to make sure at least one build is testing the current minimum 3.17.2 to ensure we don't break backcompat by accident.

spencer-lunarg (approved):

.github/workflows/vvl.yml
```diff
@@ -179,11 +154,8 @@ jobs:
   windows:
     runs-on: windows-2022
     strategy:
-      fail-fast: false
```
juan-lunarg, line 182: I left this in when I was debugging a while ago.

PR #6833, opened 2023-10-25
ci: Add dependabot

juan-lunarg:
Ensures action dependencies are kept up to date

Helps with security/performance of github action CI

Security being particularly important since we publish Android binaries via github actions


juan-lunarg:
Example of a `dependabot` PR in Vulkan-Headers:
https://github.com/KhronosGroup/Vulkan-Headers/pull/445

spencer-lunarg (approved):

Issue #6727, opened 2023-10-18
Missing VUs for bound completely and contiguously to a single VkDeviceMemory 


spencer-lunarg:
we have a `ValidateMemoryIsBoundToBuffer` function for VUs with the phrase

> If <buffer> is non-sparse then it must be bound completely and contiguously to a single **VkDeviceMemory** object

but the following seem to be missing 

- [ ] VUID-VkConditionalRenderingBeginInfoEXT-buffer-01981
- [ ] VUID-VkCopyImageToBufferInfo2-dstBuffer-00192
- [ ] VUID-VkCopyImageToBufferInfo2-srcImage-07966

PR #6835, opened 2023-10-26
layers: Remove more UNASSIGNED VUs

spencer-lunarg:
Another round to remove some UNASSIGNED VUIDs

closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6727

### DIFF
--- a/layers/core_checks/cc_synchronization.cpp
+++ b/layers/core_checks/cc_synchronization.cpp
@@ -97,8 +97,7 @@ bool SemaphoreSubmitState::ValidateBinaryWait(const Location &loc, VkQueue queue
         } else if (CannotWait(semaphore_state)) {
             const auto &vuid = GetQueueSubmitVUID(loc, SubmitError::kBinaryCannotBeSignalled);
             const LogObjectList objlist(semaphore, queue);
-            skip |= core->LogError(semaphore_state.Scope() == kSyncScopeInternal ? vuid : kVUID_Core_DrawState_QueueForwardProgress,
-                                   objlist, loc, "queue (%s) is waiting on semaphore (%s) that has no way to be signaled.",
+            skip |= core->LogError(vuid, objlist, loc, "queue (%s) is waiting on semaphore (%s) that has no way to be signaled.",
                                    core->FormatHandle(queue).c_str(), core->FormatHandle(semaphore).c_str());
         } else {
             signaled_semaphores.erase(semaphore);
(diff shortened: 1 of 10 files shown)


layers/core_checks/cc_synchronization.cpp
```diff
@@ -97,8 +97,7 @@ bool SemaphoreSubmitState::ValidateBinaryWait(const Location &loc, VkQueue queue
         } else if (CannotWait(semaphore_state)) {
             const auto &vuid = GetQueueSubmitVUID(loc, SubmitError::kBinaryCannotBeSignalled);
             const LogObjectList objlist(semaphore, queue);
-            skip |= core->LogError(semaphore_state.Scope() == kSyncScopeInternal ? vuid : kVUID_Core_DrawState_QueueForwardProgress,
```
artem-lunarg, line 100: Don't have good mental model about `kSyncScopeInternal` without digging into syncobject code, and why was separate VUID for imported semaphore. Was the idea that it could be signaled in some other way than from submit, or is it because we did not have VUID in the table?

tests/unit/renderpass.cpp
```diff
@@ -1788,44 +1788,85 @@ TEST_F(NegativeRenderPass, FramebufferIncompatible) {
...
+    vkt::RenderPass rp(*m_device, rpci);
+
+    // A compatible framebuffer.
+    VkImageObj image(m_device);
+    image.Init(32, 32, 1, VK_FORMAT_B8G8R8A8_UNORM, VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT, VK_IMAGE_TILING_OPTIMAL, 0);
+
+    VkImageView view = image.targetView(VK_FORMAT_B8G8R8A8_UNORM);
+
+    auto fci = vku::InitStruct<VkFramebufferCreateInfo>(nullptr, 0u, rp.handle(), 1u, &view, 32u, 32u, 1u);
+    vkt::Framebuffer fb(*m_device, fci);
```
artem-lunarg, line 1852: can we remove this framebuffer?

layers/core_checks/cc_synchronization.cpp
spencer-lunarg: I looked, I found no real reason why we had a a UNASSIGNED VU for this and no other spots had a ternary ... truly think this was just added for no good reason

tests/unit/renderpass.cpp
spencer-lunarg: aww yes, copied and pasted this and not used

artem-lunarg (approved):
LGTM. One test comment.

PR #6847, opened 2023-10-26
Fix acceleration structure type when copied to mutable descriptor

Tony-LunarG:
Fixes a crash in dEQP-VK.binding_model.mutable_descriptor.single.acceleration_structure_khr.update_copy.mutable_source.normal_source.pool_same_types.update_after_bind.no_array.comp
with GPU-AV


artem-lunarg (approved):
LGTM

spencer-lunarg (approved):

PR #6849, opened 2023-10-26
layers: Fix best practice check for pipeline library with no rendering

### LINKED ISSUE #6816: BestPractices-Pipeline-NoRendering triggering when using graphics pipeline library
**Environment:**
 - OS:  Windows 10 
 - GPU and driver version: NVIDIA GeForce GTX 1080 Ti VK: 1.3.242 Driver: NVIDIA 537.58
 - SDK or header version if building from repo: vulkan-sdk-1.3.268
 - Options enabled (synchronization, best practices, etc.): BestPratices + NVIDIA specific

**Describe the Issue**

When creating graphics pipeline library pipelines, the following best practice error appears starting with the current SDK version.
> UNASSIGNED-BestPractices-Pipeline-NoRendering(WARN / SPEC): msgNum: -1275560067 - Validation Warning: [ UNASSIGNED-BestPractices-Pipeline-NoRendering ] | MessageID = 0xb3f87f7d | vkCreateGraphicsPipelines(): pCreateInfos[0] renderPass is VK_NULL_HANDLE and pNext chain does not contain VkPipelineRenderingCreateInfoKHR.

However from the spec, it should be allowed to not provide VkPipelineRenderingCreateInfo unless it is the fragment output pipeline.

I think validation should check for `VkGraphicsPipelineLibraryCreateInfoEXT` and if it's in the pnext then check if has `.flags = VK_GRAPHICS_PIPELINE_LIBRARY_FRAGMENT_OUTPUT_INTERFACE_BIT_EXT`. That should cover most cases, to cover all would mean to check for `VkPipelineRenderingCreateInfo.viewMask` and only then error if the rest of the pipelines were not created with `VkPipelineRenderingCreateInfo`

**Expected behavior**

This error doesn't appear

ziga-lunarg:
Closes #6816

### DIFF
--- a/layers/best_practices/bp_pipeline.cpp
+++ b/layers/best_practices/bp_pipeline.cpp
@@ -139,8 +139,10 @@ bool BestPractices::PreCallValidateCreateGraphicsPipelines(VkDevice device, VkPi
 
         skip |= VendorCheckEnabled(kBPVendorArm) && ValidateMultisampledBlendingArm(createInfoCount, pCreateInfos, create_info_loc);
 
+        const auto* graphics_lib_info = vku::FindStructInPNextChain<VkGraphicsPipelineLibraryCreateInfoEXT>(create_info.pNext);
         if (pCreateInfos[i].renderPass == VK_NULL_HANDLE &&
-            !vku::FindStructInPNextChain<VkPipelineRenderingCreateInfoKHR>(pCreateInfos[i].pNext)) {
+            !vku::FindStructInPNextChain<VkPipelineRenderingCreateInfoKHR>(pCreateInfos[i].pNext) &&
+            (!graphics_lib_info || (graphics_lib_info->flags & VK_GRAPHICS_PIPELINE_LIBRARY_FRAGMENT_SHADER_BIT_EXT) != 0)) {
             skip |= LogWarning(kVUID_BestPractices_Pipeline_NoRendering, device, create_info_loc,
                                "renderPass is VK_NULL_HANDLE and pNext chain does not contain VkPipelineRenderingCreateInfoKHR.");
         }
--- a/tests/unit/best_practices_positive.cpp
+++ b/tests/unit/best_practices_positive.cpp
@@ -198,3 +198,23 @@ TEST_F(VkPositiveBestPracticesLayerTest, ImageInputAttachmentLayout) {
                            nullptr, 0u, nullptr, 1u, &image_memory_barrier);
     m_commandBuffer->end();
 }
+
+TEST_F(VkPositiveBestPracticesLayerTest, PipelineLibraryNoRendering) {
+    TEST_DESCRIPTION("Create a pipeline library without a render pass or rendering info");
+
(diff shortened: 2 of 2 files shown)


spencer-lunarg (approved):

layers/best_practices/bp_pipeline.cpp
```diff
@@ -139,8 +139,10 @@ bool BestPractices::PreCallValidateCreateGraphicsPipelines(VkDevice device, VkPi
 
         skip |= VendorCheckEnabled(kBPVendorArm) && ValidateMultisampledBlendingArm(createInfoCount, pCreateInfos, create_info_loc);
 
+        const auto* graphics_lib_info = vku::FindStructInPNextChain<VkGraphicsPipelineLibraryCreateInfoEXT>(create_info.pNext);
         if (pCreateInfos[i].renderPass == VK_NULL_HANDLE &&
-            !vku::FindStructInPNextChain<VkPipelineRenderingCreateInfoKHR>(pCreateInfos[i].pNext)) {
+            !vku::FindStructInPNextChain<VkPipelineRenderingCreateInfoKHR>(pCreateInfos[i].pNext) &&
+            (!graphics_lib_info || (graphics_lib_info->flags & VK_GRAPHICS_PIPELINE_LIBRARY_FRAGMENT_SHADER_BIT_EXT) != 0)) {
```
AlexRouSg, line 145: should this not be `VK_GRAPHICS_PIPELINE_LIBRARY_FRAGMENT_OUTPUT_INTERFACE_BIT_EXT`?

ziga-lunarg: Both of them actually use VkRenderPass or VkPipelineRenderingCreateInfo, I'll make another PR

AlexRouSg: hmmm from my understanding of the spec `VK_GRAPHICS_PIPELINE_LIBRARY_FRAGMENT_OUTPUT_INTERFACE_BIT_EXT` should be the only one that needs it, the rest should be optional.

AlexRouSg: src: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkGraphicsPipelineCreateInfo.html
For all stages except fragment output, it lists
> The viewMask parameter of [VkPipelineRenderingCreateInfo (formats are ignored)

Then in: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPipelineRenderingCreateInfo.html
> If this structure is not specified, and the pipeline does not include a [VkRenderPass](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkRenderPass.html), viewMask and colorAttachmentCount are 0

So it appears that it is fine to only include it in fragment output as long as viewmask of 0 is fine

ziga-lunarg: None of the stages "need it", this is just a best practice.

PR #6850, opened 2023-10-26
scripts: Skip dependabot commits

juan-lunarg:


juan-lunarg:
Fixes CI failures shown here: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6845

artem-lunarg (approved):

PR #6852, opened 2023-10-26
Fix VVL_ENABLE_ASAN

juan-lunarg:


CMakeLists.txt
```diff
@@ -111,10 +111,8 @@ endif()
 
 option(VVL_ENABLE_ASAN "Use address sanitization")
 if (VVL_ENABLE_ASAN)
-    if (MSVC)
-        add_compile_options(/fsanitize=address)
-    else()
-        add_compile_options(-fsanitize=address -fno-omit-frame-pointer)
+    add_compile_options(-fsanitize=address)
```
juan-lunarg, line 114: MSVC compiler accepts either `-` or `/` so I wrote the code to take that into account.

scripts/update_deps.py
```diff
@@ -491,22 +491,10 @@ def CMakeConfig(self, repos):
...
-                cmake_c_flags = cmake_c_flags + ' /fsanitize=address'
-            else:
-                cmake_cxx_flags = cmake_cxx_flags + ' -fsanitize=address -fno-omit-frame-pointer'
-                cmake_c_flags = cmake_c_flags + ' -fsanitize=address -fno-omit-frame-pointer'
-            cmake_cxx_flags = cmake_cxx_flags.strip()
-            cmake_c_flags = cmake_c_flags.strip()
-            cmake_cmd.append(f'-DCMAKE_CXX_FLAGS="{cmake_cxx_flags}" -DCMAKE_C_FLAGS="{cmake_c_flags}"')
-
+            cmake_cmd.append(f'-D CMAKE_CXX_FLAGS=-fsanitize=address')
+            cmake_cmd.append(f'-D CMAKE_C_FLAGS=-fsanitize=address')
```
artem-lunarg, line 495: Previous version added spaces between commands (`' /fsanitize=address'`. Do we need space before `-D` here?

juan-lunarg: > Do we need space before -D here?

The space between the -D is consistent with what is on line 490. Helps with readability IMO as well.

artem-lunarg (approved):

spencer-lunarg (approved):

scripts/update_deps.py
arno-lunarg: I wrote this convoluted logic thinking that going just for `-D CMAKE_CXX_FLAGS=-fsanitize=address` would not append to but overwrite `CMAKE_CXX_FLAGS`. So does this change really appends the sanitize flag?

arno-lunarg (approved):

juan-lunarg:
> So does this change really appends the sanitize flag?

It will just append. Feel free to try it out locally.

Issue #6838, opened 2023-10-26
VVL throws exception and crashes


zmike:
Happens in a lot of places, but can be reproduced running a [piglit](https://gitlab.freedesktop.org/mesa/piglit/) test in zink with `ZINK_DEBUG=vvl MESA_LOADER_DRIVER_OVERRIDE=zink bin/arb_texture_cube_map_array-cubemap -auto -fbo`

```
terminate called after throwing an instance of 'std::out_of_range'
  what():  map::at
```

```
#0  0x00007ffff77cd884 in __pthread_kill_implementation () from /lib64/libc.so.6
#1  0x00007ffff777cafe in raise () from /lib64/libc.so.6
#2  0x00007ffff776587f in abort () from /lib64/libc.so.6
#3  0x00007ffff74a4d19 in __gnu_cxx::__verbose_terminate_handler() [clone .cold] () from /lib64/libstdc++.so.6
#4  0x00007ffff74b4f8c in __cxxabiv1::__terminate(void (*)()) () from /lib64/libstdc++.so.6
#5  0x00007ffff74b4ff7 in std::terminate() () from /lib64/libstdc++.so.6
#6  0x00007ffff74b5258 in __cxa_throw () from /lib64/libstdc++.so.6
#7  0x00007ffff74a7738 in std::__throw_out_of_range(char const*) () from /lib64/libstdc++.so.6
#8  0x00007fffea317a25 in std::map<VkPipelineBindPoint, unsigned int, std::less<VkPipelineBindPoint>, std::allocator<std::pair<VkPipelineBindPoint const, unsigned int> > >::at (
    this=0x7fffebfbf660 <CoreChecks::ValidatePipelineBindPoint(CMD_BUFFER_STATE const*, VkPipelineBindPoint, Location const&, std::map<VkPipelineBindPoint, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::less<VkPipelineBindPoint>, std::allocator<std::pair<VkPipelineBindPoint const, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > > const&) const::flag_mask>, 
    __k=@0x7fffffffb4ac: VK_PIPELINE_BIND_POINT_GRAPHICS) at /usr/include/c++/13/bits/stl_map.h:562
[...]

PR #6855, opened 2023-10-27
layers: Refactor ValidatePipelineBindPoint

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6838

The `std::map` was getting corrupted, but looking more, I realized we really didn't need a hashmap for only 3 entries

The VUID hash map was also easy to replace with the new `Location` and we save carrying around the hash maps and move all the perf cost to error reporting time (as desired)


spencer-lunarg:
> LGTM. This kills the mystery.

Ya I know... but it as least kills it with the cleaner code it should of been

juan-lunarg (approved):
Much better. This std::map usage was questionable anyway.

PR #6860, opened 2023-10-27
syncval: Add dynamic rendering support

jzulauf-lunarg:
Includes tests from @artem-lunarg 's branch.

### DIFF
--- a/layers/sync/sync_validation.cpp
+++ b/layers/sync/sync_validation.cpp
@@ -2669,6 +2669,39 @@ void CommandBufferAccessContext::RecordDrawDynamicRenderingAttachment(ResourceUs
     }
 }
 
+ClearAttachmentInfo CommandBufferAccessContext::GetClearAttachmentInfo(const VkClearAttachment &clear_attachment,
+                                                                       const VkClearRect &rect) const {
+    // This is a NOOP if there's no renderpass nor dynamic rendering
+    // Caller must used "IsValid" to determine if clear_info contains meaningful information.
+    ClearAttachmentInfo clear_info;
+    if (current_renderpass_context_) {
+        clear_info = current_renderpass_context_->GetClearAttachmentInfo(clear_attachment, rect);
+    } else if (dynamic_rendering_info_) {
+        clear_info = dynamic_rendering_info_->GetClearAttachmentInfo(clear_attachment, rect);
+    }
+
+    return clear_info;
+}
+bool CommandBufferAccessContext::ValidateClearAttachment(const Location &loc, const VkClearAttachment &clear_attachment,
+                                                         const VkClearRect &rect) const {
+    bool skip = false;
+
+    ClearAttachmentInfo clear_info = GetClearAttachmentInfo(clear_attachment, rect);
+    if (clear_info.IsValid()) {
+        skip = ValidateClearAttachment(loc, clear_info);
+    }
+
+    return skip;
+}
+
+void CommandBufferAccessContext::RecordClearAttachment(ResourceUsageTag tag, const VkClearAttachment &clear_attachment,
+                                                       const VkClearRect &rect) {
+    ClearAttachmentInfo clear_info = GetClearAttachmentInfo(clear_attachment, rect);
+    if (clear_info.IsValid()) {
+        RecordClearAttachment(tag, clear_info);
+    }
+}
+
 QueueId CommandBufferAccessContext::GetQueueId() const { return QueueSyncState::kQueueIdInvalid; }
 
 ResourceUsageTag CommandBufferAccessContext::RecordBeginRenderPass(
@@ -3046,10 +3079,9 @@ static constexpr VkImageAspectFlags kColorAspects =
     VK_IMAGE_ASPECT_COLOR_BIT | VK_IMAGE_ASPECT_PLANE_0_BIT | VK_IMAGE_ASPECT_PLANE_1_BIT | VK_IMAGE_ASPECT_PLANE_2_BIT;
 static constexpr VkImageAspectFlags kDepthStencilAspects = VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT;
 
-static std::optional<uint32_t> GetAttachmentIndex(const RenderPassAccessContext &rp_context,
-                                                  const VkClearAttachment &clear_attachment) {
-    const auto &rpci = rp_context.GetRenderPassState()->createInfo;
-    const auto &subpass = rpci.pSubpasses[rp_context.GetCurrentSubpass()];
+uint32_t RenderPassAccessContext::GetAttachmentIndex(const VkClearAttachment &clear_attachment) const {
+    const auto &rpci = rp_state_->createInfo;
+    const auto &subpass = rpci.pSubpasses[GetCurrentSubpass()];
     uint32_t attachment_index = VK_ATTACHMENT_UNUSED;
 
     if (clear_attachment.aspectMask & VK_IMAGE_ASPECT_COLOR_BIT) {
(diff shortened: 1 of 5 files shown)


jzulauf-lunarg:
@artem-lunarg @jeremyg-lunarg -- review?

layers/sync/sync_validation.cpp
```diff
@@ -1947,6 +1987,137 @@ void AccessContext::RecordLayoutTransitions(const RENDER_PASS_STATE &rp_state, u
...
+    HazardResult hazard;
+
+    // Need to hazard detect load operations vs. the attachment views
+    const uint32_t attachment_count = static_cast<uint32_t>(info.attachments.size());
+    for (uint32_t i = 0; i < attachment_count; i++) {
+        const auto &attachment = info.attachments[i];
+        const SyncStageAccessIndex load_index = attachment.GetLoadUsage();
+        if (load_index == SYNC_ACCESS_INDEX_NONE) continue;
+
+        hazard = GetCurrentAccessContext()->DetectHazard(attachment.view_gen, load_index, SyncOrdering::kColorAttachment);
```
artem-lunarg, line 2007: `attachment.GetOrdering()` instead of `SyncOrdering::kColorAttachment`?

layers/sync/sync_validation.cpp
```diff
@@ -2289,16 +2460,122 @@ void CommandBufferAccessContext::RecordDrawVertexIndex(const std::optional<uint3
...
+    bool skip = false;
+    const auto lv_bind_point = ConvertToLvlBindPoint(VK_PIPELINE_BIND_POINT_GRAPHICS);
+    const auto &last_bound_state = cb_state_->lastBound[lv_bind_point];
+    const auto *pipe = last_bound_state.pipeline_state;
+    if (!pipe) {
+        return skip;
+    }
+
+    const auto raster_state = pipe->RasterizationState();
+    if (raster_state && raster_state->rasterizerDiscardEnable) {
```
artem-lunarg, line 2483: This expression is also used in other parts of syncval code. Should we also skip if `raster_state` is null, e.g. check `!raster_state || raster_state->rasterizerDiscardEnable`?

layers/sync/sync_validation.cpp
```diff
@@ -1947,6 +1987,137 @@ void AccessContext::RecordLayoutTransitions(const RENDER_PASS_STATE &rp_state, u
...
+                    if (hazard.IsHazard()) {
+                        Location loc = attachment.GetLocation(error_obj.location, i);
+                        skip |= report_resolve_hazard(hazard, loc.dot(vvl::Field::resolveImageView),
+                                                      attachment.resolve_view->Handle(), attachment.info.resolveMode);
+                    }
+                }
+            }
+
+            const auto store_usage = attachment.GetStoreUsage();
+            if (store_usage != SYNC_ACCESS_INDEX_NONE) {
```
artem-lunarg, line 2081: In case it's not noted in other places.
`STORE_OP_NONE` might produce a write that is equivalent to `DONT_CARE` in case render pass actually writes to the attachment.

The suggestion is to comment about limitation or to add todo (both for EndRendering's Validate and Record).

> VK_ATTACHMENT_STORE_OP_NONE specifies the contents within the render area are not accessed by the store operation as long as no values are written to the attachment during the render pass. If values are written during the render pass, this behaves identically to VK_ATTACHMENT_STORE_OP_DONT_CARE and with matching access semantics.

tests/framework/render.h
```diff
@@ -145,6 +145,17 @@ class VkRenderFramework : public VkTestFramework {
         return spv;
     }
 
+    void DeviceWaitIdle() { m_device->wait(); }
```
jzulauf-lunarg, line 148: @juan-lunarg @spencer-lunarg  -- this allowed me to factor a test help without capturing all the framework members individually.  Let me know if you'd like me to do this another way.

layers/sync/sync_validation.cpp
jzulauf-lunarg: I'll look at this as part of the attachment rules revision in the project tasks.

layers/sync/sync_validation.cpp
jzulauf-lunarg: Noted in task "Improve accuracy of attachment access semantics"

layers/sync/sync_validation.cpp
```diff
@@ -1166,20 +1181,37 @@ void AccessContext::ResolvePreviousAccesses() {
     ResolvePreviousAccess(kFullRange, &access_state_map_, &default_state);
 }
 
+static SyncStageAccessIndex GetLoadOpUsageIndex(VkAttachmentLoadOp load_op, syncval_state::AttachmentType type) {
+    SyncStageAccessIndex usage_index;
+    if (load_op == VK_ATTACHMENT_LOAD_OP_NONE_EXT) {
```
artem-lunarg, line 1186: supernit: in our code we can use enum values without extension suffixes if they exists in core (or the suffix from the most promoted extension).

layers/sync/sync_validation.cpp
```diff
@@ -2795,12 +3091,15 @@ static std::optional<uint32_t> GetAttachmentIndex(const RenderPassAccessContext
             attachment_index = subpass.pDepthStencilAttachment->attachment;
         }
     }
-    const bool invalid_index = (attachment_index == VK_ATTACHMENT_UNUSED || attachment_index >= rpci.attachmentCount);
-    return invalid_index ? std::optional<uint32_t>() : attachment_index;
+    if (attachment_index >= rpci.attachmentCount) {
```
artem-lunarg, line 3094: This check is correct, but maybe to add a comment that it also checks for `attachment_index == VK_ATTACHMENT_UNUSED`? Previously it was explicit and now it uses property that `VK_ATTACHMENT_UNUSED` is ~0, but from just from the code itself it's hard to tell if it handles `VK_ATTACHMENT_UNUSED` intentionally or it just a side effect of attempt to filter out-of-range indices.

layers/sync/sync_validation.cpp
```diff
@@ -2826,121 +3125,107 @@ static VkImageAspectFlags GetAspectsToClear(VkImageAspectFlags clear_aspect_mask
     return aspects_to_clear;
 }
 
-static std::optional<VkImageSubresourceRange> RestrictSubresourceRange(const VkImageSubresourceRange &normalized_subresource_range,
-                                                                       const VkClearRect &clear_rect) {
+std::string ClearAttachmentInfo::GetSubpassAttachmentText() const {
+    if (attachment_index == VK_ATTACHMENT_UNUSED) return std::string();
+    std::stringstream text;
+    text << " render pass attachment index " << attachment_index << "in subpass " << subpass;
```
artem-lunarg, line 3131: missing leading space "in subpass"

layers/sync/sync_validation.cpp
```diff
@@ -2826,121 +3125,107 @@ static VkImageAspectFlags GetAspectsToClear(VkImageAspectFlags clear_aspect_mask
...
+VkImageSubresourceRange ClearAttachmentInfo::RestrictSubresourceRange(const VkClearRect &clear_rect, const ImageViewState &view) {
+    const VkImageSubresourceRange &normalized_subresource_range = view.normalized_subresource_range;
+
     assert(normalized_subresource_range.layerCount != VK_REMAINING_ARRAY_LAYERS);  // contract of this function
     assert(clear_rect.layerCount != VK_REMAINING_ARRAY_LAYERS);                    // according to spec
     const uint32_t first = std::max(normalized_subresource_range.baseArrayLayer, clear_rect.baseArrayLayer);
     const uint32_t last_range = normalized_subresource_range.baseArrayLayer + normalized_subresource_range.layerCount;
     const uint32_t last_clear = clear_rect.baseArrayLayer + clear_rect.layerCount;
     const uint32_t last = std::min(last_range, last_clear);
-    std::optional<VkImageSubresourceRange> result;
```
artem-lunarg, line 2837: It looks change of behavior. For situation when `first >= last` due to bad `clear_rect` setup we returned optional and `RenderPassAccessContext::GetClearAttachmentInfo` returned empty result, but now we don't detect this and proceed with construction (line 3164):
`return ClearAttachmentInfo(clear_attachment, rect, *view_state, attachment_index, GetCurrentSubpass());`

Few mins later.

Okay I see we have 
` bool IsValid() const { return view && (aspects_to_clear != 0U) && (subresource_range.layerCount != 0U); }`

which catches this situation. I bit confused by this. Initially we had explicit object representation for invalid setup (empty optional) and did not try to proceed with ClearAttachmentInfo construction. Now we proceed with full initialization of ClearAttachmentInfo with a side effect that `VkImageSubresourceRange` will contain zeros, so we detect that later. Error detection becomes more implicit, adds a point of failure - forget to call IsValid(), harder to analyze: previously ClearAttachmentInfo could be only in a valid state and that gives good property that when I get object I can't have forks in mental model and don't need additional state to proceed further. Now having ClearAttachmentInfo you might need to think a bit, and check the context. I understand it might be not easy to change to use original approach but at least feedback on how I think it changes understanding of the code.

artem-lunarg (approved):
LGTM. Can't find correctness issues. Left feedback about more implicit style of error check.

layers/sync/sync_validation.cpp
jzulauf-lunarg: True, but `VK_ATTACHMENT_LOAD_OP_NONE_EXT` hasn't been promoted AFAICT

layers/sync/sync_validation.cpp
jzulauf-lunarg: Added comment.

layers/sync/sync_validation.cpp
jzulauf-lunarg: added clarifying comments

layers/sync/sync_validation.cpp
artem-lunarg: That's part of our implementation, not how we use vulkan api, based on this I assumed it's fine to use this as soon as we have this in the headers?

Issue #6404, opened 2023-08-31
 tests: PositiveShaderLimits::ComputeSharedMemoryWorkgroupMemoryExplicitLayoutSpec SIGFPE


HildarTheDorf:
**Environment:**
 - OS: Linux (ubuntu 20.04 'focal') using Mesa 23.1.6 (https://launchpad.net/~kisak/+archive/ubuntu/kisak-mesa)
 - GPU: AMD Ryzen™ 7 5800U
 - SDK or header version if building from repo: Build from commit 637dd02e40c368d1e7677ef228bbdf9a4fb92403
 - Options enabled (synchronization, best practices, etc.): N/A

**Describe the Issue**

The test PositiveShaderLimits::ComputeSharedMemoryWorkgroupMemoryExplicitLayoutSpec causes a divide by zero fault on my machine deep in the driver. The underlying error is that the spir-v produced by glslang lays the structure out in such a way it will blow past `maxComputeSharedMemorySize`. Laying out a struct like this using a spec constant will not work as intended as the OpMemberDecorate in the spir-v will be fixed at glsl->spir-v time, and not updated at spirv->bytecode time. This causes the entire test run to end prematurely, not running any further tests.

**Expected behavior**
Test passes, is skipped, or at least fails gracefully

spencer-lunarg:
we use `const uint32_t max_shared_memory_size = m_device->phy().properties().limits.maxComputeSharedMemorySize;` to get the size and go from there, the `glslang` we use `ProcessConfigFile(device_limits);` to set these in the GLSL we generate

Is it possible to get me the `maxComputeSharedMemorySize` on your machine and the `csSource` we generate

HildarTheDorf:
The issue is that the glsl will always specify `X.x1` to be an array of `max_shared_vec_4 + 16` `vec4`s.

For me this means the `OpMemberDecorate` in the generated spirv puts `x2` at an offset of 65792 but `maxComputeSharedMemorySize` is 65536. I think this will always invoke UB the way it is written as `sizeof(X::x1)` will always be greater than `maxComputeSharedMemorySize`, the comment in the source even claims that was intended!

I don't believe the underlying bug is environment specific, it's just some drivers are more polite with their handling of the UB.

As requested though:

`maxComputeSharedMemorySize  = 65536`
```        #version 450
        #extension GL_EXT_shared_memory_block : enable

        // will be over the max if the spec constant uses default value
[...]

PR #6863, opened 2023-10-30
layers: Use Struct Size for workgroup shared memory

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6404

The issue was the `PositiveShaderLimits.ComputeSharedMemoryWorkgroupMemoryExplicitLayoutSpec` really had an un-caught edge case of `VUID-RuntimeSpirv-Workgroup-06530`

When the GLSL looked like

```glsl
struct X {
    vec4 a;
    layout(offset = max_limit) vec4 b;
}
```

we were not using the `offset` to detect the limit

This was currently only done for Push Constants because the logic to check for this would be wasteful every time we see an `OpTypeStruct` so created a new `TypeStructSize` to have both cases use the logic


layers/state_tracker/shader_module.cpp
```diff
@@ -1923,17 +1927,52 @@ PushConstantVariable::PushConstantVariable(const SPIRV_MODULE_STATE& module_stat
...
 
+    auto struct_size = type_struct_info->GetSize(module_state);
+    offset = struct_size.offset;
+    size = struct_size.size;
+}
+
+TypeStructInfo::TypeStructInfo(const SPIRV_MODULE_STATE& module_state, const Instruction& struct_insn)
+    : id(struct_insn.Word(1)), length(struct_insn.Length() - 2), decorations(module_state.GetDecorationSet(id)) {
+    members.resize(length);
+    for (uint32_t i = 0; i < length; i++) {
```
arno-lunarg, line 1938: Use `vvl::enumerate`?

layers/state_tracker/shader_module.cpp
```diff
@@ -1923,17 +1927,52 @@ PushConstantVariable::PushConstantVariable(const SPIRV_MODULE_STATE& module_stat
...
     //    OpMemberDecorate %x 1 Offset 0
     //    OpMemberDecorate %x 0 Offset 4
+    //
+    // Info at https://gitlab.khronos.org/spirv/SPIR-V/-/issues/763
     uint32_t highest_element_index = 0;
     uint32_t highest_element_offset = 0;
-    for (uint32_t i = 0; i < type_struct_info->members.size(); i++) {
-        const auto& member = type_struct_info->members[i];
+
+    for (uint32_t i = 0; i < members.size(); i++) {
```
arno-lunarg, line 1974: Here to

arno-lunarg (approved):

layers/state_tracker/shader_module.cpp
spencer-lunarg: ya, can add

spencer-lunarg: I realize there is no need for `vvl::enumerate` this is a `std::vector`, it has it's own for-range helper

PR #6867, opened 2023-10-30
layers: Validate conservative rasterization with points and lines

ziga-lunarg:
Part of https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5507

### DIFF
--- a/layers/core_checks/cc_cmd_buffer_dynamic.cpp
+++ b/layers/core_checks/cc_cmd_buffer_dynamic.cpp
@@ -528,6 +528,29 @@ bool CoreChecks::ValidateDrawDynamicStatePipeline(const LAST_BOUND_STATE& last_b
         }
     }
 
+    if (pipeline.IsDynamic(VK_DYNAMIC_STATE_CONSERVATIVE_RASTERIZATION_MODE_EXT) &&
+        !phys_dev_ext_props.conservative_rasterization_props.conservativePointAndLineRasterization &&
+        (!pipeline.IsDynamic(VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY) ||
+         cb_state.dynamic_state_status.cb[CB_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY])) {
+        const VkPrimitiveTopology topology = pipeline.IsDynamic(VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY)
+                                                 ? cb_state.dynamic_state_value.primitive_topology
+                                                 : pipeline.topology_at_rasterizer;
+        if (IsValueIn(topology,
+                      {VK_PRIMITIVE_TOPOLOGY_POINT_LIST, VK_PRIMITIVE_TOPOLOGY_LINE_LIST, VK_PRIMITIVE_TOPOLOGY_LINE_STRIP,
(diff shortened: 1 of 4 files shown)


layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -528,6 +528,30 @@ bool CoreChecks::ValidateDrawDynamicStatePipeline(const LAST_BOUND_STATE& last_b
         }
     }
 
+    if (pipeline.IsDynamic(VK_DYNAMIC_STATE_CONSERVATIVE_RASTERIZATION_MODE_EXT) &&
+        !phys_dev_ext_props.conservative_rasterization_props.conservativePointAndLineRasterization &&
+        (!pipeline.IsDynamic(VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY) ||
```
spencer-lunarg, line 533: why would you need `(!pipeline.IsDynamic(VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY) || cb_state.dynamic_state_status.cb[CB_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY])`?
 
Is this in case they have it dynamic and break `07842`?

layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -528,6 +528,30 @@ bool CoreChecks::ValidateDrawDynamicStatePipeline(const LAST_BOUND_STATE& last_b
...
+         cb_state.dynamic_state_status.cb[CB_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY])) {
+        const VkPrimitiveTopology topology = pipeline.IsDynamic(VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY)
+                                                 ? cb_state.dynamic_state_value.primitive_topology
+                                                 : pipeline.topology_at_rasterizer;
+        if (IsValueIn(topology,
+                      {VK_PRIMITIVE_TOPOLOGY_POINT_LIST, VK_PRIMITIVE_TOPOLOGY_LINE_LIST, VK_PRIMITIVE_TOPOLOGY_LINE_STRIP,
+                       VK_PRIMITIVE_TOPOLOGY_LINE_LIST_WITH_ADJACENCY, VK_PRIMITIVE_TOPOLOGY_LINE_STRIP_WITH_ADJACENCY})) {
+            if (cb_state.dynamic_state_status.cb[CB_DYNAMIC_STATE_CONSERVATIVE_RASTERIZATION_MODE_EXT] &&
```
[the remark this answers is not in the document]
ziga-lunarg: Used `VK_DYNAMIC_STATE_CONSERVATIVE_RASTERIZATION_MODE_EXT` instead of `VK_CONSERVATIVE_RASTERIZATION_MODE_DISABLED_EXT`

layers/core_checks/cc_cmd_buffer_dynamic.cpp
ziga-lunarg: I learned from writing shader objects VUs that I should always do the check that the dynamic state was set, otherwise the log messages will print unset variable and it wont make any sense

ziga-lunarg: Without the check I also ran into situations where tests would be invalid, but would pass on debug configurations and failing only in release, but that was hard to track down

spencer-lunarg (approved):

PR #6870, opened 2023-10-30
layers: Validate sample locations enable

ziga-lunarg:
Last of #5507

### DIFF
--- a/layers/core_checks/cc_cmd_buffer_dynamic.cpp
+++ b/layers/core_checks/cc_cmd_buffer_dynamic.cpp
@@ -259,27 +259,36 @@ bool CoreChecks::ValidateDrawDynamicState(const LAST_BOUND_STATE& last_bound_sta
         }
     }
 
-    std::shared_ptr<const SPIRV_MODULE_STATE> spirv_state;
-    std::shared_ptr<const EntryPoint> entrypoint;
+    std::shared_ptr<const SPIRV_MODULE_STATE> vert_spirv_state;
+    std::shared_ptr<const EntryPoint> vert_entrypoint;
+    std::shared_ptr<const SPIRV_MODULE_STATE> frag_spirv_state;
     if (last_bound_state.pipeline_state) {
-        for (const auto &stage_state : last_bound_state.pipeline_state->stage_states) {
+        for (const auto& stage_state : last_bound_state.pipeline_state->stage_states) {
             if (stage_state.GetStage() == VK_SHADER_STAGE_VERTEX_BIT) {
-                spirv_state = stage_state.spirv_state;
-                entrypoint = stage_state.entrypoint;
+                vert_spirv_state = stage_state.spirv_state;
+                vert_entrypoint = stage_state.entrypoint;
+            }
+            if (stage_state.GetStage() == VK_SHADER_STAGE_FRAGMENT_BIT) {
+                frag_spirv_state = stage_state.spirv_state;
             }
         }
     } else {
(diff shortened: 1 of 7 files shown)


layers/state_tracker/shader_module.cpp
```diff
@@ -968,6 +968,13 @@ SPIRV_MODULE_STATE::StaticData::StaticData(const SPIRV_MODULE_STATE& module_stat
                 cooperative_matrix_inst.push_back(&insn);
                 break;
             }
+            case spv::OpExtInst: {
+                constexpr const uint32_t interpolate_at_sample = 77;
```
ziga-lunarg, line 972: I could not find this anywhere in the spirv spec

layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -261,19 +261,27 @@ bool CoreChecks::ValidateDrawDynamicState(const LAST_BOUND_STATE& last_bound_sta
 
     std::shared_ptr<const SPIRV_MODULE_STATE> spirv_state;
```
spencer-lunarg, line 262: can we call this `vert_spirv_state` then otherwise it gets confusing

layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -330,6 +338,66 @@ bool CoreChecks::ValidateDrawDynamicState(const LAST_BOUND_STATE& last_bound_sta
             }
         }
     }
+
+    if ((pipeline_state && pipeline_state->IsDynamic(VK_DYNAMIC_STATE_SAMPLE_LOCATIONS_ENABLE_EXT)) ||
+        last_bound_state.IsValidShaderBound(ShaderObjectStage::FRAGMENT)) {
```
spencer-lunarg, line 343: nit: 

we go `bool vertex_shader_bound = last_bound_state.IsValidShaderBound(ShaderObjectStage::VERTEX);` above

can you make these both be the same style (don't care which, but want to have be consistent)

layers/state_tracker/shader_module.cpp
spencer-lunarg: it is in `GLSLstd450InterpolateAtSample` inside `GLSL.std.450.h`

https://registry.khronos.org/SPIR-V/specs/unified1/GLSL.std.450.html

https://github.com/KhronosGroup/SPIRV-Guide/blob/master/chapters/extended_instruction_sets.md

tests/unit/dynamic_state.cpp
```diff
@@ -5543,3 +5543,211 @@ TEST_F(NegativeDynamicState, InvalidConservativeRasterizationMode) {
     m_commandBuffer->EndRenderPass();
     m_commandBuffer->end();
 }
+
+TEST_F(NegativeDynamicState, DynamicSampleLocationsEnable) {
```
spencer-lunarg, line 5547: nice to have 1 positive test, I have found the Sample Location to be hard to debug and nice if we have a positive test to look at

layers/state_tracker/shader_module.cpp
spencer-lunarg:
```suggestion
#include "glslang/SPIRV/GLSL.std.450.h"

if (insn.Word(4) == GLSLstd450InterpolateAtSample)
```

spencer-lunarg (approved):
Looks good now!

Please just add `GLSLstd450InterpolateAtSample` instead of a hardcoded `77` and good to go

PR #6871, opened 2023-10-30
layers: Fix tessellation validation with libraries

### LINKED ISSUE #4830: Unexpectedly receiving VUID-VkGraphicsPipelineCreateInfo-pStages-00731
**Describe the Issue**
I'm creating a new set of CTS tests that try to verify implementations correctly ignore pStages at graphics pipeline creation time when they are not needed. This is in the spirit of the current VkGraphicsPipelineCreateInfo VUs, which have some requirements about situations in which pStages has to meet certain conditions, but does not make it mandatory to be null when they are not used.

For example, when creating a graphics pipeline library with only the vertex input state, pStages should be ignored. However, if I try to do exactly that and include a tessellation shader in pStages, that should be ignored, I hit VUID-VkGraphicsPipelineCreateInfo-pStages-00731, which reads:

> If the pipeline is being created with [pre-rasterization shader state](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#pipelines-graphics-subsets-pre-rasterization) and pStages includes a tessellation control shader stage and a tessellation evaluation shader stage, pTessellationState must be a valid pointer to a valid [VkPipelineTessellationStateCreateInfo](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkPipelineTessellationStateCreateInfo) structure

Which shouldn't apply to my case as I'm not creating a pipeline with pre-rasterization shader state.

**Valid Usage ID**
VUID-VkGraphicsPipelineCreateInfo-pStages-00731

**Environment:**
 - OS: Linux
 - GPU: NVIDIA.
 - SDK or header version if building from repo: 0f839617e939b5cf2c406fb00b1ed4533f695d71
 - Options enabled (synchronization, best practices, etc.): Default ones plus synchronization and debug printf.

**Additional context**

ziga-lunarg:
Closes #4830

### DIFF
--- a/layers/core_checks/cc_pipeline_graphics.cpp
+++ b/layers/core_checks/cc_pipeline_graphics.cpp
@@ -77,6 +77,7 @@ bool CoreChecks::ValidateGraphicsPipeline(const PIPELINE_STATE &pipeline, const 
     skip |= ValidateGraphicsPipelineLibrary(pipeline, create_info_loc);
     skip |= ValidateGraphicsPipelinePreRasterState(pipeline, create_info_loc);
     skip |= ValidateGraphicsPipelineInputAssemblyState(pipeline, create_info_loc);
+    skip |= ValidateGraphicsPipelineTessellationState(pipeline, create_info_loc);
     skip |= ValidateGraphicsPipelineColorBlendState(pipeline, subpass_desc, create_info_loc);
     skip |= ValidateGraphicsPipelineRasterizationState(pipeline, subpass_desc, create_info_loc);
     skip |= ValidateGraphicsPipelineMultisampleState(pipeline, subpass_desc, create_info_loc);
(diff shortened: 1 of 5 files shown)


layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -963,6 +964,23 @@ bool CoreChecks::ValidateGraphicsPipelineInputAssemblyState(const PIPELINE_STATE
     return skip;
 }
 
+bool CoreChecks::ValidateGraphicsPipelineTessellationState(const PIPELINE_STATE &pipeline, const Location &create_info_loc) const {
+    bool skip = false;
+
+    if (pipeline.OwnsSubState(pipeline.pre_raster_state) &&
+        (pipeline.create_info_shaders & VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT) &&
+        (pipeline.create_info_shaders & VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT)) {
```
spencer-lunarg, line 972: is it valid to only have 1 tessellation stage?

spencer-lunarg (approved):

layers/core_checks/cc_pipeline_graphics.cpp
ziga-lunarg: No

spencer-lunarg: good, making sure

PR #6872, opened 2023-10-30
Syncval: Perf:  Additional (up to 25%) performance gains for certain traces

jzulauf-lunarg:
Targeted optimizations that help some traces, but don't hurt any.

1)  Use range/iterator combined walk to avoid pre-range lower_bound calls
2)  DetectHazard with range_gen/iterator walk

Benefits traces where resource access state ranges well match range generated one

### DIFF
--- a/layers/containers/range_vector.h
+++ b/layers/containers/range_vector.h
@@ -1635,29 +1635,45 @@ Iterator split(Iterator in, Map &map, const Range &range) {
 }
 
 // Apply an operation over a range map, infilling where content is absent, updating where content is present.
+// The passed pos must *either* be strictly less than range or *is* lower_bound (which may be end)
 // Trims to range boundaries.
 // infill op doesn't have to alter map, but mustn't invalidate iterators passed to it. (i.e. no erasure)
 // infill data (default mapped value or other initial value) is contained with ops.
 // update allows existing ranges to be updated (merged, whatever) based on data contained in ops.  All iterators
 // passed to update are already trimmed to fit within range.
-template <typename RangeMap, typename InfillUpdateOps>
-void infill_update_range(RangeMap &map, const typename RangeMap::key_type &range, const InfillUpdateOps &ops) {
+template <typename RangeMap, typename InfillUpdateOps, typename Iterator = typename RangeMap::iterator>
+Iterator infill_update_range(RangeMap &map, Iterator pos, const typename RangeMap::key_type &range, const InfillUpdateOps &ops) {
     using KeyType = typename RangeMap::key_type;
     using IndexType = typename RangeMap::index_type;
-    const auto map_end = map.end();
 
-    if (range.empty()) return;
+    const auto end = map.end();
+    assert((pos == end) || (pos == map.lower_bound(range)) || pos->first.strictly_less(range));
 
-    // Starting with
-    auto pos = map.lower_bound(range);
-    if ((pos != map_end) && (range.begin > pos->first.begin)) {
+    if (range.empty()) return pos;
+    if (pos == end) {
+        // Only pass pos == end for range tail after last entry
+        assert(end == map.lower_bound(range));
+    } else if (pos->first.strictly_less(range)) {
+        // pos isn't lower_bound for range (it's less than range), however, if range is monotonically increasing it's likely
+        // the next entry in the map will be the lower bound.
+
+        // If the new (pos + 1) *isn't* stricly_less and pos is,
+        // (pos + 1) must be the lower_bound, otherwise we have to look for it O(log n)
+        ++pos;
+        if (pos->first.strictly_less(range)) {
+            pos = map.lower_bound(range);
+        }
+        assert(pos == map.lower_bound(range));
+    }
+
+    if ((pos != end) && (range.begin > pos->first.begin)) {
         // lower bound starts before the range, trim and advance
         pos = map.split(pos, range.begin, sparse_container::split_op_keep_both());
         ++pos;
     }
 
     IndexType current_begin = range.begin;
-    while ((pos != map_end) && (current_begin < range.end)) {
+    while ((pos != end) && (current_begin < range.end)) {
         // The current_begin is either pointing to the next existing value to update or the beginning of a gap to infill
         assert(pos->first.begin >= current_begin);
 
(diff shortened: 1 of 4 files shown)


spencer-lunarg (approved):
LGTM from what I can see

layers/containers/range_vector.h
```diff
@@ -1640,24 +1640,33 @@ Iterator split(Iterator in, Map &map, const Range &range) {
...
+    if (pos == end) {
+        // Only pass pos == end for range tail after last entry
+        assert(end == map.lower_bound(range));
+    } else if (pos->first.strictly_less(range)) {
+        // Try the next entry in the map
+        ++pos;
+        if (pos->first.strictly_less(range)) {
+            pos = map.lower_bound(range);
+        }
+        assert(pos == map.lower_bound(range));
```
artem-lunarg, line 1659: nit: it's possible to have this assert shared for both if branches and then convert "if" to a version without "else", but I also can see that having explicit "if" for a special case can be a form of documentation.

layers/containers/range_vector.h
```diff
@@ -1640,24 +1640,33 @@ Iterator split(Iterator in, Map &map, const Range &range) {
 // infill data (default mapped value or other initial value) is contained with ops.
 // update allows existing ranges to be updated (merged, whatever) based on data contained in ops.  All iterators
 // passed to update are already trimmed to fit within range.
-template <typename RangeMap, typename InfillUpdateOps>
-void infill_update_range(RangeMap &map, const typename RangeMap::key_type &range, const InfillUpdateOps &ops) {
+template <typename RangeMap, typename InfillUpdateOps, typename Iterator = typename RangeMap::iterator>
+Iterator infill_update_range(RangeMap &map, Iterator pos, const Iterator &end, const typename RangeMap::key_type &range,
```
artem-lunarg, line 1644: Suggestion to simplify signature by removing `end` parameter. This also is not optimization for calling map.end() in the loop.

layers/containers/range_vector.h
```diff
@@ -1640,24 +1640,33 @@ Iterator split(Iterator in, Map &map, const Range &range) {
...
 
-    if (range.empty()) return;
+    if (range.empty()) return pos;
+    if (pos == end) {
+        // Only pass pos == end for range tail after last entry
+        assert(end == map.lower_bound(range));
+    } else if (pos->first.strictly_less(range)) {
+        // Try the next entry in the map
+        ++pos;
+        if (pos->first.strictly_less(range)) {
```
artem-lunarg, line 1656: If I understood this correctly, for typical range generators there won't be a gap and this condition usually will be false, and we have gains. Maybe to comment about this to make more obvious?

layers/sync/sync_validation.cpp
```diff
@@ -845,34 +858,103 @@ void AccessContext::EraseIf(Predicate &&pred) {
...
+
+        // Action is allowed to alter pos but shouldn't do so if range is strictly < pos->first
+        if (action(range, end, pos)) return true;
+    }
+
+    // Action needs to handle the "at end " condition (and can be useful for recursive actions)
     for (; range_gen->non_empty(); ++range_gen) {
-        HazardResult hazard = DetectHazard(detector, *range_gen, options);
-        if (hazard.IsHazard()) return hazard;
+        if (action(*range_gen, end, pos)) return true;
```
artem-lunarg, line 916: nit: suggestion to assert(end == pos) to make it obvious

layers/containers/range_vector.h
artem-lunarg, line 1644: One option is to have nice local variable `end` instead of parameter, templated functions are already cryptic. Not sure if that's one of those 3% of cases for optimization.

layers/sync/sync_validation.cpp
```diff
@@ -845,34 +858,103 @@ void AccessContext::EraseIf(Predicate &&pred) {
...
+// Action is const w.r.t. map
+// Action is allowed (assumed) to modify pos
+// Action must not advance pos for ranges strictly < pos->first
+// Action must handle range strictly less than pos->first correctly
+// Action must handle pos == end correctly
+// Action is assumed to only require invocation once per map entry
+// Note: If Action invocations are heavyweight and inter-entry (gap) calls are not needed
+//       add a template or function parameter to skip them. TBD.
+template <typename RangeMap, typename RangeGen, typename Action>
+bool ConstForAnyRangeGen(const RangeMap &map, RangeGen &range_gen, Action &action) {
```
artem-lunarg, line 876: Is "Const" in the name stands for "constant"? The name does not say what function does. Suggestion to use a vert to describe what function does (e.g. apply).

layers/sync/sync_validation.cpp
```diff
@@ -845,34 +858,103 @@ void AccessContext::EraseIf(Predicate &&pred) {
...
+            if (pos->first.strictly_less(range)) {
+                pos = map.lower_bound(range);
+                if (pos == end) break;
+            }
+        }
+        assert(pos == map.lower_bound(range));
+
+        // If the range intersects pos->first, consider Action performed for that map entry, and
+        // make sure not to call Action for this pos for any subsequent ranges
+        skip_limit = range.end > pos->first.begin ? pos->first.end : 0U;
```
artem-lunarg, line 908: Just to check if I understood correctly. Is it also a correct implementation to never reset `skip_limit` to 0 and only increase if necessary (monotonically increasing function)?

artem-lunarg (approved):
LGTM. Few questions and some nits but looks good.

layers/containers/range_vector.h
jzulauf-lunarg: iterator begin/end ranges pairs is idiomatic for STL style traversals, so I'll leave as is.  There is the overload that takes the map and invokes this as begin(), end() for those wanting to just traverse the whole thing

layers/sync/sync_validation.cpp
jzulauf-lunarg: Added comment that range generation must be monotonically increasing.

layers/sync/sync_validation.cpp
jzulauf-lunarg: Renamed

layers/sync/sync_validation.cpp
jzulauf-lunarg: If range_gen is non-empty we *must* be at end (given the for loop we exited).  If it's empty, pos can be anything.

layers/containers/range_vector.h
artem-lunarg: The idea of idiomatic begin/end is too separate algorithm from containers so the algorithm can accept only iterators or ranges. But here we also pass a container, so when reading the code I don’t detect such signature as idiomatic. I also spent some time scanning code to find use cases when “end” is not map.end() since that’s what you assume when see such signature, and did not find it - the point that it takes more time to understand code.

jzulauf-lunarg: You're correct.  The one thing this function won't do is function for an arbitrary container's iterators.  I was mentally struggling with correctness/predictabilty in the case end != map.end()... 

Cleaned up.

layers/containers/range_vector.h
jzulauf-lunarg: added

PR #3923, opened 2022-03-28
layers: Track and warn using unset push constants

ziga-lunarg:
Closes #3518

PR #6873, opened 2023-10-30
Add best practice warning for not setting all push constant bytes

### LINKED ISSUE #3518: Validate bound push constant size.

ziga-lunarg:
Closes #3518

Most of this was taken from #3923

### DIFF
--- a/layers/best_practices/bp_render_pass.cpp
+++ b/layers/best_practices/bp_render_pass.cpp
@@ -354,6 +354,7 @@ void BestPractices::PreCallRecordCmdBeginRenderingKHR(VkCommandBuffer commandBuf
 void BestPractices::PostCallRecordCmdNextSubpass(VkCommandBuffer commandBuffer, VkSubpassContents contents,
                                                  const RecordObject& record_obj) {
     ValidationStateTracker::PostCallRecordCmdNextSubpass(commandBuffer, contents, record_obj);
+    RecordCmdNextSubpass(commandBuffer);
 
     auto cmd_state = GetWrite<bp_state::CommandBuffer>(commandBuffer);
     auto rp = cmd_state->activeRenderPass.get();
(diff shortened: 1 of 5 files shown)


layers/best_practices/bp_render_pass.cpp
```diff
@@ -606,6 +613,55 @@ bool BestPractices::PreCallValidateCmdBeginRenderingKHR(VkCommandBuffer commandB
     return PreCallValidateCmdBeginRendering(commandBuffer, pRenderingInfo, error_obj);
 }
 
+void BestPractices::PostCallRecordCmdNextSubpass2KHR(VkCommandBuffer commandBuffer, const VkSubpassBeginInfo* pSubpassBeginInfo,
+                                                     const VkSubpassEndInfo* pSubpassEndInfo, const RecordObject& record_obj) {
+    StateTracker::PostCallRecordCmdNextSubpass2KHR(commandBuffer, pSubpassBeginInfo, pSubpassEndInfo, record_obj);
```
spencer-lunarg, line 618: just call `PostCallRecordCmdNextSubpass2(commandBuffer, pSubpassBeginInfo, pSubpassEndInfo, record_obj)`

we are removing the need to duplicate promoted extension entry points

spencer-lunarg (approved):
LGTM - with nit

PR #6874, opened 2023-10-31
testing: Use known_good.json for all testing dependencies

juan-lunarg:
Currently testing just grabs TOT main for Vulkan-Loader, Vulkan-Profile, and Vulkan-Tools.

This isn't great since it makes our testing unrepeatable.

Use known_good.json instead

### DIFF
--- a/scripts/known_good.json
+++ b/scripts/known_good.json
@@ -1,20 +1,6 @@
 {
     "repos": [
         {
-            "name": "glslang",
-            "url": "https://github.com/KhronosGroup/glslang.git",
-            "sub_dir": "glslang",
-            "build_dir": "glslang/build",
-            "install_dir": "glslang/build/install",
-            "commit": "be564292f00c5bf0d7251c11f1c9618eb1117762",
-            "cmake_options": [
-                "-DENABLE_OPT=OFF"
-            ],
-            "optional": [
-                "tests"
-            ]
-        },
-        {
             "name": "Vulkan-Headers",
             "api": "vulkan",
             "url": "https://github.com/KhronosGroup/Vulkan-Headers.git",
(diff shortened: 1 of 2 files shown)


scripts/known_good.json
```diff
@@ -87,20 +90,53 @@
             ]
         },
         {
-            "name": "mimalloc",
-            "url": "https://github.com/microsoft/mimalloc.git",
-            "sub_dir": "mimalloc",
-            "build_dir": "mimalloc/build",
-            "install_dir": "mimalloc/build/install",
+            "name": "glslang",
```
juan-lunarg, line 93: I moved all test dependencies to the bottom.

scripts/known_good.json
```diff
@@ -87,20 +90,53 @@
...
+            ]
+        },
+        {
+            "name": "Vulkan-Loader",
+            "url": "https://github.com/KhronosGroup/Vulkan-Loader.git",
+            "sub_dir": "Vulkan-Loader",
+            "build_dir": "Vulkan-Loader/build",
+            "install_dir": "Vulkan-Loader/build/install",
+            "commit": "v1.3.269",
+            "build_step": "skip",
```
juan-lunarg, line 113: For now we are skipping the build_step on these dependencies.

I'm taking this one step at a time and reproducible testing is good enough for 1 PR.

scripts/known_good.json
```diff
@@ -87,20 +90,53 @@
...
+                "tests"
+            ]
+        },
+        {
+            "name": "Vulkan-Tools",
+            "url": "https://github.com/KhronosGroup/Vulkan-Tools.git",
+            "sub_dir": "Vulkan-Tools",
+            "build_dir": "Vulkan-Tools/build",
+            "install_dir": "Vulkan-Tools/build/install",
+            "commit": "2d956672d73321bfb22b378c06033f0bd885d61c",
```
juan-lunarg, line 136: NOTE: This will mean that integrating new features isn't automatic. But it also means VVL CI won't randomly break due to a regression from tools, loader, etc.

spencer-lunarg: honestly that is ok, we only need to update these normally with new header release or extension support was added

spencer-lunarg (approved):

PR #6875, opened 2023-10-31
tests: Add Positive Parent test

spencer-lunarg:
Adds test making sure imageless framebuffer is ignored in generated Object Tracker

### DIFF
--- a/tests/unit/parent_positive.cpp
+++ b/tests/unit/parent_positive.cpp
@@ -1,0 +1,61 @@
+/*
+ * Copyright (c) 2023 Valve Corporation
+ * Copyright (c) 2023 LunarG, Inc.
+ *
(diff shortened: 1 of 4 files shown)


tests/unit/parent_positive.cpp
```diff
@@ -0,0 +1,61 @@
...
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+#include "../framework/layer_validation_tests.h"
+
+ParentTest::~ParentTest() {
+    if (m_second_device) {
```
arno-lunarg, line 15: Nit: if is not needed, calling `delete` on `nullptr` is ok

arno-lunarg (approved):

PR #6878, opened 2023-10-31
scripts: Add remove duplicates from stats script

spencer-lunarg:
adds a `-remove_duplicates` to stats script to remove VUs with same number VUID

so now `VUID-vkCmdDraw-magFilter-04553` is the only thing that comes up for `04553` and removes the other 28 current VUs with same number


mikes-lunarg (approved):
Interesting info!

arno-lunarg (approved):

PR #6883, opened 2023-10-31
gpu: Add descriptor hit tracking

jeremyg-lunarg:
Add a feedback mechanism to the instrumentation shader to allow it to report back which descriptor indices are dynamically used by a shader invocation. Then use this information to only run descriptor on validation on the used descriptors.


layers/gpu_shaders/inst_functions.comp
```diff
@@ -193,6 +198,12 @@ bool inst_bindless_check_desc(const uint shader_id, const uint inst_num, const u
                 break;
             }
         }
+        DescriptorSetOutData out_data = inst_bindless_state_buffer.desc_sets[desc_set].out_data;
+        uvec2 out_vec = uvec2(out_data);
+        if ((out_vec.x == 0u) && (out_vec.y == 0u)) {
```
spencer-lunarg, line 203: Trying to understand what is going on here, so we take a spot in the `uint[]` check first 2 bytes are `0` then set the `state_index` value to `1`?

layers/gpu_validation/gpu_descriptor_set.cpp
```diff
@@ -60,7 +61,12 @@ struct DescriptorState {
 };
 
 } // namespace gpuav_glsl
-
+  //
```
spencer-lunarg, line 64: super nit
```suggestion

```

layers/gpu_validation/gpu_descriptor_set.cpp
```diff
@@ -370,6 +376,163 @@ std::shared_ptr<gpuav_state::DescriptorSet::State> gpuav_state::DescriptorSet::G
...
+    next_state->version = cur_version;
+    next_state->allocator = gv_dev->vmaAllocator;
+
+    uint32_t descriptor_count = 0;  // Number of descriptors, including all array elements
+    for (const auto &binding : *this) {
+        // Shader instrumentation is tracking inline uniform blocks as scalers. Don't try to validate inline uniform
+        // blocks
+        if (binding->type == VK_DESCRIPTOR_TYPE_INLINE_UNIFORM_BLOCK_EXT) {
+            descriptor_count++;
+            // LogWarning(device, "UNASSIGNED-GPU-Assisted Validation Warning",
```
spencer-lunarg, line 397: remove?

layers/gpu_validation/gpu_descriptor_set.cpp
```diff
@@ -370,6 +376,163 @@ std::shared_ptr<gpuav_state::DescriptorSet::State> gpuav_state::DescriptorSet::G
...
+    result = vmaInvalidateAllocation(allocator, allocation, 0, VK_WHOLE_SIZE);
+
+    uint32_t max_binding = layout_data[0].count;
+    for (uint32_t binding = 0; binding < max_binding; binding++) {
+        uint32_t count = layout_data[binding + 1].count;
+        uint32_t start = layout_data[binding + 1].state_start;
+        for (uint32_t i = 0; i < count; i++) {
+            uint32_t pos = start + i;
+            if (data[pos]) {
+#if 0
```
spencer-lunarg, line 519: remove?

layers/gpu_validation/gpu_image_layout.cpp
```diff
@@ -0,0 +1,1648 @@
...
+        // Make sure that all image slices are updated to correct layout
+        for (uint32_t i = 0; i < regionCount; ++i) {
+            cb_state_ptr->SetImageInitialLayout(*src_image_state, pRegions[i].srcSubresource, srcImageLayout);
+            cb_state_ptr->SetImageInitialLayout(*dst_image_state, pRegions[i].dstSubresource, dstImageLayout);
+        }
+    }
+}
+
+void GpuAssisted::PreCallRecordCmdCopyImage2KHR(VkCommandBuffer commandBuffer, const VkCopyImageInfo2KHR *pCopyImageInfo, const RecordObject &record_obj) {
+    auto cb_state_ptr = GetWrite<CMD_BUFFER_STATE>(commandBuffer);
```
spencer-lunarg, line 260: I just got done with a change for this, just go 

```cpp
void GpuAssisted::PreCallRecordCmdCopyImage2KHR(VkCommandBuffer commandBuffer, const VkCopyImageInfo2KHR *pCopyImageInfo, const RecordObject &record_obj) { 
    PreCallRecordCmdCopyImage2(commandBuffer, pCopyImageInfo, record_obj);
}
```

for this and all the promoted functions below, we now have `record_obj.location` to know if it is KHR or not, no need for any chassis call (except the stateless that checks extensions requirements) to do this copy-and-paste dance

layers/gpu_validation/gpu_image_layout.cpp
```diff
@@ -0,0 +1,1648 @@
...
+            skip = ValidateDescriptors(context, binding_info, static_cast<const cvdescriptorset::SamplerBinding &>(binding), indices);
+            break;
+        case DescriptorClass::TexelBuffer:
+            skip = ValidateDescriptors(context, binding_info, static_cast<const cvdescriptorset::TexelBinding &>(binding), indices);
+            break;
+        case DescriptorClass::AccelerationStructure:
+            skip = ValidateDescriptors(context, binding_info,
+                                       static_cast<const cvdescriptorset::AccelerationStructureBinding &>(binding), indices);
+            break;
+#if 0
```
spencer-lunarg, line 656: note or remove?

layers/gpu_validation/gpu_image_layout.cpp
```diff
@@ -0,0 +1,1648 @@
...
+        if (binding_info.second.variable->is_written_to &&
+            ValidateUnprotectedBuffer(context.cb_state, *buffer_node, context.loc, context.vuids.protected_command_buffer_02712,
+                                      "Buffer is in a descriptorSet")) {
+            return true;
+        }
+    }
+    return false;
+}
+
+bool GpuAssisted::ValidateDescriptor(const DescriptorContext &context, const DescriptorBindingInfo &binding_info, uint32_t index,
```
spencer-lunarg, line 713: so what is the long term goal here, I assume we don't want 2 copies of `ValidateDescriptor` for here and `cc_descriptor.cpp` right?

jeremyg-lunarg: I don't want 2 copies of this stuff either but I've yet to come up with a good way to avoid it.

layers/gpu_shaders/inst_functions.comp
jeremyg-lunarg: Nope. The if statement is a check for a null buffer reference. Then we just set the state_index value to 1 if there is a valid output buffer.

layers/core_checks/cc_descriptor.cpp
```diff
@@ -688,13 +688,9 @@ bool CoreChecks::PreCallValidateCreateDescriptorSetLayout(VkDevice device, const
 bool CoreChecks::ValidateDrawState(const DescriptorSet &descriptor_set, const BindingVariableMap &bindings,
                                    const std::vector<uint32_t> &dynamic_offsets, const CMD_BUFFER_STATE &cb_state,
                                    const Location &loc, const DrawDispatchVuid &vuids) const {
-    std::optional<vvl::unordered_map<VkImageView, VkImageLayout>> checked_layouts;
-    if (descriptor_set.GetTotalDescriptorCount() > cvdescriptorset::PrefilterBindRequestMap::kManyDescriptors_) {
-        checked_layouts.emplace();
-    }
     bool result = false;
     VkFramebuffer framebuffer = cb_state.activeFramebuffer ? cb_state.activeFramebuffer->framebuffer() : VK_NULL_HANDLE;
-    DescriptorContext context{loc, vuids, cb_state, descriptor_set, framebuffer, true, checked_layouts};
```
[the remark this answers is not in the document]
jeremyg-lunarg: yes.

Tony-LunarG (approved):
Nice rework to share descriptor validation between core and GPU-AV, even with a few copies. I worry a little about the new GPU-AV paths are sparsely tested, but the code it was copied from has been fine.  Nice Job!

tests/unit/gpu_av.cpp
```diff
@@ -2961,3 +2960,926 @@ TEST_F(VkGpuAssistedLayerTest, SelectInstrumentedShaders) {
...
     m_errorMonitor->VerifyFound();
 }
+
+TEST_F(VkGpuAssistedLayerTest, ImageArrayDynamicIndexing) {
+    TEST_DESCRIPTION("GPU validation: test that only dynamically used indices are validated");
+
+    SetTargetApiVersion(VK_API_VERSION_1_1);
+    AddRequiredExtensions(VK_KHR_MAINTENANCE_4_EXTENSION_NAME);
+    AddRequiredExtensions(VK_EXT_DESCRIPTOR_INDEXING_EXTENSION_NAME);
+    AddRequiredExtensions(VK_KHR_TIMELINE_SEMAPHORE_EXTENSION_NAME);
```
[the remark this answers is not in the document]
jeremyg-lunarg: we shouldn't

layers/gpu_validation/gpu_image_layout.cpp
```diff
@@ -0,0 +1,732 @@
...
+                vku::FindStructInPNextChain<VkAttachmentDescriptionStencilLayout>(render_pass_info->pAttachments[i].pNext);
+            if (attachment_description_stencil_layout) {
+                stencil_layout = attachment_description_stencil_layout->stencilFinalLayout;
+            }
+            cb_state->SetImageViewLayout(*view_state, render_pass_info->pAttachments[i].finalLayout, stencil_layout);
+        }
+    }
+}
+
+// Validates the buffer is allowed to be protected
```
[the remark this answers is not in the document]
jeremyg-lunarg: Will do this in a separate PR. https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6937 so I don't forget.

layers/gpu_validation/gpu_subclasses.cpp
```diff
@@ -186,3 +116,134 @@ void gpuav_state::AccelerationStructureNV::NotifyInvalidate(const NodeList &inva
...
+                                                         di_info ? di_info->descriptor_set_buffers : empty, loc);
+                vmaUnmapMemory(device_state->vmaAllocator, cmd_info.output_mem_block.allocation);
+            }
+        }
+        // For each vkCmdBindDescriptorSets()...
+        // Some applications repeatedly call vkCmdBindDescriptorSets() with the same descriptor sets, avoid
+        // checking them multiple times.
+        vvl::unordered_set<VkDescriptorSet> validated_desc_sets;
+        for (auto &di_info : di_input_buffer_list) {
+            Location draw_loc(vvl::Func::vkCmdDraw);
```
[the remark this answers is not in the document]
jeremyg-lunarg: Yes.  I need to keep track of which draw command uses each descriptor. That needs to be a follow on. I will add a TODO.

tests/unit/gpu_av_positive.cpp
```diff
@@ -742,3 +737,100 @@ TEST_F(PositiveGpuAssistedLayer, SelectInstrumentedShaders) {
...
+    descriptor_writes[0].dstSet = descriptor_set.set_;
+    descriptor_writes[0].dstBinding = 0;
+    descriptor_writes[0].descriptorCount = 1;
+    descriptor_writes[0].descriptorType = VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER;
+    descriptor_writes[0].pBufferInfo = buffer_info;
+    vk::UpdateDescriptorSets(m_device->device(), 1, descriptor_writes, 0, NULL);
+
+    char const *shader_source = R"glsl(
+        #version 450
+        layout(set = 0, binding = 0) uniform foo_0 { int val; } doit;
```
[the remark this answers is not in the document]
jeremyg-lunarg: fixed.

tests/unit/sampler.cpp
```diff
@@ -1211,206 +1203,6 @@ TEST_F(NegativeSampler, UnnormalizedCoordinatesSeparateSamplerSharedSampler) {
...
-
-    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, "VUID-vkCmdDraw-None-08611");
-    vk::CmdDraw(m_commandBuffer->handle(), 1, 0, 0, 0);
-    m_errorMonitor->VerifyFound();
-
-    m_commandBuffer->EndRenderPass();
-    m_commandBuffer->end();
-}
-
-TEST_F(NegativeSampler, UnnormalizedCoordinatesCopyObject) {
```
spencer-lunarg, line 1313: @jeremyg-lunarg I'm confused, we moved this test and disabled it on GPU-AV, but could we still keep the non-GPU-AV ones?

tests/unit/sampler.cpp
```diff
@@ -873,8 +873,8 @@ TEST_F(NegativeSampler, UnnormalizedCoordinatesCombinedSampler) {
     const char fsSource[] = R"glsl(
         #version 450
         layout (set = 0, binding = 0) uniform sampler3D image_view_3d;
-        layout (set = 0, binding = 1) uniform sampler2D tex[2];
```
spencer-lunarg, line 876: @jeremyg-lunarg why did we need to remove the array here?

tests/unit/sampler.cpp
jeremyg-lunarg: These tests require indexing into an array. Core validation doesn't know how to determine if a descriptor array element is statically used so to avoid false positives descriptor arrays are now only validated by GPU-AV.

tests/unit/sampler.cpp
jeremyg-lunarg: again, because core validation doesn't know how to determine if an array element is statically used. This test doesn't require an array so I removed it.

tests/unit/sampler.cpp
spencer-lunarg: If I re-write this SPIR-V to not use an index, it should work then?

also confused because these aren't runtime arrays

```
%int_1 = OpConstant %int 1
%20 = OpAccessChain %ptr_uc %tex %int_1
```

these are constant indexing, this is possible to determine (understand if it isn't currently)

jeremyg-lunarg: > If I re-write this SPIR-V to not use an index, it should work then?

Yes. But when I regenerated the source shader in the comment to not use an index there was no longer an OpAccessChain instruction there.

> these are constant indexing, this is possible to determine (understand if it isn't currently)

It is possible to determine but currently not implemented.  I'm not sure how hard it would be to implement but ideally we'd get 2 things about the SPIRV out of the analysis. 1) if all accesses are static 2) if 1 is true have a static equivalent of `gpuav_state::DescriptorSet::State::UsedDescriptors()`  If that was possible then we'd only need GPU-AV if 1 was false.  There's equivalent logic in sync validation see https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6743

PR #6889, opened 2023-11-01
layers: Invert parameters order in vvl::enumerate

arno-lunarg:
The structured binding now becomes [index, element_ptr]


artem-lunarg (approved):
LGTM. It's nice that now it matches order of `std::ranges::enumerate_view`.

spencer-lunarg (approved):

PR #6894, opened 2023-11-01
layers: Validate conservative rasterization mode, add vuid 06760 06761 08892

ziga-lunarg:


layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -938,6 +938,30 @@ bool CoreChecks::ValidateGraphicsPipelineInputAssemblyState(const PIPELINE_STATE
                              "topology is %s and tessellationShader feature was not enabled.",
                              string_VkPrimitiveTopology(topology));
         }
+
+        if (!phys_dev_ext_props.conservative_rasterization_props.conservativePointAndLineRasterization &&
+            pipeline.vertex_input_state && pipeline.pre_raster_state &&
+            (pipeline.create_info_shaders & VK_SHADER_STAGE_GEOMETRY_BIT) == 0 &&
```
spencer-lunarg, line 944: is this suppose to be when Geometry is used?

ziga-lunarg: No, the VUID says: `...the pipeline does not include a geometry shader...`

spencer-lunarg: sorry, I was reading `06760`

spencer-lunarg (approved):

Issue #6698, opened 2023-10-14
Allow breaking on seq_no and reset_count


darksylinc:
**Environment:**
 - OS: Any
 - GPU and driver version: Any
 - SDK or header version if building from repo: It appears to apply to latest release.
 - Options enabled (synchronization, best practices, etc.): Any

**Describe the Issue**

Normally validation messages will look like this:

```
SYNC-HAZARD-WRITE_AFTER_WRITE(ERROR / SPEC): msgNum: -33950239 - Validation Error: [ SYNC-HAZARD-WRITE_AFTER_WRITE ] Object 0: handle = 0x277a4ff0, name = RID:541792944521253, type = VK_OBJECT_TYPE_BUFFER; | MessageID = 0xfdf9f5e1 | vkCmdCopyBuffer: Hazard WRITE_AFTER_WRITE for dstBuffer VkBuffer 0x277a4ff0[RID:541792944521253], region 0. Access info (usage: SYNC_COPY_TRANSFER_WRITE, prior_usage: SYNC_COPY_TRANSFER_WRITE, write_barriers: SYNC_VERTEX_SHADER_UNIFORM_READ|SYNC_TESSELLATION_CONTROL_SHADER_UNIFORM_READ|SYNC_TESSELLATION_EVALUATION_SHADER_UNIFORM_READ|SYNC_GEOMETRY_SHADER_UNIFORM_READ|SYNC_FRAGMENT_SHADER_UNIFORM_READ, command: vkCmdCopyBuffer, seq_no: 1347, reset_no: 5).
```

Very useful info there is that it reports the previous command, identified by `seq_no: 1347, reset_no: 5`.

However currently the layer has no way (either via vkconfig, environment variables, or via vkCmd call) to perform a breakpoint when those two conditions are hit. Right now I have to modify the source code and rebuild the layers:

```cpp
    ResourceCmdUsageRecord(vvl::Func command_, Count seq_num_, SubcommandType sub_type_, Count sub_command_,
                           const CMD_BUFFER_STATE *cb_state_, Count reset_count_)
        : command(command_),
          seq_num(seq_num_),
          sub_command_type(sub_type_),
          sub_command(sub_command_),
          cb_state(cb_state_),
          reset_count(reset_count_)
{
   if( unlikely( seq_num == 1347u && reset_count == 5u ) )
    {
        debug_break();
    }
}
```

[...]

PR #6904, opened 2023-11-02
layers: Report when syncval reaches requested command

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6698

Allows to trigger a breakpoint when a command stream reaches the location identified by the `command number`/`reset count` from the syncval error message. 

The implementation is experimental and can be replaced by a better debugging mechanism in the future. It assumes deterministic application behavior which is often not the case, but still can be useful for simpler scenarios.

In order to trigger a breakpoint for specific `command number`/`reset count` (*prior access* location):
1. Specify command to track by setting environment variables:
`VK_SYNCVAL_DEBUG_COMMAND_NUMBER` (mandatory, can be extracted from existing error message),
`VK_SYNCVAL_DEBUG_RESET_COUNT` (optional, default value is 1, can be extracted from existing error message),
`VK_SYNCVAL_DEBUG_CMDBUF_PATTERN` (optional, default is an empty string). Allows to filter command buffers by substring matching command buffer debug name (`vkSetDebugUtilsObjectNameEXT`).
2. In Vulkan Configuration set debug action as **Break** and **Info** message level. Disable other message levels to minimize spam, since each message will trigger a breakpoint.
3. When SyncVal validation reaches the requested command it will report a message that will trigger a breakpoint.
4. One limitation that any informational message will trigger a breakpoint and such false-positives should be ignored.

### DIFF
--- a/layers/error_message/logging.cpp
+++ b/layers/error_message/logging.cpp
@@ -470,6 +470,7 @@ VKAPI_ATTR VkBool32 VKAPI_CALL MessengerBreakCallback([[maybe_unused]] VkDebugUt
                                                       [[maybe_unused]] VkDebugUtilsMessageTypeFlagsEXT message_type,
                                                       [[maybe_unused]] const VkDebugUtilsMessengerCallbackDataEXT *callback_data,
                                                       [[maybe_unused]] void *user_data) {
+    // TODO: Consider to use https://github.com/scottt/debugbreak
 #ifdef VK_USE_PLATFORM_WIN32_KHR
     DebugBreak();
 #else
--- a/layers/sync/sync_commandbuffer.cpp
+++ b/layers/sync/sync_commandbuffer.cpp
@@ -868,6 +868,7 @@ ResourceUsageTag CommandBufferAccessContext::NextCommandTag(vvl::Func command, N
         access_log_->back().AddHandle(handle);
         command_handles_.emplace_back(std::move(handle));
     }
+    CheckCommandTagDebugCheckpoint();
     return next;
 }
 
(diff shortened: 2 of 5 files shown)


artem-lunarg:
I will update documentation if we decide to keep this.

jzulauf-lunarg:
Will break for any (all) command buffers at a given reset/command sequence number, without the ability to specify which.  I don't think that will meet the requested behavior of the bug. Given that it is possible to name a command buffer using the debug extension perhaps that allowing the command buffer name (with globbing?)  to be specified might meet the requirement.

artem-lunarg:
Good point. I agree this solution is far from user-friendly (but arguably better than nothing). I think in practice this still can do the job if application uses limited amount of command buffers, so you can inspect each breakpoint location.

Need to check if we have an easy way to access debug name, if it's available.

artem-lunarg:
Added `VK_SYNCVAL_DEBUG_CMDBUF_PATTERN` to specify debug command buffer name.

spencer-lunarg (approved):
@jzulauf-lunarg Talked to Artem about this in length, he plans to add another small comment, but I think this is good to go, I rather provide a small, very non-invasive system to do the debug breaking for those who might want it. The goal is to examine its useful ness overtime and the decide to

1. Remove it
2. Make it more public that this is possible
3. Find it is better to move it to a different part of code and/or modify (example https://github.com/orgs/KhronosGroup/projects/21?pane=issue&itemId=46101445)

artem-lunarg:
@jzulauf-lunarg @spencer-lunarg This definitely a useful feature. So far two users asked for this. And now the third user - me, in a personal project used to identify location of the error. Going to add this to documentation. This is mostly useful for small apps which have reproducible behavior on each run, but that's usually the type of apps that new users write when learning vulkan or when you experiment with some vulkan features.

PR #6906, opened 2023-11-02
layers: Remove unassigned image barrier aspect vuid

ziga-lunarg:
There is a VUID that prohibits image memory barrier with depth/stencil aspects on a color image, but there is nothing in the spec that says aspect can't be depth and stencil, if image only has one of them.

### DIFF
--- a/layers/state_tracker/image_state.cpp
+++ b/layers/state_tracker/image_state.cpp
@@ -52,6 +52,16 @@ VkImageSubresourceRange NormalizeSubresourceRange(const VkImageCreateInfo &image
             }
         }
     }
+    // This is required, otherwise an assert will be hit in RangeGenerator
+    if ((norm.aspectMask & VK_IMAGE_ASPECT_COLOR_BIT) != 0 && !vkuFormatIsColor(image_create_info.format)) {
+        norm.aspectMask &= ~VK_IMAGE_ASPECT_COLOR_BIT;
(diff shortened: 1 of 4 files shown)


layers/state_tracker/image_state.cpp
```diff
@@ -52,6 +52,12 @@ VkImageSubresourceRange NormalizeSubresourceRange(const VkImageCreateInfo &image
             }
         }
     }
+    if ((norm.aspectMask & VK_IMAGE_ASPECT_DEPTH_BIT) != 0 && !vkuFormatHasDepth(image_create_info.format)) {
```
ziga-lunarg, line 55: This is required, otherwise an assert will be hit in `RangeGenerator`

jeremyg-lunarg:
I think this change looks correct but it would be good to know which VUID you're referring to.

layers/state_tracker/image_state.cpp
spencer-lunarg:
```suggestion
    // This is required, otherwise an assert will be hit in RangeGenerator
    if ((norm.aspectMask & VK_IMAGE_ASPECT_DEPTH_BIT) != 0 && !vkuFormatHasDepth(image_create_info.format)) {
```

spencer-lunarg (approved):
LGTM with adding of comment

PR #6910, opened 2023-11-04
layers: Add Sync FD Semaphore type

spencer-lunarg:
part of https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5431

adds VUID-VkImportSemaphoreFdInfoKHR-handleType-03264

### DIFF
--- a/layers/state_tracker/state_tracker.cpp
+++ b/layers/state_tracker/state_tracker.cpp
@@ -3260,12 +3260,9 @@ void ValidationStateTracker::PostCallRecordImportSemaphoreFdKHR(VkDevice device,
                                pImportSemaphoreFdInfo->flags);
 }
 
-void ValidationStateTracker::RecordGetExternalSemaphoreState(VkSemaphore semaphore,
+void ValidationStateTracker::RecordGetExternalSemaphoreState(SEMAPHORE_STATE &semaphore_state,
                                                              VkExternalSemaphoreHandleTypeFlagBits handle_type) {
-    auto semaphore_state = Get<SEMAPHORE_STATE>(semaphore);
-    if (semaphore_state) {
-        semaphore_state->Export(handle_type);
-    }
+    semaphore_state.Export(handle_type);
 }
 
 #ifdef VK_USE_PLATFORM_WIN32_KHR
@@ -3280,7 +3277,10 @@ void ValidationStateTracker::PostCallRecordGetSemaphoreWin32HandleKHR(VkDevice d
                                                                       const VkSemaphoreGetWin32HandleInfoKHR *pGetWin32HandleInfo,
                                                                       HANDLE *pHandle, const RecordObject &record_obj) {
     if (VK_SUCCESS != record_obj.result) return;
-    RecordGetExternalSemaphoreState(pGetWin32HandleInfo->semaphore, pGetWin32HandleInfo->handleType);
+    auto semaphore_state = Get<SEMAPHORE_STATE>(pGetWin32HandleInfo->semaphore);
+    if (semaphore_state) {
+        RecordGetExternalSemaphoreState(*semaphore_state, pGetWin32HandleInfo->handleType);
+    }
 }
 
 void ValidationStateTracker::PostCallRecordImportFenceWin32HandleKHR(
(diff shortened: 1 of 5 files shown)


layers/state_tracker/state_tracker.cpp
```diff
@@ -3301,7 +3301,17 @@ void ValidationStateTracker::PostCallRecordGetFenceWin32HandleKHR(VkDevice devic
...
     if (VK_SUCCESS != record_obj.result) return;
-    RecordGetExternalSemaphoreState(pGetFdInfo->semaphore, pGetFdInfo->handleType);
+    auto semaphore_state = Get<SEMAPHORE_STATE>(pGetFdInfo->semaphore);
+    if (semaphore_state) {
+        ExternalOpaqueInfo external_info = {};
+        external_info.semaphore_flags = semaphore_state->flags;
+        external_info.semaphore_type = semaphore_state->type;
+
+        WriteLockGuard guard(fd_handle_map_lock_);
+        fd_handle_map_.insert_or_assign(*pFd, external_info);
```
artem-lunarg, line 3311: if this lock is not needed for Record then one option is to limit scope to map operation.

tests/unit/external_memory_sync_positive.cpp
```diff
@@ -284,3 +284,46 @@ TEST_F(PositiveExternalMemorySync, BufferDedicatedAllocation) {
...
+    VkPhysicalDeviceExternalSemaphoreInfo external_semahpore_info = vku::InitStructHelper();
+    external_semahpore_info.handleType = handle_type;
+
+    VkExternalSemaphoreProperties external_semahpore_props = vku::InitStructHelper();
+    vk::GetPhysicalDeviceExternalSemaphoreProperties(gpu(), &external_semahpore_info, &external_semahpore_props);
+    if (!(external_semahpore_props.externalSemaphoreFeatures & VK_EXTERNAL_SEMAPHORE_FEATURE_EXPORTABLE_BIT_KHR) ||
+        !(external_semahpore_props.externalSemaphoreFeatures & VK_EXTERNAL_SEMAPHORE_FEATURE_IMPORTABLE_BIT_KHR)) {
+        GTEST_SKIP() << "External semaphore does not support importing and exporting";
+    }
+    if (!(external_semahpore_props.compatibleHandleTypes & handle_type)) {
```
artem-lunarg, line 305: Not sure about special cases in drivers, but checking `externalSemaphoreFeatures` alone I think is enough to make a check for support.  `compatibleHandleTypes` is a good sanity check, supported handle shoulld be compatible with itself. Not a request for change, just dumping what I remembered from previous work.

artem-lunarg (approved):
LGTM. Comment about semaphore scope.

y-novikov:
I've bisected the failures here https://ci.chromium.org/ui/p/chromium/builders/try/dawn-linux-x64-deps-rel/42887/overview (from rolling VVL into Chrome) to this commit.
We've started seeing VUID-VkMemoryDedicatedAllocateInfo-image-01878 and VUID-VkMemoryAllocateInfo-allocationSize-01742 in some tests. Since this commit isn't intended to affect these, maybe something got wrong with state tracking?

spencer-lunarg:
@y-novikov I am build Dawn now locally and will try to re-run the tests

spencer-lunarg:
@y-novikov I think this is a real bug the Validation layers have now caught

I am seeing 

```cpp
vkGetMemoryFdKHR(); // fd == 14
vkGetMemoryFdKHR(); // fd == 15
vkGetSemaphoreFdKHR(); // fd == 23
vkGetSemaphoreFdKHR(); // fd == 24

VkImportMemoryFdInfoKHR pNext;
pNext.fd = 24
vkAllocateMemory(pNext);
```

cc @Kangz since there is a chance I am reading into the external FD spec incorrectly

y-novikov:
Thanks! I've updated crbug.com/1499919#c14 with that info.

Kangz:
Yes that's definitely our bug. In general I think it is safe for now to assume that any VVL errors around sync fds are issues in Dawn because it is a less tested path, and still exhibit various issues. It's great to see errors like this caught, in the past we had to do a lot of trial and error to figure out misuses of the Vulkan API like this one.

PR #6913, opened 2023-11-04
layer: Refactor ImageAccess to handle copies

spencer-lunarg:
Cleans up the `ImageAccess` logic to not assume common cases and instead created a proper while loop to walk the SPIR-V to find the `OpVariable`


artem-lunarg (approved):
LGTM

PR #6914, opened 2023-11-05
layers: Validate active queries with mesh shaders, add vuid 07074 07075

ziga-lunarg:


layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -2521,6 +2521,22 @@ bool CoreChecks::ValidatePipelineDrawtimeState(const LAST_BOUND_STATE &last_boun
         }
     }
 
+    if (loc.function == Func::vkCmdDrawMeshTasksEXT || loc.function == Func::vkCmdDrawMeshTasksIndirectCountEXT ||
```
spencer-lunarg, line 2524: I think this is what we have `ValidateMeshShaderStage` for (`06480` is in the same `draw_mesh_common.adoc` file in the spec)

spencer-lunarg (approved):

PR #6916, opened 2023-11-05
scripts: Fix validation stats script not handling constant values on separate line

aqnuep:


scripts/vk_validation_stats.py
```diff
@@ -171,6 +171,11 @@ def buildKvuidDict(unassigned_vuid_files):
                     if eq_pos >= 0:
                         kvuid = line[kvuid_pos:eq_pos].strip(' \t\n;"')
                         unassigned_str = line[eq_pos+1:].strip(' \t\n;"')
+                        # Handle the case when the value of the constant is on a new line due to formatting
```
spencer-lunarg, line 174: change looks harmless and good to me, but just clarifying what you mean by `constant` here? Can't find a `kVUID_` example where this is an issue

spencer-lunarg (approved):

scripts/vk_validation_stats.py
aqnuep: Just look at the `best_practices_error_enums.h` file. Lots of constants are line-broke by clang due to their length.

The script currently will consider the values of these constants to be the empty string, which will produce incorrect data in the end.

aqnuep: Of course, by currently, I mean before this fix. I found out accidentally while debugging something.

spencer-lunarg: sorry, constant as in `string const`.. yep, 100% see it now, thanks for adding this patch!

PR #6922, opened 2023-11-07
layers: Have single way to get SpecConstant values

spencer-lunarg:
- Move SpecConstant look up into a `PipelineStageState` helper function
- Only track a single set of SpecConstant mappings
- Adds more tests (most are just getting ready for a proper fix for #5911 )


artem-lunarg (approved):
LGTM. Thanks for explaining details.

PR #6923, opened 2023-11-08
tests: Draft for redundant core and extension dependencies

aqnuep:
This is just a draft to see whether this approach looks fine. This, as is, currently will fail CI, so we'll have to add some mechanisms to deal with extensions that intentionally depend on the old behavior.

What this does is that if the test requests both a core API version (target, not necessarily required) and an extension that's part of the effective core version, then it does not enable the extension, only the core version.

Maybe we should relax this and just remove the requirement, so this is for discussion for now based on our offline conversations with @spencer-lunarg.

### DIFF
--- a/scripts/generators/extension_helper_generator.py
+++ b/scripts/generators/extension_helper_generator.py
@@ -67,44 +67,29 @@ def exprToCpp(pr: ParseResults, opt = lambda x: x) -> str:
     dependCheck(pr, printExt, printOp, openParen, closeParen)
     return ''.join(r)
 
-# This class is a container for any source code, data, or other behavior that is necessary to
-# customize the generator script for a specific target API variant (e.g. Vulkan SC). As such,
-# all of these API-specific interfaces and their use in the generator script are part of the
-# contract between this repository and its downstream users. Changing or removing any of these
-# interfaces or their use in the generator script will have downstream effects and thus
-# should be avoided unless absolutely necessary.
-class APISpecific:
-    # Returns dictionary of version field names
-    @staticmethod
-    def getVersionFieldNameDict(targetApiName: str) -> dict[str, str]:
-        match targetApiName:
-
-            # Vulkan specific version field names
-            case 'vulkan':
-                return {
-                    'VK_VERSION_1_1': 'vk_feature_version_1_1',
-                    'VK_VERSION_1_2': 'vk_feature_version_1_2',
-                    'VK_VERSION_1_3': 'vk_feature_version_1_3'
-                }
-
-
-    # Returns dictionary of promoted extension array variable names
-    @staticmethod
-    def getPromotedExtensionArrayName(targetApiName: str, scope: str) -> dict[str, str]:
-        match targetApiName:
-
-            # Vulkan specific promoted extension array variable names
-            case 'vulkan':
-                return {
-                    'VK_VERSION_1_1': f'V_1_1_promoted_{scope}_apis',
-                    'VK_VERSION_1_2': f'V_1_2_promoted_{scope}_apis',
-                    'VK_VERSION_1_3': f'V_1_3_promoted_{scope}_apis'
-                }
-
 
 class ExtensionHelperOutputGenerator(BaseGenerator):
     def __init__(self):
         BaseGenerator.__init__(self)
+
+    def generatePromotedExtensionMap(self, out: list[str], type: str):
+        out.append('''
+            static const PromotedExtensionInfoMap &get_promotion_info_map() {
+                static const PromotedExtensionInfoMap promoted_map = {
+            ''')
+
+        for version in self.vk.versions.keys():
+            promoted_ext_list = [x for x in self.vk.extensions.values() if x.promotedTo == version and getattr(x, type)]
+            if len(promoted_ext_list) > 0:
+                out.append(f'{{{version.replace("VERSION", "API_VERSION")},{{"{version}",{{')
+                out.extend(['    %s,\n' % ext.nameString for ext in promoted_ext_list])
+                out.append('}}},\n')
+
+        out.append('''
+                };
+                return promoted_map;
+            }
+            ''')
 
     def generate(self):
         # [ Feature name | name in struct InstanceExtensions ]
@@ -121,8 +106,8 @@ class ExtensionHelperOutputGenerator(BaseGenerator):
                 for reqs in exprValues(parseExpr(temp)):
                     feature = self.vk.extensions[reqs] if reqs in self.vk.extensions else self.vk.versions[reqs]
                     requiredExpression[extension.name].append(feature)
-        for version, field in APISpecific.getVersionFieldNameDict(self.targetApiName).items():
-            fieldName[version] = field
+        for version in self.vk.versions.keys():
+            fieldName[version] = version.lower().replace('version', 'feature_version')
 
         out = []
         out.append(f'''// *** THIS FILE IS GENERATED - DO NOT EDIT ***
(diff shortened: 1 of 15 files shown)


aqnuep:
A particular issue with this is that some test cases rely on the KHR and/or core version of the entry points, so depending on which gets loaded and which doesn't, there may also be `NULL` entry point calls as a result, so we have to think through whether/how we'd want to deal with that.

tests/unit/instance_positive.cpp
```diff
@@ -40,22 +40,6 @@ TEST_F(PositiveInstance, TwoInstances) {
     ASSERT_NO_FATAL_FAILURE(vk::DestroyInstance(i1, nullptr));
 }
 
-TEST_F(PositiveInstance, NullFunctionPointer) {
```
aqnuep, line 43: I've removed this test, because it doesn't actually test validation layer functionality, but driver functionality.

Also, as observed, the profiles layer does override the API version of the instance, which can cause this to fail when ran with the profiles layer with a 1.3 profile.

aqnuep:
Hi, @spencer-lunarg. I believe this PR is ready for review. In the last run I still got some failures (AMD best practices layer and GPU AV), but I believe none of those are related, so I'd appreciate your feedback on those, and, of course, on the PR in general.

scripts/generators/extension_helper_generator.py
```diff
@@ -353,61 +341,56 @@ def generate(self):
...
+                if (pCreateInfo && pCreateInfo->ppEnabledExtensionNames) {
+                    for (uint32_t i = 0; i < pCreateInfo->enabledExtensionCount; i++) {
+                        if (!pCreateInfo->ppEnabledExtensionNames[i]) continue;
+                        auto info = get_info(pCreateInfo->ppEnabledExtensionNames[i]);
+                        if (info.state) this->*(info.state) = kEnabledByCreateinfo;
+                    }
+                }
+
+                // Workaround for functions being introduced by multiple extensions, until the layer is fixed to handle this correctly
+                // See https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5579 and https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5600
```
spencer-lunarg, line 372: I was literally thinking about these the other days for another issue, was curious if you had solved it, but this is fair for now... there always are exceptions :disappointed:

tests/unit/instance_positive.cpp
spencer-lunarg: agree, these types of tests have sneaked in the past

spencer-lunarg (approved):
LGTM - I can take over and figure out what is failing... honestly a lot of those Best Practice tests were not written fully correct and I've debated many times to `DISABLE` them until someone had time to look into them

spencer-lunarg:
ok, so seems it nothing to do with the best practice tests, but no matter what, the second test ran on the AMD Linux Mesa machine crashes... I can't reproduce on my Intel Linux Mesa machine, will continue to investigate, but once solved, should be able to just merge this in

tests/unit/image_drm.cpp
```diff
@@ -122,21 +122,21 @@ TEST_F(NegativeImageDrm, ImageFormatInfo) {
 
     VkImageFormatProperties2 image_format_properties = vku::InitStructHelper();
     m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkPhysicalDeviceImageFormatInfo2-tiling-02249");
-    vk::GetPhysicalDeviceImageFormatProperties2KHR(gpu(), &image_format_info, &image_format_properties);
+    vk::GetPhysicalDeviceImageFormatProperties2(gpu(), &image_format_info, &image_format_properties);
```
spencer-lunarg, line 125: honestly surprised this is ALL you found for functions that needed to be changed, I was expecting like a 100 line diff just of these

spencer-lunarg:
@aqnuep so the issue is somehow in `state_tracker.cpp` it is calling

`GetPhysicalDeviceExtProperties(physical_device, dev_ext.vk_feature_version_1_2, &phys_dev_props_core11);`

for a test that uses 1.0 and no added extension... from here seems AMD mesa will be the only driver that crashes from calling into `GetPhysicalDeviceProperties2(KHR)` without requesting it

it shows `vk_feature_version_1_3` is true in the layers which causes the layers to think it is safe to go and call `GetPhysicalDeviceProperties2` which blows things up

scripts/generators/extension_helper_generator.py
aqnuep: Hmm, yes, that's an interesting one...

I'll think about those cases as well to see if we can do something about it in a future change.

tests/unit/image_drm.cpp
aqnuep: Yes, I was surprised as well :)

Even this one only has this problem due to the fixture only supporting enablement through core API version and not through extension. If I would have changed the fixture to also require the GPDP2 extension besides the target version, then the KHR extension entry points would be populated with the core ones using the new mechanisms in this PR, so even this would be unnecessary.

In general, it will be easier to write test cases from now on that depend on functionality that is available in both extension form and as a core feature.

I do expect _some_ further exceptions that may be needed for things like draw indirect count and shader draw parameters, where feature flags only exist in core, but not in the extension, but there are no tests for those currently that currently trigger the corner cases.

My follow-up PR in a couple of weeks, where I get through all promoted extensions and making sure their tests would run if either the core version or the extension is available, will likely contain some code to handle those corner cases as well, but it was unnecessary for the scope of this PR.

aqnuep:
> @aqnuep so the issue is somehow in `state_tracker.cpp` it is calling
> 
> `GetPhysicalDeviceExtProperties(physical_device, dev_ext.vk_feature_version_1_2, &phys_dev_props_core11);`
> 
> for a test that uses 1.0 and no added extension... from here seems AMD mesa will be the only driver that crashes from calling into `GetPhysicalDeviceProperties2(KHR)` without requesting it
> 
> it shows `vk_feature_version_1_3` is true in the layers which causes the layers to think it is safe to go and call `GetPhysicalDeviceProperties2` which blows things up

Ah, yes, this may be one of those cases where the profiles layer messes things up. As mentioned in a comment, there's this weird behavior that the profiles layer will override the application-requested API version to the API version of the profile, which I believe is incorrect, and not sure why it does that.

Are these tests run with the profiles layer or just pure RADV?

spencer-lunarg:
> Are these tests run with the profiles layer or just pure RADV?

Just pure RADV, no profile layer was used at all for internal CI runs, that is only for Github actions (and local testing)

aqnuep:
> > Are these tests run with the profiles layer or just pure RADV?
> 
> Just pure RADV, no profile layer was used at all for internal CI runs, that is only for Github actions (and local testing)

I see. I'll take a look at it on a machine with AMD + RADV. Thanks!

spencer-lunarg:
one more thing, it is a Ubuntu 22.04 machine and it has the default mesa driver on it that is packaged, it hasn't been updated

aqnuep:
> one more thing, it is a Ubuntu 22.04 machine and it has the default mesa driver on it that is packaged, it hasn't been updated

Thanks. In the meantime I already root caused it. It was a legitimate failure due to a typo. Once I run the whole suite on it I'll update the PR.

aqnuep:
Okay, with the latest changes this passes everything. I rebased the PR, once more (commits keep getting in), but this is now really ready to go.

lunarpapillo:
CI build # 14551 is likely to fail due to (likely unrelated) GPU crash.  I've requeued.

PR #6924, opened 2023-11-08
scripts: Add script to help with unreleased extensions

spencer-lunarg:
The goal of this script is to call `python scripts/build_with_extension.py --spec_repo path/to/Vulkan-Docs` where the spec repo has an unreleased extensions branch so the `vk.xml` is updated.

Instead of trying to create an alternate `Vulkan-Headers` repo, this cheats a bit and just updates the `Vulkan-Headers` dependency as this is tailored towards people doing one-off VVL development

### DIFF
--- a/scripts/build_with_extension.py
+++ b/scripts/build_with_extension.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) 2023 LunarG, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import sys
+import argparse
+import shutil
+import common_ci
+
+def main():
+    parser = argparse.ArgumentParser(description='Help build with unreleased extensions')
+    parser.add_argument('--spec_repo', action='store', type=str, help='Path to Vulkan-Docs repo')
+    parser.add_argument('--build_dir', action='store', type=str, help='Defaults to ./build')
+    args = parser.parse_args()
+
+    if not args.spec_repo:
+        sys.exit('missing --spec_repo')
+
+    spec_repo_path = os.path.abspath(os.path.dirname(args.spec_repo))
+    if (not os.path.isdir(spec_repo_path)):
+        sys.exit(f'Cannot find directory {spec_repo_path}')
+
+    vk_xml_path = os.path.join(spec_repo_path, 'xml', 'vk.xml')
+    if (not os.path.isfile(vk_xml_path)):
+        sys.exit(f'Cannot find vk.xml file at {vk_xml_path}')
+
+    build_dir = args.build_dir if args.build_dir else common_ci.RepoRelative('build')
+
+    # Will grab latest Vulkan-Headers and install properly
+    common_ci.RunShellCmd(command=f'cmake -S . -B {build_dir} -D UPDATE_DEPS=ON -D BUILD_WERROR=ON -D BUILD_TESTS=ON -D CMAKE_BUILD_TYPE=Debug', verbose=True)
+
+    external_dir = 'external/Debug/64'
+    vulkan_header_path = common_ci.RepoRelative(f'{external_dir}/Vulkan-Headers')
+    if (not os.path.isdir(vulkan_header_path)):
+        sys.exit(f'Cannot find Vulkan Headers dependency directory at {vulkan_header_path}')
+
+    registry_dir = common_ci.RepoRelative(f'{external_dir}/Vulkan-Headers/registry')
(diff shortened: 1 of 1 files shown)


scripts/build_with_extension.py
```diff
@@ -0,0 +1,77 @@
...
+import shutil
+import common_ci
+
+def main():
+    parser = argparse.ArgumentParser(description='Help build with unreleased extensions')
+    parser.add_argument('--spec_repo', action='store', type=str, help='Path to Vulkan-Docs repo')
+    parser.add_argument('--build_dir', action='store', type=str, help='Defaults to ./build')
+    args = parser.parse_args()
+
+    if not args.spec_repo:
```
jeremyg-lunarg, line 29: couldn't you have a default for this, like `../vulkan` (with os.path magic of course)

scripts/build_with_extension.py
```diff
@@ -0,0 +1,77 @@
...
+            shutil.copyfile(f, os.path.join(vulkan_header_path, 'build', 'install', 'include', 'vulkan', filename))
+
+    # Generate entrypoints
+    common_ci.RunShellCmd(command=f'python3 {common_ci.RepoRelative("scripts/generate_source.py")} {registry_dir} {spirv_headers}', verbose=True)
+    # Update VUL
+    common_ci.RunShellCmd(command=f'python3 {os.path.join(vul_dir, "scripts", "generate_source.py")} {registry_dir}', verbose=True)
+    common_ci.RunShellCmd(command=f'cmake --build {os.path.join(vul_dir, "build")} --target install', verbose=True)
+
+    # Profit💲
+    common_ci.RunShellCmd(command=f'cmake --build {build_dir} --config Debug', verbose=True)
```
jeremyg-lunarg, line 74: with this step in the script, some subset of users will just rerun it every time they want to build. But I guess that's ok. Also, is that a UTF-8 comment?

scripts/build_with_extension.py
spencer-lunarg: I mean this assumes a lot about people dev setup

scripts/build_with_extension.py
spencer-lunarg: the script is called `build` ... the idea it would be advised how to just call `cmake --build .` each time afterwards

scripts/build_with_extension.py
jeremyg-lunarg: we are training them to have a sane environment and I think we should nudge them onto a common dev setup wherever we can.

scripts/build_with_extension.py
```diff
@@ -0,0 +1,77 @@
...
+def main():
+    parser = argparse.ArgumentParser(description='Help build with unreleased extensions')
+    parser.add_argument('--spec_repo', action='store', type=str, help='Path to Vulkan-Docs repo')
+    parser.add_argument('--build_dir', action='store', type=str, help='Defaults to ./build')
+    args = parser.parse_args()
+
+    if not args.spec_repo:
+        sys.exit('missing --spec_repo')
+
+    spec_repo_path = os.path.abspath(os.path.dirname(args.spec_repo))
```
mikes-lunarg, line 32: dirname is chopping off the end of the path here, just use abspath.

mikes-lunarg (commented):
Works on my machine against the current Vulkan-Docs repo (with the mentioned tweak to `spec_repo_path`).  I would really like to try it on an extension branch but I can't find any that look ready.

scripts/build_with_extension.py
```diff
@@ -0,0 +1,77 @@
...
+    vk_xml_path = os.path.join(spec_repo_path, 'xml', 'vk.xml')
+    if (not os.path.isfile(vk_xml_path)):
+        sys.exit(f'Cannot find vk.xml file at {vk_xml_path}')
+
+    build_dir = args.build_dir if args.build_dir else common_ci.RepoRelative('build')
+
+    # Will grab latest Vulkan-Headers and install properly
+    common_ci.RunShellCmd(command=f'cmake -S . -B {build_dir} -D UPDATE_DEPS=ON -D BUILD_WERROR=ON -D BUILD_TESTS=ON -D CMAKE_BUILD_TYPE=Debug', verbose=True)
+
+    external_dir = 'external/Debug/64'
```
juan-lunarg, line 45: This makes assumptions about UPDATE_DEPS.

Use UPDATE_DEPS_DIR to avoid accidental breakage.

scripts/build_with_extension.py
```diff
@@ -0,0 +1,77 @@
...
+    for filename in os.listdir(new_headers):
+        f = os.path.join(new_headers, filename)
+        if os.path.isfile(f):
+            shutil.copyfile(f, os.path.join(vulkan_header_path, 'build', 'install', 'include', 'vulkan', filename))
+
+    # Generate entrypoints
+    common_ci.RunShellCmd(command=f'python3 {common_ci.RepoRelative("scripts/generate_source.py")} {registry_dir} {spirv_headers}', verbose=True)
+    # Update VUL
+    common_ci.RunShellCmd(command=f'python3 {os.path.join(vul_dir, "scripts", "generate_source.py")} {registry_dir}', verbose=True)
+    common_ci.RunShellCmd(command=f'cmake --build {os.path.join(vul_dir, "build")} --target install', verbose=True)
```
juan-lunarg, line 71: What does VUL have to do with this?

scripts/build_with_extension.py
```diff
@@ -0,0 +1,77 @@
...
+
+    # Profit💲
```
juan-lunarg, line 73: ?

spencer-lunarg:
thanks for all the suggestions, looking and talking more, I realize this script is most likely the correct way forward

PR #6931, opened 2023-11-09
layers: Fix check for VUID-07908

aitor-lunarg:
Extension and api version was missing from the check


spencer-lunarg (approved):

tests/unit/dynamic_rendering.cpp
```diff
@@ -3680,8 +3679,11 @@ TEST_F(NegativeDynamicRendering, FragmentDensityMapAttachmentLayerCount) {
                                     VK_FORMAT_FEATURE_FRAGMENT_DENSITY_MAP_BIT_EXT)) {
         GTEST_SKIP() << "VK_FORMAT_FEATURE_FRAGMENT_DENSITY_MAP_BIT_EXT not supported";
     }
-    if (!multiview_features.multiview) {
-        GTEST_SKIP() << "Test requires (unsupported) multiview";
+    if (multiview_features.multiview || physDevProps().apiVersion >= VK_API_VERSION_1_1) {
```
spencer-lunarg, line 3682: nit: we have a `DeviceValidationVersion()`

Just add `SetTargetApiVersion(VK_API_VERSION_1_1);` to top of the test then this can only just be 

```
    if (!multiview_features.multiview) {
        GTEST_SKIP() << "Test requires (unsupported) multiview";
    }
```

spencer-lunarg, line 3682: ok, so for these we have done

```
    SetTargetApiVersion(VK_API_VERSION_1_0);
    AddRequiredExtensions(VK_EXT_FRAGMENT_DENSITY_MAP_EXTENSION_NAME);
    RETURN_IF_SKIP(InitFramework())

    if (DeviceValidationVersion() != VK_API_VERSION_1_0) {
        GTEST_SKIP() << "Tests for 1.0 only";
    }
```


there is no need to query `VkPhysicalDeviceMultiviewFeatures` if we are not enabling the extension so that line can be removed

aitor-lunarg: Applied the changed.

There's one minor comment. I believe this VUID is either malformed or adds nothing. Mainly because the VUID is only present with dynamic rendering which requires multiview, so multiview extension will always be enabled. Shouldn't the VUID check for the feature support instead of extension being enabled? The VUID states:
`If VK_KHR_multiview is not enabled`, but shouldn't it be `If VkPhysicalDeviceMultiviewFeatures.multiview is VK_FALSE`?

spencer-lunarg: there is a good chance it adds no value, we have found a few VUs that are really just caught elsewhere, but were hard to tell with the old `ifdef` hell in the spec

aitor-lunarg: https://gitlab.khronos.org/vulkan/vulkan/-/issues/3714
I'll wait for the resolution of that issue before we move in any direction in VVL

spencer-lunarg:
closing - We merged https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6489 and this VU has since been updated to match the new spec language correctly

Issue #6926, opened 2023-11-08
Tripping `assert(pos == map.lower_bound(range))` in sync_validation


amaiorano:
**Environment:**
 - OS: Linux
 - GPU and driver version: N/A
 - SDK or header version if building from repo: 70837963f82b
 - Options enabled (synchronization, best practices, etc.):

**Describe the Issue**

In Chromium tests, we are hitting an assertion that was added in [this change](https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/7d1db72436a6933820baf64c174f7037bc054b4b): `pos == map.lower_bound(range)`.

We're not sure if this is our fault, or if it's a false positive from VVL. See below for details.

**Expected behavior**
No assertion.

**Valid Usage ID**
N/A

**Additional context**

When we hit the assertion, we are on our second iteration of the for loop:
```
for (; range_gen->non_empty() && pos != end; ++range_gen) {
```

Here are the values of the variables in question:
```
range = {begin:21093088, end:21093584}
map.lower_bound(range) = {begin:21091532, end:21093248} (locally, I've stored this in a local variable `lower_bound_pos`)

map:
{impl_map_:size=31}
size=31
{begin:1508608, end:1512260}
{begin:1512260, end:1525316}
{begin:10667216, end:10667264}
{begin:10667264, end:10699984}
{begin:14664912, end:14664936}
{begin:14664936, end:14665620}
{begin:21091024, end:21091200}
{begin:21091200, end:21091532}
{begin:21091532, end:21093248} <-- lower_bound_pos, range.begin = 21093088
{begin:21093248, end:21093580} <-- pos
{begin:21093580, end:21095296} <-- range.end = 21093584
{begin:21095296, end:21095628}
{begin:21095628, end:21097344}
{begin:21097344, end:21097676}
{begin:21097676, end:21099392}
{begin:21099392, end:21099724}
{begin:21099724, end:21101440}
{begin:21101440, end:21101772}
{begin:21101772, end:21103488}
{begin:21103488, end:21103820}
{begin:21103820, end:21105536}
{begin:21105536, end:21105868}
{begin:21105868, end:21107584}
{begin:21107584, end:21107916}
{begin:21107916, end:21109632}
{begin:21109632, end:21109964}
{begin:21109964, end:21111680}
{begin:21111680, end:21112012}
{begin:21112012, end:25285328}
{begin:27712464, end:28564432}
{begin:30292944, end:30306000}
```

[...]

PR #6941, opened 2023-11-10
syncval: Remove bad assert/Avoid bad access

jzulauf-lunarg:
An assert was too broadly applied and could trigger in the valid condition that the lower_bound of a range also overlaps and early range (and thus been processed/advanced past).

Also added a missing check vs. end that could have resulted in an invalid access when `pos == map.end()`.

Addresses #6926 .


jzulauf-lunarg:
@amaiorano -- feel free to try this.

amaiorano:
> @amaiorano -- feel free to try this.

Already in the process :)

amaiorano:
It seems to be working on our end - at least I'm not reproducing the assert locally, so that's good!

spencer-lunarg (approved):

PR #6942, opened 2023-11-10
state_tracker: Store Location for queue submissions

jeremyg-lunarg:


spencer-lunarg (approved):
LGTM! I was hoping someone would help finish up this usage of `Location` in the state-tracker so thanks!

PR #6943, opened 2023-11-11
layers: Add Function support for ImageAccess

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5614

Now track functions to detect where variables are found in `ImageAccess`

### DIFF
--- a/layers/state_tracker/shader_module.cpp
+++ b/layers/state_tracker/shader_module.cpp
@@ -613,8 +613,12 @@ ImageAccess::ImageAccess(const SPIRV_MODULE_STATE& module_state, const Instructi
         while (!insn_to_search.empty()) {
             // for debugging, easier if only search one function at a time
             if (new_func) {
+                // If any function can't resolve to a variable, by design,
+                // it will kill searching other functions and those before it are now invalidated.
                 new_func = false;
                 insn = insn_to_search.front();
+                // spirv-val makes sure functions-to-functions are not recursive
+                visited.clear();
             }
 
             const uint32_t current_id = insn->ResultId();
@@ -675,6 +679,7 @@ ImageAccess::ImageAccess(const SPIRV_MODULE_STATE& module_state, const Instructi
                         variable_image_insn.push_back(insn);
                     }
                     insn_to_search.pop();
+                    new_func = true;  // keep searching if more functions
                     break;
                 }
                 default:
--- a/layers/state_tracker/shader_module.cpp
+++ b/layers/state_tracker/shader_module.cpp
@@ -17,6 +17,7 @@
 
 #include <sstream>
 #include <string>
+#include <queue>
 
 #include "state_tracker/pipeline_state.h"
 #include "state_tracker/descriptor_sets.h"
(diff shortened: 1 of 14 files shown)


layers/state_tracker/shader_module.cpp
```diff
@@ -589,15 +590,28 @@ ImageAccess::ImageAccess(const SPIRV_MODULE_STATE& module_state, const Instructi
...
+
         // Keep walking down until get to variables
-        while (true) {
-            uint32_t current_id = insn->ResultId();
-            auto visited_iter = visited.find(current_id);
+        while (!insn_to_search.empty()) {
+            // for debugging, easier if only search one function at a time
+            if (new_func) {
+                new_func = false;
+                insn = insn_to_search.front();
```
artem-lunarg, line 610: Not sure I can see clearly how it works. Could you explain if I miss something. We see new `spv::Op::OpFunctionParameter` and push potentially a list of "args" associated with a given parameter and set flag `new_func = true`. Then in the abovec code if immediately reset `new_func` and grab the instruction for the first argument. If for the previous `spv::Op::OpFunctionParameter` we added multiple arguments how  do we process the rest of them since `new_func` is false?

spencer-lunarg: I spent like 10 minutes about to write up an example and now think you are right... going to write a test that uses multiple arguments and needs the last one to find the invalid usage and report it

spencer-lunarg: ok, pushed a new commit (so you can easily see) but yes, thanks again for catching this

layers/state_tracker/shader_module.cpp
```diff
@@ -596,15 +597,32 @@ ImageAccess::ImageAccess(const SPIRV_MODULE_STATE& module_state, const Instructi
...
+    std::vector<const Instruction*> sampler_insn_to_search;
+
+    auto walk_to_variables = [this, &module_state, &sampler_insn_to_search](const Instruction* insn, bool sampler) {
         // Protect from loops
         std::unordered_set<uint32_t> visited;
 
+        // stack of function call sites to search through
+        std::queue<const Instruction*> insn_to_search;
+        insn_to_search.push(insn);
+        bool new_func = false;
```
artem-lunarg, line 610: Thanks, it's clear now how updated version works.

I was looking into this `new_func` flag and it seems it's possible to go without it using `insn == null` to indicate we need new instruction. Attached patch if you like the change. I did not test it. 
[remove_new_func_flag.patch](https://github.com/KhronosGroup/Vulkan-ValidationLayers/files/13363795/remove_new_func_flag.patch)

artem-lunarg: oh, the previous patch definitely had bug, one more.
[Uploading attemp2.patch…]()

artem-lunarg (approved):
LGTM

layers/state_tracker/shader_module.cpp
spencer-lunarg: I like this better then adding this random other boolean, will apply and see if CI is happy

spencer-lunarg: this patch fails this case

```
        #version 450
        layout (set = 0, binding = 0) uniform sampler2D tex;

        vec4 foo(sampler2D func_sampler) {
            return textureLodOffset(func_sampler, vec2(0), 0, ivec2(0));
        }

        void main() {
            vec4 x = foo(tex);
        }
```

... going to keep `new_func` for now, not the prettiest, but might be a chance with GPU-AV this gets re-touched anyway soon

PR #6946, opened 2023-11-12
gpuav: Cleanup docs and script

spencer-lunarg:
Adds more information about the GPU-AV system for instrumenting shaders

### DIFF
--- a/docs/gpu_av_shader_instrumentation.md
+++ b/docs/gpu_av_shader_instrumentation.md
@@ -1,0 +1,111 @@
+# GPU-AV Shader Instrumentation
+
+Shader instrumentation is the process of taking an existing SPIR-V file from the application and injecting additional SPIR-V instructions.
+When we can't statically determine the value that will be used in a shader, GPU-AV adds logic to detect the value and if it catches an invalid instruction, it will not actually execute it.
+
+## How Descriptor Indexing is instrumented
+
+As an exaxmple, if the incoming shader was
+
+```glsl
+#version 450
+#extension GL_EXT_nonuniform_qualifier : enable
+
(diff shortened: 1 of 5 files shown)


docs/gpu_av_shader_instrumentation.md
```diff
@@ -0,0 +1,111 @@
...
+bool inst_bindless_check_desc(const uint shader_id, const uint inst_num, const uvec4 stage_info, const uint desc_set,
+                              const uint binding, const uint desc_index, const uint byte_offset) {
+    // logic
+    return no_error_found;
+}
+```
+
+which is compiled with `glslang`'s `--no-link` option. This is done offline and the module is found in the generated directory.
+
+> Note: This uses `Linkage` which is not techincally valid Vulkan Shader `SPIR-V`, while debugging, some tools might complain
```
jeremyg-lunarg, line 49: once linking is complete, the Linkage capabilities are removed. Thus Linkange only really causes problems if you are debugging the output of the the spirv-opt instrumentation passes

docs/gpu_av_shader_instrumentation.md
```diff
@@ -0,0 +1,111 @@
...
+
+Now with the 2 modules, at runtime `GPU-AV` will call into `spirv-link` which will match up the function arguments and create the final shader which lokos like
```
filnet, line 51: typo: lokos
nit: 2 should be spelled out

docs/gpu_av_shader_instrumentation.md
filnet, line 49: typo: techincally

jeremyg-lunarg (approved):

PR #6953, opened 2023-11-13
layers: Move cvdescriptorset namespace contents to vvl::

jeremyg-lunarg:
cvdescriptorset stood for Core Validation DescriptorSet, which is not a useful grouping of things. Start to move state tracker objects into the common namespace by renaming this one to include them.

Rename all vulkan handle accessors to VkHandle(), which makes it easier to remember than the old inconsistent names that violate coding standards in some cases. (eg. GetSet(), queue(), commandBuffer()).

Note that the VulkanTypedHandle accessor, Handle(), should be used with error logging functionality, such as FormatHandle(), LogObjectList(), and LogError().  The handle type is lost when passing raw non-Dispatchable handles to these functions on 32 bit systems.

### DIFF
--- a/docs/fine_grained_locking.md
+++ b/docs/fine_grained_locking.md
@@ -189,7 +189,7 @@ The following objects have already been updated to meet the above construction r
 * VkBufferView / BUFFER_VIEW_STATE
 * VkImageView / IMAGE_VIEW_STATE
 * VkPipelineLayout / PIPELINE_LAYOUT_STATE
-* VkDescriptorSetLayout / cvdescriptorset::DescriptorSetLayout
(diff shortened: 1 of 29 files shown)


jeremyg-lunarg:
After trying this out with `vvl_state::`, `vvl::state`, and finally just `vvl::`. I think the latter is the best. Once I move different validation areas into their own namespace having state in the namespace names gives you a bad case of Pizza! Pizza! syndrome, eg:

```
namespace gpuav {
namespace state {
class Queue : public vvl::state::Queue {}
}
}
```
`state` just ends up in too many places and loses any meaning.

spencer-lunarg (approved):

artem-lunarg (approved):

PR #6957, opened 2023-11-13
gpu: Put GPUAV and DebugPrintf in namespaces

jeremyg-lunarg:
Convert `gpuav_state::` to `gpuav::`, `debug_printf_state::` to `debug_printf::` and `gpu_utils_state::` to `gpu_tracker::`. Move all definitions for each validation area into these namespaces. Having a separate `_state` namespace isn't very helpful.

Note that the 'main' objects are now called `namespace::Validator`. At some point these very large classes should be split up into at least `namespace::Instance` and `namespace::Device`, but possibly there should be a separate class for every dispatchable handle type. But this is not going to happen soon.

Add a `BaseClass` alias in each `Validator` class so that upcalls are easier to do. In the past, people have forgotten about `gpu_tracker::Validator()` and called directly into `ValidationStateTracker`, which lead to bugs.

Unfortunately, uses of the `VALTRACK_*_STATE_OBJECT()` macros must occur outside of all namespaces.

### DIFF
--- a/layers/gpu_shaders/gpu_shaders_constants.h
+++ b/layers/gpu_shaders/gpu_shaders_constants.h
@@ -20,7 +20,8 @@
 #define GPU_SHADERS_CONSTANTS_H
 
 #ifdef __cplusplus
-namespace gpuav_glsl {
+namespace gpuav {
+namespace glsl {
 using uint = unsigned int;
 
 // Upper bound for maxUpdateAfterBindDescriptorsInAllPools. This value needs to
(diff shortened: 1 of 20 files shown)


Tony-LunarG (approved):
Lot's of changes, but LGTM

jeremyg-lunarg:
> Lot's of changes, but LGTM

Yeah this is a combo of the vvl::state PR and a draft of converting gpuav over to a namespace. The final PR will be smaller.

Tony-LunarG (approved):
LGTM

PR #6958, opened 2023-11-14
layers: Refactor how semaphore binary payload is tracked

artem-lunarg:
It's enough to have a single map to track both binary payload value and the status of whether the payload was updated by this submission.

Adds comments that explain two stages of querying payload value (current submission, then other submissions).

Documents more places where we know that binary semaphore is used and not an arbitrary semaphore.

It has to be an equivalent transformation of code (until it's not).


jeremyg-lunarg (approved):
LGTM

Issue #4955, opened 2022-12-12
Please consider validating lifetime of pipeline libraries


ShabbyX:
The spec requires pipeline libraries to stay alive while the linked pipeline is. This is not validated, but it would be nice if it did. I recently hit a crash in VVL because I was destroying pipeline libraries too early, and it took a bit before I found out I should have kept them alive.

spencer-lunarg:
Agree with adding a lifetime check

> The spec requires pipeline libraries to stay alive while the linked pipeline is

You have a specific line? I want to find proper VUID to tag it with (or create a new one if missing)

ShabbyX:
First paragraph of https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap10.html#pipelines-library:

> A pipeline library is .... The application must maintain the lifetime of a pipeline library based on the pipelines that link with it.

spencer-lunarg:
so I realized this was fixed already in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6472

created a test in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6959 to confirm

PR #6959, opened 2023-11-14
tests: Add test for destroying a GPL before linking

spencer-lunarg:
To confirm we can close https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4955

### DIFF
--- a/tests/unit/graphics_library.cpp
+++ b/tests/unit/graphics_library.cpp
@@ -1772,6 +1772,63 @@ TEST_F(NegativeGraphicsLibrary, BadLibrary) {
     m_errorMonitor->VerifyFound();
 }
 
+TEST_F(NegativeGraphicsLibrary, DestroyedLibrary) {
+    TEST_DESCRIPTION("pLibraries has a destroyed pipeline");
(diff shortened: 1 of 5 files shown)


tests/unit/graphics_library.cpp
```diff
@@ -1772,6 +1772,54 @@ TEST_F(NegativeGraphicsLibrary, BadLibrary) {
     m_errorMonitor->VerifyFound();
 }
 
+TEST_F(NegativeGraphicsLibrary, DestroyedLibrary) {
```
artem-lunarg, line 1775: This test looks similar to `NegativeGraphicsLibrary.BadLibrary` (we probably could set `frag_shader_lib.pipeline_` directly to invalid value, including null?). Also it looks like the original issue described a different scenario. When executable pipeline was created, only after that Shabby tried to delete one of the contributing libraries. Does it make sense and if yes, does validation handle this scenario?

spencer-lunarg: Thanks for catching this, yes the issue was indeed the fact using the linked library and then removing a substate

will re-write test

PR #6962, opened 2023-11-14
syncval: Add end check to pos increment shortcut

jzulauf-lunarg:
Avoids potential derefrence of end.


artem-lunarg (approved):
LGTM

Tony-LunarG (approved):
LGTM and fixes CTS test

PR #6964, opened 2023-11-14
gpu: Fix handling of maximum descriptor records

jeremyg-lunarg:
gpuav_state::DescriptorHeap has to keep track of all resources that could be used in a descriptor, in gpu-accessible memory. Avoid an infinite loop if this limit is hit and raise the limit.


Tony-LunarG (approved):
LGTM - Fixes CTS

spencer-lunarg (approved):

PR #6965, opened 2023-11-15
scripts: Allow running tests on Mac

juan-lunarg:
DYLD_LIBRARY_PATH is basically the same thing as LD_LIBRARY_PATH except for .dylib

Tests will not run otherwise


juan-lunarg:
PR also removes unused MacOS code I forgot to remove

spencer-lunarg (approved):

PR #6966, opened 2023-11-15
android: Document nodelete performance benefits

juan-lunarg:
TLDR: Keep `nodelete` for Android CI performance
https://github.com/LunarG/VulkanTests/pull/432

### DIFF
--- a/tests/android/CMakeLists.txt
+++ b/tests/android/CMakeLists.txt
@@ -15,21 +15,16 @@
 # limitations under the License.
 # ~~~
 
-# NOTE: This can be removed when the minimum API level we support is 28.
-#
 # Our unit tests on Android repeatedly close/open the vulkan validation layers.
-# This causes issues with issues with thread_local variables.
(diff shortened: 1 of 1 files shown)


tests/android/CMakeLists.txt
```diff
@@ -15,21 +15,16 @@
...
-#
+# 
 # https://github.com/android/ndk/issues/360
 # https://github.com/android/ndk/issues/360#issuecomment-339107521
 # https://github.com/android/ndk/releases/tag/r26-beta1
-if ("${CMAKE_ANDROID_STL_TYPE}" MATCHES "static")
-    # nodelete : Marks that vvl shouldn't be unloaded at runtime. 
-    target_link_options(vvl PRIVATE -Wl,-z,nodelete)
-endif()
+target_link_options(vvl PRIVATE -Wl,-z,nodelete)
```
juan-lunarg, line 27: Rough estimate is that this helps test runs go from 15 minutes to 7 minutes so we should keep it regardless of the original thread_local variable issue.

PR #6970, opened 2023-11-15
tests: Cleanup wsi tests

artem-lunarg:
- Adds helper to create transition barriers (there are some subtleties there, and the tests often did that in a different ways).
- Rename  `NegativeWsi.UseSwapchainImageBeforeWait` -> `NegativeWsi.PresentImageWithWrongLayout` to show what the test actually does. It made attempt to create a "missing semaphore wait" scenario, but the actual validation was about image layout. At the moment we don't have the missing semaphore wait validation but it will be added in the future PR.

### DIFF
--- a/tests/framework/binding.cpp
+++ b/tests/framework/binding.cpp
@@ -769,6 +769,52 @@ VkImageAspectFlags Image::aspect_mask(VkFormat format) {
     return image_aspect;
 }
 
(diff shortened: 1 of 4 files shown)


spencer-lunarg (approved):
LGTM - happy to see anything improved to WSI

PR #6977, opened 2023-11-15
syncval: Add workaround to ensure monotonic rangegen

jzulauf-lunarg:
Current ImageRangeGenerator doesn't emit monotonic ranges when the implementation packs MIP levels inside the "minimum stride" empty space.  The range map traversal engines require monotonic ranges.

This workaround forces the range generator encoder setup logic to use "idealized" memory organization for images with more than one mip level.

A better solution will be needed for full accurate aliasing support.

### DIFF
--- a/layers/containers/subresource_adapter.cpp
+++ b/layers/containers/subresource_adapter.cpp
@@ -293,6 +293,10 @@ ImageRangeEncoder::ImageRangeEncoder(const IMAGE_STATE& image, const AspectParam
         }
     }
 
+    // WORKAROUND for not being able to handle packed MIPS without resulting in a non-monotonically increasing range generation
+    // Need to clean this up to correctly detect aliasing conflicts between linear image(s) and buffers
+    if (limits_.mipLevel > 1) linear_image_ = false;
+
     is_compressed_ = vkuFormatIsCompressed(image.createInfo.format);
     texel_extent_ = vkuFormatTexelBlockExtent(image.createInfo.format);
 


layers/containers/subresource_adapter.cpp
```diff
@@ -293,6 +293,10 @@ ImageRangeEncoder::ImageRangeEncoder(const IMAGE_STATE& image, const AspectParam
         }
     }
 
+    // WORKAROUND for not being able to handle packed MIPS without resulting in a non-monotonically increasing range generation
+    // Need to clean this up to correct detect aliasing conflict between linear image(s) and buffers
```
Tony-LunarG, line 297: Correct -> Correctly

Tony-LunarG (approved):
LGTM and fixes the CTS issue I was seeing

layers/containers/subresource_adapter.cpp
jzulauf-lunarg: Updated

spencer-lunarg (approved):

artem-lunarg:
What's `"minimum stride" empty space` ?

artem-lunarg (approved):

jzulauf-lunarg:
@artem-lunarg -- lots of hardware has different minimum in terms of memory alignment, image format stride and the like.  We've seen examples were two mip levels are interlaced with the same row stride, because the row stride is > the row *width*.  Image a system with a minimum stride of 256 bytes, but a minimum alignment of 128bytes.  A 32 pixel wide image with a 4 byte RGBA format would take up half that minimum stride (the first 128 bytes), with the other 128 bytes unused.  However starting the next mip level at 128 bytes offset (with the same stride) could consume a half of that unused space for half of the lines.

How this impacts range generation.... the current ImageRangeGen can handle a limited number of interleaving possibilities (for example whether y or z is interleaved for 3D textures), but assumes that MIP levels *don't* overlap, this means that the ranges it emits when incrementing *both* are non-monotonic (breaking map traversal in syncval) and non-unique (two mip level think all of pixels within rowStride belong to that mip level).

Broadly forcing linear images to the "idealized" memory organization avoids both those issues, but invalidates detailed aliasing checks vs buffers and other linear images.

artem-lunarg:
Interesting, thanks for details.

PR #6978, opened 2023-11-16
tests: Use semi-colon at end of macro

spencer-lunarg:
for @ziga-lunarg - adds a `;` to end of macro to be consistent everywhere

will add to git-blame-ignore after


artem-lunarg (approved):
LGTM. SKIPIFY™.

ziga-lunarg (approved):
Yes, this makes me happy

arno-lunarg (approved):
This also makes me happy :D

PR #6987, opened 2023-11-16
layers: De-shoutify queue_state.{h,cpp}

jeremyg-lunarg:
QUEUE_STATE -> vvl::Queue
SEMAPHORE_STATE -> vvl::Semaphore
FENCE_STATE -> vvl::Fence

### DIFF
--- a/layers/state_tracker/queue_state.h
+++ b/layers/state_tracker/queue_state.h
@@ -28,38 +28,38 @@
 #include "utils/vk_layer_utils.h"
 
 class CMD_BUFFER_STATE;
-class QUEUE_STATE;
 class ValidationStateTracker;
 
-enum SyncScope {
-    kSyncScopeInternal,
-    kSyncScopeExternalTemporary,
-    kSyncScopeExternalPermanent,
-};
+namespace vvl {
+class Queue;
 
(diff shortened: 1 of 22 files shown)


spencer-lunarg (approved):

layers/state_tracker/queue_state.h
```diff
@@ -28,38 +28,38 @@
...
-};
-
-enum FENCE_STATUS { FENCE_UNSIGNALED, FENCE_INFLIGHT, FENCE_RETIRED };
+namespace vvl {
+class Queue;
 
-class FENCE_STATE : public REFCOUNTED_NODE {
+class Fence : public REFCOUNTED_NODE {
   public:
+    enum State { kUnsignaled, kInflight, kRetired };
```
artem-lunarg, line 38: nit: we might consider to use `enum class` later if it works fine. Mostly because k-constants from both enums leak into class scope, more chances to use a wrong one (don't how well compilers detect this).

artem-lunarg (approved):
LGTM. So quiet.

arno-lunarg (approved):

layers/state_tracker/queue_state.h
jeremyg-lunarg: I think they do leak but let’s tackle that separately.

PR #6992, opened 2023-11-17
layers: Split queue_state.{h,cpp} into multiple files

jeremyg-lunarg:
The classes in this file, especially vvl::Semaphore, have gotten tricksy enough to justify having their own files.


artem-lunarg (approved):
LGTM. Thanks for splitting this.

PR #6997, opened 2023-11-20
layers: Cleanup shader_instruction.cpp

spencer-lunarg:
Various things discovered while working on GPU-AV shader instrumentation

### DIFF
--- a/layers/state_tracker/shader_instruction.cpp
+++ b/layers/state_tracker/shader_instruction.cpp
@@ -104,25 +104,6 @@ uint32_t Instruction::GetBitWidth() const {
     return bit_width;
 }
 
-AtomicInstructionInfo Instruction::GetAtomicInfo(const SPIRV_MODULE_STATE& module_state) const {
-    AtomicInstructionInfo info;
-
-    // All atomics have a pointer referenced
-    const uint32_t pointer_index = Opcode() == spv::OpAtomicStore ? 1 : 3;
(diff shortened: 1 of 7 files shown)


layers/state_tracker/shader_instruction.cpp
```diff
@@ -143,4 +124,23 @@ bool Instruction::IsImageArray() const { return (Opcode() == spv::OpTypeImage) &
 bool Instruction::IsImageMultisampled() const {
     // spirv-val makes sure that the MS operand is only non-zero when possible to be Multisampled
     return (Opcode() == spv::OpTypeImage) && (Word(6) != 0);
+}
+
+spv::StorageClass Instruction::StorageClass() const {
```
spencer-lunarg, line 129: never logically made sense to have just this one in the generated file... also this is not a function I find at risk of not having generated as it would be quite the breaking extension to add a new Storage Class based instruction

layers/state_tracker/shader_instruction.h
```diff
@@ -24,14 +24,6 @@
 #include "containers/custom_containers.h"
 #include <spirv/unified1/spirv.hpp>
 
-struct SPIRV_MODULE_STATE;
```
spencer-lunarg, line 27: circular dependency going on, `shader_instruction.h` shouldn't have to know about the Module level

artem-lunarg (approved):

Issue #6980, opened 2023-11-16
VkGpuAssistedLayerTest.GpuBufferOOB crashing on RADV release 


spencer-lunarg:
VkGpuAssistedLayerTest.GpuBufferOOB was re-enabled and now seeing issues again, failures are on RADV release

```
SPIR-V WARNING:
    In file ../src/compiler/spirv/spirv_to_nir.c:1526
    Image Type operand of OpTypeSampledImage should not have a Dim of Buffer.
    3160 bytes into the SPIR-V binary
```

PR #7004, opened 2023-11-22
tests: Fix GpuBufferOOB

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6980

There is a gap in SPIRV-Tools to check `OpTypeSampledImage` doesn't have a Dim of Buffer

while this current code was valid, this likely will fix the issues of it crashing and rather have the test ran


artem-lunarg (approved):

spencer-lunarg:
guess this didn't fix it :sob:

Issue #7002, opened 2023-11-21
Sync val crashes with PositiveDynamicRendering.MatchingAttachmentFormats2


spencer-lunarg:
if you run

`VK_LAYER_TESTS_VALIDATION_FEATURES=all ./tests/vk_layer_validation_tests --gtest_filter=PositiveDynamicRendering.MatchingAttachmentFormats2`

it will assert in sync val at

> sync_validation.cpp:9518: ImageRangeGen syncval_state::ImageViewState::MakeImageRangeGen(const VkOffset3D &, const VkExtent3D &, const VkImageAspectFlags) const: Assertion `subresource_range.aspectMask' failed.

This should be a valid test and should assume a legal Vulkan app could be doing this (so either the test needs to be fixed or sync val)

PR #7013, opened 2023-11-24
tests: Fix MatchingAttachmentFormats2

artem-lunarg:
`VkImageObj::targetView()` is an error-prone design. The method suggests it will return target view based on provided parameters but it does this only for the first call. For subsequent calls it ignores provided parameters and returns cached value. When reading code it's not clear what the result of `targetView` will be without checking previous context. 

The plan for the future change. Get rid of targetView. Use vkt::ImageView directly. VkImageObj currently provides BasicViewCreateInfo() to initialize VkImageViewCreateInfo parameterized by aspect - that's convenient and can be kept. Additional constructor can be added to vkt::ImageView that creates image view based on provided image (and also parameterize by aspect).

Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7002

### DIFF
--- a/tests/unit/dynamic_rendering_positive.cpp
+++ b/tests/unit/dynamic_rendering_positive.cpp
@@ -1086,7 +1086,8 @@ TEST_F(PositiveDynamicRendering, MatchingAttachmentFormats2) {
         m_commandBuffer->EndRendering();
 
         // Matching depth format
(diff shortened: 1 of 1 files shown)


spencer-lunarg (approved):
> VkImageObj::targetView() is an error-prone design.

:100: ... thanks for fixing this

PR #7016, opened 2023-11-25
layers: Add debug info for Instruction

spencer-lunarg:
While debugging `Instruction` is can be hard to understand what is going on with a raw `uint32_t[]` of the words

When seeing code calling `insn->Word(2)` it can be hard to understand what that is grabbing without having to execute the helper member functions in a debugger

### DIFF
--- a/layers/state_tracker/shader_instruction.h
+++ b/layers/state_tracker/shader_instruction.h
@@ -87,4 +87,12 @@ class Instruction {
     small_vector<uint32_t, word_vector_length, uint32_t> words_;
     uint32_t result_id_index_ = 0;
     uint32_t type_id_index_ = 0;
+
+#ifndef NDEBUG
+    // Helping values to make debugging what is happening in a instruction easier
+    std::string opcode_;
+    // helps people new to using SPIR-V spec to understand Word()
+    uint32_t length_;
+    uint32_t words_debug_[12];
+#endif
 };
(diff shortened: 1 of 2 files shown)


layers/state_tracker/shader_instruction.h
```diff
@@ -87,4 +87,19 @@ class Instruction {
     small_vector<uint32_t, word_vector_length, uint32_t> words_;
     uint32_t result_id_index_ = 0;
     uint32_t type_id_index_ = 0;
+
+#ifndef NDEBUG
+    // Helping values to make debugging what is happening in a instruction easier
+    std::string opcode_;
+    // helps people new to using SPIR-V spec to understand Word()
+    uint32_t length_;
+    uint32_t word_0_;
```
arno-lunarg, line 96: using `std::array<uint32_t, 8> words` should achieve the same goal (at least in visual studio it is easy to inspect a `std::array`) and be simpler that this explicit `word_n_` list?

arno-lunarg (approved):

layers/state_tracker/shader_instruction.h
spencer-lunarg: let me see how it looks in vs code

spencer-lunarg: ya, array is nicer, switching

PR #7022, opened 2023-11-27
gpu: Reorganise parts of the validation and error messages logic to simplify things

arno-lunarg:
Adding validation for a command that needs to track the lifetime of its specific validation resources is a bit painful, as you need to update "Uber" functions, and classes, mainly `AllocateValidationResources`, `GenerateValidationMessage`, and the `gpuav::CommandBuffer` class. 
This design does not scale well, this PR is the starting point I propose for an improved organisation. 
One of the root problem is how `AllocateValidationResources` is used. Every recorded command that needs to allocate validation resources and need them to survive until an eventual error message is generated uses `AllocateValidationResources`.
But not all of them want the same thing, right now this is handled with a big `if` block that allocates some resources or not depending on the command being validated. If you want to add your own stuff for some new command, you have to update that block, 
update the `CommandInfo` class with your new struct, think about making sure destruction of this struct is handled correctly, etc...

The new flow I propose starts in "record" calls. Most of them do not need specific resources, and just look like what they use to:
```cpp
void gpuav::Validator::PreCallRecordCmdDraw(VkCommandBuffer commandBuffer, uint32_t vertexCount, uint32_t instanceCount,
                                            uint32_t firstVertex, uint32_t firstInstance, const RecordObject &record_obj) {
    BaseClass::PreCallRecordCmdDraw(commandBuffer, vertexCount, instanceCount, firstVertex, firstInstance, record_obj);

    CommandResources cmd_resources =
        AllocateCommandResources(commandBuffer, VK_PIPELINE_BIND_POINT_GRAPHICS, record_obj.location.function);
    auto cmd_resources_ptr = std::make_unique<CommandResources>(cmd_resources);
    StoreCommandResources(commandBuffer, std::move(cmd_resources_ptr));
}
```

Some functions like `CmdDrawIndirect` need specific resources, and the record call looks like this now:

```cpp
void gpuav::Validator::PreCallRecordCmdDrawIndirect(VkCommandBuffer commandBuffer, VkBuffer buffer, VkDeviceSize offset,
                                                    uint32_t count, uint32_t stride, const RecordObject &record_obj) {
    BaseClass::PreCallRecordCmdDrawIndirect(commandBuffer, buffer, offset, count, stride, record_obj);

    CommandResources cmd_resources =
        AllocateCommandResources(commandBuffer, VK_PIPELINE_BIND_POINT_GRAPHICS, record_obj.location.function);
    if (!gpuav_settings.validate_indirect_buffer) {// no need to allocate specific resources 
        auto cmd_resources_ptr = std::make_unique<CommandResources>(cmd_resources);
        StoreCommandResources(commandBuffer, std::move(cmd_resources_ptr));
        return;
    }
    auto draw_resources = AllocatePreDrawIndirectValidationResources(record_obj.location.function, commandBuffer, cmd_resources,
                                                                     buffer, offset, count, VK_NULL_HANDLE, 0, stride);
    StoreCommandResources(commandBuffer, std::move(draw_resources));
}
```

So the `if command is draw indirect then allocate those validation resources` logic that used to live in `AllocateValidationResources` is now sort of baked in the correct record command.

I added a base `CommandResources` class that stores the resources every command needs, like the error output buffer, and new classes (right now`PreDrawResources`, `PreDispatchResources` and `PreTraceRaysResources`) that also need specific stuff just inherit from this class. The objects instanced from those classes are stored as pointers to this `CommandResources` class and virtual `Destroy` calls are done when cleaning resources, no need to update some random functions by hand. You just need to provide a `Destroy` function.

Class inheriting from `CommandResources` will also typically define a new error logging function, that again gets called "automatically" with a virtual call, helping to reduce reliance on Uber `GenerateValidationMessage`. 

The `AnalyzeAndGenerateMessages` and `GenerateValidationMessage` are still there, I trimmed them down a bit, but I think we need to go further. 
Error messages need to be updated to, the way they are created seems clunky and out of date when compared to how we do things in Core validation.

Again I see those changes as a starting point, a basis for a "how the validation flow in GPU-AV could be improved" discussion

**PR review note**: I tried to keep changes separated with different commits, progressively moving command validation away from the current flow, but over time rebasing become to tedious and the first commits are not clean, some classes things have been renamed in the final commit. They still have value for review purposes, but I will squash the commits together before merging anything.

### DIFF
--- a/layers/gpu_validation/gpu_subclasses.h
+++ b/layers/gpu_validation/gpu_subclasses.h
@@ -51,35 +51,6 @@ struct DescBindingInfo {
     std::vector<DescSetState> descriptor_set_buffers;
 };
 
-// State tracking needed to insert a "pre" draw call aimed at validating buffer data used to perform an indirect draw call
-struct PreDrawResources {
-    VkDescriptorPool desc_pool = VK_NULL_HANDLE;
-    VkDescriptorSet desc_set = VK_NULL_HANDLE;
-    VkBuffer indirect_buffer = VK_NULL_HANDLE;
-    VkDeviceSize indirect_buffer_offset = 0;
-    uint32_t indirect_buffer_stride = 0;
-    VkDeviceSize indirect_buffer_size = 0;
-    static constexpr uint32_t push_constant_words = 4;
-};
-
-// State tracking needed to insert a "pre" dispatch call aimed at validating buffer data used to perform an indirect dispatch call
-struct PreDispatchResources {
-    VkDescriptorPool desc_pool = VK_NULL_HANDLE;
-    VkDescriptorSet desc_set = VK_NULL_HANDLE;
-    VkBuffer indirect_buffer = VK_NULL_HANDLE;
-    VkDeviceSize indirect_buffer_offset = 0;
-    static constexpr uint32_t push_constant_words = 4;
-};
-
-// State tracking needed to insert a "pre" trace rays call aimed at validating buffer data used to perform an indirect trace rays
-// call
-struct PreTraceRaysResources {
-    VkDescriptorPool desc_pool = VK_NULL_HANDLE;
-    VkDescriptorSet desc_set = VK_NULL_HANDLE;
-    VkDeviceAddress indirect_device_address = 0;
-    static constexpr uint32_t push_constant_words = 5;
-};
-
 struct CommonDrawResources {
     // some resources can be used each time so only to need to create once
     bool initialized = false;
@@ -148,32 +119,6 @@ struct CmdIndirectState {
     VkDeviceAddress indirectDeviceAddress;
 };
 
-struct CommandInfo {
-    DeviceMemoryBlock output_mem_block;
-    PreDrawResources draw_resources;
-    PreDispatchResources dispatch_resources;
-    PreTraceRaysResources trace_rays_resources;
-    VkDescriptorSet desc_set;
-    VkDescriptorPool desc_pool;
-    VkPipelineBindPoint pipeline_bind_point;
-    bool uses_robustness;
-    vvl::Func command;
-    uint32_t desc_binding_index;
-    CommandInfo(DeviceMemoryBlock output_mem_block, PreDrawResources draw_resources, PreDispatchResources dispatch_resources,
-                PreTraceRaysResources trace_rays_resources, VkDescriptorSet desc_set, VkDescriptorPool desc_pool,
-                VkPipelineBindPoint pipeline_bind_point, bool uses_robustness, vvl::Func command, uint32_t desc_binding_index)
-        : output_mem_block(output_mem_block),
-          draw_resources(draw_resources),
-          dispatch_resources(dispatch_resources),
-          trace_rays_resources(trace_rays_resources),
-          desc_set(desc_set),
-          desc_pool(desc_pool),
-          pipeline_bind_point(pipeline_bind_point),
-          uses_robustness(uses_robustness),
-          command(command),
-          desc_binding_index(desc_binding_index){};
-};
-
 struct AccelerationStructureBuildValidationInfo {
     // The acceleration structure that is being built.
     VkAccelerationStructureNV acceleration_structure = VK_NULL_HANDLE;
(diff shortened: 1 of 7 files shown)


jeremyg-lunarg (commented):
LGTM. Agree you should squash and put at least some of the PR text into the commit log.

Tony-LunarG (approved):
LGTM - There are still a couple ARNO_TODO s

layers/gpu_validation/gpu_subclasses.h
```diff
@@ -188,10 +133,83 @@ struct AccelerationStructureBuildValidationInfo {
...
+  public:
+    virtual ~CommandResources() {}
+    virtual void Destroy(gpuav::Validator &validator);
+    CommandResources() = default;
+    CommandResources(const CommandResources &) = default;
+    CommandResources &operator=(const CommandResources &) = default;
+
+    void LogErrorIfAny(gpuav::Validator &validator, VkQueue queue, VkCommandBuffer cmd_buffer, const uint32_t operation_index);
+    // Return true iff an error has been logged
+    virtual bool LogValidationMessage(gpuav::Validator &validator, VkQueue queue, VkCommandBuffer cmd_buffer,
```
spencer-lunarg, line 148: why is this virtual but we don't have one for `Destroy`?

layers/gpu_validation/gpu_subclasses.h
```diff
@@ -188,10 +133,83 @@ struct AccelerationStructureBuildValidationInfo {
...
+    vvl::Func command = vvl::Func::Empty;  // Should probably use Location instead
+    uint32_t desc_binding_index = vvl::kU32Max;// desc_binding is only used to help generate an error message
+    std::vector<DescBindingInfo> *desc_binding_list = nullptr;
+};
+
+class PreDrawResources : public CommandResources {
+  public:
+    ~PreDrawResources() {}
+
+    VkDescriptorPool desc_pool = VK_NULL_HANDLE;
```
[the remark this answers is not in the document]
spencer-lunarg: I guess just confused if in the future I want to add something, it doesn't seem obvious when to add to `CommandResources` or to all 3 of these classes

layers/gpu_validation/gpu_subclasses.h
```diff
@@ -188,10 +133,83 @@ struct AccelerationStructureBuildValidationInfo {
...
+    bool LogValidationMessage(gpuav::Validator &validator, VkQueue queue, VkCommandBuffer cmd_buffer, const uint32_t *debug_record,
+                              const uint32_t operation_index, const LogObjectList &objlist);
+};
+
 class CommandBuffer : public gpu_tracker::CommandBuffer {
   public:
-    // per draw/dispatch command state
-    std::vector<CommandInfo> per_draw_command_infos;
+    // per validated command state
+    std::vector<std::unique_ptr<CommandResources>> per_command_infos_2;
```
spencer-lunarg, line 212:
```suggestion
    std::vector<std::unique_ptr<CommandResources>> per_command_infos;
```

the `2` is confusing, this isn't an API, we shouldn't need to 2 version of anything

layers/gpu_validation/gpu_validation.cpp
```diff
@@ -441,104 +441,224 @@ VkPipeline gpuav::Validator::GetDrawValidationPipeline(VkRenderPass render_pass)
...
+    const auto lv_bind_point = ConvertToLvlBindPoint(VK_PIPELINE_BIND_POINT_GRAPHICS);
+    auto const &last_bound = cb_node->lastBound[lv_bind_point];
+    const auto *pipeline_state = last_bound.pipeline_state;
+    const bool use_shader_objects = pipeline_state == nullptr;
+
+    AllocateSharedDrawIndirectValidationResources(use_shader_objects);
+
+    auto draw_resources = std::make_unique<PreDrawResources>();
+    CommandResources &base = *draw_resources;
+    base = cmd_resources;
```
spencer-lunarg, line 628: why did we need to set `*draw_resources` before if we are seeing it here?

layers/gpu_validation/gpu_validation.cpp
```diff
@@ -550,110 +670,128 @@ void gpuav::Validator::AllocatePreDrawValidationResources(const DeviceMemoryBloc
...
+    const auto lv_bind_point = ConvertToLvlBindPoint(VK_PIPELINE_BIND_POINT_COMPUTE);
+    auto const &last_bound = cb_node->lastBound[lv_bind_point];
+    const auto *pipeline_state = last_bound.pipeline_state;
+    const bool use_shader_objects = pipeline_state == nullptr;
+
+    AllocateSharedDispatchIndirectValidationResources(use_shader_objects);
+
+    auto dispatch_resources = std::make_unique<PreDispatchResources>();
+    CommandResources &base = *dispatch_resources;
+    base = cmd_info;
```
spencer-lunarg, line 775: I guess why pass in the `CommandResources` and not just create a `PreDispatchResources` in `PreCallRecordCmdDispatchIndirect` where this being called?

layers/gpu_validation/gpu_subclasses.h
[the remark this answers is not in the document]
arno-lunarg: Those 3 classes use a descriptor set to store different things (the logic to update those descriptor sets is not the same each time) so they all have their own. I will rename `desc_pool` and `desc_set` in each class to make what they store clearer.

Well you only add something to `CommandResources` when **every** recorded commands needs that new thing.

layers/gpu_validation/gpu_subclasses.h
arno-lunarg: That is a leftover, forgot to rename it

layers/gpu_validation/gpu_subclasses.h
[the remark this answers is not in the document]
spencer-lunarg: > Well you only add something to CommandResources when every recorded commands needs that new thing.

makes sense now! would add a comment in `CommandResources` explaining this

PR #7026, opened 2023-11-27
android: Remove redundant --tests argument from android.py

juan-lunarg:
Currently it's possible to build an APK with no testing.

Despite this being invalid/pointless.


spencer-lunarg (approved):
thanks for testing/cleaning this up, otherwise this stuff just sits around and people are afraid to touch it

PR #7039, opened 2023-11-28
layers: Add missing validation for 00995

ziga-lunarg:
VUID 00995 wasnt fully validated, because of it `VkPositiveLayerTest.DestroySwapchainWithBoundImages` was invalid and crashed on some drivers

### DIFF
--- a/tests/unit/wsi_positive.cpp
+++ b/tests/unit/wsi_positive.cpp
@@ -1001,7 +1001,7 @@ TEST_F(VkPositiveLayerTest, DestroySwapchainWithBoundImages) {
     image_create_info.samples = VK_SAMPLE_COUNT_1_BIT;
     image_create_info.tiling = VK_IMAGE_TILING_OPTIMAL;
     image_create_info.initialLayout = VK_IMAGE_LAYOUT_UNDEFINED;
-    image_create_info.usage = VK_IMAGE_USAGE_TRANSFER_SRC_BIT;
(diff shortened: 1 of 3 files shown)


tests/unit/wsi_positive.cpp
```diff
@@ -1001,7 +1001,7 @@ TEST_F(VkPositiveLayerTest, DestroySwapchainWithBoundImages) {
     image_create_info.samples = VK_SAMPLE_COUNT_1_BIT;
     image_create_info.tiling = VK_IMAGE_TILING_OPTIMAL;
     image_create_info.initialLayout = VK_IMAGE_LAYOUT_UNDEFINED;
-    image_create_info.usage = VK_IMAGE_USAGE_TRANSFER_SRC_BIT;
+    image_create_info.usage = VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT;
```
spencer-lunarg, line 1004: :100: excited to try and remove this from the blacklist!

spencer-lunarg (approved):

tests/unit/wsi_positive.cpp
ziga-lunarg: Is it blacklisted? I thought it only crashed on intel

spencer-lunarg: crashes on Mesa as well, spent like 10 minutes at it and had to move on

PR #7040, opened 2023-11-29
syncval: Correct hazard report for submit and exec

jzulauf-lunarg:
Correct and clarify hazard messages at exec and submit command buffer time.

### DIFF
--- a/layers/sync/sync_commandbuffer.h
+++ b/layers/sync/sync_commandbuffer.h
@@ -175,6 +175,24 @@ class CommandExecutionContext : public SyncValidationInfo {
     CommandExecutionContext() : SyncValidationInfo(nullptr) {}
     CommandExecutionContext(const SyncValidator *sync_validator) : SyncValidationInfo(sync_validator) {}
     virtual ~CommandExecutionContext() = default;
+
+    // Are imported command buffers Submitted (QueueBatchContext), or Executed (CommandBufferAccessContext)
+    enum ExecutionType : int {
+        kExecuted = 0,  // Recorded contexts are integrated into context during vkCmdExecuteCommands
+        kSubmitted = 1  // Recorded contexts are integrated into context during vkQueueSubmit (etc.)
+    };
+
+    virtual ExecutionType Type() const = 0;
+
+    const char *ExecutionTypeString() {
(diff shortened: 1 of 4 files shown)


layers/sync/sync_commandbuffer.h
```diff
@@ -175,6 +175,21 @@ class CommandExecutionContext : public SyncValidationInfo {
     CommandExecutionContext() : SyncValidationInfo(nullptr) {}
     CommandExecutionContext(const SyncValidator *sync_validator) : SyncValidationInfo(sync_validator) {}
     virtual ~CommandExecutionContext() = default;
+
+    // Submitted, or Executed
+    enum ExecutionType : int { kExecuted = 0, kSubmitted = 1 };
```
spencer-lunarg, line 180: is there ever a state where it is neither (like a "uninitialized")?

artem-lunarg: Not sure if I understand the new naming. In some parts of the code we use "recorded" term, this term is clear enough - the context is limited only to the command buffer where commands are only _recorded_ and queue effects are not considered yet. Here I'm not sure what "kExecuted" means. Logically it sounds like it's something that happens after submission - at first we submit commands, then they get executed, but not sure if that's the case here since it's use for `CommandBufferAccessContext `.

jzulauf-lunarg: @artem-lunarg The terminology is consistent with the API call the recorded command buffer is being referenced in --   `vkCmdExecuteCommands` or `vkQueueSubmit` (et. al.)

@spencer-lunarg  -- Given that the base class is abstract, only valid derived types can be returned.

artem-lunarg: @jzulauf-lunarg Can we add a comment that "executed" it's about secondary command buffers? Still I don't feel this is similar. Submission indeed implies execution of commands, `vkCmdExecuteCommands` in my opinion is just suboptimal naming on the API side.

artem-lunarg (approved):
LGTM. Comment about naming, but also not sure what better version can be.

jzulauf-lunarg:
Added a comment showing the connection of the "ExecutionType" to the Vulkan API calls represented.

PR #7045, opened 2023-11-29
gpuav: Custom shader instrumentation

spencer-lunarg:
Attempt to add basic support for ` VUID-RuntimeSpirv-OpTraceRayKHR-06356` and `VUID-RuntimeSpirv-OpReportIntersectionKHR-06998` and doing it all with our new internal SPIR-V instrumentation logic


layers/gpu_validation/spirv/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+target_include_directories(gpu_av_spirv PRIVATE .
+    ${CMAKE_SOURCE_DIR}/layers
+    ${CMAKE_SOURCE_DIR}/layers/${API_TYPE}
+)
+
+target_link_libraries(gpu_av_spirv PRIVATE
+    SPIRV-Headers::SPIRV-Headers
+)
+
+# Khronos validation additional dependencies
```
spencer-lunarg, line 44: @arno-lunarg what was your idea for not having to duplicate this CMake `target_link_libraries` logic?

arno-lunarg: A function that takes in a CMake target and add common compile options, libraries, etc...

layers/gpu_validation/spirv/type_manager.h
```diff
@@ -0,0 +1,166 @@
...
+    Module& module_;
+
+    vvl::unordered_map<uint32_t, const Type*> id_to_type_;
+    vvl::unordered_map<uint32_t, const Constant*> id_to_constant_;
+    vvl::unordered_map<uint32_t, const Variable*> id_to_variable_;
+
+    // Currently we don't worry about duplicated types. If duplicate types are added from the original SPIR-V, we just use the first
+    // one we fine. We should only be adding a new Type because it currently doesn't exists.
+    // Use list over vector because we only use this for storage, but use the fast-lookups below to ever transverse the list
+    std::list<Type> types_;
```
jeremyg-lunarg, line 134: I don't get why you need the lists, vs putting the storage in the maps. You might have to use std::unordered_map instead of robin_hood, to ensure stable pointers.

layers/gpu_validation/spirv/instruction.h
```diff
@@ -0,0 +1,81 @@
...
+    // When this class was created, for SPIR-V Instructions that could be used in Vulkan,
+    //   414 of 423 had 6 or less operands
+    //   361 of 423 had 5 or less operands
+    //   287 of 423 had 4 or less operands
+    // An extra word is still needed because each insturction has one word prior to the operands
+    static constexpr uint32_t word_vector_length = 7;
+
+    // Max capacity needs to be uint32_t because an instruction can have a string operand that is (2^16)-1 bytes long
+    small_vector<uint32_t, word_vector_length, uint32_t> words_;
+};
```
jeremyg-lunarg, line 78: It would be great to have a ToString() method for debugging. Eventually.

layers/gpu_validation/spirv/instruction.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+void Instruction::ToBinary(std::vector<uint32_t>& out) {
+    for (auto word : words_) {
+        out.push_back(word);
+    }
+}
+
+// The main challenge with linking to functions from 2 modules is the IDs overlap.
+// Taking advantage that we create and control all the incoming linked functions, we know the limited set of instructions that will
+// ever be used. With this, instead of trying to over-engineer something generated to find all reference IDs in each instructions,
```
jeremyg-lunarg, line 78: How do we really know this, given that the linked functions come from glsl?

layers/gpu_validation/spirv/module.h
```diff
@@ -0,0 +1,80 @@
...
+
+    // provides a way to map back and know which original SPIR-V this was from
+    const uint32_t shader_id_;
+    // Will replace the "OpDecorate DescriptorSet" for the output buffer in the incoming linked module
+    const uint32_t output_buffer_descriptor_set_;
+
+    // Replicate the "Logical Layout of a Module" of SPIR-V.
+    // Divided into sections to make easier to modify each part at different times, but still keeps it simple to write out all the
+    // instructions to a binary format
+    ModuleHeader header_;
```
jeremyg-lunarg, line 49: are you planning on leaving non-const data members as public? Are you using this direct access currently?

jeremyg-lunarg (commented):
this looks like a good start. I'm curious to see how much it would have to grow to accommodate the bindless and BDA passes. They need some additional features like injecting capabilities, extensions and addressing modes into the host module. One advantage of using the SPIRV-Tools Link() code is that it will tell you when this is messed up.

layers/gpu_validation/spirv/type_manager.h
spencer-lunarg: this originally was a vector, but the pointers weren't stable... but yes, a map could store it I realize too

layers/gpu_validation/spirv/instruction.cpp
spencer-lunarg: so the only real advantage of generating is when a new extensions adds `OpNewThing` it is handled automatically in the wild when this is used

since this is from our generated logic, we would catch the fact `OpNewThing` is not used at testing time, we don't have to worry about what the app provides us

layers/gpu_validation/spirv/module.h
spencer-lunarg: this was a hard design decision I went back and forth a lot on, so there are 2 phases

1. when we parse the original SPIR-V
2. When we do a pass/linking and want to edit

From here, I found it such code gymnastic to have the Module class be the only thing that edit this, I felt that `Module` should just be the "raw data" that the other classes decide who is in charge of writing to it, like for example, `TypeManager` is in charge of all the types/constant/variables, the `BasicBlock` is in charge or writing the instructions in it, but the `Module` is the only object the user (gpu_validation.cpp) interacts with

spencer-lunarg: thinking more about this, I might have gone to far in the fear to over abstract things...going to separate the part of `Module` that holds the spirv data and the part to orchestrates everything

juan-lunarg (changes requested):

layers/gpu_validation/spirv/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+    link.h
+    module.h
+    module.cpp
+    type_manager.h
+    type_manager.cpp
+    pass.h
+    pass.cpp
+    ray_tracing_runtime_pass.h
+    ray_tracing_runtime_pass.cpp
+    ${CMAKE_SOURCE_DIR}/layers/${API_TYPE}/generated/spirv_grammar_helper.cpp
```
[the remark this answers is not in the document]
juan-lunarg: This will break add_subdirectory / FetchContent users.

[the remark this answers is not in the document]
juan-lunarg: Use VVL_SOURCE_DIR instead.

https://cmake.org/cmake/help/latest/variable/PROJECT-NAME_SOURCE_DIR.html

spencer-lunarg:
closing, this will be attempted for real after the next SDK freeze, for now, leaving the old system in place

PR #7047, opened 2023-11-29
iOS: Update glslang / SPIRV-Tools

juan-lunarg:


.github/workflows/vvl.yml
```diff
@@ -250,7 +250,8 @@ jobs:
           -D CMAKE_OSX_ARCHITECTURES=arm64 \
           -D BUILD_WERROR=ON \
           -D UPDATE_DEPS=ON \
-          -D CMAKE_BUILD_TYPE=Debug
+          -D CMAKE_BUILD_TYPE=Debug \
+          -D BUILD_TESTS=ON
```
juan-lunarg, line 254: Before iOS testing wouldn't compile due to issues with this glslang

tests/CMakeLists.txt
juan-lunarg, line 1: These pointless stub libraries were removed:
https://github.com/KhronosGroup/glslang/pull/3426

.github/workflows/vvl.yml
juan-lunarg: Ah I completely forgot about all the other problems with iOS testing... I'll revert this bit.

PR #7048, opened 2023-11-29
test: Add RT pipeline test template

arno-lunarg:
@spencer-lunarg have a look at `PositiveRayTracing.BasicTraceRays2` to setup a ray tracing pipeline and trace one ray. I will improve this new framework as needed, tell me what you already want me to add.


tests/framework/ray_tracing_objects.h
```diff
@@ -243,8 +245,30 @@ BuildGeometryInfoKHR BuildGeometryInfoSimpleOnDeviceTopLevel(const vkt::Device&
 // on_host_bottom_level_geometry must have been built previously, and on the host
 BuildGeometryInfoKHR BuildGeometryInfoSimpleOnHostTopLevel(const vkt::Device& device,
                                                            std::shared_ptr<BuildGeometryInfoKHR> on_host_bottom_level_geometry);
-
 }  // namespace blueprint
-
 }  // namespace as
+
+namespace rt {
+class Pipeline {
```
spencer-lunarg, line 252: as much as I like this, I really really really don't want to have 2 ways to create ray tracing pipelines (pipeline_helper.h) after spending so much removing our old "2 ways to do the same thing" flows

spencer-lunarg: I guess as a first PR, I would use this new framework to replace the other (non-NV) Ray Tracing tests

tests/framework/ray_tracing_helper_nv.cpp
```diff
@@ -11,6 +11,193 @@
 
 #include "ray_tracing_helper_nv.h"
 
+RayTracingPipelineHelper::RayTracingPipelineHelper(VkLayerTest &test) : layer_test_(test) { InitInfo(); }
```
spencer-lunarg, line 14: would be nice to explicitly rename this to `RayTracingPipelineHelperNV` or something for someone with auto complete not to accidentally think this is the KHR version

tests/framework/ray_tracing_objects.cpp
```diff
@@ -949,4 +949,248 @@ BuildGeometryInfoKHR BuildGeometryInfoSimpleOnHostTopLevel(const vkt::Device &de
...
+    VkMemoryPropertyFlags ray_gen_sbt_buffer_mem_props = VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT;
+    VkMemoryAllocateFlagsInfo alloc_flags = vku::InitStructHelper();
+    alloc_flags.flags = VK_MEMORY_ALLOCATE_DEVICE_ADDRESS_BIT_KHR;
+    ray_gen_sbt_buffer_.init(device, ray_gen_sbt_buffer_info, ray_gen_sbt_buffer_mem_props, &alloc_flags);
+    auto sbt_buffer_ptr = static_cast<uint8_t *>(ray_gen_sbt_buffer_.memory().map());
+    std::memcpy(sbt_buffer_ptr, sbt_host_storage.data(), sbt_host_storage.size());
+    ray_gen_sbt_buffer_.memory().unmap();
+}
+
+void Pipeline::BasicTraceRays(vkt::Device &device, vkt::CommandBuffer &cmd_buffer, void *push_constants /*= nullptr*/,
```
spencer-lunarg, line 1120: I guess slightly against the idea of the "pipeline" class being in charge of all of this

I would expect the `Pipeline` helper to create a pipeline and for the rest of the `rt` framework to be in charge of setting up the command buffer and descriptor set and submitting it

I also have had to remove these exact type of helpers because they seem to get added for 2 or 3 tests, just for others to not know they exists and/or try to over complex them by having some other member variable un-knowingly change the values being used

tests/framework/ray_tracing_objects.cpp
```diff
@@ -949,4 +949,248 @@ BuildGeometryInfoKHR BuildGeometryInfoSimpleOnHostTopLevel(const vkt::Device &de
...
+
+        VkRayTracingShaderGroupCreateInfoKHR raygen_group_ci = vku::InitStructHelper();
+        raygen_group_ci.type = VK_RAY_TRACING_SHADER_GROUP_TYPE_GENERAL_KHR;
+        raygen_group_ci.generalShader = 0;
+        raygen_group_ci.closestHitShader = VK_SHADER_UNUSED_KHR;
+        raygen_group_ci.anyHitShader = VK_SHADER_UNUSED_KHR;
+        raygen_group_ci.intersectionShader = VK_SHADER_UNUSED_KHR;
+        shader_group_cis_.emplace_back(raygen_group_ci);
+    }
+    for (const auto [miss_shader_i, miss_shader] : vvl::enumerate(miss_shaders_.data(), miss_shaders_.size())) {
```
spencer-lunarg, line 1076: `VUID-VkRayTracingPipelineCreateInfoKHR-stage-06899`

you can only have a single `VK_SHADER_STAGE_MISS_BIT_KHR`

tests/unit/ray_tracing_pipeline_positive.cpp
```diff
@@ -185,18 +190,21 @@ TEST_F(PositiveRayTracingPipeline, GetCaptureReplayShaderGroupHandlesKHR) {
...
-    RayTracingPipelineHelper rt_pipe(*this);
-    rt_pipe.InitLibraryInfoKHR(VK_PIPELINE_CREATE_RAY_TRACING_SHADER_GROUP_HANDLE_CAPTURE_REPLAY_BIT_KHR);
-    rt_pipe.rp_ci_KHR_.stageCount = 0;
-    rt_pipe.rp_ci_KHR_.groupCount = 0;
-    rt_pipe.AddLibrary(rt_pipeline_lib);
-    rt_pipe.InitState();
-    ASSERT_EQ(VK_SUCCESS, rt_pipe.CreateKHRRayTracingPipeline());
+    vkt::rt::Pipeline rt_pipe_lib(*this);
+    rt_pipe_lib.AddCreateInfoFlags(VK_PIPELINE_CREATE_RAY_TRACING_SHADER_GROUP_HANDLE_CAPTURE_REPLAY_BIT_KHR);
+    rt_pipe_lib.InitLibraryInfo();
```
spencer-lunarg, line 195: I feel this should just be part of the `rt::Pipeline` construction, is there a reason some pipelines would not want to set this up... also if a test did care about changing the values, they can just explicitly set the `rt_pipeline_interface_info_` values

tests/unit/ray_tracing_pipeline_positive.cpp
```diff
@@ -185,18 +190,21 @@ TEST_F(PositiveRayTracingPipeline, GetCaptureReplayShaderGroupHandlesKHR) {
...
+    rt_pipe_lib.AddTopLevelAccelStructBinding(*m_device, *m_commandBuffer, 0);
+    rt_pipe_lib.SetRayGenShader(kRayTracingMinimalGlsl);
```
spencer-lunarg, line 197: for graphics/compute pipline, but default it uses the `minimal GLSL` if nothing is provided, so I would expect the RayGen could just do that too

tests/framework/ray_tracing_objects.cpp
arno-lunarg: No you can have multiples, see https://docs.vulkan.org/spec/latest/chapters/raytracing.html#_miss_shaders

tests/framework/ray_tracing_objects.cpp
```diff
@@ -949,4 +949,248 @@ BuildGeometryInfoKHR BuildGeometryInfoSimpleOnHostTopLevel(const vkt::Device &de
...
+    rt_pipeline_interface_info_ = vku::InitStructHelper();
+    rt_pipeline_interface_info_.maxPipelineRayPayloadSize = sizeof(float);  // Set according to payload defined in kRayGenShaderText
+    rt_pipeline_interface_info_.maxPipelineRayHitAttributeSize = rt_pipeline_props.maxRayHitAttributeSize;
+    AddCreateInfoFlags(VK_PIPELINE_CREATE_LIBRARY_BIT_KHR);
+    vk_info_.pLibraryInterface = &rt_pipeline_interface_info_;
+}
+
+void Pipeline::AddBinding(VkDescriptorSetLayoutBinding binding) { bindings_.emplace_back(binding); }
+
+std::shared_ptr<as::BuildGeometryInfoKHR> Pipeline::AddTopLevelAccelStructBinding(vkt::Device &device,
```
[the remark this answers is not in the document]
arno-lunarg: Yes good point

tests/unit/ray_tracing_positive.cpp
```diff
@@ -807,3 +824,22 @@ TEST_F(PositiveRayTracing, AccelerationStructuresDedicatedScratchMemory) {
...
+
+    VkPhysicalDeviceBufferDeviceAddressFeaturesKHR bda_features = vku::InitStructHelper();
+    VkPhysicalDeviceRayTracingPipelineFeaturesKHR ray_tracing_features = vku::InitStructHelper(&bda_features);
+    VkPhysicalDeviceAccelerationStructureFeaturesKHR accel_struct_features = vku::InitStructHelper(&ray_tracing_features);
+    VkPhysicalDeviceFeatures2KHR features2 = vku::InitStructHelper(&accel_struct_features);
+    RETURN_IF_SKIP(InitFrameworkForRayTracingTest(true, &features2));
+    RETURN_IF_SKIP(InitState(nullptr, &features2));
+
+    vkt::rt::Pipeline test_pipeline(*this);
+    test_pipeline.AddTopLevelAccelStructBinding(*m_device, *m_commandBuffer, 0);
```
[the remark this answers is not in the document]
arno-lunarg: Ok makes sense

tests/unit/ray_tracing_pipeline_positive.cpp
arno-lunarg: I expect the vast majority of tests to not do anything related to pipeline libraries, so I will keep this outside of any default construction

tests/framework/ray_tracing_objects.cpp
arno-lunarg: Yeah I will redo this

tests/framework/ray_tracing_helper_nv.cpp
arno-lunarg: I will put it in the `nv::rt` namespace

tests/unit/ray_tracing_pipeline_positive.cpp
arno-lunarg: I think my goal is to make this class rather bare bone, with minimal "under the hood" default values, like default shaders, default acceleration structure, etc...

Then I will do as I did with acceleration structure, and create functions that spit out ready-to-use default ray tracing pipelines. For now I do not have a good view of what a "good default" is, so I am using `vkt::rt::Pipeline` directly. When I have a more informed opinion I will go back and change those tests. It should be fairly easy, so I am ok with this 2 steps process.

tests/unit/ray_tracing_pipeline_positive.cpp
spencer-lunarg: sounds good, what I wanted to confirm

tests/framework/ray_tracing_objects.cpp
arno-lunarg: The purpose of `missIndex` is to choose a miss shader

layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -1432,6 +1432,31 @@ bool CoreChecks::ValidateDrawDynamicStateShaderObject(const LastBound& last_boun
     return skip;
 }
 
+bool CoreChecks::ValidateTraceRaysDynamicState(const LastBound& last_bound_state, const Location& loc) const {
```
spencer-lunarg, line 1435: nit: would rather this and `ValidateDynamicStateSetStatus` have similar names

so maybe `ValidateGraphicsDynamicStateSetStatus` and `ValidateRayTracingDynamicStateSetStatus`

layers/core_checks/cc_drawdispatch.cpp
```diff
@@ -1433,8 +1433,9 @@ bool CoreChecks::ValidateActionState(const vvl::CommandBuffer &cb_state, const V
         }
     }
 
-    if (VK_PIPELINE_BIND_POINT_GRAPHICS == bind_point) {
-        skip |= ValidateDrawDynamicState(last_bound_state, loc);
+    if (bind_point == VK_PIPELINE_BIND_POINT_GRAPHICS) {          // #ARNO need the same for ray tracing. Start from
+                                                                  // CoreChecks::ValidateDynamicStateSetStatus
+        skip |= ValidateDrawDynamicState(last_bound_state, loc);  // #ARNO Create ValidateTraceRaysDynamicState() ???
```
spencer-lunarg, line 1438: remove todo now :smile:

layers/state_tracker/pipeline_state.cpp
```diff
@@ -684,7 +704,7 @@ Pipeline::Pipeline(const ValidationStateTracker *state_data, const VkRayTracingP
       active_shaders(create_info_shaders),  // RTX has no linking shaders
       active_slots(GetActiveSlots(stage_states)),
       max_active_slot(GetMaxActiveSlot(active_slots)),
-      dynamic_state(0),  // RTX has no dynamic states being validated
+      dynamic_state(GetRayTracingDynamicState(*this)),  // RTX has no dynamic states being validated
```
spencer-lunarg, line 707: > // RTX has no dynamic states being validated

well this is a lie now, remove

tests/framework/ray_tracing_objects.cpp
```diff
@@ -945,8 +945,265 @@ BuildGeometryInfoKHR BuildGeometryInfoSimpleOnHostTopLevel(const vkt::Device &de
...
+    vk::CmdBindPipeline(cmd_buffer, VK_PIPELINE_BIND_POINT_RAY_TRACING_KHR, rt_pipeline_);
+
+    if (push_constants && push_constants_byte_size != 0) {
+        assert(push_constants_byte_size <= push_constant_range_size_);
+        vk::CmdPushConstants(cmd_buffer, pipeline_layout_, VK_SHADER_STAGE_RAYGEN_BIT_KHR, 0, push_constants_byte_size,
+                             push_constants);
+    }
+}
+
+void Pipeline::TraceRays(vkt::CommandBuffer &cmd_buffer) {
```
spencer-lunarg, line 1144: I don't see what the "Pipeline" class cares how the pipeline is used, it sole job should be to create a valid `VkPipeline`... how it used after it outside the scope

tests/unit/ray_tracing_pipeline.cpp
```diff
@@ -839,21 +843,23 @@ TEST_F(NegativeRayTracingPipeline, GetCaptureReplayShaderGroupHandlesKHR) {
...
+                                                        rt_pipe.GetShaderGroupsCount(),
                                                         (ray_tracing_properties.shaderGroupHandleCaptureReplaySize - 1), &buffer);
     m_errorMonitor->VerifyFound();
     m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-vkGetRayTracingCaptureReplayShaderGroupHandlesKHR-firstGroup-03483");
     m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-vkGetRayTracingCaptureReplayShaderGroupHandlesKHR-firstGroup-04051");
     // In nv::rt::CreateNVRayTracingPipelineHelper::InitKHRRayTracingPipelineInfo rp_ci_KHR_.groupCount = groups_KHR_.size();
-    uint32_t invalid_firstgroup = rt_pipe.groups_KHR_.size() + 1;
-    vk::GetRayTracingCaptureReplayShaderGroupHandlesKHR(m_device->handle(), rt_pipe.pipeline_, invalid_firstgroup, 0,
+    uint32_t invalid_firstgroup = rt_pipe.GetShaderGroupsCount() + 1;
+    vk::GetRayTracingCaptureReplayShaderGroupHandlesKHR(m_device->handle(), rt_pipe.GetPipelineHandle(), invalid_firstgroup, 0,
```
spencer-lunarg, line 862: should just be `Handle()` to be uniform with all the other helper objects

spencer-lunarg: (I have a task to one day make all `handle()` be `Handle()`)

tests/framework/ray_tracing_objects.cpp
spencer-lunarg: same with `BindResources`

tests/framework/ray_tracing_objects.cpp
```diff
@@ -945,8 +945,265 @@ BuildGeometryInfoKHR BuildGeometryInfoSimpleOnHostTopLevel(const vkt::Device &de
...
+    sbt_buffer.memory().unmap();
+}
+
+void Pipeline::BindResources(vkt::CommandBuffer &cmd_buffer, void *push_constants /*= nullptr*/,
+                             uint32_t push_constants_byte_size /*= 0*/) {
+    vk::CmdBindDescriptorSets(cmd_buffer, VK_PIPELINE_BIND_POINT_RAY_TRACING_KHR, pipeline_layout_, 0, 1, &desc_set_->set_, 0,
+                              nullptr);
+    vk::CmdBindPipeline(cmd_buffer, VK_PIPELINE_BIND_POINT_RAY_TRACING_KHR, rt_pipeline_);
+
+    if (push_constants && push_constants_byte_size != 0) {
```
spencer-lunarg, line 1137: this for me starts to cross the line of what is common and over-complexing the framework

I don't think it is that bad to call 

```c++
vk::CmdBindDescriptorSets()
vk::CmdBindPipeline()
```

manually in the test, it is clear and not that verbose

the few tests that case about push constants should set them themselves. Most tests won't need push constant, and I bet there will be one test were we only want to push "some" of the push constants and this becomes where we either need to make this small function more complex, or we won't even use it

layers/core_checks/cc_drawdispatch.cpp
arno-lunarg: Argh yes x_x

layers/state_tracker/pipeline_state.cpp
arno-lunarg: OMG

layers/core_checks/cc_cmd_buffer_dynamic.cpp
arno-lunarg: Yep better indeed

tests/framework/ray_tracing_objects.cpp
arno-lunarg: Ok will change that

tests/framework/ray_tracing_objects.cpp
arno-lunarg: I think the `CmdBindDescriptorSets` call should remain opaque as the user does not know (and 95% of the time should not care about) what descriptor sets are being used internally, and should be bound. 
Defining new `Get` functions to access those internals is I think not a good thing. 
With that in mind, might as well bundle `CmdBindPipeline` and `CmdPushConstants` in there. For now no test is doing funky things, so it should hold up, and if some test has specific needs, changing or removing this function will be super easy

spencer-lunarg: > I think the CmdBindDescriptorSets call should remain opaque as the user does not know

I guess my reasoning was because `vk::CmdBindDescriptorSets()` is explicitly called in every test we have

The type of test really makes the different here

- Tests dealing with `vkCreateRayTracingPipelinesKHR` won't need to ever set descriptors
- Tests dealing with `vkCmdTraceRays` I would hope know what it is binding and to which index

I guess rather start with less in the helpers and add more once we see a pattern emerge that many tests need... but regardless, adding `CmdBindDescriptorSets` is still outside the scope of a `pipeline` class, so if we do add it, should be somewhere else

layers/state_tracker/cmd_buffer_state.h
```diff
@@ -313,6 +313,8 @@ class CommandBuffer : public REFCOUNTED_NODE {
         uint32_t exclusive_scissor_first;
         uint32_t exclusive_scissor_count;
         std::vector<VkRect2D> exclusive_scissors;
+        // VK_DYNAMIC_STATE_RAY_TRACING_PIPELINE_STACK_SIZE_KHR
+        uint32_t rt_pipeline_stack_size;
```
spencer-lunarg, line 317: For each dynamic state we have a check that it was set with `vkCmdSet*` command

some VUs need to do additional checks at draw time to make sure the last value set is valid... this is not needed for the Stack Size, so we don't need to waste memory tracking it here  (so please remove)

tests/framework/ray_tracing_objects.cpp
```diff
@@ -945,8 +945,251 @@ BuildGeometryInfoKHR BuildGeometryInfoSimpleOnHostTopLevel(const vkt::Device &de
...
+    libraries_.emplace_back(library.rt_pipeline_);
+    pipeline_lib_info_ = vku::InitStructHelper();
+    pipeline_lib_info_.libraryCount = size32(libraries_);
+    pipeline_lib_info_.pLibraries = libraries_.data();
+    vk_info_.pLibraryInfo = &pipeline_lib_info_;
+}
+
+void Pipeline::AddDynamicState(VkDynamicState dynamic_state) { dynamic_states.emplace_back(dynamic_state); }
+
+void Pipeline::Build() {
```
spencer-lunarg, line 1028: Just from a naive view, I see sometimes we use `Build` and sometimes `BuildPipeline` and not sure why

I guess I was expecting a `CreatePipeline` call that would be essentially `Build()` and could detect if the correct state was set to do `BuildSbt`

spencer-lunarg (approved):
LGTM - Just few final nits

tests/unit/ray_tracing.cpp
```diff
@@ -3622,3 +3622,33 @@ TEST_F(NegativeRayTracing, BuildAccelerationStructuresInvalidUpdatesToGeometry)
...
+    SetTargetApiVersion(VK_API_VERSION_1_2);
+
+    VkPhysicalDeviceBufferDeviceAddressFeaturesKHR bda_features = vku::InitStructHelper();
+    VkPhysicalDeviceRayTracingPipelineFeaturesKHR ray_tracing_features = vku::InitStructHelper(&bda_features);
+    VkPhysicalDeviceAccelerationStructureFeaturesKHR accel_struct_features = vku::InitStructHelper(&ray_tracing_features);
+    VkPhysicalDeviceFeatures2KHR features2 = vku::InitStructHelper(&accel_struct_features);
+    RETURN_IF_SKIP(InitFrameworkForRayTracingTest(true, &features2));
+    RETURN_IF_SKIP(InitState(nullptr, &features2));
+
+    vkt::rt::Pipeline test_pipeline(*this, m_device);
```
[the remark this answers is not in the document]
arno-lunarg: true true

tests/framework/ray_tracing_objects.cpp
arno-lunarg: Yeah thinking about it I have no reason to publicly expose `BuildPipeline` and `BuildSbt`, (at least until some test requires it) `Build` will suffice. Also the only requirement for `BuildSbt` is to have your pipeline already built, i.e. you need to call `BuildPipeline` prior

layers/state_tracker/cmd_buffer_state.h
arno-lunarg: Ok will remove that, thank you for the explanation!

PR #7056, opened 2023-11-30
layers: Fixed a typo in the descriptor validator

travisvroman:


spencer-lunarg:
Thanks for the catch! CI is mad because you have 3 commits and need 1 with the `layers: Fixed....` style

also this small fixed seemed to cause `clang-format`  to want to refactor the entire error message function

If you could squash your PR and run `git clang-format --style=file` on the staged change, it will be ready to go

spencer-lunarg:
thanks ... but the check runs per-commit, so it fails on the first commit... sorry for yet-another-step for what should be a simple typo fix, but could you squash the commits

spencer-lunarg (approved):

PR #7059, opened 2023-11-30
tests: Replace targetView with CreateView (part 1)

artem-lunarg:
Part one of replacing `targetView` due to the issue mentioned here https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7013

While targetView is still here this PR adds asserts when it is used incorrectly. A couple of tests triggered that asserts and were fixed. `targetView` has two overloads. This PR starts with a simpler one (called only in a few places) and replaces it with `CreateView` that returns uncached `vkt::ImageView`.


spencer-lunarg (approved):
Love this type of improvement to the test framework!

PR #7062, opened 2023-12-01
printf: Remove stage specific information

### LINKED ISSUE #4350: Debug printf reports "Mixed stage shader module not supported"
**Describe the Issue**

I am trying to compile a collection of ray tracing entry points all from one .hlsl file into one .spv file, which I currently can do through the DXC compiler. I would also like to use "printf" for debugging. If I could do this, I could very closely mimic our current CUDA/OptiX workflow, but with Vulkan, and in a much more device-agnostic setup.

Unfortunately, when I try to create a shader module from this .spv while also using the GPU debug printf extension, I get the following error from vkCreateShaderModule: "Mixed stage shader module not supported"

At the moment, I'm therefore forced to either tell users to split every entry point into a separate .hlsl file---adding a tremendous amount of book keeping---or alternatively I must tell my users that we cannot use printf in any shaders due to the above error message. 

**Valid Usage ID**
ERROR: [-1804403757][UNASSIGNED-Debug-Printf] : Validation Error: [ UNASSIGNED-Debug-Printf ] Object 0: handle = 0x2be7f7e7690, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x9472fbd3 | Error during shader instrumentation: line 0: Mixed stage shader module not supported

**Environment:**
 - OS: Windows 11
 - GPU: RTX 8000 / AMD Radeon RX 6750 XT
 - SDK or header version if building from repo: 1.3.216.0

**Additional context**

Here's a little shader that I'm trying to get working:
[deviceCode.hlsl.txt](https://github.com/KhronosGroup/Vulkan-ValidationLayers/files/9184017/deviceCode.hlsl.txt)

and here's the SPIR-V generated by dxc:
[spirv.txt](https://github.com/KhronosGroup/Vulkan-ValidationLayers/files/9184015/spirv.txt)

I'm using the following DXC options:
    -spirv
    -fspv-target-env=vulkan1.1spirv1.4
    -T lib_6_3
    -fspv-extension=SPV_KHR_ray_tracing 
    -fspv-extension=SPV_KHR_ray_query 
    -fspv-extension=SPV_KHR_non_semantic_info

jeremyg-lunarg:
Remove stage specific debug info that is only needed by GPU-AV. This allows debug printfs to be used in multi stage shader modules.

Requires https://github.com/KhronosGroup/SPIRV-Tools/pull/5495

Fixes #4350


Tony-LunarG (approved):
LGTM

arno-lunarg (approved):

PR #7072, opened 2023-12-04
build: Add common_build.cmake

arno-lunarg:
File to store a function adding vvl build options that should be shared with sub targets
Made to help with the new CMakeLists added with https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7045

### DIFF
--- a/layers/CMakeLists.txt
+++ b/layers/CMakeLists.txt
@@ -16,6 +16,8 @@
 # limitations under the License.
 # ~~~
 
+include(common_build.cmake)
+
 # This variable enables downstream users to customize the CMake targets
 # based on the target API variant (e.g. Vulkan SC)
 set(LAYER_NAME "VkLayer_khronos_validation")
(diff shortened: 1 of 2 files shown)


layers/CMakeLists.txt
```diff
@@ -298,12 +300,12 @@ source_group(TREE "${CMAKE_CURRENT_SOURCE_DIR}" FILES ${LAYER_SOURCES})
 
 set_target_properties(vvl PROPERTIES OUTPUT_NAME ${LAYER_NAME})
 
+common_compile_and_link_options(vvl)
+
 if(MSVC)
     target_link_options(vvl PRIVATE /DEF:${CMAKE_CURRENT_SOURCE_DIR}/${LAYER_NAME}.def)
-    target_compile_options(vvl PRIVATE /bigobj)
```
juan-lunarg, line 303: Don't enable /bigobj unless you absolutely have to. This shouldn't be in any helper function. If anything the codegen needs work to not cause this issue.

New targets/libraries should avoid the need to use /bigobj

layers/common_build.cmake
```diff
@@ -0,0 +1,24 @@
...
+	# OS version / driver combinations. On 32-bit systems, using mimalloc cause an increase in
+	# the amount of virtual address space needed, which can also cause stability problems.
+	if (MSVC AND CMAKE_SIZEOF_VOID_P EQUAL 8)
+	   find_package(mimalloc CONFIG)
+	   option(USE_MIMALLOC "Use mimalloc, a fast malloc/free replacement library" ${mimalloc_FOUND})
+	   if (USE_MIMALLOC)
+	      target_compile_definitions(${TARGET} PRIVATE USE_MIMALLOC)
+	      target_link_libraries(${TARGET} PRIVATE mimalloc-static)
+	   endif()
+	endif()
```
juan-lunarg, line 23: See https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7077 for alternative approach.

Issue #5374, opened 2023-03-05
shader interface matching with PhysicalStorageBuffer


spencer-lunarg:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5349 turned off all checks for `PhysicalStorageBuffer` interface matching because it was giving false positive

Extra support needs to be added to walk the first struct of `PhysicalStorageBuffer` struct for matches

edit - Spec discussion  https://gitlab.khronos.org/vulkan/vulkan/-/issues/3677

spencer-lunarg:
Working group decided that Physical Storage Buffer interfaces are not allowed https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6344

PR #7073, opened 2023-12-04
tests: Add PhysicalStorageBuffer in preperation

spencer-lunarg:
Still waiting on spec change for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5374

but decided to at least get these tests in incase someone else gets to it before me, there is some content to work with. Worst case these become negative tests


spencer-lunarg:
going to just wait until the real PR to add these

PR #7077, opened 2023-12-04
cmake: Cleanup mimalloc and robin_hood code

juan-lunarg:
Use PUBLIC to ensure dependencies are propogated properly.

Libraries and targets following PUBLIC are linked to, and are made part of the link interface.

This nicely shares the code dependency so additional libraries only have to link against VkLayer_utils.

### DIFF
--- a/layers/CMakeLists.txt
+++ b/layers/CMakeLists.txt
@@ -81,6 +81,18 @@ if (USE_ROBIN_HOOD_HASHING)
     target_compile_definitions(VkLayer_utils PUBLIC USE_ROBIN_HOOD_HASHING)
 endif()
 
+# Using mimalloc on non-Windows OSes currently results in unit test instability with some
+# OS version / driver combinations. On 32-bit systems, using mimalloc cause an increase in
+# the amount of virtual address space needed, which can also cause stability problems.
+if (WIN32 AND CMAKE_SIZEOF_VOID_P EQUAL 8)
+   find_package(mimalloc CONFIG)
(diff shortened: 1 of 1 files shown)


layers/CMakeLists.txt
```diff
@@ -81,6 +81,18 @@ if (USE_ROBIN_HOOD_HASHING)
...
 
+# Using mimalloc on non-Windows OSes currently results in unit test instability with some
+# OS version / driver combinations. On 32-bit systems, using mimalloc cause an increase in
+# the amount of virtual address space needed, which can also cause stability problems.
+if (WIN32 AND CMAKE_SIZEOF_VOID_P EQUAL 8)
+   find_package(mimalloc CONFIG)
+   option(USE_MIMALLOC "Use mimalloc, a fast malloc/free replacement library" ${mimalloc_FOUND})
+   if (USE_MIMALLOC)
+      target_compile_definitions(VkLayer_utils PUBLIC USE_MIMALLOC)
+      target_link_libraries(VkLayer_utils PUBLIC mimalloc-static)
```
juan-lunarg, line 92: The important part is the PUBLIC keyword here:
https://leimao.github.io/blog/CMake-Public-Private-Interface/

TLDR: By linking against VkLayer_utils you also link against mimalloc and robinhood via the propagated dependency.

jeremyg-lunarg (approved):
LGTM

layers/CMakeLists.txt
arno-lunarg: Thank you for the explanation!

PR #7082, opened 2023-12-05
layers: Remove UNASSIGNED for non-errors

spencer-lunarg:
In effort to clean up the remaining `UNASSIGNED` VUs, I have gone and removed anything that is not an error from having UNASSIGNED (except for Best Practice, which is its own project in itself)

I few it has there are 3 prefixes

1. `VUID-` which is reserved from proper VUs in the spec
2. `UNASSIGNED-` which are errors that don't have a VU (for one reason or another)
3. everything

Things such as warnings will never have a `VUID` label, so it is not actually unassigned a VUID


layers/object_tracker/object_lifetime_validation.h
```diff
@@ -88,7 +81,7 @@ class ObjectLifetimes : public ValidationObject {
         if (!inserted) {
             // The object should not already exist. If we couldn't add it to the map, there was probably
             // a race condition in the app. Report an error and move on.
-            (void)LogError(kVUID_ObjectTracker_Info, object, loc,
+            (void)LogError("UNASSIGNED-ObjectTracker-Insert", object, loc,
```
artem-lunarg, line 84: I have impression that usually VUIDs are not assigned for race conditions. I remember Tobias said something related o gitlab (https://gitlab.khronos.org/vulkan/vulkan/-/issues/3616#note_429860). I explained this to myself that there could be many source of race condition that are impractical/impossible to enumerate all combinations, but not 100% sure.

artem-lunarg (approved):
LGTM. One comment about UNASSIGNED/VUID for race conditions.

layers/core_checks/cc_image.cpp
```diff
@@ -308,7 +308,7 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
             total_size = (total_size + ig_mask) & ~ig_mask;
 
             if (total_size > format_limits.maxResourceSize) {
-                skip |= LogWarning(kVUID_Core_Image_InvalidFormatLimitsViolation, device, error_obj.location,
+                skip |= LogWarning("WARNING-Image-InvalidFormatLimitsViolation", device, error_obj.location,
```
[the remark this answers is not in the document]
spencer-lunarg: if we add `ERROR` now, then the message will print both `Warning` and `Error` which is even more confusing

I am not against moving this to an error, but then it needs to be `LogError` and it is not a VU, it needs an explanation comment why

layers/object_tracker/object_lifetime_validation.h
spencer-lunarg: Will look into this, for now, this is just moving the string here, but agree, maybe it shouldn't be

spencer-lunarg: added a todo comment

arno-lunarg (approved):

PR #7084, opened 2023-12-05
syncval: Move SyncImageMemoryBarrier to sync_op module

artem-lunarg:
`SyncImageMemoryBarrier` and `SyncBufferMemoryBarrier` were in different files. Logically they belong to a higher-level `sync_op` module, which uses them to define other structures and defines all implementation that works with those structures.


spencer-lunarg (approved):
makes sense to my outside knowledge

jzulauf-lunarg (approved):
LGTM -- There are probably still too many `AccessContext::Detect*` variants for maintainability. Nice to remove a single use variant.

PR #7091, opened 2023-12-06
scripts: Fix mem leak in FreePnextChain

arno-lunarg:


spencer-lunarg:
could we get a test that would help prevent this from regressing

scripts/generators/safe_struct_generator.py
```diff
@@ -279,24 +279,28 @@ def generateUtil(self):
 void FreePnextChain(const void *pNext) {
     if (!pNext) return;
 
-    auto header = reinterpret_cast<const VkBaseOutStructure *>(pNext);
+    auto header = const_cast<VkBaseOutStructure *>(reinterpret_cast<const VkBaseOutStructure *>(pNext));
```
artem-lunarg, line 282: I would be nice to `reinterpret_cast` -> `static_cast`

arno-lunarg: oh yes

arno-lunarg:
> > A test would be to enable leak sanitizer, out of scope ^^"
> 
> We have adding testing in the past to address this exact kind of issue with safe structs. Why is it out of scope? We use address sanitizer which detects memory leaks ([Wiki: AddressSanitizer (google/sanitizers)](https://github.com/google/sanitizers/wiki/AddressSanitizer))
> 
> #5470

I do not see how to test for a memory leak? Also, this PR seems to prove that apparently AddressSanitizer in its current configuration does not detects leaks?

juan-lunarg:
> I do not see how to test for a memory leak? 

This test in particular doesn't check for a memory leak. It tests to ensure the deep copy is valid. The test I wrote will cause ASAN errors before the fixes. I pointed to this PR as an example of using ASAN to test safe structs. Not show a memory leak.

> Also, this PR seems to prove that apparently AddressSanitizer in its current configuration does not detects leaks?

?

The documentation clearly states ASAN detects memory leaks.

@charles-lunarg fixed leaks when helping add ASAN to vvl.
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5395/

charles-lunarg:
ASAN can only detect leaks that happen in the tests. If no test provoked the leak, then ASAN can’t catch what wasn’t happening. How was this leak found originally?

arno-lunarg:
@charles-lunarg I was just looking at the function and noticed the problem. 
With the first change I made, I just added `FreePnextChain(header->pNext)` before the `delete(header)` call, causing ASAN to catch a subsequent invalid read: after that `FreePnextChain` call, I also need to set `header->pNext` to `nullptr` to actually clear the list.
My point is, this problem ASAN caught shows that the tests are going through this code, yet ASAN could not detect the memory leak, for whatever reason. Maybe because some kind of global `free` is called before exiting the tests? I don't know. 

But I know this PR fixes a leak, and I do not think I can write a test that proves that there is no more leak?

charles-lunarg (changes requested):
Just looked at this safe structs code, I am unsure why you believe this fixes a memory leak.

Looking at the destructor for the safe struct objects, they all have `FreePnextChain(pNext);` in them. That means that this PR is more or less redundant because its manually calling `FreePnextChain(pNext)` when it would of been called in the destructor anyways. 

Granted, there *may* be a few structs that should have `FreePnextChain()` and are missing it, and I'd like to see that fixed. This PR looks more like a bandaid than a proper fix.

arno-lunarg:
Oh yes I see know, I got bamboozled by the destruction logic. When I think of destruction of a linked list, this way of doing things is not what comes to mind.
> there may be a few structs that should have FreePnextChain() and are missing it 
Do you have any examples of that? I can have a look

PR #7095, opened 2023-12-06
scripts: Move TSAN blacklist into C++ code instead of script

juan-lunarg:
Currently all of these tests aren't being run at all despite only having issues with thread sanitizer.

### DIFF
--- a/tests/unit/command.cpp
+++ b/tests/unit/command.cpp
@@ -483,6 +483,10 @@ TEST_F(NegativeCommand, NoBeginCommandBuffer) {
 }
 
 TEST_F(NegativeCommand, SecondaryCommandBufferRerecordedExplicitReset) {
+#if defined(VVL_ENABLE_TSAN)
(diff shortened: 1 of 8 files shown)


tests/unit/command.cpp
```diff
@@ -483,6 +483,10 @@ TEST_F(NegativeCommand, NoBeginCommandBuffer) {
 }
 
 TEST_F(NegativeCommand, SecondaryCommandBufferRerecordedExplicitReset) {
```
spencer-lunarg, line 485: so normally we do

```
// https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5965
TEST_F(NegativeCommand, DISABLED_SecondaryCommandBufferRerecordedExplicitReset) {
```

as it makes it easier to track how many things we have to fix

spencer-lunarg (changes requested):

tests/unit/command.cpp
spencer-lunarg: aww, fair, then looks good

spencer-lunarg (approved):

PR #7096, opened 2023-12-06
scripts: Remove sType workaround

juan-lunarg:
Now that VK_STRUCTURE_TYPE_LAYER_SETTINGS_CREATE_INFO_EXT is an official sType we no longer need this workaround in the codegen.


spencer-lunarg (approved):
wow, would have normally been left in here for years, thanks for removing

PR #7098, opened 2023-12-07
docs: Update layer configuration documentation

christophe-lunarg:
I only deleted some documentation that I believe was duplicated.

I'll add somewhere that `VK_EXT_validation_features` is deprecated.


spencer-lunarg (approved):
LGTM - @christophe-lunarg is there a reason this is still a draft?

juan-lunarg (approved):

PR #7100, opened 2023-12-07
tests: Fix core validation errors in GPU-AV tests

jeremyg-lunarg:
Note that vkt::CommandBuffer::begin() no longer sets VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT by default, but it is instead added in the places it is needed. This mostly is in arm best practices because of
 `UNASSIGNED-BestPractices-vkBeginCommandBuffer-one-time-submit`.


Tony-LunarG (approved):
LGTM

PR #7103, opened 2023-12-07
gpu: Don't wait for idle after every submit

jeremyg-lunarg:
Add virtual methods to the CommandBuffer and Queue classes that get run at PreRecord, PostRecord and Retire time. This makes it possible to remove the PostRecord wait for idle from GPU-AV submissions. Instead, at timeline semaphore is set up for host waits in the Retire() virtual method so that GPU-AV post processing for each submission is appropriately synchronized.


layers/state_tracker/queue_state.h
```diff
@@ -90,7 +91,8 @@ class Queue: public BASE_NODE {
 
     VkQueue VkHandle() const { return handle_.Cast<VkQueue>(); }
 
-    uint64_t Submit(QueueSubmission &&submission);
+    virtual uint64_t PreSubmit(std::vector<QueueSubmission> &&submissions);
```
spencer-lunarg, line 94: it wasn't obvious at first that `Pre` and `Post` here are referencing to `PreCallRecord`/`PostCallRecord`

I am still trying to understand which responsibilities we are splitting between the two

jeremyg-lunarg: in the original code, this all happened in PostCallRecordQueueSubmit(). The PreSubmit stuff is what happened before the  barrier and WaitIdle(). The PostSubmit() stuff is the barrier. Retire() is what happened after the WaitIdle()

PR #7106, opened 2023-12-08
tests: Remove unused functions from VkImageObj

spencer-lunarg:
`VkImageObj` is the last remaining object we have both in binding.cpp and render.cpp and it is a mess currently

This is a first step of cleaning it up by removing things that are not being used anymore to make the migration process a bit nicer


spencer-lunarg:
this didn't work as expected, going to try in future

PR #7108, opened 2023-12-08
tests: Create common Framebuffer init

spencer-lunarg:
Almost all use of `vkt::Framebuffer` in our tests use a simple way to init the framebuffer. This provides one more overload to reduce a lot of boilerplate and make it easier for places that just need a simple `VkFramebuffer`


artem-lunarg (approved):
LGTM! Very nice improvement.

Issue #7010, opened 2023-11-24
Missing `VkImage` object info in QueueSubmit synchronization validation error


glebov-andrey:
**Environment:**
 - VVL revision: `7adefb0aeb9034bb791437afb7ad447e0cbe5931` (current `main`)
 - Options enabled: synchronization with `VALIDATION_CHECK_ENABLE_SYNCHRONIZATION_VALIDATION_QUEUE_SUBMIT`

**Describe the Issue**

When `VALIDATION_CHECK_ENABLE_SYNCHRONIZATION_VALIDATION_QUEUE_SUBMIT` is enabled I'm getting validation errors about WAW hazards between `SYNC_IMAGE_LAYOUT_TRANSITION` usages, but the error message doesn't include the `VkImage`'s handle or debug name.

Here is an example message:
`Validation Error: [ SYNC-HAZARD-WRITE-AFTER-WRITE ] Object 0: handle = 0x1f69f5feb00, name = Graphics queue, type = VK_OBJECT_TYPE_QUEUE; | MessageID = 0x5c0ec5d6 | vkQueueSubmit2():  Hazard WRITE_AFTER_WRITE for entry 0, VkCommandBuffer 0x1f6b5fb65b0[], Recorded access info (recorded_usage: SYNC_IMAGE_LAYOUT_TRANSITION, command: vkCmdPipelineBarrier2, seq_no: 1, reset_no: 1). Access info (prior_usage: SYNC_IMAGE_LAYOUT_TRANSITION, write_barriers: 0, ).`

The `VkDebugUtilsMessengerCallbackDataEXT` also contains only one object - the `VkQueue`.

Also it looks like there's a formatting bug in `Access info` with the trailing `, `.

**Expected behavior**

The message (and callback data) contains the object info for the `VkImage` or `VkBuffer` which was specified in the pipeline barrier.

**Additional context**

<details>
<summary>Stack trace</summary>

```
[VkLayer_khronos_validation.dll] debug_log_msg(const _debug_report_data *, unsigned int, const LogObjectList &, const char *, const char *, const char *) logging.cpp:211
[VkLayer_khronos_validation.dll] LogMsg(const _debug_report_data *, unsigned int, const LogObjectList &, const Location *, basic_string_view<char, std::char_traits<char> >, const char *, char *) logging.cpp:466
[...]

PR #7109, opened 2023-12-08
layers: Use 32bit size type for handle vector instead of 8bit

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7010
Addresses this comment https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7010#issuecomment-1846751346 that  for barriers with many objects we have limitation of 8 bit type used to define size type of small vector.

The size of `small_vector<>` has not changed due to alignment gaps we had when used uint8_t (stats for msvc compiler):
uint8_t: 88 bytes
uint16_t 88 bytes
uint32_t 88 bytes

### DIFF
--- a/tests/unit/sync_val.cpp
+++ b/tests/unit/sync_val.cpp
@@ -5696,3 +5696,84 @@ TEST_F(NegativeSyncVal, AvailabilityWithoutVisibilityForBuffer) {
     m_errorMonitor->VerifyFound();
     m_commandBuffer->end();
 }
+
+TEST_F(NegativeSyncVal, ImageCopyHazardsLayoutTransition) {
+    TEST_DESCRIPTION("Copy to image and then start image layout transition without waiting for copy to end.");
+    RETURN_IF_SKIP(InitSyncValFramework());
+    RETURN_IF_SKIP(InitState());
+    vkt::Buffer buffer(*m_device, 64 * 64 * 4, VK_BUFFER_USAGE_TRANSFER_SRC_BIT);
+
+    const VkImageUsageFlags usage = VK_IMAGE_USAGE_TRANSFER_DST_BIT | VK_IMAGE_USAGE_SAMPLED_BIT;
+    VkImageObj image(m_device);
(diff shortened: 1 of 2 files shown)


tests/unit/sync_val.cpp
```diff
@@ -5696,3 +5696,84 @@ TEST_F(NegativeSyncVal, AvailabilityWithoutVisibilityForBuffer) {
     m_errorMonitor->VerifyFound();
     m_commandBuffer->end();
 }
+
+TEST_F(NegativeSyncVal, ImageCopyHazardsLayoutTransition) {
```
artem-lunarg, line 5700: This (first) test I added for completeness - to have a minimal test that reproduces a common scenario. Used it as starting point for the second test that reproduces actual issue.

tests/unit/sync_val.cpp
```diff
@@ -5696,3 +5696,84 @@ TEST_F(NegativeSyncVal, AvailabilityWithoutVisibilityForBuffer) {
...
+    // hazards with image layout transition writes
+    vk::CmdPipelineBarrier(*m_commandBuffer, VK_PIPELINE_STAGE_TRANSFER_BIT, VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT, 0, 0, nullptr, 0,
+                           nullptr, 1, &transition);
+    m_errorMonitor->VerifyFound();
+    m_commandBuffer->end();
+}
+
+// https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7010#issuecomment-1846751346
+TEST_F(NegativeSyncVal, TestMessageReportingWithManyBarriers) {
+    TEST_DESCRIPTION("Hazardous pipeline barrier contains many barrier structures (> 255). Test that reporting can handle this");
```
filnet, line 5735: nit: missing period at end of description.

artem-lunarg: Fixed.

spencer-lunarg (approved):

PR #7110, opened 2023-12-08
tests: Remove bloat from  InitRenderTarget

spencer-lunarg:
More progress towards the seemly endless cleanup of `InitRenderTarget`

instead of making the function more complex, the few tests that need one-off stuff use other helpers


tests/unit/arm_best_practices.cpp
```diff
@@ -1409,9 +1425,6 @@ TEST_F(VkArmBestPracticesLayerTest, BlitImageLoadOpLoad) {
     RETURN_IF_SKIP(InitBestPracticesFramework(kEnableArmValidation));
     InitState();
 
-    m_load_op_clear = false;  // Force LOAD_OP_LOAD
-    InitRenderTarget();
```
artem-lunarg, line 1413: Is `InitRenderTarget` called somewhere else?

artem-lunarg (approved):
LGTM. Very nice change with removing all this state from the test framework.

PR #7111, opened 2023-12-08
layers: Fix check for drawing with dynamic viewport with count and scissor with count

### LINKED ISSUE #7107: VUID-03149 Does not perform complete checks for viewport and scissor states.
**Environment:**
 - OS: Windows 11
 - GPU and driver version: NVIDIA RTX 3090, 537.34
 - SDK or header version if building from repo: 1.3.239.0
 - Options enabled (synchronization, best practices, etc.): Sync, Best Practices, GPU Assisted

**Describe the Issue**

VUID-03419 appears to not be complete in its checks.  If `vkCmdSetViewport` or `vkCmdSetScissor` is called and the pipeline was created with `VK_DYNAMIC_STATE_VIEWPORT_WITH_COUNT` or `VK_DYNAMIC_STATE_SCISSOR_WITH_COUNT` dynamic states, validation layers should emit an error. Currently, there is no output.

**Expected behavior**

I expect a validation error that `vkCmdSetViewport/ScissorWithCount` was not called, but a draw call was issued with the bound pipeline using the `_WITH_COUNT` dynamic states.

**Valid Usage ID**
VUID-03419 (error was not emitted, but checks exist in the validation layer for mismatch scissor and viewport count)

Untested, but I believe 03147 and 03148 to be affected by the same issue.

ziga-lunarg:
Closes #7107

### DIFF
--- a/layers/core_checks/cc_cmd_buffer_dynamic.cpp
+++ b/layers/core_checks/cc_cmd_buffer_dynamic.cpp
@@ -681,7 +681,7 @@ bool CoreChecks::ValidateDrawDynamicStatePipeline(const LastBound& last_bound_st
         if (dyn_viewport_count && !dyn_scissor_count) {
             const auto required_viewport_mask = (1 << viewport_state->scissorCount) - 1;
             const auto missing_viewport_mask = ~cb_state.viewportWithCountMask & required_viewport_mask;
-            if (missing_viewport_mask) {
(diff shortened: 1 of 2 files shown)


ziga-lunarg:
> Thanks so much for getting a fix out this quickly. Should the same change be applied for 03147/03148 as well? At a glance, it appears that they both need the check applied.

I don't think it's possible to hit those other 2 VUIDs with this case, without hitting another VUID before. The case with 03149 was if both states were dynamic and neither `vkCmdSetViewportWithCount` nor `vkCmdSetScissorWithCount` were called.

spencer-lunarg (approved):

PR #7112, opened 2023-12-08
ci: Update to macos-12 for iOS build

juan-lunarg:
CMake 3.28 is having issues with the macos-11 runner.


artem-lunarg (approved):

juan-lunarg:
@KhronosGroup/vvl-codeowners rebase to avoid iOS CI failures

PR #7113, opened 2023-12-08
tests: Use optimal image layout by default

artem-lunarg:
That was the majority of cases in the tests but also makes sense because optimal images are the most often used in 3d apps (due to poor mapping of linear images to gpu hardware which results in  the absence mips, limited set of supported formats, etc.).

Apologies if this conflicts with existing PRs that also modify image functionality. Lets hope it's not that bad.


spencer-lunarg (approved):
> Apologies if this conflicts with existing PRs that also modify image functionality. Lets hope it's not that bad.

Honestly the `Image` test class is in such need of help I feel it is totally acceptable!

PR #7118, opened 2023-12-10
layers: Warn if pipeline is rebound when dynamic state was set

### LINKED ISSUE #7008: Add warning to applications that modify state between pipeline rebind of the same pipeline object, where those states are not dynamic
As discussed in the internal issue (https://gitlab.khronos.org/vulkan/vulkan/-/issues/3675) and covered by the new CTS (https://gitlab.khronos.org/Tracker/vk-gl-cts/-/issues/4642), we have determined that several implementations have a bug with reapplying static pipeline state when rebinding the same pipeline object. The Vulkan working group would like a new VVL warning to be raised if an application uses the problematic sequence and the implementation does not advertise passing conformance version 1.3.8.0 or later in `VkPhysicalDeviceDriverProperties.conformanceVersion`.

The problematic sequence is:
```c++
// create "pso" with at least some static state, for example primitiveTopology=TRIANGLE_STRIP
vkCmdBindPipeline(pso)
vkCmdSetPrimitiveTopology(TRIANGLE_LIST)
vkCmdBindPipeline(pso)
vkCmdDraw() << this should draw with TRIANGLE_STRIP and not TRIANGLE_LIST
```

The specification currently allows this sequence, but it turns out several implementation use a trivial optimization such that if they see the same pipeline being rebound then they skip over applying their static state to the command buffer. The implementations need to be updated to either remove this optimization, or track the dirty state properly.

Until implementations are updated it would be most useful to applications to be warned if they hit this sequence. Since there is new CTS coverage which will be part of vulkan-cts-1.3.8.0, the warning only needs to be issued for implementations that don't pass that new version of CTS.

The VVL code can be roughly described as:
- Add a new boolean in the command buffer state tracker to indicate `bool dirtyStaticState`, which is initialized to false.
- In every `vkCmdSet*` command, check if there is a graphics pipeline bound, and if so, set dirtyStaticState to true if this vkCmdSet command affects state in the bound graphics pipeline that is not dynamic state.
- When `vkCmdBindPipeline(VK_PIPELINE_BIND_POINT_GRAPHICS)` is called, and the `VkPipeline` handle value is identical to the currently bound graphics pipeline handle, and `dirtyStaticState` is true, then issue the new WARNING if the implementation does not report a `VkPhysicalDeviceDriverProperties.conformanceVersion` equal or greater than 1.3.8.0. Clear the `dirtyStaticState` state.

ziga-lunarg:
Closes #7008

### DIFF
--- a/layers/core_checks/cc_pipeline.cpp
+++ b/layers/core_checks/cc_pipeline.cpp
@@ -25,6 +25,20 @@
 #include "generated/chassis.h"
 #include "core_validation.h"
 
+bool CoreChecks::IsBeforeCtsVersion(uint32_t major, uint32_t minor, uint32_t subminor) const {
+    // If VK_KHR_driver_properties is not enabled then conformance version will not be set
+    if (phys_dev_props_core12.conformanceVersion.major == 0) {
+        return false;
+    }
+    if (phys_dev_props_core12.conformanceVersion.major != major) {
+        return phys_dev_props_core12.conformanceVersion.major < major;
+    }
+    if (phys_dev_props_core12.conformanceVersion.minor != minor) {
+        return phys_dev_props_core12.conformanceVersion.minor < minor;
+    }
+    return phys_dev_props_core12.conformanceVersion.subminor < subminor;
+}
+
 bool CoreChecks::ValidatePipelineCacheControlFlags(VkPipelineCreateFlags2KHR flags, const Location &loc, const char *vuid) const {
     bool skip = false;
     if (enabled_features.pipelineCreationCacheControl == VK_FALSE) {
@@ -354,6 +368,21 @@ bool CoreChecks::PreCallValidateCmdBindPipeline(VkCommandBuffer commandBuffer, V
                         }
                     }
                 }
+            }
+
+            if (pipelineBindPoint == VK_PIPELINE_BIND_POINT_GRAPHICS && cb_state->GetCurrentPipeline(pipelineBindPoint) &&
+                pipeline == cb_state->GetCurrentPipeline(pipelineBindPoint)->pipeline() && cb_state->dirtyStaticState &&
+                IsBeforeCtsVersion(1, 3, 8)) {
+                const LogObjectList objlist(commandBuffer, pipeline);
+                // This catches a bug in some drivers with conformance version lower than 1.3.8
+                // https://gitlab.khronos.org/vulkan/vulkan/-/issues/3675
+                // https://gitlab.khronos.org/Tracker/vk-gl-cts/-/issues/4642
+                skip |= LogError(
+                    "UNASSIGNED-vkCmdBindPipeline-Pipeline-Rebind", objlist, error_obj.location,
+                    "The pipeline being bound (%s) is the same as the currently bound pipeline and between the calls, a "
+                    "dynamic state was set which is static in this pipeline. This might not work correctly on drivers with "
+                    "conformance version lower than 1.3.8.0.",
+                    FormatHandle(pipeline).c_str());
             }
         } else if (pipelineBindPoint == VK_PIPELINE_BIND_POINT_RAY_TRACING_KHR) {
             if (!cb_state->unprotected) {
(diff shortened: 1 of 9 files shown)


layers/best_practices/bp_pipeline.cpp
```diff
@@ -669,5 +669,16 @@ bool BestPractices::PreCallValidateCmdBindPipeline(VkCommandBuffer commandBuffer
         }
     }
 
+    if (pipelineBindPoint == VK_PIPELINE_BIND_POINT_GRAPHICS && cb->GetCurrentPipeline(pipelineBindPoint) &&
+        pipeline == cb->GetCurrentPipeline(pipelineBindPoint)->pipeline() && cb->dirtyStaticState &&
+        phys_dev_props_core12.conformanceVersion.major == 1 && phys_dev_props_core12.conformanceVersion.minor < 7) {
+        const LogObjectList objlist(commandBuffer, pipeline);
+        skip |= LogWarning(kVUID_BestPractices_Pipeline_Rebind, objlist, error_obj.location,
```
spencer-lunarg, line 676: I think this is something worth putting in the normal core layers. This is a case where we can reference https://gitlab.khronos.org/vulkan/vulkan/-/issues/3675 and https://gitlab.khronos.org/Tracker/vk-gl-cts/-/issues/4642 in a comment above why this warning belongs there

layers/best_practices/bp_pipeline.cpp
```diff
@@ -669,5 +669,16 @@ bool BestPractices::PreCallValidateCmdBindPipeline(VkCommandBuffer commandBuffer
         }
     }
 
+    if (pipelineBindPoint == VK_PIPELINE_BIND_POINT_GRAPHICS && cb->GetCurrentPipeline(pipelineBindPoint) &&
+        pipeline == cb->GetCurrentPipeline(pipelineBindPoint)->pipeline() && cb->dirtyStaticState &&
+        phys_dev_props_core12.conformanceVersion.major == 1 && phys_dev_props_core12.conformanceVersion.minor < 7) {
```
spencer-lunarg, line 674: we should have a util function like `IsBeforeCtsVersion(1, 3, 8)` (open to better names) that we can use. There it can also handle checking if `VK_KHR_driver_properties` is enabled or not as well... this might not be the last time we need add a check like this

layers/core_checks/cc_pipeline.cpp
```diff
@@ -355,6 +369,21 @@ bool CoreChecks::PreCallValidateCmdBindPipeline(VkCommandBuffer commandBuffer, V
...
+            if (pipelineBindPoint == VK_PIPELINE_BIND_POINT_GRAPHICS && cb_state->GetCurrentPipeline(pipelineBindPoint) &&
+                pipeline == cb_state->GetCurrentPipeline(pipelineBindPoint)->pipeline() && cb_state->dirtyStaticState &&
+                IsBeforeCtsVersion(1, 3, 8)) {
+                const LogObjectList objlist(commandBuffer, pipeline);
+                // This catches a bug in some drivers with conformance version lower than 1.3.8
+                // https://gitlab.khronos.org/vulkan/vulkan/-/issues/3675
+                // https://gitlab.khronos.org/Tracker/vk-gl-cts/-/issues/4642
+                skip |= LogError(
+                    "UNASSIGNED-vkCmdBindPipeline-Pipeline-Rebind", objlist, error_obj.location,
+                    "The pipeline being bound (%s) is the same as the currently bound pipeline and between the calls a "
```
spencer-lunarg, line 382:
```suggestion
                    "The pipeline being bound (%s) is the same as the currently bound pipeline and between the calls, a "
```

tests/unit/dynamic_state.cpp
```diff
@@ -5818,3 +5818,34 @@ TEST_F(NegativeDynamicState, MissingScissorWithCount) {
     m_commandBuffer->EndRenderPass();
     m_commandBuffer->end();
 }
+
+TEST_F(NegativeDynamicState, RebindSamePipeline) {
```
spencer-lunarg, line 5822: this test will need to skip if it has a 1.3.8+ conformance version

spencer-lunarg (approved):
LGTM - just fix the test to check for version and small grammar nit

Issue #7116, opened 2023-12-08
VK_EXT_calibrated_timestamps promotion to KHR


mikes-lunarg:
VK_EXT_calibrated_timestamps has been been promoted to KHR in #7115

Core validation checks for vkGetCalibratedTimestampsKHR need to be hooked up and possibly tested.

tracker - https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7116

PR #7119, opened 2023-12-11
layers: Add VK_KHR_calibrated_timestamps support

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7116

### DIFF
--- a/tests/unit/others.cpp
+++ b/tests/unit/others.cpp
@@ -2851,8 +2851,7 @@ TEST_F(VkLayerTest, GetCalibratedTimestampsDuplicate) {
     TEST_DESCRIPTION("vkGetCalibratedTimestampsEXT with duplicated timeDomain.");
     SetTargetApiVersion(VK_API_VERSION_1_1);
     AddRequiredExtensions(VK_EXT_CALIBRATED_TIMESTAMPS_EXTENSION_NAME);
-    RETURN_IF_SKIP(InitFramework());
-    RETURN_IF_SKIP(InitState());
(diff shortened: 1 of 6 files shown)


tests/unit/others.cpp
```diff
@@ -2872,12 +2871,35 @@ TEST_F(VkLayerTest, GetCalibratedTimestampsDuplicate) {
...
 }
 
+TEST_F(VkLayerTest, GetCalibratedTimestampsDuplicateKHR) {
+    TEST_DESCRIPTION("vkGetCalibratedTimestampsKHR with duplicated timeDomain.");
+    SetTargetApiVersion(VK_API_VERSION_1_1);
+    AddRequiredExtensions(VK_KHR_CALIBRATED_TIMESTAMPS_EXTENSION_NAME);
+    RETURN_IF_SKIP(Init());
+
+    uint32_t count = 0;
+    vk::GetPhysicalDeviceCalibrateableTimeDomainsKHR(gpu(), &count, nullptr);
```
artem-lunarg, line 2881: Is it possible to get `count` as zero? If yes the test should be skipped.

artem-lunarg (approved):
LGTM. One question.

tests/unit/others.cpp
spencer-lunarg: this is not a core extension, so any driver that would expose the extension, but make the only query function return nothing, would be quite scamful... also hope there is CTS for that

PR #7120, opened 2023-12-11
Fix pipelineStatistics check and test cleanup

spencer-lunarg:
This adds checks that `pipelineStatistics` can't be zero (it is not a valid `VkQueryPipelineStatisticFlags` value)

Also improves how `vkt::QueryPool` are created to make it simpler to make


layers/stateless/sl_descriptor.cpp
```diff
@@ -989,14 +989,17 @@ bool StatelessValidation::manual_PreCallValidateCreateQueryPool(VkDevice device,
...
-    if ((pCreateInfo->queryType == VK_QUERY_TYPE_PIPELINE_STATISTICS) && (pCreateInfo->pipelineStatistics != 0) &&
-        ((pCreateInfo->pipelineStatistics & (~AllVkQueryPipelineStatisticFlagBits)) != 0)) {
-        skip |= LogError("VUID-VkQueryPoolCreateInfo-queryType-00792", device, create_info_loc.dot(Field::queryType),
-                         "is VK_QUERY_TYPE_PIPELINE_STATISTICS, but "
-                         "pCreateInfo->pipelineStatistics must be a valid combination of VkQueryPipelineStatisticFlagBits "
-                         "values.");
+
+    if ((pCreateInfo->queryType == VK_QUERY_TYPE_PIPELINE_STATISTICS)) {
+        if (pCreateInfo->pipelineStatistics == 0) {
+            skip |= LogError("VUID-VkQueryPoolCreateInfo-queryType-00792", device, create_info_loc.dot(Field::queryType),
```
artem-lunarg, line 995: I find it's a not exactly obvious that `pipelineStatistics` can't be zero, or at least it requires some efforts to understand this. I'd better to write it down in case I misunderstood some parts.

> VUID-VkQueryPoolCreateInfo-queryType-00792
If queryType is VK_QUERY_TYPE_PIPELINE_STATISTICS, pipelineStatistics **must be a valid combination of [VkQueryPipelineStatisticFlagBits](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkQueryPipelineStatisticFlagBits)** values

Then section 3.7.2 includes zero in valid combination:
> Any Vk*Flags member or parameter used in the API as an input must be a valid combination of bit flags. **A valid combination is** either zero or the bitwise OR of valid bit flags.

Then section 18.4:
>  **At least one statistic counter** relevant to the operations supported on the recording command buffer must be enabled.

I guess the last one forbids zero, but it does not follow directly from 00792 which mentions only *A valid combination* of flags.

artem-lunarg (approved):
LGTM. One comment about zeros.

Issue #7123, opened 2023-12-11
vkRegisterDisplayEventEXT reporting display was created, allocated or retrieved from VkPhysicalDevice 0x0


jparimaa:
**Environment:**
 - OS: Ubuntu 20.04.6 LTS aarch64
 - GPU and driver version: 
 - SDK or header version if building from repo: https://github.com/KhronosGroup/Vulkan-ValidationLayers/releases/tag/v1.3.267 and later
 - Options enabled (synchronization, best practices, etc.): See the attached source code
 - Source: [vkcube.zip](https://github.com/KhronosGroup/Vulkan-ValidationLayers/files/13633056/vkcube.zip)
 
**Describe the Issue**

Calling `vkRegisterDisplayEventEXT` prints a warning that says

> ERROR : VALIDATION - Message Id Number: -1462709463 | Message Id Name: VUID-vkRegisterDisplayEventEXT-commonparent
> 	Validation Error: [ VUID-vkRegisterDisplayEventEXT-commonparent ] Object 0: handle = 0xaaaaaaef5c30, type = VK_OBJECT_TYPE_INSTANCE; Object 1: handle = 0xaaaaac9771d0, type = VK_OBJECT_TYPE_PHYSICAL_DEVICE; | MessageID = 0xa8d0d329 | vkRegisterDisplayEventEXT(): display (VkDisplayKHR 0xfab64d0000000002) was created, allocated or retrieved from VkPhysicalDevice 0x0[], but command is using (or its dispatchable parameter is associated with) VkPhysicalDevice 0xaaaaac9771d0[]. The Vulkan spec states: Both of device, and display must have been created, allocated, or retrieved from the same VkPhysicalDevice (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkRegisterDisplayEventEXT-commonparent)
> 
> 	Objects - 2
> 		Object[0] - VK_OBJECT_TYPE_INSTANCE, Handle 0xaaaaaaef5c30
> 		Object[1] - VK_OBJECT_TYPE_PHYSICAL_DEVICE, Handle 0xaaaaac9771d0

I think display (VkDisplayKHR 0xfab64d0000000002) was NOT created with VkPhysicalDevice 0x0

**Expected behavior**

I suspect this might be a false positive. I wasn't expecting an error message.

**Additional context**

Important part of the source code:
```
[...]

PR #7124, opened 2023-12-11
layers: Add VkDisplayKHR support for ObjectTracker

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7123

### DIFF
--- a/layers/object_tracker/object_tracker_utils.cpp
+++ b/layers/object_tracker/object_tracker_utils.cpp
@@ -57,9 +57,20 @@ bool ObjectLifetimes::CheckObjectValidity(uint64_t object_handle, VulkanObjectTy
         const auto lifetimes = other_device_data.second->GetValidationObject<ObjectLifetimes>();
         if (lifetimes && lifetimes != this && lifetimes->TracksObject(object_handle, object_type)) {
             other_lifetimes = lifetimes;
+
+            // Sometimes (calls such as vkRegisterDisplayEventEXT) interact with both the device and physical device
+            if (parent_type == kVulkanObjectTypePhysicalDevice) {
+                auto iter = other_lifetimes->object_map[object_type].find(object_handle);
+                if (iter != other_lifetimes->object_map[object_type].end()) {
+                    if (iter->second->parent_object == HandleToUint64(physical_device)) {
+                        return skip;
+                    }
+                }
+            }
             break;
         }
     }
+
     // Object was not found anywhere
     if (!other_lifetimes) {
         return LogError(invalid_handle_vuid, instance, loc, "Invalid %s Object 0x%" PRIxLEAST64 ".", object_string[object_type],
@@ -69,6 +80,7 @@ bool ObjectLifetimes::CheckObjectValidity(uint64_t object_handle, VulkanObjectTy
     if (wrong_parent_vuid == kVUIDUndefined) {
         return skip;
     }
+
     // Object found on another device
     LogObjectList objlist;
     std::string handle_str;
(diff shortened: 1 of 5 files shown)


tests/unit/wsi_positive.cpp
```diff
@@ -1208,3 +1208,27 @@ TEST_F(PositiveWsi, CreateSwapchainWithPresentModeInfo) {
 
     vk::CreateSwapchainKHR(device(), &swapchain_create_info, nullptr, &m_swapchain);
 }
+
+TEST_F(PositiveWsi, RegisterDisplayEvent) {
```
spencer-lunarg, line 1212: will work with MockICD once `https://github.com/KhronosGroup/Vulkan-Tools/pull/936` is merged and we update `known_good.json` (but tested locally)

tests/unit/parent.cpp
```diff
@@ -529,10 +528,50 @@ TEST_F(NegativeParent, PhysicalDevice_Display) {
     // display from a different physical device
     uint32_t mode_count = 0;
     m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-vkGetDisplayModePropertiesKHR-display-parent");
+    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-vkRegisterDisplayEventEXT-commonparent");
     vk::GetDisplayModePropertiesKHR(gpu(), display, &mode_count, nullptr);
```
artem-lunarg, line 532: In the process of refreshing memory how it all works. Meanwhile, should we get `VUID-vkRegisterDisplayEventEXT-commonparent` here?

layers/object_tracker/object_tracker_utils.cpp
```diff
@@ -988,14 +1000,28 @@ bool ObjectLifetimes::PreCallValidateGetPhysicalDeviceDisplayPropertiesKHR(VkPhy
     return skip;
 }
 
+void ObjectLifetimes::AllocateDisplayKHR(VkPhysicalDevice physical_device, VkDisplayKHR display, const Location &loc) {
+    auto iter = object_map[kVulkanObjectTypeDisplayKHR].find(HandleToUint64(display));
+    if (iter == object_map[kVulkanObjectTypeDisplayKHR].end()) {
```
artem-lunarg, line 1005: Do we need to report error if object already exists, similar to `InsertObject`?

tests/unit/parent.cpp
spencer-lunarg: um... that was a mistake, I didn't save, this should be removed (why CI is failing)

layers/object_tracker/object_tracker_utils.cpp
```diff
@@ -69,6 +69,18 @@ bool ObjectLifetimes::CheckObjectValidity(uint64_t object_handle, VulkanObjectTy
     if (wrong_parent_vuid == kVUIDUndefined) {
         return skip;
     }
+
+    // Sometimes (calls such as vkRegisterDisplayEventEXT) interact with both the device and physical device
+    if (parent_type == kVulkanObjectTypePhysicalDevice) {
```
artem-lunarg, line 74: I feel logic (similar or modified) should be at the beginning.

Initially we have 
```
if (TracksObject(object_handle, object_type)) {
        return skip;
    }
```
which on return signals everything is okay. The following code signals error and even if we have exception (anonimous objects, we have it there because can't move up due to logic dependencies). 

If the first `TracksObject` does not detect valid situation (object created from instance table but dispatched from device table), then if possible we should do the check before that first `TracksObject` - detect special case and call `TracksObject` from proper ObjectLifetimes instance. 

The idea is to have this:

```
ObjectLifetimes::CheckObjectValidity()
{
// GetAssociatedObjectLifetimes - returns  current lifetimes object is not special case or does search and returns
lifetimes from instance table for specific instance/physical_device. Also in this function can be other special cases if needed.
  ObjectLifetimes* associated_object_filestimes = GetAssociatedObjectLifetimes(params...); 

  // Now all code is the same as we have now with the exception we use associated_object_filestimes  in one place
 // If this instance of lifetime validation tracks the object, report success
    if (associated_object_filestimes ->TracksObject(object_handle, object_type)) {
        return skip;
    }
   ....
}

Comment (spencer-lunarg) [reply layers/object_tracker/object_tracker_utils.cpp:74]: I see your point... I guess I view this as

```
CheckObjectValidity() {

  // do common check
  if (TracksObject(object_handle, object_type)) { }
  
  // Object not found, look for it in other device object maps
  for (const auto &other_device_data : layer_data_map) {
    if (lifetimes->TracksObject(object_handle, object_type)) {
            other_lifetimes = lifetimes;
            // MOVE check here, this is were we see if it is still valid
            break;
        }
  }
  
  // everything from here is an error
  ```
  
  thoughts?

artem-lunarg: Yes this looks good.

Issue #7093, opened 2023-12-06
False positive SYNC-HAZARD-WRITE-AFTER-READ on vkCmdDispatch with writeonly SSBO


haasn:
**Environment:**
 - OS: Linux xor 6.5.8-1-default #1 SMP PREEMPT_DYNAMIC Fri Oct 20 05:40:59 UTC 2023 (51baea8) x86_64 x86_64 x86_64 GNU/Linux
 - GPU and driver version: AMD Navi 31 [Radeon RX 7900 XT/7900 XTX], amdgpu RADV Mesa 23.2.1
 - SDK or header version if building from repo: vvl 1.3.268.0-2.1
 - Options enabled (synchronization, best practices, etc.): sync + bp + gpu assisted

**Describe the Issue**

The synchronization validation thinks that access from a shader marked 'writeonly' is a conflicting *read* access. My code does, basically, the following:

0. Create vulkan device, compute pipeline, buffers bufA and bufB, etc.
1. Setup DS ds0 with DESC_IN = bufA, DESC_OUT = bufB, binding respectively the whole buffer (VK_WHOLE_SIZE)
2. vkBeginCommandBuffer, vkCmdBindPipeline, vkCmdBindDescriptorSets(ds0), vkCmdPushConstants etc.
3. vkCmdDispatch()
4. vkCmdCopyBuffer(srcBuffer = bufA, dstBuffer = bufB)
5. vkCmdPipelineBarrier( srcStage = CS|TX, dstStage = CS|TX|HOST, srcAccess = TX_W|CS_W, dstAccess = TX_R|CS_R|HOST_R )

It is worth pointing out that the memory regions written to by steps 3 and steps 4 do not overlap.

**Expected behavior**

No validation layer is emitted, as the write does not conflict any prior sync. While I would understand a false positive about a possible conflict between the two *writes* in steps 3 and 4, I don't understand at all how the VVL thinks that I am conflicting with a prior *read*.

**Valid Usage ID**
The validation error is emitted on step 4:

```
[...]

PR #7128, opened 2023-12-11
layers: Report write access for writeonly resources

artem-lunarg:
Always detect access as WRITE when resource is marked as writeonly (the choice is between READ and WRITE).
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7093

It looks like for buffers `ResourceInterfaceVariable::is_written_to` might not detect write access. Do you know about this @spencer-lunarg ? My thinking was if we have `writeonly` then it's a reliable metric and should have higher priority comparing to acess detection that can produce produce false-positive, but it's still interesting to know if `is_written_to` can detect write as in the test WriteOnlyBufferWriteHazard.

### DIFF
--- a/layers/sync/sync_commandbuffer.cpp
+++ b/layers/sync/sync_commandbuffer.cpp
@@ -32,10 +32,18 @@ SyncStageAccessIndex GetSyncStageAccessIndexsByDescriptorSet(VkDescriptorType de
         return stage_accesses.uniform_read;
     }
 
+    // Detect if variable is nonreadable (writeonly in glsl).
+    // NOTE: is_written_to can be used as a more general case instead of adding is_writeonly logic.
+    // At first we need to fix is_written_to support for buffers.
+    bool is_writeonly = variable.decorations.Has(spirv::DecorationBase::nonreadable_bit);
+    if (variable.type_struct_info) {
+        is_writeonly |= variable.type_struct_info->decorations.AllMemberHave(spirv::DecorationBase::nonreadable_bit);
+    }
+
     // If the desriptorSet is writable, we don't need to care SHADER_READ. SHADER_WRITE is enough.
     // Because if write hazard happens, read hazard might or might not happen.
     // But if write hazard doesn't happen, read hazard is impossible to happen.
-    if (variable.is_written_to) {
+    if (variable.is_written_to || is_writeonly) {
         return stage_accesses.storage_write;
     } else if (descriptor_type == VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE ||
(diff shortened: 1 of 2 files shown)


tests/unit/sync_val.cpp
```diff
@@ -5799,3 +5799,101 @@ TEST_F(NegativeSyncVal, TestMessageReportingWithManyBarriers) {
...
+
+    //
+    m_errorMonitor->SetDesiredFailureMsg(VK_DEBUG_REPORT_ERROR_BIT_EXT, "SYNC-HAZARD-WRITE-AFTER-WRITE");
+    vk::CmdCopyBuffer(*m_commandBuffer, buf_a, buf_b, 1, &region);
+    m_errorMonitor->VerifyFound();
+
+    m_commandBuffer->end();
+}
+
+TEST_F(NegativeSyncVal, WriteOnlyImageWriteHazard) {
```
artem-lunarg, line 5853: Image test is added for completeness/regression detection. `ResourceInterfaceVariable::is_written_to` detects write access for `imageStore` scenario.

layers/sync/sync_commandbuffer.cpp
```diff
@@ -32,10 +32,17 @@ SyncStageAccessIndex GetSyncStageAccessIndexsByDescriptorSet(VkDescriptorType de
         return stage_accesses.uniform_read;
     }
 
+    // Detect writeonly image
+    bool is_writeonly = variable.decorations.Has(spirv::DecorationBase::nonreadable_bit);
```
spencer-lunarg, line 36: > It looks like for buffers ResourceInterfaceVariable::is_written_to might not detect write access.

So this is something I am aware of, there are some edge cases where I realized `is_written_to` was not being marked... it really should be fixed

my only worry about adding this is

1. There is a chance it is marked as NonReadable but never actually accessed (conditional logic) and causes a false positive
2. If I do fix the `is_written_to` then this will be here and people will be afraid to remove this logic (even though it is not needed)

... I can look into this later today and use these tests you made (as well as some dedicated SPIR-V ones) and maybe the real bug is in the `is_written_to` logic

spencer-lunarg: also would approve to add a `TODO` comment pointing to an issue and we can get this in so at least things are working as expected until we fix it the "proper way"

artem-lunarg: If both `is_written_to` and `is_read_from` are supported for buffers then we can use solution that is both a false we skip further analysis. In that case we won't have situation that for `writeonly` we report READ access. This will be a good change to have since it's more general. If we can also guarantee that we can't miss reads/writes reported through is_written_to\is_read_from then writeonly logic indeed is not necessary.

spencer-lunarg (approved):

PR #7129, opened 2023-12-11
syncval: Set queue submit time sync val to default on

### LINKED ISSUE #7127: syncval: Set VALIDATION_CHECK_ENABLE_SYNCHRONIZATION_VALIDATION_QUEUE_SUBMIT to be on by default
Now that this feature has been implemented for several months and SDK releases, it seems like it should be on by default.

jzulauf-lunarg:
Set queue submit time sync validation to be on by default.

Add disable setting for those that want to turn it off for performance.

Fixes #7127

### DIFF
--- a/layers/layer_options.cpp
+++ b/layers/layer_options.cpp
@@ -33,7 +33,6 @@ const char *SETTING_VALIDATE_BEST_PRACTICES_AMD = "validate_best_practices_amd";
 const char *SETTING_VALIDATE_BEST_PRACTICES_IMG = "validate_best_practices_img";
 const char *SETTING_VALIDATE_BEST_PRACTICES_NVIDIA = "validate_best_practices_nvidia";
 const char *SETTING_VALIDATE_SYNC = "validate_sync";
-const char *SETTING_VALIDATE_SYNC_QUEUE_SUBMIT = "sync_queue_submit";
 const char *SETTING_VALIDATE_GPU_BASED = "validate_gpu_based";
 const char *SETTING_RESERVE_BINDING_SLOT = "reserve_binding_slot";
 
@@ -49,6 +48,7 @@ const char *SETTING_UNIQUE_HANDLES = "unique_handles";
 const char *SETTING_OBJECT_LIFETIME = "object_lifetime";
 const char *SETTING_CHECK_SHADERS = "check_shaders";
 const char *SETTING_CHECK_SHADERS_CACHING = "check_shaders_caching";
+const char *SETTING_VALIDATE_SYNC_QUEUE_SUBMIT = "sync_queue_submit";
 
 const char *SETTING_MESSAGE_ID_FILTER = "message_id_filter";
 const char *SETTING_CUSTOM_STYPE_LIST = "custom_stype_list";
@@ -77,6 +77,9 @@ void SetValidationDisable(CHECK_DISABLED &disable_data, const ValidationCheckDis
             break;
         case VALIDATION_CHECK_DISABLE_IMAGE_LAYOUT_VALIDATION:
             disable_data[image_layout_validation] = true;
+            break;
+        case VALIDATION_CHECK_DISABLE_SYNCHRONIZATION_VALIDATION_QUEUE_SUBMIT:
+            disable_data[sync_validation_queue_submit] = true;
             break;
         default:
             assert(true);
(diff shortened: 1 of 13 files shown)


spencer-lunarg (approved):

layers/layer_options.cpp
```diff
@@ -137,9 +140,6 @@ void SetValidationEnable(CHECK_ENABLED &enable_data, const ValidationCheckEnable
             enable_data[vendor_specific_img] = true;
             enable_data[vendor_specific_nvidia] = true;
             break;
-        case VALIDATION_CHECK_ENABLE_SYNCHRONIZATION_VALIDATION_QUEUE_SUBMIT:
-            enable_data[sync_validation_queue_submit] = true;
-            break;
```
christophe-lunarg, line 142: It's not clear to me why stuff move around but that's ok with me. Just commenting to be sure that's not an accident,

christophe-lunarg: Oh, it went from an enable to a disable ok!

tests/unit/sync_val_positive.cpp
```diff
@@ -27,18 +27,8 @@ void VkSyncValTest::InitSyncValFramework(bool enable_queue_submit_validation) {
...
-        {OBJECT_LAYER_NAME, "enables", VK_LAYER_SETTING_TYPE_STRING_EXT, 1, kEnableQueuSubmitSyncValidation}};
-    // The pNext of qs_settings is modified by InitFramework that's why it can't
-    // be static (should be separate instance per stack frame). Also we show
-    // explicitly that it's not const (InitFramework casts const pNext to non-const).
-    VkLayerSettingsCreateInfoEXT qs_settings{VK_STRUCTURE_TYPE_LAYER_SETTINGS_CREATE_INFO_EXT, nullptr,
-                                             static_cast<uint32_t>(std::size(settings)), settings};
-    if (enable_queue_submit_validation) {
-        features_.pNext = &qs_settings;
-    }
+    // QueueSubmit Validation *always* on (default behavior)
```
christophe-lunarg, line 30: It sounds like to me we need to a test to check the "disable" code path is working? If it's always on, it's never tested,

artem-lunarg (approved):

tests/unit/sync_val_positive.cpp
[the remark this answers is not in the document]
jzulauf-lunarg: Cleaned this up...

jzulauf-lunarg: done

layers/layer_options.cpp
```diff
@@ -477,6 +475,8 @@ void ProcessConfigAndEnvSettings(ConfigAndEnvSettings *settings_data) {
         SetValidationSetting(layer_setting_set, settings_data->disables, object_tracking, SETTING_OBJECT_LIFETIME);
         SetValidationSetting(layer_setting_set, settings_data->disables, shader_validation, SETTING_CHECK_SHADERS);
         SetValidationSetting(layer_setting_set, settings_data->disables, shader_validation_caching, SETTING_CHECK_SHADERS_CACHING);
+        SetValidationSetting(layer_setting_set, settings_data->disables, sync_validation_queue_submit,
```
christophe-lunarg, line 478: This is the part of the new settings and it looks correct.

christophe-lunarg (changes requested):
Sorry, I just realized : You only modified the legacy settings, that could be configured with VK_EXT_validation_features and the flags, but there is a new approach:
```
                            "settings": [
                                {
                                    "key": "sync_queue_submit",
                                    "label": "QueueSubmit Synchronization Validation",
                                    "description": "Enable synchronization validation between submitted command buffers when Synchronization Validation is enabled. This option will increase the synchronization performance cost.",
                                    "type": "BOOL",
                                    "default": false,
                                    "status": "ALPHA",
                                    "dependence": {
                                        "mode": "ALL",
                                        "settings": [
                                            {
                                                "key": "validate_sync",
                                                "value": true
                                            }
                                        ]
                                    }
                                }
                            ]
                            ```
        Replace this by :
                 ```
                            "settings": [
                                {
                                    "key": "sync_queue_submit",
                                    "label": "QueueSubmit Synchronization Validation",
                                    "description": "Enable synchronization validation between submitted command buffers when Synchronization Validation is enabled. This option will increase the synchronization performance cost.",
                                    "type": "BOOL",
                                    "default": true,
                                    "status": "STABLE",
                                    "dependence": {
                                        "mode": "ALL",
                                        "settings": [
                                            {
                                                "key": "validate_sync",
                                                "value": true
                                            }
                                        ]
                                    }
                                }
                            ]
                            ```           
                            
                            Sorry for the late feedback :/

christophe-lunarg (approved):

PR #7131, opened 2023-12-12
test: Cleanup GPU-AV BufferDeviceAddress test

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4983

The old test was incorrect how it mapped the memory (commented below)

Incase the test still hangs on CI, it will be easier to diagnosis now that things are broken up

Also created a dedicated positive file


tests/unit/gpu_av_buffer_device_address.cpp
```diff
@@ -166,102 +113,172 @@ TEST_F(NegativeGpuAVBufferDeviceAddress, DISABLED_Basic) {
...
-        m_commandBuffer->BeginRenderPass(m_renderPassBeginInfo);
-        vk::CmdBindPipeline(m_commandBuffer->handle(), VK_PIPELINE_BIND_POINT_GRAPHICS, pipe.Handle());
-        vk::CmdBindDescriptorSets(m_commandBuffer->handle(), VK_PIPELINE_BIND_POINT_GRAPHICS, pipeline_layout.handle(), 0, 1,
-                                  &descriptor_set.set_, 0, nullptr);
-        vk::CmdDraw(m_commandBuffer->handle(), 3, 1, 0, 0);
-        m_commandBuffer->EndRenderPass();
-        m_commandBuffer->end();
-
-        for (const auto &test : testcases) {
-            auto *data = static_cast<VkDeviceAddress *>(buffer0.memory().map());
```
spencer-lunarg, line 187: the data is 

```
            layout(push_constant) uniform ufoo {
                bufStruct data;
                int nWrites;
            } u_info;
```

but we were pushing 16 bytes here

tests/unit/gpu_av_buffer_device_address.cpp
```diff
@@ -102,58 +49,58 @@ TEST_F(NegativeGpuAVBufferDeviceAddress, DISABLED_Basic) {
...
-    }
-    // descriptor set version
-    {
-        OneOffDescriptorSet descriptor_set(m_device, {{0, VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER, 1, VK_SHADER_STAGE_ALL, nullptr}});
-
-        const vkt::PipelineLayout pipeline_layout(*m_device, {&descriptor_set.layout_});
-        VkDescriptorBufferInfo buffer_test_buffer_info = {};
-        buffer_test_buffer_info.buffer = buffer0.handle();
-        buffer_test_buffer_info.offset = 0;
-        buffer_test_buffer_info.range = sizeof(uint32_t);
```
spencer-lunarg, line 142: we were only updating the first 4 bytes of the buffer, not all 12 bytes, so likely some machines were just luckily pushing the whole cache line, but others were not

arno-lunarg (approved):
LGTM, wondering about the crash ...

spencer-lunarg:
ok, seems the `ReadAfterPointer` test is not be detecting on some machines and then crashing on one... so assume the issue is the shader instrumentation

jeremyg-lunarg (approved):
Thanks for figuring this out

PR #7133, opened 2023-12-12
tests: Use vkt::Queue::submit

spencer-lunarg:
make `m_default_queue` a `vkt::Queue` (instead of a `VkQueue`) 

For many tests that just need to do a "basic submit and wait" they can now just go

```
m_default_queue->submit(*m_commandBuffer);
QueueWaitIdle();
```

(this will basically be every GPU-AV test moving forward)

### DIFF
--- a/tests/framework/binding.cpp
+++ b/tests/framework/binding.cpp
@@ -359,7 +359,7 @@ VkFormatProperties Device::format_properties(VkFormat format) {
     return data;
 }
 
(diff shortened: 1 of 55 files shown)


jeremyg-lunarg:
Why not replace `QueueWaitIdle(); `while you're here?

`m_default_queue->wait_idle(); // or whatever coding standard we're using for these objects.`

PR #7135, opened 2023-12-13
layers: Update push descriptor imageInfo VUID

ziga-lunarg:
@spencer-lunarg This is the change if VU 06493 will be updated


layers/stateless/sl_descriptor.cpp
```diff
@@ -664,17 +664,10 @@ bool StatelessValidation::ValidateWriteDescriptorSet(const Location &loc, const
             (descriptor_type == VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE) || (descriptor_type == VK_DESCRIPTOR_TYPE_STORAGE_IMAGE) ||
```
spencer-lunarg, line 664: while here, lets use the new `IsValueIn` helper

spencer-lunarg (approved):

spencer-lunarg:
we agreed in the WG this is the right change, but it won't be out until a future extension is out, so in the mean time, this is correct

PR #7136, opened 2023-12-13
tests: Add more GPU-AV BDA tests

spencer-lunarg:
Adds all the tests from `inst_buff_addr_check_test.cpp` in `spirv-opt` here to be tested as an end-to-end test for GPU-AV

### DIFF
--- a/tests/unit/gpu_av_buffer_device_address_positive.cpp
+++ b/tests/unit/gpu_av_buffer_device_address_positive.cpp
@@ -19,16 +19,11 @@
 void GpuAVBufferDeviceAddressTest::InitGpuVUBufferDeviceAddress(void *p_next) {
     SetTargetApiVersion(VK_API_VERSION_1_2);
     AddRequiredExtensions(VK_KHR_BUFFER_DEVICE_ADDRESS_EXTENSION_NAME);
+    AddRequiredFeature(vkt::Feature::bufferDeviceAddress);
+    AddRequiredFeature(vkt::Feature::shaderInt64);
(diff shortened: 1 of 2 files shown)


arno-lunarg (approved):
LGTM, but the code for each test is essentially the same, it needs to be refacto

spencer-lunarg:
> LGTM, but the code for each test is essentially the same, it needs to be refacto

yep, the goal was to just get "some" tests in, now that I have a pattern, and need to add more, will DRY it up

tests/unit/gpu_av_buffer_device_address_positive.cpp
```diff
@@ -96,3 +90,277 @@ TEST_F(PositiveGpuAVBufferDeviceAddress, Basic) {
...
+    m_default_queue->wait();
+
+    in_buffer_ptr = (uint8_t *)in_buffer.memory().map();
+    float output = *((float *)(in_buffer_ptr + sizeof(VkDeviceAddress)));
+    in_buffer.memory().unmap();
+    ASSERT_TRUE(output == expected_output);
+}
+
+TEST_F(PositiveGpuAVBufferDeviceAddress, StructLoadPadded) {
+    TEST_DESCRIPTION("Do a OpLoad through a padded struct PhysicalStorageBuffer");
```
filnet, line 230: Three tests have the same description.

spencer-lunarg: good catch, will update

PR #7140, opened 2023-12-14
tests: Use new AddRequiredFeature

spencer-lunarg:
I had some extra time and did an initial pass to update tests to use the new `AddRequiredFeature` ... `-1200 net lines of code` is really  satisfy to see

### DIFF
--- a/tests/framework/render.cpp
+++ b/tests/framework/render.cpp
@@ -596,10 +596,10 @@ void VkRenderFramework::InitState(VkPhysicalDeviceFeatures *features, void *crea
                                   const VkCommandPoolCreateFlags flags) {
     if (!features && !create_device_pnext) {
         if (feature_requirements_.HasFeatures2()) {
-            if (vk::GetPhysicalDeviceFeatures2) {
-                vk::GetPhysicalDeviceFeatures2(gpu(), feature_requirements_.GetFeatures2());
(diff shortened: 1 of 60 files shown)


tests/framework/render.cpp
```diff
@@ -596,10 +596,10 @@ void VkRenderFramework::InitState(VkPhysicalDeviceFeatures *features, void *crea
                                   const VkCommandPoolCreateFlags flags) {
     if (!features && !create_device_pnext) {
         if (feature_requirements_.HasFeatures2()) {
-            if (vk::GetPhysicalDeviceFeatures2) {
-                vk::GetPhysicalDeviceFeatures2(gpu(), feature_requirements_.GetFeatures2());
-            } else {
+            if (vk::GetPhysicalDeviceFeatures2KHR) {
```
arno-lunarg, line 599: did this explode?

arno-lunarg (approved):
LGTM, is CI is happy I am happy

tests/framework/render.cpp
spencer-lunarg: yes, I forgot about this, but this is how we do it else where

basically you can have a case where you expose the core version, but never called GetDeviceProcAddress on it, so it is still null... subtle thing

Issue #7141, opened 2023-12-14
Add remaining VK_ANDROID_external_format_resolve VUs


spencer-lunarg:
remaining coverage for [VK_ANDROID_external_format_resolve](https://github.com/KhronosGroup/Vulkan-Docs/commit/e5dbdd580cf0696db8ed0aeb0736e0f512d9bbe1)

// Sync objects
- [x] VUID-vkCmdPipelineBarrier2-image-04073 (added "color resolve" to VU)
- [x] VUID-vkCmdPipelineBarrier2-image-09373
- [x] VUID-vkCmdPipelineBarrier2-image-09374

// VK_EXT_shader_object
- [x] VUID-vkCmdDraw-None-09366
- [x] VUID-vkCmdDraw-rasterizationSamples-09367
- [x] VUID-vkCmdDraw-pFragmentSize-09370
- [x] VUID-vkCmdDraw-pFragmentSize-09371

// need better way to dispatch... probably need a real device
- [ ] VUID-VkFramebufferCreateInfo-nullColorAttachmentWithExternalFormatResolve-09349
- [ ] VUID-VkRenderPassBeginInfo-framebuffer-09353
- [ ] VUID-VkRenderingAttachmentInfo-resolveMode-09330

// interact with VK_KHR_fragment_shading_rate / VK_EXT_fragment_density_map
- [x] VUID-VkRenderPassCreateInfo2-pResolveAttachments-09331
- [ ] VUID-VkSubpassDescription2-externalFormatResolve-09347

Potentially [new VUs](https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6313) to be added in 1.3.274

PR #7142, opened 2023-12-14
Cleanup and new VUs for VK_ANDROID_external_format_resolve

spencer-lunarg:
for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7141


layers/drawdispatch/drawdispatch_vuids.h
```diff
@@ -185,6 +185,10 @@ struct DrawDispatchVuid {
     const char* external_format_resolve_09368 = kVUIDUndefined;
     const char* external_format_resolve_09369 = kVUIDUndefined;
     const char* external_format_resolve_09372 = kVUIDUndefined;
+    const char* external_format_resolve_09366 = kVUIDUndefined;
```
artem-lunarg, line 188: What is the testing plan for these vuids?

spencer-lunarg: I plan to get tests for all VUs, except these draw-time shader object VUs in this PR... I don't have a way to hit it with MockICD in the current state... I am not too worried as these VUs are simple and just copies of the Pipeline versions above this one

artem-lunarg (approved):

Issue #12, opened 2018-05-14
WSI Swapchain interframe dependency tracking


karl-lunarg:
**Issue by [szdarkhack](https://github.com/szdarkhack) (MIGRATED)**
_Friday Dec 02, 2016 at 13:29 GMT_
_Originally opened as https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/issues/1225_

----

In the current implementation of core validation, when a WSI fence is retired it is not associated with any queue and thus does not retire any work, leading to incorrect errors about resources being in use when in fact they are not. It should be the case that when an acquire fence is waited upon for a particular image index, all previous work for which the respective presentation submission waited should in turn be retired. For instance:

render loop
    acquire image and wait on fence
    submit work that signals a semaphore
    queue present that waits upon that semaphore
end render loop

It should be the case that when the acquire fence is signaled the presentation must have finished, and as such the work submitted prior to the presentation (to the same image index) must also have finished. Therefore, resources used by that submission should be marked as not being in flight (command buffers etc).

karl-lunarg:
<a href="https://github.com/mark-lunarg"><img src="https://avatars2.githubusercontent.com/u/7252958?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [mark-lunarg](https://github.com/mark-lunarg) (MIGRATED)**
_Friday Dec 02, 2016 at 14:41 GMT_

----

Maybe @chrisforbes might have a go at this one?

karl-lunarg:
<a href="https://github.com/chrisforbes"><img src="https://avatars1.githubusercontent.com/u/128877?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [chrisforbes](https://github.com/chrisforbes) (MIGRATED)**
_Wednesday Dec 07, 2016 at 18:41 GMT_

----

@mark-lunarg Yes, I'll take care of this.

mark-lunarg:
Waiting on the acquire fence is not a guarantee that the submitted work has been completed, even on the same image index.  The only way to guarantee this is to wait on the signaled fence from the `QueueSubmit()`.

kvark:
[...]

PR #7145, opened 2023-12-15
layers: Retire submissions using acquire fence

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/12

The test implements and explains a scenario where image-acquire-fence can synchronize frames and guarantees that resources are not in use. Will repeat sync chain structure here too:

1) wait on the image acquire fence -> image was acquired
2) image was acquired -> image was presented in one of the previous frames
3) image was presented ->  corresponding present waited on the submit semaphore
4) submit semaphore was waited on -> corresponding submit finished execution and can be retired/reused

### DIFF
--- a/tests/framework/binding.h
+++ b/tests/framework/binding.h
@@ -384,6 +384,7 @@ class Semaphore : public internal::NonDispHandle<VkSemaphore> {
     Semaphore() = default;
     Semaphore(const Device &dev) { init(dev, vku::InitStruct<VkSemaphoreCreateInfo>()); }
     Semaphore(const Device &dev, const VkSemaphoreCreateInfo &info) { init(dev, info); }
+    Semaphore(Semaphore &&rhs) noexcept : NonDispHandle(std::move(rhs)) {}
     ~Semaphore() noexcept;
     void destroy() noexcept;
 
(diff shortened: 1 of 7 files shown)


charles-lunarg:
Because I'm a degenerate, does this have any special considerations for resizing? AKA recreating the swapchain, passing old swapchain to new one in the `oldSwapchain` member, during any of the above steps? Would be a good thing to have as a positive test regardless.

spencer-lunarg (approved):
LGTM - but not familiar with area, so would like @jeremyg-lunarg review as well

artem-lunarg:
>  does this have any special considerations for resizing? AKA recreating the swapchain, passing old swapchain to new one in the oldSwapchain member, during any of the above steps?

Not much. It could be better name for this PR is "Retire submissions based on acquire fence". All swapchain related manipulation and top level frame synchronization is on the app side. This change is about extending queue forward progress simulation logic.

It's likely there is a use case when this solution needs to be adjusted for the edge cases as you mentioned above, but right now it's hard to see what it can be, assuming the app also does sensible swapchain handling.

tests/framework/binding.h
```diff
@@ -962,6 +963,14 @@ class CommandBuffer : public internal::Handle<VkCommandBuffer> {
                            Queue *queue = nullptr) {
         Init(device, pool, level, queue);
     }
+    CommandBuffer(CommandBuffer &&rhs) noexcept : Handle(std::move(rhs)) {
```
jeremyg-lunarg, line 966: Where does this get used?

jeremyg-lunarg (approved):
LGTM

tests/framework/binding.h
artem-lunarg: Test code:

```
 std::vector<vkt::CommandBuffer> command_buffers;
    std::vector<vkt::Semaphore> submit_semaphores;
    for (size_t i = 0; i < swapchain_images.size(); i++) {
        command_buffers.emplace_back(m_device, m_commandPool);
        submit_semaphores.emplace_back(*m_device);
    }
```

artem-lunarg:
@charles-lunarg @jeremyg-lunarg Added one more tests that implements Charles's suggestion. That helped to find one issue - we could fail detect in-use error the first frame after swapchain is recreated. It is fixed now.

spencer-lunarg:
@karl-lunarg took a tad longer then expected, but we got you covered!

PR #7147, opened 2023-12-15
tests: Fix tests crashing on Intel Mesa

spencer-lunarg:
reasons for each commented on

### DIFF
--- a/layers/core_checks/cc_pipeline_graphics.cpp
+++ b/layers/core_checks/cc_pipeline_graphics.cpp
@@ -987,14 +987,11 @@ bool CoreChecks::ValidateGraphicsPipelineTessellationState(const vvl::Pipeline &
     bool skip = false;
 
     if (pipeline.OwnsSubState(pipeline.pre_raster_state) &&
-        (pipeline.create_info_shaders & VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT) &&
-        (pipeline.create_info_shaders & VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT)) {
+        (pipeline.create_info_shaders & VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT)) {
         if (!pipeline.TessellationState() && (!pipeline.IsDynamic(VK_DYNAMIC_STATE_PATCH_CONTROL_POINTS_EXT) ||
                                               !IsExtEnabled(device_extensions.vk_ext_extended_dynamic_state3))) {
             skip |= LogError("VUID-VkGraphicsPipelineCreateInfo-pStages-09022", device, create_info_loc.dot(Field::pStages),
-                             "includes a tessellation control "
-                             "shader stage and a tessellation evaluation shader stage, "
-                             "but pTessellationState is NULL.");
+                             "includes a tessellation control shader stage, but pTessellationState is NULL.");
         }
     }
     return skip;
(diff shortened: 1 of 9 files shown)


layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -987,8 +987,7 @@ bool CoreChecks::ValidateGraphicsPipelineTessellationState(const vvl::Pipeline &
     bool skip = false;
 
     if (pipeline.OwnsSubState(pipeline.pre_raster_state) &&
-        (pipeline.create_info_shaders & VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT) &&
-        (pipeline.create_info_shaders & VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT)) {
+        (pipeline.create_info_shaders & VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT)) {
```
spencer-lunarg, line 990: VU only says "control" shader, was not catching GPL test bug

tests/unit/best_practices.cpp
```diff
@@ -799,7 +799,7 @@ TEST_F(VkBestPracticesLayerTest, ClearAttachmentsAfterLoadSecondary) {
     auto image_view = image.CreateView();
 
     RenderPassSingleSubpass rp(*this);
-    rp.AddAttachmentDescription(VK_FORMAT_R8G8B8A8_UNORM, VK_IMAGE_LAYOUT_UNDEFINED, VK_IMAGE_LAYOUT_UNDEFINED,
```
spencer-lunarg, line 802: I need to do a follow-up why this wasn't caught, but you can't final layout into UNDEFINED

tests/unit/dynamic_state.cpp
```diff
@@ -5330,6 +5330,7 @@ TEST_F(NegativeDynamicState, DynamicSampleLocationsEnable) {
     CreatePipelineHelper pipe(*this);
     pipe.InitState();
     pipe.AddDynamicState(VK_DYNAMIC_STATE_SAMPLE_LOCATIONS_ENABLE_EXT);
+    pipe.AddDynamicState(VK_DYNAMIC_STATE_SAMPLE_LOCATIONS_EXT);
```
spencer-lunarg, line 5333: https://gitlab.khronos.org/vulkan/vulkan/-/issues/3732

tests/unit/wsi.cpp
```diff
@@ -2733,13 +2733,9 @@ TEST_F(NegativeWsi, CreatingWaylandSurface) {
     GTEST_SKIP() << "test not supported on platform";
 #else
     AddSurfaceExtension();
-
+    AddRequiredExtensions(VK_KHR_WAYLAND_SURFACE_EXTENSION_NAME);
```
spencer-lunarg, line 2736: not a bug, but realize the WSI tests were the old way (when debugging a test)

tests/unit/best_practices.cpp
spencer-lunarg: `VUID-VkAttachmentDescription-finalLayout-00843`

spencer-lunarg: awww, because it was best practice layer and wasn't triggering normal errors

arno-lunarg (approved):

PR #7154, opened 2023-12-18
tests: Cleanup more test init logic

spencer-lunarg:
- Adds `AddRequiredFeature` in more places
- use `Init()` when possible


arno-lunarg (approved):
Nice（￣︶￣）↗

PR #7156, opened 2023-12-18
tests: Enable pipeline rebind test on mock icd

ziga-lunarg:
Set conformance version in max_profile.json to 1.3.6.0

### DIFF
--- a/tests/device_profiles/max_profile.json
+++ b/tests/device_profiles/max_profile.json
@@ -1041,7 +1041,14 @@
                         "VK_SHADER_STAGE_CALLABLE_BIT_KHR"
                     ]
                 },
-                "VkPhysicalDeviceDriverProperties": {},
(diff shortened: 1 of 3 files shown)


tests/device_profiles/max_profile.json
```diff
@@ -1041,7 +1041,14 @@
                         "VK_SHADER_STAGE_CALLABLE_BIT_KHR"
                     ]
                 },
-                "VkPhysicalDeviceDriverProperties": {},
+                "VkPhysicalDeviceDriverProperties": {
```
spencer-lunarg, line 1044: my only concern setting this is if in the future we need to add another "check the CTS version" test, they aren't able to share this

ziga-lunarg: We could choose which of the tests we want to run, based on the coverage by the drivers in CI

PR #7163, opened 2023-12-19
layers: Fix VkDescriptorSetLayoutBinding::stageFlags

spencer-lunarg:
from https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6333

`VkDescriptorSetLayoutBinding::stageFlags` can either be `VK_SHADER_STAGE_ALL` OR another valid item

before something invalid like `0x3BADFFFF` would get covered up by the fact `VK_SHADER_STAGE_ALL` is `0x7FFFFFFF`

for other use cases, it should check `0x7FFFFFFF` as that encompasses all valid stages


scripts/generators/enum_flag_bits_generator.py
```diff
@@ -110,7 +112,13 @@ def generate(self):
...
         out.append('\n')
         out.append('// mask of all the VK_PIPELINE_STAGE_*_SHADER_BIT stages\n')
-        out.append(f'const VkPipelineStageFlagBits2 allVkPipelineShaderStageBits2 = {"|".join([flag.name for flag in self.vk.bitmasks["VkPipelineStageFlagBits2"].flags if "_SHADER_BIT" in flag.name])};\n')
+        out.append(f'const VkPipelineStageFlagBits2 AllVkPipelineShaderStageBits2 = {"|".join([flag.name for flag in self.vk.bitmasks["VkPipelineStageFlagBits2"].flags if "_SHADER_BIT" in flag.name])};\n')
+        out.append('\n')
+
+        # Special edge cases
+        out.append('// VK_SHADER_STAGE_ALL is special bit that is the collection of all bits.\n')
+        out.append('const VkShaderStageFlags AllVkShaderStageFlagBits = VK_SHADER_STAGE_ALL;\n')
+        out.append(f'const VkShaderStageFlags AllVkShaderStageFlagBitsExcludingStageAll = {"|".join([flag.name for flag in self.vk.bitmasks["VkShaderStageFlagBits"].flags if flag.name != "VK_SHADER_STAGE_ALL"])};\n')
```
artem-lunarg, line 121: The usage habit working with similar constants that `AllVk` gives the OR-ed list of separate bit fields. It can be an opportunity for mistake, since you might expect to use `AllVkShaderStageFlagBits` without looking more carefully into code. What do you think about having `AllVkShaderStageFlagBits` as before but without `VK_SHADER_STAGE_ALL` (no no other constant)? With a comment that this does not include STAGE_ALL.

artem-lunarg (approved):
LGTM. Added comment about flags constant semantics.

scripts/generators/enum_flag_bits_generator.py
spencer-lunarg: so the issue is the `AllVkShaderStageFlagBits` is the auto-gen name we used in the implicit validation logic (which needs `VK_SHADER_STAGE_ALL` as it is correct there)

the issue is in the few places we use the `AllVk` values for explicit checks (which is where `AllVkShaderStageFlagBitsExcludingStageAll` is used)

I guess no one should really ever have to use `AllVkShaderStageFlagBits` manually so while I understand the concern, its a very low risk to become an issue

PR #7165, opened 2023-12-19
tests: Revert sync_val dynamicRendering feature check

spencer-lunarg:
Galaxy s23 is failing to create a `VkDevice` (but only sometimes :cry: ) for the sync val tests

this reverts the 3 cases from https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7140/ that would likely be  a problem


arno-lunarg (approved):

artem-lunarg (approved):
This looks weird but lets check. Just not to forget about this because can hide bugs.

Issue #7169, opened 2023-12-19
Add safe struct for VkPushDescriptorSetWithTemplateInfoKHR


mikes-lunarg:
`VkPushDescriptorSetWithTemplateInfoKHR` is part of VK_KHR_maintenance6 which appeared in the 1.1.274 header.

Currently stubbed out so we can build, but it needs a custom constructor which can copy `pData` correctly.

PR #7170, opened 2023-12-19
build: Update to header 1.3.274

mikes-lunarg:
- Update known-good
- Fix crash in extension_helper_generator when an extension has an explicit dependency on VK_VERSION_1_0
- Stub out safe struct constructor for VkPushDescriptorSetWithTemplateInfoKHR until it can be properly implemented #7169 
- VUID renames:
  - VUID-VkDescriptorSetLayoutBinding-descriptorCount-00283 -> 09465
  - UNASSIGNED-vkCmdDrawIndirectByteCountEXT-counterOffset -> 09474
  - UNASSIGNED-vkCmdDrawIndirectByteCountEXT-vertexStride -> 09475
- Fix video enums that got renamed EXT -> KHR
- Generate source


scripts/generators/extension_helper_generator.py
```diff
@@ -104,6 +104,9 @@ def generate(self):
                 # This is a work around for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5372
                 temp = re.sub(r',VK_VERSION_1_\d+', '', extension.depends)
                 for reqs in exprValues(parseExpr(temp)):
+                    if reqs == 'VK_VERSION_1_0':
+                        # An explicit dependency on Vulkan 1.0 is meaningless
+                        continue
```
mikes-lunarg, line 109: @spencer-lunarg Is this comment correct? Is it worth pushing back to the WG? As far as I can tell VK_KHR_maintenance is the first extension with `depends="VK_VERSION_1_0"` in the XML

spencer-lunarg: ya... feel that was an oversite, happy to take action with the WG to fix that

spencer-lunarg (approved):

PR #7188, opened 2023-12-23
layers: Fixed bounds error in Module::StaticData constructor

syukosevn:


layers/state_tracker/shader_module.cpp
```diff
@@ -791,7 +791,7 @@ Module::StaticData::StaticData(const Module& module_state) {
     {
         std::vector<uint32_t>::const_iterator it = module_state.words_.cbegin();
         it += 5;  // skip first 5 word of header
-        while (it != module_state.words_.cend()) {
+        while (it < module_state.words_.cend()) {
```
spencer-lunarg, line 794: sorry, but I am confused what this is fixing, this has never been a problem checking for `!= list.cend()`

do you have an example to showcase where the previous code was not working?

syukosevn:
When I pass 0 code size and nullptr as code ptr, validation layer causes crash
The problem is inside these two lines:
```
it += 5;  // skip first 5 word of header
while (it != module_state.words_.cend())
```
`it` will never become equal `words_.cend()` after `+= 5`

This change probably won't save from another crash with same input, but at least not because of incorrect iterators usage.

Here are callstack and minimal example.
```
[Inlined] small_vector::emplace_back(const unsigned int &) 0x00007fff8be631a9
Instruction::Instruction(_Vector_const_iterator<…>) 0x00007fff8be63191
SPIRV_MODULE_STATE::StaticData::StaticData(const SPIRV_MODULE_STATE &) 0x00007fff8be6585f
[Inlined] SPIRV_MODULE_STATE::{ctor}(unsigned long long,const unsigned int *) 0x00007fff8be58942
[Inlined] std::_Construct_in_place(SPIRV_MODULE_STATE &, const unsigned long long &, const unsigned int *const &) 0x00007fff8be58936
[Inlined] std::_Ref_count_obj2<SPIRV_MODULE_STATE>::{ctor}(const unsigned long long &,const unsigned int *const &) 0x00007fff8be58936
std::make_shared<SPIRV_MODULE_STATE,unsigned __int64 const &,unsigned int const * const &>(const unsigned long long &,const unsigned int *const &) 0x00007fff8be58936
ValidationStateTracker::PreCallRecordCreateShaderModule(VkDevice_T *, const VkShaderModuleCreateInfo *, const VkAllocationCallbacks *, VkShaderModule_T **, void *) 0x00007fff8be943d6
CoreChecks::PreCallRecordCreateShaderModule(VkDevice_T *, const VkShaderModuleCreateInfo *, const VkAllocationCallbacks *, VkShaderModule_T **, void *) 0x00007fff8bb7f74f
vulkan_layer_chassis::CreateShaderModule(VkDevice_T *, const VkShaderModuleCreateInfo *, const VkAllocationCallbacks *, VkShaderModule_T **) 0x00007fff8bc13065
vk::raii::ShaderModule::ShaderModule(const vk::raii::Device &, const vk::ShaderModuleCreateInfo &, Optional<…>) vulkan_raii.hpp:10690
vk::raii::Device::createShaderModule(const vk::ShaderModuleCreateInfo &, Optional<…>) vulkan_raii.hpp:12408
main(int, char **) mwe.cpp:41
invoke_main() 0x00007ff675ab0ed9
__scrt_common_main_seh() 0x00007ff675ab0dbe
__scrt_common_main() 0x00007ff675ab0c7e
mainCRTStartup(void *) 0x00007ff675ab0f6e
<unknown> 0x00007ff808517344
<unknown> 0x00007ff8086626b1
```

```
#include <vulkan/vulkan_raii.hpp>
#include <array>

int main(int argc, char* argv[])
{
    vk::raii::Context context;
    std::array layers{
        "VK_LAYER_KHRONOS_validation"
    };
    vk::InstanceCreateInfo instance_create_info{
        {}, {}, layers
    };
    auto instance{context.createInstance(instance_create_info)};
    auto physical_devices{instance.enumeratePhysicalDevices()};
    if (physical_devices.empty())
        throw;

    auto queue_family_properties{
        physical_devices[0].getQueueFamilyProperties()
    };
    if (queue_family_properties.empty())
        throw;

    std::array device_queue_priorities{1.0f};
    std::array device_queue_create_infos{
        vk::DeviceQueueCreateInfo{{}, 0, device_queue_priorities}
    };
    const vk::DeviceCreateInfo device_create_info{
        {},
        device_queue_create_infos,
        layers
    };

    auto device{physical_devices[0].createDevice(device_create_info)};

    vk::ShaderModuleCreateInfo shader_module_create_info{
        {},
        0, //< code size,
        nullptr //< code
    };
    auto shader_module{device.createShaderModule(shader_module_create_info)};  //< crashes
    return 0;
}
```

syukosevn:
sdk version 1.3.268.0

spencer-lunarg:
so it is not valid to have a `codeSize` of `0` 

image

So I agree this shouldn't crash, but I feel it shouldn't have got to the `StaticData` logic, clearly there is no reason it should have started down this code path.

The reason I say this, is while we could just patch this one line, there is a very non-zero chance this code will be refactored and it will not be obvious to anyone that they are expected to handle an invalid `codeSize` value here

spencer-lunarg:
ok, I just realized for some reason ` VUID-VkShaderModuleCreateInfo-codeSize-01085` is not being validated, let me add a check for that (and a test like the one you gave) to help ensure this won't crash like this for you in the future

PR #7197, opened 2023-12-27
dynamic rendering: fix incorrect error message

evopen:


layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -972,7 +972,7 @@ bool CoreChecks::ValidateDrawDynamicStateShaderObject(const LastBound& last_boun
         !cb_state.dynamic_state_status.cb[CB_DYNAMIC_STATE_SCISSOR_WITH_COUNT]) {
         skip |= LogError(vuid.viewport_and_scissor_with_count_08635, cb_state.commandBuffer(), loc,
                          "Graphics shader objects are bound, but vkCmdSetViewportWithCount() and "
-                         "vkCmdSetViewportWithCount() were not both called.");
+                         "vkCmdSetScissorWithCount() were not both called.");
     } else if (cb_state.dynamic_state_value.viewport_count != cb_state.dynamic_state_value.scissor_count) {
         skip |= LogError(vuid.viewport_and_scissor_with_count_08635, cb_state.commandBuffer(), loc,
                          "Graphics shader objects are bound, but viewportCount set with vkCmdSetViewportWithCount() was %" PRIu32
```
spencer-lunarg, line 978: thanks for the fix!

1. you should rename the commit to `layers: Fix incorrect error message` (CI will yell at you otherwise)
2. while here, can you fix the `vkCmdSetViewportWithCount` in the VU below as well

spencer-lunarg (approved):
sorry, the CI is case sensative 

```patch
- layers: fix error message in dynamic state
+ layers: Fix error message in dynamic state
```

but once CI is happy, this is good to go!

layers/core_checks/cc_cmd_buffer_dynamic.cpp
evopen: sure. good catch!

spencer-lunarg:
Thanks again for the patch!

PR #7201, opened 2023-12-27
cmake: Use SPIRV-Tools-static for testing

juan-lunarg:
Because testing is OFF by default. We don't need to handle all the different edge cases for package managers.

Testing should just use the static binaries by default. This avoids complications for Windows / Android APK testing.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -243,11 +243,13 @@ endif()
 
 find_package(GTest CONFIG)
 find_package(glslang CONFIG)
+find_package(SPIRV-Tools CONFIG)
 
 target_link_libraries(vk_layer_validation_tests PRIVATE
     VkLayer_utils
     glslang::SPIRV
     glslang::SPVRemapper
+    SPIRV-Tools-static
     SPIRV-Headers::SPIRV-Headers
     GTest::gtest
     GTest::gtest_main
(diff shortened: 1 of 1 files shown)


tests/CMakeLists.txt
```diff
@@ -256,21 +258,6 @@ target_link_libraries(vk_layer_validation_tests PRIVATE
...
-# Try to handle all possible combinations so that we work with externally built packages.
-if (TARGET SPIRV-Tools)
-    target_link_libraries(vk_layer_validation_tests PRIVATE SPIRV-Tools)
-elseif(TARGET SPIRV-Tools-static)
-    target_link_libraries(vk_layer_validation_tests PRIVATE SPIRV-Tools-static)
-elseif(TARGET SPIRV-Tools-shared)
-    target_link_libraries(vk_layer_validation_tests PRIVATE SPIRV-Tools-shared)
-else()
-    message(FATAL_ERROR "Cannot determine SPIRV-Tools target name")
-endif()
```
juan-lunarg, line 272: This code was added to make system package managers happy. Since package managers tend to want dynamic and NOT static libraries. However, now that this CMake code is ONLY in the tests directory it doesn't matter. (Originally the code wasn't only in the test directory)

Ideally SPIRV-Tools does fix their CMake but for now just simplify the CMake here.

spencer-lunarg (approved):

PR #7203, opened 2023-12-28
layers: Add queryCount and stride check

spencer-lunarg:
adds `VUID-vkCmdCopyQueryPoolResults-queryCount-09438` and `VUID-vkGetQueryPoolResults-queryCount-09438`


tests/unit/tooling_positive.cpp
```diff
@@ -23,11 +23,11 @@ TEST_F(PositiveTooling, BasicUsage) {
         GTEST_SKIP() << "Test not supported by MockICD";
     }
 
-    const auto fpGetPhysicalDeviceToolPropertiesEXT =
```
spencer-lunarg, line 26: um.. not sure how this got in, but it was wrong, so rather just sneak in this nit fix

arno-lunarg (approved):

PR #7208, opened 2023-12-29
Cleanup shader testing helper functions

spencer-lunarg:
The `debug` was added to help debug the shaders writing tests for, but something like that is better served as a runtime setting, not a baked in compile value

### DIFF
--- a/tests/framework/render.h
+++ b/tests/framework/render.h
@@ -152,10 +152,10 @@ class VkRenderFramework : public VkTestFramework {
 
     template <typename GLSLContainer>
     std::vector<uint32_t> GLSLToSPV(VkShaderStageFlagBits stage, const GLSLContainer &code, const char *entry_point = "main",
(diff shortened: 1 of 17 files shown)


artem-lunarg:
What's the recommendation if someone needs shader with debug info? Is it something that will be implemented later?

artem-lunarg (approved):

spencer-lunarg:
> What's the recommendation if someone needs shader with debug info?

so for the glsl shaders we write, this is honestly never needed. Where this does come in handy is when dealing with the GPU-AV shader instrumentation.

The upcoming refactor will have a way to properly do this so you will be able to take your GLSL, run with 

```
  -g          generate debug information
  -gV         generate nonsemantic shader debug information
  -gVS        generate nonsemantic shader debug information with source
  ```
  
  and pipe it through if you need to understand where the input is coming from... basically if someone is doing this type of debugging, I would be suprised if they aren't comfortable using the various glsl/spirv tools

PR #7220, opened 2023-12-30
build: Remove "lib" prefix in output binary

### LINKED ISSUE #7213: `JSON_LIBRARY_PATH` doesn't match the path of the DLL when building on MSYS2
## Environment
 - System: Windows, MSYS2
 - Toolchain: `mingw-w64-x86_64`

## Description
`library_path`'s value in `VkLayer_khronos_validation.json` does not match the actual path of the library:
 - `"library_path": ".\\VkLayer_khronos_validation.dll"`
 - The actual filename of the library: `libVkLayer_khronos_validation.dll`

This causes Vulkan to complain about `VK_ERROR_LAYER_NOT_PRESENT` when the layer is clearly there, its JSON just has the wrong library path.

## Possible causes
1. These lines tell CMake what to put in `library_path`. Perhaps it is possible to add a condition that tests for MSYS2?
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/98dea769721b40ed5402b35a84867f180d8955ef/layers/CMakeLists.txt#L390-L396

2. These lines change target properties based on the system. Perhaps it is possible to add a condition that removes the `lib` prefix on MSYS2?
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/98dea769721b40ed5402b35a84867f180d8955ef/layers/CMakeLists.txt#L313-L328

charles-lunarg:
Fixes #7213

### DIFF
--- a/layers/CMakeLists.txt
+++ b/layers/CMakeLists.txt
@@ -318,6 +318,7 @@ if(MSVC)
 elseif(MINGW)
     target_sources(vvl PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/${LAYER_NAME}.def)
     target_compile_options(vvl PRIVATE -Wa,-mbig-obj)
+    set_target_properties(vvl PROPERTIES PREFIX "") # remove the prefix "lib" so the manifest json "library_path" matches
 elseif(APPLE)
     set_target_properties(vvl PROPERTIES SUFFIX ".dylib")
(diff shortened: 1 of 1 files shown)


Martysh12:
@spencer-lunarg @charles-lunarg It works!

spencer-lunarg (approved):

PR #7221, opened 2023-12-30
tests: Confirm we test for valid VkDeferredOperationKHR object

spencer-lunarg:
there is a missing `VUID-vkBuildAccelerationStructuresKHR-deferredOperation-03677` VUID but these tests prove that it is really just `VUID-vkBuildAccelerationStructuresKHR-deferredOperation-parameter` (plan to remove the `03677` from the spec)

### DIFF
--- a/tests/unit/ray_tracing.cpp
+++ b/tests/unit/ray_tracing.cpp
@@ -2340,6 +2340,22 @@ TEST_F(NegativeRayTracing, WriteAccelerationStructuresPropertiesMaintenance1) {
     }
 }
 
+TEST_F(NegativeRayTracing, BuildAccelerationStructuresDeferredOperation) {
+    TEST_DESCRIPTION("Call vkBuildAccelerationStructuresKHR with a valid VkDeferredOperationKHR object");
+    SetTargetApiVersion(VK_API_VERSION_1_1);
+    AddRequiredFeature(vkt::Feature::accelerationStructure);
+    AddRequiredFeature(vkt::Feature::accelerationStructureHostCommands);
+    AddRequiredFeature(vkt::Feature::bufferDeviceAddress);
+    RETURN_IF_SKIP(InitFrameworkForRayTracingTest());
+    RETURN_IF_SKIP(InitState());
+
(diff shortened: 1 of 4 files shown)


tests/unit/ray_tracing.cpp
```diff
@@ -3292,6 +3292,27 @@ TEST_F(NegativeRayTracing, WriteAccelerationStructuresPropertiesMaintenance1) {
...
+
+    VkDeferredOperationKHR deferred_op = CastFromUint64<VkDeferredOperationKHR>(0xdeadbeef);
+
+    vkt::as::BuildGeometryInfoKHR as_build_info = vkt::as::blueprint::BuildGeometryInfoSimpleOnDeviceBottomLevel(*m_device);
+    as_build_info.BuildHost(instance(), *m_device);
+
+    VkAccelerationStructureBuildRangeInfoKHR build_range_info;
+    const VkAccelerationStructureBuildRangeInfoKHR *p_build_range_info[1]{&build_range_info};
+    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-vkBuildAccelerationStructuresKHR-deferredOperation-parameter");
+    vk::BuildAccelerationStructuresKHR(m_device->handle(), deferred_op, 1, &as_build_info.GetInfo(), p_build_range_info);
```
arno-lunarg, line 3312: you should add `BuildGeometryInfoKHR::SetDeferredOp`, have the underlying attribute be used in `BuildGeometryInfoKHR::VkBuildAccelerationStructuresKHR`, and then use `as_build_info.BuildHost` here,

arno-lunarg (changes requested):

tests/unit/ray_tracing.cpp
spencer-lunarg: I honestly didn't think it was worth adding complexity to the framework for probably the only tests that will use `VkDeferredOperationKHR`, but will change

arno-lunarg (approved):

arno-lunarg (approved):

PR #7226, opened 2023-12-30
layers: Add misc AccelerationStructure VUs

spencer-lunarg:
adds

- `VUID-vkCopyMemoryToAccelerationStructureKHR-pInfo-03750`
- `VUID-vkCmdBuildAccelerationStructuresIndirectKHR-pIndirectStrides-03787`

### DIFF
--- a/tests/unit/ray_tracing.cpp
+++ b/tests/unit/ray_tracing.cpp
@@ -17,6 +17,7 @@
 #include "../framework/descriptor_helper.h"
 #include "../framework/shader_helper.h"
 #include "../layers/utils/vk_layer_utils.h"
+#include <cstdlib>
 
 void RayTracingTest::OOBRayTracingShadersTestBody(bool gpu_assisted) {
     SetTargetApiVersion(VK_API_VERSION_1_1);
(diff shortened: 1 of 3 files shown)


tests/unit/ray_tracing.cpp
```diff
@@ -1768,6 +1769,32 @@ TEST_F(NegativeRayTracing, WriteAccelerationStructureMemory) {
...
+    info.src = output_data;
+    info.mode = VK_COPY_ACCELERATION_STRUCTURE_MODE_DESERIALIZE_KHR;
+
+    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-vkCopyMemoryToAccelerationStructureKHR-pInfo-03750");
+    vk::CopyMemoryToAccelerationStructureKHR(device(), VK_NULL_HANDLE, &info);
+    m_errorMonitor->VerifyFound();
+}
+
+// Because of aligned_alloc
+#if defined(__linux__) && !defined(__ANDROID__)
```
spencer-lunarg, line 1797: we have this issue in the `ExternalMemorySync` tests too... we need some simple/cross-platform way to align these allocations, for now testing on Linux is sufficient, someone on Windows will need to test removing this

arno-lunarg: Can't you simply align the address returned by malloc in this case?

tests/unit/ray_tracing.cpp
```diff
@@ -2340,10 +2369,42 @@ TEST_F(NegativeRayTracing, IndirectCmdBuildAccelerationStructuresKHR) {
...
+
+    auto build_geometry_info = vkt::as::blueprint::BuildGeometryInfoSimpleOnDeviceBottomLevel(*m_device);
+    const auto &geometries = build_geometry_info.GetGeometries()[0].GetVkObj();
+    std::vector<const VkAccelerationStructureGeometryKHR *> pGeometries = {&geometries};
+
+    auto build_info = build_geometry_info.GetInfo();
+    build_info.geometryCount = 1;
+    build_info.ppGeometries = pGeometries.data();
+
+    VkDeviceAddress indirect_device_addresses{};
```
arno-lunarg, line 2399: I was expecting
> VUID-vkCmdBuildAccelerationStructuresIndirectKHR-pIndirectDeviceAddresses-parameter
pIndirectDeviceAddresses must be a valid pointer to an array of infoCount [VkDeviceAddress](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkDeviceAddress.html) values

To fire because of the null address?

Looking at what I did, actually this is also probably why the positive indirect test is crashing

tests/unit/ray_tracing.cpp
```diff
@@ -2340,10 +2369,42 @@ TEST_F(NegativeRayTracing, IndirectCmdBuildAccelerationStructuresKHR) {
...
+
+    m_commandBuffer->begin();
     m_errorMonitor->SetDesiredFailureMsg(kErrorBit,
                                          "VUID-vkCmdBuildAccelerationStructuresIndirectKHR-dstAccelerationStructure-03800");
     build_info_null_dst.BuildCmdBufferIndirect(*m_device, m_commandBuffer->handle());
     m_errorMonitor->VerifyFound();
+    m_commandBuffer->end();
+}
+
+TEST_F(NegativeRayTracing, IndirectStridesMultiple) {
```
arno-lunarg, line 2381: Not super happy about you copy pasting the internals of ray_tracing_objects here... I'd rather just add a method to set the indirect stride.
One more reason for doing this is that now, when I fix the positive tests by modifying those internals, I have to go here again to also fix it

spencer-lunarg:
closing for https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7324

PR #7232, opened 2024-01-02
layers: Add 03770

arno-lunarg:


layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -530,25 +531,44 @@ bool CoreChecks::ValidateAccelerationBuffers(uint32_t info_index, const VkAccele
...
             switch (geom_data.geometryType) {
                 case VK_GEOMETRY_TYPE_TRIANGLES_KHR:  // == VK_GEOMETRY_TYPE_TRIANGLES_NV
-                    skip |= buffer_check(geom_index, geom_data.geometry.triangles.vertexData,
-                                         geom_loc.dot(Field::geometry).dot(Field::triangles).dot(Field::vertexData));
-                    skip |= buffer_check(geom_index, geom_data.geometry.triangles.indexData,
-                                         geom_loc.dot(Field::geometry).dot(Field::triangles).dot(Field::indexData));
-                    skip |= buffer_check(geom_index, geom_data.geometry.triangles.transformData,
-                                         geom_loc.dot(Field::geometry).dot(Field::triangles).dot(Field::transformData));
+                {
+                    skip |= buffer_check(geom_i, geom_data.geometry.triangles.vertexData,
```
spencer-lunarg, line 541: since we already have `geom_loc` and this currently has to re-build the Location object 3 times to call down

```suggestion
                    const Location triangles_loc = geom_loc.dot(Field::triangles);
                    skip |= buffer_check(geom_i, geom_data.geometry.triangles.vertexData, triangles_loc.dot(Field::vertexData)
```

layers/state_tracker/state_tracker.cpp
```diff
@@ -2417,8 +2421,8 @@ void ValidationStateTracker::PostCallRecordCmdBuildAccelerationStructuresKHR(
         return;
     }
     cb_state->RecordCmd(record_obj.location.function);
-    for (uint32_t i = 0; i < infoCount; i++) {
-        RecordDeviceAccelerationStructureBuildInfo(*cb_state, pInfos[i]);
+    for (const auto [i, info] : vvl::enumerate(pInfos, infoCount)) {
```
spencer-lunarg, line 2424: this case of `vvl::enumerate` feel strange because `infoCount` controls both `pInfos` and `ppBuildRangeInfos` and to only have one object enumerated seems off (because how do you decide which

for here it is simple enough of a loop not worried, just pointing it out

layers/state_tracker/state_tracker.cpp
```diff
@@ -2433,7 +2437,7 @@ void ValidationStateTracker::PostCallRecordCmdBuildAccelerationStructuresIndirec
     }
     cb_state->RecordCmd(record_obj.location.function);
     for (uint32_t i = 0; i < infoCount; i++) {
-        RecordDeviceAccelerationStructureBuildInfo(*cb_state, pInfos[i]);
+        RecordDeviceAccelerationStructureBuildInfo(*cb_state, pInfos[i], nullptr);
```
spencer-lunarg, line 2440: instead of passing nullptr here, why not just do the `UpdateBuildRangeInfos` only in the `PostCallRecordCmdBuildAccelerationStructuresKHR` call if that is the only place that needs it

layers/core_checks/cc_ray_tracing.cpp
arno-lunarg: yeah better

layers/state_tracker/state_tracker.cpp
arno-lunarg: I agree it is a bit unusual, I chose to iterate over `pInfos` as the `ppBuildRangeInfos` are essentially an attribute of the geometries specified in `pInfos`

layers/state_tracker/state_tracker.cpp
arno-lunarg: yes better too

spencer-lunarg (approved):

Issue #7233, opened 2024-01-03
VK_EXT_shader_object and VK_EXT_descriptor_buffer are incompatible


evopen:
**Environment:**
 - OS: Linux
 - GPU and driver version: NVIDIA RTX 4080 545.29.06
 - SDK or header version if building from repo: validation layer 0e381678018f858395770d8ddd320af499006628, header 1.3.270
 - Options enabled (synchronization, best practices, etc.):

**Describe the Issue**

When using shader_object with descriptor_buffer extensions, descriptor is bounded by `vkCmdSetDescriptorBufferOffsetsEXT`. Validation reports `VkShaderEXT 0xd10d270000000018[fragment] uses set #0 but that set is not bound`.

**Expected behavior**

Validation layer will not report error.

**Valid Usage ID**
```
VUID-vkCmdDraw-None-08600(ERROR / SPEC): msgNum: 1198051129 - Validation Error: [ VUID-vkCmdDraw-None-08600 ] Object 0: handle = 0x8ec9ca0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xd10d270000000018, name = fragment, type = VK_OBJECT_TYPE_SHADER_EXT; | MessageID = 0x4768cf39 | vkCmdDraw():  VkShaderEXT 0xd10d270000000018[fragment] uses set #0 but that set is not bound. The Vulkan spec states: For each set n that is statically used by a bound shader, a descriptor set must have been bound to n at the same pipeline bind point, with a VkPipelineLayout that is compatible for set n, with the VkPipelineLayout used to create the current VkPipeline or the VkDescriptorSetLayout array used to create the current VkShaderEXT , as described in Pipeline Layout Compatibility (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdDraw-None-08600)
    Objects: 2
        [0] 0x8ec9ca0, type: 6, name: NULL
        [1] 0xd10d270000000018, type: 1000482000, name: fragment
```

**Additional context**

It seems that validation code path for shader object did not take descriptor buffer into account, unlike the pipeline code path.

spencer-lunarg:
was able to reproduce, the issue is `VkPipeline` objects are created with `VK_PIPELINE_CREATE_DESCRIPTOR_BUFFER_BIT_EXT` so we know not to check against what was set in `vkCmdBindDescriptorSets`... will have a temp fix up soon

[...]

PR #7235, opened 2024-01-03
layers: Fix ShaderObject with DescriptorBuffers

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7233

If anyone is using `vkCmdBindDescriptorBuffersEXT` with `vkCmdBindShadersEXT` we need to not try and look if the descriptor is used

Just like the Pipeline path, we have no system in play to track the descriptor buffers bound yet

### DIFF
--- a/layers/core_checks/cc_drawdispatch.cpp
+++ b/layers/core_checks/cc_drawdispatch.cpp
@@ -1,8 +1,8 @@
-/* Copyright (c) 2015-2023 The Khronos Group Inc.
- * Copyright (c) 2015-2023 Valve Corporation
- * Copyright (c) 2015-2023 LunarG, Inc.
- * Copyright (C) 2015-2023 Google Inc.
- * Modifications Copyright (C) 2020-2023 Advanced Micro Devices, Inc. All rights reserved.
+/* Copyright (c) 2015-2024 The Khronos Group Inc.
+ * Copyright (c) 2015-2024 Valve Corporation
+ * Copyright (c) 2015-2024 LunarG, Inc.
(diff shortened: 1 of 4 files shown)


layers/core_checks/cc_drawdispatch.cpp
```diff
@@ -1538,7 +1538,7 @@ bool CoreChecks::ValidateActionState(const vvl::CommandBuffer &cb_state, const V
                 if (ds.bound_descriptor_set && !ds.bound_descriptor_set->IsPushDescriptor()) {
                     const LogObjectList objlist(cb_state.Handle(), pipeline->Handle(), ds.bound_descriptor_set->Handle());
                     skip |=
-                        LogError(vuid.descriptor_buffer_set_offset_missing_08117, objlist, loc,
```
spencer-lunarg, line 1541: these were mixed and made the error message REALLY confusing

layers/core_checks/cc_drawdispatch.cpp
```diff
@@ -1629,7 +1629,11 @@ bool CoreChecks::ValidateActionState(const vvl::CommandBuffer &cb_state, const V
                 }
             }
         }
-    } else {
+    } else if (last_bound_state.cb_state.descriptor_buffer_binding_info.empty()) {
+        // TODO - VkPipeline have VK_PIPELINE_CREATE_DESCRIPTOR_BUFFER_BIT_EXT (descriptor_buffer_mode) to know if using descriptor
```
arno-lunarg, line 1633: do we need an issue for that todo?

arno-lunarg (approved):

Issue #6027, opened 2023-06-19
Add new VK_EXT_descriptor_buffer VUs


spencer-lunarg:
The following were added in 1.3.254 (https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5929)

- [x] VUID-VkDescriptorAddressInfoEXT-nullDescriptor-08938
- [x] VUID-VkDescriptorAddressInfoEXT-nullDescriptor-08939
- [x] VUID-VkDescriptorAddressInfoEXT-range-08940

> range must: not be VK_WHOLE_SIZE

is only when `VK_EXT_robustness2` is not enabled

---

The following were added in 1.3.270 (https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6255)

- [ ] VUID-VkDescriptorGetInfoEXT-type-09427
- [ ] VUID-VkDescriptorGetInfoEXT-type-09428

PR #7242, opened 2024-01-04
layers: Add maxTexelBufferElements for decriptor buffers

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6027

adds

- `VUID-VkDescriptorGetInfoEXT-type-09427`
- `VUID-VkDescriptorGetInfoEXT-type-09428`

these are similar to VUs like `VUID-VkBufferViewCreateInfo-range-00930`

while here cleaned up the function as there were a few bugs

### DIFF
--- a/layers/core_checks/cc_buffer.cpp
+++ b/layers/core_checks/cc_buffer.cpp
@@ -1,7 +1,7 @@
-/* Copyright (c) 2015-2023 The Khronos Group Inc.
- * Copyright (c) 2015-2023 Valve Corporation
- * Copyright (c) 2015-2023 LunarG, Inc.
- * Copyright (C) 2015-2023 Google Inc.
+/* Copyright (c) 2015-2024 The Khronos Group Inc.
+ * Copyright (c) 2015-2024 Valve Corporation
+ * Copyright (c) 2015-2024 LunarG, Inc.
+ * Copyright (C) 2015-2024 Google Inc.
  * Modifications Copyright (C) 2020-2022 Advanced Micro Devices, Inc. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -45,12 +45,14 @@ bool CoreChecks::ValidateBufferUsageFlags(const LogObjectList &objlist, vvl::Buf
 }
 
 bool CoreChecks::ValidateBufferViewRange(const vvl::Buffer &buffer_state, const VkBufferViewCreateInfo *pCreateInfo,
-                                         const VkPhysicalDeviceLimits *device_limits, const Location &loc) const {
+                                         const Location &loc) const {
     bool skip = false;
 
     const VkDeviceSize &range = pCreateInfo->range;
     const VkFormat format = pCreateInfo->format;
-    const uint32_t format_size = vkuFormatElementSize(format);
+    const VKU_FORMAT_INFO format_info = vkuGetFormatInfo(format);
+    // will be 1 most likely because HW don't tend to support texel buffer for compressed formats
+    VkDeviceSize texel_per_block = static_cast<VkDeviceSize>(format_info.texel_per_block);
     if (range != VK_WHOLE_SIZE) {
         // Range must be greater than 0
         if (range <= 0) {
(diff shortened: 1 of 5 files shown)


layers/core_checks/cc_descriptor.cpp
```diff
@@ -2833,51 +2841,59 @@ bool CoreChecks::PreCallValidateGetDescriptorEXT(VkDevice device, const VkDescri
...
                                  "pStorageTexelBuffer->address (%" PRIu64
                                  ") is not zero or "
                                  "an address within a buffer.",
-                                 pDescriptorInfo->data.pStorageTexelBuffer->address);
+                                 address_info->address);
             }
             break;
         case VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER:
-            if (pDescriptorInfo->data.pUniformBuffer && (pDescriptorInfo->data.pUniformBuffer->address != 0) &&
-                (GetBuffersByAddress(pDescriptorInfo->data.pStorageTexelBuffer->address).empty())) {
```
spencer-lunarg, line 2859: there were various copy-and-paste errors in this functions sadly, by using the `data_field` and `address_info` it helps remove the chance for most of these to happen again

layers/core_checks/cc_descriptor.cpp
```diff
@@ -3018,6 +3034,31 @@ bool CoreChecks::PreCallValidateGetDescriptorEXT(VkDevice device, const VkDescri
             break;
     }
 
+    if (address_info && (pDescriptorInfo->type == VK_DESCRIPTOR_TYPE_UNIFORM_TEXEL_BUFFER ||
+                         pDescriptorInfo->type == VK_DESCRIPTOR_TYPE_STORAGE_TEXEL_BUFFER)) {
+        const uint32_t format_size = vkuFormatElementSize(address_info->format);
+        if (address_info->range != VK_WHOLE_SIZE &&
+            (SafeDivision(address_info->range, format_size) > phys_dev_props.limits.maxTexelBufferElements)) {
```
arno-lunarg, line 3041: `SafeDivision(address_info->range, format_size)` does not look like the computation the VUID mentions, `⌊pUniformBuffer->range / (texel block size)⌋ × (texels per block)`?

arno-lunarg: If I am not mistaken, I think you to do need something like:
```cpp
uint32_t block_count = (uint32_t)((float)address_info->range / (float)format_size);
uint32_t texels_per_block = SomeHowGetIt;
uint32_t texels_count = block_count * texels_per_block;
if (texels_count > phys_dev_props.limits.maxTexelBufferElements) { ... }
```

spencer-lunarg: I took this logic from `VUID-VkBufferViewCreateInfo-range-00930` which is doing the same check ... let me write a test to confirm the math here

spencer-lunarg: logic updated, the reason this was "just working" is because `texels_per_block` is always `1` for non-compressed formats and no hardware I found supports compressed formats for Texel Buffers

layers/core_checks/cc_buffer.cpp
```diff
@@ -45,33 +45,34 @@ bool CoreChecks::ValidateBufferUsageFlags(const LogObjectList &objlist, vvl::Buf
...
 
 bool CoreChecks::ValidateBufferViewRange(const vvl::Buffer &buffer_state, const VkBufferViewCreateInfo *pCreateInfo,
-                                         const VkPhysicalDeviceLimits *device_limits, const Location &loc) const {
+                                         const Location &loc) const {
     bool skip = false;
 
     const VkDeviceSize &range = pCreateInfo->range;
     const VkFormat format = pCreateInfo->format;
-    const uint32_t format_size = vkuFormatElementSize(format);
+    const VKU_FORMAT_INFO format_info = vkuGetFormatInfo(format);
```
spencer-lunarg, line 53: I realize there is no way to get the `texel_per_block` value out of the Format Utils besides doing `vkuGetFormatInfo`

layers/core_checks/cc_buffer.cpp
```diff
@@ -45,33 +45,34 @@ bool CoreChecks::ValidateBufferUsageFlags(const LogObjectList &objlist, vvl::Buf
...
             skip |=
                 LogError("VUID-VkBufferViewCreateInfo-range-00929", buffer_state.buffer(), loc.dot(Field::range),
                          "(%" PRIuLEAST64 ") does not equal VK_WHOLE_SIZE, range must be a multiple of the element size (%" PRIu32
                          ") of the format %s.",
-                         range, format_size, string_VkFormat(format));
+                         range, format_info.block_size, string_VkFormat(format));
         }
-        // Range divided by the element size of format must be less than or equal to VkPhysicalDeviceLimits::maxTexelBufferElements
-        if (SafeDivision(range, format_size) > device_limits->maxTexelBufferElements) {
+        const uint32_t texels = static_cast<uint32_t>(SafeDivision(range, format_info.block_size)) * format_info.texel_per_block;
```
arno-lunarg, line 68: That division is not the one mentioned in the VU, you need to cast to float or double, do the division, then take the integer part of the result as the block count

spencer-lunarg: what you really mean is you want me to do a ceiling on the value right?

arno-lunarg (approved):

Issue #7241, opened 2024-01-04
VUID-vkSetDebugUtilsObjectNameEXT-pNameInfo-07874 behaves incorrectly


kainino0x:
**Environment:**
 - OS: Linux
 - GPU and driver version: driver agnostic - seen on nvidia, intel, and `llvmpipe: Mesa 23.2.1-1+build1 (LLVM 16.0.6) 0.0.1 [Selected]` which is the one I debugged against. (Also for some reason it does not fire when run against SwiftShader. Maybe we are just not setting up the VVL correctly in that case.)
 - SDK or header version if building from repo: This started happening in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7204
    Please see the links in the commit summary of https://chromium-review.googlesource.com/c/chromium/src/+/5166194 for the exact versions of all of the Vulkan dependencies (the later ends of all of the commit ranges linked)
 - Options enabled (synchronization, best practices, etc.):

**Describe the Issue**

When calling vkSetDebugUtilsObjectNameEXT to set the name of a VkDevice, we get the following error:

`STDERR: Error: VUID-vkSetDebugUtilsObjectNameEXT-pNameInfo-07874: Validation Error: [ VUID-vkSetDebugUtilsObjectNameEXT-pNameInfo-07874 ] Object 0: handle = 0x3b84009b4000, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0xbcf2c383 | vkSetDebugUtilsObjectNameEXT(): pNameInfo->objectHandle (VkDevice 0x3b8401e41c00) was created, allocated or retrieved from VkDevice 0x0[], but command is using (or its dispatchable parameter is associated with) VkDevice 0x3b8401e41c00[]. The Vulkan spec states: If pNameInfo->objectHandle is the valid handle of a device-level object, that object must be a descendent of the VkDevice identified by device (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkSetDebugUtilsObjectNameEXT-pNameInfo-07874)`

Note the message says the type is `VK_OBJECT_TYPE_INSTANCE` - but as far as I can tell from my debugger, this is wrong, and it is actually `VK_OBJECT_TYPE_DEVICE`.

**Expected behavior**

If this is indeed a VkDevice, then its "parent" VkDevice should be itself, not 0x0 as the error message says, which would make this call valid.

[...]

PR #7243, opened 2024-01-04
layers: Fix SetDebugUtils setting its own VkDevice

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7241

### DIFF
--- a/layers/object_tracker/object_tracker_utils.cpp
+++ b/layers/object_tracker/object_tracker_utils.cpp
@@ -1,7 +1,7 @@
-/* Copyright (c) 2015-2023 The Khronos Group Inc.
- * Copyright (c) 2015-2023 Valve Corporation
- * Copyright (c) 2015-2023 LunarG, Inc.
- * Copyright (C) 2015-2023 Google Inc.
+/* Copyright (c) 2015-2024 The Khronos Group Inc.
+ * Copyright (c) 2015-2024 Valve Corporation
+ * Copyright (c) 2015-2024 LunarG, Inc.
+ * Copyright (C) 2015-2024 Google Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


layers/object_tracker/object_tracker_utils.cpp
```diff
@@ -1187,10 +1187,23 @@ bool ObjectLifetimes::PreCallValidateSetDebugUtilsObjectNameEXT(VkDevice device,
...
-    skip |= ValidateAnonymousObject(
-        pNameInfo->objectHandle, pNameInfo->objectType, "VUID-VkDebugUtilsObjectNameInfoEXT-objectType-02590",
-        "VUID-vkSetDebugUtilsObjectNameEXT-pNameInfo-07874", error_obj.location.dot(Field::pNameInfo).dot(Field::objectHandle));
+    const VkObjectType object_type = pNameInfo->objectType;
+    const uint64_t object_handle = pNameInfo->objectHandle;
+
+    if (IsInstanceVkObjectType(object_type)) {
+        // TODO - need to check if device is from a valid instance/physical device
+        // VUID-vkSetDebugUtilsObjectNameEXT-pNameInfo-07872 /  VUID-vkSetDebugUtilsObjectNameEXT-pNameInfo-07873
+    } else if (object_type == VK_OBJECT_TYPE_DEVICE) {
```
spencer-lunarg, line 1196: note: this case is same as `SetPrivateData` and is to make a more clear/direct error message since we have both handles right here (and the `VkDevice` object hash map doesn't have itself)

artem-lunarg: Makes sense. I also learnt that VkDevice is a device-level object..

artem-lunarg (approved):
LGTM

Issue #3706, opened 2022-01-27
vkGetPhysicalDeviceToolProperties implementation in the validation layer is buggy


tomek-brcm:
I'm seeing 2 problems with the vkGetPhysicalDeviceToolProperties() implementation inside validation layer when it is called with a non-null pToolProperties pointer:

1. The implementation calls the next layer down with *pToolCount value decremented by 1, even if the value was 0 to start with.
2. The implementation increments *pToolCount value even if this would exceed the capacity. I didn't check if memory past the end of pToolProperties is trampled in this case.

The 1st bug would cause any layer below validation to receive a pointer to an array with 4 billion entries reported due to unsigned underflow. Those layer(s) will thrash memory by writing pToolProperties that way. 

The 2nd bug violates the spec ("on return the variable is overwritten with the number of structures actually written to pToolProperties") by reporting more items than actually written. Or worse: it tells the true after writing past the end of the array.

I think the Vulkan spec is somewhat unhelpful as it would be good to receive an unclamped *pToolCount value in the VK_INCOMPLETE case. The behaviour specified here matches similar queries all around Vulkan spec so it makes sense to make it consistent, even if this isn't ideal.

PR #7244, opened 2024-01-04
layers: Fix vkGetPhysicalDeviceToolProperties

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/3706

### DIFF
--- a/layers/vulkan/generated/chassis.cpp
+++ b/layers/vulkan/generated/chassis.cpp
@@ -3,11 +3,11 @@
 
 /***************************************************************************
  *
- * Copyright (c) 2015-2023 The Khronos Group Inc.
- * Copyright (c) 2015-2023 Valve Corporation
- * Copyright (c) 2015-2023 LunarG, Inc.
- * Copyright (c) 2015-2023 Google Inc.
- * Copyright (c) 2023-2023 RasterGrid Kft.
+ * Copyright (c) 2015-2024 The Khronos Group Inc.
+ * Copyright (c) 2015-2024 Valve Corporation
+ * Copyright (c) 2015-2024 LunarG, Inc.
+ * Copyright (c) 2015-2024 Google Inc.
+ * Copyright (c) 2023-2024 RasterGrid Kft.
  *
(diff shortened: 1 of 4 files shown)


layers/vulkan/generated/chassis.cpp
```diff
@@ -983,6 +983,15 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateBuffer(VkDevice device, const VkBufferCreat
 }
 
 // Handle tooling queries manually as this is a request for layer information
+static const VkPhysicalDeviceToolPropertiesEXT khronos_layer_tool_props = {
```
arno-lunarg, line 986: I do not get why this was moved?

layers/vulkan/generated/chassis.cpp
```diff
@@ -1028,6 +1027,7 @@ VKAPI_ATTR VkResult VKAPI_CALL GetPhysicalDeviceToolPropertiesEXT(VkPhysicalDevi
     if (original_pToolProperties != nullptr) {
         pToolProperties = original_pToolProperties;
     }
+    assert(*pToolCount != std::numeric_limits<uint32_t>::max());
```
arno-lunarg, line 1030: This does not fix the overflow, in case `*pToolCount` is at max u32 we should not append to the list?

spencer-lunarg: there is no mention in the spec what to do because you aren't going to have 4-billion tools to query, so if this is set to UINT_MAX, there was a bug somewhere else, but honestly not sure what you would want the layer to do in this fictional use case

layers/vulkan/generated/chassis.cpp
spencer-lunarg: there were 2 copies of it in the core and EXT function

layers/vulkan/generated/chassis.cpp
arno-lunarg: ok then

arno-lunarg (approved):

PR #7245, opened 2024-01-04
tests: Pass vkt::Device at RTX object creation

spencer-lunarg:
This is consistent with `binding.h` and `vkt::rt::Pipeline` as well

the idea is we should only need to pass in the `vkt::Device` once at the creation time, from there it can be known/moved around where needed

when calling `BuildCmdBuffer` you already should know the `VkDevice` to be used at this point


arno-lunarg (approved):

arno-lunarg:
nice

PR #7248, opened 2024-01-05
layers: Add exportFromImportedHandleTypes for SyncFD

spencer-lunarg:
Adds `VUID-VkSemaphoreGetFdInfoKHR-semaphore-01133` and `VUID-VkFenceGetFdInfoKHR-fence-01455`

### DIFF
--- a/layers/core_checks/cc_external_object.cpp
+++ b/layers/core_checks/cc_external_object.cpp
@@ -21,6 +21,24 @@
 #include "generated/chassis.h"
 #include "core_validation.h"
 
+static bool CanSemaphoreExportFromImported(VkPhysicalDevice physical_device, VkExternalSemaphoreHandleTypeFlagBits export_type,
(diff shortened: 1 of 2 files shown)


layers/core_checks/cc_external_object.cpp
```diff
@@ -21,6 +21,24 @@
 #include "generated/chassis.h"
 #include "core_validation.h"
 
+static bool CanSemaphoreExportFromImported(VkPhysicalDevice physical_device, VkExternalSemaphoreHandleTypeFlagBits export_type,
```
artem-lunarg, line 24: Just to say that `export_type` is a good name to use. The naming in the API is less clear about this and it took for me time to be sure it's not a typo in spec. A bit better naming removes any possibility of confusion.

artem-lunarg (approved):

PR #7251, opened 2024-01-05
gpu: Workaround for multiple variables using a single binding

### LINKED ISSUE #7102: False positive for image view type of dynamically unused descriptors
**Environment:**
 - OS: Windows 11
 - GPU: RTX 3070
 - SDK version: 1.3.268.0, with validation layers manually built from commit 55a016893749388c61d1bd4359b6b4b661169917
 - Options enabled: eBestPractices, eSynchronizationValidation, eGpuAssistedReserveBindingSlot, eGpuAssisted

**Describe the Issue**

I use a large descriptor binding array with descriptor indexing to simulate bindless textures.
In this descriptor binding array, I store both cubemap images (for skyboxes) and regular 2D images (for materials).

My material lighting shader may or may not sample 2D textures from this descriptor binding, depending on a signed integer index passed via an uniform buffer (-1 means "use a predefined value instead of sampling a texture").

In my test scene, I have a skybox (cubemap image stored at index 0) and a rendered object which uses a textureless material (all texture indices at -1).
In this scene, no texture read actually happen in the material lighting shader.

However, I am getting the following validation error:

> Validation Error: [ VUID-vkCmdDraw-viewType-07752 ] Object 0: handle = 0xb9181f0000000029, type = VK_OBJECT_TYPE_DESCRIPTOR_SET; Object 1: handle = 0xd22e030000001406, type = VK_OBJECT_TYPE_IMAGE_VIEW; | MessageID = 0x73ed4ff5 | vkCmdDraw():  the descriptor (VkDescriptorSet 0xb9181f0000000029[], binding 0, index 0) ImageView type is VK_IMAGE_VIEW_TYPE_CUBE but the OpTypeImage has (Dim = 2D) and (Arrayed = 0). The Vulkan spec states: If a VkImageView is accessed as a result of this command, then the image view's viewType must match the Dim operand of the OpTypeImage as described in Instruction/Sampler/Image View Validation (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdDraw-viewType-07752)

This validation error would be correct if I was actually sampling a texture from the descriptor set. However, this is not the case.

Note that this issue does **not** happen with the validation layers bundled with SDK 1.3.268.0.

**Expected behavior**

No validation error, since the cubemap at index 0 is not sampled from in the material lighting shader.

**Valid Usage ID**
VUID-vkCmdDraw-viewType-07752

jeremyg-lunarg:
For bindless, it is common for multiple variables with different  types (such as 2d and 3d images) to be attached to a single binding.  Avoid false positives in this case by looking for a compatible variable definition to validate against rather than just using the first one found.

A better fix would be to have the spirv instrumentation record the variable id that produced each access and then use that variable definiton for validation. However that will need to be done later.

Fixes #7102


spencer-lunarg (approved):
LGTM ... I rather have a proper fix in the shader instrumentation (https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7255)... but also rather have this in the next SDK, so at least you have the change isolated to help when I do the fix in the shader instrumentation soon

jeremyg-lunarg:
> LGTM ... I rather have a proper fix in the shader instrumentation (#7255)... but also rather have this in the next SDK, so at least you have the change isolated to help when I do the fix in the shader instrumentation soon

I agree. It isn't really practical to change the instrumentation before the next SDK, which is why I made this workaround PR.

Issue #6876, opened 2023-10-31
Add Missing Win32 External Memory/Sync VUs


spencer-lunarg:
These mostly have a `FD` counterpart already added, but I personally don't have a Windows device to test any changes to the Win32 side

- [x] `VUID-VkMemoryDedicatedAllocateInfo-image-01876` (like VUID-VkMemoryDedicatedAllocateInfo-image-01878)
- [x] `VUID-VkMemoryDedicatedAllocateInfo-buffer-01877` (like VUID-VkMemoryDedicatedAllocateInfo-buffer-01879)
- [x] `VUID-VkMemoryGetWin32HandleInfoKHR-handleType-00662` (like VUID-VkMemoryGetFdInfoKHR-handleType-00671)
- [x] `VUID-VkSemaphoreGetWin32HandleInfoKHR-semaphore-01128 `(like VUID-VkSemaphoreGetFdInfoKHR-semaphore-01133)
- [x] `VUID-VkFenceGetWin32HandleInfoKHR-fence-01450` (like VUID-VkFenceGetFdInfoKHR-fence-01455)
- [x] `VUID-VkMemoryAllocateInfo-memoryTypeIndex-00645` (like VUID-VkMemoryAllocateInfo-memoryTypeIndex-01744)
- [ ] `VUID-VkImportMemoryWin32HandleInfoKHR-handleType-00658` (like  VUID-VkImportMemoryFdInfoKHR-handleType-00667)

PR #7252, opened 2024-01-05
layers: Add external memory VU 00658

artem-lunarg:
Part of https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6876

This does not contain any test. For testing this requires exporting handle outside of Vulkan API. In practice this would be DX11/DX12 resource. It's possible in theory but requires support for DX initialization within test framework.


layers/core_checks/cc_device_memory.cpp
```diff
@@ -798,18 +798,39 @@ bool CoreChecks::ValidateBindBufferMemory(VkBuffer buffer, VkDeviceMemory memory
...
+            // Check if buffer can be used with memory imported from handle type
+            {
+                VkPhysicalDeviceExternalBufferInfo external_info = vku::InitStructHelper();
+                external_info.flags = buffer_state->createInfo.flags;
+                // TODO: for now, there is no VkBufferUsageFlags2KHR flag that exceeds 32-bit but should be revisited later
+                external_info.usage = VkBufferUsageFlags(buffer_state->usage);
+                external_info.handleType = mem_info->import_handle_type.value();
+                VkExternalBufferProperties external_properties = vku::InitStructHelper();
+                DispatchGetPhysicalDeviceExternalBufferProperties(physical_device, &external_info, &external_properties);
+                if (!(external_properties.externalMemoryProperties.externalMemoryFeatures &
```
spencer-lunarg, line 824: was there a reason no to use `HasExternalMemoryImportSupport` for these?

spencer-lunarg (approved):
LGTM - just 1 question first

layers/core_checks/cc_device_memory.cpp
artem-lunarg: Nope. Thanks for the suggestion.

PR #7263, opened 2024-01-08
tests: Fix closing handles in external sync test

ziga-lunarg:
On windows handle and handle2 would be the same, so closing it twice would throw an exception.


artem-lunarg:
"On windows handle and handle2 would be the same"

Is it something you experienced? I checked for this on my machine and it was different, but I see it could be just luck.

artem-lunarg (approved):

ziga-lunarg:
> "On windows handle and handle2 would be the same"
> 
> Is it something you experienced? I checked for this on my machine and the handles were different, but I see it could be just luck.

Yes, happens on my windows nvidia machine, can reproduce it 100% of the time

PR #7267, opened 2024-01-09
layers: Cleanup with External Handle messages

spencer-lunarg:
Some misc cleanup of the external handle VUs added recently


layers/core_checks/cc_device_memory.cpp
```diff
@@ -201,7 +201,7 @@ bool CoreChecks::IsZeroAllocationSizeAllowed(const VkMemoryAllocateInfo *pAlloca
     return false;
 }
 
-bool CoreChecks::HasExternalMemoryImportSupport(const vvl::Buffer &buffer, VkExternalMemoryHandleTypeFlagBits handle_type) const {
+bool CoreChecks::HasExternalMemoryBufferSupport(const vvl::Buffer &buffer, VkExternalMemoryHandleTypeFlagBits handle_type) const {
```
artem-lunarg, line 204: There is now also `HasExternalMemoryImportSupport` for images, if rename is needed probably should be renamed to. Also do we need a rename? The name was focused on import operation support (could also be export).

artem-lunarg (approved):
LGTM, the question about naming

layers/core_checks/cc_device_memory.cpp
spencer-lunarg: sorry, like wow, I literally read `Import` as `Image` and thought this was `HasExternalMemoryImageSupport`...... ok, let me revert this

PR #7277, opened 2024-01-10
Use CTest in Github Actions

spencer-lunarg:
(note first commit is to just move non-common code from `common_ci.py` to `tests.py`)

This makes use of `ctest` over the traditional `gtest` flow

The main advantage (will update time improvements when CI is done) is to have the tests run in parallel since using MockICD, there is zero risk when the needing the share the same kernel context as there is with a real GPU


spencer-lunarg:
closing, seems `ctest` is much slower (like 2x or more) on GitHub action with full parallel runs then just using gtests

PR #7283, opened 2024-01-10
tests: Test for fixed subsampled image range generation

aqnuep:


aqnuep:
@spencer-lunarg, it looks like the LunarG CI failed with an unrelated GPU AV failure. I'll do a rebasing just in case.

lunarpapillo:
LunarG CI abort seems to have been a system issue.  Retriggered.

PR #7284, opened 2024-01-10
tests: Device profile updates for video

aqnuep:
I noticed that despite the recent MockICD updates that enable the video related VVL test cases to run against MockICD, the tests still don't run on the CI.

The reason is that there are no profile-less runs, so there was a need to update the max profile to have everything in place for this to work. This also uncovered a Vulkan-Profiles bug that I've fixed here: https://github.com/KhronosGroup/Vulkan-Profiles/pull/591

With these changes there really should be 100% test coverage of video using MockICD + Profiles, not just using MockICD alone.

UPDATE: Apparently this also uncovered a bug that did not show up on our CI, despite having address sanitizer enabled (likely a non-deterministic issue), but did show up here on the github CI, so this change fixes that too.


spencer-lunarg (approved):

lunarpapillo:
LunarG CI abort seems to have been a system issue. Retriggered.

PR #7287, opened 2024-01-11
layers: Add vertex type check for dynamic vert input

spencer-lunarg:
- Adds the missing `VUID-vkCmdDraw-Input-08734`
- Improve error messages 
- Unify the logic for the Pipeline and DynamicState/ShaderObject version of vertex input checking (since they are the same)

### DIFF
--- a/layers/core_checks/cc_cmd_buffer_dynamic.cpp
+++ b/layers/core_checks/cc_cmd_buffer_dynamic.cpp
@@ -1,7 +1,7 @@
-/* Copyright (c) 2015-2023 The Khronos Group Inc.
- * Copyright (c) 2015-2023 Valve Corporation
- * Copyright (c) 2015-2023 LunarG, Inc.
- * Copyright (C) 2015-2023 Google Inc.
+/* Copyright (c) 2015-2024 The Khronos Group Inc.
+ * Copyright (c) 2015-2024 Valve Corporation
+ * Copyright (c) 2015-2024 LunarG, Inc.
+ * Copyright (C) 2015-2024 Google Inc.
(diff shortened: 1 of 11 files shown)


layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -299,44 +299,55 @@ bool CoreChecks::ValidateDrawDynamicState(const LastBound& last_bound_state, con
...
+
+                    const uint32_t var_base_type_id = variable_ptr->base_type.ResultId();
+                    const uint32_t attribute_type = spirv::GetFormatType(description.format);
+                    const auto var_numeric_type = vert_spirv_state->GetNumericType(var_base_type_id);
+
+                    const bool attribute64 = vkuFormatIs64bit(description.format);
+                    const bool shader64 = vert_spirv_state->GetBaseTypeInstruction(var_base_type_id)->GetBitWidth() == 64;
+
+                    // first type check before doing 64-bit matching
+                    if (!(attribute_type & var_numeric_type)) {
```
artem-lunarg, line 311: nit: not very familiar with this code I read this as (a && b) and could not figure out what's going on here (and the following conditions are real boolean logic, so brain gets easily confused). Realized the next day it's a bit operation :) One suggestion is to use `!= 0` to make this obvious.

artem-lunarg (approved):

layers/core_checks/cc_cmd_buffer_dynamic.cpp
spencer-lunarg: agree, there are some other "improve logic" nit around the shader interface stuff (like the fact one of these is a enum and one is a `uint32_t`)... will add this to the TODO to that upcoming PR

Issue #7282, opened 2024-01-10
Missing VUID-VkGraphicsPipelineCreateInfo-layout-07827 with null pipeline layout


rg3igalia:
**Environment:**
 - OS: Linux
 - GPU and driver version: Intel Corporation UHD Graphics 620 with the ANV Mesa driver.
 - SDK or header version if building from repo: 7f5125f4f61eba12f54b98cc82501e07cd7968e0
 - Options enabled (synchronization, best practices, etc.): defaults+synchronization

**Describe the Issue**

A common mistake in some VK-GL-CTS tests is forgetting to specify the pipeline layout in the final link step when creating a graphics pipeline using 4 graphics pipeline libraries. This is illegal according to VUID-VkGraphicsPipelineCreateInfo-layout-07827, which [reads](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkGraphicsPipelineCreateInfo.html#VUID-VkGraphicsPipelineCreateInfo-layout-07827):

> If the pipeline includes a [complete set of state](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#pipelines-graphics-subsets-complete) specified entirely by libraries, and each library was created with a [VkPipelineLayout](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPipelineLayout.html) created without VK_PIPELINE_LAYOUT_CREATE_INDEPENDENT_SETS_BIT_EXT, then layout must be [compatible](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#descriptorsets-compatibility) with the layouts in those libraries

However, the layers stay silent about this problem. It can be reproduced with VK-GL-CTS commit f005a8d65390d1a1727de63ef8a19be5cf2ce554 (main as of the time I'm writing this) when running the test dEQP-VK.pipeline.pipeline_library.vertex_input.misc.stride_change_vert_frag, for example (fix is currently under review).

**Expected behavior**

Receiving an error mentioning VUID-VkGraphicsPipelineCreateInfo-layout-07827.

**Valid Usage ID**

VUID-VkGraphicsPipelineCreateInfo-layout-07827

**Additional context**

The following patch fixes the issue for me, but I'm not sure if it's correct:

```diff
diff --git a/layers/core_checks/cc_pipeline_graphics.cpp b/layers/core_checks/cc_pipeline_graphics.cpp
[...]

PR #7289, opened 2024-01-11
layers: Add check for GPL linking layout

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7282


layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -529,6 +529,10 @@ bool CoreChecks::ValidateGraphicsPipelineLibrary(const vvl::Pipeline &pipeline,
                         LogError("VUID-VkGraphicsPipelineCreateInfo-layout-07827", objlist, create_info_loc.dot(Field::layout),
                                  "is incompatible with the layout specified in the fragment shader sub-state: %s", err_msg.c_str());
                 }
+            } else {
+                skip |= LogError("VUID-VkGraphicsPipelineCreateInfo-layout-07827", device, create_info_loc.dot(Field::layout),
+                         "is missing and therefore not ccompatible with the libraries layout");
```
rg3igalia, line 534:
```suggestion
                         "is missing and therefore not compatible with the libraries layout");
```

artem-lunarg (approved):

PR #7271, opened 2024-01-09
layers: Add 03797 03711 03735

arno-lunarg:
Part of https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/3792

**VUID-VkAccelerationStructureGeometryTrianglesDataKHR-vertexFormat-03797**
The [format features](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#resources-buffer-view-format-features) of vertexFormat must contain VK_FORMAT_FEATURE_ACCELERATION_STRUCTURE_VERTEX_BUFFER_BIT_KHR

**VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03711**
For any element of pInfos[i].pGeometries or pInfos[i].ppGeometries with a geometryType of VK_GEOMETRY_TYPE_TRIANGLES_KHR, geometry.triangles.vertexData.deviceAddress must be aligned to the size in bytes of the smallest component of the format in vertexFormat

**VUID-VkAccelerationStructureGeometryTrianglesDataKHR-vertexStride-03735**
vertexStride must be a multiple of the size in bytes of the smallest component of vertexFormat

PR #7291, opened 2024-01-11
layers: Improve buffer address validation messages

arno-lunarg:
A number of VU states something along the lines of "non sparse buffers accessed through a device address must be bound to memory" The current error message was not taking advantage of the error message generation in BufferAddressValidation, it now does

Then:
>Validation Error: [ VUID-vkCmdBindDescriptorBuffersEXT-pBindingInfos-08052 ] Object 0: handle = 0x204b7943830, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0x9fde6b0000000014, type = VK_OBJECT_TYPE_BUFFER; | MessageID = 0x17e3f381 | vkCmdBindDescriptorBuffersEXT(): pBindingInfos[0].address (VkBuffer 0x9fde6b0000000014[]) used with no memory bound and previously bound memory was freed. Memory must not be freed prior to this operation.

Now:
>Validation Error: [ VUID-vkCmdBindDescriptorBuffersEXT-pBindingInfos-08052 ] Object 0: handle = 0x9fde6b0000000014, type = VK_OBJECT_TYPE_BUFFER; | MessageID = 0x17e3f381 | vkCmdBindDescriptorBuffersEXT(): pBindingInfos[0].address (0x13ca7000) has no valid buffer associated to it. At least one buffer associated to this device address must be valid. The following buffers are not bound to memory or it has been freed:
Object 1: buffer is bound to memory (VkDeviceMemory 0xdd3a8a0000000015[]) but it has been freed

The important part is the mention that "At least one buffer associated to this device address must be valid". reminding that when multiple VUs refer to the same device address, valid usage passes if and only if at least one buffer satisfies all of those VUs

Need to update https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7290 and https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7271 after that

### DIFF
--- a/layers/core_checks/cc_buffer_address.h
+++ b/layers/core_checks/cc_buffer_address.h
@@ -59,7 +59,8 @@
    For an example of how to use BufferAddressValidation, see for instance how "VUID-VkDescriptorBufferBindingInfoEXT-usage-08122"
    and friends are validated.
  */
-template <size_t N>
+
+template <size_t N = 1>
 class BufferAddressValidation {
   public:
     // Return true if and only if VU is verified
@@ -113,6 +114,27 @@ class BufferAddressValidation {
             }
         }
         return false;
+    }
+
+    static bool ValidateMemoryBoundToBuffer(const CoreChecks& validator,
+                                            const ValidationStateTracker::BUFFER_STATE_PTR& buffer_state,
+                                            std::string* out_error_msg) {
+        if (!buffer_state->sparse && !buffer_state->IsMemoryBound()) {
+            if (out_error_msg) {
+                if (const auto mem_state = buffer_state->MemState(); mem_state && mem_state->Destroyed()) {
+                    *out_error_msg +=
+                        "buffer is bound to memory (" + validator.FormatHandle(mem_state->Handle()) + ") but it has been freed";
+                } else {
+                    *out_error_msg += "buffer has not been bound to memory";
+                }
+            }
+            return false;
+        }
+        return true;
+    }
+
+    static std::string ValidateMemoryBoundToBufferErrorMsgHeader() {
+        return "The following buffers are not bound to memory or it has been freed:\n";
     }
 
   public:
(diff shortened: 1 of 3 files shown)


layers/core_checks/cc_descriptor.cpp
```diff
@@ -2348,14 +2348,21 @@ bool CoreChecks::PreCallValidateCmdBindDescriptorBuffersEXT(VkCommandBuffer comm
...
             BufferAddressValidation<5> buffer_address_validator = {{{
                 {"VUID-vkCmdBindDescriptorBuffersEXT-pBindingInfos-08052", LogObjectList(device),
-                 [this, commandBuffer, binding_loc](const BUFFER_STATE_PTR &buffer_state, std::string *out_error_msg) {
-                     if (!out_error_msg) {
-                         return !buffer_state->sparse && buffer_state->IsMemoryBound();
-                     } else {
-                         return ValidateMemoryIsBoundToBuffer(commandBuffer, *buffer_state, binding_loc.dot(Field::address),
-                                                              "VUID-vkCmdBindDescriptorBuffersEXT-pBindingInfos-08052");
+                 [this](const BUFFER_STATE_PTR &buffer_state, std::string *out_error_msg) {
+                     if (!buffer_state->sparse && !buffer_state->IsMemoryBound()) {
```
spencer-lunarg, line 2352: it seems the logic in ` if (!buffer_state->sparse && !buffer_state->IsMemoryBound()) ` is all now unified, which is great!

but now it begs the questions, why do we copy and paste this in every lambda, this seems like it could be a separate function they all call

spencer-lunarg (approved):

spencer-lunarg (approved):

layers/core_checks/cc_descriptor.cpp
arno-lunarg: yeeeees you are right, I made two helper functions to help with copy pasta, they live in `cc_buffer_address.h`

layers/core_checks/cc_buffer_address.h
```diff
@@ -59,6 +59,9 @@
    For an example of how to use BufferAddressValidation, see for instance how "VUID-VkDescriptorBufferBindingInfoEXT-usage-08122"
    and friends are validated.
  */
+
+namespace bda {
```
spencer-lunarg, line 63: I guess this namespace feels confusing because `bda` is `Buffer Device Address` but the class is only `Buffer Address` ... but if you change it to `BufferDeviceAddress` then this namespaces feels redundant

spencer-lunarg (approved):

layers/core_checks/cc_buffer_address.h
jeremyg-lunarg: Really core validation should have its own namespace and this code should just be part of it.

PR #7292, opened 2024-01-12
sync: Fix loop and stop using uint8_t as size type

artem-lunarg:
The typo in the loop was not discovered because after running for 256 iterations the loop terminated due to uint8_t wrap around. Using uint32_t makes similar issues more noticeable.

We always had this typo. Not sure which scenarios this could affect, but as the first step this should be fixed.

Not the first issue with uint8_t we encountered, so start using uint32_t as default size type for small_vector. This does not affect the final size of the container due to alignment gaps. For specific case of small_vector<ResourceFirstAccess, 3> it's 96 bytes both for uint8_t and uint32_t size type parameterization.

### DIFF
--- a/layers/containers/custom_containers.h
+++ b/layers/containers/custom_containers.h
@@ -1,6 +1,6 @@
-/* Copyright (c) 2015-2017, 2019-2023 The Khronos Group Inc.
- * Copyright (c) 2015-2017, 2019-2023 Valve Corporation
- * Copyright (c) 2015-2017, 2019-2023 LunarG, Inc.
+/* Copyright (c) 2015-2017, 2019-2024 The Khronos Group Inc.
+ * Copyright (c) 2015-2017, 2019-2024 Valve Corporation
+ * Copyright (c) 2015-2017, 2019-2024 LunarG, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


layers/containers/custom_containers.h
```diff
@@ -118,7 +118,7 @@ using insert_iterator = std::insert_iterator<T>;
 //       MoveAssignable and MoveConstructable
 // NOTE: Unlike std::vector, iterators are invalidated by move assignment between small_vector objects effectively the
 //       "small string" allocation functions as an incompatible allocator.
-template <typename T, size_t N, typename SizeType = uint8_t>
+template <typename T, size_t N, typename SizeType = uint32_t>
```
jeremyg-lunarg, line 121: Thank you. This is like the 3-4th time we've been burned by the uint8_t here.

jeremyg-lunarg (approved):

layers/sync/sync_access_state.cpp
```diff
@@ -161,7 +161,7 @@ HazardResult ResourceAccessState::DetectHazard(const ResourceAccessState &record
             --count;
         }
 
-        for (Size i = 0; i < count; ++count) {
+        for (Size i = 0; i < count; ++i) {
```
jzulauf-lunarg, line 164: What kind of test/logic hole does this not having an obvious hang from this?

artem-lunarg: I'm not sure I follow the question. Could you clarify?

artem-lunarg: If how it worked without this, I don't know. Maybe indeed we don't have a test. Or also it's not even clear if this logic is 100% correct because there was always a bug here. So in theory fixing this could have some side effect.

jzulauf-lunarg: We were  always counting up to numeric_limits<Size>::max() which for char is *slow*, but not an apparent "hang" which it would look like for uint32_t or uint64_t.  

Apparently we don't have any tests where detection of entry > 0 hazard is required to pass.

PR #7293, opened 2024-01-12
codegen: Create vk_dispatch_table_helper.cpp

charles-lunarg:
Contains all the implementations of functions defined in vk_dispatch_table_helper.h. This improves debug test times by not having each translation unit initialize api_extension_map whenever the dll is loaded.


jeremyg-lunarg (approved):
LGTM

Issue #7261, opened 2024-01-08
Segfault when validating dynamic blend state 


gerddie:
**Environment:**
 - OS:  Linux
 - GPU and driver version: AMD, radv  (mesa-gita921a6901010)
 - SDK or header version if building from repo: 1.3.268 (but also tested with git head)  
 - Options enabled (synchronization, best practices, etc.): 
   * Core (all)
   * Handle wrapping
   * Stateless parameter 
   * Object lifetime  
   * Thread safety 
  
**Describe the Issue**

When running an apitrace of *Quern - undying thoughts* using *Mesa/Zink* the validation layer crashes in `CoreChecks::ValidateDrawDynamicStatePipeline`  on [line 524](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/02f1cb7d55aa808bbe1401cb0648f7057c98485a/layers/core_checks/cc_cmd_buffer_dynamic.cpp#L524) with a `nullptr`  dereference. (Below the backtrace is for version 1.3.268, so the line in question is 408)

**Expected behavior**

The validation layer should not crash, and possibly report an validation error if blending is enabled for non-allocated attachments (not sure about this). 

**Additional context**

The framebuffer has only one color attachment at this point, but because in `CMD_BUFFER_STATE::BeginRendering` the vector holding the attachments is allocated as `(pRenderingInfo->colorAttachmentCount + 2) * 2` and some attachments are initialized to `nullptr` in the loop that follows, because of that we end up with the `nullptr` deref. Working around this by checking for as  `nullptr` and skipping the attachment in this case results in a new validation error, because the DS surface that is also added to the vector of attachments  is checked for `VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BLEND_BIT` and this check fails.

There are two problems here: On one hand if the dynamic blend state  is enabled for an attachment that is not available we get a crash. Here the format feature test should either be skipped for this attachment or a validation error should be reported (I haven't yet found the relevant part of the spec that  tells me which would be the right answer). 

The second problem is that the validation of the dynamic blend state iterates over the whole vector of attachments as created in 
`CMD_BUFFER_STATE::BeginRendering`, but it should probably only  iterate over the first `cb_state.GetDynamicColorAttachmentCount()` elements.  

<details>
<summary>code or terminal output</summary>

```sh
4581357: error: caught an unhandled exception
[...]

PR #7297, opened 2024-01-14
layers: Fix dynamic rendering crash

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7261

### DIFF
--- a/layers/state_tracker/cmd_buffer_state.cpp
+++ b/layers/state_tracker/cmd_buffer_state.cpp
@@ -1,7 +1,7 @@
-/* Copyright (c) 2015-2023 The Khronos Group Inc.
- * Copyright (c) 2015-2023 Valve Corporation
- * Copyright (c) 2015-2023 LunarG, Inc.
- * Copyright (C) 2015-2023 Google Inc.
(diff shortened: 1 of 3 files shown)


layers/state_tracker/cmd_buffer_state.cpp
```diff
@@ -590,6 +590,9 @@ void CommandBuffer::BeginRendering(Func command, const VkRenderingInfo *pRenderi
     hasRenderPassInstance = true;
 
     active_attachments = nullptr;
+    // add 2 for the Depth and Stencil
+    // multiple by 2 because every attachment might have a resolve
+    // Currently reserve the maximum possible size for |active_attachments| so when looping, we NEED to check for null
     uint32_t attachment_count = (pRenderingInfo->colorAttachmentCount + 2) * 2;
```
spencer-lunarg, line 596: I would much rather this be only the items we need and have calls like `GetDynamicColorResolveAttachmentImageIndex()` adjust to where we put things, but would require a lot more refactoring, so for now, checking for `null` is what we need to be doing

artem-lunarg (approved):

PR #7299, opened 2024-01-15
layers: Cleanup on shader interface variables

spencer-lunarg:
@artem-lunarg pointed out some of these areas are not fast to review, so tried to clean things up a bit


artem-lunarg (approved):
Thanks for making this more explicit.

PR #7301, opened 2024-01-15
layers: Add missing implicit RasterizationStateCreateInfo

spencer-lunarg:
somehow the `VkPipelineRasterizationStateCreateInfo` implicit VUs were not added in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6824


arno-lunarg (approved):
Is there an explanation for the code gen not triggering??

spencer-lunarg:
> Is there an explanation for the code gen not triggering??

in `scripts/generators/stateless_validation_helper_generator.py` if you look above the change

> # There are cases where `noautovalidity` and `optional` are both true, but really are just guarded by other conditions.

PR #7302, opened 2024-01-15
Fix GN build after introduction of vk_dispatch_table_helper.cpp

Kangz:
@charles-lunarg please take a look! FYI @lokokung as next gardener


spencer-lunarg (approved):

charles-lunarg:
Apologies! Didn't even think to check the gn build when I made that dispatch_table_helper.cpp PR.

PR #7305, opened 2024-01-16
test: Add RETURN_IF_SKIP for Wsi

spencer-lunarg:
Cleanup for WSI tests to use the new `RETURN_IF_SKIP` to detect if the swapchain/surface failed to create


arno-lunarg (approved):
Nice!

PR #7308, opened 2024-01-17
Fix more implicit VUID labeling

spencer-lunarg:
Various fixes from adding VUID labels to adding missing VUs that were missing

### DIFF
--- a/layers/error_message/unimplementable_validation.h
+++ b/layers/error_message/unimplementable_validation.h
@@ -76,6 +76,13 @@ const char* unimplementable_validation[] = {
     "VUID-VkPipelineDepthStencilStateCreateInfo-back-parameter",
     // VkClearValue
     "VUID-VkRenderingAttachmentInfo-clearValue-parameter",
+    // VkComponentMapping
+    "VUID-VkImageViewCreateInfo-components-parameter",
+    "VUID-VkSamplerYcbcrConversionCreateInfo-components-parameter",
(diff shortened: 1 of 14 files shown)


layers/error_message/unimplementable_validation.h
```diff
@@ -76,6 +76,13 @@ const char* unimplementable_validation[] = {
     "VUID-VkPipelineDepthStencilStateCreateInfo-back-parameter",
     // VkClearValue
     "VUID-VkRenderingAttachmentInfo-clearValue-parameter",
+    // VkComponentMapping
+    "VUID-VkImageViewCreateInfo-components-parameter",
```
artem-lunarg, line 80: I'm curious what could be the meaning of `VUID-VkImageViewCreateInfo-components-parameter`? There is separate VUID for each component (r,g,b,a) and if those components are correct enum values then entire `VkComponentMapping` should be correct? Or is it more to it?

artem-lunarg (approved):

layers/error_message/unimplementable_validation.h
spencer-lunarg: as you said there are separate VUs like `VUID-VkComponentMapping-r-parameter` that check for this

This VU here is something the spec generator script does automatically, the issue is there is no `sType` and the original goal behind these VUs was to make sure you called the right struct, then inside the struct it will have it's own VUs

since there is no `sType` in `VkComponentMapping` we really don't have anything that would trigger **only** `VUID-VkImageViewCreateInfo-components-parameter`

PR #7312, opened 2024-01-17
tests: Split up ArrayOOBGraphics

jeremyg-lunarg:
This test case is huge and hard to debug, so split it into a bunch of smaller tests.

### DIFF
--- a/tests/unit/gpu_av_descriptor_indexing.cpp
+++ b/tests/unit/gpu_av_descriptor_indexing.cpp
@@ -1,7 +1,7 @@
 /*
- * Copyright (c) 2020-2023 The Khronos Group Inc.
- * Copyright (c) 2020-2023 Valve Corporation
- * Copyright (c) 2020-2023 LunarG, Inc.
+ * Copyright (c) 2020-2024 The Khronos Group Inc.
+ * Copyright (c) 2020-2024 Valve Corporation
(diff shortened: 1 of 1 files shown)


tests/unit/gpu_av_descriptor_indexing.cpp
```diff
@@ -15,6 +15,7 @@
 #include "../framework/pipeline_helper.h"
 #include "../framework/descriptor_helper.h"
 #include "../framework/gpu_av_helper.h"
+#include <iostream>
```
jeremyg-lunarg, line 18: there's a bunch of debug prints in here for now. I'd like to convert them into some sort of gtest message at some point.

arno-lunarg: are they gone now, or am I just not seeing them??

arno-lunarg: Ah I guess that there are the `SCOPED_TRACE` calls?

arno-lunarg (approved):

tests/unit/gpu_av_descriptor_indexing.cpp
jeremyg-lunarg: yes, thats the best thing I could find in gtest. I still need to repush to remove this header.

spencer-lunarg (approved):

Issue #7262, opened 2024-01-08
Validation check when using struct VkPhysicalDeviceExtendedDynamicState3FeaturesEXT


tzdimi:
If  VkPhysicalDeviceExtendedDynamicState3FeaturesEXT struct is included at VkDeviceCreateInfo .pNext chain and  VK_EXT_extended_dynamic_state3 extension is not enabled there is no error message when trying to create the device that 
the extension isn't enabled.
Most probably related with #5012

spencer-lunarg:
also related to https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7189

This has been a long todo to get the code gen working (with proper testing to catch the many nasty edge cases), unfortunately didn't make it for the upcoming SDK but my goal is to get this done for the following SDK release

Issue #7189, opened 2023-12-25
Detect extensions missing for feature structs


spencer-lunarg:
currently if you have a Vulkan 1.2 `VkDevice` (on a 1.3 capable driver), add `VkPhysicalDeviceMaintenance4FeaturesKHR`, but never add the `VK_KHR_maintenance4` extension, it will not produce a warning and in-tern, stateless VUs (like `VUID-VkBufferCreateInfo-size-06409`) rely on the fact the feature bit was passed in, but the value assumed the extension was enabled

some simple tests

```cpp
TEST_F(VkLayerTest, MissingExtensionPhysicalDeviceFeature) {
    SetTargetApiVersion(VK_API_VERSION_1_2); // VK_KHR_maintenance4 added in 1.3
    RETURN_IF_SKIP(InitFramework());
    if (!DeviceExtensionSupported(VK_KHR_MAINTENANCE_4_EXTENSION_NAME)) {
        GTEST_SKIP() << "VK_KHR_maintenance4 not supported";
    }
    VkPhysicalDeviceMaintenance4FeaturesKHR maintenance4_feature = vku::InitStructHelper();
    maintenance4_feature.maintenance4 = VK_TRUE;
    RETURN_IF_SKIP(InitState(nullptr, &maintenance4_feature));
}

TEST_F(VkLayerTest, MissingExtensionStruct) {
    SetTargetApiVersion(VK_API_VERSION_1_1);
    RETURN_IF_SKIP(InitFramework());
    if (!DeviceExtensionSupported(VK_KHR_MAINTENANCE_5_EXTENSION_NAME)) {
        GTEST_SKIP() << "VK_KHR_maintenance5 not supported";
    }
    RETURN_IF_SKIP(InitState());

    vkt::Buffer buffer(*m_device, 32, VK_BUFFER_USAGE_UNIFORM_TEXEL_BUFFER_BIT);

    VkBufferUsageFlags2CreateInfoKHR buffer_usage_flags = vku::InitStructHelper();
    buffer_usage_flags.usage = VK_BUFFER_USAGE_2_UNIFORM_TEXEL_BUFFER_BIT_KHR;

    VkBufferViewCreateInfo buffer_view_ci = vku::InitStructHelper(&buffer_usage_flags);
    buffer_view_ci.format = VK_FORMAT_R8G8B8A8_UNORM;
    buffer_view_ci.range = VK_WHOLE_SIZE;
[...]

PR #7315, opened 2024-01-18
layers: Check extensions are with features structs

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7189
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7262

When including a feature structure at `vkCreateDevice` time, add logic to validate the corresponding extension is included too


arno-lunarg (approved):
LGTM, I just think the error message could be reworked a bit

PR #7318, opened 2024-01-18
layers: Fix HasExternalMemoryImportSupport with DRM modifiers

pH5:
Add the required `VkPhysicalDeviceImageDrmFormatModifierInfoEXT` extension structure to the `VkPhysicalDeviceImageFormatInfo2` `pNext` chain when calling `vkGetPhysicalDeviceImageFormatProperties2` with tiling set to `VK_IMAGE_TILING_DRM_FORMAT_MODIFIER_EXT`.

### DIFF
--- a/layers/core_checks/cc_device_memory.cpp
+++ b/layers/core_checks/cc_device_memory.cpp
@@ -21,6 +21,7 @@
 
 #include <vulkan/vk_enum_string_helper.h>
 #include "generated/chassis.h"
+#include "generated/pnext_chain_extraction.h"
 #include "core_validation.h"
 
 // For given mem object, verify that it is not null or UNBOUND, if it is, report error. Return skip value.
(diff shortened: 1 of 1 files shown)


layers/core_checks/cc_device_memory.cpp
```diff
@@ -212,7 +212,9 @@ bool CoreChecks::HasExternalMemoryImportSupport(const vvl::Buffer &buffer, VkExt
 }
 
 bool CoreChecks::HasExternalMemoryImportSupport(const vvl::Image &image, VkExternalMemoryHandleTypeFlagBits handle_type) const {
-    VkPhysicalDeviceExternalImageFormatInfo external_info = vku::InitStructHelper();
+    VkPhysicalDeviceImageDrmFormatModifierInfoEXT drm_format_modifier_info = vku::InitStructHelper();
```
artem-lunarg, line 215: a) In similar places we have a condition that tiling is DRM_FORMAT_MODIFIER. Should the same be done here? https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/d26b50b03815ff226e6df478b4ddc4b98d8deaee/layers/core_checks/cc_image.cpp#L206

b) In similar place we also initialized other fields of `VkPhysicalDeviceImageDrmFormatModifierInfoEXT`:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/d26b50b03815ff226e6df478b4ddc4b98d8deaee/layers/core_checks/cc_image.cpp#L245

spencer-lunarg: I think this is causing the `NegativeExternalMemorySync.ImportMemoryHandleType` test to fail

artem-lunarg (approved):
LGTM

spencer-lunarg:
@pH5 thanks for the help with the DRM fixes, they are always appreciated!

PR #7321, opened 2024-01-18
gpu: Redo "Don't wait for idle after every submit"

jeremyg-lunarg:
Restore commit bfaa879ee2:

Add virtual methods to the CommandBuffer and Queue classes that get run at PreRecord, PostRecord and Retire time. This makes it possible to remove the PostRecord wait for idle from GPU-AV submissions. Instead, at timeline semaphore is set up for host waits in the Retire() virtual method so that GPU-AV post processing for each submission is appropriately synchronized.


layers/gpu_validation/gpu_state_tracker.cpp
```diff
@@ -292,51 +292,18 @@ void gpu_tracker::Validator::PreCallRecordCreateDevice(VkPhysicalDevice gpu, con
         delete modified_create_info->pEnabledFeatures;
         modified_create_info->pEnabledFeatures = new VkPhysicalDeviceFeatures(new_features);
     }
-    if (force_buffer_device_address) {
-        // TODO How to handle multi-device
-        if (api_version > VK_API_VERSION_1_1) {
-            auto *features12 = const_cast<VkPhysicalDeviceVulkan12Features *>(
-                vku::FindStructInPNextChain<VkPhysicalDeviceVulkan12Features>(modified_create_info->pNext));
-            if (features12) {
+    // TODO How to handle multi-device
```
spencer-lunarg, line 295: https://twitter.com/zeuxcg/status/1764563738001392011

image

layers/error_message/error_location.cpp
```diff
@@ -60,6 +60,28 @@ std::string Location::Message() const {
 namespace vvl {
 LocationCapture::LocationCapture(const Location& loc) { Capture(loc, 1); }
 
+LocationCapture::LocationCapture(const LocationCapture& other)
```
spencer-lunarg, line 63: was trying to find where this was used and what case we would need to copy/move

spencer-lunarg (approved):

layers/error_message/error_location.cpp
jeremyg-lunarg: good catch. I don't think it is used. I must have thought I needed it but by the end of making the PR for this i didn't.

layers/state_tracker/state_tracker.cpp
```diff
@@ -3713,15 +3727,16 @@ void ValidationStateTracker::PostCallRecordQueuePresentKHR(VkQueue queue, const
 
     auto queue_state = Get<vvl::Queue>(queue);
     Location submit_loc = record_obj.location.dot(vvl::Field::pPresentInfo);
-    vvl::QueueSubmission submission(submit_loc);
+    std::vector<vvl::QueueSubmission> submissions;
```
arno-lunarg, line 3730: Is a vector truly needed here? It seems there is always one and only one element in it

arno-lunarg (approved):
LGTM! One question

layers/state_tracker/state_tracker.cpp
jeremyg-lunarg: the interface to Queue::PreSubmit() is a vector of submission structures even though in this case its only possible ot have one element.

jeremyg-lunarg:
I remember now... The LocationCapture  move constructors are needed by several pieces of code like this:

```
        vvl::QueueSubmission submission(record_obj.location);
        submission.AddFence(Get<vvl::Fence>(fence));
        submissions.emplace_back(std::move(submission));

```
The default move constructor doesn't do the right thing and caused a crash in the last CI run.

copy is needed by core_checks\cc_synchronization.cpp(1886,13): 
```
        cb_state.cmd_execute_commands_functions.emplace_back(
            [this_ptr, loc_capture, active_subpass, sub_desc, render_pass, barrier](
                const vvl::CommandBuffer &secondary_cb, const vvl::CommandBuffer *primary_cb, const vvl::Framebuffer *fb) {
                return this_ptr->ValidateImageBarrierAttachment(loc_capture.Get(), secondary_cb, fb, active_subpass, sub_desc,
                                                                render_pass, barrier, primary_cb);
            });
```
The above isn't related to my change but was only working by accident AFAICT

Issue #7325, opened 2024-01-19
Do not call `DispatchGetAccelerationStructureDeviceAddressKHR` if acceleration structure has no memory bound


arno-lunarg:
in `PostCallRecordCreateAccelerationStructureKHR `we call `DispatchGetAccelerationStructureDeviceAddressKHR`

but there is a case where no memory has been bound to the acceleration structure, need to handle it.

some tests like `NegativeRayTracing.CopyUnboundAccelerationStructure` are doing this already

VU to be in future header update https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6400/diffs

PR #7328, opened 2024-01-19
layers: Fix accel struct device address recording

arno-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7325

If the underlying buffer is not bound to memory, do not try to call vkGetAccelerationStructureDeviceAddressKHR

### DIFF
--- a/layers/state_tracker/state_tracker.cpp
+++ b/layers/state_tracker/state_tracker.cpp
@@ -2328,9 +2328,8 @@ void ValidationStateTracker::PostCallRecordCreateAccelerationStructureNV(VkDevic
 }
 
 std::shared_ptr<vvl::AccelerationStructureKHR> ValidationStateTracker::CreateAccelerationStructureState(
-    VkAccelerationStructureKHR as, const VkAccelerationStructureCreateInfoKHR *ci, std::shared_ptr<vvl::Buffer> &&buf_state,
-    VkDeviceAddress address) {
-    return std::make_shared<vvl::AccelerationStructureKHR>(as, ci, std::move(buf_state), address);
+    VkAccelerationStructureKHR as, const VkAccelerationStructureCreateInfoKHR *ci, std::shared_ptr<vvl::Buffer> &&buf_state) {
+    return std::make_shared<vvl::AccelerationStructureKHR>(as, ci, std::move(buf_state));
 }
 
 void ValidationStateTracker::PostCallRecordCreateAccelerationStructureKHR(VkDevice device,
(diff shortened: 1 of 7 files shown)


layers/state_tracker/state_tracker.cpp
```diff
@@ -2342,7 +2342,11 @@ void ValidationStateTracker::PostCallRecordCreateAccelerationStructureKHR(VkDevi
     auto buffer_state = Get<vvl::Buffer>(pCreateInfo->buffer);
     VkAccelerationStructureDeviceAddressInfoKHR as_address_info = vku::InitStructHelper();
     as_address_info.accelerationStructure = *pAccelerationStructure;
-    const VkDeviceAddress as_address = DispatchGetAccelerationStructureDeviceAddressKHR(device, &as_address_info);
+
+    VkDeviceAddress as_address = 0;
```
spencer-lunarg, line 2346: I was shocked we could just set the `address` here to zero... I tried removing `vvl::AccelerationStructureKHR::address` and seems no one is even using it!

So I would just remove this dispatch call even if the memory is bound as seems no one cares about this address

spencer-lunarg: also remove that member variable from `vvl::AccelerationStructureKHR`

arno-lunarg: I expect GPU AV to need that at some point, so might as well keep it? So that we do not forget again about checking buffer memory state before doing the dispatch

spencer-lunarg: I rather not keep things not used just "in case" as it is not tested at all the value is ever going to be valid.

if GPU-AV needs it, it will need to know how to handle this... by that time, there will be a VU for it, we will have coverage, and it will trip that (when we do VU checking on the GPU-AV dispatch calls)

> So that we do not forget again about checking buffer memory state before doing the dispatch

We have tests on the blacklist that would catch it. I already confirmed one will pass now with this change... honestly this should have been caught sooner and those tests should never have been blacklisted, but we weren't prepared and didn't have the resources to investigate those crashes back then

arno-lunarg: ok!

spencer-lunarg: I am confused why it would be added back later, if I can remove the member variable and it is not used, why is it there, do we have a issue or a known change that will make use of it?

spencer-lunarg: if GPU-AV needs it, it should call it, not CoreCheck

arno-lunarg: Yes it seems only GPU AV will use it, so you convinced me and I will just remove this for now. With this persuasion power, you would be a brilliant politician

spencer-lunarg (approved):

layers/state_tracker/state_tracker.cpp
spencer-lunarg: me right now

image

PR #7329, opened 2024-01-19
layers: Add debug label tracking VU 01912

artem-lunarg:
The validation of labels in the primary command buffer happens on the queue level because debug regions can cross command buffer boundaries. Validation should be done at submit time.

This is more complicated comparing to a similar VU for a secondary command buffer where debug regions are not allowed to cross command buffer boundaries.

Error messages:
> Validation Error: [ VUID-vkCmdEndDebugUtilsLabelEXT-commandBuffer-01912 ] Object 0: handle = 0x20800790df0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x56146426 | vkQueueSubmit(): pSubmits[0].pCommandBuffers[0] (VkCommandBuffer 0x20800790df0[]) contains vkCmdEndDebugUtilsLabelEXT that does not have a matching vkCmdBeginDebugUtilsLabelEXT in this or one of the previously submitted command buffers. There are no previous debug regions before invalid command.

> Validation Error: [ VUID-vkCmdEndDebugUtilsLabelEXT-commandBuffer-01912 ] Object 0: handle = 0x2087d800680, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x56146426 | vkQueueSubmit(): pSubmits[0].pCommandBuffers[0] (VkCommandBuffer 0x2087d800680[]) contains vkCmdEndDebugUtilsLabelEXT that does not have a matching vkCmdBeginDebugUtilsLabelEXT in this or one of the previously submitted command buffers. The previous debug region before the invalid command is 'regionA'.

### DIFF
--- a/layers/core_checks/cc_queue.cpp
+++ b/layers/core_checks/cc_queue.cpp
@@ -1,7 +1,7 @@
-/* Copyright (c) 2015-2023 The Khronos Group Inc.
- * Copyright (c) 2015-2023 Valve Corporation
- * Copyright (c) 2015-2023 LunarG, Inc.
- * Copyright (C) 2015-2023 Google Inc.
+/* Copyright (c) 2015-2024 The Khronos Group Inc.
+ * Copyright (c) 2015-2024 Valve Corporation
+ * Copyright (c) 2015-2024 LunarG, Inc.
+ * Copyright (C) 2015-2024 Google Inc.
  * Modifications Copyright (C) 2020-2022 Advanced Micro Devices, Inc. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -32,6 +32,9 @@ struct CommandBufferSubmitState {
     QFOTransferCBScoreboards<QFOBufferTransferBarrier> qfo_buffer_scoreboards;
     std::vector<VkCommandBuffer> current_cmds;
     GlobalImageLayoutMap overlay_image_layout_map;
+    int cmdbuf_label_stack_depth;
+    std::vector<std::string> cmdbuf_label_names;
+    std::string last_closed_cmdbuf_label;
 
     // The "local" prefix is about tracking state within a *single* queue submission
     // (accross all command buffers of that submission), as opposed to globally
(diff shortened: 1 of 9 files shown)


artem-lunarg:
`NegativeGpuAVDescriptorIndexing.ArrayOOBGraphics` shenanigans

layers/core_checks/cc_queue.cpp
```diff
@@ -83,6 +93,44 @@ struct CommandBufferSubmitState {
...
+private:
+    bool ValidateCmdBufLabelMatching(const Location &loc, const vvl::CommandBuffer &cb_state) {
+        bool skip = false;
+        if (cmdbuf_label_stack_depth < 0) {
+            // If label state is already invalid it means we already reported an error.
+            // Per-queue label state is not updated after it becomes invalid, no need to perform further validation.
+            return skip;
+        }
+        bool mismatch = false;
+        for (const auto &label_cmd : cb_state.GetDebugLabelCommands()) {
```
arno-lunarg, line 106: I feel that this loop logic could be refactored using `UpdateCmdBufLabelStackDepth`?

arno-lunarg: I may have just missed something, still asking:
In addition to an error being thrown if `mismatch` is true, I was expecting an error to be thrown when `cmdbuf_label_stack_depth` is not 0 at the end of the loop. If I follow things correctly, it would mean that a `begin` label was not ended, and this is probably an error?

artem-lunarg: > I feel that this loop logic could be refactored using UpdateCmdBufLabelStackDepth?

`Update` function is for update phase (Record) - it updates state and current function is Validation (it updates only local variables). There is similarity of some control flow, but I don't think it should be the reason to introduce additional code entities (types or helper functions), since the code is simple enough.

artem-lunarg: >  error to be thrown when cmdbuf_label_stack_depth is not 0 at the end of the loop

That's not an error. It means there is an open `BeginLabel` section and it can be close in one of the next command buffers or submits. Debug regions can cross command buffer/submission boundaries.

arno-lunarg: ok, thank you for the explanation!

spencer-lunarg (approved):

arno-lunarg (approved):

PR #7343, opened 2024-01-23
layers: Add VK_EXT_debug_marker

spencer-lunarg:
`VK_EXT_debug_marker` is old and deprecated, but it had 6 simple VUs better to add then keep forever un-done

- VUID-VkDebugMarkerObjectNameInfoEXT-object-01491
- VUID-VkDebugMarkerObjectNameInfoEXT-object-01492
- VUID-VkDebugMarkerObjectNameInfoEXT-objectType-01490
- VUID-VkDebugMarkerObjectTagInfoEXT-object-01494
- VUID-VkDebugMarkerObjectTagInfoEXT-object-01495
- VUID-VkDebugMarkerObjectTagInfoEXT-objectType-01493

### DIFF
--- a/tests/unit/debug_extensions.cpp
+++ b/tests/unit/debug_extensions.cpp
@@ -16,7 +16,6 @@
 
 TEST_F(NegativeDebugExtensions, DebugMarkerName) {
     TEST_DESCRIPTION("Ensure debug marker object names are printed in debug report output");
-
     AddRequiredExtensions(VK_EXT_DEBUG_REPORT_EXTENSION_NAME);
     AddRequiredExtensions(VK_EXT_DEBUG_MARKER_EXTENSION_NAME);
     RETURN_IF_SKIP(Init());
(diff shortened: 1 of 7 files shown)


tests/unit/debug_extensions.cpp
```diff
@@ -25,46 +24,36 @@ TEST_F(NegativeDebugExtensions, DebugMarkerName) {
...
-    VkDeviceMemory memory_1, memory_2;
     std::string memory_name = "memory_name";
 
     VkBufferCreateInfo buffer_create_info = vku::InitStructHelper();
     buffer_create_info.usage = VK_BUFFER_USAGE_UNIFORM_BUFFER_BIT;
     buffer_create_info.size = 1;
-
-    vk::CreateBuffer(device(), &buffer_create_info, nullptr, &buffer);
+    vkt::Buffer buffer;
+    buffer.init_no_mem(*m_device, buffer_create_info);
```
arno-lunarg, line 33: nit, do not bother if this is the only change, but we could do this:
```suggestion
    vkt::Buffer buffer(*m_device, buffer_create_info, vkt::no_mem);
```

tests/unit/debug_extensions.cpp
```diff
@@ -25,46 +24,36 @@ TEST_F(NegativeDebugExtensions, DebugMarkerName) {
...
 
     VkMemoryRequirements memRequirements;
-    vk::GetBufferMemoryRequirements(device(), buffer, &memRequirements);
+    vk::GetBufferMemoryRequirements(device(), buffer.handle(), &memRequirements);
```
arno-lunarg, line 36: you have to specify `.handle()`, the implicit cast does not work?

arno-lunarg (approved):

tests/unit/debug_extensions.cpp
spencer-lunarg: hmm, let me look into this

tests/unit/debug_extensions.cpp
spencer-lunarg: I like this.......like a lot better... let me do a global PR :laughing: (bright side if I do that, it will update all the copyrights in the process!)

PR #7346, opened 2024-01-23
tests: Fix NegativeRT.TransformBufferInvalid

arno-lunarg:
Attempt at fixing `NegativeRayTracing.TransformBufferInvalid`


spencer-lunarg (approved):
Was there invalid Vulkan being called? (aka, are we missing a VU?)

arno-lunarg:
I _think_ that is due to how the driver recycles destroyed memory. With this new flow, I free the transform buffer memory right before issuing a build, no new buffer will be created. so there you be no chance of having the previous transform buffer address to be "reattributed" to another buffer

PR #7347, opened 2024-01-23
layers: Add 03801, cleanup RT stateless validation

arno-lunarg:
Part of https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/3792 

**VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03801**
For each element of pInfos[i].pGeometries or pInfos[i].ppGeometries with a geometryType of VK_GEOMETRY_TYPE_INSTANCES_KHR, the corresponding ppBuildRangeInfos[i][j].primitiveCount must be less than or equal to [VkPhysicalDeviceAccelerationStructurePropertiesKHR](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceAccelerationStructurePropertiesKHR.html)::maxInstanceCount

A test is not really feasible, the logic to add it would imply creating potentially enormous buffers

Also cleaned/refactored stuff in stateless ray tracing validation, especially the divergence we had for validating `pGeometries` and `ppGeometries` pointers from `VkBuildAccelerationStructureGeometryInfoKHR`

### DIFF
--- a/layers/core_checks/cc_ray_tracing.cpp
+++ b/layers/core_checks/cc_ray_tracing.cpp
@@ -495,10 +495,6 @@ bool CoreChecks::ValidateAccelerationBuffers(VkCommandBuffer cmd_buffer, uint32_
                                              const Location &info_loc) const {
     bool skip = false;
 
-    const auto geometry_count = info.geometryCount;
-    const auto *p_geometries = info.pGeometries;
-    const auto *const *const pp_geometries = info.ppGeometries;
-
     auto buffer_check = [this](uint32_t gi, const VkDeviceOrHostAddressConstKHR address, const Location &geom_loc) -> bool {
         const auto buffer_states = GetBuffersByAddress(address.deviceAddress);
         const bool no_valid_buffer_found =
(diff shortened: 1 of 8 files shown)


spencer-lunarg:
> A test is not really feasible, the logic to add it would imply creating potentially enormous buffers

There is, you could use `PFN_VkSetPhysicalDeviceProperties2EXT` in the device layers to emulate a lower value... no one has used it yet... don't think it is needed for now as these limits seem pretty hard to reach

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -520,22 +516,11 @@ bool CoreChecks::ValidateAccelerationBuffers(VkCommandBuffer cmd_buffer, uint32_
...
-    if (p_geometries) {
-        geom_accessor = [p_geometries](uint32_t i) -> const VkAccelerationStructureGeometryKHR & { return p_geometries[i]; };
-    } else if (pp_geometries) {
-        geom_accessor = [pp_geometries](uint32_t i) -> const VkAccelerationStructureGeometryKHR & {
-            // pp_geometries[i] is assumed to be a valid pointer
-            return *pp_geometries[i];
-        };
-    }
-
+    const auto geom_accessor = rt::GetGeometryAccessor(info);
```
spencer-lunarg, line 519: why do we need to have this extra indirect step here to call what is now just a normal util function?

Below can we just have

`rt::GetGeometryAccessor(geom_i)` below 

and a 

```
VkAccelerationStructureGeometryKHR GetGeometryAccessor(const VkAccelerationStructureBuildGeometryInfoKHR &info);
```

layers/stateless/stateless_validation.h
```diff
@@ -955,9 +955,14 @@ class StatelessValidation : public ValidationObject {
                                                      const VkBuffer *pBuffers, const VkDeviceSize *pOffsets,
                                                      const VkDeviceSize *pSizes, const VkDeviceSize *pStrides,
                                                      const ErrorObject &error_obj) const;
-    bool ValidateAccelerationStructureBuildGeometryInfoKHR(const VkAccelerationStructureBuildGeometryInfoKHR *pInfos,
-                                                           uint32_t infoCount, uint64_t total_primitive_count,
-                                                           const VulkanTypedHandle &handle, const Location &loc) const;
+    struct TotalPrimitiveCount {
```
spencer-lunarg, line 958: for only 2 uints, I would rather just have both in the function instead of added a global struct definition

also `total_primitive_count.total_triangles_count` is a little redundant with `total`/`count` in the name

layers/stateless/sl_ray_tracing.cpp
```diff
@@ -888,306 +896,234 @@ bool StatelessValidation::manual_PreCallValidateGetDeviceAccelerationStructureCo
...
-                    default:
-                        break;
-                }
-            }
+
+    if (check_total_primitive_count) {
+        if (total_primitive_count.total_triangles_count > phys_dev_ext_props.acc_structure_props.maxPrimitiveCount) {
+            skip |= LogError("VUID-VkAccelerationStructureBuildGeometryInfoKHR-type-03795", handle, info_loc,
+                             "total number of triangles in all geometries (%" PRIu64
+                             ") is superior to maxPrimitiveCount "
```
spencer-lunarg, line 919:
```suggestion
                             ") is larger than maxPrimitiveCount "
```

is what I think we use for other limits

layers/stateless/sl_ray_tracing.cpp
```diff
@@ -888,306 +896,234 @@ bool StatelessValidation::manual_PreCallValidateGetDeviceAccelerationStructureCo
...
+    if (info.type == VK_ACCELERATION_STRUCTURE_TYPE_BOTTOM_LEVEL_KHR &&
+        info.geometryCount > phys_dev_ext_props.acc_structure_props.maxGeometryCount) {
+        skip |= LogError("VUID-VkAccelerationStructureBuildGeometryInfoKHR-type-03793", handle, info_loc.dot(Field::type),
+                         "is VK_ACCELERATION_STRUCTURE_TYPE_BOTTOM_LEVEL_KHR but geometryCount (%" PRIu32
+                         ") is greater than maxGeometryCount (%" PRIu64 ").",
+                         info.geometryCount, phys_dev_ext_props.acc_structure_props.maxGeometryCount);
+    }
+
+    const auto geom_accessor = rt::GetGeometryAccessor(info);
+    if (geom_accessor) {
```
spencer-lunarg, line 960: I guess not sure why we need this check here, if `geometryCount > 0` and `pGeometries` and `ppGeometries` are both null, that error should have been caught already (would be good to have that as a test to confirm if we don't already)

spencer-lunarg: I see `ComputeTotalPrimitiveCountWithBuildRanges` now... let me re-evaluate

layers/stateless/sl_ray_tracing.cpp
```diff
@@ -1200,9 +1136,15 @@ bool StatelessValidation::manual_PreCallValidateCmdBuildAccelerationStructuresKH
                          "accelerationStructure feature was not enabled.");
     }
 
-    skip |= ValidateAccelerationStructureBuildGeometryInfoKHR(pInfos, infoCount, 0, error_obj.handle, error_obj.location);
+    const TotalPrimitiveCount total_primitive_count =
+        ComputeTotalPrimitiveCountWithBuildRanges(infoCount, pInfos, ppBuildRangeInfos);
```
spencer-lunarg, line 1140: am I wrong, that after this function call we could now check 

```

    if (check_total_primitive_count) {
        if (total_primitive_count.total_triangles_count > phys_dev_ext_props.acc_structure_props.maxPrimitiveCount) {
            skip |= LogError("VUID-VkAccelerationStructureBuildGeometryInfoKHR-type-03795", handle, info_loc,
                             "total number of triangles in all geometries (%" PRIu64
                             ") is superior to maxPrimitiveCount "
                             "(%" PRIu64 ")",
                             total_primitive_count.total_triangles_count, phys_dev_ext_props.acc_structure_props.maxPrimitiveCount);
        }

        if (total_primitive_count.total_aabbs_count > phys_dev_ext_props.acc_structure_props.maxPrimitiveCount) {
            skip |= LogError("VUID-VkAccelerationStructureBuildGeometryInfoKHR-type-03794", handle, info_loc,
                             "total number of AABBs in all geometries (%" PRIu64
                             ") is superior to maxPrimitiveCount "
                             "(%" PRIu64 ")",
                             total_primitive_count.total_aabbs_count, phys_dev_ext_props.acc_structure_props.maxPrimitiveCount);
        }
    }
```

I guess why not just have this function be the one that validates and reports the error, I don't see why we need to carry this information around

layers/core_checks/cc_ray_tracing.cpp
arno-lunarg: I convinced myself at some point that I needed to keep this lambda around, but yes I do no really need it and a simpler function like you suggest is perfectly fine actually

Your persuasion power hits again!

layers/stateless/sl_ray_tracing.cpp
arno-lunarg: 4 different functions need to check those 2 VUs, so I think it is better to put the checks in a function? Or am I missing the point of your question

arno-lunarg: I moved this limit validation function to a dedicated function, `ValidateTotalPrimitivesCount`, used like this in those 4 functions:
```cpp
 uint32_t total_triangles_count = 0;
 uint32_t total_aabbs_count = 0;
 ComputeTotalPrimitiveCountWithMaxPrimitivesCount(infoCount, pInfos, ppMaxPrimitiveCounts, &total_triangles_count, &total_aabbs_count);
 ValidateTotalPrimitivesCount(total_triangles_count, total_aabbs_count, error_obj.handle, error_obj.location);

 for (const auto [info_i, info] : vvl::enumerate(pInfos, infoCount)) {
     const Location info_loc = error_obj.location.dot(Field::pInfos, info_i);

     skip |= ValidateAccelerationStructureBuildGeometryInfoKHR(pInfos[info_i], error_obj.handle, info_loc);
	// ...
}

Comment (spencer-lunarg) [reply layers/stateless/sl_ray_tracing.cpp:1140]: so my question is why can't the logic of `ComputeTotalPrimitiveCountWithMaxPrimitivesCount` just be inside `ValidateTotalPrimitivesCount`?

File: layers/stateless/sl_ray_tracing.cpp (hunk 4 above)
Comment (arno-lunarg) [reply layers/stateless/sl_ray_tracing.cpp:960]: Actually I think there is no VUID to ensure that either pGeometries or ppGeometries is not NULL if geometryCount > 0

Comment (spencer-lunarg) [reply layers/stateless/sl_ray_tracing.cpp:960]: there 100% should be one 
https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6411

**Comment by arno-lunarg** [comment]
This PR now also adds a VU to make sure at least one of pGeometries or ppGeometries is not NULL if geometryCount > 0, and a corresponding test, `NegativeRayTracing, BuildNullGeometries`

File: layers/stateless/sl_ray_tracing.cpp (hunk 5 above)
Comment (arno-lunarg) [reply layers/stateless/sl_ray_tracing.cpp]: There are 2 ways to obtain the total primitive counts, one with build ranges and the other looking at a max primitive count array, so I had to separate this logic

File: layers/utils/ray_tracing_utils.cpp [hunk 6]
```diff
@@ -59,4 +59,13 @@ VkDeviceSize ComputeAccelerationStructureSize(const VkDevice device, const VkAcc
     const VkAccelerationStructureBuildSizesInfoKHR size_info = ComputeBuildSizes(device, build_info, range_infos);
     return size_info.accelerationStructureSize;
 }
+
+const VkAccelerationStructureGeometryKHR &GetGeometry(const VkAccelerationStructureBuildGeometryInfoKHR &info,
+                                                      uint32_t geometry_i) {
+    if (info.pGeometries) {
```
spencer-lunarg, line 65: nit: `return info.pGeometries ? info.pGeometries[geometry_i] : info.ppGeometries[geometry_i];`

spencer-lunarg: could just be in the header file too

layers/stateless/sl_ray_tracing.cpp
```diff
@@ -888,306 +896,249 @@ bool StatelessValidation::manual_PreCallValidateGetDeviceAccelerationStructureCo
...
+    }
+
+    if (info.geometryCount > 0 && !info.pGeometries && !info.ppGeometries) {
+        skip |= LogError("UNASSIGNED-VkAccelerationStructureBuildGeometryInfoKHR-pGeometries", handle,
+                         info_loc.dot(Field::geometryCount), "is (%" PRIu32 ") but both pGeometries and ppGeometries are NULL.",
+                         info.geometryCount);
+        return skip;
+    }
+
+        for (uint32_t geom_i = 0; geom_i < info.geometryCount; ++geom_i) {
```
spencer-lunarg, line 969: clang-format fail

layers/stateless/sl_ray_tracing.cpp
spencer-lunarg: I guess you could just have a 

```c++
ValidateTotalPrimitivesCount(
    uint32_t info_count, 
    const VkAccelerationStructureBuildGeometryInfoKHR *build_geometry_infos,
    const ErrorObject error_obj,
    
    // one is null
    const VkAccelerationStructureBuildRangeInfoKHR *const *build_ranges, 
    const uint32_t *const *max_primitives,
);
```

but happy enough with what you have

spencer-lunarg (approved):
with clang-format fixes

layers/utils/ray_tracing_utils.cpp
arno-lunarg: yes moved it

PR #7351, opened 2024-01-23
scripts: fix international Windows update_deps

lunarpapillo:
After we switched from using subprocess.Popen() to subprocess.run() for command execution in commit bdc0ce89d388155124095b3b4f090574d2d843ea, we became sensitive to non-ASCII characters being emitted to the Windows console, which can happen if a developer tries to use update_deps.py() when their Windows console is set to use a non-ASCII or non-UTF code page.

See:

    Can't update dependencies on Windows due to encoding problems
    https://gitlab.khronos.org/vulkan/Vulkan-ValidationLayers/-/issues/34

One fix here is to specify the Windows-only "oem" encoding by passing:

    encoding='oem'

to subprocess.run(); but this won't work on non-windows platforms.  The easier and more universal fix is to pass:

    errors='replace'

to subprocess.run(), which will cause any encoding errors (that would normally cause exceptions) to instead be replaced with the "unknown character" marker.  This should be sufficient for normal usage.

### DIFF
--- a/scripts/update_deps.py
+++ b/scripts/update_deps.py
@@ -1,8 +1,8 @@
 #!/usr/bin/env python3
 
 # Copyright 2017 The Glslang Authors. All rights reserved.
-# Copyright (c) 2018-2023 Valve Corporation
-# Copyright (c) 2018-2023 LunarG, Inc.
+# Copyright (c) 2018-2024 Valve Corporation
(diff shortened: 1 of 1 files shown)


artem-lunarg (approved):
LGTM

spencer-lunarg (approved):

PR #7354, opened 2024-01-24
layers: Add renderpass vkCmdPipelineBarrier VUs

spencer-lunarg:
adds `VUID-vkCmdPipelineBarrier-None-07892`/`VUID-vkCmdPipelineBarrier-None-07893` that were missing

### DIFF
--- a/layers/core_checks/cc_synchronization.cpp
+++ b/layers/core_checks/cc_synchronization.cpp
@@ -752,6 +752,30 @@ struct RenderPassDepState {
     }
 };
 
+// If inside a renderpass, validate
+bool CoreChecks::ValidateRenderPassPipelineStage(VkRenderPass render_pass, const Location &loc,
+                                                 VkPipelineStageFlags2 src_stage_mask, VkPipelineStageFlags2 dst_stage_mask) const {
+    bool skip = false;
+    const VkPipelineStageFlags2 graphics_stages = syncAllCommandStagesByQueueFlags().at(VK_QUEUE_GRAPHICS_BIT);
+    const VkPipelineStageFlags2 src_diff =
+        sync_utils::ExpandPipelineStages(src_stage_mask, VK_QUEUE_GRAPHICS_BIT) & ~graphics_stages;
+    const VkPipelineStageFlags2 dst_diff =
+        sync_utils::ExpandPipelineStages(dst_stage_mask, VK_QUEUE_GRAPHICS_BIT) & ~graphics_stages;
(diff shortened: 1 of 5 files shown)


layers/core_checks/cc_synchronization.cpp
```diff
@@ -752,6 +752,30 @@ struct RenderPassDepState {
...
+    if (src_diff != 0) {
+        const char *vuid = loc.function == Func::vkCmdPipelineBarrier ? "VUID-vkCmdPipelineBarrier-None-07892"
+                                                                      : "VUID-vkCmdPipelineBarrier2-None-07892";
+        skip = LogError(vuid, render_pass, loc.dot(Field::srcStageMask), "contains non graphics stage %s.",
+                        string_VkPipelineStageFlags2(src_diff).c_str());
+    }
+    if (dst_diff != 0) {
+        const char *vuid = loc.function == Func::vkCmdPipelineBarrier ? "VUID-vkCmdPipelineBarrier-None-07892"
+                                                                      : "VUID-vkCmdPipelineBarrier2-None-07892";
+        skip = LogError(vuid, render_pass, loc.dot(Field::dstStageMask), "contains non graphics stage %s.",
```
artem-lunarg, line 773: `skip |=` ?

layers/core_checks/cc_synchronization.cpp
```diff
@@ -804,6 +829,13 @@ bool CoreChecks::ValidateRenderPassPipelineBarriers(const Location &outer_loc, c
                                                    sub_desc, state.rp_handle, img_barrier);
         }
     }
+
+    if (sub_desc.viewMask > 1 && ((dependency_flags & VK_DEPENDENCY_VIEW_LOCAL_BIT) == 0)) {
```
artem-lunarg, line 833: `viewMask` is a bit mask not a count, as I understand it can be > 1 but with only one bit set, should we count bits here?

layers/core_checks/cc_synchronization.cpp
spencer-lunarg: yes... good catch

layers/core_checks/cc_synchronization.cpp
spencer-lunarg: awww, well this is a bug in `VUID-VkSubpassDependency-srcSubpass-00872` then too (took logic from there).. let me get a positive test for that

artem-lunarg (approved):
LGTM assuming suggested changes are made if necessary

PR #7368, opened 2024-01-25
gpuav: Add debug env variable for shaders

spencer-lunarg:
adds a `VK_LAYER_GPUAV_DEBUG_DUMP_INSTRUMENTED_SHADERS` 

also organizes GPU-AV Debug settings we might want to add to later

### DIFF
--- a/layers/gpu_validation/gpu_setup.cpp
+++ b/layers/gpu_validation/gpu_setup.cpp
@@ -104,7 +104,12 @@ void gpuav::Validator::CreateDevice(const VkDeviceCreateInfo *pCreateInfo) {
     BaseClass::CreateDevice(pCreateInfo);
     Location loc(vvl::Func::vkCreateDevice);
 
-    validate_instrumented_shaders = (GetEnvironment("VK_LAYER_GPUAV_VALIDATE_INSTRUMENTED_SHADERS").size() > 0);
+    debug_validate_instrumented_shaders = !GetEnvironment("VK_LAYER_GPUAV_DEBUG_VALIDATE_INSTRUMENTED_SHADERS").empty();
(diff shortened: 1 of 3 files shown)


arno-lunarg (approved):

layers/gpu_validation/gpu_setup.cpp
```diff
@@ -104,7 +104,12 @@ void gpuav::Validator::CreateDevice(const VkDeviceCreateInfo *pCreateInfo) {
     BaseClass::CreateDevice(pCreateInfo);
     Location loc(vvl::Func::vkCreateDevice);
 
-    validate_instrumented_shaders = (GetEnvironment("VK_LAYER_GPUAV_VALIDATE_INSTRUMENTED_SHADERS").size() > 0);
+    debug_validate_instrumented_shaders = !GetEnvironment("VK_LAYER_GPUAV_DEBUG_VALIDATE_INSTRUMENTED_SHADERS").empty();
```
jeremyg-lunarg, line 107: It would be nice if these worked like other boolean options, i.e  empty, 0, or false == false and anything else == true
Really we should make a Debug group in vkconfig and start putting these in there but that can be a different pr

jeremyg-lunarg (approved):

layers/gpu_validation/gpu_setup.cpp
spencer-lunarg: there is no rush, I rather just get the debug setup working well now... 

1. Good call with making them boolean options
2. Will look into a Debug Group for vkconfig

spencer-lunarg: for interest of having this to debug the 2 failed traces, I am going to get this is now

PR #7373, opened 2024-01-26
layers: Add pipelineStatistics zero VU

spencer-lunarg:
Adds ` VUID-VkQueryPoolCreateInfo-queryType-09534` (added in 1.3.276 spec)

### DIFF
--- a/layers/stateless/sl_descriptor.cpp
+++ b/layers/stateless/sl_descriptor.cpp
@@ -979,20 +979,17 @@ bool StatelessValidation::manual_PreCallValidateCreateQueryPool(VkDevice device,
                                                                 const VkAllocationCallbacks *pAllocator, VkQueryPool *pQueryPool,
                                                                 const ErrorObject &error_obj) const {
     bool skip = false;
-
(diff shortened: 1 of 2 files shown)


layers/stateless/sl_descriptor.cpp
```diff
@@ -979,20 +979,17 @@ bool StatelessValidation::manual_PreCallValidateCreateQueryPool(VkDevice device,
                                                                 const VkAllocationCallbacks *pAllocator, VkQueryPool *pQueryPool,
                                                                 const ErrorObject &error_obj) const {
     bool skip = false;
-
-    // Validation for parameters excluded from the generated validation code due to a 'noautovalidity' tag in vk.xml
```
spencer-lunarg, line 983: this seems old, 

1. We don't have this comment for every `manual_` function, so rather not be inconsistent
2. we won't call `manual_PreCallValidateCreateQueryPool` (or any `manual_`) function if the create parameter is invalid (null)

arno-lunarg (approved):

PR #7379, opened 2024-01-26
layers: Label QueueForwardProgress VUs

spencer-lunarg:
these VUs were discussed in https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6326

- `VUID-vkSetEvent-event-09543` came out in the 276 spec

### DIFF
--- a/layers/core_checks/cc_synchronization.cpp
+++ b/layers/core_checks/cc_synchronization.cpp
@@ -168,7 +168,7 @@ bool SemaphoreSubmitState::ValidateSignalSemaphore(const Location &signal_semaph
                         objlist.add(other_queue);
                     }
                     skip |= core->LogError(
-                        "UNASSIGNED-vkQueueSubmit-QueueForwardProgress", objlist, signal_semaphore_loc,
+                        "VUID-vkQueueSubmit-pCommandBuffers-00065", objlist, signal_semaphore_loc,
                         "(%s) is being signaled by %s, but it was previously signaled by %s and has not since been waited on",
                         core->FormatHandle(semaphore).c_str(), core->FormatHandle(queue).c_str(), initiator.str().c_str());
                 } else {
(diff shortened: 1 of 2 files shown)


layers/core_checks/cc_synchronization.cpp
```diff
@@ -1328,7 +1328,7 @@ bool CoreChecks::PreCallValidateSetEvent(VkDevice device, VkEvent event, const E
     auto event_state = Get<vvl::Event>(event);
     if (event_state) {
         if (event_state->write_in_use) {
-            skip |= LogError("UNASSIGNED-vkSetEvent-QueueForwardProgress", event, error_obj.location.dot(Field::event),
+            skip |= LogError("VUID-vkSetEvent-event-09543", event, error_obj.location.dot(Field::event),
                              "(%s) that is already in use by a command buffer.", FormatHandle(event).c_str());
```
arno-lunarg, line 1332:
```suggestion
                             "(%s) is already in use by a command buffer.", FormatHandle(event).c_str());
```

arno-lunarg: would it be possible to log the handle of the command buffer mentioned in this message?

arno-lunarg (approved):

layers/core_checks/cc_synchronization.cpp
arno-lunarg: Probably out of scope for this PR though

spencer-lunarg: nah, rather just fix it now

spencer-lunarg: ahhh, it would require more tracking, ok, will make a todo/issue for this

arno-lunarg: I think the same things could be said for the other VU, and maybe some others in this function

PR #7393, opened 2024-01-28
layers: Correcting check for vertexPipelineStoresAndAtomics

Honeybunch:
This had me stumped for a little bit and I was *sure* the validation layers couldn't be lying to me but here we are :)


spencer-lunarg (approved):
wow.. was an obvious copy-and-paste error... thanks for catching (and fixing) this!!

For the CI being mad, can you change your commit message to be

> layers: Correcting check for vertexPipelineStoresAndAtomics

and we can get this in!

spencer-lunarg:
thanks again @Honeybunch

PR #7398, opened 2024-01-29
bp: Move extension string to source file

spencer-lunarg:
- Move 2 string maps from header to source file
- Use `_EXTENSION_NAME` enum in header file instead of re-createing these strings
- Reduce the VUID to not need the api call and other info


layers/best_practices/best_practices_error_enums.h
```diff
@@ -26,30 +26,6 @@
     "BestPractices-vkCreateInstance-extension-mismatch";
 [[maybe_unused]] static const char *kVUID_BestPractices_CreateDevice_ExtensionMismatch =
     "BestPractices-vkCreateDevice-extension-mismatch";
-[[maybe_unused]] static const char *kVUID_BestPractices_CreateInstance_DeprecatedExtension =
```
spencer-lunarg, line 29: with `Location` we don't need api call in the error message

also having one for each special use is not necessary as it is easy to tell which is which from the error message

layers/vulkan/generated/best_practices.cpp
```diff
@@ -25,6 +25,166 @@
 #include "chassis.h"
 #include "best_practices/best_practices_validation.h"
 
+DeprecationData GetDeprecatedData(std::string extension_name) {
```
arno-lunarg, line 28: I think we could use `std::string_view` instead of `std::string` here, removing needless dynamic allocations

arno-lunarg (approved):
LGTM, the comment regarding using `std::string_view` could probably be made for other maps

layers/vulkan/generated/best_practices.cpp
artem-lunarg: +1 if possible to do

spencer-lunarg: So I did this change, then realized there are MANY other places we are using raw strings for the extensions

I am currently creating a `vvl::Extension` just like `vvl::Func` 

I want to merge this in now (because it also fixes a lot of other nasty script stuff best practices was doing), but please know soon there will be a replacement on top of this to use the new `vvl::Extension` instead of the current `std::string`

spencer-lunarg: https://github.com/spencer-lunarg/Vulkan-ValidationLayers/tree/spencer-lunarg-string-ext (going to make a PR if CI is happy)

PR #7400, opened 2024-01-29
layers: Remove more extension strings

spencer-lunarg:
Part 2 of https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7399

This one focuses more on `Version` and adds helper to `Requirement` to work as a union when needed

This one only saves a few extra allocations from testing, but the main thing is now I have the setup to add logic to check for flags/enums/structs missing the required extension and providing a good error message without having to bake thousands of generated strings back in


layers/vulkan/generated/spirv_validation_helper.cpp
```diff
@@ -906,94 +906,94 @@ static inline const char* SpvCapabilityRequirements(uint32_t capability) {
...
-    {"SPV_HUAWEI_subpass_shading", "VK_HUAWEI_subpass_shading"},
-    {"SPV_NV_ray_tracing_motion_blur", "VK_NV_ray_tracing_motion_blur"},
-    {"SPV_KHR_maximal_reconvergence", "VK_KHR_shader_maximal_reconvergence"},
-    {"SPV_KHR_subgroup_rotate", "VK_KHR_shader_subgroup_rotate"},
-    {"SPV_KHR_expect_assume", "VK_KHR_shader_expect_assume"},
-    {"SPV_KHR_float_controls2", "VK_KHR_shader_float_controls2"},
-    {"SPV_KHR_quad_control", "VK_KHR_shader_quad_control"},
+static inline std::string SpvExtensionRequirments(std::string_view extension) {
+    static const vvl::unordered_map<std::string_view, vvl::Requirements> table {
+    {"SPV_KHR_variable_pointers", {{vvl::Version::VERSION_1_1}, {vvl::Extension::KHR_variable_pointers}}},
```
jeremyg-lunarg, line 911: I assume you're already planning to de-string the SPV extensions?

jeremyg-lunarg (approved):
LGTM

arno-lunarg (approved):

layers/vulkan/generated/vk_api_version.h
```diff
@@ -26,12 +26,18 @@
...
-#define VK_VERSION_1_1_NAME "VK_VERSION_1_1"
-#define VK_VERSION_1_2_NAME "VK_VERSION_1_2"
-#define VK_VERSION_1_3_NAME "VK_VERSION_1_3"
-
 #define VVL_UNRECOGNIZED_API_VERSION 0xFFFFFFFF
 
+namespace vvl {
+enum class Version {
+    Empty = 0,
+    VERSION_1_0 = 4194304,
```
spencer-lunarg, line 34: was discussed, this should just be `_VK_VERSION_1_0 = (int)VK_API_VERSION_1_0,`

it can't be `VK_VERSION_1_0` because of that is taken in the namespace

but doing `name[3:]` is flimsy

Issue #7349, opened 2024-01-23
Unreported read-after-write conflict between UNIFORM_READ and TRANSFER_WRITE accesses, caused by an unrelated memory barrier


nicebyte:
**Environment:**
 - OS:  Windows 10 Home 22H2
 - GPU and driver version: GeForceRTX 2080 super, driver version 31.0.15.4617
 - SDK or header version if building from repo: SDK 1.3.275.0
 - Options enabled (synchronization, best practices, etc.): everything in Core, Synchronization, QueueSubmit Synchronization Validation, Handle Wrapping, Object Lifetime, Stateless Parameter, Fine Grained Locking, Thread Safety

**Describe the Issue**

The following commands are executed:

1. A transfer writing to buffer B;
2. A compute dispatch, reading from buffer B, using a `StructuredBuffer`;
3. Another compute dispatch, reading from buffer B, using a `ConstantBuffer`.

There is a memory barrier between 1 and 2 using the following stage/access masks:

```
srcStageMask = VK_PIPELINE_STAGE_2_ALL_TRANSFER_BIT
srcAccessMask = VK_ACCESS_2_TRANSFER_WRITE_BIT
dstStageMask = VK_PIPELINE_STAGE_2_COMPUTE_SHADER_BIT
dstAccessMask = VK_ACCESS_2_SHADER_READ_BIT
```

Note that `VK_ACCESS_2_UNIFORM_READ_BIT` is missing from the destination access mask!

**Expected behavior**

I expect validation layers to report a read-after-write hazard, because the above memory barrier does _not_ make the transfer stage's writes available/visible to the UNIFORM_READ accesses. SHADER_READ does not imply UNIFORM_READ. Quote from the spec:

> VK_ACCESS_SHADER_READ_BIT specifies read access to a uniform texel buffer, sampled image,
> storage buffer, physical storage buffer, shader binding table, storage texel buffer, or storage
> image in any shader pipeline stage

**Observed behavior**
VVL do not report any validation errors.

**Additional context**
To further check my own thinking, I have hacked the example above to use vertex attrib / index buffers instead of storage and uniform. In other words:

1. transfer writes to buffer B
2. draw call 1 uses B as vertex attribute buffer
3. draw call 2 uses B as index buffer

A memory barrier similar to the above is issued between (1) and (2) (note the missing INDEX_ATTRIBUTE_READ_BIT):
``` 
[...]

PR #7410, opened 2024-01-30
sync: Fix SHADER_READ expansion

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7349

image

### DIFF
--- a/layers/sync/sync_utils.cpp
+++ b/layers/sync/sync_utils.cpp
@@ -99,8 +99,8 @@ VkPipelineStageFlags2KHR ExpandPipelineStages(VkPipelineStageFlags2KHR stage_mas
 }
 
 static const auto kShaderReadExpandBits =
-    VK_ACCESS_2_UNIFORM_READ_BIT_KHR | VK_ACCESS_2_SHADER_SAMPLED_READ_BIT_KHR | VK_ACCESS_2_SHADER_STORAGE_READ_BIT_KHR;
-static const auto kShaderWriteExpandBits = VK_ACCESS_2_SHADER_STORAGE_WRITE_BIT_KHR;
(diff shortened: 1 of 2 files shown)


artem-lunarg:
Additionally disabled `QSTransitionHazardsPreviousBatch_TimelineSemaphore` which should be fine since we have `QSTransitionHazardsPreviousBatch_BinarySemaphore` which does the same. There is a chance timeline version causes problems on android CI but not 100% sure.

layers/sync/sync_utils.cpp
```diff
@@ -99,8 +99,8 @@ VkPipelineStageFlags2KHR ExpandPipelineStages(VkPipelineStageFlags2KHR stage_mas
 }
 
 static const auto kShaderReadExpandBits =
```
spencer-lunarg, line 101: we have this defined in the XML, for now I am happy with this change, but it is also vulnerable if a new access is added and not updated here... could we move this logic to be generated from the XML

```xml
<syncaccess name="VK_ACCESS_2_SHADER_READ_BIT" alias="VK_ACCESS_SHADER_READ_BIT">
      <syncequivalent access="VK_ACCESS_2_SHADER_SAMPLED_READ_BIT,VK_ACCESS_2_SHADER_STORAGE_READ_BIT,VK_ACCESS_2_SHADER_BINDING_TABLE_READ_BIT_KHR"/>
</syncaccess>
```

spencer-lunarg (approved):

layers/sync/sync_utils.cpp
artem-lunarg: good point

artem-lunarg: will make a separate PR for this

PR #7411, opened 2024-01-31
Cleanup of tests and add |= to skip 

spencer-lunarg:
I was trying to cleanup the Device Group tests, there I found a bug because we went `skip = ValidateSomething();` instead of `skip |=`

I have found sometimes we move around code and having `|=` everywhere is a good counter measure for future code refactoring


layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -2648,9 +2648,9 @@ bool CoreChecks::PreCallValidateCmdSetAttachmentFeedbackLoopEnableEXT(VkCommandB
                                                                       const ErrorObject& error_obj) const {
     bool skip = false;
     auto cb_state = GetRead<vvl::CommandBuffer>(commandBuffer);
-    skip = ValidateExtendedDynamicState(*cb_state, error_obj.location, enabled_features.attachmentFeedbackLoopDynamicState,
```
spencer-lunarg, line 2651: cases like this were not wrong, but the bug I had it seemed someone put an extra check above and didn't witness this was setting it

arno-lunarg (approved):

PR #7414, opened 2024-01-31
layers: Add 03731 03783 04959

arno-lunarg:


layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -1600,6 +1613,22 @@ bool CoreChecks::PreCallValidateCopyAccelerationStructureToMemoryKHR(VkDevice de
     bool skip = false;
     skip |= ValidateDeferredOperation(device, deferredOperation, error_obj.location.dot(Field::deferredOperation),
                                       "VUID-vkCopyAccelerationStructureToMemoryKHR-deferredOperation-03678");
+
```
spencer-lunarg, line 1616: nit: since we have to call into 2 functions with `error_obj.location.dot(Field::pInfo)` better to just make a 

```
const Location info_loc = error_obj.location.dot(Field::pInfo);
```

to save having to build it multiple times

layers/core_checks/cc_device_memory.cpp
```diff
@@ -93,23 +93,39 @@ bool CoreChecks::ValidateMemoryIsBoundToImage(const LogObjectList &objlist, cons
...
+                    LogError(vuid, objlist, buffer_loc, "has been created with a buffer whose bound memory is not host visible.");
             }
         }
     }
     return result;
 }
 
+bool CoreChecks::ValidateBufferMemoryIsNotMultiInstance(const vvl::Buffer &buffer, LogObjectList objlist,
+                                                        const Location &buffer_loc, const char *vuid,
+                                                        const char *error_msg_beginning) const {
```
spencer-lunarg, line 116: Only see one spot calling this, rather not pass in this string

spencer-lunarg: another option is check the `buffer_loc` where it came from and adjust the message that way

spencer-lunarg: honestly, this logic could be inlined, this function doesn't really add much IMO

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -1600,6 +1613,22 @@ bool CoreChecks::PreCallValidateCopyAccelerationStructureToMemoryKHR(VkDevice de
...
+    skip |= ValidateVkCopyAccelerationStructureToMemoryInfoKHR(*pInfo, LogObjectList(device), error_obj.location.dot(Field::pInfo));
+
+    const auto src_accel_struct = Get<vvl::AccelerationStructureKHR>(pInfo->src);
```
spencer-lunarg, line 1619: we are duplicating this `Get` inside of `ValidateVkCopyAccelerationStructureToMemoryInfoKHR`

so probably better to just pass `ValidateVkCopyAccelerationStructureToMemoryInfoKHR` the `vvl::AccelerationStructureKHR` handle (or just inline it for now)

layers/core_checks/cc_device_memory.cpp
arno-lunarg: I am also not super happy about this function, but about 14 VUIDs will need to do this check, that's why I came up with it. I will keep your comments in mind, and will probably evolve this function

layers/core_checks/cc_ray_tracing.cpp
arno-lunarg: yep!

spencer-lunarg (approved):

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -1582,24 +1581,54 @@ bool CoreChecks::PreCallValidateCopyAccelerationStructureKHR(VkDevice device, Vk
...
-            skip |= ValidateHostVisibleMemoryIsBoundToBuffer(*dst_accel_state->buffer_state, info_loc.dot(Field::dst),
-                                                             "VUID-vkCopyAccelerationStructureKHR-buffer-03728");
+            skip |= ValidateHostVisibleMemoryIsBoundToAccelStructBuffer(*dst_accel_state, info_loc.dot(Field::dst),
+                                                                        "VUID-vkCopyAccelerationStructureKHR-buffer-03728");
         }
     }
     return skip;
 }
 
+bool CoreChecks::ValidateVkCopyAccelerationStructureToMemoryInfoKHR(const vvl::AccelerationStructureKHR &src_accel_struct,
```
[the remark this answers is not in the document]
arno-lunarg: this is because I am an
image

arno-lunarg:
Oops I messed up my branch management, added 2 new commits that add validation for a bunch of null checks on RT host builds

PR #7416, opened 2024-01-31
layers: Support loading shader modules from pipeline caches

aqnuep:
This change adds infrastructure that enables loading shader module data from pipeline caches.

Currently, this is intended to be used by the Vulkan SC validation layers, but it is general enough that it could be leveraged in the Vulkan validation layers too, if the need arises in the future.

### DIFF
--- a/layers/containers/custom_containers.h
+++ b/layers/containers/custom_containers.h
@@ -875,8 +875,8 @@ class enumeration {
     iterator end() { return data_ + count_; }
     const_iterator end() const { return data_ + count_; }
 
-    T &operator[](int i) { return data_[i]; }
-    const T &operator[](int i) const { return data_[i]; }
+    T &operator[](size_t i) { return data_[i]; }
+    const T &operator[](size_t i) const { return data_[i]; }
 
     T &front() { return *data_; }
     const T &front() const { return *data_; }
--- a/layers/state_tracker/pipeline_state.h
+++ b/layers/state_tracker/pipeline_state.h
@@ -56,12 +56,23 @@ struct StageCreateInfo {
 namespace vvl {
 class PipelineCache : public StateObject {
   public:
+    const safe_VkPipelineCacheCreateInfo create_info;
+
     PipelineCache(VkPipelineCache pipeline_cache, const VkPipelineCacheCreateInfo *pCreateInfo)
         : StateObject(pipeline_cache, kVulkanObjectTypePipelineCache), create_info(pCreateInfo) {}
 
     VkPipelineCache VkHandle() const { return handle_.Cast<VkPipelineCache>(); }
 
-    const safe_VkPipelineCacheCreateInfo create_info;
+    virtual std::shared_ptr<const vvl::ShaderModule> GetStageModule(const vvl::Pipeline &pipe_state, size_t stage_index) const {
+        // This interface enables derived versions of the pipeline cache state object to return
+        // the shader module information from pipeline cache data, if available.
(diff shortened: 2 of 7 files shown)


layers/state_tracker/pipeline_state.h
```diff
@@ -56,12 +56,23 @@ struct StageCreateInfo {
...
     PipelineCache(VkPipelineCache pipeline_cache, const VkPipelineCacheCreateInfo *pCreateInfo)
         : StateObject(pipeline_cache, kVulkanObjectTypePipelineCache), create_info(pCreateInfo) {}
 
     VkPipelineCache VkHandle() const { return handle_.Cast<VkPipelineCache>(); }
 
-    const safe_VkPipelineCacheCreateInfo create_info;
+    virtual std::shared_ptr<const vvl::ShaderModule> GetStageModule(const vvl::Pipeline &pipe_state, uint32_t stage_index) const {
+        // This interface enables derived versions of the pipeline cache state object to return
+        // the shader module information from pipeline cache data, if available.
+        // This is currently used by Vulkan SC to retrieve SPIR-V module debug information when
```
spencer-lunarg, line 69: the hooking up looks good, but still not fully understanding is how one can get the `vvl::ShaderModule` state from a pipeline cache, it is not being written anywhere

spencer-lunarg: you say " may also be used by vendor-specific validation layers" so assuming there requires special knowledge of whats inside the pipeline internals?

spencer-lunarg (approved):

layers/state_tracker/pipeline_state.h
aqnuep: In Vulkan SC, pipeline caches have a semi-transparent internal layout, which itself can hold SPIR-V binaries as debug information. This is parsed downstream and will be used to do SPIR-V dependent validation.

In a similar fashion, if in Vulkan somebody would like to make their pipeline cache internal layout similarly transparent with SPIR-V debug data, then this could also be used upstream.

layers/containers/custom_containers.h
```diff
@@ -875,8 +875,8 @@ class enumeration {
     iterator end() { return data_ + count_; }
     const_iterator end() const { return data_ + count_; }
 
-    T &operator[](int i) { return data_[i]; }
-    const T &operator[](int i) const { return data_[i]; }
+    T &operator[](size_t i) { return data_[i]; }
+    const T &operator[](size_t i) const { return data_[i]; }
```
aqnuep, line 879: @spencer-lunarg, looks like the build issues are related to `vvl::span` not using index types as would be with the STL containers. We'll see whether this affects any other legacy code.

spencer-lunarg: seems to good now, thanks

aqnuep:
@spencer-lunarg, let me know if you need any action from my end (e.g. rebasing) before this can get merged.

It seems there's something wrong with the CLA job, I'm not sure whether that's an issue, as this is not my first contribution :)

spencer-lunarg:
I noticed the CLA issue earlier as well, it is still waiting.... looking online, I can't find anything, I can't even find anything in the settings for this as well, this is a `KhronosGroup` org level thing, not a VVL thing (to my knowledge).... can I ask you to rebase and push up again, I have no idea how to force this in (the I can't bypass the requirement)

aqnuep:
Rebased and pushed. Hopefully the CLA issue will be resolved as well.

lunarpapillo:
The **LunarG CI** failure above seems unrelated.  I'm restarting.

PR #7423, opened 2024-02-01
sync: Fix label issues and simplify implementation

artem-lunarg:
The previous implementation did not handle some cases when a submission
included multiple batches or a batch had multiple command buffers.
Updated implementation is also simpler (removes few complex aglorithmic
sequences).


spencer-lunarg (approved):

layers/state_tracker/cmd_buffer_state.h
```diff
@@ -696,6 +696,9 @@ class CommandBuffer : public RefcountedStateObject {
     };
     const std::vector<LabelCommand> &GetLabelCommands() const { return label_commands_; }
 
+    static void ReplayLabelCommands(const vvl::span<const LabelCommand> &label_commands, std::vector<std::string> &label_stack);
```
arno-lunarg, line 699: Maybe add a comment that briefly describes those functions intent?

arno-lunarg (approved):

layers/state_tracker/cmd_buffer_state.h
artem-lunarg: Added comments

arno-lunarg: Thank you!!

PR #7426, opened 2024-02-01
tests: Remove init from instance extensions test

ziga-lunarg:
Initialization is not required in this test and in the profiles repository had a similar test that only failed if an instance was not created before calling `vkEnumerateInstanceExtensionProperties`


spencer-lunarg (approved):

spencer-lunarg:
took a bit of CI churn, but finally removed those 2 lines!

PR #7429, opened 2024-02-01
scripts: Generate expansion for shader read/write

artem-lunarg:
Follow-up to https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7410

### DIFF
--- a/scripts/generators/sync_validation_generator.py
+++ b/scripts/generators/sync_validation_generator.py
@@ -156,6 +156,15 @@ class SyncValidationOutputGenerator(BaseGenerator):
             #include "containers/custom_containers.h"
             ''')
         out.append('// clang-format off\n')
+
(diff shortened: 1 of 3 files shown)


scripts/generators/sync_validation_generator.py
```diff
@@ -156,6 +156,15 @@ def generateHeader(self):
             #include "containers/custom_containers.h"
             ''')
         out.append('// clang-format off\n')
+
+        shader_read_access = next((a for a in self.vk.syncAccess if a.flag.name == 'VK_ACCESS_2_SHADER_READ_BIT'), None)
```
spencer-lunarg, line 160: the fact this was only 3 lines of python to get this added for some reason just made me really happy

spencer-lunarg (approved):

scripts/generators/sync_validation_generator.py
artem-lunarg: and that's for python beginner

PR #7437, opened 2024-02-04
layers: Fix a crash when image resources are deleted

corporateshark:
Fixed a crash in Validation layers when image/imageview are deleted in a descriptor indexing scenario.

**Do:**
* Install Vulkan SDK 1.3.275.0
* Run https://github.com/corporateshark/lightweightvk/blob/main/samples/Tiny_Mesh.cpp
* Press `T`

**Observe:**

A crash in Validation layers.

image


spencer-lunarg (approved):
Thanks for the fix, I will try and get a test case later to reproduce this to prevent regressions

PR #7439, opened 2024-02-04
layers: Fix how checking for required pipeline state

spencer-lunarg:
As per discussion in https://gitlab.khronos.org/vulkan/vulkan/-/issues/3764#note_451564 the understanding of GPL was wrong

### DIFF
--- a/layers/state_tracker/pipeline_state.h
+++ b/layers/state_tracker/pipeline_state.h
@@ -268,6 +268,8 @@ class Pipeline : public StateObject {
 
     // Used to know if the pipeline substate is being created (as opposed to being linked)
     // Important as some pipeline checks need pipeline state that won't be there if the substate is from linking
+    // Many VUs say "the pipeline require" which means "not being linked in as a library"
+    // If the VUs says "created with" then you should NOT use this function
     bool OwnsSubState(const std::shared_ptr<PipelineSubState> sub_state) const { return sub_state && (&sub_state->parent == this); }
 
     const std::shared_ptr<const vvl::RenderPass> RenderPassState() const {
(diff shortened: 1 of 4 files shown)


layers/state_tracker/pipeline_state.h
```diff
@@ -268,6 +268,8 @@ class Pipeline : public StateObject {
 
     // Used to know if the pipeline substate is being created (as opposed to being linked)
     // Important as some pipeline checks need pipeline state that won't be there if the substate is from linking
+    // Many VUs say "the pipeline require" which means "not being linked in as a library"
+    // If the VUs says "created with" then you should NOT use this function
```
artem-lunarg, line 272: From conversation I understood that "require" means it should be specified as part of pipeline creation (not linking).
Does this function check if the state was linked?
Why not not to use this function if "created with"?

spencer-lunarg: So there as background there are 3 ways to add each state value

1. tradition pipelines (you get all 4 states automatically)
2. with a GPL create flag (you are building just that one)
3. linking in

the goal is to match `require` to case `1` and `2`

> Why not not to use this function for "created with" then?

I am not sure I understand what you are asking, sorry

artem-lunarg: Thanks for the details.
The second question was about this comment:  // If the VUs says "created with" then you should NOT use this function
The above categorization clarifies things but I still can't see clearly the requirement of this comment.

spencer-lunarg: ... honestly I have a desire to re-do a lot of this, I wanted to unify the naming in general... things are not "stable" IMO

PR #7443, opened 2024-02-05
tests: Add GPU-AV Spir-V loop test

spencer-lunarg:
I found a nasty loop (still trying to find out if it [valid or not](https://gitlab.khronos.org/spirv/SPIR-V/-/issues/778)) but wanted to create a PR to make sure we are not crashing from it (since both the old and new shader instrumentation is not handling it with valid SPIR-V, but it "works"


spencer-lunarg:
moving to a separate PR soon

Issue #7434, opened 2024-02-02
gpuav: crash from null VMA allocation in performance traces


jeremyg-lunarg:
**Environment:**
 - OS: Windows
 - GPU and driver version: RTX 2080
 - SDK or header version if building from repo:  anything containing https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7183
 - Options enabled (synchronization, best practices, etc.): GPUAV

**Describe the Issue**
~~NOTE: You will need https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7405 to get far enough to hit this bug.~~ merged

The `thesurge2-1-10000` crashes with the following stack trace:
```
common_assert_to_message_box<wchar_t>(const wchar_t * const, const wchar_t * const, const unsigned int, void * const)
common_assert<wchar_t>(const wchar_t * const, const wchar_t * const, const unsigned int, void * const)
_wassert(const wchar_t *, const wchar_t *, unsigned int)
vmaMapMemory(VmaAllocator_T *, VmaAllocation_T *, void * *)
gpuav::CommandResources::LogErrorIfAny(gpuav::Validator &, VkQueue_T *, VkCommandBuffer_T *, const unsigned int)
gpuav::CommandBuffer::Process(VkQueue_T *, const Location &)
gpu_tracker::Validator::ProcessCommandBuffer(VkQueue_T *, VkCommandBuffer_T *, const Location &)
gpu_tracker::Validator::PostCallRecordQueueSubmit(VkQueue_T *, unsigned int, const VkSubmitInfo *, VkFence_T *, const RecordObject &)
gpuav::Validator::PostCallRecordQueueSubmit(VkQueue_T *, unsigned int, const VkSubmitInfo *, VkFence_T *, const RecordObject &)
vulkan_layer_chassis::QueueSubmit(VkQueue_T *, unsigned int, const VkSubmitInfo *, VkFence_T *)

```
in LogErrorIfAny output_memblock.allocation is nullptr, which causes VMA to assert.

Reverting #7183 makes this crash go away.

spencer-lunarg:
can reproduce with a simple test

```c++
TEST_F(PositiveGpuAV, CopyBufferToImageTwoSubmit) {
    TEST_DESCRIPTION("Make sure resources are managed correctly afer a CopyBufferToImage call.");
    RETURN_IF_SKIP(InitGpuAvFramework());
    RETURN_IF_SKIP(InitState());

[...]

PR #7183, opened 2023-12-22
gpu: Add CmdCopyBufferToImage

arno-lunarg:
Add 07931:
> If [VK_EXT_depth_range_unrestricted](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_EXT_depth_range_unrestricted.html) is not enabled, for each element of pRegions whose imageSubresource contains a depth aspect, the data in srcBuffer must be in the range [0,1]

Simplified a bit LogValidationMessage with the addition of
LogCustomValidationMessage.
While at it, removed a redundant gpu memory mapping.

Redo management of shared command validation resources:
Added a shared validation resources map to store them,
shared_validation_resources_map.
this map is cleared at PreCallRecordDestroyDevice time.
=> No need to connect new shared resources types by hand
to the Validation class.
Moved shared resources types inside the relevant classes.

PR #7444, opened 2024-02-05
gpu: Fix null CommandResources mangaement

arno-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7434

Do not store empty `CommandResources`, they do not need to be processed anyway

Before https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7183, null `CommandResources` would only be created when an error was encountered. Now, thanks to this line, a null `CommandResources` can be created when there is no need to perform validation:

```cpp
   // Only need to perform validation for depth image having a depth format that is not unsigned normalized.
    // For unsigned normalized formats, depth is by definition in range [0, 1]
    if (!IsValueIn(image_state->safe_create_info.format, {VK_FORMAT_D32_SFLOAT, VK_FORMAT_D32_SFLOAT_S8_UINT})) {
        return nullptr;
    }
```

Null `CommandResources` are now just not stored, no need to


spencer-lunarg (approved):
This fixes my local capture that was crashing

Issue #7433, opened 2024-02-02
Expecting validation error for using `VkPhysicalDeviceVulkan11Features` on Vulkan 1.1 device


MarijnS95:
The `VkPhysicalDeviceVulkan11Features` struct was only introduced with `VK_VERSION_1_2`. Testing on an Android phone with instance version `0.1.3.0`, a Vulkan device of `1.1.x`, with [the latest `1.3.268` prebuilt layers](https://github.com/KhronosGroup/Vulkan-ValidationLayers/releases/tag/vulkan-sdk-1.3.268.0) no validation errors are shown when passing `VkPhysicalDeviceVulkan11Features` to this device. When manually downgrading `VkApplicationInfo::apiVersion` which [is said to affect what the VVL checks against](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkApplicationInfo.html#_description) to `1.1`, the error is properly shown.

I'd assume that the Vulkan driver/device might or should not accept `VkPhysicalDeviceVulkan11Features`. Is this thought correct, or is the ICD converting these to the individual feature structs for older devices/drivers? If the former, should a VUID be issues to catch this?

spencer-lunarg:
this is a slightly longer-then-I-wish known issue (https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5012) I am currently [very close](https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7420) to having the code gen done for the enums/flags and will then go right to checking for the struct next!

MarijnS95:
@spencer-lunarg good stuff! I was indeed looking under the totally wrong search terms when trying to find a duplicate of this.

Do you want to close this as a duplicate?

Do you want me to test that the proposed case above indeed results in a validation layer error with #7420 merged in locally?

MarijnS95:
And should the linked PR also have `closes #5012`?

spencer-lunarg:
happy to keep this open and I plan to use this issue to write a test case for your issue

... hopefully by end of this week this will be done and I can go through and closes **all** the related issues lol (they are probably all slightly overlapping each other now)

MarijnS95:
[...]

Issue #5012, opened 2022-12-28
Missing validation for structs in pNext chains whose device extensions aren't supported by the device


charles-lunarg:
**Describe the situation in which you encountered the missing validation**
The current validation for pNext chains only verifies that they do not contain structs which never go in the chain. For example, a properties struct in a features chain (ex: VUID-VkDeviceCreateInfo-pNext-pNext). But what they don't check is whether the structs corresponding extension is supported by the implementation.

Real world use-case: If an application adds `VkPhysicalDevicePortabilitySubsetFeaturesKHR` to the pNext chain of vkGetPhysicalDeviceFeatures2 on a device that doesn't support the portability subset extension, the implementation is required to skip the struct. However, the application uses the feature flags in that struct to determine what is or isn't supported, and since it was skipped, every field will be false. (assuming the user default initialized the struct).

**Valid Usage IDs requested**
This doesn't have specific VUID's, rather it is generic spec language. New VUID's may be required, though best practices checks may be better fit.

**Additional Context**
[Vulkan Spec 3.7.2](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap3.html#fundamentals-implicit-validity)
> Any component of the implementation (the loader, any enabled layers, and drivers) must skip over, without processing (other than reading the sType and pNext members) any extending structures in the chain not defined by core versions or extensions supported by that component.

This bit of the spec states that unsupported structs should simply be skipped by layers/drivers. The language is imprecise (many discussion have been had about what 'supported' means and more specific VUID's like VUID-VkDeviceCreateInfo-pNext-pNext actually go against that rule. I think changing this bit of the spec or adding conditions to the 'skip over it' would be beneficial. 

I should say it makes sense to have this bit in the spec, as when a layer or driver gets a struct in a pnext chain that it doesn't know about, the component should ignore it. This happens quite a bit with older components not updated to the latest version.

billhollings:
> I think changing this bit of the spec or adding conditions to the 'skip over it' would be beneficial.

Agreed. I think in the final spec for the `VK_KHR_portability_subset` extension, we could modify the spec text to add something like:

> An exception to this is the `VkPhysicalDevicePortabilitySubsetFeaturesKHR` structure, which must not be submitted in the `pNext` chain of `VkPhysicalDeviceFeatures2` if the `VK_KHR_portability_subset` extension is not supported by the physical device.

[...]

PR #7448, opened 2024-02-05
layers: Add Struct extension check

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7433
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5012

Adds logic to check if a struct is being used when you didn't enable the extension

### DIFF
--- a/tests/unit/others.cpp
+++ b/tests/unit/others.cpp
@@ -63,21 +63,11 @@ TEST_F(VkLayerTest, UnsupportedPnextApiVersion) {
     SetTargetApiVersion(VK_API_VERSION_1_1);
 
     RETURN_IF_SKIP(Init());
-    VkPhysicalDeviceProperties2 phys_dev_props_2 = vku::InitStructHelper();
-    VkPhysicalDeviceVulkan12Properties bad_version_1_1_struct = vku::InitStructHelper();
-    phys_dev_props_2.pNext = &bad_version_1_1_struct;
-
-    // VkPhysDevVulkan12Props was introduced in 1.2, so try adding it to a 1.1 pNext chain
-    if (DeviceValidationVersion() >= VK_API_VERSION_1_1) {
-        m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkPhysicalDeviceProperties2-pNext-pNext");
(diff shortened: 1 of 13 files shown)


MarijnS95:
Tested, this unfortunately doesn't seem to complain when passing `VkPhysicalDeviceVulkan11Features` to a Vulkan 1.1 device, as outlined in #7433.

tests/unit/others.cpp
```diff
@@ -2004,4 +2006,55 @@ TEST_F(VkLayerTest, DescriptorBufferNoExtension) {
...
+    VkBufferViewCreateInfo buffer_view_ci = vku::InitStructHelper(&buffer_usage_flags);
+    buffer_view_ci.format = VK_FORMAT_R8G8B8A8_UNORM;
+    buffer_view_ci.range = VK_WHOLE_SIZE;
+    buffer_view_ci.buffer = buffer.handle();
+    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkBufferViewCreateInfo-pNext-pNext");
+    vkt::BufferView view(*m_device, buffer_view_ci);
+    m_errorMonitor->VerifyFound();
+}
+
+TEST_F(VkLayerTest, Features11WithoutVulkan12) {
```
spencer-lunarg, line 2034: @MarijnS95  I have a test here, I am not sure what your case is doing different

MarijnS95: @spencer-lunarg `SetTargetApiVersion()` seems to be a global (that on my setup would match what I pass to `VkApplicationInfo::apiVersion` or would be the instance version?), but as _asked_ in #7433 I **assume** this should take the _device driver_ version (`VkPhysicalDeviceProperties::version`) into account, when passing device-specific data?

spencer-lunarg: hmm, let me grab some logs and confirm the test is doing what I think it should be doing (instance version 0.1.3.0, a Vulkan device of 1.1.x)

spencer-lunarg: ok, @MarijnS95 I was able to rework the test and reproduce it and have it fixed now!

artem-lunarg (approved):
LGTM

PR #7464, opened 2024-02-07
layers: Reduce size of oversized VUs

spencer-lunarg:
```
// length of VUs
VUID-VkColorBlendEquationEXT-colorBlendOp-07361 - 1242
VUID-VkDeviceCreateInfo-pNext-pNext - 9068
VUID-VkPhysicalDeviceProperties2-pNext-pNext - 4224
```

These over size error message are quite obnoxious and doing more harm then good


artem-lunarg (approved):
LGTM. Extreme readability.

Issue #6844, opened 2023-10-26
VUID-vkCmdDrawMultiEXT-rasterizationSamples-07471 broken


zmike:
```
VUID-vkCmdDrawMultiEXT-rasterizationSamples-07471(ERROR / SPEC): msgNum: 1182038117 - Validation Error: [ VUID-vkCmdDrawMultiEXT-rasterizationSamples-07471 ] Object 0: handle = 0x7a45320, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0x74ea280000000054, type = VK_OBJECT_TYPE_PIPELINE; | MessageID = 0x46747865 | vkCmdDrawMultiEXT():  VkPhysicalDeviceFeatures::variableMultisampleRate is VK_FALSE and the rasterizationSamples set with vkCmdSetRasterizationSamplesEXT() were VK_SAMPLE_COUNT_1_BITbut a previous draw used rasterization samples 4.. The Vulkan spec states: If the bound graphics pipeline state was created with the VK_DYNAMIC_STATE_RASTERIZATION_SAMPLES_EXT state enabled, and the current subpass does not use any color and/or depth/stencil attachments, then the rasterizationSamples in the last call to vkCmdSetRasterizationSamplesEXT must follow the rules for a zero-attachment subpass (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdDrawMultiEXT-rasterizationSamples-07471)
```

The error text for this needs cleaning up, but also the error triggered is spurious/misleading/???:

`ZINK_DEBUG=vvl LIBGL_ALWAYS_SOFTWARE=1 MESA_LOADER_DRIVER_OVERRIDE=zink ./glcts -n dEQP-GLES31.functional.shaders.sample_variables.sample_pos.correctness.multisample_texture_4` to run a cts case on zink on lavapipe triggers it, but attachments are used in the subpass.

Also I don't actually see this VU anywhere in the spec? The closest I could find is `VUID-vkCmdBindPipeline-pipeline-00781`

ziga-lunarg:
I can see the VU here: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCmdDrawMultiEXT.html

Can you let me know which subpass this is and which attachment does it have? And can you verify the attachment is not VK_ATTACHMENT_UNUSED?

zmike:
[...]

PR #7468, opened 2024-02-07
layers: Fix no-attachment subpass

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6844

- We were not checking for dynamic rendering when checking for "no-attachment subpasses" so made a common util function
- `VUID-VkGraphicsPipelineCreateInfo-subpass-00758` was mislabeled as `VUID-vkCmdBindPipeline-pipeline-00781`
- Added `VUID-VkGraphicsPipelineCreateInfo-subpass-00758`

### DIFF
--- a/layers/state_tracker/render_pass_state.cpp
+++ b/layers/state_tracker/render_pass_state.cpp
@@ -1,7 +1,7 @@
-/* Copyright (c) 2015-2023 The Khronos Group Inc.
- * Copyright (c) 2015-2023 Valve Corporation
- * Copyright (c) 2015-2023 LunarG, Inc.
- * Copyright (C) 2015-2023 Google Inc.
+/* Copyright (c) 2015-2024 The Khronos Group Inc.
+ * Copyright (c) 2015-2024 Valve Corporation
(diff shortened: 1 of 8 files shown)


layers/state_tracker/render_pass_state.cpp
```diff
@@ -345,6 +345,12 @@ bool RenderPass::UsesDepthStencilAttachment(uint32_t subpass_num) const {
     return result;
 }
 
+// vkspec.html#renderpass-noattachments
+bool RenderPass::UsesNoAttachment(uint32_t subpass) const {
```
artem-lunarg, line 349: Just curious, can it be called in dynamic rendering context?

artem-lunarg (approved):

layers/state_tracker/render_pass_state.cpp
spencer-lunarg: yes, the bug was because we were mixing up "dynamic rendering" with "no attachment subpass"

arno-lunarg (approved):

PR #7473, opened 2024-02-07
build: Replace robin-hood-hashing with unordered_dense

jeremyg-lunarg:
robin-hood-hashing is no longer developed. unordered_dense is its replacement, by the same author.


jeremyg-lunarg:
This is draft because:

1.  it needs to be tested beyond CI (traces)
2. We need to talk about the CMake variable name changes and if there should be a backwards compat hack  (and possibly a generic name like USE_FAST_UNORDERED_MAP)

Issue #7441, opened 2024-02-04
SYNC-HAZARD-READ-AFTER-WRITE in vkCmdEndRendering() when resolving depth buffer


martin-ejdestig:
**Environment:**
 - OS: Windows and Linux
 - GPU and driver version: Nvidia GeForce GTK 970 and Radeon RX 6700, latest drivers as of this writing (do not think it matters)
 - SDK or header version if building from repo: 1.3.275
 - Options enabled (synchronization, best practices, etc.): `VK_VALIDATION_FEATURE_ENABLE_SYNCHRONIZATION_VALIDATION_EXT`

**Describe the Issue**

Started seeing `SYNC-HAZARD-READ-AFTER-WRITE` from validation layer when calling `vkCmdEndRendering()` and resolving depth buffer after upgrading from 1.3.268 to 1.3.275.
```
Vulkan (validation): Validation Error: [ SYNC-HAZARD-READ-AFTER-WRITE ] Object 0: handle = 0x56b9dee73c50, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0x4f8a8b0000000046, name = Main Render FB Depth Image View, type = VK_OBJECT_TYPE_IMAGE_VIEW; | MessageID = 0xe4d96472 | vkCmdEndRendering(): pDepthAttachment->imageView (VkImageView 0x4f8a8b0000000046[Main Render FB Depth Image View]), during resolve with resolveMode VK_RESOLVE_MODE_MIN_BIT. Access info (usage: SYNC_COLOR_ATTACHMENT_OUTPUT_COLOR_ATTACHMENT_READ, prior_usage: SYNC_EARLY_FRAGMENT_TESTS_DEPTH_STENCIL_ATTACHMENT_WRITE, write_barriers: 0, command: vkCmdBeginRendering, seq_no: 432, reset_no: 33, debug region: Main Render).
```
There is no warning for resolving the color buffer. There is also no warning if `VkRenderingAttachmentInfo::resolveMode` is set to `VK_RESOLVE_MODE_NONE` for the depth attachment or if MSAA is disabled entirely. 

[...]

PR #7476, opened 2024-02-07
sync: Fix ordered accesses for depth-stencil resolve

artem-lunarg:
A unique thing about resolve operations is that they happen on the `COLOR_OUTPUT` stage both for color and depth attachments and use color-specific accesses. This requirement forced ordered accesses to be also color-specific. But for depth-stencil attachment, the participation of depth-stencil accesses is still possible. In the reported issue, it is a `LoadOp::Clear` write which did not pass color ordering barrier. The fix includes both Color and Depth accesses in `OrderingBarrier` for depth-stencil attachment during resolve operation.

Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7441

@jzulauf-lunarg DEFCON 4

### DIFF
--- a/layers/sync/sync_commandbuffer.cpp
+++ b/layers/sync/sync_commandbuffer.cpp
@@ -216,6 +216,8 @@ bool CommandBufferAccessContext::ValidateEndRendering(const ErrorObject &error_o
         for (uint32_t i = 0; i < attachment_count && !skip; i++) {
             const auto &attachment = info.attachments[i];
             if (attachment.resolve_gen) {
+                const bool is_color = attachment.type == syncval_state::AttachmentType::kColor;
+                const SyncOrdering kResolveOrder = is_color ? kColorResolveOrder : kDepthStencilResolveOrder;
                 // The logic about whether to resolve is embedded in the Attachment constructor
                 assert(attachment.view);
                 HazardResult hazard = access_context->DetectHazard(attachment.view_gen, kResolveRead, kResolveOrder);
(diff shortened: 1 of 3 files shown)


layers/sync/sync_access_state.cpp
```diff
@@ -18,9 +18,17 @@
...
+    {// kNonAttachment
+     {VK_PIPELINE_STAGE_2_NONE_KHR, SyncStageAccessFlags()},
+     // kColorAttachment
      {kColorAttachmentExecScope, kColorAttachmentAccessScope},
+     // kDepthStencilAttachment
      {kDepthStencilAttachmentExecScope, kDepthStencilAttachmentAccessScope},
+     // kDepthStencilResolveAttachment: depth-stencil resolve runs (as all resolve operations) on
+     // COLOR_OUTPUT_STAGE and uses COLOR_READ/WRITE access.
+     // The prior ordered access is still a regular depth-stencil access.
+     {kColorAttachmentExecScope, kColorAttachmentAccessScope | kDepthStencilAttachmentAccessScope},
```
jzulauf-lunarg, line 30: why doesn't this include DepthExecScope? -- and wouldn't that be the same as kRaster

jzulauf-lunarg (commented):
Not sure why kRaster isn't being used instead of creating a new OrderingBarrier.

layers/sync/sync_access_state.cpp
artem-lunarg: Because Depth Resolve runs only on COLOR_OUTPUT stage, the only reason depth added to accesses because we prior access is depth-specific (from loadop for example) and current access is color_read to read multisampled depth

artem-lunarg: kRaster works and results in a more compact change. My initial motivation for new ordering mode is that it might enough to have only Color execution stage, in this regard raster has a bit wider scope. But if no issues I like the suggested approach more because it's simpler.

jzulauf-lunarg (approved):
LGTM

Issue #7361, opened 2024-01-25
CPU Ray Tracing VUs


spencer-lunarg:
See https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5858 for VUs needing GPU-AV

`not allocated with multiple instances`

- [x] VUID-vkBuildAccelerationStructuresKHR-pInfos-03775
- [x] VUID-vkBuildAccelerationStructuresKHR-pInfos-03776
- [x] VUID-vkCopyAccelerationStructureKHR-buffer-03780
- [x] VUID-vkCopyAccelerationStructureKHR-buffer-03781
- [x] VUID-vkCopyMemoryToAccelerationStructureKHR-buffer-03782
- [x] VUID-vkCopyAccelerationStructureToMemoryKHR-buffer-03783
- [x] VUID-vkWriteAccelerationStructuresPropertiesKHR-buffer-03784

-----

`must be valid accelerationStructure to get address`

- [x] VUID-vkGetAccelerationStructureDeviceAddressKHR-pInfo-09541
- [x] VUID-vkGetAccelerationStructureDeviceAddressKHR-pInfo-09542

-----

`must be bound to device memory`

- [x] VUID-vkCmdWriteAccelerationStructuresPropertiesKHR-buffer-03736

-----

`bound to host-visible device memory`

- [x] VUID-vkCopyAccelerationStructureToMemoryKHR-buffer-03731

-----

`aligned to`

- [x] VUID-vkCopyMemoryToAccelerationStructureKHR-pInfo-03750

-----

`maxInstanceCount` 

like `VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03801`

- [x] VUID-vkGetAccelerationStructureBuildSizesKHR-pBuildInfo-03785

-----

`simple feature bit`

- [x] VUID-vkGetAccelerationStructureDeviceAddressKHR-accelerationStructure-08935

-----

`check if null pointer`

- [x] VUID-vkWriteAccelerationStructuresPropertiesKHR-queryType-03449
- [x] VUID-vkWriteAccelerationStructuresPropertiesKHR-queryType-03451
- [x] VUID-vkWriteAccelerationStructuresPropertiesKHR-queryType-06732
- [x] VUID-vkWriteAccelerationStructuresPropertiesKHR-queryType-06734

-----

`have been built prior to the execution of this command`

[...]

PR #7478, opened 2024-02-08
layers: Add 07791

arno-lunarg:
Part of https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7361

### DIFF
--- a/layers/core_checks/cc_device_memory.cpp
+++ b/layers/core_checks/cc_device_memory.cpp
@@ -92,6 +92,33 @@ bool CoreChecks::ValidateMemoryIsBoundToImage(const LogObjectList &objlist, cons
     return result;
 }
 
+bool CoreChecks::ValidateAccelStructsMemoryDoNotOverlap(const Location &function_loc, LogObjectList objlist,
+                                                        const vvl::AccelerationStructureKHR &accel_struct_a, const Location &loc_a,
+                                                        const vvl::AccelerationStructureKHR &accel_struct_b, const Location &loc_b,
+                                                        const char *vuid) const {
+    bool skip = false;
+
+    const vvl::Buffer &buffer_a = *accel_struct_a.buffer_state;
+    const vvl::Buffer &buffer_b = *accel_struct_b.buffer_state;
+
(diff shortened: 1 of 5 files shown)


layers/core_checks/cc_device_memory.cpp
```diff
@@ -92,6 +92,33 @@ bool CoreChecks::ValidateMemoryIsBoundToImage(const LogObjectList &objlist, cons
...
+    const vvl::Buffer &buffer_b = *accel_struct_b.buffer_state;
+
+    const sparse_container::range<VkDeviceSize> range_a(accel_struct_a.create_infoKHR.offset, accel_struct_a.create_infoKHR.size);
+    const sparse_container::range<VkDeviceSize> range_b(accel_struct_b.create_infoKHR.offset, accel_struct_b.create_infoKHR.size);
+
+    if (const auto [memory, overlap_range] = buffer_a.GetResourceMemoryOverlap(range_a, &buffer_b, range_b);
+        memory != VK_NULL_HANDLE) {
+        objlist.add(accel_struct_a.Handle(), buffer_a.Handle(), accel_struct_b.Handle(), buffer_b.Handle());
+
+        skip |= LogError(vuid, objlist, function_loc,
```
spencer-lunarg, line 111: you can just go `loc_a.function` here instead and then you don't need to pass in `function_loc`

layers/error_message/logging.h
```diff
@@ -56,6 +56,11 @@ struct LogObjectList {
         object_list.emplace_back(object, ConvertCoreObjectToVulkanObject(VkHandleInfo<HANDLE_T>::kVkObjectType));
     }
 
+    template <typename... HANDLE_T>
+    void add(HANDLE_T... objects) {
```
spencer-lunarg, line 60: honestly, I realize I should have done this a while ago! there are for sure many other functions I can now go back and do the same pattern! (thanks for adding this)

spencer-lunarg (approved):
With removal of extra `const Location& function_loc`

layers/core_checks/cc_device_memory.cpp
arno-lunarg: Ok, will change this in a subsequent PR, I do not want to go through CI again x)

Issue #7456, opened 2024-02-06
False positive in sync validation w.r.t. sync2 access flags


HansKristian-Work:
**Environment:**
 - OS: Arch Linux
 - GPU and driver version: NVIDIA RTX 4070 535.43.25
 - SDK or header version if building from repo: VVL main
 - Options enabled (synchronization, best practices, etc.): synchronization

**Describe the Issue**

Recent VVL started to spam false positive sync validation messages that should not be possible. It's possible there are issues with cross-queue tracking.

```
#!/bin/bash

git clone https://github.com/Themaister/Granite
cd Granite
git submodule update --init
mkdir build
cd build
cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Debug
ninja latency-test
./tests/latency-test
```

**Expected behavior**

No errors

**Valid Usage ID**

```
[ERROR]: [Vulkan]: Validation Error: Validation Error: [ SYNC-HAZARD-READ-AFTER-WRITE ] Object 0: handle = 0x55c0104184a0, type = VK_OBJECT_TYPE_QUEUE; | MessageID = 0xe4d96472 | vkQueueSubmit2():  Hazard READ_AFTER_WRITE for entry 0, VkCommandBuffer 0x55c0112746c0[], Submitted access info (submitted_usage: SYNC_FRAGMENT_SHADER_SHADER_SAMPLED_READ, command: vkCmdDraw, seq_no: 4, reset_no: 1). Access info (prior_usage: SYNC_IMAGE_LAYOUT_TRANSITION, write_barriers: 0, queue: VkQueue 0x55c0104ce390[], submit: 9, batch: 0, batch_tag: 42, command: vkCmdPipelineBarrier2, command_buffer: VkCommandBuffer 0x55c010fcabb0[], seq_no: 3, VkImage 0x6c71370000000186[font], reset_no: 3).
[...]

PR #7480, opened 2024-02-09
sync: Add layout transition access to ALL_COMMANDS scope

artem-lunarg:
`ALL_COMMANDS` includes all operations performed on the gpu, not only operations performed by the official stages.
This was also mentioned here (internal link): https://gitlab.khronos.org/vulkan/vulkan/-/issues/3783#note_453947

The issue was that `ALL_COMMANDS`, when used by the semaphore, did not protect the image layout transition (WRITE) done in the first submission from the shader READ access in the second submission.

Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7456


arno-lunarg (approved):
LGTM!

PR #7487, opened 2024-02-09
layers: Improve RenderPass Compatibility messages

spencer-lunarg:
The renderpass compatibility error messages were not well written and even I had to look at the code to understand them

This both unifies how we print things and adds some useful utils for printing other error strings

### DIFF
--- a/layers/error_message/error_strings.h
+++ b/layers/error_message/error_strings.h
@@ -1,0 +1,51 @@
+/* Copyright (c) 2024 The Khronos Group Inc.
+ * Copyright (c) 2024 Valve Corporation
+ * Copyright (c) 2024 LunarG, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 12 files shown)


layers/error_message/error_strings.h
```diff
@@ -0,0 +1,51 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <string>
+
+[[maybe_unused]] static std::string string_Attachment(uint32_t attachment) {
```
spencer-lunarg, line 21: When this prints a random large value instead of the string `VK_ATTACHMENT_UNUSED` it makes the error message hard to understand

layers/error_message/error_strings.h
```diff
@@ -0,0 +1,51 @@
...
+
+[[maybe_unused]] static std::string string_VkExtent2D(VkExtent2D extent) {
+    std::stringstream ss;
+    ss << "Width = " << extent.width << ", Height = " << extent.height;
+    return ss.str();
+}
+
+[[maybe_unused]] static std::string string_VkExtent3D(VkExtent3D extent) {
+    std::stringstream ss;
+    ss << "Width = " << extent.width << ", Height = " << extent.height << ", Depth = " << extent.depth;
```
artem-lunarg, line 37: should this be capital letter consistent with VkOffset functions  (Width <-> width)?

artem-lunarg (approved):

layers/error_message/error_strings.h
spencer-lunarg: ya, I think that is a fair stance to take, make anything in this file consistent with the spec... will update

PR #7494, opened 2024-02-11
tests: Fix sparse queues

spencer-lunarg:
RADV in Mesa 24 move there Sparse Queue to be by itself (instead of combined with the graphics/compute main queue family)

This caused various tests to fail on it

### DIFF
--- a/tests/unit/sparse_buffer.cpp
+++ b/tests/unit/sparse_buffer.cpp
@@ -21,9 +21,8 @@ TEST_F(NegativeSparseBuffer, QueueBindSparseMemoryBindSize) {
     AddRequiredFeature(vkt::Feature::sparseBinding);
     RETURN_IF_SKIP(Init());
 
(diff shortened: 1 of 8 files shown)


tests/unit/sparse_buffer.cpp
```diff
@@ -52,7 +52,7 @@ TEST_F(NegativeSparseBuffer, QueueBindSparseMemoryBindSize) {
     bind_info.bufferBindCount = 1;
     bind_info.pBufferBinds = &buffer_memory_bind_info;
 
-    VkQueue sparse_queue = m_device->graphics_queues()[sparse_index.value()]->handle();
+    VkQueue sparse_queue = m_device->sparse_queues()[0]->handle();
```
artem-lunarg, line 55: nit: places like line 24 can also use sparse_queues()

artem-lunarg (approved):

tests/unit/sparse_buffer.cpp
spencer-lunarg: good call, let me fix them to be consistent and use `sparse_queues()`

Issue #7495, opened 2024-02-11
Dynamic rendering: missing validation for MSAA


QuantumDeveloper:
If MSAA is present and `ResolveImageView != null`, but `ResolveMode = None`, validation will not say anything. 
This will lead to situation when swapchain will not resolve render target and window will stays black.
Very hard to find root cause without validation error messages

Vulkan SDK: 1.3.275.0

spencer-lunarg:
Thanks for raising the issue, but I want to clarify what API calls specifically are are talking about

> MSAA is present

Did you mean you had `VkPipelineMultisampleStateCreateInfo::rasterizationSamples` to something like `VK_SAMPLE_COUNT_4_BIT` ? 

If this is what you mean, then the issue would not be noticed until a `vkCmdDraw` when you match up the Rendering object with the bound pipeline

QuantumDeveloper:
I mean this API: `VkRenderingAttachmentInfo`

![image](https://github.com/KhronosGroup/Vulkan-ValidationLayers/assets/6085009/5a850d39-f161-48c3-902b-956c400e754d)

spencer-lunarg:
So I looked at the Spec, and if `resolveMode = None` the `resolveImageView` should be ignored

![image](https://github.com/KhronosGroup/Vulkan-ValidationLayers/assets/115671160/d6ae0e7e-9fc5-4852-911a-0c3d1a6a79f8)

So I guess was the issue that you value you passed into `resolveImageView` never got written too and that is why the image is black... if this is the case, there is nothing "invalid" about this, but would be a good thing to add to best practice layers to catch

QuantumDeveloper:
Yep, at least just warn a user about this situation. This was working for nearly a year, but some time ago stopped. This issue took the whole day to find, so I thought someone could also be struggling from this issue

spencer-lunarg:
Agree... there are many other subtle "is ignored" language in the spec that I feel would also be easy to not realize like this... Will get a warning in for this case, but then will look into maybe a larger way to catch the other hundreds of cases like this

Thanks for bringing this up, I appreciate being aware of these sharp edges of using Vulkan

PR #7499, opened 2024-02-12
bp: Resolve mode none check

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7495


artem-lunarg (approved):
LGTM. In additional to situation where this helps finding error, I guess it's also possible to have a situation where it was done intentionally? Structure is pre-initialized and only resolve mode gets updated to toggle behavior. Could be one of the reason that spec says that specific fields are ignored instead of forcing them to be null.

spencer-lunarg:
@artem-lunarg yes, this is hitting a more philosophical issue, posted on internal chat, I think we basically need to move a check like this to a "opt-in, I am getting nothing to render" option

Issue #7502, opened 2024-02-12
Queue submit synchronization validation asserts/crashes in `GetDebugRegionName()`


glebov-andrey:
**Environment:**
 - OS: Linux
 - SDK or header version if building from repo: current `main` (9d00ae6c668895d3af97ad57e4f1771797cf3a16)
 - Options enabled (synchronization, best practices, etc.): synchronization (with queue submit validation enabled)

**Describe the Issue**

Synchronization validation crashes when queue submit validation is enabled with the assertion `record.label_command_index < label_commands.size()` in `GetDebugRegionName()` while formatting an error (called from `QueueBatchContext::FormatUsage()`).

There seem to be two cases this happens.

1. The first case is relatively simple - the value of `record.label_command_index` was `kU32Max`.
   I found another place where the `debug_name_provider` was set to `nullptr` in such a case, so I added
   ```c++
   const auto debug_name_provider = (record.label_command_index == vvl::kU32Max) ? nullptr : 
   access.debug_name_provider;
   ```
   to `FormatUsage()`, and it stopped crashing during the first frame.

2. In the second case `record.label_command_index` is `25` and `label_commands.size()` is `6`, which seems to be caused by the use of a timeline semaphore for device-to-host synchronization.
   I know timeline semaphores aren't properly supported by synchronization validation but it would be nice if it wouldn't crash :) Also I seem to recall that there is enough timeline semaphore state tracking for object lifetime checking, so maybe command buffer execution tracking is also possible.

   What appears to be happening is that VVL doesn't know that the command buffer's execution has been waited for by the host. As a result, when they get reused for different passes with a different amount of debug labels, this assertion is triggered. I'm pretty sure that this is due to command buffer reuse because the labels in `label_commands` are from a graphics pass while the command which triggers the error is a `vkCmdDispatch`.

   If I use an extra set of fences for device-to-host synchronization the error goes away (this doesn't affect the `kU32Max` case).

artem-lunarg:
@glebov-andrey thanks for catching this. The first part should be fixed (assuming that the issue I found is the same issue you found).

glebov-andrey:
@artem-lunarg I can confirm that #7503 fixes the first case

artem-lunarg:
@glebov-andrey The second issue with timelines, is it the same application that also had the first issue? Wondering if in the application with timeline semaphores you still can get some benefits from debug labels (or sync validation in general), for example, from the fixed first case.

[...]

PR #7504, opened 2024-02-12
sync: Fix label code crash when timeline semaphores are used

artem-lunarg:
Timeline semaphores are not supported yet but the code should not crash

Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7502


spencer-lunarg (approved):
This is a nice way to transition to timeline semaphore

PR #7505, opened 2024-02-13
scripts: Update SPIRV known good

spencer-lunarg:
`spirv-val` was updated to disallow physical storage buffer as interfaces (`VUID-StandaloneSpirv-Input-09557`)
https://github.com/KhronosGroup/SPIRV-Tools/pull/5539#event-11775591330

Move tests that are effected


jeremyg-lunarg (approved):
LGTM once you fix the pesky generate_source.py problem.

Issue #7362, opened 2024-01-25
Acceleration Structures missing in Device/Indirect call


spencer-lunarg:
These VUs we already have in `vkCmdBuildAccelerationStructuresKHR` and think they are just simply missing a case from the other 2 calls

- [x] VUID-vkBuildAccelerationStructuresKHR-pInfos-03675
- [x] VUID-vkBuildAccelerationStructuresKHR-pInfos-03668
- [x] VUID-vkBuildAccelerationStructuresKHR-pInfos-03761
- [x] VUID-vkBuildAccelerationStructuresKHR-pInfos-03762
- [x] VUID-vkBuildAccelerationStructuresKHR-pInfos-03763
- [x] VUID-vkBuildAccelerationStructuresKHR-pInfos-03764
- [x] VUID-vkBuildAccelerationStructuresKHR-pInfos-03765
- [x] VUID-vkBuildAccelerationStructuresKHR-pInfos-03766
- [x] VUID-vkBuildAccelerationStructuresKHR-pInfos-03767
- [x] VUID-vkBuildAccelerationStructuresKHR-ppBuildRangeInfos-03676
- [x] VUID-vkBuildAccelerationStructuresKHR-primitiveCount-03769
- [x] VUID-vkBuildAccelerationStructuresKHR-dstAccelerationStructure-03701
- [x] VUID-vkBuildAccelerationStructuresKHR-dstAccelerationStructure-03702
- [x] VUID-vkBuildAccelerationStructuresKHR-firstVertex-03770

(With the indirect, there might be some that are GPU-AV - please mark and ignore)

- [x] VUID-vkCmdBuildAccelerationStructuresIndirectKHR-pInfos-03671
- [x] VUID-vkCmdBuildAccelerationStructuresIndirectKHR-pInfos-03672
- [x] VUID-vkCmdBuildAccelerationStructuresIndirectKHR-geometry-03673
- [x] VUID-vkCmdBuildAccelerationStructuresIndirectKHR-pInfos-03674
- [x] VUID-vkCmdBuildAccelerationStructuresIndirectKHR-dstAccelerationStructure-03701
- [x] VUID-vkCmdBuildAccelerationStructuresIndirectKHR-dstAccelerationStructure-03702
- [x] VUID-vkCmdBuildAccelerationStructuresIndirectKHR-pInfos-03711
- [x] VUID-vkCmdBuildAccelerationStructuresIndirectKHR-pInfos-03712
[...]

PR #7508, opened 2024-02-13
layers: CommonBuildAccelerationStructureValidation

arno-lunarg:
Part of https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7362

Refacto a bunch of `vkCmdBuildAccelerationStructuresKHR` validation code to validate host and indirect builds

### DIFF
--- a/layers/core_checks/cc_ray_tracing.cpp
+++ b/layers/core_checks/cc_ray_tracing.cpp
@@ -151,7 +151,7 @@ bool CoreChecks::PreCallValidateGetAccelerationStructureHandleNV(VkDevice device
     return skip;
 }
 
-bool CoreChecks::ValidateAccelerationStructuresMemoryAlisasing(VkCommandBuffer commandBuffer, uint32_t infoCount,
+bool CoreChecks::ValidateAccelerationStructuresMemoryAlisasing(const LogObjectList &objlist, uint32_t infoCount,
                                                                const VkAccelerationStructureBuildGeometryInfoKHR *pInfos,
                                                                uint32_t info_i, const ErrorObject &error_obj) const {
     using sparse_container::range;
@@ -162,11 +162,16 @@ bool CoreChecks::ValidateAccelerationStructuresMemoryAlisasing(VkCommandBuffer c
     const auto src_as_state = Get<vvl::AccelerationStructureKHR>(info.srcAccelerationStructure);
     const auto dst_as_state = Get<vvl::AccelerationStructureKHR>(info.dstAccelerationStructure);
 
-    if (info.srcAccelerationStructure != info.dstAccelerationStructure && src_as_state && dst_as_state) {
-        skip |= ValidateAccelStructsMemoryDoNotOverlap(error_obj.location, LogObjectList(commandBuffer), *src_as_state,
+    if (info.mode == VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR &&
+        info.srcAccelerationStructure != info.dstAccelerationStructure && src_as_state && dst_as_state) {
+        const char *vuid = error_obj.location.function == Func::vkCmdBuildAccelerationStructuresKHR
+                               ? "VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03668"
+                           : error_obj.location.function == Func::vkCmdBuildAccelerationStructuresIndirectKHR
+                               ? "VUID-vkCmdBuildAccelerationStructuresIndirectKHR-pInfos-03668"
+                               : "VUID-vkBuildAccelerationStructuresKHR-pInfos-03668";
+        skip |= ValidateAccelStructsMemoryDoNotOverlap(error_obj.location, objlist, *src_as_state,
                                                        info_i_loc.dot(Field::srcAccelerationStructure), *dst_as_state,
-                                                       info_i_loc.dot(Field::dstAccelerationStructure),
-                                                       "VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03668");
+                                                       info_i_loc.dot(Field::dstAccelerationStructure), vuid);
     }
 
     for (auto [other_info_j, other_info] : vvl::enumerate(pInfos + info_i, infoCount - info_i)) {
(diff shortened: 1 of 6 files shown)


layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -162,11 +162,16 @@ bool CoreChecks::ValidateAccelerationStructuresMemoryAlisasing(VkCommandBuffer c
     const auto src_as_state = Get<vvl::AccelerationStructureKHR>(info.srcAccelerationStructure);
     const auto dst_as_state = Get<vvl::AccelerationStructureKHR>(info.dstAccelerationStructure);
 
-    if (info.srcAccelerationStructure != info.dstAccelerationStructure && src_as_state && dst_as_state) {
-        skip |= ValidateAccelStructsMemoryDoNotOverlap(error_obj.location, LogObjectList(commandBuffer), *src_as_state,
+    if (info.mode == VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR &&
+        info.srcAccelerationStructure != info.dstAccelerationStructure && src_as_state && dst_as_state) {
+        const char *vuid = error_obj.location.function == Func::vkCmdBuildAccelerationStructuresKHR
```
spencer-lunarg, line 167: realize this wasn't ideal, but for now it works

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -475,35 +511,44 @@ bool CoreChecks::ValidateAccelerationBuffers(VkCommandBuffer cmd_buffer, uint32_
...
-                         info_loc.dot(Field::scratchData).dot(Field::deviceAddress),
+        skip |= LogError(pick_vu("VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03802",
+                                 "VUID-vkCmdBuildAccelerationStructuresIndirectKHR-pInfos-03802"),
+                         device, info_loc.dot(Field::scratchData).dot(Field::deviceAddress),
                          "(0x%" PRIx64 ") has no buffer is associated with it.", info.scratchData.deviceAddress);
     } else {
-        const VkDeviceSize scratch_size = rt::ComputeScratchSize(device, info, geometry_build_ranges);
+        const VkDeviceSize scratch_size = info_loc.function == Func::vkCmdBuildAccelerationStructuresKHR
+                                              ? rt::ComputeScratchSize(rt::BuildType::Device, device, info, geometry_build_ranges)
+                                              : 1;
```
spencer-lunarg, line 539: I guess don't understand why `1` is the `scratch_size` for an indirect, worth a comment

layers/core_checks/core_validation.h
```diff
@@ -1562,7 +1562,9 @@ class CoreChecks : public ValidationStateTracker {
                                      const VkAccelerationStructureBuildGeometryInfoKHR& info,
                                      const VkAccelerationStructureBuildRangeInfoKHR* geometry_build_ranges,
                                      const Location& info_loc) const;
-    bool ValidateAccelerationStructuresMemoryAlisasing(VkCommandBuffer commandBuffer, uint32_t infoCount,
+    bool CommonBuildAccelerationStructureValidation(uint32_t info_i, const VkAccelerationStructureBuildGeometryInfoKHR& info,
```
spencer-lunarg, line 1565: can we not just use `info_loc.index` instead of having to pass in this extra `uint32_t info_i`

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -884,6 +1021,67 @@ bool CoreChecks::PreCallValidateBuildAccelerationStructuresKHR(
...
+        skip |= LogError("VUID-vkCmdBuildAccelerationStructuresIndirectKHR-commandBuffer-09547", commandBuffer, error_obj.location,
+                         "called in a protected command buffer.");
+    }
+
+    for (const auto [info_i, info] : vvl::enumerate(pInfos, infoCount)) {
+        const Location info_loc = error_obj.location.dot(Field::pInfos, info_i);
+        auto src_as_state = Get<vvl::AccelerationStructureKHR>(info->srcAccelerationStructure);
+        auto dst_as_state = Get<vvl::AccelerationStructureKHR>(info->dstAccelerationStructure);
+        skip |=
+            ValidateMemoryIsBoundToBuffer(commandBuffer, *dst_as_state->buffer_state, info_loc.dot(Field::dstAccelerationStructure),
```
spencer-lunarg, line 1045: you deference `dst_as_state` here but are checking for `!dst_as_state` below

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -244,7 +258,12 @@ bool CoreChecks::ValidateAccelerationBuffers(VkCommandBuffer cmd_buffer, uint32_
                                              const Location &info_loc) const {
     bool skip = false;
 
-    auto buffer_check = [this](uint32_t gi, const VkDeviceOrHostAddressConstKHR address, const Location &geom_loc) -> bool {
+    const auto pick_vu = [&info_loc](const char *direct_build_vu, const char *indirect_build_vu) -> const char * {
```
spencer-lunarg, line 261: nit: `pick_vuid`

VU = statement
VUID = The ID

I really like this idea and plan to use it more places now... so I want to not have to go back later and change all the spots

spencer-lunarg (commented):
Looks good, some nits

layers/core_checks/cc_ray_tracing.cpp
arno-lunarg: This is a placeholder, scratch_size cannot be computed in the indirect call. Hard coding a hypothetical minimum size of 1 byte is way easier than adding the logic to not perform validation in the indirect case, so I went for that.
I am going to add a comment!

layers/core_checks/core_validation.h
arno-lunarg: ah yes better

layers/core_checks/cc_ray_tracing.cpp
arno-lunarg: I think my checks for `as_state` being a valid pointer or no are all other the place, I will try to improve that

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -884,6 +1023,68 @@ bool CoreChecks::PreCallValidateBuildAccelerationStructuresKHR(
...
+        skip |= LogError("VUID-vkCmdBuildAccelerationStructuresIndirectKHR-commandBuffer-09547", commandBuffer, error_obj.location,
+                         "called in a protected command buffer.");
+    }
+
+    for (const auto [info_i, info] : vvl::enumerate(pInfos, infoCount)) {
+        const Location info_loc = error_obj.location.dot(Field::pInfos, info_i);
+        auto src_as_state = Get<vvl::AccelerationStructureKHR>(info->srcAccelerationStructure);
+        auto dst_as_state = Get<vvl::AccelerationStructureKHR>(info->dstAccelerationStructure);
+
+        if (dst_as_state) {
```
spencer-lunarg, line 1047: so you don't do this for `src_as_state` version of `ValidateMemoryIsBoundToBuffer`

I would just do something like

```c++
if (src_as_state) {
    // all checks using it
}

if (dst_as_state) {
    // all checks using it
}

// everything that is no dependent on the AC
```

layers/core_checks/cc_ray_tracing.cpp
spencer-lunarg: don't forget this (didn't see it last update)

arno-lunarg: oopsy, done now

layers/core_checks/cc_ray_tracing.cpp
arno-lunarg: I went for that

spencer-lunarg (approved):

PR #7510, opened 2024-02-13
layers: Fix crash with required subgroup size

ziga-lunarg:
It also seems that the shader object equivalent of this VUID is missing


spencer-lunarg (approved):
I haven't got to it yet, but we for need a better way to get the "pipeline and/or shader object info" as this is not obvious to anyone the future that event the `pipeline_create_info` can be null

Thanks for catching and patching these for the short term!

Issue #7506, opened 2024-02-13
Improve GPU-AV bindless descriptor shader instrumentation speed on larger shaders 


spencer-lunarg:
The SPIR-V from `glslang/Test/spv.subgroupExtendedTypesArithmetic.comp` produces 1518 `OpFunctionCalls`

This case takes very long (2 seconds in release, stupid long in release) to run `./tests/spirv/instrumentation in.spv -o out.spv --bindless-descriptor` on

This probably will require re-structuring the logic to be more efficient for inserting function calls

The most likely cause is how the data is held in a `std::vector` and likely is constantly having to adjust the instructions/blocks inside it

spencer-lunarg:
quick profiling, 99% of the time is spent in 

```c++
    // move all remaining instructions to the newly created merge block
    while (inst_it != original_block.instructions_.end()) {
        merge_block.instructions_.emplace_back(std::move(*inst_it));
        inst_it = original_block.instructions_.erase(inst_it);
    }
```

PR #7511, opened 2024-02-14
gpuav: Fix speed regression

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7506

The use of `insert` and `std::make_move_iterator` showed a 20x speed up and mostly got performance on par. The current way we do injection is not ideal for cases where we have tons of access in a huge Function block, but there are other things planned I have to help mitigate that

### DIFF
--- a/layers/gpu_validation/spirv/bindless_descriptor_pass.cpp
+++ b/layers/gpu_validation/spirv/bindless_descriptor_pass.cpp
@@ -429,21 +429,5 @@ bool BindlessDescriptorPass::AnalyzeInstruction(const Function& function, const 
     return true;
 }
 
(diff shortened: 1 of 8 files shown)


arno-lunarg (approved):

jeremyg-lunarg:
> CI Vulkan-ValidationLayers build # 15766 failed.

Internal CI failure is NegativeGpuAV.ValidationInlineUniformBlockAndMiscGpu  on RADV Linux which is an intermittent failure occurring in other runs.

jeremyg-lunarg (approved):
LGTM

Issue #7387, opened 2024-01-27
High memory usage


ziga-lunarg:
Forwarding an issue:

Wanted to suggest an explicit test for VVL static initializer memory usage (particularly by calls to new/malloc from statically initialized constructors) should be added. We've measured up to 12 megabytes of memory dynamically allocated (i.e., by calls to new/malloc) in constructors of statically initialized objects by VVL on our platform, and this seems to (in part) explain it. Compared to other similar layers like RenderDoc's capture layer, we would expect the amount of memory dynamically allocated by static initializes (note: this does not include the actual memory of the statically initialized objects themselves) to be on the order of several hundred kilobytes if any at all. It seems like it would be a good idea to test this in VVL to avoid future regressions.

I would also suggest a general effort to try to track down and reduce this memory usage (and ideally optimize allocation patterns to avoid many small allocations), as it significantly affects the usability of the VVL for developers on more memory constrained platforms.

spencer-lunarg:
without do any looking, one area I am curious about investigating is the current `custom_containers.h` usage of RobinHood for our maps, I know it was pointed out there is a newer version we could be upgrading to, I am curious if we are trading off a lot of memory for faster hash mapping

spencer-lunarg:
update - testing, using RobinHood over the STD implementation shows an improvement in memory footprint

spencer-lunarg:
Did some quick checking how big our state tracking really is

One issue is we have multiple copies of `ValidationStateTracker` between best practice/syncval/core checks/gpu-av so some of this info could be brought up to a global level

```
class ValidationObject {}
class ValidationStateTracker : public ValidationObject {}
class CoreChecks : public ValidationStateTracker {}

// size in bytes
Size of CoreChecks: 31296
Size of ValidationStateTracker: 30080
Size of ValidationObject: 6656
```

For our state object, there are clearly some that are larger than others
```
Size of vvl::StateObject: 160

Size of vvl::CommandBuffer: 3320
Size of vvl::Image: 952
Size of vvl::Pipeline: 688
Size of vvl::RenderPass: 656
Size of vvl::Buffer: 528
Size of vvl::ShaderObject: 472
Size of vvl::Queue: 464
Size of vvl::ImageView: 464
Size of vvl::DescriptorPool: 456
Size of vvl::Fence: 384
Size of vvl::PhysicalDevice: 376
Size of vvl::DeviceMemory: 360
Size of vvl::Semaphore: 352
[...]

PR #7513, opened 2024-02-14
layers: Add Stateless SPIR-V runtime checks

spencer-lunarg:
Part 2 of https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6230
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/2835
helps with https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7387

This (larger then expected) change is moving a lot of validation from Pipeline time to `vkCreateShaderModule` time

The main benefit are a lot of things we were tracking that we really don't need after we do a parse, the lets us do it now right away.

The large diff is because of 2 issues
1. The tests now have to check earlier for the error
2. The state of `state_tracker.h` `#include` was a little awful, created a new `chassis_modification_state.h` to hold all the various structs that are used outside


layers/state_tracker/chassis_modification_state.h
```diff
@@ -0,0 +1,115 @@
+/* Copyright (c) 2015-2024 The Khronos Group Inc.
```
jeremyg-lunarg, line 1: this file will need to be added to BUILD.gn

jeremyg-lunarg (approved):
Whew, this is a massive commit / PR. But LGTM

spencer-lunarg:
> Whew, this is a massive commit / PR

A lot of the complexity came from

1. Having to shuffle things around to not have every header need to include the `shader_module.h` header (this was what prompted my other `#include` cleanup PR)
2. A lot of shader tests were re-written to not need the pipeline to trigger things

Issue #7488, opened 2024-02-09
Remove BUFFER_STATE_PTR


spencer-lunarg:
This is just a `vvl::Buffer *` and no reason to have this alias anymore, we should be able to just use `vvl::Buffer` now

PR #7515, opened 2024-02-15
layers: Remove BUFFER_STATE_PTR

arno-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7488

Is it really an improvement, with this new `const` madness?

EDIT: Since first push, I looked further and remove a useless `const` level, so I think things are fine now. Also, I ended having to reformat entire files, sorry for the noise...

### DIFF
--- a/layers/core_checks/cc_buffer_address.h
+++ b/layers/core_checks/cc_buffer_address.h
@@ -64,30 +64,26 @@ template <size_t N = 1>
 class BufferAddressValidation {
   public:
     // Return true if and only if VU is verified
-    using ValidationFunction = std::function<bool(const ValidationStateTracker::BUFFER_STATE_PTR&, std::string* out_error_msg)>;
+    using ValidationFunction = std::function<bool(vvl::Buffer* const, std::string* out_error_msg)>;
     using ErrorMsgHeaderSuffixFunction = std::function<std::string()>;
 
     struct VuidAndValidation {
         std::string_view vuid{};
-        ValidationFunction validation_func = [](const ValidationStateTracker::BUFFER_STATE_PTR&, std::string* out_error_msg) {
-            return true;
-        };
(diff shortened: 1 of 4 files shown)


spencer-lunarg:
@jeremyg-lunarg thoughts?

arno-lunarg:
> @jeremyg-lunarg thoughts?

I got rid of the const madness, you can still have a look Jeremy but I think things are ok now

layers/core_checks/cc_buffer_address.h
```diff
@@ -64,30 +64,26 @@ template <size_t N = 1>
 class BufferAddressValidation {
   public:
     // Return true if and only if VU is verified
-    using ValidationFunction = std::function<bool(const ValidationStateTracker::BUFFER_STATE_PTR&, std::string* out_error_msg)>;
```
jeremyg-lunarg, line 67: Is it an improvement? Yes because `const ValidationStateTracker::BUFFER_STATE_PTR&` makes absolutely no sense. The 'raw' equivalent of `vvl::Buffer const *&` makes that much more obvious. I wonder what the compiler did with this weird parameter type.

jeremyg-lunarg (commented):
LGTM

spencer-lunarg (approved):

layers/core_checks/cc_buffer_address.h
spencer-lunarg: I agree, rather see the weirdness is there then some random alias name that is not like anything else

jeremyg-lunarg (approved):

layers/core_checks/cc_buffer_address.h
arno-lunarg: I made this because at some point I wanted to be able to easily be able to switch between raw and managed pointers. But now since we are sticking with raw, no need for this indeed

PR #7517, opened 2024-02-15
layers: Move extra state tracker includes

spencer-lunarg:
We have been constantly adding `#include "state_tracker/*` files everywhere in the headers... this has caused certain state tracking files to be unnecessarily included in way more files then needed

This is a slightly large (it had to span the various chassis items) but I feel this was a matter of cleaning it now before it gets progressively worse 

For some header files like `shader_module.h` this already reduces it from `102` to `50`ish files that need to rebuild


layers/gpu_validation/gpu_subclasses.h
```diff
@@ -22,9 +22,18 @@
 
 #include "gpu_validation/gpu_descriptor_set.h"
 #include "gpu_validation/gpu_state_tracker.h"
+#include "gpu_validation/gpu_resources.h"
 #include "generated/vk_object_types.h"
 #include "gpu_shaders/gpu_shaders_constants.h"
 
+// We pull in most the core state tracking files
+// gpu_subclasses.h should NOT be included by any other header file
```
spencer-lunarg, line 30: before, we were having to rebuild all of GPU_AV just because touching unrelated states, the separates the logic better

artem-lunarg (approved):
LGTM. Looks great actually. Few compile errors hopefully should be easy to fix.. It's also nice that part of implementations moved to cpp so it's easier to read what interface offers. Linkers can now inline from different compilation units, so this aspect loses its importance.

artem-lunarg (approved):

spencer-lunarg:
> Nice! How did you find unnecessary dependencies?

By hand, I have tried tools in the past, but a lot of this is just knowing the repo well enough now and looking at a header file and going "wait, why do we need this file?!?" then I have way to just build the source file which tells me quick if it was needed or not.

The way we structure some headers (like `core_validation.h`) is it has to know about everything and then we do the logic in separate `cc_*.cpp` files.. this pattern I feel main down side is it easy to just supress a warning by putting the `include` in a header and over time we built up quit the mess

spencer-lunarg:
CI Failure is for `NegativeGpuAV.ValidationInlineUniformBlockAndMiscGpu` which has been shown in other runs to be unrelated

This change should have no effects to it

Issue #4827, opened 2022-11-18
Crash in vkCreateComputePipelines when using validation layer instrumented compute shader


rumblehhh:
**Describe the Issue**
We're hitting a crash inside vkCreateComputePipelines inside the driver when using a shader module created using the attached compute shader (`acceleration_structure_handle_remap.csh`).
We compile and optimise the glsl compute shader to spv using:
1. glslangValidator -V acceleration_structure_handle_remap.csh -o acceleration_structure_handle_remap_csh-unoptimised.spv -S comp
2. spirv-opt -O acceleration_structure_handle_remap_csh-unoptimised.spv -o acceleration_structure_handle_remap_csh-optimised.spv

I can workaround the crash either by:
a. Disabling validation layers completely
b. Using validation layers and using `acceleration_structure_handle_remap_csh-unoptimised.spv` directly i.e. skipping the spirv-opt step
c. Using validation layers and using `VK_EXT_validation_features` to disable shader validation (VkValidationFeatureDisableEXT-VK_VALIDATION_FEATURE_DISABLE_SHADERS_EXT). I was surprised by this as I hadn't realised the validation layers were instrumenting shader modules by default

We're using glslangValidator, spirv-opt and the validation layers from Vulkan SDK 1.3.231.2.
I've also tried using:
 - spirv-opt from spirv-tools package on Ubuntu jammy (Version: 2022.1+1.3.204.0-2)
 - glslangValidator Glslang Version: 10:11.8.0
 - Vulkan-ValidationLayers from e14ae34206c7f1fff73356c279904f19af95217f.

I've attached the following files:
 - `acceleration_structure_handle_remap.csh` - original glsl compute shader
 - `acceleration_structure_handle_remap_csh-unoptimised.spv` - output of step 1 above.
 - `acceleration_structure_handle_remap_csh-optimised.spv` - output of step 2 above.
 - `spirv-cross_acceleration_structure_handle_remap_csh-unoptimised.csh` - spirv-cross output for `acceleration_structure_handle_remap_csh-unoptimised.spv`
 - `spirv-cross_acceleration_structure_handle_remap_csh-optimised.csh` - spirv-cross output for `acceleration_structure_handle_remap_csh-optimised.spv`
 - `spirv_dis_acceleration_structure_handle_remap_csh-optimised.spv.txt` - spirv-dis output of the spirv used when creating the shader module when using c above (disabling shader validation)
[...]

Issue #7462, opened 2024-02-06
Fix edge cases for GPU-AV shader instrumentation


spencer-lunarg:
There are some SPIR-V edge cases that need to be handled, tests were added (but disabled) in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7463)

update: likely related same issue
-  https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4577

PR #7519, opened 2024-02-16
gpuav: Fix creating invalid SPIR-V from loop headers

spencer-lunarg:
> "when you look at the edge cases and the things that make control flow hard, 95% of them come from loops"
>
> ~ https://youtu.be/7C4PmtqBVqo?si=PNeLcGijKCNgqBdV&t=973

closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4827

I found that when dealing with a `OpPhi` inside a `Loop header block` it gets really tricky to generate valid SPIR-V that still injects the needed function call.

Since this is a rare case, I rather just ignore it for now (tracked in https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7462) as generating invalid Control Flow is much worse

### DIFF
--- a/layers/gpu_validation/spirv/function_basic_block.h
+++ b/layers/gpu_validation/spirv/function_basic_block.h
@@ -49,6 +49,8 @@ struct BasicBlock {
 
     InstructionList instructions_;
     Function& function_;
+
+    bool loop_header_ = false;
 };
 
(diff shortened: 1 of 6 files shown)


jeremyg-lunarg (approved):
Failing internal CI in an unrelated test

PR #7530, opened 2024-02-19
tests: Remove VkImageObj

spencer-lunarg:
There are many misc fixes that went along my journey to this, but the main thing is `VkImageObj` is now fully gone for `vkt::Image`


artem-lunarg (approved):
LGTM. Thanks for doing this.
Another refactoring idea. It would be nice if we do something to allow types like vkt::Device to be stored as value (might need additional Init() step), to get rid of `*`s.

Issue #7386, opened 2024-01-26
Likely false-positive WRITE-RACING-WRITE error


thevaber:
I'm getting strange WRITE-RACING-WRITE validation errors in relatively simple scenarios of image layout transitions mixed together with submissions into multiple queues, and after a brief sanity-checking discussion on Discord, I'm creating an issue here. I managed to isolate a very simple case which triggers the error and implemented it within the VVL testing codebase so that it can be easily reproduced, the code is short enough to just include within this issue, and it can be simply pasted into e.g., sync_val_positive.cpp:

```cpp
TEST_F(PositiveSyncVal, TestTwoQueuesImageLayoutTransition) {
    // (boilerplate)
    SetTargetApiVersion(VK_API_VERSION_1_2);
    AddRequiredExtensions(VK_KHR_SYNCHRONIZATION_2_EXTENSION_NAME);
    AddRequiredFeature(vkt::Feature::synchronization2);
    RETURN_IF_SKIP(InitSyncValFramework());
    RETURN_IF_SKIP(InitState());
    QSTestContext test(m_device);
    if (!test.Valid()) {
        GTEST_SKIP() << "Test requires a valid queue object.";
    }

    // (initialization)
    VkImageObj image(m_device);
    image.Init(64, 64, 1, VK_FORMAT_R8_UNORM, VK_IMAGE_USAGE_TRANSFER_DST_BIT);

    vkt::Buffer buffer(*m_device, 64 * 64, VK_BUFFER_USAGE_TRANSFER_SRC_BIT);
    VkBufferImageCopy region = {};
    region.bufferRowLength = 64;
    region.bufferImageHeight = 64;
    region.imageSubresource.aspectMask = VK_IMAGE_ASPECT_COLOR_BIT;
    region.imageSubresource.layerCount = 1;
    region.imageExtent.height = 64;
    region.imageExtent.width = 64;
    region.imageExtent.depth = 1;

    ///
    // the main part
    test.BeginA();
    VkImageMemoryBarrier2KHR image_barrier = vku::InitStructHelper();
    image_barrier.srcStageMask = VK_PIPELINE_STAGE_2_NONE;
    image_barrier.dstStageMask = VK_PIPELINE_STAGE_2_TRANSFER_BIT;
    image_barrier.srcAccessMask = VK_ACCESS_2_NONE;
[...]

Issue #7455, opened 2024-02-06
False positive in sync validation for swapchain image


HansKristian-Work:
**Environment:**
 - OS: Arch Linux
 - GPU and driver version: NVIDIA RTX 4070 535.43.25
 - SDK or header version if building from repo: VVL main
 - Options enabled (synchronization, best practices, etc.): synchronization

**Describe the Issue**

Trivial use of swapchain images triggers false positive sync val errors. The frame in question only renders a single render pass to swapchain with appropriate semaphores in between. To reproduce:

```
#!/bin/bash

git clone https://github.com/Themaister/Granite
cd Granite
git submodule update --init
mkdir build
cd build
cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Debug
ninja gltf-viewer-simple
./tests/gltf-viewer-simple ~/git/glTF-Sample-Models/2.0/Suzanne/glTF/Suzanne.gltf # From KhronosGroup/glTF-Sample-Models
```

**Expected behavior**

No errors

**Valid Usage ID**

```
[ERROR]: [Vulkan]: Validation Error: Validation Error: [ SYNC-HAZARD-WRITE-RACING-WRITE ] Object 0: handle = 0x601067a25280, type = VK_OBJECT_TYPE_QUEUE; | MessageID = 0x743c6069 | vkQueueSubmit2():  Hazard WRITE_RACING_WRITE for entry 0, VkCommandBuffer 0x60106949c5a0[], Submitted access info (submitted_usage: SYNC_IMAGE_LAYOUT_TRANSITION, command: vkCmdBeginRenderPass, seq_no: 1, renderpass: VkRenderPass 0x17f18f00000003cb[], reset_no: 793). Access info (prior_usage: SYNC_PRESENT_ENGINE_SYNCVAL_PRESENT_PRESENTED_SYNCVAL, write_barriers: 0, queue: VkQueue 0x601067a25280[], submit: 143, batch: 0, batch_tag: 3229, vkQueuePresentKHR present_tag:3229, pSwapchains[0]: VkSwapchainKHR 0x948cf100000003cd[], image_index: 1image: VkImage 0xdd1b7300000003cf[backbuffer]).
[...]

PR #7532, opened 2024-02-19
sync: Filter accesses in async hazard detection

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7386, https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7455.

Only the accesses from the async queue participate in async hazard detection.
The resolved accesses from the other contexts might belong to different queues
and do not create async hazards.

The positive test is a regression test from:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7386.

The `QSWriteRacingWrite`/`QSWriteRacingWrite2` negative tests were added to
increase `WRITE-RACING-WRITE` coverage, but they do not reproduce the
issue that was fixed.

The `QSWriteRacingRead` test is a regression test for the initial
incorrect attempt to fix the issue (using scalar async tag tracking
instead of per-queue), which nevertheless did not cause any failures.

### DIFF
--- a/tests/unit/sync_val_positive.cpp
+++ b/tests/unit/sync_val_positive.cpp
@@ -1344,3 +1344,86 @@ TEST_F(PositiveSyncVal, UpdateBuffer) {
     vk::CmdCopyBuffer(*m_commandBuffer, src_buffer, dst_buffer, 1, &region);
     m_commandBuffer->end();
 }
+
+TEST_F(PositiveSyncVal, QSSynchronizedWritesAndAsyncWait) {
+    TEST_DESCRIPTION(
+        "Graphics queue: Image transition and subsequent image write are synchronized using a pipeline barrier. Transfer queue: "
+        "waits for the image layout transition using a semaphore. This should not affect pipeline barrier on the graphics queue");
+
+    SetTargetApiVersion(VK_API_VERSION_1_3);
+    VkPhysicalDeviceSynchronization2Features sync2_features = vku::InitStructHelper();
+    sync2_features.synchronization2 = VK_TRUE;
+    RETURN_IF_SKIP(InitSyncValFramework());
+    RETURN_IF_SKIP(InitState(nullptr, &sync2_features));
+
+    const std::optional<uint32_t> transfer_family =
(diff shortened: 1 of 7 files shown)


tests/unit/sync_val_positive.cpp
```diff
@@ -1152,3 +1152,88 @@ TEST_F(PositiveSyncVal, DynamicRenderingDepthResolve) {
...
+    RETURN_IF_SKIP(InitState(nullptr, &sync2_features));
+
+    const std::optional<uint32_t> transfer_family =
+        m_device->QueueFamilyMatching(VK_QUEUE_TRANSFER_BIT, (VK_QUEUE_COMPUTE_BIT | VK_QUEUE_GRAPHICS_BIT));
+    if (!transfer_family) {
+        GTEST_SKIP() << "Transfer-only queue family is not present";
+    }
+    vkt::Queue *transfer_queue = m_device->queue_family_queues(transfer_family.value())[0].get();
+
+    VkImageObj image(m_device);
```
spencer-lunarg, line 1174:
```suggestion
    vkt::Image image(*m_device, 64, 64, 1, VK_FORMAT_R8_UNORM, VK_IMAGE_USAGE_TRANSFER_DST_BIT);
```

spencer-lunarg (approved):
LGTM - @jzulauf-lunarg would be nice to confirm as well

tests/unit/sync_val_positive.cpp
artem-lunarg: Fixed

jzulauf-lunarg (changes requested):
queue_sync_tag_ is the global of each *other* queue with the one owning this QBC. Can't be scalar.  Algorithmic review with author done, for new approach.

Test gaps identified.

artem-lunarg:
Added regression test to show that previous approach with scalar asyn tag tracking is incorrect.

artem-lunarg:
@jzulauf-lunarg the author reworked the solution and also additional test was added to cause regression on initial broken implementation.

jzulauf-lunarg (approved):
LGTM -- thanks for the additional tests.

Issue #7406, opened 2024-01-30
Pipeline barrier validation seems to classify vkCmdFillBuffer() as a copy command instead of a clear command


litherum:
**Environment:**
 - OS: Windows 11 Pro 23H2 22631.3085 Windows Feature Experience Pack 1000.22684.1000.0
 - GPU and driver version: NVIDIA GeForce RTX 2080 Ti 31.0.15.5123
 - SDK or header version if building from repo: 1.3.271 
 - Options enabled (synchronization, best practices, etc.): `VK_LAYER_KHRONOS_validation`, `VK_EXT_debug_utils`, `VK_EXT_validation_features`, `Synchronization2`

**Describe the Issue**

Create 2 buffers: a source buffer and a destination buffer. The goal is to use `vkCmdFillBuffer()` to fill the first buffer with 0s, and then use `vkCmdCopyBuffer2()` to copy the 0s from the first buffer to the second buffer. Pipeline barriers are needed before and between the two commands.

The spec [lists](https://registry.khronos.org/vulkan/specs/1.3/html/chap18.html#vkCmdFillBuffer) `vkCmdFillBuffer()` as a clear command, so the first pipeline barrier needs to transition the source buffer from `VK_PIPELINE_STAGE_2_NONE`/`VK_ACCESS_2_NONE` to `VK_PIPELINE_STAGE_2_CLEAR_BIT`/`VK_ACCESS_2_TRANSFER_WRITE_BIT`. Then `vkCmdFillBuffer()` can fill it.

Then, we need to transition the source buffer to `VK_PIPELINE_STAGE_2_COPY_BIT`/`VK_ACCESS_2_TRANSFER_READ_BIT` to get it ready to be the source of `vkCmdCopyBuffer2()`. (The destination buffer also needs to be transitioned from  `VK_PIPELINE_STAGE_2_NONE`/`VK_ACCESS_2_NONE` to `VK_PIPELINE_STAGE_2_COPY_BIT`/`VK_ACCESS_2_TRANSFER_WRITE_BIT`.) Then, we can `vkCmdCopyBuffer2()` the source buffer to the destination buffer.

However, at this point, the validation layers say there is a read-after-write hazard on the source buffer, indicating that its pipeline barrier between the two commands is incorrect.

[...]

PR #7543, opened 2024-02-20
sync: Use CLEAR transfer stage for vkCmdFillBuffer

artem-lunarg:
`vkCmdFillBuffer` is a transfer clear command, and not a transfer copy command. I make this assumption because this command is described in the "Clear commands" section of the specification and not in the "Copy commands" section.

> VK_PIPELINE_STAGE_2_COPY_BIT specifies the execution of all [copy commands](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#copies)

> VK_PIPELINE_STAGE_2_CLEAR_BIT specifies the execution of [clear commands](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#clears)

Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7406

### DIFF
--- a/tests/unit/sync_val.cpp
+++ b/tests/unit/sync_val.cpp
@@ -6256,3 +6256,119 @@ TEST_F(NegativeSyncVal, UseShaderReadAccessForUniformBuffer) {
     m_errorMonitor->VerifyFound();
     m_commandBuffer->end();
 }
+
+TEST_F(NegativeSyncVal, FillBufferMissingBarrier) {
+    TEST_DESCRIPTION("Missing synchronization with vkCmdFillBuffer TRANSFER_WRITE access");
+    RETURN_IF_SKIP(InitSyncValFramework());
+    RETURN_IF_SKIP(InitState());
+
+    constexpr VkDeviceSize size = 1024;
+    vkt::Buffer src_buffer(*m_device, size, VK_BUFFER_USAGE_TRANSFER_DST_BIT | VK_BUFFER_USAGE_TRANSFER_SRC_BIT);
+    vkt::Buffer dst_buffer(*m_device, size, VK_BUFFER_USAGE_TRANSFER_DST_BIT);
+
+    VkBufferCopy region{};
+    region.size = size;
+
+    m_commandBuffer->begin();
+    vk::CmdFillBuffer(*m_commandBuffer, src_buffer, 0, size, 42);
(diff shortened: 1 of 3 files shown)


jzulauf-lunarg (changes requested):
vkCmdUpdateBuffer is also in the clear commands section, but uses `SYNC_COPY_TRANSFER_WRITE`.  Probably should also be updated.

tests/unit/sync_val_positive.cpp
```diff
@@ -1139,3 +1139,73 @@ TEST_F(PositiveSyncVal, DynamicRenderingDepthResolve) {
     vk::CmdEndRendering(*m_commandBuffer);
     m_commandBuffer->end();
 }
+
+TEST_F(PositiveSyncVal, FillBuffer) {
```
jzulauf-lunarg, line 1143: Should have a negative for this as well.

artem-lunarg: Added negative tests both for CmdUpdateBuffer and CmdFillBuffer. CmdFillBuffer already had a negative use case in `BufferCopyHazards` but it's nice to have a dedicated small test too.

tests/unit/sync_val.cpp
```diff
@@ -6256,3 +6256,45 @@ TEST_F(NegativeSyncVal, UseShaderReadAccessForUniformBuffer) {
...
+    constexpr VkDeviceSize size = 1024;
+    vkt::Buffer src_buffer(*m_device, size, VK_BUFFER_USAGE_TRANSFER_DST_BIT | VK_BUFFER_USAGE_TRANSFER_SRC_BIT);
+    vkt::Buffer dst_buffer(*m_device, size, VK_BUFFER_USAGE_TRANSFER_DST_BIT);
+
+    VkBufferCopy region{};
+    region.size = size;
+
+    m_commandBuffer->begin();
+    vk::CmdFillBuffer(*m_commandBuffer, src_buffer, 0, size, 42);
+    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "SYNC-HAZARD-READ-AFTER-WRITE");
```
jzulauf-lunarg, line 6274: What I was thinking about was something that had a barrier for the wrong stage either source or dest...

artem-lunarg: Added wrong barriers on both sides of the conflict.

spencer-lunarg (approved):
LGTM - Zulaf comments seem to be have addressed

jzulauf-lunarg (commented):
LGTM

PR #7548, opened 2024-02-21
layers: Fix VUID-vkCmdClearAttachments-pRects-00016 bounding

zmike:
a rect with geometry 0,0,4,4 is contained within an identically sized rect 0,0,4,4. this fixes false positives returned when that test returned false


zmike:
Uh, yeah, that's a good point. Disregard, I'll find better bugs.

Issue #5954, opened 2023-06-02
Unexpected VUID-vkCmdDraw-None-02721 when using dynamic vertex inputs


rg3igalia:
**Environment:**
 - OS: Linux
 - GPU: Intel UHD Graphics 620
 - SDK or header version if building from repo: 1b91bc94bef747cd460898f75e2591af9aa5f05e
 - Options enabled (synchronization, best practices, etc.): defaults+synchronization

**Describe the Issue**

While working on a new set of CTS tests related to input attribute offsets, I'm unexpectedly hitting VUID-vkCmdDraw-None-02721 on some variants of them, when the vertex input is set dynamically:

> ERROR: Validation Error: [ VUID-vkCmdDraw-None-02721 ] Object 0: handle = 0x93d4f20000000089, type = VK_OBJECT_TYPE_BUFFER; Object 1: handle = 0x8f5f070000000095, type = VK_OBJECT_TYPE_PIPELINE; | MessageID = 0x99ef63bb | vkCmdDraw: Format VK_FORMAT_R32G32B32A32_SFLOAT has an alignment of 4 but the alignment of attribAddress (17) is not aligned in pVertexAttributeDescriptions[0](binding=0 location=0) where attribAddress = vertex buffer offset (1) + binding stride (16) + attribute offset (0). The Vulkan spec states: For a given vertex buffer binding, any attribute data fetched must be entirely contained within the corresponding vertex buffer binding, as described in Vertex Input Description (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdDraw-None-02721) (code 0x00000000 from Validation at Buffer:99ef63bb)

The issue appears to be that the pipeline is created with some static vertex input state data that will not be used, because VK_DYNAMIC_STATE_VERTEX_INPUT_EXT is set, and the actual state is set later with a call to vkCmdSetVertexInputEXT. The static data contains erroneous attribute offsets, formats and strides (as seen above) that would result in invalid alignment. However, the dynamic state contains the right values that, when combined with the vertex buffer offset (1 in the error message above, which is correct) results in aligned data.

If I set pVertexInputState to nullptr when creating the dynamic state pipeline, no error is printed. I've applied this workaround in the CTS tests, but it can be disabled with a 1-line change. Let me know if you need more info to reproduce.

**Expected behavior**

No validation error reported.

**Valid Usage ID**

VUID-vkCmdDraw-None-02721

[...]

PR #7553, opened 2024-02-22
layers: Fix Dynamic Vertex Input

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5954

This is a re-working of the Vertex Input to better handle Dynamic vertex input

There are 2 dynamic input

1. Have everything dynamic via `VK_DYNAMIC_STATE_VERTEX_INPUT_EXT` / `vkCmdSetVertexInputEXT`
2. Just make the `stride` dynamic with `VK_DYNAMIC_STATE_VERTEX_INPUT_BINDING_STRIDE` / `vkCmdBindVertexBuffers2`


layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -2888,89 +2891,104 @@ bool CoreChecks::ValidatePipelineDrawtimeState(const LastBound &last_bound_state
...
-            const auto vertex_binding = pipeline->vertex_input_state->binding_descriptions[i].binding;
-            if (current_vtx_bfr_binding_info.size() < (vertex_binding + 1)) {
-                const LogObjectList objlist(cb_state.Handle(), pipeline->Handle());
-                skip |= LogError(vuid.vertex_binding_04007, objlist, loc,
-                                 "%s expects that this Command Buffer's vertex binding Index %u should be set via "
-                                 "vkCmdBindVertexBuffers. This is because pVertexBindingDescriptions[%zu].binding value is %u.",
-                                 FormatHandle(*pipeline).c_str(), vertex_binding, i, vertex_binding);
-            } else if ((current_vtx_bfr_binding_info[vertex_binding].buffer_state == nullptr) && !enabled_features.nullDescriptor) {
+    if (pipeline && pipeline->vertex_input_state) {
+        for (const auto &vertex_buffer_binding : cb_state.current_vertex_buffer_binding_info) {
```
spencer-lunarg, line 2910: the "big" change is that everything now uses `current_vertex_buffer_binding_info` which updated from the pipeline and dynamic state, so it always has the "correct" up-to-date info at draw time

layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -2888,89 +2891,104 @@ bool CoreChecks::ValidatePipelineDrawtimeState(const LastBound &last_bound_state
...
+            } else if (!Get<vvl::Buffer>(vertex_buffer_binding.second.buffer)) {
+                const LogObjectList objlist(cb_state.Handle(), pipeline->Handle());
+                skip |=
+                    LogError(vuid.vertex_binding_04007, objlist, loc,
+                             "Vertex binding %" PRIu32 " is not a valid VkBuffer. (Check the buffer set in vkCmdBindVertexBuffers)",
+                             vertex_buffer_binding.first);
             }
         }
 
+        const auto &vertex_attribute_descriptions = pipeline->IsDynamic(VK_DYNAMIC_STATE_VERTEX_INPUT_EXT)
```
spencer-lunarg, line 2926: More examples of using dynamic state to pick the correct things to loop over

layers/state_tracker/cmd_buffer_state.h
```diff
@@ -242,7 +242,7 @@ class CommandBuffer : public RefcountedStateObject {
         std::vector<VkColorComponentFlags> color_write_masks;        // VK_DYNAMIC_STATE_COLOR_WRITE_MASK_EXT
 
         // VK_DYNAMIC_STATE_VERTEX_INPUT_EXT
-        std::vector<VkVertexInputBindingDescription2EXT> vertex_binding_descriptions;
+        std::vector<uint32_t> vertex_binding_descriptions_divisor;
```
spencer-lunarg, line 245: only the `uint32_t` is checked

layers/state_tracker/vertex_index_buffer_state.h
```diff
@@ -20,41 +20,32 @@
...
 
 #include "vulkan/vulkan.h"
-#include <vector>
-#include <memory>
 
 namespace vvl {
 class Buffer;
-}  // namespace vvl
 
-struct BufferBinding {
```
spencer-lunarg, line 30: this was trying to be over clever and just made things more complex and wrong

layers/state_tracker/vertex_index_buffer_state.cpp
```diff
@@ -1,36 +0,0 @@
...
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
-#include "state_tracker/vertex_index_buffer_state.h"
-#include "state_tracker/buffer_state.h"
-
-BufferBinding::BufferBinding(const std::shared_ptr<vvl::Buffer> &buffer_state_, VkDeviceSize size_, VkDeviceSize offset_, VkDeviceSize stride_)
```
spencer-lunarg, line 24: this whole file was because we didn't want to include `buffer_state.h` in the header, but because of `nullDescriptor` we now only store `VkBuffer` so really was no need for this now

layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -177,32 +176,37 @@ bool CoreChecks::ValidateGraphicsPipelinePortability(const vvl::Pipeline &pipeli
...
-                        "(portability error): attribute.offset (%" PRIu32
-                        ") + "
-                        "sizeof(vertex_description.format) (%" PRIu32 ") is larger than the vertex stride (%" PRIu32 ").",
-                        attrib.offset, vkuFormatElementSize(attrib.format), desc.stride);
+    if (!pipeline.IsDynamic(VK_DYNAMIC_STATE_VERTEX_INPUT_EXT) &&
+        !pipeline.IsDynamic(VK_DYNAMIC_STATE_VERTEX_INPUT_BINDING_STRIDE)) {
+        // Validate vertex inputs
+        for (const auto &desc : pipeline.vertex_input_state->binding_descriptions) {
+            const uint32_t min_alignment = phys_dev_ext_props.portability_props.minVertexInputBindingStrideAlignment;
+            if ((desc.stride < min_alignment) || (min_alignment == 0) || ((desc.stride % min_alignment) != 0)) {
```
artem-lunarg, line 184: not related to this PR; for documentation purposes:
`desc.stride < min_alignment` is not needed - `(desc.stride % min_alignment) != 0` covers this well.
Also `min_alignment == 0` is not possible according to spec but lets assume it's bad driver/mock. Then suggestion for final expression is:
` if (min_alignment != 0 && desc.stride % min_alignment != 0) `

layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -2888,89 +2891,106 @@ bool CoreChecks::ValidatePipelineDrawtimeState(const LastBound &last_bound_state
     // Verify vertex & index buffer for unprotected command buffer.
     // Because vertex & index buffer is read only, it doesn't need to care protected command buffer case.
     if (enabled_features.protectedMemory == VK_TRUE) {
-        for (const auto &buffer_binding : current_vtx_bfr_binding_info) {
-            if (buffer_binding.buffer_state && !buffer_binding.buffer_state->Destroyed()) {
-                skip |= ValidateProtectedBuffer(cb_state, *buffer_binding.buffer_state, loc, vuid.unprotected_command_buffer_02707,
+        for (const auto &vertex_buffer_binding : cb_state.current_vertex_buffer_binding_info) {
+            const auto buffer_state = Get<vvl::Buffer>(vertex_buffer_binding.second.buffer);
+            if (buffer_state && !buffer_state->Destroyed()) {
```
artem-lunarg, line 2896: @spencer-lunarg what's the intuition behind additional check for `Destroyed()` that we do in other places too?

spencer-lunarg: Let me write a test, I think the idea is we destroy a `VkBuffer` before it is used the `vvl::Buffer` is valid... but then if that is the case, we should have a better system then hoping people do this everywhere.... so let me write a test or two and get back to you

artem-lunarg (approved):
LGTM. Some comments not directly related to core part of PR. I have limited understanding of this part of code but this looks like a simpler and more unified approach to handle dynamic VI.

layers/core_checks/cc_pipeline_graphics.cpp
spencer-lunarg: ok, this was a nice catch, so these `Destroy` are not needed as this caught by `VUID-vkCmdDraw-commandBuffer-recording` (have 2 new tests)

I realize the issue was before we had to store a `std::shared_ptr<vvl::Buffer>` which could have went stale, but now we hold a `VkBuffer` and check at draw time, meaning it will have been caught already

I removed these `Destroyed` from this PR

artem-lunarg: Thanks for checking this. Good information.

spencer-lunarg:
Internal Failures are for unrelated GPU-AV tests

PR #7556, opened 2024-02-22
tests: Remove duplicate handle functions

spencer-lunarg:
removes

```c++
VkDevice device() { return handle(); }
```
and
```c++-
VkImage image() const { return handle(); }
```

so that there is only `handle()` and not a duplicated way to grab it... tries to make things all consistent


artem-lunarg (approved):
image

PR #7558, opened 2024-02-22
build: Replace robin-hood-hashing with unordered_dense

arno-lunarg:
Follow up to https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7473

Code is not compiling when using `ankerl::unordered_dense::map`, but does when using `ankerl::unordered_dense::segmented_map`, [a version that better mitigates memory allocation made when adding new elements](https://github.com/martinus/unordered_dense/tree/main?tab=readme-ov-file#4-segmented_map-and-segmented_set). This is also a goal, next step is to measure if it indeed improves our memory footprint when initializing static maps.


layers/state_tracker/image_state.cpp
```diff
@@ -667,7 +667,7 @@ std::vector<VkPresentModeKHR> Surface::GetPresentModes(VkPhysicalDevice phys_dev
     assert(phys_dev);
     std::vector<VkPresentModeKHR> result;
     if (auto search = present_modes_data_.find(phys_dev); search != present_modes_data_.end()) {
-        for (auto mode = search->second.begin(); mode != search->second.end(); mode++) {
+        for (auto mode = search->second.begin(); mode != search->second.end(); ++mode) {
```
spencer-lunarg, line 670: @jeremyg-lunarg so this is the type of issue failing the build before :laughing:

tests/unit/shader_interface.cpp
```diff
@@ -540,7 +540,7 @@ TEST_F(NegativeShaderInterface, VsFsTypeMismatchBlockStruct) {
     CreatePipelineHelper::OneshotTest(*this, set_info, kErrorBit, "VUID-RuntimeSpirv-OpEntryPoint-07754");
 }
 
-TEST_F(NegativeShaderInterface, VsFsTypeMismatchBlockStruct64bit) {
+TEST_F(NegativeShaderInterface, DISABLED_VsFsTypeMismatchBlockStruct64bit) {
```
jeremyg-lunarg, line 543: Why does this test need to be disabled?

layers/CMakeLists.txt
```diff
@@ -80,11 +80,11 @@ target_link_libraries(VkLayer_utils PUBLIC
 target_include_directories(VkLayer_utils SYSTEM PRIVATE external)
 target_include_directories(VkLayer_utils PUBLIC . ${API_TYPE})
 
-find_package(robin_hood CONFIG)
-option(USE_ROBIN_HOOD_HASHING "robin_hood provides faster versions of std::unordered_map and std::unordered_set" ${robin_hood_FOUND})
-if (USE_ROBIN_HOOD_HASHING)
-    target_link_libraries(VkLayer_utils PUBLIC robin_hood::robin_hood)
-    target_compile_definitions(VkLayer_utils PUBLIC USE_ROBIN_HOOD_HASHING)
+find_package(unordered_dense CONFIG)
+option(USE_UNORDERED_DENSE "unordered_denser provides faster versions of std::unordered_map and std::unordered_set" ${unordered_dense_FOUND})
```
jeremyg-lunarg, line 84: typo: unordered_denser

scripts/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 # ~~~
-# Copyright (c) 2023 LunarG, Inc.
+# Copyright (c) 2024 LunarG, Inc.
```
jeremyg-lunarg, line 2: 2023-2204

layers/CMakeLists.txt
```diff
@@ -80,11 +80,11 @@ target_link_libraries(VkLayer_utils PUBLIC
 target_include_directories(VkLayer_utils SYSTEM PRIVATE external)
 target_include_directories(VkLayer_utils PUBLIC . ${API_TYPE})
 
-find_package(robin_hood CONFIG)
-option(USE_ROBIN_HOOD_HASHING "robin_hood provides faster versions of std::unordered_map and std::unordered_set" ${robin_hood_FOUND})
-if (USE_ROBIN_HOOD_HASHING)
-    target_link_libraries(VkLayer_utils PUBLIC robin_hood::robin_hood)
-    target_compile_definitions(VkLayer_utils PUBLIC USE_ROBIN_HOOD_HASHING)
+find_package(unordered_dense CONFIG)
```
jeremyg-lunarg, line 83: Switching from USE_ROBIN_...* to USE_UNORDERED_* is going to require changes in Internal CI and the SDK build stuff.  I wonder if we should do something like USE_CUSTOM_HASHMAP instead in case we ever decide to change implementations again.  Curious what others think about this...

tests/unit/shader_interface.cpp
spencer-lunarg: this can be re-enabled now, it was an issue with how this test was reporting different VU depending on how the hash map ordered the Locations in the interfacing logic

jeremyg-lunarg (changes requested):
I think this PR shouldn't be merged until after the upcoming SDK branches.

spencer-lunarg:
> I think this PR shouldn't be merged until after the upcoming SDK branches.

agree, but still worth getting it working and ready still

layers/CMakeLists.txt
arno-lunarg: Good idea I think. I only used this name in replacement of the C++ define, explicit mentions of unordered dense are still there in files like vvl.yml

scripts/known_good.json
```diff
@@ -46,15 +46,12 @@
...
-            "cmake_options": [
-                "-DRH_STANDALONE_PROJECT=OFF"
-            ],
-            "commit": "3.11.5"
+            "name": "unordered_dense",
+            "url": "https://github.com/martinus/unordered_dense.git",
+            "sub_dir": "unordered_dense",
+            "build_dir": "unordered_dense/build",
+            "install_dir": "unordered_dense/build/install",
+            "commit": "v4.0.4"
```
spencer-lunarg, line 54: fyi @artem-lunarg @jeremyg-lunarg 

https://github.com/martinus/unordered_dense/issues/111#issuecomment-1992635108

Looks like our compiling error **might** be something wrong with the latest unordered_dense code (version 4.4.0)

arno-lunarg: Went back to 4.0.4

arno-lunarg: well not so fast boy

spencer-lunarg: Then I will run through the performance regression suite and see what difference we get

filnet:
Would it make sense to error out (or emit some warning) if `USE_ROBIN_HOOD_HASHING` is set ?
If not some projects that build VVL might not realize they silently switched to the default map.

For example: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-vulkan-validation-layers/PKGBUILD#L68

spencer-lunarg:
> Would it make sense to error out (or emit some warning) if USE_ROBIN_HOOD_HASHING is set ?

I agree, @arno-lunarg just add

```cmake
if (USE_ROBIN_HOOD_HASHING)
  message(WARNING "USE_ROBIN_HOOD_HASHING was removed, please use USE_CUSTOM_HASH_MAP now .")
endif ()
```

filnet:
nit: there is an extraneous space before the ending colon.

> ```cmake
> if (USE_ROBIN_HOOD_HASHING)
>   message(WARNING "USE_ROBIN_HOOD_HASHING was removed, please use USE_CUSTOM_HASH_MAP now .")
> endif ()
> ```

spencer-lunarg:
closing as this has got stale - the internal testing didn't show much or any real perf increase

With various Robin Hood hashing bugs we know about (https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8501 and https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8921) , this might be desirable in the future, but can create a new PR then

Issue #7547, opened 2024-02-21
Fix how versions are handled in the test suite


spencer-lunarg:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/6db385a7ea2c85d384df3e9ccff06729f71b22e9#diff-eefbaf049195ad7b10928b277b434137d20dbbd7654d1d6bc278f689b8a5a612R318

changed how the tests handled versions, previously one could write a test case that requires descriptor indexing to work with either core or extension as follows:
```
SetTargetApiVersion(VK_API_VERSION_1_3);
AddOptionalExtensions(VK_EXT_DESCRIPTOR_INDEXING_EXTENSION_NAME);
AddRequiredFeature(vkt::Feature::descriptorIndexing);
```

but now we have turned the `SetTargetApiVersion` into "require at least this version" syntax instead of using the old `DeviceValidationVersion` API

This will likely reflect on the test coverage the `min_core.json` profile that is being skipped

Need to have a better system to allow a larger test coverage with regards to versions

aqnuep:
So the story here is that historically `SetTargetApiVersion()` only enforced the instance API version, but did not require the physical device to support the target API version. That's why the `DeviceValidationVersion()` function exists that actually verifies the physical device API version.

However, https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/6db385a7ea2c85d384df3e9ccff06729f71b22e9#diff-eefbaf049195ad7b10928b277b434137d20dbbd7654d1d6bc278f689b8a5a612R318, likely inadvertently changed this behavior to actually require the physical device to support the target API version, otherwise the test case is skipped.

This is an issue, because test cases have been written with the assumption that the test case would still continue to execute even if the physical device does not support the target API version (unless the test case explicitly guards against older versions using `DeviceValidationVersion()`), so this change practically made a lot of validation test cases to no longer execute when ran against drivers supporting order versions of Vulkan.

Furthermore, the way `SetTargetApiVersion()` originally worked was very useful, because you could actually write test cases in a fashion that ran with promoted extensions if either the core API version where the extension was promoted to or the extension was present, e.g. by writing the following:

````C++
SetTargetApiVersion(promoted_to_version);
AddOptionalExtensions(extension_name);
// + additional checks for physical device features, if applicable, e.g. using the new AddRequiedFeature function
````

Unfortunately, this is no longer possible, and the change also made the `DeviceValidationVersion()` function completely useless.

[...]

PR #7560, opened 2024-02-22
tests: Revert SetTargetApiVersion behavior and add SetRequiredApiVersion

aqnuep:
A possible resolution to issue #7547.

Please note that it is not necessarily the mechanism, or name I'm worried about.

It's just that it would be really great if applicable test cases could run if either the API version or extension is available, because that option seems to be eliminated since the behavior of `SetTargetApiVersion()` was changed.

### DIFF
--- a/docs/creating_tests.md
+++ b/docs/creating_tests.md
@@ -117,6 +117,14 @@ if (DeviceValidationVersion() >= VK_API_VERSION_1_1) {
 }
 ```
 
+Alternatively, to require the device to support the target Vulkan version, the test writer can call:
+
+```cpp
+SetRequiredApiVersion(VK_API_VERSION_1_1);
+// Will skip if the supported instance version or device supported version is not high enough
+RETURN_IF_SKIP(InitFramework());
+```
+
 ### Promoted extensions
 
 The test framework now automatically handles extensions promoted to core versions by not enabling the extensions if the target instance Vulkan version (set through `SetTargetApiVersion`) or the Vulkan version supported by the device (queriable using `DeviceValidationVersion`) already includes the instance or device extension's functionality, respectively.
--- a/tests/device_profiles/max_core.json
+++ b/tests/device_profiles/max_core.json
@@ -649,6 +649,28 @@
                     "uniformTexelBufferOffsetSingleTexelAlignment": true,
                     "maxBufferSize": 4294967000
                 },
+                "VkPhysicalDeviceSubgroupProperties": {
+                    "subgroupSize": 32,
+                    "supportedStages": [
+                        "VK_SHADER_STAGE_VERTEX_BIT",
+                        "VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT",
+                        "VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT",
+                        "VK_SHADER_STAGE_GEOMETRY_BIT",
+                        "VK_SHADER_STAGE_FRAGMENT_BIT",
+                        "VK_SHADER_STAGE_COMPUTE_BIT"
+                    ],
+                    "supportedOperations": [
+                        "VK_SUBGROUP_FEATURE_BASIC_BIT",
+                        "VK_SUBGROUP_FEATURE_VOTE_BIT",
+                        "VK_SUBGROUP_FEATURE_ARITHMETIC_BIT",
+                        "VK_SUBGROUP_FEATURE_BALLOT_BIT",
+                        "VK_SUBGROUP_FEATURE_SHUFFLE_BIT",
+                        "VK_SUBGROUP_FEATURE_SHUFFLE_RELATIVE_BIT",
+                        "VK_SUBGROUP_FEATURE_CLUSTERED_BIT",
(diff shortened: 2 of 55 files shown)


lunarpapillo:
LunarG CI failed for two reasons:
- crashed on Android device
- `PositiveGpuAV.DrawingWithUnboundUnusedSet` on **Linux-NVIDIA** was broken, causing failures on all PRs

The latter problem has been fixed, by removing the test from the test list.  I'm not restarting because the former issue (Android crash) still needs to be resolved.

aqnuep:
> LunarG CI failed for two reasons:
> 
> * crashed on Android device
> * `PositiveGpuAV.DrawingWithUnboundUnusedSet` on **Linux-NVIDIA** was broken, causing failures on all PRs
> 
> The latter problem has been fixed, by removing the test from the test list. I'm not restarting because the former issue (Android crash) still needs to be resolved.

Thank you for the information. I still have to think about the faith of this PR, because some of the testing uncovered some conceptual issues around the new `AddRequiredFeature` tool which I believe @spencer-lunarg added.

So the tricky part regarding allowing both core and extension versions of functionalities is that some extensions have additional feature caps in core, like descriptor indexing. The issue around `AddRequiredFeature` is that if, for example, I add `AddRequiredFeature(vkt::Feature::descriptorIndexing)`, then that obviously is passed down in `VkPhysicalDeviceVulkan12Features`, because it only exists there, but if the test case adds other descriptor indexing feature requirements using `AddRequiredFeature`, then it passes it down in `VkPhysicalDeviceDescriptorIndexingFeaturesEXT`, which will then result in a validation error due to the conflicting use of feature structures with overlapping feature flags.

I don't think that could be resolved in any easy way, and thus may be a fundamental limitation of `AddRequiredFeature` that may just cause troubles if we re-enable this mechanism of allowing test cases to run against both core-only and extension-only versions of a feature. @spencer-lunarg, do you have any thoughts on this?

spencer-lunarg:
@aqnuep so `AddRequiredFeature` was added by Arno as for Ray Tracing tests we have to enable a LOT of features for a single tests, but regardless I was the one who liked it and put it everywhere because it made the common test writing faster/easier

I knew it had some limitation with things like `VkPhysicalDeviceVulkan12Features` but for the few tests I ran into where it mattered I did it the "old fashion" way

A possible way maybe be to adjust the script that generates the `vkt::features` list and anything that would be problematic we don't allow, but that might lead to being confusing to someone who would assume it is just there

To be honest, people writing tests will not think about the way to make sure a feature/version/extension combo runs everywhere if it running on at least one device, I assume they run the test, see it fails because they forgot a feature and just add it anyway possible... thinking out loud, it might be better to keep the `AddRequiredFeature` interface and "make it smart" under the hood

I guess I'm still in the opinion that if this is an issue with certain extensions  like descriptor indexing, they should be mostly isolated to a single file to make it easier to contain the issue inside there

aqnuep:
> @aqnuep so AddRequiredFeature was added by Arno as for Ray Tracing tests we have to enable a LOT of features for a single tests, but regardless I was the one who liked it and put it everywhere because it made the common test writing faster/easier

I definitely agree it's a great addition. No question about that.

> To be honest, people writing tests will not think about the way to make sure a feature/version/extension combo runs everywhere if it running on at least one device, I assume they run the test, see it fails because they forgot a feature and just add it anyway possible... thinking out loud, it might be better to keep the AddRequiredFeature interface and "make it smart" under the hood

Yes, I thought about that too, in general about the whole topic of core vs extension context, as technically the promoted extension handling infrastructure I've added already does sort of do one side of this: not requiring an extension when we know the device supports a core API version that already includes the functionality.

That's why I'm second-guessing my original intention to revert the behavior of `SetTargetApiVersion` instead of doing something about that problem at the framework level.

My main concern beyond just the general issue of the lost test coverage on certain devices and profiles is that I expect that more and more test cases will end up being written with 1.3 set as the minimum API version, even when it is not really needed, which then would result in those test cases being skipped on Vulkan SC 1.0, which is based off of Vulkan 1.2.

If there is no rush to get a resolution on this, I'd like to think about this a bit more and see if we can do something better.

spencer-lunarg:
> If there is no rush to get a resolution on this, I'd like to think about this a bit more and see if we can do something better.

Sounds good... If you want, I am still happy to go through all the tests that have `SetTargetApiVersion` and remove every instance that simply works without that line... It would probably help reduce noise later when trying to find which tests are actually missing being ran

aqnuep:
I'm closing this PR as we won't need or can utilize this as is.

If we want to get to a point where we can support either core or extension versions of features as test case dependencies then we'd have to design something around the new `AddRequiredFeature` infrastructure, which will likely be superior in the end anyway.

FWIW, making `SetTargetApiVersion` mean actual required device API version did not have any relevant downstream impact on the SC VVL, so we can keep things as is and we also won't need your alternative PR for this, @spencer-lunarg.

Let's keep things as is for now and maybe think about a solution that in the future will lead us to not even needing to enable extensions or specific API versions when we only looking for support for particular features (through `AddRequiredFeature`) by implicitly enabling the extension or API version that introduces it, which should be doable.

PR #7569, opened 2024-02-23
tests: Fix sync test to use sparse binding queue

spencer-lunarg:
`PositiveSyncObject.ExternalSemaphore` was not using the sparse binding queue (was failing on RADV for Mesa 24)

### DIFF
--- a/tests/unit/sync_object_positive.cpp
+++ b/tests/unit/sync_object_positive.cpp
@@ -1168,9 +1168,9 @@ TEST_F(PositiveSyncObject, ExternalSemaphore) {
     RETURN_IF_SKIP(Init());
 
     // Check for external semaphore import and export capability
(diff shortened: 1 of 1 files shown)


tests/unit/sync_object_positive.cpp
```diff
@@ -1166,11 +1166,14 @@ TEST_F(PositiveSyncObject, ExternalSemaphore) {
     AddRequiredExtensions(VK_KHR_EXTERNAL_SEMAPHORE_EXTENSION_NAME);
 
     RETURN_IF_SKIP(Init());
+    if (m_device->sparse_queues().empty()) {
```
artem-lunarg, line 1169: The suggestion is not to skip the entire test if no sparse queues available, but to add this check to `if (m_device->phy().features().sparseBinding)`.

spencer-lunarg: @artem-lunarg good catch, fixed

artem-lunarg (approved):

PR #7576, opened 2024-02-24
Fix external_memory_sync_positive.cpp on metal platforms

OscarJPrice:
Building on Mac does not succeed:
```c++
error: called object type 'VkDevice' (aka 'VkDevice_T *') is not a function or function pointer
    const VkDevice device = device();
```
Simple change on line 353 of external_memory_sync_positive.cpp from `const VkDevice device = device();` to `const VkDevice device = this->device();` to avoid name clashing error.

### DIFF
--- a/tests/unit/external_memory_sync_positive.cpp
+++ b/tests/unit/external_memory_sync_positive.cpp
@@ -350,7 +350,7 @@ TEST_F(PositiveExternalMemorySync, ExportMetalObjects) {
         RETURN_IF_SKIP(InitState(nullptr, &features2));
     }
 
-    const VkDevice device = device();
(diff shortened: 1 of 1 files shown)


spencer-lunarg:
@OscarJPrice thanks for the fix

1. To make CI happy, please make the commit start with `tests:` and limit it to 64 characters in length (`tests: Fix building on Metal` works for me)

2. Sorry this occurred, we had a MacOS build in CI, but had to temporarily remove it because the 10x CI cost from GitHub was pushing us over the current limit we had (and are actively working to get that back into the normal CI)

OscarJPrice:
Deleted the branch because I'm an idiot

PR #7588, opened 2024-02-26
Add validation for VK_EXT_map_memory_placed

gfxstrand:
The one VU I haven't validated is the one about `pPlacedAddress` not being inside the map of another Vulkan memory object. I don't know if that kind of tracking exists inside the validation layers or how to go about adding it if it doesn't. Everything else should be validated and tested.

### DIFF
--- a/layers/core_checks/cc_device_memory.cpp
+++ b/layers/core_checks/cc_device_memory.cpp
@@ -1096,6 +1096,11 @@ bool CoreChecks::PreCallValidateMapMemory(VkDevice device, VkDeviceMemory memory
     if (mem_info) {
         skip |= ValidateMapMemory(*mem_info.get(), offset, size, error_obj.location.dot(Field::offset),
                                   error_obj.location.dot(Field::size));
+
+        if (flags & VK_MEMORY_MAP_PLACED_BIT_EXT) {
+            skip |= LogError("VUID-vkMapMemory-flags-09568", memory, error_obj.location.dot(Field::flags),
+                             "VK_MEMORY_MAP_PLACED_BIT_EXT is not allowed in vkMapMemory()");
+        }
     }
     return skip;
 }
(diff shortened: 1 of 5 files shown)


layers/core_checks/cc_device_memory.cpp
```diff
@@ -1096,6 +1096,11 @@ bool CoreChecks::PreCallValidateMapMemory(VkDevice device, VkDeviceMemory memory
     if (mem_info) {
         skip |= ValidateMapMemory(*mem_info.get(), offset, size, error_obj.location.dot(Field::offset),
                                   error_obj.location.dot(Field::size));
+
+        if (flags & VK_MEMORY_MAP_PLACED_BIT_EXT) {
+            skip |= LogError("VUID-vkMapMemory-flags-09568", memory, error_obj.location,
```
spencer-lunarg, line 1101:
```suggestion
            skip |= LogError("VUID-vkMapMemory-flags-09568", memory, error_obj.location.dot(Field::flags),
```

layers/core_checks/cc_device_memory.cpp
```diff
@@ -1105,9 +1110,73 @@ bool CoreChecks::PreCallValidateMapMemory2KHR(VkDevice device, const VkMemoryMap
...
+
+        if (pMemoryMapInfo->flags & VK_MEMORY_MAP_PLACED_BIT_EXT) {
+            if (!enabled_features.memoryMapPlaced) {
+                skip |= LogError("VUID-VkMemoryMapInfoKHR-flags-09569", pMemoryMapInfo->memory, error_obj.location,
+                                 "memoryMapPlaced is not enabled");
+            }
+
+            if (enabled_features.memoryMapRangePlaced) {
+                if (pMemoryMapInfo->offset % phys_dev_ext_props.map_memory_placed_props.minPlacedMemoryMapAlignment != 0) {
+                    skip |= LogError("VUID-VkMemoryMapInfoKHR-flags-09573", pMemoryMapInfo->memory, info_loc.dot(Field::offset),
```
spencer-lunarg, line 1125: `info_loc.dot(Field::offset)` will print out before the error message

> vkMapMemory2KHR() pMemoryMapInfo->offset 

so the error message will strange, 

suggest

"offset (0x%" PRIx64 ") is not an integer multiple of minPlacedMemoryMapAlignment (0x%" PRIx64 ") while VK_MEMORY_MAP_PLACED_BIT_EXT is set",

spencer-lunarg: same for the rest of the error messages

tests/unit/memory.cpp
```diff
@@ -323,6 +327,111 @@ TEST_F(NegativeMemory, MapMemory2WithoutHostVisibleBit) {
     m_errorMonitor->VerifyFound();
 }
 
+#ifndef VK_USE_PLATFORM_WIN32_KHR
+TEST_F(NegativeMemory, MapMemoryPlaced) {
```
spencer-lunarg, line 331: for extensions, we want a single positive test (`memory_positive.cpp`) of using the extension in a basic way to help reduce false positive regressions

tests/unit/memory.cpp
```diff
@@ -323,6 +327,111 @@ TEST_F(NegativeMemory, MapMemory2WithoutHostVisibleBit) {
...
+    TEST_DESCRIPTION("Attempt to map placed memory in a number of incorrect ways");
+    AddRequiredExtensions(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
+    AddRequiredExtensions(VK_KHR_MAP_MEMORY_2_EXTENSION_NAME);
+    AddRequiredExtensions(VK_EXT_MAP_MEMORY_PLACED_EXTENSION_NAME);
+    RETURN_IF_SKIP(InitFramework());
```
spencer-lunarg, line 336: you can just go

```patch
- RETURN_IF_SKIP(InitFramework());
- 
- VkPhysicalDeviceMapMemoryPlacedFeaturesEXT map_placed_features = vku::InitStructHelper();
- GetPhysicalDeviceFeatures2(map_placed_features);
- if (map_placed_features.memoryMapPlaced != VK_TRUE) {
-     GTEST_SKIP() << "memoryMapPlaced feature not supported";
- }
- 
- RETURN_IF_SKIP(InitState(nullptr, &map_placed_features));
+ AddRequiredFeature(vkt::Feature::memoryMapPlaced);
+ RETURN_IF_SKIP(Init());
```

layers/core_checks/cc_device_memory.cpp
```diff
@@ -1105,9 +1110,73 @@ bool CoreChecks::PreCallValidateMapMemory2KHR(VkDevice device, const VkMemoryMap
...
+            const auto placed_info = vku::FindStructInPNextChain<VkMemoryMapPlacedInfoEXT>(pMemoryMapInfo->pNext);
+            const auto addr_loc = info_loc.pNext(Struct::VkMemoryMapPlacedInfoEXT, Field::pPlacedAddress);
+            if (placed_info == NULL) {
+                skip |= LogError("VUID-VkMemoryMapInfoKHR-flags-09570", pMemoryMapInfo->memory, addr_loc,
+                                 "VK_MEMORY_MAP_PLACED_BIT_EXT was set but the pNext chain does not contain "
+                                 "VkMemoryMapPlacedInfoEXT");
+            } else if (placed_info->pPlacedAddress == NULL) {
+                skip |= LogError("VUID-VkMemoryMapInfoKHR-flags-09570", pMemoryMapInfo->memory, addr_loc,
+                                 "VK_MEMORY_MAP_PLACED_BIT_EXT was set but VkMemoryMapPlacedInfoEXT::pPlacedAddress is NULL");
+            } else if (reinterpret_cast<std::uintptr_t>(placed_info->pPlacedAddress) %
```
spencer-lunarg, line 1171: seems android/ios build aren't happy with this cast

spencer-lunarg (changes requested):
LGTM - Just a few nit fixes

layers/core_checks/cc_device_memory.cpp
```diff
@@ -1105,9 +1110,73 @@ bool CoreChecks::PreCallValidateMapMemory2KHR(VkDevice device, const VkMemoryMap
...
+                                     "VK_MEMORY_MAP_PLACED_BIT_EXT is set and offset (0x%" PRIx64 ") is not an integer "
+                                     "multiple of minPlacedMemoryMapAlignment (0x%" PRIx64 ")",
+                                     pMemoryMapInfo->offset,
+                                     phys_dev_ext_props.map_memory_placed_props.minPlacedMemoryMapAlignment);
+                }
+
+                if (pMemoryMapInfo->size != VK_WHOLE_SIZE &&
+                    pMemoryMapInfo->size % phys_dev_ext_props.map_memory_placed_props.minPlacedMemoryMapAlignment != 0) {
+                    skip |= LogError("VUID-VkMemoryMapInfoKHR-flags-09574", pMemoryMapInfo->memory, info_loc.dot(Field::size),
+                                     "VK_MEMORY_MAP_PLACED_BIT_EXT is set and size (0x%" PRIx64 ") is not VK_WHOLE_SIZE and "
```
spencer-lunarg, line 1135: another nit/question, people have complained that things like `size` and `offset` were in hex instead of decimal, as one of the main users of this extension, happy to follow your guide on this, but wanted to express this incase this was an oversight

gfxstrand: I think it's a bit of a matter of user preference. For me, when I'm looking at something that's effectively an address, especially when the validation error is that the address isn't properly aligned, having it in decimal is next to useless while hex makes the error easy to see.

layers/core_checks/cc_device_memory.cpp
gfxstrand: Fixed. It's not compatible with `PRIx64` on 32-bit builds. I've replaced it with `%p`.

layers/core_checks/cc_device_memory.cpp
spencer-lunarg: sounds good

layers/core_checks/cc_device_memory.cpp
gfxstrand: Fixed.

tests/unit/memory.cpp
gfxstrand: Done.

spencer-lunarg (approved):

PR #7520, opened 2024-02-16
tests: Add ray query gpu av tests

arno-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7500

PR #7596, opened 2024-02-27
gpu: Add OpRayQueryInitializeKHR validation

spencer-lunarg:
This adds some (will add more after we confirm this flow works) SPIR-V instrumentation to validate the values of `OpRayQueryInitializeKHR` instructions

Positive versions of tests are in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7520


tests/CMakeLists.txt
```diff
@@ -143,8 +144,6 @@ target_sources(vk_layer_validation_tests PRIVATE
     unit/query_positive.cpp
     unit/ray_tracing.cpp
     unit/ray_tracing_nv.cpp
-    unit/ray_tracing_gpu_nv.cpp
```
spencer-lunarg, line 146: this was removed in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7595 already, will be rebased

layers/gpu_validation/spirv/ray_query_pass.h
```diff
@@ -0,0 +1,44 @@
...
+
+// Create a pass to instrument SPV_KHR_ray_query instructions
+class RayQueryPass : public Pass {
+  public:
+    RayQueryPass(Module& module) : Pass(module) {}
+
+  private:
+    bool AnalyzeInstruction(const Function& function, const Instruction& inst) override;
+    uint32_t CreateFunctionCall(BasicBlock& block) override;
+    void Reset() override;
```
arno-lunarg, line 35: Nit: could use `final` instead of `override`

tests/unit/gpu_av_ray_query.cpp
```diff
@@ -0,0 +1,249 @@
...
+        layout(set = 0, binding = 1) uniform Uniforms {
+          float t_min;
+          float t_max;
+        } trace_rays_params;
+
+        void main() {
+            rayQueryEXT query;
+            rayQueryInitializeEXT(query, tlas, gl_RayFlagsTerminateOnFirstHitEXT, 0xff, vec3(0), 
+              trace_rays_params.t_min, vec3(0,0,1), trace_rays_params.t_max);
+            rayQueryProceedEXT(query);
```
arno-lunarg, line 43: So as discussed we need to make sure that passing down a null query to the driver is ok

layers/gpu_validation/gpu_error_message.cpp
```diff
@@ -436,6 +436,60 @@ bool gpuav::Validator::GenerateValidationMessage(const uint32_t *debug_record, c
                     assert(false);
             }
         } break;
+        case kInstErrorRayQueryNegativeMin: {
```
arno-lunarg, line 439: I wanted to get rid of this `GenerationValidationMessage`, now even more so. Same for `gpu_shader_constants.h`, keeping adding new constants like so is not sustainable.
But it is ok for this PR!

arno-lunarg (approved):
I will not buy any more french pastries unless you add the missing new lines at the end of files

More seriously, have you tried running this against an application using ray query? Maybe the ray query sample for instance

layers/gpu_validation/gpu_error_message.cpp
spencer-lunarg: ya, 100% agree, was going to be something I asked you tomorrow lol

tests/unit/gpu_av_ray_query.cpp
spencer-lunarg: I still need to double check the spec (also plan to run against the samples)

jeremyg-lunarg:
> I will not buy any more french pastries unless you add the missing new lines at the end of files

BRB, writing a PR to add all the missing EOF new lines.

layers/gpu_shaders/inst_ray_query.comp
```diff
@@ -0,0 +1,121 @@
...
+        }
+        if (opaque_mask != 0 && ((opaque_mask & (opaque_mask - 1)) != 0)) {
+            error = kInstErrorRayQueryOpaque;
+            param_0 = ray_flags;
+            break;
+        }
+    } while (false);
+
+    if (0u != error) {
+        uint rec_len = 9u;
```
jeremyg-lunarg, line 103: does the GLSL preprocessor support function-like macros? I know that making this a function has been problematic but it is slightly disturbing to have it inline everywhere.

layers/gpu_validation/gpu_error_message.cpp
```diff
@@ -436,6 +436,60 @@ bool gpuav::Validator::GenerateValidationMessage(const uint32_t *debug_record, c
...
+            out_vuid_msg = "VUID-RuntimeSpirv-OpRayQueryInitializeKHR-06890";
+            error_found = true;
+        } break;
+        case kInstErrorRayQueryOpaque: {
+            const uint32_t value = debug_record[kInstRayQueryOutParam0];
+            strm << "OpRayQueryInitializeKHR operand Ray Flags is 0x" << std::hex << value << ". ";
+            out_vuid_msg = "VUID-RuntimeSpirv-OpRayQueryInitializeKHR-06891";
+            error_found = true;
+        } break;
 #if 0
```
jeremyg-lunarg, line 493: #if 0???

jeremyg-lunarg (approved):

layers/gpu_shaders/inst_ray_query.comp
spencer-lunarg: we have GLSL experts around, we should ask

layers/gpu_validation/gpu_error_message.cpp
arno-lunarg: Oh this is me, a leftover, will remove it in a subsequent PR

spencer-lunarg: I need to rebase this, so can just make an additional commit and remove it

tests/unit/gpu_av_ray_query.cpp
spencer-lunarg: Ran the samples, this is working fine, last thing I want to do is still double check the spec what can be done if we find this instruction is bad

PR #7598, opened 2024-02-27
tests: Fix NegativeGpuAVOOB.ObjectUniformBufferTooSmall

jeremyg-lunarg:
This test was hitting VUID-VkGraphicsPipelineCreateInfo-renderPass-06055 It is currently the only test using the shader_objects=true path in ShaderBufferSizeTest().

### DIFF
--- a/tests/unit/gpu_av_oob.cpp
+++ b/tests/unit/gpu_av_oob.cpp
@@ -214,6 +214,18 @@ void NegativeGpuAVOOB::ShaderBufferSizeTest(VkDeviceSize buffer_size, VkDeviceSi
     pipe.InitState();
     pipe.shader_stages_ = {vs.GetStageCreateInfo(), fs.GetStageCreateInfo()};
     pipe.gp_ci_.layout = pipeline_layout.handle();
+
+    VkFormat color_formats = VK_FORMAT_UNDEFINED;
+    VkPipelineRenderingCreateInfoKHR pipeline_rendering_info = vku::InitStructHelper();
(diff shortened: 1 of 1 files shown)


tests/unit/gpu_av_oob.cpp
```diff
@@ -214,6 +214,18 @@ void NegativeGpuAVOOB::ShaderBufferSizeTest(VkDeviceSize buffer_size, VkDeviceSi
     pipe.InitState();
     pipe.shader_stages_ = {vs.GetStageCreateInfo(), fs.GetStageCreateInfo()};
     pipe.gp_ci_.layout = pipeline_layout.handle();
+
+    VkFormat color_formats = VK_FORMAT_UNDEFINED;
+    VkPipelineRenderingCreateInfoKHR pipeline_rendering_info = vku::InitStructHelper();
+    const auto depth_format = FindSupportedDepthOnlyFormat(gpu());
```
artem-lunarg, line 220: comment mostly to understand this better. reviewed spencer's PR for offensive VU, it might work without depth, I assume that pipeline helper does not create depth attachment, but might be wrong. But I'm fine with depth here if it works.

artem-lunarg (approved):
LGTM. One comment.

tests/unit/gpu_av_oob.cpp
jeremyg-lunarg: the depth might not be necessary. I copied this from `PositiveDynamicRendering.Draw` without understanding. :)

PR #7607, opened 2024-02-29
layers: Ignore pStages for non shader GPL substate

spencer-lunarg:
Spec change is here https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6515


layers/state_tracker/pipeline_sub_state.cpp
```diff
@@ -27,7 +27,7 @@ VkPipelineLayoutCreateFlags PipelineSubState::PipelineLayoutCreateFlags() const
 VertexInputState::VertexInputState(const vvl::Pipeline &p, const safe_VkGraphicsPipelineCreateInfo &create_info)
     : PipelineSubState(p) {
     for (uint32_t i = 0; i < create_info.stageCount; i++) {
-        if (create_info.pStages[i].stage == VK_SHADER_STAGE_MESH_BIT_EXT) {
+        if (create_info.pStages && create_info.pStages[i].stage == VK_SHADER_STAGE_MESH_BIT_EXT) {
```
spencer-lunarg, line 30: this spot is interesting  because at this point we don't fully know if it is a Vertex Input or not by itself ... looking into a better way to handle this code snippet in general

spencer-lunarg:
Internal CI failure from non-related GPU-AV test

PR #7609, opened 2024-02-29
layers: Check vkGetDescriptorSetLayoutSupport

spencer-lunarg:
From https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6269

Call `vkGetDescriptorSetLayoutSupport` for apps

this catches Mutable Descriptor tests failing on MESA because they don't support `VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER` (but NVIDIA does)

### DIFF
--- a/layers/core_checks/cc_descriptor.cpp
+++ b/layers/core_checks/cc_descriptor.cpp
@@ -633,6 +633,7 @@ bool CoreChecks::PreCallValidateCreateDescriptorSetLayout(VkDevice device, const
     uint32_t uniform_buffer_dynamic = pCreateInfo->bindingCount;
     uint32_t storage_buffer_dynamic = pCreateInfo->bindingCount;
     const Location create_info_loc = error_obj.location.dot(Field::pCreateInfo);
+    bool using_mutable_type = false;
 
     for (uint32_t i = 0; i < pCreateInfo->bindingCount; ++i) {
         const Location binding_loc = create_info_loc.dot(Field::pBindings, i);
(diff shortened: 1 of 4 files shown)


tests/unit/descriptors.cpp
```diff
@@ -4141,286 +4141,309 @@ TEST_F(NegativeDescriptors, BindingDescriptorSetFromHostOnlyPool) {
     m_commandBuffer->end();
 }
 
-TEST_F(NegativeDescriptors, CopyMutableDescriptors) {
```
spencer-lunarg, line 4144: this test was already basically 3 tests, so just broke it up to improve coverage on MESA (2 of the 3 work, and didn't want to skip the other 2)

layers/core_checks/cc_descriptor.cpp
```diff
@@ -759,6 +763,26 @@ bool CoreChecks::PreCallValidateCreateDescriptorSetLayout(VkDevice device, const
...
+        if (IsExtEnabledByCreateinfo(device_extensions.vk_khr_maintenance3)) {
+            DispatchGetDescriptorSetLayoutSupportKHR(device, pCreateInfo, &support);
+        } else {
+            DispatchGetDescriptorSetLayoutSupport(device, pCreateInfo, &support);
+        }
+        if (!support.supported) {
+            std::stringstream error_str;
+            error_str << "was passed into vkGetDescriptorSetLayoutSupport and returned not supported";
+            if (using_mutable_type) {
+                error_str << " (verify mutable type are supported, not every driver supports mutable types such as "
```
arno-lunarg, line 778:
```suggestion
                error_str << " (verify mutable types are supported, not every driver supports mutable types such as "
```

arno-lunarg (approved):

PR #7613, opened 2024-02-29
scripts: Use pickle to cache the VkObject

spencer-lunarg:
Currently it takes my machine `50.40 seconds` to generate all 53 files

This uses the python `pickle` module to cache the data so that we don't have to reparse the entire XML each time

This brought my machine down to `23.04` seconds

### DIFF
--- a/scripts/generate_source.py
+++ b/scripts/generate_source.py
@@ -27,10 +27,11 @@ import tempfile
 import difflib
 import json
 import common_ci
+import pickle
 from xml.etree import ElementTree
 from generate_spec_error_message import GenerateSpecErrorMessage
 
-def RunGenerators(api: str, registry: str, grammar: str, directory: str, styleFile: str, targetFilter: str):
+def RunGenerators(api: str, registry: str, grammar: str, directory: str, styleFile: str, targetFilter: str, caching: bool):
 
     has_clang_format = shutil.which('clang-format') is not None
     if not has_clang_format:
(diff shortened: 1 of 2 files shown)


scripts/generate_source.py
```diff
@@ -348,13 +354,27 @@ def RunGenerators(api: str, registry: str, grammar: str, directory: str, styleFi
...
-        reg.apiGen()
+        # The cached data is saved inside the BaseGenerator, so search for it and try
+        # to reuse the parsing for each generator file.
+        if not cacheVkObjectData:
+            if os.path.isfile(cachePath):
+                file = open(cachePath, 'rb')
+                cacheVkObjectData = pickle.load(file)
+                file.close()
+
+        if cacheVkObjectData and 'regenerate' not in generators[target]:
```
artem-lunarg, line 365: python question: is it the check for existence of the key or the value for `generate` is also checked?

artem-lunarg (approved):
LGTM. One questions. Also caching is hard, hopefully this work flawlessly against all those talks. Otherwise we can add an option to disable this...

scripts/generate_source.py
spencer-lunarg: well the idea is the absence of it is `false`

spencer-lunarg:
> LGTM. One questions. Also caching is hard, hopefully this works flawlessly against all those talks. Otherwise we can add an option to disable this...

Ya, maybe a "enable" / "opt-in" method to start would be smart... I didn't want to merge this until I could test it more places, but that might be a better way forward

PR #7617, opened 2024-02-29
tests: Fix resource in use error

artem-lunarg:
Transfer queue was not waited on. Potentially fixes CI failure.


tests/unit/sync_val_positive.cpp
```diff
@@ -1426,4 +1426,5 @@ TEST_F(PositiveSyncVal, QSSynchronizedWritesAndAsyncWait) {
     vk::QueueSubmit2(*m_default_queue, 1, &submit2, VK_NULL_HANDLE);
 
     m_default_queue->wait();
+    transfer_queue->wait();
```
artem-lunarg, line 1429: Another option to wait just on device which is equivalent to wait on all queues, but wanted to check specific queue here.

jeremyg-lunarg (approved):

Issue #7614, opened 2024-02-29
DrawIndex usage in not first pre-raster stage not validated


HansKristian-Work:
**Describe the situation in which you encountered the missing validation**
Issue description:

In a task/mesh/frag pipeline, if mesh shader uses DrawIndex, it is not valid usage, but validation does not catch it.

**Valid Usage IDs requested**
Please include the valid usage IDs for the checks you are requesting:

> When task or mesh shaders are used, only the first active stage will have proper access to the variable. The value read by other stages is undefined.

There might be a missing VU in the spec.

**Additional Context**

[repro.foz.zip](https://github.com/KhronosGroup/Vulkan-ValidationLayers/files/14449239/repro.foz.zip)

Unzip first on a driver that supports VK_EXT_mesh_shader, then run:

```
fossilize-replay repro.foz --enable-validation
```

Notice how there is no validation error.

spencer-lunarg:
@HansKristian-Work so I created a PR, but I marked it as a `LogUndefinedValue` warning instead of an "error"

I don't think this needs a VU, we have had this discussion before for other related shader issues, the "value is undefined" but it is not "undefined behavior" so technically doesn't need a VU

If hit this because a driver crashed, then clearly we don't have CTS coverage and think others would be ok then to make this just illegal outright

spencer-lunarg:
@HansKristian-Work actually @artem-lunarg brought up a good question, the spec doesn't really clarify, if the task shader doesn't access the `DrawIndex` can the mesh still access it? Or the presence of both shaders the issue?

spencer-lunarg:
discussed in https://gitlab.khronos.org/vulkan/vulkan/-/issues/3833

Going to create a VU for this, but the WG agreed that this should be fully banned so happy with getting the PR in and applying the VUID when it lands in the future spec

PR #7621, opened 2024-03-01
layers: Add Mesh DrawIndex undefined value

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7614

### DIFF
--- a/layers/core_checks/cc_pipeline_graphics.cpp
+++ b/layers/core_checks/cc_pipeline_graphics.cpp
@@ -94,6 +94,7 @@ bool CoreChecks::ValidateGraphicsPipeline(const vvl::Pipeline &pipeline, const L
     skip |= ValidateGraphicsPipelineDynamicRendering(pipeline, create_info_loc);
     skip |= ValidateGraphicsPipelineShaderState(pipeline, create_info_loc);
     skip |= ValidateGraphicsPipelineBlendEnable(pipeline, create_info_loc);
+    skip |= ValidateGraphicsPipelineMeshTask(pipeline, create_info_loc);
     skip |= ValidateGraphicsPipelineExternalFormatResolve(pipeline, subpass_desc, create_info_loc);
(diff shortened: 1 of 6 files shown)


layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -1177,6 +1178,22 @@ bool CoreChecks::ValidateGraphicsPipelineBlendEnable(const vvl::Pipeline &pipeli
     return skip;
 }
 
+bool CoreChecks::ValidateGraphicsPipelineMeshTask(const vvl::Pipeline &pipeline, const Location &create_info_loc) const {
+    bool skip = false;
+    const bool has_mesh = (pipeline.active_shaders & VK_SHADER_STAGE_MESH_BIT_EXT) != 0;
+    const bool has_task = (pipeline.active_shaders & VK_SHADER_STAGE_TASK_BIT_EXT) != 0;
+    if (has_mesh && has_task) {
```
spencer-lunarg, line 1185: there are other mesh/task VUs I need to add here, they will need to know if there is a mesh and/or task so preparing for that

artem-lunarg: > only the first active stage will have proper access to the variable

Does it mean we assume that task shader accessed draw index, so mesh shader can't use it?

spencer-lunarg: hmm, thats a good question, let me look

spencer-lunarg: confirmed the issue is when you have both a task and mesh stage in your pipeline, just having it in the mesh shader itself is illegal, there was no desire to let apps "just no access it" as built-ins are different than descriptors

artem-lunarg (approved):

PR #7624, opened 2024-03-01
gpu: Cleanup error message constants

spencer-lunarg:
This cleans up the mess of how we report back which error GPU-AV found

comments inlined


BUILD.gn
```diff
@@ -155,6 +155,7 @@ vvl_sources = [
   "layers/external/vma/vk_mem_alloc.h",
   "layers/external/vma/vma.cpp",
   "layers/external/vma/vma.h",
+  "layers/gpu_shaders/gpu_error_constants.h",
```
spencer-lunarg, line 158: didn't want to start making too many extra files, but it seems logical to split the error constant with the rest of the "misc" constants

layers/gpu_shaders/gpu_error_constants.h
```diff
@@ -0,0 +1,221 @@
...
+// Size of Common and Stage-specific Members
+const int kInstStageOutCnt = kInstCommonOutCnt + 3;
+
+// This identifies the validation error
+// We use groups to more easily mangage the many int values not conflicting
+const int kErrorGroup = kInstStageOutCnt;
+const int kErrorSubCode = kErrorGroup + 1;
+
+// Maximum Output Record Member Count
+const int kInstMaxOutCnt = kErrorSubCode + 6;
```
spencer-lunarg, line 59: everything from here and above groups the "where in the buffer" offsets, helpful having it next to each other

layers/gpu_validation/gpu_error_message.cpp
```diff
@@ -554,8 +568,23 @@ bool gpuav::Validator::AnalyzeAndGenerateMessages(VkCommandBuffer cmd_buffer, Vk
 
     std::string error_msg;
     std::string vuid_msg;
-    const bool error_found =
-        GenerateValidationMessage(debug_record, cmd_resources, descriptor_sets, error_msg, vuid_msg, oob_access);
```
spencer-lunarg, line 558: after turning this function into a single switch case that called other functions, it made no sense to have it not just be directly called here (and only pass the extra params to each `LogMessageInst*` as needed)

jeremyg-lunarg (approved):
LGTM

PR #7629, opened 2024-03-04
layers: Cleanup more error messages

spencer-lunarg:
Cleanup of some error messages


layers/core_checks/cc_cmd_buffer.cpp
```diff
@@ -874,71 +874,6 @@ bool CoreChecks::PreCallValidateCmdExecuteCommands(VkCommandBuffer commandBuffer
             skip |= LogError("VUID-vkCmdExecuteCommands-pCommandBuffers-00088", objlist, cb_loc,
                              "(%s) is not VK_COMMAND_BUFFER_LEVEL_SECONDARY.", FormatHandle(pCommandBuffers[i]).c_str());
         } else {
-            if (sub_cb_state.activeRenderPass && sub_cb_state.activeRenderPass->UsesDynamicRendering()) {
```
spencer-lunarg, line 877: these were

```
VUID-vkCmdExecuteCommands-pCommandBuffers-09504
VUID-vkCmdExecuteCommands-pCommandBuffers-09505
```

we have no tests for them I looking at the code, seems broken, removing until I get some tests in as rather not have false positives

artem-lunarg (approved):

PR #7631, opened 2024-03-04
gpu: Renamings in gpu_error_constants.h

arno-lunarg:
Renamed `gpu_error_constants.h` to `gpu_error_header.h`. Renamed constants, to clarify intents. Added `Offset` suffix to values representing offsets.
gpu_error_constants now includes `gpu_error_codes.h`, a file holding constant values used to retrieve VUIDs


layers/gpu_shaders/gpu_error_header.h
```diff
@@ -0,0 +1,179 @@
...
+// |	- Instruction Id
+// |	- Shader stage Id
+// |	- Shader stage info (3 integers)
+// | 	- Error group (Id unique to the shader/instrumentation code that wrote the error)
+// |	- subcode (maps to VUIDs)
+// | --------------------------------
+// | Error specific parameters
+// \---------------------------------
+//
+// The size of these parts depends on the validation being done.
```
arno-lunarg, line 45: I need to remove this sentence, will do in another PR, do not want to go through CI again

layers/gpu_shaders/inst_bindless_descriptor.comp
```diff
@@ -165,20 +165,22 @@ bool inst_bindless_descriptor(const uint inst_num, const uvec4 stage_info, const
         uint rec_len = 14u;
```
arno-lunarg, line 165: This `14u` now comes out of nowhere, did not bother changing it because I am already messing with record sizes in my reprise of https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7320
I am thinking about having a fixed record size big enough to hold the biggest record type. It may simplify things, and wasted space would be 1) small and 2) not important to have it since apps are not supposed to emit errors

spencer-lunarg: I agree, could be a nice followup change to this

PR #7591, opened 2024-02-27
Add VK_ARM_render_pass_striped support

mohfai-arm:
Closes #7590

PR #7634, opened 2024-03-05
build: Fix circular dependency

spencer-lunarg:
fix for https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7591#discussion_r1511292914

(cc @y-novikov)

### DIFF
--- a/layers/utils/vk_layer_utils.cpp
+++ b/layers/utils/vk_layer_utils.cpp
@@ -20,6 +20,7 @@
 #include <string.h>
 #include <sys/stat.h>
 
+#include "containers/range_vector.h"
 #include "vulkan/vulkan.h"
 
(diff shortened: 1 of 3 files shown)


arno-lunarg (approved):

layers/utils/vk_layer_utils.cpp
```diff
@@ -20,6 +20,7 @@
 #include <string.h>
 #include <sys/stat.h>
 
+#include "containers/range_vector.h"
```
y-novikov, line 23: Thanks for working on this, but now it fails here.
You can also see the error message in https://github.com/KhronosGroup/Vulkan-ValidationLayers/actions/runs/8149491090/job/22274236950?pr=7634.

PR #7640, opened 2024-03-05
gn: Use check_call for CI

spencer-lunarg:
@y-novikov I found why CI has not been failing properly

(I assume this first run will fail, from there I will turn this PR into something that fixes things)


y-novikov:
> CI Vulkan-ValidationLayers build # 16014 passed.

Hmm, doesn't look like it did.

spencer-lunarg:
> CI Vulkan-ValidationLayers build # 16014 passed.
> Hmm, doesn't look like it did.

@y-novikov this is our internal CI, but the Github Actions are now blocked (like we thought they were)

image

jeremyg-lunarg (approved):

spencer-lunarg:
Internal CI failure was from an `NegativeGpuAVRayQuery.TMaxLessThenTmin` isolated issue outside the scope of this change

y-novikov:
Super, thank you for fixing both and the build and the future checks!

PR #7641, opened 2024-03-05
tests: Cleanup pipeline class

spencer-lunarg:
This is a "remove unnecessary test code" change

The big 3 things are

1. Remove the `InitState` from the pipeline helpers... it is confusing and not needed 99% of the time
2. Remove legacy "set the color blend attachment vector" logic
3. Allow for `pNext` to be used when doing the construction (nice for dynamic rendering)

### DIFF
--- a/tests/framework/pipeline_helper.cpp
+++ b/tests/framework/pipeline_helper.cpp
@@ -12,10 +12,12 @@
 
 #include "pipeline_helper.h"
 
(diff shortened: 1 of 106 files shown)


arno-lunarg:
> Remove the InitState from the pipeline helpers... it is confusing and not needed 99% of the time
凄い !!!!!!!!!!!!

It was indeed confusing I think

arno-lunarg (approved):

PR #7653, opened 2024-03-07
layers: Check for non-zero stageCount

spencer-lunarg:
In yet-another-turn for the GPL saga, the WG decided to just ban `stageCount` being non-zero when not need to make everyone's life easier... https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6530


artem-lunarg (approved):
ZeroLittleZGIF

Issue #7592, opened 2024-02-27
layer_chassis_dispatch_generator script not generating the unwrapping code


mohfai-arm:
**Describe the Issue**
DispatchQueueSubmit2 and DispatchQueueSubmit2KHR in layer_chassis_dispatch.cpp missing unwrapping code if pCommandBufferInfos[index1].pNext contains VkRenderPassStripeSubmitInfoARM struct. 

Code from DispatchQueueSubmit2: Line 2629 and 4854
```
 if (local_pSubmits[index0].pCommandBufferInfos) {
          for (uint32_t index1 = 0; index1 < local_pSubmits[index0].commandBufferInfoCount; ++index1) {
            }
   }
```

should be

```
                 if (local_pSubmits[index0].pCommandBufferInfos) {
                     for (uint32_t index1 = 0; index1 < local_pSubmits[index0].commandBufferInfoCount; ++index1) {
+                        if (local_pSubmits[index0].pCommandBufferInfos[index1].pNext) {
+                            WrapPnextChainHandles(layer_data, local_pSubmits[index0].pCommandBufferInfos[index1].pNext);
+                        }
                     }
                 }
```

spencer-lunarg:
... so I see https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6798 got to use after all :cry:

PR #7656, opened 2024-03-07
scripts: Fix WrapPnextChainHandles

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7592

### DIFF
--- a/layers/vulkan/generated/layer_chassis_dispatch.cpp
+++ b/layers/vulkan/generated/layer_chassis_dispatch.cpp
@@ -166,6 +166,8 @@ void WrapPnextChainHandles(ValidationObject* layer_data, const void* pNext) {
                     for (uint32_t index0 = 0; index0 < safe_struct->groupCount; ++index0) {
                         if (safe_struct->pGroups[index0].pStages) {
                             for (uint32_t index1 = 0; index1 < safe_struct->pGroups[index0].stageCount; ++index1) {
+                                WrapPnextChainHandles(layer_data, safe_struct->pGroups[index0].pStages[index1].pNext);
+
                                 if (safe_struct->pGroups[index0].pStages[index1].module) {
                                     safe_struct->pGroups[index0].pStages[index1].module =
                                         layer_data->Unwrap(safe_struct->pGroups[index0].pStages[index1].module);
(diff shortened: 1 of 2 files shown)


scripts/generators/layer_chassis_dispatch_generator.py
```diff
@@ -429,13 +429,8 @@ def uniquifyMembers(self, members: list[Member], prefix: str, arrayIndex: int, i
                                     pre_code += f'{new_prefix}[{index}].initialize(&{member.name}[{index}]);\n'
                             else:
                                 pre_code += f'{new_prefix}[{index}] = {member.name}[{index}];\n'
-                            if process_pnext:
```
spencer-lunarg, line 432: I think this was just a classic python bug because this was indented an extra level, it was in the `if topLevel` block but really didn't need to be

layers/vulkan/generated/layer_chassis_dispatch.cpp
```diff
@@ -4852,9 +4852,7 @@ VkResult DispatchQueueSubmit2KHR(VkQueue queue, uint32_t submitCount, const VkSu
                 }
                 if (local_pSubmits[index0].pCommandBufferInfos) {
                     for (uint32_t index1 = 0; index1 < local_pSubmits[index0].commandBufferInfoCount; ++index1) {
-                        if (local_pSubmits[index0].pCommandBufferInfos[index1].pNext) {
```
spencer-lunarg, line 4855: no reason to check for `pNext`, if it is null, it exits `WrapPnextChainHandles` safely

scripts/generators/layer_chassis_dispatch_generator.py
```diff
@@ -429,13 +429,8 @@ def uniquifyMembers(self, members: list[Member], prefix: str, arrayIndex: int, i
...
-                                pre_code += f'WrapPnextChainHandles(layer_data, {new_prefix}[{index}].pNext);\n'
-                        if member.type == "VkCommandBufferSubmitInfo" and process_pnext:
```
artem-lunarg, line 434: The `VkCommandBufferSubmitInfo` case. Is it handled by the general code path too?

artem-lunarg (approved):

scripts/generators/layer_chassis_dispatch_generator.py
spencer-lunarg: yes, it is the spot in `DispatchQueueSubmit2KHR`

PR #7662, opened 2024-03-07
tests: Fixes related to feature requirement management

aqnuep:


tests/unit/shader_storage_image.cpp
```diff
@@ -51,6 +51,7 @@ TEST_F(NegativeShaderStorageImage, MissingFormatRead) {
                OpDecorate %12 Binding 0
                OpDecorate %22 BuiltIn WorkgroupSize
                OpDecorate %12 NonReadable
+               OpDecorate %12 NonWritable
```
spencer-lunarg, line 54: what error was this tripping?

tests/unit/descriptors.cpp
```diff
@@ -4877,6 +4877,14 @@ TEST_F(NegativeDescriptors, DescriptorIndexingMissingFeatures) {
     TEST_DESCRIPTION("Use partially bound descriptor flag without feature.");
 
     AddRequiredExtensions(VK_EXT_DESCRIPTOR_INDEXING_EXTENSION_NAME);
+    AddDisabledFeature(vkt::Feature::descriptorBindingPartiallyBound);
```
spencer-lunarg, line 4880: honestly for these "feature" specific tests, we could just do it the old fashion way, but happy for it to work either way

spencer-lunarg (approved):
LGTM - Just want clarification on the `NegativeShaderStorageImage` test

spencer-lunarg:
> CI Vulkan-ValidationLayers build # 16067 failed.

seems `PositiveGpuAV.BindingPartiallyBound` is hitting

> Validation Error: [ VUID-VkDeviceCreateInfo-sType-unique ] | MessageID = 0xfadd6202 | vkCreateDevice(): pCreateInfo->pNext chain contains duplicate structure types: VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FEATURES_2 appears multiple times.

tests/unit/shader_storage_image.cpp
aqnuep: Unsupported / not enabled feature: `shaderStorageImageWriteWithoutFormat`.

tests/unit/descriptors.cpp
aqnuep: I think the root of the problem is that if you have a required feature, then all supported features are enabled in the same struct by the framework currently, which is unexpected.

If that will be fixed, then these `AddDisabledFeature` calls would not be necessary in any test case.

aqnuep:
I've pushed a fix for `PositiveGpuAV.BindingPartiallyBound`. Hopefully CI passes now.

tests/unit/descriptors.cpp
aqnuep: The problem is that previously the "old fashioned way" did not work together with the feature requirements approach, because the latter was simply ignored in that case, which does not interact well with "implicit feature requirements" corresponding to core versions.

We really need to get the two working together (by adding support for merging them), or abandon the old fashion way.

The previous behavior where the effects of `AddRequiredFeature` would simply be silently ignored if one specifies a create info pNext chain was simply broken (not all device-create structs are feature structs, and even if they are, it should be a test case error, not a silent ignorance of explicit feature requirements).

aqnuep:
Great, finally CI passes for this! @spencer-lunarg, let me know if you have any other questions or comments.

tests/unit/shader_storage_image.cpp
spencer-lunarg: ok, we are good then... the `VK_KHR_format_feature_flags2` is a "is enabled if expose" extensions, so I have spent time in the past dealing with edge cases and wanted to confirm what was going on here

Issue #7250, opened 2024-01-05
SYNC-HAZARD-WRITE-AFTER-WRITE when using multiple threads


RYDB3RG:
Hi,

Is `SYNC-HAZARD-WRITE-AFTER-WRITE`  expected when 2 threads submit with a mutex into the same queue?

repro:

    TEST_F(NegativeSyncVal, CopyPresentLoop) {
        AddSurfaceExtension();
        RETURN_IF_SKIP(InitSyncValFramework());
        RETURN_IF_SKIP(InitState());
        if (!InitSwapchain()) {
            GTEST_SKIP() << "Cannot create surface or swapchain";
        }

        std::mutex queue_mutex;
        const auto &queues = m_device->graphics_queues();
        auto &queue0 = queues[0];
        auto &queue1 = queues[0];

        VkCommandPoolCreateInfo pool_create_info = vku::InitStructHelper();
        pool_create_info.queueFamilyIndex = m_device->graphics_queue_node_index_;
        pool_create_info.flags = VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT;
        vkt::CommandPool pool0(*m_device, pool_create_info);

        const vkt::Fence fence0(*m_device);
        vkt::CommandBuffer cmd0(m_device, &pool0);
        vkt::Buffer src(*m_device, 1024 * 128, VK_BUFFER_USAGE_TRANSFER_SRC_BIT);
        vkt::Buffer dst(*m_device, 1024 * 128, VK_BUFFER_USAGE_TRANSFER_DST_BIT | VK_BUFFER_USAGE_INDEX_BUFFER_BIT);

        std::thread thread([&]
        {
            while(true)
            {
                VkBufferCopy copy_info{};
                copy_info.size = 1024 * 128;

                cmd0.begin();
                vk::CmdCopyBuffer(cmd0.handle(), src.handle(), dst.handle(), 1, &copy_info);
                cmd0.end();

                vk::ResetFences(m_device->device(), 1, &fence0.handle());

                VkSubmitInfo submit = vku::InitStructHelper();
                submit.commandBufferCount = 1;
                submit.pCommandBuffers = &cmd0.handle();

                {
                    std::unique_lock<std::mutex> lock(queue_mutex);
                    ASSERT_EQ(VK_SUCCESS, vk::QueueSubmit(queue1->handle(), 1, &submit, fence0.handle()));
                }

[...]

PR #7663, opened 2024-03-07
sync: Fix WaitForFences leak

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7250

There are syncval endpoints that compute state in the Validate phase
and use thread local storage to propagate it to the Record phase,
which does the final update.

This change fixes one side effect of this approach in the multithreaded
environment. In some cases, the accesses that should be synchronized by
vkWaitForFences can leak through this synchronization point, and then
hazard with the following accesses.

In following scenario two threads work with the same queue.
The queue access is externally synchronized.

1. Thread A: Submits some memory accesses. Signals a fence.
2. Thread B: Submits empty command buffer. Finishes only Validate.
   This imports accesses from step 1 and the batch is stored in the
   thread local storage to be consumed later by the Record.
3. Thread A: Calls WaitForFences with a fence from step 1.
   This processes the accesses from step 1, but cannot see the copy
   of those accesses (in TLS) produced by step 2.
4. Thread B. Finishes Record. This restores accesses from step 2, that
   otherwise should have been processed by the WaitForFences.
5. Thread A. Submits more accesses. They hazard with accesses that
   were restored in step 4.

### DIFF
--- a/layers/sync/sync_validation.cpp
+++ b/layers/sync/sync_validation.cpp
@@ -36,6 +36,17 @@ void SyncValidator::ApplyTaggedWait(QueueId queue_id, ResourceUsageTag tag) {
     auto tagged_wait_op = [queue_id, tag](const std::shared_ptr<QueueBatchContext> &batch) {
         batch->ApplyTaggedWait(queue_id, tag);
         batch->Trim();
+
+        // If there is a *pending* last batch then apply tagged wait for its accesses too.
+        // A pending last batch might exist if this wait was initiated between QueueSubmit's
+        // Validate and Record phases (from a different thread). The pending last batch might
+        // contain *imported* accesses that are in the scope of this wait.
+        auto batch_queue_state = batch->GetQueueSyncState();
+        auto pending_batch = batch_queue_state ? batch_queue_state->PendingLastBatch() : nullptr;
+        if (pending_batch) {
+            pending_batch->ApplyTaggedWait(queue_id, tag);
+            pending_batch->Trim();
+        }
     };
     ForAllQueueBatchContexts(tagged_wait_op);
 }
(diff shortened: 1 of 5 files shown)


layers/sync/sync_validation.cpp
```diff
@@ -36,6 +36,17 @@ void SyncValidator::ApplyTaggedWait(QueueId queue_id, ResourceUsageTag tag) {
     auto tagged_wait_op = [queue_id, tag](const std::shared_ptr<QueueBatchContext> &batch) {
         batch->ApplyTaggedWait(queue_id, tag);
         batch->Trim();
+
+        // If there is a *pending* last batch then apply tagged wait for its accesses too.
```
artem-lunarg, line 40: That's the main part of the fix. The rest of the code is mostly concerned with providing access to the pending last batch (store it in the queue state instead of directly in the TLS).

spencer-lunarg (approved):
LGTM - but of course would like @jzulauf-lunarg approval

layers/sync/sync_validation.cpp
```diff
@@ -2794,7 +2804,7 @@ bool SyncValidator::PreCallValidateQueuePresentKHR(VkQueue queue, const VkPresen
     batch->Cleanup();
 
     if (!skip) {
-        cmd_state->present_batch = std::move(batch);
+        cmd_state->queue->SetPendingLastBatch(std::move(batch));
```
jzulauf-lunarg, line 2807: There's still a race between access context that has been copied, but not yet been set to pending, with the Wait for Fence occurs.

It seems that during batch cleanup (or even perhaps at record time), one would need to see if an intervening fence has occured, but I'm not sure that would catch all conditions.  

Maybe it's safer and more performant to defer the fence application until the next queue batch submission for a given queue, or just before the queue "head" batch is checked for async accesses.

artem-lunarg: > There's still a race between access context that has been copied, but not yet been set to pending, with the Wait for Fence occurs.

I can't see how this can be a race. Fence wait can't happen (needs write lock) at the time when submit validate holds read lock of shared mutex of validator object.

jzulauf-lunarg: Okay I'm wrong. This will work

jzulauf-lunarg (approved):

Issue #4646, opened 2022-10-11
Bug in VUID-VkDescriptorImageInfo-imageLayout-00344


forenoonwatch:
**Describe the Issue**
Validation layers emit a validation error upon calling vkCmdDrawIndexed when binding a descriptor set expecting an image that is in a layout other than undefined if the barrier transitioning the image is located in another command buffer, even if these command buffers are submitted together in the correct order.

To replicate, create 2 command buffers, emit a barrier to transition an image on command buffer 1. On command buffer 2, bind a descriptor set referencing this image at its desired layout, and invoke vkCmdDrawIndexed. Submit both of these command buffers in order as the pCommandBuffers of a single VkSubmitInfo.

**Valid Usage ID**
```
VUID-VkDescriptorImageInfo-imageLayout-00344(ERROR / SPEC): msgNum: -564812795 - Validation Error: [ VUID-VkDescriptorImageInfo-imageLayout-00344 ] Object 0: handle = 0x29359b01db8, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xde55a405 | vkCmdDrawIndexed: Cannot use VkImage 0x5c59a0000000056[] (layer=0 mip=0) with specific layout VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL that doesn't match the previously used layout VK_IMAGE_LAYOUT_UNDEFINED. The Vulkan spec states: imageLayout must match the actual VkImageLayout of each subresource accessible from imageView at the time this descriptor is accessed as defined by the image layout matching rules (https://vulkan.lunarg.com/doc/view/1.3.224.1/windows/1.3-extensions/vkspec.html#VUID-VkDescriptorImageInfo-imageLayout-00344)
    Objects: 1
        [0] 0x29359b01db8, type: 6, name: NULL
[ERROR: Validation]
```

**Environment:**
 - OS: Windows 10
 - GPU: NVIDIA GeForce GTX 760
 - SDK or header version if building from repo: 1.3.224.1
 - Options enabled (synchronization, best practices, etc.): Validation - Standard

**Additional context**
[...]

PR #7669, opened 2024-03-08
tests: Add PositiveDescriptors.ImageViewAttachmentThenRead

jeremyg-lunarg:
This tests attempts to replicate the false positive from #4646


jeremyg-lunarg:
Draft until this actually replicates the problem.

PR #7672, opened 2024-03-10
Fix validating swapchain images when they are acquired before getting them

### LINKED ISSUE #6068: Swapchain image state validations depend on vkGetSwapchainImagesKHR being called early
**Environment:**
 - OS: Windows 10 x64
 - GPU: Nvidia 4090
 - SDK or header version if building from repo: 1.3.250
 - Options enabled (synchronization, best practices, etc.): just vanilla `VK_LAYER_KHRONOS_validation`

**Describe the Issue**

Swapchain state tracking around acquired image indices cannot happen until the `SWAPCHAIN_STATE::images` value is initialized (at least partially, up to and including the index that is used by Present).

This happens because AcquireImage calls get silently dropped when the validation layer isn't aware of and tracking that image index.

https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/1541e00a63cd125f15d231d5a8059ebe66503b25/layers/state_tracker/image_state.cpp#L589-L590

Initialization of swapchain images happens in `PostCallRecordGetSwapchainImagesKHR`:

https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/1541e00a63cd125f15d231d5a8059ebe66503b25/layers/core_checks/cc_wsi.cpp#L769-L790

which calls

https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/1541e00a63cd125f15d231d5a8059ebe66503b25/layers/state_tracker/state_tracker.cpp#L5059-L5061

Sequence of events to reproduce the issue:

1. Create swapchain
2. `vkAcquireNextImageKHR(...)`
3. `vkGetSwapchainImages(...)`
4. Make a submission that at least transitions the swapchain image with index from step 2 to `PRESENT_SRC`
5. `vkQueuePresent(...)` with image index from step 2

Although the code seems vulnerable to the same issue happening with image layouts (because they may not be tracked until later), hitting it is likely impossible in practice, because the app needs to call `vkGetSwapchainImagesKHR` to perform any operation on that image that would change its layout, giving validation layers a chance to see the very first operation.

The workaround for an app is simply to call `vkGetSwapchainImagesKHR` before the first `vkAcquireNextImageKHR`.

**Expected behavior**

This sequence of events should not trigger this validation message

**Valid Usage ID**
If applicable, please include the validation messages encountered leading up to the issue

`VUID-VkPresentInfoKHR-pImageIndices-01296`
Triggered by step 5 in the repro scenario

**Additional context**

<details>
<summary>code or terminal output</summary>

```sh
ERROR & VALIDATION => Validation Error: [ VUID-VkPresentInfoKHR-pImageIndices-01296 ] Object 0: handle = 0x3fbcd60000000028, type = VK_OBJECT_TYPE_SWAPCHAIN_KHR; | MessageID = 0xc7aabc16 | vkQueuePresentKHR: pSwapchains[0] **image at index 0 was not acquired from the swapchain**. The Vulkan spec states: Each element of pImageIndices must be the index of a presentable image acquired from the swapchain specified by the corresponding element of the pSwapchains array, and the presented image subresource must be in the VK_IMAGE_LAYOUT_PRESENT_SRC_KHR layout at the time the operation is executed on a VkDevice (https://github.com/KhronosGroup/Vulkan-Docs/search?q=)VUID-VkPresentInfoKHR-pImageIndices-01296)
```

</details>

ziga-lunarg:
Closes #6068

### DIFF
--- a/layers/state_tracker/state_tracker.cpp
+++ b/layers/state_tracker/state_tracker.cpp
@@ -3807,6 +3807,13 @@ void ValidationStateTracker::RecordAcquireNextImageState(VkDevice device, VkSwap
         if (fence_state && fence_state->IsPresentSyncSwapchainChanged(swapchain_data)) {
             fence_state->SetPresentSync(vvl::PresentSync{});
         }
+        if (*pImageIndex >= swapchain_data->images.size()) {
+            uint32_t swapchainImageCount = *pImageIndex + 1;
+            DispatchGetSwapchainImagesKHR(device, swapchain, &swapchainImageCount, nullptr);
+            std::vector<VkImage> swapchainImages(swapchainImageCount);
+            DispatchGetSwapchainImagesKHR(device, swapchain, &swapchainImageCount, swapchainImages.data());
+            RecordGetSwapchainImages(device, swapchain, &swapchainImageCount, swapchainImages.data());
+        }
         swapchain_data->AcquireImage(*pImageIndex, semaphore_state, fence_state);
     }
 }
@@ -4795,10 +4802,8 @@ void ValidationStateTracker::PostCallRecordCreateShadersEXT(VkDevice device, uin
     }
 }
 
-void ValidationStateTracker::PostCallRecordGetSwapchainImagesKHR(VkDevice device, VkSwapchainKHR swapchain,
-                                                                 uint32_t *pSwapchainImageCount, VkImage *pSwapchainImages,
-                                                                 const RecordObject &record_obj) {
-    if ((record_obj.result != VK_SUCCESS) && (record_obj.result != VK_INCOMPLETE)) return;
+void ValidationStateTracker::RecordGetSwapchainImages(VkDevice device, VkSwapchainKHR swapchain, uint32_t *pSwapchainImageCount,
+                                                      VkImage *pSwapchainImages) {
     auto swapchain_state = Get<vvl::Swapchain>(swapchain);
 
     if (*pSwapchainImageCount > swapchain_state->images.size()) swapchain_state->images.resize(*pSwapchainImageCount);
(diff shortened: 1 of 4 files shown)


tests/unit/wsi.cpp
```diff
@@ -3519,4 +3519,74 @@ TEST_F(NegativeWsi, SwapchainPresentModeInfoImplicit) {
     m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkSwapchainPresentModeInfoEXT-pPresentModes-parameter");
     vk::QueuePresentKHR(m_default_queue->handle(), &present);
     m_errorMonitor->VerifyFound();
-}
\ No newline at end of file
+}
+
+TEST_F(NegativeWsi, AcquireImageBeforeGettingSwapchainImages) {
```
spencer-lunarg, line 3524: is this suppose to be a positive test?

ziga-lunarg: Yes, good catch

layers/state_tracker/state_tracker.cpp
```diff
@@ -3807,6 +3808,14 @@ void ValidationStateTracker::RecordAcquireNextImageState(VkDevice device, VkSwap
         if (fence_state && fence_state->IsPresentSyncSwapchainChanged(swapchain_data)) {
             fence_state->SetPresentSync(vvl::PresentSync{});
         }
+        if (*pImageIndex >= swapchain_data->images.size()) {
+            uint32_t swapchainImageCount = *pImageIndex + 1;
+            DispatchGetSwapchainImagesKHR(device, swapchain, &swapchainImageCount, nullptr);
+            RecordGetSwapchainImages(device, swapchain, &swapchainImageCount, nullptr);
```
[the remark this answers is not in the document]
ziga-lunarg: You're right, this can be skipped

artem-lunarg (approved):
LGTM

spencer-lunarg (approved):

PR #7678, opened 2024-03-12
layers: Improve format feature error messages

spencer-lunarg:
unify the way we print out `VkFormatFeatureFlags2` with a new line to help as these can become very verbose when we print all the features supported for a format

### DIFF
--- a/layers/core_checks/cc_cmd_buffer_dynamic.cpp
+++ b/layers/core_checks/cc_cmd_buffer_dynamic.cpp
@@ -531,37 +531,38 @@ bool CoreChecks::ValidateDrawDynamicStatePipeline(const LastBound& last_bound_st
             skip |= LogError(vuid.color_blend_enable_07627, objlist, loc,
                              "Pipeline was created with VK_DYNAMIC_STATE_COLOR_BLEND_ENABLE_EXT dynamic state, but "
                              "vkCmdSetColorBlendEnableEXT() was not called.");
-        }
-
-        const uint32_t attachment_count = static_cast<uint32_t>(cb_state.active_attachments->size());
-
-        bool advanced_blend = false;
-        for (uint32_t i = 0; i < attachment_count; ++i) {
-            if (cb_state.dynamic_state_value.color_blend_enabled[i]) {
+        } else {
+            const uint32_t attachment_count = static_cast<uint32_t>(cb_state.active_attachments->size());
+            for (uint32_t i = 0; i < attachment_count; ++i) {
+                if (!cb_state.dynamic_state_value.color_blend_enabled[i]) {
+                    continue;
+                }
                 if (cb_state.dynamic_state_value.color_blend_advanced_attachments[i]) {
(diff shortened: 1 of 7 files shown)


layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -531,39 +531,39 @@ bool CoreChecks::ValidateDrawDynamicStatePipeline(const LastBound& last_bound_st
             skip |= LogError(vuid.color_blend_enable_07627, objlist, loc,
                              "Pipeline was created with VK_DYNAMIC_STATE_COLOR_BLEND_ENABLE_EXT dynamic state, but "
                              "vkCmdSetColorBlendEnableEXT() was not called.");
-        }
-
-        const uint32_t attachment_count = static_cast<uint32_t>(cb_state.active_attachments->size());
-
-        bool advanced_blend = false;
-        for (uint32_t i = 0; i < attachment_count; ++i) {
-            if (cb_state.dynamic_state_value.color_blend_enabled[i]) {
```
spencer-lunarg, line 540: this was an extra little cleanup here

early exit on 2nd loop and moved all checks into loop (removed need for `bool advanced_blend`)

layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -531,39 +531,39 @@ bool CoreChecks::ValidateDrawDynamicStatePipeline(const LastBound& last_bound_st
...
 
                 const auto attachment = (*cb_state.active_attachments)[i];
                 if (attachment && ((attachment->format_features & VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BLEND_BIT) == 0)) {
                     const LogObjectList objlist(cb_state.Handle(), pipeline.Handle());
                     skip |= LogError(vuid.blend_feature_07470, objlist, loc,
-                                     "Attachment %" PRIu32
-                                     " format features (%s) do not include VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BLEND_BIT.",
-                                     i, string_VkFormatFeatureFlags2((*cb_state.active_attachments)[i]->format_features).c_str());
+                                     "Color Attachment %" PRIu32
+                                     " image view format (%s) do not include VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BLEND_BIT.\n"
```
arno-lunarg, line 559: More verbose, but I think I prefer wording it like so:
```suggestion
                                     "Color Attachment %" PRIu32
                                     " has an image view format (%s) that does not include VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BLEND_BIT.\n"
```

arno-lunarg (approved):

layers/core_checks/cc_cmd_buffer_dynamic.cpp
spencer-lunarg: verbose... Vulkan... never heard those 2 together

Issue #7677, opened 2024-03-12
False positive `VUID-vkCmdDraw-format-07753` when aliasing image bindings with different numeric types


glebov-andrey:
**Environment:**
 - OS: Linux & Windows
 - SDK or header version if building from repo: current `main` (441b6a90d2320dcc962622d1f3cd5bea71423c7b)
 - Options enabled (synchronization, best practices, etc.): with and without GPU-AV

**Describe the Issue**

When a shader has aliased image descriptor array bindings with different numeric types, the validation layer complains about some valid usage, even when the image views match the expected numeric types at runtime.

The errors occur both with "static" validation (GPU-AV disabled and statically known indices) and "dynamic" validation (GPU-AV enabled, dynamically determined indices).
In the case of "static" validation, it seems that both possible numeric types are checked - both accesses cause errors. In the case of GPU-AV, it seems that one numeric type is checked - one access causes an error. (see code and error messages below)

I've checked SPIR-V which `glslc` generates, and to the best of my understanding, it uses the correct alias for each sampling operation.
It seems that GPU-AV doesn't disable the equivalent Core check, resulting in duplicate errors.
Note also that even though the example uses a dispatch, GPU-AV reports the error as if for `vkCmdDraw`.

**Expected behavior**

The validation layer correctly determines which alias the image is accessed through.

**Valid Usage ID**

`VUID-vkCmdDraw-format-07753`
`VUID-vkCmdDispatch-format-07753`

**Additional context**

Here's a minimized example which causes the errors:

<details>
<summary>GLSL shader</summary>

```glsl
#version 460

#define USE_STATIC_INDEX 1 // change to 0 for GPU-AV only validation

#extension GL_EXT_samplerless_texture_functions : require

layout(set = 0, binding = 0) uniform texture2D float_textures[2];
layout(set = 0, binding = 0) uniform utexture2D uint_textures[2];
layout(set = 0, binding = 1, rgba32f) uniform restrict writeonly image2D output_image; // avoid optimization

layout(push_constant) uniform PushConstants { uint float_texture_index; }; // for the dynamic case

layout(local_size_x = 8, local_size_y = 8) in;

void main() {
    const ivec2 coords = ivec2(gl_GlobalInvocationID.xy);
    const vec4 value = texelFetch(float_textures[
                                                 #if USE_STATIC_INDEX
                                                 0
                                                 #else
                                                 float_texture_index
                                                 #endif
                                                ], coords, 0);
[...]

PR #7682, opened 2024-03-13
layers: Fix aliasing image bindings with different types

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7677

```glsl
layout(set = 0, binding = 0) uniform texture2D float_textures[2]; // only index 0 is accessed
layout(set = 0, binding = 0) uniform utexture2D uint_textures[2]; // only index 1 is accessed
```

before we were checking only by `binding`, now we make sure the `ValidateDescriptor` is checking to make sure the image was accessed in that index before trying to validate it


jeremyg-lunarg (approved):
LGTM

PR #7683, opened 2024-03-13
tests: Use Handle for pipeline helpers

spencer-lunarg:
adds a `Handle()` so no one can go `pipe.pipeline_` anymore (in the quest to make `Handle()` the only way to get a handle)


arno-lunarg (approved):

arno-lunarg:
image-asset

PR #7696, opened 2024-03-14
tests: Fix wsi test for the latest nvidia drivers

artem-lunarg:
Updated nvidia driver to version 551.67 and got support for VK_EXT_swapchain_maintenance1. This wsi test did not pass.

### DIFF
--- a/tests/unit/wsi.cpp
+++ b/tests/unit/wsi.cpp
@@ -2035,33 +2035,43 @@ TEST_F(NegativeWsi, SwapchainMaintenance1ExtensionAcquire) {
     swapchain_create_info.pNext = &present_scaling_info;
 
     m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkSwapchainPresentScalingCreateInfoEXT-presentGravityX-07765");
-    present_scaling_info.presentGravityY = 1;
+    // Disable validation that prevents testing zero gravity value on platforms that provide support for gravity values.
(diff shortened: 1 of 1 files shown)


tests/unit/wsi.cpp
```diff
@@ -2035,33 +2035,41 @@ TEST_F(NegativeWsi, SwapchainMaintenance1ExtensionAcquire) {
     swapchain_create_info.pNext = &present_scaling_info;
 
     m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "VUID-VkSwapchainPresentScalingCreateInfoEXT-presentGravityX-07765");
-    present_scaling_info.presentGravityY = 1;
+    m_errorMonitor->SetAllowedFailureMsg("VUID-VkSwapchainPresentScalingCreateInfoEXT-presentGravityX-07772");
```
jeremyg-lunarg, line 2038: It would be good to have a comment about why these errors are allowed.

jeremyg-lunarg (approved):
LGTM, but a comment about the new allowed errors would be nice

spencer-lunarg (approved):

tests/unit/wsi.cpp
artem-lunarg: Added comment. Considered also to have a conditional validation and do not validate zero value on platforms that support gravity. No need to suppress the errors then. But this will disable part of the validation, at least on nvidia, and will require some profile trickery to have the same coverage.

PR #7699, opened 2024-03-15
layers: Create VUID Map file

spencer-lunarg:
Taking inspiration from `sync_vuid_maps.cpp` I have created a `cc_vuid_maps.cpp` that fully makes use of the `Location` scheme we have

I started with `cc_copy_blit_resolve.cpp`, mainly because it needed this the most, but the goal is to move more and more VUID fetching logic here when we want to share functions between different entry points and not want to pass a `const char* vuid` everywhere

### DIFF
--- a/layers/core_checks/cc_vuid_maps.cpp
+++ b/layers/core_checks/cc_vuid_maps.cpp
@@ -1,0 +1,570 @@
+/* Copyright (c) 2024 The Khronos Group Inc.
+ * Copyright (c) 2024 Valve Corporation
+ * Copyright (c) 2024 LunarG, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


layers/core_checks/core_validation.h
```diff
@@ -1026,12 +1026,12 @@ class CoreChecks : public ValidationStateTracker {
                                                                 const char* vuid) const;
 
     template <typename HandleT>
-    bool ValidateImageMipLevel(const HandleT handle, const vvl::Image& image_state, uint32_t mip_level, const Location& mip_loc,
-                               const char* vuid) const;
```
spencer-lunarg, line 1030: with these VUID maps, we can stop passing in `const char*` to functions like these... truly moving more logic until we know there is an actual error should also give a nice small perf boost

layers/core_checks/cc_vuid_maps.cpp
```diff
@@ -0,0 +1,572 @@
...
+    if (result.empty()) {
+        static const std::string unhandled("UNASSIGNED-CoreChecks-unhandled-mip-level");
+        return unhandled;
+    }
+    return result;
+}
+
+const std::string &GeteImageArrayLayerRangeVUID(const Location &loc) {
+    static const std::array<Entry, 20> errors{{
+        // static const std::array<Entry, 20> errors{
```
spencer-lunarg, line 540: remove

arno-lunarg (approved):
Nice change!!

PR #7675, opened 2024-03-12
layers: Linear Sampler check for MinMax reduction mode

spencer-lunarg:
The 280 spec updated ` VUID-vkCmdDraw-magFilter-04553`/` VUID-vkCmdDraw-mipmapMode-04770` and then added 

- `VUID-vkCmdDraw-magFilter-09598`
- `VUID-vkCmdDraw-mipmapMode-09599`

(ref https://gitlab.khronos.org/vulkan/vulkan/-/issues/3800)

PR #7706, opened 2024-03-18
layers: Fix crash from sampler pNext

spencer-lunarg:
Fix for https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7675#discussion_r1526178065 (cc @y-novikov)

... The fact this broken is concerning, I will make an issue and go through and make sure we are properly using safe struct to capture the pNext chain in all our state tracking

### DIFF
--- a/layers/state_tracker/sampler_state.h
+++ b/layers/state_tracker/sampler_state.h
@@ -18,6 +18,7 @@
  */
 #pragma once
 #include "state_tracker/state_object.h"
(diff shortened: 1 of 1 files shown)


arno-lunarg (approved):

y-novikov:
Thank you, the fix worked https://chromium-review.googlesource.com/c/angle/angle/+/5377890!

PR #7708, opened 2024-03-18
layers: Move SubresourceRangeErrorCodes to map file

spencer-lunarg:
Move another good candidate of VUIDs to `cc_vuid_maps.cpp`

These VUs use to be scatter around the place and now are all grouped together properly

### DIFF
--- a/layers/core_checks/cc_image.cpp
+++ b/layers/core_checks/cc_image.cpp
@@ -23,6 +23,7 @@
 #include <vulkan/vk_enum_string_helper.h>
 
 #include "core_validation.h"
+#include "cc_vuid_maps.h"
 #include "generated/chassis.h"
 #include "generated/pnext_chain_extraction.h"
 #include "error_message/error_strings.h"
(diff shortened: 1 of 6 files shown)


layers/core_checks/cc_image.cpp
```diff
@@ -1493,11 +1497,14 @@ bool CoreChecks::ValidateImageSubresourceRange(const uint32_t image_mip_count, c
                 uint64_t{subresourceRange.baseArrayLayer} + uint64_t{subresourceRange.layerCount};
 
             if (necessary_layer_count > image_layer_count) {
-                skip |= LogError(errorCodes.layer_count_err, objlist, subresource_loc.dot(Field::baseArrayLayer),
+                const auto vuid = image_layer_count_var == Field::depth
```
spencer-lunarg, line 1500: this VU is just funky because of how the `maintenance1` extension was designed

artem-lunarg (approved):

PR #7710, opened 2024-03-18
scripts: Create union of pointer group

spencer-lunarg:
There is a future Extension that will have a union similar to the `VkDescriptorDataEXT` that holds a pointer depending on the `type` passed in

To both make the future PR easier and also improve the "why do we do things on this specific struct" logic in the code, I made this groupining

### DIFF
--- a/scripts/generators/safe_struct_generator.py
+++ b/scripts/generators/safe_struct_generator.py
@@ -65,6 +65,14 @@ class SafeStructOutputGenerator(BaseGenerator):
             'AHardwareBuffer',
         ]
 
+        # These 'data' union are decided by the 'type' in the same parent struct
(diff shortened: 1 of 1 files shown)


scripts/generators/safe_struct_generator.py
```diff
@@ -65,6 +65,14 @@ def __init__(self):
             'AHardwareBuffer',
         ]
 
+        # These 'data' union are decided by the 'type' in the same parent struct
+        self.union_of_pointer = [
```
arno-lunarg, line 69: Nit: 
```suggestion
        self.union_of_pointers = [
```

arno-lunarg (approved):

PR #7713, opened 2024-03-18
Fixing Encode2D/3D address calculation for compressed textures.

RamiHg:
We were getting sync write-after-write errors when copying non-overlapping blocks of 320x320 pixels from a buffer to an ETC2 RGB texture. 320x320 is clearly a multiple of the block size (4x4) - so this should not fail validation.

I was able to track the problem down to `ImageRangeEncoder::Encode2D`. It was computing the x-offset into the row by multiplying `offset.x` with `texel_Sizes_[aspect_index`]. That is incorrect. Imagine an 8x8 ETC2 RGB (or BC1) texture. Each 4x4 block is 8 bytes. The address of `(x = 4, y = 0)` is the address of the second block, 8. However, the old code computed it as `offset.x * texel_sizes_[aspect_index] = 4 * 0.5 = 2`.

The problem was that the rest of the code was correctly computing y-offsets and spans. For example, the span of a 4x4 copy is correctly computed as 8. This created a false-positive overlap if another copy happened to be exactly adjacent, because the base address was __inside__ the region of the previous copy (due to the incorrect x offset calculation).

This change correctly computes the x-offset by taking into account the stride of each block. In the example above, the x-offset at `(x = 4, y = 0)` is now correct: 8.

I also did the same change in Encode3D. 

I created a new unit test. I confirmed that the unit test fails without this change. I also ran all the Windows/Android unit tests.

### DIFF
--- a/layers/containers/subresource_adapter.cpp
+++ b/layers/containers/subresource_adapter.cpp
@@ -358,13 +358,19 @@ ImageRangeEncoder::ImageRangeEncoder(const vvl::Image& image, const AspectParame
 IndexType ImageRangeEncoder::Encode2D(const VkSubresourceLayout& layout, uint32_t layer, uint32_t aspect_index,
                                       const VkOffset3D& offset) const {
     assert(offset.z == 0U);
+    // The address offset of the beginning of offset.x's block is:
+    // floor(offset.x * / texel_extent_.width) * texel_extent_.width * texel_extent_.height * texel_sizes_[apsect].
+    // Since offset.x must be a multiple of texel_extent_.width, we can simplify the formula by canceling out texel_extent_.width.
+    double xSize = static_cast<double>(texel_extent_.height * texel_sizes_[aspect_index]);
     return layout.offset + layer * layout.arrayPitch + offset.y * layout.rowPitch +
-           (offset.x ? static_cast<IndexType>(floor(offset.x * texel_sizes_[aspect_index])) : 0U);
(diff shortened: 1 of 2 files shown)


artem-lunarg:
@RamiHg Thanks for providing a solution with a test! I'm going to look through the subresource adapter code since I did not wrote it and need to familiarize myself. Meanwhile, I have one suggestion. Could you update the test to make it positive (put it into sync_val_positive.cpp and use PositiveSyncVal) and replace command buffer part with this one:
```
m_commandBuffer->begin();
vk::CmdCopyBufferToImage(*m_commandBuffer, src_buffer, dst_image, VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL, 1, &buffer_copy[0]);
vk::CmdCopyBufferToImage(*m_commandBuffer, src_buffer, dst_image, VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL, 1, &buffer_copy[1]);
m_commandBuffer->end();
```

It's two copies instead of three and this signals intent well - two non overlapping copies should not result in WRITE-AFTER-WRITE hazard. This version works with your fix and fails with the old code.

One more suggestion is to use `optimalTilingFeatures` instead of `linearTilingFeatures`. The `dst_image` is created with optimal tiling (default settings).

artem-lunarg (approved):
LGTM. After self-education session I can see why this fix is needed. 
@RamiHg If you don't see downsides I'm going to update the test as was suggested above in a separate PR and will merge this fix now.

Issue #7707, opened 2024-03-18
investigate safe struct not being used and pNext being dropped


spencer-lunarg:
from https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7706

We need a better system to make sure safe structs are both being used and make sure all the state tracking dose it the same way

PR #7714, opened 2024-03-19
Safe Struct CreatInfo rework

spencer-lunarg:
This closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7707

> Note: no "logic" was modified, this is just a lot renaming for consistency and changing things to be a const ref of a safe_struct

We have `safe_Vk*CreateInfo` structs that are used to hold things like the pNext chains, but when using, people really just want a "normal" struct

We had zero consistency with this, and it lead to preventable bug 

This change unifies all `StateObject` classes to look like

```c++
const safe_VkBufferCreateInfo safe_create_info;
const VkBufferCreateInfo &create_info;

// when constructing
safe_create_info(pCreateInfo),
create_info(*safe_create_info.ptr()),
```

doing this ensures everyone trying to pNext look into a create info is rested assure it will work

### DIFF
--- a/layers/state_tracker/cmd_buffer_state.h
+++ b/layers/state_tracker/cmd_buffer_state.h
@@ -616,6 +616,7 @@ class CommandBuffer : public RefcountedStateObject {
     void BindShader(VkShaderStageFlagBits shader_stage, vvl::ShaderObject *shader_object_state);
 
     bool IsPrimary() const { return createInfo.level == VK_COMMAND_BUFFER_LEVEL_PRIMARY; }
+    bool IsSeconary() const { return createInfo.level == VK_COMMAND_BUFFER_LEVEL_SECONDARY; }
     void BeginLabel(const char *label_name);
     void EndLabel();
     int LabelStackDepth() const { return label_stack_depth_; }
--- a/layers/state_tracker/cmd_buffer_state.h
+++ b/layers/state_tracker/cmd_buffer_state.h
@@ -72,8 +72,8 @@ class Event : public StateObject {
     // Queue that signaled this event. It's null if event was signaled from the host
     VkQueue signaling_queue = VK_NULL_HANDLE;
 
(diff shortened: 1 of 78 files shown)


layers/state_tracker/cmd_buffer_state.h
```diff
@@ -56,6 +56,19 @@ static bool GetMetalExport(const VkEventCreateInfo *info) {
 }
 #endif  // VK_USE_PLATFORM_METAL_EXT
 
+// Only CoreChecks uses this, but the state tracker stores it.
```
spencer-lunarg, line 59: moved this here so there is only 1 `namespace vvl {`

layers/state_tracker/cmd_buffer_state.h
```diff
@@ -137,7 +130,7 @@ class CommandBuffer : public RefcountedStateObject {
     using ImageLayoutMap = vvl::unordered_map<VkImage, LayoutState>;
     using AliasedLayoutMap = vvl::unordered_map<const GlobalImageLayoutRangeMap *, std::shared_ptr<ImageSubresourceLayoutMap>>;
 
-    VkCommandBufferAllocateInfo createInfo = {};
+    VkCommandBufferAllocateInfo allocate_info;
```
spencer-lunarg, line 133: this has burnt me before, you don't "create" a command buffer

layers/state_tracker/query_state.h
```diff
@@ -79,7 +80,8 @@ class QueryPool : public StateObject {
         return QUERYSTATE_UNKNOWN;
     }
 
-    const VkQueryPoolCreateInfo createInfo;
```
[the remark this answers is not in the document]
spencer-lunarg: correction, the checks were in `PreCallValidateCreateQueryPool` so the pNext would have been ok

arno-lunarg (approved):

PR #7044, opened 2023-11-29
layers: Track VK_ERROR_DEVICE_LOST

spencer-lunarg:
First attempt at https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6310

Can reproduce the issue, but need to figure out a more automated way to find all objects at all instances of functions returning `VK_ERROR_DEVICE_LOST` and resetting things

Issue #6310, opened 2023-08-17
VVL miss a vulkan device lost check when reporting VUID-vkDestroy* error messages


HuYuxin:
**Environment:**
 - OS: Android
 - GPU: Arm
 - SDK or header version if building from repo: sdk-1.3.261. git hash: e6bdb8d71409a96a4174589ea195d0dc1e920625
 - Options enabled (synchronization, best practices, etc.): N/A

**Describe the Issue**

A [clear and concise description](https://en.wikipedia.org/wiki/Minimal_reproducible_example) of what the bug is.

When VVL processes `vkDestroy*()` calls, it doesn't seem to cover the case when the vulkan device is lost and the vulkan resources can't finish execution on the GPU due to device loss.

To reproduce:

1. Follow [ANGLE Development Setup](https://github.com/google/angle/blob/main/doc/DevSetup.md#development-setup) to get ANGLE source code.

2. Follow [Setting up the ANGLE build for Android](https://github.com/google/angle/blob/main/doc/DevSetupAndroid.md#setting-up-the-angle-build-for-android) to download Android build dependency and set up GN args for building Android target. Make sure the vulkan validation layer is enabled by adding below line to the GN arg:

```
angle_enable_vulkan_validation_layers = true
```
3. Run this dEQP test from ANGLE repo root on an Android device with Android build version of 13 or above:

```
out/Android/angle_deqp_egl_tests --gtest_filter=dEQP?EGL.functional.robustness.reset_context.shaders.out_of_bounds_non_robust.reset_status.writes.local_array.fragment --verbose --local-output --num-retries=0 --skip-clear-data
```

4. Observe that the test failed with below VVL error messages:

```
[ VUID-vkDestroyFence-fence-01120 ] Validation Error: [ VUID-vkDestroyFence-fence-01120 ] Object 0: handle = 0x280000000028, type = VK_OBJECT_TYPE_FENCE; | MessageID = 0x5d296248 | VkFence 0x280000000028[] is in use. The Vulkan spec states: All queue submission commands that refer to fence must have completed execution

[ VUID-vkDestroyPipeline-pipeline-00765 ] Validation Error: [ VUID-vkDestroyPipeline-pipeline-00765 ] | MessageID = 0x6bdce5fd | Cannot call vkDestroyPipeline on VkPipeline 0x200000000020[] that is currently in use by a command buffer. The Vulkan spec states: All submitted commands that refer to pipeline must have completed execution

[...]

PR #7715, opened 2024-03-19
layers: Don't check for used object if Device Lost

spencer-lunarg:
second attempt of https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7044 to address https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6310

- Generate checks for when a function can return `VK_ERROR_DEVICE_LOST` and sets the status
- When destroying objects, all go through `ValidateObjectNotInUse` and there we can simply not check

The test uses a special hook into MockICD to force a device lost and can confirm the cleanup is as expected

### DIFF
--- a/scripts/known_good.json
+++ b/scripts/known_good.json
@@ -136,7 +136,7 @@
             "sub_dir": "Vulkan-Tools",
             "build_dir": "Vulkan-Tools/build",
             "install_dir": "Vulkan-Tools/build/install",
-            "commit": "v1.3.280",
+            "commit": "934b5f7c13374949527b71b9732b93b5bc0fcc3e",
             "build_step": "skip",
             "optional": [
                 "tests"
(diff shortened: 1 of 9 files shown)


scripts/known_good.json
```diff
@@ -136,7 +136,7 @@
             "sub_dir": "Vulkan-Tools",
             "build_dir": "Vulkan-Tools/build",
             "install_dir": "Vulkan-Tools/build/install",
-            "commit": "v1.3.280",
+            "commit": "934b5f7c13374949527b71b9732b93b5bc0fcc3e",
```
artem-lunarg, line 139: what does it do?

spencer-lunarg: this adds the newest Vulkan-Tools that has the MockICD that the new test will be able to run

https://github.com/KhronosGroup/Vulkan-Tools/pull/971

PR #7717, opened 2024-03-19
layers: Use new safe struct code from VUL

jeremyg-lunarg:


tests/unit/other_positive.cpp
```diff
@@ -363,50 +362,6 @@ TEST_F(VkPositiveLayerTest, ExtensionXmlDependsLogic) {
 }
 
 // https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5112
```
spencer-lunarg, line 364: nit: will need to remove this after

layers/state_tracker/pipeline_state.h
```diff
@@ -76,49 +80,17 @@ class PipelineCache : public StateObject {
 };
 
 class Pipeline : public StateObject {
-  public:
```
spencer-lunarg, line 79: What this PR shows is moving safe struct wasn't hard... except for Pipeline State, it is awful here and will need to be fixed moving forward as we all know

spencer-lunarg (approved):

tests/unit/other_positive.cpp
spencer-lunarg: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7789

PR #7718, opened 2024-03-20
Consistent use of ValidationStateTracker

spencer-lunarg:
```
const ValidationStateTracker* state_tracker;
const ValidationStateTracker* state_data;
const ValidationStateTracker* dev;
const ValidationStateTracker* dev_data;
const ValidationStateTracker* device_data;
const ValidationStateTracker* state;
```

are all things I have found in the code

this PR does two things

1. Makes them all `validator`
2. Makes them reference instead of pointers consistently (where possible)

There are some more core changes I would like to make to the chassis code, but first there is a need to clean up this stuff so it is easy to understand what is going on everywhere

### DIFF
--- a/layers/best_practices/best_practices_utils.cpp
+++ b/layers/best_practices/best_practices_utils.cpp
@@ -50,10 +50,11 @@ WriteLockGuard BestPractices::WriteLock() {
 std::shared_ptr<vvl::CommandBuffer> BestPractices::CreateCmdBufferState(VkCommandBuffer handle,
                                                                         const VkCommandBufferAllocateInfo* pCreateInfo,
                                                                         const vvl::CommandPool* pool) {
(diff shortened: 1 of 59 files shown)


artem-lunarg (approved):
LGTM. So nice to get rid of dev/dev_data.

There is also this idea that there can be a value in`state_tracker` naming since state tracker is more about collecting state and less about validation (and `validator` for types that support validation). If we ever need to refactor to make more separation between validation and state tracking different naming can also help. But that's a question we can discuss later, it's already a lot of good changes. Also it's not completely clear where we have places with pure state tracking or when it's mixed with validation.

jeremyg-lunarg:
It would also be nice to split up this object into instance and device objects, as I've mentioned before. I think the dev/dev_data names were reminders that we were working with the device and not the instance.

spencer-lunarg:
closing, thinking more about what I want, going to try a different approach to this

PR #7721, opened 2024-03-20
tests: Simplify test and make it positive

artem-lunarg:
Follow-up to https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7713
It's two copies instead of three that also reproduces the issue if the above fix is not applied.

### DIFF
--- a/tests/unit/sync_val.cpp
+++ b/tests/unit/sync_val.cpp
@@ -4184,7 +4184,7 @@ TEST_F(NegativeSyncVal, TestInvalidExternalSubpassDependency) {
     m_errorMonitor->VerifyFound();
 }
 
-TEST_F(NegativeSyncVal, TestCopyingToCompressedImage) {
+TEST_F(NegativeSyncVal, CopyToCompressedImage) {
     TEST_DESCRIPTION("Copy from uncompressed to compressed image with and without overlap.");
 
     AddOptionalExtensions(VK_KHR_COPY_COMMANDS_2_EXTENSION_NAME);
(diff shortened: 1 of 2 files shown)


tests/unit/sync_val_positive.cpp
```diff
@@ -1671,3 +1671,39 @@ TEST_F(PositiveSyncVal, ThreadedSubmitAndFenceWaitAndPresent) {
     }
     thread.join();
 }
+
+// https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7713
+TEST_F(PositiveSyncVal, TestCopyingBufferToCompressedImage) {
```
spencer-lunarg, line 1676:
```suggestion
TEST_F(PositiveSyncVal, CopyingBufferToCompressedImage) {
```

tests/unit/sync_val.cpp
```diff
@@ -4292,45 +4292,6 @@ TEST_F(NegativeSyncVal, TestCopyingToCompressedImage) {
     }
 }
 
-TEST_F(NegativeSyncVal, TestCopyingBufferToCompressedImage) {
```
spencer-lunarg, line 4295: not worth keeping this one?

artem-lunarg: Thought about that, did not find enough value.

artem-lunarg: I tested different combinations of that code if that can bring new scenario but did not find. A bigger problem is that additional copy makes harder to understand what is tested, and positive test provides the same scenario but in unambiguous way!

tests/unit/sync_val_positive.cpp
artem-lunarg: Updated

spencer-lunarg (approved):

PR #7723, opened 2024-03-20
gpu: temporary removing GPU-AV NV RT acceleration structures

arno-lunarg:
Validation was only done for the Nvidia extension, which has essentially been replaced by the KHR one, so the cost of maintaining that part of the code base (changes to the logging system have to be propagated here) is not worth it

While working on the GPU-AV refactoring, it is harder to retro-fit this logic, and would be 10x easier to re-write after when the GPU-AV code base in a better state

Making a PR to make sure we can easily search and reference this

As of Vulkan 1.3.280 there is a [VK_NV_ray_tracing_validation](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_NV_ray_tracing_validation.html) which likely will work better then anything that can be done at a Layer Level


jeremyg-lunarg:
This isn't deprecating it, it is removing it.  Deprecating would be adding a LOGW that says "we're going to remove this in N SDK releases...". 

I'm not against removing it if the KHR validation is at least equivalent than what is being removed. But we probably ought to discuss this outside of code review to be sure.

spencer-lunarg (approved):

PR #7724, opened 2024-03-20
layers: Make more use of const ref

spencer-lunarg:
Found some spots using pointers that could be using a constant reference to the object. This leads to better understanding if we expect pointers to possibly be null or not in a function

### DIFF
--- a/layers/core_checks/cc_ray_tracing.cpp
+++ b/layers/core_checks/cc_ray_tracing.cpp
@@ -1438,20 +1438,20 @@ bool CoreChecks::PreCallValidateCmdWriteAccelerationStructuresPropertiesNV(
     return skip;
 }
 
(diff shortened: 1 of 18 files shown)


layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -1485,19 +1485,18 @@ bool CoreChecks::PreCallValidateCmdCopyAccelerationStructureKHR(VkCommandBuffer
     auto cb_state = GetRead<vvl::CommandBuffer>(commandBuffer);
     assert(cb_state);
     skip |= ValidateCmd(*cb_state, error_obj.location);
-    if (pInfo) {
```
spencer-lunarg, line 1488: VUs like `VUID-vkCmdCopyAccelerationStructureKHR-pInfo-parameter` catch this and this function will never be called

artem-lunarg (approved):
LGTM *&->!

PR #7725, opened 2024-03-20
layers: Remove unused file sparse_containers.h

jeremyg-lunarg:


artem-lunarg (approved):

jeremyg-lunarg:
Internal CI failure is an android test. Overriding since this change would show up as a compile error.

PR #7671, opened 2024-03-10
layers: Add validation for vkGetDescriptorSetLayoutSupport with mutable descriptors

ziga-lunarg:
Closes #7665

PR #7726, opened 2024-03-21
Add proper vkGetDescriptorSetLayoutSupport support

spencer-lunarg:
follow-up of https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7671

For Stateless/ObjectTracking/CoreChecks we now just have `vkGetDescriptorSetLayoutSupport` and `vkCreateDescriptorSetLayout` just both call into a `ValidateDescriptorSetLayoutCreateInfo` call as both share the same VUs for `VkDescriptorSetLayoutCreateInfo`

### DIFF
--- a/layers/object_tracker/object_tracker_utils.cpp
+++ b/layers/object_tracker/object_tracker_utils.cpp
@@ -656,9 +656,7 @@ bool ObjectLifetimes::PreCallValidateCreateDescriptorSetLayout(VkDevice device, 
                                                                const ErrorObject &error_obj) const {
     bool skip = false;
     // Checked by chassis: device: "VUID-vkCreateDescriptorSetLayout-device-parameter"
-    if (pCreateInfo) {
-        skip |= ValidateDescriptorSetLayoutCreateInfo(*pCreateInfo, error_obj.location.dot(Field::pCreateInfo));
-    }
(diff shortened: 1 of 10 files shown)


layers/object_tracker/object_tracker_utils.cpp
```diff
@@ -656,9 +656,7 @@ bool ObjectLifetimes::PreCallValidateCreateDescriptorSetLayout(VkDevice device,
                                                                const ErrorObject &error_obj) const {
     bool skip = false;
     // Checked by chassis: device: "VUID-vkCreateDescriptorSetLayout-device-parameter"
-    if (pCreateInfo) {
```
artem-lunarg, line 659: I wonder if we need to forbid disabling layers down the "logical" layer stack. If lifetimes or core is enabled then you then you should not be allowed to disable stateless. Enforce this in UI logic too, in vkconfig you can disable stateless and have lifetimes.

artem-lunarg (approved):

layers/object_tracker/object_tracker_utils.cpp
spencer-lunarg: I guess the idea is by default, stateless is on

I guess what we should do is have a warning that if you turn it off, basic things like these pointers are assumed valid

I can take an action to figure out a good way to do this... in general I would like to better understand how we can improve UX for people using VVL on VkConfig

PR #7733, opened 2024-03-21
layers: Mention the actual version number of the device

tintou:
Previously, the validation message could lead to something like:

> pProperties->pNext includes a pointer to a VkStructureType (VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VULKAN_1_3_PROPERTIES)
> which was added in VK_API_VERSION_1_3 but the current effective API version is 1.3.0 (0x00403000)

Use the actual version it is comparing with in the validation message.

Also took the opportunity to format the version a bit less verbose.


spencer-lunarg:
@tintou thanks for helping improve the error message!

You will need to re-generate the code and check that in as well (CI will yell soon about this)
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/docs/generated_code.md

----

What I use personally

```
python3 scripts/generate_source.py external/Vulkan-Headers/registry/ external/SPIRV-Headers/include/spirv/unified1/ --target stateless_validation_helper.cpp
```

tintou:
@spencer-lunarg change added!

spencer-lunarg (approved):
LTGM - @aqnuep can you quick confirm this version naming logic won't wreck anything downstream in Vulkan SC

arno-lunarg (approved):

aqnuep:
> LTGM - @aqnuep can you quick confirm this version naming logic won't wreck anything downstream in Vulkan SC

Thanks, for pinging me. I'd probably keep `VK_API_VERSION_1_x` in the messages instead of saying version 1.x. The former is unambiguous, but the latter is just confusing because it does not contain information about which API variant does the version number apply to.

Alternatively, you could also use `%s` -> `StringAPIVersion(VK_API_VERSION_1_x)` for these cases to output the API version of the relevant API variant.

aqnuep:
If we really want to include version number literals in the message strings, at least spell it out as "Vulkan version 1.1", instead of just "version 1.1" to make this clear.

For future-proofing, especially if Vulkan would have other variants, for example, I think VVL should consider also parameterizing the API name itself, like we did with the version through the customizable output of `StringAPIVersion`, but that's just a suggestion that may be worth considering.

spencer-lunarg (changes requested):
as per @aqnuep comments

tintou:
alright, I've kept it formatted as VK_API_VERSION_1_x

aqnuep:
Thank you!

FYI, these particular messages would never get triggered on Vulkan SC, as Vulkan SC 1.0 is based on Vulkan 1.2, my notes were more about general policy when dealing with such messages.

spencer-lunarg (approved):
Happy when CI is

PR #7736, opened 2024-03-21
layers: Store submit reference instead of SemOp in timeline

artem-lunarg:
Before starting with some core sync bugs, I decided it was a good time to make a change I had been thinking about for some time. I feel this makes concepts a bit more orthogonal and hopefully helps to dig faster into the code.

It is a refactor to store only submission references instead of SemOps inside the semaphore's timeline. The main gain is that a submission reference is a simpler abstraction than SemOp. As a bonus, this avoids some data duplication and the need to keep asserts to ensure that duplicated data is consistent.

SemOps are continued to be used outside the semaphore's timeline (completed/last operation, other contexts). 

We can avoid storing `payload` and `op_type` values inside timeline because:
* payload value is already known when we access semaphore timeline - it is the key of the timeline map to get corresponding TimePoint object.
* op_type is determined by the TimePoint's variable that stores information: signal_op, wait_ops, acquire.

Each refactor step has its commit with comments, some commits are more challenging than others, might be good to keep this way in case of regression.

That's a minimal change to get it work, some more clean up is needed, which I'd like to make separately: rename some variables from ops to refs, simplify some signatures to pass SubmissionReference instead of (queue, seq), check if there are places that don't require null check anymore.

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -244,6 +244,7 @@ vvl_sources = [
   "layers/state_tracker/state_object.h",
   "layers/state_tracker/state_tracker.cpp",
   "layers/state_tracker/state_tracker.h",
+  "layers/state_tracker/submission_reference.h",
   "layers/state_tracker/vertex_index_buffer_state.h",
   "layers/state_tracker/video_session_state.cpp",
   "layers/state_tracker/video_session_state.h",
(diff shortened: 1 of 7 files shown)


jeremyg-lunarg:
This looks good and I'm glad that making things more correct is also making them simpler.

Have you tried this with any real games or traces yet?

artem-lunarg:
I did not try, but that's a good idea and I will check CS2 and its trace, or if you have some traces in mind let me know.

artem-lunarg:
No errors CS2 game and the same errors replaying cs2-noReflex-20240304T163412 as on the main.

jeremyg-lunarg (approved):
LGTM

Issue #4136, opened 2022-05-19
layer_chassis_dispatch.cpp attempts to unwrap output handles


pdaniell-nv:
As discussed in https://gitlab.khronos.org/Tracker/vk-gl-cts/-/issues/3685#note_360445, it appears some of the generated functions in layer_chassis_dispatch.cpp are treating parameters that are meant to output a handle as inputs, and attempts to Unwrap them, when it should be doing WrapNew().

It appears the bug is in https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/c935ec7efa2e37dee67995b717a908535c713355/scripts/layer_chassis_dispatch_generator.py#L2077 where there needs to be an exception for "GetDrmDisplayEXT" and possibly "GetWinrtDisplayNV". I didn't check all functions exhaustively, so there may be others.

edit: CTS 

```
dEQP-VK.wsi.direct_drm.*
dEQP-VK.api.image_compression_control.swapchain.direct_drm.*
dEQP-VK.image.swapchain_mutable.direct_drm.*
dEQP-VK.protected_memory.interaction.wsi.direct_drm.*
```

ncesario-lunarg:
Thanks @pdaniell-nv, it seems like some additional annotation in the xml would be beneficial, but I'm not sure how realistic it would be to get something like an `in`/`out`/`inout` attribute added to each command parameter.

FYI @mikes-lunarg as he probably understands the code generation best.

mikes-lunarg:
Yeah, treating Get*Display* as "create" calls is probably the quickest fix. I'm a little surprised we don't just infer in/out from constness. It would probably be a worthwhile exercise to compare the current `iscreate` list with a list of all entry points that take a non-const handle pointer.

ncesario-lunarg:
@mikes-lunarg yes, that makes sense. I tried this out in #4138 and it seems to have the desired effect, but not sure if it's 100% correct (that code is still more mysterious to me than it should be).

math4origami:
Hi, I'm the one who opened the original issue in CTS.

To resolve this issue, I think we'll need we need to add this around here:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/scripts/thread_safety_generator.py#L1344

```
void ThreadSafety::PostCallRecordGetDrmDisplayEXT(
    VkPhysicalDevice                            physicalDevice,
    int32_t                                     drmFd,
    uint32_t                                    connectorId,
[...]

PR #7740, opened 2024-03-22
Improve Layer Chassis Dispatch script

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4136

I have gone over all the various edge cases in the `layer_chassis_dispatch_generator.py` and feel confident it is in a good state now

### DIFF
--- a/layers/layer_chassis_dispatch_manual.cpp
+++ b/layers/layer_chassis_dispatch_manual.cpp
@@ -902,7 +902,7 @@ VkResult DispatchGetPhysicalDeviceDisplayPropertiesKHR(VkPhysicalDevice physical
     if (!wrap_handles) return result;
     if ((result == VK_SUCCESS || result == VK_INCOMPLETE) && pProperties) {
         for (uint32_t idx0 = 0; idx0 < *pPropertyCount; ++idx0) {
-            pProperties[idx0].display = layer_data->MaybeWrapDisplay(pProperties[idx0].display, layer_data);
+            pProperties[idx0].display = layer_data->MaybeWrapDisplay(pProperties[idx0].display);
         }
     }
     return result;
@@ -916,8 +916,7 @@ VkResult DispatchGetPhysicalDeviceDisplayProperties2KHR(VkPhysicalDevice physica
     if (!wrap_handles) return result;
     if ((result == VK_SUCCESS || result == VK_INCOMPLETE) && pProperties) {
         for (uint32_t idx0 = 0; idx0 < *pPropertyCount; ++idx0) {
-            pProperties[idx0].displayProperties.display =
-                layer_data->MaybeWrapDisplay(pProperties[idx0].displayProperties.display, layer_data);
+            pProperties[idx0].displayProperties.display = layer_data->MaybeWrapDisplay(pProperties[idx0].displayProperties.display);
         }
     }
     return result;
@@ -932,7 +931,7 @@ VkResult DispatchGetPhysicalDeviceDisplayPlanePropertiesKHR(VkPhysicalDevice phy
     if ((result == VK_SUCCESS || result == VK_INCOMPLETE) && pProperties) {
         for (uint32_t idx0 = 0; idx0 < *pPropertyCount; ++idx0) {
             VkDisplayKHR &opt_display = pProperties[idx0].currentDisplay;
-            if (opt_display) opt_display = layer_data->MaybeWrapDisplay(opt_display, layer_data);
+            if (opt_display) opt_display = layer_data->MaybeWrapDisplay(opt_display);
         }
     }
     return result;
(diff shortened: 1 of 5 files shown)


layers/layer_chassis_dispatch_manual.cpp
```diff
@@ -340,60 +340,6 @@ void DispatchDestroyRenderPass(VkDevice device, VkRenderPass renderPass, const V
     layer_data->renderpasses_states.erase(renderPass);
 }
 
-VkResult DispatchCreateSwapchainKHR(VkDevice device, const VkSwapchainCreateInfoKHR *pCreateInfo,
```
spencer-lunarg, line 343: these are handled correctly now in the code gen, they were on the top of the "no autogen" list so probably added well before the script was improved over the years

layers/layer_chassis_dispatch_manual.cpp
```diff
@@ -902,7 +848,7 @@ VkResult DispatchGetPhysicalDeviceDisplayPropertiesKHR(VkPhysicalDevice physical
     if (!wrap_handles) return result;
     if ((result == VK_SUCCESS || result == VK_INCOMPLETE) && pProperties) {
         for (uint32_t idx0 = 0; idx0 < *pPropertyCount; ++idx0) {
-            pProperties[idx0].display = layer_data->MaybeWrapDisplay(pProperties[idx0].display, layer_data);
+            pProperties[idx0].display = layer_data->MaybeWrapDisplay(pProperties[idx0].display);
```
spencer-lunarg, line 851: cleaned up `MaybeWrapDisplay` logic, it is called from `layer_data` so didn't need it passed in as well

layers/layer_chassis_dispatch_manual.cpp
```diff
@@ -1006,18 +951,6 @@ VkResult DispatchGetDisplayModeProperties2KHR(VkPhysicalDevice physicalDevice, V
     return result;
 }
 
-VkResult DispatchEnumerateDeviceExtensionProperties(
```
spencer-lunarg, line 1009: generated the same, removed from no-autogen list

layers/layer_chassis_dispatch_manual.cpp
```diff
@@ -1110,84 +1043,6 @@ VkResult DispatchGetPhysicalDeviceToolProperties(VkPhysicalDevice physicalDevice
     return result;
 }
 
-bool NotDispatchableHandle(VkObjectType object_type) {
```
spencer-lunarg, line 1113: this is the only thing that SetPrivate/GetPrivate calls had different, this is not all auto-generated (since we have information about which handles are dispatchable now

layers/layer_chassis_dispatch_manual.cpp
```diff
@@ -1722,41 +1577,3 @@ VkResult DispatchReleasePerformanceConfigurationINTEL(
 
     return result;
 }
-
-#ifdef VK_USE_PLATFORM_XLIB_XRANDR_EXT
-VkResult DispatchGetRandROutputDisplayEXT(VkPhysicalDevice physicalDevice, Display* dpy, RROutput rrOutput,
```
spencer-lunarg, line 1727: these 3 `VkDisplayKHR` are now generated

scripts/generators/layer_chassis_dispatch_generator.py
```diff
@@ -33,15 +33,16 @@ def __init__(self):
             'vkDestroyInstance',
             'vkCreateDevice',
             'vkDestroyDevice',
-            'vkCreateSwapchainKHR',
-            'vkCreateSharedSwapchainsKHR',
+            # Need to handle Acquired swapchain image handles
```
spencer-lunarg, line 36: finally added a comment "why" we are not auto generating things

layers/vulkan/generated/chassis.h
```diff
@@ -2432,43 +2432,40 @@ class ValidationObject {
...
         unique_id = HashedUint64::hash(unique_id);
         assert(unique_id != 0);  // can't be 0, otherwise unwrap will apply special rule for VK_NULL_HANDLE
-        unique_id_mapping.insert_or_assign(unique_id, CastToUint64(newlyCreatedHandle));
+        unique_id_mapping.insert_or_assign(unique_id, CastToUint64(new_created_handle));
         return (HandleType)unique_id;
     }
 
-    // Specialized handling for VkDisplayKHR. Adds an entry to enable reverse-lookup.
-    VkDisplayKHR WrapDisplay(VkDisplayKHR newlyCreatedHandle, ValidationObject* map_data) {
+    // VkDisplayKHR objects are statically created in the driver at VkCreateInstance.
```
artem-lunarg, line 2453: Great comment, makes clear how the things work.

layers/vulkan/generated/chassis.h
```diff
@@ -2432,43 +2432,40 @@ class ValidationObject {
...
+        // See if this display is already known
+        auto it = display_id_reverse_mapping.find(handle);
+        if (it != display_id_reverse_mapping.end()) return (VkDisplayKHR)it->second;
+
+        // Unknown, so wrap
         auto unique_id = global_unique_id++;
         unique_id = HashedUint64::hash(unique_id);
-        unique_id_mapping.insert_or_assign(unique_id, CastToUint64(newlyCreatedHandle));
-        map_data->display_id_reverse_mapping.insert_or_assign(newlyCreatedHandle, unique_id);
+        unique_id_mapping.insert_or_assign(unique_id, CastToUint64(handle));
```
artem-lunarg, line 2464: Can we call `WrapNew` here to no repeat logic related to atomic update/hashing?

spencer-lunarg: so all I did was inline `WrapDisplay` in to `MaybeWrapDisplay`

but looking now, I see the only thing difference between `WrapDisplay` and `WrapNew` is it needs to call `display_id_reverse_mapping.insert_or_assign` after

... so yes, will update this to just call `WrapNew`

artem-lunarg (approved):

PR #7748, opened 2024-03-23
Remove broken shader object vuids 08875 and 08871

ziga-lunarg:
These VUs are broken, see here: https://gitlab.khronos.org/vulkan/vulkan/-/issues/3747

It is not possible to generate spirv from glsl that would follow these 2 vuids, so I think it's best to remove them until they're fixed.


spencer-lunarg (approved):

spencer-lunarg:
>  so I think it's best to remove them until they're fixed.

From looking at the internal gitlab, I agree, this should be fixed/confirmed in the spec first

If we need it back, we have this nice PR that has the test and layer logic

PR #7749, opened 2024-03-24
Move functions into DebugReportData

spencer-lunarg:
On the quest to improve logging, first need to do some house cleaning

This change

1. `typedef struct debug_report_data` -> `class DebugReportData`
2. Move things into `DebugReportData` instead of passing around a pointer reference to it

I imagine a lot of this stems back from when there were multiple layers, but instead of functions like `BeginCmdDebugUtilsLabel` being globals that pass in a `debug_report_data*`, I just moved these calls into the new `DebugReportData` class

### DIFF
--- a/layers/core_checks/cc_synchronization.cpp
+++ b/layers/core_checks/cc_synchronization.cpp
@@ -1949,9 +1949,8 @@ void CoreChecks::RecordBarriers(Func func_name, vvl::CommandBuffer &cb_state, co
 }
 
 template <typename TransferBarrier, typename Scoreboard>
-bool CoreChecks::ValidateAndUpdateQFOScoreboard(const debug_report_data *report_data, const vvl::CommandBuffer &cb_state,
-                                                const char *operation, const TransferBarrier &barrier, Scoreboard *scoreboard,
-                                                const Location &loc) const {
+bool CoreChecks::ValidateAndUpdateQFOScoreboard(const vvl::CommandBuffer &cb_state, const char *operation,
+                                                const TransferBarrier &barrier, Scoreboard *scoreboard, const Location &loc) const {
     // Record to the scoreboard or report that we have a duplication
     bool skip = false;
     auto inserted = scoreboard->emplace(barrier, &cb_state);
@@ -1993,7 +1992,7 @@ bool CoreChecks::ValidateQueuedQFOTransferBarriers(const vvl::CommandBuffer &cb_
                                    found->dstQueueFamilyIndex);
             }
         }
-        skip |= ValidateAndUpdateQFOScoreboard(report_data, cb_state, "releasing", release, &scoreboards->release, loc);
+        skip |= ValidateAndUpdateQFOScoreboard(cb_state, "releasing", release, &scoreboards->release, loc);
     }
     // Each acquire must have a matching release (ERROR)
     for (const auto &acquire : cb_barriers.acquire) {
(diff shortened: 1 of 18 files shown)


layers/error_message/logging.h
```diff
@@ -178,76 +178,30 @@ class TypedHandleWrapper {
     VulkanTypedHandle handle_;
 };
 
-typedef struct _debug_report_data {
+struct Location;
+
+class DebugReportData {
+  public:
     std::vector<VkLayerDbgFunctionState> debug_callback_list;
-    VkDebugUtilsMessageSeverityFlagsEXT active_severities{0};
```
spencer-lunarg, line 183: these I found are private as no one else touches them

layers/error_message/logging.h
```diff
@@ -260,67 +214,56 @@ typedef struct _debug_report_data {
...
+    void EndCmdDebugUtilsLabel(VkCommandBuffer command_buffer);
+    void InsertCmdDebugUtilsLabel(VkCommandBuffer command_buffer, const VkDebugUtilsLabelEXT *label_info);
+    void ResetCmdDebugUtilsLabel(VkCommandBuffer command_buffer);
+    void EraseCmdDebugUtilsLabel(VkCommandBuffer command_buffer);
 
-struct Location;
-VKAPI_ATTR bool LogMsg(const debug_report_data *debug_data, VkFlags msg_flags, const LogObjectList &objects, const Location *loc,
-                       std::string_view vuid_text, const char *format, va_list argptr);
+  private:
+    bool UpdateLogMsgCounts(int32_t vuid_hash) const;
```
spencer-lunarg, line 231: these were `static` but still required passing in a pointer, so just made them private functions

layers/error_message/logging.h
```diff
@@ -334,109 +277,3 @@ VKAPI_ATTR VkBool32 VKAPI_CALL MessengerWin32DebugOutputMsg(VkDebugUtilsMessageS
                                                             VkDebugUtilsMessageTypeFlagsEXT message_type,
                                                             const VkDebugUtilsMessengerCallbackDataEXT *callback_data,
                                                             void *user_data);
-
-template <typename Map>
-static LoggingLabelState *GetLoggingLabelState(Map *map, typename Map::key_type key, bool insert) {
```
spencer-lunarg, line 339: things like this I just moved, it is static, but everyone using it is in `logging.cpp` so no need to have it out in the header file

layers/error_message/logging.h
```diff
@@ -260,67 +214,56 @@ typedef struct _debug_report_data {
...
+    bool LogMsg(VkFlags msg_flags, const LogObjectList &objects, const Location *loc, std::string_view vuid_text,
+                const char *format, va_list argptr);
 
-template debug_report_data *GetLayerDataPtr<debug_report_data>(void *data_key,
-                                                               std::unordered_map<void *, debug_report_data *> &data_map);
+    void BeginQueueDebugUtilsLabel(VkQueue queue, const VkDebugUtilsLabelEXT *label_info);
+    void EndQueueDebugUtilsLabel(VkQueue queue);
+    void InsertQueueDebugUtilsLabel(VkQueue queue, const VkDebugUtilsLabelEXT *label_info);
 
-static inline void DebugReportFlagsToAnnotFlags(VkDebugReportFlagsEXT dr_flags, VkDebugUtilsMessageSeverityFlagsEXT *da_severity,
```
spencer-lunarg, line 268: this was just moved into `logging.cpp`

layers/core_checks/cc_synchronization.cpp
```diff
@@ -1949,9 +1949,8 @@ void CoreChecks::RecordBarriers(Func func_name, vvl::CommandBuffer &cb_state, co
 }
 
 template <typename TransferBarrier, typename Scoreboard>
-bool CoreChecks::ValidateAndUpdateQFOScoreboard(const debug_report_data *report_data, const vvl::CommandBuffer &cb_state,
```
spencer-lunarg, line 1952: `report_data` was unused here

layers/error_message/logging.h
```diff
@@ -178,76 +178,30 @@ class TypedHandleWrapper {
     VulkanTypedHandle handle_;
 };
 
-typedef struct _debug_report_data {
+struct Location;
+
+class DebugReportData {
```
artem-lunarg, line 183: nit: real nit about Data suffix, not sure if it's helpful.

artem-lunarg (approved):
LGTM

layers/error_message/logging.h
spencer-lunarg: it isn't.... let me fix it now as it will be too nit to ever touch again lol

PR #7750, opened 2024-03-24
Fix synchronization crash with dynamic rendering local read image barrier

ziga-lunarg:
I don't know if this needs to be handled differently, but at least it fixes the crash for now.


spencer-lunarg (approved):
> I don't know if this needs to be handled differently, but at least it fixes the crash for now.

Probably should, but having the test is amazing for when someone has time to look into it

PR #7753, opened 2024-03-25
layers: Add new vkCmdBeginTransformFeedbackEXT VU

spencer-lunarg:
`VUID-vkCmdBeginTransformFeedbackEXT-firstCounterBuffer-09630` was added in `1.3.281`

This might be what was causing some Transform Feedback tests to fail and be blacklisted in CI as the internal Gitlab issue showed that some driver crash and I never could figure out we were doing wrong in the tests

### DIFF
--- a/layers/core_checks/cc_cmd_buffer.cpp
+++ b/layers/core_checks/cc_cmd_buffer.cpp
@@ -1484,6 +1484,14 @@ bool CoreChecks::PreCallValidateCmdBeginTransformFeedbackEXT(VkCommandBuffer com
         }
     }
 
+    if ((counterBufferCount + firstCounterBuffer) > cb_state->transform_feedback_buffers_bound) {
+        skip |= LogError("VUID-vkCmdBeginTransformFeedbackEXT-firstCounterBuffer-09630", commandBuffer,
+                         error_obj.location.dot(Field::firstCounterBuffer),
+                         "is %" PRIu32 " and counterBufferCount is %" PRIu32
+                         " but vkCmdBindTransformFeedbackBuffersEXT only bound %" PRIu32 " buffers.",
+                         firstCounterBuffer, counterBufferCount, cb_state->transform_feedback_buffers_bound);
+    }
+
     // pCounterBuffers and pCounterBufferOffsets are optional and may be nullptr. Additionaly, pCounterBufferOffsets must be nullptr
     // if pCounterBuffers is nullptr.
     if (pCounterBuffers == nullptr) {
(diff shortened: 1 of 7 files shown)


layers/core_checks/cc_cmd_buffer.cpp
```diff
@@ -1484,6 +1484,14 @@ bool CoreChecks::PreCallValidateCmdBeginTransformFeedbackEXT(VkCommandBuffer com
         }
     }
 
+    if (counterBufferCount != 0 && !cb_state->transform_feedback_buffer_bound.test(firstCounterBuffer + counterBufferCount - 1)) {
+        skip |= LogError("VUID-vkCmdBeginTransformFeedbackEXT-firstCounterBuffer-09630", commandBuffer,
+                         error_obj.location.dot(Field::firstCounterBuffer),
+                         "is %" PRIu32 " and counterBufferCount is %" PRIu32
+                         " but vkCmdBindTransformFeedbackBuffersEXT never bound buffer %" PRIu32 ".",
+                         firstCounterBuffer, counterBufferCount, firstCounterBuffer + counterBufferCount);
```
arno-lunarg, line 1492:
```suggestion
                         firstCounterBuffer, counterBufferCount, firstCounterBuffer + counterBufferCount - 1);
```
(I think?)

arno-lunarg (approved):
LGTM! Maybe one issue with error message

layers/core_checks/cc_cmd_buffer.cpp
arno-lunarg: VUID-vkCmdBeginTransformFeedbackEXT-firstCounterBuffer-09630
The sum of firstCounterBuffer and counterBufferCount **must be less than or equal to the number of transform feedback buffers currently bound** by [vkCmdBindTransformFeedbackBuffersEXT](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCmdBindTransformFeedbackBuffersEXT.html)

So I think that `test`ing to see if `firstCounterBuffer + counterBufferCount - 1` is set is not the correct way to check?

arno-lunarg (changes requested):

layers/core_checks/cc_cmd_buffer.cpp
spencer-lunarg: there are VU to make sure you have can't bind `[2, 3]` and you need to bind buffer `[0, 1]`

So if I have bound `[0, 1]` that means I need to make sure that `firstCounterBuffer + counterBufferCount` is not greater than `2` so I guess what would be your counter proposal?

spencer-lunarg: @arno-lunarg I got clarifications and this was simpler than I realize

arno-lunarg (approved):

PR #7754, opened 2024-03-25
layers: Move some state tracking out of stateless

spencer-lunarg:
The core issue is we track the `subpass` state inside `StatelessValidation`

We currently are tracking state in `stateless_validation.h` and `StatelessValidation::RecordRenderPass`

My goal is to get rid of it, this PR removes half of it as the 2nd half will be a lot more messy to remove

### DIFF
--- a/layers/core_checks/cc_pipeline_graphics.cpp
+++ b/layers/core_checks/cc_pipeline_graphics.cpp
@@ -88,7 +88,7 @@ bool CoreChecks::ValidateGraphicsPipeline(const vvl::Pipeline &pipeline, const L
     skip |= ValidateGraphicsPipelineColorBlendState(pipeline, subpass_desc, create_info_loc);
     skip |= ValidateGraphicsPipelineRasterizationState(pipeline, subpass_desc, create_info_loc);
     skip |= ValidateGraphicsPipelineMultisampleState(pipeline, subpass_desc, create_info_loc);
-    skip |= ValidateGraphicsPipelineDepthStencilState(pipeline, create_info_loc);
+    skip |= ValidateGraphicsPipelineDepthStencilState(pipeline, subpass_desc, create_info_loc);
     skip |= ValidateGraphicsPipelineDynamicState(pipeline, create_info_loc);
     skip |= ValidateGraphicsPipelineFragmentShadingRateState(pipeline, create_info_loc);
     skip |= ValidateGraphicsPipelineDynamicRendering(pipeline, create_info_loc);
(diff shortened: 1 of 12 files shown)


layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -428,8 +427,8 @@ bool CoreChecks::ValidatePipelineLibraryCreateInfo(const vvl::Pipeline &pipeline
 
 // Using memcmp will fail sometimes as the alignment/padding in the struct can have undefined values, so because we can't ensure the
 // memory was zero-intialized, we need to just do a normal compare on each member
-static bool ComparePipelineMultisampleStateCreateInfo(VkPipelineMultisampleStateCreateInfo a,
-                                                      VkPipelineMultisampleStateCreateInfo b) {
+static bool ComparePipelineMultisampleStateCreateInfo(const VkPipelineMultisampleStateCreateInfo &a,
```
spencer-lunarg, line 430: honestly I am mad I didn't make these const ref before, we were definitely wasting time doing a copy here

layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -1298,7 +1298,7 @@ bool CoreChecks::ValidateGraphicsPipelineExternalFormatResolve(const vvl::Pipeli
         }
     } else {
         const uint64_t external_format = GetExternalFormat(pipeline.PNext());
-        const auto rendering_struct = vku::FindStructInPNextChain<VkPipelineRenderingCreateInfo>(pipeline.PNext());
+        const auto *rendering_struct = pipeline.rendering_create_info;
```
spencer-lunarg, line 1301: making use of the `rendering_create_info` we have

layers/state_tracker/pipeline_sub_state.h
```diff
@@ -202,7 +202,7 @@ struct FragmentOutputState : public PipelineSubState {
     std::unique_ptr<const safe_VkPipelineColorBlendStateCreateInfo> color_blend_state;
     std::unique_ptr<const safe_VkPipelineMultisampleStateCreateInfo> ms_state;
 
-    AttachmentVector attachments;
+    AttachmentStateVector attachment_states;
```
spencer-lunarg, line 205: `attachments` are `VkImage` in my brain (and the spec language) so made it clear these are the state info object, not the images themselves

layers/stateless/sl_pipeline.cpp
```diff
@@ -941,196 +934,21 @@ bool StatelessValidation::manual_PreCallValidateCreateGraphicsPipelines(
...
             }
 
-            const auto *rasterization_order_attachment_access_feature =
-                vku::FindStructInPNextChain<VkPhysicalDeviceRasterizationOrderAttachmentAccessFeaturesEXT>(device_createinfo_pnext);
             if (create_info.pDepthStencilState != nullptr && uses_depthstencil_attachment) {
                 const Location ds_loc = create_info_loc.dot(Field::pDepthStencilState);
                 auto const &ds_state = *create_info.pDepthStencilState;
                 skip |= ValidatePipelineDepthStencilStateCreateInfo(ds_state, ds_loc);
-
-                if ((ds_state.flags & VK_PIPELINE_DEPTH_STENCIL_STATE_CREATE_RASTERIZATION_ORDER_ATTACHMENT_DEPTH_ACCESS_BIT_EXT) !=
```
spencer-lunarg, line 960: the "core" change of this PR is moving this stuff from stateless to core checks since it actually does require state tracking

tests/unit/pipeline_advanced_blend.cpp
```diff
@@ -43,8 +43,12 @@ TEST_F(NegativePipelineAdvancedBlend, BlendOps) {
         color_blend_state.pAttachments = attachment_states;
         helper.gp_ci_.pColorBlendState = &color_blend_state;
     };
-    CreatePipelineHelper::OneshotTest(*this, set_info_different, kErrorBit,
-                                      "VUID-VkPipelineColorBlendAttachmentState-colorBlendOp-01406");
+
+    {
+        constexpr std::array vuids = {"VUID-VkGraphicsPipelineCreateInfo-renderPass-07609",
```
artem-lunarg, line 48: Why 07609 vuid was added here?

artem-lunarg (approved):
LGTM. One question.

tests/unit/pipeline_advanced_blend.cpp
spencer-lunarg: before `01406` was in stateless so it would never call `07609` in core checks

now they are both core checks, so both get triggered together

It was simpler doing this then re-writing the test to only hit `01406`

artem-lunarg: makes sense, thanks for the explanation

PR #7755, opened 2024-03-25
Clean up chassis_modification_state.h

spencer-lunarg:
This makes 2 simple, but really nice changes

1. Use the name of the struct instead of `void*` in the function parameters (this was `void*` before because we didn't have things separated and it would have required everyone to include everything)
2. Use `namespace chassis` for these structs instead of the strange names we had for them


jeremyg-lunarg (approved):
LGTM

PR #7756, opened 2024-03-25
layers: Fix arrayPitch computation for ASTC_10x10

artem-lunarg:
Fixes Angle regression https://ci.chromium.org/ui/p/angle/builders/try/linux-test/19351/test-results?sortby=&groupby=&cols=
Related conversation: https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/5df44678c4f4ec06dcad3eea308b1dd70eb8cc56#commitcomment-140093219

The history here is that we made a https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7713 PR recently, which is correct, but it exposes another existing issue. It did not cause troubles for Angle due to luck (two bugs compensated each other).

Our code assumed that rowPitch (bytes for one row of texels) is an integral number, which is true for most formats, including compressed ones, but it's not true for ASTC. For example, ASTC10x10, encodes 10x10=100 pixels in 16 bytes, so on average it's 1.6 bytes per row. Early rounding resulted in incorrect computation of a texture slice size (`arrayPitch`). The size of a 20x20 ASTC texture (4 blocks) was computed as 60 bytes instead of 16*4=64 bytes.

The ASTC addressing issue existed before https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7713 but was not covered by the tests. `CopyBufferToCompressedImageASTC` adds such a test. This will work as a regression test for 7713 fix for ASTC format.

`CopyBufferToCompressedImageASTC2` improves ASTC coverage, does not repro any issue.

`CopyBufferToCompressedImageASTC3` reproduces the issue that angle encountered after https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7713 and is fixed here.

This PR fixes `arrayPitch` computation. There is a similar `rowPitch` issue but it does not block anyone so far, so that fix is postponed.

### DIFF
--- a/layers/containers/subresource_adapter.cpp
+++ b/layers/containers/subresource_adapter.cpp
@@ -335,8 +335,15 @@ ImageRangeEncoder::ImageRangeEncoder(const vvl::Image& image, const AspectParame
                 }
             } else {
                 layout.offset += layout.size;
-                layout.rowPitch = static_cast<VkDeviceSize>(floor(subres_extent.width * texel_sizes_[aspect_index]));
-                layout.arrayPitch = layout.rowPitch * subres_extent.height;
+
+                const double row_pitch = subres_extent.width * texel_sizes_[aspect_index];
+                // TODO: layout.rowPitch is still computed incorrectly for ASTC_10x10,
+                // but it is less trivial fix comparing to arrayPitch, so will be fixed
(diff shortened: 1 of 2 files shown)


spencer-lunarg (approved):

spencer-lunarg:
Thanks for taking this on, the tests are really good too!!

PR #7758, opened 2024-03-25
layers: Avoid UB in ValidateStructTypeArray

HildarTheDorf:
The read of count in ValidateStructTypeArray can invoke UB if the application does something like this:

```c++
uint32_t num_format_props;
CHECK_RESULT(vkGetPhysicalDeviceVideoFormatPropertiesKHR(physical_device, &video_format_info, &num_format_props, nullptr));
```

By moving the count check after the nullptr check, we avoid reading a possibly uninitialized variable.

### DIFF
--- a/layers/stateless/stateless_validation.h
+++ b/layers/stateless/stateless_validation.h
@@ -200,7 +200,7 @@ class StatelessValidation : public ValidationObject {
                                  const char *stype_vuid, const char *param_vuid, const char *count_required_vuid) const {
         bool skip = false;
 
-        if ((count == 0) || (array == nullptr)) {
+        if ((array == nullptr) || (count == 0)) {
             skip |=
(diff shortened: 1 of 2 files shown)


layers/stateless/stateless_validation.h
```diff
@@ -200,7 +200,7 @@ class StatelessValidation : public ValidationObject {
                                  const char *stype_vuid, const char *param_vuid, const char *count_required_vuid) const {
         bool skip = false;
 
-        if ((count == 0) || (array == nullptr)) {
+        if ((array == nullptr) || (count == 0)) {
```
spencer-lunarg, line 203: I am happy with this, but grepping, I found `if ((count == 0) || (array == nullptr)) {` is used in 6 other places

It would be nice to update these everywhere at once so it is all unified

HildarTheDorf: Yeah, these all seem to follow the same pattern, good catch.

spencer-lunarg (approved):

PR #7760, opened 2024-03-26
Test failures

HildarTheDorf:
Few fixes for the test suite. A test that segfaults, a test that fails on my machine but shouldn't, and a test that works but I noticed doesn't check for a required feature.


spencer-lunarg (approved):

spencer-lunarg:
Thanks @HildarTheDorf - I am always happy to keep improving the tests to work on anyone's machine

PR #7762, opened 2024-03-26
gpuav: Split debug instructions and make a switch case

spencer-lunarg:
This fixes small bug were I was injecting a `OpName` after a `OpModuleProcessed`

### DIFF
--- a/layers/gpu_validation/spirv/module.cpp
+++ b/layers/gpu_validation/spirv/module.cpp
@@ -42,55 +42,87 @@ Module::Module(std::vector<uint32_t> words, uint32_t shader_id, uint32_t output_
         }
         auto new_inst = std::make_unique<Instruction>(it, instruction_count++);
 
-        SpvType spv_type = GetSpvType(new_inst->Opcode());
+        switch (opcode) {
+            case spv::OpCapability:
+                capabilities_.emplace_back(std::move(new_inst));
+                break;
+            case spv::OpExtension:
+                extensions_.emplace_back(std::move(new_inst));
+                break;
+            case spv::OpExtInstImport:
+                ext_inst_imports_.emplace_back(std::move(new_inst));
+                break;
+            case spv::OpMemoryModel:
+                memory_model_.emplace_back(std::move(new_inst));
+                break;
+            case spv::OpEntryPoint:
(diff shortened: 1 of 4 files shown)


layers/gpu_validation/spirv/module.cpp
```diff
@@ -42,55 +42,87 @@ Module::Module(std::vector<uint32_t> words, uint32_t shader_id, uint32_t output_
         }
         auto new_inst = std::make_unique<Instruction>(it, instruction_count++);
 
-        SpvType spv_type = GetSpvType(new_inst->Opcode());
-
-        if (opcode == spv::OpCapability) {
```
spencer-lunarg, line 47: this grew and expanded and better just being a switch case now, so that all I did

layers/gpu_validation/spirv/module.h
```diff
@@ -49,7 +49,9 @@ class Module {
     InstructionList memory_model_;
     InstructionList entry_points_;
     InstructionList execution_modes_;
-    InstructionList debug_infos_;
```
spencer-lunarg, line 52: image

i didn't realize that "debug" has 3 sub groups in it

layers/vulkan/generated/spirv_grammar_helper.h
```diff
@@ -839,39 +839,6 @@ static constexpr bool GroupOperation(uint32_t opcode) {
     }
 }
 
-static constexpr bool DebugOperation(uint32_t opcode) {
```
spencer-lunarg, line 842: this wasn't helpful because these are actually split up into 3

layers/vulkan/generated/spirv_grammar_helper.h
```diff
@@ -839,39 +839,6 @@ static constexpr bool GroupOperation(uint32_t opcode) {
...
-        case spv::OpLine:
-        case spv::OpNoLine:
-        case spv::OpModuleProcessed:
-            return true;
-        default:
-            return false;
-    }
-}
-
-static constexpr bool AnnotationOperation(uint32_t opcode) {
```
spencer-lunarg, line 859: this "seems" like it would be good to generate, but used one spot and there hasn't been a change to these instructions in the last 7 years, won't probably be a new one in the next 7 as well

PR #7763, opened 2024-03-26
More cleanup for cross chassis objects

spencer-lunarg:
- Makes everything `chassis_state` to better see what is coming from the chassis
- Make things const ref (assume was left over when we use to pass `void *` everywhere)
-  replace `using CreateShaderModuleStates = std::array<CreateShaderModule, 32>;` 
- Extract the `vvl::Pipeline` from things that don't need a per-layer check

### DIFF
--- a/layers/state_tracker/chassis_modification_state.h
+++ b/layers/state_tracker/chassis_modification_state.h
@@ -79,33 +79,31 @@ using ShaderModuleUniqueIds = std::unordered_map<VkShaderStageFlagBits, uint32_t
 
 struct CreateGraphicsPipelines {
     std::vector<safe_VkGraphicsPipelineCreateInfo> modified_create_infos;
-    std::vector<std::shared_ptr<vvl::Pipeline>> pipe_state;
     std::vector<ShaderModuleUniqueIds> shader_unique_id_maps;
     const VkGraphicsPipelineCreateInfo* pCreateInfos;
 };
 
 struct CreateComputePipelines {
     std::vector<safe_VkComputePipelineCreateInfo> modified_create_infos;
-    std::vector<std::shared_ptr<vvl::Pipeline>> pipe_state;
     std::vector<ShaderModuleUniqueIds> shader_unique_id_maps;  // not used, here for template function
     const VkComputePipelineCreateInfo* pCreateInfos;
 };
 
 struct CreateRayTracingPipelinesNV {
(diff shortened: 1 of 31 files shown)


layers/state_tracker/chassis_modification_state.h
```diff
@@ -72,35 +72,36 @@ struct ShaderObject {
     }
 };
 
-// One for each shader stage
-// Currently at most, we should ever have 5 stages in a pipeline/shader object
-using CreateShaderModuleStates = std::array<CreateShaderModule, 32>;
```
spencer-lunarg, line 77: so this was awful, it was only carrying around the unique shader ID 

also the `32` was suppose to be a hashmap index for `VkShaderStageFlagBits` which means clearly anything using Mesh Shading was not working as expected

layers/vulkan/generated/chassis.cpp
```diff
@@ -656,35 +656,35 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateGraphicsPipelines(VkDevice device, VkPipeli
     bool skip = false;
     ErrorObject error_obj(vvl::Func::vkCreateGraphicsPipelines, VulkanTypedHandle(device, kVulkanObjectTypeDevice));
 
-    chassis::CreateGraphicsPipelines cgpl_state[LayerObjectTypeMaxEnum]{};
```
spencer-lunarg, line 659: before we were creating everything for each layer

the goal of `chassis::CreateGraphicsPipelines` was it could be a global thing all layers used since it is at the "chassis level"

The issue is we have an different way of creating `vvl::Pipeline` state objects. These **need** to be tied to a single layer (otherwise everyone is combing them together)

jeremyg-lunarg (approved):
LGTM

scripts/generators/best_practices_generator.py
```diff
@@ -187,15 +191,20 @@ def generateSource(self):
             paramList = [param.name for param in command.params]
             paramList.append('record_obj')
             if command.name in self.extra_parameter_map:
-                paramList.append('state_data')
+                paramList.append('chassis_state')
+            if command.name in self.pipeline_parameter_map:
```
arno-lunarg, line 195:
```suggestion
            elif command.name in self.pipeline_parameter_map:
```

arno-lunarg (approved):

PR #7764, opened 2024-03-26
layers: Use vvl custom container

spencer-lunarg:
some spots I noticed we were not using `vvl::unordered_map`/`vvl::unordered_set` where we could


artem-lunarg (approved):

jeremyg-lunarg:
It would be nice if we had an automated way to catch this...

PR #7766, opened 2024-03-26
layers: Remove WSI validation dependency on GetSwapchainImages

artem-lunarg:
This is follow-up to https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7703

Found two reasons why we couldn't easily get rid of `GetSwapchainImages` from `VkRenderFramework::CreateSwapchain`:
- VUs like `VUID-VkBindImageMemorySwapchainInfoKHR-imageIndex-01644` check swapchain image count and can run before the image is acquired.
-  Swapchain image layout map initialization relied on GetSwapchainImagesKHR call.

This solution initializes swapchain image state immediately when swapchain object is created.

### DIFF
--- a/layers/state_tracker/state_tracker.cpp
+++ b/layers/state_tracker/state_tracker.cpp
@@ -3671,6 +3671,27 @@ void ValidationStateTracker::RecordCreateSwapchainState(VkResult result, const V
             std::copy(swapchain_present_modes_ci->pPresentModes, swapchain_present_modes_ci->pPresentModes + present_mode_count,
                       std::back_inserter(swapchain->present_modes));
         }
+
+        // Initialize swapchain image state
+        {
+            uint32_t swapchain_image_count = 0;
+            DispatchGetSwapchainImagesKHR(device, *pSwapchain, &swapchain_image_count, nullptr);
+            std::vector<VkImage> swapchain_images(swapchain_image_count);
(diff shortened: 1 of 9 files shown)


layers/state_tracker/state_tracker.cpp
```diff
@@ -3671,6 +3671,27 @@ void ValidationStateTracker::RecordCreateSwapchainState(VkResult result, const V
             std::copy(swapchain_present_modes_ci->pPresentModes, swapchain_present_modes_ci->pPresentModes + present_mode_count,
                       std::back_inserter(swapchain->present_modes));
         }
+
+        // Initialize swapchain image state
```
artem-lunarg, line 3675: That's a main part of this solution

tests/unit/wsi_positive.cpp
```diff
@@ -370,6 +370,8 @@ TEST_F(PositiveWsi, TransferImageToSwapchainDeviceGroup) {
     bind_info.memoryOffset = 0;
 
     vk::BindImageMemory2(device(), 1, &bind_info);
+    // Can transition layout after the memory is bound
+    peer_image.SetLayout(VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL);
```
artem-lunarg, line 374: There was missing image layout transition which was not reported previously

jeremyg-lunarg (approved):
LGTM

spencer-lunarg (approved):

PR #7769, opened 2024-03-27
gpuav: Don't create an entire spirv::Module for errors

spencer-lunarg:
When we print the error for GPU-AV and DebugPrintF we need `spirv::Instrument` to make parsing the SPIR-V easy

Before we were creating an entire `spirv::Module` in order to get `std::vector<spirv::Instruction> instructions;` while ignore all the other state tracking that `spirv::Module` does

This splits it up to make printing errors smoother

### DIFF
--- a/layers/gpu_validation/debug_printf.cpp
+++ b/layers/gpu_validation/debug_printf.cpp
@@ -78,15 +78,15 @@ void debug_printf::Validator::DestroyBuffer(BufferInfo &buffer_info) {
 }
 
 // Call the SPIR-V Optimizer to run the instrumentation pass on the shader.
-bool debug_printf::Validator::InstrumentShader(const vvl::span<const uint32_t> &input, std::vector<uint32_t> &new_pgm,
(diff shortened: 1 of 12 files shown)


layers/gpu_validation/debug_printf.cpp
```diff
@@ -78,15 +78,15 @@ void debug_printf::Validator::DestroyBuffer(BufferInfo &buffer_info) {
 }
 
 // Call the SPIR-V Optimizer to run the instrumentation pass on the shader.
-bool debug_printf::Validator::InstrumentShader(const vvl::span<const uint32_t> &input, std::vector<uint32_t> &new_pgm,
```
spencer-lunarg, line 81: took this opportunity to not use `pgm` as I still don't know what that stands for

jeremyg-lunarg: pretty sure it means 'program'

jeremyg-lunarg (approved):

PR #7774, opened 2024-03-27
layers: Semaphore cleanup

artem-lunarg:
Cleanup and refactor as follow-up to https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7736

### DIFF
--- a/layers/state_tracker/semaphore_state.cpp
+++ b/layers/state_tracker/semaphore_state.cpp
@@ -20,16 +20,6 @@
 #include "state_tracker/queue_state.h"
 #include "state_tracker/state_tracker.h"
 
-vvl::Semaphore::TimePoint::TimePoint(OpType op_type, SubmissionReference &submit_ref)
-    : signal_ref(), completed(), waiter(completed.get_future()) {
-    if (op_type == kWait) {
-        wait_refs.emplace_back(submit_ref);
-    } else {
-        assert(op_type == kSignal);
-        signal_ref.emplace(submit_ref);
-    }
-}
-
 void vvl::Semaphore::TimePoint::Notify() const {
     if (signal_ref && signal_ref->queue) {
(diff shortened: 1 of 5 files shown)


layers/state_tracker/semaphore_state.cpp
```diff
@@ -20,44 +20,74 @@
...
-    SubmissionReference signal_ref(queue, queue_seq);
-    auto result = timeline_.emplace(payload, TimePoint(kSignal, signal_ref));
-    if (!result.second) {
-        // timeline semaphore wait before signal
-        result.first->second.signal_op.emplace(signal_ref);
-    }
+    // Check there is no existing signal, validation should enforce this
+    assert(timeline_.find(payload) == timeline_.end() || !timeline_.find(payload)->second.signal_submit.has_value());
+
+    timeline_[payload].signal_submit.emplace(signal_submit);
```
artem-lunarg, line 62: Here indexing operator does what we need: if entry does not exist then create it

layers/state_tracker/semaphore_state.cpp
```diff
@@ -20,44 +20,74 @@
...
 }
 
-void vvl::Semaphore::EnqueueWait(vvl::Queue *queue, uint64_t queue_seq, uint64_t &payload) {
+void vvl::Semaphore::EnqueueWait(const SubmissionReference &wait_submit, uint64_t &payload) {
     auto guard = WriteLock();
-    SubmissionReference wait_ref(queue, queue_seq);
     if (type == VK_SEMAPHORE_TYPE_BINARY) {
+        // Timeline can be empty for the binary wait operation if the semaphore was imported.
```
artem-lunarg, line 68: Corner case that's worth to have comment about

jeremyg-lunarg (approved):
LGTM

Issue #6172, opened 2023-07-20
False negative for reusing binary semaphore in different queue after usage of vkQueueWaitIdle


sewbro01-ARM:
**Describe the Issue**
Below is scenario where issue was found, but as far as I understand it should occur on every reuse of binary semaphore in different queue:

For example we have scenario with 2 queues and 7 semaphores:
First we are using **vkQueueSubmit()** three times, twice to queue[1], with multiple **VkSubmitInfo** that are using semaphores from 0 to 4
Then we are using **vkQueueWaitIdle()**  on both queues to ensure that all submitted operations to them will finish and all used binary semaphores that are being waited for should return to unsignaled state and be good to use again.  - https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#synchronization-semaphores-waiting

The problem is when some time later we are again submiting to queues, this time once to queue[0] and once to queue[1] , and in **vkQueueSubmit()** to queue[0]  we are reusing semaphore[2]  which was previously used in queue[1] which give us VL error.
If we add new semaphore or intstead of queue[0] in second step we use queue[1], error disappears.
I also tried to use fences and  **vkWaitForFences()** but it didn't change anything.

As far as I understand, this is probably VL bug.
Thanks to usage of vkQueueWaitIdle() we are sure that semaphores that were being waited for are in unsignaled state, so basically "reset-ed" and are good for reuse, but VL seems to no respect it.

VL error and Vulkan specification clearly states:
```
VUID-vkQueueSubmit-pWaitSemaphores-00068
When a semaphore wait operation referring to a binary semaphore defined by any element of the pWaitSemaphores member of any element of pSubmits executes on queue, there must be no other queues waiting on the same semaphore
```
And the part "executes" is most important here, cause also thanks to vkQueueWaitIdle() we are sure that nothing executes after it on those queues so it should be valid to reuse semaphores in this scenario.

**Expected behavior**

VL not complaining about reuse of binary semaphore for pWaitSemaphores in different queue after usage of **vkQueueWaitIdle()** or **vkWaitForFences()** on queue that originally used it.

I believe that VL should "retire" for sure "waiting" binary semaphore and possibly also binary signaling ones with ensuring that their ratio is 1:1 as described in specification:
[...]

PR #7777, opened 2024-03-27
layers: Fix false-positive when reusing binary semaphore

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6172

The previous refactor clarified some terminology and helped with this fix. For example, `GetLastBinarySignalSubmission` was  renamed to `GetPendingBinarySignalSubmission`.  In the current code the `pending` operation usually refers to something that is still (or was recently) in the semaphore timeline (pending execution), and the `last` operation can refer either to the `pending` operation, if it exists, otherwise it's the last completed operation.

`LastOp()` is not a proper tool to implement `AnotherQueueWaits()` because we are interesting only in the pending operations. 

One direction of further code improvements is to review LastOp() usage. It's pretty complicated. Maybe it can be split into smaller scenarios similar to this fix.

### DIFF
--- a/layers/state_tracker/semaphore_state.cpp
+++ b/layers/state_tracker/semaphore_state.cpp
@@ -134,6 +134,26 @@ std::optional<vvl::SubmissionReference> vvl::Semaphore::GetPendingBinarySignalSu
     return signal_submit;
 }
 
+std::optional<vvl::SubmissionReference> vvl::Semaphore::GetPendingBinaryWaitSubmission() const {
+    assert(type == VK_SEMAPHORE_TYPE_BINARY);
+    auto guard = ReadLock();
+    if (timeline_.empty()) {
+        return {};
+    }
+    const auto &timepoint = timeline_.rbegin()->second;
+    assert(timepoint.wait_submits.empty() || timepoint.wait_submits.size() == 1);
+
+    // No waits
+    if (timepoint.wait_submits.empty()) {
+        return {};
+    }
+    // Skip waits that are not associated with a queue
+    if (timepoint.wait_submits[0].queue == nullptr) {
+        return {};
+    }
+    return timepoint.wait_submits[0];
+}
+
 vvl::Semaphore::SemOp vvl::Semaphore::Completed() const {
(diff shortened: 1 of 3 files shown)


layers/state_tracker/semaphore_state.cpp
```diff
@@ -134,6 +134,26 @@ std::optional<vvl::SubmissionReference> vvl::Semaphore::GetPendingBinarySignalSu
...
+        return {};
+    }
+    const auto &timepoint = timeline_.rbegin()->second;
+    assert(timepoint.wait_submits.empty() || timepoint.wait_submits.size() == 1);
+
+    // No waits
+    if (timepoint.wait_submits.empty()) {
+        return {};
+    }
+    // Skip waits that are not associated with a queue
```
jeremyg-lunarg, line 150: This happens from vkAcquireNextImageKHR(), right?

jeremyg-lunarg (approved):
LGTM

layers/state_tracker/semaphore_state.cpp
artem-lunarg: There are other places that do not happen on the queue (hosts command like WaitSemaphores and few special cases for signal). Image acquire does not use (null, 0) enqueueing scheme now, but a separate path, I think it should not be reported here, but did not check. That's also one direction I started to look at, to get off this special case of null queue to simplify things, image acquire is one attempt.

artem-lunarg: Hm.. vkWaitSemaphores is only for timeline semaphores. I'll check if we indeed need this.

artem-lunarg: `Semaphore::Export` is the only place that potentially can create wait with empty submit. If it becomes clear it's not possible then we can remove that check, for now will leave it here.

I also realized that I could probably use LastOp's pending parameter and filter without new helper function. I'd still like to try to decompose LastOp into simpler helpers. So will leave this for now and will make an attempt to categorize LastOp usages.

Issue #7702, opened 2024-03-17
GPU-AV vkCreateShaderModule doesn't support mixed stage shader modules


Nielsbishere:
**Environment:**
 - OS: Win11 and Win10
 - GPU and driver version: NV 3070, 3080, 4090: 551.76, AMD 6800xt and W7800: 24.2.1, Intel A770: 101.5333
 - SDK or header version if building from repo: 1.3.275
 - Options enabled (synchronization, best practices, etc.): GPU assisted layer + reserved slot, best practices, synchronization

**Describe the Issue**
A raytracing shader with multiple stages can't be a single binary, even though this is way more efficient in production environments as shaders can be large so saving compilation by reducing compiles is good practice. And it's also accepted by all drivers I've tested (Intel, AMD, NV).
```
Error: Validation Error: [ UNASSIGNED-GPU-Assisted ] | MessageID = 0xfd41b5b6 | vkCreateShaderModule():  Error during shader instrumentation: line 0: Mixed stage shader module not supported
  OpEntryPoint ClosestHitNV %mainClosestHit "mainClosestHit" %3 %4 %5 %6 %7 %F32_pi %_rwTextures2D %_tlasExt %globals %payload %payload_0 %attr %payload_1
```

Repro can be found on https://github.com/Oxsomi/rt_core, it can easily be tested by compiling this with DXC and then calling vkCreateShaderModule with it:
```hlsl
typedef float2 F32x2;
typedef uint2 U32x2;
typedef float3 F32x3;
typedef float4 F32x4;
typedef float F32;
typedef uint U32;

struct ColorPayload {
	F32x3 color;
	F32 hitT;
};

[shader("miss")] 
void mainMiss(inout ColorPayload payload) {
	payload.color = F32x3(0.25, 0.5, 1);
	payload.hitT = -1;
}

F32x3 expandBary(F32x2 bary) {
	return F32x3(1 - bary.x - bary.y, bary);
}

[shader("closesthit")]
void mainClosestHit(inout ColorPayload payload, BuiltInTriangleIntersectionAttributes attr) {
	payload.color = expandBary(attr.barycentrics);
	payload.hitT = RayTCurrent();
}

RWTexture2D<F32x4> _myTexture;
RaytracingAccelerationStructure _tlas;

[shader("raygeneration")]
void mainRaygen() {

	U32x2 id = DispatchRaysIndex().xy;
	U32x2 dims = DispatchRaysDimensions().xy;

[...]

PR #7778, opened 2024-03-28
gpuav: Add Multiple EntryPoint support

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7702

(going to add Debug PrintF support for this as well, but will be a separate PR since it requires a PR into SPIR-V Tools)


jeremyg-lunarg (approved):
LGTM

PR #7780, opened 2024-03-28
scripts: Use promoted SPIR-V names

spencer-lunarg:
Went through and made sure the promoted alias name for SPIR-V is used throughout the code

### DIFF
--- a/layers/gpu_validation/gpu_error_message.cpp
+++ b/layers/gpu_validation/gpu_error_message.cpp
@@ -57,45 +57,55 @@ static void GenerateStageMessage(const uint32_t *error_record, std::string &msg)
             strm << "Stage = Compute.  Global invocation ID (x, y, z) = (" << error_record[kHeaderInvocationIdXOffset] << ", "
                  << error_record[kHeaderInvocationIdYOffset] << ", " << error_record[kHeaderInvocationIdZOffset] << " )";
         } break;
-        case spv::ExecutionModelRayGenerationNV: {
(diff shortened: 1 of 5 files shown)


layers/gpu_validation/gpu_error_message.cpp
```diff
@@ -57,45 +57,55 @@ static void GenerateStageMessage(const uint32_t *error_record, std::string &msg)
...
                  << error_record[kHeaderRayTracingLaunchIdYOffset] << ", " << error_record[kHeaderRayTracingLaunchIdZOffset]
                  << "). ";
         } break;
-        case spv::ExecutionModelCallableNV: {
+        case spv::ExecutionModelCallableKHR: {
             strm << "Stage = Callable.  Global Launch ID (x,y,z) = (" << error_record[kHeaderRayTracingLaunchIdXOffset] << ", "
                  << error_record[kHeaderRayTracingLaunchIdYOffset] << ", " << error_record[kHeaderRayTracingLaunchIdZOffset]
                  << "). ";
         } break;
+        case spv::ExecutionModelTaskEXT: {
```
spencer-lunarg, line 90: note: common mistake, in Vulkan the `VK_SHADER_STAGE_MESH_BIT_NV`/`VK_SHADER_STAGE_MESH_BIT_EXT` are the same, but in SPIR-V the `ExecutionModel` is actually a different value

artem-lunarg (approved):

PR #7781, opened 2024-03-28
printf: Add proper vk_layer_settings support

spencer-lunarg:
Debug PrintF was using the old way to set its settings

Now it is hooked up with vk_layer_settings like everything else


jeremyg-lunarg (approved):
LGTM

PR #7787, opened 2024-03-28
scripts: Fix PlatformGuardHelper.add_guard()

jeremyg-lunarg:
Comparison to the existing guard should be `!=` not `is not`

Found during code review of https://github.com/KhronosGroup/Vulkan-Utility-Libraries/pull/187

### DIFF
--- a/scripts/generators/generator_utils.py
+++ b/scripts/generators/generator_utils.py
@@ -77,11 +77,11 @@ class PlatformGuardHelper():
 
     def add_guard(self, guard, extra_newline = False):
         out = []
-        if self.current_guard is not guard and self.current_guard is not None:
+        if self.current_guard != guard and self.current_guard is not None:
             out.append(f'#endif  // {self.current_guard}\n')
         if extra_newline:
             out.append('\n')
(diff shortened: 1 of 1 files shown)


charles-lunarg (commented):
Shouldn't this also run the code-gen? Or did the previous code not check in codegen either?

jeremyg-lunarg:
> Shouldn't this also run the code-gen? Or did the previous code not check in codegen either?

I reran the codegen. Apparently this method is no longer used in VVL code gen. There's also a github CI job that will check that codegen is up to date.

charles-lunarg (approved):
Well if CI passes, who am I to complain. I think I added it here to keep generator_utils.py the same across the various repo's.

jeremyg-lunarg:
> Well if CI passes, who am I to complain. I think I added it here to keep generator_utils.py the same across the various repo's.

I think @spencer-lunarg has a plan to move all of these scripts to 1 repo but hasn't gotten to it yet.

artem-lunarg (approved):

spencer-lunarg (approved):

PR #7788, opened 2024-03-28
layers: Expose semaphore payload instead of operation object

artem-lunarg:
Trying to remove SemOp usage in the Semaphore public interface.

In some place places where Completed() was called, the code uses "current" payload terminology, which seems logical. Hense the new name of the method.

### DIFF
--- a/layers/state_tracker/semaphore_state.h
+++ b/layers/state_tracker/semaphore_state.h
@@ -100,9 +100,9 @@ class Semaphore : public RefcountedStateObject {
     // Returns pending queue submission that waits on this binary semaphore.
     std::optional<SubmissionReference> GetPendingBinaryWaitSubmission() const;
 
-    // This is the most recently completed operation. It is returned by value so that the caller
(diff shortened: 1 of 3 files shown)


layers/state_tracker/semaphore_state.h
```diff
@@ -100,9 +100,9 @@ class Semaphore : public RefcountedStateObject {
     // Returns pending queue submission that waits on this binary semaphore.
     std::optional<SubmissionReference> GetPendingBinaryWaitSubmission() const;
 
-    // This is the most recently completed operation. It is returned by value so that the caller
-    // has a correct copy even if something else is completing on this queue in a different thread.
-    SemOp Completed() const;
+    // Current payload value.
+    // If a queue submission command is pending execution, then the returned value may immediately be out of date.
```
artem-lunarg, line 104: Comment in the spec style...

jeremyg-lunarg (approved):

PR #7791, opened 2024-03-29
layers: Rename layer_chassis to chassis

spencer-lunarg:
there is a `layer_dispatch` because it handles the dispatching from a layer scope

the chassis is an internal design we have to offload a call into a certain scope of work, it is not tied to being a Vulkan Layer directly

having half things be `chassis` and half be `layer_chassis` is very confusing and I believe just stems from the days where everything was one repo

this renames `layer_chassis` -> `chassis`


spencer-lunarg:
cc @mark-lunarg let me know if I am completely wrong here

PR #7793, opened 2024-03-29
layers: Add robustness check to drawn time vertex buffer

spencer-lunarg:
from https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6570

- Move logic to a dedicated `ValidateVertexAttribute` function
- wrap logic with a check for `robustBufferAccess` support

cc @ziga-lunarg

### DIFF
--- a/layers/core_checks/cc_pipeline_graphics.cpp
+++ b/layers/core_checks/cc_pipeline_graphics.cpp
@@ -3291,74 +3291,7 @@ bool CoreChecks::ValidatePipelineDrawtimeState(const LastBound &last_bound_state
             }
         }
 
-        const auto &vertex_attribute_descriptions = pipeline->IsDynamic(VK_DYNAMIC_STATE_VERTEX_INPUT_EXT)
-                                                        ? cb_state.dynamic_state_value.vertex_attribute_descriptions
-                                                        : pipeline->vertex_input_state->vertex_attribute_descriptions;
-        // Verify vertex attribute address alignment
-        for (uint32_t i = 0; i < vertex_attribute_descriptions.size(); i++) {
-            const auto &attribute_description = vertex_attribute_descriptions[i];
-            const uint32_t vertex_binding = attribute_description.binding;
-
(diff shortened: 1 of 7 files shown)


ziga-lunarg (approved):

artem-lunarg (approved):
LGTM. Interesting change. It is according to spec update but I find this a bit unintuitive, if that's what happens here. I look at robust buffer access as a debug feature to help application debugging (can also affect performance). If there is a VU that can detect issue earlier (and more reliably, robustness does not report back) why doesn't VU have higher priority?

spencer-lunarg:
>  look at robust buffer access as a debug feature to help application debugging (can also affect performance).

The main use case for robust buffer access to my knowledge is actually WebGPU which has to sandbox everything

The whole point of robustness is you get back a "known value" so if you try to access vertex buffer memory not there, it is required to return `0` and nothing is invalid

(not vertex buffer) but some people rely on Robustness so they can go accessing what they want and know there is a "default" when they are out of range

>  why doesn't VU have higher priority?

I guess I am not seeing where your concerns here are

artem-lunarg:
> I guess I am not seeing where your concerns here are

Hm.. still it's not that obvious. I used robustness feature to debug crashes, so application can progress further where I inspect values in the debugger instead of crashing. I think my concern is why not to allow VU validation here, so the issue will be reported immediately instead of just making it not crash and then you need to do less trivial gpu side debugging.

spencer-lunarg:
I think what this might need is a

```
if (robustness_enabled) {
    LogWarning()
} else {
    LogError()
}
```

type logic

it clearly is not an error as it is very valid and defined behavior what will happen if you have robustness on (and someone might argue it would be spam to report this if they are aware what they are doing)

but at the same time, I agree for a developer, it is useful to at least "know" which is why I think a "warning" is maybe more appropriate... the issue is we have tried to move away from warning in CoreChecks, so maybe robustness is where we draw the line for an exception

artem-lunarg:
Yeah, one scenario if for some reason application relies on robustness feature for correctness. In my comment I assumed it's not the case, but technically it's a valid behavior, even it looks like a workaround and can be not a good practice.

PR #7794, opened 2024-03-29
chassis: Create manual chassis file and folder

spencer-lunarg:
Create a `chassis` directory to group all "chassis level" logic together

Will be adding more here soon, so want to group it all in a more logical place


layers/chassis/layer_chassis_dispatch_manual.cpp
```diff
@@ -430,7 +425,7 @@ void DispatchDestroyDescriptorPool(VkDevice device, VkDescriptorPool descriptorP
     WriteLockGuard lock(dispatch_lock);
 
     // remove references to implicitly freed descriptor sets
-    for(auto descriptor_set : layer_data->pool_descriptor_sets_map[descriptorPool]) {
+    for (auto descriptor_set : layer_data->pool_descriptor_sets_map[descriptorPool]) {
```
spencer-lunarg, line 428: I was concern clang-format was acting strange, but from this line, seems clang-format was just never ran on this file before :shrug:

jeremyg-lunarg (approved):

Issue #7795, opened 2024-03-29
`vkCmdClearAttachments` crashes when using an imageless framebuffer in `VkCommandBufferInheritanceInfo`


glebov-andrey:
**Environment:**
 - OS: Windows
 - SDK or header version if building from repo: current `main`
 - Options enabled (synchronization, best practices, etc.): Core

**Describe the Issue**

When `vkCmdClearAttachments` is recorded into a secondary command buffer whose `VkCommandBufferInheritanceInfo` contains an imageless framebuffer, VVL crashes while trying to access an image view (`depth_view_state` is `nullptr`): https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/242b6f33a6a8fbb8a043391baf68ee7f6711ac4b/layers/core_checks/cc_image.cpp#L1150

Ideally, in this case, validation would use the information provided in `VkFramebufferAttachmentImageInfo` or delayed until `vkCmdExecuteCommands` - when the image views are known.

**Additional context**

<details>
<summary>The callstack</summary>

```
[VkLayer_khronos_validation.dll] CoreChecks::PreCallValidateCmdClearAttachments(VkCommandBuffer_T *, unsigned int, const VkClearAttachment *, unsigned int, const VkClearRect *, const ErrorObject &) cc_image.cpp:1150
[VkLayer_khronos_validation.dll] vulkan_layer_chassis::CmdClearAttachments(VkCommandBuffer_T *, unsigned int, const VkClearAttachment *, unsigned int, const VkClearRect *) chassis.cpp:3596
```

</details>

PR #7796, opened 2024-03-30
layers: Fix imageless framebuffer attachments clear crash

ziga-lunarg:
Part of #7795 

This PR fixes the crash, however the validation here should be moved to vkCmdExecuteCommands for imageless framebuffers


ziga-lunarg:
Seems like this also crashes in the driver

spencer-lunarg (approved):
LGTM - Test is good

we have a lot of clean up to do for secondary command buffer logic

spencer-lunarg:
Seems like the test is crashing on some devices :cry:

ziga-lunarg:
> Seems like the test is crashing on some devices 😢

Looks like a driver bug to me

PR #7802, opened 2024-04-01
wip: Fix issue 7790

artem-lunarg:


artem-lunarg:
Corresponding mock icd PR: https://github.com/KhronosGroup/Vulkan-Tools/pull/979

layers/core_checks/cc_device_memory.cpp
```diff
@@ -330,25 +330,16 @@ bool CoreChecks::ValidateSetMemBinding(VkDeviceMemory memory, const vvl::Bindabl
...
     const VkExternalMemoryHandleTypeFlags ignored_allocation = VK_EXTERNAL_MEMORY_HANDLE_TYPE_D3D11_TEXTURE_BIT |
                                                                VK_EXTERNAL_MEMORY_HANDLE_TYPE_D3D11_TEXTURE_KMT_BIT |
                                                                VK_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_RESOURCE_BIT;
-#ifdef VK_USE_PLATFORM_WIN32_KHR
     const auto import_memory_win32 = vku::FindStructInPNextChain<VkImportMemoryWin32HandleInfoKHR>(allocate_info.pNext);
     if (import_memory_win32 && (import_memory_win32->handleType & ignored_allocation) != 0) {
         return true;
     }
 #endif
-    const auto import_memory_fd = vku::FindStructInPNextChain<VkImportMemoryFdInfoKHR>(allocate_info.pNext);
```
artem-lunarg, line 343: Remove scenarios that can't use D3D handles.

layers/core_checks/cc_device_memory.cpp
```diff
@@ -330,25 +330,16 @@ bool CoreChecks::ValidateSetMemBinding(VkDeviceMemory memory, const vvl::Bindabl
     return skip;
 }
 
-bool CoreChecks::IsZeroAllocationSizeAllowed(const VkMemoryAllocateInfo &allocate_info) const {
+bool CoreChecks::IgnoreAllocationSize(const VkMemoryAllocateInfo &allocate_info) const {
```
artem-lunarg, line 333: Detects when allocation size is ignored, not only adds zero as possible value.

layers/core_checks/cc_device_memory.cpp
```diff
@@ -357,13 +348,6 @@ bool CoreChecks::IsZeroAllocationSizeAllowed(const VkMemoryAllocateInfo &allocat
             return true;
         }
     }
-
-#ifdef VK_USE_PLATFORM_FUCHSIA
```
arno-lunarg, line 361: This Fuchsia part is no longer needed??

tests/unit/memory_positive.cpp
```diff
@@ -506,3 +506,63 @@ TEST_F(PositiveMemory, DeviceImageMemoryRequirements) {
...
+    AddRequiredExtensions(VK_KHR_EXTERNAL_MEMORY_WIN32_EXTENSION_NAME);
+    RETURN_IF_SKIP(Init());
+    // Mock ICD allows to use fake DX11 handles instead of using DX11 API directly.
+    if (!IsPlatformMockICD()) {
+        GTEST_SKIP() << "This test only runs on the mock ICD";
+    }
+
+    VkExternalMemoryImageCreateInfo external_info = vku::InitStructHelper();
+    external_info.handleTypes = VK_EXTERNAL_MEMORY_HANDLE_TYPE_D3D11_TEXTURE_BIT;
+    auto image_ci = vkt::Image::ImageCreateInfo2D(32, 32, 1, 1, VK_FORMAT_B8G8R8A8_UNORM, VK_IMAGE_USAGE_SAMPLED_BIT);
```
arno-lunarg, line 522: Nit: use `DefaultImageInfo()`?

arno-lunarg (approved):
LGTM! Just a question about fuchsia

layers/core_checks/cc_device_memory.cpp
artem-lunarg: I think only `VkImportMemoryWin32HandleInfoKHR` can import D3D handles. Also based on the type system, it's the only import structure that can do this (uses HANDLE type needed to store win32 handle).

tests/unit/memory_positive.cpp
artem-lunarg: It looks like it's only for the image tests. Idk, maybe we need a global helper too.

PR #7811, opened 2024-04-08
layers: Add util for stage flags

spencer-lunarg:
moves `kShaderStageAllRayTracing` from a GPU-AV const to a layer util constant


arno-lunarg (approved):
https://www.youtube.com/watch?v=3ATBJGkdYgI

PR #7815, opened 2024-04-09
layers: Add 03704 for host accel structs builds

arno-lunarg:
VUID-vkBuildAccelerationStructuresKHR-scratchData-03704
The range of memory backing the scratchData member of any element of pInfos that is accessed by this command must not overlap the memory backing the scratchData member of any other element of pInfos, which is accessed by this command

### DIFF
--- a/layers/core_checks/cc_ray_tracing.cpp
+++ b/layers/core_checks/cc_ray_tracing.cpp
@@ -206,18 +206,18 @@ bool CoreChecks::ValidateAccelerationStructuresMemoryAlisasing(const LogObjectLi
             }
         }
 
-            // Validate that there is no destination acceleration structures' memory overlaps
-            if (dst_as_state && other_dst_as_state) {
-                const char *vuid = error_obj.location.function == Func::vkCmdBuildAccelerationStructuresKHR
-                                       ? "VUID-vkCmdBuildAccelerationStructuresKHR-dstAccelerationStructure-03702"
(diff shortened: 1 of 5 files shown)


layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -990,6 +990,33 @@ bool CoreChecks::PreCallValidateBuildAccelerationStructuresKHR(
         }
 
         skip |= ValidateAccelerationStructuresMemoryAlisasing(LogObjectList(), infoCount, pInfos, info_i, error_obj);
+        const VkDeviceSize scratch_i_size = rt::ComputeScratchSize(rt::BuildType::Host, device, *info, ppBuildRangeInfos[info_i]);
+        auto scratch_i_host_addr = reinterpret_cast<uint64_t>(info->scratchData.hostAddress);
+        const sparse_container::range<uint64_t> scratch_addr_range(scratch_i_host_addr, scratch_i_host_addr + scratch_i_size);
+
+        assert(infoCount > info_i);
+        for (auto [other_info_j, other_info] : vvl::enumerate(pInfos + info_i + 1, infoCount - (info_i + 1))) {
```
spencer-lunarg, line 998: we have `info_i + 1` 3 times here, where is this coming from? is it worth just making a variable for it as it is not obvious right away why we need this `+ 1`

spencer-lunarg (approved):

PR #7817, opened 2024-04-09
layers: Move Secondary CommandBuffer tracking to chassis

spencer-lunarg:
Currently we were tracking the command buffers for being secondary because in `VkCommandBufferBeginInfo` it says

image

This is the first a few spots like this, so I created a `chassis::HandleData` object we can put into the `RecordObject`/`ErrorObject` to pass down this information

This basically provides a chassis-level way of doing state tracking, but only passing down the information we care about

### DIFF
--- a/layers/chassis/chassis_handle_data.h
+++ b/layers/chassis/chassis_handle_data.h
@@ -1,0 +1,33 @@
+/* Copyright (c) 2024 The Khronos Group Inc.
+ * Copyright (c) 2024 Valve Corporation
+ * Copyright (c) 2024 LunarG, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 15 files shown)


layers/chassis/chassis_handle_data.h
```diff
@@ -0,0 +1,33 @@
...
+namespace chassis {
+
+struct CommandBufferData {
+    bool is_secondary;
+};
+
+// Some handles need their own "private data" that everyone needs (even statless chassis layers)
+// The goal is to have the chassis do the state tracking once, and pass down this selective information to everyone.
+// Using an union, we can leverage the ErrorObject/RecordObject to get the data everywhere
+union HandleData {
```
jeremyg-lunarg, line 29: not that it matters at this stage, but you may want to consider using std::variant<>.  The biggest advantages come if you need destructors to run on the union contents.

spencer-lunarg: for now I only have `bool is_ignored` in mind to add, but let me also look into `std::variant` as well

Issue #7786, opened 2024-03-28
False positive in minImageCount validation in EXT_swapchain_maintenance1


HansKristian-Work:
**Environment:**
 - OS: Arch Linux
 - GPU and driver version: RX 6800 Mesa main
 - SDK or header version if building from repo: VVL main
 - Options enabled (synchronization, best practices, etc.): synchronization

**Describe the Issue**

When using EXT_swapchain_maintenance1, it's possible to use a minImageCount that is per-present mode, using
VkSwapchainPresentModesCreateInfoEXT. To validate correctly, GetSurfaceCaps2 with explicit present mode must be used to determine validity.

To reproduce, https://github.com/Themaister/Granite with latency-test on Wayland with Mesa main.

**Expected behavior**

No false positive.

**Valid Usage ID**

```
"Validation Error: [ VUID-VkSwapchainCreateInfoKHR-presentMode-02839 ] | MessageID = 0xfb70ccdb | vkCreateSwapchainKHR(): pCreateInfo->minImageCount 2, which is outside the bounds returned by vkGetPhysicalDeviceSurfaceCapabilitiesKHR() (i.e. minImageCount = 4, maxImageCount = 0). The Vulkan spec states: If presentMode is not VK_PRESENT_MODE_SHARED_DEMAND_REFRESH_KHR nor VK_PRESENT_MODE_SHARED_CONTINUOUS_REFRESH_KHR, then minImageCount must be greater than or equal to the value returned in the minImageCount member of the VkSurfaceCapabilitiesKHR structure returned by vkGetPhysicalDeviceSurfaceCapabilitiesKHR for the surface (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkSwapchainCreateInfoKHR-presentMode-02839)"
```

artem-lunarg:
> To validate correctly, GetSurfaceCaps2 with explicit present mode must be used to determine validity.

@HansKristian-Work looking into the code it looks like it does exactly this:

```
const auto surface_caps2 =
    surface_state->GetCapabilities(IsExtEnabled(instance_extensions.vk_khr_get_surface_capabilities2),
                                   physical_device_state->VkHandle(), surface_caps_query_pnext, create_info_loc, this);
```
It calls GetSurfaceCaps2 if the extension is enabled. And it makes query for specific present mode:
```
VkSurfacePresentModeEXT present_mode_info = vku::InitStructHelper();
[...]

PR #7820, opened 2024-04-09
layers: Refactor surface caps caching

artem-lunarg:
Fixes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7786 (tested on Granite renderer).

This reworks surface capabilities caching. The result of a surface caps query depends on the surface info's pNext chain, each element can affect results. This solution caches data for two types of queries: surface info's pNext is a) null or b) a single `VkSurfacePresentModeEXT` structure. For other types of pNext chains caching is not performed and corresponding queries will skip the cache and run regular Vulkan API calls.

### DIFF
--- a/layers/core_checks/cc_wsi.cpp
+++ b/layers/core_checks/cc_wsi.cpp
@@ -338,15 +338,15 @@ bool CoreChecks::ValidateCreateSwapchain(const VkSwapchainCreateInfoKHR &create_
         }
     }
 
-    void *surface_caps_query_pnext = nullptr;
+    void *surface_info_pnext = nullptr;
 #if defined(VK_USE_PLATFORM_WIN32_KHR)
     VkSurfaceFullScreenExclusiveInfoEXT full_screen_info_copy = vku::InitStructHelper();
     VkSurfaceFullScreenExclusiveWin32InfoEXT win32_full_screen_info_copy = vku::InitStructHelper();
     const auto *full_screen_info = vku::FindStructInPNextChain<VkSurfaceFullScreenExclusiveInfoEXT>(create_info.pNext);
     if (full_screen_info && full_screen_info->fullScreenExclusive == VK_FULL_SCREEN_EXCLUSIVE_APPLICATION_CONTROLLED_EXT) {
         full_screen_info_copy = *full_screen_info;
-        full_screen_info_copy.pNext = surface_caps_query_pnext;
-        surface_caps_query_pnext = &full_screen_info_copy;
+        full_screen_info_copy.pNext = surface_info_pnext;
+        surface_info_pnext = &full_screen_info_copy;
 
         if (IsExtEnabled(device_extensions.vk_khr_win32_surface)) {
             const auto *win32_full_screen_info =
@@ -361,8 +361,8 @@ bool CoreChecks::ValidateCreateSwapchain(const VkSwapchainCreateInfoKHR &create_
                 }
             } else {
                 win32_full_screen_info_copy = *win32_full_screen_info;
-                win32_full_screen_info_copy.pNext = surface_caps_query_pnext;
-                surface_caps_query_pnext = &win32_full_screen_info_copy;
+                win32_full_screen_info_copy.pNext = surface_info_pnext;
+                surface_info_pnext = &win32_full_screen_info_copy;
             }
         }
     }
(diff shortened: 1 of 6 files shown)


artem-lunarg:
Need to add regression test here based on the issue.

layers/core_checks/cc_wsi.cpp
```diff
@@ -1125,14 +1114,13 @@ bool CoreChecks::ValidateAcquireNextImage(VkDevice device, VkSwapchainKHR swapch
         const uint32_t acquired_images = swapchain_data->acquired_images;
         const uint32_t swapchain_image_count = static_cast<uint32_t>(swapchain_data->images.size());
 
-        vku::safe_VkSurfaceCapabilities2KHR surface_caps2{};
```
spencer-lunarg, line 1128: during my safe struct reworking, I saw this and was going to look more why this needed to be a safe struct... glad to see you got to it first

layers/state_tracker/image_state.h
```diff
@@ -426,35 +411,49 @@ class Surface : public StateObject {
...
+  private:
+    // Contains per present mode capabilities
+    struct PresentModeInfo {
+        VkPresentModeKHR present_mode;
+        VkSurfaceCapabilitiesKHR surface_capabilities;
+        std::optional<VkSurfacePresentScalingCapabilitiesEXT> scaling_capabilities;
+        std::optional<std::vector<VkPresentModeKHR>> compatible_present_modes;
+    };
+    // Cached information per physical device. Optional indicates if element is in the cache.
+    struct PhysDevCache {
```
spencer-lunarg, line 440: I like the idea of having a physical device level cache like this

spencer-lunarg (approved):
Having a test would be great, but overall this is a nice cleanup to managing all those surface capabilities values

layers/core_checks/cc_wsi.cpp
artem-lunarg: Yeah, one reason that in the existing cases we did not use `caps pNext` info from 2KHR version, this part is just refactor that does not change behavior. Probably more important reason if some code tried to use that chained data it would be unsafe since in general we don't know the structure of `info pNext` used to get this caps, so can't reason about the meaning of caps data, it gets possible to get non-deterministic behavior when the same query gets different result depending on which code initiated caching. Looks like the better approach is to preprocess pNext info early and provide specific APIs for scenarios we are interested in.

layers/state_tracker/state_tracker.cpp
```diff
@@ -4052,24 +4050,20 @@ void ValidationStateTracker::PostCallRecordGetPhysicalDeviceSurfaceCapabilities2
...
-                                                          pSurfaceCapabilities->surfaceCapabilities, *present_scaling_caps);
+        if (!pSurfaceInfo->pNext) {
+            surface_state->UpdateCapabilitiesCache(physicalDevice, pSurfaceCapabilities->surfaceCapabilities);
+        } else if (IsExtEnabled(device_extensions.vk_ext_surface_maintenance1)) {
+            const auto *surface_present_mode = vku::FindStructInPNextChain<VkSurfacePresentModeEXT>(pSurfaceInfo->pNext);
+            if (surface_present_mode) {
+                // The surface caps caching should take into account pSurfaceInfo->pNext chain structure,
+                // because each pNext element can affect query result. Here we support caching for a common
+                // case when pNext chain is a single VkSurfacePresentModeEXT structure. Surface queries are
+                // not high frequency operations, disabling caching alltogether might be an option too.
```
Eearslya, line 4061: Just wanted to comment that, as far as I know, the purpose of caching surface capabilities isn't because they're high-frequency and caching saves API calls; the reason was an old validation false-positive. There was basically a race condition where the surface capabilities can change between the user getting the capabilities and creating the swapchain, which resulted in validation complaining about min/max extents being one or two pixels off, despite the fact that the user couldn't really do anything about that.

artem-lunarg: Doesn't this sounds like a driver (or vvl/app) bug? Until the spec allows this I thought the queries should be deterministic and disabling caching should result in the most robust behavior (no potential bugs from caching).

artem-lunarg:
Regression test is added, reproduceable on wayland.

layers/state_tracker/state_tracker.cpp
artem-lunarg: But to give a specific answer here. The caching should guarantee the same behavior as using direct API calls. That's basically definition of caching, provide the same results but in faster or more convenient way. Other behavior should be considered a bug and should be fixed.

Eearslya: I don't know if I'd call it a bug, per se? The whole issue was that the user followed the exact procedure they're supposed to. Call `vkGetPhysicalDeviceSurfaceCapabilitiesKHR`, clamp their swapchain size according to the min/max extents, and then call `vkCreateSwapchainKHR`.

The problem was that during the validation of `vkCreateSwapchainKHR`, the validation layers would call `vkGetPhysicalDeviceSurfaceCapabilitiesKHR` *again*, to know what the correct values should be. And unfortunately, in those microseconds between the user getting the caps and validation getting the caps, the surface may have actually changed.

artem-lunarg: > The problem was that during the validation of vkCreateSwapchainKHR, the validation layers would call vkGetPhysicalDeviceSurfaceCapabilitiesKHR again, to know what the correct values should be. And unfortunately, in those microseconds between the user getting the caps and validation getting the caps, the surface may have actually changed.

@Eearslya  Thanks for the explanation, it makes sense.

artem-lunarg: @spencer-lunarg if the app gets surface caps and then surface changes and then the app tries to create swapchain using queried caps. It does not sound right that you can use old surface caps after surface was change to create swapchain. But need to check the spec.

artem-lunarg: > Some people are on the fence of whether it's the user's fault for allowing the window to resize

That’s a standard thing on windows, you should be able to do resize if you want. There are events to detect resize. Of course it’s possible to get resize immediately after resize but you expect this in the code and retry if swapchain creation fails but querying surface caps again. I agree, swapchain apis and parts of OS support is not perfect.

artem-lunarg: Checking the spec I found the following part, but it is about the presentation, when swapchain is already created:
> It is the application’s responsibility to detect surface size changes and react appropriately. If presentation fails because of a mismatch in the surface and presented image sizes, a VK_ERROR_OUT_OF_DATE_KHR error will be returned.

I can't find something like this related to `vkCreateSwapchainKHR`. I think it might be harder to report an error here, similar to VK_ERROR_OUT_OF_DATE_KHR, because we did not have previous association with a surface. It's hard to distinuish between  a) surface is unchanged but the app just provided wrong arguments (that's one type of error) or b) the app computed parameters correctly but surface changed in the last moment (in this case would could reporting something similar to VK_ERROR_OUT_OF_DATE_KHR ). 

One version of the API that allows to report something similar VK_ERROR_OUT_OF_DATE_KHR for vkCreateSwapchain is to have an ability to query a token from a surface which is incremented each time when the surface changes. Then we provide that token to vkCreateSwapchain (using pNext for example), and then the system can compare that token to the current surface's token during vkCreateSwapchain and this might give the system the ability to report an error similar to VK_ERROR_OUT_OF_DATE_KHR.

PR #7821, opened 2024-04-10
scripts: Add InterceptId where possible

spencer-lunarg:
I noticed in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7817 that we have a single "manual function" for the chassis logic but we only need a subset for the `chassis_dispatch_helper.h` file because it has some functions that can still make use of the `intercept_vectors`


artem-lunarg (approved):
LGTM. It's interesting what's a ratio of the functions that have intercept vector to those that do not. Would it make sense at some point to put everything in intercept_vectors?

spencer-lunarg:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/layers/vulkan/generated/chassis.h#L4507-L4586

@artem-lunarg these are the ones that don't have an `intercept_vectors` because we have an overload

This whole `interceptId` system is before my time, assuming it still gives us a 5% boost for multi-thread apps, but anything else touching this would definitely need some profiling

PR #7825, opened 2024-04-10
layer: Cleanup of manual dispatch variables

spencer-lunarg:
For things like

```c++
vku::safe_VkDescriptorUpdateTemplateCreateInfo var_local_pCreateInfo;
vku::safe_VkDescriptorUpdateTemplateCreateInfo *local_pCreateInfo = nullptr;
```

we need to have both variables, so we generate it for everyone, but these are manual and don't need it

### DIFF
--- a/layers/chassis/layer_chassis_dispatch_manual.cpp
+++ b/layers/chassis/layer_chassis_dispatch_manual.cpp
@@ -87,18 +87,16 @@ void CopyExportMetalObjects(const void *src_chain, const void *dst_chain) {
 void DispatchExportMetalObjectsEXT(VkDevice device, VkExportMetalObjectsInfoEXT *pMetalObjectsInfo) {
     auto layer_data = GetLayerDataPtr(GetDispatchKey(device), layer_data_map);
     if (!wrap_handles) return layer_data->device_dispatch_table.ExportMetalObjectsEXT(device, pMetalObjectsInfo);
-    vku::safe_VkExportMetalObjectsInfoEXT var_local_pMetalObjectsInfo;
-    vku::safe_VkExportMetalObjectsInfoEXT *local_pMetalObjectsInfo = nullptr;
+    vku::safe_VkExportMetalObjectsInfoEXT local_pMetalObjectsInfo;
     {
         if (pMetalObjectsInfo) {
-            local_pMetalObjectsInfo = &var_local_pMetalObjectsInfo;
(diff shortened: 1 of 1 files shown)


layers/chassis/layer_chassis_dispatch_manual.cpp
```diff
@@ -1316,96 +1310,92 @@ void DispatchGetAccelerationStructureBuildSizesKHR(VkDevice device, VkAccelerati
...
-        device, buildType, (const VkAccelerationStructureBuildGeometryInfoKHR *)local_pBuildInfo, pMaxPrimitiveCounts, pSizeInfo);
+        device, buildType, (const VkAccelerationStructureBuildGeometryInfoKHR *)&local_pBuildInfo, pMaxPrimitiveCounts, pSizeInfo);
 }
 
 void DispatchGetDescriptorEXT(VkDevice device, const VkDescriptorGetInfoEXT *pDescriptorInfo, size_t dataSize, void *pDescriptor) {
     auto layer_data = GetLayerDataPtr(GetDispatchKey(device), layer_data_map);
     if (!wrap_handles) return layer_data->device_dispatch_table.GetDescriptorEXT(device, pDescriptorInfo, dataSize, pDescriptor);
-    vku::safe_VkDescriptorGetInfoEXT var_local_pDescriptorInfo;
-    vku::safe_VkDescriptorGetInfoEXT *local_pDescriptorInfo = NULL;
+    vku::safe_VkDescriptorGetInfoEXT local_pDescriptorInfo;
```
spencer-lunarg, line 1340: related to https://github.com/KhronosGroup/Vulkan-Utility-Libraries/pull/196 

This is the only place that "needed" `safe_VkDescriptorDataEXT` and the only reason we are using `safe_` here is because it handles copying `pDescriptorInfo` so we can unwrap, but the pointer to these handles don't need to be safe struct (well I guess safe union in this case)

spencer-lunarg: cc @jeremyg-lunarg

spencer-lunarg:
> Do we need both variables in the general case to be able to specify null?

I guess not sure why it would mater here, I couldn't see an edge case, maybe I am missing something, but I think these were "left in" by accident and had no purpose... which again is confusing, that is why I want to remove them if possible

artem-lunarg (approved):

PR #7827, opened 2024-04-10
Add vuid 09382, validate render pass contents

ziga-lunarg:
This PR adds validation for `VUID-VkSubpassBeginInfo-contents-09382`, which is for `vkCmdBeginRenderPass2`, but a similar VUID for `vkCmdBeginRenderPass` seems to be missing from the spec

### DIFF
--- a/tests/unit/render_pass.cpp
+++ b/tests/unit/render_pass.cpp
@@ -4312,3 +4312,21 @@ TEST_F(NegativeRenderPass, RenderPassWithRenderPassStripedQueueSubmit2) {
     m_default_queue->wait();
     m_errorMonitor->VerifyFound();
 }
+
+TEST_F(NegativeRenderPass, MissingNestedCommandBuffersFeature) {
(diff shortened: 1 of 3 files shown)


spencer-lunarg (approved):
such a nice and simple VU, thanks for adding

tests/unit/render_pass.cpp
```diff
@@ -4312,3 +4312,21 @@ TEST_F(NegativeRenderPass, RenderPassWithRenderPassStripedQueueSubmit2) {
     m_default_queue->wait();
     m_errorMonitor->VerifyFound();
 }
+
+TEST_F(NegativeRenderPass, MissingNestedCommandBuffersFeature) {
+    TEST_DESCRIPTION("Use VK_SUBPASS_CONTENTS_INLINE_AND_SECONDARY_COMMAND_BUFFERS_EXT when nextedCommandBuffers is not enabled");
+
+    AddRequiredExtensions(VK_KHR_CREATE_RENDERPASS_2_EXTENSION_NAME);
```
spencer-lunarg, line 4319: I love we can write a test like this in less than 100 lines now :parrot:

PR #7828, opened 2024-04-10
layers: Add explanation why to cache surface caps

artem-lunarg:
Based on this comment and the following discussion: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7820#discussion_r1559123395

@spencer-lunarg @Eearslya please let me know if this explanation makes sense.


Eearslya:
Sounds good to me, explains the issue well. WSI has always been the weird part of the API with probably the most spec-lawyering (except maybe synchronization).

spencer-lunarg (approved):

PR #7829, opened 2024-04-11
layers: Remove old mockicd workaround

spencer-lunarg:
Ran across this, I don't think there are actually any issues with Profiles/MockICD anymore

### DIFF
--- a/layers/containers/subresource_adapter.cpp
+++ b/layers/containers/subresource_adapter.cpp
@@ -282,16 +282,6 @@ ImageRangeEncoder::ImageRangeEncoder(const vvl::Image& image, const AspectParame
     VkImageSubresourceLayers subres_layers = {limits_.aspectMask, 0, 0, limits_.arrayLayer};
     linear_image_ = false;
 
-    // WORKAROUND for profile and mock_icd not containing valid VkSubresourceLayout yet. Treat it as optimal image.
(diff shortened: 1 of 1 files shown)


layers/containers/subresource_adapter.cpp
```diff
@@ -280,17 +280,6 @@ ImageRangeEncoder::ImageRangeEncoder(const vvl::Image& image, const AspectParame
     VkSubresourceLayout layout = {};
     VkImageSubresource subres = {};
     VkImageSubresourceLayers subres_layers = {limits_.aspectMask, 0, 0, limits_.arrayLayer};
-    linear_image_ = false;
```
artem-lunarg, line 283: I think this part still should be here. The reason why it did not change behavior of mockICD is another workaround on the line 294 that does  `linear_image_ = false;`. Yes, it unconditionally overwrites this logic (for now). But the part that is removed, it looks like it's the only place in the codebase that detects if the image is linear `linear_image_ = true;`, so we will need it. Yeah, it's a bit of a mess.

spencer-lunarg: fair, will leave this here

spencer-lunarg: >  Yeah, it's a bit of a mess.

:laughing:

PR #7830, opened 2024-04-11
layers: Create a core::CommandBuffer

spencer-lunarg:
Due to memory usage (https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7387) the `CommandBuffer` state object should be broken up as everyone inherent a lot of  unnecessary data

### DIFF
--- a/layers/core_checks/cc_state_tracker.h
+++ b/layers/core_checks/cc_state_tracker.h
@@ -1,0 +1,38 @@
+/* Copyright (c) 2015-2024 The Khronos Group Inc.
+ * Copyright (c) 2015-2024 Valve Corporation
+ * Copyright (c) 2015-2024 LunarG, Inc.
+ * Copyright (C) 2015-2024 Google Inc.
+ * Modifications Copyright (C) 2020 Advanced Micro Devices, Inc. All rights reserved.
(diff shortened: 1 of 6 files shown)


arno-lunarg (approved):

layers/core_checks/cc_state_tracker.h
```diff
@@ -0,0 +1,40 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+#include "state_tracker/state_tracker.h"
+#include "state_tracker/cmd_buffer_state.h"
+
+namespace vvl {
+namespace core {
```
jeremyg-lunarg, line 25: Why is the `core` namespace nested inside of `vvl`. I'm pretty sure the other derived state are not nested like this.

jeremyg-lunarg (approved):

layers/core_checks/cc_state_tracker.h
spencer-lunarg: good catch, ya, not sure what I was doing here, will remove so its just `core`

Issue #7737, opened 2024-03-21
GPU-Assisted seem to be checking the wrong descriptor set when binding sets before pipeline


AlexRouSg:
**Environment:**
 - OS:  Windows 10
 - GPU and driver version: NVIDIA GeForce GTX 1080 Ti Driver: NVIDIA 551.86
 - SDK or header version if building from repo: 1.3.280.0
 - Options enabled (synchronization, best practices, etc.): GPU-Assisted Preset

**Describe the Issue**

The following code triggers `VUID-vkCmdDraw-viewType-07752`.

```c++
vkCmdBindDescriptorSets();
vkCmdBindPipeline();
vkCmdDraw() ;// draw that uses 2D textures

vkCmdBindDescriptorSets();
vkCmdBindPipeline();
vkCmdDraw(); // draw that uses 3D textures
```
However swapping the order of `vkCmdBindDescriptorSets` and `vkCmdBindPipeline` does not.
```c++
vkCmdBindPipeline();
vkCmdBindDescriptorSets();
vkCmdDraw() ;// draw that uses 2D textures

vkCmdBindPipeline();
vkCmdBindDescriptorSets();
vkCmdDraw(); // draw that uses 3D textures
```
Removing the first draw call and related binding cmds that only uses 2D textures also does not trigger this VUID.

**Expected behavior**
This VUID should not trigger as I have verified the SPIRV manually that the second draw call does not contain a  `OpTypeImage 2D` nor does renderdoc detect a 3D texture being bound to the first draw call.

**Valid Usage ID**
VUID-vkCmdDraw-viewType-07752

ImageView type is VK_IMAGE_VIEW_TYPE_3D but the OpTypeImage has
(
    Dim = 2D
)
 
**Additional Info:**
Might be relevant but I am also using dynamic rendering + graphics pipeline library

spencer-lunarg:
@AlexRouSg we do shader instrumentation to detect if the texture is accessed or not. Can you provide the SPIR-V as it will be helpful to see if the bug is how we are instrumenting it

AlexRouSg:
@spencer-lunarg I have tested with a very minimal frag shader and here it is

```
; SPIR-V
; Version: 1.6
; Generator: Google Shaderc over Glslang; 11
; Bound: 34
; Schema: 0
               OpCapability Shader
          %2 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
[...]

PR #7838, opened 2024-04-12
gpu: Fix Descriptor Indexing assuming pipeline bound first

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7737

Before we only updated the `binding_req` (that part that maps the `OpVariable` from the pipeline to the `VkImage` in the descriptor set) at `vkCmdBindDescriptorSets` time, this caused the following to produce a false error

```
vkCmdBindPipeline(); // bad
vkCmdBindDescriptorSets();
vkCmdBindPipeline(); // good
vkCmdDraw()
```

This now updates the last `vkCmdBindDescriptorSets` if there is a `vkCmdBindPipeline` after it


jeremyg-lunarg (approved):
LGTM

PR #7839, opened 2024-04-12
layers: Unify all GPU-AV settings

spencer-lunarg:
- All GPU-AV settings now start with `VK_LAYER_GPUAV_` as it caused wasted time to assume this was the case as a few settings didn't follow this
  - Deprecation warning given to the few that were wrong
- It is not obvious how `vk_layer_settings` will underneath append the `VK_LAYER_`  suffix and uppercase the key name, so now all the names in `layer_options.cpp` match the actual environment variable name used to set it

### DIFF
--- a/layers/layer_options.cpp
+++ b/layers/layer_options.cpp
@@ -26,50 +26,63 @@
 #include "mimalloc-new-delete.h"
 #endif
 
-const char *SETTING_ENABLES = "enables";
-const char *SETTING_VALIDATE_BEST_PRACTICES = "validate_best_practices";
-const char *SETTING_VALIDATE_BEST_PRACTICES_ARM = "validate_best_practices_arm";
-const char *SETTING_VALIDATE_BEST_PRACTICES_AMD = "validate_best_practices_amd";
-const char *SETTING_VALIDATE_BEST_PRACTICES_IMG = "validate_best_practices_img";
-const char *SETTING_VALIDATE_BEST_PRACTICES_NVIDIA = "validate_best_practices_nvidia";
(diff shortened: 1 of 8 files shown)


layers/layer_options.cpp
```diff
@@ -26,50 +26,63 @@
...
-const char *SETTING_GPUAV_SELECT_INSTRUMENTED_SHADERS = "select_instrumented_shaders";
-const char *SETTING_GPUAV_MAX_BUFFER_DEVICE_ADDRESS_BUFFERS = "gpuav_max_buffer_device_addresses";
-const char *SETTING_GPUAV_DEBUG_VALIDATE_INSTRUMENTED_SHADERS = "gpuav_debug_validate_instrumented_shaders";
-const char *SETTING_GPUAV_DEBUG_DUMP_INSTRUMENTED_SHADERS = "gpuav_debug_dump_instrumented_shaders";
+// To enable use these names to match
+//   example: VK_LAYER_PRINTF_TO_STDOUT=1
+// This works because the key is "printf_to_stdout"
+// From here vk_layer_setting (in VUL) will do the logic:
+//   "VK_LAYER_" + toUpper(key)
+// to create the variable that needs to be set
```
arno-lunarg, line 34:
```suggestion
// To enable "my_setting",
// Set env var VK_LAYER_MY_SETTING to 1
//
// To quickly be able to find the env var corresponding to a setting,
// the following `const char*` holding setting names match their corresponding environment variable
```

arno-lunarg (approved):
One nit!

PR #7852, opened 2024-04-16
tests: Update queue family helpers

artem-lunarg:
Need to write some tests that work with different queues and it's a good opportunity to cleanup some helpers.

### DIFF
--- a/tests/framework/binding.cpp
+++ b/tests/framework/binding.cpp
@@ -317,7 +317,7 @@ const Device::QueueFamilyQueues &Device::queue_family_queues(uint32_t queue_fami
     return queue_families_[queue_family];
 }
 
-std::optional<uint32_t> Device::QueueFamilyMatching(VkQueueFlags with, VkQueueFlags without, bool all_bits) {
+std::optional<uint32_t> Device::QueueFamilyMatching(VkQueueFlags with, VkQueueFlags without, bool all_bits) const {
     for (uint32_t i = 0; i < phy_.queue_properties_.size(); i++) {
         const auto flags = phy_.queue_properties_[i].queueFlags;
         const bool matches = all_bits ? (flags & with) == with : (flags & with) != 0;
@@ -326,6 +326,14 @@ std::optional<uint32_t> Device::QueueFamilyMatching(VkQueueFlags with, VkQueueFl
         }
     }
     return {};
+}
+
+std::optional<uint32_t> Device::ComputeQueueFamily() const {
+    return QueueFamilyMatching(VK_QUEUE_COMPUTE_BIT, VK_QUEUE_GRAPHICS_BIT);
+}
+
+std::optional<uint32_t> Device::TransferQueueFamily() const {
+    return QueueFamilyMatching(VK_QUEUE_TRANSFER_BIT, VK_QUEUE_GRAPHICS_BIT | VK_QUEUE_COMPUTE_BIT);
 }
 
 bool Device::IsEnabledExtension(const char *extension) const {
--- a/tests/framework/binding.cpp
+++ b/tests/framework/binding.cpp
@@ -285,8 +285,6 @@ void Device::init_queues(const VkDeviceCreateInfo &info) {
         QueueFamilyQueues &queue_storage = queue_families_[queue_family_i];
         queue_storage.reserve(queue_create_info.queueCount);
         for (uint32_t queue_i = 0; queue_i < queue_create_info.queueCount; ++queue_i) {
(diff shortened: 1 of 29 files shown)


tests/framework/binding.cpp
```diff
@@ -311,25 +309,54 @@ void Device::init_queues(const VkDeviceCreateInfo &info) {
...
-    ASSERT_TRUE(!queues_[GRAPHICS].empty() || !queues_[COMPUTE].empty() || !queues_[DMA].empty() || !queues_[SPARSE].empty());
+    ASSERT_TRUE(!queues_[GRAPHICS].empty() || !queues_[COMPUTE].empty() || !queues_[TRANSFER].empty() || !queues_[SPARSE].empty());
 }
 
 const Device::QueueFamilyQueues &Device::queue_family_queues(uint32_t queue_family) const {
     assert(queue_family < queue_families_.size());
     return queue_families_[queue_family];
 }
 
-std::optional<uint32_t> Device::QueueFamilyMatching(VkQueueFlags with, VkQueueFlags without, bool all_bits) {
```
artem-lunarg, line 322: With this update no more places left that need to match any bit (`all_bits == false`). If there is a need `QueueFamily` can be called multiple times.

tests/framework/binding.h
```diff
@@ -220,22 +220,29 @@ class Device : public internal::Handle<VkDevice> {
     // vkGetDeviceProcAddr()
     PFN_vkVoidFunction get_proc(const char *name) const { return vk::GetDeviceProcAddr(handle(), name); }
 
-    // vkGetDeviceQueue()
-    const std::vector<Queue *> &graphics_queues() const { return queues_[GRAPHICS]; }
-    const std::vector<Queue *> &compute_queues() { return queues_[COMPUTE]; }
-    const std::vector<Queue *> &dma_queues() { return queues_[DMA]; }
-    const std::vector<Queue *> &sparse_queues() { return queues_[SPARSE]; }
+    const std::vector<Queue *> &QueuesWithGraphicsCapability() const { return queues_[GRAPHICS]; }
```
artem-lunarg, line 223: This commit comments the rename: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7852/commits/4fd928328e6bc81a057a34e7f4470b74451e4d7e

tests/unit/others.cpp
```diff
@@ -817,7 +817,7 @@ TEST_F(VkLayerTest, InvalidAllocationCallbacks) {
 
     RETURN_IF_SKIP(Init());
 
-    const std::optional queueFamilyIndex = DeviceObj()->QueueFamilyMatching(0, 0, true);
```
artem-lunarg, line 820: This expression meant any queue, just use the graphics queue to make this simpler, should not affect test coverage.

tests/framework/binding.h
```diff
@@ -220,22 +220,29 @@ class Device : public internal::Handle<VkDevice> {
...
-        return QueueFamilyMatching(VkQueueFlags(0), capabilities, true /* all_bits with */);
-    }
+    // Queue family that has "with" capabilities and optionally without "without" capabilities.
+    std::optional<uint32_t> QueueFamily(VkQueueFlags with, VkQueueFlags without = 0) const;
+
+    // Queue family that does not have "without" capabilities
+    std::optional<uint32_t> QueueFamilyWithoutCapabilities(VkQueueFlags without);
+
+    // Dedicated compute queue family: has compute but no graphics
+    std::optional<uint32_t> ComputeQueueFamily() const;
```
spencer-lunarg, line 239:
```suggestion
    std::optional<uint32_t> ComputeOnlyQueueFamily() const;
```

might be easier to tell right away it isn't "any queue that has compute"

spencer-lunarg (approved):
The queue family logic in the tests needed this!

LGTM

tests/framework/binding.h
artem-lunarg: Updated

artem-lunarg: I was thinking about "only" during development. I agree it gives proper intuition when reading code.  Was not sure if someone wants to use "only" queue to check for other caps and whether "only" can be a bit misleading because "only" queues can also have other caps. We probably can introduce more helpers if needed.

One option for vvl/driver apps is this naming: `Dedicated{Compute/Transfer}Queue`

Another option is `Async{Compute/Transfer}Queue`, but that's might be more specific for graphics apps, which usually use "only" queues for async operations. When async is not needed, they likely will run, for example, compute commands on the graphics queue, even if there is a compute-only queue.

Issue #7835, opened 2024-04-11
VUID-VkGraphicsPipelineCreateInfo-stageCount-09587 not checked for pipeline link step


rg3igalia:
**Environment:**
 - OS: Fedora Linux 39
 - GPU and driver version: Intel/Mesa (ANV)
 - SDK or header version if building from repo: 301ab9c429271c51a777ff1cf9c6ed79efad50e1
 - Options enabled (synchronization, best practices, etc.): defaults+synchronization

**Describe the Issue**

Not sure if to file this as a bug or a lack of coverage. The recently introduced `VUID-VkGraphicsPipelineCreateInfo-stageCount-09587` made all "unused shader stages" CTS tests illegal (`dEQP-VK.pipeline.*.misc.unused_shader_stages_*`). The new VUID is emitted for all tests matching the previous pattern. These will have to be removed but, in the mean time, I discovered that the layers do not emit any validation error if extra stages are included in the final link step when 4 other libraries are used to create a final pipeline, which is a case that falls under the new VUID (since all state is included as libraries, the pipeline does not require pre-rasterization or fragment shader state).

If we comment out [lines 887 and 888 in vktPipelineMiscTests.cpp](https://github.com/KhronosGroup/VK-GL-CTS/blob/b47bb2277ddd332bc9998f76ef76c753d116e568/external/vulkancts/modules/vulkan/pipeline/vktPipelineMiscTests.cpp#L887) and [lines 957 and 958 in the same file](https://github.com/KhronosGroup/VK-GL-CTS/blob/b47bb2277ddd332bc9998f76ef76c753d116e568/external/vulkancts/modules/vulkan/pipeline/vktPipelineMiscTests.cpp#L957), the layers stop emitting any error, but they should still emit them unless we also comment out [lines 979 and 980](https://github.com/KhronosGroup/VK-GL-CTS/blob/b47bb2277ddd332bc9998f76ef76c753d116e568/external/vulkancts/modules/vulkan/pipeline/vktPipelineMiscTests.cpp#L979).

**Expected behavior**

Layers emit `VUID-VkGraphicsPipelineCreateInfo-stageCount-09587` unless all 3 blocks of lines are commented out.

**Valid Usage ID**

`VUID-VkGraphicsPipelineCreateInfo-stageCount-09587`

**Additional context**

N/A

spencer-lunarg:
[...]

PR #7855, opened 2024-04-17
layers: Check linked GPL for stageCount not zero

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7835

### DIFF
--- a/layers/core_checks/cc_pipeline_graphics.cpp
+++ b/layers/core_checks/cc_pipeline_graphics.cpp
@@ -112,7 +112,11 @@ bool CoreChecks::ValidateGraphicsPipeline(const vvl::Pipeline &pipeline, const L
         }
     }
 
-    skip |= ValidateShaderModuleId(pipeline, create_info_loc);
+    // pStages are ignored if not using one of these sub states
+    if (pipeline.OwnsSubState(pipeline.fragment_shader_state) || pipeline.OwnsSubState(pipeline.pre_raster_state)) {
+        skip |= ValidateShaderModuleId(pipeline, create_info_loc);
+    }
+
     skip |= ValidatePipelineCacheControlFlags(pipeline.create_flags, create_info_loc.dot(Field::flags),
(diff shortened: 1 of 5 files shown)


layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -112,7 +112,11 @@ bool CoreChecks::ValidateGraphicsPipeline(const vvl::Pipeline &pipeline, const L
         }
     }
 
-    skip |= ValidateShaderModuleId(pipeline, create_info_loc);
+    // pStages are ignored if not using one of these sub states
+    if (pipeline.OwnsSubState(pipeline.fragment_shader_state) || pipeline.OwnsSubState(pipeline.pre_raster_state)) {
```
spencer-lunarg, line 116: this is called from Compute/RTX pipeline so rather do the logic outside where I don't need to check if from graphics pipeline there

artem-lunarg, line 116: Was trying to understand `OwnSubState` from its comment and feel a bit confused:

`// Used to know if the pipeline substate is being created (as opposed to being linked)`

Okay, the code checks if `parent` is `this` so it has to be this means that "substate is being created" or "pipeline owns substate".

`// If the VUs says "created with"...`
Okay, looks like a good use for this function, since it checks that substate is being created.

`//...  then you should NOT use this function`

artem-lunarg (approved):

layers/core_checks/cc_pipeline_graphics.cpp
spencer-lunarg: so the `OwnsSubState` was there when we first implemented GPL, the spec has since been reworked to specify when things are related to "created with" vs "linked with"

... I plan to just make a PR to unify the naming in the VVL to match the spec now as it is not obvious by any means at the moment

artem-lunarg: I see, thanks for the explanation

PR #7856, opened 2024-04-17
layer: Stop checking removed VU

gnl21:
VUID-vkCmdDraw-None-04912 has been removed from the Vulkan spec because it's redundant. Remove the check from the layer and update the tests.

### DIFF
--- a/layers/core_checks/cc_cmd_buffer_dynamic.cpp
+++ b/layers/core_checks/cc_cmd_buffer_dynamic.cpp
@@ -164,12 +164,8 @@ bool CoreChecks::ValidateGraphicsDynamicStateSetStatus(const LastBound& last_bou
 
     // VK_EXT_vertex_input_dynamic_state
     {
-        if (pipeline.IsDynamic(VK_DYNAMIC_STATE_VERTEX_INPUT_EXT) &&
+        if (!pipeline.IsDynamic(VK_DYNAMIC_STATE_VERTEX_INPUT_EXT) &&
             pipeline.IsDynamic(VK_DYNAMIC_STATE_VERTEX_INPUT_BINDING_STRIDE_EXT)) {
-            skip |= ValidateDynamicStateIsSet(state_status_cb, CB_DYNAMIC_STATE_VERTEX_INPUT_EXT, objlist, loc,
-                                              vuid.vertex_input_04912);
-        } else if (!pipeline.IsDynamic(VK_DYNAMIC_STATE_VERTEX_INPUT_EXT) &&
-                   pipeline.IsDynamic(VK_DYNAMIC_STATE_VERTEX_INPUT_BINDING_STRIDE_EXT)) {
             skip |= ValidateDynamicStateIsSet(state_status_cb, CB_DYNAMIC_STATE_VERTEX_INPUT_BINDING_STRIDE, objlist, loc,
                                               vuid.vertex_input_binding_stride_04913);
         }
--- a/layers/drawdispatch/drawdispatch_vuids.cpp
+++ b/layers/drawdispatch/drawdispatch_vuids.cpp
@@ -68,7 +68,6 @@ struct DispatchVuidsCmdDraw : DrawDispatchVuid {
         logic_op_04878                           = "VUID-vkCmdDraw-logicOp-04878";
         primitive_restart_enable_04879           = "VUID-vkCmdDraw-None-04879";
         primitive_restart_list_09637             = "VUID-vkCmdDraw-None-09637";
-        vertex_input_04912                       = "VUID-vkCmdDraw-None-04912";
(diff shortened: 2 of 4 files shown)


gnl21:
Looking at the CI results, the first test that it is complaining about is crashing for me locally with or without my change. I'm not sure what's going on with it. Can someone confirm that it is definitely my change that is causing this problem? If so then I'll look into it.

I can't test the GPU-AV tests at the moment because of problems on my implementation that I think are caused by an unrelated issue in the gpu-av code.

artem-lunarg:
@gnl21 Could you please rebase your fix on the latest code, the pushed branch is quite old. One thing to update is that the latest version of the test uses `SetDesiredError` instead of `SetDesiredFailureMsg`. It's not clear what happened on CI, could be sporadic failure.

layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -164,12 +164,8 @@ bool CoreChecks::ValidateGraphicsDynamicStateSetStatus(const LastBound& last_bou
 
     // VK_EXT_vertex_input_dynamic_state
     {
-        if (pipeline.IsDynamic(VK_DYNAMIC_STATE_VERTEX_INPUT_EXT) &&
+        if (!pipeline.IsDynamic(VK_DYNAMIC_STATE_VERTEX_INPUT_EXT) &&
             pipeline.IsDynamic(VK_DYNAMIC_STATE_VERTEX_INPUT_BINDING_STRIDE_EXT)) {
-            skip |= ValidateDynamicStateIsSet(state_status_cb, CB_DYNAMIC_STATE_VERTEX_INPUT_EXT, objlist, loc,
-                                              vuid.vertex_input_04912);
```
gnl21, line 170: This variable `vuid.vertex_input_04912` is unused now. Is that code autogenerated or should I remove that as well.

gnl21:
> @gnl21 Could you please rebase your fix on the latest code, the pushed branch is quite old.

Done, thanks.

> When `VUID-vkCmdDraw-None-04912` was removed from the spec, do you know which commit on gitlab?

It's !6604, which should be in the next released spec. I think it should be fine to accept this in advance of that though, because the check removed is redundant and if you look a few lines below this change you'll see the same check that I'm removing being executed unconditionally with a different VUID.

layers/core_checks/cc_cmd_buffer_dynamic.cpp
artem-lunarg: We add them manually (there are few of them in different structs), please remove them. Thanks!

artem-lunarg:
Should we also check for `pipeline.IsDynamic(VK_DYNAMIC_STATE_VERTEX_INPUT_EXT)` for `vertex_input_04914` that is required by this VU? I can't see how the previous condition ensures it's true, but also not an expert in this code (FYI @spencer-lunarg).

gnl21:
I don't know, and I would say that fixing that check could be a separate issue, but I looked at this when making the change and decided it was probably correct.

Lines 56-7 of the changed file have:
```
    // build the mask of what has been set in the Pipeline, but yet to be set in the Command Buffer
    const CBDynamicFlags state_status_cb = ~((cb_state.dynamic_state_status.cb ^ pipeline.dynamic_state) & pipeline.dynamic_state);
```
so I think that the check is checking whether the state was set either in the pipeline (so is non-dynamic), or by the setter command. That would mean that the unconditional check would be correct,

Almost all the other checks in this function are unconditional, which is consistent with that. The conditional in the check for 04913 is only because that state can be set in some non-obvious ways, which need to be excluded. It might actually be nicer to have the `SetVertexInputEXT` command mark the stride dynamic state as set. I think the code would be a bit nicer then, but would be a bit less like the spec, so maybe it wouldn't be an improvement,

layers/core_checks/cc_cmd_buffer_dynamic.cpp
gnl21: Done.

artem-lunarg:
> so I think that the check is checking whether the state was set either in the pipeline (so is non-dynamic), or by the setter command. That would mean that the unconditional check would be correct,

Thanks, that's good explanation.

Personally for me the above boolean expression is easier to read in this form:

```
    const CBDynamicFlags state_status_cb = ~(cb_state.dynamic_state_status.cb ^ pipeline.dynamic_state) | /*setter command*/
                                           ~ pipeline.dynamic_state; /*non-dynamic*/
```

artem-lunarg (approved):

ziga-lunarg (approved):

spencer-lunarg (approved):
Thanks @gnl21 (and @artem-lunarg for helping)

I was gone this week and didn't see the spec change, but everything looks good and thanks for making the PR ahead of time for me

PR #7863, opened 2024-04-18
tests: Add second helper queue

artem-lunarg:
Sometimes the tests check for the second graphics queue if 2 queues are needed.
This can decrease test coverage. For example, AMD windows drivers usually expose a single graphics queue.

This PR adds one more helper queue in addition to the default one. It will be a graphics queue if the second graphics queue is available, otherwise it's a compute if available, then transfer. Null when only 1 queue is exposed.

It can be used in situations when the test can work with the queues with different capabilities. Even if the queue with a specific capability is needed it might be easier to check`m_second_queue`'s  `VkQueueFlags` than iterating over `QueuesWithXXXCapability()` in order to skip the default queue.

### DIFF
--- a/tests/unit/sync_object_positive.cpp
+++ b/tests/unit/sync_object_positive.cpp
@@ -2398,14 +2398,12 @@ TEST_F(PositiveSyncObject, TwoQueuesReuseBinarySemaphore) {
     TEST_DESCRIPTION("Use binary semaphore with the first queue then re-use on a different queue");
     RETURN_IF_SKIP(Init());
 
-    if ((m_device->phy().queue_properties_.empty()) || (m_device->phy().queue_properties_[0].queueCount < 2)) {
+    if (!m_second_queue) {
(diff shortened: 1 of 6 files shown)


tests/unit/sync_object_positive.cpp
```diff
@@ -2398,14 +2398,12 @@ TEST_F(PositiveSyncObject, TwoQueuesReuseBinarySemaphore) {
     TEST_DESCRIPTION("Use binary semaphore with the first queue then re-use on a different queue");
     RETURN_IF_SKIP(Init());
 
-    if ((m_device->phy().queue_properties_.empty()) || (m_device->phy().queue_properties_[0].queueCount < 2)) {
```
artem-lunarg, line 2401: Used this test as an example (but in general need this for syncval work).
It was skipped previously on AMD hardware because of single graphics queue. Now dedicated compute queue is used as the additional queue.

spencer-lunarg (approved):

PR #7869, opened 2024-04-19
test: Set VK_LAYER_PATH automatically if missing

charles-lunarg:
Changes:
* Add mechanism to set VK_LAYER_PATH automatically if it isn't set or if it is set but doesn't contain a path to a valid layer manifest.
* Search VK_ADD_LAYER_PATH in addition to VK_LAYER_PATH when looking for the layer manifest
* Enhance searching of the layer manifest to include directory support rather than only allowing files
* Enhance VK_ICD_FILENAMES & VK_DRIVER_FILES checking to support files in addition to the existing support for directories

### DIFF
--- a/tests/framework/test_framework.cpp
+++ b/tests/framework/test_framework.cpp
@@ -18,6 +18,7 @@
 
 #include "test_framework.h"
 #include "render.h"
+#include "config.h"
 #include <filesystem>
 #include <cmath>
 #include <cstdarg>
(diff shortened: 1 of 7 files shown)


tests/framework/test_framework.cpp
```diff
@@ -71,68 +72,103 @@ static std::vector<std::string> GetVkEnvironmentVariable(const char *env_var) {
...
+    bool vk_layer_env_vars_present = false;
+    std::stringstream error_log;  // Build up error log in case the validation json cannot be found
+    for (const char *env_var : {"VK_LAYER_PATH", "VK_ADD_LAYER_PATH"}) {
+        const std::vector<std::string> vk_layer_paths = GetVkEnvironmentVariable(env_var);
+        if (!vk_layer_paths.empty()) {
+            vk_layer_env_vars_present = true;
         }
-        const std::string user_provided = "\n\nVK_LAYER_PATH = " + layer_path;
+        for (const std::string &layer_path : vk_layer_paths) {
+            const std::filesystem::path layer_dir(layer_path);
```
arno-lunarg, line 129: If I follow things correctly, `layer_dir` can be the path to a file, so the name is maybe confusing?

arno-lunarg (approved):
LGTM, one nit.
Thank you Charles!!!

tests/framework/test_framework.cpp
charles-lunarg: I can update the variable name to be clearer, but unfortunately VK_LAYER_PATH is just annoying in that it specifies a 'path' when a directory is also acceptable. Kinda confusing.

Also I should note that this *only* sets VK_LAYER_PATH. If you want to use other layers in addition, you'll need to set VK_LAYER_PATH to validation yourself (in addition to the other layers you want). This was the easiest way to get it to be automatic for the simple case while allowing more interesting cases in the future.

charles-lunarg:
Went ahead and updated the docs - realized that the docs shouldn't lie about VK_LAYER_PATH...

tests/framework/test_framework.cpp
arno-lunarg: The way VK_LAYER_PATH is automatically set or not is warranted I think, those additions you are making here are what we truly want, no need for more sophisticated behaviors!

charles-lunarg: Glad you think so - I’d definitely want to make the average dev’s experience running tests easier, and didn’t want to get in the way of more complex setups.

PR #7873, opened 2024-04-19
layers: Fix sync stage for EndRendering resolve of depth and stencil textures

nickclark2016:
Issue: Validation layers were not taking into account the image type when passing the sync stage to hazard detection. This would cause a false positive when resolving non-color texture attachments.

Fix: Switch between COLOR_ATTACHMENT_* and LATE_FRAGMENT_TESTS_DEPTH_STENCIL_* depending on if the texture attachment is a color or depth/stencil texture.

### DIFF
--- a/layers/sync/sync_commandbuffer.cpp
+++ b/layers/sync/sync_commandbuffer.cpp
@@ -224,6 +224,8 @@ bool CommandBufferAccessContext::ValidateEndRendering(const ErrorObject &error_o
             if (attachment.resolve_gen) {
                 const bool is_color = attachment.type == syncval_state::AttachmentType::kColor;
                 const SyncOrdering kResolveOrder = is_color ? kColorResolveOrder : kDepthStencilResolveOrder;
+                const SyncStageAccessIndex kResolveRead = is_color ? kColorResolveRead : kDepthStencilResolveRead;
+
                 // The logic about whether to resolve is embedded in the Attachment constructor
                 assert(attachment.view);
                 HazardResult hazard = access_context->DetectHazard(attachment.view_gen, kResolveRead, kResolveOrder);
@@ -234,6 +236,8 @@ bool CommandBufferAccessContext::ValidateEndRendering(const ErrorObject &error_o
                                                   attachment.info.resolveMode);
                 }
                 if (!skip) {
+                    const SyncStageAccessIndex kResolveWrite = is_color ? kColorResolveWrite : kDepthStencilResolveWrite;
+
                     hazard = access_context->DetectHazard(*attachment.resolve_gen, kResolveWrite, kResolveOrder);
                     if (hazard.IsHazard()) {
                         Location loc = attachment.GetLocation(error_obj.location, i);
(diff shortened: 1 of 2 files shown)


artem-lunarg:
@nickclark2016 I don't think this change is correct, please refer to section https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#renderpass-resolve-operations about details of resolve operation. 

Spefically this part:
>  Access to the single sample attachment can be synchronized with VK_ACCESS_COLOR_ATTACHMENT_READ_BIT and VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT. **These pipeline stage and access types are used whether the attachments are color or depth/stencil attachments**.

It's unintuitive but for both color and *depth* resolve operations happen on the COLOR_ATTACHMENT state and use color accesses. Potential validation errors is probably should be fixed on the application side.

nickclark2016:
If that's the case, I'd like to point out a workflow I have where the spec being worded like this makes it awkward (though there may be a better way to do what I'm doing, which I'll happily take suggestions for as well).

I have a MS depth stencil attachment that needs to be resolved. I use a pipeline barrier to ensure the resolve target is no longer in use, then I pass that image as the resolve target to vkCmdBeginRendering.

In an ideal world, my barrier would be set up like so:

```cpp
VkImageMemoryBarrier2 pre_resolve_barrier = {
    .srcStageMask = prior_stage_usage,
    .srcAccessMask = prior_usage.access_mask
    .dstStageMask = is_color ? COLOR_ATTACHMENT_OUTPUT : LATE_FRAGMENT_TEST,
    .dstAccessMask = is_color ? COLOR_ATTACHMENT_WRITE : DEPTH_ATTACHMENT_WRITE
    .newLayout = is_color ? COLOR_ATTACHMENT_OPTIMAL : DEPTH_ATTACHMENT_OPTIMAL
};
```

However, based on what you're telling me, I need to adjust my code to as follows:

```cpp
VkImageMemoryBarrier2 pre_resolve_barrier = {
    .srcStageMask = prior_stage_usage,
    .srcAccessMask = prior_usage.access_mask
    .dstStageMask = COLOR_ATTACHMENT_OUTPUT ,
    .dstAccessMask = COLOR_ATACHMENT_WRITE
    .newLayout = is_color ? COLOR_ATTACHMENT_OPTIMAL : ATTACHMENT_OPTIMAL
};
```

To me, this feels awkward, as due to the fact that I have to now set my `dstAccessMask` to `COLOR_ATTACHMENT_WRITE`, I then need to change my `dstStageMask` to `COLOR_ATTACHMENT_OUTPUT` and because I've change that, the depth layouts are no longer valid and I need to change my `newLayout` to a more general layout.

artem-lunarg:
> However, based on what you're telling me, I need to adjust my code to as follows:
VkImageMemoryBarrier2 pre_resolve_barrier = {
    .srcStageMask = prior_stage_usage,
    .srcAccessMask = prior_usage.access_mask
    .dstStageMask = COLOR_ATTACHMENT_OUTPUT ,
    .dstAccessMask = COLOR_ATACHMENT_WRITE
    .newLayout = is_color ? COLOR_ATTACHMENT_OPTIMAL : ATTACHMENT_OPTIMAL
};

That's not what I meant, resolve should not affect *destintation* part of the barrier. If you do at first resolve and then use the result of resolve as attachment during rendering and put a barrier between resolve and rendering, then barrier should wait for resolve operation, it means resolve stage/access should be *source* side of the barrier. And then the *destination side* is what was in the original example, it's not affected by the resolve.

One resource that can be useful to get answers to Vulkan related questions is Vulkan discord. There is a group related to synchronization too (https://discord.com/invite/vulkan)

nickclark2016:
The barrier is before the resolve.

artem-lunarg:
@nickclark2016 sorry, I misunderstood your setup. Often you don't need to synchronize with resolve if it's not the first access that is performed by the render pass. If `loadOp` for the depth attachment is not NONE, then the first access performed by the render pass is from the loadOp operation, and synchronization should be done with corresponding accesses. The accesses performed by the loadOp are described here: https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#renderpass-load-operations. It's likely it will be something like in your ideal scenario but the stage is `VK_PIPELINE_STAGE_EARLY_FRAGMENT_TESTS_BIT` (according to section 8.5) and the access mask depends on specific loadOp.

PR #7876, opened 2024-04-20
layers: Add Draw VK_EXT_legacy_dithering VU

spencer-lunarg:
Adds `VUID-vkCmdDraw-None-09642`/`VUID-vkCmdDraw-None-09643` which were added in 1.3.283 spec

### DIFF
--- a/layers/core_checks/cc_pipeline_graphics.cpp
+++ b/layers/core_checks/cc_pipeline_graphics.cpp
@@ -3950,31 +3950,29 @@ bool CoreChecks::ValidatePipelineDynamicRenderpassDraw(const LastBound &last_bou
             }
         }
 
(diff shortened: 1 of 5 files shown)


layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -3950,31 +3950,29 @@ bool CoreChecks::ValidatePipelineDynamicRenderpassDraw(const LastBound &last_bou
             }
         }
 
-        if (cb_state.activeRenderPass->UsesDynamicRendering()) {
```
spencer-lunarg, line 3953: This function is already guarded by `UsesDynamicRendering` so this is not required, saw and cleaned up

PR #7877, opened 2024-04-20
layers: Add dynamic state VK_EXT_conservative_rasterization

spencer-lunarg:
Adds `VUID-VkGraphicsPipelineCreateInfo-pDynamicState-09639` from 1.3.283 spec

### DIFF
--- a/layers/core_checks/cc_pipeline_graphics.cpp
+++ b/layers/core_checks/cc_pipeline_graphics.cpp
@@ -2007,6 +2007,14 @@ bool CoreChecks::ValidateGraphicsPipelineRasterizationState(const vvl::Pipeline 
                 }
             }
         }
+    } else {
+        if (pipeline.IsDynamic(VK_DYNAMIC_STATE_CONSERVATIVE_RASTERIZATION_MODE_EXT) &&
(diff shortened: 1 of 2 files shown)


layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -2007,6 +2007,14 @@ bool CoreChecks::ValidateGraphicsPipelineRasterizationState(const vvl::Pipeline
                 }
             }
         }
+    } else {
+        if (pipeline.IsDynamic(VK_DYNAMIC_STATE_CONSERVATIVE_RASTERIZATION_MODE_EXT) &&
```
arno-lunarg, line 2011: The "If the pipeline requires [pre-rasterization shader state](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#pipelines-graphics-subsets-pre-rasterization)," part of the VU has already been handled when getting here?

arno-lunarg (approved):

layers/core_checks/cc_pipeline_graphics.cpp
spencer-lunarg: yes, otherwise `RasterizationState` is ignored

Issue #7874, opened 2024-04-19
Deadlock (WriteLock followed by ReadLock) for secondary command buffers containing vkCmdBeginQuery


akeley98NV:
**Environment:**
 - OS: Ubuntu 22.04
 - GPU and driver version: RTX A5000 535.43.22
 - SDK or header version if building from repo: 0409ac63ab1a178bed275cf13acceb21d06f6091
 - Options enabled (synchronization, best practices, etc.): Whatever the NegativeQuery tests do by default

**Describe the Issue**

If I record `vkCmdBeginQuery` or such to secondary command buffers, this can cause the validation layers to deadlock or crash in the `is_query_updated_after` helper in `Queue::Retire`. This is due to the helper being run while the `WriteLock` (a.k.a. `std::unique_lock`) is held but itself requiring the `ReadLock` (a.k.a. `std::shared_lock`). Please see the repro case I attached.

**Code Analysis**

In `Queue::Retire` we have this code that runs over each secondary command buffer referenced

```C++
        for (auto *secondary_cmd_buffer : cb_state->linkedCommandBuffers) {
            auto secondary_guard = secondary_cmd_buffer->WriteLock();
            secondary_cmd_buffer->Retire(submission.perf_submit_pass, is_query_updated_after);
        }
```
Note that the `WriteLock` is held on the secondary command buffer when this is done.

The `Retire` function takes the `is_query_updated_after` lambda which itself calls `next_cb_state->UpdatesQuery`, if we look inside that function we see

```C++
bool CommandBuffer::UpdatesQuery(const QueryObject &query_obj) const {
    // Clear out the perf_pass from the caller because it isn't known when the command buffer is recorded.
    auto key = query_obj;
    key.perf_pass = 0;
    for (auto *sub_cb : linkedCommandBuffers) {
        auto guard = sub_cb->ReadLock();   // <<< RISKY
        if (sub_cb->updatedQueries.find(key) != sub_cb->updatedQueries.end()) {
            return true;
        }
    }
    return updatedQueries.find(key) != updatedQueries.end();
}
```

The line marked as `<<< RISKY` crashes the validation layers if any secondary command buffer encountered is the same as the one where the `WriteLock` is held. Which is quite likely to be the case for a real application given re-use is the main purpose of secondary command buffers.

[...]

PR #7880, opened 2024-04-21
layers: Fix deadlock triggered by queries updated in secondary command buffers.

akeley98NV:
Fix for #7874

### DIFF
--- a/layers/state_tracker/cmd_buffer_state.cpp
+++ b/layers/state_tracker/cmd_buffer_state.cpp
@@ -395,7 +395,7 @@ void CommandBuffer::BeginQuery(const QueryObject &query_obj) {
         SetQueryState(QueryObject(query_obj, perfQueryPass), QUERYSTATE_RUNNING, localQueryToStateMap);
         return false;
     });
-    updatedQueries.insert(query_obj);
+    recursiveUpdatedQueries.insert(query_obj);
     if (query_obj.inside_render_pass) {
         renderPassQueries.insert(query_obj);
     }
@@ -407,7 +407,7 @@ void CommandBuffer::EndQuery(const QueryObject &query_obj) {
                                           uint32_t perfQueryPass, QueryMap *localQueryToStateMap) {
         return SetQueryState(QueryObject(query_obj, perfQueryPass), QUERYSTATE_ENDED, localQueryToStateMap);
     });
-    updatedQueries.insert(query_obj);
+    recursiveUpdatedQueries.insert(query_obj);
     if (query_obj.inside_render_pass) {
         renderPassQueries.erase(query_obj);
     }
@@ -417,13 +417,7 @@ bool CommandBuffer::UpdatesQuery(const QueryObject &query_obj) const {
     // Clear out the perf_pass from the caller because it isn't known when the command buffer is recorded.
     auto key = query_obj;
     key.perf_pass = 0;
-    for (auto *sub_cb : linkedCommandBuffers) {
-        auto guard = sub_cb->ReadLock();
-        if (sub_cb->updatedQueries.find(key) != sub_cb->updatedQueries.end()) {
-            return true;
-        }
-    }
-    return updatedQueries.find(key) != updatedQueries.end();
+    return recursiveUpdatedQueries.find(key) != recursiveUpdatedQueries.end();
 }
 
 static bool SetQueryStateMulti(VkQueryPool queryPool, uint32_t firstQuery, uint32_t queryCount, uint32_t perfPass, QueryState value,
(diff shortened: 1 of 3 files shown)


spencer-lunarg (approved):
@akeley98NV thanks for the fix (and tests)

1. You need to run `clang-format` as CI is mad about it https://github.com/KhronosGroup/Vulkan-ValidationLayers/actions/runs/8775393117/job/24077674023?pr=7880

2. @artem-lunarg I would like to get your review on this as well

akeley98NV:
Yeah it's mad because my TEST_DESCRIPTION is too long, is it acceptable to split it across 2 lines? (Some projects don't like this as it breaks grepability). Or suggestion for a shorter description?

spencer-lunarg:
> Yeah it's mad because my TEST_DESCRIPTION is too long, is it acceptable to split it across 2 lines? (Some projects don't like this as it breaks grepability). Or suggestion for a shorter description?

It is ok, we have `clang-format` to enforce this, the description is normally used when someone fails the test or has to debug it... so a 2nd line is totally ok!

arno-lunarg (approved):

artem-lunarg:
@akeley98NV Thanks for the detailed analysis that shows the problem. I updated the test a little bit. Made it positive since we do not try to reproduce some VU error, and also used the existing helpers to make it smaller. Could you please remove the existing test from `query.cpp` and put the following updated version into a different file `query_positive.cpp`. I tested it on windows, it hangs with original code (no crash with windows runtime but a deadlock) and works with your fix.

```

// https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7874
TEST_F(PositiveQuery, ReuseSecondaryWithQueryCommand) {
    TEST_DESCRIPTION("Regression test for a deadlock when secondary command buffer is reused and records a query command");

    RETURN_IF_SKIP(Init());
    if (HasZeroTimestampValidBits()) {
        GTEST_SKIP() << "Device graphic queue has timestampValidBits of 0, skipping.";
    }

    vkt::QueryPool query_pool(*m_device, VK_QUERY_TYPE_TIMESTAMP, 1);

    vkt::CommandBuffer secondary_buffer(*m_device, m_commandPool, VK_COMMAND_BUFFER_LEVEL_SECONDARY);
    secondary_buffer.begin(VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT);
    vk::CmdWriteTimestamp(secondary_buffer.handle(), VK_PIPELINE_STAGE_ALL_GRAPHICS_BIT, query_pool, 0);
    secondary_buffer.end();

    m_commandBuffer->begin(VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT);
    vk::CmdResetQueryPool(m_commandBuffer->handle(), query_pool, 0, 1);
    vk::CmdExecuteCommands(m_commandBuffer->handle(), 1, &secondary_buffer.handle());
    m_commandBuffer->end();

    m_default_queue->submit(*m_commandBuffer);

    // Submit the command buffer again, making query pool in use and invalid to destroy
    m_default_queue->submit(*m_commandBuffer);

    m_default_queue->wait();
}
```

artem-lunarg:
I'd like to suggest a one line fix (https://github.com/artem-lunarg/Vulkan-ValidationLayers/commit/3dcbec707095a36c567ae4a3a530106cd7bfc7ed) (cc @jeremyg-lunarg).

Current vvl has this function:
```
bool CommandBuffer::UpdatesQuery(const QueryObject &query_obj) const {
    // Clear out the perf_pass from the caller because it isn't known when the command buffer is recorded.
    auto key = query_obj;
    key.perf_pass = 0;
    for (auto *sub_cb : linkedCommandBuffers) {
        auto guard = sub_cb->ReadLock();
        if (sub_cb->updatedQueries.find(key) != sub_cb->updatedQueries.end()) {
            return true;
        }
    }
    // !! No lock for  THIS command buffer while accessing the following updatedQueries. !!
    return updatedQueries.find(key) != updatedQueries.end(); 
}
```

We do lock the secondary command buffers but we do not lock *this* command buffer. *this* command buffer is a primary command buffer from one of the subsequent submissions (not from the current one that is being retired). I can speculate that the reason we do not lock it is because we assume that command buffers can not be modified while *pending* and probably we have validation for that, this code might rely on that assumption.

The part that locks secondary command buffer was added separately here (SHA=221bb37645c72db356277aa1353af4e07a75b677), my best guess it was added with the best intention to make code safer but it's not clear if there was an actual race (@jeremyg-lunarg do you have insights?).

My suggestion is to tread secondary command buffer the same way we tread primary command buffers and remove the line with a lock. This passes new regression test. If any issues we then we still need a different fix which also fixes access to *this* command buffer since it is also not locked.

akeley98NV:
If the maintainers think the original problematic lock can be removed, I'm certainly happy to have that change accepted instead of this one.

jeremyg-lunarg (approved):

jeremyg-lunarg:
I think either the solution in this PR or removing the sub cb locks as @artem-lunarg proposes are equivalent.  Really though the query state tracking code is in sorry state and needs some rethinking.

artem-lunarg:
I discussed this more with @jeremyg-lunarg and one of the goals is to to prevent crashes if external synchronization is violated (correctness is secondary then), in this regard removing lock might be unsafe. On the other hand we also don't have the lock for the primary command buffer (both approaches fix the issue but the external synchronization is also not handled ideally). I'll try to analyze this more during the week.

artem-lunarg:
Going to proceed with https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7909. @akeley98NV thanks for the help with this issue.

akeley98NV:
Sounds good, thanks!

PR #7884, opened 2024-04-22
printf: Add way to disconnect from chassis on abort

spencer-lunarg:
(plan to do with GPU-AV, but starting easy with DebugPrintF)

When we want to abort in GPU-AV/DebugPrintf it can be annoying to try and going `if (aborted) return;` everywhere. This adds a `ReleaseDeviceDispatchObject` that will remove the object from the chassis so that no function calls are ever called into after an abort

### DIFF
--- a/layers/gpu_validation/debug_printf.cpp
+++ b/layers/gpu_validation/debug_printf.cpp
@@ -25,10 +25,7 @@
 // Perform initializations that can be done at Create Device time.
 void debug_printf::Validator::CreateDevice(const VkDeviceCreateInfo *pCreateInfo, const Location &loc) {
     if (enabled[gpu_validation]) {
-        ReportSetupProblem(device, loc,
-                           "Debug Printf cannot be enabled when gpu assisted validation is enabled.  "
-                           "Debug Printf disabled.");
-        aborted = true;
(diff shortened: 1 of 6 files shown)


layers/gpu_validation/debug_printf.cpp
```diff
@@ -755,11 +741,12 @@ void debug_printf::CommandBuffer::Reset() {
...
-    if (debug_printf->aborted) {
-        return;
-    }
     for (auto &buffer_info : buffer_infos) {
         debug_printf->DestroyBuffer(buffer_info);
     }
     buffer_infos.clear();
 }
+
+void debug_printf::Validator::AbortLayer(LogObjectList objlist, const Location &loc, const char *specific_message, bool vma_fail) {
```
jeremyg-lunarg, line 750: I'm not crazy about the name of this method... at a glance it implies the whole layer will be gone. Maybe just Abort() or AbortValidation() or ???

jeremyg-lunarg: Or just leave it as ReportSetupProblem since this method doesn't do anything else

spencer-lunarg: ...ya, I think it was going to do something, but now I realize it is just a pass through, will just use `ReportSetupProblem`

jeremyg-lunarg (approved):

PR #7887, opened 2024-04-23
gpu: Fix concurrent access of cmb buffer resources

arno-lunarg:
After acquiring a command buffer's lock to call PostProcess on it, its resources could have been freed by another thread in the meantime, so PostProcess should account for that.
Bug showed up while replaying a capture of Doom 2

I looked at making sure we just do not end up having "empty" command buffers in `gpu_tracker::Queue::retiring_` but not sure it is worth to go down that route.
Would love @jeremyg-lunarg's eyes on that PR!

### DIFF
--- a/layers/gpu_validation/gpu_resources.h
+++ b/layers/gpu_validation/gpu_resources.h
@@ -24,8 +24,8 @@ class Validator;
 struct DescBindingInfo;
 
 struct DeviceMemoryBlock {
-    VkBuffer buffer;
-    VmaAllocation allocation;
+    VkBuffer buffer = VK_NULL_HANDLE;
+    VmaAllocation allocation = VK_NULL_HANDLE;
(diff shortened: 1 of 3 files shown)


jeremyg-lunarg:
concurrent use of the dispatch barrier cmd buffer seems safe to me.

arno-lunarg:
> concurrent use of the dispatch barrier cmd buffer seems safe to me.

Sorry I think I made you look at the wrong commit! It is this one that addresses the race: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7887/commits/899147687fab8ed8dbc4b6e016ef9a985d1a2ae6

spencer-lunarg (approved):
LGTM - @jeremyg-lunarg it is now a lot simple to review

jeremyg-lunarg (approved):
LGTM, thanks!

PR #7888, opened 2024-04-23
layers: Allow MapMemoryInfo::size to be allocationSize with VK_MEMORY_MAP_PLACED_BIT

gfxstrand:


layers/core_checks/cc_device_memory.cpp
```diff
@@ -1253,12 +1253,12 @@ bool CoreChecks::PreCallValidateMapMemory2KHR(VkDevice device, const VkMemoryMap
...
                 }
 
-                if (pMemoryMapInfo->size != VK_WHOLE_SIZE) {
+                if (pMemoryMapInfo->size != VK_WHOLE_SIZE && pMemoryMapInfo->size != mem_info->allocate_info.allocationSize) {
                     skip |= LogError("VUID-VkMemoryMapInfoKHR-flags-09572", pMemoryMapInfo->memory, info_loc.dot(Field::size),
                                      "(0x%" PRIx64
-                                     ") is not VK_WHOLE_SIZE while VK_MEMORY_MAP_PLACED_BIT_EXT is set and "
-                                     "memoryMapRangePlaced is not enabled.",
-                                     pMemoryMapInfo->size);
+                                     ") is not VK_WHOLE_SIZE or the size of memory (%" PRIu64
```
spencer-lunarg, line 1259:
```suggestion
                                     ") is not VK_WHOLE_SIZE and the size of memory (%" PRIu64
```

Am I wrong here, this wording doesn't match the logic above

gfxstrand: Ugh... english. The check is "is not foo and is not bar" while the text here is "is not foo or bar". I think the english means the same but if you've got a suggestion for how to phrase it better, I'm all ears.

spencer-lunarg (approved):

gfxstrand:
I think we want to leave that VU alone. However, I think we probably want a new VU for the `!memoryMapRangePlaced` case which requires `allocationSize` to be aligned.

gfxstrand:
Yeah, so I just reworked the validation code again as well as the VU on the internal spec MR. We need to require that the size is always aligned, regardless of whether it's `VK_WHOLE_SIZE` and implicitly taken from the allocation size or if it's an explicitly provided size. Given that the driver is replacing pages in the client's address space rather than allocating new ones, we don't want the driver doing the usual `nonCoherentAtomSize` trick of rounding it up. We want it to map EXACTLY the size specified.

layers/core_checks/cc_device_memory.cpp
```diff
@@ -1253,12 +1244,30 @@ bool CoreChecks::PreCallValidateMapMemory2KHR(VkDevice device, const VkMemoryMap
...
+                                     ") while VK_MEMORY_MAP_PLACED_BIT_EXT is set and memoryMapRangePlaced is not enabled.",
+                                     pMemoryMapInfo->size, mem_info->allocate_info.allocationSize);
+                }
+            }
+
+            if (pMemoryMapInfo->size == VK_WHOLE_SIZE) {
+                if (mem_info->allocate_info.allocationSize % phys_dev_ext_props.map_memory_placed_props.minPlacedMemoryMapAlignment != 0) {
+                    skip |= LogError("VUID-VkMemoryMapInfoKHR-XXXX", pMemoryMapInfo->memory, info_loc.dot(Field::size),
```
spencer-lunarg, line 1258:
```suggestion
                    // Being worked on in https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6616
                    skip |= LogError("UNASSIGNED-VkMemoryMapInfoKHR", pMemoryMapInfo->memory, info_loc.dot(Field::size),
```

This is how I do it as it helps be track these later

spencer-lunarg:
@gfxstrand now that the VUs are out in the 1.3.285 spec, if you could rebase/update from WIP, we can merge this in (even before VVL gets the new headers)

Issue #7892, opened 2024-04-23
Unexpected VUID-vkCmdDraw-None-04008


MennoVink:
**Environment:**
 - OS: Windows
 - GPU and driver version: NVIDIA GeForce RTX 3070 (driver version: 2311913472)
 - SDK or header version if building from repo: SDK 1.3.280.0 headers 1.3.275.0
 - Options enabled (synchronization, best practices, etc.):
```
VK_API_VERSION_1_2

VK_KHR_SURFACE_EXTENSION_NAME
VK_KHR_WIN32_SURFACE_EXTENSION_NAME
VK_KHR_SWAPCHAIN_EXTENSION_NAME
VK_EXT_DEBUG_UTILS_EXTENSION_NAME
VK_EXT_MEMORY_BUDGET_EXTENSION_NAME
VK_KHR_EXTERNAL_MEMORY_WIN32_EXTENSION_NAME
VK_KHR_EXTERNAL_SEMAPHORE_WIN32_EXTENSION_NAME
VK_KHR_EXTERNAL_FENCE_WIN32_EXTENSION_NAME

vk::PhysicalDeviceFeatures requiredFeatures;
requiredFeatures.fillModeNonSolid = true;
requiredFeatures.textureCompressionBC = true;
requiredFeatures.pipelineStatisticsQuery = true;

vk::PhysicalDeviceVulkan12Features vulkan12Features;
vulkan12Features.hostQueryReset = true;
vulkan12Features.descriptorIndexing = true;
vulkan12Features.runtimeDescriptorArray = true;
```

**Describe the Issue**

`[ VUID-vkCmdDraw-None-04008 ] Object 0: handle = 0x211187b9710, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xda6ce50000000156, name = MEVI_DRAW_TEXT, type = VK_OBJECT_TYPE_PIPELINE; | MessageID = 0x255241c3 | vkCmdDraw():  Vertex binding 1 is VK_NULL_HANDLE. (Most likely you forgot to call vkCmdBindVertexBuffers). The Vulkan spec states: If the nullDescriptor feature is not enabled, all vertex input bindings accessed via vertex input variables declared in the vertex shader entry point's interface must not be VK_NULL_HANDLE (https://vulkan.lunarg.com/doc/view/1.3.280.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdDraw-None-04008)`

[...]

PR #7901, opened 2024-04-24
layers: Fix drawtime vertex VU 04008

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7892

### DIFF
--- a/layers/core_checks/cc_pipeline_graphics.cpp
+++ b/layers/core_checks/cc_pipeline_graphics.cpp
@@ -3283,7 +3283,18 @@ bool CoreChecks::ValidatePipelineDrawtimeState(const LastBound &last_bound_state
 
     // Verify vertex binding
     if (pipeline && pipeline->vertex_input_state) {
+        vvl::unordered_set<uint32_t> bound_pipeline_bindings;
+        for (const auto &description : pipeline->vertex_input_state->binding_descriptions) {
+            bound_pipeline_bindings.insert(description.binding);
+        }
+
         for (const auto &vertex_buffer_binding : cb_state.current_vertex_buffer_binding_info) {
+            // Only validate the bindings from the last bound pipeline (unlesss it used VK_DYNAMIC_STATE_VERTEX_INPUT_EXT)
+            if (!pipeline->IsDynamic(VK_DYNAMIC_STATE_VERTEX_INPUT_EXT) &&
+                bound_pipeline_bindings.find(vertex_buffer_binding.first) == bound_pipeline_bindings.end()) {
+                continue;
+            }
+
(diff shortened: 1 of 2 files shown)


tests/unit/vertex_input_positive.cpp
```diff
@@ -814,3 +814,45 @@ TEST_F(PositiveVertexInput, BindVertexBufferNull) {
...
+
+    m_commandBuffer->begin();
+    m_commandBuffer->BeginRenderPass(m_renderPassBeginInfo);
+    // We bind, but rebind with valid pipeline
+    vk::CmdBindPipeline(m_commandBuffer->handle(), VK_PIPELINE_BIND_POINT_GRAPHICS, pipe0.Handle());  // invalid
+    vk::CmdBindPipeline(m_commandBuffer->handle(), VK_PIPELINE_BIND_POINT_GRAPHICS, pipe1.Handle());  // valid
+    VkDeviceSize offset = 0;
+    vk::CmdBindVertexBuffers(m_commandBuffer->handle(), 0, 1, &vtx_buf.handle(), &offset);
+    vk::CmdDraw(m_commandBuffer->handle(), 1, 0, 0, 0);
+    m_errorMonitor->VerifyFound();
```
artem-lunarg, line 854: `VerifyFound` leftover

layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -3283,7 +3283,18 @@ bool CoreChecks::ValidatePipelineDrawtimeState(const LastBound &last_bound_state
 
     // Verify vertex binding
     if (pipeline && pipeline->vertex_input_state) {
+        vvl::unordered_set<uint32_t> bound_pipeline_bindings;
```
artem-lunarg, line 3286: nit: better vector instead of map, less allocations (can be preallocated), the search is faster for this case though not important here.

artem-lunarg (approved):

layers/core_checks/cc_pipeline_graphics.cpp
artem-lunarg: but if the elements can be duplicated, then ignore!

spencer-lunarg: was just about to reply, it can be duplicated

artem-lunarg: yeah, make sense... it just this code runs often, and in practice number of bindings small, should be possible go without allocations at all by using small_vector with buffer size say 10 and some logic for deduplication (even optimized general allocators it's 100s of cycles). Maybe we can do something like that when we have proper profiling/optimization task.

PR #7918, opened 2024-04-29
icd: Create internal VVL Test Driver

spencer-lunarg:
Replaces the need for the MockICD in Vulkan-Tools and lets us do it internally using the new `VVL Test Driver`

This removes a Vulkan-Header-based-dependency

The entire implementation is done in `test_icd.cpp` and `test_icd_helper.h` with the goal that we never have to write MockICD logic in a python string anymore


artem-lunarg (approved):

tests/icd/test_icd.h
```diff
@@ -0,0 +1,79 @@
...
+#include "vulkan/vulkan.h"
+#include "vulkan/vk_icd.h"
+
+namespace icd {
+
+using mutex_t = std::mutex;
+using lock_guard_t = std::lock_guard<mutex_t>;
+using unique_lock_t = std::unique_lock<mutex_t>;
+
+static mutex_t global_lock;
```
jeremyg-lunarg, line 39: I realize that this is a straight port of mock icd, but statics in header files and driver state split among eleventy globals rather than a state structure is kinda bleah.

jeremyg-lunarg (approved):
LGTM, I want to steal it. I think mac just needs a warning surpressed...

tests/icd/test_icd.h
artem-lunarg: yes, not sure about final structure, but variables in the header in existing code it's not very conventional, moving to cpp might help to sleep better.

PR #7919, opened 2024-04-29
tests: Do not check result of submit helpers and rename

artem-lunarg:
Part 2.
This removes the checks for submit results and because I already touched a lot of these decided to rename to follow conventions.

Part 3 is going to add helpers to submit with semaphores.


spencer-lunarg (approved):
Awesome!

arno-lunarg (approved):

Issue #7797, opened 2024-04-01
False positive `UNASSIGNED-Device address out of bounds`


corporateshark:
**Environment:**
 - OS: Windows 10
 - GPU and driver version: GeForce RTX 3070 Ti; NVIDIA 551.86
 - SDK or header version if building from repo: VulkanSDK 1.3.280
 - Options enabled (synchronization, best practices, etc.): GPU-assisted validation enabled

**Describe the Issue**

Switching to VulkanSDK 1.3.280 resulted in `UNASSIGNED-Device address out of bounds` across multiple apps.

```
ERROR:
Validation layer:
 Validation Error: UNASSIGNED-Device address out of bounds
 Object 0: handle = 0000017AE1DEB630, type = VK_OBJECT_TYPE_QUEUE;
 MessageID = 000000001A898625
 vkCmdDrawIndexed():  Device address 0x13630088 access out of bounds.
Command buffer (0x17ae22564d0).
Draw Index 0.
Pipeline (Pipeline: mesh)(0xf8d1ba000000013e).
Shader Module (Shader Module: main (frag))(0xedf7820000000136).
Shader Instruction Index = 167.
Stage = Fragment. 
Fragment coord (x,y) = (252.5, 892.5).  Shader validation error occurred at line 55.
Unable to find SPIR-V OpSource.
```

The same error is encountered in more trivial apps where push constants contain only a single 64-bit buffer device address.

**Reproducible example:**

* Run [`Tiny_Mesh`](https://github.com/corporateshark/lightweightvk/blob/master/samples/Tiny_Mesh.cpp).

**Expected behavior**

No validation error.

spencer-lunarg:
@corporateshark I'm on PTO this week, but @arno-lunarg can you please take a look at this, you have a NVIDIA Windows laptop to test on and also been touching all of GPU-AV lately

spencer-lunarg:
update on this, Arno and I spent a good amount of time deep diving into this and realizing we actually have quite the large regression with BDA, we are currently writing new tests, making sure they are actually running in CI, and we will have this (and probably other issues) fixed by end of week hopefully

spencer-lunarg:
this auto closed from the PR, but that PR altered from the original goal (as we found other issues and wrote many more tests)

Keeping this open until we go back and confirm the `Tiny_Mesh` example is good now

spencer-lunarg:
[...]

PR #7924, opened 2024-04-30
gpu: Fix BDA buffer management

arno-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7797

Before the BDA buffer could be accessed by an in flight command buffer
while also being updated on the CPU.
Made the BDA buffer a per command buffer object to avoid races.

Tried to create a new test that does what I observed in LVK, but could not repro the issue. Still, I propose to add this test. No more issue observed locally on tiny mesh and tiny mesh large. (Had one in mesh large)


layers/vulkan/generated/gpu_pre_dispatch_comp.cpp
```diff
@@ -74,7 +74,7 @@
     0x00000021, 0x00000002, 0x0004002b, 0x0000000a, 0x00000024, 0x00000001, 0x0004002b, 0x0000000a, 0x00000025, 0x00000000,
     0x0004002b, 0x0000000a, 0x00000028, 0x00000006, 0x0003001d, 0x00000031, 0x0000000a, 0x0005001e, 0x00000032, 0x0000000a,
     0x0000000a, 0x00000031, 0x00040020, 0x00000033, 0x00000002, 0x00000032, 0x0004003b, 0x00000033, 0x00000034, 0x00000002,
-    0x0004002b, 0x00000018, 0x00000035, 0x00000001, 0x0004002b, 0x0000000a, 0x00000037, 0x00000010, 0x0004002b, 0x00000018,
```
spencer-lunarg, line 77: why is this being updated? Is this by change left over from when you were debugging values in the shader?

layers/gpu_validation/gpu_subclasses.cpp
```diff
@@ -183,6 +183,28 @@ void gpuav::CommandBuffer::AllocateResources() {
         }
     }
 
+    // BDA snapshot
+    {
```
spencer-lunarg, line 187:
```suggestion
```

not sure why have extra scope here

spencer-lunarg (approved):

Issue #7327, opened 2024-01-19
Possible wrong conditions for vkCmdSetRasterizationSamplesEXT/VK_EXT_extended_dynamic_state3


pawel-jastrzebski-mobica:
**Environment:**
 - OS: WIndows 10 22H2
 - GPU and driver version: nVidia RTX 3080Ti 536.67
 - Vulkan SDK 1.3.268.0 & VVL 8095b05
 - Options enabled (synchronization, best practices, etc.): N/A

**Describe the Issue**

During work on development of a sample (https://github.com/KhronosGroup/Vulkan-Samples/pull/856) to test the multi-sample (`vkCmdSetRasterizationSamplesEXT`) feature of `VK_EXT_extended_dynamic_state3` extension I discovered that there are several validation errors which I am not sure how to fix. 

Normally when multi-sampling is implemented statically (sample implementation: https://github.com/SaschaWillems/Vulkan/blob/master/examples/multisampling/multisampling.cpp): 
1) pipeline is provided with `VkPipelineMultisampleStateCreateInfo.rasterizationSamples` different form `VK_SAMPLE_COUNT_1_BIT`. 
2) render pass and frame buffer have 2 multi-sampled attachments (depth/stencil and color) with `image_create_info.samples` different form `VK_SAMPLE_COUNT_1_BIT` and a resolve attachment with `image_create_info.sample` equal to `VK_SAMPLE_COUNT_1_BIT`.

In the sample I am implementing multi-sampling using a pipeline with the dynamic state features:
1) pipeline is provided with `VkPipelineMultisampleStateCreateInfo.rasterizationSamples` equal to `VK_SAMPLE_COUNT_1_BIT` and `VK_DYNAMIC_STATE_RASTERIZATION_SAMPLES_EXT` in the dynamic state list. The pipeline is not modified in the runtime.
2) render pass and frame buffer have 2 multi-sampled attachments (depth/stencil and color) with `image_create_info.samples` can be whatever between `VK_SAMPLE_COUNT_1_BIT` and 
`VK_SAMPLE_COUNT_8_BIT` (depends on the GPU) and a resolve attachment with `image_create_info.sample` equal to `VK_SAMPLE_COUNT_1_BIT`. The are both modified on the fly when the user changes the settings, 
As stated earlier, I am not changing the pipeline.

I am seeing the following Vulkan Validation Errors in the dynamic multi-sampling scenario:
```
rasterizationSamples-07474
[...]

PR #7935, opened 2024-05-02
Check attachment depending on type

spencer-lunarg:
fixes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7327 (and the CTS tests in `dEQP-VK.pipeline.monolithic.extended_dynamic_state.misc.`)

We never had a way to know which attachments where depth vs stencil vs resolve vs color when it came to dynamic rendering

with some cleanup of the code, this logic is now all contained

### DIFF
--- a/layers/core_checks/cc_cmd_buffer_dynamic.cpp
+++ b/layers/core_checks/cc_cmd_buffer_dynamic.cpp
@@ -242,19 +242,17 @@ bool CoreChecks::ValidateDrawDynamicState(const LastBound& last_bound_state, con
     const vvl::CommandBuffer& cb_state = last_bound_state.cb_state;
     const vvl::DrawDispatchVuid& vuid = vvl::GetDrawDispatchVuid(loc.function);
     if (!pipeline_state || pipeline_state->IsDynamic(VK_DYNAMIC_STATE_COLOR_WRITE_MASK_EXT)) {
-        if (cb_state.active_attachments) {
-            for (uint32_t i = 0; i < cb_state.active_attachments->size(); ++i) {
-                const auto attachment = (*cb_state.active_attachments)[i];
-                if (attachment && attachment->create_info.format == VK_FORMAT_E5B9G9R9_UFLOAT_PACK32) {
-                    const auto color_write_mask = cb_state.dynamic_state_value.color_write_masks[i];
-                    VkColorComponentFlags rgb = VK_COLOR_COMPONENT_R_BIT | VK_COLOR_COMPONENT_G_BIT | VK_COLOR_COMPONENT_B_BIT;
-                    if ((color_write_mask & rgb) != rgb && (color_write_mask & rgb) != 0) {
-                        skip |= LogError(vuid.color_write_mask_09116, cb_state.Handle(), loc,
-                                         "Render pass attachment %" PRIu32
-                                         " has format VK_FORMAT_E5B9G9R9_UFLOAT_PACK32, but the corresponding element of "
-                                         "pColorWriteMasks is %s.",
-                                         i, string_VkColorComponentFlags(color_write_mask).c_str());
-                    }
+        for (uint32_t i = 0; i < cb_state.active_attachments.size(); ++i) {
+            const auto* attachment = cb_state.active_attachments[i].image_view;
+            if (attachment && attachment->create_info.format == VK_FORMAT_E5B9G9R9_UFLOAT_PACK32) {
+                const auto color_write_mask = cb_state.dynamic_state_value.color_write_masks[i];
+                VkColorComponentFlags rgb = VK_COLOR_COMPONENT_R_BIT | VK_COLOR_COMPONENT_G_BIT | VK_COLOR_COMPONENT_B_BIT;
+                if ((color_write_mask & rgb) != rgb && (color_write_mask & rgb) != 0) {
+                    skip |= LogError(vuid.color_write_mask_09116, cb_state.Handle(), loc,
+                                     "Render pass attachment %" PRIu32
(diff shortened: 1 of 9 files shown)


layers/state_tracker/cmd_buffer_state.cpp
```diff
@@ -39,6 +39,56 @@ static ShaderObjectStage inline ConvertToShaderObjectStage(VkShaderStageFlagBits
     return ShaderObjectStage::LAST;
 }
 
+// For Traditional RenderPasses, the index is simply the index into the VkRenderPassCreateInfo::pAttachments,
+// but for dynamic rendering, there is no "standard" way to map the index, instead we have our own custom indexing and it is not
+// obvious at all to the user where it came from
+std::string AttachmentInfo::Describe(AttachmentSource source, uint32_t index) const {
```
spencer-lunarg, line 45: Before the error message said

> vkCmdDraw(): Render pass attachment 1 samples VK_SAMPLE_COUNT_1_BIT does not match samples VK_SAMPLE_COUNT_4_BIT set with vkCmdSetRasterizationSamplesEXT()

but the `attachment 1` here means nothing with dynamic rendering, so now we have a better error message (plan to roll out all over after)

> vkCmdDraw(): VkRenderingInfo::pColorAttachments[0].imageView attachment samples VK_SAMPLE_COUNT_1_BIT does not match samples VK_SAMPLE_COUNT_4_BIT set with vkCmdSetRasterizationSamplesEXT()
>
> vkCmdDraw(): VkRenderPassCreateInfo::pAttachments[0] (Color) attachment samples VK_SAMPLE_COUNT_1_BIT does not match samples VK_SAMPLE_COUNT_4_BIT set with vkCmdSetRasterizationSamplesEXT()

layers/state_tracker/cmd_buffer_state.cpp
```diff
@@ -456,10 +507,14 @@ void CommandBuffer::ResetQueryPool(VkQueryPool queryPool, uint32_t firstQuery, u
     });
 }
 
-void CommandBuffer::UpdateSubpassAttachments(const vku::safe_VkSubpassDescription2 &subpass) {
+void CommandBuffer::UpdateSubpassAttachments() {
+    const auto &subpass = activeRenderPass->create_info.pSubpasses[GetActiveSubpass()];
+    assert(active_subpasses.size() == active_attachments.size());
```
spencer-lunarg, line 512: This kind of sums things up

`active_attachments` is a vector of `AttachmentInfo` which is used everything
`active_subpasses` is a vector of `SubpassInfo` which is when we are not using dynamic rendering

layers/state_tracker/cmd_buffer_state.h
```diff
@@ -360,9 +392,8 @@ class CommandBuffer : public RefcountedStateObject {
     // The RenderPass created from vkCmdBeginRenderPass or vkCmdBeginRendering
     std::shared_ptr<vvl::RenderPass> activeRenderPass;
     // Used for both type of renderPass
-    // TOOD - Tried to get rid of this shared_ptr as it didn't seem to be needed, but would hit "Assertion failed: vector subscript
```
artem-lunarg, line 363: 🎉

layers/state_tracker/cmd_buffer_state.cpp
artem-lunarg: Really cool improvement of the error message.

layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -847,16 +846,17 @@ bool CoreChecks::ValidateDrawDynamicStatePipeline(const LastBound& last_bound_st
...
-                    }
+            for (uint32_t i = 0; i < cb_state.active_attachments.size(); ++i) {
+                const AttachmentInfo& attachment_info = cb_state.active_attachments[i];
+                const auto* attachment = attachment_info.image_view;
+                if (attachment && !attachment_info.IsInput() && !attachment_info.IsResolve() &&
+                    cb_state.dynamic_state_value.rasterization_samples != attachment->samples) {
+                    skip |=
+                        LogError(vuid.rasterization_sampled_07474, cb_state.Handle(), loc,
+                                 "%s attachment samples %s does not match samples %s set with vkCmdSetRasterizationSamplesEXT().",
+                                 attachment_info.Describe(cb_state.attachment_source, i).c_str(),
```
artem-lunarg, line 857: Trying to understand indexing, do we pass raw index to `Describe` because it's a color attachment? Would it be correct to use here `GetDynamicColorAttachmentImageIndex` to be more expressive?

spencer-lunarg: This could be the depth/stencil as well (for this VU... this might change when I apply everywhere)

- if Renderpass, it will be the right index still, but we print out the `(Depth)` part
- if Dynamic Rendering, the depth/stencil don't use the `index` so its just ignored

artem-lunarg: One more question. It has to be that this place guarantees that we can't have `Type::ColorResolve` attachment because otherwise we need to use `GetDynamicColorResolveAttachmentImageIndex()` to get proper index, is it correct?

spencer-lunarg: yes, but we do a `!IsResolve()` check above so can make the guarantee here

(the core issue this is solving is we never had a way to see if these attachments are resolve or non-resolve and this VU was giving false positive that the Resolve attachment was `VK_SAMPLE_COUNT_1_BIT`)

... The usage of `Describe` is a little flimsy for now, but when I start applying it everywhere, I might discover how it needs to be restructured

artem-lunarg (approved):

PR #7942, opened 2024-05-02
tests: Remove operator bool

artem-lunarg:
It's unsafe. Can hide errors when an argument is of a wrong type but conversion operator finds a way.


spencer-lunarg (approved):
honestly never knew this was here

PR #7943, opened 2024-05-02
Feature request: Add ApplicationName information to the log messages

kocdemir:
Hi, 
Adding this draft CL as a feature request to check if it'd be possible to add application names from InstanceInfo to the log messages. This would be very helpful for debugging issues where multiple Vulkan instances are being created from different applications, such as in emulators. 
To avoid clutter and changing the default behavior, names can be printed out only when there are multiple instances with VVL in flight, or the feature can be put behind a global flag. 
Thanks

Example output with [AppName: ..]:  
`
UNASSIGNED-CoreValidation-DrawState-InvalidImageLayout(ERROR / SPEC): msgNum: 1303270965 - [AppName: com.android.systemui] Validation Error: [ UNASSIGNED-CoreValidation-DrawState-InvalidImageLayout ] Object 0: handle = 0x562767ac4740, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xc88e49000000081d, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x4dae5635 | vkQueueSubmit(): 
...
`

### DIFF
--- a/layers/error_message/logging.cpp
+++ b/layers/error_message/logging.cpp
@@ -106,6 +106,9 @@ bool DebugReport::UpdateLogMsgCounts(int32_t vuid_hash) const {
     }
 }
 
+// Global instance name, set to the last created instance's name
+std::string instance_application_name = "";
+
 bool DebugReport::DebugLogMsg(VkFlags msg_flags, const LogObjectList &objects, const char *message, const char *text_vuid) const {
     bool bail = false;
     std::vector<VkDebugUtilsLabelEXT> queue_labels;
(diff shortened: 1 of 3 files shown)


spencer-lunarg:
@kocdemir thanks for this, I agree this is a great simple thing to add, I quick threw something together here https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7946

I have a overdue todo to update how we do formatting of the error messages and this is a simple good start for me to start working on it. For now I have this behind a global setting flag as this will just be more noise to the current messages for most people

kocdemir:
Thank you, #7946 looks good

Issue #7932, opened 2024-05-01
Descriptor Indexing not detecting when sets are shared


spencer-lunarg:
With a shader like

```glsl
        layout (set = 0, binding = 0) uniform texture2D kTextures2D[];
        layout (set = 0, binding = 1) uniform sampler kSamplers[];
        layout (set = 1, binding = 0) uniform textureCube kTexturesCube[];

        vec4 textureBindlessCube(uint textureid, uint samplerid) {
            return texture(samplerCube(kTexturesCube[textureid], kSamplers[samplerid]), vec3(0.0));
        }
        vec4 textureBindless(uint textureid, uint samplerid) {
            return texture(sampler2D(kTextures2D[textureid], kSamplers[samplerid]), vec2(0.0));
        }
```

if set `0` and `1` are pointed to the same descriptor set and the descriptor array looks like

`[VK_IMAGE_VIEW_TYPE_2D, ...., VK_IMAGE_VIEW_TYPE_CUBE_ARRAY]`

We are giving a false positive. I have a test about to push that can reproduce it because we don't have a good way currently to distinguish what `set` it was when doing post processing

cc @corporateshark this is currently being triggered in `Tiny_MeshLarge` (goal is to have it fixed by next SDK, but thanks for having an easy demo to debug and reproduce this)

corporateshark:
@spencer-lunarg I think the shader is missing `nonuniformEXT` qualifiers.

PR #7949, opened 2024-05-03
gpuav: Track which set is used during descriptor indexing

spencer-lunarg:
potential fix for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7932


layers/gpu_shaders/inst_bindless_descriptor.comp
```diff
@@ -154,7 +154,9 @@ bool inst_bindless_descriptor(const uint inst_num, const uvec4 stage_info, const
         if ((out_vec.x == 0u) && (out_vec.y == 0u)) {
             break;
         }
-        out_data.data[state_index] = 1;
+        // desc_sets[0] and desc_sets[1] could be the same pointer if shared VkDescriptorSet handles
+        // this buffer is all zero and any non-zero value means we are accessing this index
```
jeremyg-lunarg, line 158: what if instead you wrote` out_data.data[state_index] =  (1u << 31) | desc_set`  I think that would be clearer than using ~0 for 0. Also there are many spare upper bits you could eventually use for something else as well.

arno-lunarg (approved):

Issue #7923, opened 2024-04-30
Regression: incorrect warnings claiming shaders aren't writing to attachments beyond the first one.


mbechard:
**Environment:**
 - OS: Windows 10
 - GPU and driver version: Quadro 6000 Ada, 522.22
 - SDK or header version if building from repo: 1.3.280.0
 - Options enabled (synchronization, best practices, etc.):

**Describe the Issue**

Starting in SDK 1.3.275.0 (and not occurring in 1.3.268.0) I now get incorrect validation warnings for attachments that the validation thinks aren't getting written by shaders.

```
WARNING : VALIDATION - Message Id Number: 1203141749 | Message Id Name: Undefined-Value-ShaderInputNotProduced
        Validation Warning: [ Undefined-Value-ShaderInputNotProduced ] Object 0: handle = 0xf39fcf00000020fb, type = VK_OBJECT_TYPE_SHADER_MODULE; | MessageID = 0x47b67c75 | vkCreateGraphicsPipelines(): pCreateInfos[0] Attachment 1 not written by fragment shader; undefined values will be written to attachment
```

The below simple shader will produce the warning
```
                              SourceExtension  "GL_EXT_nonuniform_qualifier"
                              SourceExtension  "GL_EXT_scalar_block_layout"
                              SourceExtension  "GL_GOOGLE_cpp_style_line_directive"
                              SourceExtension  "GL_GOOGLE_include_directive"
                              Name 10  "main"
                              Name 29  "fragColor"
                              ModuleProcessed  "auto-map-bindings"
                              ModuleProcessed  "auto-map-locations"
                              ModuleProcessed  "client vulkan100"
                              ModuleProcessed  "target-env spirv1.3"
                              ModuleProcessed  "target-env vulkan1.1"
                              ModuleProcessed  "entry-point main"
                              ModuleProcessed  "client vulkan100"
                              ModuleProcessed  "target-env spirv1.3"
                              ModuleProcessed  "target-env vulkan1.1"
                              ModuleProcessed  "entry-point main"
                              Decorate 29(fragColor) Location 0
               8:             TypeVoid
               9:             TypeFunction 8
              12:             TypeFloat 32
              13:             TypeVector 12(float) 4
              23:   12(float) Constant 1065353216
              24:   13(fvec4) ConstantComposite 23 23 23 23
              25:             TypeInt 32 0
              26:     25(int) Constant 2
              27:             TypeArray 13(fvec4) 26
[...]

PR #7954, opened 2024-05-06
Prevent false positive when using fragment output array

spencer-lunarg:
Temp false positive patch for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7923 

I plan to go back in and just redo all the logic as I realize there is a lot of gaps with how we are checking Fragment output                                                                                                                                                                                                                                                                    matching. But want to at least remove false positive for the new SDK


artem-lunarg (approved):
LGTM

PR #7955, opened 2024-05-06
gpu: Renamings in descriptor set validation

arno-lunarg:
And another commit where I enforce that the buffer device address extension is always added to GPU AV tests


layers/drawdispatch/descriptor_validator.h
```diff
@@ -35,7 +35,10 @@ class CommandBuffer;
 class Sampler;
 class DescriptorSet;
 
-using DescriptorBindingInfo = std::pair<uint32_t, std::vector<DescriptorRequirement>>;
```
spencer-lunarg, line 38: we should run this through the perf suit

I tried to do

```c++
-    using DescriptorBindingInfo = std::pair<const uint32_t, DescriptorRequirement>;
+    using DescriptorBindingInfo = std::pair<const uint32_t, const ResourceInterfaceVariable*>;
```
 
 https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/898721a7a9e98689e0b8ec75ebc77896a295cdf4#diff-5fe986fb6847842fb6c34866bd2fc5d71c994c71cac11440dc78a88a14dd39e2L765

 
 and spent a lot of time trying to understand why it was giving us a 20% perf hit in Doom
 
 This is very perf critical code (as it is checking against millions of descriptors weather to re-validate them or not so want to take extra precaution changing this from a `std::pair` to `struct`

layers/gpu_validation/gpu_descriptor_set.cpp
```diff
@@ -435,48 +426,48 @@ std::shared_ptr<gpuav::DescriptorSet::State> gpuav::DescriptorSet::GetOutputStat
...
-    result = vmaInvalidateAllocation(allocator, allocation, 0, VK_WHOLE_SIZE);
+    uint32_t *shader_set_numbers = nullptr;
+    result = vmaMapMemory(allocator, gpu_records_allocation, reinterpret_cast<void **>(&shader_set_numbers));
+    result = vmaInvalidateAllocation(allocator, gpu_records_allocation, 0, VK_WHOLE_SIZE);
 
     uint32_t max_binding = layout_data[0].count;
     for (uint32_t binding = 0; binding < max_binding; binding++) {
-        uint32_t count = layout_data[binding + 1].count;
-        uint32_t start = layout_data[binding + 1].state_start;
+        const uint32_t count = layout_data[binding + 1].count;
```
arno-lunarg, line 454: @jeremy-lunarg I failed to find better names for `count`, `start`, `pos`, `i` as I do not fully grasp what they represent. Do you have any suggestions please?

layers/drawdispatch/descriptor_validator.h
spencer-lunarg: confirmed that Perf tests with ok with this

spencer-lunarg (approved):

tests/unit/gpu_av_positive.cpp
```diff
@@ -43,6 +43,9 @@ VkValidationFeaturesEXT GpuAVTest::GetGpuAvValidationFeatures() {
 // This checks any requirements needed for GPU-AV are met otherwise devices not meeting them will "fail" the tests
 void GpuAVTest::InitGpuAvFramework(void *p_next) {
     SetTargetApiVersion(VK_API_VERSION_1_1);
+    // Buffer device address is used by enough tests to just be considered required
+    AddRequiredExtensions(VK_KHR_BUFFER_DEVICE_ADDRESS_EXTENSION_NAME);
```
spencer-lunarg, line 47: I thought we figured out this is something that GPU-AV should require

This is not related to renaming and is changing how every GPU-AV test is ran

Issue #7957, opened 2024-05-06
Compilation fails due to missing kInstBufferOOBEnable


aabdolrashidi:
Hello,
In a recent update to SPIRV-Tools (https://github.com/KhronosGroup/SPIRV-Tools/commit/9241a58a8028c49510bc174b6c970e3c2b4b8e51), the parameter `kInstBufferOOBEnable` has been removed from `spvtools`. However, this value is still being used in `Vulkan-ValidationLayers` (https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/layers/gpu_validation/gpu_validation.cpp#L619), which leads to compilation errors like the following:

```
../../third_party/vulkan-deps/vulkan-validation-layers/src/layers/gpu_validation/gpu_validation.cpp:621:78: error: no member named 'kInstBufferOOBEnable' in namespace 'spvtools'
```

@jeremyg-lunarg

aabdolrashidi:
Thank you!

PR #7960, opened 2024-05-07
gpuav: Use internal stream out constants

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7957

These use to live in spirv-tools `instrument.hpp` but we can now safely move to VVL


arno-lunarg (approved):
Nice, I wanted to get rid of those `spvtools` constants!

PR #7961, opened 2024-05-07
tests: Remove deprecated m_commandPool

spencer-lunarg:
followup to @artem-lunarg https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7958/

This removes `m_commandPool` and finally makes `vkt::CommandBuffer` take a reference


arno-lunarg (approved):
noice

PR #7967, opened 2024-05-07
layers: Cleanup old subpass UNASSIGNED VUs

spencer-lunarg:
Takes care of 3 `UNASSIGNED-CoreValidation-DrawState-InvalidRenderpass`

### DIFF
--- a/layers/core_checks/cc_render_pass.cpp
+++ b/layers/core_checks/cc_render_pass.cpp
@@ -1229,11 +1229,10 @@ static bool FindDependency(const uint32_t index, const uint32_t dependent, const
     return false;
 }
 
-bool CoreChecks::CheckDependencyExists(const VkRenderPass renderpass, const uint32_t subpass, const VkImageLayout layout,
-                                       const std::vector<SubpassLayout> &dependent_subpasses,
-                                       const std::vector<DAGNode> &subpass_to_node, const Location &attachment_loc,
-                                       bool &skip) const {
-    bool result = true;
+bool CoreChecks::ValidateDependencyExists(const VkRenderPass renderpass, const uint32_t subpass, const VkImageLayout layout,
+                                          const std::vector<SubpassLayout> &dependent_subpasses,
+                                          const std::vector<DAGNode> &subpass_to_node, const Location &attachment_loc) const {
+    bool skip = false;
     const bool b_image_layout_read_only = IsImageLayoutReadOnly(layout);
     // Loop through all subpasses that share the same attachment and make sure a dependency exists
     for (uint32_t k = 0; k < dependent_subpasses.size(); ++k) {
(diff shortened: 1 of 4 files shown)


layers/core_checks/cc_render_pass.cpp
```diff
@@ -1250,51 +1249,15 @@ bool CoreChecks::CheckDependencyExists(const VkRenderPass renderpass, const uint
...
-    if (result && depth > 0) {
-        bool has_preserved = false;
-        for (uint32_t j = 0; j < subpass.preserveAttachmentCount; ++j) {
-            if (subpass.pPreserveAttachments[j] == attachment) {
-                has_preserved = true;
-                break;
-            }
-        }
-        if (!has_preserved) {
-            skip |= LogError("UNASSIGNED-CoreValidation-DrawState-InvalidRenderpass", renderpass, attachment_loc,
```
spencer-lunarg, line 1293: so this logic does recursively calls into `CheckPreserved` and never returned `true`, so it is not even possible to hit now (and we have zero tests for it)

Looking more into this, this logic doesn't consider a lot of things, like `VK_ATTACHMENT_STORE_OP_NONE_EXT`/`VK_ATTACHMENT_LOAD_OP_NONE_EXT` which were added to replace Preserve Attachments

Also this is not invalid, if we want to try and correctly add this back, it would better suited for Best Practice layers

layers/core_checks/cc_render_pass.cpp
```diff
@@ -2300,15 +2261,9 @@ bool CoreChecks::ValidateDependencies(const vvl::Framebuffer &framebuffer_state,
             for (auto overlapping_attachment : attachments[attachment].overlapping) {
                 attachments[overlapping_attachment].outputs.emplace_back(sp);
             }
-
-            if (attachment_indices.count(attachment)) {
```
spencer-lunarg, line 2304: I tried, and there are so many VUs stopping you from this.

This itself is not invalid, but to find a format that is both capable of supporting everything needed for `color` and `depth` attachment is just not feasible

I imagine this was added a long time ago as an overall catching of things, but the spec VUs and VVL have since been cleaned up and catch things like this early

artem-lunarg:
Is this going to report an error when subpasses write to different attachments?  Each attachment is used only by the subpass that writes to it and not by other subpasses. No dependencies are defined.

spencer-lunarg:
> Is this going to report an error when subpasses write to different attachments

So I guess not sure if we check that already, but this is not removing any logic that would have checked that

artem-lunarg (approved):
LGTM as update of existing logic

spencer-lunarg:
> My current thinking that the above case is fine (render to two different attachment that do not affect each other), so don't expect an error

Yes, all I did for that was

1. Write a test that could hit it
2. Change the VUID string to be unique and more descriptive

The internal logic of that UNASSIGNED VU is something I would like to have listed, as you said, I am also not sure what is valid or not for subpass dependency

Issue #7975, opened 2024-05-08
Code that does `auto &state_object = *Get<>(handle);` risks use after free bugs.


jeremyg-lunarg:
Validation code must hold a reference to any state objects it is using.  This pattern avoids that and risks use after free bugs from invalid applications.  Current list of places this is happening. There could be others that aren't easily findable with grep.
```
:~/src/Vulkan-ValidationLayers>git grep "\*Get<"
layers/best_practices/bp_cmd_buffer.cpp:    const auto& query_pool_state = *Get<vvl::QueryPool>(queryPool);
layers/core_checks/cc_descriptor.cpp:    const auto &src_set = *Get<vvl::DescriptorSet>(update.srcSet);
layers/core_checks/cc_descriptor.cpp:    const auto &dst_set = *Get<vvl::DescriptorSet>(update.dstSet);
layers/core_checks/cc_descriptor.cpp:        const auto &set_node = *Get<vvl::DescriptorSet>(dst_set);
layers/core_checks/cc_device_memory.cpp:            !HasExternalMemoryImportSupport(*Get<vvl::Image>(dedicated_image), VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD_BIT)) {
layers/core_checks/cc_device_memory.cpp:            !HasExternalMemoryImportSupport(*Get<vvl::Buffer>(dedicated_buffer), VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD_BIT)) {
layers/core_checks/cc_query.cpp:    const auto &query_pool_state = *Get<vvl::QueryPool>(queryPool);
layers/core_checks/cc_query.cpp:    const auto &query_pool_state = *Get<vvl::QueryPool>(queryPool);
layers/core_checks/cc_query.cpp:    const auto &query_pool_state = *Get<vvl::QueryPool>(queryPool);
layers/core_checks/cc_query.cpp:    const auto &query_pool_state = *Get<vvl::QueryPool>(queryPool);
layers/core_checks/cc_query.cpp:    const auto &query_pool_state = *Get<vvl::QueryPool>(queryPool);
layers/core_checks/cc_query.cpp:    const auto &query_pool_state = *Get<vvl::QueryPool>(queryPool);
layers/core_checks/cc_query.cpp:    const auto &query_pool_state = *Get<vvl::QueryPool>(query_obj.pool);
[...]

PR #7980, opened 2024-05-09
tests: Add extra object lifetime tests

spencer-lunarg:
as a reaction from https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7975 I wanted to make sure we had good test coverage and realize we do test these object lifetime cases pretty well already

### DIFF
--- a/tests/unit/object_lifetime.cpp
+++ b/tests/unit/object_lifetime.cpp
@@ -1232,3 +1232,24 @@ TEST_F(NegativeObjectLifetime, FreeDescriptorSetsNull) {
     vk::FreeDescriptorSets(device(), ds_pool.handle(), 1, &invalid_set);
     m_errorMonitor->VerifyFound();
 }
+
+TEST_F(NegativeObjectLifetime, DescriptorBufferInfoUpdate) {
+    TEST_DESCRIPTION("Destroy a buffer then try to update it in the descriptor set");
+    RETURN_IF_SKIP(Init());
+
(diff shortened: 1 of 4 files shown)


tests/unit/object_lifetime.cpp
```diff
@@ -1232,3 +1232,24 @@ TEST_F(NegativeObjectLifetime, FreeDescriptorSetsNull) {
...
+    RETURN_IF_SKIP(Init());
+
+    vkt::Buffer buffer(*m_device, 32, VK_BUFFER_USAGE_UNIFORM_BUFFER_BIT);
+    VkBuffer invalid_buffer = CastToHandle<VkBuffer, uintptr_t>(0xbaadbeef);
+
+    OneOffDescriptorSet descriptor_set(m_device, {
+                                                     {0, VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER, 1, VK_SHADER_STAGE_ALL, nullptr},
+                                                     {1, VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER, 1, VK_SHADER_STAGE_ALL, nullptr},
+                                                 });
+    descriptor_set.WriteDescriptorBufferInfo(0, buffer.handle(), 0, VK_WHOLE_SIZE, VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER);
```
artem-lunarg, line 1247: awww, was confused initially that we use DescriptorBuffer naming here, probably was added before *descriptor buffers* were introduced. For future clean up maybe just to use WriteBufferInfo or other alternative.

artem-lunarg (approved):

Issue #7977, opened 2024-05-08
Adding debug marker object name/tag to swapchain image causes validation error


SRSaunders:
**Environment:**
 - OS: All (tested on macOS Ventura)
 - GPU and driver version: All
 - SDK or header version if building from repo: SDK 1.3.280
 - Options enabled (synchronization, best practices, etc.): standard

**Describe the Issue**

Adding a debug marker object name or tag to a swapchain image results in spurious validation layer messages 01492 or 01495.  This is because `PreCallValidateDebugMarkerSetObjectNameEXT()` and `PreCallValidateDebugMarkerSetObjectTagEXT()` only check the `object_map` and not the `swapchain_image_map` for object containment.

**Expected behavior**

Adding a debug marker object name or tag to a swapchain image should not result in a validation layer error message.

**Valid Usage ID**

If applicable, please include the validation messages encountered leading up to the issue:

`[Vulkan] ERROR location=0x4abd241d code=0, layerPrefix='Validation'] Validation Error: [ VUID-VkDebugMarkerObjectNameInfoEXT-object-01492 ] | MessageID = 0x4abd241d | vkDebugMarkerSetObjectNameEXT(): pNameInfo->objectType doesn't match the object. The Vulkan spec states: object must be a Vulkan object of the type associated with objectType as defined in VkDebugReportObjectTypeEXT and Vulkan Handle Relationship (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkDebugMarkerObjectNameInfoEXT-object-01492)
`

**Additional context**

Possible fix in **object_tracker_utils** (could possibly add check for `object_type == kVulkanObjectTypeImage` before checking `swapchain_image_map` but not sure if really needed):

```
bool ObjectLifetimes::PreCallValidateDebugMarkerSetObjectTagEXT(VkDevice device, const VkDebugMarkerObjectTagInfoEXT *pTagInfo, const ErrorObject &error_obj) const {
...
        } else if (!object_map[object_type].contains(pTagInfo->object) && !swapchain_image_map.contains(pTagInfo->object)) {
                                                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
            skip |= LogError("VUID-VkDebugMarkerObjectTagInfoEXT-object-01495", device,
[...]

PR #7981, opened 2024-05-09
layers: Fix Swapchain DebugMarker false positive

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7977


artem-lunarg (approved):

SRSaunders:
Shouldn't the code (for both `pTagInfo` and `pNameInfo`) be:

```
if (object_type != kVulkanObjectTypeImage || !swapchain_image_map.contains(pTagInfo->object)) {
                                          ^^
```

Which is the expansion of:

```
if !(object_type == kVulkanObjectTypeImage && swapchain_image_map.contains(pTagInfo->object)) {
```

spencer-lunarg:
@SRSaunders ... this is why I should have wrote a negative test checking this, got the fix (and a test) in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7984

SRSaunders:
Thanks very much.

PR #7986, opened 2024-05-10
gpuav: Support all stages for gpuav

spencer-lunarg:
There is no reason I know of that these should be excluding geometry/tess stages as GPU-AV should work for all stages

cc @arno-lunarg


arno-lunarg (approved):
But nobody loves the geometry shader!

spencer-lunarg:
> But nobody loves the geometry shader!

You clearly never had to benchmark the manhattan benchmark

Issue #5631, opened 2023-04-12
Add subpass self dependency VUs


spencer-lunarg:
[Internal spec change](https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5687/diffs)
[Released in  1.3.242](https://github.com/KhronosGroup/Vulkan-Docs/commit/f3e686fc22251102713a2a1a1cd071f60ab6efd8#diff-c0f9e36aa254a211f561ac0f4b0027c7ce5fdaa0da3fc0ffbf95ffa7bf5c802c)

- [x] VUID-vkCmdPipelineBarrier2-None-07890 (https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6649)
- [x] VUID-vkCmdPipelineBarrier2-dependencyFlags-07891
- [x] VUID-vkCmdPipelineBarrier2-None-07892
- [x] VUID-vkCmdPipelineBarrier-None-07889

related:

`VUID-vkCmdPipelineBarrier2-pDependencies-02285` is similar, but it missing the `VUID-vkCmdPipelineBarrier2-None-07889` variation

```
ifndef::VK_KHR_multiview,VK_VERSION_1_1[]
  * [[VUID-{refpage}-pDependencies-02285]]
endif::VK_KHR_multiview,VK_VERSION_1_1[]
ifdef::VK_KHR_multiview,VK_VERSION_1_1[]
  * [[VUID-{refpage}-None-07889]]
endif::VK_KHR_multiview,VK_VERSION_1_1[]
```

ziga-lunarg:
It seems to me that these VUIDs are impossible to hit without hitting similar VUIDs on render pass create, for example to hit 07890 you will hit `VUID-VkSubpassDependency-srcSubpass-06809` when you create the required render pass.

spencer-lunarg:
what if you use dynamic rendering and have no subpasses?

ziga-lunarg:
You can't call `vkCmdPipelineBarrier()` inside a dynamic render pass.

spencer-lunarg:
awww, ok, will take a double look Monday and make a spec MR to remove them

spencer-lunarg:
update, so I think I realize this original spec change was for the `VK_KHR_dynamic_rendering_local_read` that came out and were

PR #7991, opened 2024-05-10
tests: Add missing subpass pipeline barrier tests

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5631

Confirmed the VUs are actually done and wrote extra tests to confirm

### DIFF
--- a/tests/unit/dynamic_rendering_local_read.cpp
+++ b/tests/unit/dynamic_rendering_local_read.cpp
@@ -226,15 +226,8 @@ TEST_F(NegativeDynamicRenderingLocalRead, ImageBarrier) {
     AddRequiredFeature(vkt::Feature::dynamicRendering);
     AddRequiredFeature(vkt::Feature::shaderTileImageColorReadAccess);
     RETURN_IF_SKIP(Init());
-    InitRenderTarget();
-
+    InitDynamicRenderTarget();
     m_commandBuffer->begin();
-
-    VkRenderingInfoKHR begin_rendering_info = vku::InitStructHelper();
-    VkClearRect clear_rect = {{{0, 0}, {m_width, m_height}}, 0, 1};
-    begin_rendering_info.renderArea = clear_rect.rect;
-    begin_rendering_info.layerCount = 1;
-
     vkt::Image image(*m_device, 32, 32, 1, VK_FORMAT_R8G8B8A8_UNORM, VK_IMAGE_USAGE_STORAGE_BIT);
 
     VkImageMemoryBarrier img_barrier = vku::InitStructHelper();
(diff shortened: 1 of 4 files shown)


tests/unit/subpass.cpp
```diff
@@ -1261,4 +1261,47 @@ TEST_F(NegativeSubpass, FramebufferNoAttachmentsSampleCounts) {
     m_errorMonitor->SetDesiredError("VUID-VkGraphicsPipelineCreateInfo-subpass-00758");
     pipe.CreateGraphicsPipeline();
     m_errorMonitor->VerifyFound();
-}
\ No newline at end of file
+}
+
+// Need clarification about VUID-vkCmdPipelineBarrier2-None-07893
+TEST_F(NegativeSubpass, DISABLED_DependencyViewLocalSubpass) {
```
spencer-lunarg, line 1267: I will follow-up with this next week, might need slight tweaking of the VU

tests/unit/dynamic_rendering_local_read.cpp
```diff
@@ -1106,3 +1078,82 @@ TEST_F(NegativeDynamicRenderingLocalRead, AttachmentLocationsMax) {
...
+    memory_barrier_2.dstAccessMask = VK_ACCESS_2_COLOR_ATTACHMENT_READ_BIT;
+
+    VkDependencyInfoKHR dependency_info = vku::InitStructHelper();
+    dependency_info.dependencyFlags = VK_DEPENDENCY_VIEW_LOCAL_BIT;
+    dependency_info.memoryBarrierCount = 1;
+    dependency_info.pMemoryBarriers = &memory_barrier_2;
+    dependency_info.bufferMemoryBarrierCount = 0;
+    dependency_info.imageMemoryBarrierCount = 0;
+
+    m_errorMonitor->SetDesiredError("VUID-vkCmdPipelineBarrier2-dependencyFlags-07891");
```
artem-lunarg, line 1104: Is it something from the spec that we need to test this VU with local read?

artem-lunarg (approved):

tests/unit/dynamic_rendering_local_read.cpp
spencer-lunarg: yes, so you will hit `VUID-vkCmdPipelineBarrier2-None-09553` if you try to do a pipeline barrier in a renderpass

actually the `DependencyViewLocalOutsideRendering` **could** be in another file, but it made sense to keep these together

Williams23678: How can you make 10k in a day?

PR #7992, opened 2024-05-10
layers: Refactor RequiredQueueFlags check

spencer-lunarg:
Instead of passing ad-hoc strings down, this separates `ValidateCmdQueueFlags` into a  `HasRequiredQueueFlags` and `DescribeRequiredQueueFlag` logic

While it seems like more code, it prevents having to pass strings around while we are still validating and saves that until we need to print the error out


artem-lunarg (approved):
LGTM. That's really nice, I like this separation into small hot validation functions and larger cold code for reporting.

PR #7998, opened 2024-05-13
Use empty() over size() == 0

spencer-lunarg:
While trying to follow up for the disabled test in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7991 I found some spots where `empty()` should have been used

slightly overloaded, but threw in the test change as well


tests/unit/subpass.cpp
```diff
@@ -1262,46 +1262,3 @@ TEST_F(NegativeSubpass, FramebufferNoAttachmentsSampleCounts) {
     pipe.CreateGraphicsPipeline();
     m_errorMonitor->VerifyFound();
 }
-
-// Need clarification about VUID-vkCmdPipelineBarrier2-None-07893
```
spencer-lunarg, line 1266: The VU was correct and is already validated and was related to Multiview, not subpasses

artem-lunarg (approved):

PR #7999, opened 2024-05-13
tests: Fix SelectInstrumentedShaders tests

spencer-lunarg:
The `SelectInstrumentedShaders` have not been running properly and this fixes them

### DIFF
--- a/tests/unit/gpu_av.cpp
+++ b/tests/unit/gpu_av.cpp
@@ -90,9 +90,8 @@ TEST_F(NegativeGpuAV, ValidationFeatures) {
 TEST_F(NegativeGpuAV, SelectInstrumentedShaders) {
     TEST_DESCRIPTION("GPU validation: Validate selection of which shaders get instrumented for GPU-AV");
     SetTargetApiVersion(VK_API_VERSION_1_2);
-    AddRequiredExtensions(VK_EXT_LAYER_SETTINGS_EXTENSION_NAME);
     const VkBool32 value = true;
-    const VkLayerSettingEXT setting = {OBJECT_LAYER_NAME, "select_instrumented_shaders", VK_LAYER_SETTING_TYPE_BOOL32_EXT, 1,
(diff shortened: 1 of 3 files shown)


tests/unit/gpu_av.cpp
```diff
@@ -90,9 +90,8 @@ TEST_F(NegativeGpuAV, ValidationFeatures) {
 TEST_F(NegativeGpuAV, SelectInstrumentedShaders) {
     TEST_DESCRIPTION("GPU validation: Validate selection of which shaders get instrumented for GPU-AV");
     SetTargetApiVersion(VK_API_VERSION_1_2);
-    AddRequiredExtensions(VK_EXT_LAYER_SETTINGS_EXTENSION_NAME);
```
spencer-lunarg, line 93: The validation layers expose this extension so it will always be supported

tests/unit/gpu_av.cpp
```diff
@@ -90,9 +90,8 @@ TEST_F(NegativeGpuAV, ValidationFeatures) {
...
     const VkBool32 value = true;
-    const VkLayerSettingEXT setting = {OBJECT_LAYER_NAME, "select_instrumented_shaders", VK_LAYER_SETTING_TYPE_BOOL32_EXT, 1,
```
spencer-lunarg, line 95: were getting deprecation warning (we renamed this variable)

tests/unit/gpu_av.cpp
```diff
@@ -162,7 +161,7 @@ TEST_F(NegativeGpuAV, SelectInstrumentedShaders) {
     m_commandBuffer->end();
     // Should get a warning since shader was instrumented
     m_errorMonitor->ExpectSuccess(kWarningBit | kErrorBit);
-    m_errorMonitor->SetDesiredFailureMsg(kWarningBit, "VUID-vkCmdDraw-storageBuffers-06936");
+    m_errorMonitor->SetDesiredWarning("VUID-vkCmdDraw-storageBuffers-06936", 3);
```
spencer-lunarg, line 164: clearly this was never being ran for a while in CI (because `AddRequiredExtensions(VK_EXT_LAYER_SETTINGS_EXTENSION_NAME);` was blocking it) and now is doing things as expected

artem-lunarg (approved):

PR #8000, opened 2024-05-13
Cleanup for Shader Object tests

spencer-lunarg:
We will want to have more and more Shader Object tests for GPU-AV and to help with that, created some helpers to allow things other than `NegativeShaderObject`/`PositiveShaderObject` to use them


artem-lunarg (approved):
LGTM
image
👍

Issue #8001, opened 2024-05-13
Correctly propagate 'enabled_features' modifications


arno-lunarg:
Setting `enabled_features.bufferDeviceAddress` to true in `GpuShaderInstrumentor::PreCallRecordCreateDevice`
when adding missing features will modify another validator object, one associated to VkInstance,
and "this" validator is associated to a device. enabled_features is not inherited, and besides
would be reset in `GetEnabledDeviceFeatures`.
The switch from one to the other happens in `state_tracker.cpp`, `ValidationStateTracker::PostCallRecordCreateDevice`
TL;DR is the following type of sanity check is currently invalid, but it would be nice to have

spencer-lunarg:
so I don't get the problem here fully

yes, there are currently two `DeviceFeatures enabled_features`, one for the Instance level and one for Device level, but the Instance Level one should never matter for GPU-AV

I think the only call that we use as an instance level is `PostCallRecordGetPhysicalDeviceProperties` and it doesn't care about the `enabled_features` there

-----

So the flow should be

- We call `GetEnabledDeviceFeatures()` which sets it
- We check if not enabled, and set it to true and enable it
- It is marked as `true` everywhere as expected

arno-lunarg:
I updated the issue

PR #8002, opened 2024-05-13
gpu: Remove feature check

arno-lunarg:
Feature check is invalid, see https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8001

### DIFF
--- a/layers/gpu_validation/gpu_validation.cpp
+++ b/layers/gpu_validation/gpu_validation.cpp
@@ -45,12 +45,15 @@
 namespace gpuav {
 
 VkDeviceAddress Validator::GetBufferDeviceAddress(VkBuffer buffer, const Location &loc) const {
-    if (!enabled_features.bufferDeviceAddress) {
-        assert(false);
-        ReportSetupProblem(buffer, loc, "bufferDeviceAddress feature not enabled, calling GetBufferDeviceAddress is invalid.");
-        aborted = true;
-        return 0;
-    }
+    // https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8001
+    // Setting enabled_features.bufferDeviceAddress to true in GpuShaderInstrumentor::PreCallRecordCreateDevice
+    // when adding missing features will modify another validator object, one associated to VkInstance,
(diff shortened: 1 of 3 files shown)


layers/gpu_validation/gpu_validation.cpp
```diff
@@ -45,12 +45,14 @@
 namespace gpuav {
 
 VkDeviceAddress Validator::GetBufferDeviceAddress(VkBuffer buffer, const Location &loc) const {
-    if (!enabled_features.bufferDeviceAddress) {
-        assert(false);
-        ReportSetupProblem(buffer, loc, "bufferDeviceAddress feature not enabled, calling GetBufferDeviceAddress is invalid.");
-        aborted = true;
-        return 0;
-    }
+    // https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8001
```
spencer-lunarg, line 48: I don't get why this is not just a 

`assert(enabled_features.bufferDeviceAddress)`

I don't know why we are adding this comment now if we are assuming the feature is enabled

layers/gpu_validation/gpu_validation.cpp
```diff
@@ -45,12 +45,14 @@
...
+    // Setting enabled_features.bufferDeviceAddress to true in GpuShaderInstrumentor::PreCallRecordCreateDevice
+    // when adding missing features will modify another validator object, one associated to VkInstance,
+    // and "this" validator is associated to a device. enabled_features is not inherited, and besides
+    // would be reset in GetEnabledDeviceFeatures.
+    // The switch from the instance validator object to the device one happens in
+    // `state_tracker.cpp`, `ValidationStateTracker::PostCallRecordCreateDevice`
+    // TL;DR is the following type of sanity check is currently invalid, but it would be nice to have
```
spencer-lunarg, line 55: this is no pointing to anything now

spencer-lunarg (approved):

layers/gpu_validation/gpu_validation.cpp
arno-lunarg: Issue description was bad, I updated it hopefully it is better

PR #8006, opened 2024-05-14
layers: Add vk_struct_compare.h

spencer-lunarg:
There are some other structs that we will need to do "compare the structs are identical" and want to make a separate PR first to move the logic to its own file


arno-lunarg (changes requested):
Nice, but functions definitions need to be in their own .cpp to not worsen compilation time

spencer-lunarg:
> Nice! But functions definitions need to be in their own .cpp to not worsen compilation time

good call, done

arno-lunarg (approved):

Issue #7993, opened 2024-05-10
Assertion when enabling GPU-VA with Linear Memory Allocation Mode disabled


f32by:
**Environment:**
 - OS: Windows 11
 - GPU and driver version: NVIDIA 552.44
 - SDK or header version if building from repo: built from source
 - Options enabled (synchronization, best practices, etc.): GPU-VA but with Linear Memory Allocation Mode disabled

**Describe the Issue**

Assertion when enabling GPU-VA with Linear Memory Allocation Mode disabled

**Expected behavior**

At least GPU-VA should work without Linear Memory Allocation Mode

**Valid Usage ID**
None

**Additional context**

Looks like [creation of `output_buffer_pool`](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/e74bfc215bf985d5cb515dcbd972b7b33611ba63/layers/gpu_validation/gpu_setup.cpp#L225) is guarded under `gpuav_settings.vma_linear_output` at [line 206](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/e74bfc215bf985d5cb515dcbd972b7b33611ba63/layers/gpu_validation/gpu_setup.cpp#L206) in `gpu_setup.cpp` and if Linear Memory Allocation Mode is disabled, the VMA pool will not be created and assetion at [line 269](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/e74bfc215bf985d5cb515dcbd972b7b33611ba63/layers/gpu_validation/gpu_setup.cpp#L269) fails.

spencer-lunarg:
@f32by thanks for bringing this up, @arno-lunarg and I were literally talking last week how we have some settings (like VMA Linear Memory Allocation Mode) that we are not properly testing values outside the defaults. Regardless, Arno will get this fixed quick next week I assume

PR #8007, opened 2024-05-14
gpu: Fix VMA linear pool flag setting handling

arno-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7993


spencer-lunarg (approved):
LGTM - From now on, during SDK release we need to add a "run without Linear Pool setting"

PR #8008, opened 2024-05-14
layers: Fix Image/Buffer comparing

spencer-lunarg:
from https://gitlab.khronos.org/vulkan/Vulkan-ValidationLayers/-/issues/35

These VUs say `identical` but it didn't mean the VkHandle, but the creation was the same

image

There is still a discussion which pNext this is suppose to account for, but for now covered the main cases

### DIFF
--- a/layers/state_tracker/buffer_state.cpp
+++ b/layers/state_tracker/buffer_state.cpp
@@ -73,6 +73,31 @@ Buffer::Buffer(ValidationStateTracker &dev_data, VkBuffer handle, const VkBuffer
     }
 }
 
+// This function is only used for comparing Imported External Dedicated Memory
+bool Buffer::CompareCreateInfo(const Buffer &other) const {
+    bool valid_queue_family = true;
+    if (create_info.sharingMode == VK_SHARING_MODE_CONCURRENT) {
+        if (create_info.queueFamilyIndexCount != other.create_info.queueFamilyIndexCount) {
+            valid_queue_family = false;
+        } else {
+            for (uint32_t i = 0; i < create_info.queueFamilyIndexCount; i++) {
+                if (create_info.pQueueFamilyIndices[i] != other.create_info.pQueueFamilyIndices[i]) {
+                    valid_queue_family = false;
+                    break;
+                }
+            }
+        }
+    }
+
+    // There are limitations what actually needs to be compared, so for simplicity (until found otherwise needed), we only need to
+    // check the ExternalHandleType and not other pNext chains
+    const bool valid_external = GetExternalHandleTypes(&create_info) == GetExternalHandleTypes(&other.create_info);
+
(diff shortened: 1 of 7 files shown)


layers/state_tracker/buffer_state.cpp
```diff
@@ -73,6 +73,30 @@ Buffer::Buffer(ValidationStateTracker &dev_data, VkBuffer handle, const VkBuffer
     }
 }
 
+// This Function is only used for comparing Imported External Dedicated Memory so only need to compare limited things and don't need
+// a complete check of all the pNexts
+bool Buffer::Compare(const Buffer &other) const {
```
arno-lunarg, line 78: I think the method name should reflect the limitations mentioned in the comment, otherwise I feel this is going to be misused

layers/state_tracker/image_state.cpp
```diff
@@ -371,6 +371,34 @@ void Image::SetSwapchain(std::shared_ptr<vvl::Swapchain> &swapchain, uint32_t sw
     bind_swapchain->AddParent(this);
 }
 
+// This Function is only used for comparing Imported External Dedicated Memory so only need to compare limited things and don't need
+// a complete check of all the pNexts
+bool Image::Compare(const Image &other) const {
```
arno-lunarg, line 376: Same

layers/state_tracker/buffer_state.cpp
arno-lunarg: Maybe something like `CompareBaseCreateInfos`?

tests/unit/external_memory_sync_positive.cpp
```diff
@@ -517,3 +572,57 @@ TEST_F(PositiveExternalMemorySync, MultipleExportOpaqueFd) {
...
+
+    dedicated_info.buffer = buffer2.handle();
+
+    VkImportMemoryFdInfoKHR import_info = vku::InitStructHelper(&dedicated_info);
+    import_info.handleType = VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD_BIT;
+    import_info.fd = fd;
+
+    alloc_info = vkt::DeviceMemory::get_resource_alloc_info(*m_device, buffer2.memory_requirements(), 0, &import_info);
+    vkt::DeviceMemory memory_import(*m_device, alloc_info);
+}
```
arno-lunarg, line 628: image

arno-lunarg (approved):

layers/state_tracker/buffer_state.cpp
spencer-lunarg: I'll rename and re-comment this

basically we never should need to compare two object create info like this, outside these rare external memory VUs, those themselves have limitation which pNext are even allowed

PR #8012, opened 2024-05-15
scripts: Generate Dynamic State error messages

spencer-lunarg:
This provides a better error message for dynamic state related messages

### DIFF
--- a/layers/core_checks/cc_cmd_buffer_dynamic.cpp
+++ b/layers/core_checks/cc_cmd_buffer_dynamic.cpp
@@ -50,8 +50,9 @@ bool CoreChecks::ValidateGraphicsDynamicStateSetStatus(const LastBound& last_bou
         (cb_state.dynamic_state_status.pipeline ^ pipeline.dynamic_state) & cb_state.dynamic_state_status.pipeline;
     if (unset_status_pipeline.any()) {
         skip |= LogError(vuid.dynamic_state_setting_commands_08608, objlist, loc,
-                         "%s doesn't set up %s, but it calls the related dynamic state setting commands.",
-                         FormatHandle(pipeline).c_str(), DynamicStatesToString(unset_status_pipeline).c_str());
+                         "%s doesn't set up %s, but it calls the related dynamic state setting commands (%s).",
+                         FormatHandle(pipeline).c_str(), DynamicStatesToString(unset_status_pipeline).c_str(),
+                         DynamicStatesCommandsToString(unset_status_pipeline).c_str());
     }
 
     // build the mask of what has been set in the Pipeline, but yet to be set in the Command Buffer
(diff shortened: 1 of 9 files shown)


layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -31,7 +31,8 @@
 bool CoreChecks::ValidateDynamicStateIsSet(CBDynamicFlags state_status_cb, CBDynamicState dynamic_state,
                                            const LogObjectList& objlist, const Location& loc, const char* vuid) const {
     if (!state_status_cb[dynamic_state]) {
-        return LogError(vuid, objlist, loc, "%s state not set for this command buffer.", DynamicStateToString(dynamic_state));
```
spencer-lunarg, line 34: before

> VK_DYNAMIC_STATE_LINE_WIDTH state not set for this command buffer

now

> VK_DYNAMIC_STATE_LINE_WIDTH state is dynamic, but the command buffer never called vkCmdSetLineWidth

scripts/generators/dynamic_state_generator.py
```diff
@@ -17,6 +17,227 @@
 import os
 from generators.base_generator import BaseGenerator
 
+# TODO - Get in vk.xml
+# This is a representation of all the dynamic state information
+dynamic_state_map = {
```
spencer-lunarg, line 22: the goal is to build this up until I know all things needed, from there, I can make an effort to bring to the `vk.xml`, but then it will be simple as it will just be replacing this python dict for another python dict

layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -49,8 +50,9 @@ bool CoreChecks::ValidateGraphicsDynamicStateSetStatus(const LastBound& last_bou
         (cb_state.dynamic_state_status.pipeline ^ pipeline.dynamic_state) & cb_state.dynamic_state_status.pipeline;
     if (unset_status_pipeline.any()) {
         skip |= LogError(vuid.dynamic_state_setting_commands_08608, objlist, loc,
```
spencer-lunarg, line 52: Before

> vkCmdDraw():  VkPipeline 0xd5b26f0000000010[] doesn't set up VK_DYNAMIC_STATE_SCISSOR|VK_DYNAMIC_STATE_LINE_WIDTH, but it calls the related dynamic state setting commands.

Now 

> vkCmdDraw():  VkPipeline 0xd5b26f0000000010[] doesn't set up VK_DYNAMIC_STATE_SCISSOR|VK_DYNAMIC_STATE_LINE_WIDTH, but it calls the related dynamic state setting commands (vkCmdSetScissor, vkCmdSetLineWidth)

layers/core_checks/cc_cmd_buffer_dynamic.cpp
artem-lunarg: AI is great, but this is better

artem-lunarg (approved):

Issue #8020, opened 2024-05-16
Apply shader instrumentation caching determined on settings


spencer-lunarg:
from https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8016

We should take in account the GPU-AV setting to determined the shader hashing otherwise we hit nasty edge cases as found in the issue above

PR #8021, opened 2024-05-16
gpu: Fix shader caching

arno-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8020

Take into account GPU AV settings

Tested on lightweight VK, with this change caching works as expected, removing false positives

### DIFF
--- a/layers/gpu_validation/gpu_setup.cpp
+++ b/layers/gpu_validation/gpu_setup.cpp
@@ -235,9 +235,10 @@ void Validator::CreateDevice(const VkDeviceCreateInfo *pCreateInfo, const Locati
 
         std::ifstream file_stream(instrumented_shader_cache_path, std::ifstream::in | std::ifstream::binary);
         if (file_stream) {
-            char inst_shader_hash[sizeof(INST_SHADER_GIT_HASH)];
+            ShaderCacheHash shader_cache_hash(gpuav_settings);
+            char inst_shader_hash[sizeof(shader_cache_hash)];
             file_stream.read(inst_shader_hash, sizeof(inst_shader_hash));
-            if (!strncmp(inst_shader_hash, INST_SHADER_GIT_HASH, sizeof(INST_SHADER_GIT_HASH))) {
+            if (std::memcmp(inst_shader_hash, reinterpret_cast<char *>(&shader_cache_hash), sizeof(shader_cache_hash)) == 0) {
                 uint32_t num_shaders = 0;
                 file_stream.read(reinterpret_cast<char *>(&num_shaders), sizeof(uint32_t));
                 for (uint32_t i = 0; i < num_shaders; ++i) {
--- a/layers/gpu_validation/gpu_record.cpp
+++ b/layers/gpu_validation/gpu_record.cpp
@@ -179,7 +179,8 @@ void Validator::PreCallRecordDestroyDevice(VkDevice device, const VkAllocationCa
     if (gpuav_settings.cache_instrumented_shaders && !instrumented_shaders.empty()) {
         std::ofstream file_stream(instrumented_shader_cache_path, std::ofstream::out | std::ofstream::binary);
         if (file_stream) {
-            file_stream.write(INST_SHADER_GIT_HASH, sizeof(INST_SHADER_GIT_HASH));
+            ShaderCacheHash shader_cache_hash(gpuav_settings);
+            file_stream.write(reinterpret_cast<const char *>(&shader_cache_hash), sizeof(shader_cache_hash));
             uint32_t datasize = static_cast<uint32_t>(instrumented_shaders.size());
             file_stream.write(reinterpret_cast<char *>(&datasize), sizeof(uint32_t));
             for (auto &record : instrumented_shaders) {
--- a/layers/gpu_validation/gpu_settings.h
+++ b/layers/gpu_validation/gpu_settings.h
@@ -18,6 +18,8 @@
 #pragma once
 // Default values for those settings should match layers/VkLayer_khronos_validation.json.in
 
+#include "generated/gpu_inst_shader_hash.h"
+
 struct GpuAVSettings {
     bool shader_instrumentation_enabled = true;
     bool validate_descriptors = true;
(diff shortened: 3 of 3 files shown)


layers/gpu_validation/gpu_setup.cpp
```diff
@@ -235,9 +235,16 @@ void Validator::CreateDevice(const VkDeviceCreateInfo *pCreateInfo, const Locati
 
         std::ifstream file_stream(instrumented_shader_cache_path, std::ifstream::in | std::ifstream::binary);
         if (file_stream) {
-            char inst_shader_hash[sizeof(INST_SHADER_GIT_HASH)];
-            file_stream.read(inst_shader_hash, sizeof(inst_shader_hash));
-            if (!strncmp(inst_shader_hash, INST_SHADER_GIT_HASH, sizeof(INST_SHADER_GIT_HASH))) {
+            struct ShaderCacheHash {
+                GpuAVSettings gpuav_settings{};
+                const char inst_shader_git_hash[sizeof(INST_SHADER_GIT_HASH)] = INST_SHADER_GIT_HASH;
+                char _pad[Align<size_t>(sizeof(GpuAVSettings) + sizeof(INST_SHADER_GIT_HASH), 4) - sizeof(GpuAVSettings) -
```
artem-lunarg, line 241: `_pad` - starting with underscore is probably reserved for the complier in c++.

If  `Align<size_t>(sizeof(GpuAVSettings) + sizeof(INST_SHADER_GIT_HASH), 4) - sizeof(GpuAVSettings) - sizeof(INST_SHADER_GIT_HASH)`  is zero then it is illegal to declare array of size zero. Probably it is not zero now, but might become if other fields change.

arno-lunarg: yeah this pad thing is garbage but does the job and seems to compile fine in CI... basically I just want to pad if needed. Probably some template shenanigans would allow me to do that, but I do not know what : /

artem-lunarg: Maybe the above can be achieved by using `struct alignas(4) ShaderCacheHash`?? This should ensure that struct size is also multiple of 4 in order to have correct alignment when structs are put in array.

arno-lunarg: What is want is access to the padding byte, to make sure they are always 0, because of the memory write and compare we do in this portion of code. 
If I do not explicitly add padding, not sure the trailing bytes are guaranteed to be 0?

arno-lunarg: struct mem layout looks like this in VS:
image

arno-lunarg: and `sizeof(ShaderCacheHash)`, without explicit padding, still gives me 64

arno-lunarg: But you make me realize that instead of adjusting by adding explicit padding, I could adjust the memory write and compare sizes, it will be less messy

arno-lunarg:
I am missing the shader cache hash write, in the process of adding that....

arno-lunarg:
Should be good now, I made sure that we hit the cache when we need, and only then

layers/gpu_validation/gpu_setup.cpp
```diff
@@ -235,9 +235,12 @@ void Validator::CreateDevice(const VkDeviceCreateInfo *pCreateInfo, const Locati
 
         std::ifstream file_stream(instrumented_shader_cache_path, std::ifstream::in | std::ifstream::binary);
         if (file_stream) {
-            char inst_shader_hash[sizeof(INST_SHADER_GIT_HASH)];
-            file_stream.read(inst_shader_hash, sizeof(inst_shader_hash));
-            if (!strncmp(inst_shader_hash, INST_SHADER_GIT_HASH, sizeof(INST_SHADER_GIT_HASH))) {
+            ShaderCacheHash shader_cache_hash{};
+            shader_cache_hash.gpuav_settings = gpuav_settings;
+            constexpr size_t unpadded_shader_cash_hash_byte_size = sizeof(GpuAVSettings) + sizeof(INST_SHADER_GIT_HASH);
```
artem-lunarg, line 240: This assumes there is no padding between GpuAvSettings and GIT_HASH fields, not true in general.. I dont know if expression Obj obj{} initializes padding to zero, if yes, might help.

This stuff is always complicated, now I think maybe not that bad. I also like solution where you compute pad size manually and use constant and then after structure have static_assert(sizeof(struct) == sizeof(field1) + sizeof(field2) + padding_array_size_constant) so it will always complain at compile time and struct definition will be simple, but will need adjustment to padding_array_size_constant if structure definition changes.

arno-lunarg: I went for `#pragma pack(push, 1)` approach, makes things simpler indeed!

artem-lunarg (approved):

spencer-lunarg (commented):
code is fine, but before we merge this, please let me test this tomorrow on Linux to make sure it works as intended

layers/gpu_validation/gpu_setup.cpp
```diff
@@ -235,9 +235,10 @@ void Validator::CreateDevice(const VkDeviceCreateInfo *pCreateInfo, const Locati
 
         std::ifstream file_stream(instrumented_shader_cache_path, std::ifstream::in | std::ifstream::binary);
         if (file_stream) {
-            char inst_shader_hash[sizeof(INST_SHADER_GIT_HASH)];
+            ShaderCacheHash shader_cache_hash(gpuav_settings);
+            char inst_shader_hash[sizeof(shader_cache_hash)];
             file_stream.read(inst_shader_hash, sizeof(inst_shader_hash));
-            if (!strncmp(inst_shader_hash, INST_SHADER_GIT_HASH, sizeof(INST_SHADER_GIT_HASH))) {
+            if (!strncmp(inst_shader_hash, reinterpret_cast<char *>(&shader_cache_hash), sizeof(shader_cache_hash))) {
```
spencer-lunarg, line 241: before we had a `uint64_t` we were reading, but not we have a struct with bool that I assume are `0x0` should we be using `memcmp` as I am not sure why this is a string compare to be honest

spencer-lunarg: yes, confirmed this was happening to me and so this is not working as intended

image

spencer-lunarg (changes requested):
need to use `memcmp`

layers/gpu_validation/gpu_setup.cpp
arno-lunarg: Oh yes good catch!

arno-lunarg: (before I was an actual string so `strncmp` usage was warranted)

spencer-lunarg (approved):

PR #8023, opened 2024-05-16
Null check all Get<> to state object

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7975

This was me going through every `Get<vvl::*>` to make sure we are null-checking them


artem-lunarg (approved):
LGTM

PR #8024, opened 2024-05-17
More StateTracker Null Check cleanup

spencer-lunarg:
Small Null Check follow-up

### DIFF
--- a/layers/best_practices/bp_instance_device.cpp
+++ b/layers/best_practices/bp_instance_device.cpp
@@ -167,7 +167,7 @@ bool BestPractices::PreCallValidateCreateDevice(VkPhysicalDevice physicalDevice,
     }
 
     const auto bp_pd_state = Get<bp_state::PhysicalDevice>(physicalDevice);
-    if ((bp_pd_state->vkGetPhysicalDeviceFeaturesState == UNCALLED) && (pCreateInfo->pEnabledFeatures != NULL)) {
+    if (bp_pd_state && (bp_pd_state->vkGetPhysicalDeviceFeaturesState == UNCALLED) && (pCreateInfo->pEnabledFeatures != NULL)) {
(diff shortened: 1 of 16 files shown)


layers/best_practices/bp_instance_device.cpp
```diff
@@ -167,7 +167,7 @@ bool BestPractices::PreCallValidateCreateDevice(VkPhysicalDevice physicalDevice,
     }
 
     const auto bp_pd_state = Get<bp_state::PhysicalDevice>(physicalDevice);
-    if ((bp_pd_state->vkGetPhysicalDeviceFeaturesState == UNCALLED) && (pCreateInfo->pEnabledFeatures != NULL)) {
+    if (bp_pd_state && (bp_pd_state->vkGetPhysicalDeviceFeaturesState == UNCALLED) && (pCreateInfo->pEnabledFeatures != NULL)) {
```
spencer-lunarg, line 170: we shouldn't need this for `bp_state::PhysicalDevice` since its the dispatchable handle.... but making them all consistent for now until we change things to be tied to the dispatchable handles

layers/core_checks/cc_descriptor.cpp
```diff
@@ -1127,21 +1127,20 @@ bool CoreChecks::ValidateCopyUpdate(const VkCopyDescriptorSet &update, const Loc
 // Validate given sampler. Currently this only checks to make sure it exists in the samplerMap
 bool CoreChecks::ValidateSampler(const VkSampler sampler) const { return Get<vvl::Sampler>(sampler).get() != nullptr; }
 
-bool CoreChecks::ValidateImageUpdate(VkImageView image_view, VkImageLayout image_layout, VkDescriptorType type,
```
spencer-lunarg, line 1130: there were a few functions where we were doing the `Get<>` inside but also everyone calling it already had it, so was no need

artem-lunarg (approved):

PR #8026, opened 2024-05-17
gpu: Have InternalError disconnect from chassis

spencer-lunarg:
- rename `ReportSetupProblem` to `InternalError` (as it does more then just during setup)
- Call `ReleaseDeviceDispatchObject` that will prevent any more `Pre/Post` chassis calls from getting called

For `aborted` we still need it, but not to "gaurd" a function, but to get out of a function, example

```
void DeepDownFunction() {
    AllocationOfThing();
    if (aborted) return;
}
```

We might decide in the future to do a `[[no-discard]]` bool return from these functions, but that is orthogonal to the scope of this change

### DIFF
--- a/layers/gpu_validation/debug_printf.cpp
+++ b/layers/gpu_validation/debug_printf.cpp
@@ -25,7 +25,7 @@
 // Perform initializations that can be done at Create Device time.
 void debug_printf::Validator::CreateDevice(const VkDeviceCreateInfo *pCreateInfo, const Location &loc) {
     if (enabled[gpu_validation]) {
-        ReportSetupProblem(device, loc, "Debug Printf cannot be enabled when gpu assisted validation is enabled.");
+        InternalError(device, loc, "Debug Printf cannot be enabled when gpu assisted validation is enabled.");
         return;
     }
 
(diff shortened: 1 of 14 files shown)


tests/unit/gpu_av.cpp
```diff
@@ -57,9 +57,23 @@ TEST_F(NegativeGpuAV, ValidationAbort) {
     features.vertexPipelineStoresAndAtomics = false;
     features.fragmentStoresAndAtomics = false;
     fpvkSetPhysicalDeviceFeaturesEXT(gpu(), features);
-    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "GPU-Assisted Validation disabled");
+    m_errorMonitor->SetDesiredFailureMsg(kErrorBit, "GPU-AV is being disabled");
     RETURN_IF_SKIP(InitState());
     m_errorMonitor->VerifyFound();
+
+    // Still make sure we can use Vulkan as expected without errors
```
arno-lunarg, line 64: Nice addition !

arno-lunarg (approved):

layers/gpu_validation/debug_printf.cpp
```diff
@@ -743,10 +747,3 @@ void debug_printf::CommandBuffer::ResetCBState() {
     }
     buffer_infos.clear();
 }
-
-// TODO - Remove this function once ReleaseDeviceDispatchObject() is added to GPU-AV as well
```
arno-lunarg, line 747: the-thing

layers/gpu_validation/gpu_shader_instrumentor.cpp
```diff
@@ -948,3 +926,27 @@ void GpuShaderInstrumentor::PostCallRecordPipelineCreations(const uint32_t count
...
+        error_message += " VMA statistics = ";
+        error_message += stats_string;
+        vmaFreeStatsString(vmaAllocator, stats_string);
+    }
+
+    char const *layer_name = container_type == LayerObjectTypeDebugPrintf ? "Debug PrintF" : "GPU-AV";
+    char const *vuid =
+        container_type == LayerObjectTypeDebugPrintf ? "UNASSIGNED-DEBUG-PRINTF" : "UNASSIGNED-GPU-Assisted-Validation";
+
+    LogError(vuid, objlist, loc, "Setup Error, %s is being disabled. Detail: (%s)", layer_name, error_message.c_str());
```
arno-lunarg, line 946: Maybe:
```suggestion
    LogError(vuid, objlist, loc, "Setup Error, %s is being disabled. Detail:\n%s", layer_name, error_message.c_str());
```

PR #8036, opened 2024-05-20
Fix VkPhysicalDeviceIDProperties sType validation

krOoze:
`VkPhysicalDeviceIDProperties` is provided from multiple extensions. Add handling of this cornercase into layers. Fix tests that were defective to discover it.

### DIFF
--- a/layers/vulkan/generated/stateless_validation_helper.cpp
+++ b/layers/vulkan/generated/stateless_validation_helper.cpp
@@ -4184,11 +4184,14 @@ bool StatelessValidation::ValidatePnextPropertyStructContents(const Location& lo
         // Validation code for VkPhysicalDeviceIDProperties structure members
         case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_ID_PROPERTIES: {  // Covers VUID-VkPhysicalDeviceIDProperties-sType-sType
 
-            if (!IsExtEnabled(instance_extensions.vk_khr_external_fence_capabilities)) {
+            if (!IsExtEnabled(instance_extensions.vk_khr_external_fence_capabilities) &&
+                !IsExtEnabled(instance_extensions.vk_khr_external_memory_capabilities) &&
+                !IsExtEnabled(instance_extensions.vk_khr_external_semaphore_capabilities)) {
                 skip |= LogError(pnext_vuid, instance, loc.dot(Field::pNext),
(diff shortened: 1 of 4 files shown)


layers/vulkan/generated/stateless_validation_helper.cpp
```diff
@@ -4184,7 +4185,9 @@ bool StatelessValidation::ValidatePnextPropertyStructContents(const Location& lo
         // Validation code for VkPhysicalDeviceIDProperties structure members
         case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_ID_PROPERTIES: {  // Covers VUID-VkPhysicalDeviceIDProperties-sType-sType
 
-            if (!IsExtEnabled(instance_extensions.vk_khr_external_fence_capabilities)) {
+            if (!IsExtEnabled(instance_extensions.vk_khr_external_semaphore_capabilities) &&
```
spencer-lunarg, line 4188: :100:

spencer-lunarg (approved):
thanks @krOoze - this is the change I wanted to have, just didn't have time, but this looks awesome and does what it should have done the first time

scripts/generators/stateless_validation_helper_generator.py
```diff
@@ -1001,24 +1003,27 @@ def genStructBody(self, struct: Struct, non_prop_feature: bool):
...
-            else:
-                extension_check = f'if (!IsExtEnabled(instance_extensions.{extension.name.lower()})) {{'
+            extension_conditinals = list()
+            for ext_name in ext_names:
+                extension = self.vk.extensions[ext_name]
+                if extension.device:
+                    extension_conditinals.append( f'((is_physdev_api && !SupportedByPdev(physical_device, vvl::Extension::_{extension.name})) || (!is_physdev_api && !IsExtEnabled(device_extensions.{extension.name.lower()})))' )
+                else:
+                    extension_conditinals.append( f'!IsExtEnabled(instance_extensions.{extension.name.lower()})' )
+            extension_check = f'if ({' && '.join(extension_conditinals)}) {{'
```
spencer-lunarg, line 1026:
```suggestion
            extension_check = f'if ({" && ".join(extension_conditinals)}) {{'
```

should work... this is a python 3.12 feature I think but we are still at 3.10 python as the minimum requirements

spencer-lunarg: (CI is at 3.10 which is failing on this)

krOoze: I squashed it in

PR #8040, opened 2024-05-21
layers: Invalidate dynamic state from previous static state

spencer-lunarg:
For https://github.com/KhronosGroup/Vulkan-Docs/issues/2276 (internal spec change https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6666)

The following case was never being validated

```
vkCmdSet()
vkCmdBindPipeline() // static
vkCmdBindPipeline() // dynamic
vkCmdDraw()
```

This also adds better error messaging to let the developer know that dynamic state was invalidated

### DIFF
--- a/layers/state_tracker/cmd_buffer_state.h
+++ b/layers/state_tracker/cmd_buffer_state.h
@@ -186,13 +186,20 @@ class CommandBuffer : public RefcountedStateObject {
 
     // Track status of all vkCmdSet* calls, if 1, means it was set
     struct DynamicStateStatus {
-        CBDynamicFlags cb;        // for lifetime of CommandBuffer
+        CBDynamicFlags cb;        // for lifetime of CommandBuffer (invalidated if static pipeline is bound)
         CBDynamicFlags pipeline;  // for lifetime since last bound pipeline
+
+        CBDynamicFlags history;  // for lifetime of CommandBuffer, regardless if invalidated, used for better error messages
 
         // There is currently only a single non-graphics dynamic state, for now manage manually to save memory
         bool rtx_stack_size_cb;        // for lifetime of CommandBuffer
         bool rtx_stack_size_pipeline;  // for lifetime since last bound pipeline
     } dynamic_state_status;
+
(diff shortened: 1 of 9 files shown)


layers/state_tracker/cmd_buffer_state.h
```diff
@@ -186,7 +186,8 @@ class CommandBuffer : public RefcountedStateObject {
 
     // Track status of all vkCmdSet* calls, if 1, means it was set
     struct DynamicStateStatus {
-        CBDynamicFlags cb;        // for lifetime of CommandBuffer
+        CBDynamicFlags all;       // for lifetime of CommandBuffer, regardless if invalidated, used for better error messages
```
artem-lunarg, line 189: nit: nice name but can it be too casual so can be used by mistake, if it's only for error reporting? Would you like something more scary here in a style of cb_potentially_invalidated.

layers/state_tracker/cmd_buffer_state.h
```diff
@@ -310,31 +311,56 @@ class CommandBuffer : public RefcountedStateObject {
 
         // When the Command Buffer resets, the value most things in this struct don't matter because if they are read without
         // setting the state, it will fail in ValidateDynamicStateIsSet() for us. Some values (ex. the bitset) are tracking in
-        // replacement for static_status/dynamic_status so this needs to reset along with those
-        void reset() {
+        // replacement for static_status/dynamic_status so this needs to reset along with those.
+        //
+        // The only time this is reset is when the command buffer is reset, and vkCmdBindPipeline for static state
+        void reset(CBDynamicFlags mask) {
```
artem-lunarg, line 317: Parameter name. Name that suggests this is the flags to _keep_ might be more self documenting.

artem-lunarg (approved):

layers/state_tracker/cmd_buffer_state.h
spencer-lunarg: yep, good call (also for the `reset` as well)

PR #8041, opened 2024-05-21
layers: Add Shader Val Cache for ShaderObject

spencer-lunarg:
Adds `ValidationCache` to Shader Object so we don't repeat running `spirv-val` on binaries across multiple runs

### DIFF
--- a/layers/core_checks/cc_shader_object.cpp
+++ b/layers/core_checks/cc_shader_object.cpp
@@ -288,16 +288,8 @@ bool CoreChecks::PreCallValidateCreateShadersEXT(VkDevice device, uint32_t creat
         }
         const Location create_info_loc = error_obj.location.dot(Field::pCreateInfos, i);
 
-        const uint32_t hash = cache ? hash_util::ShaderHash(create_info.pCode, create_info.codeSize) : 0;
-        if (!cache || !cache->Contains(hash)) {
-            spv_const_binary_t binary{static_cast<const uint32_t*>(create_info.pCode), create_info.codeSize / sizeof(uint32_t)};
-            skip |= RunSpirvValidation(binary, create_info_loc);
-        }
-
-        // No point to cache anything that is not valid
-        if (!skip && cache) {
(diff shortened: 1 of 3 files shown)


layers/core_checks/cc_shader_object.cpp
```diff
@@ -270,61 +276,73 @@ bool CoreChecks::PreCallValidateCreateShadersEXT(VkDevice device, uint32_t creat
...
-                        skip |= LogError("VUID-VkShaderCreateInfoEXT-codeType-08873", device, create_info_loc.dot(Field::stage),
-                                         "is VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT, but orientation is not specified.");
-                    }
-                    if (stage_state.entrypoint->execution_mode.tessellation_spacing == 0) {
-                        skip |= LogError("VUID-VkShaderCreateInfoEXT-codeType-08874", device, create_info_loc.dot(Field::stage),
-                                         "is VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT, but spacing is not specified.");
-                    }
-                }
+        // No point to cache anything that is not valid
+        if (!skip && cache) {
```
arno-lunarg, line 298: If I am correct, `skip` could be `false` and the shader could still be invalid, if for instance the VUID that has been emitted was in the ignore list. 
But then you could say that if a VUID is in the ignore list, we consider that a shader emitting is still valid...

Maybe a more general question is then: do we consider that "something" is valid if and only if the `skip` value computed during validation if `false`?

arno-lunarg (approved):
LGTM, one question

layers/core_checks/cc_shader_object.cpp
spencer-lunarg: so yes you are correct

1. I'm going just move the `ValidationCache` inside `RunSpirvValidation` so we know for sure
2. yes, we currently associate `valid` with `skip == false` which I think it what it should be, but needs some tweaking with the internal of `LogError`

Issue #8052, opened 2024-05-24
Crash in DescriptorSet::PerformWriteUpdate


ChunkTreasure1:
**Environment:**
 - OS: Windows 11
 - GPU and driver version: RTX 3080, Driver: 545.84
 - SDK or header version if building from repo: 1.3.283.0
 - Options enabled (synchronization, best practices, etc.): Core

**Describe the Issue**

A crash happens within the validation layers when trying to call vkUpdateWriteDescriptors with a VkWriteDescriptorSet that has a dstArrayElement larger than or equal to the max array size of the dstBinding.

**Expected behavior**

This should not crash, and instead output the validation error VUID-VkWriteDescriptorSet-dstArrayElement-00321.

**Valid Usage ID**
None

**Additional context**
The crash happens in layers/state_tracker/descriptor_sets.cpp and in the function PerformWriteUpdate.

PR #8066, opened 2024-05-29
layer: Add ASSERT_AND_RETURN

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8052

follow-up of  https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8024

There are a LOT of early return and assert and no idea which are "valid to hit" and which we should never hit (but clearly in the issue we do sometimes). This will make it more clear when we should be asserting or not... but if in a release build and the assert doesn't trigger, we still need to fail gracefully

### DIFF
--- a/layers/utils/vk_layer_utils.h
+++ b/layers/utils/vk_layer_utils.h
@@ -55,6 +55,33 @@
 #endif
 #endif
 
+// There are many times we want to assert, but also it is highly important to not crash for release builds.
+// This Macro also makes it more obvious if we are returning early because of a known situation or if we are just guarding against
+// something wrong actually happening.
+#define ASSERT_AND_RETURN(cond) \
+    do {                        \
+        if (!(cond)) {          \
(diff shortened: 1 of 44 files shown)


layers/utils/vk_layer_utils.h
```diff
@@ -55,6 +55,27 @@
 #endif
 #endif
 
+// There are many times we want to assert, but also it is highly important to not crash for release builds.
+// This Macro also makes it more obvious if we are returning early because of a known situation or if we are just guarding against
+// something wrong actually happening.
+#define ASSERT_AND_RETURN(cond) \
```
jeremyg-lunarg, line 61: Usually it is good to enclose multi-line macros into a `do { } while(0);` "loop" see: https://hownot2code.wordpress.com/2016/12/05/do-while-0-in-macros/

Issue #6702, opened 2023-10-16
Validation for VK_EXT_nested_command_buffer


spencer-lunarg:
Should be the VUs that need to be added for the extension

- [x] VUID-vkCmdExecuteCommands-contents-06018
- [x] VUID-vkCmdExecuteCommands-commandBuffer-09375
- [ ] VUID-vkCmdExecuteCommands-nestedCommandBuffer-09376
- [ ] VUID-vkCmdExecuteCommands-nestedCommandBufferRendering-09377
- [ ] VUID-vkCmdExecuteCommands-nestedCommandBufferSimultaneousUse-09378
- [x] VUID-vkCmdBeginRendering-commandBuffer-06068
- [x] VUID-VkRenderingInfo-flags-09381
- [x] VUID-VkSubpassBeginInfo-contents-09382

PR #8070, opened 2024-05-30
layers: Add remaining VK_EXT_nested_command_buffer

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6702

### DIFF
--- a/layers/core_checks/cc_cmd_buffer.cpp
+++ b/layers/core_checks/cc_cmd_buffer.cpp
@@ -813,9 +813,19 @@ bool CoreChecks::PreCallValidateCmdExecuteCommands(VkCommandBuffer commandBuffer
         skip |= viewport_scissor_inheritance.VisitPrimary(cb_state);
     }
 
-    if (!cb_state.IsPrimary() && !enabled_features.nestedCommandBuffer) {
-        skip |= LogError("VUID-vkCmdExecuteCommands-commandBuffer-09375", commandBuffer,
-                         error_obj.location.dot(Field::commandBuffer), "is a secondary command buffer.");
+    if (!cb_state.IsPrimary()) {
+        if (!enabled_features.nestedCommandBuffer) {
+            skip |= LogError("VUID-vkCmdExecuteCommands-commandBuffer-09375", commandBuffer,
+                             error_obj.location.dot(Field::commandBuffer),
+                             "is a secondary command buffer (and the nestedCommandBuffer feature was not enabled).");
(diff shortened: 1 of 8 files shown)


pdaniell-nv (approved):
LGTM. Many thanks!

layers/core_checks/cc_cmd_buffer.cpp
```diff
@@ -915,6 +925,16 @@ bool CoreChecks::PreCallValidateCmdExecuteCommands(VkCommandBuffer commandBuffer
                     }
                 }
 
+                if ((sub_cb_state.beginInfo.flags & VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT) &&
+                    !enabled_features.nestedCommandBufferSimultaneousUse) {
+                    skip |=
+                        LogError("VUID-vkCmdExecuteCommands-nestedCommandBufferSimultaneousUse-09378", pCommandBuffers[i], cb_loc,
+                                 "(%s) is was recorded with VkCommandBufferBeginInfo::flag including "
```
artem-lunarg, line 932: "is was recorded" typo

layers/core_checks/cc_cmd_buffer.cpp
```diff
@@ -1209,6 +1229,19 @@ bool CoreChecks::PreCallValidateCmdExecuteCommands(VkCommandBuffer commandBuffer
                         }
                     }
                 }
+
+                // spec: "A maxCommandBufferNestingLevel of UINT32_MAX means there is no limit to the nesting level"
+                if (enabled_features.nestedCommandBuffer &&
+                    phys_dev_ext_props.nested_command_buffer_props.maxCommandBufferNestingLevel != UINT32_MAX) {
+                    if ((sub_cb_state.nesting_level + 1) >
```
artem-lunarg, line 1236: If someone creates a command buffer with nested level of UINT32_MAX, `sub_cb_state.nesting_level + 1` will result in 0, which won't trigger an error. This is not something one can do in practice, but this mostly to simplify code validation (you still need to think a little bit about this).
` if (sub_cb_state.nesting_level >=  phys_dev_ext_props.nested_command_buffer_props.maxCommandBufferNestingLevel)` does not wrap around.

spencer-lunarg: good catch

artem-lunarg (approved):

Issue #6730, opened 2023-10-18
missing VK_EXT_fragment_density_map / VK_KHR_fragment_shading_rate (RenderPass/Framebuffer)


spencer-lunarg:
There are surely more missing, mostly dealing with SPIR-V and AHB, but here are some I found that are focused on the Renderpass/Framebuffer side

- [x] VUID-VkFramebufferCreateInfo-flags-03196 (https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8230)
- [x] VUID-VkFramebufferCreateInfo-flags-03197 (https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8230)
- [ ] VUID-VkFramebufferCreateInfo-flags-04587
- [x] VUID-VkFramebufferCreateInfo-pAttachments-02552 (https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8227)
- [x] VUID-VkFramebufferCreateInfo-renderPass-02553
- [x] VUID-VkRenderPassCreateInfo2-pAttachments-09387
- [x] VUID-VkRenderPassFragmentDensityMapCreateInfoEXT-fragmentDensityMapAttachment-02548 (https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8252)

spencer-lunarg:
@r-potter note this issue was open, I marked off the few you added recently

r-potter:
Thanks @spencer-lunarg. We have something for 02548 on the way. 04587 needed a spec change to the VUID, which was signed off but not published last week

spencer-lunarg:
https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6766 will replace `VUID-VkFramebufferCreateInfo-renderPass-08921` with `VUID-VkFramebufferCreateInfo-flags-04587` so this is issue is done and closing

PR #8071, opened 2024-05-31
layers: Add VUID-VkFramebufferCreateInfo-renderPass-02553 check

ludviglindau-arm:
Adding a check for VUID-VkFramebufferCreateInfo-renderPass-02553 as we had some erroneous bug reports, I see it's on a list of unimplemented VUID's https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6730

I haven't created a test yet since the relevant tests seem to be disabled (https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/tests/unit/fragment_shading_rate.cpp#L2377)

### DIFF
--- a/layers/core_checks/cc_render_pass.cpp
+++ b/layers/core_checks/cc_render_pass.cpp
@@ -4507,47 +4507,56 @@ bool CoreChecks::PreCallValidateCreateFramebuffer(VkDevice device, const VkFrame
 
             if (enabled_features.fragmentDensityMap) {
                 const auto *fdm_attachment = vku::FindStructInPNextChain<VkRenderPassFragmentDensityMapCreateInfoEXT>(rpci->pNext);
-                if (fdm_attachment && fdm_attachment->fragmentDensityMapAttachment.attachment == i) {
-                    uint32_t ceiling_width = vvl::GetQuotientCeil(
-                        pCreateInfo->width, phys_dev_ext_props.fragment_density_map_props.maxFragmentDensityTexelSize.width);
-                    if (mip_width < ceiling_width) {
+                if (fdm_attachment && fdm_attachment->fragmentDensityMapAttachment.attachment != VK_ATTACHMENT_UNUSED) {
+                    if (fdm_attachment->fragmentDensityMapAttachment.attachment == i) {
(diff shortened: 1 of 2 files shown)


layers/core_checks/cc_render_pass.cpp
```diff
@@ -4507,47 +4507,58 @@ bool CoreChecks::PreCallValidateCreateFramebuffer(VkDevice device, const VkFrame
...
+                               (ici.flags & VK_IMAGE_CREATE_SUBSAMPLED_BIT_EXT) == 0) {
                         LogObjectList objlist(pCreateInfo->renderPass, image_views[i], ivci.image);
-                        skip |= LogError("VUID-VkFramebufferCreateInfo-renderPass-02746", objlist, attachment_loc,
-                                         "is referenced by "
-                                         "VkRenderPassFragmentDensityMapCreateInfoEXT::fragmentDensityMapAttachment in "
-                                         "the pNext chain, but it was create with subresourceRange.layerCount (%" PRIu32
-                                         ") different from 1.",
-                                         view_state->normalized_subresource_range.layerCount);
+                        skip |= LogError("VUID-VkFramebufferCreateInfo-renderPass-02553", objlist, attachment_loc,
+                                         "pAttachments[%" PRIu32
```
spencer-lunarg, line 4557: so `attachment_loc` will print out `pCreateInfo->pAttachments[0]` in the error message so this can just be `"is not created...."`

spencer-lunarg:
> I haven't created a test yet since the relevant tests seem to be disabled (https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/tests/unit/fragment_shading_rate.cpp#L2377)

So that test has one bad part, and it is why it **needs** to be broken up in to smaller tests, as we are currently not testing it all because of one small section... So it would be ideal to have a single test just for `VUID-VkFramebufferCreateInfo-renderPass-02553` called `TEST_F(NegativeFragmentShadingRate, FragmentDensityMapNonSubsampledImages)`

spencer-lunarg (approved):
Thanks for adding the test!

Once CI is all happy will get this in

PR #8087, opened 2024-06-04
tests: Fix tests on RADV

spencer-lunarg:
some edge cases that were failing on RADV

### DIFF
--- a/tests/unit/shader_object.cpp
+++ b/tests/unit/shader_object.cpp
@@ -5376,7 +5376,7 @@ TEST_F(NegativeShaderObject, InvalidColorWriteMask) {
         GTEST_SKIP() << "image format not supported as color attachment.";
     }
 
-    vkt::Image image(*m_device, 32, 32, 1, format, VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT);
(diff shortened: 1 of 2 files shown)


tests/unit/shader_object.cpp
```diff
@@ -6703,6 +6703,8 @@ TEST_F(NegativeShaderObject, MissingComputeFullSubgroups) {
     TEST_DESCRIPTION("Create shader with invalid flags when computeFullSubgroups is not enabled.");
 
     m_errorMonitor->SetDesiredError("VUID-VkShaderCreateInfoEXT-flags-09405");
+    // if subgroup size is not 32
+    m_errorMonitor->SetUnexpectedError("VUID-VkShaderCreateInfoEXT-flags-08417");
```
artem-lunarg, line 6707: If this test is about subgroups how important to specify valid number here? The popular options on the  market for the width of execution unit: 32 and 64 (nvidia/amd, and I guess intel too). Vulkan database shows there are mobile devices that advertise 128.

spencer-lunarg: I'm just going to make it 64, and query if it is over 64 we skip

artem-lunarg: good idea

spencer-lunarg: so found we had a more dynamic way to do this in another test

artem-lunarg (approved):

PR #8088, opened 2024-06-04
github: Remove redundant CI matrix

spencer-lunarg:
Help reduce CI minute pressure by removing our longest run that has caught nothing for us


artem-lunarg (approved):
LGTM. If the future technology improvements will solve CI performance, then it might be a good idea to enable this. Mostly from the perspective that debug build has different timings so can catch different type of threading errors than release build.

spencer-lunarg:
@artem-lunarg agree, we might just be moving this to an internal CI machine where it will run faster (and not hit our limits)

Issue #5208, opened 2023-02-02
Validate maxFragmentDualSrcAttachments when using dynamic blend state


spencer-lunarg:
`DISABLED_MaxFragmentDualSrcAttachmentsDynamicBlendEnable` was added to test to using dynamic blend state to test for `VUID-RuntimeSpirv-Fragment-06427`

to fix https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5199 this is still not working.

The current issue is managing the interactions with `VK_EXT_extended_dynamic_state3` and `VK_EXT_graphics_pipeline_library`

spencer-lunarg:
Spec change here https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6104

PR #8092, opened 2024-06-05
Add Dual Source Blending check

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5208

This adds `VUID-vkCmdDraw-maxFragmentDualSrcAttachments-09239`

Now that the dynamic state situation is better, was finally able to add this check in

### DIFF
--- a/layers/core_checks/cc_drawdispatch.cpp
+++ b/layers/core_checks/cc_drawdispatch.cpp
@@ -1533,6 +1533,7 @@ bool CoreChecks::ValidateActionState(const vvl::CommandBuffer &cb_state, const V
         skip |= ValidateDrawDynamicState(last_bound_state, vuid);
         skip |= ValidateDrawPrimitivesGeneratedQuery(last_bound_state, vuid);
         skip |= ValidateDrawProtectedMemory(last_bound_state, vuid);
+        skip |= ValidateDrawDualSourceBlend(last_bound_state, vuid);
 
         if (pipeline) {
             skip |= ValidateDrawPipeline(last_bound_state, *pipeline, vuid);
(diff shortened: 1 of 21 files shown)


layers/core_checks/cc_drawdispatch.cpp
```diff
@@ -1939,3 +1940,91 @@ bool CoreChecks::ValidateDrawProtectedMemory(const LastBound &last_bound_state,
...
+
+bool CoreChecks::ValidateDrawDualSourceBlend(const LastBound &last_bound_state, const vvl::DrawDispatchVuid &vuid) const {
+    bool skip = false;
+    const vvl::CommandBuffer &cb_state = last_bound_state.cb_state;
+    const auto *pipeline = last_bound_state.pipeline_state;
+    if (pipeline && !pipeline->ColorBlendState()) return skip;
+
+    // TODO - have better way to get fragment shader state
+    std::shared_ptr<const spirv::EntryPoint> entrypoint;
+    if (last_bound_state.pipeline_state) {
```
artem-lunarg, line 1952: There's `pipeline` shortcut and is already null-checked. Not sure if that's something temporary due to TODO.

spencer-lunarg: good catch

layers/core_checks/cc_drawdispatch.cpp
```diff
@@ -1939,3 +1940,91 @@ bool CoreChecks::ValidateDrawProtectedMemory(const LastBound &last_bound_state,
...
+        for (const auto &stage : last_bound_state.pipeline_state->stage_states) {
+            if (stage.entrypoint && (stage.entrypoint->stage == VK_SHADER_STAGE_FRAGMENT_BIT)) {
+                entrypoint = stage.entrypoint;
```
artem-lunarg, line 1955: nit: no `break`, probably fine for short loops, just highlighted to make sure that's something we do

spencer-lunarg: yep, should be a `break`

layers/core_checks/cc_drawdispatch.cpp
```diff
@@ -1939,3 +1940,91 @@ bool CoreChecks::ValidateDrawProtectedMemory(const LastBound &last_bound_state,
...
+        }
+    }
+    if (max_fragment_location < phys_dev_props.limits.maxFragmentDualSrcAttachments) return skip;
+
+    // If color blend is disabled, the blend equation doesn't matter
+    const bool dynamic_blend_enable = !pipeline || pipeline->IsDynamic(CB_DYNAMIC_STATE_COLOR_BLEND_ENABLE_EXT);
+    const bool dynamic_blend_equation = !pipeline || pipeline->IsDynamic(CB_DYNAMIC_STATE_COLOR_BLEND_EQUATION_EXT);
+    const uint32_t attachment_count = pipeline ? pipeline->ColorBlendState()->attachmentCount
+                                               : (uint32_t)cb_state.dynamic_state_value.color_blend_equations.size();
+    for (uint32_t i = 0; i < attachment_count; ++i) {
```
artem-lunarg, line 1981: It's a bit unclear  from VUID description if this should be reporte for every attachment in a loop, or the loop should be used to find an attachment that uses secondary blend source, and then if found report an error (assuming  max_fragment_location < phys_dev_props.limits.maxFragmentDualSrcAttachments). Did not think much about this, ignore if makes no sense, then I missed the point.

spencer-lunarg: I had to re-read the VU a few times and even looked at the gitlab issue that added it

in simple english - the idea is if any attachment is found to be blending with a dual-source blend factor, then "dual source blending" is enabled... from here, the max `Location` for the entire shader must be `maxFragmentDualSrcAttachments`

so the if `maxFragmentDualSrcAttachments` is `2`, the "dual source" attachment can be at `Location 0` or `2` ... still an error

there are tests for this as well

artem-lunarg: Okay, I thought that if dual source blending is determined as enabled, then there should be no attachment index greater-equal than the limit independent if it uses dual blend or not, but again see this the first time…

spencer-lunarg: image

This part of spec mentions why any dual source blending will just lower the overall max fragment output limit

spencer-lunarg: also remember, this was an open issue for a year because I never grasped dual source blending until spending an hour reading internal spec issues :laughing:

PR #8094, opened 2024-06-05
sync: Use vector to return last batches

artem-lunarg:
Started this as cleanup to remove specialized templates that generalize some helpers. Inline code directly where it's needed - results in simpler code and it's less code.
Adds alias names for smart pointers. 
Replaces map with a vector to store per-queue last batch list, which usually is few elements. Run Doom/cs2 performance traces.


spencer-lunarg (approved):

spencer-lunarg:
Internal CI is failing from Android issues, ran locally on android and had no issues

PR #8083, opened 2024-06-03
Use ALooper_pollOnce instead of ALooper_pollAll

jrprice:
The latter is deprecated and will be removed in Android 34.

PR #8097, opened 2024-06-05
Revert "Use ALooper_pollOnce instead of ALooper_pollAll"

spencer-lunarg:
Reverts KhronosGroup/Vulkan-ValidationLayers#8083

Internal CI was never ran and I broke CI :cry:


jeremyg-lunarg (approved):
LGTM once the commit message is happy

PR #8099, opened 2024-06-06
layers: Add maxFragmentOutputAttachments

spencer-lunarg:
We never were checking for `maxFragmentOutputAttachments`, but we do now

### DIFF
--- a/tests/unit/shader_interface.cpp
+++ b/tests/unit/shader_interface.cpp
@@ -1377,7 +1377,7 @@ TEST_F(NegativeShaderInterface, AlphaToCoverageArrayIndex) {
 
     char const *fsSource = R"glsl(
         #version 450
-        layout(location=1) out vec4 fragData[4];
+        layout(location=1) out vec4 fragData[3];
         void main() {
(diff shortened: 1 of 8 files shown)


tests/unit/shader_interface.cpp
```diff
@@ -1490,6 +1490,10 @@ TEST_F(NegativeShaderInterface, MultidimensionalArray64bit) {
     RETURN_IF_SKIP(Init());
     InitRenderTarget();
 
+    if (m_device->phy().limits_.maxFragmentOutputAttachments < 25) {
```
artem-lunarg, line 1493: Do we have a configuration we this does not skip?

artem-lunarg (approved):
LGTM. One question about test.

tests/unit/shader_interface.cpp
spencer-lunarg: not yet... i'll explain offline

spencer-lunarg: (at our next weekly meeting)

PR #8102, opened 2024-06-06
gpu: Refacto

arno-lunarg:
Redo GPU resources management:
-> Get rid of CommandResources classes
-> Now error logic and needed resources are stored in lambdas, no need to create a new class and modify many files (per_command_error_loggers in gpuav::CommandBuffer) 
-> Add GpuResourcesManager and SharedResourcesManager to help with GPU resources lifetime management

Next thing to refacto would be error messages logic for validation done by instrumenting shaders


layers/CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ target_sources(VkLayer_utils PRIVATE
     error_message/error_strings.h
     error_message/record_object.h
     external/xxhash.h
+    external/inplace_function.h
```
spencer-lunarg, line 33: missing in `BUILD.gn`

arno-lunarg: argh

layers/gpu/cmd_validation/gpuav_copy_buffer_to_image.cpp
```diff
@@ -18,62 +18,167 @@
...
-    CommandResources cmd_resources;
-    cmd_resources.command = loc.function;
+    auto &shared_copy_validation_resources = gpuav.shared_resources_manager.Get<SharedCopyBufferToImageValidationResources>(
+        gpuav, cb_state->GetValidationCmdCommonDescriptorSetLayout(), loc);
 
-    auto copy_buffer_to_img_resources = std::make_unique<PreCopyBufferToImageResources>();
-    CommandResources &base = *copy_buffer_to_img_resources;
-    base = cmd_resources;
-    copy_buffer_to_img_resources->src_buffer = copy_buffer_to_img_info->srcBuffer;
+    assert(shared_copy_validation_resources.IsValid());
```
spencer-lunarg, line 172: great use of the new `ASSERT_AND_RETURN` macro here

arno-lunarg: (〜￣▽￣)〜

arno-lunarg: honestly I do not like ASSERT_AND_RETURN, since the condition is not in the assert it will not appear in the error message

layers/gpu/resources/gpuav_subclasses.h
```diff
@@ -122,6 +125,11 @@ class CommandBuffer : public gpu_tracker::CommandBuffer {
     void Destroy() final;
     void Reset() final;
 
+    gpu::GpuResourcesManager gpu_resources_manager;
+    using ErrorLoggerFunc =
```
spencer-lunarg, line 129: So I know what this is because we talked about it, I guess worth adding a small comment here explaining to the future dev "why are we using this implace_function library)

layers/gpu/resources/gpuav_subclasses.h
```diff
@@ -71,11 +72,13 @@ struct CmdIndirectState {
 class CommandBuffer : public gpu_tracker::CommandBuffer {
   public:
     // per validated command state
-    std::vector<std::unique_ptr<CommandResources>> per_command_resources;
+    // std::vector<std::unique_ptr<CommandResources>> per_command_resources;
```
spencer-lunarg, line 75: remove?

layers/gpu/cmd_validation/gpuav_copy_buffer_to_image.cpp
spencer-lunarg: We can add `__LINE__` and `__FILE__` to it

as I have found from my assert in `ReplaceLinkedId` no one seems to read the asserts in details anyway and is just using a debugger

spencer-lunarg (approved):
small nits, but otherwise LGTM

layers/gpu/cmd_validation/gpuav_copy_buffer_to_image.cpp
arno-lunarg: I was just reading one in the CI logs for this PR, actually one for those new `assert(shared_resources.IsValid())`.
I may be missing something here, what problem does `ASSERT_AND_RETURN` solves?

layers/gpu/resources/gpuav_subclasses.h
arno-lunarg: ok!

layers/gpu/resources/gpuav_subclasses.h
arno-lunarg: yes

layers/gpu/cmd_validation/gpuav_copy_buffer_to_image.cpp
spencer-lunarg: we have a LOT of early returns all over, I never know if they are because 
1. it is a valid situation that we got there 
2. if it should NEVER happen, and we only early return to not crash incase we get there

We also had a lot of times were we just did `assert()` and crashed when it hit it in a release build in the wild

I wanted a clear way to say "We should not get here, so please assert, but if in a release build, don't just crash here and return/continue on"

What you have here is find, its logically the same thing, I just wrapped it in a macro for ease of adding it in a hundred spots

artem-lunarg: And my comment too! The fact that we have assert-and-return in many places suggests something should be fixed at some point of time. Assert is fine in localized parts of code to express uncertainty or complexity of the code but if large part of the system if covered with assert this could the reason to solve this ultimately, (proof behavior of some code, centralize some functionality or categorize cases). To be honest (after few discussions) I still don't have good intuition of the reason of null object queries, there were probably different answers to that question, which probably is a valid situation for asserts. But I like to look at this as temporary solution, ideally the need for assert-and-return should be rare enough not to have separate helper for it, scattering many asserts usually suggests about some fragility (maybe Jeremy's idea about splitting validation objects solves this).

spencer-lunarg: adding to weekly meeting discussion, we are not going to solve this here, and what you have here @arno-lunarg is fine

PR #8105, opened 2024-06-07
Re-enable DISABLED tests

spencer-lunarg:
Two commits

1. We had off-by-one error for `vkCmdClearAttachments`
2. I went through the `DISABLED` tests and the ones without an issue (and still open) I tried to re-enable


tests/unit/object_lifetime.cpp
```diff
@@ -20,25 +20,6 @@
 #include "../framework/descriptor_helper.h"
 #include "../framework/external_memory_sync.h"
 
-// Validation of dispatchable handles is not performed until VVL's chassis will be
-// able to do this validation (if ever) instead of crashing (which is also an option).
-// If vulkan's loader trampoline is active, then it's also the place where invalid
-// dispatchable handle can cause a crash.
-TEST_F(NegativeObjectLifetime, DISABLED_CreateBufferUsingInvalidDevice) {
```
spencer-lunarg, line 27: this will never be test-able in validation, the loader now provide an error message, but this doesn't belong in VVL anymore

tests/unit/query.cpp
```diff
@@ -2004,121 +2004,6 @@ TEST_F(NegativeQuery, CommandBufferInheritanceFlags) {
     m_errorMonitor->VerifyFound();
 }
 
-// Doesn't seem like these VUs are suppose to be in the spec
-// https://gitlab.khronos.org/vulkan/vulkan/-/issues/3733
-TEST_F(NegativeQuery, DISABLED_MultiviewEndQuery) {
```
spencer-lunarg, line 2009: tldr, this is being validated elsewhere and this test is not needed anymore

tests/unit/buffer_positive.cpp
```diff
@@ -87,7 +87,7 @@ TEST_F(PositiveBuffer, TexelBufferAlignmentIn13) {
     CreateBufferViewTest(*this, &buff_view_ci, {});
 }
 
-TEST_F(PositiveBuffer, DISABLED_PerfGetBufferAddressWorstCase) {
```
arno-lunarg, line 90: Please do not re-enable those tests, they are meant to be used locally only, they date back to an issue a client had, long story...

arno-lunarg (changes requested):

tests/unit/buffer_positive.cpp
spencer-lunarg: > I should have added a comment!!!

tell me what to write here and I'll add it

arno-lunarg:
```
// The two PerfGetBufferAddress tests are intended to be used locally to monitor performance of the internal address -> buffer map
```

spencer-lunarg: `// The two PerfGetBufferAddress tests are intended to be used locally to monitor performance of the internal address -> buffer map` #first

layers/core_checks/cc_image.cpp
```diff
@@ -1215,6 +1216,16 @@ bool CoreChecks::PreCallValidateCmdClearAttachments(VkCommandBuffer commandBuffe
             }
         }
 
+        auto describe_color_count = [is_dynamic_rendering, &cb_state]() {
```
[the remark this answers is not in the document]
spencer-lunarg: ya, I thought I was gonna use it elsewhere, but no

arno-lunarg (approved):

PR #8114, opened 2024-06-09
build: added missing -soname link option for Android builds

Izhido:


Izhido:
The format of the message of the commit seems to be wrong. Closing and retrying.

PR #8115, opened 2024-06-09
build: added missing -soname link option for Android builds

Izhido:


Izhido:
Once again, the commit message is wrong. trying again.

PR #8116, opened 2024-06-09
build: Added missing -soname link option for Android builds

Izhido:


CMakeLists.txt
```diff
@@ -122,6 +122,7 @@ if(${CMAKE_CXX_COMPILER_ID} MATCHES "(GNU|Clang)")
         -Wextra
         -Wpointer-arith
     )
+    add_link_options(-Wl,-soname=libVkLayer_khronos_validation.so)
```
spencer-lunarg, line 125: So this is failing for the normal Linux build

so in `layers/CMakeLists.txt` we have `target_link_options` calls, I think this needs to be

```
if(ANDROID)
    target_link_options(vvl PRIVATE -Wl,-soname=libVkLayer_khronos_validation.so))
endif()
```

Izhido:
As this was superseded by the latest changes by @spencer-lunarg , this is no longer required. Closing.

Issue #8084, opened 2024-06-03
Syncval SSBO/UBO regression


ShabbyX:
Since dc5d54248 (@spencer-lunarg), we (ANGLE) see a new syncval error in a few traces like this:

```
[ SYNC-HAZARD-WRITE-AFTER-READ ] Validation Error: [ SYNC-HAZARD-WRITE-AFTER-READ ] Object 0: handle = 0x1dd00000001dd, type = VK_OBJECT_TYPE_BUFFER; | MessageID = 0x376bc9df | vkCmdDispatch():  Hazard WRITE_AFTER_READ for VkBuffer 0x1dd00000001dd[] in VkCommandBuffer 0x5649ae338100[], VkPipeline 0xa3e0000000a3e[], and VkDescriptorSet 0xb7c0000000b7c[], type: VK_DESCRIPTOR_TYPE_STORAGE_BUFFER, binding #4 index 0. Access info (usage: SYNC_COMPUTE_SHADER_SHADER_STORAGE_WRITE, prior_usage: SYNC_COMPUTE_SHADER_UNIFORM_READ, read_barriers: VkPipelineStageFlags2(0), command: vkCmdDispatch, seq_no: 25, reset_no: 9)
```

Looking at one of the traces (`warcraft_rumble`, if you have access), this looks like a tracking bug. Like, one of the buffers for which this error is generated is used as such:

![image](https://github.com/KhronosGroup/Vulkan-ValidationLayers/assets/1067408/604306bf-b7d0-44a7-b8a1-b0f9ec9a83bb)

EID 83: 
![image](https://github.com/KhronosGroup/Vulkan-ValidationLayers/assets/1067408/ef3beb82-a33e-4286-b8ea-fe82f3524b35)

EID 158: 
![image](https://github.com/KhronosGroup/Vulkan-ValidationLayers/assets/1067408/d8edb187-2aac-488f-905f-c55f8380bc17)

EID 183: 
![image](https://github.com/KhronosGroup/Vulkan-ValidationLayers/assets/1067408/f09edf7d-f01f-4d48-aa03-4f28db310739)

As you can see, there is no overlap in the ranges of the buffer used in these draw calls, but EID 183 produces the following error in RenderDoc:

```
[...]

PR #8117, opened 2024-06-09
sync: Apply descriptor dynamic offset

artem-lunarg:
Addresses https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8084 (at least non-overlapping accesses part)

### DIFF
--- a/layers/state_tracker/descriptor_sets.cpp
+++ b/layers/state_tracker/descriptor_sets.cpp
@@ -519,6 +519,21 @@ void vvl::DescriptorSet::NotifyInvalidate(const NodeList &invalid_nodes, bool un
     }
 }
 
+uint32_t vvl::DescriptorSet::GetDynamicOffsetIndexFromBinding(uint32_t dynamic_binding) const {
+    const uint32_t index = layout_->GetIndexFromBinding(dynamic_binding);
+    if (index == bindings_.size()) {  // binding not found
+        return vvl::kU32Max;
+    }
+    assert(IsDynamicDescriptor(bindings_[index]->type));
+    uint32_t dynamic_offset_index = 0;
+    for (uint32_t i = 0; i < index; i++) {
+        if (IsDynamicDescriptor(bindings_[i]->type)) {
+            dynamic_offset_index += bindings_[i]->count;
+        }
+    }
+    return dynamic_offset_index;
+}
+
 void vvl::DescriptorSet::Destroy() {
     for (auto &binding : bindings_) {
         binding->RemoveParent(this);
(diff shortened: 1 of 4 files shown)


layers/state_tracker/descriptor_sets.cpp
```diff
@@ -519,6 +519,21 @@ void vvl::DescriptorSet::NotifyInvalidate(const NodeList &invalid_nodes, bool un
     }
 }
 
+uint32_t vvl::DescriptorSet::GetDynamicOffsetIndexFromBinding(uint32_t dynamic_binding) const {
+    const uint32_t index = layout_->GetIndexFromBinding(dynamic_binding);
+    if (index == bindings_.size()) {  // binding not found
+        return vvl::kU32Max;
+    }
+    assert(IsDynamicDescriptor(bindings_[index]->type));
```
spencer-lunarg, line 527: if we are only adding the index from `[0, index - 1]` below, why does this need to be true?

artem-lunarg: `uint32_t GetDynamicOffsetIndexFromBinding(uint32_t dynamic_binding)` is defined only for dynamic descriptor bindings. Non-dynamic binding is not associated with any dynamic offset. Added a comment.

tests/unit/sync_val_positive.cpp
```diff
@@ -1534,3 +1534,113 @@ TEST_F(PositiveSyncVal, SignalUnsignalSignalSingleSubmit) {
...
+    RETURN_IF_SKIP(InitState());
+
+    // 32 bytes
+    const VkDeviceSize uniform_data_size = 8 * sizeof(uint32_t);
+    // 128 bytes or more (depending on minUniformBufferOffsetAlignment)
+    const VkDeviceSize copy_dst_area_size = std::max((VkDeviceSize)128, m_device->phy().limits_.minUniformBufferOffsetAlignment);
+    // 160 bytes or more (depending on minUniformBufferOffsetAlignment)
+    const VkDeviceSize size = copy_dst_area_size + uniform_data_size;
+
+    // At first we have at least 128 bytes copy destinatin region, followed by 32 bytes of uniform data.
```
arno-lunarg, line 1550:
```suggestion
    // At first we have at least 128 bytes copy destination region, followed by 32 bytes of uniform data.
```

arno-lunarg (approved):

tests/unit/sync_val_positive.cpp
artem-lunarg: Updated

spencer-lunarg (approved):

PR #8119, opened 2024-06-10
gpu: Call InternalError in GetOutputState

arno-lunarg:
Also cleanup some error messages

### DIFF
--- a/layers/gpu/cmd_validation/gpuav_draw.cpp
+++ b/layers/gpu/cmd_validation/gpuav_draw.cpp
@@ -87,7 +87,7 @@ struct SharedDrawValidationResources final {
             shader_ci.pPushConstantRanges = &push_constant_range;
             result = DispatchCreateShadersEXT(device, 1u, &shader_ci, nullptr, &shader_object);
             if (result != VK_SUCCESS) {
-                gpuav.InternalError(device, loc, "Unable to create shader object. Aborting GPU-AV");
+                gpuav.InternalError(device, loc, "Unable to create shader object. Aborting GPU-AV.");
                 return;
             }
         } else {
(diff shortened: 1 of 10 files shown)


layers/gpu/descriptor_validation/gpuav_descriptor_set.cpp
```diff
@@ -412,10 +412,13 @@ std::shared_ptr<DescriptorSet::State> DescriptorSet::GetOutputState() {
     alloc_info.requiredFlags = VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_CACHED_BIT;
     VkResult result =
         vmaCreateBuffer(next_state->allocator, &buffer_info, &alloc_info, &next_state->buffer, &next_state->allocation, nullptr);
+    assert(result == VK_SUCCESS);
```
artem-lunarg, line 415: I can see this assert is how it's done in other places, but failure to allocate memory on gpu sounds like a realistic situation, if happens should not be considered a bug in the code, more like a request to buy better gpus for CI.

artem-lunarg (approved):

layers/gpu/cmd_validation/gpuav_draw.cpp
```diff
@@ -201,7 +201,7 @@ void InsertIndirectDrawValidation(Validator &gpuav, const Location &loc, VkComma
 
     auto cb_state = gpuav.GetWrite<CommandBuffer>(cmd_buffer);
     if (!cb_state) {
-        gpuav.InternalError(cmd_buffer, loc, "Unrecognized command buffer");
+        gpuav.InternalError(cmd_buffer, loc, "Unrecognized command buffer. Aborting GPU-AV.");
```
spencer-lunarg, line 204: we should just put the `Aborting GPU-AV` substring inside the `InternalError` call

arno-lunarg: Yes

PR #8123, opened 2024-06-10
sync: Turn QueueBatchContext cmdbuf state into a local variable

artem-lunarg:
We have a set of state in `QueueBatchContext` that assists during validation period and then should not be used. Trying to make it local when possible.


artem-lunarg:
Lord of determinism please make it pass this time

spencer-lunarg (approved):

Issue #7385, opened 2024-01-26
Version 1.3.268 does not load in app for which 1.3.236 loads correctly (Android 12)


Izhido:
**Environment:**
 - OS: Android 12 (Vulkan 1.1.128)
 - GPU and driver version: Adreno 740, Driver 512.746.0

**Describe the Issue**

Loading version 1.3.268 of the validation layer lib causes an java.lang.UnsatisfiedLinkError at runtime. Latest known version that works, with unchanged settings, is 1.3.236.

**Expected behavior**

**Valid Usage ID**

**Additional context**

<details>
<summary>The reported error is as follows:</summary>

```sh
E  FATAL EXCEPTION: main
Process: com.my_namespace.my_app, PID: 16821
java.lang.UnsatisfiedLinkError: Unable to load native library "/data/app/~~eBT-ClU8d8rElTi09Pz4Rw==/com.my_namespace.my_app-9oOtV6cikWELznRSV6vYYA==/base.apk!/lib/arm64-v8a/libmy_app_native.so": dlopen failed: library "D:/Projects/MyApp/app/src/main/cpp/../../../../android-binaries-vulkan-sdk-1.3.268.0/arm64-v8a/libVkLayer_khronos_validation.so" not found: needed by /data/app/~~eBT-ClU8d8rElTi09Pz4Rw==/com.my_namespace.my_app-9oOtV6cikWELznRSV6vYYA==/base.apk!/lib/arm64-v8a/libmy_app_native.so in namespace classloader-namespace
	at android.app.NativeActivity.onCreate(NativeActivity.java:178)
	at android.app.Activity.performCreate(Activity.java:8057)
	at android.app.Activity.performCreate(Activity.java:8037)
	at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1341)
	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3705)
	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3881)
	at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:103)
	at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
	at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2270)
	at android.os.Handler.dispatchMessage(Handler.java:106)
[...]

PR #8124, opened 2024-06-10
build: List soname for Android

spencer-lunarg:
for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7385 - v2 of https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8116

@Izhido can you confirm this works for you, works for me locally

### DIFF
--- a/layers/CMakeLists.txt
+++ b/layers/CMakeLists.txt
@@ -392,6 +392,11 @@ else()
     target_link_options(vvl PRIVATE LINKER:--version-script=${CMAKE_CURRENT_SOURCE_DIR}/${LAYER_NAME}.map)
 endif()
 
+if(ANDROID)
+    # See https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7385#issuecomment-2155788105
+    target_link_options(vvl PRIVATE -Wl,-soname=libVkLayer_khronos_validation.so)
+endif()
+
 target_compile_options(vvl PRIVATE "$<IF:$<CXX_COMPILER_ID:MSVC>,/wd4100,-Wno-unused-parameter>")
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")


Izhido:
> for #7385 - v2 of #8116
> 
> @Izhido can you confirm this works for you, works for me locally

Yes. I can confirm now this change works.

layers/CMakeLists.txt
```diff
@@ -392,6 +392,11 @@ else()
     target_link_options(vvl PRIVATE LINKER:--version-script=${CMAKE_CURRENT_SOURCE_DIR}/${LAYER_NAME}.map)
 endif()
 
+if(ANDROID)
+    # See https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7385#issuecomment-2155788105
+    target_link_options(vvl PRIVATE -Wl,-soname=libVkLayer_khronos_validation.so)
```
juan-kitware, line 397: Why is this needed? CI currently works fine without `-soname` being specified. Under what exact circumstances is this actually needed?

CMake deliberately omits -soname for MODULE libraries from what I can tell.

I'd like to understand what is breaking because -soname is not being added.

spencer-lunarg: https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7385#issuecomment-2155788105

I will check tomorrow, but curious if the newest NDK breaks something (I have 25.1.8937393 locally and the issue was using 26.3)

DanAlbert: > CMake deliberately omits -soname for MODULE libraries from what I can tell.

Which https://cmake.org/cmake/help/latest/command/add_library.html says is something only loadable by `dlopen`, not by `-l`:

> A plugin that may not be linked by other targets, but may be dynamically loaded at runtime using dlopen-like functionality.

That's how the validation layers work though, right? It looks like the @lzhido linked directly to the validation layers? Don't do that, just add them to the `jniLibs` directory of your project: https://developer.android.com/ndk/guides/graphics/validation-layer#prebuilt-binaries. Your C++ build should be completely ignorant of their existence; they only need to be packaged.

spencer-lunarg:
closing from https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7385#issuecomment-2259145247

PR #8127, opened 2024-06-10
layers: Add more details about VariableBase access getters

spencer-lunarg:
per @artem-lunarg request, hopefully this sums up all the "fun" around SPIR-V variable accesses

### DIFF
--- a/layers/state_tracker/shader_module.h
+++ b/layers/state_tracker/shader_module.h
@@ -193,8 +193,6 @@ const uint32_t read = 1 << 0;
 const uint32_t write = 1 << 1;
 const uint32_t atomic_read = 1 << 2;
 const uint32_t atomic_write = 1 << 3;
-// For Images, we might only want to know if the read/write came from an image operation
-// (a storage image is always "loaded" before it is written, but that is not a image read, just a memory read)
 const uint32_t image_read = 1 << 4;
 const uint32_t image_write = 1 << 5;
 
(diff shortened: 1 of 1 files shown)


layers/state_tracker/shader_module.h
```diff
@@ -306,10 +304,24 @@ struct VariableBase {
...
+    // -------
+    // Accessed == (read | write | atomic)
+    // It is possible to have descriptors/inout/push constant/etc declared but never used
     bool IsAccessed() const { return access_mask != AccessBit::empty; }
+    // Atomics are really both a read/write, but some HW doesn't support atomic on all int/float bit-width
     bool IsAtomic() const { return access_mask & AccessBit::atomic_mask; }
+    // Read/Write here refer to the variable itself. For a buffer this means the memory has been accessed. For an Image, this only
+    // means the descriptor itself has been accessed
     bool IsReadFrom() const { return access_mask & AccessBit::read_mask; }
     bool IsWrittenTo() const { return access_mask & AccessBit::write_mask; }
```
artem-lunarg, line 319: Is there a setup with IsWrittenTo == true for the image?

layers/state_tracker/shader_module.h
```diff
@@ -306,10 +304,24 @@ struct VariableBase {
...
+    // Images are special and will first have a read/write to the descriptor, then an Image Operation to the image memory itself.
+    // - Some operations such as ImageSize() will read data from the descriptor and never actually read the image memory (it would
+    // return false for IsImageReadFrom()).
+    // - A storage image is always "loaded" before it is written, but it will only return true for IsImageReadFrom() if the image
```
artem-lunarg, line 323: "A storage image is always "loaded" before it is written"
Where is this from? Let's say I have a storage image and write directly to specific texel location, can't see how this can cause a read.

spencer-lunarg: There is an `OpLoad` to the descriptor, so therefor the descriptor was read, but the memory was written

Again, the issue is from this point in SPIR-V, I can't tell if this OpLoad is touching a buffer's memory or the image descriptor, so I have to mark the `AccessBit::read`

layers/state_tracker/shader_module.h
spencer-lunarg: I don't think there is a way for `IsWrittenTo()` to be true if it is an image, there are enough restrictions in SPIR-V

artem-lunarg: Interesting, thanks for confirmation.

layers/state_tracker/shader_module.h
spencer-lunarg: There is an `OpLoad` to the descriptor, so therefor the descriptor was read, but the image memory was not written

Again, the issue is from this point in SPIR-V, I can't tell if this OpLoad is touching a buffer's memory or the image descriptor, so I have to mark the `AccessBit::read`v

artem-lunarg: Thanks for explanation, that's makes sense. I'm not sure if there are scenarios on the client side (I guess that's syncval at the moment, but maybe some shader code too?) that need to know about access to a descriptor instead of the resource memory. If that's the case, the ideal interface for me is where all getters report only accesses to resource memory, so you don't need to think about descriptors at all, why to filter out something that you don't work with. Maybe I'm mistaken and descriptor access are check by some code or maybe that's not in scope of spirv.

artem-lunarg (approved):
LGTM. Thanks for documenting this. Already clarified some details.

layers/state_tracker/shader_module.h
spencer-lunarg: example ` VUID-vkCmdDraw-None-08600` - If you have an image in your shader, but your descriptor in your pipeline layout is a buffer, that is an error, only if the descriptor is statically used, so if the never touch it, we ignored it, but if they do an `imageSize` it has not been accessed and "statically used"

So the issue is you (sync val) clearly only cares about touching the memory, but some validation cares about any access because if a descriptor is accessed at all, the driver's compiler must care about that in its shader interface with the driver

PR #8131, opened 2024-06-11
layers: Fix subpass contents check

spencer-lunarg:
fix from https://github.com/KhronosGroup/Vulkan-Docs/issues/2372 that was in 1.3.287 spec

`VUID-vkCmdExecuteCommands-contents-06018` was reworked and split into

- VUID-vkCmdExecuteCommands-contents-09680
- VUID-vkCmdExecuteCommands-None-09681

### DIFF
--- a/layers/core_checks/cc_cmd_buffer.cpp
+++ b/layers/core_checks/cc_cmd_buffer.cpp
@@ -841,15 +841,27 @@ bool CoreChecks::PreCallValidateCmdExecuteCommands(VkCommandBuffer commandBuffer
     }
 
     if (cb_state.activeRenderPass) {
-        if (!cb_state.activeRenderPass->UsesDynamicRendering() && cb_state.IsPrimary() &&
-            (cb_state.activeSubpassContents != VK_SUBPASS_CONTENTS_SECONDARY_COMMAND_BUFFERS &&
-             cb_state.activeSubpassContents != VK_SUBPASS_CONTENTS_INLINE_AND_SECONDARY_COMMAND_BUFFERS_EXT)) {
-            const LogObjectList objlist(commandBuffer, cb_state.activeRenderPass->Handle());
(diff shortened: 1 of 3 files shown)


layers/core_checks/cc_cmd_buffer.cpp
```diff
@@ -841,15 +841,27 @@ bool CoreChecks::PreCallValidateCmdExecuteCommands(VkCommandBuffer commandBuffer
...
+                                     "when calling vkCmdExecuteCommands() within the first subpass.");
+                }
+            } else {
+                if (cb_state.activeSubpassContents != VK_SUBPASS_CONTENTS_SECONDARY_COMMAND_BUFFERS &&
+                    cb_state.activeSubpassContents != VK_SUBPASS_CONTENTS_INLINE_AND_SECONDARY_COMMAND_BUFFERS_EXT) {
+                    const LogObjectList objlist(commandBuffer, cb_state.activeRenderPass->Handle());
+                    skip |= LogError("VUID-vkCmdExecuteCommands-None-09681", objlist, error_obj.location,
+                                     "contents must be set to VK_SUBPASS_CONTENTS_SECONDARY_COMMAND_BUFFERS or "
+                                     "VK_SUBPASS_CONTENTS_INLINE_AND_SECONDARY_COMMAND_BUFFERS_EXT"
+                                     "when calling vkCmdExecuteCommands() within the non-first subpass (currently subpass %" PRIu32
```
gamagan, line 860: For the first subpass, it is correct to write `...within the first subpass.`, like it is in the code, because there is only one first subpass in a RenderPass. However, since there can be more than one non-first subpass, the correct article to use is `a`, not `the`. Thus:
`...within a non-first subpass...` instead of `...within the non-first subpass...`

spencer-lunarg: good grammar catch!

artem-lunarg (approved):

PR #8133, opened 2024-06-11
layers: Move logic to stateless

spencer-lunarg:
Moves some logic from core checks to stateless


artem-lunarg (approved):
LGTM. Lets add meeting item how to decide when to move things to stateless.

PR #8135, opened 2024-06-11
layers: Create combined dynamic state draw check

spencer-lunarg:
This change is from https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6345 (and there are more coming, just doing 3 to start)

The overall goal is to combine all the drawtime dynamic state checks from Shader Object and Pipeline into a single check

### DIFF
--- a/layers/vulkan/generated/dynamic_state_helper.cpp
+++ b/layers/vulkan/generated/dynamic_state_helper.cpp
@@ -22,6 +22,7 @@
 // NOLINTBEGIN
 
 #include "core_checks/core_validation.h"
+#include "state_tracker/pipeline_state.h"
 
 VkDynamicState ConvertToDynamicState(CBDynamicState dynamic_state) {
     switch (dynamic_state) {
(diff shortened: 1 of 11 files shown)


layers/vulkan/generated/dynamic_state_helper.cpp
```diff
@@ -593,4 +594,54 @@ std::string DescribeDynamicStateCommand(CBDynamicState dynamic_state) {
     return ss.str();
 }
 
+std::string DescribeDynamicStateDependency(CBDynamicState dynamic_state, const vvl::Pipeline* pipeline) {
```
spencer-lunarg, line 597: this is generated, and the goal is to explain "where" the dependency came from, so example of 2 error messages that could come from there

```
vkCmdDraw():  VK_DYNAMIC_STATE_DEPTH_COMPARE_OP state is dynamic, but the command buffer never called vkCmdSetDepthCompareOp.
vkCmdSetRasterizerDiscardEnable last set rasterizerDiscardEnable to VK_FALSE.
vkCmdSetDepthTestEnable last set depthTestEnable to VK_TRUE.

vkCmdDraw():  VK_DYNAMIC_STATE_DEPTH_COMPARE_OP state is dynamic, but the command buffer never called vkCmdSetDepthCompareOp.
VkPipelineRasterizationStateCreateInfo::rasterizerDiscardEnable was VK_FALSE in the last bound graphics pipeline.
VkPipelineDepthStencilStateCreateInfo::depthTestEnable was VK_TRUE in the last bound graphics pipeline.
```

artem-lunarg (approved):

PR #8137, opened 2024-06-11
sync: Make batch state local and devirtualization

artem-lunarg:
More changes to reduce state and simplify analysis.
They are good to have on its own, but that's mostly the parts that will interact with timeline pipeline.

### DIFF
--- a/layers/sync/sync_commandbuffer.h
+++ b/layers/sync/sync_commandbuffer.h
@@ -215,11 +215,7 @@ class CommandExecutionContext : public SyncValidationInfo {
     virtual const SyncEventsContext *GetCurrentEventsContext() const = 0;
     virtual QueueId GetQueueId() const = 0;
 
-    ResourceUsageRange ImportRecordedAccessLog(const CommandBufferAccessContext &recorded_context);
-
-    virtual ResourceUsageTag GetTagLimit() const = 0;
     virtual VulkanTypedHandle Handle() const = 0;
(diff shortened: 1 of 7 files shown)


layers/sync/sync_commandbuffer.h
```diff
@@ -215,11 +215,7 @@ class CommandExecutionContext : public SyncValidationInfo {
     virtual const SyncEventsContext *GetCurrentEventsContext() const = 0;
     virtual QueueId GetQueueId() const = 0;
 
-    ResourceUsageRange ImportRecordedAccessLog(const CommandBufferAccessContext &recorded_context);
-
-    virtual ResourceUsageTag GetTagLimit() const = 0;
     virtual VulkanTypedHandle Handle() const = 0;
-    virtual void InsertRecordedAccessLogEntries(const CommandBufferAccessContext &cb_context) = 0;
```
artem-lunarg, line 222: and this too

layers/sync/sync_submit.h
```diff
@@ -305,7 +297,6 @@ class QueueBatchContext : public CommandExecutionContext, public std::enable_sha
     std::vector<ResourceUsageTag> queue_sync_tag_;
 
     // Clear these after validation and import, not valid after.
-    BatchAccessLog::BatchRecord batch_;  // Holds the cumulative tag bias, and command buffer counts for Import support.
```
artem-lunarg, line 308: and this field with temporary lifetime become a local variable. The entire block in the end might gone eventually.

spencer-lunarg (approved):

Issue #8138, opened 2024-06-12
First vkGetSwapchainImagesKHR false positive


Goshido:
**Environment:**
 - OS: _Android 11_
 - GPU and driver version: _GPU_: _Mali G76_, driver: _26.0.0_
 - SDK or header version if building from repo: _1.3.287_, _SHA-1_ `15ba84ba94ff2ed473474deba2b4f2af4705b1fb`
 - Options enabled (synchronization, best practices, etc.): `VK_VALIDATION_FEATURE_ENABLE_BEST_PRACTICES_EXT` and `VK_VALIDATION_FEATURE_ENABLE_SYNCHRONIZATION_VALIDATION_EXT`

**Describe the Issue**

I've just updated to latest _VVL_ and got best practices validation warning which seems odd:

```txt
severity: Warning
type: Validation
message ID name: BestPractices-SwapchainPriorCount
message ID: 0xEE8A3A07
queues: N/A
command buffers: N/A
objects: N/A
message: Validation Warning: [ BestPractices-SwapchainPriorCount ] | MessageID = 0xee8a3a07 |
    vkGetSwapchainImagesKHR():  called with non-NULL pSwapchainImageCount; but no prior positive value has
    been seen for pSwapchainImages.
```

it's literally first call to ask the presentation engine how many swapchain images it has.

What I gonna do? I'm so confused. Would you kindly help?

**Expected behavior**

No validation warnings for first call `vkGetSwapchainImagesKHR`.

**Valid Usage ID**
_N/A_

**Additional context**

Previous compatible _VVL_ version: _1.3.285_, `df717cc614855e2b81660798da349043bb37472a`

spencer-lunarg:
> called with non-NULL pSwapchainImageCount

looking at the logic, the code is not checking that, seems to be a clear bug, will get a fix

spencer-lunarg:
actually, I argue this check is something that would best be generated as this is for ANY call that requires a "query to get the count from the driver"

Goshido:
The issue has been fixed. 👍

PR #8142, opened 2024-06-12
bp: Remove incorrect check to WSI query

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8138

If we want to have a Best Practice to checks that the app called the query function once before, that is something that could be generated as there are many other spots this occur


layers/best_practices/bp_wsi.cpp
```diff
@@ -73,19 +73,11 @@ bool BestPractices::PreCallValidateGetSwapchainImagesKHR(VkDevice device, VkSwap
...
-                           "been seen for pSwapchainImages.");
-    }
-
-    if (*pSwapchainImageCount > swapchain_state->get_swapchain_image_count) {
-        skip |= LogWarning(kVUID_BestPractices_Swapchain_InvalidCount, device, error_obj.location,
-                           "called with non-NULL pSwapchainImages, and with pSwapchainImageCount set to a "
-                           "value (%" PRId32 ") that is greater than the value (%" PRId32
-                           ") that was returned when pSwapchainImages was NULL.",
-                           *pSwapchainImageCount, swapchain_state->get_swapchain_image_count);
+    if (pSwapchainImageCount && *pSwapchainImageCount > swapchain_state->get_swapchain_image_count) {
```
artem-lunarg, line 76: Do we need the second check? Can't see why it's correct. If that's the first call that returns count number, the memory pointed by `pSwapchainImageCount` can contain arbitrary value.

spencer-lunarg: Happy to remove it as well, these seem old and has no tests either

artem-lunarg (approved):

PR #8144, opened 2024-06-13
layers: Add missing parent VUs

spencer-lunarg:
VUID will be added here https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6711

This is adding some (more to come later, will take time) missing checks where we aren't checking if the `VkDevice` is the same between 2 handles used in a command (this has been discussed before in https://gitlab.khronos.org/vulkan/vulkan/-/issues/3553)

### DIFF
--- a/layers/vulkan/generated/object_tracker.cpp
+++ b/layers/vulkan/generated/object_tracker.cpp
@@ -386,7 +386,8 @@ bool ObjectLifetimes::PreCallValidateFlushMappedMemoryRanges(VkDevice device, ui
         for (uint32_t index0 = 0; index0 < memoryRangeCount; ++index0) {
             [[maybe_unused]] const Location index0_loc = error_obj.location.dot(Field::pMemoryRanges, index0);
             skip |= ValidateObject(pMemoryRanges[index0].memory, kVulkanObjectTypeDeviceMemory, false,
-                                   "VUID-VkMappedMemoryRange-memory-parameter", kVUIDUndefined, index0_loc.dot(Field::memory));
+                                   "VUID-VkMappedMemoryRange-memory-parameter",
+                                   "UNASSIGNED-vkFlushMappedMemoryRanges-memory-device", index0_loc.dot(Field::memory));
         }
     }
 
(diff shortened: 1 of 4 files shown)


scripts/generators/object_tracker_generator.py
```diff
@@ -209,20 +209,14 @@ def __init__(self,
         # Structures that do not define parent/commonparent VUIDs for vulkan handles.
         # This overlaps with https://gitlab.khronos.org/vulkan/vulkan/-/issues/3553#note_424431
         self.structs_that_forgot_about_parent_vuids = [
-            'VkMappedMemoryRange',
```
spencer-lunarg, line 212: the goal is to remove these so this list is minimal as possible

scripts/generators/object_tracker_generator.py
```diff
@@ -623,29 +620,75 @@ def hasParameterParentVUID(self, parameter: Member, commandName: str) -> bool:
         # For other dispatchable handles it depends on the API function
         return commandName in self.dispatchable_has_parent_vuid_commands
 
+    # It is very complex for the spec handle walking through structs and finding Handles and generating implicit VUs,
+    # We instead just have to do it manually for now.
+    # (details at https://gitlab.khronos.org/vulkan/vulkan/-/issues/3553#note_424431)
+    def getManualParentVUID(self, memberName: str, structName: str, commandName: str):
```
spencer-lunarg, line 626: instead of trying to create some crazy complex pytthon dict/list, decided it was easier to reason with things in a function for what VUID to assign... maybe in future this can be reduce somehow

ziga-lunarg (approved):
LGTM

layers/vulkan/generated/object_tracker.cpp
```diff
@@ -2406,14 +2413,16 @@ bool ObjectLifetimes::PreCallValidateCmdPipelineBarrier2(VkCommandBuffer command
...
+                                   "UNASSIGNED-vkCmdPipelineBarrier2-commandBuffer-commonparent", index1_loc.dot(Field::buffer));
             }
         }
         if (pDependencyInfo->pImageMemoryBarriers) {
             for (uint32_t index1 = 0; index1 < pDependencyInfo->imageMemoryBarrierCount; ++index1) {
                 [[maybe_unused]] const Location index1_loc = pDependencyInfo_loc.dot(Field::pImageMemoryBarriers, index1);
                 skip |= ValidateObject(pDependencyInfo->pImageMemoryBarriers[index1].image, kVulkanObjectTypeImage, false,
-                                       "VUID-VkImageMemoryBarrier2-image-parameter", kVUIDUndefined, index1_loc.dot(Field::image));
+                                       "VUID-VkImageMemoryBarrier2-image-parameter",
+                                       "UNASSIGNED-vkCmdPipelineBarrier2-commandBuffer-commonparent", index1_loc.dot(Field::image));
```
artem-lunarg, line 2425: `commandBuffer-commonparent` suffix. It looks like in most places in the spec they use just `-commonparent` after the command name, just to make sure new naming is fine.

spencer-lunarg: I'm not too concerned as these are very temporary names (the real VUID will be replaced when they spec is has them, hopefully by 1.3.288)

For now, I just let it get generated as `UNASSIGNED-{func}-{parameter}-commonparent` for ease of not changing the code gen too much

artem-lunarg (approved):

layers/vulkan/generated/object_tracker.cpp
artem-lunarg: Makes sense

PR #8145, opened 2024-06-13
sync: Turn remaining QueueBatchContext temp state into local variables

artem-lunarg:


layers/sync/sync_submit.h
```diff
@@ -295,10 +295,6 @@ class QueueBatchContext : public CommandExecutionContext, public std::enable_sha
     SyncEventsContext events_context_;
     BatchAccessLog batch_log_;
     std::vector<ResourceUsageTag> queue_sync_tag_;
-
-    // Clear these after validation and import, not valid after.
```
artem-lunarg, line 299: This block is finally removed

PR #8148, opened 2024-06-13
gpu: Fix missing SPV_KHR_physical_storage_buffer

spencer-lunarg:
was getting spirv-val errors from instrumented shaders 

> 1st operand of Capability: operand PhysicalStorageBufferAddresses(5347) requires one of these extensions: SPV_EXT_physical_storage_buffer SPV_KHR_physical_storage_buffer
  OpCapability PhysicalStorageBufferAddresses

so now add `OpExtension` if we are adding the `OpCapability` for the user


artem-lunarg (approved):
LGTM

Issue #8120, opened 2024-06-10
Buffer content validation disturbs push constants set by application


pc-john:
**Environment:**
 - OS: Windows 10
 - GPU and driver version: Quadro RTX 3000, drivers: 538.27
 - SDK or header version if building from repo: SDK 1.3.283.0

**Short description**

My application does not render anything when I enable "Buffer content validation" and "Indirect draws parameters". It seems that push constants are disturbed by VVL and not restored properly after performing validation.

**Long description**

My application stops rendering when I enable "Buffer content validation" and "Indirect draws parameters". If I enable just ApiDump, the application works correctly and I see following Vulkan calls:

<pre>
vkCmdBindPipeline(...)
vkCmdPushConstants(...)
vkCmdDrawIndirect(...)
</pre>

If I enable validation but not "Indirect draws parameters", I see the following:

<pre>
vkCmdBindPipeline(...)
vkCmdPushConstants(...)
vkAllocateDescriptorSets(...)   // <- probably inserted by VVL
vkUpdateDescriptorSets(...)   // <- probably inserted by VVL
vkCmdBindDescriptorSets(...)   // <- probably inserted by VVL
vkCmdDrawIndirect(...)
</pre>

My application still works. But if I enable even "Indirect draws parameters", I see the following:

<pre>
vkCmdBindPipeline(...)
vkCmdPushConstants(...)
vkCreateDescriptorSetLayout(...)   // <- probably inserted by VVL
vkCreatePipelineLayout(...)   // <- probably inserted by VVL
vkCreateShaderModule(...)   // <- probably inserted by VVL
vkCreateGraphicsPipelines(...)   // <- probably inserted by VVL
vkAllocateDescriptorSets(...)   // <- probably inserted by VVL
vkUpdateDescriptorSets(...)   // <- probably inserted by VVL
vkCmdBindPipeline(...)   // <- probably inserted by VVL
vkCmdPushConstants(...)   // <- probably inserted by VVL
vkCmdBindDescriptorSets(...)   // <- probably inserted by VVL
vkCmdBindDescriptorSets(...)   // <- probably inserted by VVL
vkCmdDraw(...)   // <- probably inserted by VVL
vkAllocateDescriptorSets(...)   // <- probably inserted by VVL
vkUpdateDescriptorSets(...)   // <- probably inserted by VVL
vkCmdBindDescriptorSets(...)   // <- probably inserted by VVL
[...]

PR #8149, opened 2024-06-13
gpu: Fix user push constants restore - and state tracking in Core validation

arno-lunarg:
- Push constants state restoration was not working correctly
- Reworked the state tracking, also in best practices
- Added 2 tests, `RestorePushConstants` is not working on my PC, vertex shader runs correctly but no fragment are emitted when doing an indirect draw. Things work as expected when doing a normal draw. It is not related to GPU-AV, as disabling it does not solve the problem. My indirect buffer looks ok, so I will see what CI tells me.

Will see if it closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8120

### DIFF
--- a/layers/best_practices/bp_state.h
+++ b/layers/best_practices/bp_state.h
@@ -198,8 +198,10 @@ class CommandBuffer : public vvl::CommandBuffer {
     bool uses_vertex_buffer = false;
     uint32_t small_indexed_draw_call_count = 0;
 
-    std::vector<uint8_t> push_constant_data_set;
-    void UnbindResources() { push_constant_data_set.clear(); }
+    // This function used to not be empty. It has been left empty because
+    // the logic to decide to call this function is not simple, so adding this
+    // function back could tedious.
+    void UnbindResources() {}
 
     struct SignalingInfo {
         // True, if the event's first state change within a command buffer is a signal (SetEvent)
--- a/layers/state_tracker/pipeline_layout_state.h
+++ b/layers/state_tracker/pipeline_layout_state.h
@@ -73,7 +73,7 @@ class PipelineLayout : public StateObject {
     using SetLayoutVector = std::vector<std::shared_ptr<vvl::DescriptorSetLayout const>>;
     const SetLayoutVector set_layouts;
     // canonical form IDs for the "compatible for set" contents
-    const PushConstantRangesId push_constant_ranges;
+    const PushConstantRangesId push_constant_ranges_layout;
     // table of "compatible for set N" cannonical forms for trivial accept validation
     const std::vector<PipelineLayoutCompatId> set_compat_ids;
     VkPipelineLayoutCreateFlags create_flags;
--- a/layers/state_tracker/state_tracker.cpp
+++ b/layers/state_tracker/state_tracker.cpp
@@ -2842,27 +2842,29 @@ void ValidationStateTracker::PostCallRecordCmdPushConstants(VkCommandBuffer comm
 
     cb_state->RecordCmd(record_obj.location.function);
     auto layout_state = Get<vvl::PipelineLayout>(layout);
-    cb_state->ResetPushConstantDataIfIncompatible(layout_state.get());
+    cb_state->ResetPushConstantRangesLayoutIfIncompatible(*layout_state);
 
-    auto &push_constant_data = cb_state->push_constant_data;
-    assert((offset + size) <= static_cast<uint32_t>(push_constant_data.size()));
-    std::memcpy(push_constant_data.data() + offset, pValues, static_cast<std::size_t>(size));
+    vvl::CommandBuffer::PushConstantData push_constant_data;
+    push_constant_data.layout = layout;
+    push_constant_data.stage_flags = stageFlags;
+    push_constant_data.offset = offset;
+    push_constant_data.values.resize(size);
+    auto byte_values = static_cast<const std::byte *>(pValues);
+    std::copy(byte_values, byte_values + size, push_constant_data.values.data());
+    // Always add submitted push constant values, even if the same data is already stored.
+    // Storing duplicated data, or data submitted by one vkCmdPushConstants call
+    // and overridden by a subsequent one is not a problem.
+    // push_constant_data_chunks is intended to be parsed from 0 to N,
+    // thus going through the history in order, so even though it is
+    // possibly suboptimal push constant data is correct.
+    cb_state->push_constant_data_chunks.emplace_back(push_constant_data);
 }
 
 void ValidationStateTracker::PostCallRecordCmdPushConstants2KHR(VkCommandBuffer commandBuffer,
                                                                 const VkPushConstantsInfoKHR *pPushConstantsInfo,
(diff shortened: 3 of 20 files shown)


tests/unit/gpu_av_positive.cpp
```diff
@@ -1261,3 +1261,189 @@ INSTANTIATE_TEST_SUITE_P(GpuAvBufferContentValidationSettings, PositiveGpuAVPara
...
+    // Bind pipeline 2, draw
+    vk::CmdBindPipeline(m_commandBuffer->handle(), VK_PIPELINE_BIND_POINT_GRAPHICS, pipe_2.Handle());
+    vk::CmdDrawIndirect(m_commandBuffer->handle(), indirect_draw_parameters_buffer.handle(), 0, 1, sizeof(VkDrawIndirectCommand));
+
+    m_commandBuffer->EndRenderPass();
+    m_commandBuffer->end();
+
+    m_default_queue->Submit(*m_commandBuffer);
+    m_default_queue->Wait();
+}
```
spencer-lunarg, line 1449: I would add another test for push constant that mixes push constants for Compute and Graphics as that is a fun case we have no GPU-AV tests for

Have both push constant write data to Storage Buffer (different for compute and graphics) and make sure we didn't clobber the values by accident 

```c++
VkPipelineLayout layout_graphics; // VK_SHADER_STAGE_FRAGMENT_BIT
VkPipelineLayout layout_compute;  // VK_SHADER_STAGE_COMPUTE_BIT

// vkBeginCommandBuffer()
vkCmdBindPipeline(pipeline_graphics); // layout_graphics
vkCmdBindPipeline(pipeline_compute);  // layout_compute

vkCmdPushConstants(layout_graphics); // VK_SHADER_STAGE_FRAGMENT_BIT
vkCmdDraw();

vkCmdPushConstants(layout_compute); // VK_SHADER_STAGE_COMPUTE_BIT
vkCmdDispatch(); // valid
```

layers/state_tracker/state_tracker.cpp
```diff
@@ -2842,27 +2842,23 @@ void ValidationStateTracker::PostCallRecordCmdPushConstants(VkCommandBuffer comm
...
-    assert((offset + size) <= static_cast<uint32_t>(push_constant_data.size()));
-    std::memcpy(push_constant_data.data() + offset, pValues, static_cast<std::size_t>(size));
+    vvl::CommandBuffer::PushConstantData push_constant_data;
+    push_constant_data.layout = layout;
+    push_constant_data.stage_flags = stageFlags;
+    push_constant_data.offset = offset;
+    push_constant_data.values.resize(size);
+    auto byte_values = static_cast<const std::byte *>(pValues);
+    std::copy(byte_values, byte_values + size, push_constant_data.values.data());
+    cb_state->push_constant_data_chunks.emplace_back(push_constant_data);
```
spencer-lunarg, line 2854: If I go

```c++
vkCmdPushConstants(offset: 0, size: 16, value = [0, 0, 0, 0]);
vkCmdDraw(); // values = [0, 0, 0, 0]

vkCmdPushConstants(offset: 4, size: 8, value = [1 ,1]);
vkCmdDraw(); // values = [0, 1, 1, 0]
```

you shuold be overriding the values, this seem like it will make 2 copies where the first is actually outdated

arno-lunarg: it will indeed not override, but there is no need to take care of this case: when restoring, I will do the exact same 2 dispatches you describe since I lookup elements in `push_constant_data_chunks` from 0 to N

tests/unit/gpu_av_positive.cpp
arno-lunarg: yes good test case!

layers/state_tracker/state_tracker.cpp
spencer-lunarg: 1. This needs a `// comment`
2. I think this will work, the only downside is we will have to track every `vkCmdPushConstants` call made so **if** they call this 100 times in the same range, we are now tracking all that memory and have to transverse it.... I have zero knowledge how many times an app will update the `vkCmdPushConstants` for a given range, so for now happy with this approuch (but leave a comment explaining all of this for the sad soul who has to touch this code next after us 2 lol)

arno-lunarg: A comment is welcome indeed.
Not to worried about not optimizing this, not doing the same call multiple times is probably more up to the application to avoid doing. And even with a 1000 elements, it is still a small number and traversing the vector should not be a problem.

tests/unit/gpu_av_positive.cpp
arno-lunarg: Added it as `RestoreUserPushConstants2`

layers/best_practices/bp_state.h
```diff
@@ -198,8 +198,7 @@ class CommandBuffer : public vvl::CommandBuffer {
     bool uses_vertex_buffer = false;
     uint32_t small_indexed_draw_call_count = 0;
 
-    std::vector<uint8_t> push_constant_data_set;
-    void UnbindResources() { push_constant_data_set.clear(); }
+    void UnbindResources() {}
```
spencer-lunarg, line 201: why keep this?

layers/state_tracker/pipeline_layout_state.h
```diff
@@ -73,7 +73,7 @@ class PipelineLayout : public StateObject {
     using SetLayoutVector = std::vector<std::shared_ptr<vvl::DescriptorSetLayout const>>;
     const SetLayoutVector set_layouts;
     // canonical form IDs for the "compatible for set" contents
-    const PushConstantRangesId push_constant_ranges;
+    const PushConstantRangesId push_constant_ranges_layout;
```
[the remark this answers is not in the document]
spencer-lunarg: ping on answer for this

layers/best_practices/bp_state.h
arno-lunarg: When I looked at the spot where `UnbindResources` is called, I felt that the logic was logic to call it or not was not easy to grasp, so I left this empty function to ease the process of adding such a function back if it is needed

spencer-lunarg: maybe a comment... this will be strange if I came across this in the future

even a `// todo - remove` type comment

arno-lunarg: done

spencer-lunarg (changes requested):
I added the 3 positive test to ToT and ran and they pass

So I guess if we are marking this as a "clean up" then this would be ok, but we are claiming this "fixes" the push constant restore, yet we added no test to prove what this fixed (and prevent regression)

(otherwise this looks great and ready to approve and merge)

arno-lunarg:
> I added the 3 positive test to ToT and ran and they pass
> 
> So I guess if we are marking this as a "clean up" then this would be ok, but we are claiming this "fixes" the push constant restore, yet we added no test to prove what this fixed (and prevent regression)
> 
> (otherwise this looks great and ready to approve and merge)

The po

PR #8154, opened 2024-06-14
build: Update to header 1.3.288

mikes-lunarg:


spencer-lunarg (approved):
pretty boring header update... feel we are making your job too easy, but lgtm

PR #8158, opened 2024-06-17
layers: Remove old Sync Val enable enum

spencer-lunarg:
@jzulauf-lunarg @mark-lunarg so I just came across https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/1976 and there is both a `VK_VALIDATION_FEATURE_ENABLE_SYNCHRONIZATION_VALIDATION` and `VK_VALIDATION_FEATURE_ENABLE_SYNCHRONIZATION_VALIDATION_EXT` that was added. 

But the logic for `SetValidationFeatureEnable2` make no sense because `VK_VALIDATION_FEATURE_ENABLE_SYNCHRONIZATION_VALIDATION` is `0` but it will hit `VK_VALIDATION_FEATURE_ENABLE_GPU_ASSISTED_EXT` first (that is also `0`)

... basically I think I should be able to remove `SetValidationFeatureEnable2`and the original `VK_VALIDATION_FEATURE_ENABLE_SYNCHRONIZATION_VALIDATION`

### DIFF
--- a/layers/layer_options.cpp
+++ b/layers/layer_options.cpp
@@ -208,16 +208,6 @@ void SetValidationFeatureEnable(CHECK_ENABLED &enable_data, const VkValidationFe
     }
 }
 
-void SetValidationFeatureEnable2(CHECK_ENABLED &enable_data, const VkValidationFeatureEnable feature_enable) {
-    switch (feature_enable) {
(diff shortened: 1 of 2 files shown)


artem-lunarg (approved):
LGTM if no objections from others.

PR #8160, opened 2024-06-17
layers: Avoid using references to out-of-scope strings

kocdemir:
Fixes possible crashes when using message filters and other string lists.


spencer-lunarg (approved):
just change your commit message to `layers: Avoid ...` and CI will be happy

artem-lunarg:
I'd like to point out that the original code is correct. The suggested change is more common though. C++ allows to bind `rvalues` to `const reference`. This often looks unusual but the reference will stay alive. 
 
https://stackoverflow.com/questions/11560339/returning-temporary-object-and-binding-to-const-reference

kocdemir:
Right, it should be working fine but I somehow experienced a crash on the usage of the string built this way on MacOS arm64 with the distributed SDK binaries, in an app with multiple vulkan instances spawned. I can debug a bit more if it happens again to see if there are some other underlying issues, but I believe not having references fixed the issue and makes the code more readable as well. 
Thanks

artem-lunarg (approved):

Issue #8103, opened 2024-06-06
debug printfs from compute shaders seemingly not outputting to console


AlexRouSg:
**Environment:**
 - OS:  Windows 10
 - GPU and driver version: NVIDIA GeForce GTX 1080 Ti, Driver: NVIDIA 555.99
 - SDK or header version if building from repo: 1.3.283
 - Options enabled (synchronization, best practices, etc.):  Debug Printf Preset

**Describe the Issue**

I am trying to debug printf from a compute shader but no messages are showing up in the console, however prints from a vertex shader do show up. Also worth noting that renderdoc captures the compute printf and there is only one message so it should not be buffer size related which is set at 1048576.

I have even tried on a very simple compute shader with no other extensions. 

**Expected behavior**

Printfs from compute shaders show up in the console.

spencer-lunarg:
@AlexRouSg I can confirm locally that `1.3.280.1`, `1.3.283` and ToT all work with compute with something like

```
#version 450
#extension GL_EXT_debug_printf : enable
void main() {
    float myfloat = 3.1415f;
    debugPrintfEXT("float == %f", myfloat);
}
 ```

Can you give me the SPIR-V that is not working, I'm curious if our implementation of Debug Printf is missing something

**Comment by AlexRouSg** [comment]
Workgroup size: 32x32x1
Workgroup count: 32x32x1
```
; SPIR-V
; Version: 1.6
; Generator: Google Shaderc over Glslang; 11
; Bound: 43
; Schema: 0
               OpCapability Shader
               OpExtension "SPV_KHR_non_semantic_info"
          %2 = OpExtInstImport "GLSL.std.450"
         %30 = OpExtInstImport "NonSemantic.DebugPrintf"
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %main "main" %gl_GlobalInvocationID %_ %t
               OpExecutionModeId %main LocalSizeId %8 %9 %10
          %1 = OpString "water.comp"
         %29 = OpString "TEST"
               OpSource GLSL 450 %1 "#version 450
#pragma shader_stage(compute)
#extension GL_EXT_debug_printf : require

#define FREQ 0.25

layout(local_size_x_id = 0, local_size_y_id = 1, local_size_z_id = 2) in;

layout(push_constant) uniform constants {
	float deltaTime;
};

layout(binding = 0, r32f) uniform restrict image2D t;

void main() {
[...]

PR #8162, opened 2024-06-17
Add StatelessData when shader is inline in pNext

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8103

`VK_KHR_maintenance5` and `VK_EXT_graphics_pipeline_library` added the ability to pass `VkShaderModuleCreateInfo` in the `pNext` of `VkPipelineShaderStageCreateInfo` (removing the need creating a `VkShaderModule`)

We did not have this all plumbed out. This adds full support for it now

### DIFF
--- a/layers/core_checks/cc_pipeline_compute.cpp
+++ b/layers/core_checks/cc_pipeline_compute.cpp
@@ -54,6 +54,14 @@ bool CoreChecks::PreCallValidateCreateComputePipelines(VkDevice device, VkPipeli
                 vku::FindStructInPNextChain<VkPipelineRobustnessCreateInfoEXT>(pCreateInfos[i].pNext)) {
             skip |= ValidatePipelineRobustnessCreateInfo(*pipeline, *pipeline_robustness_info, create_info_loc);
         }
+
+        // From dumping traces, we found almost all apps only create a 1 pipeline at a time. To greatly simplify the logic, only
+        // check the stateless validation in the pNext chain for the first pipeline. (The core issue is because we parse the SPIR-V
+        // at state tracking time, and we state track pipelines first)
+        if (i == 0 && chassis_state.stateless_data.pipeline_pnext_module) {
+            skip |= ValidateSpirvStateless(*chassis_state.stateless_data.pipeline_pnext_module, chassis_state.stateless_data,
+                                           create_info_loc.dot(Field::stage).pNext(Struct::VkShaderModuleCreateInfo, Field::pCode));
+        }
     }
     return skip;
 }
--- a/layers/core_checks/cc_pipeline_compute.cpp
+++ b/layers/core_checks/cc_pipeline_compute.cpp
@@ -55,7 +55,7 @@ bool CoreChecks::PreCallValidateCreateComputePipelines(VkDevice device, VkPipeli
             skip |= ValidatePipelineRobustnessCreateInfo(*pipeline, *pipeline_robustness_info, create_info_loc);
         }
 
-        // From dumping traces, we found almost all apps only create a 1 pipeline at a time. To greatly simplify the logic, only
+        // From dumping traces, we found almost all apps only create one pipeline at a time. To greatly simplify the logic, only
         // check the stateless validation in the pNext chain for the first pipeline. (The core issue is because we parse the SPIR-V
         // at state tracking time, and we state track pipelines first)
         if (i == 0 && chassis_state.stateless_data.pipeline_pnext_module) {
--- a/layers/core_checks/cc_pipeline_graphics.cpp
+++ b/layers/core_checks/cc_pipeline_graphics.cpp
@@ -47,6 +47,19 @@ bool CoreChecks::PreCallValidateCreateGraphicsPipelines(VkDevice device, VkPipel
         const Location create_info_loc = error_obj.location.dot(Field::pCreateInfos, i);
         skip |= ValidateGraphicsPipeline(*pipeline_states[i].get(), create_info_loc);
         skip |= ValidateGraphicsPipelineDerivatives(pipeline_states, i, create_info_loc);
+
+        // From dumping traces, we found almost all apps only create a 1 pipeline at a time. To greatly simplify the logic, only
+        // check the stateless validation in the pNext chain for the first pipeline. (The core issue is because we parse the SPIR-V
+        // at state tracking time, and we state track pipelines first)
+        if (i == 0) {
+            for (uint32_t stage = 0; stage < pCreateInfos[0].stageCount; stage++) {
+                if (chassis_state.stateless_data[stage].pipeline_pnext_module) {
+                    skip |= ValidateSpirvStateless(
+                        *chassis_state.stateless_data[stage].pipeline_pnext_module, chassis_state.stateless_data[stage],
+                        create_info_loc.dot(Field::pStages, stage).pNext(Struct::VkShaderModuleCreateInfo, Field::pCode));
+                }
+            }
+        }
     }
     return skip;
 }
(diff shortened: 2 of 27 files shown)


layers/gpu/instrumentation/gpu_shader_instrumentor.h
```diff
@@ -29,6 +29,11 @@ class Validator;
 }
 
 namespace gpu {
+
+// With VK_KHR_maintenance5 you can inline your VkShaderModuleCreateInfo via VkPipelineShaderStageCreateInfo::pNext
+// We need a way to that a "null shader module" doesn't always mean "from shader object"
```
artem-lunarg, line 34: Not sure what's here, "We need a way to tell.." ?
Also if that's a random constant then one suggestion is to use full 64 bit 0xeeeeeeeeeeeeeeee to reduce chance of collision.

layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -47,6 +47,19 @@ bool CoreChecks::PreCallValidateCreateGraphicsPipelines(VkDevice device, VkPipel
         const Location create_info_loc = error_obj.location.dot(Field::pCreateInfos, i);
         skip |= ValidateGraphicsPipeline(*pipeline_states[i].get(), create_info_loc);
         skip |= ValidateGraphicsPipelineDerivatives(pipeline_states, i, create_info_loc);
+
+        // From dumping traces, we found almost all apps only create a 1 pipeline at a time. To greatly simplify the logic, only
+        // check the stateless validation in the pNext chain for the first pipeline. (The core issue is because we parse the SPIR-V
+        // at state tracking time, and we state track pipelines first)
+        if (i == 0) {
+            for (uint32_t stage = 0; stage < pCreateInfos[0].stageCount; stage++) {
+                if (stage < spirv::kCommonMaxGraphicsShaderStages && chassis_state.stateless_data[stage].pipeline_pnext_module) {
```
artem-lunarg, line 56: nit: first part of `if` can be included in `for`:
```
uint32_t stage_count = std::min(pCreateInfos[0].stageCount, spirv::kCommonMaxGraphicsShaderStages);
for (uint32_t stage = 0; stage < stage_count; stage++)
```

layers/core_checks/cc_pipeline_compute.cpp
```diff
@@ -54,6 +54,14 @@ bool CoreChecks::PreCallValidateCreateComputePipelines(VkDevice device, VkPipeli
                 vku::FindStructInPNextChain<VkPipelineRobustnessCreateInfoEXT>(pCreateInfos[i].pNext)) {
             skip |= ValidatePipelineRobustnessCreateInfo(*pipeline, *pipeline_robustness_info, create_info_loc);
         }
+
+        // From dumping traces, we found almost all apps only create a 1 pipeline at a time. To greatly simplify the logic, only
+        // check the stateless validation in the pNext chain for the first pipeline. (The core issue is because we parse the SPIR-V
```
artem-lunarg, line 59: Can we schedule a knowledge minute for the next meeting about order of validation of pipeline<->spirv?

spencer-lunarg: sure, will be happy to rant/explain... on schedule

layers/gpu/instrumentation/gpu_shader_instrumentor.h
spencer-lunarg: updated to `0xEEEEEEEEEEEEEEEE`

also updated comment to

```

// There are 3 ways to have a null VkShaderModule
// 1. Use GPL for something like Vertex Input which won't have a shader 
// 2. Use Shader Objects
// 3. Use VK_KHR_maintenance5 and inline your VkShaderModuleCreateInfo via VkPipelineShaderStageCreateInfo::pNext
//
// The first is handled because you have to link it in the end, but we need a way to differentiate 2 and 3
```

layers/core_checks/cc_pipeline_graphics.cpp
spencer-lunarg: ya, way better

arno-lunarg (approved):

layers/gpu/instrumentation/gpu_shader_instrumentor.cpp
```diff
@@ -689,57 +694,69 @@ void GpuShaderInstrumentor::PreCallRecordPipelineCreations(uint32_t count, const
...
+                auto sm_ci =
+                    const_cast<vku::safe_VkShaderModuleCreateInfo *>(reinterpret_cast<const vku::safe_VkShaderModuleCreateInfo *>(
+                        vku::FindStructInPNextChain<VkShaderModuleCreateInfo>(stage_ci.pNext)));
+
+                if (sm_ci) {
+                    // Skip if we are in selective mode
+                    if (gpuav_settings.select_instrumented_shaders && !CheckForGpuAvEnabled(sm_ci->pNext)) continue;
+                } else {
+                    // If missing VkShaderModuleCreateInfo, for sure in GPL and not only want to check if there would be a shader
+                    // anyway
```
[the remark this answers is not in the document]
spencer-lunarg: fixed

layers/state_tracker/shader_module.h
```diff
@@ -38,6 +38,9 @@ struct Module;
 
 static constexpr uint32_t kInvalidValue = std::numeric_limits<uint32_t>::max();
 
+// It is very rare to have more then 3 stages (really only geo/tess) and better to save memory/time for the 99% use cases
```
arno-lunarg, line 41:
```suggestion
// It is very rare to have more than 3 stages (really only geo/tess) and better to save memory/time for the 99% use cases
```

arno-lunarg: parks-and-rec-ron-swanson

layers/state_tracker/pipeline_sub_state.cpp
```diff
@@ -191,8 +197,14 @@ void SetFragmentShaderInfoPrivate(FragmentShaderState &fs_state, const Validatio
                 // This support was also added in VK_KHR_maintenance5
                 if (const auto shader_ci = vku::FindStructInPNextChain<VkShaderModuleCreateInfo>(create_info.pStages[i].pNext)) {
                     // don't need to worry about GroupDecoration in GPL
-                    auto spirv_module = std::make_shared<spirv::Module>(shader_ci->codeSize, shader_ci->pCode);
+                    spirv::StatelessData *stateless_data_stage =
```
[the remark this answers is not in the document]
spencer-lunarg: no, this is all as designed... its aweful, basically there are too many ways to get a shader through Vulkan

layers/gpu/debug_printf/debug_printf.cpp
```diff
@@ -310,7 +310,7 @@ void Validator::AnalyzeAndGenerateMessage(VkCommandBuffer command_buffer, VkQueu
             shader_object_handle = it->second.shader_object;
             instrumented_spirv = it->second.instrumented_spirv;
         }
-        assert(instrumented_spirv.size() != 0);
+        ASSERT_AND_RETURN(!instrumented_spirv.empty());
```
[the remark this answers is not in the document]
spencer-lunarg: if you look at the logic

```
if (!(cond)) {          
    assert(false);      
    return;             
}        
```

it does an assert AND then returns

so it would be `ASSERT_ELSE_DONT_RETURN` then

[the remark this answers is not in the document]
arno-lunarg: But when you use this macro, your intention is to say "assert that this condition is true, else return". The underlying code, and where we actually do assert, is not really important here. The name should describe the intention, not the internal logic

Issue #7388, opened 2024-01-27
Validate VUID-RuntimeSpirv-Location-06428


Rua:
This VUID is not validated anywhere in the layer's source code:

> VUID-RuntimeSpirv-Location-06428
> The maximum number of storage buffers, storage images, and output `Location` decorated color attachments written to in the `Fragment Execution Model` **must** be less than or equal to [maxFragmentCombinedOutputResources](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#limits-maxFragmentCombinedOutputResources)

spencer-lunarg:
yes, this is similar to the `maxFragmentDualSrcAttachments` one (`VUID-RuntimeSpirv-Fragment-06427`)

I remember I was going to do this one, but needed spec clarification (which was added 
https://gitlab.khronos.org/vulkan/vulkan/-/issues/2236) and just never got around back to adding this

spencer-lunarg:
wow, so **finally** got to this and tried to implement it as 

```

    // maxFragmentCombinedOutputResources
    if (stage == VK_SHADER_STAGE_FRAGMENT_BIT) {
        // Variables can be aliased, so use Location to mark things as unique
        vvl::unordered_set<uint32_t> color_attachments;
        vvl::unordered_set<uint32_t> storage_buffers;
        vvl::unordered_set<uint32_t> storage_images;
        for (const auto *variable : entrypoint.user_defined_interface_variables) {
            if (variable->storage_class == spv::StorageClassOutput && variable->decorations.location != spirv::kInvalidValue) {
                // even if using an array of attachments in the shader, each used variable of the array is represented by a single variable
                color_attachments.insert(variable->decorations.location);
            }
        }

        for (const auto& variable : entrypoint.resource_interface_variables) {
            if (variable.decorations.location == spirv::kInvalidValue) continue;
            if (variable.is_storage_buffer) {
                storage_buffers.insert(variable.decorations.location);
            } else if (variable.is_storage_image) {
                storage_images.insert(variable.decorations.location);
            }
        }
        const uint32_t total_output = (uint32_t)(color_attachments.size() + storage_buffers.size() + storage_images.size());
        if (total_output > limits.maxFragmentCombinedOutputResources) {
            skip |= LogError("VUID-RuntimeSpirv-Location-06428", module_state.handle(), loc,
[...]

PR #8164, opened 2024-06-18
layers: Add MaxFragmentCombinedOutputResources check

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7388


layers/state_tracker/shader_module.cpp
```diff
@@ -1934,8 +1934,11 @@ bool ResourceInterfaceVariable::IsStorageBuffer(const ResourceInterfaceVariable&
     const bool physical_storage_buffer = variable.storage_class == spv::StorageClassPhysicalStorageBuffer;
     const bool storage_buffer = variable.storage_class == spv::StorageClassStorageBuffer;
     const bool uniform = variable.storage_class == spv::StorageClassUniform;
-    const bool buffer_block = variable.decorations.Has(DecorationSet::buffer_block_bit);
-    const bool block = variable.decorations.Has(DecorationSet::block_bit);
+    const bool buffer_block =
```
spencer-lunarg, line 1937: this wasn't checking if the struct also had the block decoration, only seems the Atomic tests were effected, so wrote the missing test proving this was wrong

artem-lunarg (approved):

PR #8165, opened 2024-06-18
layers: Fix checks with runtime arrays

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7677

We didn't have `OpTypeRuntimeArray` tracked well, but as shown, for some VUs, it is just not possible without a lot of GPU-AV tracking, but for now want to just remove the false positive for the edge case and keep things working for the other 99% cases

### DIFF
--- a/layers/drawdispatch/descriptor_validator.cpp
+++ b/layers/drawdispatch/descriptor_validator.cpp
@@ -343,7 +343,10 @@ bool vvl::DescriptorValidator::ValidateDescriptor(const DescriptorBindingInfo &b
                                       string_SpvDim(dim), is_image_array);
         }
 
-        if ((variable->info.image_format_type & image_view_state->descriptor_format_bits) == 0) {
+        // Because you can have a runtime array with different types in it, without extensive GPU-AV tracking, we have no way to
+        // detect if the types match up in a given index
+        if (variable->array_length != spirv::kRuntimeArray &&
+            ((variable->info.image_format_type & image_view_state->descriptor_format_bits) == 0)) {
             const bool signed_override =
                 ((variable->info.image_format_type & spirv::NumericTypeUint) && variable->info.is_sign_extended);
             const bool unsigned_override =
--- a/layers/state_tracker/shader_module.h
+++ b/layers/state_tracker/shader_module.h
@@ -38,6 +38,9 @@ struct Module;
 
 static constexpr uint32_t kInvalidValue = std::numeric_limits<uint32_t>::max();
 
+// Need to find a way to know if actually array lenght of zero, or a runtime array.
+static constexpr uint32_t kRuntimeArray = std::numeric_limits<uint32_t>::max();
+
 // This is the common info for both OpDecorate and OpMemberDecorate
 // Used to keep track of all decorations applied to any instruction
 struct DecorationBase {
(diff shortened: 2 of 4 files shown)


layers/drawdispatch/descriptor_validator.cpp
```diff
@@ -343,7 +343,10 @@ bool vvl::DescriptorValidator::ValidateDescriptor(const DescriptorBindingInfo &b
                                       string_SpvDim(dim), is_image_array);
         }
 
-        if ((variable->info.image_format_type & image_view_state->descriptor_format_bits) == 0) {
+        // Because you can a runtime array with different types in it, without extensive GPU-AV tracking, we have no way to detect
```
filnet, line 346: typo: you can [have] a runtime

layers/state_tracker/shader_module.h
```diff
@@ -38,6 +38,9 @@ struct Module;
 
 static constexpr uint32_t kInvalidValue = std::numeric_limits<uint32_t>::max();
 
+// Need to a way to know if actually array lenght of zero, or a runtime array.
```
filnet, line 41: to [find] a way

layers/drawdispatch/descriptor_validator.cpp
```diff
@@ -343,7 +343,10 @@ bool vvl::DescriptorValidator::ValidateDescriptor(const DescriptorBindingInfo &b
                                       string_SpvDim(dim), is_image_array);
         }
 
-        if ((variable->info.image_format_type & image_view_state->descriptor_format_bits) == 0) {
+        // Because you can have a runtime array with different types in it, without extensive GPU-AV tracking, we have no way to
```
artem-lunarg, line 346: What's the example of different types? As in different textures (2d, 3d) in a single array?

That's also interesting tendency with validation of runtime arrays. We also disabled array processing in syncval for descriptor validation.

artem-lunarg (approved):

layers/drawdispatch/descriptor_validator.cpp
spencer-lunarg:
```glsl
// valid GLSL and SPIR-V
layout(set = 0, binding = 0) uniform texture2D float_textures[];
layout(set = 0, binding = 0) uniform utexture2D uint_textures[];
 ```
 
 so in your array the app know which index contains the float or int, but since they can share the same set/binding  (this is the issue), then we have no really good way to know what is being accessed
 
 we would have to save the texture type (float/int/etc) in the GPU-AV output, which is not impossible (we do it for the `set`/`binding`),  but currently we are trying to reduce our GPU-AV overhead and need to be smarter how we can go about writting out all this information
 
 For this one edge case, I rather just be simple and not report the false positive

artem-lunarg: > since they can share the same set/binding
🤯

artem-lunarg: > For this one edge case, I rather just be simple and not report the false positive

100% agree. Personally I prefer 80% good validation then 100% with rare false-positives.

spencer-lunarg: yes, I was shocked this was allowed as well... it only works because the way the `float`/`int` type in SPIR-V is tied to the image access (ex `texelFetch()`)  and not the variable itself

Issue #6937, opened 2023-11-10
Need GPU-AV bindless descriptor test for VUID 02707


jeremyg-lunarg:
This can be based on NegativeProtectedMemory.MixingProtectedResources

PR #8168, opened 2024-06-19
gpu: Fix protected memory on descriptor indexing

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6937

Added 2 tests and found the protected memory check was wrapped in the wrong `if` so fixed that as well

### DIFF
--- a/layers/drawdispatch/descriptor_validator.cpp
+++ b/layers/drawdispatch/descriptor_validator.cpp
@@ -582,16 +582,17 @@ bool vvl::DescriptorValidator::ValidateDescriptor(const DescriptorBindingInfo &b
                 }
             }
         }
-        if (dev_state.enabled_features.protectedMemory == VK_TRUE) {
-            if (dev_state.ValidateProtectedImage(cb_state, *image_view_state->image_state, loc,
-                                                 vuids.unprotected_command_buffer_02707, " (Image is in a descriptorSet)")) {
-                return true;
-            }
-            if (variable->IsWrittenTo() &&
-                dev_state.ValidateUnprotectedImage(cb_state, *image_view_state->image_state, loc,
-                                                   vuids.protected_command_buffer_02712, " (Image is in a descriptorSet)")) {
-                return true;
-            }
+    }
+
+    if (dev_state.enabled_features.protectedMemory == VK_TRUE) {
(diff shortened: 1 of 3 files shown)


tests/unit/gpu_av_descriptor_indexing.cpp
```diff
@@ -2365,3 +2365,91 @@ TEST_F(NegativeGpuAVDescriptorIndexing, MultipleOOBTypesInOneCmdBuffer) {
     m_default_queue->Wait();
     m_errorMonitor->VerifyFound();
 }
+
+TEST_F(NegativeGpuAVDescriptorIndexing, MixingProtectedResources) {
+    TEST_DESCRIPTION("Have protected resources, but don't actually access it so no VU is triggered");
```
arno-lunarg, line 2370:
```suggestion
    TEST_DESCRIPTION("Have protected resources, but don't actually access it so no VU is triggered");
```
Update description

arno-lunarg (approved):

layers/drawdispatch/descriptor_validator.cpp
```diff
@@ -582,16 +582,17 @@ bool vvl::DescriptorValidator::ValidateDescriptor(const DescriptorBindingInfo &b
...
-            if (dev_state.ValidateProtectedImage(cb_state, *image_view_state->image_state, loc,
-                                                 vuids.unprotected_command_buffer_02707, " (Image is in a descriptorSet)")) {
-                return true;
-            }
-            if (variable->IsWrittenTo() &&
-                dev_state.ValidateUnprotectedImage(cb_state, *image_view_state->image_state, loc,
-                                                   vuids.protected_command_buffer_02712, " (Image is in a descriptorSet)")) {
-                return true;
-            }
+    }
```
jeremyg-lunarg, line 585: It took me a while to realize the only non-whitespace part of this change was moving the closing brace!

jeremyg-lunarg (approved):
LGTM but the PR title is... odd

spencer-lunarg:
@jeremyg-lunarg updated PR title

layers/drawdispatch/descriptor_validator.cpp
spencer-lunarg: ya, the negative test was a good thing to add, this logic was inside the parent `if ()` for no reason

PR #8169, opened 2024-06-19
layers: Add vkGetDeviceFaultInfoEXT check

spencer-lunarg:
simple check, we already have `is_device_lost` piped out so


artem-lunarg (approved):
LGTM. Cool extension. From desktop vendors I can see AMD reports, and it looks like no support on NV, Intel.

PR #8175, opened 2024-06-20
layers: Remove unused unistd.h includes

pinotree:
The sources seem to build fine even without them (at least on Linux), so drop these OS-specific includes.

### DIFF
--- a/layers/gpu/core/gpuav.cpp
+++ b/layers/gpu/core/gpuav.cpp
@@ -22,9 +22,6 @@
 
 #include <cmath>
 #include <fstream>
-#if defined(__linux__) || defined(__FreeBSD__) || defined(__OpenBSD__)
-#include <unistd.h>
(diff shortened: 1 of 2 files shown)


spencer-lunarg (approved):
LGTM - @pinotree I'm curious how you noticed this, I enjoy removing unused `#include` where possible

pinotree:
> I'm curious how you noticed this

Because of my GNU/Hurd porting (#8177): rather than adding more `ifdef`s, I decided to drop the ones that were not needed (at least as it seemed).

spencer-lunarg:
> CI Vulkan-ValidationLayers build # 16879 failed.

The error logs are showing strange github/internet issue

> ssh: Could not resolve hostname github.com: Temporary failure in name resolution

will wait a bit and kick it off again and get this in soon

PR #8179, opened 2024-06-20
layers: Extra null checks

spencer-lunarg:
Found these while working on other things, seem like spots were I couldn't quickly confirm we will always have a valid pointer so added assert/null checks

### DIFF
--- a/layers/best_practices/bp_pipeline.cpp
+++ b/layers/best_practices/bp_pipeline.cpp
@@ -114,9 +114,10 @@ bool BestPractices::PreCallValidateCreateGraphicsPipelines(VkDevice device, VkPi
     for (uint32_t i = 0; i < createInfoCount; i++) {
         const Location create_info_loc = error_obj.location.dot(Field::pCreateInfos, i);
         const auto& create_info = pCreateInfos[i];
-        const auto& pipeline = *pipeline_states[i].get();
+        const auto pipeline = pipeline_states[i].get();
+        ASSERT_AND_CONTINUE(pipeline);
 
-        if (!(pipeline.active_shaders & VK_SHADER_STAGE_MESH_BIT_EXT) && create_info.pVertexInputState) {
+        if (!(pipeline->active_shaders & VK_SHADER_STAGE_MESH_BIT_EXT) && create_info.pVertexInputState) {
             const auto& vertex_input = *create_info.pVertexInputState;
             uint32_t count = 0;
             for (uint32_t j = 0; j < vertex_input.vertexBindingDescriptionCount; j++) {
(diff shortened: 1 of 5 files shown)


layers/best_practices/bp_pipeline.cpp
```diff
@@ -114,9 +114,10 @@ bool BestPractices::PreCallValidateCreateGraphicsPipelines(VkDevice device, VkPi
     for (uint32_t i = 0; i < createInfoCount; i++) {
         const Location create_info_loc = error_obj.location.dot(Field::pCreateInfos, i);
         const auto& create_info = pCreateInfos[i];
-        const auto& pipeline = *pipeline_states[i].get();
+        const auto pipeline = pipeline_states[i].get();
+        ASSERT_AND_CONTINUE(pipeline);
```
arno-lunarg, line 118:
```suggestion
        const auto *pipeline_ptr = pipeline_states[i].get();
        ASSERT_AND_CONTINUE(pipeline_ptr);
        const auto &pipeline = *pipeline_ptr;
```
So that the code preserves the usage of a const ref, I think it was better this way

layers/best_practices/bp_wsi.cpp
```diff
@@ -277,7 +277,7 @@ void BestPractices::ManualPostCallRecordGetPhysicalDeviceSurfaceFormatsKHR(VkPhy
     if (auto bp_pd_data = Get<bp_state::PhysicalDevice>(physicalDevice)) {
         auto& call_state = bp_pd_data->vkGetPhysicalDeviceSurfaceFormatsKHRState;
 
-        if (*pSurfaceFormatCount) {
+        if (pSurfaceFormatCount) {
```
arno-lunarg, line 280:
```suggestion
        if (*pSurfaceFormatCount) {
```
The meaning is not the same. I think the original was checking for `*pSurfaceFormatCount > 0`, which here is equivalent to just checking `*pSurfaceFormatCount`. So I think we need to revert this change

layers/best_practices/bp_wsi.cpp
```diff
@@ -297,7 +297,7 @@ void BestPractices::ManualPostCallRecordGetPhysicalDeviceSurfaceFormats2KHR(VkPh
                                                                             VkSurfaceFormat2KHR* pSurfaceFormats,
                                                                             const RecordObject& record_obj) {
     if (auto bp_pd_data = Get<bp_state::PhysicalDevice>(physicalDevice)) {
-        if (*pSurfaceFormatCount) {
+        if (pSurfaceFormatCount) {
```
arno-lunarg, line 300: same

arno-lunarg (changes requested):

layers/best_practices/bp_pipeline.cpp
spencer-lunarg: honestly in the future if someone works on BP again, there is a 100% chance that Pipeline becomes a large mess like it is in core check.. I'm just going to call into another function, because it isolates a single `const VkGraphicsPipelineCreateInfo* pCreateInfos` like we normally do

spencer-lunarg: made separate commit

arno-lunarg (approved):

PR #8181, opened 2024-06-21
tests: Cleanup test to use framework more

spencer-lunarg:
just a friendly 1000+ line clean up of test code to be more unify in using the framework


tests/unit/buffer_positive.cpp
```diff
@@ -109,7 +109,6 @@ TEST_F(PositiveBuffer, DISABLED_PerfGetBufferAddressWorstCase) {
     constexpr size_t N = 1400;
     std::vector<vkt::Buffer> buffers(N);
     VkBufferCreateInfo buffer_ci = vku::InitStructHelper();
-    buffer_ci.sharingMode = VK_SHARING_MODE_EXCLUSIVE;
```
spencer-lunarg, line 112: `InitStructHelper` sets everything to zero

except for the few tests who need to care about `sharingMode`, having it here just causes it to get copy-and-pasted for no good reason

artem-lunarg (approved):

Issue #5596, opened 2023-04-07
Unable to find suitable #line directive in SPIR-V source


DethRaid:
**Environment:**
 - OS: Windows 11
 - GPU: RTX 2080 Super
 - SDK or header version if building from repo: 1.3.243.0
 - Options enabled (synchronization, best practices, etc.): Synchronization, GPU-assisted validation

**Describe the Issue**

GPU assisted validation reports a shader file name and line, but this line is incorrect. It also reports that it's unable to find a suitable line directive in my shader

**Expected behavior**

GPU assisted validation should either report a correct shader file name and line, or it should report that it's unable to determine the correct shader file name or line

**Valid Usage ID**
Validation Error: [ UNASSIGNED-Descriptor uninitialized ] Object 0: handle = 0x1954061ece0, name = Graphics queue, type = VK_OBJECT_TYPE_QUEUE; | MessageID = 0x893513c7 | Descriptor index 1 is uninitialized. Command buffer (0x19555250b40). Compute Dispatch Index 0x2. Pipeline (Inject into GV)(0x67022e000000004b). Shader Module (0x7323f50000000048). Shader Instruction Index = 379.  Stage = Compute.  Global invocation ID (x, y, z) = (1, 3, 0 ) Shader validation error occurred in file D:/Source/SahRenderer/Windows/../RenderCore/shaders/lpv/inject_into_gv.comp at line 78.Unable to find suitable #line directive in SPIR-V OpSource.

**Additional context**

I have some problem in my shader. I enabled GPU assisted validation to help debug it. I received the above message

I don't know how to interpret it. It reports that the shader validation error occurred in a certain line of a certain file, but then it immediately states that it's unable to find a suitable line directive. There's no space before that last sentence - is this a bug with how the message is being output? 

If GPU assisted validation is able to determine the shader file and line where the error occurred, why does it report that it's unable to find a suitable line directive? If it's not able to determine where the error occurred, why does it report a shader file and line? The whole thing is very confusing

juan-lunarg:
Can you provide a [SSCCE](http://www.sscce.org/) of the shader in question? 

NOTE: We don't need your actual shader. Just a shader that can reproduce the issue.

Tony-LunarG:
If it could have found a #line directive in the SPIR-V OpSource, it would have printed the line of source code where it detected the error. I'd agree that the message could be clearer.

DethRaid:
[...]

PR #8182, opened 2024-06-21
gpu: Fix/Improve shader error messages

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5596

Cleans up the error message debug info for GPU-AV / DebugPrintF

Adds a suite of tests to make sure we are reporting the correct data in the future as well


layers/gpu/error_message/gpuav_error_message.cpp
```diff
@@ -588,45 +417,215 @@ bool Validator::AnalyzeAndGenerateMessage(VkCommandBuffer cmd_buffer, const LogO
...
+                     debug_info_message.c_str());
         }
     }
 
     return error_found;
 }
-
 }  // namespace gpuav
+
+namespace gpu {
```
arno-lunarg, line 457: This does not belong here, please move it to `gpu_shader_instrumentor.cpp`, maybe close to the definition of `InternalError` and `InternalWarning`

layers/gpu/error_message/gpuav_error_message.cpp
```diff
@@ -588,45 +417,215 @@ bool Validator::AnalyzeAndGenerateMessage(VkCommandBuffer cmd_buffer, const LogO
...
+           << LookupDebugUtilsNameNoLock(debug_report, HandleToUint64(commandBuffer)) << "(" << HandleToUint64(commandBuffer)
+           << ")\n";
+        assert(true);
+    } else {
+        std::unique_lock<std::mutex> lock(debug_report->debug_output_mutex);
+        ss << "Command buffer " << LookupDebugUtilsNameNoLock(debug_report, HandleToUint64(commandBuffer)) << "("
+           << HandleToUint64(commandBuffer) << ")\n";
+
+        ss << std::dec << std::noshowbase;
+        ss << "\t";  // help show the index is with respect to the command buffer
```
arno-lunarg, line 480:
```suggestion
        ss << "\t";  // helps to show that the index is expressed with respect to the command buffer
```

layers/gpu/error_message/gpuav_error_message.cpp
```diff
@@ -588,45 +417,215 @@ bool Validator::AnalyzeAndGenerateMessage(VkCommandBuffer cmd_buffer, const LogO
...
+            }
+            ss << "\n";
+            break;
+        }
+        // OpString can only be in the debug section, so can break early if not found
+        if (insn.Opcode() == spv::OpFunction) break;
+    }
+
+    if (!found_opstring) {
+        ss << "Unable to find SPIR-V OpString for file id " << reported_file_id << " from OpLine instruction.\n";
```
arno-lunarg, line 565:
```suggestion
        ss << "Unable to find SPIR-V OpString for file from OpLine instruction.\n";
```
file id already reported below

layers/gpu/error_message/gpuav_error_message.cpp
```diff
@@ -588,45 +417,215 @@ bool Validator::AnalyzeAndGenerateMessage(VkCommandBuffer cmd_buffer, const LogO
...
+                        saved_opsource_offset = std::distance(opsource_lines.begin(), it);
+                        found_best_line = true;
+                    }
+                }
+            }
+        }
+
+        if (found_best_line) {
+            assert(reported_line_number >= saved_line_number);
+            std::vector<std::string>::size_type opsource_index =
```
arno-lunarg, line 607:
```suggestion
            size_t opsource_index =
```

tests/unit/gpu_av_shader_debug_info.cpp
```diff
@@ -0,0 +1,584 @@
...
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+#include "../framework/layer_validation_tests.h"
+#include "../framework/pipeline_helper.h"
+#include "../framework/descriptor_helper.h"
+#include "../framework/gpu_av_helper.h"
+
+void NegativeGpuAVShaderDebugInfo::BasicTestComputeOOO(const char *shader, const char *error) {
```
arno-lunarg, line 18: Please add a description telling what kind of shaders with what resources need to be passed to this function

tests/unit/gpu_av_shader_debug_info.cpp
```diff
@@ -0,0 +1,584 @@
...
+        layout(set = 0, binding = 0) buffer foo {
+            IndexBuffer data;
+            int x;
+        };
+        void main()  {
+            x = data.indices[16];
+        }
+    )glsl";
+
+    CreateComputePipelineHelper pipe(*this);
```
arno-lunarg, line 345: Could BasicTestComputeOOO be used? Same questions for tests below

arno-lunarg (approved):
Nice changes, only thing I really want is to move `GpuShaderInstrumentor::GenerateDebugInfoMessage`

tests/unit/gpu_av_shader_debug_info.cpp
spencer-lunarg: it has to call `SetDebugUtilsObjectNameEXT` in the middle, debated about starting to add lots of optional parameters to `BasicTestComputeOOO` for fear it will because a mess, but for now it is only 1 optional argument so probably ok

layers/gpu/error_message/gpuav_error_message.cpp
spencer-lunarg: ya, agree, it was easier when doing since it was just replacing the old commands, but yes, doesn't actually belong here

PR #8185, opened 2024-06-24
build: Automate self vvl renaming

spencer-lunarg:
This will `-DBUILD_SELF_VVL=ON` in CI to be able to build the Self Validation so we can run on it easier

cc @arno-lunarg


arno-lunarg (approved):
nice!

y-novikov:
ANGLE isn't able to load VK_LAYER_KHRONOS_validation after this:
https://chromium-review.googlesource.com/c/angle/angle/+/5650353

Could you please update BUILD.gn similarly to CMakeLists.txt?
@cnorthrop

y-novikov:
@dneto0 in case you wonder why https://chromium-review.googlesource.com/c/chromium/src/+/5650456 failed.

spencer-lunarg:
Created https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8188

I only grepped in `BUILD.gn` but seems there is some extra script (wasn't aware of) that is where these JSON files are injected with the string

Issue #8140, opened 2024-06-12
Allow smarter error message checking for tests


spencer-lunarg:
Currently for GPU-AV, it would be nice to have a way for it to make sure we test the error message is good

My advice is to always have the VUID (so we know where the test coverage is for any VUID) so it could look like

`m_errorMonitor->SetDesiredError("VUID-xxx", "regex to error message");`

this can be done easily because in `DebugCallback()` we can just do the VUID matching in `VkDebugUtilsMessengerCallbackDataEXT::pMessageIdName` and the rest would be for the error message

PR #8187, opened 2024-06-24
tests: Add support for regex lookup in error msg

arno-lunarg:
Proposal for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8140

For now only tried it in NegativeGpuAVIndirectBuffer.DrawCount

### DIFF
--- a/tests/framework/error_monitor.cpp
+++ b/tests/framework/error_monitor.cpp
@@ -25,7 +25,7 @@ static VKAPI_ATTR VkBool32 VKAPI_CALL DebugCallback(VkDebugReportFlagsEXT messag
     auto *error_monitor = reinterpret_cast<ErrorMonitor *>(user_data);
 
     if (message_flags & error_monitor->GetMessageFlags()) {
-        return error_monitor->CheckForDesiredMsg(message);
+        return error_monitor->CheckForDesiredMsg("", message);
     }
     return VK_FALSE;
 }
@@ -54,11 +54,12 @@ static VKAPI_ATTR VkBool32 VKAPI_CALL DebugCallback(VkDebugUtilsMessageSeverityF
                                                     VkDebugUtilsMessageTypeFlagsEXT message_types,
                                                     const VkDebugUtilsMessengerCallbackDataEXT *callback_data, void *user_data) {
     const auto message_flags = DebugAnnotFlagsToMsgTypeFlags(message_severity, message_types);
+    const char *vuid = callback_data->pMessageIdName;
     const char *message = callback_data->pMessage;
     auto *error_monitor = reinterpret_cast<ErrorMonitor *>(user_data);
 
     if (message_flags & error_monitor->GetMessageFlags()) {
-        return error_monitor->CheckForDesiredMsg(message);
+        return error_monitor->CheckForDesiredMsg(vuid, message);
     }
     return VK_FALSE;
 }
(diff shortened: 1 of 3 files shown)


tests/framework/error_monitor.h
```diff
@@ -19,6 +19,7 @@
 
 #include "test_common.h"
 #include <unordered_set>
+#include <regex>
```
spencer-lunarg, line 22: this needs to be in the header? not in `error_monitor.cpp`?

spencer-lunarg (approved):
It looks good, one nit about the `include`

nice part is this is forward looking and won't break anything we already have

tests/unit/gpu_av_indirect_buffer.cpp
```diff
@@ -241,7 +241,10 @@ TEST_F(NegativeGpuAVIndirectBuffer, DrawCount) {
     pipe.gp_ci_.layout = pipeline_layout.handle();
     pipe.CreateGraphicsPipeline();
 
-    m_errorMonitor->SetDesiredError("VUID-vkCmdDrawIndirectCount-countBuffer-03122");
+    m_errorMonitor->SetDesiredErrorRegex(
+        "VUID-vkCmdDrawIndirectCount-countBuffer-03122",
+        "Indirect draw count of 2 would exceed buffer size 16 of buffer VkBuffer 0x[a-f0-9]*\\[\\] stride = 16 offset = 0 "
```
spencer-lunarg, line 246: will be fun copying `0x[a-f0-9]*\\[\\] ` all over :laughing:

tests/framework/error_monitor.h
arno-lunarg: yes, for `std::regex` definition

tests/unit/gpu_av_indirect_buffer.cpp
arno-lunarg: oh I left that... honestly ` 0x[a-f0-9]*\\[\\] ` could just be `.*`

spencer-lunarg: ... that would feel nicer in my soul to copy in the future

tests/framework/error_monitor.h
spencer-lunarg: missed that

tests/framework/error_monitor.cpp
```diff
@@ -149,6 +165,12 @@ void ErrorMonitor::SetDesiredError(const char *msg, uint32_t count) {
     }
 }
 
+void ErrorMonitor::SetDesiredErrorRegex(const char *vuid, std::string msg_regex, uint32_t count /*= 1*/) {
+    for (uint32_t i = 0; i < count; i++) {
+        SetDesiredFailureMsgRegex(kErrorBit, vuid, msg_regex);
```
filnet, line 170: if count is greater than one then you might end up "compiling" the regex string multiple times.

spencer-lunarg: yes, I agree, will take a look later, for the short term, most things use count of `1` and the few GPU-AV tests that use it is `3` so we shouldn't accidentally hit a crazy regex compiling bottleneck, but I always monitor how long tests take to run regardless, so good thing to note as a possible spot to look at

spencer-lunarg: nvm, @arno-lunarg is a better man than me https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8206

PR #8188, opened 2024-06-24
build: Fix GN build for JSON_LAYER_NAME

spencer-lunarg:
Should be a fix for https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8185#issuecomment-2186884474

@y-novikov if you want to try this


artem-lunarg (approved):

y-novikov:
Thanks, seem like this fixed it https://chromium-review.googlesource.com/c/angle/angle/+/5650184.

PR #8193, opened 2024-06-25
gpuav: Cleanup

arno-lunarg:
Removed last methods that were in `gpuav.h` but did not belong here (shader instrumentation error logging logic). Now it only contains virtual methods. The `Validator` class was becoming too big, so I moved everything that did not strictly need to be there outside of it.

Also remove abstraction level to set output buffer size (error buffer for GPU AV, log buffer for printf), it did not add anything of value


layers/gpu/shaders/cmd_validation/buffer_device_address_validation.comp
```diff
@@ -0,0 +1,123 @@
+// Copyright (c) 2024 The Khronos Group Inc.
```
spencer-lunarg, line 1: was this file added by accident?

spencer-lunarg (approved):
LGTM - pending why we now have two `buffer_device_address.comp` files

layers/gpu/shaders/cmd_validation/buffer_device_address_validation.comp
arno-lunarg: oopsy yes

PR #8194, opened 2024-06-25
tests: Improve error checking/reporting in test surface creation

r-potter:
Without this change, running tests headlessly via SSH will fail to create a valid surface, but report success and subsequently crash in the ICD.


spencer-lunarg (approved):

spencer-lunarg:
> Author r-potter not on autobuild list. 

I added you to the list and manually kicked off CI as well

spencer-lunarg:
I think a better fix is going


```patch
-  ASSERT_EQ(VK_SUCCESS, CreateSurface(surface_context, instance2_surface.handle, instance2));
+   if (CreateSurface(surface_context, instance2_surface.handle, instance2) != VK_SUCCESS) {
+        GTEST_SKIP() << "Cannot create 2nd surface";
+   }
  ```
  
  on them all

PR #8195, opened 2024-06-25
tests: Fix 'mismatch' typos

r-potter:


spencer-lunarg (approved):
100% my doing... one day I will take those English classes

PR #8198, opened 2024-06-26
layers: Add Dispatch helpers for GPDP2 calls

spencer-lunarg:
We make a lot of internal calls to `VK_KHR_get_physical_device_properties2` functions, but we need to make sure to use the `DispatchGetPhysicalDeviceFormatProperties2` vs `DispatchGetPhysicalDeviceFormatProperties2KHR` depending if still using Vulkan 1.0 or not

This adds a DispatchGetPhysicalDevice*2Helper` for all those calls and substitutes them everywhere

### DIFF
--- a/layers/core_checks/cc_image.cpp
+++ b/layers/core_checks/cc_image.cpp
@@ -69,11 +69,11 @@ bool CoreChecks::ValidateImageFormatFeatures(const VkImageCreateInfo &create_inf
         if (has_format_feature2) {
             VkDrmFormatModifierPropertiesList2EXT fmt_drm_props = vku::InitStructHelper();
             VkFormatProperties2 fmt_props_2 = vku::InitStructHelper(&fmt_drm_props);
-            DispatchGetPhysicalDeviceFormatProperties2(physical_device, image_format, &fmt_props_2);
+            DispatchGetPhysicalDeviceFormatProperties2Helper(physical_device, image_format, &fmt_props_2);
             std::vector<VkDrmFormatModifierProperties2EXT> drm_properties;
             drm_properties.resize(fmt_drm_props.drmFormatModifierCount);
             fmt_drm_props.pDrmFormatModifierProperties = drm_properties.data();
-            DispatchGetPhysicalDeviceFormatProperties2(physical_device, image_format, &fmt_props_2);
+            DispatchGetPhysicalDeviceFormatProperties2Helper(physical_device, image_format, &fmt_props_2);
 
             for (uint32_t i = 0; i < fmt_drm_props.drmFormatModifierCount; i++) {
                 if (drm_format_modifiers.find(fmt_drm_props.pDrmFormatModifierProperties[i].drmFormatModifier) !=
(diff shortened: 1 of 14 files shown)


layers/vulkan/generated/chassis.cpp
```diff
@@ -143,6 +143,72 @@ static void InitDeviceObjectDispatch(ValidationObject* instance_interceptor, Val
     }
 }
 
+void ValidationObject::DispatchGetPhysicalDeviceFeatures2Helper(VkPhysicalDevice physicalDevice,
+                                                                VkPhysicalDeviceFeatures2* pFeatures) const {
+    if (api_version != VK_API_VERSION_1_0) {
```
aqnuep, line 148: I still think `>= VK_API_VERSION_1_1` would better describe the situation: i.e. this is only supported since Vulkan 1.1.

It would also avoid having problems from `api_version` containing patch numbers, as 1.0.268 != 1.0.0.

layers/state_tracker/state_tracker.cpp
```diff
@@ -1052,7 +1053,7 @@ void ValidationStateTracker::PostCreateDevice(const VkDeviceCreateInfo *pCreateI
     }
 
     // Query queue family extension properties
-    {
+    if (IsExtEnabled(device_extensions.vk_khr_get_physical_device_properties2)) {
```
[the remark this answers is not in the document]
spencer-lunarg: sorry, I guess I can move it below where the old check was around `DispatchGetPhysicalDeviceQueueFamilyProperties2`
 
note the `IsExtEnabled` call returns true if either it was added via extension OR via api version (there is a `IsExtEnabledByCreateinfo` as well)

layers/vulkan/generated/chassis.cpp
spencer-lunarg: ok

layers/state_tracker/state_tracker.cpp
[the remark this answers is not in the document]
aqnuep: Oh, you're right. I forgot about that one. Yeah, it should be fine then.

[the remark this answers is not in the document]
spencer-lunarg: ya, I didn't think there was any use to do any part of the block of code... the `IsExtEnabled` vs `IsExtEnabledByCreateinfo` can easy to mix up, but basically it is always `IsExtEnabled` unless you know you need it to be only the extension

layers/core_checks/cc_image.cpp
```diff
@@ -255,8 +255,9 @@ bool CoreChecks::PreCallValidateCreateImage(VkDevice device, const VkImageCreate
     // Exit early if any thing is not succesful
     VkResult result = VK_SUCCESS;
     if (pCreateInfo->tiling != VK_IMAGE_TILING_DRM_FORMAT_MODIFIER_EXT) {
-        if (IsExtEnabled(device_extensions.vk_khr_get_physical_device_properties2)) {
-            result = DispatchGetPhysicalDeviceImageFormatProperties2(physical_device, &image_format_info, &image_format_properties);
+        if (IsExtEnabled(instance_extensions.vk_khr_get_physical_device_properties2)) {
+            result = DispatchGetPhysicalDeviceImageFormatProperties2Helper(physical_device, &image_format_info,
```
artem-lunarg, line 259: Can we have here only `DispatchGetPhysicalDeviceImageFormatProperties2Helper` without both `if` and `else`?

spencer-lunarg: no

there are 3 cases

1. Vulkan 1.0 - no extension (use `vkGetPhysicalDeviceImageFormatProperties()`)
2. Vulkan 1.0 - extension (use `vkGetPhysicalDeviceImageFormatProperties2KHR()`)
3. Vulkan 1.1 (use `vkGetPhysicalDeviceImageFormatProperties2()`)

the goal of this PR was some spots were check to call `1` vs `2`, but not in others... it is not obvious, so the helper helps decide... this in theory could be done everywhere, but `VK_KHR_get_physical_device_properties2` is the majority of `Dispatch` calls we make in Core Validation

artem-lunarg (approved):

spencer-lunarg:
@aqnuep just a friendly ping on this, (no rush as **will** take a look at it until Monday, but just peeking at open PRs right now)

aqnuep (approved):
You've addressed my only comments, @spencer-lunarg, so this is fine by me. Thanks!

PR #8201, opened 2024-06-26
sync: Update allocation scheme of HandleRecord objects

artem-lunarg:
This gets rid of allocations when we need to store many resource handles, and as a bonus reduces size of structures that store resource handles:
`NamedHandle` (new name `HandleRecord`): 64 bytes -> 16 bytes
`ResourceCmdUsageRecord`: 128 bytes -> 40 bytes

This size reduction means that even if we store few handles per API call in order to have them in error messages (single handle structure `NamedHandle` is 16 bytes on most platforms) it will still use less memory then original implementation that did store most of the handles.

This change adds resource handles only to a single API call as an example and to be able to run the test. Separate PR will add a bunch of those.


layers/sync/sync_validation.cpp
```diff
@@ -236,7 +236,10 @@ void SyncValidator::PreCallRecordCmdCopyBuffer(VkCommandBuffer commandBuffer, Vk
     auto *context = cb_context->GetCurrentAccessContext();
 
     auto src_buffer = Get<vvl::Buffer>(srcBuffer);
+    cb_context->AddCommandHandle(tag, src_buffer->Handle());
```
artem-lunarg, line 239: This is just an example for a single API call (vkCmdCopyBuffer). Similar code should be added in other places.

tests/unit/sync_val.cpp
```diff
@@ -6228,3 +6228,35 @@ TEST_F(NegativeSyncVal, RenderPassStoreOpNone) {
...
+    vk::CmdCopyBuffer(cb0, buffer_a, buffer_b, 1, &region);
+    cb0.end();
+    m_default_queue->Submit(cb0);
+
+    vkt::CommandBuffer cb1(*m_device, m_command_pool);
+    cb1.SetName(m_device->handle(), "CB1");
+    cb1.begin();
+    vk::CmdCopyBuffer(cb1, buffer_a, buffer_b, 1, &region);
+    cb1.end();
+    m_errorMonitor->SetDesiredErrorRegex("SYNC-HAZARD-WRITE-AFTER-WRITE", "BufferA.*BufferB");
```
artem-lunarg, line 6258: Regex in action

spencer-lunarg (approved):
This looks like a good first step, excited to see the next PR.

PR #8206, opened 2024-06-27
tests: Do not create lookup regex more that once

arno-lunarg:


tests/framework/error_monitor.h
```diff
@@ -82,6 +82,7 @@ class ErrorMonitor {
     void ExpectSuccess(VkDebugReportFlagsEXT const message_flag_mask = kErrorBit);
 
   private:
+    void SetDesiredFailureMsgRegex(const VkFlags msg_flags, const char *vuid, std::string msg_regex, std::regex regex);
```
spencer-lunarg, line 85: i was worried we added yet-another-way to set the error, but having it in `private` calmed me down

spencer-lunarg (approved):

PR #8207, opened 2024-06-27
sync: Add basic stats support

artem-lunarg:
Some basic scaffolding that's enough for the current task, but the plan is to evolve this further.

Stats are disabled by default (`VVL_ENABLE_SYNCVAL_STATS` in `sync_stats.h`)

### DIFF
--- a/layers/sync/sync_stats.cpp
+++ b/layers/sync/sync_stats.cpp
@@ -1,0 +1,65 @@
+/* Copyright (c) 2024 The Khronos Group Inc.
+ * Copyright (c) 2024 Valve Corporation
+ * Copyright (c) 2024 LunarG, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "sync_stats.h"
+
+#if VVL_ENABLE_SYNCVAL_STATS != 0
+#include "sync_commandbuffer.h"
+#include "utils/vk_layer_utils.h"
+
+namespace syncval_stats {
+
+void Value32::Update(uint32_t new_value) { u32.store(new_value); }
+uint32_t Value32::Add(uint32_t n) { return u32.fetch_add(n); }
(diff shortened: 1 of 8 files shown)


layers/sync/sync_stats.cpp
```diff
@@ -0,0 +1,64 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "sync_stats.h"
+
+#if ENABLE_SYNCVAL_STATS != 0
```
spencer-lunarg, line 20:
```suggestion
#if VVL_ENABLE_SYNCVAL_STATS != 0
```

layers/sync/sync_stats.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+std::string Stats::CreateReport() {
+    uint32_t handle_record = handle_record_counter.value.u32;
+    uint64_t handle_record_memory = handle_record * sizeof(HandleRecord);
+
+    uint32_t handle_record_max = handle_record_counter.max_value.u32;
+    uint64_t handle_record_max_memory = handle_record_max * sizeof(HandleRecord);
+
+    std::ostringstream str;
+    str << "HandleRecord count = " << handle_record << "\n";
```
spencer-lunarg, line 56: 100% suggestion, I have zero idea which is better in practice

```
    str << "HandleRecord:\n";
    str << "\tcount = " << handle_record << "\n";
    str << "\tmemory = " << handle_record_memory << " bytes\n";
    str << "\tmax_count = " << handle_record_max << "\n";
    str << "\tmax_memory = " << handle_record_max_memory << " bytes\n";
```

layers/utils/vk_layer_utils.h
```diff
@@ -521,11 +521,21 @@ const typename T::value_type *DataOrNull(const T &container) {
...
 }
 
-
 // Workaround for static_assert(false) before C++ 23 arrives
 // https://en.cppreference.com/w/cpp/language/static_assert
 // https://cplusplus.github.io/CWG/issues/2518.html
 template <typename>
 inline constexpr bool dependent_false_v = false;
 
+// Until C++ 26 std::atomic<T>::fetch_max arrives
```
spencer-lunarg, line 530: image

spencer-lunarg (approved):

layers/utils/vk_layer_utils.h
artem-lunarg: I agree, c++ 26 doesn't sound too bad.

layers/sync/sync_stats.cpp
artem-lunarg: This looks better

layers/sync/sync_stats.cpp
artem-lunarg: Updated

Issue #8204, opened 2024-06-27
VUID-VkSwapchainCreateInfoKHR-presentMode-01281


water-chika:
**Describe the situation in which you encountered the missing validation**
Issue description:
When I try to run an app that will create a swapchain with immediate present mode. But on Wayland, it may does not support immediate present mode. Then it will crash because it will try to access a NULL pointer.

**Valid Usage IDs requested**
Please include the valid usage IDs for the checks you are requesting:
VUID-VkSwapchainCreateInfoKHR-presentMode-01281

**Additional Context**

spencer-lunarg:
We are already checking for `VUID-VkSwapchainCreateInfoKHR-presentMode-01281` but curious what is happening. Will try to run some tests locally with Wayland and `VK_PRESENT_MODE_IMMEDIATE_KHR` to see if I can reproduce

spencer-lunarg:
ok, so this is a bug, we have this check, but it was accidentally moved behind a `if VK_EXT_swapchain_maintenance1` check, will get some tests in and fix this

spencer-lunarg:
trying to write tests, just want to point out that Wayland can support `VK_PRESENT_MODE_IMMEDIATE_KHR`, looking at the mesa source code it is

```c
if (display.tearing_control_manager)
    present_modes[present_modes_count++] = VK_PRESENT_MODE_IMMEDIATE_KHR;
```

which depends on if your compositor supports https://wayland.app/protocols/tearing-control-v1

spencer-lunarg:
@water-chika thanks again for raising this

The new layer are being built and can be grabbed from https://github.com/KhronosGroup/Vulkan-ValidationLayers/actions/runs/9713251358 when done

also this will now be fixed in the new SDK release

PR #8208, opened 2024-06-27
layers: Fix PresentMode check being hidden

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8204

This check was in `ValidateSwapchainPresentModesCreateInfo` which only got called if `VK_EXT_swapchain_maintenance1` was set

### DIFF
--- a/layers/core_checks/cc_wsi.cpp
+++ b/layers/core_checks/cc_wsi.cpp
@@ -68,60 +68,49 @@ static VkImageCreateInfo GetSwapchainImpliedImageCreateInfo(const VkSwapchainCre
     return result;
 }
 
+// Validate VkSwapchainPresentModesCreateInfoEXT data
 bool CoreChecks::ValidateSwapchainPresentModesCreateInfo(VkPresentModeKHR present_mode, const Location &create_info_loc,
                                                          const VkSwapchainCreateInfoKHR &create_info,
+                                                         const std::vector<VkPresentModeKHR> &present_modes,
                                                          const vvl::Surface *surface_state) const {
     bool skip = false;
-    std::vector<VkPresentModeKHR> present_modes{};
-    if (surface_state) {
-        present_modes = surface_state->GetPresentModes(physical_device, create_info_loc, this);
-    } else if (IsExtEnabled(instance_extensions.vk_google_surfaceless_query)) {
-        present_modes = physical_device_state->surfaceless_query_state.present_modes;
-    }
-
-    if (std::find(present_modes.begin(), present_modes.end(), present_mode) == present_modes.end()) {
-        skip |= LogError("VUID-VkSwapchainCreateInfoKHR-presentMode-01281", device, create_info_loc,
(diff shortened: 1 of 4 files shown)


layers/core_checks/cc_wsi.cpp
```diff
@@ -68,61 +68,50 @@ static VkImageCreateInfo GetSwapchainImpliedImageCreateInfo(const VkSwapchainCre
...
-    }
-
-    if (std::find(present_modes.begin(), present_modes.end(), present_mode) == present_modes.end()) {
-        skip |= LogError("VUID-VkSwapchainCreateInfoKHR-presentMode-01281", device, create_info_loc,
-                         "called with a non-supported presentMode (%s).", string_VkPresentModeKHR(present_mode));
-    }
-
-    // Validate VkSwapchainPresentModesCreateInfoEXT data
     auto swapchain_present_modes_ci = vku::FindStructInPNextChain<VkSwapchainPresentModesCreateInfoEXT>(create_info.pNext);
-    if (swapchain_present_modes_ci) {
```
spencer-lunarg, line 89: I just returned early and shifted everything in

artem-lunarg (approved):
LGTM. Good fix, actually when I read the `ValidateSwapchainPresentModesCreateInfo` name it suggests it's only about `VkSwapchainPresentModesCreateInfoEXT`, so makes everything more logical.

tests/unit/wsi.cpp
```diff
@@ -3468,4 +3468,51 @@ TEST_F(NegativeWsi, SwapchainPresentModeInfoImplicit) {
...
+    AddSurfaceExtension();
+    RETURN_IF_SKIP(Init());
+    RETURN_IF_SKIP(InitSurface());
+
+    uint32_t count;
+    vk::GetPhysicalDeviceSurfacePresentModesKHR(gpu(), m_surface, &count, nullptr);
+    std::vector<VkPresentModeKHR> present_modes(count);
+    vk::GetPhysicalDeviceSurfacePresentModesKHR(gpu(), m_surface, &count, present_modes.data());
+    for (auto present_mode : present_modes) {
+        if (present_mode == VK_PRESENT_MODE_IMMEDIATE_KHR) {
```
artem-lunarg, line 3484: The ticket was about immediate mode, but for example on windows both amd and nvidia support it. Can we have some version less well supported modes on windows. MAILBOX is not reported on AMD on my machine, or one of the more esoteric modes like `VK_PRESENT_MODE_SHARED_DEMAND_REFRESH_KHR` in case you can specify it without extension.

artem-lunarg: I can submit it separately if you prefer this.

spencer-lunarg: yes, I tested this by hacking at test icd, but more WSI tests improvements would be AMAZING!!

spencer-lunarg: @artem-lunarg I merged this so the issue is fixed, everything else we add is just an enhancement

Issue #8202, opened 2024-06-27
Crash with dynamic rendering + advanced blend + graphics pipeline library


ShabbyX:
When VK_EXT_graphics_pipeline_library, VK_EXT_blend_operation_advanced and VK_KHR_dynamic_rendering are simultaneously used, the following piece of code in `cc_pipeline_graphics.cpp` crashes during draw-time validation:

```
            if (advance_blend) {
                const uint32_t color_attachment_count = pipeline.rendering_create_info
                                                            ? pipeline.rendering_create_info->colorAttachmentCount
                                                            : subpass_desc->colorAttachmentCount;
```

The crux of the issue is that `pipeline` is created out of libraries. The libraries have rendering create info on them, but the linked pipeline neither has a rendering create info nor a render pass. The code above assumes there must be a render pass because there is no rendering create info, which leads to crash.

The color attachment count should be taken from the fragment output library per https://docs.vulkan.org/spec/latest/chapters/pipelines.html#pipelines-graphics-subsets-fragment-output

ShabbyX:
Should be pretty easy to repro in a test, but if you want something ready to test, take this work-in-progress ANGLE change: https://chromium-review.googlesource.com/c/angle/angle/+/5637155 and run:

```
$ ./angle_deqp_khr_gles31_tests --use-angle=swiftshader --deqp-case=KHR-GLES31.core.blend_equation_advanced.blend_all.*
```

Note that support for VK_KHR_dynamic_rendering_local_read was recently added to SwiftShader for testing purposes, so you'd need to sync to ToT of ANGLE, and don't forget `gclient sync`!

spencer-lunarg:
thanks, taking a look now, but looking at the code more now, the other issue is we split up pipeline and mix-and-match the renderpass and dynamic-rendering case and it opens to not catching a case like this... going to slightly clean things up to be 2 separate path (with reference, not pointer) so it is clear where in the code a future check is added

PR #8210, opened 2024-06-28
Cleanup RenderPass logic for Pipeline

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8202

The main goal was to try and make it clear in the pipeline validation where/when we have a renderpass/dynamic rendering/GPL being used

The core issue was we were mixing up

- `rp_state == NULL` when using something like GPL Vertex Input and there is no state **OR** there is a bad handle to VkRenderPass
- `pipeline.GraphicsCreateInfo().renderPass == VK_NULL_HANDLE` when the passed in value is null (which can be from Dynamic Rendering OR GPL)
- `rp_state && rp_state.UsesDynamicRendering()` which is when the pipeline passes in null, but we created a `vvl::RenderPass` at `vkCmdBeginRendering` and so there is a state object

Core issue is that `vvl::RenderPass` is used for dynamic rendering where there might not be an actual `VkRenderPass VkHandle()` under it

### DIFF
--- a/layers/core_checks/cc_pipeline_graphics.cpp
+++ b/layers/core_checks/cc_pipeline_graphics.cpp
@@ -68,14 +68,12 @@ bool CoreChecks::PreCallValidateCreateGraphicsPipelines(VkDevice device, VkPipel
 bool CoreChecks::ValidateGraphicsPipeline(const vvl::Pipeline &pipeline, const Location &create_info_loc) const {
     bool skip = false;
     vku::safe_VkSubpassDescription2 *subpass_desc = nullptr;
-
     const auto rp_state = pipeline.RenderPassState();
-    if (!rp_state) {
-        // Things such as GPL Vertex Input will not have a render pass state.
-    } else if (rp_state->UsesDynamicRendering()) {
+
+    if (rp_state && rp_state->UsesDynamicRendering()) {
         skip |= ValidateGraphicsPipelineExternalFormatResolveDynamicRendering(pipeline, create_info_loc);
-    } else {
-        const auto subpass = pipeline.Subpass();
+    } else if (rp_state && !rp_state->UsesDynamicRendering()) {
+        const uint32_t subpass = pipeline.Subpass();
         if (subpass >= rp_state->create_info.subpassCount) {
             skip |=
                 LogError("VUID-VkGraphicsPipelineCreateInfo-renderPass-06046", rp_state->Handle(),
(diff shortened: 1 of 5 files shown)


layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -481,17 +495,18 @@ bool CoreChecks::ValidateGraphicsPipelineVertexInputState(const vvl::Pipeline &p
...
     if (!pipeline.IsRenderPassStateRequired()) return skip;
 
-    if (pipeline.GraphicsCreateInfo().renderPass == VK_NULL_HANDLE) {
+    if (pipeline.IsRenderPassNull()) {
         if (!enabled_features.dynamicRendering) {
             skip |= LogError("VUID-VkGraphicsPipelineCreateInfo-dynamicRendering-06576", device,
                              create_info_loc.dot(Field::renderPass), "is NULL, but the dynamicRendering feature was not enabled");
         }
     } else if (!pipeline.RenderPassState()) {
+        // invalid render pass
```
arno-lunarg, line 509: We get here is render pas is **not** null, so is it invalid?

layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -1991,6 +1906,77 @@ bool CoreChecks::ValidateGraphicsPipelineRasterizationState(const vvl::Pipeline
...
+        }
+    }
+
+    // If subpass uses color attachments, pColorBlendState must be valid pointer
+    if (pipeline.fragment_output_state && !pipeline.fragment_output_state->color_blend_state &&
+        !pipeline.IsColorBlendStateDynamic()) {
+        uint32_t color_attachment_count = 0;
+        for (uint32_t i = 0; i < subpass_desc.colorAttachmentCount; ++i) {
+            if (subpass_desc.pColorAttachments[i].attachment != VK_ATTACHMENT_UNUSED) {
+                ++color_attachment_count;
```
arno-lunarg, line 1964:
```suggestion
                has_color_attachment_count = true;
                break;
```
or print out `color_attachment_count` in error message

layers/core_checks/cc_pipeline_graphics.cpp
spencer-lunarg: so again welcome to this awful situation

`null handle` != `null state object`

`null handle` - that is what dynamic state has
`null state object` - we failed the `Get<vvl::RenderPass>` call (because it was either null OR invalid)

The above `if` check for the `null handle`, this now checks if we had an `invalid state object` (from a bad `VkRenderPass` handle)

layers/core_checks/cc_pipeline_graphics.cpp
spencer-lunarg: ya, I just moved this, but agree, let me improve this while here

layers/core_checks/cc_pipeline_graphics.cpp
spencer-lunarg: ... going to add a comment now btw

PR #8216, opened 2024-06-28
layers: Ignore pipeline state with dynamic state

spencer-lunarg:
There were a few places were we were checking the value of the pipeline state, but it should have been ignore because the corresponding `VK_DYNAMIC_STATE_*` was used


artem-lunarg (approved):
LGTM 🌴🍸

PR #8217, opened 2024-06-28
build: Update to header 1.3.289

mikes-lunarg:


spencer-lunarg (approved):

scripts/known_good.json
```diff
@@ -7,15 +7,18 @@
             "sub_dir": "Vulkan-Headers",
             "build_dir": "Vulkan-Headers/build",
             "install_dir": "Vulkan-Headers/build/install",
-            "commit": "v1.3.288"
+            "cmake_options": [
+                "-DVULKAN_HEADERS_ENABLE_MODULE=OFF"
```
spencer-lunarg, line 11: this wasn't off by default? What happens if it was on? Where does it blow up

mikes-lunarg: It usually doesn't hurt but the latest version of the module crashes the version of MSVC that is in the github actions runner: https://github.com/KhronosGroup/Vulkan-Headers/commit/e3c37e6e184a232e10b01dff5a065ce48c047f88

PR #8219, opened 2024-07-01
layers: Add VK_KHR_maintenance7

spencer-lunarg:
Most of `VK_KHR_maintenance7` for validation is just allowing things that were not previously allowed, but now drivers will support (if using the extension/feature)

### DIFF
--- a/layers/core_checks/cc_descriptor.cpp
+++ b/layers/core_checks/cc_descriptor.cpp
@@ -4002,16 +4002,6 @@ bool CoreChecks::PreCallValidateCreatePipelineLayout(VkDevice device, const VkPi
                          sum_all_stages[VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER], phys_dev_props.limits.maxDescriptorSetUniformBuffers);
     }
 
-    // Dynamic uniform buffers
-    if (sum_all_stages[VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER_DYNAMIC] > phys_dev_props.limits.maxDescriptorSetUniformBuffersDynamic) {
-        skip |= LogError("VUID-VkPipelineLayoutCreateInfo-descriptorType-03030", device, error_obj.location,
-                         "sum of dynamic uniform buffer bindings among all stages (%" PRIu32
-                         ") exceeds device "
-                         "maxDescriptorSetUniformBuffersDynamic limit (%" PRIu32 ").",
-                         sum_all_stages[VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER_DYNAMIC],
-                         phys_dev_props.limits.maxDescriptorSetUniformBuffersDynamic);
-    }
-
     // Storage buffers
     if (sum_all_stages[VK_DESCRIPTOR_TYPE_STORAGE_BUFFER] > phys_dev_props.limits.maxDescriptorSetStorageBuffers) {
         skip |= LogError("VUID-VkPipelineLayoutCreateInfo-descriptorType-03031", device, error_obj.location,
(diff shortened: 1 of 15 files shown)


layers/core_checks/cc_descriptor.cpp
```diff
@@ -4021,14 +4011,59 @@ bool CoreChecks::PreCallValidateCreatePipelineLayout(VkDevice device, const VkPi
...
+            skip |= LogError("VUID-VkPipelineLayoutCreateInfo-maintenance7-10004", device, error_obj.location,
+                             "sum of dynamic storage buffer bindings among all stages (%" PRIu32
+                             ") exceeds device "
+                             "maxDescriptorSetTotalStorageBuffersDynamic limit (%" PRIu32 ").",
+                             sum_all_storage_buffer_dynamic,
+                             phys_dev_ext_props.maintenance7_props.maxDescriptorSetTotalStorageBuffersDynamic);
+        }
+
+        // prevent overflow
+        const uint64_t total_sum =
```
artem-lunarg, line 4038: Suggestion not worry about overflow here. 4 billion of descriptors, that's up to 100Gb memory to store only descriptors, not counting the resources themselves. The problem here if we decide to do this in this place, the other places should also be updated (will have similar probability of overflows): 

when separate `sum_all_xx` values are so large, so their sum overflows, then separate values itself can overflow with similar probability (if there are 3bln of descriptors of a single type, then  it does not look unlikely to have 5lbn of those descriptors), and those computations should also be done in 64 bit. I guess API choice of uint32 in the property structs suggests that's practical upper limit.

layers/stateless/sl_instance_device.cpp
```diff
@@ -908,3 +908,33 @@ bool StatelessValidation::manual_PreCallValidateSetDebugUtilsObjectTagEXT(VkDevi
...
+                                                                             VkPhysicalDeviceProperties2 *pProperties,
+                                                                             const ErrorObject &error_obj) const {
+    bool skip = false;
+    const auto *api_props_lists = vku::FindStructInPNextChain<VkPhysicalDeviceLayeredApiPropertiesListKHR>(pProperties->pNext);
+    if (api_props_lists && api_props_lists->pLayeredApis) {
+        for (uint32_t i = 0; i < api_props_lists->layeredApiCount; i++) {
+            if (const auto *api_vulkan_props = vku::FindStructInPNextChain<VkPhysicalDeviceLayeredApiVulkanPropertiesKHR>(
+                    api_props_lists->pLayeredApis[i].pNext)) {
+                const VkBaseOutStructure *current =
+                    reinterpret_cast<const VkBaseOutStructure *>(api_vulkan_props->properties.pNext);
```
artem-lunarg, line 922: nit: something I annoyed Arno in the past, in c++ it's enough static_cast<const type*>(const void*), reinterpreset_cast is more stronk.

artem-lunarg (approved):
LGTM. Minor suggestions. 🏄‍♂️🍹

layers/core_checks/cc_descriptor.cpp
spencer-lunarg: `maxDescriptorSetTotalStorageBuffersDynamic` was added for certain vendors, but I assume it will be the same as `maxDescriptorSetStorageBuffers` for most people, which can be `UINT32_MAX` (https://www.vulkan.gpuinfo.org/listreports.php?limit=maxDescriptorSetStorageBuffers&value=4294967295&platform=all)

so adding these together will for sure overflow... I expect NO ONE to hit this limit, but could see someone getting a strange false positive from the overflow

layers/stateless/sl_instance_device.cpp
spencer-lunarg: I just blindly took from safe_struct in VUL, make it to `static_cast`

layers/core_checks/cc_descriptor.cpp
artem-lunarg: I could misunderstood the code, if we add MAX values then that's definitely true.

PR #8221, opened 2024-07-01
sync: Report descriptor resources

artem-lunarg:
One more PR to track resource handles for submit time validation. This adds the resource handles referenced by the descriptors. New option `sync_report_descriptor_resources` controls this feature (disabled by default).

The previous PRs added handle reporting for handles specified directly through API commands, as in `vkCmdCopyBuffer(buffer_a, buffer_b)`. Those handles have fixed and small memory footprint and they are always collected.

Shader can reference a lot of resources and it can take significant amount of memory depending on the application. That's the reason configuration option is introduced. Some numbers for the required amount of memory for descriptor resource handles from the test sessions: CS2: 70 Mb, Doom eternal: 600 Mb.

The next PR will use the information about tracked handles to detect a specific resource that caused hazard during submit time validation.

### DIFF
--- a/layers/layer_options.cpp
+++ b/layers/layer_options.cpp
@@ -23,6 +23,8 @@
 #include "gpu/core/gpu_settings.h"
 #include "error_message/logging.h"
 
+#include "sync/sync_settings.h"
+
 // Include new / delete overrides if using mimalloc. This needs to be include exactly once in a file that is
 // part of the VVL but not the layer utils library.
 #if defined(USE_MIMALLOC) && defined(_WIN64)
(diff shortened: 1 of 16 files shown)


layers/sync/sync_commandbuffer.cpp
```diff
@@ -493,6 +493,9 @@ void CommandBufferAccessContext::RecordDispatchDrawDescriptorSet(VkPipelineBindP
                         } else {
                             current_context_->UpdateAccessState(*img_view_state, sync_index, SyncOrdering::kNonAttachment, tag);
                         }
+                        if (sync_state_->syncval_settings.report_descriptor_resources) {
```
artem-lunarg, line 496: Main part of the PR: register handle if option is enabled.

layers/layer_options.cpp
```diff
@@ -91,6 +93,10 @@ const char *VK_LAYER_GPUAV_VMA_LINEAR_OUTPUT = "gpuav_vma_linear_output";
 const char *VK_LAYER_GPUAV_DEBUG_VALIDATE_INSTRUMENTED_SHADERS = "gpuav_debug_validate_instrumented_shaders";
 const char *VK_LAYER_GPUAV_DEBUG_DUMP_INSTRUMENTED_SHADERS = "gpuav_debug_dump_instrumented_shaders";
 
+// SyncVal
+// ---
+const char *VK_LAYER_SYNCVAL_REPORT_DESCRIPTOR_RESOURCES = "sync_report_descriptor_resources";
```
artem-lunarg, line 98: Thanks @spencer-lunarg for the help with configuration. In the end I decided to remove "artem" from variables.

spencer-lunarg: we can make `VK_LAYER_SYNC_ARTEM_*` namespace for the super secret hidden cool features!

spencer-lunarg (approved):

PR #8225, opened 2024-07-01
layers: Fix VUID 09680 after spec update

ziga-lunarg:


artem-lunarg (approved):
LGTM. There is similar change here https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8219/files#diff-f8843b8d19e072640e9193c84a85ff9102adf454fd4b4c7970d17f34dd0a4860R843. It looks like this version changes both the first and non-first subpass cases. Spencer's changes updates name with KHR suffix in one place.

PR #8228, opened 2024-07-02
scripts: Print VUID text correctly in vk_validation_stats.py

r-potter:
In html and csv output cases the vk_validation_stats.py script will print out the VUID text as the last column. However, using text output mode the "ext" field was printed previously.


arno-lunarg (approved):

arno-lunarg:
Thank you for the fix!

PR #8235, opened 2024-07-03
layers: Miscellaneous fixes

aqnuep:
This change contains various fixes, including fixes for the following problems:
- incorrect query entry point used by vkGetCalibratedTimestampsKHR validation
- incorrect generator script logic for the malfunctioning generator script caching
- incorrect handling of empty structextends lists
- incorrect assumptions about entry points with VkObjectType parameters
- API parameterization changes for the Test ICD
- incorrect test cases still expecting "stateless" SPIR-V validation at pipeline create call
- incorrect test case dependencies

### DIFF
--- a/layers/state_tracker/pipeline_state.h
+++ b/layers/state_tracker/pipeline_state.h
@@ -389,6 +389,19 @@ class Pipeline : public StateObject {
         return rt->sType;
     }
 
+    const void *GetCreateInfoPNext() const {
+        const auto *gfx = std::get_if<vku::safe_VkGraphicsPipelineCreateInfo>(&create_info);
+        if (gfx) {
+            return gfx->pNext;
+        }
+        const auto *cmp = std::get_if<vku::safe_VkComputePipelineCreateInfo>(&create_info);
+        if (cmp) {
+            return cmp->pNext;
+        }
+        const auto *rt = std::get_if<vku::safe_VkRayTracingPipelineCreateInfoCommon>(&create_info);
+        return rt->pNext;
+    }
+
     bool BlendConstantsEnabled() const { return fragment_output_state && fragment_output_state->blend_constants_enabled; }
 
     bool SampleLocationEnabled() const { return fragment_output_state && fragment_output_state->sample_location_enabled; }
--- a/layers/vulkan/generated/layer_chassis_dispatch.cpp
+++ b/layers/vulkan/generated/layer_chassis_dispatch.cpp
@@ -334,7 +334,7 @@ void WrapPnextChainHandles(ValidationObject* layer_data, const void* pNext) {
     }
 }
 
-static bool NotDispatchableHandle(VkObjectType object_type) {
+[[maybe_unused]] static bool NotDispatchableHandle(VkObjectType object_type) {
     switch (object_type) {
         case VK_OBJECT_TYPE_INSTANCE:
         case VK_OBJECT_TYPE_PHYSICAL_DEVICE:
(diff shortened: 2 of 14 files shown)


scripts/generators/base_generator.py
```diff
@@ -36,7 +36,7 @@
 
 # Helpers to keep things cleaner
 def splitIfGet(elem, name):
-    return elem.get(name).split(',') if elem.get(name) is not None else None
+    return elem.get(name).split(',') if elem.get(name) is not None and elem.get(name) != '' else None
```
spencer-lunarg, line 39: the xml has empty string `name` tags now?

spencer-lunarg (approved):

layers/vulkan/generated/layer_chassis_dispatch.cpp
```diff
@@ -2494,12 +2494,7 @@ VkResult DispatchSetPrivateData(VkDevice device, VkObjectType objectType, uint64
     auto layer_data = GetLayerDataPtr(GetDispatchKey(device), layer_data_map);
     if (!wrap_handles)
         return layer_data->device_dispatch_table.SetPrivateData(device, objectType, objectHandle, privateDataSlot, data);
-    {
-        if (NotDispatchableHandle(objectType)) {
-            objectHandle = layer_data->Unwrap(objectHandle);
```
spencer-lunarg, line 2499: this being removed seems wrong and all the `NegativeTooling.PrivateDataDestroyHandle` tests are failing now because of it I think

scripts/generators/base_generator.py
aqnuep: Empty `structextends` attribute, which can result from if you generate a target API set where the struct has a `structextends` attribute, but none are actually present in the target API. It does happen with some structures today in Vulkan SC, and the pre-refactored script handled it correctly. But it is not specific to Vulkan SC, Vulkan may similarly have in the future a scenario when a structure is enabled to extend a Vulkan SC-only structure, for example, and that's the only structure in the `structextends` attribute, resulting in `structextends=""` for Vulkan.

The new code simply doesn't handle the case when the attribute ends up being empty after the target API stripping logic  done by the spec tools (which, again, was handled correctly by `base_generator.py` before recent refactorings).

layers/vulkan/generated/layer_chassis_dispatch.cpp
aqnuep: Sorry about that. It shouldn't have happened, I probably made a mistake there.

However, the problem with the original script code is that if it detects an entry point with any `VkObjectType` typed parameter, then it generates this code, even though it's not really a general rule. For example, Vulkan SC has this entry point:

```c
VkResult vkGetPhysicalDeviceRefreshableObjectTypesKHR(
    VkPhysicalDevice                            physicalDevice,
    uint32_t*                                   pRefreshableObjectTypeCount,
    VkObjectType*                               pRefreshableObjectTypes);
```

This certainly shouldn't have that hard-coded code injected into it that is only meant to apply to the private data entry points.

Anyway, I'll fix the generator script.

scripts/generators/base_generator.py
spencer-lunarg: sounds good... I still want to get `base_generator.py` in the header as an edge case like this, is nice to have solved in a single spot as someone editing the code gen would likely not have seen this

> which, again, was handled correctly by base_generator.py before recent refactorings

this is something stopping me from putting it in the headers, I would ideally like to add some testing to not break this with the ever expanding ways to correctly parse the XML

layers/state_tracker/pipeline_state.h
```diff
@@ -389,6 +389,19 @@ class Pipeline : public StateObject {
         return rt->sType;
     }
 
+    const void *GetCreateInfoPNext() const {
```
spencer-lunarg, line 392: this is a nice helper, will try and get a change soon to use it around the current code (also partly so no one accidentally deletes this because they don't think its being used anywhere)

PR #8240, opened 2024-07-05
gpu: Add GPU to CPU buffer atomic exchange read test

arno-lunarg:
Nothing to see here, just want to run this test on CI machines

### DIFF
--- a/tests/unit/gpu_av_positive.cpp
+++ b/tests/unit/gpu_av_positive.cpp
@@ -21,6 +21,11 @@
 #include "../framework/gpu_av_helper.h"
 #include "../../layers/gpu/shaders/gpu_shaders_constants.h"
 
+#include <atomic>
+#include <chrono>
+#include <thread>
+#include <cinttypes>
+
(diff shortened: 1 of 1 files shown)


tests/unit/gpu_av_positive.cpp
```diff
@@ -1732,3 +1736,239 @@ TEST_F(PositiveGpuAV, PipelineLayoutMixing) {
...
+void main() {
+    const uint values [8] = {
+      0xBBAADDCC,
+      0xFFFFFFFF,
+      0xAABBCCDD,
+      0xEEFFBBAA,
+      0x00112233,
+      0x44556677,
+      0x8899AABB,
+      0xCCDDEEFF
```
filnet, line 1766: You are missing `0xDEADBEEF` and a bunch others : https://en.wikipedia.org/wiki/Hexspeak
Have a good weekend.

arno-lunarg: Ahah thanks for the hint ^^

PR #8242, opened 2024-07-05
Remove legacy obsolete-file handling from the GN build

### LINKED ISSUE #8238: BUILD.gn "vulkan_clean_old_validation_layer_objects" removing files unrelated to VVL
`action("vulkan_clean_old_validation_layer_objects") `

is nuking files by glob which can match files unrelated to VVL

https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/e0f6324b75b14466dea699ba4979cfe2096ee5e9/BUILD.gn#L40-L42

This is creating a race condition between build rules resulting in flaky build failures in ANGLE: https://g-issues.angleproject.org/issues/350816748#comment5

@mark-lunarg

mark-lunarg:
Due to recent changes the remove-files script created unexpected interactions with other layer build artifacts. The script and build action are no longer needed and so have been removed.

Fixes #8238


artem-lunarg (approved):

arno-lunarg:
He answered the VVL call...
batman

spencer-lunarg (approved):

BUILD.gn
```diff
@@ -20,28 +20,7 @@ if (is_win) {
   vulkan_undefine_configs += [ "//build/config/win:unicode" ]
 }
 
-vulkan_gen_dir = "$target_gen_dir/$vulkan_gen_subdir"
-raw_vulkan_gen_dir = rebase_path(vulkan_gen_dir, root_build_dir)
-
 vulkan_data_dir = "$root_out_dir/$vulkan_data_subdir"
```
romanl-g, line 23: Oops this breaks build because this var is now unused:
```
/b/s/w/ir/cache/builder/angle/buildtools/linux64/gn gen /b/s/w/ir/cache/builder/angle/out/Release --check
  -> returned 1
ERROR at //third_party/vulkan-validation-layers/src/BUILD.gn:23:19: Assignment had no effect.
vulkan_data_dir = "$root_out_dir/$vulkan_data_subdir"
                  ^----------------------------------
You set the variable "vulkan_data_dir" here and it was unused before it went
out of scope.
```

spencer-lunarg: I guess 

1. @mark-lunarg can you help get a fix (or @romanl-g if you can)
2. Why is this not failing in CI (https://github.com/KhronosGroup/Vulkan-ValidationLayers/actions/runs/9812111518/job/27095568479) - we don't have the same warning-as-error flag?

romanl-g: Looks like our builds run `gn gen ... --check`, maybe it's only a failure with `--check`?

The fix should be just removing that line given that it's a now unused var.

spencer-lunarg: checked with https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/dd72a8050aac28e19a1603329c4f4f62e339a6f9 and seems we are doing `--check`

I can't get an error locally neither to reproduce

romanl-g: Sorry, this appears to be Android-specific. This should be sufficient to repro:
```
target_os = "android"
dcheck_always_on=true
```

romanl-g: Would it be possible for you to also run gn gen for Android on CI?

romanl-g:
@mikes-lunarg thanks for all the fixes in #8254 !

PR #8243, opened 2024-07-05
sync: Make QueueId deterministic

artem-lunarg:
This allows syncval to have deterministic `QueueId` between runs that simplifies debugging. It can be frustrating when during long debug session you learn how QueueId numbers map to specific queues and on the next run the mapping changes.

`vvl::Queue` adds `queue_index` field that is the index of the queue within a queue family.

### DIFF
--- a/layers/state_tracker/queue_state.h
+++ b/layers/state_tracker/queue_state.h
@@ -85,10 +85,11 @@ struct PreSubmitResult {
 
 class Queue: public StateObject {
   public:
-    Queue(ValidationStateTracker &dev_data, VkQueue handle, uint32_t index, VkDeviceQueueCreateFlags flags,
-          const VkQueueFamilyProperties &queueFamilyProperties)
+    Queue(ValidationStateTracker &dev_data, VkQueue handle, uint32_t family_index, uint32_t queue_index,
+          VkDeviceQueueCreateFlags flags, const VkQueueFamilyProperties &queueFamilyProperties)
         : StateObject(handle, kVulkanObjectTypeQueue),
-          queueFamilyIndex(index),
+          queueFamilyIndex(family_index),
+          queue_index(queue_index),
           flags(flags),
           queueFamilyProperties(queueFamilyProperties),
           dev_data_(dev_data) {}
@@ -114,7 +115,12 @@ class Queue: public StateObject {
     // Helper that combines Notify and Wait
     void NotifyAndWait(const Location &loc, uint64_t until_seq = kU64Max);
 
+    // Queue family index. As queueFamilyIndex parameter in vkGetDeviceQueue.
     const uint32_t queueFamilyIndex;
+
+    // Index of the queue within a queue family. As queueIndex parameter in vkGetDeviceQueue.
+    const uint32_t queue_index;
+
     const VkDeviceQueueCreateFlags flags;
     const VkQueueFamilyProperties queueFamilyProperties;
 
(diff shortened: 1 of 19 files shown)


layers/state_tracker/state_tracker.cpp
```diff
@@ -1046,7 +1047,7 @@ void ValidationStateTracker::PostCreateDevice(const VkDeviceCreateInfo *pCreateI
                     DispatchGetDeviceQueue(device, queue_info.queue_family_index, i, &queue);
                 }
                 assert(queue != VK_NULL_HANDLE);
-                Add(CreateQueue(queue, queue_info.queue_family_index, queue_info.flags,
+                Add(CreateQueue(queue, queue_info.queue_family_index, i, queue_info.flags,
```
artem-lunarg, line 1050: That's the place where we specify queue index (`i`).

layers/state_tracker/state_tracker.cpp
```diff
@@ -1522,26 +1523,27 @@ void ValidationStateTracker::PostCallRecordGetFenceStatus(VkDevice device, VkFen
...
 
-        Add(CreateQueue(queue, queue_family_index, flags, queue_family_properties_list[queue_family_index]));
+        Add(CreateQueue(queue, queue_family_index, queue_index, flags, queue_family_properties_list[queue_family_index]));
     }
 }
 
 void ValidationStateTracker::PostCallRecordGetDeviceQueue(VkDevice device, uint32_t queueFamilyIndex, uint32_t queueIndex,
                                                           VkQueue *pQueue, const RecordObject &record_obj) {
-    RecordGetDeviceQueueState(queueFamilyIndex, {}, *pQueue);
+    RecordGetDeviceQueueState(queueFamilyIndex, queueIndex, {}, *pQueue);
```
artem-lunarg, line 1541: and also when `vkGetDeviceQueue` is called.

layers/sync/sync_validation.cpp
```diff
@@ -605,9 +605,21 @@ void SyncValidator::PostCreateDevice(const VkDeviceCreateInfo *pCreateInfo, cons
     // The state tracker sets up the device state
     StateTracker::PostCreateDevice(pCreateInfo, loc);
 
-    ForEachShared<vvl::Queue>([this](const std::shared_ptr<vvl::Queue> &queue_state) {
-        queue_sync_states_.emplace_back(std::make_shared<QueueSyncState>(queue_state, queue_id_limit_++));
-    });
+    // Returns queues in the same order as advertised by the driver.
+    // This allows to have deterministic QueueId between runs that simplifies debugging.
+    auto get_sorted_queues = [this]() {
```
artem-lunarg, line 610: This parts sorts queue at first by queue family index then by index within a queue family, so the order is the same for the given gpu/drivers.

layers/state_tracker/queue_state.h
```diff
@@ -115,6 +116,7 @@ class Queue: public StateObject {
     void NotifyAndWait(const Location &loc, uint64_t until_seq = kU64Max);
 
     const uint32_t queueFamilyIndex;
+    const uint32_t queue_index;
```
spencer-lunarg, line 119: 1. maybe worth adding a comment about the difference between this and `queueFamilyIndex`
2. (not asking you to clean up here in the PR) but it would be nice to have `queue_family_index` ... maybe a future small PR to unify these member names will help as well

spencer-lunarg (approved):

layers/state_tracker/queue_state.h
artem-lunarg: good points

artem-lunarg: updated

PR #8245, opened 2024-07-06
layers: Reorganize renderpass framebuffer checks

spencer-lunarg:
The Renderpass/Framebuffer creations checks were getting long and a tad out-of-hand

This does 3 things

- Moves more checks to stateless where possible
- Groups logical extension logic to own function
- Pass by reference as much as possible (instead of pointers)


artem-lunarg (approved):

spencer-lunarg:
@r-potter heads up - if you had a change in the work, might have to rebase around this... don't plan to touch this code much anymore and if you also find things worth cleaning up around the framebuffer / renderpass logic, please don't hesitate

Issue #8232, opened 2024-07-02
VUID-vkCmdDrawMultiIndexedEXT-dynamicPrimitiveTopologyUnrestricted-07500 bug?


zmike:
When running `ZINK_DEBUG=vvl MESA_LOADER_DRIVER_OVERRIDE=zink ./glcts  -n KHR-GL46.shader_viewport_layer_array.ShaderViewportIndexTestCase` I get this error:

```
VUID-vkCmdDrawMultiIndexedEXT-dynamicPrimitiveTopologyUnrestricted-07500(ERROR / SPEC): msgNum: -272496202 - Validation Error: [ VUID-vkCmdDrawMultiIndexedEXT-dynamicPrimitiveTopologyUnrestricted-07500 ] Object 0: handle = 0x4019bf0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xe7713700000000ae, type = VK_OBJECT_TYPE_PIPELINE; | MessageID = 0xefc209b6 | vkCmdDrawMultiIndexedEXT():  the last primitive topology VK_PRIMITIVE_TOPOLOGY_PATCH_LIST state set by vkCmdSetPrimitiveTopology is not compatible with the pipeline topology VK_PRIMITIVE_TOPOLOGY_TRIANGLE_STRIP. The Vulkan spec states: If the bound graphics pipeline state was created with the VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY dynamic state enabled and the dynamicPrimitiveTopologyUnrestricted is VK_FALSE, then the primitiveTopology parameter of vkCmdSetPrimitiveTopology must be of the same topology class as the pipeline VkPipelineInputAssemblyStateCreateInfo::topology state (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdDrawMultiIndexedEXT-dynamicPrimitiveTopologyUnrestricted-07500)
```

which I believe to be spurious. The bound pipeline is GPL with an input state that sets `VK_PRIMITIVE_TOPOLOGY_PATCH_LIST`, matching the dynamic state value.

This happens on RADV with mesa main and VVL 803ef169856697d93f69b5d01489763a114e81cb

PR #8253, opened 2024-07-09
layers: Fix GPL with tessellation topology

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8232 (@zmike enjoy these quick GPL fixes, they won't last long :laughing: )


artem-lunarg (approved):
GLTM. So in this case input assembly topology takes precedence over shaders? Which shaders can do this, geometry?

spencer-lunarg:
> Which shaders can do this

tessellation... the source of all complications

Issue #8108, opened 2024-06-07
OpLoad wrongly classified as a shader access for images


jrprice:
It looks like VVL treats `OpLoad` of an image handle as an access to the underlying texture storage, which is leading to false positives (e.g. `SYNC-HAZARD-WRITE-AFTER-READ`).

For example, a shader that only queries the dimensions of an image does not actually perform any accesses to that image. The Vulkan spec says:
> OpImageQuerySize [...] return properties of the image descriptor that would be accessed. The image itself is not accessed.

I created this unit test for `sync_val_positive.cpp` which I believe captures the issue:
```

TEST_F(PositiveSyncVal, WriteAfterImageUsedInShaderWithoutAccess) {
    TEST_DESCRIPTION("Test that image queries are not considered to be accesses");
    RETURN_IF_SKIP(InitSyncValFramework());
    RETURN_IF_SKIP(InitState());

    vkt::Buffer copy_source(*m_device, 32 * 32 * 4, VK_BUFFER_USAGE_TRANSFER_SRC_BIT | VK_BUFFER_USAGE_STORAGE_BUFFER_BIT);

    vkt::Image image(*m_device, 32, 32, 1, VK_FORMAT_R8G8B8A8_UNORM, VK_IMAGE_USAGE_STORAGE_BIT | VK_IMAGE_USAGE_TRANSFER_DST_BIT);
    image.SetLayout(VK_IMAGE_LAYOUT_GENERAL);
    vkt::ImageView view = image.CreateView();

    OneOffDescriptorSet descriptor_set(m_device, {
                                                     {0, VK_DESCRIPTOR_TYPE_STORAGE_IMAGE, 1, VK_SHADER_STAGE_COMPUTE_BIT, nullptr},
                                                 });
    descriptor_set.WriteDescriptorImageInfo(0, view, VK_NULL_HANDLE, VK_DESCRIPTOR_TYPE_STORAGE_IMAGE, VK_IMAGE_LAYOUT_GENERAL);
    descriptor_set.UpdateDescriptorSets();

    const char* cs_source = R"glsl(
        #version 450
        layout(set = 0, binding = 0, rgba8) writeonly uniform image2D image;
        void main(){
            uvec2 size = imageSize(image);
        }
    )glsl";
    CreateComputePipelineHelper pipe(*this);
[...]

PR #8256, opened 2024-07-09
sync: Ignore not accessed variables and descriptor-only access

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8108

### DIFF
--- a/layers/sync/sync_commandbuffer.cpp
+++ b/layers/sync/sync_commandbuffer.cpp
@@ -28,6 +28,9 @@
 SyncStageAccessIndex GetSyncStageAccessIndexsByDescriptorSet(VkDescriptorType descriptor_type,
                                                              const spirv::ResourceInterfaceVariable &variable,
                                                              VkShaderStageFlagBits stage_flag) {
+    if (!variable.IsAccessed()) {
+        return SYNC_ACCESS_INDEX_NONE;
+    }
     if (descriptor_type == VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT) {
         assert(stage_flag == VK_SHADER_STAGE_FRAGMENT_BIT);
         return SYNC_FRAGMENT_SHADER_INPUT_ATTACHMENT_READ;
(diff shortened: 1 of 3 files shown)


layers/sync/sync_commandbuffer.cpp
```diff
@@ -49,6 +51,10 @@ SyncStageAccessIndex GetSyncStageAccessIndexsByDescriptorSet(VkDescriptorType de
                descriptor_type == VK_DESCRIPTOR_TYPE_UNIFORM_TEXEL_BUFFER) {
         return stage_accesses.sampled_read;
     } else {
+        if (variable.IsImage() && !variable.IsImageReadFrom()) {
```
artem-lunarg, line 54: Just a point fix for a specific issue.  This entire function probably can be rearranged and probably some other checks can be done in a better way.

layers/sync/sync_commandbuffer.cpp
```diff
@@ -28,6 +28,9 @@
 SyncStageAccessIndex GetSyncStageAccessIndexsByDescriptorSet(VkDescriptorType descriptor_type,
                                                              const spirv::ResourceInterfaceVariable &variable,
                                                              VkShaderStageFlagBits stage_flag) {
+    if (!variable.IsAccessed()) {
```
artem-lunarg, line 31: Not access variables were previously reported as READ (everything not write were considered a read). This can produce another set of false-positives (but not the one reported in the issue), so in theory this will reduce the number of false-positives. But if any issue for someone we already have a next hammer to fix...

spencer-lunarg (approved):

tests/unit/sync_val_positive.cpp
```diff
@@ -1705,3 +1705,43 @@ TEST_F(PositiveSyncVal, WriteAndReadNonOverlappedDynamicUniformBufferRegions2) {
 
     m_commandBuffer->end();
 }
+
+TEST_F(PositiveSyncVal, ImageUsedInShaderWithoutAccess) {
+    TEST_DESCRIPTION("Test that image queries are not considered to be accesses");
```
filnet, line 1710: typo: `accesses` => `accessed`

artem-lunarg: I'm not a native speaker but I think this specific case is correct. The "accesses" here is not about whether the image queries were *accessed* but rather that they should not be classified as *accesses*.

filnet: Same here, not native...
Sorry for the noise.

spencer-lunarg: I'm the native speaker, and I'm the one who needs to have their English corrected :sob:

PR #8170, opened 2024-06-19
gpu: Add experimental non-conditional injection

spencer-lunarg:
This adds the ability to toggle on non-contional checking for GPU-AV

so instead of 

```
if (!PassFunction(original_value)) {
   value = original_value;
} else {
   value = safe_value;
}
```

we can do

```
PassFunction(original_value)
value = original_value;
```

this will allow @arno-lunarg to experiment with this and robustness

PR #8259, opened 2024-07-09
gpu: Add experimental non-conditional injection

spencer-lunarg:
2nd attempt at https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8170

### DIFF
--- a/layers/gpu/spirv/bindless_descriptor_pass.cpp
+++ b/layers/gpu/spirv/bindless_descriptor_pass.cpp
@@ -78,7 +78,7 @@ uint32_t BindlessDescriptorPass::FindTypeByteSize(uint32_t type_id, uint32_t mat
 // Find outermost buffer type and its access chain index.
 // Because access chains indexes can be runtime values, we need to build arithmetic logic in the SPIR-V to get the runtime value of
 // the indexing
-uint32_t BindlessDescriptorPass::GetLastByte(BasicBlock& block) {
+uint32_t BindlessDescriptorPass::GetLastByte(BasicBlock& block, InstructionIt* inst_it) {
     const Type* pointer_type = module_.type_manager_.FindTypeById(var_inst_->TypeId());
     const Type* descriptor_type = module_.type_manager_.FindTypeById(pointer_type->inst_.Word(3));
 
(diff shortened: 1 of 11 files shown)


layers/gpu/spirv/bindless_descriptor_pass.cpp
```diff
@@ -78,7 +78,7 @@ uint32_t BindlessDescriptorPass::FindTypeByteSize(uint32_t type_id, uint32_t mat
 // Find outermost buffer type and its access chain index.
 // Because access chains indexes can be runtime values, we need to build arithmetic logic in the SPIR-V to get the runtime value of
 // the indexing
-uint32_t BindlessDescriptorPass::GetLastByte(BasicBlock& block) {
+uint32_t BindlessDescriptorPass::GetLastByte(BasicBlock& block, InstructionIt* inst_it) {
```
arno-lunarg, line 81:
```suggestion
uint32_t BindlessDescriptorPass::GetLastByte(BasicBlock& block, InstructionIt& inst_it) {
```
Unless it can sometimes be null

layers/gpu/spirv/function_basic_block.cpp
```diff
@@ -50,6 +50,36 @@ BasicBlock::BasicBlock(Module& module, Function& function) : function_(function)
...
+            case spv::OpSwitch:
+            case spv::OpReturn:
+            case spv::OpReturnValue:
+            case spv::OpKill:
+            case spv::OpUnreachable:
+            case spv::OpTerminateInvocation:
+                break;
+            default:
+                // this works because we know we are not at rend() and there MUST be at least one termination instruction
+                return (inst_it).base();
```
arno-lunarg, line 77:
```suggestion
                return inst_it.base();
```

arno-lunarg (approved):
Well for now I can only trust you on this!
I see `uint32_t` being used everywhere, even though the underlying meaning can differ, leading to potentially mixing apples and oranges... Have you considered creating dedicated types to avoid that?

spencer-lunarg:
> I see uint32_t being used everywhere, even though the underlying meaning can differ, leading to potentially mixing apples and oranges... Have you considered creating dedicated types to avoid that?

we use to do this (I think SPIRV-Tools did also), it honestly was more annoying then helpful  ... basically while it looks like you would mix it up, not once in my 3 years touching SPIR-V can I recall having a bug because I thought this random non-SPIRV thing was actually a SPIR-V id

PR #8261, opened 2024-07-10
gpu: Don't run spirv-opt on non-instrumented shaders

spencer-lunarg:
Currently for ALL shaders in GPU-AV (even thoughs we don't instrument) we are still running the re-building the SPIR-V and then running it through spirv-opt

This will make things faster when we are running on shaders that don't have anything to touch

it will also allow other changes in the shader instrumentation to freely do things knowing it will be ignored if it is never used

### DIFF
--- a/layers/gpu/spirv/module.h
+++ b/layers/gpu/spirv/module.h
@@ -65,6 +65,7 @@ class Module {
     // Order of functions that will try to be linked in
     std::vector<LinkInfo> link_info_;
     void LinkFunction(const LinkInfo& info);
+    bool IsInstrumented() const { return !link_info_.empty(); }
 
     // The class is designed to be written out to a binary file.
     void ToBinary(std::vector<uint32_t>& out);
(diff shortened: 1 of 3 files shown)


layers/gpu/spirv/module.h
```diff
@@ -65,6 +65,7 @@ class Module {
     // Order of functions that will try to be linked in
     std::vector<LinkInfo> link_info_;
     void LinkFunction(const LinkInfo& info);
+    bool HadBeenInstrumented() { return !link_info_.empty(); }
```
arno-lunarg, line 68:
```suggestion
    bool IsInstrumented() { return !link_info_.empty(); }
```

arno-lunarg (approved):

layers/gpu/spirv/module.h
artem-lunarg: I can contribute a const: `bool IsInstrumented() const { return !link_info_.empty(); }`

PR #8264, opened 2024-07-10
gpu: Add setting to override disable GPU-AV

spencer-lunarg:
The app might be setting `VK_VALIDATION_FEATURE_ENABLE_GPU_ASSISTED_EXT` and there is no `VkValidationFeatureDisableEXT` to disable it, so this adds a `VK_LAYER_GPUAV_DEBUG_DISABLE_ALL` option

this is useful for debugging if GPU-AV is the issue or not for app that is forcing it on

### DIFF
--- a/layers/layer_options.cpp
+++ b/layers/layer_options.cpp
@@ -90,6 +90,7 @@ const char *VK_LAYER_GPUAV_BUFFER_COPIES = "gpuav_buffer_copies";
 const char *VK_LAYER_GPUAV_RESERVE_BINDING_SLOT = "gpuav_reserve_binding_slot";
 const char *VK_LAYER_GPUAV_VMA_LINEAR_OUTPUT = "gpuav_vma_linear_output";
 
+const char *VK_LAYER_GPUAV_DEBUG_DISABLE_ALL = "gpuav_debug_disable_all";
 const char *VK_LAYER_GPUAV_DEBUG_VALIDATE_INSTRUMENTED_SHADERS = "gpuav_debug_validate_instrumented_shaders";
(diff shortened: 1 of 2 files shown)


artem-lunarg (approved):

layers/layer_options.cpp
```diff
@@ -578,6 +579,15 @@ void ProcessConfigAndEnvSettings(ConfigAndEnvSettings *settings_data) {
         SetValidationFlags(settings_data->disables, validation_flags_ext);
     }
 
+    // if app is setting VK_VALIDATION_FEATURE_ENABLE_GPU_ASSISTED_EXT, we can use this to disable it for debugging
+    if (vkuHasLayerSetting(layer_setting_set, VK_LAYER_GPUAV_DEBUG_DISABLE_ALL)) {
+        bool disable_gpu = false;
```
arno-lunarg, line 584:
```suggestion
        bool disable_gpuav = false;
```

layers/layer_options.cpp
```diff
@@ -578,6 +579,15 @@ void ProcessConfigAndEnvSettings(ConfigAndEnvSettings *settings_data) {
...
+        vkuGetLayerSettingValue(layer_setting_set, VK_LAYER_GPUAV_DEBUG_DISABLE_ALL, disable_gpu);
+        if (disable_gpu) {
+            settings_data->enables[gpu_validation] = false;
```
arno-lunarg, line 587: While in here （￣︶￣）↗　
```suggestion
            settings_data->enables[gpuav] = false;
```

arno-lunarg (approved):

layers/layer_options.cpp
spencer-lunarg: this is a out of scope change

image

arno-lunarg: 4pk9q1-4122960734

PR #8268, opened 2024-07-10
sync: Make queue submit validation regular layer option

artem-lunarg:
Remove some legacy configuration and make queue submit validation a regular layer setting of synchronization validation.

New setting is called: `syncval_submit_time_validation`.

Decided to go with `Submit time validation` term. One example it is used here: https://vulkan.org/user/pages/09.events/vulkanised-2024/vulkanised-2024-john-zulauf-lunarg.pdf. The `time` suffix is often added because it is more about the place where the validation happens and to less degree about validating vkQueueSubmit itself. The validation also happens for presentation request, so not only vkQueueSubmit(2) specific.

### DIFF
--- a/layers/layer_options.cpp
+++ b/layers/layer_options.cpp
@@ -59,7 +59,6 @@ const char *VK_LAYER_UNIQUE_HANDLES = "unique_handles";
 const char *VK_LAYER_OBJECT_LIFETIME = "object_lifetime";
 const char *VK_LAYER_CHECK_SHADERS = "check_shaders";
 const char *VK_LAYER_CHECK_SHADERS_CACHING = "check_shaders_caching";
-const char *VK_LAYER_VALIDATE_SYNC_QUEUE_SUBMIT = "sync_queue_submit";
 
 const char *VK_LAYER_MESSAGE_ID_FILTER = "message_id_filter";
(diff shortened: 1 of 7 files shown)


layers/layer_options.cpp
```diff
@@ -613,8 +618,6 @@ void ProcessConfigAndEnvSettings(ConfigAndEnvSettings *settings_data) {
         SetValidationSetting(layer_setting_set, settings_data->disables, object_tracking, VK_LAYER_OBJECT_LIFETIME);
         SetValidationSetting(layer_setting_set, settings_data->disables, shader_validation, VK_LAYER_CHECK_SHADERS);
         SetValidationSetting(layer_setting_set, settings_data->disables, shader_validation_caching, VK_LAYER_CHECK_SHADERS_CACHING);
-        SetValidationSetting(layer_setting_set, settings_data->disables, sync_validation_queue_submit,
```
spencer-lunarg, line 616: see `DEPRECATED_GPUAV_VMA_LINEAR_OUTPUT` ... we should keep both settings for a while as if someone is trying to use it via command line it will be awful to debug

artem-lunarg: Thanks for pointing this. Nice solution for transitioning period.

artem-lunarg: Updated. Going update docs later too.

PR #8269, opened 2024-07-11
Correct error message when attempting to build for Android

clayjohn:
Clarify that version under NDK 25 are not supported

I noticed this while attempting to build with NDK 23. The error tells you to use NDK 20 or higher, when in reality, you need NDK 25 or higher.


spencer-lunarg (approved):
LGTM - To fix CI change your commit message to
`build: Correct Android NDK version in error message`

clayjohn:
> LGTM - To fix CI change your commit message to `build: Correct Android NDK version in error message`

Done! Thank you for the feedback

PR #8270, opened 2024-07-11
gpu: Pipe Location through for internal errors

spencer-lunarg:
For our `InternalError` we seem to have random `Location` used. When I hit this error I expected to know which original incoming function was coming in. For things like `Reset()` there are many things that can reset the command buffer and very helpful to have that information

### DIFF
--- a/layers/gpu/descriptor_validation/gpuav_descriptor_set.cpp
+++ b/layers/gpu/descriptor_validation/gpuav_descriptor_set.cpp
@@ -374,7 +374,7 @@ std::shared_ptr<DescriptorSet::State> DescriptorSet::GetCurrentState() {
     return next_state;
 }
 
-std::shared_ptr<DescriptorSet::State> DescriptorSet::GetOutputState(Validator &gpuav) {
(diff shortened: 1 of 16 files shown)


layers/gpu/descriptor_validation/gpuav_descriptor_set.cpp
```diff
@@ -412,10 +412,8 @@ std::shared_ptr<DescriptorSet::State> DescriptorSet::GetOutputState(Validator &g
     alloc_info.requiredFlags = VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_CACHED_BIT;
     VkResult result =
         vmaCreateBuffer(next_state->allocator, &buffer_info, &alloc_info, &next_state->buffer, &next_state->allocation, nullptr);
-    assert(result == VK_SUCCESS);
```
spencer-lunarg, line 415: I found no reason to have an assert here as the `InternalError` will be taking care of handling this

arno-lunarg (approved):

PR #8272, opened 2024-07-11
gpu: Fix when OpStore is used in first block

spencer-lunarg:
Fixes issue where you have

```swift
%firstBlock = OpLabel
OpStore %a %b // don't inject anything
OpStore %c %d // inject check
```

### DIFF
--- a/tests/unit/gpu_av_spirv_positive.cpp
+++ b/tests/unit/gpu_av_spirv_positive.cpp
@@ -353,3 +353,45 @@ TEST_F(PositiveGpuAVSpirv, VulkanMemoryModelDeviceScope) {
     m_default_queue->Submit(*m_commandBuffer);
     m_default_queue->Wait();
 }
+
+TEST_F(PositiveGpuAVSpirv, FindMultipleStores) {
+    TEST_DESCRIPTION("Catches bug when various OpStore are in top of a function");
+    AddDisabledFeature(vkt::Feature::robustBufferAccess);
+    RETURN_IF_SKIP(InitGpuAvFramework());
+    RETURN_IF_SKIP(InitState());
+
+    const char shader_source[] = R"glsl(
+      #version 450
+      layout(set = 0, binding = 0) buffer StorageBuffer { uint data[]; } Data;  // data[4]
+
+      int foo() {
+            return (gl_WorkGroupSize.x > 1) ? 1 : 0;
+      }
+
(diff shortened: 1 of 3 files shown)


tests/unit/gpu_av_spirv_positive.cpp
```diff
@@ -353,3 +353,45 @@ TEST_F(PositiveGpuAVSpirv, VulkanMemoryModelDeviceScope) {
...
+    const char shader_source[] = R"glsl(
+      #version 450
+      layout(set = 0, binding = 0) buffer StorageBuffer { uint data[]; } Data;  // data[4]
+
+      int foo() {
+            return (gl_WorkGroupSize.x > 1) ? 1 : 0;
+      }
+
+      void main() {
+            int index = foo(); // first OpStore is not instrumented
```
artem-lunarg, line 372: Is the idea that instrumentation can distinguish between data[0] and data[1] store?

artem-lunarg (approved):
LGTM.

tests/unit/gpu_av_spirv_positive.cpp
spencer-lunarg: the issue (and didn't know how to describe it simply) was

`index = foo()` is a store, but we don't instrument it
`data[index]` is a store, and we do instrument it

for over-complex spirv instrumentation reasons, we need to find the first instrumented `OpStore` 

while looping, the logic was accidentally grabbing the first `OpStore` of any kind, and then was instrumenting `index = foo()` and everything went wrong from there

this is because `OpStore` in SPIR-V doesn't have a return id (basically a `void` function mentally) so while trying to be fast and checking for the result id, all of them are `zero` so that is why I found the first one

now I just do a "deep check" by checking that all bytes in the `OpStore` instruction are the same to confirm which `OpStore` it was

artem-lunarg: > index = foo() is a store, but we don't instrument it

Interesting, I somehow imagined that OpStore applies only "real write to memory" (buffer/image), so not only that.

spencer-lunarg: we for Buffer Device Address stuff, we only check if the `OpStore` is storing into a pointer address (and we make sure that address is valid)

for "normal" stores, that is checked all statically with `spirv-val`

PR #8277, opened 2024-07-12
tests: Add several scenarios with shader heuristic issues

artem-lunarg:
Tried to classify the problems when syncval shader access tracking does not work. Some scenarios are fixable. The rest will be in the category "too complex for heuristic". Not an attempt to provide full coverage of syncval shader access tracking. Includes a scenario reported recently by the client, and several variations of that.

### DIFF
--- a/tests/unit/sync_val.cpp
+++ b/tests/unit/sync_val.cpp
@@ -6448,3 +6448,81 @@ TEST_F(NegativeSyncVal, DebugDescriptorBufferName3) {
     m_errorMonitor->VerifyFound();
     m_default_queue->Wait();
 }
+
+// TODO: this very simple case should cause WRITE-AFTER-WRITE hazard but it passes successfully.
+// This happens because of SupressedBoundDescriptorWAW(). Remove SupressedBoundDescriptorWAW()
+// after SDK release. SupressedBoundDescriptorWAW works only for command buffer validation,
+// that's why the next test (WriteSameLocationFromTwoSubmits) that uses Submit does not have this problem.
+//
+// Prevention of tricky/complex false-positives should be done by having descriptor validation feature
+// turned off as default (current plan), as opposite to solutions where syncval behaves unreliable even
+// in simple scenarios. Simple should work. Complex should be manageable.
+TEST_F(NegativeSyncVal, WriteSameLocationFromTwoDispatches) {
+    TEST_DESCRIPTION("Not synchronized write to the same location causes WAW hazard");
+    RETURN_IF_SKIP(InitSyncVal());
+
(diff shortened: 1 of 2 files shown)


tests/unit/sync_val.cpp
```diff
@@ -6448,3 +6448,81 @@ TEST_F(NegativeSyncVal, DebugDescriptorBufferName3) {
...
+
+// TODO: this very simple case should cause WRITE-AFTER-WRITE hazard but it passes successfully.
+// This happens because of SupressedBoundDescriptorWAW(). Remove SupressedBoundDescriptorWAW()
+// after SDK release. SupressedBoundDescriptorWAW works only for command buffer validation,
+// that's why the next test (WriteSameLocationFromTwoSubmits) that uses Submit does not have this problem.
+//
+// Prevention of tricky/complex false-positives should be done by having descriptor validation feature
+// turned off as default (current plan), as opposite to solutions where syncval behaves unreliable even
+// in simple scenarios. Simple should work. Complex should be manageable.
+TEST_F(NegativeSyncVal, WriteSameLocationFromTwoDispatches) {
```
artem-lunarg, line 6460: This scenario should be handled by syncval heuristic (if not this then what?). Should be easy to fix.

tests/unit/sync_val_positive.cpp
```diff
@@ -1746,3 +1746,170 @@ TEST_F(PositiveSyncVal, ImageUsedInShaderWithoutAccess) {
...
+    m_default_queue->Submit(m_command_buffer);
+    m_default_queue->Wait();
+}
+
+// TODO: this test does not work due to SupressedBoundDescriptorWAW(). That workaround should be removed.
+// Two possible solutions:
+// a) Try to detect if there is atomic operation in the buffer access chain. If yes, skip validation.
+// b) If a) is hard to do, then this case is in the category that is not handled by the current heuristic
+//    and is part of "Shader access heuristic" is disabled by default direction.
+TEST_F(PositiveSyncVal, AtomicAccessFromTwoDispatches2) {
```
artem-lunarg, line 1834: This one might require more tricky SPIRV analysis comparing to the previous comment. Still it's the static SPIRV analysis, so possible in theory, need to properly ask @spencer-lunarg :).

spencer-lunarg: I would expect the current `i_am_counter` to report `true` in `VariableBase::IsAtomic()` (which I realize should be renamed to `IsAtomicAccessed()` or something better)

spencer-lunarg (approved):

tests/unit/sync_val_positive.cpp
artem-lunarg: Okay! if it's possible to transfer that information to `i_am_data` when it is indexed by the counter, some additional flag that atomic participates in the variable's access chain that would be great (tm).

PR #8279, opened 2024-07-12
sync: Make shader_access_heuristic disabled by default

artem-lunarg:
Improvements to SPIR-V static analysis allowed to properly identify
some accesses as writes (previously there were erroneously classified
as reads). This increased the number of cases where current validation
of shader accesses results in false-positive reports. This makes
current heuristic less suitable to be enabled by default.

Some additional improvements to the heuristic are planned. It can be
useful in some scenarios, but fundamentally it can't be a
false-positive free solution.

### DIFF
--- a/layers/sync/sync_settings.h
+++ b/layers/sync/sync_settings.h
@@ -19,5 +19,5 @@
 
 struct SyncValSettings {
     bool submit_time_validation = true;
-    bool shader_access_heuristic = true;
+    bool shader_access_heuristic = false;
 };
--- a/tests/unit/sync_val_positive.cpp
+++ b/tests/unit/sync_val_positive.cpp
@@ -17,6 +17,7 @@
 #include "../framework/render_pass_helper.h"
 #include "../framework/thread_helper.h"
 #include "../framework/queue_submit_context.h"
+#include "../layers/sync/sync_settings.h"
 
 class PositiveSyncVal : public VkSyncValTest {};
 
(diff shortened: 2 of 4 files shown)


layers/sync/sync_settings.h
```diff
@@ -19,5 +19,5 @@
 
 struct SyncValSettings {
     bool submit_time_validation = true;
-    bool shader_access_heuristic = true;
+    bool shader_access_heuristic = false;
```
spencer-lunarg, line 22: you want to update it in the `VkLayer_khronos_validation.json.in` file... this will set it by default if not using `vkconfig`, but `vkconfig` will use the json to actively set it to true/false

artem-lunarg: Good catch, forgot about it

artem-lunarg: Updated

tests/unit/sync_val_positive.cpp
```diff
@@ -17,43 +17,54 @@
...
-    static const VkBool32 submit_time_validation = false;
-    if (disable_queue_submit_validation) {
-        settings[setting_count++] = {OBJECT_LAYER_NAME, "syncval_submit_time_validation", VK_LAYER_SETTING_TYPE_BOOL32_EXT, 1,
-                                     &submit_time_validation};
-    }
+void VkSyncValTest::InitSyncValFramework(const SyncValSettings *p_sync_settings) {
+    std::vector<VkLayerSettingEXT> settings;
+
+    static const SyncValSettings test_default_sync_settings = [] {
+        SyncValSettings settings;
```
spencer-lunarg, line 35: doesn't this line have default settings already? I guess seems confusing to have this test code here re-set the defaults to something that is different from what it is normally

tests/unit/sync_val_positive.cpp
```diff
@@ -17,43 +17,54 @@
 #include "../framework/render_pass_helper.h"
 #include "../framework/thread_helper.h"
 #include "../framework/queue_submit_context.h"
+#include "../layers/sync/sync_settings.h"
```
spencer-lunarg, line 20: :100: for this idea, seems like a much more scalable way to do setting adjustment .... ~might~ will use in GPU-AV

artem-lunarg: Yeah, something I decided to try. It's more options in gpu-av, will be interesting if no pitfalls with this approach. Initially I passed Settings object by reference initialized with default values, but syncval tests needs different defaults than the main layer configuration, that's the reason of using a pointer.

tests/unit/sync_val_positive.cpp
artem-lunarg: Oh, I actually answered below. Initially I passed object with default values. But tests need different defaults. For example all tests written assuming that `shader_access_heuristic` is enabled, that's why it's option is enabled by default for tests.

spencer-lunarg: agree now... maybe worth a comment

spencer-lunarg (approved):

tests/unit/sync_val_positive.cpp
artem-lunarg: Definitely, since this raises the questions.

PR #8290, opened 2024-07-15
printf: More test and doc cleanup

spencer-lunarg:
- Update documentation
- Improve `VK_LAYER_PRINTF_BUFFER_SIZE` overflow warning
- More tests!

### DIFF
--- a/docs/debug_printf.md
+++ b/docs/debug_printf.md
@@ -1,6 +1,6 @@
 <!-- markdownlint-disable MD041 -->
-<!-- Copyright 2020-2023 LunarG, Inc. -->
-<!-- Copyright 2020-2023 Valve Corporation -->
+<!-- Copyright 2020-2024 LunarG, Inc. -->
+<!-- Copyright 2020-2024 Valve Corporation -->
 [![Khronos Vulkan][1]][2]
 
 [1]: https://vulkan.lunarg.com/img/Vulkan_100px_Dec16.png "https://www.khronos.org/vulkan/"
@@ -25,6 +25,22 @@ The Debug Printf settings are managed by configuring the Validation Layer. These
 
 Debug Printf settings can also be managed using the [Vulkan Configurator](https://vulkan.lunarg.com/doc/sdk/latest/windows/vkconfig.html) included with the Vulkan SDK.
 
+### Settings
+
+> All settings are found in Vulkan Configurator (`VkConfig`)
+
+There are currently 3 environment variables that are used for settings in Debug Printf
+
+- `VK_LAYER_PRINTF_TO_STDOUT` will print to `stdout` **instead** of the normal Debug Callback
+    - `VK_LAYER_PRINTF_TO_STDOUT=1` turn on
+    - `VK_LAYER_PRINTF_TO_STDOUT=0` turn off
+- `VK_LAYER_PRINTF_VERBOSE` will print extra information (pipeline, shader, command, etc)
+    - `VK_LAYER_PRINTF_VERBOSE=1` turn on
+    - `VK_LAYER_PRINTF_VERBOSE=0` turn off
+- `VK_LAYER_PRINTF_BUFFER_SIZE` size of the buffer used to store Printf messages (buffer is shared across all calls in a single `vkQueueSubmit`).
+    - Default: 1024 bytes
+    - `set VK_LAYER_PRINTF_BUFFER_SIZE=4096` (example of making it larger)
+
 ## Using Debug Printf in GLSL Shaders
 
 To use Debug Printf in GLSL shaders, you need to enable the GL_EXT_debug_printf extension.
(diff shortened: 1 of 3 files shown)


docs/debug_printf.md
```diff
@@ -25,6 +25,21 @@ The Debug Printf settings are managed by configuring the Validation Layer. These
 
 Debug Printf settings can also be managed using the [Vulkan Configurator](https://vulkan.lunarg.com/doc/sdk/latest/windows/vkconfig.html) included with the Vulkan SDK.
 
+### Settings
+
+> All settings are found in Vulkan Configurator (`VkConfig`)
+
+There are currently 3 environmental variables that are used for settings in Debug PrintF
```
filnet, line 32: the uppercase F in PrintF is strange...

docs/debug_printf.md
```diff
@@ -25,6 +25,21 @@ The Debug Printf settings are managed by configuring the Validation Layer. These
...
+
+- `VK_LAYER_PRINTF_TO_STDOUT` will print to `stdout` **instead** of the normal Debug Callback
+    - `VK_LAYER_PRINTF_TO_STDOUT=1` turn on
+    - `VK_LAYER_PRINTF_TO_STDOUT=0` turn off
+- `VK_LAYER_PRINTF_VERBOSE` will print extra information which pipeline, shader, command, etc, the printf message came from
```
filnet, line 37: does not read well / is not syntactically correct

docs/debug_printf.md
```diff
@@ -25,6 +25,21 @@ The Debug Printf settings are managed by configuring the Validation Layer. These
...
+    - `VK_LAYER_PRINTF_VERBOSE=1` turn on
+    - `VK_LAYER_PRINTF_VERBOSE=0` turn off
+- `VK_LAYER_PRINTF_BUFFER_SIZE` size of buffer to allocate to put printf message too (this is shared across all calls in a single `vkQueueSubmit`)
```
filnet, line 40: `too` => `to` or `into`

docs/debug_printf.md
```diff
@@ -191,8 +211,7 @@ unrelated to Debug Printf, making it difficult to find the desired output.
 Documentation for the GL_EXT_debug_printf extension can be found
 [here](https://github.com/KhronosGroup/GLSL/blob/main/extensions/ext/GLSL_EXT_debug_printf.txt)
 
-There is a validation layer test that demonstrates the simple and programmatic use of Debug
-Printf. It is called "NegativeDebugPrintf.BasicUsage" and is in `tests/unit/debug_printf.cpp` 
-in the Vulkan-ValidationLayers repository.
+There are many validation layer test that demonstrates the simple and programmatic use of Debug
```
filnet, line 214: plural: `many [...] tests [...] that demonstrate`

layers/gpu/debug_printf/debug_printf.cpp
```diff
@@ -397,8 +397,10 @@ void Validator::AnalyzeAndGenerateMessage(VkCommandBuffer command_buffer, VkQueu
         index += debug_record->size;
     }
     if ((index - spvtools::kDebugOutputDataOffset) != expect) {
-        InternalWarning(queue, loc,
-                        "Debug Printf message was truncated, likely due to a buffer size that was too small for the message");
+        std::stringstream message;
+        message << "Debug Printf message was truncated, likely due to a buffer size (" << printf_settings.buffer_size
```
filnet, line 401: Is it possible to also show the size of the message that was too long ?
Will help adjust the buffer size if needed.

filnet, line 401: The use of `likely` entails that there can be other causes. Is that true ?

filnet, line 401: Sorry for spamming but the message could be better formulated.
Not all users of DebugPrintf are aware that the buffer will accumulate multiple messages and will in some cases be filled.
Once the buffer is filled all subsequent calls to DebugPrintf will be truncated/ignored.
Saying that the buffer "was to small for the message" is a bit confusing.
If the buffer is 1024 bytes and the message is 5 bytes, then the buffer *is* big enough for the message.
The issue is that the buffer does not have enough space left for the message...

filnet: My understanding of message buffering is:
* if the buffer does not have enough free space for the message a truncated version of the message will be written (and the buffer will be full)
* after that message are ignored (?)
If the above is true, then how do you handle the case where the last message was big enough to exactly fill the buffer.
I.e. there will be no room at all to write the next message or a truncated version of the next message.
CPU side will not be able to detect that a message was truncated.
PS : I have not done a deep dive in the code so I am probably totally of here..

filnet: So I believe the warning should say that *a* message was truncated because the buffer is full and that subsequent messages were ignored.

filnet: Nothing that a few global counters cannot handle.

docs/debug_printf.md
filnet, line 40: plural: messages

layers/gpu/debug_printf/debug_printf.cpp
spencer-lunarg: I doubt it, this was from the original addition of DebugPrintf... safe to say this is the only cause, otherwise yes, it makes the user think "wait, what else is there!?"

spencer-lunarg: agree, going to update this fully

docs/debug_printf.md
arno-lunarg, line 32:
```suggestion
There are currently 3 environment variables that are used for settings in Debug PrintF
```

docs/debug_printf.md
arno-lunarg, line 40:
```suggestion
- `VK_LAYER_PRINTF_BUFFER_SIZE` size of the buffer used to store Printf messages (buffer is shared across all calls in a single `vkQueueSubmit`). Default: 1024 bytes.
```

docs/debug_printf.md
```diff
@@ -25,6 +25,21 @@ The Debug Printf settings are managed by configuring the Validation Layer. These
...
+There are currently 3 environmental variables that are used for settings in Debug PrintF
+
+- `VK_LAYER_PRINTF_TO_STDOUT` will print to `stdout` **instead** of the normal Debug Callback
+    - `VK_LAYER_PRINTF_TO_STDOUT=1` turn on
+    - `VK_LAYER_PRINTF_TO_STDOUT=0` turn off
+- `VK_LAYER_PRINTF_VERBOSE` will print extra information which pipeline, shader, command, etc, the printf message came from
+    - `VK_LAYER_PRINTF_VERBOSE=1` turn on
+    - `VK_LAYER_PRINTF_VERBOSE=0` turn off
+- `VK_LAYER_PRINTF_BUFFER_SIZE` size of buffer to allocate to put printf message too (this is shared across all calls in a single `vkQueueSubmit`)
+    - `VK_LAYER_PRINTF_BUFFER_SIZE=1024` default size
```
arno-lunarg, line 41:
```suggestion
```

docs/debug_printf.md
```diff
@@ -35,35 +50,35 @@ Here is a very simple example:
...
+Then use `glslangValidator` to generate SPIR-V to use in `vkCreateShaderModule`.
+`glslangvalidator --target-env vulkan1.2 -e main -o shader.vert.spv shader.vert` would be one
 example of compiling shader.vert
 
-Note that every time this shader is executed, "My float is 3.141500" will be printed. If this were
+Note that every time this shader is executed, `My float is 3.141500` will be printed. If this were
 in a vertex shader and a triangle was drawn, it would be printed 3 times.
 
-Note also that the VK_KHR_shader_non_semantic_info device extension must be enabled in
+Note also that the `VK_KHR_shader_non_semantic_info` device extension must be enabled in
```
arno-lunarg, line 64: This info seems quite important, but lost at the end of the paragraph, I think it should be elsewhere

arno-lunarg: It is also repeated in the `Using Debug Printf in SPIR-V Shaders` section

arno-lunarg (approved):
Nice update!
Meanwhile:
image

docs/debug_printf.md
spencer-lunarg: will remove the duplicate, but I need to look into this, not full sure this is needed as a requirement since we implement it in the shader instrumentation

layers/gpu/debug_printf/debug_printf.cpp
spencer-lunarg: just heads up, I have a larger follow-up PR that will do more around error messages as a whole and plan to improve this still

PR #8294, opened 2024-07-16
printf: Add Basic DebugPrintf string validation

spencer-lunarg:
Currently we do zero validation for debug printf, things like `debugPrintfEXT("vector of %f", myInt);` or other variations will blindly start printing garbage and there is nothing validating it

Checked and Renderdoc follows the same formatting so this should be universal (left comment)

To start these are all warnings, but some things feel like errors (so we might change them in future)

### DIFF
--- a/layers/core_checks/cc_spirv.cpp
+++ b/layers/core_checks/cc_spirv.cpp
@@ -2884,6 +2884,267 @@ bool CoreChecks::ValidateEmitMeshTasksSize(const spirv::Module &module_state, co
     return skip;
 }
 
+bool CoreChecks::ValidateDebugPrint(const spirv::Module &module_state, const spirv::StatelessData &stateless_data,
+                                    const Location &loc) const {
+    bool skip = false;
+    if (stateless_data.debug_printf_import_id == 0) return skip;
+
+    // Strictly speaking - the format given in GLSL_EXT_debug_printf is a client side implementation of SPIR-V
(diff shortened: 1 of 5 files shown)


layers/core_checks/cc_spirv.cpp
```diff
@@ -2884,6 +2884,262 @@ bool CoreChecks::ValidateEmitMeshTasksSize(const spirv::Module &module_state, co
...
+        uint32_t string_id = inst->Word(5);
+
+        const char *op_string = nullptr;
+        for (const spirv::Instruction *string_inst : stateless_data.string_inst) {
+            if (string_inst->ResultId() == string_id) {
+                op_string = string_inst->GetAsString(2);
+                break;
+            }
+        }
+        ASSERT_AND_CONTINUE(op_string);
```
artem-lunarg, line 2923: Is it possible to get empty string `""` (still not null)? If yes, should be accounted for, `op_string_len - 1` in the loop will be  problematic.

artem-lunarg (approved):

layers/core_checks/cc_spirv.cpp
spencer-lunarg: will add tests and make sure

spencer-lunarg: it was going `OpString "" contains "%" which is missing a valid specifier (d, i, o, u, x, X, a, A, e, E, f, F, g, or G).`

so now added a check and test for empty strings (if you go `debugPrintfEXT("");`)

Issue #7143, opened 2023-12-14
debugPrintfEXT with %g formatting on GPU doubles, dvec2, dvec4 etc is broken


mlfarrell:
**Environment:**
 - OS: Linux
 - GPU and driver version:  NVIDIA 535.86.05

**Describe the Issue**

Sorry for being brief, i dont have a ton of time these days.

```
layout(binding = 0) readonly buffer SegmentedCurveA
{
    int numPoints;
    dvec2 curve[];
} curveA;

//...

debugPrintfEXT("%g", curveA.curve[0].x);
```

**Expected behavior**

Double appears correctly

**Additional context**
Note, that I believe the issue is in https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/4bd62a370f8ed8e5c7e349d345a665e2644b0c15/layers/gpu_validation/debug_printf.cpp#L175

varfloat is implemented on doubles using float * pointer casts on doubles which is incorrect.

spencer-lunarg:
So I looked into this now, so the [GLSL_EXT_debug_printf](https://github.com/KhronosGroup/GLSL/blob/master/extensions/ext/GLSL_EXT_debug_printf.txt) spec says

>  Interpretation of the format specifiers is specified by the client API.
    The set of format specifiers is implementation-dependent, but must
    include at least "%d" and "%i" (int), "%u" (uint), and "%f" (float).
    
and I see we only ever added `%lu` and `%lx` for 64-bit unsigned ints, and 64-bit doubles were never added

So I guess is not "broken" and just "currently not supported". I do plan to get to this, just not sure how soon

spencer-lunarg:
so update on this, spent a lot of time this week on DebugPrintf

so the core issue is how this is we use `snprintf` on the CPU side (RenderDoc does as well) so there `%f` or `%g` works for 32-bit and 64-bit floats.... but in GPU world, not all devices even support 64-bit floats

the current implementation doesn't know if the contents are 32-bit or 64-bit because of the way we swap out the non-semantic SPIR-V op for a function call to do our own custom way to write out to the buffer

... basically this works on Renderdoc and really a VVL bug and needs to be fixed and I have tests, just will be a bit of re-working (which luckily is already needed to allow DebugPrintf to work with GPU-AV)

PR #8296, opened 2024-07-16
printf: Cleanup Printf string message parsing logic

spencer-lunarg:
The logic for `AnalyzeAndGenerateMessage` in debug printf was a little cryptic and out-of-date. Reorganize it with the intentions to finally get a fix for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7143 soon

### DIFF
--- a/tests/unit/debug_printf.cpp
+++ b/tests/unit/debug_printf.cpp
@@ -463,7 +463,8 @@ TEST_F(NegativeDebugPrintf, DISABLED_Float64) {
         #extension GL_EXT_shader_explicit_arithmetic_types_float64 : enable
         void main() {
             float64_t foo = 1.23456789;
-            debugPrintfEXT("float64 == %lf", foo);
+            float bar = 1.23456789;
+            debugPrintfEXT("floats and doubles %f %f %f", foo, bar, foo);
         }
     )glsl";
     BasicComputeTest(shader_source, "float64 == 1.23456789");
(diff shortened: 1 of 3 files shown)


tests/unit/debug_printf.cpp
```diff
@@ -463,7 +463,8 @@ TEST_F(NegativeDebugPrintf, DISABLED_Float64) {
         #extension GL_EXT_shader_explicit_arithmetic_types_float64 : enable
         void main() {
             float64_t foo = 1.23456789;
-            debugPrintfEXT("float64 == %lf", foo);
+            float bar = 1.23456789;
+            debugPrintfEXT("floats and doubles %f %f %f", foo, bar, foo);
         }
     )glsl";
     BasicComputeTest(shader_source, "float64 == 1.23456789");
```
filnet, line 470: that test should fail ?

filnet, line 470: that test should fail ?

tests/unit/debug_printf.cpp
```diff
@@ -463,7 +463,8 @@ TEST_F(NegativeDebugPrintf, DISABLED_Float64) {
         #extension GL_EXT_shader_explicit_arithmetic_types_float64 : enable
         void main() {
             float64_t foo = 1.23456789;
-            debugPrintfEXT("float64 == %lf", foo);
+            float bar = 1.23456789;
+            debugPrintfEXT("floats and doubles %f %f %f", foo, bar, foo);
```
filnet, line 467: why `foo` twice with the same `%f`

tests/unit/debug_printf.cpp
spencer-lunarg: its disabled, but I wanted to test we can correctly pull out a 64-bit float between two 32-bit floats

this is really just here for me to test with, hense why it is still disabled

artem-lunarg (approved):

Issue #8167, opened 2024-06-18
[FR] build AAR for android binaries, publish to github pacakges?


DanAlbert:
Not a bug, so neither template seems to apply.

Using the validation layers on Android requires jumping through a few hoops: https://developer.android.com/ndk/guides/graphics/validation-layer

This repo does already publish prebuilt libraries (thank you, that does simplify their use a lot), but just as a bare zip/tarball of libraries (and chrome flags the download as malware). Users have to vendor these libraries into their repo and remember to keep them up to date.

If they were instead an AAR and published to https://docs.github.com/en/packages ([free for public repos](https://github.com/features/packages#pricing)), those instructions reduce down to just defining the repo and declaring the dependency as described in https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-gradle-registry#using-a-published-package. Android Studio would also highlight the dependency as out of date whenever a new version was available (I think so, anyway, maybe it only does that for specific libraries).

Is that something the maintainers here would be open to? Looking at https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/scripts/android.py and https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/.github/workflows/sdk_android_build.yml, it's not a whole lot of work to produce the AAR, and I can volunteer to do that part, but an owner here would need to enable the github package repo.

spencer-lunarg:
we are always happy to help here in anyway with Android, unfortunately I don't personally do much Android development as I use to, so my knowledge of the latest best practices for build flow/processes are low so having your help do the work would be awesome and I can help enable and repo settings needed

DanAlbert:
SG, I'll send a PR soonish. It may be a few weeks (US holidays coming up next week).

DanAlbert:
Uploaded https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8303. As I mention in the PR, GHR doesn't work as well as I'd hoped (authentication is required even to _read_ public packages), so the simplest option isn't actually very good.

DanAlbert:
The last thing I was going to try (publish to Maven Central) requires that the packages be signed, and I know just enough to know that I don't know enough to set up a secure signing workflow in GHA (in particular, I don't know how to protect keys from exposure in malicious PRs), so unfortunately I don't think I'm going to make any more progress here.

[...]

PR #8303, opened 2024-07-19
Publish an AAR for Android developers.

DanAlbert:
This extends the existing Android workflow to collect the validation layer libraries into an AAR that is published to the repository's GitHub Packages Repository. This way, Android Gradle users can point their build at this repository and add a new dependency to their build.gradle.kts file rather than needing to download and manage the artifacts manually.

Fixes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8167

The alternate implementation would be to restructure all of the Android build as a Gradle project. Most of this workflow (and both of the Python scripts it calls) could be deleted since this behavior is all included in the Android Gradle Plugin (AGP). I opted not to do that because it's much more invasive, and people that aren't already Gradle users generally don't like Gradle being added to their projects :) It'd also have an impact on the build speed of the project (but only for this workflow, which only runs for new tags), since AGP builds each ABI in series rather than in parallel like the current workflow does. If you would actually prefer I do it that way, lmk, I don't think it's actually much work to migrate to that, since it's mostly deleting code.

You can see this in action on my fork: https://github.com/DanAlbert/Vulkan-ValidationLayers/actions/runs/9999981314 publishes https://github.com/DanAlbert/Vulkan-ValidationLayers/packages/2208155, which can be used like https://github.com/android/ndk-samples/pull/1072.

Unfortunately I've just found out that GitHub requires authentication to _read_ public packages for maven/gradle registries (everything but docker images?)

https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-apache-maven-registry

> You need an access token to publish, **install**, and delete private, internal, and public packages.

That actually makes this quite annoying to use (especially for anyone that would want their repository to be easy to build and clone without configuring that, which is probably most). I'd suggested GHR in the feature request because I'd thought it would be easiest. Since that seems to not be true, this should probably be tweaked to publish somewhere like Maven Central?

### DIFF
--- a/.github/workflows/sdk_android_build.yml
+++ b/.github/workflows/sdk_android_build.yml
@@ -19,7 +19,10 @@ name: SDK Android Build
 # artifacts to the release when a Vulkan SDK tag is pushed. The
 # Vulkan SDK does not include binaries for Android, so we publish
 # them here to provide Android binaries built from the same source
-# used to build the Vulkan SDK.
+# used to build the Vulkan SDK. The artifacts will also be bundled into an AAR
+# (the library counterpart the APK application format) and uploaded to GitHub
+# Packages so app developers can include the validation layers in their
+# application the same way they would any Java dependencies.
 #
 # The tag needs to be pushed by name, as `git push --tags` to push all
 # tags does not appear to trigger the action.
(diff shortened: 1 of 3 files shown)


DanAlbert:
> Since that seems to not be true, this should probably be tweaked to publish somewhere like Maven Central?

The PR doesn't have to change much for this, but someone on the Khronos side would need to set up an account somewhere (I've never done it, so I'm not sure exactly what's involved, but if you're okay with doing that I'll figure it out as part of reworking the PR and let you know, I expect it's easy). Maven Central would be the easiest on developers since every AGP project already has that repo configured, but if there's a preference for a different host, lmk.

DanAlbert:
(can ignore the CI bot for now, I'll un-draft this once it's ready for review, but I have to push to test it)

DanAlbert:
I haven't had any time to work on it, unfortunately. None of the easy solutions worked. I think the last option (use maven central) isn't too much work but it's definitely all paged out of my brain :( If someone else wants to pick up the work I'd certainly be grateful, but if not I'll be able to get back to it eventually (though maybe not even this year tbh).

PR #8305, opened 2024-07-19
printf: Move implementation inside VVL

spencer-lunarg:
This was brought over from the current implementations in SPIRV-Tools (`source/opt/inst_debug_printf_pass.cpp`) for 3 reasons (after lots of discussions)

1. There are bugs that need be fixed and easier to iterate without having to grab a dependency
2. To get GPU-AV and work with DebugPrintf we need a more unified way the shaders are instrumented
3. Allows future maintainer to not need to learn yet-another-code-base to make changes

There is more I want to fix/add but this is the initial "get it to work the same" change... there are limitations (as noted from disabled tests) of the old implementation and plan to fix those once this is in

### DIFF
--- a/layers/gpu/debug_printf/debug_printf.cpp
+++ b/layers/gpu/debug_printf/debug_printf.cpp
@@ -16,11 +16,14 @@
  */
 
 #include "gpu/debug_printf/debug_printf.h"
-#include "spirv-tools/instrument.hpp"
-#include <iostream>
 #include "generated/layer_chassis_dispatch.h"
 #include "state_tracker/shader_stage_state.h"
 #include "chassis/chassis_modification_state.h"
+#include "gpu/spirv/module.h"
+#include "gpu/shaders/gpu_error_header.h"
+
+#include <iostream>
+#include <fstream>
 
 namespace debug_printf {
 
@@ -47,8 +50,8 @@ void Validator::PostCreateDevice(const VkDeviceCreateInfo *pCreateInfo, const Lo
         use_stdout = true;
     }
 
-    const uint32_t kDebugOutputPrintfStream = 3;  // from instrument.hpp
-    VkDescriptorSetLayoutBinding binding = {kDebugOutputPrintfStream, VK_DESCRIPTOR_TYPE_STORAGE_BUFFER, 1,
+    binding_slot_ = (uint32_t)instrumentation_bindings_.size();  // get next free binding
+    VkDescriptorSetLayoutBinding binding = {binding_slot_, VK_DESCRIPTOR_TYPE_STORAGE_BUFFER, 1,
                                             kShaderStageAllGraphics | VK_SHADER_STAGE_COMPUTE_BIT | kShaderStageAllRayTracing,
                                             nullptr};
     instrumentation_bindings_.push_back(binding);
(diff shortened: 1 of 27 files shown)


layers/gpu/spirv/debug_printf_pass.cpp
```diff
@@ -0,0 +1,457 @@
...
+        check_block->CreateInstruction(spv::OpULessThanEqual, {bool_type_id, less_than_equal_id, int_add_id, array_length_id});
+
+        const uint32_t merge_block_label_id = merge_block->GetLabelId();
+        check_block->CreateInstruction(spv::OpSelectionMerge, {merge_block_label_id, spv::SelectionControlMaskNone});
+
+        const uint32_t store_block_label_id = store_block->GetLabelId();
+        check_block->CreateInstruction(spv::OpBranchConditional, {less_than_equal_id, store_block_label_id, merge_block_label_id});
+    }
+
+    // Store how many bytes
```
jeremyg-lunarg, line 338: Is this really bytes or is it 32 bit words?

layers/gpu/spirv/debug_printf_pass.cpp
```diff
@@ -0,0 +1,457 @@
...
+
+        const uint32_t access_chain_id = module_.TakeNextId();
+        store_block->CreateInstruction(spv::OpAccessChain,
+                                       {pointer_type_id, access_chain_id, output_buffer_variable_id_, one_id, int_add_id});
+
+        const uint32_t shader_id = module_.type_manager_.GetConstantUInt32(module_.shader_id_).Id();
+        store_block->CreateInstruction(spv::OpStore, {access_chain_id, shader_id});
+    }
+
+    // Write a byte for each argument
```
jeremyg-lunarg, line 363: bytes?

jeremyg-lunarg (approved):
LGTM this looks like what I remember of the spirv-tools implementation but simpler.

layers/gpu/spirv/debug_printf_pass.cpp
spencer-lunarg: yes, how many 32-bit words, will update

arno-lunarg:
Will review tomorrow, sorry I did not take the time this week (￣(工)￣)

layers/gpu/spirv/function_basic_block.cpp
```diff
@@ -117,6 +117,13 @@ BasicBlockIt Function::InsertNewBlock(BasicBlockIt it) {
     return new_block_it;
 }
 
+void Function::InitBlocks(uint32_t count) {
+    for (uint32_t i = 0; i < count; i++) {
```
arno-lunarg, line 121:
```suggestion
    blocks_.reserve(blocks_.size() + count);
    for (uint32_t i = 0; i < count; i++) {
```

layers/gpu/spirv/module.cpp
```diff
@@ -227,28 +226,68 @@ void Module::AddExtension(const char* extension) {
...
+    debug_name_.emplace_back(std::move(new_inst));
+}
+
+void Module::AddDecoration(uint32_t target_id, spv::Decoration decoration, const std::vector<uint32_t>& operands) {
+    auto new_inst = std::make_unique<Instruction>((uint32_t)(operands.size() + 3), spv::OpDecorate);
+    new_inst->Fill({target_id, (uint32_t)decoration});
+    if (!operands.empty()) {
+        new_inst->Fill(operands);
+    }
+    annotations_.push_back(std::move(new_inst));
```
arno-lunarg, line 243: Nit:
```suggestion
    annotations_.emplace_back(std::move(new_inst));
```

arno-lunarg (approved):
LGTM! Thank you for doing this, must have been as tedious as it was needed!

layers/gpu/spirv/function_basic_block.cpp
spencer-lunarg: I need to go around and see if we can do this reserve more often for SPIR-V stuff

layers/gpu/spirv/bindless_descriptor_pass.cpp
```diff
@@ -424,5 +425,9 @@ bool BindlessDescriptorPass::AnalyzeInstruction(const Function& function, const
     return true;
 }
 
+void BindlessDescriptorPass::PrintDebugInfo() {
+    std::cout << "BindlessDescriptorPass\n\tinstrumentation count: " << instrumented_count_ << "\n";
```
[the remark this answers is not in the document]
spencer-lunarg: going to do a big find-and-replace after, we do this a lot actually

layers/gpu/debug_printf/debug_printf.cpp
```diff
@@ -80,48 +83,66 @@ void Validator::DestroyBuffer(BufferInfo &buffer_info) {
...
                                  std::vector<uint32_t> &out_instrumented_spirv) {
     if (input[0] != spv::MagicNumber) return false;
 
-    // Load original shader SPIR-V
-    out_instrumented_spirv.clear();
-    out_instrumented_spirv.reserve(input.size());
-    out_instrumented_spirv.insert(out_instrumented_spirv.end(), &input.front(), &input.back() + 1);
+    std::vector<uint32_t> binary;
+    binary.reserve(input.size());
+    binary.insert(binary.end(), &input.front(), &input.back() + 1);
```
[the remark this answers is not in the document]
spencer-lunarg: I know before we had an issue with how `spirv-opt` needed the input to be a certain type (I think `std::vector<uint32_t>`) but now that this is all gone, I need to look into making sure there is no copy all the way down the chain from when we get the SPIR-V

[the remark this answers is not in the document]
spencer-lunarg: idk if we need a copy or not anymore neither, I **think** the issue was before we might have touched that vector, but now I consume things in its own representation, so it **should** be a const ref... need to make sure (here and normal GPU-AV)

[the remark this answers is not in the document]
arno-lunarg: worth looking into!

PR #8307, opened 2024-07-21
bp: Move PreCallRecord to PostCallRecord

spencer-lunarg:
This moves some `PreCallRecord` in Best Practices to `PostCallRecord` where it didn't need to be in `PreCall`

### DIFF
--- a/layers/best_practices/bp_drawdispatch.cpp
+++ b/layers/best_practices/bp_drawdispatch.cpp
@@ -83,26 +83,25 @@ bool BestPractices::ValidatePushConstants(VkCommandBuffer cmd_buffer, const Loca
     return skip;
 }
 
-void BestPractices::RecordCmdDrawType(VkCommandBuffer cmd_buffer, uint32_t draw_count) {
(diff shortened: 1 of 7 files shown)


layers/best_practices/bp_drawdispatch.cpp
```diff
@@ -83,26 +83,25 @@ bool BestPractices::ValidatePushConstants(VkCommandBuffer cmd_buffer, const Loca
     return skip;
 }
 
-void BestPractices::RecordCmdDrawType(VkCommandBuffer cmd_buffer, uint32_t draw_count) {
-    auto cb_state = GetWrite<bp_state::CommandBuffer>(cmd_buffer);
+void BestPractices::RecordCmdDrawType(bp_state::CommandBuffer& cb_state, uint32_t draw_count) {
```
spencer-lunarg, line 86: some of these best practices where calling the `GetWrite<bp_state::CommandBuffer>` inside these helper functions which was hiding a way to deadlock on this, so now it is called at the top function and passed in as ref (like we do everywhere else)

artem-lunarg (approved):

PR #8312, opened 2024-07-21
scripts: Add missing parent VUs

spencer-lunarg:
this is a follow-up to https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8144 (based from https://gitlab.khronos.org/vulkan/vulkan/-/issues/3553)

The core issue is the spec doesn't generate proper implicit VUs for `common parent` checks

We (correctly) are ignoring and throwing warnings when we don't recognize the missing VU and with that we have been missing a LOT of checks for `common parent" checks.

Before we just threw the struct name in the `self.structs_that_forgot_about_parent_vuids` and moved on

this manually add them in a way that one day (probably not anytime soon tbh) when someone goes and fix the spec generating logic, we have a more descriptive breakdown of all the VUs missing

(tl;dr - lots of VUID churn 😢)

### DIFF
--- a/tests/unit/parent.cpp
+++ b/tests/unit/parent.cpp
@@ -180,6 +180,26 @@ TEST_F(NegativeParent, BindPipeline) {
     m_commandBuffer->end();
 }
 
+TEST_F(NegativeParent, PipelineShaderStageCreateInfo) {
+    RETURN_IF_SKIP(Init());
+    auto features = m_device->phy().features();
+    m_second_device = new vkt::Device(gpu_, m_device_extension_names, &features, nullptr);
+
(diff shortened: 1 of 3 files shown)


tests/unit/parent.cpp
```diff
@@ -180,6 +180,26 @@ TEST_F(NegativeParent, BindPipeline) {
     m_commandBuffer->end();
 }
 
+TEST_F(NegativeParent, PipelineShaderStageCreateInfo) {
```
spencer-lunarg, line 183: because we were ignoring the warning and just printing `kVUIDUndefined` we were never validating this

the main benefit of this change is adding lots of checks that were not being checked (and we have seen CTS hit these bugs before)

arno-lunarg (approved):

PR #8313, opened 2024-07-22
layers: Integrate Tracy profiler

arno-lunarg:
See `profiling.md` for details

Only made one interesting change to the code outside of Tracy: Removed all global objects with a constructor that dynamically allocates memory, placed those objects inside dedicated access functions.
This approach allows to pay the cost of construction only when truly needed.


layers/state_tracker/shader_module.cpp
```diff
@@ -852,6 +852,7 @@ Module::StaticData::StaticData(const Module& module_state, StatelessData* statel
     {
         std::vector<uint32_t>::const_iterator it = module_state.words_.cbegin();
         it += 5;  // skip first 5 word of header
+        instructions.reserve(module_state.words_.size() * 4);
```
arno-lunarg, line 855: This was among low hanging optimisations. Turns out this is a hot path, so reserving memory, even with a heuristic this dumb, saves millions of small allocations/mem moves

layers/layer_options.cpp
```diff
@@ -31,6 +32,224 @@
...
+            if (!tracy::IsProfilerStarted()) {
+                tracy::StartupProfiler();
+            }
+            break;
+        case DLL_THREAD_ATTACH:
+            break;
+        case DLL_THREAD_DETACH:
+            break;
+        case DLL_PROCESS_DETACH:
+#if 0
```
spencer-lunarg, line 142: still need?

layers/layer_options.cpp
```diff
@@ -31,6 +32,224 @@
 #include "mimalloc-new-delete.h"
 #endif
 
+#include <cstdlib>
+
+#if defined(TRACY_ENABLE)
+
+#include "profiling/profiling.h"
+
+void *operator new(std ::size_t size) {
```
spencer-lunarg, line 41: why does this stuff need to be in `layer_options.cpp` (compared to another file)

scripts/generators/layer_chassis_generator.py
```diff
@@ -1135,7 +1148,7 @@ def generateSource(self):
                 if (local_disables[handle_wrapping]) {
                     wrap_handles = false;
                 }
-
+
```
spencer-lunarg, line 1151: revert :smile:

scripts/generators/spirv_validation_generator.py
```diff
@@ -187,25 +187,31 @@ def generate(self):
...
+        out.append('const std::unordered_multimap<std::string_view, RequiredSpirvInfo>& GetSpirvExtensions() {\n')
         out.append('// clang-format off\n')
-        out.append('static const std::unordered_multimap<std::string_view, RequiredSpirvInfo> spirvExtensions = {\n')
+        out.append('    static const std::unordered_multimap<std::string_view, RequiredSpirvInfo> spirv_extensions = {\n')
         for spirv in [x for x in self.vk.spirv if x.extension]:
             for enable in spirv.enable:
                 out.append(f'    {{"{spirv.name}", {self.createMapValue(spirv.name, enable, True)}}},\n')
-        out.append('};\n')
+        out.append('    };\n')
         out.append('// clang-format on\n')
```
spencer-lunarg, line 212: nit: put down 

```
        out.append('}\n')
        out.append('// clang-format on\n')
        out.append('\n')
```

CMakeLists.txt
```diff
@@ -76,6 +76,28 @@ else()
     endif()
 endif()
 
+#Tracy
+option(VVL_ENABLE_TRACY "Enable profiling with Tracy" OFF)
+option(VVL_ENABLE_TRACY_CPU_MEMORY "Enable profiling memory with Tracy" OFF)
+option(VVL_TRACY_CALLSTACK "Size of the collected collected call stacks" 48)
```
spencer-lunarg, line 82: https://cmake.org/cmake/help/v3.22/command/option.html

older cmake version (this is the 22.04 Ubuntu default version) will set this as a `OFF` and this failed to build ... so either need to bump the version to `3.24` or just have user manually set this... 

I think the easiest option is go

```
if (VVL_TRACY_CALLSTACK EQUALS "OFF") 
   Error("explain they have an old CMake and need to manually set this")
```

layers/layer_options.cpp
```diff
@@ -31,6 +32,224 @@
...
+}
+
+void operator delete[](void *p, const std::nothrow_t &) noexcept {
+    VVL_TracyFree(p);
+    free(p);
+}
+
+#if (__cplusplus > 201402L || defined(__cpp_aligned_new))
+void *operator new(std::size_t size, std::align_val_t al) noexcept(false) {
+    auto ptr = _aligned_malloc(size, size_t(al));
```
spencer-lunarg, line 87: fails to build with `clang` 

https://stackoverflow.com/questions/29247065/compiler-cant-find-aligned-alloc-function

```
#if !defined(_WIN32)
#define _aligned_malloc(size, alignment) aligned_alloc(alignment, size)
#define _aligned_free(ptr) free(ptr)
#endif
```

is my hot fix

layers/CMakeLists.txt
```diff
@@ -417,6 +418,25 @@ target_link_libraries(vvl PRIVATE
...
+if (VVL_ENABLE_TRACY)
+    include(FetchContent)
+    FetchContent_Declare(
+        tracy
+        GIT_REPOSITORY https://github.com/wolfpld/tracy.git
+        GIT_TAG 075395620a504c0cdcaf9bab3d196db16a043de7
+        GIT_SHALLOW TRUE
+        GIT_PROGRESS TRUE)
+
+    #set(TRACY_CALLSTACK ON CACHE BOOL "Tracy callstack")
```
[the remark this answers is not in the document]
arno-lunarg: The Tracy setup is just somewhat working for now, so I want to leave those `set` in the cmakelists.txt, as well as dead code in `DllMain`, for I may need to reactivate those in the future as part of fixing this broken setup. Figuring out that I could do those things was a bit tedious, so I want to keep them for now

layers/CMakeLists.txt
```diff
@@ -417,6 +418,25 @@ target_link_libraries(vvl PRIVATE
     VkLayer_utils
 )
 
+if (VVL_ENABLE_TRACY)
+    include(FetchContent)
+    FetchContent_Declare(
```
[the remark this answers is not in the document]
arno-lunarg: Just as you say, it is dev only, simple and working as we need it to, so I'd rather stay away from update_deps.py

layers/layer_options.cpp
arno-lunarg: (first comment)

layers/layer_options.cpp
arno-lunarg: it needs to be in some translation unit, and only one, since mimalloc was there I decided to do my changes here. I could probably add this in a dedicated file, will do

layers/layer_options.cpp
arno-lunarg: Oh yes I forgot about that ￣へ￣

CMakeLists.txt
arno-lunarg: I think it is still valuable to have a default value:
```
    if (VVL_TRACY_CALLSTACK EQUALS "OFF") 
        set(VVL_TRACY_CALLSTACK 48)
    endif()
```

layers/CMakeLists.txt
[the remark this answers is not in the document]
spencer-lunarg: comment then please in the code, this looks like leftover stuff

layers/layer_options.cpp
spencer-lunarg: comment why we are leaving this

layers/layer_options.cpp
spencer-lunarg: if leaving it here, comment why we are leaving it here cause it didn't make sense to me

CMakeLists.txt
spencer-lunarg: ok, this works!

layers/layer_options.cpp
arno-lunarg: moved it to `profiling.cpp`

layers/CMakeLists.txt
```diff
@@ -417,6 +418,26 @@ target_link_libraries(vvl PRIVATE
...
 )
 
+if (VVL_ENABLE_TRACY)
+    # Rely on FetchContent to get Tracy, it is simple
+    # update_deps.py is too chaotic. Quite ok since Tracy is intended as a dev tool
+    include(FetchContent)
+    FetchContent_Declare(
+        tracy
+        GIT_REPOSITORY https://github.com/wolfpld/tracy.git
+        GIT_TAG 075395620a504c0cdcaf9bab3d196db16a043de7
```
spencer-lunarg, line 428: would just make `v0.11.0`

scripts/generators/layer_chassis_generator.py
```diff
@@ -1034,8 +1041,8 @@ def generateSource(self):
             }
 
             VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL GetInstanceProcAddr(VkInstance instance, const char* funcName) {
-                const auto& item = name_to_funcptr_map.find(funcName);
-                if (item != name_to_funcptr_map.end()) {
+                const auto& item = GetNameToFuncPtrMap().find(funcName);
```
spencer-lunarg, line 1044: we have to loop through every function with `GetInstanceProcAddr` and everytime we call `GetNameToFuncPtrMap()` we are having to reallocate the `static const vvl::unordered_map<std::string, function_data> name_to_func_ptr_map` map it seems on my machine

arno-lunarg:
Closing if favor of https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8322

Issue #8291, opened 2024-07-15
False-positive READ-AFTER-WRITE hazard and incorrect resource label


StefanPoelloth:
**Environment:**
 - OS: Windows 10 22H2
 - GPU and driver version: 2080Ti 556.12
 - SDK or header version if building from repo: d0a37c686 (and 1.3.283)
 - Options enabled (synchronization, best practices, etc.): synchronization, queue submit validation

**Describe the Issue**

I can provide an api dump privately if needed.
The message is invalid/incorrect, it specifies 2 different buffers 0x6eea0c0000000185 and 0x7d8104000000018d which are both alive (not destroyed). 
The prior_usage specifies that buffer 0x7d8104000000018d was used with vkCmdCopyBuffer, which is wrong.
Using the API dump I made sure that no vkCmdCopyBuffer was called for buffer 0x7d8104000000018d  (its a mapped buffer, updated with the mapped pointer and used in CmdDispatchIndirect).

**Expected behavior**

Either a correct sync hazard message or no message.

**Valid Usage ID**

validation layer: Validation Error: [ SYNC-HAZARD-READ-AFTER-WRITE ] Object 0: handle = 0x2d4ecb7c6c0, type = VK_OBJECT_TYPE_QUEUE; | MessageID = 0xe4d96472 | vkQueueSubmit():  Hazard READ_AFTER_WRITE for entry 0, VkCommandBuffer 0x2d4ef7fd510[], Submitted access info (submitted_usage: SYNC_COPY_TRANSFER_READ, command: vkCmdCopyBuffer, seq_no: 2, reset_no: 10, resource: VkBuffer 0x6eea0c0000000185[GpuArray MatrixStore _primaryArray]). Access info (prior_usage: SYNC_COPY_TRANSFER_WRITE, write_barriers: SYNC_VERTEX_SHADER_SHADER_STORAGE_READ|SYNC_COMPUTE_SHADER_SHADER_STORAGE_READ, queue: VkQueue 0x2d4ecb7c6c0[], submit: 22, batch: 0, batch_tag: 571, command: vkCmdCopyBuffer, command_buffer: VkCommandBuffer 0x2d4ef775860[], seq_no: 5, reset_no: 8, resource: VkBuffer 0x7d8104000000018d[PrimaryCullPass occludedCount]).

artem-lunarg:
[...]

PR #8317, opened 2024-07-24
🌴sync: Ensure handle index is valid

artem-lunarg:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8291 discovered two problems:
1) There is a potential syncval bug. Will be fixed in another PR, we have a deterministic repro case.
2) If syncval resource reporting invariants get invalidated this leads to crash/assert. The above syncval bug leads to such bad state and it will be fixed. But the same happens when we have unhandled core validation error (when core validation is disabled). The convention that syncval should handle this gracefully without crashing, although does not guarantee that reporting is correct.

This PR addresses problem 2.


spencer-lunarg (approved):
best PTO code I have ever seen

Issue #7562, opened 2024-02-22
Performance issue with debugPrintfEXT using Vulkan 1.1


SaschaWillems:
**Environment:**
 - OS: Windows 11
 - GPU and driver version: NVIDIA RTX 4070 538.42
 - SDK or header version if building from repo: SDK 1.3.275
 - Options enabled (synchronization, best practices, etc.): -

**Describe the Issue**

When using `debugPrintfEXT` along with `VkValidationFeaturesEXT` to fetch the debug output with a debug messenger callback, performance is so severely impacted, that the application is no longer interactive.

The setup:

Application:

```cpp
	std::vector<VkValidationFeatureEnableEXT>  validation_feature_enables = {VK_VALIDATION_FEATURE_ENABLE_DEBUG_PRINTF_EXT};
	std::vector<VkValidationFeatureDisableEXT> disables{};

	VkValidationFeaturesEXT validation_features{VK_STRUCTURE_TYPE_VALIDATION_FEATURES_EXT};
	validation_features.enabledValidationFeatureCount = 1;
	validation_features.pEnabledValidationFeatures    = validation_feature_enables.data();
	validation_features.pDisabledValidationFeatures   = 0;
	validation_features.pDisabledValidationFeatures   = disables.data();

	VkApplicationInfo app_info{VK_STRUCTURE_TYPE_APPLICATION_INFO};
	app_info.pApplicationName = "Shader debugprintf";
	app_info.pEngineName      = "Vulkan Samples";
	app_info.apiVersion       = VK_API_VERSION_1_2;

	std::vector<const char *> validation_layers = {"VK_LAYER_KHRONOS_validation"};

	VkInstanceCreateInfo instance_create_info{VK_STRUCTURE_TYPE_INSTANCE_CREATE_INFO};
	instance_create_info.ppEnabledExtensionNames = enabled_extensions.data();
	instance_create_info.enabledExtensionCount   = static_cast<uint32_t>(enabled_extensions.size());
	instance_create_info.pApplicationInfo        = &app_info;
	instance_create_info.pNext                   = &validation_features;
	instance_create_info.ppEnabledLayerNames     = validation_layers.data();
	instance_create_info.enabledLayerCount       = static_cast<uint32_t>(validation_layers.size());

	VkInstance vulkan_instance;
	result = vkCreateInstance(&instance_create_info, nullptr, &vulkan_instance);

[...]

PR #8318, opened 2024-07-24
layers: Remove Stub from dispatch table helper

spencer-lunarg:
Want to see if these `Stub` calls are needed

Currently they are the source of hidding a nasty bug (https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7562) when we call dispatch internally with the wrong version of the API

It might be better to hide the invalid function with a null call

### DIFF
--- a/layers/vulkan/generated/vk_dispatch_table_helper.cpp
+++ b/layers/vulkan/generated/vk_dispatch_table_helper.cpp
@@ -23,555 +23,8 @@
 
 // NOLINTBEGIN
 #include "vk_dispatch_table_helper.h"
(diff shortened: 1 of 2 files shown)


arno-lunarg:
Is it a draft?

spencer-lunarg:
Making a draft now

seems that this fails on Android (due to the Android Loader setup) and need to go do some git digging to understand why these Stub functions were added and if/how we can actually remove them

spencer-lunarg:
closing - Android loader issues preventing this from going in, maybe will try to come back to this one day

PR #8322, opened 2024-07-25
layers: Integrate Tracy profiler

arno-lunarg:
Cleaned up https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8313

See `profiling.md` for details

Only made one interesting change to the code outside of Tracy: Removed all global objects with a constructor that dynamically allocates memory, placed those objects inside dedicated access functions.
This approach allows to pay the cost of construction only when truly needed.


layers/profiling/profiling.cpp
```diff
@@ -0,0 +1,148 @@
...
+    free(ptr);
+}
+
+#if (__cplusplus > 201402L || defined(__cpp_aligned_new))
+
+#if defined(_WIN32)
+#define vvl_aligned_malloc(size, alignment) _aligned_malloc(size, alignment)
+#define vvl_aligned_free(ptr) _aligned_free(ptr)
+#else
+#define vvl_aligned_malloc(size, alignment) aligned_alloc(size, alignment)
```
spencer-lunarg, line 76: I haven't been able to figure out why this breaks things, a workaround of 

```patch
-#define vvl_aligned_malloc(size, alignment) aligned_alloc(size, alignment)
-#define vvl_aligned_free(ptr) free(ptr)
+ // TODO - Need to understand why on Linux sometimes calling aligned_alloc causes corruption
+void *vvl_aligned_malloc(std::size_t size, std::size_t al) {
+    void *mem = malloc(size+al+sizeof(void*));
+    void **ptr = (void**)((uintptr_t)((uintptr_t)mem+al+sizeof(void*)) & ~(al-1));
+    ptr[-1] = mem;
+    return ptr;
+}
+
+void vvl_aligned_free(void *ptr) {
+    free(((void**)ptr)[-1]);
+}
```

layers/profiling/profiling.cpp
```diff
@@ -0,0 +1,148 @@
...
+    }
+    return TRUE;
+}
+#else
+
+__attribute__((constructor)) static void so_attach(void) { tracy::StartupProfiler(); }
+
+#endif
+
+#endif
```
spencer-lunarg, line 146: with this many macros going on a `#endif // TRACY_MANUAL_LIFETIME` is helpful

layers/profiling/profiling.cpp
```diff
@@ -0,0 +1,148 @@
...
+ * limitations under the License.
+ */
+
+#include "profiling/profiling.h"
+
+#include <cstdlib>
+
+#if defined(TRACY_ENABLE)
+
+#include "profiling/profiling.h"
```
spencer-lunarg, line 24: there are 2 `#include "profiling/profiling.h"` in this file

layers/profiling/profiling.h
```diff
@@ -0,0 +1,44 @@
...
+
+#pragma once
+
+#if defined(TRACY_ENABLE)
+#include "tracy/Tracy.hpp"
+#include "tracy/TracyC.h"
+#include "tracy/../client/TracyProfiler.hpp"
+
+#define VVL_ZoneScoped ZoneScoped
+#define VVL_TracyCZone(zone_name, active) TracyCZone(zone_name, active)
```
spencer-lunarg, line 26: these don't work for me, I think it is because it already in a zone

using `ZoneNamedN` worked for me and could easily search it

scripts/generators/layer_chassis_generator.py
```diff
@@ -1097,6 +1099,7 @@ def generateSource(self):
 
             VKAPI_ATTR VkResult VKAPI_CALL CreateInstance(const VkInstanceCreateInfo* pCreateInfo, const VkAllocationCallbacks* pAllocator,
                                                         VkInstance* pInstance) {
+                VVL_ZoneScoped;
```
spencer-lunarg, line 1102: (ya I know the `chassis.cpp` code is awful for this) but you missed the `VVL_ZoneScoped` in the manual chassis hooks such as `CreateGraphicsPipelines`

scripts/generators/layer_chassis_generator.py
```diff
@@ -1226,6 +1229,7 @@ def generateSource(self):
             }
 
             VKAPI_ATTR void VKAPI_CALL DestroyInstance(VkInstance instance, const VkAllocationCallbacks* pAllocator) {
+                VVL_TracyCZone(tracy_zone_precall, true);
```
spencer-lunarg, line 1232: (ya I know the `chassis.cpp` code is awful for this) but you missed the `VVL_TracyCZone` in the manual chassis hooks such as `CreateGraphicsPipelines`

scripts/generators/layer_chassis_generator.py
```diff
@@ -1625,14 +1639,17 @@ def generateSource(self):
 
                 // Special extra check if SPIR-V itself fails runtime validation in PreCallRecord
                 if (chassis_state.skip) return VK_ERROR_VALIDATION_FAILED_EXT;
-
+                VVL_TracyCZoneEnd(tracy_zone_precall);
+                VVL_TracyCZone(tracy_zone_dispatch, true);
```
spencer-lunarg, line 1643: you have these different zones, but it just ends up being 3 `CreateShaderModule` in the profiler and I was confused until now and thought it was just tracy being buggy

there a way to color them (if naming is too much)

arno-lunarg: I moved away from The `C` zones: having to manually end a zone is prone to errors, as you have to take care of early returns. I just scoped the `PreCallValidate`, `PreCallRecord`, `Dispatch` and `PostCallRecord` loops/calls, and used `ZoneScopedN` there, so that with RAII I do not need to care about properly freeing stuff

spencer-lunarg (approved):

PR #8323, opened 2024-07-25
gpu: Fix a race condition in submission processing

### LINKED ISSUE #7562: Performance issue with debugPrintfEXT using Vulkan 1.1

jeremyg-lunarg:
Move most submission post processing into the PostCallRecord hooks. This avoids adding submission state in the PreCall hook and then looking for it in the PostCall hook. It also avoids adding bad state in case  the submit call fails.

The remaining work in the PreCall hook is setting up resources used by the instrumented shaders we're about to execute and that must happen  before the submit is dispatched.

This problem was hidden by sometimes calling the wrong version (core vs KHR) of vkWaitSemaphores(), which would end up calling a no-op stub set up elsewhere in VVL (and being fixed separately). So if we're using vulkan 1.2 or greater we use the core versions, if we're in 1.1 and have enabled the extensions, use the KHR versions.

Fixes #7562

### DIFF
--- a/layers/gpu/core/gpu_state_tracker.cpp
+++ b/layers/gpu/core/gpu_state_tracker.cpp
@@ -25,9 +25,10 @@ CommandBuffer::CommandBuffer(gpu::GpuShaderInstrumentor &shader_instrumentor, Vk
     : vvl::CommandBuffer(shader_instrumentor, handle, pCreateInfo, pool) {}
 
 Queue::Queue(gpu::GpuShaderInstrumentor &shader_instrumentor, VkQueue q, uint32_t family_index, uint32_t queue_index,
-             VkDeviceQueueCreateFlags flags, const VkQueueFamilyProperties &queueFamilyProperties)
+             VkDeviceQueueCreateFlags flags, const VkQueueFamilyProperties &queueFamilyProperties, bool timeline_khr)
     : vvl::Queue(shader_instrumentor, q, family_index, queue_index, flags, queueFamilyProperties),
-      shader_instrumentor_(shader_instrumentor) {}
+      shader_instrumentor_(shader_instrumentor),
+      timeline_khr_(timeline_khr) {}
 
 Queue::~Queue() {
     if (barrier_command_buffer_) {
@@ -153,7 +154,12 @@ void Queue::Retire(vvl::QueueSubmission &submission) {
         wait_info.semaphoreCount = 1;
         wait_info.pSemaphores = &barrier_sem_;
         wait_info.pValues = &submission.seq;
-        DispatchWaitSemaphoresKHR(shader_instrumentor_.device, &wait_info, 1000000000);
+
+        if (timeline_khr_) {
+            DispatchWaitSemaphoresKHR(shader_instrumentor_.device, &wait_info, 1000000000);
+        } else {
+            DispatchWaitSemaphores(shader_instrumentor_.device, &wait_info, 1000000000);
+        }
 
         for (auto &cbs : retiring_) {
             for (auto &cb : cbs) {
--- a/layers/gpu/core/gpu_state_tracker.cpp
+++ b/layers/gpu/core/gpu_state_tracker.cpp
@@ -28,7 +28,19 @@ Queue::Queue(gpu::GpuShaderInstrumentor &shader_instrumentor, VkQueue q, uint32_
              VkDeviceQueueCreateFlags flags, const VkQueueFamilyProperties &queueFamilyProperties, bool timeline_khr)
     : vvl::Queue(shader_instrumentor, q, family_index, queue_index, flags, queueFamilyProperties),
       shader_instrumentor_(shader_instrumentor),
-      timeline_khr_(timeline_khr) {}
+      timeline_khr_(timeline_khr) {
+    VkSemaphoreTypeCreateInfoKHR semaphore_type_create_info = vku::InitStructHelper();
+    semaphore_type_create_info.semaphoreType = VK_SEMAPHORE_TYPE_TIMELINE_KHR;
+    semaphore_type_create_info.initialValue = 0;
+
+    VkSemaphoreCreateInfo semaphore_create_info = vku::InitStructHelper(&semaphore_type_create_info);
+
+    auto result = DispatchCreateSemaphore(shader_instrumentor_.device, &semaphore_create_info, nullptr, &barrier_sem_);
+    if (result != VK_SUCCESS) {
+        Location loc(vvl::Func::vkGetDeviceQueue);
+        shader_instrumentor_.InternalError(shader_instrumentor_.device, loc, "Unable to create barrier semaphore.");
+    }
+}
 
 Queue::~Queue() {
     if (barrier_command_buffer_) {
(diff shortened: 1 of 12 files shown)


layers/gpu/core/gpu_state_tracker.cpp
```diff
@@ -25,9 +25,35 @@ CommandBuffer::CommandBuffer(gpu::GpuShaderInstrumentor &shader_instrumentor, Vk
     : vvl::CommandBuffer(shader_instrumentor, handle, pCreateInfo, pool) {}
 
 Queue::Queue(gpu::GpuShaderInstrumentor &shader_instrumentor, VkQueue q, uint32_t family_index, uint32_t queue_index,
-             VkDeviceQueueCreateFlags flags, const VkQueueFamilyProperties &queueFamilyProperties)
+             VkDeviceQueueCreateFlags flags, const VkQueueFamilyProperties &queueFamilyProperties, bool timeline_khr)
     : vvl::Queue(shader_instrumentor, q, family_index, queue_index, flags, queueFamilyProperties),
+<<<<<<< HEAD
```
spencer-lunarg, line 30: image

layers/gpu/core/gpu_state_tracker.cpp
```diff
@@ -25,9 +25,24 @@ CommandBuffer::CommandBuffer(gpu::GpuShaderInstrumentor &shader_instrumentor, Vk
...
+      shader_instrumentor_(shader_instrumentor),
+      timeline_khr_(timeline_khr) {
+    VkSemaphoreTypeCreateInfoKHR semaphore_type_create_info = vku::InitStructHelper();
+    semaphore_type_create_info.semaphoreType = VK_SEMAPHORE_TYPE_TIMELINE_KHR;
+    semaphore_type_create_info.initialValue = 0;
+
+    VkSemaphoreCreateInfo semaphore_create_info = vku::InitStructHelper(&semaphore_type_create_info);
+
+    auto result = DispatchCreateSemaphore(shader_instrumentor_.device, &semaphore_create_info, nullptr, &barrier_sem_);
+    assert(result == VK_SUCCESS);
```
spencer-lunarg, line 39: we don't need this assert if we have the full blown `InternalError` to catch it below

layers/gpu/core/gpu_state_tracker.cpp
```diff
@@ -25,9 +25,24 @@ CommandBuffer::CommandBuffer(gpu::GpuShaderInstrumentor &shader_instrumentor, Vk
...
+    if (result != VK_SUCCESS) {
+        Location loc(vvl::Func::vkGetDeviceQueue);
+        shader_instrumentor_.InternalError(shader_instrumentor_.device, loc, "Unable to create barrier semaphore. result=%d",
+                                           result);
```
spencer-lunarg, line 43: 1. `string_VkResult()`
2. `InternalError` is not a variadic function, `result` here is setting a boolean

layers/gpu/core/gpu_state_tracker.cpp
```diff
@@ -116,44 +116,38 @@ void Queue::SubmitBarrier(const Location &loc, uint64_t seq) {
         submit_info.signalSemaphoreCount = 1;
         submit_info.pSignalSemaphores = &barrier_sem_;
 
-        DispatchQueueSubmit(vvl::Queue::VkHandle(), 1, &submit_info, VK_NULL_HANDLE);
+        auto result = DispatchQueueSubmit(vvl::Queue::VkHandle(), 1, &submit_info, VK_NULL_HANDLE);
+        assert(result == VK_SUCCESS);
```
spencer-lunarg, line 120: CI is about to barf that `result` is unused in release build

layers/gpu/core/gpu_state_tracker.cpp
```diff
@@ -116,44 +116,38 @@ void Queue::SubmitBarrier(const Location &loc, uint64_t seq) {
...
     vvl::Queue::Retire(submission);
     retiring_.emplace_back(submission.cbs);
     if (submission.end_batch) {
         VkSemaphoreWaitInfo wait_info = vku::InitStructHelper();
         wait_info.semaphoreCount = 1;
         wait_info.pSemaphores = &barrier_sem_;
-        wait_info.pValues = &submission.seq;
-        DispatchWaitSemaphoresKHR(shader_instrumentor_.device, &wait_info, 1000000000);
+        wait_info.pValues = &wait_seq;
+        VkResult result;
```
spencer-lunarg, line 144: CI is about to barf that `result` is unused in release build

probably good to replace assert with `InternalError` call

spencer-lunarg:
@jeremyg-lunarg confirmed this change fixes the issue!

layers/gpu/core/gpu_state_tracker.cpp
jeremyg-lunarg: bleah

jeremyg-lunarg: I've never actually seen this fail so I'm going to skip the pain of printing the value of result.

layers/gpu/core/gpu_state_tracker.cpp
jeremyg-lunarg: oh i missed this one. That is a good idea.

spencer-lunarg (approved):

PR #8324, opened 2024-07-25
layers: Use emplace_back in DescriptorSet

spencer-lunarg:
small perf fix found from profiling around


arno-lunarg:
It makes a difference?? push_back should just do an inlined call to emplace_back anyway

arno-lunarg (approved):

spencer-lunarg:
> It makes a difference?? push_back should just do an inlined call to emplace_back anyway

for me I saw it go from 2 allocations to zero

Issue #8224, opened 2024-07-01
O(n^2) algorithm in `CoreChecks::RecordCmdCopyBuffer`


Axel-Reactor:
**Environment:**
 - OS: doesn't matter
 - GPU and driver version: doesn't matter
 - SDK or header version if building from repo: 1.3.283.0
 - Options enabled (synchronization, best practices, etc.): just the defaults

**Describe the Issue**

The checks in `CoreChecks::RecordCmdCopyBuffer` effectively hang the validation layers when calling `vkCmdCopyBuffer` with ~100 of regions.

**Expected behavior**

`O(log(n))`

spencer-lunarg:
@arno-lunarg has this been fixed in the refactor from last SDK?

arno-lunarg:
not at all, I will take a look but for now it is not a priority

Novum:
This was for buffer ranges, not texels.

jeremyg-lunarg:
@arno-lunarg I think you can make this better by changing the inner loop of ` CoreChecks::RecordCmdCopyBuffer`

```
    for (uint32_t i = 0; i < regionCount; ++i) {
         const auto &src = src_ranges[i]; 
         for (uint32_t j = 0; j < regionCount; ++j) {  
```
Starting from j = i + 1 should give the same results.

Maybe it is possible to get to O(log(n)) as suggested but that would require more than 5 minutes of pondering, at least for me.

arno-lunarg:
Ah wow I just confused myself, it is not gpu-av related at all.... Sorry, I will try to take a look this week!

arno-lunarg:
Also since it is a perf problem in core validation, it now becomes a priority

spencer-lunarg:
sorry, want to take blame, I labeled this as GPU-AV and that rightfully confused Arno... sorry, Arno has profiling setup now so should be able to get this improved!

arno-lunarg:
I made some optimisations, hopefully things get better on your hand. If they don't, please feel free to re-open the issues. Again sorry for the brain fart I had, I should have worked on this earlier.

PR #8325, opened 2024-07-26
layers: Improve perf for sparse buffer copies validation

arno-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8224

Stress test:
For 10,000 copies, went down from +15 minutes (stopped measured) down to 60 ms
For 4000, went down from +6 minutes to 10ms

For the stress test, the source and destination buffer are bound to only one VkDeviceMemory, I need to make another one where they are bound to multiples to see if perf is bad

~~Need to port this logic to the non sparse buffer case, will make another PR as users do not appear to be blocked on that~~ done, see `BufferCopiesValidationStressTest2 `

### DIFF
--- a/layers/core_checks/cc_copy_blit_resolve.cpp
+++ b/layers/core_checks/cc_copy_blit_resolve.cpp
@@ -1826,41 +1826,68 @@ void CoreChecks::RecordCmdCopyBuffer(VkCommandBuffer commandBuffer, VkBuffer src
     auto dst_buffer_state = Get<vvl::Buffer>(dstBuffer);
     ASSERT_AND_RETURN(src_buffer_state && dst_buffer_state);
 
-    if (src_buffer_state->sparse || dst_buffer_state->sparse) {
+    if (regionCount > 0 && (src_buffer_state->sparse || dst_buffer_state->sparse)) {
         auto cb_state = Get<vvl::CommandBuffer>(commandBuffer);
 
-        std::vector<sparse_container::range<VkDeviceSize>> src_ranges;
-        std::vector<sparse_container::range<VkDeviceSize>> dst_ranges;
+        using BufferRange = vvl::BindableMemoryTracker::BufferRange;
+
+        std::vector<BufferRange> src_ranges(regionCount);
+        std::vector<BufferRange> dst_ranges(regionCount);
+        BufferRange src_ranges_bounds(pRegions[0].srcOffset, pRegions[0].srcOffset + pRegions[0].size);
+        BufferRange dst_ranges_bounds(pRegions[0].dstOffset, pRegions[0].dstOffset + pRegions[0].size);
 
         for (uint32_t i = 0; i < regionCount; ++i) {
             const RegionType &region = pRegions[i];
-            src_ranges.emplace_back(sparse_container::range<VkDeviceSize>{region.srcOffset, region.srcOffset + region.size});
-            dst_ranges.emplace_back(sparse_container::range<VkDeviceSize>{region.dstOffset, region.dstOffset + region.size});
+            src_ranges[i] = sparse_container::range<VkDeviceSize>{region.srcOffset, region.srcOffset + region.size};
+            dst_ranges[i] = sparse_container::range<VkDeviceSize>{region.dstOffset, region.dstOffset + region.size};
+
+            src_ranges_bounds.begin = std::min(src_ranges_bounds.begin, region.srcOffset);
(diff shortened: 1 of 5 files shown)


layers/state_tracker/device_memory_state.cpp
```diff
@@ -192,6 +198,59 @@ BoundMemoryRange vvl::BindableSparseMemoryTracker::GetBoundMemoryRange(const Mem
     return mem_ranges;
 }
 
+BoundRanges vvl::BindableSparseMemoryTracker::GetBoundRanges(const BufferRange &ranges_bounds,
+                                                             const std::vector<BufferRange> &buffer_ranges) const {
+    BoundRanges memory_to_bound_ranges_map;
+    auto guard = ReadLockGuard{binding_lock_};
+
+    auto bound_memory_ranges = binding_map_.bounds(ranges_bounds);
```
arno-lunarg, line 206: I think most of the gains comes from the fact that `bounds` will be called only twice (once for the source ranges, once for the destination ranges), when it used to be called for every pair of range

layers/state_tracker/device_memory_state.h
```diff
@@ -149,6 +155,10 @@ class BindableLinearMemoryTracker : public BindableMemoryTracker {
                     VkDeviceSize resource_offset, VkDeviceSize size) override;
 
     BoundMemoryRange GetBoundMemoryRange(const MemoryRange &range) const override;
+    BoundRanges GetBoundRanges(const BufferRange &ranges_bounds, const std::vector<BufferRange> &ranges) const override {
```
arno-lunarg, line 158: ~~GetBoundRanges has for now has no reason to be called on non sparse memory~~ Well it will soon

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -1819,48 +1821,78 @@ void CoreChecks::PostCallRecordCmdCopyImage2(VkCommandBuffer commandBuffer, cons
 template <typename RegionType>
 void CoreChecks::RecordCmdCopyBuffer(VkCommandBuffer commandBuffer, VkBuffer srcBuffer, VkBuffer dstBuffer, uint32_t regionCount,
                                      const RegionType *pRegions, const Location &loc) {
+    VVL_ZoneScoped;
```
spencer-lunarg, line 1824: we plan to keep this here?

layers/profiling/profiling.h
```diff
@@ -22,19 +22,49 @@
...
+// Define CPU zones
 #define VVL_ZoneScoped ZoneScoped
 #define VVL_ZoneScopedN(name) ZoneScopedN(name)
 #define VVL_TracyCZone(zone_name, active) TracyCZone(zone_name, active)
 #define VVL_TracyCZoneEnd(zone_name) TracyCZoneEnd(zone_name)
 #define VVL_TracyCFrameMark TracyCFrameMark
+
+// Print messages
 #define VVL_TracyMessage TracyMessage
+#define VVL_TracyMessageStream(message)                \
```
spencer-lunarg, line 34: this should be in a separate commit IMO

layers/core_checks/cc_copy_blit_resolve.cpp
arno-lunarg: Could remove yes

layers/profiling/profiling.h
arno-lunarg: ok!

layers/core_checks/cc_copy_blit_resolve.cpp
arno-lunarg: can remove

spencer-lunarg: thanks, same as below, I really don't want to start having random `VVL_ZoneScopedN` scatter everywhere

arno-lunarg:
On my AMD laptot, new test `PositiveSparseBuffer, BufferCopiesValidationStressTest2` is causing a `vk_timeout` when waiting for the fence after submitting memory bindings, but this timeout is hit immediately, not after the supplied delay of 10 seconds. No validation error. Can't see what is wrong with the test

layers/state_tracker/device_memory_state.h
spencer-lunarg: maybe worth adding a `todo` comment here incase you don't get back to fixing this soon and someone just sees this random function that shouldn't exists and get confused

spencer-lunarg (approved):

PR #8332, opened 2024-07-29
layers: Use emplace_back where possible

spencer-lunarg:
Basic attempt to use `emplace_back` where it should have an obvious benefits

prepared to criticism and "well technically in c++... " responses :shield:

### DIFF
--- a/layers/error_message/logging.cpp
+++ b/layers/error_message/logging.cpp
@@ -343,7 +343,7 @@ void DebugReport::BeginQueueDebugUtilsLabel(VkQueue queue, const VkDebugUtilsLab
     if (nullptr != label_info && nullptr != label_info->pLabelName) {
         auto *label_state = GetLoggingLabelState(&debug_utils_queue_labels, queue, /* insert */ true);
         assert(label_state);
-        label_state->labels.emplace_back(label_info);
+        label_state->labels.push_back(label_info);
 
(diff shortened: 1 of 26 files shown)


arno-lunarg (approved):
image

layers/error_message/logging.cpp
```diff
@@ -377,7 +377,7 @@ void DebugReport::BeginCmdDebugUtilsLabel(VkCommandBuffer command_buffer, const
     if (nullptr != label_info && nullptr != label_info->pLabelName) {
         auto *label_state = GetLoggingLabelState(&debug_utils_cmd_buffer_labels, command_buffer, /* insert */ true);
         assert(label_state);
-        label_state->labels.push_back(LoggingLabel(label_info));
+        label_state->labels.emplace_back(label_info);
```
artem-lunarg, line 380: IMO this gets less readable. Before it was clear we put LoggingLabel objects, now the first impression we put VkDebugUtilsLabelEXT pointers there.

One of the reasons of introduction of emplace methods is to facilitate meta programming where specific types are not known or even not visible and require decltype.  In that case emplace does not reduce code clarity.

arno-lunarg: 🌴

spencer-lunarg:
ok, no more spending time worrying about `emplace_back` vs `push_back` for at least a year now :upside_down_face:

Issue #8328, opened 2024-07-26
Add Sparse VU 09491 09492


arno-lunarg:
`VUID-VkSparseMemoryBind-resourceOffset-09491`
If the resource being bound is a VkBuffer, resourceOffset, memoryOffset and size must be an integer multiple of the alignment of the [VkMemoryRequirements](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkMemoryRequirements.html) structure returned from a call to [vkGetBufferMemoryRequirements](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetBufferMemoryRequirements.html) with the buffer resource

`VUID-VkSparseMemoryBind-resourceOffset-09492`
If the resource being bound is a VkImage, resourceOffset and memoryOffset must be an integer multiple of the alignment of the [VkMemoryRequirements](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkMemoryRequirements.html) structure returned from a call to [vkGetImageMemoryRequirements](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetImageMemoryRequirements.html) with the image resource

spencer-lunarg:
note this was added in https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6323 and touched up a bit in https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6555

These are the same idea as VUs such as `VUID-vkBindBufferMemory-memoryOffset-01036`

arno-lunarg:
you should have the title of "VUID archeologist"

PR #8333, opened 2024-07-29
Arno add 09491 09492

arno-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8328


layers/core_checks/cc_device_memory.cpp
```diff
@@ -2094,6 +2094,67 @@ void CoreChecks::PostCallRecordBindImageMemory2KHR(VkDevice device, uint32_t bin
     PostCallRecordBindImageMemory2(device, bindInfoCount, pBindInfos, record_obj);
 }
 
+bool CoreChecks::ValidateBufferSparseMemoryBindAlignments(const VkSparseMemoryBind &bind, const vvl::Buffer &buffer,
+                                                          const Location &bind_loc, const Location &buffer_bind_info_loc) const {
+    bool skip = false;
+    auto mem_state = Get<vvl::DeviceMemory>(bind.memory);
+    if (!mem_state) return skip;
```
spencer-lunarg, line 2101: maybe a comment here, `mem_state` is not used so not 100% sure why we need to return here... we want a helper call like `IsBound` or something ... or a comment is good... or is this actually not needed?

spencer-lunarg (approved):
happy when CI is happy

PR #8335, opened 2024-07-29
Combined and improve InstrumentShader

spencer-lunarg:
This is a series of commits that merge the `InstrumentShader` between GPU-AV and DebugPrintf

In the process I also removed the extra copy of SPIR-V so when we instrument the shader, we pass the `vvl::span` down and never need to copy it to a `std::vector<uint32_t>` (which we needed for `spirv-opt` in the past)


arno-lunarg:
image

arno-lunarg (approved):

PR #8337, opened 2024-07-29
printf: Add stage info to verbose info

spencer-lunarg:
Adds back the stage info to `DebugPrintf` when printing the verbose mode

Uses same system as GPU-AV to print it


layers/gpu/debug_printf/debug_printf.cpp
```diff
@@ -314,18 +314,10 @@ std::string Validator::FindFormatString(const std::vector<spirv::Instruction> &i
 
 void Validator::AnalyzeAndGenerateMessage(VkCommandBuffer command_buffer, VkQueue queue, BufferInfo &buffer_info,
                                           uint32_t operation_index, uint32_t *const debug_output_buffer, const Location &loc) {
-    // Word         Content
```
spencer-lunarg, line 317: this was already outdated, `OutputRecord` has the correct info

layers/gpu/shaders/gpu_error_header.h
```diff
@@ -63,56 +63,12 @@ const int kHeaderInstructionIdOffset = 2;
...
-const int kHeaderStageInfoOffset_2 = 6;
-
-// Stage-specific Error record offsets
-// ---
 // Each stage will contain different values in the next set of words of the
 // record used to identify which instantiation of the shader generated the
 // validation error.
-//
-// Vertex Shader Output Record Offsets
-const int kHeaderVertexIndexOffset = kHeaderStageInfoOffset_0;
```
spencer-lunarg, line 77: these alias "seems" good in theory, but are actually quite useless... you will still end up reading the contents of `GenerateStageMessage` to know what is what

layers/gpu/debug_printf/debug_printf.cpp
```diff
@@ -374,36 +366,39 @@ void Validator::AnalyzeAndGenerateMessage(VkCommandBuffer command_buffer, VkQueu
...
-                        needed = std::snprintf(nullptr, 0, substring.string.c_str(), *static_cast<uint32_t *>(values)) + 1;
+                        const uint32_t value = *static_cast<uint32_t *>(current_value);
+                        needed = std::snprintf(nullptr, 0, substring.string.c_str(), value) + 1;
                         temp_string.resize(needed);
-                        std::snprintf(&temp_string[0], needed, substring.string.c_str(), *static_cast<uint32_t *>(values));
+                        std::snprintf(&temp_string[0], needed, substring.string.c_str(), value);
 
                     } else if (substring.type == NumericTypeSint) {
-                        needed = std::snprintf(nullptr, 0, substring.string.c_str(), *static_cast<int32_t *>(values)) + 1;
+                        const uint32_t value = *static_cast<int32_t *>(current_value);
```
arno-lunarg, line 387:
```suggestion
                        const int32_t value = *static_cast<int32_t *>(current_value);
```

layers/gpu/instrumentation/gpu_shader_instrumentor.cpp
```diff
@@ -1022,16 +1023,98 @@ static bool GetLineAndFilename(const std::string &string, uint32_t *linenumber,
...
+               << stage_info_2 << "). ";
+        } break;
+        case spv::ExecutionModelGeometry: {
+            ss << "Stage = Geometry.  Primitive ID = " << stage_info_0 << " Invocation ID = " << stage_info_1 << ". ";
+        } break;
+        case spv::ExecutionModelFragment: {
+            float x_coord;
+            float y_coord;
+            std::memcpy(&x_coord, &stage_info_0, sizeof(float));
+            std::memcpy(&y_coord, &stage_info_1, sizeof(float));
```
arno-lunarg, line 1050:
```suggestion
            const float x_coord = std::bit_cast<float>(stage_info_0);
            const float y_coord = std::bit_cast<float>(stage_info_1);
```

arno-lunarg (changes requested):
One typo for `int32_t` that needs to be fixed, otherwise LGTM!

layers/gpu/instrumentation/gpu_shader_instrumentor.cpp
spencer-lunarg: aww, figured this was a better way to do this (wrote a test to make sure it works too)

spencer-lunarg: wait, nope this is a c++ 20 only feature

PR #8338, opened 2024-07-30
Improve buffer copy validation perf

arno-lunarg:
Took care of the sparse buffer case, this PR modifies validation for non sparse buffers.
For 1000 copies, went from **24 seconds** down to **0.5 ms**. Perf gains seem even better than what I got in the sparse case

### DIFF
--- a/layers/core_checks/cc_copy_blit_resolve.cpp
+++ b/layers/core_checks/cc_copy_blit_resolve.cpp
@@ -685,7 +685,6 @@ bool CoreChecks::ValidateCmdCopyBufferBounds(VkCommandBuffer cb, const vvl::Buff
                                              const Location &loc) const {
     bool skip = false;
     const bool is_2 = loc.function == Func::vkCmdCopyBuffer2 || loc.function == Func::vkCmdCopyBuffer2KHR;
-    const char *vuid;
 
     VkDeviceSize src_buffer_size = src_buffer_state.create_info.size;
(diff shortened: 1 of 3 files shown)


layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -685,37 +685,54 @@ bool CoreChecks::ValidateCmdCopyBufferBounds(VkCommandBuffer cb, const vvl::Buff
...
     const LogObjectList src_objlist(cb, dst_buffer_state.Handle());
     const LogObjectList dst_objlist(cb, dst_buffer_state.Handle());
+
+    const auto *src_binding = src_buffer_state.Binding();
+    const auto *dst_binding = dst_buffer_state.Binding();
+
+    const bool validate_no_memory_overlaps =
+        !are_buffers_sparse && (regionCount > 0) && src_binding && (src_binding->memory_state == dst_binding->memory_state);
+
+    auto cb_state = Get<vvl::CommandBuffer>(cb);
```
spencer-lunarg, line 702: nit: `ValidateCmdCopyBufferBounds` should just take a `const vvl::CommandBuffer &cb_state`

arno-lunarg: yes leftover!

spencer-lunarg (approved):

PR #8339, opened 2024-07-30
gpu: Improve naming between shader for Descriptor Indexing

spencer-lunarg:
So this was me fully trying to understand the logic in `bindless_descriptor.comp` and where it mapped to

I moved things out of `gpuav_subclasses.h` that didn't deal with state objects

I created a `gpu_shader_resources.h` that defines all the structures/data that we use in the GLSL

I updated `bindless_descriptor.comp` with some comments to help link where in the source code all the things get mapped to for easier search in the future


jeremyg-lunarg (approved):
LGTM

PR #8341, opened 2024-07-31
printf: Support 64-bit floats

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7143

### DIFF
--- a/layers/gpu/debug_printf/debug_printf.cpp
+++ b/layers/gpu/debug_printf/debug_printf.cpp
@@ -344,14 +344,14 @@ void Validator::AnalyzeAndGenerateMessage(VkCommandBuffer command_buffer, VkQueu
         auto format_substrings = ParseFormatString(format_string);
         void *current_value = static_cast<void *>(&debug_record->values);
         // Sprintf each format substring into a temporary string then add that to the message
-        for (auto &substring : format_substrings) {
+        for (size_t i = 0; i < format_substrings.size(); i++) {
+            auto &substring = format_substrings[i];
             std::string temp_string;
             size_t needed = 0;
 
             if (substring.needs_value) {
                 if (substring.is_64_bit) {
                     assert(substring.type != NumericTypeSint);  // not supported
-
                     if (substring.type == NumericTypeUint) {
                         std::vector<std::string> format_strings = {"%ul", "%lu", "%lx"};
                         for (const auto &ul_string : format_strings) {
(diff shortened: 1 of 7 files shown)


layers/gpu/spirv/module.cpp
```diff
@@ -635,5 +637,8 @@ void Module::PostProcess() {
     }
 }
 
+// TODO - Currently this is not tied into validation layers callback, should be using LogWarning/LogError
+void InternalWarning(const char* message) { std::cout << "[Internal Shader Instrumentation Warning] " << message << '\n'; }
```
spencer-lunarg, line 641: next PR after this will be to hook in the call back so I can move all the validation from https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8294 into here as it should be done in this part of the code anyway

layers/gpu/debug_printf/debug_printf.cpp
```diff
@@ -344,14 +344,14 @@ void Validator::AnalyzeAndGenerateMessage(VkCommandBuffer command_buffer, VkQueu
         auto format_substrings = ParseFormatString(format_string);
         void *current_value = static_cast<void *>(&debug_record->values);
         // Sprintf each format substring into a temporary string then add that to the message
-        for (auto &substring : format_substrings) {
+        for (size_t i = 0; i < format_substrings.size(); i++) {
```
arno-lunarg, line 347: use `vvl::enumerate`

layers/gpu/spirv/debug_printf_pass.cpp
```diff
@@ -112,18 +115,31 @@ void DebugPrintfPass::CreateFunctionParams(uint32_t argument_id, const Type& arg
...
             } else if (width == 64) {
+                if (expanded_parameter_count_ > 31) {
+                    // It is very unlikely to hit this, if the user does, they can just split the very long printf() into 2 printf()
+                    // calls ontop of each other
+                    InternalWarning("More than 32 expanded parameters, can't properly detect 64-bit float");
+                }
+                double_bitmask_ |= 1 << expanded_parameter_count_;
+
+                // TODO (Also in GPU-AV) need to detect if shaderInt64 is supported or not before forcing this on
+                module_.AddCapability(spv::CapabilityInt64);
```
arno-lunarg, line 133: What happens if `CapabilityInt64` is not supported? Will the user get noticed?

arno-lunarg (approved):

layers/gpu/spirv/debug_printf_pass.cpp
spencer-lunarg: its just technically invalid SPIR-V and a VU like anything else

from what I could see, `Float64` is more rare than `Int64` so assume they would support this... but we should add a way to check

layers/gpu/debug_printf/debug_printf.cpp
spencer-lunarg: next PR is going to touch this so will fix it there (save a CI run)

PR #8345, opened 2024-08-01
layers: Handle sparse/not sparse buffer copies

arno-lunarg:
Handle copies between buffers where only one is sparse

### DIFF
--- a/tests/unit/sparse_buffer.cpp
+++ b/tests/unit/sparse_buffer.cpp
@@ -290,8 +290,6 @@ TEST_F(NegativeSparseBuffer, OverlappingBufferCopy) {
 
     vkt::DeviceMemory buffer_mem;
     buffer_mem.init(*m_device, buffer_mem_alloc);
-    vkt::DeviceMemory buffer_mem2;
-    buffer_mem2.init(*m_device, buffer_mem_alloc);
 
     VkSparseMemoryBind buffer_memory_bind = {};
     buffer_memory_bind.size = buffer_mem_reqs.size;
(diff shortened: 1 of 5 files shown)


tests/unit/sparse_buffer.cpp
```diff
@@ -510,6 +508,83 @@ TEST_F(NegativeSparseBuffer, OverlappingBufferCopy3) {
     sparse_queue->Wait();
 }
 
+TEST_F(NegativeSparseBuffer, OverlappingBufferCopy4) {
```
spencer-lunarg, line 511: `OverlappingBufferCopy2` and `OverlappingBufferCopy3` both had a positive test, not sure if relevant here or not

spencer-lunarg (approved):

tests/unit/sparse_buffer.cpp
```diff
@@ -510,6 +508,83 @@ TEST_F(NegativeSparseBuffer, OverlappingBufferCopy3) {
...
+    copy_info.size = 256;
+
+    VkBufferCreateInfo b_info =
+        vkt::Buffer::create_info(copy_info.size, VK_BUFFER_USAGE_TRANSFER_SRC_BIT | VK_BUFFER_USAGE_TRANSFER_DST_BIT, nullptr);
+    vkt::Buffer buffer_not_sparse(*m_device, b_info, vkt::no_mem);
+    b_info.flags = VK_BUFFER_CREATE_SPARSE_BINDING_BIT;
+    vkt::Buffer buffer_sparse(*m_device, b_info, vkt::no_mem);
+
+    VkMemoryRequirements buffer_mem_reqs;
+    vk::GetBufferMemoryRequirements(device(), buffer_sparse.handle(), &buffer_mem_reqs);
```
spencer-lunarg, line 535: seems `buffer_not_sparse.bind_memory(buffer_mem, 0);` is being bound with the sparse buffer queried memory

why it failed on Android CI

arno-lunarg: yes that's the goal. But this could totally be illegal, tried my luck with CI and yes Android is not happy, need to fix.

tests/unit/sparse_buffer.cpp
arno-lunarg: Could a positive test yes

PR #8351, opened 2024-08-02
layers: Remove unused includes

spencer-lunarg:
Installed the new Zed editor for Linux, tried it out, found some unused `#include` and spent some time removing them to hopefully improve compiling speed a bit (and keep things more tidy)

### DIFF
--- a/layers/core_checks/cc_drawdispatch.cpp
+++ b/layers/core_checks/cc_drawdispatch.cpp
@@ -20,7 +20,6 @@
 #include "generated/chassis.h"
 #include "drawdispatch/drawdispatch_vuids.h"
 #include "core_validation.h"
-#include "state_tracker/image_state.h"
 #include "state_tracker/buffer_state.h"
 #include "state_tracker/shader_object_state.h"
 #include "state_tracker/descriptor_sets.h"
--- a/layers/state_tracker/device_memory_state.cpp
+++ b/layers/state_tracker/device_memory_state.cpp
@@ -17,7 +17,6 @@
  * limitations under the License.
  */
 #include "state_tracker/device_memory_state.h"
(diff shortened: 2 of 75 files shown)


layers/core_checks/cc_drawdispatch.cpp
```diff
@@ -1738,4 +1737,4 @@ bool CoreChecks::ValidateDrawDualSourceBlend(const LastBound &last_bound_state,
     }
 
     return skip;
-}
\ No newline at end of file
+}
```
spencer-lunarg, line 1740: idk why git displays it as this, but @arno-lunarg seems Zed enforces the new-line at the end for you

layers/state_tracker/device_memory_state.cpp
```diff
@@ -373,4 +372,3 @@ void vvl::Bindable::CacheInvalidMemory() const {
         }
     }
 }
-
```
spencer-lunarg, line 376: even removes when there are 2 new lines at the end to just 1

layers/state_tracker/shader_stage_state.h
```diff
@@ -17,7 +17,8 @@
  */
 
 #pragma once
-#include "utils/vk_layer_utils.h"
```
spencer-lunarg, line 20: didn't have time, but we are bad at relying on things in `vk_layer_utils.h` to be consumed everywhere... makes touching that file awful as you have to rebuild everything

layers/core_checks/cc_drawdispatch.cpp
arno-lunarg: <proceeds to wipe out windows and immediately install linux with this code editor>

arno-lunarg (approved):

PR #8356, opened 2024-08-05
spirv: Rerun generate_spirv.py

spencer-lunarg:
the change in `glslang` from https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8340 just rearranged which order the `OpDecorate` were printed

from before

image

to after

image


jeremyg-lunarg (approved):
LGTM

PR #8357, opened 2024-08-05
spirv: Create dedicated InjectConditionalFunctionPass class

spencer-lunarg:
This creates a dedicated `InjectConditionalFunctionPass` and `InjectFunctionPass` to allow future passes to better separate the logic if it wants to use 

```
if (check_logic()) {
   // original code
} else {
   // safe code
}
```

vs

```
check_logic();
// original_code
```


jeremyg-lunarg (approved):
LGTM

PR #8174, opened 2024-06-20
gpu: Fix enabled_features

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8001

This correctly passes down the `VkDeviceCreateInfo` down to GPU so `enabled_features` works as expected

PR #8360, opened 2024-08-06
gpu: More cleanup around descriptor indexing

spencer-lunarg:
Some more cleanup separated in a separate PR

### DIFF
--- a/layers/gpu/cmd_validation/gpuav_cmd_validation_common.cpp
+++ b/layers/gpu/cmd_validation/gpuav_cmd_validation_common.cpp
@@ -115,29 +115,4 @@ void RestorablePipelineState::Restore() const {
     }
 }
 
-VkDeviceAddress GetBufferDeviceAddress(Validator &gpuav, VkBuffer buffer, const Location &loc) {
-    // https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8001
-    // Setting enabled_features.bufferDeviceAddress to true in GpuShaderInstrumentor::PreCallRecordCreateDevice
-    // when adding missing features will modify another validator object, one associated to VkInstance,
-    // and "this" validator is associated to a device. enabled_features is not inherited, and besides
-    // would be reset in GetEnabledDeviceFeatures.
-    // The switch from the instance validator object to the device one happens in
-    // `state_tracker.cpp`, `ValidationStateTracker::PostCallRecordCreateDevice`
-    // TL;DR is the following type of sanity check is currently invalid, but it would be nice to have
-    // assert(enabled_features.bufferDeviceAddress);
-
-    VkBufferDeviceAddressInfo address_info = vku::InitStructHelper();
-    address_info.buffer = buffer;
-    if (gpuav.api_version >= VK_API_VERSION_1_2) {
-        return DispatchGetBufferDeviceAddress(gpuav.device, &address_info);
-    }
-    if (IsExtEnabled(gpuav.device_extensions.vk_ext_buffer_device_address)) {
-        return DispatchGetBufferDeviceAddressEXT(gpuav.device, &address_info);
-    }
-    if (IsExtEnabled(gpuav.device_extensions.vk_khr_buffer_device_address)) {
-        return DispatchGetBufferDeviceAddressKHR(gpuav.device, &address_info);
-    }
-    return 0;
-}
-
 }  // namespace gpuav
--- a/layers/gpu/descriptor_validation/gpuav_descriptor_validation.cpp
+++ b/layers/gpu/descriptor_validation/gpuav_descriptor_validation.cpp
@@ -23,17 +23,12 @@
 #include "gpu/resources/gpu_shader_resources.h"
 
 namespace gpuav {
-void UpdateBoundPipeline(Validator &gpuav, VkCommandBuffer cb, VkPipelineBindPoint pipeline_bind_point, VkPipeline pipeline,
+void UpdateBoundPipeline(Validator &gpuav, CommandBuffer &cb_state, VkPipelineBindPoint pipeline_bind_point, VkPipeline pipeline,
                          const Location &loc) {
     if (!gpuav.gpuav_settings.validate_descriptors) return;
-    auto cb_state = gpuav.GetWrite<CommandBuffer>(cb);
-    if (!cb_state) {
-        gpuav.InternalError(cb, loc, "Unrecognized command buffer");
-        return;
-    }
 
     const auto lv_bind_point = ConvertToLvlBindPoint(pipeline_bind_point);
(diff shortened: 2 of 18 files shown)


layers/gpu/cmd_validation/gpuav_cmd_validation_common.cpp
```diff
@@ -116,29 +116,4 @@ void RestorablePipelineState::Restore() const {
     }
 }
 
-VkDeviceAddress GetBufferDeviceAddress(Validator &gpuav, VkBuffer buffer, const Location &loc) {
-    // https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8001
```
spencer-lunarg, line 120: this was all solved/cleaned-up in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8174

layers/gpu/resources/gpu_resources.h
```diff
@@ -116,6 +108,7 @@ class SharedResourcesManager {
         bool operator()(TypeInfoRef lhs, TypeInfoRef rhs) const { return lhs.get() == rhs.get(); }
     };
 
+    // Fails to compile on Windows with vvl
```
jeremyg-lunarg, line 111: what's up with this comment?

layers/gpu/resources/gpu_shader_resources.h
```diff
@@ -15,12 +15,15 @@
  * limitations under the License.
  */
 
+// This fill helps maps the layout of resources that we find in the instrumented shaders
```
jeremyg-lunarg, line 18: This file?

jeremyg-lunarg (commented):
code looks good. There's some unclear comments that need fixing tho

layers/gpu/resources/gpu_resources.h
spencer-lunarg: failed to save

layers/gpu/resources/gpu_shader_resources.h
spencer-lunarg: yes, fixed, you can replace Arno short term for my English Checker

jeremyg-lunarg (approved):

layers/gpu/instrumentation/gpu_shader_instrumentor.cpp
```diff
@@ -1096,6 +1096,19 @@ bool GpuShaderInstrumentor::InstrumentShader(const vvl::span<const uint32_t> &in
     return true;
 }
 
+VkDeviceAddress GpuShaderInstrumentor::GetBufferDeviceAddressHelper(VkBuffer buffer) const {
+    VkBufferDeviceAddressInfo address_info = vku::InitStructHelper();
+    address_info.buffer = buffer;
+
+    // We cannot rely on device_extensions here, since we may be enabling BDA support even
+    // though the application has not requested it.
```
arno-lunarg, line 1104: So per this comment, when we force enable an extension, this change is not reflected in `device_extensions`?
Seems counter intuitive

layers/gpu/descriptor_validation/gpuav_descriptor_validation.cpp
```diff
@@ -64,27 +59,26 @@ void UpdateBoundPipeline(Validator &gpuav, VkCommandBuffer cb, VkPipelineBindPoi
     }
 }
 
-void UpdateBoundDescriptors(Validator &gpuav, VkCommandBuffer cb, VkPipelineBindPoint pipeline_bind_point, const Location &loc) {
-    auto cb_state = gpuav.GetWrite<CommandBuffer>(cb);
-    if (!cb_state) {
-        gpuav.InternalError(cb, loc, "Unrecognized command buffer");
+void UpdateBoundDescriptors(Validator &gpuav, CommandBuffer &cb_state, VkPipelineBindPoint pipeline_bind_point,
+                            const Location &loc) {
+    if (!gpuav.gpuav_settings.validate_descriptors || !gpuav.force_buffer_device_address_) {
```
arno-lunarg, line 64: Not a fan of adding a check for `!gpuav.force_buffer_device_address_`, without context it looks arbitrary

We should force `gpuav.gpuav_settings.validate_descriptors` to false when conditions for validation are not met, and only care about checking for this setting to know whether or not we need to perform validation

arno-lunarg: We even confuse ourselves with that since in `UpdateBoundPipeline` just above we only check for `gpuav.gpuav_settings.validate_descriptors`

arno-lunarg (commented):
LGTM, at least one more thing to improve stands out for me

layers/gpu/instrumentation/gpu_shader_instrumentor.h
```diff
@@ -169,6 +169,8 @@ class GpuShaderInstrumentor : public ValidationStateTracker {
     void PreCallRecordQueueSubmit2(VkQueue queue, uint32_t submitCount, const VkSubmitInfo2 *pSubmits, VkFence fence,
                                    const RecordObject &record_obj) override;
 
+    VkDeviceAddress GetBufferDeviceAddressHelper(VkBuffer buffer) const;
```
arno-lunarg, line 172: This belongs as a free function, those GPU-AV classes are filled to the brim, only virtual functions, that cannot be moved elsewhere, should be declared here. 
I spent a lot of time cleaning this, I do not want to do it again!

layers/gpu/instrumentation/gpu_shader_instrumentor.cpp
spencer-lunarg: sorry I just copied this, `device_extensions` should work, will update this comment

layers/gpu/descriptor_validation/gpuav_descriptor_validation.cpp
spencer-lunarg: yep, already on this! (this is just the code we already had prior to this PR, I just moved it up to the top of the function where I noticed the same thing)

layers/gpu/instrumentation/gpu_shader_instrumentor.h
spencer-lunarg: sure, can move up, was greedy that didn't want to pass in the `VkDevice`

spencer-lunarg: no this makes sense in a `ValidationObject` because it needs the `api_version` as well... we have other `Dispatch Helpers` in these virtual functions, not the first time we have done this

arno-lunarg: just pass in `ValidationObject &gpuav`, no need for a method
I will die on this hill
tenor

PR #8361, opened 2024-08-06
gpu: Avoid multiple command buffer lookups while instrumenting

jeremyg-lunarg:
This avoids lookup and locking churn, and makes it more obvious that the state object is in use by various instrumentation subroutines.


spencer-lunarg (approved):
was just thinking about doing this change soon, glad you beat me to it

PR #8362, opened 2024-08-06
printf: Fix 8-bit 16-bit signed ints

spencer-lunarg:
As my original test showed, this was never working and now will print the correct values out


arno-lunarg (approved):
Nice thank you!

PR #8363, opened 2024-08-07
layers: Apply some clang-tidy fixes

spencer-lunarg:
using `Zed` took some time to look through clang-tidy warnings, a LOT were not worth fixing, but here are some good low-hanging fruit

### DIFF
--- a/layers/state_tracker/image_state.cpp
+++ b/layers/state_tracker/image_state.cpp
@@ -57,8 +57,8 @@ VkImageSubresourceRange NormalizeSubresourceRange(const VkImageCreateInfo &image
 }
 
 static bool IsDepthSliced(const VkImageCreateInfo &image_create_info, const VkImageViewCreateInfo &create_info) {
-    auto kDepthSlicedFlags = VK_IMAGE_CREATE_2D_ARRAY_COMPATIBLE_BIT | VK_IMAGE_CREATE_2D_VIEW_COMPATIBLE_BIT_EXT;
-    return ((image_create_info.flags & kDepthSlicedFlags) != 0) &&
+    auto depth_slice_flag = VK_IMAGE_CREATE_2D_ARRAY_COMPATIBLE_BIT | VK_IMAGE_CREATE_2D_VIEW_COMPATIBLE_BIT_EXT;
+    return ((image_create_info.flags & depth_slice_flag) != 0) &&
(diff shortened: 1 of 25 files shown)


layers/state_tracker/image_state.cpp
```diff
@@ -57,8 +57,8 @@ VkImageSubresourceRange NormalizeSubresourceRange(const VkImageCreateInfo &image
 }
 
 static bool IsDepthSliced(const VkImageCreateInfo &image_create_info, const VkImageViewCreateInfo &create_info) {
-    auto kDepthSlicedFlags = VK_IMAGE_CREATE_2D_ARRAY_COMPATIBLE_BIT | VK_IMAGE_CREATE_2D_VIEW_COMPATIBLE_BIT_EXT;
-    return ((image_create_info.flags & kDepthSlicedFlags) != 0) &&
+    auto depth_slice_flag = VK_IMAGE_CREATE_2D_ARRAY_COMPATIBLE_BIT | VK_IMAGE_CREATE_2D_VIEW_COMPATIBLE_BIT_EXT;
```
artem-lunarg, line 60: does it mean we stop using "k" convention for constants due to underscore style?

layers/state_tracker/image_state.h
```diff
@@ -99,10 +99,10 @@ class Image : public Bindable {
     const VkFormatFeatureFlags2KHR format_features;
     // Need to memory requirments for each plane if image is disjoint
     const bool disjoint;  // True if image was created with VK_IMAGE_CREATE_DISJOINT_BIT
-    static constexpr int MAX_PLANES = 3;
-    using MemoryReqs = std::array<VkMemoryRequirements, MAX_PLANES>;
+    static constexpr int kMaxPlanes = 3;
```
artem-lunarg, line 102: attaching this example to the above comment

artem-lunarg (approved):

layers/state_tracker/image_state.cpp
spencer-lunarg: I though `k` was for global constant, this is just any `const uint32_t` type we use inside a function

layers/state_tracker/image_state.h
spencer-lunarg: ... will bring up on meeting, we are pretty bad at this consistency

PR #8364, opened 2024-08-07
gpu: fix deferred pipeline creation

arno-lunarg:
Fix management of deferred pipeline creation
- Per the spec, when doing a deferred build, we have to maintain resources passed down to `vkCreateRayTracingPipelines` alive is creation is deferred, including pointers. This is done through maintaining `chassis_state` in a shared pointer that is captured in a lambda that is called upon getting the result of the deferred operation
- We had an issue in the notification code for when a deferred operation is completed: We also need to call the backlog of functions stored in `deferred_operation_post_completion` at `DispatchGetDeferredOperationResultKHR` time, to make sure our internal pipeline states are updated
- `chassis::CreateRayTracingPipelinesKHR` was also storing a vector of `safe_VkRayTracingPipelineCreateInfoCommon` instead of `safe_VkRayTracingPipelineCreateInfoKHR`, the former has an additional member that made `reinterpret_cast` its content to a `VkRayTracingPipelineCreateInfoKHR *` invalid

Issues showed up in Doom Eternal, they are gone now.

As of writing the new test I added is crashing on my AMD pc, can't find why, I need to run through CI to see if nvidia machines also have this problem

EDIT: Solved, do not pass a temporary handle to vkCreateRayTracingPipelines


layers/chassis/layer_chassis_dispatch_manual.cpp
```diff
@@ -1203,10 +1203,10 @@ VkResult DispatchDeferredOperationJoinKHR(VkDevice device, VkDeferredOperationKH
 
     // If this thread completed the operation, free any retained memory.
     if (result == VK_SUCCESS) {
-        auto iter = layer_data->deferred_operation_post_completion.pop(operation);
-        if (iter != layer_data->deferred_operation_post_completion.end()) {
-            for (auto &cleanup_fn : iter->second) {
-                cleanup_fn();
+        auto post_op__completion_fns = layer_data->deferred_operation_post_completion.pop(operation);
```
spencer-lunarg, line 1206:
```suggestion
        auto post_op_completion_fns = layer_data->deferred_operation_post_completion.pop(operation);
```

arno-lunarg: That was quick!

layers/gpu/spirv/inject_function_pass.cpp
```diff
@@ -31,7 +31,18 @@ bool InjectFunctionPass::Run() {
             auto& block_instructions = (*block_it)->instructions_;
             for (auto inst_it = block_instructions.begin(); inst_it != block_instructions.end(); ++inst_it) {
                 // Every instruction is analyzed by the specific pass and lets us know if we need to inject a function or not
-                if (!AnalyzeInstruction(*function, *(inst_it->get()))) continue;
+                if (!AnalyzeInstruction(*function, *(inst_it->get()))) {
```
spencer-lunarg, line 34: this needs to be a different PR, this was a sloppy hack and needs to be added in a more clean way

layers/state_tracker/state_tracker.cpp
```diff
@@ -2023,18 +2023,26 @@ bool ValidationStateTracker::PreCallValidateCreateRayTracingPipelinesKHR(
 void ValidationStateTracker::PostCallRecordCreateRayTracingPipelinesKHR(
     VkDevice device, VkDeferredOperationKHR deferredOperation, VkPipelineCache pipelineCache, uint32_t count,
     const VkRayTracingPipelineCreateInfoKHR *pCreateInfos, const VkAllocationCallbacks *pAllocator, VkPipeline *pPipelines,
-    const RecordObject &record_obj, PipelineStates &pipeline_states, chassis::CreateRayTracingPipelinesKHR &chassis_state) {
-    const bool operation_is_deferred = (deferredOperation != VK_NULL_HANDLE && record_obj.result == VK_OPERATION_DEFERRED_KHR);
+    const RecordObject &record_obj, PipelineStates &pipeline_states,
+    std::shared_ptr<chassis::CreateRayTracingPipelinesKHR> chassis_state) {
+    // #ARNO_TODO maybe need to copy
```
spencer-lunarg, line 2028: copy what?

arno-lunarg: ah leftover

layers/gpu/spirv/inject_function_pass.cpp
arno-lunarg: I need it to be able to run Doom :3

spencer-lunarg: well since currently for the short term, no one is using `inject_function_pass.cpp` and are using `inject_conditional_function_pass.cpp` so this isn't even in the correct spot

... I guess this is a "fix" but not in scope with fixing `deferred pipeline creation`

arno-lunarg: that's why it is in a different commit!

spencer-lunarg: sure, we can leave this but

1. I still think this is in the wrong file (should be in `inject_conditional_function_pass.cpp` as that is the path that caused this issue)
2. Then please add this comment

```
    // TODO - This should be cleaned up then having it injected here
    // 
    // we can have a situation where the incoming SPIR-V looks like
    //
    // %a = OpSampledImage %type %image %sampler
    // ... other stuff we inject a function around
    // %b = OpImageSampleExplicitLod %type2 %a %3893 Lod %3918
    //
    // and we get an error "All OpSampledImage instructions must be in the same block in which their Result <id> are consumed"
    // to get around this we inject a OpCopyObject right after the OpSampledImage
    ```

File: layers/gpu/spirv/inject_function_pass.cpp [hunk 4]
```diff
@@ -31,7 +31,26 @@ bool InjectFunctionPass::Run() {
...
```
spencer-lunarg, line 34: This is still in the wrong file, should be in `InjectConditionalFunctionPass::Run()`

arno-lunarg: Are you sure? I just applied the patch you gave me

spencer-lunarg: yes, that patch was before we made https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8357

spencer-lunarg (approved):

PR #8366, opened 2024-08-07
gpu: Fix buffer OOB to use range, not size

spencer-lunarg:
- We were using the "buffer size", not the "bound range", to detect buffer OOB in GPU-AV
- We didn't print the buffer handle in the GPU-AV error (yet we know it)
- Also for `VkDescriptorBufferInfo` we were not handling `VK_WHOLE_SIZE` correctly

image

### DIFF
--- a/layers/gpu/instrumentation/gpuav_instrumentation.cpp
+++ b/layers/gpu/instrumentation/gpuav_instrumentation.cpp
@@ -294,10 +294,13 @@ bool LogMessageInstBindlessDescriptor(Validator &gpuav, const uint32_t *error_re
                     error_found = true;
                     break;
                 }
-                case vvl::DescriptorClass::TexelBuffer:
+                case vvl::DescriptorClass::TexelBuffer: {
+                    const vvl::BufferView *buffer_view_state =
+                        static_cast<const vvl::TexelBinding *>(binding_state)->descriptors[desc_index].GetBufferViewState();
                     strm << "(set = " << set_num << ", binding = " << binding_num << ") Descriptor index " << desc_index
-                         << " access out of bounds. Descriptor size is " << size << " texels and highest texel accessed was "
-                         << offset;
+                         << " access out of bounds. The descriptor texel buffer ("
+                         << gpuav.FormatHandle(buffer_view_state->Handle()) << ") size is " << size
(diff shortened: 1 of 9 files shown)


layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -276,9 +276,15 @@ bool LogMessageInstBindlessDescriptor(const uint32_t *error_record, std::string
...
+                case vvl::DescriptorClass::GeneralBuffer: {
+                    const vvl::Buffer *buffer_state =
+                        static_cast<const vvl::BufferBinding *>(binding_state)->descriptors[desc_index].GetBufferState();
+                    assert(buffer_state);
+
                     strm << "(set = " << set_num << ", binding = " << binding_num << ") Descriptor index " << desc_index
-                         << " access out of bounds. Descriptor size is " << size << " and highest byte accessed was " << offset;
+                         << " access out of bounds. The descriptor buffer (" << gpuav.FormatHandle(buffer_state->Handle())
+                         << ") size is " << buffer_state->create_info.size << " bytes, " << size
+                         << " bytes were bound, and the highest out of bounds access was at [" << offset << "] bytes";
```
spencer-lunarg, line 287: newer error message

> vkCmdDraw(): (set = 0, binding = 0) Descriptor index 0 access out of bounds. The descriptor buffer (VkBuffer 0x301e6c0000000022[]) size is 8 bytes, 4 bytes were bound, and the highest out of bounds access was at [7] bytes

layers/state_tracker/descriptor_sets.cpp
```diff
@@ -1088,7 +1099,7 @@ void vvl::MutableDescriptor::WriteUpdate(DescriptorSet &set_state, const Validat
         case DescriptorClass::GeneralBuffer: {
             const auto &buffer_info = update.pBufferInfo[index];
             offset_ = buffer_info.offset;
-            range_ = buffer_info.range;
```
jeremyg-lunarg, line 1091: why did you stop setting `range_` here?

spencer-lunarg: because my `ctrl+z` buffer broke... good catch

jeremyg-lunarg (approved):

PR #8368, opened 2024-08-08
gpu: Add feature support checks for shader instrumentation

spencer-lunarg:
Create a setting struct that we can pass things down like required features support


layers/gpu/spirv/module.h
```diff
@@ -35,12 +34,20 @@ struct ModuleHeader {
     uint32_t schema;
 };
 
+struct Settings {
```
arno-lunarg, line 37: Missing default values

arno-lunarg (approved):

PR #8372, opened 2024-08-09
corechecks: Replace VUID 07825 with 08798

jeremyg-lunarg:
The latter is the same as the former but with added text about vkCmdBindVertexBuffer2(), which validation already handles.

Hopefully 07825 will be removed from the spec soon.


artem-lunarg (approved):

jeremyg-lunarg:
I filed https://gitlab.khronos.org/vulkan/vulkan/-/issues/3961 to ask for 07825 to be removed.

PR #8374, opened 2024-08-09

### LINKED ISSUE #2492: GPU-AV check index buffer out-of-bounds

@Tony-LunarG @greg-lunarg

jeremyg-lunarg:
This involves splitting up draw.vert into separate shaders for non-indexed indirect, indexed (both indirect and direct) and mesh draw calls.

Fixes #2492


jeremyg-lunarg:
This has many TODOs but works well enough I want a backup just in case my computer gets hit with an EMP

layers/gpu/cmd_validation/gpuav_draw.cpp
```diff
@@ -191,23 +195,79 @@ void DestroyRenderPassMappedResources(Validator &gpuav, VkRenderPass render_pass
...
-                                  VkDeviceSize indirect_offset, uint32_t draw_count, VkBuffer count_buffer,
-                                  VkDeviceSize count_buffer_offset, uint32_t stride) {
-    if (!gpuav.gpuav_settings.validate_indirect_draws_buffers) {
-        return;
+struct IndirectInfo {
+    IndirectInfo(Validator &gpuav, VkBuffer buffer_, VkDeviceSize offset_, size_t struct_size_, uint32_t stride_)
+        : buffer(buffer_), offset(offset_), stride(stride_) {
+        // Buffer size must be >= (stride * (drawCount - 1) + offset + sizeof(cmd))
+        auto buffer_state = gpuav.Get<vvl::Buffer>(buffer);
+        if (buffer_state) {
```
spencer-lunarg, line 203: do we need an assert here, the `buffer_size` not being set just seems to cascade strange values for the rest below?

spencer-lunarg: could we pass in a `const vvl::Buffer& buffer_state`?

layers/vulkan/generated/instrumentation_ray_query_comp.cpp
```diff
@@ -49,15 +49,15 @@
     0x69746341, 0x6e496e6f, 0x42786564, 0x65666675, 0x00000072, 0x00050006, 0x000000fb, 0x00000000, 0x65646e69, 0x00000078,
     0x00090005, 0x000000fd, 0x74736e69, 0x7463615f, 0x5f6e6f69, 0x65646e69, 0x75625f78, 0x72656666, 0x00000000, 0x00090047,
     0x0000000f, 0x00000029, 0x74736e69, 0x7961725f, 0x6575715f, 0x635f7972, 0x00706d6f, 0x00000000, 0x00040047, 0x000000aa,
-    0x00000006, 0x00000004, 0x00030047, 0x000000ab, 0x00000002, 0x00050048, 0x000000ab, 0x00000000, 0x00000023, 0x00000000,
```
spencer-lunarg, line 52: you have  the latest `glslang`  ... we just adjusted this, did it revert? Is it dependent on something else?

tests/unit/gpu_av_indirect_buffer.cpp
```diff
@@ -607,6 +632,247 @@ TEST_F(NegativeGpuAVIndirectBuffer, FirstInstance) {
     m_errorMonitor->VerifyFound();
 }
 
+TEST_F(NegativeGpuAVIndirectBuffer, IndexBufferOOB) {
```
spencer-lunarg, line 635: I doubt we have a positive GPU test that uses an index buffer, might be good to add one

tests/unit/gpu_av_indirect_buffer.cpp
```diff
@@ -607,6 +632,247 @@ TEST_F(NegativeGpuAVIndirectBuffer, FirstInstance) {
...
+    TEST_DESCRIPTION("Validate overruning the index buffer");
+    AddRequiredExtensions(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
+    RETURN_IF_SKIP(InitGpuAvFramework());
+
+    AddDisabledFeature(vkt::Feature::drawIndirectFirstInstance);
+    RETURN_IF_SKIP(InitState(nullptr));
```
spencer-lunarg, line 641: nit:

```suggestion
    RETURN_IF_SKIP(InitState());
```

layers/vulkan/generated/instrumentation_ray_query_comp.cpp
jeremyg-lunarg: I'll double check but I did move some constant definitions out of gpu_error_codes.h which is globally included, that might have cause the spirv to change for this shader.

layers/gpu/cmd_validation/gpuav_draw.cpp
jeremyg-lunarg: I suppose, no one else needs it but I guess it does make the error handling better

layers/gpu/cmd_validation/gpuav_draw.cpp
```diff
@@ -370,12 +364,189 @@ void InsertIndirectDrawValidation(Validator &gpuav, const Location &loc, Command
...
+attribAddress = bufferBindingAddress + effectiveVertexOffset + attribDesc.offset;
+
+end = size = offset[binding] + (index * stride) + attrib.offset + attrib.size;
+
+end = (size - offset[binding]) = (index * stride) + attrib.offset + attrib.size;
+
+max_index = (size - offset - attrib.offset - attrib.size) / stride
+*/
+#if 0
+static uint32_t MinVertexBufferLen(const vvl::CommandBuffer &cb_state, const vvl::Pipeline &pipeline) {
```
jeremyg-lunarg, line 434: This was my WIP attempt to figure out the max vertex buffer length

layers/gpu/cmd_validation/gpuav_draw.cpp
```diff
@@ -370,12 +364,189 @@ void InsertIndirectDrawValidation(Validator &gpuav, const Location &loc, Command
...
+                push_constants.index_width = 16;
+                break;
+            case VK_INDEX_TYPE_UINT8_KHR:
+                push_constants.index_width = 8;
+                break;
+            default:
+                gpuav.InternalError(cb_state.Handle(), loc, "Unsupported indexType");
+                return;
+        }
+        push_constants.vertex_buffer_size = 3;  // TODO vertex size calc
```
jeremyg-lunarg, line 513: This is the important TODO, actually send down the max vertex buffer size that will work across all bindings

tests/unit/gpu_av_positive.cpp
```diff
@@ -1732,3 +1732,53 @@ TEST_F(PositiveGpuAV, PipelineLayoutMixing) {
...
+        indexed_draw_ptr->vertexOffset = 0;
+        indexed_draw_ptr->firstInstance = 0;
+        indexed_draw_buffer.memory().unmap();
+    }
+
+    VkCommandBufferBeginInfo begin_info = vku::InitStructHelper();
+    m_commandBuffer->begin(&begin_info);
+    m_commandBuffer->BeginRenderPass(m_renderPassBeginInfo);
+    vk::CmdBindPipeline(m_commandBuffer->handle(), VK_PIPELINE_BIND_POINT_GRAPHICS, pipe.Handle());
+    const uint32_t kNumVertices = 12;
```
jeremyg-lunarg, line 1769: This test makes sure the TODO gets fixed.

jeremyg-lunarg:
@arno-lunarg you may want to take a look at the latest push. I fixed up @spencer-lunarg's previous review comments

layers/gpu/cmd_validation/gpuav_draw.cpp
arno-lunarg: If I am not mistaken, multiple vertex buffers can be bound, each being an input to a different `location` in the shader, so we have to account for that?

arno-lunarg: Ah but it seems your WIP is taking care of that!

layers/gpu/shaders/cmd_validation/indexed_draw.vert
```diff
@@ -0,0 +1,154 @@
...
+        // host should have caught this
+        return false;
+    }
+
+    if ((fi + ic) > index_len) {
+        GpuavLogError(kErrorGroupGpuPreDraw, kErrorSubCodePreDrawIndexBuffer, fi, ic);
+        return false;
+    }
+
+    for (uint j = 0; j < ic; j++) {
```
arno-lunarg, line 95: I think having one GPU thread check all indices will make things really slow, I am thinking about moving the index check to a dedicated validation vertex shader, and dispatch a draw on it with a number of vertices specified on the CPU for non indirect draws, and computed on the GPU for indirect ones

layers/gpu/cmd_validation/gpuav_draw.cpp
```diff
@@ -38,13 +41,14 @@ struct SharedDrawValidationResources final {
     VkDevice device = VK_NULL_HANDLE;
 
     SharedDrawValidationResources(Validator &gpuav, VkDescriptorSetLayout error_output_desc_set_layout, bool use_shader_objects,
-                                  const Location &loc)
+                                  const Location &loc, const uint32_t *shader, uint32_t shader_size)
```
arno-lunarg, line 44: I should have added comments, but this will not work as expected x_x

Only one `SharedDrawValidationResources` object is created, it is then cached and retrieved when needed. So the first call to `gpuav.shared_resources_manager.Get<SharedDrawValidationResources>(
       /*...*/, shader, shader_size);` will set the validation shader, and calling it again with a different shader will be a no-op and just retrieve the cached entry......

layers/gpu/shaders/cmd_validation/indirect_draw.vert
```diff
@@ -0,0 +1,80 @@
...
+    uint count_buffer;
+};
+
+// binding 2 is for the index buffer which isn't used
+
+void main() {
+    if (gl_VertexIndex == 0) {
+        uint draw_count = 0;
+
+        if ((push_data.flags & kPreDrawSelectCountBuffer) != 0) {
```
arno-lunarg, line 53: Validation code is now duplicated, I don't think this is ideal?
I really like how you broke up the logic in `gpuav_draw.cpp`, but here, the old Uber shader was not that bad I think

layers/gpu/cmd_validation/gpuav_draw.cpp
arno-lunarg: The key in the cache is only based on the type of the stored object, ideally I would also use the passed in parameters to form the key, but I don't even know if that is possible

layers/gpu/cmd_validation/gpuav_draw.cpp
arno-lunarg: Just thought about the case where there is an index buffer but no vertex buffer. I think given how you set min_vertex_count to UINT32_MAX, it is properly handled? Ideally in this case we should not even bother doing validation

layers/gpu/shaders/cmd_validation/indirect_draw.vert
jeremyg-lunarg: The answer to this will depend on how much additional validation needs to happen in each draw type.  Also how many push constants become necessary and when that starts to exceeds device limits. It seems to me that at least mesh shaders would benefit from being separate.

layers/gpu/cmd_validation/gpuav_draw.cpp
jeremyg-lunarg: Yes but the WIP is handling VK_VERTEX_INPUT_RATE_INSTANCE and the divisor.  Those could be split out to a separate value. I think these can be checked on the CPU for direct calls but not for indirect.

layers/gpu/cmd_validation/gpuav_draw.cpp
jeremyg-lunarg: Then you can just turn off kPreDrawSelectIndexBuffer for indirect and skip the diagnostic draw for vkCmdDrawIndexed()

layers/gpu/shaders/cmd_validation/indexed_draw.vert
jeremyg-lunarg: Good point.   You could just map the index buffer as a vertex buffer and have a separate thread check each index. And have the other checking only happen in thread 0. Or are you thinking about having 2 shaders?

layers/gpu/shaders/cmd_validation/indirect_draw.vert
jeremyg-lunarg: The other reason I split up the shaders was that there are fields with the same name but different offsets in VkDrawIndirectCommand and VkDrawIndexedIndirectCommand. And then VkDrawMeshTasksIndirectCommandEXT is completely different. The old shader code got down the offset of the firstInstance member in a push constant for whatever draw was being validated. But... I wanted to access multiple fields and not deal with the mess of managing different offsets in the same shader.

layers/gpu/cmd_validation/gpuav_draw.cpp
jeremyg-lunarg: It doesn't look like it is possible as written.  If it was up to me, I'd ditch the typeid magic and just make a key enum.

PR #8383, opened 2024-08-12
core: Use robin hood map to improve perf

arno-lunarg:
Turns out adding/updating elements to this map is an hot spot for Doom Eternal. Did not improve game frame rate when actually playing the game, at least in the menu, but with a GFXR trace, the difference is drastic: Saved 19 seconds on a replay that is less than 2 minutes long 

image

### DIFF
--- a/layers/state_tracker/query_state.h
+++ b/layers/state_tracker/query_state.h
@@ -154,7 +154,7 @@ inline bool operator==(const QueryObject &query1, const QueryObject &query2) {
     return ((query1.pool == query2.pool) && (query1.slot == query2.slot) && (query1.perf_pass == query2.perf_pass));
 }
 
-typedef std::map<QueryObject, QueryState> QueryMap;
+using QueryMap = vvl::unordered_map<QueryObject, QueryState>;
 
 enum QueryResultType {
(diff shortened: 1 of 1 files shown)


spencer-lunarg (approved):
always happy to remove un-needed `std::map` like this

layers/state_tracker/query_state.h
```diff
@@ -154,7 +154,7 @@ inline bool operator==(const QueryObject &query1, const QueryObject &query2) {
     return ((query1.pool == query2.pool) && (query1.slot == query2.slot) && (query1.perf_pass == query2.perf_pass));
 }
 
-typedef std::map<QueryObject, QueryState> QueryMap;
```
artem-lunarg, line 157: Any usages of `QueryMap` that rely on ordered property of std::map?
If no, then `bool operator<(const QueryObject &rhs) const` can be removed.

arno-lunarg: ah good question, need to investigate

artem-lunarg: yeah, I briefly scanned through the code and could not see such places, but definitely not a thorough check.

PR #8387, opened 2024-08-12
gpu: Combine ShaderObject in Printf/GPU-AV

spencer-lunarg:
Combine the `PreCallRecordCreateShadersEXT` for GPU-AV and DebugPrintf as they use the same flow now

Also make it look like the Pipeline path for code consistency

### DIFF
--- a/layers/chassis/chassis_modification_state.h
+++ b/layers/chassis/chassis_modification_state.h
@@ -42,6 +42,13 @@ struct CreateShaderModule {
     spirv::StatelessData stateless_data;
 };
 
+struct ShaderObjectInstrumentationMetadata {
+    // Need to hold this in memory between the PreCallRecord and the Dispatch
+    std::vector<uint32_t> instrumented_spirv;
+    // Need to hold this in memory between the PreCallRecord and the PostCallRecord
+    uint32_t unique_shader_ids;
+};
+
 // VkShaderEXT (VK_EXT_shader_object)
 struct ShaderObject {
     // allows PreCallRecord to return a value like PreCallValidate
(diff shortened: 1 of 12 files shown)


layers/chassis/chassis_modification_state.h
```diff
@@ -42,27 +42,33 @@ struct CreateShaderModule {
     spirv::StatelessData stateless_data;
 };
 
+struct ShaderObjectInstrumentationMetadata {
+    // Need to hold this in memory between the PreCallRecord and the Dispatch
+    std::vector<uint32_t> instrumented_spirv;
+    // Need to hold this in memory between the PreCallRecord and the PostCallRecord
+    uint32_t unique_shader_ids;
```
arno-lunarg, line 49:
```suggestion
    uint32_t unique_shader_id;
```

layers/chassis/chassis_modification_state.h
```diff
@@ -42,27 +42,33 @@ struct CreateShaderModule {
     spirv::StatelessData stateless_data;
 };
 
+struct ShaderObjectInstrumentationMetadata {
```
arno-lunarg, line 45:
```suggestion
struct ShaderObjectInstrumentationData {
```
Since this holds `instrumented_spirv`, it does not only hold _Meta_ data, but really instrumentation data. Yesterday it was truly "Metadata", here not so much : 3

layers/chassis/chassis_modification_state.h
```diff
@@ -42,27 +42,33 @@ struct CreateShaderModule {
...
 
-    // Pass the instrumented SPIR-V info from PreCallRecord to Dispatch (so GPU-AV logic can run with it)
-    VkShaderCreateInfoEXT* instrumented_create_info = nullptr;
-    std::vector<std::vector<uint32_t>> instrumented_spirv;
+    std::vector<VkShaderCreateInfoEXT> modified_create_infos;
+    const VkShaderCreateInfoEXT* pCreateInfos = nullptr;
 
-    std::vector<VkDescriptorSetLayout> new_layouts;
+    // Pass the instrumented SPIR-V info from PreCallRecord to Dispatch (so GPU-AV logic can run with it)
+    std::vector<ShaderObjectInstrumentationMetadata> shader_instrumentations_metadata;
```
arno-lunarg, line 66: instrumentation_data
```
    std::vector<ShaderObjectInstrumentationData> instrumentations_data;
```

layers/gpu/instrumentation/gpu_shader_instrumentor.cpp
```diff
@@ -492,15 +481,59 @@ void GpuShaderInstrumentor::PostCallRecordCreateShaderModule(VkDevice device, co
...
 
+void GpuShaderInstrumentor::PreCallRecordShaderObjectInstrumentation(
+    VkShaderCreateInfoEXT &create_info, const Location &create_info_loc,
+    chassis::ShaderObjectInstrumentationMetadata &shader_instrumentation_metadata) {
+    if (gpuav_settings.select_instrumented_shaders && !IsSelectiveInstrumentationEnabled(create_info.pNext)) return;
+    uint32_t unique_shader_id = 0;
+    bool cached = false;
+    bool pass = false;
+    std::vector<uint32_t> &instrumented_spirv = shader_instrumentation_metadata.instrumented_spirv;
+    if (gpuav_settings.cache_instrumented_shaders) {
```
arno-lunarg, line 492: The whole cache logic needs to be centralized I think, if this is indeed another duplicate?

arno-lunarg (approved):

layers/gpu/instrumentation/gpu_shader_instrumentor.cpp
spencer-lunarg: yes, my doing this I realize it could be more centralized, will take a stab at it after this

PR #8381, opened 2024-08-12
gpu: Move Instrumentation to Pipeline creation

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7804

This move is done because the next change to improve descriptor indexing checks will requires use to know if `VK_DESCRIPTOR_BINDING_PARTIALLY_BOUND_BIT_EXT` is used or not

Luckily because of `VK_KHR_maintenance5`

image

we already had a path to do stuff at pipeline creation time

This change actually removed a lot of ugly tracking we did to match the `vkCreateShaderModule` at create pipeline time, having it all in a single chassis call allows us to save more data on the stack and remove it when the pipeline is created

PR #8389, opened 2024-08-13
gpu: Remove extra copy of safe structs

spencer-lunarg:
cleanup from https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8381#discussion_r1714718004

### DIFF
--- a/layers/chassis/chassis_modification_state.h
+++ b/layers/chassis/chassis_modification_state.h
@@ -67,8 +67,10 @@ struct ShaderObject {
 };
 
 struct ShaderInstrumentationMetadata {
-    // Maps the SPIR-V to a specific VkShaderModule/VkPipeline/VkShaderObject/etc
+    // We need one for each stage (Vertex, Fragment, etc) and the index is VkGraphicsPipelineCreateInfo::pStages[i]
+    // Maps the SPIR-V to a specific VkShaderModule/VkPipeline/etc handle
     std::vector<uint32_t> spirv_unique_id_map;
+
     // Used to know if VkShaderModuleCreateInfo is passed down VkPipelineShaderStageCreateInfo
     bool passed_in_shader_stage_ci = false;
 };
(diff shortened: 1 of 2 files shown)


layers/gpu/instrumentation/gpu_shader_instrumentor.cpp
```diff
@@ -949,7 +959,11 @@ void GpuShaderInstrumentor::PreCallRecordPipelineCreationShaderInstrumentation(
             } else if (sm_ci) {
                 // The user is inlining the Shader Module into the pipeline, so just need to update the spirv
                 shader_instrumentation_metadata.passed_in_shader_stage_ci = true;
-                // TODO - This makes a copy, but could save on Chassis stack instead (then remove function from VUL)
+                // TODO - This makes a copy, but could save on Chassis stack instead (then remove function from VUL).
+                // The core issue is we always use std::vector<uint32_t> but Safe Struct manages its own version of the pCode
+                // memory. It would be much harder to change everything from std::vector and instead to adjust Safe Struct to not
+                // double-free the memory on us. If making any changes, we have to consider a case where the user inlines the
+                // fragment shader, but use a normal VkShaderModule in the vertex shader.
                 sm_ci->SetCode(instrumented_spirv);
```
spencer-lunarg, line 967: we could actually replace `SetCode` to not copy, but set a "Don't own" and consume the pointer, and not destroy it... probably the simplest way to remove this copy since there is already a side-function for us to modify

layers/chassis/chassis_modification_state.h
```diff
@@ -67,8 +67,10 @@ struct ShaderObject {
 };
 
 struct ShaderInstrumentationMetadata {
-    // Maps the SPIR-V to a specific VkShaderModule/VkPipeline/VkShaderObject/etc
+    // We need one for each stage (Vertex, Fragment, etc) and the index is VkGraphicsPipelineCreateInfo::pStages[i]
+    // Maps the SPIR-V to a specific VkShaderModule/VkPipeline/etc handle
```
arno-lunarg, line 71:
```suggestion
    // Holds a unique shader id for each VkGraphicsPipelineCreateInfo::pStages[i]
    // Unique shader ids are used to map SPIR-V to a specific VkShaderModule/VkPipeline/etc handle
```

arno-lunarg (approved):

PR #8390, opened 2024-08-13

### LINKED ISSUE #2492: GPU-AV check index buffer out-of-bounds

arno-lunarg:
This involves splitting up draw.vert into separate shaders for non-indexed indirect, indexed (both indirect and direct) and mesh draw calls.

Fixes #2492

Takeover of https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8374


jeremyg-lunarg:
I think I may have a slightly better version of this code to push still, I will push to my PR and you can figure out if that's really the case.

arno-lunarg:
closing for now, still WIP and no need to run through CI

PR #8391, opened 2024-08-13
state: Improve storage of vertex input state

### LINKED ISSUE #5281: syncval: Add support for cmdBindVertexBuffers2 to syncval
Bind Vertex Buffers2 allows for restricting the scope of the buffer bound which provides a workaround for false positives where the whole of the buffer is considered used.

jeremyg-lunarg:
Rather than just storing arrays of VkVertexInputBindingDescription and VkVertexAttributeBindingDescription structures that have to be corellated to do anything useful, group all of this state by binding and location for easier lookup.

Fixes #5281

### DIFF
--- a/layers/core_checks/cc_cmd_buffer_dynamic.cpp
+++ b/layers/core_checks/cc_cmd_buffer_dynamic.cpp
@@ -1013,52 +1013,54 @@ bool CoreChecks::ValidateDrawDynamicState(const LastBound& last_bound_state, con
                 continue;
             }
             bool location_provided = false;
-            for (uint32_t i = 0; i < cb_state.dynamic_state_value.vertex_attribute_descriptions.size(); ++i) {
-                const auto& description = cb_state.dynamic_state_value.vertex_attribute_descriptions[i];
-                if (variable_ptr->decorations.location == description.location) {
+            if (const auto* binding_state =
+                    vvl::Find(cb_state.dynamic_state_value.vertex_bindings, variable_ptr->decorations.binding)) {
+                if (const auto* attrib = vvl::Find(binding_state->locations, variable_ptr->decorations.location)) {
                     location_provided = true;
 
                     const uint32_t var_base_type_id = variable_ptr->base_type.ResultId();
-                    const uint32_t attribute_type = spirv::GetFormatType(description.format);
+                    const uint32_t attribute_type = spirv::GetFormatType(attrib->desc.format);
                     const uint32_t var_numeric_type = vert_spirv_state->GetNumericType(var_base_type_id);
 
-                    const bool attribute64 = vkuFormatIs64bit(description.format);
+                    const bool attribute64 = vkuFormatIs64bit(attrib->desc.format);
                     const bool shader64 = vert_spirv_state->GetBaseTypeInstruction(var_base_type_id)->GetBitWidth() == 64;
 
                     // first type check before doing 64-bit matching
                     if ((attribute_type & var_numeric_type) == 0) {
                         if (!enabled_features.legacyVertexAttributes || shader64) {
-                            skip |=
-                                LogError(vuid.vertex_input_08734, vert_spirv_state->handle(), vuid.loc(),
-                                         "vkCmdSetVertexInputEXT set pVertexAttributeDescriptions[%" PRIu32 "].location (%" PRIu32
-                                         ") with format %s but the vertex shader input is numeric type %s",
-                                         i, description.location, string_VkFormat(description.format),
-                                         vert_spirv_state->DescribeType(var_base_type_id).c_str());
+                            skip |= LogError(
+                                vuid.vertex_input_08734, vert_spirv_state->handle(), vuid.loc(),
+                                "vkCmdSetVertexInputEXT set pVertexAttributeDescriptions[%" PRIu32 "] (binding %" PRIu32
+                                ", location %" PRIu32 ") with format %s but the vertex shader input is numeric type %s",
+                                attrib->index, attrib->desc.binding, attrib->desc.location, string_VkFormat(attrib->desc.format),
+                                vert_spirv_state->DescribeType(var_base_type_id).c_str());
                         }
                     } else if (attribute64 && !shader64) {
-                        skip |= LogError(vuid.vertex_input_format_08936, vert_spirv_state->handle(), vuid.loc(),
-                                         "vkCmdSetVertexInputEXT set pVertexAttributeDescriptions[%" PRIu32 "].location (%" PRIu32
-                                         ") with a 64-bit format (%s) but the vertex shader input is 32-bit type (%s)",
-                                         i, description.location, string_VkFormat(description.format),
-                                         vert_spirv_state->DescribeType(var_base_type_id).c_str());
+                        skip |= LogError(
+                            vuid.vertex_input_format_08936, vert_spirv_state->handle(), vuid.loc(),
+                            "vkCmdSetVertexInputEXT set pVertexAttributeDescriptions[%" PRIu32 "] (binding %" PRIu32
+                            ", location %" PRIu32 ") with a 64-bit format (%s) but the vertex shader input is 32-bit type (%s)",
+                            attrib->index, attrib->desc.binding, attrib->desc.location, string_VkFormat(attrib->desc.format),
+                            vert_spirv_state->DescribeType(var_base_type_id).c_str());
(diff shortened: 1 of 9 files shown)


layers/state_tracker/pipeline_sub_state.h
```diff
@@ -52,17 +52,18 @@ struct PipelineSubState {
     VkPipelineLayoutCreateFlags PipelineLayoutCreateFlags() const;
 };
 
+struct VertexBindingState {
+    vku::safe_VkVertexInputBindingDescription2EXT binding;
+    std::map<uint32_t, vku::safe_VkVertexInputAttributeDescription2EXT> locations;
```
spencer-lunarg, line 57: can this not be `vvl::unordered_map`? I didn't spot anywhere the ordered mattered here (maybe I just forgot, but maybe a comment would here would help)

(same right below)

layers/sync/sync_commandbuffer.cpp
```diff
@@ -574,16 +574,16 @@ bool CommandBufferAccessContext::ValidateDrawVertex(const std::optional<uint32_t
         return skip;
     }
 
-    // TODO - doesn't consider dynamic vertex binding input
```
spencer-lunarg, line 577: this was such a nasty ticking time bomb for sync val to hit

layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -4103,71 +4105,74 @@ bool CoreChecks::ValidateDrawPipelineVertexAttribute(const vvl::CommandBuffer &c
...
             break;
-        } else if (vertex_binding_map_it->second.buffer == VK_NULL_HANDLE && !enabled_features.nullDescriptor) {
+        }
+        const auto &vbb = vertex_binding_map_it->second;
+        if (vbb.buffer == VK_NULL_HANDLE && !enabled_features.nullDescriptor) {
             const LogObjectList objlist(cb_state.Handle(), pipeline.Handle());
             skip |= LogError(vuid.vertex_binding_attribute_02721, objlist, vuid.loc(),
-                             "pVertexAttributeDescriptions[%" PRIu32 "].binding (%" PRIu32 ") points to a VK_NULL_HANDLE buffer.",
-                             vertex_binding, i);
+                             "pVertexAttributeDescriptions (binding %" PRIu32 ") points to a VK_NULL_HANDLE buffer.", binding_num);
```
spencer-lunarg, line 4127: Why did we lose the information about which `pVertexAttributeDescriptions` index it was from?

spencer-lunarg (approved):
some nits/concerns, but if addresses LGTM

layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -1013,9 +1013,11 @@ bool CoreChecks::ValidateDrawDynamicState(const LastBound& last_bound_state, con
                 continue;
             }
             bool location_provided = false;
-            for (uint32_t i = 0; i < cb_state.dynamic_state_value.vertex_attribute_descriptions.size(); ++i) {
-                const auto& description = cb_state.dynamic_state_value.vertex_attribute_descriptions[i];
-                if (variable_ptr->decorations.location == description.location) {
+            auto binding_state = cb_state.dynamic_state_value.vertex_bindings.find(variable_ptr->decorations.binding);
+            if (binding_state != cb_state.dynamic_state_value.vertex_bindings.end()) {
```
artem-lunarg, line 1017: It should be possible to use `vvl::Find` by sacrificing  usual stl style. The benefits are that there's no need to repeat the variable name both in find() and end(), reduces scope of variable, avoids generic `second` naming. I started to use it in my code but I also don't like much stl style, so can be biased. Something like this:
```
if (const auto* binding_state = vvl::Find(cb_state.dynamic_state_value.vertex_bindings, variable_ptr->decorations.binding) {
      if (const auto* attrib_state = vvl::Find(binding_state->locations, variable_ptr->decorations.location) {
                          ....
}
```

artem-lunarg (approved):

layers/core_checks/cc_pipeline_graphics.cpp
jeremyg-lunarg: The old code stored the arrays passed by the caller to vkCmdSetVertexInput() or in the vertex input create info, so we had the index simply because all of the data was in the original order.   I could store them in the new VertexBindingState struct if you think it is worth it.

layers/core_checks/cc_cmd_buffer_dynamic.cpp
jeremyg-lunarg: TIL about vvl::Find()

layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -4103,71 +4105,78 @@ bool CoreChecks::ValidateDrawPipelineVertexAttribute(const vvl::CommandBuffer &c
...
+        const auto binding_num = binding_desc.binding;
 
-        const auto &vertex_binding_map_it = cb_state.current_vertex_buffer_binding_info.find(vertex_binding);
-        if (vertex_binding_map_it == cb_state.current_vertex_buffer_binding_info.cend()) {
+        const auto *vertex_buffer = vvl::Find(cb_state.current_vertex_buffer_binding_info, binding_num);
+        if (!vertex_buffer) {
             const LogObjectList objlist(cb_state.Handle(), pipeline.Handle());
-            skip |=
-                LogError(vuid.vertex_binding_attribute_02721, objlist, vuid.loc(),
-                         "pVertexAttributeDescriptions[%" PRIu32 "].binding (%" PRIu32 ") is an invalid value.", vertex_binding, i);
```
jeremyg-lunarg, line 4119: Note that the binding number and index into pVertexAttributeDescriptions were swapped throughout this method.

layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -4103,71 +4105,78 @@ bool CoreChecks::ValidateDrawPipelineVertexAttribute(const vvl::CommandBuffer &c
...
-                                 vertex_binding, vertex_buffer_stride, String(vuid.function), i, attribute_binding_extent);
+            const VkDeviceSize attribute_offset = attr_desc.offset;
+            const VkFormat attribute_format = attr_desc.format;
+            const VkDeviceSize vertex_buffer_stride = vertex_buffer->stride;
+            if (pipeline.IsDynamic(CB_DYNAMIC_STATE_VERTEX_INPUT_BINDING_STRIDE)) {
+                const VkDeviceSize attribute_binding_extent = attribute_offset + vkuFormatElementSize(attribute_format);
+                if (vertex_buffer_stride != 0 && vertex_buffer_stride < attribute_binding_extent) {
+                    const LogObjectList objlist(cb_state.Handle(), pipeline.Handle());
+                    skip |= LogError("VUID-vkCmdBindVertexBuffers2-pStrides-06209", objlist, vuid.loc(),
+                                     "(binding %" PRIu32 "location %" PRIu32 ") The pStrides value (%" PRIu64
```
jeremyg-lunarg, line 4152: here too

layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -4103,71 +4105,78 @@ bool CoreChecks::ValidateDrawPipelineVertexAttribute(const vvl::CommandBuffer &c
...
 
-        const VkDeviceSize attribute_offset = attribute_description.offset;
-        const VkFormat attribute_format = attribute_description.format;
-        const VkDeviceSize vertex_buffer_stride = vertex_binding_map_it->second.stride;
-        if (pipeline.IsDynamic(CB_DYNAMIC_STATE_VERTEX_INPUT_BINDING_STRIDE)) {
-            const VkDeviceSize attribute_binding_extent = attribute_offset + vkuFormatElementSize(attribute_format);
-            if (vertex_buffer_stride != 0 && vertex_buffer_stride < attribute_binding_extent) {
-                const LogObjectList objlist(cb_state.Handle(), pipeline.Handle());
-                skip |= LogError("VUID-vkCmdBindVertexBuffers2-pStrides-06209", objlist, vuid.loc(),
-                                 "The pStrides[%" PRIu32 "] (%" PRIu64
```
jeremyg-lunarg, line 4145: pStrides came from the call to vkCmdSetVertexBuffers2(). But we were using the binding number as the index here too.

Issue #8319, opened 2024-07-25
VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500


RikiLee:
**Environment:**
 - OS: Windows 11
 - GPU and driver version: NVIDIA RTX4090
 - SDK or header version if building from the repo: 1.3.290.0
 - Options enabled (synchronization, best practices, etc.):

**Describe the Issue**

Unexcepted error message when I use a pipeline without the dynamic state VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY, I only enable the VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY for another pipeline.

**Expected behavior**

No VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500 errors.

**Valid Usage ID**
VUID-vkCmdDraw-dynamicPrimitiveTopologyUnrestricted-07500

spencer-lunarg:
@RikiLee can you please provide the error message that was given

spencer-lunarg:
an [API Dump](https://vulkan.lunarg.com/doc/view/latest/windows/api_dump_layer.html) or GFXReconstruct capture would also help... we have tests for this VU and I see the ` if (pipeline.IsDynamic(CB_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY)` around it, so not sure how you go in this VU reported by accident

RikiLee:
> @RikiLee can you please provide the error message that was given

[...]

PR #8394, opened 2024-08-15
layers: Fix false positive for dynamic Geometry Topology

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8319

For geometry shaders, which have an input and output topology, we were grabbing the last one listed and not the dedicated input one

### DIFF
--- a/layers/state_tracker/shader_module.h
+++ b/layers/state_tracker/shader_module.h
@@ -710,6 +710,7 @@ struct Module {
     std::string DescribeType(uint32_t type) const;
     std::string DescribeVariable(uint32_t id) const;
 
+    // Note that some shaders can have an input and output topology
     std::optional<VkPrimitiveTopology> GetTopology(const EntryPoint &entrypoint) const;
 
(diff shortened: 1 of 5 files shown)


layers/state_tracker/shader_module.h
```diff
@@ -710,7 +710,9 @@ struct Module {
     std::string DescribeType(uint32_t type) const;
     std::string DescribeVariable(uint32_t id) const;
 
+    // Geometry shaders can have an input and output topology
     std::optional<VkPrimitiveTopology> GetTopology(const EntryPoint &entrypoint) const;
```
arno-lunarg, line 714: Should this be called `GetOutputTopology` now?

spencer-lunarg: <enter the rabbit hole>

so for Vertex shaders, there is just "Topology" (which is implicitly input)

for Geom there is a dedicated Input/Output variation of the topology

so no, I wish it was that simple

Issue #8375, opened 2024-08-11
Timeline signaling validation breaks after win32 handle export.


rygo6:
**Environment:**
 - OS: Win11
 - GPU and driver version: 4090
 - SDK or header version if building from repo: 1.3.290.0
 - Options enabled (synchronization, best practices, etc.): Fine Grained Locking. Core.

**Describe the Issue**

Signaling a semaphore with:
```
  const VkSemaphoreSignalInfo semaphoreSignalInfo = {
      .sType = VK_STRUCTURE_TYPE_SEMAPHORE_SIGNAL_INFO,
      .semaphore = timeline,
      .value = signalValue,
  };
  MIDVK_REQUIRE(vkSignalSemaphore(device, &semaphoreSignalInfo));
```
Gives the following error:
```
MemorVUID-VkSemaphoreSignalInfo-value-03259(ERROR / SPEC): msgNum: 1227785937 - Validation Error: [ VUID-VkSemaphore
SignalInfo-value-03259 ] Object 0: handle = 0x28ff7d9a370, name = CompTimelineSemaphore, type = VK_OBJECT_TYPE_SEMAPHORE
; | MessageID = 0x492e86d1 | vkSignalSemaphore(): pSignalInfo->value (1) must be less than value of any pending signal o
peration (0) for semaphore VkSemaphore 0x28ff7d9a370[CompTimelineSemaphore]. The Vulkan spec states: value must be less
than the value of any pending semaphore signal operations (https://vulkan.lunarg.com/doc/view/1.3.290.0/windows/1.3-exte
nsions/vkspec.html#VUID-VkSemaphoreSignalInfo-value-03259)
    Objects: 1
        [0] 0x28ff7d9a370, type: 5, name: CompTimelineSemaphore
```
However, it ONLY has this issue if I create the semaphore as an exported win32 handle and then actually `Get` the handle via:
```
      const VkSemaphoreGetWin32HandleInfoKHR getWin32HandleInfo = {
          .sType = VK_STRUCTURE_TYPE_SEMAPHORE_GET_WIN32_HANDLE_INFO_KHR,
          .semaphore = *pSemaphore,
          .handleType = MIDVK_EXTERNAL_SEMAPHORE_HANDLE_TYPE,
      };
      MIDVK_INSTANCE_FUNC(GetSemaphoreWin32HandleKHR);
[...]

Issue #8212, opened 2024-06-28
Inconsistent behavior when signaling exported timeline semaphore (VUID-VkSemaphoreSignalInfo-value-03259)


jerry411:
**Environment:**
 - OS: Windows 11
 - GPU and driver version: Nvidia RTX 3070Ti; 551.86
 - SDK or header version if building from repo: SDK version 1.3.283.0
 - Options enabled (synchronization, best practices, etc.): Validation configuration with Standard preset 

**Describe the Issue**

I have a really weird issue with VK_KHR_external_semaphore. All I want to do is a simple test. As a setup, I:
- create an exportable timeline semaphore on one VkInstance
- export the semaphore's handle (vkGetSemaphoreWin32HandleKHR with VK_EXTERNAL_FENCE_HANDLE_TYPE_OPAQUE_WIN32_BIT), - and import it on another instance.

Then I want to:
- query both semaphore's counter values (vkGetSemaphoreCounterValue)
- signal the first semaphore (with greater value),
- query both their counter values again and see whether those values match (as I expect they should since the second semaphore references the payload of the first one).

However, when signaling a semaphore, I get a validation layer error: "VUID-VkSemaphoreSignalInfo-value-03259" - "value must be less than the value of any pending semaphore signal operations". But there are not any pending signal operations! All operations I do in this simple test are mentioned above.

But wait, it gets weirder! When I do NOT call the first "vkGetSemaphoreCounterValue" (the one before signaling), no validation layer error appears! And if I DO call vkGetSemaphoreCounterValue but don't get a handle to the first semaphore, it is fine as well. Only a combination of getting handle and getting counter value triggers this validation layer error. As a cherry on top, if I query for counter value BEFORE creating a handle, it is all fine. At this point, I think this is a bug in the validation layers.

To summarize:
* create semaphore -> get a handle -> get counter value -> signal - throws aforementioned validation layer error
* create semaphore -> get counter value -> get handle -> signal - is fine
* create semaphore -> get handle -> signal - is fine
* create semaphore -> get counter value -> signal - is fine

**Expected behavior**

I think I should not get this validation layer error at all. And even if it is correct (I would love to know the reason because I really do not understand it at this point), it should be consistent. Changing the order of "get a handle" and "get counter value" operations should not affect a validation layer error being thrown when signaling a semaphore.

**Valid Usage ID**

[...]

PR #8396, opened 2024-08-15
sync: Fix vkGetSemaphoreCounterValue leaves orphaned signal

artem-lunarg:
Addresses  https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8212 and potentially https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8375.

### DIFF
--- a/layers/state_tracker/semaphore_state.cpp
+++ b/layers/state_tracker/semaphore_state.cpp
@@ -286,18 +286,15 @@ void vvl::Semaphore::NotifyAndWait(const Location &loc, uint64_t payload) {
                                " completed_.payload=%" PRIu64 " wait_payload=%" PRIu64,
                                completed_.payload, payload);
         }
-    } else {
+    } else {  // TODO: ensure atomicity of this entire else branch
         // For external timeline semaphores we should bump the completed payload to whatever the driver
-        // tells us. That value may originate from an external process that imported the semaphore and
-        // might not be signaled through the application queues.
-        //
-        // However, there is one exception. The current process can still signal the semaphore, even if
-        // it was imported. The queue's semaphore signal should not be overwritten by a potentially
-        // external signal. Otherwise, queue information (queue/seq) can be lost, which may prevent the
-        // advancement of the queue simulation.
-        const auto it = timeline_.find(payload);
-        const bool already_signaled = it != timeline_.end() && it->second.signal_submit.has_value();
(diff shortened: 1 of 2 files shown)


layers/state_tracker/semaphore_state.cpp
```diff
@@ -286,18 +286,15 @@ void vvl::Semaphore::NotifyAndWait(const Location &loc, uint64_t payload) {
                                " completed_.payload=%" PRIu64 " wait_payload=%" PRIu64,
                                completed_.payload, payload);
         }
-    } else {
+    } else {  // TODO: ensure atomicity of this entire else branch
```
artem-lunarg, line 289: It should be possible to have race conditions in this else block, but that's something we had before and not related to current issue. So  making progress by adding TODO!

layers/state_tracker/semaphore_state.cpp
```diff
@@ -286,18 +286,15 @@ void vvl::Semaphore::NotifyAndWait(const Location &loc, uint64_t payload) {
...
-        const auto it = timeline_.find(payload);
-        const bool already_signaled = it != timeline_.end() && it->second.signal_submit.has_value();
-        if (!already_signaled) {
+        // tells us. That value may originate from an external process and is not trackable by vvl.
+        // There is one exception, though. If the current program signaled the imported semaphore,
+        // then this signal should not be overwritten by a potential external signal. Otherwise, signal's
+        // queue information (queue/seq) can be lost and this may block queue forward progress simulation.
+        const TimePoint *timepoint = vvl::Find(timeline_, payload);
+        const bool already_signaled = timepoint && timepoint->signal_submit.has_value();
+        if (!already_signaled && CurrentPayload() < payload) {
```
artem-lunarg, line 297: The only thing that is changed here (the rest is comment update and minor cleanup) is additional condition `CurrentPayload() < payload`. 

`Semaphore::Retire` skips signals that are a not larger than current completed value (early exit):
```
void vvl::Semaphore::Retire(vvl::Queue *current_queue, const Location &loc, uint64_t payload) {
    auto guard = WriteLock();
    if (payload <= completed_.payload) {
        return;
    }
    ...
```

`CurrentPayload() < payload` prevents adding signal that does not make forward progress. It is ignored by Retire and gets stuck in semaphore timeline.

spencer-lunarg (approved):

PR #8397, opened 2024-08-15
gpu: Remove self validation from bindless shader

spencer-lunarg:
Removes some size from `inst_bindless_descriptor` where we don't need it


layers/gpu/shaders/instrumentation/bindless_descriptor.comp
```diff
@@ -79,18 +79,7 @@ bool inst_bindless_descriptor(const uint inst_num, const uvec4 stage_info, const
     uint param_0 = 0u;
     uint param_1 = 0u;
     do {
-        if (desc_set >= 32u) {
```
spencer-lunarg, line 82: these are both known at instrumentation time, moved check to the instrumentation pass and it will just not attempt to instrumentation that spot and add a warning

layers/gpu/shaders/instrumentation/bindless_descriptor.comp
```diff
@@ -79,18 +79,7 @@ bool inst_bindless_descriptor(const uint inst_num, const uvec4 stage_info, const
...
-            error = kErrorSubCodeBindlessDescriptorBounds;
-            param_0 = desc_set;
-            break;
-        }
         DescriptorLayoutData layout_data = bindless_state_buffer.desc_sets[desc_set].layout_data;
-        // Check for null pointer like this to not enforce needing Int64 support
-        uvec2 ptr_check = uvec2(layout_data);
-        if ((ptr_check.x == 0u) && (ptr_check.y == 0u)) {
```
spencer-lunarg, line 90: 1. These should never be null, we have (and if don't should) wrap all our internal calls to `vkMapMemory` and `vkGetDeviceAddress` such that this will get reported first on the CPU
2. Soon with debugprintf, we can just print these out if we are still concerned while debugging
3. Until we find it useful, I don't want to have the complexity of a macro and a "debug" version of these shaders

layers/gpu/shaders/instrumentation/bindless_descriptor.comp
spencer-lunarg: Also note, no tests fail because we actually had no way to test these and are likely to blow up before here

layers/gpu/spirv/bindless_descriptor_pass.cpp
```diff
@@ -428,6 +429,11 @@ bool BindlessDescriptorPass::AnalyzeInstruction(const Function& function, const
         }
     }
 
+    if (descriptor_set_ >= gpuav::glsl::kDebugInputBindlessMaxDescSets) {
+        module_.InternalWarning("BindlessDescriptorPass", "Tried to use a descriptor slot over the current max limit");
```
arno-lunarg, line 433: Print `descriptor_set_` in message

arno-lunarg (approved):

layers/gpu/spirv/bindless_descriptor_pass.cpp
spencer-lunarg: next PR will be adjusting this function and will have it printed

spencer-lunarg: (I was adjusting this code and thought the same thing, so it is already in the next change lol)

Issue #8376, opened 2024-08-11
False positive when calling vkResetFences right after vkDeviceWaitIdle when the fence was used in VkSwapchainPresentFenceInfoEXT


knokko:
**Environment:**
 - OS: Windows 10, possibly more
 - GPU and driver version: NVIDIA GeForce RTX 3070 Ti driver version 552.44.0.0
 - SDK or header version if building from repo: 1.3.290
 - Options enabled (synchronization, best practices, etc.): synchronization and best practices

**Describe the Issue**
When my applications calls `vkResetFences` right after `vkDeviceWaitIdle`, the validation layer reports:
```
[ VUID-vkResetFences-pFences-01123 ] Object 0: handle = 0xd10d270000000018, name = PresentFence-1, type = VK_OBJECT_TYPE_FENCE; | MessageID = 0x68a5074e | vkResetFences(): pFences[1] (VkFence 0xd10d270000000018[PresentFence-1]) is in use. The Vulkan spec states: Each element of pFences must not be currently associated with any queue command that has not yet completed execution on that queue (https://vulkan.lunarg.com/doc/view/1.3.290.0/windows/1.3-extensions/vkspec.html#VUID-vkResetFences-pFences-01123)
```
I believe this is a false positive because `vkDeviceWaitIdle` should ensure that all commands have completed execution. The validation layer goes away when I explicitly call `vkWaitForFences` between `vkDeviceWaitIdle` and `vkResetFences`.

I don't know how to make a **minimal** reproduce example in a reasonable about of time, but I did publish a [more complicated reproducible example](https://github.com/knokko/vk-boiler/tree/validation-layer-false-positive).

**Expected behavior**

No validation errors should be reported.

**Valid Usage ID**
- During initialization, `BestPractices-vkCreateSwapchainKHR-no-VkSwapchainPresentModesCreateInfoEXT-provided` is reported. (Probably not important.)
- This bug report is about `VUID-vkResetFences-pFences-01123`

**Additional context**

The offending fences are used in `VkSwapchainPresentFenceInfoEXT`. No validation errors are reported for any other fences, which suggests that there is something special about fences used in `VkSwapchainPresentFenceInfoEXT`. This also means that **a driver with swapchain maintenance support is required to reproduce this**.

This bug did not yet happen with the validation layers from 1.3.268. I haven't tried the SDKs between 1.3.268 and 1.3.290.

<details>
<summary>api dump + validation output</summary>

```sh
[...]

PR #8400, opened 2024-08-16
layers: Wait for present fences in QueueWaitIdle

artem-lunarg:
Fixes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8376
Based on this discussion: https://gitlab.khronos.org/vulkan/vulkan/-/issues/3962

This adds `QueueSubmission` object per swapchain (instead of per present as it was before). The reason is that present fence is defined per swapchain. This allows to use `QueueSubmission::fence` field to keep present fence and reuse existing fence tracking code path without additional changes. The first commit does `QueueSubmission` per swapchain change. The fix is the second commit, just adds the fence to `QueueSubmission`.

### DIFF
--- a/layers/state_tracker/state_tracker.cpp
+++ b/layers/state_tracker/state_tracker.cpp
@@ -3735,16 +3735,25 @@ void ValidationStateTracker::PostCallRecordQueuePresentKHR(VkQueue queue, const 
     }
 
     auto queue_state = Get<vvl::Queue>(queue);
-    Location submit_loc = record_obj.location.dot(vvl::Field::pPresentInfo);
-    std::vector<vvl::QueueSubmission> submissions;
-    submissions.emplace_back(submit_loc);
-    vvl::PresentSync present_sync;
+    const Location present_loc = record_obj.location.dot(vvl::Field::pPresentInfo);
+
+    // Add submission record per swapchain. Each swapchain can have its own fence to wait for.
+    std::vector<vvl::QueueSubmission> present_submissions;
+    present_submissions.reserve(pPresentInfo->swapchainCount);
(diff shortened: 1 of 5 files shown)


layers/state_tracker/state_tracker.cpp
```diff
@@ -3735,16 +3735,31 @@ void ValidationStateTracker::PostCallRecordQueuePresentKHR(VkQueue queue, const
...
+    // Create present QueueSubmission objects
+    std::vector<vvl::QueueSubmission> present_submissions;  // TODO: use small_vector. Need to update interfaces to use vvl::span
+    small_vector<std::shared_ptr<vvl::Fence>, 1> present_fences;
+    const auto *present_fence_info = vku::FindStructInPNextChain<VkSwapchainPresentFenceInfoEXT>(pPresentInfo->pNext);
+    for (uint32_t i = 0; i < pPresentInfo->swapchainCount; ++i) {
+        present_submissions.emplace_back(present_loc.dot(vvl::Field::pSwapchains, i));
+        if (present_fence_info) {
+            present_fences.emplace_back(Get<vvl::Fence>(present_fence_info->pFences[i]));
+            auto moveable_fence = present_fences.back();
+            present_submissions.back().AddFence(std::move(moveable_fence));
```
artem-lunarg, line 3749: This line adds present fence to present `QueueSubmission`. Then the fence in-use status will be updated by existing code as in the regular submissions.

artem-lunarg:
CI issues on Android with swapchain maintenance1. Blacklisting on GalaxyS24/Pixel8 similar to other swapchain maintenance1 tests.

spencer-lunarg (approved):

PR #8402, opened 2024-08-16
gpu: Add NonBindlessOOBBuffer Pass

spencer-lunarg:
WIP


layers/gpu/spirv/non_bindless_oob_texel_buffer_pass.cpp
```diff
@@ -0,0 +1,234 @@
...
+    const uint32_t image_word = OpcodeImageAccessPosition(opcode);
+
+    image_inst_ = function.FindInstruction(inst.Word(image_word));
+    if (!image_inst_) return false;
+    const Type* image_type = module_.type_manager_.FindTypeById(image_inst_->TypeId());
+    if (!image_type) return false;
+
+    const uint32_t dim = image_type->inst_.Operand(1);
+    if (dim != spv::DimBuffer) {
+        return false;  // It is a Storage Image
```
spencer-lunarg, line 102: this actually opened my eyes that we don't do OOB checking on Storage Images right now neither, this will require having to pass in more information into the shader.

... pointing this out because having these descriptor types in separate passes (instead in a mega pass like bindless is) makes it easier to spot this

scripts/generate_source.py
```diff
@@ -379,6 +379,10 @@ def main(argv):
         'instrumentation_buffer_device_address_comp.cpp',
         'instrumentation_bindless_descriptor_comp.h',
         'instrumentation_bindless_descriptor_comp.cpp',
+        'instrumentation_non_bindless_oob_buffer_comp.h',
+        'instrumentation_non_bindless_oob_buffer_comp.cpp',
+        'instrumentation_non_bindless_oob_texel_buffer_comp.h',
+        'instrumentation_non_bindless_oob_texel_buffer_comp.cpp',
```
spencer-lunarg, line 385: plan to make another change to remove needing to list yet-another-place when adding a new instrumentation

layers/gpu/instrumentation/gpu_shader_instrumentor.cpp
```diff
@@ -499,10 +500,24 @@ void GpuShaderInstrumentor::PreCallRecordShaderObjectInstrumentation(
...
+            }
+            if (has_bindless_descriptors) break;
+        }
+    }
+
     if (!cached) {
         pass = InstrumentShader(
             vvl::make_span(static_cast<const uint32_t *>(create_info.pCode), create_info.codeSize / sizeof(uint32_t)),
-            unique_shader_id, create_info_loc, instrumented_spirv);
+            unique_shader_id, has_bindless_descriptors, create_info_loc, instrumented_spirv);
```
arno-lunarg, line 520: passing `has_bindless_descriptors` here does not feel right, as this info is specific to well bindless descriptors validation. 
To me it looks like this info is only needed in `RunPassBindlessDescriptor`, quite further down the `InstrumentShader` rabbit hole is

layers/gpu/spirv/non_bindless_oob_texel_buffer_pass.cpp
```diff
@@ -0,0 +1,234 @@
...
+        spv::OpFunctionCall,
+        {bool_type, function_result, function_def, injection_data.inst_position_id, injection_data.stage_info_id,
+         descriptor_array_size_id_, set_constant.Id(), binding_constant.Id(), descriptor_index_id, descriptor_offset_id_},
+        inst_it);
+
+    return function_result;
+}
+
+void NonBindlessOOBTexelBufferPass::Reset() {
+    access_chain_inst_ = nullptr;
```
arno-lunarg, line 77: I think you could go:
```suggestion
*this = NonBindlessOOBTexelBufferPass();
```
Simpler and makes sure no member is forgotten.
But not 100% sure as `NonBindlessOOBTexelBufferPass` is inheriting from `Pass`

layers/gpu/spirv/non_bindless_oob_texel_buffer_pass.cpp
```diff
@@ -0,0 +1,234 @@
...
+    const Type* image_type = module_.type_manager_.FindTypeById(image_inst_->TypeId());
+    if (!image_type) return false;
+
+    const uint32_t dim = image_type->inst_.Operand(1);
+    if (dim != spv::DimBuffer) {
+        return false;  // It is a Storage Image
+    }
+    const uint32_t depth = image_type->inst_.Operand(2);
+    const uint32_t arrayed = image_type->inst_.Operand(3);
+    const uint32_t ms = image_type->inst_.Operand(4);
```
arno-lunarg, line 106:
```suggestion
    const uint32_t multi_sampling = image_type->inst_.Operand(4);
```
if it is indeed multi sampling?

layers/gpu/spirv/non_bindless_oob_texel_buffer_pass.cpp
```diff
@@ -0,0 +1,234 @@
...
+        const Variable* variable = module_.type_manager_.FindVariableById(var_inst_->Operand(0));
+        if (!variable) {
+            module_.InternalError("NonBindlessOOBTexelBufferPass", "OpAccessChain base is not a variable");
+            return false;
+        }
+        var_inst_ = &variable->inst_;
+
+        const Type* pointer_type = variable->PointerType(module_.type_manager_);
+        if (pointer_type->inst_.Opcode() == spv::OpTypeRuntimeArray) {
+            return false;  // Currently we mark these are "bindless"
```
arno-lunarg, line 153:
```suggestion
            return false;  // Currently we mark these as "bindless"
```

tests/unit/gpu_av_oob.cpp
```diff
@@ -1075,7 +1075,64 @@ TEST_F(NegativeGpuAVOOB, TexelFetch) {
...
+}
+
+TEST_F(NegativeGpuAVOOB, TexelFetchArray) {
+    TEST_DESCRIPTION("index into texelFetch OOB for an array of TexelBuffers");
+    SetTargetApiVersion(VK_API_VERSION_1_2);
+    AddDisabledFeature(vkt::Feature::robustBufferAccess);
+    RETURN_IF_SKIP(InitGpuAvFramework());
+    RETURN_IF_SKIP(InitState());
+
+    char const *shader_source = R"glsl(
```
arno-lunarg, line 1091:
```suggestion
    char const *cs_source = R"glsl(
```
It is the name I saw the most for compute shaders
No relation to
header-3477294717

arno-lunarg (approved):
image

layers/gpu/spirv/non_bindless_oob_texel_buffer_pass.cpp
spencer-lunarg: this would reset `link_function_id` back to `0`, that is suppose to be the same between runs... agree this is error prone, but also the way around it would be to have a struct in the class that could be reset like this

spencer-lunarg: will add to a future enhancement todo list

PR #8410, opened 2024-08-19
layers: Spec fixes for 1.3.293

spencer-lunarg:
2 fixes

1. ` VUID-vkCmdDraw-None-06666` now requires `VK_EXT_sample_locations` (because when using Shader Object that extension has no feature bit to turn on)
2. Fixes some VUID churn around the Object Tracker parent VUIDs not being assigned yet

### DIFF
--- a/layers/core_checks/cc_cmd_buffer_dynamic.cpp
+++ b/layers/core_checks/cc_cmd_buffer_dynamic.cpp
@@ -21,6 +21,7 @@
 #include "generated/chassis.h"
 #include "core_validation.h"
 #include "drawdispatch/drawdispatch_vuids.h"
+#include "generated/vk_extension_helper.h"
 #include "state_tracker/image_state.h"
 #include "state_tracker/render_pass_state.h"
 #include "state_tracker/shader_object_state.h"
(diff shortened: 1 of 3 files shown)


layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -248,7 +249,7 @@ bool CoreChecks::ValidateGraphicsDynamicStateSetStatus(const LastBound& last_bou
             skip |= ValidateDynamicStateIsSet(last_bound_state, state_status_cb, CB_DYNAMIC_STATE_FRONT_FACE, vuid);
         }
 
-        if (last_bound_state.IsSampleLocationsEnable()) {
+        if (last_bound_state.IsSampleLocationsEnable() && IsExtEnabled(device_extensions.vk_ext_sample_locations)) {
```
artem-lunarg, line 252: Is it possible to do this check inside `IsSampleLocationsEnable`? Is the condition only for the case when `pipeline_state` is null?

spencer-lunarg: ya, I guess we can, let me do that then

spencer-lunarg: actually no, the `device_extensions` is not part of the `LastBound` object

also this starts to get in the grey area of "what is the default value"

artem-lunarg (approved):

PR #8412, opened 2024-08-19
gpu: Pass all tests with self val

spencer-lunarg:
All GPU-AV tests pass with Self Validation now on Mesa Intel

### DIFF
--- a/layers/gpu/instrumentation/gpuav_instrumentation.cpp
+++ b/layers/gpu/instrumentation/gpuav_instrumentation.cpp
@@ -174,11 +174,14 @@ void SetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer &cb_sta
     const uint32_t error_logger_i = static_cast<uint32_t>(cb_state.per_command_error_loggers.size());
     const std::array<uint32_t, 2> dynamic_offsets = {
         {operation_index * static_cast<uint32_t>(sizeof(uint32_t)), error_logger_i * static_cast<uint32_t>(sizeof(uint32_t))}};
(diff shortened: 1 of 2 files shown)


layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -174,11 +174,14 @@ void SetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer &cb_sta
     const uint32_t error_logger_i = static_cast<uint32_t>(cb_state.per_command_error_loggers.size());
     const std::array<uint32_t, 2> dynamic_offsets = {
         {operation_index * static_cast<uint32_t>(sizeof(uint32_t)), error_logger_i * static_cast<uint32_t>(sizeof(uint32_t))}};
-    if ((pipeline_layout && pipeline_layout->set_layouts.size() <= gpuav.desc_set_bind_index_) &&
-        pipeline_layout_handle != VK_NULL_HANDLE) {
-        DispatchCmdBindDescriptorSets(cb_state.VkHandle(), bind_point, pipeline_layout_handle, gpuav.desc_set_bind_index_, 1,
-                                      &instrumentation_desc_set, static_cast<uint32_t>(dynamic_offsets.size()),
-                                      dynamic_offsets.data());
+    if (pipeline_layout && pipeline_layout_handle != VK_NULL_HANDLE) {
```
spencer-lunarg, line 177: this will not be the only spot, but we have disabled tests and will need to make a better "if shader object or pipeline path" (to be fair, this is still needed in core val as well)

arno-lunarg (approved):

Issue #8415, opened 2024-08-20
GPU-AV Shader caching: Account for driver version


arno-lunarg:
Delete cache if driver version has changed

spencer-lunarg:
Also need to account for the instrumentation passes that are done

if you have 2 shaders that are the same input, but the first is only using `--buffer-device-address` pass, and the second only uses `--bindless-descriptor` pass, it will be cached from the first one and produce the incorrect output

spencer-lunarg:
^ Nvm - I forgot this was already done in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8021

spencer-lunarg:
@arno-lunarg this actually makes no sense why we would do this

the GPU-AV shader cache just decides skips running the shader instrumentation code or not, but the final SPIR-V is still the same regardless... what is the goal of deleting this cache if the driver changes, the output SPIR-V we send down will look **the same** to the driver and deleting the cache with a new driver version will have zero effect

I think this was created because of the bug that was fixed in our caching (in #8416)

arno-lunarg:
Ah yes exactly

PR #8416, opened 2024-08-20
gpu: Fix GPU-AV Caching

spencer-lunarg:
Combination of https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8021 and https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8381 broke how caching works

also will solve most issues that caused https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8415 to be raised

### DIFF
--- a/layers/gpu/core/gpu_shader_cache_hash.h
+++ b/layers/gpu/core/gpu_shader_cache_hash.h
@@ -25,16 +25,9 @@
 // or a GPUAV shader changes.
 #pragma pack(push, 1)
 struct ShaderCacheHash {
-    ShaderCacheHash(const GpuAVSettings& gpuav_settings)
-        : validate_descriptors_(gpuav_settings.validate_descriptors),
-          warn_on_robust_oob_(gpuav_settings.warn_on_robust_oob),
-          validate_bda_(gpuav_settings.validate_bda),
-          validate_ray_query_(gpuav_settings.validate_ray_query) {}
+    ShaderCacheHash(const GpuAVSettings::ShaderInstrumentation& settings) : shader_instrumentation_settings(settings) {}
     // Settings that are part of shader instrumentation that would need us to invalidate the cache
-    bool validate_descriptors_;
-    bool warn_on_robust_oob_;
-    bool validate_bda_;
-    bool validate_ray_query_;
+    const GpuAVSettings::ShaderInstrumentation shader_instrumentation_settings;
     const char gpu_av_shader_git_hash[sizeof(GPU_AV_SHADER_GIT_HASH)] = GPU_AV_SHADER_GIT_HASH;
 };
 #pragma pack(pop)
--- a/layers/gpu/core/gpu_shader_cache_hash.h
+++ b/layers/gpu/core/gpu_shader_cache_hash.h
@@ -25,8 +25,16 @@
 // or a GPUAV shader changes.
 #pragma pack(push, 1)
 struct ShaderCacheHash {
-    ShaderCacheHash(const GpuAVSettings& gpuav_settings) : gpuav_settings(gpuav_settings) {}
-    GpuAVSettings gpuav_settings{};
+    ShaderCacheHash(const GpuAVSettings& gpuav_settings)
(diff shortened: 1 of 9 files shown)


layers/gpu/core/gpu_shader_cache_hash.h
```diff
@@ -25,8 +25,16 @@
 // or a GPUAV shader changes.
 #pragma pack(push, 1)
 struct ShaderCacheHash {
-    ShaderCacheHash(const GpuAVSettings& gpuav_settings) : gpuav_settings(gpuav_settings) {}
-    GpuAVSettings gpuav_settings{};
```
spencer-lunarg, line 29: tried `const GpuAVSettings gpuav_settings` but this seems to constantly be giving random values on initialization causing this to sometimes fail

at the same time, toggling non-shader instrumentation settings should not effect caching

spencer-lunarg: (todo will be to maybe rename/regroup the settings as we add more and more shader instrumentation options)

layers/gpu/instrumentation/gpu_shader_instrumentor.cpp
```diff
@@ -509,7 +509,7 @@ void GpuShaderInstrumentor::PreCallRecordShaderObjectInstrumentation(
         instrumentation_data.unique_shader_id = unique_shader_id;
         create_info.pCode = instrumented_spirv.data();
         create_info.codeSize = instrumented_spirv.size() * sizeof(uint32_t);
-        if (gpuav_settings.cache_instrumented_shaders) {
+        if (gpuav_settings.cache_instrumented_shaders && !cached) {
```
spencer-lunarg, line 512: if we found in cache, no need to re-add it (we do this correctly for pipeline path, was missing for Shader Object path)

layers/gpu/instrumentation/gpu_shader_instrumentor.cpp
```diff
@@ -970,7 +970,8 @@ void GpuShaderInstrumentor::PreCallRecordPipelineCreationShaderInstrumentation(
         bool pass = false;
         std::vector<uint32_t> instrumented_spirv;
         if (gpuav_settings.cache_instrumented_shaders) {
-            unique_shader_id = hash_util::ShaderHash(module_state->spirv->words_.data(), module_state->spirv->words_.size());
+            unique_shader_id =
+                hash_util::ShaderHash(module_state->spirv->words_.data(), module_state->spirv->words_.size() * sizeof(uint32_t));
```
spencer-lunarg, line 974: we were only caching 1/4th of the incoming shader, which is why some things were cache hitting when it should have been a miss

layers/gpu/core/gpu_shader_cache_hash.h
```diff
@@ -25,8 +25,16 @@
...
+    ShaderCacheHash(const GpuAVSettings& gpuav_settings)
+        : validate_descriptors_(gpuav_settings.validate_descriptors),
+          warn_on_robust_oob_(gpuav_settings.warn_on_robust_oob),
+          validate_bda_(gpuav_settings.validate_bda),
+          validate_ray_query_(gpuav_settings.validate_ray_query) {}
+    // Settings that are part of shader instrumentation that would need us to invalidate the cache
+    bool validate_descriptors_;
```
arno-lunarg, line 34: We will 100% forget to update this struct when needed
Maybe put all shader instrumentation related settings in one struct `Foo`, then have `gpuav_settings` and the shader cache have one `Foo`, and just go `shader_cache.foo(gpuav_settings.foo)` in the constructor

spencer-lunarg: ya, will do that as a 2nd commit in this PR

spencer-lunarg: created commit, honestly having a dedicated `GpuAVSettings::ShaderInstrumentation` is good long term to know which settings are actually for instrumentation

arno-lunarg (approved):

PR #8421, opened 2024-08-20
docs: GPU-AV Bindless Descriptor Design Doc

spencer-lunarg:
My attempt to break down the bindless validation (vs out-of-bounds validation) for GPU-AV in the most simplest, concise way possible


docs/gpu_av_bindless.md
```diff
@@ -0,0 +1,196 @@
...
+
+It is very easy to mix up what is invalid because it is `out-of-bounds` or `invalid binding`.
+
+```glsl
+layout(set = 0, binding = 0) buffer StorageBuffer {
+    uint x;
+} Data[2];
+
+void main() {
+    Data[index].x = 0;
```
jeremyg-lunarg, line 33: You lost the definition for `index` here which would make it clear that its coming from a UBO and not known until runtime.

docs/gpu_av_bindless.md
```diff
@@ -0,0 +1,196 @@
...
+
+## When we validate
+
+The following is a break down of "when" we can validate descriptors:
+
+![alt_text](images/gpu_av_bindless_when_to_validate.png "When to validate diagram")
+
+All 3 spots should run the same "validate descriptor" logic
+
+The big thing to take from here is that `UPDATE_AFTER_BIND` doesn't mean we need to validate on the GPU, but can be done on the CPU if desired.
```
jeremyg-lunarg, line 136: ... but not until `vkQueueSubmit` is called.

docs/gpu_av_bindless.md
```diff
@@ -0,0 +1,196 @@
...
+
+In the above shader, in theory, this could be valid if the user never enters `if (some_condition)` where an invalid access will occur.
+
+Instead of dealing with some theoretical edge cases or trying to validate some things on the CPU vs GPU, the following is used to make things simpler for everyone:
+
+- If `PARTIALLY_BOUND` is **not** used, `Data` is accessed with a constant, validate that element is bound on the CPU.
+- If `PARTIALLY_BOUND` is **not** used, `Data` is accessed with a non-constant, we check all elements are bound on the CPU.
+    - Even if user "knows" what the non-constant will be, it is invalid, they need to use descriptor indexing!
+- If `PARTIALLY_BOUND` is used, move all work to GPU
+
```
jeremyg-lunarg, line 166: Its important to note that one of our worst perf cases, Doom Eternal, uses UPDATE_AFTER_BIND, but NOT PARTIALLY_BOUND and arrays of image 1 million image descriptors. Even though the above is all technically correct, it is useful to only check the descriptors (for things like matching image layout) that were actually used (as told to us by GPU-AV) since that's all that REALLY matters.

docs/gpu_av_bindless.md
```diff
@@ -0,0 +1,196 @@
...
+> `accessed` means there is a load/store/atomic on it in any function, even if that function is never called by the main function. If the `spirv-opt` wasn't used there might be a ghost binding declared, but never used, this is a case of not being an access.
+
```
jeremyg-lunarg, line 168: Another thing that should be discussed here is the SPIRV NonUniform decoration. If that's present, each shader invocation could produce a different index.

spencer-lunarg: yes, this also requires features bits from descriptor indexing to be on... will add something

jeremyg-lunarg (approved):

Issue #8418, opened 2024-08-20
VUID-VkPipelineShaderStageCreateInfo-pNext-02755 message is confusing


ziga-lunarg:
The error message printed will be:
`vkCreateComputePipelines(): pCreateInfos[0].stage subgroupSizeControl was not enabled. The Vulkan spec states: If a VkPipelineShaderStageRequiredSubgroupSizeCreateInfo structure is included in the pNext chain, the subgroupSizeControl feature must be enabled, and stage must be a valid bit specified in requiredSubgroupSizeStages (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkPipelineShaderStageCreateInfo-pNext-02755)`

The first part that comes from VVL is `pCreateInfos[0].stage subgroupSizeControl was not enabled`

This made me think that subgroupSizeControl was a properties flag where the pCreateInfos[0].stage has to be included, but that was not it.

PR #8427, opened 2024-08-21
layers: Improve RequiredSubgroupSizeCreateInfo messages

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8418

### DIFF
--- a/layers/core_checks/cc_spirv.cpp
+++ b/layers/core_checks/cc_spirv.cpp
@@ -2388,11 +2388,7 @@ bool CoreChecks::ValidateShaderStage(const ShaderStageState &stage_state, const 
                              local_size_x, local_size_y, local_size_z, phys_dev_props.limits.maxComputeWorkGroupInvocations);
         }
 
-        if (const auto *required_subgroup_size_features =
-                vku::FindStructInPNextChain<VkPipelineShaderStageRequiredSubgroupSizeCreateInfoEXT>(stage_state.GetPNext())) {
-            skip |= ValidateRequiredSubgroupSize(module_state, stage_state, *required_subgroup_size_features, invocations,
-                                                 local_size_x, local_size_y, local_size_z, loc);
-        }
+        skip |= ValidateRequiredSubgroupSize(module_state, stage_state, invocations, local_size_x, local_size_y, local_size_z, loc);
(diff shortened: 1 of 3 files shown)


layers/core_checks/cc_spirv.cpp
```diff
@@ -2388,11 +2388,7 @@ bool CoreChecks::ValidateShaderStage(const ShaderStageState &stage_state, const
                              local_size_x, local_size_y, local_size_z, phys_dev_props.limits.maxComputeWorkGroupInvocations);
         }
 
-        if (const auto *required_subgroup_size_features =
-                vku::FindStructInPNextChain<VkPipelineShaderStageRequiredSubgroupSizeCreateInfoEXT>(stage_state.GetPNext())) {
-            skip |= ValidateRequiredSubgroupSize(module_state, stage_state, *required_subgroup_size_features, invocations,
-                                                 local_size_x, local_size_y, local_size_z, loc);
-        }
+        skip |= ValidateRequiredSubgroupSize(module_state, stage_state, invocations, local_size_x, local_size_y, local_size_z, loc);
```
spencer-lunarg, line 2391: found that just doing the quick "if no pnext, quick return" for things like this is better to contain the fact that everything inside that function pretends to the `pNext` (and is makes it more obvious to write better error messages)

layers/core_checks/cc_spirv.cpp
```diff
@@ -2595,58 +2591,60 @@ bool CoreChecks::PreCallValidateGetShaderModuleCreateInfoIdentifierEXT(VkDevice
 }
 
 bool CoreChecks::ValidateRequiredSubgroupSize(const spirv::Module &module_state, const ShaderStageState &stage_state,
-                                              const VkPipelineShaderStageRequiredSubgroupSizeCreateInfoEXT &required_subgroup_size,
                                               uint64_t invocations, uint32_t local_size_x, uint32_t local_size_y,
                                               uint32_t local_size_z, const Location &loc) const {
     bool skip = false;
 
-    const uint32_t requiredSubgroupSize = required_subgroup_size.requiredSubgroupSize;
```
spencer-lunarg, line 2603: while here, updated `requiredSubgroupSize` to `required_subgroup_size`

ziga-lunarg (approved):

PR #8434, opened 2024-08-22
layers: Separate CreateBuffer for descriptor buffers

spencer-lunarg:
This moves the `VK_EXT_descriptor_buffer` (and `VK_KHR_buffer_device_address` as well) so that future extensions can not worry about blending things together

### DIFF
--- a/layers/core_checks/cc_buffer.cpp
+++ b/layers/core_checks/cc_buffer.cpp
@@ -20,6 +20,7 @@
 #include <string>
 
 #include <vulkan/vk_enum_string_helper.h>
+#include "error_message/error_location.h"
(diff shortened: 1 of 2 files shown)


layers/core_checks/cc_buffer.cpp
```diff
@@ -20,6 +20,7 @@
 #include <string>
 
 #include <vulkan/vk_enum_string_helper.h>
+#include "error_message/error_location.h"
```
spencer-lunarg, line 23: `clangd` added this for me, its a case were we were indirectly including this but better to have it explicit here

artem-lunarg (approved):

PR #8436, opened 2024-08-22
GPU-AV: Fix problems discovered with self validation

arno-lunarg:


layers/state_tracker/cmd_buffer_state.h
```diff
@@ -510,6 +510,7 @@ class CommandBuffer : public RefcountedStateObject {
         std::vector<std::byte> values{};
     };
     std::vector<PushConstantData> push_constant_data_chunks;
+    std::array<VkPipelineLayout, BindPoint_Count> push_constant_latest_used_layout{};
```
spencer-lunarg, line 513: this probably could have been moved to the GPU-AV command buffer object, and in GPU-AV in `PostCallRecordCmdPushConstants` update it, now CoreCheck also has to allocate yet-more memory per command buffer

spencer-lunarg (approved):

layers/state_tracker/cmd_buffer_state.h
arno-lunarg: True true, but I feel that 24 bytes for an object that does not necessitate dynamic allocations is ok-ish, just a drop in the massive pond that cmd buffer state is

artem-lunarg: yes, dynamic allocation even without performance cost takes additional memory for metadata, which I can only guess, is least 8 bytes

spencer-lunarg: The `CommandBuffer` object thing is a mess, I wasn't that hard on this because there is a bigger refactor needed here

PR #8437, opened 2024-08-22
gpu: Have all Passes have a name function

spencer-lunarg:
As we add more passes, and have common `Pass` member functions that might have to report an error, it is nice to have a `Name()`function to get which pass it was

### DIFF
--- a/layers/gpu/spirv/bindless_descriptor_pass.h
+++ b/layers/gpu/spirv/bindless_descriptor_pass.h
@@ -27,6 +27,7 @@ namespace spirv {
 class BindlessDescriptorPass : public InjectConditionalFunctionPass {
   public:
     BindlessDescriptorPass(Module& module) : InjectConditionalFunctionPass(module) {}
+    const char* Name() const final { return "BindlessDescriptorPass"; }
     void PrintDebugInfo() final;
 
   private:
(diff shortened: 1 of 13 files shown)


layers/gpu/spirv/bindless_descriptor_pass.h
```diff
@@ -27,6 +27,10 @@ namespace spirv {
 class BindlessDescriptorPass : public InjectConditionalFunctionPass {
   public:
     BindlessDescriptorPass(Module& module) : InjectConditionalFunctionPass(module) {}
+    const char* Name() const final {
+        static const std::string_view name{"BindlessDescriptorPass"};
+        return name.data();
```
arno-lunarg, line 32:
```suggestion
       return "BindlessDescriptorPass";
```

arno-lunarg (commented):
Just no need to bring out the big guns ^^

arno-lunarg (approved):

layers/gpu/spirv/bindless_descriptor_pass.h
spencer-lunarg: ya, probably just as good... also this is really only for error/warnings anyway

PR #8442, opened 2024-08-23
layers: Print ppEnabledExtensionNames string if unknown

spencer-lunarg:
If the extension is not known, it will print out `INVALID_EMPTY` (which is not helpful IMO)

We have the original `const char*` so this prints it out instead

### DIFF
--- a/layers/vulkan/generated/chassis.cpp
+++ b/layers/vulkan/generated/chassis.cpp
@@ -299,7 +299,7 @@ static void InstanceExtensionWhitelist(ValidationObject* layer_data, const VkIns
                                    loc.dot(vvl::Field::pCreateInfo).dot(vvl::Field::ppEnabledExtensionNames, i),
                                    "%s is not supported by this layer.  Using this extension may adversely affect validation "
                                    "results and/or produce undefined behavior.",
-                                   String(extension));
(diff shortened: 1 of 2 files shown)


layers/vulkan/generated/chassis.cpp
```diff
@@ -299,7 +299,7 @@ static void InstanceExtensionWhitelist(ValidationObject* layer_data, const VkIns
                                    loc.dot(vvl::Field::pCreateInfo).dot(vvl::Field::ppEnabledExtensionNames, i),
                                    "%s is not supported by this layer.  Using this extension may adversely affect validation "
                                    "results and/or produce undefined behavior.",
-                                   String(extension));
+                                   pCreateInfo->ppEnabledExtensionNames[i]);
```
artem-lunarg, line 302: Check this for nullptr in case stateless validation is disabled?

artem-lunarg (approved):

layers/vulkan/generated/chassis.cpp
spencer-lunarg: I checked, this will crash elsewhere (actually a lot of spots, not just in VVL) if there is a null item in an array of things that are not suppose to be null

PR #8453, opened 2024-08-26
layers: Added present fences to thread safety checks

artem-lunarg:
This should be merged upon a confirmation here https://gitlab.khronos.org/vulkan/vulkan/-/issues/3980


artem-lunarg:
It's confirmed now https://gitlab.khronos.org/vulkan/vulkan/-/issues/3980#note_498278

arno-lunarg (approved):

PR #7694, opened 2024-03-14
layers: Add VkSwapchainPresentFenceInfoEXT fence sync support

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7025
Also fixes https://github.com/KhronosGroup/Vulkan-Samples/pull/617 (@ShabbyX)

Pretty happy that `PresentSync` pass can be used here (12 strikes positively back).

PR #8454, opened 2024-08-26
layers: Track present fences (better)

artem-lunarg:
It's the second attempt after: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8400

Original present fence synchronization was implemented here: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7694.
It was based on PresentSync, although PresentSync was not designed for this (only for acquire-based synchronization), still it worked.

Then we accidentally ended up with two solutions to handle present fences: `QueueSubmission` and `PresentSync`. This solution removes `PresentSync` code path and uses `QueueSubmission` fence instead.

Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8376
Potentially fixes the issue ANGLE had with the previous approach. Two solutions worked at the same time and caused raced condition. It worth to say that race condition could be detected by the thread safety validation, but we don't have it for the present fence. If we get confirmation it will be added in a separate PR.

### DIFF
--- a/layers/state_tracker/queue_state.cpp
+++ b/layers/state_tracker/queue_state.cpp
@@ -87,7 +87,6 @@ vvl::SubmitResult vvl::Queue::PostSubmit(std::vector<vvl::QueueSubmission> &&sub
         }
         {
             auto guard = Lock();
-            result.last_submission_seq = submission.seq;
             PostSubmit(submission);
             submissions_.emplace_back(std::move(submission));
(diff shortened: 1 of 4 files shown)


artem-lunarg:
During ~1 hour ANGLE test session the issue is not reproducible anymore.

spencer-lunarg (approved):

PR #8455, opened 2024-08-26
layers: Add numerical_types

arno-lunarg:
Monday morning, Arno is tired of having VS chokes on the completion of `uint32_t` for the 1341234 times since he is here, time to act.

```cpp

#pragma once

#include <cstdint>

using u8 = uint8_t;
using u16 = uint16_t;
using u32 = uint32_t;
using u64 = uint64_t;

using i8 = int8_t;
using i16 = int16_t;
using i32 = int32_t;
using i64 = int64_t;

using f32 = float;
using f64 = double;
```

I have not made the switch from `float/double` to `f32/f64` yet, as there is no formal guarantee for those types are respectively 32 and 64 bits long. 
I think this is always the case on the platforms we target, so can I go ahead?


artem-lunarg:
>  VS choke on the completion of uint32_t

What does it mean?

arno-lunarg:
> > VS choke on the completion of uint32_t
> 
> What does it mean?

They are too many symbols starting with `uint`, so VS pretty much always does not suggest what I want. Like so:
image

artem-lunarg:
How dare you to suggest replacement of almost standard types that were blessed by the committee and are well known to an ordinary people due to the issue with a proprietary tool!

I'm skeptical, but also curious about other opinions. There are languages  that natively use this naming convention, including f32/f64 and in some cases even in c++ codebases, not sure about the exacts reason of the latter. In some cases I might be about better cross-platform guarantees,  IIRC c++ does not guarantee uintN_t are always available, and another reason to distance from c++ std library if codebase provides custom solution.

arno-lunarg:
> How dare you to suggest replacement of almost standard types that were blessed by the committee and are well known to an ordinary people due to the issue with a proprietary tool!

12592250_903691643076358_6115484917659803342_n820-4146048689

arno-lunarg:
> How dare you to suggest replacement of almost standard types that were blessed by the committee and are well known to an ordinary people due to the issue with a proprietary tool!
> 
> I'm skeptical, but also curious about other opinions. There are languages that natively use this naming convention, including f32/f64 and in some cases even in c++ codebases, not sure about the exacts reason of the latter. In some cases I might be about better cross-platform guarantees, IIRC c++ does not guarantee uintN_t are always available, and another reason to distance from c++ std library if codebase provides custom solution.

Since I am basically just adding aliases for the `<u>intN_t` types, we already had this problem of working with those types without having the guarantee that they are defined then.
I know I am touching the "holy" standard types, but these days those aliases are quite common, and make our programmer's life a bit easier, so I think this is quite worth it!

PR #8463, opened 2024-08-27
layers: Allow SPIRV-Tools commit to be a tag

spencer-lunarg:
For people working on private extension (and possibly if SPIR-V Tools adds tags in near future) we will want to go

```patch
diff --git a/scripts/known_good.json b/scripts/known_good.json
index 4237ef3ee..66cc20a39 100755
--- a/scripts/known_good.json
+++ b/scripts/known_good.json
@@ -43,7 +43,7 @@
                 "-DSPIRV_SKIP_TESTS=ON",
                 "-DSPIRV_SKIP_EXECUTABLES=OFF"
             ],
-            "commit": "363486479d4c8dfbfdf2e2dc397cd10aa15f80b0"
+            "commit": "some tag name"
         },
```

so this removes the (honestly over complex) logic to get the git sha into the `UUID` for shader caching and just takes the string and uses the same internal hashing we use for everything else

### DIFF
--- a/layers/utils/shader_utils.cpp
+++ b/layers/utils/shader_utils.cpp
@@ -24,27 +24,16 @@
 
 #include "generated/spirv_tools_commit_id.h"
 
-void ValidationCache::GetUUID(uint8_t *uuid) {
-    const char *sha1_str = SPIRV_TOOLS_COMMIT_ID;
-    // Convert sha1_str from a hex string to binary. We only need VK_UUID_SIZE bytes of
-    // output, so pad with zeroes if the input string is shorter than that, and truncate
-    // if it's longer.
-#if defined(__GNUC__) && (__GNUC__ > 8)
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wstringop-truncation"
-#endif
-    char padded_sha1_str[2 * VK_UUID_SIZE + 1] = {};  // 2 hex digits == 1 byte
-    std::strncpy(padded_sha1_str, sha1_str, 2 * VK_UUID_SIZE);
-#if defined(__GNUC__) && (__GNUC__ > 8)
-#pragma GCC diagnostic pop
-#endif
-    for (uint32_t i = 0; i < VK_UUID_SIZE; ++i) {
-        const char byte_str[] = {padded_sha1_str[2 * i + 0], padded_sha1_str[2 * i + 1], '\0'};
-        uuid[i] = static_cast<uint8_t>(std::strtoul(byte_str, nullptr, 16));
-    }
+void ValidationCache::SetUUID(uint8_t *uuid) {
+    // Currently not using the last 8 bytes, so make sure they are zero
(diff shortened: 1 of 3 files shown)


artem-lunarg (approved):

jeremyg-lunarg:
Are you ok if the tag moves? Eg if it points at a dev branch that gets force pushed.

spencer-lunarg:
> Are you ok if the tag moves? Eg if it points at a dev branch that gets force pushed.

ugh, why you have to come in with the hard hitting questions... ya, I guess that might effect things, let me digest this more

jeremyg-lunarg:
https://stackoverflow.com/questions/1862423/how-to-tell-which-commit-a-tag-points-to-in-git/1862542#1862542 says you do:
```
 git rev-list -n 1 origin/sdk-1.1.101
6471d151f6882a0f93a6919a33255615a5d1aa59
```

Ideally this would be handled by cmake somehow and remove the need for generate_spirv.py after each update of SPIRV-Tools. But I don't have a magic recipe for that...

spencer-lunarg:
I spent way to much time on this, my CMAKE foo is not good enough, but wanted to do something like

```cmake
# We need the hash of the SPIRV-Tools to know when to invalidate caches that rely on it.
# Because the known_good.json might use a development tag name, the sha might change underneath it, so we need to call git to get the real sha.
# We do it in the cmake, as it is easy to forget to re-run a python script and it will not change correctly.
# We have the check the code in so that BUILD.gn build works, but luckily no one is developing in that path so doesn't need to be update each time there.
get_target_property(SPIRV_TOOLS_OPT_LOCATION SPIRV-Tools-opt LOCATION)
get_filename_component(SPIRV_TOOLS_DIR "${SPIRV_TOOLS_OPT_LOCATION}" DIRECTORY)
set(SPIRV_TOOLS_COMMIT_FILE "${CMAKE_CURRENT_SOURCE_DIR}/${API_TYPE}/generated/spirv_tools_commit_id.h")

set(SHA_COMMAND "${CMAKE_COMMAND} -E echo \"#define SPIRV_TOOLS_COMMIT_SHA \\\"\$(git -C ${SPIRV_TOOLS_DIR} rev-parse HEAD)\\\"\" > ${SPIRV_TOOLS_COMMIT_FILE}")

add_custom_command(
    OUTPUT ${SPIRV_TOOLS_COMMIT_FILE}
    COMMAND ${SHA_COMMAND}
)

add_custom_target(update_spirv_tools_commit ALL DEPENDS ${SPIRV_TOOLS_COMMIT_FILE})
add_dependencies(vvl update_spirv_tools_commit)
```

and have it just write the file out

The issue is we need the sha checked in so things like `BUILD.gn` sees it... but it is hard to get cmake to re-run based off the sha from another repo

spencer-lunarg:
closing - as much as I would want to do this, someone with more CMake/Build.gn knowledge will need to do, for now, internal changes will just have to keep using commit hashes

PR #8465, opened 2024-08-27
ci: Add some job dependency

spencer-lunarg:
1. Move the formatting action over to the rest of the actions
2. Build a dependency workgraph to save on CI runtime for clearly bad PRs

The new flow is

Code formatting -> Check VVL (code gen is good) -> Builds


.github/workflows/vvl.yml
```diff
@@ -59,6 +74,7 @@ jobs:
       - run: cmake --install build --prefix /tmp
 
   linux:
+    needs: check_vvl
```
artem-lunarg, line 77: Did we run `check_vvl`  before?

artem-lunarg (approved):
LGTM. The flow looks exactly as described in the PR.

.github/workflows/vvl.yml
spencer-lunarg: we did, it was always just in parallel  though

Issue #8459, opened 2024-08-26
Move GPL instrumentation to linking time


spencer-lunarg:
related to https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7804

Some workflows will create 20x pipeline libraries then will link. GPU-AV takes a long time to load a gfxreconstruct trace because we are instrumenting (and compiling) these large instrumentation at startup time.

One idea is to move the GPL instrumentation to linking time (which would be at draw time) so that instead of a LARGE startup cost, we can move things over to a runtime/pipeline hitching cost

PR #8468, opened 2024-08-28
gpu: Defer GPL instrumenation until linking

spencer-lunarg:
For https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8459

Currently a little rough, if this works well, will want to cleanup the somewhat duplicate code path (already wanted to do it with shader object path)

### DIFF
--- a/layers/gpu/instrumentation/gpu_shader_instrumentor.cpp
+++ b/layers/gpu/instrumentation/gpu_shader_instrumentor.cpp
@@ -510,19 +510,7 @@ void GpuShaderInstrumentor::PreCallRecordShaderObjectInstrumentation(
         unique_shader_id = unique_shader_module_id_++;
     }
 
-    bool has_bindless_descriptors = false;
-    for (const auto [layout_i, set_layout] : vvl::enumerate(create_info.pSetLayouts, create_info.setLayoutCount)) {
-        if (auto set_layout_state = Get<vvl::DescriptorSetLayout>(*set_layout)) {
-            for (uint32_t i = 0; i < set_layout_state->GetBindingCount(); i++) {
-                const VkDescriptorBindingFlags flags = set_layout_state->GetDescriptorBindingFlagsFromIndex(i);
-                if (vvl::IsBindless(flags)) {
-                    has_bindless_descriptors = true;
-                    break;
-                }
-            }
-            if (has_bindless_descriptors) break;
-        }
-    }
+    const bool has_bindless_descriptors = HasBindlessDescriptors(create_info);
 
     if (!cached) {
         pass = InstrumentShader(
@@ -642,11 +630,20 @@ void GpuShaderInstrumentor::PreCallRecordCreateGraphicsPipelines(VkDevice device
         auto &new_pipeline_ci = chassis_state.modified_create_infos[i];
         new_pipeline_ci.initialize(&pipeline_state->GraphicsCreateInfo());
 
+        if (!NeedPipelineCreationShaderInstrumentation(*pipeline_state)) {
+            continue;
+        }
+
         const Location create_info_loc = record_obj.location.dot(vvl::Field::pCreateInfos, i);
         auto &shader_instrumentation_metadata = chassis_state.shader_instrumentations_metadata[i];
 
-        PreCallRecordPipelineCreationShaderInstrumentation(pAllocator, *pipeline_state, new_pipeline_ci, create_info_loc,
-                                                           shader_instrumentation_metadata);
+        if (pipeline_state->linking_shaders != 0) {
+            PreCallRecordPipelineCreationShaderInstrumentationGPL(pAllocator, *pipeline_state, new_pipeline_ci, create_info_loc,
+                                                                  shader_instrumentation_metadata);
+        } else {
+            PreCallRecordPipelineCreationShaderInstrumentation(pAllocator, *pipeline_state, new_pipeline_ci, create_info_loc,
+                                                               shader_instrumentation_metadata);
+        }
     }
 
     chassis_state.pCreateInfos = reinterpret_cast<VkGraphicsPipelineCreateInfo *>(chassis_state.modified_create_infos.data());
@@ -670,6 +667,10 @@ void GpuShaderInstrumentor::PreCallRecordCreateComputePipelines(VkDevice device,
         // Need to make a deep copy so if SPIR-V is inlined, user doesn't see it after the call
         auto &new_pipeline_ci = chassis_state.modified_create_infos[i];
         new_pipeline_ci.initialize(&pipeline_state->ComputeCreateInfo());
+
+        if (!NeedPipelineCreationShaderInstrumentation(*pipeline_state)) {
+            continue;
+        }
 
         const Location create_info_loc = record_obj.location.dot(vvl::Field::pCreateInfos, i);
         auto &shader_instrumentation_metadata = chassis_state.shader_instrumentations_metadata[i];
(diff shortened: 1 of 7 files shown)


tests/unit/gpu_av_oob.cpp
```diff
@@ -887,6 +887,101 @@ TEST_F(NegativeGpuAVOOB, GPLImageLoadStoreIndependentSets) {
...
+    descriptor_set.WriteDescriptorBufferInfo(1, write_buffer.handle(), 0, VK_WHOLE_SIZE, VK_DESCRIPTOR_TYPE_STORAGE_BUFFER);
+    descriptor_set.UpdateDescriptorSets();
+
+    uint32_t *data = (uint32_t *)offset_buffer.memory().map();
+    *data = 8;
+    offset_buffer.memory().unmap();
+
+    static const char vertshader[] = R"glsl(
+        #version 450
+        layout(set = 0, binding = 0) uniform Foo { uint index[]; };
```
arno-lunarg, line 916:
```suggestion
        layout(set = 0, binding = 0) uniform Uniform { uint offset_buffer[]; };
```

tests/unit/gpu_av_oob.cpp
```diff
@@ -887,6 +887,101 @@ TEST_F(NegativeGpuAVOOB, GPLImageLoadStoreIndependentSets) {
...
+        layout(set = 0, binding = 1) buffer StorageBuffer { uint data[]; };
```
arno-lunarg, line 917:
```suggestion
        layout(set = 0, binding = 1) buffer StorageBuffer { uint write_buffer[]; };
```

tests/unit/gpu_av_oob.cpp
arno-lunarg, line 916: Really we have this problem in every test where we bound resources to shader, where with the names quite bad and inconsistent, the mapping between the CPU descriptor set and the in shader layout needlessly becomes a guessing game

tests/unit/gpu_av_oob.cpp
```diff
@@ -887,6 +887,101 @@ TEST_F(NegativeGpuAVOOB, GPLImageLoadStoreIndependentSets) {
...
+    vkt::Buffer write_buffer(*m_device, 16, VK_BUFFER_USAGE_STORAGE_BUFFER_BIT, reqs);
+
+    OneOffDescriptorSet descriptor_set(m_device, {{0, VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER, 1, VK_SHADER_STAGE_ALL, nullptr},
+                                                  {1, VK_DESCRIPTOR_TYPE_STORAGE_BUFFER, 1, VK_SHADER_STAGE_ALL, nullptr}});
+    const vkt::PipelineLayout pipeline_layout(*m_device, {&descriptor_set.layout_});
+    descriptor_set.WriteDescriptorBufferInfo(0, offset_buffer.handle(), 0, VK_WHOLE_SIZE);
+    descriptor_set.WriteDescriptorBufferInfo(1, write_buffer.handle(), 0, VK_WHOLE_SIZE, VK_DESCRIPTOR_TYPE_STORAGE_BUFFER);
+    descriptor_set.UpdateDescriptorSets();
+
+    uint32_t *data = (uint32_t *)offset_buffer.memory().map();
```
arno-lunarg, line 910: Same complaints about namings here. I know you copy pasted it, but `data` should be just be illegal as a variable name x)
`data` => `offset_buffer_ptr`

tests/unit/gpu_av_oob.cpp
```diff
@@ -887,6 +887,101 @@ TEST_F(NegativeGpuAVOOB, GPLImageLoadStoreIndependentSets) {
...
+    m_commandBuffer->begin();
+    m_commandBuffer->BeginRenderPass(m_renderPassBeginInfo);
+    vk::CmdBindPipeline(m_commandBuffer->handle(), VK_PIPELINE_BIND_POINT_GRAPHICS, exe_pipe.handle());
+    vk::CmdBindDescriptorSets(m_commandBuffer->handle(), VK_PIPELINE_BIND_POINT_GRAPHICS, pipeline_layout.handle(), 0, 1,
+                              &descriptor_set.set_, 0, nullptr);
+    vk::CmdDraw(m_commandBuffer->handle(), 3, 1, 0, 0);
+    m_commandBuffer->EndRenderPass();
+    m_commandBuffer->end();
+
+    m_errorMonitor->SetDesiredError("VUID-vkCmdDraw-storageBuffers-06936", 3);
```
arno-lunarg, line 978: Again again, same complaints could be made elsewhere: what should be the source of this error? Not clear 
Having it said in some comment would be nice, it helps with reviews. I feel that to make a good review, you must fist identify the intentions, and then see if code matches those intentions. If you start guessing what the code tries to do from the code itself too much, review quality suffers

arno-lunarg (approved):
LGTM, no need to address my complaints in the test now, it was more of a global rant about some problems I think we have

tests/unit/gpu_av_oob_positive.cpp
```diff
@@ -299,6 +299,98 @@ TEST_F(PositiveGpuAVOOB, GPL) {
     m_default_queue->Wait();
 }
 
+TEST_F(PositiveGpuAVOOB, GPLNonInlined) {
+    TEST_DESCRIPTION("Make suree GPL works when shader modules are not inlined at pipeline creation time");
```
arno-lunarg, line 303: Explain the expected test error

layers/gpu/instrumentation/gpu_shader_instrumentor.cpp
```diff
@@ -764,8 +773,17 @@ void GpuShaderInstrumentor::PostCallRecordCreateGraphicsPipelines(VkDevice devic
...
         ASSERT_AND_CONTINUE(pipeline_state);
+
+        if (pipeline_state->create_flags & VK_PIPELINE_CREATE_LIBRARY_BIT_KHR) {
+            continue;  // Move all instrumentation until the final linking time
+        }
+
         auto &shader_instrumentation_metadata = chassis_state.shader_instrumentations_metadata[i];
-        PostCallRecordPipelineCreationShaderInstrumentation(*pipeline_state, shader_instrumentation_metadata);
+        if (pipeline_state->linking_shaders != 0) {
+            PostCallRecordPipelineCreationShaderInstrumentationGPL(*pipeline_state, pAllocator, shader_instrumentation_metadata);
```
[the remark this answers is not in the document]
spencer-lunarg: - so you create 4 pipeline libraries (all `VkPipeline` but can't be used)
- then you create a `VkPipeline` that links them all (I call it the `executable pipeline` a lot)

so 
- `VK_PIPELINE_CREATE_LIBRARY_BIT_KHR` means you are just a library and can't be used
- `if (pipeline_state->linking_shaders != 0)` means you have pipelines you are linking in (so you are the executable)

spencer-lunarg:
> no need to address my complaints in the test now

I need to make the other nit updates (also realized one more test I want to add) so happy to do it... but agree it is quite a mess and we only copy-and-paste it around like wild fire

layers/gpu/instrumentation/gpu_shader_instrumentor.cpp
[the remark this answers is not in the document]
spencer-lunarg: also you can have libraries that are partially linking other things in 

...... this needs a diagram, I have spent hours grasping this, the spec has all the information, but in the most scatter, "you need to be a spec lawyer" way to understand it.... but for short term, just trust me on this one, I have spent too many hours on GPL sadly to know this is logically the right thing

tests/unit/gpu_av_oob.cpp
spencer-lunarg: https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8558

Issue #8469, opened 2024-08-28
BestPractices-SyncObjects-HighNumberOfFences : off by one error


filnet:
**Environment:**
 - OS: Windows
 - GPU and driver version: NVidia
 - SDK or header version if building from repo: 1.3.290.0
 - Options enabled (synchronization, best practices, etc.): best practices

**Describe the Issue**

The BestPractices-SyncObjects-HighNumberOfFences warning triggers after creating the fifth fence and not after creating the fourth (the limit is defined as 3).
The check is done in a PreCallValidateCreateFence so maybe before the count is updated...

Side note: would be nice if the warning message included the limit and the current count.

PS: other such checks might have the same issue.

filnet:
This check seems a bit hard to not trigger as fences are not used only for in flight frames... I allow only two in flight frames so have only 2 fences for that. But I also use fences for waiting for queued commands to finish executing. So currently I have 7 fences and trigger this warning.

spencer-lunarg:
Made a PR to fix the error message, not fully sure if there is threading issues or what, but the `VkAmdBestPracticesLayerTest.NumSyncPrimitives` test shows we report on the 4th fence there

filnet:
My code is not multi threaded during the initialization phase.
I added a debug printf prior to every call to create_fence and the output shows that the warning is emitted on the 5th call and thereafter.
The warning is emmitted 3 times for 7 created fences. So one short.

```
	Line 111: CREATE FENCE
	Line 113: CREATE FENCE
	Line 193: CREATE FENCE
	Line 234: CREATE FENCE
	Line 377: CREATE FENCE
	Line 378: [Warning][Performance]"Validation Performance Warning: [ BestPractices-SyncObjects-HighNumberOfFences ] | MessageID = 0x
	Line 379: a9f4ff68 | vkCreateFence():  [AMD] [NVIDIA] High number of VkFence objects created.Minimize the amount of CPU-GPU synchr
	Line 380: onization that is used. Semaphores and fences have overhead. Each fence has a CPU and GPU cost with it."
	Line 381: CREATE FENCE
	Line 382: [Warning][Performance]"Validation Performance Warning: [ BestPractices-SyncObjects-HighNumberOfFences ] | MessageID = 0x
	Line 383: a9f4ff68 | vkCreateFence():  [AMD] [NVIDIA] High number of VkFence objects created.Minimize the amount of CPU-GPU synchr
[...]

PR #8470, opened 2024-08-28
bp: Improve max sync object message

spencer-lunarg:
for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8469

now prints

> Validation Performance Warning: [ BestPractices-SyncObjects-HighNumberOfFences ] | MessageID = 0xa9f4ff68 | vkCreateFence():  [AMD] [NVIDIA] High number of VkFence objects created. 4 created, but recommanded max is 3. Minimize the amount of CPU-GPU synchronization that is used. Semaphores and fences have overhead. Each fence has a CPU and GPU cost with it.

### DIFF
--- a/layers/best_practices/bp_synchronization.cpp
+++ b/layers/best_practices/bp_synchronization.cpp
@@ -517,12 +517,16 @@ bool BestPractices::PreCallValidateCreateSemaphore(VkDevice device, const VkSema
                                                    const ErrorObject& error_obj) const {
     bool skip = false;
     if (VendorCheckEnabled(kBPVendorAMD) || VendorCheckEnabled(kBPVendorNVIDIA)) {
-        if (Count<vvl::Semaphore>() > kMaxRecommendedSemaphoreObjectsSizeAMD) {
+        const size_t count = Count<vvl::Semaphore>();
+        if (count > kMaxRecommendedSemaphoreObjectsSizeAMD) {
             skip |= LogPerformanceWarning("BestPractices-SyncObjects-HighNumberOfSemaphores", device, error_obj.location,
                                           "%s %s High number of vkSemaphore objects created. "
+                                          "%zu created, but recommanded max is %" PRIu32
+                                          ". "
                                           "Minimize the amount of queue synchronization that is used. "
                                           "Semaphores and fences have overhead. Each fence has a CPU and GPU cost with it.",
-                                          VendorSpecificTag(kBPVendorAMD), VendorSpecificTag(kBPVendorNVIDIA));
+                                          VendorSpecificTag(kBPVendorAMD), VendorSpecificTag(kBPVendorNVIDIA), count,
+                                          kMaxRecommendedSemaphoreObjectsSizeAMD);
         }
     }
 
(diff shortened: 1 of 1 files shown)


layers/best_practices/bp_synchronization.cpp
```diff
@@ -517,12 +517,16 @@ bool BestPractices::PreCallValidateCreateSemaphore(VkDevice device, const VkSema
...
+        if (count > kMaxRecommendedSemaphoreObjectsSizeAMD) {
             skip |= LogPerformanceWarning("BestPractices-SyncObjects-HighNumberOfSemaphores", device, error_obj.location,
                                           "%s %s High number of vkSemaphore objects created. "
+                                          "%zu created, but recommanded max is %" PRIu32
+                                          ". "
                                           "Minimize the amount of queue synchronization that is used. "
                                           "Semaphores and fences have overhead. Each fence has a CPU and GPU cost with it.",
-                                          VendorSpecificTag(kBPVendorAMD), VendorSpecificTag(kBPVendorNVIDIA));
+                                          VendorSpecificTag(kBPVendorAMD), VendorSpecificTag(kBPVendorNVIDIA), count,
+                                          kMaxRecommendedSemaphoreObjectsSizeAMD);
```
artem-lunarg, line 529: not related to PR, but that's interesting that `kMaxRecommendedSemaphoreObjectsSizeAMD` constant is AMD branded but the condition above checks both AMD and NVIDIA.

artem-lunarg (approved):

layers/best_practices/bp_synchronization.cpp
spencer-lunarg: AMD added this first, then NVIDIA came in and went "ya we agree as well" and really don't have a "cross-vendor" constant naming convention yet, so for now it is fine

artem-lunarg: Makes sense, piece of vvl history.

filnet:
There are similar limits checked for semaphores, barriers, etc.

layers/best_practices/bp_synchronization.cpp
```diff
@@ -517,12 +517,16 @@ bool BestPractices::PreCallValidateCreateSemaphore(VkDevice device, const VkSema
...
     if (VendorCheckEnabled(kBPVendorAMD) || VendorCheckEnabled(kBPVendorNVIDIA)) {
-        if (Count<vvl::Semaphore>() > kMaxRecommendedSemaphoreObjectsSizeAMD) {
+        const size_t count = Count<vvl::Semaphore>();
+        if (count > kMaxRecommendedSemaphoreObjectsSizeAMD) {
             skip |= LogPerformanceWarning("BestPractices-SyncObjects-HighNumberOfSemaphores", device, error_obj.location,
                                           "%s %s High number of vkSemaphore objects created. "
+                                          "%zu created, but recommanded max is %" PRIu32
+                                          ". "
                                           "Minimize the amount of queue synchronization that is used. "
                                           "Semaphores and fences have overhead. Each fence has a CPU and GPU cost with it.",
```
filnet, line 527: "Each fence" => "Each semaphore".

spencer-lunarg: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8474

PR #8485, opened 2024-09-02
build: Update to header 1.3.295

mikes-lunarg:


layers/vulkan/generated/layer_chassis_dispatch.cpp
```diff
@@ -262,118 +262,6 @@ void UnwrapPnextChainHandles(ValidationObject* layer_data, const void* pNext) {
                     safe_struct->queryPool = layer_data->Unwrap(safe_struct->queryPool);
                 }
             } break;
-            case VK_STRUCTURE_TYPE_COMPUTE_PIPELINE_CREATE_INFO: {
```
mikes-lunarg, line 265: This looks suspect, I think it means that we won't unwrap handles in VkPipelineCreateInfoKHR properly

layers/vulkan/generated/object_tracker.cpp
```diff
@@ -4450,68 +4392,8 @@ void ObjectLifetimes::PreCallRecordDestroyPipelineBinaryKHR(VkDevice device, VkP
     RecordDestroyObject(pipelineBinary, kVulkanObjectTypePipelineBinaryKHR);
 }
 
-bool ObjectLifetimes::PreCallValidateGetPipelineKeyKHR(VkDevice device, const VkPipelineCreateInfoKHR* pPipelineCreateInfo,
```
spencer-lunarg, line 4453: this is also strange, not sure what is fully going on tbh, curious what CI will show

mikes-lunarg: I had to do this in safe_struct: [fbb4db92c6b2ac09003b2b8e5ceb978f4f2dda71](https://github.com/KhronosGroup/Vulkan-Utility-Libraries/commit/fbb4db92c6b2ac09003b2b8e5ceb978f4f2dda71)

 probably need something similar here

spencer-lunarg (approved):

layers/vulkan/generated/layer_chassis_dispatch.cpp
spencer-lunarg: just hit this now with https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8540

confirmed this should be here... everything is messed up because we made the API to act like an output where it is normally an input to the driver

PR #8489, opened 2024-09-03

### LINKED ISSUE #2492: GPU-AV check index buffer out-of-bounds

arno-lunarg:
Fixes #2492

Completely rework gpuav_draw, to make it easier to add validation, and debug existing one.
When adding validation for some VkBuffer, the 2 more importants parts are 1) creating the validation shader and 2) have an informative error message. gpuav_draw.cpp has been redesigned around that idea. Each shader is in charge of its small validation, no uber shader trying to cater to all needs. Doing that ends up being a mess, with the need to correctly pipe data to the validation shader.
Validation shaders are now more fine grained.
The C++ has been reworked around some template to streamline the process of adding validation. The general idea is the following: 1) Declase a new "shader" struct representing validation shader bindings and push contants
2) fill that struct
3) bind this struct and the validation pipeline
4) dispatch the appropriate number of draws

About 4): we used to dispatch one draw, and do for loop in the shader to scan buffer elements. GPU are not good at that, instead a proper amount of draws should be dispatched, each in charge of doing it's small validation logic.

Example of out of bounds Vertex Index error message:
```
Validation Error: [ VUID-vkCmdDrawIndexed-None-02721 ] Object 0: handle = 0x29bdd8f9570, type = VK_OBJECT_TYPE_QUEUE; Object 1: handle = 0x29bdd8ad0c0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x24afafc5 | vkCmdDrawIndexed():  Vertex index 666 is not within the smallest bound vertex buffer.
index_buffer[1] (666) + vertexOffset (0) = Vertex index 666
Smallest vertex buffer binding info:
- Buffer: VkBuffer 0xa21a4e0000000030[]
- Binding: 0
- Binding offset: 0
- Binding size: 36 bytes
Index buffer binding info:
- Buffer: VkBuffer 0x980b0000000002e[]
- Index type: VK_INDEX_TYPE_UINT32
- Binding offset: 0
- Binding size: 12 bytes (or 3 VK_INDEX_TYPE_UINT32)
. The Vulkan spec states: If robustBufferAccess is not enabled, and that pipeline was created without enabling VK_PIPELINE_ROBUSTNESS_BUFFER_BEHAVIOR_ROBUST_BUFFER_ACCESS_EXT for vertexInputs, then for a given vertex buffer binding, any attribute data fetched must be entirely contained within the corresponding vertex buffer binding, as described in Vertex Input Description (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdDrawIndexed-None-02721)
```

Out of bounds index:
```
Validation Error: [ VUID-VkDrawIndexedIndirectCommand-robustBufferAccess2-08798 ] Object 0: handle = 0x13f72c903a0, type = VK_OBJECT_TYPE_QUEUE; Object 1: handle = 0x13f72c9aee0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x294f0c80 | vkCmdDrawIndexedIndirect():  Index 4 is not within the bound index buffer. Computed from VkDrawIndexedIndirectCommand[0] (.firstIndex = 1, .indexCount = 3), stored in VkBuffer 0xa43473000000002d[]
Index buffer binding info:
- Buffer: VkBuffer 0x4b7df1000000002f[]
- Index type: VK_INDEX_TYPE_UINT32
- Binding offset: 0
- Binding size: 12 bytes (or 3 VK_INDEX_TYPE_UINT32)
Supplied buffer parameters in indirect command: offset = 0, stride = 20 bytes. The Vulkan spec states: If robustBufferAccess2 is not enabled, (indexSize ├ù (firstIndex + indexCount) + offset) must be less than or equal to the size of the bound index buffer, with indexSize being based on the type specified by indexType, where the index buffer, indexType, and offset are specified via vkCmdBindIndexBuffer or vkCmdBindIndexBuffer2KHR. If vkCmdBindIndexBuffer2KHR is used to bind the index buffer, the size of the bound index buffer is vkCmdBindIndexBuffer2KHR::size (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkDrawIndexedIndirectCommand-robustBufferAccess2-08798)
```

Still missing (planned for followup PRs):
- Index out of bounds of index buffer for non indirect indexed draw (indirect draw done)
- Need to redo `DestroyRenderPassMappedResources`
- Need to add last used pipeline/shader object to error messages


layers/gpu/cmd_validation/gpuav_draw.h
```diff
@@ -21,13 +21,67 @@
 
 struct Location;
 
+namespace vvl {
+class Buffer;
```
arno-lunarg, line 25: need to remove that

layers/gpu/cmd_validation/gpuav_draw.cpp
```diff
@@ -15,143 +15,77 @@
...
+#include "gpu/shaders/cmd_validation/draw_push_data.h"
+#include "generated/cmd_validation_draw_mesh_indirect_vert.h"
+#include "generated/cmd_validation_first_instance_vert.h"
+#include "generated/cmd_validation_count_buffer_vert.h"
+#include "generated/cmd_validation_draw_indexed_vert.h"
+#include "generated/cmd_validation_draw_indexed_indirect_index_buffer_vert.h"
+#include "generated/cmd_validation_draw_indexed_indirect_vertex_buffer_vert.h"
 
 namespace gpuav {
+namespace bufval {
```
spencer-lunarg, line 38:
```suggestion
namespace cmdval {
```

I guess not sure what "buffer" refers to when literally everything is becoming a buffer now

layers/gpu/cmd_validation/gpuav_draw.cpp
```diff
@@ -161,307 +95,995 @@ static VkPipeline GetDrawValidationPipeline(Validator &gpuav, SharedDrawValidati
...
+                                   smallest_vertex_buffer_binding.binding, smallest_vertex_buffer_binding.binding_info.offset,
+                                   smallest_vertex_buffer_binding.binding_info.size,
+
+                                   // Index buffer binding info
+                                   gpuav.FormatHandle(index_buffer_binding.buffer).c_str(),
+                                   string_VkIndexType(index_buffer_binding.index_type), index_buffer_binding.offset,
+                                   index_buffer_binding.size, max_indices_in_buffer,
+                                   string_VkIndexType(index_buffer_binding.index_type));
+                    break;
+                    break;
```
spencer-lunarg, line 914: BREAK... BREAK.... BREAKKKKKKK!

```suggestion
```

layers/gpu/cmd_validation/gpuav_draw.cpp
```diff
@@ -470,4 +1092,262 @@ void InsertIndirectDrawValidation(Validator &gpuav, const Location &loc, Command
...
+};
+
+void DrawIndexedIndirectVertexBuffer(Validator &gpuav, CommandBuffer &cb_state, const Location &loc, VkBuffer draw_buffer,
+                                     VkDeviceSize draw_buffer_offset, uint32_t draw_cmds_byte_stride, uint32_t draw_count,
+                                     VkBuffer count_buffer, VkDeviceSize count_buffer_offset, const char *vuid_oob_vertex) {
+    if (!gpuav.gpuav_settings.validate_indirect_draws_buffers) {
+        return;
+    }
+
+    if (gpuav.enabled_features.robustBufferAccess) {
```
spencer-lunarg, line 1154: need to double check (and maybe worth a quick comment as the robustness spec is not fun to read or know) but is it really vertex buffer need `robustBufferAccess` but above you use `robustBufferAccess2` for the index buffer?

layers/gpu/core/gpuav_record.cpp
```diff
@@ -367,7 +375,15 @@ void Validator::PreCallRecordCmdDrawIndirect(VkCommandBuffer commandBuffer, VkBu
         return;
     }
 
-    InsertIndirectDrawValidation(*this, record_obj.location, *cb_state, buffer, offset, count, VK_NULL_HANDLE, 0, stride);
+    auto indirect_buffer_state = Get<vvl::Buffer>(buffer);
+    if (!indirect_buffer_state) {
+        InternalError(LogObjectList(commandBuffer, buffer), record_obj.location, "buffer must be a valid VkBuffer handle");
```
spencer-lunarg, line 380:
```suggestion
        InternalError(commandBuffer, record_obj.location, "buffer must be a valid VkBuffer handle");
```

if we can't find it here, the `LogObjectList` isn't gonna be able to look it up neither

... I also wouldn't worry **too** much on these error, these are really just asserts and already better information then going `assert(false)`

layers/gpu/cmd_validation/gpuav_draw.cpp
```diff
@@ -161,307 +95,995 @@ static VkPipeline GetDrawValidationPipeline(Validator &gpuav, SharedDrawValidati
...
+        };
+
+    cb_state.per_command_error_loggers.emplace_back(std::move(error_logger));
+}
+
+template <>
+void FirstInstance<VkDrawIndirectCommand>(Validator &gpuav, CommandBuffer &cb_state, const Location &loc, VkBuffer draw_buffer,
+                                          VkDeviceSize draw_buffer_offset, uint32_t draw_count, VkBuffer count_buffer,
+                                          VkDeviceSize count_buffer_offset, const char *vuid) {
+    FirstInstance(gpuav, cb_state, loc, draw_buffer, draw_buffer_offset, sizeof(VkDrawIndirectCommand), "VkDrawIndirectCommand", 3,
```
spencer-lunarg, line 357:
```suggestion
    FirstInstance(gpuav, cb_state, loc, draw_buffer, draw_buffer_offset, sizeof(VkDrawIndirectCommand), vvl::Struct::VkDrawIndirectCommand, 3,
```

layers/gpu/cmd_validation/gpuav_draw.cpp
```diff
@@ -161,307 +95,995 @@ static VkPipeline GetDrawValidationPipeline(Validator &gpuav, SharedDrawValidati
...
+        desc_writes[1].pBufferInfo = &count_buffer_binding.info;
+
+        return desc_writes;
+    }
+};
+
+void FirstInstance(Validator &gpuav, CommandBuffer &cb_state, const Location &loc, VkBuffer draw_buffer,
+                   VkDeviceSize draw_buffer_offset, uint32_t draw_cmds_byte_stride, const char *draw_indirect_struct_name,
+                   uint32_t first_instance_member_pos, uint32_t draw_count, VkBuffer count_buffer, VkDeviceSize count_buffer_offset,
+                   const char *vuid) {
```
spencer-lunarg, line 263: Why do we need to pass in the VUID, can't we just grab it here with something like

`const char* vuid = in_struct == vvl::Struct::VkDrawIndexedIndirectCommand ? "VUID-VkDrawIndexedIndirectCommand-firstInstance-00554" ? "VUID-VkDrawIndirectCommand-firstInstance-00501"`

and save a parameter in this large function call

layers/gpu/core/gpuav_record.cpp
```diff
@@ -403,8 +425,18 @@ void Validator::PreCallRecordCmdDrawIndirectCount(VkCommandBuffer commandBuffer,
...
-                                 countBufferOffset, stride);
+
+    auto indirect_buffer_state = Get<vvl::Buffer>(buffer);
+    if (!indirect_buffer_state) {
+        InternalError(LogObjectList(commandBuffer, buffer), record_obj.location, "buffer must be a valid VkBuffer handle");
+        return;
+    }
+
+    bufval::CountBuffer(*this, *cb_state, record_obj.location, buffer, offset, sizeof(VkDrawIndirectCommand),
+                        "VkDrawIndirectCommand", stride, countBuffer, countBufferOffset,
```
spencer-lunarg, line 436:
```suggestion
                        vvl::Struct::VkDrawIndirectCommand, stride, countBuffer, countBufferOffset,
```

tests/unit/gpu_av_positive.cpp
```diff
@@ -1732,3 +1732,53 @@ TEST_F(PositiveGpuAV, PipelineLayoutMixing) {
     m_default_queue->Submit(*m_commandBuffer);
     m_default_queue->Wait();
 }
+
+TEST_F(PositiveGpuAV, VertexIndex) {
```
jeremyg-lunarg, line 1736: since you made gpu_av_index_buffer_positive.cpp, you might as well move this test there.

layers/gpu/resources/gpuav_subclasses.h
```diff
@@ -94,7 +94,7 @@ class CommandBuffer : public gpu_tracker::CommandBuffer {
     gpu::GpuResourcesManager gpu_resources_manager;
     // Using stdext::inplace_function over std::function to allocate memory in place
     using ErrorLoggerFunc =
-        stdext::inplace_function<bool(Validator &gpuav, const uint32_t *error_record, const LogObjectList &objlist), 128>;
+        stdext::inplace_function<bool(Validator &gpuav, const uint32_t *error_record, const LogObjectList &objlist), 192>;
```
spencer-lunarg, line 97: for my own info, what is 192 here and why did you know to increase it to this?

layers/gpu/shaders/cmd_validation/draw_indexed_indirect_index_buffer.vert
```diff
@@ -0,0 +1,83 @@
...
+
+#include "common.h"
+#include "draw_push_data.h"
+
+layout(push_constant) uniform UniformInfo {
+    DrawIndexedIndirectIndexBufferPushData pc;
+};
+
+/*
+struct VkDrawIndexedIndirectCommand {
```
spencer-lunarg, line 29: remove?

layers/gpu/shaders/cmd_validation/draw_mesh_indirect.vert
```diff
@@ -0,0 +1,75 @@
...
+	} 
+	if (y > pc.max_workgroup_count_y) {
+	    GpuavLogError2(kErrorGroupGpuPreDraw, kErrorSubCodePreDrawGroupCountY, y, gl_VertexIndex);
+	} 
+	if (z > pc.max_workgroup_count_z) {
+	    GpuavLogError2(kErrorGroupGpuPreDraw, kErrorSubCodePreDrawGroupCountZ, z, gl_VertexIndex);
+	} 
+	
+	uint total = x * y * z;
+    if (total > pc.max_workgroup_total_count) {
```
spencer-lunarg, line 71: note-to-self, get a clang-format flow for GLSL inplace

tests/unit/gpu_av_indirect_buffer.cpp
```diff
@@ -607,6 +667,566 @@ TEST_F(NegativeGpuAVIndirectBuffer, FirstInstance) {
...
+
+    m_commandBuffer->EndRenderPass();
+    m_commandBuffer->end();
+
+    m_default_queue->Submit(*m_commandBuffer);
+    m_default_queue->Wait();
+    m_errorMonitor->VerifyFound();
+}
+
+#if 0
```
spencer-lunarg, line 1144: Use `DISABLED_` with a comment on top

tests/unit/gpu_av_indirect_buffer.cpp
```diff
@@ -426,54 +439,67 @@ TEST_F(NegativeGpuAVIndirectBuffer, Mesh) {
...
         m_default_queue->Submit(*m_commandBuffer);
         m_default_queue->Wait();
         m_errorMonitor->VerifyFound();
-
         draw_ptr[2] = 0;
-        draw_ptr[1] = 0;
-        draw_ptr[0] = 0;
     }
+// total count can end up being really high, draw takes too long and times out
+#if 0
```
spencer-lunarg, line 483: remove?

tests/unit/gpu_av_indirect_buffer.cpp
```diff
@@ -607,6 +667,566 @@ TEST_F(NegativeGpuAVIndirectBuffer, FirstInstance) {
...
+
+    VkPipelineLayoutCreateInfo pipelineLayoutCreateInfo = vku::InitStructHelper();
+    vkt::PipelineLayout pipeline_layout(*m_device, pipelineLayoutCreateInfo);
+
+    char const *vsSource = R"glsl(
+        #version 450
+        
+        layout(location=0) in vec3 pos;
+        
+		void main() {
```
spencer-lunarg, line 743: slightly ill-formed

jeremyg-lunarg (commented):
Is there any way to measure the perf difference

layers/gpu/cmd_validation/gpuav_draw.cpp
jeremyg-lunarg: could also be 'drawval'

layers/gpu/cmd_validation/gpuav_draw.cpp
arno-lunarg: https://www.youtube.com/watch?v=NFeUko-lQHg

layers/gpu/cmd_validation/gpuav_draw.cpp
spencer-lunarg: actually why do we need `val`... everything is validation in GPU-AV (it is the `V` in it lol)

can this just be `gpuav::draw`?

layers/CMakeLists.txt
```diff
@@ -213,8 +213,18 @@ target_sources(vvl PRIVATE
     ${API_TYPE}/generated/cmd_validation_copy_buffer_to_image_comp.cpp
     ${API_TYPE}/generated/cmd_validation_dispatch_comp.h
     ${API_TYPE}/generated/cmd_validation_dispatch_comp.cpp
-    ${API_TYPE}/generated/cmd_validation_draw_vert.h
-    ${API_TYPE}/generated/cmd_validation_draw_vert.cpp
+	${API_TYPE}/generated/cmd_validation_count_buffer_vert.h
```
spencer-lunarg, line 216:
```suggestion
    ${API_TYPE}/generated/cmd_validation_count_buffer_vert.h
```

layers/gpu/cmd_validation/gpuav_draw.cpp
spencer-lunarg: wait, I just realize we have `GpuVuid::index_buffer_size` for this but it is not being used in here

spencer-lunarg (commented):
There is a LOT going on here, overall really really happy how well this turned out, it is all well localized to `gpuav_draw.cpp` (thanks to our work with the refactor for the last months!)

Some things to cleanup, but hopefully we can get this in soon and push from here as a new basepoint, this PR is really pushing things in a good direction for GPU-AV!

layers/gpu/cmd_validation/gpuav_draw.cpp
arno-lunarg: Depending on the VUID, you need to check for either `robustBufferAccess` or `robustBufferAccess2`. I double checked, and I think I got this right

layers/gpu/core/gpuav_record.cpp
arno-lunarg: ah yes good point!

layers/gpu/cmd_validation/gpuav_draw.cpp
arno-lunarg: What you point is part of the goal of this PR: get rid of IMO unnecessary abstraction levels.

The issue I see is that we have prime access to an info (here, the precise VUID we target), we dilute it in a blob we pass to a function (here, the function location), only to retrieve it in the function.
It makes code harder to follow, and more complex to write. 

For this reason I have `GpuVuid` at gunpoint, and want to get rid of that in the future. Usage is gone in the `cmd_validation` part, the same needs to be done for instrumentation. Doing this will imply getting rid of  abstraction around `LogInstrumentationError`. The architecture for error message logging needs to mimic the one in `cmd_validation`. Right now for instrumentation there is only one generic `error_logger`, I think there should be one per validation type. It would also allow to write better error message, each `error_logger` lambda being able to capture whatever info is needed
to print a good error message for one specific validation type. 
It will not be easy to do, but I think it would result in clearer, easier to maintain code, and more importantly, better error messages

layers/gpu/cmd_validation/gpuav_draw.cpp
```diff
@@ -161,307 +95,995 @@ static VkPipeline GetDrawValidationPipeline(Validator &gpuav, SharedDrawValidati
...
+};
+
+struct FirstInstanceValidationShader {
+    static size_t GetSpirvDwordsCount() { return cmd_validation_first_instance_vert_size; }
+    static const uint32_t *GetSpirv() { return cmd_validation_first_instance_vert; }
+
+    static const uint32_t desc_set_id = gpuav::glsl::kDiagPerCmdDescriptorSet;
+
+    glsl::FirstInstancePushData push_constants{};
+    gpu::BoundStorageBuffer draw_buffer_binding = {gpuav::glsl::kPreDrawBinding_IndirectBuffer};
```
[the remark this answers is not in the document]
arno-lunarg: you cannot call a constructor at member declaration site : /
For now `gpu::BoundStorageBuffer` is only used in this file, and this way, we can always add the constructor when it becomes truly needed

layers/gpu/resources/gpuav_subclasses.h
arno-lunarg: Here 192 is the byte size (as in the `sizeof`) of the closure type that `ErrorLoggerFunc` will store.  `inplace_function` allocates memory for closures, guess what, in place, so size needs to be statically defined.
Effectively, it dictates the memory amount you can capture in your lambda. I had to grow it to fit the additional captured data. But it is getting quite big, I will rethink that to save memory.

layers/gpu/shaders/cmd_validation/draw_indexed_indirect_index_buffer.vert
arno-lunarg: No, Jeremy added that as an explanation for the constants just below, it is good info

tests/unit/gpu_av_indirect_buffer.cpp
arno-lunarg: I want to leave this as info

layers/gpu/shaders/cmd_validation/draw_push_data.h
```diff
@@ -0,0 +1,102 @@
...
+
+// Bindings for all pre draw types
+const uint kPreDrawBinding_IndirectBuffer = 0;
+const uint kPreDrawBinding_CountBuffer = 1;
+const uint kPreDrawBinding_IndexBuffer = 2;
+
+// Flag values for all pre draw types
+
+// Set if the count buffer is bound
+const uint kPreDrawValidationFlags_CountBuffer = (1 << 0);
```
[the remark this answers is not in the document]
arno-lunarg: nope they were not, got rid of them!

layers/gpu/shaders/cmd_validation/draw_indexed_indirect_vertex_buffer.vert
```diff
@@ -0,0 +1,118 @@
...
+
+layout(set = kDiagPerCmdDescriptorSet, binding = kPreDrawBinding_IndexBuffer)
+readonly buffer IndexBuffer32 {
+    uint index_buffer[];
+};
+
+uint get_vertex_index(uint i) {
+    if (pc.index_width == 32) {
+        return index_buffer[i];
+    } else if (pc.index_width == 16) {
```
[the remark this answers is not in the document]
arno-lunarg: Ah yes my editor was not correctly configured... will go by hand and change that

layers/gpu/cmd_validation/gpuav_draw.cpp
```diff
@@ -15,143 +15,77 @@
...
+    gpu::DeviceMemoryBlock dummy_buffer;  // Used to fill unused buffers in validation pipelines
+    bool valid = false;
 
-        VkDescriptorSetLayoutCreateInfo ds_layout_ci = vku::InitStructHelper();
-        ds_layout_ci.bindingCount = static_cast<uint32_t>(bindings.size());
-        ds_layout_ci.pBindings = bindings.data();
-        result = DispatchCreateDescriptorSetLayout(device, &ds_layout_ci, nullptr, &ds_layout);
+    SharedDrawValidationResources(Validator &gpuav, const Location &loc) : vma_allocator(gpuav.vma_allocator_) {
+        VkBufferCreateInfo buffer_info = vku::InitStructHelper();
+        buffer_info.size = 64;
```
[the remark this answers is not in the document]
arno-lunarg: Buffer content is never used, but you still need have every storage buffer bound to use a shader, hence this dummy buffer. Size could be anything (valid)

layers/gpu/resources/gpu_resources.cpp
```diff
@@ -17,6 +17,7 @@
 
 #include "gpu/resources/gpu_resources.h"
 
+#include "gpu/instrumentation/gpu_shader_instrumentor.h"
```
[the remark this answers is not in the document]
arno-lunarg: terminated-the-apprentice-178992476

tests/framework/buffer_helper.h
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright (c) 2015-2016, 2020-2024 The Khronos Group Inc.
```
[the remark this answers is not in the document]
arno-lunarg: And

newyearnewme-1458901666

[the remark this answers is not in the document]
arno-lunarg: why is this so big though

tests/framework/buffer_helper.h
```diff
@@ -0,0 +1,36 @@
...
+
+#pragma once
+
+#include "binding.h"
+
+namespace vkt {
+
+Buffer VertexBuffer(const Device &dev, const std::vector<float> &vertices);
+template <typename IndexT>
+Buffer IndexBuffer(const Device &dev, const std::vector<IndexT> &indices) {
```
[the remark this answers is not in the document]
arno-lunarg: Hummm you made me think about it, and now `VertexBuffer` too is a template function, so I got rid of `buffer_helper.cpp`

layers/state_tracker/state_tracker.cpp
```diff
@@ -4964,8 +4964,10 @@ void ValidationStateTracker::PostCallRecordCmdBindVertexBuffers2(VkCommandBuffer
         auto buffer_state = Get<vvl::Buffer>(pBuffers[i]);
         vvl::VertexBufferBinding &vertex_buffer_binding = cb_state->current_vertex_buffer_binding_info[i + firstBinding];
         vertex_buffer_binding.buffer = pBuffers[i];
-        vertex_buffer_binding.size = (pSizes) ? pSizes[i] : VK_WHOLE_SIZE;
         vertex_buffer_binding.offset = pOffsets[i];
+        vertex_buffer_binding.size = (pSizes) ? pSizes[i] : VK_WHOLE_SIZE;
```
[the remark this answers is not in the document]
arno-lunarg: ah I add forgotten something, now it looks like:
```cpp
        vertex_buffer_binding.size = pSizes ? pSizes[i] : VK_WHOLE_SIZE;
        if (vertex_buffer_binding.size == VK_WHOLE_SIZE) {
            vertex_buffer_binding.size = vvl::Buffer::ComputeSize(buffer_state, pOffsets[i], VK_WHOLE_SIZE);
        }
```

layers/gpu/cmd_validation/gpuav_draw.cpp
```diff
@@ -15,143 +15,77 @@
...
+#include "generated/cmd_validation_first_instance_vert.h"
+#include "generated/cmd_validation_count_buffer_vert.h"
+#include "generated/cmd_validation_draw_indexed_vert.h"
+#include "generated/cmd_validation_draw_indexed_indirect_index_buffer_vert.h"
+#include "generated/cmd_validation_draw_indexed_indirect_vertex_buffer_vert.h"
 
 namespace gpuav {
+namespace bufval {
 
 struct SharedDrawValidationResources final {
```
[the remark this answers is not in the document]
arno-lunarg: removed

tests/CMakeLists.txt
```diff
@@ -38,6 +38,8 @@ target_sources(vk_layer_validation_tests PRIVATE
     framework/render.h
     framework/binding.h
     framework/binding.cpp
+	framework/buffer_helper.h
+	framework/buffer_helper.cpp
     framework/test_framework.cpp
     framework/ray_tracing_objects.h
     framework/ray_tracing_objects.cpp
```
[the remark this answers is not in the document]
arno-lunarg: argh

layers/gpu/shaders/cmd_validation/draw_push_data.h
```diff
@@ -0,0 +1,102 @@
...
+const uint kFirstInstanceFlags_DrawCountFromBuffer = uint(1) << 0;
+struct FirstInstancePushData {
+    uint flags;
+    uint draw_cmds_stride_dwords;
+    uint first_instance_member_pos;
+};
+
+struct CountBufferPushData {
+    uint draw_cmds_byte_stride;
+    uint64_t draw_buffer_offset;
```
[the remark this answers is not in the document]
arno-lunarg: I added logic in `Validator::PreCallRecordCreateDevice` to force feature on if available

arno-lunarg:
> Is there any way to measure the perf difference of the new multi-shader approach vs what was happening before?

I probably can come up with some small stress test to figure this out. And will also have a look at Doom Eternal and the lightweight VK sample

layers/gpu/cmd_validation/gpuav_draw.cpp
spencer-lunarg: sure, we can move this to outside this PR, I think the "how we pass VUID strings" is a massive issue that we have solved 10 different ways in 10 different places, so fine for now having a 11th way

tests/framework/buffer_helper.h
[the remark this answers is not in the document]
spencer-lunarg: that works, consistency is what I like

arno-lunarg:
So @jeremyg-lunarg for some perf stats:
- Index buffer validation costs around 4 ms on the Tiny mesh large lightweight VK sample. On my AMD laptop, I go from a frame time of 24 ms to 28 ms.
- On Doom Eternal, performance is.... better. Actually on main, there seem to be some kind of leak in GPU-AV buffer validation code. Frame time keeps increasing over time. It does not happen on this branch. And I checked with some custom logging that validation was actually performed. Since this branch fixes this "leak", I did not investigate it

jeremyg-lunarg:
> * On Doom Eternal, performance is.... better. Actually on main, there seem to be some kind of leak in GPU-AV buffer validation code. Maybe something happening in the GPU. Frame time keeps increasing over time. It does not happen on this branch. And I checked with some custom logging that validation was actually performed. Since this branch fixes this "leak", I did not investigate it

That's disturbing! But yeah since you've fixed it not worth figuring out what was happening in the old code.

layers/gpu/resources/gpuav_subclasses.cpp
```diff
@@ -438,6 +438,100 @@ void CommandBuffer::ClearCmdErrorsCountsBuffer(const Location &loc) const {
...
+        }
+
+        switch (dynamic_state) {
+            case CB_DYNAMIC_STATE_VIEWPORT: {
+                DispatchCmdSetViewport(VkHandle(), dynamic_state_value.first_viewport,
+                                       uint32_t(dynamic_state_value.viewports.size()), dynamic_state_value.viewports.data());
+                break;
+            }
+            case CB_DYNAMIC_STATE_SCISSOR: {
+                // DispatchCmdSetScissor(VkHandle(), dynamic_state_value.scissor_count
```
jeremyg-lunarg, line 454: is this supposed to be commented out?

tests/unit/gpu_av_indirect_buffer.cpp
```diff
@@ -607,6 +667,569 @@ TEST_F(NegativeGpuAVIndirectBuffer, FirstInstance) {
...
```
jeremyg-lunarg, line 817: this is poorly formatted

tests/unit/gpu_av_indirect_buffer.cpp
```diff
@@ -607,6 +667,569 @@ TEST_F(NegativeGpuAVIndirectBuffer, FirstInstance) {
...
+    InitRenderTarget();
+
+    VkPipelineLayoutCreateInfo pipelineLayoutCreateInfo = vku::InitStructHelper();
+    vkt::PipelineLayout pipeline_layout(*m_device, pipelineLayoutCreateInfo);
+
+    char const *vsSource = R"glsl(
+        #version 450
+        
+        layout(location=0) in vec3 pos;
+
```
jeremyg-lunarg, line 892: here too

tests/unit/gpu_av_indirect_buffer.cpp
```diff
@@ -607,6 +667,569 @@ TEST_F(NegativeGpuAVIndirectBuffer, FirstInstance) {
...
+    RETURN_IF_SKIP(InitState());
+    InitRenderTarget();
+
+    VkPipelineLayoutCreateInfo pipelineLayoutCreateInfo = vku::InitStructHelper();
+    vkt::PipelineLayout pipeline_layout(*m_device, pipelineLayoutCreateInfo);
+
+    char const *vsSource = R"glsl(
+        #version 450
+        
+        layout(location=0) in vec3 pos;
```
jeremyg-lunarg, line 963: actually these are all the same shader. can it be factored out to a single string constant?

arno-lunarg: I also want that, but then not having shader code inline makes it not obvious what `location=0` is, and thus how to properly fill the corresponding Vulkan vertex binding description. And having a helper a function that would basically spit out a valid graphics pipeline based on this single shader seems overkill?

layers/gpu/resources/gpuav_subclasses.cpp
arno-lunarg: Nope leftover! For now caring about setting back scissor state is not needed.

tests/unit/gpu_av_indirect_buffer.cpp
jeremyg-lunarg: you could always name is something that makes it obvious, like `kVertexShaderVec3Location0`. But not a big deal either way

layers/CMakeLists.txt
spencer-lunarg: @arno-lunarg still need to fix this

layers/gpu/core/gpuav_setup.cpp
```diff
@@ -104,13 +104,31 @@ void Validator::PreCallRecordCreateDevice(VkPhysicalDevice physicalDevice, const
...
-        add_missing_features();
+        add_bda_feature();
+    }
+
+    // shaderInt64
+    VkPhysicalDeviceFeatures physical_device_features{};
+    DispatchGetPhysicalDeviceFeatures(physicalDevice, &physical_device_features);
+    if (physical_device_features.shaderInt64) {
+        if (auto enabled_features = const_cast<VkPhysicalDeviceFeatures*>(modified_create_info->pEnabledFeatures)) {
+            enabled_features->shaderInt64 = VK_TRUE;
```
spencer-lunarg, line 121: will mark as a TODO (For me) after, but we need to cleanup how we do this, like we are not warning users if we are setting things to true under them

layers/gpu/resources/gpuav_subclasses.cpp
```diff
@@ -438,6 +438,124 @@ void CommandBuffer::ClearCmdErrorsCountsBuffer(const Location &loc) const {
...
+                }
+            } break;
+            case CB_DYNAMIC_STATE_FRONT_FACE: {
+                if (IsExtEnabledByCreateinfo(gpuav_.device_extensions.vk_ext_extended_dynamic_state)) {
+                    DispatchCmdSetFrontFaceEXT(VkHandle(), dynamic_state_value.front_face);
+                } else {
+                    DispatchCmdSetFrontFace(VkHandle(), dynamic_state_value.front_face);
+                }
+            } break;
+            default:
```
spencer-lunarg, line 553: there are a LOT of dynamic states missing here, like `VK_DYNAMIC_STATE_STENCIL_OP`

spencer-lunarg: if wanted, happy to fix up after

layers/state_tracker/cmd_buffer_state.h
```diff
@@ -216,6 +216,10 @@ class CommandBuffer : public RefcountedStateObject {
         bool depth_test_enable;
         // VK_DYNAMIC_STATE_DEPTH_BOUNDS_TEST_ENABLE
         bool depth_bounds_test_enable;
+        // VK_DYNAMIC_STATE_DEPTH_COMPARE_OP
+        VkCompareOp depth_compare_op;
```
spencer-lunarg, line 220: so `CommandBuffer` is already by far the largest memory object we have, this is going to add a lot of memory overhead

I knowing we are still missing 50+ states for the RestoreDynamicState and my worry is we are going to really bloat this up

we either should

1. Move these to the GPU-AV `CommandBuffer` state object
2. We just serialize the `vkCmdSet` calls and replay them back

... please note there is a LOT of nasty dynamic state you have not touched yet (see Viewport with Count)

layers/gpu/cmd_validation/gpuav_cmd_validation_common.cpp
```diff
@@ -78,11 +78,29 @@ void RestorablePipelineState::Create(vvl::CommandBuffer &cb_state, VkPipelineBin
...
+        rendering_info.layerCount = 1;
+        rendering_info.viewMask = 0;
+        rendering_info.colorAttachmentCount = 0;
+        DispatchCmdBeginRendering(cb_state.VkHandle(), &rendering_info);
+    }
 }
 
 void RestorablePipelineState::Restore() const {
+    if (rendering_info_) {
+        DispatchCmdEndRendering(cb_state_.VkHandle());
```
spencer-lunarg, line 98: discussing with driver people

1. This will trash runtime speed on some drivers if we are going

```c++
vkCmdBeginRendering();
vkCmdDrawIndexed();
vkCmdDrawIndexed();
vkCmdDrawIndexed();
vkCmdDrawIndexed();
vkCmdDrawIndexed();
vkCmdEndRendering():
```

too

```c++
vkCmdBeginRendering();
vkCmdDrawIndexed();
vkCmdEndRendering():
vkCmdBeginRendering();
vkCmdDrawIndexed();
vkCmdEndRendering():
vkCmdBeginRendering();
vkCmdDrawIndexed();
vkCmdEndRendering():
vkCmdBeginRendering();
vkCmdDrawIndexed();
vkCmdEndRendering():
vkCmdBeginRendering();
vkCmdDrawIndexed();
vkCmdEndRendering():
```

2. There might be dependency (via `VK_KHR_dynamic_rendering_local_read`)  that are going to just blow up if we don't chain it for them

arno-lunarg:
Working on a new PR

PR #8490, opened 2024-09-03
tests: Add some larger Descriptor Buffer tests

spencer-lunarg:
This adds some more full tests for descriptor buffers

this was half me trying to make sure i grasped the API before digging more into issues around it... half, when I want to write a test, I have a starting boilerplate to go off of

### DIFF
--- a/tests/unit/descriptor_buffer_positive.cpp
+++ b/tests/unit/descriptor_buffer_positive.cpp
@@ -11,6 +11,7 @@
 
 #include "../framework/layer_validation_tests.h"
 #include "../framework/pipeline_helper.h"
+#include "utils/vk_layer_utils.h"
 
 void DescriptorBufferTest::InitBasicDescriptorBuffer() {
     SetTargetApiVersion(VK_API_VERSION_1_2);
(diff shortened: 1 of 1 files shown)


tests/unit/descriptor_buffer_positive.cpp
```diff
@@ -137,3 +137,332 @@ TEST_F(PositiveDescriptorBuffer, BindingMidBuffer) {
     vk::CmdBindDescriptorBuffersEXT(m_commandBuffer->handle(), 1, &dbbi);
     m_commandBuffer->end();
 }
+
+static inline VkDeviceSize aligned_size(VkDeviceSize value, VkDeviceSize alignment) {
```
arno-lunarg, line 141: Use `Align` from `vk_layer_utils`

arno-lunarg (approved):

tests/unit/descriptor_buffer_positive.cpp
spencer-lunarg: I knew we had one, couldn't find it

Issue #8458, opened 2024-08-26
vkCmdSetVertexInputEXT is not validated correctly


ziga-lunarg:
There are many CTS tests hitting `VUID-vkCmdDraw-Input-07939` when I don't think they should, one example is `dEQP-VK.draw.renderpass.instanced.dynamic_state_draw_vk_primitive_topology_point_list`.

Here is a test to reproduce:

```
TEST_F(PositiveDynamicState, VertexInputMultiple) {
    TEST_DESCRIPTION("Test CmdSetVertexInputEXT with multiple bindings");
    SetTargetApiVersion(VK_API_VERSION_1_1);

    AddRequiredExtensions(VK_EXT_VERTEX_INPUT_DYNAMIC_STATE_EXTENSION_NAME);
    AddRequiredFeature(vkt::Feature::vertexInputDynamicState);
    RETURN_IF_SKIP(Init());
    InitRenderTarget();

    struct PerVertex {
        int a;
        float b;
    };

    struct PerInstance {
        uint32_t c;
        float d;
    };

    VkVertexInputBindingDescription2EXT bindings[2];
    bindings[0] = vku::InitStructHelper();
    bindings[0].binding = 0u;
    bindings[0].stride = sizeof(PerVertex);
    bindings[0].inputRate = VK_VERTEX_INPUT_RATE_VERTEX;
    bindings[0].divisor = 1u;
    bindings[1] = vku::InitStructHelper();
    bindings[1].binding = 1u;
    bindings[1].stride = 0u;
    bindings[1].inputRate = VK_VERTEX_INPUT_RATE_INSTANCE;
    bindings[1].divisor = 1u;

    VkVertexInputAttributeDescription2EXT attributes[4];
    attributes[0] = vku::InitStructHelper();
    attributes[0].location = 0u;
    attributes[0].binding = 0u;
    attributes[0].format = VK_FORMAT_R8_SINT;
    attributes[0].offset = offsetof(PerVertex, a);

    attributes[1] = vku::InitStructHelper();
    attributes[1].location = 1u;
    attributes[1].binding = 0u;
    attributes[1].format = VK_FORMAT_R32_SFLOAT;
    attributes[1].offset = offsetof(PerVertex, b);

    attributes[2] = vku::InitStructHelper();
    attributes[2].location = 2u;
    attributes[2].binding = 1u;
    attributes[2].format = VK_FORMAT_R8_UINT;
[...]

PR #8496, opened 2024-09-04
layers: Fix vkCmdSetVertexInputEXT false positive

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8458

Also made sure it fixed `dEQP-VK.draw.renderpass.instanced.dynamic_state_draw_vk_primitive_topology_point_list`

### DIFF
--- a/layers/core_checks/cc_cmd_buffer_dynamic.cpp
+++ b/layers/core_checks/cc_cmd_buffer_dynamic.cpp
@@ -1014,55 +1014,54 @@ bool CoreChecks::ValidateDrawDynamicState(const LastBound& last_bound_state, con
                 continue;
             }
             bool location_provided = false;
-            if (const auto* binding_state =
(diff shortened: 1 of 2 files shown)


layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -1014,55 +1014,54 @@ bool CoreChecks::ValidateDrawDynamicState(const LastBound& last_bound_state, con
                 continue;
             }
             bool location_provided = false;
-            if (const auto* binding_state =
```
spencer-lunarg, line 1017: simple fix is just changing this line to 

```c++
for (const auto& binding_state : cb_state.dynamic_state_value.vertex_bindings) {
``` 

(what we do everywhere else)

the shader has the `location` but the `variable_ptr->decorations.binding` is ment for descriptor (not vertex input) and was always zero, so it would not detect anything in binding 1 or above

while here, I did a quick return to remove a level of indenting

ziga-lunarg (approved):

Issue #8492, opened 2024-09-04
ERROR_DEVICE_LOST with "Descriptor indexing" enabled when using VK_EXT_descriptor_buffer


StefanPoelloth:
**Environment:**
 - OS: Windows 10 22H2 19045.4780
 - GPU and driver version: 560.94
 - SDK or header version if building from repo: af7b0a35d009b5ad6e0b280a5b81388608ebfe39
 - Options enabled (synchronization, best practices, etc.): Descriptor Indexing

**Describe the Issue**
When using VK_EXT_descriptor_buffer and VVL option "Descriptor indexing", the driver resets and the app errors with `ERROR_DEVICE_LOST`.
It happens in my production code and can be reproduced with the sample: [descriptor_buffer_basic.cpp](https://github.com/KhronosGroup/Vulkan-Samples/blob/main/samples/extensions/descriptor_buffer_basic/descriptor_buffer_basic.cpp )

**Expected behavior**

Either no crash and working or a message that descriptor indexing checks are not possible with descriptor buffers and disabling it internally to not crash.

spencer-lunarg:
We currently don't support `VK_EXT_descriptor_buffer` at all on GPU-AV (aka Descriptor indexing)

it should be producing a warning, but I need to just write actual Descriptor Buffer GPU-AV tests and confirm it will "work as normal" and not just crash on you like this

Thanks for reporting, will get fixed

StefanPoelloth:
@spencer-lunarg I can confirm that your PR fixes the crash for us.

PR #8500, opened 2024-09-05
gpu: Fix Descriptor Buffer GPU-AV crash

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8492

We will have to redo the entire Shader Instrumentation Flow to handle Descriptor Buffers

For now make use of `shader_instrumentation_enabled` to prevent any shader instrumentation hooks to take place when turned off

### DIFF
--- a/layers/gpu/core/gpu_settings.h
+++ b/layers/gpu/core/gpu_settings.h
@@ -16,6 +16,7 @@
  * limitations under the License.
  */
 #pragma once
+#include <cstdint>
 // Default values for those settings should match layers/VkLayer_khronos_validation.json.in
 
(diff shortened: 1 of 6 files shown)


arno-lunarg (approved):
Just use `IsShaderInstrumentationEnabled` instead of `shader_instrumentation_enabled`

spencer-lunarg:
> Just use IsShaderInstrumentationEnabled instead of shader_instrumentation_enabled

discussed that this is a bigger problem and that we should make a PR that just removes one (we both prefer keeping the `IsShaderInstrumentationEnabled()`)

PR #8510, opened 2024-09-09
gpu: Fix crash in `Pass::ConvertTo32()`

corporateshark:
Fixed a crash in `Pass::ConvertTo32()`.

Reproduce: https://github.com/corporateshark/lightweightvk/blob/master/samples/004_YUV.cpp


spencer-lunarg (approved):
will need to apply to `CastToUint32` as well, but I can do that after as it might be nice to reduce some copying-and-pasting in this code

Issue #8367, opened 2024-08-08
format-parameter VUs not checking for maintenance5


ziga-lunarg:
The following VUIDs do not check for maintenance5:

`VUID-vkGetPhysicalDeviceFormatProperties-format-parameter`
`VUID-vkGetPhysicalDeviceFormatProperties2-format-parameter`
`VUID-vkGetPhysicalDeviceImageFormatProperties-format-parameter`
`VUID-VkPhysicalDeviceImageFormatInfo2-format-parameter`
`VUID-vkGetPhysicalDeviceSparseImageFormatProperties-format-parameter`
`VUID-VkPhysicalDeviceSparseImageFormatInfo2-format-parameter`
`VUID-vkGetPhysicalDeviceImageFormatProperties-usage-parameter`
`VUID-VkPhysicalDeviceImageFormatInfo2-usage-parameter`
`VUID-vkGetPhysicalDeviceSparseImageFormatProperties-usage-parameter`
`VUID-VkPhysicalDeviceSparseImageFormatInfo2-usage-parameter`

With maintenance5 the following was added to the spec:

> Additionally, if maintenance5 is supported, any integer value representable in the range valid for the
> defined type is valid when used in a function that has a VkPhysicalDevice object as its first
> parameter. Physical device queries will either return results indicating lack of support, or ignore
> unsupported values when used as a bit flag in a Vk*Flags* parameter.

spencer-lunarg:
wow, did not even know that was part of `maintenance5`

spencer-lunarg:
@ziga-lunarg could you grab the CTS coverage that was hitting this (if you still remember), @jeremyg-lunarg is making re-works to the pNext stuff and we want to make sure we are not too strict now with how maintenance5 was suppose to work

ziga-lunarg:
> [@ziga-lunarg](https://github.com/ziga-lunarg) could you grab the CTS coverage that was hitting this (if you still remember), [@jeremyg-lunarg](https://github.com/jeremyg-lunarg) is making re-works to the pNext stuff and we want to make sure we are not too strict now with how maintenance5 was suppose to work

The tests are `dEQP-VK.api.maintenance5.format.*`

PR #8514, opened 2024-09-09
Fix maintenance5 PhysicalDevice checks

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8367

This passes through the `PhysicalDevice` handle when we validate enums/flags


ziga-lunarg (approved):
Fixes the false positives in CTS

Issue #8476, opened 2024-08-29
vkSignalSemaphore should not initiate forward progress in queue thread


artem-lunarg:
vkSignalSemaphore state tracking  uses vvl::Semaphore::Retire() which marks resources as not in use. The expected behavior that signal should be registered without initiation of forward progress.

```
TEST_F(PositiveSyncObject, NotVeryReliableTestButDemonstratesTheIssue) {
    SetTargetApiVersion(VK_API_VERSION_1_2);
    AddRequiredFeature(vkt::Feature::timelineSemaphore);
    RETURN_IF_SKIP(Init());

    vkt::Buffer buffer_a(*m_device, 256, VK_BUFFER_USAGE_TRANSFER_SRC_BIT);
    vkt::Buffer buffer_b(*m_device, 256, VK_BUFFER_USAGE_TRANSFER_DST_BIT);
    m_command_buffer.begin();
    m_command_buffer.Copy(buffer_a, buffer_b);
    m_command_buffer.end();

    vkt::Semaphore semaphore(*m_device, VK_SEMAPHORE_TYPE_TIMELINE);
    m_default_queue->SubmitWithTimelineSemaphore(m_command_buffer, vkt::wait, semaphore, 1);

    // This starts retiring queue operations similar to vkQueueWaitIdle, except there is no wait at the end
    semaphore.Signal(1);
    
    // Magic line. In most cases this will allow the test to pass which is not correct, resources should be 
    // considered in-use until we use some form of host wait (vkWaitSemaphores, vkDeviceWaitIdle, etc.).
    std::this_thread::sleep_for(std::chrono::milliseconds{100});
}
```

Issue #8370, opened 2024-08-09
Incorrect resource in-use validation when timeline semaphore is used


artem-lunarg:
Something I noticed some time ago but was lazy to report. The test comments explain the issue. 

```
TEST_F(PositiveSyncObject, Test) {
    SetTargetApiVersion(VK_API_VERSION_1_2);
    AddRequiredFeature(vkt::Feature::timelineSemaphore);
    RETURN_IF_SKIP(Init());
    if (!m_second_queue) {
        GTEST_SKIP() << "Test requires two queues";
    }

    vkt::Buffer buffer_a(*m_device, 256, VK_BUFFER_USAGE_TRANSFER_SRC_BIT);
    vkt::Buffer buffer_b(*m_device, 256, VK_BUFFER_USAGE_TRANSFER_DST_BIT);
    m_command_buffer.begin();
    m_command_buffer.Copy(buffer_a, buffer_b);
    m_command_buffer.end();

    vkt::Semaphore semaphore(*m_device, VK_SEMAPHORE_TYPE_TIMELINE);

    // A) Submitting like this will corectly generate in-use error since
    // we dont wait for m_default_queue before starting destroying resources.
    // m_default_queue->Submit(m_command_buffer);

    // B) This example shows the issue. We initiate validation forward progress on the
    // m_second_queue by calling Wait() and this will properly mark m_second_queue related
    // resources as not in use. The problem that semaphore signal automatically initiates
    // validation forward progress on m_default_queue. m_default_queue will mark all resources
    // as not in-use, even this queue is not synchronized with Wait(). This tests finishes
    // successfully, but it should cause in-use error.
    m_default_queue->SubmitWithTimelineSemaphore(m_command_buffer, vkt::wait, semaphore, 1);
    m_second_queue->SubmitWithTimelineSemaphore(vkt::no_cmd, vkt::signal, semaphore, 1);
    m_second_queue->Wait();
}
```

Issue #7941, opened 2024-05-02
Deadlock when waiting on the queue


artem-lunarg:
This test reproduces the issue:
```
TEST_F(PositiveSyncObject, TwoQueueSubmitsWaitBeforeSignal) {
    AddRequiredExtensions(VK_KHR_TIMELINE_SEMAPHORE_EXTENSION_NAME);
    AddRequiredFeature(vkt::Feature::timelineSemaphore);
    all_queue_count_ = true;
    RETURN_IF_SKIP(Init());

    if (!m_second_queue) {
        GTEST_SKIP() << "2 queues are needed";
    }
    vkt::Semaphore semaphore(*m_device, VK_SEMAPHORE_TYPE_TIMELINE_KHR);
    m_default_queue->SubmitWithTimelineSemaphore(vkt::no_cmd, vkt::wait, semaphore, 1);
    m_second_queue->SubmitWithTimelineSemaphore(vkt::no_cmd, vkt::signal, semaphore, 2);
    m_default_queue->Wait();
}
```
The scenario is almost identical to https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7904 but waiting on a single queue instead of `vkDeviceWaitIdle`. 

This use case looks more tricky comparing to a deadlock with `vkDeviceWaitIdle`. In the latter case it was enough to `Notify()` all the queues. In the above test it's not enough to notify `m_default_queue`, because it depends on `m_second_queue`, and somehow we need to initiate forward progress on the second queue too. But it's also incorrect to initiate forward progress on all the queues.  In the general case this can retire the queues that are not in the dependency chain and can legally be pending after the wait.

artem-lunarg:
TODO: negative tests that adds the third queue and it does not interact with the above two queues and is still pending after Wait() so we have troubles to delete some resources used by the third queue.

Issue #8461, opened 2024-08-26
Bugs with CTS synchronization tests


ziga-lunarg:
Running timeline semaphore tests in CTS reports `UNASSIGNED-VkSemaphore-state-timeout` and the message says `This is most likely a validation bug`.

One such test for example is `dEQP-VK.synchronization.timeline_semaphore.device_host.write_copy_buffer_read_copy_buffer.buffer_16384`

And the full message is:

> UNASSIGNED-VkSemaphore-state-timeout(ERROR / SPEC): msgNum: 1474714163 - Validation Error: [ UNASSIGNED-VkSemaphore-state-timeout ] Object 0: handle = 0x40b43c0000000049, type = VK_OBJECT_TYPE_SEMAPHORE; | MessageID = 0x57e65a33 | vkWaitSemaphores(): pWaitInfo->pValues[0] Timeout waiting for timeline semaphore state to update. This is most likely a validation bug. completed_.payload=0 wait_payload=141
>     Objects: 1
>         [0] 0x40b43c0000000049, type: 5, name: NULL

There are other VUIDs that follow, but I assume they are caused by this first one.

jeremyg-lunarg:
I think this might be a tests where many many semaphores a created and then it takes a very long time for them to be signaled.  Possibly the timer for this error needs to be made longer.

artem-lunarg:
It's a regression after 882f09f1e8751a8738ae1c3dd6bf0e8ceb5898bc

PR #8519, opened 2024-09-09
layers: Update semaphore state tracking

artem-lunarg:
Broadly this fixes the following categories of issues:
* semaphore signal on one queue could initiate forward progress on a different queue and mark resources there as not in-use in the scenarios where the second queue is not synchronized with a host in any way.
* various combinations of timeline semaphore host signal/wait led to deadlock (timeout and internal vvl error).
* issues with timeline synchronization when signal and wait values are not equal (larger signals also resolve the waits).

The main observation that notifying timeline's timepoint should notify only resolving signals but not the waits (which can trigger forward progress on other queues even without host wait). The waits are naturally processed when forward progress on particular queue goes over submissions one by one.

Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7941
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8370
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8476

### DIFF
--- a/tests/unit/sync_object.cpp
+++ b/tests/unit/sync_object.cpp
@@ -12,6 +12,7 @@
  *     http://www.apache.org/licenses/LICENSE-2.0
  */
 
+#include <thread>
 #include "utils/cast_utils.h"
 #include "../framework/layer_validation_tests.h"
 #include "../framework/barrier_queue_family.h"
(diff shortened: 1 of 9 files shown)


tests/unit/sync_object_positive.cpp
```diff
@@ -2086,3 +2089,184 @@ TEST_F(PositiveSyncObject, GetCounterValueOfExportedSemaphore2) {
...
+TEST_F(PositiveSyncObject, TimelineHostWaitThenSubmitSignal) {
+    TEST_DESCRIPTION("Wait on the host then submit signal");
+    SetTargetApiVersion(VK_API_VERSION_1_2);
+    AddRequiredFeature(vkt::Feature::timelineSemaphore);
+    RETURN_IF_SKIP(Init());
+
+    vkt::Semaphore semaphore(*m_device, VK_SEMAPHORE_TYPE_TIMELINE_KHR);
+    auto thread = [&semaphore]() { semaphore.Wait(1, vvl::kU64Max); };
+    std::thread t(thread);
+    std::this_thread::sleep_for(std::chrono::milliseconds{50});
```
artem-lunarg, line 2102: This delays can't cause false-positives. A simple solution to have high enough probability that wait started before signal, but in the rare case of computer/scheduling glitches the wait and signal will be swapped and the test will test boring scenario of wait after signal.

artem-lunarg:
So far tested this only with VVL tests, tomorrow will test some games/apps/cts/angle.

spencer-lunarg (approved):

tests/unit/sync_object.cpp
```diff
@@ -3674,3 +3675,62 @@ TEST_F(NegativeSyncObject, ImageBarrierStageNotSupportedByQueue) {
...
+    AddRequiredFeature(vkt::Feature::timelineSemaphore);
+    RETURN_IF_SKIP(Init());
+
+    vkt::Semaphore semaphore(*m_device, VK_SEMAPHORE_TYPE_TIMELINE);
+    VkSemaphore handle = semaphore.handle();
+
+    m_default_queue->SubmitWithTimelineSemaphore(vkt::no_cmd, vkt::wait, semaphore, 1);
+    semaphore.Signal(1);
+
+    // This wait should NOT help the above submit to retire.
```
jeremyg-lunarg, line 3691: do you mean the Signal(1) shouldn't help the submit retire?

tests/unit/sync_object_positive.cpp
jeremyg-lunarg: this should be a comment in the test

tests/unit/sync_object.cpp
artem-lunarg: Yeah, need to rewrite. The idea if there is a regression and Signal() incorrectly initiates retirement then the test fails because DestroySemaphore runs without error. Delay helps this regression, so the queue thread has more time to mark objects as not in-use. There is no guarantee, but few ms is enough in most cases.  In case of bad scheduling that queue thread did not finish in this amount of time then test will just pass, so in rare cases could be false negatives, but not false-positives. 

That's the idea in most tests. Technically the test will run correctly if delays is removed, but this increases the rate that the test can pass when there is regression.

jeremyg-lunarg: > That's the idea in most tests. Technically the test will run correctly if delays is removed, but this increases the rate that the test can pass when there is regression.

That idea is very sound. I was just tripped up by the comment.

artem-lunarg:
CS2/Dota tests, quake3, angle end2end tests with `--gtest_filter=*Sync*` runs fine.
CTS with `-n=dEQP-VK.synchronization.timeline_semaphore*` triggers existing https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8461, will check it later.
Doom capture reports new VUID-vkAcquireNextImageKHR-semaphore-01779 `Semaphore must not have any pending operations` , this change can expose such issues but need to investigate.

tests/unit/sync_object_positive.cpp
artem-lunarg: done

tests/unit/sync_object.cpp
artem-lunarg: done

tests/unit/sync_object_positive.cpp
```diff
@@ -2086,3 +2089,184 @@ TEST_F(PositiveSyncObject, GetCounterValueOfExportedSemaphore2) {
...
+
+TEST_F(PositiveSyncObject, TimelineHostSignalThenHostWait) {
+    TEST_DESCRIPTION("Signal on the host then wait on the host");
+    SetTargetApiVersion(VK_API_VERSION_1_2);
+    AddRequiredFeature(vkt::Feature::timelineSemaphore);
+    RETURN_IF_SKIP(Init());
+
+    vkt::Semaphore semaphore(*m_device, VK_SEMAPHORE_TYPE_TIMELINE_KHR);
+    semaphore.Signal(1);
+    semaphore.Wait(1, vvl::kU64Max);
```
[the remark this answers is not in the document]
artem-lunarg: done

artem-lunarg:
I just got a blue screen (windows, nvidia) when run *SyncObject* tests, and there was another one few days ago. This needs more testing.

artem-lunarg:
Investigated new VUID-vkAcquireNextImageKHR-semaphore-01779 from internal capture (doometernal_370_frames_20210224T125732.gfxr). The conclusion it is a valid error.

Attached is submits/acquire/present from the first two frames that led to error in frame 2 during vkAcquireNextImageKHR and explanation that causes the problem.
[doom-first-frames-sync.txt](https://github.com/user-attachments/files/16962655/doom-first-frames-sync.txt)

jeremyg-lunarg (approved):

Issue #8523, opened 2024-09-10
Incorrect VUID-vkCmdDrawIndexed-Input-08734 VVL error


HuYuxin:
**Environment:**
 - OS: Linux
 - GPU and driver version: NVIDIA, Driver Version: 535.183.01. Vulkan Instance Version: 1.3.283
 - SDK or header version if building from repo: 9abfdd289
 - Options enabled (synchronization, best practices, etc.):

**Describe the Issue**

We found that after applying this change  https://chromium.googlesource.com/external/github.com/KhronosGroup/Vulkan-ValidationLayers/+/9abfdd2897a8a130524bb31caf934440804068d2, some ANGLE tests start to hit VVL error VUID-vkCmdDrawIndexed-Input-08734:

```
vkCmdSetVertexInputEXT set pVertexAttributeDescriptions[1](binding 1, location 1) with format VK_FORMAT_R8G8B8A8_UINT but the vertex shader input is numeric type vec4 of float32. 
```

**Expected behavior**

In the renderdoc capture, here are the shader vertex input format:

```
Location 0, Binding 0, R16G16B16A16_SNORM
Location 1, Binding 1, R16G16B16_FLOAT
Location 2, Binding 2, R16G16B16A16_FLOAT
Location 3, Binding 3, R16G16_SNORM
Location 4, Binding 4, R16G16_SNORM
Location 5, Binding 5, R8G8B8A8_UINT
Location 6, Binding 6, R8G8B8A8_UNORM
```

And here are the pVertexAttributeDescriptions values passed o vkCmdSetVertexInputEXT:

```
Location 0, Binding 0, VK_FORMAT_R16G16B16A16_SNORM
Location 1, Binding 1, VK_FORMAT_R16G16B16_SFLOAT
Location 2, Binding 2, VK_FORMAT_R16G16B16A16_SFLOAT
Location 3, Binding 3, VK_FORMAT_R16G16_SNORM
Location 4, Binding 4, VK_FORMAT_R16G16_SNORM
Location 5, Binding 5, VK_FORMAT_R8G8B8A8_UINT
Location 6, Binding 6, VK_FORMAT_R8G8B8A8_UNORM
```

The formats at Location 1 binding 1 (R16G16B16_FLOAT and VK_FORMAT_R16G16B16_SFLOAT) seem to match, and they don't match with any of the formats mentioned in VVL error (VK_FORMAT_R8G8B8A8_UINT, vec4 of float32).

[...]

PR #8526, opened 2024-09-12
layers: Fix multiple calls to CmdSetVertexInputEXT

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8523

When we called `vkCmdSetVertexInputEXT` twice in the same command buffer, we were not actually updating the values since `emplace()` does not override the value in the `unordered_map`

### DIFF
--- a/layers/state_tracker/state_tracker.cpp
+++ b/layers/state_tracker/state_tracker.cpp
@@ -5255,14 +5255,14 @@ void ValidationStateTracker::PostCallRecordCmdSetVertexInputEXT(
     }
     auto &vertex_bindings = cb_state->dynamic_state_value.vertex_bindings;
     for (const auto [i, bd] : vvl::enumerate(pVertexBindingDescriptions, vertexBindingDescriptionCount)) {
-        vertex_bindings.emplace(bd->binding, VertexBindingState(i, bd));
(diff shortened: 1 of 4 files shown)


layers/state_tracker/state_tracker.cpp
```diff
@@ -5255,14 +5255,24 @@ void ValidationStateTracker::PostCallRecordCmdSetVertexInputEXT(
     }
     auto &vertex_bindings = cb_state->dynamic_state_value.vertex_bindings;
     for (const auto [i, bd] : vvl::enumerate(pVertexBindingDescriptions, vertexBindingDescriptionCount)) {
-        vertex_bindings.emplace(bd->binding, VertexBindingState(i, bd));
+        // Since we don't want a default constructor for VertexBindingState we need to search
+        if (vertex_bindings.find(bd->binding) == vertex_bindings.end()) {
```
artem-lunarg, line 5259: Can `insert_or_assign` be used here?

artem-lunarg (approved):
LGTM. Suggestion about map insertion.

layers/state_tracker/state_tracker.cpp
spencer-lunarg: ......I literally spent like 10 minutes trying to find `insert_or_assign` and couldn't... ya, 100% lets do that

PR #8527, opened 2024-09-12
tests: Add new test helpers from timeline branch

artem-lunarg:
Some general helpers to minimize diff.

### DIFF
--- a/layers/sync/sync_stats.cpp
+++ b/layers/sync/sync_stats.cpp
@@ -26,8 +26,16 @@
 namespace syncval_stats {
 
 void Value32::Update(uint32_t new_value) { u32.store(new_value); }
-uint32_t Value32::Add(uint32_t n) { return u32.fetch_add(n); }
-uint32_t Value32::Sub(uint32_t n) { return u32.fetch_sub(n); }
+
+uint32_t Value32::Add(uint32_t n) {
+    // fetch_add returns initial value (before increment), add one to get the new value
+    return u32.fetch_add(n) + 1;
+}
+
+uint32_t Value32::Sub(uint32_t n) {
+    // fetch_sub returns initial value (before the decrement), subtract one to get the new value
+    return u32.fetch_sub(n) - 1;
+}
 
 void ValueMax32::Update(uint32_t new_value) {
     value.Update(new_value);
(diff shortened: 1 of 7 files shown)


layers/sync/sync_stats.cpp
```diff
@@ -26,8 +26,8 @@
 namespace syncval_stats {
 
 void Value32::Update(uint32_t new_value) { u32.store(new_value); }
-uint32_t Value32::Add(uint32_t n) { return u32.fetch_add(n); }
-uint32_t Value32::Sub(uint32_t n) { return u32.fetch_sub(n); }
+uint32_t Value32::Add(uint32_t n) { return u32.fetch_add(n) + 1; }
+uint32_t Value32::Sub(uint32_t n) { return u32.fetch_sub(n) - 1; }
```
jeremyg-lunarg, line 30: probably should have a comment about what the return values are for these, just to save people from having to go look up fetch_add & fetch_sub.

tests/framework/binding.h
```diff
@@ -420,6 +420,7 @@ class Semaphore : public internal::NonDispHandle<VkSemaphore> {
     VkResult WaitKHR(uint64_t value, uint64_t timeout);
     VkResult Signal(uint64_t value);
     VkResult SignalKHR(uint64_t value);
+    uint64_t GetCounterValue(bool use_khr = false) const;
```
jeremyg-lunarg, line 423: could `use_khr_` be part of the Semaphore object? Then you could get rid of the KHR versions of Wait and Signal as well.

artem-lunarg: It's a good idea. I'd postpone this to a separate PR, currently we have other styles here like KHR in the signature name, and also submit helpers. Have a plan to update submit helpers too so can do as part of that task.

layers/sync/sync_stats.cpp
artem-lunarg: done

tests/framework/binding.h
jeremyg-lunarg: Possibly we need an even more general solution. A similar question just popped up in internal chat. And I could use smarter dispatch in both CDL and VEL.

artem-lunarg: @jeremyg-lunarg could you approve this nevertheless, I'm trying to move to the main branch what is not essential to timeline validation, for simplicity reasons.

jeremyg-lunarg (approved):

tests/framework/binding.h
jeremyg-lunarg: done

Issue #8420, opened 2024-08-20
Misc error message improvements


artem-lunarg:
1. Aspect mask is printed as a number instead of readable flag names. Search this pattern for places to fix: `aspectMask 0x%x`

2.  When I run validation through vkconfig in the end of the message I get the list of objects.  The suggestion is to replace type number with a readable name, assuming that it is VVL responsible for this part.
```
   Objects: 2
        [0] 0x242784aee70, type: 6, name: NULL
        [1] 0xf443490000000006, type: 9, name: NULL
```

ziga-lunarg:
`VUID-VkMemoryAllocateFlagsInfo-flags-parameter` for example is similar, the message only says:

> vkAllocateMemory(): pAllocateInfo->pNext<VkMemoryAllocateFlagsInfo>.flags (0x2) has VkMemoryAllocateFlagBits values that requires the extensions VK_KHR_buffer_device_address.

PR #8534, opened 2024-09-12
layers: Improve error message with flags

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8420

>  vkCreateDescriptorSetLayout(): pCreateInfo->flags has VkDescriptorSetLayoutCreateFlagBits values (VK_DESCRIPTOR_SET_LAYOUT_CREATE_UPDATE_AFTER_BIND_POOL_BIT) that requires the extensions VK_EXT_descriptor_indexing.


layers/best_practices/bp_render_pass.cpp
```diff
@@ -825,11 +825,12 @@ bool BestPractices::ValidateCmdEndRenderPass(VkCommandBuffer commandBuffer, cons
             if (untouched_aspects) {
                 skip |= LogPerformanceWarning(
                     "BestPractices-vkCmdEndRenderPass-redundant-attachment-on-tile", commandBuffer, loc,
-                    "%s %s: Render pass was ended, but attachment #%u (format: %u, untouched aspects 0x%x) "
+                    "%s %s: Render pass was ended, but attachment #%u (format: %u, untouched aspects %s) "
```
ziga-lunarg, line 828: Should these `u` also be replaced with `PRIu32`?

spencer-lunarg: yess... but we need to do that everywhere... I have a todo to just make a large find-and-replace

ziga-lunarg (approved):

PR #8536, opened 2024-09-13
Improve enum extension error message

spencer-lunarg:
similar to https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8534 but for the enums

old message

> [ VUID-VkImageCreateInfo-format-parameter ] | vkCreateImage(): pCreateInfo->format (1000340001) requires the extensions VK_EXT_4444_formats.

now

> [ VUID-VkImageCreateInfo-format-parameter ] | vkCreateImage(): pCreateInfo->format (VK_FORMAT_A4B4G4R4_UNORM_PACK16) requires the extensions VK_EXT_4444_formats.

### DIFF
--- a/layers/stateless/stateless_validation.h
+++ b/layers/stateless/stateless_validation.h
@@ -407,7 +407,8 @@ class StatelessValidation : public ValidationObject {
         } else if (result == ValidValue::NoExtension && device != VK_NULL_HANDLE) {
             // If called from an instance function, there is no device to base extension support off of
             auto extensions = GetEnumExtensions(value);
-            skip |= LogError(vuid, device, loc, "(%" PRIu32 ") requires the extensions %s.", value, String(extensions).c_str());
+            skip |=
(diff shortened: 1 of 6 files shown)


layers/stateless/stateless_validation.h
```diff
@@ -489,6 +490,8 @@ class StatelessValidation : public ValidationObject {
     ValidValue IsValidEnumValue(T value) const;
     template <typename T>
     vvl::Extensions GetEnumExtensions(T value) const;
+    template <typename T>
+    std::string DescribeEnum(T value) const;
```
arno-lunarg, line 494:
```suggestion
    const char * DescribeEnum(T value) const;
```

arno-lunarg (approved):
Just use `const char *`!

layers/stateless/stateless_validation.h
spencer-lunarg: ya, idk why I was in a string mood last night

PR #8546, opened 2024-09-16
layers: Remove workaround for handling missing signals

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8461

Discussed with ~@jeremy-lunarg~ @jeremyg-lunarg  reverted commit.  PreRecord guarantees that any dependent PostRecord will see registered signals/waits in semaphore state object. There is a code that relies on this (when retiring non-external semaphore wait, there should be a signal).

### DIFF
--- a/layers/state_tracker/semaphore_state.cpp
+++ b/layers/state_tracker/semaphore_state.cpp
@@ -59,22 +59,6 @@ void vvl::Semaphore::EnqueueSignal(const SubmissionReference &signal_submit, uin
     assert(timeline_.find(payload) == timeline_.end() || !timeline_.find(payload)->second.signal_submit.has_value());
 
     timeline_[payload].signal_submit.emplace(signal_submit);
-
-    // WORKAROUND for situation when WaitSemaphores wait is over but it can't find
-    // registered signal because QueueSubmit PostRecord has not started yet.
-    // For in-use tracking the solution is to register signal in PreRecord,
-    // so subsequent PostRecord calls will have access to the expected state
-    // (One of the PostRecords will retire the signal, but then the others will see
-    // this as part of semaphore completed state, for example, other retire calls
-    // will return early.
-    if (type == VK_SEMAPHORE_TYPE_TIMELINE) {
-        for (auto &[p, t] : timeline_) {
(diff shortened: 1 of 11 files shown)


layers/state_tracker/semaphore_state.cpp
```diff
@@ -224,26 +208,46 @@ bool vvl::Semaphore::CanRetireBinaryWait(TimePoint &timepoint) const {
...
-            // Initiate forward progress on signaling queue and ask the caller to wait.
-            if (t.signal_submit->queue != current_queue) {
-                t.Notify();
-                return false;
-            }
+
+    // In the correct program the resolving signal is the next signal on the timeline,
+    // otherwise this violates the rule of strictly increasing signal values.
+    const TimePoint *timepoint = nullptr;
+    for (auto it = timeline_.find(payload); it != timeline_.end(); ++it) {
```
artem-lunarg, line 215: Just realized this loop is basically what we had before, just with a placeholder to add error reporting later (probably not for the ongoing SDK). The main difference is removed workaround.

artem-lunarg:
Investigating crashes on CI machines.

artem-lunarg:
The problem with CI crashes was that last change I did before opening this PR to cleanup things a bit...

tests/unit/sync_object_positive.cpp
```diff
@@ -1313,6 +1313,10 @@ TEST_F(PositiveSyncObject, FenceSemThreadRace) {
     AddRequiredFeature(vkt::Feature::timelineSemaphore);
     RETURN_IF_SKIP(Init());
 
+    if (IsPlatformMockICD()) {
+        GTEST_SKIP() << "Test not supported by MockICD (WaitSemaphores)";
```
jeremyg-lunarg, line 1317: note: I have code in the CDL test framework that implements semaphores for the mock ICD. Could be a reason for us to re-converge them.

jeremyg-lunarg (approved):
LGTM.

tests/unit/sync_object_positive.cpp
artem-lunarg: Okay. Are you implementing mock synchronization for mock ICD?

PR #8547, opened 2024-09-16
layers: Improve bindless descriptor error messages

spencer-lunarg:
Some improvements to the error message logic around bindless descriptors


layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -261,13 +261,6 @@ bool LogMessageInstBindlessDescriptor(Validator &gpuav, const uint32_t *error_re
             out_vuid_msg = "UNASSIGNED-Descriptor destroyed";
             error_found = true;
         } break;
-        case kErrorSubCodeBindlessDescriptorNullPointer: {
```
spencer-lunarg, line 264: we removed this check already

layers/gpu/shaders/instrumentation/bindless_descriptor.comp
```diff
@@ -80,10 +80,6 @@ bool inst_bindless_descriptor(const uint inst_num, const uvec4 stage_info, const
     uint param_1 = 0u;
     do {
         DescriptorLayoutData layout_data = bindless_state_buffer.desc_sets[desc_set].layout_data;
-        if (binding >= layout_data.num_bindings) {
```
spencer-lunarg, line 83: I wrote a test, there is no way to hit this unless we (VVL) mess up, so removing because we don't have luxary to keep VVL asserts here when the compiler bloat is already too large

(added a test below to confirm this as well)

layers/gpu/shaders/instrumentation/bindless_descriptor.comp
```diff
@@ -100,8 +96,6 @@ bool inst_bindless_descriptor(const uint inst_num, const uvec4 stage_info, const
         uint desc_type = (in_data.data[state_index].x & ~kDebugInputBindlessSkipId) >> kDescBitShift;
         if (desc_id == 0u) {
             error = kErrorSubCodeBindlessDescriptorUninit;
-            param_0 = binding_state.y;
```
spencer-lunarg, line 103: we are not printing these values (they don't provide any information worth printing anyways) so removing to reduce compiled size

tests/unit/gpu_av_descriptor_indexing.cpp
```diff
@@ -2503,4 +2503,68 @@ TEST_F(NegativeGpuAVDescriptorIndexing, PartialBoundDescriptorSSBO) {
...
+    pipe.cp_ci_.layout = pipeline_layout.handle();
+    pipe.CreateComputePipeline();
+
+    m_commandBuffer->begin();
+    vk::CmdBindPipeline(m_commandBuffer->handle(), VK_PIPELINE_BIND_POINT_COMPUTE, pipe.Handle());
+    vk::CmdBindDescriptorSets(m_commandBuffer->handle(), VK_PIPELINE_BIND_POINT_COMPUTE, pipeline_layout.handle(), 0, 1,
+                              &descriptor_set.set_, 0, nullptr);
+    vk::CmdDispatch(m_commandBuffer->handle(), 1, 1, 1);
+    m_commandBuffer->end();
+    m_errorMonitor->SetDesiredErrorRegex("VUID-vkCmdDispatch-None-08114", "Descriptor index 1 is uninitialized");
```
arno-lunarg, line 2566: Include `set` and `binding` in error message, and don't use SetDesiredError`Regex`

arno-lunarg (approved):

Issue #8477, opened 2024-08-29
Setting report_flags through an env variable or VK_EXT_layer_settings doesn't work


kanashimia:
**Environment:**
 - OS: NixOS 24.11
 - SDK or header version: 1.3.283.0
 - Options enabled (synchronization, best practices, etc.): All of them

**Describe the Issue**

VVL docs list that you can use `VK_KHRONOS_VALIDATION_REPORT_FLAGS` env variable to set the report flags:
https://vulkan.lunarg.com/doc/sdk/1.3.283.0/windows/khronos_validation_layer.html

But you can't just enable report_flags for example like this:
```bash
export VK_LOADER_LAYERS_ENABLE=VK_LAYER_KHRONOS_validation
export VK_KHRONOS_VALIDATION_REPORT_FLAGS=perf
```

You will see nothing, as if that env variable doesn't exist.

It wont work if you set it through the VK_EXT_layer_settings API too, even though examples clearly show that you should be able to.
https://vulkan.lunarg.com/doc/sdk/1.3.283.0/windows/layer_configuration.html

```c
// I heard you like macros? I wrote some macros for you:

#define ARRAY_LEN(arr) (sizeof(arr) / sizeof(*arr))
#define VK_ARRAY(...) ARRAY_LEN((__VA_ARGS__)), (__VA_ARGS__)

#define VK_DBG_TYPE_MAP(val) _Generic((val), \
    VkBool32: VK_LAYER_SETTING_TYPE_BOOL32_EXT, \
    str: VK_LAYER_SETTING_TYPE_STRING_EXT )

#define VK_DBG_OPT(name, ...) { \
    .pLayerName = "VK_LAYER_KHRONOS_validation", \
    .pSettingName = name, \
    .type = VK_DBG_TYPE_MAP(*(__VA_ARGS__)), \
    .valueCount = ARRAY_LEN((__VA_ARGS__)), (__VA_ARGS__) }

auto layer_settings_info = (VkLayerSettingsCreateInfoEXT){
    .sType = VK_STRUCTURE_TYPE_LAYER_SETTINGS_CREATE_INFO_EXT,
    .settingCount = VK_ARRAY((VkLayerSettingEXT[]){
        VK_DBG_OPT("validate_core", (VkBool32[]){ true }),
        VK_DBG_OPT("validate_sync", (VkBool32[]){ true }),
        // VK_DBG_OPT("thread_safety", (VkBool32[]){ true }),
[...]

PR #8549, opened 2024-09-17
layers: Move all settings to VK_EXT_layer_settings

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8477

I have wanted to do this for a long time `VK_LAYER_LOG_FILENAME`, `VK_LAYER_DEBUG_ACTION`, and `VK_LAYER_REPORT_FLAG` could never be set by environment variables and used old code that was solved by using `vkuCreateLayerSettingSet`via `VK_EXT_layer_setting`

This moves the code over, as well as adds some more useful info when setting non-default settings


layers/error_message/logging.cpp
```diff
@@ -676,6 +676,49 @@ VKAPI_ATTR VkBool32 VKAPI_CALL MessengerBreakCallback([[maybe_unused]] VkDebugUt
     return false;
 }
 
+static void PrintMessageSeverity(VkFlags vk_flags, char *msg_flags) {
```
spencer-lunarg, line 679: just moving here as only place it is used

layers/vk_layer_config.h
```diff
@@ -16,64 +16,9 @@
  **************************************************************************/
```
spencer-lunarg, line 16: this file basically has just become a way to call `GetEnvironment`

might move this logic to another file, but for now going to just leave here

layers/vk_layer_config.cpp
```diff
@@ -250,197 +127,6 @@ static inline bool IsHighIntegrity() {
 }
 #endif
 
-string ConfigFile::FindSettings() {
```
spencer-lunarg, line 253: all this logic was replaced with `VK_EXT_layer_settings`, we were duplicating its logic here

layers/layer_options.cpp
```diff
@@ -508,6 +520,167 @@ static std::string Merge(const std::vector<std::string> &strings) {
...
+    }
+    if (report_flags & kVerboseBit) {
+        dbg_create_info.messageSeverity |= VK_DEBUG_UTILS_MESSAGE_SEVERITY_VERBOSE_BIT_EXT;
+    }
+
+    VkDebugUtilsMessengerEXT messenger = VK_NULL_HANDLE;
+    if (debug_action & VK_DBG_LAYER_ACTION_LOG_MSG) {
+        FILE *log_output = GetLayerLogOutput(log_filename.c_str());
+        if (log_output != stdout) {
+            printf("Validation Layer Info - Logging validation error to %s\n", log_filename.c_str());
```
spencer-lunarg, line 657: this single handly is so nice, i have sometimes had issues trying to log to a file and having confirmation that it is hooked up is awesome

tests/framework/test_framework.cpp
```diff
@@ -16,8 +16,8 @@
  * limitations under the License.
  */
 
-#include "test_framework.h"
```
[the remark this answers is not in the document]
jeremyg-lunarg: agreed.  Is vk_layer_config.h needed in test_framework.h?  test_framework.h should stand on its own and having it first in its .cpp file is a good way to ensure that.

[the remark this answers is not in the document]
spencer-lunarg: will undo this, not sure what happen here

scripts/generators/layer_chassis_generator.py
```diff
@@ -982,35 +981,11 @@ def generateSource(self):
...
-                        default:
-                            settings_status.append("default location (current working directory).");
-                            break;
-                    }
-                }
-
                 Location loc(vvl::Func::vkCreateInstance);
                 // Output layer status information message
                 context->LogInfo("WARNING-CreateInstance-status-message", context->instance, loc,
-                    "Khronos Validation Layer Active:\\n    Settings File: %s\\n    Current Enables: %s.\\n    Current Disables: %s.\\n",
```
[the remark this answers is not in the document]
spencer-lunarg: yes, before we (VVL) would detect where the setting file came from (vkconfig or not)

but now that is all done is the layer settings

... honestly this whole `LogInfo` seems like it should just be replaced with a system where we go "here are all the settings set" verbose mode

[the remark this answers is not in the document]
spencer-lunarg: https://github.com/KhronosGroup/Vulkan-Utility-Libraries/issues/188

docs/settings.md
```diff
@@ -0,0 +1,78 @@
+# Settings
+
+There are many settings that can be set for Validation layers. This is a brief overview how to use them.
+
+There are 4 ways to set settings: `vkconfig`, `application defined`, `vk_layer_settings.txt`, `environment variables`
+
+## VkConfig
+
+We suggest people to use [VkConfig](https://www.lunarg.com/introducing-the-new-vulkan-configurator-vkconfig/).
```
ziga-lunarg, line 9: Missing word between `suggest` and `people`? I would just say `We suggest using`

spencer-lunarg: will change to `We suggest that people use`

docs/settings.md
```diff
@@ -0,0 +1,78 @@
+# Settings
+
+There are many settings that can be set for Validation layers. This is a brief overview how to use them.
+
+There are 4 ways to set settings: `vkconfig`, `application defined`, `vk_layer_settings.txt`, `environment variables`
```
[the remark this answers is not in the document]
filnet: 4 => four ?

docs/settings.md
```diff
@@ -0,0 +1,78 @@
...
+const VkLayerSettingEXT layer_setting = {"VK_LAYER_KHRONOS_validation", "printf_verbose", VK_LAYER_SETTING_TYPE_BOOL32_EXT, 1, &verbose_value};
+VkLayerSettingsCreateInfoEXT layer_settings_create_info = {VK_STRUCTURE_TYPE_LAYER_SETTINGS_CREATE_INFO_EXT, nullptr, 1, &layer_setting};
+
+VkInstanceCreateInfo instance_ci = GetYourCreateInfo();
+instance_ci.pNext = &layer_settings_create_info;
+```
+
+## vk_layer_settings.txt
+
+There is info [elsewhere](https://vulkan.lunarg.com/doc/view/latest/windows/layer_configuration.html) to describe this file, but the short answer is to go
```
filnet, line 29: Looks like this sentence is unfinished.

docs/settings.md
```diff
@@ -0,0 +1,78 @@
...
+
+```
+# windows
+set VK_LAYER_SETTINGS_PATH=C:\path\to\vk_layer_settings.txt
+
+# linux
+export VK_LAYER_SETTINGS_PATH=/path/to/vk_layer_settings.txt
+```
+
+and it will set things for you there. We have a [default example](../layers/vk_layer_settings.txt) file you can start with
```
filnet, line 39: missing colon

docs/settings.md
```diff
@@ -0,0 +1,78 @@
...
+# Linux
+export VK_LAYER_MESSAGE_ID_FILTER=VUID-VkInstanceCreateInfo-pNext-pNext
+
+# Android
+adb setprop debug.vulkan.khronos_validation.message_id_filter=VUID-VkInstanceCreateInfo-pNext-pNext
+```
+
+## Finding available settings
+
+How I suggest finding them:
```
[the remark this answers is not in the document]
spencer-lunarg:
```suggestion
How we suggest finding them:
```

docs/settings.md
spencer-lunarg:
```suggestion
and it will set things for you in that file. We have a [default example](../layers/vk_layer_settings.txt) file you can start with.
```

docs/settings.md
spencer-lunarg:
```suggestion
There is info [elsewhere](https://vulkan.lunarg.com/doc/view/latest/windows/layer_configuration.html) to describe this file, but the short answer is to set the `VK_LAYER_SETTINGS_PATH` like the following:
```

PR #8550, opened 2024-09-17
gpu: Print internal shader ID in error message

spencer-lunarg:
now the error message will display

> Shader Module (4fac1c0000000032) (internal ID 1)

which is **SUPER** nice when you have large app and do a `VK_LAYER_GPUAV_DEBUG_DUMP_INSTRUMENTED_SHADERS` and need to line up which shader matches the error

### DIFF
--- a/layers/gpu/debug_printf/debug_printf.cpp
+++ b/layers/gpu/debug_printf/debug_printf.cpp
@@ -340,10 +340,10 @@ void Validator::AnalyzeAndGenerateMessage(VkCommandBuffer command_buffer, VkQueu
         }
 
         if (verbose) {
-            std::string debug_info_message =
-                GenerateDebugInfoMessage(command_buffer, instructions, debug_record->stage_id, debug_record->stage_info_0,
-                                         debug_record->stage_info_1, debug_record->stage_info_2, debug_record->instruction_position,
-                                         tracker_info, buffer_info.pipeline_bind_point, buffer_info.action_command_index);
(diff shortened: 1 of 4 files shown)


layers/gpu/debug_printf/debug_printf.cpp
```diff
@@ -340,10 +340,10 @@ void Validator::AnalyzeAndGenerateMessage(VkCommandBuffer command_buffer, VkQueu
         }
 
         if (verbose) {
-            std::string debug_info_message =
-                GenerateDebugInfoMessage(command_buffer, instructions, debug_record->stage_id, debug_record->stage_info_0,
-                                         debug_record->stage_info_1, debug_record->stage_info_2, debug_record->instruction_position,
-                                         tracker_info, buffer_info.pipeline_bind_point, buffer_info.action_command_index);
+            std::string debug_info_message = GenerateDebugInfoMessage(
```
arno-lunarg, line 343: This function call is getting longer than my shopping list, maybe worth having some `DebugInfo` struct and pass it? I feel that it is likely we are going to pass even more info in the future, having a struct makes adding stuff a tad easier

arno-lunarg (approved):

layers/gpu/debug_printf/debug_printf.cpp
spencer-lunarg: ya, part of this I think will shrink when debugprint and GPU-AV are merged as they are the only 2 who call it

PR #8552, opened 2024-09-17
gpu: Print debug info on one line

arno-lunarg:
To make lookups easier


spencer-lunarg (approved):
I think my original idea was it would have more info and you would want to separate them by line, but agree this ruins grepping at all

Issue #4047, opened 2022-04-21
missing sparse binding validation


zmike:
**Describe the situation in which you encountered the missing validation**
Running sparse binding glcts tests on zink

**Valid Usage IDs requested**

- [ ] VUID-VkSparseImageMemoryBind-memory-01104
- [x] VUID-VkSparseImageMemoryBind-memory-01105
- [x] VUID-VkSparseImageMemoryBind-subresource-01106
- [x] VUID-VkSparseImageMemoryBind-offset-01107
- [x] VUID-VkSparseImageMemoryBind-extent-01108
- [x] VUID-VkSparseImageMemoryBind-offset-01109
- [x] VUID-VkSparseImageMemoryBind-extent-01110
- [x] VUID-VkSparseImageMemoryBind-offset-01111
- [x] VUID-VkSparseImageMemoryBind-extent-01112
- [ ] ~~VUID-vkCreateBuffer-flags-00911~~ see https://gitlab.khronos.org/vulkan/vulkan/-/issues/2403

**Additional Context**
Probably some of the buffer ones too, but I haven't checked.

ncesario-lunarg:
@aitor-lunarg are some of your recent changes related to these?

aitor-lunarg:
No, but VUID-VkSparseImageMemoryBind-memory-01104 can easily be solved with the changes I will introduce on memory overlap checking. There's still work to be done regarding validation of sparse memory, which I was planning to take a look at after introducing the memory overlap checks (needed for sparse memory validation).

PR #8555, opened 2024-09-17
Improve coverage for sparse image binds

r-potter:


spencer-lunarg:
noting this will mark off a VU from https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4047

r-potter:
@spencer-lunarg Do you happen to know what implementation the address sanitizer builds run on? One of the added tests appears to be failing to emit an expected validation error, and one possible reason might be that the image memory alignment requirement is one byte (and so it's impossible to misalign the bind). I'd like to confirm that before making any changes

spencer-lunarg:
> Do you happen to know what implementation the address sanitizer builds run on?

that is just `TestICD` (aka `MockICD`) (since it is on github and there are not real GPUs there)

it is failing on the `Thread sanitizer` build as well, so it is not an issue with `address sanitizer` but rather Linux with TestICD

I see it is failing other real GPU as well that is not getting the check

My suggest would be to move the 2 new "sub-tests" to just be its own `TEST_F` for each variation, then it will be much easier to detect what is wrong with which one (currently I am not sure which of the 2 new sub-test is not hitting the VU, hence why I prefer smaller tests in general)

spencer-lunarg:
@r-potter I can try and took a look locally later today why both tests are failing

r-potter:
> @r-potter I can try and took a look locally later today why both tests are failing

Thanks @spencer-lunarg but hopefully no need. I just pushed what *should* be the solution. It appears to be minimum alignment of 1 making it impossible to violate the VU on those devices, so just added a skip in that case

r-potter:
@spencer-lunarg I guess I have to take that back. MockICD tests are resolved, but I have no way to verify what the issue with the AMD 6800 in CI is

spencer-lunarg:
>  but I have no way to verify what the issue with the AMD 6800 in CI is

We need to fix how the webhooks don't print the actual error, but it is

> Validation Error: [ VUID-vkAllocateMemory-deviceCoherentMemory-02790 ] | MessageID = 0x8830dc95 | vkAllocateMemory(): pAllocateInfo->memoryTypeIndex 4 includes the VK_MEMORY_PROPERTY_DEVICE_COHERENT_BIT_AMD memory property, but the deviceCoherentMemory feature is not enabled.

I am going to just mark this to skip on AMD Windows for now, I will soon have a better way to reproduce, otherwise you are just going to be sending guesses to get it to work and rather have someone with a local device reproduce it

Issue #7600, opened 2024-02-27
sync: False-positives in Granite samples when timeline semaphores are used


artem-lunarg:
The scope of this issue is restricted to two samples from the Granite renderer (https://github.com/Themaister/Granite):
`gltf-viewer-simple` and `latency-test`.

These samples are good examples of synchronization validation false positives due to missing timeline semaphore support. The above samples run error-free when binary semaphores are used.

`context.cpp`: `ext.vk12_features.timelineSemaphore = VK_FALSE` can be used to enable binary semaphores.

@HansKristian-Work for your visibility

devshgraphicsprogramming:
Btw I have a pretty nice repro of latched submits [multithreaded_histogram.zip](https://github.com/KhronosGroup/Vulkan-ValidationLayers/files/15149875/multithreaded_histogram.zip)

artem-lunarg:
thanks @devshgraphicsprogramming!

devshgraphicsprogramming:
Source is here btw: https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/blob/master/07_StagingAndMultipleQueues/main.cpp

devshgraphicsprogramming:
This code change silences the validation layer (if I treat every GPU as I'd treat Intel): https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/commit/4e48f0b27213fae6ce032bae4ff2ff0622921afc 

Basially my code is a repro for validation layers choking on timeline semaphores + out-of-order-submits to different queues

devshgraphicsprogramming:
Funny thing, renderdoc chokes on this usage as well, but it actually freezes up completely.

I guess its because its not been coded to be able to defer a capture until later after a submission.

imrodriguezro:
Hi!
out of curiosity, do you guys have a rough estimate of when these false-positives related to timeline semaphores will be fixed?

artem-lunarg:
Hi @imrodriguezro, actually we have some estimates. Syncval timelines semaphores support won't be available in the next SDK (later this month) but it will be in the SDK after that. The implementation should land in the main branch somewhere in August, so it will be possible to test this before the autumn SDK too.

imrodriguezro:
thanks for the info @artem-lunarg 👍

PR #8559, opened 2024-09-18
sync: Timeline semaphore support for syncval

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7600


artem-lunarg:
Removed malformed tests: violated rule increasing signal values. Caught by AMD driver which reported error from QueueSubmit.

tests/unit/sync_val.cpp
```diff
@@ -5709,63 +5709,6 @@ TEST_F(NegativeSyncVal, QSDebugRegion_Secondary) {
     m_default_queue->Wait();
 }
 
-// TODO: this test should be removed after timeline semaphore support is added to sync validation
```
spencer-lunarg, line 5712: good thing for `todo` :laughing:

spencer-lunarg (approved):

tests/unit/sync_val.cpp
artem-lunarg: as was designed it stopped working, that’s how I found it

artem-lunarg:
Found issue (unresolved batches are not getting removed) in the latest version of Granite renderer. Investigating.

artem-lunarg:
The issue is fixed.

artem-lunarg:
Fixed leak when resolved wait-before-signal did not retire older signals

artem-lunarg:
Cleanup and refactoring.

artem-lunarg:
Found one more potential issue when testing Granite renderer (unresolved batches are not getting resolved). Investigating.

artem-lunarg:
The issue is fixed (external semaphore case).

jeremyg-lunarg (approved):
LGTM. It would be good to have:
-@jzulauf-lunarg eyes on it
-A follow-on change to [docs/synchronization.md](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/docs/synchronization.md) at some point to describe how this works.

artem-lunarg:
Additional renderers/games used for testing:
* Granite renderer (gltf-viewer-simple, latency-test)
* Tiny Glade Demo 
* Return to Monkey Island (only debug vvl build)
* CS2/Dota2 (does not use timelines but stability/memory usage checks)
* Quake 3 Arena Kenny Edition (does not use timelines but stability/memory usage checks)

The feature documentation will be added later.

Issue #8192, opened 2024-06-25
Allow GPU-AV and DebugPrintF to be used at the same time


arno-lunarg:
Things to do:
- Remove `VUID-VkValidationFeaturesEXT-pEnabledValidationFeatures-02968` (no GPU AV and DebugPrintF at the same time)

spencer-lunarg:
so yes, I think we will want to fully confirm we have it working with both, but then I will remove/update the VUs in the spec for `VUID-VkValidationFeaturesEXT-*`

PR #8561, opened 2024-09-19
Move DebugPrintf Validation Object into GPU-AV

spencer-lunarg:
In goal of https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8192 this doesn't quite let you use both DebugPrintf and GPU-AV together, but it gets **very** close by removing the DebugPrintf Validation Object and has it merged into GPU-AVs path

currently this uses the `enabled[debug_printf_validation]` (what turned it on before) as a way to funnel what gets turned on and off

Added one test (disabled) but the goal is after this PR, we can start to merge things in together since now they share all the objects and it has become just another GPU-AV setting


spencer-lunarg:
closing, going to redo with post SDK branch

Issue #8377, opened 2024-08-11
VK_ERROR_DEVICE_LOST when enabling descriptors indexing validation


Trider12:
**Environment:**
 - OS: Windows 11
 - GPU and driver version: AMD Radeon RX 6950 XT, 2.0.299
 - SDK or header version if building from repo: 1.3.290.0
 - Options enabled (synchronization, best practices, etc.): Descriptors indexing

**Describe the Issue**

I have two graphics pipelines - A and B. A uses sets X and Y, B uses set X. X and Y are bound to sets 0 and 1 respectively. Here's the code:

```c++
VkDescriptorSetLayout setLayouts[] { setLayoutX, setLayoutY };
VkPipelineLayoutCreateInfo pipelineLayoutCreateInfo {};
pipelineLayoutCreateInfo.pSetLayouts = setLayouts;
pipelineLayoutCreateInfo.setLayoutCount = 2;
vkCreatePipelineLayout(device, &pipelineLayoutCreateInfo, nullptr, &pipelineLayoutA);
// create pipeline A using layout A

pipelineLayoutCreateInfo.setLayoutCount = 1;
vkCreatePipelineLayout(device, &pipelineLayoutCreateInfo, nullptr, &pipelineLayoutB);
// create pipeline B using layout B
```
```c++
VkDescriptorSet descriptorSets[] { setX, setY };
vkCmdBindDescriptorSets(cmd, VK_PIPELINE_BIND_POINT_GRAPHICS, pipelineLayoutA, 0, 2, descriptorSets, 0, nullptr);

vkCmdBindPipeline(cmd, VK_PIPELINE_BIND_POINT_GRAPHICS, pipelineA);
vkCmdDraw(...); // uses sets 0 and 1

vkCmdBindPipeline(cmd, VK_PIPELINE_BIND_POINT_GRAPHICS, pipelineB);
vkCmdDraw(...); // uses set 0
```
Submitting this command buffer results in `VK_ERROR_DEVICE_LOST` with descriptors indexing validation enabled. There's no error with it disabled. The error can be avoided by creating pipeline **_B_** with pipeline layout **_A_**.

According to the spec my usage is fine, because layouts A and B are compatible for set 0:

[...]

PR #8565, opened 2024-09-20
gpu: Fix instrumentation desc set binding logic

arno-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8377
This binding can disturb application binding in some circumstances
Add SharedPipelineLayoutSubset tests

### DIFF
--- a/layers/gpu/instrumentation/gpuav_instrumentation.cpp
+++ b/layers/gpu/instrumentation/gpuav_instrumentation.cpp
@@ -22,11 +22,130 @@
 #include "gpu/error_message/gpuav_vuids.h"
 #include "gpu/resources/gpu_shader_resources.h"
 #include "gpu/shaders/gpu_error_header.h"
+#include "state_tracker/shader_object_state.h"
 
 namespace gpuav {
 
-void SetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer &cb_state, VkPipelineBindPoint bind_point,
-                                         const Location &loc) {
+// If application is using shader objects, bindings count will be computed from bound shaders
+static uint32_t GetBindingsCountFromLastBoundPipelineOrShader(Validator &gpuav, VkPipelineBindPoint bind_point,
+                                                              CommandBuffer &cb_state, const LastBound &last_bound) {
+    if (last_bound.pipeline_state && last_bound.pipeline_state->PreRasterPipelineLayoutState()) {
+        return static_cast<uint32_t>(last_bound.pipeline_state->PreRasterPipelineLayoutState()->set_layouts.size());
+    }
+
+    if (const vvl::ShaderObject *main_bound_shader = last_bound.GetFirstShader(bind_point)) {
+        return static_cast<uint32_t>(main_bound_shader->set_layouts.size());
+    }
+
+    // Should not get there, it would mean no pipeline nor shader object was bound
+    assert(false);
+    return 0;
+}
+
+static VkPipelineLayout CreateInstrumentationPipelineLayout(Validator &gpuav, VkPipelineBindPoint bind_point, const Location &loc,
+                                                            const LastBound &last_bound,
+                                                            VkDescriptorSetLayout dummy_desc_set_layout,
+                                                            VkDescriptorSetLayout instrumentation_desc_set_layout,
+                                                            uint32_t inst_desc_set_binding) {
+    // If not using shader objects, GPU-AV should be able to retrieve a pipeline layout from last bound pipeline
+    VkPipelineLayoutCreateInfo pipe_layout_ci = vku::InitStructHelper();
+    std::shared_ptr<const vvl::PipelineLayout> last_bound_pipeline_pipe_layout;
+    if (last_bound.pipeline_state && last_bound.pipeline_state->PreRasterPipelineLayoutState()) {
+        last_bound_pipeline_pipe_layout = last_bound.pipeline_state->PreRasterPipelineLayoutState();
+    }
+    if (last_bound_pipeline_pipe_layout) {
+        // Application is using classic pipelines, compose a pipeline layout from last bound pipeline
+        // ---
+        pipe_layout_ci.flags = last_bound_pipeline_pipe_layout->create_flags;
+        std::vector<VkPushConstantRange> ranges;
+        if (last_bound_pipeline_pipe_layout->push_constant_ranges_layout) {
+            ranges.reserve(last_bound_pipeline_pipe_layout->push_constant_ranges_layout->size());
+            for (const VkPushConstantRange &range : *last_bound_pipeline_pipe_layout->push_constant_ranges_layout) {
+                ranges.push_back(range);
+            }
+        }
+        pipe_layout_ci.pushConstantRangeCount = static_cast<uint32_t>(ranges.size());
+        pipe_layout_ci.pPushConstantRanges = ranges.data();
+        std::vector<VkDescriptorSetLayout> set_layouts;
+        set_layouts.reserve(inst_desc_set_binding + 1);
+        for (const auto &set_layout : last_bound_pipeline_pipe_layout->set_layouts) {
+            set_layouts.push_back(set_layout->VkHandle());
+        }
+        for (uint32_t set_i = static_cast<uint32_t>(last_bound_pipeline_pipe_layout->set_layouts.size());
+             set_i < inst_desc_set_binding; ++set_i) {
+            set_layouts.push_back(dummy_desc_set_layout);
+        }
+        set_layouts.push_back(instrumentation_desc_set_layout);
+        pipe_layout_ci.setLayoutCount = static_cast<uint32_t>(set_layouts.size());
+        pipe_layout_ci.pSetLayouts = set_layouts.data();
+        VkPipelineLayout pipe_layout_handle;
+        VkResult result = DispatchCreatePipelineLayout(gpuav.device, &pipe_layout_ci, VK_NULL_HANDLE, &pipe_layout_handle);
+        if (result != VK_SUCCESS) {
+            gpuav.InternalError(gpuav.device, loc, "Failed to create instrumentation pipeline layout");
+            return VK_NULL_HANDLE;
+        }
+
+        return pipe_layout_handle;
+    } else {
+        // Application is using shader objects, compose a pipeline layout from bound shaders
+        // ---
+
+        const vvl::ShaderObject *main_bound_shader = last_bound.GetFirstShader(bind_point);
+        if (!main_bound_shader) {
+            // Should not get there, it would mean no pipeline nor shader object was bound
+            gpuav.InternalError(gpuav.device, loc, "Could not retrieve last bound computer/vertex/mesh shader");
+            return VK_NULL_HANDLE;
+        }
+
+        // From those VUIDs:
+        // VUID-vkCmdDraw-None-08878:
+        // - All bound graphics shader objects must have been created with identical or identically defined push constant ranges
+        // VUID-vkCmdDraw-None-08879
+        // - All bound graphics shader objects must have been created with identical or identically defined arrays of descriptor set
+        // layouts
+        // => To compose a VkPipelineLayout, only need to get compute or vertex/mesh shader and look at their bindings,
+        // no need to check other shaders.
+        const vvl::ShaderObject::SetLayoutVector *set_layouts = &main_bound_shader->set_layouts;
+        PushConstantRangesId push_constants_layouts = main_bound_shader->push_constant_ranges;
+
+        if (last_bound.desc_set_pipeline_layout) {
+            std::shared_ptr<const vvl::PipelineLayout> last_bound_desc_set_pipe_layout =
+                gpuav.Get<vvl::PipelineLayout>(last_bound.desc_set_pipeline_layout);
+            if (last_bound_desc_set_pipe_layout) {
+                pipe_layout_ci.flags = last_bound_desc_set_pipe_layout->CreateFlags();
+            }
+        }
+        std::vector<VkDescriptorSetLayout> set_layout_handles;
+        if (set_layouts) {
+            set_layout_handles.reserve(inst_desc_set_binding + 1);
+            for (const auto &set_layout : *set_layouts) {
+                set_layout_handles.push_back(set_layout->VkHandle());
+            }
+            for (uint32_t set_i = static_cast<uint32_t>(set_layouts->size()); set_i < inst_desc_set_binding; ++set_i) {
+                set_layout_handles.push_back(dummy_desc_set_layout);
+            }
+            set_layout_handles.push_back(instrumentation_desc_set_layout);
+            pipe_layout_ci.setLayoutCount = static_cast<uint32_t>(set_layout_handles.size());
+            pipe_layout_ci.pSetLayouts = set_layout_handles.data();
+        }
+
+        if (push_constants_layouts) {
+            pipe_layout_ci.pushConstantRangeCount = static_cast<uint32_t>(push_constants_layouts->size());
+            pipe_layout_ci.pPushConstantRanges = push_constants_layouts->data();
+        }
+        VkPipelineLayout pipe_layout_handle;
+        VkResult result = DispatchCreatePipelineLayout(gpuav.device, &pipe_layout_ci, VK_NULL_HANDLE, &pipe_layout_handle);
+        if (result != VK_SUCCESS) {
+            gpuav.InternalError(gpuav.device, loc, "Failed to create instrumentation pipeline layout");
+            return VK_NULL_HANDLE;
+        }
+
+        return pipe_layout_handle;
+    }
+}
+
+void PreCallSetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer &cb_state, VkPipelineBindPoint bind_point,
+                                                const Location &loc) {
     if (!gpuav.gpuav_settings.shader_instrumentation_enabled) {
         return;
     }
@@ -36,9 +155,8 @@ void SetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer &cb_sta
 
     const auto lv_bind_point = ConvertToLvlBindPoint(bind_point);
     auto const &last_bound = cb_state.lastBound[lv_bind_point];
-    const auto *pipeline_state = last_bound.pipeline_state;
 
-    if (!pipeline_state && !last_bound.HasShaderObjects()) {
+    if (!last_bound.pipeline_state && !last_bound.HasShaderObjects()) {
         gpuav.InternalError(cb_state.VkHandle(), loc, "Neither pipeline state nor shader object states were found.");
         return;
     }
@@ -151,24 +269,6 @@ void SetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer &cb_sta
         DispatchUpdateDescriptorSets(gpuav.device, static_cast<uint32_t>(desc_writes.size()), desc_writes.data(), 0, nullptr);
     }
 
-    auto pipeline_layout = pipeline_state ? pipeline_state->PipelineLayoutState()
-                                          : gpuav.Get<vvl::PipelineLayout>(last_bound.desc_set_pipeline_layout);
-    // If GPL is used, it's possible the pipeline layout used at pipeline creation time is null. If CmdBindDescriptorSets has
-    // not been called yet (i.e., state.pipeline_null), then fall back to the layout associated with pre-raster state,
-    // or the last specified pipeline layout in vkCmdPushConstantRanges.
-    // PipelineLayoutState should be used for the purposes of determining the number of sets in the layout, but this layout
-    // may be a "pseudo layout" used to represent the union of pre-raster and fragment shader layouts, and therefore have a
-    // null handle.
-    VkPipelineLayout pipeline_layout_handle = VK_NULL_HANDLE;
-    if (last_bound.desc_set_pipeline_layout) {
-        pipeline_layout_handle = last_bound.desc_set_pipeline_layout;
-    } else if (pipeline_state && !pipeline_state->PreRasterPipelineLayoutState()->Destroyed()) {
-        pipeline_layout_handle = pipeline_state->PreRasterPipelineLayoutState()->VkHandle();
-    } else if (cb_state.push_constant_latest_used_layout[lv_bind_point] != VK_NULL_HANDLE) {
-        pipeline_layout_handle = cb_state.push_constant_latest_used_layout[lv_bind_point];
-        pipeline_layout = gpuav.Get<vvl::PipelineLayout>(pipeline_layout_handle);
-    }
-
     uint32_t operation_index = 0;
     if (bind_point == VK_PIPELINE_BIND_POINT_GRAPHICS)
         operation_index = cb_state.draw_index++;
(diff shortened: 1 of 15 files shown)


layers/state_tracker/pipeline_state.cpp
```diff
@@ -1213,6 +1215,83 @@ bool LastBound::IsAnyGraphicsShaderBound() const {
         IsValidShaderBound(ShaderObjectStage::MESH);
 }
 
+uint32_t LastBound::GetBoundShaderObjectsSetLayoutSize(VkPipelineBindPoint bind_point) const {
```
spencer-lunarg, line 1218: see my attempt in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8581/files

see `VUID-vkCmdDraw-None-08879` we only need to get the `Vertex`/`Mesh` and use that since they will all be the same

arno-lunarg: Ah yes looking at the vertex shader should be enough!

layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -230,6 +432,77 @@ void SetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer &cb_sta
...
+    const LvlBindPoint lv_bind_point = ConvertToLvlBindPoint(bind_point);
+    const LastBound &last_bound = cb_state.lastBound[lv_bind_point];
+
+    if (!last_bound.pipeline_state && !last_bound.HasShaderObjects()) {
+        gpuav.InternalError(cb_state.VkHandle(), loc, "Neither pipeline state nor shader object states were found.");
+        return;
+    }
+
+    std::shared_ptr<const vvl::PipelineLayout> instrumentation_desc_set_binding_pipe_layout_state;
+    const PipelineLayoutSource instrumentation_desc_set_binding_pipe_layout_src = GetInstrumentionDescSetBindingPipelineLayout(
```
spencer-lunarg, line 453: I still not sure why we have all this code in the`Post` it should really just be

```c++
    if (cb_state.reset_disturbed_sets) {
        cb_state.reset_disturbed_sets = false;
        const auto lv_bind_point = ConvertToLvlBindPoint(bind_point);
        auto const &last_bound = cb_state.lastBound[lv_bind_point];
        const auto *pipeline_state = last_bound.pipeline_state;
        uint32_t current_set = 0;
        if (pipeline_state) {
            current_set = (uint32_t)pipeline_state->PipelineLayoutState()->set_layouts.size();
        } else {
            const vvl::ShaderObject *shader_object = last_bound.GetFirstShaderState();
            current_set = (uint32_t)shader_object->set_layouts.size();
        }

        for (size_t i = current_set; i < last_bound.per_set.size(); i++) {
            const auto &set_info = last_bound.per_set[i];
            const VkDescriptorSet set_handle = set_info.bound_descriptor_set->VkHandle();
            DispatchCmdBindDescriptorSets(cb_state.VkHandle(), bind_point, last_bound.desc_set_pipeline_layout, current_set, 1,
                                          &set_handle, static_cast<uint32_t>(set_info.dynamicOffsets.size()),
                                          set_info.dynamicOffsets.data());
        }
    }
```

layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -25,8 +25,174 @@
...
+static std::shared_ptr<const vvl::PipelineLayout> GetLayoutFromLastBoundPipeline(Validator &gpuav, const LastBound &last_bound) {
+    if (last_bound.pipeline_state && last_bound.pipeline_state->PipelineLayoutState()) {
+        return last_bound.pipeline_state->PipelineLayoutState();
+    } else if (last_bound.pipeline_state && last_bound.pipeline_state->PreRasterPipelineLayoutState()) {
+        return last_bound.pipeline_state->PreRasterPipelineLayoutState();
+    }
+    return nullptr;
+}
+
+#if 0
```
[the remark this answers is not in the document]
arno-lunarg: whoopsie-daisy

layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -25,8 +25,174 @@
 
 namespace gpuav {
 
-void SetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer &cb_state, VkPipelineBindPoint bind_point,
-                                         const Location &loc) {
+enum class PipelineLayoutSource { NoPipelineLayout, LastBoundPipeline, LastBoundDescriptorSet, LastPushedConstants };
```
spencer-lunarg, line 28: this whole change is just around descriptor sets, you can't mix and match Push Constants in `vkCmdBindDescriptorSet` and `vkCmdBindPipeline` layouts

layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -25,8 +25,174 @@
...
+    return PipelineLayoutSource::NoPipelineLayout;
+}
+
+// If application is using shader objects, bindings count will be computed from bound shaders
+static uint32_t GetBindingsCountFromLastBoundPipeline(Validator &gpuav, VkPipelineBindPoint bind_point, CommandBuffer &cb_state,
+                                                      const LastBound &last_bound) {
+    if (last_bound.pipeline_state && last_bound.pipeline_state->PipelineLayoutState()) {
+        return static_cast<uint32_t>(last_bound.pipeline_state->PipelineLayoutState()->set_layouts.size());
+    } else if (last_bound.pipeline_state && last_bound.pipeline_state->PreRasterPipelineLayoutState()) {
+        return static_cast<uint32_t>(last_bound.pipeline_state->PreRasterPipelineLayoutState()->set_layouts.size());
```
spencer-lunarg, line 62: we should only use `PreRasterPipelineLayoutState()` there is no reason we would need `PipelineLayoutState()`

both GPL and non-GPL will be able to get it from `PreRasterPipelineLayoutState()`

so can just go

```c++
// Grab from PreRasterPipelineLayoutState to ensure you get it even for GPL
    auto inst_bind_pipeline_layout = pipeline_state ? pipeline_state->PreRasterPipelineLayoutState()
                                                    : gpuav.Get<vvl::PipelineLayout>(last_bound.desc_set_pipeline_layout);
```

layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -25,8 +25,174 @@
...
+    if (result != VK_SUCCESS) {
+        gpuav.InternalError(gpuav.device, loc, "Failed to create instrumentation pipeline layout");
+        return VK_NULL_HANDLE;
+    }
+
+    return pipe_layout_handle;
+}
+#endif
+
+static VkPipelineLayout Foo(Validator &gpuav, VkPipelineBindPoint bind_point, const Location &loc, const LastBound &last_bound,
```
spencer-lunarg, line 116: Foo?

layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -36,9 +202,8 @@ void SetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer &cb_sta
 
     const auto lv_bind_point = ConvertToLvlBindPoint(bind_point);
     auto const &last_bound = cb_state.lastBound[lv_bind_point];
-    const auto *pipeline_state = last_bound.pipeline_state;
```
spencer-lunarg, line 39: why did we get rid of this, there is only 1 pipeline state, why do we need to type `last_bound.pipeline_state` everywhere. The everywhere else (outside of GPU-AV) does it like this as well

layers/gpu/instrumentation/gpuav_instrumentation.cpp
arno-lunarg: You can have a pipeline layout with no descriptor sets, but push constants ranges. We then are going to look for a pipeline layout in the last call to `vkCmdPushConstants`

layers/gpu/instrumentation/gpuav_instrumentation.cpp
arno-lunarg: Argh

layers/gpu/instrumentation/gpuav_instrumentation.cpp
spencer-lunarg: we have a test for this?

layers/gpu/instrumentation/gpuav_instrumentation.cpp
arno-lunarg: I simplified my logic

layers/gpu/instrumentation/gpuav_instrumentation.cpp
spencer-lunarg: sure, don't disagree, then in future would like to remove `const auto *pipeline_state = last_bound.pipeline_state;` the other few spots we have in GPU-AV or it will keep spreading

layers/state_tracker/pipeline_state.cpp
spencer-lunarg: well vertex/mesh/compute ... you will need one of those

layers/gpu/instrumentation/gpuav_instrumentation.cpp
arno-lunarg: Nice, simplified things

layers/gpu/instrumentation/gpuav_instrumentation.cpp
arno-lunarg: (O_O)

layers/gpu/instrumentation/gpuav_instrumentation.cpp
arno-lunarg: yes, that and probably every variable that goose `SomeType some_type;`

layers/state_tracker/pipeline_state.cpp
```diff
@@ -1213,6 +1215,55 @@ bool LastBound::IsAnyGraphicsShaderBound() const {
...
+        if (vvl::ShaderObject *ms = GetShaderStateIfValid(ShaderObjectStage::MESH)) {
+            return static_cast<uint32_t>(ms->set_layouts.size());
+        }
+        return 0;
+    }
+    assert(false);
+    return 0;
+}
+
+std::pair<const vvl::ShaderObject::SetLayoutVector *, PushConstantRangesId> LastBound::GetBoundShaderObjectsLayoutInfo(
```
spencer-lunarg, line 1242: Can we just please return `vvl::ShaderObject*` instead of `std::pair<const vvl::ShaderObject::SetLayoutVector *, PushConstantRangesId>`

This is how we do things now elsewhere and the second I need something else, I don't have a way to add to the pair... plus it is MUCH easier to check if it is null on return

spencer-lunarg: also when do we need both a `GetBoundShaderObjectsLayoutInfo` and `GetBoundShaderObjectsSetLayoutSize` this is VERY specific to GPU-AV 

we really just need to get the `vvl::ShaderObject*`  back

layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -230,6 +382,57 @@ void SetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer &cb_sta
...
+            uint32_t first_set = 0;
+
+            if (uint32_t bindings_counts_from_last_pipeline =
+                    GetBindingsCountFromLastBoundPipeline(gpuav, bind_point, cb_state, last_bound);
+                bindings_counts_from_last_pipeline <
+                static_cast<uint32_t>(last_bound_desc_set_pipe_layout_state->set_layouts.size())) {
+                bindings_count = static_cast<uint32_t>(last_bound_desc_set_pipe_layout_state->set_layouts.size() -
+                                                       bindings_counts_from_last_pipeline);
+                first_set = bindings_counts_from_last_pipeline;
+            }
```
spencer-lunarg, line 423: will need Shader Object version here

layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -230,6 +382,57 @@ void SetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer &cb_sta
...
+
+    const LvlBindPoint lv_bind_point = ConvertToLvlBindPoint(bind_point);
+    const LastBound &last_bound = cb_state.lastBound[lv_bind_point];
+
+    if (!last_bound.pipeline_state && !last_bound.HasShaderObjects()) {
+        gpuav.InternalError(cb_state.VkHandle(), loc, "Neither pipeline state nor shader object states were found.");
+        return;
+    }
+
+    // Only need to rebind application desc sets if they have been disturbed by GPU-AV binding its instrumentation desc set.
```
spencer-lunarg, line 402: can we have a `cb_state.distrubed_state` bool to pass the "`PreCall` did something so `PostCall` now needs to do something"

looking at this function **only**, without looking up when `last_bound.desc_set_pipeline_layout` and `last_bound_desc_set_pipe_layout_state` I really have no idea if this will be called for all cases or only for the case we care about

layers/state_tracker/pipeline_state.cpp
```diff
@@ -1213,6 +1215,55 @@ bool LastBound::IsAnyGraphicsShaderBound() const {
...
+    VkPipelineBindPoint bind_point) const {
+    if (bind_point == VK_PIPELINE_BIND_POINT_COMPUTE) {
+        if (vvl::ShaderObject *cs = GetShaderStateIfValid(ShaderObjectStage::COMPUTE)) {
+            return {&cs->set_layouts, cs->push_constant_ranges};
+        }
+    } else if (bind_point == VK_PIPELINE_BIND_POINT_GRAPHICS) {
+        // VUID-vkCmdDraw-None-08878:
```
spencer-lunarg, line 1249: Just adding these VUs doesn't explain **why** we care

```suggestion
        // VUID-vkCmdDraw-None-08878 and VUID-vkCmdDraw-None-08879 for all VkShaderEXT to have same Push Constant Ranges and Descriptor Sets, so we only need to check one
```

layers/state_tracker/pipeline_state.h
```diff
@@ -26,6 +26,7 @@
 #include "utils/shader_utils.h"
 #include "state_tracker/state_tracker.h"
 #include "state_tracker/shader_stage_state.h"
+#include "state_tracker/shader_object_state.h"
```
spencer-lunarg, line 29: I tried hard to really not have headers include state_tracker objects like this

Another reason why we should just return `vvl::ShaderObject` instead of exposing the inside of it as a return value

layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -25,8 +25,133 @@
 
 namespace gpuav {
 
-void SetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer &cb_state, VkPipelineBindPoint bind_point,
-                                         const Location &loc) {
+enum class PipelineLayoutSource { NoPipelineLayout, LastBoundPipeline, LastBoundDescriptorSet, LastPushedConstants };
+static PipelineLayoutSource GetInstrumentionDescSetBindingPipelineLayout(
```
spencer-lunarg, line 29: honestly this function seems to add more complexity then keeping it inlined

the fact it has to pass in 5 arguments (and has an output from both the return and an output argument) is a LOT for what is really 12 lines of code

if we are now not using this function more than once, I think it is simpler to keep inlined (or a lambda if you are really egar to section it off)

layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -25,8 +25,133 @@
...
+}
+
+static VkPipelineLayout CreateInstrumentationPipelineLayout(Validator &gpuav, VkPipelineBindPoint bind_point, const Location &loc,
+                                                            const LastBound &last_bound,
+                                                            VkDescriptorSetLayout dummy_desc_set_layout,
+                                                            VkDescriptorSetLayout instrumentation_desc_set_layout,
+                                                            uint32_t inst_desc_set_binding) {
+    // If not using shader objects, GPU-AV should be able to retrieve a pipeline layout from last bound pipeline
+    VkPipelineLayoutCreateInfo pipe_layout_ci = vku::InitStructHelper();
+    if (auto last_bound_pipeline_pipe_layout = GetLayoutFromLastBoundPipeline(gpuav, last_bound)) {
```
spencer-lunarg, line 80: if this function is just a big `if/else` between pipeline and Shader Object, I would just create 2 functions overloaded (it is what we have been doing for other Shader Object things)

The caller can check easily with `LastBound::pipeline_state == nullptr` (or even `LastBound::HasShaderObjects()`)

layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -177,45 +287,87 @@ void SetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer &cb_sta
...
+                            gpuav, bind_point, loc, last_bound, gpuav.dummy_desc_layout_,
+                            gpuav.GetInstrumentationDescriptorSetLayout(), gpuav.instrumentation_desc_set_bind_index_);
+
+                        if (instrumentation_pipe_layout != VK_NULL_HANDLE) {
+                            DispatchCmdBindDescriptorSets(cb_state.VkHandle(), bind_point, instrumentation_pipe_layout,
+                                                          gpuav.instrumentation_desc_set_bind_index_, 1, &instrumentation_desc_set,
+                                                          static_cast<uint32_t>(dynamic_offsets.size()), dynamic_offsets.data());
+                            DispatchDestroyPipelineLayout(gpuav.device, instrumentation_pipe_layout, nullptr);
+                        } else {
+                            return;
```
spencer-lunarg, line 338: need a comment or a `LogError` as if we are hitting this, something is wrong and blowing up aftewards probably

layers/state_tracker/pipeline_state.cpp
arno-lunarg: Yes agreed, no need to pollute state tracking with that. I added `GetMainShader` ( name is meh, feel free to find something better (～￣▽￣)～ ) to retrieve bound compute or vertex/mesh shader, and then GPU-AV does its cooking

layers/gpu/instrumentation/gpuav_instrumentation.cpp
arno-lunarg: No need, we get what we want from `GetBindingsCountFromLastBoundPipelineOrShader`

layers/gpu/instrumentation/gpuav_instrumentation.cpp
arno-lunarg: No, no need for additional state. I removed 
```cpp
    if (!last_bound.pipeline_state && !last_bound.HasShaderObjects()) {
        gpuav.InternalError(cb_state.VkHandle(), loc, "Neither pipeline state nor shader object states were found.");
        return;
    }
```
it is not needed anymore, that was your pain point?

layers/state_tracker/pipeline_state.cpp
arno-lunarg: sure

layers/gpu/instrumentation/gpuav_instrumentation.cpp
arno-lunarg: Oh yeah given what you told me I can simplify this quite a bit

layers/gpu/instrumentation/gpuav_instrumentation.cpp
arno-lunarg: Can add a comment, error already has been emitted in `CreateInstrumentationPipelineLayout` at this point. This return is just poor man error propagation ￣へ￣

layers/state_tracker/pipeline_state.cpp
```diff
@@ -1155,6 +1155,36 @@ vvl::ShaderObject *LastBound::GetShaderState(ShaderObjectStage stage) const {
...
+        // - All bound graphics shader objects must have been created with identical or identically defined arrays of descriptor set
+        // layouts
+        if (vvl::ShaderObject *vs = GetShaderStateIfValid(ShaderObjectStage::VERTEX)) {
+            return vs;
+        }
+
+        if (vvl::ShaderObject *ms = GetShaderStateIfValid(ShaderObjectStage::MESH)) {
+            return ms;
+        }
+        return nullptr;
```
spencer-lunarg, line 1181: remove so it hits the assert in the bottom if this happens (as it shouldn't)

layers/state_tracker/pipeline_state.cpp
```diff
@@ -1155,6 +1155,36 @@ vvl::ShaderObject *LastBound::GetShaderState(ShaderObjectStage stage) const {
...
 }
 
+vvl::ShaderObject *LastBound::GetShaderStateIfValid(ShaderObjectStage stage) const {
+    if (!shader_object_bound[static_cast<uint32_t>(stage)]) {
+        return nullptr;
+    }
+    return shader_object_states[static_cast<uint32_t>(stage)];
+}
+
+const vvl::ShaderObject *LastBound::GetMainShader(VkPipelineBindPoint bind_point) const {
```
spencer-lunarg, line 1165: I would call this `GetFirstShader`. There is no concept of a "main shader" anywhere in the spec

layers/gpu/instrumentation/gpuav_instrumentation.cpp
spencer-lunarg: the issue if I was only to look at 

```c++

    if (last_bound.desc_set_pipeline_layout) {
        std::shared_ptr<const vvl::PipelineLayout> last_bound_desc_set_pipe_layout_state =
            gpuav.Get<vvl::PipelineLayout>(last_bound.desc_set_pipeline_layout);
        if (last_bound_desc_set_pipe_layout_state) {
```

I really don't know "when" in the conditions in PreCall it will get here

this should only happen where things are disturbed

having a `bool cb_state.sets_are_disturbed` would make it easy to know where in `PreCall` we turn this on, and obvious in `PostCall` what this is for

the reason is I can easily forsee more logic needing to go into `PostCall` that is **not** related to shader disturbing and someone will see this `if` and think "ya, can probably wrap in this as well"

layers/gpu/instrumentation/gpuav_instrumentation.cpp
spencer-lunarg: If someone report an issue, It would be nice to see the error message knowing it got here... hurts nothing to add it

```suggestion
                            gpuav.InternalError(cb_state.Handle(), loc, "Could not create instrumentation pipeline layout");
                            return;
```

layers/state_tracker/pipeline_state.cpp
```diff
@@ -1155,6 +1155,36 @@ vvl::ShaderObject *LastBound::GetShaderState(ShaderObjectStage stage) const {
...
+    if (bind_point == VK_PIPELINE_BIND_POINT_COMPUTE) {
+        return GetShaderStateIfValid(ShaderObjectStage::COMPUTE);
+    } else if (bind_point == VK_PIPELINE_BIND_POINT_GRAPHICS) {
+        // VUID-vkCmdDraw-None-08878:
```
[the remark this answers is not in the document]
spencer-lunarg: aww, I see you have the `VUID-vkCmdDraw-None-08878` comments now at the caller, so ya, need to remove these here then

layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -177,45 +280,113 @@ void SetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer &cb_sta
     else if (bind_point == VK_PIPELINE_BIND_POINT_RAY_TRACING_KHR)
         operation_index = cb_state.trace_rays_index++;
 
+    const bool uses_shader_object = last_bound.pipeline_state == nullptr;
+
+    // Bind instrumentation descriptor set, using an appropriate pipeline layout
+    // ---
+
+    // First find this appropriate pipeline layout
+    // If GPL is used, it's possible the pipeline layout used at pipeline creation time is null.
```
spencer-lunarg, line 289: this whole thing is trying to say "use `PreRasterPipelineLayoutState() instead of `PipelineLayoutState()`" but does a bad job at it IMO

layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -22,11 +22,133 @@
...
 
 namespace gpuav {
 
-void SetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer &cb_state, VkPipelineBindPoint bind_point,
-                                         const Location &loc) {
+// If application is using shader objects, bindings count will be computed from bound shaders
+static uint32_t GetBindingsCountFromLastBoundPipelineOrShader(Validator &gpuav, VkPipelineBindPoint bind_point,
+                                                              CommandBuffer &cb_state, const LastBound &last_bound) {
+    if (last_bound.pipeline_state && last_bound.pipeline_state->PipelineLayoutState()) {
+        return static_cast<uint32_t>(last_bound.pipeline_state->PipelineLayoutState()->set_layouts.size());
```
spencer-lunarg, line 33: I thought we said somewhere else to always just call `PreRasterPipelineLayoutState` (and add a comment) because it will work for both normal Pipeline and GPL

arno-lunarg: aaaaaaah I remember doing it, but it was somewhere else then : //

layers/state_tracker/pipeline_state.h
```diff
@@ -729,6 +729,9 @@ struct LastBound {
     bool ValidShaderObjectCombination(const VkPipelineBindPoint bind_point, const DeviceFeatures &device_features) const;
     VkShaderEXT GetShader(ShaderObjectStage stage) const;
     vvl::ShaderObject *GetShaderState(ShaderObjectStage stage) const;
+    vvl::ShaderObject *GetShaderStateIfValid(ShaderObjectStage stage) const;
```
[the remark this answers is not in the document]
arno-lunarg: It could make sense if we care about retrieving shader state whether or not it is valid, but then that seems looks like a weird design decision. From a quick look at how `GetShaderState` is used, not sure I can just replace it with `GetShaderIfValid`

I at least made `GetShaderStateIfValid` return a `const` pointer

layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -230,6 +396,51 @@ void SetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer &cb_sta
...
+
+            if (uint32_t bindings_counts_from_last_pipeline =
+                    GetBindingsCountFromLastBoundPipelineOrShader(gpuav, bind_point, cb_state, last_bound);
+                bindings_counts_from_last_pipeline <
+                static_cast<uint32_t>(last_bound_desc_set_pipe_layout_state->set_layouts.size())) {
+                bindings_count = static_cast<uint32_t>(last_bound_desc_set_pipe_layout_state->set_layouts.size() -
+                                                       bindings_counts_from_last_pipeline);
+                first_set = bindings_counts_from_last_pipeline;
+            }
+            for (uint32_t set_i = 0; set_i < bindings_count; ++set_i) {
```
spencer-lunarg, line 432: ok, so I am about to sign off on this PR except for this... 

it is not clear to anyone new to the code (or probably me in the near future when there is a change) that inside `PostCallSetupShaderInstrumentationResources` that everything above this line is not actually needed unless the sizes are different

like for the normal case (because I am still of high belief most people don't try to create compatible layouts and design their shades in such a way that this whole issue caused) we still need to call `GetBindingsCountFromLastBoundPipelineOrShader` and do all that checking... but it is not actutaly needed

so for both performance (aka, we don't need to run all the state object `Get` each draw call, as well as clarity that this `PostCall` logic is only for the well explained comment above, I would **really** prefer some `bool` to communicate that we have disturbed the descriptor set and need to fix it... then at the top of this function we have a simple 

```c++
if (cb_state.disturbed_descriptor_set) {
   cb_state.disturbed_descriptor_set = true;
   // rest of logic
}
```

this way if this function grows larger and larger, it is clear why/when this occurs

arno-lunarg: No need to add state to store something we can just cheaply compute. No state means no need to take care of it, make sure it is always set, make sure it is no used at another spot because wait it was not intended to be used this way, etc... 
I can add other comments on this loop to clarify things further. 
This decision makes sense now, if in the future having to maintain state is a good trade off, then we will do it, but for now, nope

spencer-lunarg: > No need to add state to store something we can just cheaply compute.

I could make the argument for EVERY thing then... when is it now "not worth it"

We have to communicate between the `PreCall`/`PostCall` boundary, a simple boolean seems like the best way to do that.

spencer-lunarg: thanks @jeremyg-lunarg 

- no to the boolean
- yes to `bool bindings_disturbed = <complicated logic>; ` 

and I am happy

jeremyg-lunarg: > * no to the boolean

You mean no to the cb_state boolean?

spencer-lunarg: > I'm REALLY not a fan of adding a flag to cb_state for this.

sorry, was trying to say "ok, I concede and won't ask for the boolean" as long as the other things you said are met

layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -230,6 +396,53 @@ void SetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer &cb_sta
...
+        std::shared_ptr<const vvl::PipelineLayout> last_bound_desc_set_pipe_layout_state =
+            gpuav.Get<vvl::PipelineLayout>(last_bound.desc_set_pipeline_layout);
+        if (last_bound_desc_set_pipe_layout_state) {
+            const uint32_t bindings_counts_from_last_pipeline =
+                GetBindingsCountFromLastBoundPipelineOrShader(gpuav, bind_point, cb_state, last_bound);
+
+            const bool any_disturbed_bindings = bindings_counts_from_last_pipeline <
+                                                static_cast<uint32_t>(last_bound_desc_set_pipe_layout_state->set_layouts.size());
+
+            if (any_disturbed_bindings) {
```
spencer-lunarg, line 427: :100:

spencer-lunarg (approved):

PR #8572, opened 2024-09-21
tests: Attempt to fix KhronosTimelineSemaphoreExample

artem-lunarg:
* Add actual work to submissions in case there is some issue with synchronization  of empty batches.
* Add iteration count to test more combinations of threading behavior.


artem-lunarg:
Does not fix the issue. Can reproduce locally.

artem-lunarg:
Lets stress test CI machines a bit on weekends.

PR #8575, opened 2024-09-22
Fix matching shader object input output interface

ziga-lunarg:


spencer-lunarg (approved):
its always the Geometry shaders breaking things...

ziga-lunarg:
> its always the Geometry shaders breaking things...

No I found this bug with your favourite: tessellation shaders. A geometry shader test is just easier to write.

PR #8577, opened 2024-09-22
Fix VUID 07483, only check if sample locations is enabled

ziga-lunarg:


layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -582,8 +582,9 @@ bool CoreChecks::ValidateGraphicsDynamicStateValue(const LastBound& last_bound_s
                     string_VkSampleCountFlagBits(pipeline.MultisampleState()->rasterizationSamples),
                     cb_state.DescribeInvalidatedState(CB_DYNAMIC_STATE_SAMPLE_LOCATIONS_EXT).c_str());
             }
-        } else if (cb_state.dynamic_state_value.sample_locations_info.sampleLocationsPerPixel !=
-                   cb_state.dynamic_state_value.rasterization_samples) {
+        } else if (last_bound_state.IsSampleLocationsEnable() &&
```
spencer-lunarg, line 585: I think `IsSampleLocationsEnable()` should be moved  also be for `if (pipeline.IsDynamic(CB_DYNAMIC_STATE_SAMPLE_LOCATIONS_EXT) && IsSampleLocationsEnable()) {`

because I imagine `07482` will suffer the same issue

I am going to make a spec fix to have both `07482` and `07483` clearly state that `sampleLocationsEnable` needs to be `VK_TRUE`

spencer-lunarg: https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6896

spencer-lunarg (approved):

PR #8581, opened 2024-09-23
[Reference] gpu: Fix instrumentation desc set binding logic 

spencer-lunarg:
Takes https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8565 and adds my try on top of it 

Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8377

The basic idea is if we go

```
vkCmdBindDescriptors([0, 1, 2, 3])
vkCmdBindPipeline([0, 1])
vkCmdDraw()
```

we will now use the layout from `vkCmdBindPipeline` (and re-created it if it was destroyed) 

afterwards we will re-set the sets `[2, 3]` since they will have been disturbed

### DIFF
--- a/layers/gpu/core/gpuav.h
+++ b/layers/gpu/core/gpuav.h
@@ -147,59 +147,114 @@ class Validator : public gpu::GpuShaderInstrumentor {
         const RecordObject& record_obj) final;
     void PreCallRecordCmdDraw(VkCommandBuffer commandBuffer, uint32_t vertexCount, uint32_t instanceCount, uint32_t firstVertex,
                               uint32_t firstInstance, const RecordObject& record_obj) final;
+    void PostCallRecordCmdDraw(VkCommandBuffer commandBuffer, uint32_t vertexCount, uint32_t instanceCount, uint32_t firstVertex,
+                               uint32_t firstInstance, const RecordObject& record_obj) final;
     void PreCallRecordCmdDrawMultiEXT(VkCommandBuffer commandBuffer, uint32_t drawCount, const VkMultiDrawInfoEXT* pVertexInfo,
                                       uint32_t instanceCount, uint32_t firstInstance, uint32_t stride,
                                       const RecordObject& record_obj) final;
+    void PostCallRecordCmdDrawMultiEXT(VkCommandBuffer commandBuffer, uint32_t drawCount, const VkMultiDrawInfoEXT* pVertexInfo,
+                                       uint32_t instanceCount, uint32_t firstInstance, uint32_t stride,
+                                       const RecordObject& record_obj) final;
     void PreCallRecordCmdDrawIndexed(VkCommandBuffer commandBuffer, uint32_t indexCount, uint32_t instanceCount,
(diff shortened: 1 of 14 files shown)


tests/unit/gpu_av.cpp
```diff
@@ -18,30 +18,6 @@
 
 class NegativeGpuAV : public GpuAVTest {};
 
-TEST_F(NegativeGpuAV, DestroyedPipelineLayout) {
```
spencer-lunarg, line 21: moved this test to `NegativeGpuAVDescriptorIndexing, DestroyedPipelineLayou`

it does the same thing, but now we don't report the error, handle it correctly, and are still able to do normal GPU-AV checks afterwards!

arno-lunarg:
It does not handle shader objects...

layers/gpu/instrumentation/gpu_shader_instrumentor.h
```diff
@@ -225,6 +227,8 @@ class GpuShaderInstrumentor : public ValidationStateTracker {
     std::atomic<uint32_t> unique_shader_module_id_ = 1;  // zero represents no shader module found
     // The descriptor slot we will be injecting our error buffer into
     uint32_t desc_set_bind_index_ = 0;
+    // This is a layout used to "pad" a pipeline layout to fill in any gaps to the selected bind index
+    VkDescriptorSetLayout dummy_desc_layout_ = VK_NULL_HANDLE;
```
arno-lunarg, line 231: For concurrency reasons, this needs to be stored in CommanBuffer state

layers/gpu/core/gpuav.h
```diff
@@ -326,6 +404,11 @@ class Validator : public gpu::GpuShaderInstrumentor {
     std::optional<DescriptorHeap> desc_heap_{};  // optional only to defer construction
     gpu::SharedResourcesManager shared_resources_manager;
 
+    bool reset_disturbed_sets = false;
+    // Only used to build a quick layout between a Pre/Post boundry.
+    // The PostCall function is responsible to destroy clean this up.
+    VkPipelineLayout temp_pipeline_layout = VK_NULL_HANDLE;
```
arno-lunarg, line 410: Those cannot live here for concurrency reasons, need to be moved to CommandBuffer

layers/state_tracker/pipeline_state.cpp
```diff
@@ -1155,6 +1155,13 @@ vvl::ShaderObject *LastBound::GetShaderState(ShaderObjectStage stage) const {
     return shader_object_states[static_cast<uint32_t>(stage)];
 }
 
+// Apps are required to have either a Vertex or Mesh shader object (when using shader objects).
+vvl::ShaderObject *LastBound::GetFirstShaderState() const {
```
spencer-lunarg, line 1159: Needs to add Compute as well

spencer-lunarg:
closing for https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8565

PR #8589, opened 2024-09-24
tests: Simplify threading configuration

artem-lunarg:
When this test was introduced running time could vary 10x depending on the device. Reduce the number of threads to reduce threading contention (but still keep high enough) with a hope to make running time a bit less unpredictable.


spencer-lunarg (approved):

artem-lunarg:
The updated test takes around 4 seconds on internal CI Android devices.

PR #8593, opened 2024-09-25
Combining more Dynamic State for upcoming header

spencer-lunarg:
The following were approved and will be merged in for the next headers

- https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6882
- https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6872
- https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6871
- https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6876

This combines them accordingly

### DIFF
--- a/layers/core_checks/cc_cmd_buffer_dynamic.cpp
+++ b/layers/core_checks/cc_cmd_buffer_dynamic.cpp
@@ -185,6 +185,9 @@ bool CoreChecks::ValidateDynamicStateIsSet(const LastBound& last_bound_state, co
             case CB_DYNAMIC_STATE_VIEWPORT_COARSE_SAMPLE_ORDER_NV:
                 vuid_str = vuid.set_viewport_coarse_sample_order_09233;
                 break;
+            case CB_DYNAMIC_STATE_VIEWPORT_SHADING_RATE_PALETTE_NV:
+                vuid_str = vuid.set_viewport_shading_rate_palette_09234;
+                break;
             default:
                 assert(false);
                 break;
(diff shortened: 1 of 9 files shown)


layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -343,10 +380,21 @@ bool CoreChecks::ValidateGraphicsDynamicStateSetStatus(const LastBound& last_bou
...
         skip |= ValidateDynamicStateIsSet(last_bound_state, state_status_cb, CB_DYNAMIC_STATE_TESSELLATION_DOMAIN_ORIGIN_EXT, vuid);
     }
 
+    if (tesc_shader_bound) {
+        // Don't call GetPrimitiveTopology() because want to view the Topology from the dynamic state for ShaderObjects
+        const VkPrimitiveTopology topology =
+            (!last_bound_state.pipeline_state || last_bound_state.pipeline_state->IsDynamic(CB_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY))
+                ? cb_state.dynamic_state_value.primitive_topology
+                : last_bound_state.pipeline_state->topology_at_rasterizer;
+        if (topology == VK_PRIMITIVE_TOPOLOGY_PATCH_LIST) {
```
ziga-lunarg, line 393: Is this check necessary? It is the only valid topology and everything else should already be caught before this point

tests/unit/shader_object_positive.cpp
```diff
@@ -1632,3 +1632,37 @@ TEST_F(PositiveShaderObject, DrawWithVertGeomFragShaderObjects) {
     m_commandBuffer->EndRendering();
     m_commandBuffer->end();
 }
+
+TEST_F(PositiveShaderObject, MissingCmdSetPatchControlPointsEXT) {
+    RETURN_IF_SKIP(InitBasicShaderObject());
+    VkPhysicalDeviceFeatures features;
+    GetPhysicalDeviceFeatures(&features);
+    if (features.tessellationShader == VK_FALSE) {
```
ziga-lunarg, line 1640: Why not use `AddRequiredFeature`

tests/unit/shader_object_positive.cpp
```diff
@@ -1632,3 +1632,37 @@ TEST_F(PositiveShaderObject, DrawWithVertGeomFragShaderObjects) {
...
+    m_commandBuffer->begin();
+    m_commandBuffer->BeginRenderingColor(GetDynamicRenderTarget(), GetRenderTargetArea());
+    SetDefaultDynamicStatesExclude({VK_DYNAMIC_STATE_PATCH_CONTROL_POINTS_EXT});
+    const VkShaderStageFlagBits stages[] = {VK_SHADER_STAGE_VERTEX_BIT, VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT,
+                                            VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT, VK_SHADER_STAGE_GEOMETRY_BIT,
+                                            VK_SHADER_STAGE_FRAGMENT_BIT};
+    const VkShaderEXT shaders[] = {vertShader.handle(), tescShader.handle(), teseShader.handle(), VK_NULL_HANDLE,
+                                   fragShader.handle()};
+    vk::CmdBindShadersEXT(m_commandBuffer->handle(), 5u, stages, shaders);
+    // valid because this is not VK_PRIMITIVE_TOPOLOGY_PATCH_LIST
```
ziga-lunarg, line 1661: I think there should be a VUID against this, because it doesn't make sense to draw with tessellation shaders and use anything but VK_PRIMITIVE_TOPOLOGY_PATCH_LIST

layers/core_checks/cc_cmd_buffer_dynamic.cpp
spencer-lunarg: the spec say

image

tests/unit/shader_object_positive.cpp
spencer-lunarg: because I just copy and pasted your test :laughing: 

let me update both

tests/unit/shader_object_positive.cpp
spencer-lunarg: I think the topology patch has some quirks to get out... I am going to pull that commit out and make after getting the other ones in

PR #8595, opened 2024-09-25
tests: Remove incorrect timeline semaphore tests

artem-lunarg:
The mistake was that one signal was prefered over the other, but without additional synchronization, if two signals can resolve the wait then any of them can become a resolving signal. The specification does not define which one.


spencer-lunarg (approved):

artem-lunarg:
Merging anyway, failure due to CI machine connectivity problem.

PR #8599, opened 2024-09-26
gpu: Force shaderInt64 on

spencer-lunarg:
attempt 2 from https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8596 (cc @arno-lunarg )

### DIFF
--- a/layers/gpu/instrumentation/gpu_shader_instrumentor.cpp
+++ b/layers/gpu/instrumentation/gpu_shader_instrumentor.cpp
@@ -191,6 +191,10 @@ void GpuShaderInstrumentor::PreCallRecordCreateDevice(VkPhysicalDevice physicalD
                             "Forcing VkPhysicalDeviceFeatures::vertexPipelineStoresAndAtomics to VK_TRUE");
             enabled_features->vertexPipelineStoresAndAtomics = VK_TRUE;
         }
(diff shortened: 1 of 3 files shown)


spencer-lunarg:
(will wait until after https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8565 to bring this in)

layers/gpu/instrumentation/gpu_shader_instrumentor.cpp
```diff
@@ -191,6 +191,10 @@ void GpuShaderInstrumentor::PreCallRecordCreateDevice(VkPhysicalDevice physicalD
                             "Forcing VkPhysicalDeviceFeatures::vertexPipelineStoresAndAtomics to VK_TRUE");
             enabled_features->vertexPipelineStoresAndAtomics = VK_TRUE;
         }
+        if (supported_features.shaderInt64 && !enabled_features->shaderInt64) {
```
arno-lunarg, line 194: (Can't comment above)
Don't we also need to look for features in a potential `VkPhysicalDeviceFeatures2` that would be found in `modified_create_info->pNext`?

arno-lunarg (approved):
Not sure not looking in `VkPhysicalDeviceFeatures2` is ok? But this problem was here before

Issue #8600, opened 2024-09-26
Apply 296 changes to VK_EXT_swapchain_maintenance1


spencer-lunarg:
`VK_EXT_swapchain_maintenance1` was updated the [1.3.296](https://github.com/KhronosGroup/Vulkan-Docs/commit/486e4b289053a7d64784e7ce791711843c60c235) ([internal MR](https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6199))

- [x] `VUID-VkSwapchainPresentFenceInfoEXT-pFences-07758` and `VUID-VkSwapchainPresentFenceInfoEXT-pFences-07759` need to check for `VK_NULL_HANDLE`
- [x] `VUID-VkSwapchainPresentScalingCreateInfoEXT-scalingBehavior-07770` can be zero
- [x] `VUID-VkSwapchainPresentScalingCreateInfoEXT-scalingBehavior-07771` can be zero
- [x] `VUID-VkSwapchainPresentScalingCreateInfoEXT-scalingBehavior-07772` can be zero
- [x] `VUID-VkSwapchainPresentScalingCreateInfoEXT-scalingBehavior-07773` can be zero
- [x] `VUID-VkSwapchainPresentScalingCreateInfoEXT-scalingBehavior-07774` can be zero
- [x] `VUID-VkSwapchainPresentScalingCreateInfoEXT-scalingBehavior-07775` can be zero
- [ ]  Split ` VUID-VkSwapchainCreateInfoKHR-presentMode-02839` into `VUID-VkSwapchainCreateInfoKHR-swapchainMaintenance1-10155` and ` VUID-VkSwapchainCreateInfoKHR-swapchainMaintenance1-10156`
- [ ] Add `VUID-VkSwapchainPresentScalingCreateInfoEXT-swapchainMaintenance1-10154`
- [ ] Add `VUID-VkSwapchainCreateInfoKHR-swapchainMaintenance1-10157`
- [ ] Add ` VUID-VkPresentInfoKHR-swapchainMaintenance1-10158`

spencer-lunarg:
closing - the remaining VUs are spec bugs and need to be resolved

PR #8601, opened 2024-09-26
layers: Update VK_EXT_swapchain_maintenance1 VUs

spencer-lunarg:
For https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8600

### DIFF
--- a/tests/unit/wsi.cpp
+++ b/tests/unit/wsi.cpp
@@ -3581,3 +3581,30 @@ TEST_F(NegativeWsi, NonSupportedPresentMode) {
     vk::CreateSwapchainKHR(device(), &swapchain_create_info, nullptr, &swapchain);
     m_errorMonitor->VerifyFound();
 }
+
+TEST_F(NegativeWsi, SwapchainMaintenance1DeferredMemoryFlags) {
+    AddSurfaceExtension();
+    RETURN_IF_SKIP(Init());
+    RETURN_IF_SKIP(InitSurface());
+    InitSwapchainInfo();
+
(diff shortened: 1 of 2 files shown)


tests/unit/wsi.cpp
```diff
@@ -3581,3 +3581,30 @@ TEST_F(NegativeWsi, NonSupportedPresentMode) {
...
+    swapchain_ci.imageArrayLayers = 1;
+    swapchain_ci.imageUsage = VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT;
+    swapchain_ci.imageSharingMode = VK_SHARING_MODE_EXCLUSIVE;
+    swapchain_ci.preTransform = VK_SURFACE_TRANSFORM_IDENTITY_BIT_KHR;
+    swapchain_ci.compositeAlpha = m_surface_composite_alpha;
+    swapchain_ci.presentMode = m_surface_non_shared_present_mode;
+    swapchain_ci.clipped = VK_FALSE;
+    swapchain_ci.oldSwapchain = 0;
+
+    m_errorMonitor->SetDesiredError("VUID-VkSwapchainCreateInfoKHR-flags-parameter");
```
artem-lunarg, line 3607: Is it because extension was not enabled?

artem-lunarg (approved):
LGTM

tests/unit/wsi.cpp
spencer-lunarg: yes, there was a dedicated `VUID-VkSwapchainCreateInfoKHR-swapchainMaintenance1-10157` added and I realize that VU is not needed as we cover checking for enum with the `flags-parameter` implicit checks (so kept test as proof it is being tested)

PR #8612, opened 2024-09-27
ci: Remove redundant Thread Sanitize testing

spencer-lunarg:
Based on  https://github.com/KhronosGroup/Vulkan-ValidationLayers/actions/runs/11071704738/usage 

as stated in comment, the goal is to save 15+ minutes of CI time per run reducing non-useful testing


artem-lunarg (approved):
LGTM. I wonder if it's possible to exclude directly in matrix/exclude section.

spencer-lunarg:
> LGTM. I wonder if it's possible to exclude directly in matrix/exclude section.

I assume there is a lot of ways to do it, but I'm only a part-time `part-time CI Engineer` so already ran out of my allocated time to look more into :shrug:

PR #8617, opened 2024-09-30
printf: Ray tracing tests

arno-lunarg:
Added new ray tracing tests, that actually trace a ray into a triangle and expect to report intersections. It showed some flaws in the ray tracing framework, main one being a synchronisation issue with the function in charge of return a ready to use TLAS


layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -744,7 +744,7 @@ bool CoreChecks::CommonBuildAccelerationStructureValidation(const VkAcceleration
                     objlist.add(info.srcAccelerationStructure);
                     skip |= LogError(vuid, objlist, geom_loc.dot(Field::flags), "is %s but was last specified as %s.",
                                      string_VkGeometryFlagsKHR(updated_geometry.flags).c_str(),
-                                     string_VkGeometryFlagsKHR(last_geometry.geometryType).c_str());
```
arno-lunarg, line 747: A shame this was not a compile time error....

spencer-lunarg (approved):

Issue #8505, opened 2024-09-06
remove shader_instrumentation_enabled for IsShaderInstrumentationEnabled


spencer-lunarg:
from https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8500/

We currently have both a `IsShaderInstrumentationEnabled` and `shader_instrumentation_enabled` and should just remove the `shader_instrumentation_enabled` but need to go through and make sure every spot is good with in

PR #8620, opened 2024-09-30
gpu: Move DebugPrintf Validation Object into GPU-AV

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8505
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8192 ( :partying_face: )

From here you will now be able to use both DebugPrintf and GPU-AV together at the same time!

image


docs/debug_printf.md
```diff
@@ -25,9 +25,13 @@ The Debug Printf settings are managed by configuring the Validation Layer. These
 
 Debug Printf settings can also be managed using the [Vulkan Configurator](https://vulkan.lunarg.com/doc/sdk/latest/windows/vkconfig.html) included with the Vulkan SDK.
 
+For those who "just need to quick use it" there is also a `set VK_LAYER_PRINTF_ONLY_PRESET=1` environment variable that will turn on DebugPrintf and turn off all of the other validation logic.
+
 ### Settings
 
 > All settings are found in Vulkan Configurator (`VkConfig`)
+>
+> Even if you use `VK_LAYER_PRINTF_ONLY_PRESET` you need to set this settings yourself as desired
```
arno-lunarg, line 34:
```suggestion
> Even if you use `VK_LAYER_PRINTF_ONLY_PRESET` you need to set those settings yourself as desired
```

layers/gpu/instrumentation/gpu_shader_instrumentor.cpp
```diff
@@ -1368,33 +1371,34 @@ bool GpuShaderInstrumentor::InstrumentShader(const vvl::span<const uint32_t> &in
 
     gpu::spirv::Module module(input_spirv, debug_report, module_settings);
 
-    // For now, we don't yet support (or have tested) combining GPU-AV and DebugPrintf, so have 2 paths here
-    const bool is_debug_printf = debug_printf_enabled;
-
+    GpuAVSettings::ShaderInstrumentation &shader_instrumentation = gpuav_settings.shader_instrumentation;
```
arno-lunarg, line 1374: Just use `gpuav_settings.shader_instrumentation` directly, no need to lose info for the sake of having to type less characters

layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -168,13 +167,35 @@ void PreCallSetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer
         return;
     }
 
+    gpu::DeviceMemoryBlock debug_printf_output_buffer = {};
```
arno-lunarg, line 170: We probably should not do that if printf is not enabled?

layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -168,13 +167,35 @@ void PreCallSetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer
...
     // Update instrumentation descriptor set
     {
         // Pathetic way of trying to make sure we take care of updating all
         // bindings of the instrumentation descriptor set
-        assert(gpuav.instrumentation_bindings_.size() == 6);
+        assert(gpuav.instrumentation_bindings_.size() == 7);
         std::vector<VkWriteDescriptorSet> desc_writes = {};
 
+        // DebugPrintf output buffer
+        VkDescriptorBufferInfo debug_printf_desc_buffer_info = {};
```
arno-lunarg, line 184: Same here. I think the logic to update/not update GPU-AV or printf descriptors needs to be cleaned up here?

layers/gpu/resources/gpuav_subclasses.cpp
```diff
@@ -458,14 +478,26 @@ bool CommandBuffer::NeedsPostProcess() { return !error_output_buffer_.IsNull();
...
+
+    // For the given command buffer, map its debug data buffers and read their contents for analysis.
+    for (auto &buffer_info : debug_printf_buffer_infos) {
+        char *data;
+
+        VkResult result = vmaMapMemory(gpuav->vma_allocator_, buffer_info.output_mem_block.allocation, (void **)&data);
+        if (result == VK_SUCCESS) {
+            debug_printf::AnalyzeAndGenerateMessage(*gpuav, VkHandle(), queue, buffer_info, (uint32_t *)data, loc);
+            vmaUnmapMemory(gpuav->vma_allocator_, buffer_info.output_mem_block.allocation);
+        }
```
arno-lunarg, line 491:
```suggestion
        } else {
        gpuav.InternalError(queue, loc, "Unable to map device memory allocated for DebugPrintf buffer.", true);
        return;
    }
```

layers/gpu/resources/gpuav_subclasses.h
```diff
@@ -126,6 +103,7 @@ class CommandBuffer : public gpu_tracker::CommandBuffer {
     const gpu::DeviceMemoryBlock &GetBdaRangesSnapshot() const { return bda_ranges_snapshot_; }
 
     void ClearCmdErrorsCountsBuffer(const Location &loc) const;
+    void UpdateCommandCount(VkPipelineBindPoint bind_point);
```
arno-lunarg, line 106: This is really cool, never took the time to do that!

layers/layer_options.cpp
```diff
@@ -1012,11 +1016,20 @@ void ProcessConfigAndEnvSettings(ConfigAndEnvSettings *settings_data) {
                              VK_LAYER_VALIDATE_BEST_PRACTICES_NVIDIA);
         SetValidationSetting(layer_setting_set, settings_data->enables, sync_validation, VK_LAYER_VALIDATE_SYNC);
 
-        if (vkuHasLayerSetting(layer_setting_set, VK_LAYER_VALIDATE_GPU_BASED)) {
+        // Before GPU-AV and DebugPrintf were megered, we used this enum to set GPU-AV and DebugPrintf in vkconfig.
```
arno-lunarg, line 1019:
```suggestion
        // Before GPU-AV and DebugPrintf were merged, we used this enum to set GPU-AV and DebugPrintf in vkconfig.
```

layers/layer_options.cpp
```diff
@@ -1012,11 +1016,20 @@ void ProcessConfigAndEnvSettings(ConfigAndEnvSettings *settings_data) {
...
+        // This code should in theory be dead since removing it from vkconfig, but keep just incase for a bit
```
arno-lunarg, line 1020: Maybe make an issue to remember to remove that?

layers/layer_options.h
```diff
@@ -69,6 +69,7 @@ enum EnableFlags {
     vendor_specific_amd,
     vendor_specific_img,
     vendor_specific_nvidia,
+    // DebugPrintf was merged with GPU-AV, so we now just use this funnel into the new settings
```
arno-lunarg, line 72: Can you rephrase that?

arno-lunarg (commented):
Nice to see it through!!! I only really have 2 requests:
1) `PreCallSetupShaderInstrumentationResources` should probably manage resources depending on what is enabled, not always do both GPU-AV and printf. Just use dummy resources when one is disabled.
2) Need new tests similar to the exist on/off settings tests (see `ShaderInstrumentationMainSettings.PositiveGpuAVParameterized`) with debug_printf option

layers/gpu/instrumentation/gpuav_instrumentation.cpp
spencer-lunarg: So a big issue we have currently is even if I turn off various parts of GPU-AV, we still allocate for everything here, so yes, I want/plan to reduce to wasteful allocation, but this is a larger GPU-AV issue already

layers/gpu/resources/gpuav_subclasses.h
spencer-lunarg: ya, it should just always be at the end, otherwise easy to have cases where you do an early return in `PreCallSetupShaderInstrumentationResources` and it wasn't being increased

layers/layer_options.cpp
spencer-lunarg: I will add a "deprecated after 1.3.296 SDK" like we do with the other settings and maybe we will just set 1 year as the "time to remove" for all the other deprecated settings as well

layers/gpu/resources/gpuav_subclasses.cpp
spencer-lunarg: I realize we should just be checking `if (result != VK_SUCCESS) {` everywhere instead (we are not consistent on this)

layers/gpu/instrumentation/gpu_shader_instrumentor.cpp
```diff
@@ -452,7 +455,7 @@ void GpuShaderInstrumentor::PreCallRecordCreatePipelineLayout(VkDevice device, c
                                                               const VkAllocationCallbacks *pAllocator,
                                                               VkPipelineLayout *pPipelineLayout, const RecordObject &record_obj,
                                                               chassis::CreatePipelineLayout &chassis_state) {
-    if (gpuav_settings.shader_instrumentation_enabled) {
+    if (gpuav_settings.IsShaderInstrumentationEnabled() || gpuav_settings.debug_printf_enabled) {
```
jeremyg-lunarg, line 458: It would be nice if there was a method that just returned this frequently repeated expression so that some one doesn't forget it somewhere.

spencer-lunarg: happy to do it, need a name... `IsShaderInstrumentationOrDebugPrintfEnabled()`?

spencer-lunarg: (this is because we decided to not mix/match "shader instrumentation" and "Debug printf" in the settings for `vkconfig` and want to keep it consistent here

spencer-lunarg: `IsSpirvModified()`

tests/unit/debug_printf.cpp
```diff
@@ -1464,7 +1464,7 @@ TEST_F(NegativeDebugPrintf, GPLFragment) {
     m_errorMonitor->VerifyFound();
 }
 
-TEST_F(NegativeDebugPrintf, GPLFragmentIndependentSets) {
+TEST_F(NegativeDebugPrintf, DISABLED_GPLFragmentIndependentSets) {
```
spencer-lunarg, line 1467: waiting on fix from https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8630

layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -144,11 +145,106 @@ static VkPipelineLayout CreateInstrumentationPipelineLayout(Validator &gpuav, Vk
     }
 }
 
+void ActionCommandUpdate(Validator &gpuav, CommandBuffer &cb_state, VkDescriptorSet instrumentation_desc_set, const Location &loc) {
```
arno-lunarg, line 148:
```suggestion
void UpdateInstrumentationDescSet(Validator &gpuav, CommandBuffer &cb_state, VkDescriptorSet instrumentation_desc_set, const Location &loc) {
```

layers/gpu/debug_printf/debug_printf.cpp
```diff
@@ -350,360 +332,77 @@ void Validator::AnalyzeAndGenerateMessage(VkCommandBuffer command_buffer, VkQueu
...
-
-CommandBuffer::~CommandBuffer() { Destroy(); }
+    memset(data, 0, gpuav.gpuav_settings.debug_printf_buffer_size);
+    vmaUnmapMemory(gpuav.vma_allocator_, debug_printf_output_buffer.allocation);
 
-void CommandBuffer::Destroy() {
     {
-        auto guard = WriteLock();
-        ResetCBState();
+        VkDescriptorBufferInfo debug_printf_desc_buffer_info = {};
```
arno-lunarg, line 388: Nit: Can remove embracing `{ /*...*/ }` for this code block

arno-lunarg: (not talking about the `= {};`)

arno-lunarg (approved):

layers/gpu/instrumentation/gpuav_instrumentation.cpp
spencer-lunarg: ya, like this name better

PR #8621, opened 2024-10-01
tests: Fix dgc token offset limit test

ziga-lunarg:


arno-lunarg (approved):

arno-lunarg:
LunarG CI did not catch this PR??

ziga-lunarg:
> LunarG CI did not catch this PR??

I think there might not be a driver in there that has maxIndirectCommandsTokenOffset == UINT32_MAX

spencer-lunarg (approved):

spencer-lunarg:
> I think there might not be a driver in there that has maxIndirectCommandsTokenOffset == UINT32_MAX

There is not, this is only being tested on TestICD and locally on my Lavapipe driver with it... so happy to just go with Github CI here

Issue #8618, opened 2024-09-30
rbtree infinite loop/crash


zmike:
As I pasted to @spencer-lunarg, running this this [piglit](https://gitlab.freedesktop.org/mesa/piglit) test with VVL causes an extended loop inside VVL on shutdown which eventually results in a crash:

 `ZINK_DEBUG=vvl MESA_LOADER_DRIVER_OVERRIDE=zink bin/object-namespace-pollution glBitmap framebuffer -auto -fbo`

spencer-lunarg:
(notes) 

If I turn off everything but Core Checks ( `VK_LAYER_DISABLES=VK_VALIDATION_FEATURE_DISABLE_SHADERS_EXT,VK_VALIDATION_FEATURE_DISABLE_THREAD_SAFETY_EXT,VK_VALIDATION_FEATURE_DISABLE_OBJECT_LIFETIMES_EXT,VK_VALIDATION_FEATURE_DISABLE_UNIQUE_HANDLES_EXT,VK_VALIDATION_FEATURE_DISABLE_API_PARAMETERS_EXT`) it will still crash

If I turn off core checks (`VK_LAYER_DISABLES=VK_VALIDATION_FEATURE_DISABLE_CORE_CHECKS_EXT`) it works as expected... so seems to be isolated to just core checks

I noticed from API dump it goes

```
vkBindBufferMemory()
vkBindBufferMemory()
PIGLIT: {"result": "pass" }
vkCreateImageView()
vkGetPhysicalDeviceImageFormatProperties2KHR()
vkCreateImage()
vkGetImageMemoryRequirements2()
vkAllocateMemory()
vkBindImageMemory()
vkCmdPipelineBarrier2() <---- memory corruption noticed here
```

spencer-lunarg:
seems all `static const std::map` are being corrupted, even in things in like `GetAccessMask2CommonMap()`, not sure yet when that corruption starts

spencer-lunarg:
so Piglet is calling `exit(0)`

```
#0 __run_exit_handlers stdlib/exit.c:108
#1 __GI_exit stdlib/exit.c:138
#2 piglit_report_result piglit/tests/util/piglit-util.c:245-
```

but it is then handling it `piglit-framework-gl.c` via the `destory()` but our `std::map` are being destroyed as a library under us during that `exit(0)` call

spencer-lunarg:
[...]

PR #8639, opened 2024-10-02
layers: Hack to not have piglet exit() destroy maps under us

spencer-lunarg:
Really just an attempt to showcase "a fix" for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8618

### DIFF
--- a/layers/core_checks/cc_device.cpp
+++ b/layers/core_checks/cc_device.cpp
@@ -1023,3 +1023,36 @@ bool CoreChecks::PreCallValidateCreatePipelineBinariesKHR(VkDevice device, const
 
     return skip;
 }
+
+class StaticAllocatedObject {
+  public:
+    StaticAllocatedObject() {
+        printf("--- StaticAllocatedObject() --\n");
+        fflush(stdout);
+    };
+    ~StaticAllocatedObject() {
+        // This is used to detect if the program has called exit() and we need to cleanup the layers. If we don't the app might have
+        // a fucntion in atexit() trying to still make Vulkan calls and if it hits our freed heap memory (such as our std::map) it
+        // will crash and will get a false report it "crashes only with VVL" (Currently this is only required in CoreChecks)
+        exit_called = true;
+        printf("--- ~~~StaticAllocatedObject() --\n");
+        fflush(stdout);
+    }
+    volatile bool exit_called = false;
+};
+
+const StaticAllocatedObject &GetStaticAllocatedObject() {
+    static const StaticAllocatedObject object;
+    return object;
+}
+
+// If the app has called exit() it should (and can) still do cleanup (vkDestory/vkFree) but we need to worry about it trying to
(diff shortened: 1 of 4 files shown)


layers/core_checks/cc_device.cpp
```diff
@@ -1023,3 +1023,36 @@ bool CoreChecks::PreCallValidateCreatePipelineBinariesKHR(VkDevice device, const
...
+        // will crash and will get a false report it "crashes only with VVL" (Currently this is only required in CoreChecks)
+        exit_called = true;
+        printf("--- ~~~StaticAllocatedObject() --\n");
+        fflush(stdout);
+    }
+    volatile bool exit_called = false;
+};
+
+const StaticAllocatedObject &GetStaticAllocatedObject() {
+    static const StaticAllocatedObject object;
```
artem-lunarg, line 1045: One thing to check if this is important for issue's use case. Constructor for static object in function scope runs the first time the function is called. If the function is called the first time after exit() is called then `exit_called` will stay false.

artem-lunarg: I'm a bit a afraid of this kind of stuff, if @jeremyg-lunarg is available to check it would be nice.

jeremyg-lunarg:
Could you also do this with an https://en.cppreference.com/w/c/program/atexit callback?

spencer-lunarg:
> Could you also do this with an https://en.cppreference.com/w/c/program/atexit callback?

the issue is it replays it back in the order they are created, so piglet does it sometime after `vkCreateDevice()`

so it ends looking like

```c++
vvl::CreateDevice(); // VVL hooks into atexit()
PigletSetup(); // piglet hooks into atexit()

// ... app

exit()

PigletDestroy(); // runs their atexit() function
// <----- Crash here
vvl::OnExit(); // We never get to run ours
```

layers/core_checks/cc_device.cpp
```diff
@@ -1023,3 +1023,36 @@ bool CoreChecks::PreCallValidateCreatePipelineBinariesKHR(VkDevice device, const
...
+
+class StaticAllocatedObject {
+  public:
+    StaticAllocatedObject() {
+        printf("--- StaticAllocatedObject() --\n");
+        fflush(stdout);
+    };
+    ~StaticAllocatedObject() {
+        // This is used to detect if the program has called exit() and we need to cleanup the layers. If we don't the app might have
+        // a fucntion in atexit() trying to still make Vulkan calls and if it hits our freed heap memory (such as our std::map) it
```
filnet, line 1035:
```suggestion
        // a function in atexit() trying to still make Vulkan calls and if it hits our freed heap memory (such as our std::map) it
```

layers/core_checks/cc_device.cpp
artem-lunarg: Dll boundaries might create additional difficulties that are not necessarily covered by the standard, just something to pay attention too and to check platform/compiler specific behavior.

Issue #8605, opened 2024-09-26
Missing VK_EXT_device_generated_commands checks


spencer-lunarg:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8604 added most VUs, some (mostly GPU-AV) were not added in the initial release and are tracked here

## GPU-AV 

- [ ] VUID-vkCmdExecuteGeneratedCommandsEXT-* (all common draw VUs)
- [ ] VUID-vkCmdExecuteGeneratedCommandsEXT-preprocessAddress-11142
- [ ] VUID-VkBindIndexBufferIndirectCommandEXT-None-11117
- [ ] VUID-VkBindIndexBufferIndirectCommandEXT-bufferAddress-11118
- [ ] VUID-VkBindIndexBufferIndirectCommandEXT-None-11119
- [ ] VUID-VkBindVertexBufferIndirectCommandEXT-None-11120
- [ ] VUID-VkBindVertexBufferIndirectCommandEXT-None-11121
- [ ] VUID-VkDrawIndirectCountIndirectCommandEXT-None-11122
- [ ] VUID-VkDrawIndirectCountIndirectCommandEXT-None-11123
- [ ] VUID-VkGeneratedCommandsInfoEXT-indirectCommandsLayout-11065
- [ ] VUID-VkGeneratedCommandsInfoEXT-indirectCommandsLayout-11066
- [ ] VUID-VkGeneratedCommandsInfoEXT-sequenceCountAddress-11068

## fine grain sync required

- [ ] VUID-vkUpdateIndirectExecutionSetPipelineEXT-None-11038
- [ ] VUID-VkWriteIndirectExecutionSetPipelineEXT-index-11029
- [ ] VUID-VkWriteIndirectExecutionSetShaderEXT-index-11034
- [ ] VUID-vkCmdExecuteGeneratedCommandsEXT-isPreprocessed-11046
- [ ] VUID-vkCmdExecuteGeneratedCommandsEXT-isPreprocessed-11048
- [ ] VUID-vkCmdExecuteGeneratedCommandsEXT-isPreprocessed-11049
- [ ] VUID-vkCmdExecuteGeneratedCommandsEXT-isPreprocessed-11051
- [ ] VUID-vkCmdExecuteGeneratedCommandsEXT-isPreprocessed-11052
- [ ] VUID-vkCmdExecuteGeneratedCommandsEXT-isPreprocessed-11055 (queue submit update helper)

[...]

PR #8647, opened 2024-10-03
layers: Fix Layout Set Compatible check

spencer-lunarg:
there were some VUs in https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8605 that needed this fix

### DIFF
--- a/layers/state_tracker/pipeline_state.cpp
+++ b/layers/state_tracker/pipeline_state.cpp
@@ -1350,6 +1350,32 @@ std::string LastBound::DescribeNonCompatibleSet(uint32_t set, const vvl::ShaderO
     return ss.str();
 }
 
+bool IsPipelineLayoutSetCompatible(uint32_t set, const vvl::PipelineLayout *a, const vvl::PipelineLayout *b) {
+    if (!a || !b) {
+        return false;
+    }
+    if ((set >= a->set_compat_ids.size()) || (set >= b->set_compat_ids.size())) {
+        return false;
+    }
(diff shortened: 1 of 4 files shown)


layers/state_tracker/pipeline_state.cpp
```diff
@@ -1350,6 +1350,32 @@ std::string LastBound::DescribeNonCompatibleSet(uint32_t set, const vvl::ShaderO
...
 }
 
+bool IsPipelineLayoutSetCompatible(uint32_t set, const vvl::PipelineLayout *a, const vvl::PipelineLayout *b) {
+    if (!a || !b) {
+        return false;
+    }
+    if ((set >= a->set_compat_ids.size()) || (set >= b->set_compat_ids.size())) {
+        return false;
+    }
+    return *(a->set_compat_ids[set]) == *(b->set_compat_ids[set]);
```
spencer-lunarg, line 1360: This was `return a->set_compat_ids[set] == b->set_compat_ids[set];` before which was wrong, it was only checking the shared pointers were different, not that they had different contents

spencer-lunarg: (plenty of Device Generated Commands tests are coming, they were failing for this reason I learned)

artem-lunarg: > This was `return a->set_compat_ids[set] == b->set_compat_ids[set];` before which was wrong, it was only checking the shared pointers were different, not that they had different contents

I have dejavu we fixed this some time ago.

artem-lunarg (approved):

PR #8652, opened 2024-10-04
tests: More WSI cleanup

artem-lunarg:
Some remaining parts. We also need to unify helpers for swapchain image layout transition and create info initialization but probably for another time.


spencer-lunarg (approved):
いいよ！

PR #8653, opened 2024-10-04
gpuav: Create VMA wrappers for DeviceMemoryBlock

spencer-lunarg:
(Broken up into nice commits)

The goal of this was to have `DeviceMemoryBlock` wrap VMA so we can now just go

```c++
    debug_printf_output_buffer.CreateBuffer(loc, &buffer_info, &alloc_info);

    uint32_t *data;
    debug_printf_output_buffer.MapMemory(loc, reinterpret_cast<void **>(&data));
    memset(data, 0, gpuav.gpuav_settings.debug_printf_buffer_size);
    debug_printf_output_buffer.UnmapMemory();
```

without the many (error-prone copy-and-pasted) "is VMA returning VK_SUCCESS" checks

While at it, I had to move a lot of stuff out of `GpuShaderInstrumentor` to the main GPU-AV class object (because now there is no trying to duplicate work for DebugPrintf) ... we can on the future of `GpuShaderInstrumentor` later, but this now leaves it where it is only dealing with "shader instrumentation" and not allocating memory and other things not really related to that scope

### DIFF
--- a/layers/gpu/cmd_validation/gpuav_cmd_validation_common.cpp
+++ b/layers/gpu/cmd_validation/gpuav_cmd_validation_common.cpp
@@ -19,9 +19,9 @@
 
 #include "gpu/core/gpuav.h"
 #include "gpu/core/gpuav_constants.h"
+#include "gpu/resources/gpuav_subclasses.h"
 #include "gpu/shaders/gpuav_shaders_constants.h"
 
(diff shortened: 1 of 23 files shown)


jeremyg-lunarg (approved):
LGTM. The many commits did make this much easier to understand.

arno-lunarg (approved):

Issue #8656, opened 2024-10-06
debugPrintfEXT breaks strangely with 64-bit integer values


buzmeg:
**Environment:**
 - OS: Ubuntu 24.04.1 LTS
 - GPU and driver version: Intel UHD Graphics 620 (WHL GT2)
 - SDK or header version if building from repo:  Vulkan Instance Version: 1.3.290
 - Options enabled (synchronization, best practices, etc.): Validation with Debug Printf preset

```
        apiVersion        = 1.3.274 (4206866)
        driverVersion     = 24.0.9 (100663305)
```

**Describe the Issue**
debugPrintfEXT garbles values after a 64-bit value

The following compute shader code
```
    uint64_t idx_64 = 42;
    uint32_t idx_32 = 42;
    
    if (gl_LocalInvocationIndex == 0) {
        debugPrintfEXT(" BAD: %8ld  %8x  %8x  %8x  %8x\n", idx_64, rc.exclusion_left, rc.exclusion_right, rc.exclusion_top, rc.exclusion_bottom);  // This line is broken
        debugPrintfEXT("OKAY:           %8x  %8x  %8x  %8x  %8ld\n", rc.exclusion_left, rc.exclusion_right, rc.exclusion_top, rc.exclusion_bottom, idx_64);  // This line works
        debugPrintfEXT("GOOD: %8d  %8x  %8x  %8x  %8x\n", idx_32, rc.exclusion_left, rc.exclusion_right, rc.exclusion_top, rc.exclusion_bottom);  // This line works
        debugPrintfEXT("GOOD:           %8x  %8x  %8x  %8x  %8d\n", rc.exclusion_left, rc.exclusion_right, rc.exclusion_top, rc.exclusion_bottom, idx_32);  // This line works
    }

```

Produces the following output:
```
 BAD:       42         0     10000     10001     10002
OKAY:              10000     10001     10002     10003        42
GOOD:       42     10000     10001     10002     10003
GOOD:              10000     10001     10002     10003        42
```

**Expected behavior**
The BAD/OKAY lines should match the GOOD/GOOD lines.

Thanks.

spencer-lunarg:
So this "should" be fixed in the next SDK coming out this week, while making the fixes for [float 64-bit](https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8341) I realized we had issues with how we handled no-32bit values and have tests

... if the SDK doesn't fix it, can look more

[...]

PR #8660, opened 2024-10-07
printf: Add Signed Int64 support

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8656

Adds 64-int support using `%ld` 

We had ability to support it, just were not parsing it correctly (or attempted at all)

### DIFF
--- a/layers/gpu/debug_printf/debug_printf.cpp
+++ b/layers/gpu/debug_printf/debug_printf.cpp
@@ -231,7 +231,6 @@ void AnalyzeAndGenerateMessage(Validator &gpuav, VkCommandBuffer command_buffer,
 
             if (substring.needs_value) {
                 if (substring.is_64_bit) {
-                    assert(substring.type != NumericTypeSint);  // not supported
                     if (substring.type == NumericTypeUint) {
                         std::array<std::string_view, 3> format_strings = {{"%ul", "%lu", "%lx"}};
                         for (const auto &ul_string : format_strings) {
@@ -250,6 +249,28 @@ void AnalyzeAndGenerateMessage(Validator &gpuav, VkCommandBuffer command_buffer,
                         needed = std::snprintf(nullptr, 0, substring.string.c_str(), value) + 1;
                         temp_string.resize(needed);
                         std::snprintf(&temp_string[0], needed, substring.string.c_str(), value);
+                    } else if (substring.type == NumericTypeSint) {
+                        size_t ld_pos = substring.string.find("%ld");
+                        if (ld_pos != std::string::npos) {
+                            substring.string.replace(ld_pos + 1, 2, PRId64);
+                        } else {
+                            gpuav.InternalWarning(command_buffer, loc,
+                                                  "Trying to DebugPrintf a 64-bit signed int but not using \"%%ld\" to print it.");
+                        }
+
+                        const uint32_t *current_ptr = static_cast<uint32_t *>(current_value);
+                        const uint32_t low = *current_ptr;
+                        const uint32_t high = *(current_ptr + 1);
+                        // Need to shift into uint before casting to signed int to avoid undefined behavior
+                        // https://learn.microsoft.com/en-us/cpp/cpp/left-shift-and-right-shift-operators-input-and-output?view=msvc-170#footnotes
+                        const uint64_t value_unsigned = (static_cast<uint64_t>(high) << 32) | low;
+                        const int64_t value = static_cast<int64_t>(value_unsigned);
+
+                        needed = std::snprintf(nullptr, 0, substring.string.c_str(), value) + 1;
+                        temp_string.resize(needed);
+                        std::snprintf(&temp_string[0], needed, substring.string.c_str(), value);
+                    } else {
+                        assert(false);  // non-supported type
                     }
                 } else {
                     if (substring.type == NumericTypeUint) {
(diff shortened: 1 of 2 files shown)


layers/gpu/debug_printf/debug_printf.cpp
```diff
@@ -250,6 +249,25 @@ void AnalyzeAndGenerateMessage(Validator &gpuav, VkCommandBuffer command_buffer,
...
+                            substring.string.replace(ld_pos + 1, 2, PRId64);
+                        } else {
+                            gpuav.InternalWarning(command_buffer, loc,
+                                                  "Trying to DebugPrintf a 64-bit signed int but not using \"%%ld\" to print it.");
+                        }
+
+                        const uint32_t *current_ptr = static_cast<uint32_t *>(current_value);
+                        const uint32_t low = *current_ptr;
+                        const uint32_t high = *(current_ptr + 1);
+                        const int64_t value = (static_cast<int64_t>(high) << 32) | low;
```
artem-lunarg, line 264: Is it possible to use uin64_t here? From what I read it can be undefined behavior (that's how I understood https://learn.microsoft.com/en-us/cpp/cpp/left-shift-and-right-shift-operators-input-and-output?view=msvc-170#footnotes).

arno-lunarg: Why not just go:
```cpp
const int64_t *i64_ptr = static_cast<int64_t *>(current_value);
```

artem-lunarg: > Why not just go:
> 
> ```c++
> const auto i64_ptr = static_cast<int64_t *>(current_value);
> ```

This makes assumption that machine uses little endian endianness. In the current code low/high values are stored in little-endian fashion (low part then high part), but then `const int64_t value` initialization will do proper thing independent of the machine endianess.

arno-lunarg: The endianness problem is also here when we read back the other types right? 
We shoud read bytes, not DWORDS, and reorder by hand? Otherwise I don't see why reading 2 `uint32_t` and assembling a 64 bits value from that is any more valid than straight up reading a 64 bits value

It looks like this code has been written with little endian in mind, and would break if this was not the case

arno-lunarg: yes, but don't we already make this assumption?

artem-lunarg: where?

artem-lunarg: I don't know how `current_ptr` data is stored. I assumed it's according to some protocol, so some code explicitly stored at first low 32 bit then high and it is decoded that way. But if 64 bit value was just written directly, then you are right.

arno-lunarg: > I don't know how current_ptr data is stored. I assumed it's according to some protocol, so some code explicitly stored at first low 32 bit then high and it is decoded that way. But if 64 bit value was just written directly, then you are right. I obviously don't know details of this code, please ignore my comments if I missed something.

Not sure either. Spencer, how does a 64 bits value as seen in a shader gets written to printf buffer?

arno-lunarg: Looks like we write data 32 bits at a time, so yes we should read 2 DWORDS to get that 64 bits int

spencer-lunarg: we write it like

```c++
bool debugPrintf(uint metaData, uint word, uint moreWords) {
    out_buffer[0] = metaData;
    out_buffer[1] = word;
    out_buffer[2] = moreWords;
}
```

layers/gpu/debug_printf/debug_printf.cpp
```diff
@@ -250,6 +249,25 @@ void AnalyzeAndGenerateMessage(Validator &gpuav, VkCommandBuffer command_buffer,
                         needed = std::snprintf(nullptr, 0, substring.string.c_str(), value) + 1;
                         temp_string.resize(needed);
                         std::snprintf(&temp_string[0], needed, substring.string.c_str(), value);
+                    } else if (substring.type == NumericTypeSint) {
+                        size_t ld_pos = substring.string.find("%ld");
+                        if (ld_pos != std::string::npos) {
+                            substring.string.replace(ld_pos + 1, 2, PRId64);
```
artem-lunarg, line 255: With MSVC compiler `#define PRId64       "lld"` which is 3 symbols.

artem-lunarg (approved):
LGTM with unsigned cast as we discussed

layers/gpu/debug_printf/debug_printf.cpp
spencer-lunarg: added and left a comment to the undefined behavior note/link from above

layers/gpu/debug_printf/debug_printf.cpp
artem-lunarg: Forgot about this part. Do we need to preserve string length or is it fine to replace 2 chars with 3 chars if necessary?

artem-lunarg: My initial concern was that replace can  override something by replacing 2 chars with 3, but it looks like I did not know how replace works.

PR #8661, opened 2024-10-07
gpu: Cleanup in DeviceMemoryBlock

arno-lunarg:
- no public member in DeviceMemoryBlock
 => either everything public or nothing
- MapMemory simply returns a void* pointer
 => Just as the test framework. This pointer is immediately cast to appropriate type. Thought about using a template method to specify the type, but a simple manual cast is probably better here

### DIFF
--- a/layers/gpu/descriptor_validation/gpuav_descriptor_set.cpp
+++ b/layers/gpu/descriptor_validation/gpuav_descriptor_set.cpp
@@ -57,8 +57,7 @@ VkDeviceAddress DescriptorSet::GetLayoutState(Validator &gpuav, const Location &
     alloc_info.requiredFlags = VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT;
     layout_.CreateBuffer(loc, &buffer_info, &alloc_info);
 
-    glsl::BindingLayout *layout_data;
-    layout_.MapMemory(loc, reinterpret_cast<void **>(&layout_data));
+    auto layout_data = (glsl::BindingLayout *)layout_.MapMemory(loc);
     memset(layout_data, 0, static_cast<size_t>(buffer_info.size));
 
     layout_data[0].count = num_bindings;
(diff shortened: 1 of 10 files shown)


layers/gpu/descriptor_validation/gpuav_descriptor_set.cpp
```diff
@@ -252,8 +251,7 @@ std::shared_ptr<DescriptorSet::State> DescriptorSet::GetCurrentState(Validator &
     alloc_info.requiredFlags = VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_CACHED_BIT;
     next_state->buffer.CreateBuffer(loc, &buffer_info, &alloc_info);
 
-    glsl::DescriptorState *data{nullptr};
-    next_state->buffer.MapMemory(loc, reinterpret_cast<void **>(&data));
```
arno-lunarg, line 256: `buffer` is a bad name here, but I do not understand things enough to be able to change it. @spencer-lunarg or @jeremyg-lunarg any suggestion?

spencer-lunarg: its the `DeviceMemoryBlock` used by the descriptor validation here... so it consists of a Buffer and Memory... `next_state->memory_block`?

arno-lunarg: memory_block is just as vague IMO 
what does this buffer holds? The name should stem from that

spencer-lunarg (approved):

layers/gpu/descriptor_validation/gpuav_descriptor_set.cpp
spencer-lunarg: it holds the `DescriptorState` data (hense why we are mapping the memory above it)

it is a mapping, so for `Uniform Buffers` it prints out the size of the buffer, `Texel Buffer` it has the range, `Samplers` it holds the sampler ID, etc

spencer-lunarg: we call this all `struct State {` inside `class DescriptorSet` and this is just the memory backing it... I guess unless you want to call it `DescriptorSet::State::memory_backing_the_state` I don't have a good idea and throwing my hands up with "naming sucks" unfortunately

PR #8667, opened 2024-10-07
scripts: Don't reformat json on gen

mikes-lunarg:
Update json.in files as strings instead of round-tripping through the python json module. The old way implicitly formats the file and adds a lot of vertical whitespace that makes these hard to edit by hand.


spencer-lunarg (approved):
This is smart, I like this

PR #8669, opened 2024-10-08
gpu: Fix abort if using Vulkan 1.0

arno-lunarg:


layers/gpu/core/gpuav_setup.cpp
```diff
@@ -219,6 +219,11 @@ void Validator::PreCallRecordCreateDevice(VkPhysicalDevice physicalDevice, const
                                           const RecordObject &record_obj, vku::safe_VkDeviceCreateInfo *modified_create_info) {
     BaseClass::PreCallRecordCreateDevice(physicalDevice, pCreateInfo, pAllocator, pDevice, record_obj, modified_create_info);
 
+    // If using 1.0, GPU-AV will error inside Validator::PostCallRecordCreateDevice
```
spencer-lunarg, line 222: This is in `PreCallRecordCreateDevice`

What is crashing here? We have the other check for this reason in PostCall

spencer-lunarg: ok, then we should re-comment these

```c++
// in PreCallRecordCreateDevice

// GPU-AV not supported, exit early or will potentially create invalid Vulkan
 if (api_version < VK_API_VERSION_1_1) {

---------


// in PosCallRecordCreateDevice

// GPU-AV not supported, exit early or it will error inside Validator::PostCallRecordCreateDevice
 if (api_version < VK_API_VERSION_1_1) {
 ```

**Review by spencer-lunarg — APPROVED** [review]

File: layers/gpu/core/gpuav_setup.cpp [hunk 2]
```diff
@@ -219,6 +219,12 @@ void Validator::PreCallRecordCreateDevice(VkPhysicalDevice physicalDevice, const
                                           const RecordObject &record_obj, vku::safe_VkDeviceCreateInfo *modified_create_info) {
     BaseClass::PreCallRecordCreateDevice(physicalDevice, pCreateInfo, pAllocator, pDevice, record_obj, modified_create_info);
 
+    // If using 1.0, GPU-AV will error inside Validator::PostCallRecordCreateDevice,
```
spencer-lunarg, line 222:
```suggestion
    // GPU-AV not supported, exit early or will potentially create invalid Vulkan.
```

layers/gpu/core/gpuav_setup.cpp
```diff
@@ -375,6 +381,12 @@ void Validator::PreCallRecordCreateDevice(VkPhysicalDevice physicalDevice, const
 
 // Perform initializations that can be done at Create Device time.
 void Validator::PostCreateDevice(const VkDeviceCreateInfo *pCreateInfo, const Location &loc) {
+    // GPU-AV requires at least Vulkan 1.1
```
spencer-lunarg, line 384:
```suggestion
    // GPU-AV not supported, exit early or it will error inside Validator::PostCallRecordCreateDevice
```

PR #8670, opened 2024-10-08
printf: Add deferred RT pipeline test

arno-lunarg:
Also misc cleanup I did while chasing a bug in debug printf that turned out to be an output buffer too small

### DIFF
--- a/layers/gpu/debug_printf/debug_printf.cpp
+++ b/layers/gpu/debug_printf/debug_printf.cpp
@@ -193,14 +193,14 @@ struct OutputRecord {
 void AnalyzeAndGenerateMessage(Validator &gpuav, VkCommandBuffer command_buffer, VkQueue queue,
                                gpuav::DebugPrintfBufferInfo &buffer_info, uint32_t *const debug_output_buffer,
                                const Location &loc) {
-    uint32_t output_record_counts = debug_output_buffer[gpuav::kDebugPrintfOutputBufferSize];
-    if (!output_record_counts) return;
+    uint32_t output_buffer_dwords_counts = debug_output_buffer[gpuav::kDebugPrintfOutputBufferDWordsCount];
+    if (!output_buffer_dwords_counts) return;
 
-    uint32_t index = gpuav::kDebugPrintfOutputBufferData;  // get first OutputRecord index
-    while (debug_output_buffer[index]) {
+    uint32_t output_record_i = gpuav::kDebugPrintfOutputBufferData;  // get first OutputRecord index
+    while (debug_output_buffer[output_record_i]) {
         std::stringstream shader_message;
 
-        OutputRecord *debug_record = reinterpret_cast<OutputRecord *>(&debug_output_buffer[index]);
+        OutputRecord *debug_record = reinterpret_cast<OutputRecord *>(&debug_output_buffer[output_record_i]);
         // Lookup the VkShaderModule handle and SPIR-V code used to create the shader, using the unique shader ID value returned
         // by the instrumented shader.
         const gpuav::GpuAssistedShaderTracker *tracker_info = nullptr;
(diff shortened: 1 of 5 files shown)


layers/gpu/spirv/instruction.cpp
```diff
@@ -290,7 +291,6 @@ void GenerateInstructions(const vvl::span<const uint32_t>& spirv, std::vector<In
         auto new_insn = instructions.emplace_back(it, instruction_count++);
         it += new_insn.Length();
     }
-    instructions.shrink_to_fit();
```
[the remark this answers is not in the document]
arno-lunarg: Also this function could probably return a std::vector instead of expecting an empty one and then modifying it. I expect compilers to not make unnecessary deep copy

layers/gpu/debug_printf/debug_printf.cpp
```diff
@@ -193,47 +193,48 @@ struct OutputRecord {
...
-    while (debug_output_buffer[index]) {
+    uint32_t output_record_i = gpuav::kDebugPrintfOutputBufferData;  // get first OutputRecord index
+    while (debug_output_buffer[output_record_i]) {
         std::stringstream shader_message;
 
-        OutputRecord *debug_record = reinterpret_cast<OutputRecord *>(&debug_output_buffer[index]);
+        OutputRecord *debug_record = reinterpret_cast<OutputRecord *>(&debug_output_buffer[output_record_i]);
         // Lookup the VkShaderModule handle and SPIR-V code used to create the shader, using the unique shader ID value returned
         // by the instrumented shader.
-        const gpuav::GpuAssistedShaderTracker *tracker_info = nullptr;
```
spencer-lunarg, line 206: I liked `tracker` we use `shader` in a million places to mean the SPIR-V or `VkShaderEXT`, the `GpuAssistedShaderTracker` was designed to bundle pipeline, shader object, SPIR-V etc... `shader_tracker_info` works, but don't want yet-another `shader_something` in our code if we don't have to

layers/gpu/debug_printf/debug_printf.cpp
```diff
@@ -193,47 +193,48 @@ struct OutputRecord {
...
 
         std::vector<gpuav::spirv::Instruction> instructions;
-        gpuav::spirv::GenerateInstructions(tracker_info->instrumented_spirv, instructions);
+        gpuav::spirv::GenerateInstructions(shader_info->instrumented_spirv, instructions);
 
         // Search through the shader source for the printf format string for this invocation
         const std::string format_string = FindFormatString(instructions, debug_record->format_string_id);
         // Break the format string into strings with 1 or 0 value
         auto format_substrings = ParseFormatString(format_string);
+        assert(!format_substrings.empty());
```
spencer-lunarg, line 225: lets report a warning and return, don't want just another random assert here that won't get seen in release builds

arno-lunarg: actually I just need to remove that I think, printing an empty string should be ok, and lead to an empty format_substrings

spencer-lunarg:
```c++
if (format_substrings.empty());
     gpuav.InternalWarning(queue, loc, "Can't parse any formatted strings from: " + format_string);
     return;
}
```

spencer-lunarg: > printing an empty string should be ok, and lead to an empty format_substrings

we have `NegativeDebugPrintf, MisformattedEmptyString` so you can test yourself

layers/gpu/debug_printf/debug_printf.cpp
arno-lunarg: I went back to `tracker_info`

layers/gpu/spirv/instruction.cpp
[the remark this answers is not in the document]
spencer-lunarg: so last thing will point out is `instructions.reserve(spirv.size() * 4);` allocates way too much memory, if each instruction is average 4 DWORds then the vector will be 4 times bigger, but not sure this is a bottleneck to spend time finding a clever way to do this

spencer-lunarg (approved):

PR #8672, opened 2024-10-09
tests: Initialize buffer with data

artem-lunarg:
Need to write some tests that that need buffers with specific data.


tests/framework/binding.h
```diff
@@ -586,6 +586,11 @@ class Buffer : public internal::NonDispHandle<VkBuffer> {
              &allocate_flag_info);
     }
 
+    explicit Buffer(const Device &dev, VkBufferUsageFlags usage, const void *data, size_t data_size,
```
spencer-lunarg, line 589: is there an example in this PR using this?

spencer-lunarg (approved):

tests/framework/binding.h
artem-lunarg: Going to use it in the next PR!

PR #8681, opened 2024-10-10
tests: GPU-AV tests not checking if can run

spencer-lunarg:
These tests were asserting on my local TestICD, but never should have been running


tests/unit/ray_tracing.cpp
```diff
@@ -3312,6 +3312,9 @@ TEST_F(NegativeRayTracing, DynamicRayTracingPipelineStack) {
     AddRequiredFeature(vkt::Feature::accelerationStructure);
     AddRequiredFeature(vkt::Feature::bufferDeviceAddress);
     RETURN_IF_SKIP(InitFrameworkForRayTracingTest());
+    if (IsPlatformMockICD()) {
```
spencer-lunarg, line 3315: I was hitting this

```
if (!std::align(rt_pipeline_props.shaderGroupBaseAlignment, rt_pipeline_props.shaderGroupHandleSize, sbt_buffer_ptr,
                sbt_buffer_space_left)) {
    assert(false);
    return;
}
```

but not consistently, but for short term just marking this as it is being tested on real hardware

arno-lunarg (approved):

Issue #8678, opened 2024-10-10
Differentiate between NV and EXT DGC features


ShabbyX:
Context: https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6933

Like BDA, the EXT version of GDC uses the same feature names as the NV version, but changes the semantics.

VVL needs to differentiate between these features now. That starts with `scripts/generators/device_features_generator.py` needing this change:

```
        self.identical_but_different_features = {
            'VkPhysicalDeviceBufferDeviceAddressFeaturesEXT': 'EXT',
+           'VkPhysicalDeviceDeviceGeneratedCommandsFeaturesNV': 'NV',
        }
```

and autogenerate takes care of everything except the existing VVL checks for NV and EXT version of the extension. The NV ones would have to be changed to check the NV-specific feature.

spencer-lunarg:
I am like 10000000% sure I added this to `device_features_generator.py` ... but seems we have 2 lists, I added it to `feature_requirements.py` only

PR #8683, opened 2024-10-10
scripts: Add deviceGeneratedCommands alias

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8678

### DIFF
--- a/layers/vulkan/generated/device_features.h
+++ b/layers/vulkan/generated/device_features.h
@@ -200,9 +200,11 @@ struct DeviceFeatures {
     // VkPhysicalDeviceDeviceGeneratedCommandsComputeFeaturesNV
     bool deviceGeneratedComputePipelines;
     // VkPhysicalDeviceDeviceGeneratedCommandsFeaturesEXT
-    bool dynamicGeneratedPipelineLayout;
-    // VkPhysicalDeviceDeviceGeneratedCommandsFeaturesEXT, VkPhysicalDeviceDeviceGeneratedCommandsFeaturesNV
     bool deviceGeneratedCommands;
+    // VkPhysicalDeviceDeviceGeneratedCommandsFeaturesEXT
+    bool dynamicGeneratedPipelineLayout;
+    // VkPhysicalDeviceDeviceGeneratedCommandsFeaturesNV
+    bool deviceGeneratedCommandsNV;
     // VkPhysicalDeviceDeviceMemoryReportFeaturesEXT
     bool deviceMemoryReport;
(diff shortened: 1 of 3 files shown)


layers/vulkan/generated/device_features.h
```diff
@@ -200,9 +200,11 @@ struct DeviceFeatures {
     // VkPhysicalDeviceDeviceGeneratedCommandsComputeFeaturesNV
     bool deviceGeneratedComputePipelines;
     // VkPhysicalDeviceDeviceGeneratedCommandsFeaturesEXT
-    bool dynamicGeneratedPipelineLayout;
-    // VkPhysicalDeviceDeviceGeneratedCommandsFeaturesEXT, VkPhysicalDeviceDeviceGeneratedCommandsFeaturesNV
     bool deviceGeneratedCommands;
+    // VkPhysicalDeviceDeviceGeneratedCommandsFeaturesEXT
```
artem-lunarg, line 204: nit: `// VkPhysicalDeviceDeviceGeneratedCommandsFeaturesEXT` goes after `deviceGeneratedCommands`.

spencer-lunarg: so this is now (broken out to see)

```c++
    // VkPhysicalDeviceDeviceGeneratedCommandsFeaturesEXT
    bool deviceGeneratedCommands;
    // VkPhysicalDeviceDeviceGeneratedCommandsFeaturesEXT
    bool dynamicGeneratedPipelineLayout;
 
    // VkPhysicalDeviceDeviceGeneratedCommandsFeaturesNV
    bool deviceGeneratedCommandsNV;
 ```
 
 note the EXT has 2 feature bits (`dynamicGeneratedPipelineLayout`) so that is why it looks like it both above and below it

Comment (artem-lunarg) [reply layers/vulkan/generated/device_features.h:204]: yeah, could be confused diff...

File: scripts/generators/device_features_generator.py [hunk 2]
```diff
@@ -33,6 +33,7 @@ def __init__(self):
         # semantically different.  They are given a suffix to be distinguished.
         self.identical_but_different_features = {
             'VkPhysicalDeviceBufferDeviceAddressFeaturesEXT': 'EXT',
+            'VkPhysicalDeviceDeviceGeneratedCommandsFeaturesNV': 'NV',
```
ShabbyX, line 36: To avoid future mistakes, maybe pull this out to somewhere shared between this and `feature_requirements.py`. I hear mesh shader promotion has the same problem and some upcoming ray tracing extension too, but I haven't seen evidence to it.

Issue #8684, opened 2024-10-10
Stop enabling all features in AddRequiredFeature


spencer-lunarg:
We have many tests that we have to go

```c++
AddRequiredFeature(vkt::Feature::meshShader);
AddDisabledFeature(vkt::Feature::multiviewMeshShader);
AddDisabledFeature(vkt::Feature::primitiveFragmentShadingRateMeshShader);
```

because when enable a single feature for `VkPhysicalDeviceMeshShaderFeaturesEXT` it will expose anything else supported... they should be defaulted to disabled

This makes it very easy for someone new to not realize features are being enabled from underneath them

arno-lunarg:
I am all for this change, but my hunch is that many tests will break because of this

spencer-lunarg:
>  but my hunch is that many tests will break because of this

The majority of feature structs have only 1 feature, some will break, but should be easy to catch/fix

PR #8687, opened 2024-10-12
tests: Only enable required features

arno-lunarg:
Let's blow up CI on a weekend! (closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8684)

### DIFF
--- a/layers/gpu/core/gpuav_setup.cpp
+++ b/layers/gpu/core/gpuav_setup.cpp
@@ -224,18 +224,25 @@ void Validator::PreCallRecordCreateDevice(VkPhysicalDevice physicalDevice, const
     VkPhysicalDeviceFeatures supported_features{};
     DispatchGetPhysicalDeviceFeatures(physicalDevice, &supported_features);
 
-    if (auto enabled_features = const_cast<VkPhysicalDeviceFeatures *>(modified_create_info->pEnabledFeatures)) {
+    VkPhysicalDeviceFeatures *enabled_features = const_cast<VkPhysicalDeviceFeatures *>(modified_create_info->pEnabledFeatures);
+    if (!enabled_features) {
+        // Look in a potential VkPhysicalDeviceFeatures2 feature struct
+        if (auto *enabled_features_2 = const_cast<VkPhysicalDeviceFeatures2 *>(
+                vku::FindStructInPNextChain<VkPhysicalDeviceFeatures2>(modified_create_info->pNext))) {
+            enabled_features = &enabled_features_2->features;
+        }
+    }
+    if (enabled_features) {
         if (supported_features.fragmentStoresAndAtomics && !enabled_features->fragmentStoresAndAtomics) {
-            InternalWarning(device, record_obj.location, "Forcing VkPhysicalDeviceFeatures::fragmentStoresAndAtomics to VK_TRUE");
+            InternalWarning(device, record_obj.location, "Forcing fragmentStoresAndAtomics to VK_TRUE");
             enabled_features->fragmentStoresAndAtomics = VK_TRUE;
         }
         if (supported_features.vertexPipelineStoresAndAtomics && !enabled_features->vertexPipelineStoresAndAtomics) {
-            InternalWarning(device, record_obj.location,
-                            "Forcing VkPhysicalDeviceFeatures::vertexPipelineStoresAndAtomics to VK_TRUE");
+            InternalWarning(device, record_obj.location, "Forcing vertexPipelineStoresAndAtomics to VK_TRUE");
             enabled_features->vertexPipelineStoresAndAtomics = VK_TRUE;
         }
         if (supported_features.shaderInt64 && !enabled_features->shaderInt64) {
-            InternalWarning(device, record_obj.location, "Forcing VkPhysicalDeviceFeatures::shaderInt64 to VK_TRUE");
+            InternalWarning(device, record_obj.location, "Forcing shaderInt64 to VK_TRUE");
             enabled_features->shaderInt64 = VK_TRUE;
         }
     }
--- a/tests/framework/feature_requirements.h
+++ b/tests/framework/feature_requirements.h
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 2023 Valve Corporation
- * Copyright (c) 2023 LunarG, Inc.
+ * Copyright (c) 2024 Valve Corporation
+ * Copyright (c) 2024 LunarG, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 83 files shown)


layers/gpu/core/gpuav_setup.cpp
```diff
@@ -224,7 +224,15 @@ void Validator::PreCallRecordCreateDevice(VkPhysicalDevice physicalDevice, const
     VkPhysicalDeviceFeatures supported_features{};
     DispatchGetPhysicalDeviceFeatures(physicalDevice, &supported_features);
 
-    if (auto enabled_features = const_cast<VkPhysicalDeviceFeatures *>(modified_create_info->pEnabledFeatures)) {
+    VkPhysicalDeviceFeatures *enabled_features = const_cast<VkPhysicalDeviceFeatures *>(modified_create_info->pEnabledFeatures);
+    if (!enabled_features) {
+        // Look in a potential VkPhysicalDeviceFeatures2 feature struct
+        if (auto *enabled_features_2 = const_cast<VkPhysicalDeviceFeatures2 *>(
```
spencer-lunarg, line 230: this will make the `"Forcing VkPhysicalDeviceFeatures::fragmentStoresAndAtomics to VK_TRUE"` probably more confusing... ugh, happy for now though, better something than nothing

tests/framework/feature_requirements.h
```diff
@@ -23,23 +23,20 @@ class FeatureRequirements {
     ~FeatureRequirements();
     // Add a feature required for the test to be executed
     void AddRequiredFeature(APIVersion api_version, vkt::Feature feature);
-    // Add a feature that will be disabled when creating the device
-    void AddDisabledFeature(APIVersion api_version, vkt::Feature feature);
```
spencer-lunarg, line 27: we will 100% still need this, we currently just pass in the core 1.0 `VkPhysicalDeviceFeatures` so we need a way to disable those features... so maybe call it a different name (or just change the comment) but this is most likely still required

tests/unit/shader_compute.cpp
```diff
@@ -684,7 +684,7 @@ TEST_F(NegativeShaderCompute, LocalSizeIdExecutionModeMaintenance5) {
     SetTargetApiVersion(VK_API_VERSION_1_3);
     AddRequiredExtensions(VK_KHR_MAINTENANCE_5_EXTENSION_NAME);
     AddRequiredFeature(vkt::Feature::maintenance5);
-    AddDisabledFeature(vkt::Feature::maintenance4);
```
spencer-lunarg, line 687: this test is testing

> VUID-RuntimeSpirv-LocalSizeId-06434
>
> If Execution Mode LocalSizeId is used, maintenance4 must be enabled

when using 1.3, so we need to manually disable this test for 1.3, so likely will not work

tests/framework/feature_requirements.h
spencer-lunarg: I think instead of doing this "try and find-and-replace" everything, I would just focus on 

`multiviewMeshShader` and `multiviewGeometryShader` as those are key examples of the problem I want to  solve

after that, we can spend time trying to remove other `AddDisabledFeature` that are "no-ops", but doing this way will just be a lot more work to slowly figure out each test failing CI

arno-lunarg: No feature but the one mentioned in the `AddRequiredFeature` will be enabled, including those in the original `VkPhysicalDeviceFeatures`

layers/gpu/core/gpuav_setup.cpp
arno-lunarg: Oh yes, I removed `VkPhysicalDeviceFeatures::`
Also we were talking about that not long ago, and you told me that looking in a potential `VkPhysicalDeviceFeatures2` was not needed. A test in `DebugPrintf` broke after disabling default features, making me realize that either 1) turns out we need to look for this `VkPhysicalDeviceFeatures2` or 2) there is something wrong happening before that, so we don't end up having a `VkPhysicalDeviceFeatures` (the original) but it is not there

tests/unit/shader_compute.cpp
arno-lunarg: maintenance4 will still be disabled here, so it should be fine?

layers/core_checks/cc_render_pass.cpp
```diff
@@ -3089,7 +3089,7 @@ bool CoreChecks::ValidateBeginRenderingFragmentDensityMap(VkCommandBuffer comman
         if (rendering_info.pDepthAttachment && (rendering_info.pDepthAttachment->imageView != VK_NULL_HANDLE)) {
             auto depth_view_state = Get<vvl::ImageView>(rendering_info.pDepthAttachment->imageView);
             if (depth_view_state && !(depth_view_state->image_state->create_info.flags & VK_IMAGE_CREATE_SUBSAMPLED_BIT_EXT)) {
-                const LogObjectList objlist(commandBuffer, rendering_info.pStencilAttachment->imageView);
+                const LogObjectList objlist(commandBuffer, rendering_info.pDepthAttachment->imageView);
```
arno-lunarg, line 3092: wow was not expecting to catch this one x)

tests/framework/feature_requirements.h
spencer-lunarg: I see, that is "ideal" but can see from the TestICD runs a LOT of old tests will need to have their `VkPhysicalDeviceFeatures` features enabled now, which isn't necessarily a bad thing, just more work to do all at once

layers/core_checks/cc_render_pass.cpp
spencer-lunarg: can you just quick make this as a 1 line fix PR (and it will just rebase), rather just get this in and not have to wait until we pass CI for all these tests

tests/framework/render.h
```diff
@@ -186,7 +184,15 @@ class VkRenderFramework : public VkTestFramework {
     VkInstance instance_;
     VkPhysicalDevice gpu_ = VK_NULL_HANDLE;
     VkPhysicalDeviceProperties physDevProps_;
-    vkt::FeatureRequirements feature_requirements_;
+    // This set of required features is used for the features query.
+    // If any required feature is not available, test will fail
+    vkt::FeatureRequirements required_features_;
+    // This is the set of features that will be enable.
```
filnet, line 190:
```suggestion
    // This is the set of features that will be enabled.
```

tests/unit/dynamic_state_positive.cpp
```diff
@@ -409,6 +409,8 @@ TEST_F(PositiveDynamicState, SetDepthBias2EXTDepthBiasClampEnabled) {
     pipe.AddDynamicState(VK_DYNAMIC_STATE_DEPTH_BIAS);
     VkPipelineRasterizationStateCreateInfo raster_state = vku::InitStructHelper();
     raster_state.depthBiasEnable = VK_TRUE;
+    raster_state.lineWidth = 1.0f;
+    ;
```
filnet, line 413: ?

arno-lunarg: !

spencer-lunarg (approved):
Went through and all changes look good, just will need to keep adding missing features until CI is happy

PR #8694, opened 2024-10-14
layers: Header cleanup

spencer-lunarg:
Remove more `#include` headers that are just including the world everywhere

### DIFF
--- a/layers/best_practices/best_practices_validation.h
+++ b/layers/best_practices/best_practices_validation.h
@@ -25,6 +25,7 @@
 #include <string>
 #include <deque>
 #include <chrono>
+#include <set>
 
 static const uint32_t kMemoryObjectWarningLimit = 250;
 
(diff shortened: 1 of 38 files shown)


layers/best_practices/best_practices_validation.h
```diff
@@ -25,6 +25,7 @@
 #include <string>
 #include <deque>
 #include <chrono>
+#include <set>
```
artem-lunarg, line 28: Does this change remove includes that are already included by some other headers? It looks like `std::set` is used in this file.

artem-lunarg: In general it's not recommended to rely on include dependencies and include what is used (can depend on compiler), but not sure what clang tool is actually doing.

artem-lunarg (approved):
LGTM assuming it's CI issue

layers/best_practices/best_practices_validation.h
spencer-lunarg: as you might have found out, the `custom_containers.h` wasn't hiding `<set>`/`<map>` in the `#ifdef USE_ROBIN_HOOD_HASHING` else check

spencer-lunarg:
CI is failing for Video test (unrelated) and tracking it here https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8696

PR #8701, opened 2024-10-15
gpu: Descriptor Indexing cleanup

spencer-lunarg:
In efforts to create a post processing output that is independent, went in and had to clean up some of the descriptor indexing. Was able to replace things like `DescriptorSet::State` with our new `DeviceMemoryBlock` class

### DIFF
--- a/layers/gpu/descriptor_validation/gpuav_descriptor_set.cpp
+++ b/layers/gpu/descriptor_validation/gpuav_descriptor_set.cpp
@@ -38,7 +38,7 @@ DescriptorSet::DescriptorSet(const VkDescriptorSet handle, vvl::DescriptorPool *
     : vvl::DescriptorSet(handle, pool, layout, variable_count, state_data), layout_(*static_cast<Validator *>(state_data)) {}
 
 DescriptorSet::~DescriptorSet() {
-    Destroy();
+    last_used_block_.reset();
     layout_.DestroyBuffer();
 }
 
@@ -217,11 +217,12 @@ void FillBindingInData(const vvl::InlineUniformBinding &binding, glsl::Descripto
 
 std::shared_ptr<DescriptorSet::State> DescriptorSet::GetCurrentState(Validator &gpuav, const Location &loc) {
     auto guard = Lock();
-    uint32_t cur_version = current_version_.load();
-    if (last_used_state_ && last_used_state_->version == cur_version) {
-        return last_used_state_;
+    uint32_t current_version = current_version_.load();
+    if (last_used_block_ && last_used_version_ == current_version) {
+        return last_used_block_;
     }
-    auto next_state = std::make_shared<State>(VkHandle(), cur_version, gpuav);
+    last_used_version_ = current_version;
+    auto next_block = std::make_shared<State>(gpuav);
 
     uint32_t descriptor_count = 0;  // Number of descriptors, including all array elements
     if (GetBindingCount() > 0) {
@@ -237,8 +238,8 @@ std::shared_ptr<DescriptorSet::State> DescriptorSet::GetCurrentState(Validator &
     }
     if (descriptor_count == 0) {
         // no descriptors case, return a dummy state object
-        last_used_state_ = next_state;
-        return last_used_state_;
+        last_used_block_ = next_block;
+        return last_used_block_;
     }
 
     VkBufferCreateInfo buffer_info = vku::InitStruct<VkBufferCreateInfo>();
@@ -249,9 +250,9 @@ std::shared_ptr<DescriptorSet::State> DescriptorSet::GetCurrentState(Validator &
     // and manually flushing it at the end of the state updates is faster than using HOST_COHERENT.
     VmaAllocationCreateInfo alloc_info{};
     alloc_info.requiredFlags = VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_CACHED_BIT;
-    next_state->buffer.CreateBuffer(loc, &buffer_info, &alloc_info);
+    next_block->buffer.CreateBuffer(loc, &buffer_info, &alloc_info);
 
-    auto data = (glsl::DescriptorState *)next_state->buffer.MapMemory(loc);
+    auto data = (glsl::DescriptorState *)next_block->buffer.MapMemory(loc);
 
     uint32_t index = 0;
     for (uint32_t i = 0; i < bindings_.size(); i++) {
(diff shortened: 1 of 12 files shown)


layers/gpu/descriptor_validation/gpuav_descriptor_set.cpp
```diff
@@ -287,20 +299,18 @@ std::shared_ptr<DescriptorSet::State> DescriptorSet::GetCurrentState(Validator &
...
+std::shared_ptr<DeviceMemoryBlock> DescriptorSet::GetPostProcessBuffer(Validator &gpuav, const Location &loc) {
     auto guard = Lock();
-    uint32_t cur_version = current_version_.load();
-    if (output_state_) {
-        return output_state_;
+    if (post_process_block_) {
+        return post_process_block_;
     }
-    auto next_state = std::make_shared<State>(VkHandle(), cur_version, gpuav);
+    post_process_block_ = std::make_shared<DeviceMemoryBlock>(gpuav);
```
arno-lunarg, line 313: If I follow thins somewhat correctly, this GPU buffer is essentially used like a map, indexed by a descriptor index, and storing... something 
Is this buffer mapping to `DescriptorSetOutData` (seen in `bindless_descriptor.comp`)?
More generally, the names of the buffer bindings appearing in the validation shaders should just map to the actual buffers as seen in C++ code, unless the same buffer is used in different ways in a shader 
I see a lot of `data[]`  in the shaders too, would be really nice to get rename those too : 3

layers/gpu/resources/gpuav_shader_resources.h
```diff
@@ -28,14 +28,12 @@
 namespace gpuav {
 
 struct DescSetState {
-    uint32_t num = 0;
     std::shared_ptr<DescriptorSet> state = {};
-    BindingVariableMap binding_req = {};
+    BindingVariableMap binding_req_map = {};
```
arno-lunarg, line 32: BindingVariableMap is defined as:
```cpp
// < binding index (of descriptor set) : meta data >
typedef std::unordered_multimap<uint32_t, DescriptorRequirement> BindingVariableMap;
```
So I sense that it should be renamed to something like `DescSetBindingToRequirementsMap`. With an `s` at `Requirement` to hint at the fact that this is a multi map

layers/gpu/descriptor_validation/gpuav_descriptor_set.cpp
```diff
@@ -249,9 +261,9 @@ std::shared_ptr<DescriptorSet::State> DescriptorSet::GetCurrentState(Validator &
     // and manually flushing it at the end of the state updates is faster than using HOST_COHERENT.
     VmaAllocationCreateInfo alloc_info{};
     alloc_info.requiredFlags = VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_CACHED_BIT;
-    next_state->buffer.CreateBuffer(loc, &buffer_info, &alloc_info);
+    input_block_.CreateBuffer(loc, &buffer_info, &alloc_info);
```
arno-lunarg, line 264: same comment as `post_process_block_`

arno-lunarg (approved):
W
But some things could use even more love!

layers/gpu/resources/gpuav_shader_resources.h
spencer-lunarg: I did this to match what we did in core validation, I also don't want to use `DescSet` more than I need (we sprinkle that everywhere at a very inconsistent matter)

Things like this need a lot more work because you need to touch 5+ year old core validation code as well

... basically happy to change `BindingVariableMap`, but want more time to think what to rename it before having to re-re-rename it because it wasn't good ... also I which there was comment telling me "why" it is a  multi map

layers/gpu/descriptor_validation/gpuav_descriptor_set.cpp
spencer-lunarg: the problem is the indiretions

so there is 

```
// glsl
struct DescriptorSetRecord {
    DescriptorLayoutData layout_data;
    DescriptorSetInData in_data;
    DescriptorSetOutData out_data;
};

// c++
struct DescriptorSetRecord {
    VkDeviceAddress layout_data;
    VkDeviceAddress in_data;
    VkDeviceAddress out_data;
};
```

the `post_process_block_`  is the buffer holding/backing the memory in 

```
layout(buffer_reference, buffer_reference_align = 8, std430) buffer DescriptorSetOutData {
    uint data[];
};
```

which is used like

```
        DescriptorSetOutData out_data = bindless_state_buffer.desc_sets[desc_set].out_data;
        out_data.data[state_index] = kDescriptorSetWrittenMask | desc_set;
 ```

Comment (spencer-lunarg) [reply layers/gpu/descriptor_validation/gpuav_descriptor_set.cpp:313]: ok, so basically after this change my goal is to rip things apart

1. The "output" is going to be a separate Post Process Pass and not part of this
2. having a "layout" and "input" name is awful, both are "input" but one is an indirect LUT

File: layers/gpu/descriptor_validation/gpuav_descriptor_validation.cpp [hunk 4]
```diff
@@ -136,32 +135,64 @@ void UpdateBoundDescriptors(Validator &gpuav, CommandBuffer &cb_state, VkPipelin
...
+                bindless_state->desc_sets[i].out_data = set_buffer.post_process_buffer->Address();
             }
         }
         cmd_info.bindless_state.UnmapMemory();
     }
     return true;
 }
+}  // namespace descriptor
+
+static std::map<uint32_t, std::vector<uint32_t>> UsedDescriptors(const Location &loc, const DeviceMemoryBlock &layout,
```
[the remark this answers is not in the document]
spencer-lunarg: my goal was to just move this out of being nested in a class, next PR will alter what this is actually going to do and consume and will add comments!

layers/gpu/descriptor_validation/gpuav_descriptor_set.cpp
arno-lunarg: Looking at:
```
DescriptorSetOutData out_data = bindless_state_buffer.desc_sets[desc_set].out_data;
out_data.data[state_index] = kDescriptorSetWrittenMask | desc_set;
```
it seems that this is the only thing we write as an output, or at least the only reference I found to `DescriptorSetOutData`. So looks like we could be more specific, and instead of `out_data.data`, rename to something like 
`written_descriptors`, if it is indeed the only thing we store?

PR #8704, opened 2024-10-16
Add a common spirv_logging.cpp helper

spencer-lunarg:
I recently added ShaderDebugInfo in error message for GPU-AV https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8691

There are **many** normal core validation error messages that also should have source code printed in the error message

I have a follow-up PR to do it for some VUs, but want to get this in first which is just splitting the GPU-AV logging code for SPIR-V into a dedicated `spirv_logging` helper file

### DIFF
--- a/layers/gpu/instrumentation/gpuav_shader_instrumentor.h
+++ b/layers/gpu/instrumentation/gpuav_shader_instrumentor.h
@@ -18,10 +18,16 @@
 
 #include "containers/custom_containers.h"
 #include "generated/chassis.h"
+#include "state_tracker/shader_instruction.h"
 #include "state_tracker/state_tracker.h"
-#include "gpu/spirv/instruction.h"
 
 #include <vector>
+
+// There is a spirv::Instruction used for normal validation.
+// There is a gpuav::spirv::Instruction that is ONLY intended for shader instrumentation (designed so we can build the shader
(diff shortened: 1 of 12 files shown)


layers/gpu/instrumentation/gpuav_shader_instrumentor.h
```diff
@@ -18,11 +18,17 @@
...
 #include "state_tracker/state_tracker.h"
-#include "gpu/spirv/instruction.h"
 
 #include <vector>
 
+// There is a spirv::Instruction used for normal validation.
+// There is a gpuav::spirv::Instruction that is ONLY intended for shader instrumentation (designed so we can build the shader
+// instrumentation as a seperate library). For logging GPU-AV will want to make use of the normal validaiton instruction class, just
+// alias it with "Instruction" as that name shouldn't collide with anything.
+using Instruction = ::spirv::Instruction;
```
spencer-lunarg, line 30: basically either

1. Have a single `Instruction` class which turns into a build dependency mess when I just want the `./tests/spirv/instrumentation` executable (so can't do that)
2. make "normal" validation `vvl::spirv::Instrumentation` which I tried first and it ended up making a LOT of churn
3. Do this as the need to use the `spirv` namespace is just for logging as seen by the fact only `GenerateDebugInfoMessage` was changed

artem-lunarg (approved):

PR #8705, opened 2024-10-16
Update deps enhancements

MathiasMagnus:
This PR contains fixes to two annoyances of ours, both arguably bug candidates:

- Dependency fetching breaks when the remote URL is changed in `update_deps.py`, which can easily be handled by wiping the clone dir.
- Both documentation and code suggests that the `XYZ_INSTALL_DIR` variables can be used to override where dependencies should be taken from, but the dep fetching machinery outright overwrites all user-provided paths.

### DIFF
--- a/scripts/update_deps.py
+++ b/scripts/update_deps.py
@@ -422,8 +422,15 @@ class GoodRepo(object):
         if VERBOSE:
             print('Checking out {n} in {d}'.format(n=self.name, d=self.repo_dir))
 
-        if self._args.do_clean_repo:
+        if os.path.exists(os.path.join(self.repo_dir, '.git')):
+            url_changed = command_output(['git', 'config', '--get', 'remote.origin.url'], self.repo_dir).strip() != self.url
+        else:
+            url_changed = False
+
+        if self._args.do_clean_repo or url_changed:
             if os.path.isdir(self.repo_dir):
+                if VERBOSE:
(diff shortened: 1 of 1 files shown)


MathiasMagnus:
By the way, has LunarG considered externalizing all the build related utilities to a dedicated repository? This way it would be easier to propagate changes to all the repositories, but the biggest gain would be to have a dedicated place where the utility scripts could be tested (for eg. if the repo URL is changed, can one rerun `update_deps.py`? Other ecosystems have profited greatly from such externalized collection of helpers, such as [rocm-cmake](https://github.com/ROCm/rocm-cmake) that all ROCm components use to build in a uniform style.

spencer-lunarg (approved):
LGTM 

@charles-lunarg any objections?

something to be aware of is unfortunately this `update_deps.py` is a example of copy-and-paste gone wrong and you may want to upstream fixes everywhere else

https://github.com/KhronosGroup/Vulkan-Loader/pull/1469#issuecomment-2057769741

spencer-lunarg:
> By the way, has LunarG considered externalizing all the build related utilities to a dedicated repository? 

Yes, we have considered and very aware of the problem, I think the issue is just having someone spend the engineering hours going around and updating things and making sure they break everyone as they are doing it

charles-lunarg (approved):

Issue #8689, opened 2024-10-14
Wrong error detection on vkFreeCommandBuffers


asuessenbach:
**Environment:**
 - OS: Win10
 - GPU and driver version: NVIDIA RTX A3000 Laptop GPU, 560.76
 - SDK or header version if building from repo: 1.3.296.0
 - Options enabled (synchronization, best practices, etc.): none

**Describe the Issue**

With this code sequence in a sample project
```
	vkEndCommandBuffer(commandBuffer);

	const VkSubmitInfo submitInfo = {
		.sType = VK_STRUCTURE_TYPE_SUBMIT_INFO,
		.pNext = nullptr,
		.waitSemaphoreCount = 0,
		.pWaitSemaphores = nullptr,
		.pWaitDstStageMask = nullptr,
		.commandBufferCount = 1,
		.pCommandBuffers = &commandBuffer,
		.signalSemaphoreCount = 0,
		.pSignalSemaphores = nullptr
	};

	vkQueueSubmit(vkDev.graphicsQueue, 1, &submitInfo, VK_NULL_HANDLE);
	vkQueueWaitIdle(vkDev.graphicsQueue);

	vkFreeCommandBuffers(vkDev.device, vkDev.commandPool, 1, &commandBuffer);
```
I get this VVL error message:
```
Validation Error: [ VUID-vkFreeCommandBuffers-pCommandBuffers-00047 ] Object 0: handle = 0x23b2a8996b8, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xcad092000000000d, type = VK_OBJECT_TYPE_COMMAND_POOL; | MessageID = 0x1ab902fc | vkFreeCommandBuffers(): pCommandBuffers[0] (VkCommandBuffer 0x23b2a8996b8[]) is in use.
The Vulkan spec states: All elements of pCommandBuffers must not be in the pending state (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkFreeCommandBuffers-pCommandBuffers-00047)
```
The mentioned command buffer and command pool are the ones used in that call to `vkFreeCommandBuffers`.

**Expected behavior**

No error message or warning, as after the call to `vkQueueWaitIdle`, the `commandBuffer` should not be in the pending state.

artem-lunarg:
[...]

PR #8713, opened 2024-10-17
layers: Break dependency of in-use tracking on GPU state

artem-lunarg:
Inspired by https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8689 which was confusing and time consuming.

The idea is that in-use tracking is not a GPU side system. Evaluate resource in-use status based on CPU side API usage pattern. Make GPU/driver errors a separate axis of error handling.

The problem with mixing these two types of errors handling is that it leads to confusion in practice. For example, if the application is correctly structured to handle resource lifetimes or image layouts transition, but it did poor job with handling VkResult, and the error message reports that resource is still in use (but we just called vkDeviceWaitIdle!) or layout is incorrect when all transitions are properly specified. This misleading messages can focus attention on the wrong area (like spending few days debugging multithreaded system trying to understand why the resource is not tracked properly).

If necessary we might provide an option to detect non successful VkResult checks (have this already?) for critical places like submit/wait calls.

For now this change is experimental to observe how it works in practice.

### DIFF
--- a/layers/state_tracker/state_tracker.cpp
+++ b/layers/state_tracker/state_tracker.cpp
@@ -1589,15 +1589,12 @@ void ValidationStateTracker::PostCallRecordGetDeviceQueue2(VkDevice device, cons
 }
 
 void ValidationStateTracker::PostCallRecordQueueWaitIdle(VkQueue queue, const RecordObject &record_obj) {
-    if (VK_SUCCESS != record_obj.result) return;
     if (auto queue_state = Get<vvl::Queue>(queue)) {
         queue_state->NotifyAndWait(record_obj.location);
     }
 }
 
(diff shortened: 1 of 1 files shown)


spencer-lunarg (approved):
good thought, these are showing work on the GPU status, and we are tracking if the queue is done with work, not if it was "successful" at the work

artem-lunarg:
> good thought, these are showing work on the GPU status, and we are tracking if the queue is done with work, not if it was "successful" at the work

Yes, I can say even more lose guarantee. Checks that you used API correctly to ensure safe resource access. If there is GPU error it’s a different problem, resource can still be in use by the GPU. Fix those problems with the help of GPU-AV :) then the original code will indeed guarantee resource is safe to access.

PR #8715, opened 2024-10-17
gpu: Add PostProcessDescriptorIndexPass

spencer-lunarg:
This takes a

```glsl
vec4 color = texture(s, vec3(0.0));
```

and turns it into the following to be used by itself (for things like Sync Val)

```glsl
inst_post_process_descriptor_index(0u, 0u, 0u); // (set, binding, index)
color = texture(s, vec3(0.0));
```

then if we have "normal all of GPU-AV turned on" it looks like

```glsl
bool check = inst_bindless_descriptor(28u, uvec4(4u, floatBitsToUint(gl_FragCoord).xy, 0u), 0u, 0u, 0u, 0u);
vec4 color;
if (check) {
    inst_post_process_descriptor_index(0u, 0u, 0u); // (set, binding, index)
    color = texture(s, vec3(0.0));
} else {
    color = vec4(0.0);
}
```

so either way, we are marking which descriptor is used


layers/gpu/spirv/post_process_descriptor_indexing.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+#include "generated/instrumentation_post_process_descriptor_index_comp.h"
+#include "gpu/shaders/gpuav_shaders_constants.h"
+
+namespace gpuav {
+namespace spirv {
+
+PostProcessDescriptorIndexingPass::PostProcessDescriptorIndexingPass(Module& module) : Pass(module) { module.use_bda_ = true; }
+
+static LinkInfo link_info = {instrumentation_post_process_descriptor_index_comp,
```
arno-lunarg, line 28: move this variable to `GetLinkFunctionId`

layers/gpu/spirv/post_process_descriptor_indexing.cpp
```diff
@@ -0,0 +1,191 @@
...
+    const uint32_t function_result = module_.TakeNextId();
+    const uint32_t function_def = GetLinkFunctionId();
+    const uint32_t void_type = module_.type_manager_.GetTypeVoid().Id();
+
+    block.CreateInstruction(
+        spv::OpFunctionCall,
+        {void_type, function_result, function_def, set_constant.Id(), binding_constant.Id(), descriptor_index_id}, inst_it);
+}
+
+bool PostProcessDescriptorIndexingPass::AnalyzeInstruction(const Function& function, const Instruction& inst) {
```
arno-lunarg, line 58: Would rename this function to `InstrumentInstruction`, since that is what happens here if I am correct.
That make the code code block below clearer:
```cpp
 if (!InstrumentInstruction(*function, *(inst_it->get()))) continue;
                instrumented_count_++;
```
I also see that `AnalyzeInstruction` is already being used in spirv tools, for a different job, so not sure conflating names is a good idea

layers/gpu/spirv/post_process_descriptor_indexing.cpp
```diff
@@ -0,0 +1,191 @@
...
+    descriptor_index_id_ = 0;
+}
+
+bool PostProcessDescriptorIndexingPass::Run() {
+    for (const auto& function : module_.functions_) {
+        for (auto block_it = function->blocks_.begin(); block_it != function->blocks_.end(); ++block_it) {
+            auto& block_instructions = (*block_it)->instructions_;
+            for (auto inst_it = block_instructions.begin(); inst_it != block_instructions.end(); ++inst_it) {
+                if (!AnalyzeInstruction(*function, *(inst_it->get()))) continue;
+                instrumented_count_++;
```
arno-lunarg, line 175:
```suggestion
                instrumentations_count_++;
```

layers/gpu/shaders/instrumentation/post_process_descriptor_index.comp
```diff
@@ -0,0 +1,78 @@
...
+    // BindingLayout[0]
+    uint num_bindings;
+    uint pad; // always zero, used to keep things aligned
+
+    // BindingLayout[1] - BindingLayout[N]
+    // struct glsl::BindingLayout {
+    //   x: count
+    //   y: state_start
+    // }
+    uvec2 data[];
```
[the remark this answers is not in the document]
spencer-lunarg: 1. just copy-and-pasting it for this PR
2. this is going to be gone with https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8714

layers/gpu/spirv/post_process_descriptor_indexing.cpp
spencer-lunarg: agree, but this is how we do it for the other passes, would like to fix all at once afterwards (as this PR is being consistent)

layers/gpu/spirv/post_process_descriptor_indexing.cpp
spencer-lunarg: agree, but this is how we do it for the other passes, would like to fix all at once afterwards (as this PR is being consistent)

layers/gpu/spirv/post_process_descriptor_indexing.cpp
spencer-lunarg: no, this function has one job, decide **if** a given instruction needs to be modified, it itself does no instrumentation

layers/gpu/shaders/instrumentation/post_process_descriptor_index.comp
```diff
@@ -0,0 +1,78 @@
...
+layout(buffer_reference, buffer_reference_align = 8, std430) buffer DescriptorSetOutData {
+    // size of descriptor count (including all array elements)
+    // Used to mark which indexes were accessed
+    uint data[];
+};
+
+struct DescriptorSetRecord {
+    DescriptorLayoutData layout_data; // TODO - remove with LUT send into instrumentation of shader
+    DescriptorSetInData in_data; // unused
+    DescriptorSetOutData out_data;
```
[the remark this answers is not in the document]
spencer-lunarg: again, copy and pasting what we had to get this PR to work, 100% want to fix this, trying to make small, manageable PR,  and will be part of https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8714

layers/gpu/spirv/post_process_descriptor_indexing.cpp
arno-lunarg: Then probably rename to `<Instruction>NeedsIntrumentation` or anything that better captures the intent, since the task is precisely defined

arno-lunarg (approved):

spencer-lunarg:
Going to make a PR for renaming today to address all valid comments

PR #8718, opened 2024-10-17
gpu: Cleanup DescriptorSet class

spencer-lunarg:
This just removes `struct DescSetState` which was just a wrapper around the `gpu::DescriptorSet` class, so this just moves that logic into `gpu::DescriptorSet` so that we have a single interface

### DIFF
--- a/layers/gpu/descriptor_validation/gpuav_descriptor_validation.cpp
+++ b/layers/gpu/descriptor_validation/gpuav_descriptor_validation.cpp
@@ -54,7 +54,7 @@ void UpdateBoundPipeline(Validator &gpuav, CommandBuffer &cb_state, VkPipelineBi
                     // TODO - Hit crash running with Dota2, this shouldn't happen, need to look into
                     continue;
                 }
-                descriptor_set_buffers[update_index++].binding_req_map = slot->second;
+                descriptor_set_buffers[update_index++]->binding_req_map = slot->second;
             }
         }
     }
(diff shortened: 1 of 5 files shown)


layers/gpu/descriptor_validation/gpuav_descriptor_validation.cpp
```diff
@@ -102,26 +102,25 @@ void UpdateBoundDescriptors(Validator &gpuav, CommandBuffer &cb_state, VkPipelin
             continue;
         }
 
-        DescSetState desc_set_state;
-        desc_set_state.state = std::static_pointer_cast<DescriptorSet>(last_bound_set.bound_descriptor_set);
-        bindless_state->desc_sets[i].layout_data = desc_set_state.state->GetLayoutAddress(gpuav, loc);
+        std::shared_ptr<DescriptorSet> ds_state = std::static_pointer_cast<DescriptorSet>(last_bound_set.bound_descriptor_set);
```
spencer-lunarg, line 105: the main goal was to only have a single `descriptor set state` variable and it is the State Object itself

layers/gpu/descriptor_validation/gpuav_descriptor_validation.cpp
```diff
@@ -135,34 +134,35 @@ void UpdateBoundDescriptors(Validator &gpuav, CommandBuffer &cb_state, VkPipelin
...
         }
         cmd_info.bindless_state.UnmapMemory();
     }
     return true;
 }
 }  // namespace descriptor
 
-static std::map<uint32_t, std::vector<uint32_t>> UsedDescriptors(const Location &loc, const DeviceMemoryBlock &layout,
-                                                                 const DeviceMemoryBlock &output_state, uint32_t shader_set) {
+static std::map<uint32_t, std::vector<uint32_t>> UsedDescriptors(Validator &gpuav, const Location &loc, DescriptorSet &ds_state,
```
spencer-lunarg, line 152: this function is still going to be updated, just an intermediate step

layers/gpu/resources/gpuav_shader_resources.h
```diff
@@ -27,20 +27,11 @@
 
 namespace gpuav {
 
-struct DescSetState {
```
spencer-lunarg, line 30: main goal of PR

PR #8725, opened 2024-10-18
layers: Validate monotonic increase of timeline values

artem-lunarg:
This introduces new type of submit validation (will see how it flies) and it is done and reported by the Queue threads. This should be taken into account by the tests. During QueueSubmit call no error might be reported when validation is performed by the Queue thread. `VerifyFound` should be put after device/queue `Wait` which guarantees that Queue thread finished its job.

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -126,6 +126,8 @@ vvl_sources = [
   "layers/core_checks/cc_spirv.cpp",
   "layers/core_checks/cc_state_tracker.cpp",
   "layers/core_checks/cc_state_tracker.h",
+  "layers/core_checks/cc_submit.cpp",
+  "layers/core_checks/cc_submit.h",
   "layers/core_checks/cc_synchronization.cpp",
   "layers/core_checks/cc_video.cpp",
   "layers/core_checks/cc_vuid_maps.cpp",
--- a/layers/core_checks/cc_state_tracker.h
+++ b/layers/core_checks/cc_state_tracker.h
@@ -18,8 +18,10 @@
  * limitations under the License.
  */
 #pragma once
+#include "cc_submit.h"
 #include "state_tracker/state_tracker.h"
 #include "state_tracker/cmd_buffer_state.h"
+#include "state_tracker/queue_state.h"
 
 namespace core {
 
@@ -33,6 +35,20 @@ class CommandBuffer : public vvl::CommandBuffer {
 
     void RecordWaitEvents(vvl::Func command, uint32_t eventCount, const VkEvent* pEvents,
                           VkPipelineStageFlags2KHR src_stage_mask) override;
+};
+
+// Override Retire to validate submissions in the order defined by synchronization
+class Queue : public vvl::Queue {
+  public:
+    Queue(ValidationStateTracker& dev_data, VkQueue handle, uint32_t family_index, uint32_t queue_index,
+          VkDeviceQueueCreateFlags flags, const VkQueueFamilyProperties& queue_family_properties,
+          const ValidationObject& error_logger);
+
+  private:
+    void Retire(vvl::QueueSubmission&) override;
+
+  private:
+    QueueSubmissionValidator queue_submission_validator_;
 };
 
 }  // namespace core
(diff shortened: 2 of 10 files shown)


layers/state_tracker/state_tracker.cpp
```diff
@@ -1180,7 +1180,7 @@ void ValidationStateTracker::PreCallRecordQueueSubmit(VkQueue queue, uint32_t su
     }
     // Now process each individual submit
     for (uint32_t submit_idx = 0; submit_idx < submitCount; submit_idx++) {
-        Location submit_loc = record_obj.location.dot(vvl::Field::pSubmits, submit_idx);
+        Location submit_loc = record_obj.location.dot(vvl::Struct::VkSubmitInfo, vvl::Field::pSubmits, submit_idx);
```
artem-lunarg, line 1183: Added Struct to make `GetQueueSubmitVUID` work.

BUILD.gn
```diff
@@ -126,6 +126,8 @@ vvl_sources = [
   "layers/core_checks/cc_spirv.cpp",
   "layers/core_checks/cc_state_tracker.cpp",
   "layers/core_checks/cc_state_tracker.h",
+  "layers/core_checks/cc_submit_time_validation.cpp",
```
spencer-lunarg, line 129:
```suggestion
  "layers/core_checks/cc_submit_time.cpp",
```

I feel it gets slippery to start calling any file in the Validation Layers as `validation`

layers/core_checks/cc_submit_time_validation.cpp
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ */
+
+#include "cc_submit_time_validation.h"
+#include "state_tracker/queue_state.h"
+#include "sync/sync_vuid_maps.h"
+#include "generated/chassis.h"  // ValidationObject
+
+static Location GetSignaledSemaphoreLocation(const Location& submit_loc, uint32_t index) {
+    if (submit_loc.function == vvl::Func::vkQueueSubmit2) {
```
spencer-lunarg, line 24: Watch up for `vvl::Func::vkQueueSubmit2KHR`... I would just flip the logic around

```suggestion
    if (submit_loc.function != vvl::Func::vkQueueSubmit) {
```

or 

`auto field = (submit_loc.function == vvl::Func::vkQueueSubmit) ? vvl::Field::pSignalSemaphores : vvl::Field::pSignalSemaphoreInfos`

artem-lunarg: pSignalSemaphores also work for sparse. Maybe make this more explicit.

BUILD.gn
artem-lunarg: cc_submit

artem-lunarg: Updated

layers/core_checks/cc_submit_time_validation.cpp
artem-lunarg: Updated

layers/core_checks/cc_state_tracker.h
```diff
@@ -35,4 +37,17 @@ class CommandBuffer : public vvl::CommandBuffer {
                           VkPipelineStageFlags2KHR src_stage_mask) override;
 };
 
-}  // namespace core
\ No newline at end of file
+class Queue : public vvl::Queue {
```
spencer-lunarg, line 40: worth adding some comment here explaining why we have a `vvl` and `core` Queue object (until there is a day where every object base is `vvl` and `core` has it's own copy)

layers/core_checks/cc_submit.cpp
```diff
@@ -0,0 +1,47 @@
...
+    } else {
+        assert(false && "Unhandled signaling function");
+    }
+    return submit_loc.dot(field, index);
+}
+
+void QueueSubmissionValidator::Validate(const vvl::QueueSubmission& submission) const {
+    for (uint32_t i = 0; i < (uint32_t)submission.signal_semaphores.size(); ++i) {
+        const auto& signal = submission.signal_semaphores[i];
+        const Location signal_semaphore_loc = GetSignaledSemaphoreLocation(submission.loc.Get(), i);
```
spencer-lunarg, line 38: move into the `if (signal.payload < current_payload) {` so only called if there is an error to print

spencer-lunarg (approved):
with last few nits

layers/core_checks/cc_submit.cpp
artem-lunarg: good catch

layers/core_checks/cc_state_tracker.h
artem-lunarg: Done

PR #8726, opened 2024-10-18
gpu: Better naming for Descriptor Indexing shaders

spencer-lunarg:
```patch
struct DescriptorSetRecord {
-   DescriptorLayoutData layout_data;
-   DescriptorSetInData in_data;
-   DescriptorSetOutData out_data; 

+   DescriptorIndexLUT descriptor_index_lut;
+   DescriptorSetType ds_type;
+   DescriptorIndexPostProcess descriptor_index_post_process; 
};
```

note in the future once I split things with https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8714 then there will be no `DescriptorIndexLUT`


spencer-lunarg:
@arno-lunarg - hope this is satisfactory!

layers/gpu/shaders/instrumentation/bindless_descriptor.comp
```diff
@@ -43,15 +43,15 @@ layout(buffer_reference, buffer_reference_align = 8, std430) buffer DescriptorLa
     uvec2 data[];
 };
 
-layout(buffer_reference, buffer_reference_align = 8, std430) buffer DescriptorSetInData {
+layout(buffer_reference, buffer_reference_align = 8, std430) buffer DescriptorSetType {
     // struct glsl::DescriptorState {
     //   x: id
     //   y: extra data depending on the descriptor type
     // }
     uvec2 data[];
```
arno-lunarg, line 51: Did you forget to use a struct here or are you planning on doing it later?

layers/gpu/shaders/instrumentation/bindless_descriptor.comp
```diff
@@ -110,7 +110,7 @@ bool inst_bindless_descriptor(const uint inst_num, const uvec4 stage_info, const
         }
         if (desc_type == kImageSamplerDesc) {
             // check sampler
-            desc_id = in_data.data[state_index].y;
+            desc_id = ds_type.data[state_index].y;
```
arno-lunarg, line 113: From the description of `DescriptorSetInData`, `id` is stored in `.x`. So having `desc_id` coming from `.y` seems weird. Is everything ok here?

layers/gpu/shaders/instrumentation/bindless_descriptor.comp
```diff
@@ -61,9 +61,9 @@ layout(buffer_reference, buffer_reference_align = 8, std430) buffer GlobalState
 };
 
 struct DescriptorSetRecord {
-    DescriptorLayoutData layout_data;
-    DescriptorSetInData in_data;
-    DescriptorSetOutData out_data; // unused
+    DescriptorIndexLUT descriptor_index_lut;
+    DescriptorSetType ds_type;
```
arno-lunarg, line 65: What do you think of adding `<In/Out/Inout>Ptr` as a suffix for pointer types? It would add meaningful info, and would help with the lack of `*` in glsl to denote pointers. And having `In/Out/InOut`  helps with understanding what will be done with those pointers. Good for everyone.

arno-lunarg (commented):
Still need to get rid of **all** the remaining `data[]` in the shaders, but that is a net improvement!

arno-lunarg (approved):

layers/gpu/shaders/instrumentation/bindless_descriptor.comp
spencer-lunarg: planning later, this is purely a naming change, no semantics differences

layers/gpu/shaders/instrumentation/bindless_descriptor.comp
spencer-lunarg: so the idea is we are packing 2 things

1. Which descriptor type it is (sampler, texel buffer, image, etc)
2. Any meta data for it ... which we currently use only 1 DWORD but we need more then that, for Stroage Images I need both the width, height, depth, array layer.... so this whole design needs to be updated

layers/gpu/shaders/instrumentation/bindless_descriptor.comp
spencer-lunarg: like the idea, but the issue is having 

```
layout(buffer_reference, buffer_reference_align = 8, std430) buffer DescriptorIndexLUTInPtr {
    uvec2 data[];
};
```

seems more confusing because this struct type now is a Ptr?

spencer-lunarg:
the naming is an on-going thing, this is an improvement so merging, happy to always keep improving it afterwards

layers/gpu/shaders/instrumentation/bindless_descriptor.comp
arno-lunarg: Well what you point at is a pointer type right? I don't see the issue
When you say "this struct", you mean `DescriptorIndexLUTInPtr` ?

PR #8740, opened 2024-10-22
Fix checking image use for images in VkReleaseSwapchainImagesInfoEXT

ziga-lunarg:


layers/core_checks/cc_wsi.cpp
```diff
@@ -1043,7 +1043,7 @@ bool CoreChecks::PreCallValidateReleaseSwapchainImagesEXT(VkDevice device, const
                              pReleaseInfo->pImageIndices[i]);
         }
 
-        if (swapchain_state->images[i].image_state->InUse()) {
+        if (swapchain_state->images[pReleaseInfo->pImageIndices[i]].image_state->InUse()) {
```
spencer-lunarg, line 1046: (not required, but while here) want to just create a `const uint32_t image_indices = pReleaseInfo->pImageIndices[i]` at top of loop and use that everywhere

also if we fail `if (pReleaseInfo->pImageIndices[i] >= swapchain_state->images.size()) {` this will be an OOB, so we need to do a `continue;`  after the `VUID-VkReleaseSwapchainImagesInfoEXT-pImageIndices-07785` VU

spencer-lunarg (approved):

artem-lunarg (approved):

tests/unit/wsi_positive.cpp
```diff
@@ -1581,3 +1581,82 @@ TEST_F(PositiveWsi, DifferentPerPresentModeImageCount) {
...
+    const uint32_t imageCount = 4;
+
+    if (info.surface_capabilities.maxImageCount < imageCount) {
+        GTEST_SKIP() << "Test maxImageCount to be at least 3";
+    }
+    InitSwapchainInfo();
+
+    VkSwapchainCreateInfoKHR swapchain_create_info = vku::InitStructHelper();
+    swapchain_create_info.surface = m_surface;
+    swapchain_create_info.minImageCount = imageCount;
```
spencer-lunarg, line 1610: From CI failing

> Validation Error: [ VUID-VkSwapchainCreateInfoKHR-presentMode-02839 ] | MessageID = 0xfb70ccdb | vkCreateSwapchainKHR(): pCreateInfo->minImageCount 4, which is outside the bounds returned by vkGetPhysicalDeviceSurfaceCapabilitiesKHR() (i.e. minImageCount = 6, maxImageCount = 64).

so I think just need to query test for `minImageCount`/`maxImageCount`

ziga-lunarg:
New test fails on CI on android, but passes for me locally

spencer-lunarg:
> New test fails on CI on android, but passes for me locally

failing on `m_swapchain.initialized()` is strange, let me build and test locally on my pixel 8

spencer-lunarg:
CI failed on Android devices due old Android version

The test work fine with newer Android versions of same device, so added to blacklist

PR #8741, opened 2024-10-22
Track releasing swapchain images

ziga-lunarg:


layers/state_tracker/image_state.cpp
```diff
@@ -584,6 +584,15 @@ void Swapchain::PresentImage(uint32_t image_index, uint64_t present_id) {
     }
 }
 
+void Swapchain::ReleaseImage(uint32_t image_index) {
+    if (image_index >= images.size()) return;
```
artem-lunarg, line 588: Assuming this function is called in PostRecord index has to be correct, assert can be better, otherwise can hide bugs.
Probably we have similar check for image index in other places, I never had to a chance to investigate/fix that.

artem-lunarg: Indeed other functions also start with this check. Feel free to ignore comment, maybe it's better to update this all together later.

artem-lunarg (approved):

PR #8736, opened 2024-10-22
Fix semaphore use with VkSwapchainPresentFenceInfoEXT 

ziga-lunarg:
Closes #8578

Issue #8578, opened 2024-09-22
Semaphores in VkPresentInfoKHR not handled correctly when used with VkSwapchainPresentFenceInfoEXT 


ziga-lunarg:
Destroying a semaphore that was used in `VkPresentInfoKHR::pWaitSemaphores`, when one of the fences in `VkSwapchainPresentFenceInfoEXT::pFences` already signaled, will report a false positive `VUID-vkDestroySemaphore-semaphore-05149`, unless all of the fences are waited on.

The relevant part of the spec under `VkSwapchainPresentFenceInfoEXT` says: 
> The application can destroy the wait semaphores associated with a given presentation operation when at least one of the associated fences is signaled

Test to reproduce:

```
TEST_F(PositiveWsi, WaitSemaphoreAndPresentFence) {
    TEST_DESCRIPTION("vkQueuePresent waits on a semaphore and signals a fence");
    AddSurfaceExtension();
    AddRequiredExtensions(VK_EXT_SURFACE_MAINTENANCE_1_EXTENSION_NAME);
    AddRequiredExtensions(VK_EXT_SWAPCHAIN_MAINTENANCE_1_EXTENSION_NAME);
    RETURN_IF_SKIP(Init());
    RETURN_IF_SKIP(InitSwapchain());

    constexpr uint32_t swapchain_count = 2u;

    SurfaceContext surface_contexts[swapchain_count];
    VkSurfaceKHR surfaces[swapchain_count];
    VkSwapchainKHR swapchains[swapchain_count] = {};
    for (uint32_t i = 0u; i < swapchain_count; ++i) {
        CreateSurface(surface_contexts[i], surfaces[i]);
        CreateSwapchain(surfaces[i], VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT, VK_SURFACE_TRANSFORM_IDENTITY_BIT_KHR, swapchains[i]);
    }

    vkt::Semaphore acquire_semaphores[swapchain_count];
    vkt::Semaphore submit_semaphores[swapchain_count];
    std::vector<VkImage> swapchain_images[swapchain_count];
    for (uint32_t i = 0u; i < swapchain_count; ++i) {
        acquire_semaphores[i] = vkt::Semaphore(*m_device);
        submit_semaphores[i] = vkt::Semaphore(*m_device);
        swapchain_images[i] = GetSwapchainImages(swapchains[i]);
    }

    m_command_buffer.begin();

[...]

PR #8744, opened 2024-10-23
layers: Fix present fence does not retire present semaphores

artem-lunarg:
Fixes the root cause of issue https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8578 and is a candidate to replace PR https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8736

Thanks @ziga-lunarg for providing test and alternative solution, it helped to understand the problem. The test in this PR is a bit modified comparing to the original one, but the original test also passes.

The previous iterations related to presentation fences:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8400
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8454


ziga-lunarg (approved):
Fixes the CTS tests I was looking at

artem-lunarg:
Need to fix Android issue.

PR #8747, opened 2024-10-24
tests: Remove incorrect event test

artem-lunarg:
This test failed for me on CI so I removed it. The comment in the test suggests it failed before too, so less troubles for everyone.

That's correct but not a real reason. It's incorrect to do this according: `VUID-vkSetEvent-event-09543`

Seemingly we don't have VUID-vkSetEvent-event-09543 implemented fully.

more spec notes:
> If a command buffer is waiting for an event to be signaled from the host, the application must signal the event before submitting the command buffer, as described in the [queue forward progress](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#commandbuffers-submission-progress) section.

> Since the original users of this functionality were highly limited and platform-specific, this functionality is now considered defunct and should not be used.

### DIFF
--- a/tests/unit/sync_object_positive.cpp
+++ b/tests/unit/sync_object_positive.cpp
@@ -1119,28 +1119,6 @@ TEST_F(PositiveSyncObject, BasicSetAndWaitEvent2) {
     m_device->Wait();
 }
 
-TEST_F(PositiveSyncObject, WaitEventThenSet) {
-#if defined(VVL_ENABLE_TSAN)
-    // NOTE: This test in particular has failed sporadically on CI when TSAN is enabled.
-    GTEST_SKIP() << "https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/5965";
-#endif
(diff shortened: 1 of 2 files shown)


tests/unit/sync_object_positive.cpp
```diff
@@ -1119,28 +1119,6 @@ TEST_F(PositiveSyncObject, BasicSetAndWaitEvent2) {
     m_device->Wait();
 }
 
-TEST_F(PositiveSyncObject, WaitEventThenSet) {
```
spencer-lunarg, line 1122: > Seemingly we don't have VUID-vkSetEvent-event-09543 implemented fully.

Can we move this to a negative test with an TODO (pointed to a new issue) to finish `VUID-vkSetEvent-event-09543` properly

artem-lunarg: Updated

spencer-lunarg (approved):

jeremyg-lunarg (approved):

PR #8757, opened 2024-10-25
build: Update to header 1.3.300

mikes-lunarg:


spencer-lunarg (approved):
Code looks good, but lack of 300 reference left me empty inside :cry:

PR #8769, opened 2024-10-29
layers: Combine Dynamic Patch Control

spencer-lunarg:
2nd attempt at https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8593/commits/b3bd2ba934b3725281490d37cb8ea1978f5e25a7

### DIFF
--- a/layers/stateless/sl_pipeline.cpp
+++ b/layers/stateless/sl_pipeline.cpp
@@ -552,8 +552,11 @@ bool StatelessValidation::manual_PreCallValidateCreateGraphicsPipelines(
                 skip |= ValidatePipelineTessellationStateCreateInfo(*create_info.pTessellationState,
                                                                     create_info_loc.dot(Field::pTessellationState));
 
-                if (create_info.pTessellationState->patchControlPoints == 0 ||
-                    create_info.pTessellationState->patchControlPoints > device_limits.maxTessellationPatchSize) {
+                const bool has_dynamic_patch_control_points =
+                    vvl::Contains(dynamic_state_map, VK_DYNAMIC_STATE_PATCH_CONTROL_POINTS_EXT);
+                if (!has_dynamic_patch_control_points &&
+                    (create_info.pTessellationState->patchControlPoints == 0 ||
+                     create_info.pTessellationState->patchControlPoints > device_limits.maxTessellationPatchSize)) {
                     skip |= LogError("VUID-VkPipelineTessellationStateCreateInfo-patchControlPoints-01214", device,
                                      create_info_loc.dot(Field::pTessellationState).dot(Field::patchControlPoints),
                                      "is %" PRIu32 ", but should be between 0 and maxTessellationPatchSize (%" PRIu32 ").",
--- a/tests/unit/shader_object_positive.cpp
+++ b/tests/unit/shader_object_positive.cpp
@@ -1670,3 +1670,33 @@ TEST_F(PositiveShaderObject, DiscardRectangleModeEXT) {
     m_command_buffer.EndRendering();
     m_command_buffer.End();
 }
+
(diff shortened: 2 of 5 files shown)


tests/unit/shader_object_positive.cpp
```diff
@@ -1670,3 +1670,34 @@ TEST_F(PositiveShaderObject, DiscardRectangleModeEXT) {
...
+    m_command_buffer.begin();
+    m_command_buffer.BeginRenderingColor(GetDynamicRenderTarget(), GetRenderTargetArea());
+    SetDefaultDynamicStatesExclude({VK_DYNAMIC_STATE_PATCH_CONTROL_POINTS_EXT});
+    const VkShaderStageFlagBits stages[] = {VK_SHADER_STAGE_VERTEX_BIT, VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT,
+                                            VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT, VK_SHADER_STAGE_GEOMETRY_BIT,
+                                            VK_SHADER_STAGE_FRAGMENT_BIT};
+    const VkShaderEXT shaders[] = {vertShader.handle(), tescShader.handle(), teseShader.handle(), VK_NULL_HANDLE,
+                                   fragShader.handle()};
+    vk::CmdBindShadersEXT(m_command_buffer.handle(), 5u, stages, shaders);
+    // valid because this is not VK_PRIMITIVE_TOPOLOGY_PATCH_LIST
```
ziga-lunarg, line 1695: Not sure I understand this comment, is it saying that `VK_PRIMITIVE_TOPOLOGY_PATCH_LIST` is not valid?

ziga-lunarg (commented):
LGTM, but I hope there will be a VUID for this

tests/unit/shader_object_positive.cpp
spencer-lunarg: because I set `VK_PRIMITIVE_TOPOLOGY_TRIANGLE_LIST` the 

`and the current value of primitiveTopology is VK_PRIMITIVE_TOPOLOGY_PATCH_LIST` part of the VU is not true and can ignore setting the `vkCmdSetPatchControlPointsEXT`

ziga-lunarg: This makes sense, but I took the comment to mean setting `vkCmdSetPrimitiveTopologyEXT` to anything other than `VK_PRIMITIVE_TOPOLOGY_PATCH_LIST` is valid

spencer-lunarg: @ziga-lunarg decided to just make this test as expected, I have a raised MR (https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6981) and after that gets approved and a new VU added will write a negative test for it

layers/stateless/sl_pipeline.cpp
```diff
@@ -552,8 +552,11 @@ bool StatelessValidation::manual_PreCallValidateCreateGraphicsPipelines(
                 skip |= ValidatePipelineTessellationStateCreateInfo(*create_info.pTessellationState,
                                                                     create_info_loc.dot(Field::pTessellationState));
 
-                if (create_info.pTessellationState->patchControlPoints == 0 ||
-                    create_info.pTessellationState->patchControlPoints > device_limits.maxTessellationPatchSize) {
+                const bool has_dynamic_patch_control_points =
+                    vvl::Contains(dynamic_state_map, VK_DYNAMIC_STATE_PATCH_CONTROL_POINTS_EXT);
+                if (!has_dynamic_patch_control_points &&
```
spencer-lunarg, line 557: the one test showed we were not ignoring `patchControlPoints` when `VK_DYNAMIC_STATE_PATCH_CONTROL_POINTS_EXT` is set

PR #8771, opened 2024-10-29
layers: Combine Stipple Enable Dynamic State

spencer-lunarg:
Combines `vkCmdSetLineStippleKHR` from  https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6956
cc @ziga-lunarg

### DIFF
--- a/layers/core_checks/cc_cmd_buffer_dynamic.cpp
+++ b/layers/core_checks/cc_cmd_buffer_dynamic.cpp
@@ -209,14 +209,18 @@ bool CoreChecks::ValidateDynamicStateIsSet(const LastBound& last_bound_state, co
             case CB_DYNAMIC_STATE_DISCARD_RECTANGLE_MODE_EXT:
                 vuid_str = vuid.dynamic_discard_rectangle_mode_07881;
                 break;
+            case CB_DYNAMIC_STATE_LINE_STIPPLE_KHR:
+                vuid_str = vuid.dynamic_line_stipple_ext_07849;
+                break;
             default:
                 assert(false);
                 break;
         }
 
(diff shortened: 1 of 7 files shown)


layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -209,14 +209,18 @@ bool CoreChecks::ValidateDynamicStateIsSet(const LastBound& last_bound_state, co
...
                 assert(false);
                 break;
         }
 
-        return LogError(vuid_str, objlist, vuid.loc(), "%s state is dynamic, but the command buffer never called %s.\n%s",
+        return LogError(vuid_str, objlist, vuid.loc(), "%s state is dynamic, but the command buffer never called %s.\n%s%s",
                         DynamicStateToString(dynamic_state), DescribeDynamicStateCommand(dynamic_state).c_str(),
-                        DescribeDynamicStateDependency(dynamic_state, pipeline).c_str());
+                        DescribeDynamicStateDependency(dynamic_state, pipeline).c_str(),
+                        last_bound_state.cb_state.DescribeInvalidatedState(dynamic_state).c_str());
```
spencer-lunarg, line 223: This is the code that makes it say

> (There was a call to vkCmdBindPipeline with VkPipeline 0x27d60e0000000019[Single Dynamic] that didn't have VK_DYNAMIC_STATE_LINE_STIPPLE_KHR and invalidated the prior vkCmdSetLineStippleKHR call).

... seems I never had this added to the new code path :cry: (the test caught it, I had a test to make sure we had this, but used Stipple Enable so just saw it now)

ziga-lunarg (approved):
LGTM

PR #8775, opened 2024-10-30
scripts: Cleanup generate_spec_error_message.py

spencer-lunarg:
In preparation for https://github.com/KhronosGroup/Vulkan-Docs/pull/2453 I went through and removed old stuff we use to do because the `validusage.json` use to be different before we allowed `ifdef` inside VUs

cc @mikes-lunarg


artem-lunarg (approved):
LGTM

PR #8779, opened 2024-10-30
layers: Move hostImageCopy feature check to image

spencer-lunarg:
for future 1.3.301 header - https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6982

Move the `hostImageCopy` to image creation time

cc @ziga-lunarg (since you caught this)


ziga-lunarg (approved):
LGTM

PR #8781, opened 2024-10-30
layers: Add ShaderObject Tess Topology draw VU

spencer-lunarg:
For future header fix in https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6981

prevents you to draw with shader object with tessellation without setting `VK_PRIMITIVE_TOPOLOGY_PATCH_LIST`

cc @ziga-lunarg who caught this


ziga-lunarg (approved):
LGTM

Issue #8753, opened 2024-10-24
Spurious VUID-vkAcquireNextImageKHR-surface-07783 Error


greg-churchillnavigation:
**Environment:**
 - OS: Windows 11 Pro build 22631.4317
 - GPU: nVidia RTX 2000 driver v31.0.15.5274
 - SDK: 1.3.296.0
 - Options enabled: validation, synchronization:

**Describe the Issue**

App has 2x OS windows and swapchains
When one window is minimized, vkQueuePresentKHR() returns VK_ERROR_OUT_OF_DATE_KHR.
App sees swapchain current surface extent 0,0, so skips rendering that window.
Later, window is restored, (as code was skipping zero size window, nothing was done)
swapchain vkAcquireNextImageKHR() returns VK_ERROR_OUT_OF_DATE_KHR, so re-creates swapchain.
	
Logging my calls shows all Acquires have a matching Presents.

Validation error:
```
Validation Error: [ VUID-vkAcquireNextImageKHR-surface-07783 ] Object 0: handle = 0xe88693000000000c, type = VK_OBJECT_TYPE_SWAPCHAIN_KHR; | MessageID = 0xad0e15f6 | vkAcquireNextImageKHR():  Application has already previously acquired 3 images from swapchain. Only 3 are available to be acquired using a timeout of UINT64_MAX (given the swapchain has 4, and VkSurfaceCapabilitiesKHR::minImageCount is 2).
The Vulkan spec states: If forward progress cannot be guaranteed for the surface used to create the swapchain member of pAcquireInfo, timeout must not be UINT64_MAX (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkAcquireNextImageKHR-surface-07783)
```
Occurs at vkAcquireNextImageKHR() for OTHER window swapchain, not the one that was minimized and restored.
Despite validation error, vkAcquireNextImageKHR() returns VK_SUCCESS.
	
Work around:
	In call to vkQueuePresentKHR()
	Pass not null array via VkPresentInfoKHR .pResults, to accept per swapchain image present results.
That's all! Don't actually do anything with the returned results.

Does not occur with 
GPU: Intel Graphics driver 32.0.101.6078

**Expected behavior**

Don't expect a validation error. No apparent leaked acquired images.
Do not expect optionally returned Present results to influence.

ziga-lunarg:
[...]

PR #8783, opened 2024-10-30
layers: Ensure QueuePresent has access to per-swapchain results

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8753

The issue was caused by presenting multiple swapchains with a single `QueuePresent` command. If one presentation failed, state tracking was disabled for all swapchains, even those without presentation errors. This led to false positives for "good" swapchains that had no presentation issues. This problem occurred when the optional `VkPresentInfoKHR::pResults` was not specified.

### DIFF
--- a/tests/framework/render.cpp
+++ b/tests/framework/render.cpp
@@ -697,6 +697,10 @@ void VkRenderFramework::InitSurface() {
 LRESULT CALLBACK WindowProc(HWND hwnd, UINT uMsg, WPARAM wParam, LPARAM lParam) {
     return DefWindowProc(hwnd, uMsg, wParam, lParam);
 }
+
+void SurfaceContext::Resize(uint32_t width, uint32_t height) {
+    ::SetWindowPos(m_win32Window, NULL, 0, 0, (int)width, (int)height, SWP_NOMOVE);
+}
 #endif  // VK_USE_PLATFORM_WIN32_KHR
 
 VkResult VkRenderFramework::CreateSurface(SurfaceContext &surface_context, VkSurfaceKHR &surface, VkInstance custom_instance) {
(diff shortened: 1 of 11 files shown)


tests/framework/render.cpp
```diff
@@ -922,6 +927,18 @@ void VkRenderFramework::DestroySwapchain() {
     }
 }
 
+void VkRenderFramework::SupportMultiSwapchain() {
```
artem-lunarg, line 930: As a separate change need to update other tests to use this and to remove them from blacklist.

tests/framework/render.h
```diff
@@ -50,10 +50,17 @@ struct SurfaceContext {
 #if defined(VK_USE_PLATFORM_XCB_KHR)
     xcb_connection_t *m_surface_xcb_conn{};
 #endif
-
 #if defined(VK_USE_PLATFORM_METAL_EXT)
     void *caMetalLayer;
 #endif
+
+#if defined(VK_USE_PLATFORM_WIN32_KHR)
+    static bool CanResize() { return true; }
```
artem-lunarg, line 58: For now for windows. If it's stable then need to implement the same for other platforms.

layers/chassis/chassis_modification_state.h
```diff
@@ -135,4 +135,19 @@ struct CreateBuffer {
     VkBufferCreateInfo modified_create_info;
 };
 
+struct QueuePresentKHR {
```
spencer-lunarg, line 138: My suggestion is to just inline this all into the Chassis function, the main purpose of `chassis::` structs was it allowed the `PreCall` and `PostCall` functions to consume them, but if the chassis call is the only thing that needs to know about `chassis::QueuePresentKHR` then we can just leave it defined for that call

spencer-lunarg: maybe better way to say it: Because `PostCallRecordQueuePresentKHR` just takes a normal `VkPresentInfoKHR` and not `chassis::QueuePresentKHR` I don't see a need to have this in this header file

artem-lunarg: Makes sense, thanks for explanation.

spencer-lunarg (approved):

Issue #8763, opened 2024-10-28
debugPrintfEXT - missing output for pass with previously bound but unused descriptor set 


flagrans-dev:
**Environment:**
 - OS: Windows 11
 - GPU and driver version: NVIDIA RTX 2070 Super 561.09
 - SDK or header version if building from repo: 1.3.296
 - Options enabled (synchronization, best practices, etc.): `debugPrintfEXT`

**Describe the Issue**

Compute dispatches that do not use any descriptor sets (e.g., because buffers are accessed with a buffer device address) do not output any messages from `debugPrintfEXT` **if a descriptor set was bound for a previous compute pass**.
```c++
// 1st compute pass: binds a descriptor set
vkCmdBindPipeline(cmd, pipeline0, VK_PIPELINE_BIND_POINT_COMPUTE);
vkCmdBindDescriptorSet(cmd, VK_PIPELINE_BIND_POINT_COMPUTE, ...);
vkCmdDispatch(cmd, ...); // debugPrintfEXT here works

// 2nd compute pass: does not require any descriptor sets (but previous set is still bound)
vkCmdBindPipeline(cmd, pipeline1, VK_PIPELINE_BIND_POINT_COMPUTE);
vkCmdDispatch(cmd, ...); // !!! debugPrintfEXT here does not output anything !!!
```
I tried the following things:
- Disable the extension and calls to `debugPrintfEXT` in the first compute shader: still no output
- Remove the first `vkCmdDispatch`: still no output
- Remove the `vkCmdBindDescriptorSet`: it works
- Swap the order of the two passes: debug outputs for both passes work
- Change the pipeline layout and shaders for the second pass to use the same descriptor set as the first pass: debug outputs for both passes work
- Change the pipeline layout and shaders for the second pass to use a different descriptor set and bind it with a second call of `vkCmdBindDescriptorSet` to the same slot: debug outputs for both passes work
- In all of the cases above: record a renderpass and graphics pipeline before the compute passes with the same or different descriptor set layouts: doesn't interfere with what happens to the compute passes

So to me it seems that the call to `vkCmdBindDescriptorSet` is the cause of the issue, as it binds the descriptor set to the compute bind point and there is no way to "unbind" it for the successive passes. If I understood the way `debugPrintfEXT` works correctly, the validation layer inserts another descriptor set for the message buffer. Maybe this actually unused descriptor set messes something up?
[...]

PR #8784, opened 2024-10-31
DebugPrintf and Setting tweaks

spencer-lunarg:
This started as trying to add a test for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8763 

but then realize a few things

- Printf should never use Duplicated Message Limit as it will only possibly cause people to think their shader is broken and not printing
- We don't have a dedicated file to test settings
- the `enable_message_limit` didn't actually do anything and setting it outside of `vkconfig` didn't actually turn off the duplicate message limit


layers/layer_options.cpp
```diff
@@ -662,12 +662,13 @@ static void ProcessDebugReportSettings(ConfigAndEnvSettings *settings_data, VkuL
                 "will not be seen.");
             report_flags |= kInformationBit;
         }
-        if (!settings_data->gpuav_settings->debug_printf_to_stdout && settings_data->gpuav_settings->debug_printf_verbose &&
-            debug_report->duplicate_message_limit != 0) {
-            // If verbose is disabled, it bypasses the LogMsgEnabled() check and not important
+        // If any non-stdout debugprintf is being used, just turn off duplicate_message_limit, it will prevent people thinking
+        // DebugPrintf is broken because nothing is printing.
+        if (!settings_data->gpuav_settings->debug_printf_to_stdout && debug_report->duplicate_message_limit != 0) {
```
artem-lunarg, line 667: why special case for stdout?

artem-lunarg (approved):

layers/layer_options.cpp
spencer-lunarg: the code looks like

```c++
 if (use_stdout) {
                std::cout << shader_message.str();
            } else {
                gpuav.debug_report->DebugLogMsg(kInformationBit, {}, shader_message.str().c_str(), "VVL-DEBUG-PRINTF");
            }
```

we only print to stdout or your callback, not both

spencer-lunarg: nope, but stdout is also an opt-in feature

the idea of the duplication limit is by default we don't spam people with 1000 of the same error because it happens 60 fps

Issue #8772, opened 2024-10-29
VkSwapchainPresentModesCreateInfoEXT with swapchain recreation produces wrong message


Agrael1:
**Environment:**
 - OS: Windows 11
 - GPU and driver version: NVidia A4000, 560.x
 - SDK or header version if building from repo: 1.3.296
 - Options enabled (synchronization, best practices, etc.): yes

**Describe the Issue**

I had created the swapchain with `VkSwapchainPresentModesCreateInfoEXT` initially, but on resize I forgot to add the same pNext and got weird error:
```
[Validation layer]: VUID-VkSwapchainCreateInfoKHR-pNext-07781
 [Message]:Validation Error: [ VUID-VkSwapchainCreateInfoKHR-pNext-07781 ] | MessageID = 0x4c8929c1 | vkCreateSwapchainKHR(): pCreateInfo->imageExtent (width = 2560, height = 1349), which is outside the bounds returned by vkGetPhysicalDeviceSurfaceCapabilitiesKHR(): currentExtent = (width = 1920, height = 1080), minImageExtent = (width = 1920, height = 1080), maxImageExtent = (width = 1920, height = 1080).
The Vulkan spec states: If a VkSwapchainPresentScalingCreateInfoEXT structure was not included in the pNext chain, or it is included and VkSwapchainPresentScalingCreateInfoEXT::scalingBehavior is zero then imageExtent must be between minImageExtent and maxImageExtent, inclusive, where minImageExtent and maxImageExtent are members of the VkSurfaceCapabilitiesKHR structure returned by vkGetPhysicalDeviceSurfaceCapabilitiesKHR for the surface (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-VkSwapchainCreateInfoKHR-pNext-07781)
```
although the resize code contains `vkGetPhysicalDeviceSurfaceCapabilitiesKHR` and returned size is correct.

**Expected behavior**

A message stating that I have forgot to include the same pnexts or nothing.

artem-lunarg:
> A message stating that I have forgot to include the same pnexts or nothing.

@Agrael1 validation layer probably can't use previous behavior to make suggestion about API use, it validates API usage only based on specification. 

As I understand there first part of the message is problematic:
[...]

PR #8789, opened 2024-10-31
layers: Fix wsi image extent caching

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8772

Also move all image extent checks to `ValidateSwapchainImageExtent`.

### DIFF
--- a/layers/state_tracker/image_state.cpp
+++ b/layers/state_tracker/image_state.cpp
@@ -774,7 +774,23 @@ const Surface::PhysDevCache *Surface::GetPhysDevCache(VkPhysicalDevice phys_dev)
 
 void Surface::UpdateCapabilitiesCache(VkPhysicalDevice phys_dev, const VkSurfaceCapabilitiesKHR &surface_caps) {
     auto guard = Lock();
-    cache_[phys_dev].capabilities = surface_caps;
+    PhysDevCache &cache = cache_[phys_dev];
+
+    cache.capabilities = surface_caps;
+
+    // The properties that vary per present mode are (check VkSurfacePresentModeEXT documentation):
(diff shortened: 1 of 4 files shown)


layers/state_tracker/image_state.cpp
```diff
@@ -774,7 +774,23 @@ const Surface::PhysDevCache *Surface::GetPhysDevCache(VkPhysicalDevice phys_dev)
...
+
+    // The properties that vary per present mode are (check VkSurfacePresentModeEXT documentation):
+    //  VkSurfaceCapabilitiesKHR::minImageCount
+    //  VkSurfaceCapabilitiesKHR::maxImageCount
+    //  VkSurfacePresentScalingCapabilitiesEXT::minScaledImageExtent
+    //  VkSurfacePresentScalingCapabilitiesEXT::maxScaledImageExtent
+    //
+    // Update extent properties since they do not vary per present mode.
+    // NOTE: other present mode independent properties can be added too if needed for validation
+    for (PresentModeInfo &present_mode_info : cache.present_mode_infos) {
```
artem-lunarg, line 789: Main part of the fix. We have two groups of cached properties: present mode independent and per present mode. Image extent properties do not depend on present mode so we can synchronize them between two groups. This fixes issues with swapchain image size validation.

I know there is a discussion how to properly handle present mode dependent properties here https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8321 and here https://gitlab.khronos.org/vulkan/vulkan/-/issues/3963 but that's orthogonal question and caching should be adjusted depending on what they decide.

spencer-lunarg (approved):

PR #8790, opened 2024-10-31
layers: Revise VU 08358 to include equality

lolzballs:
In Vulkan 1.3.299, VkVideoEncodeRateControlInfoKHR VU 08358 was updated to allow initialVirtualBufferSizeInMs to equal virtualBufferSizeInMs.

Also add a positive test case for this.


spencer-lunarg (approved):
LGTM @aqnuep can you quick confirm this is good to you

PR #8794, opened 2024-11-01
layers: Better surface caps caching

artem-lunarg:
Replaces yesterday's fix (https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8789) with a different one (so the same tests).
Why better:
- Simpler then yesterday's fix (no need to sync between two groups)
- Potentially fixes most blacklisted Android WSI tests. Big if true. (according to my experiments  https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8792 there's a chance, but need to test with current changes).
- Potentially solves a problem raised by Ralph or at least handles additional cases (failures of our Android WSI is likely one of the manifestation of the problem): https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8321

### DIFF
--- a/layers/core_checks/cc_wsi.cpp
+++ b/layers/core_checks/cc_wsi.cpp
@@ -368,7 +368,7 @@ bool CoreChecks::ValidateCreateSwapchain(const VkSwapchainCreateInfoKHR &create_
     }
 #endif
     VkSurfacePresentModeEXT present_mode_info = vku::InitStructHelper();
-    if (IsExtEnabled(device_extensions.vk_ext_surface_maintenance1)) {
+    if (surface_state->IsLastCapabilityQueryUsedPresentMode(physical_device_state->VkHandle())) {
         present_mode_info.presentMode = create_info.presentMode;
         present_mode_info.pNext = surface_info_pnext;
         surface_info_pnext = &present_mode_info;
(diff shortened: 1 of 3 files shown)


artem-lunarg:
The following test can be unblacklisted from GalaxyS24Ultra/Pixel_8:
    `NegativeWsi.SwapchainMaintenance1ExtensionRelease`
    `NegativeWsi.SwapchainPresentModeInfoImplicit`
    `PositiveWsi.PresentFenceWaitsForSubmission`
    `PositiveWsi.PresentFenceRetiresPresentQueueOperation`
    `PositiveWsi.QueueWaitsForPresentFence`
    `PositiveWsi.ReleaseAndAcquireSwapchainImages`

layers/core_checks/cc_wsi.cpp
```diff
@@ -368,7 +368,7 @@ bool CoreChecks::ValidateCreateSwapchain(const VkSwapchainCreateInfoKHR &create_
     }
 #endif
     VkSurfacePresentModeEXT present_mode_info = vku::InitStructHelper();
-    if (IsExtEnabled(device_extensions.vk_ext_surface_maintenance1)) {
+    if (surface_state->LastCapabilitiesQueryUsedPresentMode(physical_device_state->VkHandle())) {
```
artem-lunarg, line 371: That's the main fix. Swapchain i created based on parameters from `GetPhysicalDeviceSurfaceCapabilities` query. If that  query specified present mode then we take present mode into account during validation. Otherwise, do not take it into account. In this case VVL validation will be in sync with the parameters that the client uses.

spencer-lunarg (approved):
agree, much simpler

PR #8797, opened 2024-11-03
tests: Fix crash on some multi gpu configurations

artem-lunarg:
Crash on the system where the first physical device is AMD GPU and the second one NVIDIA GPU.


spencer-lunarg (approved):

artem-lunarg:
@spencer-lunarg updated PR with fixing all such places and added helper function

PR #8808, opened 2024-11-04
scripts: Let generate_source.py be called with a single path

jeremyg-lunarg:
In most cases both Vulkan-Headers and SPIRV-Headers are in a single base directory, such as external/. If only 1 path is passed in, figure out the registry and grammar files relative to it, otherwise you must pass in the 2 correct directories as before.

### DIFF
--- a/scripts/generate_source.py
+++ b/scripts/generate_source.py
@@ -401,8 +401,9 @@ def main(argv):
                         default='vulkan',
                         choices=['vulkan'],
                         help='Specify API name to generate')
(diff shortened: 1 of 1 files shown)


mikes-lunarg:
I think the next logical step would be to allow calling generate_source.py with 0 paths and default to `external`.

That would require a change to update_deps.py which currently maintains a source checkout per build configuration. Instead, it would be nice if we could have a single source directory for all build variants and just keep the build directories separated by variant.

spencer-lunarg (approved):

PR #8809, opened 2024-11-04
layers: Check null pApplicationName

kocdemir:
Small change to add nullptr check for pApplicationName to avoid crashes at string assignment for display_application_name setting.


spencer-lunarg (approved):
Thanks for adding!

lunarpapillo:
The **LunarG CI** failure seems to be due to an unrelated GPU crash on the system.  I'm restarting CI.

PR #8812, opened 2024-11-05
gpuav: Handle VK_PIPELINE_COMPILE_REQUIRED

arno-lunarg:


spencer-lunarg (approved):
This is good, I looked at `ValidationStateTracker::PostCallRecordCreateGraphicsPipelines` and it just checks if the `VkPipeline` handle is NULL and that is how it has been handling this

PR #8814, opened 2024-11-06
sync: Flags2 and KHR types cleanup

artem-lunarg:


scripts/generators/sync_validation_generator.py
```diff
@@ -281,12 +281,13 @@ def generateSource(self):
...
 
         # syncStageAccessMaskByAccessBit
-        out.append('const std::map<VkAccessFlags2, SyncStageAccessFlags>& syncStageAccessMaskByAccessBit() {\n')
-        out.append('    static const std::map<VkAccessFlags2, SyncStageAccessFlags> variable = {\n')
+        out.append('const std::map<VkAccessFlagBits2, SyncStageAccessFlags>& syncStageAccessMaskByAccessBit() {\n')
+        out.append('    static const std::map<VkAccessFlagBits2, SyncStageAccessFlags> variable = {\n')
         access_to_stageAccess = {}
         for stageAccess_info in self.stageAccessCombo:
             access = stageAccess_info['access']
-            if access == 'VK_ACCESS_2_FLAG_NONE_KHR': continue
```
spencer-lunarg, line 289: I see `VK_ACCESS_2_FLAG_NONE_KHR` was wrong from the start lol

layers/vulkan/generated/sync_validation_types.h
```diff
@@ -472,22 +472,22 @@ static const SyncStageAccessFlags syncStageAccessWriteMask = ( //  Mask of all w
 );
 
 // Bit order mask of stage_access bit for each stage. Order matters, don't try to use vvl::unordered_map
-const std::map<VkPipelineStageFlags2, SyncStageAccessFlags>& syncStageAccessMaskByStageBit();
```
spencer-lunarg, line 475: glad we didn't have a bug were two similar flags were in keys, nice catch!

spencer-lunarg (approved):

scripts/generators/sync_validation_generator.py
artem-lunarg: Yes, it was needed and it was wrong!

layers/vulkan/generated/sync_validation_types.h
artem-lunarg: Unfortunately a compiler can't detect this issue in sync2 (https://chat.google.com/room/AAAAOXVAYGg/vcRNb6Sd7cY/ZzUuE6Z_wDI?cls=10), but at least it's visible now in the source code.

layers/sync/sync_utils.cpp
```diff
@@ -80,45 +80,45 @@ VkPipelineStageFlags2KHR ExpandPipelineStages(VkPipelineStageFlags2KHR stage_mas
...
 
-VkAccessFlags2KHR ExpandAccessFlags(VkAccessFlags2KHR access_mask) {
-    VkAccessFlags2KHR expanded = access_mask;
+VkAccessFlags2 ExpandAccessFlags(VkAccessFlags2 access_mask) {
+    VkAccessFlags2 expanded = access_mask;
 
-    if (VK_ACCESS_2_SHADER_READ_BIT_KHR & access_mask) {
-        expanded = expanded & ~VK_ACCESS_2_SHADER_READ_BIT_KHR;
+    if (VK_ACCESS_2_SHADER_READ_BIT & access_mask) {
+        expanded = expanded & ~VK_ACCESS_2_SHADER_READ_BIT;
```
filnet, line 105: use `&=`

PR #8815, opened 2024-11-06
tests: Fix tests on Windows Arm

spencer-lunarg:
Fixes a few tests that were failing on Windows Arm we are trying to add into CI


tests/unit/shader_storage_image.cpp
```diff
@@ -154,6 +154,7 @@ TEST_F(NegativeShaderStorageImage, MissingFormatReadForFormat) {
     // Find storage formats with & without read without format support
     for (uint32_t fmt = VK_FORMAT_R4G4_UNORM_PACK8; fmt < VK_FORMAT_D16_UNORM; fmt++) {
         if (has_without_format_test && has_with_format_test) break;
+        if (!vkuFormatIsSampledFloat((VkFormat)fmt)) continue;
```
artem-lunarg, line 157: Comment would be nice

artem-lunarg (approved):

tests/unit/shader_storage_image.cpp
spencer-lunarg: I think this test is awful and needs a re-write

PR #8817, opened 2024-11-06
gpuav: Fix descriptor set bindings restore

arno-lunarg:


layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -432,25 +432,31 @@ void PostCallSetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer
...
-                    last_bound_desc_set_pipe_layout_state->set_layouts.size() - bindings_counts_from_last_pipeline);
-                const uint32_t first_disturbed_set = bindings_counts_from_last_pipeline;
+                    last_bound_desc_set_pipe_layout_state->set_layouts.size() - desc_set_bindings_counts_from_last_pipeline);
+                const uint32_t first_disturbed_set = desc_set_bindings_counts_from_last_pipeline;
 
                 for (uint32_t set_i = 0; set_i < disturbed_bindings_count; ++set_i) {
                     const uint32_t last_bound_set_i = set_i + first_disturbed_set;
-                    VkDescriptorSet desc_set = last_bound.per_set[last_bound_set_i].bound_descriptor_set->VkHandle();
+                    const auto &last_bound_set_state = last_bound.per_set[last_bound_set_i].bound_descriptor_set;
+                    // last_bound.per_set is a LUT, and descriptor sets before the last one could be unbound.
```
spencer-lunarg, line 450: > last_bound.per_set is a LUT

I don't think that is accurate, it is an vector of each of the descriptor slots, `set = 4` in your GLSL is `per_set[4]`, not indirect lookup

```suggestion
                    // There can be gaps in the middle of your valid descriptor sets
```

layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -30,8 +30,8 @@
 namespace gpuav {
 
 // If application is using shader objects, bindings count will be computed from bound shaders
-static uint32_t GetBindingsCountFromLastBoundPipelineOrShader(Validator &gpuav, VkPipelineBindPoint bind_point,
-                                                              CommandBuffer &cb_state, const LastBound &last_bound) {
+static uint32_t GetDescSetBindingsCountFromLastBoundPipelineOrShader(Validator &gpuav, VkPipelineBindPoint bind_point,
```
spencer-lunarg, line 33: honestly `PipelineOrShader` is a terminal term, I think `GetDescSetBindingsCountFromLastBound` is find, we imply this means "pipeline or shader" everywhere else so doing it here feels inconsistent

spencer-lunarg (approved):
LGTM - some nit comments

layers/gpu/instrumentation/gpuav_instrumentation.cpp
arno-lunarg: I know long name, would love to cut that, but it must be clear that we are retrieving bindings count from last bound pipeline/shader, and not from last bound descriptor sets. That's the whole reason for this mess of descriptor set restoration : /

layers/gpu/instrumentation/gpuav_instrumentation.cpp
spencer-lunarg: I could argue a pointer is a LUT, everything is going to "look up" ... Like I guess I would be willing to call everything a LUT then... Like every state object map is a LUT, there is no end if we start calling thing a LUT

arno-lunarg: No, we do have genuine arrays/list where the index of an element has no real meanings,  and we have maps/LUTs, where the index does have a meaning

PR #8827, opened 2024-11-07
gpu: Cleanup DescriptorClass usage

spencer-lunarg:
Two small commits to improve the usage of `DescriptorClass`

Main thing is removing the `default` in the switch cases so it is easier to see which descriptors are covered or not in various spots


jeremyg-lunarg (commented):
# #

jeremyg-lunarg (approved):
LGTM

PR #8828, opened 2024-11-08
tests: Add inline uniform buffer for GPU-AV

spencer-lunarg:
Inline Uniform Buffers don't have anyway to be OOB (pending https://gitlab.khronos.org/vulkan/vulkan/-/issues/4083) so the only thing we "could" do is check if the value you are reading from was not updated, and give a warning the data was uninitialized... but for now, added tests to make sure we don't generate false positive for Inline Uniform Buffers

### DIFF
--- a/layers/core_checks/cc_descriptor.cpp
+++ b/layers/core_checks/cc_descriptor.cpp
@@ -16,6 +16,7 @@
  * limitations under the License.
  */
 
+#include <sstream>
 #include <valarray>
 
 #include "containers/custom_containers.h"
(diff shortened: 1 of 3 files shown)


layers/core_checks/cc_descriptor.cpp
```diff
@@ -1884,14 +1885,24 @@ bool CoreChecks::ValidateWriteUpdate(const DescriptorSet &dst_set, const VkWrite
             if (current_binding->count > 0) {
                 // Check for consistent stageFlags and descriptorType
                 if ((current_binding->stage_flags != stage_flags) || (current_binding->type != descriptor_type)) {
+                    std::stringstream extra;
+                    // If using inline, easy to go outside of its range and not realize you are in the next descriptor
```
spencer-lunarg, line 1889: I hit this writing the test, it is not obvious what is going on so added this to clarify the error message

layers/core_checks/cc_descriptor.cpp
```diff
@@ -1884,14 +1885,24 @@ bool CoreChecks::ValidateWriteUpdate(const DescriptorSet &dst_set, const VkWrite
...
+                    if (descriptor_type == VK_DESCRIPTOR_TYPE_INLINE_UNIFORM_BLOCK) {
+                        extra << " For inline uniforms blocks, you might have your VkWriteDescriptorSet::dstArrayElement ("
+                              << update.dstArrayElement << ") plus VkWriteDescriptorSet::descriptorCount ("
```
arno-lunarg, line 1892:
```suggestion
                              << update.dstArrayElement << ") + VkWriteDescriptorSet::descriptorCount ("
```

arno-lunarg (approved):

layers/core_checks/cc_descriptor.cpp
spencer-lunarg: going to make a quick follow-up PR, we use ` plus ` in other VUs, should just make them all `+`

PR #8829, opened 2024-11-08
gpuav: Fix handling of PIPELINE_COMPILE_REQUIRED

arno-lunarg:
- Remove VK_PIPELINE_CREATE_FAIL_ON_PIPELINE_COMPILE_REQUIRED_BIT from instrumented pipeline libray
GPU-AV needs to succeed in creating them no matter driver cache state

- Renaming for clarification

### DIFF
--- a/layers/gpu/instrumentation/gpuav_shader_instrumentor.cpp
+++ b/layers/gpu/instrumentation/gpuav_shader_instrumentor.cpp
@@ -454,8 +454,11 @@ void GpuShaderInstrumentor::PreCallRecordCreateGraphicsPipelines(VkDevice device
         auto &shader_instrumentation_metadata = chassis_state.shader_instrumentations_metadata[i];
 
         if (pipeline_state->linking_shaders != 0) {
-            PreCallRecordPipelineCreationShaderInstrumentationGPL(pAllocator, *pipeline_state, new_pipeline_ci, create_info_loc,
-                                                                  shader_instrumentation_metadata);
+            bool success = PreCallRecordPipelineCreationShaderInstrumentationGPL(pAllocator, *pipeline_state, new_pipeline_ci,
+                                                                                 create_info_loc, shader_instrumentation_metadata);
+            if (!success) {
+                return;
+            }
         } else {
             PreCallRecordPipelineCreationShaderInstrumentation(pAllocator, *pipeline_state, new_pipeline_ci, create_info_loc,
                                                                shader_instrumentation_metadata);
@@ -1010,12 +1013,13 @@ void GpuShaderInstrumentor::PostCallRecordPipelineCreationShaderInstrumentation(
     }
 }
 
-// While have an almost duplicated funciton is not ideal, the core issue is we have a single, templated function designed for
+// While have an almost duplicated function is not ideal, the core issue is we have a single, templated function designed for
 // Graphics, Compute, and Ray Tracing. GPL is only for graphics, so we end up needing this "side code path" for graphics only and it
 // doesn't fit in the "all pipeline" templated flow.
-void GpuShaderInstrumentor::PreCallRecordPipelineCreationShaderInstrumentationGPL(
-    const VkAllocationCallbacks *pAllocator, vvl::Pipeline &pipeline_state, vku::safe_VkGraphicsPipelineCreateInfo &new_pipeline_ci,
-    const Location &loc, std::vector<chassis::ShaderInstrumentationMetadata> &shader_instrumentation_metadata) {
+bool GpuShaderInstrumentor::PreCallRecordPipelineCreationShaderInstrumentationGPL(
+    const VkAllocationCallbacks *pAllocator, vvl::Pipeline &pipeline_state,
+    vku::safe_VkGraphicsPipelineCreateInfo &modified_pipeline_ci, const Location &loc,
+    std::vector<chassis::ShaderInstrumentationMetadata> &shader_instrumentation_metadata) {
     // Init here instead of in chassis so we don't pay cost when GPU-AV is not used
     const size_t total_stages = pipeline_state.stage_states.size();
     shader_instrumentation_metadata.resize(total_stages);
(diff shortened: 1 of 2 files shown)


layers/gpu/instrumentation/gpuav_shader_instrumentor.cpp
```diff
@@ -1010,83 +1013,98 @@ void GpuShaderInstrumentor::PostCallRecordPipelineCreationShaderInstrumentation(
...
+         ++to_instrument_lib_i) {
+        const auto to_instrument_lib = Get<vvl::Pipeline>(to_instrument_pipeline_lib_ci->pLibraries[to_instrument_lib_i]);
+        if (!to_instrument_lib) continue;
+        if (to_instrument_lib->stage_states.empty()) continue;
+
+        vku::safe_VkGraphicsPipelineCreateInfo to_instrument_pipeline_ci(to_instrument_lib->GraphicsCreateInfo());
+        // If the application supplied pipeline might be interested in failing to be created
+        // if the driver does not find it in its cache, GPU-AV needs to succeed in the instrumented pipeline library
+        // creation process no matter caching state.
+        to_instrument_pipeline_ci.flags &= ~VK_PIPELINE_CREATE_FAIL_ON_PIPELINE_COMPILE_REQUIRED_BIT;
```
arno-lunarg, line 1047: Here is the one line fix

layers/gpu/instrumentation/gpuav_shader_instrumentor.cpp
```diff
@@ -454,8 +454,11 @@ void GpuShaderInstrumentor::PreCallRecordCreateGraphicsPipelines(VkDevice device
         auto &shader_instrumentation_metadata = chassis_state.shader_instrumentations_metadata[i];
 
         if (pipeline_state->linking_shaders != 0) {
-            PreCallRecordPipelineCreationShaderInstrumentationGPL(pAllocator, *pipeline_state, new_pipeline_ci, create_info_loc,
-                                                                  shader_instrumentation_metadata);
+            bool success = PreCallRecordPipelineCreationShaderInstrumentationGPL(pAllocator, *pipeline_state, new_pipeline_ci,
+                                                                                 create_info_loc, shader_instrumentation_metadata);
+            if (!success) {
+                return;
```
spencer-lunarg, line 460: if you tried to create more than 1 pipelines in a single `vkCreateGraphicsPipelines` call, this would just skip the the rest, so should be a `continue;`

... but also at that point, what is the need to `continue` if there is nothing left in the `for` loop to hit

layers/gpu/instrumentation/gpuav_shader_instrumentor.cpp
```diff
@@ -1010,83 +1013,98 @@ void GpuShaderInstrumentor::PostCallRecordPipelineCreationShaderInstrumentation(
...
     // Init here instead of in chassis so we don't pay cost when GPU-AV is not used
     const size_t total_stages = pipeline_state.stage_states.size();
     shader_instrumentation_metadata.resize(total_stages);
 
     InstrumentationDescriptorSetLayouts instrumentation_dsl;
     BuildDescriptorSetLayoutInfo(pipeline_state, instrumentation_dsl);
 
-    auto library_create_info = const_cast<VkPipelineLibraryCreateInfoKHR *>(
-        vku::FindStructInPNextChain<VkPipelineLibraryCreateInfoKHR>(new_pipeline_ci.pNext));
+    auto to_instrument_pipeline_lib_ci = const_cast<VkPipelineLibraryCreateInfoKHR *>(
```
spencer-lunarg, line 1030: for consistency, we have used `modified_` (ex `modified_create_info`) so rather we just do that then have some that are `to_instrument` and some that are `modified_`

... also we "instrument SPIR-V" we don't instrument a pNext struct

```suggestion
    auto modified_pipeline_library_ci = const_cast<VkPipelineLibraryCreateInfoKHR *>(
```

layers/gpu/instrumentation/gpuav_shader_instrumentor.cpp
```diff
@@ -1010,83 +1013,98 @@ void GpuShaderInstrumentor::PostCallRecordPipelineCreationShaderInstrumentation(
...
+                    is_shader_cached = true;
                 }
             } else {
                 unique_shader_id = unique_shader_module_id_++;
             }
-            if (!cached) {
-                pass =
-                    InstrumentShader(module_state->spirv->words_, unique_shader_id, instrumentation_dsl, loc, instrumented_spirv);
+            if (!is_shader_cached) {
+                was_shader_instrumented = InstrumentShader(to_instrument_module_state->spirv->words_, unique_shader_id,
```
spencer-lunarg, line 1102: mixing `was`/`is` is strange, `is_shader_instrumented` probably works

layers/gpu/instrumentation/gpuav_shader_instrumentor.cpp
arno-lunarg: if success if `false`, that means GPU-AV had to abort, so might as well return early and not instrument other pipelines?

layers/gpu/instrumentation/gpuav_shader_instrumentor.cpp
arno-lunarg: ok!

layers/gpu/instrumentation/gpuav_shader_instrumentor.cpp
spencer-lunarg: I feel in practice, we should just have an `InternalWarning`... we can still in theory instrument other pipelines

... I guess if we are aborting, there is a `aborted_` you can check up here

if we are going to do a `if (!success) return` then might as well keep the function void and just go `if (aborted_) return`

arno-lunarg: Yeah we could just warn, but looking at the error conditions in this function, I feel that if they are emitted, something went wrong and I'd rather sink our boat : 3
Not a super fan of looking at `aborted_`, what would be the rule to know when to or not to look when calling a GPU-AV function?

spencer-lunarg: sure, just leave this for now, but please leave a small comment when this (and not the non-GPL) version has a return like this

layers/gpu/instrumentation/gpuav_shader_instrumentor.cpp
```diff
@@ -454,11 +454,17 @@ void GpuShaderInstrumentor::PreCallRecordCreateGraphicsPipelines(VkDevice device
         auto &shader_instrumentation_metadata = chassis_state.shader_instrumentations_metadata[i];
 
         if (pipeline_state->linking_shaders != 0) {
-            PreCallRecordPipelineCreationShaderInstrumentationGPL(pAllocator, *pipeline_state, new_pipeline_ci, create_info_loc,
-                                                                  shader_instrumentation_metadata);
+            bool success = PreCallRecordPipelineCreationShaderInstrumentationGPL(pAllocator, *pipeline_state, new_pipeline_ci,
```
spencer-lunarg, line 457: want to just go

```
bool success = false;

if (pipeline_state->linking_shaders != 0) {
} else {
}

if (!success) return;
```

layers/gpu/instrumentation/gpuav_shader_instrumentor.cpp
```diff
@@ -889,89 +904,99 @@ void GpuShaderInstrumentor::BuildDescriptorSetLayoutInfo(const vvl::DescriptorSe
...
+                           selected_instrumented_shaders.end()) {
+                    continue;
+                }
             }
         }
-
         uint32_t unique_shader_id = 0;
-        bool cached = false;
-        bool pass = false;
+        bool is_instrumented_shader_cached = false;
```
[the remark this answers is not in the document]
arno-lunarg: yep!

PR #8830, opened 2024-11-08
Remove including state tracker in header 

spencer-lunarg:
Broken into commits, main goal was 

Remove the `#include <state_tracker.h>` from the `descriptor_validator.h` header


jeremyg-lunarg (approved):
LGTM, I think we already hit the renaming quota for the week

PR #8832, opened 2024-11-08
scripts: Remove most raw new/deletes from handle wrapping

jeremyg-lunarg:
Most of these can be replaced directly with small_vector<>. But, deferred operations still require use of raw pointers so that clean up can be ... deferred.


spencer-lunarg (approved):
`+1` for code
`+` :100:  for "deferred" joke

arno-lunarg (approved):

spencer-lunarg:
CI failure is from the GalaxyS24 not running, but the Pixel ran and was fine, so had Android coverage

Issue #8497, opened 2024-09-05
Reachable "should never happen" assertion error in cc_descriptor.cpp


NBickford-NV:
**Environment:**
 - OS: Windows 10 x86_64 (my repro should work on Windows 11 and Linux as well)
 - GPU and driver version: NVIDIA RTX 3090, 552.88
 - SDK or header version if building from repo: main (commit af7b0a35d009b5ad6e0b280a5b81388608ebfe39, 2024-09-03)
 - Options enabled (synchronization, best practices, etc.): None

**Describe the Issue**

Hi Vulkan Validation Layers team!

I managed to find a way to reach this assertion error at the end of `CoreChecks::VerifySetLayoutCompatibility()` in cc_descriptor.cpp:

```cpp
    // No detailed check should succeed if the trivial check failed -- or the dictionary has failed somehow.
    bool compatible = true;
    assert(!compatible);
    return compatible;
```

This can happen if an app:
1. creates a sampler;
2. creates two identical descriptor set layouts, pipeline layouts, and descriptor sets, each using a combined image sampler using the sampler from step 1;
3. then calls `vkCmdBindDescriptorSets()` with a pipeline layout from the first group and a descriptor set from the second:

```cpp
    // Create a sampler. The repro doesn't depend on the parameters.
    VkSamplerCreateInfo samplerInfo{ .sType = VK_STRUCTURE_TYPE_SAMPLER_CREATE_INFO };
    VkSampler sampler;
    NVVK_CHECK(vkCreateSampler(device, &samplerInfo, nullptr, &sampler));
...
    const uint32_t kCopies = 2;
    std::array<VkDescriptorSetLayout, kCopies> descriptorSetLayouts;
    std::array<VkPipelineLayout, kCopies> pipelineLayouts;
    std::array<VkDescriptorSet, kCopies> descriptorSets;
    for (uint32_t i = 0; i < kCopies; i++) {
        // Create identical descriptor set layouts, pipeline layouts, and descriptor sets.
        // pipelineLayouts[i] and descriptorSets[i] use descriptorSetLayouts[i], and
        // descriptorSetLayouts[i] uses `sampler`.
        // See attached repro for full details.
    }
...
    // Bind pipeline layout copy 0 and descriptor set copy 1:
    vkCmdBindDescriptorSets(cmd,        // commandBuffer
        VK_PIPELINE_BIND_POINT_COMPUTE,	// pipelineBindPoint
[...]

PR #8834, opened 2024-11-08
layers: Fix VerifySetLayoutCompatibility for Samplers

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8497

Add exception to check the `VkSamplerCreateInfo` of 2 `pImmutableSamplers` to see if they are the same for Descriptor Set Layout Compatibility

### DIFF
--- a/layers/state_tracker/descriptor_sets.h
+++ b/layers/state_tracker/descriptor_sets.h
@@ -275,6 +275,9 @@ static inline bool operator==(const DescriptorSetLayoutDef &lhs, const Descripto
         if (l.pImmutableSamplers) {
             for (uint32_t s = 0; s < l.descriptorCount; s++) {
                 if (l.pImmutableSamplers[s] != r.pImmutableSamplers[s]) {
+                    // https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8497
+                    // This just checks pointers, but two different VkSampler handles could be created with same createInfo.
+                    // Since this is rare enough, mark as "not the same" and check later when checking for compatibility.
                     return false;
                 }
             }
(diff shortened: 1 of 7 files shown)


layers/state_tracker/descriptor_sets.h
```diff
@@ -275,6 +275,9 @@ static inline bool operator==(const DescriptorSetLayoutDef &lhs, const Descripto
         if (l.pImmutableSamplers) {
             for (uint32_t s = 0; s < l.descriptorCount; s++) {
                 if (l.pImmutableSamplers[s] != r.pImmutableSamplers[s]) {
+                    // https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8497
```
artem-lunarg, line 278: The suggestion is to add a short description for the entire function that mentions this exception, so you can get correct idea of what the function does without checking all details.

artem-lunarg (approved):

layers/state_tracker/descriptor_sets.h
spencer-lunarg: I guess the function is just a `operator==` of `DescriptorSetLayoutDef`... I honestly don't want anyone to care about these details unless they need too... also no one calls this, this function is called via some hashing compare magic, so its not even obvious when you would call this

artem-lunarg: > I honestly don't want anyone to care about these details unless they need too

 I don't see here exception comparing to any other function. Special cases happens, lets assume each one is rare, but they happen often because there are a lot of them (one specific is rare, but probability of *any* is high). That's why in practice we constantly dealing with special cases, because they are not in total.

Issue #8837, opened 2024-11-09
False positive VUID-VkPresentInfoKHR-pNext-pNext


mahkoh:
**Environment:**
 - OS: n/a
 - GPU and driver version: n/a
 - SDK or header version if building from repo: 1.3.290
 - Options enabled (synchronization, best practices, etc.): none

**Describe the Issue**

When the pNext chain of VkPresentInfoKHR is VkPresentRegionsKHR -> VkSwapchainPresentFenceInfoEXT, the validation layers report

>VUID-VkPresentInfoKHR-pNext-pNext(ERROR / SPEC): msgNum: 748584181 - Validation Error: [ VUID-VkPresentInfoKHR-pNext-pNext ] | MessageID = 0x2c9e7cf5 | vkQueuePresentKHR(): pNext\<VkPresentRegionsKHR>.pNext must be NULL. This error is based on the Valid Usage documentation for version 290 of the Vulkan header.  It is possible that you are using a struct from a private extension or an extension that was added to a later version of the Vulkan header, in which case the use of pNext\<VkPresentRegionsKHR>.pNext is undefined and may not work correctly with validation enabled. The Vulkan spec states: Each pNext member of any structure (including this one) in the pNext chain must be either NULL or a pointer to a valid instance of VkDeviceGroupPresentInfoKHR, VkDisplayPresentInfoKHR, VkFrameBoundaryEXT, VkPresentFrameTokenGGP, VkPresentIdKHR, VkPresentRegionsKHR, VkPresentTimesInfoGOOGLE, VkSwapchainPresentFenceInfoEXT, or VkSwapchainPresentModeInfoEXT (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkPresentInfoKHR-pNext-pNext)

**Expected behavior**

The quoted spec states that this chain is valid.

**Valid Usage ID**

VUID-VkPresentInfoKHR-pNext-pNext

**Additional context**

The backtrace shows https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/6c3011be72dd3f418a8902acab340e8c1a00864d/layers/stateless/sl_wsi.cpp#L160-L162

PR #8838, opened 2024-11-10
Fix stateless code gen for structs

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8837


artem-lunarg (approved):

arno-lunarg (commented):
No test? Asking because in the issue you mention adding one

spencer-lunarg:
> No test? Asking because in the issue you mention adding one

So there are a LOT of `VUID-*-pNext-pNext` we don't tests, just adding a random one I don't find that helpful to catch future regressions, I was going to try and setup something to test them all at once in its own test file

PR #8841, opened 2024-11-11
layers: Use + instead of plus

spencer-lunarg:
Use `+` instead of `plus`


ziga-lunarg (approved):

arno-lunarg (approved):
I knew this PR was coming!!!

Issue #8835, opened 2024-11-09
Spurious VUID-vkCmdBindDescriptorBuffersEXT-maxResourceDescriptorBufferBindings-08049(ERROR / SPEC)


buzmeg:
**Environment:**
 -   OS: Ubuntu 24.04.1 LTS (Linux hostname 6.8.0-45-generic)
 -    GPU and driver version: Intel UHD Graphics 620 (WHL GT2)
 -    SDK or header version if building from repo: Vulkan Instance Version: 1.3.290
 -    Options enabled (synchronization, best practices, etc.): Validation

I received:

VUID-vkCmdBindDescriptorBuffersEXT-maxResourceDescriptorBufferBindings-08049(ERROR / SPEC): msgNum: -132400417 - Validation Error: [ VUID-vkCmdBindDescriptorBuffersEXT-maxResourceDescriptorBufferBindings-08049 ] Object 0: handle = 0x7fff88037490, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xf81bbadf | vkCmdBindDescriptorBuffersEXT():  Number of resource buffers is 2. There must be no more than maxResourceDescriptorBufferBindings (1) descriptor buffers containing resource descriptor data bound. List of resource buffers:
VkBuffer 0x2cfba2000000001c[]
VkBuffer 0x2cfba2000000001c[]

I'm pretty sure those are the same buffer.  :)

spencer-lunarg:
There is a TODO in the code

> // TODO - Understand what "descriptor buffer" means toward the limit, does the same VkBuffer reused count twice

I need to raise a spec issue to clarify if "resource descriptor buffers" cares about the number of "things bound" or "unique handles bound"

buzmeg:
Oddly, I have a different program that violates that limit significantly with completely different buffers, and yet works perfectly on my integrated Intel hardware.

Which, to a certain extent, sorta makes sense. Everything is indirected through VkBuffers allocated by the user and VkDeviceAddresses.  At bottom, it's just pointer chasing.  There is no inherent reason why everything has to come off of the same VkBuffer as long as the pointers and addresses all land in the memory that the card can use for descriptors.

In the mesa vulkan driver in anv_physical_device.c, for example, it's completely arbitrary:
```
      if (pdevice->uses_ex_bso) {
         props->maxDescriptorBufferBindings = MAX_SETS;
[...]

PR #8849, opened 2024-11-12
layers: Clarify Descriptor Buffer Binding limits

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8835

### DIFF
--- a/tests/unit/descriptor_buffer.cpp
+++ b/tests/unit/descriptor_buffer.cpp
@@ -17,6 +17,7 @@
 #include "../framework/layer_validation_tests.h"
 #include "../framework/pipeline_helper.h"
 #include "../framework/ray_tracing_objects.h"
+#include "utils/vk_layer_utils.h"
 
 class NegativeDescriptorBuffer : public DescriptorBufferTest {};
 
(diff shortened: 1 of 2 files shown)


tests/unit/descriptor_buffer.cpp
```diff
@@ -1668,3 +1669,58 @@ TEST_F(NegativeDescriptorBuffer, TexelBufferFormat) {
...
+        GTEST_SKIP() << "maxResourceDescriptorBufferBindings  is not 1";
+    }
+
+    VkDescriptorSetLayoutBinding binding = {0, VK_DESCRIPTOR_TYPE_STORAGE_BUFFER, 1, VK_SHADER_STAGE_ALL, nullptr};
+    vkt::DescriptorSetLayout ds_layout(*m_device, binding, VK_DESCRIPTOR_SET_LAYOUT_CREATE_DESCRIPTOR_BUFFER_BIT_EXT);
+
+    VkDeviceSize ds_layout_size = 0;
+    vk::GetDescriptorSetLayoutSizeEXT(device(), ds_layout.handle(), &ds_layout_size);
+
+    ds_layout_size = Align(ds_layout_size, descriptor_buffer_properties.descriptorBufferOffsetAlignment);
```
artem-lunarg, line 1689: `descriptorBufferOffsetAlignment` should be used to ensure that buffer address  `VkDescriptorBufferBindingInfoEXT::address` (line 1713) is aligned properly. Also can't find that spec guarantees that buffer address that corresponds to the beginning of the buffer has correct alignment. This is used in other tests, can be fixed as a separate cleanup.

When I experimented with this I assumed that mapped buffer pointer satisfies this requirement (based on `VK_BUFFER_USAGE_RESOURCE_DESCRIPTOR_BUFFER_BIT_EXT` flag), but that's probably not a robust solution: https://github.com/kennyalive/vulkan-base/blob/f1f9fb2562d2eecb2eae8215f9e02ef82c27bbc3/src/demo.cpp#L187

artem-lunarg (approved):
LGTM. One comment how alignment requirement is treated by the descriptor buffer tests.

tests/unit/descriptor_buffer.cpp
artem-lunarg: I would be happy if spec guarantees some alignments of buffer address of the beginning of the buffer. Let me know if you can find something. If no such thing, it actually feels like a spec design issue.

spencer-lunarg: will take a look, this is a copy-and-paste test so need to fix everywhere

Issue #8850, opened 2024-11-12
GPU-AV Instrumentation Cache not working correctly


spencer-lunarg:
Currently if you clear the cache and run

`--gtest_filter=NegativeGpuAVDescriptorIndexing.ArrayOOBRuntime:NegativeGpuAVDescriptorIndexing.ArrayOOBVariableDescriptorCountAllocate`

you will see the 2nd test fails because we now inline our LUT into the instrumented shader and the two tests have different values in the LUT that cause the wrong errors

We need to move the instrumentation cache hashing into `InstrumentShader` and use the `spirv::Settings` we pass into `spirv::Module`

PR #8856, opened 2024-11-13
gpu: Remove Shader Instrumentation Cache

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8850

The Shader Instrumentation Cache was added back when things were simpler, now as we move more and more things into the shader instrumentation, we need to be more careful of how we hash everything. We have recently been constantly getting false positives trying to maintain the complexity of this caching.

From some experiments, running one of our heavier traces of Doom External with 1500+ shaders we found on average the use of `VK_LAYER_GPUAV_CACHE_INSTRUMENTED_SHADERS` was only getting us from about 9.5 to 8.3 seconds in the load time. This shows that even at a worst case of having to load 1500+ shaders at once, it only takes a second longer to load the trace.

We have decide the complexity of trying to get this correct is currently not worth the small perf gain as shader instrumentation is not our bottleneck currently.


arno-lunarg (approved):
_So long old friend!_


(_**good riddance eheh**_)

PR #8857, opened 2024-11-13
layers: Removed unmaintained doxygen comments

spencer-lunarg:
These comments have been lingering around from day one of validation. They are not maintained, not complete, we aren't even confident they are accurate, and its not consistent with the rest of the code base

### DIFF
--- a/layers/stateless/stateless_validation.h
+++ b/layers/stateless/stateless_validation.h
@@ -82,83 +82,49 @@ class StatelessValidation : public ValidationObject {
     StatelessValidation() { container_type = LayerObjectTypeParameterValidation; }
     ~StatelessValidation() {}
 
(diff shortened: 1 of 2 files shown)


layers/stateless/stateless_validation.h
```diff
@@ -415,35 +285,14 @@ class StatelessValidation : public ValidationObject {
         return skip;
     }
 
-    /**
-     * Validate an array of Vulkan enumeration value.
-     *
-     * Process all enumeration token values in the specified array and generate a warning if a value
-     * does not fall within the core enumeration begin and end token values, and was not added to
```
spencer-lunarg, line 422: > enumeration begin and end token values

ya, these are very old :laughing:

artem-lunarg (approved):

PR #8860, opened 2024-11-14
layers: Add Array check for Inline Uniform Blocks

spencer-lunarg:
From https://gitlab.khronos.org/vulkan/vulkan/-/issues/4083

Add check to prevent having an array of descriptor when using `VK_DESCRIPTOR_TYPE_INLINE_UNIFORM_BLOCK`

### DIFF
--- a/layers/core_checks/cc_spirv.cpp
+++ b/layers/core_checks/cc_spirv.cpp
@@ -26,6 +26,7 @@
 
 #include <vulkan/vk_enum_string_helper.h>
 #include <vulkan/vulkan_core.h>
+#include "core_checks/cc_vuid_maps.h"
 #include "core_validation.h"
 #include "generated/spirv_grammar_helper.h"
 #include "state_tracker/shader_stage_state.h"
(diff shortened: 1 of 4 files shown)


layers/core_checks/cc_spirv.cpp
```diff
@@ -1944,6 +1945,16 @@ bool CoreChecks::ValidateShaderInterfaceVariablePipeline(const spirv::Module &mo
                    : "VUID-VkRayTracingPipelineCreateInfoNV-layout-07991";
     };
 
+    auto get_vuid_inline = [&pipeline]() {
```
arno-lunarg, line 1948: put lambda declaration right before it is needed. This things ain't free!

spencer-lunarg: I'm just copy and pasting the above code... this whole function I think has a stupid way of doing VUs... we need a better solution

spencer-lunarg: Will try and improve this in this PR though, while I'm here

PR #8864, opened 2024-11-14
gpu: Fix Sampler array checks

spencer-lunarg:
We had some disabled tests for GPU-AV because it was getting through `DescriptorValidator::ValidateDescriptor()` correctly.

The core issue is we ignore things when we don't know the index into an array on the CPU, but that can be ignored post process in GPU-AV because we are passing in the index

Added various tests to help with coverage

### DIFF
--- a/layers/drawdispatch/descriptor_validator.cpp
+++ b/layers/drawdispatch/descriptor_validator.cpp
@@ -296,6 +296,8 @@ static const spirv::ResourceInterfaceVariable *FindMatchingImageVar(const std::v
 
 bool DescriptorValidator::ValidateDescriptor(const DescriptorBindingInfo &binding_info, uint32_t index,
                                              VkDescriptorType descriptor_type, const ImageDescriptor &image_descriptor) const {
+    // We skip various parts of checks for core check to prevent false positive when we don't know the index
+    const bool is_gpu_av = dev_state.container_type == LayerObjectTypeGpuAssisted;
     std::vector<const Sampler *> sampler_states;
     const VkImageView image_view = image_descriptor.GetImageView();
     const ImageView *image_view_state = image_descriptor.GetImageViewState();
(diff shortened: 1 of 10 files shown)


tests/unit/gpu_av.cpp
```diff
@@ -879,11 +878,174 @@ TEST_F(NegativeGpuAV, DISABLED_YcbcrDrawFetchIndexed) {
...
     m_command_buffer.End();
+
+    m_errorMonitor->SetDesiredError("VUID-vkCmdDraw-None-06550");
+    m_default_queue->Submit(m_command_buffer);
+    m_default_queue->Wait();
+    m_errorMonitor->VerifyFound();
+}
+
+TEST_F(NegativeGpuAV, NonMultisampleMismatchWithPipeline) {
+    TEST_DESCRIPTION("Shader uses non-Multisample, but image view is.");
```
arno-lunarg, line 979:
```suggestion
    TEST_DESCRIPTION("Shader uses non-Multisample, but image view is ???.");
```

layers/drawdispatch/descriptor_validator.cpp
```diff
@@ -338,8 +340,9 @@ bool DescriptorValidator::ValidateDescriptor(const DescriptorBindingInfo &bindin
     const auto *variable = FindMatchingImageVar(binding_info.second, image_view_ci);
```
arno-lunarg, line 340:
```suggestion
    const auto *image_variable = FindMatchingImageVar(binding_info.second, image_view_ci);
```
or maybe better:
```suggestion
    const auto *image_spirv_var = FindMatchingImageVar(binding_info.second, image_view_ci);
```
since variable is a bloated term

arno-lunarg (approved):

tests/unit/gpu_av.cpp
spencer-lunarg: > Shader uses non-Multisample, but image view is <multisampled>>

I stole this from other test, can update

spencer-lunarg:
> CI Vulkan-ValidationLayers build # 18035 failed.

Un-related CI failure - all tests passed

PR #8867, opened 2024-11-15
layers: Renames and cleanup of ownership transfer code

artem-lunarg:
@spencer-lunarg This is some renames (it took me probably a year to realize that `OwnershipTransferBarrier` is better a match than  `QueueFamilyBarrier`) and some untemplation and simplification. But none of the potential changes we discussed earlier today. Thinking to do that after I fix the issue this PR is preparing for. Also found we don't have tests for some VUIDs, that's also in the plans.

### DIFF
--- a/layers/containers/qfo_transfer.h
+++ b/layers/containers/qfo_transfer.h
@@ -23,8 +23,7 @@
 
 // Types to store queue family ownership (QFO) Transfers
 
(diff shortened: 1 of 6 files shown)


spencer-lunarg (approved):
years overdue... great cleanup

PR #8871, opened 2024-11-18
layers: Fix message formatting in WSI validation

Eearslya:
Minor fix to a couple format strings in WSI validation. Extraneous `%s` in the stringstream output would result in messages like:
```
pCreateInfo->preTransform is not supported, support values are:
VK_SURFACE_TRANSFORM_IDENTITY_BIT_KHR%s
. The Vulkan spec states: ...
```


artem-lunarg (approved):
LGTM. @Eearslya thanks for the fix.

PR #8873, opened 2024-11-18
tests: Add ownership transfer tests for 09049 and 09052

artem-lunarg:
There is validation but not the tests for:
VUID-VkBufferMemoryBarrier-None-09049
VUID-VkImageMemoryBarrier-None-09052


spencer-lunarg (approved):
adding missing tests always makes me happy!

artem-lunarg:
VVL developers after new test is added
more-tests

PR #8877, opened 2024-11-18
GPU-AV: Add debug label regions to instrumentation error messages

arno-lunarg:
Still need to add it to debug printf

### DIFF
--- a/layers/gpu/instrumentation/gpuav_instrumentation.cpp
+++ b/layers/gpu/instrumentation/gpuav_instrumentation.cpp
@@ -25,6 +25,7 @@
 #include "gpu/shaders/gpuav_error_header.h"
 #include "gpu/debug_printf/debug_printf.h"
 
+#include "state_tracker/cmd_buffer_state.h"
 #include "state_tracker/shader_object_state.h"
 
 namespace gpuav {
@@ -394,18 +395,42 @@ void PreCallSetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer 
     const bool uses_robustness = (gpuav.enabled_features.robustBufferAccess || gpuav.enabled_features.robustBufferAccess2 ||
                                   (last_bound.pipeline_state && last_bound.pipeline_state->uses_pipeline_robustness));
 
+    std::optional<vvl::LabelCommand> deepest_opened_label_region = std::nullopt;
+    {
+        int32_t ended_label_regions_count = 0;
+        if (auto deepest_opened_label_region_rit =
+                std::find_if(cb_state.GetLabelCommands().rbegin(), cb_state.GetLabelCommands().rend(),
+                             [&ended_label_regions_count](const vvl::LabelCommand &label_cmd) {
+                                 if (label_cmd.begin) {
+                                     if (ended_label_regions_count == 0) {
+                                         return true;
+                                     } else {
+                                         --ended_label_regions_count;
+                                         return false;
+                                     }
+                                 } else {
+                                     ++ended_label_regions_count;
+                                     return false;
+                                 }
+                             });
+            deepest_opened_label_region_rit != cb_state.GetLabelCommands().rend()) {
+            deepest_opened_label_region = *deepest_opened_label_region_rit;
+        }
+    }
+
     CommandBuffer::ErrorLoggerFunc error_logger =
         [loc, descriptor_binding_index, descriptor_binding_list = &cb_state.descriptor_command_bindings,
-         cb_state_handle = cb_state.VkHandle(), bind_point, operation_index, uses_shader_object,
+         cb_state_handle = cb_state.VkHandle(), bind_point, deepest_opened_label_region, operation_index, uses_shader_object,
          uses_robustness](Validator &gpuav, const uint32_t *error_record, const LogObjectList &objlist) {
             bool skip = false;
 
             const DescriptorCommandBinding *descriptor_command_binding =
                 descriptor_binding_index != vvl::kU32Max ? &(*descriptor_binding_list)[descriptor_binding_index] : nullptr;
-            skip |= LogInstrumentationError(gpuav, cb_state_handle, objlist, operation_index, error_record,
-                                            descriptor_command_binding ? descriptor_command_binding->bound_descriptor_sets
-                                                                       : std::vector<DescriptorCommandBountSet>(),
-                                            bind_point, uses_shader_object, uses_robustness, loc);
+            skip |=
+                LogInstrumentationError(gpuav, cb_state_handle, objlist, deepest_opened_label_region, operation_index, error_record,
+                                        descriptor_command_binding ? descriptor_command_binding->bound_descriptor_sets
+                                                                   : std::vector<DescriptorCommandBountSet>(),
+                                        bind_point, uses_shader_object, uses_robustness, loc);
             return skip;
         };
 
@@ -518,7 +543,7 @@ bool LogMessageInstDescriptorIndexingOOB(Validator &gpuav, const uint32_t *error
             error_found = true;
         } break;
     }
-    out_error_msg = strm.str();
+    out_error_msg += strm.str();
     return error_found;
 }
 
@@ -588,7 +613,7 @@ bool LogMessageInstDescriptorClass(Validator &gpuav, const uint32_t *error_recor
             assert(false);  // other OOB checks are not implemented yet
     }
 
-    out_error_msg = strm.str();
+    out_error_msg += strm.str();
     return error_found;
 }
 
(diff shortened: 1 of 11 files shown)


layers/gpu/instrumentation/gpuav_instrumentation.h
```diff
@@ -17,13 +17,18 @@
 
 #pragma once
 
+#include "state_tracker/cmd_buffer_state.h"
```
spencer-lunarg, line 20: can we use forward reference on the Optional? I really really prefer to not start including large state objects in headers as this will not require a lot to depending on state object header

layers/gpu/instrumentation/gpuav_shader_instrumentor.cpp
```diff
@@ -1450,9 +1450,10 @@ static std::string FindShaderSource(std::ostringstream &ss, const std::vector<In
 
 // Where we build up the error message with all the useful debug information about where the error occured
 std::string GpuShaderInstrumentor::GenerateDebugInfoMessage(
-    VkCommandBuffer commandBuffer, const std::vector<Instruction> &instructions, uint32_t stage_id, uint32_t stage_info_0,
-    uint32_t stage_info_1, uint32_t stage_info_2, uint32_t instruction_position, const InstrumentedShader *instrumented_shader,
-    uint32_t shader_id, VkPipelineBindPoint pipeline_bind_point, uint32_t operation_index) const {
+    VkCommandBuffer commandBuffer, std::optional<vvl::CommandBuffer::LabelCommand> label_cmd,
```
spencer-lunarg, line 1453: why does this need to be an `optional`?

It really should be a `const reference` but seems you are using this because you haven't got DebugPrintf plumbed 

I guess I would have just had a pointer, since we do that already for `instrumented_shader` here and there are only 2 spots that call this function anyway

layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -694,7 +719,8 @@ bool LogMessageInstRayQuery(const uint32_t *error_record, std::string &out_error
 // sure it is available when the pipeline is submitted.  (The ShaderModule tracking object also
 // keeps a copy, but it can be destroyed after the pipeline is created and before it is submitted.)
 //
-bool LogInstrumentationError(Validator &gpuav, VkCommandBuffer cmd_buffer, const LogObjectList &objlist, uint32_t operation_index,
+bool LogInstrumentationError(Validator &gpuav, VkCommandBuffer cmd_buffer, const LogObjectList &objlist,
+                             std::optional<vvl::CommandBuffer::LabelCommand> label_command, uint32_t operation_index,
```
spencer-lunarg, line 723: we have `label_cmd`, `label_command` and `deepest_opened_label_region` for what is the same value being passed around... can be get a single consistent naming scheme

layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -394,18 +394,43 @@ void PreCallSetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer
     const bool uses_robustness = (gpuav.enabled_features.robustBufferAccess || gpuav.enabled_features.robustBufferAccess2 ||
                                   (last_bound.pipeline_state && last_bound.pipeline_state->uses_pipeline_robustness));
 
+    std::optional<vvl::CommandBuffer::LabelCommand> deepest_opened_label_region = std::nullopt;
+    {
+        int32_t ended_label_regions_count = 0;
+        if (auto deepest_opened_label_region_rit =
```
spencer-lunarg, line 400: nit: maybe try `const auto& label_commands = cb_state.GetLabelCommands()` or something, this is a crazy nesting inline going on, it is isolated, so not too worried

tests/framework/error_monitor.h
```diff
@@ -56,6 +56,9 @@ class ErrorMonitor {
     void SetDesiredFailureMsg(const VkFlags msg_flags, const char *const msg_string);
     void SetDesiredFailureMsg(const VkFlags msg_flags, const std::string &msg);
     void SetDesiredFailureMsgRegex(const VkFlags msg_flags, const char *vuid, std::string regex_str);
+    // Any error message matching undesired_regex_str will provoke a test failure
```
spencer-lunarg, line 59: All for this idea, but why not just have a dedicated `SetUndesiredMsgRegex`

Now in order to use this in another test,  I need to also use the failure regex when it would be nice to just go

```c++
SetDesiredError("VUID-Something");
SetUndesiredMsgRegex("Some thing I don't want to see in the error message");
```

layers/gpu/instrumentation/gpuav_instrumentation.h
arno-lunarg: no because `LabelCommand` is withing `vvl::CommandBuffer` : /

layers/gpu/instrumentation/gpuav_shader_instrumentor.cpp
arno-lunarg: We have to carry a value in the logging lambda, a pointer/reference would be dangling at error logging time. And that value may not be there if not within a debug label region, so have to use an optional value

layers/gpu/instrumentation/gpuav_instrumentation.cpp
arno-lunarg: I will remove `label_command`, but keep the others

layers/gpu/instrumentation/gpuav_instrumentation.cpp
arno-lunarg: I have to use reverse iterators, doing a reverse loop by hand is messy, it looks ass but at least I am sure the reverse loop is not messed up : /

tests/framework/error_monitor.h
arno-lunarg: I had that lol, but then I was like "what I really want is for some string to not be in one particular message, not necessarily all of them"

layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -394,18 +394,43 @@ void PreCallSetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer
...
+    }
+
     CommandBuffer::ErrorLoggerFunc error_logger =
         [loc, descriptor_binding_index, descriptor_binding_list = &cb_state.descriptor_command_bindings,
-         cb_state_handle = cb_state.VkHandle(), bind_point, operation_index, uses_shader_object,
+         cb_state_handle = cb_state.VkHandle(), bind_point, deepest_opened_label_region, operation_index, uses_shader_object,
          uses_robustness](Validator &gpuav, const uint32_t *error_record, const LogObjectList &objlist) {
             bool skip = false;
 
+            // #ARNO_TODO pass down label
```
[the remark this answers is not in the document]
arno-lunarg: ah leftover!

layers/gpu/instrumentation/gpuav_instrumentation.h
spencer-lunarg: so? That does not happen in `gpuav_instrumentation.h`... this header shouldn't need to know about `LabelCommand`, just forward reference it, all the writing happens inside the source files

layers/gpu/instrumentation/gpuav_instrumentation.cpp
spencer-lunarg: ya, fair, just caught me off guard, but again, its at least isolated

tests/framework/error_monitor.h
spencer-lunarg: ok, for now ok with this, not 100% sure how many other tests actually want this yet to decide on a good way to set the API

layers/gpu/instrumentation/gpuav_instrumentation.h
arno-lunarg: [For the compiler to know that `LabelCommand` is within ](https://stackoverflow.com/questions/1021793/how-do-i-forward-declare-an-inner-class)

arno-lunarg: But I can just move `LabelCommand` outside `vvl::CommandBuffer`

spencer-lunarg: for now, just add a TODO above this `#include` - this will effect Sync Val as well

spencer-lunarg: I rather get "sometime" working, can cleanup this stuff after, just don't want to forget, but ya, I think `LabelCommand` should live in `layers/error_message` somewhere

spencer-lunarg (approved):
LGTM when all comments are addressed

arno-lunarg:
Will add support for debug printf and tests with secondary command buffers tomorrow

layers/gpu/instrumentation/gpuav_instrumentation.h
arno-lunarg: aaaand did that

PR #8878, opened 2024-11-18
layers: Fix and cleanup ownership transfer code

artem-lunarg:
This removes `ValidatorState` ownership transfer helper. It was mostly additional abstraction layer that copied local variables and the result was stored as another local variable.

Also replaced `VUID-vkQueueSubmit-pSubmits-04626` with unassigned VUID when resource is EXCLUSIVE. 04626 is only about checking CONCURRENT resource properties. It also checks resource itself but not a barrier properties which is needed in unassigned VUID.

Unassigned VUID is about this requirement from `7.7.4. Queue Family Ownership Transfer`:
> The srcQueueFamilyIndex parameter of the barrier must be the source queue family index, and the dstQueueFamilyIndex parameter to the destination queue family index.

Opened gitlab issue: https://gitlab.khronos.org/vulkan/vulkan/-/issues/4092. Will open a MR too.

### DIFF
--- a/layers/core_checks/cc_synchronization.cpp
+++ b/layers/core_checks/cc_synchronization.cpp
@@ -1944,7 +1944,76 @@ void CoreChecks::EnqueueSubmitTimeValidateImageBarrierAttachment(const Location 
     }
 }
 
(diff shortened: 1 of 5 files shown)


spencer-lunarg (approved):
nice cleanup

PR #8876, opened 2024-11-18
gpu: Prevent false positives for aliased bindings

spencer-lunarg:
Prevents false positives - tracking issue https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8875

Issue #8875, opened 2024-11-18
Handle alias descriptor bindings


spencer-lunarg:
Things like https://godbolt.org/z/Kdh85abj8 will throw false positives in GPU-AV

The issue is we just get "binding 0 was accessed" out from GPU-AV post processing, but don't provide information which `OpVariable` was accessed and are just blindly taking the first one we find.

PR #8879, opened 2024-11-19
gpu: Remove Alias support to suppress false positives

spencer-lunarg:
We added a quick hack for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8875 in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8876, but this takes it a step further and just removes alias binding checks all together. This will allow us to more easily get other things correct, and then can return and try to apply them back in

### DIFF
--- a/layers/drawdispatch/descriptor_validator.cpp
+++ b/layers/drawdispatch/descriptor_validator.cpp
@@ -57,12 +57,9 @@ std::string DescriptorValidator::DescribeDescriptor(const DescriptorBindingInfo 
     ss << dev_state.FormatHandle(descriptor_set.Handle()) << " [Set " << set_index << ", Binding " << binding_info.first
        << ", Index " << index;
 
-    // If multiple variables tied to a binding, don't attempt to detect which one
-    if (binding_info.second.size() == 1) {
-        auto variable = binding_info.second[0].variable;
-        if (variable && !variable->debug_name.empty()) {
-            ss << ", variable \"" << variable->debug_name << "\"";
-        }
+    auto variable = binding_info.second.variable;
+    if (variable && !variable->debug_name.empty()) {
+        ss << ", variable \"" << variable->debug_name << "\"";
     }
     ss << "]";
     return ss.str();
@@ -152,6 +149,12 @@ bool DescriptorValidator::ValidateBindingDynamic(const DescriptorBindingInfo &bi
     auto binding_ptr = descriptor_set.GetBinding(binding_info.first);
     ASSERT_AND_RETURN_SKIP(binding_ptr);
     auto &binding = *binding_ptr;
+
+    // If false, we have already validated (and updated the state) for this descriptor on the CPU
+    if (!descriptor_set.ValidateBindingOnGPU(binding, binding_info.second.variable->is_dynamic_accessed)) {
+        return skip;
+    }
+
     switch (binding.descriptor_class) {
         case DescriptorClass::GeneralBuffer:
             skip |= ValidateDescriptorsDynamic(binding_info, static_cast<const BufferBinding &>(binding), indices);
@@ -224,11 +227,9 @@ bool DescriptorValidator::ValidateDescriptor(const DescriptorBindingInfo &bindin
             return true;
         }
         bool is_written_to = false;
-        for (const auto &req : binding_info.second) {
-            if (req.variable->IsWrittenTo()) {
-                is_written_to = true;
-                break;
-            }
+        const auto &descriptor_req = binding_info.second;
+        if (descriptor_req.variable->IsWrittenTo()) {
+            is_written_to = true;
         }
         if (is_written_to && dev_state.ValidateUnprotectedBuffer(cb_state, *buffer_node, loc, vuids.protected_command_buffer_02712,
                                                                  " (Buffer is in a descriptorSet)")) {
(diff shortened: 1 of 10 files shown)


layers/drawdispatch/descriptor_validator.cpp
```diff
@@ -152,6 +149,12 @@ bool DescriptorValidator::ValidateBindingDynamic(const DescriptorBindingInfo &bi
     auto binding_ptr = descriptor_set.GetBinding(binding_info.first);
     ASSERT_AND_RETURN_SKIP(binding_ptr);
     auto &binding = *binding_ptr;
+
+    // If false, we have already validated (and updated the state) for this descriptor on the CPU
+    if (!descriptor_set.ValidateBindingOnGPU(binding, binding_info.second.variable->is_dynamic_accessed)) {
```
spencer-lunarg, line 154: this is actually a BIG motivation for this change. I want to make the exact same call we do in the CPU to defer work off to the GPU to also no repeat work

Some false positives are things that we were validating on the CPU, just to have it re-run `ValidateDescriptor` in GPU-AV and get it wrong

layers/drawdispatch/descriptor_validator.cpp
```diff
@@ -238,61 +239,57 @@ bool DescriptorValidator::ValidateDescriptor(const DescriptorBindingInfo &bindin
     return false;
 }
 
-static const spirv::ResourceInterfaceVariable *FindMatchingImageVariable(const std::vector<DescriptorRequirement> &reqs,
+static const spirv::ResourceInterfaceVariable *FindMatchingImageVariable(const DescriptorRequirement &descriptor_req,
```
spencer-lunarg, line 242: this function (and `FindMatchingTexelVariable`) is strange, I "think" it could be removed, but wanted to try that in another PR as it might uncover some other issue

layers/gpu/descriptor_validation/gpuav_descriptor_validation.cpp
```diff
@@ -188,11 +189,17 @@ void UpdateBoundDescriptors(Validator &gpuav, CommandBuffer &cb_state, VkPipelin
...
                 while (iter != bound_descriptor_set.binding_req_map.end() && iter->first == binding) {
-                    binding_info.second.emplace_back(iter->second);
+                    descriptor_req_found++;
+                    binding_info.second = iter->second;
                     ++iter;
                 }
-                context.ValidateBindingDynamic(binding_info, u.second);
+                // TODO - https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8875
+                // Currently if we find multiple variables (aka aliased bindings) we will skip validating
+                if (descriptor_req_found == 1) {
```
spencer-lunarg, line 200: I am not sure if this can be zero, it shouldn't be able to, but will need more testing

tests/unit/gpu_av.cpp
```diff
@@ -693,116 +689,6 @@ TEST_F(NegativeGpuAV, UnnormalizedCoordinatesSeparateSamplerSharedSampler) {
...
-    descriptor_set.WriteDescriptorImageInfo(1, image_view, VK_NULL_HANDLE, VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE);
-    descriptor_set.UpdateDescriptorSets();
-
-    m_command_buffer.Begin();
-    m_command_buffer.BeginRenderPass(m_renderPassBeginInfo);
-    vk::CmdBindDescriptorSets(m_command_buffer.handle(), VK_PIPELINE_BIND_POINT_GRAPHICS, pipeline_layout.handle(), 0, 1,
-                              &descriptor_set.set_, 0, nullptr);
-    vk::CmdBindPipeline(m_command_buffer.handle(), VK_PIPELINE_BIND_POINT_GRAPHICS, g_pipe.Handle());
-
-    if (!m_gpuav_disable_core) {
```
spencer-lunarg, line 793: we had these in some test... this was sign that this is actually just validated on the CPU and doesn't need GPU-AV

layers/drawdispatch/descriptor_validator.cpp
spencer-lunarg: the tests failing are showing me that there is actually just some larger problems at play... glad we had tests, but things are kinda a mess right now trying to share this `ValidateDescriptor` on the CPU/GPU and having subtle things in it only for one or the other code path

layers/gpu/descriptor_validation/gpuav_descriptor_validation.cpp
```diff
@@ -16,6 +16,7 @@
  */
 
 #include "gpu/descriptor_validation/gpuav_descriptor_validation.h"
+#include <cstdint>
```
arno-lunarg, line 19: should not be needed?

layers/gpu/descriptor_validation/gpuav_descriptor_validation.cpp
```diff
@@ -188,11 +189,17 @@ void UpdateBoundDescriptors(Validator &gpuav, CommandBuffer &cb_state, VkPipelin
                 auto iter = bound_descriptor_set.binding_req_map.find(binding);
                 vvl::DescriptorBindingInfo binding_info;
                 binding_info.first = binding;
+                uint32_t descriptor_req_found = 0;
                 while (iter != bound_descriptor_set.binding_req_map.end() && iter->first == binding) {
-                    binding_info.second.emplace_back(iter->second);
+                    descriptor_req_found++;
+                    binding_info.second = iter->second;
```
arno-lunarg, line 195: Needless `vector` copies will be done here, just have one copy during the last iteration
Also, do we truly need a deep copy?

layers/state_tracker/descriptor_sets.h
```diff
@@ -983,7 +983,7 @@ class DescriptorSet : public StateObject {
         return DescriptorIterator<ConstBindingIterator>(*this, binding, index);
     }
 
-    virtual bool SkipBinding(const DescriptorBinding &binding, bool is_dynamic_accessed) const {
+    inline bool ValidateBindingOnGPU(const DescriptorBinding &binding, bool is_dynamic_accessed) const {
```
arno-lunarg, line 986:
```suggestion
   bool ValidateBindingOnGPU(const DescriptorBinding &binding, bool is_dynamic_accessed) const {
```
Function defined in a header are likely to be inlined anyway, and `inline` being just a hint, just let the compiler decide!

arno-lunarg (commented):
I think there are needless deep vector copies, aside that LGTM

layers/state_tracker/descriptor_sets.h
artem-lunarg: Yeah, inline is mostly outdated now, compilers inline not only little but medium sized functions too, results in mostly large functions in the output, and function calls are less expensive as a result 🙃

artem-lunarg: I guess another part of inline that deals with One Definition rule can more central reason to use it

spencer-lunarg:
closing - taking different approach

PR #8881, opened 2024-11-19
layers: Renaming related to debug label regions

arno-lunarg:
Multiple renaming relate to debug label regions.

I was working on adding support for debug label regions in GPU-AV, and looking for what we already had was hard. Few of the old names related to Vulkan functions/structs mentioned in the specification, making looking for things needlessly cumbersome. 
`LabelCommand` for instance appears nowhere in the spec, and this struct essentially is a wrapper for `VkDebugUtilsLabelEXT`, so I just renamed it to `DebugUtilsLabel` for easy lookup. I tried to base all renaming I did from Vulkan spec excerpts: things like "debug label region", "label region". 
Following the [Google style guide](https://google.github.io/styleguide/cppguide.html#General_Naming_Rules), I also tried to use `label_region` instead of the full blown `debug_label_region` for local variables, I think the small scope allows to use the shorter version.

Doing this allowed me to find that Artem has done some great work in that area  that GPU-AV could use. So it was a shame that I could not find it easily!

### DIFF
--- a/layers/error_message/logging.h
+++ b/layers/error_message/logging.h
@@ -105,10 +105,12 @@ static inline uint64_t HandleToUint64(HANDLE_T h) {
 
 static inline uint64_t HandleToUint64(uint64_t h) { return h; }
 
+// #ARNO_TODO Move LabelCommand/DebugUtilsLabel here
+
 // Data we store per label for logging
 struct LoggingLabel {
-    std::string name;
-    std::array<float, 4> color;
+    std::string name{};
+    std::array<float, 4> color{};
 
     void Reset() { *this = LoggingLabel(); }
     bool Empty() const { return name.empty(); }
@@ -125,8 +127,6 @@ struct LoggingLabel {
         if (label_info && label_info->pLabelName) {
             name = label_info->pLabelName;
             std::copy_n(std::begin(label_info->color), 4, color.begin());
-        } else {
-            Reset();
         }
     }
 
@@ -144,20 +144,13 @@ struct LoggingLabelState {
     LoggingLabel insert_label;
 
     // Export the labels, but in reverse order since we want the most recent at the top.
-    std::vector<VkDebugUtilsLabelEXT> Export() const {
-        size_t count = labels.size() + (insert_label.Empty() ? 0 : 1);
-        std::vector<VkDebugUtilsLabelEXT> out(count);
-
-        if (!count) return out;
-
-        size_t index = count - 1;
+    void Export(std::vector<VkDebugUtilsLabelEXT> &exported_labels) const {
         if (!insert_label.Empty()) {
-            out[index--] = insert_label.Export();
+            exported_labels.emplace_back(insert_label.Export());
         }
-        for (const auto &label : labels) {
-            out[index--] = label.Export();
-        }
-        return out;
+        exported_labels.reserve(exported_labels.size() + labels.size());
+        std::for_each(labels.rbegin(), labels.rend(),
+                      [&exported_labels](const LoggingLabel &label) { exported_labels.emplace_back(label.Export()); });
     }
 };
 
--- a/layers/state_tracker/cmd_buffer_state.cpp
+++ b/layers/state_tracker/cmd_buffer_state.cpp
@@ -299,7 +299,7 @@ void CommandBuffer::ResetCBState() {
     // Clean up the label data
     debug_label.Reset();
     label_stack_depth_ = 0;
-    label_commands_.clear();
+    debug_label_regions_.clear();
 
     nesting_level = 0;
 
@@ -1142,7 +1142,8 @@ void CommandBuffer::ExecuteCommands(vvl::span<const VkCommandBuffer> secondary_c
         }
 
         label_stack_depth_ += sub_cb_state->label_stack_depth_;
-        label_commands_.insert(label_commands_.end(), sub_cb_state->label_commands_.begin(), sub_cb_state->label_commands_.end());
+        debug_label_regions_.insert(debug_label_regions_.end(), sub_cb_state->debug_label_regions_.begin(),
+                                    sub_cb_state->debug_label_regions_.end());
     }
 }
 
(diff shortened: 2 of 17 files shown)


layers/error_message/logging.h
```diff
@@ -105,10 +105,12 @@ static inline uint64_t HandleToUint64(HANDLE_T h) {
 
 static inline uint64_t HandleToUint64(uint64_t h) { return h; }
 
+// #ARNO_TODO Move LabelCommand/DebugUtilsLabel here
```
arno-lunarg, line 108: Doing this was my first goal, then things got where they are : 3

layers/error_message/logging.h
```diff
@@ -144,20 +144,13 @@ struct LoggingLabelState {
...
 
     // Export the labels, but in reverse order since we want the most recent at the top.
-    std::vector<VkDebugUtilsLabelEXT> Export() const {
-        size_t count = labels.size() + (insert_label.Empty() ? 0 : 1);
-        std::vector<VkDebugUtilsLabelEXT> out(count);
-
-        if (!count) return out;
-
-        size_t index = count - 1;
+    void Export(std::vector<VkDebugUtilsLabelEXT> &exported_labels) const {
```
arno-lunarg, line 147: Reworked `Export` to avoid needless copies

layers/error_message/logging.h
```diff
@@ -144,20 +144,13 @@ struct LoggingLabelState {
...
-            out[index--] = insert_label.Export();
+            exported_labels.emplace_back(insert_label.Export());
         }
-        for (const auto &label : labels) {
-            out[index--] = label.Export();
-        }
-        return out;
+        exported_labels.reserve(exported_labels.size() + labels.size());
+        std::for_each(labels.rbegin(), labels.rend(),
+                      [&exported_labels](const LoggingLabel &label) { exported_labels.emplace_back(label.Export()); });
```
arno-lunarg, line 153: The original code was not doing that, but I think I need to check that `label` is not `Empty` before exporting it?

layers/state_tracker/cmd_buffer_state.cpp
```diff
@@ -1810,21 +1811,21 @@ void CommandBuffer::GetCurrentPipelineAndDesriptorSets(VkPipelineBindPoint pipel
...
-void CommandBuffer::ReplayLabelCommands(const vvl::span<const LabelCommand> &label_commands,
-                                        std::vector<std::string> &label_stack) {
-    for (const LabelCommand &command : label_commands) {
-        if (command.begin) {
-            label_stack.emplace_back(command.label_name.empty() ? "(empty label)" : command.label_name);
+void CommandBuffer::ReplayDebugLabelRegions(const vvl::span<const DebugUtilsLabel> &label_regions,
+                                            std::vector<std::string> &label_stack) {
+    for (const DebugUtilsLabel &label_region : label_regions) {
+        if (label_region.begin) {
+            label_stack.emplace_back(label_region.label_name.empty() ? "(empty label)" : label_region.label_name);
```
arno-lunarg, line 1828: I could just now rename `DebugUtilsLabel::label_name` to `DebugUtilsLabel::name`, no need to reming `label_`
```suggestion
            label_stack.emplace_back(label_region.name.empty() ? "(empty label)" : label_region.name);
```

layers/state_tracker/cmd_buffer_state.h
```diff
@@ -150,7 +150,7 @@ class CommandPool : public StateObject {
     void Destroy() override;
 };
 
-struct LabelCommand {
+struct DebugUtilsLabel {
```
artem-lunarg, line 153: What I sent in internal chat:
> I think it's wrong to rename LabelCommand into DebugUtilsLabel because label command is just a command CmdBeginLabel or CmdEndLabel, and is not a debug region. Debug region is described at least by two commands, but in general stack of commands define the region.

artem-lunarg: Another example , LabelCommand associated with vkCmdEndDebugUtilsLabelEXT does not have a label_name and just a boolean flag. In order to properly define a a debug region we need an entire sequence of label command (begin/end and how are relative to one another, e.g. could be (begin begin end begin end end) and only having all sequence allows to define the final label region. This structure is needed to track not only debug region within a command buffer, but only between command bufers and queue submissions to a single queue.

arno-lunarg: Oh ok that's why you named it like so
But yeah it just wraps a VkDebugUtilsLabelEXT, something that is use both to define regions and also just insert a label. `DebugUtilsLabel` itself does not directly denote debug regions.
`std::vector<DebugUtilsLabel> debug_label_regions_` holds the regions, hence `regions` in the name

layers/sync/sync_commandbuffer.cpp
```diff
@@ -983,10 +983,10 @@ void CommandBufferAccessContext::AddSubcommandHandle(ResourceUsageTag tag, const
     }
 }
 
-std::string CommandBufferAccessContext::GetDebugRegionName(const ResourceUsageRecord &record) const {
-    const bool use_proxy = !proxy_label_commands_.empty();
-    const auto &label_commands = use_proxy ? proxy_label_commands_ : cb_state_->GetLabelCommands();
-    return vvl::CommandBuffer::GetDebugRegionName(label_commands, record.label_command_index);
+std::string CommandBufferAccessContext::GetStackedDebugLabelRegionName(const ResourceUsageRecord &record) const {
```
artem-lunarg, line 986: > GetStackedDebugLabelRegionName

There is no such notion as "stecked debug region name",  we don't have a non-stacked debug region. Debug region is always defined with regard to all previous history of label commands.

For the context. Development of this feature took around of entire month and there was few iterations of naming. There are some areas related to naming that could be improved (like shorter names in some places) but I don't think that change to general terms here is a right direction, I see it this a mostly misleading change. Maybe more comments can be added to clarify terminology.

arno-lunarg: Yeah I was not sure about this one, if a debug region is defined as you say then `stacked` adds no value.
But I don't want to go back to `LabelCommand` , again it appears nowhere in the spec so you have to know the VVL code to re-learn how debug regions works. Also `labe_command` was turning into just `command` at some spots, adding to the confusion.

layers/error_message/logging.h
artem-lunarg: Well, actually this should be investigated. It looks Queue works with final name as std::string, and the reason LabelCommands are inside command buffer is because that's the place where label commands are recorded.  It should be fine to move this definition to some error/debug file, but command buffer is probably is still the primary object that owns them, and other types access them when needed. When there is no need to track separate command a we can get string the command can be converted to std::string, so the only reason it's array of command because in that context we are not ready yet to convert to final string.

layers/sync/sync_commandbuffer.cpp
artem-lunarg: I don't follow. This change started with completely wrong definitions, that's the opposite of improving. My suggestion is to spend enough time working with specific piece of codebase before doing major cleanups to have enough confidence of how it works and understanding. 

Also longer and more detailed names are not always improvement. It could be additional noise. When working on part of new codebase it's fine that initially thing are not very familiar. Should not be a trigger to make a change. One of my experiences working on syncval, where some parts are not-obvious, it helps to work with those parts for months and then it's usually more measured decisions if something needs cleanup or renames.

artem-lunarg: > What are the wrong definitions?

LabelCommand  was renamed to DebugUtilsLabel. It's not clarification. Hypothetically leaving this new name makes code logic wrong in a sense the name does not correlate with what the code does.

> Are you not managing debug label regions in syncval?

There is some code there, but the foundational are in the core. I think core does all the necessary state tracking related to labels. Then syncval  queries what it needs, maybe does additional operations specific to syncval.

PR #8883, opened 2024-11-19
layers: Rework LoggingLabel::Export

arno-lunarg:
Remove the need for a staging std::vector


layers/error_message/logging.h
```diff
@@ -144,20 +142,16 @@ struct LoggingLabelState {
...
+    void Export(std::vector<VkDebugUtilsLabelEXT> &exported_labels) const {
         if (!insert_label.Empty()) {
-            out[index--] = insert_label.Export();
+            exported_labels.emplace_back(insert_label.Export());
         }
-        for (const auto &label : labels) {
-            out[index--] = label.Export();
-        }
-        return out;
+        exported_labels.reserve(exported_labels.size() + labels.size());
```
artem-lunarg, line 149: nit: since reserve is used, maybe it can be reserved before `insert_label` so removes one more allocation

artem-lunarg (approved):

layers/error_message/logging.h
arno-lunarg: True!

PR #8884, opened 2024-11-19
gpu: Reorganize BDA used in shader

spencer-lunarg:
Currently we do our updating at `vkCmdBindDescriptorSets` time, but there are things we care about at a per-draw time

This tries and removes the Post Processing from the "What descriptor type is it".

Trying to find a balance of sharing logic, but also keeping things separated to allow different parts of GPU-AV to be more efficient/accurate where needed


layers/gpu/resources/gpuav_shader_resources.h
```diff
@@ -50,18 +50,18 @@ struct DescriptorCommandBinding {
...
-// Not all GPU-AV code uses each, but group together for ease of managing the memory
-struct DescriptorSetRecord {
-    // The type information will change with UpdateAfterBind so will need to update this before submitting the to the queue
-    VkDeviceAddress ds_type;  // input data
-
-    VkDeviceAddress descriptor_index_post_process;  // output data
-};
-
 // Shared among all Descriptor Indexing GPU-AV checks (so we only have to create a single buffer)
 struct DescriptorStateSSBO {
```
spencer-lunarg, line 54: main idea behind this change, the idea is all we are passing in our BDA pointer addresses and no shader needed both `ds_type` and `descriptor_index_post_process` so they can just be their own arrays (and might even just be moved out as different bindings in the future)

layers/gpu/resources/gpuav_subclasses.h
```diff
@@ -50,9 +50,18 @@ struct DebugPrintfBufferInfo {
...
 class CommandBuffer : public vvl::CommandBuffer {
   public:
-    // per vkCmdBindDescriptorSet() state
+    // One item per vkCmdBindDescriptorSet() called
+    // We really need this information at the draw/dispatch/action, but assume most apps will either
+    //  1. Update once and only change pipelines, so this information doesn't change between draws
+    //  2. Change once per draw, then this would be the same length as doing this per-draw
+    //
+    // Note: If the app calls vkCmdBindDescriptorSet 10 times to set descriptor set [0, 9] one at a time instead of setting [0, 9]
+    // in a single vkCmdBindDescriptorSet call than this will allocate a lot of redundant memory
```
jeremyg-lunarg, line 59: s/than/then/

jeremyg-lunarg (approved):

layers/gpu/resources/gpuav_subclasses.h
spencer-lunarg: applied fix to https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8890

spencer-lunarg:
> CI Vulkan-ValidationLayers build # 18059 failed.

Unrelated CI failure on Android

PR #8886, opened 2024-11-19
gpuav: Update debug label regions management

arno-lunarg:
Took some of what Artem did for syncval
Add debug label regions for debug printf
Still need to add support for secondary command buffers, not so easy

EDIT: Updated PR to add support for debug label regions starting in one (primary) command buffer and ending in another following one. To do that I drew inspiration from sync val, but had to come up with my own (little) system, for GPU-AV has different constraints. I described the problem in a comment above this new `vvl::CommandBuffer` method: https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/19390298820aceb3881e9865f05eabdd27cecf41/layers/state_tracker/cmd_buffer_state.h#L710

EDIT 2: Reworked again to account for Artem's comments. Removed debug label region tracking from the `CommandBuffer` class, now it is tracked on a "per command buffer submission" basis, in the `CommandBufferSubmission` struct. Added new test `StageInfoWithDebugLabel7` that was not working without this enhancement. 
Realized that support for debug printf was broken (https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8924) so removed it

### DIFF
--- a/layers/gpu/instrumentation/gpuav_instrumentation.cpp
+++ b/layers/gpu/instrumentation/gpuav_instrumentation.cpp
@@ -411,44 +411,26 @@ void PreCallSetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer 
     const bool uses_robustness = (gpuav.enabled_features.robustBufferAccess || gpuav.enabled_features.robustBufferAccess2 ||
                                   (last_bound.pipeline_state && last_bound.pipeline_state->uses_pipeline_robustness));
 
-    std::optional<vvl::LabelCommand> deepest_opened_label_region = std::nullopt;
-    {
-        int32_t ended_label_regions_count = 0;
-        if (auto deepest_opened_label_region_rit =
-                std::find_if(cb_state.GetLabelCommands().rbegin(), cb_state.GetLabelCommands().rend(),
-                             [&ended_label_regions_count](const vvl::LabelCommand &label_cmd) {
-                                 if (label_cmd.begin) {
-                                     if (ended_label_regions_count == 0) {
-                                         return true;
-                                     } else {
-                                         --ended_label_regions_count;
-                                         return false;
-                                     }
-                                 } else {
-                                     ++ended_label_regions_count;
-                                     return false;
-                                 }
-                             });
-            deepest_opened_label_region_rit != cb_state.GetLabelCommands().rend()) {
-            deepest_opened_label_region = *deepest_opened_label_region_rit;
-        }
-    }
+    const uint32_t last_label_command_i =
+        !cb_state.GetLabelCommands().empty() ? uint32_t(cb_state.GetLabelCommands().size() - 1) : vvl::kU32Max;
 
-    CommandBuffer::ErrorLoggerFunc error_logger =
-        [loc, descriptor_binding_index, descriptor_binding_list = &cb_state.descriptor_command_bindings,
-         cb_state_handle = cb_state.VkHandle(), bind_point, deepest_opened_label_region, operation_index, uses_shader_object,
-         uses_robustness](Validator &gpuav, const uint32_t *error_record, const LogObjectList &objlist) {
-            bool skip = false;
+    CommandBuffer::ErrorLoggerFunc error_logger = [loc, descriptor_binding_index,
+                                                   descriptor_binding_list = &cb_state.descriptor_command_bindings, bind_point,
+                                                   last_label_command_i, operation_index, uses_shader_object,
+                                                   uses_robustness](Validator &gpuav, const CommandBuffer &cb_state,
+                                                                    const uint32_t *error_record, const LogObjectList &objlist,
+                                                                    const std::vector<std::string> &initial_label_stack) {
+        bool skip = false;
 
-            const DescriptorCommandBinding *descriptor_command_binding =
-                descriptor_binding_index != vvl::kU32Max ? &(*descriptor_binding_list)[descriptor_binding_index] : nullptr;
-            skip |=
-                LogInstrumentationError(gpuav, cb_state_handle, objlist, deepest_opened_label_region, operation_index, error_record,
+        const DescriptorCommandBinding *descriptor_command_binding =
+            descriptor_binding_index != vvl::kU32Max ? &(*descriptor_binding_list)[descriptor_binding_index] : nullptr;
+        skip |= LogInstrumentationError(gpuav, cb_state, objlist, initial_label_stack, last_label_command_i, operation_index,
+                                        error_record,
                                         descriptor_command_binding ? descriptor_command_binding->bound_descriptor_sets
                                                                    : std::vector<std::shared_ptr<DescriptorSet>>(),
                                         bind_point, uses_shader_object, uses_robustness, loc);
-            return skip;
-        };
+        return skip;
+    };
 
     cb_state.per_command_error_loggers.emplace_back(error_logger);
 }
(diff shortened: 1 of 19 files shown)


layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -395,42 +395,23 @@ void PreCallSetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer
...
-                                     return false;
-                                 }
-                             });
-            deepest_opened_label_region_rit != cb_state.GetLabelCommands().rend()) {
-            deepest_opened_label_region = *deepest_opened_label_region_rit;
-        }
+    std::string debug_region_name;
+    if (!cb_state.GetLabelCommands().empty()) {
+        const uint32_t last_label_command_i = static_cast<uint32_t>(cb_state.GetLabelCommands().size() - 1);
+        debug_region_name = cb_state.GetDebugRegionName(cb_state.GetLabelCommands(), last_label_command_i);
```
arno-lunarg, line 401: This is how I use `GetDebugRegionName`, anything looking wrong @artem-lunarg ?

artem-lunarg: Looks good. Not sure what is the scope in gpu-av. If gpu-av needs also to consider global label state then `GetDebugRegionName()` provides  `initial_label_stack` parameter and it can be used to add that support to.

Labels command influence is not limited to command buffer in general. EndLabel can be in different command buffer. So after specific queue submission the queue can have current non-empty label name (per-queue state). It can be used as `initial_label_stack` to prepend when replaying command from current command buffer.

arno-lunarg: Ultimately we want the same level of support as sync val, so this is valuable info! I need to look at how you do those things and will probably ping you again on this topic

artem-lunarg: Just refreshed memory how queue label state is tracked. It's vvl::Queue and `cmdbuf_label_stack` field.
Example how it is used in syncval: https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/fdbdd359fb2bbb50352e84b259171813957d1ee3/layers/sync/sync_validation.cpp#L3202

One difference of this queue state form command buffer label tracking that on queue level it's enough to keep std::vector<string> of nested labels, in contrast to LabelCommand in the command buffer. The reason is because after command buffer is submitted we know exact label hierarchy (just replay label commands from entire submitted command buffer) so we just save result of this replay.

artem-lunarg (approved):
LGTM if CI is happy

layers/state_tracker/cmd_buffer_state.h
```diff
@@ -698,6 +699,15 @@ class CommandBuffer : public RefcountedStateObject {
 
     bool IsPrimary() const { return allocate_info.level == VK_COMMAND_BUFFER_LEVEL_PRIMARY; }
     bool IsSecondary() const { return allocate_info.level == VK_COMMAND_BUFFER_LEVEL_SECONDARY; }
+
+    // Specifically made for GPU-AV, for it has unique problems: Error reporting is done *after* command buffer submissions,
+    // not at Pre/PostCall time. Contrary to sync val, GPU-AV cannot just look at `GetQueueState()->cmdbuf_label_stack`
+    // to construct an initial label stack. sync-val can do that because validation and error reporting is done
+    // *before* a command buffer list is submitted: validation is performed one command buffer at a time,
+    // and `GetQueueState()->cmdbuf_label_stack` is updated between those validations.
+    // When GPU-AV starts doing error reporting, the label stack info stored in Queue state is lost.
```
artem-lunarg, line 708: @arno-lunarg could you provide more details on the following parts, to understand this better

> GPU-AV cannot just look at `GetQueueState()->cmdbuf_label_stack`

> // When GPU-AV starts doing error reporting, the label stack info stored in Queue state is lost.

From what I remember, Queue tracks label stack globally, it reflects state after the labels from the previously submitted command buffers are applied. And then when we start processing new command buffer then we use global queue state to and call it the initial label state (at the beginning of current command buffer). It's not clear how the point (Pre/Post) that was used to update Queue can affect processing of subsequent command buffer.

artem-lunarg: > the label stack info stored in Queue state is lost.

This also not clear. Label stack in Queue is never lost. It reflects current queue state after all past submissions.

layers/state_tracker/state_tracker.cpp
```diff
@@ -1264,7 +1264,8 @@ void ValidationStateTracker::PostCallRecordQueueSubmit(VkQueue queue, uint32_t s
     for (uint32_t submit_idx = 0; submit_idx < submitCount; submit_idx++) {
         const VkSubmitInfo *submit = &pSubmits[submit_idx];
         for (uint32_t i = 0; i < submit->commandBufferCount; i++) {
-            if (auto cb_state = GetRead<vvl::CommandBuffer>(submit->pCommandBuffers[i])) {
+            if (auto cb_state = GetWrite<vvl::CommandBuffer>(submit->pCommandBuffers[i])) {
+                cb_state->SetPostProcessInitialLabelStack(queue_state->cmdbuf_label_stack);
```
artem-lunarg, line 1268: Related to the previous comment, this place is not obvious. I guess in general we don't update command buffers during QueueSubmit since they can be shared.

layers/state_tracker/cmd_buffer_state.h
artem-lunarg: Had an internal discussion. Arno explained how GPU-AV reads results, and it's different timeline then where we track queue progress.

arno-lunarg:
Had a call with Artem, maintaining some state per command buffer looks wrong. And state needs to be maintained per command buffer _submission_, not per command buffer: the same command buffer could be submitted multiple times, so maintaining some state inside a command buffer would have it overwritten when looping through the submitted command buffers

arno-lunarg:
EDIT 2: Reworked again to account for Artem's comments. Removed debug label region tracking from the `CommandBuffer` class, now it is tracked on a "per command buffer submission" basis, in the `CommandBufferSubmission` struct. Added new test `StageInfoWithDebugLabel7` that was not working without this enhancement.

layers/state_tracker/queue_state.h
```diff
@@ -34,11 +35,33 @@ namespace vvl {
...
+    // Specifically made for GPU-AV, for it has unique problems: Error reporting is done *after*
+    // command buffer submissions, not at Pre/PostCall time.
+    // Contrary to sync val, GPU-AV cannot just look at `GetQueueState()->cmdbuf_label_stack`
+    // to construct an initial label stack. sync-val can do that because validation and error reporting is done
+    // *before* a command buffer list is submitted: validation is performed one command buffer at a time,
+    // and `GetQueueState()->cmdbuf_label_stack` is updated between those validations.
+    // When GPU-AV starts doing error reporting, when command buffers have completed,
+    // the label stack info stored in Queue state is lost.
+    // => GPU-AV needs to track this initial label stack per command buffer submission.
+    std::vector<std::string> initial_label_stack;
```
artem-lunarg, line 49: As a future improvement I wonder if this can be put into GPU-AV subclasses, to minimize changes to core types.

layers/state_tracker/state_tracker.cpp
```diff
@@ -1198,38 +1214,39 @@ void ValidationStateTracker::PreCallRecordQueueSubmit(VkQueue queue, uint32_t su
...
-                value = timeline_semaphore_submit->pWaitSemaphoreValues[i];
+                (wait_sem_i < timeline_semaphore_submit->waitSemaphoreValueCount)) {
+                value = timeline_semaphore_submit->pWaitSemaphoreValues[wait_sem_i];
             }
-            submission.AddWaitSemaphore(Get<vvl::Semaphore>(submit->pWaitSemaphores[i]), value);
+            submission.AddWaitSemaphore(Get<vvl::Semaphore>(submit->pWaitSemaphores[wait_sem_i]), value);
         }
 
-        for (uint32_t i = 0; i < submit->signalSemaphoreCount; ++i) {
+        for (uint32_t sig_sem_i = 0; sig_sem_i < submit->signalSemaphoreCount; ++sig_sem_i) {
```
artem-lunarg, line 1231: I can't say it's easier to read this rename, I actually starting to spend time to decode sig_sem_i sequence. IMO it's valuable property or very nested scopes is that they allow to use shorter names.

artem-lunarg (approved):
LGTM. One future suggestion to consider if it's possible to store initial label stack in parallel GPU-AV specific structures.

layers/state_tracker/state_tracker.cpp
arno-lunarg: Yes true, as long as it is not `index` or `idx` I am happy!

arno-lunarg:
Update: Rename `wait_sem_i` and `sig_sem_i` to `i`, replaced some `auto` usage in `queue_state.cpp`

Issue #8745, opened 2024-10-23
VUID-vkDestroyFence-fence-01120 message


ziga-lunarg:
I'm running into `VUID-vkDestroyFence-fence-01120`, but the message doesn't make much sense. 

> Validation Error: [ VUID-vkDestroyFence-fence-01120 ] | MessageID = 0x5d296248 | vkDestroyFence(): fence can't be called on VkFence 0xf56c9b0000000004[] that is currently in use by VkFence 0xf56c9b0000000004[].

The word `fence` seems extra in that sentence, but more importantly the fence is in use by itself?

Test to reproduce:

```
TEST_F(PositiveSyncObject, DestroyFenceInUse) {
    RETURN_IF_SKIP(Init());

    VkFenceCreateInfo fci = vku::InitStructHelper();
    fci.flags = VK_FENCE_CREATE_SIGNALED_BIT;
    vkt::Fence fence(*m_device, fci);

    vk::WaitForFences(device(), 1, &fence.handle(), VK_TRUE, kWaitTimeout);
    vk::ResetFences(device(), 1u, &fence.handle());

    m_default_queue->Submit(vkt::no_cmd, fence);
}
```

artem-lunarg:
It looks similar issues with other objects, saw the same in Semaphore messages.

User21T:
Good day. The fence is possibly in state where it can't be destroyed. Maybe some process, e.g. submission command at the end of the test, is still using it. I think we could repeat `vk::WaitForFences` after `m_default_queue->Submit(vkt::no_cmd, fence)` to make sure this command has finished and no processes use the fence.

PR #8892, opened 2024-11-20
layers: Fix in-use message for fence and semaphore

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8745

before: fence can't be called on VkFence that is in use by VkFence 🥹
> Validation Error: [ VUID-vkDestroyFence-fence-01120 ] | MessageID = 0x5d296248 | vkDestroyFence(): fence can't be called on VkFence 0xf56c9b0000000004[] that is currently in use by VkFence 0xf56c9b0000000004[].
The Vulkan spec states: All queue submission commands that refer to fence must have completed execution (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkDestroyFence-fence-01120)

after fix:
> Validation Error: [ VUID-vkDestroyFence-fence-01120 ] | MessageID = 0x5d296248 | vkDestroyFence():  can't be called on VkFence 0xf56c9b0000000004[] that is currently in use by VkQueue 0x25822fafa70[].
The Vulkan spec states: All queue submission commands that refer to fence must have completed execution (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkDestroyFence-fence-01120)

### DIFF
--- a/layers/state_tracker/semaphore_state.cpp
+++ b/layers/state_tracker/semaphore_state.cpp
@@ -54,6 +54,46 @@ vvl::Semaphore::Semaphore(ValidationStateTracker &dev, VkSemaphore handle, const
       dev_data_(dev) {
 }
 
+const VulkanTypedHandle *vvl::Semaphore::InUse() const {
+    auto guard = ReadLock();
+    // Semaphore does not have a parent (in the sense of a VVL state object), and the value returned
+    // by the base class InUse is not useful for reporting (it is the semaphore's own handle)
+    const bool in_use = RefcountedStateObject::InUse() != nullptr;
+    if (!in_use) {
+        return nullptr;
+    }
+    // Scan timeline to find the first queue that uses the semaphore
+    for (const auto &[_, timepoint] : timeline_) {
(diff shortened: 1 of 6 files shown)


layers/state_tracker/semaphore_state.cpp
```diff
@@ -54,6 +54,46 @@ vvl::Semaphore::Semaphore(ValidationStateTracker &dev, VkSemaphore handle, const
...
+            return &timepoint.signal_submit->queue->Handle();
+        } else {
+            for (const SubmissionReference &wait_submit : timepoint.wait_submits) {
+                if (wait_submit.queue) {
+                    return &wait_submit.queue->Handle();
+                }
+            }
+        }
+    }
+    // NOTE: In current implementation timepoints represent pending state. In-use tracking
```
artem-lunarg, line 77: Just realized something, so a big comment to make it visible during future cleanups.

arno-lunarg (approved):
🚧

PR #8894, opened 2024-11-20
layers: Remove extra space

artem-lunarg:

### DIFF
--- a/layers/error_message/error_location.cpp
+++ b/layers/error_message/error_location.cpp
@@ -52,7 +52,12 @@ std::string Location::Message() const {
     std::stringstream out;
     out << StringFunc() << "(): ";
     AppendFields(out);
-    return out.str();
+    std::string message = out.str();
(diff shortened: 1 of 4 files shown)


layers/error_message/error_location.cpp
```diff
@@ -52,7 +52,12 @@ std::string Location::Message() const {
     std::stringstream out;
     out << StringFunc() << "(): ";
     AppendFields(out);
-    return out.str();
+    std::string message = out.str();
+    // Remove space in the end when no fields are added
```
artem-lunarg, line 56: All places that I found add " " after this call.

artem-lunarg: Except `GpuShaderInstrumentor::ValidateCmdWaitEvents` which appends ": recorded with..."

spencer-lunarg (approved):
looks good, saving the world, one unused `char` at a time

PR #8895, opened 2024-11-20
tests: Fix mutable descriptor test

ziga-lunarg:
This fixes the test on some drivers which otherwise fail to allocate descriptor sets


tests/unit/descriptors.cpp
```diff
@@ -4203,7 +4203,7 @@ TEST_F(NegativeDescriptors, CopyMutableDescriptors) {
 
         VkDescriptorPoolSize pool_size = {};
         pool_size.type = VK_DESCRIPTOR_TYPE_MUTABLE_EXT;
-        pool_size.descriptorCount = 4;
+        pool_size.descriptorCount = 8;
```
spencer-lunarg, line 4206: was this failing to allocate the descriptor set? I almost feel we need a comment here, I would have guessed 4 was enough since its 2 sets with 2 descriptor in each

ziga-lunarg: Yes, I've seen this in a lot of tests on a certain driver.

spencer-lunarg (approved):

tests/unit/descriptors.cpp
spencer-lunarg: will look into this later and figure out with the WG if this should be excepted or not (and figure out if VU, spec language, best practice, etc)

spencer-lunarg: aww, it was related to the mutable list https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/7036

PR #8904, opened 2024-11-21
scripts: Split up chasssis.{h,cpp}

jeremyg-lunarg:
Move the giant blocks of 'manual' code to files under layers/chassis

DispatchObject and ValidationObject are a big mix of manual and generated code. The class definition headers #include a file defining generated methods.

chassis.h is now a manual file that only defines some shared data structures and the manual defined functions needed by the generated code.


layers/core_checks/cc_buffer.cpp
```diff
@@ -21,7 +21,6 @@
 
 #include <vulkan/vk_enum_string_helper.h>
 #include "error_message/error_location.h"
-#include "generated/chassis.h"
```
spencer-lunarg, line 24: not having to whole-include the chassis here makes me happy

spencer-lunarg (approved):
LGTM (assuming things are just copy-and-pasted where I expect it to be)

Did a double check locally and seems things are split up well, of course will want @aqnuep sign off as well as someone who will have to ingest this downstream

layers/chassis/dispatch_object_manual.cpp
```diff
@@ -17,12 +17,232 @@
...
+    ConfigAndEnvSettings config_and_env_settings_data{
+        OBJECT_LAYER_DESCRIPTION, pCreateInfo,     enabled,          disabled, debug_report,
+        &global_settings,         &gpuav_settings, &syncval_settings};
+    ProcessConfigAndEnvSettings(&config_and_env_settings_data);
+
+    if (disabled[handle_wrapping]) {
+        wrap_handles = false;
+    }
+
+    // Add VOs to dispatch vector. Order here will be the validation dispatch order!
```
jeremyg-lunarg, line 62: This part needs to go back to generated code because it is different for SC

layers/chassis/dispatch_object_manual.cpp
```diff
@@ -17,12 +17,232 @@
...
+
+void DispatchObject::DestroyDevice(VkDevice device, const VkAllocationCallbacks* pAllocator) {
+   device_dispatch_table.DestroyDevice(device, pAllocator);
+}
+
+void DispatchObject::InitDeviceValidationObjects(DispatchObject* instance_dispatch) {
+    // Note that this DEFINES THE ORDER IN WHICH THE LAYER VALIDATION OBJECTS ARE CALLED
+    disabled = instance_dispatch->disabled;
+    enabled = instance_dispatch->enabled;
+    if (!disabled[thread_safety]) {
```
jeremyg-lunarg, line 131: This part needs to go back to generated code because it is different for SC

layers/chassis/dispatch_object_manual.cpp
```diff
@@ -760,383 +958,371 @@ void *BuildUnwrappedUpdateTemplateBuffer(DispatchObject *layer_data, uint64_t de
...
 }
 
-VkResult DispatchBeginCommandBuffer(VkCommandBuffer commandBuffer, const VkCommandBufferBeginInfo *pBeginInfo, bool is_secondary) {
-    auto layer_data = GetLayerDataPtr(GetDispatchKey(commandBuffer), layer_data_map);
-    if (!wrap_handles || !is_secondary) return layer_data->device_dispatch_table.BeginCommandBuffer(commandBuffer, pBeginInfo);
+VkResult DispatchObject::BeginCommandBuffer(VkCommandBuffer commandBuffer, const VkCommandBufferBeginInfo *pBeginInfo,
+                                            bool &is_secondary) {
+    {
+        auto lock = ReadLockGuard(secondary_cb_map_mutex);
+        is_secondary = secondary_cb_map.find(commandBuffer) != secondary_cb_map.end();
```
jeremyg-lunarg, line 1267: secondary command buffer tracking was happening in chassis.cpp. It is needed here for handle wrapping weirdness.   Unfortunately, thread safety and object tracking were relying on `is_secondary` as well so I had to hack to pass it back out of this method :(

jeremyg-lunarg:
> LGTM (assuming things are just copy-and-pasted where I expect it to be)
> 
> Did a double check locally and seems things are split up well, of course will want @aqnuep sign off as well as someone who will have to ingest this downstream

The files under layers/chassis and scripts/generators need the most review IMO. But that can wait until I clean up the epic fails I'm getting in CI

layers/chassis/dispatch_object_manual.cpp
spencer-lunarg: this is because `VkCommandBufferBeginInfo::pInheritanceInfo` can be a garbage pointer, we need to ignore it (including Safe Struct) if it is not a secondary pointer

image

Going to create an issue if we can require this to be null as the app really should clear it out and not require everyone to state track to know if we can deference it or not

spencer-lunarg: https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/7037

spencer-lunarg (approved):

PR #8905, opened 2024-11-21
layers: Cleanup of Queue and ImageLayout checks

spencer-lunarg:
Before I go into this to fix things for GPU-AV, noticed some small QoL fixes wanted to apply first

### DIFF
--- a/layers/core_checks/cc_queue.cpp
+++ b/layers/core_checks/cc_queue.cpp
@@ -30,9 +30,10 @@
 #include "state_tracker/buffer_state.h"
 #include "sync/sync_vuid_maps.h"
 
+// Holds common information between all command buffers being submitted
 struct CommandBufferSubmitState {
     const CoreChecks &core;
-    const vvl::Queue *queue_state;
+    const vvl::Queue &queue_state;
     QFOTransferCBScoreboards<QFOImageTransferBarrier> qfo_image_scoreboards;
     QFOTransferCBScoreboards<QFOBufferTransferBarrier> qfo_buffer_scoreboards;
     std::vector<VkCommandBuffer> current_cmds;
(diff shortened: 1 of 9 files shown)


layers/core_checks/cc_queue.cpp
```diff
@@ -30,9 +30,10 @@
 #include "state_tracker/buffer_state.h"
 #include "sync/sync_vuid_maps.h"
 
+// Holds common information between all command buffers being submitted
 struct CommandBufferSubmitState {
     const CoreChecks &core;
-    const vvl::Queue *queue_state;
+    const vvl::Queue &queue_state;
```
spencer-lunarg, line 36: found no reason this was a pointer

layers/core_checks/cc_queue.cpp
```diff
@@ -147,40 +148,71 @@ bool CoreChecks::PreCallValidateQueueSubmit(VkQueue queue, uint32_t submitCount,
...
-                }
-                if (cb_state->resumesRenderPassInstance) {
-                    if (!suspended_render_pass_instance) {
-                        skip |=
-                            LogError("VUID-VkSubmitInfo-pCommandBuffers-06193", queue, submit_loc.dot(Field::pCommandBuffers, i),
+            ASSERT_AND_CONTINUE(cb_state);
+
+            skip |= cb_submit_state.Validate(cb_loc, *cb_state, perf_pass);
+
+            // Make sure command buffers are all protected or unprotected
```
spencer-lunarg, line 182: we were looping `commandBufferCount` twice in QueueSubmit, but in QueueSumbit2 we combine the Protected command buffer checks in the same loop, so moved that flow over here as well

layers/core_checks/cc_queue.cpp
```diff
@@ -348,40 +348,38 @@ bool CoreChecks::ValidateQueueSubmit2(VkQueue queue, uint32_t submitCount, const
                                                                 "VUID-VkCommandBufferSubmitInfo-deviceMask-03891");
             }
 
-            if (cb_state != nullptr) {
```
spencer-lunarg, line 351: we were derferencing this above, so clearly it is not null

jeremyg-lunarg (approved):

artem-lunarg (approved):

Issue #8880, opened 2024-11-19
False positive vkQueueSubmit VUID-vkQueueSubmit-pCommandBuffers-00071


fridenmf:
**Environment:**
 - OS: Windows
 - GPU and driver version: NVIDIA GeForce RTX 2080 Ti
 - SDK or header version if building from repo: 1.3.296.0
 - Options enabled (synchronization, best practices, etc.): validate_core

**Describe the Issue**

vkQueueSubmit trigger a false positive VUID-vkQueueSubmit-pCommandBuffers-00071 validation error:
```
Validation Error: [ VUID-vkQueueSubmit-pCommandBuffers-00071 ] | MessageID = 0x2e2f4d65 | vkQueueSubmit(): pSubmits[0].pCommandBuffers[0] VkCommandBuffer 0x1417c843360[CommandBuffer0] is already in use and is not marked for simultaneous use.

The Vulkan spec states: If any element of the pCommandBuffers member of any element of pSubmits was not recorded with the VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT, it must not be in the pending state (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkQueueSubmit-pCommandBuffers-00071)
```

The application create one command buffer for each swapchain image outside the main loop, responsible for all rendering that swapchain index, and re-submits a command buffer each frame, choosing which one depending on the index acquired by vkAcquireNextImageKHR.

Each vkAcquireNextImageKHR is followed by a vkWaitForFences to make sure that the presentation engine is done presenting that particular image the previous time it was used for presentation, following the principles of the [swapchain_recreation](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/samples/api/swapchain_recreation) sample in Vulkan-Samples:
```
In this sample, a workaround is implemented where a fence signaled by vkAcquireNextImageKHR is used to determine when the previous present job involving the same image index has been completed. This is often much later than the point where the present resources can be freed.
```

The bug is present in all available Vulkan SDK installers tested since at least a year back:
* VulkanSDK-1.3.296.0-Installer.exe
* VulkanSDK-1.3.290.0-Installer.exe
* VulkanSDK-1.3.283.0-Installer.exe
* VulkanSDK-1.3.280.0-Installer.exe
* VulkanSDK-1.3.275.0-Installer.exe
[...]

PR #8906, opened 2024-11-21
layers: Change ownership of AcquireFenceSync object

artem-lunarg:
`AcquireNextImage` does not necessarily use the same fence with a specific
swapchain image. The previous implementation made this (incorrect) assumption,
and this allowed to use Fence object as the owner of `AcquireFenceSync`.

A better solution is that Present call stores `AcquireFenceSync` in a
swapchain image. When swapchain image is re-acquired, it assigns
`AcquireFenceSync` to a Fence object, if any.

This also removes the need to handle a special case that swapchain can
be re-created, and in that case we had to invalidate `AcquireFenceSync`
stored inside Fence. Now Fence object is automatically associated
with correct swapchain.

Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8880

### DIFF
--- a/layers/state_tracker/fence_state.cpp
+++ b/layers/state_tracker/fence_state.cpp
@@ -173,8 +173,3 @@ void vvl::Fence::SetAcquireFenceSync(const AcquireFenceSync &acquire_fence_sync)
 
     acquire_fence_sync_ = acquire_fence_sync;
 }
-
-bool vvl::Fence::IsAcquireFenceSyncSwapchainChanged(const std::shared_ptr<vvl::Swapchain> &current_swapchain) const {
-    auto guard = ReadLock();
-    return acquire_fence_sync_.swapchain != current_swapchain;
-}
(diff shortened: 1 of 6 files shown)


layers/state_tracker/fence_state.cpp
```diff
@@ -173,8 +173,3 @@ void vvl::Fence::SetAcquireFenceSync(const AcquireFenceSync &acquire_fence_sync)
 
     acquire_fence_sync_ = acquire_fence_sync;
 }
-
-bool vvl::Fence::IsAcquireFenceSyncSwapchainChanged(const std::shared_ptr<vvl::Swapchain> &current_swapchain) const {
```
artem-lunarg, line 177: This special case has gone now

layers/state_tracker/state_tracker.cpp
```diff
@@ -3865,12 +3865,8 @@ void ValidationStateTracker::PostCallRecordQueuePresentKHR(VkQueue queue, const
         if (local_result != VK_SUCCESS && local_result != VK_SUBOPTIMAL_KHR) continue;  // this present didn't actually happen.
         // Mark the image as having been released to the WSI
         if (auto swapchain_data = Get<vvl::Swapchain>(pPresentInfo->pSwapchains[i])) {
-            if (const auto &acquire_fence = swapchain_data->images[pPresentInfo->pImageIndices[i]].acquire_fence) {
```
artem-lunarg, line 3868: Assignment of SetAcquireFenceSync is simpler too. Assuming this solution works, it handles more cases and is simpler!

PR #8915, opened 2024-11-25
layers: Rename LastBound::PER_SET

spencer-lunarg:
`LastBound::PER_SET` is not helpful name, changed to `DescriptorSetSlot`

also instead of going `last_bound.ds_slots[index].bound_descriptor_set` it is just `last_bound.ds_slots[index].ds_state` to try and reduce the `bound` being littered everything it doesn't need to be (since you are already actively grabbing `LastBound`)

### DIFF
--- a/layers/state_tracker/pipeline_state.h
+++ b/layers/state_tracker/pipeline_state.h
@@ -678,13 +678,14 @@ struct LastBound {
         uint32_t index = 0;
         VkDeviceSize offset = 0;
     };
-    // Ordered bound set tracking where index is set# that given set is bound to
-    struct PER_SET {
-        std::shared_ptr<vvl::DescriptorSet> bound_descriptor_set;
-        std::optional<DescriptorBufferBinding> bound_descriptor_buffer;
+
+    // Each command buffer has a "slot" to hold a descriptor set binding. This "slot" also might be empty
+    struct DescriptorSetSlot {
+        std::shared_ptr<vvl::DescriptorSet> ds_state;
+        std::optional<DescriptorBufferBinding> descriptor_buffer_binding;
 
         // one dynamic offset per dynamic descriptor bound to this CB
-        std::vector<uint32_t> dynamicOffsets;
+        std::vector<uint32_t> dynamic_offsets;
         PipelineLayoutCompatId compat_id_for_set{0};
 
         // Cache most recently validated descriptor state for ValidateActionState/UpdateDrawState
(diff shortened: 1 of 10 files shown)


layers/state_tracker/pipeline_state.h
```diff
@@ -678,13 +678,14 @@ struct LastBound {
         uint32_t index = 0;
         VkDeviceSize offset = 0;
     };
-    // Ordered bound set tracking where index is set# that given set is bound to
-    struct PER_SET {
-        std::shared_ptr<vvl::DescriptorSet> bound_descriptor_set;
-        std::optional<DescriptorBufferBinding> bound_descriptor_buffer;
+
+    // Each command buffer has a "slot" to hold a descriptor set binding. This "slot" also might be empty
+    struct DescriptorSetSlot {
```
artem-lunarg, line 683: Was always confused by PER_SET, nice rename

arno-lunarg (approved):
ohnepixel-csgo

layers/state_tracker/pipeline_state.h
```diff
@@ -678,13 +678,14 @@ struct LastBound {
...
+        std::shared_ptr<vvl::DescriptorSet> ds_state;
```
artem-lunarg, line 684: `ds_state` - what do you think about `descriptor_set_state`) ? The next name is not abbreviated (`descriptor_buffer_binding`), should we use this style? I presonally prefer longer name but I did not really work with this code and for could be the best for usability.

layers/state_tracker/pipeline_state.h
spencer-lunarg: there is no spec concept of a "slot" and but I feel that is the best way to describe it... but anything is better than `PER_SET` IMO

layers/state_tracker/pipeline_state.h
artem-lunarg: to be honest, in my brain had burned some patterns for DS and it's DEPTH-STENCIL! probably the reason of nitpicking

spencer-lunarg: so `descriptor_buffer_binding` is not `ds` it is for `VK_EXT_descriptor_buffer` and honestly there is already a MUCH larger issue with the fact a "descriptor buffer" is a spec term that doesn't mean quite a `VkDescriptorSet` nor `VkBuffer` but both and has already caused confusion (https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/7010)

I guess my main goal is reduce `set` and `binding` where possible as there are so many things that have them in their name where it is not the focus... the `state` is the focus, because I want to know I am grabbing the state object

artem-lunarg (approved):

layers/state_tracker/pipeline_state.h
arno-lunarg: IMO here it is ok to go for `ds_state` (makes sense from context), but still have `descriptor_buffer_binding` has long name. `ds_state` is commonly used, `descriptor_buffer_binding` is not

spencer-lunarg: lol that is fair, I guess luckily Depth Stencil is does not have a state (also your brain doesn't think DepthStencil when you are in `LastBound`... if I was in the format or image layout classes, different story lol)

PR #8917, opened 2024-11-25
Spencer lunarg more descriptors55

spencer-lunarg:
Calls `SetImageViewInitialLayout` directly instead of a callback. Pass things in as reference


jeremyg-lunarg (approved):
LGTM. MOAR CHEEZBURGUR

PR #8930, opened 2024-11-26
gpu: Handle Alias variables

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8875

when dealing with something like

```glsl
layout(set = 0, binding = 2) uniform sampler3D tex3d[];
layout(set = 0, binding = 2) uniform sampler2D tex[];
```

we need to store the `OpVariable` in Post Processing to be able to map back... once we know we have the `OpVariable` it makes the rest of the code simpler!


layers/core_checks/cc_descriptor.cpp
```diff
@@ -826,22 +826,22 @@ bool CoreChecks::ValidateDrawState(const DescriptorSet &descriptor_set, uint32_t
                                             const_cast<DescriptorSet &>(descriptor_set), set_index, framebuffer, loc);
 
     for (const auto &binding_pair : bindings) {
+        ASSERT_AND_CONTINUE(binding_pair.second.variable);
+        const auto &resource_variable = *binding_pair.second.variable;
```
spencer-lunarg, line 830: before we were passing in `DescriptorBindingInfo` but now that we can get the `OpVariable` for sure, we can remove having to carry around a map for it

layers/drawdispatch/descriptor_validator.cpp
```diff
@@ -247,72 +227,14 @@ bool DescriptorValidator::ValidateDescriptor(const DescriptorBindingInfo &bindin
     return false;
 }
 
-static const spirv::ResourceInterfaceVariable *FindMatchingImageVariable(const std::vector<DescriptorRequirement> &reqs,
```
spencer-lunarg, line 250: these were old hacks (not correct even) to try and find the `OpVariable` before we tracked it in this PR

layers/core_checks/cc_descriptor.cpp
```diff
@@ -826,22 +826,22 @@ bool CoreChecks::ValidateDrawState(const DescriptorSet &descriptor_set, uint32_t
                                             const_cast<DescriptorSet &>(descriptor_set), set_index, framebuffer, loc);
 
     for (const auto &binding_pair : bindings) {
```
arno-lunarg, line 828:
```suggestion
    for (const auto &[desc_set_binding_i, desc_set_reqs] : bindings) {
```
and bindings should be renamed, (actually its type, `BindingVariableMap`, should too) to something like `desc_set_binding_to_requirements_map`

layers/core_checks/cc_descriptor.cpp
```diff
@@ -826,22 +826,22 @@ bool CoreChecks::ValidateDrawState(const DescriptorSet &descriptor_set, uint32_t
...
+        ASSERT_AND_CONTINUE(binding_pair.second.variable);
+        const auto &resource_variable = *binding_pair.second.variable;
+
         const auto *binding = descriptor_set.GetBinding(binding_pair.first);
         if (!binding) {  //  End at construction is the condition for an invalid binding.
             auto set = descriptor_set.Handle();
             result |= LogError(vuids.descriptor_buffer_bit_set_08114, set, loc, "%s %s is invalid.", FormatHandle(set).c_str(),
-                               binding_pair.second.variable->DescribeDescriptor().c_str());
+                               resource_variable.DescribeDescriptor().c_str());
```
arno-lunarg, line 836: In this context `variable` is too vague.  Maybe try `spv_var` or `spv_variable`.

layers/drawdispatch/descriptor_validator.cpp
```diff
@@ -232,13 +218,7 @@ bool DescriptorValidator::ValidateDescriptor(const DescriptorBindingInfo &bindin
...
             return true;
         }
-        bool is_written_to = false;
-        for (const auto &req : binding_info.second) {
-            if (req.variable->IsWrittenTo()) {
-                is_written_to = true;
-                break;
-            }
-        }
+        const bool is_written_to = resource_variable.IsWrittenTo();
```
arno-lunarg, line 221: Remove this variable, `resource_variable.IsWrittenTo()` is a constant

layers/drawdispatch/descriptor_validator.cpp
```diff
@@ -321,11 +243,8 @@ bool DescriptorValidator::ValidateDescriptor(const DescriptorBindingInfo &bindin
         // https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8922
         // To not give false positve, we will skip all Sampler related checks since sampler_states will be empty
     } else {
-        for (const auto &req : binding_info.second) {
-            if (!req.variable || req.variable->samplers_used_by_image.size() <= index) {
-                continue;
-            }
-            for (const auto &desc_index : req.variable->samplers_used_by_image[index]) {
+        if (index < resource_variable.samplers_used_by_image.size()) {
```
arno-lunarg, line 246: Rename `index`, too vague given the size of the function

layers/gpu/descriptor_validation/gpuav_descriptor_validation.cpp
```diff
@@ -237,18 +238,32 @@ void UpdateBoundDescriptors(Validator &gpuav, CommandBuffer &cb_state, VkPipelin
...
+                     iter != binding_req_map->end() && iter->first == descriptor_access.binding; ++iter) {
+                    if (iter->second.variable->id == descriptor_access.variable_id) {
+                        resource_variable = iter->second.variable;
+                        break;
+                    }
+                }
+                ASSERT_AND_CONTINUE(resource_variable);
+
+                // If we already validated/updated the descriptor on the CPU, don't redo it now in GPU-AV Post Processing
+                if (!bound_descriptor_set->ValidateBindingOnGPU(*descriptor_binding, resource_variable->is_dynamic_accessed)) {
```
arno-lunarg, line 262:
```suggestion
                if (!bound_descriptor_set->ValidateBindingOnGPU(*descriptor_binding, resource_variable->is_dynamic_access)) {
```
or
```suggestion
                if (!bound_descriptor_set->ValidateBindingOnGPU(*descriptor_binding, resource_variable->is_dynamically_accessed)) {
```

layers/gpu/resources/gpuav_shader_resources.h
```diff
@@ -79,11 +79,16 @@ struct BindingLayout {
     uint32_t count;
 };
 
-// For each descriptor index we have a 32-bit slot to mark what happend on the GPU.
-// Since most devices can only support 32 descriptor sets, we can start to be clever and compress info into these 32-bits by
-// exploiting the fact certain data we are saving doesn't need a full 32-bit to save. GLSL doesn't have bitfields and don't want to
-// make this structs in case we can't fit everything in 32-bits anymore.
-typedef uint32_t PostProcessDescriptorIndexSlot;
+struct PostProcessDescriptorIndexSlot {
+    // For each descriptor index we have a 32-bit slot to mark what happend on the GPU.
```
arno-lunarg, line 83:
```suggestion
    // For each descriptor index we have a 32-bit slot to mark what happend on the GPU.
```
This part is no longer true given your changes? Should be "... we have 64 bits ..."?

arno-lunarg (approved):

layers/core_checks/cc_descriptor.cpp
spencer-lunarg: I thought alot about this, `spv_variable` doesn't tell me anything, is it for Push Constants, Descriptors, Input/Output, etc (this is why we have a `spv::VariableBase` class)

"Resource" is the spec term around descriptors and the class is `spv::ResourceInterfaceVariable` but "interface" is kinda implied as there is there is no other way to get a descriptor in your SPIR-V

layers/gpu/descriptor_validation/gpuav_descriptor_validation.cpp
spencer-lunarg: ya, I like `is_dynamically_accessed` better

layers/gpu/resources/gpuav_shader_resources.h
spencer-lunarg: the struct is 64-bit, but the first `32-bit` is still being sectioned off... will update comment a bit

layers/core_checks/cc_descriptor.cpp
spencer-lunarg: overall, the issue is everything slowly was becoming `binding` because it works for so many things in this area of code

layers/core_checks/cc_descriptor.cpp
spencer-lunarg: I guess I can't find any other code in VVL that uses "variable" in the variable name except for the SPIR-V stuff, so not sure what you will get it mixed up with

arno-lunarg: I clearly only worked a little with spirv, so you have the final call, but when just looking at this change in the git diff, it is really not obvious that `resource_variable` designates a spirv variable. `resource_spv_var` is much clearer IMO

layers/core_checks/cc_descriptor.cpp
spencer-lunarg: also this syntax is much better than `.first`/`.second`!

layers/core_checks/cc_descriptor.cpp
spencer-lunarg: > it is really not obvious that resource_variable designates a spirv variable.

my counter point is what did you potentially think it would have been otherwise? Unlike the million things that we have that are a `binding` or `index` I really don't know what the ambiguity is here with `resource_variable`

PR #8936, opened 2024-11-27
sync: Categorize error messages

artem-lunarg:
That's all syncval error messages that have different structure. More work is needed to classify this further.

The idea that this should simplify further work when it's in a single place (already this helped to detect some inconsistencies and opportunities to organize this better).

Additional progress on this might allow to filter error messages (as ANGLE does) in more reliable way. Currently ANGLE specifies multiple parts of the error message to identify specific use case. A potential improvement is to filter in a more structured way and to specify more stable values than just arbitrary text snippets.

Initial motivation behind the change is to minimize breaking changes for the clients that filter error messages based on current structure. Having this in a centralized place it should be easier to provide some backward compatibility option if necessary while this is in active development.

TESTING: For testing purposes on the local machine I forced dispatchable handles to have the same value. This allowed to ensure that outputs of all negative syncval tests before and after this change are binary identical (the only potential bug if two dispatchable handles are swapped somehow).

### DIFF
--- a/layers/sync/sync_error_messages.cpp
+++ b/layers/sync/sync_error_messages.cpp
@@ -1,0 +1,368 @@
+/* Copyright (c) 2024 The Khronos Group Inc.
+ * Copyright (c) 2024 Valve Corporation
+ * Copyright (c) 2024 LunarG, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "sync/sync_error_messages.h"
+#include "sync/sync_commandbuffer.h"
+#include "sync/sync_image.h"
+#include "state_tracker/buffer_state.h"
+#include "state_tracker/descriptor_sets.h"
+
+#include <cassert>
+#include <cinttypes>
+#include <cstdarg>
+#include <sstream>
+
+// TODO: update algorith in logging.cpp to be similar to this version.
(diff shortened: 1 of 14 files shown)


layers/sync/sync_error_messages.cpp
```diff
@@ -0,0 +1,368 @@
...
+#include "sync/sync_image.h"
+#include "state_tracker/buffer_state.h"
+#include "state_tracker/descriptor_sets.h"
+
+#include <cassert>
+#include <cinttypes>
+#include <cstdarg>
+#include <sstream>
+
+// TODO: update algorith in logging.cpp to be similar to this version.
```
artem-lunarg, line 29: Added comment. After some testing period this formatting code can be used to update existing code, there are some C++ spec problems with it (still works in practice), but also leads to some hacky manipulations that are not needed here.

layers/sync/sync_error_messages.cpp
```diff
@@ -0,0 +1,368 @@
...
+// logging.cpp's vsnprintf usage in some places assumes that std::string
+// storage reserves space for null terminator but it is not guaranteed
+// (although it is in practise). vsnprintf there writes 0 terminator
+// in the char indexed as str[std::string::size()] and it should not be
+// accesses in this way by client code, only by std::string itself.
+
+static std::string Format(const char* format, ...) {
```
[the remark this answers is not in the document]
spencer-lunarg: makes sense... good to add as a comment here

layers/sync/sync_renderpass.cpp
```diff
@@ -302,11 +294,9 @@ bool RenderPassAccessContext::ValidateStoreOperation(const CommandBufferAccessCo
                 const char *const op_type_string = checked_stencil ? "stencilStoreOp" : "storeOp";
                 const char *const store_op_string = string_VkAttachmentStoreOp(checked_stencil ? ci.stencilStoreOp : ci.storeOp);
```
spencer-lunarg, line 295: (can be done in a followup PR) but we could pass these in as `vvl::Field::stencilStoreOp` and then use the `String` method to print it later

spencer-lunarg (approved):
With the number of various errors we have in sync val, the idea of categorizing them is a good idea!

Not sure if this will end up being the best way or not to do it, but it seems like a good, logical next step

Issue #6944, opened 2023-11-11
Detect if descriptor is statically indexed


spencer-lunarg:
from https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6883 we only detect if the descriptor is an array or not.

This leaves gaps in validation when we know the size of the array and have constants access chains into it.

It will need to be investigated if these edge cases are better to leave only for GPU-AV or if it is easily detectable

spencer-lunarg:
discussed with Jeremy, decided we will have to have some level of "we tried" and then quickly switch over to needing GPU-AV support

spencer-lunarg:
update - there are 2 disabled tests (`NegativeDescriptors.ConstantArrayElementNotBound` and `NegativeShaderImageAccess.SampledImageShareBindingArray`) that tests this. The question is if `is_dynamically_accessed` is currently too strict or not (https://gitlab.khronos.org/vulkan/vulkan/-/issues/4098)

PR #8937, opened 2024-11-28
gpu: Fix static array checking

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/6944

Now that GPU-AV is in a better shape, was able to actually try and re-work this logic. Tests added and re-enabled to confirm things work as expected (still some open edge cases, but will be worked out after getting some spec clarification)


layers/gpu/spirv/descriptor_indexing_oob_pass.cpp
```diff
@@ -109,68 +108,99 @@ void DescriptorIndexingOOBPass::Reset() {
...
+        }
+        var_inst_ = &variable->inst_;
+
+        const Type* pointer_type = variable->PointerType(module_.type_manager_);
+        if (!pointer_type) {
+            module_.InternalError(Name(), "Pointer type not found");
+            return false;
+        }
+
+        if (pointer_type->inst_.IsArray() && access_chain_inst->Length() >= 5) {
```
arno-lunarg, line 139: Some comment on this 5 maybe?

layers/gpu/spirv/descriptor_indexing_oob_pass.cpp
```diff
@@ -109,68 +108,99 @@ void DescriptorIndexingOOBPass::Reset() {
...
         }
 
-        const Variable* variable = module_.type_manager_.FindVariableById(access_chain_inst_->Operand(0));
+        const Variable* variable = module_.type_manager_.FindVariableById(access_chain_inst->Operand(0));
         if (!variable) {
             return false;
         }
         var_inst_ = &variable->inst_;
 
         uint32_t storage_class = variable->StorageClass();
```
arno-lunarg, line 161:
```suggestion
        const uint32_t storage_class = variable->StorageClass();
```

arno-lunarg (approved):

Issue #8900, opened 2024-11-21
Alleged `VUID-vkQueuePresentKHR-pWaitSemaphores-03268` false-positive


MarijnS95:
Separating this out from https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8590#issuecomment-2486378161.  Downloading a new validation layer from CI, ever since https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8739 was merged, seems to cause this false-positive in our application:

> Validation Error: [ VUID-vkQueuePresentKHR-pWaitSemaphores-03268 ] Object 0: handle = 0xb12fb2000000002c, name = Present ready 0, type = VK_OBJECT_TYPE_SEMAPHORE; Object 1: handle = 0xb9181f0000000029, type = VK_OBJECT_TYPE_SEMAPHORE; Object 2: handle = 0x125ee2133d8, type = VK_OBJECT_TYPE_QUEUE; | MessageID = 0x251f8f7a | vkQueuePresentKHR(): pPresentInfo->pWaitSemaphores[0] queue (VkQueue 0x125ee2133d8[]) is waiting on binary semaphore (VkSemaphore 0xb12fb2000000002c[Present ready 0]) that has an associated signal but it depends on timeline semaphore wait (VkSemaphore 0xb9181f0000000029[], wait value = 1) that does not have resolving signal submitted yet.
The Vulkan spec states: All elements of the pWaitSemaphores member of pPresentInfo must reference a semaphore signal operation that has been submitted for execution and any semaphore signal operations on which it depends must have also been submitted for execution (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkQueuePresentKHR-pWaitSemaphores-03268)

I am quite sure that we signal the timeline semaphore to the desired wait value _in a prior submit_ as outlined in https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8590#issuecomment-2486571063, and @artem-lunarg acknowledged that this case should work.

In order to have a useful reproducer that I could contribute back to this repository, I set out to write some positive tests for this feature that reproduce my case, and the tests seem to succeed meaning that this VUID _does not_ trigger: https://github.com/MarijnS95/Vulkan-ValidationLayers/compare/test. Let me know if you think this is useful, and I'll contribute the tests back.

[...]

PR #8939, opened 2024-11-28
layers: Fix detection of dependent timeline signal during binary wait

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8900

### DIFF
--- a/layers/state_tracker/queue_state.cpp
+++ b/layers/state_tracker/queue_state.cpp
@@ -141,19 +141,38 @@ std::optional<vvl::SemaphoreInfo> vvl::Queue::FindTimelineWaitWithoutResolvingSi
     // binary semaphores are used this will return immediately.
     uint32_t processed_waits = 0;
 
-    auto guard = Lock();
-    for (auto it = submissions_.rbegin(); it != submissions_.rend() && processed_waits < timeline_wait_count_; ++it) {
-        const vvl::QueueSubmission &submission = *it;
-        if (submission.seq > until_seq) {
-            continue;
-        }
-        for (const auto &wait_info : submission.wait_semaphores) {
-            if (wait_info.semaphore->type == VK_SEMAPHORE_TYPE_TIMELINE) {
-                if (!wait_info.semaphore->HasResolvingTimelineSignal(wait_info.payload)) {
-                    return wait_info;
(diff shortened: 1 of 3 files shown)


artem-lunarg:
Looking into thread-sanitizer lock-inversion error. This PR does not introduce new synchronization but test activated new code path.

spencer-lunarg (approved):

artem-lunarg:
Added separate commit to fix lock-inversion found by thread sanitizer

layers/state_tracker/queue_state.cpp
```diff
@@ -141,21 +141,40 @@ std::optional<vvl::SemaphoreInfo> vvl::Queue::FindTimelineWaitWithoutResolvingSi
...
-    for (auto it = submissions_.rbegin(); it != submissions_.rend() && processed_waits < timeline_wait_count_; ++it) {
-        const vvl::QueueSubmission &submission = *it;
-        if (submission.seq > until_seq) {
-            continue;
-        }
-        for (const auto &wait_info : submission.wait_semaphores) {
-            if (wait_info.semaphore->type == VK_SEMAPHORE_TYPE_TIMELINE) {
-                if (!wait_info.semaphore->HasResolvingTimelineSignal(wait_info.payload)) {
-                    return wait_info;
+    // Run algorithm in two separate steps to avoid lock-inversion with Semaphore::RetireWait:
```
artem-lunarg, line 144: That's the fix for thread sanitizer. From what I checked it's indeed a potential deadlock. This function was added not long ago, but thread sanitizer did not detect any issues previously. It looks like we did not have a test to activate a necessary code path, and the test from this PR did the trick.

layers/state_tracker/semaphore_state.cpp
```diff
@@ -288,6 +288,12 @@ void vvl::Semaphore::GetLastBinarySignalSource(VkQueue &queue, vvl::Func &acquir
 bool vvl::Semaphore::HasResolvingTimelineSignal(uint64_t wait_payload) const {
     assert(type == VK_SEMAPHORE_TYPE_TIMELINE);
     auto guard = ReadLock();
+
+    // Check if completed payload value (which includes initial value) resolves the wait.
+    if (wait_payload <= completed_.payload) {
```
artem-lunarg, line 293: That's the main fix for false positive.

Issue #8934, opened 2024-11-27
Increased CPU consumption over time


fridenmf:
**Environment:**
 - OS: Windows 11
 - GPU and driver version: NVIDIA GeForce RTX 2080 Ti (560.94)
 - SDK or header version if building from repo: 1.3.296.0
 - Options enabled (synchronization, best practices, etc.): synchronization, best practices

**Describe the Issue**

The validation layers consume more CPU time the longer an application runs if an application often create and destroys resources. The first seconds the validation layers take up less than 1 ms CPU time, but after about 40 seconds it takes about 8 ms, and it continues to increase the longer you run. An application starting on 200 FPS gets down to about 40 FPS after about 2 minutes.

There is a video under **Additional context** demonstrating this, as well as Visual Studio performance profile results showing the amount of time spent in the validation layers compared to a "dummy 4 ms function", where the dummy function takes up more time than the validation layers in the first seconds of the application, but only a fraction of the time of the validation layers 40 seconds later.

**Expected behavior**

The validation performance should not get worse over time if the total number of resources used in a frame does not increase over time.

**Additional context**

<details>

Video:\
![vvl_performance_loss](https://github.com/user-attachments/assets/4eb41ade-90eb-4b40-a432-8abe5b54c217)

Visual Studio performance profile after 8 seconds:\
![after_8_seconds](https://github.com/user-attachments/assets/b9257718-e859-4735-8575-65668a76661e)

Visual Studio performance profile after 40 seconds:\
![after_40_seconds](https://github.com/user-attachments/assets/5fafb363-d16c-49c0-8d83-d06ca490224f)

</details>

spencer-lunarg:
Thanks to bringing this to our attentions, we will take a look into this locally and see if we can reproduce and then hopefully track down what is going on

arno-lunarg:
Would you be able to provide an executable so we can have a proper look at how VVL behaves with your application please?

artem-lunarg:
@fridenmf Is synchronization validation enabled? If yes, does it behave the same if only standard validation is enabled?

fridenmf:
@arno-lunarg Sure! The following app has the ability to load the validation layers from the application if the "validate"-checkbox is ticked. Leave it unticked if you reproduce it vkconfig.exe instead\
[TODO.zip](https://github.com/user-attachments/files/17940333/TODO.zip)

[...]

PR #8948, opened 2024-12-02
sync: Remove access entries of destroyed resources

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8934

Syncval cleanups old access entries during host side synchronization
operations. For example, when frame synchronization uses `vkWaitForFences`
to wait for the QueueSubmit's fence, this guarantees that device memory
accesses are completed and it is safe to remove corresponding tracking
state.

There are synchronization mechanisms that are not associated directly
with a queue (acquire fence sync) and do not guarantee completion
of memory accesses (provide only execution dependency). In this case
we need to cleanup access entries when resource is destroyed, because
there is no guarantee they will be removed later.

### DIFF
--- a/tests/framework/binding.h
+++ b/tests/framework/binding.h
@@ -613,21 +613,8 @@ class Buffer : public internal::NonDispHandle<VkBuffer> {
         InitHostVisibleWithData(dev, usage, data, data_size, queue_families);
     }
 
-    Buffer(Buffer &&rhs) noexcept : NonDispHandle(std::move(rhs)) {
-        create_info_ = std::move(rhs.create_info_);
-        internal_mem_ = std::move(rhs.internal_mem_);
-    }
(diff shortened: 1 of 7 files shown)


tests/framework/binding.h
```diff
@@ -746,6 +746,40 @@ class Image : public internal::NonDispHandle<VkImage> {
     explicit Image(const Device &dev, const VkImageCreateInfo &info, NoMemT);
     explicit Image(const Device &dev, const VkImageCreateInfo &info, SetLayoutT);
 
+    Image(Image &&rhs) noexcept : NonDispHandle(std::move(rhs)) {
```
spencer-lunarg, line 749: nit would be to move the logic to `binding.cpp`

spencer-lunarg (approved):

PR #8949, opened 2024-12-02
chassis: Simplify Dispatch and Validation object tracking

jeremyg-lunarg:
Split Device and Instance state into separate maps. Make map
declarations static instead of global. Use concurrent_unordered_map
with a TLS cache instead of non-threadsafe small_map.

Store DispatchObjects and ValidationObjects in std::unique_ptr<>s to simplify memory management.


scripts/generators/dispatch_object_generator.py
```diff
@@ -171,6 +171,8 @@ def __init__(self):
         # Dispatch functions that need special state tracking variables passed in
         self.custom_definition = {}
 
+        self.dispatchable_handles = ('VkInstance', 'VkPhysicalDevice', 'VkDevice', 'VkQueue', 'VkCommandBuffer')
```
spencer-lunarg, line 174: you can loop `self.vk.handles` to check for the `dispatchable` field (that way if someone does add a new dispatch handle they would see it)

(probably can do this in a single python list comprehension)

layers/containers/custom_containers.h
```diff
@@ -768,7 +768,7 @@ class value_type_helper_map {
         // stand-in for placement new and a small-block allocator, so the const_cast
         // is minimal, contained, valid, and allows operators * and -> to avoid copies
         const_cast<Key &>(lhs.first) = rhs.first;
-        lhs.second = rhs.second;
+        lhs.second = std::move(rhs.second);
```
jeremyg-lunarg, line 771: this isn't needed and doesn't work

layers/chassis/dispatch_object_manual.cpp
```diff
@@ -53,7 +120,7 @@ DispatchObject::DispatchObject(const VkInstanceCreateInfo *pCreateInfo) : is_ins
     // create all enabled validation, which is API specific
     InitInstanceValidationObjects();
 
-    for (auto* vo : object_dispatch) {
+    for (auto &vo : object_dispatch) {
         vo->dispatch_ = this;
```
spencer-lunarg, line 124: if `vo` is a reference now, why we still using `->` here?

spencer-lunarg (approved):

layers/chassis/dispatch_object_manual.cpp
jeremyg-lunarg: it is a reference to a unique_ptr<>

jeremyg-lunarg: I updated the commit message to reflect what is happening but forgot to update the PR description until just now.

PR #8953, opened 2024-12-03
GPU-AV: Rework validation commands. Add some Index and vertex buffer validation.

### LINKED ISSUE #2492: GPU-AV check index buffer out-of-bounds

arno-lunarg:
Still a bunch of `#ARNO_TODO` to address, I know for a fact I have one buffer alignment issue on my nvidia machine to fix, but tests are green on my AMD one. So I think it should be reviewable. The diff with `gpuav_draw` will be stupid, I touched everything, I advise looking at the file directly!

Fixes #2492

Completely rework gpuav_draw, to make it easier to add validation, and debug existing one.
When adding validation for some VkBuffer, the 2 more importants parts are 1) have an informative error message and 2) creating the validation shader. gpuav_draw.cpp has been redesigned around that idea. Each shader is in charge of its small validation, no uber shader trying to cater to all needs. Doing that ends up being a mess, with the need to correctly pipe data to the validation shader.
Validation shaders are now more fine grained.
The C++ has been reworked around some template to streamline the process of adding validation. The general idea is the following: 1) Declase a new "shader" struct representing validation shader bindings and push contants
2) fill that struct
3) bind this struct and the validation pipeline
4) dispatch the appropriate number of draws

About 4): we used to dispatch one draw, and do for loop in the shader to scan buffer elements. GPU are not good at that, instead a proper amount of draws should be dispatched, each in charge of doing it's small validation logic.

Contrary to the previous attempt (https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8489), validation commands insertion is now done _after_ a render pass is ended, and only using compute pipelines. It avoids the need to restore graphics pipeline state. I use lambdas for that: when a command is intercepted, I capture the relevant validation info in the lambda, and based on that the appropriate validation commands will be inserted when a render pass is ended. 

Tested again CS2 (Train map), Doom Eternal, and Tiny Mesh Large (lightweight VK sample). CS2 runs at ~4 FPS (~100 FPS without VVL), Doom (with ray tracing) at ~6 FPS ( ~28 FPS without VVL), and lightweight VK performance is the same with or without this validation.
No error detected for lightweight VK, only one for Doom, and a bunch for CS2 (I have to investigate to see if these are not false positives)

Here is an error message generated by Index buffer validation:
```
Validation Error: [ VUID-vkCmdDrawIndexed-None-02721 ] Object 0: handle = 0x15a1abb9300, type = VK_OBJECT_TYPE_QUEUE; Object 1: handle = 0x15a2a98a110, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x24afafc5 | vkCmdDrawIndexed(): Vertex index 1 is not within the smallest bound vertex buffer.
index_buffer[ 1 ] (1) + vertexOffset (0) = Vertex index 1
Smallest vertex buffer binding info, causing OOB access:
- Buffer: VkBuffer 0xea7170000000031[]
- Binding: 0
- Binding size (effective): 32 bytes
- Binding offset: 64 bytes
- Binding stride: 32 bytes
- Vertices count: 1
At least the following vertex attribute caused OOB access:
- location: 2
- binding: 0
- format: VK_FORMAT_R32G32B32_SFLOAT
- offset: 20 bytes
Index buffer binding info:
- Buffer: VkBuffer 0xa808d50000000033[]
- Index type: VK_INDEX_TYPE_UINT16
- Binding offset: 0 bytes
- Binding size: 6 bytes (or 3 VK_INDEX_TYPE_UINT16)
Note: Vertex buffer binding size is the effective, valid one, based on how the VkBuffer was created and vertex buffer binding parameters. So it can be clamped up to 0 if binding was invalid.
The Vulkan spec states: If the robustBufferAccess feature is not enabled, and that pipeline was created without enabling VK_PIPELINE_ROBUSTNESS_BUFFER_BEHAVIOR_ROBUST_BUFFER_ACCESS for vertexInputs, then for a given vertex buffer binding, any attribute data fetched must be entirely contained within the corresponding vertex buffer binding, as described in Vertex Input Description (https://www.khronos.org/registry/vulkan/specs/1.4-extensions/html/vkspec.html#VUID-vkCmdDrawIndexed-None-02721)
```

Coming up next:
- Debug label regions in error messages
- Perf improvements targeted at CS2: Using Tracy I noticed that in the draw indexed indirect case, my validation dispatches are too big: I use the size of the index buffer as dispatch size, since the used index count is stored on the GPU. It ends up being really really big. So I need to switch to an indirect dispatch for my validation command, with a group count obtained from reading the VkBuffer stored `indicesCount`, to just validate `indicesCount` indices.
Corresponding code: https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/7eedb32e91d69e64ab30830d8ea6a50539fc8471/layers/gpu/cmd_validation/gpuav_draw.cpp#L1380


layers/CMakeLists.txt
```diff
@@ -222,8 +222,18 @@ target_sources(vvl PRIVATE
     ${API_TYPE}/generated/cmd_validation_copy_buffer_to_image_comp.cpp
     ${API_TYPE}/generated/cmd_validation_dispatch_comp.h
     ${API_TYPE}/generated/cmd_validation_dispatch_comp.cpp
-    ${API_TYPE}/generated/cmd_validation_draw_vert.h
-    ${API_TYPE}/generated/cmd_validation_draw_vert.cpp
+	${API_TYPE}/generated/cmd_validation_count_buffer_comp.h
```
spencer-lunarg, line 225:
```suggestion
    ${API_TYPE}/generated/cmd_validation_count_buffer_comp.h
```

layers/gpu/error_message/gpuav_vuids.cpp
```diff
@@ -49,6 +49,7 @@ struct GpuVuidsCmdDrawIndexed : GpuVuid {
         storage_access_oob_08613 = "VUID-vkCmdDrawIndexed-None-08613";
         invalid_descriptor_08114 = "VUID-vkCmdDrawIndexed-None-08114";
         descriptor_index_oob_10068 = "VUID-vkCmdDrawIndexed-None-10068";
+		vertex_index_oob = "VUID-vkCmdDrawIndexed-None-02721";
```
spencer-lunarg, line 52:
```suggestion
        vertex_index_oob = "VUID-vkCmdDrawIndexed-None-02721";
```

layers/gpu/core/gpuav_setup.cpp
```diff
@@ -373,6 +373,26 @@ void Validator::PreCallRecordCreateDevice(VkPhysicalDevice physicalDevice, const
...
+                       vku::FindStructInPNextChain<VkPhysicalDeviceFeatures2>(modified_create_info->pNext))) {
+            features->features.shaderInt64 = VK_TRUE;
+        } else {
+            // Need to add a VkPhysicalDeviceFeatures pointer
+            enabled_features = new VkPhysicalDeviceFeatures;
+            enabled_features->shaderInt64 = VK_TRUE;
+            modified_create_info->pEnabledFeatures = enabled_features;
+        }
+    } else {
+        InternalWarning(device, record_obj.location, "shaderInt64 feature not available, countBuffer (as seen in commands like vkCmdDrawIndexedIndirectCount) validation will not be performed.");
```
spencer-lunarg, line 394: `Validator::PreCallRecordCreateDevice` should be about trying to enable things

`Validator::InitSettings` is where you give warnings likes this

layers/gpu/resources/gpuav_state_trackers.cpp
```diff
@@ -430,6 +430,124 @@ void CommandBuffer::UpdateCommandCount(VkPipelineBindPoint bind_point) {
     }
 }
 
+void CommandBuffer::RestoreDynamicStates(const Validator &gpuav_) {
```
spencer-lunarg, line 433: I'm still worried about this, there is no way we have correctly covered all 73 dynamic states here

layers/utils/vk_layer_utils.h
```diff
@@ -282,6 +282,21 @@ static inline uint32_t GetIndexAlignment(VkIndexType indexType) {
...
+    switch (indexType) {
+        case VK_INDEX_TYPE_UINT16:
+            return 16;
+        case VK_INDEX_TYPE_UINT32:
+            return 32;
+        case VK_INDEX_TYPE_NONE_KHR:
+            return 0;
+        case VK_INDEX_TYPE_UINT8_KHR:
+            return 8;
+        default:
```
spencer-lunarg, line 295: shouldn't need `default` and will help throw compiler error in future if a new one is added

layers/gpu/resources/gpuav_state_trackers.cpp
arno-lunarg: Ah we don't need that anymore

layers/utils/vk_layer_utils.h
arno-lunarg: Yes good point!

layers/gpu/cmd_validation/gpuav_draw.cpp
```diff
@@ -74,394 +108,1125 @@ struct SharedDrawValidationResources final {
...
+            const VkDeviceSize attribute_size = vkuFormatElementSize(attrib.desc.format);
+
+            const VkDeviceSize stride =
+                vbb.stride != 0 ? vbb.stride : attribute_size;  // Tracked stride should already handle all possible value origin
+
+            if (vbb.size < (vbb.offset + attrib.desc.offset)) {
+                // overflow will occur
+                continue;
+            }
+            VkDeviceSize vertex_buffer_remaining_size = vbb.size - vbb.offset - attrib.desc.offset;
```
arno-lunarg, line 827: This vbb.size - vbb.offset looks completely wrong ...

layers/gpu/core/gpuav_setup.cpp
```diff
@@ -563,6 +581,11 @@ void Validator::InitSettings(const Location &loc) {
                 "Device does not support the minimum range of push constants (32 bytes). No indirect buffer checking will be "
                 "attempted");
         }
+        if (!supported_features.shaderInt64) {
+            InternalWarning(device, loc,
+                            "shaderInt64 feature not available, countBuffer (as seen in commands like "
+                            "vkCmdDrawIndexedIndirectCount) validation will not be performed.");
```
spencer-lunarg, line 587: I think you forgot call `gpuav_settings.SetBufferValidationEnabled(false);` in this  `if (!supported_features.shaderInt64) {`

also it would be nice if the above message were worded the same since they both do the same things

also it should be a `if () { } else (!supported_features.shaderInt64) { }` otherwise you will get told twice it is disabled for no reason

layers/gpu/resources/gpuav_vulkan_objects.h
```diff
@@ -69,7 +69,9 @@ class Buffer {
     void Destroy();
 
     bool IsDestroyed() const { return buffer == VK_NULL_HANDLE; }
+    VkBuffer &VkHandle() { return buffer; }
```
spencer-lunarg, line 72: why do we need to expose a non-const version here?

layers/gpu/shaders/cmd_validation/draw_indexed_indirect_index_buffer.comp
```diff
@@ -0,0 +1,72 @@
...
+void main() {
+    uint draw_count = 0;
+    if ((pc.flags & kIndexedIndirectDrawFlags_DrawCountFromBuffer) == 0) {
+        draw_count = pc.cpu_draw_count;
+    } else {
+        draw_count = count_buffer[pc.count_buffer_dwords_offset];
+    }
+
+    for (uint draw_i = 0; draw_i < draw_count; ++draw_i) {
+        const uint draw_i_index_count = draw_indexed_indirect_cmds[draw_i * pc.draw_cmds_stride_dwords + kIndexCount + pc.draw_indexed_indirect_cmds_dwords_offset];
```
spencer-lunarg, line 65: I would make a 

`const uint some_name = (draw_i * pc.draw_cmds_stride_dwords) + pc.draw_indexed_indirect_cmds_dwords_offset;`

intermittent value

layers/gpu/shaders/cmd_validation/draw_indexed_indirect_vertex_buffer.comp
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    
+    for (uint draw_i = 0; draw_i < draw_count; ++draw_i) {
+        const uint draw_i_index_count = draw_indexed_indirect_cmds[draw_i * pc.draw_cmds_stride_dwords + kIndexCount + pc.draw_indexed_indirect_cmds_dwords_offset];
+        const uint draw_i_first_index = draw_indexed_indirect_cmds[draw_i * pc.draw_cmds_stride_dwords + kFirstIndex + pc.draw_indexed_indirect_cmds_dwords_offset];
+        
+        // Is index within the range of indices to check?
+        if (index_buffer_offset >= draw_i_first_index && index_buffer_offset < (draw_i_first_index + draw_i_index_count)) {
+            const int vertex_offset = int(draw_indexed_indirect_cmds[draw_i * pc.draw_cmds_stride_dwords + kVertexOffset + pc.draw_indexed_indirect_cmds_dwords_offset]);
```
spencer-lunarg, line 112:
```suggestion
            const int vertex_offset = int(draw_indexed_indirect_cmds[draw_i * pc.draw_cmds_stride_dwords + kVertexOffset + pc.draw_indexed_indirect_cmds_dwords_offset]);
```

layers/state_tracker/cmd_buffer_state.h
```diff
@@ -222,6 +222,10 @@ class CommandBuffer : public RefcountedStateObject {
         bool depth_test_enable;
         // VK_DYNAMIC_STATE_DEPTH_BOUNDS_TEST_ENABLE
         bool depth_bounds_test_enable;
+        // VK_DYNAMIC_STATE_DEPTH_COMPARE_OP
```
spencer-lunarg, line 225: now that we are not attempting to restore dynamic state, did we need these state tracker changes?

tests/unit/gpu_av_index_buffer_positive.cpp
```diff
@@ -0,0 +1,178 @@
...
+    RETURN_IF_SKIP(InitGpuAvFramework());
+
+    RETURN_IF_SKIP(InitState(nullptr));
+    InitRenderTarget();
+
+    VkPipelineLayoutCreateInfo pipelineLayoutCreateInfo = vku::InitStructHelper();
+    vkt::PipelineLayout pipeline_layout(*m_device, pipelineLayoutCreateInfo);
+
+    CreatePipelineHelper pipe(*this);
+    pipe.gp_ci_.layout = pipeline_layout.handle();
```
spencer-lunarg, line 32: nit - you shouldn't need this line, we create a dummy PipelineLayout like you have above if none is provided

tests/unit/gpu_av_index_buffer_positive.cpp
```diff
@@ -0,0 +1,178 @@
...
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+#include "../framework/layer_validation_tests.h"
+#include "../framework/pipeline_helper.h"
+#include "../framework/buffer_helper.h"
+
+class PositiveGpuAVIndexBuffer : public GpuAVTest {};
```
spencer-lunarg, line 18: 100% should have a Mesh and RTX positive tests that just use them and make sure with GPU-AV the indirect buffer doesn't blowup

tests/unit/gpu_av_indirect_buffer.cpp
```diff
@@ -421,60 +440,84 @@ TEST_F(NegativeGpuAVIndirectBuffer, Mesh) {
...
+            m_errorMonitor->VerifyFound();
+
+            draw_ptr[2] = 1;
+            draw_ptr[1] = 1;
+            draw_ptr[0] = 1;
+        }
+    }
+#endif
+    memset(draw_ptr, 0, buffer_size);
+// TODO Fix task shader part of the test - it stalls GPU
```
spencer-lunarg, line 519: honestly just make a `NegativeGpuAVIndirectBuffer.DISABLED_MeshTask` test or something and move that there

layers/gpu/core/gpuav_record.cpp
```diff
@@ -763,8 +864,28 @@ void Validator::PreCallRecordCmdDrawMeshTasksIndirectCountEXT(VkCommandBuffer co
...
     }
-    InsertIndirectDrawValidation(*this, record_obj.location, *cb_state, buffer, offset, maxDrawCount, countBuffer,
-                                 countBufferOffset, stride);
+    auto indirect_buffer_state = Get<vvl::Buffer>(buffer);
+    if (!indirect_buffer_state) {
+        InternalError(commandBuffer, record_obj.location, "buffer must be a valid VkBuffer handle");
+        return;
+    }
+
+    valcmd::DrawMeshIndirectVuids vuids;
```
spencer-lunarg, line 873: if these are the same VUs, why do we both sending them into `DrawMeshIndirect` the same both places instead of just having inside `DrawMeshIndirect` itself?

layers/gpu/cmd_validation/gpuav_cmd_validation_common.cpp
```diff
@@ -78,11 +79,29 @@ void RestorablePipelineState::Create(vvl::CommandBuffer &cb_state, VkPipelineBin
     if (last_bound.push_descriptor_set) {
         push_descriptor_set_writes_ = last_bound.push_descriptor_set->GetWrites();
     }
+
+    // Do not handle cb_state.activeRenderPass->use_dynamic_rendering_inherited for now
```
spencer-lunarg, line 83: I would log a warning if `use_dynamic_rendering_inherited` is true

layers/gpu/cmd_validation/gpuav_draw.cpp
```diff
@@ -15,57 +15,91 @@
...
-        : device(gpuav.device) {
-        VkResult result = VK_SUCCESS;
+    SharedDrawValidationResources(Validator &gpuav, const Location &loc)
+        : vma_allocator(gpuav.vma_allocator_), dummy_buffer(gpuav) {
+        VkBufferCreateInfo dummy_buffer_info = vku::InitStructHelper();
+        dummy_buffer_info.size = 64;// whatever
+        dummy_buffer_info.usage = VK_BUFFER_USAGE_STORAGE_BUFFER_BIT;
+        VmaAllocationCreateInfo alloc_info = {};
+        dummy_buffer_info.size = dummy_buffer_info.size;
+        VkResult result = vmaCreateBuffer(gpuav.vma_allocator_, &dummy_buffer_info, &alloc_info, &dummy_buffer.VkHandle(),
```
spencer-lunarg, line 58: the big point of `vko::Buffer` was it could call `vma` calls for use (and wrap the error messages)

You can just pass in `SharedDrawValidationResources(Validator &gpuav)` if you need to as well

layers/gpu/cmd_validation/gpuav_draw.cpp
```diff
@@ -74,394 +108,1125 @@ struct SharedDrawValidationResources final {
...
-    const bool use_shader_objects = pipeline_state == nullptr;
-
-    auto &shared_draw_resources = gpuav.shared_resources_manager.Get<SharedDrawValidationResources>(
-        gpuav, cb_state.GetErrorLoggingDescSetLayout(), use_shader_objects, loc);
+    if (gpuav.enabled_features.drawIndirectFirstInstance) return;
 
-    assert(shared_draw_resources.IsValid());
-    if (!shared_draw_resources.IsValid()) {
+    if (((draw_buffer_offset % sizeof(uint32_t)) != 0) || ((count_buffer_offset % sizeof(uint32_t)) != 0)) {
         return;
```
spencer-lunarg, line 244: this feel worth logging a warning as this is invalid to my knowledge things like `VUID-vkCmdDrawIndexedIndirect-offset-02710` make this invalid

layers/gpu/core/gpuav_setup.cpp
arno-lunarg: I do it on purpose, only count buffer validation is not possible without shaderInt64

layers/gpu/cmd_validation/gpuav_draw.cpp
arno-lunarg: Ah yes oversight. That's also why I added non const versions of `vko::Buffer::VkHandle`

layers/gpu/cmd_validation/gpuav_cmd_validation_common.cpp
arno-lunarg: Ah again leftover from previous attempt, I can just remove that

layers/gpu/cmd_validation/gpuav_draw.cpp
arno-lunarg: Now that you make me look at this again, I think I should just remove that. No point in doing core validation's job here

layers/state_tracker/cmd_buffer_state.h
arno-lunarg: leftover

tests/unit/gpu_av_indirect_buffer.cpp
arno-lunarg: done

tests/unit/gpu_av_index_buffer_positive.cpp
arno-lunarg: Done, with `gpu_av_indirect_buffer_positive`

layers/gpu/resources/gpuav_vulkan_objects.h
```diff
@@ -64,8 +64,8 @@ class Buffer {
     void FlushAllocation(const Location &loc, VkDeviceSize offset = 0, VkDeviceSize size = VK_WHOLE_SIZE) const;
     void InvalidateAllocation(const Location &loc, VkDeviceSize offset = 0, VkDeviceSize size = VK_WHOLE_SIZE) const;
 
-    void Create(const Location &loc, const VkBufferCreateInfo *buffer_create_info,
-                const VmaAllocationCreateInfo *allocation_create_info);
+    [[nodiscard]] bool Create(const Location &loc, const VkBufferCreateInfo *buffer_create_info,
```
spencer-lunarg, line 67: just curious if this is something you actually hit, or just something you want to prevent in the case it did happen

tests/unit/gpu_av_indirect_buffer_positive.cpp
```diff
@@ -0,0 +1,189 @@
...
+ */
+
+#include "../framework/layer_validation_tests.h"
+#include "../framework/pipeline_helper.h"
+#include "../framework/buffer_helper.h"
+#include "../framework/ray_tracing_objects.h"
+
+class PositiveGpuAVIndirectBuffer : public GpuAVTest {
+  public:
+    // Turned off in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8586
```
spencer-lunarg, line 21: this PR is turning things back on, so we shouldn't need this hack anymore

tests/CMakeLists.txt
```diff
@@ -117,6 +117,8 @@ target_sources(vk_layer_validation_tests PRIVATE
     unit/gpu_av_descriptor_post_process_positive.cpp
     unit/gpu_av_image_layout.cpp
     unit/gpu_av_indirect_buffer.cpp
+    unit/gpu_av_indirect_buffer_positive.cpp
+    unit/gpu_av_index_buffer_positive.cpp
```
spencer-lunarg, line 121: assuming we also wanted a `gpu_av_index_buffer.cpp` tests for the negative tests

layers/state_tracker/state_tracker.cpp
```diff
@@ -5187,8 +5184,12 @@ void ValidationStateTracker::PostCallRecordCmdBindVertexBuffers2(VkCommandBuffer
         auto buffer_state = Get<vvl::Buffer>(pBuffers[i]);
         vvl::VertexBufferBinding &vertex_buffer_binding = cb_state->current_vertex_buffer_binding_info[i + firstBinding];
         vertex_buffer_binding.buffer = pBuffers[i];
-        vertex_buffer_binding.size = (pSizes) ? pSizes[i] : VK_WHOLE_SIZE;
         vertex_buffer_binding.offset = pOffsets[i];
+        vertex_buffer_binding.size = pSizes ? pSizes[i] : VK_WHOLE_SIZE;
+        if (vertex_buffer_binding.size == VK_WHOLE_SIZE) {
+            vertex_buffer_binding.size = vvl::Buffer::ComputeSize(buffer_state, pOffsets[i], VK_WHOLE_SIZE);
```
spencer-lunarg, line 5190: looked, no one in core checked seemed to care about this size being the "real" size or not (some VUs and error messages we look for `VK_WHOLE_SIZE`)

but in `vvl::VertexBufferBinding` please add a comment to `VkDeviceSize size;` that if it is `VK_WHOLE_SIZE` it will have been calculated

spencer-lunarg: ok, just realized this is matching what `PreCallRecordCmdBindVertexBuffers(1)` was doing... cool, ok, well if you can still put a comment while here, awesome, otherwise ignore

spencer-lunarg (approved):

layers/gpu/resources/gpuav_vulkan_objects.h
arno-lunarg: No I did not hit this, just realized that `Create` was not propagating errors so I just updated that!

tests/unit/gpu_av_indirect_buffer_positive.cpp
arno-lunarg: indeed

tests/CMakeLists.txt
arno-lunarg: Yes I will move new relevant tests to this file

layers/state_tracker/state_tracker.cpp
arno-lunarg: To remove some noise from this PR, I will do this in a subsequent PR

PR #8954, opened 2024-12-03
sync: Replace bitmask with index where possible

artem-lunarg:
Updated some places where `SyncStageAccessFlags` (`std::bitset`) always contains a single bit (similar to Vulkan `FlagBits`). Use index instead. Saves around 20 bytes. The main goal here is not optimization but to simplify things. Having a bitset type, it's often not clear whether it can contain a collection of bits or only a single access.

Some central types got smaller:
HazardState: 64 bytes -> 40 bytes
ReadState   : 72 bytes - > 56 bytes
ResourceAccessState: 536 bytes -> 488 bytes

Potentially can bring small performance improvement. The following was observed: doom capture 4 fps -> 4.4 fps, cs2: 7.9 -> 8.1), but this is not very reliable, I have some issues with local machine, need to recheck later.

### DIFF
--- a/layers/sync/sync_access_state.cpp
+++ b/layers/sync/sync_access_state.cpp
@@ -63,7 +63,7 @@ HazardResult ResourceAccessState::DetectHazard(const SyncStageAccessInfoType &us
         if (last_reads.size()) {
             for (const auto &read_access : last_reads) {
                 if (IsReadHazard(usage_stage, read_access)) {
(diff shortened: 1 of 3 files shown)


spencer-lunarg (approved):
There is actually quite a few spots in the SPIR-V area (non-GPU-AV) that could also use some bit masks like this

PR #8957, opened 2024-12-04
gpu: Fix GPU-AV Perf Regression

spencer-lunarg:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8937 introduced a huge perf regression, the way around it is to just check the total Descriptor Count and if high, defer work to the GPU


arno-lunarg (approved):

artem-lunarg:
I noticed this regression in my local testing too, in dota2 and doom captures. Can confirm this PR fixes the issue.

jeremyg-lunarg (approved):

Issue #8940, opened 2024-11-29
False positive with `VUID-vkCmdDraw-None-09600`


ShabbyX:
Encountered in ANGLE end2end test. To reproduce, unsuppress `VUID-vkCmdDraw-None-09600` (previously `UNASSIGNED-CoreValidation-DrawState-InvalidImageLayout`) in `vk_renderer.cpp` and run:

```
./angle_end2end_tests --gtest_filter=VulkanExternalImageTest.PreInitializedOnGLImport/ES2_Vulkan
```

VVL complains that:

```
[ VUID-vkCmdDraw-None-09600 ] Object 0: handle = 0x56051096b820, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0x1d000000001d, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x46582f7b | vkQueueSubmit(): pSubmits[0].pCommandBuffers[0] command buffer VkCommandBuffer 0x56051096b820[] expects VkImage 0x1d000000001d[] (subresource: aspectMask VK_IMAGE_ASPECT_COLOR_BIT array layer 0, mip level 0) to be in layout VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL--instead, current layout is VK_IMAGE_LAYOUT_UNDEFINED.
```

The test does not do any draw calls at all! In fact, the only thing in this submission is an image layout transition (and a global memory barrier), where the image transitions from COLOR_ATTACHMENT_OPTIMAL to TRANSFER_DST. What's tricky perhaps is the fact that the test uses two VkDevices, which may be the source of bugs?

This VVL failure started happening since we enabled sync2 (but not actually use it) which may be another clue.

spencer-lunarg:
Can reproduce with simple test

```c++
TEST_F(PositiveImage, AngleIssue) {
    RETURN_IF_SKIP(Init());
    m_command_buffer.Begin();

    VkImageCreateInfo image_create_info = vku::InitStructHelper();
    image_create_info.imageType = VK_IMAGE_TYPE_2D;
    image_create_info.format = VK_FORMAT_R8G8B8A8_UNORM;
    image_create_info.extent = {1, 1, 1};
    image_create_info.mipLevels = 1;
    image_create_info.arrayLayers = 1;
    image_create_info.samples = VK_SAMPLE_COUNT_1_BIT;
[...]

PR #8960, opened 2024-12-04
layers: Don't validate external image layouts

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8940 (well fixes the false positives for now)

### DIFF
--- a/layers/core_checks/cc_image_layout.cpp
+++ b/layers/core_checks/cc_image_layout.cpp
@@ -201,6 +201,12 @@ bool CoreChecks::ValidateCmdBufImageLayouts(const Location &loc, const vvl::Comm
         const auto image_state = Get<vvl::Image>(image);
         if (!image_state) continue;
 
+        // TODO - things like ANGLE might have external images which have their layouts transitioned implicitly
+        // https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8940
+        if (image_state->external_memory_handle_types != 0) {
+            continue;
+        }
+
         const auto &layout_map = layout_map_entry.second.map->GetLayoutMap();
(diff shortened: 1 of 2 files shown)


layers/gpu/descriptor_validation/gpuav_image_layout.cpp
```diff
@@ -161,6 +161,13 @@ static bool VerifyImageLayoutRange(const Validator &gpuav, const vvl::CommandBuf
     if (!subresource_map) {
         return skip;
     }
+
+    // TODO - things like ANGLE might have external images which have their layouts transitioned implicitly
+    // https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8940
+    if (image_state.external_memory_handle_types != 0) {
```
artem-lunarg, line 167: I wonder if it's possible to do this in theory. At  least for synchronization, in some cases we can't track what happens to external object and just skip validation.

artem-lunarg (approved):

layers/core_checks/cc_image_layout.cpp
```diff
@@ -201,6 +201,12 @@ bool CoreChecks::ValidateCmdBufImageLayouts(const Location &loc, const vvl::Comm
         const auto image_state = Get<vvl::Image>(image);
         if (!image_state) continue;
 
+        // TODO - things like ANGLE might have external images which have their layouts transitioned implicitly
```
ShabbyX, line 204: A more accurate description could be something like "validation layers cannot currently track layout transitions that happen on another Vulkan device (or a GL device through GL_EXT_semaphore)"

PR #8962, opened 2024-12-04
tests: Use profiles layer support for video

aqnuep:
Now that the Vulkan Profiles layer supports video profiles, switch to using that functionality for video capability and format emulation instead of manually written Test ICD code.


spencer-lunarg (approved):
JSON seems better then hardcoded C++ values

aqnuep:
> JSON seems better then hardcoded C++ values

Certainly, but it wasn't quite as simple as you would think to add this support to Vulkan-Profiles. The layer code size almost doubled with the support for video profiles, let alone complexity :)

spencer-lunarg:
@aqnuep I created this branch https://github.com/spencer-lunarg/Vulkan-ValidationLayers/tree/spencer-lunarg-profile-test-1 and it shows the same issue (all I did was update the `tests.py` and `known_good.json`)

So the issue is around the new changes in the profile layer logic it seems

aqnuep:
> @aqnuep I created this branch https://github.com/spencer-lunarg/Vulkan-ValidationLayers/tree/spencer-lunarg-profile-test-1 and it shows the same issue (all I did was update the `tests.py` and `known_good.json`)
> 
> So the issue is around the new changes in the profile layer logic it seems

You know what's funny?

I also have this: https://github.com/RasterGrid/Vulkan-ValidationLayers/commit/1a0c32460861338bc973292467a78049d430c738

Which is just updating the dependency, and it runs fine (I mean fast).

Then there is this: https://github.com/RasterGrid/Vulkan-ValidationLayers/actions/runs/12177110404/job/33964322240

Which also adds the JSON changes, but without enabling the simulation of video caps and it runs slow.

Also, the branch you've created updates the dependency, enables simulation of video caps, but does not actually includes video caps in the max_profile.json, and it runs slow

The interesting thing is that if the profile JSON does not contain any video caps, then enabling the simulation of video caps is a no-op, so it should actually run fast as my first attempt.

Now I'm not sure what to make of that.

aqnuep:
To summarize:

1. New Profiles layer code _without_ video-specific JSON and _without_ enabling video-cap simulation => TSAN runs _fast_ (my first test)
2. New Profiles layer code **with** video-specific JSON and _without_ enabling video-cap simulation => TSAN runs **slow** (my second test)
3. New Profiles layer code _without_ video-specific JSON and **with** enabling video-cap simulation => TSAN runs **slow** (your test)
4. New Profiles layer code **with** video-specific JSON and **with** enabling video-cap simulation => TSAN runs **slow** (this PR)

I'm really scratching my head, considering all the code is ASAN clean and the Profiles layer has no threading hazards as it always locks a global mutex in every entry point...

aqnuep:
Okay, in the meantime I ran some more variations of your test (3):

  3. (A) New Profiles layer code _without_ video-specific JSON and **with** enabling video-cap-only, but not video-format simulation => TSAN runs **slow**
  3. (B) New Profiles layer code _without_ video-specific JSON and **with** enabling video-format-only, but not video-cap simulation => TSAN runs _fast_

Go figure...

aqnuep:
@spencer-lunarg, looks there there's a failure on your CI, but only due to an update_deps error (probably network related).

There are also seemingly new failures of `NegativeShaderObject.InvalidColorWriteMask`, which:

- should not be affected by the change
- did not show up earlier, only since my rebases this week
- do not reproduce locally, which is weird, considering that these tests should only be dependent on the profile and the test ICD

Did you encounter this anywhere else?

spencer-lunarg:
> Did you encounter this anywhere else?

No, the past tests skipped it actually (on purpose if you look at the ICD `GetPhysicalDeviceImageFormatProperties` code

image

I pulled your change down now, I can reproduce ... the issue is before we used the `GetPhysicalDeviceImageFormatProperties2` inside of TestICD which has its own hard coded values, but seems your Profiles change now has it call into the `GetPhysicalDeviceImageFormatProperties2` inside profiles layer instead which overwrites it

.my advise is put a `if (IsPlatformMockICD())` on that test with a comment explaining this, the fact this test was not even tested before makes me less worried about adding it

the "core" issue is we don't have a proper test ICD where we can configure things at the start of test yet, so we still have random things hardcoded everywhere

aqnuep:
> No, the past tests skipped it actually (on purpose if you look at the ICD GetPhysicalDeviceImageFormatProperties code

Yes, it is still skipped when ran locally or on your own CI.

> I pulled your change down now, I can reproduce ...

You can? I could not. What system are you testing on? Anything specific I should

> the issue is before we used the GetPhysicalDeviceImageFormatProperties2 inside of TestICD which has its own hard coded values, but seems your Profiles change now has it call into the GetPhysicalDeviceImageFormatProperties2 inside profiles layer instead which overwrites it

That doesn't seem right. The GetPhysicalDeviceImageFormatProperties2 was coming from the Profiles layer for a very long time (if not since the beginning, blame says the corresponding profiles code is at least 2 years old).

The only change I made to the profiles there is that GetPhysicalDeviceImageFormatProperties2 will look at any video profiles included in the pNext chain to further restrict the caps, but it does not apply to this test case.

spencer-lunarg:
> You can? I could not. What system are you testing on? Anything specific I should

I have a bash script to set my env up, here is what it is on right now

```
VK_ICD_FILENAMES=/github/Vulkan-ValidationLayers/build/tests/icd/VVL_Test_ICD.json
VK_INSTANCE_LAYERS=VK_LAYER_KHRONOS_validation:VK_LAYER_KHRONOS_profiles
VK_LAYER_PATH=/github/Vulkan-ValidationLayers/build/layers:/github/Vulkan-Profiles/build/layer/
VK_KHRONOS_PROFILES_PROFILE_FILE=/github/Vulkan-ValidationLayers/tests/device_profiles/max_profile.json
```

(note, I build profiles in own repo instead of using the pulled down version CI does, but I matched the sha you had in this PR)

spencer-lunarg:
ah, yes, I have it at 

`export VK_KHRONOS_PROFILES_SIMULATE_CAPABILITIES=SIMULATE_API_VERSION_BIT,SIMULATE_FEATURES_BIT,SIMULATE_PROPERTIES_BIT,SIMULATE_EXTENSIONS_BIT,SIMULATE_FORMATS_BIT,SIMULATE_QUEUE_FAMILY_PROPERTIES_BIT,SIMULATE_VIDEO_CAPABILITIES_BIT,SIMULATE_VIDEO_FORMATS_BIT`

which matches the tests.py logic

aqnuep:
Intriguing... I wonder then how it does not reproduce on my system where the test case is correctly skipped (Ubuntu 22.04, practically with the same env setup).

aqnuep:
I looked at `max_profile` and it contained a lot of bogus format feature flags for `VK_FORMAT_E5B9G9R9_UFLOAT_PACK32`.

Apparently, the same is true for most other formats, but for now I attempted removing them at least for this format to see if it resolves the test case issue.

I'm kind of still surprised how this happens only now and why does it not reproduce locally. But let's see what the CI says to this.

aqnuep:
Ah, I know why.

There was a long standing issue in the Profiles layer where all flags have been handled as 32-bit values. I fixed that to correctly return 64-bit values.

So probably what we see here is that >32 bits specified in the `max_profile` were disappearing in the Profiles layer before, but now they are not, hence show new capabilities.

I still don't understand how this does not reproduce for me locally, but it's worth looking into.

aqnuep:
Okay, in the end I ended up removing support in `max_profiles` for RGB9E5, because it will not work with the hard-coded logic in the Test ICD.

The Test ICD intentionally does not support RGB9E5 (hard-coded), and because the Profiles layer overrides it to support those feature flags, GPDIFP2 will return `VK_SUCCESS` (correctly, according to Profiles layer logic), but the values returned in the image format properties (such as `maxMipLevels`) will be zero (again, hard-coded, coming from the Test ICD), and this is what causes the failures.

The above is still true: this was not observable as a test failure before because the Profiles layer only supported the first 32 bits of all flags, even 64-bit ones like the `VkFormatFeatureFlags2`, but now that it does, the inconsistencies between the Test ICD and `max_profile.json` surfaced.

spencer-lunarg:
@aqnuep thanks for helping look into this

PR #8968, opened 2024-12-05
gpu: Support Slang global BDA style pointers

spencer-lunarg:
Found out with Slang you can have a BDA reference not tied to a struct (https://godbolt.org/z/1eKesYaPW)

Currently we were not validating it, this adds logic to detect it as well as tests

### DIFF
--- a/layers/gpu/spirv/instruction.cpp
+++ b/layers/gpu/spirv/instruction.cpp
@@ -126,6 +126,17 @@ void Instruction::ReplaceOperandId(uint32_t old_word, uint32_t new_word) {
     }
 }
 
+bool Instruction::IsArray() const {
+    const uint32_t opcode = Opcode();
+    return opcode == spv::OpTypeArray || opcode == spv::OpTypeRuntimeArray;
+}
+
+bool Instruction::IsAccessChain() const {
+    const uint32_t opcode = Opcode();
+    return opcode == spv::OpAccessChain || opcode == spv::OpPtrAccessChain || opcode == spv::OpInBoundsAccessChain ||
+           opcode == spv::OpInBoundsPtrAccessChain;
+}
+
 // The main challenge with linking to functions from 2 modules is the IDs overlap.
 // TODO - Use the new generated operand to find the IDs.
 void Instruction::ReplaceLinkedId(vvl::unordered_map<uint32_t, uint32_t>& id_swap_map) {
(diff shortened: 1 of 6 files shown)


layers/gpu/spirv/instruction.cpp
```diff
@@ -126,6 +126,17 @@ void Instruction::ReplaceOperandId(uint32_t old_word, uint32_t new_word) {
     }
 }
 
+bool Instruction::IsArray() const {
```
artem-lunarg, line 129: Is this method for future uses?

artem-lunarg (approved):

tests/unit/gpu_av_buffer_device_address.cpp
```diff
@@ -1427,3 +1427,75 @@ TEST_F(NegativeGpuAVBufferDeviceAddress, LoadAlignment) {
...
+    pipe.descriptor_set_->UpdateDescriptorSets();
+
+    m_command_buffer.Begin();
+    vk::CmdBindPipeline(m_command_buffer.handle(), VK_PIPELINE_BIND_POINT_COMPUTE, pipe.Handle());
+    vk::CmdBindDescriptorSets(m_command_buffer.handle(), VK_PIPELINE_BIND_POINT_COMPUTE, pipe.pipeline_layout_.handle(), 0, 1,
+                              &pipe.descriptor_set_->set_, 0, nullptr);
+    vk::CmdDispatch(m_command_buffer.handle(), 1, 1, 1);
+    m_command_buffer.End();
+
+    m_errorMonitor->SetDesiredError("UNASSIGNED-Device address out of bounds");
```
arno-lunarg, line 1497: Add OOB error info to error message detection
```suggestion
    m_errorMonitor->SetDesiredError("UNASSIGNED-Device address out of bounds");
```

arno-lunarg (approved):

layers/gpu/spirv/instruction.cpp
spencer-lunarg: it was just inlined in the header before, was always there

tests/unit/gpu_av_buffer_device_address.cpp
spencer-lunarg: I have an upcoming PR to clean up the GPU-AV tests, for now this is unfortunately consistent with the other tests

PR #8976, opened 2024-12-07
settings: Revert enable_message_limit

spencer-lunarg:
reverts https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8784/commits/ff8e3acfcac128f9faabb74d8b62206ff0195939

Now logic is 

- default is limit to 10
   - This is new, as pointed out that we have tests failing because of this
- `enable_message_limit = false`  there is no limit
- `duplicate_message_limit = 0` (which is possible outside of vkconfig) there is no limit

@ziga-lunarg @artem-lunarg you 2 wanted this back (and agree) but please verify this is what you wanted

(the vkconfig GUI looks the exact same it was before)


ziga-lunarg (approved):

artem-lunarg (approved):

spencer-lunarg:
> CI Vulkan-ValidationLayers build # 18307 aborted.

is just a networking issues with on of the linux machines that passed on the 64-bit run

PR #8977, opened 2024-12-07
layers: Misc cleanup of Copy Buffer Image code

spencer-lunarg:
This does a lot of things around the Copy Buffer/Image code

1. Reduce places we loop the `pRegion`
2. Keep src and dst together (prevents cache trashing going between)
3. Make more use of VUID maps to not need to pass `const char*` everywhere possible


layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -2430,76 +2485,61 @@ void CoreChecks::PostCallRecordCmdCopyBufferToImage2(VkCommandBuffer commandBuff
...
-}
-template <>
-VkImage GetImage<VkCopyMemoryToImageInfo>(VkCopyMemoryToImageInfo data) {
-    return data.dstImage;
-}
-template <>
-VkImage GetImage<VkCopyImageToMemoryInfo>(VkCopyImageToMemoryInfo data) {
-    return data.srcImage;
-}
+VkImageLayout GetImageLayout(VkCopyMemoryToImageInfo data) { return data.dstImageLayout; }
```
artem-lunarg, line 2529: 👍

artem-lunarg (approved):

Issue #8969, opened 2024-12-05
Post 1.4 cleanup


arno-lunarg:
- https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8966#discussion_r1871146500
  - Make the `VkPhysicalDeviceVulkan14Features` device check look the same for older versions (VUID-VkDeviceCreateInfo-pNext-10360)

PR #8980, opened 2024-12-07
Fix VUID for 1.4 device features

ziga-lunarg:


spencer-lunarg (approved):
Part of https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8969

... wow, missed this, but ya, these were the wrong version features, thanks for fixing

Issue #1998, opened 2020-07-05
UNASSIGNED-CoreValidation-DrawState-InvalidImageLayout false positive with array textures


Ralith:
To reproduce:
1. Create an array texture
2. Initialize a subset of the layers
3. Read from those layers in a shader
4. Observe flood of validation errors regarding unread layers

This can be worked around by transitioning the entire image up front, but that can be inconvenient.

locke-lunarg:
@Ralith @mark-lunarg 

I added a test to reproduce this issue, but it couldn't.
The image's arrayLayer is 2. It only changes layer:0 into correct layout. layer:1 is still wrong layout. ImageView uses only layer:0, so it supposed to trigger the false positive. But it doesn't. Could you help me to reproduce the false positive? Thanks.

https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/3d7dd9aab255b7229cdbe9b93d15af5acf2cd0c2

mark-lunarg:
@Ralith, can you please take a look at the test @locke-lunarg created, and see if you can tell how it differs from your use-case?  Also, which version of the layers are generating these errors?

Ralith:
> ImageView uses only layer:0

I'm having a difficult time getting master to build, but this is probablythe divergence. In my case, the image view covers the entire array, but the shader only dynamically accesses layers in the correct layout.

locke-lunarg:
@Ralith 
Yes. If the imageview covers the entire array, and some layers don't change to the correct layout, it will trigger positive because this validation only verify VkImageView data. It doesn't verify what happen in shader. Could you give me a sample how you do in shader "only dynamically accesses layers in the correct layout"? Thanks.

Ralith:
Right, and that's an annoying false positive, to the point that I've had to complicate my code to hack around the issue by performing unnecessary layout transitions.

My use case is a very large array where each layer corresponds to a heightmap for a particular region of terrain. As the viewpoint moves around, data is streamed into layers that aren't currently being sampled. An instance buffer instructs drawing commands which layers to sample from, so layers that have not been initialized or which are currently being written are never read from.

locke-lunarg:
Sorry for inconvenient. We'll schedule for it.

Ralith:
Maybe a good approach here would be to emit an error iff no subresource is in the correct layout? That should still catch most common errors while being conservative when there's clever shader stuff going on.

locke-lunarg:
[...]

PR #8982, opened 2024-12-09
tests: Prove image layout works with descriptor indexing

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/1998

Just adding tests (which we really have none for image layout) to prove things work now


tests/unit/gpu_av.cpp
```diff
@@ -1421,3 +1421,83 @@ TEST_F(NegativeGpuAV, AliasImageMultisampleDescriptorSetsPartiallyBound) {
     m_default_queue->Wait();
     m_errorMonitor->VerifyFound();
 }
+
+// move to future gpu_av_image_layout.cpp
```
spencer-lunarg, line 1425: post - https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8971 (plan to add more tests anyway afterwards)

PR #8983, opened 2024-12-09
Collapse the image layout map value

spencer-lunarg:
1. `ImageSubresourceLayoutMap` is not a map, it contains the `LayoutMap` and it is confusing, renamed to reflect that
2. Moved the `uint32_t` (`IdType`) inside `ImageSubresourceLayoutInfo` so that the other `image_layout_map` is simpler to use

### DIFF
--- a/layers/state_tracker/cmd_buffer_state.h
+++ b/layers/state_tracker/cmd_buffer_state.h
@@ -41,7 +41,7 @@ class VideoSessionParameters;
 
 // Only CoreChecks uses this, but the state tracker stores it.
 constexpr static auto kInvalidLayout = image_layout_map::kInvalidLayout;
-using ImageSubresourceLayoutMap = image_layout_map::ImageSubresourceLayoutMap;
+using ImageSubresourceLayoutInfo = image_layout_map::ImageSubresourceLayoutInfo;
 
 struct EventInfo {
     VkPipelineStageFlags2 src_stage_mask = VK_PIPELINE_STAGE_2_NONE;
@@ -161,10 +161,10 @@ class CommandBuffer : public RefcountedStateObject {
   public:
     struct LayoutState {
         StateObject::IdType id;
-        std::shared_ptr<ImageSubresourceLayoutMap> map;
+        std::shared_ptr<ImageSubresourceLayoutInfo> info;
     };
     using ImageLayoutMap = vvl::unordered_map<VkImage, LayoutState>;
-    using AliasedLayoutMap = vvl::unordered_map<const GlobalImageLayoutRangeMap *, std::shared_ptr<ImageSubresourceLayoutMap>>;
+    using AliasedLayoutMap = vvl::unordered_map<const GlobalImageLayoutRangeMap *, std::shared_ptr<ImageSubresourceLayoutInfo>>;
 
     VkCommandBufferAllocateInfo allocate_info;
     VkCommandBufferBeginInfo beginInfo;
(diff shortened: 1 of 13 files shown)


layers/state_tracker/cmd_buffer_state.h
```diff
@@ -159,12 +159,8 @@ struct LabelCommand {
 class CommandBuffer : public RefcountedStateObject {
     using Func = vvl::Func;
   public:
-    struct LayoutState {
```
spencer-lunarg, line 162: removing this because the `id` was just the `image_state.id` found inside the old `ImageSubresourceLayoutMap`

layers/state_tracker/cmd_buffer_state.h
```diff
@@ -579,9 +575,9 @@ class CommandBuffer : public RefcountedStateObject {
 
     void ResetPushConstantRangesLayoutIfIncompatible(const vvl::PipelineLayout &pipeline_layout_state);
 
-    std::shared_ptr<const ImageSubresourceLayoutMap> GetImageSubresourceLayoutMap(VkImage image) const;
-    std::shared_ptr<ImageSubresourceLayoutMap> GetImageSubresourceLayoutMap(const vvl::Image &image_state);
-    const ImageLayoutMap &GetImageSubresourceLayoutMap() const;
+    std::shared_ptr<const ImageSubresourceLayoutInfo> GetImageSubresourceLayoutInfo(VkImage image) const;
+    std::shared_ptr<ImageSubresourceLayoutInfo> GetOrCreateImageSubresourceLayoutInfo(const vvl::Image &image_state);
```
spencer-lunarg, line 579: named `GetOrCreateImageSubresourceLayoutInfo` because it is not easy to realize right away one is might create up under you otherwise

layers/state_tracker/state_object.h
```diff
@@ -49,7 +49,6 @@ class StateObject: public std::enable_shared_from_this<StateObject>, public Type
     // parent types without locking every weak_ptr.
     using NodeMap = unordered_map<VulkanTypedHandle, std::weak_ptr<StateObject>>;
     using NodeList = small_vector<std::shared_ptr<StateObject>, 4, uint32_t>;
-    using IdType = uint32_t;
```
spencer-lunarg, line 52: I found this not helpful at all, things that don't want to include all of `state_object.h` now either need to redefine a `uint32_t`

tests/unit/image_layout.cpp
```diff
@@ -243,6 +243,69 @@ TEST_F(NegativeImageLayout, Compute11) {
     m_errorMonitor->VerifyFound();
 }
 
+TEST_F(NegativeImageLayout, TrackingCommand) {
```
spencer-lunarg, line 246: Currently using this test to debug things, while not related to this PR, got commited by accident and honestly more tests is always nice :laughing:

artem-lunarg: Maybe a similar comment for the test? Sometimes I add tests that serve more of a debugging helper, then the comment helps to find the purpose later.

layers/state_tracker/image_layout_map.h
```diff
@@ -51,7 +51,8 @@ struct InitialLayoutState {
     InitialLayoutState() : image_view(VK_NULL_HANDLE), aspect_mask(0), label() {}
 };
 
-class ImageSubresourceLayoutMap {
```
artem-lunarg, line 54: > ImageSubresourceLayoutMap is not a map, it contains the LayoutMap and it is confusing, renamed to reflect that

One can argue it's an abstraction around map and indicates it stores layout info for each subresource (in case if that's what it describes, I don't know for sure tbh).
New name sounds it's about layout of a *single* image subresource and not about all layouts. Can this be in plural form, ImageSubresourceLayoutInfos or simpler variants like ImageLayoutRegistry, ImageLayouts, etc.?

tests/unit/image_layout.cpp
spencer-lunarg: good call, can add

layers/state_tracker/image_layout_map.h
spencer-lunarg: > ImageLayoutRegistry

:100: better already, going to use that

artem-lunarg (approved):
LGTM. One remaining nitpick about registry word is that it sounds a bit singleton-ish but not sure about better options.

Issue #8931, opened 2024-11-26
Fix TSAN bogging down runs with Video Test Framework


spencer-lunarg:
The simple change in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8929 reduced the Max Profile Thread Sanitizer run from close to 3 hours down to 30 minutes

There is something going on in the Video Test framework that is making TSAN not happy. The tests slowly get slower and slower and then tests after that normally take 100ms to run take multiple minutes to run

jeremyg-lunarg:
part of the problem is that the `if` guard on [2 of the test runs](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/.github/workflows/vvl.yml#L119) is not working as intended.

somewhat of a separate issue that what happened with the video tests but it makes them run 3 times instead of 1.

PR #8986, opened 2024-12-09
ci: Reduce TSAN to tests that need it

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8931

This limits TSAN to threading, sync val, sync object, and WSI tests for now (which is 348 tests currently)

cc @artem-lunarg @aqnuep


artem-lunarg (approved):
LGTM. I agree it's a reasonable subset of the test for TSAN.

PR #8987, opened 2024-12-09
tests: Add test for GPL and legacy dithering

Maro1:
The VK_PIPELINE_CREATE_2_ENABLE_LEGACY_DITHERING_BIT_EXT flag is part of the fragment output state, but is not tracked by the fragment output state of the validation layer. This can cause incorrect validation layer errors when the flag is included when creating a pipeline library, while not including it in the final pipeline. This PR fixes this by adding the flag to the fragment output state, and adds a test to the positive graphics pipeline library test class.

Addresses issue: #8988


spencer-lunarg (approved):
looks good, thanks for writing a test as well!

PR #8996, opened 2024-12-10
layers: Cleanup Buffer::ComputeSize

arno-lunarg:
It always has been `ComputeValidSize` under the hood, so remove needless abstractions

### DIFF
--- a/layers/state_tracker/buffer_state.h
+++ b/layers/state_tracker/buffer_state.h
@@ -53,16 +53,20 @@ class Buffer : public Bindable {
     }
 
     VkBuffer VkHandle() const { return handle_.Cast<VkBuffer>(); }
-    std::optional<VkDeviceSize> ComputeValidSize(VkDeviceSize offset, VkDeviceSize size) const {
-        return ::ComputeValidSize(offset, size, create_info.size);
+
+    VkDeviceSize GetRegionSize(VkDeviceSize offset, VkDeviceSize size) const {
+        if (offset < create_info.size) {
+            if (size == VK_WHOLE_SIZE) {
+                return create_info.size - offset;
+            } else if ((offset + size) <= create_info.size) {
+                return size;
+            }
+        }
+        return 0;
     }
-    VkDeviceSize ComputeSize(VkDeviceSize offset, VkDeviceSize size) const {
-        std::optional<VkDeviceSize> valid_size = ComputeValidSize(offset, size);
-        return valid_size.has_value() ? *valid_size : VkDeviceSize(0);
-    }
-    static VkDeviceSize ComputeSize(const std::shared_ptr<const Buffer> &buffer_state, VkDeviceSize offset,
-                                    VkDeviceSize size) {
-        return buffer_state ? buffer_state->ComputeSize(offset, size) : VkDeviceSize(0);
+
+    static VkDeviceSize GetRegionSize(const std::shared_ptr<const Buffer> &buffer_state, VkDeviceSize offset, VkDeviceSize size) {
+        return buffer_state ? buffer_state->GetRegionSize(offset, size) : 0;
     }
 
     sparse_container::range<VkDeviceAddress> DeviceAddressRange() const {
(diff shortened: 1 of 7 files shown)


layers/state_tracker/buffer_state.h
```diff
@@ -53,16 +53,21 @@ class Buffer : public Bindable {
...
 
     VkBuffer VkHandle() const { return handle_.Cast<VkBuffer>(); }
-    std::optional<VkDeviceSize> ComputeValidSize(VkDeviceSize offset, VkDeviceSize size) const {
-        return ::ComputeValidSize(offset, size, create_info.size);
-    }
-    VkDeviceSize ComputeSize(VkDeviceSize offset, VkDeviceSize size) const {
-        std::optional<VkDeviceSize> valid_size = ComputeValidSize(offset, size);
-        return valid_size.has_value() ? *valid_size : VkDeviceSize(0);
+
+    VkDeviceSize ComputeValidSize(VkDeviceSize offset, VkDeviceSize size) const {
```
artem-lunarg, line 57: Is it possible to compute invlalid size?

arno-lunarg: Valid is related to both input being invalid (offset past buffer, size too big for buffer), or size being `VK_WHOLE_SIZE`
So yes maybe `Valid` is not the best choice here...
Maybe `ComputeEffectiveSize`? Or `ComputeEffectiveSizeOrZero`? But I feel that `OrZero` is not strictly necessary here

artem-lunarg: >  input being invalid (offset past buffer, size too big for buffer),

This part of invalid input sometimes just asserted and should be handled by the caller. Is it something that fits current implementation?

artem-lunarg: > Maybe ComputeEffectiveSize? Or ComputeEffectiveSizeOrZero? 

It would be nice to have just "buffer size". Invalid input should not change buffer size. Having function in the Buffer interface with "effective" make thing look a bit more complex.

arno-lunarg: I think it has to be clear that this is not `create_info.size`, but an actual valid/effective size, so going for `BufferSize()` or `ComputeBufferSize()` is not a good option, too easy to think "oh I can use that instead of accessing create_info.size"

artem-lunarg: Hm.. Isn't buffer size is exactly "create_info.size" ? Yes, you can allocate more memory, but it's what drivers needs. For client perspective it's exactly create_info.size. Or is it about something else?

artem-lunarg: Okay, then last comment :) All interfaces with buffer->GetSize()/GetEffectiveSize() you read this as this is size of the buffer. If the meaning of the returned value is actually not a buffer size but a region within a buffer defined by parameters then maybe indicated this in the name. Similar to GetBufferRegionSize()?

artem-lunarg (approved):
LGTM. All feedback is left, naming is hard. One more idea, maybe this function is better as a free one and not part of the buffer interface.

layers/state_tracker/buffer_state.h
arno-lunarg: Aaah yes `GetBufferRegionSize` better encapsulates function intents. I will go for that!

layers/state_tracker/vertex_index_buffer_state.h
```diff
@@ -26,11 +26,13 @@ class Buffer;
 
 struct VertexBufferBinding {
     VkBuffer buffer;  // VK_NULL_HANDLE is valid if using nullDescriptor
-    VkDeviceSize size;
+    // Binding valid size: 0 if buffer is not tracked, actual size if VK_WHOLE_SIZE was specified,
+    // clamped up to 0 if specified size is greater than the size actually left in the buffer
+    VkDeviceSize valid_size;
```
spencer-lunarg, line 31: For `VK_WHOLE_SIZE` there are things in VVL we call `effective_range`

image

... I would much prefer this be called `effective_size` than `valid_size`

spencer-lunarg (approved):
LGTM - but still for the `effective_size` name to match closer to spec terms around `VK_WHOLE_SIZE`

layers/state_tracker/vertex_index_buffer_state.h
arno-lunarg: Ok, will change!

PR #8997, opened 2024-12-10
gpuav: Add debug label regions to validation cmds

arno-lunarg:


layers/error_message/error_location.h
```diff
@@ -47,6 +48,7 @@ struct Location {
     Location(const Location& prev_loc, vvl::Struct s, vvl::Field f, uint32_t i, bool p)
         : function(prev_loc.function), structure(s), field(f), index(i), isPNext(p), prev(&prev_loc) {}
 
+    void SetDebugRegion(std::string& debug_region) { this->debug_region = &debug_region; }
```
arno-lunarg, line 51: No `const std::string&` to make passing temporary `std::string` illegal: pointer would be dangling

layers/error_message/error_location.h
```diff
@@ -32,13 +32,14 @@ struct Location {
...
-    const vvl::Field field;
-    const uint32_t index;  // optional index if checking an array.
-    const bool isPNext;    // will print the struct is from a 'pNext` chain
-    const Location* prev;
+    const vvl::Struct structure{};
+    const vvl::Field field{};
+    const uint32_t index{};  // optional index if checking an array.
+    const bool isPNext{};    // will print the struct is from a 'pNext` chain
+    const Location* prev{};
+    mutable const std::string* debug_region{};
```
arno-lunarg, line 42: `Location` always being passed around as `const`, I made `debug_region` mutable to avoid having to modify all `const Location&`

spencer-lunarg: The issue I forsee (and why I made things const) is if you go

```c++
skip |= LogError(VUID-A, loc);
skip |= ValidateThings(loc);
skip |= LogError(VUID-C, loc);
```

will the `ValidateThings` ever unexpectedly change `loc` on you in an undesirable way

spencer-lunarg: I think this "just works" because a `Location` is created at a function entry level and we don't have Debug Region that change as you walk down a struct/pNext chain

tests/CMakeLists.txt
```diff
@@ -118,6 +118,7 @@ target_sources(vk_layer_validation_tests PRIVATE
     unit/gpu_av_image_layout.cpp
     unit/gpu_av_indirect_buffer.cpp
     unit/gpu_av_indirect_buffer_positive.cpp
+    unit/gpu_av_index_buffer.cpp
```
spencer-lunarg, line 121: aww, so these were  not being ran before :laughing:

spencer-lunarg (approved):

tests/CMakeLists.txt
arno-lunarg: ごめんなさい 🙇‍♂️🙇‍♂️🙇‍♂️

layers/error_message/error_location.h
arno-lunarg: This is a valid concern, so I switched to creating a temporary `Location` storing debug region name for no longer than needed:
```cpp
Location loc_with_debug_region(loc, debug_region_name);
```

spencer-lunarg: ok, ya, we do this for the VUID maps as well.. I'm ok spending the extra little time here when reporting an error

Issue #8989, opened 2024-12-09
Validation Layer Crash


mbaxmann:
**Environment:**
 - OS: Windows 10
 - GPU and driver version: NVIDIA GeForce RTX 2070 SUPER 32.0.15.6614
 - SDK or header version if building from repo: Vulkan SDK 1.3.296
 - Options enabled (synchronization, best practices, etc.): Standard vulkan configurator with synchronization, submit time validation and shader accesses heuristic.

**Describe the Issue**

Validation layer will crash randomly at vvl::Semaphore::WaitRetire(). It will throw the following exception:
```
timepoint.wait_submits.working_store_ is nullptr.
```

**Expected behavior**

No validation error should be reported.

<details>
<summary>code or terminal output</summary>

```sh
VkLayer_khronos_validation.dll!vvl::Semaphore::RetireWait(vvl::Queue * current_queue, unsigned __int64 payload, const Location & loc, bool queue_thread) Ligne 288	C++
VkLayer_khronos_validation.dll!vvl::Queue::ThreadFunc() Ligne 229	C++
VkLayer_khronos_validation.dll!vvl::Queue::ThreadFunc() Ligne 229	C++
```

</details>

artem-lunarg:
@mbaxmann this does not provide much context of what's going on. Could you describe your scenario or provide minimal sample or reference to the project? Is the project open source?

mbaxmann:
I think I managed to find wich queue submission cause the issue
by looking at the submission.wait_semaphore struct in the vvl::Queue::Retire() function.

submission.wait_semaphore will always contain 3 semaphore when it crash. The first and second
semaphore are BIRNARY semaphore, where the third is a TIMELINE semaphore.
The second BINARY semaphore is the one who crash.

I am confident it's from my forward render pass command buffer

```	void Renderer::endForwardScene(void)
	{
		AVALON_PROFILE_FUNCTION();

		//Prepare Composite Quad
		glm::mat4 model = glm::scale(glm::mat4(1.0f), glm::vec3(2.0f));
		drawQuad(model, glm::vec4(0.0f, 0.0f, 0.0f, 1.0f), -1);
		_data.quadIndexCount -= 6;

		flush2D();
		flushForward3D();

		RenderCommand::nextSubpass(_sceneData->activeCommandBuffer);
		flushTransparent();

		RenderCommand::nextSubpass(_sceneData->activeCommandBuffer);
		flushComposite();

[...]

PR #8998, opened 2024-12-10
layers: Add assert for external semaphore code path

artem-lunarg:
Related to https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8989

The above issue has similar symptoms as infamous `PositiveSyncObject.KhronosTimelineSemaphoreExample` test. In both cases code can take "impossible" branch (like in the case when memory corruption changes object state). The added assert will help to detect such cases (`scope_` gets corrupted and has different value than `kInternal` for regular non-external semaphore) .

The original `assert(payload == 0);` assert was not correct. When binary and timeline semaphores are mixed, the semaphore value entries for binary semaphores should be ignored because they can contain arbitrary values.


artem-lunarg:
Closing this. `PositiveSparseBuffer.NonOverlappingBufferCopy` provides example when it's possible to reach that place with regular semaphore. Going to update code in a separate PR.

PR #9000, opened 2024-12-11
layers: Fix empty timeline assumption in Semaphore::EnqueueWait

artem-lunarg:
Working on https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8989 I found an incorrect assumption about when a specific code branch can be active. This PR provides a fix and adds a regression test to catch the issue (would trigger an assert in the original code).

Also found a hint why issues with `PositiveSyncObject.KhronosTimelineSemaphoreExampl`  on CI can be our problem and how to debug it. Will work on this next.

### DIFF
--- a/layers/state_tracker/semaphore_state.cpp
+++ b/layers/state_tracker/semaphore_state.cpp
@@ -113,16 +113,22 @@ void vvl::Semaphore::EnqueueSignal(const SubmissionReference &signal_submit, uin
 void vvl::Semaphore::EnqueueWait(const SubmissionReference &wait_submit, uint64_t &payload) {
     auto guard = WriteLock();
     if (type == VK_SEMAPHORE_TYPE_BINARY) {
-        // Timeline can be empty for the binary wait operation if the semaphore was imported.
-        // Otherwise timeline should contain a binary signal.
         if (timeline_.empty()) {
-            assert(payload == 0);
-            completed_ = SemOp(kWait, wait_submit, 0);
-            return;
+            if (scope_ != vvl::Semaphore::kInternal) {
+                // for external semaphore mark wait as completed, no guarantee of signal visibility
+                completed_ = SemOp(kWait, wait_submit, 0);
+                return;
+            } else {
+                // generate binary payload value from the last completed signals
+                assert(completed_.op_type == kSignal);
+                payload = completed_.payload;
+            }
+        } else {
+            // generate binary payload value from the most recent pending binary signal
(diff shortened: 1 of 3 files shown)


layers/state_tracker/state_tracker.cpp
```diff
@@ -1291,23 +1291,24 @@ void ValidationStateTracker::PreCallRecordQueueSubmit(VkQueue queue, uint32_t su
...
         const VkSubmitInfo *submit = &pSubmits[submit_i];
-        auto *timeline_semaphore_submit = vku::FindStructInPNextChain<VkTimelineSemaphoreSubmitInfo>(submit->pNext);
+        auto *timeline_info = vku::FindStructInPNextChain<VkTimelineSemaphoreSubmitInfo>(submit->pNext);
         for (uint32_t i = 0; i < submit->waitSemaphoreCount; ++i) {
+            auto wait_semaphore = Get<vvl::Semaphore>(submit->pWaitSemaphores[i]);
             uint64_t value{0};
-            if (timeline_semaphore_submit && timeline_semaphore_submit->pWaitSemaphoreValues != nullptr &&
-                (i < timeline_semaphore_submit->waitSemaphoreValueCount)) {
-                value = timeline_semaphore_submit->pWaitSemaphoreValues[i];
+            if (wait_semaphore->type == VK_SEMAPHORE_TYPE_TIMELINE && timeline_info && timeline_info->pWaitSemaphoreValues &&
```
artem-lunarg, line 1298: That's additional minor issue, we did not check semaphore type and did not ingore `VkTimelineSemaphoreSubmitInfo` values for binary semaphore. Do the same logic as we already have for QueueSubmit2.

layers/state_tracker/semaphore_state.cpp
```diff
@@ -113,16 +113,22 @@ void vvl::Semaphore::EnqueueSignal(const SubmissionReference &signal_submit, uin
 void vvl::Semaphore::EnqueueWait(const SubmissionReference &wait_submit, uint64_t &payload) {
     auto guard = WriteLock();
     if (type == VK_SEMAPHORE_TYPE_BINARY) {
-        // Timeline can be empty for the binary wait operation if the semaphore was imported.
```
artem-lunarg, line 116: This comment was a lie. `if (timeline_.empty())` can be true here for regular non-external semaphores.  One example is when semaphore signal submissions is followed by a queue wait.

spencer-lunarg (approved):

PR #9004, opened 2024-12-11
layers: Do not create wait timepoint after signal retired

artem-lunarg:
This is follow-up to https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9000 and reproduces another assert https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8989#issuecomment-2535685344

Wait timepoint should not be registered when signal is already retired.
In this case wait should be resolved againt `completed_` state.

There is assert in `Semaphore::CanBinaryBeSignaled` that checks that
binary timepoint has at least signal op. The new tests will trigger this
assert in case of a regression.

There might be other places to adjust, will continue to inspect code.

Addresses assert part of https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9009

### DIFF
--- a/layers/state_tracker/semaphore_state.cpp
+++ b/layers/state_tracker/semaphore_state.cpp
@@ -128,11 +128,19 @@ void vvl::Semaphore::EnqueueWait(const SubmissionReference &wait_submit, uint64_
             assert(timeline_.rbegin()->second.HasSignaler());
             payload = timeline_.rbegin()->first;
         }
-    } else {  // timeline semaphore
-        if (payload <= completed_.payload) {
-            return;
-        }
     }
+
+    if (payload <= completed_.payload) {
+        // Signal is already retired and its timepoint removed. Mark wait as completed.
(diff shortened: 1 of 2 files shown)


layers/state_tracker/semaphore_state.cpp
```diff
@@ -128,11 +128,19 @@ void vvl::Semaphore::EnqueueWait(const SubmissionReference &wait_submit, uint64_
             assert(timeline_.rbegin()->second.HasSignaler());
             payload = timeline_.rbegin()->first;
         }
-    } else {  // timeline semaphore
-        if (payload <= completed_.payload) {
-            return;
-        }
     }
+
+    if (payload <= completed_.payload) {
```
artem-lunarg, line 133: Previously this was only for timeline semaphores. The test shows it's also needed for binary semaphores.

artem-lunarg:
Additionally closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9009

spencer-lunarg (approved):

PR #9006, opened 2024-12-12
tests: Remove invalid positive array layer tests

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/1998

from https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8982

Was discussed in https://gitlab.khronos.org/vulkan/vulkan/-/issues/4110 that all image array layers in a single ImageView must be valid if any part of the descriptor is accessed. While this might not effect drivers that don't care about Image Layout, those that do care said this likely will break and the advice to either

1. Full transition the entire Image View where doing a sub-transition
2. Create a separate Image Views for each arrayLayer and use a descriptor indexing method on them (since things like Descriptor Indexing were designed to solve being able to have this dynamic access)
3. For more future things like `VK_EXT_descriptor_buffer` there is actually a `descriptorBufferImageLayoutIgnored` feature that was designed for things like this

cc @Ralith


tests/unit/gpu_av_image_layout_positive.cpp
```diff
@@ -93,79 +93,3 @@ TEST_F(PositiveGpuAVImageLayout, DescriptorArrayLayout) {
     m_default_queue->Submit(m_command_buffer);
     m_default_queue->Wait();
 }
-
-// Waiting to hear from https://gitlab.khronos.org/vulkan/vulkan/-/issues/4110
-TEST_F(PositiveGpuAVImageLayout, DISABLED_MultiArrayLayers) {
```
spencer-lunarg, line 98: we have negative variation of the tests already when these were added

artem-lunarg (approved):

Ralith:
> Was discussed in https://gitlab.khronos.org/vulkan/vulkan/-/issues/4110 that all image array layers in a single ImageView must be valid if any part of the descriptor is accessed

Where is this specified? What is the exact, formal, requirement? At face value, this seems to make it impossible to stream layers in an array texture concurrently with rendering, which is a painful constraint.

PR #9012, opened 2024-12-12
tests: Split up AttachmentDescriptionFinalLayout

spencer-lunarg:
makes `AttachmentDescriptionFinalLayout` 3 smaller tests for better maintainability


tests/unit/render_pass.cpp
```diff
@@ -167,173 +257,167 @@ TEST_F(NegativeRenderPass, AttachmentDescriptionFinalLayout) {
         GTEST_SKIP() << "Couldn't find a stencil only image format";
     }
 
-    if (separate_depth_stencil_layouts_features.separateDepthStencilLayouts) {
```
spencer-lunarg, line 170: we set `separateDepthStencilLayouts` to false above, so this whole section was not even being ran, this is what prompted me to make this PR

artem-lunarg (approved):

Issue #9014, opened 2024-12-12
Divide by zero crash validating single-draw vkCmdDrawMeshTasksIndirectEXT


baldurk:
**Environment:**
 - OS: Windows 11
 - GPU and driver version: AMD 7900 XTX driver 24.10.1
 - SDK or header version if building from repo: Built from b10d21aa975d5cf4ecf4d8d3c7ba4610e151bce6
 - Options enabled (synchronization, best practices, etc.): GPU validation & synchronization

**Describe the Issue**

If calling `vkCmdDrawMeshTasksIndirectEXT` (possibly others, but this was where it crashed for me) with `drawCount == 1` the spec says `stride` is ignored, but on this line:

https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/07e3442eac85dc9a1984ba057c849d492fe8b974/layers/gpu/cmd_validation/gpuav_draw.cpp#L657

There's an unconditional divide by stride, I believe it's to clamp the draw count if he draws read out of bounds. In the event that stride is set to 0 this causes a crash.

**Expected behavior**

When calling an indirect draw with count of 1 stride is not required to be non-zero so this should be ignored.

**Valid Usage ID**
N/A

**Additional context**
Adding a simple `drawCount > 1` to the check worked for me, though I'm assuming there's earlier validation that stride is properly set when requried.

spencer-lunarg:
Thanks for catching this (we literately just merged this change in this week)

We will get some mesh tests (and other things that use`drawCount`) in for this

arno-lunarg:
Thank you for spotting that! Especially since the error spat at me when testing that was awful, and the debugger would not even break ...

spencer-lunarg:
Keeping open until we add tests for `vkCmdDrawIndirect` and `vkCmdDrawIndexedIndirect` to confirm they aren't broken as well

PR #9019, opened 2024-12-13
gpuav: Fix / 0 in draw mesh indirect validation

arno-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9014

### DIFF
--- a/layers/gpu/cmd_validation/gpuav_draw.cpp
+++ b/layers/gpu/cmd_validation/gpuav_draw.cpp
@@ -597,7 +597,7 @@ void DrawMeshIndirect(Validator &gpuav, CommandBuffer &cb_state, const Location 
     const bool is_task_shader = (stages & VK_SHADER_STAGE_TASK_BIT_EXT) == VK_SHADER_STAGE_TASK_BIT_EXT;
 
     CommandBuffer::ValidationCommandFunc validation_cmd =
-        [draw_buffer, draw_buffer_size = draw_buffer_state->create_info.size, draw_buffer_offset, draw_cmds_byte_stride,
+        [draw_buffer, draw_buffer_full_size = draw_buffer_state->create_info.size, draw_buffer_offset, draw_cmds_byte_stride,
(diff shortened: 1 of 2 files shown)


layers/gpu/cmd_validation/gpuav_draw.cpp
```diff
@@ -653,8 +653,13 @@ void DrawMeshIndirect(Validator &gpuav, CommandBuffer &cb_state, const Location
...
 
                 uint32_t max_held_draw_cmds = 0;
-                if (draw_buffer_size > draw_buffer_offset) {
-                    max_held_draw_cmds = static_cast<uint32_t>((draw_buffer_size - draw_buffer_offset) / draw_cmds_byte_stride);
+                if (draw_buffer_full_size > draw_buffer_offset) {
+                    if (draw_cmds_byte_stride > 0) {
+                        max_held_draw_cmds =
+                            static_cast<uint32_t>((draw_buffer_full_size - draw_buffer_offset) / draw_cmds_byte_stride);
+                    } else {
+                        max_held_draw_cmds = 1;
```
artem-lunarg, line 661: nit: quote section from the spec: "If drawCount is less than or equal to one, stride is ignored." ?

arno-lunarg: Yes good point!

tests/unit/gpu_av_indirect_buffer_positive.cpp
```diff
@@ -176,3 +176,67 @@ TEST_F(PositiveGpuAVIndirectBuffer, Mesh) {
     m_default_queue->Submit(m_command_buffer);
     m_default_queue->Wait();
 }
+
+TEST_F(PositiveGpuAVIndirectBuffer, MeshSingleCommand) {
```
spencer-lunarg, line 180: can we add a test for `vkCmdDrawIndirect` and `vkCmdDrawIndexedIndirect` which also can have this draw, you only added a code change to the `DrawMeshIndirect` code

Issue #9018, opened 2024-12-13
Possibly incorrect length check in BDA validation


baldurk:
**Environment:**
 - OS: Windows 11
 - GPU and driver version: AMD 7900 XTX driver 24.10.1
 - SDK or header version if building from repo: Built from b10d21aa975d5cf4ecf4d8d3c7ba4610e151bce6
 - Options enabled (synchronization, best practices, etc.): GPU validation & synchronization

**Describe the Issue**

When running with GPU-AV I encountered an assert inside the BDA SPIR-V pass:

https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/f00d38b79154d717654f1e513460bf4471e46013/layers/gpu/spirv/buffer_device_address_pass.cpp#L80-L84

In this case the OpLoad was 5 words long but then an access to word index [5] was out of bounds. There was no subsequent crash but I'm guessing that was just luck.

**Expected behavior**

There would be no assert when processing shaders.

**Valid Usage ID**
N/A

**Additional context**

I think the problem here is that although OpLoad is normally only 4 words, it can be 5 words if there are memory operands which do not include `Aligned` (since [4] is the memory operands and [5] if any will be the parameters) - in my case this is some generated SPIR-V inside RenderDoc and it always emits the memory operands parameter even if it's `None`, but the same could happen if there was a parameter-less memory operand. The length check should maybe be `<=` to cover this case.

I'm not sure if you want to also add a check that the memory operands have `Aligned` set as well, since although e.g. `MakePointerAvailable` would also have a param the pointer can't validly be BDA if it doesn't include `Aligned` (whose operand would come first). The 'alignment' value loaded in that case would not be used for anything currently, but might be confusing?

spencer-lunarg:
Yep, a simple fix would to make sure the 2nd last operand is `Aligned` actually

Issue #9029, opened 2024-12-16
Add Atomic support for BDA in GPU-AV


spencer-lunarg:
Currently we don't support an atomic load or store from a BDA address

Note that atomic load/stores are naturally aligned (need to find spec text yet)

Examples https://godbolt.org/z/EhvfsvTva

PR #9030, opened 2024-12-16
gpu: Fix BDA not checking if memory operand is Aligned

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9018

we handled something like

`%29 = OpLoad %uint %28 Aligned 4` 

fine, but something like

`%29 = OpLoad %uint %28 MakePointerAvailable|NonPrivatePointer %uint_5` 

was blowing up


layers/gpu/core/gpuav_setup.cpp
```diff
@@ -268,6 +268,29 @@ void Validator::PreCallRecordCreateDevice(VkPhysicalDevice physicalDevice, const
             new_ts_features.timelineSemaphore = VK_TRUE;
             vku::AddToPnext(*modified_create_info, new_ts_features);
         }
+
+        // need vulkanMemoryModelDeviceScope feature to let us call atomicAdd to the output buffer
```
spencer-lunarg, line 272: this was a problem because the new tests was setting `vulkanMemoryModel` to true, but `vulkanMemoryModelDeviceScope` to false

... this is just being consistent with how we add things for `timelineSemaphore`, but would like to clean this up after

(we need a better system to mark which features are "required" or "some features will be turned off" and then have a simpler way to map which Vulkan features map to which GPU-AV features)

tests/unit/gpu_av_buffer_device_address.cpp
```diff
@@ -1494,6 +1494,126 @@ TEST_F(NegativeGpuAVBufferDeviceAddress, NonStructPointer) {
...
+    pipe.descriptor_set_->UpdateDescriptorSets();
+
+    m_command_buffer.Begin();
+    vk::CmdBindPipeline(m_command_buffer.handle(), VK_PIPELINE_BIND_POINT_COMPUTE, pipe.Handle());
+    vk::CmdBindDescriptorSets(m_command_buffer.handle(), VK_PIPELINE_BIND_POINT_COMPUTE, pipe.pipeline_layout_.handle(), 0, 1,
+                              &pipe.descriptor_set_->set_, 0, nullptr);
+    vk::CmdDispatch(m_command_buffer.handle(), 1, 1, 1);
+    m_command_buffer.End();
+
+    m_errorMonitor->SetDesiredError("UNASSIGNED-Device address out of bounds", 2);
```
spencer-lunarg, line 1552: This is being consistent with the other error messages in these tests, still need to go back and update with regex, but still need to add https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9029 as well so can try and do it then

layers/gpu/core/gpuav_setup.cpp
arno-lunarg: You enable those new features without checking for support - are they known to be available if using 1.1 or higher?

arno-lunarg (approved):

layers/gpu/core/gpuav_setup.cpp
spencer-lunarg: yes, `vulkanMemoryModel` is just like `timelineSemaphore` in this caes

Issue #9028, opened 2024-12-14
Invalid settings from VK_EXT_layer_settings are not reported


elasota:
**Environment:**
 - OS: Windows 10
 - GPU and driver version: AMD Radeon RX 6700, driver 24.9.1
 - SDK or header version if building from repo: 1.3.296.0

**Describe the Issue**
If a setting is supplied VK_EXT_layer_settings that is not actually supported by the validation layer, no warning message is generated.  Since VK_EXT_layer_settings itself doesn't supply a reflection API to enumerate settings, there isn't really a way to validate that the settings being passed were correct.

Passing values with the wrong type also results in unusual behavior, e.g. passing "thread_safety" with the type set to VK_LAYER_SETTING_TYPE_INT32_EXT results in the setting being disabled.

**Expected behavior**
Passing an unsupported or incorrectly-typed setting should produce a warning.

**Valid Usage ID**
There does not appear to be a VUID for this.

spencer-lunarg:
This a really good question and opened a lot of dialog

I created this issue https://github.com/KhronosGroup/Vulkan-Utility-Libraries/issues/254 but for the near term, I am going to just hand write it for the Validation Layers (and then swap it out for the VUL version) as I agree 100% you should be warned if you are setting invalid or  non-existent settings

spencer-lunarg:
@elasota so added a fix to at least give you a warning if you provide a setting we don't have or the wrong type

As for a "reflection" it was brought up internally that during the design of `VK_EXT_layer_settings` that came up and the decision was made (which I agree with now) that if you don't know a setting you shouldn't just have code randomly try setting it

elasota:
> As for a "reflection" it was brought up internally that during the design of `VK_EXT_layer_settings` that came up and the decision was made (which I agree with now) that if you don't know a setting you shouldn't just have code randomly try setting it

FWIW one of the cases I'm concerned about in wanting this is the setting not being supported by the SDK/driver version.  In that case, correct usage will still fail, and it would be nice to have a way (similar to layer/extension enumeration) to detect that the requested setting is unavailable and inform the user.

spencer-lunarg:
So the issue becomes we currently don't have a mechanism to expose all the settings

We would need any layer that uses `VK_EXT_layer_settings` to provide a `vkGetLayerSettings` type query

>  I'm concerned about in wanting this is the setting not being supported by the SDK/driver version

[...]

PR #9032, opened 2024-12-16
layers: Warn if VkLayerSettingEXT has wrong type or duplicated

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9028

We ultimately need a more streamline, auto-generated way of doing this type of checks (else adding a setting only gets more complicated again)

For now, at least warn the user if they have the wrong `VkLayerSettingTypeEXT` or duplicated a setting

### DIFF
--- a/layers/layer_options.cpp
+++ b/layers/layer_options.cpp
@@ -509,18 +509,22 @@ static bool ValidateLayerSettingsCreateInfo(const VkLayerSettingsCreateInfoEXT *
 
 // TODO - This should be in https://github.com/KhronosGroup/Vulkan-Utility-Libraries/issues/254
 // Doing here in VVL until added in VUL
-// If it will not be don in VUL for a while, we should at least generate this from the JSON
+// If it will not be done in VUL for a while, we should at least generate this from the JSON
 static void ValidateLayerSettingsProvided(const VkLayerSettingsCreateInfoEXT *layer_setting_create_info,
                                           VkuLayerSettingSet layer_setting_set, std::vector<std::string> &setting_warnings) {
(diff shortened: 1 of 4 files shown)


layers/layer_options.cpp
```diff
@@ -504,6 +507,138 @@ static bool ValidateLayerSettings(const VkLayerSettingsCreateInfoEXT *layer_sett
     return valid;
 }
 
+// TODO - This should be in https://github.com/KhronosGroup/Vulkan-Utility-Libraries/issues/254
+// Doing here in VVL until added in VUL
+// If it will not be don in VUL for a while, we should at least generate this from the JSON
```
filnet, line 512:
```suggestion
// If it will not be done in VUL for a while, we should at least generate this from the JSON
```

layers/layer_options.cpp
```diff
@@ -504,6 +507,138 @@ static bool ValidateLayerSettings(const VkLayerSettingsCreateInfoEXT *layer_sett
...
+static void ValidateLayerSettingsProvided(const VkLayerSettingsCreateInfoEXT *layer_setting_create_info,
+                                          VkuLayerSettingSet layer_setting_set, std::vector<std::string> &setting_warnings) {
+    if (!layer_setting_create_info) return;
+    // set of const char* is not cross platform and Windows doesn't detect duplicates
```
artem-lunarg, line 516: nit: as other points set of const char* is cross platform. Might be more precice to mention string literal handling by the compiler

artem-lunarg (approved):

PR #9036, opened 2024-12-16
tests: Disable message limit in tests

ziga-lunarg:
The message limit should be disabled in the tests by default. Some tests are currently failing when reaching this limit.


tests/framework/layer_validation_tests.cpp
```diff
@@ -320,6 +320,16 @@ VkSamplerCreateInfo SafeSaneSamplerCreateInfo() {
 }
 
 void VkLayerTest::Init(VkPhysicalDeviceFeatures *features, VkPhysicalDeviceFeatures2 *features2, void *instance_pnext) {
+    const VkBool32 enable_message_limit = false;
```
spencer-lunarg, line 323: putting it in here seems like a great way for people to get burned if they go

```
RETURN_IF_SKIP(InitFramework(something_else));
RETURN_IF_SKIP(InitState());
```

I also created a `static VkLayerSettingsCreateInfoEXT kDisableMessageLimit` that can be used here

PR #9038, opened 2024-12-16
tests: Skip test if descriptor set alloc fails

ziga-lunarg:
This test would fail on one of my devices.

The test `PositiveYcbcr.ImageQuerySizeLod` has the same check.


spencer-lunarg (approved):

spencer-lunarg:
checked internal CI passed

PR #9042, opened 2024-12-17
gpu: Fix Chaining of Accesses for General Buffer

spencer-lunarg:
This fixes case when you go

```
%36 = OpAccessChain %23 %35 0
%37 = OpAccessChain %20 %36 1
```

instead of `%36 = OpAccessChain %23 %35 0 1` 

While here, added lots of test to fix checking when copying between Array/Structs

### DIFF
--- a/layers/gpu/spirv/pass.cpp
+++ b/layers/gpu/spirv/pass.cpp
@@ -14,6 +14,7 @@
  */
 
 #include "pass.h"
+#include "instruction.h"
 #include "module.h"
 #include "gpu/shaders/gpuav_error_codes.h"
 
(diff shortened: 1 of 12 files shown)


layers/gpu/spirv/pass.cpp
```diff
@@ -205,22 +209,110 @@ const Instruction* Pass::GetMemeberDecoration(uint32_t id, uint32_t member_index
...
+                const Type* vector_type = module_.type_manager_.FindTypeById(type.inst_.Word(2));
+                return vector_type->inst_.Word(3) * matrix_stride;
+            }
+        }
+        case SpvType::kVector: {
+            uint32_t size = type.inst_.Word(3);
+            const Type* component_type = module_.type_manager_.FindTypeById(type.inst_.Word(2));
+            // if vector in row major matrix, the vector is strided so return the number of bytes spanned by the vector
+            if (in_matrix && !col_major && matrix_stride > 0) {
+                return (size - 1) * matrix_stride + FindTypeByteSize(component_type->Id());
```
artem-lunarg, line 237: Does this mean that the last row is computed as not strided for size computation?

artem-lunarg (approved):

layers/gpu/spirv/pass.cpp
artem-lunarg: Okay, I see this code was just moved, so just curiosity question.

spencer-lunarg: So I still am trying to understand the "right" way to calculate the size of a matrix (when at the end of a struct), we have 2 different ways in this Codebase, currently getting WG to confirm, then will be adding lots of tests lol

PR #9046, opened 2024-12-18
Mark PDB install as optional

charles-lunarg:
Checking if the PDB file exists doesn't work because the PDB file won't exist yet if the build has not occurred yet.


spencer-lunarg (approved):

spencer-lunarg:
fix commit message... feel free to make a new PR if you can't do it easily with Github online editor

artem-lunarg (approved):

PR #9048, opened 2024-12-18
sync: Add more properties to filter syncval errors

artem-lunarg:
Adds properties to filter common types of syncval error messages. Tested on ANGLE's error supression list. Later, after more testing, will provide to ANGLE an example/patch how to update suppression list to use new properties. This will make the project less affected by the changes to error messages.

### DIFF
--- a/layers/sync/sync_error_messages.cpp
+++ b/layers/sync/sync_error_messages.cpp
@@ -27,6 +27,17 @@
 #include <cstdarg>
 #include <sstream>
 
+constexpr const char* kPropertyMessageType = "message_type";
+constexpr const char* kPropertyLoadOp = "load_op";
+constexpr const char* kPropertyStoreOp = "store_op";
+constexpr const char* kPropertyResolveMode = "resolve_mode";
+constexpr const char* kPropertyOldLayout = "old_layout";
+constexpr const char* kPropertyNewLayout = "new_layout";
+constexpr const char* kPropertyResourceParameter = "resource_parameter";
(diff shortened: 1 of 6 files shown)


layers/sync/sync_error_messages.cpp
```diff
@@ -27,6 +27,17 @@
 #include <cstdarg>
 #include <sstream>
 
+constexpr const char* kPropertyMessageType = "message_type";
+constexpr const char* kPropertyLoadOp = "load_op";
```
spencer-lunarg, line 31: more of a ignorant question, why is this not `"LoadOp"` or `"load op"`?

spencer-lunarg (approved):

layers/sync/sync_error_messages.cpp
artem-lunarg: That's optional section of key-values pairs that is not part of the main error message (turned off by default). It's for parsing purposes to filter messages and also for additional debug information.

The keys are without spaces to simplify parsing and to have them visually in a single piece. Camel/underscores I think that's convention/taste aspect here. That's what syncval used so far (`seq_no`, `reset_no`). I feel this style is used sometimes for property names in different tech stacks (in layer configuration `syncval_submit_time_validation`, java lang properties...).

The promise that optional properties section will contain data in structured form. And the main message will move more in the direction of unstructured text where we explain things in english. The goal also to remove syncval internal implementation names from the main message (like `SYNC_INDEX_INPUT_INDEX_READ` and use there only terms/constants from Vulkan specification)!

Issue #9033, opened 2024-12-16
Add maxMemoryAllocationSize check when allocating memory


Vovan675:
# Issue description
I am attempting to allocate a VkImage with dimensions 16,384 x 16,384 and 4 array layers, but the call to `vkAllocateMemory` results in `VK_ERROR_UNKNOWN`. After further investigation, I discovered that memory allocation succeeds (`VK_SUCCESS`) if I do not use `VkMemoryDedicatedAllocateInfoKHR`. However, this seems to be undefined behavior, as the expected result should be `VK_ERROR_OUT_OF_DEVICE_MEMORY`.

The root of the issue appears to be related to the `maxMemoryAllocationSize` limit, which on my GPU is set to 2,147,483,648 (2 GB). The image I am trying to allocate exceeds this size, yet there is no validation layer error triggered. This makes diagnosing the problem challenging.

I believe there is a need to improve validation for vkAllocateMemory to account for the maxMemoryAllocationSize limit to prevent undefined behavior.

# Context
Tested on GPU AMD RX 6700 XT. Enabled VK_LAYER_KHRONOS validation layer

# Reproduce code
```
VkImageCreateInfo info{};
info.sType = VK_STRUCTURE_TYPE_IMAGE_CREATE_INFO;
info.imageType = VK_IMAGE_TYPE_2D;
info.extent.width = 16384;
info.extent.height = 16384;
info.extent.depth = 1;
info.mipLevels = 5;
info.arrayLayers = 4;
info.format = VK_FORMAT_R16G16_SFLOAT;
info.tiling = VK_IMAGE_TILING_OPTIMAL;
info.initialLayout = VK_IMAGE_LAYOUT_UNDEFINED;
info.usage = VK_IMAGE_USAGE_SAMPLED_BIT | VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT;
info.samples = VK_SAMPLE_COUNT_1_BIT;
info.sharingMode = VK_SHARING_MODE_EXCLUSIVE;

// 1. Create image
VkImage image;
VkResult result = vkCreateImage(VkWrapper::device->logicalHandle, &info, nullptr, &image);

// 2. Get memory requirements for the image
VkMemoryRequirements memRequirements;
vkGetImageMemoryRequirements(VkWrapper::device->logicalHandle, image, &memRequirements);

// 3. Allocate memory
VkMemoryAllocateInfo allocInfo = {};
[...]

PR #9052, opened 2024-12-19
layers: Check maxMemoryAllocationSize

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9033

### DIFF
--- a/layers/core_checks/cc_device_memory.cpp
+++ b/layers/core_checks/cc_device_memory.cpp
@@ -758,6 +758,25 @@ bool CoreChecks::PreCallValidateAllocateMemory(VkDevice device, const VkMemoryAl
     return skip;
 }
 
+void CoreChecks::PostCallRecordAllocateMemory(VkDevice device, const VkMemoryAllocateInfo *pAllocateInfo,
+                                              const VkAllocationCallbacks *pAllocator, VkDeviceMemory *pMemory,
+                                              const RecordObject &record_obj) {
+    StateTracker::PostCallRecordAllocateMemory(device, pAllocateInfo, pAllocator, pMemory, record_obj);
+
+    if (VK_SUCCESS != record_obj.result) {
+        // Discussed in https://gitlab.khronos.org/vulkan/vulkan/-/issues/4119
+        // We should help hint the application if this is the reason they fail.
+        // Note, some drivers don't properly report VK_ERROR_OUT_OF_DEVICE_MEMORY so check any error code
+        if (IsExtEnabled(device_extensions.vk_khr_maintenance3) &&
+            pAllocateInfo->allocationSize > phys_dev_props_core11.maxMemoryAllocationSize) {
+            LogWarning("WARNING-CoreValidation-AllocateMemory-maxMemoryAllocationSize", device,
+                       record_obj.location.dot(Field::pAllocateInfo).dot(Field::allocationSize),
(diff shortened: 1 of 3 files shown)


layers/core_checks/cc_device_memory.cpp
```diff
@@ -758,6 +758,25 @@ bool CoreChecks::PreCallValidateAllocateMemory(VkDevice device, const VkMemoryAl
...
+                                              const RecordObject &record_obj) {
+    StateTracker::PostCallRecordAllocateMemory(device, pAllocateInfo, pAllocator, pMemory, record_obj);
+
+    if (VK_SUCCESS != record_obj.result) {
+        // Discussed in https://gitlab.khronos.org/vulkan/vulkan/-/issues/4119
+        // We should help hint the application if this is the reason they fail.
+        // Note, some drivers don't properly report VK_ERROR_OUT_OF_DEVICE_MEMORY so check any error code
+        if (IsExtEnabled(device_extensions.vk_khr_maintenance3) &&
+            pAllocateInfo->allocationSize >= phys_dev_props_core11.maxMemoryAllocationSize) {
+            LogWarning("WARNING-CoreValidation-AllocateMemory-maxMemoryAllocationSize", device,
```
artem-lunarg, line 772: Did not check discussion, but any arguments to use warning intead of error? 
Although no vuid but there is text `maxMemoryAllocationSize is the maximum size of a memory allocation that can be created, even if there is more space available in the heap.`

layers/core_checks/cc_device_memory.cpp
```diff
@@ -758,6 +758,25 @@ bool CoreChecks::PreCallValidateAllocateMemory(VkDevice device, const VkMemoryAl
...
+                                              const VkAllocationCallbacks *pAllocator, VkDeviceMemory *pMemory,
+                                              const RecordObject &record_obj) {
+    StateTracker::PostCallRecordAllocateMemory(device, pAllocateInfo, pAllocator, pMemory, record_obj);
+
+    if (VK_SUCCESS != record_obj.result) {
+        // Discussed in https://gitlab.khronos.org/vulkan/vulkan/-/issues/4119
+        // We should help hint the application if this is the reason they fail.
+        // Note, some drivers don't properly report VK_ERROR_OUT_OF_DEVICE_MEMORY so check any error code
+        if (IsExtEnabled(device_extensions.vk_khr_maintenance3) &&
+            pAllocateInfo->allocationSize >= phys_dev_props_core11.maxMemoryAllocationSize) {
```
artem-lunarg, line 771: Should be strongly greater: `>`

layers/core_checks/cc_device_memory.cpp
spencer-lunarg: we have another spot we do this (see `WARNING-CoreValidation-AllocateDescriptorSets-WrongType`) so we keeping it consistent

There is no VU because the behavior is defined, you will get an error message back (like you will never blowup or device lost)

Added to discussion for next VVL meeting (as want input from Mark/Mike/Charles on this)

layers/core_checks/cc_device_memory.cpp
spencer-lunarg: good catch

artem-lunarg (approved):

Issue #8869, opened 2024-11-17
False `DrawState-InvalidImageLayout` error


knokko:
**Environment:**
 - OS: Windows 10
 - GPU and driver version: NVIDIA GeForce RTX 3070 Ti 561.9.0.0
 - SDK or header version if building from repo: 1.3.296.0
 - Options enabled (synchronization, best practices, etc.): gpu-assisted, synchronization, and best practices

**Describe the Issue**

When my  application draws an image, the validation layer reports a validation error that I believe is incorrect:
```
Validation Error: [ UNASSIGNED-CoreValidation-DrawState-InvalidImageLayout ] Object 0: handle = 0x2243b06c390, name = GameWindowBuffer0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0x2e2cd000000002b, name = Bc1Image0, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x4dae5635 | vkCmdDraw():  command buffer VkCommandBuffer 0x2243b06c390[GameWindowBuffer0] expects VkImage 0x2e2cd000000002b[Bc1Image0] (subresource: aspectMask VK_IMAGE_ASPECT_COLOR_BIT array layer 0, mip level 0) to be in layout VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL--instead, current layout is VK_IMAGE_LAYOUT_UNDEFINED.
```
It states that my image `Bc1Image0` has the current layout `VK_IMAGE_LAYOUT_UNDEFINED`, but I believe that it's layout is `VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL` because the following commands have been executed and completed prior to the draw command submission:
<details>
<summary>the image layout transition commands</summary>

```
Thread 1, Frame 0:
vkCmdPipelineBarrier(commandBuffer, srcStageMask, dstStageMask, dependencyFlags, memoryBarrierCount, pMemoryBarriers, bufferMemoryBarrierCount, pBufferMemoryBarriers, imageMemoryBarrierCount, pImageMemoryBarriers) returns void:
    commandBuffer:                  VkCommandBuffer = 000002243B5C0C60 [UiStagingTransfer]
    srcStageMask:                   VkPipelineStageFlags = 1 (VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT)
[...]

Issue #8688, opened 2024-10-12
False positives in `UNASSIGNED-CoreValidation-DrawState-InvalidImageLayout` when rendering using descriptor indexing


corporateshark:
**Environment:**
 - OS: Windows 10
 - GPU and driver version: NVIDIA 565.90
 - SDK or header version if building from repo: 1.3.296
 - Options enabled (synchronization, best practices, etc.):  Enabled: `GPA/Shader Instrumentation/Descriptor Indexing`

**Describe the Issue**

1) Run [Chapter08/01_DescriptorIndexing](https://github.com/PacktPublishing/3D-Graphics-Rendering-Cookbook-Second-Edition/tree/main/Chapter08/01_DescriptorIndexing)

2) Observe validation errors:
```
Validation layer:
 Validation Error: UNASSIGNED-CoreValidation-DrawState-InvalidImageLayout
 Object 0: handle = 000001DDE46EC3E0, type = VK_OBJECT_TYPE_COMMAND_BUFFER;
 MessageID = 0000000000000000
 vkCmdDraw():  command buffer VkCommandBuffer 0x1dde46ec3e0[] expects
  VkImage 0x7222ad000000040f[Image: deps/src/explosion/explosion1/explosion01-frame015.tga]
  (subresource: aspectMask VK_IMAGE_ASPECT_COLOR_BIT array layer 0, mip level 0)
  to be in layout  VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL -- 
  instead, current layout is VK_IMAGE_LAYOUT_UNDEFINED.
```

**Expected behavior**

Previous SDK versions have no errors. 

**Additional context**

Behavior changes when toggling `GPA/Shader Instrumentation/Descriptor Indexing`.

corporateshark:
@spencer-lunarg Any chance this can be looked into before the next Vulkan SDK release?

spencer-lunarg:
@corporateshark yes, I have just started looking at it late last week, I have a few other (https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8940) issues around it and its just around how we are doing the state tracking divide between CPU and GPU-AV... so should be ready by the January SDK

spencer-lunarg:
ok pinned this down I think (and noting for myself)

[...]

PR #9055, opened 2024-12-19
gpu: Default Image Layout to off

spencer-lunarg:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8688
and 
https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8869

are showing that we still have some false positives around Image Layouts with GPU-AV

For short term, created a setting (which we probably should have had anyway) and made the default off

### DIFF
--- a/tests/unit/layer_settings_positive.cpp
+++ b/tests/unit/layer_settings_positive.cpp
@@ -25,7 +25,7 @@ TEST_F(PositiveLayerSettings, AllSettings) {
     const VkBool32 disable = VK_FALSE;
     const uint32_t one = 1;
     const uint32_t one_k = 1024;
-    std::array<VkLayerSettingEXT, 51> settings = {{
+    std::vector<VkLayerSettingEXT> settings = {{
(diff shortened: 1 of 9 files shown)


arno-lunarg (approved):
It should be mentioned somewhere in the setting description that this part of validation is still unreliable

tests/unit/layer_settings_positive.cpp
```diff
@@ -25,7 +25,7 @@ TEST_F(PositiveLayerSettings, AllSettings) {
     const VkBool32 disable = VK_FALSE;
     const uint32_t one = 1;
     const uint32_t one_k = 1024;
-    std::array<VkLayerSettingEXT, 51> settings = {{
+    std::array<VkLayerSettingEXT, 52> settings = {{
```
arno-lunarg, line 28:
```suggestion
    std::vector<VkLayerSettingEXT> settings = {
```

spencer-lunarg: ya, this is better just a vector at this point since we will keep changing it (and it can't infer the type)

spencer-lunarg:
> It should be mentioned somewhere in the setting description that this part of validation is still unreliable

Yep - will add

PR #9057, opened 2024-12-19
gpu: Add support for multiple access chains

spencer-lunarg:
Adds more Slang shaders that use chains of `OpAccessChains` and finally fixed the various TODO to add support for it


artem-lunarg (approved):
LGTM

PR #9059, opened 2024-12-19
chassis: Clean up handle wrapping and Log method definitions

jeremyg-lunarg:
Add DispatchObject methods for common handle wrapping operations that were being done inline in many places.

Add Logger class to define LogError(), etc. so that the can only be defined once.

### DIFF
--- a/layers/chassis/dispatch_object.h
+++ b/layers/chassis/dispatch_object.h
@@ -173,6 +173,28 @@ class DispatchObject : public Logger {
         return (HandleType)unique_id;
     }
 
+    template <typename HandleType>
+    HandleType Find(HandleType wrapped_handle) const {
+        uint64_t id = CastToUint64(wrapped_handle);
+        auto iter = unique_id_mapping.find(id);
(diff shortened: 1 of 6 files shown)


layers/chassis/dispatch_object.h
```diff
@@ -190,71 +211,6 @@ class DispatchObject {
 
     bool IsSecondary(VkCommandBuffer cb) const;
 
-    // Debug Logging Helpers
-    bool DECORATE_PRINTF(5, 6)
```
spencer-lunarg, line 194: I've wanted to remove this from the chassis code forever lol

spencer-lunarg (approved):

layers/chassis/dispatch_object.h
jeremyg-lunarg: Yeah I wish I didn't have to use inheritance but its the cleanest way to avoid changing 95% of the error messages. But now it is easier to make other code log without being or having a ValidationObject.

PR #9061, opened 2024-12-19
layers: Remove win32 handle on memory free

### LINKED ISSUE #8874: False positive vkAllocateMemory VUID-VkMemoryAllocateInfo-allocationSize-01743
**Environment:**
 - OS: Windows 11 Pro 23H2
 - GPU and driver version: GeForce RTX 3070 / 560.94
 - SDK or header version if building from repo: 1.3.296.0

**Describe the Issue**

When `VkImportMemoryWin32HandleInfoKHR.handle` has a `VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32_BIT` from a second process which matches a previously created and closed handle from the validated process, the validation layer incorrectly reports a mismatch between `VkMemoryAllocateInfo.allocationSize` and the size when the payload was created.

**Expected behavior**

No validation error should be reported.

**Valid Usage ID**

**Additional context**

Process 1: allocates 3360 bytes, calls `vkGetMemoryWin32HandleKHR()` -> `0xdb4`
Process 1: ...
Process 1: `CloseHandle(0xdb4)`, `vkFreeMemory()`
Process 2:  allocates 41760 bytes, calls `vkGetMemoryWin32HandleKHR()` -> `0x1698`
Process 2: `DuplicateHandle(0x1698)` -> `0xdb4`
Process 2: sends `0xdb4` to Process 1
Process 1: `vkAllocateMemory()` with `VkImportMemoryWin32HandleInfoKHR.handle` `0xdb4` triggers false positive 3360 != 41760 but 3360 is the allocation of a previously deleted resource in Process 1

````
ERROR : VALIDATION - Message Id Number: -1224652074 | Message Id Name: VUID-VkMemoryAllocateInfo-allocationSize-01743
	Validation Error: [ VUID-VkMemoryAllocateInfo-allocationSize-01743 ] | MessageID = 0xb7014ad6 | vkAllocateMemory(): pAllocateInfo->allocationSize allocationSize (125280) does not match pAllocateInfo->pNext<VkImportMemoryWin32HandleInfoKHR>.handle (0x15fc) of type VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32_BIT allocationSize (41760).
The Vulkan spec states: If the parameters define an import operation, the external handle was created by the Vulkan API, and the external handle type is VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32_BIT or VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32_KMT_BIT, then the values of allocationSize and memoryTypeIndex must match those specified when the payload being imported was created (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-VkMemoryAllocateInfo-allocationSize-01743)
````

### PR DESCRIPTION
**Pull request by ziga-lunarg**
Closes #8874 

I have a test that reproduced this, but I can't submit it, because it is not reliable. 

```
TEST_F(PositiveExternalMemorySync, Test) {
    TEST_DESCRIPTION("");

    {
        AddRequiredExtensions(VK_KHR_EXTERNAL_MEMORY_WIN32_EXTENSION_NAME);
        RETURN_IF_SKIP(Init());

        const auto handle_type = VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32_BIT;

        VkExternalMemoryBufferCreateInfo external_buffer_info = vku::InitStructHelper();
        external_buffer_info.handleTypes = handle_type;
        auto buffer_info = vkt::Buffer::CreateInfo(4096u, VK_BUFFER_USAGE_TRANSFER_SRC_BIT | VK_BUFFER_USAGE_TRANSFER_DST_BIT);
        buffer_info.pNext = &external_buffer_info;
        vkt::Buffer buffer_export(*m_device, buffer_info, vkt::no_mem);

        auto alloc_info = vkt::DeviceMemory::GetResourceAllocInfo(*m_device, buffer_export.MemoryRequirements(), 0u);

        VkExportMemoryAllocateInfo export_info = {VK_STRUCTURE_TYPE_EXPORT_MEMORY_ALLOCATE_INFO_KHR, nullptr, handle_type};
        alloc_info.pNext = &export_info;

        vkt::DeviceMemory memory_export;
        memory_export.init(*m_device, alloc_info);

        // Bind exported memory
        buffer_export.BindMemory(memory_export, 0);

        VkMemoryGetWin32HandleInfoKHR mghi = vku::InitStructHelper();
        mghi.memory = memory_export.handle();
        mghi.handleType = handle_type;
        HANDLE handle;
        vk::GetMemoryWin32HandleKHR(device(), &mghi, &handle);
        printf("Test1 handle: %p\n", handle);
        prevHandle = handle;
        CloseHandle(handle);
    }
    {
        const auto handle_type = VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32_BIT;

        VkExternalMemoryBufferCreateInfo external_buffer_info = vku::InitStructHelper();
        external_buffer_info.handleTypes = handle_type;
        auto buffer_info = vkt::Buffer::CreateInfo(8192u, VK_BUFFER_USAGE_TRANSFER_SRC_BIT | VK_BUFFER_USAGE_TRANSFER_DST_BIT);
        buffer_info.pNext = &external_buffer_info;
        vkt::Buffer buffer_export(*m_device, buffer_info, vkt::no_mem);

        auto alloc_info = vkt::DeviceMemory::GetResourceAllocInfo(*m_device, buffer_export.MemoryRequirements(), 0u);

        VkExportMemoryAllocateInfo export_info = vku::InitStructHelper();
        export_info.handleTypes = handle_type;
        alloc_info.pNext = &export_info;

        vkt::DeviceMemory memory_export;
        memory_export.init(*m_device, alloc_info);

        // Bind exported memory
        buffer_export.BindMemory(memory_export, 0);

        VkMemoryGetWin32HandleInfoKHR mghi = vku::InitStructHelper();
        mghi.memory = memory_export.handle();
        mghi.handleType = handle_type;
        HANDLE handle;
        vk::GetMemoryWin32HandleKHR(device(), &mghi, &handle);
        std::vector<HANDLE> duplicated_handles;

        HANDLE same_handle;
        if (handle == prevHandle) {
            same_handle = handle;
        } else {
            while (true) {
                HANDLE duplicated;
                DuplicateHandle(GetCurrentProcess(),   // Source process handle
                                handle,                // Handle to duplicate
                                GetCurrentProcess(),   // Target process handle (same process here)
                                &duplicated,           // Pointer to the new handle
                                0,                     // Desired access (0 means same as source handle)
                                FALSE,                 // Handle is not inheritable
                                DUPLICATE_SAME_ACCESS  // Options (use same access rights as source handle)
                );

                duplicated_handles.push_back(duplicated);
                if (duplicated == prevHandle) {
                    same_handle = duplicated;
                    break;
                }
                if (duplicated_handles.size() > 20) {
                    for (const auto& duplicated_handle : duplicated_handles) {
                        CloseHandle(duplicated_handle);
                    }
                    duplicated_handles.clear();
                }
            }
        }
        printf("Test2 handle: %p\n", same_handle);

        {
            auto import_buffer_info =
                vkt::Buffer::CreateInfo(16386u, VK_BUFFER_USAGE_TRANSFER_SRC_BIT | VK_BUFFER_USAGE_TRANSFER_DST_BIT);
            import_buffer_info.pNext = &external_buffer_info;
            vkt::Buffer buffer_import(*m_device, import_buffer_info, vkt::no_mem);

            auto import_alloc_info = vkt::DeviceMemory::GetResourceAllocInfo(*m_device, buffer_import.MemoryRequirements(), 0u);

            VkImportMemoryWin32HandleInfoKHR import_info = vku::InitStructHelper();
            import_info.handleType = handle_type;
            import_info.handle = same_handle;
            import_alloc_info.pNext = &import_info;

            vkt::DeviceMemory memory_import;
            memory_import.init(*m_device, import_alloc_info);

            // Bind exported memory
            buffer_import.BindMemory(memory_import, 0);
        }

        for (const auto& duplicated_handle : duplicated_handles) {
            CloseHandle(duplicated_handle);
        }
        CloseHandle(handle);
    }
}
```

### DIFF
--- a/layers/state_tracker/state_tracker.cpp
+++ b/layers/state_tracker/state_tracker.cpp
@@ -1444,6 +1444,30 @@ void ValidationStateTracker::PreCallRecordFreeMemory(VkDevice device, VkDeviceMe
     if (auto mem_info = Get<vvl::DeviceMemory>(mem)) {
         fake_memory.Free(mem_info->fake_base_address);
     }
+    {
+        WriteLockGuard guard(fd_handle_map_lock_);
+        for (auto it = fd_handle_map_.begin(); it != fd_handle_map_.end();) {
+            if (it->second.device_memory == mem) {
+                it = fd_handle_map_.erase(it);
+                break;
+            } else {
+                ++it;
+            }
+        }
+    }
+#ifdef VK_USE_PLATFORM_WIN32_KHR
+    {
+        WriteLockGuard guard(win32_handle_map_lock_);
+        for (auto it = win32_handle_map_.begin(); it != win32_handle_map_.end();) {
+            if (it->second.device_memory == mem) {
+                it = win32_handle_map_.erase(it);
+                break;
+            } else {
+                ++it;
+            }
+        }
+    }
+#endif
     Destroy<vvl::DeviceMemory>(mem);
 }
 
@@ -3788,6 +3812,7 @@ void ValidationStateTracker::PostCallRecordGetMemoryWin32HandleKHR(VkDevice devi
         external_info.memory_type_index = memory_state->allocate_info.memoryTypeIndex;
         external_info.dedicated_buffer = memory_state->GetDedicatedBuffer();
         external_info.dedicated_image = memory_state->GetDedicatedImage();
+        external_info.device_memory = pGetWin32HandleInfo->memory;
 
         WriteLockGuard guard(win32_handle_map_lock_);
         // `insert_or_assign` ensures that information is updated when the system decides to re-use
@@ -3809,6 +3834,7 @@ void ValidationStateTracker::PostCallRecordGetMemoryFdKHR(VkDevice device, const
         external_info.memory_type_index = memory_state->allocate_info.memoryTypeIndex;
         external_info.dedicated_buffer = memory_state->GetDedicatedBuffer();
         external_info.dedicated_image = memory_state->GetDedicatedImage();
+        external_info.device_memory = memory_state->VkHandle();
 
         WriteLockGuard guard(fd_handle_map_lock_);
         // `insert_or_assign` ensures that information is updated when the system decides to re-use
--- a/layers/state_tracker/state_tracker.h
+++ b/layers/state_tracker/state_tracker.h
@@ -1735,6 +1735,7 @@ class ValidationStateTracker : public ValidationObject {
         uint32_t memory_type_index;
         VkBuffer dedicated_buffer;
         VkImage dedicated_image;
+        VkDeviceMemory device_memory;
 
         // External Semaphore
         VkSemaphoreCreateFlags semaphore_flags = 0;

### CODE REVIEW

**Comment by artem-lunarg** [comment]
We have this comment:
```
        // `insert_or_assign` ensures that information is updated when the system decides to re-use
        // closed handle value for a new handle. The validation layer does not track handle close operation
        // which is performed by 'CloseHandle' system call.
        win32_handle_map_.insert_or_assign(*pHandle, external_info);
```
@ziga-lunarg Do we have a situation when the program at first *exports* HANDLE and frees memory (the handles entry stay in the map) and the tries to *import* the HANDLE with the same value from different app? That looks like a possible scenario for the error, sorry did not go through the tests details!

**Comment by artem-lunarg** [comment]
Well I guess that that situation the described in the issue.

**Review by artem-lunarg — APPROVED** [review]

**Comment by ziga-lunarg** [comment]
> We have this comment:
> 
> ```
>         // `insert_or_assign` ensures that information is updated when the system decides to re-use
>         // closed handle value for a new handle. The validation layer does not track handle close operation
>         // which is performed by 'CloseHandle' system call.
>         win32_handle_map_.insert_or_assign(*pHandle, external_info);
> ```
> 
> @ziga-lunarg Do we have a situation when the program at first _exports_ HANDLE, then frees memory, and then during _import_ gets the HANDLE with the same value from different app? That looks like a possible scenario for this error, sorry did not go through the tests details!

What that comment doesn't consider is `DuplicateHandle()`, that is another function that we are not able to track and which is where this issue comes from. 

What you described is what I think my test does.

The only possible false positive I see is if VkDeviceMemory is created, a handle is retrieved from it. Then the handle is destroyed. Then a different VkDeviceMemory is created and a handle for it is retrieved. Then `DuplicateHandle()` is called which returns the same handle as the first one was, and this one is used to import the memory. But for this case I don't think there is anything we can do.

File: layers/state_tracker/state_tracker.cpp [hunk 1]
```diff
@@ -1444,6 +1444,16 @@ void ValidationStateTracker::PreCallRecordFreeMemory(VkDevice device, VkDeviceMe
     if (auto mem_info = Get<vvl::DeviceMemory>(mem)) {
         fake_memory.Free(mem_info->fake_base_address);
     }
+#ifdef VK_USE_PLATFORM_WIN32_KHR
+    for (auto it = win32_handle_map_.begin(); it != win32_handle_map_.end();) {
```
spencer-lunarg, line 1448: throw a comment to either to the issue explaining why this is... because I will see this in the future and think "wait, its only on windows, not FD Handles??"

spencer-lunarg (approved):

layers/state_tracker/state_tracker.cpp
artem-lunarg: @ziga-lunarg In the scenario from my above comment https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9061#issuecomment-2555572998 I did not use option of handle duplication and considered a use case that OS can re-use HANDLE after it is closed. So you have handle 0x42, close it, then potentially OS can return 0x42. I don't know if different OSes guarantees not to do this. It looks for me that if this happens it can be the same problem for FD?

ziga-lunarg: I am not as familiar with FD to be able to answer this. But I think we can remove the handles just in case?

ziga-lunarg: The only issue here is with DuplicateHandle(), because VVL cannot track that. If the same handle is retrieve through `vkGetMemoryWin32HandleKHR`, or the FD version, then VVL will know about it and overwrite the old info.

artem-lunarg: Maybe I misunderstood some part, but lets sure we are on the same page. I have impression you described the same app exports the same handle two times. My scenario is when it at first exports handle, then imports the same handle. Sorry if I'm really confused here.

So, this scenario:

* App A: **Exports** handle. This puts handle in the map.
* App B. Imports handle.
* App A: Deletes resource, App B also stoped using it, Handle is completly released on the OS level, no app is using it. App A still has that handle in the map.
* App A. **Imports** handle from App X. Gets the same handle that it already has in the map and it raises the error.

artem-lunarg: To be clear, no questions for  the correctness of this PR. Just discussion of possible scenarios.

layers/state_tracker/state_tracker.cpp
```diff
@@ -1444,6 +1444,30 @@ void ValidationStateTracker::PreCallRecordFreeMemory(VkDevice device, VkDeviceMe
     if (auto mem_info = Get<vvl::DeviceMemory>(mem)) {
         fake_memory.Free(mem_info->fake_base_address);
     }
+    {
+        WriteLockGuard guard(fd_handle_map_lock_);
+        for (auto it = fd_handle_map_.begin(); it != fd_handle_map_.end();) {
```
artem-lunarg, line 1449: nit: since we break after `erase` it should be possible to use classic loop form:
```
for (auto it = fd_handle_map_.begin(); it != fd_handle_map_.end(); ++it) {
        if (it->second.device_memory == mem) {
            fd_handle_map_.erase(it);
            break;
        }
}
```

artem-lunarg (approved):

layers/state_tracker/state_tracker.cpp
ziga-lunarg: I find this better than iterating with an index. I can't use foreach loops and erase inside of it, I believe that is UB.

PR #9063, opened 2024-12-19
Fix enumerating performance queries

### LINKED ISSUE #5894: State tracking and validation for performance query counters is incorrect
**Environment:**
 - OS: All
 - GPU: All
 - SDK or header version if building from repo: latest main (8f4fc4c98acf7a6a2c77524a389969a651b7a8c9)
 - Options enabled (synchronization, best practices, etc.): none

**Describe the Issue**

The set of valid performance counters (introduced by the `VK_KHR_performance_query` extension) is incorrectly tracked by the state tracker of VVL.

Currently, what it does is that it records the set of performance query counters in `RecordEnumeratePhysicalDeviceQueueFamilyPerformanceQueryCounters` when any of the corresponding enumerate functions are called by the application, but these will only ever contain the query counters the application actually queried, and won't necessarily contain all the query counters supported for the queue family by the implementation.

This will manifest in false positives of the following VUIDs:

* `VUID-VkQueryPoolPerformanceCreateInfoKHR-queueFamilyIndex-03236` - if using any (otherwise valid) queue family index, for which no performance query counter enumeration call was issued, or for which such enumeration was only called to determine the query counter count, but not actually retrieve any counters
* `VUID-VkQueryPoolPerformanceCreateInfoKHR-pCounterIndices-03321` - if using any query counter for which the application didn't explicitly enumerate its properties

**Expected behavior**

The VUIDs above should not be triggered when otherwise the corresponding valid usage criteria are met, no matter whether the application enumerated the available counters or not, as that's not a requirement.

If anything, a best practices warning could be generated if trying to use a performance query counter that wasn't queried before, but the false positive VUIDs are unrelated to that.

In order to fix this issue VVL should enumerate the available performance query counters for each queue family itself (either up-front, or when the information is first needed to check the VUIDs above).

Also, for VUID 03236 specifically, the check should validate the queue family index itself, indifferent of the performance query counters available, as a queue family index is not invalid just because the queue family does not support any performance query counters.

We also need to make sure that these VUIDs are covered with tests which don't ever call the performance query counter enumeration functions, as currently all tests do that which masked the underlying issue.

**Valid Usage ID**

* `VUID-VkQueryPoolPerformanceCreateInfoKHR-queueFamilyIndex-03236`
* `VUID-VkQueryPoolPerformanceCreateInfoKHR-pCounterIndices-03321`

**Additional context**

See above.

ziga-lunarg:
Closes #5894

### DIFF
--- a/layers/state_tracker/state_tracker.cpp
+++ b/layers/state_tracker/state_tracker.cpp
@@ -1220,6 +1220,23 @@ void ValidationStateTracker::PostCreateDevice(const VkDeviceCreateInfo *pCreateI
         DispatchGetPhysicalDeviceQueueFamilyProperties2Helper(physical_device, &queue_family_count, props.data());
     }
 
+    if (IsExtEnabled(dev_ext.vk_khr_performance_query)) {
+        uint32_t queue_family_count = (uint32_t)physical_device_state->queue_family_properties.size();
+        for (uint32_t i = 0; i < queue_family_count; ++i) {
+            uint32_t counterCount;
+            DispatchEnumeratePhysicalDeviceQueueFamilyPerformanceQueryCountersKHR(physical_device, i, &counterCount, nullptr,
+                                                                                  nullptr);
+
+            std::unique_ptr<QueueFamilyPerfCounters> queue_family_counters(new QueueFamilyPerfCounters());
+            queue_family_counters->counters.resize(counterCount);
+
+            DispatchEnumeratePhysicalDeviceQueueFamilyPerformanceQueryCountersKHR(physical_device, i, &counterCount,
+                                                                                  queue_family_counters->counters.data(), nullptr);
+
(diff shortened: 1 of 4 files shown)


layers/state_tracker/state_tracker.cpp
```diff
@@ -1220,6 +1220,23 @@ void ValidationStateTracker::PostCreateDevice(const VkDeviceCreateInfo *pCreateI
         DispatchGetPhysicalDeviceQueueFamilyProperties2Helper(physical_device, &queue_family_count, props.data());
     }
 
+    if (IsExtEnabled(dev_ext.vk_khr_performance_query)) {
```
spencer-lunarg, line 1223: I didn't realize until now that there are only 3 `vkEnumeratePhysicalDevice*` calls and ya, this makes sense to just be done once for the app at device creation time

spencer-lunarg (approved):

PR #9067, opened 2024-12-20
gpuav: Fix selection of instrumentation desc set pipeline layout

arno-lunarg:
Added a test that showed issues with the selection of the pipeline layout used to bind the instrumentation descriptor set. Now, when trying to get pipeline layout from last bound pipeline, if it has been destroyed it is temporarily re-created and used to bind instrumentation descriptor set.
Trying to get a pipeline layout from another source (for the new test, pipeline layout specified in last vkCmdPushConstants call) lead to the dreaded vkCmdDraw-None-08600


tests/unit/gpu_av_buffer_device_address_positive.cpp
```diff
@@ -1642,4 +1642,102 @@ TEST_F(PositiveGpuAVBufferDeviceAddress, PieceOfDataPointerInStruct) {
     auto out_buffer_ptr = static_cast<uint32_t *>(out_buffer.Memory().Map());
     ASSERT_TRUE(out_buffer_ptr[0] == 33);
     out_buffer.Memory().Unmap();
-}
\ No newline at end of file
+}
+
+TEST_F(PositiveGpuAVBufferDeviceAddress, SharedPipelineLayoutSubsetGraphicsPushConstants) {
```
arno-lunarg, line 1647: Goal of moving this test here was to have shaders actually be instrumented to provoke the "pipeline layout incompatibility" self validation error

arno-lunarg:
I forgot about `CreateInstrumentationPipelineLayout`

PR #9072, opened 2024-12-20
gpuav: Fix selection of instrumentation desc set pipeline layout

arno-lunarg:
Added tests that showed issues with the selection of the pipeline layout used to bind the instrumentation descriptor set. Now, when trying to get pipeline layout from last bound pipeline, if it has been destroyed it is temporarily re-created and used to bind instrumentation descriptor set.
Trying to get a pipeline layout from another source (for the new test, pipeline layout specified in last vkCmdPushConstants call) lead to the dreaded vkCmdDraw-None-08600

### DIFF
--- a/layers/gpu/instrumentation/gpuav_instrumentation.cpp
+++ b/layers/gpu/instrumentation/gpuav_instrumentation.cpp
@@ -31,14 +31,31 @@
 namespace gpuav {
 
 // If application is using shader objects, bindings count will be computed from bound shaders
-static uint32_t GetDescSetBindingsCountFromLastBoundPipelineOrShader(Validator &gpuav, VkPipelineBindPoint bind_point,
-                                                                     CommandBuffer &cb_state, const LastBound &last_bound) {
+static uint32_t LastBoundPipelineOrShaderDescSetBindingsCount(Validator &gpuav, VkPipelineBindPoint bind_point,
+                                                              CommandBuffer &cb_state, const LastBound &last_bound) {
     if (last_bound.pipeline_state && last_bound.pipeline_state->PreRasterPipelineLayoutState()) {
         return static_cast<uint32_t>(last_bound.pipeline_state->PreRasterPipelineLayoutState()->set_layouts.size());
     }
 
     if (const vvl::ShaderObject *main_bound_shader = last_bound.GetFirstShader(bind_point)) {
         return static_cast<uint32_t>(main_bound_shader->set_layouts.size());
+    }
+
+    // Should not get there, it would mean no pipeline nor shader object was bound
+    assert(false);
+    return 0;
+}
+
+// If application is using shader objects, bindings count will be computed from bound shaders
+static uint32_t LastBoundPipelineOrShaderPushConstantsRangesCount(Validator &gpuav, VkPipelineBindPoint bind_point,
+                                                                  CommandBuffer &cb_state, const LastBound &last_bound) {
+    if (last_bound.pipeline_state && last_bound.pipeline_state->PreRasterPipelineLayoutState()) {
+        return static_cast<uint32_t>(
+            last_bound.pipeline_state->PreRasterPipelineLayoutState()->push_constant_ranges_layout->size());
+    }
+
+    if (const vvl::ShaderObject *main_bound_shader = last_bound.GetFirstShader(bind_point)) {
+        return static_cast<uint32_t>(main_bound_shader->push_constant_ranges->size());
     }
 
     // Should not get there, it would mean no pipeline nor shader object was bound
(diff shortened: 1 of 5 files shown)


layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -340,62 +357,66 @@ void PreCallSetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer
...
+                    GetDescSetBindingsCountFromLastBoundPipelineOrShader(gpuav, bind_point, cb_state, last_bound);
+                const uint32_t push_constants_ranges_count_from_last_pipeline =
+                    GetPushConstantsRangesCountFromLastBoundPipelineOrShader(gpuav, bind_point, cb_state, last_bound);
+
+                // If the number of binding of the currently bound pipeline's layout (or the equivalent for shader objects) is
+                // less that the number of bindings in the pipeline layout used to bind descriptor sets,
+                // GPU-AV needs to create a temporary pipeline layout matching the the currently bound pipeline's layout
+                // to bind the instrumentation descriptor set
+                if (bindings_count_from_last_pipeline < (uint32_t)inst_binding_pipe_layout_state->set_layouts.size() ||
+                    push_constants_ranges_count_from_last_pipeline <
```
arno-lunarg, line 396: This is the only addition. Rest is re-organisation around early exit

layers/gpu/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -340,62 +357,66 @@ void PreCallSetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer
...
+                // To alleviate that, since we could not get a pipeline layout from last pipeline binding (it was either
+                // destroyed, or never has been created if using shader objects), a pipeline layout matching bindings of last
+                // bound pipeline or
+                // last bound shader objects is created and used.
+                // If will also be cached: heuristic is next action command will likely need the same.
+
+                const uint32_t bindings_count_from_last_pipeline =
+                    GetDescSetBindingsCountFromLastBoundPipelineOrShader(gpuav, bind_point, cb_state, last_bound);
+                const uint32_t push_constants_ranges_count_from_last_pipeline =
+                    GetPushConstantsRangesCountFromLastBoundPipelineOrShader(gpuav, bind_point, cb_state, last_bound);
```
spencer-lunarg, line 389: these names are getting long and hard to parse IMO (for now in this PR it is fine)

We have a `class LastBound` in CoreChecks and I would much rather have a `LastBound last_bound` object than every other variable/function being called `from_last_pipeline` when really it is `LastBoundPiplineOrShader` and we are never going to enforce this consistently... so having a `last_bound.push_constants_ranges_count` would be both easier to parse, and also be consistent with how we do it in CoreChecks

tests/unit/gpu_av_descriptor_indexing_positive.cpp
```diff
@@ -1438,95 +1437,6 @@ TEST_F(PositiveGpuAVDescriptorIndexing, SharedPipelineLayoutSubsetGraphicsShader
     m_default_queue->Wait();
 }
 
-TEST_F(PositiveGpuAVDescriptorIndexing, SharedPipelineLayoutSubsetGraphicsPushConstants) {
```
spencer-lunarg, line 1441: why did we remove this test?

layers/state_tracker/pipeline_layout_state.cpp
```diff
@@ -68,14 +68,36 @@ std::string PipelineLayoutCompatDef::DescribeDifference(const PipelineLayoutComp
...
         ss << "The set " << set << " is different from the non-compatible pipeline layout (" << other.set << ")\n";
     } else if (push_constant_ranges != other.push_constant_ranges) {
-        ss << "The set push constant ranges is different from the non-compatible pipeline layout push constant ranges\n";
+        ss << "Pipeline layout pipeline bound with last call to vkCmdBindDescriptorSets has following push constant ranges:\n";
+        if (push_constant_ranges->empty()) {
+            ss << "Empty\n";
+        } else {
+            for (const auto &[pcr_i, pcr] : vvl::enumerate(push_constant_ranges->data(), push_constant_ranges->size())) {
+                ss << "VkPushConstantRange[ " << pcr_i << " ]:\n";
+                ss << "- stageFlags: " << string_VkShaderStageFlags(pcr->stageFlags) << '\n';
```
spencer-lunarg, line 77: I added a `string_VkPushConstantRange` in `error_strings.h` it would be nice to try and have a single spot to print things like this out (so they all stay the same)... for now happy to leave this as what you have, but would be much nicer to fix that function up so it all was the same

layers/gpu/instrumentation/gpuav_instrumentation.cpp
arno-lunarg: I don't want to pollute `LastBound` with something only used here, but I can shorten names, quite ok given scope of usage

layers/state_tracker/pipeline_layout_state.cpp
arno-lunarg: I can use that yeah!

layers/gpu/instrumentation/gpuav_instrumentation.cpp
spencer-lunarg: We would have our own for GPU-AV was my thought... or like other things like `CommandBuffer` state object, can make it extend

spencer-lunarg (approved):

PR #9085, opened 2024-12-22
layers: Validate surface counters

### LINKED ISSUE #9082: Add missing VK_EXT_display_control VU
`VUID-VkSwapchainCounterCreateInfoEXT-surfaceCounters-01244` is missing a check

ziga-lunarg:
Closes #9082

### DIFF
--- a/layers/core_checks/cc_wsi.cpp
+++ b/layers/core_checks/cc_wsi.cpp
@@ -694,6 +694,29 @@ bool CoreChecks::ValidateCreateSwapchain(const VkSwapchainCreateInfoKHR &create_
                          "contains VkImageCompressionControlEXT, but imageCompressionControlSwapchain is not enabled");
     }
 
+    const auto *swapchain_counter = vku::FindStructInPNextChain<VkSwapchainCounterCreateInfoEXT>(create_info.pNext);
+    if (swapchain_counter) {
+        VkSurfaceCapabilities2EXT surface_capabilities = vku::InitStructHelper();
+        const VkResult result =
+            DispatchGetPhysicalDeviceSurfaceCapabilities2EXT(physical_device, create_info.surface, &surface_capabilities);
+        if (result != VK_SUCCESS) {
+            skip |= LogError(
+                "VUID-VkSwapchainCounterCreateInfoEXT-surfaceCounters-01244", device,
+                create_info_loc.pNext(Struct::VkSwapchainPresentModesCreateInfoEXT, Field::surfaceCounters),
+                "is %s, but the counters are not supported because the vkGetPhysicalDeviceSurfaceCapabilities2EXT query failed",
+                string_VkSurfaceCounterFlagsEXT(swapchain_counter->surfaceCounters).c_str());
+        } else {
+            if ((swapchain_counter->surfaceCounters & surface_capabilities.supportedSurfaceCounters) !=
+                swapchain_counter->surfaceCounters) {
+                skip |= LogError("VUID-VkSwapchainCounterCreateInfoEXT-surfaceCounters-01244", device,
+                                 create_info_loc.pNext(Struct::VkSwapchainPresentModesCreateInfoEXT, Field::surfaceCounters),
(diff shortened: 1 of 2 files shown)


layers/core_checks/cc_wsi.cpp
```diff
@@ -694,6 +694,20 @@ bool CoreChecks::ValidateCreateSwapchain(const VkSwapchainCreateInfoKHR &create_
...
 
+    const auto *swapchain_counter = vku::FindStructInPNextChain<VkSwapchainCounterCreateInfoEXT>(create_info.pNext);
+    if (swapchain_counter) {
+        VkSurfaceCapabilities2EXT surface_capabilities = vku::InitStructHelper();
+        DispatchGetPhysicalDeviceSurfaceCapabilities2EXT(physical_device, create_info.surface, &surface_capabilities);
+        if ((swapchain_counter->surfaceCounters & surface_capabilities.supportedSurfaceCounters) !=
+            swapchain_counter->surfaceCounters) {
+            skip |= LogError("VUID-VkSwapchainCounterCreateInfoEXT-surfaceCounters-01244", device,
+                             create_info_loc.dot(Struct::VkSwapchainPresentModesCreateInfoEXT, Field::surfaceCounters),
+                             "is %s, but physical device supports only %s",
```
spencer-lunarg, line 705: Maybe say something like

```suggestion
                             "is %s, but calling vkGetPhysicalDeviceSurfaceCapabilities2EXT shows only %s is supported",
```

layers/core_checks/cc_wsi.cpp
```diff
@@ -694,6 +694,20 @@ bool CoreChecks::ValidateCreateSwapchain(const VkSwapchainCreateInfoKHR &create_
...
     }
 
+    const auto *swapchain_counter = vku::FindStructInPNextChain<VkSwapchainCounterCreateInfoEXT>(create_info.pNext);
+    if (swapchain_counter) {
+        VkSurfaceCapabilities2EXT surface_capabilities = vku::InitStructHelper();
+        DispatchGetPhysicalDeviceSurfaceCapabilities2EXT(physical_device, create_info.surface, &surface_capabilities);
+        if ((swapchain_counter->surfaceCounters & surface_capabilities.supportedSurfaceCounters) !=
+            swapchain_counter->surfaceCounters) {
+            skip |= LogError("VUID-VkSwapchainCounterCreateInfoEXT-surfaceCounters-01244", device,
+                             create_info_loc.dot(Struct::VkSwapchainPresentModesCreateInfoEXT, Field::surfaceCounters),
```
spencer-lunarg, line 704:
```suggestion
                             create_info_loc.pNext(Struct::VkSwapchainPresentModesCreateInfoEXT, Field::surfaceCounters),
```

layers/core_checks/cc_wsi.cpp
```diff
@@ -694,6 +694,20 @@ bool CoreChecks::ValidateCreateSwapchain(const VkSwapchainCreateInfoKHR &create_
                          "contains VkImageCompressionControlEXT, but imageCompressionControlSwapchain is not enabled");
     }
 
+    const auto *swapchain_counter = vku::FindStructInPNextChain<VkSwapchainCounterCreateInfoEXT>(create_info.pNext);
+    if (swapchain_counter) {
+        VkSurfaceCapabilities2EXT surface_capabilities = vku::InitStructHelper();
+        DispatchGetPhysicalDeviceSurfaceCapabilities2EXT(physical_device, create_info.surface, &surface_capabilities);
```
spencer-lunarg, line 700: we should check if the result is not `VK_SUCCESS` (in case they provide something strange) and report another `01244` error basically saying "the counters are not supported because the query failed"

ziga-lunarg: I did think of that, but the only thing that can be invalid here is the surface and `VUID-VkSwapchainCreateInfoKHR-surface-parameter` will catch that

spencer-lunarg: ok, well thanks for doing it.. there is a long "todo" to have a better way to return things if we get a non-success call in our internal `Dispatch`

spencer-lunarg (approved):

PR #9088, opened 2024-12-24
Add robustness supported VUIDs

### LINKED ISSUE #5657: VUID-VkPipelineRobustnessCreateInfoEXT-robustImageAccess2-06934 not check correctly
`VUID-VkPipelineRobustnessCreateInfoEXT-robustImageAccess2-06934` and other similar VUIDs are not checked correctly, the VUIDs says that the feature must be supported, but validation layers check if it's enabled.

ziga-lunarg:
Closes #5657

### DIFF
--- a/layers/stateless/sl_instance_device.cpp
+++ b/layers/stateless/sl_instance_device.cpp
@@ -246,8 +246,15 @@ void StatelessValidation::CommonPostCallRecordEnumeratePhysicalDevice(const VkPh
                     discard_rectangles_extension_version = ext_props[j].specVersion;
                 } else if (extension == vvl::Extension::_VK_NV_scissor_exclusive) {
                     scissor_exclusive_extension_version = ext_props[j].specVersion;
+                } else if (extension == vvl::Extension::_VK_EXT_robustness2 &&
+                           IsExtEnabled(instance_extensions.vk_khr_get_physical_device_properties2)) {
+                    VkPhysicalDeviceRobustness2FeaturesEXT robustness_2_features = vku::InitStructHelper();
+                    VkPhysicalDeviceFeatures2 features2 = vku::InitStructHelper(&robustness_2_features);
+                    DispatchGetPhysicalDeviceFeatures2Helper(phys_device, &features2);
+                    device_supported_robustness_2_features[phys_device] = robustness_2_features;
                 }
             }
+
             device_extensions_enumerated[phys_device] = std::move(dev_exts_enumerated);
         }
     }
--- a/layers/stateless/stateless_validation.h
+++ b/layers/stateless/stateless_validation.h
@@ -34,13 +34,17 @@ class StatelessValidation : public ValidationObject {
 
   public:
     StatelessValidation(vvl::dispatch::Device *dev, StatelessValidation *instance_vo)
-        : BaseClass(dev, LayerObjectTypeParameterValidation) {}
+        : BaseClass(dev, LayerObjectTypeParameterValidation),
+          device_supported_robustness_2_features(instance_vo->device_supported_robustness_2_features) {}
     StatelessValidation(vvl::dispatch::Instance *inst) : BaseClass(inst, LayerObjectTypeParameterValidation) {}
     ~StatelessValidation() {}
 
     VkPhysicalDeviceLimits device_limits = {};
     vvl::unordered_map<VkPhysicalDevice, VkPhysicalDeviceProperties *> physical_device_properties_map;
     vvl::unordered_map<VkPhysicalDevice, vvl::unordered_set<vvl::Extension>> device_extensions_enumerated{};
+    // Validation requires special handling for VkPhysicalDeviceRobustness2FeaturesEXT, because for some cases robustness features
+    // need to only be supported, not enabled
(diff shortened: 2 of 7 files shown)


layers/stateless/sl_instance_device.cpp
```diff
@@ -258,7 +258,16 @@ void StatelessValidation::CommonPostCallRecordEnumeratePhysicalDevice(const VkPh
                     scissor_exclusive_extension_version = ext_props[j].specVersion;
                 }
             }
+
+            VkPhysicalDeviceRobustness2FeaturesEXT robustness_2_features = vku::InitStructHelper();
+            auto enum_iter = dev_exts_enumerated.find(vvl::Extension::_VK_EXT_robustness2);
```
spencer-lunarg, line 263: instead of finding here, maybe just after the `if (extension == vvl::Extension::_VK_EXT_discard_rectangles) {` do a `else if (extension == vvl::Extension::_VK_EXT_robustness2) {` and do the dispatch in there

(mainly so we only loop the extensions in a single location)

layers/stateless/stateless_validation.h
```diff
@@ -35,6 +35,7 @@ class StatelessValidation : public ValidationObject {
     VkPhysicalDeviceLimits device_limits = {};
     vvl::unordered_map<VkPhysicalDevice, VkPhysicalDeviceProperties *> physical_device_properties_map;
     vvl::unordered_map<VkPhysicalDevice, vvl::unordered_set<vvl::Extension>> device_extensions_enumerated{};
+    vvl::unordered_map<VkPhysicalDevice, VkPhysicalDeviceRobustness2FeaturesEXT> device_supported_robustness_2_features{};
```
spencer-lunarg, line 38: add a comment here explaining how this is a special case (so people don't think this is an accident this is here in the future

layers/vulkan/generated/dispatch_object.cpp
```diff
@@ -71,7 +71,8 @@ void DispatchObject::InitDeviceValidationObjects(DispatchObject* instance_dispat
             new ThreadSafety(static_cast<ThreadSafety*>(instance_dispatch->GetValidationObject(LayerObjectTypeThreading))));
     }
     if (!disabled[stateless_checks]) {
-        object_dispatch.emplace_back(new StatelessValidation);
+        object_dispatch.emplace_back(new StatelessValidation(
+            static_cast<StatelessValidation*>(instance_dispatch->GetValidationObject(LayerObjectTypeParameterValidation))));
```
spencer-lunarg, line 75: want @jeremyg-lunarg opinion on this as a way to pass the instance level gather data to the device level

I feel there has to be a better way of communicating this

spencer-lunarg: ok, going to wait until https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9096 lands first, that really changes how this will all work

spencer-lunarg: (the rest of the PR looks good btw!)

jeremyg-lunarg: yeah my PR will do this for all the validation objects

spencer-lunarg: @ziga-lunarg you can rebase this now and should be good to go

layers/stateless/stateless_validation.h
```diff
@@ -34,13 +34,17 @@ class StatelessValidation : public ValidationObject {
...
+          device_supported_robustness_2_features(instance_vo->device_supported_robustness_2_features) {}
     StatelessValidation(vvl::dispatch::Instance *inst) : BaseClass(inst, LayerObjectTypeParameterValidation) {}
     ~StatelessValidation() {}
 
     VkPhysicalDeviceLimits device_limits = {};
     vvl::unordered_map<VkPhysicalDevice, VkPhysicalDeviceProperties *> physical_device_properties_map;
     vvl::unordered_map<VkPhysicalDevice, vvl::unordered_set<vvl::Extension>> device_extensions_enumerated{};
+    // Validation requires special handling for VkPhysicalDeviceRobustness2FeaturesEXT, because for some cases robustness features
+    // need to only be supported, not enabled
+    vvl::unordered_map<VkPhysicalDevice, VkPhysicalDeviceRobustness2FeaturesEXT> device_supported_robustness_2_features{};
```
spencer-lunarg, line 47: Ok, so I found in `state_tracker.h` we actually have

```c++
    // TODO - Issue 5657
    // bool has_robust_image_access2;  // VK_EXT_robustness2
    // bool has_robust_buffer_access2; // VK_EXT_robustness2
```

as this isn't the first time we had this issue. So instead of doing things this way, look at `has_robust_image_access` and we should mimic that style (to be consistent) for detecting if the feature bit is set or not

ziga-lunarg: I don't think just checked if the extension is enabled works for VK_EXT_robustness2.

It works for robustImageAccess because: 

> If VK_EXT_image_robustness is supported, robustImageAccess must be supported

But with VK_EXT_robustness2 not all features are required:

> If VK_EXT_robustness2 is supported, robustBufferAccess2, robustImageAccess2, or nullDescriptor must be supported

spencer-lunarg: agree, we can update things in the future, but rather get something that we know works in now

PR #9089, opened 2024-12-24
layers: Fix checking if event is in use

### LINKED ISSUE #8748: Validation of VUID 09543 does not handle all cases
`VUID-vkSetEvent-event-09543` should be triggered for the test removed in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8747

ziga-lunarg:
Closes #8748

### DIFF
--- a/layers/core_checks/cc_synchronization.cpp
+++ b/layers/core_checks/cc_synchronization.cpp
@@ -1407,7 +1407,7 @@ void CoreChecks::PreCallRecordCmdPipelineBarrier2(VkCommandBuffer commandBuffer,
 bool CoreChecks::PreCallValidateSetEvent(VkDevice device, VkEvent event, const ErrorObject &error_obj) const {
     bool skip = false;
     if (auto event_state = Get<vvl::Event>(event)) {
-        if (event_state->write_in_use) {
+        if (event_state->InUse()) {
             skip |= LogError("VUID-vkSetEvent-event-09543", event, error_obj.location.dot(Field::event),
                              "(%s) that is already in use by a command buffer.", FormatHandle(event).c_str());
         }
--- a/tests/unit/sync_object.cpp
+++ b/tests/unit/sync_object.cpp
@@ -3821,7 +3821,9 @@ TEST_F(NegativeSyncObject, ResetEventThenSet) {
     m_default_queue->Wait();
 }
 
-// TODO: https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8748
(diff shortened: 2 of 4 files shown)


layers/core_checks/cc_synchronization.cpp
```diff
@@ -1405,7 +1405,7 @@ void CoreChecks::PreCallRecordCmdPipelineBarrier2(VkCommandBuffer commandBuffer,
 bool CoreChecks::PreCallValidateSetEvent(VkDevice device, VkEvent event, const ErrorObject &error_obj) const {
     bool skip = false;
     if (auto event_state = Get<vvl::Event>(event)) {
-        if (event_state->write_in_use) {
+        if (event_state->InUse()) {
```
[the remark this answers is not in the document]
ziga-lunarg: Good point

jeremyg-lunarg (approved):

tests/unit/sync_object.cpp
```diff
@@ -3775,6 +3789,15 @@ TEST_F(NegativeSyncObject, DISABLED_WaitEventThenSet) {
     m_errorMonitor->SetDesiredError("VUID-vkSetEvent-event-09543");
     vk::SetEvent(device(), event.handle());
     m_errorMonitor->VerifyFound();
+
+    vkt::CommandPool second_pool(*m_device, m_second_queue->family_index);
```
artem-lunarg, line 3793: This mechanism to resolve wait form a different queue works for semaphores but not for events (events communicate within a single queue). I don't know a legal mechanism to signal event after the wait, it should be ordered (did not think about this when wrote initial version of the test). I think it's correct behavior that submission hangs in this scenario. One option is to have this test disabled by default only for manual inspection. Not sure if it causes timeout, if yes, we could rely on it in the test, even though that's not great.

Not sure why originally `write_in_use` was introduced. It was in 2016. Here's that commit for documentation purposes: https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/860b0fe7c661ba7f7ee72b45488a6a7f9758c8c5

In that commit `writeEventsBeforeWait` and `waitedEvents` were also introduced together with `write_in_use`. If we don't need write_in_use then it could be need to remove these variables too.

jeremyg-lunarg: sadly, I don't think we have a PR for that commit. One possibility is that the in use tracking worked differently back then and that's why `write_in_use` and `writeEventsBeforeWait`, etc. were needed.

PR #9092, opened 2024-12-26
layers: Fix checking for swapchain protected bit support

### LINKED ISSUE #4692: Validation layers give VU error when creating protected swapchain on devices that don't support VK_KHR_surface_protected_capabilities
**Describe the Issue**
When doing vkCreateSwapchainKHR with VK_SWAPCHAIN_PROTECTED_BIT_KHR set on a Vulkan instance that doesn't support VK_KHR_surface_protected_capabilities, the layers will always log error VUID-VkSwapchainCreateInfoKHR-flags-03187. This is incorrect, since that VUID was added to the spec in VK_KHR_surface_protected_capabilities; before that, there was no such restriction on VK_SWAPCHAIN_PROTECTED_BIT_KHR.

**Valid Usage ID**
VUID-VkSwapchainCreateInfoKHR-flags-03187

**Environment:**
 - OS: Fuchsia
 - GPU: ARM Mali
 - SDK or header version if building from repo: 1.3.231
 - Options enabled (synchronization, best practices, etc.): Default

**Additional context**

ziga-lunarg:
Closes #4692


layers/stateless/sl_wsi.cpp
```diff
@@ -118,6 +119,35 @@ bool StatelessValidation::ValidateSwapchainCreateInfo(const VkSwapchainCreateInf
...
         }
     }
+
+    if (create_info.flags & VK_SWAPCHAIN_CREATE_PROTECTED_BIT_KHR) {
+        const bool is_required_ext_supported =
+            SupportedByPdev(physical_device, vvl::Extension::_VK_KHR_surface_protected_capabilities);
+        const bool is_required_ext_enabled = IsExtEnabled(instance_extensions.vk_khr_surface_protected_capabilities);
+
+        // Assume that the "protected" flag is not supported if VK_KHR_surface_protected_capabilities is not enabled
+        bool log_error = is_required_ext_supported && !is_required_ext_enabled;
```
spencer-lunarg, line 129: so this whole VU (and honestly the way `VK_KHR_surface_protected_capabilities` was added) is not obvious/easy to understand

I would just have a dedicated

```
if (is_required_ext_supported && !is_required_ext_enabled) {
    LogError("VUID-VkSwapchainCreateInfoKHR-flags-03187", ...., "explain what is going on here and how by not enabling VK_KHR_surface_protected_capabilities and checking for the flag, we are assuming it is not valid"
} 
```

and then another check inside the `if (is_required_ext_enabled)` case

ziga-lunarg: `if (is_required_ext_enabled)` would also need `is_required_ext_supported` imo, otherwise you can't really do anything about it

spencer-lunarg: `is_required_ext_enabled && !is_required_ext_supported`  is not a valid case, you can't enable an extension not supported and then none of this should matter

ziga-lunarg: Right, it is confusing me

spencer-lunarg: feel free to just redo this logic... this is a very special case, so no other code to really compare too and will be un-fun to read regardless

PR #9093, opened 2024-12-27
sync: Fix creating dynamic rendering attachments in sync val

ziga-lunarg:
Syncval would assert if dynamic rendering contains a depth attachment without depth aspect or stencil attachment without stencil aspect

### DIFF
--- a/layers/sync/sync_renderpass.cpp
+++ b/layers/sync/sync_renderpass.cpp
@@ -976,9 +976,11 @@ syncval_state::DynamicRenderingInfo::Attachment::Attachment(const SyncValidator 
     if (view) {
         if (type == AttachmentType::kColor) {
             view_gen = view->MakeImageRangeGen(offset, extent);
-        } else if (type == AttachmentType::kDepth) {
+        } else if (type == AttachmentType::kDepth &&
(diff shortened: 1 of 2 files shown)


layers/sync/sync_renderpass.cpp
```diff
@@ -976,9 +976,11 @@ syncval_state::DynamicRenderingInfo::Attachment::Attachment(const SyncValidator
     if (view) {
         if (type == AttachmentType::kColor) {
             view_gen = view->MakeImageRangeGen(offset, extent);
-        } else if (type == AttachmentType::kDepth) {
+        } else if (type == AttachmentType::kDepth &&
```
artem-lunarg, line 979: Is it a core validation error when depth attachment does not have depth aspect (and similar for stencil)?

If this above check is to skip situation that is not valid in core checks, that one option to consider not to create `DynamicRenderingInfo::Attachment` objects at all, if that's something that works with current design.

ziga-lunarg: There was no VUID reported in core validation when I hit this

artem-lunarg: Asked question here https://gitlab.khronos.org/vulkan/vulkan/-/issues/4130, but that's a separate issue, syncval should handle invalid configurations without crashes/asserts in case core checks are disabled.

artem-lunarg (approved):

PR #9095, opened 2024-12-27
gpuav: Always create gpuav helper pipeline

ziga-lunarg:


arno-lunarg (approved):
You remind me that I need to update `gpuav_dispatch` so that it matches how we do things now in `gpuav_draw`!

tests/unit/gpu_av_shader_object_positive.cpp
```diff
@@ -453,3 +454,59 @@ TEST_F(PositiveGpuAVShaderObject, RestoreUserPushConstants2) {
     }
     graphics_storage_buffer.Memory().Unmap();
 }
+
+TEST_F(PositiveGpuAVShaderObject, Test) {
```
spencer-lunarg, line 458: better name for test

spencer-lunarg (approved):

tests/unit/gpu_av_shader_object_positive.cpp
ziga-lunarg: Oops, that is my placeholder name

PR #9096, opened 2024-12-27
chassis: Split up DeviceObject

jeremyg-lunarg:
Create separate vvl::dispatch::Instance and vvl::dispatch::Device objects with most initializtion happening in constructors.

Add Instance and Device specific construtors to ValidiationObject, even though it isn't yet split up.


scripts/generators/dispatch_object_generator.py
```diff
@@ -224,15 +233,12 @@ def generate(self):
...
-    def generateMethods(self):
+    def generateMethods(self, want_instance):
         out = []
-        out.append('''
-            // This file contains methods for class DispatchObject and it is designed to ONLY be
-            // included into dispatch_object.h.
-            ''')
-
         guard_helper = PlatformGuardHelper()
         for command in self.vk.commands.values():
```
spencer-lunarg, line 239: `for command in [x for x in self.vk.commands.values() if x.instance == want_instance]:`

spencer-lunarg (approved):

PR #9097, opened 2024-12-28
sync: Add missing command in assert

ziga-lunarg:


layers/sync/sync_op.cpp
```diff
@@ -835,6 +835,7 @@ bool SyncOpResetEvent::DoValidate(const CommandExecutionContext &exec_context, c
             default:
```
spencer-lunarg, line 835: I would just change this to be

```
case vvl::Func::Empty:
case vvl::Func::vkCmdResetEvent:
case vvl::Func::vkCmdResetEvent2KHR:
case vvl::Func::vkCmdResetEvent2:
    break; // valid, but nothing to d
default:
    assert(false);
    break;
```

instead of doing this assert here like this

spencer-lunarg (approved):

PR #9112, opened 2024-12-31
tests: Remove using std::

spencer-lunarg:


ziga-lunarg (approved):
LGTM, very much in favour of not using `using namespace`

PR #9131, opened 2025-01-04
layers: Improve Push Descriptor error messages

spencer-lunarg:
This mainly cleans up the `VkWriteDescriptorSet` checks to provide a better error message when using `VK_KHR_push_descriptor` (now promoted to 1.4)

the tl;dr is for the normal flow, you look at `VkWriteDescriptorSet::dstSet` to get the VkDescriptorSetLayout... but with Push Descriptors, you use the `VkPipelineLayout` at the `vkCmdPushDescriptorSet` call

Many error messages were really confusing because it would talk about the `VkDescriptorSet`, but there isn't one with Push Descriptors

### DIFF
--- a/layers/core_checks/cc_descriptor.cpp
+++ b/layers/core_checks/cc_descriptor.cpp
@@ -1,7 +1,7 @@
-/* Copyright (c) 2015-2024 The Khronos Group Inc.
- * Copyright (c) 2015-2024 Valve Corporation
- * Copyright (c) 2015-2024 LunarG, Inc.
- * Copyright (C) 2015-2024 Google Inc.
+/* Copyright (c) 2015-2025 The Khronos Group Inc.
+ * Copyright (c) 2015-2025 Valve Corporation
+ * Copyright (c) 2015-2025 LunarG, Inc.
+ * Copyright (C) 2015-2025 Google Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -897,9 +897,9 @@ bool CoreChecks::VerifyUpdateConsistency(const vvl::DescriptorSet &set, uint32_t
     if (skip) {
         std::stringstream error_str;
         if (set.IsPushDescriptor()) {
-            error_str << " push descriptors";
+            error_str << "(push descriptors)";
         } else {
-            error_str << " descriptor set " << FormatHandle(set);
+            error_str << FormatHandle(set);
         }
         error_str << " binding #" << orig_binding.binding << " with #" << update_count
                   << " descriptors being updated but this update oversteps the bounds of this binding and the next binding is "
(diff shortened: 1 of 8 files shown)


layers/core_checks/cc_descriptor.cpp
```diff
@@ -3364,7 +3348,8 @@ bool CoreChecks::PreCallValidateAllocateDescriptorSets(VkDevice device, const Vk
 
         if (ds_layout_state->IsPushDescriptor()) {
             skip |= LogError("VUID-VkDescriptorSetAllocateInfo-pSetLayouts-00308", pAllocateInfo->pSetLayouts[i], set_layout_loc,
-                             "(%s) was created with VK_DESCRIPTOR_SET_LAYOUT_CREATE_PUSH_DESCRIPTOR_BIT.",
+                             "(%s) was created with VK_DESCRIPTOR_SET_LAYOUT_CREATE_PUSH_DESCRIPTOR_BIT. (For Push Descriptors, "
+                             "you don't allocate a VkDescriptorSet and VkWriteDescriptorSet::dstSet is ignored)",
```
spencer-lunarg, line 3352: I would love to find more spots like this that we could add hints, just hard to maintain (as extensions warp the API more) outside some obvious spots like this

layers/core_checks/cc_descriptor.cpp
```diff
@@ -3734,78 +3725,84 @@ bool CoreChecks::ValidateCmdPushDescriptorSetWithTemplate(VkCommandBuffer comman
 
     const bool is_2 =
         loc.function != Func::vkCmdPushDescriptorSetWithTemplateKHR && loc.function != Func::vkCmdPushDescriptorSetWithTemplate;
-    auto layout_data = Get<vvl::PipelineLayout>(layout);
```
spencer-lunarg, line 3737: This is all just making this (`ValidateCmdPushDescriptorSetWithTemplate`) and `ValidateCmdPushDescriptorSet` the same code path (tried to combine, but will need some more plumbing to get the correct VUID to each error message)

layers/core_checks/cc_descriptor.cpp
```diff
@@ -1670,57 +1662,77 @@ bool CoreChecks::ValidateBufferUpdate(const VkDescriptorBufferInfo &buffer_info,
...
+                        "%s has been destroyed.%s", FormatHandle(dst_layout->Handle()).c_str(),
+                        dsl_error_source.PrintMessage(*this).c_str());
     }
 
     const Location dst_binding_loc = write_loc.dot(Field::dstBinding);
     if (dst_layout->GetBindingCount() == 0) {
-        return LogError("VUID-VkWriteDescriptorSet-dstBinding-10009", objlist, write_loc.dot(Field::dstSet),
-                        "was created with %s that has a binding count of zero.", FormatHandle(dst_layout->Handle()).c_str());
+        return LogError("VUID-VkWriteDescriptorSet-dstBinding-10009", dst_layout->Handle(), dsl_error_source.ds_loc_,
+                        "%s was created with bindingCount of zero.%s", FormatHandle(dst_layout->Handle()).c_str(),
```
spencer-lunarg, line 1680: the "main" thing going on here is the `write_loc.dot(Field::dstSet)` is wrong when using Push Descriptors

so now the error message will point to `dstSet` if normal, but won't if using Push Descriptors

The `dsl_error_source.PrintMessage` at the end will then print if the `VkDescriptorSetLayout` is from the `VkDescriptorSet` or the `VkPipelineLayout` if using push descriptors

artem-lunarg (approved):

PR #9132, opened 2025-01-05
layers: Fix getting bind memory status result

ziga-lunarg:
When handle wrapping was enabled, `VkBindMemoryStatus` was not handled correctly


spencer-lunarg:
do you have test to reproduce what was going on here?

spencer-lunarg (approved):
wow, I totally forgot about `VkBindMemoryStatus` ... this looks good, thanks!

the test is good too, can reproduce the issue

PR #9134, opened 2025-01-05
gpuav: Fix enabling device features

ziga-lunarg:


layers/gpu/core/gpuav_setup.cpp
```diff
@@ -423,6 +423,7 @@ void Validator::PreCallRecordCreateDevice(VkPhysicalDevice physicalDevice, const
         } else {
             // Need to add a VkPhysicalDeviceFeatures pointer
             enabled_features = new VkPhysicalDeviceFeatures;
```
spencer-lunarg, line 425:
```suggestion
 ```
 
 I think this memset is doing what the `new` was trying to do, so not sure why keep both

ziga-lunarg: `enabled_features` is `NULL` here, there is a check 7 lines above

spencer-lunarg (approved):

PR #9135, opened 2025-01-05
bp: Fix assert when buffer can be null

ziga-lunarg:
If the maintenance6 feature is enabled, buffer can be VK_NULL_HANDLE. If buffer is VK_NULL_HANDLE and the nullDescriptor feature is enabled, every index fetched results in a value of zero.

### DIFF
--- a/layers/best_practices/bp_drawdispatch.cpp
+++ b/layers/best_practices/bp_drawdispatch.cpp
@@ -1,6 +1,6 @@
-/* Copyright (c) 2015-2024 The Khronos Group Inc.
- * Copyright (c) 2015-2024 Valve Corporation
- * Copyright (c) 2015-2024 LunarG, Inc.
+/* Copyright (c) 2015-2025 The Khronos Group Inc.
+ * Copyright (c) 2015-2025 Valve Corporation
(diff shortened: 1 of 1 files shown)


layers/best_practices/bp_drawdispatch.cpp
```diff
@@ -214,7 +214,9 @@ bool BestPractices::ValidateIndexBufferArm(const bp_state::CommandBuffer& cmd_st
 
     // check for sparse/underutilised index buffer, and post-transform cache thrashing
     const auto ib_state = Get<vvl::Buffer>(cmd_state.index_buffer_binding.buffer);
-    ASSERT_AND_RETURN_SKIP(ib_state);
+    if (!ib_state) {
```
spencer-lunarg, line 217: > If the maintenance6 feature is enabled, buffer can be VK_NULL_HANDLE. If buffer is VK_NULL_HANDLE and the nullDescriptor feature is enabled, every index fetched results in a value of zero.

This is good information, could you leave a comment here explaining that

spencer-lunarg (approved):

spencer-lunarg:
Thanks, I like slowly learning where things **can** be null and where we should be asserting!

PR #9136, opened 2025-01-05
bp: Fix index buffer scan when robustness is enabled

ziga-lunarg:
Check if the memory allocation size is large enough for all indices

### DIFF
--- a/layers/best_practices/bp_drawdispatch.cpp
+++ b/layers/best_practices/bp_drawdispatch.cpp
@@ -1,6 +1,6 @@
-/* Copyright (c) 2015-2024 The Khronos Group Inc.
- * Copyright (c) 2015-2024 Valve Corporation
- * Copyright (c) 2015-2024 LunarG, Inc.
+/* Copyright (c) 2015-2025 The Khronos Group Inc.
+ * Copyright (c) 2015-2025 Valve Corporation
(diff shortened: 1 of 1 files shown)


layers/best_practices/bp_drawdispatch.cpp
```diff
@@ -234,126 +234,129 @@ bool BestPractices::ValidateIndexBufferArm(const bp_state::CommandBuffer& cmd_st
...
-                primitive_restart_value = 0xFF;
-            } else if (ib_type == VK_INDEX_TYPE_UINT16) {
-                scan_index = *reinterpret_cast<const uint16_t*>(scan_ptr);
-                primitive_restart_value = 0xFFFF;
-            } else {
-                scan_index = *reinterpret_cast<const uint32_t*>(scan_ptr);
-                primitive_restart_value = 0xFFFFFFFF;
-            }
+        // Check if all indices are within the memory allocation size, if robustness is enabled they might not be
+        if ((firstIndex + indexCount) * scan_stride <= ib_mem_state->allocate_info.allocationSize) {
```
spencer-lunarg, line 238: maybe just flip the logic and `return skip;` instead of adding another indent level here (since the `ValidateIndexBufferArm` all seems to rely on this being true in order to check anything)

spencer-lunarg (approved):

PR #9140, opened 2025-01-06
bp: Fix getting primitive restart enable

ziga-lunarg:


layers/best_practices/bp_drawdispatch.cpp
```diff
@@ -228,11 +228,13 @@ bool BestPractices::ValidateIndexBufferArm(const bp_state::CommandBuffer& cmd_st
     bool primitive_restart_enable = false;
 
     const auto* pipeline_state = cmd_state.GetCurrentPipeline(VK_PIPELINE_BIND_POINT_GRAPHICS);
-    ASSERT_AND_RETURN_SKIP(pipeline_state);
-
-    const auto* ia_state = pipeline_state->InputAssemblyState();
-    if (ia_state) {
-        primitive_restart_enable = ia_state->primitiveRestartEnable == VK_TRUE;
+    if (pipeline_state && pipeline_state->IsDynamic(CB_DYNAMIC_STATE_PRIMITIVE_RESTART_ENABLE)) {
```
spencer-lunarg, line 231: shouldn't it be 

```suggestion
    if (pipeline_state && !pipeline_state->IsDynamic(CB_DYNAMIC_STATE_PRIMITIVE_RESTART_ENABLE)) {
```

we want to use `cmd_state.dynamic_state_value.primitive_restart_enable` if it is dynamic

spencer-lunarg (commented):
ugh, just realize the dynamic state stuff hasn't been ported to BP... there might be many more subtle things like this

layers/best_practices/bp_drawdispatch.cpp
```diff
@@ -228,11 +228,13 @@ bool BestPractices::ValidateIndexBufferArm(const bp_state::CommandBuffer& cmd_st
     bool primitive_restart_enable = false;
 
     const auto* pipeline_state = cmd_state.GetCurrentPipeline(VK_PIPELINE_BIND_POINT_GRAPHICS);
-    ASSERT_AND_RETURN_SKIP(pipeline_state);
-
-    const auto* ia_state = pipeline_state->InputAssemblyState();
-    if (ia_state) {
-        primitive_restart_enable = ia_state->primitiveRestartEnable == VK_TRUE;
+    if (pipeline_state && !pipeline_state->IsDynamic(CB_DYNAMIC_STATE_PRIMITIVE_RESTART_ENABLE)) {
```
spencer-lunarg, line 231: sorry didn't just say this first, but we should have things like we do in `LastBound::IsDepthTestEnable()`... but seems the BP code lacks any plumbing of `LastBound` which it should... I can take care of this afterwards, for no this is fine short term

spencer-lunarg (approved):

PR #9144, opened 2025-01-06
Fix crash when dst buffer has no memory

ziga-lunarg:
@spencer-lunarg There was no test for `VUID-vkCmdCopyBuffer-dstBuffer-00121` and it would have found this


spencer-lunarg (approved):

spencer-lunarg:
@ziga-lunarg thanks for fixing this!!

>  There was no test for VUID-vkCmdCopyBuffer-dstBuffer-00121

I need to re-look at this then, I didn't have a good way to detect automatically which things like `VUID-vkCmdCopyBuffer` were missing a test vs just having the `vkCmdCopyBuffer2` version tested

PR #9146, opened 2025-01-06
Fixes for VkBindMemoryStatus

spencer-lunarg:
`VkBindMemoryStatus` was added in `VK_KHR_maintenance6` (which is now part of 1.4)

I allows cases where people batch many things in `vkBindImageMemory2` but provides a way to know which buffer/images actually bound memory or not

This was not well plumbed out throughout the VVL

### DIFF
--- a/tests/unit/memory_positive.cpp
+++ b/tests/unit/memory_positive.cpp
@@ -1,8 +1,8 @@
 /*
- * Copyright (c) 2023-2024 The Khronos Group Inc.
- * Copyright (c) 2023-2024 Valve Corporation
- * Copyright (c) 2023-2024 LunarG, Inc.
- * Copyright (c) 2023-2024 Collabora, Inc.
+ * Copyright (c) 2023-2025 The Khronos Group Inc.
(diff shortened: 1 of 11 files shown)


tests/unit/memory_positive.cpp
```diff
@@ -556,3 +556,68 @@ TEST_F(PositiveMemory, BindMemoryDX12Handle) {
...
+    bind_info.buffer = buffer;
+    bind_info.memory = memory;
+    bind_info.memoryOffset = 0u;
+    vk::BindBufferMemory2(device(), 1u, &bind_info);
+
+    ASSERT_NE(result, VK_RESULT_MAX_ENUM);
+}
+
+TEST_F(PositiveMemory, BindMemoryStatusImage) {
+    TEST_DESCRIPTION("Use VkBindMemoryStatus when binding buffer to memory.");
```
arno-lunarg, line 596:
```suggestion
    TEST_DESCRIPTION("Use VkBindMemoryStatus when binding image to memory.");
```

arno-lunarg (approved):

PR #9148, opened 2025-01-06
bp: Use LastBound for dynamic state values

spencer-lunarg:
followup of https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9140

Make use of the `LastBound` harness to get the value from dynamic state or pipeline state


arno-lunarg (approved):
Since this new function is only called once, unless it is planned to use it at multiple other spots, I don't think adding another dedicated method is a good idea. It just adds another level of indirection.

spencer-lunarg:
> I don't think adding another dedicated method is a good idea. It just adds another level of indirection.

After dealing with Dynamic State, it is SUPER easy to not get it right, having it all in a single location (`LastBound`) has been the only way to make sure we don't make mistakes (because this logic before was actually slightly wrong, but hard to get right)

PR #9154, opened 2025-01-07
Fix C++23 build error in function_base_block.h

victorvianna:
Trying to build with C++23 fails due to the Instruction type being incomplete (probably std::unique_ptr<> requirements became stricter). This patch fixes the error by #including the type instead of forward declaring.

There are a few other C++23 build errors related to the use of std::aligned_storage, but fixing those is beyond the scope of this patch.


victorvianna:
CC @pkasting

arno-lunarg (approved):

PR #9155, opened 2025-01-07
Fix checking mesh pipeline dynamic states

ziga-lunarg:


spencer-lunarg:
> eww... this is nasty

dynamic state, your code is beautiful :rainbow:

PR #9156, opened 2025-01-07
layers: Fix checking discard rectangle dynamic state

ziga-lunarg:
If discard rectangles enable is false, then discard rectangles don't have to be set


spencer-lunarg (approved):
> If discard rectangles enable is false, then discard rectangles don't have to be set

Will update the spec VUID

Issue #9141, opened 2025-01-06
CTS draw test gets a VK_ERROR_DEVICE_LOST when run with gpuav index buffers


ziga-lunarg:
The test `dEQP-VK.draw.renderpass.indirect_draw.indexed_draw_count_clamping.triangle_list` with get a `VK_ERROR_DEVICE_LOST` if gpuav is enabled with `Index buffers`

arno-lunarg:
Issue fixed on my end, @ziga-lunarg can you still reproduce?

PR #9161, opened 2025-01-07
gpuav: Fix handling of count buffer

arno-lunarg:
Half of the fix for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9141
Second half will be disabling GPU-AV vertex buffer fetching validation.

- Clamp manually read count buffer value to maxDrawCount
- Fix count buffer offset being accounted for 2 times, once in count buffer binding, and another in shaders reading said buffer


layers/gpu/cmd_validation/gpuav_draw.cpp
```diff
@@ -261,13 +261,14 @@ void FirstInstance(Validator &gpuav, CommandBuffer &cb_state, const Location &lo
...
             shader_resources.push_constants.draw_cmds_stride_dwords = draw_cmds_byte_stride / sizeof(uint32_t);
+            shader_resources.push_constants.cpu_draw_count = draw_count;
             shader_resources.push_constants.first_instance_member_pos = first_instance_member_pos;
 
             shader_resources.draw_buffer_binding.info = {draw_buffer, 0, VK_WHOLE_SIZE};
             shader_resources.push_constants.draw_buffer_dwords_offset = (uint32_t)draw_buffer_offset / sizeof(uint32_t);
             if (count_buffer) {
                 shader_resources.push_constants.flags |= gpuav::glsl::kFirstInstanceFlags_DrawCountFromBuffer;
-                shader_resources.count_buffer_binding.info = {count_buffer, 0, sizeof(uint32_t) + count_buffer_offset};
+                shader_resources.count_buffer_binding.info = {count_buffer, 0, sizeof(uint32_t)};
```
spencer-lunarg, line 271: if we are making a change like this, we should have positive tests that would not work without this PR

arno-lunarg: Or rather they may provoke device losses, or just nothing, at least they would not be predectible

arno-lunarg: So I could add them, but they would not be reliable in detecting errors. Maybe still better than nothing?

spencer-lunarg (approved):

layers/gpu/cmd_validation/gpuav_draw.cpp
spencer-lunarg: up to you, I was under the assumption your PR here is fixing the CTS from crashing, if that is the case, having a test that mimic that CTS test in our test suite would be nice

PR #9162, opened 2025-01-07
gpuav: Disable vertex buffer fetching validation

arno-lunarg:
Second half of the fix for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9141
It can be too slow, stalling the GPU. I am moving validation to instrumentation

### DIFF
--- a/layers/gpu/core/gpuav_record.cpp
+++ b/layers/gpu/core/gpuav_record.cpp
@@ -1,6 +1,6 @@
-/* Copyright (c) 2018-2024 The Khronos Group Inc.
- * Copyright (c) 2018-2024 Valve Corporation
- * Copyright (c) 2018-2024 LunarG, Inc.
+/* Copyright (c) 2018-2025 The Khronos Group Inc.
+ * Copyright (c) 2018-2025 Valve Corporation
+ * Copyright (c) 2018-2025 LunarG, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


layers/gpu/core/gpuav_record.cpp
```diff
@@ -526,8 +526,10 @@ void Validator::PreCallRecordCmdDrawIndexedIndirect(VkCommandBuffer commandBuffe
 
     valcmd::DrawIndexedIndirectIndexBuffer(*this, *cb_state, record_obj.location, buffer, offset, stride, count, VK_NULL_HANDLE, 0,
                                            "VUID-VkDrawIndexedIndirectCommand-robustBufferAccess2-08798");
+#if 0
     valcmd::DrawIndexedIndirectVertexBuffer(*this, *cb_state, record_obj.location, buffer, offset, stride, count, VK_NULL_HANDLE, 0,
```
spencer-lunarg, line 530: I would much rather prefer

```suggestion
    // TODO - Some Issue number
    // valcmd::DrawIndexedIndirectVertexBuffer(*this, *cb_state, record_obj.location, buffer, offset, stride, count, VK_NULL_HANDLE, 0,
```

tests/unit/gpu_av_index_buffer.cpp
```diff
@@ -55,7 +55,7 @@ TEST_F(NegativeGpuAVIndexBuffer, IndexBufferOOB) {
     m_errorMonitor->VerifyFound();
 }
 
-TEST_F(NegativeGpuAVIndexBuffer, IndirectDrawBadVertexIndex32) {
+TEST_F(NegativeGpuAVIndexBuffer, DISABLED_IndirectDrawBadVertexIndex32) {
```
spencer-lunarg, line 58: `// todo issue` above these, otherwise I never know why things are DISABLED

spencer-lunarg (approved):

PR #9167, opened 2025-01-08
Small misc Sync Test fixes

spencer-lunarg:
- Move 2 tests to Sync Object
- Use `vkt::Event` the same everywhere


artem-lunarg:
+1. I think there are parts of core checks implementation in sycnval for some reason, should be moved eventually. Events in general need additional iteration of improvemetns.

artem-lunarg (approved):

PR #9172, opened 2025-01-08
vvl: Test a new .clang-format

arno-lunarg:


arno-lunarg:
So for instance I like that for GPU-AV specifically because it turns this:
image

into this:

image

So parsing the captured variables list in the lambda is way easier

layers/best_practices/bp_state.h
```diff
@@ -20,56 +20,72 @@
...
+#include "state_tracker/device_state.h"
+#include "state_tracker/image_state.h"
+#include "state_tracker/state_tracker.h"
 
 class BestPractices;
 
-namespace bp_state {
-class Image : public vvl::Image {
+namespace bp_state
+{
```
jeremyg-lunarg, line 32: moving brackets around seems very unnecessary to me

layers/best_practices/bp_state.h
```diff
@@ -80,133 +96,164 @@ class Image : public vvl::Image {
...
 // used to track state regarding render pass heuristic checks
-struct RenderPassState {
-    bool depthAttachment = false;
-    bool colorAttachment = false;
-    bool depthOnly = false;
-    bool depthEqualComparison = false;
-    uint32_t numDrawCallsDepthOnly = 0;
+struct RenderPassState
+{
+    bool     depthAttachment               = false;
```
jeremyg-lunarg, line 163: 🤮

layers/best_practices/bp_state.h
```diff
@@ -80,133 +96,164 @@ class Image : public vvl::Image {
...
-
-    std::optional<float> dynamic_priority;  // VK_EXT_pageable_device_local_memory priority
+    DeviceMemory(VkDeviceMemory                         handle,
+                 const VkMemoryAllocateInfo*            allocate_info,
+                 uint64_t                               fake_address,
+                 const VkMemoryType&                    memory_type,
+                 const VkMemoryHeap&                    memory_heap,
+                 std::optional<vvl::DedicatedBinding>&& dedicated_binding,
+                 uint32_t                               physical_device_count) :
+        vvl::DeviceMemory(handle,
```
jeremyg-lunarg, line 138: this part is mostly nicer

layers/best_practices/bp_state.h
arno-lunarg: Is the issue the blanks between the variable name and its default value?

jeremyg-lunarg: All the extra spaces on lines like this. trying to make vertical columns will lead to lots of churn whenever someone adds or removes a declaration that causes the columns to change. And inevitably someone who isn't fully up to speed on `git clang-format` will waste a bunch of time trying to fix the columns manually.

layers/best_practices/bp_ray_tracing.cpp
```diff
@@ -21,35 +21,48 @@
...
+    uint32_t                                               infoCount,
+    const VkAccelerationStructureBuildGeometryInfoKHR*     pInfos,
+    const VkAccelerationStructureBuildRangeInfoKHR* const* ppBuildRangeInfos,
+    const ErrorObject&                                     error_obj) const {
     return ValidateBuildAccelerationStructure(commandBuffer, error_obj.location);
 }
 
 bool BestPractices::ValidateBuildAccelerationStructure(VkCommandBuffer commandBuffer, const Location& loc) const {
-    bool skip = false;
+    bool skip     = false;
```
spencer-lunarg, line 58: I don't get why we need this white space like this... I'm gonna just go crazy and end up trying to add more newlines to satisfy my OCD to not have this like this

layers/best_practices/bp_state.h
```diff
@@ -156,50 +172,54 @@ struct CommandBufferStateNV {
...
+                  const VkCommandBufferAllocateInfo* allocate_info,
+                  const vvl::CommandPool*            pool);
 
-    RenderPassState render_pass_state;
+    RenderPassState      render_pass_state;
     CommandBufferStateNV nv;
-    uint64_t num_submits = 0;
-    bool uses_vertex_buffer = false;
-    uint32_t small_indexed_draw_call_count = 0;
+    uint64_t             num_submits                   = 0;
```
spencer-lunarg, line 220: sorry, not of fan have to highlight over a line to make sure I know which type it is once these get many lines stack on it

layers/best_practices/bp_state.h
spencer-lunarg: well until you have a 13-inch laptop screen and this takes up half of it :man_shrugging:

spencer-lunarg:
just want to say if we needed to revert something like https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9200 that would have been a nightmare/impossible if we touch every file like this.... I just don't see what we gain from this, the whole idea of clang-format is we picked something and stick with it, we plan to update this clang-format again in 2 years when someone else suggests it is better, if we change it now, what justification is there to **not** change it again in the future

PR #9200, opened 2025-01-10
layers: Revert all recent chassis changes

jeremyg-lunarg:
The very first change in this series is causing intermittent problems in someone's CI downstream. Revert the entire house of cards until it can be figured out.

Note that this undoes https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9132, which is not strictly related but will need to be redone on the old code.


jeremyg-lunarg:
reducing defcon level for now

PR #9202, opened 2025-01-10
gpu: Remove passthrough to run a Pass

spencer-lunarg:
We use to add this 

```
bool Module::RunPassDescriptorIndexingOOB() {
    DescriptorIndexingOOBPass pass(*this);
    const bool changed = pass.Run();
    if (print_debug_info_) {
        pass.PrintDebugInfo();
    }
    return changed;
}
```

type logic for every pass we added, removed it (and just passed in `enabled_features`) so it is more streamline to quickly/easily add a new SPIR-V pass


layers/gpuav/spirv/pass.h
```diff
@@ -33,9 +33,17 @@ struct InjectionData {
...
 class Pass {
   public:
+    // Needed to know where an error/warning comes from
     virtual const char* Name() const = 0;
+    // Return true if code was instrumented/modified in anyway
+    virtual bool Run() = 0;
+    // Require passes to print info as extremely helpful for debugging
+    virtual void PrintDebugInfo() const = 0;
+    // Wrapper that each pass can use to start
+    bool Start();
```
arno-lunarg, line 46: a `Start` calls for a `End` or a `Finish`, so IMO inappropriate here. 
If I follow things correctly, `Run` is now only called within `Start`.
So I would expect calls now looking like `modified |= texel_buffer_pass.Start();` to stay similar to how they were, by going `modified |= texel_buffer_pass.Run();`
And here `Run` would look like your `Start`:
```cpp
bool Pass::Run() {
    // I too would like to have "Instrument" be called `Run`, but since we can't I guess we can only cope 
    // Also "Instrument" should probably be a `protected` method
    const bool modified = Instrument();
    if (module_.print_debug_info_) {
        PrintDebugInfo();
    }
    return modified;
}
```

layers/gpuav/spirv/post_process_descriptor_indexing_pass.cpp
```diff
@@ -0,0 +1,200 @@
+/* Copyright (c) 2024-2025 LunarG, Inc.
```
arno-lunarg, line 1: Anything new here or just moved code?

arno-lunarg (approved):

layers/gpuav/spirv/module.h
```diff
@@ -113,8 +103,7 @@ class Module {
     const uint32_t output_buffer_descriptor_set_;
 
     const bool support_non_semantic_info_;
```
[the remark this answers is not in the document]
spencer-lunarg: this is from an extension, was debating to do it, but doubt other passes care about extension, but 100% will care about feature bits

layers/gpuav/spirv/pass.h
spencer-lunarg: the `Run()` is what `spirv-opt` uses to "run a pass"

honestly `Run` and `Instrument` are better... I just picked `Start` because I needed a name

layers/gpuav/spirv/post_process_descriptor_indexing_pass.cpp
spencer-lunarg: file name changed in last commit, was missing `_pass` in the name

Issue #9191, opened 2025-01-09
VUID-vkQueueSubmit-pCommandBuffers-00065 is validating the wrong thing


ziga-lunarg:
`NegativeSyncObject.QueueForwardProgressFenceWait` has a todo in it:

```
    // TODO: the test works according to description but that's not what VUID describes
    const char *queue_forward_progress_message = "VUID-vkQueueSubmit-pCommandBuffers-00065";
```

The VVL message does not match the spec message:

> Validation Error: [ VUID-vkQueueSubmit-pCommandBuffers-00065 ] Object 0: handle = 0xf56c9b0000000004, type = VK_OBJECT_TYPE_SEMAPHORE; Object 1: handle = 0x1bdd1e14370, type = VK_OBJECT_TYPE_QUEUE; Object 2: handle = 0x1bdd1e14370, type = VK_OBJECT_TYPE_QUEUE; | MessageID = 0x71565eaf | vkQueueSubmit(): pSubmits[0].pSignalSemaphores[0] (VkSemaphore 0xf56c9b0000000004[]) is being signaled by VkQueue 0x1bdd1e14370[], but it was previously signaled by VkQueue 0x1bdd1e14370[] and has not since been waited on.
> The Vulkan spec states: Any calls to vkCmdSetEvent, vkCmdResetEvent or vkCmdWaitEvents that have been recorded into any of the command buffer elements of the pCommandBuffers member of any element of pSubmits, must not reference any VkEvent that is referenced by any of those commands in a command buffer that has been submitted to another queue and is still in the pending state (https://docs.vulkan.org/spec/latest/chapters/cmdbuffers.html#VUID-vkQueueSubmit-pCommandBuffers-00065)

PR #9205, opened 2025-01-10
layers: Use correct VUID for signaling signaled semaphore

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9191

### DIFF
--- a/layers/core_checks/cc_synchronization.cpp
+++ b/layers/core_checks/cc_synchronization.cpp
@@ -237,8 +237,9 @@ bool SemaphoreSubmitState::ValidateSignalSemaphore(const Location &signal_semaph
                         initiator << core.FormatHandle(other_queue);
                         objlist.add(other_queue);
                     }
+                    const auto &vuid = GetQueueSubmitVUID(signal_semaphore_loc, sync_vuid_maps::SubmitError::kSemAlreadySignalled);
                     skip |= core.LogError(
-                        "VUID-vkQueueSubmit-pCommandBuffers-00065", objlist, signal_semaphore_loc,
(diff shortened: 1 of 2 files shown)


layers/core_checks/cc_synchronization.cpp
```diff
@@ -237,8 +237,9 @@ bool SemaphoreSubmitState::ValidateSignalSemaphore(const Location &signal_semaph
                         initiator << core.FormatHandle(other_queue);
                         objlist.add(other_queue);
                     }
+                    const auto &vuid = GetQueueSubmitVUID(signal_semaphore_loc, sync_vuid_maps::SubmitError::kSemAlreadySignalled);
                     skip |= core.LogError(
-                        "VUID-vkQueueSubmit-pCommandBuffers-00065", objlist, signal_semaphore_loc,
```
ziga-lunarg, line 241: Are we now missing validation for 00065?

ziga-lunarg (approved):

layers/core_checks/cc_synchronization.cpp
artem-lunarg: Yes. From what I remember from the last year the events in core checks has implementation issues. I used BP as playground to implement "proper" event tracking here https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8033 (in addition to command primary buffer tracking, track secondary cmdbufs and state between submissions), and the idea was later to do similar thing in core checks. This includes moving parts of code from syncval (for some reason part of even code are there...). If you notice something fix is limited buy current implementation it's probably can be postponed.

PR #9210, opened 2025-01-13
layers: Combine vkCmdSetDiscardRectangleEXT check

spencer-lunarg:
For change in https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/7099

New VU error looks like

> [ VUID-vkCmdDraw-None-07751 ] vkCmdDraw(): vkCmdSetDiscardRectangleEXT was not set for discard rectangle index 1 for this command buffer. Need to set once for each in discardRectangleCount (2). (There was a call to vkCmdBindPipeline that didn't have VK_DYNAMIC_STATE_DISCARD_RECTANGLE_EXT and invalidated the prior vkCmdSetDiscardRectangleEXT call).

### DIFF
--- a/layers/core_checks/cc_cmd_buffer_dynamic.cpp
+++ b/layers/core_checks/cc_cmd_buffer_dynamic.cpp
@@ -20,6 +20,7 @@
 #include <vulkan/vk_enum_string_helper.h>
 #include "core_validation.h"
 #include "drawdispatch/drawdispatch_vuids.h"
+#include "generated/error_location_helper.h"
 #include "generated/vk_extension_helper.h"
 #include "generated/dispatch_functions.h"
 #include "state_tracker/image_state.h"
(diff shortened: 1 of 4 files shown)


layers/core_checks/cc_cmd_buffer_dynamic.cpp
```diff
@@ -390,6 +391,35 @@ bool CoreChecks::ValidateGraphicsDynamicStateSetStatus(const LastBound& last_bou
...
+                    // vkCmdSetDiscardRectangleEXT needs to be set on each rectangle
+                    for (uint32_t i = 0; i < rect_limit; i++) {
+                        if (!cb_state.dynamic_state_value.discard_rectangles.test(i)) {
+                            const vvl::Field limit_name =
+                                has_pipeline ? vvl::Field::discardRectangleCount : vvl::Field::maxDiscardRectangles;
+                            const char* vuid2 =
+                                has_pipeline ? vuid.dynamic_discard_rectangle_07751 : vuid.set_discard_rectangle_09236;
+                            skip |= LogError(vuid2, cb_state.Handle(), vuid.loc(),
+                                             "vkCmdSetDiscardRectangleEXT was not set for discard rectangle index %" PRIu32
+                                             " for this command buffer. Need to set once for each in %s (%" PRIu32 ").%s",
```
ziga-lunarg, line 416: `Need to set once for each` is slightly grammatically incorrect

ziga-lunarg (approved):

layers/core_checks/cc_cmd_buffer_dynamic.cpp
spencer-lunarg: what would you suggest? I want to make it clear then need to set it from `0` to `limit - 1`

ziga-lunarg: `It needs to be set once for each rectangle in`
or something like that

spencer-lunarg: done, agree that sounds better

PR #9212, opened 2025-01-13
layers: Add warning for using Pointer in shader interface

spencer-lunarg:
(long details in https://gitlab.khronos.org/spirv/SPIR-V/-/issues/779)

but the tl;dr is we have no real CTS coverage or spec language around using a pointer for a shader interface and how it works for things like the amount of `Location` slots used

For the time being, we should at least warn the users as this is ticking timebomb for someone to get trapped in. Also having these tests help regression around the shader interface code that has to dance around this


spencer-lunarg:
> CI Vulkan-ValidationLayers build # 18634 failed.

Seems CI also shows that using `PhysicalStorageBuffer` can blow up at vkCreateShaderModule time even!

artem-lunarg (approved):

PR #9213, opened 2025-01-13
tests: Add SSBO Length DebugPrintf tests

spencer-lunarg:
Some local tests I found, were used to make sure DebugPrintf worked with the `OpArrayLength` correctly and figured it is worth adding


spencer-lunarg:
created https://gitlab.khronos.org/vulkan/vulkan/-/issues/4143 - internal CI is showing some drivers return different results

artem-lunarg (approved):

PR #9218, opened 2025-01-14
Fix WSI tests on MockICD X11

spencer-lunarg:
Running X11, these WSI tests were not working

### DIFF
--- a/tests/framework/binding.h
+++ b/tests/framework/binding.h
@@ -1495,6 +1495,13 @@ class Surface {
         return *this;
     }
 
+    // This is ONLY for tests that need a way test destroying an instance and leak the Surface object (and calling
+    // vkDestroySurfaceKHR will be invalid)
+    void DestroyExplicitly() {
+        handle_ = VK_NULL_HANDLE;
+        instance_ = VK_NULL_HANDLE;
+    }
+
   private:
     VkInstance instance_ = VK_NULL_HANDLE;
     VkSurfaceKHR handle_ = VK_NULL_HANDLE;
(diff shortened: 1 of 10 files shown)


tests/framework/binding.h
```diff
@@ -1495,6 +1495,13 @@ class Surface {
         return *this;
     }
 
+    // This is ONLY for tests that need a way test destroying an instance and leak the Surface object (and calling
+    // vkDestroySurfaceKHR will be invalid)
+    void DestroyExplicitly() {
```
artem-lunarg, line 1500: I think I understand purpose - not to use stale instance handle to destroy surface, but the method name itself is not clear, do we destroy surface explicitly?

spencer-lunarg: (happy to change this, but also the Return on Investment might be low)

Unlike every other handle, there is no single `vkCreateSurface`  and we basically need to use the `vkt::Surface` object, which means we need a way to not have it call `vkDestroySurface()` for the one or two tests need to let it

Wanted to give it somewhat of a non-useful name so no one tries using it blindly

artem-lunarg: If this solves the problem, maybe for special cases it's fine not to use helper objects, at some point handling special cases might start adding complexity to common code.

spencer-lunarg: Idk how else to solve it without a helper function

1. Make the private variables not private
2. Make the test only work for Wayland or Win32 so I know I can create the `VkSurfaceKHR` object for them

Issue #9150, opened 2025-01-06
Add missing sync object tests


spencer-lunarg:
Queue Family

- [x] VUID-VkBufferMemoryBarrier-None-09097 (https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9189)
- [x] VUID-VkImageMemoryBarrier2-None-09119
- [x] VUID-VkBufferMemoryBarrier-None-09098
- [x] VUID-VkImageMemoryBarrier2-None-09120
- [x] VUID-VkBufferMemoryBarrier-srcQueueFamilyIndex-09099
- [x] VUID-VkImageMemoryBarrier2-srcQueueFamilyIndex-09121
- [x] VUID-VkBufferMemoryBarrier-dstQueueFamilyIndex-09100
- [x] VUID-VkImageMemoryBarrier2-dstQueueFamilyIndex-09122

Feature Bits

- [x] VUID-vkCmdPipelineBarrier-dstStageMask-04091 (https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9192)
- [x] VUID-vkCmdPipelineBarrier-dstStageMask-04092
- [x] VUID-vkCmdPipelineBarrier-dstStageMask-04093
- [x] VUID-vkCmdPipelineBarrier-dstStageMask-04094
- [x] VUID-vkCmdPipelineBarrier-dstStageMask-03937
- [x] VUID-vkCmdPipelineBarrier-dstStageMask-07318
- [x] VUID-VkBufferMemoryBarrier2-dstStageMask-04957
- [x] VUID-VkBufferMemoryBarrier2-dstStageMask-04995

access mask and pipeline stages

- [x] VUID-VkMemoryBarrier2-dstAccessMask-03918 (https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9193)
- [x] VUID-VkMemoryBarrier2-dstAccessMask-03919
- [x] VUID-VkMemoryBarrier2-dstAccessMask-03920
- [x] VUID-VkMemoryBarrier2-dstAccessMask-04747
- [x] VUID-VkMemoryBarrier2-dstAccessMask-03926

no memory bound

- [x] VUID-VkBufferMemoryBarrier-buffer-01931 (https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9190)

unsignaled

- [ ] VUID-vkQueueSubmit-pSignalSemaphores-00067

PR #9223, opened 2025-01-15
tests: Test signaling already signaled semaphore

ziga-lunarg:
Last of #9150

### DIFF
--- a/tests/unit/sync_object.cpp
+++ b/tests/unit/sync_object.cpp
@@ -4799,3 +4799,20 @@ TEST_F(NegativeSyncObject, UnsupportedBufferMemoryBarrier2Stages) {
 
     m_command_buffer.End();
 }
+
+TEST_F(NegativeSyncObject, SignalSemaphoreTwice) {
+    RETURN_IF_SKIP(Init());
+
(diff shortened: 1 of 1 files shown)


tests/unit/sync_object.cpp
```diff
@@ -4799,3 +4799,20 @@ TEST_F(NegativeSyncObject, UnsupportedBufferMemoryBarrier2Stages) {
 
     m_command_buffer.End();
 }
+
+TEST_F(NegativeSyncObject, SignalSemaphoreTwice) {
```
artem-lunarg, line 4803: Is it different from `NegativeSyncObject.SignalSignaledSemaphore` ?

ziga-lunarg: Didn't realize this VUID was already covered

spencer-lunarg: me neither, sorry for adding to the issue if that was a mistake

artem-lunarg: no worries, it was added few days ago, probably when issue already existed

PR #9227, opened 2025-01-15
sync: Implement new formatting of syncval accesses

artem-lunarg:
Use native Vulkan stage/access constants instead of synval internal representation in the error messages.

Using native constants is more familiar to the users but also provides opportunity to use meta stages to have simpler messages.

Log of doom capture error messages is reduced from 105Kb -> 69Kb.

`write_barriers` extra property before:
> write_barriers = SYNC_COMPUTE_SHADER_SHADER_BINDING_TABLE_READ|SYNC_COMPUTE_SHADER_SHADER_SAMPLED_READ|SYNC_COMPUTE_SHADER_SHADER_STORAGE_READ|SYNC_COMPUTE_SHADER_SHADER_STORAGE_WRITE|SYNC_COMPUTE_SHADER_UNIFORM_READ|SYNC_ACCELERATION_STRUCTURE_COPY_ACCELERATION_STRUCTURE_READ|SYNC_ACCELERATION_STRUCTURE_COPY_ACCELERATION_STRUCTURE_WRITE|SYNC_ACCELERATION_STRUCTURE_COPY_TRANSFER_READ|SYNC_ACCELERATION_STRUCTURE_COPY_TRANSFER_WRITE|SYNC_COPY_TRANSFER_READ|SYNC_COPY_TRANSFER_WRITE|SYNC_RESOLVE_TRANSFER_READ|SYNC_RESOLVE_TRANSFER_WRITE|SYNC_BLIT_TRANSFER_READ|SYNC_BLIT_TRANSFER_WRITE|SYNC_CLEAR_TRANSFER_WRITE

after:
> write_barriers = VK_PIPELINE_STAGE_2_COMPUTE_SHADER_BIT(VK_ACCESS_2_UNIFORM_READ_BIT|VK_ACCESS_2_SHADER_READ_BIT|VK_ACCESS_2_SHADER_STORAGE_WRITE_BIT):VK_PIPELINE_STAGE_2_ALL_TRANSFER_BIT(ALL_ACCESSES)

### DIFF
--- a/layers/utils/vk_layer_utils.h
+++ b/layers/utils/vk_layer_utils.h
@@ -1,6 +1,6 @@
-/* Copyright (c) 2015-2017, 2019-2024 The Khronos Group Inc.
- * Copyright (c) 2015-2017, 2019-2024 Valve Corporation
- * Copyright (c) 2015-2017, 2019-2024 LunarG, Inc.
+/* Copyright (c) 2015-2017, 2019-2025 The Khronos Group Inc.
+ * Copyright (c) 2015-2017, 2019-2025 Valve Corporation
+ * Copyright (c) 2015-2017, 2019-2025 LunarG, Inc.
  * Modifications Copyright (C) 2022 RasterGrid Kft.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


spencer-lunarg (approved):
LGTM - awesome seeing more formatting improvements!

artem-lunarg:
Going to do testing on additional captures. Might also wait until the next ANGLE roll. After this PR they need to update their error filters, this was discussed in our internal channel.

layers/utils/vk_layer_utils.h
```diff
@@ -405,12 +405,18 @@ static inline VkDeviceSize SafeDivision(VkDeviceSize dividend, VkDeviceSize divi
     return result;
 }
 
-// Only 32 bit fields should need a bit count
 static inline uint32_t GetBitSetCount(uint32_t field) {
     std::bitset<32> view_bits(field);
     return static_cast<uint32_t>(view_bits.count());
 }
 
+static inline uint32_t GetBitSetCount64(uint64_t field) {
```
arno-lunarg, line 413: Nit: Maybe just have templated `GetBitSetCount` and `IsSingleBitSet`, for consistency with other similar functions?

arno-lunarg (approved):
Much more readable!

layers/utils/vk_layer_utils.h
artem-lunarg: That's tricky question. In this example there are instructions for bit count, so in case current code does not generate them it's easier to update functions in case this is needed for performance (if compiler does not generate popcnt or similar, then emulation is not that cheap). Another  aspect if you change parameter (say pass 64 bit value to 32 bit parameter) you get warning from compiler. Template will just work. Could be proc and cons, but I feel low level code should be less smart. This does not mean template version could be better here, just not obvious.

artem-lunarg:
ANGEL just made a roll. Merging this. The idea to merge immediately after roll, in case of any issues there might be a day or two to fix.

layers/utils/vk_layer_utils.h
artem-lunarg: Went with the suggestion and used templates

PR #9231, opened 2025-01-16
layers: Move maxMemoryAllocationSize check

spencer-lunarg:
This was discussed a LOT in the working group last two weeks (https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/7073)

Basically this **should** be a VU, but can't because of technically it is "defined" what the app will return, but this is impossible for CTS to catch and there is no hardware vendor had a reason why an app should ever try to go above this limit, so we are adding it as a pro-active error

### DIFF
--- a/layers/core_checks/cc_device_memory.cpp
+++ b/layers/core_checks/cc_device_memory.cpp
@@ -406,13 +406,27 @@ bool CoreChecks::PreCallValidateAllocateMemory(VkDevice device, const VkMemoryAl
                                                const ErrorObject &error_obj) const {
     bool skip = false;
     if (Count<vvl::DeviceMemory>() >= phys_dev_props.limits.maxMemoryAllocationCount) {
-        skip |= LogError(
-            "VUID-vkAllocateMemory-maxMemoryAllocationCount-04101", device, error_obj.location,
-            "vkAllocateMemory: Number of currently valid memory objects is not less than maxMemoryAllocationCount (%" PRIu32 ").",
-            phys_dev_props.limits.maxMemoryAllocationCount);
+        skip |=
+            LogError("VUID-vkAllocateMemory-maxMemoryAllocationCount-04101", device, error_obj.location,
+                     "The number of currently valid memory objects (%zu) is not less than maxMemoryAllocationCount (%" PRIu32 ").",
+                     Count<vvl::DeviceMemory>(), phys_dev_props.limits.maxMemoryAllocationCount);
     }
 
     const Location allocate_info_loc = error_obj.location.dot(Field::pAllocateInfo);
+    if (IsExtEnabled(device_extensions.vk_khr_maintenance3) &&
+        pAllocateInfo->allocationSize > phys_dev_props_core11.maxMemoryAllocationSize) {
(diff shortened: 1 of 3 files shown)


layers/core_checks/cc_device_memory.cpp
```diff
@@ -406,13 +406,27 @@ bool CoreChecks::PreCallValidateAllocateMemory(VkDevice device, const VkMemoryAl
...
-            "vkAllocateMemory: Number of currently valid memory objects is not less than maxMemoryAllocationCount (%" PRIu32 ").",
-            phys_dev_props.limits.maxMemoryAllocationCount);
+        skip |=
+            LogError("VUID-vkAllocateMemory-maxMemoryAllocationCount-04101", device, error_obj.location,
+                     "The number of currently valid memory objects (%zu) is not less than maxMemoryAllocationCount (%" PRIu32 ").",
+                     Count<vvl::DeviceMemory>(), phys_dev_props.limits.maxMemoryAllocationCount);
     }
 
     const Location allocate_info_loc = error_obj.location.dot(Field::pAllocateInfo);
+    if (IsExtEnabled(device_extensions.vk_khr_maintenance3) && phys_dev_props_core11.maxMemoryAllocationSize != 0 &&
```
artem-lunarg, line 416: `maxMemoryAllocationSize !=` is it a workaround for bad driver/something else?

artem-lunarg: Asking because it's a property with `min` value of 1GB.

spencer-lunarg: ya, there is no good reason, will remove

artem-lunarg: How does min_core.json work if it does not define this value?

artem-lunarg (approved):

layers/core_checks/cc_device_memory.cpp
```diff
@@ -406,13 +406,27 @@ bool CoreChecks::PreCallValidateAllocateMemory(VkDevice device, const VkMemoryAl
...
     }
 
     const Location allocate_info_loc = error_obj.location.dot(Field::pAllocateInfo);
+    if (IsExtEnabled(device_extensions.vk_khr_maintenance3) &&
+        pAllocateInfo->allocationSize > phys_dev_props_core11.maxMemoryAllocationSize) {
+        // Discussed in https://gitlab.khronos.org/vulkan/vulkan/-/issues/4119 and finalized in
+        // https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/7073 This is a limit the Working Group feels should be alerted
+        // to the user as if it was a VU. While some drivers should report VK_ERROR_OUT_OF_DEVICE_MEMORY, each platform has small
+        // quirks to it and there is no way for us to test the drivers will return the correct VkError here.
+        LogError("UNASSIGNED-vkAllocateMemory-maxMemoryAllocationSize", device, allocate_info_loc.dot(Field::allocationSize),
```
artem-lunarg, line 422: Did you consider LogWarning? (basically previous version)

artem-lunarg: Asking based on Ziga comment that it's valid but returns VK error

PR #9234, opened 2025-01-16
tests: Enabled disabled BDA tests

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/7462

enabled these, they seem to work now, likely implicitly from the re-working of all the SPIR-V in the last year


artem-lunarg (approved):
LGTM if no issues on internal CI.

PR #9235, opened 2025-01-16
syncval: Split Device and Instance ValidationObjects

jeremyg-lunarg:
This is the start of splitting class ValidationObject into separate Device and Instance classes. Syncval has nothing happening at Instance level so it gets to go first. Once all the leaf classes are split, ValidationStateTracker and ValidationObject can be split. This seems backwards but allows it to happen incrementally rather than in 1 giant PR

### DIFF
--- a/layers/sync/sync_validation.h
+++ b/layers/sync/sync_validation.h
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 2019-2024 Valve Corporation
- * Copyright (c) 2019-2024 LunarG, Inc.
+ * Copyright (c) 2019-2025 Valve Corporation
+ * Copyright (c) 2019-2025 LunarG, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


layers/sync/sync_validation.h
```diff
@@ -27,6 +27,14 @@
 #include "sync/sync_stats.h"
 #include "sync/sync_submit.h"
 
+namespace syncval {
+// sync validation has no instance-level functionality
+class Instance : public ValidationStateTracker {
```
artem-lunarg, line 32: Got confused that comment says no instance functionality but also create Instance class. In what direction this this going to evolve?

artem-lunarg: Do we also need to derive instead of using vvl::dispatch::Instance? I did not follow previous sorry, so just random questions to get some idea.

jeremyg-lunarg: There will be a couple of phases:
1) First this will inherit from the Instance object of the state tracker, something like `vvl::state::Instance`
2) Then there will be a common state tracker and I think this class will go away.  Since syncval does nothing here, your 'device' level objects (SyncValidators) will just point at the shared state trackers Device and Instance objects

jeremyg-lunarg: Other validation types will still have to have Instance classes so that they can do their current instance level functionality.

artem-lunarg (approved):

PR #9236, opened 2025-01-16
scripts: Automated copyright notice update

aitor-lunarg:
Relevant script is `update_copyright.py`

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -1,4 +1,4 @@
-# Copyright (C) 2018-2021 The ANGLE Project Authors.
+# Copyright (C) 2018-2025 The ANGLE Project Authors.
 # Copyright (C) 2019-2025 LunarG, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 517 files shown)


aitor-lunarg:
As of now, the script does not care much about which company as long as it sees a `Copyright`. We may want to limit to just check for relevant ones? (Easily addressable)

jeremyg-lunarg:
We should limit it to the companies that are checked by check_code_format.py

Also IMO it would be much better if this only updated the copyrights on files that have been changed. There's code in check_code_format.py that could be used. In fact it might be better if this functionality was part of that script with a command line argument to enable.

spencer-lunarg:
One idea I have had was a command that would add a git hook on your local machine that would run `check_code_format.py` for you upon making a commit and something like this could not just warn you, but do it for you

jeremyg-lunarg:
> One idea I have had was a command that would add a git hook on your local machine that would run `check_code_format.py` for you upon making a commit and something like this could not just warn you, but do it for you

Auto format all the things!!!

aitor-lunarg:
Closing in favor of suggested approach. I will see if I have some cycles in the coming days to do the change, but otherwise, feel free to go ahead and do it!

PR #9241, opened 2025-01-16
gpuav: Remove legacy Alias check

spencer-lunarg:
I reworked things in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8864 that I wasn't sure if we still needed `is_descriptor_potentially_aliased`

Added tests and see it was too restrictive and we could remove it

### DIFF
--- a/tests/unit/gpu_av_descriptor_post_process.cpp
+++ b/tests/unit/gpu_av_descriptor_post_process.cpp
@@ -1,8 +1,8 @@
 /*
- * Copyright (c) 2020-2024 The Khronos Group Inc.
- * Copyright (c) 2020-2024 Valve Corporation
- * Copyright (c) 2020-2024 LunarG, Inc.
- * Copyright (c) 2020-2023 Google, Inc.
+ * Copyright (c) 2020-2025 The Khronos Group Inc.
+ * Copyright (c) 2020-2025 Valve Corporation
+ * Copyright (c) 2020-2025 LunarG, Inc.
+ * Copyright (c) 2020-2025 Google, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


tests/unit/gpu_av_descriptor_post_process.cpp
```diff
@@ -1210,6 +1210,151 @@ TEST_F(NegativeGpuAVDescriptorPostProcess, MultipleCommandBuffersSameDescriptorS
     m_default_queue->Wait();
 }
 
+TEST_F(NegativeGpuAVDescriptorPostProcess, AliasImageBindingRuntimeArray) {
```
spencer-lunarg, line 1213: we already have the Positive version of these in `gpu_av_descriptor_post_process_positive.cpp`

tests/unit/gpu_av_descriptor_post_process.cpp
```diff
@@ -1210,6 +1210,151 @@ TEST_F(NegativeGpuAVDescriptorPostProcess, MultipleCommandBuffersSameDescriptorS
...
+    vkt::Image uint_image(*m_device, image_ci, vkt::set_layout);
+    vkt::ImageView uint_image_view = uint_image.CreateView();
+
+    vkt::Buffer buffer(*m_device, 64, VK_BUFFER_USAGE_STORAGE_BUFFER_BIT);
+
+    descriptor_set.WriteDescriptorImageInfo(0, float_image_view, VK_NULL_HANDLE, VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE,
+                                            VK_IMAGE_LAYOUT_GENERAL, 0);
+    descriptor_set.WriteDescriptorImageInfo(0, uint_image_view, VK_NULL_HANDLE, VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE,
+                                            VK_IMAGE_LAYOUT_GENERAL, 1);
+    descriptor_set.WriteDescriptorImageInfo(0, float_image_view, VK_NULL_HANDLE, VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE,
```
arno-lunarg, line 1340: Since this test seems about partially bound runtime descriptor arrays, maybe this last `WriteDescriptorImageInfo` is unintended? 
Otherwise descriptor set is fully bound.
Or you just care about having VK_DESCRIPTOR_BINDING_PARTIALLY_BOUND_BIT set?

arno-lunarg (approved):
LGTM, one question

tests/unit/gpu_av_descriptor_post_process.cpp
spencer-lunarg: > Or you just care about having VK_DESCRIPTOR_BINDING_PARTIALLY_BOUND_BIT set?

it is this

if you go `layout(set = 0, binding = 0) uniform utexture2D uint_textures[3]`

by definitions, all 3 image descriptors **MUST** be valid **regardless** if they are accessed or not

Using `VK_DESCRIPTOR_BINDING_PARTIALLY_BOUND_BIT` forces us to defer to the GPU-AV as now these 3 images don't have to be valid unless accessed

PR #9252, opened 2025-01-17
vvl: Add Tracy GPU profiling

arno-lunarg:
Add initial support for Tracy GPU profiling
=> Allows to get GPU timings for GPU commands. For now, only draws, dispatches and trace rays commands are profiled.

Some implementation details:

GPU profiling relies on timestamp queries. Timestamps have to be added to command buffers, their status is then queried from the host, and available queried are reset after inspection, with a call to `vkCmdResetQueryPool`.
A worker thread is in charge of scanning query results, doing it on the main threads showed serious performance issues.
From testing against GFXR traces threading seems to work, still need to test on live games. A review on this threading part would be especially appreciated!

And if you can time the time to pull this PR and test it locally, that would be awesome, since no new CI tests are added for this.

### DIFF
--- a/layers/state_tracker/state_tracker.cpp
+++ b/layers/state_tracker/state_tracker.cpp
@@ -43,6 +43,8 @@
 #include "chassis/chassis_modification_state.h"
 #include "spirv-tools/optimizer.hpp"
 
+#include "chassis/chassis.h"
+
 ValidationStateTracker::~ValidationStateTracker() { DestroyObjectMaps(); }
 
 // NOTE:  Beware the lifespan of the rp_begin when holding  the return.  If the rp_begin isn't a "safe" copy, "IMAGELESS"
@@ -697,6 +699,34 @@ void ValidationStateTracker::PostCallRecordCreateDevice(VkPhysicalDevice gpu, co
 
     // finish setup in the object representing the device
     device_state->PostCreateDevice(pCreateInfo, record_obj.location);
+
+#if defined(VVL_TRACY_GPU)
+    std::vector<VkTimeDomainKHR> time_domains;
+    uint32_t time_domain_count = 0;
+    VkResult result = DispatchGetPhysicalDeviceCalibrateableTimeDomainsEXT(gpu, &time_domain_count, nullptr);
+    assert(result == VK_SUCCESS);
+    time_domains.resize(time_domain_count);
+    result = DispatchGetPhysicalDeviceCalibrateableTimeDomainsEXT(gpu, &time_domain_count, time_domains.data());
+    assert(result == VK_SUCCESS);
+
+    bool found_tracy_required_time_domain = false;
+    for (VkTimeDomainEXT time_domain : time_domains) {
+#if defined(VK_USE_PLATFORM_WIN32_KHR)
+        if (time_domain == VK_TIME_DOMAIN_QUERY_PERFORMANCE_COUNTER_EXT) {
+            found_tracy_required_time_domain = true;
+            break;
+        }
+#else
+        if (time_domain == VK_TIME_DOMAIN_CLOCK_MONOTONIC_RAW_EXT) {
+            found_tracy_required_time_domain = true;
+            break;
+        }
+#endif
+    }
+    (void)found_tracy_required_time_domain;
+    assert(found_tracy_required_time_domain);
+
+#endif
 }
 
 std::shared_ptr<vvl::Queue> ValidationStateTracker::CreateQueue(VkQueue handle, uint32_t family_index, uint32_t queue_index,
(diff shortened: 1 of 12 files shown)


layers/vulkan/generated/validation_object.cpp
```diff
@@ -3,9 +3,9 @@
 
 /***************************************************************************
  *
- * Copyright (c) 2015-2024 The Khronos Group Inc.
- * Copyright (c) 2015-2024 Valve Corporation
- * Copyright (c) 2015-2024 LunarG, Inc.
+ * Copyright (c) 2015-2025 The Khronos Group Inc.
```
arno-lunarg, line 6: Those noticed got updated because `layer_chassis_generator.py` was updated

layers/vulkan/generated/chassis.cpp
```diff
@@ -3483,6 +3489,7 @@ VKAPI_ATTR void VKAPI_CALL CmdDraw(VkCommandBuffer commandBuffer, uint32_t verte
     }
     {
         VVL_ZoneScopedN("Dispatch");
+        VVL_TracyVkZone(GetTracyVkCtx(), commandBuffer, "gpu_vkCmdDraw");
```
artem-lunarg, line 3492: What's the purpose of such fine grained profiling (draw level command). Is it mostly GPU-AV?

I'm a bit skeptical about such fine grained (single draw command profiling). Draw commands always run in groups (unusual to see dozens on them run in parallel), in that case such profiling deserializes that grouping.  Also profiling tools light Radeon GPU profiling/Nsight not only do this but also provide visualization (Radeon profiler especially good at showing how commands run in parallel, so can be used to validate if current impl deserializes command, but maybe Nsight too). So yeah, I'm affraid there could be significant difference between what the above tools show comparing to this method for non-trivial workloads. It would be nice to do the comparison. It could be a good idea to profile Draws based on render pass granularity.

arno-lunarg: Yes primary goal is to measure impact of GPU AV instrumentation on in pipeline execution.
And very good point on serialization, I did not think of that, thank you for bringing this up !
Tracy offers no other way to profile GPU workloads than the TracyVkzone objects, that rely on RAII to insert begin and end timestamps. So its API promotes per cmd profiling. 
I will try to hack it to profile render passes instead.

layers/chassis/chassis.h
```diff
@@ -20,6 +20,8 @@
  ****************************************************************************/
 #include <vulkan/vulkan.h>
 
+#include "profiling/profiling.h"
```
spencer-lunarg, line 23: why is this needed in this file (trying to prevent everyone having to pull this in if it is in the header)

layers/state_tracker/state_tracker.cpp
```diff
@@ -4163,6 +4193,63 @@ void ValidationStateTracker::PostCallRecordCreateInstance(const VkInstanceCreate
...
+        vku::FindStructInPNextChain<VkPhysicalDeviceHostQueryResetFeatures>(pCreateInfo->pNext));
+    if (host_query_reset_feature) {
+        host_query_reset_feature->hostQueryReset = VK_TRUE;
+    } else {
+        VkPhysicalDeviceHostQueryResetFeatures new_host_query_reset_feature = vku::InitStructHelper();
+        new_host_query_reset_feature.hostQueryReset = VK_TRUE;
+        vku::AddToPnext(*modified_create_info, new_host_query_reset_feature);
+    }
+
+#else
```
spencer-lunarg, line 4235: why are we leaving this code here? what is the use of a `#if 1` (I have no idea when to use this code or not)

layers/vulkan/generated/chassis.cpp
artem-lunarg: There is another aspect. The reason we add scopes for CPU profiling because that's how tracy works, but we do this so in the end can use Tracy App to view results. But GPU profilers don't need this intermediate step of modifying code, you use GPU profiling APP directly to view all data without any need of registration in the source code. The GPU profilers also more efficient, tracy might add some small overhead but GPU profilers don't insert any commands and register timespamps in GPU frontend - they are at least microsecond precise. So that's my first impression it's wrong tool. But I did not check if people do this. If you know about similar use cases please provide reference!

arno-lunarg: Yes of course Tracy won't be as precise or efficient as IHV profiler. But it has the merit of being portable across OS/GPU, and directly be integrated with the CPU profiler. Having everything in one spot will help coming up with scripts to streamline the process of profiling and comparing traces. We don't want number precise up to nanoseconds level, the idea is to have an easy way to see if some new code is showing a noticeable regression.

layers/chassis/chassis.h
arno-lunarg: mistake, mistake!

layers/vulkan/generated/chassis.cpp
artem-lunarg: > We don't want number precise up to nanoseconds level

My point was not about precision, that only one perk, I'm okay if tracy result at least roughtly correllate and not totally off. Concern about separate commands. Maybe work with bigger scopes: submission, command buffer, render pass, debug label scopes. Some interactive apps have built in GPU profuilers too, but that's always on bigger scopes.

arno-lunarg: Ah sorry! Yes doing broader scope profiling seems more appropriate, I need to try that

artem-lunarg: I quickly checked Tracy documentation on GPU profiling, it looks good and timestamps is the method that is used by most apps that want build-in gpu profilers. Was positively surprised. But we need to do something with per-draw profiling, that's a big no go. I think the question here is not about being imprecise but completely off numbers. Single draw call it's easily could be < 1microsecond, and measurement error can be easily greater than this. But bigger problem that gpu does not run single draw call in complex scenarios, it's not correct to apply CPU model that profiles  relatively serial execution of commands to GPU.

If we go in bigger scopes it can be a typical build-in GPU profiler system. In addition to other scopes,  debug labels can be good granularity, if apps support them. Apps usually assign them to logical blocks that represent separate techniques.

arno-lunarg:
I now only profile rendering zones as defined by `vkCmd<Begin, End>RengerPass` and `VkCmd<Begin, End>Rendering`

PR #9253, opened 2025-01-17
sync: Make the first step to validate ray tracing API

artem-lunarg:
Adds the first syncval ray tracing related validation. The goal is to validate ray tracing memory accesses for the next SDK.

### DIFF
--- a/layers/sync/sync_validation.cpp
+++ b/layers/sync/sync_validation.cpp
@@ -24,6 +24,7 @@
 #include "sync/sync_image.h"
 #include "state_tracker/buffer_state.h"
 #include "utils/convert_utils.h"
+#include "utils/ray_tracing_utils.h"
 #include "vk_layer_config.h"
 
 SyncValidator::~SyncValidator() {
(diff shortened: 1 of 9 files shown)


layers/sync/sync_validation.cpp
```diff
@@ -3572,6 +3573,85 @@ void SyncValidator::PostCallRecordGetSwapchainImagesKHR(VkDevice device, VkSwapc
     }
 }
 
+bool SyncValidator::PreCallValidateCmdBuildAccelerationStructuresKHR(
```
spencer-lunarg, line 3576: are you going to add host side `vkBuildAccelerationStructuresKHR` as well?

artem-lunarg: I did not think about it yet, it depends if that's something syncval can track. If that's just CPU memory (not VkBuffer) then probably no, can't track reliably accesses there.

spencer-lunarg: I guess this is where my knowledge falls short, but does syncval need to track the memory was updated 

like incase someone went

```
RecordCB();
vkBuildAccelerationStructuresKHR()
Submit();
```

and there is some race condition there... honestly I am not aware of the RTX sync rules around when you are allowed to build a AS or not

artem-lunarg: SyncVal validates memory accesses as they are defined by Vulkan API. like vkCmdCopyBuffer specifies region that copy touches. It can be one direction to try to track virtual memory pages and accesses but it's looks like a different direction. more what TSAN or similar tools do, but even they make mistakes. So my current intuition SyncVal has to do it's best what is defined by the API, not every accesses.

You are right we can miss races from host memory accesses (again, I did not check details of vkBuildAccelerationStructuresKHR). The goal not to produce false-positives (error that did not happen), it's okay to have false-negatives (error but we did not know about).

artem-lunarg: We have in the list of known limitations that host accesses are not supported.

artem-lunarg: Sorry, I think we had that but it was probably lost during documentation update. Will add.

spencer-lunarg (approved):

layers/sync/sync_validation.cpp
```diff
@@ -3572,6 +3573,85 @@ void SyncValidator::PostCallRecordGetSwapchainImagesKHR(VkDevice device, VkSwapc
...
+        const auto &info = pInfos[i];
+        const auto scratch_buffers = GetBuffersByAddress(info.scratchData.deviceAddress);
+        if (scratch_buffers.empty()) {
+            continue;
+        }
+        if (scratch_buffers.size() > 1) {
+            continue;  // postpone handling this case until syncval memory aliasing support is ready
+        }
+        const VkDeviceSize scratch_size = rt::ComputeScratchSize(rt::BuildType::Device, device, info, ppBuildRangeInfos[i]);
+        const vvl::Buffer &scratch_buffer = *scratch_buffers[i];
```
arno-lunarg, line 3595: Shouldn't it be scratch_buffers[0]?

arno-lunarg (commented):
I will try to finish this review by monday, I guess since I have added good chunks of RT validation I might be able to do a more detailed review

layers/sync/sync_validation.cpp
```diff
@@ -3572,6 +3573,85 @@ void SyncValidator::PostCallRecordGetSwapchainImagesKHR(VkDevice device, VkSwapc
...
+        // Skip invalid configurations
+        {
+            const sparse_container::range<VkDeviceSize> scratch_range(info.scratchData.deviceAddress,
+                                                                      info.scratchData.deviceAddress + scratch_size);
+            if (!scratch_buffer.DeviceAddressRange().includes(scratch_range)) {
+                continue;  // [core validation check]: invalid scratch range
+            }
+        }
+
+        const ResourceAccessRange range = MakeRange(scratch_buffer.deviceAddress - info.scratchData.deviceAddress, scratch_size);
```
arno-lunarg, line 3606: We have `scratch_buffer.deviceAddress` <= `info.scratchData.deviceAddress` so I think you intended to have range begin be `info.scratchData.deviceAddress - scratch_buffer.deviceAddress`?

layers/sync/sync_validation.cpp
```diff
@@ -3572,6 +3573,85 @@ void SyncValidator::PostCallRecordGetSwapchainImagesKHR(VkDevice device, VkSwapc
...
```
arno-lunarg, line 3648: Same comment as above here

arno-lunarg (commented):
So cool to see the start of sync val for ray tracing!!!
I spotted a 3 things that, at least on a Saturday after waking up not so long ago, look like bugs

layers/sync/sync_validation.cpp
artem-lunarg: bad mistake, good catch!

artem-lunarg: Updated

arno-lunarg (approved):
LGTM!
If you want other "basic" ray tracing positive tests, IMO good ones to take are:
PositiveRayTracing, BuildAccelerationStructuresList
PositiveRayTracing, AccelerationStructuresOverlappingMemory

Or actually all of them, but I should refactor the ray tracing positive test file then

PR #9256, opened 2025-01-17
printf: Don't do work when there is no printf instruction

spencer-lunarg:
Found using `VK_LAYER_PRINTF_ONLY_PRESET` slows down things way too much by default. With this change, find that when an app uses no printf in their shader (because assuming only 1 of there many pipelines will even have it) you regain near "normal" FPS again

note - `VK_LAYER_PRINTF_ONLY_PRESET` turns off everything but DebugPrintf just like vkconfig does

### DIFF
--- a/layers/gpuav/instrumentation/gpuav_instrumentation.cpp
+++ b/layers/gpuav/instrumentation/gpuav_instrumentation.cpp
@@ -1,6 +1,6 @@
-/* Copyright (c) 2020-2024 The Khronos Group Inc.
- * Copyright (c) 2020-2024 Valve Corporation
- * Copyright (c) 2020-2024 LunarG, Inc.
+/* Copyright (c) 2020-2025 The Khronos Group Inc.
+ * Copyright (c) 2020-2025 Valve Corporation
+ * Copyright (c) 2020-2025 LunarG, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -305,8 +305,13 @@ void PreCallSetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer 
     // bindings of the instrumentation descriptor set
     assert(gpuav.instrumentation_bindings_.size() == 8);
 
+    // If nothing was updated, we don't want bind anything
+    if (!last_bound.WasInstrumented()) return;
+
     if (gpuav.gpuav_settings.debug_printf_enabled) {
         if (!debug_printf::UpdateInstrumentationDescSet(gpuav, cb_state, instrumentation_desc_set, bind_point, loc)) {
+            // TODO - need cleaner way to indicate if we want to return because of an error or because we want to save from doing
+            // unnecessary work
             return;
         }
     }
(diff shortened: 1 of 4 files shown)


layers/gpuav/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -305,15 +305,23 @@ void PreCallSetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer
     // bindings of the instrumentation descriptor set
     assert(gpuav.instrumentation_bindings_.size() == 8);
 
-    if (gpuav.gpuav_settings.debug_printf_enabled) {
+    bool updated_desc_set = false;
```
arno-lunarg, line 308: Why not just go `if (!last_bound.pipeline_state.instrumentation_data.was_instrumented) return;`?

arno-lunarg: (and equivalent for shader objects)

arno-lunarg (changes requested):
Why not just go `if (!last_bound.pipeline_state.instrumentation_data.was_instrumented) return;`?

spencer-lunarg:
> It shows that we still have problems with the architecture : /

Honestly I don't know what to change?

1. We need to instrument at `PreCallRecordPipeline`
2. We need to save the state at `PostCallRecordPipeline`
3. We need to refer to the data at Draw time to know if it is to skip or not

So we have 2 spots (between 1-and-2 and 2-and-3) to hold this state, not sure how else to get around it

layers/gpuav/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -305,15 +305,23 @@ void PreCallSetupShaderInstrumentationResources(Validator &gpuav, CommandBuffer
...
+    if (gpuav.gpuav_settings.debug_printf_enabled && last_bound.HasDebugPrintf()) {
```
spencer-lunarg, line 309: > Why not just go if (!last_bound.pipeline_state.instrumentation_data.was_instrumented) return; in <Pre, Post>CallSetupShaderInstrumentationResources?

If you have both GPU-AV and DebugPrintf checks, then we will not have added the GPU-AV part of things

layers/gpuav/instrumentation/gpuav_shader_instrumentor.cpp
```diff
@@ -306,7 +306,7 @@ void GpuShaderInstrumentor::PreCallRecordShaderObjectInstrumentation(
     const uint32_t unique_shader_id = unique_shader_module_id_++;
     const bool is_shader_instrumented =
         InstrumentShader(vvl::make_span(static_cast<const uint32_t *>(create_info.pCode), create_info.codeSize / sizeof(uint32_t)),
-                         unique_shader_id, instrumentation_dsl, create_info_loc, instrumented_spirv);
+                         unique_shader_id, instrumentation_dsl, create_info_loc, instrumentation_data.status, instrumented_spirv);
```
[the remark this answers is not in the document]
spencer-lunarg: > std::vector<uint32_t> &instrumented_spirv = instrumentation_data.instrumented_spirv;

It does, the reason I can't pass in `instrumentation_data` is because it is different between Pipeline and Shader Objects

layers/gpuav/instrumentation/gpuav_instrumentation.cpp
arno-lunarg: If pipeline was not instrumented, there is nothing to do? It feels that is this is not true, we messed up somewhere

arno-lunarg (approved):

PR #9261, opened 2025-01-17
layers: Add VK_KHR_maintenance8

spencer-lunarg:
Add checks for VK_KHR_maintenance8

### DIFF
--- a/layers/core_checks/cc_copy_blit_resolve.cpp
+++ b/layers/core_checks/cc_copy_blit_resolve.cpp
@@ -31,6 +31,7 @@
 #include "state_tracker/buffer_state.h"
 #include "state_tracker/device_state.h"
 #include "generated/dispatch_functions.h"
+#include "utils/vk_layer_utils.h"
 
 struct ImageRegionIntersection {
     VkImageSubresourceLayers subresource = {};
@@ -1467,6 +1468,27 @@ bool CoreChecks::ValidateCopyImageCommon(HandleT handle, const vvl::Image &src_i
     return skip;
 }
 
+// From table in spec vkspec.html#formats-compatible-zs-color
+// TODO - We should generate the spec and move to format helper in VUL
+static bool IsCompatibleDepthStencilWithColor(VkFormat color_format, VkFormat ds_format) {
+    switch (ds_format) {
+        case VK_FORMAT_D32_SFLOAT:
+        case VK_FORMAT_D32_SFLOAT_S8_UINT:
+            return IsValueIn(color_format, {VK_FORMAT_R32_SFLOAT, VK_FORMAT_R32_SINT, VK_FORMAT_R32_UINT});
+        case VK_FORMAT_X8_D24_UNORM_PACK32:
+        case VK_FORMAT_D24_UNORM_S8_UINT:
+            return IsValueIn(color_format, {VK_FORMAT_R32_SFLOAT, VK_FORMAT_R32_SINT, VK_FORMAT_R32_UINT});
+        case VK_FORMAT_D16_UNORM:
+        case VK_FORMAT_D16_UNORM_S8_UINT:
+            return IsValueIn(color_format, {VK_FORMAT_R16_SFLOAT, VK_FORMAT_R16_UNORM, VK_FORMAT_R16_SNORM, VK_FORMAT_R16_UINT,
+                                            VK_FORMAT_R16_SINT});
+        case VK_FORMAT_S8_UINT:
+            return IsValueIn(color_format, {VK_FORMAT_R8_UINT, VK_FORMAT_R8_SINT, VK_FORMAT_R8_UNORM, VK_FORMAT_R8_SNORM});
+        default:
+            return false;
+    }
+}
+
 template <typename RegionType>
 bool CoreChecks::ValidateCmdCopyImage(VkCommandBuffer commandBuffer, VkImage srcImage, VkImageLayout srcImageLayout,
                                       VkImage dstImage, VkImageLayout dstImageLayout, uint32_t regionCount,
@@ -1620,12 +1642,13 @@ bool CoreChecks::ValidateCmdCopyImage(VkCommandBuffer commandBuffer, VkImage src
         const VkImageAspectFlags dst_aspect = dst_subresource.aspectMask;
         if (!vkuFormatIsMultiplane(src_format) && !vkuFormatIsMultiplane(dst_format)) {
             // If neither image is multi-plane the aspectMask member of src and dst must match
-            if (src_aspect != dst_aspect) {
+            if (src_aspect != dst_aspect && !enabled_features.maintenance8) {
                 vuid = is_2 ? "VUID-VkCopyImageInfo2-srcImage-01551" : "VUID-vkCmdCopyImage-srcImage-01551";
-                skip |= LogError(vuid, all_objlist, src_subresource_loc.dot(Field::aspectMask), "(%s) does not match %s (%s).",
-                                 string_VkImageAspectFlags(src_aspect).c_str(),
-                                 dst_subresource_loc.dot(Field::aspectMask).Fields().c_str(),
-                                 string_VkImageAspectFlags(dst_aspect).c_str());
+                skip |= LogError(
+                    vuid, all_objlist, src_subresource_loc.dot(Field::aspectMask),
+                    "(%s) does not match %s (%s). (This can be allowed in some cases if maintenance8 feature is enabled)",
+                    string_VkImageAspectFlags(src_aspect).c_str(), dst_subresource_loc.dot(Field::aspectMask).Fields().c_str(),
+                    string_VkImageAspectFlags(dst_aspect).c_str());
             }
         } else {
             // Source image multiplane checks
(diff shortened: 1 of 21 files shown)


tests/unit/copy_buffer_image.cpp
```diff
@@ -3234,3 +3234,141 @@ TEST_F(NegativeCopyBufferImage, CopyToBufferWithoutMemoryBound) {
 
     m_command_buffer.End();
 }
+
+TEST_F(NegativeCopyBufferImage, CopyColorToDepthMaintenacne8AspectMask) {
```
arno-lunarg, line 3238:
```suggestion
TEST_F(NegativeCopyBufferImage, CopyColorToDepthMaintenance8AspectMask) {
```

tests/unit/shader_spirv.cpp
```diff
@@ -2606,6 +2606,48 @@ TEST_F(NegativeShaderSpirv, ScalarBlockLayoutShaderCache) {
...
+    %v4float = OpTypeVector %float 4
+       %main = OpFunction %void None %4
+          %6 = OpLabel
+         %12 = OpLoad %9 %tex
+         %21 = OpImageSampleExplicitLod %v4float %12 %15 Lod|Offset %float_0 %19
+               OpReturn
+               OpFunctionEnd
+    )";
+
+    m_errorMonitor->SetAllowedFailureMsg("VUID-StandaloneSpirv-Offset-04663");  // TODO - Remove with new spirv-val
```
arno-lunarg, line 2645: Do you have an internal note for that?

tests/unit/sync_object_positive.cpp
```diff
@@ -2370,6 +2370,90 @@ TEST_F(PositiveSyncObject, KhronosTimelineSemaphoreExample) {
...
+    vkt::CommandPool compute_pool(*m_device, compute_only_family.value());
+    vkt::CommandBuffer compute_cb(*m_device, compute_pool);
+
+    vkt::Buffer buffer(*m_device, 256, VK_BUFFER_USAGE_TRANSFER_DST_BIT);
+
+    VkBufferMemoryBarrier barrier = vku::InitStructHelper();
+    barrier.srcAccessMask = VK_ACCESS_SHADER_READ_BIT;
+    barrier.dstAccessMask = VK_ACCESS_TRANSFER_WRITE_BIT;
+    barrier.srcQueueFamilyIndex = 0;
+    barrier.dstQueueFamilyIndex = 0;
```
arno-lunarg, line 2445: I don't have the details, but since this test is about `QUEUE_FAMILY_OWNERSHIP_TRANSFER`, the src and dst queue family indices should maybe be different?

arno-lunarg (approved):

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -1467,6 +1467,28 @@ bool CoreChecks::ValidateCopyImageCommon(HandleT handle, const vvl::Image &src_i
...
+    switch (ds_format) {
+        case VK_FORMAT_D32_SFLOAT:
+        case VK_FORMAT_D32_SFLOAT_S8_UINT:
+            return color_format == VK_FORMAT_R32_SFLOAT || color_format == VK_FORMAT_R32_SINT || color_format == VK_FORMAT_R32_UINT;
+        case VK_FORMAT_X8_D24_UNORM_PACK32:
+        case VK_FORMAT_D24_UNORM_S8_UINT:
+            return color_format == VK_FORMAT_R32_SFLOAT || color_format == VK_FORMAT_R32_SINT || color_format == VK_FORMAT_R32_UINT;
+        case VK_FORMAT_D16_UNORM:
+        case VK_FORMAT_D16_UNORM_S8_UINT:
+            return color_format == VK_FORMAT_R16_SFLOAT || color_format == VK_FORMAT_R16_UNORM ||
```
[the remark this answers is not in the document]
spencer-lunarg: I think the goal was to have this in VUL (which wouldn't have `IsValueIn`) but that effort to generate this from the spec might be a while, so gonna just using `IsValueIn` here

tests/unit/shader_spirv.cpp
spencer-lunarg: yes, once https://github.com/KhronosGroup/SPIRV-Tools/pull/5951 is merged, I will be doing a spirv-tools dep update and will make sure to purge this

tests/unit/sync_object_positive.cpp
spencer-lunarg: going to add this, create an issue, need extra eyes from the sync val master himself here

(updating test to show it is a test where using the `QUEUE_FAMILY_OWNERSHIP_TRANSFER` as a no-op still works)

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -1663,6 +1686,60 @@ bool CoreChecks::ValidateCmdCopyImage(VkCommandBuffer commandBuffer, VkImage src
             }
         }
 
+        if (enabled_features.maintenance8) {
+            const VkImageAspectFlags both_depth_and_stencil = VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT;
+            const VkImageAspectFlags valid_aspect_mask = VK_IMAGE_ASPECT_COLOR_BIT | both_depth_and_stencil;
+            if (src_aspect == VK_IMAGE_ASPECT_COLOR_BIT) {
+                if ((dst_aspect & valid_aspect_mask) == 0) {
+                    vuid = is_2 ? "VUID-VkCopyImageInfo2-srcSubresource-10209" : "VUID-vkCmdCopyImage-srcSubresource-10209";
```
artem-lunarg, line 1694: Can't see how 10209 can be validated here. Since maintenance8 is enabled in this branch then DEPTH/STENCIL are allowed as dst aspect.  I think logic should start with checking maintenance8 not enabled: 
`if maintenance8 not enable and if src_aspect is color then if dst_aspect is depth/stencil -> report error`.

layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -1746,13 +1823,31 @@ bool CoreChecks::ValidateCmdCopyImage(VkCommandBuffer commandBuffer, VkImage src
     // because because both texels are 4 bytes in size.
     if (!vkuFormatIsMultiplane(src_format) && !vkuFormatIsMultiplane(dst_format)) {
         const char *compatible_vuid = is_2 ? "VUID-VkCopyImageInfo2-srcImage-01548" : "VUID-vkCmdCopyImage-srcImage-01548";
-        // Depth/stencil formats must match exactly.
-        if (vkuFormatIsDepthOrStencil(src_format) || vkuFormatIsDepthOrStencil(dst_format)) {
+        const bool is_src_depth_stencil = vkuFormatIsDepthOrStencil(src_format);
+        const bool is_dst_depth_stencil = vkuFormatIsDepthOrStencil(dst_format);
+        if (is_src_depth_stencil && is_dst_depth_stencil) {
             if (src_format != dst_format) {
```
artem-lunarg, line 1829: Question. 01548 is defined through `size-compatible` term. It seems `Size Compatibility` section does not define what it means if both formats are depth-stencil. Then `48.1.8. Format Compatibility Classes` says `Each depth/stencil format is only compatible with itself` but it's not clear for me if format `compatibility` it's the same as format `size compatibility`. ??

spencer-lunarg: So the `Size Compatibility` section is a big spec can-of-worms fixing right now.

From what I can tell, what am i checking here is what was intended by maintenance8

spencer-lunarg: > but it's not clear for me if format compatibility it's the same as format size compatibility

This is a known spec issue that needs to be clarified

tests/unit/sync_object_positive.cpp
artem-lunarg: SyncVal from what I know does not have support for detection of races related to ownership transfer. That was on near-future todo list but it's again postponed because the issue we thought is about missing support in syncval went away when we implemented timeline semaphores.

spencer-lunarg:
mingw was just a cache failure

layers/sync/sync_utils.h
```diff
@@ -82,22 +82,26 @@ struct MemoryBarrier {
     VkAccessFlags2 srcAccessMask;
     VkPipelineStageFlags2 dstStageMask;
     VkAccessFlags2 dstAccessMask;
+    VkDependencyFlags dependencyFlags = 0;
```
artem-lunarg, line 85: `MemoryBarrier` and other structures in this are wrappers around regular Vulkan barrier structures and the main purpose they unify Sync1/Sync2 types, so it's possible to write code without using templates (where template parameter is eitehr Sync1 or Sync2 type). `dependecyFlag` does not look a good place here, Vulkan barrier structures does not have this member.

If we want to get this in quickly, I'm fine with this approach. If needed I can play with this solution and try to move out this field (the simplest idea is to pass this as separate parameter).

layers/core_checks/cc_copy_blit_resolve.cpp
artem-lunarg: Yes, I saw this, that's between color<->depth/stencil.

Issue #9268, opened 2025-01-20
Cleanup of transfer flags for VK_KHR_maintenance8


spencer-lunarg:
From https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9261#discussion_r1922662566 

We should explore better options for checking for how we do check for `VK_DEPENDENCY_QUEUE_FAMILY_OWNERSHIP_TRANSFER_USE_ALL_STAGES_BIT_KHR`

For `PositiveSyncObject.CmdPipelineBarrierQueueFamilyTransferAllStages` we should also have testing of using this flag with an actual transfer

original spec PRs (https://gitlab.khronos.org/vulkan/vulkan/-/issues/3752 and https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6563)

also note - VU fix already up for https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/7118

artem-lunarg:
Waiting for the answer https://gitlab.khronos.org/vulkan/vulkan/-/issues/4153

spencer-lunarg:
> Waiting for the answer https://gitlab.khronos.org/vulkan/vulkan/-/issues/4153

I think the original idea of this issue is done, PR was awesome, will wait for spec issue to resolve and act accordingly

PR #9275, opened 2025-01-20
layers: Cleanup usage of ownership transfer flag

artem-lunarg:
Addresses https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9268

### DIFF
--- a/layers/core_checks/cc_synchronization.cpp
+++ b/layers/core_checks/cc_synchronization.cpp
@@ -25,7 +25,6 @@
 #include <vulkan/vk_enum_string_helper.h>
 #include "core_checks/cc_synchronization.h"
 #include "core_checks/core_validation.h"
-#include "generated/vk_extension_helper.h"
 #include "sync/sync_utils.h"
 #include "sync/sync_vuid_maps.h"
 #include "generated/enum_flag_bits.h"
@@ -973,7 +972,7 @@ bool CoreChecks::ValidateRenderPassPipelineBarriers(const Location &outer_loc, c
                          dep_info.bufferMemoryBarrierCount, state.active_subpass, FormatHandle(state.rp_handle).c_str());
     }
     for (uint32_t i = 0; i < dep_info.imageMemoryBarrierCount; ++i) {
-        const auto img_barrier = ImageBarrier(dep_info.pImageMemoryBarriers[i], dep_info.dependencyFlags);
+        const auto img_barrier = ImageBarrier(dep_info.pImageMemoryBarriers[i]);
         const Location barrier_loc = outer_loc.dot(Struct::VkImageMemoryBarrier2, Field::pImageMemoryBarriers, i);
 
         skip |= state.ValidateStage(barrier_loc, img_barrier.srcStageMask, img_barrier.dstStageMask);
@@ -1075,13 +1074,9 @@ bool CoreChecks::ValidatePipelineStageFeatureEnables(const LogObjectList &objlis
 }
 
 bool CoreChecks::ValidatePipelineStage(const LogObjectList &objlist, const Location &stage_mask_loc, VkQueueFlags queue_flags,
-                                       VkPipelineStageFlags2KHR stage_mask, VkDependencyFlags dependency_flags) const {
+                                       VkPipelineStageFlags2KHR stage_mask) const {
     bool skip = false;
-    // VK_KHR_maintenance8 provides way for pipeline barriers to not need this check, but add check here in VVL as this is where
-    // everything is funneled to
-    if ((dependency_flags & VK_DEPENDENCY_QUEUE_FAMILY_OWNERSHIP_TRANSFER_USE_ALL_STAGES_BIT_KHR) == 0) {
-        skip |= ValidateStageMasksAgainstQueueCapabilities(objlist, stage_mask_loc, queue_flags, stage_mask);
-    }
+    skip |= ValidateStageMasksAgainstQueueCapabilities(objlist, stage_mask_loc, queue_flags, stage_mask);
     skip |= ValidatePipelineStageFeatureEnables(objlist, stage_mask_loc, stage_mask);
     return skip;
 }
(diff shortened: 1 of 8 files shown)


layers/core_checks/cc_synchronization.cpp
```diff
@@ -1547,6 +1540,14 @@ static inline VkQueueFlags SubpassToQueueFlags(uint32_t subpass) {
 bool CoreChecks::ValidateSubpassDependency(const ErrorObject &error_obj, const Location &in_loc,
                                            const VkSubpassDependency2 &dependency) const {
     bool skip = false;
+
+    if (dependency.dependencyFlags & VK_DEPENDENCY_QUEUE_FAMILY_OWNERSHIP_TRANSFER_USE_ALL_STAGES_BIT_KHR) {
```
artem-lunarg, line 1544: Moved this part from `ValidateRenderPassDAG` to `ValidateSubpassDependency`. The former is more focused with validation related to "graph structure" of the subpasses.

layers/core_checks/cc_synchronization.cpp
```diff
@@ -2599,13 +2598,16 @@ bool CoreChecks::ValidateMemoryBarrier(const LogObjectList &objects, const Locat
...
     if (is_sync2) {
-        if (ownership_transfer_op != OwnershipTransferOp::acquire) {
-            skip |= ValidatePipelineStage(objects, barrier_loc.dot(Field::srcStageMask), queue_flags, barrier.srcStageMask,
-                                          barrier.dependencyFlags);
+        const bool allow_all_stages =
+            (barrier_loc.function == vvl::Func::vkCmdPipelineBarrier2 ||
+             barrier_loc.function == vvl::Func::vkCmdPipelineBarrier2KHR) &&
+            (dependency_flags & VK_DEPENDENCY_QUEUE_FAMILY_OWNERSHIP_TRANSFER_USE_ALL_STAGES_BIT_KHR) != 0;
+
+        if (ownership_transfer_op != OwnershipTransferOp::acquire || allow_all_stages) {
```
artem-lunarg, line 2606: That's the place for 09675/09676 VUIDS. It already had conditions related to `If a buffer or image memory barrier does not specify an acquire release operation` parts of the VUs, so added here the part related to USE_ALL_STAGES.

tests/unit/sync_object_positive.cpp
```diff
@@ -2370,58 +2370,6 @@ TEST_F(PositiveSyncObject, KhronosTimelineSemaphoreExample) {
     }
 }
 
-TEST_F(PositiveSyncObject, BufferBarrierStageNotSupportedByQueue) {
```
artem-lunarg, line 2373: 09675/09676 says
   if 
    a) barrier does not define ownership transfer
   or
    b) it defines ownership transfer and USE_ALL_STAGES is used.

`NegativeSyncObject.BufferBarrierStageNotSupportedByQueue` already checked part a).
Part b) should also be a negative test that uses ownership transfer to generate the same VUIDs.
The new test is `NegativeSyncObject.BufferOwnershipTransferStageNotSupportedByQueue`.

tests/unit/sync_object.cpp
```diff
@@ -3982,54 +3982,114 @@ TEST_F(NegativeSyncObject, BufferBarrierStageNotSupportedByQueue) {
...
+    AddRequiredExtensions(VK_KHR_MAINTENANCE_8_EXTENSION_NAME);
+    RETURN_IF_SKIP(Init());
 
     vkt::Buffer buffer(*m_device, 256, VK_BUFFER_USAGE_TRANSFER_DST_BIT);
 
     VkBufferMemoryBarrier barrier = vku::InitStructHelper();
-    barrier.srcAccessMask = VK_ACCESS_SHADER_READ_BIT;
+    barrier.srcAccessMask = VK_ACCESS_SHADER_WRITE_BIT;
     barrier.dstAccessMask = VK_ACCESS_TRANSFER_WRITE_BIT;
-    barrier.srcQueueFamilyIndex = 0;
```
artem-lunarg, line 4003: No need to define ownership transfer here just to check for validity of the flag usage

tests/unit/sync_object.cpp
```diff
@@ -3982,54 +3982,114 @@ TEST_F(NegativeSyncObject, BufferBarrierStageNotSupportedByQueue) {
     compute_cb.End();
 }
 
-TEST_F(NegativeSyncObject, CmdPipelineBarrierQueueFamilyTransferAllStagesFeature) {
+TEST_F(NegativeSyncObject, BufferOwnershipTransferStageNotSupportedByQueue) {
```
artem-lunarg, line 3985: Wrote this test by following stricly the spec language for 09675/09676.
This test should be changed if the spec is wrong: https://gitlab.khronos.org/vulkan/vulkan/-/issues/4153

tests/unit/sync_object_positive.cpp
spencer-lunarg: Do you feel we have a positive test that properly uses `VK_DEPENDENCY_QUEUE_FAMILY_OWNERSHIP_TRANSFER_USE_ALL_STAGES_BIT_KHR`?

spencer-lunarg (approved):

spencer-lunarg:
rebase to fix CI

tests/unit/sync_object_positive.cpp
artem-lunarg: Hm.. Good question. I think it's not really possible to come up with a *regression test* here, in the sense that when you remove `VK_DEPENDENCY_QUEUE_FAMILY_OWNERSHIP_TRANSFER_USE_ALL_STAGES_BIT_KHR` flag from that test then it starts failing. Replacing the flag with zero just fallbacks to the old behavior of ownership transfer that ignores the stages on one side of the barrier. And when stages are ignored it's hard to use this behavior to generate regression error.

We still have a positive test you added that the flag does not affect anything if there is no ownershp transfer operation.
I'll re-add a test similar to the one I removed, with valid ownership transfer that uses a flag.

artem-lunarg: Added positive test `PositiveSyncObject.OwnershipTransferUseAllStages` based on `NegativeSyncObject.BufferOwnershipTransferStageNotSupportedByQueue`.

PR #9283, opened 2025-01-21
layers: Update how we do format size checks

spencer-lunarg:
Uses 

- https://github.com/KhronosGroup/Vulkan-Utility-Libraries/pull/261
- https://github.com/KhronosGroup/Vulkan-Utility-Libraries/pull/262
- https://github.com/KhronosGroup/Vulkan-Utility-Libraries/pull/263

Main aim to get rid of things like `vkuFormatElementSize` (as `element` is not a proper spec term) and stick to `texel` and `texel blocks`

Also where these are used (ex - vertex input, texel buffer, etc) have different inherent spec restrictions that the current code was assuming, but hard to track down

### DIFF
--- a/layers/core_checks/cc_buffer.cpp
+++ b/layers/core_checks/cc_buffer.cpp
@@ -50,30 +50,30 @@ bool CoreChecks::ValidateBufferViewRange(const vvl::Buffer &buffer_state, const 
 
     const VkDeviceSize &range = create_info.range;
     const VkFormat format = create_info.format;
-    const VKU_FORMAT_INFO format_info = vkuGetFormatInfo(format);
-    // will be 1 most likely because HW don't tend to support texel buffer for compressed formats
-    VkDeviceSize texel_per_block = static_cast<VkDeviceSize>(format_info.texel_per_block);
+    // will be 1 because  block-compressed format are not supported for Texe l Buffer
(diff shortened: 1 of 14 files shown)


layers/core_checks/cc_buffer.cpp
```diff
@@ -381,14 +381,14 @@ bool CoreChecks::PreCallValidateCreateBufferView(VkDevice device, const VkBuffer
     }
 
     if (enabled_features.texelBufferAlignment) {
-        VkDeviceSize element_size = vkuFormatElementSize(pCreateInfo->format);
```
artem-lunarg, line 384: From error message:
"If the size of a texel is a multiple of three bytes, then the size of a  single component of format is used instead"

The vairable here means either a texel block size or the size of a single component. Rename to something something related to alignment?

artem-lunarg (approved):

layers/core_checks/cc_buffer.cpp
spencer-lunarg: So I have a spec issue open - https://gitlab.khronos.org/vulkan/vulkan/-/issues/4155 

I think these 2 VUs around `VK_EXT_texel_buffer_alignment` need clarification as well as both the error message and VU statement are unclear to me

... so will update once I know what to actually check and print out

PR #9291, opened 2025-01-22
layers: More cleaup of format size compatibility

spencer-lunarg:
More cleanup around using `texel` and `texel block` correctly and consistently everywhere

### DIFF
--- a/layers/core_checks/cc_copy_blit_resolve.cpp
+++ b/layers/core_checks/cc_copy_blit_resolve.cpp
@@ -346,7 +346,6 @@ bool CoreChecks::ValidateHeterogeneousCopyData(const HandleT handle, const Regio
         IsValueIn(region_loc.function, {Func::vkCopyMemoryToImageEXT, Func::vkCopyImageToMemory, Func::vkCopyImageToMemoryEXT});
 
     const Location subresource_loc = region_loc.dot(Field::imageSubresource);
-    const VkImageAspectFlags region_aspect_mask = region.imageSubresource.aspectMask;
     if (image_state.create_info.imageType == VK_IMAGE_TYPE_1D) {
         if ((region.imageOffset.y != 0) || (region.imageExtent.height != 1)) {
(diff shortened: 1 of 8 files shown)


layers/core_checks/cc_copy_blit_resolve.cpp
```diff
@@ -569,12 +569,9 @@ bool CoreChecks::ValidateHeterogeneousCopyData(const HandleT handle, const Regio
 
     // *RowLength divided by the texel block extent width and then multiplied by the texel block size of the image must be
     // less than or equal to 2^31-1
-    const uint32_t element_size =
-        vkuFormatIsDepthOrStencil(image_format)
-            ? 0
-            : vkuFormatElementSizeWithAspect(image_format, static_cast<VkImageAspectFlagBits>(region_aspect_mask));
+    const uint32_t texel_block_size = vkuFormatTexelBlockSize(image_format);
```
artem-lunarg, line 572: The deprecated function like `vkuFormatElementSizeWithAspect` used aspectMask paremeter. The table lookup in `vkuFormatTexelBlockSize` does not depend on aspectMask. Was the previous behavior incorrect or we should still need aspect related code?

spencer-lunarg: yes, so the aspect mask was used for 2 reason

1. Depth/Stencil
2. Multi-plane

For 1, depth/stencil have restrictions all over when it can and can't be used. So doing a copy, there is a very restricted way you can do it (with VK_KHR_maintenance8) 

For 2,  same idea, for copying, you only copy a single plane and you need to know which plane is selected

For this VU (`09108`) it talks about the `texel block size`, well this is the same regardless which plane you select, so there is no need for worrying about plane aspect. Also for depth/stencil, you copy it to the entire thing, there is currently no way to "only update the stencil bytes" in a vkCmdBufferToImage call

artem-lunarg (approved):

PR #9296, opened 2025-01-23
layers: Update vvl::enumerate to return reference

artem-lunarg:
This matches behavior of `std::views::enumerate` - enumerated object is returned as a reference instead of a pointer.
Would minimize changes if we start using C++ 23 in the future.


arno-lunarg (approved):
Nice!! I think I tried to do this 2 or 3 times but gave up in front of the storm of compilation issues

artem-lunarg:
It's definitely not C++ standard compatible, you need C++ PhD for that (https://brevzin.github.io/c++/2022/12/05/enumerate/). Hopefully the calling sites will be less affected after this change if we need more fixes. We might need to add ability to `std::move` referenced element if needed, std::views::enumerate supports this.

PR #9302, opened 2025-01-23
stateless: Remove ValidatePnextPropertyStructContents()

jeremyg-lunarg:
Property pNext checking was implemented incorrectly for a long time and recently fixed by commit da1b58841. However requiring extensions enabled for querying properties is very cumbersome and contradicts this statement from the specification:

       Any component of the implementation (the loader, any enabled layers,
       and drivers) must skip over, without processing (other than reading the
       sType and pNext members) any extending structures in the chain not
       defined by core versions or extensions supported by that component.

It is possible that this could also be relaxed for other queries in the future.


scripts/generators/stateless_validation_helper_generator.py
```diff
@@ -442,6 +442,9 @@ def generateSource(self):
         property_structs = [x for x in extended_structs if x.extends == ["VkPhysicalDeviceProperties2"]]
         other_structs = [x for x in extended_structs if x not in feature_structs and x not in property_structs and x.name not in self.structsWithManualChecks]
 
+        # NOTE: property structs are no longer checked. Previously we only checked if the corresponding extension was supported
```
spencer-lunarg, line 445: maybe worth putting this above `property_structs` declaration and point to link to https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9302 for future reference

spencer-lunarg (approved):

PR #9307, opened 2025-01-24
gpuav: Improve shader map management

arno-lunarg:
Deleting an element from instrumented_shaders_map_ was too slow

### DIFF
--- a/layers/gpuav/instrumentation/gpuav_shader_instrumentor.cpp
+++ b/layers/gpuav/instrumentation/gpuav_shader_instrumentor.cpp
@@ -356,6 +356,7 @@ void GpuShaderInstrumentor::PostCallRecordCreateShadersEXT(VkDevice device, uint
         }
         if (const auto &shader_object_state = Get<vvl::ShaderObject>(pShaders[i])) {
             shader_object_state->instrumentation_data.was_instrumented = true;
+            shader_object_state->instrumentation_data.unique_shader_id = instrumentation_data.unique_shader_id;
         }
 
         instrumented_shaders_map_.insert_or_assign(instrumentation_data.unique_shader_id, VK_NULL_HANDLE, VK_NULL_HANDLE,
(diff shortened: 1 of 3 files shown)


jeremyg-lunarg (approved):

layers/state_tracker/shader_object_state.h
```diff
@@ -52,6 +52,7 @@ struct ShaderObject : public StateObject {
     // TOOD Create a shader object inherited class
     struct InstrumentationData {
         bool was_instrumented = false;
+        uint32_t unique_shader_id = 0xffff'ffff;
```
spencer-lunarg, line 55: unique_shader_id of `0` is used in `IsInstrumented` to mean "not valid"

so just use `0`, it is already defined as the "invalid" value

layers/state_tracker/pipeline_state.h
```diff
@@ -138,7 +138,7 @@ class Pipeline : public StateObject {
     // ment for GPU-AV
     struct InstrumentationData {
         // We create a VkShaderModule that is instrumented and need to delete before leaving the pipeline call
-        std::vector<VkShaderModule> instrumented_shader_module;
+        std::vector<std::pair<uint32_t, VkShaderModule>> instrumented_shader_modules;
```
spencer-lunarg, line 141: comment what this pair is

layers/gpuav/instrumentation/gpuav_shader_instrumentor.cpp
```diff
@@ -673,15 +672,10 @@ void GpuShaderInstrumentor::PostCallRecordCreateRayTracingPipelinesKHR(
...
-    auto to_erase =
-        instrumented_shaders_map_.snapshot([pipeline](const InstrumentedShader &entry) { return entry.pipeline == pipeline; });
-    for (const auto &entry : to_erase) {
-        instrumented_shaders_map_.erase(entry.first);
-    }
-
     if (auto pipeline_state = Get<vvl::Pipeline>(pipeline)) {
-        for (auto shader_module : pipeline_state->instrumentation_data.instrumented_shader_module) {
-            DispatchDestroyShaderModule(device, shader_module, pAllocator);
+        for (auto shader_module : pipeline_state->instrumentation_data.instrumented_shader_modules) {
```
spencer-lunarg, line 676: can we do the `auto [unique_shader_id, shader_module_handle]` syntax because it is a std::pair?

spencer-lunarg (approved):

PR #9308, opened 2025-01-24
build: Update to header 1.4.306

mikes-lunarg:


spencer-lunarg (approved):
another simple header... make @mikes-lunarg life too easy these days

PR #9313, opened 2025-01-25
tests: Add CreateMinimalShaders for shader object tests

spencer-lunarg:
Found creating the vertex/frag shader for many tests was not part what the we were trying to test so wrapped it up in a `CreateMinimalShaders` call

### DIFF
--- a/tests/framework/layer_validation_tests.h
+++ b/tests/framework/layer_validation_tests.h
@@ -374,6 +374,11 @@ class ShaderObjectTest : public virtual VkLayerTest {
   public:
     void InitBasicShaderObject();
     void InitBasicMeshShaderObject(APIVersion target_api_version);
+
+    // Many tests just need a basic vert/frag shader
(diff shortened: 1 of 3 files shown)


ziga-lunarg (approved):

arno-lunarg (approved):
Nit for style: I think it is easier to understand code flow if you have this new helper return the shaders it creates. Then you don't need to know that to grab them you must use internal members, you just naturally grab returned shader.
Let functional programming shine!

spencer-lunarg:
> I think it is easier to understand code flow if you have this new helper return the shaders it creates

I feel then we are back where we are now, we just reduced the typing

I think the ship has sailed and we already have things like `Init` and `InitDynamicRenderTarget();`

I made sure these are **only** for the shader object tests, just because we have like 50 tests for dynamic state and non of this even matters.. I was debating about just putting it in `InitBasicShaderObject` at first as well

PR #9316, opened 2025-01-25
layers: Fix determining device api version

### LINKED ISSUE #9277: False positive for VUID-VkPipelineLayoutCreateInfo-descriptorType-03024
**Environment:**
 - OS: Android 13
 - GPU and driver version: Adreno (TM) 620 512.490.0
 - SDK or header version if building from repo: 1.4.304.0

**Describe the Issue**

After updating from 1.3.296.0 layers, we started hitting messages like:
```
[ VUID-VkPipelineLayoutCreateInfo-descriptorType-03024 ] | MessageID = 0xeb2bfadc | vkCreatePipelineLayout(): max per-stage storage buffer bindings count (1) exceeds device maxPerStageDescriptorUpdateAfterBindStorageBuffers limit (0).
The Vulkan spec states: The total number of descriptors with a descriptorType of VK_DESCRIPTOR_TYPE_STORAGE_BUFFER and VK_DESCRIPTOR_TYPE_STORAGE_BUFFER_DYNAMIC accessible to any given shader stage across all elements of pSetLayouts must be less than or equal to VkPhysicalDeviceDescriptorIndexingProperties::maxPerStageDescriptorUpdateAfterBindStorageBuffers (https://docs.vulkan.org/spec/latest/chapters/descriptorsets.html#VUID-VkPipelineLayoutCreateInfo-descriptorType-03024)
```
And equivalents for other descriptor types, ie uniform buffer, sampled image, etc.

This makes no sense at all, as we are not utilising VK_EXT_descriptor_indexing as it is not supported on this device.

I would instead expect it to follow these guidelines from the spec, as we are not using `VK_DESCRIPTOR_SET_LAYOUT_CREATE_UPDATE_AFTER_BIND_POOL_BIT`:
```
The total number of descriptors in descriptor set layouts created without the VK_DESCRIPTOR_SET_LAYOUT_CREATE_UPDATE_AFTER_BIND_POOL_BIT bit set with a descriptorType of VK_DESCRIPTOR_TYPE_STORAGE_BUFFER and VK_DESCRIPTOR_TYPE_STORAGE_BUFFER_DYNAMIC accessible to any given shader stage across all elements of pSetLayouts must be less than or equal to VkPhysicalDeviceLimits::maxPerStageDescriptorStorageBuffers
```

**Expected behavior**
We do not get spurious validation messages for features we are not using.

**Valid Usage ID**
`VUID-VkPipelineLayoutCreateInfo-descriptorType-03024`

**Additional context**

<details>
<summary>code or terminal output</summary>

```sh
# callstacks, crashes, etc.
# EX:
Validation Error: [ VUID-vkCmdDrawMultiEXT-colorAttachmentCount-06188 ] Object 0: handle = 0x3d47e60 ...
```

</details>

ziga-lunarg:
Might close #9277, would be great if author of the issue can confirm

### DIFF
--- a/layers/chassis/dispatch_object_manual.cpp
+++ b/layers/chassis/dispatch_object_manual.cpp
@@ -270,16 +270,15 @@ Device::Device(Instance *instance, VkPhysicalDevice gpu, const VkDeviceCreateInf
     : HandleWrapper(instance->debug_report),
       settings(instance->settings),
       dispatch_instance(instance),
-      api_version(instance->api_version),
       physical_device(gpu) {
     // Get physical device limits for device
     VkPhysicalDeviceProperties device_properties = {};
     dispatch_instance->instance_dispatch_table.GetPhysicalDeviceProperties(gpu, &device_properties);
 
     // Setup the validation tables based on the application API version from the instance and the capabilities of the device driver
-    auto effective_api_version = std::min(APIVersion(device_properties.apiVersion), dispatch_instance->api_version);
+    api_version = std::min(APIVersion(device_properties.apiVersion), dispatch_instance->api_version);
 
-    extensions = DeviceExtensions(dispatch_instance->extensions, effective_api_version, pCreateInfo);
+    extensions = DeviceExtensions(dispatch_instance->extensions, api_version, pCreateInfo);
 
     InitValidationObjects();
     InitObjectDispatchVectors();


spencer-lunarg (approved):
If CI is happy, then so am I.

I still want to spend more time investigating as well, going to get older Mesa driver to test with and well as messing around more with profiles (so we can add a small test run to CI)

ziga-lunarg:
> If CI is happy, then so am I.
> 
> I still want to spend more time investigating as well, going to get older Mesa driver to test with and well as messing around more with profiles (so we can add a small test run to CI)

The problem is a test for this would not run in CI as is, we would have to change our framework.

spencer-lunarg:
> we would have to change our framework.

We can (and do) have tests that don't use the `Init()` logic, so we can still do whatever we want with "raw" vulkan for a given test

I feel the framework being designed for the 99% of test is important, these 1 tests, we could put in a new file (and test class)

like instead of `class VkTestFramework : public ::testing::Test {` we would make a new class and tests dedicated for these very aweful version edge cases

jeremyg-lunarg (approved):

PR #9330, opened 2025-01-28
layers: Fix crash in shader objects

spencer-lunarg:
for internal issue still needed to be understood


artem-lunarg (approved):

ziga-lunarg (approved):

ziga-lunarg:
I hope this does actually solve the problem temporarily, but I fear it might just crash in another place instead

spencer-lunarg:
> I hope this does actually solve the problem temporarily, but I fear it might just crash in another place instead

IMO the issue is how we did Render Passes (with that `activeRenderpass` pointer)

I created https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9334 - will take a look at it more later this week

Issue #9334, opened 2025-01-28
More robust way to access renderpass


spencer-lunarg:
Things like https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9330 happen non-zero number of times.

For the state getters (like `Get<vvl::Image>()`) we have added wrapping to check for null pointers, but for `activeRenderpass` we tend to blindly use it everywhere

1. Need to write tests to understand when/why it is sometimes null
2. Create a better interface for someone using it to know to check for it

PR #9337, opened 2025-01-28
layers: Full cleanup how we do vvl::RenderPass

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9334

so `vvl::RenderPass`

- For normal vulkan1.0 we create this at `vkCreateRenderPass` time
- For Dynamic Rendering, we create the object on the spot
- There are 2 places that consume render passes (command buffer and pipeline)
    - Both need to either consume a `VkRenderPass` handle or create a `vvl::RenderPass` on the spot
- The command buffer state had some issue tracking what is a
    - RenderPass state value (something you would find in  vkCreateRenderPass)
    - RenderPass begin value (something at vkCmdBeginRenderPass/vkCmdBeginRendering)
- I try to pass around references/null check things early like we do for the other "normal" state objects

### DIFF
--- a/layers/best_practices/bp_drawdispatch.cpp
+++ b/layers/best_practices/bp_drawdispatch.cpp
@@ -28,7 +28,7 @@ bool BestPractices::ValidateCmdDrawType(VkCommandBuffer cmd_buffer, const Locati
     bool skip = false;
     const auto cb_state = GetRead<bp_state::CommandBuffer>(cmd_buffer);
     if (const auto* pipe = cb_state->GetCurrentPipeline(VK_PIPELINE_BIND_POINT_GRAPHICS)) {
-        if (const auto rp_state = pipe->RenderPassState()) {
+        if (const auto rp_state = cb_state->active_render_pass.get()) {
             for (uint32_t i = 0; i < rp_state->create_info.subpassCount; ++i) {
                 const auto& subpass = rp_state->create_info.pSubpasses[i];
                 const auto* ds_state = pipe->DepthStencilState();
--- a/layers/core_checks/cc_cmd_buffer.cpp
+++ b/layers/core_checks/cc_cmd_buffer.cpp
@@ -833,19 +833,20 @@ bool CoreChecks::PreCallValidateCmdExecuteCommands(VkCommandBuffer commandBuffer
         }
     }
 
-    if (cb_state.activeRenderPass) {
-        if (!cb_state.activeRenderPass->UsesDynamicRendering() && cb_state.IsPrimary()) {
+    const vvl::RenderPass *rp_state = cb_state.active_render_pass.get();
+    if (rp_state) {
+        if (!rp_state->UsesDynamicRendering() && cb_state.IsPrimary()) {
             // check if first subpass
             if (cb_state.activeSubpassContents != VK_SUBPASS_CONTENTS_SECONDARY_COMMAND_BUFFERS &&
                 cb_state.activeSubpassContents != VK_SUBPASS_CONTENTS_INLINE_AND_SECONDARY_COMMAND_BUFFERS_KHR) {
                 if (cb_state.GetActiveSubpass() == 0) {
-                    const LogObjectList objlist(commandBuffer, cb_state.activeRenderPass->Handle());
+                    const LogObjectList objlist(commandBuffer, rp_state->Handle());
                     skip |= LogError("VUID-vkCmdExecuteCommands-contents-09680", objlist, error_obj.location,
                                      "contents must be set to VK_SUBPASS_CONTENTS_SECONDARY_COMMAND_BUFFERS or "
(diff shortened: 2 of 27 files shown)


layers/core_checks/cc_image.cpp
```diff
@@ -1438,15 +1439,15 @@ void CoreChecks::PostCallRecordCmdClearAttachments(VkCommandBuffer commandBuffer
...
+                        secondary, prim_cb->active_render_pass->dynamic_rendering_begin_rendering_info.renderArea,
+                        prim_cb->active_render_pass->dynamic_rendering_begin_rendering_info.layerCount, rectCount,
                         clear_rect_copy->data(), record_obj.location);
                 };
                 cb_state_ptr->cmd_execute_commands_functions.emplace_back(val_fn);
             }
         }
-    } else if (cb_state.activeRenderPass->use_dynamic_rendering == false) {
-        const VkRenderPassCreateInfo2 *renderpass_create_info = cb_state.activeRenderPass->create_info.ptr();
+    } else if (rp_state->use_dynamic_rendering == false) {
```
ziga-lunarg, line 1449: For consistency I would prefer `!rp_state->use_dynamic_rendering`

layers/core_checks/cc_pipeline.cpp
```diff
@@ -247,6 +248,69 @@ bool CoreChecks::PreCallValidateDestroyPipeline(VkDevice device, VkPipeline pipe
     return skip;
 }
 
+bool CoreChecks::ValidateCmdBindPipelineRenderPass(const vvl::CommandBuffer &cb_state, const vvl::Pipeline &pipeline_state,
+                                                   const vvl::RenderPass &rp_state, const Location &loc) const {
+    bool skip = false;
+    const auto *multisample_state = pipeline_state.MultisampleState();
+    if (!multisample_state) return skip;
```
ziga-lunarg, line 255: This might cause issues, if someone adds something at the bottom of this function and doesn't see this skip

spencer-lunarg: Let me just rename this function `ValidateCmdBindPipelineRenderPassMultistate` for clarity then

layers/best_practices/bp_drawdispatch.cpp
```diff
@@ -28,7 +28,7 @@ bool BestPractices::ValidateCmdDrawType(VkCommandBuffer cmd_buffer, const Locati
     bool skip = false;
     const auto cb_state = GetRead<bp_state::CommandBuffer>(cmd_buffer);
     if (const auto* pipe = cb_state->GetCurrentPipeline(VK_PIPELINE_BIND_POINT_GRAPHICS)) {
-        if (const auto rp_state = pipe->RenderPassState()) {
+        if (const auto rp_state = cb_state->active_render_pass.get()) {
```
spencer-lunarg, line 31: subtle, but this was a bug, added comment above `RenderPassState()` about it's usage (and checked it is good everywhere else)

layers/core_checks/cc_cmd_buffer.cpp
```diff
@@ -883,131 +883,127 @@ bool CoreChecks::PreCallValidateCmdExecuteCommands(VkCommandBuffer commandBuffer
             const LogObjectList objlist(commandBuffer, pCommandBuffers[i]);
             skip |= LogError("VUID-vkCmdExecuteCommands-pCommandBuffers-00088", objlist, cb_loc,
                              "(%s) is not VK_COMMAND_BUFFER_LEVEL_SECONDARY.", FormatHandle(pCommandBuffers[i]).c_str());
-        } else {
```
spencer-lunarg, line 886: just collapsed the `if/else` one thing here

layers/core_checks/cc_device.cpp
```diff
@@ -606,9 +605,7 @@ bool CoreChecks::PreCallValidateCmdSetDeviceMask(VkCommandBuffer commandBuffer,
     skip |= ValidateDeviceMaskToPhysicalDeviceCount(deviceMask, objlist, loc, "VUID-vkCmdSetDeviceMask-deviceMask-00108");
     skip |= ValidateDeviceMaskToZero(deviceMask, objlist, loc, "VUID-vkCmdSetDeviceMask-deviceMask-00109");
     skip |= ValidateDeviceMaskToCommandBuffer(cb_state, deviceMask, objlist, loc, "VUID-vkCmdSetDeviceMask-deviceMask-00110");
-    if (cb_state.activeRenderPass) {
```
spencer-lunarg, line 609: moved inside the function (see change right above)

artem-lunarg (approved):
LGTM assuming @ziga-lunarg feedback is addressed.

spencer-lunarg:
> LGTM assuming @ziga-lunarg feedback is addressed.

it is

PR #9340, opened 2025-01-28
state_tracker: Split Instance and Device

jeremyg-lunarg:


layers/state_tracker/state_tracker.h
```diff
@@ -1981,3 +2065,73 @@ class ValidationStateTracker : public ValidationObject {
     };
     FakeAllocator fake_memory;
 };
+
+// Get buffer size from VkBufferImageCopy / VkBufferImageCopy2KHR structure, for a given format
+template <typename RegionType>
+static inline VkDeviceSize GetBufferSizeFromCopyImage(const RegionType& region, VkFormat image_format,
```
spencer-lunarg, line 2071: you find a reason why we have this here?

spencer-lunarg (approved):
Gebben killing `ValidationStateTracker` colorized

layers/state_tracker/state_tracker.h
jeremyg-lunarg: it is used by syncval and core validation so it has to be somewhere common?

PR #9346, opened 2025-01-29
layers: Add back 06750

spencer-lunarg:
https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/7138/diffs was approved

### DIFF
--- a/tests/unit/wsi.cpp
+++ b/tests/unit/wsi.cpp
@@ -3698,14 +3698,16 @@ TEST_F(NegativeWsi, SwapchainCreateMissingMaintenanc1Feature) {
     m_errorMonitor->VerifyFound();
 }
 
-// TODO: https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/7138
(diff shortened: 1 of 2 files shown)


tests/unit/wsi.cpp
```diff
@@ -3698,14 +3698,16 @@ TEST_F(NegativeWsi, SwapchainCreateMissingMaintenanc1Feature) {
     m_errorMonitor->VerifyFound();
 }
 
-// TODO: https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/7138
-TEST_F(NegativeWsi, DISABLED_ImageCompressionPropertiesSwapchainWithoutFeature) {
+TEST_F(NegativeWsi, ImageCompressionPropertiesSwapchainWithoutFeature) {
```
artem-lunarg, line 3701: nit: test name and test description still use feature term

artem-lunarg (approved):

tests/unit/wsi.cpp
spencer-lunarg: good catch

PR #9349, opened 2025-01-29
sync: Add example of new style of syncval reporting

artem-lunarg:
This demonstrates new style of synval error messages. Only one specific use case is updated. The following work will involve converting other messages and reworking reporting helpers.

OLD:
> vkCmdCopyBuffer(): Hazard WRITE_AFTER_WRITE for dstBuffer VkBuffer 0xf443490000000006[], region 0. Access info (usage: SYNC_COPY_TRANSFER_WRITE, prior_usage: SYNC_COPY_TRANSFER_WRITE, write_barriers: VK_ACCESS_2_TRANSFER_WRITE_BIT accesses on VK_PIPELINE_STAGE_2_CLEAR_BIT stage, command: vkCmdCopyBuffer, resource: VkBuffer 0xf443490000000006[]).

NEW:
> vkCmdCopyBuffer(): WRITE_AFTER_WRITE hazard detected. vkCmdCopyBuffer writes to VkBuffer 0xf443490000000006[], which was written earlier by another vkCmdCopyBuffer command. The existed synchronization protects VK_ACCESS_2_TRANSFER_WRITE_BIT accesses at VK_PIPELINE_STAGE_2_CLEAR_BIT stage but not at VK_PIPELINE_STAGE_2_COPY_BIT stage.

p.s. still need to add `region` info to new message

### DIFF
--- a/layers/sync/sync_error_messages.cpp
+++ b/layers/sync/sync_error_messages.cpp
@@ -1,6 +1,6 @@
-/* Copyright (c) 2024 The Khronos Group Inc.
- * Copyright (c) 2024 Valve Corporation
- * Copyright (c) 2024 LunarG, Inc.
+/* Copyright (c) 2025 The Khronos Group Inc.
+ * Copyright (c) 2025 Valve Corporation
+ * Copyright (c) 2025 LunarG, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,6 +19,7 @@
 #include "sync/sync_commandbuffer.h"
 #include "sync/sync_image.h"
 #include "sync/sync_reporting.h"
+#include "sync/sync_validation.h"
 #include "state_tracker/buffer_state.h"
 #include "state_tracker/descriptor_sets.h"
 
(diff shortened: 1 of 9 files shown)


spencer-lunarg (approved):

layers/sync/sync_error_messages.cpp
```diff
@@ -148,21 +149,53 @@ std::string ErrorMessages::BufferError(const HazardResult& hazard, VkBuffer buff
...
+        const ReportUsageInfo usage_info = cb_context.GetReportUsageInfo(hazard.TagEx());
+        std::stringstream ss;
+        ss << string_SyncHazard(hazard.Hazard()) << " hazard detected. ";
+        ss << vvl::String(command) << " writes to " << validator_.FormatHandle(buffer);
+        ss << ", which was written earlier by ";
+        if (usage_info.command == command) {
+            ss << "another ";
+        }
+        ss << vvl::String(usage_info.command) << " command. ";
+        ss << "The existed synchronization protects ";
```
filnet, line 176: Did you mean "exited" ?

artem-lunarg: No, I meant currently applied synchronization. I think I'll replace "protect" wording so this will be updated. Thanks!

artem-lunarg: It should be "existing", right?

filnet: Yes, "existing" makes sense in the context.
"existed" does not sound correct and would mean "the synchronization that existed" and does not exist anymore.

PR #9350, opened 2025-01-30
layers: Move syncval format helper to shared utils

artem-lunarg:
Add helper to share implementation between logging.cpp and syncval. The helper also addresses some issues in the implementation from logging.cpp.

### DIFF
--- a/layers/sync/sync_error_messages.cpp
+++ b/layers/sync/sync_error_messages.cpp
@@ -22,45 +22,13 @@
 #include "sync/sync_validation.h"
 #include "state_tracker/buffer_state.h"
 #include "state_tracker/descriptor_sets.h"
+#include "utils/text_utils.h"
+using text::Format;
 
 #include <cassert>
 #include <cinttypes>
 #include <cstdarg>
 #include <sstream>
-
-// TODO: update algorith in logging.cpp to be similar to this version.
-// logging.cpp's vsnprintf usage in some places assumes that std::string
-// storage reserves space for null terminator but it is not guaranteed
-// (although it is in practise). vsnprintf there writes 0 terminator
-// in the char indexed as str[std::string::size()] and it should not be
-// accesses in this way by client code, only by std::string itself.
-
-static std::string Format(const char* format, ...) {
-    const int initial_max_symbol_count = 1024;
-    std::vector<char> buffer(initial_max_symbol_count + 1 /*null terminator*/);
-
-    va_list argptr;
-    va_start(argptr, format);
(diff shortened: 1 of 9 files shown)


layers/sync/sync_error_messages.cpp
```diff
@@ -22,46 +22,14 @@
...
 #include "state_tracker/descriptor_sets.h"
+#include "utils/text_utils.h"
+using text::Format;
 
 #include <cassert>
 #include <cinttypes>
 #include <cstdarg>
 #include <sstream>
 
-// TODO: update algorith in logging.cpp to be similar to this version.
```
artem-lunarg, line 31: Did this PR to remove this TODO. Two problems with original Format(): a) assumes internal std::string storage is null terminated b) performs additional resizes which are not needed if implemented differently (only one potential resize is needed).

layers/utils/text_utils.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include "text_utils.h"
+#include <algorithm>
+#include <cassert>
+#include <vector>
+
+namespace text {
+
+std::string VFormat(const char *format, va_list argptr) {
+    const int initial_max_symbol_count = 1024;
+    std::vector<char> buffer(initial_max_symbol_count + 1 /*null terminator*/);
```
artem-lunarg, line 27: This implementation works with `std::vector<char>` so does not need to make risky assumption about std::string internal representation and 0 symbol storage availability. At the end use buffer's `const char*` to initialize `std::string`.

layers/utils/vk_layer_utils.h
```diff
@@ -524,18 +524,6 @@ bool RangesIntersect(int64_t x, uint64_t x_size, int64_t y, uint64_t y_size);
 
 namespace vvl {
 
-static inline void ToLower(std::string &str) {
```
spencer-lunarg, line 527: all for removing things from the random melting pot that is `vk_layer_utils.h` to a dedicate file!

layers/utils/text_utils.cpp
spencer-lunarg: make sense! worth adding a comment to the code about this?

spencer-lunarg (approved):

layers/utils/text_utils.cpp
artem-lunarg: yes, good point. Very error prone api. When we switch to c++ 20 we should be able to get rid of this by using std::format, std::vformat.

PR #9356, opened 2025-01-30
gpuav: Fix Post Processing failing on zero bindingCount

spencer-lunarg:
Hit a crash in a trace using GPU-AV, root issue was the Descriptor Set was created with `VkDescriptorSetLayoutCreateInfo::bindingCount = 0` 

In this case, we save resources and don't create a Post Processing buffer (there are no descriptors in it!)

but then we were mixing up this case with a case where the somehow we did lose track of the descriptor

### DIFF
--- a/layers/gpuav/descriptor_validation/gpuav_descriptor_validation.cpp
+++ b/layers/gpuav/descriptor_validation/gpuav_descriptor_validation.cpp
@@ -1,6 +1,6 @@
-/* Copyright (c) 2020-2024 The Khronos Group Inc.
- * Copyright (c) 2020-2024 Valve Corporation
- * Copyright (c) 2020-2024 LunarG, Inc.
+/* Copyright (c) 2020-2025 The Khronos Group Inc.
+ * Copyright (c) 2020-2025 Valve Corporation
+ * Copyright (c) 2020-2025 LunarG, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


layers/gpuav/descriptor_validation/gpuav_descriptor_validation.cpp
```diff
@@ -231,8 +231,12 @@ void UpdateBoundDescriptors(Validator &gpuav, CommandBuffer &cb_state, VkPipelin
                 // validating the 2nd instance of it
                 continue;
             }
-            validated_desc_sets.emplace(bound_descriptor_set->VkHandle());
+
             if (!bound_descriptor_set->HasPostProcessBuffer()) {
+                if (!bound_descriptor_set->CanPostProcess()) {
```
arno-lunarg, line 236: Reading those 2 lines, it is strange to be in the condition `not bound_descriptor_set->HasPostProcessBuffer()` and yet be checking whether or not you can `PostProcess` : ?

arno-lunarg (approved):

layers/gpuav/descriptor_validation/gpuav_descriptor_validation.cpp
spencer-lunarg: Has == Is it Null
Can == Is it allowed to be Null

Not sure if it would be better to flip it?

arno-lunarg: I currently read this code as: "If bound descriptor set doesn't have a post process buffer and if bound descriptor set can be post processed, then do your stuff"

And was expecting something like "If If bound descriptor set **does** have a post process buffer and if bound descriptor set can be post processed, then do your stuff"

PR #9358, opened 2025-01-30
chassis: Split ValidationObject into Instance and Device

jeremyg-lunarg:


layers/chassis/validation_object.h
```diff
@@ -77,12 +77,12 @@ enum class ValidValue {
     NoExtension,  // trying to use a proper value, but the extension is required
 };
 
-// Layer chassis validation object base class definition
-class ValidationObject : public Logger {
+namespace vvl::vo_base {
```
jeremyg-lunarg, line 80: I really don't like this namespace name but can't come up with a better one. `vvl::validation_object`  seems way too long.

spencer-lunarg (approved):

layers/chassis/validation_object.h
spencer-lunarg:
```suggestion
// Validation Object Base
namespace vvl::vo_base {
```

I think if you do that, I'm happy

honestly `vobase` is also just as good... I'm not too worried about this name, as long as it is distinct

artem-lunarg: vvl::base 😎

PR #9087, opened 2024-12-22
Validate required dedicated allocations

ziga-lunarg:
Closes #9083

Issue #9083, opened 2024-12-22
Add missing VK_KHR_dedicated_allocation VUs


spencer-lunarg:
- [ ] VUID-VkBindBufferMemoryInfo-buffer-01444
- [ ] VUID-VkBindImageMemoryInfo-image-01445

KorsarOfficial:
did it support VKsub.buffer outwice?

spencer-lunarg:
@KorsarOfficial I am not sure what `VKsub.buffer` is

PR #9361, opened 2025-01-31
layers: Add Dedicated Buffer bind check

spencer-lunarg:
Taking https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9087 and adding just the buffer part (will add image after, but buffers are simpler)

adds `VUID-vkBindBufferMemory-buffer-01444` for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9083


tests/unit/android_hardware_buffer.cpp
```diff
@@ -1089,6 +1089,7 @@ TEST_F(NegativeAndroidHardwareBuffer, InvalidBindBufferMemory) {
     }
 
     if (mem_reqs.alignment > 1) {
+        m_errorMonitor->SetAllowedFailureMsg("VUID-vkBindBufferMemory-buffer-01444");  // required dedicated
```
ziga-lunarg, line 1092: Would it not be better to query this, and if it is required use `SetDesiredError`?

spencer-lunarg: it would, 100% just me being lazy because my Pixel 8 device doesn't need this, but the Galaxy s24 does, so just wanted something to work... feel free to steal my code and do it in #9087 - I just wanted to get traction on this and started with buffers

ziga-lunarg: Maybe add a TODO explaining what is needed to remove `SetAllowedFailureMsg`

ziga-lunarg (approved):

tests/unit/android_hardware_buffer.cpp
spencer-lunarg: I at least put the comment, most `SetAllowedFailureMsg` are missing anything (and might not even be needed anymore)

Part of me wants to just "do it correctly", but we already are hitting the other VU we are desiring and honestly dealing with External Memory (and Android HardwareBuffer) I am more willing to take slight shortcuts here, just for sanity

PR #9362, opened 2025-01-31
tests: Check if AS are supported for mutable

spencer-lunarg:
RADV doesn't support `VK_DESCRIPTOR_TYPE_ACCELERATION_STRUCTURE_KHR` for `VK_DESCRIPTOR_TYPE_MUTABLE_EXT` and needs to be checked

### DIFF
--- a/tests/unit/descriptors_positive.cpp
+++ b/tests/unit/descriptors_positive.cpp
@@ -572,6 +572,12 @@ TEST_F(PositiveDescriptors, CopyAccelerationStructureMutableDescriptors) {
     create_info.bindingCount = bindings.size();
     create_info.pBindings = bindings.data();
 
+    VkDescriptorSetLayoutSupport dsl_support = vku::InitStructHelper();
+    vk::GetDescriptorSetLayoutSupport(device(), &create_info, &dsl_support);
+    if (!dsl_support.supported) {
+        GTEST_SKIP() << "Acceleration Structure not supported for mutable";
+    }
+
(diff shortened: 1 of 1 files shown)


tests/unit/descriptors_positive.cpp
```diff
@@ -572,6 +572,12 @@ TEST_F(PositiveDescriptors, CopyAccelerationStructureMutableDescriptors) {
     create_info.bindingCount = bindings.size();
     create_info.pBindings = bindings.data();
 
+    VkDescriptorSetLayoutSupport dsl_support = vku::InitStructHelper();
+    vk::GetDescriptorSetLayoutSupport(device(), &create_info, &dsl_support);
+    if (!dsl_support.supported) {
```
artem-lunarg, line 577: missed opportunity to blame the driver in the comment?

artem-lunarg (approved):

tests/unit/descriptors_positive.cpp
spencer-lunarg: no, `VK_DESCRIPTOR_TYPE_ACCELERATION_STRUCTURE_KHR` is an optional type for Mutable Descriptors, we (as a bad test) did not check

https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7609 would have caught it, but need to get that PR up again

artem-lunarg: Ah, good to know

PR #9363, opened 2025-01-31
layers: Rewrite inplace_function default alignment

victorvianna:
No behavior change. Replaces a variadic template with a constexpr function and a for loop, which is simpler. The logic was added in 0d99e5aad4299d8e4b3bcf4bf1e1bea4a32a9044.


arno-lunarg (approved):
Nice, much more readable!

lucaslugao:
Great work! Hiding loops behind variadic templates is much harder to read :+1:

PR #9366, opened 2025-01-31
build: Try disabling CMAKE_INTERPROCEDURAL_OPTIMIZATION_RELEASE

jeremyg-lunarg:


artem-lunarg:
For documention purposes. A benchmark session with cs2/dota2/doom gfxr captures did not detect performance difference  with LTCG on and off in syncval and core checks. The flag affects only release build.

Issue #9368, opened 2025-01-31
PositiveRayTracing.StridedDeviceAddressRegion is invalid and nothing catches it


spencer-lunarg:
Full details in https://gitlab.freedesktop.org/mesa/mesa/-/issues/12554

`PositiveRayTracing.StridedDeviceAddressRegion` is invalid and need to understand

1. Are there missing VUs that need to be added to the spec?
2. Are there VUs we don't have that would have caught this (and then add a negative test, can start as disabled, until we add code to detect it)

PR #9378, opened 2025-02-03
layers: Validate invalid ray gen SBT address

arno-lunarg:
Part of https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9368

### DIFF
--- a/layers/core_checks/cc_drawdispatch.cpp
+++ b/layers/core_checks/cc_drawdispatch.cpp
@@ -901,6 +901,13 @@ bool CoreChecks::ValidateCmdTraceRaysKHR(const Location &loc, const vvl::Command
                                                             : "VUID-vkCmdTraceRaysKHR-pRayGenShaderBindingTable-03680";
         const char *vuid_binding_table_flag = is_indirect ? "VUID-vkCmdTraceRaysIndirectKHR-pRayGenShaderBindingTable-03681"
                                                           : "VUID-vkCmdTraceRaysKHR-pRayGenShaderBindingTable-03681";
+        // https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9368
+        // TODO - waiting for https://gitlab.khronos.org/vulkan/vulkan/-/issues/4173
+        if (const auto buffers = GetBuffersByAddress(pRaygenShaderBindingTable->deviceAddress); buffers.empty()) {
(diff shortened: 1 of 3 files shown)


layers/core_checks/cc_drawdispatch.cpp
```diff
@@ -901,6 +901,13 @@ bool CoreChecks::ValidateCmdTraceRaysKHR(const Location &loc, const vvl::Command
                                                             : "VUID-vkCmdTraceRaysKHR-pRayGenShaderBindingTable-03680";
         const char *vuid_binding_table_flag = is_indirect ? "VUID-vkCmdTraceRaysIndirectKHR-pRayGenShaderBindingTable-03681"
                                                           : "VUID-vkCmdTraceRaysKHR-pRayGenShaderBindingTable-03681";
+        // https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9368
+        // TODO - waiting for https://gitlab.khronos.org/vulkan/vulkan/-/issues/4173
+        if (const auto buffers = GetBuffersByAddress(pRaygenShaderBindingTable->deviceAddress); buffers.empty()) {
+            skip |= LogError("UNASSIGNED-TraceRays-InvalidRayGenSBTAddress", cb_state.Handle(), table_loc.dot(Field::deviceAddress),
+                             " (0x%" PRIx64 ") does not belong to a currently alive VkBuffer.",
```
artem-lunarg, line 908: nit/question: Is the word "alive" something we use in such contexts?

artem-lunarg (approved):

layers/core_checks/cc_drawdispatch.cpp
arno-lunarg: I am also not a fan, but could not come up with an alternative : /
I wanted something reminiscent of the "object lifetime" part of validation

spencer-lunarg: From everything we doing around object lifetime the mental model (I thought) is

1. a `VkBuffer` is undefined at start
2. It is created
3. It can be destroyed
4. for various reasons it can be invalidated (like destroying an Image under an ImageView makes the ImageView invalid)

We don't really need to worry about `1`, so these cases are either `3` or `4`

I would just say a `valid VkBuffer` to be honest (mainly because it is consistent with everything else)

spencer-lunarg:
```suggestion
                             "(0x%" PRIx64 ") does not belong to a currently alive VkBuffer.",
```

Issue #9364, opened 2025-01-31
Not detecting PhysicalStorageBuffer as a 64-bit element


spencer-lunarg:
(Currently created internal spec issue to clarify - https://gitlab.khronos.org/vulkan/vulkan/-/issues/4172)

In slang you can have something like (https://godbolt.org/z/8r8nYv7dW)

```c++
RWStructuredBuffer<int> result;

[shader("compute")]
[numthreads(64,1,1)]
void main(
    uint8_t* output,
) {
    result[0] = (int)output[0];
}
```

this will trip up `VUID-RuntimeSpirv-storagePushConstant8-06330` because it will read

```
%14 = OpTypeInt 8 0
%4 = OpTypePointer PhysicalStorageBuffer %14
%6 = OpTypeStruct %4 %8
```

as an 8-bit element in the push constant, but really it is a 64-bit pointer

PR #9380, opened 2025-02-03
layers: Fix checking pointers 8/16 bit elements

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9364

### DIFF
--- a/layers/core_checks/cc_spirv.cpp
+++ b/layers/core_checks/cc_spirv.cpp
@@ -29,6 +29,7 @@
 #include "core_checks/cc_vuid_maps.h"
 #include "core_validation.h"
 #include "generated/spirv_grammar_helper.h"
+#include "state_tracker/shader_instruction.h"
 #include "state_tracker/shader_stage_state.h"
 #include "utils/hash_util.h"
 #include "chassis/chassis_modification_state.h"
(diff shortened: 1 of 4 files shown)


layers/core_checks/cc_spirv.cpp
```diff
@@ -1790,18 +1791,96 @@ struct VariableInstInfo {
...
+            }
+            const uint32_t base_insn_id = module_state.GetBaseType(member_insn);
+            const spirv::Instruction *base_insn = module_state.FindDef(base_insn_id);
             GetVariableInfo(module_state, base_insn, info);
         }
     }
 }
 
+// This is to validate the VK_KHR_8bit_storage and VK_KHR_16bit_storage extensions
+bool CoreChecks::Validate8And16BitStorage(const spirv::Module &module_state, const spirv::Instruction &var_insn,
```
spencer-lunarg, line 1814: created new function as have follow-up PR to provide better error messages using Shader DebugInfo and wanted to isolate this for that

layers/core_checks/cc_spirv.cpp
```diff
@@ -1790,18 +1791,96 @@ struct VariableInstInfo {
...
+    } else if (insn->Opcode() == spv::OpTypePointer) {
+        return;
     } else if (insn->Opcode() == spv::OpTypeFloat || insn->Opcode() == spv::OpTypeInt) {
         const uint32_t bit_width = insn->Word(2);
         info.has_8bit |= (bit_width == 8);
         info.has_16bit |= (bit_width == 16);
     } else if (insn->Opcode() == spv::OpTypeStruct) {
         for (uint32_t i = 2; i < insn->Length(); i++) {
-            const spirv::Instruction *base_insn = module_state.GetBaseTypeInstruction(insn->Word(i));
+            const spirv::Instruction *member_insn = module_state.FindDef(insn->Word(i));
```
artem-lunarg, line 1802: A question to understand what's going on.
Is it enough to have only the above `} else if (insn->Opcode() == spv::OpTypePointer) {` to handle example from the 
issue? If yes, do we need this OpTypeStruct check to handle additional cases?

artem-lunarg (approved):

layers/core_checks/cc_spirv.cpp
spencer-lunarg: valid question

for Push Constants (which is the issue) we need this one here in the struct (because Push Constants are always in structs)

For some of the other things in 8/16 bit storage I have an open spec issue (https://gitlab.khronos.org/vulkan/vulkan/-/issues/4172) as need to better understand what is going on, but if the Variable is just a normal pointer (and not in a struct) it also shouldn't report anything

... basically I am be extra cautious here to not have more false positives around pointers until I get more tests/clarification what is suppose to happen here

PR #9383, opened 2025-02-03
ci: Update runners to use 24.04

spencer-lunarg:
Use 24.04 Ubuntu to make sure we compile with latest version of GCC/Clang on those machines

### DIFF
--- a/.github/workflows/vvl.yml
+++ b/.github/workflows/vvl.yml
@@ -36,7 +36,7 @@ permissions:
 
 jobs:
   code-format:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
         with:
@@ -51,7 +51,7 @@ jobs:
 
   linux:
     needs: check_vvl
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     name: "linux (address sanitizer, ${{matrix.config}}, robinhood ${{matrix.robin_hood}} )"
     strategy:
       fail-fast: false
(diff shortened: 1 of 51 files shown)


.github/workflows/vvl.yml
```diff
@@ -51,7 +51,7 @@ jobs:
 
   linux:
     needs: check_vvl
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
```
jpr42, line 54: If you update the runner. You can update the `sudo sysctl vm.mmap_rnd_bits=28` workaround affecting older linux kernels.

.github/workflows/vvl.yml
```diff
@@ -93,7 +93,7 @@ jobs:
   # Make sure a debug version builds, but no need to run tests on it (since we do with release already)
   linux-debug:
     needs: check_vvl
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
```
jpr42, line 96: It's worth keeping 1 build that runs `22.04`

The `BUILD.md` states that the latest 2 LTS releases are supported.

.github/workflows/vvl.yml
spencer-lunarg: o ya, good call!

.github/workflows/vvl.yml
spencer-lunarg: yes, I was going to add it (we just removed https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/6cf616f131e9870c499a50441bca2d07ccda9733 and this was a goal to swap)

... working through gcc errors right now!

.github/workflows/vvl.yml
```diff
@@ -148,13 +142,39 @@ jobs:
...
+        uses: lukka/get-cmake@latest
+        with:
+          cmakeVersion: 3.22.1
+      - uses: hendrikmuhs/ccache-action@v1.2
+        with:
+          key: linux_back_compat
+      - run: sudo apt-get -qq update && sudo apt-get install -y libwayland-dev xorg-dev
+      # This is to combat a bug when using 6.6 linux kernels with thread/address sanitizer
+      # https://github.com/google/sanitizers/issues/1716
+      - run: sudo sysctl vm.mmap_rnd_bits=28
```
jpr42, line 162: This can be removed. Since you aren't running tests.

artem-lunarg (approved):
LGTM. I assume the most files will dissappear from the diff after rebase.

PR #9385, opened 2025-02-03
chassis: Move enabled_features to vvl::dispatch::Device

jeremyg-lunarg:
Restructure common data used by both stateless and state_tracker  based validation.


jeremyg-lunarg:
There's more behind this for properties, but this chunk is relatively contained and working now. This is necessary because the features and properties are touched all over the place and making them common will make it easier to change the class hierarchy for things using the state tracker.

spencer-lunarg (approved):

spencer-lunarg:
`cc_android.cpp` got you lol

scripts/generators/dispatch_object_generator.py
```diff
@@ -284,7 +292,8 @@ def generateFunctions(self):
             // make vulkan calls.
 
             #include "chassis/dispatch_object.h"
-    
+
+            #pragma once
```
arno-lunarg, line 296: Should probably be moved up?

arno-lunarg (approved):

scripts/generators/dispatch_object_generator.py
jeremyg-lunarg: It can't just be moved up without ending up in the .cpp file but the other .h files should have it too

jeremyg-lunarg: oh and yeah it should be above the #include too

PR #9381, opened 2025-02-03
gpuav: Keep instruction count to make error message faster

spencer-lunarg:
Found by @arno-lunarg in profiling that we really freeze up in `GenerateInstructions` and this hopefully solves it

PR #9390, opened 2025-02-04
gpuav: Do GPU-AV Post Processing with uint32_t only

spencer-lunarg:
@arno-lunarg found https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9381 was not good enough and we need to not use `spirv::Instruction` at all and resort to doing old-fashion `vector<uint32_t>` parsing

Added some helpers and I think it is tolerable (as parsing SPIR-V can be)

### DIFF
--- a/layers/error_message/spirv_logging.cpp
+++ b/layers/error_message/spirv_logging.cpp
@@ -17,6 +17,7 @@
 
 #include "spirv_logging.h"
 #include <string>
+#include <cstring>
 
 // TODO https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9384
 // Fix GCC 13 issues with regex
@@ -32,6 +33,13 @@
 #include "state_tracker/shader_instruction.h"
 #include <spirv/unified1/NonSemanticShaderDebugInfo100.h>
 #include <spirv/unified1/spirv.hpp>
+#include "generated/spirv_grammar_helper.h"
+
+namespace spirv {
+
+static const int kModuleStartingOffset = 5;  // first 5 words of module are the headers
+static inline uint32_t Opcode(uint32_t instruction) { return instruction & 0x0ffffu; }
+static inline uint32_t Length(uint32_t instruction) { return instruction >> 16; }
 
 struct SpirvLoggingInfo {
     uint32_t file_string_id = 0;  // OpString with filename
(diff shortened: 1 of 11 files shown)


layers/gpuav/instrumentation/gpuav_shader_instrumentor.cpp
```diff
@@ -1290,21 +1290,32 @@ static std::string LookupDebugUtilsNameNoLock(const DebugReport *debug_report, c
...
-                if (insn.Opcode() == spv::OpFunction) break;  // early exit when possible
-                if (insn.Opcode() == spv::OpEntryPoint) {
+
+            uint32_t offset = ::spirv::kModuleStartingOffset;
+            while (offset < instrumented_spirv.size()) {
+                const uint32_t instruction = instrumented_spirv[offset];
+                const uint32_t length = ::spirv::Length(instruction);
+                const uint32_t opcode = ::spirv::Opcode(instruction);
+
+                // if here, seen all OpString and can return early
```
spencer-lunarg, line 1305: Will fix (and run CI again) after confirm this is actually the way forward

```suggestion
                // if here, seen all OpEntryPoint and can return early
```

layers/error_message/spirv_logging.cpp
```diff
@@ -42,46 +44,41 @@ struct SpirvLoggingInfo {
...
 // Read the contents of the SPIR-V OpSource instruction and any following continuation instructions.
 // Split the single string into a vector of strings, one for each line, for easier processing.
-static void ReadOpSource(const std::vector<spirv::Instruction> &instructions, const uint32_t reported_file_id,
+static void ReadOpSource(const std::vector<uint32_t> &instructions, const uint32_t reported_file_id,
                          std::vector<std::string> &out_source_lines) {
-    for (size_t i = 0; i < instructions.size(); i++) {
-        const auto &insn = instructions[i];
-        if ((insn.Opcode() != spv::OpSource) || (insn.Length() < 5) || (insn.Word(3) != reported_file_id)) {
+    uint32_t offset = kModuleStartingOffset;
+    while (offset < instructions.size()) {
```
arno-lunarg, line 52: We can do better than this while loop, this old style is error prone: you have to think about updating length, updating offset every time you have a continue or at loop end, you have to update the correct length/offset, etc ...
Use a full fledged `Instruction` iterator class, that would benefit for this very cool `Instruction` class, and come up with something that does this for instance:
```
for (const Instruction& instruction : InstructionsSpan(instructions))
```
Definitely more work, but easier to use, way less error prone

Or if you want to go fast with something not as elegant, at least go for that instead of the `while` loops:
```cpp
    for ( size_t offset = kModuleStartingOffset, length = Length(instruction[offset]); offset < instructions.size(); (offset += length, length = Length(instruction[offset])) ) 
```

layers/error_message/spirv_logging.cpp
```diff
@@ -42,46 +44,41 @@ struct SpirvLoggingInfo {
...
+        in_stream.str(str);
         while (std::getline(in_stream, current_line)) {
             out_source_lines.emplace_back(current_line);
         }
 
-        for (size_t k = i + 1; k < instructions.size(); k++) {
-            const auto &continue_insn = instructions[k];
-            if (continue_insn.Opcode() != spv::OpSourceContinued) {
+        // Look for OpSourceContinued, it must be right after
+        while (offset < instructions.size()) {
```
arno-lunarg, line 70: Looking at the original logic, you should use a copy of `offset` to loop here. So yeah we definitely need to move away from those while loop!

arno-lunarg (changes requested):

arno-lunarg (commented):
.

layers/error_message/spirv_logging.cpp
arno-lunarg, line 70: Styling wise, having this loop be within the top one is confusing because of the `return` right after: when entering this loop, you will never go back to the enclosing one. So you probably should move it out

layers/error_message/spirv_logging.cpp
```diff
@@ -90,46 +87,58 @@ static void ReadOpSource(const std::vector<spirv::Instruction> &instructions, co
...
         while (std::getline(in_stream, current_line)) {
             out_source_lines.emplace_back(current_line);
         }
 
-        for (size_t k = i + 1; k < instructions.size(); k++) {
-            const auto &continue_insn = instructions[k];
-            if (continue_insn.Opcode() != spv::OpExtInst ||
-                continue_insn.Word(4) != NonSemanticShaderDebugInfo100DebugSourceContinued) {
+        // Look for DebugSourceContinued, it must be right after
+        while (offset < instructions.size()) {
```
arno-lunarg, line 125: same for this loop

layers/error_message/spirv_logging.cpp
spencer-lunarg: 1. I moved **all** the SPIR-V uint32_t parsing inside `spirv_logging.cpp`
2. I cleaned up the aweful nested while loops

Then I tried to do this iterator suggestions above (both the C++ and C style), I found adding it made things harder to read and while doing it, I had bug and it was not even obvious what was going on, so I reverted that part of the change

Every `while (offset < instruction.size())` loop is 15 to 20 lines, with mainly a single `offset += length` and feel it's simplicity (and consistency) is good enough here

You might disagree, but I am satisfied with how the change is now and have no plan to update things, not "just" because this PR gets the job done, but because anyone who edits this can easily understand what is going on, it is the only file that does this, no one is going to honestly editing these isolated small functions (they been untouched forever since i added them). If this logic was actually doing more complex things (like the other parts of our SPIR-V code using the `Instruction` class, I would probably reconsider, but they literally just need to loop and pick out a single instruction, spirv-tool, spirv-reflect, spirv-cross, etc do the exact same thing here... its just the nature of parsing a IR

arno-lunarg: Then at least move to the `for` loop I described. It is just safer, and moves the iteration logic where it should be

arno-lunarg: Mmmmh you mean you can't go:
```cpp
// first loop
    for ( size_t offset = kModuleStartingOffset, length = Length(instruction[offset]); offset < instructions.size(); (offset += length, length = Length(instruction[offset])) ) 

// second loop
    for ( ; offset < instructions.size(); (offset += length, length = Length(instruction[offset])) ) 
```
Looking at the code, I bet you 5$ you can

spencer-lunarg: I feel this is harder to grasp, trying to shove all this stuff in the for loop header, but honestly sure, will just do it at this point because I need this PR in to move forward to add more Shader DebugInfo support

arno-lunarg: The while loop with manual increment is just copium, `for` loop where created to handle those kind of "oh but here I need to `continue;` so need to think about incrementing" problems

arno-lunarg: And looking through our code we definitely have existing bad `while` loop usage where a `for` loop would be better

spencer-lunarg: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9390/commits/8aac0c21d718e17a47ef0c5284da47395a6ef296

layers/error_message/spirv_logging.cpp
```diff
@@ -42,99 +51,107 @@ struct SpirvLoggingInfo {
...
-            const auto &continue_insn = instructions[k];
-            if (continue_insn.Opcode() != spv::OpSourceContinued) {
-                return;
-            }
-            in_stream.clear();  // without, will fail getline
-            in_stream.str(continue_insn.GetAsString(1));
-            while (std::getline(in_stream, current_line)) {
-                out_source_lines.emplace_back(current_line);
-            }
+        next_offset = offset + length;  // point to next instruction after OpSource
```
spencer-lunarg, line 75: ugh, now this is a subtle bug, we need to nest these for loops else we need to track the next instruction here

jeremyg-lunarg (approved):
SPIRV parsing is such a PITA

arno-lunarg (approved):

PR #9391, opened 2025-02-05
vvl: Faster get device

arno-lunarg:


layers/chassis/dispatch_object_manual.cpp
```diff
@@ -72,21 +73,25 @@ Device *GetData(VkQueue queue) { return GetDeviceFromKey(GetDispatchKey(queue));
...
+    }
 }
 
 void FreeAllData() {
     {
         WriteLockGuard lock(device_mutex);
-        device_data.clear();
+        for (auto [key, device] : device_data) {
+            delete device;
+        }
```
jeremyg-lunarg, line 94: you should still call `device_data.clear()` after this loop

jeremyg-lunarg (changes requested):

PR #9395, opened 2025-02-05
sync: Cleanup formatting code

artem-lunarg:
Cleanup buffer message formatting code to prepare it for re-use in other places.

### DIFF
--- a/layers/sync/sync_access_state.cpp
+++ b/layers/sync/sync_access_state.cpp
@@ -1060,3 +1060,33 @@ const char *string_SyncHazardVUID(SyncHazard hazard) {
     }
     return "SYNC-HAZARD-INVALID";
 }
+
+SyncHazardInfo GetSyncHazardInfo(SyncHazard hazard) {
+    switch (hazard) {
+        case SyncHazard::NONE:
+            return SyncHazardInfo{};
+        case SyncHazard::READ_AFTER_WRITE:
+            return SyncHazardInfo{false, true};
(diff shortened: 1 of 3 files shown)


layers/sync/sync_access_state.cpp
```diff
@@ -1060,3 +1060,33 @@ const char *string_SyncHazardVUID(SyncHazard hazard) {
     }
     return "SYNC-HAZARD-INVALID";
 }
+
+SyncHazardInfo GetSyncHazardInfo(SyncHazard hazard) {
+    switch (hazard) {
+        case SyncHazard::NONE:
+            return SyncHazardInfo{};
```
spencer-lunarg, line 1067: you think it is better to just set these all here like `return SyncHazardInfo{false, false, false};` and set the fields as const?

```
struct SyncHazardInfo {
    const bool is_write = false;
    const bool is_prior_write = false;
    const bool is_racing_hazard = false;
```

spencer-lunarg (approved):
One suggestion, but LGTM

layers/sync/sync_access_state.cpp
artem-lunarg: Not sure, I feel that simple/POD data structures in general should be flexible and const applied to the structure itself `const SyncHazardInfo& param`. Then if some helper function has to modify fields it should be possible without. I think I put `const` more often on members of more high-level types, so const helps them to simplify by marking ummovable types.

PR #9398, opened 2025-02-06
vvl: Faster get device

arno-lunarg:
- use small unordered map
- cache last used device

### DIFF
--- a/layers/chassis/dispatch_object_manual.cpp
+++ b/layers/chassis/dispatch_object_manual.cpp
@@ -20,6 +20,7 @@
 #include "chassis/dispatch_object.h"
 #include <vulkan/utility/vk_safe_struct.hpp>
 #include "state_tracker/pipeline_state.h"
+#include "containers/custom_containers.h"
 
 #define OBJECT_LAYER_DESCRIPTION "khronos_validation"
 
@@ -39,7 +40,7 @@ static std::shared_mutex instance_mutex;
 static vvl::unordered_map<void *, std::unique_ptr<Instance>> instance_data;
 
 static std::shared_mutex device_mutex;
-static vvl::unordered_map<void *, std::unique_ptr<Device>> device_data;
+static small_unordered_map<void *, Device *, 1> device_data;
 
 static Instance *GetInstanceFromKey(void *key) {
     ReadLockGuard lock(instance_mutex);
@@ -63,7 +64,7 @@ void FreeData(void *key, VkInstance instance) {
 
 static Device *GetDeviceFromKey(void *key) {
     ReadLockGuard lock(device_mutex);
-    return device_data[key].get();
+    return device_data[key];
 }
 
 Device *GetData(VkDevice device) { return GetDeviceFromKey(GetDispatchKey(device)); }
@@ -72,20 +73,26 @@ Device *GetData(VkQueue queue) { return GetDeviceFromKey(GetDispatchKey(queue));
 
 Device *GetData(VkCommandBuffer cb) { return GetDeviceFromKey(GetDispatchKey(cb)); }
 
-void SetData(VkDevice device, std::unique_ptr<Device> &&data) {
+void SetData(VkDevice device, Device *data) {
     void *key = GetDispatchKey(device);
     WriteLockGuard lock(device_mutex);
-    device_data[key] = std::move(data);
+    device_data[key] = data;
 }
 
 void FreeData(void *key, VkDevice device) {
     WriteLockGuard lock(device_mutex);
-    device_data.erase(key);
+    if (device_data.contains(key)) {
+        delete device_data[key];
+        device_data.erase(key);
+    }
 }
 
 void FreeAllData() {
     {
         WriteLockGuard lock(device_mutex);
+        for (auto [key, device] : device_data) {
+            delete device;
+        }
         device_data.clear();
     }
     {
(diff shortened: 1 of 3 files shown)


layers/chassis/dispatch_object_manual.cpp
```diff
@@ -62,8 +66,14 @@ void FreeData(void *key, VkInstance instance) {
 }
 
 static Device *GetDeviceFromKey(void *key) {
+    Device *last_device = last_used_device.load();
+    if (last_device && GetDispatchKey(last_device->device) == key) {
```
artem-lunarg, line 70: Can `GetDeviceFromKey` be used from multiple threads?

Depending on the answer and how this function is used the followed can be non-thread safe:
`last_used_device && GetDispatchKey()` is not an atomic expression and atomic `last_used_device` does not help much. One solution to cache in non-atomic variable and use mutex, but if the point of this solution to go away from mutexes then 🤷.

layers/chassis/dispatch_object_manual.cpp
```diff
@@ -62,8 +66,14 @@ void FreeData(void *key, VkInstance instance) {
 }
 
 static Device *GetDeviceFromKey(void *key) {
+    Device *last_device = last_used_device.load();
```
arno-lunarg, line 69: Don't know what type of `std::memory_order` I should use here?

layers/chassis/dispatch_object_manual.cpp
artem-lunarg: Used mutex for caching surfaces here, maybe can do something smart in specific cases but in general mutex gets less surprises.
[Just ex](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/af84fab9c18963ccfd2e8116dd6144b31fb1e257/layers/state_tracker/image_state.cpp#L866)

arno-lunarg: It can be called from multiple threads, the way I see it caching like this should be ok?
Only considering concurrent `GetDeviceFromKey` calls (concurrency managment with the `FreeData` functions is up to the application to manage, you shouldn't try to use a device from one thread and destroy it in another): 
- If everyone uses the same key, no problem
- If different keys: a thread with a different key will fail this `if (last_device && ...)`, and will update the cache, atomically so concurrent accessed to `last_used_device` are well defined

I don't see a scenario where there is an issue, do you see one?

artem-lunarg: Okay, so if that's used by multiple thread then it looks like a bug.

the problem that `last_device && GetDispatchKey(last_device->device)` expression is not atomic. It possible when thread evaluates only `last_device` before `&&` and then another thread takes place which updates `last_device` value. Then the first thread gets control back and evaluates `GetDispatchKey` using stale `last_device` value.

the atomic variable doesn't help much here with threading behavior, regular variable will probably be to equal degree not thread safe here.

artem-lunarg: Problem even in an atomicity of (a && b) expression. Probaly any other combination (e.g. just `GetDispatchKey(last_device->device)` call) has the same problem. The root cause that update of `last_device` is not guarantee to be properly synchronized - caching can't use this variable if another thread can modify it, that's why lock my be needed.

artem-lunarg: > mutexes are too slow in this case.

Modern mutex can be pretty fast. There is one legend about slow mutexed when about the synchronization primitives that used to sync different threads and had to to kernel switch so could be thousand of cycles.

The mutexes with uses here is to sync the same process. Their speed can be of order magnitude of the atomic operation (like few atomics accesses). I vaguely remember one example when mutex was either 2 or 3 atomic accesses. This mutexes can be constly when there is contention (many theads try to execute the same place), but API call are quite big in terms of instructions, so hopefully this is not a center of contention.

arno-lunarg: > Then the first thread gets control back and evaluates GetDispatchKey using stale last_device value.

In this precise context I don't think there is a notion of stale value: I only care about protecting updates to the cached device. After Thread 1 loads the cache value, the cache being updated is unimportant. It just means that the last device used has now been updated by Thread 2, but all Thread 1 cares about is retrieving the device object corresponding to the passed in `key`. The only issue I see would be that the `if (last_device && GetDispatchKey(last_device->device) == key)` being totally not thread safe, the device could be destroyed when Thread 1 starts doing those operations. But this concurrency problem is up to the app to manage

And as for mutexes being slow, I actually measured it, it is super slow: I measured upwards of 50ms in a Doom frame spent acquiring and releasing lock.

artem-lunarg: Mutexes being slow is also depend on the locked region. If **large** region is locked that it's not about mutex performance but how locking is organized. it's nice to lock for short period of times. But yeah, in some situations it's defintely slow.

layers/chassis/dispatch_object_manual.cpp
arno-lunarg: Ok thank you for those details!! I did not think of ARM...

layers/chassis/dispatch_object_manual.cpp
artem-lunarg: > After Thread 1 loads the cache value, the cache being potentially updated by the time we get to the if is unimportant. It just means that the last device used has now been updated by Thread 2, but all Thread 1 cares about is retrieving the device object corresponding to the passed in key, and it will work no matter "last used device" cache state

Okay, makes sense. I a bit scared of sufficiently advanced lock free code, easy to make mistake. It would be nice to have @jeremyg-lunarg eyes on this.

PR #9400, opened 2025-02-06
vvl: Add last used device caching

arno-lunarg:
3rd try, 2nd was https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9398
Got rid of the `small_unordered_map` change, not using smart pointers leads to memory leaks, and using them is not possible with the current implementation of `small_unordered_map`
So just add caching, it should be enough anyway, will make sure

### DIFF
--- a/layers/chassis/dispatch_object_manual.cpp
+++ b/layers/chassis/dispatch_object_manual.cpp
@@ -21,6 +21,8 @@
 #include <vulkan/utility/vk_safe_struct.hpp>
 #include "state_tracker/pipeline_state.h"
 
+#include <atomic>
+
 #define OBJECT_LAYER_DESCRIPTION "khronos_validation"
 
 #define DISPATCH_MAX_STACK_ALLOCATIONS 32
@@ -40,6 +42,7 @@ static vvl::unordered_map<void *, std::unique_ptr<Instance>> instance_data;
 
 static std::shared_mutex device_mutex;
 static vvl::unordered_map<void *, std::unique_ptr<Device>> device_data;
+static std::atomic<Device *> last_used_device = nullptr;
 
 static Instance *GetInstanceFromKey(void *key) {
     ReadLockGuard lock(instance_mutex);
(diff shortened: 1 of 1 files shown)


layers/chassis/dispatch_object_manual.cpp
```diff
@@ -62,7 +65,13 @@ void FreeData(void *key, VkInstance instance) {
 }
 
 static Device *GetDeviceFromKey(void *key) {
+    Device *last_device = last_used_device.load();
+    if (last_device && GetDispatchKey(last_device->device) == key) {
```
arno-lunarg, line 69: Link to relevant discussion in 2nd PR: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9398#discussion_r1944425569

jeremyg-lunarg: could there still be a race where 1 thread has loaded `last_device` but another thread is freeing it?

jeremyg-lunarg (commented):
this needs to be tested with angle end2end tests before merging

layers/chassis/dispatch_object_manual.cpp
arno-lunarg: To me this case can only occur when you have that:
```cpp
// thread 1
vkUseDevice(device);
// thread 2
vkDestroyDevice(device);
```
And it's up to the application to make sure this does not happen?

arno-lunarg:
Confirmed that just having this caching mechanism solves the performance. (Well still ~4.5 ms spent on this in some frames, still too much but at this point there are other areas to improve before getting to that)

arno-lunarg:
I ran the end2end test suite for ~20 minutes without issues, takes too long to finish. Made sure using loader debug info that I was loading the correct VVL build. Unless you had specific tests in mind, things look ok!

jeremyg-lunarg (approved):

PR #9403, opened 2025-02-06
chassis: Move PostCreateDevice() out of the state tracker

jeremyg-lunarg:
Most of the time, Instance::PostCallRecordCreateDevice() is used to set up stuff in the Device object, which is awkward. Move it up, in part so that when the state tracker is running on its own, it can use this hook.


layers/chassis/chassis_manual.cpp
```diff
@@ -334,8 +334,11 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateDevice(VkPhysicalDevice gpu, const VkDevice
     vvl::dispatch::SetData(*pDevice, std::move(device_dispatch));
     for (auto& vo : instance_dispatch->object_dispatch) {
         // Send down modified create info as we want to mark enabled features that we sent down on behalf of the app
-        vo->PostCallRecordCreateDevice(gpu, reinterpret_cast<VkDeviceCreateInfo*>(&modified_create_info), pAllocator, pDevice,
-                                       record_obj);
+        vo->PostCallRecordCreateDevice(gpu, modified_create_info.ptr(), pAllocator, pDevice, record_obj);
+    }
+    // Note: device_dispatch is no longer valid since it was a std::move source above.
+    for (auto& vo : vvl::dispatch::GetData(*pDevice)->object_dispatch) {
+        vo->PostCreateDevice(modified_create_info.ptr(), record_obj.location);
```
artem-lunarg, line 341: `PostCallRecordCreateDevice` and `PostCreateDevice` are now on the same level and are very similar. More distinctive names might be a good idea, not sure what it could be though.

artem-lunarg (approved):
LGTM, one naming highlight

layers/object_tracker/object_lifetime_validation.h
```diff
@@ -205,8 +205,6 @@ class Device : public vvl::base::Device {
 
     object_list_map_type linked_graphics_pipeline_map;
 
-    bool null_descriptor_enabled{false};
```
spencer-lunarg, line 208: :100: for removing this hack I added from combining everything

spencer-lunarg (approved):

arno-lunarg (approved):

layers/chassis/chassis_manual.cpp
spencer-lunarg: `PostCallRecord` is a chassis prefix name

Maybe `SetDeviceState` or maybe `PostDeviceCreation` (just so it is not `CreateDevice`)

artem-lunarg: some names with finish/finalize: FinishDeviceSetup, FinalizeDeviceSetup.

layers/chassis/chassis_manual.cpp
```diff
@@ -334,8 +334,11 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateDevice(VkPhysicalDevice gpu, const VkDevice
     vvl::dispatch::SetData(*pDevice, std::move(device_dispatch));
     for (auto& vo : instance_dispatch->object_dispatch) {
         // Send down modified create info as we want to mark enabled features that we sent down on behalf of the app
-        vo->PostCallRecordCreateDevice(gpu, reinterpret_cast<VkDeviceCreateInfo*>(&modified_create_info), pAllocator, pDevice,
-                                       record_obj);
+        vo->PostCallRecordCreateDevice(gpu, modified_create_info.ptr(), pAllocator, pDevice, record_obj);
+    }
+    // Note: device_dispatch is no longer valid since it was a std::move source above.
+    for (auto& vo : vvl::dispatch::GetData(*pDevice)->object_dispatch) {
+        vo->FinishDeviceSetup(modified_create_info.ptr(), record_obj.location);
```
spencer-lunarg, line 341: :100: points to @artem-lunarg for suggesting `Finish`

spencer-lunarg (approved):

PR #9404, opened 2025-02-07
spirv: Change file name to match gcc/clang

spencer-lunarg:
main thing is just like I would get a gcc/clang error `layers/core_checks/cc_spirv.cpp:400:30` I would expect the same from the Shader Debug Info as well

also add more tests

### DIFF
--- a/tests/unit/atomics.cpp
+++ b/tests/unit/atomics.cpp
@@ -179,18 +179,6 @@ TEST_F(NegativeAtomic, Int64) {
     AddRequiredFeature(vkt::Feature::shaderInt64);
     RETURN_IF_SKIP(Init());
 
-    // For sanity check without GL_EXT_shader_atomic_int64
-    std::string cs_positive = R"glsl(
-        #version 450
-        #extension GL_EXT_shader_explicit_arithmetic_types_int64 : enable
-        #extension GL_KHR_memory_scope_semantics : enable
-        shared uint64_t x;
-        layout(set = 0, binding = 0) buffer ssbo { uint64_t y; };
-        void main() {
-           y = x + 1;
-        }
-    )glsl";
-
     std::string cs_base = R"glsl(
         #version 450
         #extension GL_EXT_shader_explicit_arithmetic_types_int64 : enable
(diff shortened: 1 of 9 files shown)


tests/unit/atomics.cpp
```diff
@@ -179,18 +179,6 @@ TEST_F(NegativeAtomic, Int64) {
     AddRequiredFeature(vkt::Feature::shaderInt64);
     RETURN_IF_SKIP(Init());
 
-    // For sanity check without GL_EXT_shader_atomic_int64
```
spencer-lunarg, line 182: we already have positive test for this

tests/unit/gpu_av_shader_debug_info.cpp
```diff
@@ -753,7 +752,7 @@ void main() {
                OpFunctionEnd
     )";
 
-    BasicSingleStorageBufferComputeOOB(shader_source, "    x = data.indices[16];\n             ^");
+    BasicSingleStorageBufferComputeOOB(shader_source, "    x = data.indices[16];\n                 ^");
```
spencer-lunarg, line 755: there were bad, were not accounting for the `11: ` at the front of printing the line

tests/unit/shader_debug_info.cpp
```diff
@@ -10,11 +10,237 @@
...
+               OpFunctionEnd
+    )";
+
+    // VUID-RuntimeSpirv-vulkanMemoryModel-06265
+    m_errorMonitor->SetDesiredError("Error occurred at in.comp:4");
+    VkShaderObj const cs(this, source, VK_SHADER_STAGE_COMPUTE_BIT, SPV_ENV_VULKAN_1_2, SPV_SOURCE_ASM);
+    m_errorMonitor->VerifyFound();
+}
+
+TEST_F(NegativeShaderDebugInfo, FileName2) {
```
spencer-lunarg, line 73: to give an idea what the error message of these look like

image

arno-lunarg (approved):
Nice update!!

PR #9407, opened 2025-02-07
Update Tracy GPU profiling

arno-lunarg:
- Tracy GPU profiling API has some flaws described in `profiling.md`, so I switched to my own fork to add what I want: I now profile GPU workload at `vkQueueSubmits` level.
- Added first iteration of scripts to help with the process of getting statistics and comparing traces. They are AI generated, for scripts it should be ok?
- Renamed CPU zones for clarity, it was not obvious what zones like "Dispatch", "PreCallRecord" referred to


layers/profiling/stats.py
```diff
@@ -0,0 +1,234 @@
...
+            sys.exit(1)
+    else:
+        print("No zone substrings provided (--zones or --zones_json). Exiting.")
+        sys.exit(0)
+
+    results = {}
+
+    # Process GPU data in parallel.
+    with ThreadPoolExecutor() as executor:
+        gpu_futures = {
```
arno-lunarg, line 120: Honestly impressed that prompting to parallelize workload just worked!

layers/profiling/profiling.cpp
```diff
@@ -264,14 +336,15 @@ void TracyVkCollector::Create(VkDevice device, VkQueue queue, uint32_t queue_fam
     });
 
     queue_to_collector_map.emplace_back(std::move(collector));
-    VVL_TracyMessageStream("[tracyvk] Added TracyVkCollector (now " << queue_to_collector_map.size()
-                                                                    << " of them): device: " << device << " queue: " << queue
-                                                                    << " queue family i:" << queue_family_i);
+    // VVL_TracyMessageStream("[tracyvk] Added TracyVkCollector (now " << queue_to_collector_map.size()
```
spencer-lunarg, line 339: remove?

layers/profiling/profiling.cpp
```diff
@@ -264,14 +336,15 @@ void TracyVkCollector::Create(VkDevice device, VkQueue queue, uint32_t queue_fam
...
+    //                                                                 << " of them): device: " << device << " queue: " << queue
+    //                                                                 << " queue family i:" << queue_family_i);
 }
 
 void TracyVkCollector::Destroy(TracyVkCollector &collector) {
-    VVL_TracyMessageStream("[tracyvk] Destroying TracyVkCollector: device: " << collector.device << " queue: " << collector.queue);
-
+    // VVL_TracyMessageStream("[tracyvk] Destroying TracyVkCollector: device: " << collector.device << " queue: " <<
```
spencer-lunarg, line 345: same, remove?

layers/profiling/profiling.cpp
```diff
@@ -320,7 +393,7 @@ std::optional<std::pair<VkCommandBuffer, VkFence>> TracyVkCollector::TryGetColle
 
 // Not thread safe for now, should be fine to profile GFXR traces
 TracyVkCollector &TracyVkCollector::GetTracyVkCollector(VkQueue queue) {
-    VVL_TracyMessageStream("[tracyvk] Getting TracyVkCollector for queue " << queue);
+    // VVL_TracyMessageStream("[tracyvk] Getting TracyVkCollector for queue " << queue);
```
spencer-lunarg, line 396: remove

layers/profiling/stats.py
```diff
@@ -0,0 +1,234 @@
...
+import sys
+import argparse
+import pandas as pd
+from statistics import median
+from concurrent.futures import ThreadPoolExecutor, as_completed
+import math
+import csv
+
+# Path to the tracy-csvexport executable
+TRACY_CSVEXPORT_PATH = r"C:\Users\Arno\lunarg\tracy-arno\csvexport\build\Release\tracy-csvexport.exe"
```
spencer-lunarg, line 15: this won't work on my machine

layers/profiling/compare.py
```diff
@@ -0,0 +1,166 @@
...
+        B = RB
+
+    # ANSI 24-bit color escape sequence.
+    color_code = f"\033[38;2;{R};{G};{B}m"
+    reset_code = "\033[0m"
+    diff_str = f"{perc_diff:+.2f}% ({diff:+.2f} ms)"
+    return f"{color_code}{diff_str}{reset_code}"
+
+def main():
+    parser = argparse.ArgumentParser(
```
spencer-lunarg, line 102: like you did in `stats.py` (and we do in other scripts)

either

1. pass the args in like `def main(argv):`
2. Call main after parsing

(this is good practice if one day we want to call a python script from another python script instead of the command line)

layers/profiling/profiling.cpp
arno-lunarg: Given the issues I had, I would like to keep those for potential debugging purposes. Testing on different configurations, with different games, showed different behaviors and I need to easily toggle debugging on again. I could maybe wrap this in some `TRACY_GPU_PROFILING_DEBUG` macro instead of commenting it

layers/profiling/stats.py
arno-lunarg: Ah yikes literally did not look at this code and it shows

layers/profiling/compare.py
arno-lunarg: ... and I actually plan to call those scripts from other ones, so need to fix!

layers/profiling/profiling.cpp
spencer-lunarg: I get it, but just leaving code commented out like this, no context, just looks sloppy, if it is just 1 spot, sure add a comment, but if its 3 or plan to have more, really should have something like `TRACY_GPU_PROFILING_DEBUG`

layers/CMakeLists.txt
```diff
@@ -454,8 +454,8 @@ if (VVL_ENABLE_TRACY)
     include(FetchContent)
     FetchContent_Declare(
         tracy
-        GIT_REPOSITORY https://github.com/wolfpld/tracy.git
-        GIT_TAG v0.11.1
+        GIT_REPOSITORY https://github.com/arno-lunarg/tracy.git
```
jeremyg-lunarg, line 457: should we host this fork  under github.com/LunarG?

Do you plan on eventually making a PR to get this into upstream?

spencer-lunarg (approved):

jeremyg-lunarg (commented):
chassis code LGTM

layers/CMakeLists.txt
arno-lunarg: I see this as temporary (famous last words), eventually I would love to just have this in the main repo, it would also benefit others. I need to discuss it with the maintainers, so I have to admit that it is not much of a priority for now ...

PR #9408, opened 2025-02-07
layers: Validate render pass begin info flags

### LINKED ISSUE #9406: Add new VK_RENDERING_CONTENTS_SECONDARY_COMMAND_BUFFERS_BIT check
`VUID-vkCmdDraw-flags-10582` was added in 1.4.308 spec

ziga-lunarg:
Closes #9406

### DIFF
--- a/layers/core_checks/cc_pipeline_graphics.cpp
+++ b/layers/core_checks/cc_pipeline_graphics.cpp
@@ -3468,6 +3468,12 @@ bool CoreChecks::ValidateDrawPipelineDynamicRenderpass(const LastBound &last_bou
                              ") must be equal to VkRenderingInfo::viewMask ([%" PRIu32 ")",
                              FormatHandle(pipeline).c_str(), pipeline_rendering_ci.viewMask, rendering_view_mask);
         }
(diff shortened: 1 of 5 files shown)


layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -3468,6 +3468,12 @@ bool CoreChecks::ValidateDrawPipelineDynamicRenderpass(const LastBound &last_bou
                              ") must be equal to VkRenderingInfo::viewMask ([%" PRIu32 ")",
                              FormatHandle(pipeline).c_str(), pipeline_rendering_ci.viewMask, rendering_view_mask);
         }
+        if (cb_state.IsPrimary() && (rendering_info.flags & VK_RENDERING_CONTENTS_SECONDARY_COMMAND_BUFFERS_BIT) > 0 &&
```
spencer-lunarg, line 3471: nit question: did you mean `!= 0` instead of `> 0`... same thing, but just never seen greater than used

spencer-lunarg (approved):
LGTM - You beat the headers to the fix lol

layers/core_checks/cc_pipeline_graphics.cpp
ziga-lunarg: Fixed. But there are also other places where it is used like this, for example cc_descriptor.cpp:1198

PR #9409, opened 2025-02-07
sync: Error messages update

artem-lunarg:
* Extracted formatting code that potentially can be used in many error messages.
* Update some of the video error messages to use this shared helper (video tests use the most general `ErrorMessages::Error` message type and it was the simplest place to start).

OLD:
> vkCmdEncodeVideoKHR(): pEncodeInfo->pSetupReferenceSlot->pPictureResource Hazard WRITE_AFTER_READ for reconstructed picture. Access info (usage: SYNC_VIDEO_ENCODE_VIDEO_ENCODE_WRITE, prior_usage: SYNC_VIDEO_ENCODE_VIDEO_ENCODE_READ, read_barriers: VkPipelineStageFlags2(0), command: vkCmdEncodeVideoKHR).

NEW:
> vkCmdEncodeVideoKHR(): WRITE_AFTER_READ hazard detected. vkCmdEncodeVideoKHR writes to reconstructed picture pEncodeInfo->pSetupReferenceSlot->pPictureResource {VkImageView 0xd5b26f0000000010[], codedOffset (x = 0, y = 0), codedExtent (width = 160, height = 64), baseArrayLayer = 0}, which was previously read by another vkCmdEncodeVideoKHR command. No sufficient synchronization is present to ensure that a write (VK_ACCESS_2_VIDEO_ENCODE_WRITE_BIT_KHR) at VK_PIPELINE_STAGE_2_VIDEO_ENCODE_BIT_KHR does not conflict with a prior read (VK_ACCESS_2_VIDEO_ENCODE_READ_BIT_KHR) at the same stage. An execution dependency is sufficient to prevent this hazard.

### DIFF
--- a/layers/sync/sync_validation.cpp
+++ b/layers/sync/sync_validation.cpp
@@ -2430,23 +2430,35 @@ bool SyncValidator::PreCallValidateCmdDecodeVideoKHR(VkCommandBuffer commandBuff
     if (dst_resource) {
         auto hazard = context->DetectHazard(*vs_state, dst_resource, SYNC_VIDEO_DECODE_VIDEO_DECODE_WRITE);
         if (hazard.IsHazard()) {
-            const auto error =
-                error_messages_.Error(hazard, "decode output picture", *cb_access_context, error_obj.location.function);
-            skip |= SyncError(hazard.Hazard(), dst_resource.image_view_state->Handle(),
-                              decode_info_loc.dot(Field::dstPictureResource), error);
+            std::stringstream ss;
+            ss << "decode output picture ";
+            ss << Location(Func::Empty, Field::pDecodeInfo).dot(Field::dstPictureResource).Fields();
+            ss << " ";
+            FormatVideoPictureResouce(*this, pDecodeInfo->dstPictureResource, ss);
+            const std::string resouce_description = ss.str();
+            const auto error = error_messages_.Error(hazard, resouce_description, *cb_access_context, error_obj.location.function);
+            skip |= SyncError(hazard.Hazard(), dst_resource.image_view_state->Handle(), error_obj.location, error);
         }
     }
 
     if (pDecodeInfo->pSetupReferenceSlot != nullptr && pDecodeInfo->pSetupReferenceSlot->pPictureResource != nullptr) {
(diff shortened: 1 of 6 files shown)


layers/sync/sync_validation.cpp
```diff
@@ -2592,12 +2616,15 @@ bool SyncValidator::PreCallValidateCmdEncodeVideoKHR(VkCommandBuffer commandBuff
                 auto hazard = context->DetectHazard(*image_view_state, offset, extent, SYNC_VIDEO_ENCODE_VIDEO_ENCODE_READ,
                                                     SyncOrdering::kOrderingNone);
                 if (hazard.IsHazard()) {
-                    // TODO: there are no tests for this error
```
artem-lunarg, line 2595: actually there is a tests for this error, but need to be run on AMD in order not to be skipped.

layers/error_message/error_location.cpp
```diff
@@ -53,7 +53,9 @@ std::string Location::Message() const {
     if (debug_region && !debug_region->empty()) {
         out << "[ Debug region: " << *debug_region << " ] ";
     }
-    out << StringFunc() << "(): ";
+    if (function != vvl::Func::Empty) {
```
spencer-lunarg, line 56: Sorry, I misspoke earlier, I think this is wrong, there should always be a `Func` I think and you should use `Fields()` instead

`Message()` is really reserved for the `LogError` call itself, but this is likely not well documented

artem-lunarg: Does it break something?

This adds new capability, so in the formatting code is did:
```
 ss << Location(Func::Empty, Field::pDecodeInfo)
                          .dot(Field::pSetupReferenceSlot)
                          .dot(Field::pPictureResource)
                          .Message();
```

artem-lunarg: Ah, I got it!

spencer-lunarg: replace `Message()` with `Fields()` and you should get the exact same output

artem-lunarg: Updated

spencer-lunarg (approved):

Issue #9248, opened 2025-01-16
Have better system to skip validation if an error occurs


spencer-lunarg:
We have spots all over like this

```c++
skip |= ValidateCmd(*cb_state, error_obj.location);
if (skip) return skip;  // basic validation failed, might have null pointers
```

For testing, this works because we have it so `LogError` will return `true`, but that is not always the case for a real usage

We need to create a way to propogate up if there was a validation error recorded or not. This might also be as easy as marking something in the `Logger` or `ErrorObject` class

spencer-lunarg:
Looked into more today, some options

1. (What is I think is the **correct** way)

Create a

```c++
typedef enum VvlResultBits {
    VVL_SUCCESS = 0x00,
    // A LogError has been called and we are in undefined behavior land
    VVL_ERROR = 0x01,
    // We want to skip the dispatch
    VVL_SKIP_DISPATCH = 0x02,
} VvlResultBits;
typedef uint8_t VvlResult;
```

and replace all `PreCallValidate` and `LogError` to return a `VvlResult` instead of `bool` and the `LogError` looks like

```c++
VvlResult result = VVL_ERROR;
const bool skip = debug_report->LogMsg(kErrorBit, objlist, loc, vuid_text, format, argptr);
if (skip) result |= VVL_SKIP_DISPATCH;
return result;
```

The main down side is we literally need to update EVERYTHING

--------

2. Add a need `bool& skip` as parameter to `LogError` and have it set it (then the old `skip` just means "something was logged")

This could be annoying to pipe more things down, so thoughts are to make it maybe part of the `ErrorObject err_object` everyone has. Currently it is const, but could just remove the const (because the fields are const)

PR #9411, opened 2025-02-07
layers: Seperating skip from a logged error

spencer-lunarg:
for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9248

1st commit just removes `const` from ErrorObject
2nd commit is me playing around with idea to use `ErrorObject` to detect if an error has been logged


spencer-lunarg:
closing and putting in the "one day...." pile

PR #9421, opened 2025-02-12
layers: Move range outside of sparse namespace

artem-lunarg:
The issue with `sparse_container::range` is that it associates the range type with sparse operations, but a range is a more general (and lower-level) concept. Seeing `sparse_container` in the code might require an additional check to determine whether it's a high level sparse storage operation or simply the way to represent an interval use two numbers.

* Move range type to `containters/range.h`
* Rename `containers/range_vector.h` to `containters/range_map.h`


layers/containers/range_map.h
```diff
@@ -343,9 +224,10 @@ class range_map {
     }
 
     // do an ranged insert that splits existing ranges at the boundaries, and writes value to any non-initialized sub-ranges
-    range<ImplIterator> infill_and_split(const key_type &bounds, const mapped_type &value, ImplIterator lower, bool split_bounds) {
+    vvl::range<ImplIterator> infill_and_split(const key_type &bounds, const mapped_type &value, ImplIterator lower,
```
jeremyg-lunarg, line 227: couldn't you do `using range = vvl::range` to avoid most of the changes in this class?

jeremyg-lunarg (approved):

layers/containers/range_map.h
artem-lunarg: Thanks, good point. I forgot that you can limit `using` to a scope.

arno-lunarg (approved):

PR #9422, opened 2025-02-12
tests: Ray tracing framework renaming

arno-lunarg:
Rename AccelerationStructureKHR::Build to Create, because that's what it does. Build is done after, so name was confusing

### DIFF
--- a/tests/framework/ray_tracing_objects.cpp
+++ b/tests/framework/ray_tracing_objects.cpp
@@ -250,8 +250,6 @@ GeometryKHR &GeometryKHR::SetInstanceShaderBindingTableRecordOffset(uint32_t ins
     return *this;
 }
 
-GeometryKHR &GeometryKHR::Build() { return *this; }
-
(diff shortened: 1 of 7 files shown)


tests/framework/ray_tracing_objects.cpp
```diff
@@ -512,22 +510,17 @@ void BuildGeometryInfoKHR::UpdateDstAccelStructSize() {
 }
 
 void BuildGeometryInfoKHR::SetupBuild(bool is_on_device_build, bool use_ppGeometries /*= true*/) {
-    // Build geometries
-    for (auto &geometry : geometries_) {
-        geometry.Build();
```
artem-lunarg, line 517: Is this called in a different place?

artem-lunarg: it looks like it was always no op

arno-lunarg: Yes exactly, saw that while doing this renaming

PR #9429, opened 2025-02-13
tests: Update shader object tests

### LINKED ISSUE #9315: Cleanup Shader Object tests
kinda of like https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9313

@ziga-lunarg now that its been a while, it would be good to go through the tests and

1. Make anything consistent you found you did better after making more tests. (Basically assume people are just going to copy and paste what these tests do for future tests)
2. Some of the newer test have this pipeline barrier, which I know you said was because you caught it with Sync Val.. this is good, but we should have a more streamline way of adding it everywhere.

ziga-lunarg:
Closes #9315

### DIFF
--- a/tests/framework/binding.h
+++ b/tests/framework/binding.h
@@ -892,8 +892,11 @@ class ShaderModule : public internal::NonDispHandle<VkShaderModule> {
 class Shader : public internal::NonDispHandle<VkShaderEXT> {
   public:
     Shader() = default;
+    Shader(const Device &dev, VkShaderEXT shader) { NonDispHandle::init(dev.handle(), shader); }
     Shader(const Device &dev, const VkShaderCreateInfoEXT &info) { init(dev, info); }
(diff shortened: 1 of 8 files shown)


tests/framework/binding.h
```diff
@@ -1162,8 +1165,15 @@ class CommandBuffer : public internal::Handle<VkCommandBuffer> {
     void BeginRenderingColor(const VkImageView imageView, VkRect2D render_area);
     void EndRendering();
 
-    void BindVertFragShader(const vkt::Shader &vert_shader, const vkt::Shader &frag_shader);
+    void BindShaders(const vkt::Shader &vert_shader, const vkt::Shader &frag_shader);
```
spencer-lunarg, line 1168: these binds helpers are nice, will make my next Shader Object test nicer

spencer-lunarg (approved):
Thanks @ziga-lunarg - this is a good cleanup and we can now copy-and-pasting from anywhere in the shader object file lol

PR #9431, opened 2025-02-14
layers: Check if pSetupReferenceSlot is NULL for 0752.

zzoon:
The Spec says:

 ```
 If pDecodeInfo->pSetupReferenceSlot is NULL or pDecodeInfo->pSetupReferenceSlot->pPictureResource 
does not refer to the same image subresource as pDecodeInfo->dstPictureResource, then the image subresource
referred to by pDecodeInfo->dstPictureResource must be in the VK_IMAGE_LAYOUT_VIDEO_DECODE_DST_KHR layout
at the time the video decode operation is executed on the device 
(https://docs.vulkan.org/spec/latest/chapters/videocoding.html#VUID-vkCmdDecodeVideoKHR-pDecodeInfo-07252)

```

I found when testing ANV.
This causes a false alarm on ANV, which supports VK_VIDEO_DECODE_CAPABILITY_DPB_AND_OUTPUT_COINCIDE_BIT_KHR.

### DIFF
--- a/layers/core_checks/cc_video.cpp
+++ b/layers/core_checks/cc_video.cpp
@@ -5111,7 +5111,7 @@ bool CoreChecks::PreCallValidateCmdDecodeVideoKHR(VkCommandBuffer commandBuffer,
                              FormatHandle(dst_resource.image_state->Handle()).c_str(), FormatHandle(*vs_state).c_str());
         }
 
-        bool dst_same_as_setup = (setup_resource == dst_resource);
+        bool dst_same_as_setup = (setup_resource == dst_resource) || !setup_resource;
 
         VkImageLayout expected_layout =
             dst_same_as_setup ? VK_IMAGE_LAYOUT_VIDEO_DECODE_DPB_KHR : VK_IMAGE_LAYOUT_VIDEO_DECODE_DST_KHR;


aqnuep:
It is not a false alarm, even if COINCIDE is supported, the setup resource and the decode output must both be specified (pointing to the same resource) and, evidently, both have to be in the DST layout.

The setup resource can only be omitted if the video session does not use any DPB slots at all (intra-only decoding).

What exactly is your setup that you believe is legal? To me it sounds like you simply forgot to set `pSetupReferenceSlot` to point to the info specifying the same resource as the decode output (as you use COINCIDE) and the target slot index.

zzoon:
> It is not a false alarm, even if COINCIDE is supported, the setup resource and the decode output must both be specified (pointing to the same resource) and, evidently, both have to be in the same layout (DPB in this case).
> 
> The setup resource can only be omitted if the video session does not use any DPB slots at all (intra-only decoding).
> 

I found this via dEQP-VK.video.decode.av1.allintra_nosetup_8_separated_dpb, which can be omitted for the setup resource.


> What exactly is your setup that you believe is legal? To me it sounds like you simply forgot to set `pSetupReferenceSlot` to point to the info specifying the same resource as the decode output (as you use COINCIDE) and the target slot index.

aqnuep:
> I found this via dEQP-VK.video.decode.av1.allintra_nosetup_8_separated_dpb, which can be omitted for the setup resource.

For the full history, originally it was expected that `pSetupReferenceSlot` could be `NULL` when the decoded frame is not planned to be set up as a reference frame, but that was changed because HW always require a DPB slot to target (at least as "scratch space"), even if the decoded frame is not set up as reference, so the rules have been clarified.

Per that, even if you don't intend to use the decoded frame as a reference in the future, you still have to specify a non-`NULL` `pSetupReferenceSlot`, because that struct tells the DPB slot index to use for decoding (same goes for encoding).

It is possible that the CTS test case you mention does not handle the interactions with different capabilities and does not set `pSetupReferenceSlot` correctly, but considering that the test case you mention uses a DPB with 8 slots, `pSetupReferenceSlot` is required to be non-`NULL`. There's even a VU for this:

>VUID-vkCmdDecodeVideoKHR-pDecodeInfo-08376
> pDecodeInfo->pSetupReferenceSlot must not be NULL unless the bound video session was created with [VkVideoSessionCreateInfoKHR](https://registry.khronos.org/vulkan/specs/latest/man/html/VkVideoSessionCreateInfoKHR.html)::maxDpbSlots equal to zero

This seems like a CTS test bug, not a validation bug.

zzoon:
> For the full history, originally it was expected that `pSetupReferenceSlot` could be `NULL` when the decoded frame is not planned to be set up as a reference frame, but that was changed because HW always require a DPB slot to target (at least as "scratch space"), even if the decoded frame is not set up as reference, so the rules have been clarified.
> 
> Per that, even if you don't intend to use the decoded frame as a reference in the future, you still have to specify a non-`NULL` `pSetupReferenceSlot`, because that struct tells the DPB slot index to use for decoding (same goes for encoding).
> 

Thank you for the details.

> It is possible that the CTS test case you mention does not handle the interactions with different capabilities and does not set `pSetupReferenceSlot` correctly, but considering that the test case you mention uses a DPB with 8 slots, `pSetupReferenceSlot` is required to be non-`NULL`. There's even a VU for this:
> 
> > VUID-vkCmdDecodeVideoKHR-pDecodeInfo-08376
> > pDecodeInfo->pSetupReferenceSlot must not be NULL unless the bound video session was created with [VkVideoSessionCreateInfoKHR](https://registry.khronos.org/vulkan/specs/latest/man/html/VkVideoSessionCreateInfoKHR.html)::maxDpbSlots equal to zero
> 
> This seems like a CTS test bug, not a validation bug.

Thanks again. Going to close this issue.

aqnuep:
> Sorry. I checked the cts but it sets 0 to both VkVideoSessionCreateInfoKHR::maxDpbSlots and VkVideoSessionCreateInfoKHR::maxActiveReferencePictures, which is valid.

If that is indeed the case then I'll add a proper fix for that and some test cases. I guess we didn't handle this correctly when the spec clarification landed. Let me take care of it.

The test case name `allintra_nosetup_8_separated_dpb` made me think it uses a DPB with 8 slots.

aqnuep:
Summa summarum: the validation layers do exactly what the spec (including VUs) say:

1. If `pSetupReferenceSlot == NULL` then the decode output must be in `DECODE_DST` layout, indifferent of the supported decode output mode
2. If `pSetupReferenceSlot == NULL` then the decode output must be in `DECODE_DST` if it is different from the reconstructed picture (requires `DISTINCT` mode)
3. If `pSetupReferenceSlot == NULL` then the decode output must be in `DECODE_DPB` if it is the same as the reconstructed picture (requires `COINCIDE` mode)

This is probably a CTS bug which doesn't handle image layout correctly in this case.

zzoon:
Thanks for the detailed elaborating. I think it makes sense.   
 I'll look into the cts case further.

PR #9434, opened 2025-02-14
tests: Remove sync val errors in tests

spencer-lunarg:


tests/unit/ycbcr.cpp
```diff
@@ -1442,10 +1442,15 @@ TEST_F(NegativeYcbcr, MultiplaneImageViewAspectMasks) {
     if (IsExtensionsEnabled(VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME)) {
         GTEST_SKIP() << "VK_KHR_portability_subset enabled, can hit issues with imageViewFormatReinterpretation";
     }
+    const VkFormat mp_format = VK_FORMAT_G8_B8_R8_3PLANE_420_UNORM;
```
spencer-lunarg, line 1445: so this one is actually from failing on lavapipe, but regardless, same effect of "making tests more valid"

tests/framework/binding.cpp
```diff
@@ -2066,6 +2066,16 @@ void CommandBuffer::Copy(const Buffer &src, const Buffer &dst) {
 
 void CommandBuffer::ExecuteCommands(const CommandBuffer &secondary) { vk::CmdExecuteCommands(handle(), 1, &secondary.handle()); }
 
+// For positive tests, if you run with sync val, ideally want no errors.
+// Many tests need a simple quick way to sync multiple commands
+void CommandBuffer::FullMemoryBarrier() {
```
artem-lunarg, line 2071: Sometimes for debugging purposes to protect everything the recommendation is to use VK_ACCESS_MEMORY_READ_BIT|VK_ACCESS_MEMORY_WRITE_BIT on ALL_COMMANDS stage (https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/docs/syncval_usage.md#debugging-tips). 

Not sure if that's the intention here. If not, and only transfer accesses are considered, then suggestion to rename FullMemoryBarrier to AllTransferMemoryBarrier or similar (FullMemoryBarrier name is associated with VK_ACCESS_MEMORY_READ_BIT|VK_ACCESS_MEMORY_WRITE_BIT which can be misleading for transfer only).

artem-lunarg (approved):
LGTM, one suggestion is either to update naming or to use barrier that protects everything

tests/framework/binding.cpp
spencer-lunarg: Will just use `VK_ACCESS_MEMORY_READ_BIT`, the intent is we can do a quick full sync

PR #9441, opened 2025-02-15
gn: Remove unsupported files for Fuchsia

gnoliyil:
vk_function_pointers.{cpp,h} is not supported on Fuchsia yet. Since this is only used in tests and we don't yet build Validation Layers tests on Fuchsia, it's okay to skip these files when building the validation layer library.

Test: Vulkan-ValidationLayers v1.4.304.1 with this change builds on Fuchsia tip-of-tree.

Bug: https://fxbug.dev/378964821

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -443,6 +443,16 @@ vvl_sources = [
   "layers/vulkan/generated/vk_validation_error_messages.h",
 ]
 
+if (is_fuchsia) {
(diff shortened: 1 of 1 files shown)


gnoliyil:
cc: @cdotstout

spencer-lunarg (approved):

gnoliyil:
Thank you for your quick review!

PR #9442, opened 2025-02-15
sync: vkQueuePresentKHR: Skip updating queue state if queue is invalid.

gnoliyil:
We caught this issue in a UBsan build of the Vulkan-ValidationLayer on Fuchsia where `PostCallRecordQueuePresentKHR()` tried to update a null QueueSyncState pointer, causing a nullptr dereference error. This change fixes the issue by skipping the update if `queue_state` is invalid.

### DIFF
--- a/layers/sync/sync_validation.cpp
+++ b/layers/sync/sync_validation.cpp
@@ -3192,6 +3192,7 @@ void SyncValidator::PostCallRecordQueuePresentKHR(VkQueue queue, const VkPresent
 
     // Update the state with the data from the validate phase
     std::shared_ptr<QueueSyncState> queue_state = std::const_pointer_cast<QueueSyncState>(std::move(cmd_state->queue));
+    if (!queue_state) return;  // Invalid Queue
     ApplySignalsUpdate(cmd_state->signals_update, queue_state->PendingLastBatch());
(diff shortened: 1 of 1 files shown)


gnoliyil:
cc: @cdotstout

artem-lunarg:
@gnoliyil Does the regular (coer) validation (that includes Stateless paramerers) passes successfully (it can detect null values)? I'm trying to understand how zero queue is possible here.
image.

Also if that's something specific to Fuchsia I'd love to know the details.

artem-lunarg (approved):
LGTM. Should be investiated later.

gnoliyil:
We were running the following test on Fuchsia: https://cs.opensource.google/fuchsia/fuchsia/+/main:src/lib/vulkan/tests/test_hermetic_swapchain.cc;l=577-666;drc=ccfae79f76565ac03ed35f3ce6f392c6d8e2542a

From the test output (https://pastebin.com/cVezuNT9, Line 568-658), we saw some `VUID-vkAcquireNextImageKHR-semaphore-01780` errors (expected) but the `VkQueue` should always be valid (and it didn't trigger a stateless check).

The test is Fuchsia-specific (we use a Fuchsia-only swapchain layer) so I am not sure if this could be reproduced on other platforms...

PR #9443, opened 2025-02-15
state: Support VkSemaphores imported with Zircon handles

gnoliyil:
This adds support for vkImportSemaphoreZirconHandleFUCHSIA() and vkGetSemaphoreZirconHandleFUCHSIA() to state_tracker. The logic is similar to fd and win32 handles.

Bug: https://fxbug.dev/379153784
Test: Fuchsia image-pipe-swapchain-test [1] passed with Vulkan- ValidationLayers vulkan-sdk-1.4.304.1 patched with this change.

[1] https://cs.opensource.google/fuchsia/fuchsia/+/main:src/lib/vulkan/tests/test_swapchain.cc

### DIFF
--- a/layers/state_tracker/state_tracker.cpp
+++ b/layers/state_tracker/state_tracker.cpp
@@ -3385,6 +3385,25 @@ void Device::PostCallRecordGetFenceWin32HandleKHR(VkDevice device, const VkFence
 }
 #endif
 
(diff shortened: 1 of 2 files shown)


gnoliyil:
cc: @cdotstout

spencer-lunarg (approved):

PR #9452, opened 2025-02-17
tests: Split large draw/dispatch test

artem-lunarg:
Split 1 huge test into 11 screen size tests.


spencer-lunarg (approved):
:100: - debugging large tests when they fail is the worst

Issue #9384, opened 2025-02-03
Add UBSAN testing


jpr42:
Similar to the existing address sanitization and thread sanitization it's worth adding undefined behavior sanitization testing:
 https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html

Running VVL tests with UBSAN does show failing tests.

Here is an example:
```
[ RUN      ] NegativeCommand.MultiDraw
/home/juaramos/projects/Vulkan-ValidationLayers/layers/core_checks/cc_drawdispatch.cpp:284:96: runtime error: member access within misaligned address 0x7ffdc5bf777d for type 'const struct VkMultiDrawIndexedInfoEXT', which requires 4 byte alignment
0x7ffdc5bf777d: note: pointer points here
 00 00 00 00 00 00 00  01 00 00 00 00 00 00 00  00 00 00 00 01 00 00 00  00 00 00 00 00 00 00 00  00
             ^ 
Command "/home/juaramos/projects/Vulkan-ValidationLayers/build-ci/install/bin/vk_layer_validation_tests" failed with return code -6
```

NOTE: This doesn't need to be a new CI pass. This can just be added to ASAN testing.

GCC / Clang accept the following: `-fsanitize=address,undefined`.

NOTE: I had to set the environment variable `UBSAN_OPTIONS` to `abort_on_error=1:halt_on_error=1` to get lldb to quickly point me to the exact problem when running the tests in vscode.

EDIT: For further reference glslang implemented it in this PR https://github.com/KhronosGroup/glslang/pull/3654

spencer-lunarg:
Did a quick attempt in a branch (https://github.com/spencer-lunarg/Vulkan-ValidationLayers/actions/runs/13181233782/job/36792380702)

Getting strange linking errors

`/usr/bin/ld: source/opt/libSPIRV-Tools-opt.a(pass_manager.cpp.o):(.data.rel+0x58): undefined reference to `typeinfo for spvtools::utils::Timer'`

I had similar in Glslang (which I see you caught) but seems we might have to fix SPIRV-Tools as well first

jpr42:
> Getting strange linking errors

It's related to RTTI from what I could tell:
https://stackoverflow.com/questions/37358573/why-does-fsanitize-undefined-cause-undefined-reference-to-typeinfo

jpr42:
Also I looked at your branch. You will want to set the environment variable UBSAN_OPTION when running the tests.

[...]

PR #9453, opened 2025-02-17
ci: Run Undefined runner

spencer-lunarg:
Attempt to add a runner for Undefined Behavior for #9384

### DIFF
--- a/.github/workflows/vvl.yml
+++ b/.github/workflows/vvl.yml
@@ -87,27 +87,34 @@ jobs:
         env:
           VK_KHRONOS_PROFILES_PROFILE_FILE: ${{ github.workspace }}/tests/device_profiles/min_core.json
 
-  # Make sure a debug version builds, but no need to run tests on it (since we do with release already)
-  linux-debug:
+  linux-undefined:
     needs: check_vvl
     runs-on: ubuntu-24.04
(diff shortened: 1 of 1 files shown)


.github/workflows/vvl.yml
```diff
@@ -87,27 +87,34 @@ jobs:
         env:
           VK_KHRONOS_PROFILES_PROFILE_FILE: ${{ github.workspace }}/tests/device_profiles/min_core.json
 
-  # Make sure a debug version builds, but no need to run tests on it (since we do with release already)
-  linux-debug:
```
artem-lunarg, line 91: Does this remove debug build?

spencer-lunarg: ya, we have debug builds tested inside internal CI, we just recently demoted the Debug run to a Debug build and honestly no sure what it offers (since it is fully tested internally)

So replacing those CI minutes with an Undefined Behavior runner seems better use of the credits

artem-lunarg: I don't remember details about replacing debug build, but if was about timing, then trading timing to debug build validity checks (c-runtime, stl) is a bad move.

artem-lunarg (approved):

.github/workflows/vvl.yml
spencer-lunarg: discussed, will keep the debug build

PR #9465, opened 2025-02-18
layers: video_maintenance2 depends on video_queue

### LINKED ISSUE #9464: Codegen - Handle case where extension needs A or B extension as dependency
**Environment:**
 - OS: Linux
 - GPU and driver version:
 - SDK or header version if building from repo: 1.4.304

**Describe the Issue**

In the case of a decode operation such as `dEQP-VK.video.decode.h264.inline_session_params_layered_dpb`, the validation layer complains that the maintenance2 extension requires VK_KHR_video_encode_queue which should not be mandatory.

**Expected behavior**

No validation layer issue if `vk_khr_video_queue` is enabled

**Valid Usage ID**
VUID-vkCreateDevice-ppEnabledExtensionNames-01387

**Additional context**

```
VUID-vkCreateDevice-ppEnabledExtensionNames-01387(ERROR / SPEC): msgNum: 307460652 - Validation Error: [ VUID-vkCreateDevice-ppEnabledExtensionNames-01387 ] Object 0: handle = 0x55555b309110, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0x12537a2c | vkCreateDevice(): pCreateInfo->ppEnabledExtensionNames[3] Missing extension required by the device extension VK_KHR_video_maintenance2: VK_KHR_video_encode_queue.
```

dabrain34:
The vk_khr_video_maintenance2 extension should depend only vk_khr_video_queue.

Fixes #9464


layers/vulkan/generated/vk_extension_helper.h
```diff
@@ -995,8 +995,7 @@ struct DeviceExtensions : public InstanceExtensions {
              Info(&DeviceExtensions::vk_khr_maintenance8, {{{&DeviceExtensions::vk_feature_version_1_1, "VK_VERSION_1_1"}}})},
             {vvl::Extension::_VK_KHR_video_maintenance2,
              Info(&DeviceExtensions::vk_khr_video_maintenance2,
-                  {{{&DeviceExtensions::vk_khr_video_decode_queue, VK_KHR_VIDEO_DECODE_QUEUE_EXTENSION_NAME},
-                    {&DeviceExtensions::vk_khr_video_encode_queue, VK_KHR_VIDEO_ENCODE_QUEUE_EXTENSION_NAME}}})},
+                  {{{&DeviceExtensions::vk_khr_video_queue, VK_KHR_VIDEO_QUEUE_EXTENSION_NAME}}})},
```
spencer-lunarg, line 998: this is generated code, which means if this is wrong it needs to be fixed in the `vk.xml` 

@aqnuep can you take a look

dabrain34: This is not the right repository to fix the issue. Closing

spencer-lunarg: yes, make a fix in https://github.com/KhronosGroup/Vulkan-Docs for the `vk.xml` as that is how we generate this

aqnuep: No, the `vk.xml` is correct. It uses `,` (i.e. OR), not `+` (i.e. AND), but the code-gen is incorrect as it apparently doesn't differentiate between the two.

Issue #9412, opened 2025-02-08
CTS test crashes when both sync and gpuav are enabled


ziga-lunarg:
The test `dEQP-VK.api.device_init.create_instance_device_intentional_alloc_fail.basic` will crash if ran with everything enabled

ziga-lunarg:
This is probably not gpu-av related, but general vvl framework.

PR #9466, opened 2025-02-18
vvl: Fix crash

arno-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9412
Handle the fact that GPU-AV can be removed
from the object_dispatch list

### DIFF
--- a/layers/chassis/chassis_manual.cpp
+++ b/layers/chassis/chassis_manual.cpp
@@ -337,8 +337,16 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateDevice(VkPhysicalDevice gpu, const VkDevice
         vo->PostCallRecordCreateDevice(gpu, modified_create_info.ptr(), pAllocator, pDevice, record_obj);
     }
     // Note: device_dispatch is no longer valid since it was a std::move source above.
-    for (auto& vo : vvl::dispatch::GetData(*pDevice)->object_dispatch) {
+    // Note 2: GPU-AV can be disabled if some conditions are not met,
+    // and removed from object_dispatches, thus disallowing proper usage of a range-based
+    // for loop => just use a classic one instead
+    auto& object_dispatches = vvl::dispatch::GetData(*pDevice)->object_dispatch;
+    for (size_t i = 0; i < object_dispatches.size(); ++i) {
+        auto& vo = object_dispatches[i];
         vo->FinishDeviceSetup(modified_create_info.ptr(), record_obj.location);
+        if (!object_dispatches[i]) {
+            --i;
+        }
     }
 
     return result;


layers/chassis/chassis_manual.cpp
```diff
@@ -337,7 +337,12 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateDevice(VkPhysicalDevice gpu, const VkDevice
         vo->PostCallRecordCreateDevice(gpu, modified_create_info.ptr(), pAllocator, pDevice, record_obj);
     }
     // Note: device_dispatch is no longer valid since it was a std::move source above.
-    for (auto& vo : vvl::dispatch::GetData(*pDevice)->object_dispatch) {
+    // Note 2: GPU-AV can be disabled if some conditions are not met,
```
artem-lunarg, line 340: what happens here so range-base loop does not work?
Does disabling gpu-av remove element from `object_dispatch` array?

arno-lunarg: Yes exactly

spencer-lunarg (approved):
seems logically correct to me, but @jeremyg-lunarg to take a look as the `git blame` of this whole file now

layers/chassis/chassis_manual.cpp
arno-lunarg: And it should not be necessary to handle this case everywhere, since GPU-AV can only be disabled at device creation time

artem-lunarg: If this works as it could be an issue here too because after removing element say 3, on the next iteration we should still use 3, or do I miss something?

arno-lunarg: Ah yes that's an issue x_x

arno-lunarg: I think I can cope by doing:
```cpp
    auto& object_dispatches = vvl::dispatch::GetData(*pDevice)->object_dispatch;
    for (size_t i = 0; i < object_dispatches.size(); ++i) {
        auto& vo = object_dispatches[i];
        vo->FinishDeviceSetup(modified_create_info.ptr(), record_obj.location);
       // If object dispatch was removed, next iteration need to again look a object_dispatches[i]
       if (vo != object_dispatches[i]) {
            --i;
       }
    }
```
The comparison should work since we compare `std::unique_ptr`s

artem-lunarg: how does it remove elememnt, using vector::erase? one question if `auto& vo` reference is valid.

arno-lunarg: aaaah yes, cannot rely on `vo` anymore after the call to `FinishDeviceSetup`... But then I just need to go
```cpp
if (!object_dispatches[i]) {
    --i;
}
```

jeremyg-lunarg: oh it happens in the FinishDeviceSetup call. yuck

layers/chassis/chassis_manual.cpp
```diff
@@ -337,7 +337,12 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateDevice(VkPhysicalDevice gpu, const VkDevice
...
+    // and removed from object_dispatches, thus disallowing proper usage of a range-based
+    // for loop => just use a classic one instead
+    auto& object_dispatches = vvl::dispatch::GetData(*pDevice)->object_dispatch;
+    for (size_t i = 0; i < object_dispatches.size(); ++i) {
```
jeremyg-lunarg, line 344: I don't think this will work correctly if gpuav and syncval are enabled at the same time, because syncval is after gpuav `object_dispatch`. When gpu-av is erased, syncval will move forward in the vector and its FinishDeviceSetup call will not happen. I think if you copy the object_dispatch to a local variable (instead of just a reference) it'll work correctly no matter which type of loop is used.

However, the erasing could happen in many other calls and doing this copy trick every time would kill performance.

jeremyg-lunarg:
give me a second and I'll put up an alternative fix.

layers/chassis/chassis_manual.cpp
arno-lunarg: Ah yes GPU-AV can be removed at any time because it can `abort`, that sucks...

jeremyg-lunarg:
I think the right fix is to just null out entries inside of ReleaseValidationObject. Then the loops need to become:

```
for (auto& vo : object_dispatch) {
   if (vo) {
       vo->Something();
   }
}
```
This:
```
if (!object_dispatches[i]) {
    --i;
}
```
will cause syncval's call to be skipped.

jeremyg-lunarg:
> Isn't it UB to use a range-based for loop, remove an element from the vector, and proceed with the loop?

yes. that's why I was trying to say that ReleaseValidationObject() needs to just set the `object_dispatch` element that is aborting to `nullptr`, NOT remove an element.

PR #9470, opened 2025-02-18
chassis: Fix ReleaseValidationObject()

### LINKED ISSUE #9412: CTS test crashes when both sync and gpuav are enabled

jeremyg-lunarg:
GPU-AV can detect problems and remove itself from object_dispatch and intercept_vectors. Make these lists rather than vectors so that removal doesn't invalidate iterators. Removal can happen during the FinishDeviceSetup() loop or possibly any other vulkan call. GPU-AV is not the last entry in these lists and calls still need to happen for anything after it (currently only syncval), so playing games with vector indexing will not work.

Fixes #9412

### DIFF
--- a/layers/chassis/dispatch_object.h
+++ b/layers/chassis/dispatch_object.h
@@ -21,6 +21,7 @@
 
 #pragma once
 #include <atomic>
+#include <list>
 
 #include <vulkan/vulkan.h>
 #include <vulkan/vk_enum_string_helper.h>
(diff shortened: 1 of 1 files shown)


layers/chassis/dispatch_object.h
```diff
@@ -232,7 +233,7 @@ class Instance : public HandleWrapper {
     APIVersion api_version;
     DeviceExtensions extensions{};
 
-    mutable std::vector<std::unique_ptr<base::Instance>> object_dispatch;
+    mutable std::list<std::unique_ptr<base::Instance>> object_dispatch;
```
artem-lunarg, line 236: std::list is a rare guest, maybe a comment why we use it.

artem-lunarg (approved):
LGTM

arno-lunarg (changes requested):
I can still repro the crash : /
image

arno-lunarg:
I think the fix from my PR now updated to manage the issues Artem should work, but the for loop update needs to be propagated to *every* functions -_-

jeremyg-lunarg:
yes this approach won't work either. :(

PR #9471, opened 2025-02-18
gpuav: Add a BDA Root Node

spencer-lunarg:
This is first step towards making more use of BDA to do descriptor management.

I used BDA as the first example, because it doesn't need to be updated each draw as it gets updated before `vkQueueSumbit`


docs/gpu_av_shader_instrumentation.md
```diff
@@ -146,4 +146,12 @@ OpBranchConditional %compare %L2 %L3
 
 %L3 = OpLabel
 OpReturn
-```
\ No newline at end of file
+```
+
+## Creating a BDA Root Node
+
+We have used descriptors in the past to bind our buffers, but are moving to doing it all ourselves with Buffer Device Address. This provides control and lets us control the indirect binding look ups ourselves.
```
arno-lunarg, line 153: Wording for "This provides control and lets us control ..."

layers/gpuav/core/gpuav.h
```diff
@@ -446,6 +446,8 @@ class Validator : public GpuShaderInstrumentor {
     VmaPool output_buffer_pool_ = VK_NULL_HANDLE;
     std::unique_ptr<vko::DescriptorSetManager> desc_set_manager_;
 
+    vko::Buffer root_node_;
```
arno-lunarg, line 449: Comment describing usage would be nice. 
Actually given broad usage scope of all members in this class, they should all have a comment...

layers/gpuav/core/gpuav_setup.cpp
```diff
@@ -528,6 +527,21 @@ void Validator::FinishDeviceSetup(const VkDeviceCreateInfo *pCreateInfo, const L
...
     }
 
+    // Create buffer for Root Note
+    {
+        VkBufferCreateInfo buffer_info = vku::InitStructHelper();
+        buffer_info.usage = VK_BUFFER_USAGE_STORAGE_BUFFER_BIT | VK_BUFFER_USAGE_SHADER_DEVICE_ADDRESS_BIT;
+        buffer_info.size = sizeof(glsl::RootNode);
+
+        VmaAllocationCreateInfo alloc_info{};
+        alloc_info.requiredFlags = VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT;
```
arno-lunarg, line 537:
```suggestion
        alloc_info.requiredFlags = VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT;
        alloc_info.preferredFlags = the_device_local_flag;
```

layers/gpuav/instrumentation/gpuav_shader_instrumentor.h
```diff
@@ -219,6 +221,10 @@ class GpuShaderInstrumentor : public vvl::Device {
     // sense too)
     mutable bool aborted_ = false;
 
+    // We need a copy here because we can't have the vko::Buffer object in this class, but also need to pass this in during shader
+    // instrumentation
+    uint64_t root_node_address_;
```
arno-lunarg, line 226: Why not
```suggestion
    VkDeviceAddress root_node_address_;
```

layers/gpuav/shaders/instrumentation/buffer_device_address.comp
```diff
@@ -40,8 +38,8 @@ struct Range {
 // Ranges are supposed to:
 // 1) be stored for low to high
 // 2) not overlap
-layout(set = kInstDefaultDescriptorSet, binding = kBindingInstBufferDeviceAddress, std430) buffer BuffAddrInputBuffer {
-    uint64_t bda_ranges_count;
+layout(buffer_reference, std430) buffer BDAInputBuffer {
+    uint64_t bda_range_count;
```
arno-lunarg, line 42:
```suggestion
    uint64_t bda_ranges_count;
```

layers/gpuav/spirv/module.cpp
```diff
@@ -500,6 +503,10 @@ void Module::LinkFunction(const LinkInfo& info) {
                 // Replace LinkConstants
                 if (constant_value == glsl::kLinkShaderId) {
                     new_inst->UpdateWord(3, shader_id_);
+                } else if (constant_value == glsl::kLinkRootNodeAddressLow) {
+                    new_inst->UpdateWord(3, static_cast<uint32_t>(root_node_address_ & 0xFFFFFFFF));
```
arno-lunarg, line 507:
```suggestion
                    new_inst->UpdateWord(3, static_cast<uint32_t>(root_node_address_));
```
No need to mask

layers/gpuav/spirv/module.cpp
```diff
@@ -489,7 +490,9 @@ void Module::LinkFunction(const LinkInfo& info) {
             // for simplicity, just create a new constant for things other than 32-bit OpConstant as there are rarely-to-none
             // composite/null/true/false constants in linked functions. The extra logic to try and find them is much larger and cost
             // time failing most the searches.
-            if (opcode == spv::OpConstant) {
+            //
+            // If length is 5, it is a 64-bit constant, which we don't care about
```
arno-lunarg, line 494:
```suggestion
            // If length is 4, it is a 64-bit constant, which we don't care about
```
or should the check be `new_inst->Length() == 4`?

layers/gpuav/shaders/instrumentation/buffer_device_address.comp
artem-lunarg: FWIW, singluar form is used with count and plural with words like number:
`bda_range_count`, `num_bda_ranges`

arno-lunarg: Ok, I don't get this convention since grammar rules tell that countable nouns should use plural form in sentences like "the count of bananas"

arno-lunarg: Here the correct English would be "Buffer device address ranges' count"

artem-lunarg: Yeah, I'm also interested why is that.  It was a big open question for me due to non-native specifics, and I was looking different code bases and came to this conclusion. Do people also use plural with count in the code?

artem-lunarg: Something from doom codebase:
https://github.com/TTimo/doom3.gpl/blob/aaa855815ab484d5bd095f347163194ac569abcc/neo/renderer/Image.h#L112C5-L112C33
`mip_count`

https://github.com/TTimo/doom3.gpl/blob/aaa855815ab484d5bd095f347163194ac569abcc/neo/renderer/Image.h#L194C28-L194C36
`refCount`

https://github.com/TTimo/doom3.gpl/blob/aaa855815ab484d5bd095f347163194ac569abcc/neo/renderer/RenderWorld_local.h#L61
`viewCount`

For mip/ref I think it sounds good in singular form. Similar in other codebases, doom just the first one.

spencer-lunarg: I was using the Vulkan Spec naming convention 

`VkPipelineViewportStateCreateInfo::viewportCount` and `pViewports`

layers/gpuav/spirv/module.h
```diff
@@ -111,6 +112,8 @@ class Module {
     // instrumentation
     bool has_bindless_descriptors_ = false;
 
+    uint64_t root_node_address_;
```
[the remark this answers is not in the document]
spencer-lunarg: It eventually gets casted somewhere from a `VkDeviceAddress` to a `uint64_t` because we load it up as a `uvec2` so I can just cast at the end, was trying to cast it just once when stored, but likely just better to keep in `VkDeviceAddress`

layers/gpuav/spirv/module.cpp
spencer-lunarg: - length() == 4 == 32-bit
- length() == 5 == 64-bit

I only want to check for 32 bits, will just adjust it to be `new_inst->Length() < 5`

layers/gpuav/shaders/instrumentation/buffer_device_address.comp
```diff
@@ -54,15 +54,19 @@ bool inst_buffer_device_address(
     if ((addr & (alignment - 1)) != 0) {
         error_sub_code = kErrorSubCodeBufferDeviceAddressAlignment;
     } else {
-        const Range cache_range = bda_ranges[index_cache];
+        RootNode root_node = RootNode(root_node_address);
+        BDAInputBuffer in_buffer = root_node.binding_buffer_device_address;
+
+        const Range cache_range = in_buffer.bda_ranges[index_cache];
```
spencer-lunarg, line 60: so this seems to be a compiler bug (going to report to RADV later)

using debug printf, I can see the `index_cache` is zero, I can access it, copy it, etc, but once I try to deference the array (inside a BDA) with the workgroup storage class... :boom: 

workaround we have an extra byte to store it in the BDA

layers/gpuav/resources/gpuav_state_trackers.cpp
```diff
@@ -308,11 +309,13 @@ bool CommandBuffer::UpdateBdaRangesBuffer(const Location &loc) {
...
     // ---
-    auto bda_table_ptr = (VkDeviceAddress *)bda_ranges_snapshot_.GetMappedPtr();
+    auto bda_table_ptr = (uint32_t *)bda_ranges_snapshot_.GetMappedPtr();
 
     // Buffer device address table layout
     // Ranges are sorted from low to high, and do not overlap
-    // QWord 0 | Number of *ranges* (1 range occupies 2 QWords)
+    // QWord 0 | split up into two dwords
+    //     DWord 0 | Number of *ranges* (1 range occupies 2 QWords)
+    //     DWord 1 | unused
```
spencer-lunarg, line 318: This is really to just not waste casting the range from uint64_t to a uint32_t  later in the shader

layers/gpuav/shaders/instrumentation/buffer_device_address.comp
```diff
@@ -28,16 +28,25 @@ struct Range {
...
     Range bda_ranges[];
 };
 
 // It is common that an app only has a single BDA address and it is used to poke inside a struct.
-// This likely means there is only a single range being accessed, and for shader that do multiple checks,
-// we can hopefully speed up runtime perf by hitting this early and leaving fast
+// It has been observed from dumps of shaders, apps tend to poke at the same BDA multiple times together.
+// This means having a simple "last used" cache can help speed up apps
+//
+// TODO - Play around more with having the cache be the last 2 or 4 elements as well as having no cache
```
spencer-lunarg, line 43: This actually can be really simple (and powerful), like we know the first usage can be a no-cache version. Also we can detect in a single block if the address is coming from the same area, but future work

layers/gpuav/shaders/instrumentation/buffer_device_address.comp
```diff
@@ -28,16 +28,25 @@ struct Range {
...
+//        (and picking depending on what we see instrumenting)
+//
+// Note - This NEEDS to be initialized with zero otherwise found to crash drivers
+//        (it will print as zero, but if used to index into an array, will just crash).
+//        GLSL lacks the ability to use the Initializer ID to a OpVariable, so while linking,
+//        we will adjust the SPIR-V to set this to zero to start
 uint index_cache;
```
spencer-lunarg, line 50: So @arno-lunarg (@jeremyg-lunarg) fun learning here, this was crashing on RADV when indexing, the issue is GLSL

https://godbolt.org/z/7of7aEnqT

we now add logic in the linking to make it look like

```
%uint_0 = OpConstant %uint 0
%_ptr_Private_uint = OpTypePointer Private %uint
%index_cache = OpVariable %_ptr_Private_uint Private %uint_0
```

so that the first time a thread/invocation gets to it, it will be set

layers/vulkan/generated/instrumentation_descriptor_indexing_oob_bindless_combined_image_sampler_comp.cpp
```diff
@@ -64,14 +64,14 @@
     0x00040047, 0x00000021, 0x00000006, 0x00000004, 0x00030047, 0x00000022, 0x00000002, 0x00050048, 0x00000022, 0x00000000,
     0x00000023, 0x00000000, 0x00040047, 0x00000024, 0x00000021, 0x00000003, 0x00040047, 0x00000024, 0x00000022, 0x00000007,
     0x00040047, 0x00000089, 0x00000006, 0x00000004, 0x00030047, 0x0000008a, 0x00000002, 0x00050048, 0x0000008a, 0x00000000,
-    0x00000023, 0x00000000, 0x00040047, 0x0000008c, 0x00000021, 0x00000006, 0x00040047, 0x0000008c, 0x00000022, 0x00000007,
+    0x00000023, 0x00000000, 0x00040047, 0x0000008c, 0x00000021, 0x00000005, 0x00040047, 0x0000008c, 0x00000022, 0x00000007,
```
spencer-lunarg, line 67: these are just because we adjustedm things in like `
gpuav_shaders_constants.h`

layers/gpuav/resources/gpuav_state_trackers.cpp
```diff
@@ -342,6 +345,9 @@ bool CommandBuffer::UpdateBdaRangesBuffer(const Location &loc) {
     bda_ranges_snapshot_.FlushAllocation(loc);
     bda_ranges_snapshot_version_ = gpuav->buffer_device_address_ranges_version;
 
+    auto root_node_ptr = (glsl::RootNode *)gpuav->root_node_.GetMappedPtr();
+    root_node_ptr->binding_buffer_device_address = bda_ranges_snapshot_.Address();
```
arno-lunarg, line 349: This update `root_node_` update could happen while `root_node_ptr->binding_buffer_device_address` is already in use by a submitted command buffer? 
I may just not follow things correctly, but with our current descriptor set system, we ensure that each draw receives its own descriptor set, thus in a way "its own pointer to the BDA table".

With this update system, each draw always looks at `root_node_ptr->binding_buffer_device_address`, but they *all* share this `root_node_ptr`, so modifying it for one will modify it for other

arno-lunarg (commented):
Not sure it actually works?

spencer-lunarg:
closing as will need to retry this

PR #9474, opened 2025-02-18
gpuav: Make Module Setting fields not be repeated

spencer-lunarg:
Noticed while doing https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9471

We have this `Settings` struct, but were still duplicating everything inside another private field member of `Module`

### DIFF
--- a/layers/gpuav/spirv/module.cpp
+++ b/layers/gpuav/spirv/module.cpp
@@ -32,13 +32,9 @@ Module::Module(vvl::span<const uint32_t> words, DebugReport* debug_report, const
                const DeviceFeatures& enabled_features,
                const std::vector<std::vector<BindingLayout>>& set_index_to_bindings_layout_lut)
     : type_manager_(*this),
-      max_instrumentations_count_(settings.max_instrumentations_count),
-      shader_id_(settings.shader_id),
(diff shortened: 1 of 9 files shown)


layers/gpuav/spirv/module.cpp
```diff
@@ -32,13 +32,9 @@ Module::Module(vvl::span<const uint32_t> words, DebugReport* debug_report, const
                const DeviceFeatures& enabled_features,
                const std::vector<std::vector<BindingLayout>>& set_index_to_bindings_layout_lut)
     : type_manager_(*this),
-      max_instrumentations_count_(settings.max_instrumentations_count),
-      shader_id_(settings.shader_id),
-      output_buffer_descriptor_set_(settings.output_buffer_descriptor_set),
-      support_non_semantic_info_(settings.support_non_semantic_info),
+      settings_(settings),
       enabled_features_(enabled_features),
       has_bindless_descriptors_(settings.has_bindless_descriptors),
```
spencer-lunarg, line 37: this one is a `TODO` in the field, this is non-const and can now re-work the way we pass in the LUT and that will just remove this one

Issue #9317, opened 2025-01-26
Crash: Samsung S24 seems to crash with validation layer


Nielsbishere:
**Environment:**
 - OS: Android 14 (Samsung S24)
 - GPU and driver version:  Xclipse 940 (build number UP1A.231005.007.S921BXXS5AXKA)
 - SDK or header version if building from repo: latest (happens on either main manual build or the release)
 - Options enabled (synchronization, best practices, etc.): Yes (all of them)

**Describe the Issue**

Creating this specific spirv compute pipeline (vkCreateComputePipelines) will crash the validation layer on android. Both HLSL and SPIRV are available and of course a minimum repro as well.
This seems to be android specific and the validation layer supplied with my current desktop environment seems to behave fine with both the minimum repro and the real final executable.

**Expected behavior**

Not crashing.

**Valid Usage ID**
```
01-26 00:42:24.033 22352 22401 W OxC3    : Validation Warning: [ WARNING-GPU-Assisted-Validation ] Object 0: handle = 0xb400006f7760c110, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0x24b5c69f | vkCreateDevice(): Internal Warning: Forcing fragmentStoresAndAtomics to VK_TRUE
01-26 00:42:24.033 22352 22401 W OxC3    : Validation Warning: [ WARNING-GPU-Assisted-Validation ] Object 0: handle = 0xb400006f7760c110, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0x24b5c69f | vkCreateDevice(): Internal Warning: Forcing vertexPipelineStoresAndAtomics to VK_TRUE
01-26 00:42:24.033 22352 22401 W OxC3    : Validation Warning: [ WARNING-GPU-Assisted-Validation ] Object 0: handle = 0xb400006f7760c110, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0x24b5c69f | vkCreateDevice(): Internal Warning: Forcing shaderInt64 to VK_TRUE
[...]

PR #9475, opened 2025-02-18
tests: Add new stress test

spencer-lunarg:
(inspired by https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9317) New stress tests and some numbers

This is O(n^2) level bad and why we will need a fix for those that are unfortunate and don't get `DontInline` (the worst part is there is no way to query the driver if it "actually" supports it)

(Time it took to run the test, which checked, it is reflecting **Pipeline Compiling times**)

No instrumentation - **0.2 seconds**

50 instrumentation of PostProcess - **0.2 seconds**
50 instrumentation of GeneralBuffer - **0.6 seconds**
50 instrumentation of DescriptorIndexing - **1.2 seconds**
50 instrumentation of PostProcess/GeneralBuffer/DescriptorIndexing - **2.0 seconds**

100 instrumentation of PostProcess - **0.3 seconds**
100 instrumentation of GeneralBuffer - **1.8 seconds**
100 instrumentation of DescriptorIndexing - **4.6 seconds**
100 instrumentation of PostProcess/GeneralBuffer/DescriptorIndexing - **8.5 seconds**

300 instrumentation of PostProcess - **0.4 seconds**
300 instrumentation of GeneralBuffer - **25.5 seconds**
300 instrumentation of DescriptorIndexing - **86.2 seconds**
300 instrumentation of PostProcess/GeneralBuffer/DescriptorIndexing - **164.0 seconds**

### DIFF
--- a/tests/unit/gpu_av_descriptor_indexing_positive.cpp
+++ b/tests/unit/gpu_av_descriptor_indexing_positive.cpp
@@ -828,8 +828,9 @@ TEST_F(PositiveGpuAVDescriptorIndexing, SampledImageShareBindingBDA) {
     m_default_queue->Wait();
 }
 
+// Disabled as this is a perf testing test, not much value in normal CI
 // If on Mesa, also add MESA_SHADER_CACHE_DISABLE=1
-TEST_F(PositiveGpuAVDescriptorIndexing, Stress) {
+TEST_F(PositiveGpuAVDescriptorIndexing, DISABLED_Stress) {
     TEST_DESCRIPTION("Do many indexing into the shader");
     RETURN_IF_SKIP(InitGpuVUDescriptorIndexing());
     InitRenderTarget();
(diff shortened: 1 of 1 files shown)


tests/unit/gpu_av_descriptor_indexing_positive.cpp
```diff
@@ -944,6 +944,71 @@ TEST_F(PositiveGpuAVDescriptorIndexing, Stress) {
     m_default_queue->Wait();
 }
 
+TEST_F(PositiveGpuAVDescriptorIndexing, Stress2) {
```
arno-lunarg, line 947: Should it be disabled? Useful for local testing, but unless test time gets monitored during CI runs, this test does not add much value?

arno-lunarg (approved):

tests/unit/gpu_av_descriptor_indexing_positive.cpp
spencer-lunarg: the other test is there, I guess they could just be DISABLED, let me do that (to booth)

PR #9476, opened 2025-02-19
vvl: Fix crash

arno-lunarg:
3rd attempt, closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9412
Handle the fact that GPU-AV can be removed
from the object_dispatch list


layers/chassis/dispatch_object_manual.cpp
```diff
@@ -718,7 +718,8 @@ void Device::ReleaseValidationObject(LayerObjectTypeId type_id) const {
         if ((*object_it)->container_type == type_id) {
             auto object = std::move(*object_it);
 
-            object_dispatch.erase(object_it);
+            // Do not remove the smart pointer, to not invalidate object_dispatch iterators
+            // object_dispatch.erase(object_it);
```
arno-lunarg, line 722: So no more vector erasure, and iterators should be preserved

jeremyg-lunarg: couldn't we have the same problem in the intercept vectors (line 728/729)?

arno-lunarg: aaaah looks like so? It did not blow up when testing though ...

arno-lunarg: I will make the same change here

jeremyg-lunarg: it would be much harder to blow up here. GPU-AV would have to abort from some other random call not vkCreateDevice

scripts/generators/layer_chassis_generator.py
```diff
@@ -329,6 +329,10 @@ def generateSource(self):
                 out.append('    auto lock = vo->ReadLock();\n')
             else:
                 out.append(f'for (const auto& vo : {dispatch}->object_dispatch) {{\n')
+                out.append('    if (!vo) {\n')
```
spencer-lunarg, line 332: thoughts? 

```suggestion
                out.append('    if (!vo) continue;\n')
```

arno-lunarg: for debugging reasons, I want `continue;` on its own line, to easily be able to set a break point

arno-lunarg:
Added a new test, and made sure it reproduces the observed crash without the fix

tests/unit/gpu_av.cpp
```diff
@@ -616,3 +616,33 @@ TEST_F(NegativeGpuAV, UseAllDescriptorSlotsPipelineLayout) {
...
+    validation_features.pEnabledValidationFeatures = validation_enables.data();
+    if (m_gpuav_disable_core) {
+        validation_features.disabledValidationFeatureCount = size32(validation_disables);
+        validation_features.pDisabledValidationFeatures = validation_disables.data();
+    }
+
+    RETURN_IF_SKIP(InitFramework(&validation_features));
+
+    m_errorMonitor->SetDesiredError("UNASSIGNED-GPU-Assisted-Validation");
+    RETURN_IF_SKIP(InitState(nullptr));
```
spencer-lunarg, line 646: did you want (for completeness) do a basic dispatch or something to show even after `vkCreateDevice` it still doesn't blow up

jeremyg-lunarg (approved):

PR #9478, opened 2025-02-19
sync: Refactor clear attachment code

artem-lunarg:
Some improvements of how clear attachment code is structured before updating corresponding error messages.

### DIFF
--- a/layers/sync/sync_commandbuffer.cpp
+++ b/layers/sync/sync_commandbuffer.cpp
@@ -675,6 +675,8 @@ bool CommandBufferAccessContext::ValidateDrawAttachment(const Location &loc) con
 }
 
 bool CommandBufferAccessContext::ValidateDrawDynamicRenderingAttachment(const Location &location) const {
+    // TODO: Add tests. This is never called by existing tests.
+    // TODO: Check for opportunities to improve error message after this covered by the tests.
     bool skip = false;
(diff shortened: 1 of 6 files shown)


layers/sync/sync_renderpass.h
```diff
@@ -75,30 +74,6 @@ struct BeginRenderingCmdState {
 void InitSubpassContexts(VkQueueFlags queue_flags, const vvl::RenderPass &rp_state, const AccessContext *external_context,
                          std::vector<AccessContext> &subpass_contexts);
 
-struct ClearAttachmentInfo {
```
artem-lunarg, line 78: One change is that this structure was updated (and moved to sync_commandbuffer.h) and now is a simple 5 LOC type. The code reorganization allowed to replace some pointer with references.

layers/sync/sync_commandbuffer.cpp
```diff
@@ -675,6 +675,8 @@ bool CommandBufferAccessContext::ValidateDrawAttachment(const Location &loc) con
 }
 
 bool CommandBufferAccessContext::ValidateDrawDynamicRenderingAttachment(const Location &location) const {
+    // TODO: Add tests. This is never called by existing tests.
```
arno-lunarg, line 678: Maybe create a corresponding issue?

arno-lunarg (approved):

layers/sync/sync_commandbuffer.cpp
artem-lunarg: yes

Issue #9129, opened 2025-01-04
`OpTypeImage`'s format mismatching with that of the bound VkImageView is ignored.


nicebyte:
**Environment:**
SDK ver 1.3.296

**Describe the Issue**

Consider the following HLSL:

```
[[vk::binding(0, 0)]] RWTexture2D<float4> outputImage;

[numthreads(4, 4, 1)] void CSMain(uint3 tid
                                  : SV_DispatchThreadID) {
  outputImage[tid.xy] = float4(sin(tid.x),cos(tid.y),0.0,1.0);
}
```

note the absence of a format qualifier for `outputImage`.

This code produces the following SPIR-V:

```
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %CSMain "CSMain" %gl_GlobalInvocationID
               OpExecutionMode %CSMain LocalSize 4 4 1
               OpDecorate %gl_GlobalInvocationID BuiltIn GlobalInvocationId
               OpDecorate %outputImage DescriptorSet 0
               OpDecorate %outputImage Binding 0
      %float = OpTypeFloat 32
    %float_0 = OpConstant %float 0
    %float_1 = OpConstant %float 1
%type_2d_image = OpTypeImage %float 2D 2 0 0 2 Rgba32f
%_ptr_UniformConstant_type_2d_image = OpTypePointer UniformConstant %type_2d_image
       %uint = OpTypeInt 32 0
     %v3uint = OpTypeVector %uint 3
%_ptr_Input_v3uint = OpTypePointer Input %v3uint
       %void = OpTypeVoid
         %15 = OpTypeFunction %void
    %v4float = OpTypeVector %float 4
     %v2uint = OpTypeVector %uint 2
%outputImage = OpVariable %_ptr_UniformConstant_type_2d_image UniformConstant
%gl_GlobalInvocationID = OpVariable %_ptr_Input_v3uint Input
     %CSMain = OpFunction %void None %15
         %18 = OpLabel
         %19 = OpLoad %v3uint %gl_GlobalInvocationID
         %20 = OpCompositeExtract %uint %19 0
         %21 = OpConvertUToF %float %20
         %22 = OpExtInst %float %1 Sin %21
         %23 = OpCompositeExtract %uint %19 1
         %24 = OpConvertUToF %float %23
         %25 = OpExtInst %float %1 Cos %24
[...]

PR #9487, opened 2025-02-20
layers: Add warning for storage image formats

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9129
(Discussed in https://gitlab.khronos.org/vulkan/vulkan/-/issues/4128)

This will now properly warn the user when they accidentally mix the formats up and are going to get the wrong values


layers/drawdispatch/descriptor_validator.cpp
```diff
@@ -364,32 +368,49 @@ bool DescriptorValidator::ValidateDescriptor(const spirv::ResourceInterfaceVaria
                                       string_VkImageViewType(image_view_ci.viewType), string_SpvDim(dim), is_image_array);
         }
 
-        if ((resource_variable.info.image_format_type & image_view_state->descriptor_format_bits) == 0) {
+        const uint32_t view_numeric_type = spirv::GetFormatType(image_view_ci.format);
+        const uint32_t variable_numeric_type = resource_variable.info.image_sampled_type_numeric;
+        // When the image has a external format the views format must be VK_FORMAT_UNDEFINED and it is required to use a sampler
+        // Ycbcr conversion. Thus we can't extract any meaningful information from the format parameter.
+        if (image_view_ci.format != VK_FORMAT_UNDEFINED && ((variable_numeric_type & view_numeric_type) == 0)) {
```
spencer-lunarg, line 375: This is replacing the `descriptor_format_bits(im->HasAHBFormat(), ...` logic we had before, in theory something in the future that is not AHB could also have `VK_FORMAT_UNDEFINED`

layers/drawdispatch/descriptor_validator.cpp
```diff
@@ -909,7 +930,7 @@ bool DescriptorValidator::ValidateDescriptor(const spirv::ResourceInterfaceVaria
         }
     }
 
-    return false;
+    return skip;
```
spencer-lunarg, line 933: I have a todo after this to fix these function to use `skip` instead of early returning

layers/state_tracker/shader_module.cpp
```diff
@@ -1997,14 +1998,6 @@ const Instruction& ResourceInterfaceVariable::FindBaseType(ResourceInterfaceVari
     return *type;
 }
 
-uint32_t ResourceInterfaceVariable::FindImageSampledTypeWidth(const Module& module_state, const Instruction& base_type) {
```
spencer-lunarg, line 2000: Just moved the logic inline where it is used

layers/drawdispatch/descriptor_validator.cpp
```diff
@@ -364,32 +368,48 @@ bool DescriptorValidator::ValidateDescriptor(const spirv::ResourceInterfaceVaria
...
-                                          string_VkFormat(image_view_ci.format));
+                                          spirv::string_NumericType(variable_numeric_type), string_VkFormat(image_view_ci.format));
             }
         }
 
+        if (image_view_ci.format != VK_FORMAT_UNDEFINED && resource_variable.info.image_format != VK_FORMAT_UNDEFINED &&
+            image_view_ci.format != resource_variable.info.image_format) {
+            auto set = descriptor_set.Handle();
+            const LogObjectList objlist(set, image_view);
+            skip |= dev_state.LogUndefinedValue(
```
artem-lunarg, line 394: just to confirm that this will be reported in the standard configuration according recent PRs?

artem-lunarg (approved):

layers/drawdispatch/descriptor_validator.cpp
spencer-lunarg: yes, it still gets reported, after this, going to just make this function all use `skip |=` and there will be no confusion for the next person

PR #9491, opened 2025-02-20
layers: Cleanup extension name check

spencer-lunarg:
@jeremyg-lunarg @mikes-lunarg @charles-lunarg thoughts on this

I found that 3 of these 4 the desktop loader just takes care of for you (not 100% sure about Android loader)

Tried to provide a better message for those that might hit this somehow


charles-lunarg (approved):
The loader will ignore unknown instance extensions if asked nicely. The loader doesn't care about known device extensions, so this check is still very useful.

jeremyg-lunarg (approved):

Issue #9428, opened 2025-02-13
Bad performance around validation for acceleration structures


Jasper-Bekkers:
![Image](https://github.com/user-attachments/assets/18a4e192-71ae-41b5-9c45-c872aa953c57)

This ends up taking around ~20s with the validation layer and less then 1s without.

arno-lunarg:
Hello! Do you have a code sample for me to try and reproduce this, or a GFXReconstruct trace I could look at please?

arno-lunarg:
Or actually if you could just further extend the call stacks in Superliminal so I can have more information on what part of the the validation function is causing issues

Jasper-Bekkers:
![Image](https://github.com/user-attachments/assets/85fb31ec-adc7-4fce-b080-d35aff78f8af)

I can also send you the superluminal capture entirely if you'd like. (Reminder to self: `ValidationLayer.slp`)

Jasper-Bekkers:
Seems to be a few odd things are happening at least. 

- There's code referencinging `VkDisplayModeKHR_T` is in here, 
- It seems to be spending a lot of time formatting strings in `ValidateScratchMemoryNoOverlap` even though there aren't any errors being reported.
- `LogErrorsIfNoValidBuffer` takes significant time.
- There's also significant time being spent initializing a `save_VkAccelerationStructureGeometryKHR` object
- There are 4 driver calls being made in `rt::ComputeBuildSizes`.

arno-lunarg:
Thank you very much for those details! Local testing showed the same issues, working on it... If you can send the trace at `arno AT lunarg DOT com` that would be a plus, but not necessary

PR #9492, opened 2025-02-20
vvl: Rework accel structs mem aliasing validation

arno-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9428

In acceleration structures build commands, when looking for overlaps between source acceleration structures, destination acceleration structures, and scratch buffers, do it using device address ranges instead of relying on the device memory objects, and sort device address ranges.
=> Since overlap detection is done when a new device address range is inserted among the previously seen ones, and insertion is done using dichotomy (faster and ok since address ranges are sorted), not all overlap will be detected, but if there is any, it will be detected.
For instance, if all ranges are equal, inserting yet another equal range will likely result in it being directly placed at the beginning or end of the list, with only one overlap detected with first of last element.
Given the performance gains, and the fact that most calls done by apps to `vkCmdBuildAccelerationStructuresKHR` are expected to be valid so no memory overlaps will be detected, this compromise is well worth it.

Doing some stress test, where I build 8000 acceleration structures in one go, `PreCallValidateCmdBuildAccelerationStructuresKHR` went from ~4 minutes down to ~200 ms, and the overlap validation itself is about ~100 ms

Still need to do the same change at other spots, but this one in particular was the issue, so let's get this in first

### DIFF
--- a/layers/core_checks/cc_ray_tracing.cpp
+++ b/layers/core_checks/cc_ray_tracing.cpp
@@ -29,6 +29,7 @@
 #include "error_message/error_strings.h"
 
 #include <algorithm>
+#include <iostream>
 
 bool CoreChecks::ValidateInsertAccelerationStructureMemoryRange(VkAccelerationStructureNV as, const vvl::DeviceMemory &mem_info,
                                                                 VkDeviceSize mem_offset, const Location &loc) const {
@@ -270,6 +271,8 @@ bool CoreChecks::ValidateAccelerationStructuresDeviceScratchBufferMemoryAliasing
             src_as_state && info.mode == VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR) {
             const vvl::range<VkDeviceAddress> src_as_range = src_as_state->GetDeviceAddressRange();
 
+            std::cout << "pInfos[" << info_i << "].src_as addr range: " << string_range_hex(src_as_range) << std::endl;
+
             if (dst_as_state && dst_as_state->VkHandle() != src_as_state->VkHandle()) {
                 const vvl::range<VkDeviceAddress> dst_as_range = dst_as_state->GetDeviceAddressRange();
 
@@ -327,6 +330,10 @@ bool CoreChecks::ValidateAccelerationStructuresDeviceScratchBufferMemoryAliasing
         if (dst_as_state) {
             const AddressRange dst_as_address_range = {dst_as_state->GetDeviceAddressRange(), info_i,
                                                        AddressRangeOrigin::DstAccelStruct};
+
+            std::cout << "pInfos[" << info_i << "].dst_as addr range: " << string_range_hex(dst_as_address_range.range)
+                      << std::endl;
+
             const std::optional<AddressRange> overlapped_address_range = insert_address(address_ranges, dst_as_address_range);
             if (overlapped_address_range.has_value()) {
                 switch (overlapped_address_range->origin) {
@@ -387,6 +394,9 @@ bool CoreChecks::ValidateAccelerationStructuresDeviceScratchBufferMemoryAliasing
                                                                info.scratchData.deviceAddress + assumed_scratch_size};
 
             const AddressRange scratch_address_range = {scratch_range, info_i, AddressRangeOrigin::Scratch};
+
+            std::cout << "pInfos[" << info_i << "].scratch addr range: " << string_range_hex(scratch_range) << std::endl;
+
             const std::optional<AddressRange> overlapped_address_range = insert_address(address_ranges, scratch_address_range);
             if (overlapped_address_range.has_value()) {
                 switch (overlapped_address_range->origin) {
--- a/layers/core_checks/cc_ray_tracing.cpp
+++ b/layers/core_checks/cc_ray_tracing.cpp
@@ -28,6 +28,8 @@
 #include "state_tracker/ray_tracing_state.h"
 #include "error_message/error_strings.h"
 
+#include <algorithm>
+
 bool CoreChecks::ValidateInsertAccelerationStructureMemoryRange(VkAccelerationStructureNV as, const vvl::DeviceMemory &mem_info,
                                                                 VkDeviceSize mem_offset, const Location &loc) const {
     return ValidateInsertMemoryRange(VulkanTypedHandle(as, kVulkanObjectTypeAccelerationStructureNV), mem_info, mem_offset, loc);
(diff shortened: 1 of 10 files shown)


arno-lunarg:
The number of detected overlaps in tests will also now depends on how `std::lower_bound` is implemented, hopefully every implementation works the same way

artem-lunarg:
@arno-lunarg `std::lower_bound` (and other stl algorithm) has some guarantees how it works. We can rely on them. If that's something not coverd by those rules (or can't be derived from them) then we can't.
https://en.cppreference.com/w/cpp/algorithm/lower_bound

tests/unit/ray_tracing.cpp
```diff
@@ -2044,15 +2049,15 @@ TEST_F(NegativeRayTracing, AccelerationStructuresOverlappingMemory3) {
...
-            m_errorMonitor->SetDesiredError("VUID-vkCmdBuildAccelerationStructuresKHR-scratchData-03704");
-        }
-        for (size_t i = 0; i < build_info_count; ++i) {
-            m_errorMonitor->SetDesiredError("VUID-vkCmdBuildAccelerationStructuresKHR-dstAccelerationStructure-03703");
-        }
+        // 03704 *should* be triggered for each pair of elements in `build_infos`. 03703 *should* also be triggered for individual
+        // elements.
+        // => due to validation code optimisations, not *all* overlaps will be detected,
+        // but if there is *at least one*, it will *always+ be detected.
+        m_errorMonitor->SetDesiredError("VUID-vkCmdBuildAccelerationStructuresKHR-dstAccelerationStructure-03703");
```
spencer-lunarg, line 2056:
```suggestion
        m_errorMonitor->SetDesiredError("VUID-vkCmdBuildAccelerationStructuresKHR-dstAccelerationStructure-03703", 3);
```

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -215,67 +217,219 @@ bool CoreChecks::ValidateAccelerationStructuresMemoryAlisasing(const LogObjectLi
...
+            const AddressRange src_as_address_range = {src_as_range, info_i, AddressRangeOrigin::SrcAccelStruct};
+            const std::optional<AddressRange> overlapped_address_range = insert_address(address_ranges, src_as_address_range);
+            if (overlapped_address_range.has_value()) {
+                switch (overlapped_address_range->origin) {
+                    case AddressRangeOrigin::SrcAccelStruct: {
+                        // Valid overlap, source acceleration structures being read only
+                        break;
+                    }
+                    case AddressRangeOrigin::DstAccelStruct: {
+                        const auto other_dst_as_state =
```
spencer-lunarg, line 295: can this be null?

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -215,67 +217,219 @@ bool CoreChecks::ValidateAccelerationStructuresMemoryAlisasing(const LogObjectLi
...
+                            LogError("VUID-vkCmdBuildAccelerationStructuresKHR-dstAccelerationStructure-03701",
+                                     LogObjectList(cmd_buffer, dst_as_state->VkHandle(), other_src_as_state->VkHandle()),
+                                     info_i_loc.dot(Field::dstAccelerationStructure),
+                                     ", overlaps with srcAccelerationStructure of pInfos[%" PRIu32
+                                     "], which is in update mode, on device address range %s.",
+                                     overlapped_address_range->info_i, string_range_hex(dst_as_other_src_as_intersection).c_str());
+                        break;
+                    }
+                    case AddressRangeOrigin::DstAccelStruct: {
+                        const auto other_dst_as_state =
```
spencer-lunarg, line 348: same, can this be null?

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -215,67 +217,219 @@ bool CoreChecks::ValidateAccelerationStructuresMemoryAlisasing(const LogObjectLi
...
+                    default:
+                        assert(false);
+                        break;
+                }
+            }
+        }
+
+        {
+            const rt::BuildType rt_build_type =
+                error_obj.location.function == Func::vkBuildAccelerationStructuresKHR ? rt::BuildType::Host : rt::BuildType::Device;
```
spencer-lunarg, line 380: can this be removed, this check is only for device call now I think

layers/state_tracker/state_tracker.cpp
```diff
@@ -368,6 +368,21 @@ std::shared_ptr<Buffer> Device::CreateBufferState(VkBuffer handle, const VkBuffe
     return std::make_shared<Buffer>(*this, handle, create_info);
 }
 
+void Device::PreCallRecordCreateBuffer(VkDevice device, const VkBufferCreateInfo *pCreateInfo,
+                                       const VkAllocationCallbacks *pAllocator, VkBuffer *pBuffer, const RecordObject &record_obj,
+                                       chassis::CreateBuffer &chassis_state) {
+    if (pCreateInfo->usage & VK_BUFFER_USAGE_ACCELERATION_STRUCTURE_STORAGE_BIT_KHR) {
```
spencer-lunarg, line 374: this is playing a bit with fire, and there is a chance this might cause a false positive if we have now altered what the drive reports back and the user will see a different result with and without validation

Another option/idea is if they don't have the `VK_BUFFER_USAGE_SHADER_DEVICE_ADDRESS_BIT` flag, we just don't check it.

I guess personally want to be on the side of "no false positive" to trying to validate everything, but at the same time, not sure if this will "save many mistakes" that otherwise would be missed if we don't enforce it

arno-lunarg: Yes it is a compromise, for now I deem it well worth it. VVL will still report an error when application forgot to set this flag, I don't expect major problems around here, but we will see.
It is impossible to not have shader device address when using acceleration structure, at least because you need them for the scratch buffer

layers/core_checks/cc_ray_tracing.cpp
arno-lunarg: yes indeed!

layers/state_tracker/state_tracker.cpp
arno-lunarg: And this would not be the first time a debug/validation tool modifies the behavior of the targeted app, just look at GPU-AV, we do things far more horrible!

spencer-lunarg: yes, but this is not GPU-AV and we have had issues where pipeline creation was setting ignored fields to null and caused lots of headaches

basically people don't expect core validation to alter things, just a read only validation... again, should be fine, but will keep our eyes on the lookout for trouble

spencer-lunarg (approved):
LGTM - when null check comments are addressed

layers/core_checks/cc_ray_tracing.cpp
spencer-lunarg: just a `ASSERT_AND_RETURN` type thing should be good

layers/state_tracker/state_tracker.cpp
arno-lunarg: Yes I will keep this change in mind should ray tracing weirdness come

arno-lunarg:
After getting some info back from the working group, it seems that relying on device addresses to look for memory overlaps will not work, at least because of this:
If you have 2 sparse buffers, bound to the same memory, they can still return different base device addresses....

So I need to revamp this PR

Issue #9386, opened 2025-02-04
SDK 1.4.304.0 breaks VVL DEBUG_PRINTF feature on macOS


SRSaunders:
**Environment:**
 - OS: macOS Ventura
 - GPU and driver version: AMD 6600XT
 - SDK or header version if building from repo: SDK 1.4.304.0
 - Options enabled (synchronization, best practices, etc.):

**Describe the Issue**

The VVL version contained in SDK 1.4.304.0 now implicitly requests `VkPhysicalDeviceVulkanMemoryModelFeatures::vulkanMemoryModel` and `VkPhysicalDeviceVulkanMemoryModelFeatures::vulkanMemoryModelDeviceScope` device features when enabling the VVL DEBUG_PRINTF feature.  This was not the case in previous SDKs (e.g. 1.3.296.0 and earlier).  Unfortunately this breaks the VVL DEBUG_PRINTF feature on macOS/MoltenVK where these device features are not available.

**Expected behavior**

Previously these `VkPhysicalDeviceVulkanMemoryModelFeatures` device features were not implicitly requested and the DEBUG_PRINTF feature worked properly on macOS.  Are these features absolutely necessary for VVL DEBUG_PRINTF to work with new SDKs?

**Valid Usage ID**
If applicable, please include the validation messages encountered leading up to the issue

```
[warning] 1985515673 - WARNING-DEBUG-PRINTF: Validation Warning: [ WARNING-DEBUG-PRINTF ] | MessageID = 0x76589099 | vkCreateDevice(): Internal Warning: Adding a VkPhysicalDeviceVulkanMemoryModelFeatures to pNext with vulkanMemoryModel and vulkanMemoryModelDeviceScope set to VK_TRUE

[error] 0 - mvk-error: VK_ERROR_FEATURE_NOT_PRESENT: vkCreateDevice(): Requested physical device feature specified by the 1st flag in VkPhysicalDeviceVulkanMemoryModelFeatures is not available on this device.
[error] 0 - mvk-error: VK_ERROR_FEATURE_NOT_PRESENT: vkCreateDevice(): Requested physical device feature specified by the 2nd flag in VkPhysicalDeviceVulkanMemoryModelFeatures is not available on this device.

```

spencer-lunarg:
[...]

PR #9494, opened 2025-02-20
gpuav: Cleanup how feature are forced on and checked

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9386

We still need a more stream line way to add features, but this tries to 

1. Fix the bug forcing `vulkanMemoryModel` without checking
2. Makes `bufferDeviceAddress` a proper requirement now
3. Tries to start creating a way to one day just have a simple, automatic way to add a feature needed


layers/gpuav/core/gpuav_features.cpp
```diff
@@ -0,0 +1,258 @@
...
+            return true;
+        }
+    }
+
+    return false;
+}
+
+// In PreCallRecord we try to turn on as many features as possible on behalf of the app (and warn the user if we are doing it).
+// Later after device creation, we can decide if the required Vulkan feature for each GPU-AV setting is found and report errors
+void Instance::AddFeatures(VkPhysicalDevice physical_device, vku::safe_VkDeviceCreateInfo *modified_create_info,
```
spencer-lunarg, line 52: this is an ugly function, should/could be auto generated, but for now, moved to own file

layers/gpuav/core/gpuav_settings.h
```diff
@@ -47,13 +47,13 @@ struct GpuAVSettings {
         bool descriptor_checks = true;
         bool buffer_device_address = true;
         bool ray_query = true;
-        bool post_process_descriptor_index = true;
+        bool post_process_descriptor_indexing = true;
```
spencer-lunarg, line 50: this was not matching the name `indexing` we use everywhere else

layers/gpuav/core/gpuav_setup.cpp
```diff
@@ -554,59 +319,69 @@ void Validator::FinishDeviceSetup(const VkDeviceCreateInfo *pCreateInfo, const L
...
+}  // namespace setting
+
+// At this point extensions/features may have been turned on by us in PreCallRecord.
+// Now that we have all the information, here is where we might disable GPU-AV settings that are missing requirements
+void Validator::InitSettings(const Location &loc) {
+    setting::BufferDeviceAddress buffer_device_address;
+    setting::RayQuery ray_query;
+    setting::BufferCopies buffer_copies;
+    setting::BufferContent buffer_content;
+    std::array<setting::Setting *, 4> all_settings = {&buffer_device_address, &ray_query, &buffer_copies, &buffer_content};
```
spencer-lunarg, line 379: this is my attempt to try and make it more streamline to add a feature, it is better than the manual mess we had, but still not the final form I would want

layers/gpuav/core/gpuav_features.cpp
```diff
@@ -0,0 +1,258 @@
...
+                           const Location &loc) {
+    // Query things here to make sure we don't attempt to add a feature this is just not supported
+    VkPhysicalDevice8BitStorageFeatures supported_8bit_feature = vku::InitStructHelper();
+    VkPhysicalDeviceBufferDeviceAddressFeatures supported_bda_feature = vku::InitStructHelper(&supported_8bit_feature);
+    VkPhysicalDeviceVulkanMemoryModelFeatures supported_memory_model_feature = vku::InitStructHelper(&supported_bda_feature);
+    VkPhysicalDeviceTimelineSemaphoreFeatures supported_timeline_feature = vku::InitStructHelper(&supported_memory_model_feature);
+    VkPhysicalDeviceFeatures2 features_2 = vku::InitStructHelper(&supported_timeline_feature);
+    DispatchGetPhysicalDeviceFeatures2(physical_device, &features_2);
+
+    // Fist core features
```
jeremyg-lunarg, line 62: Fist?

layers/gpuav/core/gpuav_features.cpp
```diff
@@ -0,0 +1,258 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "gpuav/core/gpuav.h"
+
+namespace gpuav {
+
+static std::vector<VkExtensionProperties> GetExtensions(VkPhysicalDevice physical_device) {
```
jeremyg-lunarg, line 22: Why not use `DeviceExtensions`?

layers/gpuav/instrumentation/gpuav_shader_instrumentor.cpp
```diff
@@ -71,20 +71,35 @@ WriteLockGuard GpuShaderInstrumentor::WriteLock() {
 void GpuShaderInstrumentor::FinishDeviceSetup(const VkDeviceCreateInfo *pCreateInfo, const Location &loc) {
     BaseClass::FinishDeviceSetup(pCreateInfo, loc);
 
-    VkPhysicalDeviceFeatures supported_features{};
-    DispatchGetPhysicalDeviceFeatures(physical_device, &supported_features);
-    if (!supported_features.fragmentStoresAndAtomics) {
+    // Check hard requirements for GPU-AV (after BaseClass::FinishDeviceSetup as set enabled_features)
```
jeremyg-lunarg, line 74: should this be "has set"?

layers/gpuav/core/gpuav_features.cpp
spencer-lunarg: I just moved this here, probably predated, will try to use it

jeremyg-lunarg: You can make a DeviceExtensions that represents what the PhysicalDevice supports:
```
   DeviceExtensions(const InstanceExtensions &instance_ext, APIVersion requested_api_version,
                     const std::vector<VkExtensionProperties> &props);
```
This is used in stateless. Maybe it could be streamlined even more.

layers/gpuav/core/gpuav_features.cpp
spencer-lunarg: I tried, the single issue was we do a `InternalWarning("We forced this on for you")` here and couldn't quickly come up with a way to merge it... but overall I moved this code to a new file so if we do use auto-generated stuff, it is just removing this file

Issue #8627, opened 2024-10-01
VUID-VkExportMemoryAllocateInfo-handleTypes-00656 broken


zmike:
VVL attempts to validate this at vkBindImageMemory time, but this is invalid, as there is no way to pass e.g., `VkImageFormatListCreateInfo` which is needed by drivers to validate format properties.

zmike:
Also affects `VUID-VkImportMemoryWin32HandleInfoKHR-handleType-00658`

spencer-lunarg:
.... this again, seems we never actually added a "full solution" last time :cry:

zmike:
Just hit it again today.

spencer-lunarg:
@zmike I think we squashed this, like I took a big hammer out and tried smashing it multiple times, if it still is there, let me know

PR #9503, opened 2025-02-21
layers: Fix pNext chaining for DRM

spencer-lunarg:
hopefully closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8627

### DIFF
--- a/layers/core_checks/cc_device_memory.cpp
+++ b/layers/core_checks/cc_device_memory.cpp
@@ -27,6 +27,7 @@
 #include "state_tracker/image_state.h"
 #include "state_tracker/buffer_state.h"
 #include "state_tracker/ray_tracing_state.h"
+#include "error_message/error_strings.h"
 
 // For given mem object, verify that it is not null or UNBOUND, if it is, report error. Return skip value.
 bool CoreChecks::VerifyBoundMemoryIsValid(const vvl::DeviceMemory *memory_state, const LogObjectList &objlist,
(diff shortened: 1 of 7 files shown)


layers/core_checks/cc_device_memory.cpp
```diff
@@ -383,6 +384,25 @@ bool CoreChecks::HasExternalMemoryImportSupport(const vvl::Image &image, VkExter
     info.usage = image.create_info.usage;
     info.flags = image.create_info.flags;
 
+    // TODO - Want to use vvl::PnextChainExtract, but would need to cleanup (and test) rest of how we add the other pNext here
+    VkImageFormatListCreateInfo format_list = vku::InitStructHelper();
+    if (auto original_format_list = vku::FindStructInPNextChain<VkImageFormatListCreateInfo>(image.create_info.pNext)) {
+        format_list.pViewFormats = original_format_list->pViewFormats;
+        format_list.viewFormatCount = original_format_list->viewFormatCount;
+        vvl::PnextChainAdd(&external_info, &format_list);
```
artem-lunarg, line 392: Can we add directly looked up pnext structures instead of initializing a fresh copies? Is it to avoid using potentially not null pNext from pNext?

tests/unit/image_drm_positive.cpp
```diff
@@ -282,7 +261,99 @@ TEST_F(PositiveImageDrm, MutableFormat) {
...
+    VkFormat formats = VK_FORMAT_R8G8B8A8_SNORM;
+    VkImageFormatListCreateInfo format_list = vku::InitStructHelper();
+    format_list.viewFormatCount = 1;
+    format_list.pViewFormats = &formats;
+
+    VkExternalMemoryImageCreateInfo external_info = vku::InitStructHelper(&format_list);
+    external_info.handleTypes = VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD_BIT;
+
+    // handleTypes needs to be assigned to trigger the behavior we want
+    assert(external_info.handleTypes);
```
artem-lunarg, line 297: supernit: these two lines can be removed, there is initializatoin above.

layers/core_checks/cc_device_memory.cpp
spencer-lunarg: will add a comment about that here

artem-lunarg: makes sense, thanks

PR #9511, opened 2025-02-24
sync: Validate acceleration structure vertex/index input buffers

artem-lunarg:


layers/sync/sync_validation.cpp
```diff
@@ -3648,6 +3648,74 @@ void SyncValidator::PostCallRecordGetSwapchainImagesKHR(VkDevice device, VkSwapc
...
+    const vvl::Buffer *index_data = nullptr;
+    ResourceAccessRange index_range;
+};
+
+static std::optional<AccelerationStructureGeometryInfo> GetValidGeometryInfo(
+    const vvl::Device &device, const VkAccelerationStructureGeometryKHR &geometry,
+    const VkAccelerationStructureBuildRangeInfoKHR &range_info) {
+    if (geometry.geometryType == VK_GEOMETRY_TYPE_TRIANGLES_KHR) {
+        const VkAccelerationStructureGeometryTrianglesDataKHR &triangles = geometry.geometry.triangles;
+        const uint32_t vertex_position_size = vkuGetFormatInfo(triangles.vertexFormat).texel_block_size;
```
spencer-lunarg, line 3677: 1. It is the logic
2. See comment in function, basically vertex formats have restrictions and we can make assumptions

```suggestion
        const uint32_t vertex_position_size = GetVertexInputFormatSize(triangles.vertexFormat);
```

scripts/generators/sync_validation_generator.py
```diff
@@ -490,22 +490,11 @@ def createStageAccessCombinations(self):
...
 
-        # ACCELERATION_STRUCTURE_BUILD accesses input geometry buffers using SHADER_READ access.
-        # The core of validation logic works with expanded stages and SHADER_STORAGE_READ was
-        # choosen to represent SHADER_READ access. Then SHADER_READ itself is handled correctly:
-        # validation logic automatically includes SHADER_READ when its sub-access is used.
-        #
-        # This new stage-access pair should be added manually because vk.xml does not define
-        # ACCELERATION_STRUCTURE_BUILD for SHADER_READ's sub-accesses.
-        #
-        # TODO: with this change SHADER_STORAGE_READ also passes validation (in addition to SHADER_READ),
```
spencer-lunarg, line 501: good 2024 Artem!

spencer-lunarg (approved):
LGTM (with one VUL nit)

layers/sync/sync_validation.cpp
artem-lunarg: "Vertex input" this sounds it more related to vertex input stages of fixed-fucntion graphic pipeline. These stages are not used in ray tracing, so it might be a good idea not to re-use this. For example, in graphics pipeline there are rules about conversion rules of vertex attributes to the types defined by the shaders (just an example of one difference) and there is no such thing for ray tracing - it's a regular buffers that do not interract with fixed-function functionality.

I did not check the spec. If it turns out formats of the vertex data that can be used by the acceleration structure build process are related to graphics pipeline vertex input stage then it makes sense to re-use format related helpers. For now leave as it is, should be easy to update if needed.

PR #9512, opened 2025-02-24
layer: Cleanup logging code

spencer-lunarg:
This change is just moving things around for future change, comments on code

### DIFF
--- a/layers/error_message/logging.cpp
+++ b/layers/error_message/logging.cpp
@@ -111,18 +111,25 @@ bool DebugReport::UpdateLogMsgCounts(int32_t vuid_hash) const {
     }
 }
 
-bool DebugReport::DebugLogMsg(VkFlags msg_flags, const LogObjectList &objects, const char *msg, const char *text_vuid) const {
+bool DebugReport::LogMessage(VkFlags msg_flags, const LogObjectList &objects, const Location &loc, std::string_view vuid_text,
+                             const std::string &main_message) {
+    assert(*(vuid_text.data() + vuid_text.size()) == '\0');
+
+    VkDebugUtilsMessageSeverityFlagsEXT msg_severity;
+    VkDebugUtilsMessageTypeFlagsEXT msg_type;
+
+    // Convert the info to the VK_EXT_debug_utils format
+    DebugReportFlagsToAnnotFlags(msg_flags, &msg_severity, &msg_type);
+    std::unique_lock<std::mutex> lock(debug_output_mutex);
+    // Avoid logging cost if msg is to be ignored
+    const uint32_t vuid_hash = (vuid_text.empty()) ? 0 : hash_util::VuidHash(vuid_text);
+    if (!LogMsgEnabled(vuid_hash, msg_severity, msg_type)) {
+        return false;
+    }
+
     bool bail = false;
     std::vector<VkDebugUtilsLabelEXT> queue_labels;
     std::vector<VkDebugUtilsLabelEXT> cmd_buf_labels;
-
-    // Convert the info to the VK_EXT_debug_utils format
-    VkDebugUtilsMessageTypeFlagsEXT msg_type;
-    VkDebugUtilsMessageSeverityFlagsEXT msg_severity;
-    DebugReportFlagsToAnnotFlags(msg_flags, &msg_severity, &msg_type);
-    if (!(active_msg_severities & msg_severity) || !(active_msg_types & msg_type)) {
-        return false;  // quick check again to make sure user wants these printed
-    }
 
     std::vector<std::string> object_labels;
     // Ensures that push_back will not reallocate, thereby providing pointer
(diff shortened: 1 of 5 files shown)


layers/error_message/logging.cpp
```diff
@@ -111,19 +111,26 @@ bool DebugReport::UpdateLogMsgCounts(int32_t vuid_hash) const {
     }
 }
 
-bool DebugReport::DebugLogMsg(VkFlags msg_flags, const LogObjectList &objects, const char *msg, const char *text_vuid) const {
-    bool bail = false;
-    std::vector<VkDebugUtilsLabelEXT> queue_labels;
-    std::vector<VkDebugUtilsLabelEXT> cmd_buf_labels;
+bool DebugReport::LogMessage(VkFlags msg_flags, const LogObjectList &objects, const Location &loc, std::string_view vuid_text,
```
spencer-lunarg, line 114: this is the new "log" function, no `LogMsg`/`DebugLogMsg` combo, just a single thing to rule them all

layers/error_message/logging.cpp
```diff
@@ -111,19 +111,26 @@ bool DebugReport::UpdateLogMsgCounts(int32_t vuid_hash) const {
...
+    assert(*(vuid_text.data() + vuid_text.size()) == '\0');
 
-    // Convert the info to the VK_EXT_debug_utils format
-    VkDebugUtilsMessageTypeFlagsEXT msg_type;
     VkDebugUtilsMessageSeverityFlagsEXT msg_severity;
+    VkDebugUtilsMessageTypeFlagsEXT msg_type;
+
+    // Convert the info to the VK_EXT_debug_utils format
     DebugReportFlagsToAnnotFlags(msg_flags, &msg_severity, &msg_type);
-    if (!(active_msg_severities & msg_severity) || !(active_msg_types & msg_type)) {
```
spencer-lunarg, line 123: we were doing this check twice because the logic was half split in the two old functions

layers/error_message/logging.cpp
```diff
@@ -187,10 +192,61 @@ bool DebugReport::DebugLogMsg(VkFlags msg_flags, const LogObjectList &objects, c
     callback_data.objectCount = static_cast<uint32_t>(object_name_infos.size());
     callback_data.pObjects = object_name_infos.data();
 
+    std::string full_message = CreateMessageText(msg_flags, loc, object_name_infos, vuid_hash, vuid_text, main_message);
```
spencer-lunarg, line 195: following PR will have a `CreateMessageJson`!

layers/error_message/logging.cpp
```diff
@@ -572,93 +630,26 @@ VKAPI_ATTR void DeactivateInstanceDebugCallbacks(DebugReport *debug_report) {
...
-                full_message.append(spec_url_section);
-                full_message.append("#");  // CMake hates hashes
-                full_message.append(vuid_text);
-                full_message.append(")");
-            }
-        }
-    }
-
-    return DebugLogMsg(msg_flags, objects, full_message.c_str(), vuid_text.data());
+bool DebugReport::LogMessageVaList(VkFlags msg_flags, const LogObjectList &objects, const Location &loc, std::string_view vuid_text,
```
spencer-lunarg, line 649: all this does is call the `text::VFormat`

I want to create a `LogError` after then that will take in `std::string` or a string stream, basically this is setting up note being stuck to c-style printf

arno-lunarg: Just only use an allocating string representation if absolutely needed! We tend to overuse them when sometimes a plain `const char*` or a `std::string_view` is enough

layers/error_message/logging.cpp
```diff
@@ -111,19 +111,26 @@ bool DebugReport::UpdateLogMsgCounts(int32_t vuid_hash) const {
     }
 }
 
-bool DebugReport::DebugLogMsg(VkFlags msg_flags, const LogObjectList &objects, const char *msg, const char *text_vuid) const {
-    bool bail = false;
-    std::vector<VkDebugUtilsLabelEXT> queue_labels;
-    std::vector<VkDebugUtilsLabelEXT> cmd_buf_labels;
+bool DebugReport::LogMessage(VkFlags msg_flags, const LogObjectList &objects, const Location &loc, std::string_view vuid_text,
+                             const std::string &main_message) {
+    assert(*(vuid_text.data() + vuid_text.size()) == '\0');
```
artem-lunarg, line 116: Is there option not to do this? If function works with any std::string_view no need to worry about issues, and it should be easier to write code for any std::string_view than for a C-compatible version of string representation.

artem-lunarg (approved):
Finally this code gets update. One remark about std::string_view.

layers/error_message/logging.cpp
spencer-lunarg: honestly I just copy and pasted this around, I agree this is likely not needed. Will have a follow-up cleanup PR and remove this

layers/error_message/logging.cpp
```diff
@@ -250,53 +306,55 @@ bool DebugReport::DebugLogMsg(VkFlags msg_flags, const LogObjectList &objects, c
...
-#ifdef VK_USE_PLATFORM_ANDROID_KHR
-    if (force_default_log_callback) {
-        use_default_callbacks = true;
-    }
+        // Construct and append the specification text and link to the appropriate version of the spec
+        if (nullptr != spec_text) {
+#ifdef ANNOTATED_SPEC_LINK
+            std::string spec_url_base = ANNOTATED_SPEC_LINK;
+#else
+            std::string spec_url_base = "https://docs.vulkan.org/spec/latest/";
```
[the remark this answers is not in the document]
spencer-lunarg: agree (and your other overall comment about use using `std::string` too much

This PR is just about re-arranging code, going to merge, then apply the various fixes as different independent PRs (I can fix the std::string abuse and add JSON support in 2 PR after this that won't conflict)

PR #9515, opened 2025-02-24
layers: Small followup from 9512

spencer-lunarg:
small fix to address @artem-lunarg @arno-lunarg nit in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9512

### DIFF
--- a/layers/error_message/logging.cpp
+++ b/layers/error_message/logging.cpp
@@ -111,7 +111,7 @@ bool DebugReport::UpdateLogMsgCounts(int32_t vuid_hash) const {
     }
 }
 
-bool DebugReport::LogMessage(VkFlags msg_flags, const LogObjectList &objects, const Location &loc, std::string_view vuid_text,
+bool DebugReport::LogMessage(VkFlags msg_flags, std::string_view vuid_text, const LogObjectList &objects, const Location &loc,
                              const std::string &main_message) {
     // Convert the info to the VK_EXT_debug_utils format
     VkDebugUtilsMessageSeverityFlagsEXT msg_severity;
@@ -324,7 +324,7 @@ std::string DebugReport::CreateMessageText(VkFlags msg_flags, const Location &lo
         }
 
         // Construct and append the specification text and link to the appropriate version of the spec
-        if (nullptr != spec_text) {
+        if (spec_text && spec_url_section) {
 #ifdef ANNOTATED_SPEC_LINK
             const char *spec_url_base = ANNOTATED_SPEC_LINK;
 #else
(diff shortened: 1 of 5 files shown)


layers/error_message/logging.cpp
```diff
@@ -111,23 +111,21 @@ bool DebugReport::UpdateLogMsgCounts(int32_t vuid_hash) const {
     }
 }
 
-bool DebugReport::LogMessage(VkFlags msg_flags, const LogObjectList &objects, const Location &loc, std::string_view vuid_text,
+bool DebugReport::LogMessage(VkFlags msg_flags, std::string_view vuid_text, const LogObjectList &objects, const Location &loc,
```
spencer-lunarg, line 114: wanted to match the way we do `LogError(vuid, objects, loc)` to just be the same order here (for consistency)

layers/error_message/logging.cpp
```diff
@@ -113,21 +113,19 @@ bool DebugReport::UpdateLogMsgCounts(int32_t vuid_hash) const {
...
     VkDebugUtilsMessageTypeFlagsEXT msg_type;
-
-    // Convert the info to the VK_EXT_debug_utils format
     DebugReportFlagsToAnnotFlags(msg_flags, &msg_severity, &msg_type);
+
     std::unique_lock<std::mutex> lock(debug_output_mutex);
+
     // Avoid logging cost if msg is to be ignored
-    const uint32_t vuid_hash = (vuid_text.empty()) ? 0 : hash_util::VuidHash(vuid_text);
+    const uint32_t vuid_hash = hash_util::VuidHash(vuid_text);
```
artem-lunarg, line 124: If we can have empty vuid for some reason (probably not?), this will likely result in a different hash value than before (0). Just to document theoretical change of behavior.

layers/error_message/logging.cpp
```diff
@@ -315,7 +314,7 @@ std::string DebugReport::CreateMessageText(VkFlags msg_flags, const Location &lo
         uint32_t num_vuids = sizeof(vuid_spec_text) / sizeof(vuid_spec_text_pair);
         const char *spec_text = nullptr;
         // Only the Antora site will make use of the sections
-        std::string spec_url_section;
+        const char *spec_url_section = nullptr;
```
arno-lunarg, line 317:
```suggestion
        const char *spec_url_section = "";
```
This will not get assigned for UNASSIGNED vuids for instance, so need to prevent nullptr dereference

spencer-lunarg: it will never make it here, we wrap this with `if ((vuid_text.find("VUID-") != std::string::npos)) {`

spencer-lunarg: but in the case we can't find the VUID, agree this is bad to be null, I will keep as `nullptr` and do a check to skip the whole "construct the spec link" code if `spec_url_section` is null

spencer-lunarg: now do a `if (spec_text && spec_url_section) {` check

arno-lunarg (approved):

PR #9516, opened 2025-02-24
vvl: Rework accel structs mem aliasing validation, re-vengeance 

arno-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9428

Turns out what I observed in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9492 WRT buffers not sharing the same base address when bound to the same memory is in practice only true for for sparse buffers, and the reworked validation in this PR does not attempt to validate sparse buffers anymore, since it is something you can only safely do a `vkQueueSubmit` time when all memory bindings can be resolved for sparse buffers/images. The code to handle the specific sparse case has never been there, we used to validate sparse buffers with the "normal" path, a mistake.

### DIFF
--- a/layers/core_checks/cc_ray_tracing.cpp
+++ b/layers/core_checks/cc_ray_tracing.cpp
@@ -28,6 +28,8 @@
 #include "state_tracker/ray_tracing_state.h"
 #include "error_message/error_strings.h"
 
+#include <algorithm>
+
 bool CoreChecks::ValidateInsertAccelerationStructureMemoryRange(VkAccelerationStructureNV as, const vvl::DeviceMemory &mem_info,
                                                                 VkDeviceSize mem_offset, const Location &loc) const {
     return ValidateInsertMemoryRange(VulkanTypedHandle(as, kVulkanObjectTypeAccelerationStructureNV), mem_info, mem_offset, loc);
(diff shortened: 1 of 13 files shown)


layers/state_tracker/ray_tracing_state.h
```diff
@@ -75,12 +75,27 @@ class AccelerationStructureNV : public Bindable {
...
+          buffer_state(buf_state) {
+        // If the buffer's device address has not been queried,
+        // get it here. Since it is used for the purpose of
+        // validation, do not try to update buffer_state, since
+        // it only tracks application state.
+        if (buffer_state && buffer_state->deviceAddress == 0) {
+            VkBufferDeviceAddressInfo address_info = vku::InitStructHelper();
+            address_info.buffer = buffer_state->VkHandle();
+
+            if (api_version >= VK_API_VERSION_1_2) {
```
spencer-lunarg, line 92: we have `GetBufferDeviceAddressHelper` for this, I would almost expect `Device::CreateAccelerationStructureState` to be in charge of calling this and passing in the `buffer_device_address` (then it can also just `const`)

arno-lunarg: This function is stuck in GPU-AV, the plumbing to share code is not really worth it. Should we have this a third time, I would

arno-lunarg: But I need if the Vulkan API version is below 1.2, I need to make sure the device address extension has been added before trying to do this stuff, otherwise 💥

arno-lunarg: okkkk in the end you were right, I just moved the GPU-AV function, it is simpler!

layers/state_tracker/ray_tracing_state.h
```diff
@@ -120,11 +121,30 @@ class AccelerationStructureKHR : public StateObject {
...
+        return {buffer_device_address + safe_create_info.offset,
+                buffer_device_address + safe_create_info.offset + safe_create_info.size};
+    }
+
     const vku::safe_VkAccelerationStructureCreateInfoKHR safe_create_info;
     const VkAccelerationStructureCreateInfoKHR &create_info;
 
     uint64_t opaque_handle = 0;
     std::shared_ptr<vvl::Buffer> buffer_state{};
+    VkDeviceAddress buffer_device_address =
```
spencer-lunarg, line 146: nit: could be `const  VkDeviceAddress buffer_device_address;` now right?

side nit: move comment to above for formatting sake

layers/core_checks/core_validation.h
```diff
@@ -1689,9 +1689,9 @@ class CoreChecks : public vvl::Device {
     bool ValidateAccelerationStructuresMemoryAlisasing(const LogObjectList& objlist, uint32_t infoCount,
                                                        const VkAccelerationStructureBuildGeometryInfoKHR* pInfos, uint32_t info_i,
                                                        const ErrorObject& error_obj) const;
-    bool ValidateAccelerationStructuresDeviceScratchBufferMemoryAlisasing(
-        const LogObjectList& objlist, uint32_t info_count, const VkAccelerationStructureBuildGeometryInfoKHR* p_infos,
-        uint32_t info_i, const VkAccelerationStructureBuildRangeInfoKHR* const* pp_range_infos, const ErrorObject& error_obj) const;
+    bool ValidateAccelerationStructuresDeviceScratchBufferMemoryAliasing(
+        VkCommandBuffer cmd_buffer, uint32_t info_count, const VkAccelerationStructureBuildGeometryInfoKHR* p_infos,
+        const VkAccelerationStructureBuildRangeInfoKHR* const* pp_range_infos, const ErrorObject& error_obj) const;
```
spencer-lunarg, line 1694: realize you didn't do it, but while here, can we pass in `const Location& loc` instead of `ErrorObject&`

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -215,67 +217,234 @@ bool CoreChecks::ValidateAccelerationStructuresMemoryAlisasing(const LogObjectLi
...
 
-    // Loop on other acceleration structure builds info.
-    // Given that comparisons are commutative, only need to consider elements after info_i
-    assert(info_count > info_i);
-    for (uint32_t other_info_j = info_i + 1; other_info_j < info_count; ++other_info_j) {
-        // Validate that scratch buffer's memory does not overlap destination acceleration structure's memory, or source
-        // acceleration structure's memory if build mode is update, or other scratch buffers' memory.
-        // Here validation is pessimistic: if one buffer associated to pInfos[other_info_j].scratchData.deviceAddress has an
-        // overlap, an error will be logged.
+    for (const auto [info_i, info] : vvl::enumerate(p_infos, info_count)) {
```
spencer-lunarg, line 264: assuming the `info_count` here aren't too large (may be wrong) but now we are looping this twice in `PreCallValidateCmdBuildAccelerationStructuresKHR`

not sure if there is a way to to keep this logic in the other loop we have before the function is called

Agree this keeps the logic more encapsulated, so happy to just leave until we find it is actually a perf issue

tests/unit/ray_tracing.cpp
```diff
@@ -1910,12 +1916,13 @@ TEST_F(NegativeRayTracing, AccelerationStructuresOverlappingMemory) {
...
-            m_errorMonitor->SetDesiredError("VUID-vkCmdBuildAccelerationStructuresKHR-dstAccelerationStructure-03701");
-            m_errorMonitor->SetDesiredError("VUID-vkCmdBuildAccelerationStructuresKHR-dstAccelerationStructure-03702");
-        }
+        // => due to validation code optimisations, not *all* overlaps described by 03701 and 03702 will be detected,
+        // but if there is *at least one*, it will *always+ be detected.
+        m_errorMonitor->SetDesiredError("VUID-vkCmdBuildAccelerationStructuresKHR-dstAccelerationStructure-03701", 2);
+        m_errorMonitor->SetDesiredError("VUID-vkCmdBuildAccelerationStructuresKHR-dstAccelerationStructure-03702", 2);
+
         for (size_t i = 0; i < build_info_count; ++i) {
             m_errorMonitor->SetDesiredError("VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03668");
```
spencer-lunarg, line 1927:
```suggestion
            m_errorMonitor->SetDesiredError("VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03668", build_info_count);
```

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -215,67 +217,234 @@ bool CoreChecks::ValidateAccelerationStructuresMemoryAlisasing(const LogObjectLi
...
+                            if (const auto other_dst_as_state = Get<vvl::AccelerationStructureKHR>(
+                                    p_infos[overlapped_address_range->info_i].dstAccelerationStructure)) {
+                                const vvl::range<VkDeviceAddress> src_as_other_dst_as_intersection =
+                                    src_as_address_range.range & overlapped_address_range->range;
+
+                                skip |= LogError(
+                                    "VUID-vkCmdBuildAccelerationStructuresKHR-dstAccelerationStructure-03701",
+                                    LogObjectList(cmd_buffer, src_as_state->VkHandle(), other_dst_as_state->VkHandle()),
+                                    info_i_loc.dot(Field::srcAccelerationStructure),
+                                    ", in update mode, overlaps with dstAccelerationStructure of pInfos[%" PRIu32
```
spencer-lunarg, line 306: this looks like

> vkCmdBuildAccelerationStructuresKHR(): pInfos[1].srcAccelerationStructure , in update mode, overlaps with dstAccelerationStructure of pInfos[0], on device address range [0x761a654a8000, 0x761a654a8090).

So maybe 

```suggestion
                                    "is in update mode, which overlaps with dstAccelerationStructure of pInfos[%" PRIu32
```

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -215,67 +217,234 @@ bool CoreChecks::ValidateAccelerationStructuresMemoryAlisasing(const LogObjectLi
...
+                    case AddressRangeOrigin::SrcAccelStruct: {
+                        if (const auto other_src_as_state = Get<vvl::AccelerationStructureKHR>(
+                                p_infos[overlapped_address_range->info_i].srcAccelerationStructure)) {
+                            const vvl::range<VkDeviceAddress> scratch_other_src_as_intersection =
+                                scratch_address_range.range & overlapped_address_range->range;
+
+                            skip |= LogError(
+                                "VUID-vkCmdBuildAccelerationStructuresKHR-scratchData-03705",
+                                LogObjectList(cmd_buffer, other_src_as_state->VkHandle()), info_i_loc.dot(Field::scratchData),
+                                ", overlaps with srcAccelerationStructure of pInfos[%" PRIu32
```
spencer-lunarg, line 413:
```suggestion
                                "overlaps with srcAccelerationStructure of pInfos[%" PRIu32
```

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -215,67 +217,234 @@ bool CoreChecks::ValidateAccelerationStructuresMemoryAlisasing(const LogObjectLi
...
+                            if (const auto other_src_as_state = Get<vvl::AccelerationStructureKHR>(
+                                    p_infos[overlapped_address_range->info_i].srcAccelerationStructure)) {
+                                const vvl::range<VkDeviceAddress> dst_as_other_src_as_intersection =
+                                    dst_as_address_range.range & overlapped_address_range->range;
+
+                                skip |= LogError(
+                                    "VUID-vkCmdBuildAccelerationStructuresKHR-dstAccelerationStructure-03701",
+                                    LogObjectList(cmd_buffer, dst_as_state->VkHandle(), other_src_as_state->VkHandle()),
+                                    info_i_loc.dot(Field::dstAccelerationStructure),
+                                    ", overlaps with srcAccelerationStructure of pInfos[%" PRIu32
```
spencer-lunarg, line 348:
```suggestion
                                    "overlaps with srcAccelerationStructure of pInfos[%" PRIu32
```

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -215,67 +217,234 @@ bool CoreChecks::ValidateAccelerationStructuresMemoryAlisasing(const LogObjectLi
...
+    const auto insert_address = [](std::vector<AddressRange> &address_ranges, const AddressRange address_range) {
+        std::optional<AddressRange> overlapped_range = std::nullopt;
+        auto insert_pos = std::lower_bound(
+            address_ranges.begin(), address_ranges.end(), address_range,
+            [&overlapped_range](const AddressRange iter, const AddressRange new_elt) -> bool {
+                const vvl::range<VkDeviceAddress> intersection = iter.range & new_elt.range;
+                if (intersection.non_empty()) {
+                    const bool both_src_as_ranges =
+                        iter.origin == AddressRangeOrigin::SrcAccelStruct && new_elt.origin == AddressRangeOrigin::SrcAccelStruct;
+                    // #ARNO_TODO make sure VUID 03668 is correctly handled
```
spencer-lunarg, line 245: correctly handled as in what?

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -215,67 +217,234 @@ bool CoreChecks::ValidateAccelerationStructuresMemoryAlisasing(const LogObjectLi
...
+                                src_as_address_range.range & overlapped_address_range->range;
+                            skip |= LogError("VUID-vkCmdBuildAccelerationStructuresKHR-scratchData-03705",
+                                             LogObjectList(cmd_buffer, src_as_state->VkHandle()),
+                                             info_i_loc.dot(Field::srcAccelerationStructure),
+                                             "overlaps with scratch buffer of pInfos[%" PRIu32 "] on device address range %s.",
+                                             overlapped_address_range->info_i,
+                                             string_range_hex(src_as_other_scratch_intersection).c_str());
+                            break;
+                        }
+                        default:
```
spencer-lunarg, line 323: super nit: Compiler can make sure no one tries to add a 5th enum value

```suggestion
                        case AddressRangeOrigin::Undefined:
```

layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -215,67 +217,234 @@ bool CoreChecks::ValidateAccelerationStructuresMemoryAlisasing(const LogObjectLi
...
-                other_info_in_update_mode ? other_src_as_state.get() : nullptr,
-                other_info_j_loc.dot(Field::srcAccelerationStructure), *other_dst_as_state,
-                other_info_j_loc.dot(Field::dstAccelerationStructure), other_info_scratches, other_info.scratchData.deviceAddress,
-                assumed_other_scratch_size, &other_scratch_address_loc);
+        // Do not attempt looking for memory overlaps here if any buffer associated to scratch address is sparse
+        const auto scratch_buffers = GetBuffersByAddress(info.scratchData.deviceAddress);
+        const bool any_scratch_sparse = std::any_of(scratch_buffers.begin(), scratch_buffers.end(),
+                                                    [](const vvl::Buffer *buffer) { return buffer && buffer->sparse; });
+        if (!any_scratch_sparse) {
+            assert(error_obj.location.function != Func::vkBuildAccelerationStructuresKHR);
```
spencer-lunarg, line 393: either remove or put at top of function

spencer-lunarg (approved):
with comments addressed

layers/state_tracker/ray_tracing_state.h
arno-lunarg: yes, I always forget that const members are a thing

layers/core_checks/cc_ray_tracing.cpp
arno-lunarg: the `which` makes you think that what is overlapping is "update mode"
But I can go "is in update mode, **but** overlaps with ..."

layers/core_checks/cc_ray_tracing.cpp
arno-lunarg: As in "I forgot to remove that comment" x)

layers/core_checks/cc_ray_tracing.cpp
arno-lunarg: yes better!

layers/core_checks/cc_ray_tracing.cpp
arno-lunarg: Moved it to top

layers/core_checks/cc_ray_tracing.cpp
arno-lunarg: Ah yes good point I overlooked that... I need to call `ValidateAccelerationStructuresDeviceScratchBufferMemoryAliasing` at other spots, it is likely to change a bit, I will also look at that issue

Issue #9499, opened 2025-02-21
GPU-AV debugprintf fails when DEVICE_LOCAL | HOST_COHERENT | HOST_VISIBLE heap is full


danginsburg:
**Environment:**
 - OS: Windows
 - GPU and driver version: GTX 5080 571.32
 - SDK or header version if building from repo: ff2042e742e676f1c12054695df6e5c0167124f8
 - Options enabled (synchronization, best practices, etc.): VK_VALIDATION_FEATURE_ENABLE_DEBUG_PRINTF_EXT

**Describe the Issue**

When I enable VK_VALIDATION_FEATURE_ENABLE_DEBUG_PRINTF_EXT in our engine I get the following error from validation and then debugprintf does not work:

```
VK ERROR : VALIDATION - Message Id Number: -1841738615 | Message Id Name: UNASSIGNED-DEBUG-PRINTF
    Validation Error: [ UNASSIGNED-DEBUG-PRINTF ] | MessageID = 0x92394c89 | vkAllocateCommandBuffers(): Internal VMA Error, DebugPrintf is being disabled. Details:
Unable to allocate device memory for internal buffer. VMA statistics = {
  "General": {
    "API": "Vulkan", 
    "apiVersion": "1.4.303", 
    "GPU": "NVIDIA Graphics Device", 
    "deviceType": 2, 
    "maxMemoryAllocationCount": 4294967295, 
    "bufferImageGranularity": 1024, 
    "nonCoherentAtomSize": 64, 
    "memoryHeapCount": 3, 
    "memoryTypeCount": 5
  }, 
  "Total": {
    "BlockCount": 210, 
    "BlockBytes": 208569936, 
    "AllocationCount": 910, 
    "AllocationBytes": 135208048, 
    "UnusedRangeCount": 92, 
    "AllocationSizeMin": 32768, 
    "AllocationSizeMax": 1048576, 
    "UnusedRangeSizeMin": 32768, 
    "UnusedRangeSizeMax": 16024432
  }, 
  "MemoryInfo": {
    "Heap 0": {
      "Flags": ["DEVICE_LOCAL"], 
      "Size": 16660824064, 
      "Budget": {
        "BudgetBytes": 13328659251, 
        "UsageBytes": 0
      }, 
      "Stats": {
        "BlockCount": 0, 
        "BlockBytes": 0, 
        "AllocationCount": 0, 
        "AllocationBytes": 0, 
[...]

PR #9518, opened 2025-02-24
gpuav: Remove VMA pool

arno-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9499

Went down the "Do not use VMA pools" route.
Still need to investigate our VMA usage further. 
Ideally our buffers should always use device local memory, otherwise perf can get stupid.
Buffers need to be cached rather than created/destroyed at will, it can happen a lot during a frame, so much so that in games like Doom Eternal it can become the bottleneck


spencer-lunarg:
before we remove, could we try to add a test that uses VMA (or maybe not even that), allocates all the memory, and would normally fail without this check

spencer-lunarg (approved):
LGTM

PR #9520, opened 2025-02-24
gpu: Remove some inherited state objects

jeremyg-lunarg:
Remove inheritance from state objects used in descriptor validation.


layers/state_tracker/descriptor_sets.h
```diff
@@ -997,6 +1038,12 @@ class DescriptorSet : public StateObject {
 
     bool ValidateBindingOnGPU(const DescriptorBinding &binding, bool is_runtime_descriptor_array) const;
 
+    void SetSubState(LayerObjectTypeId id, std::unique_ptr<DescriptorSetProxy> &&substate) {
```
jeremyg-lunarg: This is what I'm thinking to do for state objects that have large amounts of inherited data, such as VkCommandBuffer, VkQueue and possibly VkImage. The code for managing SubStates can probably be pulled out and reused somehow.

layers/gpuav/core/gpuav.h
```diff
@@ -28,21 +28,13 @@ struct ShaderObject;
 }  // namespace chassis
 
 namespace gpuav {
-class Buffer;
-class BufferView;
```
jeremyg-lunarg, line 32: most of these state objects just added a single field to hold the descriptor id,

layers/gpuav/descriptor_validation/gpuav_descriptor_set.cpp
```diff
@@ -63,95 +62,98 @@ void DescriptorSet::BuildBindingLayouts() {
...
 
-static glsl::DescriptorState GetInData(const vvl::BufferDescriptor &desc) {
-    auto buffer_state = static_cast<const Buffer *>(desc.GetBufferState());
+static glsl::DescriptorState GetInData(DescriptorHeap &heap, const vvl::BufferDescriptor &desc) {
+    auto buffer_state = desc.GetBufferState();
     if (!buffer_state) {
         return glsl::DescriptorState(DescriptorClass::GeneralBuffer, glsl::kNullDescriptor, vvl::kU32Max);
     }
-    return glsl::DescriptorState(DescriptorClass::GeneralBuffer, buffer_state->id, static_cast<uint32_t>(desc.GetEffectiveRange()));
+    return glsl::DescriptorState(DescriptorClass::GeneralBuffer, heap.GetId(buffer_state->Handle()),
```
jeremyg-lunarg, line 70: instead the descriptor id gets lazily allocated when each object is first used.

PR #9521, opened 2025-02-24
layers: Use LogMessage with std::string where possible

spencer-lunarg:
Now that `LogMessage` is consuming a `const std::string&` we had a few spots where we going from `std::string` to `const char` back to `std::string`

### DIFF
--- a/layers/gpuav/spirv/debug_printf_pass.cpp
+++ b/layers/gpuav/spirv/debug_printf_pass.cpp
@@ -624,7 +624,7 @@ bool DebugPrintfPass::Validate(const Function& current_function) {
         if (modifier == ' ') {
             std::string err_msg =
                 "OpString \"" + print_op_string() + "\" contains a isolated % which is missing the modifier (to escape use %%)";
-            module_.InternalError(tag, err_msg.c_str());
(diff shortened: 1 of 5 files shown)


layers/gpuav/spirv/debug_printf_pass.cpp
```diff
@@ -624,7 +624,7 @@ bool DebugPrintfPass::Validate(const Function& current_function) {
         if (modifier == ' ') {
             std::string err_msg =
                 "OpString \"" + print_op_string() + "\" contains a isolated % which is missing the modifier (to escape use %%)";
-            module_.InternalError(tag, err_msg.c_str());
```
spencer-lunarg, line 627: When we create a new/better `LogError` flow, we can apply it to these few edge cases that by-pass the `LogError` and call `LogMessage` directly

jeremyg-lunarg (approved):

spencer-lunarg:
> CI Vulkan-ValidationLayers build # 19186 failed.

unrelated `PositiveSyncValTimelineSemaphore.KhronosTimelineSemaphoreExample` failure

PR #9529, opened 2025-02-25
printf: Extra testing and error message fix

spencer-lunarg:
1. Add some basic Mesh/Task tests for Printf/GPU-AV
2. Fix how we report error for `printf` when user messes up our `%v3f` style syntax

### DIFF
--- a/layers/gpuav/spirv/debug_printf_pass.cpp
+++ b/layers/gpuav/spirv/debug_printf_pass.cpp
@@ -19,6 +19,7 @@
 #include <spirv/unified1/NonSemanticDebugPrintf.h>
 #include <cstring>
 #include <iostream>
+#include <string>
 
 #include "generated/device_features.h"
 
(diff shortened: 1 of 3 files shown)


layers/gpuav/spirv/debug_printf_pass.cpp
```diff
@@ -698,8 +699,14 @@ bool DebugPrintfPass::Validate(const Function& current_function) {
...
                 }
                 default:
-                    std::string err_msg = "OpString \"" + print_op_string() + "\" contains a \"" + modifier +
+                    // Need to escape, other error makes no sense
+                    std::string err_modifier = (modifier == '\n') ? "\\n" : (modifier == '\t') ? "\\t" : std::to_string(modifier);
+                    std::string err_msg = "OpString \"" + print_op_string() + "\" contains a \"" + err_modifier +
                                           "\" modifier which is an unknown modifier.";
+                    if (param_info.vector_size > 0) {
+                        // Help explain our custom vector format rules
+                        err_msg += " (for vectors you need something like %v3f which is just %f with a special v3 prefix)";
```
arno-lunarg, line 708: Probably needs to be mentioned in the doc instead of here, with all the details

arno-lunarg (approved):

layers/gpuav/spirv/debug_printf_pass.cpp
spencer-lunarg: we do https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/docs/debug_printf.md#debug-printf-format-string

just people are not always going to read it, and having a hint here when they hit it is helpful

Issue #9222, opened 2025-01-15
GPU-AV: `Post Process Descriptor Indexing` is broken in Vulkan SDK 1.4


corporateshark:
**Environment:**
 - OS: Windows 10 64-bit
 - GPU and driver version: RTX 3070 Ti, 566.36
 - SDK or header version if building from repo: 1.4.304.0
 - Options enabled (synchronization, best practices, etc.): GPU-Assisted - Shader Instrumentation - Post Process Descriptor Indexing

**Describe the Issue**

Run [3D-Graphics-Rendering-Cookbook-Second-Edition/Chapter11/06_FinalDemo](https://github.com/PacktPublishing/3D-Graphics-Rendering-Cookbook-Second-Edition/tree/main/Chapter11/06_FinalDemo).

Device Lost or (sometimes) extremely slow rendering.

**Expected behavior**

Disable `Post Process Descriptor Indexing`.

![Image](https://github.com/user-attachments/assets/d6fec27c-16ff-4da6-9bcc-fa05a8cbc9ad)

**Valid Usage ID**

No validation messages.

corporateshark:
@spencer-lunarg If you plan to reproduce this with the latest version, please note that this validation check is now deliberately [disabled in LVK](https://github.com/corporateshark/lightweightvk/blob/fb2f0f519987c3e148b75d7594d4f7e91e285dbc/lvk/vulkan/VulkanClasses.cpp#L5861) as a workaround.

spencer-lunarg:
@corporateshark thanks (for reporting this providing samples for us to use)

I have found some other issues in traces that might be related

corporateshark:
Just checked the new Vulkan SDK `1.4.304.1`. Still negative function. Device lost.

spencer-lunarg:
yes, the `1.4.304.1` SDK only had a 2 commit cherry-picked (we don't try adding to much drastic changes in a minor release), one for pointer crash and one to point the URL to a better/faster page to load the VU error message

I still think this might be fixed (haven't had time to repro yet) but known to not be in the 1.4.304.1 SDK

arno-lunarg:
Notes for us: Issues seems to come from device address validation, it completely stalls some shader(s).
Related SPIRV, before/after instrumentation:

[issue_9222_device_addresses_validation_spv_dumps.zip](https://github.com/user-attachments/files/18944194/issue_9222_device_addresses_validation_spv_dumps.zip)

[...]

PR #9532, opened 2025-02-25
gpuav: Support all atomics for BDA check

spencer-lunarg:
part of https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9222

We were handling `atomicAdd(value, bda_pointer);` but not `atomicAdd(bda_pointer, value);`

The fix (thankfully) is very simple and wrote tests to prove it works (note, tried HLSL/Slang and they generate the same code for atomics like this)


jeremyg-lunarg:
> We were handling atomicAdd(value, bda_pointer); but not atomicAdd(bda_pointer, value);

Is this line correct or did you mean a different atomic was handled? That's what it looks like from the code change.

spencer-lunarg:
@jeremyg-lunarg look at https://godbolt.org/z/P4YYPKdbd

image

So when you do an atomic (outside of store,load,exchange) you need both a `Pointer` and `Value`

So `atomicAdd(value, bda_pointer);` will have a `OpLoad %uint %_ Aligned 16` load prior to handing it (because it is only a `read` operation)

but when it is a `Pointer` the "load" is a `atomic` access (load and then store), so the access is there

jeremyg-lunarg (approved):

Issue #9527, opened 2025-02-25
GPL validation broken again.


zmike:
You *cannot* illegally read fields from `VkPipelineRenderingCreateInfo` when using GPL, and color attachment formats is definitely illegal in everything but the fragment output state.

```
17797	    }
17798	    if (!custom_init) {
17799	        // The custom iniitalization was not used, so do the regular initialization
17800	        if (in_struct->pColorAttachmentFormats) {
17801	            pColorAttachmentFormats = new VkFormat[in_struct->colorAttachmentCount];
17802	            memcpy((void*)pColorAttachmentFormats, (void*)in_struct->pColorAttachmentFormats,
17803	                   sizeof(VkFormat) * in_struct->colorAttachmentCount);
17804	        }
17805	    }
17806	}


#0  0x00007ffff7954a8d in __memmove_avx_unaligned_erms () from /lib64/libc.so.6
#1  0x00007fffea83e586 in vku::safe_VkPipelineRenderingCreateInfo::safe_VkPipelineRenderingCreateInfo (this=0x7fff8003c250, in_struct=0x7fffd15fd410, 
    copy_state=0x0, copy_pnext=false)
    at /home/zmike/src/Vulkan-ValidationLayers/external/Debug/64/Vulkan-Utility-Libraries/src/vulkan/vk_safe_struct_core.cpp:17802
#2  0x00007fffea8bc8a4 in vku::SafePnextCopy (pNext=0x7fffd15fd410, copy_state=0x0)
    at /home/zmike/src/Vulkan-ValidationLayers/external/Debug/64/Vulkan-Utility-Libraries/src/vulkan/vk_safe_struct_utils.cpp:364
#3  0x00007fffea909795 in vku::safe_VkGraphicsPipelineCreateInfo::initialize (this=0x7fff80019da8, in_struct=0x7fffd15fd160, uses_color_attachment=true, 
    uses_depthstencil_attachment=true, copy_state=0x0)
    at /home/zmike/src/Vulkan-ValidationLayers/external/Debug/64/Vulkan-Utility-Libraries/src/vulkan/vk_safe_struct_manual.cpp:707
#4  0x00007fffe9688ade in vvl::dispatch::Device::CreateGraphicsPipelines (this=0x9e8cc0, device=0xa2c700, pipelineCache=0xed362f0000000133, createInfoCount=1, 
[...]

PR #9533, opened 2025-02-25
layers: Handle yet-another ignored pointer

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9527

We ignored when `VkPipelineRenderingCreateInfo::pColorAttachmentFormats` is garbage in like 3 out of 4 cases, we missed one, this just brings the big hammer and nulls the pointer for the user if they are supplying it with garbage (we tried to not do this before and it worked well for the non-pNext structs, but this one is harder because its in the pNext)

### DIFF
--- a/layers/state_tracker/state_tracker.cpp
+++ b/layers/state_tracker/state_tracker.cpp
@@ -1746,11 +1746,19 @@ bool Device::PreCallValidateCreateGraphicsPipelines(VkDevice device, VkPipelineC
         if (pCreateInfos[i].renderPass != VK_NULL_HANDLE) {
             render_pass = Get<RenderPass>(create_info.renderPass);
         } else if (enabled_features.dynamicRendering) {
-            auto dynamic_rendering = vku::FindStructInPNextChain<VkPipelineRenderingCreateInfo>(create_info.pNext);
-            const bool rasterization_enabled = Pipeline::EnablesRasterizationStates(*this, create_info);
+            auto pipeline_rendering_ci = vku::FindStructInPNextChain<VkPipelineRenderingCreateInfo>(create_info.pNext);
             const bool has_fragment_output_state =
                 Pipeline::ContainsSubState(this, create_info, VK_GRAPHICS_PIPELINE_LIBRARY_FRAGMENT_OUTPUT_INTERFACE_BIT_EXT);
-            render_pass = std::make_shared<RenderPass>(dynamic_rendering, rasterization_enabled && has_fragment_output_state);
+            const bool rasterization_enabled =
+                has_fragment_output_state && Pipeline::EnablesRasterizationStates(*this, create_info);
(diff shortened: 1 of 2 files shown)


layers/state_tracker/state_tracker.cpp
```diff
@@ -1746,11 +1746,19 @@ bool Device::PreCallValidateCreateGraphicsPipelines(VkDevice device, VkPipelineC
...
         } else if (enabled_features.dynamicRendering) {
-            auto dynamic_rendering = vku::FindStructInPNextChain<VkPipelineRenderingCreateInfo>(create_info.pNext);
-            const bool rasterization_enabled = Pipeline::EnablesRasterizationStates(*this, create_info);
+            auto pipeline_rendering_ci = vku::FindStructInPNextChain<VkPipelineRenderingCreateInfo>(create_info.pNext);
             const bool has_fragment_output_state =
                 Pipeline::ContainsSubState(this, create_info, VK_GRAPHICS_PIPELINE_LIBRARY_FRAGMENT_OUTPUT_INTERFACE_BIT_EXT);
-            render_pass = std::make_shared<RenderPass>(dynamic_rendering, rasterization_enabled && has_fragment_output_state);
+            const bool rasterization_enabled =
+                has_fragment_output_state && Pipeline::EnablesRasterizationStates(*this, create_info);
+            if (pipeline_rendering_ci && pipeline_rendering_ci->pColorAttachmentFormats && !rasterization_enabled) {
```
artem-lunarg, line 1754: My attempt to read the issue description failed on "cannot illegally" part. Why pColorAttachmentFormats is not valid here?

artem-lunarg (approved):
LGTM if CI happy

layers/state_tracker/state_tracker.cpp
artem-lunarg: Also do we correct client structures in other parts of validation?

spencer-lunarg: > Also do we correct client structures in other parts of validation?

No, we have a handful of hacks in SafeStruct to ignore the various pointers of `VkGraphicsPipelineCreateInfo`

the issue here is this is a pNext struct, so we would need even more magic to pass this info to the automated pNext checking logic

... so just giving up and doing this one here

> Why pColorAttachmentFormats is not valid here?

If you are just doing a vertex shader GPL, there is no Dynamic Rendering stuff and this struct is just never consumed by the driver and because GPL is "fun" we (as a tool) have to not parse this, which Safe Struct is bad at doing

artem-lunarg: Thanks for the details. GPL cambrian explosion keeps giving life to new type of bugs.

spencer-lunarg: GPL will be providing bugs at least until 2030

PR #9536, opened 2025-02-26
layers: Fix warning for Storage Image

spencer-lunarg:
Fixes issue with warning, the user might use `Rgba8` in their shader which maps to `VK_FORMAT_R8G8B8A8_UNORM` but the user can still use `VK_FORMAT_B8G8R8A8_UNORM` (basically was doing exact format check instead of using `vkuFormatCompatibilityClass`

### DIFF
--- a/layers/drawdispatch/descriptor_validator.cpp
+++ b/layers/drawdispatch/descriptor_validator.cpp
@@ -17,7 +17,9 @@
 
 #include "descriptor_validator.h"
 #include <vulkan/vulkan_core.h>
+#include <sstream>
 #include "generated/spirv_grammar_helper.h"
+#include "generated/spirv_validation_helper.h"
 #include "state_tracker/shader_stage_state.h"
 #include "error_message/error_strings.h"
 #include "state_tracker/descriptor_sets.h"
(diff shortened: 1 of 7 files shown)


layers/drawdispatch/descriptor_validator.cpp
```diff
@@ -387,15 +388,26 @@ bool DescriptorValidator::ValidateDescriptor(const spirv::ResourceInterfaceVaria
...
-                DescribeDescriptor(resource_variable, index, descriptor_type).c_str(),
-                string_VkFormat(resource_variable.info.image_format), dev_state.FormatHandle(image_view).c_str(),
-                string_VkFormat(image_view_ci.format));
+            std::stringstream msg;
+            msg << "the " << DescribeDescriptor(resource_variable, index, descriptor_type)
+                << " is accessed by a OpTypeImage that has a Format operand equivalent to "
+                << string_VkFormat(resource_variable.info.image_format) << " which doesn't match the "
+                << dev_state.FormatHandle(image_view) << " format (" << string_VkFormat(image_view_ci.format)
+                << "). Any loads or stores with the variable will produce undefined values.";
+            if (vkuFormatCompatibilityClass(image_view_ci.format) ==
```
spencer-lunarg, line 400: This is 100% going to trip people up as a "wait I was not allowed to do this?" so the new message tries to help explain more "what to do" here

>vkCmdDispatch(): the storage image descriptor [VkDescriptorSet 0xee647e0000000009, Set 0, Binding 0, Index 0, variable "si0"] is accessed by a OpTypeImage that has a Format operand equivalent to VK_FORMAT_R8G8B8A8_UNORM which doesn't match the VkImageView 0xf443490000000006 format (VK_FORMAT_B8G8R8A8_UNORM). Any loads or stores with the variable will produce undefined values. While the formats are compatible, Storage Images must exactly match. Two ways to resolve this are
>
> 1. Set your ImageView to VK_FORMAT_R8G8B8A8_UNORM and swizzle the values in the shader to match the desired results.
>
> 2. Use the Unknown format in your shader (will need the widely supported shaderStorageImageWriteWithoutFormat feature)
>
> Spec information at https://docs.vulkan.org/spec/latest/chapters/textures.html#textures-format-validation

artem-lunarg (approved):

PR #9538, opened 2025-02-26
tests: Fix RT tests sporadic failures

arno-lunarg:


tests/unit/ray_tracing.cpp
```diff
@@ -1870,7 +1870,7 @@ TEST_F(NegativeRayTracing, AccelerationStructuresOverlappingMemory) {
         // of elements in `build_infos`
         // => due to validation code optimisations, not *all* overlaps will be detected,
         // but if there is *at least one*, it will *always+ be detected.
-        m_errorMonitor->SetDesiredError("VUID-vkCmdBuildAccelerationStructuresKHR-dstAccelerationStructure-03702", 2);
+        m_errorMonitor->SetAllowedFailureMsg("VUID-vkCmdBuildAccelerationStructuresKHR-dstAccelerationStructure-03702");
```
spencer-lunarg, line 1873: What is the point of this test if everything is `SetAllowedFailureMsg`

It basically means "sometimes this tests passes, sometimes it falis"

arno-lunarg: The emitted errors are un-predictable, it is just the nature of validation: it will depend on the addresses the driver returns, and the order in which memory overlaps are found

spencer-lunarg (approved):

PR #9540, opened 2025-02-26
sync: Update renderpass layout transition messages

artem-lunarg:
> vkCmdBeginRenderPass(): WRITE_AFTER_READ hazard detected. vkCmdBeginRenderPass performs image layout transition in subpass 0 on attachment 0 (VkImageView 0xcfef35000000000a, VkImage 0xf56c9b0000000004, oldLayout VK_IMAGE_LAYOUT_UNDEFINED, newLayout VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL), which was previously read by vkCmdCopyImage.
No sufficient synchronization is present to ensure that a layout transition does not conflict with a prior read (VK_ACCESS_2_TRANSFER_READ_BIT) at VK_PIPELINE_STAGE_2_COPY_BIT. An execution dependency is sufficient to prevent this hazard.

### DIFF
--- a/layers/sync/sync_renderpass.cpp
+++ b/layers/sync/sync_renderpass.cpp
@@ -164,19 +164,28 @@ bool RenderPassAccessContext::ValidateLayoutTransitions(const CommandBufferAcces
         }
         auto hazard = access_context.DetectSubpassTransitionHazard(*track_back, attachment_views[transition.attachment]);
         if (hazard.IsHazard()) {
+            const SyncValidator &sync_state = cb_context.GetSyncState();
             const Location loc(command);
+
+            const syncval_state::ImageViewState *attachment_view = attachment_views[transition.attachment].GetViewState();
+            std::stringstream ss;
+            ss << "in subpass " << subpass << " on attachment " << transition.attachment << " (";
+            ss << sync_state.FormatHandle(attachment_view->Handle());
+            ss << ", " << sync_state.FormatHandle(attachment_view->GetImageState()->Handle());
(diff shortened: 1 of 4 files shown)


layers/sync/sync_renderpass.cpp
```diff
@@ -725,17 +733,25 @@ bool RenderPassAccessContext::ValidateFinalSubpassLayoutTransitions(const Comman
...
             const Location loc(command);
-            if (hazard.Tag() == kInvalidTag) {
-                // Hazard vs. store/resolve
-                const auto error = cb_context.GetSyncState().error_messages_.RenderPassFinalLayoutTransitionVsStoreOrResolveError(
-                    hazard, cb_context, transition.prev_pass, transition.attachment, transition.old_layout, transition.new_layout,
-                    command);
+
+            std::stringstream ss;
+            ss << "on attachment " << transition.attachment << " (";
+            ss << cb_context.GetSyncState().FormatHandle(view_gen.GetViewState()->Handle());
```
spencer-lunarg, line 739: nit (and feel free to ignore) but curious if it would be cleaner to go

`const SyncValidator &sync_state = cb_context.GetSyncState();`

once at the top somewhere

spencer-lunarg (approved):

layers/sync/sync_renderpass.cpp
artem-lunarg: definitely, updated

PR #9544, opened 2025-02-27
sync: Update barrier messages

artem-lunarg:
Temporary contains commits from https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9543

OLD:
> vkCmdPipelineBarrier(): Hazard WRITE_AFTER_WRITE for image barrier 0 VkImage 0xf443490000000006. Access info (usage: SYNC_IMAGE_LAYOUT_TRANSITION, prior_usage: SYNC_COPY_TRANSFER_WRITE, write_barriers: 0, command: vkCmdCopyBufferToImage, resource: VkImage 0xf443490000000006).

NEW:
> vkCmdPipelineBarrier(): WRITE_AFTER_WRITE hazard detected. vkCmdPipelineBarrier performs image layout transition on the VkImage 0xf443490000000006, which was previously written by vkCmdCopyBufferToImage.
No sufficient synchronization is present to ensure that a layout transition does not conflict with a prior write (VK_ACCESS_2_TRANSFER_WRITE_BIT) at VK_PIPELINE_STAGE_2_COPY_BIT.
Image barrier: {
  barrier index = 0,
  src access = 0,
  dst access = 0,
  srcStageMask = VK_PIPELINE_STAGE_2_ALL_TRANSFER_BIT,
  dstStageMask = VK_PIPELINE_STAGE_2_TOP_OF_PIPE_BIT,
}

### DIFF
--- a/layers/sync/sync_error_messages.cpp
+++ b/layers/sync/sync_error_messages.cpp
@@ -105,8 +105,10 @@ static void FormatCommonMessage(const HazardResult& hazard, const std::string& r
     const SyncAccessFlags write_barriers = hazard.State().access_state->GetWriteBarriers();
     const VkPipelineStageFlags2 read_barriers = hazard.State().access_state->GetReadBarriers(hazard.State().prior_access_index);
 
+    // TODO: BOTTOM_OF_PIPE part will go away when syncval switches internally to use NONE/ALL for everything
     const bool missing_synchronization = (hazard_info.IsPriorWrite() && write_barriers.none()) ||
-                                         (hazard_info.IsPriorRead() && read_barriers == VK_PIPELINE_STAGE_2_NONE);
+                                         (hazard_info.IsPriorRead() && (read_barriers == VK_PIPELINE_STAGE_2_NONE ||
+                                                                        read_barriers == VK_PIPELINE_STAGE_2_BOTTOM_OF_PIPE_BIT));
 
     // Brief description of what happened
     ss << string_SyncHazard(hazard_type) << " hazard detected";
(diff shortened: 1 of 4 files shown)


layers/sync/sync_error_messages.cpp
```diff
@@ -478,23 +505,94 @@ std::string ErrorMessages::RenderPassStoreOpError(const HazardResult& hazard, co
...
+
+std::string ErrorMessages::ImagePipelineBarrierError(const HazardResult& hazard, const CommandBufferAccessContext& cb_context,
+                                                     vvl::Func command, const std::string& resource_description,
+                                                     const SyncImageMemoryBarrier& barrier) const {
+    AdditionalMessageInfo additional_info;
+    additional_info.access_action = "performs image layout transition on the";
+
+    std::stringstream ss;
+    ss << "\nImage barrier: {\n";
+    ss << "  barrier index = " << barrier.barrier_index << ",\n";
```
spencer-lunarg, line 581: Not 100% what `index` means here, if it is the `pImageMemoryBarriers[index]` then maybe go

```
pImageMemoryBarriers[3] : {
 srcAcesss = ....
  ....
}
```

as it maps to the code better

layers/sync/sync_error_messages.cpp
```diff
@@ -478,23 +505,94 @@ std::string ErrorMessages::RenderPassStoreOpError(const HazardResult& hazard, co
...
+    ss << "  src access = "
```
spencer-lunarg, line 582: below we use `srcStageMask` (which maps to the API name) so would do

```suggestion
    ss << "  srcAccessMask = "
```

spencer-lunarg (approved):
with comments considered

layers/sync/sync_error_messages.cpp
artem-lunarg: Yeah, was aware of this area (explanation below).. This part is something that need to be improved but trying to finish with the remaining messages to be able to remove old helpers, then the plan to do a round of refactorings and more polish.

The current tracking stores exact form of stages (srcStage/dstStage) but accesses are stored in *internal* form. That's why different naming. This will require a bit of refactoring, for now I will update naming further to show it's not a variable name (*source accesses*). The corresponding fields are printed in the form stage(access1|sccess2).

layers/sync/sync_error_messages.cpp
artem-lunarg: agree, that's better

PR #9555, opened 2025-03-03
docs: Fix a typo on vk_layer_settings for gpuav_image_layout

Lephar:
Change 

> Use GPU-AV to detect which descriptors **where** accessed.

 to 

> Use GPU-AV to detect which descriptors **were** accessed.


artem-lunarg (approved):

artem-lunarg:
@Lephar thanks for the fix!

Issue #9531, opened 2025-02-25
False zero alignment error for BDA GPU-AV


spencer-lunarg:
```glsl
layout(buffer_reference) buffer DrawIndirectBuffer {
    uint drawCount;
    DrawIndirect commands[];
};

layout(buffer_reference) buffer PerDrawBuffer {
    PerDrawInfo info[];
};

struct FrustumCullingInfo {
    vec4 planes[6];
    uint64_t address;
};

layout(set = 0, binding = 0) restrict readonly buffer FrustumCulling {
    uint count;
    FrustumCullingInfo info[];
} frustumCulling;

void main() {
    FrustumCullingInfo frustumCullingInfo = frustumCulling.info[frustumIndex];

    uint drawIndex = atomicAdd(DrawIndirectBuffer(frustumCullingInfo.address).drawCount, 1);

    DrawIndirectBuffer(frustumCullingInfo.address).commands[drawIndex] = inDrawIndirect.commands[objectIndex];
    PerDrawBuffer(frustumCullingInfo.address + DRAW_INDIRECT_MAX_ENTITIES_SIZE).info[drawIndex] = inPerDraw.info[objectIndex];
}
```

generates

> VUID-RuntimeSpirv-PhysicalStorageBuffer64-06315(ERROR / SPEC): vkCmdDispatch(): Unaligned pointer access: The OpStore at buffer device address 0x15208004 is not aligned to the instruction Aligned operand of 0.

which is bogus

ntsh-oni:
Full shader, for information:

	#version 460
	#extension GL_EXT_shader_explicit_arithmetic_types_int64 : require
	#extension GL_EXT_buffer_reference2 : require

	#define DRAW_INDIRECT_MAX_ENTITIES_SIZE 81936

	layout(local_size_x = 8, local_size_y = 8) in;

	struct FrustumCullingInfo {
		vec4 planes[6];
		uint64_t address;
	};

	struct ObjectInfo {
		vec3 position;
		mat4 rotation;
		vec3 scale;
		vec3 aabbMin;
		vec3 aabbMax;
	};

	layout(set = 0, binding = 0) restrict readonly buffer FrustumCulling {
		uint count;
		FrustumCullingInfo info[];
	} frustumCulling;

	layout(set = 0, binding = 1) restrict readonly buffer Objects {
		ObjectInfo info[];
	} objects;

	struct DrawIndirect {
		uint indexCount;
		uint instanceCount;
[...]

PR #9575, opened 2025-03-04
gpuav: Guard against invalid Aligned operands

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9531

Going to add spec text and `spirv-val` support for https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/7210

### DIFF
--- a/layers/gpuav/spirv/buffer_device_address_pass.cpp
+++ b/layers/gpuav/spirv/buffer_device_address_pass.cpp
@@ -69,6 +69,8 @@ void BufferDeviceAddressPass::Reset() {
     type_length_ = 0;
 }
 
+constexpr bool IsPowerOfTwo(uint32_t x) { return x && !(x & (x - 1)); }
+
 bool BufferDeviceAddressPass::RequiresInstrumentation(const Function& function, const Instruction& inst) {
     const uint32_t opcode = inst.Opcode();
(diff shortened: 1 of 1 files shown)


layers/gpuav/spirv/buffer_device_address_pass.cpp
```diff
@@ -69,6 +69,8 @@ void BufferDeviceAddressPass::Reset() {
     type_length_ = 0;
 }
 
+constexpr bool IsPowerOfTwo(uint32_t x) { return x && !(x & (x - 1)); }
```
artem-lunarg, line 72: IsPowerOfTwo from vk_layer_utils.h?

artem-lunarg (approved):

layers/gpuav/spirv/buffer_device_address_pass.cpp
spencer-lunarg: it is, but didn't want to include it here... basically I need to break up more of `vk_layer_utils.h` as its kinda too large of a garbage bin of utils, but its on my post-SDK to do that and then have in notes to fix this spot

artem-lunarg: would be nice to have some math module

PR #9578, opened 2025-03-04
layers: Remove pNext Struct Extension check

spencer-lunarg:
@jeremyg-lunarg @aqnuep here is a quick demonstration of removing the "your pNext struct is missing an extension" validation

Now looking at the tests, something like `VkLayerTest.MissingExtensionStruct` scares me to apply this PR because it would hide an obvious bug if the user tried to use `VkBufferUsageFlags2CreateInfo` and didn't understand why it wasn't working

### DIFF
--- a/scripts/generators/stateless_validation_helper_generator.py
+++ b/scripts/generators/stateless_validation_helper_generator.py
@@ -1060,49 +1060,6 @@ class StatelessValidationHelperOutputGenerator(BaseGenerator):
         if struct.name == 'VkPhysicalDeviceLayeredApiPropertiesListKHR':
             return ""
 
-        if struct.sType and struct.version and all(not x.promotedTo for x in struct.extensions):
-            pNextCheck += f'''
-                if (extensions.api_version < {struct.version.nameApi}) {{
-                    skip |= log.LogError(
-                            pnext_vuid, error_obj.handle, loc.dot(Field::pNext),
-                            "includes a pointer to a VkStructureType ({struct.sType}) which was added in {struct.version.nameApi} but the "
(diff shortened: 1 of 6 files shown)


aqnuep:
I understand the concerns, and I wasn't aware that this validation was even a thing. But I still believe that we shouldn't generate validation errors beyond what the spec requires, even if it may, in some cases, indicate that the application might have forgot to enable something.

Posting my concerns here too related to validating the existence of structs in pNext chains without extension/feature enablement:

1. Nothing in the spec actually requires this, there's no correlation between extensions (not even features) and allowed structs in the pNext chain, and for good reason
2. There's a general API design recommendation that the presence of a struct in the pNext chain on its own should not alter behavior, only setting the fields of that struct to specific behaviors that have preconditions
3. Chaining can often be a "compile-time" choice by applications, e.g. in vulkan hpp it's a template parameter list that controls the structure chain, so by making chaining legality conditioned based on extension/feature enablement can increase application complexity and break existing apps

scripts/generators/stateless_validation_helper_generator.py
```diff
@@ -1060,49 +1060,6 @@ def genStructBody(self, struct: Struct, nonPropFeature: bool, context: str):
         if struct.name == 'VkPhysicalDeviceLayeredApiPropertiesListKHR':
             return ""
 
-        if struct.sType and struct.version and all(not x.promotedTo for x in struct.extensions):
```
spencer-lunarg, line 1063: because this is the only script code that needs to be added (if we want to revert for some reason) it is more confident to do this

PR #9581, opened 2025-03-05
tests: Improve Simple Test creation helpers

spencer-lunarg:
Things like `CreateSamplerTest` and `CreateBufferTest` were a little more clunky then needed to

1. Moved to inside `VkLayerTest` (so don't need to pass it in as a first argument)
2. Remove default string being null (Positive tests can just use `vkt::`)
3. Pass by reference


tests/unit/image_positive.cpp
```diff
@@ -761,3 +761,107 @@ TEST_F(PositiveImage, BlitMaintenance8) {
                      VK_FILTER_NEAREST);
     m_command_buffer.End();
 }
+
+TEST_F(PositiveImage, ImageViewIncompatibleFormat) {
```
spencer-lunarg, line 765: found some Image tests that were doing a lot of positive testing in the negative tests so just moved it here

artem-lunarg (approved):

PR #9591, opened 2025-03-05
cmake: Use generator expression

jpr42:
Now this works on Ninja Multi-Config generator on linux.


CMakeLists.txt
```diff
@@ -160,12 +160,9 @@ if(${CMAKE_CXX_COMPILER_ID} MATCHES "(GNU|Clang)")
             -Wconversion
             -Wimplicit-fallthrough
             -Wstring-conversion
-        )
-
-        if (CMAKE_BUILD_TYPE MATCHES "Debug")
```
jpr42, line 165: Generally speaking using CMAKE_BUILD_TYPE is almost always a bad idea. Use generator expressions.

PR #9595, opened 2025-03-06
tests: Adds a negative testcase with RateControl disabled

zzoon:
Hi, 
This is not only a PR but rather a question.

This test doesn't pass but I think it should pass. That's why I suspect it's about VVL's bug.   
But unfortunately I didn't manage to fix how.   

What do you think? Is it a real bug of VVL or there's something I'm missing.

### DIFF
--- a/tests/unit/video_encode.cpp
+++ b/tests/unit/video_encode.cpp
@@ -1083,6 +1083,41 @@ TEST_F(NegativeVideoEncode, RateControlStateMismatchNotDefault) {
     m_device->Wait();
 }
 
+TEST_F(NegativeVideoEncode, RateControlDisabledStateMismatch) {
+    TEST_DESCRIPTION("vkCmdBeginVideoCodingKHR - A case where rate control state specified as disabled");
+    RETURN_IF_SKIP(Init());
+
+    VideoConfig config =
+        GetConfig(GetConfigsWithRateControl(GetConfigsEncode(), VK_VIDEO_ENCODE_RATE_CONTROL_MODE_DISABLED_BIT_KHR));
+
+    if (!config) {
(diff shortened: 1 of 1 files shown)


aqnuep:
No, this shouldn't pass.

I think the confusion stems from the misunderstanding about the default rate control mode.

When the video session is created, the default rate control mode is **not** `VK_VIDEO_ENCODE_RATE_CONTROL_MODE_DISABLED_BIT_KHR` (which is the constant QP mode), but `VK_VIDEO_ENCODE_RATE_CONTROL_MODE_DEFAULT_KHR` (which is the default, implementation-specific rate control mode).

So this test will fail because you provided an RC mode in the BeginInfo that does not match the actual RC mode currently set for the video session (i.e. the default mode that video sessions are in after creation).

In order to set the `VK_VIDEO_ENCODE_RATE_CONTROL_MODE_DISABLED_BIT_KHR` mode, you have to use `vkCmdControlVideoKHR` to update the RC mode.

Then, the next time you begin a new video coding scope, you should specify the previously set mode, but the first time around (before setting any RC mode), the mode should be `VK_VIDEO_ENCODE_RATE_CONTROL_MODE_DEFAULT_KHR` (which is the default assumed if you don't specify any RC info at video-coding-scope-begin-time).

zzoon:
Thanks for the explaination.
I rewrote an negative test regarding this.   
If you don't think it makes sense to land, feel free to close this.

aqnuep:
There are already similar tests, but there should be no harm adding this too.

spencer-lunarg (approved):
@aqnuep did this look good to you to add now?

aqnuep:
Sure, to some extent it is redundant (there's a more generic test case), but it does cover a special case explicitly with its own test, so should be fine to merge.

PR #9598, opened 2025-03-06
layers: Extract stateless device data initialization

aqnuep:
Here's the promised follow-up refactoring extracting the stateless device data into its own atomically initialized class.

Unfortunately, I had to keep `extensions` and `enabled_features` mutable, because there's some patching done in the chassis of `CreateDevice`.

### DIFF
--- a/layers/chassis/dispatch_object.h
+++ b/layers/chassis/dispatch_object.h
@@ -68,6 +68,10 @@ namespace base {
 class Instance;
 class Device;
 }  // namespace base
+namespace dispatch {
+class Instance;
+class Device;
+}  // namespace dispatch
 
 // Device extension properties -- storing properties gathered from VkPhysicalDeviceProperties2::pNext chain
 // TODO: this could be defined and initialized via generated code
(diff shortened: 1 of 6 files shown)


aqnuep:
Btw, that patching in the chassis seems to be only needed by GPUAV to force-enable some used features, which is reasonable, but I'm not sure whether the enabled extensions/features, as seen by the rest of the validation layers should actually reflect the GPUAV patched values. IMO validation should only see the application provided values, even if GPUAV force-enables those things when calling down into the ICD itself. @jeremyg-lunarg, any thoughts?

jeremyg-lunarg (approved):
LGTM, thanks!

jeremyg-lunarg:
> IMO validation should only see the application provided values, even if GPUAV force-enables those things when calling down into the ICD itself. @jeremyg-lunarg, any thoughts?

I agree. I'll make an issue for that and one of us working on GPU-AV can take it on.

spencer-lunarg (approved):

PR #9603, opened 2025-03-06
gpuav: Skip State Tracking Image Layout when off

spencer-lunarg:
@arno-lunarg showed we were spending a lot of time in `UpdateDrawState` in GPU-AV even when Image Layout validation is off

1. Check if setting is off and skip work
2. Rename function to be clear this is only updating Image Layout state

### DIFF
--- a/layers/drawdispatch/descriptor_validator.cpp
+++ b/layers/drawdispatch/descriptor_validator.cpp
@@ -202,16 +202,20 @@ bool DescriptorValidator::ValidateBindingDynamic(const spirv::ResourceInterfaceV
             skip |= ValidateDescriptorsDynamic(resource_variable, static_cast<const BufferBinding &>(binding), index);
             break;
         case DescriptorClass::ImageSampler: {
-            auto &imgs_binding = static_cast<ImageSamplerBinding &>(binding);
-            auto &descriptor = imgs_binding.descriptors[index];
-            descriptor.UpdateDrawState(cb_state);
-            skip |= ValidateDescriptorsDynamic(resource_variable, imgs_binding, index);
+            auto &img_sampler_binding = static_cast<ImageSamplerBinding &>(binding);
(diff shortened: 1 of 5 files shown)


layers/drawdispatch/descriptor_validator.cpp
```diff
@@ -202,16 +202,20 @@ bool DescriptorValidator::ValidateBindingDynamic(const spirv::ResourceInterfaceV
...
             break;
         case DescriptorClass::ImageSampler: {
-            auto &imgs_binding = static_cast<ImageSamplerBinding &>(binding);
-            auto &descriptor = imgs_binding.descriptors[index];
-            descriptor.UpdateDrawState(cb_state);
-            skip |= ValidateDescriptorsDynamic(resource_variable, imgs_binding, index);
+            auto &img_sampler_binding = static_cast<ImageSamplerBinding &>(binding);
+            if (dev_state.gpuav_settings.validate_image_layout) {
+                auto &descriptor = img_sampler_binding.descriptors[index];
+                descriptor.UpdateImageLayoutDrawState(cb_state);
```
arno-lunarg, line 208: The logic here is weird, essentially you have `A.Foo(B)` and `Foo(B)` just goes `B.Bar(A)`.
But when you see `A.Foo(B)` you expect A to update its state after looking at B, not the reverse.

IMO would make more sense to here just have:
`cb_state.UpdateImageLayout(descriptor);`
It is a more natural pattern

arno-lunarg (approved):
And nice renaming, `UpdateDrawState` was king

layers/drawdispatch/descriptor_validator.cpp
spencer-lunarg: not the first time something around Image Layout was non-ideal, agree should fix if anyone gets to it

PR #9605, opened 2025-03-06
sync: Fix getting SyncAccessFlags from VkAccessFlags2

artem-lunarg:
Fixes major gap in syncval ray tracing testing, where simple scenario did not work at all


layers/sync/sync_utils.cpp
```diff
@@ -62,6 +62,9 @@ VkPipelineStageFlags2 DisabledPipelineStages(const DeviceFeatures &features, con
     if (!features.rayTracingMaintenance1) {
         result |= VK_PIPELINE_STAGE_2_ACCELERATION_STRUCTURE_COPY_BIT_KHR;
     }
+    if (!features.micromap) {
```
artem-lunarg, line 65: side effect of the fix: this has to be here but everything worked without it and now we finally have real use case and test to show it's needed.

spencer-lunarg (approved):

PR #9607, opened 2025-03-06
layers: Cleanup for deferring to GPU-AV

spencer-lunarg:
Taking the good out of https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9604

### DIFF
--- a/layers/core_checks/cc_descriptor.cpp
+++ b/layers/core_checks/cc_descriptor.cpp
@@ -846,7 +846,7 @@ bool CoreChecks::ValidateDrawState(const vvl::DescriptorSet &descriptor_set, uin
             return result;
         }
 
-        if (descriptor_set.ValidateBindingOnGPU(*binding, resource_variable.is_runtime_descriptor_array)) {
+        if (descriptor_set.ValidateBindingOnGPU(*binding, resource_variable)) {
             continue;
         }
 
--- a/layers/state_tracker/descriptor_sets.cpp
+++ b/layers/state_tracker/descriptor_sets.cpp
@@ -698,9 +698,10 @@ void vvl::DescriptorSet::UpdateImageLayoutDrawStates(vvl::Device *device_data, v
         ASSERT_AND_CONTINUE(binding);
 
         // core validation doesn't handle descriptor indexing, that is only done by GPU-AV
-        if (ValidateBindingOnGPU(*binding, binding_req_pair.second.variable->is_runtime_descriptor_array)) {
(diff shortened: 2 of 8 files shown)


layers/core_checks/cc_descriptor.cpp
```diff
@@ -845,9 +845,7 @@ bool CoreChecks::ValidateDrawState(const vvl::DescriptorSet &descriptor_set, uin
             return result;
         }
 
-        if (descriptor_set.ValidateBindingOnGPU(*binding, resource_variable.is_runtime_descriptor_array)) {
-            continue;
```
jeremyg-lunarg, line 849: not crazy about single line ifs, being able to set a breakpoint here could be occasionally useful.

layers/state_tracker/descriptor_sets.cpp
```diff
@@ -732,13 +731,19 @@ void vvl::DescriptorSet::UpdateDrawStates(vvl::Device *device_data, vvl::Command
...
-bool vvl::DescriptorSet::ValidateBindingOnGPU(const DescriptorBinding &binding, bool is_runtime_descriptor_array) const {
+bool vvl::DescriptorSet::ValidateBindingOnGPU(const DescriptorBinding &binding,
+                                              const spirv::ResourceInterfaceVariable &variable) const {
     // Some applications (notably Doom Eternal) might have large non-bindless descriptors attached (basically doing Descriptor
     // Indexing without the extension). Trying to loop through these on the CPU will bring FPS down by over 50% so we make use of
     // the post processing to detect which descriptors were actually accessed
     static constexpr uint32_t max_descriptor_on_cpu = 1024;
-    return GetNonInlineDescriptorCount() > max_descriptor_on_cpu || IsBindless(binding.binding_flags) ||
-           is_runtime_descriptor_array;
+    return
```
jeremyg-lunarg, line 740: 🤯  You could write this more clearly as:
```
if (thing1) {
    return true;
}
if (thing2) {
   return true;
}
return thing3;
```

layers/core_checks/cc_descriptor.cpp
arno-lunarg: +1 !

layers/state_tracker/descriptor_sets.cpp
arno-lunarg: +1

arno-lunarg (approved):
LGTM with changes mentioned by Jeremy

PR #9608, opened 2025-03-06
layers: Improve small_vector initialization for better memory handling

nrmu9:
This pull request includes an update to the `small_vector` class in the `layers/containers/custom_containers.h` file to improve the initialization process and ensure proper memory allocation. Fixes an issue where ninja would fail to compile the x64 windows release triplet.

Changes to `small_vector` class:

* Added a check to ensure the `reserve` method is called only if the size is greater than zero, preventing unnecessary memory allocation.
* Fixed the initialization loop to correctly construct elements in the allocated memory using the `placement new` operator.

### DIFF
--- a/layers/containers/custom_containers.h
+++ b/layers/containers/custom_containers.h
@@ -160,13 +160,14 @@ class small_vector {
     }
 
     small_vector(size_type size, const value_type &value = value_type()) : size_(0), capacity_(N), working_store_(GetSmallStore()) {
-        reserve(size);
-        auto dest = GetWorkingStore();
-        for (size_type i = 0; i < size; i++) {
-            new (dest) value_type(value);
-            ++dest;
+        if (size > 0) {
+            reserve(size);
+            auto dest = GetWorkingStore();
+            for (size_type i = 0; i < size; i++) {
+                new (&dest[i]) value_type(value);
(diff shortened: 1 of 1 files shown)


jeremyg-lunarg:
Thanks for the PR, it looks like a good change to me.

But you have a merge commit in this PR. Can you please do a `git fetch origin; git rebase origin/main` and re-push?

jeremyg-lunarg (approved):

jeremyg-lunarg:
> @jeremyg-lunarg I would also highly recommend you push an update to `vcpkg` with the latest commits if possible. Building x64 windows static release fails on MSVC 14.44.34823 and C/C++ Optimizing Compiler Version 19.44.34823.2 for x64 without them.

@nrmu9 can you share the error messages you got?

jeremyg-lunarg:
14.44.34823 is a preview release of MSVC, so I reported this internal error to MS:

https://developercommunity.visualstudio.com/t/Vulkan-ValidationLayers-internal-compile/10867048

Hopefully they'll be able to fix it before making the official release.

jeremyg-lunarg:
Also, this will be in the next SDK release which is happening shortly. vckpg should be updated at that time

PR #9609, opened 2025-03-06
gpuav: Store modified extensions and features locally

### LINKED ISSUE #9599: gpuav: Force enabled extensions and features should be 'hidden' from the rest of validation
In https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9598, it was pointed out that GPU-AV modifies the `DeviceExtensions` and `DeviceFeatures` data used by the rest of validation. This could cause incorrect results, most likely false negatives, if an application uses features without enabling them itself.  GPU-AV should track the extensions and features it enabled separately from the main data structures.

jeremyg-lunarg:
Rather than putting stuff that was turned on by GPU-AV into `StatelessDeviceData`, store it separately so that the rest of validation can use what the application enabled.

This exposed quite a few validation errors in the unit tests.

Fixes #9599


layers/gpuav/core/gpuav_setup.cpp
```diff
@@ -63,7 +63,7 @@ std::shared_ptr<vvl::AccelerationStructureKHR> Validator::CreateAccelerationStru
         if (buf_state->deviceAddress != 0) {
             buffer_address = buf_state->deviceAddress;
         } else if (buf_state->Binding()) {
-            buffer_address = GetBufferDeviceAddressHelper(buf_state->VkHandle());
+            buffer_address = GetBufferDeviceAddressHelper(buf_state->VkHandle(), &modified_extensions);
```
jeremyg-lunarg, line 66: I'm not happy with this part yet. GPU-AV turns on BDA and so it needs to use the modified extensions in the helper.

arno-lunarg: Since `GetBufferDeviceAddressHelper` is a methog of `validator` and  `modified_extensions` one of its member, no need to pass it as a parameter?

arno-lunarg (approved):

layers/gpuav/core/gpuav_setup.cpp
jeremyg-lunarg: It is in `vvl::Device` now, and used somewhere at that level, which is why it is a problem.

jeremyg-lunarg:
made this draft because it needs to wait until after the sdk

tests/unit/gpu_av.cpp
```diff
@@ -66,6 +66,7 @@ TEST_F(NegativeGpuAV, SelectInstrumentedShaders) {
     TEST_DESCRIPTION("GPU validation: Validate selection of which shaders get instrumented for GPU-AV");
     SetTargetApiVersion(VK_API_VERSION_1_2);
     AddRequiredFeature(vkt::Feature::robustBufferAccess);
+    AddRequiredFeature(vkt::Feature::vertexPipelineStoresAndAtomics);
```
spencer-lunarg, line 69: So we can't expect user to go and add features in order to use GPU-AV

The idea is we have a part where we do force features on for them, because they are likely supported, just not on

I think what we need is a `modified_features` copy with the `enabled_features`. The idea is we pass through `modified_features` down to the dispatch of device creation, but for Core Checks, it can just use the `enabled_feature` and pretend it wasn't enabled.

GPU-AV can be cautious what things were enabled from the user vs us (I already have plans to have that difference for something like robustness)

spencer-lunarg (changes requested):
This is a bit off, we should not require apps to enable features to use GPU-AV

tests/unit/gpu_av.cpp
arno-lunarg: Also where does those new feature requirements come from?

jeremyg-lunarg: `vertexPipelineStoresAndAtomics specifies whether storage buffers and images support stores and atomic operations in the vertex, tessellation, and geometry shader stages. If this feature is not enabled, all storage image, storage texel buffer, and storage buffer variables used by these stages in shader modules must be decorated with the NonWritable decoration (or the readonly memory qualifier in GLSL).`

This shader and all the other ones where I added features, are storing to a storage buffer and they must have this feature to be valid even without GPU-AV. The ones with read only access I fixed by adding `readonly` to the declarations.

jeremyg-lunarg: See also: VUID-RuntimeSpirv-NonWritable-06340 and VUID-RuntimeSpirv-NonWritable-06341

spencer-lunarg (approved):

PR #9612, opened 2025-03-06
layers: Better ShaderInputNotProduced error message

spencer-lunarg:
new message

```
Validation Warning: [ Undefined-Value-ShaderInputNotProduced-DynamicRendering ] | MessageID = 0x733b5ff8
vkCmdDrawIndexed(): Inside the fragment shader, the output Locaiton 1 was never written to. This means the bound VkRenderingInfo::pColorAttachments[1].imageView (VkImageView 0x3a6cbb0000000025) will have undefined values written to it. The pipeline was created with pColorBlendState->pAttachments[1].colorWriteMask set to 0xf so setting it to zero is one way to prevent undefined values overriding your color attachment (this will require independentBlend, which is basically supported everywhere, to have some attachment have different colorWriteMask).
Spec information at https://docs.vulkan.org/spec/latest/chapters/interfaces.html#interfaces-fragmentoutput
Objects: 2
    [0] VkCommandBuffer 0x644d8413f0e0
    [1] VkPipeline 0xc25f26000000009c
```

### DIFF
--- a/layers/core_checks/cc_shader_interface.cpp
+++ b/layers/core_checks/cc_shader_interface.cpp
@@ -516,19 +516,8 @@ bool CoreChecks::ValidateFsOutputsAgainstRenderPass(const spirv::Module &module_
         const VkAttachmentDescription2 *attachment = attachment_info.attachment;
         const spirv::StageInterfaceVariable *output = attachment_info.output;
         if (attachment && !output) {
-            const auto &attachment_states = pipeline.AttachmentStates();
-            if (location < attachment_states.size() && attachment_states[location].colorWriteMask != 0) {
-                skip |= LogUndefinedValue(
-                    "Undefined-Value-ShaderInputNotProduced", module_state.handle(), create_info_loc,
-                    "Inside the fragment shader, the output Locaiton %" PRIu32
-                    " was never written to. This means anything future VkSubpassDescription::pColorAttachments[%" PRIu32
-                    "] will have undefined values written to it. The pipeline was created with "
-                    "pColorBlendState->pAttachments[%" PRIu32 "].colorWriteMask set to 0x%" PRIx32
-                    " so setting it to zero is one way to prevent undefined values overriding your color attachment.\nSpec "
-                    "information at "
-                    "https://docs.vulkan.org/spec/latest/chapters/interfaces.html#interfaces-fragmentoutput",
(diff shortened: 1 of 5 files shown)


layers/core_checks/cc_shader_interface.cpp
```diff
@@ -605,14 +606,29 @@ bool CoreChecks::ValidateDrawDynamicRenderingFsOutputs(const LastBound &last_bou
             const auto image_view_state = Get<vvl::ImageView>(attachment_info.rendering_attachment_info->imageView);
             const VkColorComponentFlags color_write_mask = last_bound_state.GetColorWriteMask(location);
             if (color_write_mask != 0) {
+                std::ostringstream msg;
+                msg << "Inside the fragment shader, the output Locaiton " << location
```
arno-lunarg, line 610:
```suggestion
                msg << "Inside the fragment shader, the output Location" << location
```

layers/core_checks/cc_shader_interface.cpp
```diff
@@ -605,14 +606,29 @@ bool CoreChecks::ValidateDrawDynamicRenderingFsOutputs(const LastBound &last_bou
...
+                    msg << "The pipeline was created with pColorBlendState->pAttachments[" << location
+                        << "].colorWriteMask set to ";
+                } else {
+                    msg << "The last call to vkCmdSetColorWriteMaskEXT for attachment " << location
+                        << " set the colorWriteMask to ";
+                }
+                msg << "0x" << std::hex << (uint32_t)color_write_mask
+                    << " so setting it to zero is one way to prevent undefined values overriding your color attachment";
+                if (color_attachment_count > 1) {
+                    msg << " (this will require independentBlend, which is basically supported everywhere, to have some attachment "
```
arno-lunarg, line 624:
```suggestion
                    msg << " (this will require independentBlend, which is basically supported everywhere, to have some attachments "
```
(not sure?)

arno-lunarg (approved):

layers/core_checks/cc_shader_interface.cpp
spencer-lunarg: no, I want it to say `Location 3` not `Location3`

layers/core_checks/cc_shader_interface.cpp
spencer-lunarg: will fix

PR #9637, opened 2025-03-11
gpuav: Misc optimisations

arno-lunarg:
Some small optimizations I made while investigating GPU-AV performance. For the lightest instrumentation, done for "out of bounds vertex attributes fetching", went from 20 -> 30 fps on Doom Eternal. 
When enabling other types of instrumentation, unfortunately bottlenecks totally shift to other spots specific to instrumentation types.

### DIFF
--- a/layers/gpuav/descriptor_validation/gpuav_descriptor_set.cpp
+++ b/layers/gpuav/descriptor_validation/gpuav_descriptor_set.cpp
@@ -219,7 +219,7 @@ VkDeviceAddress DescriptorSet::GetTypeAddress(Validator &gpuav, const Location &
         return 0;
     }
 
-    auto data = (glsl::DescriptorState *)input_buffer_.MapMemory(loc);
+    auto data = (glsl::DescriptorState *)input_buffer_.GetMappedPtr(loc);
 
     uint32_t index = 0;
     for (uint32_t i = 0; i < bindings_.size(); i++) {
@@ -256,7 +256,6 @@ VkDeviceAddress DescriptorSet::GetTypeAddress(Validator &gpuav, const Location &
 
     // Flush the descriptor state buffer before unmapping so that the new state is visible to the GPU
     input_buffer_.FlushAllocation(loc);
-    input_buffer_.UnmapMemory();
 
     return input_buffer_.Address();
 }
@@ -297,11 +296,10 @@ VkDeviceAddress DescriptorSet::GetPostProcessBuffer(Validator &gpuav, const Loca
 
     VVL_TracyPlot("Post process buffer size (bytes)", int64_t(buffer_info.size));
 
-    void *data = post_process_buffer_.MapMemory(loc);
+    void *data = post_process_buffer_.GetMappedPtr(loc);
     memset(data, 0, static_cast<size_t>(buffer_info.size));
 
     post_process_buffer_.FlushAllocation(loc);
-    post_process_buffer_.UnmapMemory();
 
     return post_process_buffer_.Address();
 }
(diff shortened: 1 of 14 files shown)


layers/state_tracker/queue_state.cpp
```diff
@@ -260,6 +262,8 @@ void vvl::Queue::Retire(QueueSubmission &submission) {
 }
 
 void vvl::Queue::ThreadFunc() {
+    VVL_TracySetThreadName(__FUNCTION__);
```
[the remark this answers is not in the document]
arno-lunarg: We could steal the implementation, but not directly use `VVL_TracySetThreadName` or the underlying `tracy::SetThreadName` function since they belong to Tracy

layers/gpuav/descriptor_validation/gpuav_descriptor_set.cpp
```diff
@@ -293,24 +294,26 @@ VkDeviceAddress DescriptorSet::GetPostProcessBuffer(Validator &gpuav, const Loca
...
     post_process_buffer_.FlushAllocation(loc);
-    post_process_buffer_.UnmapMemory();
 
     return post_process_buffer_.Address();
 }
 
 // cross checks the two buffers (our layout with the output from the GPU-AV run) and builds a map of which indexes in which binding
 // where accessed
 std::vector<DescriptorAccess> DescriptorSet::GetDescriptorAccesses(const Location &loc, uint32_t shader_set) const {
+    VVL_ZoneScoped;
```
spencer-lunarg, line 310: remove

layers/gpuav/descriptor_validation/gpuav_descriptor_set.cpp
```diff
@@ -293,24 +294,26 @@ VkDeviceAddress DescriptorSet::GetPostProcessBuffer(Validator &gpuav, const Loca
         return 0;
     }
 
-    void *data = post_process_buffer_.MapMemory(loc);
+    VVL_TracyPlot("Post process buffer size (bytes)", int64_t(buffer_info.size));
```
spencer-lunarg, line 297: remove

layers/gpuav/descriptor_validation/gpuav_descriptor_set.cpp
```diff
@@ -23,6 +23,8 @@
 #include "gpuav/shaders/gpuav_shaders_constants.h"
 #include "state_tracker/descriptor_sets.h"
 
+#include "profiling/profiling.h"
```
spencer-lunarg, line 26: remove

layers/gpuav/descriptor_validation/gpuav_descriptor_validation.cpp
```diff
@@ -23,6 +23,8 @@
 #include "gpuav/resources/gpuav_shader_resources.h"
 #include "state_tracker/shader_module.h"
 
+#include "profiling/profiling.h"
```
spencer-lunarg, line 26: remove

layers/gpuav/descriptor_validation/gpuav_descriptor_set.cpp
arno-lunarg: No for every Tracy comment, we need to keep an eye on those problematic areas

spencer-lunarg: sure

layers/gpuav/descriptor_validation/gpuav_descriptor_set.cpp
spencer-lunarg: these `VVL_TracyPlot` I rather not keep, we have nothing automatically tracking this... one day when you are done with Post Process no one will remove this and we will have a world where no one knows which `VVL_TracyPlot` to  keep or not

spencer-lunarg: unless we agree every so often we just go by and clear them all, when do we ever remove them, or is the idea to just leave them anywhere anyone wants?

arno-lunarg: That's a legit worry, and it also applies to the manually placed `VVL_ZoneScoped`. I will remove this one, since we are likely to make changes in that area this plot might soon get obsolete. 
The idea behind keeping a plot would be "is it always useful to know that or just needed for my current work". Always plotting instrumentations counts could be nice, it would help correlate to pipeline compilation times.

spencer-lunarg (approved):

PR #9642, opened 2025-03-11
gpuav: Dont check combined image sampler when possible

spencer-lunarg:
HLSL/Slang doesn't use Combined Image Samplers, but we have extra checks in every descriptor indexing for it

What makes it worse is we recently added 2 checks for Sampled Image and the Sampler so it was doubling the amount of times we didn't even need this part of the check

See some good small perf win for compile times with this


layers/gpuav/shaders/instrumentation/descriptor_indexing_oob_bindless_combined_image_sampler.comp
```diff
@@ -0,0 +1,129 @@
+// Copyright (c) 2024-2025 The Khronos Group Inc.
```
spencer-lunarg, line 1: this is copy-and-paste junk, but makes it more appealing to get Slang working so we can do proper modules and need less copy-and-pasting like this

layers/gpuav/spirv/descriptor_indexing_oob_pass.cpp
```diff
@@ -32,17 +33,43 @@ uint32_t DescriptorIndexingOOBPass::GetLinkFunctionId() {
...
                                               instrumentation_descriptor_indexing_oob_non_bindless_comp_size, 0,
                                               "inst_descriptor_indexing_oob_non_bindless"};
 
-    if (link_function_id == 0) {
-        link_function_id = module_.TakeNextId();
-        LinkInfo& link_info = module_.has_bindless_descriptors_ ? link_info_bindless : link_info_non_bindless;
-        link_info.function_id = link_function_id;
-        module_.link_info_.push_back(link_info);
+    uint32_t link_id = 0;
+    if (!module_.has_bindless_descriptors_) {
```
spencer-lunarg, line 47: didn't know how to abstract this, we have 3 functions that can be created, so each has their own "last used" member variable

layers/gpuav/spirv/descriptor_indexing_oob_pass.h
```diff
@@ -43,6 +45,7 @@ class DescriptorIndexingOOBPass : public InjectConditionalFunctionPass {
     uint32_t descriptor_binding_ = 0;
     uint32_t descriptor_index_id_ = 0;
 
+    bool is_combined_image_sampler_ = false;
```
spencer-lunarg, line 48: so thinking more about the Unsafe PR, I think the issue is we have member variables that are "per instruction" and some that are "per pass" and a follow up PR I think of trying to make it more obvious, that way things like `RequireInstrumentation` can take a "per instruction" object

tests/unit/gpu_av_descriptor_indexing.cpp
```diff
@@ -3491,6 +3491,54 @@ TEST_F(NegativeGpuAVDescriptorIndexing, SamplersMixedBindless2) {
...
+
+    const char cs_source[] = R"glsl(
+        #version 460
+        layout(set = 0, binding = 0) uniform texture2D kTextures2D;
+        layout(set = 0, binding = 1) uniform sampler kSampler;
+        layout(set = 0, binding = 2) uniform sampler2D kCombinedImageSampler;
+
+        void main() {
+            vec4 out_color = texture(sampler2D(kTextures2D, kSampler), vec2(0));
+            out_color += texture(kCombinedImageSampler, vec2(0));
```
spencer-lunarg, line 3509: this now looks like the following instrumented

```
    bool _49 = inst_descriptor_indexing_oob_bindless(41u, _41, 0u, 0u, 0u, 1u, 0u);
    bool _52 = inst_descriptor_indexing_oob_bindless(41u, _41, 0u, 1u, 0u, 1u, 1u);
    vec4 _46;
    if (_49 && _52) {
        _46 = texture(sampler2D(kTextures2D, kSampler), vec2(0.0));
    } else {
        _46 = vec4(0.0);
    }
    out_color = _46;

    bool _60 = inst_descriptor_indexing_oob_bindless_combined_image_sampler(44u, _41, 0u, 2u, 0u, 1u, 2u);
    vec4 _58;
    if (_60) {
        _58 = texture(kCombinedImageSampler, vec2(0.0));
    } else {
        _58 = vec4(0.0);
    }
```

PR #9644, opened 2025-03-11
gpuav: Add more Post Processing tests

spencer-lunarg:
3 extra tests for @arno-lunarg to test with

### DIFF
--- a/tests/unit/gpu_av_descriptor_post_process.cpp
+++ b/tests/unit/gpu_av_descriptor_post_process.cpp
@@ -1555,3 +1555,243 @@ TEST_F(NegativeGpuAVDescriptorPostProcess, ImageViewArrayAliasBinding) {
     m_default_queue->Wait();
     m_errorMonitor->VerifyFound();
 }
+
+// TODO - Currently don't detect the 2nd use of the descriptor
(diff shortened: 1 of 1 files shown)


tests/unit/gpu_av_descriptor_post_process.cpp
```diff
@@ -1491,3 +1491,242 @@ TEST_F(NegativeGpuAVDescriptorPostProcess, DISABLED_DescriptorIndexingSlang) {
     m_default_queue->Wait();
     m_errorMonitor->VerifyFound();
 }
+
+// TODO - Currently don't detect the 2nd use of the descriptor
+TEST_F(NegativeGpuAVDescriptorPostProcess, DISABLED_SameDescriptorAcrossSets) {
```
spencer-lunarg, line 1496: we have this code

```
            if (validated_desc_sets.count(bound_descriptor_set->VkHandle()) > 0) {
                // TODO - If you share two VkDescriptorSet across two different sets in the SPIR-V, we are not going to be
                // validating the 2nd instance of it
                continue;
            }
```

that I would like to hopefully remove, while here a test for it

arno-lunarg (approved):

PR #9646, opened 2025-03-12
printf: Add support for printing pointers

spencer-lunarg:
sick of not being able to go `%p` to print a pointer... well now we can


arno-lunarg (approved):
Nice!

PR #9651, opened 2025-03-12
layers: Remove state object inheritance

jeremyg-lunarg:
In order to have a shared state tracker, customization of state objects by inheritance must end. Instead add sub states, which allow each validation type to add their own data to the common state object.


layers/state_tracker/state_object.h
```diff
@@ -143,4 +145,21 @@ class RefcountedStateObject : public StateObject {
...
     const VulkanTypedHandle* InUse() const override { return ((in_use_.load() > 0) || StateObject::InUse()) ? &Handle() : nullptr; }
 };
+
+template <typename T>
+class SubStateManager {
+  public:
+    void SetSubState(LayerObjectTypeId id, std::unique_ptr<T> &&substate) { substates.emplace(id, std::move(substate)); }
+    void RemoveSubState(LayerObjectTypeId id) { substates.erase(id); }
+
+    T *SubState(LayerObjectTypeId id) { return substates[id].get(); }
```
artem-lunarg, line 155: is different implementation here comparing to const for simplicity? I understand why this can't be used in const version, but also non-const version has ability to insert elements comparing to const version. Is this insertion used anywhere? If not, I personally would prefer more verbose const impementation just because it does not modify.

layers/chassis/dispatch_object.h
```diff
@@ -35,18 +35,7 @@
...
-    LayerObjectTypeThreading,            // Instance or device threading layer object
-    LayerObjectTypeParameterValidation,  // Instance or device parameter validation layer object
-    LayerObjectTypeObjectTracker,        // Instance or device object tracker layer object
-    LayerObjectTypeCoreValidation,       // Instance or device core validation layer object
-    LayerObjectTypeBestPractices,        // Instance or device best practices layer object
-    LayerObjectTypeGpuAssisted,          // Instance or device gpu assisted validation layer object
-    LayerObjectTypeSyncValidation,       // Instance or device synchronization validation layer object
-    LayerObjectTypeMaxEnum,              // Max enum count
-};
+#include "layer_object_id.h"
```
spencer-lunarg, line 38: heads up @aqnuep, I assume you have this modified in the fork, just gonna be in a separate file now

layers/state_tracker/image_state.h
```diff
@@ -119,7 +121,6 @@ class Image : public Bindable {
 #endif  // VK_USE_PLATFORM_METAL
 
     const image_layout_map::Encoder subresource_encoder;                             // Subresource resolution encoder
-    std::unique_ptr<const subresource_adapter::ImageRangeEncoder> fragment_encoder;  // Fragment resolution encoder
```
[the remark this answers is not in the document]
jeremyg-lunarg: moved to the base class

layers/gpuav/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -35,7 +35,7 @@ namespace gpuav {
 
 // If application is using shader objects, bindings count will be computed from bound shaders
 static uint32_t LastBoundPipelineOrShaderDescSetBindingsCount(Validator &gpuav, VkPipelineBindPoint bind_point,
-                                                              CommandBuffer &cb_state, const LastBound &last_bound) {
+                                                              CommandBufferSubState &cb_state, const LastBound &last_bound) {
```
spencer-lunarg, line 38: an exercise for me afterwards, we don't even use `cb_state` here... will be something for GPU-AV I will need to scrub afterwards (just noting out loud)

layers/state_tracker/state_object.h
```diff
@@ -143,4 +145,21 @@ class RefcountedStateObject : public StateObject {
...
+    const T *SubState(LayerObjectTypeId id) const {
+        auto iter = substates.find(id);
+        return iter != substates.end() ? iter->second.get() : nullptr;
+    }
+
+  protected:
+    std::map<LayerObjectTypeId, std::unique_ptr<T>> substates;
```
artem-lunarg, line 162: Not an ask to change something here, but an idea and maybe to experiment later. Since we have only a few of `LayerObjectTypeId` (currently 7) we can have a `std::array<std::unique_ptr<T>, 7>.  Then we can index there directly by LayerObjectTypeId. Memory wise it will be 7 * 8 (typical size of unique ptr) = 56 bytes. std::map object on  my machine it's 24 bytes plus at least single node that allocates (allocation metadata) and single unique_ptr that's already 40 bytes. If more than one substate this can take more since than std::array, but direct indexing would be much faster.

layers/state_tracker/descriptor_sets.cpp
```diff
@@ -752,13 +758,13 @@ bool vvl::DescriptorSet::ValidateBindingOnGPU(const DescriptorBinding &binding,
 // src and dst are shared pointers.
 template <typename T>
 static void ReplaceStatePtr(DescriptorSet &set_state, T &dst, const T &src, bool is_bindless) {
-    if (dst && !is_bindless) {
```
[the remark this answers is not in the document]
jeremyg-lunarg: To get rid of the many state object inheritances in GPU-AV, I have to track the bindless descriptor state instead of skipping it. I can go back and remove the parameter

layers/state_tracker/state_object.h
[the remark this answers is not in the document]
spencer-lunarg: more naming nit: we use `sub_state` and `SubState` everywhere else so should be 2 words here

jeremyg-lunarg: Hmm. Maybe and really only a subset of the members of  `LayerObjectTypeId `  (core, bp, syncval, gpuav) will use the state tracker so it could be even smaller. But need to come up with a way to express that.

spencer-lunarg (approved):
- Overall really glad we are getting this!
- Glanced at BP, looks fine
- GPU-AV state tracking all looks good
- Descriptor tracking looks good
- The Queue SubState is still fuzzy to me how it worked, but seemed good
- Quick glanced Sync code, assume Artem will help verify it

layers/state_tracker/descriptor_sets.cpp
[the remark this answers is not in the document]
spencer-lunarg: I think I saw "why" after I commented, but ya, just remove the parameter then

layers/state_tracker/state_object.h
artem-lunarg: Yeah, but if memory question is important. I was thinking about performance with direct indexing no need to have conditions. Might depends on what what we need more.

jeremyg-lunarg: I'll play around with this in a separate PR

artem-lunarg (approved):
SyncVal code looks good, it's a lot of changes but implementation is not complicated and nice to read.

PR #9653, opened 2025-03-12
gpuav: Remove more pass member variables

spencer-lunarg:
Most passes have a common trend

1. Call `RequiresInstrumentation` and build up values
2. Call `CreateFunctionCall` and use those values

Currently we just stored everything as a global member variable which made it hard to grasp the lifetime

This removes all the "instruction only" lifetime variables to a struct that gets passed around as a reference to explain where the life time starts/ends

Chose `Meta` because it is unique, accurate , and doesn't add yet-another `instruction` name to the name space


layers/gpuav/spirv/descriptor_class_general_buffer_pass.cpp
```diff
@@ -345,14 +341,14 @@ bool DescriptorClassGeneralBufferPass::Instrument() {
                 // Do here before we inject instructions into the block list below
                 for (auto inst_it = block_instructions.begin(); inst_it != block_instructions.end(); ++inst_it) {
                     // Every instruction is analyzed by the specific pass and lets us know if we need to inject a function or not
-                    if (!RequiresInstrumentation(*function, *(inst_it->get()), true)) continue;
-                    Reset();
+                    if (!RequiresInstrumentation(*function, *(inst_it->get()), meta, true)) continue;
```
arno-lunarg, line 344: Looks like `meta` should be declared at this loop level?
=> if `not RequiresInstrumentation`, meta is not modified, and we directly explore the next instruction
=> if it does, `meta` is supposedly written in `RequiresInstrumentation` but then is immediately reset.

Well actually writing that I realize we just don't need `meta` here, it is a dummy object?

layers/gpuav/spirv/descriptor_class_general_buffer_pass.cpp
```diff
@@ -345,14 +341,14 @@ bool DescriptorClassGeneralBufferPass::Instrument() {
...
+                    meta.Reset();
                 }
             }
 
             for (auto inst_it = block_instructions.begin(); inst_it != block_instructions.end(); ++inst_it) {
                 // Every instruction is analyzed by the specific pass and lets us know if we need to inject a function or not
-                if (!RequiresInstrumentation(*function, *(inst_it->get()), false)) continue;
+                if (!RequiresInstrumentation(*function, *(inst_it->get()), meta, false)) continue;
```
arno-lunarg, line 351: Same, have `meta` be declared at the top of this loop, and then also no need for an explicit `Reset`

arno-lunarg (approved):
One thing regarding how `meta` is used needs to be fixed, but otherwise nice change, functional programming > oop

layers/gpuav/spirv/descriptor_class_general_buffer_pass.cpp
arno-lunarg, line 351: Actually you should just return a `meta` object, not pass it around like that. You could have `RequiresInstrumentation` return a `std::optional<InstrumentationMeta>`, that's more functional programming like thus more comprehensible

layers/gpuav/spirv/descriptor_class_general_buffer_pass.cpp
spencer-lunarg: it is just a dummy object

PR #9658, opened 2025-03-13
gpuav: Hash Post Processing once per block

spencer-lunarg:
While looking at traces, found one where we could reduce almost half the post processing instrumentation just doing a simple hash lookup on the 4 variables we were going to pass into the function (as it means they are all going to mark the same things)

**Before**
Vert Shader - PostProcessDescriptorIndexingPass instrumentation count: **127**
Geom Shader - PostProcessDescriptorIndexingPass instrumentation count: **8**
Frag Shader - PostProcessDescriptorIndexingPass instrumentation count: **57**

**After**
Vert Shader - PostProcessDescriptorIndexingPass instrumentation count: **78**
Geom Shader - PostProcessDescriptorIndexingPass instrumentation count: **2**
Frag Shader - PostProcessDescriptorIndexingPass instrumentation count: **30**

### DIFF
--- a/layers/gpuav/spirv/post_process_descriptor_indexing_pass.h
+++ b/layers/gpuav/spirv/post_process_descriptor_indexing_pass.h
@@ -40,7 +40,8 @@ class PostProcessDescriptorIndexingPass : public Pass {
         uint32_t variable_id = 0;
     };
 
-    bool RequiresInstrumentation(const Function& function, const Instruction& inst, InstructionMeta& meta);
+    bool RequiresInstrumentation(const Function& function, const Instruction& inst, InstructionMeta& meta,
+                                 vvl::unordered_set<uint32_t>& found_in_block_set);
     void CreateFunctionCall(BasicBlockIt block_it, InstructionIt* inst_it, const InstructionMeta& meta);
 
     uint32_t link_function_id = 0;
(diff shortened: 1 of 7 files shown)


layers/utils/hash_util.cpp
```diff
@@ -37,12 +37,12 @@ uint32_t VuidHash(std::string_view vuid) {
     return XXH32(vuid.data(), vuid.size(), seed);
 }
 
-uint32_t ShaderHash(const void *pCode, const size_t codeSize) {
```
spencer-lunarg, line 40: these both just took a `void*` and `size_t` and instead of creating a duplicate function with same name, just made a generic name

layers/gpuav/spirv/post_process_descriptor_indexing_pass.h
```diff
@@ -45,6 +45,9 @@ class PostProcessDescriptorIndexingPass : public Pass {
 
     uint32_t link_function_id = 0;
     uint32_t GetLinkFunctionId();
+
+    // We only need to instrument the set/binding/inde/variable combo once per block
+    vvl::unordered_set<uint32_t> found_in_block_set_;
```
spencer-lunarg, line 50: this doesn't have to be a member variable, can be on stack

arno-lunarg (approved):

PR #9673, opened 2025-03-13
chassis: Shared state tracking

jeremyg-lunarg:
Rather than each type of validation (eg. core checks, gpu-av) having its own state tracking via inheritance, they now share
a single state tracker, which is now dispatched separately just like all the other validation types. Validation that needs state
tracking now inherits from vvl::InstanceProxy and vvl::DeviceProxy, which provide the same interface for the most commonly used functionality like the Get methods.

The order in which the state tracker runs is important. For most commands, it runs before any validation types that use it. However, for Destroy and Free commands, it runs last so that the state object being destroyed still exists until the other active validation types run. This is implemented by changing the order in the dispatch vector, see dispatch_vector_generator.py and dispatch_vector.cpp.

The first commit in this PR renames vvl::Device to vvl::DeviceState for clarity and its best to review the 2 commits separately.

### DIFF
--- a/layers/chassis/chassis_manual.cpp
+++ b/layers/chassis/chassis_manual.cpp
@@ -450,7 +450,7 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateGraphicsPipelines(VkDevice device, VkPipeli
     bool skip = false;
     ErrorObject error_obj(vvl::Func::vkCreateGraphicsPipelines, VulkanTypedHandle(device, kVulkanObjectTypeDevice));
 
-    PipelineStates pipeline_states[LayerObjectTypeMaxEnum];
+    PipelineStates pipeline_states;
     chassis::CreateGraphicsPipelines chassis_state(pCreateInfos);
 
     {
@@ -461,8 +461,7 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateGraphicsPipelines(VkDevice device, VkPipeli
             }
             auto lock = vo->ReadLock();
             skip |= vo->PreCallValidateCreateGraphicsPipelines(device, pipelineCache, createInfoCount, pCreateInfos, pAllocator,
-                                                               pPipelines, error_obj, pipeline_states[vo->container_type],
-                                                               chassis_state);
+                                                               pPipelines, error_obj, pipeline_states, chassis_state);
             if (skip) return VK_ERROR_VALIDATION_FAILED_EXT;
         }
     }
@@ -476,7 +475,7 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateGraphicsPipelines(VkDevice device, VkPipeli
             }
             auto lock = vo->WriteLock();
             vo->PreCallRecordCreateGraphicsPipelines(device, pipelineCache, createInfoCount, pCreateInfos, pAllocator, pPipelines,
-                                                     record_obj, pipeline_states[vo->container_type], chassis_state);
+                                                     record_obj, pipeline_states, chassis_state);
         }
     }
 
@@ -503,7 +502,7 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateGraphicsPipelines(VkDevice device, VkPipeli
             }
             auto lock = vo->WriteLock();
             vo->PostCallRecordCreateGraphicsPipelines(device, pipelineCache, createInfoCount, pCreateInfos, pAllocator, pPipelines,
-                                                      record_obj, pipeline_states[vo->container_type], chassis_state);
+                                                      record_obj, pipeline_states, chassis_state);
         }
     }
     return result;
@@ -519,7 +518,7 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateComputePipelines(VkDevice device, VkPipelin
     bool skip = false;
     ErrorObject error_obj(vvl::Func::vkCreateComputePipelines, VulkanTypedHandle(device, kVulkanObjectTypeDevice));
 
-    PipelineStates pipeline_states[LayerObjectTypeMaxEnum];
+    PipelineStates pipeline_states;
     chassis::CreateComputePipelines chassis_state(pCreateInfos);
 
     {
@@ -530,8 +529,7 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateComputePipelines(VkDevice device, VkPipelin
             }
             auto lock = vo->ReadLock();
             skip |= vo->PreCallValidateCreateComputePipelines(device, pipelineCache, createInfoCount, pCreateInfos, pAllocator,
-                                                              pPipelines, error_obj, pipeline_states[vo->container_type],
-                                                              chassis_state);
+                                                              pPipelines, error_obj, pipeline_states, chassis_state);
             if (skip) return VK_ERROR_VALIDATION_FAILED_EXT;
         }
     }
@@ -545,7 +543,7 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateComputePipelines(VkDevice device, VkPipelin
             }
             auto lock = vo->WriteLock();
             vo->PreCallRecordCreateComputePipelines(device, pipelineCache, createInfoCount, pCreateInfos, pAllocator, pPipelines,
-                                                    record_obj, pipeline_states[vo->container_type], chassis_state);
+                                                    record_obj, pipeline_states, chassis_state);
         }
     }
 
@@ -571,7 +569,7 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateComputePipelines(VkDevice device, VkPipelin
             }
             auto lock = vo->WriteLock();
             vo->PostCallRecordCreateComputePipelines(device, pipelineCache, createInfoCount, pCreateInfos, pAllocator, pPipelines,
-                                                     record_obj, pipeline_states[vo->container_type], chassis_state);
+                                                     record_obj, pipeline_states, chassis_state);
         }
     }
     return result;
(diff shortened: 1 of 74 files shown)


layers/state_tracker/state_tracker.h
```diff
@@ -2013,8 +2057,108 @@ class Device : public vvl::base::Device {
         std::atomic<VkDeviceSize> free_{1U << 20};  // start at 1mb to leave room for a NULL address
     };
     FakeAllocator fake_memory;
+
+    std::map<LayerObjectTypeId, DeviceProxy*> proxies;
```
spencer-lunarg, line 2061: I think the future person looking at the code without knowledge of this PR will get here and if there was a comment explaining briefly what this PR does here I think it would be good to document it in the code.

Basically why this can't be an unordered_map, the order we create/destroy matters, etc

layers/state_tracker/state_tracker.h
```diff
@@ -2013,8 +2057,108 @@ class Device : public vvl::base::Device {
...
+    }
+    virtual bool VerifyImageLayout(const vvl::CommandBuffer& cb_state, const vvl::ImageView& image_view_state,
+                                   VkImageLayout explicit_layout, const Location& image_loc, const char* mismatch_layout_vuid,
+                                   bool* error) const {
+        return false;
+    }
 };
 
+template <typename State, std::enable_if_t<HasSubStates<State>::value, bool>>
+void DeviceState::NotifyCreated(State& state_object) {
```
spencer-lunarg, line 2156: ok with this name, but thoughts on `NotifyProxies` and `item.second->CreateProxy()`

scripts/generators/dispatch_vector_generator.py
```diff
@@ -25,92 +25,7 @@
...
-            default:
-                /* Chassis codegen needs to be updated for unknown validation object type */
-                assert(0);
-            }
-        }
-    };
-
-    intercept_vectors.resize(InterceptIdCount);
-'''
+from generators.dispatch_object_generator import APISpecific
```
spencer-lunarg, line 28: for now this is fine, but hate the idea that the generators are not isolated

scripts/generators/dispatch_vector_generator.py
```diff
@@ -219,17 +134,96 @@ def generateSource(self):
...
+    };
+
+    intercept_vectors.resize(InterceptIdCount);
+    ''')
+
 
         guard_helper = PlatformGuardHelper()
         for command in [x for x in self.vk.commands.values() if not x.instance and x.name not in self.skip_intercept_id_functions]:
             out.extend(guard_helper.add_guard(command.protect))
+            macro = 'BUILD_DESTROY_DISPATCH_VECTOR' if ('Destroy' in command.name or 'Free' in command.name) else 'BUILD_DISPATCH_VECTOR'
```
spencer-lunarg, line 220: we have this issue detecting these in threading scripts

```python
        if ('Destroy' in command.name or 'Free' in command.name or 'ReleasePerformanceConfigurationINTEL' in command.name) and prefix == 'Finish':
            out.append(f'DestroyObject{parent_instance}({param.name});\n')
elif param.pointer and ('Create' in command.name or 'Allocate' in command.name or 'AcquirePerformanceConfigurationINTEL' in command.name) and prefix == 'Finish':
```

I "think" we are find because we dont state track the Release/Acquire things, but that stuff feels very under tested as it is

layers/state_tracker/state_object.h
```diff
@@ -149,6 +149,8 @@ class RefcountedStateObject : public StateObject {
 template <typename T>
 class SubStateManager {
   public:
+    using SubStateType = T;
```
spencer-lunarg, line 152: could compile without this, was this left over?

layers/state_tracker/state_tracker.h
```diff
@@ -121,6 +121,13 @@ struct TraitsBase {
     using ReadLockedType = LockedSharedPtr<const StateType, ReadLockGuard>;
     using WriteLockedType = LockedSharedPtr<StateType, WriteLockGuard>;
 };
+
+template <typename T, typename = void>
```
spencer-lunarg, line 125: could compile without this, was this left over?

layers/state_tracker/state_tracker.h
```diff
@@ -474,16 +529,19 @@ class Device : public vvl::base::Device {
...
           has_format_feature2(dev->stateless_device_data.has_format_feature2),
           has_robust_image_access(dev->stateless_device_data.has_robust_image_access),
           has_robust_image_access2(dev->stateless_device_data.has_robust_image_access2),
           has_robust_buffer_access2(dev->stateless_device_data.has_robust_buffer_access2) {
         physical_device_state = instance_state->Get<vvl::PhysicalDevice>(physical_device).get();
     }
-    ~Device();
+    ~DeviceState();
+
+    void AddProxy(DeviceProxy* proxy);
```
spencer-lunarg, line 543: reason they cant be a reference instead of a pointer?

layers/state_tracker/state_tracker.h
```diff
@@ -2013,8 +2057,108 @@ class Device : public vvl::base::Device {
...
+    WriteLockedType GetWrite(typename Traits::HandleType handle) {
+        return device_state->GetWrite<State>(handle);
+    }
+
+    template <typename State>
+    size_t Count() const {
+        return device_state->Count<State>();
+    }
+    template <typename State>
+    bool AnyOf(std::function<bool(const State& s)> fn) const {
```
spencer-lunarg, line 2110: who consumes this?

layers/gpuav/instrumentation/gpuav_shader_instrumentor.cpp
```diff
@@ -1247,6 +1247,7 @@ void GpuShaderInstrumentor::InternalError(LogObjectList objlist, const Location
 
     LogError(vuid, objlist, loc, "Internal Error, %s is being disabled. Details:\n%s", layer_name, error_message.c_str());
 
+    device_state->RemoveProxy(this);
```
spencer-lunarg, line 1250: this is error prone doing it in here, you missed it in `InternalVmaError`... there a way we can do it in ReleaseValidationObject?

layers/chassis/chassis_manual.cpp
```diff
@@ -632,7 +630,7 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateRayTracingPipelinesKHR(VkDevice device, VkD
     bool skip = false;
     ErrorObject error_obj(vvl::Func::vkCreateRayTracingPipelinesKHR, VulkanTypedHandle(device, kVulkanObjectTypeDevice));
 
-    PipelineStates pipeline_states[LayerObjectTypeMaxEnum];
```
spencer-lunarg, line 635: :100:

layers/state_tracker/state_tracker.cpp
```diff
@@ -1712,29 +1735,23 @@ bool Device::PreCallValidateCreateGraphicsPipelines(VkDevice device, VkPipelineC
             // The rasterization_enabled is our way to hint to vvl::RenderPass to ignore a possible VkPipelineRenderingCreateInfo
             // that contains bad pointers (when using GPL)
             const bool has_fragment_output_state =
-                Pipeline::ContainsSubState(this, create_info, VK_GRAPHICS_PIPELINE_LIBRARY_FRAGMENT_OUTPUT_INTERFACE_BIT_EXT);
+                Pipeline::ContainsSubState(*this, create_info, VK_GRAPHICS_PIPELINE_LIBRARY_FRAGMENT_OUTPUT_INTERFACE_BIT_EXT);
             const bool rasterization_enabled =
                 has_fragment_output_state && Pipeline::EnablesRasterizationStates(*this, create_info);
 
             render_pass = std::make_shared<RenderPass>(pipeline_rendering_ci, rasterization_enabled);
-        } else {
```
spencer-lunarg, line 1720: why was this removed?

spencer-lunarg (approved):
questions and nits, but overall cool to see this and LGTM

layers/state_tracker/state_tracker.h
jeremyg-lunarg: good point

layers/state_tracker/state_object.h
jeremyg-lunarg: yep will remove

layers/vulkan/generated/dispatch_vector.cpp
```diff
@@ -51,44 +48,73 @@ void Device::InitObjectDispatchVectors() {
...
+
+                    case LayerObjectTypeGpuAssisted:
+                        if (ga_typeid != vo_typeid) intercept_vector->push_back(item);
+                        break;
+
+                    case LayerObjectTypeSyncValidation:
+                        if (sv_typeid != vo_typeid) intercept_vector->push_back(item);
+                        break;
+
+                    default:
```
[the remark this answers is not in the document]
jeremyg-lunarg: I think so...

layers/state_tracker/state_tracker.h
jeremyg-lunarg: `layers/core_checks/cc_wsi.cpp:        const bool is_supported = AnyOf<vvl::Queue>([this, surface_state](const vvl::Queue &queue_state) {`

layers/state_tracker/state_tracker.h
```diff
@@ -1996,6 +2027,13 @@ class DeviceState : public vvl::base::Device {
     VALSTATETRACK_MAP_AND_TRAITS(VkIndirectExecutionSetEXT, vvl::IndirectExecutionSet, indirect_execution_set_ext_map_)
     VALSTATETRACK_MAP_AND_TRAITS(VkIndirectCommandsLayoutEXT, vvl::IndirectCommandsLayout, indirect_commands_layout_ext_map_)
 
+    // For state objects that have sub states.  Requires the definition of DeviceProxy, which is below.
+    template <typename State, std::enable_if_t<HasSubStates<State>::value, bool> = true>
+    void NotifyCreated(State& state_object);
+
+    template <typename State, std::enable_if_t<not HasSubStates<State>::value, bool> = true>
```
y-novikov, line 2034: This fails to build with MSVC when we roll into ANGLE:
https://chromium-review.googlesource.com/c/angle/angle/+/6425459
https://ci.chromium.org/ui/p/angle/builders/try/win-msvc-x86-dbg-compile/23700/overview

PR #9688, opened 2025-03-16
layer: Improve YCbCr error messages

spencer-lunarg:
Provide better error messages around YCbCr


layers/stateless/sl_descriptor.cpp
```diff
@@ -1055,31 +1060,50 @@ bool Device::manual_PreCallValidateCreateSamplerYcbcrConversion(VkDevice device,
         // depth/stencil = no [g,b,a] (shouldn't ever occur, but no VU preventing it)
         const uint32_t component_count = (vkuFormatIsDepthOrStencil(format) == true) ? 1 : vkuFormatComponentCount(format);
 
-        if ((component_count < 4) && ((components.r == VK_COMPONENT_SWIZZLE_A) || (components.g == VK_COMPONENT_SWIZZLE_A) ||
-                                      (components.b == VK_COMPONENT_SWIZZLE_A))) {
+        if (format == VK_FORMAT_UNDEFINED) {
+            // If using external format they should have been caught in GetExternalFormat above. This means the user forgot to pass
```
spencer-lunarg, line 1064: This the main thing added, it wasn't obvious for a user this error occurred because they forgot to pass in the `VkExternalFormatANDROID` and we were checking `VK_FORMAT_UNDEFINED` here which is nonsensical

layers/stateless/sl_descriptor.cpp
```diff
@@ -1034,17 +1034,22 @@ bool Device::manual_PreCallValidateCreateSamplerYcbcrConversion(VkDevice device,
...
-                LogError(vuid, device, create_info_loc,
-                         "The ycbcrModel (%s) is not VK_SAMPLER_YCBCR_MODEL_CONVERSION_RGB_IDENTITY so components.r (%s), "
-                         "components.g (%s), nor components.b (%s) can't be VK_COMPONENT_SWIZZLE_ZERO or VK_COMPONENT_SWIZZLE_ONE.",
-                         string_VkSamplerYcbcrModelConversion(pCreateInfo->ycbcrModel), string_VkComponentSwizzle(components.r),
-                         string_VkComponentSwizzle(components.g), string_VkComponentSwizzle(components.b));
+        if (components.r == VK_COMPONENT_SWIZZLE_ZERO || components.g == VK_COMPONENT_SWIZZLE_ZERO ||
+            components.b == VK_COMPONENT_SWIZZLE_ZERO) {
+            skip |= LogError(vuid, device, create_info_loc,
+                             "The ycbcrModel (%s) is not VK_SAMPLER_YCBCR_MODEL_CONVERSION_RGB_IDENTITY so the r, g, and b in "
+                             "pCreateInfo->components all can't be VK_COMPONENT_SWIZZLE_ZERO\n%s",
```
artem-lunarg, line 1043: `all can't be` - mostly my english, does this mean that none of the components can be SWIZZLE_ZERO (as opposed to all of the components can't be SWIZZLE_ZERO at the same time)?

artem-lunarg (approved):

layers/stateless/sl_descriptor.cpp
ziga-lunarg: I would suggest `all can't` => `can't all`

spencer-lunarg: `can't all` :100:

PR #9689, opened 2025-03-16
gpuav: Use EntryPoint to map shader to post process

spencer-lunarg:
The goal is to remove the `BindingVariableMap` when doing Post Processing because

1. It is the wrong level of detail to save for the given pipeline
2. It doesn't work at all with Shader Object (main reason we can't support it currently)
3. It reduces a lot of random checks because we now can directly get the `const spirv::ResourceInterfaceVariable &` right away

### DIFF
--- a/layers/gpuav/descriptor_validation/gpuav_descriptor_set.cpp
+++ b/layers/gpuav/descriptor_validation/gpuav_descriptor_set.cpp
@@ -22,6 +22,7 @@
 #include "gpuav/resources/gpuav_shader_resources.h"
 #include "gpuav/shaders/gpuav_shaders_constants.h"
 #include "state_tracker/descriptor_sets.h"
+#include "state_tracker/shader_module.h"
 
 #include "profiling/profiling.h"
 
@@ -306,7 +307,8 @@ VkDeviceAddress DescriptorSet::GetPostProcessBuffer(Validator &gpuav, const Loca
 
 // cross checks the two buffers (our layout with the output from the GPU-AV run) and builds a map of which indexes in which binding
 // where accessed
-std::vector<DescriptorAccess> DescriptorSet::GetDescriptorAccesses(const Location &loc, uint32_t shader_set) const {
+std::vector<DescriptorAccess> DescriptorSet::GetDescriptorAccesses(
+    const Location &loc, uint32_t shader_set, const std::vector<const ::spirv::EntryPoint *> &entry_points) const {
     VVL_ZoneScoped;
     std::vector<DescriptorAccess> descriptor_accesses;
     if (post_process_buffer_.IsDestroyed()) {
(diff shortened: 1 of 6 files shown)


layers/gpuav/descriptor_validation/gpuav_descriptor_set.cpp
```diff
@@ -322,7 +324,14 @@ std::vector<DescriptorAccess> DescriptorSet::GetDescriptorAccesses(const Locatio
             const glsl::PostProcessDescriptorIndexSlot slot = slot_ptr[binding_layout.start + descriptor_i];
             if (slot.descriptor_set & glsl::kDescriptorSetAccessedMask) {
                 if ((slot.descriptor_set & glsl::kDescriptorSetSelectionMask) == shader_set) {
-                    descriptor_accesses.emplace_back(DescriptorAccess{binding, descriptor_i, slot.variable_id});
+                    // TODO - Currently we don't save the shader stage in the shader so we rely on resource OpVariable IDs not being
```
spencer-lunarg, line 327: we currently have this problem already, but there was no way to solve it before, now we can check for it (but currently not main worry)

layers/gpuav/descriptor_validation/gpuav_descriptor_validation.cpp
```diff
@@ -250,33 +234,18 @@ void UpdateBoundDescriptors(Validator &gpuav, CommandBuffer &cb_state, VkPipelin
...
                                              action_command_snapshot.shader_handle, draw_loc);
 
-            auto descriptor_accesses = bound_descriptor_set->GetDescriptorAccesses(loc, set_index);
+            std::vector<DescriptorAccess> descriptor_accesses =
+                bound_descriptor_set->GetDescriptorAccesses(loc, set_index, action_command_snapshot.entry_points);
             for (const auto &descriptor_access : descriptor_accesses) {
                 auto descriptor_binding = bound_descriptor_set->GetBinding(descriptor_access.binding);
                 ASSERT_AND_CONTINUE(descriptor_binding);
 
-                // There is a chance two descriptor bindings are aliased to each other.
```
spencer-lunarg, line 258: the main goal was to remove all this junk (and more above) as it adds more complexity to what is all going on here

layers/gpuav/descriptor_validation/gpuav_descriptor_validation.cpp
```diff
@@ -250,33 +234,18 @@ void UpdateBoundDescriptors(Validator &gpuav, CommandBuffer &cb_state, VkPipelin
...
-                //   layout(set = 0, binding = 2) uniform sampler3D tex3d[];
-                //   layout(set = 0, binding = 2) uniform sampler2D tex[];
```
spencer-lunarg, line 260: Now that we just do the lookup from the `OpVariable` ID, the alias case like this automatically get solved because this information is already known from the shader

arno-lunarg (approved):

PR #9690, opened 2025-03-17
layers: Cleanup of error messages

spencer-lunarg:


layers/core_checks/cc_device.cpp
```diff
@@ -286,7 +286,7 @@ bool core::Instance::ValidateDeviceQueueCreateInfos(const vvl::PhysicalDevice &p
                     extensions.vk_khr_get_physical_device_properties2 ? " or vkGetPhysicalDeviceQueueFamilyProperties2[KHR]" : "";
                 const std::string count_note =
                     queue_family_has_props
-                        ? "i.e. is not less than or equal to " + std::to_string(requested_queue_family_props.queueCount)
+                        ? "needs to be less than or equal to " + std::to_string(requested_queue_family_props.queueCount)
```
artem-lunarg, line 289: `queueFamilyIndex}` - is right brace here needed?

spencer-lunarg: aww, I see below, no let me remove that as well!

artem-lunarg (approved):

PR #9693, opened 2025-03-17
layers: Multiview with Mesh Shader in GLSL

spencer-lunarg:
Discussed in https://gitlab.khronos.org/vulkan/vulkan/-/issues/4194

This shows how you can use Mesh Shading with Multiview in GLSL and not trigger the Layer VU

### DIFF
--- a/layers/core_checks/cc_pipeline_graphics.cpp
+++ b/layers/core_checks/cc_pipeline_graphics.cpp
@@ -4154,15 +4154,16 @@ bool CoreChecks::ValidateMultiViewShaders(const vvl::Pipeline &pipeline, const L
         // Stage may not have SPIR-V data (e.g. due to the use of shader module identifier or in Vulkan SC)
         if (!stage.spirv_state) continue;
 
-        // This is being discussed in https://gitlab.khronos.org/vulkan/vulkan/-/issues/4194
-        // As a temporary solution, ignore this case to prevent false positives.
-        if (stage.GetStage() == VK_SHADER_STAGE_MESH_BIT_EXT) continue;
-
         if (stage.spirv_state->static_data_.has_builtin_layer) {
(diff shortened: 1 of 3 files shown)


layers/core_checks/cc_pipeline_graphics.cpp
```diff
@@ -4154,15 +4154,16 @@ bool CoreChecks::ValidateMultiViewShaders(const vvl::Pipeline &pipeline, const L
...
+            // Special case for GLSL and Mesh Shading discussed in https://gitlab.khronos.org/vulkan/vulkan/-/issues/4194
             const char *vuid = dynamic_rendering ? "VUID-VkGraphicsPipelineCreateInfo-renderPass-06059"
                                                  : "VUID-VkGraphicsPipelineCreateInfo-renderPass-06050";
-            skip |= LogError(vuid, device, multiview_loc, "is 0x%" PRIx32 " but %s stage contains a Layer decorated OpVariable.",
-                             view_mask, string_VkShaderStageFlagBits(stage.GetStage()));
+            skip |= LogError(vuid, device, multiview_loc, "is 0x%" PRIx32 " but %s stage contains a Layer decorated OpVariable.%s",
+                             view_mask, string_VkShaderStageFlagBits(stage.GetStage()),
+                             stage.GetStage() == VK_SHADER_STAGE_MESH_BIT_EXT
+                                 ? "(If hitting this with Mesh Shading using GLSL you can explicitly leave out Layer, see "
+                                   "https://godbolt.org/z/av9zsxT8G as an example)"
```
artem-lunarg, line 4165: that's interesting that godbolt becomes part of infrustructure, in our source code I can also see xxhash does this too.

artem-lunarg (approved):

layers/core_checks/cc_pipeline_graphics.cpp
spencer-lunarg: ya, well it is the best way to show case things like this honestly lol

PR #9696, opened 2025-03-17
scripts: Fix for Version1.0 in VulkanObject

spencer-lunarg:
@aqnuep @charles-lunarg the fix for VVL was not bad at all

This change with https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/7235 applied produce the same code

@aqnuep please let me know which part of these would need to be put into something `APISpecific`


aqnuep:
Please let's not add more hard-coded checks of `== 'VK_VERSION_1_0`. We'd have to fork this.

I think we should just leave version to be `None` for the initial version. This causes more problems than it solves, unfortunately.

spencer-lunarg:
closing as discussed

Issue #9587, opened 2025-03-05
False positive for destroying a semaphohre that should no longer be in use


knokko:
**Environment:**
 - OS: PopOS 22.04
 - GPU and driver version: NVIDIA Quadro M1200 565.77.0.0 (but same error happens on my integrated GPU & CPU)
 - SDK or header version if building from repo: 1.4.304
 - Options enabled (synchronization, best practices, etc.): none

**Describe the Issue**

When I call `vkDestroySemaphore` on a semaphore that I believe is no longer in use, I get the following validation error:
```
Validation Error: [ VUID-vkDestroySemaphore-semaphore-05149 ] | MessageID = 0x93e24db1 | vkDestroySemaphore(): can't be called on VkSemaphore 0xdd3a8a0000000015[PresentSemaphore-HelloTriangle-16071629517007] that is currently in use by VkQueue 0x79fb752d5570[].
The Vulkan spec states: All submitted batches that refer to semaphore must have completed execution (https://vulkan.lunarg.com/doc/view/1.4.304.0/linux/1.4-extensions/vkspec.html#VUID-vkDestroySemaphore-semaphore-05149)
```

I believe that this semaphore `PresentSemaphore-HelloTriangle-16071629517007` is no longer in use because of the following trace. It is created here:
```
Thread 0, Frame 0:
vkCreateSemaphore(device, pCreateInfo, pAllocator, pSemaphore) returns VkResult VK_SUCCESS (0):
    device:                         VkDevice = 0x79fb751e6620
    pCreateInfo:                    const VkSemaphoreCreateInfo* = 0x79fb742ee3f0:
        sType:                          VkStructureType = VK_STRUCTURE_TYPE_SEMAPHORE_CREATE_INFO (9)
        pNext:                          const void* = NULL
        flags:                          VkSemaphoreCreateFlags = 0
    pAllocator:                     const VkAllocationCallbacks* = NULL
    pSemaphore:                     VkSemaphore* = 0xdd3a8a0000000015

Thread 0, Frame 0:
vkSetDebugUtilsObjectNameEXT(device, pNameInfo) returns VkResult VK_SUCCESS (0):
    device:                         VkDevice = 0x79fb751e6620
    pNameInfo:                      const VkDebugUtilsObjectNameInfoEXT* = 0x79fb7b3fe300:
[...]

PR #9698, opened 2025-03-17
layers: Fix present semaphore in use error after acquire fence wait

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9587

This adds one more use case of synchronization and simplifies implementation a bit. The new use case helped to realize it's possible to use simpler implementation that unifies both previous and new sync scenario  (assuming no regression, which is hard to guarantee for such code). Some code simplifications:
* Remove `AcquireFenceSync` type. Now it can be replaced with a single instance of `SubmissionReference`
* Remove `Semaphore::GetPendingBinaryWaitSubmission`

### DIFF
--- a/layers/state_tracker/fence_state.cpp
+++ b/layers/state_tracker/fence_state.cpp
@@ -25,14 +25,14 @@ static VkExternalFenceHandleTypeFlags GetExportHandleTypes(const VkFenceCreateIn
     return export_info ? export_info->handleTypes : 0;
 }
 
-vvl::Fence::Fence(vvl::Device &dev, VkFence handle, const VkFenceCreateInfo *pCreateInfo)
+vvl::Fence::Fence(Logger &logger, VkFence handle, const VkFenceCreateInfo *pCreateInfo)
     : RefcountedStateObject(handle, kVulkanObjectTypeFence),
       flags(pCreateInfo->flags),
       export_handle_types(GetExportHandleTypes(pCreateInfo)),
       state_((pCreateInfo->flags & VK_FENCE_CREATE_SIGNALED_BIT) ? kRetired : kUnsignaled),
       completed_(),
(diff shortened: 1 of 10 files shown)


layers/state_tracker/state_tracker.cpp
```diff
@@ -3607,12 +3607,8 @@ void Device::PostCallRecordQueuePresentKHR(VkQueue queue, const VkPresentInfoKHR
         }
     }
 
-    AcquireFenceSync acquire_fence_sync;
     for (uint32_t i = 0; i < pPresentInfo->waitSemaphoreCount; ++i) {
         if (auto semaphore_state = Get<Semaphore>(pPresentInfo->pWaitSemaphores[i])) {
-            if (auto submission_ref = semaphore_state->GetPendingBinarySignalSubmission()) {
```
artem-lunarg, line 3613: The main fix is in this function. Instead of tracking multiple submissions associated with each present wait semaphore, now we track only a single submission associated with presentation operation itself. This makes things simpler (many -> one) and also fixes the issue that present semaphore was reported as in-use. The reason it fixes the issue is because when we wait on Image Acquire Fence the presentation submission is automatically picked up by our pipeline which marks semaphores as not in use.

layers/state_tracker/fence_state.cpp
```diff
@@ -25,14 +25,14 @@ static VkExternalFenceHandleTypeFlags GetExportHandleTypes(const VkFenceCreateIn
     return export_info ? export_info->handleTypes : 0;
 }
 
-vvl::Fence::Fence(vvl::Device &dev, VkFence handle, const VkFenceCreateInfo *pCreateInfo)
```
spencer-lunarg, line 28: :100:

Issue #9699, opened 2025-03-17
Detect failed instrumented shader and create original shader as fallback


spencer-lunarg:
We will have cases where after instrumenting a shader we will cause a `vkCreate*Pipeline` to fail and we should detect that and replace the shader with the original and allow the app to move forward

PR #9702, opened 2025-03-17
gpuav: Retry Pipeline if failed with modification

spencer-lunarg:
Found Doom on RADV doesn't like our changes to the RTX shader (and blows up), so added logic to check if pipelines fail and to try and re-run them so things are a bit more robust

closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9699

### DIFF
--- a/layers/chassis/chassis_manual.cpp
+++ b/layers/chassis/chassis_manual.cpp
@@ -20,9 +20,7 @@
  ****************************************************************************/
 #include "chassis.h"
 
-#include <array>
 #include <cstring>
-#include <mutex>
 
 #include "chassis/dispatch_object.h"
 #include "chassis/validation_object.h"
@@ -485,6 +483,13 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateGraphicsPipelines(VkDevice device, VkPipeli
         VVL_ZoneScopedN("Dispatch_CreateGraphicsPipelines");
         result = device_dispatch->CreateGraphicsPipelines(device, pipelineCache, createInfoCount, chassis_state.pCreateInfos,
                                                           pAllocator, pPipelines);
+
+        // If we have modified the pCreateInfos caused things to fail, revert to allow the app to continue
+        if (result < VK_SUCCESS && chassis_state.is_modified) {
+            chassis_state.is_modified = false;
+            result = device_dispatch->CreateGraphicsPipelines(device, pipelineCache, createInfoCount, pCreateInfos, pAllocator,
+                                                              pPipelines);
+        }
     }
     record_obj.result = result;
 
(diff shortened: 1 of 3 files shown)


layers/chassis/chassis_manual.cpp
```diff
@@ -485,6 +483,13 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateGraphicsPipelines(VkDevice device, VkPipeli
         VVL_ZoneScopedN("Dispatch_CreateGraphicsPipelines");
         result = device_dispatch->CreateGraphicsPipelines(device, pipelineCache, createInfoCount, chassis_state.pCreateInfos,
                                                           pAllocator, pPipelines);
+
+        // If we have modified the pCreateInfos caused things to fail, revert to allow the app to continue
+        if (result < VK_SUCCESS && chassis_state.is_modified) {
```
arno-lunarg, line 488:
```suggestion
        if (result != VK_SUCCESS && chassis_state.is_modified) {
```
?

layers/chassis/chassis_modification_state.h
```diff
@@ -63,6 +63,8 @@ struct ShaderObject {
     std::vector<std::shared_ptr<spirv::Module>> module_states;  // contains SPIR-V to validate
     std::vector<spirv::StatelessData> stateless_data;
 
+    // When using GPU-AV the pCreateInfo is modified on the user
+    bool is_modified = false;
```
arno-lunarg, line 67: Kind of vague, we could go `was_instrumented`?

layers/gpuav/instrumentation/gpuav_shader_instrumentor.cpp
```diff
@@ -560,6 +568,9 @@ void GpuShaderInstrumentor::PostCallRecordCreateGraphicsPipelines(VkDevice devic
                                                                   chassis::CreateGraphicsPipelines &chassis_state) {
     BaseClass::PostCallRecordCreateGraphicsPipelines(device, pipelineCache, count, pCreateInfos, pAllocator, pPipelines, record_obj,
                                                      pipeline_states, chassis_state);
+    // This can occur if the driver failed to compile the instrumented shader or if a PreCall step failed
+    if (!chassis_state.is_modified) return;
```
arno-lunarg, line 572: This check relies on supplied parameters and is not as generic as `if (!gpuav_settings.IsSpirvModified()) return;`, so I would put it after. Elsewhere we tend to check for settings before looking at objects.

arno-lunarg (approved):

layers/chassis/chassis_modification_state.h
spencer-lunarg: thought about it, but the real thing it is check is if we modified the create info in the `modified_create_info` member... so seemed "modified" fit with those concepts (in theory, something else in the future can modify the create info as well)

layers/gpuav/instrumentation/gpuav_shader_instrumentor.cpp
spencer-lunarg: yep, will put lower

layers/chassis/chassis_manual.cpp
spencer-lunarg: added to the comment

PR #9704, opened 2025-03-18
layers: Readd removed state creation abstractions

aqnuep:
The whole point of these abstractions is to enable subclasses to provide customized creation to customized subclasses of the state objects themselves, so it would be bad design to remove this ability. For example, this is not only used by layer subclasses such as the best practices layers, but also by the Vulkan SC validation layers to ammend SC-specific data to the state objects through subclassing.

Having overrideable factory methods for the state objects is paramount to enable extensibility.

### DIFF
--- a/layers/state_tracker/state_tracker.cpp
+++ b/layers/state_tracker/state_tracker.cpp
@@ -1089,8 +1089,8 @@ void Device::PostCallRecordAllocateMemory(VkDevice device, const VkMemoryAllocat
         WriteLockGuard guard(fd_handle_map_lock_);
         fd_handle_map_.erase(import_memory_fd_info->fd);
     }
-    Add(std::make_shared<DeviceMemory>(*pMemory, pAllocateInfo, fake_address, memory_type, memory_heap,
-                                       std::move(dedicated_binding), physical_device_count));
+    Add(CreateDeviceMemoryState(*pMemory, pAllocateInfo, fake_address, memory_type, memory_heap, std::move(dedicated_binding),
+                                physical_device_count));
     return;
 }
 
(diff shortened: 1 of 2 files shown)


spencer-lunarg (approved):
@jeremyg-lunarg did you have a reason we shouldn't add this back?

aqnuep:
I think @jeremyg-lunarg's thought was that since best-practices no longer going to subclass state tracking once it moves outside of the validation layer class hierarchy, then nobody will subclass it, but Vulkan SC does have to subclass state tracking (and various state objects) in order to be able to extend those with SC-specific data/behavior, that's why we would need these overridable factory methods to continue to exist.

jeremyg-lunarg:
Yeah sorry about that. There's a bigger problem coming with the shared state tracker: It is not possible in C++ for multiple types of validation to subclass state objects via inheritance at the same time. https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9651 is how I want to solve this by adding 'SubState' objects for validation-type specific data.  I've only added the SubState functionality to places where code in this repo needs it. I'll take a look at the SC code and add it where that code is overriding as well.  A follow on PR https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9673 is where the state tracker actually becomes shared. The `Create` callbacks change in that PR so that all state tracker clients have a chance to attach substate.

jeremyg-lunarg (approved):

jeremyg-lunarg:
> Yeah sorry about that. There's a bigger problem coming with the shared state tracker: It is not possible in C++ for multiple types of validation to subclass state objects via inheritance at the same time. #9651 is how I want to solve this by adding 'SubState' objects for validation-type specific data. I've only added the SubState functionality to places where code in this repo needs it. I'll take a look at the SC code and add it where that code is overriding as well. A follow on PR #9673 is where the state tracker actually becomes shared. The `Create` callbacks change in that PR so that all state tracker clients have a chance to attach substate.

Actually, looking at the SC code it seems like it would be better to let it keep overriding the `Create` methods even in the shared state tracker. I'll adjust #9651 to account for that.

aqnuep:
Right, @jeremyg-lunarg, we don't want to subclass state objects for individual validation layers. I understand that won't work with the shared state tracker. We expect to only subclass the shared state tracker itself for the SC version of state tracking, as today, so that won't be an issue.

Issue #9703, opened 2025-03-18
[Android] Support 16 KB page sizes


jsyjbaba:
Beginning with Android 15, AOSP supports devices that are configured to use a page size of 16 KB (16 KB devices). If your app uses any [NDK](https://developer.android.com/ndk) libraries, either directly or indirectly through an SDK, then you will need to rebuild your app for it to work on these 16 KB devices.

As device manufacturers continue to build devices with larger amounts of physical memory (RAM), many of these devices will adopt 16 KB (and eventually greater) page sizes to optimize the device's performance. Adding support for 16 KB page size devices enables your app to run on these devices and helps your app benefit from the associated performance improvements. Without recompiling, apps might not work on 16 KB devices when they are productionized in future Android releases.

https://developer.android.com/guide/practices/page-sizes

The .so file built for 4KB does not work properly on a 16KB device. However, the Vulkan validation .so built for 16KB works on both 4KB and 16KB devices. Therefore, the Vulkan validation layer (.so) should be built for 16KB.

Update the CMakeLists.txt to enable 16KB ELF alignment.
<img width="1173" alt="Image" src="https://github.com/user-attachments/assets/989b2660-76dc-4427-9f9a-b877935dfe36" />

As I checked this way is working fine,
Before(Using NDK25)
<img width="680" alt="Image" src="https://github.com/user-attachments/assets/eb5794a3-4ef7-4421-9fc8-b3dfe9908406" />

After (Using NDK25) - CMakeLists.txt modified
<img width="683" alt="Image" src="https://github.com/user-attachments/assets/660a2da9-7b82-4b0e-b61c-5828b1bee281" />

Issue #9170, opened 2025-01-08
Android binaries don't support 16KiB page sizes


alusch:
Starting with Android 15, devices may use 16KiB pages instead of 4KiB, which requires native libraries to be recompiled to align the ELF sections to 16KiB boundaries (only on 64-bit architectures). https://developer.android.com/guide/practices/page-sizes has more details.

The latest binary release is only aligned to 4KiB (2**12) and so will fail to load on such devices:
```
> objdump -p android-binaries-1.3.296.0/arm64-v8a/libVkLayer_khronos_validation.so | grep LOAD
    LOAD off    0x0000000000000000 vaddr 0x0000000000000000 paddr 0x0000000000000000 align 2**12
    LOAD off    0x00000000013a9e90 vaddr 0x00000000013aae90 paddr 0x00000000013aae90 align 2**12
    LOAD off    0x00000000014e8b90 vaddr 0x00000000014eab90 paddr 0x00000000014eab90 align 2**12

> objdump -p android-binaries-1.3.296.0/x86_64/libVkLayer_khronos_validation.so | grep LOAD
    LOAD off    0x0000000000000000 vaddr 0x0000000000000000 paddr 0x0000000000000000 align 2**12
    LOAD off    0x00000000015a3510 vaddr 0x00000000015a4510 paddr 0x00000000015a4510 align 2**12
    LOAD off    0x00000000016e1410 vaddr 0x00000000016e3410 paddr 0x00000000016e3410 align 2**12
```

Per the linked document above, since you're using CMake and NDK 25, you should set

```cmake
target_link_options(${CMAKE_PROJECT_NAME} PRIVATE "-Wl,-z,max-page-size=16384")
```

for 64-bit builds (and depending on the version of `ld` may also need to set `common-page-size=16384`).

Not the most urgent issue since the validation layers are only used for testing and such devices probably don't exist yet, but would be good to be prepared for them.

spencer-lunarg:
@alusch thanks for this, good to know and we can get this fix in for future devices

webez:
@spencer-lunarg Any ETA on this? We want to start using 16KiB pages and it would be great if validation layers binaries already have the support

spencer-lunarg:
@webez I am a bit behind on things (said every engineer ever)

[...]

PR #9705, opened 2025-03-18
build: Use 16k pages in Android

spencer-lunarg:
should close https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9703 and https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9170

@jsyjbaba and @alusch @webez can you help verify this fixes the problem (my issue is I can't reproduce it currently and seems it is a "simple" fix)

### DIFF
--- a/layers/CMakeLists.txt
+++ b/layers/CMakeLists.txt
@@ -424,7 +424,7 @@ elseif(ANDROID)
     # Need 16k pages in Android 15
     # https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9703
     # https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9170
-    target_link_options(${CMAKE_PROJECT_NAME} PRIVATE "-Wl,-z,max-page-size=16384")
+    target_link_options(vvl PRIVATE "-Wl,-z,max-page-size=16384")
     
     target_link_options(vvl PRIVATE LINKER:--version-script=${CMAKE_CURRENT_SOURCE_DIR}/${LAYER_NAME}-android.map)
 else()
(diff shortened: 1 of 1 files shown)


layers/CMakeLists.txt
```diff
@@ -421,6 +421,11 @@ endif()
     # Both Apple and IOS
     target_link_options(vvl PRIVATE -exported_symbols_list ${CMAKE_CURRENT_SOURCE_DIR}/${LAYER_NAME}.exp)
 elseif(ANDROID)
+    # Need 16k pages in Android 15
+    # https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9703
+    # https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9170
+    target_link_options(${CMAKE_PROJECT_NAME} PRIVATE "-Wl,-z,max-page-size=16384")
```
spencer-lunarg, line 427:
```suggestion
    target_link_options(vvl PRIVATE "-Wl,-z,max-page-size=16384")
```

jsyjbaba:
Yes, it looks like the issue should be resolved with the provided fix.

artem-lunarg (approved):
LGTM

Aksing AI about potential issues I got this, not sure if this can affects us

> Compatibility with Older Android Versions: Binaries built with larger page sizes may face compatibility issues on devices running older Android versions that expect smaller page sizes. For example, certain unwinder implementations used by crash handlers on versions of Android prior to API 29 did not correctly handle binaries built with read-only segments (rosegment), which is enabled by default for LLD (LLVM's linker).

> Solution: When targeting older Android versions, consider using linker flags like --no-rosegment to maintain compatibility

jsyjbaba (approved):
LGTM 
It looks like the issue should be resolved with the provided fix.

PR #9709, opened 2025-03-18
gpuav: No OOB vertex validation when robustness on

arno-lunarg:
VUs from the 02721 family should not be checked when robustness is on


spencer-lunarg (approved):

spencer-lunarg (changes requested):
Actually I want a GPU-AV Positive test that would be OOB, but the test uses `robustBufferAccess` and we Don't report an error

spencer-lunarg (approved):

PR #9710, opened 2025-03-18
gpuav: VUID-vkCmdDrawMultiIndexedEXT-None-02721

arno-lunarg:


layers/gpuav/core/gpuav_record.cpp
```diff
@@ -483,11 +483,9 @@ void Validator::PreCallRecordCmdDrawMultiIndexedEXT(VkCommandBuffer commandBuffe
         InternalError(commandBuffer, record_obj.location, "Unrecognized command buffer.");
         return;
     }
-    for (uint32_t i = 0; i < drawCount; i++) {
```
arno-lunarg, line 486: looping like this is not needed, `CmdDrawMultiIndexedEXT` is indeed in the spirit like submitting multiple draws with one command, but for the purpose of validation, it is just another draw command

spencer-lunarg (approved):

PR #9715, opened 2025-03-19
tests: Help setup printf for GPU-AV

arno-lunarg:


tests/unit/gpu_av_positive.cpp
```diff
@@ -21,17 +21,26 @@
 #include "../framework/gpu_av_helper.h"
 #include "../../layers/gpuav/shaders/gpuav_shaders_constants.h"
 
+// #define PRINTF_FOR_GPU_AV_TESTS
```
spencer-lunarg, line 24: remove?

spencer-lunarg (approved):
I will say, I will personally never use this because of the macro, I will not stop you if you think this works for you, but honestly what I would do instead is 

1. Just force `VK_KHR_shader_non_semantic_info` on for people when they use debug printf (**we really should** anyway, it is an annoying error to get) https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9716
2. enable it with `VK_LAYER_PRINTF_ENABLE` locally instead of `VK_VALIDATION_FEATURE_ENABLE_DEBUG_PRINTF_EXT` in the code
3. add a `--print-info` type flag that will adjust the ErrorMonitor class to always print Info warning messages (or for me, I just use `VK_LAYER_PRINTF_TO_STDOUT` instead as it faster and doesn't need a recompile)

PR #9717, opened 2025-03-19
gpuav: Rework of Post Processing

spencer-lunarg:
This PR brings down the perf time of Post Processing on traces while maintaining the core idea

What has happened

- We don't loop for each "action command snapshot", instead we just bake the command index into the Post Processing Slot in the shader
- To find the pipeline/shaderObject, we save the `unique_shader_id` from the shader to do our lookup to find it
- As an optimization (which can change later easily) we have an atomicMin/Max in the shader to know where the range to look is. From profiling, found things with large descriptor array will use a small window and can make the scanning faster


layers/drawdispatch/descriptor_validator.h
```diff
@@ -49,6 +49,11 @@ class DescriptorValidator : public Logger {
     bool ValidateBindingDynamic(const spirv::ResourceInterfaceVariable& binding_info, DescriptorBinding& binding,
                                 const uint32_t index);
 
+    // For GPU-AV, these can become aliased and need to be mutable between descriptor accesses
```
spencer-lunarg, line 52: this isn't ideal, but can figure out later if it makes more sense to pass these in as arguments to `ValidateBindingDynamic` or to just re-create the `DescriptorValidator` class each time (which doesn't sound ideal)

thing to note, there are only 2 spots using this class (static and dynamic) so it is easy to control who is actually setting these

layers/gpuav/descriptor_validation/gpuav_descriptor_validation.cpp
```diff
@@ -218,33 +210,87 @@ void UpdateBoundDescriptors(Validator &gpuav, CommandBuffer &cb_state, VkPipelin
...
+            // For each shader ID we can do the state object lookup once, then validate all the accesses inside of it
+            for (const auto &[shader_id, descriptor_accesses] : descriptor_access_map) {
+                auto it = state_.instrumented_shaders_map_.find(shader_id);
+                if (it == state_.instrumented_shaders_map_.end()) {
+                    assert(false);
                     continue;
                 }
 
-                context.ValidateBindingDynamic(descriptor_access.resource_variable, *descriptor_binding, descriptor_access.index);
+                const vvl::Pipeline *pipeline_state = nullptr;
```
spencer-lunarg, line 242: would love to have combined this more, but we have an overall issue in Core Val still that `vvl::Pipeline` and `vvl::ShaderObject` don't share a similar enough interface

layers/gpuav/resources/gpuav_state_trackers.cpp
```diff
@@ -426,6 +425,7 @@ void CommandBuffer::ResetCBState(bool should_destroy) {
 
 void CommandBuffer::IncrementCommandCount(VkPipelineBindPoint bind_point) {
     action_command_count++;
+    assert(action_command_count < glsl::kMaxActionsPerCommandBuffer);
```
spencer-lunarg, line 428: `kMaxActionsPerCommandBuffer` is 8k and we currently only allocate enough for 8k worth of draws in `GetCmdErrorsCountsBufferByteSize` so this is really also asserting that as well

(follow-up PR I want to focus on fixing/improving the command buffer labels)

tests/unit/gpu_av_descriptor_indexing_positive.cpp
```diff
@@ -795,7 +795,7 @@ TEST_F(PositiveGpuAVDescriptorIndexing, SampledImageShareBindingBDA) {
     image_ci.arrayLayers = 6;
     vkt::Image image_cube(*m_device, image_ci);
     image_cube.SetLayout(VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL);
-    vkt::ImageView image_view_cube = image_cube.CreateView(VK_IMAGE_VIEW_TYPE_CUBE_ARRAY);
+    vkt::ImageView image_view_cube = image_cube.CreateView(VK_IMAGE_VIEW_TYPE_CUBE);
```
spencer-lunarg, line 798: caught a bug we had with new Post Processing that was being hidden before

tests/unit/gpu_av_descriptor_post_process.cpp
```diff
@@ -2039,7 +2042,8 @@ TEST_F(NegativeGpuAVDescriptorPostProcess, SameDescriptorAcrossStagesFragment) {
     m_errorMonitor->VerifyFound();
 }
 
-TEST_F(NegativeGpuAVDescriptorPostProcess, VariableIdClash) {
+// TODO - Need way in pipeline to distinguish which stage it came from
```
arno-lunarg, line 2045: Link an issue?

spencer-lunarg:
going to wait and rebase this on https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9651

spencer-lunarg:
closing for https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9748

PR #9720, opened 2025-03-19
gpuav: Skip TexelBuffer OOB if robustness

spencer-lunarg:
We can skip texel buffer checks if robustness is on

### DIFF
--- a/tests/unit/gpu_av_descriptor_class_texel_buffer.cpp
+++ b/tests/unit/gpu_av_descriptor_class_texel_buffer.cpp
@@ -481,8 +481,8 @@ TEST_F(NegativeGpuAVDescriptorClassTexelBuffer, ImageStore) {
         layout(set = 0, binding = 0, r32f) uniform imageBuffer s_buffer;  // texel_buffer[4]
 
         void main() {
-            vec4 x = imageLoad(s_buffer, 4); // invalid load
-            imageStore(s_buffer, 5, x); // invalid store
+            vec4 x = imageLoad(s_buffer, 3); // valid load
+            imageStore(s_buffer, 4, x); // invalid store
         }
     )glsl";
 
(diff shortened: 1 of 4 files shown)


tests/unit/gpu_av_descriptor_class_texel_buffer.cpp
```diff
@@ -509,65 +509,7 @@ TEST_F(NegativeGpuAVDescriptorClassTexelBuffer, ImageStore) {
...
     m_command_buffer.End();
 
-    m_errorMonitor->SetDesiredError("UNASSIGNED-Descriptor Texel Buffer texel out of bounds", 2);
-    m_default_queue->Submit(m_command_buffer);
-    m_default_queue->Wait();
-    m_errorMonitor->VerifyFound();
-}
-
-// This will hang a GPU
-// Texel Buffers might not properly under robustBufferAccess like we think, need to investigate
```
spencer-lunarg, line 519: this was a Descriptor Indexing check, moved to test to other test file and confirmed we do catch it

tests/unit/gpu_av_descriptor_class_texel_buffer.cpp
```diff
@@ -481,8 +481,8 @@ TEST_F(NegativeGpuAVDescriptorClassTexelBuffer, ImageStore) {
         layout(set = 0, binding = 0, r32f) uniform imageBuffer s_buffer;  // texel_buffer[4]
 
         void main() {
-            vec4 x = imageLoad(s_buffer, 4); // invalid load
-            imageStore(s_buffer, 5, x); // invalid store
+            vec4 x = imageLoad(s_buffer, 3); // valid load
```
spencer-lunarg, line 484: the test above is called `ImageLoad` and testing the load, this test should just test the store

arno-lunarg (approved):

PR #9723, opened 2025-03-19
tests: Remove unnecessary layout transitions

artem-lunarg:
This removes unnecessary `SetLayout` calls that submit layout transition jobs on the queue for the tests that perform validation of *command buffer recording*. All such tests can't access current layout state of the image. If any recording function needs image layout then it is provided as a parameter (for example, `imageLayout` parameter from `vkCmdClearDepthStencilImage`).

The following tests usually need image layout and were not updated:
* tests that call QueueSubmit
* gpu-av tests which are extremely submit friendly
* host copy which does not work with command buffers and runs on the host, so it has access to current image layout.

Currently we have two kinds of `SetLayout` - one which records to command buffer and another one which immediately applies layout by doing internal submission to the queue. One idea (but not here) is to rename the latter to something like SubmitLayoutTransition to show it performs queue submission and changes global image state, so you can't be mistaken what you are doing.


spencer-lunarg (approved):
I hate these, please purge as many as you can

PR #9724, opened 2025-03-19
gpuav: Force on Robustness 2 as well

spencer-lunarg:
For those using `VK_LAYER_GPUAV_FORCE_ON_ROBUSTNESS` this will now also turn on `robustBufferAccess2` and `robustImageAccess2` as well

### DIFF
--- a/layers/gpuav/spirv/descriptor_class_general_buffer_pass.cpp
+++ b/layers/gpuav/spirv/descriptor_class_general_buffer_pass.cpp
@@ -30,9 +30,7 @@ namespace gpuav {
 namespace spirv {
 
 DescriptorClassGeneralBufferPass::DescriptorClassGeneralBufferPass(Module& module)
-    : Pass(module),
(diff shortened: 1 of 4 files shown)


layers/gpuav/spirv/descriptor_class_general_buffer_pass.cpp
```diff
@@ -30,9 +30,7 @@ namespace gpuav {
 namespace spirv {
 
 DescriptorClassGeneralBufferPass::DescriptorClassGeneralBufferPass(Module& module)
-    : Pass(module),
-      // robustBufferAccess safety to got into unsafe mode to improve performance
-      unsafe_mode_(module.settings_.unsafe_mode || module.enabled_features_.robustBufferAccess) {
```
spencer-lunarg, line 35: we already made it now so the whole Pass is just skipped, so need to set the unsafe mode with it

arno-lunarg (approved):

PR #9728, opened 2025-03-20
layers: Remove layout state from vkt::Image

artem-lunarg:
This stops using `Image::image_layout_` state and instead the test directly specifies current layout when necessary. Usually the test knows exactly what is the current layout. The layout member is not removed yet and does some assert checks to help with this cleanup, but it is not used by any logic anymore.

Some reasons why not to track layout inside image object:
* It was very common that intead of using, for example, VK_LAYOUT_GENERAL constant directly, the test at first initiated real layout transition (`SetLayout`) only because this also sets `image_layout_` and then it was read using `Layout()` getter (`Layout()` getter is gone too), so very advanced way to get a constant value.
* Another potential issue is related to Vulkan design conventions. In Vulkan VkImage does not track current layout on the API level, so the image can be used in threaded environment where each thread tracks the layout per command buffer. By tracking state in our wrapper object this creates potential issues for some advanced test scenarios. Also hidden state makes the tests just harder to follow. One idea if we need some convenience of automatic layout tracking we can provide higher level helper which will combine vkt::Image and current layout state.


spencer-lunarg (approved):
if CI is happy, I am happy

artem-lunarg:
Yeah, some places where SetLaout is called that runs only on specific configuraition. Need to catch them all.

PR #9729, opened 2025-03-20
gpuav: Add an EarlySkip option in the pass

spencer-lunarg:
Talking with @arno-lunarg about how to streamline the Passes, I wanted to move this check inside the pass as it might change between shaders, but also not tied to the SPIR-V itself and falls into this strange grey area

This just adds an option so each pass (which actually knows what it does) is in charge of deciding to early return, not the "chassis" calling it

### DIFF
--- a/layers/gpuav/spirv/descriptor_indexing_oob_pass.cpp
+++ b/layers/gpuav/spirv/descriptor_indexing_oob_pass.cpp
@@ -151,6 +151,13 @@ uint32_t DescriptorIndexingOOBPass::CreateFunctionCall(BasicBlock& block, Instru
     return function_result;
 }
 
+bool DescriptorIndexingOOBPass::EarlySkip() const {
+    if (module_.set_index_to_bindings_layout_lut_.empty()) {
+        return true;  // If there is no bindings, nothing to instrument
+    }
+    return false;
+}
+
 void DescriptorIndexingOOBPass::NewBlock(const BasicBlock&, bool is_original_new_block) {
     // Don't clear if the new block occurs from control flow breaking one up
     if (is_original_new_block) {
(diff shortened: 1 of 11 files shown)


layers/gpuav/spirv/pass.cpp
```diff
@@ -25,7 +25,13 @@ namespace gpuav {
 namespace spirv {
 
 bool Pass::Run() {
-    const bool modified = Instrument();
+    bool modified = false;
+    if (EarlySkip()) {
+        return modified;
+    }
+
+    modified = Instrument();
```
arno-lunarg, line 33: just having `Instrument()` do any "early skip" check is simpler than adding a dedicated function?

spencer-lunarg: Yes, the problem is `InjectConditionalFunctionPass` is the one calling `Instrument()`

`BufferDeviceAddressPass` and `DescriptorIndexingOOBPass` both don't have a direct `Instrument` to it

But the early skip only works for `DescriptorIndexingOOBPass`

----

The issue is either we have every pass redo the Conditional Function injection logic

or

We keep the unified `Instrument` call an allow things like this

... what I really want is to redo the part of `InjectConditionalFunctionPass` everyone inherits from, but Control Flow is just not fun and other fires to put out

layers/gpuav/spirv/descriptor_indexing_oob_pass.cpp
```diff
@@ -151,6 +151,13 @@ uint32_t DescriptorIndexingOOBPass::CreateFunctionCall(BasicBlock& block, Instru
     return function_result;
 }
 
+bool DescriptorIndexingOOBPass::EarlySkip() const {
```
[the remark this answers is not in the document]
arno-lunarg: ah ok! Well let's get this in as is then

arno-lunarg (approved):

PR #9735, opened 2025-03-20
tests: Fix image CreateInfo incorrect use

artem-lunarg:
Sometimes the code assumed that `CreateInfo()` returns the image's `create_info_`, but in reality, it returns default create info structure.

Rename `CreateInfo()` to `DefaultCreateInfo()` to avoid confusion, and reintroduce `CreateInfo()`, which now returns the actual image create info.

### DIFF
--- a/tests/framework/binding.h
+++ b/tests/framework/binding.h
@@ -796,7 +796,8 @@ class Image : public internal::NonDispHandle<VkImage> {
                             VkPipelineStageFlags src_stages = VK_PIPELINE_STAGE_ALL_COMMANDS_BIT,
                             VkPipelineStageFlags dest_stages = VK_PIPELINE_STAGE_ALL_COMMANDS_BIT);
 
-    static VkImageCreateInfo CreateInfo();
+    const VkImageCreateInfo &CreateInfo() const { return create_info_; }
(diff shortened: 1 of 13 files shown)


tests/framework/binding.h
```diff
@@ -1291,14 +1292,15 @@ inline VkQueryPoolCreateInfo QueryPool::CreateInfo(VkQueryType type, uint32_t sl
     return info;
 }
 
-inline VkImageCreateInfo Image::CreateInfo() {
+inline VkImageCreateInfo Image::DefaultCreateInfo() {
```
spencer-lunarg, line 1295: I think `Simple2DCreateInfo` is a better name... hard to guess what a "default image" is otherwise

spencer-lunarg (approved):
LGTM - happy with either name (

tests/framework/binding.h
artem-lunarg: It's still a mess with these names, we also have existing `ImageCreateInfo2D` :( Maybe it's worth to get rid of this one in favor of  ImageCreateInfo2D?...

`Simple2DCreateInfo` might suggests that it's a ready to use object? DefaultCreateInfo does not fully initialize object, specifically the fields that are usually changes.

PR #9736, opened 2025-03-20
Validate mesh shader memory size

ziga-lunarg:
Part of #6656

### DIFF
--- a/layers/core_checks/cc_spirv.cpp
+++ b/layers/core_checks/cc_spirv.cpp
@@ -1241,11 +1241,6 @@ bool CoreChecks::ValidateWorkgroupSharedMemory(const spirv::Module &module_state
                                                uint32_t total_workgroup_shared_memory, const Location &loc) const {
     bool skip = false;
 
-    // If not found before with spec constants, find here
-    if (total_workgroup_shared_memory == 0) {
-        total_workgroup_shared_memory = module_state.CalculateWorkgroupSharedMemory();
-    }
-
     switch (stage) {
         case VK_SHADER_STAGE_COMPUTE_BIT: {
             if (total_workgroup_shared_memory > phys_dev_props.limits.maxComputeSharedMemorySize) {
@@ -1681,6 +1676,7 @@ bool CoreChecks::ValidateShaderStage(const ShaderStageState &stage_state, const 
     uint32_t local_size_y = 0;
     uint32_t local_size_z = 0;
     uint32_t total_workgroup_shared_memory = 0;
+    uint32_t total_task_payload_memory = 0;
 
     // If specialization-constant instructions are present in the shader, the specializations should be applied.
     if (module_state.static_data_.has_specialization_constants) {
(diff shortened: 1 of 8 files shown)


layers/core_checks/cc_spirv.cpp
```diff
@@ -2401,3 +2414,20 @@ bool CoreChecks::ValidateEmitMeshTasksSize(const spirv::Module &module_state, co
...
+
+bool CoreChecks::ValidateMeshMemorySize(const spirv::Module &module_state, uint32_t total_workgroup_shared_memory,
+                                        uint32_t total_task_payload_memory, const Location &loc) const {
+    bool skip = false;
+
+    if (total_task_payload_memory + total_workgroup_shared_memory >
+        phys_dev_ext_props.mesh_shader_props_ext.maxMeshPayloadAndSharedMemorySize) {
+        skip |=
+            LogError("VUID-RuntimeSpirv-maxMeshPayloadAndSharedMemorySize-08755", module_state.handle(), loc,
+                     "SPIR-V uses %" PRIu32
```
spencer-lunarg, line 2426: I would print how much of each (payload and workgroup) is used

layers/core_checks/cc_spirv.cpp
```diff
@@ -1838,6 +1843,12 @@ bool CoreChecks::ValidateShaderStage(const ShaderStageState &stage_state, const
         }
     } else {
         module_state.FindLocalSize(entrypoint, local_size_x, local_size_y, local_size_z);
+
+        total_workgroup_shared_memory = module_state.CalculateWorkgroupSharedMemory();
```
spencer-lunarg, line 1847: in theory, we can remove the other `CalculateWorkgroupSharedMemory` we called in `ValidateWorkgroupSharedMemory`... I like this approach doing it here better

tests/unit/shader_mesh.cpp
```diff
@@ -287,4 +287,43 @@ TEST_F(NegativeShaderMesh, MeshAndTaskShaderDerivatives) {
     m_errorMonitor->SetDesiredError("VUID-RuntimeSpirv-meshAndTaskShaderDerivatives-10153");
     VkShaderObj ms(this, ms_source, VK_SHADER_STAGE_MESH_BIT_EXT, SPV_ENV_VULKAN_1_2, SPV_SOURCE_ASM);
     m_errorMonitor->VerifyFound();
-}
\ No newline at end of file
+}
+
+TEST_F(NegativeShaderMesh, MeshShaderPayloadMemoryOverLimit) {
```
spencer-lunarg, line 292: we should have as addition to this test

1. Shader Object version of this test
2. Something like `SharedMemorySpecConstantSet` where we use spec constant to set the limit
3. A positive test that is at the limit exactly

spencer-lunarg (changes requested):
Looks great (thanks for taking a stab at these again!) just some few nits

layers/state_tracker/shader_module.cpp
```diff
@@ -1511,6 +1511,23 @@ uint32_t Module::CalculateWorkgroupSharedMemory() const {
     return total_size;
 }
 
+uint32_t Module::CalculateTaskPayloadMemory() const {
+    uint32_t total_size = 0;
+
+    for (const Instruction* insn : static_data_.variable_inst) {
+        // StorageClass Workgroup is shared memory
+        if (insn->StorageClass() == spv::StorageClassTaskPayloadWorkgroupEXT) {
+            const uint32_t result_type_id = insn->Word(1);
```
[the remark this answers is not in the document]
ziga-lunarg: Those 2 things don't seem to be equal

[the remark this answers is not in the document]
spencer-lunarg: Ya, it is `TypeId()`, sorry

... regardless, other spots I can clean up afterwards, this works for now what you have

spencer-lunarg (approved):

PR #9739, opened 2025-03-21
build: Use Custom hash build name

spencer-lunarg:
taking idea from https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7558

This change the cmake flag from `USE_ROBIN_HOOD_HASHING` to `USE_CUSTOM_HASH_MAP` for when we try and use something other than robin hood

Note that this will still allow `-D DUSE_ROBIN_HOOD_HASHING=OFF` to turn it off to not break any tooling that was doing that while we get them to swap

### DIFF
--- a/layers/CMakeLists.txt
+++ b/layers/CMakeLists.txt
@@ -88,11 +88,19 @@ target_link_libraries(VkLayer_utils PUBLIC
 target_include_directories(VkLayer_utils SYSTEM PRIVATE external)
 target_include_directories(VkLayer_utils PUBLIC . ${API_TYPE})
 
+option(USE_CUSTOM_HASH_MAP "Use a custom hashing library that is faster than std::unordered_map and std::unordered_set (default: ON)" ON)
+
 find_package(robin_hood CONFIG)
+
+# Keep this option as legacy for people to turn off custom hashing
 option(USE_ROBIN_HOOD_HASHING "robin_hood provides faster versions of std::unordered_map and std::unordered_set" ${robin_hood_FOUND})
-if (USE_ROBIN_HOOD_HASHING)
+if (NOT USE_ROBIN_HOOD_HASHING)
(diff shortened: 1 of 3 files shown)


layers/CMakeLists.txt
```diff
@@ -82,11 +82,19 @@ target_link_libraries(VkLayer_utils PUBLIC
...
 target_include_directories(VkLayer_utils PUBLIC . ${API_TYPE})
 
+option(USE_CUSTOM_HASH_MAP "Use a custom hashing library that is faster than std::unordered_map and std::unordered_set (default: ON)" ON)
+
 find_package(robin_hood CONFIG)
+
+# Keep this option as legacy for people to turn off custom hashing
 option(USE_ROBIN_HOOD_HASHING "robin_hood provides faster versions of std::unordered_map and std::unordered_set" ${robin_hood_FOUND})
-if (USE_ROBIN_HOOD_HASHING)
+if (NOT USE_ROBIN_HOOD_HASHING)
```
artem-lunarg, line 91: Any chance this will disable `USE_CUSTOM_HASH_MAP` in the scenario when we set `USE_CUSTOM_HASH_MAP` but not `USE_ROBIN_HOOD_HASHING`?

spencer-lunarg: only if it can't find Robin Hood in your external dep, which is then currently also would turn off

artem-lunarg (approved):

spencer-lunarg:
> @spencer-lunarg
[Merge branch 'main' into spencer-lunarg-custom-hash-name](https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9739/commits/42e71fc752d197d6e8a5a69793070e5c3a793eeb)

..... sometimes i REALLY hate github... need to repush without this commit else CI will not be happy...

PR #9754, opened 2025-03-21
gpuav: Create common Pre/Post Action spot

spencer-lunarg:
in `gpuav_record.cpp` creates a single `PreCallActionCommand`/`PostCallActionCommand` spot so next time I need to update something I don't have to copy-and-probably-mis-paste in one of the 23 variation of draw calls we have

### DIFF
--- a/layers/gpuav/core/gpuav.h
+++ b/layers/gpuav/core/gpuav.h
@@ -151,6 +151,12 @@ class Validator : public GpuShaderInstrumentor {
         VkCommandBuffer commandBuffer,
         const VkBindDescriptorBufferEmbeddedSamplersInfoEXT* pBindDescriptorBufferEmbeddedSamplersInfo,
         const RecordObject& record_obj) final;
+
+    void PreCallActionCommand(Validator& gpuav, CommandBufferSubState& cb_state, VkPipelineBindPoint bind_point,
+                              const Location& loc);
+    void PostCallActionCommand(Validator& gpuav, CommandBufferSubState& cb_state, VkPipelineBindPoint bind_point,
+                               const Location& loc);
+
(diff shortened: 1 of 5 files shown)


layers/gpuav/descriptor_validation/gpuav_descriptor_validation.cpp
```diff
@@ -32,36 +32,6 @@
...
-        return;
-    }
-
-    // Should have just been updated
-    if (!last_bound.pipeline_state && !last_bound.HasShaderObjects()) {
-        gpuav.InternalError(gpuav.device, loc, "Unrecognized pipeline nor shader object");
-        return;
-    }
-
-    // TODO - Add Shader Object support
```
spencer-lunarg, line 48: this is not true, we support shader object now and this whole function can be gone, this prompt me to not have to re-add 23 `PreCallActionCommand` calls again when we suddenly want something

layers/gpuav/core/gpuav.h
```diff
@@ -151,6 +151,13 @@ class Validator : public GpuShaderInstrumentor {
         VkCommandBuffer commandBuffer,
         const VkBindDescriptorBufferEmbeddedSamplersInfoEXT* pBindDescriptorBufferEmbeddedSamplersInfo,
         const RecordObject& record_obj) final;
+
+    // Chassis for command things to do before/after all Draw/Dispatch/TraceRays action commands
```
arno-lunarg, line 155: 英語

jeremyg-lunarg: Don't call it Chassis, that already has a very specific meaning in this codebase.  "Methods for command things..."

arno-lunarg (approved):

PR #9757, opened 2025-03-22
gpuav: Use API Prefix for naming

spencer-lunarg:
Trying to add a fix in the validation cmds is hard because how many conflating names have `buffer` and `draw` in them.. sometime we say "cpu" but did that mean what the app or layer did on the cpu?

This does 2 things

1. For the "stride", "offset", "buffer", "count" etc that come from the **API** (`vkCmdDraw...`) I prefix with `api_`
2. For things where we want to know if byte vs dwords, I put as a suffix (sometimes it was randomly in the middle)


arno-lunarg (approved):
Thank you for the renamings, I like that you put all byte/dwords as a suffix. Not too sure about `api_buffer` though, you lose info here. I will just revert the removal of a new line for long `layout(set = ..., binding = ...)`, it deteriorates readability

PR #9758, opened 2025-03-22
tests: Fix TestICD freeing for aligned_alloc

aqnuep:
The Test ICD incorrectly uses `free` to free memory allocated with `_aligned_alloc` on Windows.

I don't know how this didn't cause any issue for you, but for us it resulted in hanging CI on Github Actions and in our own CI.


spencer-lunarg (approved):
I have no idea how this was getting through, I guess we don't run the test on Windows with TestICD on CI (we run with real GPUs locally)

aqnuep:
I see. We do have both Linux and Windows tests with the Test ICD downstream.

PR #9759, opened 2025-03-22
layers: Fix pCode typo

TrentShailer:


spencer-lunarg (approved):
thanks! confirmed these are the only 2 spots

PR #9765, opened 2025-03-24
gpuav: Initialize Private Variable

spencer-lunarg:
Takes code from https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9471 that allows to proper initialize the private thread variable to prevent strange undefined behavior on the GPU

(Honestly a cool check in general and created issue for it https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9766)


layers/gpuav/spirv/module.cpp
```diff
@@ -510,12 +513,23 @@ void Module::LinkFunction(const LinkInfo& info) {
...
-            AddInterfaceVariables(new_result_id, (spv::StorageClass)new_inst->Word(3));
+            const spv::StorageClass storage_class = new_inst->StorageClass();
+            AddInterfaceVariables(new_result_id, storage_class);
             id_swap_map[old_result_id] = new_result_id;
             new_inst->ReplaceResultId(new_result_id);
             new_inst->ReplaceLinkedId(id_swap_map);
 
             const Type* type = type_manager_.FindTypeById(new_inst->TypeId());
+
+            if (storage_class == spv::StorageClassPrivate && type->spv_type_ == SpvType::kPointer &&
```
arno-lunarg, line 524: Could we at least assert if we find a private variable that is not an int? We probably would want to initialize it to a well defined value too. I can see that bite us in the future 🐍

arno-lunarg (approved):

layers/gpuav/spirv/module.cpp
spencer-lunarg: sure can do that

PR #9767, opened 2025-03-24
gpuav: Remove unnecessary link id member variable

spencer-lunarg:
Realized there is zero reason to have a `link_function_id` member variable when we literally are saving the information in the `static LinkInfo` already

### DIFF
--- a/layers/gpuav/spirv/descriptor_indexing_oob_pass.cpp
+++ b/layers/gpuav/spirv/descriptor_indexing_oob_pass.cpp
@@ -27,49 +27,43 @@
 namespace gpuav {
 namespace spirv {
 
+// This pass has 2 variations of GLSL we can pull in. Non-Bindless is simpler and we want to use when possible
+static LinkInfo link_info_bindless = {instrumentation_descriptor_indexing_oob_bindless_comp,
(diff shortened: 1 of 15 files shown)


layers/gpuav/spirv/descriptor_indexing_oob_pass.cpp
```diff
@@ -27,49 +27,43 @@
...
+
+DescriptorIndexingOOBPass::DescriptorIndexingOOBPass(Module& module) : InjectConditionalFunctionPass(module) {
+    // reset each pass
+    link_info_bindless.function_id = 0;
+    link_info_bindless_combined_image_sampler.function_id = 0;
+    link_info_non_bindless.function_id = 0;
+}
+
 // By appending the LinkInfo, it will attempt at linking stage to add the function.
 uint32_t DescriptorIndexingOOBPass::GetLinkFunctionId(bool is_combined_image_sampler) {
```
spencer-lunarg, line 52: The reality is we are going to have more passes where there are multiple functions that might be linked

To make this code simpler, we can just point to a `LinkInfo` and update it instead of trying to dance between the `LinkInfo::function_id` and some random member variable

arno-lunarg (approved):

Issue #9721, opened 2025-03-19
GPU-AV Index Buffer slowness


spencer-lunarg:
the [computecullandlod](https://github.com/SaschaWillems/Vulkan/blob/master/examples/computecullandlod/computecullandlod.cpp) example on my machine runs at 9 FPS with GPU-AV on but setting `VK_LAYER_GPUAV_INDEX_BUFFERS=0` gets me back to 30 FPS

Need to understand where this slowness is coming from and what we can do to fix it

PR #9772, opened 2025-03-24
gpuav: Improve perf for draw indexed indirect val

arno-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9721


layers/gpuav/shaders/validation_cmd/setup_draw_indexed_indirect_index_buffer.comp
```diff
@@ -0,0 +1,72 @@
...
+    if ((pc.flags & kIndexedIndirectDrawFlags_DrawCountFromBuffer) == 0) {
+        draw_count = pc.api_draw_count;
+    } else {
+        // Need to clamp draw count value stored in count buffer to the maxDrawCount
+        // that was specified at vkCmdDrawIndexedIndirect time.
+        // pc.api_draw_count is used to store maxDrawCount
+        draw_count = min(count_buffer[pc.api_count_buffer_offset_dwords], pc.api_draw_count);
+    }
+    
+    dispatch_indirect_x = (draw_count / 64) + uint((draw_count % 64) > 0);
```
spencer-lunarg, line 69: is this `64` just tied to the thread count in the other shader? should we at least have a comment, or better, a `shader_constant.h` they both use

layers/gpuav/shaders/validation_cmd/setup_draw_indexed_indirect_index_buffer.comp
```diff
@@ -0,0 +1,72 @@
...
+Matching Struct from the API
+struct VkDrawIndexedIndirectCommand {
+    uint indexCount;
+    uint instanceCount;
+    uint firstIndex;
+    uint vertexOffset;
+    uint firstInstance;
+};
+*/
+const uint kIndexCount = 0;
```
spencer-lunarg, line 38: remove for keeping the shader minimal?

layers/gpuav/shaders/validation_cmd/draw_indexed_indirect_index_buffer.comp
```diff
@@ -41,16 +41,18 @@ const uint kFirstIndex = 2;
...
 
 // CountBuffer won't be bound for non-count draws
-layout(set = kDiagPerCmdDescriptorSet, binding = kPreDrawBinding_CountBuffer) readonly buffer CountBuffer {
+layout(set = kDiagPerCmdDescriptorSet, binding = kPreDrawBinding_CountBuffer) 
+readonly buffer CountBuffer {
     uint count_buffer[];
 };
 
-layout(local_size_x = 1, local_size_y = 1, local_size_z = 1) in;
+layout(local_size_x = 64, local_size_y = 1, local_size_z = 1) in;
```
spencer-lunarg, line 55: why only 64? comment to at least say "just from testing" or something if there is required meaning it couldn't be updated in the future for some reason

layers/gpuav/validation_cmd/gpuav_draw.cpp
```diff
@@ -870,42 +915,94 @@ void DrawIndexedIndirectIndexBuffer(Validator &gpuav, CommandBufferSubState &cb_
...
 
-            shader_resources.draw_buffer_binding.info = {api_buffer, 0, VK_WHOLE_SIZE};
-            shader_resources.push_constants.api_offset_dwords = (uint32_t)api_offset / sizeof(uint32_t);
+            setup_validation_dispatch_pipeline.BindShaderResources(gpuav, cb_state, draw_i, error_logger_i,
+                                                                   setup_validation_shader_resources);
 
-            validation_pipeline.BindShaderResources(gpuav, cb_state, draw_i, error_logger_i, shader_resources);
+            DispatchCmdBindPipeline(cb_state.VkHandle(), VK_PIPELINE_BIND_POINT_COMPUTE,
+                                    setup_validation_dispatch_pipeline.pipeline);
+            DispatchCmdDispatch(cb_state.VkHandle(), 1, 1, 1);
```
spencer-lunarg, line 975: maybe a quick comment for those only reading the CPU code quick that we need to do this because the draw_count is a GPU decided value we need to get out

layers/gpuav/shaders/validation_cmd/setup_draw_indexed_indirect_index_buffer.comp
```diff
@@ -0,0 +1,72 @@
...
+buffer DipatchIndirectBuffer {
+    uint dispatch_indirect_x;
+    uint dispatch_indirect_y;
+    uint dispatch_indirect_z;
+};
+
+layout(local_size_x = 1, local_size_y = 1, local_size_z = 1) in;
+void main() {
+    uint draw_count = 0;
+    if ((pc.flags & kIndexedIndirectDrawFlags_DrawCountFromBuffer) == 0) {
```
spencer-lunarg, line 60: quick question,  I really only saw one test before this PR, are we actually testing both code paths here?

layers/gpuav/shaders/validation_cmd/draw_indexed_indirect_index_buffer.comp
arno-lunarg: That's wave size on nvidia, AMD is 32 I think

artem-lunarg: nvidia was always 32, AMD on GCC 64 and RDNA supports both 64/32 but I guess 32 is native.

artem-lunarg: but 64 is good default for local size group to run well on different devices.

layers/gpuav/shaders/validation_cmd/setup_draw_indexed_indirect_index_buffer.comp
arno-lunarg: not here I think, this count buffer logic is the same for every check in `gpuav_draw` so I did not bother to do that here

spencer-lunarg (approved):
can confirm gets perf increase locally

PR #9775, opened 2025-03-24
tests: Add helper for sync2 barriers

artem-lunarg:
The helpers allow to skip `VkDependencyInfo` part of initialization when barrier command specifies a single barrier, which is a common case in the tests.

For now, there are no helpers for sync1 barriers. They can't be as nice as sync2 helpers, because you also need two additional parameter for the stage masks, so the signature will need 4 parameters in total and it's not obvious if this brings benefits comparing to using sync1 api directly. We can reevaluate this later if needed.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -46,8 +46,8 @@ target_sources(vk_layer_validation_tests PRIVATE
     framework/ray_tracing_helper_nv.cpp
     framework/external_memory_sync.h
     framework/external_memory_sync.cpp
-    framework/barrier_queue_family.h
-    framework/barrier_queue_family.cpp
+    framework/sync_helper.h
+    framework/sync_helper.cpp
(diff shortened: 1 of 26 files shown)


tests/CMakeLists.txt
```diff
@@ -46,8 +46,8 @@ target_sources(vk_layer_validation_tests PRIVATE
     framework/ray_tracing_helper_nv.cpp
     framework/external_memory_sync.h
     framework/external_memory_sync.cpp
-    framework/barrier_queue_family.h
```
spencer-lunarg, line 49: I've been afraid to touch this file for years.... :pray:

artem-lunarg: Just a rename, the code still there, but the feeling most of it should go away, eventually...

spencer-lunarg (approved):

PR #9781, opened 2025-03-25
gpuav: Compress error output messages

spencer-lunarg:
I looked into reduce how much we log in our Error Record, the issue is we need all the information and trying to isolate it doesn't work as it would get inlined anyway.

This attempts to compress the data so we don't waste an entire dword on values we know are only ever in a smaller range

The main thing is getting the `kHeaderSize` from `11` to `7` which removes 4 writes per instrumentation function

With both unsafe off and on, I see about a compile times increase about 15-20% on MESA


layers/gpuav/shaders/instrumentation/buffer_device_address.comp
```diff
@@ -105,24 +105,19 @@ bool inst_buffer_device_address(
...
-        inst_errors_buffer.data[write_pos + kHeaderCommandResourceIdOffset] = inst_cmd_resource_index_buffer.index[0];
+        inst_errors_buffer.data[write_pos + kHeaderActionIdOffset] = (inst_action_index_buffer.index[0] << kActionIdShift) | inst_cmd_resource_index_buffer.index[0];
 
         inst_errors_buffer.data[write_pos + kInstBuffAddrUnallocDescPtrLoOffset] = uint(addr);
         inst_errors_buffer.data[write_pos + kInstBuffAddrUnallocDescPtrHiOffset] = uint(addr >> 32u);
-        inst_errors_buffer.data[write_pos + kInstBuffAddrAccessByteSizeOffset] = access_byte_size;
-        inst_errors_buffer.data[write_pos + kInstBuffAddrAccessOpcodeOffset] = access_opcode;
-        inst_errors_buffer.data[write_pos + kInstBuffAddrAccessAlignmentOffset] = alignment;
+
+        const uint payload = access_byte_size | (alignment << kInstBuffAddrAccessPayloadShiftAlignment) | (is_write << 31);
```
arno-lunarg, line 119: `access_byte_size` and `is_write` should probably also specify some constant shift, seems arbitrary to only have it for alignment

arno-lunarg (approved):
hope our tests are working well and helped in making those changes

spencer-lunarg:
> CI Vulkan-ValidationLayers build # 19559 failed.

Hit known issue being looked in of Window Arm failing `PositiveMemory.BindMemoryStatusBuffer` (GPU-AV tests all passed on the run)

PR #9783, opened 2025-03-25
msvc: Add natvis file for custom vvl types

artem-lunarg:
This adds Visual Studio's natvis file to better visualize custom vvl types in the msvc debugger. As the first example I added `sparse_container::small_range_map`, it stores data in generric uint8_t array, so by default debugger can't see anything.

before:

after:

Initial visualization is still available in `[Raw View]` field.


spencer-lunarg (approved):

artem-lunarg:
Added small_vector support

PR #9784, opened 2025-03-25
layers: Cleanup BothRangeMap

artem-lunarg:
Remove kTrisolar state. `BothRangeMap` has a single constructor that creates either a small or a big map, no other option is possible.  More streamlined assignment/copy implementation - in some places no need to do conditions for small or big use case if member-wise assignment gives the same result and potentially is more performant (no conditions).


spencer-lunarg (approved):
`git blame` for image layouts is now Artem!!!

PR #9790, opened 2025-03-26
layers: Use Parallel Hashmap

arno-lunarg:
Finish up https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9740

### DIFF
--- a/layers/containers/custom_containers.h
+++ b/layers/containers/custom_containers.h
@@ -26,7 +26,7 @@
 #include <utility>
 
 #ifdef USE_CUSTOM_HASH_MAP
-#include "robin_hood.h"
+#include "parallel_hashmap/phmap.h"
 #else
 #include <map>
 #include <set>
@@ -41,54 +41,17 @@ namespace vvl {
 
 #ifdef USE_CUSTOM_HASH_MAP
 template <typename T>
-using hash = robin_hood::hash<T>;
+using hash = phmap::Hash<T>;
 
-template <typename Key, typename Hash = robin_hood::hash<Key>, typename KeyEqual = std::equal_to<Key>>
-using unordered_set = robin_hood::unordered_set<Key, Hash, KeyEqual>;
+template <typename Key, typename Hash = phmap::Hash<Key>, typename KeyEqual = std::equal_to<Key>>
+using unordered_set = phmap::flat_hash_set<Key, Hash, KeyEqual>;
 
-template <typename Key, typename T, typename Hash = robin_hood::hash<Key>, typename KeyEqual = std::equal_to<Key>>
-using unordered_map = robin_hood::unordered_map<Key, T, Hash, KeyEqual>;
+template <typename Key, typename T, typename Hash = phmap::Hash<Key>, typename KeyEqual = std::equal_to<Key>>
+using unordered_map = phmap::flat_hash_map<Key, T, Hash, KeyEqual>;
 
 template <typename Key, typename T>
-using map_entry = robin_hood::pair<Key, T>;
+using map_entry = phmap::Pair<Key, T>;
 
-// robin_hood-compatible insert_iterator (std:: uses the wrong insert method)
-// NOTE: std::iterator was deprecated in C++17, and newer versions of libstdc++ appear to mark this as such.
-template <typename T>
-struct insert_iterator {
-    using iterator_category = std::output_iterator_tag;
-    using value_type = typename T::value_type;
-    using iterator = typename T::iterator;
-    using difference_type = void;
-    using pointer = void;
-    using reference = T &;
-
-    insert_iterator(reference t, iterator i) : container(&t), iter(i) {}
-
-    insert_iterator &operator=(const value_type &value) {
-        auto result = container->insert(value);
-        iter = result.first;
-        ++iter;
-        return *this;
-    }
-
-    insert_iterator &operator=(value_type &&value) {
-        auto result = container->insert(std::move(value));
-        iter = result.first;
-        ++iter;
-        return *this;
-    }
-
-    insert_iterator &operator*() { return *this; }
-
-    insert_iterator &operator++() { return *this; }
-
-    insert_iterator &operator++(int) { return *this; }
-
-  private:
-    T *container;
-    iterator iter;
-};
 #else
 template <typename T>
 using hash = std::hash<T>;
@@ -102,8 +65,6 @@ using unordered_map = std::unordered_map<Key, T, Hash, KeyEqual>;
 template <typename Key, typename T>
 using map_entry = std::pair<Key, T>;
 
-template <typename T>
-using insert_iterator = std::insert_iterator<T>;
 #endif
 
 #if 1
--- a/layers/external/parallel_hashmap/meminfo.h
+++ b/layers/external/parallel_hashmap/meminfo.h
@@ -1,0 +1,185 @@
+#if !defined(spp_memory_h_guard)
+#define spp_memory_h_guard
+
+#include <cstdint>
+#include <cstring>
+#include <cstdlib>
+
+#if defined(_WIN32) || defined(__CYGWIN__)
+#define SPP_WIN
+#endif
+
+#ifdef SPP_WIN
+#include <windows.h>
+#include <Psapi.h>
+#undef min
+#undef max
+#elif defined(__linux__)
+#include <sys/types.h>
+#include <sys/sysinfo.h>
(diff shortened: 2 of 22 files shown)


layers/containers/custom_containers.h
```diff
@@ -26,7 +26,7 @@
 #include <utility>
 
 #ifdef USE_CUSTOM_HASH_MAP
-#include "robin_hood.h"
+#include "external/parallel_hashmap/phmap.h"
```
spencer-lunarg, line 29: you shouldn't need `external` here, we already have the folder included as a target include already

```suggestion
#include "parallel_hashmap/phmap.h"
```

layers/external/parallel_hashmap/meminfo.h
```diff
@@ -0,0 +1,185 @@
+#if !defined(spp_memory_h_guard)
```
spencer-lunarg, line 1: are we running these through clang-format? (I don't want to, but we don't have a "exclude" list in the CI check I think)

layers/containers/custom_containers.h
arno-lunarg: I did that for some test complaining about not finding `parallel_hashmap/phmap.h`

spencer-lunarg: I hope no test needs `custom_container.h`

if they do, you should update the `target_include_directories` in the CMake like you did for the gpuav-spirv cmake

BUILD.gn
```diff
@@ -166,6 +166,15 @@ vvl_sources = [
   "layers/external/vma/vma.cpp",
   "layers/external/vma/vma.h",
   "layers/external/xxhash.h",
+  "layers/external/parallel_hashmap/btree.h",
```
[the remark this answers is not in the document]
arno-lunarg: The new files should not be included and compiled if `USE_CUSTOM_HASH_MAP` is false, so it should not be a problem?

layers/containers/custom_containers.h
arno-lunarg: I will fix the test

arno-lunarg: > C:\Users\Arno\lunarg\Vulkan-ValidationLayers\layers\containers\custom_containers.h(29,10): error C1083: Cannot open include file: 'parallel_hashmap/phmap.h': No such file or directory [C:\Users\Arno\
lunarg\Vulkan-ValidationLayers\build\debug\tests\layers\VkLayer_device_profile_api.vcxproj]
  (compiling source file '../../../../tests/layers/device_profile_api.cpp')

spencer-lunarg: this seems like bad header includes, I see no use of `vvl::` inside `device_profile_api.cpp`

arno-lunarg: `device_profile_api.cpp` includes `vk_dispatch_table_helper.h` that includes `vk_extension_helper.h` that includes `custom_containers.h`

arno-lunarg: but then I just need to add this:
```
target_include_directories(VkLayer_device_profile_api PRIVATE .
    ${VVL_SOURCE_DIR}/layers/external
)
```

spencer-lunarg: No, just move the header to the source and forward declare instead of hacking this with CMake

```diff
diff --git a/layers/vulkan/generated/vk_dispatch_table_helper.cpp b/layers/vulkan/generated/vk_dispatch_table_helper.cpp
index c2146c620..625846c12 100644
--- a/layers/vulkan/generated/vk_dispatch_table_helper.cpp
+++ b/layers/vulkan/generated/vk_dispatch_table_helper.cpp
@@ -1957,6 +1957,7 @@ const auto& GetApiExtensionMap() {
     return api_extension_map;
 }
 
+#include "vk_extension_helper.h"
 // Using the above code-generated map of APINames-to-parent extension names, this function will:
 //   o  Determine if the API has an associated extension
 //   o  If it does, determine if that extension name is present in the passed-in set of device or instance enabled_ext_names
diff --git a/layers/vulkan/generated/vk_dispatch_table_helper.h b/layers/vulkan/generated/vk_dispatch_table_helper.h
index 14d671e4a..e57140aac 100644
--- a/layers/vulkan/generated/vk_dispatch_table_helper.h
+++ b/layers/vulkan/generated/vk_dispatch_table_helper.h
@@ -30,7 +30,8 @@
 #include <cstring>
 #include <string>
 #include "vk_layer_dispatch_table.h"
-#include "vk_extension_helper.h"
+
+struct DeviceExtensions;
 
 // Using the above code-generated map of APINames-to-parent extension names, this function will:
 //   o  Determine if the API has an associated extension
```

arno-lunarg: `vk_extension_helper.h` is being included in tests, with no way to remove references to `custom_containers.h` since we need vvl maps here:
image

so we need `custom_containers.h` in tests

arno-lunarg: Still I made the change you mentioned to `vk_dispatch_table_helper`, one less include is always nice

spencer-lunarg: ... this is 100% getting outside the scope of this PR, can rry and cleanup after, I already spotted many random includes in the test code now... so sure, signed off for this part of the change, thanks for making the small change while at it

spencer-lunarg (approved):

BUILD.gn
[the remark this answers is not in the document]
jeremyg-lunarg: USE_CUSTOM_HASH_MAP should stay off in BUILD.gn and these files shouldn't be included here. The underlying problem in the angle bug is not solved by switching to a different unapproved pile of code.

jeremyg-lunarg (changes requested):

BUILD.gn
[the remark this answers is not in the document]
spencer-lunarg: yes, good catch, so 

1. Remove these includes here
2. Add these new header files to `scripts/gn/export_targets.py` where `b'robin_hood.h',` was to be `phmap.h` instead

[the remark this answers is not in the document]
jeremyg-lunarg: why would we need to list these files when we weren't listing RH files? Also google really doesn't want to use non-approved code like this and NOT listing these files would help catch if they start getting used somehow.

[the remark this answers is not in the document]
spencer-lunarg: ok, fair point, so then just removing the files in the `build.gn` good enough?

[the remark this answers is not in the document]
jeremyg-lunarg: I think so?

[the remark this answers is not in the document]
arno-lunarg: In the linux gn build github action I get:

```
Run python scripts/gn/gn.py --args='target_os="linux" target_cpu="x64"'
Cloning into 'depot_tools'...
Starting builds in /home/runner/work/Vulkan-ValidationLayers/Vulkan-ValidationLayers/external
WARNING: Your metrics.cfg file was invalid or nonexistent. A new one will be created.
________ running 'python3 build/linux/sysroot_scripts/install-sysroot.py --arch=x64' in '/home/runner/work/Vulkan-ValidationLayers/Vulkan-ValidationLayers/.'
Installing Debian bullseye amd64 root image: /home/runner/work/Vulkan-ValidationLayers/Vulkan-ValidationLayers/build/linux/debian_bullseye_amd64-sysroot
Downloading https://commondatastorage.googleapis.com/chrome-linux-sysroot/dec7a3a0fc5b83b909cba1b6d119077e0429a138eadef6bf5a0f2e03b1904631
________ running 'python3 tools/clang/scripts/update.py' in '/home/runner/work/Vulkan-ValidationLayers/Vulkan-ValidationLayers/.'
Downloading https://commondatastorage.googleapis.com/chromium-browser-clang/Linux_x64/clang-llvmorg-19-init-14561-gecea8371-2.tar.xz .......... Done.
Done. Made 57 targets from 65 files in 170ms
Importing graph
  ('gn', 'desc', '--format=json', 'out', '*')

Processing graph
  shared_library //:VkLayer_khronos_validation
Warning in //:VkLayer_khronos_validation: layers/containers/custom_containers.h: Included file must be listed in the GN target or its public dependency: b'parallel_hashmap/phmap.h'
```

if I do not list the files, `gn` does not care about the file being actually included or not : /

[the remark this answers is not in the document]
arno-lunarg: ah I did not see Spencer's comment, NVM

jeremyg-lunarg (approved):

PR #9793, opened 2025-03-26
layers: Cleanup image barrier validation

artem-lunarg:
Initial motivation here is to deal with confusing function name `UpdateCommandBufferImageLayoutMap` which does **not** update command buffer layout map. But this ended up in additional cleanup of image barrier validation.

p.s. these PRs are a side effect of me going of image barrier/layout code to fix some synchronization issues.

### DIFF
--- a/layers/core_checks/cc_synchronization.cpp
+++ b/layers/core_checks/cc_synchronization.cpp
@@ -1577,9 +1577,9 @@ bool CoreChecks::ValidateSubpassDependency(const ErrorObject &error_obj, const L
     return skip;
 }
 
-// Verify an ImageMemoryBarrier's old/new ImageLayouts are compatible with the Image's ImageUsageFlags.
-bool CoreChecks::ValidateBarrierLayoutToImageUsage(const Location &layout_loc, VkImage image, VkImageLayout layout,
-                                                   VkImageUsageFlags usage_flags) const {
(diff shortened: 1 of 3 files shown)


layers/core_checks/cc_synchronization.cpp
```diff
@@ -2375,8 +2362,7 @@ bool CoreChecks::ValidateBarriers(const Location &outer_loc, const vvl::CommandB
         const ImageBarrier barrier(pImageMemBarriers[i], src_stage_mask, dst_stage_mask);
         const OwnershipTransferOp transfer_op = barrier.TransferOp(cb_state.command_pool->queueFamilyIndex);
         skip |= ValidateMemoryBarrier(objects, barrier_loc, cb_state, barrier, transfer_op);
-        skip |= ValidateImageBarrier(objects, barrier_loc, cb_state, barrier);
-        skip |= ValidateBarriersToImages(barrier_loc, cb_state, barrier, layout_updates_state);
+        skip |= ValidateImageBarrier(objects, cb_state, barrier, barrier_loc, layout_updates_state);
```
artem-lunarg, line 2365: Instead of two calls to validate image barrier now it's a single `ValidateImageBarrier` which includes the second call.

spencer-lunarg (approved):

PR #9794, opened 2025-03-26
Fix getting vertex attrib divisor properties

ziga-lunarg:
The issue is if both EXT and KHR extensions are enabled, I believe we should load the properties from KHR in that case, because it contains `supportsNonZeroFirstInstance`, which the EXT does not.


spencer-lunarg (approved):
had similar issue with https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9622/files

didn't think about the EXT/KHR combined case so this is better to catch now

ziga-lunarg:
> had similar issue with https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9622/files
> 
> didn't think about the EXT/KHR combined case so this is better to catch now

The simpler fix might have been just to set `supportsNonZeroFirstInstance` to `VK_TRUE` for EXT, since that is how it behaves anyway.

Issue #8089, opened 2024-06-04
GPU-AV  - Handle instrumenting inside a struct


spencer-lunarg:
a follow-up of https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8073

Was attempted in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/8075/commits/3fbcdeef44aff0b5db560c1863b9d7f1e03c4b35 but would provide an error of

> Out of bounds access: 64 bytes read at buffer device address 0x0

which still have't figured out if a driver compiler bug or not (as we seemed to be doing things correctly)

spencer-lunarg:
Update - basically it comes down in the example https://godbolt.org/z/cd7e78z3K is this a full struct load or do we need to detect only the parts used

PR #9797, opened 2025-03-27
Validate struct acceses with BDA

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8089

This is to detect if the user is trying to copy the struct. 

The core issue is Glslang has limitations (https://godbolt.org/z/s78Mo5s49) that if you only access the first part of a BDA struct, you really are accessing everything. This is something that Slang recently fixed to generate SPIR-V to match what the compiler does allowing it to not have this problem

Because I assume people will try to raise false positives, I added a `is_struct` bit in the BDA pass to help report to users what is going on here


layers/gpuav/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -817,6 +818,11 @@ bool LogMessageInstBufferDeviceAddress(const uint32_t *error_record, std::string
             uint64_t address = *reinterpret_cast<const uint64_t *>(error_record + kInstBuffAddrUnallocDescPtrLoOffset);
             strm << "Out of bounds access: " << byte_size << " bytes " << access_type << " at buffer device address 0x" << std::hex
                  << address << '.';
+            if (is_struct) {
```
spencer-lunarg, line 821: @artem-lunarg @arno-lunarg @ziga-lunarg need help from people who didn't spend 4 straight hours staring at SPIR-V. So imagine you are not using Slang (which 2 SDK ago, had the same problem) and went  https://godbolt.org/z/cd7e78z3K

Notice I never use `vec4 frustum[6]; // [64, 160]` and you "might" think you need to not bind that memory, even looking at Mesa's output, you notice it is not accessed... but confirmed we can NOT assume things are optimized out... it is up to the toolchains to generate SPIR-V that only access the single part of the struct as the SPIR-V does a "full struct load" here so therefor you have accessed all 160 bytes

I want to provide a language agnostic message to basically explain this in a way they know we are not generating a False Positive (because recently I suppressed this until I confirmed what we are checking is correct)

spencer-lunarg: cc @charles-lunarg as you will be customer support likely

spencer-lunarg: sorry - context - the PR here in `if (is_struct)` will produce an error message

> Validation Error: [ UNASSIGNED-Device address out of bounds ] | MessageID = 0x1a898625
vkCmdDispatch(): Out of bounds access: 4 bytes read at buffer device address 0x484000040. The access was from accessing the entire struct, while not all of the struct might be accessed, it is up to the source langauge or tooling to detect that and reflect it in the SPIR-V.

I am asking if this makes enough sense imaging you had this shader https://godbolt.org/z/cd7e78z3K

arno-lunarg: What about:

vkCmdDispatch(): Out of bounds access: 4 bytes read at buffer device address 0x484000040. This read corresponds to a full OpTypeStruct load (<hint in message for corresponding OpTypeStruct name>). While not all members of the struct might be accessed, it is up to the source langauge or tooling to detect that and reflect it in the SPIR-V.

arno-lunarg: > They will 100% need Shader Debug Info on for me to tell them unfortunately

Yep but IMO message wording needs to account for that, I really don't like when a message has something like `this/the/that thing` as if you are suppose to know but actually you have no clue, quite frustrating

spencer-lunarg: The (current) issue is at this point in the error message, I don't know if we are able to find the line in the source or not

The 

> SPIR-V Instruction Index = 75
Unable to source. Build shader with debug info to get source information.

Is that, I expect people to use Shader Debug Info (cause there is no other information to report) and I don't want to try saying "use that" in multiple places, it is said where we check for it

spencer-lunarg: sounds good, updated!

artem-lunarg (approved):

PR #9798, opened 2025-03-27
Fix indirect AS build crash

ziga-lunarg:


layers/core_checks/cc_ray_tracing.cpp
```diff
@@ -553,7 +553,7 @@ bool CoreChecks::ValidateAccelerationBuffers(VkCommandBuffer cmd_buffer, uint32_
 
                 if (geom_data.geometry.triangles.indexType != VK_INDEX_TYPE_NONE_KHR) {
                     auto index_buffer_states = GetBuffersByAddress(geom_data.geometry.triangles.indexData.deviceAddress);
-                    if (index_buffer_states.empty() && geometry_build_ranges[geom_i].primitiveCount > 0) {
+                    if (index_buffer_states.empty() && geometry_build_ranges && geometry_build_ranges[geom_i].primitiveCount > 0) {
```
spencer-lunarg, line 556: while this is good, there are multiple spots in this function trying to access `geometry_build_ranges` that also need this check it seems, but the are all `geometry_build_ranges[geom_i]`

so maybe during the ` for (uint32_t geom_i = 0; geom_i < info.geometryCount; ++geom_i) {` do a 

```
const VkAccelerationStructureBuildRangeInfoKHR* geometry_build_range = ? geometry_build_ranges : geometry_build_ranges[geom_i] : nullptr;
```

then you can make this at least

```suggestion
                    if (index_buffer_states.empty() && geometry_build_range && geometry_build_range.primitiveCount > 0) {
```

or maybe something else .... but basically we need fix this everywhere in this function

spencer-lunarg (approved):

layers/core_checks/cc_ray_tracing.cpp
spencer-lunarg: nice, getting it to an uint

PR #9799, opened 2025-03-27
gpuav: Add debug info to spirv dumps

arno-lunarg:
- Auto dump invalid instrumented spirv
- Add instrumentation info to dumped file names

I did this by using an horrible thread_local global variable, for it is the less intrusive way to try to have passes communicate whatever info they want to something in charge of gathering and formatting this data before dumping it

### DIFF
--- a/layers/gpuav/instrumentation/gpuav_instrumentation_debug.cpp
+++ b/layers/gpuav/instrumentation/gpuav_instrumentation_debug.cpp
@@ -1,0 +1,90 @@
+/* Copyright (c) 2020-2025 The Khronos Group Inc.
+ * Copyright (c) 2020-2025 Valve Corporation
+ * Copyright (c) 2020-2025 LunarG, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "gpuav/instrumentation/gpuav_instrumentation_debug.h"
+
+#include <cassert>
+#include <sstream>
+#include <fstream>
+
+namespace gpuav {
+
+thread_local InstrumentedShaderDebugInfo tl_instrumentation_debug_info;
+
+std::string InstrumentedShaderDebugInfo::ToStringFileSuffix() const {
+    std::stringstream ss;
+
+    ss << "bda_" << bda << "_dbgpf_" << dbgpf << "_dcgb_" << dcgb << "_dctb_" << dctb << "_dioob_" << dioob << "_postp_" << postp
+       << "_rayq_" << rayq << "_vaoob_" << vaoob;
+
+    switch (shader_stage) {
+        case VK_SHADER_STAGE_VERTEX_BIT:
+            ss << "_vert";
+            break;
+        case VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT:
+            ss << "_tess_ctrl";
+            break;
+        case VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT:
+            ss << "_tess_eval";
+            break;
+        case VK_SHADER_STAGE_GEOMETRY_BIT:
+            ss << "_geom";
+            break;
+        case VK_SHADER_STAGE_FRAGMENT_BIT:
+            ss << "_frag";
+            break;
+        case VK_SHADER_STAGE_COMPUTE_BIT:
+            ss << "_comp";
+            break;
+        case VK_SHADER_STAGE_RAYGEN_BIT_KHR:
+            ss << "_rgen";
+            break;
+        case VK_SHADER_STAGE_ANY_HIT_BIT_KHR:
+            ss << "_ahit";
+            break;
+        case VK_SHADER_STAGE_CLOSEST_HIT_BIT_KHR:
+            ss << "_chit";
+            break;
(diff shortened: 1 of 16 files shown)


layers/gpuav/instrumentation/gpuav_instrumentation_debug.h
```diff
@@ -0,0 +1,48 @@
...
+#include <filesystem>
+#include <string>
+
+namespace gpuav {
+
+struct InstrumentedShaderDebugInfo {
+    VkShaderStageFlagBits shader_stage = VK_SHADER_STAGE_FLAG_BITS_MAX_ENUM;
+
+    // instrumenation counts
+    uint32_t bda{};
```
spencer-lunarg, line 31: We have added 3 or 4 new spots to add when adding a new pass, which we just complained was already hard to add

spencer-lunarg (changes requested):
We have redone a lot of work `VK_LAYER_GPUAV_DEBUG_PRINT_INSTRUMENTATION_INFO` is doing

I would be way more incline to improve what happens when that is enabled

layers/gpuav/instrumentation/gpuav_shader_instrumentor.cpp
```diff
@@ -1216,8 +1225,10 @@ bool GpuShaderInstrumentor::InstrumentShader(const vvl::span<const uint32_t> &in
     module.ToBinary(out_instrumented_spirv);
 
     if (gpuav_settings.debug_dump_instrumented_shaders) {
-        std::string file_name = "dump_" + std::to_string(unique_shader_id) + "_after.spv";
-        DumpSpirvToFile(file_name, reinterpret_cast<const char *>(out_instrumented_spirv.data()),
+        const std::string instrumentation_info = tl_instrumentation_debug_info.ToStringFileSuffix();
```
spencer-lunarg, line 1228: I would want an option to NOT print this out as I rather not have file names that are full or random abbreviations and info I don't want

arno-lunarg:
With your input:
- I will keep `VK_LAYER_GPUAV_DEBUG_PRINT_INSTRUMENTATION_INFO`, but will use what I introduce instead to do that. Issue with current implementation is correlating multiple debug info belonging to one shader, the current `PrintDebugInfo` flow relying on `std::cout` does not allow to do that, the new one does
- If this option is on, I add instrumentation info to dumped file names

spencer-lunarg:
I think we should improve when you go `VK_LAYER_GPUAV_DEBUG_PRINT_INSTRUMENTATION_INFO` it prints out the Shader Id and stages in it

as for `PrintDebugInfo`, we either need to say 

1. the pass is charge or printing out what it wants
2. The Module prints it out, and only gets data from the Pass

not both, so maybe it makes sense to have a `PrintDebugInfo` be a higher level things, the problem is we destroy the pass after it is done, so it would need to be saved with some Module level information

> If this option is on, I add instrumentation info to dumped file names

I still don't know why you want it in the file name, if it is being printed out already with `VK_LAYER_GPUAV_DEBUG_PRINT_INSTRUMENTATION_INFO` and assuming `VK_LAYER_GPUAV_DEBUG_PRINT_INSTRUMENTATION_INFO` now prints the stage info

arno-lunarg:
> I still don't know why you want it in the file name,

info will get lost in the output if not in the file name. Sometimes you don't have the possibility to look at info printed to `std::cout`, if I just launch a game I have no terminal to look at. So having the option to print instrumentation info in the dumped file names will make sure we can always find it.

spencer-lunarg:
> So having the option to print instrumentation info in the dumped file names will make sure we can always find it.

or just move the `std::cout` to use our callback or file... I'm not gonna stop you from adding them, just I don't want it (so add it as an opt in) and will just need to remember to update it when adding a new pass

spencer-lunarg:
If you wanted, since we print the `OpName`, we could easily write a python tool that just searched all `OpFunctionCall` and printed out how many of each functions were called, that is the same thing and would work automatically for each thing added

so you go `scripts/print_gpuav_info.py dump_100_after.spv` and it could print

>  "inst_buffer_device_address" called 15 times

and it would be automatically granular to each type of passes that have multiple possible function calls

layers/gpuav/instrumentation/gpuav_instrumentation_debug.h
```diff
@@ -0,0 +1,48 @@
...
+    uint32_t rayq{};
+    uint32_t vaoob{};
+
+    void Clear() { *this = InstrumentedShaderDebugInfo(); }
+    std::string ToStringFileSuffix() const;
+};
+
+extern thread_local InstrumentedShaderDebugInfo tl_instrumentation_debug_info;
+
+void DumpSpirvToFile(const std::filesystem::path &file_path, const char *spirv_data, size_t spirv_size);
```
spencer-lunarg, line 46: can we leave in `shader_utils.h`

I have used this a few time for core validation and tests to debug things

layers/gpuav/instrumentation/gpuav_instrumentation_debug.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+#include "gpuav/instrumentation/gpuav_instrumentation_debug.h"
+
+#include <cassert>
+#include <sstream>
+#include <fstream>
+
+namespace gpuav {
+
+thread_local InstrumentedShaderDebugInfo tl_instrumentation_debug_info;
```
spencer-lunarg, line 26: why make this `thread_local`? Just have it be part of `gpauv::spirv::Module` as that is the true lifetime of it anyways

arno-lunarg:
I will keep the "dump invalid spirv" part and put it in another PR

spencer-lunarg:
> I will keep the "dump invalid spirv" part and put it in another PR

awesome

> That's probably the best way to go about it, not intrusive, and with python it will be easier to also come up with more global statistics on the overall instrumentations counts! I will close this PR in favor of that

Created issue, https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9800 - I should be able to quickly put something together today/tomorrw... the nice part of the script is it works on anything instrumented, you don't need to remember to set a flag

Issue #9585, opened 2025-03-05
False positive 09555


ziga-lunarg:
Setting an image barrier during dynamic rendering produces a false positive `VUID-vkCmdPipelineBarrier-image-09555`.

```
TEST_F(PositiveDynamicRenderingLocalRead, BarrierDuringDynamicRenderPass) {
    AddRequiredExtensions(VK_KHR_SYNCHRONIZATION_2_EXTENSION_NAME);
    AddRequiredFeature(vkt::Feature::synchronization2);
    RETURN_IF_SKIP(InitBasicDynamicRenderingLocalRead());

    vkt::Image image(*m_device, 32, 32, 1, VK_FORMAT_R8G8B8A8_UNORM,
                     VK_IMAGE_USAGE_INPUT_ATTACHMENT_BIT | VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT | VK_IMAGE_USAGE_TRANSFER_SRC_BIT |
                         VK_IMAGE_USAGE_TRANSFER_DST_BIT | VK_IMAGE_USAGE_SAMPLED_BIT);

    vkt::ImageView image_view = image.CreateView();

    m_command_buffer.Begin();

    VkImageMemoryBarrier image_memory_barrier = vku::InitStructHelper();
    image_memory_barrier.srcAccessMask = VK_ACCESS_NONE;
    image_memory_barrier.dstAccessMask = VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT;
    image_memory_barrier.oldLayout = VK_IMAGE_LAYOUT_UNDEFINED;
    image_memory_barrier.newLayout = VK_IMAGE_LAYOUT_GENERAL;
    image_memory_barrier.image = image.handle();
    image_memory_barrier.subresourceRange.aspectMask = VK_IMAGE_ASPECT_COLOR_BIT;
    image_memory_barrier.subresourceRange.baseMipLevel = 0u;
    image_memory_barrier.subresourceRange.levelCount = 1u;
    image_memory_barrier.subresourceRange.baseArrayLayer = 0u;
    image_memory_barrier.subresourceRange.layerCount = 1u;
    vk::CmdPipelineBarrier(m_command_buffer.handle(), VK_PIPELINE_STAGE_NONE, VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT, 0u, 0u, nullptr, 0u,
[...]

PR #9803, opened 2025-03-27
layers: Fix image barrier layout validation

artem-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9585

Moved some code, the fix is the best visible in its separate commit.

### DIFF
--- a/layers/core_checks/cc_synchronization.cpp
+++ b/layers/core_checks/cc_synchronization.cpp
@@ -1655,37 +1655,6 @@ bool CoreChecks::ValidateImageLayoutAgainstImageUsage(const Location &layout_loc
     return skip;
 }
 
-std::vector<uint32_t> GetUsedAttachments(const vvl::CommandBuffer &cb_state) {
-    std::set<uint32_t> unique;
-
-    for (size_t i = 0; i < cb_state.rendering_attachments.color_locations.size(); ++i) {
-        const uint32_t unmapped_color_attachment = cb_state.rendering_attachments.color_locations[i];
-        if (unmapped_color_attachment != VK_ATTACHMENT_UNUSED) {
-            unique.insert(unmapped_color_attachment);
-        }
-    }
-
-    for (size_t i = 0; i < cb_state.rendering_attachments.color_indexes.size(); ++i) {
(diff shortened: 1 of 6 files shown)


layers/core_checks/cc_synchronization.cpp
```diff
@@ -1783,41 +1752,19 @@ bool CoreChecks::ValidateImageBarrierAgainstImage(const vvl::CommandBuffer &cb_s
...
-            const auto &color_attachment = rendering_info.pColorAttachments[color_attachment_idx];
-            if (color_attachment.imageView == VK_NULL_HANDLE) {
-                continue;
-            }
-            const auto image_view_state = Get<vvl::ImageView>(color_attachment.imageView);
-            ASSERT_AND_CONTINUE(image_view_state);
-            const auto &image_view_image_state = image_view_state->image_state;
-
-            if (image == image_view_image_state->VkHandle()) {
-                auto guard = image_view_image_state->layout_range_map->ReadLock();
```
artem-lunarg, line 1810: The main issue was that global layout map was which should be used during submit time validation. Added comment about this. Command buffer recording validation should use command buffer layout map.

tests/unit/dynamic_rendering_local_read.cpp
```diff
@@ -477,7 +477,10 @@ TEST_F(NegativeDynamicRenderingLocalRead, ImageBarrierRequireFeature) {
     m_command_buffer.End();
 }
 
-TEST_F(NegativeDynamicRenderingLocalRead, ImageBarrierInProperLayout) {
+// TODO: temporary disabled: the first layout transition in the command buffer can be validation only when submitted to the queue.
+// The original validation code was incorrect: ccan't use global image layout map (updated during submit time) to validate
+// record-time state.
+TEST_F(NegativeDynamicRenderingLocalRead, DISABLED_ImageBarrierInProperLayout) {
```
artem-lunarg, line 483: This test should be rewritten. It's not possible to validate the first layout transition in command buffer against current image layout state because such type of validation can be done only during submit time. Submit time validation is not implemented yet, but going to add it. This test worked because it relied on the bug that was fixed.

spencer-lunarg (approved):

PR #9809, opened 2025-03-28
gpuav: Separated generated link info from dynamic

spencer-lunarg:
replacement for https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9808

Separates out what is generated (and can be `const static`) vs the data that is per-pass

### DIFF
--- a/layers/gpuav/spirv/link.h
+++ b/layers/gpuav/spirv/link.h
@@ -26,20 +26,25 @@ enum LinkFlags {
     ZeroInitializeUintPrivateVariables = 0x00000001,
 };
 
-struct LinkInfo {
+// This struct is for things that are generated and therefor can be created once statically
+struct OfflineLinkInfo {
     // SPIR-V module to link in
     const uint32_t* words;
     const uint32_t word_count;
-
-    // Information about the function it has
-    // (Will change each time pass is ran)
-    uint32_t function_id;
 
     // used for debugging
     const char* opname;
 
     // Optional things to be done when linking
-    uint32_t flags = 0;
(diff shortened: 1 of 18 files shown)


layers/gpuav/spirv/link.h
```diff
@@ -26,20 +26,25 @@ enum LinkFlags {
...
-
     // used for debugging
     const char* opname;
 
     // Optional things to be done when linking
-    uint32_t flags = 0;
+    const uint32_t flags = 0;
+};
+
+struct LinkInfo {
```
spencer-lunarg, line 42: this was the main issue, everything is generated and we want to make it `const static` in the source, but we need to update the `function_id` each pass as it is unique to the SPIR-V

layers/gpuav/spirv/link.h
```diff
@@ -26,20 +26,25 @@ enum LinkFlags {
     ZeroInitializeUintPrivateVariables = 0x00000001,
 };
 
-struct LinkInfo {
+// This struct is for things that are generated and therefor can be created once statically
+struct GeneratedLinkInfo {
```
arno-lunarg, line 30:
```suggestion
// This struct is for things that are generated offline and therefore can be created once statically
struct StaticLinkInfo {
```

spencer-lunarg: I had "Static" but changed  to "generated" in fear of someone not liking it... will change it (I don't really care what it is called of the 2 options)

arno-lunarg (approved):

PR #9812, opened 2025-03-28
layers: Add Parallel Hashmap includes to BUILD.gn

y-novikov:
Fixes #9790


spencer-lunarg (approved):

y-novikov:
@arno-lunarg @spencer-lunarg PTAL

y-novikov:
It sounds like we don't want this, because GN build doesn't define USE_CUSTOM_HASH_MAP by default like CMake build does.

PR #9816, opened 2025-03-28
gpuav: Auto dump invalid spirv

arno-lunarg:


layers/utils/shader_utils.cpp
```diff
@@ -187,7 +187,7 @@ void AdjustValidatorOptions(const DeviceExtensions &device_extensions, const Dev
 }
 
 // This is used to help dump SPIR-V while debugging intermediate phases of any altercations to the SPIR-V
-void DumpSpirvToFile(std::string file_name, const char *spirv_data, size_t spirv_size) {
-    std::ofstream debug_file(file_name, std::ios::out | std::ios::binary);
-    debug_file.write(spirv_data, spirv_size);
+void DumpSpirvToFile(const std::filesystem::path &file_path, const uint32_t *spirv, size_t spirv_dwords_count) {
```
spencer-lunarg, line 190: why does this need to consume a `std::filesystem::path` if you don't have the callers do anything with it

I rather you just do the `std::filesystem::absolute(` once in this function and leave all the callers to just use `std::string` to path in the file name

layers/gpuav/instrumentation/gpuav_shader_instrumentor.cpp
```diff
@@ -1215,25 +1217,32 @@ bool GpuShaderInstrumentor::InstrumentShader(const vvl::span<const uint32_t> &in
...
         std::string spirv_val_error;
-        if (!GpuValidateShader(out_instrumented_spirv, extensions.vk_khr_relaxed_block_layout,
-                               extensions.vk_ext_scalar_block_layout, target_env, spirv_val_error)) {
+
+        is_instrumented_spirv_valid = GpuValidateShader(out_instrumented_spirv, extensions.vk_khr_relaxed_block_layout,
+                                                        extensions.vk_ext_scalar_block_layout, target_env, spirv_val_error);
+        if (!is_instrumented_spirv_valid) {
+            const std::filesystem::path instrumented_spirv_file =
+                std::filesystem::absolute("dump_" + std::to_string(unique_shader_id) + "_after_invalid.spv");
+            DumpSpirvToFile(instrumented_spirv_file, out_instrumented_spirv.data(), out_instrumented_spirv.size());
```
spencer-lunarg, line 1231: did you want to have a `  if (!gpuav_settings.debug_dump_instrumented_shaders) {` check here and print out the `_before` as well?

spencer-lunarg, line 1231: did you want to have a `  if (!gpuav_settings.debug_dump_instrumented_shaders) {` check here and print out the `_before` as well?

layers/utils/shader_utils.cpp
arno-lunarg: I print the path in the internal error message

layers/gpuav/instrumentation/gpuav_shader_instrumentor.cpp
arno-lunarg: Ah yes!

layers/utils/shader_utils.cpp
arno-lunarg: I also want to be able to use it for debugging, and don't want to guess where my dumps ends up (is it the path based on the working dir, the exe dir?), so being able to directly give a fixed absolute path helps.
Also looking at the official doc they recommend using `namespace fs = std::filesystem`, I now do that, so basically you can go:
`DumpSpirvToFile(fs::absolute("foo.spv"), spirv.data(), spirv.size());`
Note that with the current code an equivalent is the longer:
`DumpSpirvToFile("foo.spv", reinterpret_cast<const char *>(spirv.data()), spirv.size() * sizeof(uint32_t));`
So overall this function feels easier to use?

spencer-lunarg (approved):

PR #9826, opened 2025-04-01
gpuav: Swap in Private Variable at Linking and multiple link functions

spencer-lunarg:
Some basic infrastructure used for upcoming PR to move error message to own dedicated function

This also adds a way to have a single GLSL file per pass where we have multiple functions that a Pass might need to link into


arno-lunarg (approved):

layers/gpuav/spirv/module.cpp
```diff
@@ -344,19 +354,9 @@ void Module::AddInterfaceVariables(uint32_t id, spv::StorageClass storage_class)
...
-uint32_t Module::GetLinkFunction(uint32_t& link_function_id, const OfflineLinkInfo& offline_info) {
-    if (link_function_id == 0) {
-        link_function_id = TakeNextId();
-        link_info_.emplace_back(LinkInfo{offline_info, link_function_id});
-    }
-    return link_function_id;
-}
-
 // Takes the current module and injects the function into it
 // This is done by first apply any new Types/Constants/Variables and then copying in the instructions of the Function
```
arno-lunarg, line 358:
```suggestion
// Link functions into this module
// First, any new Types/Constants/Variables are inserted, then functions' instructions 
```

layers/gpuav/spirv/module.cpp
```diff
@@ -461,12 +461,15 @@ void Module::LinkFunction(const LinkInfo& info) {
...
                 }
                 case SpvType::kStruct: {
-                    // For OpTypeStruct, we just add it regardless since low chance to find for the amount of time to search all
-                    // struct (which there can be quite a bit of)
-                    type_id = TakeNextId();
-                    new_inst->ReplaceResultId(type_id);
-                    new_inst->ReplaceLinkedId(id_swap_map);
-                    type_manager_.AddType(std::move(new_inst), spv_type).Id();
+                    // For OpTypeStruct, unless asked for, we just add it regardless since low chance to find for the amount of time
+                    // to search all struct (which there can be quite a bit of)
```
arno-lunarg, line 465:
```suggestion
                    // For OpTypeStruct, always add them.
                    // Only try to look for definition duplication if asked explicitly. 
                    // Chances of adding a struct already defined in original SPIR-V are low, and struct definitions exploration can be expensive
```

layers/gpuav/spirv/module.cpp
```diff
@@ -544,28 +547,34 @@ void Module::LinkFunction(const LinkInfo& info) {
...
-                assert(pointer_type && pointer_type->spv_type_ == SpvType::kInt);
-                if (pointer_type->spv_type_ == SpvType::kInt) {
-                    const uint32_t uint32_0_id = type_manager_.GetConstantZeroUint32().Id();
-                    new_inst->AppendWord(uint32_0_id);
+            if (storage_class == spv::StorageClassPrivate && ((info.module.flags & SwapPrivateVariable) != 0)) {
+                // Variable already is in shader, just mark the new result ID
+                AddInterfaceVariables(info.private_variable_id, storage_class);
+                id_swap_map[old_result_id] = info.private_variable_id;
+            } else {
+                // Add in all variables outside of functions
```
arno-lunarg, line 556: "outside of functions?"

spencer-lunarg: the variable doesn't live inside an `OpFunction` block... will try to reword

Issue #9764, opened 2025-03-24
Add command and debug label to post processing


spencer-lunarg:
We now collect the "action index" inside the post processing buffer (https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/fe25d5e4bb2663e2a0fcea929f28bdb879331af0/layers/gpuav/descriptor_validation/gpuav_descriptor_set.cpp#L326)

Now with this, we should be able to back track which `vkCmdDraw`/`vkCmdDispatc`/ etc the error came from and most importantly, we should be able to add tests to make sure we print out the correct debug region label as well inside the command buffere
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/fe25d5e4bb2663e2a0fcea929f28bdb879331af0/layers/gpuav/descriptor_validation/gpuav_descriptor_validation.cpp#L194

PR #9828, opened 2025-04-01
gpuav: Add debug labels to post processing

arno-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9764


layers/drawdispatch/descriptor_validator.h
```diff
@@ -85,7 +86,7 @@ class DescriptorValidator : public Logger {
     vvl::CommandBuffer& cb_state;
     vvl::DescriptorSet& descriptor_set;
     const VkFramebuffer framebuffer;
-    const Location& loc;
+    const Location* loc = nullptr;
```
spencer-lunarg, line 89: I hate this, not that you did this, but that we have to do this here... We have made `Location` a `const` to prevent things in Core Validation from accidentally altering it under you

Since Location are suppose to be cheap to create, could we just go `Location loc` and update it with a new Location each time? Everything inside `Location` is already const, so we don't need to make this field `const` because it clearly is not a `const` here anymore if we need to update it

layers/gpuav/descriptor_validation/gpuav_descriptor_set.cpp
```diff
@@ -323,7 +323,8 @@ DescriptorAccessMap DescriptorSetSubState::GetDescriptorAccesses(const Location
             const glsl::PostProcessDescriptorIndexSlot slot = slot_ptr[binding_layout.start + descriptor_i];
             if (slot.meta_data & glsl::kPostProcessMetaMaskAccessed) {
                 const uint32_t shader_id = slot.meta_data & glsl::kShaderIdMask;
-                const uint32_t action_index = slot.meta_data & glsl::kPostProcessMetaMaskActionIndex;
+                const uint32_t action_index =
```
spencer-lunarg, line 326: image

spencer-lunarg (approved):
LGTM with thoughts on needing the `Location` being a pointer

layers/gpuav/descriptor_validation/gpuav_descriptor_set.cpp
arno-lunarg: <( ‵□′)>───Ｃε(┬﹏┬)3

layers/drawdispatch/descriptor_validator.h
arno-lunarg: Yeah copying is probably better, I switched to that. Still, we don't deep copy the `Location` so we could have dangling pointers ...

layers/gpuav/descriptor_validation/gpuav_descriptor_validation.cpp
```diff
@@ -259,6 +259,15 @@ void UpdateBoundDescriptors(Validator &gpuav, CommandBufferSubState &cb_state, V
...
                     assert(context.shader_handle);
+
+                    std::string debug_region_name;
+                    if (auto found_label_cmd_i = label_logging.action_cmd_i_to_label_cmd_i_map.find(descriptor_access.action_index);
+                        found_label_cmd_i != label_logging.action_cmd_i_to_label_cmd_i_map.end()) {
+                        debug_region_name = GetDebugLabelRegion(found_label_cmd_i->second, label_logging.initial_label_stack);
+                    }
+
+                    Location access_loc(loc, debug_region_name);
+                    context.SetLocation(access_loc);
```
spencer-lunarg, line 270: why do we need a `SetLocation` helper here? We already direct set things like `context.set_index =`

spencer-lunarg: also adds an extra copy I think

layers/error_message/error_location.h
```diff
@@ -33,15 +33,16 @@ struct Location {
     static const uint32_t kNoIndex = vvl::kU32Max;
 
     // name of the vulkan function we're checking
-    const vvl::Func function{};
+    vvl::Func function{};
```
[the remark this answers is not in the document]
arno-lunarg: I removed those changes!

layers/gpuav/descriptor_validation/gpuav_descriptor_validation.cpp
arno-lunarg: We don't really need to expose the variable, only write to it, so I wanted to keep it `private`

spencer-lunarg: sure, then make `set_index` private as well

I don't mind the getter/setter, I mind have half the members use getter/setters and half being public exposed

layers/drawdispatch/descriptor_validator.h
```diff
@@ -48,6 +48,7 @@ class DescriptorValidator : public Logger {
     // The main reason we can't combine is one function needs to be const and the other is non-const.
     bool ValidateBindingDynamic(const spirv::ResourceInterfaceVariable& binding_info, DescriptorBinding& binding,
                                 const uint32_t index);
+    void SetLocation(const Location& loc) { this->loc = LocationCapture(loc); }
 
     // For GPU-AV, these can become aliased and need to be mutable between descriptor accesses
     uint32_t set_index;
```
spencer-lunarg, line 54: maybe just have a `SetThingsForGPUAV` setter for all of these and move these back as private

spencer-lunarg (approved):

PR #9837, opened 2025-04-02
Validate shader object nextStage

### LINKED ISSUE #9823: VkShaderCreateInfoEXT.nextStage is not validated
When using shader objects, it is mandatory to specify the possible next stages of each shader using `VkShaderCreateInfoEXT.nextStage` :

> The stage of the shader being compiled is specified by stage. Applications must also specify which stage types will be allowed to immediately follow the shader being created. For example, a vertex shader might specify a nextStage value of VK_SHADER_STAGE_FRAGMENT_BIT to indicate that the vertex shader being created will always be followed by a fragment shader (and never a geometry or tessellation shader). Applications that do not know this information at shader creation time or need the same shader to be compatible with multiple subsequent stages can specify a mask that includes as many valid next stages as they wish. For example, a vertex shader can specify a nextStage mask of VK_SHADER_STAGE_GEOMETRY_BIT | VK_SHADER_STAGE_FRAGMENT_BIT to indicate that the next stage could be either a geometry shader or fragment shader (but not a tessellation shader).

https://docs.vulkan.org/features/latest/features/proposals/VK_EXT_shader_object.html

This does not seem to be currently tested by the validation layer. Using RADV with a graphic pipeline consisting of vertex, geometry, and fragment shaders without specifying `VkShaderCreateInfoEXT.nextStage` causes a segfault on the draw call without triggering the validation layer.

ziga-lunarg:
Closes #9823

### DIFF
--- a/layers/core_checks/cc_shader_object.cpp
+++ b/layers/core_checks/cc_shader_object.cpp
@@ -25,30 +25,28 @@
 #include "drawdispatch/drawdispatch_vuids.h"
 #include "containers/limits.h"
 
-VkShaderStageFlags FindNextStage(uint32_t createInfoCount, const VkShaderCreateInfoEXT* pCreateInfos, VkShaderStageFlagBits stage) {
-    constexpr uint32_t graphics_stages_count = 5;
-    constexpr uint32_t mesh_stages_count = 3;
-    const VkShaderStageFlagBits graphics_stages[graphics_stages_count] = {
-        VK_SHADER_STAGE_VERTEX_BIT, VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT, VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT,
-        VK_SHADER_STAGE_GEOMETRY_BIT, VK_SHADER_STAGE_FRAGMENT_BIT};
-    const VkShaderStageFlagBits mesh_stages[mesh_stages_count] = {VK_SHADER_STAGE_TASK_BIT_EXT, VK_SHADER_STAGE_MESH_BIT_EXT,
-                                                                  VK_SHADER_STAGE_FRAGMENT_BIT};
+// In order of how stages are linked together
+static const std::array graphics_stages = {VK_SHADER_STAGE_VERTEX_BIT, VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT,
+                                           VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT, VK_SHADER_STAGE_GEOMETRY_BIT,
+                                           VK_SHADER_STAGE_FRAGMENT_BIT};
+static const std::array mesh_stages = {VK_SHADER_STAGE_TASK_BIT_EXT, VK_SHADER_STAGE_MESH_BIT_EXT, VK_SHADER_STAGE_FRAGMENT_BIT};
 
-    uint32_t graphic_index = graphics_stages_count;
-    uint32_t mesh_index = mesh_stages_count;
-    for (uint32_t i = 0; i < graphics_stages_count; ++i) {
+VkShaderStageFlags FindNextStage(uint32_t createInfoCount, const VkShaderCreateInfoEXT* pCreateInfos, VkShaderStageFlagBits stage) {
+    uint32_t graphic_index = static_cast<uint32_t>(graphics_stages.size());
+    uint32_t mesh_index = static_cast<uint32_t>(mesh_stages.size());
+    for (uint32_t i = 0; i < graphics_stages.size(); ++i) {
         if (graphics_stages[i] == stage) {
             graphic_index = i;
             break;
         }
-        if (i < mesh_stages_count && mesh_stages[i] == stage) {
+        if (i < mesh_stages.size() && mesh_stages[i] == stage) {
             mesh_index = i;
             break;
         }
     }
 
-    if (graphic_index < graphics_stages_count) {
-        while (++graphic_index < graphics_stages_count) {
+    if (graphic_index < graphics_stages.size()) {
+        while (++graphic_index < graphics_stages.size()) {
             for (uint32_t i = 0; i < createInfoCount; ++i) {
                 if (pCreateInfos[i].stage == graphics_stages[graphic_index]) {
                     return graphics_stages[graphic_index];
@@ -56,7 +54,7 @@ VkShaderStageFlags FindNextStage(uint32_t createInfoCount, const VkShaderCreateI
             }
         }
     } else {
-        while (++mesh_index < mesh_stages_count) {
+        while (++mesh_index < mesh_stages.size()) {
             for (uint32_t i = 0; i < createInfoCount; ++i) {
                 if (pCreateInfos[i].stage == mesh_stages[mesh_index]) {
                     return mesh_stages[mesh_index];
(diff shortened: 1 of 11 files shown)


layers/core_checks/cc_shader_object.cpp
```diff
@@ -558,6 +558,53 @@ bool CoreChecks::PreCallValidateGetShaderBinaryDataEXT(VkDevice device, VkShader
...
+    }
+    ShaderObjectStage previous_stage = ShaderObjectStage::LAST;
+    for (const auto& stage : stages) {
+        if (last_bound_state.IsValidShaderBound(stage)) {
+            const auto previous_state = last_bound_state.GetShaderStateIfValid(previous_stage);
+            if (previous_stage != ShaderObjectStage::LAST &&
+                (previous_state->create_info.flags & VK_SHADER_CREATE_LINK_STAGE_BIT_EXT) == 0 &&
+                (previous_state->create_info.nextStage & ShaderObjectStageToVkShaderStage(stage)) == 0) {
+                const auto state = last_bound_state.GetShaderStateIfValid(stage);
+                skip |= LogError(vuid.pipeline_or_shaders_bound_08607, last_bound_state.cb_state.Handle(), vuid.loc(),
```
spencer-lunarg, line 589: my only real concern/question is why we now have 2 large checks for `08607`... I don't think anything to solve here, but it causes these types of gaps... might be a spec issue

Like I get this error

>  vkCmdDraw(): The combination of shader objects bound to VK_PIPELINE_BIND_POINT_GRAPHICS is invalid, because shader stages VK_SHADER_STAGE_VERTEX_BIT (VkShaderEXT 0x70000000007) and VK_SHADER_STAGE_FRAGMENT_BIT (VkShaderEXT 0x80000000008) are bound with no other stages between them, but VK_SHADER_STAGE_VERTEX_BIT was created with nextStage VkShaderStageFlags(0).

I am still confused what is suppose to be done or not... do I need a non-zero `nextStage`? Do I need to bind another shader?

layers/core_checks/cc_shader_object.cpp
```diff
@@ -558,6 +558,53 @@ bool CoreChecks::PreCallValidateGetShaderBinaryDataEXT(VkDevice device, VkShader
     return skip;
 }
 
+bool CoreChecks::ValidateShaderObjectNextStage(const LastBound& last_bound_state, const VkPipelineBindPoint bind_point,
+                                               const vvl::DrawDispatchVuid& vuid) const {
+    bool skip = false;
+
+    std::vector<ShaderObjectStage> stages;
```
spencer-lunarg, line 565: can we just have 2 `const static std::array` and then pick which one to loop

also we have a 

```
    const std::array graphics_stages = {VK_SHADER_STAGE_VERTEX_BIT, VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT,
                                        VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT, VK_SHADER_STAGE_GEOMETRY_BIT,
                                        VK_SHADER_STAGE_FRAGMENT_BIT};
```

already in this file, maybe make this global to this file as we seem to need these list a lot

layers/core_checks/cc_shader_object.cpp
```diff
@@ -558,6 +558,53 @@ bool CoreChecks::PreCallValidateGetShaderBinaryDataEXT(VkDevice device, VkShader
...
+    } else {
+        stages = {
+            ShaderObjectStage::TASK,
+            ShaderObjectStage::MESH,
+            ShaderObjectStage::FRAGMENT,
+        };
+    }
+    ShaderObjectStage previous_stage = ShaderObjectStage::LAST;
+    for (const auto& stage : stages) {
+        if (last_bound_state.IsValidShaderBound(stage)) {
```
spencer-lunarg, line 583:
```suggestion
        if (!last_bound_state.IsValidShaderBound(stage)) continue;
```

tests/framework/shader_object_helper.cpp
```diff
@@ -13,13 +13,55 @@
 #include "shader_object_helper.h"
 #include <vulkan/utility/vk_struct_helper.hpp>
 
+void SetNextStage(VkShaderCreateInfoEXT& info) {
```
spencer-lunarg, line 16: why do we have this and `Shader::SetNextStage` ... they are similar but different?

layers/core_checks/cc_shader_object.cpp
```diff
@@ -558,6 +558,53 @@ bool CoreChecks::PreCallValidateGetShaderBinaryDataEXT(VkDevice device, VkShader
...
+            const auto previous_state = last_bound_state.GetShaderStateIfValid(previous_stage);
```
spencer-lunarg, line 584:
```suggestion
            const auto previous_state = last_bound_state.GetShaderStateIfValid(previous_stage);
            ASSERT_AND_CONTINUE(previous_state);
```

layers/core_checks/cc_shader_object.cpp
```diff
@@ -558,6 +558,53 @@ bool CoreChecks::PreCallValidateGetShaderBinaryDataEXT(VkDevice device, VkShader
     return skip;
```
spencer-lunarg, line 558: not sure where to put, but I fail

```
[  FAILED  ] NegativeShaderObject.MissingCmdSetRasterizationStreamEXT
[  FAILED  ] NegativeShaderObject.MissingLineStippleEnable
[  FAILED  ] NegativeShaderObject.SetPrimitiveTopologyNonPatch
[  FAILED  ] PositiveShaderObject.DrawWithAllGraphicsShaderStagesUsed
[  FAILED  ] PositiveShaderObject.IndirectDraw
```

on ToT Lavapipe with this change (getting the new error)

layers/core_checks/cc_shader_object.cpp
```diff
@@ -558,6 +558,53 @@ bool CoreChecks::PreCallValidateGetShaderBinaryDataEXT(VkDevice device, VkShader
...
+        if (last_bound_state.IsValidShaderBound(stage)) {
+            const auto previous_state = last_bound_state.GetShaderStateIfValid(previous_stage);
+            if (previous_stage != ShaderObjectStage::LAST &&
+                (previous_state->create_info.flags & VK_SHADER_CREATE_LINK_STAGE_BIT_EXT) == 0 &&
+                (previous_state->create_info.nextStage & ShaderObjectStageToVkShaderStage(stage)) == 0) {
+                const auto state = last_bound_state.GetShaderStateIfValid(stage);
+                skip |= LogError(vuid.pipeline_or_shaders_bound_08607, last_bound_state.cb_state.Handle(), vuid.loc(),
+                                 "The combination of shader objects bound to %s is invalid, because shader stages %s (%s) and "
+                                 "%s (%s) are bound with no other stages between them, but %s was created with nextStage %s.",
+                                 string_VkPipelineBindPoint(bind_point),
```
spencer-lunarg, line 592: can this not be Graphics? If it can be compute, write a test, otherwise just hardcode to Graphics

spencer-lunarg (changes requested):

layers/core_checks/cc_shader_object.cpp
ziga-lunarg: I agree there could be a more specific VU for this, but at the moment there is not

tests/framework/shader_object_helper.cpp
ziga-lunarg: They are similar, but one knows if tessellation and geometry shader features are enabled, and the other does not

layers/core_checks/cc_shader_object.cpp
ziga-lunarg: Only graphics pipeline bind point applies here, done.

tests/framework/shader_object_helper.cpp
spencer-lunarg: can we either 
- get it down to 1 (with some param)
- Create different names
- add comment why we have 2 in 2 different helpers

layers/core_checks/cc_shader_object.cpp
spencer-lunarg: I guess my question/issue is reading this VU message, it doesn't tell me what is wrong... like it would be nice here to explain if we detect `nextStage == 0` what the user likely might need to do

Change the `nextStage` or bind other stages? (I realize the issue reference the extension proposal)

ziga-lunarg: I've updated the message

spencer-lunarg (approved):
Thanks, this is now a lot cleaner and will test on RADV for my issues once this is in

Issue #9802, opened 2025-03-27
vkQueuePresentKHR semaphores being signaled again before wait doesn't trigger validation error.


danginsburg:
I just spent a bunch of time on debugging a GPU deadlock that didn't give a validation error (well, to be fair, it didn't reproduce with validation on unless I turned off a lot of checks, so I can't be sure validation wouldn't have caught it).   To summarize the problem, I have a sequence like the following:

vkAcquireNextImage -> Sem A
vkQueueSubmit(Graphics) -> Wait Sem A / Signal Sem B, Signal Sem C
vkQueueSubmit(Compute) -> Wait Sem B / Signal Sem D
vkQueuePresentKHR -> Wait Sem C, Wait Sem D

In the next frame, I signal Sem C and Sem D again in vkQueueSubmit and will wait for them in vkQueuePresentKHR.  For what it's worth, if you ignore the Graphics/Compute aspect, this is actually how Dota/CS2 currently work as well, which has never caused a visible problem nor failed validation, but when interacting with Graphics/Compute it is causing a deadlock on NVIDIA (both Windows + Linux).

I believe the problem is basically described here: https://github.com/KhronosGroup/Vulkan-Docs/issues/2007.  I need to wait on a fence in vkQueuePresentKHR (by VK_EXT_swapchain_maintenance1) before recycling sems C/D.  Or, I need to wait until I've waited on the vkAcquireNextImageKHR again.

Assuming my understanding here is correct, it seems that the current behavior even of Dota/CS2 is invalid and should be caught by validation, but it isn't.  It would be good if validation could catch this, given how tricky it is.

artem-lunarg:
@danginsburg We merged the fix. Feel free to  close the ticket if the solution works for you.

> Or, I need to wait until I've waited on the vkAcquireNextImageKHR again.

In a few threads I saw a comment about waiting on the fence form Acquire, and maybe it can be useful in some scenarios. But also in common frame organization setups often it's not needed and instead to use *semaphore* from Acquire and wait on it in Submit call. That's in addition to having a *present* semaphore per swapchain image. The reason we can do this i because VUIDs like `VUID-vkQueueSubmit-pSignalSemaphores-00067` specify that semaphore should be unsignaled when `the semaphore signal operation it defines is executed on the device` and not during submission time. So it's enough to wait on semaphore from Acquire in one of the submits. Maybe there are scenarios when it's easier to wait on the Acquire fence on CPU side.

PR #9840, opened 2025-04-02
layers: Track semaphore is in use by swapchain

artem-lunarg:
Addresses https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9802

The main idea is to set a boolean flag as part of the semaphore state between QueuePresent and the next AcquireNextImage that returns the same image. If any QueueSubmit uses this semaphore as a signal and the semaphore's flag is set then it means  it can still be used by the presentation.

### DIFF
--- a/layers/core_checks/cc_synchronization.cpp
+++ b/layers/core_checks/cc_synchronization.cpp
@@ -243,6 +243,19 @@ bool SemaphoreSubmitState::ValidateSignalSemaphore(const Location &signal_semaph
                         vuid, objlist, signal_semaphore_loc,
                         "(%s) is being signaled by %s, but it was previously signaled by %s and has not since been waited on",
                         core.FormatHandle(handle).c_str(), core.FormatHandle(queue).c_str(), initiator.str().c_str());
+                } else if (semaphore_state.IsInUseBySwapchain()) {
+                    const auto &vuid = GetQueueSubmitVUID(signal_semaphore_loc, sync_vuid_maps::SubmitError::kSemAlreadySignalled);
+                    skip |= core.LogError(
+                        vuid, objlist, signal_semaphore_loc,
+                        "(%s) is being signaled by %s, but it may still be in use by the swapchain because the "
+                        "corresponding swapchain image has not yet been re-acquired.\n"
+                        "Vulkan insight: Common methods to ensure that a semaphore passed to vkQueuePresentKHR is not in "
+                        "use and can be safely reused:\n"
+                        "\ta) Use a separate semaphore per swapchain image. Index these semaphores using the "
+                        "index of the acquired image.\n"
+                        "\tb) Consider the VK_EXT_swapchain_maintenance1 extension. It allows using VkFence with the presentation "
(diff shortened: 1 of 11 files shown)


jeremyg-lunarg (approved):

artem-lunarg:
Keeping this for a bit to do some apps testing

layers/core_checks/cc_synchronization.cpp
```diff
@@ -243,6 +243,19 @@ bool SemaphoreSubmitState::ValidateSignalSemaphore(const Location &signal_semaph
                         vuid, objlist, signal_semaphore_loc,
                         "(%s) is being signaled by %s, but it was previously signaled by %s and has not since been waited on",
                         core.FormatHandle(handle).c_str(), core.FormatHandle(queue).c_str(), initiator.str().c_str());
+                } else if (semaphore_state.IsInUseBySwapchain()) {
+                    const auto &vuid = GetQueueSubmitVUID(signal_semaphore_loc, sync_vuid_maps::SubmitError::kSemAlreadySignalled);
+                    skip |= core.LogError(
+                        vuid, objlist, signal_semaphore_loc,
+                        "(%s) is being signaled by %s, but it may still be in use by the swapchain because the "
+                        "corresponding swapchain image has not yet been re-acquired.\n"
+                        "Vulkan insight: Common methods to ensure that a semaphore passed to vkQueuePresentKHR is not in "
```
spencer-lunarg, line 252:
```suggestion
                        "Vulkan insight: you are doing crazy stuff, please stop, you will regret things, keep it simple! "
```

spencer-lunarg (approved):

layers/core_checks/cc_synchronization.cpp
artem-lunarg: Well, didn't know how to say exactly this. Looks like this can affect some people (many?), that's  why additional section.

spencer-lunarg: I'm all for it!!! I can just feel the "I've spent hours making sure this is correct according to the spec" energy in this error message

artem-lunarg:
Something is not right, need to investigate.

artem-lunarg:
It was bug in my own program. I expected it can have this issue, but I did not expect it can have issue in swapchain configuration with only 2 images (I expected issues when swapchain has 3 images and the app double buffer other frame resources). It turns out that even of 2-image swapchain vkAcquireNextImage can return the same image index multiple times in a row. It looks like this happens on the first frames, maybe the dricer/OS does some warp up.

But in the end VVL caught the issue. I fixed it, solution is simple: https://github.com/kennyalive/vulkan-base/commit/27bcaad9d519cc2f9c5cde4872742d4a5212eee6. The intuition many people can find similar bug but hopefully the fix is relatively simple. Probably many people's will have similar reaction it's somethign with VVL!

artem-lunarg:
Will make an attempt to add information about acquired image indices and maybe detect some mapping of indices to semaphores to provide additional hints.

PR #9841, opened 2025-04-02
gpuav: Create a single LogError for passes

spencer-lunarg:
This creates a single `LogError` like function at the end so instead of looking like

```glsl
void inst_oob_check(uint a, uint b) {
    uint error = 0;
    if (/*bad a and b*/) {
        error = kErrorName;
    }

    if (error) {
        // ... error logging
    }
}

void main() {
    inst_oob_check(index, x); 
    index[x] = 0;
        
    inst_oob_check(index, y); 
    index[y] = 0;
    
    inst_oob_check(index, z); 
    index[z] = 0;
}
```

it now looks like

```glsl
struct ErrorPayload {
    uint payload;
} error_payload;

void inst_oob_check(uint a, uint b) {
    if (/*bad a and b*/) {
        error_payload = ErrorPayload(kErrorName);
    }
}

void inst_log_error() {
    if (error_payload) {
        // ... error logging
    }
}

void main() {
   error_payload = ErrorPayload(0); // clear at start of entry point

    inst_oob_check(index, x); 
    index[x] = 0;
        
    inst_oob_check(index, y); 
    index[y] = 0;
    
    inst_oob_check(index, z); 
    index[z] = 0;

    inst_log_error();
}
```

From testing, this does give an nice improvement in compile time (around like 40% on Mesa) (at the cost of having only 1 error reported)

Found the time saved for Safe Mode is very drastic (like 3x) so definitely a improvement


layers/gpuav/instrumentation/gpuav_instrumentation.cpp
```diff
@@ -807,16 +815,17 @@ bool LogMessageInstBufferDeviceAddress(const uint32_t *error_record, std::string
     bool error_found = true;
     std::ostringstream strm;
 
-    const uint32_t payload = error_record[kInstBuffAddrAccessPayloadOffset];
+    const uint32_t payload = error_record[kInstLogErrorParameterOffset_2];
     const bool is_write = ((payload >> kInstBuffAddrAccessPayloadShiftIsWrite) & 1) != 0;
     const bool is_struct = ((payload >> kInstBuffAddrAccessPayloadShiftIsStruct) & 1) != 0;
 
+    uint64_t address = *reinterpret_cast<const uint64_t *>(error_record + kInstLogErrorParameterOffset_0);
```
arno-lunarg, line 822:
```suggestion
    const uint64_t address = *reinterpret_cast<const uint64_t *>(error_record + kInstLogErrorParameterOffset_0);
```

layers/gpuav/spirv/log_error_pass.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+    const uint32_t function_result = module_.TakeNextId();
+    const uint32_t void_type = module_.type_manager_.GetTypeVoid().Id();
+    block.CreateInstruction(spv::OpFunctionCall, {void_type, function_result, link_function_id_, stage_info_id}, inst_it);
+}
+
+// This is a hard thing to fully get right, as things like OpTerminateInvocation/OpKill can end the invocation, but not the shader.
+// The main thing we are trying to find here is if this is our last chance to report an error message
+bool LogErrorPass::IsShaderExiting(const Instruction& inst) const {
+    return inst.Opcode() == spv::OpReturn || inst.Opcode() == spv::OpReturnValue;
```
arno-lunarg, line 90: If say all fragments within a fragment shader invocation call discard, will we miss the opportunity to log an error?

layers/state_tracker/descriptor_sets.h
```diff
@@ -936,6 +936,8 @@ class DescriptorSet : public StateObject, public SubStateManager<DescriptorSetSu
     // The caller has to ensure that binding has dynamic descriptor type.
     uint32_t GetDynamicOffsetIndexFromBinding(uint32_t dynamic_binding) const;
 
+    std::pair<uint32_t, uint32_t> GetBinidngAndIndex(const uint32_t global_descriptor_index) const;
```
arno-lunarg, line 939:
```suggestion
    std::pair<uint32_t, uint32_t> GetBindingAndIndex(const uint32_t global_descriptor_index) const;
```

arno-lunarg (approved):

layers/gpuav/spirv/log_error_pass.cpp
spencer-lunarg: well.... see this is why I am avoiding this, so what we can (and we have ZERO tests for this anywhere) is check for `OpTerminateInvocation` and warn the users... or we could add it before that spot, again, would need to test... will create an issue to track

spencer-lunarg: 100% a "not worth the time and will just accept some silent errors" for the short term

PR #9845, opened 2025-04-03
state_tracker: Use ! instead of not in HasSubState template

jeremyg-lunarg:
This appeared to be causing problems with some MSVC 32 bit builds.


spencer-lunarg (approved):
!

PR #9846, opened 2025-04-03
docs: Add chassis and shared state tracking documentation

jeremyg-lunarg:


docs/chassis.md
```diff
@@ -0,0 +1,60 @@
+## Introduction
+
+The Vulkan-ValidationLayers project has had its architecture evolve several times since it was started. Initially, validation was implemented as independent layers, which could be enabled using the vulkan loader.  This design had high overhead and was difficult to test since the layers could be enabled in any order, which could cause very different behaviors. 
+
+So all of the separate layers were combined into a single layer. The original layers were now called validation objects. A component called the chassis was made which intercepts vulkan commands and let validation objects do validation and track state. Several of the validation objects used common code to do state tracking, but since state tracking was implemented in an abstract base class, each validation object had its own copy of the state data. This was redundant both for memory and CPU overhead.
+
+Recently, state tracking was changed so that all validation objects now share a single state tracker. This phase of the architecture is described here.
```
spencer-lunarg, line 7: maybe quick mention this was done to both reduce memory and allow sharing state between Validation Objects

spencer-lunarg (approved):

docs/chassis.md
```diff
@@ -0,0 +1,60 @@
+## Introduction
+
+The Vulkan-ValidationLayers project has had its architecture evolve several times since it was started. Initially, validation was implemented as independent layers, which could be enabled using the vulkan loader.  This design had high overhead and was difficult to test since the layers could be enabled in any order, which could cause very different behaviors. 
+
+So all of the separate layers were combined into a single layer. The original layers were now called validation objects. A component called the chassis was made which intercepts vulkan commands and let validation objects do validation and track state. Several of the validation objects used common code to do state tracking, but since state tracking was implemented in an abstract base class, each validation object had its own copy of the state data. This was redundant both for memory and CPU overhead.
```
artem-lunarg, line 5: "abstract base class" - nit and it could be just me,  but check if we can go without word "abstract". Some people can associate abstract with more stateless than stateful class, so potential confusion with "each validation object had its own copy of the state data" as in is this data mostly defined by the derived classes or the abstract base class defined some state.

docs/chassis.md
```diff
@@ -0,0 +1,58 @@
+## Introduction
+
+The Vulkan-ValidationLayers project has had its architecture evolve several times since it was started. Initially, validation was implemented as independent layers, which could be enabled using the vulkan loader.  This design had high overhead and was difficult to test since the layers could be enabled in any order, which could cause very different behaviors.
```
lunarpapillo, line 3: vulkan loader => Vulkan loader

docs/chassis.md
```diff
@@ -0,0 +1,58 @@
+## Introduction
+
+The Vulkan-ValidationLayers project has had its architecture evolve several times since it was started. Initially, validation was implemented as independent layers, which could be enabled using the vulkan loader.  This design had high overhead and was difficult to test since the layers could be enabled in any order, which could cause very different behaviors. 
+
+So all of the separate layers were combined into a single layer. The original layers were now called validation objects. A component called the chassis was made which intercepts vulkan commands and let validation objects do validation and track state. Several of the validation objects used common code to do state tracking, but since state tracking was implemented in a base class, each validation object had its own copy of the state data. This was redundant both for memory and CPU overhead.
```
lunarpapillo, line 5: vulkan commands => Vulkan commands

docs/chassis.md
```diff
@@ -0,0 +1,58 @@
...
+    return
+for all validation objects:
+	PreCallRecordFoo()
+// Call into the ICD, wrapping and unwrapping Vulkan handles (if enabled)
+DispatchFoo()
+for all validation objects:
+    PostCallRecordFoo()
+```
+
+![](images/chassis-class-interaction.png)
```
lunarpapillo, line 29: In the future... GitHub supports inline Mermaid diagrams, which can render images like this in a way that make it easier to change the underlying structure when necessary.

PR #9818, opened 2025-03-28
gpuav: Fix crash with GPL

spencer-lunarg:
Fixes many random crashes I was getting, normally when reducing certain instrumentation settings to suddenly find our GPL flow blowing up

PR #9847, opened 2025-04-03
gpuav: Fix crash with GPL

spencer-lunarg:
redo of https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9818 but found the "other" issue (wrote a test) and should be good to go now


layers/gpuav/instrumentation/gpuav_shader_instrumentor.cpp
```diff
@@ -1030,30 +1047,32 @@ bool GpuShaderInstrumentor::PreCallRecordPipelineCreationShaderInstrumentationGP
...
-        if (result != VK_SUCCESS || instrumented_pipeline_lib == VK_NULL_HANDLE) {
-            InternalError(device, loc, "Failed to create instrumentde pipeline library.");
-            return false;
-        }
+        // Create instrumented pipeline library if we have instrumented one of the libraries inside of it
+        if (need_new_pipeline) {
+            VkPipeline instrumented_pipeline_lib = VK_NULL_HANDLE;
+            const VkResult result = DispatchCreateGraphicsPipelines(device, VK_NULL_HANDLE, 1, modified_pipeline_ci.ptr(),
+                                                                    pAllocator, &instrumented_pipeline_lib);
+            if (result != VK_SUCCESS || instrumented_pipeline_lib == VK_NULL_HANDLE) {
```
arno-lunarg, line 1062: Can we have a success and no handle?

layers/gpuav/instrumentation/gpuav_shader_instrumentor.cpp
```diff
@@ -1030,30 +1047,32 @@ bool GpuShaderInstrumentor::PreCallRecordPipelineCreationShaderInstrumentationGP
...
+                InternalError(device, loc, "Failed to create instrumentde pipeline library.");
```
arno-lunarg, line 1063:
```suggestion
                InternalError(device, loc, "Failed to recreate instrumented pipeline library.");
```
(not sure about `recreate`?)

tests/unit/gpu_av_descriptor_indexing.cpp
```diff
@@ -4294,4 +4294,87 @@ TEST_F(NegativeGpuAVDescriptorIndexing, OpInBoundsAccessChain) {
...
+    VkPipeline libraries[2] = {
+        combined_lib.Handle(),
+        frag_out_lib.Handle(),
+    };
+    VkPipelineLibraryCreateInfoKHR link_info = vku::InitStructHelper();
+    link_info.libraryCount = size32(libraries);
+    link_info.pLibraries = libraries;
+
+    // Destroy VkShaderModule as not required to have when linking
+    vs.destroy();
```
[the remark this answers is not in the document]
spencer-lunarg: yes... if they provide the module name, we will need to tie it to the vvl::Pipeline state object

(haven't looked at your PR yet) but i would imagine for GPL it would be the same thing, since we wait until the final pipeline is built, we might need to track names for GPL Libraries and Shader Modules being used to build it

layers/gpuav/instrumentation/gpuav_shader_instrumentor.cpp
spencer-lunarg: that would be a hard driver bug

spencer-lunarg: I just realize we do the double check here... I guess being over cautious, will add comment saying that

PR #9850, opened 2025-04-04
scripts: Warning and misc cleanup

spencer-lunarg:
So misc cleanup from my Ruff linter, also inline the large strings to make the python code more easy to follow indentation


scripts/generators/spirv_tool_commit_id_generator.py
```diff
@@ -34,7 +34,7 @@ def generate(self):
             commit_id = [x for x in json_data['repos'] if x['name'] == 'SPIRV-Tools'][0]['commit']
 
         try:
-            str_as_int = int(commit_id, 16)
+            int(commit_id, 16)
```
jeremyg-lunarg, line 37: I guess python doesn't optimize this away?

jeremyg-lunarg (approved):

Issue #9771, opened 2025-03-24
Selective shaders based on debug handle


spencer-lunarg:
Have an option to pass in your debug handle for a pipeline/shaderModule/shaderObject and run selective shader instrumentation on those. This would make it easier to have people just list handle names and we could use that. This would make it faster as you would not need to re-compile the application

PR #9855, opened 2025-04-04
gpuav: Upgrade selective shader instrumentation

arno-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9771

Add the possibility to selectively instrument shaders with a debug name matching any regex from a list.

Still need to add support for shader objects, as the flow will be different
Edit: Also no support for GPL when shader modules have been destroyed before linking - we will miss instrumentation in this case

### DIFF
--- a/layers/gpuav/core/gpuav_settings.h
+++ b/layers/gpuav/core/gpuav_settings.h
@@ -17,6 +17,15 @@
  */
 #pragma once
 #include <cstdint>
+// Fix GCC 13 issues with regex
+#if defined(__GNUC__) && (__GNUC__ > 12)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+#endif
+#include <regex>
+#if defined(__GNUC__) && (__GNUC__ > 12)
+#pragma GCC diagnostic pop
+#endif
 // Default values for those settings should match layers/VkLayer_khronos_validation.json.in
 
 struct GpuAVSettings {
@@ -25,6 +34,7 @@ struct GpuAVSettings {
     bool force_on_robustness = false;
     uint32_t max_bda_in_use = 10000;
     bool select_instrumented_shaders = false;
+    std::vector<std::regex> shader_selection_regexes{};
 
     bool validate_indirect_draws_buffers = true;
     bool validate_indirect_dispatches_buffers = true;
@@ -80,6 +90,12 @@ struct GpuAVSettings {
         validate_indirect_trace_rays_buffers = enabled;
         validate_buffer_copies = enabled;
         validate_index_buffers = enabled;
+    }
+    void GetShaderSelectionRegexes(const std::vector<std::string>& instrumented_shaders) {
+        for (const std::string& shader_name_pattern : instrumented_shaders) {
+            shader_selection_regexes.emplace_back(
+                std::regex(shader_name_pattern, std::regex_constants::ECMAScript | std::regex_constants::optimize));
+        }
     }
 
     // For people who are using VkValidationFeatureEnableEXT to set only DebugPrintf (and want the rest of GPU-AV off)
--- a/tests/unit/gpu_av.cpp
+++ b/tests/unit/gpu_av.cpp
@@ -113,6 +113,146 @@ TEST_F(NegativeGpuAV, SelectInstrumentedShaders) {
     m_errorMonitor->VerifyFound();
 }
 
+TEST_F(NegativeGpuAV, SelectInstrumentedShadersRegex) {
+    TEST_DESCRIPTION(
+        "Selectively instrument shaders for validation, using regexes: all shaders matching regexes must be instrumented. Here it "
+        "means they should emit errors");
+    SetTargetApiVersion(VK_API_VERSION_1_2);
+    AddRequiredExtensions(VK_EXT_DEBUG_UTILS_EXTENSION_NAME);
+    AddRequiredFeature(vkt::Feature::vertexPipelineStoresAndAtomics);
+    std::vector<VkLayerSettingEXT> layer_settings(2);
+    layer_settings[0] = {OBJECT_LAYER_NAME, "gpuav_select_instrumented_shaders", VK_LAYER_SETTING_TYPE_BOOL32_EXT, 1, &kVkTrue};
+    std::array<const char *, 2> shader_regexes = {{"vertex_foo", "fragment_.*"}};
+    layer_settings[1] = {OBJECT_LAYER_NAME, "gpuav_shaders_to_instrument", VK_LAYER_SETTING_TYPE_STRING_EXT, size32(shader_regexes),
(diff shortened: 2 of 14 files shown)


tests/unit/gpu_av.cpp
```diff
@@ -115,6 +115,76 @@ TEST_F(NegativeGpuAV, SelectInstrumentedShaders) {
     m_errorMonitor->VerifyFound();
```
spencer-lunarg, line 115: Just dumping thoughts out loud, for things we should test (without looking what you added yet)... and maybe some are redundant 

- [ ] They use `vkCreateShaderModule`, set name to `VkShaderModule`
- [ ] They use `vkCreateShaderModule`, set name to `VkShaderModule`, but destroy after creating pipeline
- [ ] They use `vkCreateShaderModule`, set name to `VkPipeline`
- [ ] They use `VK_KHR_maintenance5` to "inline", set name to `VkPipeline`
- [ ] They use GPL, name the `VkShaderModule` in a library and destroy it
- [ ] They use GPL, name the `VkPipeline` of the library (not the final pipeline)
- [ ] They use Shader Objects and set name to `VkShaderEXT`

arno-lunarg: > They use vkCreateShaderModule, set name to VkShaderModule
They use vkCreateShaderModule, set name to VkShaderModule, but destroy after creating pipeline

Will add test for 2nd case, it should not have an impact

> They use vkCreateShaderModule, set name to VkPipeline
They use VK_KHR_maintenance5 to "inline", set name to VkPipeline

You set a VkPipeline name after creating it, so after instrumentation, can't easily go back to instrument?

> They use GPL, name the VkShaderModule in a library and destroy it

Not handled, I think I need new state tracking to do that? Since we instrument at pipeline linking time, at this point shader names will be lost along their corresponding module

> They use GPL, name the VkPipeline of the library (not the final pipeline)

I can add that

> They use Shader Objects and set name to VkShaderEXT

Again issue of "object is instrumented at its creation time, so it can't have a debug name yet", would probably need a new system that upon first shader object usage, tries to see if it's eventual debug name appears in the list of shaders to instrument, and replace it with a newly instrumented shader?

spencer-lunarg: > You set a VkPipeline name after creating it, so after instrumentation, can't easily go back to instrument?

Ya, this one is tricky, we will need to re-create the pipeline... with handle wrapping, we can swap it.. maybe we might need to carry both handles around

> Not handled, I think I need new state tracking to do that? Since we instrument at pipeline linking time, at this point shader names will be lost along their corresponding module

We know we need to save it though... also we save the underlying SPIR-V for the pipeline, that is how we are able to do this at all right now without the selective shader

.github/workflows/vvl.yml
```diff
@@ -67,11 +67,8 @@ jobs:
         with:
           key: ${{ matrix.config }}-address-${{matrix.custom_hash_map}}
       - run: sudo apt-get -qq update && sudo apt-get install -y libwayland-dev xorg-dev
-      - run: python scripts/tests.py --build --config ${{ matrix.config }} --cmake='-DUSE_CUSTOM_HASH_MAP=${{matrix.custom_hash_map}}'
+      - run: python scripts/tests.py --build --config ${{ matrix.config }} --cmake='-DUSE_CUSTOM_HASH_MAP=${{matrix.custom_hash_map}} -DVVL_ENABLE_ASAN=ON'
```
spencer-lunarg, line 70: 1. Why change this here (out of scope of PR to mess with CI)
2. I am 99% sure this will not build any of the deps like VUL with ASAN, which is not what we want

arno-lunarg: It works, I have been using this flow locally for about a year now. I can move this change to its own PR, or maybe in its own commit no problem!

The way we currently enable ASAN in the 2 github actions is hackish, and would not work with MSVC for instance. We solved this problem with the `VVL_ENABLE_ASAN` option, so we should just use that.

The cmake command ran by the ASAN Github actions is the following:


> CICMD(['cmake', '-S', '/home/runner/work/Vulkan-ValidationLayers/Vulkan-ValidationLayers', '-B', '/home/runner/work/Vulkan-ValidationLayers/Vulkan-ValidationLayers/build-ci/vvl', '-D', 'CMAKE_BUILD_TYPE=release', '-D', 'BUILD_TESTS=ON', '-D', 'UPDATE_DEPS=ON', '-D', 'UPDATE_DEPS_DIR=/home/runner/work/Vulkan-ValidationLayers/Vulkan-ValidationLayers/build-ci/external', '-D', 'BUILD_WERROR=ON', '-D', 'INSTALL_VVL_TEST_ICD=ON', '-DUSE_CUSTOM_HASH_MAP=OFF', '-DVVL_ENABLE_ASAN=ON'], env=None)


It will enable ASAN for both VVL and its dependencies. 
Enabling for VVL is done in the root `CMakeLists.txt`:
```cmake
if (VVL_ENABLE_ASAN)
    add_compile_options(-fsanitize=address)
    if (NOT MSVC)
        add_link_options(-fsanitize=address)
    endif()
endif()
```
Dependencies will get it by going here in `scripts\CMakeLists.txt`: 
```cmake
    if (VVL_ENABLE_ASAN)
        list(APPEND update_dep_command "--asan")
    endif()
```

layers/gpuav/core/gpuav_settings.h
```diff
@@ -17,6 +17,15 @@
  */
 #pragma once
 #include <cstdint>
+// Fix GCC 13 issues with regex
```
spencer-lunarg, line 20: image

tests/unit/gpu_av_ray_query_positive.cpp
```diff
@@ -20,13 +20,13 @@
...
 
-void GpuAVRayQueryTest::InitGpuAVRayQuery() {
+void GpuAVRayQueryTest::InitGpuAVRayQuery(void *p_next /*= nullptr*/) {
     SetTargetApiVersion(VK_API_VERSION_1_2);
     AddRequiredExtensions(VK_KHR_RAY_QUERY_EXTENSION_NAME);
     AddRequiredFeature(vkt::Feature::rayQuery);
     AddRequiredFeature(vkt::Feature::accelerationStructure);
     AddRequiredFeature(vkt::Feature::bufferDeviceAddress);
-    RETURN_IF_SKIP(InitGpuAvFramework());
+    RETURN_IF_SKIP(InitGpuAvFramework(p_next));
```
spencer-lunarg, line 29: when you rebase, just need to pass in a `std::vector<VkLayerSettingEXT>` so we can add safe mode on (or not)

spencer-lunarg (approved):

layers/gpuav/core/gpuav_settings.h
arno-lunarg: (┬┬﹏┬┬)

PR #9859, opened 2025-04-04
corechecks: Fix bad instance cast

### LINKED ISSUE #9854: vkGetDeviceGroupSurfacePresentModesKHR crashes
Any call to `vkGetDeviceGroupSurfacePresentModesKHR()` seems to crash in `vku::concurrent::unordered_map`

```
TEST_F(PositiveWsi, GetDeviceGroupSurfacePresentModes) {
    AddSurfaceExtension();
    RETURN_IF_SKIP(Init());
    RETURN_IF_SKIP(InitSwapchain());

    VkDeviceGroupPresentModeFlagsKHR present_mode_flags;
    vk::GetDeviceGroupSurfacePresentModesKHR(m_device->handle(), m_surface.Handle(), &present_mode_flags);
}

```

jeremyg-lunarg:
CoreChecks::PreCallValidateGetDeviceGroupSurfacePresentModesKHR wants to call a validation method in the core::Instance class but it was using instance_state, which is the state tracker instead of in instance_proxy which is the core::Instance object. Bleah.

Fixes #9854

### DIFF
--- a/layers/core_checks/cc_wsi.cpp
+++ b/layers/core_checks/cc_wsi.cpp
@@ -1507,7 +1507,7 @@ bool CoreChecks::PreCallValidateGetDeviceGroupSurfacePresentModesKHR(VkDevice de
                                                                      VkDeviceGroupPresentModeFlagsKHR *pModes,
                                                                      const ErrorObject &error_obj) const {
     bool skip = false;
-    const auto *core_instance = reinterpret_cast<core::Instance *>(instance_state);
+    const auto *core_instance = static_cast<core::Instance *>(instance_proxy);
     if (device_state->physical_device_count == 1) {
         skip |= core_instance->ValidatePhysicalDeviceSurfaceSupport(
             physical_device, surface, "VUID-vkGetDeviceGroupSurfacePresentModesKHR-surface-06212", error_obj.location);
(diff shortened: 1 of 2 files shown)


ziga-lunarg (approved):

layers/core_checks/cc_wsi.cpp
```diff
@@ -1507,7 +1507,7 @@ bool CoreChecks::PreCallValidateGetDeviceGroupSurfacePresentModesKHR(VkDevice de
                                                                      VkDeviceGroupPresentModeFlagsKHR *pModes,
                                                                      const ErrorObject &error_obj) const {
     bool skip = false;
-    const auto *core_instance = reinterpret_cast<core::Instance *>(instance_state);
+    const auto *core_instance = dynamic_cast<core::Instance *>(instance_proxy);
```
spencer-lunarg, line 1510: image

spencer-lunarg (approved):

layers/core_checks/cc_wsi.cpp
artem-lunarg, line 1510: If result of dymamic_cast is not checked because we know the type then it's the case for:
image

jeremyg-lunarg: well clearly we didn't know the type in the original code and dynamic_cast will at least return null

artem-lunarg (approved):

layers/core_checks/cc_wsi.cpp
jeremyg-lunarg: c++ casts are the worst. I guess I will change this to static_cast so I stop getting memed

Issue #9725, opened 2025-03-20
Eliminate Buffer Device Address instrumentation with unsafe mode


spencer-lunarg:
Found in a trace where it is

```
Shader 329
BufferDeviceAddressPass instrumentation count: 522
CreateGraphicsPipelines: 4.84878 seconds
```

And this could be reduced when using Unsafe Mode as all 522 usages are pointing to a single struct

spencer-lunarg:
Example test to help showcase simple example

```
TEST_F(PositiveGpuAVBufferDeviceAddress, x) {
    RETURN_IF_SKIP(InitGpuVUBufferDeviceAddress());

    char const *shader_source = R"glsl(
        #version 450
        #extension GL_EXT_buffer_reference : enable

        layout(buffer_reference) buffer BDA {
            vec3 x;
            vec3 payload[4096];
        };

        layout(push_constant) uniform Uniforms {
            BDA ptr;
        };

        void main() {
            vec3 a = vec3(0);
            a += fma(vec3(ptr.payload[0].x, ptr.payload[0].y, ptr.payload[0].z),
                        vec3(ptr.payload[1].x, ptr.payload[1].y, ptr.payload[1].z),
                        vec3(ptr.payload[2].x, ptr.payload[2].y, ptr.payload[2].z));
            ptr.x = a;
        }
    )glsl";
    VkPushConstantRange pc_range = {VK_SHADER_STAGE_COMPUTE_BIT, 0, sizeof(VkDeviceAddress)};
    const vkt::PipelineLayout pipeline_layout(*m_device, {}, {pc_range});

    CreateComputePipelineHelper pipe(*this);
    pipe.cp_ci_.layout = pipeline_layout;
    pipe.cs_ = std::make_unique<VkShaderObj>(this, shader_source, VK_SHADER_STAGE_COMPUTE_BIT);
    pipe.CreateComputePipeline();
}
```

PR #9863, opened 2025-04-04
gpuav: Reduce BDA when statically can detect range

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9725

We found shaders that would look "like"

```glsl
layout(buffer_reference) buffer BDA {
    uint x[4096];
};

void main() {
    BDA ptr;
    ptr.x[4] = ptr.x[8] + ptr.x[12]
}
```

and our check would look like

```glsl
inst_bda(ptr + 6, 4);
inst_bda(ptr + 8, 4);
inst_bda(ptr + 12, 4);
ptr.x[6] = ptr.x[8] + ptr.x[12]
```

and would easily get 500+ instrumentation in a single shader, but just getting the being/end range statically it will now look like

```glsl
inst_bda(ptr + 6, 28); // covers range from x[4] to x[12]
ptr.x[6] = ptr.x[8] + ptr.x[12]
```

I got `PositiveGpuAVBufferDeviceAddress.Stress` down from 2 minutes to 0.3 seconds with this

I have a fossilize of the 5 worst pipeline in a large trace, it went from single-threaded 15.2 seconds down to 3 seconds

### DIFF
--- a/layers/gpuav/spirv/buffer_device_address_pass.cpp
+++ b/layers/gpuav/spirv/buffer_device_address_pass.cpp
@@ -47,7 +47,7 @@ uint32_t BufferDeviceAddressPass::CreateFunctionCall(BasicBlock& block, Instruct
     const uint32_t address_id = module_.TakeNextId();
     block.CreateInstruction(spv::OpConvertPtrToU, {uint64_type.Id(), address_id, pointer_id}, inst_it);
 
-    const Constant& length_constant = module_.type_manager_.GetConstantUInt32(meta.type_length);
+    const uint32_t access_size_id = module_.type_manager_.GetConstantUInt32(meta.access_size).Id();
     const uint32_t opcode = meta.target_instruction->Opcode();
 
     uint32_t access_type_value = 0;
@@ -63,12 +63,35 @@ uint32_t BufferDeviceAddressPass::CreateFunctionCall(BasicBlock& block, Instruct
     const uint32_t inst_position = meta.target_instruction->GetPositionIndex();
     const uint32_t inst_position_id = module_.type_manager_.CreateConstantUInt32(inst_position).Id();
 
-    const uint32_t function_range_result = module_.TakeNextId();
+    uint32_t function_range_result = 0;  // only take next ID if needed
     const uint32_t function_range_id = GetLinkFunction(function_range_id_, kOfflineFunctionRange);
-    block.CreateInstruction(
-        spv::OpFunctionCall,
-        {bool_type, function_range_result, function_range_id, inst_position_id, address_id, access_type.Id(), length_constant.Id()},
-        inst_it);
+
+    if (module_.settings_.safe_mode || block_skip_list_.find(inst_position) == block_skip_list_.end()) {
+        // "normal" check
+        function_range_result = module_.TakeNextId();
+        block.CreateInstruction(
+            spv::OpFunctionCall,
+            {bool_type, function_range_result, function_range_id, inst_position_id, address_id, access_type.Id(), access_size_id},
+            inst_it);
+    } else {
+        // Find if this is the lowest pointer access in the struct
+        for (const auto& [struct_id, range] : block_struct_range_map_) {
+            // This is only for unsafe mode, so we can ignore all other instructions
+            if (range.min_instruction != inst_position) {
+                continue;
+            }
+
+            function_range_result = module_.TakeNextId();
+            // If there is only a single access found, range diff is zero and this becomes a "normal" check automatically
+            const uint32_t full_access_range = (range.min_struct_offsets - range.max_struct_offsets) + meta.access_size;
+            const uint32_t full_range_id = module_.type_manager_.GetConstantUInt32(full_access_range).Id();
+            block.CreateInstruction(spv::OpFunctionCall,
+                                    {bool_type, function_range_result, function_range_id, inst_position_id, address_id,
+                                     access_type.Id(), full_range_id},
+                                    inst_it);
+            break;
+        }
+    }
 
     const Constant& alignment_constant = module_.type_manager_.GetConstantUInt32(meta.alignment_literal);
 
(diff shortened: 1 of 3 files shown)


layers/gpuav/spirv/buffer_device_address_pass.cpp
```diff
@@ -64,12 +65,33 @@ uint32_t BufferDeviceAddressPass::CreateFunctionCall(BasicBlock& block, Instruct
...
+        function_range_result = module_.TakeNextId();
+        block.CreateInstruction(
+            spv::OpFunctionCall,
+            {bool_type, function_range_result, function_range_id, inst_position_id, address_id, access_type.Id(), access_size_id},
+            inst_it);
+    } else {
+        // Find if this is the lowest pointer access in the struct
+        for (const auto& [struct_id, range] : block_struct_range_map_) {
+            // This is only for unsafe mode, so we can ignore all other instructions
+            if (range.begin_instruction != inst_position) continue;
```
arno-lunarg, line 82:
```suggestion
            if (range.begin_instruction != inst_position) {
            continue;
            }
```

layers/gpuav/spirv/buffer_device_address_pass.cpp
```diff
@@ -169,6 +191,50 @@ bool BufferDeviceAddressPass::Instrument() {
...
+                            if (struct_offset == 0) continue;
+                            uint32_t inst_position = meta.target_instruction->GetPositionIndex();
+                            block_skip_list_.insert(inst_position);
+
+                            Range& range = block_struct_range_map_[struct_type->Id()];
+                            if (struct_offset < range.begin) {
+                                range.begin_instruction = inst_position;
+                                range.begin = struct_offset;
+                            }
+                            range.end = std::max(range.end, struct_offset);
```
arno-lunarg, line 232:
```suggestion
                            if (struct_offset < range.begin) {
                                range.begin_instruction = inst_position;
                                range.begin = struct_offset;
                            }
                            range.end = std::max(range.end, struct_offset);
```
both `range.begin` and `range.end` can end up equaling `struct_offset`, looks weird? I expected `range.end` to be something like `struct_offset + some_size`

arno-lunarg:
We need tests for that new option

spencer-lunarg:
>  We need tests for that new option

That is `PositiveGpuAVBufferDeviceAddress.Stress`

layers/gpuav/spirv/buffer_device_address_pass.cpp
spencer-lunarg: its the range of offsets

if you have

```
layout(buffer_reference) buffer BDA {
    uint x[4096];
};

void main() {
    BDA ptr;
    ptr.x[0] = 0
}
```

the Range begin/end would be the same, it works out because we only use the `begin` to really know where the `begin_instruction` is, and we use the `end` to know the actual length/range the check

will add a comment that `begin` here is not actually used other then to find `begin_instruction`

layers/gpuav/spirv/buffer_device_address_pass.h
```diff
@@ -34,17 +35,26 @@ class BufferDeviceAddressPass : public Pass {
...
+    uint32_t CreateFunctionCall(BasicBlock& block, InstructionIt* inst_it, const InstructionMeta& meta);
 
     // Function IDs to link in
     uint32_t function_range_id_ = 0;
     uint32_t function_align_id_ = 0;
+
+    struct Range {
+        uint32_t begin_instruction = 0;
+        uint32_t begin = vvl::kU32Max;
+        uint32_t end = 0;
```
arno-lunarg, line 54:
```suggestion
        uint32_t begin = vvl::kU32Max;
        uint32_t end = 0;
```
Given how these are used, they probably should be named something like `min/max_struct_offsets`?

spencer-lunarg: sure, `min`/`max` probably better

arno-lunarg:
> > We need tests for that new option
> 
> That is `PositiveGpuAVBufferDeviceAddress.Stress`

This test is disabled, so we don't test the unsafe path. Shouldn't most of the buffer address tests "just work" for both modes? If so, then those tests should run for both of them, you can come up with something using the Gtest `INSTANTIATE_TEST_SUITE_P` feature

layers/gpuav/spirv/buffer_device_address_pass.h
```diff
@@ -34,17 +35,26 @@ class BufferDeviceAddressPass : public Pass {
...
+    uint32_t CreateFunctionCall(BasicBlock& block, InstructionIt* inst_it, const InstructionMeta& meta);
 
     // Function IDs to link in
     uint32_t function_range_id_ = 0;
     uint32_t function_align_id_ = 0;
+
+    struct Range {
+        uint32_t min_instruction = 0;  // used to only instrument at the lowest offset
+        uint32_t min_struct_offsets = vvl::kU32Max;
+        uint32_t max_struct_offsets = 0;
```
arno-lunarg, line 54:
```suggestion
    struct StructOffsets? Range is too generic {
        uint32_t min_instruction = 0;  // used to only instrument at the lowest offset
        uint32_t min_struct_offset = vvl::kU32Max;
        uint32_t max_struct_offset = 0;
```

spencer-lunarg: what do you actually want, we need this to PR to get perf up in real traces and we have spent a lot of time bike-shedding this name... `StructOffsets::min_struct_offset` is not ideal

```
struct StructOffsets {
   uint min;
   uint max;
}
```

good enough?

This `Range` is inside the BDA Pass only, it is only isolated in a small code, it just has a simple job, track the lowest and highest access of a pointer from in a BDA Struct

it is already in a map call `block_struct_range_map_`... didn't want `block_struct_struct_offset_map_` originally

arno-lunarg (approved):

PR #9871, opened 2025-04-07
chassis: Fix Device and Instance destroy order

jeremyg-lunarg:
Make sure the state tracker runs last in vkDestroyDevice() and vkDestroyInstance().

Make sure syncval cleans up its custom state tracking in vkDestroyDevice().

### DIFF
--- a/layers/sync/sync_validation.cpp
+++ b/layers/sync/sync_validation.cpp
@@ -755,6 +755,15 @@ void SyncValidator::FinishDeviceSetup(const VkDeviceCreateInfo *pCreateInfo, con
     text::ToLower(debug_cmdbuf_pattern);
 }
 
+void SyncValidator::PreCallRecordDestroyDevice(VkDevice device, const VkAllocationCallbacks *pAllocator,
+                                               const RecordObject &record_obj) {
+    queue_sync_states_.clear();
+    binary_signals_.clear();
+    timeline_signals_.clear();
+    waitable_fences_.clear();
+    host_waitable_semaphores_.clear();
+}
+
 void SyncValidator::PostCallRecordCreateSemaphore(VkDevice device, const VkSemaphoreCreateInfo *pCreateInfo,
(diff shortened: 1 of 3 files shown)


spencer-lunarg (approved):
LGTM - but want @artem-lunarg confirmation as it is touching his turf

layers/sync/sync_validation.cpp
```diff
@@ -755,6 +755,15 @@ void SyncValidator::FinishDeviceSetup(const VkDeviceCreateInfo *pCreateInfo, con
     text::ToLower(debug_cmdbuf_pattern);
 }
 
+void SyncValidator::PreCallRecordDestroyDevice(VkDevice device, const VkAllocationCallbacks *pAllocator,
+                                               const RecordObject &record_obj) {
+    queue_sync_states_.clear();
```
artem-lunarg, line 760: Any chance this object SyncValidator will be used again, so CreateDevice will start new life cycle? (mostly thinking if queue_id_limit_ should also be reset). 

What's the purpose of reseting state?

artem-lunarg (approved):
LGTM, some questions.

layers/sync/sync_validation.cpp
jeremyg-lunarg: after vkDestroyDevice() this SyncValidator object should be c++ deleted. But there's possible ordering problems with the state tracker destruction.

Resetting the various state maps gets rid of some shared_ptrs that were causing vvl::Queue objects to be destroyed too late. See the stack trace in the issue. Probably only queue_states_ really needs to be cleared to avoid the bug, but better safe than sorry. The state tracker clears all of its object maps in this method as well.

artem-lunarg: > Resetting the various state maps gets rid of some shared_ptrs that were causing vvl::Queue objects to be destroyed too late. 

That's a subtle point

Issue #9852, opened 2025-04-04
Replace assert in gpuav_state_tracker.cpp


ziga-lunarg:
There are a few CTS tests hitting: assert(action_command_count < glsl::kMaxActionsPerCommandBuffer);

Some of the tests are:
dEQP-VK.api.command_buffers.record_many_draws_primary_1
dEQP-VK.dgc.ext.compute.misc.execute_many_8192_universal_queue
dEQP-VK.memory.pipeline_barrier.host_write_uniform_buffer.1048576

spencer-lunarg:
We could write a test to hit this limit, agree we don't need an assert, just 1 warning after they hit the limit to let them know that errors will not be tracked properly for those draws/dispatches

PR #9873, opened 2025-04-07
gpuav: Warn on reaching max actions cmd validation

arno-lunarg:
Closes https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9852

Actually nothing was done when reaching this limit, so I added the logic to skip validation when it is reached. It was more cumbersome to do than what I would like it to be...
The warning:

Validation Warning: [ GPU-AV::Max action per command buffer reached ] | MessageID = 0x275e89d
vkCmdDraw(): Reached maximum validation commands count for command buffer ( 8192 ). No more draw/dispatch/trace rays commands will be validated inside this command buffer.
Objects: 1
    [0] VkCommandBuffer 0x1eaf1c58960

### DIFF
--- a/layers/gpuav/core/gpuav_record.cpp
+++ b/layers/gpuav/core/gpuav_record.cpp
@@ -378,14 +378,20 @@ void Validator::PreCallRecordCmdBindDescriptorBufferEmbeddedSamplers2EXT(
 // Common logic before any draw/dispatch/traceRays
 void Validator::PreCallActionCommand(Validator &gpuav, CommandBufferSubState &cb_state, VkPipelineBindPoint bind_point,
                                      const Location &loc) {
+    if (cb_state.max_actions_cmd_validation_reached_) {
+        return;
+    }
     PreCallSetupShaderInstrumentationResources(gpuav, cb_state, bind_point, loc);
 }
 
 // Common logic after any draw/dispatch/traceRays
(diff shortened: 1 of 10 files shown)


layers/gpuav/core/gpuav_record.cpp
```diff
@@ -378,14 +378,20 @@ void Validator::PreCallRecordCmdBindDescriptorBufferEmbeddedSamplers2EXT(
...
 
 // Common logic after any draw/dispatch/traceRays
 void Validator::PostCallActionCommand(Validator &gpuav, CommandBufferSubState &cb_state, VkPipelineBindPoint bind_point,
                                       const Location &loc) {
+    if (cb_state.max_actions_cmd_validation_reached_) {
+        return;
+    }
     PostCallSetupShaderInstrumentationResources(gpuav, cb_state, bind_point, loc);
-    cb_state.IncrementCommandCount(bind_point);
+    cb_state.max_actions_cmd_validation_reached_ = !cb_state.IncrementCommandCount(gpuav, bind_point, loc);
```
spencer-lunarg, line 394: `IncrementCommandCount` in a `CommandBufferSubState` function, can't it just set `max_actions_cmd_validation_reached_` instead of having it return a bool here

spencer-lunarg (approved):
LGTM - But think `IncrementCommandCount` can just be `void` to make things simpler

layers/gpuav/core/gpuav_record.cpp
arno-lunarg: yeah true will do!

